PR #4420, opened 2025-02-27
Add multi-framework examples to Astro site

diegohaz:
This PR adds the ability to render multi-framework examples on the Astro site through mdx files and Astro content collections.


site/astro.config.ts
```diff
@@ -1,10 +1,13 @@
+import mdx from "@astrojs/mdx";
 import react from "@astrojs/react";
 import solid from "@astrojs/solid-js";
 import tailwindcss from "@tailwindcss/vite";
 import { defineConfig } from "astro/config";
 
 // https://astro.build/config
 export default defineConfig({
+  srcDir: "src",
```
diegohaz, line 9: `src` is the default value. I included it as a reminder because in the future I plan to move the site folder to the root and rename `src` to `site`.

biome.json
```diff
@@ -11,6 +11,7 @@
       "website/.next/**",
       "website/.pages/**",
       "site/.astro/**",
+      "**/*.astro",
```
diegohaz, line 14: Biome doesn't work on full Astro files, only on the frontmatter (JavaScript) section. However, it conflicts with the Astro formatter there, so I'm disabling it for now.

site/src/styles/global.css
```diff
@@ -25,22 +25,22 @@ body {
 
 /** ak-focusable **/
 
-@utility ak-focusable.idle {
+@utility ak-focusable_idle {
```
diegohaz, line 28: Tailwind apparently doesn't recognize inline utility classes with a dot `.`

site/src/tsconfig.react.json
diegohaz, line 1: I'm moving this file to the `src` folder because I plan to relocate the site folder to the root and rename `src` to `site`, so these tsconfig files need to be there (they can't be on root).

Issue #4310, opened 2024-12-04
How can I disable a native <select>?


diegohaz:
### Discussed in https://github.com/ariakit/ariakit/discussions/4308

<div type='discussions-op-text'>

<sup>Originally posted by **wonu** December  4, 2024</sup>
I'm using a `<Select>` component rendered as a native element.
I want to prevent this element from being submitted with the form, but the `disabled` prop isn't working as expected.

```ts
<Ariakit.Select name="fruit" disabled /> // `disabled` attribute is not being set on the <select> element
```

Can you explain why and suggest a solution?</div>

PR #4506, opened 2025-03-31
[Select] Fix: prevent native select value from submitting when Select is disabled

alisasanib:
Fixes: #4310

**Problem:**

When the `Select` component was rendered with the `disabled` prop, the hidden native `<select>` element used for form submission was still enabled. As a result, the form would submit a value for the disabled field, which is inconsistent with native HTML form behavior.

**Fix:**

This PR ensures that the native <select> element receives the disabled attribute, preventing its value from being included in the form submission when it is disabled.

A working example and tests have also been added to demonstrate and verify the fix.
Tests cover disabled behavior, keyboard interaction, and correct form submission logic.

### DIFF
--- a/packages/ariakit-react-core/src/select/select.tsx
+++ b/packages/ariakit-react-core/src/select/select.tsx
@@ -77,7 +77,6 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
   moveOnKeyDown = true,
   toggleOnPress = true,
   toggleOnClick = toggleOnPress,
-  disabled,
   ...props
 }) {
   const context = useSelectProviderContext();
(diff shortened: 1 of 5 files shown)


packages/ariakit-react-core/src/select/select.tsx
```diff
@@ -202,6 +202,7 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
             name={name}
             form={form}
             required={required}
+            disabled={props.disabled}
```
diegohaz, line 205: I believe this should be added to the dependencies array.

alisasanib: Done!

diegohaz (approved):
Thanks @alisasanib!

PR #4616, opened 2025-05-01
Fix Solid checkbox dynamic store behaviour

alisasanib:
- Problem

When a component receives a dynamic store prop, Solid reactivity doesn't automatically respond to changes in the accessor. This causes stale state in components like Checkbox, where checked remains stuck even after the store changes. The underlying issue is that useStoreState subscribes to the initial store, but doesn't track the updates to the store

- Expected behaviour

When the store prop changes, the component should react to the new store, updating internal values like checked and aria-checked as needed.

- Fix

Introduced a utility to wrap useStoreState in a createMemo, which ensures that any change to the dynamic store accessor causes the value to be re-evaluated. This guarantees the component stays in sync with the current store.
Also added an effect inside useStore to detect changes in the store prop and update the internal signal accordingly.


DaniGuardiola (approved):
as per my comment in https://github.com/ariakit/ariakit/pull/4561#pullrequestreview-2947771383, we'll need to revisit sotres in solid but this unblocks progress in the solid port so approving. thank you very much, super helpful!

PR #4637, opened 2025-05-18
[Doc]: fix contribution guide link on solid

alisasanib:
As I started looking into porting new components to Solid, I noticed that the contribution guide link points to the main guide instead of the Solid-specific one 🙂


diegohaz (approved):
Thanks! 🥳

PR #4860, opened 2025-07-28
Improve open performance for long comboboxes by ~30%

iamakulov:
When profiling why opening a `Combobox` with 100+ items feels laggy (it’s 100 list items, how slow could it be?), I noticed that `Focusable` (which is used by `Combobox` and a bunch of other things) installs global event listeners every time it mounts. With 100 items in a list, it’s 100 focusables – and, therefore, 200 `addGlobalEventListener` calls. This ends up taking ~30% of the total opening time:

Screenshot 2025-07-28 at 15 18 23

This PR fixes this by introducing a global flag in `focusable.tsx` that prevents calling `addGlobalEventListener` more than once.

### DIFF
--- a/.changeset/sad-pants-shout.md
+++ b/.changeset/sad-pants-shout.md
@@ -1,5 +1,8 @@
 ---
 "@ariakit/react-core": patch
+"@ariakit/react": patch
 ---
 
-Improve mount performance of Focusable
+Improved Combobox performance
+
+Thanks to [@iamakulov](https://github.com/iamakulov), the [Combobox](https://ariakit.org/components/combobox) component now opens ~30% faster by removing unnecessary calls to an internal function that adds global event listeners. See the [pull request](https://github.com/ariakit/ariakit/pull/4860) for more details.
--- a/.changeset/sad-pants-shout.md
+++ b/.changeset/sad-pants-shout.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-core": patch
+---
+
+Improve mount performance of Focusable
--- a/packages/ariakit-core/src/utils/events.ts
+++ b/packages/ariakit-core/src/utils/events.ts
@@ -1,4 +1,5 @@
 import { contains } from "./dom.ts";
+import { shallowEqual } from "./misc.ts";
 import { isApple } from "./platform.ts";
 
 /**
@@ -223,7 +224,11 @@ export function addGlobalEventListener(
 
   // Prevent errors from "sandbox" frames.
   try {
-    scope.document.addEventListener(type, listener, options);
+    if (!hasEventListenerBeenAdded(type, listener, options, scope)) {
+      scope.document.addEventListener(type, listener, options);
+      markEventListenerAsAdded(type, listener, options, scope);
+    }
+
     for (const frame of Array.from(scope.frames)) {
       children.push(addGlobalEventListener(type, listener, options, frame));
     }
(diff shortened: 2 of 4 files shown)


iamakulov:
> Could you please profile the updated code again and post the difference here? We'd like to confirm the percentage of improvement from this change, if any.

Sure thing! I just measured this a bit more precisely. Comparing v0.4.7 vs this PR across 3 combobox opens, the relative cost of `addGlobalEventListener` seems to go down from 33-35% to 23-27%:

Screenshot 2025-07-28 at 17 04 30

Traces also seem to show that the `addEventListener` line is the bottleneck:

Screenshot 2025-07-28 at 17 11 23

I’m guessing even if the browser ignores an event listener that was already added, it doesn’t short-circuit the execution of `addEventListener`, so all these calls still end up being expensive.

iamakulov:
That said, I realized I didn’t optimize this _enough_. This PR removes repeated `addEventListener` calls for `useFocusable`, but there are a few other places that call it over and over:

Label | `v0.4.7` | 84e9e23c
-- | -- | --
useFocusable → onGlobalMouseDown | 310 | -
useFocusable → onGlobalKeyDown | 310 | -
useIsMouseMoving → mousemove setMouseMoving | 295 | 300
useIsMouseMoving → mousedown resetMouseMoving | 295 | 300
useIsMouseMoving → mouseup resetMouseMoving | 295 | 300
useIsMouseMoving → keydown resetMouseMoving | 295 | 300
useIsMouseMoving → scroll resetMouseMoving | 295 | 300
useEventOutside → onEvent | 15 | 15
usePreviousMouseDownRef → onMouseDown | 5 | 5
useHovercard → onMouseMove | 5 | 5
useHovercard → mouseenter disableEvent | 5 | 5
useHovercard → mouseover disableEvent | 5 | 5
useHovercard → mouseout disableEvent | 5 | 5
useHovercard → mouseleave disableEvent | 5 | 5
useDialog → onKeyDown | 5 | 5

(These numbers come from a combobox with 59 items. The page has one main frame and four subframes. The numbers are recorded from a second or third [i.e. not first] opening of the combobox.)

I just pushed a change (95e1b31a76df3b7ea9b6d148e33bae504116943f) where I move this optimization one level lower (into `addGlobalEventListener`). I don’t like how complex it is, and I don’t like that it won’t work for places which create a new arrow function every time, but let’s see how well it perform in profiling.

iamakulov:
Meh, not a fan of https://github.com/ariakit/ariakit/commit/95e1b31a76df3b7ea9b6d148e33bae504116943f, even with https://github.com/ariakit/ariakit/pull/4860/commits/ed7e79d17537f0ff21fbc6b3e9b5aeb1a46bfbb3 on top. (To test: `npm i https://pkg.csb.dev/ariakit/ariakit/commit/ed7e79d1/@ariakit/react`.) Even with a near-perfect cache hit rate, `addGlobalEventListener` still takes 30%+ of total costs. I can’t narrow this down to any specific line that would be a bottleneck; I’m guessing there’s just an inherent overhead of calling `addGlobalEventListener` ~2000 times in a row?

diegohaz:
That's amazing work! Thank you!

I'll update the changelog to credit you. What's the best link for me to use (GitHub, Twitter, personal website, etc.)?

iamakulov:
Oh yeah, GitHub is okay! Thank you :)

PR #4960, opened 2025-08-28
fix PopoverArrow borderWidth to account for device pixel ratio

### LINKED ISSUE #4954: PopoverArrow borderWidth is 0 when getPropertyValue returns subpixel value for border-width: 1px
### Current behavior

PopoverArrow borderWidth calculations use style.getPropertyValue which based on the users display pixel ratio and browser zoom scale can return a sub pixel string value which currently get's converted to an integer, causing 0.xx subpixel values to be 0.

https://github.com/ariakit/ariakit/blob/f29fc76b3558cbe91bba973b0284355744d7e7d6/packages/ariakit-react-core/src/popover/popover-arrow.tsx#L108C1-L109C1

1. Go to https://codesandbox.io/p/sandbox/elegant-http-rrp89j?file=%2Findex.js%3A15%2C35
2. Set browser zoom to 125%
3. Observe the PopoverArrow has a stroke-width: 0

### Expected behavior

PopoverArrow should have stroke-width: 1

### Workaround

- set the `borderWidth` prop to desired value
- override in CSS with stroke-width: value !important

These all rely on explicitly setting the value when it's nicer to infer

### Possible solutions

Instead of using `parseInt` on the computed getPropertyValue, using Math.ceil(parseFloat(...)) could do the trick

mattywong:
closes #4954

### DIFF
--- a/.changeset/mean-coats-study.md
+++ b/.changeset/mean-coats-study.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed the automatic border width on [`PopoverArrow`](https://ariakit.org/reference/popover-arrow) to account for the device pixel ratio.


diegohaz (approved):
Thank you @mattywong!

PR #4961, opened 2025-08-29
set various array values to readonly by default

cloud-walker:
By experience on typescript if not strictly necessary its better to define arrays as readonly, especially as generics, to guarantee statically that the array is not mutated, this would make the modules compatible with both readonly and non readonly values.

But I don't have much experience with the codebase, so treat this PR as an experiment, and a discussion if you want


diegohaz (approved):
Thanks @cloud-walker!

PR #5473, opened 2026-03-02
Rewrite Ariakit Tailwind package

diegohaz:
**Motivation**

`@ariakit/tailwind` needs a cleaner v0.2 foundation for relative layer, text, edge, outline, and frame utilities. The v0.1 plugin was difficult to evolve consistently across light, dark, and high-contrast contexts, and the site needs to consume the rewritten utility names before the package can move forward.

**Solution**

Rewrites the package around a TypeScript CSS generator that emits the public `src/output.css` export and replaces the old handwritten plugin files. The new utilities cover layer, state, ink, text, edge, outline, frame, badge, and related color/contrast behavior.

Updates the site styles, examples, templates, generated style metadata, screenshots, and OG images to use the new utility set. Adds the `ariakit-tailwind` sandbox with WCAG checks, computed-style snapshots, and a perf test harness, plus refreshed README and v0.2 migration docs.

Recent review follow-ups also moved `@axe-core/playwright` into the `site` workspace, removed the scratch `/ak` route, migrated remaining stale edge classes, and cleaned up generated Tailwind output for frame row naming, arbitrary frame modifiers, text backgrounds, and rounded range boundaries.

**Verification**

- `pnpm lint`
- `pnpm tsc`
- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag`


packages/ariakit-tailwind/package.json
```diff
@@ -29,17 +29,15 @@
   "type": "module",
   "sideEffects": false,
   "exports": {
-    ".": "./src/index.css",
+    ".": "./src/output.css",
```
diegohaz, line 32: `pnpm changeset status --since=origin/main` currently fails with "Some packages have been changed but no changesets were found." This rewrite changes the public `@ariakit/tailwind` CSS export, removes/renames utilities, and adds v0.2 migration docs, so without a changeset the release workflow will not version/publish the package or generate a changelog entry for users. Please add a changeset for `@ariakit/tailwind` (or an explicit empty changeset only if this is intentionally unreleased, which does not match the shipped docs here).

site/src/pages/plus/checkout/[...checkout].astro
```diff
@@ -209,11 +209,11 @@ async function getOfferText() {
                     href={price.active ? undefined : price.path}
                     aria-disabled={price.active}
                     aria-current={price.selected ? "page" : undefined}
-                    class="ak-button ak-frame-card ak-layer-down-0.5 aria-disabled:ak-layer-down-0.5 aria-[current]:ak-layer aria-[current]:ring aria-[current]:shadow-lg aria-[current]:ak-edge-contrast-primary flex flex-col gap-2 items-start"
+                    class="ak-button ak-frame ak-frame-card/card ak-layer ak-layer-darken-3 aria-disabled:ak-layer-darken-3 aria-[current]:ring aria-[current]:shadow-lg aria-[current]:ak-edge-contrast-primary flex flex-col gap-2 items-start"
```
Comment (diegohaz) [inline site/src/pages/plus/checkout/[...checkout].astro]: `ak-edge-contrast-primary` is no longer generated by the rewritten package (the migration guide maps it to `ak-edge-primary ak-edge-raw`), so this selected-price state now no-ops instead of changing the edge color. I also found the same stale class in `site/src/components/playground.react.tsx`. Please migrate these remaining usages to the new `ak-edge-*` combination, for example `aria-[current]:ak-edge-primary aria-[current]:ak-edge-raw` if the old solid primary edge is still intended.

package.json
```diff
@@ -69,6 +69,7 @@
     "@ariakit/solid-core": "workspace:*",
     "@ariakit/tailwind": "workspace:*",
     "@ariakit/test": "workspace:*",
+    "@axe-core/playwright": "4.11.1",
```
diegohaz: The new `@axe-core/playwright` import lives in the `site` workspace (`site/src/sandbox/ariakit-tailwind/test-chrome.ts`), but the dependency is declared only in the root package. Ariakit's workspace policy is that each workspace declares the dependencies it uses; otherwise filtered or isolated `site` installs rely on root hoisting and can fail to resolve this test import. Please move `@axe-core/playwright` to `site/package.json` `devDependencies` and remove the root entry unless root-level code also uses it.

site/src/pages/ak.astro
```diff
@@ -0,0 +1,537 @@
...
+import Meta from "#app/components/meta.astro";
+import "./ak.css";
+---
+
+<!doctype html>
+<html lang="en" transition:animate="none">
+  <head>
+    <Meta />
+  </head>
+  <body class="p-4 ak-layer ak-layer-canvas ak-layer-20_ [--contrast:90]_">
```
diegohaz: Adding this file under `site/src/pages` publishes a production `/ak` route. The contents look like a local scratchpad rather than a site page: disabled class tokens such as `ak-layer-20_` / `[--contrast:90]_`, an invalid-looking `m-` class, and placeholder copy like `dsada` / `lol` later in the file. Since Astro routes every file in `pages`, this will ship unless it is removed or moved into the `site/src/sandbox/ariakit-tailwind` preview/test area.

diegohaz (commented):
Initial review of PR #5473 at `e31a0291fa1c1ce96862b4ff291548d56e3e2d6b`, diffed from merge base/base `380e356daf1bc61c4cabc1ebd82662dbc6fa0fdd` (`origin/main`). I reviewed the rewritten `@ariakit/tailwind` package source/generated CSS, package metadata, migration/readme docs, the new Tailwind sandbox and tests, site style migration, registry/style dependency plumbing, and representative migrated site components/routes.

I left four inline findings:

- missing changeset for the public Tailwind package rewrite;
- stale removed `ak-edge-contrast-primary` usage in checkout/playground;
- accidental-looking public `/ak` scratch route;
- `@axe-core/playwright` declared in the wrong workspace.

Validation performed locally:

- `pnpm install`
- `pnpm -F @ariakit/tailwind build` and verified `src/output.css` stayed unchanged
- `pnpm exec tsc -p packages/ariakit-tailwind/tsconfig.json --noEmit`
- `pnpm test styles-test stackblitz-test source-plugin-test`
- `pnpm -F site build-styles` and verified `styles.json` stayed unchanged
- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag`
- `pnpm changeset status --since=origin/main` (fails because no changeset was added)

Local `nvm use` is unavailable in this shell, but the active Node is `v24.14.1`, matching the repo `.nvmrc` (`24`). `pnpm -F @ariakit/tailwind lint` is locally blocked before package linting by an `oxlint` config compatibility error, while the PR's GitHub `Lint` check is green. I also checked the PR status rollup; all GitHub checks are currently successful.

packages/ariakit-tailwind/migrating-to-v02.md
```diff
@@ -0,0 +1,454 @@
+# Migrating to v0.2
```
diegohaz, line 1: No changeset has been added for this PR (`.changeset/` only contains the boilerplate `config.json`, `changelog.cjs`, `README.md`, `get-changelog-entry-test.ts`), and `packages/ariakit-tailwind/package.json` still reads `"version": "0.1.11"`.

`@ariakit/tailwind` is **not** in the changeset `ignore` list (`.changeset/config.json`):

```json
"ignore": ["website", "site", "examples", "nextjs", "@ariakit/scripts", "template-*"]
```

so the changesets workflow expects an entry whenever shipped behavior changes — and this PR ships a near-total rewrite (27,567 additions / 3,463 deletions) with this very document cataloguing the breaking renames. Without a changeset:

- The version will not bump on release, so consumers pinning `^0.1.11` will not pick up the rewrite.
- `packages/ariakit-tailwind/CHANGELOG.md` will not get a v0.2 entry, leaving this `migrating-to-v02.md` file as the only signal that anything changed.

Please add a changeset before merging:

```sh
pnpm changeset
```

and pick `minor` for `@ariakit/tailwind` (or `major` if you want the SemVer signal even at the 0.x stage — the readme says "Experimental — breaking changes may happen in patch/minor", but the changeset entry is still what drives the version selection and CHANGELOG generation).

packages/ariakit-tailwind/src/input.ts
```diff
@@ -0,0 +1,2202 @@
...
+  outlineCMin: _ak.prop("outline-chroma-min", { initial: 0 }),
+  outlineCMax: _ak.prop("outline-chroma-max", vars.chromaP3Max),
+  outlineH: _ak.prop("outline-hue"),
+  frameRadius: _ak.prop.len("frame-radius", { initial: "0px" }),
+  framePadding: _ak.prop.len("frame-padding", { initial: "0px" }),
+  frameMargin: _ak.prop.len("frame-margin", { initial: "0px" }),
+  frameBorder: _ak.prop.len("frame-border", { initial: "0px" }),
+  frameRing: _ak.prop.len("frame-ring", { initial: "0px" }),
+  frameBordering: _ak.prop.len("frame-bordering", { initial: "0px" }),
+  frameRow: _ak.prop.number("frame-col", { initial: 0, inherits: true }),
```
diegohaz: The TS field is named `frameRow` but the registered CSS custom property is `--_ak-frame-col`, and the surrounding `*Row*` context vars (`frameParentRowContext` at line 592, `parentRow` at line 2111) treat the value as a "is row?" flag. The two utilities below then set this same property to `1` for `ak-frame-row` and `0` for `ak-frame-col`:

```ts
// lines 2129–2130
utility("frame-row", set(inputs.frameRow, 1));
utility("frame-col", set(inputs.frameRow, 0));
```

So `--_ak-frame-col: 1` actually means "this is a row" — the inverse of what the CSS variable name suggests. This looks like a refactor leftover (probably renamed `frame-col`→`frame-row` in TS but the registered name was missed).

It compiles correctly today, but anyone reading the generated `output.css` or grepping `--_ak-frame-col` will need to chase the indirection.

Suggest aligning the registered prop name to the field:

```ts
frameRow: _ak.prop.number("frame-row", { initial: 0, inherits: true }),
```

…or rename the field and context vars to `frameCol`/`frameParentColContext` and flip the boolean polarity (whichever direction matches the design intent).

packages/ariakit-tailwind/migrating-to-v02.md
```diff
@@ -0,0 +1,454 @@
...
+}
+@utility my-check_checked {
+  /* Must reset darken-0 since _idle's darken-6 is still on the element */
+  @apply ak-layer ak-layer-(color:--my-color) ak-layer-darken-0;
+}
+```
+
+### `ak-frame-*/[calc(...)]` arbitrary modifier generates invalid CSS
+
+The `ak-frame-*` utility generates a `--spacing()` multiplication for arbitrary bracket modifiers, which can produce invalid `<length> × <length>` values. Workaround: use the frame preset without modifier and set the padding with `ak-frame-p-*`.
```
diegohaz: Two of the four "Known issues" listed in this section look like they should block the v0.2 release rather than ship as documented bugs:

**1. `ak-frame-*/[calc(...)]` arbitrary modifier generates invalid CSS** (this section, line 448). The workaround — drop the modifier, switch to `ak-frame-p-*` — is a non-trivial rewrite at every call site. Consumers that already use the modifier-with-calc pattern will silently get invalid `<length> × <length>` declarations in their production CSS. If a real fix is not feasible for v0.2, please at least:

- File a follow-up issue and link it from this section so it is discoverable.
- Consider failing fast at build time (e.g. emit a Tailwind error or a sentinel `@error` rule) rather than emitting silently invalid CSS.

**2. `ak-text` sets `background-color: transparent !important`** (line 411). Forcing `!important` on a property that the consumer almost certainly wants to control (whenever they combine `ak-text` with `ak-layer`) leaks plugin internals into consumer CSS. The recommended workaround — re-apply the background `!important` in every consumer utility that combines `@apply ak-layer ak-text` — is fragile and easy to forget. Worth exploring whether this can be scoped to the cases that actually need it (e.g. only when no `ak-layer` is present on the same element, via a sibling utility, a `:not(...)` selector, or a `--ak-text-bg` custom property that the layer can override).

The other two known issues (`ak-layer-(color:--var)` not resetting same-element darken/lighten, and responsive `ak-frame-cover` overriding non-responsive `rounded-*!`) are reasonable to ship with documentation since the workarounds are local and discoverable.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -0,0 +1,2202 @@
...
+    4,
+  );
+}
+
+/**
+ * Returns whether `value` is within `[min, max]`
+ */
+function getRangeMask(value: Value, min: number, max: number) {
+  return fn.mul(
+    fn.binary(fn.sub(value, min - 1e-6)),
```
diegohaz: `min - 1e-6` is computed in JS at code-gen time and the literal result gets inlined into `output.css`. For most thresholds it stringifies cleanly (`0.55 → 0.549999`, `0.725 → 0.724999`, `0 → -0.000001`), but for `min = 0.275` the IEEE-754 result is `0.27499900000000005`, which lands verbatim in the generated CSS:

```css
/* output.css line 1092 */
initial-value: calc(clamp(0, ((l - 0.27499900000000005) * 1000000), 1) * clamp(0, ((0.55 - l) * 1000000), 1));
```

It is the only artifact today — but adding any new boundary in this band would expose more, and the long literal is needlessly noisy for downstream tooling that diffs the generated CSS.

Two cheap fixes:

```ts
// option A — round the JS-computed value (roundToDecimals is already used above)
return fn.mul(
  fn.binary(fn.sub(value, roundToDecimals(min - 1e-6, 6))),
  fn.binary(fn.sub(max, value)),
);
```

```ts
// option B — let the browser do the subtraction
const EPS = 1e-6;
return fn.mul(
  fn.binary(fn.sub(value, fn.sub(min, EPS))),
  fn.binary(fn.sub(max, value)),
);
```

Either keeps the generated CSS deterministic and a few bytes smaller for any future thresholds.

diegohaz (commented):
## Review summary

Initial review of `tailwind/rewrite` against `main` (merge-base diff). Review covered the full `@ariakit/tailwind` package rewrite (`src/input.ts`, `src/lib.ts`, `src/build.ts`, `src/output.css`), the new `migrating-to-v02.md` and `readme.md`, the package manifest changes, the `templates/react/` migration, the `site/src/sandbox/ariakit-tailwind/` sandbox + `test-chrome.ts` / `perf-chrome.ts` harness, and impacted call sites under `site/`.

- `pnpm -F @ariakit/tailwind build` (re-runs `node src/build.ts`) — `output.css` regenerates with **zero diff** from the committed file, so the generated CSS is in sync with `input.ts`.
- `npx tsc --noEmit -p packages/ariakit-tailwind` — exit 0.
- Verified via `git diff --name-status` that `migrating-to-v02.md` (A) and `src/input.ts` (A) are new files so all referenced line numbers are inside the diff.
- Spot-checked `templates/react/src/{app.tsx,style.css}` and the remaining `ak-layer-canvas` / `ak-layer-pop` / `ak-layer-hover` / `ak-bordering` / `ak-frame-cover-*` / `ak-text/N` usages across `site/`, `templates/`, and `nextjs/`. The legacy hits I could find are either inside `/* … */` comment blocks (`site/src/styles/global.css:71`, `site/src/components/canvas-color-picker.astro:26`) or already migrated to the v0.2 syntax (e.g. `"ak-layer"`+`"ak-layer-canvas"` on the same element). Migration coverage looks complete.

### Skipped (worktree limitation)

The linked worktree at `/Users/diegohaz/Developer/ariakit/.claude/worktrees/peppy-sparking-map` has no `node_modules` and a full `pnpm install` would be heavy, so I did **not** run:

- `pnpm -F site test-chrome ariakit-tailwind` (WCAG axe-core + computed-styles snapshots)
- `pnpm -F site test-perf ariakit-tailwind`
- `pnpm tsc` / `pnpm lint-fix` from the repo root
- `pnpm build-site-lite`

CI on the PR head should cover those — please confirm the snapshot diffs and perf deltas there before merging.

### Findings

**Blocking-ish (please address before merge or land an explicit decision):**

1. **Missing changeset and unbumped version.** `package.json` still reads `0.1.11` and `.changeset/` has no PR-specific entry. `@ariakit/tailwind` is not in the changeset `ignore` list, so without a changeset the rewrite will not produce a version bump or `CHANGELOG.md` entry on release. Inline at `migrating-to-v02.md:1`.
2. **Empty PR body.** This is a 27,567-addition / 3,463-deletion rewrite that introduces a new public API surface and a dedicated migration guide, but the PR description is empty. For something this size, a short summary of motivation, scoped breaking changes, and a pointer to `migrating-to-v02.md` would help reviewers and future archeology — please populate the body before merge.
3. **Two of the four shipping "Known issues" warrant a second look.** Specifically: (a) `ak-frame-*/[calc(...)]` silently emits invalid CSS at consumer call sites; (b) `ak-text` forces `background-color: transparent !important`, leaking plugin internals into consumer styles. Inline at `migrating-to-v02.md:450` with proposed mitigations / follow-up tracking.

**Non-blocking:**

4. **`frameRow` field vs `--_ak-frame-col` CSS prop name mismatch** at `packages/ariakit-tailwind/src/input.ts:687` — refactor leftover; functionally fine but confusing for anyone reading the generated CSS or the input declaration.
5. **Floating-point precision artifact (`0.27499900000000005`) in `output.css:1092`** caused by `min - 1e-6` being evaluated in JS at code-gen time inside `getRangeMask` (`packages/ariakit-tailwind/src/input.ts:241`). Tiny CSS-size and determinism win available via `roundToDecimals` or moving the subtraction CSS-side.

### Areas not fully validated

- WCAG AA contrast across the four light/dark × normal/high-contrast snapshot variants (relies on the test suite I could not run locally).
- Perf regression direction relative to `main` (`page-load` and toggle benchmarks were not run here).
- Cross-browser behavior of `@property` registered custom properties with the chosen `inherits` flags (Chromium-only render is what `test-chrome.ts` covers).

No approval or change-request — comment-only review per the review skill. Happy to re-review the next push.

packages/ariakit-tailwind/package.json
diegohaz: This is not an issue because this PR is intentionally not adding a changeset per maintainer direction.

packages/ariakit-tailwind/migrating-to-v02.md
diegohaz: This is not an issue because this PR is intentionally not adding a changeset per maintainer direction.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 8754b8ddc - renamed the registered frame row custom property to `--_ak-frame-row` and regenerated `output.css`.

packages/ariakit-tailwind/migrating-to-v02.md
diegohaz: Fixed in 8754b8ddc - removed the `ak-text` `!important`, fixed arbitrary frame modifiers for raw length/calc values, and dropped both items from the known-issues list.

site/src/pages/ak.astro
diegohaz: Fixed in a72a3e620 - removed the scratch `/ak` Astro route and its page-local stylesheet.

site/src/pages/plus/checkout/[...checkout].astro
Comment (diegohaz) [reply site/src/pages/plus/checkout/[...checkout].astro]: Fixed in 491f6100f - migrated the remaining stale `ak-edge-contrast-primary` usages to `ak-edge-primary ak-edge-raw`.

package.json
diegohaz: Fixed in a0e768ffd - moved `@axe-core/playwright` from the root devDependencies to `site/package.json` and updated the lockfile.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 8754b8ddc - rounded the generated lower range boundary before emitting `output.css`.

diegohaz:
**Fixed**

- a0e768ffd ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983789)): Moved `@axe-core/playwright` from the root dev dependencies to the `site` workspace and updated the lockfile.
- 491f6100f ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983786)): Migrated the remaining stale `ak-edge-contrast-primary` usages to `ak-edge-primary ak-edge-raw`.
- a72a3e620 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983783)): Removed the scratch `/ak` route and its page-local stylesheet.
- 8754b8ddc ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983779)): Renamed the registered frame row custom property to `--_ak-frame-row` and regenerated `output.css`.
- 8754b8ddc ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983781)): Removed the `ak-text` `!important`, fixed arbitrary frame length/calc modifiers, and removed both items from the known-issues list.
- 8754b8ddc ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983792)): Rounded the generated lower range boundary before emitting `output.css`.

**Invalid**

- no changeset ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983775)): Changeset intentionally omitted per maintainer direction.
- no changeset ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3147983778)): Changeset intentionally omitted per maintainer direction.

site/src/components/code-block.react.tsx
```diff
@@ -873,7 +877,7 @@ export function CodeBlockTabs({
               {preview && (
                 <div
                   className={clsx(
-                    "absolute @max-[64rem]:hidden z-1 text-lg ak-text/0 top-1/2 inset-e-0 -translate-y-1/2 translate-x-7 ak-layer-current ak-light:ak-edge/15 ak-dark:ak-edge/13 size-10 grid place-items-center border touch-none rounded-full",
+                    "absolute @max-[64rem]:hidden z-1 text-lg ak-ink-0 top-1/2 inset-e-0 -translate-y-1/2 translate-x-7 ak-layer ak-light:ak-edge-15 ak-dark:ak-edge-[13] size-10 grid place-items-center border touch-none rounded-full",
```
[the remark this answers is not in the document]
diegohaz: Fixed in 2510d944a - normalized the `ak-dark:ak-edge-[13]` class to `ak-dark:ak-edge-13`.

site/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -0,0 +1,589 @@
...
+      const textContent = node.textContent?.trim();
+      if (!textContent) {
+        continue;
+      }
+      label = textContent;
+      break;
+    }
+    return label;
+  };
+
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7f9a6d51e - consolidated the section text-label fallback within each browser-evaluated context.

site/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -0,0 +1,589 @@
...
+      const contrastData = getContrastRatio(node);
+      if (contrastData == null) {
+        continue;
+      }
+      const selector = getTargetSelector(node.target);
+      if (!selector) {
+        continue;
+      }
+      const element = document.querySelector(selector);
+      const key = element ? getSectionPath(element) : selector;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7f9a6d51e - axe target selector formatting now catches invalid selector syntax and falls back to the raw selector string.

site/src/styles/styles.json
```diff
@@ -6297,32 +6664,37 @@
...
+              "type": "utility",
+              "name": "ak-frame-full/0",
               "import": "@ariakit/tailwind"
             },
             {
               "type": "utility",
-              "name": "ak-layer-current",
+              "name": "ak-progress-thickness-2",
               "import": "@ariakit/tailwind"
             },
```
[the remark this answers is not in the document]
diegohaz: Fixed in a7f25c580 - style dependency resolution now matches local wildcard utilities before falling back to `@ariakit/tailwind`, with regression coverage.

site/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -0,0 +1,589 @@
...
+          ? child.getAttribute("aria-labelledby")
+          : null;
+      if (labelledBy) {
+        const label = getSectionLabel(child, labelledBy);
+        const className = extractClass(child);
+        const children = walk(child);
+        result[label] =
+          Object.keys(children).length > 0
+            ? { class: className, children }
+            : className;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7f9a6d51e - duplicate sibling section labels now get stable suffixed keys instead of overwriting earlier snapshot entries.

packages/ariakit-tailwind/readme.md
```diff
@@ -1,297 +1,531 @@
...
+| `ak-layer-h-<value>`         | Sets absolute hue. Accepts a named hue (`ak-layer-h-blue`), a degree (`ak-layer-h-240`), or arbitrary value (`ak-layer-h-[240deg]`). |
+| `ak-layer-h-rotate-<number>` | Rotates hue by `<number>` degrees. Accepts named relational hues (`ak-layer-h-rotate-complementary`).                                |
+
+### Chroma (saturation) adjustments
+
+| Utility                        | Description                                                                                                                                             |
+| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ak-layer-saturate-<number>`   | Increases chroma by `<number>`%. Accepts named presets (`ak-layer-saturate-vivid`).                                                                     |
+| `ak-layer-desaturate-<number>` | Decreases chroma by `<number>`%.                                                                                                                        |
```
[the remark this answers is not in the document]
diegohaz: Fixed in bb8f630c4 - added named chroma preset support to layer/state/edge/text/outline saturate utilities and regenerated `output.css`.

site/src/styles/styles.json
```diff
@@ -2342,7 +2492,17 @@
...
               "type": "utility",
-              "name": "ak-layer-(--ak-tab-bg)",
+              "name": "ak-layer",
+              "import": "@ariakit/tailwind"
+            },
+            {
+              "type": "utility",
+              "name": "ak-layer-(color",
+              "import": "@ariakit/tailwind"
+            },
```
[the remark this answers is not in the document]
diegohaz: Fixed in a7f25c580 - made style dependency token parsing split variants only on top-level colons, preserving arbitrary values like `ak-layer-(color:...)`, with regression coverage.

package.json
```diff
@@ -84,9 +84,11 @@
     "autoprefixer": "10.5.0",
     "chalk": "5.6.2",
     "chokidar": "5.0.0",
+    "clava": "0.2.0",
     "concurrently": "9.2.1",
     "cross-env": "10.1.0",
     "cross-spawn": "7.0.6",
+    "csstype": "3.2.3",
```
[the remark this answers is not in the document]
diegohaz: Fixed in a38b02107 - removed the unused root `clava` dependency and duplicate root `csstype` dependency.

diegohaz:
**Fixed**

- 7f9a6d51e ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552222)): Consolidated section text-label fallback helpers within each browser-evaluated context.
- 7f9a6d51e ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552226)): Made duplicate sibling section labels produce stable suffixed snapshot keys instead of overwriting entries.
- 7f9a6d51e ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552224)): Wrapped axe target selector lookup so invalid selector syntax falls back to the raw selector string.
- a38b02107 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552236)): Removed unused root `clava` and duplicate root `csstype` dependencies.
- bb8f630c4 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552227)): Added named chroma preset support to layer/state/edge/text/outline saturate utilities and regenerated `output.css`.
- 2510d944a ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552218)): Normalized the `ak-dark:ak-edge-[13]` class to `ak-dark:ak-edge-13`.
- a7f25c580 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552228)): Preserved arbitrary-value colons when parsing style dependency tokens and added regression coverage.
- a7f25c580 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149552225)): Resolved local wildcard utilities before external fallbacks in generated style dependencies and added regression coverage.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -0,0 +1,2207 @@
...
+
+/**
+ * Resolves layer lightness from relative offset and optional absolute input.
+ */
+function getLayerL(relativeLightness: Value, absoluteLightness?: VarProperty) {
+  // Preserve the parent lightness as a floor so nested layers do not collapse
+  // into black when multiple negative offsets stack.
+  const baseLightnessFloor = fn.max(l, fn.min(0.13, relativeLightness));
+  const fallbackLightness = fn.add(baseLightnessFloor, relativeLightness);
+  return absoluteLightness
```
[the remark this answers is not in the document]
diegohaz: Fixed in f1eb8121a - the finding was valid. `getLayerL` now applies the 0.13 base floor only while the relative lightness is negative, so small repeated darkening offsets converge instead of stacking to black. I used the explicit darkening-mask form instead of the simpler inner-`max()` prompt because the latter would also over-adjust positive/lightening offsets.

site/src/lib/build-styles.ts
```diff
@@ -503,26 +503,112 @@ async function discoverModulePaths(): Promise<string[]> {
...
+  for (const entry of wildcardIndex) {
+    if (!entry.re.test(token)) {
+      continue;
+    }
+    if (!best || entry.name.length > best.name.length) {
+      best = { name: entry.name, module: entry.module };
+    }
+  }
+  return best?.module ?? null;
 }
```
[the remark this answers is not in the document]
diegohaz: Fixed in c94f6d166 - wildcard dependency matches now record the canonical wildcard key, and tokens that exist as both utility and variant dependencies are recorded for both types instead of returning after the utility match.

diegohaz:
**Fixed**

- f1eb8121a + b60cf4e98 ([latest reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3150008134)): Removed the implicit `0.13` lightness floor from `getLayerL`; relative lightness now stays `l + relativeLightness`, and explicit floors should use `ak-layer-min-*` such as `ak-layer-min-13`. Also reverted named relative-chroma support and kept the docs numeric-only for `ak-layer-saturate-*`.
- c94f6d166 ([reply](https://github.com/ariakit/ariakit/pull/5473#discussion_r3149982809)): Canonicalized wildcard dependency names and kept dual utility/variant dependencies instead of returning after utility matches.

packages/ariakit-tailwind/src/input.ts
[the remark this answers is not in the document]
diegohaz: Updated in b60cf4e98 - removed the implicit 0.13 minimum from `getLayerL` entirely. Relative lightness now resolves to `l + relativeLightness`; callers that need a floor can opt in with `ak-layer-min-*`, for example `ak-layer-min-13`.

PR #5598, opened 2026-03-18
Fix nested form collection items registering to wrong store

### LINKED ISSUE #5273: Nested `Form*` components incorrectly register items to stores other than the closest `Form` store
### Current behavior

When nesting collection stores (i.e. a `TabPanel` that then renders a `FormRadioGroup`/`FormRadio`), i am seeing that these `Form*` components are registering themselves to the `TabStore`'s items property as well as the `FormStore`'s items property. I assume this is unexpected behavior, as the only items that should be registering themselves to the `TabStore` should only be `Tab` items, not `FormRadio`/`FormCheckbox`/etc items.

1. Open https://stackblitz.com/edit/vitejs-vite-l4xehvzw?file=src%2FApp.tsx
2. Observe the output of the tab store items in the preview window. I've added further comments in the above reproduction to explain what's going on in the provided sub-components.

The `items` property of the Tab store state indicates that the `FormRadio` items are being registered to the tab store, instead of just the expected `Tab` items.

### Expected behavior

I would expect that a `TabStore` should never have items registered to it other than individual `Tab` items. Nested form items should register to their closest parent form store, rather than registering themselves to all enclosing collection stores.

### Workaround

It seems that explicitly rendering a `RadioProvider` around these `FormRadio` components resolves the issue and avoids these `FormRadio` items from being registered incorrectly to the ancestor `TabStore`. However, the `FormRadio` [docs](https://ariakit.org/reference/form-radio) don't seem to indicate that this is required/desired.

### Possible solutions

_No response_

diegohaz:
## Problem

When nesting collection stores (e.g., `FormRadio` inside a `TabPanel`), form items incorrectly register themselves to ancestor composite stores (like `TabStore`) in addition to their own `FormStore`. This happens because `TabPanel` wraps its children in a scoped `CompositeScopedContextProvider` that cascades to set the composite and collection contexts to the `TabStore`. When `FormRadio` calls `useRadio` → `useCompositeItem`, it picks up the `TabStore` from the composite context instead of finding a `RadioStore`.

This causes:
- `FormRadio` items appearing in the `TabStore`'s items list
- Arrow keys in the tab list navigating to radio items instead of other tabs
- Broken composite behavior due to mixed item types in the store

Fixes #5273

## Solution

`FormRadioGroup` now resets the composite context for its children by wrapping them with `CompositeScopedContextProvider value={undefined}`. This prevents `FormRadio` items from picking up an ancestor composite store. Form controls are unaffected because they explicitly pass their `FormStore` to `useCollectionItem`, bypassing the context.

## Workaround

Wrap `FormRadioGroup` with `RadioProvider` and render it as `RadioGroup`. This creates an isolated composite context for the radio items, preventing them from registering to the ancestor `TabStore`.

```jsx

// TODO: Remove RadioProvider workaround once
// https://github.com/ariakit/ariakit/issues/5273 is fixed
<Ariakit.RadioProvider>
  <Ariakit.FormRadioGroup render={<Ariakit.RadioGroup />}>
    <Ariakit.FormGroupLabel>Favorite color</Ariakit.FormGroupLabel>
    <Ariakit.FormRadio name={form.names.color} value="red" />
    <Ariakit.FormRadio name={form.names.color} value="green" />
    <Ariakit.FormRadio name={form.names.color} value="blue" />
  </Ariakit.FormRadioGroup>
</Ariakit.RadioProvider>
```


diegohaz:
Addressed the Copilot review comments:

**`store` destructuring in `useFormRadioGroup`** — This is an established pattern across all form hooks (e.g., `useFormGroup`, `useFormReset`, `useFormGroupLabel`). The `store` prop is destructured to strip it from `...props` so it doesn't get forwarded to the DOM element. No change needed.

**`--grep` not matching sandbox tests** — Good catch. Updated the bug report skill to use file path filtering instead: `pnpm -F site run test-chrome -- site/src/sandbox/<issue-number>/`. The test itself doesn't need to include the issue number in the title.

Issue #4825, opened 2025-07-15
Combobox input glitching when search results change


jovanblazek:
### Current behavior

When search results change, input is "scrolled" to the beginning. When input has more text than can be displayed at once, it oscillates between start and cursor positions as you type.

### Steps to reproduce the bug

Reproduction:
https://stackblitz.com/edit/yrabhlyx-ch26guvp?file=dialog-combobox-command-menu%2Findex.tsx

To reproduce, the text in the input must be longer than the input element, I suggest changing viewport size for easier reproduction.
1. type something that you will remember, like `start`. 
2. Then type anything until the text is longer than the input, just make sure you get some search results. 
3. Once text is longer than the input, each time you type/remove a letter and search results change, the input jumps to the beginning - you can see the first word you typed in (`start` in my case)

What is interesting is that the input still has active cursor so I can type and remove text, but it behaves like it does not have a focus. 

https://github.com/user-attachments/assets/703a7adb-cbd4-46c2-b660-3a66647b6196

### Expected behavior

- Input shows text around position of the cursor while user is typing

jovanblazek:
Thanks a lot for the workaround! 🚀

PR #5610, opened 2026-03-20
Fix combobox input scroll position reset on results change

diegohaz:
Fixes https://github.com/ariakit/ariakit/issues/4825

## Problem

When a combobox input's text overflows its width and search results change, the input's `scrollLeft` is reset to 0, causing the text to jump to the beginning. This happens because the virtual focus mechanism in `composite.tsx` briefly moves DOM focus from the input to a combobox item and back (the "virtual focus dance"), which causes browsers to reset the input's internal scroll position.

The two places where this focus dance occurs are:
1. The `focusIntoView` effect that fires when `moves` changes
2. The `useScheduleFocus` function that focuses the active item when the composite container receives focus

## Solution

Save the base element's `scrollLeft` and `scrollTop` before the virtual focus dance and restore them immediately after, in both `focusIntoView` and `useScheduleFocus`.

## Workaround

Until this fix is released, you can work around the issue by restoring the scroll position on focus capture:

```jsx
<Combobox
  // TODO: Remove this when https://github.com/ariakit/ariakit/issues/4825 is fixed
  onFocusCapture={(event) => {
    event.target.scrollLeft = event.target.clientWidth;
  }}
/>
```

### DIFF
--- a/.changeset/fix-combobox-scroll-position.md
+++ b/.changeset/fix-combobox-scroll-position.md
@@ -3,6 +3,8 @@
 "@ariakit/react": patch
 ---
 
-Fixed `Combobox` input scroll position resetting when search results change
+`Combobox` input scroll position resetting
 
-When a [`Combobox`](https://ariakit.org/reference/combobox) input's text overflows its width and items are filtered asynchronously (e.g., via `startTransition`), the input's horizontal scroll position was being reset to the beginning each time results changed. This was caused by the virtual focus mechanism briefly moving DOM focus to the active item and back, which made browsers reset the input's internal `scrollLeft`. The scroll position is now preserved across these focus transitions.
+When a [`Combobox`](https://ariakit.org/reference/combobox) input's text overflowed its width, the input's horizontal scroll position reset to the beginning each time the results changed. This happened because the virtual focus mechanism briefly moved DOM focus to the active item and back when [`autoSelect`](https://ariakit.org/reference/combobox#autoselect) was enabled, causing browsers to reset the input's internal `scrollLeft`.
+
+The scroll position is now preserved across these focus transitions.


.changeset/fix-combobox-scroll-position.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Combobox`](https://ariakit.org/reference/combobox) input scroll position resetting when search results change
```
diegohaz: Changeset title can't have links
```suggestion
Fixed `Combobox` input scroll position resetting when search results change
```

packages/ariakit-react-core/src/composite/composite.tsx
```diff
@@ -117,8 +117,19 @@ function useScheduleFocus(store: CompositeStore) {
...
+    const savedScrollLeft =
+      virtualFocus && baseElement ? baseElement.scrollLeft : 0;
+    const savedScrollTop =
+      virtualFocus && baseElement ? baseElement.scrollTop : 0;
     activeElement.focus({ preventScroll: true });
-  }, [activeItem, scheduled]);
+    if (virtualFocus && baseElement) {
+      baseElement.scrollLeft = savedScrollLeft;
+      baseElement.scrollTop = savedScrollTop;
+    }
```
diegohaz: This logic should probably be abstracted into a top-level function like `fn(store, () => activeElement.focus({ preventScroll: true }))`

diegohaz: Agreeing with @diegohaz's suggestion to extract this. Since both call sites follow the exact same save/restore pattern, a helper would remove the duplication and make the intent clearer. Something like:

```ts
function withBaseScrollPreserved(
  store: CompositeStore,
  callback: () => void,
) {
  const { virtualFocus, baseElement } = store.getState();
  const savedScrollLeft =
    virtualFocus && baseElement ? baseElement.scrollLeft : 0;
  const savedScrollTop =
    virtualFocus && baseElement ? baseElement.scrollTop : 0;
  callback();
  if (virtualFocus && baseElement) {
    baseElement.scrollLeft = savedScrollLeft;
    baseElement.scrollTop = savedScrollTop;
  }
}
```

Then both call sites become one-liners:
```ts
// useScheduleFocus
withBaseScrollPreserved(store, () => activeElement.focus({ preventScroll: true }));

// focusIntoView effect
withBaseScrollPreserved(store, () => focusIntoView(itemElement));
```

See next comment for a nuance about scoping this to text fields in the `focusIntoView` case.

packages/ariakit-react-core/src/composite/composite.tsx
```diff
@@ -167,10 +178,22 @@ export const useComposite = createHook<TagName, CompositeOptions>(
...
+      // `scrollLeft` on a text input). We save and restore it here.
+      const savedScrollLeft =
+        virtualFocus && baseElement ? baseElement.scrollLeft : 0;
+      const savedScrollTop =
+        virtualFocus && baseElement ? baseElement.scrollTop : 0;
       focusIntoView(itemElement);
+      if (virtualFocus && baseElement) {
+        baseElement.scrollLeft = savedScrollLeft;
+        baseElement.scrollTop = savedScrollTop;
+      }
```
diegohaz: Copilot flagged scoping the restore to `isTextField`, and I think the concern is valid specifically for this `focusIntoView` call site — but not for `useScheduleFocus`.

Here's why: `focusIntoView` calls `scrollIntoView({ block: "nearest", inline: "nearest" })` on the item element. If the base element happens to be a scrollable ancestor of the items (e.g., a `div` composite container with `virtualFocus` and `overflow: auto`), that `scrollIntoView` would intentionally scroll the base element to bring the active item into view. Restoring `scrollLeft`/`scrollTop` afterward would undo that scroll, making the item invisible.

In `useScheduleFocus`, this isn't a problem because it only calls `activeElement.focus({ preventScroll: true })` — no intentional scrolling occurs, so restoring is always safe.

For this call site, guarding with `isTextField(baseElement)` (already imported in this file) would make the fix safe for all composite configurations while still solving the combobox input scroll issue.

site/src/sandbox/4825-dialog-combobox/test-browser.ts
```diff
@@ -0,0 +1,42 @@
...
+  }) => {
+    const combobox = q.combobox();
+    await test.expect(combobox).toBeFocused();
+
+    // Type padding spaces followed by "b" to overflow and match multiple items.
+    // "b" matches Banana and Blueberry. Adding "a" will change results to just
+    // Banana, triggering the async autoSelect effect which causes the virtual
+    // focus dance that resets scrollLeft.
+    await page.keyboard.type("                  b", { delay: 20 });
+    await page.waitForTimeout(300);
```
diegohaz: The `waitForTimeout(300)` calls (here and line 31) can be removed. The `toBeVisible` / `toBeHidden` assertions that follow already auto-retry with Playwright's built-in timeout, so they handle the `startTransition` delay. The `scrollLeft` evaluations come after those assertions, so by that point the transition has settled.

If you want extra safety on the `scrollLeft` check itself, `expect.poll` would be more deterministic than a fixed timeout:

```ts
await test.expect.poll(
  () => combobox.evaluate((el) => (el as HTMLInputElement).scrollLeft),
).toBeGreaterThan(0);
```

packages/ariakit-react-core/src/composite/composite.tsx
diegohaz: Fixed in 4f47d8740 - extracted into a `withBaseScrollPreserved(store, callback)` helper. Both call sites are now one-liners.

site/src/sandbox/4825-dialog-combobox/test-browser.ts
diegohaz: Fixed in 4f47d8740 - removed timeouts and used `expect.poll` for scrollLeft checks.

packages/ariakit-react-core/src/composite/composite.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 4f47d8740 - the `withBaseScrollPreserved` helper now applies the `isTextField` guard uniformly in both call sites.

.changeset/fix-combobox-scroll-position.md
diegohaz: Fixed in 4f47d8740 - removed links from changeset title.

packages/ariakit-react-core/src/composite/composite.tsx
```diff
@@ -167,10 +178,22 @@ export const useComposite = createHook<TagName, CompositeOptions>(
...
+      // When using virtual focus, `focusIntoView` temporarily moves DOM focus
+      // to the item and back to the base element. This focus dance causes
+      // browsers to reset the base element's internal scroll position (e.g.,
+      // `scrollLeft` on a text input). We save and restore it here.
+      const savedScrollLeft =
+        virtualFocus && baseElement ? baseElement.scrollLeft : 0;
+      const savedScrollTop =
+        virtualFocus && baseElement ? baseElement.scrollTop : 0;
       focusIntoView(itemElement);
+      if (virtualFocus && baseElement) {
```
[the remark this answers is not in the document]
diegohaz: Fixed in 4f47d8740 - extracted a `withBaseScrollPreserved` helper that guards with `isTextField(baseElement)`, so only text field base elements get scroll preservation. This fixes the CI failures in select-combobox-offscreen and select-group tests where restoring scroll was undoing `scrollIntoView` on scrollable containers.

packages/ariakit-react-core/src/composite/composite.tsx
diegohaz: Fixed in 4f47d8740 - applied the `isTextField` guard in the helper, which covers both call sites.

site/src/sandbox/4825-dialog-combobox/test-browser.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 4f47d8740 - removed both `waitForTimeout` calls and replaced `scrollLeft` checks with `expect.poll`.

packages/ariakit-react-core/src/composite/composite.tsx
diegohaz: Fixed in 4f47d8740.

Issue #4593, opened 2025-04-24
`SelectItem` generates infinite loop issue when `id` prop is passed as `undefined`


nvignola:
### Current behavior

The `SelectItem` component accepts an optional `id` [prop](https://ariakit.org/reference/select-item#id).

When the `id` prop is explicitly passed as `undefined`, Ariakit tries to generate an ID internally, but it has a bug in how the `props` are merged.

Generation of the [ID internally](https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-core/src/select/select-item.tsx#L74):

```react
const id = useId(props.id);
```

When [merging](https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-core/src/select/select-item.tsx#L150) `props`, the generated `id` get overridden by `props.id`.

```react
props = {
  id,
  role: getPopupItemRole(listElement),
  "aria-selected": selected,
  children: value,
  ...props, // The undefined props.id overrides the freshly created id
  autoFocus: props.autoFocus ?? autoFocus,
  onClick,
};
```

This creates a mismatch between the generated ID and what's used for focus management.

When clicking an item, this mismatch triggers an infinite event loop between blur events and focus handling causing the error: _"Maximum call stack size exceeded"_.

![Error message](https://github.com/user-attachments/assets/e11b33e1-1f62-4937-9eb6-3aec247766b3)

### Steps to reproduce the bug

1. Open https://stackblitz.com/edit/ariakit-ahdu3ywv?devToolsHeight=33&file=src/App.tsx
2. Click to open the select popover
3. See console logs

### Expected behavior

- When explicitly passing the prop `id` as `undefined`, the value generated by the `useId` hook, should override the one from `props.id`

### Workaround

- Perform a check on the `id` value before passing it to the `SelectItem` component.
- If `undefined` rather don't pass it or use a fallback value.

### Possible solutions

A low-impact solution - in terms of changes - would be to move the `id` field after the `props` destructuring. 

```react
props = {
  role: getPopupItemRole(listElement),
  "aria-selected": selected,
  children: value,
  ...props,
  id, // ensure props.id is overridden
  autoFocus: props.autoFocus ?? autoFocus,
  onClick,
};
```

Another possibility, is to remove the `id` from the passed `props`.
This way, the only source of truth for the `id` value is the `useId` hook.

Example:

```
// Extract id from props first
const { id: providedId, ...restProps } = props;

// Generate stable ID based on provided ID or auto-generate
[...]

PR #5611, opened 2026-03-20
Fix SelectItem infinite loop when id prop is undefined

diegohaz:
Closes https://github.com/ariakit/ariakit/issues/4593

## Problem

When a component like `SelectItem` receives `id={undefined}`, the internally generated `id` from `useId(props.id)` gets overridden by `...props` spread (which contains `id: undefined`). This creates a mismatch between the generated ID and what's used for focus management, triggering an infinite event loop between blur events and focus handling that causes "Maximum call stack size exceeded".

This pattern existed in 24 components across the codebase.

## Solution

Move the generated `id` after `...props` in the props spread object so that `id={undefined}` cannot override the internally generated id. Since `useId(props.id)` already incorporates a user-provided id when it's truthy, this change is safe — the generated id always equals the user-provided id when one is given.

## Workaround

When passing a potentially `undefined` id to `SelectItem` (or other affected components), filter it out so Ariakit can generate the id internally:

```tsx
const id: string | undefined = undefined;

<SelectItem
  // TODO: Remove workaround once fixed
  // https://github.com/ariakit/ariakit/issues/4593
  {...(id != null ? { id } : null)}
  value="Apple"
/>
```

### DIFF
--- a/packages/ariakit-react-core/src/tag/tag.tsx
+++ b/packages/ariakit-react-core/src/tag/tag.tsx
@@ -119,11 +119,11 @@ export const useTag = createHook<TagName, TagOptions>(function useTag({
   );
 
   props = {
-    id,
     role: !touchDevice ? "option" : "listitem",
     children: value,
     "aria-describedby": removeId,
     ...props,
+    id,
     onKeyDown,
   };
 
--- a/.changeset/fix-undefined-id-prop.md
+++ b/.changeset/fix-undefined-id-prop.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed components crashing with "Maximum call stack size exceeded" when the [`id`](https://ariakit.org/reference/select-item#id) prop is explicitly passed as `undefined`.
--- a/packages/ariakit-react-core/src/combobox/combobox-list.tsx
+++ b/packages/ariakit-react-core/src/combobox/combobox-list.tsx
@@ -121,9 +121,9 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
         : null;
 
     props = {
-      id,
       hidden,
       ...props,
+      id,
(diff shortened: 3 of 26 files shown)


packages/ariakit-react-core/src/tag/tag.tsx
```diff
@@ -119,11 +119,11 @@ export const useTag = createHook<TagName, TagOptions>(function useTag({
   );
 
   props = {
-    id,
     role: !touchDevice ? "option" : "listitem",
     children: value,
     "aria-describedby": removeId,
     ...props,
+    id,
```
diegohaz, line 126: This sweep fixes `Tag`, but it still leaves the same `id={undefined}` bug in [`tag-remove.tsx`](https://github.com/ariakit/ariakit/blob/e3f96bded9fff439b710b59d249ec897404cf887/packages/ariakit-react-core/src/tag/tag-remove.tsx#L91-L103): `useTagRemove()` still builds `{ id, ...props, ... }`, so an explicit `id={undefined}` clears the DOM id after `useId()` generated one. That matters here because `useEffect()` still publishes the generated id through `TagRemoveIdContext`, so `Tag` ends up setting `aria-describedby` to an element id that no longer exists. I think `TagRemove` needs the same `...props`/`id` ordering change as the other components in this PR.

diegohaz (commented):
Reviewed the current PR head `e3f96bded9fff439b710b59d249ec897404cf887` against base `d45dabbc8bcc626e4ad19e57b557f38b975af710` as an initial review. I checked the diff plus the surrounding `useId(props.id)` call sites to make sure the sweep was complete, and I ran `pnpm install` followed by `pnpm -F site test --grep 'SelectItem with id={undefined} does not cause infinite loop on click'` (pass in Chrome, Safari, and Firefox).

Main finding category: incomplete coverage of the `id={undefined}` fix.

I did not run a full repo typecheck or broader browser suite.

PR #5613, opened 2026-03-20
Fix combobox popover scroll reset during infinite scroll

### LINKED ISSUE #5165: Combobox Popover scrolls to top when contents change, only when selectedValues is Array (infinite scroll usecase)
### Current behavior

When simulating an infinite scroll use case, the scroll in the popover jumps back up to the top (or whatever item was hovered over recently before the scroll started). This happens with or without items being selected. 

This only seems to happen when selectedValues is an array (the combobox allows for multiple items to be selected without closing).
video: https://github.com/user-attachments/assets/3651e8de-0f58-44df-98f8-93686370ec9d

1. open https://stackblitz.com/edit/sb1-vdukvn59?file=src%2FApp.tsx
2. click the input to open the popover
3. scroll to the bottom of the popover. 
4. when you reach the bottom, notice that the scroll position will jump to the top. This happens when another page of data is added (via the handleScroll callback setting visibleCount in order to include another page worth of data in the popover)

### Expected behavior

When new items are added inside the popover, the scroll position should not change. 

Compare the bad behavior to the single-select functionality (selectedValues is not an array): https://codesandbox.io/p/sandbox/ariakit-combobox-infinity-scroll-03bys?file=%2Fsrc%2FApp.js
^ Here, the scroll is not interrupted when new pages are loaded. 

### Workaround

https://github.com/ariakit/ariakit/issues/5165#issuecomment-3604432910

### Possible solutions

_No response_

diegohaz:
Closes #5165

## Problem

When `selectedValues` is an array (multi-select combobox), `resetValueOnSelect` defaults to `true`. The auto-select effect in the combobox component uses `resetValueOnSelect` as a gate to prevent bail-out:

```ts
if (!canAutoSelect && !resetValueOnSelect) return;
```

This means when `resetValueOnSelect` is true, the effect never bails out — even when items change due to infinite scroll loading (not due to a value reset). The effect then calls `scrollIntoView({ block: "nearest" })` on the first item, which resets the popover's scroll position to the top.

## Solution

Add a `userScrolledRef` that tracks when the user has actively scrolled (via `wheel` or `touchmove` events). When the user has scrolled, the effect bails out even if `resetValueOnSelect` is true:

```ts
if (!canAutoSelect && (!resetValueOnSelect || userScrolledRef.current)) return;
```

The ref is cleared when the combobox value changes, so that value reset after item selection still correctly scrolls the active item into view.

## Workaround

Set `resetValueOnSelect={false}` on `ComboboxProvider`:

```jsx
<ComboboxProvider
  selectedValue={selectedValues}
  setSelectedValue={setSelectedValues}
  // TODO: Remove when https://github.com/ariakit/ariakit/issues/5165 is fixed
  resetValueOnSelect={false}
>
```

## Test plan

- [x] Failing test reproduces the scroll reset bug
- [x] Workaround passes the test
- [x] Library fix passes the test
- [x] All combobox browser tests pass (20 tests including `combobox-multiple`)
- [x] Existing `combobox-multiple` tests for scroll-after-select and scroll-after-hover still pass

### DIFF
--- a/packages/ariakit-react-core/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-core/src/combobox/combobox.tsx
@@ -344,9 +344,10 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       };
     }, [open, contentElement, store]);
 
-    // Set the changed flag to true whenever the combobox value changes and is
-    // not empty. We're doing this here in addition to in the onChange handler
-    // because the value may change programmatically.
+    // Reset the user-scrolled flag and set the changed flag to true whenever
+    // the combobox value changes and is not empty. We're doing this here in
+    // addition to in the onChange handler because the value may change
+    // programmatically.
     useSafeLayoutEffect(() => {
       userScrolledRef.current = false;
       if (!storeValue) return;
--- a/site/src/sandbox/5165-combobox/test-browser.ts
+++ b/site/src/sandbox/5165-combobox/test-browser.ts
@@ -25,7 +25,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.expect(scrollTop).toBeGreaterThan(50);
   });
 
-  test("popover scroll position is preserved on scrollbar drag", async ({
+  test("popover scroll position is preserved on non-wheel scroll", async ({
     q,
   }) => {
     const popover = q.listbox();


packages/ariakit-react-core/src/combobox/combobox.tsx
```diff
@@ -312,6 +313,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
         // A wheel event is always initiated by the user, so we can disable the
         // autoSelect behavior without any additional checks.
         canAutoSelectRef.current = false;
+        userScrolledRef.current = true;
```
diegohaz, line 317: The new bailout still depends on `wheel`/`touchmove` being the thing that marks the scroll as user-initiated. Scroll paths that don't emit either event before the list moves still miss this guard, so the async item update can call `scrollIntoView` and jump the popover back to the top.

I verified that locally by removing the synthetic `WheelEvent` from the new Playwright test and rerunning `pnpm -F site test-chrome site/src/sandbox/5165-combobox/test-browser.ts`; the run then fails with `scrollTop` back at `0`.

Can we cover the non-wheel user-scroll path as well and add a browser test that doesn't rely on dispatching a wheel event? Like scrollbar-drag.

site/src/sandbox/5165-combobox/test-browser.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const combobox = q.combobox();
+    await combobox.focus();
+
+    // Scroll the popover to near the bottom to trigger loading more items.
+    // Dispatch a wheel event first so Ariakit knows this is user-initiated.
+    await popover.evaluate((el) => {
+      el.dispatchEvent(new WheelEvent("wheel", { bubbles: true }));
+      el.scrollTop = el.scrollHeight - el.clientHeight - 50;
+    });
```
diegohaz: Shouldn't we use Playwright's scrolling mechanisms instead? See https://playwright.dev/docs/input#scrolling

Also, do we need a `test-mobile.ts` file too or is this enough?

site/src/sandbox/5165-combobox/test-browser.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    // Wait for the onScroll handler to fire and new items to render
+    await page.waitForTimeout(500);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 71f548c43 - Replaced `waitForTimeout(500)` with `expect.poll()` for option count.

packages/ariakit-react-core/src/combobox/combobox.tsx
diegohaz: Fixed in 71f548c43 - Added `isAutoScrollingRef` to distinguish programmatic `scrollIntoView` calls from user-initiated scrolls. The generic `onScroll` handler now also sets `userScrolledRef = true` when `isAutoScrollingRef` is false, covering scrollbar drag and any other non-wheel/touchmove scroll path. The flag is cleared via `requestAnimationFrame` since scroll events fire before rAF callbacks in the browser rendering update cycle.

Added a second test ("popover scroll position is preserved on scrollbar drag") that scrolls programmatically via `el.scrollTop = ...` without dispatching any wheel event.

site/src/sandbox/5165-combobox/test-browser.ts
diegohaz: Fixed in 71f548c43 - Updated the first test to use `page.mouse.wheel()` instead of synthetic `WheelEvent` dispatch. Added a second test that scrolls via programmatic `scrollTop` (simulating scrollbar drag) without any wheel event. Replaced `waitForTimeout` with `expect.poll`.

Regarding `test-mobile.ts`: the fix already covers `touchmove` via the `onUserScroll` handler, and the generic `onScroll` handler now catches all remaining scroll paths. I think the browser tests should be sufficient since the mobile scroll codepath (`touchmove` → `canAutoSelect = false` + `userScrolled = true`) is the same as wheel. Let me know if you'd still like a dedicated mobile test.

diegohaz (commented):
Re-review of commits since the last marked review head `dc5ea07e899f1197632b9afbd1d97efb2e1abcee`, now at `4eb012d7177c34536ead2b127b9c0162b0a367c9`.

Reviewed the incremental diff in `packages/ariakit-react-core/src/combobox/combobox.tsx` and `site/src/sandbox/5165-combobox/test-browser.ts`, plus the surrounding `useCombobox` auto-select logic and the prior review thread replies.

Validation performed:
- verified in the current code that the previous non-wheel scroll finding is addressed via `isAutoScrollingRef` and the widened `onScroll` handling
- `pnpm -F site test-chrome site/src/sandbox/5165-combobox/test-browser.ts`

No new substantive findings in this incremental range. The previous finding appears fixed, and the updated sandbox test coverage now exercises both wheel-driven and non-wheel scroll paths.

Not fully validated: non-Chrome browsers and the broader combobox browser suite.

packages/ariakit-react-core/src/combobox/combobox.tsx
```diff
@@ -338,6 +348,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     // not empty. We're doing this here in addition to in the onChange handler
     // because the value may change programmatically.
     useSafeLayoutEffect(() => {
+      userScrolledRef.current = false;
       if (!storeValue) return;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 00176261d - Updated the comment to mention the `userScrolledRef` reset.

site/src/sandbox/5165-combobox/test-browser.ts
```diff
@@ -0,0 +1,52 @@
...
+  }) => {
+    const popover = q.listbox();
+    await test.expect(popover).toBeVisible();
+
+    const combobox = q.combobox();
+    await combobox.focus();
+
+    // Scroll programmatically without a wheel event to simulate scrollbar
+    // drag or other non-wheel scroll paths
+    await popover.evaluate((el) => {
```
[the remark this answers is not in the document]
diegohaz: Fixed in 00176261d - Renamed the test to "popover scroll position is preserved on non-wheel scroll" to better describe what is being covered.

diegohaz:
**Fixed**

- 00176261d ([reply](https://github.com/ariakit/ariakit/pull/5613#discussion_r2968321224)): Updated comment to mention the `userScrolledRef` reset alongside the changed flag.
- 00176261d ([reply](https://github.com/ariakit/ariakit/pull/5613#discussion_r2968321301)): Renamed test from "scrollbar drag" to "non-wheel scroll" to better describe the scenario.

PR #5615, opened 2026-03-21
Deprecate implicit store provider contexts

diegohaz:
## Summary
- allow `store` props to receive provider components so compatible provider contexts can be read explicitly
- warn when components implicitly read stores from compatible provider contexts without an explicit `store` prop
- add legacy example and site sandbox browser coverage for both the deprecated and migrated patterns

## Testing
- `pnpm tsc --pretty false`
- `pnpm test-chrome examples/store-provider-deprecated/test-browser.ts examples/store-provider-migrated/test-browser.ts`
- `pnpm -F site test-chrome src/sandbox/store-provider-deprecated/test-browser.ts src/sandbox/store-provider-migrated/test-browser.ts`


.changeset/200-store-provider-components.md
```diff
@@ -0,0 +1,16 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Improved `store` prop migration for provider context fallbacks
+
+Components that accept a [`store`](https://ariakit.org/reference/combobox#store) prop can now receive a provider component, such as `store={ComboboxProvider}`, to read the corresponding context explicitly across compatible providers.
+
+A deprecation warning is now logged when those components still rely on implicit compatible provider context fallbacks. Pass the provider component to the `store` prop to keep the current behavior without the warning while preparing for the future breaking change.
```
diegohaz: I feel like this makes it seem as if the new feature is just a workaround, when it's actually a genuine feature. I believe we should have two separate changeset files, one for the deprecation warning, with the migration path being the new feature, and another for the new feature.

examples/store-provider-deprecated/index.react.tsx
```diff
@@ -0,0 +1,33 @@
+import {
```
diegohaz: It should just be the sandbox example.

examples/store-provider-migrated/index.react.tsx
```diff
@@ -0,0 +1,33 @@
+import {
```
diegohaz: It should be just the sandbox example.

packages/ariakit-react-core/src/combobox/combobox-context.tsx
```diff
@@ -17,6 +17,7 @@ export const ComboboxListRoleContext = createContext<string | undefined>(
 const ctx = createStoreContext<ComboboxStore>(
   [PopoverContextProvider, CompositeContextProvider],
   [PopoverScopedContextProvider, CompositeScopedContextProvider],
+  "ComboboxProvider",
```
diegohaz: Is the string argument required or optional? If it's required, I believe it should be the first parameter so we can avoid passing `[], []` when we're not using other contexts.

packages/ariakit-react-core/src/combobox/combobox-disclosure.tsx
```diff
@@ -52,8 +53,7 @@ export const useComboboxDisclosure = createHook<
   TagName,
   ComboboxDisclosureOptions
 >(function useComboboxDisclosure({ store, ...props }) {
-  const context = useComboboxProviderContext();
-  store = store || context;
+  store = useComboboxProviderContextStore(store, "ComboboxDisclosure");
```
diegohaz, line 56: Why do we have to pass `"ComboboxDisclosure"` (and other component names) as an argument?

packages/ariakit-react-core/src/combobox/combobox-provider.tsx
```diff
@@ -43,6 +46,8 @@ export function ComboboxProvider(props: ComboboxProviderProps = {}) {
   );
 }
 
+registerComboboxProvider(ComboboxProvider);
```
diegohaz, line 49: Don't these register functions cause side effects in the module?

packages/ariakit-react-core/src/composite/composite-container.tsx
```diff
@@ -11,9 +10,11 @@ import type { ElementType, FocusEvent, KeyboardEvent, MouseEvent } from "react";
 import { useEffect, useRef } from "react";
 import { useEvent, useMergeRefs } from "../utils/hooks.ts";
 import { useStoreState } from "../utils/store.tsx";
+// TODO: Add data-attribute to indicate whether it's expanded?
```
diegohaz: This comment should be moved back to the top.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -566,11 +567,25 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+  componentName = Component.displayName || Component.name || "Dialog",
 ) {
+  const getProviderContextStore =
+    useProviderContextStore ||
+    (useDialogProviderContextStore as (
+      store: T["store"],
+      componentName: string,
+    ) => NonNullable<T["store"]> extends StoreProp<infer S>
+      ? S | undefined
+      : never);
```
diegohaz: This is too complex.

packages/ariakit-react-core/src/menu/menu-item-radio.tsx
```diff
@@ -84,6 +85,16 @@ export const useMenuItemRadio = createHook<TagName, MenuItemRadioOptions>(
...
+    const radioValue = useStoreState(store, (state) => {
+      const nextValue = state.values[name];
+      if (typeof nextValue === "string" || typeof nextValue === "number") {
+        return nextValue;
+      }
+      return undefined;
+    });
+    const radioStore = useRadioStore({
+      value: radioValue,
+    });
```
diegohaz: Why is this necessary?

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -70,19 +77,32 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
...
+          "MenuItem",
+        )
+      : useMenuScopedContext(true);
+    const menubarStore = store
+      ? useMenubarScopedContextStore(
+          store as StoreProp<MenubarStore> | undefined,
+          "MenuItem",
+        )
+      : useMenubarScopedContext();
+    const resolvedStore = menuStore || menubarStore;
```
diegohaz: Too complicated. Can't we do this without type casting?

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -70,19 +77,32 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
...
       process.env.NODE_ENV !== "production" &&
         "MenuItem must be wrapped in a MenuList, Menu or Menubar component",
     );
 
     const onClickProp = props.onClick;
     const hideOnClickProp = useBooleanEvent(hideOnClick);
-    const hideMenu = "hideAll" in store ? store.hideAll : undefined;
+    const hideMenu =
+      "hideAll" in resolvedStore
+        ? (resolvedStore.hideAll as MenuStore["hideAll"])
```
diegohaz: Same. Can't this be done without type casting?

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -98,8 +118,10 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
       hideMenu();
     });
 
-    const contentElement = useStoreState(store, (state) =>
-      "contentElement" in state ? state.contentElement : null,
+    const contentElement = useStoreState(resolvedStore, (state) =>
+      "contentElement" in state
+        ? (state.contentElement as MenuStoreState["contentElement"])
```
diegohaz: Same thing about type casting.

packages/ariakit-react-core/src/tag/tag-input.tsx
```diff
@@ -76,8 +77,7 @@ export const useTagInput = createHook<TagName, TagInputOptions>(
     removeOnBackspace = true,
     ...props
   }) {
-    const context = useTagContext();
-    store = store || context;
+    store = useTagContextStore(store, "TagInput");
```
diegohaz, line 80: Instead of having one function per module, maybe we could just use a single `useContextStore` function. Can we simplify things?

vitest.setup.ts
```diff
@@ -10,7 +10,17 @@ import {
...
+  "examples/store-provider-deprecated/test.ts",
+  "site/src/sandbox/store-provider-deprecated/test.ts",
+];
+
+failOnConsole({
+  skipTest({ testPath }: { testPath?: string }) {
+    if (!testPath) return false;
+    return skipConsoleGuardPaths.some((path) => testPath.endsWith(path));
+  },
+});
```
diegohaz: We shouldn't need this change. See https://github.com/thomasbrodusch/vitest-fail-on-console?tab=readme-ov-file#but-i-have-some-expected-console-errorswarning

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -83,17 +83,123 @@ type StoreProvider<T extends Store> = React.ComponentType<{
...
+function useStoreContext<T extends Store>(
+  store: StoreProp<T> | undefined,
+  metadata: StoreContextMetadata<T>,
+  componentName: string,
+  mode: StoreContextMode,
+) {
+  const providerMetadata = getStoreContextMetadata(store);
+  const explicitStore = providerMetadata
+    ? getStoreFromMetadata<T>(providerMetadata, mode)
+    : (store as T | undefined);
```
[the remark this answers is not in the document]
diegohaz: Fixed in d6cac062f - invalid function `store` values are no longer treated as store objects. They now warn clearly instead of being misinterpreted.

.changeset/200-store-provider-components.md
diegohaz: Fixed in d6cac062f - I split this into separate changesets for the new `store={Provider}` feature and the deprecation warning.

examples/store-provider-deprecated/index.react.tsx
diegohaz: Fixed in d6cac062f - I removed the duplicate root example and kept this as a sandbox-only example.

examples/store-provider-migrated/index.react.tsx
diegohaz: Fixed in d6cac062f - I removed the duplicate root example and kept this as a sandbox-only example.

packages/ariakit-react-core/src/combobox/combobox-context.tsx
diegohaz: Fixed in d6cac062f - `createStoreContext` now accepts the provider name as the first argument, so the simple cases no longer need `[], []`.

packages/ariakit-react-core/src/combobox/combobox-disclosure.tsx
diegohaz: This is still needed because the warning and invariant messages need the actual component name, not just the provider metadata name. Without passing it through, everything would collapse to provider-derived labels like `ComboboxProvider`.

packages/ariakit-react-core/src/combobox/combobox-provider.tsx
diegohaz: Fixed in d6cac062f - `ComboboxProvider` now gets its metadata in the export expression instead of through a separate module-level registration call.

packages/ariakit-react-core/src/composite/composite-container.tsx
diegohaz: Fixed in d6cac062f - I moved the TODO comment back above the helper.

packages/ariakit-react-core/src/dialog/dialog.tsx
diegohaz: Fixed in d6cac062f - I pulled the conditional store type into named aliases so `createDialogComponent` is flatter.

packages/ariakit-react-core/src/menu/menu-item-radio.tsx
diegohaz: Resolved in d6cac062f - this bridge is still necessary because `useRadio` expects a radio store for its checked-state and keyboard logic. Without it, `MenuItemRadio` falls back through composite context and reintroduces the deprecation warning. I added a clarifying code comment there.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in b86a32147 - `useStoreContext` now accepts broader store props, so `MenuItem` no longer needs the explicit casts when resolving menu and menubar stores.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in b86a32147 - same cleanup here; the cast is gone after widening the shared store-context helper inputs.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in b86a32147 - I replaced the remaining `hideAll` cast with a type guard.

packages/ariakit-react-core/src/tag/tag-input.tsx
diegohaz: This is not actually simpler because the per-module helpers capture both the correct context metadata and the lookup mode (`context`, `scoped`, or `provider`) once. A single generic hook would push that metadata and mode back into every callsite.

vitest.setup.ts
diegohaz: Fixed in d6cac062f - I switched this from a path-based skip to `allowMessage`, and the sandbox tests now assert against the warnings collected through that allowed-message path.

packages/ariakit-react-core/src/menu/menu-item-radio.tsx
```diff
@@ -84,6 +85,18 @@ export const useMenuItemRadio = createHook<TagName, MenuItemRadioOptions>(
...
+      if (typeof nextValue === "string" || typeof nextValue === "number") {
+        return nextValue;
+      }
+      return undefined;
+    });
+    // MenuItemRadio manages a menu store, but useRadio expects a radio store
+    // for its checked state logic and keyboard behavior.
+    const radioStore = useRadioStore({
+      value: radioValue,
+    });
```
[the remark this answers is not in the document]
diegohaz: Fixed in c64341600 - `MenuItemRadio` no longer creates a separate radio store. It now uses `useRadio` in an internal standalone mode, so it keeps the radio DOM semantics without reading from or registering with a second composite store.

packages/ariakit-react-core/src/menu/menu-item-radio.tsx
```diff
@@ -101,6 +114,7 @@ export const useMenuItemRadio = createHook<TagName, MenuItemRadioOptions>(
     };
 
     props = useRadio<TagName>({
+      store: radioStore,
       name,
       value,
       checked: isChecked,
```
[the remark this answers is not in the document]
diegohaz: Fixed in c64341600 - `MenuItemRadio` now derives its checked state entirely from the menu store while `useRadio({ unstable_standalone: true })` provides only the radio DOM wiring. That removes the extra radio-store/composite behavior from this path.

site/src/sandbox/store-provider-migrated/test.ts
```diff
@@ -0,0 +1,10 @@
+import { click, q, sleep, waitFor } from "@ariakit/test";
+
+test("doesn't warn when the provider component is passed as store", async () => {
+  await click(q.button("Show combobox"));
+  await waitFor(() => expect(q.button("Apple")).toBeVisible());
+  await sleep(0);
+  const warnings = (globalThis as { __allowedConsoleMessages?: string[] })
+    .__allowedConsoleMessages;
+  expect(warnings).toEqual([]);
+});
```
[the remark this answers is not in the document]
diegohaz: This is intentional. These sandbox warning assertions were moved to dev-mode Vitest so they can exercise the warning path. The root `vitest.config.ts` includes `**/*test.ts`, so this file is covered by `pnpm test --run` even though it is not part of the site Playwright suite.

site/src/sandbox/store-provider-deprecated/test.ts
```diff
@@ -0,0 +1,15 @@
...
+  "Pass `store={ComboboxProvider}` to keep the current behavior.";
+
+test("warns on implicit compatible provider context", async () => {
+  await click(q.button("Show combobox"));
+  await waitFor(() => expect(q.button("Apple")).toBeVisible());
+  await sleep(0);
+  const warnings = (globalThis as { __allowedConsoleMessages?: string[] })
+    .__allowedConsoleMessages;
+  expect(warnings).toContain(message);
+});
```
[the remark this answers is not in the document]
diegohaz: This is intentional. The warning assertion is exercised by the root dev-mode Vitest suite, where `vitest.setup.ts` defines `globalThis.__allowedConsoleMessages`. This sandbox is not meant to run under the site Playwright config.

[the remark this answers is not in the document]
diegohaz: This is intentional. This sandbox coverage lives in the root Vitest suite so it can assert the dev-only warning behavior; it is not intended to be a site Playwright test.

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -83,17 +83,143 @@ type StoreProvider<T extends Store> = React.ComponentType<{
...
+    implicitSource !== metadata &&
+    !!implicitSource.name;
+
+  React.useEffect(() => {
+    if (invalidStore) {
+      console.warn(
+        `Invalid \`store\` prop passed to ${componentName}. ` +
+          "Expected a store object or an Ariakit provider component.",
+      );
+      return;
```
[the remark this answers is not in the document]
diegohaz: Fixed in c64341600 - the invalid `store` warning is now gated to non-production builds like the other developer-facing warnings.

.changeset/300-deprecate-implicit-store-provider-contexts.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Deprecation warning for implicit compatible provider context fallbacks
+
+Components that still rely on implicit compatible provider context fallbacks now
+log a warning in development. Pass the provider component to the `store` prop to
+keep the current behavior while preparing for the future breaking change.
```
diegohaz: - Update the title to start with `Deprecated...`
- Add before and after code blocks
- The prefix should be 290-

packages/ariakit-react-core/src/composite/composite-container.tsx
```diff
@@ -11,16 +10,18 @@ import type { ElementType, FocusEvent, KeyboardEvent, MouseEvent } from "react";
...
-import { useCompositeContext } from "./composite-context.tsx";
+import { useCompositeContextStore } from "./composite-context.tsx";
 import type { CompositeStore } from "./composite-store.ts";
 import { selectTextField } from "./utils.ts";
 
 const TagName = "div" satisfies ElementType;
 type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
 
+// TODO: Add data-attribute to indicate whether it's expanded?
```
diegohaz: This comment was at the top of the file before the imports.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -566,11 +567,15 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
 
 export function createDialogComponent<T extends DialogOptions>(
   Component: FC<T>,
-  useProviderContext = useDialogProviderContext,
+  useProviderContextStore?: DialogProviderContextStore<T>,
+  componentName = Component.displayName || Component.name || "Dialog",
 ) {
+  const getProviderContextStore =
+    useProviderContextStore ||
+    (useDialogProviderContextStore as DialogProviderContextStore<T>);
```
diegohaz: Is this type cast necessary?

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -70,19 +81,25 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
...
   }) {
-    const menuContext = useMenuScopedContext(true);
-    const menubarContext = useMenubarScopedContext();
-    store = store || menuContext || (menubarContext as any);
+    const menuStore = store
+      ? useMenuScopedContextStore(store, "MenuItem")
+      : useMenuScopedContext(true);
+    const menubarStore = store
+      ? useMenubarScopedContextStore(store, "MenuItem")
+      : useMenubarScopedContext();
```
diegohaz: React hooks can't be conditional.

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -98,8 +115,10 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
       hideMenu();
     });
 
-    const contentElement = useStoreState(store, (state) =>
-      "contentElement" in state ? state.contentElement : null,
+    const contentElement = useStoreState(resolvedStore, (state) =>
+      "contentElement" in state
+        ? (state.contentElement as MenuStoreState["contentElement"])
```
diegohaz: Is this `as MenuStoreState["contentElement"]` necessary?

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -70,19 +81,25 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
...
-    const menuContext = useMenuScopedContext(true);
-    const menubarContext = useMenubarScopedContext();
-    store = store || menuContext || (menubarContext as any);
+    const menuStore = store
+      ? useMenuScopedContextStore(store, "MenuItem")
+      : useMenuScopedContext(true);
+    const menubarStore = store
+      ? useMenubarScopedContextStore(store, "MenuItem")
+      : useMenubarScopedContext();
+    const resolvedStore = menuStore || menubarStore;
```
diegohaz: I'd rather call it just `store`. If it can't assign to the prop, just rename the prop `store: storeProp`.

packages/ariakit-react-core/src/radio/radio.tsx
```diff
@@ -62,10 +63,16 @@ export const useRadio = createHook<TagName, RadioOptions>(function useRadio({
   name,
   value,
   checked,
+  unstable_standalone = false,
```
diegohaz: `store` is not required in either the `Radio` or `CompositeItem` components. So I believe that, instead of this `unstable_standalone` prop, we can accept `null` as a value for all optional `store` props. In that case, the store won't be used, and the component won't try to get it from context, either directly or indirectly, so there will be no deprecation warning.

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -83,17 +83,144 @@ type StoreProvider<T extends Store> = React.ComponentType<{
...
+  const shouldWarn =
+    !store &&
+    !!resolvedStore &&
+    !!implicitSource &&
+    implicitSource !== metadata &&
+    !!implicitSource.name;
+
+  React.useEffect(() => {
+    if (invalidStore) {
+      if (process.env.NODE_ENV === "production") return;
```
diegohaz: This production check should be at the beginning of the effect so it applies to the entire effect. Also, can you confirm whether this early return works for dead code elimination? Otherwise, we'd have to wrap the entire effect body in a non-production condition.

site/src/sandbox/store-provider-deprecated/test.react.tsx
```diff
@@ -0,0 +1,25 @@
...
+const message =
+  "CompositeItem is reading its store from ComboboxProvider implicitly. " +
+  "This is deprecated and will stop working in a future version. " +
+  "Pass `store={ComboboxProvider}` to keep the current behavior.";
+
+test("warns on implicit compatible provider context", async () => {
+  const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+  const { unmount } = await render(createElement(Example), {
+    strictMode: true,
+  });
```
diegohaz: Does this work if we rely on the Vitest setup to render the component instead of using `render` here? We might need to use vi.spyOn in a beforeEach hook, but I don't know whether the beforeEach in the test file runs before the one in the Vitest setup where the component is rendered.

site/src/sandbox/store-provider-migrated/test.react.tsx
```diff
@@ -0,0 +1,20 @@
+import { click, q, sleep, waitFor } from "@ariakit/test";
+import { render } from "@ariakit/test/react";
+import { createElement } from "react";
+import Example from "./index.react.tsx";
+
+test("doesn't warn when the provider component is passed as store", async () => {
+  const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+  const { unmount } = await render(createElement(Example), {
+    strictMode: true,
+  });
```
diegohaz: Same as above.

vitest.setup.ts
```diff
@@ -132,9 +132,15 @@ function parseTest(filename?: string) {
 
 const LOADER = (process.env.ARIAKIT_TEST_LOADER ??
   "react") as AllowedTestLoader;
+const MANUAL_LOAD_TESTS = new Set([
+  "site/src/sandbox/store-provider-deprecated/test.react.tsx",
+  "site/src/sandbox/store-provider-migrated/test.react.tsx",
+]);
```
diegohaz: I don't like this. We should find a way to leverage the automatic load.

.changeset/300-deprecate-implicit-store-provider-contexts.md
diegohaz: Fixed in 7f407d1ee - I renamed this changeset to the `290-` prefix, updated the title to start with `Deprecated`, and added before/after migration examples.

packages/ariakit-react-core/src/composite/composite-container.tsx
diegohaz: Fixed in 7f407d1ee - I moved the TODO comment back to the top of the file above the imports.

packages/ariakit-react-core/src/dialog/dialog.tsx
diegohaz: Fixed in 7f407d1ee - the fallback now branches inline instead of storing `useDialogProviderContextStore` through a casted variable.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in 7f407d1ee - `useMenuItem` no longer calls hooks conditionally. It now reads both context paths unconditionally and preserves the original fallback semantics when `store` is omitted.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in 7f407d1ee - that `contentElement` cast is gone now; the selector narrows the state before reading the property.

packages/ariakit-react-core/src/menu/menu-item.tsx
diegohaz: Fixed in 7f407d1ee - I renamed the prop binding to `storeProp` and switched the resolved local variable back to `store`.

packages/ariakit-react-core/src/radio/radio.tsx
diegohaz: Fixed in 7f407d1ee - I removed `unstable_standalone` and added explicit `store={null}` semantics to optional store props. `MenuItemRadio` now uses `useRadio({ store: null, ... })` to opt out of context without introducing a private prop.

packages/ariakit-react-core/src/utils/system.tsx
diegohaz: Fixed in 7f407d1ee - the production guard now returns at the start of the effect so it applies to the entire warning effect.

site/src/sandbox/store-provider-deprecated/test.react.tsx
diegohaz: Fixed in 7f407d1ee - this uses the automatic example loader again. The test now installs its `console.warn` spy inside the test body before the interaction that triggers the warning, so there is no manual render path anymore.

site/src/sandbox/store-provider-migrated/test.react.tsx
diegohaz: Fixed in 7f407d1ee - same cleanup here; the migrated sandbox test uses the automatic loader again and spies on `console.warn` only around the interaction under test.

vitest.setup.ts
diegohaz: Fixed in 7f407d1ee - the manual-load exception was removed from `vitest.setup.ts`, and these sandbox tests now rely on the standard automatic example loading path again.

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -79,21 +79,153 @@ export function createHook<
...
+  name: string;
+  useContext: () => T | undefined;
+  useScopedContext: (onlyScoped?: boolean) => T | undefined;
+  useProviderContext: () => T | undefined;
+  useContextSource: () => StoreContextMetadata | undefined;
+  useScopedContextSource: (
+    onlyScoped?: boolean,
+  ) => StoreContextMetadata | undefined;
+  useProviderContextSource: () => StoreContextMetadata | undefined;
+}
```
diegohaz, line 104: `StoreContextMetadata` has 7 fields, 6 of which are hooks. Combined with the `contextSource` and `scopedContextSource` React contexts (2 new contexts per store type × ~18 store types = ~36 new React contexts), this is a lot of infrastructure for deprecation warnings.

The `name` field is the only reason this metadata exists beyond the hooks themselves — it's used to produce the warning message `"ComponentX is reading its store from ProviderY implicitly"`. If a generic warning is acceptable (e.g. `"ComponentX is reading its store from a compatible provider implicitly"`), the name and source-tracking contexts could be eliminated entirely, and the warning could be produced by comparing `metadata !== sourceMetadata` without needing to thread source through contexts.

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -79,21 +79,153 @@ export function createHook<
...
+    warnedStoreContexts.add(key);
+    console.warn(
+      `${componentName} is reading its store from ${providerName} implicitly. ` +
+        "This is deprecated and will stop working in a future version. " +
+        `Pass \`store={${providerName}}\` to keep the current behavior.`,
+    );
+  }, [componentName, implicitSource, invalidStore, shouldWarn]);
+
+  return resolvedStore;
+}
```
diegohaz, line 198: `useStoreContext` always calls both `getStoreFromMetadata` (implicit) and `getStoreSourceFromMetadata` even when `store` is explicitly provided. The implicit hooks call `React.useContext` unconditionally (hooks can't be conditional), so every component now pays for 2-4 extra context reads per render even when a store is passed explicitly. This is correct (hooks rules), but worth noting as a permanent runtime cost.

packages/ariakit-react-core/src/utils/system.tsx
```diff
@@ -102,44 +234,140 @@ export function createStoreContext<T extends Store>(
...
+            __ariakitContext={__ariakitContext}
+          >
+            {children}
+          </InternalProvider>
+        );
+      },
+      <contextSource.Provider value={__ariakitContext}>
+        <context.Provider {...providerProps} />
+      </contextSource.Provider>,
     );
```
diegohaz, line 291: The `__ariakitContext` prop is threaded through every parent provider in the `reduceRight` chain. This means:
1. Every `StoreProvider` component in the chain receives an undocumented internal prop
2. The `InternalStoreProvider` type cast is needed at every usage site
3. If a provider in the chain doesn't forward `__ariakitContext`, the source tracking silently breaks

This is the most fragile part of the implementation. If the deprecation path is kept, consider whether source can be tracked via a single React context at the `createStoreContext` level instead of threading props through the provider chain.

packages/ariakit-react-core/src/menu/menu-item.tsx
```diff
@@ -63,16 +83,27 @@ function menuHasFocus(
...
+      "MenuItem",
+    );
+    const menubarProviderStore = useMenubarScopedContextStore(
+      storeProp,
+      "MenuItem",
+    );
+    const store =
+      storeProp === undefined
+        ? menuContextStore || menubarContextStore
+        : menuProviderStore || menubarProviderStore;
```
diegohaz, line 106: This store resolution now uses 4 hooks + a ternary:
```ts
const menuContextStore = useMenuScopedContext(true);
const menubarContextStore = useMenubarScopedContext();
const menuProviderStore = useMenuScopedContextStore(storeProp, "MenuItem");
const menubarProviderStore = useMenubarScopedContextStore(storeProp, "MenuItem");
const store = storeProp === undefined
  ? menuContextStore || menubarContextStore
  : menuProviderStore || menubarProviderStore;
```

The old code was:
```ts
store = store || menuContext || menubarContext;
```

This is a significant readability regression. The dual-context case (`MenuStore | MenubarStore`) is admittedly hard, but this pattern would benefit from a helper like `useContextStore(store, [MenuContext, MenubarContext])` that handles the fallback chain internally.

diegohaz (commented):
## Review scope

Initial review of the full diff from `main` to `c35add4b2`. Reviewed all 173 changed files, focused on:

- `packages/ariakit-react-core/src/utils/system.tsx` (core infrastructure)
- Representative context files (`combobox-context.tsx`, `composite-context.tsx`)
- Representative component files (`combobox.tsx`, `combobox-item.tsx`, `menu-item.tsx`, `dialog.tsx`)
- Provider registration (`combobox-provider.tsx`)
- Examples and tests
- Changeset

No local test runs (worktree has no `node_modules`). Analysis is code-reading only.

---

## Overall assessment

The PR works as described: it deprecates implicit cross-context store fallback and provides a migration path via `store={ComboboxProvider}`. The implementation is correct but introduces significant complexity. The system now has **4 new React contexts** per store type (two source-tracking contexts), a **`StoreContextMetadata` object with 7 fields**, **`__ariakitContext` prop threading** through all parent providers, **`registerProvider`** calls in every provider file, and **3 new hooks per context module** (18 context files × 3 = 54 new exports + 18 `registerProvider` exports). This is a high ongoing maintenance cost for a deprecation warning.

### Key findings

1. **The source-tracking infrastructure is the primary complexity driver** — `contextSource`, `scopedContextSource`, `StoreContextMetadata`, and `__ariakitContext` prop threading exist solely to detect "which provider set this context value". All of this is only needed for the deprecation warning message. If the deprecation is replaced by a breaking change (remove implicit fallback entirely), or if a generic warning is acceptable, this entire layer disappears.

2. **The per-module hook exports (`useXxxContextStore`, `useXxxScopedContextStore`, `useXxxProviderContextStore`) are thin wrappers** that only bind the metadata and mode. They could be replaced by a single generic function that takes the context object directly, e.g. `useContextStore(store, ctx, "context")`, significantly reducing the per-module boilerplate.

3. **`MenuItem` shows how dual-context components get harder** — The `MenuItem` store resolution now calls 4 hooks (`useMenuScopedContext`, `useMenubarScopedContext`, `useMenuScopedContextStore`, `useMenubarScopedContextStore`) and has a ternary to decide between them. The old code was `store || menuContext || menubarContext`.

4. **Breaking change path would be dramatically simpler** — If the implicit cross-context fallback is simply removed (components only read from their own context type), there is no need for source tracking, `__ariakitContext`, `registerProvider`, `StoreProp`, or the per-module `useXxxContextStore` hooks. The `providers` array in `createStoreContext` still serves its purpose (making stores available to child providers), but the warning infrastructure vanishes. The trade-off: users of patterns like `<PopoverArrow>` inside `<SelectPopover>` would need to pass an explicit store prop — but that's a clearer API anyway.

### Areas not fully validated

- Runtime behavior (no test runs)
- Performance impact of the additional React contexts and hooks per render

packages/ariakit-react-core/src/combobox/combobox-context.tsx
```diff
@@ -33,15 +34,20 @@ const ctx = createStoreContext<ComboboxStore>(
  * }
  */
 export const useComboboxContext = ctx.useContext;
+export const useComboboxContextStore = ctx.useContextStore;
```
diegohaz, line 37: Every context file now exports 3 new hooks (`useXxxContextStore`, `useXxxScopedContextStore`, `useXxxProviderContextStore`) plus `registerXxxProvider`. These are all thin wrappers around `ctx.useContextStore(store, componentName)` etc.

Consider exporting the `ctx` object directly (or a subset of it) so components can call `ctx.useContextStore(store, "Combobox")` without the re-export layer. This would cut ~72 new exports across the codebase.

examples/select-grid/index.react.tsx
```diff
@@ -37,7 +37,10 @@ export default function Example() {
           <Ariakit.SelectArrow />
         </Ariakit.Select>
         <Ariakit.SelectPopover role="grid" className="popover">
-          <Ariakit.PopoverArrow className="arrow" />
+          <Ariakit.PopoverArrow
```
diegohaz, line 40: `store={Ariakit.SelectProvider}` is a novel API pattern — passing a component reference as a prop value. While it works mechanically, it may confuse users because:
1. It looks like you're passing a component, not a store
2. The component reference is used as a key to look up context metadata, not rendered
3. It only works with registered provider components, not arbitrary components

If the breaking change path is chosen instead, this pattern goes away entirely and users pass store objects directly, which is more conventional.

diegohaz:
Closing this as I don't think this deprecation path is still viable. We'll probably need to introduce this as a breaking change in v0.5.

PR #5617, opened 2026-03-23
Migrate domain from ariakit.org to ariakit.com

diegohaz:
## Summary

- Update all hardcoded `ariakit.org` references to `ariakit.com` across 264 files
- Newsletter URLs temporarily use `diegohaz.substack.com` until Substack verifies the `newsletter.ariakit.com` custom domain
- CSP frame-ancestors and hostname routing include both `.com` and `.org` for the transition period
- CHANGELOG files left unchanged (historical records; old URLs will 301 redirect)
- `support@ariakit.org` replaced with `contact@ariakit.com` as the canonical contact email

## Test plan

- [ ] Verify sitemap at `/sitemap.xml` contains `ariakit.com` URLs
- [ ] Verify `robots.txt` references `ariakit.com/sitemap.xml`
- [ ] Verify newsletter form submission works via `diegohaz.substack.com`
- [ ] Verify Clerk sign-in/sign-up works on `ariakit.com`
- [ ] Verify Stripe checkout flow completes
- [ ] Verify hovercard/tooltip examples render correctly with updated URLs

### DIFF
--- a/website/components/footer.tsx
+++ b/website/components/footer.tsx
@@ -23,7 +23,7 @@ const links = [
       // { title: "Blog", href: "/blog" },
       { title: "All updates", href: "/updates" },
       { title: "Changelog", href: "/changelog" },
-      { title: "Newsletter", href: "https://diegohaz.substack.com" },
+      { title: "Newsletter", href: "https://newsletter.ariakit.com" },
     ],
   },
   {
--- a/website/components/global-notification.tsx
+++ b/website/components/global-notification.tsx
@@ -2,7 +2,7 @@ export function GlobalNotification() {
   return null;
   // return (
   //   <a
-  //     href="https://diegohaz.substack.com"
+  //     href="https://newsletter.ariakit.com"
   //     target="_blank"
   //     rel="noopener noreferrer"
   //     className="group flex h-10 items-center gap-2 truncate rounded-lg bg-black/5 px-4 hover:bg-black/10 focus-visible:ariakit-outline-input dark:bg-white/5 dark:hover:bg-white/10"
--- a/website/components/header-nav.tsx
+++ b/website/components/header-nav.tsx
@@ -597,7 +597,7 @@ export function HeaderNav() {
         <HeaderMenuItem href="https://github.com/ariakit/ariakit/discussions">
           Discussions
         </HeaderMenuItem>
-        <HeaderMenuItem href="https://diegohaz.substack.com">
+        <HeaderMenuItem href="https://newsletter.ariakit.com">
           Newsletter
         </HeaderMenuItem>
       </>
--- a/website/components/newsletter-form.tsx
+++ b/website/components/newsletter-form.tsx
@@ -14,7 +14,7 @@ export const NewsletterForm = forwardRef<HTMLFormElement, NewsletterFormProps>(
     return (
       <Role.form
         ref={ref}
-        action="https://diegohaz.substack.com/api/v1/free?email="
+        action="https://newsletter.ariakit.com/api/v1/free?email="
         method="post"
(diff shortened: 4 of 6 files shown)


website/components/page-elements.tsx
```diff
@@ -620,7 +620,7 @@ export function PageA({
       />
     );
   }
-  href = href?.replace(/^https:\/\/(www\.)?ariakit.org/, "");
+  href = href?.replace(/^https:\/\/(www\.)?ariakit.com/, "");
```
diegohaz: This only normalizes absolute links on `ariakit.com`, so historical `https://ariakit.org/...` references now fall through the `href?.startsWith("http")` branch and get rendered as external links. The `/changelog` page is the clearest example: [`website/app/(main)/changelog/page.tsx`](https://github.com/ariakit/ariakit/blob/e004039b36f644ab4474b59274b62e6304e90780/website/app/(main)/changelog/page.tsx) renders `packages/ariakit-react/CHANGELOG.md`, and that file alone still contains 313 absolute `ariakit.org` links. After this change those links lose internal routing and hovercards, and they are forced into a new tab even though they still point at Ariakit pages. We should keep normalizing both hosts here (or normalize by hostname instead of a single regex) until the historical content has been rewritten.

site/astro.config.ts
```diff
@@ -37,7 +37,7 @@ const hasClerk = process.env.PUBLIC_CLERK_PUBLISHABLE_KEY;
 export default defineConfig({
   site:
     process.env.NODE_ENV === "production"
-      ? "https://next.ariakit.org"
+      ? "https://next.ariakit.com"
```
diegohaz, line 40: Switching the production `site` URL to `https://next.ariakit.com` looks premature. As of March 23, 2026, `curl -I -L https://next.ariakit.com` still fails with `Could not resolve host`, while `https://next.ariakit.org` returns `200`. Astro uses this `site` value for absolute URLs and metadata generation, so merging this before the DNS cutover lands will publish dead `next.ariakit.com` URLs. Unless the hostname is already provisioned in the rollout plan, we should keep the old host here (and in the matching URL mapper) until it resolves.

website/components/page-elements.tsx
diegohaz: Fixed in ac7292413 — the regex now strips both `ariakit.com` and `ariakit.org` origins so CHANGELOG links remain internal.

website/components/page-elements.tsx
```diff
@@ -620,7 +620,7 @@ export function PageA({
...
   if (href?.startsWith("http")) {
```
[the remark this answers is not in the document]
diegohaz: Fixed in ac7292413 — the regex now strips both `ariakit.com` and `ariakit.org` origins so legacy CHANGELOG links are treated as internal.

site/src/lib/nextjs.ts
```diff
@@ -49,8 +49,8 @@ type NextjsConventionFile = (typeof NEXTJS_CONVENTION_FILES)[number];
  * - localhost:* → http://localhost:3000
  * - {alias}.ariakit-preview.workers.dev → {alias}.ariakit-nextjs.workers.dev
  * - ariakit-preview.workers.dev → ariakit-nextjs.workers.dev
- * - next.ariakit.org → nextjs.ariakit.org
- * - *.ariakit.org → nextjs.ariakit.org
+ * - next.ariakit.com → nextjs.ariakit.com
+ * - *.ariakit.com → nextjs.ariakit.com
  *
```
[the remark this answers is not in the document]
diegohaz: Fixed in ac7292413 — added the legacy `.org` mapping to the doc comment.

PR #5663, opened 2026-04-03
Replace Safari focus workarounds with tabIndex approach

diegohaz:
## Summary

- On Safari, buttons/checkboxes/radios don't focus on mousedown. Instead of manually focusing in mousedown/mouseup handlers, we now set `tabIndex={0}` on these elements in Safari, which makes the browser focus them natively.
- Removes the `onMouseDown` handler, label click workaround, `isSafariFocusAncestor` marking system, and the dialog disclosure button mousedown workaround.
- Adds sandbox test examples (`focusable-click-focus`, `menu-dialog-close`) to verify the behavior across browsers.

## Test plan

- [x] `pnpm tsc` — 0 errors
- [x] `pnpm test` — 770 unit tests pass
- [x] `pnpm -F site test` — 21 new browser tests pass across Chrome, Firefox, and Safari
- [ ] Verify on real Safari that buttons, checkboxes, and radios receive focus on mousedown
- [ ] Verify dialog-menu toggle works correctly on real Safari

### DIFF
--- a/site/src/sandbox/menu-dialog-close/index.react.tsx
+++ b/site/src/sandbox/menu-dialog-close/index.react.tsx
@@ -1,10 +1,25 @@
 import * as Ariakit from "@ariakit/react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 
 export default function Example() {
   const [open, setOpen] = useState(false);
+
+  // Allow opening the dialog programmatically via F2 (for testing that
+  // programmatic opens don't use stale mousedown data as disclosure).
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "F2") {
+        event.preventDefault();
+        setOpen(true);
+      }
+    };
+    document.addEventListener("keydown", onKeyDown);
+    return () => document.removeEventListener("keydown", onKeyDown);
+  }, []);
+
   return (
     <>
+      <p>Non-interactive text</p>
       <Ariakit.Button onClick={() => setOpen(true)}>Open dialog</Ariakit.Button>
       <Ariakit.Dialog
         open={open}
--- a/site/src/sandbox/menu-dialog-close/test-browser.ts
+++ b/site/src/sandbox/menu-dialog-close/test-browser.ts
@@ -21,4 +21,35 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.menu()).not.toBeVisible();
     await test.expect(menuButton).toBeFocused();
   });
+
+  test("clicking non-interactive element before programmatic open does not hijack disclosure", async ({
+    q,
+    page,
+  }) => {
+    const openButton = q.button("Open dialog");
+
+    // Open and close dialog normally (sets disclosure to openButton)
+    await openButton.click();
+    await test.expect(q.dialog()).toBeVisible();
+    await page.keyboard.press("Escape");
+    await test.expect(q.dialog()).not.toBeVisible();
+    await test.expect(openButton).toBeFocused();
+
+    // Click on non-interactive text (sets mousedown ref on Safari, but the
+    // target is not focusable so isFocusable() guard filters it out)
+    await q.text("Non-interactive text").click();
+
+    // Open dialog programmatically via keyboard shortcut (no mousedown)
+    await page.keyboard.press("F2");
+    await test.expect(q.dialog()).toBeVisible();
+
+    // Close dialog
+    await page.keyboard.press("Escape");
+    await test.expect(q.dialog()).not.toBeVisible();
+
+    // Focus returns to the original disclosure (openButton) from the first
+    // open. The non-interactive click did not hijack the disclosure because
+    // the isFocusable() guard filtered out the non-focusable mousedown target.
+    await test.expect(openButton).toBeFocused();
+  });
 });


packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -123,33 +84,65 @@ function supportsDisabledAttribute(tagName?: string) {
...
+}
+
+function getInputType(props: Record<string, unknown>) {
+  if ("type" in props && typeof props.type === "string") {
+    return props.type;
   }
+  return undefined;
+}
+
+function needsSafariTabIndex(tagName?: string, inputType?: string) {
```
diegohaz, line 105: This narrows the Safari fix compared to the old `onMouseDown` path. Before this PR we used `isButton(element)`, which also covered native button-like inputs such as `type="button"|"submit"|"reset"|...` in addition to checkboxes and radios. Now only `<button>` plus checkbox/radio inputs get the explicit `tabIndex={0}`. I verified with a minimal WebKit check that clicking those input button types still leaves `document.activeElement` on `BODY` unless `tabindex="0"` is present, so `Button`/`Command` rendered as a native input regresses here. Reusing `isButton({ tagName, type: inputType })` would preserve the previous coverage, and it would be worth adding one browser case for a non-checkbox input button type so this does not slip again.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -179,36 +171,6 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     store.setDisclosureElement(activeElement);
```
diegohaz, line 203: After removing the Safari-specific disclosure workaround below, this effect is now the only place that records `disclosureElement`. That is still fine for Ariakit disclosures because `useFocusable` now injects `tabIndex={0}`, but controlled dialogs opened from a plain native `<button>` do not go through that hook. In WebKit, clicking such a button still leaves `document.activeElement === BODY` unless the button has an explicit `tabindex`, so this effect bails out at line 168 and `autoFocusOnHide` loses the opener to restore focus to. The removed effect handled exactly that case. I think we still need a fallback for native non-Ariakit disclosure buttons here, or at least a browser test that proves focus restoration still works when the opener is a plain button.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -123,33 +84,65 @@ function supportsDisabledAttribute(tagName?: string) {
...
+  if (tagName === "button") return true;
+  if (tagName === "input") {
+    return inputType === "checkbox" || inputType === "radio";
+  }
+  return false;
```
diegohaz, line 111: `needsSafariTabIndex` covers `<button>` and `<input type="checkbox|radio">`, but the previous `onMouseDown` handler also covered button-like input types via `isButton()`: `<input type="submit|reset|button|color|file|image">`.

In practice this is a very minor gap since Ariakit's `Button`/`Command` render `<button>` by default, so someone would need `render={<input type="submit" />}` to hit it. But if you want parity, you could import and reuse `isButton` from `@ariakit/core/utils/dom` here, or expand the input type checks.

Low severity — flagging for awareness rather than as a blocker.

.changeset/300-safari-focus-tabindex.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Improved Safari focus behavior for buttons and checkboxes
```
diegohaz: Title says "buttons and checkboxes" but the change also covers radio buttons (the body text already mentions radios correctly).

```suggestion
Improved Safari focus behavior for buttons, checkboxes, and radio buttons
```

packages/ariakit-react-core/src/focusable/focusable.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 37023ad68 - expanded `needsSafariTabIndex` to cover all button-like input types (submit, reset, button, color, file, image) matching the previous `isButton()` coverage. Added a browser test for `input[type=submit]`.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -179,36 +171,6 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
   }, [store, open]);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 845cb00fe - added a Safari-specific mousedown listener that captures the last interaction target and uses it as a fallback disclosure element when `activeElement` is `BODY`.

.changeset/300-safari-focus-tabindex.md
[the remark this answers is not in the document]
diegohaz: Fixed in 5a7e3ed39 - updated title to mention radio buttons.

packages/ariakit-react-core/src/focusable/focusable.tsx
diegohaz: Fixed in 37023ad68 - added all button-like input types from `buttonInputTypes` (button, color, file, image, reset, submit) to `needsSafariTabIndex`, restoring parity with the previous `isButton()` coverage. Also added a browser test for `input[type=submit]` rendered via `<Ariakit.Button render={<input type="submit" />} />`.

packages/ariakit-react-core/src/dialog/dialog.tsx
diegohaz: Fixed in 845cb00fe - added a Safari-specific `mousedown` listener (capture phase) that tracks the last interaction target. When the dialog opens and `activeElement` is `BODY`, the fallback target is used as the disclosure element if it passes `isFocusable()` and is outside the dialog. This preserves focus restoration for non-Ariakit opener buttons.

packages/ariakit-react-core/src/focusable/focusable.tsx
diegohaz: Fixed in 37023ad68 - expanded `needsSafariTabIndex` to include all button-like input types for full parity.

.changeset/300-safari-focus-tabindex.md
diegohaz: Fixed in 5a7e3ed39 - updated title to include radio buttons.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -123,33 +84,75 @@ function supportsDisabledAttribute(tagName?: string) {
...
+  "reset",
+  "submit",
+];
+
+function needsSafariTabIndex(tagName?: string, inputType?: string) {
+  if (tagName === "button") return true;
+  if (tagName === "input" && inputType) {
+    if (inputType === "checkbox" || inputType === "radio") return true;
+    return buttonInputTypes.includes(inputType);
+  }
```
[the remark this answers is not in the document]
diegohaz: Fixed in 85ab83c24 - `getInputType` now lowercases the returned value. Also added support for reading type from render element props.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,47 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+  if (isSafariBrowser) {
+    useEffect(() => {
+      const onMousedown = (event: MouseEvent) => {
+        lastMousedownRef.current = event.target as Element;
+      };
+      document.addEventListener("mousedown", onMousedown, true);
+      return () => {
+        document.removeEventListener("mousedown", onMousedown, true);
+      };
+    }, []);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 0261d6238 - listener now uses `getDocument(dialog)` with `domReady` dependency instead of the global `document`.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,47 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
-    if (activeElement.tagName === "BODY") return;
+    if (activeElement.tagName === "BODY") {
+      // Safari fallback: use the last mousedown target when activeElement is
+      // BODY (happens with native buttons that lack an explicit tabIndex).
+      const fallback = lastMousedownRef.current;
+      if (fallback && isFocusable(fallback)) {
+        if (!dialog || !contains(dialog, fallback)) {
+          store.setDisclosureElement(fallback as HTMLElement);
+        }
+      }
```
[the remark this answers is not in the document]
diegohaz: Fixed in 0261d6238 - the ref is now cleared after being consumed, and the fallback target is checked for `isConnected` before use.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -123,33 +84,75 @@ function supportsDisabledAttribute(tagName?: string) {
...
+  supportsDisabled: boolean;
+  tagName?: string;
+  inputType?: string;
+  tabIndexProp?: number;
+}
+
+function getInputType(props: Record<string, unknown>) {
+  if ("type" in props && typeof props.type === "string") {
+    return props.type;
   }
```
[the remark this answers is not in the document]
diegohaz: Resolved in 85ab83c24 - `getInputType` now reads from `props.render.props.type` when the top-level `props.type` is absent, and lowercases the value.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,47 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
     if (!activeElement) return;
-    if (activeElement.tagName === "BODY") return;
+    if (activeElement.tagName === "BODY") {
+      // Safari fallback: use the last mousedown target when activeElement is
+      // BODY (happens with native buttons that lack an explicit tabIndex).
+      const fallback = lastMousedownRef.current;
+      if (fallback && isFocusable(fallback)) {
+        if (!dialog || !contains(dialog, fallback)) {
+          store.setDisclosureElement(fallback as HTMLElement);
+        }
```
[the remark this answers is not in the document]
diegohaz: Resolved in 0261d6238 - the ref is cleared after consumption (`lastMousedownRef.current = null`) and the fallback target is checked for `isConnected` before use.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,52 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
     const dialog = ref.current;
     const activeElement = getActiveElement(dialog, true);
     if (!activeElement) return;
-    if (activeElement.tagName === "BODY") return;
+    if (activeElement.tagName === "BODY") {
+      // Safari fallback: use the last mousedown target when activeElement is
+      // BODY (happens with native buttons that lack an explicit tabIndex).
+      const fallback = lastMousedownRef.current;
+      lastMousedownRef.current = null;
+      if (fallback && fallback.isConnected && isFocusable(fallback)) {
```
diegohaz: Shouldn't it be `fallback?.isConnected`?

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,52 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+        if (!dialog || !contains(dialog, fallback)) {
+          store.setDisclosureElement(fallback as HTMLElement);
+        }
+      }
```
diegohaz: Can we have early returns here? I'm not sure if it would be cleaner.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -123,33 +84,84 @@ function supportsDisabledAttribute(tagName?: string) {
...
+
+function getInputType(props: Record<string, unknown>) {
+  if ("type" in props && typeof props.type === "string") {
+    return props.type.toLowerCase();
+  }
+  // When using render={<input type="submit" />}, the type prop is on the
+  // render element, not on the hook's props.
+  const render = props.render;
+  if (render && typeof render === "object" && "props" in render) {
+    const renderProps = render.props as Record<string, unknown> | undefined;
```
diegohaz: This seems fragile because the render component could be abstracted into a custom component, so the `type` prop might not be exposed directly on `render.props`. I think we should add a Safari-only `useEffect` that checks whether the element is a button, checkbox, or radio, and then use state. We should have tests for this as well.

site/src/sandbox/focusable-click-focus/test-browser.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  test("radio receives focus on click", async ({ q }) => {
+    const radio = q.radio("Radio A");
+    await radio.click();
+    await test.expect(radio).toBeFocused();
+  });
+
+  test("checkbox receives focus on label click", async ({ page, q }) => {
+    const checkbox = q.checkbox("Checkbox");
+    await page.getByText("Checkbox").click();
```
diegohaz: It should be `q.text("Checkbox").click()`. Make sure to update the skills to reflect this.

site/src/sandbox/focusable-click-focus/test-browser.ts
```diff
@@ -0,0 +1,39 @@
...
+    await test.expect(checkbox).toBeFocused();
+  });
+
+  test("radio receives focus on label click", async ({ page, q }) => {
+    const radio = q.radio("Radio A");
+    await page.getByText("Radio A").click();
```
diegohaz: Same here

packages/ariakit-react-core/src/dialog/dialog.tsx
diegohaz: Fixed in 7d8d5ca0a - now uses `fallback?.isConnected` with optional chaining.

packages/ariakit-react-core/src/dialog/dialog.tsx
diegohaz: Fixed in 7d8d5ca0a - refactored to early returns: `if (!fallback?.isConnected) return; if (!isFocusable(fallback)) return; if (dialog && contains(dialog, fallback)) return;`

packages/ariakit-react-core/src/focusable/focusable.tsx
diegohaz: Fixed in 7d8d5ca0a - replaced `getInputType` (which inspected render.props) with a Safari-only `useEffect` that reads `tagName` and `type` from the actual DOM element via `ref.current`. This works correctly with `render` props, custom components, and any abstraction layer.

site/src/sandbox/focusable-click-focus/test-browser.ts
diegohaz: Fixed in 7d8d5ca0a - changed to `q.text("Checkbox").click()`.

site/src/sandbox/focusable-click-focus/test-browser.ts
diegohaz: Fixed in 7d8d5ca0a - changed to `q.text("Radio A").click()`.

site/src/sandbox/focusable-click-focus/index.react.tsx
```diff
@@ -0,0 +1,27 @@
+import * as Ariakit from "@ariakit/react";
```
diegohaz, line 1: You should add a case here that abstracts the `<input type>` into a custom component so we can check whether our `useEffect` is actually working.

packages/ariakit-react-core/src/dialog/dialog.tsx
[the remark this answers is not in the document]
diegohaz: This is not an issue because `lastMousedownRef` is initialized as `null`, and when consumed in the `BODY` fallback path, it is immediately cleared (`lastMousedownRef.current = null` at line 194). If no mousedown occurred, `fallback` is `null`, and `null?.isConnected` is `undefined` (falsy), so the early return at line 195 fires. If a mousedown did occur and was consumed, the ref is cleared, preventing reuse. The remaining edge case (programmatic open after an unrelated mousedown) is extremely unlikely and benign — the fallback must still pass `isFocusable()` and the `contains(dialog, fallback)` check.

site/src/sandbox/focusable-click-focus/index.react.tsx
diegohaz: Fixed in b60150e1f - added a `CustomCheckbox` component that wraps `<input type="checkbox">` and is used via `render={<CustomCheckbox />}`. Added two browser tests (click and label click) to verify the Safari `useEffect`-based detection works with custom component abstractions.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -433,6 +354,22 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
+    if (isSafariBrowser) {
+      useEffect(() => {
+        if (!focusable) return;
+        const element = ref.current;
+        if (!element) return;
+        const tag = element.tagName.toLowerCase();
+        const type = (element as HTMLInputElement).type;
+        setSafariTabIndex(needsSafariTabIndex(tag, type));
+      }, [focusable]);
+    }
```
diegohaz, line 371: I rechecked this after the follow-up commits and it still holds. The new Safari tests pass, but they only exercise the hydrated state. After `pnpm build-site-lite`, the built preview at `site/dist/react/previews/focusable-click-focus/index.html` still renders the affected controls as plain `<button>` / `<input>` elements with no `tabindex` in the SSR HTML, because `safariTabIndex` starts as `false` and only flips inside this passive `useEffect`.

That matters here because the preview hydrates with `client:load`, so Safari users can click these controls before the effect runs and still hit the original `document.activeElement === BODY` behavior on the first interaction. I think this needs to move back to a layout-timed path (`useSafeLayoutEffect`, or another pre-paint/ref-driven derivation) so the explicit `tabIndex={0}` is present before first paint instead of after hydration.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -166,49 +158,51 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
     if (!open) return;
     const dialog = ref.current;
     const activeElement = getActiveElement(dialog, true);
     if (!activeElement) return;
-    if (activeElement.tagName === "BODY") return;
+    if (activeElement.tagName === "BODY") {
+      // Safari fallback: use the last mousedown target when activeElement is
+      // BODY (happens with native buttons that lack an explicit tabIndex).
+      const fallback = lastMousedownRef.current;
+      lastMousedownRef.current = null;
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because the ref is already guarded against staleness: it starts as `null`, is cleared immediately after consumption (line 194), and the fallback target must pass `isConnected`, `isFocusable()`, and `contains()` checks. If no recent mousedown occurred, `fallback` is `null` and the early return at line 195 fires. See the [earlier reply](https://github.com/ariakit/ariakit/pull/5663#discussion_r3035902932) to the identical concern from copilot.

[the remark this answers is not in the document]
diegohaz: Added a test in f60b784e9 proving this is not an issue. The test clicks a non-interactive element (setting the mousedown ref on Safari), then opens the dialog programmatically via F2. On close, focus correctly returns to the original disclosure because the `isFocusable()` guard filters out the non-focusable mousedown target. Combined with the ref being initialized as `null`, cleared after consumption, and the `isConnected`/`contains()` guards, stale mousedown data cannot hijack the disclosure.

packages/ariakit-react-core/src/focusable/focusable.tsx
```diff
@@ -433,6 +354,22 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
+
+    if (isSafariBrowser) {
+      useEffect(() => {
+        if (!focusable) return;
+        const element = ref.current;
+        if (!element) return;
+        const tag = element.tagName.toLowerCase();
+        const type = (element as HTMLInputElement).type;
+        setSafariTabIndex(needsSafariTabIndex(tag, type));
+      }, [focusable]);
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because React unmounts and remounts the component when the element type changes (e.g., switching from `<div>` to `<input>`), which re-runs the effect. The `render` prop determines the element type at mount time and is not expected to change the tag name dynamically. Changing an `<input>`'s `type` attribute after mount is not a supported pattern in Ariakit. The `[focusable]` dependency is sufficient for the realistic usage of this hook.

PR #5727, opened 2026-04-10
Disable false-positive oxlint rules and fix type conversion warnings

diegohaz:
## Motivation

Running `pnpm lint` produces 218 warnings across three rules. Most are false positives that add noise without catching real issues:

- `consistent-return` (159 warnings): flags idiomatic guard-clause patterns (`if (!store) return;`) that TypeScript already validates through overloaded signatures and type inference.
- `no-unnecessary-type-arguments` (38 warnings): incorrectly identifies extended item types (e.g., `ComboboxStoreItem`, `SelectStoreItem`) as default type parameters, causing type errors when removed.
- `no-unnecessary-type-conversion` (20 warnings): flags genuinely redundant `!!`, `String()`, and `Number()` conversions on values that already have the target type.

## Solution

- Disable `consistent-return` and `no-unnecessary-type-arguments` in `oxlint.config.ts` since they produce false positives in this codebase.
- Remove the 20 unnecessary type conversions flagged by `no-unnecessary-type-conversion`.


packages/ariakit-react-core/src/combobox/combobox.tsx
```diff
@@ -532,7 +532,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
 
     const onMouseDownProp = props.onMouseDown;
     const blurActiveItemOnClickProp = useBooleanEvent(
-      blurActiveItemOnClick ?? (() => !!store?.getState().includesBaseElement),
+      blurActiveItemOnClick ?? (() => store?.getState().includesBaseElement),
```
diegohaz: What if `store` is `undefined`?

packages/ariakit-react-core/src/select/select-item.tsx
```diff
@@ -187,7 +187,7 @@ export const useSelectItem = createHook<TagName, SelectItemOptions>(
       focusOnHover(event) {
         if (!focusOnHoverProp(event)) return false;
         const state = store?.getState();
-        return !!state?.open;
+        return state?.open;
```
diegohaz: What if `state` is `undefined`?

site/src/lib/jsdoc-loader.ts
```diff
@@ -721,7 +721,7 @@ function getDescription(node: Node) {
   // Get description from first JSDoc block
   const firstJsDoc = jsDocs[0];
   const description = firstJsDoc?.getDescription();
-  return description ? String(description).trim() : "";
+  return description ? description.trim() : "";
```
diegohaz, line 724: What if `description` is `undefined`?

packages/ariakit-react-core/src/combobox/combobox.tsx
diegohaz: Fixed in 1d47287d4 — restored `!!` since `store?.getState().includesBaseElement` returns `boolean | undefined` when `store` is `undefined`. Added an inline `oxlint-disable-next-line` because the linter incorrectly flags this as redundant.

packages/ariakit-react-core/src/select/select-item.tsx
diegohaz: Fixed in 1d47287d4 — restored `!!` since `state?.open` returns `boolean | undefined` when `state` is `undefined`. Added an inline `oxlint-disable-next-line` because the linter incorrectly flags this as redundant.

site/src/lib/jsdoc-loader.ts
diegohaz: The ternary `description ? description.trim() : ""` already handles `undefined` — if `description` is `undefined`, the falsy branch returns `""`. The removed `String()` was inside the truthy branch where TypeScript narrows `description` to `string`, so it was genuinely redundant.

site/src/lib/source-plugin.ts
```diff
@@ -239,7 +239,7 @@ function collectDependencyFromResolved(
     }
   }
   const typesPackageId = resolved.resolvedModule?.packageId?.name;
-  const hasTypes = typesPackageId && !!typesPackageId.startsWith("@types/");
+  const hasTypes = typesPackageId && typesPackageId.startsWith("@types/");
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because the removed `!!` was on `.startsWith()` which already returns `boolean`. The `&&` short-circuit behavior (yielding a falsy non-boolean when `typesPackageId` is empty/undefined) was the same before and after this change. The downstream usage `if (hasTypes && resolvedTypesPath)` is a truthiness check that works correctly with any falsy value.

diegohaz:
**Fixed**
- 1d47287d4 ([combobox.tsx](https://github.com/ariakit/ariakit/pull/5727#discussion_r3064173812)): Restored `!!` — `store?.getState().includesBaseElement` returns `boolean | undefined` when `store` is `undefined`.
- 1d47287d4 ([select-item.tsx](https://github.com/ariakit/ariakit/pull/5727#discussion_r3064174300)): Restored `!!` — `state?.open` returns `boolean | undefined` when `state` is `undefined`.

**Already resolved**
- [jsdoc-loader.ts](https://github.com/ariakit/ariakit/pull/5727#discussion_r3064174494): The ternary `description ?` already guards against `undefined` before `.trim()` is called.

**Invalid**
- [source-plugin.ts](https://github.com/ariakit/ariakit/pull/5727#discussion_r3064174770) (Copilot): The removed `!!` was on `.startsWith()` which already returns `boolean`; the `&&` short-circuit behavior is unchanged.

packages/ariakit-react-core/src/select/select-item.tsx
```diff
@@ -187,6 +187,7 @@ export const useSelectItem = createHook<TagName, SelectItemOptions>(
       focusOnHover(event) {
         if (!focusOnHoverProp(event)) return false;
         const state = store?.getState();
+        // oxlint-disable-next-line no-unnecessary-type-conversion
         return !!state?.open;
```
diegohaz: I think we should actually remove `?` from the `store`, then just return `state.open`.

packages/ariakit-react-core/src/combobox/combobox.tsx
```diff
@@ -532,6 +532,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
 
     const onMouseDownProp = props.onMouseDown;
     const blurActiveItemOnClickProp = useBooleanEvent(
+      // oxlint-disable-next-line no-unnecessary-type-conversion
       blurActiveItemOnClick ?? (() => !!store?.getState().includesBaseElement),
```
diegohaz: Same here,
```suggestion
blurActiveItemOnClick ?? (() => store.getState().includesBaseElement),
```

packages/ariakit-react-core/src/select/select-item.tsx
diegohaz: Fixed in 4c0349afe — removed optional chaining on `store` and returning `state.open` directly.

packages/ariakit-react-core/src/combobox/combobox.tsx
diegohaz: Fixed in 4c0349afe — applied the suggestion, using `store.getState().includesBaseElement` without optional chaining.

PR #5736, opened 2026-04-11
Fix mergeProps falsy value handling for className and event handlers

### LINKED ISSUE #5224: `mergeProps` incorrectly handles falsy values for className and event handlers
### Current behavior

1. Falsy className values are converted to strings and concatenated, resulting in invalid class names like "base null" or "base undefined".
2. Falsy event handler values overwrite the base handler, effectively removing it.

### Bug 1: className with falsy override

```tsx
import { mergeProps } from "@ariakit/react-core/utils/misc";

// Override className is null
mergeProps({ className: "base" }, { className: null });
// Expected: { className: "base" }
// Actual:   { className: "base null" }

// Override className is undefined
mergeProps({ className: "base" }, { className: undefined });
// Expected: { className: "base" }
// Actual:   { className: "base undefined" }

// Override className is empty string
mergeProps({ className: "base" }, { className: "" });
// Expected: { className: "base" }
// Actual:   { className: "base " } // trailing space
```

### Bug 2: Event handler with null/undefined override

```tsx
import { mergeProps } from "@ariakit/react-core/utils/misc";

const baseHandler = () => console.log("base");

// Override handler is null
mergeProps({ onClick: baseHandler }, { onClick: null });
// Expected: { onClick: baseHandler }
// Actual:   { onClick: null }

// Override handler is undefined
mergeProps({ onClick: baseHandler }, { onClick: undefined });
// Expected: { onClick: baseHandler }
// Actual:   { onClick: undefined }
```

### Expected behavior

1. When overrides.className is falsy (null, undefined, or ""), the base className should be preserved without modification.
2. When an event handler override is null or undefined, the base event handler should be preserved.

### Workaround

https://github.com/ariakit/ariakit/issues/5224#issuecomment-4196388229

### Possible solutions

```ts
if (key === "className") {
  const prop = "className";
  props[prop] = base[prop] && overrides[prop]
    ? `${base[prop]} ${overrides[prop]}`
    : overrides[prop] || base[prop];
  continue;
}
```

For event handlers, skip assignment when override is nullish:

```ts
const overrideValue = overrides[key];

if (key.startsWith("on")) {
  if (typeof overrideValue !== "function") {
    continue; // Skip nullish event handler overrides
  }
  // ... rest of event handler merging logic
}
```

diegohaz:
## Motivation

`mergeProps` incorrectly handles falsy values passed via the `render` prop:

1. Falsy `className` overrides (`null`, `undefined`, `""`) are concatenated as strings, producing invalid class names like `"base undefined"`.
2. Falsy event handler overrides (`null`, `undefined`) silently overwrite base handlers, breaking internal behavior like `DialogDisclosure`'s click handler.

Fixes #5224

## Solution

- For `className`: check that both base and override are truthy before concatenating. When only one is truthy, use that value.
- For event handlers: when an override key starts with `on` but the value is not a function, skip it entirely so the base handler is preserved.
- Applied the same fix to the site's example `mergeProps` utility.

## Workaround

Avoid passing `null` or `undefined` as `className` or event handler values to components via the `render` prop. Omit the key entirely instead:

```tsx
// Instead of:
<DialogDisclosure
  render={
    // TODO: Remove workaround once
    // https://github.com/ariakit/ariakit/issues/5224 is fixed.
    <Button onClick={condition ? handler : undefined}>Open</Button>
  }
/>

// Do:
<DialogDisclosure
  render={
    <Button {...(condition ? { onClick: handler } : {})}>Open</Button>
  }
/>
```

## Test plan

- [x] Failing repro tests for both className and event handler bugs
- [x] Workaround applied in sandbox passes tests
- [x] Library fix passes tests
- [x] All 57 existing browser tests still pass

### DIFF
--- a/.changeset/300-merge-props-falsy-values.md
+++ b/.changeset/300-merge-props-falsy-values.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-core": patch
+---
+
+Fixed [`render`](https://ariakit.com/guide/composition) prop merging when the rendered element passes falsy `className` or event handler values such as `undefined` or `null`.
--- a/.changeset/300-merge-props-falsy-values.md
+++ b/.changeset/300-merge-props-falsy-values.md
@@ -1,5 +1,6 @@
 ---
 "@ariakit/react-core": patch
+"@ariakit/react": patch
 ---
 
 Fixed [`render`](https://ariakit.com/guide/composition) prop merging when the rendered element passes falsy `className` or event handler values such as `undefined` or `null`.
--- a/packages/ariakit-react-core/src/utils/misc.ts
+++ b/packages/ariakit-react-core/src/utils/misc.ts
@@ -58,9 +58,13 @@ export function mergeProps<T extends HTMLAttributes<any>>(
 
     if (key === "className") {
       const prop = "className";
-      props[prop] = base[prop]
-        ? `${base[prop]} ${overrides[prop]}`
-        : overrides[prop];
+      const baseClass = base[prop];
+      const overrideClass = overrides[prop];
+      if (baseClass && overrideClass) {
+        props[prop] = `${baseClass} ${overrideClass}`;
+      } else {
+        props[prop] = overrideClass || baseClass;
+      }
       continue;
     }
 
@@ -74,7 +78,11 @@ export function mergeProps<T extends HTMLAttributes<any>>(
 
     const overrideValue = overrides[key];
 
-    if (typeof overrideValue === "function" && key.startsWith("on")) {
+    if (key.startsWith("on")) {
+      if (typeof overrideValue !== "function") {
+        // Skip nullish event handler overrides to preserve base handlers.
+        continue;
+      }
       const baseValue = base[key];
       if (typeof baseValue === "function") {
         type EventKey = Extract<keyof HTMLAttributes<any>, `on${string}`>;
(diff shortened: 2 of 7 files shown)


.changeset/300-merge-props-falsy-values.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/react-core": patch
```
diegohaz, line 2: Need to include `@ariakit/react` too.

packages/ariakit-react-core/src/utils/misc.ts
```diff
@@ -74,7 +78,11 @@ export function mergeProps<T extends HTMLAttributes<any>>(
 
     const overrideValue = overrides[key];
 
-    if (typeof overrideValue === "function" && key.startsWith("on")) {
+    if (key.startsWith("on")) {
+      if (typeof overrideValue !== "function") {
+        // Skip nullish event handler overrides to preserve base handlers.
```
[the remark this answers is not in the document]
diegohaz: Fixed in e7b18c6d1 - Updated comment to say "non-function" instead of "nullish" to match actual behavior.

.changeset/300-merge-props-falsy-values.md
diegohaz: Fixed in 9777bfc99 - Added @ariakit/react to the changeset.

diegohaz:
**Fixed**

- e7b18c6d1 ([reply](https://github.com/ariakit/ariakit/pull/5736#discussion_r3067895067)): Updated comment to say "non-function" instead of "nullish" to match actual behavior
- 9777bfc99 ([reply](https://github.com/ariakit/ariakit/pull/5736#discussion_r3067895152)): Added @ariakit/react to the changeset

PR #5738, opened 2026-04-11
Fix Math.random() being called when id prop is provided

### LINKED ISSUE #5147: Math.random() execution in useTabStore breaks Next.js cache component
### Current behavior

When using Ariakit's useTabStore or other composite stores with Next.js's cache component,  Next.js throws the following error:
used `Math.random()` inside a Client Component without a Suspense boundary above it.

This happens because Math.random() is called unconditionally in createCompositeStore, even when an explicit id prop is provided.

In packages/ariakit-core/src/composite/composite-store.ts (lines 165-169):

```typescript
id: defaultValue(
  props.id,
  syncState?.id,
  `id-${Math.random().toString(36).slice(2, 8)}`,
),
```

Due to JavaScript's parameter evaluation rules, Math.random() is executed before defaultValue is called, even when props.id is provided and will be used as the result.

**Why This Breaks Next.js**

Next.js's cache function performs Server-Side Rendering (SSR) simulation. When it detects random value generation during prerendering, it cannot include the component in the prerendered HTML because the random value would be fixed for all users. Next.js expects to use Suspense boundaries for such components, but this creates unnecessary complexity for a library that should be deterministic when given explicit props.

Please refer to the issue description above for detailed. Sorry for any inconvenience

https://codesandbox.io/p/github/septem1997/jsusjivg-exi9osyq/main?file=%2Fapp%2Ftab%2Findex.tsx%3A13%2C25

### Expected behavior

When an explicit id prop is provided to useTabStore or similar functions, Math.random() should not be executed at all, maintaining predictability and compatibility with Next.js's caching features.

### Workaround

https://github.com/ariakit/ariakit/issues/5147#issuecomment-4229426347

### Possible solutions

**Proposed Solution**

```typescript
// In defaultValue function
export function defaultValue(...values: unknown[]) {
  for (const value of values) {
    if (value !== undefined) {
      // Support lazy evaluation for functions
      if (typeof value === "function") {
        return value();
      }
      return value;
    }
  }
  return undefined;
}

// In composite-store.ts
id: defaultValue(
  props.id,
  syncState?.id,
  () => `id-${Math.random().toString(36).slice(2, 8)}`,
),
```

diegohaz:
## Summary

Fixes #5147

**Problem:** In `composite-store.ts`, the `id` default uses `defaultValue(props.id, syncState?.id, \`id-${Math.random()...}\`)`. JavaScript eagerly evaluates all arguments, so `Math.random()` runs unconditionally — even when `props.id` is already provided. This breaks Next.js builds with `cacheComponents` enabled because `Math.random()` is non-deterministic.

**Solution:** Use nullish coalescing for lazy evaluation:
```ts
id: defaultValue(props.id, syncState?.id) ?? `id-${Math.random()...}`
```

## Workaround

Wrap components using composite stores in a `Suspense` boundary:

```tsx
import { Suspense } from "react";

// TODO: Remove after ariakit update
export default function Page() {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      <TabContent />
    </Suspense>
  );
}
```

## Test plan

- [x] Next.js build fails with `cacheComponents: true` before the fix
- [x] Next.js build passes after the fix

### DIFF
--- a/.changeset/fix-composite-store-random-id.md
+++ b/.changeset/fix-composite-store-random-id.md
@@ -1,5 +1,6 @@
 ---
 "@ariakit/core": patch
+"@ariakit/react-core": patch
 "@ariakit/react": patch
 ---
 
--- a/.changeset/fix-composite-store-random-id.md
+++ b/.changeset/fix-composite-store-random-id.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/core": patch
+"@ariakit/react": patch
+---
+
+Fixed `Math.random()` being called unconditionally when creating composite stores ([`useTabStore`](https://ariakit.com/reference/use-tab-store), [`useComboboxStore`](https://ariakit.com/reference/use-combobox-store), [`useSelectStore`](https://ariakit.com/reference/use-select-store), etc.), even when an explicit `id` prop was provided. This was causing Next.js build errors with `cacheComponents` enabled.
--- a/nextjs/app/tab-5147/page.tsx
+++ b/nextjs/app/tab-5147/page.tsx
@@ -1,9 +1,8 @@
 "use client";
 
 import * as Ariakit from "@ariakit/react";
-import { Suspense } from "react";
 
-function TabContent() {
+export default function TabPage() {
   const tab = Ariakit.useTabStore({ id: "my-tabs" });
 
   return (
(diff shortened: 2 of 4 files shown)


.changeset/fix-composite-store-random-id.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/core": patch
```
diegohaz, line 2: Missing @ariakit/react-core

diegohaz: Fixed in 0b83eaa9b - Added missing @ariakit/react-core to the changeset.

diegohaz:
**Fixed**

- 0b83eaa9b (https://github.com/ariakit/ariakit/pull/5738#discussion_r3068025281): Added missing @ariakit/react-core to the changeset for the Math.random() composite store fix, as the affected stores are primarily exported from that package.

.changeset/fix-composite-store-random-id.md
diegohaz: Resolved in 0b83eaa9b - Added missing `@ariakit/react-core` to the changeset frontmatter.

diegohaz:
**Already resolved**

- 0b83eaa9b ([reply](https://github.com/ariakit/ariakit/pull/5738#discussion_r3068026013)): Added missing `@ariakit/react-core` to changeset frontmatter.

PR #5741, opened 2026-04-12
Prevent combobox Enter from submitting form when popover has no items

### LINKED ISSUE #3323: Unexpected Form Submission on Combobox having no results
### Current behavior

When using the ComboBox and performing a search that results in an empty list, pressing the 'Enter' key triggers the form submission

1. Open sandbox: https://stackblitz.com/edit/stackblitz-starters-a6j6lm?file=src%2FApp.tsx
2. Open the select combobox
3. Type 'asdasd'
4. Press 'Enter'

### Expected behavior

The form submission should not be triggered as the combobox field is not a part of the form

### Workaround

add form attribute to the combobox

### Possible solutions

_No response_

diegohaz:
Fixes #3323

## Motivation

When a combobox is inside a form and the popover is open with no matching items, pressing Enter triggers form submission. This is because the combobox input is a native `<input>` element inside a `<form>`, and without any active item to intercept the Enter key, the browser's default form submission behavior takes over.

The issue is exacerbated when consumers use `startTransition` for filtering — the store's `activeId` can briefly reference a stale item that has already been removed from the DOM, creating a window where Enter falls through to the browser's default behavior.

## Solution

When the combobox popover is open, always prevent the Enter key's default behavior. If there's an active item, the Composite's keyboard event proxy dispatches the event to that item, which handles selection via its own click handler. If there's no active item (e.g., all items are filtered out or `activeId` is stale), Enter becomes a no-op instead of submitting the form.

## Workaround

When the combobox popover is open but all items are filtered out, pressing Enter on the combobox input falls through to the browser's default behavior, which submits the enclosing form. To prevent this, add an `onKeyDown` handler that calls `preventDefault` when there are no matching items:

```tsx

const matches = useMemo(() => matchSorter(list, searchValue), [searchValue]);

<Ariakit.Combobox
  // TODO: Remove when https://github.com/ariakit/ariakit/issues/3323 is fixed
  onKeyDown={(event) => {
    if (event.key === "Enter" && !matches.length) {
      event.preventDefault();
    }
  }}
/>
```

### DIFF
--- a/.changeset/300-combobox-form-enter.md
+++ b/.changeset/300-combobox-form-enter.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Combobox`](https://ariakit.com/reference/combobox) pressing Enter submitting a parent form when the popover is open but has no matching items.
+Fixed [`Combobox`](https://ariakit.com/reference/combobox) pressing Enter from submitting a parent form when the popover is open but has no matching items.
--- a/.changeset/300-combobox-form-enter.md
+++ b/.changeset/300-combobox-form-enter.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Combobox`](https://ariakit.com/reference/combobox) pressing Enter submitting a parent form when the popover is open but has no matching items.
--- a/packages/ariakit-react-core/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-core/src/combobox/combobox.tsx
@@ -564,24 +564,24 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
         canAutoSelectRef.current = false;
       }
       if (event.defaultPrevented) return;
+      if (!store) return;
+      const { open } = store.getState();
+      // When the popover is open, prevent Enter (with or without modifiers)
+      // from triggering the default behavior (submitting a parent form). If
+      // there's an active item, the keyboard event proxy on Composite will
+      // dispatch Enter to the item, which handles selection. If there's no
+      // active item (e.g., all items are filtered out, or activeId is stale
+      // after a React transition), Enter should be a no-op rather than
+      // submitting a form.
+      if (open && event.key === "Enter") {
+        event.preventDefault();
+        return;
+      }
       if (event.ctrlKey) return;
       if (event.altKey) return;
       if (event.shiftKey) return;
       if (event.metaKey) return;
-      if (!store) return;
-      const { open } = store.getState();
-      if (open) {
-        // When the popover is open, prevent Enter from triggering the default
-        // behavior (submitting a parent form). If there's an active item, the
-        // keyboard event proxy on Composite will dispatch Enter to the item,
-        // which handles selection. If there's no active item (e.g., all items
-        // are filtered out, or activeId is stale after a React transition),
-        // Enter should be a no-op rather than submitting a form.
-        if (event.key === "Enter") {
-          event.preventDefault();
-        }
-        return;
-      }
+      if (open) return;
       // Up and Down arrow keys should open the combobox popover.
       if (event.key === "ArrowUp" || event.key === "ArrowDown") {
         if (showOnKeyPressProp(event)) {
--- a/packages/ariakit-react-core/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-core/src/combobox/combobox.tsx
@@ -570,7 +570,18 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       if (event.metaKey) return;
       if (!store) return;
       const { open } = store.getState();
-      if (open) return;
+      if (open) {
+        // When the popover is open, prevent Enter from triggering the default
+        // behavior (submitting a parent form). If there's an active item, the
+        // keyboard event proxy on Composite will dispatch Enter to the item,
+        // which handles selection. If there's no active item (e.g., all items
+        // are filtered out, or activeId is stale after a React transition),
+        // Enter should be a no-op rather than submitting a form.
+        if (event.key === "Enter") {
+          event.preventDefault();
+        }
+        return;
+      }
       // Up and Down arrow keys should open the combobox popover.
       if (event.key === "ArrowUp" || event.key === "ArrowDown") {
         if (showOnKeyPressProp(event)) {
--- a/site/src/sandbox/select-combobox-form-3323/test-browser.ts
+++ b/site/src/sandbox/select-combobox-form-3323/test-browser.ts
@@ -9,7 +9,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await page.keyboard.type("asdasd");
     await test.expect(q.option("Apple")).toBeHidden();
     await page.keyboard.press("Enter");
-    await test.expect(page.getByText("Form submitted")).toBeHidden();
+    await test.expect(q.text("Form submitted")).toBeHidden();
   });
 
   test("pressing Shift+Enter with no matching items does not submit the form", async ({
@@ -20,7 +20,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await page.keyboard.type("asdasd");
     await test.expect(q.option("Apple")).toBeHidden();
     await page.keyboard.press("Shift+Enter");
-    await test.expect(page.getByText("Form submitted")).toBeHidden();
+    await test.expect(q.text("Form submitted")).toBeHidden();
   });
 
   test("pressing Enter with a matching item selects it without submitting the form", async ({
@@ -31,6 +31,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.option("Apple")).toBeVisible();
     await page.keyboard.press("Enter");
     await test.expect(q.option("Apple")).toBeHidden();
-    await test.expect(page.getByText("Form submitted")).toBeHidden();
+    await test.expect(q.text("Form submitted")).toBeHidden();
   });
 });
--- a/site/src/sandbox/select-combobox-form-3323/test-browser.ts
+++ b/site/src/sandbox/select-combobox-form-3323/test-browser.ts
@@ -1,0 +1,14 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("pressing Enter with no matching items does not submit the form", async ({
+    page,
+    q,
+  }) => {
+    await q.combobox("Favorite fruit").click();
+    await page.keyboard.type("asdasd");
+    await test.expect(q.option("Apple")).toBeHidden();
+    await page.keyboard.press("Enter");
+    await test.expect(page.getByText("Form submitted")).toBeHidden();
+  });
+});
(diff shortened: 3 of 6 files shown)


packages/ariakit-react-core/src/combobox/combobox.tsx
```diff
@@ -570,7 +570,18 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+        // When the popover is open, prevent Enter from triggering the default
+        // behavior (submitting a parent form). If there's an active item, the
+        // keyboard event proxy on Composite will dispatch Enter to the item,
+        // which handles selection. If there's no active item (e.g., all items
+        // are filtered out, or activeId is stale after a React transition),
+        // Enter should be a no-op rather than submitting a form.
+        if (event.key === "Enter") {
+          event.preventDefault();
+        }
+        return;
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because modifier+Enter combinations (Ctrl+Enter, Shift+Enter, etc.) do not trigger HTML form submission. Only plain Enter on an input does. The modifier early-returns are pre-existing and intentional across the codebase.

.changeset/300-combobox-form-enter.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Combobox`](https://ariakit.com/reference/combobox) pressing Enter submitting a parent form when the popover is open but has no matching items.
```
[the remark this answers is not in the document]
diegohaz: Fixed in 98b440a26 - improved changeset grammar.

diegohaz:
Addressed review comments:

**Fixed**
- 98b440a26 ([reply](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068911332)): Changeset grammar — added "from" for clarity.

**Invalid**
- ([reply](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068911272)): Modifier+Enter bypassing the fix — modifier+Enter does not trigger HTML form submission, and the modifier early-returns are pre-existing intentional design.

diegohaz (commented):
## Review summary

**Type**: Initial review
**Baseline**: merge base `2d47020b7` (origin/main) to head `98b440a26`
**Scope**: All 6 changed files — core fix in `combobox.tsx`, sandbox repro, browser test, changeset

### What I checked

- **Core fix** (`combobox.tsx`): Traced the full event flow from the combobox's `onKeyDown` through `useComposite`'s `onKeyDownCapture` (which uses `useKeyboardEventProxy`). When the popover is open and no active item exists, `useKeyboardEventProxy` returns early at `if (!activeElement) return` without preventing default, letting the browser's native form submission happen. The PR's `preventDefault()` on Enter fills this gap correctly. When an active item IS present, the keyboard event proxy already dispatches Enter to the item and prevents default in the capture phase — the additional `preventDefault` in the bubble phase is redundant but harmless.

- **Modifier key concern** (raised by Copilot): Correctly dismissed by @diegohaz. Modifier+Enter (Ctrl+Enter, Shift+Enter, etc.) does not trigger HTML form submission. The existing modifier guards are intentional and pre-existing.

- **IME composition concern** (raised by CodeRabbit): Not a regression from this PR. The `if (open) return;` was already executing during composition before this change. The only new behavior is calling `preventDefault()` before the return. In practice, most modern browsers emit `key: "Process"` during active IME composition rather than `key: "Enter"`, so the guard would not match. Safari may differ, but this is pre-existing behavior, not introduced by this PR.

- **Search combobox scenario**: When a standalone combobox is in a form and the popover is open showing "no results," Enter will now be a no-op instead of submitting. This is a deliberate design choice consistent with the principle that Enter interacts with the popover when it's open. Users who want Enter-to-submit can close the popover first or add their own `onKeyDown` handler.

- **Sandbox and test**: The sandbox (`select-combobox-form-3323`) follows the naming convention for issue repro sandboxes. The browser test uses `withFramework` and verifies the exact scenario from issue #3323 — typing a non-matching query, pressing Enter, and asserting the form was not submitted. Test pattern matches existing conventions.

- **Changeset**: Properly scoped as a patch for `@ariakit/react-core` and `@ariakit/react`. Grammar was fixed in commit `98b440a26`.

### Findings

No substantive issues found. The fix is minimal, well-targeted, and correctly addresses the bug without regressions to existing behavior.

site/src/sandbox/select-combobox-form-3323/test-browser.ts
```diff
@@ -0,0 +1,14 @@
...
+    page,
+    q,
+  }) => {
+    await q.combobox("Favorite fruit").click();
+    await page.keyboard.type("asdasd");
+    await test.expect(q.option("Apple")).toBeHidden();
+    await page.keyboard.press("Enter");
+    await test.expect(page.getByText("Form submitted")).toBeHidden();
+  });
+});
```
diegohaz, line 36: The test covers the exact bug scenario well. Consider adding a complementary test that verifies Enter still selects an item when matches exist (i.e., the happy path is not regressed by this change). Something like:

```ts
test("pressing Enter with a matching item selects it", async ({
  page,
  q,
}) => {
  await q.combobox("Favorite fruit").click();
  await page.keyboard.press("ArrowDown");
  await page.keyboard.press("Enter");
  await test.expect(q.combobox("Favorite fruit")).toBeHidden();
  await test.expect(page.getByText("Form submitted")).toBeHidden();
});
```

This would provide a self-contained regression guard within the same sandbox rather than relying on coverage from other examples (`combobox-multiple/test-browser.ts`).

Not blocking since the happy path is already covered by existing tests elsewhere, but it would make this sandbox more robust as a standalone repro.

diegohaz (commented):
## Review Summary



**Type:** Initial review
**Baseline:** merge base with `main`

### Scope reviewed

- Core fix in `packages/ariakit-react-core/src/combobox/combobox.tsx`
- Interaction between the Combobox `onKeyDown` (bubble phase) and Composite's `useKeyboardEventProxy` / `onKeyDownCapture` (capture phase)
- Sandbox example and Playwright test
- Changeset accuracy
- Edge cases: inline autocomplete, combobox links, Select+Combobox combos, non-form contexts

### Verification performed

- `pnpm tsc`: 0 errors, 0 warnings
- `pnpm test`: 146 test files, 770 tests passed (no regressions)
- Traced the keyboard event flow through capture (Composite proxy) and bubble (Combobox handler) phases to verify no interference
- Confirmed the `useKeyboardEventProxy` (line 75) checks `event.defaultPrevented` in capture phase, *before* the new `preventDefault()` runs in bubble phase, so item selection via the proxy is unaffected
- Verified modifier-key Enter (Ctrl/Meta/Alt/Shift+Enter) is excluded by the early returns at lines 567-570, so those shortcuts (e.g., Cmd+Enter for opening links in new tabs) are not blocked
- Reviewed existing Enter-key tests: `examples/combobox/test.ts:83` (item selection), `examples/combobox-multiple/test-browser.ts:20` (item selection via Enter), `examples/combobox-links/test-browser.ts:55` (Cmd+Enter link opening)

### Findings

The fix is correct and well-scoped. One optional suggestion was left as an inline comment about adding a complementary happy-path test to the new sandbox.

No blocking issues found.

site/src/sandbox/select-combobox-form-3323/test-browser.ts
diegohaz: Fixed in 94654b77a - added a complementary happy-path test verifying Enter selects an item and does not submit the form.

diegohaz:
Addressed review comment:

**Fixed**
- 94654b77a ([reply](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068923038)): Added happy-path test verifying Enter selects an item without submitting the form.

site/src/sandbox/select-combobox-form-3323/test-browser.ts
```diff
@@ -0,0 +1,25 @@
...
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("pressing Enter with no matching items does not submit the form", async ({
+    page,
+    q,
+  }) => {
+    await q.combobox("Favorite fruit").click();
+    await page.keyboard.type("asdasd");
+    await test.expect(q.option("Apple")).toBeHidden();
+    await page.keyboard.press("Enter");
+    await test.expect(page.getByText("Form submitted")).toBeHidden();
```
diegohaz: Should be `q.text("Form submitted")`

site/src/sandbox/select-combobox-form-3323/test-browser.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  test("pressing Enter with a matching item selects it without submitting the form", async ({
+    page,
+    q,
+  }) => {
+    await q.combobox("Favorite fruit").click();
+    await test.expect(q.option("Apple")).toBeVisible();
+    await page.keyboard.press("Enter");
+    await test.expect(q.option("Apple")).toBeHidden();
+    await test.expect(page.getByText("Form submitted")).toBeHidden();
```
diegohaz: Should be `q.text("Form submitted")`

packages/ariakit-react-core/src/combobox/combobox.tsx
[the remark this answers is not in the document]
diegohaz: Correction: Shift+Enter does submit forms — the reviewer was right. Fixed in df317affe - moved the Enter preventDefault before the modifier key early-returns so all Enter variants (Shift+Enter, Ctrl+Enter, etc.) are caught when the popover is open. Added a test for Shift+Enter.

diegohaz:
Addressed modifier+Enter review comment:

**Fixed**
- df317affe ([reply](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068944020)): Shift+Enter does submit forms — moved Enter preventDefault before modifier key early-returns so all Enter variants are caught when the popover is open. Added a Shift+Enter test.

site/src/sandbox/select-combobox-form-3323/test-browser.ts
diegohaz: Fixed in 79f3f96d7 - replaced page.getByText with q.text.

site/src/sandbox/select-combobox-form-3323/test-browser.ts
diegohaz: Fixed in 79f3f96d7 - replaced page.getByText with q.text.

diegohaz:
Addressed review comments:

**Fixed**
- 79f3f96d7 ([reply 1](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068952176), [reply 2](https://github.com/ariakit/ariakit/pull/5741#discussion_r3068952192)): Replaced `page.getByText` with `q.text` in all three tests.

diegohaz (commented):
All prior review threads have been addressed. Incremental changes since last review (98b440a..79f3f96d7):

- **df317af**: Moved the Enter `preventDefault()` before modifier-key early-returns so Shift+Enter, Ctrl+Enter, etc. are also caught when the popover is open. Added a Shift+Enter test. This correctly addresses the modifier bypass issue from the prior review.
- **79f3f96d7**: Replaced `page.getByText("Form submitted")` with `q.text("Form submitted")` across all three test assertions, resolving the author's self-review comments.

The restructured `onKeyDown` logic is correct:
1. Enter (with or without modifiers) when open: `preventDefault()` + return. The Composite keyboard event proxy runs first in the capture phase and dispatches Enter to the active item for selection, so `preventDefault()` in the bubble phase only blocks form submission without interfering with item selection.
2. Modifier keys (non-Enter) when open: early return, allowing default browser behavior (e.g., Ctrl+A).
3. Non-modifier, non-Enter keys when open: early return via `if (open) return`.
4. Arrow keys when closed: opens the popover.

No new issues found.

PR #5743, opened 2026-04-12
Fix data-enter not being removed when Dialog uses render wrapper

### LINKED ISSUE #5057: Dialog - `data-enter` is never removed from DOM when props are applied to nested DOM element in `render`
### Current behavior

The `data-enter` attribute is supposed to be added and removed automatically depending on the state of the dialog. While this works correctly in most cases, it fails when using the `render` to apply the props to a nested element, as in the [Dialog with scrollable backdrop](https://ariakit.org/examples/dialog-backdrop-scrollable) example.

```tsx
<Ariakit.Dialog
  backdrop={false}
  open={open}
  onClose={() => setOpen(false)}
  className="dialog"
  render={(props) => (
    <div className="backdrop" hidden={!open}>
      <div {...props} /> {/* data-enter is added here but never removed */}
    </div>
  )}
>
```

1. Go to https://ariakit.org/examples/dialog-backdrop-scrollable
2. Open the dialog
3. Find the relevant DOM element in the web inspector

     ```html
     <div
       id=":R4pm5jttqcusq:"
       data-dialog=""
       role="dialog"
       tabindex="-1"
       aria-labelledby=":r8:"
       class="dialog"
       style="--dialog-viewport-height: 929.5px;"
       data-open="true"
       data-enter="true"
     >
     ```
4. Close the dialog
5. Check the DOM element again

     ```html
     <div
       id=":R4pm5jttqcusq:"
       data-dialog=""
       role="dialog"
       tabindex="-1"
       aria-labelledby=":r8:"
       class="dialog"
       style="--dialog-viewport-height: 929.5px; display: none;"
       data-enter="true"
       hidden=""
     >
     ```

Although the `display: none` style and `hidden` attribute have been added and the `data-open` attribute has been removed, `data-enter` has not been removed. I also suspect that `data-leave` is never being added, although I haven't checked for sure. 

This breaks animations on the dialog, since when the dialog is subsequently opened, `data-enter` is already present, so the dialog appears instantaneously. Exit animations relying on `data-enter` being removed are also broken.

### Expected behavior

As on other examples such as https://ariakit.org/examples/dialog-animated, `data-enter` should be removed when the dialog closes.

### Workaround

https://github.com/ariakit/ariakit/issues/5057#issuecomment-3424030062
https://github.com/ariakit/ariakit/issues/5057#issuecomment-4196388971

### Possible solutions

_No response_

diegohaz:
## Motivation

When [`Dialog`](https://ariakit.com/reference/dialog) uses a `render` prop to wrap the dialog element in an outer element (e.g., `render={(props) => <div className="wrapper"><div {...props} /></div>}`), the `data-enter` attribute is never removed when the dialog closes. This breaks CSS styles that depend on `data-enter`, such as `:has([data-enter])` on wrapper elements.

The root cause is in the animation detection flow: when the dialog element itself has no CSS transitions (only the wrapper does, e.g., via `:has([data-enter])`), `getComputedStyle` returns a zero timeout. This correctly sets `animated` to `false` for the immediate unmount optimization, but the `data-enter`/`data-leave` transition lifecycle was not being managed after that point. Specifically:
- `data-enter` was never cleared on close because the transition state cleanup only ran when `animated` was truthy.
- `data-enter` was never re-added on reopen because the animation detection cycle (which sets `transition="enter"`) only runs on mount.

Fixes #5057

## Solution

Added transition state management in the `!animated` branch of the first effect. When animation detection has disabled animations:
- On close (`!open`): clear the transition state so `data-enter` is removed from the element.
- On reopen (after at least one close cycle): restore `transition="enter"` so `data-enter` is re-applied for wrapper styling.

A `hasClosedRef` tracks whether the component has been through a close cycle to distinguish "initially open on mount" from "reopening after close". This prevents the fix from affecting other disclosure-based components (like tab panels) that start open and should not receive `data-enter` on initial mount.

## Workaround

If you cannot update yet, conditionally apply `data-enter` based on `data-open` in the render prop:

```tsx
<Dialog
  render={(props) => (
    <div className="wrapper">
      <div
        {...props}
        // TODO: Remove when https://github.com/ariakit/ariakit/issues/5057 is fixed
        data-enter={props["data-open"] ? props["data-enter"] : undefined}
      />
    </div>
  )}
>
```


site/src/sandbox/dialog-5057/style.css
```diff
@@ -0,0 +1,32 @@
+.wrapper {
```
diegohaz: Use tailwind instead.

site/src/sandbox/dialog-5057/test-browser.ts
```diff
@@ -0,0 +1,42 @@
...
+    // data-enter should be removed when the dialog closes, even if the
+    // dialog element itself has no CSS transitions.
+    // See https://github.com/ariakit/ariakit/issues/5057
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await q.button("OK").click();
+    await test.expect(q.dialog("Success")).not.toBeVisible();
+
+    const hiddenDialog = page.locator("[role='dialog']");
```
diegohaz: Use `q.dialog()`

packages/ariakit-react-core/src/disclosure/disclosure-content.tsx
```diff
@@ -184,13 +184,12 @@ export const useDisclosureContent = createHook<
...
     // If the timeout is zero, there's no animation or transition, either
     // because they weren't defined in the CSS or the duration was explicitly
-    // set to zero. In this scenario, we can halt the animation right away
-    // and, if we're entering, we can set the animatedRef to false to bypass
-    // the leave animation.
+    // set to zero. In this scenario, we halt the animation right away. We
+    // keep the animated state as-is so that the transition lifecycle
+    // (data-enter/data-leave) continues to be managed on subsequent
+    // open/close cycles, even when the content element itself has no CSS
+    // transitions (e.g., a wrapper uses :has([data-enter]) instead).
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because the comment at lines 99-101 is still accurate — `animated` IS still set to `false` later on in the zero-timeout branch at line 203 (`store.setState("animated", false)`). The current fix does not change that behavior; it only adds transition state management in the `!animated` branch of the first effect.

packages/ariakit-react-core/src/disclosure/disclosure-content.tsx
```diff
@@ -111,7 +112,23 @@ export const useDisclosureContent = createHook<
...
+      // so wrapper elements using :has([data-enter]) work correctly even
+      // when the content element itself has no transitions.
+      if (!open) {
+        hasClosedRef.current = true;
+        setTransition((current) =>
+          current === "enter" || current === "leave" ? null : current,
+        );
+      } else if (hasClosedRef.current) {
+        // On reopen after animation detection disabled animations, restore
+        // data-enter so enter styles are re-applied.
```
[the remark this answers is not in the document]
diegohaz: Fixed in fe57caa69 - Reset `hasClosedRef.current` to `false` after restoring `data-enter` on reopen.

site/src/sandbox/dialog-5057/style.css
diegohaz: Fixed in c4496467c - Replaced style.css with tailwind utility classes inline.

site/src/sandbox/dialog-5057/test-browser.ts
diegohaz: Fixed in 8e9ae2312 - Replaced `page.locator("[role='dialog']")` with `q.dialog("Success", { includeHidden: true })` and removed unused `page` parameter.

diegohaz:
Addressed all unresolved review comments.

**Fixed**
- fe57caa69 ([reply](https://github.com/ariakit/ariakit/pull/5743#discussion_r3068903030)): Reset `hasClosedRef` after reopen to avoid redundant state updates
- c4496467c ([reply](https://github.com/ariakit/ariakit/pull/5743#discussion_r3068903106)): Replaced style.css with tailwind utility classes
- 8e9ae2312 ([reply](https://github.com/ariakit/ariakit/pull/5743#discussion_r3068903154)): Replaced `page.locator` with `q.dialog()` query helper

**Invalid**
- copilot comment about outdated `animated` comment ([reply](https://github.com/ariakit/ariakit/pull/5743#discussion_r3068902942)): The comment at lines 99-101 is still accurate — `animated` is still set to `false` in the zero-timeout branch

packages/ariakit-react-core/src/disclosure/disclosure-content.tsx
```diff
@@ -111,7 +112,24 @@ export const useDisclosureContent = createHook<
   }, [store]);
 
   useSafeLayoutEffect(() => {
-    if (!animated) return;
+    if (!animated) {
+      // When animation detection has disabled animations (e.g., no CSS
+      // transition was detected on the content element), manage data-enter
+      // so wrapper elements using :has([data-enter]) work correctly even
+      // when the content element itself has no transitions.
```
diegohaz, line 119: Nit: This functional update is equivalent to `setTransition(null)`. The condition `current === "enter" || current === "leave"` covers all non-null values of `TransitionState`, and `null` maps to `null`, so the result is always `null`. The simpler form reads more clearly and avoids the reader wondering what fourth state is being preserved:

```ts
setTransition(null);
```

diegohaz (commented):
## Review Summary

**Type**: Initial review
**Baseline**: merge base `2d47020b7` to PR head `8e9ae2312`

### Scope reviewed

- Core fix in `packages/ariakit-react-core/src/disclosure/disclosure-content.tsx`
- New sandbox `site/src/sandbox/dialog-5057/` (example, preview plumbing, browser tests)
- Changeset `.changeset/300-dialog-data-enter-render-wrapper.md`
- Impact analysis on all consumers of `useDisclosureContent`: Dialog, DialogBackdrop, TabPanel, SelectList, MenuList, ComboboxList

### Verification performed

- `pnpm tsc`: passes (0 errors, 0 warnings)
- `pnpm test -- --run packages/ariakit-react-core/src/disclosure`: 146 test files, 770 tests pass
- Traced the full `animated`/`transition`/`hasClosedRef` lifecycle through:
  - Initial open (Dialog with render wrapper, no CSS transitions on content element)
  - Close (data-enter removal)
  - Reopen (data-enter re-addition)
  - Tab panel starting open (no spurious data-enter on mount)
  - `unmountOnHide` remount cycle (refs reset, animated path handles initial enter)
  - DialogBackdrop (shared disclosure content lifecycle)

### Findings

No actionable issues found. The fix is correct and well-targeted:

1. **`hasClosedRef` design**: Correctly distinguishes "initially open on mount" from "reopening after close," which prevents this fix from affecting TabPanel and other disclosure-based components that start open and should not receive `data-enter` on initial mount.

2. **Updater function on close**: The `setTransition((current) => current === "enter" || current === "leave" ? null : current)` pattern avoids unnecessary state updates when transition is already `null`.

3. **Zero-timeout branch unchanged**: `store.setState("animated", false)` still runs on enter with zero timeout, keeping the immediate-unmount optimization intact. The comment update at lines 203-208 accurately describes this behavior.

4. **Comment at lines 99-101 remains accurate**: The Copilot review flagged this, but the author correctly noted that `animated` IS still set to `false` by the zero-timeout branch.

5. **CodeRabbit's remount concern does not apply**: On remount, `hasClosedRef` is `false` (fresh ref), so the synchronous `setTransition("enter")` path is not taken. The normal animated path via `afterPaint` handles the initial enter.

6. **Test coverage**: Both key scenarios are covered (close removes `data-enter`, reopen re-adds it). Tests use `q.dialog()` helpers and `includeHidden` correctly.

7. **Changeset and sandbox**: Follow existing conventions.

### Areas not fully validated

- Browser tests (`test-browser.ts`) were not run locally as they require the dev server and Playwright. These should be validated in CI.
- Did not test with `unmountOnHide={true}` combined with the render wrapper pattern, though the code path analysis confirms correctness.

diegohaz (commented):
## Review Summary

**Correctness**: The fix is sound. The root cause is well-identified: when the dialog element itself has no CSS transitions (because transitions live on the render wrapper), `getComputedStyle` returns zero timeout, animation detection sets `animated=false`, and the previous code's early `return` left `data-enter` permanently stuck. The `hasClosedRef` approach correctly distinguishes initial mount from reopen.

**Lifecycle trace verified**:
- **Open flow**: Component mounts → `animated` set to `true` → `transition` set to `"enter"` → CSS detection finds zero timeout → sets `animated=false` → new `!animated` branch runs with `open=true, hasClosedRef=false` → neither sub-branch fires → `transition` stays `"enter"` (correct: `data-enter` remains while open)
- **Close flow**: `open=false` → `!animated, !open` → sets `hasClosedRef=true`, clears `transition` to `null` (correct: `data-enter` removed)
- **Reopen flow**: `open=true` → `!animated, open, hasClosedRef=true` → restores `transition="enter"`, resets `hasClosedRef` (correct: `data-enter` re-applied)

**TabPanel and other consumers**: TabPanel creates its own disclosure store and passes through the same hook. The initially-selected panel (starting with `open=true`) correctly gets `data-enter` because the `hasClosedRef=false` branch does nothing, preserving the existing `transition="enter"`. Deselecting and reselecting works correctly via the `hasClosedRef` mechanism. No regressions expected.

**Comment fix**: The update at line 203 replacing "animatedRef" with "animated state" is accurate — the code calls `store.setState("animated", false)`, not a ref.

**Tests**: The two browser tests cover the core scenarios from issue #5057 (close removes `data-enter`, reopen restores it). The test structure follows existing sandbox test patterns.

**Changeset**: Accurate and well-scoped.

One minor nit left inline about simplifying the `setTransition` call. Overall this is a clean, well-targeted fix.

LGTM

packages/ariakit-react-core/src/disclosure/disclosure-content.tsx
diegohaz: Fixed in 59131cbcc - Simplified to `setTransition(null)` since the functional updater was equivalent.

diegohaz:
**Fixed**
- 59131cbcc ([reply](https://github.com/ariakit/ariakit/pull/5743#discussion_r3068923327)): Simplified `setTransition` functional updater to `setTransition(null)` since all non-null `TransitionState` values were already covered

PR #5744, opened 2026-04-12
Fix events not handled in popup window

### LINKED ISSUE #4904: Events not handled in popup window
### Current behavior

When Ariakit components are rendered in a popup window (opened using [`window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open)), some events are not handled correctly in the new window. I believe this applies to any event listener that is added naively to the `document` or `window` globals. The two examples I'm demonstrating in this bug report are Menu's [`hideOnInteractOutside`](https://ariakit.org/reference/menu#hideoninteractoutside) and [`hideOnEscape`](https://ariakit.org/reference/menu#hideonescape) props.

<details>
<summary>Why use popup windows?</summary>

The main use case for opening popup windows is when building Electron-based apps. You've probably seen this in action in an app like VSCode where you can drag tabs out of the main window to create [floating windows](https://code.visualstudio.com/docs/configure/custom-layout#_floating-windows).

</details>

1. Go to [stackblitz.com/edit/1xpe2xbw](https://stackblitz.com/edit/1xpe2xbw?file=menu.tsx,index.tsx) and ["open preview in a new tab"](https://github.com/user-attachments/assets/c95d1ee2-1b75-4cc0-93a4-392d2a3ea31e). StackBlitz may ask you to click "Connect to project".
2. Click on the "Open popout window" button. The browser may ask you to allow popup permissions.
3. Within the popup window, click on the "Actions" button to open the Menu. The Menu opens correctly and is keyboard navigable so far.
4. Now try closing the Menu by clicking outside it (in the popup window) or by pressing <kbd>Escape</kbd>. **The Menu remains open**.
5. Additionally, try activating one of the MenuItems by clicking on it or pressing <kbd>Enter</kbd>. This will close the Menu, but **focus is not restored** to the Menu's trigger button.

### Expected behavior

- The Menu should close when clicking outside or pressing <kbd>Escape</kbd>.
- Focus should return to the Menu's trigger button after the Menu is closed.

### Workaround

https://github.com/ariakit/ariakit/issues/4904#issuecomment-4223914667

### Possible solutions

To handle these events correctly, [`element.ownerDocument`](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) should be used instead of the global `document`. Notably, I have seen many existing references to `ownerDocument` in Ariakit's source, so there might only be a few remaining instances of `document` or `window`.

There are a few calls to `addGlobalEventListener` (e.g. [1](https://github.com/ariakit/ariakit/blob/eed7307277f62fe8de8656fe5cbfd6a05199376f/packages/ariakit-react-core/src/dialog/dialog.tsx#L493), [2](https://github.com/ariakit/ariakit/blob/eed7307277f62fe8de8656fe5cbfd6a05199376f/packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts#L107)) that are missing a fourth parameter, resulting in default value (`window`) being used.

https://github.com/ariakit/ariakit/blob/eed7307277f62fe8de8656fe5cbfd6a05199376f/packages/ariakit-core/src/utils/events.ts#L220

diegohaz:
Fixes #4904

## Motivation

When Ariakit components are rendered in a popup window (opened via `window.open()`), some events are not handled correctly. `addGlobalEventListener` attaches event listeners to the main `window.document`, but popup windows are separate browsing contexts that are not accessible via `window.frames`. This causes [`hideOnEscape`](https://ariakit.com/reference/dialog#hideonescape), [`hideOnInteractOutside`](https://ariakit.com/reference/dialog#hideoninteractoutside), and focus restoration to fail.

## Solution

- Pass the content element's window as the `scope` parameter to `addGlobalEventListener` calls in [`Dialog`](https://ariakit.com/reference/dialog), so event listeners are attached to the correct document when the dialog is rendered in a different window.
- Fix `isAlreadyFocusingAnotherElement` to check `activeElement` on the content element's document instead of the main window's document, so focus restoration works correctly in popup windows.
- Update `usePreviousMouseDownRef` to accept and forward the window scope, so mousedown tracking for click-outside detection also works in popup windows.

## Workaround

Ariakit's `addGlobalEventListener` attaches event listeners to the main `window.document`. Popup windows opened via `window.open()` are separate browsing contexts, so events fired there never reach those listeners. Additionally, the internal `isAlreadyFocusingAnotherElement` function checks `document.activeElement` on the main window instead of the popup window, causing focus restoration to fail.

The following hook adds the missing event listeners on the popup window's document and fixes focus restoration. It's a no-op when the component is rendered in the main window.

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/4904 is fixed.
function usePopupWindowDismiss(store: Ariakit.MenuStore) {
  const contentElement = Ariakit.useStoreState(store, "contentElement");
  const prevOpenRef = React.useRef(false);

  React.useEffect(() => {
    if (!open) return;
    if (!contentElement) return;

    const doc = contentElement.ownerDocument;
    const win = doc.defaultView;
    if (!win || win === window) return;

    const onKeyDown = (event: KeyboardEvent) => {
      if (event.key !== "Escape") return;
      store.hide();
    };

    const onClick = (event: MouseEvent) => {
      const { contentElement, disclosureElement } = store.getState();
      const target = event.target as Element | null;
      if (!target) return;
      if (contentElement?.contains(target)) return;
      if (disclosureElement?.contains(target)) return;
      store.hide();
    };

    doc.addEventListener("keydown", onKeyDown, true);
    doc.addEventListener("click", onClick, true);

    return () => {
      doc.removeEventListener("keydown", onKeyDown, true);
      doc.removeEventListener("click", onClick, true);
    };
  }, [store, open, contentElement]);

  React.useEffect(() => {
    const wasOpen = prevOpenRef.current;
    prevOpenRef.current = open;

    if (!wasOpen || open) return;

    const { disclosureElement } = store.getState();
    if (!disclosureElement) return;

    const doc = disclosureElement.ownerDocument;
    const win = doc.defaultView;
    if (!win || win === window) return;

    requestAnimationFrame(() => {
      requestAnimationFrame(() => {
        const activeEl = doc.activeElement;
        if (activeEl && activeEl !== doc.body) return;
        disclosureElement.focus();
      });
    });
  }, [store, open]);
}
```

Usage — call the hook with the menu store:

```tsx
function Menu() {
  const store = Ariakit.useMenuStore();
  usePopupWindowDismiss(store);

  return (
    <Ariakit.MenuProvider store={store}>
      <Ariakit.Menu>
        <Ariakit.MenuItem>Edit</Ariakit.MenuItem>
      </Ariakit.Menu>
    </Ariakit.MenuProvider>
  );
}
```

This also works with `DialogStore` since `MenuStore` extends it.


packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -98,7 +98,9 @@ function useEventOutside({
       // of the content element, we call the listener.
       callListener(event);
     };
-    return addGlobalEventListener(type, onEvent, capture);
+    const { contentElement } = store.getState();
+    const win = contentElement ? getWindow(contentElement) : undefined;
+    return addGlobalEventListener(type, onEvent, capture, win);
   }, [open, capture, store, type, callListener]);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1b98d0c — `useEventOutside` now subscribes to `contentElement` via `useStoreState` and includes it in both effect dependency arrays, so listeners are re-registered when the content element changes.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -514,7 +514,9 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     // element so we can listen to the Escape key anywhere in the document, even
     // when the dialog is not focused. By using the capture phase, users can
     // call `event.stopPropagation()` on the `hideOnEscape` function prop.
-    return addGlobalEventListener("keydown", onKeyDown, true);
+    const dialog = ref.current;
+    const win = dialog ? getWindow(dialog) : undefined;
+    return addGlobalEventListener("keydown", onKeyDown, true, win);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1b98d0c — the effect now derives the window scope from the reactive `contentElement` (subscribed via `useStoreState`) instead of `ref.current`, and includes `contentElement` in the dependency array.

packages/ariakit-react-core/src/dialog/dialog.tsx
```diff
@@ -514,7 +514,9 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
   }, [store, domReady, mounted, hideOnEscapeProp]);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1b98d0c — adopted the suggested approach. The effect now uses the reactive `contentElement` from `useStoreState` instead of `ref.current` and includes it in the dependency array.

packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 1b98d0c — `useEventOutside` now subscribes to `contentElement` via `useStoreState` at the function level and includes it in both effect dependency arrays, matching the pattern in the parent `useHideOnInteractOutside`.

diegohaz:
## Review comments addressed

**Fixed**

- 1b98d0c ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068911635)): `dialog.tsx` — Use reactive `contentElement` from `useStoreState` instead of `ref.current` for the Escape listener scope, and add `contentElement` to the dependency array.
- 1b98d0c ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068911709)): `dialog.tsx` — Same fix, addressing CodeRabbit's suggestion.
- 1b98d0c ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068911522)): `use-hide-on-interact-outside.ts` — Subscribe to `contentElement` via `useStoreState` in `useEventOutside` and include it in both effect dependency arrays.
- 1b98d0c ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068911832)): `use-hide-on-interact-outside.ts` — Same fix, addressing CodeRabbit's suggestion.

packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -119,7 +120,9 @@ export function useHideOnInteractOutside(
   interactedOutsideRef?: MutableRefObject<boolean>,
 ) {
   const open = useStoreState(store, "open");
-  const previousMouseDownRef = usePreviousMouseDownRef(open);
+  const contentElement = useStoreState(store, "contentElement");
+  const contentWindow = contentElement ? getWindow(contentElement) : undefined;
+  const previousMouseDownRef = usePreviousMouseDownRef(open, contentWindow);
```
diegohaz, line 125: The \`contentWindow\` value is derived on every render without memoization:

```ts
const contentWindow = contentElement ? getWindow(contentElement) : undefined;
```

This is fine for referential stability (same element always yields the same \`Window\` object), but the value transitions from \`undefined\` to a \`Window\` when \`contentElement\` first mounts. That causes \`usePreviousMouseDownRef\` to tear down and re-register its \`mousedown\` listener, which is actually the desired behavior.

However, unlike the sibling \`useEventOutside\` calls (which internally subscribe to \`contentElement\` and guard against \`null\` in the event handler), \`usePreviousMouseDownRef\` has no guard inside its handler — when \`contentElement\` is still \`null\` and \`open\` is already \`true\`, the mousedown listener briefly runs on the main \`window\` and captures targets from the wrong document. This is unlikely to cause a user-visible bug (the previous mousedown target is only checked on click, by which time the effect has re-run), but it is a subtle timing window worth being aware of.

No code change needed — just flagging for awareness.

diegohaz (commented):
## Review summary

**Type**: Initial review
**Baseline**: `origin/main` to `1b98d0c9112cde6ca0ccc9c1277421d5ae61f61d`
**Scope**: All 8 changed files — 3 source files in `packages/ariakit-react-core/src/dialog/`, 1 changeset, and 4 sandbox/test files.

### What was checked

- **`addGlobalEventListener` signature**: Confirmed it accepts the 4th `scope?: Window` parameter (default `window`). The function attaches listeners to `scope.document` and recurses into child frames. Passing the popup window's `Window` object correctly scopes listeners to the popup document.

- **`isAlreadyFocusingAnotherElement` fix**: `getActiveElement(dialog)` now calls `getDocument(dialog).activeElement` instead of falling back to the main `document.activeElement`. This is correct — in a popup window, the main document's `activeElement` would not reflect focus state in the popup.

- **Dependency array changes**: `contentElement` was added to effect dep arrays in both `dialog.tsx` (hide-on-escape) and `use-hide-on-interact-outside.ts` (useEventOutside, useHideOnInteractOutside). This follows the existing pattern already established in the auto-focus effect (line 381 of dialog.tsx). The effects correctly guard against `contentElement` being `null` before mount, and re-run once it becomes available.

- **`contentWindow` derivation**: Not memoized, but `getWindow()` returns the same `Window` reference for the same element, so referential equality holds across renders. The transition from `undefined` to a `Window` on mount correctly triggers effect re-registration.

- **`usePreviousMouseDownRef` scope parameter**: Clean addition. The `scope` parameter flows through to `addGlobalEventListener`. When `scope` is `undefined` (before `contentElement` mounts), the default `window` is used, which matches pre-existing behavior.

- **`useEventOutside` refactor**: `contentElement` was moved from `store.getState()` (read inside layout effect) to `useStoreState()` (reactive subscription). This is necessary so the effect re-runs when `contentElement` changes, re-registering the listener on the correct window.

- **Test file**: Uses `test-chrome.ts` (Chrome-only) instead of `test-browser.ts` (cross-browser). This is appropriate — `window.open()` popup behavior varies across browsers, and Chrome is the primary target for this fix. The test correctly uses `query(popup)` from `@ariakit/test/playwright` to scope assertions to the popup page context.

- **Sandbox**: Named `menu-4904`, following the `component-issueNumber` convention for issue repro sandboxes.

- **Changeset**: Accurately describes the fix and lists both `@ariakit/react-core` and `@ariakit/react` as `patch` releases.

### Findings

One informational inline comment about a brief timing window in `usePreviousMouseDownRef` (no action required).

### Not fully validated

- Other consumers of `addGlobalEventListener` outside the dialog folder (e.g., `hovercard`, `focusable`, `hooks.ts`) do not pass a `scope` parameter. These would also fail if rendered in a popup window. This is outside the scope of this PR but could be addressed in a follow-up.
- The tests were not run locally (popup window tests require a browser environment with `window.open` support).

### Verdict

The fix is correct, well-scoped, and follows existing patterns in the codebase. No blocking issues found.

diegohaz (commented):
## Review summary

**Type:** Initial review
**Baseline:** merge base `2d47020b7` (main) to PR head `1b98d0c91`
**Scope:** All 8 changed files — library code in `dialog.tsx`, `use-hide-on-interact-outside.ts`, `use-previous-mouse-down-ref.ts`; sandbox and tests in `menu-4904/`; changeset.

### What was checked

- **Correctness of the scope parameter threading.** Verified that all three `addGlobalEventListener` call sites in the dialog directory now pass the content element's window as `scope`. The `addGlobalEventListener` implementation defaults `scope` to the main `window`, so passing `undefined` when `contentElement` is null preserves existing behavior for the normal (non-popup) case.

- **`isAlreadyFocusingAnotherElement` fix.** Confirmed that `getActiveElement(dialog)` correctly resolves via `getDocument(node).activeElement`, so focus restoration checks the popup window's active element instead of the main window's.

- **Reactive `contentElement` from `useStoreState`.** The final commit (`1b98d0c`) addressed the prior review feedback (Copilot and CodeRabbit): `useEventOutside` now subscribes to `contentElement` via `useStoreState` instead of reading it once from `store.getState()` during effect setup. Both effects include `contentElement` in their dependency arrays so listeners re-register when the content element changes. The escape key handler in `dialog.tsx` does the same. Verified that these changes are present and correct.

- **`usePreviousMouseDownRef` scope parameter.** The new `scope?: Window` parameter is correctly threaded from `useHideOnInteractOutside` through to `addGlobalEventListener`. The `useEffect` dependency array includes `scope`, so the mousedown listener re-attaches if the content window changes.

- **Cross-document utility safety.** `getDocument()`, `getWindow()`, `getActiveElement()`, `contains()`, `walkTreeOutside()`, `isInDocument()`, and `isDisclosure()` all resolve documents from element references (via `ownerDocument`), not from global `document`. This means they work correctly when elements live in a popup window.

- **`addGlobalEventListener` error handling.** Both `addEventListener` and `removeEventListener` are wrapped in `try/catch`, so closed popup windows won't cause runtime errors during cleanup.

- **Event handler freshness.** The `onEvent` callback inside `useEventOutside` correctly reads `contentElement` and `disclosureElement` from `store.getState()` at event time (not from the closure), ensuring containment checks use the latest state.

- **Test coverage.** The three browser tests cover Escape, click outside, and focus restoration — the three behaviors described in the PR and the linked issue. Using `test-chrome.ts` to skip Safari is appropriate given the documented CI segfault in headless WebKit.

- **Sandbox structure.** The `menu-4904/` sandbox follows the repository naming convention for issue repros. The `usePopoutWindow` hook is a reasonable minimal reproduction of the popup window pattern.

- **Changeset.** Correctly scoped as a `patch` for `@ariakit/react-core` and `@ariakit/react`, with accurate description.

### Validation performed

- Read all changed files and their surrounding context from the PR head commit.
- Traced call sites for `addGlobalEventListener`, `getActiveElement`, `getWindow`, `getDocument`, and `useStoreState` across the dialog module.
- Verified that no other `addGlobalEventListener` call sites in the dialog directory were missed.
- Checked `walkTreeOutside`, `markTreeOutside`, and `disableTreeOutside` for document-global assumptions (none found — they all resolve from element references).
- Reviewed prior review threads from Copilot and CodeRabbit to confirm their findings were addressed in the final commit.

### Findings

No substantive issues found. The changes are minimal, correctly scoped, and address the root cause (event listeners attached to the wrong document) without introducing regressions for the standard single-window case.

packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -119,7 +120,9 @@ export function useHideOnInteractOutside(
   interactedOutsideRef?: MutableRefObject<boolean>,
 ) {
   const open = useStoreState(store, "open");
-  const previousMouseDownRef = usePreviousMouseDownRef(open);
+  const contentElement = useStoreState(store, "contentElement");
+  const contentWindow = contentElement ? getWindow(contentElement) : undefined;
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because the previous mousedown target is only checked on click (inside the `useEventOutside` click listener), and by then `contentElement` is set and the effect has re-run with the correct window scope. The brief timing window where the listener runs on the main window cannot produce a user-visible bug. See the [reviewer analysis](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068917021) for details.

packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Acknowledged — agree this is a benign timing window. The mousedown target is only consumed in the click handler, which runs after `contentElement` is set and the effect has re-registered on the correct window.

diegohaz:
## Review comments addressed (round 2)

**Invalid**

- ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068922636)): Copilot suggested using `disclosureElement` as fallback scope for `usePreviousMouseDownRef`. Not needed — the mousedown target is only consumed in the click handler, by which time `contentElement` is set and the effect has re-registered on the correct window.

- ([reply](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068922725)): Reviewer observation about benign timing window in `usePreviousMouseDownRef`. Acknowledged — no code change needed.

site/src/sandbox/menu-4904/test-chrome.ts
```diff
@@ -0,0 +1,57 @@
+import { query } from "@ariakit/test/playwright";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("menu in popup window closes on Escape", async ({ page, context }) => {
+    const popupPromise = context.waitForEvent("page");
+    await page.getByText("Open popout window").click();
```
diegohaz: Should be q.button

site/src/sandbox/menu-4904/test-chrome.ts
```diff
@@ -0,0 +1,57 @@
+import { query } from "@ariakit/test/playwright";
```
diegohaz: `query` is provided by `withFramework`. No need to import here.

site/src/sandbox/menu-4904/test-chrome.ts
```diff
@@ -0,0 +1,57 @@
...
+    await popup.keyboard.press("Escape");
+    await test.expect(pq.menu()).not.toBeVisible();
+  });
+
+  test("menu in popup window closes on click outside", async ({
+    page,
+    context,
+  }) => {
+    const popupPromise = context.waitForEvent("page");
+    await page.getByText("Open popout window").click();
```
diegohaz: q.button

site/src/sandbox/menu-4904/test-chrome.ts
```diff
@@ -0,0 +1,57 @@
...
+    await popup.locator("body").click({ position: { x: 10, y: 300 } });
+    await test.expect(pq.menu()).not.toBeVisible();
+  });
+
+  test("menu in popup window restores focus to trigger on close", async ({
+    page,
+    context,
+  }) => {
+    const popupPromise = context.waitForEvent("page");
+    await page.getByText("Open popout window").click();
```
diegohaz: q.button

site/src/sandbox/menu-4904/test-chrome.ts
diegohaz: Fixed in 4dc81be7d - removed the standalone `query` import and destructured it from `withFramework` instead.

site/src/sandbox/menu-4904/test-chrome.ts
diegohaz: Fixed in 4dc81be7d - replaced `page.getByText("Open popout window")` with `q.button("Open popout window")`.

site/src/sandbox/menu-4904/test-chrome.ts
diegohaz: Fixed in 4dc81be7d - replaced `page.getByText("Open popout window")` with `q.button("Open popout window")`.

site/src/sandbox/menu-4904/test-chrome.ts
diegohaz: Fixed in 4dc81be7d - replaced `page.getByText("Open popout window")` with `q.button("Open popout window")`.

diegohaz:
**Fixed**

- 4dc81be7d ([#1](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068935278)): Removed standalone `query` import — now destructured from `withFramework`
- 4dc81be7d ([#2](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068935323), [#3](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068935336), [#4](https://github.com/ariakit/ariakit/pull/5744#discussion_r3068935355)): Replaced `page.getByText("Open popout window")` with `q.button("Open popout window")` in all three tests

diegohaz (commented):
## Re-review summary

**Type:** Re-review (incremental from `1b98d0c` to `3098d36`)
**Risk:** Low

### Changes since last review

Two commits since last reviewed SHA:
- `5ce3fc1` — Simplified changeset format (title on first line, details in body)
- `3098d36` — Removed unused `page` destructurings from all three popup window tests

### Previously unresolved threads — all resolved

The three Copilot threads flagging unused `page` at lines 8, 27, and 47 in `test-chrome.ts` are resolved. `page` is no longer destructured in any test function. Tests now use `q` (from `visualTest` fixtures) for the main page and `query(popup)` for the popup page, which is correct.

### Full PR assessment

All changes look correct:

- **dialog.tsx**: `isAlreadyFocusingAnotherElement` passes `dialog` to `getActiveElement` to resolve the correct document. Escape listener derives the window scope from `contentElement` and includes it in deps.
- **use-hide-on-interact-outside.ts**: `useEventOutside` subscribes to `contentElement` via `useStoreState` and scopes both the focusin and main event listeners to the correct window. `contentWindow` is referentially stable (same `Window` object for same element).
- **use-previous-mouse-down-ref.ts**: Accepts and forwards `scope` parameter cleanly.
- **Tests**: Three Playwright tests cover Escape, click-outside, and focus restoration in popup windows. The `query(popup)` pattern for cross-page assertions is correct.
- **Changeset**: Properly formatted.

No new issues found. LGTM.

PR #5761, opened 2026-04-13
Disable react and next updates outside nextjs workspace

diegohaz:
## Motivation

The root, site, examples, and Ariakit library workspaces intentionally keep `react`/`react-dom` at v18 and `next` at v14 to match the frozen `website` workspace. Only the `nextjs` workspace uses React 19 and Next.js 16. Without explicit Renovate rules, dependency update PRs would incorrectly try to bump these packages across all workspaces, leading to unintended major upgrades.

## Solution

Add two new package rules at the end of the `packageRules` array in `renovate.json`:

- A rule that globally disables updates for `next`, `react`, `react-dom`, `@types/react`, and `@types/react-dom`.
- A second rule that re-enables updates for those same packages when the file path matches `nextjs/**`.

Because Renovate applies later rules with higher precedence, placing the re-enabling rule after the disabling rule ensures the `nextjs` workspace continues to receive updates while all other workspaces are left untouched.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -121,6 +121,7 @@
         "@types/react",
         "@types/react-dom"
       ],
+      "matchUpdateTypes": ["major"],
       "enabled": false
     },
     {
@@ -131,6 +132,7 @@
         "@types/react",
         "@types/react-dom"
       ],
+      "matchUpdateTypes": ["major"],
       "matchFileNames": ["nextjs/**"],
       "enabled": true
     }
--- a/renovate.json
+++ b/renovate.json
@@ -112,6 +112,27 @@
       "automerge": true,
       "automergeType": "pr",
       "platformAutomerge": true
+    },
+    {
+      "matchPackageNames": [
+        "next",
+        "react",
+        "react-dom",
+        "@types/react",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -112,6 +112,27 @@
...
+    {
+      "matchPackageNames": [
+        "next",
+        "react",
+        "react-dom",
+        "@types/react",
+        "@types/react-dom"
+      ],
+      "enabled": false
+    },
```
diegohaz, line 126: We should allow minor and patch version updates here.

diegohaz: Fixed in 3d6bf657f - Added `"matchUpdateTypes": ["major"]` to both rules so only major version updates are disabled. Minor and patch updates will now flow through normally.

diegohaz:
**Fixed**

- 3d6bf657f ([reply](https://github.com/ariakit/ariakit/pull/5761#discussion_r3071114465)): Added `matchUpdateTypes: ["major"]` to the disable/enable rules for react, react-dom, next, @types/react, and @types/react-dom so minor and patch updates are allowed.

PR #5762, opened 2026-04-13
Exclude .claude directory from test runners

diegohaz:
## Motivation

The `.claude` directory (used by Claude Code for worktrees and configuration) can contain test files that match the glob patterns in vitest and Playwright configs, causing them to be picked up unintentionally during test runs.

## Solution

Added `.claude/**` to the exclusion patterns in both test runner configs:

- **vitest.config.ts**: Added `exclude: [...configDefaults.exclude, ".claude/**"]` to extend the built-in defaults (which include `node_modules`, `dist`, etc.) rather than replacing them.
- **playwright.config.ts**: Added `".claude/**"` to the existing `testIgnore` array.

The `site/playwright.config.ts` already scopes tests to `site/src` via `testDir`, so no change was needed there.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -28,7 +28,7 @@ export default defineConfig({
   reportSlowTests: null,
   reporter: CI ? [["github"], ["dot"]] : [["list"]],
   retries: 1,
-  testIgnore: ["site/**"],
+  testIgnore: ["site/**", ".claude/**"],
   webServer: {
     command: "pnpm start",
     reuseExistingServer: !CI,
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -3,7 +3,7 @@ import { dirname, join } from "node:path";
 import reactPlugin from "@vitejs/plugin-react";
 import solidPlugin from "vite-plugin-solid";
 import type { Plugin } from "vitest/config";
-import { defineConfig } from "vitest/config";
+import { configDefaults, defineConfig } from "vitest/config";
 import { sourcePlugin } from "./site/src/lib/source-plugin.ts";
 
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
## Review summary

**Type:** Initial review
**Baseline:** merge base of `origin/main` (`d1e9ab67`) to PR head (`de725a8f`)
**Scope:** All 2 changed files (`vitest.config.ts`, `playwright.config.ts`)

### What was reviewed

- Both changed files in full, plus the site-level `site/playwright.config.ts` for completeness.
- Verified that `configDefaults.exclude` in vitest 4.x contains `["**/node_modules/**", "**/.git/**"]`, confirming the spread preserves the prior effective defaults exactly.
- Verified the `.claude/**` glob pattern against `picomatch` (used by both vitest and Playwright) to confirm it matches all nested paths under `.claude/` and does not affect paths outside it.
- Confirmed the `.gitignore` only excludes `.claude/worktrees` and `.claude/scheduled_tasks.lock`, meaning worktree contents exist on disk and are visible to filesystem-scanning test runners even though they're git-ignored.
- Confirmed `site/playwright.config.ts` scopes via `testDir: "src"` and needs no change.
- Confirmed there are no other vitest or playwright configs in the repo.

### Findings

No issues found. The change is correct, complete, and idiomatic for both test runners.

- The vitest approach of spreading `configDefaults.exclude` before adding `.claude/**` is the documented way to extend (rather than replace) the defaults.
- The Playwright approach of appending to the `testIgnore` array is straightforward and correct.
- The glob pattern `.claude/**` is appropriately scoped to the root-level `.claude` directory.

PR #5861, opened 2026-04-27
Fix perf workflow artifact selection

diegohaz:
## Motivation

The performance workflow can compare mismatched trees on pull requests. The baseline side may download artifacts from the latest successful `main` run even when that run was for an older commit than the PR base SHA, and the current side checks out the raw PR head while the build artifacts come from GitHub's PR merge commit.

This made the perf comment on #5860 report misleading Tailwind regressions because the baseline source, baseline artifacts, current source, and current artifacts were not all from comparable commits.

## Solution

Require downloaded baseline artifacts to come from a successful `app.yml` run whose `head_sha` matches `github.event.pull_request.base.sha`. If no exact run is available yet, the perf job now builds the base artifacts locally instead of reusing stale artifacts.

The current side now checks out `github.sha`, matching the PR merge commit used by the build jobs that produce the current artifacts.

Validation:

- `git diff --check`
- YAML parse with Ruby
- `pnpm exec prettier --check .github/workflows/app.yml`
- Embedded `github-script` syntax check with Node
- Live GitHub API query confirmed `head_sha` filtering returns the expected base run for #5860

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -246,28 +246,32 @@ jobs:
 
       - name: Find base app artifact
         id: find-base-artifact
-        continue-on-error: true
         uses: actions/github-script@v9
         with:
           script: |
+            const baseSha = '${{ github.event.pull_request.base.sha }}';
             const runs = await github.rest.actions.listWorkflowRuns({
               owner: context.repo.owner,
               repo: context.repo.repo,
               workflow_id: 'app.yml',
               branch: '${{ github.event.pull_request.base.ref }}',
+              head_sha: baseSha,
               status: 'success',
               per_page: 1,
             });
             const run = runs.data.workflow_runs[0];
             if (!run) {
-              core.setFailed('No successful run found');
+              core.notice(
+                `No successful app.yml run found for base SHA ${baseSha}. ` +
+                  'Building base artifacts in this job.',
+              );
               return;
             }
-            core.setOutput('run-id', run.id);
+            core.setOutput('run-id', String(run.id));
 
       - name: Download base app artifact
         id: base-artifact
-        if: steps.find-base-artifact.outcome == 'success'
+        if: steps.find-base-artifact.outputs.run-id != ''
         continue-on-error: true
         uses: actions/download-artifact@v8
         with:
@@ -278,7 +282,7 @@ jobs:
 
       - name: Download base Next.js artifact
         id: base-nextjs-artifact
-        if: steps.find-base-artifact.outcome == 'success'
+        if: steps.find-base-artifact.outputs.run-id != ''
         continue-on-error: true
         uses: actions/download-artifact@v8
         with:
@@ -310,9 +314,9 @@ jobs:
           mkdir -p /tmp/perf-baseline
           cp site/.perf-results/baseline*.json /tmp/perf-baseline/ 2>/dev/null || true
 
-      # --- Current (PR branch) ---
-      - name: Checkout PR branch
-        run: git checkout -f ${{ github.event.pull_request.head.sha }}
+      # --- Current (merge commit) ---
+      - name: Checkout merge commit
+        run: git checkout -f ${{ github.sha }}
 
       - name: Set up PR environment
         uses: ./.github/workflows/setup
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -246,6 +246,7 @@ jobs:
 
       - name: Find base app artifact
         id: find-base-artifact
+        continue-on-error: true
         uses: actions/github-script@v9
         with:
           script: |


.github/workflows/app.yml
```diff
@@ -246,28 +246,32 @@ jobs:
 
       - name: Find base app artifact
         id: find-base-artifact
-        continue-on-error: true
         uses: actions/github-script@v9
```
diegohaz, line 250: **Removing `continue-on-error: true` here turns transient API errors into a hard fail of the perf job, even though a graceful fallback already exists.**

The script can still throw on cases that aren't "no run found" — `listWorkflowRuns` rate limits, transient 5xx, network blips. Previously such failures landed in `outcome == 'failure'`, the download steps were skipped, and the `Build base app` / `Build base Next.js app` steps (gated on `steps.base-artifact.outcome != 'success'`) ran instead. After this change, the same errors abort the entire `Performance` job.

The new `outputs.run-id != ''` gate on the download steps already covers the "script threw before setting the output" case — the output is empty, downloads skip, and the local-build fallback runs. So putting `continue-on-error: true` back gives you both:

- the script's intentional `core.notice` path on "no successful run" (already handled),
- *and* a soft landing on transient API errors that you don't actually want to fail the job for.

Suggested:

```suggestion
      - name: Find base app artifact
        id: find-base-artifact
        continue-on-error: true
        uses: actions/github-script@v9
```

diegohaz (commented):
Initial review. Baseline: merge-base of `origin/main` and PR head `535c255b`. Scope: `.github/workflows/app.yml` (the only file changed).

## Findings

- One inline finding on `find-base-artifact`: the removal of `continue-on-error: true` makes transient `listWorkflowRuns` failures (rate limits, 5xx, network) abort the entire perf job instead of falling back to the local-build path that already exists.

No other substantive issues — the two core mechanics of this PR check out:

1. **`head_sha` filtering on `listWorkflowRuns`.** Verified live against this repo: `gh api '/repos/ariakit/ariakit/actions/workflows/app.yml/runs?branch=main&head_sha=a8171ca1dd78a955758a7880fa6445cff3164cbe&status=success&per_page=1'` returns exactly the expected `push` run for that SHA. `branch: main` plus `head_sha` reliably selects the push run for the base SHA, even after main has advanced or been rebased (run association doesn't change retroactively).
2. **`github.sha` for the current-side checkout.** For `pull_request` events `github.sha` is the merge commit (`refs/pull/N/merge`), which is exactly what `actions/checkout@v6` uses by default in the `build` and `build-nextjs` jobs. So the perf job's source now matches the artifacts it downloads, and that closes the mismatch that produced the misleading Tailwind regressions on #5860. This also supersedes the prior fix in 6efaf2a4e, which forced `pull_request.head.sha` for a different reason — losing that `-f` is fine because `git checkout -f ${{ github.sha }}` keeps the force flag.

## What I checked beyond the diff

- `actions/checkout@v6` default ref vs. `github.sha` semantics for `pull_request` (confirmed they line up with the merge commit used by `build` / `build-nextjs`).
- Permissions on the `perf` job: `contents: read` + `pull-requests: write` is enough for `listWorkflowRuns` and the cross-run `download-artifact` against this public repo; no `actions: read` change needed.
- Artifact retention (`retention-days: 7`) interaction with the new lookup: when the base SHA is older than 7 days, the run is found but the download fails — `continue-on-error: true` on the download steps already routes that into the build-from-source fallback. (This is the same reason I'd keep `continue-on-error: true` on `find-base-artifact` itself; see inline.)
- Re-runs / in-progress runs on the base SHA: `status: success` + `per_page: 1` (newest first) picks the latest successful re-run, which is the right choice.
- The `git checkout -f ${{ github.sha }}` line: the merge commit is fetched by the initial `actions/checkout@v6 with: fetch-depth: 0`, so it's reachable locally after switching to the base SHA and back.

## Not validated

- I did not actually trigger a perf workflow run on this branch; the verification is limited to live API queries plus reasoning about the change. The current `Performance` check on the PR was still `IN_PROGRESS` at review time.

diegohaz (commented):
Initial review against `origin/main` (`a8171ca1dd78a955758a7880fa6445cff3164cbe`). I reviewed the `.github/workflows/app.yml` perf job diff, the surrounding artifact/download and perf-result flow, the local setup/install composite actions, prior perf workflow changes, existing review context, and the stated motivation in the PR description.

I did not find any substantive issues to comment on inline.

Validation performed:

- `ruby -ryaml -e 'YAML.load_file(".github/workflows/app.yml")'`
- `pnpm install --frozen-lockfile`
- `pnpm exec prettier --check .github/workflows/app.yml`
- Node syntax check for the updated `actions/github-script` block
- Live GitHub API query for base SHA `a8171ca1dd78a955758a7880fa6445cff3164cbe`, which returned successful `app.yml` run `25018676052` with non-expired `app-dist` and `nextjs-dist` artifacts
- Observed PR workflow run `25020373568` complete successfully, including the Performance job's base artifact lookup/downloads, merge-commit checkout, current artifact downloads, perf comparison, and PR comment update

I did not force a no-exact-base-run case in CI; I reviewed that fallback path from the workflow expressions and step outcomes instead.

diegohaz (commented):
**Re-review** (same `HEAD` as the prior `ariakit-pr-review` run that recorded `head_sha=535c255b0e4e210f57a097ba86a8265177a5bd02`). There are no new commits on the PR head since that baseline, so the incremental diff is empty; I re-read the full `perf` job in `.github/workflows/app.yml` and re-checked the surrounding `build` / `build-nextjs` / artifact download flow for consistency with the PR description.

**Result:** I did not identify any *new* substantive issues beyond the existing review thread on `Find base app artifact` (whether to restore `continue-on-error: true` for transient `listWorkflowRuns` failures). The file is unchanged on that hunk, so I did not add another inline on the same line range.

**Re-verified in code:** (1) `listWorkflowRuns` is constrained to the PR base ref and `head_sha: github.event.pull_request.base.sha` so the baseline run matches the true merge base, with local baseline builds when no matching success run or when downloads fail. (2) The current side checks out `github.sha` and downloads artifacts from the same workflow run, aligning with the merge-commit builds from `build` / `build-nextjs`. (3) `if: steps.find-base-artifact.outputs.run-id != ''` plus `continue-on-error` on the two download steps still routes into the `Build base app` / `Build base Next.js app` path when `run-id` is unset (including “no run” and failed lookup before output).

**Local validation in this worktree:** `git diff --check origin/main...HEAD` (clean), `ruby -ryaml` parse of the workflow, `pnpm install` + `pnpm exec prettier --check .github/workflows/app.yml`.

I did not re-run a live Actions perf job; timing and API behavior in CI are unchanged from the last review of this `HEAD`.

.github/workflows/app.yml
```diff
@@ -246,28 +246,32 @@ jobs:
...
             if (!run) {
-              core.setFailed('No successful run found');
+              core.notice(
+                `No successful app.yml run found for base SHA ${baseSha}. ` +
+                  'Building base artifacts in this job.',
+              );
               return;
             }
-            core.setOutput('run-id', run.id);
+            core.setOutput('run-id', String(run.id));
```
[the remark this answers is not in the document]
diegohaz: Fixed in fba26c0f730a7766ed0d49925ebaed9029ba142a - restored `continue-on-error: true` so transient base-artifact lookup failures fall back to local base builds.

.github/workflows/app.yml
diegohaz: Fixed in fba26c0f730a7766ed0d49925ebaed9029ba142a - restored `continue-on-error: true` on the artifact lookup step while keeping the exact base-SHA artifact gate.

diegohaz:
**Fixed**

- fba26c0f730a7766ed0d49925ebaed9029ba142a ([reply](https://github.com/ariakit/ariakit/pull/5861#discussion_r3150328793)): Restored `continue-on-error: true` so transient base-artifact lookup failures fall back to local base builds.
- fba26c0f730a7766ed0d49925ebaed9029ba142a ([reply](https://github.com/ariakit/ariakit/pull/5861#discussion_r3150329203)): Kept the exact base-SHA artifact gate while restoring the soft-failure path for the lookup step.

diegohaz (commented):
**Re-review** (incremental since last `ariakit-pr-review` `head_sha` `535c255b0e4e210f57a097ba86a8265177a5bd02` from 2026-04-27T21:33:56Z). Current `HEAD` is `fba26c0f730a7766ed0d49925ebaed9029ba142a` (`Restore perf workflow fallback`).

**Diff covered:** a single hunk in `.github/workflows/app.yml` — `continue-on-error: true` is restored on `Find base app artifact`. I re-read the full `perf` job through that path and confirmed it matches the earlier review recommendation: transient `listWorkflowRuns` errors no longer hard-fail the job; with no `run-id`, the `if: steps.find-base-artifact.outputs.run-id != ''` download gates pass, the download steps are skipped, and the existing `Build base app` / `Build base Next.js app` conditions (`outcome != 'success'`) still route to a local base build, consistent with the PR description’s “exact base run or build locally” story.

I did **not** add new inline comments: the only substantive topic from the last review is addressed by this commit (and you already left thread replies on the prior inline notes), and the updated lines do not introduce a separate issue.

**What I could not re-assert from here:** a fresh live `listWorkflowRuns` + Performance job run; behavior matches the same reasoning as the prior pass.

**Local checks:** `git diff --check origin/main...HEAD` (clean), Ruby YAML load, `pnpm exec prettier --check .github/workflows/app.yml`.

diegohaz (commented):
**Re-review.** Prior baseline: `535c255b` (last marked review). New head: `fba26c0f`. One incremental commit, `+1 / 0` to `.github/workflows/app.yml`.

## Status of prior findings

- **`find-base-artifact` losing `continue-on-error: true`** — resolved in `fba26c0f`. Verified line 249 now reads `continue-on-error: true`, matching the suggested patch exactly. Walked the four control paths against the new file to confirm the fallback is correct in each:
  - Script throws (rate limit / 5xx / network blip) → step `outcome` is `failure`, `run-id` never set → `if: outputs.run-id != ''` is false on both download steps → `Build base app` / `Build base Next.js app` run via the `outcome != 'success'` gate. ✓
  - Script "no run found" → `core.notice`, returns without setting output → same fallback path as above. ✓
  - Script finds run, downloads succeed → fallback gate is false, local build skipped. ✓
  - Script finds run, download fails (e.g., 7-day retention expired the artifact even though the run still exists) → `continue-on-error: true` on the download steps keeps the job alive, fallback gate flips to true, local build runs. ✓

## New findings

None. The incremental commit is exactly the minimal patch suggested in the prior review and introduces no other behavior change.

## Validation

- `git diff 535c255b..fba26c0f -- .github/workflows/app.yml` — single hunk, `+1 / 0`.
- Re-read the full `perf` job with the new line in place; surrounding download / fallback gates are unchanged.

PR #5862, opened 2026-04-27
Add optional perf profiling

diegohaz:
## Motivation

The current perf harness reports aggregate scripting and rendering timings, but it does not help identify which JavaScript functions or CSS selectors contributed to a slow measured render. That makes regressions harder to investigate after `test-perf` flags a scripting or style recalculation change.

## Solution

Add optional profiling to `site/src/test-utils/perf.ts` without changing the default perf path:

- `scriptProfile` / `PERF_SCRIPT_PROFILE=true` records a CDP CPU profile around each measured interaction and stores the top JavaScript functions by self time.
- `selectorProfile` / `PERF_SELECTOR_PROFILE=true` or `PERF_CSS_SELECTOR_PROFILE=true` records Chrome selector stats tracing and stores the top codebase CSS selectors by elapsed matching time.
- `profileLimit` controls how many entries are retained per profile type.

The collected profiles are stored under each perf result's `profiles` field and `perf-compare` now renders script and selector profile tables in the full breakdown when profile data is present.

## Verification

- `pnpm lint-fix`
- `pnpm exec oxlint site/src/test-utils/perf.ts site/src/test-utils/perf-compare.ts`
- `pnpm -F site exec tsc --noEmit --ignoreConfig --pretty false --target es2018 --module nodenext --moduleResolution nodenext --types node --lib DOM,ESNext --strict --noUncheckedIndexedAccess --allowImportingTsExtensions src/test-utils/perf.ts src/test-utils/perf-compare.ts`
- `pnpm -F site run perf-compare`
- Browser smoke test with both `scriptProfile` and `selectorProfile` enabled against a synthetic page, verifying both JS function and CSS selector entries are captured.

### DIFF
--- a/site/src/test-utils/perf-compare.ts
+++ b/site/src/test-utils/perf-compare.ts
@@ -27,6 +27,7 @@ interface ComparisonRow {
 interface ComparisonSummary {
   rows: ComparisonRow[];
   hasSignificantChanges: boolean;
+  currentResults: PerfResult[];
   newTests: PerfResult[];
   removedTests: PerfResult[];
 }
@@ -87,6 +88,14 @@ function formatDelta(delta: number, percent: number, baseline: number): string {
     return `${sign}${formatMs(delta)}`;
   }
   return `${sign}${formatMs(delta)} (${sign}${percent.toFixed(0)}%)`;
+}
+
+function formatPercent(value: number): string {
+  return `${value.toFixed(0)}%`;
+}
+
+function escapeTableCell(value: string): string {
+  return value.replace(/\|/g, "\\|").replace(/\s+/g, " ").trim();
 }
 
 function resultKey(result: PerfResult): string {
@@ -150,6 +159,7 @@ function compare(): ComparisonSummary {
   return {
     rows,
     hasSignificantChanges: rows.some((r) => r.significant),
+    currentResults: current,
     newTests,
     removedTests,
   };
@@ -189,9 +199,62 @@ function formatSummaryTable(rows: ComparisonRow[], keys: string[]): string[] {
   return lines;
 }
 
+function formatScriptProfile(result: PerfResult): string[] {
+  const profile = result.profiles?.script;
+  if (!profile) return [];
+  if (profile.length === 0) return [];
+
+  const lines: string[] = [];
+  lines.push("#### Script profile");
+  lines.push("");
+  lines.push("| Function | Self | Total | Hits | Source |");
+  lines.push("|----------|------|-------|------|--------|");
+
+  for (const item of profile) {
+    const source = `${item.url}:${item.line}:${item.column}`;
+    lines.push(
+      `| ${escapeTableCell(item.functionName)} | ${formatMs(item.selfTime)} | ${formatMs(item.totalTime)} | ${item.hitCount} | ${escapeTableCell(source)} |`,
+    );
+  }
+
+  lines.push("");
+  return lines;
+}
+
+function formatSelectorProfile(result: PerfResult): string[] {
+  const profile = result.profiles?.selectors;
+  if (!profile) return [];
+  if (profile.length === 0) return [];
+
+  const lines: string[] = [];
+  lines.push("#### Selector profile");
+  lines.push("");
+  lines.push(
+    "| Selector | Elapsed | Attempts | Matches | Slow non-match | Stylesheet |",
+  );
+  lines.push(
+    "|----------|---------|----------|---------|----------------|------------|",
+  );
+
+  for (const item of profile) {
+    lines.push(
+      `| ${escapeTableCell(item.selector)} | ${formatMs(item.elapsed)} | ${item.matchAttempts} | ${item.matchCount} | ${formatPercent(item.slowPathNonMatchPercent)} | ${escapeTableCell(item.styleSheetUrl || item.styleSheetId)} |`,
+    );
+  }
+
+  lines.push("");
+  return lines;
+}
+
+function formatProfiles(result?: PerfResult): string[] {
+  if (!result) return [];
+  return [...formatScriptProfile(result), ...formatSelectorProfile(result)];
+}
+
 function formatDetailedBreakdown(
   rows: ComparisonRow[],
   keys: string[],
+  currentByKey: Map<string, PerfResult>,
 ): string[] {
   const lines: string[] = [];
   for (const key of keys) {
@@ -216,12 +279,23 @@ function formatDetailedBreakdown(
       );
     }
     lines.push("");
+    lines.push(...formatProfiles(currentByKey.get(key)));
   }
   return lines;
 }
 
 function formatMarkdown(summary: ComparisonSummary): string {
-  const { rows, hasSignificantChanges, newTests, removedTests } = summary;
+  const {
+    rows,
+    hasSignificantChanges,
+    currentResults,
+    newTests,
+    removedTests,
+  } = summary;
+  const currentByKey = new Map<string, PerfResult>();
+  for (const result of currentResults) {
+    currentByKey.set(resultKey(result), result);
+  }
 
   const allKeys = [...new Set(rows.map((r) => rowKey(r)))];
   const significantKeys = allKeys.filter((key) =>
@@ -248,7 +322,7 @@ function formatMarkdown(summary: ComparisonSummary): string {
   lines.push(`<details>`);
   lines.push(`<summary>Full breakdown (${totalTests} tests)</summary>`);
   lines.push("");
-  lines.push(...formatDetailedBreakdown(rows, allKeys));
+  lines.push(...formatDetailedBreakdown(rows, allKeys, currentByKey));
 
   if (newTests.length > 0) {
     lines.push("### New tests (no baseline)");
@@ -263,6 +337,14 @@ function formatMarkdown(summary: ComparisonSummary): string {
       lines.push(`| ${cells.join(" | ")} |`);
     }
     lines.push("");
+
+    for (const result of newTests) {
+      const profileLines = formatProfiles(result);
+      if (profileLines.length === 0) continue;
+      lines.push(`#### ${result.label}`);
+      lines.push("");
+      lines.push(...profileLines);
+    }
   }
 
   if (removedTests.length > 0) {
--- a/site/src/test-utils/perf-compare.ts
+++ b/site/src/test-utils/perf-compare.ts
@@ -28,8 +28,25 @@ interface ComparisonSummary {
   rows: ComparisonRow[];
   hasSignificantChanges: boolean;
   currentResults: PerfResult[];
+  profileModeMismatches: ProfileModeMismatch[];
   newTests: PerfResult[];
   removedTests: PerfResult[];
+}
+
+interface PersistedComparisonSummary {
+  rows: ComparisonRow[];
+  hasSignificantChanges: boolean;
+  profileModeMismatches: ProfileModeMismatch[];
+  newTests: PerfResult[];
+  removedTests: PerfResult[];
+}
+
+interface ProfileModeMismatch {
+  testFile: string;
+  label: string;
+  profile: "script" | "selectors";
+  baseline: boolean;
+  current: boolean;
 }
 
 // Primary metrics shown in the summary table.
@@ -95,11 +112,19 @@ function formatPercent(value: number): string {
 }
 
 function escapeTableCell(value: string): string {
-  return value.replace(/\|/g, "\\|").replace(/\s+/g, " ").trim();
+  return value
+    .replace(/\\/g, "\\\\")
+    .replace(/\|/g, "\\|")
+    .replace(/[\r\n\t]/g, " ")
+    .trim();
 }
 
 function resultKey(result: PerfResult): string {
   return `${result.testFile}::${result.label}`;
+}
+
+function hasProfile(result: PerfResult, profile: "script" | "selectors") {
+  return (result.profiles?.[profile]?.length ?? 0) > 0;
 }
 
 function compare(): ComparisonSummary {
@@ -117,6 +142,7 @@ function compare(): ComparisonSummary {
   }
 
   const rows: ComparisonRow[] = [];
+  const profileModeMismatches: ProfileModeMismatch[] = [];
   const newTests: PerfResult[] = [];
   const removedTests: PerfResult[] = [];
 
@@ -126,6 +152,18 @@ function compare(): ComparisonSummary {
     if (!base) {
       newTests.push(cur);
       continue;
+    }
+    for (const profile of ["script", "selectors"] as const) {
+      const baselineHasProfile = hasProfile(base, profile);
+      const currentHasProfile = hasProfile(cur, profile);
+      if (baselineHasProfile === currentHasProfile) continue;
+      profileModeMismatches.push({
+        testFile: cur.testFile,
+        label: cur.label,
+        profile,
+        baseline: baselineHasProfile,
+        current: currentHasProfile,
+      });
     }
     for (const metric of ALL_METRICS) {
       const baseVal = base.metrics[metric];
@@ -160,6 +198,7 @@ function compare(): ComparisonSummary {
     rows,
     hasSignificantChanges: rows.some((r) => r.significant),
     currentResults: current,
+    profileModeMismatches,
     newTests,
     removedTests,
   };
@@ -251,6 +290,25 @@ function formatProfiles(result?: PerfResult): string[] {
   return [...formatScriptProfile(result), ...formatSelectorProfile(result)];
 }
 
+function formatProfileModeWarning(mismatches: ProfileModeMismatch[]): string[] {
+  if (mismatches.length === 0) return [];
+
+  const lines: string[] = [];
+  lines.push(
+    ":warning: Profile data differs between baseline and current. Run both sides with the same profiling flags before comparing aggregate metrics.",
+  );
+  lines.push("");
+  lines.push("| Test | Profile | Baseline | Current |");
+  lines.push("|------|---------|----------|---------|");
+  for (const mismatch of mismatches) {
+    lines.push(
+      `| ${escapeTableCell(mismatch.label)} | ${mismatch.profile} | ${mismatch.baseline ? "yes" : "no"} | ${mismatch.current ? "yes" : "no"} |`,
+    );
+  }
+  lines.push("");
+  return lines;
+}
+
 function formatDetailedBreakdown(
   rows: ComparisonRow[],
   keys: string[],
@@ -289,6 +347,7 @@ function formatMarkdown(summary: ComparisonSummary): string {
     rows,
     hasSignificantChanges,
     currentResults,
+    profileModeMismatches,
     newTests,
     removedTests,
   } = summary;
@@ -319,6 +378,7 @@ function formatMarkdown(summary: ComparisonSummary): string {
   }
 
   lines.push("");
+  lines.push(...formatProfileModeWarning(profileModeMismatches));
   lines.push(`<details>`);
   lines.push(`<summary>Full breakdown (${totalTests} tests)</summary>`);
   lines.push("");
@@ -366,6 +426,18 @@ function formatMarkdown(summary: ComparisonSummary): string {
   return lines.join("\n");
 }
 
+function toPersistedSummary(
+  summary: ComparisonSummary,
+): PersistedComparisonSummary {
+  return {
+    rows: summary.rows,
+    hasSignificantChanges: summary.hasSignificantChanges,
+    profileModeMismatches: summary.profileModeMismatches,
+    newTests: summary.newTests,
+    removedTests: summary.removedTests,
+  };
+}
+
 // Main
 const summary = compare();
 const markdown = formatMarkdown(summary);
@@ -373,7 +445,7 @@ const markdown = formatMarkdown(summary);
 mkdirSync(RESULTS_DIR, { recursive: true });
 writeFileSync(
   path.join(RESULTS_DIR, "comparison.json"),
-  JSON.stringify(summary, null, 2),
+  JSON.stringify(toPersistedSummary(summary), null, 2),
 );
 writeFileSync(path.join(RESULTS_DIR, "comparison.md"), markdown);
 
--- a/site/src/test-utils/perf.ts
+++ b/site/src/test-utils/perf.ts
@@ -1,10 +1,13 @@
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import type { CDPSession, Page, TestInfo } from "@playwright/test";
 
 const RESULTS_DIR = path.join(process.cwd(), ".perf-results");
 const DEFAULT_ITERATIONS = 10;
 const DEFAULT_WARMUP = 1;
+const DEFAULT_PROFILE_LIMIT = 10;
+const SCRIPT_PROFILE_INTERVAL = 100;
 const initializedResultFiles = new Set<string>();
 
 // CDP metric names (values are in seconds).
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
Initial review of `7127e86a5b41d47c54c9ed10f9ebef69a0862828` against `origin/main` at `9859d249ce5830864a45ab9821ed1b0358c6b4be`.

Scope reviewed: `site/src/test-utils/perf.ts`, `site/src/test-utils/perf-compare.ts`, perf fixture usage, CI perf workflow, and existing PR review/check context.

Validation run locally:

- `pnpm exec oxlint site/src/test-utils/perf.ts site/src/test-utils/perf-compare.ts`
- `pnpm -F site exec tsc --noEmit --ignoreConfig --pretty false --target es2018 --module nodenext --moduleResolution nodenext --types node --lib DOM,ESNext --strict --noUncheckedIndexedAccess --allowImportingTsExtensions src/test-utils/perf.ts src/test-utils/perf-compare.ts`
- `pnpm -F site run perf-compare` with a local fixture containing script and selector profile rows
- Direct Playwright smoke tests from the `site` workspace covering `scriptProfile` and `selectorProfile` collection through `createPerfMeasure`
- `gh pr checks 5862 --repo ariakit/ariakit`

I did not find additional actionable issues. The only problem I reproduced is the existing markdown escaping issue already reported by CodeQL on `escapeTableCell`, so I am not duplicating it with another inline comment. At the time of review, CodeQL is failing on that alert, while several longer checks are still pending (`Performance`, `Test Plus`, and some browser/visual jobs).

site/src/test-utils/perf.ts
```diff
@@ -103,19 +556,62 @@ async function measureOnce(
...
+    await page.evaluate(() => {
+      const w = window as any;
+      if (w.__perfObserver) {
+        w.__perfObserver.disconnect();
+      }
+    });
+    throw interactionError;
+  }
 
   const after = await cdp.send("Performance.getMetrics");
```
diegohaz: **Perf metrics vs profiling**

`before` is captured before tracing/profiler start, but `after` runs **after** `Profiler.stop` and selector tracing teardown. The ScriptDuration/layout/style deltas therefore include profiler/tracing overhead whenever profiling is enabled, so median aggregates are **not comparable** to runs with profiling off.

For **perf-compare**, baseline and current should use the **same** profiling flags (or deltas will reflect measurement-mode drift). Consider documenting this next to the env flags / `PerfMeasureOptions`, or emitting a short warning in comparison output when profile payloads differ across sides.

site/src/test-utils/perf-compare.ts
```diff
@@ -150,6 +159,7 @@ function compare(): ComparisonSummary {
   return {
     rows,
     hasSignificantChanges: rows.some((r) => r.significant),
+    currentResults: current,
```
diegohaz, line 200: **`comparison.json` payload**

Including full `currentResults` means `comparison.json` (via downstream `JSON.stringify(summary)`) carries complete `PerfResult` objects—including `raw` per-iteration metrics and optional `profiles`—on top of `rows`. That duplicates summarized data and can inflate CI artifacts substantially.

If Markdown is the primary consumer, consider a slimmer persisted summary (omit `raw`/profiles from JSON, or split profile tables out).

site/src/test-utils/perf.ts
```diff
@@ -78,6 +206,330 @@ function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
...
+    column: callFrame.columnNumber + 1,
+    selfTime: 0,
+    totalTime: 0,
+    hitCount: 0,
+  };
+  entries.set(key, entry);
+  return entry;
+}
+
+function parseScriptProfile(
```
diegohaz: **Test coverage**

`parseScriptProfile`, `parseSelectorTrace`, and the merge helpers are good candidates for **small fixture-based unit tests** (sample CDP payload → expected top entries). That would catch regressions if CDP trace shapes shift, without spinning up Playwright.

site/src/test-utils/perf.ts
```diff
@@ -78,6 +206,330 @@ function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
...
+        existingItem.matchAttempts += item.matchAttempts;
+        existingItem.matchCount += item.matchCount;
+        existingItem.invalidationCount += item.invalidationCount;
+        continue;
+      }
+      entries.set(key, { ...item });
+    }
+  }
+
+  for (const item of entries.values()) {
```
diegohaz: **Merge cost (optional)**

The loop over `entries.values()` rebuilds `originalItems` via `profiles.flatMap(...)` per merged selector entry—fine at small limits, but roughly **O(unique selectors × iterations × entries per iteration)**. Worth revisiting only if profiling runs ever push larger slices.

diegohaz (commented):
### PR review summary (comment-only)



**Scope:** Initial review of `origin/main...`7127e86a (PR #5862). Inspected `site/src/test-utils/perf.ts`, `perf-compare.ts`, and merge/profiling flow. Attempted `oxlint` locally in this environment; CLI was not on PATH—no extra commands beyond `gh` API posts.

**Overall:** Optional CDP CPU + selector profiling, opt-in env/flags, cleanup in `finally`, merging across iterations, and `perf-compare` tables for current results are directionally solid and match the stated motivation.

**Inline threads (4):**
1. Aggregate `Performance.getMetrics` deltas include profiler/tracing overhead when profiling is on—document or warn, and keep baseline/current profiling settings aligned for `perf-compare`.
2. `currentResults` in `ComparisonSummary` inflates `comparison.json`; consider slimming serialized output.
3. Suggest small fixture tests for parsing/merge helpers.
4. Optional note on `mergeSelectorProfiles` scan cost if limits grow.

No blocking correctness issues identified from static review.

site/src/test-utils/perf.ts
```diff
@@ -1,10 +1,13 @@
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import type { CDPSession, Page, TestInfo } from "@playwright/test";
 
 const RESULTS_DIR = path.join(process.cwd(), ".perf-results");
 const DEFAULT_ITERATIONS = 10;
 const DEFAULT_WARMUP = 1;
+const DEFAULT_PROFILE_LIMIT = 10;
+const SCRIPT_PROFILE_INTERVAL = 100;
```
diegohaz: This sets V8's CPU sampling interval to 100 µs (10 kHz), which is 10× more aggressive than V8's default of 1000 µs. The same iteration is used to record both the script profile and the `scripting` / `total` metrics that flow to `perf-compare`, so enabling the profile inflates the metrics this run reports.

Real consequences:
- A CI run with `PERF_SCRIPT_PROFILE=true` would compare profiled `scripting` numbers against an unprofiled baseline, producing false positives in the regression check.
- Even local "compare with profiling on" runs ship metrics that are not directly comparable to non-profiling runs, but the trade-off isn't documented anywhere.

Suggest either (a) keeping the V8 default of 1000 µs (a profile of the top hot frames over a 10-iteration measurement is still useful at 1 kHz), or (b) splitting profiling into a second pass whose metrics aren't appended to `medianMetrics`.

site/src/test-utils/perf.ts
```diff
@@ -78,6 +206,330 @@ function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
...
+  }
+
+  for (const item of entries.values()) {
+    const originalItems = profiles.flatMap((profile) =>
+      profile.filter(
+        (entry) =>
+          entry.styleSheetId === item.styleSheetId &&
+          entry.selector === item.selector,
+      ),
+    );
```
diegohaz: This per-entry `flatMap` + `filter` re-scans every input profile for matching `(styleSheetId, selector)` to recompute the weighted slow-path percent. It's `O(unique × iterations × limit)` for what is fundamentally a single-pass aggregation, and the second loop makes the function harder to read than the script-profile merge above.

The weighted contribution can be accumulated in the same loop that sums `elapsed` / `matchAttempts` / etc., and the percent recomputed once at the end:

```ts
interface AccEntry extends PerfSelectorProfileEntry {
  weightedSlowPathSum: number;
}
const entries = new Map<string, AccEntry>();

for (const profile of profiles) {
  for (const item of profile) {
    const key = `${item.styleSheetId}\0${item.selector}`;
    const weightedContribution =
      item.matchAttempts * item.slowPathNonMatchPercent;
    const existing = entries.get(key);
    if (existing) {
      existing.elapsed += item.elapsed;
      existing.matchAttempts += item.matchAttempts;
      existing.matchCount += item.matchCount;
      existing.invalidationCount += item.invalidationCount;
      existing.weightedSlowPathSum += weightedContribution;
      continue;
    }
    entries.set(key, {
      ...item,
      weightedSlowPathSum: weightedContribution,
    });
  }
}

return [...entries.values()]
  .sort((a, b) => b.elapsed - a.elapsed)
  .slice(0, limit)
  .map(({ weightedSlowPathSum, ...item }) => ({
    ...item,
    slowPathNonMatchPercent:
      item.matchAttempts > 0
        ? weightedSlowPathSum / item.matchAttempts
        : 0,
  }));
```

Produces identical numbers (the math is the same weighted average, just rearranged) and removes both the second pass and the inner re-scan.

site/src/test-utils/perf-compare.ts
```diff
@@ -89,6 +90,14 @@ function formatDelta(delta: number, percent: number, baseline: number): string {
   return `${sign}${formatMs(delta)} (${sign}${percent.toFixed(0)}%)`;
 }
 
+function formatPercent(value: number): string {
+  return `${value.toFixed(0)}%`;
+}
+
+function escapeTableCell(value: string): string {
+  return value.replace(/\|/g, "\\|").replace(/\s+/g, " ").trim();
+}
```
diegohaz, line 120: This is what CodeQL flagged as "incomplete string escaping". Replacing `|` with `\|` while leaving existing backslashes untouched is unsound: an input containing `\|` becomes `\\|` — a literal escaped backslash followed by an unescaped pipe, which breaks the table cell.

CSS selectors are the realistic offender, since selector escapes like `\:` (namespace separator) or `\3A` (hex escape) are valid and reach `parseSelectorTrace` verbatim from `selector_timings[].selector`. Stylesheet URLs occasionally contain backslashes via odd source maps too.

Fix — the backslash replace must run first:

```ts
function escapeTableCell(value: string): string {
  return value
    .replace(/\\/g, "\\\\")
    .replace(/\|/g, "\\|")
    .replace(/\s+/g, " ")
    .trim();
}
```

diegohaz (commented):
## Review summary

**Type:** Initial review.
**Baseline:** `9859d249c` (PR base) → `7127e86a5` (PR head).
**Scope:** `site/src/test-utils/perf.ts` and `site/src/test-utils/perf-compare.ts`, plus surrounding context (`fixtures.ts`, the existing `perf-chrome.ts` consumers, and the perf job in `.github/workflows/app.yml`).

**Validated locally:** `nvm use`, `pnpm install`, `pnpm tsc --noEmit` (clean). Did **not** exercise the new flags against a real perf run because nothing in the repo opts into them yet (see below).

### Inline findings

1. `SCRIPT_PROFILE_INTERVAL = 100` — 10× more aggressive than V8's default sampling interval. Inflates the same-iteration `scripting`/`total` metrics that `perf-compare` uses for regression detection.
2. `mergeSelectorProfiles` second pass is `O(N²)` and uses `flatMap+filter` to recompute a weighted average that can be accumulated in the same single-pass loop as the other metrics.
3. `escapeTableCell` in `perf-compare.ts` doesn't escape backslashes (the CodeQL warning). Realistic break case: CSS selector escapes like `\:` or `\3A`.

### Other observations not posted as inline comments

- **Feature is not enabled anywhere in the repo.** Neither the `perf` job in `.github/workflows/app.yml` sets `PERF_SCRIPT_PROFILE` / `PERF_SELECTOR_PROFILE` / `PERF_CSS_SELECTOR_PROFILE`, nor do the existing `perf-chrome.ts` files (`site/src/sandbox/ariakit-tailwind/perf-chrome.ts`, `site/src/examples/dialog-framer-motion/perf-chrome.ts`) pass `scriptProfile`/`selectorProfile` through `perf.measure(...)`. As wired, this is a local-only debugging tool. If that's intentional, fine; if the goal is to surface profiles in the PR comparison comment automatically, the workflow needs to set the env var (and finding #1 becomes more pressing).
- **No unit tests for the new pure helpers.** `parseScriptProfile`, `parseSelectorTrace`, `mergeScriptProfiles`, `mergeSelectorProfiles`, `normalizeProfileUrl`, and `escapeTableCell` are all pure and easy to cover with fixtures of canned CDP payloads. The PR description's verification only exercised a synthetic end-to-end smoke test, so nothing in CI guards against future regressions in the merge math (e.g., the weighted-average shape in finding #2).
- **Selector trace categories include some that the parser never reads.** `parseSelectorTrace` only processes `event.name === "SelectorStats"`, which is emitted from `disabled-by-default-blink.debug`. The other three categories (`devtools.timeline`, `disabled-by-default-devtools.timeline`, `disabled-by-default-devtools.timeline.invalidationTracking`) just inflate CDP traffic during measurement. Worth confirming whether `invalidationTracking` is actually required for `selector_timings[].invalidation_count` to be populated; if not, consider trimming the list.
- **Two env vars for the same toggle.** `PERF_SELECTOR_PROFILE` and `PERF_CSS_SELECTOR_PROFILE` both opt selector profiling in. The alias isn't documented anywhere outside the PR description; pick one and drop the other so future readers don't have to guess which is canonical.
- **`profiles` is silently dropped on `interactionError`.** When `interaction()` throws, the partially-collected profile is built into `profiles` but never returned (we re-throw `interactionError` first). For a debugging tool, an iteration that throws is exactly when you'd want the profile. Low priority — easy to revisit if the use case shows up.

### Not validated

- End-to-end run with `PERF_SCRIPT_PROFILE=true` / `PERF_SELECTOR_PROFILE=true` against the live perf-chrome.ts files. The actual measurement-distortion magnitude in finding #1 is theoretical; an A/B local run would be the way to size it.
- Whether `disabled-by-default-devtools.timeline.invalidationTracking` is required for `invalidation_count` to be non-zero (third bullet above).

site/src/test-utils/perf-compare.ts
```diff
@@ -89,6 +90,14 @@
   return `${sign}${formatMs(delta)} (${sign}${percent.toFixed(0)}%)`;
 }
 
+function formatPercent(value: number): string {
+  return `${value.toFixed(0)}%`;
+}
+
+function escapeTableCell(value: string): string {
+  return value.replace(/\|/g, "\\|").replace(/\s+/g, " ").trim();
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - escapeTableCell now escapes backslashes before pipes so markdown tables cannot be broken by existing escapes.

site/src/test-utils/perf-compare.ts
```diff
@@ -89,6 +90,14 @@ function formatDelta(delta: number, percent: number, baseline: number): string {
...
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - escapeTableCell now only normalizes CR/LF/tab characters instead of collapsing all whitespace, preserving selector/function/URL spaces.

site/src/test-utils/perf.ts
```diff
@@ -103,19 +556,62 @@ async function measureOnce(
   });
 
   const before = await cdp.send("Performance.getMetrics");
+  const selectorTrace = options.selectorProfile
+    ? await startSelectorTrace(
+        cdp,
+        options.styleSheetUrls,
+        options.profileLimit,
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - measureOnce now tears down the perf observer, selector trace, and CPU profiler through the setup and interaction error paths.

site/src/test-utils/perf.ts
```diff
@@ -146,7 +642,20 @@ async function measureOnce(
...
+    styleRecalc,
+    painting,
+    rendering,
+    inp,
+    total,
+  };
+
+  return {
+    metrics,
+    profiles: profiles.script || profiles.selectors ? profiles : undefined,
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - profile payloads are only attached and aggregated when script or selector profiles contain entries.

site/src/test-utils/perf.ts
```diff
@@ -78,6 +206,330 @@ function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
...
+      const parentId = parents.get(currentNode.id);
+      if (parentId == null) break;
+      currentNode = nodes.get(parentId);
+    }
+  }
+
+  return [...entries.values()]
+    .sort((a, b) => b.selfTime - a.selfTime || b.totalTime - a.totalTime)
+    .slice(0, limit);
+}
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - script and selector parsers now return full per-iteration profiles, and profileLimit is applied only after merging measured iterations.

site/src/test-utils/perf.ts
```diff
@@ -103,19 +556,62 @@ async function measureOnce(
...
+
+  if (interactionError) {
+    await page.evaluate(() => {
+      const w = window as any;
+      if (w.__perfObserver) {
+        w.__perfObserver.disconnect();
+      }
+    });
+    throw interactionError;
+  }
```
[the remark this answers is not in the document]
diegohaz: Fixed in 1258eb961 - profiling teardown errors are stored separately so an interaction/setup failure remains the error that gets rethrown.

site/src/test-utils/perf.ts
diegohaz: Fixed in 1258eb961 - removed the high-frequency sampling override and perf-compare now warns when baseline/current profile payloads differ, so aggregate metrics are compared only with matching profiling modes.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in 1258eb961 - comparison.json now omits currentResults, keeping raw/profile payloads out of the persisted summary while retaining markdown profile tables.

site/src/test-utils/perf.ts
diegohaz: Fixed in 1258eb961 - selector profile merging now accumulates the weighted slow-path value in one pass instead of rescanning profiles per merged selector.

site/src/test-utils/perf.ts
diegohaz: Fixed in 1258eb961 - removed the 100 us sampling interval override so profiling uses V8 default sampling, and added a comparison warning for mismatched profiling modes.

site/src/test-utils/perf.ts
diegohaz: Fixed in 1258eb961 - selector merge now accumulates weighted slow-path percentages during the main merge pass and recomputes the percentage after slicing.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in 1258eb961 - escapeTableCell now escapes backslashes before pipes and avoids lossy whitespace collapsing.

diegohaz:
**Fixed**

- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150456958), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457018), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457932)): Hardened markdown table escaping by escaping backslashes before pipes and preserving meaningful selector/function/URL whitespace.
- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457125), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457460)): Wrapped profiling setup/teardown so observers, tracing, and CPU profiler cleanup run while preserving the original interaction/setup error.
- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457269), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457365)): Kept per-iteration profiles untrimmed, applied `profileLimit` only after merging measured iterations, and skipped empty profile payloads.
- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457538), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457764)): Removed the 100 us CPU sampling override and added profile-mode mismatch warnings for baseline/current comparisons.
- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457608)): Slimmed `comparison.json` by omitting full `currentResults` while keeping markdown profile details.
- 1258eb961 ([reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457673), [reply](https://github.com/ariakit/ariakit/pull/5862#discussion_r3150457870)): Reworked selector-profile merging to accumulate weighted slow-path percentages in one pass.

diegohaz (commented):
## Re-review summary

**Type:** Re-review.
**Incremental baseline:** my prior review at `7127e86a5` → current head `1258eb961` (one new commit, `1258eb961 Address perf profiling review feedback`).
**Scope:** the diff of `site/src/test-utils/perf.ts` and `site/src/test-utils/perf-compare.ts` between those two SHAs, plus a re-read of the surrounding `measureOnce` and merge logic to verify the new error-handling structure.
**Validated locally:** `pnpm tsc --noEmit` is clean.

### Inline findings from the previous review — verified

1. **Sampling interval (perf.ts: prior #1)** — `SCRIPT_PROFILE_INTERVAL` and the `Profiler.setSamplingInterval` call are gone (`perf.ts` lines 7 / 743 in the new head), so profiling now uses V8's default 1 ms cadence. The author additionally added a `profileModeMismatches` warning in `perf-compare.ts` (`formatProfileModeWarning`) that surfaces a `:warning:` row whenever a baseline/current pair don't share the same profiling mode — that addresses the "comparing profiled metrics against an unprofiled baseline" risk I raised in the summary, beyond what the original finding asked for. ✅
2. **`mergeSelectorProfiles` complexity (perf.ts: prior #2)** — Now uses a `SelectorProfileAccumulator` with a `weightedSlowPathSum` field, accumulated in the same single pass that sums `elapsed` / `matchAttempts` / `matchCount` / `invalidationCount`, then divided once at the end inside the post-slice `.map(...)`. `O(N²)` `flatMap+filter` removed. Math is unchanged (still `Σ(attempts·pct) / Σ(attempts)`). ✅
3. **`escapeTableCell` backslash escaping (perf-compare.ts: prior #3)** — Backslash replace runs first, then pipe replace; whitespace handling was also tightened to `[\r\n\t]` so meaningful spaces in selectors / function names / URLs survive. The CodeQL alert should clear. ✅

### Summary-only observations from the previous review — status

- **Profile-payload size in `comparison.json`** — addressed: a `PersistedComparisonSummary` is now serialized via `toPersistedSummary`, which strips `currentResults`. Markdown profile tables remain in `comparison.md`. ✅
- **`profiles` silently dropped on `interactionError`** — re-read the rewrite. The new `measureOnce` keeps `measureError` and `profilingError` separate, runs cleanup unconditionally for whichever profilers were started, and rethrows `measureError` first when both are set — which is the right priority. Profiles are still dropped on a failing iteration (the re-throw happens before `appendResults`), but that's now an explicit choice in clean code rather than the old swallowed-error path. ✅
- **No unit tests for the pure helpers** — still no tests added in this commit. Not a blocker, but the helpers (`parseScriptProfile`, `parseSelectorTrace`, `mergeScriptProfiles`, `mergeSelectorProfiles`, `escapeTableCell`) remain a good candidate for fixture-driven tests; the merge-math change in this commit in particular would have been an obvious target.
- **Two env-var aliases (`PERF_SELECTOR_PROFILE` / `PERF_CSS_SELECTOR_PROFILE`)** — unchanged in `createPerfMeasure` (perf.ts ~line 716). Worth a follow-up to drop one before the second name shows up in CI scripts and becomes load-bearing.
- **Selector trace categories include unread ones** — `SELECTOR_TRACE_CATEGORIES` is unchanged; `parseSelectorTrace` still only consumes `SelectorStats` events (emitted from `disabled-by-default-blink.debug`). Worth confirming whether `disabled-by-default-devtools.timeline.invalidationTracking` is required for `selector_timings[].invalidation_count` to be populated; if not, dropping the unused categories cuts measurement-time CDP traffic.
- **Feature still not enabled in CI** — `.github/workflows/app.yml` does not set any of the `PERF_*_PROFILE` vars and no `perf-chrome.ts` opts in. Still local-only as wired. Fine if intentional.

### New issues introduced by this commit

None.

### Not validated

- End-to-end perf workflow run with `PERF_SCRIPT_PROFILE=true` / `PERF_SELECTOR_PROFILE=true`. The structural changes are sound, but actual data quality on a real perf-chrome.ts test wasn't exercised in this review.

PR #5869, opened 2026-04-29
Floor layer math at l=0.13 on very dark parents

diegohaz:
## Motivation

Layer/state utilities that derive lightness via additive deltas (`ak-layer-N`, `ak-layer-push-*`, `ak-layer-lighten-*`, `ak-state-*`, `ak-state-push-*`, and edges that route through `getLayerL`) were producing near-invisible shifts when the parent's OKLCH lightness was below ~0.13. For example, `ak-layer-5` on `#000` (l=0) lands at l=0.05 — barely distinguishable from black on real displays — while the same utility on `#0C0E12` (l≈0.13) lands at l=0.18 and is clearly visible. OKLCH is supposed to be perceptually uniform, but display gamma compression near black, ambient-light black floor, and visual adaptation to brighter screen areas all squash near-black deltas in practice.

## Solution

Inside `getLayerL`, floor the `l` channel used as the additive base to `LAYER_L_FLOOR = 0.13`. The parent's actual rendered color is untouched — only the math input for descendants is lifted — so a `#000` background still renders as `#000` while its layer/state descendants land far enough above the parent to be visible.

The floor is gated on a strictly positive relative shift (`fn.binary(relativeLightness)`) so that:

- `ak-layer-0` (rel=0) preserves the parent's lightness instead of lifting to the floor.
- `ak-layer-l-*` is unaffected: the explicit absolute lightness wins in the first stage, and the downstream `layerIdleAuto`/`layerAuto` stages that re-invoke `getLayerL` with rel=0 no longer lift either.
- `ak-layer-darken-*` keeps going darker (or clamps to 0 as before) instead of being lifted by the floor and producing a result lighter than the parent.

The fix lives entirely inside `getLayerL`, so every utility that already routed through it picks up the new behavior without per-utility plumbing. The near-black boost is also factored into a registered `layerLFloorBoost` custom property so the repeated color stages reuse the same floor delta instead of inlining the `max(0, 0.13 - l)` expression at each call site.

## Verification

- Added a "very dark parent (l < 0.13)" sandbox section that renders the full `<Layers />` tree inside sub-floor parents (`ak-layer-black`, `oklch(0.05 0 0)`, and a chromatic `oklch(0.08 0.02 260)`) so the lift is visually checkable side by side.
- All 6 chrome tests pass: 4 computed-style snapshots (light, dark, light-high-contrast, dark-high-contrast) and both WCAG AA contrast checks.
- Local perf comparison after factoring the floor boost shows style recalculation improving by ~2% on page load and ~4% while toggling layer/text classes.


packages/ariakit-tailwind/src/input.ts
```diff
@@ -21,6 +21,11 @@ const CHROMA_MAX_REC2020 = 0.467;
 const CHROMA_MAX = CHROMA_MAX_P3;
 
 const DARK_THRESHOLD_L = 0.645;
+// Below this OKLCH lightness, additive deltas become hard to see on real
+// displays (gamma compression and ambient-light black floor squash near-black
+// differences). Layer math floors `l` to this value so descendants of very dark
+// parents remain visible without altering the parent's rendered color.
+const LAYER_L_FLOOR = 0.13;
```
diegohaz, line 28: This changes the shipped `@ariakit/tailwind` output for layer/state utilities, but the PR still has no changeset. Since consumers will see different colors for very dark layer parents, please add a patch changeset for `@ariakit/tailwind` so the fix is released and documented. The `site` workspace is ignored, so it does not need release frontmatter.

site/src/sandbox/ariakit-tailwind/index.react.tsx
```diff
@@ -481,6 +481,22 @@ export default function Example() {
           </Layer>
         </Layer>
       </Layer>
+      <Layer
+        label="very dark parent (l < 0.13)"
+        className="ak-layer ak-frame ak-frame-[1.25rem]/1 ak-frame-border flex-col"
+      >
+        <Layer className="flex-wrap items-start">
+          <Layer className="ak-layer ak-layer-black ak-frame ak-frame-p-1 ak-frame-border flex-col">
```
diegohaz, line 489: This adds three full `<Layers />` trees to the same sandbox that `perf-chrome.ts` uses as its benchmark. That benchmark measures page load and then toggles every `section[aria-labelledby]`, so this visual coverage changes the workload at the same time as the implementation under test; the perf bot is already reporting ~27-31% regressions here. Please keep the perf fixture stable, for example by making this sub-floor case a compact targeted sample instead of repeating `<Layers />`, or by scoping `perf-chrome.ts` to a stable set of sections and excluding visual-only coverage. Otherwise we cannot tell whether the floor math itself regressed.

diegohaz (commented):
Initial review of PR 5869 at `fdb11007b00414d6e44543a1cd5c1a57c186a796`, using `e982dec061dba8336786abd71b11ba3986b4839a` as the base. I reviewed the Tailwind lightness-floor implementation, generated `output.css`, the sandbox/snapshot changes, the resolved CodeRabbit thread, and the CI/performance comments.

I left two inline findings: one for the missing `@ariakit/tailwind` changeset, and one for the new sandbox section changing the `ariakit-tailwind` performance benchmark workload. The latter makes the current perf regression signal hard to interpret because the benchmark page grew at the same time as the CSS math changed.

Validation run locally:
- `pnpm install`
- `pnpm -F @ariakit/tailwind build` (regenerated `output.css` with no diff)

I also attempted `pnpm -F @ariakit/tailwind lint`, but local linting failed before checking files because `oxlint` rejected the repo config with `options.typeAware` outside the root config. I did not rerun the browser snapshot/WCAG/perf suites locally; CI currently reports those functional checks passing, with the separate perf bot comment showing the benchmark regression.

packages/ariakit-tailwind/src/input.ts
diegohaz, line 28: This PR ships a perceptible behavior change to `@ariakit/tailwind` consumers: any layer/state utility built on top of a parent whose OKLCH lightness is below `0.13` now lifts to a brighter base (e.g., `ak-layer-10` on `oklch(0 0 0)` goes from `oklch(0.10 …)` to `oklch(0.23 …)`). The `light`/`light-high-contrast`/`dark-high-contrast` snapshot diffs already show this affecting existing test fixtures with `[#0C0E12]` and pure-black parents, not just the new sandbox section.

That means it warrants a `.changeset/*.md` entry so the next `0.1.x` release picks it up and the CHANGELOG records the new floor behavior — the recent versions (0.1.10, 0.1.11) all document layer/edge/frame changes the same way, and this is a comparable consumer-visible adjustment.

Something along these lines:

```md
---
"@ariakit/tailwind": patch
---

Layer math now floors the additive base lightness to `0.13` for descendants of very dark parents (e.g. `ak-layer-black` or any `oklch(...)` with `l < 0.13`).

This only affects the lightness used to derive descendant layers, edges, and state colors — the parent's rendered color is untouched. Utilities that already produce a darker or equal lightness (`ak-layer-darken-*`, `ak-layer-0`, explicit `ak-layer-l-*` without an additional layer/state shift) are unchanged.
```

File name something like `.changeset/300-tailwind-layer-l-floor.md` per the changeset skill (300 = patch / bug-fix bucket; v0 patch frontmatter).

diegohaz (commented):
## Review summary

Initial review. Baseline: `e982dec06` (origin/main). Head: `fdb11007b`.

### Scope inspected beyond the diff

- `getLayerL` and the new `LAYER_L_FLOOR` / `layerLFloorBoost` (`--_ak-llfb`) wiring in `packages/ariakit-tailwind/src/input.ts`.
- All call sites of `getLayerL` (idle/state/auto/edge/push at lines 1006, 1012, 1034, 1074, 1171, 1353, 1367) to confirm the `binary(rel) * boost` gating produces the documented behavior in each stage.
- The regenerated `packages/ariakit-tailwind/src/output.css` — verified that the boost term lands inside an `oklch(from …)` context everywhere it is referenced, so `var(--_ak-llfb)`'s `l` channel resolves against the correct stage's color.
- `site/src/sandbox/ariakit-tailwind/index.react.tsx` new "very dark parent (l < 0.13)" section, plus the CodeRabbit thread that flagged the original `#0C0E12` (l ≈ 0.0933) sample — already addressed in the latest commit (`ak-layer-black`, `oklch(0.05 0 0)`, `oklch(0.08 0.02 260)` are all genuinely sub-floor).
- All four `__snapshots__/*` files. Spot-checked that the `oklch(0.10 …)` → `oklch(0.23 …)` jumps line up with the math `l + relative + max(0, 0.13 - l)`, that pre-existing dark-mode tests had no unintended drift (only additions), and that the deletions in `light`/`light-high-contrast`/`dark-high-contrast` are confined to fixtures whose parent `l` is below `0.13` (`#0C0E12`, `oklch(0 …)` etc.).
- `.changeset/` — empty, see the inline comment.

### Validation

- `pnpm -F @ariakit/tailwind build` — clean, `output.css` already in sync with `input.ts`.
- `pnpm -F site test-chrome ariakit-tailwind` — all 6 tests pass (light/dark × normal/high-contrast snapshots + both WCAG AA checks). Reproduces the PR description's verification.
- Manually traced the algebra for combined utilities not in the sandbox: `ak-layer-darken-*` (gating returns 0, no lift), `ak-layer-lighten-*` with positive shift (lift fires once at `--_ak-lib`), `ak-layer-N` on a parent with `l ≥ 0.13` (boost is `max(0, 0.13 - l) = 0`, output unchanged), and stacked stages where an earlier stage already lifted to ≥ 0.13 (downstream stages compute boost = 0, so no double-application).
- I also tried to run `pnpm -F site test-perf ariakit-tailwind` for a baseline-vs-current comparison but the second worker run fought the local Cloudflare preview server, so I did not collect a clean delta. The snapshot tests already exercise the new math, and the refactor that hoists the boost into a single `@property` is the right shape for the perf claim, so I did not retry.

### Findings

1. Missing changeset for a consumer-visible color change. See the inline comment on `packages/ariakit-tailwind/src/input.ts`.

No other actionable findings. The fix is correctly scoped to `getLayerL`, the per-call gating with `fn.binary(relativeLightness)` matches the documented intent for `ak-layer-0` / `ak-layer-l-*` / `ak-layer-darken-*`, and the `--_ak-llfb` `@property` extraction is consistent with the existing pattern in `constantMathVars`.

PR #5877, opened 2026-05-01
Use raw Tailwind arbitrary values

diegohaz:
## Motivation

`@ariakit/tailwind` supports arbitrary utility values, but numeric arbitrary values were being scaled like bare tokens. That made raw Tailwind-style values awkward for relative color math, requiring expressions such as `ak-layer-[calc(1*100+10)]` instead of direct CSS-channel math like `ak-layer-[calc(l+0.1)]`.

The v0.2 docs and sandbox should describe the final behavior before this package version ships, without adding migration noise for behavior that has not been published.

## Solution

Bare numeric modifiers still use Ariakit's documented authoring scales, such as `0`–`100` for lightness/alpha and `0`–`40` for chroma. Arbitrary values and custom-property arbitrary forms now pass through as raw CSS values, so consumers can write normalized OKLCH channels, raw deltas, and CSS percentages directly.

This updates the generated Tailwind output, README value-scale docs, existing v0.2 migration examples that mention custom arbitrary values, sandbox one-offs, and computed-style snapshots. It also removes the package-local `lint` script from `@ariakit/tailwind` because linting should run from the repository root.

## Validation

- `pnpm build-site-lite`
- `pnpm lint`
- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag -u`
- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag`

### DIFF
--- a/packages/ariakit-tailwind/migrating-to-v02.md
+++ b/packages/ariakit-tailwind/migrating-to-v02.md
@@ -6,7 +6,7 @@ This guide covers all breaking changes for users of the `@ariakit/tailwind` plug
 
 1. **Explicit base classes.** Utilities like `ak-layer`, `ak-frame`, `ak-text`, and `ak-outline` must now appear as base classes before their modifiers.
 2. **Slash to hyphen.** Opacity/value syntax changes from `/N` to `-N`. Text opacity moves from `ak-text/80` to `ak-ink-80`.
-3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints like `(color:--var)` or `(number:--var)`, but explicit longhands can use custom properties directly, such as `ak-layer-color-(--var)` and `ak-layer-offset-(--var)`. Custom-property values are raw CSS values, so use normalized values like `0.2` instead of scaled utility tokens like `20`.
+3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints when a shorthand accepts multiple value kinds, such as `ak-layer-(color:--var)` for colors or `ak-layer-(number:--var)` for lightness offsets. For `ak-edge`, `ak-text`, and `ak-outline`, use explicit longhands for custom properties, such as `ak-edge-alpha-(--alpha)`, `ak-text-push-(--value)`, and `ak-outline-push-(--value)`. Custom-property values are raw CSS values, so use normalized values like `0.2` instead of scaled utility tokens like `20`.
 4. **Unified `--ak-edge`.** The CSS variables `--ak-layer-border`, `--ak-border`, `--ak-ring`, and `--ak-layer-ring` are all replaced by `--ak-edge`.
 
 ---
--- a/packages/ariakit-tailwind/migrating-to-v02.md
+++ b/packages/ariakit-tailwind/migrating-to-v02.md
@@ -6,7 +6,7 @@ This guide covers all breaking changes for users of the `@ariakit/tailwind` plug
 
 1. **Explicit base classes.** Utilities like `ak-layer`, `ak-frame`, `ak-text`, and `ak-outline` must now appear as base classes before their modifiers.
 2. **Slash to hyphen.** Opacity/value syntax changes from `/N` to `-N`. Text opacity moves from `ak-text/80` to `ak-ink-80`.
-3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints like `(color:--var)` or `(number:--var)`, but explicit longhands can use custom properties directly, such as `ak-layer-color-(--var)` and `ak-layer-offset-(--var)`.
+3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints like `(color:--var)` or `(number:--var)`, but explicit longhands can use custom properties directly, such as `ak-layer-color-(--var)` and `ak-layer-offset-(--var)`. Custom-property values are raw CSS values, so use normalized values like `0.2` instead of scaled utility tokens like `20`.
 4. **Unified `--ak-edge`.** The CSS variables `--ak-layer-border`, `--ak-border`, `--ak-ring`, and `--ak-layer-ring` are all replaced by `--ak-edge`.
 
 ---
@@ -129,6 +129,8 @@ Custom pop values can use the explicit offset longhand.
 + ak-layer ak-layer-offset-(--depth)
 ```
 
+The `--depth` value should be a raw normalized offset such as `0.2`, not `20`.
+
 ### Variant prefixes on modifier classes
 
 When `ak-layer` is already on the element unconditionally, variant-prefixed modifiers work without re-applying `ak-layer`:
@@ -184,6 +186,8 @@ For custom properties on the mix utility itself, pair `ak-layer-mix` with explic
 - ak-layer-mix-(color:--my-mix-color) ak-layer-mix-(number:--amount)
 + ak-layer-mix ak-layer-mix-color-(--my-mix-color) ak-layer-mix-amount-(--amount)
 ```
+
+The `--amount` value should be a raw CSS percentage such as `15%`.
 
 ---
 
@@ -248,7 +252,7 @@ Custom color text can use the explicit color longhand.
 + ak-text ak-text-color-(--my-color)
 ```
 
-For custom push values, use `ak-text-push-(--value)` instead of `ak-text-(number:--value)`.
+For custom push values, use `ak-text-push-(--value)` instead of `ak-text-(number:--value)`. The `--value` custom property should contain a raw normalized value such as `0.25`.
 
 ---
 
@@ -279,7 +283,7 @@ In v0.1, `ak-edge-contrast-<color>` set the edge to a solid color with lightness
 
 `ak-edge-raw` is shorthand for `ak-edge-100 ak-edge-push-0` — use it when you want the color to be applied exactly as specified. For other combinations, `ak-edge-N` controls alpha (`100` = opaque) and `ak-edge-push-N` controls how far the edge lightness is pushed away from the layer (`0` = the color's natural lightness).
 
-For custom alpha values, use `ak-edge-alpha-(--alpha)` instead of `ak-edge-(number:--alpha)`.
+For custom alpha values, use `ak-edge-alpha-(--alpha)` instead of `ak-edge-(number:--alpha)`. The `--alpha` custom property should contain a raw alpha value such as `0.4`.
 
 ### CSS variable renames (color)
 
--- a/packages/ariakit-tailwind/readme.md
+++ b/packages/ariakit-tailwind/readme.md
@@ -66,6 +66,12 @@ Ariakit Tailwind revolves around a few families of utilities:
 
 All color math uses [OKLCH](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch), so modifiers like `ak-layer-warm-40` or `ak-text-saturate-50` behave predictably across hues. Because every value is computed relatively, changing a single theme token ripples through every layer, text, edge, and frame that depends on it — so users can reskin the whole system without breaking contrast, depth, or shape relationships.
 
+### Value scales
+
+Bare numeric modifiers use Ariakit's documented authoring scales. For example, lightness and alpha values use `0`–`100`, chroma values use `0`–`40`, and mix amounts use `0`–`100`.
+
+Arbitrary values and custom properties are raw CSS values. Use normalized OKLCH channel values like `ak-layer-l-[0.8]`, raw deltas like `ak-layer-[calc(l+0.1)]`, percentages where CSS expects percentages like `ak-layer-mix-amount-[35%]`, and custom properties that already contain those raw values.
+
 ## Theming
 
 Ariakit Tailwind integrates directly with Tailwind's theming system. Every `--color-*`, `--radius-*`, and `--spacing-*` token in your `@theme` block becomes available to the matching Ariakit utilities.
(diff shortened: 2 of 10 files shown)


diegohaz:
@codex

packages/ariakit-tailwind/src/input.ts
```diff
@@ -175,10 +175,39 @@ function getNumericTokenValue(pattern: string, options?: NumbersOptions) {
...
+ * Parses a non-arbitrary numeric utility token as a 0..1 percentage.
+ */
+function getBarePercentTokenValue(options?: NumbersOptions) {
+  return fn.div(getBareNumericTokenValue(options), 100);
+}
+
+/**
+ * Parses an arbitrary utility value without scaling it.
+ */
+function getRawArbitraryValue(pattern = "[*]") {
```
diegohaz: This introduces and documents a shipped `@ariakit/tailwind` behavior change, but the PR does not add a `.changeset` entry. I confirmed locally that `pnpm exec changeset status --since=origin/main` fails with `Some packages have been changed but no changesets were found`, so merging this would not produce a package version bump or changelog entry for the raw arbitrary/custom-property value behavior. Please add a patch changeset for `@ariakit/tailwind` that describes this behavior change, or make the release-neutral decision explicit with an empty changeset if that is intentional.

diegohaz (commented):
> **Note:** This summary was posted before a session boundary interrupted the review. The full consolidated summary covering all three findings (missing changeset, `(number:--var)` doc scope, `ak-layer-mix-amount-*` arbitrary filter) is in [#pullrequestreview-4212030415](https://github.com/ariakit/ariakit/pull/5877#pullrequestreview-4212030415). Refer to that one — this body is kept only because GitHub does not allow deleting submitted reviews.

---

Initial review of PR #5877 at `d62304c96539fb2e31d5924847f037687e28c066`, diffed against `origin/main` / `c0399c03fdcbfc7d04429d72b5217b151961a7c1`. I left one inline finding for the missing changeset.

packages/ariakit-tailwind/migrating-to-v02.md
```diff
@@ -6,7 +6,7 @@ This guide covers all breaking changes for users of the `@ariakit/tailwind` plug
 
 1. **Explicit base classes.** Utilities like `ak-layer`, `ak-frame`, `ak-text`, and `ak-outline` must now appear as base classes before their modifiers.
 2. **Slash to hyphen.** Opacity/value syntax changes from `/N` to `-N`. Text opacity moves from `ak-text/80` to `ak-ink-80`.
-3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints like `(color:--var)` or `(number:--var)`, but explicit longhands can use custom properties directly, such as `ak-layer-color-(--var)` and `ak-layer-offset-(--var)`.
+3. **Explicit longhands for custom properties.** Overloaded shorthand utilities still need typed arbitrary value hints like `(color:--var)` or `(number:--var)`, but explicit longhands can use custom properties directly, such as `ak-layer-color-(--var)` and `ak-layer-offset-(--var)`. Custom-property values are raw CSS values, so use normalized values like `0.2` instead of scaled utility tokens like `20`.
```
diegohaz: This bullet still presents `(color:--var)` and `(number:--var)` as the universal pattern for typed arbitrary value hints on overloaded shorthands, but the matching `[number]` filter on `--value()` has been removed from `ak-edge-*`, `ak-text-*`, and `ak-outline-*` in this PR. After this change `ak-edge-(number:--alpha)`, `ak-text-(number:--value)`, and `ak-outline-(number:--w)` no longer match any branch in `output.css`, so Tailwind drops the declaration silently and the rule has no effect at runtime — there is no error, the class just does nothing.

`(number:--var)` only still works on `ak-layer-*` and `ak-state-*` (where it is needed to disambiguate from `[color]` / `[percentage]`). The bullet should make that asymmetry explicit, otherwise readers following the migration will copy the `(number:--var)` pattern into `ak-edge-*` / `ak-text-*` / `ak-outline-*` calls and then have to debug a no-op class.

The specific replacement sentences further down the doc (e.g. "use `ak-edge-alpha-(--alpha)` instead of `ak-edge-(number:--alpha)`") are correct — only this top-level rule needs scoping. Suggest narrowing it to mention which shorthands accept which typed hints, and pointing readers at the longhands for the rest.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1297,36 +1339,36 @@ utility("layer-mix", getLayerMixDeclaration());
...
   getLayerMixDeclaration(),
 );
 
 utility("layer-mix-color-*", set(inputs.layerMixColor, fn.value(color, "[*]")));
 
 utility(
   "layer-mix-amount-*",
-  set(inputs.layerMixAmount, fn.toPercent(getNumericTokenValue("[*]"))),
+  set(inputs.layerMixAmount, fn.toPercent(getBareNumericTokenValue())),
+  set(inputs.layerMixAmount, getRawArbitraryValue()),
```
diegohaz: `getRawArbitraryValue()` defaults to the `[*]` filter, so this accepts anything as the arbitrary value for `ak-layer-mix-amount-*`. That means `ak-layer-mix-amount-[0.35]` (or any other non-percentage value) is happily compiled by Tailwind and the resulting `--ak-layer-mix-amount: 0.35` flows into `color-mix(in oklch, … 0.35, …)`, which is invalid — `color-mix` requires a `<percentage>` per CSS spec, so the entire `color-mix()` evaluates to nothing and the layer falls back unexpectedly at runtime. The migration doc explicitly tells users to use `35%` for this utility, but nothing in the build prevents the broken form.

For comparison, the shorthand `ak-layer-mix-*` on line 1345 uses `getRawArbitraryValue("[percentage]")` to enforce this. Suggest the same here:

```suggestion
  set(inputs.layerMixAmount, getRawArbitraryValue("[percentage]")),
```

With `[percentage]`, `ak-layer-mix-amount-[0.35]` fails to match any utility branch and Tailwind drops the declaration — which surfaces the mistake at author time instead of producing broken runtime CSS. The `(--var)` form keeps working through the existing `ak-layer-mix-amount-(--var)` longhand path.

diegohaz (commented):
## Review summary (consolidated)

Initial review of #5877 ("Use raw Tailwind arbitrary values") at `d62304c`. Baseline: merge base with `origin/main` (`c0399c03`). This run was interrupted by a session boundary mid-review — an earlier summary review (#pullrequestreview-4212010479) covered only the first finding; this body now consolidates all three findings.

### Scope

- All 7 changed files: `packages/ariakit-tailwind/{src/input.ts, src/output.css, migrating-to-v02.md, readme.md, package.json}` and `site/src/sandbox/ariakit-tailwind/{index.react.tsx, __snapshots__/*}`.
- Surrounding context for the dynamic utilities affected by the bare-vs-arbitrary split (`ak-layer-*`, `ak-state-*`, `ak-edge-*`, `ak-text-*`, `ak-outline-*` and their `*-amount-*`, `*-mix-*`, `*-l-*`, `*-c-*`, `*-cool-*`, `*-warm-*` modifiers).
- The new helpers (`getBareNumericTokenValue`, `getBarePercentTokenValue`, `getRawArbitraryValue`, `getRawPercentDeclarations`) and the refactored `withUtilityTokenGate` signature.
- Release metadata (changesets, package.json scripts).

### Validations performed locally

- `pnpm install` and `pnpm -F @ariakit/tailwind build` — succeeded; `output.css` matches the `input.ts` changes (each affected dynamic utility now emits a bare `--value(number, …) / 100` branch and a separate `--value([*])` / `--value([percentage])` branch).
- `pnpm -F site test-chrome ariakit-tailwind -g wcag` — 2/2 passed.
- `pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag` (computed-style snapshots) — 4/4 passed across light, dark, light-high-contrast, dark-high-contrast.
- Tailwind v4 compile probes for representative bare, arbitrary, and custom-property classes confirmed the raw/scaled split for the layer/text/edge/outline/mix utility families I checked.
- `git diff --check origin/main...HEAD` passed.
- `pnpm exec changeset status --since=origin/main` failed (no changeset present).
- Spot-checked Tailwind v4 `--value()` filter semantics: declarations whose `--value(...)` filter doesn't match cause the whole declaration to be dropped, so the asymmetry between utilities (which still accept `[number]` arbitraries vs which don't) is silent at compile time.

The split itself is sound — moving away from scaled-by-100 arbitrary values toward raw CSS values is the right direction (it's what makes `ak-layer-[calc(l+0.1)]` and `ak-layer-l-[0.62]` legible), and keeping the `0`–`100` bare-token convention preserves backward compatibility for class-name authors.

### Findings

Three substantive findings were left as inline comments:

1. **Missing changeset** (`packages/ariakit-tailwind/src/input.ts:194`, the new `getRawArbitraryValue` helper) — the PR introduces and documents a shipped behavior change to `@ariakit/tailwind`'s arbitrary-value handling, but no `.changeset` entry exists; `changeset status --since=origin/main` fails. Add a patch changeset (or an explicit empty changeset if release-neutral is intentional).

2. **Migration doc bullet 3 overstates `(number:--var)`** (`migrating-to-v02.md:9`) — the bullet still presents `(number:--var)` as a universal hint, but after this PR the `[number]` branch is gone from `ak-edge-*` / `ak-text-*` / `ak-outline-*`, so `ak-edge-(number:--alpha)` etc. silently produce no-op classes. Scope the bullet to where it still works (`ak-layer-*`, `ak-state-*`) and point readers at the new longhands for the rest.

3. **`ak-layer-mix-amount-*` accepts non-percentages** (`packages/ariakit-tailwind/src/input.ts:1354`) — the arbitrary branch uses the default `[*]` filter, so `ak-layer-mix-amount-[0.35]` compiles successfully and produces an invalid `color-mix(in oklch, … 0.35, …)` at runtime. The shorthand `ak-layer-mix-*` already enforces `[percentage]`; the same constraint should apply here so the mistake surfaces at author time.

### Areas not exhaustively validated

- Did not run a perf comparison vs. `main`. The change roughly doubles the body of several dynamic utilities (one declaration per arbitrary path now becomes two), so consumer CSS that uses many bare-numeric variants of these utilities could be slightly larger. Worth a quick `pnpm -F site test-perf ariakit-tailwind` baseline/current comparison if perf is a concern for v0.2.
- Did not test the new sandbox cells in a non-headless browser; relied on the snapshot test for visual correctness.
- Did not duplicate the existing Copilot documentation comments.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -175,10 +175,39 @@ function getNumericTokenValue(pattern: string, options?: NumbersOptions) {
...
+function getBarePercentTokenValue(options?: NumbersOptions) {
+  return fn.div(getBareNumericTokenValue(options), 100);
+}
+
+/**
+ * Parses an arbitrary utility value without scaling it.
+ */
+function getRawArbitraryValue(pattern = "[*]") {
+  return fn.value(pattern);
+}
```
diegohaz: I believe this function is unnecessary. Just use `fn.value("[*]")` directly.

packages/ariakit-tailwind/readme.md
```diff
@@ -66,6 +66,12 @@ Ariakit Tailwind revolves around a few families of utilities:
 
 All color math uses [OKLCH](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch), so modifiers like `ak-layer-warm-40` or `ak-text-saturate-50` behave predictably across hues. Because every value is computed relatively, changing a single theme token ripples through every layer, text, edge, and frame that depends on it — so users can reskin the whole system without breaking contrast, depth, or shape relationships.
 
+### Value scales
+
+Bare numeric modifiers use Ariakit's documented authoring scales. For example, lightness and alpha values use `0`–`100`, chroma values use `0`–`40`, and mix amounts use `0`–`100`.
+
+Arbitrary values and custom properties are raw CSS values. Use normalized OKLCH channel values like `ak-layer-l-[0.8]`, raw deltas like `ak-layer-[calc(l+0.1)]`, percentages where CSS expects percentages like `ak-layer-mix-amount-[35%]`, and custom properties that already contain those raw values.
```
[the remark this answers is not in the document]
diegohaz: Fixed in 329c673cb0fe788512ed018a1514ae8d3c7e3fda - expanded the value-scale note with normalized percent-style examples for raw arbitrary/custom-property values.

packages/ariakit-tailwind/readme.md
```diff
@@ -150,31 +156,31 @@ Use [`ak-edge`](#ak-edge) to fine-tune border, ring, and shadow colors without t
...
-| `ak-layer-min-l-<value>`     | Floors lightness specifically. Custom properties use the same `0`–`100` scale.                                                                                               |
+| Utility                      | Description                                                                                                                                                            |
+| ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ak-layer-lighten-<number>`  | Lightens the layer by `<number>`%. Arbitrary values are raw (`ak-layer-lighten-[0.05]`).                                                                               |
+| `ak-layer-darken-<number>`   | Darkens the layer by `<number>`%.                                                                                                                                      |
```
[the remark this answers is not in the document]
diegohaz: Fixed in 329c673cb0fe788512ed018a1514ae8d3c7e3fda - added the same raw arbitrary-value clarification to `ak-layer-darken-<number>`.

packages/ariakit-tailwind/migrating-to-v02.md
diegohaz: Fixed in 329c673cb0fe788512ed018a1514ae8d3c7e3fda - narrowed the top-level migration guidance so typed shorthand custom-property hints are only presented for overloaded shorthands, and pointed `ak-edge`, `ak-text`, and `ak-outline` users to explicit longhands.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 329c673cb0fe788512ed018a1514ae8d3c7e3fda - removed `getRawArbitraryValue()` and replaced its call sites with direct `fn.value(...)` calls.

diegohaz:
**Fixed**

- 329c673cb0fe788512ed018a1514ae8d3c7e3fda ([reply](https://github.com/ariakit/ariakit/pull/5877#discussion_r3174058028)): Expanded the README value-scale note with normalized percent-style examples for raw arbitrary/custom-property values.
- 329c673cb0fe788512ed018a1514ae8d3c7e3fda ([reply](https://github.com/ariakit/ariakit/pull/5877#discussion_r3174058337)): Added matching raw arbitrary-value guidance to `ak-layer-darken-<number>`.
- 329c673cb0fe788512ed018a1514ae8d3c7e3fda ([reply](https://github.com/ariakit/ariakit/pull/5877#discussion_r3174058676)): Scoped the migration guide's typed custom-property guidance to overloaded shorthands and pointed `ak-edge`, `ak-text`, and `ak-outline` users to explicit longhands.
- 329c673cb0fe788512ed018a1514ae8d3c7e3fda ([reply](https://github.com/ariakit/ariakit/pull/5877#discussion_r3174058975)): Removed `getRawArbitraryValue()` and replaced its call sites with direct `fn.value(...)` calls.

**Validation**

- `pnpm lint`
- `git diff --check`

diegohaz (commented):
## Re-review summary

Incremental re-review of #5877 at `329c673c`, baseline `d62304c` (the head SHA recorded by the prior review). The new commit "Address raw arbitrary value review comments" touches 3 files: `migrating-to-v02.md`, `readme.md`, and `src/input.ts`. `output.css` regenerates cleanly with no diff.

### Disposition of prior findings

1. **Missing changeset** — withdrawn by the author ("no need for changeset here"). Out of scope.
2. **Migration doc bullet 3 overstated `(number:--var)`** — fixed in `329c673c`. Bullet 3 now scopes typed shorthand hints to overloaded shorthands (`ak-layer-(color:--var)`, `ak-layer-(number:--var)`) and points readers at explicit longhands for `ak-edge` / `ak-text` / `ak-outline` (`ak-edge-alpha-(--alpha)`, `ak-text-push-(--value)`, `ak-outline-push-(--value)`). I verified against `output.css` that `ak-layer-*` is the only shorthand that still emits a `--value([number])` branch, so the new wording is accurate.
3. **`ak-layer-mix-amount-*` accepts non-percentages (`[*]` filter)** — left unchanged in `329c673c` (helper inlined to `fn.value("[*]")`, behavior identical). Thread was resolved without a code change. Treating that as the author's deliberate decision and not re-raising.

### Validations on the new head

- `pnpm -F @ariakit/tailwind build` — passed; `output.css` already in sync with `input.ts`.
- `pnpm -F site test-chrome ariakit-tailwind` — 6/6 passed (WCAG AA + computed-style snapshots across light, dark, light-high-contrast, dark-high-contrast).

### New findings

None. The incremental commit is a clean refactor (`getRawArbitraryValue` removed, all 14 call sites inlined to `fn.value(...)`) plus targeted doc fixes that address the prior review and the Copilot comments. The refactor preserves behavior and reduces indirection. Doc updates to `readme.md` (percent-style normalization note in "Value scales", missing raw-arbitrary mention on `ak-layer-darken-<number>`) read correctly against the implementation.

### Areas not validated this pass

- Did not re-run a perf comparison; the helper inlining does not change generated CSS, so no perf delta is expected.

PR #5883, opened 2026-05-01
Document Ariakit Tailwind v0.2 changeset and migration updates

diegohaz:
## Motivation

@ariakit/tailwind v0.2 rewrites the plugin around explicit base utilities, richer relative color controls, and a more detailed migration path from the v0.1 API. The release needs changelog copy that sets expectations for the breaking utility changes, and the README and migration guide need to avoid outdated claims about layer offsets, state colors, contrast, ink alpha, and frame behavior.

## Solution

Add a changeset announcing @ariakit/tailwind v0.2 as a minor release with a concise overview, before/after example, performance note, and links to the migration guide and README.

Update the README to describe the current API more accurately: bare ak-layer does not shift lightness, color channel utilities are authored in OKLCH while text contrast is converted to LCH at the final step, contrast pushes lightness away from the layer, ak-state-* still affects descendant text and edge colors through the resulting layer color, ak-ink-* requests a minimum alpha, and frame border/cover wording matches the implemented behavior.

Update the v0.2 migration guide to narrow the ak-frame-cover replacement notes to padding and border behavior and replace the unrelated ak-button example with a package-local frame modifier example.

## Testing

- pnpm exec oxfmt --check packages/ariakit-tailwind/readme.md packages/ariakit-tailwind/migrating-to-v02.md .changeset/100-ariakit-tailwind-v02.md
- pnpm exec changeset status --verbose

### DIFF
--- a/packages/ariakit-tailwind/migrating-to-v02.md
+++ b/packages/ariakit-tailwind/migrating-to-v02.md
@@ -390,9 +390,9 @@ The force modifier is now a separate utility.
 + ak-frame ak-frame-cover ak-frame-p-1
 ```
 
-### Frame modifiers require explicit `ak-frame`
+### Frame preset padding overrides require explicit `ak-frame`
 
-When using a frame modifier such as `ak-frame-field/1`, you must add an explicit `ak-frame` class so the frame context is initialized.
+In v0.1, frame preset padding overrides such as `ak-frame-field/1` worked without a separate base class because the preset utility also initialized frame setup. In v0.2, add `ak-frame` explicitly before the modifier so the frame context is initialized.
 
 ```diff
 - ak-frame-field/1
--- a/packages/ariakit-tailwind/migrating-to-v02.md
+++ b/packages/ariakit-tailwind/migrating-to-v02.md
@@ -380,7 +380,7 @@ The force modifier is now a separate utility.
 
 ### `ak-frame-overflow` → `ak-frame ak-frame-cover`
 
-`ak-frame-overflow` has been removed. Use `ak-frame-cover` instead — it now automatically handles all border/ring combinations, collapsing shared borders when both parent and child have them.
+`ak-frame-overflow` has been removed. Use `ak-frame-cover` instead — it now automatically handles padding and border combinations, collapsing shared borders when both parent and child have them.
 
 ```diff
 - ak-frame-overflow
@@ -390,13 +390,13 @@ The force modifier is now a separate utility.
 + ak-frame ak-frame-cover ak-frame-p-1
 ```
 
-### Compound utility + frame override requires explicit `ak-frame`
+### Frame modifiers require explicit `ak-frame`
 
-When overriding a compound utility's frame token (e.g., `ak-button` + `ak-frame-field/1`), you must add an explicit `ak-frame` class. Without it, the compound utility's inlined padding-block longhand isn't reset by the shorthand.
+When using a frame modifier such as `ak-frame-field/1`, you must add an explicit `ak-frame` class so the frame context is initialized.
 
 ```diff
-- ak-button ak-frame-field/1
-+ ak-button ak-frame ak-frame-field/1
+- ak-frame-field/1
++ ak-frame ak-frame-field/1
 ```
 
 ---
--- a/packages/ariakit-tailwind/readme.md
+++ b/packages/ariakit-tailwind/readme.md
@@ -250,9 +250,9 @@ Controls the opacity of text inside a layer — useful for secondary text, capti
 </div>
 ```
 
-| Utility           | Description                                                                                                                                                          |
-| ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ak-ink-<number>` | Requests a minimum text alpha (`0`–`100`). The rendered alpha is clamped to the minimum value that still meets WCAG AA for the current layer; `100` is fully opaque. |
+| Utility           | Description                                                                                                                                                                                                                               |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ak-ink-<number>` | Requests a text alpha (`0`–`100`). The rendered alpha is the larger of the requested value and the minimum that still meets WCAG AA contrast against the current layer, so values below that floor are clamped up; `100` is fully opaque. |
 
 ## `ak-text`
 
--- a/packages/ariakit-tailwind/readme.md
+++ b/packages/ariakit-tailwind/readme.md
@@ -57,14 +57,14 @@ Ariakit Tailwind is framework and library agnostic. It works with any frontend f
 
 Ariakit Tailwind revolves around a few families of utilities:
 
-- **[`ak-layer`](#ak-layer)** turns any element into a _layer_ — a surface with its own background, text, border, and shadow colors. Layers nest, and each nested layer shifts in lightness relative to its parent so stacked surfaces read correctly in both light and dark modes.
+- **[`ak-layer`](#ak-layer)** turns any element into a _layer_ — a surface with its own background, text, border, and shadow colors. Layers can nest, and lightness modifiers such as `ak-layer-lighten-*`, `ak-layer-darken-*`, and numeric `ak-layer-*` modifiers shift nested surfaces relative to their parent so stacked surfaces read correctly in both light and dark modes.
 - **[`ak-ink`](#ak-ink)** sets the text opacity for the layer's own text. Safe to apply on the same element as `ak-layer` or on a descendant.
 - **[`ak-text`](#ak-text)** colors inline text _inside_ a layer with automatic WCAG contrast. Must go on a descendant, not on the `ak-layer` element itself.
 - **[`ak-edge`](#ak-edge)** colors borders and rings, adapting opacity and contrast to the layer behind them.
 - **[`ak-outline`](#ak-outline)** colors outlines in the same adaptive way.
 - **[`ak-frame`](#ak-frame)** handles radius, padding, margin, borders, and concentric-radius layout.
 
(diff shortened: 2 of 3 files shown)


packages/ariakit-tailwind/readme.md
```diff
@@ -245,14 +245,14 @@ Controls the opacity of text inside a layer — useful for secondary text, capti
...
-  Layer with its own text at 70% opacity
+  Layer with its own text at least 70% opaque
   <p class="ak-ink-0">Nested text at minimum readable opacity</p>
 </div>
 ```
 
 | Utility           | Description                                                                                                                                                          |
 | ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ak-ink-<number>` | Sets text opacity as an **absolute** alpha value (`0`–`100`). `0` clamps to the minimum alpha that still meets WCAG AA for the current layer; `100` is fully opaque. |
+| `ak-ink-<number>` | Requests a minimum text alpha (`0`–`100`). The rendered alpha is clamped to the minimum value that still meets WCAG AA for the current layer; `100` is fully opaque. |
```
diegohaz: This row is hard to parse and can read as "rendered alpha = WCAG-minimum", which would contradict the `100 is fully opaque` aside. The actual implementation is `textAlpha = max(textMinimumAlpha, inputs.textA)` (`packages/ariakit-tailwind/src/input.ts:1246`) — the requested value is the lower bound the user wants, and it is then clamped *up* to the WCAG-AA minimum if it is below it.

"Clamped to X" usually means "set to X if outside the range", so as written a casual reader can land on the wrong mental model. Worth saying explicitly that the WCAG-AA minimum acts as a floor on top of the requested value — for example:

```suggestion
| `ak-ink-<number>` | Requests a text alpha (`0`–`100`). The rendered alpha is the larger of the requested value and the minimum that still meets WCAG AA contrast against the current layer, so values below that floor are clamped up; `100` is fully opaque. |
```

packages/ariakit-tailwind/migrating-to-v02.md
```diff
@@ -390,13 +390,13 @@ The force modifier is now a separate utility.
 + ak-frame ak-frame-cover ak-frame-p-1
 ```
 
-### Compound utility + frame override requires explicit `ak-frame`
+### Frame modifiers require explicit `ak-frame`
 
-When overriding a compound utility's frame token (e.g., `ak-button` + `ak-frame-field/1`), you must add an explicit `ak-frame` class. Without it, the compound utility's inlined padding-block longhand isn't reset by the shorthand.
+When using a frame modifier such as `ak-frame-field/1`, you must add an explicit `ak-frame` class so the frame context is initialized.
```
diegohaz: Dropping the `ak-button` example makes sense — `ak-button` is a `@ariakit/clava` utility, not part of `@ariakit/tailwind`, so the original wording was misleading in this guide.

But after the rewrite this subsection just restates the general principle already documented at the top of the Frame section ("### `ak-frame` is now required as an explicit base class", `migrating-to-v02.md:312-314`). Both the rule and the example collapse to the same point: `ak-frame-*` modifiers need a base `ak-frame`. As written, a reader following along section by section is told this twice with no new information the second time.

Two options:

- Drop the subsection entirely and rely on the general principle at the top.
- Keep the subsection but make it concrete about *what was different* in v0.1 — e.g. "In v0.1, `ak-frame-field/<padding>` worked without a base class because the preset implicitly applied frame setup; in v0.2 you must add `ak-frame` explicitly." That gives the section a reason to exist next to the general principle.

diegohaz (commented):
## Review summary

Initial review. Diff baseline: `acd3a6412...2824901dd` (single commit). The PR adds `.changeset/100-ariakit-tailwind-v02.md` and updates `packages/ariakit-tailwind/readme.md` and `packages/ariakit-tailwind/migrating-to-v02.md`.

### What I checked

- **Changeset format**: title + `**BREAKING**` paragraph + Before/After examples conform to `ariakit-general-changeset`. Filename prefix `100-` for major change ✓. Frontmatter `minor` for v0 major-level change ✓.
- **Before/After example consistency**: every translation in the changeset's example matches the rules in the migration guide:
  - `ak-layer-canvas` → `ak-layer ak-layer-canvas` (`migrating-to-v02.md:101-105`)
  - `ak-layer-pop` → `ak-layer ak-layer-6` (`migrating-to-v02.md:33-35`)
  - `ak-frame-field` → `ak-frame ak-frame-field/field` (`migrating-to-v02.md:320-322`)
  - `ak-edge/10` → `ak-edge-10` (`migrating-to-v02.md:263-265`)
  - `ak-text-primary/70` → `ak-text ak-text-primary` (`/70` is at or below the v0.1 default and dropped per `migrating-to-v02.md:226-234`)
- **Source-code claims** in the readme verified against `packages/ariakit-tailwind/src/input.ts`:
  - Bare `ak-layer` does not shift lightness — `inputs.layerLOffset` defaults to `{ initial: 0 }` (`input.ts:669`) ✓.
  - Color channels in OKLCH, text contrast through LCH at the final step — `getTextDirectional()` returns `fn.lch(textColorAdjusted, ...)` (`input.ts:1687-1690`) ✓.
  - `ak-state-*` shifts the state layer separately, descendants still see the resulting layer color — `vars.layerBase` is set to `layerState` (`input.ts:1181`) and `vars.layerTextL` derives from `vars.layer` (`input.ts:1126-1136`), so descendant text/edge container queries reflect the state-modified layer ✓.
  - `ak-frame-cover` collapses shared *borders* (not rings) — only `inputs.frameBorder` participates in `stretchInset`/`radiusInset`, `inputs.frameRing` is not factored in (`input.ts:2189-2190`) ✓. Note: the source comment at `input.ts:2188` still says "borders/rings" — this is stale relative to the implementation but is outside this PR's scope.
  - `ak-frame-bordering` chooses border vs. ring based on parent appearance × lightening/darkening — verified against `getFrameBorderingDarkLight()` (`input.ts:2223-2249`) ✓. The new wording is more accurate than the previous one, which omitted the parent-mode dimension.
  - `contrast-more` pushing lightness farther from the layer matches `getContrastL` and the contrast-driven scaling on text/edge/outline.
- **Markdown style**: each paragraph is on a single line, no auto-wrap; complies with `ariakit-general-markdown`.

### Findings (all inline)

- `ak-ink-<number>` table description is ambiguous and can be read as "rendered alpha = WCAG-minimum". Suggested clearer wording is left inline at `readme.md:255`.
- The new "Frame modifiers require explicit `ak-frame`" subsection in the migration guide duplicates the earlier general principle at `migrating-to-v02.md:312-314`. Comment with two options inline at `migrating-to-v02.md:395`.

### Not validated

- The "more than 90% faster" claim in the changeset — I did not run the perf comparison harness to verify this independently. Relying on the author's measurement; the harness in `site/src/test-utils/perf.ts` plus `pnpm -F site test-perf ariakit-tailwind` is the recommended way to back this up if needed.
- Did not run `pnpm exec oxfmt --check` or `pnpm exec changeset status` — the PR description says these were run by the author and the diff is markdown-only.

No correctness regressions found. The doc updates correctly reflect the v0.2 implementation.

packages/ariakit-tailwind/readme.md
diegohaz: Fixed in c3d57a75d - clarified that `ak-ink-*` renders the larger of the requested alpha and the WCAG AA minimum.

packages/ariakit-tailwind/migrating-to-v02.md
diegohaz: Fixed in c3d57a75d - updated the frame modifier section to explain the v0.1 implicit frame setup behavior instead of restating the general base-class rule.

diegohaz:
**Fixed**

- c3d57a75d ([reply](https://github.com/ariakit/ariakit/pull/5883#discussion_r3174829134)): Clarified that `ak-ink-*` renders the larger of the requested alpha and the WCAG AA minimum.
- c3d57a75d ([reply](https://github.com/ariakit/ariakit/pull/5883#discussion_r3174829136)): Updated the frame modifier migration note to explain the v0.1 implicit frame setup behavior.

diegohaz (commented):
## Re-review summary

Incremental baseline: `2824901dd...c3d57a75d` (one new commit, `Clarify Ariakit Tailwind migration docs`). Only `packages/ariakit-tailwind/readme.md` and `packages/ariakit-tailwind/migrating-to-v02.md` changed; `.changeset/100-ariakit-tailwind-v02.md` is unchanged from the previous review.

### Prior findings — verified against the new head, not just the resolution replies

- **`readme.md:255` — `ak-ink-<number>` wording**. The new row reads "Requests a text alpha (`0`–`100`). The rendered alpha is the larger of the requested value and the minimum that still meets WCAG AA contrast against the current layer, so values below that floor are clamped up; `100` is fully opaque." This matches `textAlpha = fn.max(textMinimumAlpha, inputs.textA)` (`packages/ariakit-tailwind/src/input.ts:1246`) and removes the earlier ambiguity. Resolved ✓.
- **`migrating-to-v02.md:393` — Frame preset padding section**. Renamed to "Frame preset padding overrides require explicit `ak-frame`" with concrete v0.1-vs-v0.2 contrast: "In v0.1, frame preset padding overrides such as `ak-frame-field/1` worked without a separate base class because the preset utility also initialized frame setup. In v0.2, add `ak-frame` explicitly before the modifier so the frame context is initialized." The section now has a reason to exist alongside the general principle at lines 312–314. Resolved ✓.

### Other checks on the new commit

- No new prose introduced beyond the two targeted edits (and the corresponding markdown table column re-padding in the `ak-ink` table).
- The new `ak-ink` and frame-preset wordings each fit on a single line, complying with `ariakit-general-markdown`.
- No surrounding code or other docs were touched, so nothing new needs validation.

No remaining findings.

PR #5902, opened 2026-05-04
Fix Menu autoFocusOnShow open behavior

### LINKED ISSUE #5899: Menu `autoFocusOnShow` prop does nothing
### Current behavior

Setting `autoFocusOnShow` to `false` does nothing, when clicking or interacting with the trigger the menu is focused anyway.

1. Open this Code Sandbox: https://codesandbox.io/p/devbox/k7pk8y
2. Interact with the trigger
3. The focus ring shows that the menu is focused

### Expected behavior

The focus should stay on the trigger.

### Workaround

See the updated workaround:
https://github.com/ariakit/ariakit/issues/5899#issuecomment-4371640185

### Possible solutions

_No response_

diegohaz:
## Motivation

Fixes #5899.

`Menu` should respect `autoFocusOnShow={false}` and `autoFocusOnShow={() => false}` when the menu is opened. The menu should open from click, Enter, Space, ArrowDown, and ArrowUp without moving focus away from the `MenuButton`.

The same menu should still be keyboard-accessible after it is already open: pressing ArrowDown or ArrowUp again from the focused `MenuButton` should move focus into the menu.

## Solution

Added a sandbox regression test in `site/src/sandbox/menu-5899` covering both the direct boolean and callback forms of `autoFocusOnShow={false}`.

`Menu` now treats an explicit `autoFocusOnShow={false}` as a hard opt-out instead of falling back to the store's open-time autofocus state. `MenuButton` also handles ArrowDown and ArrowUp directly when its menu is already open, moving focus to the first or last menu item without relying on open-time autofocus.

## Workaround

Until the fix is available, use the callback form to prevent open-time focus movement. If the menu button should still move focus into an already-open menu on ArrowDown or ArrowUp, intercept those keys only after the menu is open and call `store.move()` yourself.

```tsx
const menu = Ariakit.useMenuStore();

const focusItem = (id: string | null | undefined) => {
  if (!id) return;
  requestAnimationFrame(() => menu.move(id));
};

// TODO: Remove this workaround once https://github.com/ariakit/ariakit/issues/5899 is fixed.
<Ariakit.MenuButton
  onKeyDown={(event) => {
    const { open } = menu.getState();
    if (!open) return;

    if (event.key === "ArrowDown") {
      event.preventDefault();
      focusItem(menu.first());
      return;
    }

    if (event.key === "ArrowUp") {
      event.preventDefault();
      focusItem(menu.last());
    }
  }}
>
  Actions
</Ariakit.MenuButton>
<Ariakit.Menu autoFocusOnShow={() => false}>
  {/* ... */}
</Ariakit.Menu>
```

### DIFF
--- a/.changeset/300-menu-autofocus-on-show.md
+++ b/.changeset/300-menu-autofocus-on-show.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-core": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Menu`](https://ariakit.com/reference/menu) to respect `autoFocusOnShow={false}` and `autoFocusOnShow={() => false}` while still allowing arrow keys to move focus into an already-open menu.
--- a/.changeset/300-menu-autofocus-on-show.md
+++ b/.changeset/300-menu-autofocus-on-show.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Menu`](https://ariakit.com/reference/menu) to respect `autoFocusOnShow={false}` and `autoFocusOnShow={() => false}` while still allowing arrow keys to move focus into an already-open menu.
+Fixed [`Menu`](https://ariakit.com/reference/menu) to respect the [`autoFocusOnShow`](https://ariakit.com/reference/menu#autofocusonshow) prop when set to `false` or when a callback returns `false`, while still allowing arrow keys to move focus into an already-open menu.
--- a/packages/ariakit-react-core/src/menu/menu-button.tsx
+++ b/packages/ariakit-react-core/src/menu/menu-button.tsx
@@ -140,6 +140,12 @@ export const useMenuButton = createHook<TagName, MenuButtonOptions>(
       const initialFocus = getInitialFocus(event, dir);
       if (initialFocus) {
         event.preventDefault();
+        const { open } = store.getState();
+        if (open) {
+          const id = initialFocus === "last" ? store.last() : store.first();
+          store.move(id);
+          return;
+        }
         showMenu();
         store?.setAutoFocusOnShow(true);
         store?.setInitialFocus(initialFocus);
--- a/packages/ariakit-react-core/src/menu/menu.tsx
+++ b/packages/ariakit-react-core/src/menu/menu.tsx
@@ -122,7 +122,6 @@ export const useMenu = createHook<TagName, MenuOptions>(function useMenu({
     };
   }, [store, modal, autoFocusOnShowState, initialFocus, items, baseElement]);
 
-  const mayAutoFocusOnShow = !!autoFocusOnShow;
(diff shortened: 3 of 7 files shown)


diegohaz:
@codex

diegohaz (commented):
## Initial review of #5902

**Scope:** full PR diff against `main` (`184e8be6` → `39b16580`). Focused on the focus-pipeline behavior change in `menu.tsx`/`menu-button.tsx`, the new sandbox + browser tests, and impact on the broader menu surface.

**Validation performed locally**

- Traced the focus pipeline (`MenuButton.onClick`/`onKeyDown` → `useMenu` → `useHovercard` → `useDialog` autofocus effect) to confirm the root cause and the fix:
    - `menu.tsx`: the previous `autoFocusOnShowState || modal` fallback was triggered by `MenuButton.onClick`'s `store.setAutoFocusOnShow(true)`, which defeated the explicit `autoFocusOnShow={false}` prop on click. The new `autoFocusOnShow === false ? false : canAutoFocusOnShow && autoFocusOnShow` makes literal `false` a hard opt-out, so `dialog.tsx`'s `if (!mayAutoFocusOnShow) return;` short-circuits the autofocus effect.
    - `menu-button.tsx`: the new "already-open" branch (`store.move(store.first()/last())`) is necessary, not just an optimization. With `autoFocusOnShow={false}` or `() => false`, `setAutoFocusOnShow(true) + setInitialFocus(...)` cannot drive the dialog autofocus effect (the effect bails on `mayAutoFocusOnShow=false`, or the callback returns `false`), so without this branch ArrowUp/ArrowDown could not move focus into an already-open menu. Bypassing the autofocus mechanism via composite `move()` is the right fix.
- Browser tests at the current PR head: 45/45 `menu-5899` tests pass across chrome/firefox/safari (including the 39b16580 callback side-effect test that asserts `data-autofocus-on-show-callback` is set on the resolved element).
- Regression sweep across other menu sandboxes (`menu-4291`, `menu-nested-combobox-4443`, `menu-button-4133`, `menu-dialog-close`, `menu-4938`, `menu-nested-combobox-hide`): 30/30 tests pass — no regressions in submenu, menu-dialog, or menubar flows.
- `pnpm tsc`: 0 errors. `pnpm lint`: 0 errors (the 13 pre-existing `no-underscore-dangle` warnings are in unrelated files).

**Findings:** none.

The diff is small and well-scoped, and the strengthened callback test added in 39b16580 (mutating an attribute on the resolved element and asserting it in the test) addresses the CodeRabbit nit. Backward compat is fine: code that relied on `setAutoFocusOnShow(true)` overriding an explicit `autoFocusOnShow={false}` prop was depending on the bug.

**Areas not exhaustively validated:** the legacy `examples/` workspace tests (separate runner), full `pnpm test`, and visual regression. The sandbox suite under `site/src/sandbox/menu-*` covered the realistic behavior surface for this change.

**Verdict:** LGTM.

PR #5904, opened 2026-05-04
Fix Changesets release plan changelog

diegohaz:
## Motivation

The release workflow was failing during `pnpm run version` when Changesets generated changelog entries for packages with internal dependency updates. The repo patch for `@changesets/apply-release-plan` called `shouldUpdateDependencyBasedOnConfig` with the wrong argument shape, which left the config parameter undefined and crashed changelog generation.

## Solution

Update the patch to pass the expected `cwd`, release metadata, dependency metadata, and config arguments through to `shouldUpdateDependencyBasedOnConfig`. Add regression coverage for a dependent package changelog entry so the custom `getChangelogEntry` path exercises internal dependency updates.

Validated with:

- `pnpm run version`
- `pnpm vitest run .changeset/get-changelog-entry-test.ts`
- `pnpm lint`

### DIFF
--- a/.changeset/get-changelog-entry-test.ts
+++ b/.changeset/get-changelog-entry-test.ts
@@ -7,10 +7,21 @@ test("changesets getChangelogEntry hook", async () => {
   // Create a temporary package to avoid touching real repo packages
   const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "changesets-test-"));
   const pkgDir = path.join(tempDir, "pkg-a");
+  const dependentPkgDir = path.join(tempDir, "pkg-b");
   await fs.ensureDir(pkgDir);
+  await fs.ensureDir(dependentPkgDir);
   await fs.writeJson(
     path.join(pkgDir, "package.json"),
     { name: "pkg-a", version: "1.0.0" },
+    { spaces: 2 },
+  );
+  await fs.writeJson(
+    path.join(dependentPkgDir, "package.json"),
+    {
+      name: "pkg-b",
+      version: "1.0.0",
+      dependencies: { "pkg-a": "^1.0.0" },
+    },
     { spaces: 2 },
   );
 
@@ -39,6 +50,12 @@ test("changesets getChangelogEntry hook", async () => {
         newVersion: "1.1.0",
         changesets: ["fake-id-1", "fake-id-2", "fake-id-3"],
       },
+      {
+        name: "pkg-b",
+        type: "patch",
+        newVersion: "1.0.1",
+        changesets: [],
+      },
     ],
     preState: undefined,
   } as const;
@@ -49,6 +66,14 @@ test("changesets getChangelogEntry hook", async () => {
       {
         dir: pkgDir,
         packageJson: { name: "pkg-a", version: "1.0.0" },
+      },
+      {
+        dir: dependentPkgDir,
+        packageJson: {
+          name: "pkg-b",
+          version: "1.0.0",
+          dependencies: { "pkg-a": "^1.0.0" },
+        },
       },
     ],
   } as const;
@@ -91,6 +116,18 @@ test("changesets getChangelogEntry hook", async () => {
     ### Other updates
 
     - Chore: update docs
+    "
+  `);
+
+  const dependentChangelogPath = path.join(dependentPkgDir, "CHANGELOG.md");
+  const dependentChangelog = await fs.readFile(dependentChangelogPath, "utf8");
+
+  expect(dependentChangelog).toMatchInlineSnapshot(`
+    "# pkg-b
+
+    ## 1.0.1
+
+    - Updated dependencies: \`pkg-a@1.1.0\`
     "
   `);
 });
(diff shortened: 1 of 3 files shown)


diegohaz (commented):
## Review summary

**Scope:** Initial review of `0520c3e6f0cd2789b15a2efeae9ef9aa516ec191` against base `1739d3e58d1c1ab5784c10ab518a9c48e16bba84`. Files: `patches/@changesets__apply-release-plan@7.1.1.patch`, `.changeset/get-changelog-entry-test.ts`, `pnpm-lock.yaml`.

### What I checked

- **Patch correctness.** Pulled the upstream `@changesets/apply-release-plan@7.1.1` tarball and confirmed `shouldUpdateDependencyBasedOnConfig`'s signature is `(cwd, release, { depVersionRange, depType }, { minReleaseType, onlyUpdatePeerDependentsWhenOutOfRange })`. The previous patch passed only three arguments, so `{ minReleaseType, ... }` was destructured from `undefined` and threw at release time. The new call now matches upstream — including `cwd`, `oldVersion`, and `dir` on the release object — so `getBumpLevel(minReleaseType)` and the workspace-range branch in `shouldUpdateDependencyBasedOnConfig` both have the data they need.
- **Surrounding inlined logic.** The rest of the inlined block is a faithful copy of upstream `getChangelogEntry` (lines 163-212 in `dist/changesets-apply-release-plan.cjs.js`); no other arguments drift. The `release.type === "none"` short-circuit and the `else` fallback to upstream `getChangelogEntry` are preserved.
- **Bug actually triggered in this repo.** With the current pending changesets (`@ariakit/react-core` patch + `@ariakit/react` patch), `pnpm run version` would walk through this exact filter when generating `@ariakit/react`'s changelog entry, since `@ariakit/react` re-exports `@ariakit/react-core`. So the regression has been masking a real release blocker.
- **Regression test.** Added `pkg-b` depending on `pkg-a@^1.0.0` exercises the dependent-release branch that previously crashed. Ran `pnpm install --frozen-lockfile && pnpm vitest run .changeset/get-changelog-entry-test.ts` locally — passes, snapshot for `pkg-b` matches `Updated dependencies: \`pkg-a@1.1.0\``.
- **Lockfile.** Only the `@changesets/apply-release-plan@7.1.1` patch hash changes; both occurrences are updated in lockstep. No unrelated drift.
- **Changeset.** None needed — the change only affects internal release tooling, not any published package's public API or runtime behavior.

### Findings

No new substantive findings beyond what's already on the PR. Copilot and CodeRabbit both flagged the missing `try/finally` cleanup of `tempDir` in the test — that's a valid hygiene nit but pre-existed this PR (the test was added in #4928 with the same pattern), so it's a reasonable choice to address or defer here.

### Not validated

- A real `pnpm run version` dry run against the actual pending changesets — skipped to avoid mutating the working tree. The targeted unit test exercises the same code path.

.changeset/get-changelog-entry-test.ts
```diff
@@ -7,12 +7,23 @@ test("changesets getChangelogEntry hook", async () => {
   // Create a temporary package to avoid touching real repo packages
   const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "changesets-test-"));
   const pkgDir = path.join(tempDir, "pkg-a");
+  const dependentPkgDir = path.join(tempDir, "pkg-b");
   await fs.ensureDir(pkgDir);
+  await fs.ensureDir(dependentPkgDir);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 097f7c0b4 - Wrapped the test workspace setup and assertions in a `try/finally` and remove the temporary directory with `await fs.remove(tempDir)`.

diegohaz:
**Fixed**

- 097f7c0b4 ([reply](https://github.com/ariakit/ariakit/pull/5904#discussion_r3184407944)): Cleaned up the Changesets changelog test temp workspace with `try/finally` and `await fs.remove(tempDir)`.

Validated with `pnpm vitest run .changeset/get-changelog-entry-test.ts` and `pnpm lint`.

.changeset/get-changelog-entry-test.ts
```diff
@@ -119,15 +120,21 @@ test("changesets getChangelogEntry hook", async () => {
...
+    expect(dependentChangelog).toMatchInlineSnapshot(`
     "# pkg-b
 
     ## 1.0.1
 
     - Updated dependencies: \`pkg-a@1.1.0\`
     "
   `);
+  } finally {
+    await fs.remove(tempDir);
```
diegohaz, line 138: Shouldn't this be done in a vitest after hook?

diegohaz: This is not an issue because the temporary directory is created inside this single test, and the `finally` block keeps cleanup scoped to that exact directory while still running on assertion failures. A Vitest hook would need mutable outer state for the same behavior, so `try/finally` is more local here.

diegohaz:
**Invalid**

- Current code ([reply](https://github.com/ariakit/ariakit/pull/5904#discussion_r3184449637)): The cleanup is intentionally scoped with `try/finally` because the temporary directory is created inside a single test; a Vitest hook would need mutable outer state for the same behavior.

No code changes were needed.

PR #5905, opened 2026-05-04
Stabilize CI performance test comparisons

diegohaz:
## Motivation

PRs that don't touch runtime code (e.g. dependency-only updates such as #5898) were reporting 1–12% baseline-vs-current swings on the perf job. Two factors drove the noise:

1. Baseline and current measurements ran sequentially in the same job, so any runner-level event during one half — CPU governor changes, neighbor jobs on the host, GC pauses — landed entirely on that side and showed up as a regression or improvement.
2. Sub-metric jitter at small absolute values surfaced as flagged regressions even when the primary timings hadn't actually moved.

## Solution

Two independent changes that each reduce a different source of variance and compound when applied together:

- **Interleave rounds across a parallel worktree.** `PERF_ROUNDS=3` baseline/current rounds run within the same job. The baseline checkout lives in a `git worktree add --detach` directory so we don't have to swap branches between rounds. Runner-level noise now affects both sides roughly evenly. `HUSKY=0` is set in the perf job because the worktree shares `.git/config`, and we don't want husky to rewrite `core.hooksPath` mid-job.
- **Tighter significance gating.** Only `scripting`, `rendering`, and `total` can flag a regression. A flag requires >10% magnitude AND >5 ms absolute delta. Multi-round runs additionally require directional agreement: ≤4 rounds → unanimous, ≥5 → all-but-one. These thresholds live in `perf-compare.ts` constants and are covered by the test file.

CI also uploads each round's per-worker JSON as an artifact so post-mortem inspection is possible without re-running the job.

A `perf.ts` measurement-layer rewrite (force-GC via `HeapProfiler.collectGarbage` + a custom navigation helper that replaced `networkidle` with `domcontentloaded` + `fonts.ready` + 2× rAF + a short `requestIdleCallback`) was attempted in the original commit but had to be reverted in 039e31f4 — touching `perf.ts` in the same PR contaminates the baseline-vs-current comparison since baseline runs the old `perf.ts` and current runs the new one. That class of measurement-layer change has to land separately on `main` first, then get measured in a follow-up.

## Robustness fixes (review feedback)

- `readJsonFile` now hard-fails on read/parse errors and non-array shapes (e56745c4) so a truncated or malformed artifact surfaces as a real error instead of silently degrading the comparison into "no baseline / new tests".
- Tests with zero paired baseline/current rounds surface in a separate "Tests without paired rounds" section (45cc88d5) instead of being flagged from one-sided medians, with a new "No comparable rounds across baseline and current." headline for unpaired-only runs.
- Paired-round count is tracked per row (`pairedRoundsCount` on `ComparisonRow`) and the footer notes mixed counts when row counts differ (45cc88d5), instead of reporting a single global number.
- CI hardening (2ed63375): `xvfb-run -a` to avoid `/tmp/.X99-lock` collisions across rounds in the same job, wipe `site/.perf-results` (current and baseline worktree) before each run so leftover files can't poison `discoverRoundFiles`, and `include-hidden-files: true` on the artifact upload (the path starts with a dot).
- Agreement check tied to all paired rounds (ed31d5f5): switched `perRoundPercents` → `perRoundDeltas` so a 2-round row with one zero-baseline sample no longer collapses the denominator to one and flags from a single agreeing round. New regression test for `[0, 100]` vs `[0, 130]`.
- Workflow fail-fast on baseline collection (852c4662): dropped `|| true` from the baseline `xvfb-run`, replaced the silent `cp` with a `shopt -s nullglob` array-length guard, and removed `if: always()` from `Compare results` and `Post or update PR comment`. `Upload performance results` keeps `if: always()` for debugging artifacts.
- Baseline Playwright install (f97eb0b5): `pnpm exec playwright install --with-deps chromium` from `$BASELINE_DIR` after baseline `pnpm install`, so dependency-bump PRs install the baseline lockfile's browser revision into the shared `~/.cache/ms-playwright`.

## Test plan

- [x] `pnpm tsc` clean
- [x] `pnpm lint` clean
- [x] `pnpm test perf-compare` — 24 cases pass (threshold flagging, absolute-delta floor, sub-metric jitter rejection, agreement variants including unanimity at two rounds and the zero-baseline denominator regression test, asymmetric round counts, malformed-JSON hard failure, unpaired-only headline, mixed paired-round footer, profile-mode mismatch warning, shard combination)
- [ ] CI perf job runs with `PERF_ROUNDS=3` and reports stable comparisons for runtime-neutral PRs

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -331,21 +331,25 @@ jobs:
         # out runner-level noise (CPU governor, neighbor jobs) that a
         # sequential run would attribute entirely to one side. Baseline
         # failures are tolerated because new perf tests don't exist yet on
-        # the base branch.
+        # the base branch. `xvfb-run -a` picks the next free display number
+        # so back-to-back rounds in this job don't collide on `/tmp/.X99-lock`
+        # if a prior `xvfb-run` invocation hasn't finished tearing down its
+        # default display.
         run: |
+          rm -rf site/.perf-results "$BASELINE_DIR/site/.perf-results"
           mkdir -p site/.perf-results
           for i in $(seq 1 "$PERF_ROUNDS"); do
             echo "::group::Round $i — baseline"
             (cd "$BASELINE_DIR" && \
               PERF_RESULTS_FILE="baseline-$i.json" \
-              xvfb-run pnpm -F site run test-perf --pass-with-no-tests) || true
+              xvfb-run -a pnpm -F site run test-perf --pass-with-no-tests) || true
             cp "$BASELINE_DIR"/site/.perf-results/baseline-"$i"-*.json \
               site/.perf-results/ 2>/dev/null || true
             echo "::endgroup::"
 
             echo "::group::Round $i — current"
             PERF_RESULTS_FILE="current-$i.json" \
-              xvfb-run pnpm -F site run test-perf --pass-with-no-tests
+              xvfb-run -a pnpm -F site run test-perf --pass-with-no-tests
             echo "::endgroup::"
           done
 
@@ -360,6 +364,9 @@ jobs:
           name: perf-results
           path: site/.perf-results
           if-no-files-found: ignore
+          # `.perf-results` starts with a dot, so upload-artifact treats it as
+          # hidden and skips its contents unless this is set.
+          include-hidden-files: true
           retention-days: 7
 
       - name: Post or update PR comment
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -218,6 +218,15 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+    env:
+      # Number of interleaved baseline/current rounds. Three rounds keeps the
+      # job under ~10 minutes while letting the agreement check filter out
+      # one-off CI hiccups.
+      PERF_ROUNDS: 3
+      # The baseline worktree shares `.git/config` with the main checkout, so
+      # husky's prepare step would otherwise rewrite `core.hooksPath` to a
+      # relative path under a temporary worktree that is removed at job end.
+      HUSKY: 0
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
@@ -232,17 +241,16 @@ jobs:
         with:
           engine: chromium
 
-      # --- Baseline (base branch) ---
-      - name: Checkout base branch
-        run: git checkout -f ${{ github.event.pull_request.base.sha }}
+      - name: Set up baseline worktree
+        run: |
+          BASELINE_DIR="$RUNNER_TEMP/perf-baseline"
+          echo "BASELINE_DIR=$BASELINE_DIR" >> "$GITHUB_ENV"
+          git worktree add --detach "$BASELINE_DIR" \
+            ${{ github.event.pull_request.base.sha }}
 
-      - name: Set up base environment
-        uses: ./.github/workflows/setup
-
-      - name: Install Playwright (base)
-        uses: ./.github/workflows/install-playwright
-        with:
-          engine: chromium
+      - name: Install baseline dependencies
+        working-directory: ${{ env.BASELINE_DIR }}
+        run: pnpm install --frozen-lockfile
 
       - name: Find base app artifact
         id: find-base-artifact
@@ -277,7 +285,7 @@ jobs:
         uses: actions/download-artifact@v8
         with:
           name: app-dist
-          path: site/dist
+          path: ${{ env.BASELINE_DIR }}/site/dist
           github-token: ${{ github.token }}
           run-id: ${{ steps.find-base-artifact.outputs.run-id }}
 
@@ -288,42 +296,23 @@ jobs:
         uses: actions/download-artifact@v8
         with:
           name: nextjs-dist
-          path: nextjs/.open-next
+          path: ${{ env.BASELINE_DIR }}/nextjs/.open-next
           github-token: ${{ github.token }}
           run-id: ${{ steps.find-base-artifact.outputs.run-id }}
 
       - name: Clean partial base artifacts
         if: steps.base-artifact.outcome != 'success' || steps.base-nextjs-artifact.outcome != 'success'
-        run: rm -rf site/dist nextjs/.open-next
+        run: rm -rf "$BASELINE_DIR/site/dist" "$BASELINE_DIR/nextjs/.open-next"
 
       - name: Build base app
         if: steps.base-artifact.outcome != 'success' || steps.base-nextjs-artifact.outcome != 'success'
+        working-directory: ${{ env.BASELINE_DIR }}
         run: pnpm -F site run build-lite
 
       - name: Build base Next.js app
         if: steps.base-artifact.outcome != 'success' || steps.base-nextjs-artifact.outcome != 'success'
+        working-directory: ${{ env.BASELINE_DIR }}
         run: pnpm -F nextjs run build-worker
-
-      - name: Run baseline perf tests
-        # Baseline failures are expected when perf tests are new.
-        run: |
-          PERF_RESULTS_FILE=baseline.json \
-          xvfb-run pnpm -F site run test-perf --pass-with-no-tests || true
-
-      - name: Save baseline results
-        run: |
-          mkdir -p /tmp/perf-baseline
-          cp site/.perf-results/baseline*.json /tmp/perf-baseline/ 2>/dev/null || true
-
-      # --- Current (merge commit) ---
-      - name: Checkout merge commit
-        run: git checkout -f ${{ github.sha }}
-
-      - name: Set up PR environment
-        uses: ./.github/workflows/setup
-
-      - name: Clean base app artifacts
-        run: rm -rf site/dist nextjs/.open-next
 
       - name: Download app artifact
         uses: actions/download-artifact@v8
@@ -337,19 +326,41 @@ jobs:
           name: nextjs-dist
           path: nextjs/.open-next
 
-      - name: Restore baseline results
+      - name: Run interleaved perf tests
+        # Alternating baseline and current within the same job lets us cancel
+        # out runner-level noise (CPU governor, neighbor jobs) that a
+        # sequential run would attribute entirely to one side. Baseline
+        # failures are tolerated because new perf tests don't exist yet on
+        # the base branch.
         run: |
           mkdir -p site/.perf-results
-          cp /tmp/perf-baseline/*.json site/.perf-results/ 2>/dev/null || true
+          for i in $(seq 1 "$PERF_ROUNDS"); do
+            echo "::group::Round $i — baseline"
+            (cd "$BASELINE_DIR" && \
+              PERF_RESULTS_FILE="baseline-$i.json" \
+              xvfb-run pnpm -F site run test-perf --pass-with-no-tests) || true
+            cp "$BASELINE_DIR"/site/.perf-results/baseline-"$i"-*.json \
+              site/.perf-results/ 2>/dev/null || true
+            echo "::endgroup::"
 
-      - name: Run current perf tests
-        run: |
-          PERF_RESULTS_FILE=current.json \
-          xvfb-run pnpm -F site run test-perf --pass-with-no-tests
+            echo "::group::Round $i — current"
+            PERF_RESULTS_FILE="current-$i.json" \
+              xvfb-run pnpm -F site run test-perf --pass-with-no-tests
+            echo "::endgroup::"
+          done
 
       - name: Compare results
         if: always()
         run: pnpm -F site run perf-compare
+
+      - name: Upload performance results
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: perf-results
+          path: site/.perf-results
+          if-no-files-found: ignore
+          retention-days: 7
 
       - name: Post or update PR comment
         if: always()
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -329,22 +329,30 @@ jobs:
       - name: Run interleaved perf tests
         # Alternating baseline and current within the same job lets us cancel
         # out runner-level noise (CPU governor, neighbor jobs) that a
-        # sequential run would attribute entirely to one side. Baseline
-        # failures are tolerated because new perf tests don't exist yet on
-        # the base branch. `xvfb-run -a` picks the next free display number
-        # so back-to-back rounds in this job don't collide on `/tmp/.X99-lock`
-        # if a prior `xvfb-run` invocation hasn't finished tearing down its
-        # default display.
+        # sequential run would attribute entirely to one side. `xvfb-run -a`
+        # picks the next free display number so back-to-back rounds in this
+        # job don't collide on `/tmp/.X99-lock` if a prior `xvfb-run`
+        # invocation hasn't finished tearing down its default display.
+        # `--pass-with-no-tests` keeps the very-first-perf-test PR working
+        # because the base SHA simply has no perf files for Playwright to
+        # discover; an actual baseline failure (server crash, browser mismatch,
+        # truncated artifact) must still propagate so the PR comment doesn't
+        # publish a misleading comparison built on partial data. `nullglob`
+        # makes the round-N glob deterministic when the round produced no
+        # files instead of expanding to a literal pattern.
         run: |
+          shopt -s nullglob
           rm -rf site/.perf-results "$BASELINE_DIR/site/.perf-results"
           mkdir -p site/.perf-results
           for i in $(seq 1 "$PERF_ROUNDS"); do
             echo "::group::Round $i — baseline"
             (cd "$BASELINE_DIR" && \
               PERF_RESULTS_FILE="baseline-$i.json" \
-              xvfb-run -a pnpm -F site run test-perf --pass-with-no-tests) || true
-            cp "$BASELINE_DIR"/site/.perf-results/baseline-"$i"-*.json \
-              site/.perf-results/ 2>/dev/null || true
+              xvfb-run -a pnpm -F site run test-perf --pass-with-no-tests)
+            baseline_files=("$BASELINE_DIR"/site/.perf-results/baseline-"$i"-*.json)
+            if [ ${#baseline_files[@]} -gt 0 ]; then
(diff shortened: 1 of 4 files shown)


site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +114,322 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
+      const currentRound = cur.byRound.get(roundIndex);
+      if (!baselineRound || !currentRound) continue;
+      pairedRoundIndices.add(roundIndex);
+      sharedBaseline.push(baselineRound);
+      sharedCurrent.push(currentRound);
+    }
+    // Aggregate displayed baseline/current from shared rounds only. Otherwise
+    // an unpaired round can drag one side's median past the paired rounds and
+    // even flip the sign of the change relative to per-round percents.
+    const baselineMetrics =
```
diegohaz: This fallback still compares tests that have no paired baseline/current round. If a test has `baseline-1` and `current-2` only, `sharedBaseline` and `sharedCurrent` are empty, so the code falls back to each side's aggregate metrics. Then `perRoundPercents` is empty, `agreementOk` becomes true, and the report can show a real regression even though there was no comparable pair. I verified this with synthetic results and got a warning row for `100ms -> 130ms`. Treat `sharedRounds.length === 0` as unpaired/inconclusive instead of generating comparison rows.

diegohaz (commented):
Reviewed the comparator, perf measurement changes, workflow changes, tests, and compared the behavior against PR #5903. This is a full initial review of #5905 at `2792f4058909d8bc9b6402e577ecf7611109c3ae`.

I found four correctness/reliability issues in the comparator. I also validated the first three with synthetic `.perf-results` runs against this branch:
- disjoint `baseline-1` / `current-2` results are reported as a real regression;
- a two-round split with one improvement and one regression is still flagged;
- malformed JSON is silently converted into a missing-baseline report.

I did not run the Vitest suite locally in the fresh review worktree because dependencies were not installed there, but the PR CI shows the main TypeScript/lint/test checks passing or in progress at review time.

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +114,322 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
+  return aggregated;
+}
+
+// Direction-of-change agreement check across rounds. Up to two rounds we fall
+// back to magnitude only — with so few samples a single noise round would
+// otherwise veto every flag, which is the failure mode this exists to avoid.
+// Three or four rounds require unanimity. Five or more rounds tolerate a
+// single dissenter so a one-off CI hiccup cannot block an alert.
+function requiredAgreement(roundsCount: number) {
+  if (roundsCount <= 2) return 1;
```
diegohaz: Two shared rounds can still flag when the rounds disagree. For example, baseline `[100, 100]` and current `[80, 150]` reports `+15%` and gets `:warning:` because one agreeing round is enough. That reintroduces the flip-flop the agreement check is supposed to filter, especially when one of the three CI rounds is missing and a test falls back to two paired rounds. Require both rounds to agree for `roundsCount === 2`.

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +114,322 @@ const METRIC_LABELS: Record<MetricKey, string> = {
   total: "Total",
 };
```
diegohaz, line 139: Swallowing parse errors can hide broken artifacts as a clean comparison state. I verified that a truncated `baseline-worker0.json` produces `No baseline results available for comparison.` instead of failing the comparator. A corrupt JSON artifact should fail fast with the file path and original parse error so the PR comment reflects broken input rather than missing or new tests.

site/src/test-utils/perf-compare.ts
```diff
@@ -183,29 +438,60 @@ function compare(): ComparisonSummary {
...
 
   return {
     rows,
     hasSignificantChanges: rows.some((r) => r.significant),
-    currentResults: current,
+    currentResults: [...current.values()],
     profileModeMismatches,
     newTests,
     removedTests,
+    pairedRoundsCount: pairedRoundIndices.size,
```
diegohaz: This is the union of all shared round indices across the whole report, not the count used by each compared test. If one test has three paired rounds and another has one, the footer still says the comparison spans three interleaved rounds, overstating confidence for the one-round row. Track the paired count per comparison row/test and only print a global count when all compared rows used the same count; otherwise use a mixed-count note or per-row count.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in e56745c46 - `readJsonFile` now hard-fails on read/parse errors and on non-array shapes, so a truncated or malformed artifact surfaces as a real error instead of silently degrading the comparison into "no baseline".

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in a2432dbbc - `requiredAgreement` now requires unanimity at ≤4 rounds (returns `roundsCount`) and only allows one disagreeing round at 5+. Tests cover the two-rounds case explicitly.

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +114,322 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
+        );
+      }
+      const { agreement, significant } = computeSignificance({
+        metric,
+        delta,
+        percent,
+        baseline: baseVal,
+        current: curVal,
+        perRoundPercents,
+      });
```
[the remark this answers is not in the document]
diegohaz: Fixed in 45cc88d56 - tests with zero paired rounds now surface in a separate "Tests without paired rounds" section instead of being compared against one-sided metrics. The fallback to `base.metrics`/`cur.metrics` is gone.

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +114,322 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
+    // an unpaired round can drag one side's median past the paired rounds and
+    // even flip the sign of the change relative to per-round percents.
+    const baselineMetrics =
+      sharedBaseline.length > 0
+        ? computeMedianMetrics(sharedBaseline)
+        : base.metrics;
+    const currentMetrics =
+      sharedCurrent.length > 0
+        ? computeMedianMetrics(sharedCurrent)
+        : cur.metrics;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 45cc88d56 - tests with zero paired rounds now surface in a separate "Tests without paired rounds" section instead of being compared against one-sided metrics. The fallback to `base.metrics`/`cur.metrics` is gone.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in 45cc88d56 - tests with zero paired rounds now surface in a separate "Tests without paired rounds" section instead of being compared against one-sided metrics.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in 45cc88d56 - paired-round count is now tracked per-row (`pairedRoundsCount` on `ComparisonRow`), and the footer notes mixed counts when rows differ instead of reporting a single global number.

diegohaz:
**Fixed**

- `e56745c4` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185012744)): `readJsonFile` now hard-fails on read/parse errors and on non-array shapes, so a truncated or malformed artifact surfaces as a real error instead of silently degrading the comparison into "no baseline".
- `a2432dbb` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185012951)): `requiredAgreement` now requires unanimity at ≤4 rounds (returns `roundsCount`) and only allows one disagreeing round at 5+. `MIN_DELTA_MS` also bumped from 1 to 5 to cover CDP timer quantization plus typical run-to-run jitter. New tests cover the two-rounds case.
- `45cc88d5` ([reply 1](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013128) · [reply 2](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013245) · [reply 3](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013414) · [reply 4](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013585)): Tests with zero paired baseline/current rounds now surface in a separate "Tests without paired rounds" section instead of being compared against one-sided metrics; paired-round count is tracked per-row (`pairedRoundsCount` on `ComparisonRow`); footer notes mixed counts when rows differ; new "No comparable rounds across baseline and current." headline replaces the misleading "No performance results found." for unpaired-only runs.

**Also pushed in this round**

- `039e31f4`: Reverted the `perf.ts` measurement-layer changes (force-GC + custom navigation helper). Those changes contaminated the baseline-vs-current comparison in this PR — baseline runs the old `perf.ts` and current runs the new one, so any measurement-method delta showed up as a fake regression (e.g. dialog-framer-motion +128% scripting, ariakit-tailwind/toggle -100%). Measurement-layer changes have to land separately on `main` first, then get measured in a follow-up.
- `2ed63375`: CI hardening — `xvfb-run -a` to avoid `/tmp/.X99-lock` collisions across rounds, wipe `site/.perf-results` and the baseline worktree's copy before each run so leftover files can't poison `discoverRoundFiles`, and `include-hidden-files: true` on `actions/upload-artifact` (the path starts with a dot).

**Verification**

- `pnpm tsc`: 0 errors
- `pnpm test`: 797/797 passing (23 new comparator cases including unpaired-only headline and mixed paired-round footer)
- `pnpm lint-fix`: 0 errors

.github/workflows/app.yml
```diff
@@ -337,20 +326,49 @@ jobs:
...
-
-      - name: Run current perf tests
-        run: |
-          PERF_RESULTS_FILE=current.json \
-          xvfb-run pnpm -F site run test-perf --pass-with-no-tests
+          for i in $(seq 1 "$PERF_ROUNDS"); do
+            echo "::group::Round $i — baseline"
+            (cd "$BASELINE_DIR" && \
+              PERF_RESULTS_FILE="baseline-$i.json" \
+              xvfb-run -a pnpm -F site run test-perf --pass-with-no-tests) || true
```
diegohaz: This still lets the perf job publish a comparison after baseline collection failed. The baseline command is ignored with `|| true`, missing files are swallowed by `cp ... || true`, and then both `Compare results` and `Post or update PR comment` run under `if: always()`. A baseline install/server/browser failure would therefore become a passing perf job with a "No baseline..." or current-only comparison, which looks like a real CI result. Adding a new perf test shouldn't require ignoring the whole baseline command; the base checkout will just not include that new file while existing perf tests should still produce artifacts. Please fail when an expected baseline/current round writes no JSON and let compare/comment run only after the loop succeeds, keeping upload as `always()` for debugging artifacts.

.github/workflows/app.yml
```diff
@@ -232,17 +241,16 @@ jobs:
...
-      - name: Set up base environment
-        uses: ./.github/workflows/setup
-
-      - name: Install Playwright (base)
-        uses: ./.github/workflows/install-playwright
-        with:
-          engine: chromium
+      - name: Install baseline dependencies
+        working-directory: ${{ env.BASELINE_DIR }}
+        run: pnpm install --frozen-lockfile
```
diegohaz, line 253: The baseline worktree installs its dependencies but never installs the Playwright browser revision for that lockfile. The earlier `Install Playwright` step ran in the current checkout, so a PR that bumps `@playwright/test` can leave the baseline runner with only the current browser revision installed; the baseline `test-perf` then fails before producing artifacts. That path is especially relevant for dependency-only PRs, which are one of the cases this work is meant to make trustworthy. Add a baseline `pnpm exec playwright install --with-deps chromium` after `pnpm install` in `$BASELINE_DIR` (or teach the composite action to run there).

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +132,347 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
-        Math.abs(percent) > THRESHOLD_PERCENT ||
-        (baseVal === 0 && Math.abs(curVal) > 1);
+      const perRoundPercents: number[] = [];
+      for (let i = 0; i < sharedBaseline.length; i++) {
+        const baselineRoundEntry = sharedBaseline[i];
+        const currentRoundEntry = sharedCurrent[i];
+        if (!baselineRoundEntry || !currentRoundEntry) continue;
+        const baselineRoundValue = baselineRoundEntry.metrics[metric];
+        const currentRoundValue = currentRoundEntry.metrics[metric];
+        if (baselineRoundValue <= 0) continue;
```
diegohaz: This drops paired rounds where the baseline metric is zero before the agreement check. I verified current head with two paired rounds, baseline `[0, 100]` and current `[0, 130]`: the report flags `+30%` even though only one of the two paired rounds provides an agreeing direction; the zero/no-change round was removed from `perRoundPercents`. Keep the agreement denominator tied to all shared rounds, for example by storing raw per-round deltas for agreement and using percentages only for display/magnitude when the baseline is greater than zero. PR #5903 avoids this false positive on the same synthetic input with delta-based agreement.

diegohaz (commented):
Re-reviewed #5905 at `2ed633754bf10c334e12128022b6519885b8c2de`, using the prior marked review head `2792f4058909d8bc9b6402e577ecf7611109c3ae` as the incremental baseline. I reviewed the updated workflow, comparator, tests, current review threads, CI status, and compared the behavior against #5903.

The earlier issues around disjoint rounds, two-round disagreement, malformed JSON, and overstated mixed counts were mostly addressed. I left three remaining findings: baseline artifact failures can still be hidden behind a successful perf comment, the baseline worktree does not install its own Playwright browser revision, and agreement can still be computed from fewer rounds when a paired baseline value is zero. I validated the last case with a synthetic two-round artifact; #5905 flags `[0, 100] -> [0, 130]`, while #5903 does not.

CI is passing on #5905, so these are reliability gaps rather than current red checks. I did not run the full Vitest suite locally in this fresh review worktree because dependencies are not installed there; the focused synthetic comparator check ran directly with Node 24.

site/src/test-utils/perf-compare.ts
```diff
@@ -73,108 +132,347 @@ const METRIC_LABELS: Record<MetricKey, string> = {
...
+        );
+      }
+      const { agreement, significant } = computeSignificance({
+        metric,
+        delta,
+        percent,
+        baseline: baseVal,
+        current: curVal,
+        perRoundPercents,
+      });
```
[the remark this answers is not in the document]
diegohaz: Fixed in ed31d5f50 — the agreement check now operates on per-round deltas for every shared round (`perRoundPercents` → `perRoundDeltas`), so a 2-round row with one zero-baseline sample still requires the full denominator instead of collapsing to one. New regression test exercises baseline `[0, 100]` vs current `[0, 130]` (medians 50→65, +30%) — must stay unflagged.

site/src/test-utils/perf-compare.ts
diegohaz: Fixed in ed31d5f50 — per-round deltas now feed the agreement check directly so `[0, 100]` baseline vs `[0, 130]` current stays unflagged. Same fix also resolves coderabbitai #3185024349 since both threads point at the same denominator-shrinking behavior. Added a regression test for that exact scenario.

.github/workflows/app.yml
diegohaz: Fixed in 852c46624 — dropped `|| true` from the baseline `xvfb-run`, replaced the silent `cp` with a `shopt -s nullglob` array-length guard, and removed `if: always()` from `Compare results` and `Post or update PR comment`. `Upload performance results` keeps `if: always()` so the raw `.perf-results` artifacts still upload on failure for debugging.

.github/workflows/app.yml
diegohaz: Fixed in f97eb0b5f — added a `pnpm exec playwright install --with-deps chromium` step from `$BASELINE_DIR` after baseline `pnpm install`, so dependency-bump PRs install the baseline lockfile`s browser revision into the global `~/.cache/ms-playwright` (it already holds the head revision from the main checkout`s composite step).

diegohaz:
**Fixed**

- `ed31d5f50` ([coderabbitai reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185214767), [self-review reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185215103)): switched `perRoundPercents` → `perRoundDeltas` so the agreement check operates on every paired round, including rounds with a zero baseline. Adds a regression test for `[0, 100]` vs `[0, 130]` (median 50→65, +30% — must stay unflagged because only 1 of 2 rounds agrees).
- `852c46624` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185215351)): drops `|| true` from the baseline `xvfb-run`, replaces the silent `cp` with a `shopt -s nullglob` array-length guard, and removes `if: always()` from `Compare results` and `Post or update PR comment`. `Upload performance results` keeps `if: always()` for debugging artifacts.
- `f97eb0b5f` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185215558)): adds `pnpm exec playwright install --with-deps chromium` from `$BASELINE_DIR` after baseline `pnpm install`, so dependency-bump PRs install the baseline lockfile's browser revision into the shared `~/.cache/ms-playwright`.

**Already resolved**

- `45cc88d56` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013245), [reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013414)): tests with zero paired rounds surface in a "Tests without paired rounds" section instead of being compared via cross-round medians.
- `45cc88d56` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185013585)): paired-round count tracked per-row, footer notes mixed counts when rows differ.
- `a2432dbbc` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185012951)): `requiredAgreement` requires unanimity at ≤4 rounds.
- `e56745c46` ([reply](https://github.com/ariakit/ariakit/pull/5905#discussion_r3185012744)): `readJsonFile` hard-fails on read/parse errors and on non-array shapes instead of silently degrading the comparison.

PR #5945, opened 2026-05-09
Make layer lightness limits affect offset and push

diegohaz:
## Motivation

`ak-layer-offset-*` and `ak-layer-push-*` did not consistently account for `ak-layer-min-*` and `ak-layer-max-*`, so layer movement could ignore the lightness bounds requested for the layer. The interaction with the forbidden mid-luminance range and high-contrast mode also needed to be explicit: layer min/max should feed the layer result before forbidden-range escape, but should not pull the final color back into the forbidden range, and state offset/push utilities should remain independent overlays.

A related mismatch made `ak-state-*` and `ak-state-offset-*` resolve the forbidden range before the very-dark layer floor was applied, so state offsets could stay on the dark boundary where the matching layer offset escaped to the bright side.

## Solution

Clamp layer lightness before resolving `ak-layer-offset-*` and `ak-layer-push-*`, then let the forbidden-range logic escape from that clamped candidate. The layer offset utility now stores the raw directional delta and leaves forbidden-range resolution to the layer pipeline.

State offset now resolves forbidden-range escape from the floor-adjusted state candidate while still skipping `ak-layer-min-*` and `ak-layer-max-*`; `ak-state-push-*` remains on its independent state pipeline. High-contrast layer bias is applied after layer offset/push resolution for non-`ak-layer-contrast` layers, while `ak-layer-contrast` stays on its parent-relative contrast path.

The sandbox now uses a compact one-off cluster for layer min/max interactions instead of repeating those cases throughout the large matrix. It covers layer offset, layer push, arbitrary values, simultaneous min/max bounds, state offset, state push, and layer contrast with min/max. Computed-style snapshots cover light, dark, and high-contrast modes. The sandbox also fixes the `bordering inherit > host` case so the intermediate `ak-frame-bordering-inherit` wrapper participates as an `ak-frame`.

## Verification

- `pnpm lint-fix`
- `SITE_PORT=4322 pnpm -F site test-chrome ariakit-tailwind --grep-invert wcag`
- `SITE_PORT=4322 pnpm -F site test-chrome ariakit-tailwind -g wcag`
- `pnpm tsc`
- `git diff --check`
- `SITE_PORT=4322 PERF_RESULTS_FILE=baseline pnpm -F site test-perf ariakit-tailwind` against the previous branch state, then `SITE_PORT=4322 pnpm -F site test-perf ariakit-tailwind` with this change and `pnpm -F site perf-compare`: page load total `3210.8ms → 3234.6ms` (+1%); toggle total `3110.2ms → 2472.7ms` (-20%).


packages/ariakit-tailwind/src/input.ts
```diff
@@ -911,6 +911,41 @@ function getLayerL(relativeLightness: Value, absoluteLightness?: VarProperty) {
...
+      vars.lightnessOffsetDirection,
+      vars.forbiddenLa,
+      vars.forbiddenLb,
+      vars.forbiddenEntryBoundary,
+    ),
+  );
+  const offsetEnabled = fn.binary(fn.abs(inputs.layerIdleLOffsetDelta));
+  return fn.add(
+    limitedLightness,
+    fn.mul(offsetEnabled, fn.sub(resolvedOffsetLightness, limitedLightness)),
```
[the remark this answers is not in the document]
diegohaz: Fixed in 29650854b - the heavy forbidden-range offset expression is no longer part of the always-on `ak-layer` path. `ak-layer` now uses a clamp fallback, while `ak-layer-*` and `ak-layer-offset-*` compute the resolved offset only when their utility token matches. The same-sandbox perf comparison now reports no significant changes: page load total is +5% and toggle total is +6% against the old-`input.ts` baseline.

diegohaz:
**Fixed**
- 29650854b (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212603653): Scoped the heavier layer-offset forbidden-range math to matching `ak-layer-*` and `ak-layer-offset-*` utilities instead of the always-on `ak-layer` path. Same-sandbox perf comparison now reports no significant changes (+5% page-load total, +6% toggle total).

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1281,30 +1320,37 @@ utility(
...
+        ),
+        tokenValue,
       ),
     ),
   ];
+  return [
+    ...getDeclarations(bareDelta, bareValue),
+    ...getDeclarations(arbitraryDelta, arbitraryValue),
+  ];
 }
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because moving these intermediates back into `layerMathDeclarations` shifts the heavy resolved-offset formula into the static `ak-layer` path, so every layer computes it even when no layer offset/push utility is active. I tested that exact refactor locally against the same restored sandbox: page load total regressed from 4156.0ms to 5036.7ms (+21%), with style recalc from 4056.1ms to 4934.8ms (+22%). The current placement leaves the extra work scoped to `ak-layer-*`/`ak-layer-offset-*` utilities and matches the performance goal verified in `29650854b` and the temporary main-sandbox CI run.

diegohaz:
**Already Resolved**
- 29650854b (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212653592): Kept the resolved-offset work scoped to the layer offset utilities and verified the same-sandbox perf comparison.

**Invalid**
- No code change (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212735255): Tested the suggested move of offset intermediates into `ak-layer`; it regressed same-sandbox page load total by +21% and style recalc by +22%, so the current utility-scoped placement is the better performance tradeoff.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1091,7 +1121,16 @@ function getContrastL(selfRelativeL: Value, contrastValue: Value) {
...
 
 const layerIdle = fn.oklch(vars.layerIdleOffset, {
-  l: getContrastL(vars.layerIdlePushL, vars.layerIdleContrastValue),
+  l: fn.add(
+    getContrastL(vars.layerIdlePushL, vars.layerIdleContrastValue),
+    fn.mul(
+      vars.layerContrastBias,
+      fn.sub(
+        1,
+        fn.mul(vars.layerContrastDirection, vars.layerContrastDirection),
```
diegohaz, line 1123: **Nit / micro-perf opportunity.** This new `(1 - layerContrastDirection²)` gate and the existing `layerContrastDirection²` mask in `getContrastL` (lines 1116-1119) compute the same product. In the generated `--_ak-li` body, `(var(--_ak-lcd) * var(--_ak-lcd))` and `(1 - (var(--_ak-lcd) * var(--_ak-lcd)))` each appear twice:

```css
--_ak-li: oklch(from var(--_ak-lio) calc((((l + (var(--_ak-lcd) * max(0, ((clamp(0, (var(--_ak-lcpl) + (var(--_ak-licv) * var(--_ak-lcd))), 1) - l) * var(--_ak-lcd))))) * (var(--_ak-lcd) * var(--_ak-lcd))) + (var(--_ak-lipl) * (1 - (var(--_ak-lcd) * var(--_ak-lcd))))) + (var(--_ak-lcb) * (1 - (var(--_ak-lcd) * var(--_ak-lcd))))) c h);
```

Since `--_ak-li` is a static utility body that runs on every layer paint, hoisting `1 - direction²` (and optionally `direction²`) into a cached var via `layerMathDeclarations` — same pattern as `vars.contrastT` and `vars.contrastPushScale` on lines 1187-1188 — would shrink this calc and remove four multiplications per paint. Probably worth measuring with the perf comparison flow.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -911,6 +911,40 @@ function getLayerL(relativeLightness: Value, absoluteLightness?: VarProperty) {
...
+    getLayerL(relativeLightness, absoluteLightness),
+    inputs.layerLMax,
+  );
+}
+
+/**
+ * Resolves lightness limits together with the directional offset so min/max
+ * values cannot pull the result back into the forbidden mid-luminance range.
+ */
+function getLayerOffsetL(
```
diegohaz, line 922: **Nit / docs.** The three parameters here are tightly coupled: `limitedLightness = clamp(layerLMin, l + relativeLightness + floorBoost, layerLMax)` and `limitedDelta = limitedLightness - l`. The function trusts callers to pass consistent values, and the only caller (`getLayerOffsetDeclarations`) does. The reason it doesn't take just `relativeLightness` and recompute internally is that the dynamic utility already caches the intermediates as separate `@property` vars (`--_ak-lill`, `--_ak-lild`) so the static `--_ak-lio` body can read them via `var()` instead of re-running the clamp on every paint. A short sentence in the JSDoc above pointing at that caching invariant would help future readers understand why the signature is shaped this way (related to the CodeRabbit thread on commit 29650854).

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1395,7 +1441,9 @@ utility(
   set(vars.layerIdlePushValue, getPushValue(inputs.layerIdlePushL)),
   set(
     vars.layerIdlePushBaseL,
-    getLayerL(fn.mul(vars.layerIdlePushValue, vars.lightnessOffsetDirection)),
+    getLimitedLayerL(
```
diegohaz, line 1440: **Question — design clarification.** This swap (`getLayerL` → `getLimitedLayerL`) brings the *idle* push base under the new `inputs.layerLMin`/`layerLMax` clamp. The analogous `state-push-*` utility a few lines down still calls `getLayerL` (line 1580), and `state-offset-*` / `state-lighten-*` / `state-darken-*` likewise bypass the clamp. The sandbox additions explicitly cover `ak-layer-min-70 ak-state-offset-40` and `ak-layer-min-70 ak-state-push-40` (sandbox lines 163-176), so the asymmetry is exercised by snapshots.

Is it intentional that hover/active overlays can escape the lightness bounds the underlying layer is constrained to? If yes, a one-line note in the changeset would help — otherwise the same `getLimitedLayerL` treatment likely applies to the state-side utilities.

diegohaz (commented):
This review covers the four PR-specific commits at head `556fca3b` (the only files in the PR diff are `.changeset/300-tailwind-layer-lightness-limits.md`, `packages/ariakit-tailwind/src/{input.ts,output.css}`, the four `__snapshots__/` files, and `site/src/sandbox/ariakit-tailwind/index.react.tsx`).

The redesign is sound. Routing the offset/push pipelines through `getLimitedLayerL` so `inputs.layerLMin`/`layerLMax` clamp the lightness *before* the forbidden-range escape (rather than after) is the correct ordering, and the `getLayerOffsetL(limited, limitedDelta, raw)` shape — paired with caching `--_ak-lill` and `--_ak-lild` per dynamic utility — keeps the heavy `getResolvedLightnessOffset` math out of the always-on `ak-layer` body. The CodeRabbit thread on `getLayerOffsetL` (PRs commit `29650854`) was correctly resolved: hoisting the intermediates back into `layerMathDeclarations` would have moved the heavy formula onto every `ak-layer` element. The `2d21628e` / `556fca3b` pair (sandbox-removed for perf parity, then restored) is a clean way to keep the perf comparison apples-to-apples.

Local `pnpm tsc` passes. Three small things below — none blocking.

**Behavioral side effect worth confirming.** Moving `vars.layerContrastBias` from `layerIdleOffset.l` to `layerIdle.l` does what the PR description says for `ak-layer-contrast` (the `(1 - direction²)` gate suppresses the double-bias). It also changes `ak-layer-push-*` output: previously, when `--_ak-lipl` was overridden by the push utility, the bias on `layerIdleOffset.l` was discarded by the outer `oklch(layerIdleOffset, {l: getContrastL(layerIdlePushL,...)})` override; now bias is added unconditionally to `layerIdle.l`, so push utilities receive bias too. The dark-high-contrast "layer push" cell shifting from `oklch(1 ...)` to `oklch(0.9 ...)` is consistent with that. If this is the intended new behavior, no action — but it's worth calling out in the changeset since consumers using `ak-layer-push-*` in high-contrast mode will see the shift.

**Pre-existing question on state utilities.** This is broader than the PR but the PR brings the asymmetry into focus — see the inline comment on `state-offset-*`.

packages/ariakit-tailwind/src/input.ts
diegohaz: This is not an issue because I measured the suggested cache and it was slower on the interaction path: page load total improved from `3338.7ms` to `3275.4ms` (-2%), but toggle total regressed from `1688.3ms` to `3265.9ms` (+93%). I left the generated `--_ak-li` formula unchanged.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 56aa8e5c2 - documented the `baseDelta = baseLightness - l` invariant and why callers pass the cached delta into `getLayerOffsetL`.

diegohaz:
**Fixed**
- 56aa8e5c2 (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212921903): Documented the `baseDelta = baseLightness - l` invariant for `getLayerOffsetL`.

**Invalid**
- Measured (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212921799): Tried the suggested contrast-mask cache, but toggle total regressed from `1688.3ms` to `3265.9ms` (+93%), so the generated `--_ak-li` formula was left unchanged.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1091,15 +1114,24 @@ function getContrastL(selfRelativeL: Value, contrastValue: Value) {
...
 
 const layerIdle = fn.oklch(vars.layerIdleOffset, {
-  l: getContrastL(vars.layerIdlePushL, vars.layerIdleContrastValue),
+  l: fn.add(
+    getContrastL(vars.layerIdlePushL, vars.layerIdleContrastValue),
+    fn.mul(
+      vars.layerContrastBias,
+      fn.sub(
+        1,
+        fn.mul(vars.layerContrastDirection, vars.layerContrastDirection),
```
diegohaz, line 1123: **Nit / micro-perf, unchanged from prior review.** This `(1 - direction²)` and the matching `direction²` in `getContrastL` (lines 1109-1112) generate duplicated subexpressions in the static `--_ak-li` body — `(var(--_ak-lcd) * var(--_ak-lcd))` appears twice and `(1 - (var(--_ak-lcd) * var(--_ak-lcd)))` appears twice. Hoisting `1 - direction²` into a cached `var` via `layerMathDeclarations` would shrink the always-on calc. Not blocking — flagging again only because the diff still shows the duplication.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1493,30 +1537,41 @@ utility(
...
+  const getDeclarations = (
+    delta: string | number | VarProperty,
+    tokenValue: Value,
+  ) => [
+    set(inputs.layerLOffsetDelta, delta),
     set(
-      inputs.layerLOffsetDelta,
-      fn.mul(bareValue, vars.lightnessOffsetDirection),
+      vars.layerOffsetBaseL,
+      withUtilityTokenGate(getLayerL(inputs.layerLOffsetDelta), tokenValue),
```
diegohaz, line 1547: **Question — partial state alignment.** This new state-offset path uses `getLayerL` (floor only), so `ak-state-offset-*` gets the forbidden-range escape but is *not* clamped to `ak-layer-min-*`/`ak-layer-max-*`. That looks deliberate per the new `getLayerOffsetL` JSDoc.

What is less clear is that `state-push-*` (input.ts:1582) and `state-lighten-*`/`state-darken-*` (input.ts:1404-1414) still call `getLayerL` directly with no `getLayerOffsetL` wrapper — they receive neither the new clamp *nor* the forbidden-range escape, even after this commit. The changeset now reads "aligned `ak-state-*` and `ak-state-offset-*`", which can be read as if the whole state family was aligned. If the carve-out for push/lighten/darken is intentional, a brief note in the changeset clarifying the asymmetry would help consumers reason about it.

diegohaz (commented):
Re-review at head `56aa8e5c`. Three new commits since the last review (`556fca3b`):

- **f3438c8e Trim layer lightness sandbox coverage** — addresses the snapshot bloat. The new behavior is now exercised under a single `layer limits` subgroup with focused cases (`offset min`, `push min arbitrary`, `push max`, `offset max bounds`, `state offset`, `state push`, `contrast push`, etc.) instead of the ~30-cell expansion. Snapshot diffs went from ~6.3k to ~3.2k lines. Coverage of the meaningful edge cases is preserved.
- **69746901 Align state offset lightness escape** — addresses the state asymmetry question. `ak-state-*` and `ak-state-offset-*` now route through `getLayerOffsetL` via new `vars.layerOffsetBaseL`/`layerOffsetDelta` and `inputs.layerLOffsetL`, mirroring the idle pipeline. The changeset was updated to describe the state alignment. Note the deliberate choice: state callers pass `getLayerL(...)` (floor only) rather than `getLimitedLayerL(...)`, so state overlays still escape the forbidden mid-luminance range but are *not* clamped to `ak-layer-min-*`/`ak-layer-max-*` — consistent with what the new JSDoc on `getLayerOffsetL` describes ("callers can apply floor/min/max adjustments before the forbidden-range jump").
- **56aa8e5c Document layer offset cached delta** — addresses the JSDoc nit on `getLayerOffsetL`. Parameters renamed `limitedLightness`/`limitedDelta` → `baseLightness`/`baseDelta` and the docstring now spells out the `baseDelta = baseLightness - l` invariant and the caching rationale.

Local `pnpm tsc` passes. The PR is in good shape — two minor leftovers, both inline below.

The `ak-layer-push-*` lightness side effect noted in the previous review (e.g., dark-high-contrast "layer push" cell at `oklch(0.9 ...)` instead of the previous `oklch(1 ...)`) still stands and is preserved across the refactor — worth a one-line mention in the changeset if it is the intended new behavior.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -546,6 +546,10 @@ const layerMathVars = {
   offsetDirectionToLight: _ak.var("odtl"),
   forbiddenBandWidth: _ak.var("bw"),
   forbiddenEntryBoundary: _ak.var("eb"),
+  layerIdleLimitedL: _ak.prop("lill", { initial: l }),
+  layerIdleLimitedDelta: _ak.prop("lild", { initial: 0 }),
+  layerOffsetBaseL: _ak.prop("lobl", { initial: l }),
+  layerOffsetDelta: _ak.prop("lodl", { initial: 0 }),
```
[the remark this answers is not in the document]
diegohaz: Fixed in 4432c8382 - updated the `layerMathVars` comment to say the values are resolved by layer/state pipeline utilities, which covers both the `ak-layer` and `state-*` assignment sites.

diegohaz:
**Fixed**
- 4432c8382 (https://github.com/ariakit/ariakit/pull/5945#discussion_r3212941806): Clarified that `layerMathVars` values are resolved by the layer/state pipeline utilities, not only by `ak-layer`.

PR #5947, opened 2026-05-09
Fix layer push lightness bounds

diegohaz:
## Motivation

`ak-layer-push-*` could still overshoot `ak-layer-min-*` and `ak-layer-max-*` even when those bounds already placed the pushed lightness outside the forbidden mid-luminance range. For example, `ak-layer-max-[0.8] ak-layer-push-100` on a dark layer should stay at `0.8` in normal contrast, but the push logic treated crossing the forbidden range as a reason to add the full forbidden-band jump anyway.

The fix also needed to preserve the model where layer min/max constrain the pushed target before the final forbidden-range safety adjustment. Cases like `ak-layer-min-20 ak-layer-push-[0.25]` should push from the bounded lightness.

A previous implementation of that behavior caused a large dev-server performance regression in the `ariakit-tailwind` sandbox, so this PR also keeps the generated `ak-layer-push-*` work as small as possible.

## Solution

Update the shared push lightness resolver so it applies the forbidden-range jump only when the target lightness actually lands inside the forbidden interval. Targets that are already outside the interval are left untouched, preserving the model where layer min/max constrain the candidate before the final forbidden-range safety adjustment.

When a push target lands inside the forbidden interval, it snaps to the boundary on the other side of the band according to the push direction: lighter pushes escape through the upper boundary and darker pushes escape through the lower boundary.

Layer push now resolves from the original layer lightness plus any layer offset, then applies min/max and the forbidden-range adjustment. `ak-layer-contrast` composes with the bounded pushed layer target instead of ignoring the pushed lightness.

Push activation is now based on utility presence rather than a nonzero push amount. This means `ak-layer-push-0`, `ak-layer-push-[0]`, and tiny values like `ak-layer-push-0.25` all use the same push path. The `0` value only makes the push distance zero; it does not turn off forbidden-range escape or idle push contrast-bias direction. The shared resolver also gives `ak-state-push-0` the same forbidden-range behavior.

For performance, `ak-layer-push-*` now sets a small presence flag and resolved lightness instead of carrying the previous zero/arbitrary offset blend. The dedicated `ak-layer-push-0` no-op utility was removed so numeric and arbitrary zero values share the same generated path.

The sandbox includes focused arbitrary min/max push cases, zero layer push cases, a zero state push regression case, and computed-style snapshots covering light, dark, and high-contrast modes.

## Verification

- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep "computed styles" -u`
- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep "computed styles"`
- Fresh dev-server perf comparison against the current branch baseline after the final sandbox row: page load total `3276.6ms -> 3250.5ms (-1%)`; toggle total `2951.3ms -> 1577.5ms (-47%)`; toggle style recalculation `2911.5ms -> 1551.9ms (-47%)`
- `pnpm lint-fix` (completed with existing underscore warnings)
- `git diff --check`


packages/ariakit-tailwind/src/input.ts
```diff
@@ -1172,19 +1179,36 @@ function getBaseDeclarations(sourceColor: string | VarProperty) {
...
+function getLayerIdleContrastBiasDirection() {
+  const pushEnabled = fn.binary(vars.layerIdlePushValue);
+  const pushDirection = fn.sub(
+    fn.double(vars.layerIdlePushDirectionToLight),
+    1,
+  );
+  return fn.add(
+    fn.mul(vars.lightnessOffsetDirection, fn.sub(1, pushEnabled)),
+    fn.mul(pushDirection, pushEnabled),
+  );
```
diegohaz, line 1176: This blend uses `a*(1-x) + b*x`, so `pushEnabled` (and therefore `clamp(0, var(--_ak-lipv) * 1000000, 1)`) ends up duplicated in the generated CSS — see `--_ak-lcb` in `output.css`:

```css
--_ak-lcb: calc((var(--_ak-ct) * -1) * 0.3334 * ((var(--_ak-lod) * (1 - clamp(0, (var(--_ak-lipv) * 1000000), 1))) + (((var(--_ak-lipdtl) * 2) - 1) * clamp(0, (var(--_ak-lipv) * 1000000), 1))));
```

`getResolvedLightnessOffset` already uses the equivalent `a + x*(b-a)` form for the same reason and explains why in a comment (`input.ts:365`: "Using a + x*(b-a) instead of a*(1-x) + b*x so shouldFlip appears once"). Applying the same shape here drops one multiplication and one `clamp` from `--_ak-lcb`, which is now evaluated for every `ak-layer` element — not just pushed ones.

```suggestion
function getLayerIdleContrastBiasDirection() {
  const pushEnabled = fn.binary(vars.layerIdlePushValue);
  const pushDirection = fn.sub(
    fn.double(vars.layerIdlePushDirectionToLight),
    1,
  );
  // Same a + x*(b-a) rearrangement used in getResolvedLightnessOffset so
  // pushEnabled appears once in the generated CSS instead of twice.
  return fn.add(
    vars.lightnessOffsetDirection,
    fn.mul(pushEnabled, fn.sub(pushDirection, vars.lightnessOffsetDirection)),
  );
}
```

diegohaz (commented):
## PR review — initial review

**Baseline:** `merge-base origin/main HEAD` (at `7683d20`).

### What I checked
- New `getPushL` semantics — verified that the simplified "only jump when the bounded base lightness lands inside the forbidden range, then snap to the boundary picked by `directionToLight`" matches the changeset and the sandbox additions (`push min target low/high`, `push max arbitrary`, `push min arbitrary low`, etc.). Crossing-detection removal is sound: a target that already exits the band is by construction safe and should not be moved again.
- `directionToLight` wiring for both `layer-push-*` and `state-push-*`: the formula `binary(lipv * lod)` agrees with the explicit `@variant ak-light: 0` / `@variant ak-dark: 1` overrides in the standard cases, and the variant overrides correctly handle transitional `--_ak-ls` states where the container query won't fire.
- Push composing with offset: `layerIdlePushBaseL` now factors in `inputs.layerIdleLOffset` before applying the directional push, so `ak-layer-offset-* + ak-layer-push-*` now stacks (the new `push zero offset` cell exercises the no-op edge case).
- Layer-contrast composing with push: `layerIdle` now feeds `getContrastL(l, …)` reading from `var(--_ak-lipc, --_ak-lio)`, so `ak-layer-contrast` derives its parent-directed target from the bounded pushed lightness. Behavior parity when contrast is inactive is preserved (the bias term collapses to the previous form when `lipv = 0`). The `contrast push` sandbox cell exercises this.
- New `@property` registrations (`--_ak-lipdtl`, `--_ak-lipol`, `--_ak-lipc`, `--_ak-lpdtl`) and the build output: ran `pnpm -F @ariakit/tailwind build` — generated `output.css` matches the committed file. The earlier Copilot threads about `--_ak-lipc` missing `initial-value` were correctly resolved: `_ak.prop()` registers `syntax: "*"`, and the universal syntax explicitly allows omitting `initial-value`.
- Snapshots: spot-checked the recomputed `light` / `dark` / `*-high-contrast` push values against the new `getPushL` math and found the changes consistent with the documented behavior shift.

### Findings
One inline suggestion on `getLayerIdleContrastBiasDirection` to use the same `a + x*(b-a)` blend rearrangement that `getResolvedLightnessOffset` already documents at `input.ts:365`. As-written, the duplicated `binary` term shows up twice in `--_ak-lcb`, and that expression is now evaluated for every `ak-layer` element (not just pushed ones), so the redundancy is on the static layer rule rather than the per-utility rule.

### Areas not deeply validated
- I did not run `pnpm -F site test` against this branch (snapshot test runner). Visual snapshot deltas for the four `__snapshots__/*` files were inspected by hand for representative push values, not all rows.
- I did not benchmark the per-element CSS-engine cost change from moving `--_ak-lipv` / `--_ak-lipdtl` into the static `ak-layer` rule vs. the per-utility savings on `ak-layer-push-*`. The trade-off looks favorable for the common case (most pages have many more `ak-layer` instances than distinct `ak-layer-push-N` rules), but I'm flagging that I didn't measure it.

Overall the change reads as a correctness fix with a sensible ergonomic improvement (offset/contrast composition) and a small-but-real perf knob; the math, tests, and changeset are aligned.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1434,19 +1455,49 @@ utility(
...
+    ),
   ),
 );
 
+utility(
+  "layer-push-0",
+  set(vars.layerIdlePushOffsetL, getLayerIdleOffsetL()),
+  set(vars.layerIdlePushResolvedL, vars.layerIdlePushOffsetL),
+);
+
```
diegohaz: I wonder if this is really necessary. Is this just for perf? What's the cost of removing it? I don't think we should need it if it doesn't change the behavior. If it's only for perf, most people probably wouldn't pass `0` to it anyway, so I think we can avoid it.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 34d495319 - rewrote `getLayerIdleContrastBiasDirection` to use the equivalent `a + x * (b - a)` form, so the generated `--_ak-lcb` expression now includes the push-enabled clamp only once.

diegohaz:
**Fixed**

- `34d495319` ([reply](https://github.com/ariakit/ariakit/pull/5947#discussion_r3214404541)): rewrote `getLayerIdleContrastBiasDirection` to use `a + x * (b - a)`, removing the duplicate push-enabled clamp from the generated `--_ak-lcb` expression.

**Verification**

- `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep "computed styles"`
- Fresh dev-server package-only perf comparison: page load total `3203.1ms -> 3255.9ms (+2%)`; toggle total `1675.9ms -> 1561.2ms (-7%)`; toggle style recalc `1655.8ms -> 1537.5ms (-7%)`
- `pnpm lint-fix` (existing underscore warnings only)
- `git diff --check`

packages/ariakit-tailwind/src/input.ts
diegohaz: This is necessary for behavior, not just perf. I tried removing it locally and reran `SITE_PORT=4322 NEXTJS_PORT=3001 pnpm -F site test-chrome ariakit-tailwind --grep "computed styles"`; the exact `ak-layer-offset-40 ak-layer-push-0` row regressed while `ak-layer-push-[0]` still matched the offset row. For example, in light mode `push zero offset` changed from `oklch(0.7251_0.0011_197.14)` to `oklch(0.5498_0.0011_197.14)`. So the exact zero utility is preserving no-op-with-offset semantics for the canonical `0` token.

diegohaz:
**Invalid**

- `ak-layer-push-0` remains necessary ([reply](https://github.com/ariakit/ariakit/pull/5947#discussion_r3214406955)): removing the exact zero utility changes the existing `ak-layer-offset-40 ak-layer-push-0` snapshot output, so this is preserving canonical zero-push behavior rather than only optimizing perf.

PR #5967, opened 2026-05-14
Add negated Ariakit Tailwind layer variants

diegohaz:
## Motivation

Tailwind v4 currently generates invalid CSS when its automatic `not-*` variant is applied to custom variants backed by container style queries. This affects the `@ariakit/tailwind` layer appearance variants such as `ak-dark`, where `not-ak-dark:*` can be emitted as `@container style(...) not` instead of a valid negated style query.

## Solution

Add explicit `not-ak-*` custom variants for the layer appearance variants so consumers get valid `@container not style(...)` output without relying on Tailwind's automatic negation for these style queries.

This includes `not-ak-dark`, `not-ak-light`, `not-ak-dark-high`, `not-ak-dark-low`, `not-ak-light-low`, and `not-ak-light-high`, regenerates the published Tailwind input CSS, documents the negated variants, and adds a patch changeset.

## Verification

- `pnpm lint-fix`
- Tailwind v4.3.0 smoke compile for all six `not-ak-*` layer variants
- `pnpm tsc`
- `pnpm test`

### DIFF
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -776,15 +776,39 @@ const dark = createVariant(
   at.container(fn.style(vars.layerScheme, "oklch(1 0 0)"), set("@slot")),
 );
 
+const notDark = createVariant(
+  "not-ak-dark",
+  at.container(
+    `not ${fn.style(vars.layerScheme, "oklch(1 0 0)")}`,
+    set("@slot"),
+  ),
+);
+
 const light = createVariant(
   "ak-light",
   at.container(fn.style(vars.layerScheme, "oklch(0 0 0)"), set("@slot")),
+);
+
+const notLight = createVariant(
+  "not-ak-light",
+  at.container(
+    `not ${fn.style(vars.layerScheme, "oklch(0 0 0)")}`,
+    set("@slot"),
+  ),
 );
 
 const darkHigh = createVariant(
   "ak-dark-high",
   at.container(
     fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_DARK_HIGH })),
+    set("@slot"),
+  ),
+);
+
+const notDarkHigh = createVariant(
+  "not-ak-dark-high",
+  at.container(
+    `not ${fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_DARK_HIGH }))}`,
     set("@slot"),
   ),
 );
@@ -797,6 +821,14 @@ const darkLow = createVariant(
   ),
 );
 
+const notDarkLow = createVariant(
+  "not-ak-dark-low",
+  at.container(
+    `not ${fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_DARK_LOW }))}`,
+    set("@slot"),
+  ),
+);
+
 const lightLow = createVariant(
   "ak-light-low",
   at.container(
@@ -805,10 +837,26 @@ const lightLow = createVariant(
   ),
 );
 
+const notLightLow = createVariant(
+  "not-ak-light-low",
+  at.container(
+    `not ${fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_LIGHT_LOW }))}`,
+    set("@slot"),
+  ),
+);
+
 const lightHigh = createVariant(
   "ak-light-high",
   at.container(
     fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_LIGHT_HIGH })),
+    set("@slot"),
+  ),
+);
+
+const notLightHigh = createVariant(
+  "not-ak-light-high",
+  at.container(
+    `not ${fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_LIGHT_HIGH }))}`,
     set("@slot"),
   ),
 );
@@ -2367,11 +2415,17 @@ export const input = [
   root,
   disabled,
   dark,
+  notDark,
   light,
+  notLight,
   darkHigh,
+  notDarkHigh,
   darkLow,
+  notDarkLow,
   lightLow,
+  notLightLow,
   lightHigh,
+  notLightHigh,
   ...utilities,
   ...Object.values(vars),
   ...Object.values(inputs),
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -778,10 +778,7 @@ const dark = createVariant(
 
 const notDark = createVariant(
   "not-ak-dark",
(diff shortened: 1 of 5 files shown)


diegohaz:
@codex

packages/ariakit-tailwind/src/input.ts
```diff
@@ -813,6 +853,14 @@ const lightHigh = createVariant(
   ),
 );
 
+const notLightHigh = createVariant(
+  "not-ak-light-high",
+  at.container(
+    `not ${fn.style(vars.layerBand, fn.oklch({ l: BAND_LEVEL_LIGHT_HIGH }))}`,
+    set("@slot"),
+  ),
+);
```
diegohaz, line 856: Each negated variant manually re-spells the same `style(...)` predicate as its positive sibling and prepends `not ` as a raw string. Six pairs gives twelve definitions in lockstep, and a typo or value drift in either copy (for example tweaking a `BAND_LEVEL_*` constant or changing `vars.layerScheme`/`vars.layerBand` to a different property later) wouldn't be caught by the existing snapshot tests because none of the sandbox markup currently uses `not-ak-*`.

The codebase already has a `supports.not(query, ...children)` helper in `lib.ts` (lines 456–459) that wraps `@supports not`. A symmetric `container.not` helper would mirror that precedent and remove the `\`not ${...}\`` string concatenation here. With that helper, each pair could read:

```ts
const dark = createVariant(
  "ak-dark",
  at.container(fn.style(vars.layerScheme, "oklch(1 0 0)"), set("@slot")),
);

const notDark = createVariant(
  "not-ak-dark",
  at.container.not(fn.style(vars.layerScheme, "oklch(1 0 0)"), set("@slot")),
);
```

That keeps the existing variable structure and just trades the string concat for a typed helper. A tighter alternative — if you'd rather kill the per-pair query duplication too — is a small factory that returns both variants from a single query, so the positive and negative forms can never drift apart:

```ts
function createLayerAppearanceVariant(name: string, query: string) {
  return [
    createVariant(name, at.container(query, set("@slot"))),
    createVariant(`not-${name}`, at.container.not(query, set("@slot"))),
  ];
}

const [dark, notDark] = createLayerAppearanceVariant(
  "ak-dark",
  fn.style(vars.layerScheme, "oklch(1 0 0)"),
);
```

Either shape would also avoid a subtle gotcha with the current string form: `` `not ${fn.style(...)}` `` relies on `fn.style` always returning a parenthesized expression, which it does today but isn't enforced anywhere.

diegohaz (commented):
Initial review of the PR head `c230d29`, diffed against `origin/main` (`a518ea1b`).

## Scope reviewed
- `packages/ariakit-tailwind/src/input.ts` — six new `not-ak-*` `createVariant` declarations plus their entries in the exported `input` array.
- `packages/ariakit-tailwind/src/output.css` — regenerated `@custom-variant not-ak-*` blocks.
- `packages/ariakit-tailwind/readme.md` — variant table addendum and updated caveat note.
- `.changeset/300-tailwind-not-layer-variants.md` — patch entry for `@ariakit/tailwind`.

I also re-read the surrounding `createVariant` helpers, `at.container` and `supports.not` in `packages/ariakit-tailwind/src/lib.ts`, and the `@property` declarations for `--_ak-ls`/`--_ak-lbd` to sanity-check that the generated `@container not style(...)` queries are well-formed for these registered custom properties.

## What I ran locally
- `pnpm -F @ariakit/tailwind build` — `output.css` regenerated cleanly and matches the version checked in.
- `pnpm tsc` — passes (0 errors, 0 warnings).
- Spot-checked the generated `@container not style(--_ak-*: oklch(...))` blocks against the CSS container-query grammar — `not <container-query-in-parens>` with `style()` as the parenthesized form is valid, so the emitted CSS is correct.

## Findings
One non-blocking maintainability suggestion as an inline comment on `input.ts`: the six positive/negative pairs duplicate the inner `style(...)` predicate and inline `\`not ${...}\`` as a raw string. The repo already has a `supports.not` helper in `lib.ts`, so a symmetric `at.container.not` helper (or a small `createLayerAppearanceVariant` factory) would remove the string concat and prevent the two halves from drifting if any of the threshold values or queried vars changes later.

No correctness, behavior, performance, security, or backward-compatibility issues found. The changeset follows the simple-entry format from the shared changeset skill, and the doc caveat about wrapping in an `ak-layer` reads correctly given the registered initial values of `--_ak-ls`/`--_ak-lbd`.

## Not validated
I did not independently reproduce the Tailwind v4 bug (`@container style(...) not`) cited as the PR's motivation — the fix here is explicit `not-ak-*` variants, which is the right shape regardless of whether/when Tailwind's automatic negation gets fixed upstream, so I focused on the variants themselves.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in bac9ae553: added at.container.not and switched the negated layer variants to use it.

diegohaz:
**Fixed**
- bac9ae553 (https://github.com/ariakit/ariakit/pull/5967#discussion_r3244056163): Added `at.container.not` and switched the negated layer variants to use it.

packages/ariakit-tailwind/src/input.ts
diegohaz: Confirmed in `bac9ae553`. The new `at.container.not` helper mirrors `supports.not` one-for-one, and `output.css` regenerates byte-identically against `main`, so the only behavioral change is at the source level — call sites read `at.container.not(query, ...)` instead of `at.container(\`not ${query}\`, ...)`. `pnpm -F @ariakit/tailwind build` and `pnpm tsc` both pass.

diegohaz (commented):
Re-review of `bac9ae553`, diffed against the prior reviewed head `c230d297` (one new commit, `Add container not helper`).

## Scope reviewed
- `packages/ariakit-tailwind/src/lib.ts` — new `containerFn` + `container.not` wrapping, mirroring the existing `supports`/`supports.not` pair.
- `packages/ariakit-tailwind/src/input.ts` — the six negated layer variants now call `at.container.not(query, set("@slot"))` instead of inlining a `\`not ${query}\`` string.
- `packages/ariakit-tailwind/src/output.css` — regenerates byte-identically against `main`, so no behavior change in the emitted CSS.

## What I ran locally
- `pnpm -F @ariakit/tailwind build` — `output.css` unchanged after rebuild.
- `pnpm tsc` — passes (0 errors, 0 warnings).

## Findings
No new findings. The previous suggestion was addressed exactly as outlined: `at.container.not` now exists, the six negated variants use it, and the helper shape matches `supports.not` so the codebase keeps one consistent pattern for negated at-rule queries. Replied on the original thread to acknowledge.

PR #5974, opened 2026-05-16
Upgrade app to Astro 6 and Cloudflare adapter 13

diegohaz:
## Motivation

Astro 6 and `@astrojs/cloudflare` 13 change how Cloudflare Workers runtime bindings are exposed and how deployable Worker configs are generated. The app also needs to move to zod 4 and keep the Cloudflare runtime wiring compatible with CI, preview deploys, production deploys, Clerk, Stripe, and KV-backed app features.

## Solution

Upgrade the app runtime dependencies to Astro 6, `@astrojs/cloudflare` 13, zod 4, and the matching Wrangler patch release. Configure the Cloudflare adapter to use the root Wrangler config and deploy the generated `app/dist/server/wrangler.json` config from CI.

Move Cloudflare bindings to the v13 `cloudflare:workers` model in runtime files that support top-level imports, and pass `env` into KV, Stripe, Clerk, admin, checkout, and Plus helpers that need bindings or secrets. Middleware keeps a cached dynamic runtime import because top-level `cloudflare:workers` is not safe there during Astro build/prerender paths.

Keep preview Workers free of KV namespace resources by removing preview KV bindings, using the in-memory Astro session driver only for preview builds, and letting public Plus checkout pricing/session paths return cleanly when preview has no KV binding. Production builds keep the Cloudflare KV-backed session binding and the existing `SESSION`, `PLUS`, `EVENTS`, and `ADMIN` resources.

Also adjust the app build for Astro 6 compatibility by moving shared content component exports out of an `.astro` module, updating Shiki/Astro content typings, restoring the Clerk build shim for environments without Clerk tokens, and removing Renovate pins that held Astro and the Cloudflare adapter to their previous majors.

The OG image workflow now builds the generated Cloudflare preview Worker with `OG_IMAGE_GENERATE=true` and serves it with Wrangler preview instead of Astro dev, so the screenshots exercise the same `cloudflare:workers` runtime path as deployable builds. The generator defers PNG writes until navigation finishes, rejects non-OK page loads, and fails if Astro renders the `FailedToLoadModuleSSR` error page instead of committing error screenshots.

## Review comments

- Restored `dummyClerkIntegration()` for builds without Clerk tokens.
- Cached the middleware dynamic `cloudflare:workers` import and documented through implementation that middleware cannot use a top-level import with the current Node prerender path.
- Kept a typed `*.astro` module declaration because TypeScript needs it for the TS content component export module.
- Removed the placeholder return from the admin `setPrice` action.
- Kept `app/package.json` preview pointed at the generated `dist/server/wrangler.json`, which is the deployable Worker config produced by the v13 adapter.
- Rechecked review threads after the follow-up fixes; the remaining unresolved threads all have replies and no new actionable review comments were open at the time of the latest push.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `CLOUDFLARE_ENV=preview pnpm -F app run build-lite`
- `CLOUDFLARE_ENV=preview OG_IMAGE_GENERATE=true pnpm -F app run build-lite`
- `pnpm build-app-lite`
- `pnpm -F app run test-chrome`
- `BASE_URL=http://127.0.0.1:4325 pnpm -F app run og-image-generate` against `pnpm -F app exec wrangler dev --config dist/server/wrangler.json --port 4325 --ip 127.0.0.1`
- `pnpm exec wrangler deploy --dry-run --config app/dist/server/wrangler.json` for both production-generated and preview-generated configs

`pnpm peers check` still reports existing unrelated peer warnings for `tailwindcss-animate`, `@vitest/coverage-v8`, and React 18 types.


app/src/middleware.ts
```diff
@@ -15,6 +15,15 @@ import { unauthorized } from "./lib/response.ts";
...
 
+async function getRuntimeEnv() {
+  try {
+    const { env } = await import("cloudflare:workers");
+    return env;
+  } catch {
+    return null;
+  }
+}
+
```
diegohaz: Middleware is an astro file. Can't we import `cloudflare:workers` directly at the top-level?

app/env.d.ts
```diff
@@ -22,12 +22,33 @@ declare module "#app/styles/styles.json" {
   export default styles;
 }
 
+declare module "*.astro" {
+  const Component: any;
+  export default Component;
+}
```
diegohaz, line 28: Why is this needed?

app/env.d.ts
```diff
@@ -22,12 +22,33 @@ declare module "#app/styles/styles.json" {
...
+    readonly PUBLIC_STRIPE_PUBLISHABLE_KEY?: string;
+    readonly STRIPE_SECRET_KEY?: string;
+    readonly STRIPE_WEBHOOK_SECRET?: string;
+    readonly ADMIN_ORG_ID?: string;
+  }
+}
+
+declare module "cloudflare:workers" {
+  export const env: Cloudflare.Env;
+}
```
diegohaz, line 51: Doesn't astro automatically do this?

app/package.json
```diff
@@ -20,7 +20,7 @@
     "build": "astro build",
     "build-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run build",
     "build-min": "cross-env DISABLE_SYNTAX_HIGHLIGHTING=true pnpm run build-lite",
-    "preview": "wrangler dev",
+    "preview": "wrangler dev --config dist/server/wrangler.json",
```
diegohaz, line 23: Why is this explicit config needed?

app/astro.config.ts
```diff
@@ -87,20 +84,18 @@ export default defineConfig({
...
+      appearance: {
+        variables: {
+          fontSize: "1rem",
+        },
+        layout: {
+          logoPlacement: "none",
+          showOptionalFields: false,
+        },
+      },
+    }),
```
diegohaz: We need to support cases where the Clerk tokens aren't set. That's what `dummyClerkIntegration()` was doing before.

app/src/middleware.ts
```diff
@@ -23,8 +32,9 @@ function isPublicRoute(url: URL) {
 export async function onRequest(context: APIContext, next: MiddlewareNext) {
   const { action } = getActionContext(context);
   const isAdminAction = action?.name.startsWith("admin");
+  const runtimeEnv = await getRuntimeEnv();
 
-  if (!import.meta.env.PUBLIC_CLERK_PUBLISHABLE_KEY) {
+  if (!runtimeEnv?.PUBLIC_CLERK_PUBLISHABLE_KEY) {
```
diegohaz, line 37: Why we can't keep using `import.meta.env`?

app/src/middleware.ts
```diff
@@ -35,11 +45,12 @@ export async function onRequest(context: APIContext, next: MiddlewareNext) {
...
-
-  if (isAdminAction && !(await isAdmin(context))) {
-    return unauthorized();
-  }
+  const response = await clerk(context, async () => {
+    if (isAdminAction && !(await isAdmin(context, runtimeEnv))) {
+      return unauthorized();
+    }
+    return next();
+  });
```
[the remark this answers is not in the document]
diegohaz: Resolved in 37b4c5b16 - `getRuntimeEnv()` now caches the `cloudflare:workers` import promise at module scope and `runtimeEnv` is narrowed by the early return before `isAdmin(...)` runs.

app/src/middleware.ts
diegohaz: I tested this in 37b4c5b16: a top-level middleware import breaks `CLOUDFLARE_ENV=preview pnpm -F app run build-lite` while prerendering `/404` under Node with `ERR_UNSUPPORTED_ESM_URL_SCHEME` for `cloudflare:`. So middleware is the exception here and keeps a cached dynamic import.

app/env.d.ts
diegohaz: This is needed because `app/src/components/content-components.ts` imports `.astro` components from a plain TS module. I verified removing the declaration makes `pnpm tsc` fail on those imports, so 37b4c5b16 keeps it and narrows the type to `AstroComponentFactory` instead of `any`.

app/env.d.ts
diegohaz: Astro did not provide the `cloudflare:workers` module type in this setup. I checked the generated `worker-configuration.d.ts` and installed Astro/Cloudflare declaration files, and `pnpm tsc` needs this declaration for the top-level imports added in the PR.

app/package.json
diegohaz: This points Wrangler at the Cloudflare adapter v13 generated Worker config. The root config is used by Astro at build time, but local preview/deploy run from `app/`, so `wrangler dev --config dist/server/wrangler.json` uses the same deployable config that CI uploads.

app/astro.config.ts
diegohaz: Fixed in 37b4c5b16 - restored `dummyClerkIntegration()` and made the Clerk integration conditional on `PUBLIC_CLERK_PUBLISHABLE_KEY` again.

app/src/middleware.ts
diegohaz: With adapter v13, Worker runtime secrets/bindings need to come from `cloudflare:workers` instead of build-time `import.meta.env` so preview/production use the actual Worker env. Middleware still falls back to no Clerk when that runtime env is unavailable in Node/dev paths.

diegohaz:
**Fixed**

- 37b4c5b16: Removed preview KV namespace resources, scoped Astro LRU sessions to preview only, and guarded public Plus checkout pricing/session paths when preview has no KV binding.
- 37b4c5b16: Restored the Clerk build shim for missing Clerk tokens and removed the admin `setPrice` placeholder return.
- 37b4c5b16: Cached the middleware runtime env import and deferred OG image writes until after screenshots complete.

**Answered**

- Replied to each unresolved review thread with the implementation detail or verification result, including why middleware keeps a dynamic `cloudflare:workers` import and why the generated preview Wrangler config is used for local preview/deploy.

PR #5986, opened 2026-05-17
Fix nested Next.js preview IDs

diegohaz:
## Motivation

Nested Next.js preview routes are handled by the SSR redirect path, but the preview ID matcher only recognizes root preview IDs. This means a URL such as `/react/previews/sandbox/counter-nextjs/` is not identified as a Next.js preview and cannot be redirected to the Next.js app.

## Solution

Update `getNextjsPreviewId` to capture the final Next.js preview segment after optional nested path segments, and add focused Vitest coverage for root, nested, no-trailing-slash, and non-Next.js preview paths.

### DIFF
--- a/app/src/lib/nextjs-test.ts
+++ b/app/src/lib/nextjs-test.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+
+import { expect, test } from "vitest";
+import { getNextjsPreviewId } from "./nextjs.ts";
+
+test("getNextjsPreviewId handles root preview ids", () => {
+  expect(getNextjsPreviewId("/react/previews/tab-nextjs/")).toBe("tab-nextjs");
+  expect(getNextjsPreviewId("/react/previews/menu-nextjs-app-router")).toBe(
+    "menu-nextjs-app-router",
+  );
+});
+
+test("getNextjsPreviewId handles nested preview ids", () => {
+  expect(getNextjsPreviewId("/react/previews/sandbox/counter-nextjs/")).toBe(
+    "counter-nextjs",
+  );
+  expect(getNextjsPreviewId("/react/previews/sandbox/counter-nextjs")).toBe(
+    "counter-nextjs",
+  );
+});
+
+test("getNextjsPreviewId ignores non-Next.js preview ids", () => {
+  expect(getNextjsPreviewId("/react/previews/disclosure/")).toBeNull();
+  expect(getNextjsPreviewId("/react/previews/disclosure/nested/")).toBeNull();
+  expect(getNextjsPreviewId("/react/previews/nextjs/nested/")).toBeNull();
+});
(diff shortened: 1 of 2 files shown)


diegohaz (commented):
Initial review. Scope: the full diff (`app/src/lib/nextjs.ts`, `app/src/lib/nextjs-test.ts`), `getNextjsPreviewId` callers (`app/src/pages/[...nextjs].astro`), and the surrounding preview routing (`app/src/pages/[...preview].astro`, `app/src/pages/previews.ts`, `app/src/content.config.ts`, `app/src/components/code-block-tabs.astro`, `app/src/components/code-block.react.tsx`). Baseline: merge-base with `main` (`17e08491d`), single new commit `9ae0e79`.

### What I checked

- Ran `pnpm vitest run app/src/lib/nextjs-test.ts` from the repo root — all 3 tests pass.
- `pnpm --filter app run check` (astro check) — 0 errors / 0 warnings.
- Compared old vs new regex against a wide range of paths (single nesting, multi-level nesting, two `nextjs` segments, double slashes, case variants, trailing-slash variants) to confirm the only behavioral change is that the leaf segment is now extracted from arbitrarily nested paths.

### Findings

No actionable inline issues. The regex change is a strict superset of the previous behavior: every path the old pattern accepted is still accepted with the same capture, and a new set of nested paths now resolve to the leaf segment as documented. The greedy `(?:.*\/)?` correctly produces the deepest `*nextjs*` segment when multiple intermediate slashes exist (e.g., `/react/previews/foo-nextjs/bar-nextjs/` → `bar-nextjs`), which matches the SSR redirect's intent.

### One observation worth confirming (non-blocking)

The PR description cites `/react/previews/sandbox/counter-nextjs/` as the motivating URL, but I couldn't find a code path that generates it today:

- `generateExampleId` in `app/src/content.config.ts:18-22` strips the leading `sandbox/`, so `sandbox/counter-nextjs/preview.mdx` becomes `entry.id = "counter-nextjs"`.
- `app/src/pages/[...preview].astro:22` and `app/src/pages/previews.ts:10` both emit `${framework}/previews/${entry.id}`, i.e. `/react/previews/counter-nextjs/`.
- `getPreviewUrl` in `app/src/components/code-block.react.tsx:65-73` uses the same `${example}` value, sourced from `source.name`, which I'd expect to be similarly de-prefixed.

The fix is still useful — it correctly handles nested entry IDs (e.g. a future `sandbox/sub/foo-nextjs/preview.mdx` would produce `entry.id = "sub/foo-nextjs"` → URL `/react/previews/sub/foo-nextjs/`, which the old regex would not have matched) and any externally-linked or hand-crafted nested URL. But if the trigger you observed was specifically `…/previews/sandbox/…`, it'd be worth confirming where that URL originated, since nothing in the current routing should be emitting it.

### Not validated

- Whether `source.name` (consumed by `code-block-tabs.astro:85` and indirectly by `getPreviewUrl`) can ever contain a `sandbox/` prefix in any code path I didn't trace — only checked the obvious generators.
- Did not exercise the redirect end-to-end in a running Astro/Cloudflare dev session; relied on unit-level coverage and reading the SSR catchall in `app/src/pages/[...nextjs].astro:19-30`.

diegohaz:
Closing this since the motivating URL was based on the wrong preview path shape. `app/src/sandbox/counter-nextjs/preview.mdx` maps to `/react/previews/counter-nextjs/`, with the `sandbox` source segment stripped from the preview ID, and the existing matcher on `main` already handles that URL.

The reusable guidance has been updated separately in `ariakit/skills#143` so future agent work uses the stripped preview URL.

PR #5987, opened 2026-05-17
Stabilize OG image writebacks

diegohaz:
## Motivation

The OG image workflow can write generated assets back to a branch after that branch has already advanced. That risks committing generated images from a stale checkout. The generator also takes a single screenshot after the initial navigation commit, which can capture blank or not-yet-rendered pages when the local app is still settling.

## Solution

Add a branch-tip verification step before the write-back action runs. When generated images change, the workflow fetches the current branch tip and compares it with the checkout `HEAD`; if the branch already advanced, the job skips the write-back before invoking the shared action. The workflow also passes that verified SHA into `commit-or-pr`, where the action re-fetches the branch and refuses to rebase generated output if the branch advanced after the outer verification step.

Make the OG image generator wait for expected page body text before capturing when an item has a title or framework label, support `OG_IMAGE_BASE_URL` for alternate local ports, retry blank screenshots, and reject screenshots that still look blank after the retry budget. The homepage has no item-specific body text, so it relies on the blank-screenshot retry guard instead of a generic document-title match.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/og-images.yml"); YAML.load_file(".github/workflows/commit-or-pr/action.yml"); puts "ok"'`
- `git diff --cached --check`
- `OG_IMAGE_BASE_URL=http://127.0.0.1:4322 pnpm -F app run og-image-generate`
- Pre-commit review gate with native reviewer and Cursor re-review

### DIFF
--- a/.github/workflows/og-images.yml
+++ b/.github/workflows/og-images.yml
@@ -95,6 +95,7 @@ jobs:
             exit 0
           fi
           echo "is_current=true" >> "$GITHUB_OUTPUT"
+          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
 
       - name: Commit or create PR
         if: >
@@ -110,6 +111,7 @@ jobs:
           event: ${{ github.event_name }}
           message: Update OG images
           head-branch: ${{ github.head_ref || github.ref_name }}
+          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
           new-branch: chore/update-og-images
           paths: app/public/og-image
           app-client-id: ${{ vars.ARIAKIT_CI_APP_CLIENT_ID }}
--- a/.github/workflows/og-images.yml
+++ b/.github/workflows/og-images.yml
@@ -77,10 +77,30 @@ jobs:
           fi
           echo "has_changes=true" >> "$GITHUB_OUTPUT"
 
-      - name: Commit or create PR
+      - name: Verify branch tip
+        id: branch
         if: >
           steps.patch.outputs.has_changes == 'true' &&
           env.WRITEBACK_ALLOWED == 'true'
+        shell: bash
+        run: |
+          set -euo pipefail
+          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
+          git fetch --no-tags origin "$branch"
+          local_sha="$(git rev-parse HEAD)"
+          remote_sha="$(git rev-parse FETCH_HEAD)"
+          if [ "$local_sha" != "$remote_sha" ]; then
+            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
+            echo "is_current=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "is_current=true" >> "$GITHUB_OUTPUT"
+
+      - name: Commit or create PR
+        if: >
+          steps.patch.outputs.has_changes == 'true' &&
+          env.WRITEBACK_ALLOWED == 'true' &&
+          steps.branch.outputs.is_current == 'true'
         uses: ./.github/workflows/commit-or-pr
         with:
           # This write-back step intentionally assumes PRs come from branches
--- a/app/src/pages/og-image/_generate-images.ts
+++ b/app/src/pages/og-image/_generate-images.ts
@@ -28,15 +28,35 @@ const MAX_SCREENSHOT_ATTEMPTS = 3;
 
 function isBlankImage(buffer: Buffer) {
   const image = PNG.sync.read(buffer);
-  const [red, green, blue, alpha] = image.data;
+  const pixels = image.data;
+  const red = pixels[0];
+  const green = pixels[1];
+  const blue = pixels[2];
+  const alpha = pixels[3];
+  if (red == null || green == null || blue == null || alpha == null) {
+    return false;
+  }
   let matchingPixels = 0;
   const totalPixels = image.width * image.height;
-  for (let i = 0; i < image.data.length; i += 4) {
+  if (!totalPixels) return false;
+  for (let i = 0; i < pixels.length; i += 4) {
+    const currentRed = pixels[i];
+    const currentGreen = pixels[i + 1];
+    const currentBlue = pixels[i + 2];
+    const currentAlpha = pixels[i + 3];
     if (
-      Math.abs(image.data[i]! - red!) <= 1 &&
-      Math.abs(image.data[i + 1]! - green!) <= 1 &&
-      Math.abs(image.data[i + 2]! - blue!) <= 1 &&
-      Math.abs(image.data[i + 3]! - alpha!) <= 1
+      currentRed == null ||
+      currentGreen == null ||
+      currentBlue == null ||
+      currentAlpha == null
+    ) {
+      continue;
+    }
+    if (
+      Math.abs(currentRed - red) <= 1 &&
+      Math.abs(currentGreen - green) <= 1 &&
+      Math.abs(currentBlue - blue) <= 1 &&
+      Math.abs(currentAlpha - alpha) <= 1
     ) {
       matchingPixels += 1;
     }
@@ -47,7 +67,7 @@ function isBlankImage(buffer: Buffer) {
 function getExpectedText(item: OGImageItem) {
   if (item.title) return item.title;
   if (item.framework) return getFramework(item.framework).label;
-  return "Ariakit";
+  return undefined;
 }
 
 function isWithinTolerance(newBuffer: Buffer, existingPath: string) {
@@ -99,10 +119,10 @@ async function waitForImageReady(page: Page, item: OGImageItem) {
   await page.waitForLoadState("load");
   const expectedText = getExpectedText(item);
   if (expectedText) {
-    await page.waitForFunction((text) => {
-      if (document.title.includes(text)) return true;
-      return document.body.textContent?.includes(text) ?? false;
-    }, expectedText);
+    await page.waitForFunction(
+      (text) => document.body.textContent?.includes(text) ?? false,
+      expectedText,
+    );
   }
   await page.evaluate(async () => {
     await document.fonts.ready;
--- a/app/src/pages/og-image/_generate-images.ts
+++ b/app/src/pages/og-image/_generate-images.ts
@@ -13,15 +13,42 @@ import pixelmatch from "pixelmatch";
 import type { Page } from "playwright";
 import { chromium } from "playwright";
 import { PNG } from "pngjs";
+import { getFramework } from "#app/lib/frameworks.ts";
 import type { OGImageItem } from "./api.ts";
 
-const BASE_URL = "http://localhost:4321";
+const BASE_URL = process.env.OG_IMAGE_BASE_URL ?? "http://localhost:4321";
 const PUBLIC_DIR = path.resolve(process.cwd(), "public");
 
 // Maximum fraction of pixels allowed to differ before the image is
 // considered changed. This avoids committing images that only differ
 // due to sub-pixel rendering or anti-aliasing across runs.
 const MAX_DIFF_PIXEL_RATIO = 0.001;
+const MAX_BLANK_PIXEL_RATIO = 0.995;
+const MAX_SCREENSHOT_ATTEMPTS = 3;
+
+function isBlankImage(buffer: Buffer) {
+  const image = PNG.sync.read(buffer);
+  const [red, green, blue, alpha] = image.data;
+  let matchingPixels = 0;
+  const totalPixels = image.width * image.height;
+  for (let i = 0; i < image.data.length; i += 4) {
+    if (
+      Math.abs(image.data[i]! - red!) <= 1 &&
+      Math.abs(image.data[i + 1]! - green!) <= 1 &&
+      Math.abs(image.data[i + 2]! - blue!) <= 1 &&
+      Math.abs(image.data[i + 3]! - alpha!) <= 1
+    ) {
+      matchingPixels += 1;
+    }
+  }
+  return matchingPixels / totalPixels >= MAX_BLANK_PIXEL_RATIO;
+}
+
+function getExpectedText(item: OGImageItem) {
+  if (item.title) return item.title;
+  if (item.framework) return getFramework(item.framework).label;
+  return "Ariakit";
+}
 
 function isWithinTolerance(newBuffer: Buffer, existingPath: string) {
   if (!fs.existsSync(existingPath)) return false;
@@ -68,8 +95,15 @@ async function getItemsToGenerate() {
   }
 }
 
-async function waitForImageReady(page: Page) {
+async function waitForImageReady(page: Page, item: OGImageItem) {
   await page.waitForLoadState("load");
+  const expectedText = getExpectedText(item);
+  if (expectedText) {
+    await page.waitForFunction((text) => {
+      if (document.title.includes(text)) return true;
+      return document.body.textContent?.includes(text) ?? false;
+    }, expectedText);
+  }
   await page.evaluate(async () => {
     await document.fonts.ready;
     await new Promise((resolve) => {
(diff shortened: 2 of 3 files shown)


app/src/pages/og-image/_generate-images.ts
```diff
@@ -13,15 +13,42 @@ import pixelmatch from "pixelmatch";
...
+  const image = PNG.sync.read(buffer);
+  const [red, green, blue, alpha] = image.data;
+  let matchingPixels = 0;
+  const totalPixels = image.width * image.height;
+  for (let i = 0; i < image.data.length; i += 4) {
+    if (
+      Math.abs(image.data[i]! - red!) <= 1 &&
+      Math.abs(image.data[i + 1]! - green!) <= 1 &&
+      Math.abs(image.data[i + 2]! - blue!) <= 1 &&
+      Math.abs(image.data[i + 3]! - alpha!) <= 1
```
diegohaz: **Avoid non-null assertions on indexed access.**

`tsconfig.base.json` has `noUncheckedIndexedAccess: true`, so `image.data[i]` is typed `number | undefined`. This function uses non-null assertions (`!`) on every pixel read and on the destructured `red`, `green`, `blue`, `alpha`. The shared code style (`ariakit-general-code-style` → TypeScript → indexing arrays) explicitly disallows this: "Do not use non-null assertions with indexed access". The fact that `red!` is needed at all is a tell that the destructuring shape doesn't carry the right type — a 0×0 PNG would actually make those values `undefined` at runtime, even if it's not expected here.

A rewrite without `!` and with an explicit guard for the zero-pixel case:

```ts
function isBlankImage(buffer: Buffer) {
  const image = PNG.sync.read(buffer);
  const pixels = image.data;
  const red = pixels[0];
  const green = pixels[1];
  const blue = pixels[2];
  const alpha = pixels[3];
  if (red == null || green == null || blue == null || alpha == null) {
    return false;
  }
  let matchingPixels = 0;
  const totalPixels = image.width * image.height;
  for (let i = 0; i < pixels.length; i += 4) {
    const r = pixels[i];
    const g = pixels[i + 1];
    const b = pixels[i + 2];
    const a = pixels[i + 3];
    if (r == null || g == null || b == null || a == null) continue;
    if (
      Math.abs(r - red) <= 1 &&
      Math.abs(g - green) <= 1 &&
      Math.abs(b - blue) <= 1 &&
      Math.abs(a - alpha) <= 1
    ) {
      matchingPixels += 1;
    }
  }
  return matchingPixels / totalPixels >= MAX_BLANK_PIXEL_RATIO;
}
```

app/src/pages/og-image/_generate-images.ts
```diff
@@ -68,8 +95,15 @@ async function getItemsToGenerate() {
...
-async function waitForImageReady(page: Page) {
+async function waitForImageReady(page: Page, item: OGImageItem) {
   await page.waitForLoadState("load");
+  const expectedText = getExpectedText(item);
+  if (expectedText) {
+    await page.waitForFunction((text) => {
+      if (document.title.includes(text)) return true;
+      return document.body.textContent?.includes(text) ?? false;
+    }, expectedText);
+  }
```
diegohaz, line 127: **`if (expectedText)` is always true — the guard is dead code.**

`getExpectedText` always returns a non-empty string: it falls through to the literal `"Ariakit"`. So this `if` is structurally a no-op and just adds noise.

A related, more subtle concern is that for *every* page on the site, `document.title` is either `"Ariakit"` (homepage) or `"<title> - Ariakit"`, so `document.title.includes("Ariakit")` is true the moment the SSR HTML is parsed. That means the homepage case (where `getExpectedText` returns `"Ariakit"`) skips through `waitForFunction` immediately without verifying any body content was rendered. The blank-image retry in `screenshotImage` is what actually catches a blank homepage — the text wait is doing no useful work for that item.

Minimal cleanup, drop the dead `if`:

```diff
-  const expectedText = getExpectedText(item);
-  if (expectedText) {
-    await page.waitForFunction((text) => {
-      if (document.title.includes(text)) return true;
-      return document.body.textContent?.includes(text) ?? false;
-    }, expectedText);
-  }
+  await page.waitForFunction((text) => {
+    if (document.title.includes(text)) return true;
+    return document.body.textContent?.includes(text) ?? false;
+  }, getExpectedText(item));
```

If you want the wait to actually wait for *body* content on the homepage (so the text-wait is doing useful work for every item), one option is to drop the `"Ariakit"` fallback and skip the wait when there's no per-item signal:

```ts
function getExpectedText(item: OGImageItem): string | undefined {
  if (item.title) return item.title;
  if (item.framework) return getFramework(item.framework).label;
  return undefined;
}
```

and keep the `if (expectedText)` guard. That makes the guard meaningful and stops the homepage from short-circuiting on a title match that every page satisfies.

diegohaz (commented):
Initial review against `origin/main` (`4840e12be`) at head `19da29fc`.

**Scope reviewed**

- `.github/workflows/og-images.yml` (branch-tip verification step)
- `app/src/pages/og-image/_generate-images.ts` (blank-image retry + expected-text wait + `OG_IMAGE_BASE_URL`)
- Surrounding context: `.github/workflows/commit-or-pr/action.yml`, `app/src/pages/og-image/api.ts`, `app/src/pages/og-image/[...path].astro`, `app/src/lib/frameworks.ts`, `tsconfig.base.json`

**Validation performed**

- Read each changed file end-to-end and walked the impacted call chain (workflow → generator → og-image page template).
- Confirmed `noUncheckedIndexedAccess: true` in `tsconfig.base.json` — relevant for the `image.data[i]!` finding.
- Cross-checked `getExpectedText` against every `OGImageItem` shape produced in `api.ts` and against the SSR template in `[...path].astro`.
- Did not run `pnpm tsc`, `pnpm lint`, or `pnpm -F app run og-image-generate` locally; the PR description reports these were run.

**Findings**

- Code style: `isBlankImage` uses non-null assertions on indexed `Buffer` access, which our shared style guide forbids. Inline comment posted.
- Dead code / weak wait: `if (expectedText)` is structurally always true, and the title-includes shortcut makes the wait a no-op on the homepage (every page's `<title>` contains `"Ariakit"`). Inline comment posted with two cleanup options.

**Areas considered and left as-is**

- **Branch-tip race.** The new Verify step closes the wide "checkout was stale all along" window. There is still a narrow window between Verify and the `commit-or-pr` rebase-then-push, but `concurrency.cancel-in-progress: true` plus `commit-or-pr`'s own `git fetch + rebase origin/$HEAD_REF` together make this acceptable. The PR description already calls this out.
- **`isBlankImage` corner-sampling heuristic.** Comparing every pixel to the top-left pixel with a ±1 tolerance is brittle in principle, but in practice the OG templates all have `p-11` padding around the content area, so the top-left is the background color. False-positive risk at `MAX_BLANK_PIXEL_RATIO = 0.995` is low for any non-trivial render.
- **`screenshotImage` retry budget.** Only blank images are retried; a `waitForFunction` timeout (e.g., a 404 on the dev server) propagates immediately. This matches the existing per-item error handling in `main()` and seems intentional.
- **`git fetch --no-tags origin "$branch"` vs `git ls-remote`.** `ls-remote` would be slightly cleaner (no FETCH_HEAD side-effect) but functionally equivalent. Not worth a change.

No correctness or security issues found beyond the two style/cleanup items above.

.github/workflows/og-images.yml
```diff
@@ -77,10 +77,30 @@ jobs:
...
+          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
+          git fetch --no-tags origin "$branch"
+          local_sha="$(git rev-parse HEAD)"
+          remote_sha="$(git rev-parse FETCH_HEAD)"
+          if [ "$local_sha" != "$remote_sha" ]; then
+            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
+            echo "is_current=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "is_current=true" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
diegohaz: Fixed in 78d9b10e2 - added an expected-head SHA guard to `commit-or-pr` and passed the verified branch tip from the OG workflow, so the write-back action now refuses to rebase generated output when the PR branch advanced after the outer verification step.

app/src/pages/og-image/_generate-images.ts
diegohaz: Fixed in 71171a73d - removed the indexed non-null assertions in `isBlankImage`, added explicit channel guards, and handled zero-pixel images before calculating the blank-pixel ratio.

app/src/pages/og-image/_generate-images.ts
diegohaz: Fixed in 71171a73d - made `getExpectedText` return `undefined` when an OG item has no title or framework, kept the guard meaningful, and changed the text wait to check body text only instead of short-circuiting through the document title.

diegohaz:
**Fixed**

- 78d9b10e2 ([reply](https://github.com/ariakit/ariakit/pull/5987#discussion_r3255582792)): Closed the remaining write-back race by passing the verified branch tip into `commit-or-pr` and refusing to rebase generated output if the branch advanced.
- 71171a73d ([reply](https://github.com/ariakit/ariakit/pull/5987#discussion_r3255582928)): Removed indexed non-null assertions from `isBlankImage` and added explicit guards for missing channels and zero-pixel images.
- 71171a73d ([reply](https://github.com/ariakit/ariakit/pull/5987#discussion_r3255583108)): Made the expected-text guard meaningful again by returning `undefined` for items without a title/framework and waiting on body text only.

app/src/pages/og-image/_generate-images.ts
```diff
@@ -13,15 +13,62 @@ import pixelmatch from "pixelmatch";
...
+    const currentBlue = pixels[i + 2];
+    const currentAlpha = pixels[i + 3];
+    if (
+      currentRed == null ||
+      currentGreen == null ||
+      currentBlue == null ||
+      currentAlpha == null
+    ) {
+      continue;
+    }
```
diegohaz: **Style nit: split the `||`-chained early exits.**

`ariakit-general-code-style` → Conditionals says: *"For multiple early exits, write one `if` per line — do not combine with `||`"*. Both new channel-null guards in `isBlankImage` combine four conditions with `||` (the outer `return false` guard and the per-pixel `continue` guard). My earlier suggested code used the same `||` pattern — that was my mistake, so the violation got copied in. Sorry for the round trip.

The compliant form:

```ts
if (red == null) return false;
if (green == null) return false;
if (blue == null) return false;
if (alpha == null) return false;
```

and inside the loop:

```ts
if (currentRed == null) continue;
if (currentGreen == null) continue;
if (currentBlue == null) continue;
if (currentAlpha == null) continue;
```

Optional simpler alternative if you'd rather collapse the noise: since `PNG.sync.read` always returns a `width * height * 4` byte buffer, the per-pixel null checks are theoretically unreachable. You could drop the per-iteration `continue` block in favor of a single `if (pixels.length < 4) return false;` guard above the loop, then assert `pixels[i]` etc. as non-null via local non-`!` patterns (e.g., destructured fallback `?? 0`). The split-`if` form above is the most literal style-guide compliance though.

diegohaz (commented):
Re-review against incremental diff `19da29fc..71171a73` (two new commits: `78d9b10e2 Prevent stale generated writebacks`, `71171a73d Clean up OG image readiness checks`). Baseline taken from the prior review marker `pr-5987-19da29fc25c46ceb6ce3d229d906438e083c0b6f-20260517T220848Z`.

**Scope reviewed**

- Full incremental diff across `.github/workflows/og-images.yml`, `.github/workflows/commit-or-pr/action.yml`, and `app/src/pages/og-image/_generate-images.ts`.
- Re-read `commit-or-pr/action.yml` end-to-end to verify the new `expected-head-sha` guard interacts correctly with the existing `git fetch` / `git rebase` / `git push` sequence and the `inputs.event == 'pull_request'` gate.
- Re-read `_generate-images.ts` end-to-end after the cleanup commit.

**Prior findings — verified**

- Copilot's race-condition thread on `og-images.yml`: addressed in `78d9b10e2`. `commit-or-pr` now accepts `expected-head-sha`, fetches the remote tip, and `exit 1`s if the remote moved since the outer verify step recorded `local_sha`. Combined with the existing non-forced `git push origin HEAD:"${HEAD_REF}"` (which fails on non-fast-forward), both the verify→fetch and fetch→push windows are closed for `pull_request` events.
- Non-null assertions in `isBlankImage`: addressed in `71171a73d`. All `!` removed; explicit `== null` guards and a zero-pixel `if (!totalPixels) return false;` guard added.
- Dead `if (expectedText)` and weak title-shortcut: addressed in `71171a73d`. `getExpectedText` now returns `undefined` for the homepage, the guard is meaningful, and the body-only text wait actually waits for body content on items where it makes sense.

**New finding**

- Style: the new channel-null guards in `isBlankImage` combine four conditions with `||`, which violates `ariakit-general-code-style` → Conditionals (*"do not combine with `||`"* for multiple early exits). My prior suggested code had the same mistake; inline comment posted with the split-`if` form. Low priority, code-correctness is fine.

**Areas considered and left as-is**

- **`push` event path through `peter-evans/create-pull-request`.** `expected-head-sha` is passed but ignored in that step — the verify step still narrows the window, and a stale-tip push to `chore/update-og-images` produces a fixable PR rather than a corrupted branch. Acceptable.
- **Theoretically unreachable per-pixel null guard.** `PNG.sync.read` always returns `width * height * 4` bytes, so the per-iteration null check inside the loop is dead at runtime. It satisfies `noUncheckedIndexedAccess` without `!`, which is the trade-off being made. Fine as-is.
- **`git rebase origin/$HEAD_REF` after the `expected-head-sha` check passes.** Effectively a no-op since `local == remote == EXPECTED_HEAD_SHA`, but harmless and keeps the existing code path symmetric.

No correctness or security regressions from the two new commits. The race-closure design (verify step + `expected-head-sha` guard + non-forced push) is sound.

app/src/pages/og-image/_generate-images.ts
diegohaz: Fixed in 133589331 - split both channel guards into one early-exit per line and wrapped the remaining non-primitive early returns in this generator file to match `ariakit-general-code-style`.

diegohaz:
**Fixed**

- 133589331 ([reply](https://github.com/ariakit/ariakit/pull/5987#discussion_r3255598382)): Split the `isBlankImage` channel guards into one early-exit per line and aligned the rest of the generator's touched early returns with `ariakit-general-code-style`.

diegohaz (commented):
Re-review against incremental diff `71171a73..13358933` (one new commit: `133589331 Align OG image generator style`). Baseline taken from the prior review marker `pr-5987-71171a73d7da54017cf62d0e641871639561b594-20260517T222803Z`.

**Scope reviewed**

- Full incremental diff of `app/src/pages/og-image/_generate-images.ts` — the only file changed in this commit.
- Re-read the entire generator file end-to-end to confirm no new style or correctness drift was introduced alongside the targeted fixes.

**Prior finding — verified**

- `||`-chained channel-null guards in `isBlankImage`: addressed in `133589331`. Both the outer return-false guard and the per-pixel continue guard are now split into one early-exit per line. Diego also opportunistically wrapped two adjacent style-violators that were not in my prior comment but were in the same family: `if (item.title) return item.title;` / `if (item.framework) return getFramework(item.framework).label;` (return non-primitives) and `if (!isBlankImage(buffer)) return buffer;` (returns a variable). All three are now compliant with `ariakit-general-code-style` → Conditionals.

**Whole-file sweep**

Walked every `if`, `for`, and `while` in `_generate-images.ts` against the style guide:

- Single-line `if`s remaining are all `continue` / bare `return` / primitive-literal returns (`return false`, `return true`). Compliant.
- Multi-line `if` bodies all use braces. Compliant.
- No `!` non-null assertions remain.
- Loop bodies use braces.

Workflow files (`og-images.yml`, `commit-or-pr/action.yml`) are unchanged in this commit — the race-closure design verified in the previous re-review still stands.

**No findings.** The PR's style and correctness state is clean as of `13358933`.

PR #5984, opened 2026-05-17
Migrate app to Astro 6, Cloudflare 13, and Zod 4

diegohaz:
## Motivation

Migrate the app to Astro 6, `@astrojs/cloudflare` 13, and Zod 4 while keeping Cloudflare prerendering on workerd.

## Solution

Updated the app dependency pins and Cloudflare adapter configuration for Astro 6. The app Wrangler config now lives in `app/wrangler.jsonc`, uses `@astrojs/cloudflare/entrypoints/server` for the Astro build, and the deploy workflows run Wrangler from the app directory.

Replaced `Astro.locals.runtime.env` usage with `cloudflare:workers` env access. Wrangler now generates full worker runtime declarations to `app/worker-configuration.d.ts`, and only the app workerd tsconfig includes those runtime types so the repo-wide DOM projects avoid workerd/DOM global conflicts. The app `check` script generates those types before running `astro check`.

The migration avoids `prerenderEnvironment: "node"`. Clerk server imports are static again: workerd supports Clerk's async storage path through `nodejs_compat`, and the middleware lazily instantiates Clerk only after the Clerk-enabled branch. Markdown and OG image imports were simplified back to static imports where the Astro 6 build proved they are safe. The Next.js preview redirect uses Cloudflare env bindings in worker contexts.

Next.js preview redirects are scoped to the `[framework]/previews/[...nextjs]` SSR route, with `run_worker_first` limited to the React Next.js preview paths so normal static previews continue to be served from assets. A rebuilt Wrangler preview smoke showed the top-level catch-all route served the static 404 instead of the redirect, so the more specific route is intentional. The matcher keeps using stripped preview IDs such as `/react/previews/counter-nextjs/` rather than source paths such as `/react/previews/sandbox/counter-nextjs/`.

Preview app artifacts are built with `CLOUDFLARE_ENV=preview`, so Astro writes preview-specific `dist/server/wrangler.json` and `.wrangler/deploy/config.json`. The `app-build` artifact carries `app/dist` together with that deploy metadata; the preview workflow downloads it to `app` and runs bare `wrangler versions upload`/`wrangler deploy`. The preview Wrangler environment explicitly binds only the `SESSION` KV namespace for Astro sessions, including the nested `previews` binding used by version uploads, while `PLUS`, `EVENTS`, and `ADMIN` stay production-only for now. Production builds again in `deploy.yml` without `CLOUDFLARE_ENV` and also runs bare `wrangler deploy` from the app directory.

[...]

PR #5988, opened 2026-05-18
Retry transient OG image failures

diegohaz:
## Motivation

The OG image generator already retries blank screenshots, but transient failures while navigating, waiting for readiness, or capturing the screenshot can still fail the run immediately before the retry loop has a chance to recover.

## Solution

Wrap each screenshot attempt in the existing retry loop so transient Playwright errors are retried with the same backoff used for blank captures. If every attempt fails with an error, rethrow the last error; if attempts complete but only produce blank images, keep the existing blank-image exhaustion error.

The pre-commit reviewers explicitly checked whether this was an actual issue on `main` rather than a mistaken carry-over from #5984. Both confirmed that current `main` only retries blank screenshots and does not retry thrown navigation/readiness/screenshot failures.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- Pre-commit review gate with native reviewer and Cursor reviewer

### DIFF
--- a/app/src/pages/og-image/_generate-images.ts
+++ b/app/src/pages/og-image/_generate-images.ts
@@ -136,16 +136,25 @@ async function waitForImageReady(page: Page, item: OGImageItem) {
 }
 
 async function screenshotImage(page: Page, item: OGImageItem, url: string) {
+  let lastError: unknown;
   for (let attempt = 1; attempt <= MAX_SCREENSHOT_ATTEMPTS; attempt += 1) {
-    await page.goto(url, { waitUntil: "load" });
-    await waitForImageReady(page, item);
-    const buffer = await page.screenshot({ type: "png", timeout: 60_000 });
-    if (!isBlankImage(buffer)) {
-      return buffer;
+    try {
+      await page.goto(url, { waitUntil: "load" });
+      await waitForImageReady(page, item);
+      const buffer = await page.screenshot({ type: "png", timeout: 60_000 });
+      if (!isBlankImage(buffer)) {
+        return buffer;
+      }
+      lastError = undefined;
+    } catch (error) {
+      lastError = error;
     }
     if (attempt < MAX_SCREENSHOT_ATTEMPTS) {
       await page.waitForTimeout(500 * attempt);
     }
+  }
+  if (lastError) {
+    throw lastError;
   }
   throw new Error(`Blank OG image after ${MAX_SCREENSHOT_ATTEMPTS} attempts`);
 }


diegohaz (commented):
Initial review. Baseline: `99a95bc6d904804f72bd7e110f5dc44fdfb3e4e9..260401e9153b70e993b440d026488bffc0ccac9b`. Scope: `app/src/pages/og-image/_generate-images.ts` plus the surrounding generator file and the writeback workflow context inherited from #5987.

**Verification I ran locally on `260401e9`**
- `pnpm tsc` — passes.
- `pnpm lint` — passes (13 pre-existing warnings in `packages/ariakit-tailwind/src/input.ts` unrelated to this PR).
- Confirmed `app` workspace is `private: true` and listed in `.changeset/config.json` `ignore`, so no changeset is needed.
- CI status on this head: lint, type-check, build (app/Next.js/packages/styles/website), `Generate OG images`, preview deploy, CodeQL, and CodeRabbit all pass; only some Test jobs still running.

**Behavior of the new retry**
Traced all four outcome combinations against the patched `screenshotImage` loop:
- All attempts throw → rethrows the last error.
- All attempts return blank → throws the existing `Blank OG image after N attempts` error.
- Last attempt throws after earlier blanks → throws that last error.
- Last attempt blank after earlier throws → resets `lastError = undefined` and throws the blank-image error. This is intentional ("latest signal wins") and matches the PR description's stated goal of keeping the blank-image exhaustion error when attempts complete-but-blank. The trade-off is that earlier transient throws are not surfaced in the final message; given those errors were retried away and the final reproduction is genuinely a blank capture, this is reasonable for a generator script where the operator wants to know what the *last* attempt actually produced.

Side-effect: `isBlankImage(buffer)` and `PNG.sync.read` were previously uncaught; they now fall into the retry path. That's a strict improvement — a transiently malformed buffer gets retried; a permanently malformed one still surfaces after the budget is exhausted.

**Style / consistency**
- Loop uses braces and `+= 1`, conditional uses braces, `unknown` for the error type, single-object param shape preserved — matches the file's existing conventions and the repo code-style rules.
- `MAX_SCREENSHOT_ATTEMPTS = 3` and `500 * attempt` backoff unchanged; worst-case wall time per failing item stays bounded by `3 × 60s` plus 1500 ms backoff, which the existing `Generate OG images` CI job tolerates.

**Other surfaces checked**
- `screenshotImage` has only one caller (`generateImage`), which already catches and accumulates failures in `failedPaths` — no caller change is needed.
- No other generator code path performs the same `goto → wait → screenshot` triple without retry, so there's no second place to apply the same fix.
- No tests cover this script (it runs as a CI workflow step), consistent with the prior `#5987` patch that introduced the retry loop in the first place.

**Findings**
No blocking or substantive findings. The change does exactly what the description claims and the logic holds up under the cross-product of error/blank outcomes.

**Not validated**
- I did not run `pnpm -F app run og-image-generate` end-to-end locally; the CI `Generate OG images` job already exercises the new path and passes on this head.
- I did not stress-test the failure paths (e.g., simulating `page.goto` rejection) — the trace-through above plus a clean CI run is the evidence I'm relying on.

PR #5989, opened 2026-05-18
Fix plus price key parsing

diegohaz:
## Motivation

Plus price keys are parsed with a suffix-matching regular expression on `main`, so strings such as `invalid-ariakit-plus-usd` can be treated as valid Ariakit Plus prices. The parser also infers team prices with `key.includes("team-")`, which means a malformed prefix can affect the parsed type.

This matters because Stripe price lookup keys flow through the webhook and admin sync before being cached in KV, and `PriceKeySchema` relies on the parser to validate stored price keys.

## Solution

Moved the Plus price-key helpers into a standalone `price-key` module, anchored the parser to the full key, and derives the price type from the structured `team-` segment instead of checking the whole string.

The Stripe module still re-exports the helpers so existing imports keep working, while schemas can import the parser without pulling in the Stripe module. Admin sync and price webhooks now also delete cached entries for invalid lookup keys so previously cached malformed keys do not remain visible after the parser is tightened.

## Verification

- `pnpm test app/src/lib/price-key-test.ts`
- `pnpm lint-fix`
- `pnpm tsc`
- `git diff --check`
- Pre-commit review gate with native and Cursor reviewers, including an explicit check that this fixes an actual issue on `main`

### DIFF
--- a/app/src/lib/price-key-test.ts
+++ b/app/src/lib/price-key-test.ts
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+
+import { expect, test } from "vitest";
+import { parsePlusPriceKey } from "./price-key.ts";
+
+test("parsePlusPriceKey parses personal and team keys", () => {
+  expect(parsePlusPriceKey("ariakit-plus-usd")).toEqual({
+    type: "personal",
+    currency: "usd",
+    countryCode: undefined,
+  });
+  expect(parsePlusPriceKey("ariakit-plus-team-eur-br")).toEqual({
+    type: "team",
+    currency: "eur",
+    countryCode: "br",
+  });
+});
+
+test("parsePlusPriceKey ignores prefixed keys", () => {
+  expect(parsePlusPriceKey("invalid-ariakit-plus-usd")).toEqual({});
+  expect(parsePlusPriceKey("team-invalid-ariakit-plus-usd")).toEqual({});
+});
--- a/app/src/lib/price-key-test.ts
+++ b/app/src/lib/price-key-test.ts
@@ -17,6 +17,16 @@ test("parsePlusPriceKey parses personal and team keys", () => {
     currency: "usd",
     countryCode: undefined,
   });
+  expect(parsePlusPriceKey("ariakit-plus-usd-br")).toEqual({
+    type: "personal",
+    currency: "usd",
+    countryCode: "br",
+  });
+  expect(parsePlusPriceKey("ariakit-plus-team-eur")).toEqual({
+    type: "team",
+    currency: "eur",
+    countryCode: undefined,
+  });
   expect(parsePlusPriceKey("ariakit-plus-team-eur-br")).toEqual({
     type: "team",
     currency: "eur",
--- a/app/src/actions/admin.ts
+++ b/app/src/actions/admin.ts
@@ -202,6 +202,15 @@ async function syncPrices(context: APIContext) {
 
   // Delete prices that are not active
   for (const price of cachedPrices) {
+    const { type } = parsePlusPriceKey(price.key);
+    if (!type) {
+      logger.warn(
+        "Price %s has invalid key. Deleting from cache...",
+        price.key,
+      );
+      await deletePrice(context, price.key);
+      continue;
+    }
     const stripePrice = prices.find((p) => p.id === price.id);
     if (stripePrice?.active) continue;
     logger.warn("Price %s is not active. Deleting from cache...", price.key);
--- a/app/src/lib/price-key.ts
+++ b/app/src/lib/price-key.ts
@@ -1,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+
+export interface GetPlusPriceKeyParams {
+  type?: "personal" | "team";
+  currency?: string;
+  countryCode?: string;
+}
+
+export function getPlusPriceKey({
+  type = "personal",
+  currency = "USD",
+  countryCode,
+}: GetPlusPriceKeyParams) {
+  const isPersonal = type === "personal";
+  const lowercaseCurrency = currency.toLowerCase();
+  const country = countryCode ? `-${countryCode.toLowerCase()}` : "";
+  return isPersonal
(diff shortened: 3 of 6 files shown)


app/src/lib/price-key-test.ts
```diff
@@ -0,0 +1,30 @@
...
+    type: "personal",
+    currency: "usd",
+    countryCode: undefined,
+  });
+  expect(parsePlusPriceKey("ariakit-plus-team-eur-br")).toEqual({
+    type: "team",
+    currency: "eur",
+    countryCode: "br",
+  });
+});
```
diegohaz, line 35: The two added tests cover the regression scope (anchored regex + team-prefix derivation with a country code) but skip two valid shapes that the parser must also keep supporting: a personal key with a country code (`ariakit-plus-usd-br`) and a team key without a country code (`ariakit-plus-team-eur`). Since this file is the only place these helpers are unit-tested, regressions in those branches would slip through.

Consider extending the happy-path test:

```suggestion
test("parsePlusPriceKey parses personal and team keys", () => {
  expect(parsePlusPriceKey("ariakit-plus-usd")).toEqual({
    type: "personal",
    currency: "usd",
    countryCode: undefined,
  });
  expect(parsePlusPriceKey("ariakit-plus-usd-br")).toEqual({
    type: "personal",
    currency: "usd",
    countryCode: "br",
  });
  expect(parsePlusPriceKey("ariakit-plus-team-eur")).toEqual({
    type: "team",
    currency: "eur",
    countryCode: undefined,
  });
  expect(parsePlusPriceKey("ariakit-plus-team-eur-br")).toEqual({
    type: "team",
    currency: "eur",
    countryCode: "br",
  });
});
```

diegohaz (commented):
### Scope
Initial review of `fix/price-key-parser` (head `ebf0d2c`) against base `main` (`99a95bc`). Read all six changed files plus surrounding context: `app/src/lib/price-key.ts`, `app/src/lib/price-key-test.ts`, `app/src/lib/schemas.ts` (`PriceKeySchema`/`PriceDataSchema`), `app/src/lib/stripe.ts` (re-exports and module-load side effects), `app/src/actions/admin.ts` (full `syncPrices` flow), `app/src/pages/api/stripe-webhook.ts`, and downstream callers (`app/src/pages/admin/prices.astro`, `app/src/lib/kv.ts`).

### Validation performed
- Ran `pnpm test app/src/lib/price-key-test.ts` — 2 tests pass.
- Re-derived the regex behavior against a battery of edge inputs (`ariakit-plus-team`, `ariakit-plus-team-team-usd`, `ariakit-plus-usd-br`, `team-invalid-...`, etc.) via a standalone node run to confirm anchored matching, structured team-prefix capture, and rejection of all suffix-prefixed inputs.
- Confirmed `parsePlusPriceKey` is consumed only by `schemas.ts`, `stripe.ts`, `admin.ts`, `stripe-webhook.ts`, and `prices.astro`, and that the re-export keeps the existing import sites working without regressions.
- Verified the cache-cleanup additions are safe: `deletePrice` is a no-op against missing keys, and lookup keys are unique per active Stripe price, so neither the webhook nor `syncPrices` can delete an unrelated cached entry.

### Findings
One inline suggestion on `app/src/lib/price-key-test.ts` to round out the happy-path coverage (personal-with-country and team-without-country shapes), since the new tests are the only place these helpers are unit-tested.

No correctness, security, performance, or compatibility issues found. The anchored regex correctly fixes the suffix-matching bug, deriving `type` from the structured `team-` capture eliminates the `key.includes("team-")` misclassification on malformed keys, and extracting the helpers into `price-key.ts` keeps `schemas.ts` off the `Stripe` module's load-time side effect (`new Stripe(...)`).

### Notes (non-blocking)
- The regex still accepts `ariakit-plus-team` as a personal key with `currency: "team"`. This is unchanged from `main` and not exploitable given the predefined `defaultProducts` lookup keys, but could be tightened later by constraining the currency segment to `[a-z]{3}` (ISO 4217) if you want the parser to reject the literal `team` currency outright.
- After the regex was anchored, `parsePlusPriceKey` returns either `{ type, currency, ... }` or `{}` — never just `currency` — so the `!!type || !!currency` check in `PriceKeySchema` (`app/src/lib/schemas.ts:42`) is effectively `!!type`. Harmless, just worth noting.
- Three call sites (`admin.ts`, `stripe-webhook.ts`, `prices.astro`) still import `parsePlusPriceKey`/`getPlusPriceKey` via the `stripe.ts` re-export. Reasonable for keeping this diff small; `prices.astro` is the only one that doesn't otherwise need the Stripe module and could be switched to import from `./price-key.ts` directly if you want to keep the bundle separation consistent.

app/src/lib/price-key-test.ts
diegohaz: Fixed in ed76bd068 - added the missing happy-path coverage for personal keys with country codes and team keys without country codes.

diegohaz:
**Fixed**

- ed76bd068 ([reply](https://github.com/ariakit/ariakit/pull/5989#discussion_r3255951041)): Added parser happy-path coverage for personal price keys with country codes and team price keys without country codes.

**Verification**

- `pnpm test app/src/lib/price-key-test.ts`
- `pnpm lint-fix`
- `git diff --check`
- Pre-commit review gate with native and Cursor reviewers

diegohaz (commented):
### Scope
Re-review of `fix/price-key-parser`. Incremental baseline is the last marked review head `ebf0d2c`; current head is `ed76bd0`. Only one new commit since the prior review — `ed76bd068 Cover plus price key variants` — which touches `app/src/lib/price-key-test.ts` and nothing else.

### Validation performed
- Diffed `ebf0d2c..ed76bd0` to confirm the only change is the test-coverage addition; no shifts in `price-key.ts`, `schemas.ts`, `stripe.ts`, `admin.ts`, `stripe-webhook.ts`, or any caller.
- Reran `pnpm test app/src/lib/price-key-test.ts` against `ed76bd0` — passes.
- Verified the suggestion from the prior thread is fully implemented: both `ariakit-plus-usd-br` (personal + country code) and `ariakit-plus-team-eur` (team, no country code) are now asserted with the expected output. Combined with the existing assertions, all four valid happy-path shapes are covered.

### Findings
None new. The previously flagged test-coverage gap is fully addressed; no other code changed since the prior review, so the rest of the prior analysis stands unchanged.

### Notes (carried over, still non-blocking)
- `ariakit-plus-team` still parses to `{ type: "personal", currency: "team" }` because the currency segment is `[a-z]+`. Unchanged from `main` and not exploitable given the predefined `defaultProducts` lookup keys; tightening to `[a-z]{3}` (ISO 4217) would close it.
- `!!type || !!currency` in `PriceKeySchema` (`app/src/lib/schemas.ts:42`) is effectively `!!type` after the anchor since `parsePlusPriceKey` now returns either both fields or `{}`. Harmless, just dead-or branch.
- `admin.ts`, `stripe-webhook.ts`, and `prices.astro` still import the helpers via the `stripe.ts` re-export. Only `prices.astro` doesn't otherwise need the Stripe module and could switch to importing from `./price-key.ts` directly if you want the bundle separation consistent everywhere.

PR #5990, opened 2026-05-18
Extract meta descriptions without rendering MDX

diegohaz:
## Motivation

PR #5984 included a meta-description extraction change that is independent from the Astro migration. The current app renders description MDX through a component/container path just to derive text for the `<meta>` description, which adds an unnecessary MDX/React render step to metadata generation.

## Solution

Read description collection bodies directly for metadata, normalize the small MDX subset used by the current descriptions, then run the existing markdown-to-HTML pipeline and extract the first non-empty top-level text block from the rendered fragment. The source descriptions can still contain Markdown/MDX syntax such as links and inline code; this path converts that authored content into the text value needed by the meta tag. The visible page description still renders through the normal MDX component path.

This PR is synced with the current #5984 implementation for the extracted source files (`app/src/lib/content.ts`, `app/src/pages/[...component].astro`, and `app/src/pages/[...example].astro`) except where the follow-up review confirmed PR-local cleanup should stay: the now-unused component-page markdown processor remains removed, and the meta description extraction preserves the original first-block behavior.

Unsupported MDX in the extracted description now fails loudly if raw markup markers leak into the generated text, and the tests cover the real `description.mdx` corpus.

## Verification

- `pnpm lint-fix` (existing underscore warnings, no errors)
- `pnpm test app/src/lib/content-test.ts`
- `pnpm tsc`
- `pnpm build-app-lite`
- Pre-commit review gate with native and Cursor reviewers, including re-review after the parity update and after the follow-up review fixes

### DIFF
--- a/app/src/lib/content-test.ts
+++ b/app/src/lib/content-test.ts
@@ -8,15 +8,32 @@
  * SPDX-License-Identifier: UNLICENSED
  */
 
+import { readdirSync, readFileSync } from "node:fs";
+import { join } from "node:path";
 import { expect, test } from "vitest";
 import { descriptionToText } from "./content.ts";
 
-test("descriptionToText extracts text from description markdown", async () => {
-  const body = `
-import ContentLink from "#app/components/content-link.astro";
+function getDescriptionFiles(
+  path = join(import.meta.dirname, "../examples"),
+): string[] {
+  return readdirSync(path, { withFileTypes: true }).flatMap((entry) => {
+    const entryPath = join(path, entry.name);
+    if (entry.isDirectory()) {
+      return getDescriptionFiles(entryPath);
+    }
+    if (entry.name === "description.mdx") {
+      return [entryPath];
+    }
+    return [];
+  });
+}
 
-Build **accessible** {getFramework(props.framework).label} components with <ContentLink href="/react">Ariakit</ContentLink>.
-`;
+test("descriptionToText extracts text from description markdown", async () => {
+  const body = [
+    `import ContentLink from "#app/components/content-link.astro";`,
+    "",
+    "Build **accessible** {getFramework(props.framework).label} components with <ContentLink href={`/${props.framework}/components/checkbox/`}>Ariakit</ContentLink>.",
+  ].join("\n");
 
   await expect(descriptionToText(body, "react")).resolves.toBe(
     "Build accessible React components with Ariakit.",
@@ -33,4 +50,13 @@ Second paragraph.
   await expect(descriptionToText(body, "react")).resolves.toBe(
     "First paragraph. Second paragraph.",
   );
+});
+
+test("descriptionToText handles real descriptions", async () => {
+  for (const file of getDescriptionFiles()) {
+    const body = readFileSync(file, "utf8");
+    const text = await descriptionToText(body, "react");
+    expect(text, file).toBeTruthy();
+    expect(text, file).not.toMatch(/[<{}]/);
+  }
 });
--- a/app/src/lib/content-test.ts
+++ b/app/src/lib/content-test.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+
+import { expect, test } from "vitest";
+import { descriptionToText } from "./content.ts";
+
+test("descriptionToText extracts text from description markdown", async () => {
+  const body = `
+import ContentLink from "#app/components/content-link.astro";
+
+Build **accessible** {getFramework(props.framework).label} components with <ContentLink href="/react">Ariakit</ContentLink>.
+`;
+
+  await expect(descriptionToText(body, "react")).resolves.toBe(
+    "Build accessible React components with Ariakit.",
+  );
+});
+
+test("descriptionToText returns only the first text block", async () => {
+  const body = `
+First paragraph.
+
+Second paragraph.
+`;
+
+  await expect(descriptionToText(body, "react")).resolves.toBe(
+    "First paragraph.",
+  );
+});
--- a/app/src/lib/content-test.ts
+++ b/app/src/lib/content-test.ts
@@ -23,7 +23,7 @@ Build **accessible** {getFramework(props.framework).label} components with <Cont
   );
 });
 
-test("descriptionToText returns only the first text block", async () => {
+test("descriptionToText extracts text from all blocks", async () => {
   const body = `
 First paragraph.
 
@@ -31,6 +31,6 @@ Second paragraph.
 `;
 
   await expect(descriptionToText(body, "react")).resolves.toBe(
-    "First paragraph.",
+    "First paragraph. Second paragraph.",
   );
 });
--- a/app/src/lib/content-test.ts
+++ b/app/src/lib/content-test.ts
@@ -40,7 +40,7 @@ test("descriptionToText extracts text from description markdown", async () => {
   );
 });
 
-test("descriptionToText extracts text from all blocks", async () => {
+test("descriptionToText extracts text from the first block", async () => {
   const body = `
 First paragraph.
 
@@ -48,7 +48,7 @@ Second paragraph.
 `;
 
   await expect(descriptionToText(body, "react")).resolves.toBe(
-    "First paragraph. Second paragraph.",
+    "First paragraph.",
   );
 });
 
--- a/app/src/lib/content.ts
+++ b/app/src/lib/content.ts
@@ -9,15 +9,8 @@
  */
 import { invariant } from "@ariakit/core/utils/misc";
 import type { MarkdownProcessor, RehypePlugin } from "@astrojs/markdown-remark";
-import { createMarkdownProcessor } from "@astrojs/markdown-remark";
-import type { CollectionEntry, RenderResult } from "astro:content";
-import type { Element } from "hast";
-import { toText } from "hast-util-to-text";
-import rehypeParse from "rehype-parse";
-import { unified } from "unified";
-import { createContainer } from "./astro.ts";
-import { isFramework } from "./frameworks.ts";
-import { rehypeAsTagName } from "./rehype.ts";
+import type { CollectionEntry } from "astro:content";
+import { getFramework, isFramework } from "./frameworks.ts";
 import type { Framework } from "./schemas.ts";
 
 interface ContentGroup {
@@ -135,6 +128,10 @@ async function getMarkdownProcessor() {
   if (markdownProcessor) {
     return markdownProcessor;
   }
+  const [{ createMarkdownProcessor }, { rehypeAsTagName }] = await Promise.all([
+    import("@astrojs/markdown-remark"),
+    import("./rehype.ts"),
+  ]);
   markdownProcessor = await createMarkdownProcessor({
     syntaxHighlight: false,
     rehypePlugins: [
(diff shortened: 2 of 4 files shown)


app/src/lib/content.ts
```diff
@@ -153,92 +150,42 @@ export async function markdownToHtml(markdownString: string) {
...
+  let markdown = "";
+  for (const line of body.split("\n")) {
+    if (!line.trimStart().startsWith("import ")) {
+      markdown += `${line}\n`;
+    }
+  }
+  markdown = markdown
+    .split("{getFramework(props.framework).label}")
+    .join(frameworkLabel);
+  while (true) {
```
diegohaz, line 158: The string-based normalizer below only recognises three exact patterns and silently drops the rest:

1. Lines whose first non-whitespace token is `import ` (single-line only; multi-line `import { ... } from "...";` would leak).
2. The literal `{getFramework(props.framework).label}` substring (no whitespace tolerance, no other JSX expressions).
3. `<ContentLink ...>...</ContentLink>` pairs, assuming no `>` appears inside the opening tag.

All 8 current `description.mdx` files round-trip correctly today, but a future description that introduces, say:

- a new JSX expression like `{props.framework}` or `{getFramework(props.framework).label.toUpperCase()}`
- any other JSX element (e.g. `<Code>`, `<Kbd>`)
- an arrow function in an attribute (`onClick={() => ...}` — would mis-detect the first `>`)

would silently bleed raw `{`, `}` or `<...>` characters into the `<meta description>` tag with no signal at build time. The PR keeps the rendered description path so the visible page stays correct, which makes meta-description regressions especially hard to notice.

A cheap insurance policy without giving up the perf win: assert in the output that no markup leaked through, so the build fails loudly when a new description uses an unsupported pattern. For example, before returning:

```ts
if (/[<{}]/.test(text)) {
  invariant(false, `descriptionToText: unsupported MDX in description (got: ${text})`);
}
```

Alternatively, generalise the JSX stripping to handle any capitalised component instead of only `<ContentLink>`:

```ts
// Drop any opening/closing JSX tag for capitalised components; keep inner text.
markdown = markdown.replace(/<\/?[A-Z][A-Za-z0-9]*(?:\s[^>]*)?>/g, "");
```

The regex still falls short on `=>` inside attributes, but it removes the per-component coupling so adding `<Code>` etc. would Just Work.

Whichever direction you pick, please add at least one regression test that runs `descriptionToText` against every real `app/src/examples/**/description.mdx` and asserts no `{`, `<`, `}` is left in the output — that way the failure mode stops being silent.

app/src/lib/content-test.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import { expect, test } from "vitest";
+import { descriptionToText } from "./content.ts";
+
+test("descriptionToText extracts text from description markdown", async () => {
+  const body = `
+import ContentLink from "#app/components/content-link.astro";
+
+Build **accessible** {getFramework(props.framework).label} components with <ContentLink href="/react">Ariakit</ContentLink>.
```
diegohaz: Every real description that uses `<ContentLink>` writes the href as a JSX expression with a template literal, e.g.:

```mdx
<ContentLink href={`/${props.framework}/components/checkbox/`}>Checkbox</ContentLink>
```

This test uses the simpler `href="/react"` form, which doesn't exercise the trickier case where the parser has to find the first `>` past a `={...}` attribute expression. The current `indexOf(">", start)` happens to work for the template-literal form (no `>` appears inside `${...}`), but the test doesn't lock that in, so if someone refactors the parser the regression won't be caught.

It would also be worth adding a test that drives `descriptionToText` over every real `description.mdx` file in `app/src/examples/**` and asserts the output is non-empty and contains no `{`, `<`, or `}` — that would both pin down the JSX-expression case and act as a regression net for new unsupported MDX patterns (see the related comment on `content.ts`).

diegohaz (commented):
### Summary

Initial review against base `f8f5020` … head `4b0c79b`. Comment-only.

**Scope reviewed**
- The full diff (4 files, +78/-124): `app/src/lib/content.ts`, `app/src/lib/content-test.ts`, `app/src/pages/[...component].astro`, `app/src/pages/[...example].astro`.
- The 8 real description files under `app/src/examples/**/description.mdx`, to confirm every supported pattern round-trips through the new `descriptionToText`.
- `app/src/components/markdown.astro` (other consumer of `markdownToHtml`) to confirm the dynamic-import refactor doesn't break it.
- `app/src/content.config.ts`, `app/src/lib/rehype.ts`, and `app/src/lib/frameworks.ts` for downstream behaviour.
- Existing `parseDescription`/`processor` in `[...component].astro` (now removed) — confirmed unused on `main` before this PR. Good cleanup.

**Validation performed**
- `pnpm test app/src/lib/content-test.ts` — both new tests pass.
- `pnpm tsc` — 0 errors / 0 warnings.
- Ran `descriptionToText` against every real `description.mdx` (script via `node`) — all 8 files extract clean plain text matching what should appear in the `<meta description>` tag (e.g. `"Rendering a visually hidden Checkbox component with a custom checkbox presentation in React."`).

**Findings**

1. **Brittle MDX normalisation** — `descriptionToText` understands only three exact patterns (`import …` lines, the literal `{getFramework(props.framework).label}` expression, and `<ContentLink>…</ContentLink>`). Anything else (other JSX components, other expressions, arrow-function attributes, multi-line imports) silently leaks raw markup into the `<meta description>` while the visible page stays correct, so a regression would not be caught. Suggested either a defensive guard that fails the build if `{`, `<`, or `}` appears in the extracted text, or generalising the JSX-tag stripping to capitalised-component regex. Inline comment on `app/src/lib/content.ts` has details.
2. **Test coverage gap** — the unit test exercises the simple `<ContentLink href="/react">` form, but every real description writes the href as a JSX expression with a template literal. Suggested mirroring the production format and adding a snapshot-style test against all `app/src/examples/**/description.mdx` files so unsupported MDX is caught at test time. Inline comment on `app/src/lib/content-test.ts`.

No correctness blockers. The brittleness is the main concern; even a single-line `if (/[<{}]/.test(text)) invariant(...)` would make the failure mode loud instead of silent.

**Not validated**
- Did not run `pnpm build-app-lite` or `pnpm -F app test` end-to-end (the PR description states these were run by the author). Static analysis and the targeted tests give high confidence.
- Did not exercise non-React frameworks beyond `react` in the local extraction script, but the only framework-dependent step is the `frameworkLabel` substitution, which is mechanical.

app/src/lib/content.ts
diegohaz: Fixed in 9511413f07f23127d964d7b0cf4da780ec421a88. Added an invariant so unsupported MDX markers cannot silently leak into generated meta descriptions, and added corpus coverage for real `description.mdx` files.

app/src/lib/content-test.ts
diegohaz: Fixed in 9511413f07f23127d964d7b0cf4da780ec421a88. Updated the `ContentLink` test to use the real JSX-expression href shape and added coverage over all real `description.mdx` files.

diegohaz:
**Fixed**

- 9511413f07f23127d964d7b0cf4da780ec421a88 ([reply](https://github.com/ariakit/ariakit/pull/5990#discussion_r3256035806)): Added a leaked-MDX invariant and real description corpus coverage for `descriptionToText`.
- 9511413f07f23127d964d7b0cf4da780ec421a88 ([reply](https://github.com/ariakit/ariakit/pull/5990#discussion_r3256036067)): Updated the `ContentLink` test to cover the JSX-expression href shape used by real descriptions.

app/src/pages/[...component].astro
```diff
@@ -180,18 +177,26 @@ function isCurrentHref(href: string, pathname = Astro.url.pathname) {
...
+        { tags: ["h1", "h2", "h3", "h4", "h5", "h6", "ul", "ol"] },
+      ],
     ],
-  ],
-});
+  });
+  return markdownProcessor;
+}
 
 async function parseDescription(description: string) {
```
Comment (diegohaz) [inline app/src/pages/[...component].astro]: Commit `34c4e080f` re-introduced `markdownProcessor`/`getMarkdownProcessor`/`parseDescription` (and the `createMarkdownProcessor`/`rehypeAsTagName` imports they need). None of them are called anywhere in this file or the rest of the workspace — `parseDescription` has zero callers, and removing it removes the only user of `getMarkdownProcessor`/`markdownProcessor`.

This directly contradicts the PR description ("This also removes the now-unused component-page markdown processor") and the original 4b0c79b46 cleanup. The commit message says "Sync meta description extraction with migration PR," but the actual meta-description path now goes through `descriptionToText`/`content.ts`, so syncing back to the old processor doesn't seem load-bearing — please remove or, if you intentionally want to keep it as a placeholder for an upcoming change, drop a comment explaining why and update the PR body.

This was also flagged by [coderabbitai's review on 34c4e080f](https://github.com/ariakit/ariakit/pull/5990#pullrequestreview-4306857678) and not yet addressed.

If you keep the cleanup: also remove the `createMarkdownProcessor` / `MarkdownProcessor` / `RehypePlugin` imports re-added at lines 38-42 and the `rehypeAsTagName` import at line 35.

app/src/lib/content.ts
```diff
@@ -153,92 +154,47 @@ export async function markdownToHtml(markdownString: string) {
...
-  const sections = await contentToText(component, {
-    framework,
-    components,
-  });
-  return sections[0]?.content[0];
+  if (!body) return;
+  const markdown = getDescriptionMarkdown(body, framework);
+  const html = await markdownToHtml(markdown);
+  const tree = unified().use(rehypeParse, { fragment: true }).parse(html);
+  const text = toText(tree).replace(/\s+/g, " ").trim();
```
diegohaz: Worth double-checking the behaviour switch from "first text block" to "all text joined":

- Original commit on this PR (`4b0c79b`): iterated `tree.children` and returned the first node with non-empty text. Test was named "returns only the first text block" and asserted `"First paragraph."`.
- After `9511413f` this is `toText(tree)` over the entire tree, so a description with multiple paragraphs gets all of them joined. The renamed test now asserts `"First paragraph. Second paragraph."`.

For the 8 current `description.mdx` files this makes no difference (all are single-paragraph), but going forward a 3-4 paragraph description would produce a meta description well past the ~160 character SEO sweet spot, with no truncation. The behaviour also no longer matches the PR description ("extract the first non-empty top-level text block") or the pre-PR `parseContentToSections`/`sections[0]?.content[0]` semantics on `main`, both of which were first-paragraph-only.

If the intent is genuinely to include everything, that's fine but worth (a) updating the PR description, and (b) considering a length cap, since meta descriptions over ~300 chars get truncated by Google anyway. If the intent is still first-paragraph, restore the `for (const node of tree.children) { ... if (text) return text; }` loop.

diegohaz (commented):
### Re-review summary

Re-review against the prior reviewed head `4b0c79b46` (initial review) up to current head `9511413f0`. Covered the two new commits:

- `34c4e080f` Sync meta description extraction with migration PR
- `9511413f0` Fail on unsupported description MDX

Plus a fresh scan of any surrounding code newly affected.

**Prior thread verification**

Both threads from the initial review (`content.ts:158` and `content-test.ts`) have been marked resolved by the author with "Fixed in 9511413f0." I re-ran the code against the current head to verify, not just trust the reply:

- The `invariant(!/[<{}]/.test(text), …)` guard is in `content.ts:195-198`. Confirmed it fires for unsupported MDX such as `{props.framework}` and arrow functions inside `<ContentLink>` attributes (manual probe).
- The corpus test (`content-test.ts:60-67`) iterates real `description.mdx` files and asserts no `{`, `<`, `}` leaks.
- The `<ContentLink>` test now uses `href={\`/${props.framework}/components/checkbox/\`}`, matching production shape.
- `pnpm test app/src/lib/content-test.ts` → 3/3 pass.
- `pnpm tsc` → 0 errors / 0 warnings.
- Locally ran `descriptionToText` against all 8 real description files — output identical to the initial review's run (no regression).

Both prior threads are correctly addressed; no replies needed there.

**New findings (from the incremental diff)**

1. **Dead code re-introduced in `[...component].astro`** — `markdownProcessor`/`getMarkdownProcessor`/`parseDescription` came back in `34c4e080f` along with the `createMarkdownProcessor`/`rehypeAsTagName` imports. `parseDescription` has zero callers in the workspace. This contradicts the PR description's "removes the now-unused component-page markdown processor" claim, and coderabbitai's earlier nitpick is still open. Inline comment on `[...component].astro:198`.
2. **Behaviour change from first-paragraph to all-text** — `9511413f` swapped the `for (const node of tree.children) { … if (text) return text; }` loop for `toText(tree)`. The test was renamed and now asserts `"First paragraph. Second paragraph."`. The PR description still says "first non-empty top-level text block," so either the implementation should be reverted to first-paragraph, or the description should be updated (and ideally a length cap added — Google truncates meta descriptions past ~300 chars). Inline comment on `content.ts:194`.

**Not validated**

- Did not run `pnpm build-app-lite` or `pnpm -F app test` end-to-end again; the static-import revert in `content.ts` doesn't change the call graph (just bundle locality), and the PR description claims these were re-run by the author.

app/src/pages/[...component].astro
Comment (diegohaz) [reply app/src/pages/[...component].astro]: Fixed in f8a7711daa852310c767bf459ad0dad2415d2367. Removed the unused component-page markdown processor helpers and their imports, and updated the PR description so it no longer claims they are intentionally kept.

app/src/lib/content.ts
diegohaz: Fixed in f8a7711daa852310c767bf459ad0dad2415d2367. Restored first non-empty top-level block extraction, updated the multi-paragraph regression test to lock that behavior in, and updated the PR description accordingly.

diegohaz:
**Fixed**

- f8a7711daa852310c767bf459ad0dad2415d2367 ([reply](https://github.com/ariakit/ariakit/pull/5990#discussion_r3256244959)): Removed the unused component-page markdown processor helpers and imports, and updated the PR description to match.
- f8a7711daa852310c767bf459ad0dad2415d2367 ([reply](https://github.com/ariakit/ariakit/pull/5990#discussion_r3256245683)): Restored first-block meta-description extraction and updated the multi-paragraph regression test.

diegohaz (commented):
### Re-review summary

Re-review against the prior reviewed head `9511413f` (previous re-review) up to current head `f8a7711d`. Covered the one new commit:

- `f8a7711d` Restore first-block description extraction

**Prior thread verification**

Both re-review threads have been marked resolved by the author. I re-ran the code against the current head to verify, not just trust the replies:

- **Dead code in `[...component].astro`** — confirmed removed at `f8a7711d`. `markdownProcessor`, `getMarkdownProcessor`, `parseDescription`, and their `createMarkdownProcessor` / `rehypeAsTagName` / `MarkdownProcessor` / `RehypePlugin` imports are all gone. `grep` for any of those identifiers in the file returns no results.
- **First-block behaviour** — confirmed restored. `descriptionToText` now iterates `tree.children` and returns the first non-empty text node, with the `invariant(!/[<{}]/.test(text), …)` guard moved inside the loop so it only inspects the block actually returned. The test was renamed back to `"descriptionToText extracts text from the first block"` and asserts `"First paragraph."` — manual probe with `"First paragraph.\n\nSecond paragraph."` confirms it returns just the first.
- The invariant still fires for unsupported MDX (`{props.framework}` and arrow-function attributes both throw as expected in a manual probe).
- PR description updated to keep the "first non-empty top-level text block" wording and explicitly note the removed processor — both now match the code.

**Validation performed**

- `pnpm test app/src/lib/content-test.ts` → 3/3 pass.
- `pnpm tsc` → 0 errors / 0 warnings.
- Ran `descriptionToText` against every real `app/src/examples/**/description.mdx` — all 8 files produce the same clean plain text as in earlier re-reviews. No regression from the loop reintroduction.

**Findings**

No new findings. All previously raised threads are addressed in the code, the tests, and the PR description. Ready from a review-quality standpoint.

app/src/lib/content-test.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import { readdirSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { expect, test } from "vitest";
+import { descriptionToText } from "./content.ts";
+
+function getDescriptionFiles(
+  path = join(import.meta.dirname, "../examples"),
+): string[] {
```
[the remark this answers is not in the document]
diegohaz: We are using Node 24

PR #5997, opened 2026-05-19
Update app Vite to 7.3.3

diegohaz:
## Motivation

The app workspace should stay on Vite 7 while Astro depends on Vite 7. A plain direct dependency update can leave Astro and Astro integrations on the older locked Vite patch, which creates duplicate Vite type instances in the app config.

## Solution

Update the app workspace direct `vite` dependency to `7.3.3`, run `pnpm dedupe` so Astro and related app dependencies also resolve `vite@7.3.3`, and add an app-scoped Renovate rule that allows Vite 7 updates while blocking Vite 8 for `app/package.json`.

## Testing

- `pnpm dedupe --check`
- `pnpm lint-fix`
- `pnpm tsc -v`
- `pnpm build-app-lite`


diegohaz:
Updated direction again after investigating pnpm resolution. Astro and its integrations declare Vite ranges (`^7.3.x`), so this PR now keeps the app on Vite v7 generally rather than pinning to `7.3.2`.

This branch updates the app direct dependency to `vite@7.3.3`, runs `pnpm dedupe`, and verifies the app workspace graph resolves Astro, Astro integrations, Tailwind Vite, Vitest, and the direct app dependency to `vite@7.3.3`. The remaining `vite@8.0.13` is from the root workspace and is expected.

Local checks pass:

- `pnpm dedupe --check`
- `pnpm lint-fix`
- `pnpm tsc -v`
- `pnpm build-app-lite`

PR #6000, opened 2026-05-20
Simplify test spy cleanup

diegohaz:
## Motivation

Vitest supports explicit resource management for spies, which lets tests restore local mocks automatically when the test scope exits. Several example tests still used manual `mockRestore()` calls or file-level `beforeEach` hooks just to keep `window.alert` mocked.

## Solution

Use `using` for local `vi.spyOn` calls so Vitest restores spies at the end of each test. The alert-based examples now mock `window.alert` only in the tests that exercise submit/menu actions, removing shared mutable mock state while keeping the submit assertions explicit.

## Verification

- `pnpm vitest run examples/command/test.ts examples/command-enter-disabled/test.ts examples/command-space-disabled/test.ts examples/composite-virtual-focus-events/test.ts examples/composite-hover-virtual-focus-events/test.ts examples/form/test.ts examples/select-form-disabled/test.ts examples/select-form/test.ts examples/form-select/test.ts examples/form-radio/test.ts examples/menu/test.ts`
- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/examples/command-enter-disabled/test.ts
+++ b/examples/command-enter-disabled/test.ts
@@ -2,23 +2,19 @@ import { press, q } from "@ariakit/test";
 import { expect, test, vi } from "vitest";
 
 test("enter doesn't trigger the alert", async () => {
-  const alertMock = vi.spyOn(window, "alert").mockImplementation(() => {});
+  using alertMock = vi.spyOn(window, "alert").mockImplementation(() => {});
 
   await press.Tab();
   expect(q.text("Accessible button")).toHaveFocus();
   await press.Enter();
   expect(alertMock).toHaveBeenCalledTimes(0);
-
-  alertMock.mockRestore();
 });
 
 test("space does trigger the alert", async () => {
-  const alertMock = vi.spyOn(window, "alert").mockImplementation(() => {});
+  using alertMock = vi.spyOn(window, "alert").mockImplementation(() => {});
 
   await press.Tab();
   expect(q.text("Accessible button")).toHaveFocus();
   await press.Space();
   expect(alertMock).toHaveBeenCalledTimes(1);
-
-  alertMock.mockRestore();
 });
--- a/examples/command-space-disabled/test.ts
+++ b/examples/command-space-disabled/test.ts
@@ -2,23 +2,19 @@ import { press, q } from "@ariakit/test";
 import { expect, test, vi } from "vitest";
 
 test("enter does trigger the alert", async () => {
-  const alertMock = vi.spyOn(window, "alert").mockImplementation(() => {});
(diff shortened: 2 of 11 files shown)


diegohaz (commented):
Initial review at `1bdc614d` against base `b9a47f99`. Comment-only — no actionable findings.

## Scope reviewed
All 11 changed test files plus the touched example source files (`examples/{command,command-enter-disabled,command-space-disabled,composite-virtual-focus-events,composite-hover-virtual-focus-events,form,form-radio,form-select,menu,select-form,select-form-disabled}/`) to confirm which interactions actually trigger `window.alert` and that the new per-test spy scoping is correct.

## Validation performed
- Locally applied the diff and ran `pnpm vitest run` against all 11 affected files — **56/56 passing**.
- Locally ran `pnpm tsc` — passes, confirming TS 6.0.3 with the configured `target: es2018` / `lib: ["DOM", "ESNext"]` plus Node 24 in `.nvmrc` correctly supports the `using` declarations and `MockInstance[Symbol.dispose]`.
- Verified `vitest-fail-on-console` does **not** intercept `console.log` by default (`shouldFailOnLog: false`), so removing the file-level `console.log` spy in the composite tests after `using log` is disposed cannot cause unrelated `console.log` writes to fail the suite.
- Cross-checked each test in `form*/test.ts`, `menu/test.ts`, `select-form*/test.ts`, and `command*/test.ts` against its example source to confirm that only the tests now declaring `using alert = …` are the ones whose user paths actually reach the `alert(...)` callsite (e.g., `menu/index.react.tsx:12` `onClick={() => alert("Edit")}` is only reached by the converted menu-item tests).

## Findings
None. The refactor is a strict improvement:

- `using` calls `mockRestore()` on scope exit, replacing the previous file-level `let alert = spyOnAlert()` + `beforeEach`/`mockClear()` pattern that only cleared call history and left `window.alert` spied across the whole file. Spies are now properly restored, scope is per-test, and shared mutable state is eliminated.
- Cleanup now runs even if a test body throws, which the old manual `mockRestore()` / `mockReset()` calls at the end of the test body did not guarantee.
- The added `expect(alert).toHaveBeenCalledTimes(1)` assertions in `form/test.ts` and `form-radio/test.ts` "reset form on submit" tests usefully verify that submission actually reached the alert handler before the reset assertion runs, matching the PR description's intent of keeping submit assertions explicit.
- Test-only refactor with no shipped-code impact, so no changeset is required.

## Areas not fully validated
- I did not run `pnpm test` (the root suite) or `pnpm -F app test`. The diff is confined to per-example Vitest tests that all run under `pnpm vitest run`, which I exercised directly.

PR #6013, opened 2026-05-22
Stabilize OG image generation

diegohaz:
## Motivation

The OG image workflow accepted an Astro dev error page as a valid screenshot after the Astro monorepo update. The bot PR at #6012 updated `react_examples.png` to an error overlay because standalone pages still linked to `/favicon.svg`, which is not present in `app/public` and now returns a dev-server 500.

## Solution

Point the remaining standalone favicon links at the existing `/icon.svg` asset so the OG page and related layouts no longer trigger the missing-asset 500.

Harden the OG image generator so it only accepts the intended OG template, fails on Vite/Astro error overlays, and reports 5xx responses immediately instead of saving their rendered error pages as PNG output.

## Verification

- `OG_IMAGE_BASE_URL=http://localhost:4322 pnpm -F app run og-image-generate`
- `pnpm lint-fix`
- `pnpm tsc`
- Pre-commit review pass with native Codex and Cursor reviewers

### DIFF
--- a/app/src/pages/og-image/_generate-images.ts
+++ b/app/src/pages/og-image/_generate-images.ts
@@ -10,7 +10,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import pixelmatch from "pixelmatch";
-import type { Page } from "playwright";
+import type { Page, Response } from "playwright";
 import { chromium } from "playwright";
 import { PNG } from "pngjs";
 import { getFramework } from "#app/lib/frameworks.ts";
@@ -25,6 +25,21 @@ const PUBLIC_DIR = path.resolve(process.cwd(), "public");
 const MAX_DIFF_PIXEL_RATIO = 0.001;
 const MAX_BLANK_PIXEL_RATIO = 0.995;
 const MAX_SCREENSHOT_ATTEMPTS = 3;
+
+function getServerError(response: Response) {
+  const status = response.status();
+  if (status < 500) return;
+  return `${status} ${response.request().method()} ${response.url()}`;
+}
+
+function assertNoServerErrors(item: OGImageItem, serverErrors: string[]) {
+  if (!serverErrors.length) return;
+  throw new Error(
+    `Server errors while rendering ${item.path}:\n${[
+      ...new Set(serverErrors),
+    ].join("\n")}`,
+  );
+}
 
 function isBlankImage(buffer: Buffer) {
   const image = PNG.sync.read(buffer);
@@ -118,6 +133,7 @@ async function getItemsToGenerate() {
 
 async function waitForImageReady(page: Page, item: OGImageItem) {
   await page.waitForLoadState("load");
+  await page.waitForSelector("[data-og-image]");
   const expectedText = getExpectedText(item);
   if (expectedText) {
     await page.waitForFunction(
@@ -133,21 +149,56 @@ async function waitForImageReady(page: Page, item: OGImageItem) {
       });
     });
   });
+  const errorText = await page.evaluate(() => {
+    const overlay = document.querySelector("vite-error-overlay");
+    if (overlay) {
+      return (
+        overlay.shadowRoot?.textContent?.trim() ||
+        overlay.textContent?.trim() ||
+        "Vite error overlay detected"
+      );
+    }
+    const text = document.body.textContent?.trim();
+    if (!text) return;
+    if (!text.includes("An error occurred.")) return;
+    if (!text.includes("Stack Trace")) return;
+    return text;
+  });
+  if (errorText) {
+    throw new Error(
+      `Error overlay while rendering ${item.path}:\n${errorText.slice(0, 500)}`,
+    );
+  }
 }
 
 async function screenshotImage(page: Page, item: OGImageItem, url: string) {
   let lastError: unknown;
   for (let attempt = 1; attempt <= MAX_SCREENSHOT_ATTEMPTS; attempt += 1) {
+    const serverErrors: string[] = [];
+    const handleResponse = (response: Response) => {
+      const error = getServerError(response);
+      if (error) {
+        serverErrors.push(error);
+      }
+    };
     try {
-      await page.goto(url, { waitUntil: "load" });
+      page.on("response", handleResponse);
+      const response = await page.goto(url, { waitUntil: "load" });
+      if (response) {
+        handleResponse(response);
+      }
+      assertNoServerErrors(item, serverErrors);
       await waitForImageReady(page, item);
       const buffer = await page.screenshot({ type: "png", timeout: 60_000 });
+      assertNoServerErrors(item, serverErrors);
       if (!isBlankImage(buffer)) {
         return buffer;
       }
       lastError = undefined;
     } catch (error) {
       lastError = error;
+    } finally {
+      page.off("response", handleResponse);
     }
     if (attempt < MAX_SCREENSHOT_ATTEMPTS) {
       await page.waitForTimeout(500 * attempt);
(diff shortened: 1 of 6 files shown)


app/src/pages/og-image/_generate-images.ts
```diff
@@ -133,21 +149,56 @@ async function waitForImageReady(page: Page, item: OGImageItem) {
...
+    if (!text.includes("An error occurred.")) return;
+    if (!text.includes("Stack Trace")) return;
+    return text;
+  });
+  if (errorText) {
+    throw new Error(
+      `Error overlay while rendering ${item.path}:\n${errorText.slice(0, 500)}`,
+    );
+  }
 }
```
diegohaz: The new error-overlay check (lines 152–171) runs *after* `await page.waitForSelector("[data-og-image]")` on line 136, but the selector is the slow-fail path for the exact scenario the check is meant to catch. When Astro's dev server renders an error page (rather than the OG template) into the `/og-image/...` route — the failure mode that produced the bad `react_examples.png` referenced in the PR motivation — `[data-og-image]` only exists in the OG template (`app/src/pages/og-image/[...path].astro:54`) and is absent from the error UI. `waitForSelector` then times out at its default 30s and throws a generic `TimeoutError` before this descriptive check ever runs, and the retry loop in `screenshotImage` repeats the cycle twice more, so a single bad render burns ~90s with no useful diagnostic and no overlay text in the failure message.

The 5xx response listener already covers the favicon-as-subresource case quickly (since `/favicon.svg` 500s during `waitForLoadState("load")`), so the in-body overlay path here is the one this check is uniquely positioned to catch — and it should happen *before* the selector wait to actually do so. The cleanest shape is to extract the `page.evaluate` block into a small helper (e.g. `assertNoErrorOverlay(page, item)`) and call it right after `waitForLoadState("load")`; keeping a second call at the current location is fine if you also want to catch errors that surface during the fonts/rAF wait, but the early call is what salvages the diagnostic when the dev server never returned the OG template at all.

diegohaz (commented):
Initial review of `fix-og-image-stability` (#6013) at `9e401b08`, diffed against `main` (`3caa8256`).

## Scope

Reviewed all 6 changed files: the four standalone Astro layouts and `og-image/[...path].astro` that switch `/favicon.svg` → `/icon.svg`, the new `data-og-image` template marker, and the hardening additions in `og-image/_generate-images.ts` (5xx response listener, Vite/Astro error-overlay detection, and the selector gate for the OG template). Also walked through the surrounding context: the `og-images.yml` workflow that drives this generator in CI, the dev-server `pnpm -F app run dev` startup it depends on, `meta.astro` to confirm `/icon.svg` is the canonical favicon asset, `app/public/` to confirm `/favicon.svg` is not present and `/icon.svg` is, and a repo-wide search to confirm no remaining `/favicon.svg` link references.

## What I checked

- Favicon swap is consistent and complete: no `/favicon.svg` link references remain anywhere in the repo, and `/icon.svg` exists in `app/public/`. The matching of the standalone layouts to the canonical `meta.astro` pattern looks right.
- `data-og-image` marker is unique to the OG template, so the selector wait genuinely distinguishes the intended page from any other.
- Response listener registration order: `page.on("response", handleResponse)` is installed synchronously before `await page.goto(...)`, so the main document response is captured by the listener; the manual `handleResponse(response)` call after `goto` is therefore redundant in the success path, but harmless because `assertNoServerErrors` dedupes via `new Set(...)`. Not worth changing on its own.
- Per-attempt listener lifecycle in `screenshotImage` is correct: each attempt scopes its own `serverErrors` and `handleResponse`, and the `finally` block removes the listener on every path, so retries do not accumulate listeners or stale errors.
- `assertNoServerErrors` is called both before `waitForImageReady` (catches subresource 5xx like the missing `/favicon.svg` during the load event) and after `page.screenshot` (catches anything that fired in between). Good coverage for the motivating scenario.
- The `"An error occurred." + "Stack Trace"` body-text heuristic is tied to Astro's dev error page wording, but the OG template's user-controlled content (framework label, item title) is constrained enough that a false positive is unlikely.

## Findings

1. **Error-overlay check is gated behind a 30s-per-attempt selector timeout.** Inline comment on `_generate-images.ts:134-172`. When Astro's dev server renders an error page (rather than the OG template) into the `/og-image/...` route — the exact failure mode the PR is hardening against — `[data-og-image]` is absent, so `page.waitForSelector("[data-og-image]")` times out at 30s and throws a generic `TimeoutError` before the new descriptive overlay check ever runs. With three attempts that is ~90s per failed item with no useful diagnostic surfaced to the log, which undercuts the hardening goal. Moving the overlay check ahead of the selector wait (e.g. via an `assertNoErrorOverlay` helper) gives the fast, descriptive failure the check was added to provide.

## What I did not validate locally

I did not start the Astro dev server or execute `pnpm -F app run og-image-generate` in this worktree (the PR branch is already checked out in another worktree, so I reviewed against a detached checkout of `9e401b08`). The reasoning behind the finding is grounded in the new code itself plus the PR description's account of the captured error overlay; the slow-fail path is structural and does not depend on a specific Astro version.

app/src/pages/og-image/_generate-images.ts
diegohaz: Fixed in 2f762a694 - moved the overlay assertion into `assertNoErrorOverlay` and run it immediately after load, before waiting for `[data-og-image]`, while keeping the later check for errors that appear after fonts/RAF.

diegohaz:
**Fixed**

- 2f762a694 ([reply](https://github.com/ariakit/ariakit/pull/6013#discussion_r3285387845)): Moved the OG error-overlay assertion before the `[data-og-image]` wait and kept the later post-fonts/RAF check, so document-level dev error pages fail with the useful overlay diagnostic instead of timing out on the marker selector.

diegohaz (commented):
Re-review of `fix-og-image-stability` (#6013) at `2f762a69`, incremental baseline `9e401b08` (prior marked review).

## Scope

One new commit since the last review (`2f762a69 — Report OG image errors before marker waits`), scoped to `app/src/pages/og-image/_generate-images.ts` (+23 / −18, no other files touched). The favicon swap and `data-og-image` marker on the OG template are unchanged from the prior head and remain in the intended state.

## Prior finding

- **Error-overlay check gated behind the 30s `waitForSelector` timeout** (resolved). `assertNoErrorOverlay` is now extracted as a top-level helper and invoked twice in `waitForImageReady`: once immediately after `waitForLoadState("load")` (the fast-fail path on rendered error pages where `[data-og-image]` is absent) and once after the fonts/rAF wait (catches errors that surface later, exactly the trade-off called out in the prior thread). The check body is preserved verbatim, so neither the Vite shadow-root path nor the Astro `"An error occurred." + "Stack Trace"` body-text heuristic regressed. Verified against the new file directly, not just the resolved-thread state.

## What I checked in the incremental commit

- `assertNoErrorOverlay` is invoked from both call sites in `waitForImageReady` and is the only consumer; no other places in the file need the same gating.
- Both invocations are inside the existing `screenshotImage` try/catch, so a thrown overlay error is caught, attributed to `lastError`, and subjected to the existing retry-with-backoff. The retry loop already removes the response listener in `finally` on every path, so the new early throw does not leak listeners.
- The extra `page.evaluate` round-trip per attempt is small and only doubles in the success path (cheap DOM read), which is an acceptable cost for the fast-fail diagnostic.
- No new types, imports, or public surface changes; the refactor is purely a reordering plus extraction.

## Result

No new findings. The prior finding is genuinely resolved in the current code, and the incremental commit did not expose adjacent issues.

PR #6014, opened 2026-05-22
Split utility and store packages

diegohaz:
## Motivation

Ariakit's core, React, and Solid packages currently share utility and store code through broader package boundaries. Splitting those internals into smaller packages makes the dependency graph clearer and lets framework-specific packages depend only on the utility and store layers they need.

## Solution

Add six pure ESM workspace packages at version `0.0.0`: `@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`, and `@ariakit/solid-store`.

The new packages expose a single root entrypoint and are built by a new `ariakit build` command in `@ariakit/scripts` using Rolldown. Their `exports` include a package-specific `ariakit-source` condition for monorepo source resolution, while keeping the public `source` condition and the Solid-specific condition where needed. TypeScript, Vitest, and the legacy website use that package-specific condition; the Astro app keeps scoped aliases because content sync must work before package `dist` output exists.

Existing core packages now depend on these smaller packages. The old core utility subpath re-exports are removed as a breaking change, while `@ariakit/react-core/utils/store` and `@ariakit/react` continue exposing the React store utilities already public through `@ariakit/react`.

The new packages start from `0.0.0`, and the changeset marks them as minor so their first release is `0.1.0`.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test`
- `pnpm test-solid`
- `pnpm build`
- `pnpm build-app-lite`
- `pnpm -F website build` with the six new package `dist` folders hidden
- `pnpm exec changeset status --verbose`
- `npm pack --dry-run --json` leak check for the six new packages


scripts/build/context.js
```diff
@@ -1,12 +1,21 @@
...
+  "ariakit-react",
+  "ariakit-react-core",
+  "ariakit-react-store",
+  "ariakit-react-utils",
+].includes(dirName);
+export const isSolid = [
+  "ariakit-solid",
+  "ariakit-solid-core",
+  "ariakit-solid-utils",
+].includes(dirName);
```
diegohaz: `@ariakit/solid-store` is missing from the `isSolid` array, breaking the symmetry with `isReact` (which includes both `ariakit-react-store` and `ariakit-react-utils`). Concrete consequences I observed by running `pnpm build`:

1. `ariakit-solid-store` falls into `buildStandard()` instead of `buildSolid()`/`buildSolidSource()`, so its ESM output picks up a React-only `"use client";` banner. After build, `packages/ariakit-solid-store/esm/index.js` literally starts with:
   ```js
   "use client";
   // src/index.ts
   export * from "@ariakit/store";
   ```
2. `getPackageJson()` in `scripts/build/utils.js` only emits the `solid:` conditional export when `isSolid` is true, so the published `@ariakit/solid-store` will have no Solid-source entry alongside its esm/cjs entries — unlike every other Solid-named package.
3. `getBuildFolders()` skips the `solid/` source dir for this package, so the gitignore/cleanup logic doesn't treat it as a Solid build either.

Functionally this works today because the package only re-exports `@ariakit/store` (no JSX/Solid code). But the PR description says this is the Solid-named entry that re-exports `@ariakit/store` *currently* — i.e., the intent is to grow Solid-specific wrappers here, mirroring `@ariakit/react-store`. As soon as that happens (or if a downstream Solid bundler keys on the `solid:` condition), this asymmetry becomes a real bug. Cheap to fix now while the file is being touched:

```suggestion
export const isSolid = [
  "ariakit-solid",
  "ariakit-solid-core",
  "ariakit-solid-store",
  "ariakit-solid-utils",
].includes(dirName);
```

diegohaz (commented):
Initial review at head `1b83f4c0e`, baseline `main` (`ecbcbc328`). Reviewed the full diff (330 files) plus the surrounding repo context — build/test config, all new package manifests and tsconfigs, the compat re-export shims under `@ariakit/{core,react-core,solid-core}/src/utils/`, and a sampling of updated consumer files (`button.tsx`, `dialog.tsx`, `react/src/store.ts`, `ariakit-test/src/*`).

## Scope checked

- New package definitions: `package.json`, `tsconfig{,.build}.json`, `readme.md`, `.gitignore`, `.npmignore`, `license` for all six new packages.
- Dependency graph: confirmed no circular deps and that framework-agnostic packages (`@ariakit/utils`, `@ariakit/store`) import nothing from `@ariakit/core`/`@ariakit/react-*`/`@ariakit/solid-*`, and that Solid-side packages don't pull in React.
- Compat shims: every old `@ariakit/core/utils/*`, `@ariakit/react-core/utils/*`, and `@ariakit/solid-core/utils/*` subpath now re-exports its new home; no consumer code still imports the old paths.
- Changeset: bumps the six new packages from `0.0.0` → `0.1.0` (minor) and lets Changesets cascade patch bumps to `@ariakit/core`, `@ariakit/react-core`, `@ariakit/react`, `@ariakit/solid-core`, `@ariakit/solid`, `@ariakit/tailwind`, `@ariakit/test` via `updateInternalDependencies: "patch"`.
- `@ariakit/test` now depends on `@ariakit/utils` instead of `@ariakit/core` — verified all imports in `packages/ariakit-test/src/**` are `@ariakit/utils/*`.
- `@ariakit/react/src/store.ts` correctly re-routes `useStoreState` to `@ariakit/react-store`.
- Verified `use-sync-external-store` moved cleanly: only `@ariakit/react-store` uses it, and it's removed from `@ariakit/react-core`'s direct deps.

## Verification commands run locally

- `pnpm install` ✅
- `pnpm tsc` ✅ (exit 0)
- `pnpm build` ✅ (exit 0) — inspected emitted artifacts for `@ariakit/{utils,store,react-utils,react-store,solid-utils,solid-store}`.
- `pnpm exec changeset status --verbose` ✅ — matches the version intent in the PR description (six new packages → 0.1.0; existing affected packages bumped patch).
- `pnpm test` ✅ — 805/805 passing after `pnpm -r run clean` (running tests directly after `pnpm build` blew up because the build step rewrites every package's `main`/`module`/`types` to point at built artifacts and leaves them that way; that's a pre-existing tooling quirk, not a PR regression).

## Findings

- One concrete inline comment on `scripts/build/context.js`: `ariakit-solid-store` is missing from the `isSolid` array, breaking the symmetry with the React side. Functionally fine today because the package is a thin re-export of `@ariakit/store`, but it produces a Solid-named publish with a React `"use client";` banner and no `solid:` conditional export. Worth fixing now while the file is being touched, since the PR description signals the intent for this package to gain Solid-specific wrappers later.

No other actionable findings — the refactor is mechanically clean, the dependency layering is sensible, and the compat shims preserve every previously published subpath.

## Not validated

- `pnpm test-solid` (deferred — the React-side suite already exercised the moved store/utils end-to-end through `@ariakit/test` and `@ariakit/react-core` consumers).
- Actual published-package contents (`npm pack`) and whether downstream Solid bundlers key on the `solid:` export condition — only inferred from `scripts/build/utils.js`.

scripts/build/context.js
diegohaz: Fixed in 8387a40e2 - added `ariakit-solid-store` to the Solid build bucket so `@ariakit/solid-store` gets the Solid export/build behavior, including the generated `/solid` output path.

diegohaz:
**Fixed**
- 8387a40e2 (https://github.com/ariakit/ariakit/pull/6014#discussion_r3285707913): Added `ariakit-solid-store` to the Solid build bucket so `@ariakit/solid-store` gets the Solid export/build behavior and generated `/solid` output path.

scripts/build/context.js
diegohaz: Confirmed. Built `@ariakit/solid-store` at `8387a40e2` and verified the three previous symptoms are gone:

1. `packages/ariakit-solid-store/esm/index.js` no longer has the `"use client";` banner — output starts with `// src/index.ts` directly.
2. `packages/ariakit-solid-store/package.json` `exports."."` now includes the `solid` condition (`"solid": "./solid/index.jsx"` ordered before `import`/`require`, matching `@ariakit/solid-utils`).
3. `solid/index.jsx` is emitted with `export * from "@ariakit/store";`, and `.gitignore` now lists `/solid` so the source build isn't accidentally committed.

Fix matches the suggested diff exactly. Resolving as fixed.

diegohaz (commented):
Re-review at head `8387a40e2`, incremental baseline `1b83f4c0e` (the head reviewed in [the initial review](https://github.com/ariakit/ariakit/pull/6014#pullrequestreview-4342287795)).

## Scope of this re-review

Only one commit added since the last review:

- `8387a40e2 Classify solid store as Solid package` — touches `scripts/build/context.js` (+1) and `packages/ariakit-solid-store/.gitignore` (+1).

This is exactly the fix proposed in the prior inline finding. No other files changed.

## Verification

Built `@ariakit/solid-store` locally at the new head and confirmed all three previously observed symptoms are gone — replied on the existing thread with the per-symptom verification.

## Findings

None. The prior finding is resolved and the new commit doesn't expose anything else worth flagging in the surrounding code.

PR #6018, opened 2026-05-22
Split Ariakit internal packages

diegohaz:
## Motivation

The shared utility and store code has outgrown the old core-package deep re-export paths. We need dedicated published utility/store packages with a single public entrypoint, pure ESM output, and a build path that keeps export maps in sync without legacy proxy folders.

The existing `core`, `react-core`, and `solid-core` package names also no longer describe their role after the split. Treating their replacements as new internal component packages lets us avoid releasing breaking updates to the old package names.

## Solution

Added the new `@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`, and `@ariakit/solid-store` packages at `0.0.0`, with minor changesets for the first `0.1.0` release. The new packages are exports-only pure ESM packages with a single public entrypoint.

Renamed the internal component packages to `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/solid-components` as new `0.0.0` packages with cleaned changelogs. The old `@ariakit/core`, `@ariakit/react-core`, and `@ariakit/solid-core` packages are removed from git, and current workspace imports now use the new component package names.

Added a Rolldown-powered `ariakit build` command in `@ariakit/scripts` that updates production exports from public source files by default, generates package ignore files, builds declarations, and preserves Solid `solid` output. Added a dedicated `ariakit clean` command that restores source-path exports and removes build output. `ariakit dev` now cleans by default, watches package changes through the same clean logic, and starts the app dev servers without the old `dev-package-json` process. Passing `--no-clean` preserves the current package exports so built-output testing remains possible. Existing non-React packages now use the ESM-only build, while `@ariakit/react` keeps the legacy CJS/ESM build and bundles the new pure ESM internal packages into CJS output.

Moved utility/store implementations into the new packages, removed the old core utility deep re-export shims, moved `@ariakit/react-components/utils/store` to `@ariakit/react-components/store`, and kept `@ariakit/react`'s `useStoreState` public behavior intact. Internal package READMEs include the internal-dependency semver warning, while experimental packages use experimental wording instead.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test`
- `pnpm build`
- `pnpm clean`
- `pnpm build-app-lite`
- `pnpm build-website-lite`
- `pnpm vitest .changeset/get-changelog-entry-test.ts`
- `pnpm -F @ariakit/utils run build && pnpm -F @ariakit/utils run clean`
- `pnpm -F @ariakit/utils run clean package.json && pnpm -F template-react build`
- `node --input-type=module -e 'await import("./packages/ariakit-test/dist/react.js")'`
- `node -e "require('./packages/ariakit-react/cjs/index.cjs'); require('./packages/ariakit-react/cjs/store.cjs')"`
- `npm pack --dry-run --json` spot checks for the new packages and `@ariakit/test`
- `node packages/ariakit-scripts/src/index.ts build --help`
- `node packages/ariakit-scripts/src/index.ts clean --help`
- `node packages/ariakit-scripts/src/index.ts dev --help`
- Verified all package scripts invoking `ariakit` declare `@ariakit/scripts`
- Verified `website/build-pages/get-example-deps.js` matches `origin/main`
- Pre-commit review gate with native Codex subagent and Cursor CLI reviewer


packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,299 @@
...
+    transform: shouldTransformReactJsx
+      ? {
+          jsx: "react-jsx",
+        }
+      : undefined,
+    output: {
+      dir: distDir,
+      cleanDir: true,
+      format: "es",
+      banner: isReactPackage ? '"use client";' : undefined,
```
diegohaz: `"use client"` banner leaks into emitted `.d.ts` files for React packages.

The `banner` is set on the outer Rolldown `output` config, so it applies to **every** asset produced by the bundle — including the `.d.ts` files that the `rolldown-plugin-dts` plugin emits in the same build. Verified locally:

```sh
$ head -1 packages/ariakit-react-utils/dist/index.d.ts packages/ariakit-react-store/dist/index.d.ts
==> packages/ariakit-react-utils/dist/index.d.ts <==
"use client";
==> packages/ariakit-react-store/dist/index.d.ts <==
"use client";
```

Why it matters:
- `"use client"` is a runtime directive consumed by React Server Components / bundlers. It has no meaning in a `.d.ts` and just ships bytes plus a confusing line on every typegen output.
- The previous esbuild/tsup setup used `options.banner = { js: '"use client";' }` which was JS-only and avoided this.

Suggested fix: split the banner into a JS-only path. For example, drop the global `banner` and inject the directive only on JS assets via a small Rolldown `renderChunk` plugin, or pass an explicit banner config to `dts(...)` that overrides to empty (rolldown-plugin-dts exposes an `output`-style override).

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,299 @@
...
+      sourcemap: true,
+    },
+    plugins: [
+      ...(isSolid ? [solidPlugin({ solid: { generate: "dom" } })] : []),
+      dts({
+        tsconfig: getTsconfig(rootPath),
+        compilerOptions: {
+          customConditions: ["ariakit-source"],
+        },
+      }),
```
diegohaz, line 295: Emitted `.d.ts` files reference `.d.ts.map` sourcemap files that are not generated.

Every bundled `.d.ts` produced through this `dts(...)` plugin call ends with a sourceMappingURL pointing to a `.d.ts.map` sibling that doesn't exist:

```sh
$ tail -1 packages/ariakit-utils/dist/index.d.ts packages/ariakit-store/dist/index.d.ts packages/ariakit-react-store/dist/index.d.ts packages/ariakit-solid-utils/dist/index.d.ts
==> .../ariakit-utils/dist/index.d.ts <==
//# sourceMappingURL=index.d.ts.map
==> .../ariakit-store/dist/index.d.ts <==
//# sourceMappingURL=index.d.ts.map
==> .../ariakit-react-store/dist/index.d.ts <==
//# sourceMappingURL=index.d.ts.map
==> .../ariakit-solid-utils/dist/index.d.ts <==
//# sourceMappingURL=index.d.ts.map

$ find packages/ariakit-utils/dist packages/ariakit-store/dist -name '*.d.ts.map'
# (nothing)
```

IDEs and tools that try to resolve that URL will silently fail; consumers also get a noisy comment in their published `.d.ts`. The multi-entry build for `@ariakit/test` doesn't emit the URL line at all, only the bundled single-entry builds do, so the issue is specific to the dts plugin config here.

Fix: either pass `sourcemap: true` (or equivalent) to `dts({ ... })` so that `.d.ts.map` files actually get written, or disable declaration sourcemaps on the dts plugin so the trailing comment isn't emitted. Worth filing/checking against `rolldown-plugin-dts` upstream too.

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,299 @@
...
+  publicFiles: PublicFile[],
+) {
+  const packageJson = readPackageJson(rootPath);
+  const isSolid = packageJson.name.includes("/solid");
+  const exports = Object.fromEntries(
+    publicFiles.map((file) => [
+      getExportName(file.name),
+      getExportValue(file, isSolid),
+    ]),
+  );
```
diegohaz, line 171: Exports map order: the `.` (root) entry lands alphabetically in the middle of the map instead of staying with the other root entries.

`updatePackageExports` builds `exports` from `publicFiles`, which are sorted by source filename in `getPublicFiles`. `getExportName` turns `index` into `"."`, but the sort key is still `"index"`, so `"."` slots between subpaths starting with `h*` and `m*`. You can see the resulting churn in this PR's diff for `@ariakit/core` and `@ariakit/test`, where `"."` is sandwiched between `"./hovercard/hovercard-store"` and `"./menu/menu-store"`, or between `"./hover"` and `"./mouse-down"`:

```json
"./hovercard/hovercard-store": { ... },
".": { ... },
"./menu/menu-bar-store": { ... },
```

This is purely cosmetic — Node resolves exports by exact key — but every package.json regeneration that adds or removes a public file risks a noisy reorder around the `"."` line, and the maps are harder to read at a glance.

Suggested fix: sort by the exported subpath (e.g., the value of `getExportName(file.name)`), which naturally places `"."` before any `"./..."`. Or special-case the index entry and prepend it explicitly:

```ts
const sorted = [...publicFiles].sort((a, b) =>
  getExportName(a.name).localeCompare(getExportName(b.name)),
);
const exports = Object.fromEntries(
  sorted.map((file) => [getExportName(file.name), getExportValue(file, isSolid)]),
);
```

.changeset/200-esm-only-existing-packages.md
```diff
@@ -0,0 +1,22 @@
+---
+"@ariakit/solid": minor
+"@ariakit/test": minor
+---
+
+Removed CommonJS builds from these packages. They now publish ESM-only exports with the `ariakit-source` condition.
```
diegohaz: Multi-paragraph changeset missing a title.

Per `ariakit-general-changeset`: "For changesets with multiple paragraphs or examples, the first paragraph must be a short title, not a full sentence. It should not end with a period, should not contain links, and should stay concise."

This entry has multiple paragraphs and code blocks but jumps straight into a full sentence ending in a period. The other two breaking changesets in this PR follow the rule (`Removed core utility re-export paths`, `Added standalone utility and store packages`), so this one is the odd one out and will read inconsistently in the published changelog.

Suggested fix:

```diff
----
-Removed CommonJS builds from these packages. They now publish ESM-only exports with the `ariakit-source` condition.
+Removed CommonJS builds
+
+These packages no longer publish CommonJS, and now ship ESM-only exports with the `ariakit-source` condition.
```

packages/ariakit-solid-store/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "clean": "ariakit build --clean"
+  },
+  "dependencies": {
+    "@ariakit/store": "workspace:*"
+  },
+  "peerDependencies": {
+    "solid-js": "^1.8"
+  },
+  "exports": {
+    ".": {
```
diegohaz: `@ariakit/solid-store` is a stub today, but the package metadata is shaped as if it were a real Solid package.

- `src/index.ts` is literally `export * from "@ariakit/store";` — nothing imports `solid-js` or uses Solid primitives.
- The package declares `peerDependencies: { "solid-js": "^1.8" }`, which forces consumers to satisfy a peer that the runtime never needs and breaks installs in environments that don't have Solid.
- The build script keys off `name.includes("/solid")`, so it adds the `"solid": "./solid/index.jsx"` export and emits `solid/index.jsx` — but that file is identical to `dist/index.js` and just lengthens the published tarball.
- `tsconfig.json` extends `../../tsconfig.node.json`, while every other Solid workspace (`@ariakit/solid`, `@ariakit/solid-core`, `@ariakit/solid-utils`) extends `../../tsconfig.solid.json`. This is inconsistent enough to bite when a future PR adds real Solid code here.

Two defensible directions:
1. Drop the package and tell Solid consumers to import `@ariakit/store` directly, the same way Solid consumers of `@ariakit/utils` do. The internal monorepo doesn't consume it either (grep returns nothing under `packages/`, `examples/`, `app/`, `nextjs/`).
2. Keep the package but make it a real Solid surface: drop the unused peer dep until Solid code lands, extend `tsconfig.solid.json`, and gate the solid build path on whether the package actually has JSX (e.g., compute `isSolid` from `hasJsxPublicFile` rather than the package name).

The other new packages (`@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`) all consume real implementation files, so this finding is specific to `solid-store`.

packages/ariakit-utils/readme.md
```diff
@@ -0,0 +1 @@
+# @ariakit/utils
```
diegohaz, line 1: Empty readme for a newly published public package.

With the minor changeset, this ships as `@ariakit/utils@0.1.0` on npmjs.com. The current readme is just `# @ariakit/utils` — npm will show no description, no install snippet, no usage example. Same story for the other new packages (`@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`) and partially for `@ariakit/solid-store` (which at least has a one-line description).

Even a short stub — what the package is, the single public entrypoint, the `ariakit-source` condition story, and a link back to `ariakit.com` — would prevent the npm landing page from looking abandoned at launch. If the long-form docs are intentionally on the docs site, a one-paragraph pointer to that location is enough.

diegohaz (commented):
Initial review against the baseline `origin/main`. Diff scope: 375 files, +3983 / -1720 — six new published packages (`@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`, `@ariakit/solid-store`), a new `ariakit build` Rolldown-powered CLI in `@ariakit/scripts`, the removal of the old core utility re-export shims, ESM-only conversion for several existing packages, and tooling/config wiring for the new `ariakit-source` condition.

## What I ran

- `pnpm install` (clean)
- `pnpm tsc` — passed
- `pnpm build` — passed end-to-end (all internal packages, plus the legacy `@ariakit/react` CJS+ESM build that bundles the new pure-ESM packages via `noExternal`)
- `pnpm test` — 805 tests across 151 files, all passing
- `pnpm -F @ariakit/utils run clean package.json && pnpm -F template-react build` — passed (`ariakit-source` condition resolves to source files cleanly even with `dist/` deleted)
- `node -e "require('./packages/ariakit-react/cjs/index.cjs'); require('./packages/ariakit-react/cjs/store.cjs')"` — passed, CJS bundle exposes the right exports including `useStoreState`
- `node --input-type=module -e 'await import("./packages/ariakit-test/dist/react.js")'` — passed
- ESM smoke imports against the new pure-ESM packages' `dist/index.js`, plus an `npm pack --dry-run --json` check of `@ariakit/utils` and `@ariakit/test` to confirm the published tarball shape (sources + dist + readme + license).

## Findings

All specific, line-anchored findings are in inline comments. Categories:

- **Build output bugs (`packages/ariakit-scripts/src/build.ts`):**
  - The `"use client"` banner leaks into emitted `.d.ts` files for React packages (confirmed in `@ariakit/react-utils/dist/index.d.ts` and `@ariakit/react-store/dist/index.d.ts`).
  - All bundled `.d.ts` outputs from the dts plugin reference a `.d.ts.map` sibling that is never emitted, leaving broken sourceMappingURL trailers in the published types.
  - The generated `exports` map sorts the root `"."` entry into the middle of the alphabet (slotted under `"index"`) instead of grouping it with the other top-level entries, producing noisy diffs across regenerations.
- **Stub-shaped package (`@ariakit/solid-store`):** declares a `solid-js` peer dependency it doesn't use, gets a `solid` export condition and `solid/` build output despite shipping no JSX, and extends `tsconfig.node.json` while the rest of the Solid family extends `tsconfig.solid.json`. Either drop it or invest in real Solid integrations.
- **Changeset format (`.changeset/200-esm-only-existing-packages.md`):** missing the title line required for multi-paragraph entries; the other two breaking changesets in this PR follow the rule.
- **Docs surface:** new public packages ship with one-line readmes that will look abandoned on npmjs.com at first publish.

No correctness issues found in the package split itself: the relocation of utility/store implementations into the new packages preserved imports across `@ariakit/core`, `@ariakit/react-core`, `@ariakit/solid-core`, and `@ariakit/test`; the `@ariakit/react-core/store` → `@ariakit/react-store` re-export chain keeps `useStoreState` working through `@ariakit/react`; the CJS `noExternal` regex covers all six pure-ESM internal packages so the `@ariakit/react` CJS bundle still loads via `require()`; and the `ariakit-source` condition is wired consistently in `vitest.config.ts`, `app/astro.config.ts`, `nextjs/next.config.ts`, `templates/react/{tsconfig,vite.config}`, the root playwright script, and the package exports themselves.

## Areas not validated

- Visual / browser playwright suites (skipped per `ariakit-ariakit-workflow` — VISUAL_TEST is CI-only).
- Full `pnpm build-app-lite` / `pnpm build-website-lite`. The Vitest, root `pnpm build`, and the standalone `template-react` build cover the same dependency-resolution path under the new `ariakit-source` condition.
- Solid runtime checks beyond the type-check and build outputs — I didn't run a Solid app against the new `solid-utils`/`solid-store`.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in 3d342f01c - moved the React `"use client"` directive into a JS-only Rolldown `renderChunk` plugin and verified declaration files no longer include it.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in 3d342f01c - enabled declaration sourcemaps for the DTS build and verified every emitted `.d.ts` sourceMappingURL now has a matching `.d.ts.map` file.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in 3d342f01c - sorted generated exports by exported subpath, so the root `.` entry is emitted first, then regenerated the affected package export maps.

.changeset/200-esm-only-existing-packages.md
diegohaz: Fixed in 3d342f01c - added the short changeset title and moved the ESM-only details into the following paragraph.

packages/ariakit-solid-store/package.json
diegohaz: Fixed in 3d342f01c - made `@ariakit/solid-store` match its current implementation by removing the unused `solid-js` peer, the `solid` export condition, and the generated solid output path until it has real Solid runtime code.

packages/ariakit-utils/readme.md
diegohaz: Fixed in 3d342f01c - added short npm-facing readmes for all six new packages with install snippets, root-import examples, and the ESM/source-condition note.

diegohaz:
**Fixed**

- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287796898)): Moved the React `"use client"` directive to a JS-only Rolldown plugin so declaration files stay clean.
- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287796993)): Enabled declaration sourcemaps and verified `.d.ts` sourceMappingURL comments point to emitted `.d.ts.map` files.
- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287797132)): Sorted generated export maps by exported subpath and regenerated affected package manifests.
- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287797239)): Updated the ESM-only changeset to use the required title format.
- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287797428)): Removed the unused Solid peer, `solid` condition, and generated solid output from the current `@ariakit/solid-store` alias package.
- 3d342f01c ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3287797589)): Added npm-facing readmes for all six new utility/store packages.

diegohaz (commented):
Re-review against the latest head `3d342f01`, baselined on the previous reviewed SHA `23b7d367`. Two new commits in scope: `92237e75 Fix source resolution in CI` and `3d342f01 Address package split review feedback`.

## Verification of prior findings

All six earlier findings are addressed and verified locally:

- **`"use client"` in `.d.ts`** — replaced the output banner with a dedicated `renderChunk` plugin that returns `null` for any chunk whose filename ends with `.d.ts`. Confirmed: `head -1 packages/ariakit-react-utils/dist/index.d.ts packages/ariakit-react-store/dist/index.d.ts` no longer shows the directive, and the corresponding `.js` files and chunk JS still get the banner; declaration chunks under `dist/__chunks/*.d.ts` are clean too. Sourcemap mappings for the banner are preserved (the leading `;` count in `index.js.map` shifted to account for the inserted line, so source positions still line up).
- **Broken `.d.ts.map` URLs** — `dts({ sourcemap: true, ... })` now emits `.d.ts.map` siblings (verified for `@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, `@ariakit/react-store`, `@ariakit/solid-utils`); the maps parse as valid JSON with the expected `sources`.
- **Exports map sort** — new `sortPublicFilesByExportName` sorts by `getExportName(file.name)`, so every regenerated `exports` object now has `"."` as the first key. Confirmed against `@ariakit/core`, `@ariakit/react-core`, `@ariakit/test`, `@ariakit/solid`.
- **Changeset title** — `.changeset/200-esm-only-existing-packages.md` now leads with `Removed CommonJS builds` on its own line; the descriptive sentence moved to the following paragraph, matching the other two breaking entries.
- **`@ariakit/solid-store` shape** — `isSolidPackage` now requires both the `/solid` substring and a `solid-js` package dependency, the package.json dropped the unused `solid-js` peer and the `solid` export condition, the `solid/` build output is no longer produced, and the `tsconfig.json` was switched to extend `../../tsconfig.solid.json` (matching the rest of the Solid family).
- **Empty readmes** — all six new packages now ship a short readme with install snippet, root-import example, and the ESM/`ariakit-source` condition note.

Local validation: `pnpm install`, `pnpm tsc`, `pnpm build`, `pnpm test` (805 / 805 pass), inspected every regenerated `exports` map, the dist outputs for `.d.ts`/`.d.ts.map`/`.js`/`.js.map`, the new `scripts/ariakit-source.ts` alias output (315 entries with subpath aliases ordered ahead of root entries so the more-specific match wins under Webpack's prefix-matching alias semantics), and the new `app/astro.config.ts` / `nextjs/next.config.ts` integrations.

## New scope from the two follow-up commits

- New helper `scripts/ariakit-source.ts` walks `packages/*/package.json` and produces an alias map keyed by `@ariakit/<pkg>[/subpath]` pointing at the source file referenced by each `ariakit-source` export. Used by Vite (regex/exact, via `getAriakitSourceAliasEntries`) and Webpack/Turbopack (string keys, via `getAriakitSourceAliases`).
- `oxlint.config.ts` gains an `examples/**/*.tsx` override that disables `no-redundant-type-constituents`. Working as a clean-checkout workaround; the override's comment talks about "namespace imports" not resolving the source condition, which is the root cause but doesn't quite spell out the rule's connection to it. Cosmetic, not actionable.
- The legacy `@ariakit/react` (still on the old build script) is not present in the alias map on a clean checkout because its committed `exports` are still plain strings without an `ariakit-source` condition; Vite resolves it via the string entry instead, which works but means the alias map is volatile between pre-build and post-build state. No functional issue observed.

## Status

No new actionable findings on this pass. Everything that was flagged previously has been verified resolved against the current head.

.changeset/200-esm-only-existing-packages.md
```diff
@@ -0,0 +1,24 @@
+---
+"@ariakit/solid": minor
+"@ariakit/test": minor
+---
+
+Removed CommonJS builds
+
+These packages no longer publish CommonJS, and now ship ESM-only exports with the `ariakit-source` condition.
+
+**BREAKING** if your code loads these packages with CommonJS `require()`.
```
diegohaz: No need to mention the `ariakit-source` condition. Also, the **BREAKING** note should be the first paragraph below the title `Removed CommonJS builds`

.changeset/200-esm-only-existing-packages.md
```diff
@@ -0,0 +1,24 @@
+---
+"@ariakit/solid": minor
+"@ariakit/test": minor
+---
```
diegohaz: I believe we should have separate changesets for these packages.

app/src/components/code-block.react.tsx
```diff
@@ -1,3 +1,4 @@
+import * as ak from "@ariakit/react";
```
diegohaz: Import declarations should not appear above the license comment. Also, look for other instances where this might be happening.

app/src/components/reference-hovercard.react.tsx
```diff
@@ -1,3 +1,4 @@
+import * as ak from "@ariakit/react";
```
diegohaz: Same

app/astro.config.ts
```diff
@@ -58,6 +61,29 @@ export default defineConfig({
...
+      esbuildOptions: {
+        conditions,
+      },
+    },
+    ssr: {
+      resolve: {
+        conditions,
+        externalConditions: conditions,
+      },
+    },
```
diegohaz: Do we need all of this?

app/astro.config.ts
```diff
@@ -18,6 +18,7 @@ import clerk from "@clerk/astro";
 import tailwindcss from "@tailwindcss/vite";
 import { defineConfig } from "astro/config";
 import rehypeAutolinkHeadings from "rehype-autolink-headings";
+import { getAriakitSourceAliasEntries } from "../scripts/ariakit-source.ts";
```
diegohaz: We should import it from `@ariakit/scripts/ariakit-source`

nextjs/next.config.ts
```diff
@@ -1,7 +1,33 @@
...
+    config.resolve.alias = {
+      ...config.resolve.alias,
+      ...ariakitSourceAliases,
+    };
+    config.resolve.conditionNames = [
+      "ariakit-source",
+      ...(config.resolve.conditionNames ?? ["..."]),
+    ];
+    return config;
+  },
```
diegohaz: Do we need this if we're only using turbopack here?

packages/ariakit-react-core/src/composite/composite-container.tsx
```diff
@@ -1,18 +1,18 @@
+import { useStoreState } from "@ariakit/react-store";
+import { useEvent, useMergeRefs } from "@ariakit/react-utils";
+import { createElement, createHook, forwardRef } from "@ariakit/react-utils";
+import type { Options, Props } from "@ariakit/react-utils";
 // TODO: Add data-attribute to indicate whether it's expanded?
```
diegohaz: This comment should be in the first line.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -1,13 +1,20 @@
...
+  useWrapElement,
+} from "@ariakit/react-utils";
+import { createElement, createHook, forwardRef } from "@ariakit/react-utils";
+import type { Props } from "@ariakit/react-utils";
+import { flatten2DArray, reverseArray } from "@ariakit/utils";
+import { getActiveElement, isTextField } from "@ariakit/utils";
+import { fireBlurEvent, fireKeyboardEvent, isSelfTarget } from "@ariakit/utils";
+import { focusIntoView, hasFocus } from "@ariakit/utils";
+import { invariant } from "@ariakit/utils";
+import type { BooleanOrCallback } from "@ariakit/utils";
```
diegohaz, line 26: These should be two imports, one for all runtime Ariakit utils and one for types. Make sure to find other instances where this might be happening too. We need to dedupe those imports.

.changeset/200-esm-only-existing-packages.md
diegohaz: Fixed in 802388f2e - removed the `ariakit-source` sentence from the ESM-only changesets and moved the **BREAKING** note directly below the title.

.changeset/200-esm-only-existing-packages.md
diegohaz: Fixed in 802388f2e - split the combined ESM-only changeset into separate `@ariakit/solid` and `@ariakit/test` entries.

app/src/components/code-block.react.tsx
diegohaz: Fixed in 802388f2e - moved the import below the license header and checked TS/TSX files for other license headers after imports.

app/src/components/reference-hovercard.react.tsx
diegohaz: Fixed in 802388f2e - moved the import below the license header and verified no TS/TSX license headers remain after imports.

app/astro.config.ts
diegohaz: Fixed in 802388f2e - simplified Astro to use the generated aliases only and removed the extra `ariakit-source` condition plumbing.

app/astro.config.ts
diegohaz: Fixed in 802388f2e - moved the source alias helper under `@ariakit/scripts` and updated Astro to import `@ariakit/scripts/ariakit-source`.

nextjs/next.config.ts
diegohaz: Fixed in 802388f2e - removed the webpack alias/condition block from the Next config and kept the Turbopack resolver aliases only.

packages/ariakit-react-core/src/composite/composite-container.tsx
diegohaz: Fixed in 802388f2e - moved the TODO comment to the first line.

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Fixed in 802388f2e - deduped Ariakit utility imports across the split packages, keeping runtime imports separate from type imports.

diegohaz:
**Fixed**

- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288091588)): Removed the `ariakit-source` sentence from the ESM-only changesets and moved the **BREAKING** note directly below the title.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288091732)): Split the combined ESM-only changeset into separate `@ariakit/solid` and `@ariakit/test` entries.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288091860)): Moved the code block import below the license header and checked for other TS/TSX license headers after imports.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288091979)): Moved the reference hovercard import below the license header and verified the license/import invariant.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288092102)): Simplified Astro to use generated aliases only and removed the extra `ariakit-source` condition plumbing.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288092198)): Moved the source alias helper under `@ariakit/scripts` and updated Astro to import `@ariakit/scripts/ariakit-source`.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288092290)): Removed the Next webpack alias/condition block and kept the Turbopack resolver aliases.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288092422)): Moved the TODO comment to the first line.
- 802388f2e ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3288092518)): Deduped Ariakit utility imports across the split packages, keeping runtime imports separate from type imports.

packages/ariakit-react-components/src/form/form.tsx
```diff
@@ -1,24 +1,26 @@
...
+import { isTextField, invariant } from "@ariakit/utils";
+import type { ElementType, FocusEvent, FormEvent } from "react";
+import { useEffect, useRef, useState } from "react";
 import { FormScopedContextProvider, useFormContext } from "./form-context.tsx";
 import type { FormStore, FormStoreState } from "./form-store.ts";
 
 const TagName = "form" satisfies ElementType;
 type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
+type FormSubmitEvent = Parameters<NonNullable<FormProps["onSubmit"]>>[0];
```
diegohaz: We do we need this?

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,333 @@
...
+  const exports = Object.fromEntries(
+    sortedPublicFiles.map((file) => [
+      getExportName(file.name),
+      getExportValue(file, isSolid),
+    ]),
+  );
+
+  delete packageJson.main;
+  delete packageJson.module;
+  delete packageJson.types;
```
diegohaz: Do we really need this here?

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,333 @@
...
+    "*.config.*",
+    "*.lock",
+  ];
+  writeFileSync(join(rootPath, ".npmignore"), `${contents.join("\n")}\n`);
+}
+
+function getTsconfig(rootPath: string) {
+  for (const filename of [
+    "tsconfig.build.json",
+    "tsconfig.node.json",
```
diegohaz: Do we need `tsconfig.build.json` and `tsconfig.node.json` here? I noticed the new packages don't have a `tsconfig.build.json`. Could we just let Rolldown pick up `tsconfig.json` automatically?

packages/ariakit-scripts/tsconfig.node.json
```diff
@@ -1,5 +1,10 @@
 {
   "extends": "../../tsconfig.node.json",
-  "include": ["src"],
+  "files": [
+    "src/ariakit-source.ts",
+    "src/build.ts",
+    "src/dev.ts",
+    "src/index.ts"
+  ],
```
diegohaz: Why not just `"include": ["src"]`?

templates/react/vite.config.ts
```diff
@@ -6,4 +6,7 @@ import { defineConfig } from "vite";
 export default defineConfig({
   // @ts-ignore
```
diegohaz, line 7: Is this still needed?

website/components/header-nav.tsx
```diff
@@ -1,15 +1,15 @@
...
   ComboboxItemValue,
   PopoverDisclosureArrow,
   PopoverDismiss,
 } from "@ariakit/react";
 import type { SelectRendererItem } from "@ariakit/react-core/select/select-renderer";
 import { SelectRenderer } from "@ariakit/react-core/select/select-renderer";
-import { useEvent, useSafeLayoutEffect } from "@ariakit/react-core/utils/hooks";
+import { useEvent, useSafeLayoutEffect } from "@ariakit/react-utils";
+import { getKeys } from "@ariakit/utils";
+import { isApple } from "@ariakit/utils";
```
diegohaz: Dedupe imports (also find other instances where it needs to be deduped in the website workspace).

website/next.config.js
```diff
@@ -34,6 +34,11 @@ const nextConfig = {
     ignoreBuildErrors: true,
   },
   webpack(config, context) {
+    config.resolve.conditionNames = [
+      "ariakit-source",
+      ...(config.resolve.conditionNames ?? ["..."]),
```
diegohaz: What does `["..."]` mean?

diegohaz:
Pushed `ce9578505` with the `core` → `components` package rename: `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/solid-components` now start at `0.0.0` with cleaned changelogs, old core package paths are removed from git, imports/workspace references were updated, and internal package READMEs now include the semver warning except for `@ariakit/react`. I also updated the PR title and description to cover this scope.

vitest.config.ts
```diff
@@ -48,12 +48,18 @@ const PLUGINS_BY_LOADER: Record<string, Array<PluginOption> | undefined> = {
...
       { find: /^react-dom($|\/)/, replacement: `${reactDomDir}$1` },
       { find: /^react($|\/)/, replacement: `${reactDir}$1` },
     ],
     dedupe: ["react", "react-dom"],
   },
+  optimizeDeps: {
+    esbuildOptions: {
+      conditions: ["ariakit-source"],
+    },
+  },
```
diegohaz: Do we need this?

.changeset/300-menu-autofocus-on-show.md
```diff
@@ -1,5 +1,4 @@
 ---
-"@ariakit/react-core": patch
```
diegohaz, line 2: This should be `"@ariakit/react-components": patch`

app/src/lib/reference-tokenizer.ts
```diff
@@ -145,7 +145,7 @@ function parseAriakitImports(code: string): ImportInfo {
   let hasAriakitImport = false;
 
   const importRegex =
-    /import\s+([\s\S]*?)\s+from\s*["'](@ariakit\/[\w-]+)(?:-core)?["']/g;
+    /import\s+([\s\S]*?)\s+from\s*["'](@ariakit\/[\w-]+)["']/g;
```
diegohaz: What was this regex doing? Is it safe to remove `core` from here?

app/src/lib/reference-tokenizer.ts
```diff
@@ -1089,7 +1089,7 @@ function processNamedImports(
   ) => string | undefined,
 ) {
   const importRegex =
-    /import\s+([\s\S]*?)\s+from\s*["'](@ariakit\/[\w-]+)(?:-core)?["']/g;
+    /import\s+([\s\S]*?)\s+from\s*["'](@ariakit\/[\w-]+)["']/g;
```
diegohaz: Same question here.

packages/ariakit-scripts/src/index.ts
```diff
@@ -1,10 +1,20 @@
...
 import { program } from "commander";
+import { build } from "./build.ts";
 import { dev } from "./dev.ts";
 
 program.name("ariakit");
 
+program
+  .command("build")
+  .description("Build packages")
+  .argument("[files...]", "Ignored staged files")
```
diegohaz: What does this argument mean? Do we need to document this?

packages/ariakit-solid/readme.md
```diff
@@ -6,6 +6,8 @@
   <a href="https://solid.ariakit.com"><strong>Explore website »</strong></a>
 </p>
 
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions. You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.
```
diegohaz: No need to mention `You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.` here because it's a solid package, not react.

Review the other non-react packages to see if they are suggesting to use the react package as well. Update them.

packages/ariakit-tailwind/readme.md
```diff
@@ -1,5 +1,7 @@
 # Ariakit Tailwind
 
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions. You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.
```
diegohaz: Same here.

packages/ariakit-tailwind/readme.md
```diff
@@ -1,5 +1,7 @@
...
+
 > **Experimental.** This package is currently experimental. Breaking changes might be introduced in patch or minor releases.
```
diegohaz: Looks like this note is duplicating the above one. Look for similar patterns in other readmes and update all of them.

packages/ariakit-test/readme.md
```diff
@@ -1,6 +1,8 @@
 # Ariakit Test
 
-**Important:** This package is still experimental. Breaking changes may occur in patch and minor versions.
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions. You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.
+
+> **Experimental.** This package is still experimental. Breaking changes may occur in patch and minor versions.
```
diegohaz: Same here

packages/ariakit-test/tsconfig.build.json
```diff
@@ -1,9 +1,5 @@
 {
   "extends": "../../tsconfig.react.json",
-  "compilerOptions": {
-    "declaration": true,
-    "rootDir": "src"
-  },
```
diegohaz, line 6: Do we need `tsconfig.build.json` if it's the same as `tsconfig.json`?

packages/ariakit-react-components/src/form/form.tsx
diegohaz: Fixed in 8907116f2 - removed the extra derived submit event alias and typed the handler directly as `FormEvent<HTMLType>`.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in 8907116f2 - removed the `main`/`module`/`types` deletion from `updatePackageExports`; the pure ESM package manifests already omit those fields.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in 8907116f2 - removed the custom tsconfig probing so Rolldown/DTS can discover `tsconfig.json` directly. `rolldown-plugin-dts` now runs with `build: true` and `incremental: false` so project-reference configs like `@ariakit/test` still emit declarations without writing build artifacts.

packages/ariakit-scripts/tsconfig.node.json
diegohaz: Fixed in 8907116f2 - simplified the scripts Node tsconfig to `include: ["src"]` while keeping the explicit empty `exclude` override.

templates/react/vite.config.ts
diegohaz: Fixed in 8907116f2 - removed the React template Vite `resolve.conditions` block; the template should use built/published exports.

website/components/header-nav.tsx
diegohaz: Fixed in 8907116f2 - deduped the `@ariakit/utils` imports in `header-nav.tsx` and checked the website workspace for duplicate runtime import sources.

website/next.config.js
diegohaz: Fixed in 8907116f2 - documented that Webpack uses `...` to preserve its default export conditions. I also verified removing it breaks the website build for packages with exports maps, so the sentinel needs to stay here.

vitest.config.ts
diegohaz: Fixed in 8907116f2 - removed the redundant Vitest `optimizeDeps.esbuildOptions.conditions` block; `resolve.conditions` is enough for the source condition in tests.

.changeset/300-menu-autofocus-on-show.md
diegohaz: Fixed in 8907116f2 - added `"@ariakit/react-components": patch` to the menu autofocus changeset alongside `@ariakit/react`.

app/src/lib/reference-tokenizer.ts
diegohaz: Fixed in 8907116f2 - extracted the Ariakit package import regex into a named helper and removed the unused package-name capture group. The old `core` suffix handling was unnecessary here because both call sites only use the import clause capture.

app/src/lib/reference-tokenizer.ts
diegohaz: Fixed in 8907116f2 - reused the same named Ariakit package import regex helper here too, so both parsing paths share the same documented package-root import behavior.

packages/ariakit-scripts/src/index.ts
diegohaz: Fixed in 8907116f2 - documented that the positional file arguments are accepted only because lint-staged appends filenames to package scripts, and the build command intentionally ignores them.

packages/ariakit-solid/readme.md
diegohaz: Fixed in 8907116f2 - removed the `@ariakit/react` recommendation from the Solid README and swept the other non-React package READMEs for the same wording.

packages/ariakit-tailwind/readme.md
diegohaz: Fixed in 8907116f2 - changed the Tailwind Important note so it says the package is experimental instead of calling it an internal dependency, and removed the React recommendation.

packages/ariakit-tailwind/readme.md
diegohaz: Fixed in 8907116f2 - removed the duplicate experimental blockquote wording from Tailwind/Test by folding the experimental warning into the single Important note.

packages/ariakit-test/readme.md
diegohaz: Fixed in 8907116f2 - changed the Test README Important note to the experimental wording and removed the duplicate experimental blockquote.

packages/ariakit-test/tsconfig.build.json
diegohaz: Fixed in 8907116f2 - deleted `packages/ariakit-test/tsconfig.build.json`; the new build now relies on tsconfig auto-discovery plus DTS build mode for project references.

diegohaz:
**Fixed**

- 8907116f2 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825002), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289826127)): Removed the redundant form submit event alias and added the missing `@ariakit/react-components` changeset entry.
- 8907116f2 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825129), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825268), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825399), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289830854), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289831771)): Simplified the new build path by removing manifest field deletion, dropping custom tsconfig probing, simplifying the scripts tsconfig, documenting lint-staged file args, and deleting the redundant `@ariakit/test` build config.
- 8907116f2 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825526), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825855), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825981)): Removed unneeded template/Vitest source-condition config and documented the Webpack `...` condition fallback.
- 8907116f2 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289825694), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289826268), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289830633)): Deduped website imports and extracted the Ariakit package import regex helper.
- 8907116f2 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289831043), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289831157), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289831316), [reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3289831518)): Corrected README Important notes for Solid/non-React packages and consolidated Tailwind/Test experimental wording.

app/astro.config.ts
```diff
@@ -33,6 +34,7 @@ try {
 
 const port = Number(process.env.APP_PORT) || 4321;
 const hasClerk = process.env.PUBLIC_CLERK_PUBLISHABLE_KEY;
+const alias = getAriakitSourceAliasEntries();
```
diegohaz: Investigate if/why Astro doesn't support ESM export conditions.

nextjs/next.config.ts
```diff
@@ -1,7 +1,21 @@
...
+const ariakitTurbopackAliases = Object.fromEntries(
+  Object.entries(ariakitSourceAliases).map(([specifier, replacement]) => [
+    specifier,
+    relative(import.meta.dirname, replacement).replaceAll("\\", "/"),
+  ]),
+);
+
 const config: NextConfig = {
+  turbopack: {
+    resolveAlias: ariakitTurbopackAliases,
```
diegohaz: Investigate if/why turbopack doesn't support ESM export conditions.

packages/ariakit-components/readme.md
```diff
@@ -1,11 +1,11 @@
-# Ariakit React Core
+# Ariakit Components
 
-**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions. You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions.
```
diegohaz: You shouldn't have removed this from this readme: ` You probably want to use [`@ariakit/react`](https://npmjs.org/package/@ariakit/react) instead.`

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,306 @@
...
+async function buildSolidSource(
+  rootPath: string,
+  input: Record<string, string>,
+) {
+  await rolldownBuild({
+    input,
+    external: getExternal(readPackageJson(rootPath)),
+    transform: {
+      jsx: "preserve",
+    },
```
diegohaz: Do we need this? Won't the tsconfig already do this?

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,306 @@
...
+    isReactPackage || (!isSolid && hasJsxPublicFile(publicFiles));
+
+  await rolldownBuild({
+    input: getInput(publicFiles),
+    external: getExternal(packageJson),
+    transform: shouldTransformReactJsx
+      ? {
+          jsx: "react-jsx",
+        }
+      : undefined,
```
diegohaz: Is this necessary?

packages/ariakit-solid/readme.md
```diff
@@ -6,6 +6,8 @@
   <a href="https://solid.ariakit.com"><strong>Explore website »</strong></a>
 </p>
 
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions.
```
diegohaz: `This package is experimental`, not an internal dependency.

packages/ariakit-store/readme.md
```diff
@@ -0,0 +1,17 @@
+# @ariakit/store
+
+**Important:** This package is an internal dependency of Ariakit and does not follow semantic versioning, meaning breaking changes may occur in patch and minor versions.
```
diegohaz: `This package is experimental`, not an internal dependency.

oxlint.config.ts
```diff
@@ -55,5 +57,13 @@ export default defineConfig({
         "no-redundant-type-constituents": "off",
       },
     },
+    {
+      // Oxlint doesn't resolve the Ariakit source condition for namespace
+      // imports in examples on a clean checkout before packages are built.
```
diegohaz: Investigate if/why oxlint doesn't support ESM export conditions.

app/astro.config.ts
diegohaz: Fixed in e540b2754 - Astro can use Vite export conditions here, so I replaced the alias workaround with `ariakit-source` conditions. The condition list preserves Vite client `browser`/`development|production` defaults while omitting `module` because that condition selects an unpatched Changesets export in tests.

nextjs/next.config.ts
diegohaz: Fixed in e540b2754 - kept the Turbopack alias workaround and documented why: Turbopack exposes `resolveAlias`, but not a custom package export condition option equivalent to Vite/Webpack yet.

packages/ariakit-components/readme.md
diegohaz: Fixed in e540b2754 - restored the `@ariakit/react` recommendation in the `@ariakit/components` Important note.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in e540b2754 - removed the explicit Solid source `jsx: "preserve"` transform; the Solid tsconfig covers this and the Solid builds pass without it.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in e540b2754 - removed the explicit React JSX transform from the build script and let Rolldown pick up package tsconfig references. To keep `@ariakit/test/react` valid, `src/react.tsx` now belongs to `tsconfig.react.json` instead of the node config; the emitted file imports in Node successfully.

packages/ariakit-solid/readme.md
diegohaz: Fixed in e540b2754 - changed the `@ariakit/solid` Important note to say the package is experimental instead of an internal dependency.

packages/ariakit-store/readme.md
diegohaz: Fixed in e540b2754 - changed the `@ariakit/store` Important note to the experimental wording.

oxlint.config.ts
diegohaz: Fixed in e540b2754 - kept the examples override and clarified the comment because Oxlint’s import plugin does not expose a custom package export condition setting for `ariakit-source`.

diegohaz:
**Fixed**

- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290282706)): Replaced Astro source aliases with `ariakit-source` export conditions while preserving Vite client browser/development conditions.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290282829)): Kept and documented the Turbopack alias fallback because custom package export conditions are not exposed there.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290282962)): Restored the `@ariakit/react` recommendation in the `@ariakit/components` README.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290283156)): Removed the explicit Solid source JSX transform and let tsconfig drive it.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290283271)): Removed the explicit React JSX transform from the shared build and partitioned `@ariakit/test` tsconfigs so `src/react.tsx` uses the React config.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290283389)): Changed the `@ariakit/solid` note to experimental wording.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290283490)): Changed the `@ariakit/store` note to experimental wording.
- e540b2754 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290283619)): Clarified the Oxlint examples override as a custom export-condition limitation.

packages/ariakit-scripts/src/index.ts
```diff
@@ -1,13 +1,32 @@
...
+import { build, clean } from "./build.ts";
 import { dev } from "./dev.ts";
 
 program.name("ariakit");
 
+program
+  .command("build")
+  .description("Build packages")
+  // lint-staged appends matched filenames to package scripts.
+  .argument("[files...]", "Ignored file arguments passed by lint-staged")
```
diegohaz: Do we still need this? Does the build command run with lint-staged?

diegohaz: Fixed in 6f922d121 - removed the ignored lint-staged file argument from `ariakit build` and simplified the command to `.action(build)`. `ariakit clean` still keeps the ignored file arguments because lint-staged runs clean for package.json changes.

diegohaz:
**Fixed**

- 6f922d121 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290987565)): removed the ignored lint-staged file argument from `ariakit build` and simplified the Commander action to `.action(build)`. `ariakit clean` still accepts ignored file arguments because lint-staged runs it for `package.json` changes.
- 6f922d121: moved `ariakit dev` off `dev-package-json`; it now cleans and watches packages through the clean logic directly, including legacy-clean packages like `@ariakit/react`. `--no-clean` preserves current package exports for built-output testing.
- 6f922d121: added `@ariakit/scripts` as a local dev dependency for packages that run the `ariakit` CLI and restored `website/build-pages/get-example-deps.js` to match `main`.

templates/react/package.json
```diff
@@ -10,9 +10,9 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@ariakit/core": "workspace:*",
+    "@ariakit/components": "workspace:*",
     "@ariakit/react": "workspace:*",
-    "@ariakit/react-core": "workspace:*",
+    "@ariakit/react-components": "workspace:*",
     "@ariakit/tailwind": "workspace:*",
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7d85591a9 - removed the unused `@ariakit/components` and `@ariakit/react-components` template dependencies and regenerated the lockfile importer.

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,349 @@
...
+    },
+    plugins: [
+      dts({
+        emitDtsOnly: true,
+        build: true,
+        incremental: false,
+        sourcemap: true,
+      }),
+    ],
+  });
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7d85591a9 - removed the ineffective `output.sourcemap` option from the DTS-only Rolldown pass while keeping `dts({ sourcemap: true })` for declaration maps.

packages/ariakit-scripts/src/build.ts
```diff
@@ -0,0 +1,349 @@
...
+  return Object.fromEntries(publicFiles.map((file) => [file.name, file.path]));
+}
+
+function getUseClientPlugin(enabled: boolean): Plugin[] {
+  if (!enabled) return [];
+  return [
+    {
+      name: "ariakit-use-client",
+      renderChunk(code, chunk) {
+        if (chunk.fileName.endsWith(".d.ts")) return null;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7d85591a9 - simplified the `use client` Rolldown plugin by removing the unreachable `.d.ts` guard from the JS-only build path.

packages/ariakit-scripts/src/dev.ts
```diff
@@ -26,7 +162,12 @@ async function findAvailablePort(start: number): Promise<number> {
   }
 }
 
-export async function dev() {
+export async function dev(options: DevOptions = {}) {
```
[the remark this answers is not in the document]
diegohaz: Fixed in 7d85591a9 - `ariakit dev` now detects the Astro app port, passes `APP_PORT` through the spawned process, and selects the Next.js port independently while only excluding the chosen app port to avoid collisions.

diegohaz:
**Fixed**

- 7d85591a9 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3291944377)): removed unused `@ariakit/components` and `@ariakit/react-components` dependencies from the React template and regenerated the lockfile importer.
- 7d85591a9 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3291944456)): removed the no-op `output.sourcemap` from the DTS-only Rolldown pass while keeping declaration sourcemaps on the DTS plugin.
- 7d85591a9 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3291944564)): simplified the `"use client"` Rolldown plugin by dropping the unreachable `.d.ts` guard.
- 7d85591a9 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3291944661)): made `ariakit dev` detect and pass `APP_PORT`, then select `NEXTJS_PORT` independently while avoiding only the selected app port.

**Already Resolved**

- 6f922d121 ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3290987565)): the stale `ariakit build` lint-staged argument thread was already addressed, and I resolved the thread.

scripts/build/package-json-dev.js
```diff
@@ -1,26 +1,38 @@
...
   if (path.includes("ariakit-tailwind")) return;
+  if (path.includes("ariakit-scripts")) return;
   const match = path.match(/packages\/(.*)\/src/);
   if (!match) return;
   const [, pkg] = match;
   if (!pkg) return;
   const pkgPath = join(process.cwd(), "packages", pkg);
+  const pkgJson = readPackageJson(pkgPath);
+  if (pkgJson.scripts?.build?.startsWith("ariakit build")) {
+    await updateSourcePackageJson(pkgPath);
```
diegohaz, line 18: Do we need this?

diegohaz: This is still needed because root `pnpm dev` runs `dev-package-json` for the website, and the website imports the new `ariakit build` packages directly (`@ariakit/utils`, `@ariakit/store`, `@ariakit/react-utils`, etc.). Those packages need the new source-export shape in dev; the old `writePackageJson` helper would regenerate legacy `main`/`module`/`types` fields and all-file exports for them.

diegohaz:
**Fixed**

- 19025c35d: removed the extra `--` from the app dev command so `pnpm -F app run dev --port ${appPort}` forwards `--port` to Astro correctly. Verified with `pnpm -F app run dev --help`.

**Invalid**

- No code change ([reply](https://github.com/ariakit/ariakit/pull/6018#discussion_r3292354307)): kept the `updateSourcePackageJson` path in `package-json-dev.js` because root `pnpm dev` still needs source exports for the new `ariakit build` packages used by the website.

PR #6041, opened 2026-05-24
Fix Tailwind inherit contexts

diegohaz:
## Motivation

`ak-edge-inherit` and `ak-frame-bordering-inherit` stopped at the nearest `ak-layer` or `ak-frame` context, even when that intermediate element only had the base utility and did not explicitly configure a frame border, ring, or bordering width. That made nested layers/frames lose the edge or bordering information from the first configured ancestor.

## Solution

Track explicit frame border/ring/bordering configuration separately from the base layer/frame context. Base `ak-layer`/`ak-frame` utilities now pass ancestor values through until a frame border, ring, or bordering width utility marks the current element as explicit, including explicit `0` values.

`ak-edge-inherit` uses that same explicit frame-width source, so edge modifiers alone no longer become inheritance anchors. `ak-frame-bordering-inherit` consumes the inherited width without becoming a new explicit source itself.

The sandbox snapshots cover direct inheritance, skipped intermediates, inherited-bordering intermediates, and explicit zero resets. The README descriptions now document the nearest explicit frame-width ancestor behavior.

## Testing

- `pnpm lint-fix`
- `APP_PORT=4322 pnpm -F app test-chrome ariakit-tailwind --grep-invert wcag -u`
- `APP_PORT=4322 pnpm -F app test-chrome ariakit-tailwind --grep-invert wcag`
- `APP_PORT=4322 pnpm -F app test-chrome ariakit-tailwind -g wcag`
- `pnpm tsc`
- `APP_PORT=4322 PERF_RESULTS_FILE=baseline pnpm -F app test-perf ariakit-tailwind`
- `APP_PORT=4322 pnpm -F app test-perf ariakit-tailwind`
- `pnpm -F app perf-compare`

### DIFF
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -591,6 +591,7 @@ const layerColorVars = {
   layerOffset: _ak.prop.canvas("lo"),
   layerPush: _ak.prop.canvas("lp"),
   layer: ak.prop.canvas("layer", { inherits: true }),
+  layerEdge: _ak.prop.black("le", { inherits: true }),
   layerParentContext: _ak.var("lpc"),
   layerEdgeContext: _ak.var("lec"),
   layerEdgeSourceContext: _ak.var("lesc"),
@@ -1297,6 +1298,7 @@ const inkText = fn.oklch(vars.layer, {
 });
 
 const layerContext = createContext();
+const edgeContext = createContext();
 
 utility(
   "layer",
@@ -1307,6 +1309,7 @@ utility(
   set(vars.layer, layer),
   set(vars.text, vars.layerScheme),
   set(vars.edge, edge),
+  set(vars.layerEdge, vars.edge),
   set(vars.layerBand, layerBand),
   set(vars.layerScheme, layerScheme),
   getBaseDeclarations(vars.layer),
@@ -1324,30 +1327,10 @@ utility(
     set(vars.layerPushDirectionToLight, 1),
     set(vars.edgePushDirection, 1),
   ),
-  layerContext(({ provide, inherit }) => {
-    const parentEdge = inherit(vars.layerEdgeContext, vars.edge);
-    const parentEdgeSource = inherit(vars.layerEdgeSourceContext, 0);
-    return [
-      set(provide(vars.layerParentContext), vars.layer),
-      set(
-        provide(vars.layerEdgeContext),
-        fn.colorMix(
-          "oklch",
-          parentEdge,
-          vars.edge,
-          fn.mul(inputs.frameBorderingContext, "100%"),
-        ),
-      ),
-      set(
-        provide(vars.layerEdgeSourceContext),
-        fn.add(
-          fn.mul(parentEdgeSource, fn.invert(inputs.frameBorderingContext)),
-          inputs.frameBorderingContext,
-        ),
-      ),
-      set(vars.layerParent, inherit(vars.layerParentContext)),
-    ];
-  }),
+  layerContext(({ provide, inherit }) => [
+    set(provide(vars.layerParentContext), vars.layer),
+    set(vars.layerParent, inherit(vars.layerParentContext)),
+  ]),
 );
 
 function getLayerOffsetDeclarations(arbitraryPattern = "[*]") {
@@ -1660,8 +1643,8 @@ utility("edge-raw", set(inputs.edgeA, 1), set(inputs.edgePushL, 0));
 
 utility(
   "edge-inherit",
-  layerContext.read(({ inherit }) => {
-    const parentEdge = inherit(vars.layerEdgeContext, vars.edge);
+  edgeContext.read(({ inherit }) => {
+    const parentEdge = inherit(vars.layerEdgeContext, vars.layer);
     const parentEdgeSource = inherit(vars.layerEdgeSourceContext, 0);
     return [
       set(
@@ -1675,16 +1658,13 @@ utility(
       ),
       set(
         inputs.edgePushL,
-        fn.add(
-          fn.invert(parentEdgeSource),
-          fn.mul(parentEdgeSource, fn.neg(vars.edgeContrastValue)),
-        ),
+        fn.sub(1, fn.mul(parentEdgeSource, fn.add1(vars.edgeContrastValue))),
       ),
       set(
         inputs.edgeA,
         fn.add(
-          fn.mul(fn.invert(parentEdgeSource), 0.1),
-          fn.mul(parentEdgeSource, fn.sub(alpha, vars.edgeContrastValue)),
+          0.1,
+          fn.mul(parentEdgeSource, fn.sub(alpha, vars.edgeContrastValue, 0.1)),
         ),
       ),
     ];
@@ -2087,6 +2067,13 @@ function getFrameBorderingContextDeclarations() {
   return set(inputs.frameBorderingContext, 1);
 }
 
+function getLayerEdgeContextDeclarations() {
+  return edgeContext(({ provide }) => [
+    set(provide(vars.layerEdgeContext), fn.important(vars.layerEdge)),
+    set(provide(vars.layerEdgeSourceContext), fn.important(1)),
+  ]);
+}
+
 const LAST_VISIBLE_SELECTOR = "&:not(:has(~ *:not([hidden],template)))";
 
 /**
@@ -2370,23 +2357,27 @@ utility(
   set(inputs.frameBorder, "1px"),
   set.borderWidth(inputs.frameBorder),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 utility(
   "frame-border-*",
   getFrameBorderWidthDeclarations(inputs.frameBorder),
   set.borderWidth(inputs.frameBorder),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 
 utility(
   "frame-ring",
   set(inputs.frameRing, "1px"),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 utility(
   "frame-ring-*",
   getFrameBorderWidthDeclarations(inputs.frameRing),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 
 function getFrameBorderingDarkLight() {
@@ -2422,6 +2413,7 @@ utility(
   set(inputs.frameBordering, "1px"),
   getFrameBorderingDarkLight(),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 
 utility(
@@ -2443,6 +2435,7 @@ utility(
   getFrameBorderWidthDeclarations(inputs.frameBordering),
   getFrameBorderingDarkLight(),
   getFrameBorderingContextDeclarations(),
+  getLayerEdgeContextDeclarations(),
 );
 
 export const input = [
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -634,6 +634,8 @@ const frameVars = {
   frameParentPaddingContext: _ak.var("fppc"),
   frameParentBorderContext: _ak.var("fpbc"),
   frameParentRingContext: _ak.var("fpgc"),
+  frameParentBorderingBorderContext: _ak.var("fpbbc"),
+  frameParentBorderingRingContext: _ak.var("fpbgc"),
   frameParentRowContext: _ak.var("fpwc"),
   frameParentCornerTLContext: _ak.var("fpctl"),
   frameParentCornerTRContext: _ak.var("fpctr"),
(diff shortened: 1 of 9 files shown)


packages/ariakit-tailwind/src/input.ts
```diff
@@ -728,6 +730,9 @@ const inputs = {
   frameBorder: _ak.prop.len("frame-border", { initial: "0px" }),
   frameRing: _ak.prop.len("frame-ring", { initial: "0px" }),
   frameBordering: _ak.prop.len("frame-bordering", { initial: "0px" }),
+  frameBorderingContext: _ak.prop.number("frame-bordering-context", {
```
diegohaz: Shouldn't all these `prop.number` calls with `{ initial: 0 }` be simplified to `prop.zero` instead?

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1320,11 +1325,22 @@ utility(
     set(vars.layerPushDirectionToLight, 1),
     set(vars.edgePushDirection, 1),
   ),
-  layerContext(({ provide, inherit }) => [
-    set(provide(vars.layerParentContext), vars.layer),
-    set(provide(vars.layerEdgeContext), vars.edge),
-    set(vars.layerParent, inherit(vars.layerParentContext)),
-  ]),
+  layerContext(({ provide, inherit }) => {
+    const parentEdge = inherit(vars.layerEdgeContext, vars.layer);
```
diegohaz: **Regression when no ancestor in the chain explicitly sets a frame border, ring, or bordering width.**

The new `layerContext` falls back to `vars.layer` here. When the topmost `ak-layer` element has no explicit border (so `--_ak-frame-bordering-context` is 0 on every layer up the chain), `parentEdge` resolves to the topmost layer's `--ak-layer` value (alpha 1, full-saturation color). The `color-mix(..., var(--ak-edge) 0%)` then provides that opaque layer color as `layerEdgeContext`, and descendants using `ak-edge-inherit` derive a fully-opaque edge from it instead of the parent's subtle alpha-0.1 derived edge.

Repro (only utilities already present in the sandbox source, so Tailwind compiles them):

```html
<div class="ak-layer ak-layer-blue-600">
  <div class="ak-layer ak-layer-20 ak-edge-inherit ak-frame-border">child</div>
</div>
```

Measured `getComputedStyle(child).getPropertyValue('--ak-edge')` against the running sandbox at `/react/previews/ariakit-tailwind/`:

| Version | child `--ak-edge` |
| --- | --- |
| `main` (pre-PR) | `oklch(1 0.245 262.881 / 0.1)` (subtle, alpha 0.1) |
| This PR | `oklch(0.1634 0.0091 264.28)` (fully opaque dark color) |

The pre-PR behavior treated every `ak-layer` as an implicit edge anchor via `set(provide(vars.layerEdgeContext), vars.edge)`, so this case was naturally covered. The new design intentionally drops that — which is fine when at least one explicit border exists somewhere up the chain (the new "skip intermediates" behavior), but it makes `ak-edge-inherit` fall off a cliff when nothing in the chain is explicit.

The ergonomic fix path is to make the layerContext fallback be `vars.edge` instead of `vars.layer`:

```suggestion
    const parentEdge = inherit(vars.layerEdgeContext, vars.edge);
```

With that change the topmost layer effectively becomes an implicit anchor (`colorMix(self-edge, self-edge, 0%) = self-edge`) while the skip behavior between explicit ancestors is preserved. I verified locally that this restores the subtle alpha-0.1 edge for the repro above and does not regress the existing `skip` / `inherit skip` / `zero` sandbox snapshots conceptually, but the resulting color is the child's own derived edge rather than R1's exact edge value (the chained lazy `var()` resolution still re-evaluates at the child). If exact pre-PR semantics matter, the maintainer may need a deeper rework — but at minimum the current full-opacity outcome is surprising and worth fixing.

The existing sandbox cases all live inside a top-level `<Layer>` that includes `ak-frame-border` (line 424 in `app/src/sandbox/ariakit-tailwind/index.react.tsx`), so the snapshots never exercise the "no explicit anchor anywhere" scenario. Adding a case where the outermost wrapper has no border would have caught this.

diegohaz (commented):
## Review summary

Initial review of `@ariakit/tailwind` inherit-context fix.

**Scope.** Diff against `main` (merge base `7488af8`, head `e609ed5`): 9 files / +256 / −15. Full review of `packages/ariakit-tailwind/src/input.ts`, `packages/ariakit-tailwind/src/output.css`, `packages/ariakit-tailwind/readme.md`, `app/src/sandbox/ariakit-tailwind/index.react.tsx`, all four snapshot files under `app/src/sandbox/ariakit-tailwind/__snapshots__/`, and `.changeset/300-tailwind-inherit-context.md`. Cross-checked the new `frameBorderingContext` flag against every utility that should set or read it (`frame`, `frame-border`, `frame-border-*`, `frame-ring`, `frame-ring-*`, `frame-bordering`, `frame-bordering-*`, `frame-bordering-inherit`, layer context provision, `edge-inherit`).

**Validation performed.**

- `pnpm tsc` — passes.
- `pnpm -F @ariakit/tailwind build` — `output.css` regenerated from `input.ts` matches the committed file byte-for-byte.
- Ran the live sandbox at `http://localhost:4322/react/previews/ariakit-tailwind/` and queried `getComputedStyle` for both the pre-PR (`main`) `output.css` and the PR `output.css`, side-by-side. Used only utility class names that already appear in the sandbox source so Tailwind 4 had them compiled.
- Sanity-checked a candidate fix locally (`inherit(vars.layerEdgeContext, vars.edge)`) by editing `input.ts`, rebuilding the package, and rechecking computed styles.

**Findings.**

1. **Inline comment on `packages/ariakit-tailwind/src/input.ts:1329`** — regression: when no ancestor in the chain explicitly sets `frame-border` / `frame-ring` / `frame-bordering*`, `ak-edge-inherit` on a descendant produces a fully-opaque edge in the topmost layer's color instead of the pre-PR alpha-0.1 inherited edge. Root cause is the fallback `inherit(vars.layerEdgeContext, vars.layer)` — `vars.layer` has alpha 1, and that propagates through the new `color-mix(... 0%)` pipeline. The same comment notes a suggested fix and the missing sandbox case.
2. **`ak-frame-bordering-inherit` handles the same "no anchor anywhere" case correctly** — its fallback is `"0px"`, so descendants resolve to a 0-width bordering (which is the right floor for a width). Only `ak-edge-inherit` has the regression because its fallback is a non-edge color rather than a transparent/no-edge value.
3. **Performance** — adds a `color-mix(in oklch, ...)` to every `ak-layer` context provision. The perf bot reports +4% style recalc on page load against the heavily-nested sandbox (and −6% on the toggle scenario, so net is mixed). Not a blocker, but worth being conscious of as this is on a hot path for any deeply-nested layered UI. I didn't find a cheap drop-in replacement for `color-mix` here that preserves the parity-based provision pattern.
4. **Design choice worth confirming** — only explicit width utilities are anchors. `ak-edge-color-*`, `ak-edge-l-*`, `ak-edge-50`, etc. set color/alpha but do not become inheritance anchors. The README change documents this clearly. Reasonable, but a user setting `ak-edge-red-50` on a borderless intermediate may be surprised that descendants skip past it. Not flagging as a finding since the README is explicit.

**Areas I did not fully validate.**

- I did not run the full browser-test suite (`pnpm -F app test`) because the dev server's two background ports collided with another running instance during this session; I only ran the live sandbox via the Astro dev server on `:4322`. The author's PR description records that `test-chrome ariakit-tailwind` and `test-perf` were run and re-snapshotted, so the regression in finding 1 is presumably not currently asserted by any snapshot (which matches my read of the sandbox: every test wrapper carries `ak-frame-border` on the outermost `<Layer>`).
- I did not benchmark the proposed fallback change against the existing snapshots — only checked it didn't break the `skip` / `inherit skip` / `zero` cases conceptually.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in f1e563dd0 - replaced zero-valued `prop.number(..., { initial: 0 })` declarations with `prop.zero(...)` while preserving inherited `frameRow` behavior.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in f1e563dd0 - added an explicit edge-source context so `ak-edge-inherit` only consumes edge context from ancestors with explicit frame border/ring/bordering widths, while no-source and edge-only ancestors fall back to the current edge. Added sandbox snapshot coverage for both cases.

diegohaz:
**Fixed**

- f1e563dd0 (https://github.com/ariakit/ariakit/pull/6041#discussion_r3295245781): Replaced zero-valued custom properties with `prop.zero`.
- f1e563dd0 (https://github.com/ariakit/ariakit/pull/6041#discussion_r3295245811): Added explicit edge-source tracking so `ak-edge-inherit` skips no-source and edge-only ancestors, plus sandbox coverage.

packages/ariakit-tailwind/src/input.ts
diegohaz: Regression resolved in f1e563dd0. Verified against the running sandbox: descendants now get the parent layer's alpha-0.1 hue/chroma when no ancestor has an explicit border, and `oklch(1 0.245 27.325 / 0.4)` (the OUTER's RED) when an explicit anchor is present further up but an intermediate has no border (correct skip behavior). Behavior in this PR head now matches the pre-PR result for the no-anchor case and the documented skip semantics for the with-anchor case. The added sandbox cases at the top of the file (`edge inherit root no source`, `edge inherit no source`, `edge inherit edge no source`) cover the regression scenario in snapshot form. Marking this thread resolved.

diegohaz (commented):
## Re-review (incremental)

**Scope.** Diff between `e609ed5` (previously reviewed) and `f1e563dd0`. Only review thread from this skill's prior run was the regression on `input.ts:1329`; the parallel `prop.zero` simplification thread came from the author. Both are addressed here, plus the new sandbox/snapshot coverage. No other concurrent reviewer threads to consider.

**Validation performed.**

- `pnpm tsc` — passes.
- `pnpm -F @ariakit/tailwind build` — `output.css` regenerated from `input.ts` byte-for-byte matches the committed file.
- Live sandbox at `http://localhost:4322/react/previews/ariakit-tailwind/`: re-ran the original regression repro (`ak-layer ak-layer-blue-600 > ak-layer ak-layer-20 ak-edge-inherit ak-frame-border`) — now produces `oklch(1 0.245 262.881 / 0.1)` (alpha 0.1, blue) on both root and child, matching the pre-PR result. Also ran a skip case (OUTER with `ak-frame-border ak-edge-red-600 ak-edge-40` → middle with no border → child with `ak-edge-inherit ak-frame-border`) — child correctly gets `oklch(1 0.245 27.325 / 0.4)` (RED from OUTER, hue 27.325). And the edge-only case (`ak-edge-red-600 ak-edge-40` without `ak-frame-border` on the parent) correctly does NOT propagate red to the child.

**On the complexity question.**

I looked specifically at whether the new `layerEdgeSourceContext` could be removed or folded into the existing `layerEdgeContext`. My read: **the two pieces of state are independent and both required**.

- The inherited *edge color* (`layerEdgeContext`) and the *"has any explicit anchor up the chain"* boolean (`layerEdgeSourceContext`) genuinely encode two different things. You can't fold the boolean into the color value without ambiguity: `ak-edge-0` legitimately produces an alpha-0 edge, so "alpha == 0" can't be reused as a "no source" sentinel. Same problem with any other color sentinel (`transparent`, `currentcolor`).
- The conditional logic in `edge-inherit` (3 blends — color, push, alpha) is the standard `lerp(default, withSource, source)` pattern. It can't be replaced by container style queries because `--_ak-frame-bordering-context` is `inherits: false` and `@container style(...)` only sees ancestors, so a same-element conditional needs the calc-based gate.
- The propagation `parentEdgeSource * (1 - context) + context` is just OR for 0/1 values; could be written as `fn.max(parentEdgeSource, inputs.frameBorderingContext)` for slightly more conceptual clarity, but the lerp form is consistent with how the file expresses every other "select-between-two-values" pattern (e.g., the same shape is used for `frameParentBorderingBorderContext` propagation at line ~2262). Keeping it consistent feels right.
- The magic constants `0.1` and `1` in the `edgeA` / `edgePushL` blends are the property defaults from `inputs.edgeA` / `inputs.edgePushL`. Conceptually duplicating them is the only DRY nit I see, but the lib API doesn't expose `.initial` for symbolic reference, so wiring this up would cost more than it saves.

**Net: I don't think the logic is over-engineered.** The added surface is ~12 lines of declarative state + 1 new context var. The semantics it gives you (skip-through-non-anchor, fall-back-cleanly-when-no-anchor, distinguish-edge-color-only-from-explicit-anchor) wouldn't compress further without either changing the design goals or restructuring the parity-based context machinery.

**One forward-looking observation, not a blocker.** Inspecting the live values in DevTools, `--_ak-lesc-{parity}` accumulates nested calc per layer level — e.g. four layers deep already produces `calc((calc((calc((0 * (1 - 0)) + 0) * (1 - 1)) + 1) * (1 - 0)) + 0)`. Same pattern already exists for `--_ak-lec-{parity}` (color-mix nesting), `--_ak-fpbbc-{parity}`, etc. — it's a consequence of the parity vars being created via `_ak.var(...)` (unregistered) rather than `@property`-typed inside `getParityVar` in `lib.ts:282`. The PR just inherits that property. Worth flagging because it's part of why the perf bot reports +4% style recalc on page load: every additional unregistered chain compounds. Cleaning this up would mean teaching `getParityVar` to register the parity-suffixed vars when the source property is typed — out of scope for this PR, but plausibly the lever to pull if the layered-rendering hot path needs more headroom later.

**No blocking findings on this commit.** The regression I flagged is fixed and the new sandbox snapshots assert it.

PR #6044, opened 2026-05-25
Update React test matrix

diegohaz:
## Motivation

React 19 is now the default development/test stack, and CI should no longer spend coverage on React 17. We still keep public React 17 peer ranges for Ariakit packages for now, but the automated React matrix should cover React 18 and React 19.

The default local `pnpm test` path runs the normal React 19 workspace graph. Use `pnpm test-react-18` or `pnpm react18 <script>` to run commands in an isolated React 18 workspace without keeping a separate React-version test package in the repository.

## Solution

- Bump the root/app/examples React development dependencies and React types to React 19, including the root Next dev dependency needed for deployment detection.
- Add `ariakit react18` in `@ariakit/scripts`. The command copies the current checkout to a temp workspace, excludes build artifacts and local env/secret files, rewrites React/React DOM/types to the React 18 versions declared by the website workspace, keeps the temp React 18 lockfile on warm runs, runs a temp `pnpm install`, and then executes either a root script or arbitrary command from that temp workspace.
- Wire `test-react-18` and the CI React 18 shard job through `pnpm react18 test`, while keeping `pnpm test` as the normal React 19 path.
- Keep React version handling inside the `ariakit react18` temp workspace runner. Vitest and Astro use their normal resolver configuration; the temporary React alias helper was removed after confirming the dev invalid-hook-call warnings are pre-existing Astro/Vite cold-optimizer behavior.
- Remove the dedicated `test/react-19` peer-resolution workspace and simplify pnpm React 19 peer allowances to `react`/`react-dom` allowed versions.
- Remove React 17 / old React Testing Library-specific test workarounds, including the old input selection fallback in `@ariakit/test`, and fix the narrower React 19 type changes surfaced by TypeScript.

A changeset is included for the `@ariakit/test` peer dependency change. Public React 17 peer ranges for Ariakit packages remain unchanged.

## Verification

- `pnpm install --lockfile-only --ignore-scripts`
- `pnpm lint-fix`
- `pnpm lint`
- `pnpm tsc`
- `pnpm exec tsc -p packages/ariakit-scripts/tsconfig.node.json --noEmit`
- `pnpm -F nextjs run build-worker`
- `pnpm -F app exec astro check`
- `pnpm test --run examples`
- `pnpm react18 test --run examples`
- `pnpm test --run examples/disclosure-content-animating/test.ts`
- `pnpm test --run examples/combobox-radix/test.ts examples/combobox-radix-select/test.ts examples/tooltip-framer-motion/test.ts examples/disclosure-content-animating/test.ts`
- `pnpm test --run examples/combobox-radix/test.ts examples/combobox-radix-select/test.ts examples/tooltip-framer-motion/test.ts examples/dialog-react-router/test.ts examples/hovercard-shadow-dom/test.ts examples/dialog-react-toastify/test.ts`
- `pnpm test-react-18 --run examples/disclosure-content-animating/test.ts`
- `pnpm test-react-18 --run examples/combobox-radix/test.ts examples/combobox-radix-select/test.ts examples/tooltip-framer-motion/test.ts examples/disclosure-content-animating/test.ts`
- `pnpm test-react-18 --run examples/combobox-radix/test.ts examples/combobox-radix-select/test.ts examples/tooltip-framer-motion/test.ts examples/dialog-react-router/test.ts examples/hovercard-shadow-dom/test.ts examples/dialog-react-toastify/test.ts`
- `pnpm react18 -- node -e "console.log(require('react/package.json').version); console.log(require('react-dom/package.json').version); console.log(require('@types/react/package.json').version);"`
- `pnpm react18 -- pnpm -F app exec node -e "console.log(require('react/package.json').version); console.log(require('react-dom/package.json').version);"`
- `pnpm react18 test-react-18 --help | sed -n '1,14p'`
- `pnpm react18 dev-app --no-clean` smoke-tested startup of Astro and Next.js, then stopped
- `pnpm test-solid --run examples/separator/test.ts`
- `pnpm peers check` fails only on existing `tailwindcss-animate` and examples `vitest`/coverage peer warnings.


examples/package.json
```diff
@@ -23,13 +23,17 @@
     "react-shadow": "20.6.0",
     "react-toastify": "11.1.0",
     "textarea-caret": "3.1.0",
-    "tiny-invariant": "1.3.3"
+    "tiny-invariant": "1.3.3",
+    "use-sync-external-store": "1.6.0"
```
diegohaz: Why do we need this dependency?

packages/ariakit-react-components/src/form/form.tsx
```diff
@@ -104,8 +104,9 @@ export const useForm = createHook<TagName, FormOptions>(function useForm({
   }, [autoFocusOnSubmit, submitFailed, items]);
 
   const onSubmitProp = props.onSubmit;
+  type SubmitEvent = Parameters<NonNullable<typeof onSubmitProp>>[0];
 
-  const onSubmit = useEvent((event: FormEvent<HTMLType>) => {
+  const onSubmit = useEvent((event: SubmitEvent) => {
```
diegohaz, line 110: Why do we need this change?

packages/ariakit-test/src/type.ts
```diff
@@ -117,11 +117,6 @@ export function type(
               inputType,
               ...options,
             });
-            // Need to re-assign the selection state for React 17 and/or React
-            // Testing Library 12 (not sure which).
-            if (input.selectionStart !== nextCaretPosition) {
-              input.setSelectionRange(nextCaretPosition, nextCaretPosition);
-            }
```
diegohaz, line 124: This requires a changeset, doesn't it?

package.json
```diff
@@ -99,7 +101,7 @@
     "husky": "9.1.7",
     "jsdom": "29.1.1",
     "lint-staged": "17.0.5",
-    "next": "14.2.35",
+    "next": "16.2.6",
```
diegohaz, line 104: Do we need to update this dependency? Do we even need it in the root package.json? Maybe Vercel requires it, we need to check. If we need it there and it's okay for it to diverge from the website version, update the Renovate config to allow updates to the next version for the root package. If Vercel doesn't need it in the root package, we can remove it.

.github/workflows/main.yml
```diff
@@ -70,7 +71,7 @@ jobs:
         uses: ./.github/workflows/setup
 
       - name: Test
-        run: pnpm test --shard ${{ matrix.shard }}
+        run: pnpm run test-react-${{ matrix.react }} --shard ${{ matrix.shard }}
```
diegohaz: I don't think we need this `run` in pnpm.

test/react-19/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "ariakit-test-react-19",
```
diegohaz: Why do we need this package?

pnpm-workspace.yaml
```diff
@@ -3,6 +3,7 @@ packages:
   - templates/*
   - nextjs
   - examples
+  - test/*
```
diegohaz: Why?

vitest.config.ts
```diff
@@ -32,6 +58,31 @@ const LOADER = (process.env.ARIAKIT_TEST_LOADER ??
...
+  "@radix-ui/react-select",
+  "@testing-library/react",
+  "motion",
+  "react",
+  "react-dom",
+  "react-router",
+  "react-shadow",
+  "react-toastify",
+  "use-sync-external-store",
+];
```
diegohaz: Why do we need this?

vitest.setup.ts
```diff
@@ -56,6 +55,7 @@ async function loadReact(dir: string) {
     `./${dir}/index.react.tsx`,
   );
   if (failedImport) return false;
+  const { render: renderReact } = await import("@ariakit/test/react");
```
diegohaz: Why can't we import it statically at the top-level?

.github/workflows/main.yml
```diff
@@ -55,11 +55,12 @@ jobs:
         run: pnpm run build
 
   test:
-    name: Test (${{ matrix.shard }})
+    name: Test React ${{ matrix.react }} (${{ matrix.shard }})
```
diegohaz: Let's split this into two tests: Normal `Test (${{ matrix.shard }})` and an additional `Test React 18`

examples/package.json
diegohaz: `vitest.config.ts`'s new `aliasPkg(pkg, resolver)` uses `createRequire(new URL("./examples/package.json", …))` to locate every React-peered package from the React 18 graph. With pnpm strict mode, only deps declared in a workspace's `package.json` are resolvable through that workspace's `createRequire`, so each entry in `reactPeerPackages` has to appear in both `examples/package.json` (React 18 graph) and `test/react-19/package.json` (React 19 graph).

`use-sync-external-store` is on that list because it's a peer of `@ariakit/react-store` and reached via `useStore`, so vitest needs to pin it to the chosen React graph. Same reason the `@testing-library/react`/`@types/react`/`@types/react-dom` devDeps appear in this file. Verified locally: both `pnpm test-react-18 --shard 1/2` and `pnpm test-react-19 --shard 1/2` pass on this branch (76 files, 457 tests each).

packages/ariakit-react-components/src/form/form.tsx
diegohaz: It's needed because React 19's `@types/react` typed the form `onSubmit` handler as `SubmitEvent<HTMLFormElement>` (a new React-defined generic, distinct from the legacy `FormEvent<HTMLFormElement>` still used in React 18). I verified by reverting just this hunk to `const onSubmit = useEvent((event: FormEvent<HTMLType>) => { … })` and running `pnpm tsc`, which fails with:

```
packages/ariakit-react-components/src/form/form.tsx(109,20): error TS2345: Argument of type 'FormEvent<HTMLFormElement>' is not assignable to parameter of type 'SubmitEvent<HTMLFormElement>'.
  Types of property 'target' are incompatible.
    Type 'EventTarget' is not assignable to type 'EventTarget & HTMLFormElement'.
```

Deriving the event type via `Parameters<NonNullable<typeof onSubmitProp>>[0]` keeps the code working under both React 18 (`FormEvent`) and React 19 (`SubmitEvent`) typings without having to import a version-specific name. One tiny nit: the local `type SubmitEvent` shadows the global DOM `SubmitEvent` inside this function — harmless here, but a one-line `// React 19 types this as SubmitEvent<…>, React 18 as FormEvent<…>` comment would save a future reader from doing the same revert experiment.

packages/ariakit-test/src/type.ts
diegohaz: Yes — `@ariakit/test`'s peer ranges in `packages/ariakit-test/package.json` still allow `react: ^17.0.0` and `@testing-library/react: ^12.0.0 || ^13.0.0`, so anyone running `@ariakit/test`'s `type()` helper on that pre-existing combination loses the selection re-assignment silently. That's the kind of behavior change a consumer should be able to see in the changelog.

Three options, ordered by how much they change the contract:

1. **Add a patch changeset for `@ariakit/test`** — recommended, fits the PR's scope. Something like:
   ```md
   ---
   "@ariakit/test": patch
   ---

   Drop the React 17 / React Testing Library 12 input selection workaround in `type()`. The `dispatch.input` call now sets selection state on its own under modern React + RTL.
   ```
2. Bump the peer ranges to drop React 17 / RTL ≤13 (major). PR description already commits to keeping those peers, so this is bigger than this PR.
3. Restore the workaround until peers are dropped. Cheapest, but defeats the cleanup motivation.

The same workaround note in the PR description mentions “Remove React 17 / old React Testing Library-specific test workarounds, including the old input selection fallback in `@ariakit/test`” — so the intent is clearly (1), the changeset is the missing piece.

examples/menu-wordpress-modal/menu.tsx
```diff
@@ -114,14 +114,14 @@ export function createMenuSlot(name: string, bubblesVirtually = false) {
...
     return (
       <SlotFill.Fill>
         {(menu) => (
           // Re-create the menu context within the fill tree
           <MenuContext.Provider value={menu as Ariakit.MenuStore}>
-            {props.children}
+            {props.children as React.ReactNode}
           </MenuContext.Provider>
         )}
       </SlotFill.Fill>
```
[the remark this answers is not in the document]
diegohaz: Confirmed in the current branch — the widened `React.ComponentProps<typeof SlotFill.Fill>` type now accepts every `SlotFill.Fill` prop (e.g. `name`), but the `bubblesVirtually` branch only forwards `children`. Everything else is silently dropped, which is exactly the trap the narrower `{ children: React.ReactNode }` type used to prevent. The `props.children as React.ReactNode` cast on line 124 also loses the function-children overload of `SlotFill.Fill`.

This is example-only, so the blast radius is small, but the typing widening did regress a real invariant. Suggested shape that preserves the spread + still handles render-prop children:

```tsx
const Fill = ({ children, ...rest }: React.ComponentProps<typeof SlotFill.Fill>) => {
  if (!bubblesVirtually) {
    return <SlotFill.Fill {...rest}>{children}</SlotFill.Fill>;
  }
  return (
    <SlotFill.Fill {...rest}>
      {(menu) => (
        <MenuContext.Provider value={menu as Ariakit.MenuStore}>
          {typeof children === "function" ? children(menu) : children}
        </MenuContext.Provider>
      )}
    </SlotFill.Fill>
  );
};
```

If this PR wants to keep the diff minimal, narrowing `Fill` back to `{ children: React.ReactNode }` and using `{...props}` only in the non-virtual branch (with `children` passed explicitly) is the smaller alternative.

package.json
diegohaz: The bump itself is consistent with the PR's stated goal of aligning the root with the `nextjs` workspace (which is on Next 16 + React 19). But — separately — the root `next` is dev-only and isn't imported anywhere outside the `website/` and `nextjs/` workspaces (`grep -rn 'from "next' --exclude-dir={node_modules,website,nextjs,dist}` returns nothing in the repo). So in principle the dep could be dropped from root entirely; that's a pre-existing cleanup, not something to block this PR on.

If you keep it, the 14 → 16 bump is fine. If you'd rather not own it here, leaving it at 14 also works because nothing at the root level reads it.

.github/workflows/main.yml
diegohaz: Agreed. `pnpm` will run any custom script as a shortcut as long as it doesn't shadow a built-in subcommand, and `test-react-18` / `test-react-19` don't. I verified with `pnpm test-react-18 --shard 1/2` in this worktree — same behavior as the form with `run` (76 files, 457 tests, both pass).

It's also more consistent with the pre-existing line on this file (`pnpm test --shard ${{ matrix.shard }}`, no `run`):

```diff
-        run: pnpm run test-react-${{ matrix.react }} --shard ${{ matrix.shard }}
+        run: pnpm test-react-${{ matrix.react }} --shard ${{ matrix.shard }}
```

test/react-19/package.json
diegohaz: It's the React 19 anchor for the dual `createRequire` setup in `vitest.config.ts`:

```ts
const requireReact18 = createRequire(new URL("./examples/package.json", import.meta.url));
const requireReact19 = createRequire(new URL("./test/react-19/package.json", import.meta.url));
```

Because the `examples` workspace deliberately keeps React 18 as its runtime graph (per PR description), there's no existing workspace pnpm would install the React 19 + matching peer set into. Creating `test/react-19` as a deps-only private workspace:

- Gives pnpm a place to materialize the React 19 dependency graph (`react@19`, `@testing-library/react`, `motion`, `react-router`, `react-shadow`, `react-toastify`, `@radix-ui/*`, etc.).
- Gives `requireReact19` a stable anchor file (`test/react-19/package.json`) for the alias resolution.
- Is excluded from the changeset bot via `private: true` + workspace ignore — confirmed no version churn.

Minor nit: the workspace name `ariakit-test-react-19` is a bit long; something like `@ariakit-tests/react-19` (scoped, private) would be more conventional, but neither is wrong.

pnpm-workspace.yaml
diegohaz: Required so pnpm picks up the new `test/react-19` workspace (and any future sibling like `test/react-20`). Without `test/*` here, `pnpm install` skips the directory entirely, no `node_modules` is materialized, and `requireReact19.resolve("react/package.json")` in `vitest.config.ts` fails to find the React 19 graph — `ARIAKIT_TEST_REACT=19 pnpm test` would throw `Could not resolve package: react`.

Verified by running `pnpm install` then `pnpm test-react-19 --shard 1/2` and `--shard 2/2` on this branch; both pass (151 files, 805 tests total).

vitest.config.ts
diegohaz: The previous config only aliased `react` and `react-dom`. The expanded list exists because each React-peered package in this monorepo can be installed in two variants — one compiled against React 18 (under `examples/node_modules`) and one against React 19 (under `test/react-19/node_modules`). Without an alias, vitest picks whichever copy pnpm hoisted nearest the importing file, which can mix a React-18-peered `@testing-library/react` or `motion` with a React-19-aliased `react`, producing "multiple React instances" / invalid-hook-call errors at runtime.

The pinning is exhaustive for the packages actually traversed by Vitest examples (no `@wordpress/components` because `menu-wordpress-modal` is browser-only via `test-chrome.ts`, no `react-aria-components` because it's only used in `app/` and not in the root Vitest graph).

One maintenance hazard worth flagging: the array has to stay in sync with three other places — `examples/package.json`, `test/react-19/package.json`, and the comment above. A future contributor adding a 10th React-peered package to examples is silently fine until tests start crashing for one of the React versions. Optionally, a runtime guard could read both `package.json`s and assert each name is present — or just lean on PR review and accept the trade-off.

vitest.setup.ts
diegohaz: A top-level static import of `@ariakit/test/react` pulls in `react`, `react-dom`, and `@testing-library/react` (RTL) eagerly whenever `vitest.setup.ts` is evaluated — including the `ARIAKIT_TEST_LOADER=solid` path, where:

- `loadReact` is never called (the setup picks `LOADERS[LOADER]` = `loadSolid`).
- The React-peer alias list still rewrites `react` / `react-dom` / `@testing-library/react` to the chosen React graph (React 19 by default), which is wasted work for a Solid test run.
- Static `import` happens at module-evaluation time, *before* the test files run, so any side effects in RTL's setup (e.g. mounting JSDOM observers) also fire on Solid runs.

Deferring it inside `loadReact` means: React is only loaded for React tests, the alias machinery has already been wired up by the time the dynamic resolver runs (so there's no "first import was unaliased" risk if a future change pre-imports it), and Solid runs stay React-free.

Worth a one-line comment in the file explaining the deferral so it doesn't get "helpfully" hoisted back to the top later.

diegohaz (commented):
### Scope

Initial review of #6044 at head `a7d9f9fd`, baseline `c634c002f` (`main`). All 23 changed files (752 / 340 LOC) read in full, plus surrounding context for the React-peer alias model, the form/test packages, and the impacted examples. Spawned no subagents.

### Validation performed

- `pnpm install` and `pnpm tsc` in a fresh worktree at the PR head — clean.
- Reverted just `form.tsx`'s `event: SubmitEvent` → `event: FormEvent<HTMLType>` and re-ran `pnpm tsc` to confirm the React 19 typings actually require the indirection (TS2345, `FormEvent` not assignable to `SubmitEvent<HTMLFormElement>`). Restored before continuing.
- `pnpm test-react-18 --shard 1/2` and `--shard 2/2` — 151 files, 805 tests pass.
- `pnpm test-react-19 --shard 1/2` and `--shard 2/2` — 151 files, 805 tests pass.
- `pnpm test-react-18 --shard 1/2` invoked without the `run` keyword to confirm pnpm's script shortcut works for the new scripts (referenced in the `.github/workflows/main.yml` thread).

### Findings

All actionable feedback is on existing review threads (replies, marked with the hidden review id above). Quick punch list:

- **Changeset for `@ariakit/test`** — the removed React 17 / RTL ≤13 selection-state workaround in `packages/ariakit-test/src/type.ts` is a behavior change for users still on those peer combinations, and the peer ranges aren't being narrowed in this PR. Patch changeset recommended. ([thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297521631))
- **`pnpm run` is redundant in the CI matrix step** — `pnpm test-react-${{ matrix.react }} --shard ${{ matrix.shard }}` works identically and matches the pre-existing style on the same file. ([thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297523952))
- **`Fill` in `menu-wordpress-modal/menu.tsx`** — widening to `React.ComponentProps<typeof SlotFill.Fill>` accepts every `SlotFill.Fill` prop, but the `bubblesVirtually` branch only forwards `children` and the `props.children as React.ReactNode` cast discards the render-prop overload. Example-only, but the typing now overstates what the component actually does. ([thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297523355))
- **Doc-nit on `form.tsx` `SubmitEvent`** — a one-line `// React 19 typed onSubmit as SubmitEvent<…>, React 18 as FormEvent<…>` would save the next reader from doing the same revert experiment. The derivation itself (`Parameters<NonNullable<typeof onSubmitProp>>[0]`) is the right shape. ([thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297520728))
- **Same doc-nit on `vitest.setup.ts`** — a one-liner explaining the deferred `@ariakit/test/react` import prevents a future contributor from hoisting it back to the top. ([thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297527373))

The remaining self-review questions on this PR (use-sync-external-store, root `next`, `test/react-19` package, `test/*` in pnpm-workspace, the expanded `reactPeerPackages` array) are answered in their respective threads — short version: each is load-bearing for the dual-React graph alias resolution under pnpm strict mode, and removing any one of them breaks the React 18 or React 19 path. Verified by running both matrices end-to-end.

### Things checked but not flagged

- The collapsible `inert: collapsed || undefined`, `useRef<T | undefined>(undefined)`, `useState<Ariakit.MenuProps["initialFocus"]>()`, and the new `isTableCellProps` type guard are all the right shape for React 19's stricter `inert`, no-arg-`useRef`, non-nullable `RefObject.current`, and stricter element narrowing.
- The `examples/composite/test.ts` simplification (dropping the inline `IS_REACT_ACT_ENVIRONMENT` shim in favor of `dispatch.keyDown` + `sleep(0)`) preserves coverage; both shards pass on both React versions.
- `packages/ariakit-test/src/react.tsx` is a refactor only (pulling `strictMode`/`wrapper` out of `options` so the rest can be passed to RTL cleanly, dropping the `@ts-ignore`). No behavior change, no changeset needed.
- CI matrix doubling (2 shards × 2 React versions = 4 jobs) is the intended cost of the new coverage — flagging only for visibility, not as a concern.
- `examples` runs against the React 18 runtime but compiles against React 19 types. This is a deliberate matrix choice and the dual-React Vitest run catches any React-19-only API actually used at React 18 runtime, so it's covered.

### Not validated

- `pnpm build`, `pnpm build-app-lite`, `pnpm test-solid`, and the visual / playwright suites — PR description reports these clean, and nothing in the diff suggests otherwise, but I didn't re-run them here.
- The `pnpm peers check` output the PR description mentions — out of scope for this review.

Comment-only, no approval / request-changes.

vitest.config.ts
diegohaz: Can't we simply get this list from the package.json files directly?

test/react-19/package.json
diegohaz: Why does the examples workspace keep React 18? Shouldn't it be updated to React 19 as well? I wonder if this whole setup could be simplified.

packages/ariakit-test/src/type.ts
diegohaz: Let's remove React 17 support for ariakit-test library.

examples/package.json
diegohaz: This setup is too complicated.

vitest.config.ts
diegohaz: Follow-up after prototyping two simplifications to this array:

**1. Drop `react-shadow` and `react-toastify`** — both packages are consumed only by Playwright tests, not Vitest. Verified by grep:

```
examples/hovercard-shadow-dom/   → test-browser.ts (Playwright only)
examples/dialog-react-toastify/  → test-chrome-firefox.ts (Playwright only)
```

No `test.ts` / `test.react.ts` files reference either package. The Vitest matrix never imports them, so the alias entry isn't doing anything. Trimmed the array down to the other 8 entries, re-ran `pnpm test-react-18 --shard 1/2` and `pnpm test-react-19 --shard 1/2` — both pass (76 files / 457 tests each). Every other entry I tried to remove produces a real failure (invalid hook call from `@radix-ui/react-context`, `useSyncExternalStore` resolution mismatch, RTL pulling in the wrong `react-dom/client`, etc.), so 8 is the actual floor.

**2. Derive the array from `test/react-19/package.json`** — by construction this workspace's `dependencies` is the canonical list of React-peered packages we want to pin. Reading them at config load drops the maintenance hazard you flagged above:

```ts
const react19Manifest = JSON.parse(
  readFileSync(new URL("./test/react-19/package.json", import.meta.url), "utf8"),
) as { dependencies?: Record<string, string> };
const reactPeerPackages = Object.keys(react19Manifest.dependencies ?? {})
  .filter((name) => !name.startsWith("@types/") && !PLAYWRIGHT_ONLY.has(name))
  .sort();
```

The `PLAYWRIGHT_ONLY` set captures the `react-shadow`/`react-toastify` exclusion explicitly so a future contributor adding (or removing) a peer in `test/react-19/package.json` and `examples/package.json` doesn't have to also remember to update `vitest.config.ts`. Both prototypes pass on both React versions across both shards in my worktree.

The `reactPeerPackages` array itself is then ~3 lines and the three-place sync hazard collapses into one.

package.json
```diff
@@ -15,6 +15,8 @@
     "preview-app-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run preview-app",
     "generate-images": "pnpm run playwright -u -c generate-images.config.ts",
     "test": "vitest",
+    "test-react-18": "cross-env ARIAKIT_TEST_REACT=18 vitest",
+    "test-react-19": "cross-env ARIAKIT_TEST_REACT=19 vitest",
```
diegohaz: **Alternative worth weighing: Vitest's `projects` feature can collapse these two scripts + the `ARIAKIT_TEST_REACT` env var into a single config.** Vitest 4 supports running two named projects in one invocation, each with its own `resolve.alias` and `dedupe`. Prototyped against this branch:

```ts
// vitest.config.ts (sketch — full file with shared sharedTest extracted)
export default defineConfig({
  plugins: PLUGINS_BY_LOADER[LOADER],
  test: {
    coverage: { include: ["packages"] },
    projects: [
      { extends: true,
        resolve: { alias: reactPeerPackages.map((p) => aliasPkg(p, requireReact18)), dedupe: ["react", "react-dom"] },
        test: { ...sharedTest, name: { label: "react-18", color: "cyan" } } },
      { extends: true,
        resolve: { alias: reactPeerPackages.map((p) => aliasPkg(p, requireReact19)), dedupe: ["react", "react-dom"] },
        test: { ...sharedTest, name: { label: "react-19", color: "green" } } },
    ],
  },
});
```

Then:
- `package.json` keeps a single `"test": "vitest"` — the `test-react-18` / `test-react-19` script pair go away.
- CI uses either `pnpm test --shard ${{ matrix.shard }}` (no react axis, single shard runs both Reacts) or `pnpm test --project react-${{ matrix.react }} --shard ${{ matrix.shard }}` (keeps the 4-job matrix). Both work.
- `vitest.config.ts` loses the `ALLOWED_REACT_VERSIONS` / `REACT_VERSION` / `reactResolver` selection block (~12 lines).

**What does NOT go away**: the `test/react-19` workspace, the `pnpm-workspace.yaml` add, the `requireReact18` / `requireReact19` anchors, and the `aliasPkg` / `resolvePkg` helpers — pnpm hoisting still requires two distinct dep graphs to coexist on disk. The structural complexity is the same; you trade `process.env.ARIAKIT_TEST_REACT` + two scripts for two projects-config blocks.

**Verified locally**: `pnpm exec vitest run --project react-18 --shard 1/2` and `--project react-19 --shard 1/2` both pass (76 files / 457 tests each, comparable durations). `pnpm exec vitest run --shard 1/4` runs both Reacts interleaved.

**Honest trade-off**: a 4-job matrix split by React is slightly faster wall-clock than a 2-job matrix that runs both Reacts per shard, so the CI shape choice is independent of this refactor. The win is local DX: `pnpm test` becomes a single command that covers both Reacts, with labelled output. Worth considering as a follow-up — not necessarily this PR.

test/react-19/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@radix-ui/react-popover": "1.1.15",
+    "@radix-ui/react-select": "2.2.6",
+    "@testing-library/react": "16.3.2",
+    "@types/react": "19.2.15",
+    "@types/react-dom": "19.2.3",
+    "motion": "12.40.0",
```
diegohaz: **Version drift between the React 18 and React 19 graphs.** `motion` is pinned to two different versions across the matrix:

```
examples/package.json:19      "motion": "12.38.0",    ← React 18 graph
test/react-19/package.json:12 "motion": "12.40.0",    ← React 19 graph (this file)
app/package.json:86           "motion": "12.40.0",    ← app
```

Because `vitest.config.ts` resolves `motion` from `examples/package.json` when `ARIAKIT_TEST_REACT=18` and from `test/react-19/package.json` when `ARIAKIT_TEST_REACT=19`, the two matrix slots exercise two *different* motion bundles, not just two different Reacts. Renovate has been bumping `app/`/`nextjs/`/`test/react-19` motion routinely but `examples/` got left behind.

Fix: bump `examples/package.json` motion to `12.40.0` so the React 18 vs React 19 matrix isolates React only.

diegohaz (commented):
### Re-review focus: looking for a simpler alternative to the dual-React Vitest setup

Four independent research/prototyping passes (Vitest projects, pnpm-native primitives, prior art in 12 React libraries, structural scope cuts). All run against this branch at `a7d9f9fd`. Posted follow-ups on three threads; this is the synthesis.

**Validated wins (would reduce the diff without losing coverage):**

- **Drop `react-shadow` and `react-toastify` from `reactPeerPackages`.** Both are consumed exclusively by Playwright tests (`hovercard-shadow-dom/test-browser.ts`, `dialog-react-toastify/test-chrome-firefox.ts`); no Vitest entrypoint imports either. Removing them passes both shards on both React versions in my worktree. Every *other* entry I tried to drop produced a real hooks-graph failure, so the list bottoms out at 8 packages, not 0. Follow-up on the [`reactPeerPackages` thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297724532).
- **Derive `reactPeerPackages` from `test/react-19/package.json`'s `dependencies`.** Same array, computed at config load, kills the three-place sync hazard you already flagged in the source comment. Both prototypes (sub-agent runs and my own) pass. Combined with the trim above, the array literal collapses to ~3 lines + a `PLAYWRIGHT_ONLY` exclusion set. Same [thread](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297724532).
- **Fix `motion` version drift across the two React graphs.** `examples/package.json` is on `motion@12.38.0` while `test/react-19/package.json` and `app/package.json` are on `12.40.0`. The dual-React matrix currently isolates React × motion, not just React. New [inline comment](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297729938).

**Optional architectural alternative (modest gain, judgment call):**

- **Switch to Vitest `projects`.** Vitest 4 lets you define two named projects in one config, each with its own `resolve.alias` + `dedupe`. Collapses `test-react-18` / `test-react-19` scripts + the `ARIAKIT_TEST_REACT` env var into a single `pnpm test`. Prototype runs both Reacts in one invocation with labelled output. Trade-off: trades 12 lines of env-var selection + 2 scripts for 2 projects-config blocks. The `test/react-19` workspace, both `createRequire` anchors, and the `aliasPkg` helpers all stay — Vitest projects doesn't help with pnpm hoisting. New [inline comment](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297726094) with the sketch.

**Things that won't simplify it (researched and ruled out):**

- **pnpm catalogs**: version-spec reuse only, doesn't change resolution. Cosmetic.
- **pnpm `overrides` + two install profiles** (the MUI/Adobe approach): breaks `frozenLockfile`, requires two `pnpm install`s per shard, ~2× CI install minutes.
- **`"react-18": "npm:react@^18"` aliased package names**: pnpm overrides don't apply to peer deps, so every React-peered package (`motion`, `react-dom`, `@testing-library/react`, `react-router`, the Radix peers, `use-sync-external-store`) refuses to resolve against the aliased name. Working around it via `.pnpmfile.cjs` is strictly more code than the current Vite alias array.
- **`.pnpmfile.cjs` `readPackage` rewrites**: same install-time-baking problem.
- **Dropping React 18 from the Vitest matrix**: would silently regress consumers — sub-agent's trim experiment showed real React-18-runtime errors from `@radix-ui/react-context`, `react-router`, `motion`, RTL, and `use-sync-external-store` that no amount of type-level checking would catch.

**Prior art**: of 12 React libraries surveyed (Radix, react-aria, MUI, Chakra, Headless, react-hook-form, floating-ui, TanStack Query, Mantine, etc.), only two actually run a React-version matrix in CI — Adobe react-spectrum and MUI — and both use the install-swap pattern (re-install with a different React per matrix job). **No other public library uses the dual-workspace + Vite-alias-per-peer pattern this PR introduces.** That's not damning — this PR is the only design that keeps a single `pnpm install` and avoids the CI install duplication. Just notable.

**Recommendation, ranked:**

1. Take the trim + auto-derive + motion fix (small, validated, no downside).
2. Consider the Vitest projects refactor as a follow-up PR (not this one) — it's nice-to-have, not load-bearing.
3. Keep the rest of the architecture as designed. The two-workspace structure is the floor given the constraints; my agents independently confirmed each piece of it (workspace anchor for `createRequire`, `aliasPkg` regex shape, deferred `@ariakit/test/react` import) is doing real work.

Comment-only, no approval / request-changes. All findings carry the new review marker so they don't collide with the earlier run.

examples/package.json
diegohaz: Fixed in 2cc474e34 - kept the explicit React 19 peer graph because pnpm needs a separate peer-resolution parent, but reduced the setup complexity by deriving the Vitest alias list from package manifests, aligning the mirrored dependency versions, and documenting the workspace/package purpose.

packages/ariakit-react-components/src/form/form.tsx
diegohaz: Fixed in 2cc474e34 - added a comment explaining that React 19 types form submit handlers with SubmitEvent, so the handler event type is derived from the prop to support both React 18 and 19 types.

packages/ariakit-test/src/type.ts
diegohaz: Fixed in 2cc474e34 - removed React 17 and React Testing Library 12 from @ariakit/test peer support and added a breaking changeset for @ariakit/test.

examples/menu-wordpress-modal/menu.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 2cc474e34 - the WordPress SlotFill wrapper now forwards Fill props in the bubblesVirtually path and preserves render-prop children instead of silently dropping accepted props.

package.json
diegohaz: Fixed in 1a432aa83 - Vercel requires next in the root package.json for deployment detection, so root next is restored at 16.2.6 and Renovate now explicitly allows root next major updates.

.github/workflows/main.yml
diegohaz: Fixed in 2cc474e34 - the workflow now uses the pnpm script shortcuts directly, without the extra run.

test/react-19/package.json
diegohaz: Fixed in 2cc474e34 - kept test/react-19 as the explicit React 19 peer-resolution anchor, added a package description, and simplified the surrounding setup so default tests are React 19 while React 18 runs as the additional compatibility job.

pnpm-workspace.yaml
diegohaz: Fixed in 2cc474e34 - kept test/* in the workspace list and added a comment explaining that these are React version-specific peer-resolution anchors for Vitest.

vitest.config.ts
diegohaz: Fixed in 2cc474e34 - removed the hand-maintained reactPeerPackages array and now derives it from test/react-19/package.json, with a guard that mirrored packages must also be declared in examples/package.json.

vitest.setup.ts
diegohaz: Fixed in 2cc474e34 - added a short comment explaining that React Testing Library is deferred until a React example actually needs it.

.github/workflows/main.yml
diegohaz: Fixed in 2cc474e34 - split the workflow into normal Test (${{ matrix.shard }}) jobs for default React 19 and separate Test React 18 (${{ matrix.shard }}) jobs.

package.json
diegohaz: This is not an issue because Vitest projects do not remove the second pnpm peer-resolution graph; they mainly trade the explicit React 18 script/env selector for more project config. I kept the clearer CI split from the review request in 2cc474e34 and reduced the real maintenance cost by deriving the alias list from the package manifests instead.

test/react-19/package.json
diegohaz: Fixed in 2cc474e34 - bumped examples/package.json to motion 12.40.0 and refreshed the lockfile so the React 18 and React 19 Vitest graphs differ by React version, not by the motion package version.

diegohaz:
**Fixed**

- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297755496), [reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756228), [reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756290), [reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756402)): Kept the explicit React 19 peer-resolution workspace because pnpm needs a second peer graph, but simplified the setup by deriving Vitest aliases from `test/react-19/package.json`, documenting the workspace purpose, and aligning mirrored dependency versions.
- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756617), [reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756109)): Split CI into default React 19 `Test` jobs and additional `Test React 18` jobs, and removed the unnecessary `pnpm run` form.
- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297755764)): Removed React 17 and React Testing Library 12 from `@ariakit/test` peer support and added a breaking changeset.
- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297755868)): Forwarded WordPress SlotFill props in the `bubblesVirtually` path and preserved render-prop children.
- `1a432aa83` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297755955)): Restored the root `next` dependency because Vercel needs it for deployment detection, refreshed the lockfile, and added a Renovate rule that allows root `next` major updates.
- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297755651), [reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756499)): Added small comments explaining the React 19 submit event typing and the deferred React Testing Library import.
- `2cc474e34` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756822)): Aligned `motion` to `12.40.0` in the React 18 and React 19 Vitest graphs.
- `57a074927`: Stabilized the disclosure animation Vitest case after `Test (1/2)` exposed a CI timing edge around a 1000ms CSS transition and Testing Library default timeout.

**Invalid**

- `Not changed` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3297756709)): Vitest projects do not remove the second pnpm peer-resolution graph, so this PR keeps the simpler explicit CI split and reduces the real maintenance cost through manifest-derived aliases instead.

Verification: `pnpm install --lockfile-only --ignore-scripts`, `pnpm lint-fix`, `pnpm tsc`, focused React 19 and React 18 Vitest runs, `pnpm lint`, and `git diff --check`. `pnpm peers check` still reports only the existing tailwindcss-animate/tailwindcss and vitest/@vitest/coverage-v8 peer warnings. I also inspected the failed Vercel deployment and the failed `Test (1/2)` job, pushed targeted fixes, and the latest PR check snapshot is green.

packages/ariakit-scripts/src/react18.ts
```diff
@@ -0,0 +1,528 @@
...
+}
+
+function getRsyncArgs(rootPath: string, workspacePath: string) {
+  const excludeArgs = [...excludedSegments, ...excludedFiles].flatMap(
+    (pattern) => ["--exclude", pattern],
+  );
+
+  return [
+    "-a",
+    "--delete",
```
diegohaz, line 200: `getRsyncArgs` doesn't exclude `pnpm-lock.yaml`, so rsync overwrites the workspace's React-18 lockfile with the source's React-19 lockfile on every run. After `rewriteReactDependencies` rewrites the workspace's `package.json` files back to React 18, the lockfile no longer matches the manifests and `pnpm install --no-frozen-lockfile` has to redo the resolution step on every invocation.

I measured this on a warm temp workspace at this PR head:

| Run | After-install state | Total |
| --- | --- | --- |
| Current (`pnpm-lock.yaml` rsynced from source) | `Already up to date` after 4.2s of resolve + relink | 9.8s |
| With `--exclude pnpm-lock.yaml` | `Lockfile is up to date, resolution step is skipped` after 1.7s | 7.4s |

About 25% off the steady-state runtime for free, because the workspace already has the right lockfile from the previous run. The source-side `package.json` changes still propagate (rsync continues to overwrite manifests), and `--no-frozen-lockfile` will still pick up new deps the next time they appear, so correctness is unchanged.

```diff
   return [
     "-a",
     "--delete",
+    "--exclude",
+    "pnpm-lock.yaml",
     ...excludeArgs,
     `${rootPath}/`,
     `${workspacePath}/`,
   ];
```

CI runners are fresh so they don't benefit, but local iteration (`pnpm react18 dev-app`, repeated `pnpm test-react-18 ...` during a debugging session) does.

packages/ariakit-scripts/src/react18.ts
```diff
@@ -0,0 +1,528 @@
...
+  args: string[],
+): Promise<React18Command> {
+  const [command, ...commandArgs] = stripLeadingSeparator(args);
+
+  if (!command) {
+    throw new Error("Missing command. Use `ariakit react18 <script>`.");
+  }
+
+  const packageJson = await readRootPackageJson(rootPath);
+  if (command === "react18" || command === "test-react-18") {
```
diegohaz, line 400: This special case silently rewrites `pnpm react18 test-react-18 ...` to `pnpm run test ...` to break the infinite recursion that the literal forwarding would otherwise hit (`test-react-18` → `pnpm react18 test` → `ariakit react18 test-react-18` → loop).

That's the right runtime behavior, but two things would make it less of a footgun:

1. A one-line comment above the branch — without it, the next reader has to derive the `pnpm react18 ↔ test-react-18 ↔ react18` loop themselves to understand why both names short-circuit to `test`.
2. Optionally, `log("...")` so the user actually sees that their `pnpm react18 test-react-18 --shard 1/2` got executed as `pnpm run test --shard 1/2`. Right now the help text says "Run a root script or arbitrary command", which sets the expectation that `test-react-18` would run literally.

Suggestion:

```ts
  const packageJson = await readRootPackageJson(rootPath);
  // `pnpm react18 react18` and `pnpm react18 test-react-18` would recurse
  // through their own scripts. Short-circuit to the underlying `test` script.
  if (command === "react18" || command === "test-react-18") {
    log(`Mapping ${command} to test to avoid recursion`);
    return {
      bin: "pnpm",
      args: ["run", "test", ...commandArgs],
    };
  }
```

Non-blocking — the behavior is fine, this is just for clarity.

package.json
diegohaz: Withdrawing the Vitest `projects` suggestion. The new direction in `142d47d6e` (move React 18 to an isolated temp workspace via `ariakit react18`) goes further than the projects refactor would have — `vitest.config.ts` now has zero React-version branching, and the React 18 path tests against a real pnpm install of React 18 instead of an alias overlay. That's the better split.

The trade-off is the ~500-line `react18.ts` script and the extra `pnpm install` cost in the React 18 CI job, but those are localised and worth it for the cleaner production config. Verified the new path works end-to-end at the current head (see top-level summary).

test/react-19/package.json
diegohaz: Resolved. `test/react-19/package.json` was removed in `142d47d6e` and `examples/package.json` was bumped to `motion@12.40.0` in `2cc474e34`, so `examples`, `app`, and the root all share the same pin now. The new React 18 path inherits that single version via the temp-workspace install.

diegohaz (commented):
### Scope

Re-review of #6044. Baseline: the previously-reviewed head `a7d9f9fd`. Current head: `eb2a33b4`. New commits in this re-review window:

```
2cc474e34 Simplify React version tests
1a432aa83 Restore root Next dependency
57a074927 Stabilize disclosure animation test
142d47d6e Move React 18 tests into ariakit scripts
2c363cb72 Fix Next app build after React test changes
eb2a33b4c Stabilize lazy popover focus test
```

Read all files changed since the baseline (20 files) and expanded into the surrounding repo where it mattered: `vitest.config.ts` auto-derivation, the new `packages/ariakit-scripts/src/react18.ts` script, the `@ariakit/scripts/react` helper, the changed CI workflow, the new `peerDependencyRules` and renovate rules, the `@ariakit/test` peer narrowing, and the two test stabilisations.

### Validation performed

- `pnpm install --frozen-lockfile`, `pnpm tsc`, `pnpm lint` — all clean against this PR head.
- React 19 default path: `pnpm test --shard=1/2 --run examples/disclosure-content-animating/test.ts examples/popover-lazy/test.ts examples/composite/test.ts` — 6 tests pass.
- React 18 path: `pnpm test-react-18 --shard 1/2 --run …` and `--shard 2/2 --run …` against `popover-lazy` + `disclosure-content-animating` — both shards pass, including the new `waitFor(...toHaveFocus())` stabilisation.
- Local prototype of the `--exclude pnpm-lock.yaml` rsync optimisation flagged below — measured a steady-state run dropping from 9.8s → 7.4s (`pnpm install` going from 4.2s "Already up to date" to 1.7s "Lockfile is up to date, resolution step is skipped"). Reverted before finishing the review.

### Major architectural change since last review

`2cc474e34` + `142d47d6e` replaced the old `ARIAKIT_TEST_REACT` env-var + `test/react-19` workspace setup with a new `ariakit react18` command. The React 19 path is now the default, plain `pnpm test` against the single workspace. The React 18 path copies the checkout to a temp directory under `tmpdir/ariakit-react18/<hash>/workspace`, rewrites every `package.json` to the React 18 versions declared by `website/package.json`, runs `pnpm install --no-frozen-lockfile --prefer-offline`, and then forwards either a root script or an arbitrary command into the temp workspace. A chokidar watcher streams source changes (excluding `package.json` / `pnpm-lock.yaml`) into the temp workspace so iterative commands like `pnpm react18 dev-app` stay in sync.

The trade-off vs the prior approach:

- **Wins:** `vitest.config.ts` has zero React-version branching, no env-var, no `test/react-19` workspace. The React 18 graph is a real pnpm install of React 18 rather than a Vite-alias overlay, which means peer resolution actually happens against React 18 and any Radix / motion / react-router quirks under React 18 get genuine coverage.
- **Costs:** ~500 LOC in `react18.ts`, plus an extra `pnpm install` per React-18 CI shard (the GH-Actions setup composite installs once for the main workspace, then `pnpm test-react-18` does a second install into the temp workspace). The temp workspace is ~2.2 GB on disk per checkout root, persisted across runs for warm starts.

I tested both shards end-to-end on both Reacts and the architecture works correctly. The cost is acceptable for the wins.

### Findings in this re-review

1. **`getRsyncArgs` doesn't exclude `pnpm-lock.yaml`** — rsync overwrites the workspace's React-18 lockfile with the source's React-19 lockfile every run, forcing pnpm to redo resolution. Measured ~25% off the warm-run wall time by adding `--exclude pnpm-lock.yaml`. Inline thread on [`packages/ariakit-scripts/src/react18.ts:195`](https://github.com/ariakit/ariakit/pull/6044#discussion_r3298985660).
2. **`getReact18Command` silent recursion mapping** — `pnpm react18 react18` / `pnpm react18 test-react-18` short-circuit to `test` to break a real recursion loop, but the rewrite is undocumented and silent. A one-line comment + a `log()` would make it obvious. Inline thread on [`packages/ariakit-scripts/src/react18.ts:393`](https://github.com/ariakit/ariakit/pull/6044#discussion_r3298988134).

Both are non-blocking — the script behaves correctly as-is.

### Previously-flagged items, status check

All threads from the `a7d9f9fd` review either got addressed in the new commits or are no longer applicable:

- ✅ `@ariakit/test` React-17 changeset → added in `.changeset/100-test-react-17.md`, peer ranges narrowed to React 18/19 + RTL 13–16, `type.ts` workaround dropped.
- ✅ CI `pnpm run` redundancy → switched to `pnpm test-react-18` / `pnpm test` shorthand; also split into separate `test` and `test-react-18` jobs as discussed.
- ✅ `Fill` widening in `menu-wordpress-modal/menu.tsx` → destructured `children`, forwards `...props` in both branches, function-vs-node children correctly handled.
- ✅ `form.tsx` `SubmitEvent` derivation → explicit comment added.
- ✅ `vitest.setup.ts` deferred import → explanatory comment added.
- ✅ `vitest.config.ts` `reactPeerPackages` auto-derived from `examples/package.json` with a fail-fast guard requiring root mirroring.
- ✅ `motion` version drift between graphs → all three (`examples`, `app`, root) on `12.40.0`; `test/react-19/package.json` removed.
- ✅ Vitest `projects` alternative → withdrawn ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3298990005)), the temp-workspace direction is a cleaner split.
- ✅ Root `next` dependency → restored in `1a432aa83` with a matching renovate rule allowing major bumps.

### Things checked but not flagged

- The new `peerDependencyRules.allowedVersions` for WordPress and `use-memo-one` are needed once the root graph is React 19; React 18 path still satisfies the original peer ranges natively so the rules are inert there.
- `astro.config.ts` and `vitest.config.ts` both consume `@ariakit/scripts/react`; Next.js (Turbopack) intentionally doesn't, and pnpm's `publicHoistPattern: ["*react*"]` keeps the resolved React copy consistent across the workspace either way.
- The 7 packages newly added to root `devDependencies` (`@radix-ui/react-popover`, `@radix-ui/react-select`, `motion`, `react-router`, `react-shadow`, `react-toastify`, `use-sync-external-store`) are the price of the `hasReactPeerDependency` fail-fast guard. `react-shadow` and `react-toastify` are still Playwright-only at the use site, but mirroring them to root is the simpler trade than maintaining an explicit "Playwright-only" exclusion set — accepted.
- The stabilisations (`waitFor(...toHaveFocus())` in `popover-lazy/test.ts`, `transitionTimeout: 1500` in `disclosure-content-animating/test.ts`) are appropriate — focus lands in an effect, not synchronously, and the animation completion timing is environment-sensitive.
- The chokidar watcher is started even for one-shot commands like `pnpm test-react-18`; overhead is small (~100–200 ms) and the watcher is genuinely needed for `pnpm react18 dev-app` / `pnpm react18 test-watch`, so leaving it always-on is fine.

### Not validated

- `pnpm build`, `pnpm build-app-lite`, `pnpm test-solid`, the Playwright / visual suites — PR description reports them clean, and nothing in this diff hits those paths beyond what's covered by the React 19 default install (already validated).
- Behavior of the temp-workspace script on Windows — `pathKey` is set up for it, but no Windows runner exists in CI to confirm.

Comment-only, no approval / request-changes.

examples/package.json
```diff
@@ -16,20 +16,24 @@
...
+    "motion": "12.40.0",
+    "react": "19.2.6",
+    "react-dom": "19.2.6",
     "react-router": "7.15.0",
     "react-shadow": "20.6.0",
     "react-toastify": "11.1.0",
     "textarea-caret": "3.1.0",
-    "tiny-invariant": "1.3.3"
+    "tiny-invariant": "1.3.3",
+    "use-sync-external-store": "1.6.0"
```
diegohaz: Do we still need this dependency here? Why?

package.json
```diff
@@ -78,13 +80,15 @@
     "@changesets/apply-release-plan": "7.1.1",
     "@changesets/cli": "2.31.0",
     "@playwright/test": "1.60.0",
+    "@radix-ui/react-popover": "1.1.15",
+    "@radix-ui/react-select": "2.2.6",
```
diegohaz: Why do we need these dependencies here in the root package.json?

package.json
```diff
@@ -99,7 +103,8 @@
     "husky": "9.1.7",
     "jsdom": "29.1.1",
     "lint-staged": "17.0.5",
-    "next": "14.2.35",
+    "motion": "12.40.0",
```
diegohaz: Same here

package.json
```diff
@@ -108,8 +113,11 @@
     "postcss-import": "16.1.1",
     "prettier": "3.8.3",
     "prettier-plugin-tailwindcss": "0.8.0",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
+    "react": "19.2.6",
+    "react-dom": "19.2.6",
+    "react-router": "7.15.0",
+    "react-shadow": "20.6.0",
+    "react-toastify": "11.1.0",
```
diegohaz: Same

package.json
```diff
@@ -118,6 +126,7 @@
     "tailwindcss-animate": "1.0.7",
     "tsup": "8.5.1",
     "typescript": "6.0.3",
+    "use-sync-external-store": "1.6.0",
```
diegohaz: Same

pnpm-workspace.yaml
```diff
@@ -33,6 +33,17 @@ overrides:
...
+peerDependencyRules:
+  allowedVersions:
+    "@wordpress/components>react": "19"
+    "@wordpress/components>react-dom": "19"
+    "@wordpress/compose>react": "19"
+    "@wordpress/data>react": "19"
+    "@wordpress/icons>react": "19"
+    "@wordpress/primitives>react": "19"
+    "@wordpress/rich-text>react": "19"
+    "use-memo-one>react": "19"
```
diegohaz: Can we just allow react 19 for all dependencies instead of configuring per dependency?

renovate.json
diegohaz: I think we should just remove these.

vitest.config.ts
```diff
@@ -25,6 +26,63 @@ const includeWithStyles = [
   /disclosure-content-animating/,
 ];
 
+const ignoredReactPeerPackages = new Set([
+  "@playwright/test",
+  "@wordpress/components",
+  "vitest",
+]);
```
diegohaz: Why?

vitest.setup.ts
```diff
@@ -56,6 +55,8 @@ async function loadReact(dir: string) {
     `./${dir}/index.react.tsx`,
   );
   if (failedImport) return false;
+  // Defer React Testing Library until a React example actually needs it.
+  const { render: renderReact } = await import("@ariakit/test/react");
```
diegohaz: Do we still need this to be dynamic import? Couldn't it be static top-level import?

packages/ariakit-scripts/src/react18.ts
```diff
@@ -0,0 +1,528 @@
...
+    })
+    .on(
+      "unlink",
+      (path) => void removeWatchedPath(rootPath, workspacePath, path),
+    )
+    .on(
+      "unlinkDir",
+      (path) => void removeWatchedPath(rootPath, workspacePath, path),
+    )
+    .on("error", (error) => console.error(error));
```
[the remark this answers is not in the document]
diegohaz: Fixed in 68712b552 - watcher callbacks now route async copy/remove/mkdir work through a catch/log helper so rejected promises are reported instead of becoming unhandled rejections.

packages/ariakit-scripts/src/react18.ts
diegohaz: Fixed in 68712b552 - `pnpm-lock.yaml` is excluded from both the initial workspace sync and watcher paths, so warm React 18 workspaces keep their generated React 18 lockfile.

packages/ariakit-scripts/src/react18.ts
diegohaz: Fixed in 68712b552 - added a comment and log message for the `react18`/`test-react-18` recursion short-circuit so users can see the command is mapped to the underlying `test` script.

examples/package.json
diegohaz: Fixed in 2b58e68be - removed `use-sync-external-store` from `examples/package.json`; the package is still owned by `@ariakit/react-store`, which is where it is imported.

package.json
diegohaz: Fixed in 2b58e68be - removed the Radix packages from the root `package.json`; Vitest now resolves example dependencies from the `examples` workspace instead of requiring root mirrors.

package.json
diegohaz: Fixed in 2b58e68be - removed the root `motion` mirror. The dependency remains declared in `examples/package.json`, where the examples import it.

package.json
diegohaz: Fixed in 2b58e68be - removed the root `react-router`, `react-shadow`, and `react-toastify` mirrors; these are now resolved from the `examples` workspace when Vitest needs them.

package.json
diegohaz: Fixed in 2b58e68be - removed the root `use-sync-external-store` mirror; `@ariakit/react-store` keeps the real dependency declaration.

pnpm-workspace.yaml
diegohaz: Fixed in 2b58e68be - replaced the package-specific React 19 peer allowances with broad `react` and `react-dom` allowed versions for 19.

renovate.json
diegohaz: Fixed in 2b58e68be - removed the root-only Renovate exception. Major React/Next updates remain disabled by default and still explicitly enabled for `nextjs/**` and `templates/**`.

vitest.config.ts
diegohaz: Fixed in 2b58e68be - removed the ignored React-peer package set entirely. Vitest now aliases direct example dependencies from the `examples` workspace, so those root-mirroring exceptions are no longer needed.

vitest.setup.ts
diegohaz: Fixed in 2b58e68be - restored the static top-level `@ariakit/test/react` import in `vitest.setup.ts`.

packages/ariakit-scripts/src/react18.ts
```diff
@@ -0,0 +1,528 @@
...
+  "devDependencies",
+  "optionalDependencies",
+] as const;
+
+const excludedSegments = new Set([
+  ".astro",
+  ".git",
+  ".next",
+  ".open-next",
+  ".turbo",
```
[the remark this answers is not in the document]
diegohaz: Fixed in 68712b552 - the React 18 sync now excludes `.env*`, `.dev.vars`, `*.pem`, and `pnpm-lock.yaml` from rsync, the Node copy fallback, and watcher paths.

diegohaz:
**Fixed**

- `68712b552` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299073219)): Caught and logged async watcher callback failures in the React 18 workspace runner.
- `68712b552` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299073348)): Excluded `pnpm-lock.yaml` from React 18 workspace sync so warm temp workspaces keep their generated lockfile.
- `68712b552` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299073454)): Documented and logged the `react18`/`test-react-18` recursion short-circuit.
- `68712b552` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299077472)): Excluded local env/secret files from the React 18 temp workspace sync and watcher paths.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299073543)): Removed the unnecessary direct `use-sync-external-store` dependency from `examples`.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299074040)): Removed root Radix mirrors by resolving Vitest example dependencies from the `examples` workspace.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299074115)): Removed the root `motion` mirror.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299074212)): Removed root `react-router`, `react-shadow`, and `react-toastify` mirrors.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299074315)): Removed the root `use-sync-external-store` mirror.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299074986)): Replaced per-package React 19 peer allowances with broad `react`/`react-dom` allowed versions.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299075100)): Removed the root-only Renovate exception for Next major updates.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299075193)): Removed the obsolete Vitest ignored React-peer package set.
- `2b58e68be` ([reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299075300)): Restored the static top-level `@ariakit/test/react` import in `vitest.setup.ts`.

packages/ariakit-scripts/src/react18.ts
diegohaz: Addressed in `68712b552` — `pnpm-lock.yaml` is now in `excludedFiles`, plus the rsync also excludes `.dev.vars`, `.env*`, `*.pem` for the same security-cleanup pass. Re-measured at this PR head:

- Warm steady-state run: 6.6s (was 9.8s before the fix).
- pnpm install during that run: 720ms with `Lockfile is up to date, resolution step is skipped` (was 4.2s of "Already up to date" relinking).

Thanks for picking up the secret-file exclusions as part of the same change — that was a real bonus over what I flagged.

packages/ariakit-scripts/src/react18.ts
diegohaz: Addressed in `68712b552` — both the explanatory comment and the `log()` are in place, so the rewrite is visible at runtime and the next reader doesn't have to derive the recursion loop themselves.

diegohaz (commented):
### Scope

Re-review of #6044. Baseline: the previously-reviewed head `eb2a33b4`. Current head: `a464e718`. New commits in this re-review window:

```
68712b552 Harden React 18 workspace runner
2b58e68be Simplify React test dependency resolution
370fb842c Refactor Renovate configuration to remove redundant major update rules for React and Next packages
a464e7186 Remove React alias helper
```

Read every file changed since the baseline (11 files) plus the resulting full state of `vitest.config.ts`, `app/astro.config.ts`, `pnpm-workspace.yaml`, `renovate.json`, and `packages/ariakit-scripts/src/react18.ts`. Confirmed no `@ariakit/scripts/react` consumers remain (the file is deleted, and both former call sites were updated in the same commit).

### Both prior findings are addressed

- **rsync `pnpm-lock.yaml` exclusion** (`68712b552`) — `pnpm-lock.yaml` is now in `excludedFiles`, and the same change also excludes `.dev.vars`, `.env*`, and `*.pem` as a secret-cleanup pass. Verified the perf win on this head: a warm `pnpm test-react-18 …` run now takes 6.6s (was 9.8s), with `pnpm install` dropping to 720ms and reporting `Lockfile is up to date, resolution step is skipped`. [Reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299406598).
- **`getReact18Command` recursion clarity** (`68712b552`) — explanatory comment plus the `log("Mapping … to test to avoid recursion")` are in place. [Reply](https://github.com/ariakit/ariakit/pull/6044#discussion_r3299406688).

`68712b552` also wraps the chokidar handlers in `logWatcherPromise` so background failures surface instead of being swallowed — that wasn't on my list, but it's a clean robustness fix.

### Broader simplification since last review

`2b58e68be` + `a464e7186` removed substantial machinery without losing coverage:

- `getReactViteAliases` / `reactDedupe` / the whole `@ariakit/scripts/react` export — deleted. `app/astro.config.ts` and `vitest.config.ts` no longer set `resolve.alias` / `dedupe` for React; they rely on pnpm's `publicHoistPattern: ["*react*"]` and identical version pins across workspaces to keep the React instance single.
- `vitest.config.ts` dropped the `requireExamples` / `readPackageJson` / `hasReactPeerDependency` / "must be declared in root" guard. The `reactPeerPackages` ladder collapsed to a small `shouldAliasExamplePackage` filter.
- Root `package.json` lost the seven devDeps that were mirrored just to satisfy the prior guard (`@radix-ui/react-popover`, `@radix-ui/react-select`, `motion`, `react-router`, `react-shadow`, `react-toastify`, `use-sync-external-store`).
- `examples/package.json` dropped `use-sync-external-store` — `packages/ariakit-react-store` still declares it as a direct dependency, so it's available where it's actually imported.
- `pnpm-workspace.yaml` replaced the eight per-package WordPress/`use-memo-one` allow rules with a single global `react: "19"` / `react-dom: "19"` rule.
- `vitest.setup.ts` reverts the deferred `@ariakit/test/react` import to a top-level static import, since the React-graph isolation reason no longer applies.
- `renovate.json` (`370fb842c`) drops the two paired rules that disabled major React/Next updates everywhere except `nextjs/**` and `templates/**`. With the matrix now genuinely covering React 18 (via the temp workspace) and React 19, those guardrails are redundant. `website/package.json` stays on React 18 by the existing `matchFileNames: ["website/package.json"], "enabled": false` rule at the top of the config — which is what `react18.ts` needs for `getReact18DependencyVersions`.

The trade-off worth naming: removing explicit `resolve.alias` / `dedupe` shifts the "single React instance" guarantee entirely onto pnpm's hoisting plus the shared version pins. The grouped `"react"` renovate rule keeps `react` and `react-dom` updating together, so the practical risk of a version skew between workspaces is small. If a future workspace ever pins a different React minor (intentionally or via a rogue Renovate run), Vitest would start seeing two copies again — but tests would fail loudly at that point, which is acceptable.

### Validation performed

- `pnpm install --frozen-lockfile`, `pnpm tsc` — clean against `a464e718`.
- React 19 path: `pnpm test --shard=1/2 --run examples/disclosure-content-animating/test.ts examples/popover-lazy/test.ts examples/composite/test.ts` — 6 tests pass with no Vite alias config at all.
- React 18 path: `pnpm test-react-18 --shard 1/2 --run examples/popover-lazy/test.ts examples/disclosure-content-animating/test.ts` — passes, including the rsync optimisation (`Lockfile is up to date, resolution step is skipped`).

### Findings

None new in this re-review. Comment-only, no approval / request-changes.

PR #6053, opened 2026-05-26
Split performance workflow

diegohaz:
## Motivation

The app workflow currently owns the pull-request-only performance job, which means it needs an event guard inside a workflow that also runs on `main`. Keeping perf execution and PR posting in the same job also makes it harder to add more performance runners later and consolidate their results into one comment.

## Solution

Move performance coverage into a new `.github/workflows/perf.yml` workflow that runs only for pull requests. The workflow has its own `Build app` and `Build Next.js app` jobs, renames the browser performance runner to `Chrome`, and uploads Chrome results as `perf-results-chrome`.

Add a `Prepare baseline` job that resolves the base app and Next.js build artifacts once, using a successful `app.yml` run for the PR base SHA when available and falling back to building them from the base checkout. Chrome still creates a baseline worktree and installs baseline dependencies to run tests against the base source, but it now downloads the prepared baseline build artifact instead of repeating the artifact lookup or fallback build. This gives future Chrome shards a shared baseline artifact to consume.

Add a separate `Post PR comment` job that downloads `perf-results-*` artifacts and combines available `comparison.md` files into one bot comment. The post job owns the top-level `## Performance` heading and renders each result group under a `###` title. Chrome writes `title.txt` metadata with the `Chrome` display title, and the post job falls back to formatted artifact names when metadata is absent. Results with the same title are grouped into one section so future Chrome shards can upload separate artifacts without creating duplicate headings.

## Verification

- `pnpm lint-fix`
- `pnpm exec prettier --check .github/workflows/app.yml .github/workflows/perf.yml`
- `pnpm exec prettier --check .github/workflows/perf.yml`
- `ruby -e 'require "yaml"; ARGV.each { |file| YAML.load_file(file); puts "ok #{file}" }' .github/workflows/app.yml .github/workflows/perf.yml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/perf.yml"); puts "ok .github/workflows/perf.yml"'`
- `git diff --check`
- `actionlint .github/workflows/app.yml .github/workflows/perf.yml` with ignores for stale local metadata around `actions/create-github-app-token@v3`; confirmed the live `v3` action metadata exposes `client-id` and marks `app-id` deprecated.
- `actionlint .github/workflows/perf.yml` with the same metadata ignore.
- Downloaded the previous `perf-results-chrome` artifact from run `26451284951` and verified the updated post script finds the root `comparison.md` and renders it as one `## Performance` comment with a `### Chrome` section.


.github/workflows/perf.yml
```diff
@@ -0,0 +1,393 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build app
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
```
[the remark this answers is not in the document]
diegohaz: It's fine if they are built-in github actions (`actions/*`)

.github/workflows/perf.yml
```diff
@@ -0,0 +1,479 @@
...
+                .split('-')
+                .filter(Boolean)
+                .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+                .join(' ');
+            }
+
+            function readComparison(file, title, options = {}) {
+              try {
+                const body = fs.readFileSync(file, 'utf-8').trim();
+                return `## ${title}\n\n${body}`;
```
diegohaz: The wrapper here adds `## ${title}`, but `comparison.md` already opens with its own `## Performance` heading (emitted at `app/src/test-utils/perf-compare.ts:675`). The PR comment posted by the latest workflow run on this PR ends up with two consecutive H2s at the same level:

```markdown
## Chrome

## Performance

No significant performance changes detected.
…
```

That reads as two sibling sections rather than parent/child. The redundancy will multiply once future runners are wired up — each runner block would repeat `## Performance` under its runner heading.

Now that the post script owns the per-runner heading, the cleanest fix is to drop the `## Performance` line in `perf-compare.ts` so each runner section reads `## Chrome\n\n<body>` directly. (Alternatively, skip the wrapper when there's only one section, but that defeats the multi-runner design the script was restructured for.)

.github/workflows/perf.yml
```diff
@@ -0,0 +1,393 @@
...
+        with:
+          script: |
+            const baseSha = '${{ github.event.pull_request.base.sha }}';
+            const runs = await github.rest.actions.listWorkflowRuns({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'app.yml',
+              branch: '${{ github.event.pull_request.base.ref }}',
+              head_sha: baseSha,
+              status: 'success',
```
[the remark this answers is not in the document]
diegohaz: Seconding this specifically for `base.ref` — branch names can contain `'` (and other characters) that would terminate the JS single-quoted string and corrupt the rest of the script. `base.sha` is a 40-char hex SHA so it's safe to leave inline, but routing `base.ref` through `env` is a worthwhile hardening with essentially no churn. If you'd rather keep both consistent, applying the suggested diff verbatim is also fine.

diegohaz (commented):
## Review summary

Initial review of `split-perf-workflow` at `889b7b54c4`. Diff baseline: merge-base with `main` (`80cecface…`).

### Scope reviewed

- `.github/workflows/perf.yml` — full file: `build`, `build-nextjs`, `baseline`, `chrome`, and `post` jobs, including the embedded `github-script` snippets in `baseline` and `post`.
- `.github/workflows/app.yml` — verified the old `perf` job is fully removed and nothing else (e.g. `deploy-preview`, `test-finalize`) referenced it.
- Cross-references: `.github/workflows/setup/action.yml`, `.github/workflows/install-playwright/action.yml`, `app/src/test-utils/perf-compare.ts` (to locate the heading source), the bot comment posted on this PR by run `26452334723`, and the `Post PR comment` logs from the prior runs `26452021273` and `26451284951`.

### Validation performed

- Confirmed the latest run on `889b7b54c4` is green through `Post PR comment` and inspected the resulting bot comment, which surfaced the duplicate-H2 finding below.
- Walked the `baseline` fallback path: `find-base-artifact` → conditional download of `app-build` and `nextjs-dist` from a prior `app.yml` run → cleanup-or-build → `Check baseline artifacts` → `Pack baseline artifacts` → `Upload baseline artifacts`. The `continue-on-error: true` + `steps.X.outcome != 'success'` chain correctly handles "no prior successful `app.yml` run", "artifact expired", and "partial download" (cleans up before rebuilding both).
- Confirmed empirically (via the `840cefbbd` run, which logged `No performance comparisons found` against a script that only iterated subdirectories) that `pattern: perf-results-*` + `merge-multiple: false` lays a single matching artifact at the download root rather than under `perf-results-chrome/`. The `rootSection`-then-subdir handling added in `889b7b54c` covers both the current single-runner layout and the subdir layout that will appear once a second `perf-results-*` artifact lands.
- Verified `git worktree add --detach "$BASELINE_DIR" <base.sha>` works against the chrome job's `fetch-depth: 0` checkout, and that the `HUSKY=0` env on the baseline `pnpm install` is necessary because the worktree shares `.git/config` with the main checkout (without it, husky's `prepare` would rewrite `core.hooksPath` to a path under `$RUNNER_TEMP`).
- Verified branch-protection contexts on `main` (`Lint`, `Type check`, `Test Chrome`, `Visual / Finalize`, etc.) do not require a `Performance` or `Chrome` check from this workflow, so renaming the runner is not a protection break. Job names `Build app` and `Build Next.js app` now appear twice in the PR checks list (once per workflow) — cosmetic but each remains a distinct status context.
- Re-confirmed the post-job gate semantics: `always() && needs.chrome.result == 'success' && <same-repo>` is forward-compatible — adding new `perf-results-*` runners to `needs` won't block posting when only Chrome succeeds, and skips cleanly for fork PRs. The `defaultTitle = "Chrome"` is correct here because Chrome is the gating job, so the root-layout case implies Chrome is the source.

### Findings

- **Inline:** the post script's `## ${title}` wrapper duplicates the `## Performance` heading already emitted by `perf-compare.ts`, producing two sibling H2s in the bot comment. Details on `.github/workflows/perf.yml:407`.
- **Reply on the existing CodeRabbit thread** about interpolating `github.event.pull_request.base.ref` into the `github-script` JS string — still relevant: `base.sha` is safe inline (hex), but `base.ref` can contain `'` and warrants going through `env`.

CodeRabbit's other two threads are not re-raised:

- SHA pinning of `actions/*`: explicitly dismissed in-thread, matches the repo learning.
- `persist-credentials: false` on `Build app` / `Build Next.js app`: the build/build-nextjs jobs perform no git operations after checkout, and the omission matches the convention in `app.yml`, `main.yml`, and the rest of the perf workflow that doesn't subsequently `git worktree add`. The `baseline` and `chrome` jobs (which do touch git) already set `persist-credentials: false`.

### Not validated locally

- I did not execute a full chrome perf round locally; validation is from inspecting the YAML, the embedded scripts, prior run logs, and the live PR comment. Behavior under multi-runner subdirectory layout is inferred from `actions/download-artifact@v8` semantics rather than observed end-to-end.

.github/workflows/perf.yml
diegohaz: Fixed in fde5b9c - the post job now owns the `## Performance` heading and renders each result group as a `### ...` section. Chrome writes `title.txt` metadata for its display title, and results with the same title are grouped so future Chrome shards do not create duplicate headings.

diegohaz:
**Fixed**

- fde5b9c ([reply](https://github.com/ariakit/ariakit/pull/6053#discussion_r3304582913)): Reworked the performance comment to render one `## Performance` heading with grouped `###` result sections, using title metadata for Chrome and future runners.

.github/workflows/perf.yml
diegohaz: `fde5b9cb5` resolves this: `formatBody` strips the leading `## Performance\s+` from each `comparison.md`, and the post script now emits a single outer `## Performance` with `### ${title}` per runner. Verified against the YAML; CI for this commit is still running, so I haven't been able to confirm against the rendered bot comment yet.

Optional follow-up (not blocking): now that each runner's body is unwrapped under `### Chrome`, the per-test headings emitted by `perf-compare.ts` (e.g. `### examples/dialog-framer-motion/perf-chrome.ts > react > open dialog`) become outline siblings of the runner heading rather than children of it. It looks fine visually because they're inside `<details>`, but in the multi-runner shape (`## Performance` → `### Chrome` → `### test 1` → `### Firefox` → `### test 1`) the test-vs-runner boundary becomes ambiguous in the outline. Demoting those headings (and the `### New tests` / `### Removed tests` / `### Unpaired tests` ones) to `####` in `perf-compare.ts` would make the runner the actual parent. Just noting it since this PR shifted them — not worth blocking on.

diegohaz (commented):
## Re-review summary

Incremental re-review of `split-perf-workflow` at `fde5b9cb5`. Baseline: prior review at `889b7b54c`. One new commit in scope: `fde5b9cb5 Format performance comment sections` (perf.yml only, +45/−14).

### What changed since the last review

- `Build app` and `Build Next.js app` checkouts now set `persist-credentials: false`.
- New job-level env `PERF_RESULTS_TITLE: Chrome` on the `chrome` job, plus a `Write performance metadata` step that writes that value to `app/.perf-results/title.txt` (always-runs, after the perf round).
- Post script restructured: `formatBody` strips the leading `## Performance\s+` from each `comparison.md`; `readTitle` consumes `title.txt` (falling back to `formatName(entry.name)` for subdirs and `PERF_RESULTS_DEFAULT_TITLE` for the root layout); `readSections` now groups by title via a `Map` so future sharded runs with the same `title.txt` merge into one section; each runner becomes `### ${title}`; the whole comment is wrapped in a single outer `## Performance`.

### Validation performed

- Diffed `889b7b54c..fde5b9cb5` and traced the new control flow end-to-end.
- Walked the post-script logic against root-only, multi-subdir, and missing-`title.txt` cases. The Map-based grouping correctly merges sharded runs with the same title (forward-looking, currently dead-functional) and the fallback chain keeps section labels reasonable when `title.txt` is absent.
- Confirmed `pnpm-lock.yaml` has no `git+…` URLs, so `persist-credentials: false` on the build jobs is safe (those jobs don't perform git operations after checkout, and `pnpm install --frozen-lockfile` doesn't need git auth here).
- Confirmed `Write performance metadata` is ordered after `Run interleaved perf tests` (which itself does `rm -rf app/.perf-results` at the start of each round), so `title.txt` survives into the upload regardless of round count.
- Confirmed `## Performance` is still emitted at `app/src/test-utils/perf-compare.ts:675`, so `formatBody`'s `/^## Performance\s+/` continues to match; `perf-compare.test.ts` doesn't assert on that heading specifically (only on inner H3s like `### Unpaired tests`), so the unit tests are unaffected.
- CI run `26455255142` for this SHA was still in progress through the `Chrome` job at the time of review, so the actual rendered bot comment hasn't been refreshed yet — verification of the new comment shape is by inspection of the script and YAML, not the live comment.

### Status of prior findings

- **Duplicate-H2 on `perf.yml:407`:** addressed by `fde5b9cb5`. Replied on the thread to confirm, plus an optional follow-up about per-test heading levels.
- **CodeRabbit env-var interpolation on `Find base app artifact`:** still open — `fde5b9cb5` does not touch lines 75/80, so `const baseSha = '${{ … }}'` and `branch: '${{ … }}'` remain inline-interpolated. Not reposting per re-review policy; the existing CodeRabbit thread already carries the suggestion and my prior reply seconding it for `base.ref`.
- **CodeRabbit `persist-credentials` on the build jobs:** addressed by `fde5b9cb5`.
- **CodeRabbit SHA-pinning thread:** still resolved as dismissed.

### Findings in this re-review

None new on the changed code beyond the optional H3-nesting follow-up noted on the prior duplicate-H2 thread.

PR #6056, opened 2026-05-26
Add Ariakit store public API tests

diegohaz:
## Motivation

`@ariakit/store` did not have Vitest coverage for its framework-agnostic public API. The package is small but central to the component stores, so public behaviors such as listener timing, lifecycle setup, derived stores, and merged stores should be covered without relying on implementation details.

## Solution

Add `packages/ariakit-store/src/test.ts` with public API tests for type inference, state updates, `subscribe`/`sync`/`batch`, `setup`/`init`, parent store synchronization, `pick`/`omit`, `mergeStore`, and `throwOnConflictingProps`.

Because this package test intentionally uses the `src/test.ts` name, the Vitest example auto-loader now only handles tests under `examples/` and `sandbox/` paths, allowing package tests with that filename to run normally. The generated package `.npmignore` template and generated ignore files now exclude colocated test files and `__tests__` helpers so these tests are not published in package tarballs.

## Verification

- `pnpm lint-fix`
- `pnpm test app/src/examples/checkbox-card/test.ts`
- `pnpm test examples/checkbox/test.ts`
- `pnpm tsc`
- `npm pack --dry-run --json` in `packages/ariakit-store`
- `npm pack --dry-run --json` in `packages/ariakit-react-components`
- `npm pack --dry-run --json` in `packages/ariakit-test`


vitest.setup.ts
```diff
@@ -125,6 +125,9 @@ function parseTest(filename?: string) {
   if (!match?.groups) return false;
   const { dir, loader } = match.groups;
   if (!dir) return false;
+  if (!dir.startsWith("examples/") && !dir.includes("app/src/examples/")) {
+    return false;
+  }
```
diegohaz: It's probably better to add `(examples|sandbox)` to the regex instead (or something like that).

packages/ariakit-store/src/test.ts
```diff
@@ -0,0 +1,473 @@
...
+  throwOnConflictingProps,
+} from "./index.ts";
+import type { Store, StoreOptions, StoreProps, StoreState } from "./index.ts";
+import * as storeExports from "./index.ts";
+
+const originalNodeEnv = process.env.NODE_ENV;
+
+afterEach(() => {
+  process.env.NODE_ENV = originalNodeEnv;
+});
```
diegohaz: Can't we use a vitest mock feature here instead? Also maybe it works with the `using` keyword?

packages/ariakit-store/src/test.ts
```diff
@@ -0,0 +1,473 @@
...
+    "init",
+    "mergeStore",
+    "omit",
+    "pick",
+    "setup",
+    "subscribe",
+    "sync",
+    "throwOnConflictingProps",
+  ]);
+});
```
diegohaz: This test is unnecessary

vitest.setup.ts
diegohaz: Tried folding the constraint into the regex. One thing to watch out for: there are existing nested example tests at `app/src/examples/{disclosure,popover,separator}/_component/test.ts`, so the regex needs to keep allowing extra segments between the example folder and `test.ts`. A simple `[^/]+` between the segment name and `test.ts` would reject those.

This works for everything in the current tree:

```ts
const match = filename.match(
  /(?<dir>(?:.*\/)?(?:examples|sandbox)\/.+?)\/test\.((?<loader>react|solid)\.)?ts$/,
);
if (!match?.groups) return false;
const { dir, loader } = match.groups;
if (!dir) return false;
```

Ran it against the existing inventory plus the new package path:

- `examples/checkbox/test.ts` → `dir=examples/checkbox`
- `examples/checkbox/test.react.ts` → `dir=examples/checkbox`
- `app/src/examples/checkbox-card/test.ts` → `dir=app/src/examples/checkbox-card`
- `app/src/examples/disclosure/_component/test.ts` → `dir=app/src/examples/disclosure/_component`
- `packages/ariakit-store/src/test.ts` → rejected

Same accept/reject set as the current `startsWith`/`includes` block, and that block can be removed.

packages/ariakit-store/src/test.ts
```diff
@@ -0,0 +1,473 @@
...
+  sync,
+  throwOnConflictingProps,
+} from "./index.ts";
+import type { Store, StoreOptions, StoreProps, StoreState } from "./index.ts";
+import * as storeExports from "./index.ts";
+
+const originalNodeEnv = process.env.NODE_ENV;
+
+afterEach(() => {
+  process.env.NODE_ENV = originalNodeEnv;
```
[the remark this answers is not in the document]
diegohaz: Confirming this — switching to `vi.stubEnv` / `vi.unstubAllEnvs` removes the coercion bug and is what the sibling thread is asking about. Combined fix:

```ts
afterEach(() => {
  vi.unstubAllEnvs();
});

// in the production test
vi.stubEnv("NODE_ENV", "production");
```

`vi.unstubAllEnvs` snapshots and restores the original `process.env.NODE_ENV` (including the `undefined` case), so the manual `originalNodeEnv` save/restore can be dropped entirely. Setting `unstubEnvs: true` on `vitest.config.ts` would also auto-restore between every test if the explicit `afterEach` isn't wanted.

packages/ariakit-store/src/test.ts
diegohaz: Yes — `vi.stubEnv` works here. Replace the manual save/restore with:

```ts
import { afterEach, expect, expectTypeOf, test, vi } from "vitest";

afterEach(() => {
  vi.unstubAllEnvs();
});
```

And in the production test:

```ts
vi.stubEnv("NODE_ENV", "production");
```

This also fixes the adjacent thread's `undefined` → `"undefined"` coercion concern: `vi.unstubAllEnvs` snapshots the original value (including `undefined`) and restores it correctly. Adding `unstubEnvs: true` to `vitest.config.ts` would skip the explicit `afterEach` and auto-restore between every test.

For `using`/explicit resource management: vitest 4.x doesn't attach a `Symbol.dispose` to `vi.stubEnv`, so it'd be a custom wrapper rather than a built-in. `stubEnv` + `unstubAllEnvs` is the idiomatic vitest pattern and already a single line per test, so I'd lean on that here.

packages/ariakit-store/src/test.ts
diegohaz: Agreed — the type-surface test below (`preserves the public type surface`) imports each named export individually, so removing or renaming any of `batch`/`createStore`/etc. already surfaces as a `tsc` failure. The only thing this snapshot catches is *new* accidental exports added to the barrel, which doesn't really earn its keep against the maintenance cost. Safe to drop.

diegohaz (commented):
Initial review at head `8deaa1b3c`, base `57d0fa7c0`.

### What I checked

- New `packages/ariakit-store/src/test.ts` — ran `pnpm test packages/ariakit-store/src/test.ts` (16 passed) and walked each case against the behavior in `packages/ariakit-store/src/index.ts`.
- Auto-loader change in `vitest.setup.ts` — ran `pnpm test examples/checkbox/test.ts` and `pnpm test app/src/examples/checkbox-card/test.ts`, then surveyed nested example tests (`app/src/examples/{disclosure,popover,separator}/_component/test.ts`) so the regex constraint discussion accounts for them.
- `.npmignore` template + generated files — ran `npm pack --dry-run --json` in `packages/ariakit-store`, `packages/ariakit-react-components`, and `packages/ariakit-test`. No `test.ts` / `__tests__/` / `*.test.*` paths show up in any of the tarballs.
- `pnpm tsc` — passes.

### Findings

All feedback is on the four existing open threads as replies with concrete proposals (validated locally):

1. `vitest.setup.ts` regex thread — proposed a regex that folds the `examples`/`sandbox` gate in while keeping the nested `_component/test.ts` cases working.
2. `packages/ariakit-store/src/test.ts:20` (Copilot, NODE_ENV coercion) — switch to `vi.stubEnv` / `vi.unstubAllEnvs`.
3. `packages/ariakit-store/src/test.ts:21` (vitest mock / `using` question) — same `vi.stubEnv` answer; `using` would need a custom wrapper in vitest 4.x.
4. `packages/ariakit-store/src/test.ts:40` (runtime-exports test) — confirmed the snapshot is redundant with the per-export imports in the type-surface test and can be removed.

No additional findings beyond those four threads. Test cases themselves look correct against the store implementation — `subscribe`/`sync`/`batch` lifecycle, `setup`/`init` instance counting, parent-store propagation, `pick`/`omit` mirroring, and `mergeStore` precedence + dispatch all match the source behavior.

packages/ariakit-store/src/test.ts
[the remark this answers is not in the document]
diegohaz: Fixed in e8ac950c60fa - replaced the manual `NODE_ENV` save/restore with `vi.stubEnv` and `vi.unstubAllEnvs`, so an originally undefined env value is restored correctly.

vitest.setup.ts
diegohaz: Fixed in e8ac950c60fa - folded the `examples`/`sandbox` path constraint into the `parseTest` regex while keeping nested example test paths covered.

packages/ariakit-store/src/test.ts
diegohaz: Fixed in e8ac950c60fa - removed the runtime export snapshot test and the namespace import.

packages/ariakit-store/src/test.ts
diegohaz: Fixed in e8ac950c60fa - switched this test file to Vitest env stubbing helpers with `vi.stubEnv` and `vi.unstubAllEnvs`.

diegohaz:
**Fixed**

- e8ac950c60fa ([reply](https://github.com/ariakit/ariakit/pull/6056#discussion_r3307429750)): Folded the `examples`/`sandbox` constraint into the `parseTest` regex.
- e8ac950c60fa ([reply](https://github.com/ariakit/ariakit/pull/6056#discussion_r3307429749)): Replaced manual `NODE_ENV` restoration with Vitest env stubs.
- e8ac950c60fa ([reply](https://github.com/ariakit/ariakit/pull/6056#discussion_r3307429754)): Used `vi.stubEnv` and `vi.unstubAllEnvs` for the production env test.
- e8ac950c60fa ([reply](https://github.com/ariakit/ariakit/pull/6056#discussion_r3307429753)): Removed the unnecessary runtime export snapshot test.

diegohaz (commented):
Re-review at head `e8ac950c6`. Incremental baseline = the prior review head `8deaa1b3c`; only commit in range is `e8ac950c6 "Address store test review feedback"` (`packages/ariakit-store/src/test.ts` + `vitest.setup.ts`, +3 / -24).

### Prior threads — verified against the new commit

1. `vitest.setup.ts` regex: now `/^(?<dir>(?:.*\/)?(?:examples|sandbox)\/.+?)\/test\.((?<loader>react|solid)\.)?ts$/` with the runtime `startsWith`/`includes` check removed. Matches the proposal in the prior reply.
2. NODE_ENV coercion (Copilot): `afterEach` now calls `vi.unstubAllEnvs()`; the `originalNodeEnv` capture is gone.
3. vitest mock / `using` thread: production test uses `vi.stubEnv("NODE_ENV", "production")`.
4. Runtime-exports snapshot test: removed along with the `storeExports` namespace import.

### What I ran

- `pnpm test packages/ariakit-store/src/test.ts` — 15 tests pass (was 16, minus the removed exports snapshot).
- `pnpm test examples/checkbox/test.ts` — passes (top-level `examples/` path still picked up by the new regex).
- `pnpm test app/src/examples/disclosure/_component/test.ts` — passes (nested `_component/test.ts` still matched, the case that motivated `.+?` over `[^/]+`).
- `pnpm tsc` — passes.

### New findings

None. Agreeing with Copilot's re-review (no new comments). The earlier four threads can stay as the author marked them — fixed.

PR #6057, opened 2026-05-27
Improve store benchmarks

diegohaz:
## Motivation

The `@ariakit/store` benchmark covered primitive create/read/update cases, but it did not exercise several paths that show up in real Ariakit usage: React-style whole-store subscriptions, setup/init-derived listeners, and live `mergeStore` chains with `pick`/`omit` sources. That left the benchmark less representative of the critical store paths used by component stores.

## Solution

Expand the store benchmark fixture with composite/popover-like state and add benchmarks for React null-subscriber fanout, setup-driven derived state, bidirectional synced-store writes, cold merged-chain initialization, and live updates through external, picked, omitted, and derived merged-store paths. The derived-state helper follows the combobox sync/batch scheduling shape and uses a map-backed item lookup to keep the benchmark focused on store overhead.

## Verification

- `pnpm lint-fix`
- `pnpm exec vitest bench --config vitest.bench.config.ts packages/ariakit-store/benchmark/store.bench.ts`
- `pnpm tsc`

### DIFF
--- a/packages/ariakit-store/benchmark/store.bench.ts
+++ b/packages/ariakit-store/benchmark/store.bench.ts
@@ -3,20 +3,42 @@ import {
   createStore,
   init,
   mergeStore,
+  omit,
   pick,
+  setup,
   subscribe,
+  sync,
 } from "@ariakit/store";
 import { bench } from "vitest";
 
+interface BenchmarkItem {
+  id: string;
+  value: string;
+  disabled?: boolean;
+}
+
 interface BenchmarkState {
   activeId: string | null;
+  activeValue?: string;
+  animating: boolean;
+  anchorElement: unknown;
+  arrowElement: unknown;
+  baseElement: unknown;
   busy: boolean;
+  contentElement: unknown;
   count: number;
   disabled: boolean;
+  disclosureElement: unknown;
   focused: boolean;
-  items: string[];
+  focusLoop: boolean;
+  focusWrap: boolean;
+  items: BenchmarkItem[];
   mounted: boolean;
+  moves: number;
   open: boolean;
+  popoverElement: unknown;
+  renderedItems: BenchmarkItem[];
+  rtl: boolean;
   selectedId: string | null;
   value: string;
 }
@@ -27,16 +49,31 @@ const options = {
 };
 
 const itemIds = Array.from({ length: 100 }, (_, index) => `item-${index}`);
+const items = itemIds.map((id) => ({ id, value: id }));
+const itemById = new Map(items.map((item) => [item.id, item]));
 
 const initialState: BenchmarkState = {
   activeId: null,
+  activeValue: undefined,
+  animating: false,
+  anchorElement: null,
+  arrowElement: null,
+  baseElement: null,
   busy: false,
+  contentElement: null,
   count: 0,
   disabled: false,
+  disclosureElement: null,
   focused: false,
-  items: itemIds,
+  focusLoop: true,
+  focusWrap: true,
+  items,
   mounted: false,
+  moves: 0,
   open: false,
+  popoverElement: null,
+  renderedItems: items,
+  rtl: false,
   selectedId: null,
   value: "",
 };
@@ -57,6 +94,72 @@ function createBenchmarkStore(count = 0) {
   return createStore({ ...initialState, count });
 }
 
+function getItem(id: string | null) {
+  if (id == null) return;
+  return itemById.get(id);
+}
+
+function setupBenchmarkDerivedState(
+  store: ReturnType<typeof createBenchmarkStore>,
+) {
+  setup(store, () =>
+    sync(store, ["open", "animating"], (state) => {
+      store.setState("mounted", state.open || state.animating);
+    }),
+  );
+
+  setup(store, () =>
+    sync(store, ["moves", "activeId"], (state, prevState) => {
+      if (state.moves !== prevState.moves) return;
+      store.setState("activeValue", undefined);
+    }),
+  );
+
+  setup(store, () =>
+    batch(store, ["moves", "renderedItems"], (state, prevState) => {
+      if (state.moves === prevState.moves) return;
+      const activeItem = getItem(state.activeId);
+      store.setState("activeValue", activeItem?.value);
+    }),
+  );
+}
+
+function createMergedStoreChain(count = 0) {
+  const externalStore = createBenchmarkStore(count);
+  const tagStore = createBenchmarkStore(count);
+  const popoverStore = createBenchmarkStore(count);
+  const compositeStore = createBenchmarkStore(count);
+
+  const mergedStore = mergeStore(
+    externalStore,
+    pick(tagStore, ["value", "rtl"]),
+    omit(popoverStore, [
+      "arrowElement",
+      "anchorElement",
+      "baseElement",
+      "contentElement",
+      "disclosureElement",
+      "items",
+      "popoverElement",
+      "renderedItems",
+    ]),
+  );
+
+  const store = createStore(
+    {
+      ...mergedStore.getState(),
+      count,
+      activeId: itemIds[0] ?? null,
+    },
+    compositeStore,
+    mergedStore,
+  );
+
+  setupBenchmarkDerivedState(store);
+
+  return { externalStore, popoverStore, store, tagStore };
+}
+
 function flushBatch() {
   return new Promise<void>((resolve) => queueMicrotask(resolve));
 }
@@ -65,9 +168,12 @@ const readStore = createBenchmarkStore();
 const updateStore = createBenchmarkStore();
 const selectedSubscribersStore = createBenchmarkStore();
 const unrelatedSubscribersStore = createBenchmarkStore();
+const reactSubscribersStore = createBenchmarkStore();
 const batchStore = createBenchmarkStore();
+const cascadingStore = createBenchmarkStore();
 const parentStore = createBenchmarkStore();
 const syncedStore = createStore({ ...initialState }, parentStore);
+const mergedChain = createMergedStoreChain();
 
 for (let i = 0; i < 200; i += 1) {
   subscribe(selectedSubscribersStore, ["count"], (state) => {
@@ -76,13 +182,20 @@ for (let i = 0; i < 200; i += 1) {
   subscribe(unrelatedSubscribersStore, ["activeId"], (state) => {
     consume(state.activeId);
   });
+  subscribe(reactSubscribersStore, null, () => {
+    consume(counter);
+  });
 }
 
 batch(batchStore, ["count", "value"], (state) => {
   consume(state.count);
 });
 
+setupBenchmarkDerivedState(cascadingStore);
+
 init(syncedStore);
+init(cascadingStore);
+init(mergedChain.store);
 
 bench(
   "create store",
@@ -142,11 +255,58 @@ bench(
 );
 
 bench(
-  "set synced parent state",
+  "set synced child state",
   async () => {
     syncedStore.setState("count", nextValue());
     await flushBatch();
     consume(parentStore.getState().count);
+  },
+  options,
+);
+
+bench(
+  "set synced source state",
+  async () => {
+    parentStore.setState("count", nextValue());
+    await flushBatch();
+    consume(syncedStore.getState().count);
+  },
+  options,
+);
+
+bench(
+  "set state with React subscribers",
+  async () => {
+    reactSubscribersStore.setState(
+      "activeId",
+      itemIds[nextValue() % itemIds.length] ?? null,
+    );
+    await flushBatch();
+  },
+  options,
+);
+
+bench(
+  "set unrelated state with React subscribers",
+  async () => {
+    reactSubscribersStore.setState("count", nextValue());
+    await flushBatch();
+  },
+  options,
+);
+
+bench(
+  "cascade setup listeners",
+  async () => {
+    const value = nextValue();
+    cascadingStore.setState("open", value % 2 === 0);
+    cascadingStore.setState(
+      "activeId",
+      itemIds[value % itemIds.length] ?? null,
+    );
+    cascadingStore.setState("moves", value);
+    await flushBatch();
+    consume(cascadingStore.getState().activeValue);
   },
   options,
 );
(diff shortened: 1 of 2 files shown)


packages/ariakit-store/benchmark/store.bench.ts
```diff
@@ -57,6 +94,72 @@ function createBenchmarkStore(count = 0) {
...
+    sync(store, ["moves", "activeId"], (state, prevState) => {
+      if (state.moves !== prevState.moves) return;
+      store.setState("activeValue", undefined);
+    }),
+  );
+
+  setup(store, () =>
+    batch(store, ["moves", "renderedItems"], (state, prevState) => {
+      if (state.moves === prevState.moves) return;
+      const activeItem = getItem(state.activeId);
```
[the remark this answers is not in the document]
diegohaz: That's true, we should include the benchmark folder in the tsconfig.

packages/ariakit-store/benchmark/store.bench.ts
```diff
@@ -181,4 +352,50 @@ bench(
...
+      itemIds[value % itemIds.length] ?? null,
+    );
+    mergedChain.store.setState("moves", value);
+    await flushBatch();
+    consume(mergedChain.store.getState().activeValue);
+  },
+  options,
+);
+
 export { sink };
```
diegohaz, line 392: Are all 18 benchmark tests here really necessary, or are some of them redundant and could be deduplicated or merged with others?

packages/ariakit-store/benchmark/store.bench.ts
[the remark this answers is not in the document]
diegohaz: Worth flagging that adding `benchmark/` to the tsconfig will surface more errors than just this line. I ran `tsc` against the file with the same compiler options as `tsconfig.base.json` and got:

```
benchmark/store.bench.ts(121,40): error TS2339: Property 'activeId' does not exist on type 'Pick<...>'.
benchmark/store.bench.ts(158,30): error TS2345: Argument of type 'Store<{ count, activeId, value, rtl }>' is not assignable to parameter of type 'Store<BenchmarkState>'.
benchmark/store.bench.ts(380,42): error TS2339: Property 'open' does not exist on type '{ count, activeId, value, rtl }'.
benchmark/store.bench.ts(389,32): error TS2345: Argument of type '"open"' is not assignable to parameter of type '"value" | "count" | "activeId" | "rtl"'.
benchmark/store.bench.ts(394,32): error TS2345: Argument of type '"moves"' is not assignable to parameter of type '"value" | "count" | "activeId" | "rtl"'.
benchmark/store.bench.ts(396,42): error TS2339: Property 'activeValue' does not exist on type '{ count, activeId, value, rtl }'.
```

Lines 158/380/389/394/396 all stem from `mergeStore` inferring `S` from the intersection of the three argument states (`externalStore`, `pick(tagStore, ["value", "rtl"])`, `omit(popoverStore, [...])`), which narrows to `{ count, activeId, value, rtl }`. At runtime the merged store actually has all keys (mergeStore's `Object.assign` loop), but TypeScript can't see that.

For this line specifically, switching to `store.getState().activeId` (matches the pattern in `combobox-store.ts` lines 191-196) fixes the type error without changing keys.

For the merged-chain errors, a typed assertion on `mergedStore` (or on the returned `store`) is the lowest-noise fix:

```ts
const store = createStore(
  {
    ...mergedStore.getState(),
    count,
    activeId: itemIds[0] ?? null,
  } as BenchmarkState,
  compositeStore,
  mergedStore,
);
```

Probably worth handling in this PR rather than deferring, since the tsconfig change won't land cleanly without these fixes anyway.

packages/ariakit-store/benchmark/store.bench.ts
diegohaz: Quick observations from a deduplication pass:

**Strongest redundancy candidate:** `set state with React subscribers` (sets `activeId`) and `set unrelated state with React subscribers` (sets `count`) measure essentially the same code path. Null-keyed subscribers (`subscribe(store, null, ...)`) fire on every `setState` regardless of which key changed — the `keys.some(updated)` filter in `createStore`'s `run` is short-circuited by `!keys`. The current perf-results agree: 121,925 vs 122,377 ops/sec (~0.4% delta, statistical noise).

If the intent is to make the point "React-style subscribers can't optimize by key" (mirroring the `selected/unrelated` keyed-subscriber pair), you could keep both and rename for clarity (e.g., `set changed key with React subscribers` / `set other key with React subscribers`). Otherwise, dropping one is fine.

**Borderline — but I'd keep:** the four `set merged ...` benchmarks. Each does test a different propagation depth/path (source: 3-level, picked: 4-level via `pick`, omitted: 4-level via `omit`, derived: 4-level + setup cascades). The perf numbers reflect that (~250k / 242k / 65k / 14k).

One caveat there: `set merged omitted source state` toggles `open`, which trips the `setup` cascade (`open` → `mounted` derivation). That means it's measuring "omit propagation + cascade" rather than just "omit propagation," so it's not directly comparable against `set merged picked source state` (which sets `value`, no cascade). If you want the omit/pick comparison to isolate path overhead, switching the omit benchmark to a non-derived key like `count` would help. If you intentionally want it to measure realistic popover-store usage, the current key is fine — worth a one-line comment.

**Safe to keep, distinct:** everything else exercises a different combination of listener count, listener type, sync direction, or merge composition.

packages/ariakit-store/benchmark/store.bench.ts
```diff
@@ -57,6 +94,72 @@ function createBenchmarkStore(count = 0) {
...
+      store.setState("activeValue", activeItem?.value);
+    }),
+  );
+}
+
+function createMergedStoreChain(count = 0) {
+  const externalStore = createBenchmarkStore(count);
+  const tagStore = createBenchmarkStore(count);
+  const popoverStore = createBenchmarkStore(count);
+  const compositeStore = createBenchmarkStore(count);
```
diegohaz, line 134: `compositeStore` is created here and passed to `createStore` on line 154 as a synced source, but it isn't returned (line 160 returns `{ externalStore, popoverStore, store, tagStore }`) and no benchmark ever writes to or reads from it. It contributes to `init`/`destroy` overhead in `initialize merged store chain` and adds one extra propagation target to every top-level `mergedChain.store.setState`, without exercising any meaningful behavior the other stores don't already cover.

If this is meant to model a real "composite layered under popover" topology (mirroring `createComboboxStore`'s composite + popover layout), this stops being free filler and starts being load-bearing — in which case it's worth a one-line comment so the intent is preserved against future cleanup. Otherwise, removing it would simplify the chain without changing what the benchmarks actually measure.

```suggestion
function createMergedStoreChain(count = 0) {
  const externalStore = createBenchmarkStore(count);
  const tagStore = createBenchmarkStore(count);
  const popoverStore = createBenchmarkStore(count);

  const mergedStore = mergeStore(
    externalStore,
    pick(tagStore, ["value", "rtl"]),
    omit(popoverStore, [
      "arrowElement",
      "anchorElement",
      "baseElement",
      "contentElement",
      "disclosureElement",
      "items",
      "popoverElement",
      "renderedItems",
    ]),
  );

  const store = createStore(
    {
      ...mergedStore.getState(),
      count,
      activeId: itemIds[0] ?? null,
    },
    mergedStore,
  );
```

diegohaz (commented):
Reviewed the expansion of `packages/ariakit-store/benchmark/store.bench.ts`. The added benchmarks mirror real Ariakit derivation patterns (the `sync`+`batch` `activeValue` shape matches `combobox-store.ts:181-197`, and the `mounted` derivation matches `disclosure-store.ts:65-66`), and the perf-bot run shows no regressions on existing benchmarks.

**Scope:** initial review against `main` at `8f32ce5829648d82c7782c73a7d619348ad9a1b8`; PR head `6544624c0f30d7f94f4ededa31eb519a419c55f3`. Reviewed the diff plus surrounding `@ariakit/store` internals (`createStore`, `mergeStore`, `pick`, `omit`, `init`'s `desyncs` chain), the real combobox/popover/disclosure stores, the workspace `tsconfig.json` (which currently includes only `src`, not `benchmark`), and `vitest.bench.config.ts`.

**Validation performed:**
- Ran `tsc --noEmit` with `tsconfig.base.json`-equivalent flags against `benchmark/store.bench.ts` to enumerate the type errors that would surface if `benchmark/` is added to the tsconfig (six errors across lines 121, 158, 380, 389, 394, 396 — all stem from two root causes; details in the line-121 thread).
- Confirmed runtime behavior is fine: the module's top-level setup (creating `mergedChain`, registering 200 null-keyed subscribers, calling `init`) executes without error when loaded outside vitest — the type errors are purely TS inference, since `mergeStore`'s `Object.assign` loop copies every key at runtime.
- Traced propagation depths for the merged-chain benchmarks against `createStore`'s `setState`/`init` implementation (`packages/ariakit-store/src/index.ts:80-211`) to verify the perf-results are explained by chain depth + cascade triggers.

**Main findings (all posted as inline threads):**
1. The Copilot-flagged TS error on line 121 is the surface of a larger issue — there are 6 type errors total, all of which the planned "include benchmark in tsconfig" follow-up will have to address. Suggested a low-noise fix (matching the real combobox's `store.getState().activeId` pattern + a typed initial-state assertion on the merged chain's outer `createStore`). Probably worth doing in this PR so the tsconfig change can land cleanly. (Reply on the existing thread at line 121.)
2. `compositeStore` in `createMergedStoreChain` is created and wired in as a synced source but never returned, written, or read by any benchmark — pure overhead unless the intent is to model the composite-under-popover topology, in which case it deserves a one-line comment. (Inline at line 131.)
3. Re: the author's own deduplication question — `set state with React subscribers` and `set unrelated state with React subscribers` measure the same code path, since null-keyed subscribers fire on every `setState` regardless of which key changed (perf-results agree to within ~0.4%). Also noted that `set merged omitted source state` conflates omit-path overhead with the `mounted` derivation cascade because it sets `open`; using `count` (or a comment explaining the intent) would make the omit vs. pick comparison cleaner. (Reply on the existing thread at line 401.)

**Not flagged:**
- No changeset — correct, this is benchmark-only and not user-facing per `ariakit-general-workflow` guidance.
- The `set synced parent state` → `set synced child state` rename + new `set synced source state` are accurate per the `createStore(initialState, ...stores)` "stores to extend" terminology; the consumer/extending store is the child, the extended store is the source.
- Code style is consistent with `ariakit-general-code-style` (early returns, brace usage, JSDoc not required for benchmark internals, `function` for top-level + arrow for nested setup callbacks).

**Did not validate:** I didn't run the full benchmark in CI mode (no baseline available locally for a comparison), so I relied on the perf-bot's published numbers to confirm direction. The interpretation of those numbers in the merged-chain comment is from tracing `setState`/`init` chains, not from instrumented measurement.

packages/ariakit-store/benchmark/store.bench.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 1e88bfb61 - included the benchmark folder in the package tsconfig, read activeId from the full store state in the batch listener, and typed the merged-chain initial state as BenchmarkState without an unsafe assertion.

packages/ariakit-store/benchmark/store.bench.ts
diegohaz: Fixed in 1e88bfb61 - kept the direct composite source and added a comment explaining that it models component stores that extend composite state alongside merged external state.

packages/ariakit-store/benchmark/store.bench.ts
diegohaz: Fixed in 1e88bfb61 - removed the redundant React null-subscriber benchmark and changed the omitted-source benchmark to update focused, with a comment documenting that it isolates omit propagation instead of also triggering the open/mounted cascade.

diegohaz:
**Fixed**
- 1e88bfb61 ([reply](https://github.com/ariakit/ariakit/pull/6057#discussion_r3307770748)): Included `benchmark` in the package tsconfig and fixed the batch listener/merged-chain typing issues without unsafe assertions.
- 1e88bfb61 ([reply](https://github.com/ariakit/ariakit/pull/6057#discussion_r3307770750)): Removed the redundant React null-subscriber benchmark and isolated the omitted-source benchmark on a non-derived key.
- 1e88bfb61 ([reply](https://github.com/ariakit/ariakit/pull/6057#discussion_r3307770749)): Documented why `compositeStore` remains in the merged-chain fixture.

diegohaz (commented):
Re-reviewed `1e88bfb61` against prior head `6544624c0`. All three findings from the earlier review are addressed; nothing new to flag.

**Scope:** incremental diff from `6544624c0..1e88bfb61` covering `packages/ariakit-store/tsconfig.json` and `packages/ariakit-store/benchmark/store.bench.ts`, plus a fresh `tsc` run over the whole package and the full project `pnpm tsc`.

**Verifications performed:**
- Ran `pnpm exec tsc -p packages/ariakit-store/tsconfig.json --noEmit` and `pnpm tsc` — both clean. With `benchmark` now in the package's `include`, all six TS errors I enumerated previously are gone. The batch listener uses `const { activeId } = store.getState()` (matching `combobox-store.ts:191-196`), and the merged-chain `createStore` receives a `state: BenchmarkState`-typed object built by spreading `initialState` underneath `mergedStore.getState()` — cleaner than the cast I'd sketched, and avoids any `as` escape hatch.
- Traced the renamed omit benchmark: `focused` isn't in the omit list and isn't a dependency of any `setupBenchmarkDerivedState` listener, so this now isolates omit propagation without tripping the `mounted` cascade. The `nextValue() % 2 === 0` toggling guarantees the value actually changes each call.
- Confirmed the React-subscriber benchmark is now a single case (the old `activeId`-setting version was dropped; the count-setting version retained the name).
- The new comment on `compositeStore` accurately describes the modeled topology (component stores like combobox extend a composite store plus the merged external state — see `combobox-store.ts:198`).
- CI checks at re-review time: `Type check`, `Lint`, `Test`, `Test Chrome`, `Test Firefox`, `Build packages`, `Build Next.js app`, `Build styles` all green; `Test Chrome (perf)` and `Node` benchmark jobs still running, which is expected.

**Not validated:** the new perf-bot comparison hadn't published when I re-reviewed, so I couldn't confirm that the omit benchmark now lands near the picked benchmark (~242k ops/sec) as expected after dropping the `open` cascade. That'll show up in the next ariakito comment.

LGTM.

PR #6061, opened 2026-05-27
Improve store update performance

diegohaz:
## Motivation

The `@ariakit/store` benchmark suite ([added recently](https://github.com/ariakit/ariakit/pull/6054), expanded in [#6057](https://github.com/ariakit/ariakit/pull/6057), with a first round of perf work in [#6058](https://github.com/ariakit/ariakit/pull/6058)) made several remaining hot paths easy to measure. Listener dispatch was paying for a WeakMap lookup and a `void` setter per invocation; every `setState` queued a microtask even when no batch listener was registered.

## Solution

Independent optimizations that work together to land a wide speedup. None of them change observable behavior for code that follows the documented store contract.

### `setState` hot path

- Inline `applyState`: drop the function-call indirection (and the unused lazy `currentValue` branch — store state never holds lazy values).
- Cache `state[key]` once into `currentValue` to avoid the double property access.
- Guard the `for (const store of stores)` loop with `stores.length` so isolated stores skip an iterator allocation.

### Listener dispatch (`runListeners`)

- Switch `disposables` from `WeakMap<Listener, void | (() => void)>` to `Map<Listener, () => void>`, populated only when a listener actually returns a cleanup. Skips a WeakMap `set` per invocation in the (very common) cleanup-less case.
- Inline the former `run` helper into `runListeners`. (Extracting it as a helper regressed React subscribers ~30%.)
- Slow-path optimization: use `allKeysListeners.has(listener)` to skip the keys lookup for all-keys listeners while still iterating `group.listeners` so re-keyed listeners are visited via their updated `listenerKeys`.

### Batch microtask

- Replace the `lastUpdate` counter with a `batchPending` boolean — exactly one microtask is scheduled per batch period instead of one per `setState`.
- Skip the entire batch bookkeeping (microtask, `updatedKeys`, `prevStateBatch` updates) when no batch listener is registered. `setState` still keeps `prevStateBatch = state` current at the end of every outermost setState so a future batch listener — registered idle or mid-dispatch — observes the correct baseline.
- Track `inDispatch` so a batch listener registered during a setState observes the same `(state, prevState)` diff a HEAD-style always-queued microtask would have delivered.
- The microtask anchors `prevStateBatch` to the pre-flush snapshot only when no in-flight refresh happened (compare `prevStateBatch === prevStateBatchBefore`), so a listener that unsubscribes itself or registers a successor during the flush doesn't lose the correct baseline.

### Correctness anchors

Targeted regression tests added for each correctness anchor:

- `re-registers a listener without dragging a stale cleanup forward`
- `registers a batch listener during dispatch and still sees the in-flight diff`
- `keeps the batch baseline current across idle setStates between subscriptions`
- `keeps the batch baseline current when a batch listener unsubscribes itself and setStates mid-flush`
- `keeps the batch baseline current when the only batch listener unsubscribes itself mid-flush`
- `keeps the batch baseline current when a batch listener registers a successor mid-flush`
- `syncs reentrant parent setStates during a child's initial push`
- `syncs reentrant parent setStates to an earlier key during a child's initial push`
- `fires a sync listener registered on the parent after init with the child already updated`
- `fires a re-keyed listener for the currently dispatched key`
- `unsubscribes a keyed listener from inside another keyed listener`

## Benchmarks

All numbers from `pnpm bench`, two confirmed runs each, Node 24 on a macOS machine. Baseline is `90c640b4` (the commit before this PR).

| Benchmark | Baseline (hz) | This PR (hz) | Speedup |
|---|---:|---:|---:|
| set state | 3,587,318 | 5,760,829 | **+61%** |
| set state with selected subscribers (200 keyed) | 333,062 | 1,680,896 | **~5x** |
| set state with React subscribers (200 all-keys) | 207,780 | 556,745 | **~2.7x** |
| set state with unrelated subscribers | 3,503,297 | 5,949,854 | **+70%** |
| batch multiple updates | 1,981,191 | 2,599,113 | **+31%** |
| set synced child state | 1,517,463 | 2,395,067 | **+58%** |
| set synced source state | 1,917,324 | 3,121,099 | **+63%** |
| cascade setup listeners | 755,981 | 1,084,697 | **+43%** |
| create picked store | 687,164 | 741,729 | +8% |
| initialize merged store chain | 21,403 | 25,386 | +19% |
| set merged source state | 730,846 | 896,836 | **+23%** |
| set merged picked source state | 611,937 | 786,706 | **+29%** |
| set merged omitted source state | 521,406 | 636,809 | **+22%** |
| set merged derived state | 39,652 | 48,522 | **+22%** |

No regressions; `create store` and `read state` are within noise.

## Test plan

- [x] `pnpm test` — 846 tests pass (existing + 11 new regression tests in `packages/ariakit-store/src/test.ts`).
- [x] `pnpm bench` — confirmed numbers above.
- [x] `pnpm tsc -b --noEmit packages/ariakit-store` — clean.
- [x] `pnpm lint` — clean.

### DIFF
--- a/.changeset/201-store-update-performance.md
+++ b/.changeset/201-store-update-performance.md
@@ -1,5 +1,0 @@
----
-"@ariakit/store": patch
----
-
-Improved store update performance with a focus on listener dispatch, microtask batching, and merged store initialization.
--- a/.changeset/201-store-update-performance.md
+++ b/.changeset/201-store-update-performance.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/store": patch
+---
+
+Improved store update performance with a focus on listener dispatch, microtask batching, and merged store initialization.
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -323,23 +323,6 @@ export function createStore<S extends State>(
       return;
     }
     const allKeysListeners = group.allKeysListeners;
-    if (allKeysListeners && allKeysListeners.size === group.listeners.size) {
-      // Fast path: every listener is an all-keys listener, no keys to check.
-      // Listeners re-keyed to a non-null key during this iteration are dropped
-      // from allKeysListeners and therefore skipped for the current update;
-      // they fire normally on subsequent setStates.
-      for (const listener of allKeysListeners) {
-        const cleanup = disposables.size && disposables.get(listener);
-        if (cleanup) cleanup();
-        const result = listener(state, prevState);
-        if (result) {
-          disposables.set(listener, result);
-        } else if (cleanup) {
-          disposables.delete(listener);
-        }
-      }
-      return;
-    }
     for (const listener of group.listeners) {
       if (!allKeysListeners?.has(listener)) {
         const keys = group.listenerKeys.get(listener);
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -1,7 +1,6 @@
 import {
   omit as _omit,
   pick as _pick,
-  applyState,
   chain,
   getKeys,
   hasOwnProperty,
@@ -15,6 +14,12 @@ type Listener<S> = (state: S, prevState: S) => void | (() => void);
 type Sync<S, K extends keyof S> = (
   keys: K[] | null,
   listener: Listener<Pick<S, K>>,
+) => () => void;
+
+type ParentSyncListener<S> = (state: S, updatedKey: keyof S | null) => void;
+type ParentSync<S> = (
+  keys: Array<keyof S>,
+  listener: ParentSyncListener<S>,
 ) => () => void;
 
 type StoreSetup = (callback: () => void | (() => void)) => () => void;
@@ -31,13 +36,14 @@ type StoreOmit<
   K extends ReadonlyArray<keyof S> = ReadonlyArray<keyof S>,
 > = (keys: K) => Store<Omit<S, K[number]>>;
 type ListenerMap<S> = Map<keyof S, Set<Listener<S>>>;
+type ParentSyncMap<S> = Map<keyof S, Set<ParentSyncListener<S>>>;
 type UpdatedKey<S> = keyof S | Set<keyof S>;
 
 interface ListenerGroup<S> {
   listeners: Set<Listener<S>>;
   listenersByKey?: ListenerMap<S>;
   allKeysListeners?: Set<Listener<S>>;
-  disposables: WeakMap<Listener<S>, void | (() => void)>;
+  disposables: Map<Listener<S>, () => void>;
   listenerKeys: WeakMap<Listener<S>, Array<keyof S> | null>;
 }
 
@@ -49,6 +55,7 @@ interface StoreInternals<S = State> {
   batch: StoreBatch<S>;
   pick: StorePick<S>;
   omit: StoreOmit<S>;
+  parentSync: ParentSync<S>;
 }
 
 function getInternal<K extends keyof StoreInternals>(
@@ -119,22 +126,27 @@ export function createStore<S extends State>(
 ): Store<S> {
   let state = initialState;
   let prevStateBatch = state;
-  let lastUpdate = 0;
   let destroy = noop;
+  let batchPending = false;
+  let inDispatch = false;
+  let updatedKeys = new Set<keyof S>();
   const instances = new Set<symbol>();
-  const updatedKeys = new Set<keyof S>();
 
   const setups = new Set<() => void | (() => void)>();
   const syncListeners: ListenerGroup<S> = {
     listeners: new Set(),
-    disposables: new WeakMap(),
+    disposables: new Map(),
     listenerKeys: new WeakMap(),
   };
   const batchListenerGroup: ListenerGroup<S> = {
     listeners: new Set(),
-    disposables: new WeakMap(),
+    disposables: new Map(),
     listenerKeys: new WeakMap(),
   };
+  // Per-key fan-out for child stores that extend this one. Stays undefined
+  // until a child store registers a parent-sync listener so isolated stores
+  // pay no extra cost on every setState.
+  let parentSyncByKey: ParentSyncMap<S> | undefined;
 
   const storeSetup: StoreSetup = (callback) => {
     setups.add(callback);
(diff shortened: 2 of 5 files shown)


diegohaz:
@codex

.changeset/201-store-update-performance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/store": patch
+---
+
+Improved store update performance with a focus on listener dispatch, microtask batching, and merged store initialization.
```
diegohaz: Just update the existing changeset (as if it were just one piece of work) instead of creating a new one. Also add a changeset for `@ariakit/react` since it might affect that package as well.

packages/ariakit-store/src/index.ts
```diff
@@ -323,24 +419,71 @@ export function createStore<S extends State>(
...
+    try {
+      runListeners(syncListeners, prevState, key);
+
+      // Direct fan-out to child stores extending this one. Each parentSync
+      // listener receives the changed key and forwards the value to the
+      // child's setState in O(1).
+      const parentSyncs = parentSyncByKey?.get(key);
+      if (parentSyncs) {
+        for (const listener of parentSyncs) {
+          listener(state, key);
```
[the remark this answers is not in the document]
diegohaz: Confirmed locally against `origin/main` and this head. With `init(child); sync(parent, ["count"], () => { calls.push(child.getState().count); }); parent.setState("count", 1);`, `origin/main` records `[0, 1]`, while this PR records `[0, 0]` even though the child is updated to `1` afterward.

The regression is that parent-to-child forwarding is no longer part of the parent's normal sync listener order. Child init used to register the forwarding listener in the same `syncListeners` set, so listeners registered after `init(child)` observed the child already synced. Running `parentSyncByKey` only after `runListeners(syncListeners, ...)` changes that ordering for extended, picked, and merged stores. Please keep parent-sync dispatch in registration order with normal sync listeners, or otherwise preserve the old ordering; simply moving this block before all normal listeners would flip the behavior for listeners registered before child init.

packages/ariakit-store/src/index.ts
```diff
@@ -281,40 +343,74 @@ export function createStore<S extends State>(
...
+      }
+      return;
+    }
+    const allKeysListeners = group.allKeysListeners;
+    if (allKeysListeners && allKeysListeners.size === group.listeners.size) {
+      // Fast path: every listener is an all-keys listener, no keys to check.
+      // Listeners re-keyed to a non-null key during this iteration are dropped
+      // from allKeysListeners and therefore skipped for the current update;
+      // they fire normally on subsequent setStates.
+      for (const listener of allKeysListeners) {
```
diegohaz: This fast path changes dispatch semantics when listener registrations are mutated during a dispatch. I validated a small repro against `origin/main` and this head:

```ts
const second = () => events.push("second");
subscribe(store, null, () => {
  events.push("first");
  subscribe(store, ["count"], second);
});
subscribe(store, null, second);
store.setState("count", 1);
```

On `origin/main`, `second` still runs for this update because dispatch iterates `group.listeners` and checks the listener's current keys when it reaches it. On this PR, the first listener removes `second` from `allKeysListeners`, so the `for (const listener of allKeysListeners)` iterator never reaches it. That can drop a listener that was re-keyed to the key currently being updated. Please route this case through the generic `group.listeners` path, or otherwise re-check the current `listenerKeys` when the all-keys set changes during dispatch, so the optimization preserves existing re-registration semantics.

diegohaz (commented):
Reviewed as an initial full review for PR #6061 at `0e253770ee67f96276ae8e1c76833b1b12c10ecd` against merge base `90c640b46807b7b0214951e07ac68fd0b43958bd` / `origin/main`. I read the PR description, diff, `@ariakit/store` internals, surrounding tests, benchmark context, existing review threads, and CI status.

Posted 2 actionable findings:

- Confirmed the existing parent-sync ordering thread: child propagation now runs after later parent sync listeners.
- Added a new all-keys fast-path finding: re-keying a listener during dispatch can skip a listener that used to run for the current update.

Validation performed:

- `pnpm install`
- `pnpm test packages/ariakit-store/src/test.ts --run`
- `pnpm exec tsc -b --noEmit packages/ariakit-store`
- `gh pr checks 6061 --repo ariakit/ariakit`
- Node probes comparing `origin/main` with this head for the two dispatch-order regressions

I did not rerun `pnpm bench`; I relied on the PR/CI benchmark results for throughput and focused local validation on semantics.

packages/ariakit-store/src/index.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 4c20ab9 — reverted the parentSync internal API and went back to the original sync()-based registrations in storeInit, so parent listeners registered after init(child) once again observe the child already updated. Regression test added in packages/ariakit-store/src/test.ts.

packages/ariakit-store/src/index.ts
diegohaz: Fixed in 769c9271 — removed the fast all-keys path that iterated allKeysListeners directly. The remaining slow path iterates group.listeners and uses allKeysListeners.has(listener) only as a fast-skip for the key check, so a listener re-keyed to the dispatched key during iteration is still visited via its updated listenerKeys. Regression test added with the exact repro you posted.

.changeset/201-store-update-performance.md
diegohaz: Fixed in 7b645b9 — updated the existing .changeset/200-store-performance.md to cover the combined work and added .changeset/200-react-store-performance.md for @ariakit/react. The 201- entry was removed.

diegohaz:
**Fixed**

- 4c20ab9 ([reply](https://github.com/ariakit/ariakit/pull/6061#discussion_r3312233470)): reverted the parentSync internal API so parent-to-child propagation goes back to the original sync()-based registrations in storeInit. A sync listener registered on the parent after init(child) once again observes the child already updated. New regression test `fires a sync listener registered on the parent after init with the child already updated`.
- 769c9271 ([reply](https://github.com/ariakit/ariakit/pull/6061#discussion_r3312235682)): removed the fast all-keys path. The slow path with `allKeysListeners?.has(listener)` correctly visits listeners re-keyed to the dispatched key during iteration. New regression test `fires a re-keyed listener for the currently dispatched key` using the exact repro posted in the review.
- 7b645b9 ([reply](https://github.com/ariakit/ariakit/pull/6061#discussion_r3312239395)): consolidated the changeset into the existing `200-store-performance.md` and added `200-react-store-performance.md` for `@ariakit/react`. The `201-` entry was removed.

After the reverts, the bench summary moved slightly (`set state with React subscribers` is now ~2.7× vs the previous ~8×; `initialize merged store chain` is ~+19% vs the previous ~2×). All other gains are largely preserved.

diegohaz (commented):
Re-reviewed PR #6061 at `7b645b93c3e0401a584de608f492587a66e4a239`, using the previous marked review head `0e253770ee67f96276ae8e1c76833b1b12c10ecd` as the incremental baseline.

Scope reviewed:

- Commits `4c20ab9`, `769c927`, and `7b645b9`
- Incremental diff in `packages/ariakit-store/src/index.ts`, `packages/ariakit-store/src/test.ts`, and `.changeset/*`
- Prior review threads and replies for parent propagation ordering, all-keys listener dispatch, and changeset consolidation
- Current PR description, changesets, and CI status

Result: no new substantive findings. I verified the two prior semantic issues are fixed in the current code: parent listeners registered after `init(child)` now observe the child already updated again, and the re-keyed listener repro now fires the listener for the currently dispatched key. The changeset consolidation also matches the prior feedback.

Validation performed:

- `pnpm test packages/ariakit-store/src/test.ts --run`
- `pnpm exec tsc -b --noEmit packages/ariakit-store`
- Targeted Node probes for the two prior dispatch-order regressions
- `gh pr checks 6061 --repo ariakit/ariakit`

At the time of this re-review, the PR checks were passing except the `Node` job was still pending, so I did not treat CI as fully complete.

PR #6065, opened 2026-05-28
Fix Dialog unmount animation without explicit store

### LINKED ISSUE #3403: unmountOnHide doesn't respect closing animation when Dialog is used without explicit store
### Current behavior

Closing a dialog that has `open`, `onClose` and `unmountOnHide` props (see example below) doesn't wait for the closing animation before unmounting. When using `store` and `unmountOnHide` props, everything works as expected.

1. Open https://stackblitz.com/edit/gntorz?file=dialog-animated%2Fstyle.css,dialog-animated%2Findex.tsx&theme=dark
2. Click the first button and then dismiss the dialog - it animates correctly
3. Click the second button and then dismiss the dialog - there's no closing animation

### Expected behavior

Dialogs should finish animation regardless of whether store is passed as a prop or is created internally.

### Workaround

Construct `useStore` manually.

### Possible solutions

_No response_

diegohaz:
## Motivation

Fixes #3403.

A controlled `Dialog` using `open`, `onClose`, and `unmountOnHide` currently unmounts immediately when it closes if the consumer does not pass an explicit dialog store. That skips the leave transition, while the same animated dialog works when controlled through `useDialogStore`.

## Solution

This adds a sandbox regression for the reported behavior in `app/src/sandbox/dialog-3403`. The sandbox compares the explicit-store dialog with the controlled no-explicit-store dialog and verifies both stay mounted with `data-leave` during the closing animation before unmounting. It also includes a route-shaped controlled dialog without `unmountOnHide` to cover the unchanged default close behavior.

The `Dialog` component now creates and passes an internal dialog store through the same mounted-state gate used by explicit stores only when `unmountOnHide` is enabled and no `store` prop or `DialogProvider` context is available. This lets `unmountOnHide` observe `mounted` while the leave animation is running instead of falling back to the controlled `open` prop after it has already changed to `false`, while preserving the previous path for dialogs that do not use `unmountOnHide`.

## Workaround

Create the dialog store manually with the controlled `open` state and pass it to `Dialog` instead of passing `open` and `onClose` directly to `Dialog`:

```tsx
const [open, setOpen] = useState(false);
// TODO: Remove this explicit store workaround once
// https://github.com/ariakit/ariakit/issues/3403 is fixed.
const dialog = Ariakit.useDialogStore({
  open,
  setOpen,
});

<Ariakit.Button onClick={() => setOpen(true)}>Show dialog</Ariakit.Button>
<Ariakit.Dialog store={dialog} unmountOnHide>
  {/* ... */}
</Ariakit.Dialog>
```

### DIFF
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -628,13 +628,30 @@ export function createDialogComponent<T extends DialogOptions>(
  * </Dialog>
  * ```
  */
-export const Dialog = createDialogComponent(
-  forwardRef(function Dialog(props: DialogProps) {
-    const htmlProps = useDialog(props);
-    return createElement(TagName, htmlProps);
-  }),
+const DialogImpl = forwardRef(function DialogImpl(props: DialogProps) {
+  const htmlProps = useDialog(props);
+  return createElement(TagName, htmlProps);
+});
+
+const DialogWithStore = createDialogComponent(
+  DialogImpl,
   useDialogProviderContext,
 );
+
+const DialogWithInternalStore = forwardRef(function DialogWithInternalStore(
+  props: DialogProps,
+) {
+  const store = useDialogStore({ open: props.open });
+  return <DialogWithStore {...props} store={store} />;
+});
+
+export const Dialog = forwardRef(function Dialog(props: DialogProps) {
+  const context = useDialogProviderContext();
+  if (props.store || context) {
+    return <DialogWithStore {...props} />;
+  }
+  return <DialogWithInternalStore {...props} />;
+});
 
 export interface DialogOptions<T extends ElementType = TagName>
   extends FocusableOptions<T>, PortalOptions<T>, DisclosureContentOptions<T> {
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -647,7 +647,8 @@ const DialogWithInternalStore = forwardRef(function DialogWithInternalStore(
 
 export const Dialog = forwardRef(function Dialog(props: DialogProps) {
   const context = useDialogProviderContext();
-  if (props.store || context) {
+  // The hoisted store is only needed for the unmountOnHide mounted-state gate.
+  if (props.store || context || !props.unmountOnHide) {
     return <DialogWithStore {...props} />;
   }
   return <DialogWithInternalStore {...props} />;
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -608,6 +608,23 @@ export function createDialogComponent<T extends DialogOptions>(
   });
 }
 
+const DialogImpl = forwardRef(function DialogImpl(props: DialogProps) {
+  const htmlProps = useDialog(props);
+  return createElement(TagName, htmlProps);
+});
+
+const DialogWithStore = createDialogComponent(
+  DialogImpl,
+  useDialogProviderContext,
+);
+
+const DialogWithInternalStore = forwardRef(function DialogWithInternalStore(
+  props: DialogProps,
+) {
+  const store = useDialogStore({ open: props.open });
+  return <DialogWithStore {...props} store={store} />;
+});
+
 /**
  * Renders a dialog similar to the native `dialog` element that's rendered in a
  * [`portal`](https://ariakit.com/reference/dialog#portal) by default.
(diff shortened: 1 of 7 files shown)


packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -628,14 +628,31 @@ export function createDialogComponent<T extends DialogOptions>(
...
 );
 
+const DialogWithInternalStore = forwardRef(function DialogWithInternalStore(
+  props: DialogProps,
+) {
+  const store = useDialogStore({ open: props.open });
+  return <DialogWithStore {...props} store={store} />;
+});
+
+export const Dialog = forwardRef(function Dialog(props: DialogProps) {
```
diegohaz, line 648: The JSDoc block that documents the public `Dialog` component now sits above `DialogImpl`, the internal helper. Hovering `<Ariakit.Dialog>` in an editor or auto-extracting reference docs from this source will pick up nothing because the export below has no docstring of its own.

Move the block down so it lands immediately above `export const Dialog = forwardRef(...)`. Suggested order:

```tsx
const DialogImpl = forwardRef(function DialogImpl(props: DialogProps) {
  const htmlProps = useDialog(props);
  return createElement(TagName, htmlProps);
});

const DialogWithStore = createDialogComponent(
  DialogImpl,
  useDialogProviderContext,
);

const DialogWithInternalStore = forwardRef(function DialogWithInternalStore(
  props: DialogProps,
) {
  const store = useDialogStore({ open: props.open });
  return <DialogWithStore {...props} store={store} />;
});

/**
 * Renders a dialog similar to the native `dialog` element that's rendered in a
 * [`portal`](https://ariakit.com/reference/dialog#portal) by default.
 *
 * The dialog can be either
 * [`modal`](https://ariakit.com/reference/dialog#modal) or non-modal. The
 * visibility state can be controlled with the
 * [`open`](https://ariakit.com/reference/dialog#open) and
 * [`onClose`](https://ariakit.com/reference/dialog#onclose) props.
 * @see https://ariakit.com/components/dialog
 * @example
 * ```jsx {4-6}
 * const [open, setOpen] = useState(false);
 *
 * <button onClick={() => setOpen(true)}>Open dialog</button>
 * <Dialog open={open} onClose={() => setOpen(false)}>
 *   Dialog
 * </Dialog>
 * ```
 */
export const Dialog = forwardRef(function Dialog(props: DialogProps) {
  const context = useDialogProviderContext();
  if (props.store || context) {
    return <DialogWithStore {...props} />;
  }
  return <DialogWithInternalStore {...props} />;
});
```

The sibling components in this directory (e.g. `DialogDisclosure`) keep the docstring right above the exported `forwardRef`, so this mirrors the established pattern.

diegohaz (commented):
## Review summary

**Type:** initial review against `b992669` (compared to `origin/main` @ `1d99543`).

**Scope reviewed:**
- Library change in `packages/ariakit-react-components/src/dialog/dialog.tsx`.
- New sandbox under `app/src/sandbox/dialog-3403/` and its `test-browser.ts`.
- Changeset `300-dialog-unmount-animation.md`.
- Surrounding context: `useDialog`, `createDialogComponent`, `useDialogStore`, `useDisclosureContent`, `mergeStore`/`createStore` propagation in `@ariakit/store`.

**What I validated:**
- Traced the unmount path through both stores: when no external store or provider is present, `DialogWithInternalStore` creates an outer store whose `mounted` stays true while `animating` is true, so `createDialogComponent`'s `useStoreState(store, state => …)` correctly observes the leave transition. The inner store created inside `useDialog` propagates `setState("animating", false)` back up to that outer store via `createStore`'s upstream propagation (the `if (!fromStores && stores.length)` branch in `setState`), so the wrapper eventually re-renders with `state.mounted === false` and unmounts. The control flow is correct.
- `pnpm tsc` is clean.
- Confirmed all `Test Chrome/Firefox/Safari/macOS/Android/React 18` checks pass on CI; the only failing job (`Test Plus`) is an unrelated Stripe/Clerk billing test and not caused by this change.
- Compared sibling components: `useMenu`/`usePopover`/`useTooltip` all guard with `invariant(store, …)`, so they cannot be used controlled without an explicit store or provider. The bug is therefore Dialog-specific and the fix's narrow scope is appropriate; there is no follow-up needed for the other `createDialogComponent` consumers.
- Considered alternative shapes: (1) always wrapping with an internal store (extra store layer in the common case where a store is already provided, slightly less efficient than the conditional render); (2) creating the fallback store inside `createDialogComponent` itself (would change the behavior of unrelated consumers and always pay the store-creation cost). The PR's conditional render — `Dialog` reading context once and switching between `DialogWithStore` and `DialogWithInternalStore` — is the best trade-off here.

**Sandbox / test coverage:** the new sandbox covers both the explicit-store and the controlled-without-store paths, asserts the dialog stays attached and gets `data-leave="true"` during the leave animation, then asserts it eventually detaches. The query uses `includeHidden: true` correctly because `disableTree` sets `inert` on the dialog during the leave phase. Test design is solid.

**Changeset:** filename, prefix `300-`, frontmatter (`@ariakit/react-components` + `@ariakit/react`), and link-style content follow the changeset skill. `@ariakit/react` is correctly included because it re-exports `Dialog` from `@ariakit/react-components/dialog/dialog` in `packages/ariakit-react/src/dialog.ts:5`.

**Findings:** one actionable, posted inline:
- JSDoc misplacement on `Dialog` — the public component's docstring now sits above the private `DialogImpl` helper instead of above `export const Dialog`. Hover and reference-doc extraction will no longer pick it up. Fix is to move the block down. See inline comment for the suggested ordering.

No other substantive correctness, performance, security, or consistency issues found.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in d01fe413f - moved the public Dialog JSDoc block back above `export const Dialog` so editor hover/reference extraction attaches it to the exported component.

diegohaz:
**Fixed**

- d01fe413f ([thread](https://github.com/ariakit/ariakit/pull/6065#discussion_r3318779153)): moved the public `Dialog` JSDoc back above the exported component and resolved the review thread.

PR #6066, opened 2026-05-28
Fix Button form status on keyboard submit

### LINKED ISSUE #4540: Button doesn't work with useFormStatus on keyboard click
### Current behavior

The Ariakit Button doesn't work as expected when combined with [`useFormStatus`](https://react.dev/reference/react-dom/hooks/useFormStatus).

1. Open https://stackblitz.com/edit/giwgk7h1-5wysplkc?file=button%2Findex.tsx
2. Focus on `AriakitButton`
3. Press <kbd>Enter</kbd>
4. Notice that the label doesn't change to `Pending` (bug)
5. Focus on `AriakitButton (focusable)`
6. Press <kbd>Enter</kbd> and <kbd>Tab</kbd>
7. Notice how the label reverts when the button loses focus (bug)

### Expected behavior

Follow the reproduction steps on the `NativeButton` buttons to see the expected behavior.

### Workaround

Passing `onFocusVisible={(event) => event.preventDefault()}` to the Ariakit Button prevents the issue.

### Possible solutions

The internal focus-visible handling logic might conflict with the new form features. We should create an internal example using `useFormStatus` and test it.

diegohaz:
## Motivation

Ariakit [`Button`](https://ariakit.com/reference/button) does not work correctly with React's `useFormStatus` when the form is submitted from the keyboard. Pressing Enter on the button does not consistently expose the pending state, and an `accessibleWhenDisabled` button can lose the pending label after focus moves away.

Fixes #4540.

## Solution

Added a React sandbox that reproduces the keyboard submission behavior with `useFormStatus` and covers both the regular submit button and the `accessibleWhenDisabled` case.

`Focusable` now preserves the imperative `data-focus-visible` behavior for native submit controls without committing a React focus-visible state update while React form status is pending. It keeps a ref-backed focus-visible flag so repeat keydown handling and hidden/disabled cleanup still behave like the state-backed path, and it removes the sandbox workaround so the tests exercise the library fix directly.

## Workaround

Preventing Ariakit's focus-visible handler from updating internal focus-visible state lets React keep the form submission pending state when the button is submitted from the keyboard.

```tsx
<Button
  type="submit"
  disabled={pending}
  // TODO: Remove once https://github.com/ariakit/ariakit/issues/4540 is fixed.
  onFocusVisible={(event) => event.preventDefault()}
>
  {pending ? "Pending" : children}
</Button>
```

### DIFF
--- a/app/src/sandbox/button-form-status-4540/index.react.tsx
+++ b/app/src/sandbox/button-form-status-4540/index.react.tsx
@@ -72,7 +72,7 @@ export default function Example() {
         <AriakitButton>AriakitButton</AriakitButton>
       </Form>
       <Form label="AriakitButton focusable">
-        <AriakitButton accessibleWhenDisabled>
+        <AriakitButton accessibleWhenDisabled onClick={countClicks}>
           AriakitButton (focusable)
         </AriakitButton>
       </Form>
--- a/app/src/sandbox/button-form-status-4540/index.react.tsx
+++ b/app/src/sandbox/button-form-status-4540/index.react.tsx
@@ -5,12 +5,7 @@ import { useFormStatus } from "react-dom";
 function AriakitButton(props: Ariakit.ButtonProps) {
   const { pending } = useFormStatus();
   return (
-    <Ariakit.Button
-      type="submit"
-      disabled={pending}
-      onFocusVisible={(event) => event.preventDefault()}
-      {...props}
-    >
+    <Ariakit.Button type="submit" disabled={pending} {...props}>
       {pending ? "Pending" : props.children}
     </Ariakit.Button>
   );
--- a/app/src/sandbox/button-form-status-4540/index.react.tsx
+++ b/app/src/sandbox/button-form-status-4540/index.react.tsx
@@ -1,0 +1,69 @@
+import * as Ariakit from "@ariakit/react";
+import type { ComponentProps } from "react";
+import { useFormStatus } from "react-dom";
+
+function AriakitButton(props: Ariakit.ButtonProps) {
+  const { pending } = useFormStatus();
+  return (
+    <Ariakit.Button type="submit" disabled={pending} {...props}>
+      {pending ? "Pending" : props.children}
+    </Ariakit.Button>
+  );
+}
+
+interface NativeButtonProps extends ComponentProps<"button"> {
+  accessibleWhenDisabled?: boolean;
+}
+
+function NativeButton({ accessibleWhenDisabled, ...props }: NativeButtonProps) {
+  const { pending } = useFormStatus();
+  if (accessibleWhenDisabled) {
+    props["aria-disabled"] = pending;
+  } else {
+    props.disabled = pending;
+  }
+  return (
+    <button type="submit" {...props}>
+      {pending ? "Pending" : props.children}
+    </button>
+  );
+}
+
+interface FormProps extends ComponentProps<"form"> {
+  label: string;
+}
+
+function Form({ label, ...props }: FormProps) {
+  return (
+    <form
+      aria-label={label}
+      action={async () => {
+        await new Promise((resolve) => setTimeout(resolve, 3000));
+      }}
+      {...props}
+    />
+  );
+}
+
+export default function Example() {
+  return (
+    <div style={{ display: "flex", flexDirection: "column", gap: 16 }}>
+      <Form label="AriakitButton">
+        <AriakitButton>AriakitButton</AriakitButton>
+      </Form>
+      <Form label="AriakitButton focusable">
+        <AriakitButton accessibleWhenDisabled>
+          AriakitButton (focusable)
+        </AriakitButton>
+      </Form>
+      <Form label="NativeButton">
+        <NativeButton>NativeButton</NativeButton>
+      </Form>
+      <Form label="NativeButton focusable">
+        <NativeButton accessibleWhenDisabled>
+          NativeButton (focusable)
+        </NativeButton>
+      </Form>
+    </div>
+  );
+}
--- a/app/src/sandbox/button-form-status-4540/index.react.tsx
+++ b/app/src/sandbox/button-form-status-4540/index.react.tsx
@@ -5,7 +5,12 @@ import { useFormStatus } from "react-dom";
 function AriakitButton(props: Ariakit.ButtonProps) {
   const { pending } = useFormStatus();
   return (
-    <Ariakit.Button type="submit" disabled={pending} {...props}>
+    <Ariakit.Button
+      type="submit"
+      disabled={pending}
+      onFocusVisible={(event) => event.preventDefault()}
+      {...props}
+    >
       {pending ? "Pending" : props.children}
     </Ariakit.Button>
   );
--- a/app/src/sandbox/button-form-status-4540/index.react.tsx
+++ b/app/src/sandbox/button-form-status-4540/index.react.tsx
@@ -1,5 +1,5 @@
 import * as Ariakit from "@ariakit/react";
-import type { ComponentProps } from "react";
+import type { ComponentProps, MouseEvent } from "react";
 import { useFormStatus } from "react-dom";
 
 function AriakitButton(props: Ariakit.ButtonProps) {
@@ -15,7 +15,11 @@ interface NativeButtonProps extends ComponentProps<"button"> {
   accessibleWhenDisabled?: boolean;
 }
 
-function NativeButton({ accessibleWhenDisabled, ...props }: NativeButtonProps) {
+function NativeButton({
+  accessibleWhenDisabled,
+  onClick,
+  ...props
+}: NativeButtonProps) {
   const { pending } = useFormStatus();
   if (accessibleWhenDisabled) {
     props["aria-disabled"] = pending;
@@ -23,7 +27,17 @@ function NativeButton({ accessibleWhenDisabled, ...props }: NativeButtonProps) {
     props.disabled = pending;
   }
   return (
-    <button type="submit" {...props}>
+    <button
+      type="submit"
+      {...props}
+      onClick={(event) => {
+        if (accessibleWhenDisabled && pending) {
+          event.preventDefault();
+          return;
+        }
+        onClick?.(event);
+      }}
+    >
       {pending ? "Pending" : props.children}
     </button>
   );
@@ -45,6 +59,12 @@ function Form({ label, ...props }: FormProps) {
   );
 }
 
+function countClicks(event: MouseEvent<HTMLButtonElement>) {
+  const { currentTarget } = event;
+  const clickCount = Number(currentTarget.dataset.clickCount ?? 0) + 1;
+  currentTarget.dataset.clickCount = String(clickCount);
+}
+
 export default function Example() {
   return (
     <div style={{ display: "flex", flexDirection: "column", gap: 16 }}>
@@ -60,7 +80,7 @@ export default function Example() {
         <NativeButton>NativeButton</NativeButton>
       </Form>
       <Form label="NativeButton focusable">
-        <NativeButton accessibleWhenDisabled>
+        <NativeButton accessibleWhenDisabled onClick={countClicks}>
           NativeButton (focusable)
         </NativeButton>
       </Form>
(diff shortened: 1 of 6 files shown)


app/src/sandbox/button-form-status-4540/index.react.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+function NativeButton({ accessibleWhenDisabled, ...props }: NativeButtonProps) {
+  const { pending } = useFormStatus();
+  if (accessibleWhenDisabled) {
+    props["aria-disabled"] = pending;
+  } else {
+    props.disabled = pending;
+  }
+  return (
+    <button type="submit" {...props}>
```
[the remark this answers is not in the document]
diegohaz: Fixed in a020e56f4 - the native focusable sandbox button now prevents activation while pending, and the browser test covers repeated Enter activation.

diegohaz:
**Fixed**

- a020e56f4 ([reply](https://github.com/ariakit/ariakit/pull/6066#discussion_r3318791262)): prevented repeated activation for the focusable native sandbox button while React form status is pending, and added browser coverage for repeated Enter activation.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -225,13 +237,21 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
     // here by checking the element's disabled attribute.
     useEffect(() => {
       if (!focusable) return;
-      if (trulyDisabled && focusVisible) {
+      if (!trulyDisabled) return;
+      const element = ref.current;
+      nativeSubmitObserverCleanupRef.current?.();
+      nativeSubmitObserverCleanupRef.current = null;
+      focusVisibleRef.current = false;
+      element?.removeAttribute("data-focus-visible");
```
diegohaz: The 4-line cleanup sequence used here — disconnect the native-submit observer, null the cleanup ref, clear `focusVisibleRef.current`, and remove the `data-focus-visible` attribute — is repeated almost verbatim in three places: this `trulyDisabled` effect (lines 250–253), the `IntersectionObserver` callback inside `handleFocusVisible` (lines 323–326), and `onBlur` (lines 383–386).

Consider extracting it into a small helper to keep the cleanup invariant in one place. A future change (e.g., adding another cleanup step) would otherwise have to be made in all three sites.

```ts
const cleanupFocusVisible = (element: HTMLElement | null) => {
  nativeSubmitObserverCleanupRef.current?.();
  nativeSubmitObserverCleanupRef.current = null;
  focusVisibleRef.current = false;
  element?.removeAttribute("data-focus-visible");
};
```

The three sites would then reduce to:

- `trulyDisabled` effect: `cleanupFocusVisible(ref.current); if (focusVisible) setFocusVisible(false);`
- Observer callback: `if (isFocusable(element)) return; cleanupFocusVisible(element);`
- `onBlur`: `cleanupFocusVisible(event.currentTarget); setFocusVisible(false);`

Aligns with the shared code-style guidance about extracting repeated logic, and makes the cleanup contract explicit in one spot.

app/src/sandbox/button-form-status-4540/test-browser.ts
```diff
@@ -0,0 +1,49 @@
...
+    await button.focus();
+    await page.keyboard.press("Enter");
+    await test.expect(button).toHaveText("Pending");
+    await test.expect(button).toHaveAttribute("data-click-count", "1");
+
+    await page.keyboard.press("Enter");
+
+    await test.expect(button).toHaveAttribute("data-click-count", "1");
+  });
+});
```
diegohaz, line 66: This test sets a useful reference baseline for what an `accessibleWhenDisabled` submit button should do when Enter is pressed twice during pending — only one form submission. Consider adding the symmetric assertion for the `AriakitButton focusable` form so the library side is also exercised.

Ariakit currently prevents the second submission via `Focusable`'s `useDisableEvent`-driven `onClickCapture` (which calls `preventDefault`/`stopPropagation` when the focusable is disabled). That path is unrelated to the focus-visible change in this PR, but a regression test would guard against future refactors of `useFocusable` / `Command` accidentally weakening it, and it would round out the parity between the native and Ariakit reference forms in the sandbox.

It would just need wiring an `onClick={countClicks}` (or equivalent) into the `AriakitButton` shape and asserting `data-click-count` stays at `"1"` after a second Enter press.

Not a blocker — flagging because the test naming and the two halves of the sandbox suggest this is the intended symmetry, and it's cheap to add.

diegohaz (commented):
Initial review against `a020e56f4b474e7ad72b06450ebf722167d71f5b`, baseline `origin/main` (`1d9954353`). Read the diff and surrounding context: `useFocusable`, `useButton`, `useCommand`, `useDisableEvent`, the existing focus-visible IntersectionObserver effect, the consumers that read `data-focus-visible` (`tooltip.tsx`, `hovercard-disclosure.tsx`), and the linked issue #4540. Ran `tsc` against `@ariakit/react-components` — passes. Did not run the new browser test locally; reasoned about the timing instead.

## Scope reviewed

- `packages/ariakit-react-components/src/focusable/focusable.tsx` — adds `isNativeSubmitControl`, a `focusVisibleRef` mirror of the state, and a parallel imperative cleanup path (DOM attribute + ref + observer) for native submit controls so `setFocusVisible(true/false)` is never called for them. Updates the `trulyDisabled` effect, the existing IntersectionObserver effect, `onKeyDownCapture`, and `onBlur` to keep the ref in sync. Adds an unmount cleanup effect.
- `app/src/sandbox/button-form-status-4540/` — new sandbox with both `AriakitButton` and `NativeButton` variants (with and without `accessibleWhenDisabled`) inside a form whose `action` resolves after 3s. Folder name + structure follow the convention used by `tab-form-5273` and similar sandbox-for-issue paths.
- `app/src/sandbox/button-form-status-4540/test-browser.ts` — three browser tests covering (1) pending state appears on Enter, (2) pending state survives Tab on the `accessibleWhenDisabled` variant, and (3) the sandbox's native button doesn't double-activate during pending.
- `.changeset/300-button-form-status.md` — patch entry for `@ariakit/react-components` + `@ariakit/react`, single past-tense sentence matching the changeset skill.
- Latest commit `a020e56f4` adds `preventDefault` on the sandbox's `NativeButton` click for `accessibleWhenDisabled + pending` and the corresponding test, establishing reference baseline behavior.

## Bug analysis confirmed

Verified the mechanism: when the form action is pending, the `trulyDisabled` effect or the `onBlur` handler would call `setFocusVisible(false)` while the action is mid-flight. That state update appears to interfere with React 19's `useFormStatus` tracking, dropping the pending state. The fix is correct: for native submit controls it sets `data-focus-visible` imperatively and skips the `useState` update entirely, so no state churn happens during the pending window. The state-based path (non-submit `Focusable` elements) is unchanged.

Also confirmed:
- `data-focus-visible` set via `element.dataset.focusVisible = "true"` is not clobbered on subsequent React renders because the JSX value is `undefined` and stripped by `removeUndefinedValues` before reconciliation — React never tracks the attribute, so the imperative value persists.
- `tooltip.tsx`'s `"focusVisible" in anchorElement.dataset` and `hovercard-disclosure.tsx`'s `MutationObserver({ attributeFilter: ["data-focus-visible"] })` both still work since they read DOM state, not React state.
- For `<button>` without an explicit `type`, `HTMLButtonElement.type` returns `"submit"` by default, so `isNativeSubmitControl` correctly covers the default-type case as well as `type="submit"` and `<input type="submit"|"image">`.
- The inline `IntersectionObserver` in `handleFocusVisible` (lines 320–330) duplicates the shape of the existing observer effect, but is necessary: the existing one is gated on `focusVisible` state being `true`, which never happens for native submit controls. Keeping both is the symmetric choice — fine as-is.

## Findings

Two non-blocking inline suggestions:

1. Extract the 4-line "cleanup native-submit focus-visible state" sequence that's repeated in three places (focusable.tsx).
2. Add a symmetric repeat-submission assertion for `AriakitButton focusable` to mirror the new native-button test (test-browser.ts).

No correctness, security, or backwards-compatibility issues found. The fix is targeted and minimal for the bug it addresses.

## Not validated

- Did not run the new browser test locally — the worktree is on Node 24 but the app's `check` script picks up a port-bound inspector from another running dev server in this environment.
- Did not independently verify React 19's internal reason for losing the pending state on intervening `setState`; reasoned from observed behavior described in the issue and the targeted fix.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in d8e77c2e0 - extracted the shared focus-visible cleanup into a single helper used by the disabled, observer, and blur cleanup paths.

app/src/sandbox/button-form-status-4540/test-browser.ts
diegohaz: Fixed in d8e77c2e0 - added the symmetric Ariakit focusable repeated-Enter assertion and wired the click counter into that sandbox case.

diegohaz:
**Fixed**

- d8e77c2e0 ([cleanup reply](https://github.com/ariakit/ariakit/pull/6066#discussion_r3318934696)): extracted the shared focus-visible cleanup helper.
- d8e77c2e0 ([test reply](https://github.com/ariakit/ariakit/pull/6066#discussion_r3318934892)): added symmetric repeated-Enter coverage for the Ariakit focusable submit button.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Resolved in d8e77c2e0. `cleanupFocusVisible` is now extracted as a `useEvent`-wrapped helper at focusable.tsx:242 and the three sites (`trulyDisabled` effect, the inline `IntersectionObserver` callback in `handleFocusVisible`, and `onBlur`) all delegate to it. `useEvent` gives the helper a stable identity, matching the existing pattern used for `onKeyDownCapture` / `onFocusCapture` / `onBlur` / `autoFocusRef` in this file, so adding it to the `trulyDisabled` effect's deps doesn't cause extra runs. Behavior is preserved.

app/src/sandbox/button-form-status-4540/test-browser.ts
diegohaz: Resolved in d8e77c2e0. New test `focusable button prevents activation while pending` at test-browser.ts:33-48 mirrors the native-button test against the Ariakit `AriakitButton focusable` form, with `onClick={countClicks}` wired into the sandbox. The assertion that `data-click-count` stays at `"1"` after the second Enter exercises Focusable's `useDisableEvent`-driven `onClickCapture` `preventDefault`/`stopPropagation` path while the form is pending. Verified the test setup and counter mechanics line up with the native equivalent.

diegohaz (commented):
Re-review of the incremental changes since my prior review at `a020e56f4`, against new head `d8e77c2e0`. Diff scope: `app/src/sandbox/button-form-status-4540/index.react.tsx`, `app/src/sandbox/button-form-status-4540/test-browser.ts`, `packages/ariakit-react-components/src/focusable/focusable.tsx`. Ran `tsc` on `@ariakit/react-components` — passes.

## Prior threads

Both of my previous findings were addressed by `d8e77c2e0` ("Address Button form status review comments"):

- `focusable.tsx` cleanup helper: `cleanupFocusVisible` extracted as a `useEvent`-wrapped helper and called from all three sites. Confirmed identical behavior and noted that `useEvent` matches the file's existing convention for stable callbacks. Replied on the thread.
- `test-browser.ts` Ariakit-side repeat-submission test: added as `focusable button prevents activation while pending`, with `onClick={countClicks}` now wired into the `AriakitButton focusable` form. Replied on the thread.

## New review

Nothing new in the incremental diff to flag. Spot-checked that:

- The dependency change at `focusable.tsx:259` (`cleanupFocusVisible` added to the `trulyDisabled` effect deps) is harmless because the helper has a stable identity via `useEvent`.
- The observer-callback usage at `focusable.tsx:325` and the `onBlur` usage at `focusable.tsx:382` correctly pass a non-null `HTMLElement`; the helper's optional chaining still does the right thing for the effect site where `ref.current` may be `null`.
- The new test reuses the `"AriakitButton focusable"` form already covered by `focusable button keeps pending state after losing focus`; tests don't share state since `beforeEach` reloads the preview.

No correctness, security, or backwards-compat concerns. From my side this looks ready.

PR #6067, opened 2026-05-28
Fix tooltip controlled open crash

### LINKED ISSUE #4894: total window crash on multiple tooltips with prop `open`
### Current behavior

It seems like my browser is crashing when I have more than one tooltip opened at the same time.

Works
```
<>
      <Ariakit.TooltipProvider>
        <Ariakit.TooltipAnchor>one</Ariakit.TooltipAnchor>
        <Ariakit.Tooltip>HELLO!</Ariakit.Tooltip>
      </Ariakit.TooltipProvider>
      <Ariakit.TooltipProvider>
        <Ariakit.TooltipAnchor>two</Ariakit.TooltipAnchor>
        <Ariakit.Tooltip>HELLO 222!</Ariakit.Tooltip>
      </Ariakit.TooltipProvider>
    </>
```

Crashes the window
```
<>
      <Ariakit.TooltipProvider open>
        <Ariakit.TooltipAnchor>one</Ariakit.TooltipAnchor>
        <Ariakit.Tooltip>HELLO!</Ariakit.Tooltip>
      </Ariakit.TooltipProvider>
      <Ariakit.TooltipProvider open>
        <Ariakit.TooltipAnchor>two</Ariakit.TooltipAnchor>
        <Ariakit.Tooltip>HELLO 222!</Ariakit.Tooltip>
      </Ariakit.TooltipProvider>
    </>
```

1. See: https://stackblitz.com/edit/ariakit-ptxwqlja?file=src%2FApp.tsx

### Expected behavior

Well, it should not crash, instead just have both tooltips open at the same time.

### Workaround

None found :(

### Possible solutions

_No response_

diegohaz:
## Motivation

Issue #4894 reports that rendering multiple forced-open tooltips can freeze or crash the page. The tooltip global active-store logic tries to hide the previously mounted tooltip, but `open={true}` immediately forces it open again, which can create a re-entrant loop.

Fixes #4894

## Solution

This adds a `tooltip-4894` sandbox with both browser coverage and a Vitest test that also runs under the React 18 test matrix.

The sandbox covers three cases:

- Multiple tooltips forced open with `open={true}` stay visible without looping.
- Controlled tooltips with `open` and `setOpen` still participate in the default one-open-tooltip behavior when they accept `setOpen(false)`.
- Managed controlled tooltips still close each other even after forced-open tooltips reopen.

The library fix keeps the normal active-tooltip `hide()` path. When a store that Ariakit just tried to hide is forced open again in the same turn, Ariakit leaves the current active store in place instead of issuing a reciprocal `hide()` that would restart the loop.

## Workaround

Until the library fix is released, avoid forcing multiple tooltip providers open at the same time. Instead, keep one active tooltip in state and derive each provider's `open` prop from that state:

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/4894 is fixed.
const [activeTooltip, setActiveTooltip] = useState<"one" | "two" | null>(null);

<TooltipProvider open={activeTooltip === "one"}>
  <TooltipAnchor onClick={() => setActiveTooltip("one")}>one</TooltipAnchor>
  <Tooltip>HELLO!</Tooltip>
</TooltipProvider>
<TooltipProvider open={activeTooltip === "two"}>
  <TooltipAnchor onClick={() => setActiveTooltip("two")}>two</TooltipAnchor>
  <Tooltip>HELLO 222!</Tooltip>
</TooltipProvider>
```

This workaround trades simultaneous visible tooltips for a stable one-at-a-time controlled tooltip flow.

### DIFF
--- a/.changeset/300-tooltip-controlled-open.md
+++ b/.changeset/300-tooltip-controlled-open.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid a re-entrant loop when multiple tooltips are forced open at the same time while preserving the default one-open-tooltip behavior for controlled tooltips that accept close updates.
+Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid a re-entrant loop when multiple tooltips are forced open at the same time.
--- a/.changeset/300-tooltip-controlled-open.md
+++ b/.changeset/300-tooltip-controlled-open.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid closing other controlled open tooltips when multiple tooltips are mounted at the same time.
--- a/.changeset/300-tooltip-controlled-open.md
+++ b/.changeset/300-tooltip-controlled-open.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid closing other controlled open tooltips when multiple tooltips are mounted at the same time.
+Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid a re-entrant loop when multiple tooltips are forced open at the same time while preserving the default one-open-tooltip behavior for controlled tooltips that accept close updates.
--- a/app/src/sandbox/tooltip-4894/test.ts
+++ b/app/src/sandbox/tooltip-4894/test.ts
@@ -1,9 +1,34 @@
-import { q } from "@ariakit/test";
+import { click, q } from "@ariakit/test";
 import { expect, test } from "vitest";
 
 // See https://github.com/ariakit/ariakit/issues/4894
-test("does not hide multiple controlled open tooltips", () => {
-  expect(q.tooltip("HELLO!")).toBeVisible();
-  expect(q.tooltip("HELLO 222!")).toBeVisible();
-  expect(q.text("Close requests: 0")).toBeVisible();
+test("does not loop when multiple tooltips are forced open", async () => {
+  await click(q.button("Show forced tooltips"));
+  expect(q.tooltip("FORCED ONE")).toBeVisible();
+  expect(q.tooltip("FORCED TWO")).toBeVisible();
+  expect(q.tooltip("FORCED THREE")).toBeVisible();
+  expect(q.text(/Forced close requests: [1-9]\d*/)).toBeVisible();
+});
+
+test("hides controlled tooltips that accept setOpen updates", async () => {
+  await click(q.button("Open managed one"));
+  expect(q.tooltip("MANAGED ONE")).toBeVisible();
+  expect(q.tooltip("MANAGED TWO")).not.toBeInTheDocument();
+
+  await click(q.button("Open managed two"));
+  expect(q.tooltip("MANAGED ONE")).not.toBeInTheDocument();
+  expect(q.tooltip("MANAGED TWO")).toBeVisible();
+});
+
+test("keeps managed tooltips active after forced tooltips reopen", async () => {
+  await click(q.button("Show forced tooltips"));
+  await click(q.button("Open managed one"));
+  expect(q.tooltip("MANAGED ONE")).toBeVisible();
+
+  await click(q.button("Open managed two"));
+  expect(q.tooltip("MANAGED ONE")).not.toBeInTheDocument();
+  expect(q.tooltip("MANAGED TWO")).toBeVisible();
+  expect(q.tooltip("FORCED ONE")).toBeVisible();
+  expect(q.tooltip("FORCED TWO")).toBeVisible();
+  expect(q.tooltip("FORCED THREE")).toBeVisible();
 });
--- a/app/src/sandbox/tooltip-4894/test.ts
+++ b/app/src/sandbox/tooltip-4894/test.ts
@@ -1,13 +1,9 @@
-import { click, q } from "@ariakit/test";
+import { q } from "@ariakit/test";
 import { expect, test } from "vitest";
 
 // See https://github.com/ariakit/ariakit/issues/4894
-test("works around the crash by opening one controlled tooltip at a time", async () => {
-  await click(q.button("Open first tooltip"));
+test("does not hide multiple controlled open tooltips", () => {
   expect(q.tooltip("HELLO!")).toBeVisible();
-  expect(q.tooltip("HELLO 222!")).not.toBeInTheDocument();
-
-  await click(q.button("Open second tooltip"));
-  expect(q.tooltip("HELLO!")).not.toBeInTheDocument();
   expect(q.tooltip("HELLO 222!")).toBeVisible();
+  expect(q.text("Close requests: 0")).toBeVisible();
 });
--- a/app/src/sandbox/tooltip-4894/test.ts
+++ b/app/src/sandbox/tooltip-4894/test.ts
@@ -1,0 +1,9 @@
+import { q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+// See https://github.com/ariakit/ariakit/issues/4894
+test("does not hide multiple controlled open tooltips", () => {
+  expect(q.tooltip("HELLO!")).toBeVisible();
+  expect(q.tooltip("HELLO 222!")).toBeVisible();
+  expect(q.text("Close requests: 0")).toBeVisible();
+});
--- a/app/src/sandbox/tooltip-4894/test.ts
+++ b/app/src/sandbox/tooltip-4894/test.ts
@@ -1,9 +1,13 @@
-import { q } from "@ariakit/test";
+import { click, q } from "@ariakit/test";
 import { expect, test } from "vitest";
 
 // See https://github.com/ariakit/ariakit/issues/4894
-test("does not hide multiple controlled open tooltips", () => {
+test("works around the crash by opening one controlled tooltip at a time", async () => {
+  await click(q.button("Open first tooltip"));
   expect(q.tooltip("HELLO!")).toBeVisible();
+  expect(q.tooltip("HELLO 222!")).not.toBeInTheDocument();
+
+  await click(q.button("Open second tooltip"));
+  expect(q.tooltip("HELLO!")).not.toBeInTheDocument();
   expect(q.tooltip("HELLO 222!")).toBeVisible();
-  expect(q.text("Close requests: 0")).toBeVisible();
 });
--- a/app/src/sandbox/tooltip-4894/test.ts
+++ b/app/src/sandbox/tooltip-4894/test.ts
@@ -2,7 +2,10 @@ import { click, q } from "@ariakit/test";
 import { expect, test } from "vitest";
 
 // See https://github.com/ariakit/ariakit/issues/4894
-test("does not loop when multiple tooltips are forced open", async () => {
+// In jsdom, this documents the stable forced-open state. The mixed test below
+// fails without the re-entrant loop guard, and test-browser.ts covers the pure
+// forced-open flow in a real browser.
+test("keeps multiple forced tooltips visible", async () => {
   await click(q.button("Show forced tooltips"));
   expect(q.tooltip("FORCED ONE")).toBeVisible();
   expect(q.tooltip("FORCED TWO")).toBeVisible();
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -13,6 +13,7 @@ import { useEffect, useRef } from "react";
 import type { HovercardAnchorOptions } from "../hovercard/hovercard-anchor.tsx";
 import { useHovercardAnchor } from "../hovercard/hovercard-anchor.tsx";
 import { useTooltipProviderContext } from "./tooltip-context.tsx";
+import { isTooltipStoreOpenControlled } from "./tooltip-store.ts";
 import type { TooltipStore } from "./tooltip-store.ts";
 
 const TagName = "div" satisfies ElementType;
@@ -81,7 +82,10 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           // active one and set the current one as the active tooltip.
           if (state.mounted) {
             const { activeStore } = globalStore.getState();
-            if (activeStore !== store) {
+            if (
+              activeStore !== store &&
+              !isTooltipStoreOpenControlled(activeStore)
+            ) {
               activeStore?.hide();
             }
             return globalStore.setState("activeStore", store);
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -79,7 +79,9 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
         sync(store, ["mounted", "skipTimeout"], (state) => {
           if (!store) return;
           // If the current tooltip is open, we should immediately hide the
-          // active one and set the current one as the active tooltip.
+          // active one and set the current one as the active tooltip. We
+          // skip controlled open stores because useStoreProps would force
+          // them open again, causing a re-entrant loop between tooltips.
           if (state.mounted) {
             const { activeStore } = globalStore.getState();
             if (
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -39,6 +39,8 @@ function hideStore(store: TooltipStore | null) {
   if (!store) return;
   hidingStores.add(store);
   store.hide();
+  // Cleanup runs after `hide()` so controlled `open` props still see this
+  // store in `hidingStores` when their batch microtask forces them open again.
   queueMicrotask(() => hidingStores.delete(store));
 }
 
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -96,11 +96,8 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           if (state.mounted) {
             const { activeStore } = globalStore.getState();
             if (activeStore !== store) {
-              if (hidingStores.has(store)) {
-                return;
-              } else {
-                hideStore(activeStore);
-              }
+              if (hidingStores.has(store)) return;
+              hideStore(activeStore);
             }
             return globalStore.setState("activeStore", store);
           }
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -13,7 +13,6 @@ import { useEffect, useRef } from "react";
 import type { HovercardAnchorOptions } from "../hovercard/hovercard-anchor.tsx";
 import { useHovercardAnchor } from "../hovercard/hovercard-anchor.tsx";
 import { useTooltipProviderContext } from "./tooltip-context.tsx";
-import { isTooltipStoreOpenControlled } from "./tooltip-store.ts";
 import type { TooltipStore } from "./tooltip-store.ts";
 
 const TagName = "div" satisfies ElementType;
@@ -26,12 +25,21 @@ const globalStore = createStore<{ activeStore: TooltipStore | null }>({
   activeStore: null,
 });
 
+const hidingStores = new WeakSet<TooltipStore>();
+
 function createRemoveStoreCallback(store: TooltipStore) {
   return () => {
     const { activeStore } = globalStore.getState();
     if (activeStore !== store) return;
     globalStore.setState("activeStore", null);
   };
+}
+
+function hideStore(store: TooltipStore | null) {
+  if (!store) return;
+  hidingStores.add(store);
+  store.hide();
+  queueMicrotask(() => hidingStores.delete(store));
 }
 
 /**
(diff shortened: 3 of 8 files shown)


packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
```diff
@@ -81,7 +82,10 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           // active one and set the current one as the active tooltip.
```
diegohaz: The new `!isTooltipStoreOpenControlled(activeStore)` guard is correct, but the rationale lives entirely in the PR/issue — anyone reading this code in isolation lands on a check that quietly skips `hide()` for some active stores without a stated reason. They have to grep into `tooltip-store.ts`, find the WeakSet, and reconstruct the re-entrant loop (`hide()` → `useStoreProps` forces `open=true` → other tooltip mounts again → its `mounted` sync listener calls `hide()` back) to understand why this branch exists.

The codebase already inlines similarly non-obvious rationales (e.g., the `canShowOnHoverRef` block a few lines above), so it would be consistent — and per the codebase's "only comment when the WHY is non-obvious" rule, justified — to extend the existing comment here. Something like:

```suggestion
          // active one and set the current one as the active tooltip. We
          // skip `hide()` when the active store's `open` prop is controlled
          // because `useStoreProps` would force it open again, creating a
          // re-entrant loop between mounted tooltips (see #4894).
```

diegohaz (commented):
## Summary

Initial review of PR #6067 against `origin/main` (baseline `1d99543539feab457c8a03b7e49a6bd9a0ea18b9`, head `5953660ae55d18f3e5a46904ff7259c2437647a8`). Scope covered: the changeset, the new `tooltip-4894` sandbox (component, preview wrappers, browser + Vitest tests), and the library fix in `tooltip-anchor.tsx` and `tooltip-store.ts`. Surrounding context reviewed: the tooltip global active-store machinery, `useStoreProps`/`useStore` semantics in `@ariakit/react-store`, the disclosure/hovercard store cascade that drives `mounted` from `open`, and the existing tooltip sandboxes and example tests for convention comparison.

The fix is well-scoped and correct. The `controlledOpenStores` WeakSet is populated in a `useSafeLayoutEffect` registered before the open-related `useStoreProps` calls, so it is committed before any `mounted` sync listener in `tooltip-anchor.tsx` ever fires (layout effects run before `useEffect`, and across sibling providers the WeakSet is fully populated before the second tooltip's anchor effect registers). The guard in `tooltip-anchor.tsx` then skips the `hide()` call only when the active store's `open` prop is controlled, preserving the existing one-tooltip-at-a-time behavior for uncontrolled stores. I considered alternatives (mutating the store object with an `openControlled` flag, lifting the flag into the global store, deferring `hide()` to a microtask, removing the active-store hide entirely) and the WeakSet approach is the cleanest of the lot — it leaves the store API untouched, is O(1) per check, and doesn't change uncontrolled behavior.

## Validation performed

- `pnpm vitest run app/src/sandbox/tooltip-4894/test.ts` — passes on PR head.
- Reverted the fix files to `origin/main` while keeping the new test, and re-ran the same command — the test fails with `Close requests: 1`, confirming the test catches the regression (the `setOpen(false)` counter pattern is a nice trick; in StrictMode the loop is bounded to one iteration before unmount cleanup tears the sync listeners down, but a non-zero count is enough to detect any spurious close).
- `pnpm react18 test app/src/sandbox/tooltip-4894/test.ts` — passes on the React 18 matrix as well.
- `pnpm vitest run examples/menu-tooltip examples/menu-tooltip-various examples/tooltip examples/tooltip-label examples/tooltip-instant examples/tooltip-framer-motion` — all 50 existing tooltip tests pass.
- `pnpm tsc` — clean.

## Findings

- One inline comment on `tooltip-anchor.tsx:82` suggesting the existing comment be extended so the WHY of the new `!isTooltipStoreOpenControlled(activeStore)` guard is captured at the call site (currently a reader has to grep into `tooltip-store.ts` and reconstruct the re-entrant loop to understand the branch). This is consistent with the codebase's "only comment when the WHY is non-obvious" rule and with the existing inline rationale for `canShowOnHoverRef` a few lines above.

No correctness, regression, perf, compatibility, or test-coverage issues otherwise. The changeset format, sandbox layout, browser+integration test pair, public surface (no new export from `@ariakit/react`), and React 18 coverage all look right.

## Areas I did not fully validate

- I did not run `test-browser.ts` against the dev app preview (Playwright path). The Vitest/jsdom test already exercises the same StrictMode behavior, and the visible-text assertions match, so I expect this to be redundant in practice, but I didn't actually start `pnpm dev-app` to confirm.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
diegohaz: Fixed in 5de851304 - added an inline rationale explaining why controlled open stores skip hide() to avoid the re-entrant tooltip loop.

diegohaz:
**Fixed**

- 5de851304 ([reply](https://github.com/ariakit/ariakit/pull/6067#discussion_r3318997997)): Added an inline rationale for the controlled-open active-store guard so the re-entrant tooltip loop is explained at the branch that avoids it.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
diegohaz: Confirmed in 5de851304 — the extended comment captures the WHY at the call site and matches the surrounding rationale style. The test still passes on the new head. No further follow-up here.

diegohaz (commented):
## Summary

Re-review of PR #6067 at head `5de8513041c500a97def0d424a349e4b0278aa53`, incremental against the prior reviewed SHA `5953660ae55d18f3e5a46904ff7259c2437647a8`. One new commit (5de851304 "Explain controlled tooltip active store guard") modifying only the comment block on `packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx:81-84`. No other code changes since the last review.

The new commit addresses the single finding from the previous review by extending the inline rationale at the call site. I checked the new wording at lines 82-84 — it captures the WHY (`controlled open stores skipped because useStoreProps would force them open again, causing a re-entrant loop`) at the call site, matches the surrounding comment style, and is a fair abbreviation of the suggested wording (drops the explicit `#4894` reference, but the rest is preserved). Test still passes on the new head (`pnpm vitest run app/src/sandbox/tooltip-4894/test.ts` → 1/1). Acknowledged on the thread (https://github.com/ariakit/ariakit/pull/6067#discussion_r3319010478).

Reviewed the rest of the unchanged diff again for any newly exposed surrounding issues — nothing new turned up. The library fix, sandbox, browser + Vitest tests, and changeset remain in the state I approved in the prior review.

No new findings. The PR looks ready from my side.

.changeset/300-tooltip-controlled-open.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid a re-entrant loop when multiple tooltips are forced open at the same time while preserving the default one-open-tooltip behavior for controlled tooltips that accept close updates.
```
diegohaz: No need to mention that we're preserving behavior.

```suggestion
Fixed [`TooltipProvider`](https://ariakit.com/reference/tooltip-provider) to avoid a re-entrant loop when multiple tooltips are forced open at the same time.
```

diegohaz: Fixed in 1b1eaee98 - simplified the changeset wording to focus on the forced-open re-entrant loop.

diegohaz:
**Fixed**

- 1b1eaee98 ([reply](https://github.com/ariakit/ariakit/pull/6067#discussion_r3319330594)): Simplified the changeset wording to focus on the forced-open re-entrant loop.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
```diff
@@ -33,6 +35,13 @@ function createRemoveStoreCallback(store: TooltipStore) {
   };
 }
 
+function hideStore(store: TooltipStore | null) {
+  if (!store) return;
+  hidingStores.add(store);
+  store.hide();
+  queueMicrotask(() => hidingStores.delete(store));
+}
```
diegohaz, line 45: The `queueMicrotask(() => hidingStores.delete(store))` cleanup must be scheduled **after** `store.hide()`, not before — and this isn't obvious from reading the four lines in isolation. Worth a one-liner so a future refactor doesn't reorder them.

`store.hide()` calls `setState("open", false)`. For controlled stores, `useStoreProps`'s batch listener (registered in `packages/ariakit-react-store/src/index.tsx`) means that `setState` schedules its own microtask to re-sync the prop back to `true`. Microtasks fire FIFO, so scheduling order decides what happens on re-entry:

- **Current (cleanup after `hide`)**: the re-sync microtask fires first → triggers the mount sync at line 87 → `hidingStores.has(store)` is still `true` → early return. Then this cleanup runs and clears the flag. No loop.
- **Reversed (cleanup before `hide`)**: this cleanup microtask fires first → flag cleared → the re-sync microtask then fires → mount sync sees `hidingStores.has(store) === false` → the re-entrant loop restarts.

So the ordering inside this helper is what actually breaks the cycle, not the `hidingStores.has(store)` check on its own (which only short-circuits if the flag survives long enough). I'd recommend locking that in with a brief comment, e.g.:

```suggestion
function hideStore(store: TooltipStore | null) {
  if (!store) return;
  hidingStores.add(store);
  store.hide();
  // Cleanup runs after `hide()` so it fires after the batch microtask
  // `setState` queues to re-sync controlled `open` props (via
  // `useStoreProps`). The re-sync's mount sync then still sees the flag
  // and short-circuits, breaking the re-entrant loop.
  queueMicrotask(() => hidingStores.delete(store));
}
```

diegohaz (commented):
## Summary

Re-review of PR #6067 at head `1b1eaee982baa14c86b24945ac99df30500fc349`, incremental against the prior reviewed SHA `5de8513041c500a97def0d424a349e4b0278aa53`. Two new commits since then:

- `584bf0c5a` ("Preserve managed controlled tooltip behavior") — substantively rewrites the fix.
- `1b1eaee98` ("Tighten tooltip changeset wording") — applies your own earlier suggestion on the changeset; no code change.

### The library fix

The `584bf0c5a` commit reverts the previously-reviewed `controlledOpenStores` WeakSet in `tooltip-store.ts` and introduces a different mechanism entirely in `tooltip-anchor.tsx`: a module-level `hidingStores` `WeakSet` plus a `hideStore` helper that adds the store to the set, calls `store.hide()`, then schedules a `queueMicrotask` to clear it. The mount-sync handler then short-circuits when the just-mounted store is already in `hidingStores`.

**This is a substantive improvement over the previously-reviewed approach.** The old guard blanket-skipped `hide()` whenever the active store's `open` prop was controlled, which would have prevented managed controlled tooltips from closing each other when a new one opens — the case the new third test covers. The new approach still calls `hide()` on every active-store swap and only short-circuits on the re-entrant re-mount within the same microtask turn. That narrows the protection to the actual re-entrant case (controlled `open` whose `setOpen` won't accept `false`) while preserving one-active-tooltip behavior for controlled tooltips that do accept it.

**Why the mechanism works.** When `hideStore(active)` runs, `active.hide()` calls `setState("open", false)`, which (for controlled stores with a `useStoreProps` batch listener) schedules its own batch microtask to re-sync the prop back to `true`. `hideStore` then schedules its cleanup microtask. Microtasks fire FIFO, so the batch microtask runs first, re-syncs `state.open=true`, triggers the mount sync, and the mount sync sees `hidingStores.has(activeStore) === true` and returns without restarting the loop. The cleanup microtask then clears the flag. Managed tooltips that *do* accept `setOpen(false)` settle on the next React render because their prop value (re-read from parent state updated via the `setValue` sync listener) now matches the closed state.

I considered the alternatives — keeping the prior `controlledOpenStores` flag (fails the managed third test), lifting state into the global store, deferring hide via a deferred microtask/timeout (visual flicker), removing the active-store hide entirely (loses one-tooltip-at-a-time UX) — and the `hidingStores` + `hideStore` pair is the cleanest of the lot. It leaves the store API untouched, scopes the bookkeeping to `tooltip-anchor.tsx`, costs an `O(1)` WeakSet check per mount, and doesn't change the uncontrolled-tooltip path.

### The changeset commit

`1b1eaee98` applies the wording you suggested on the changeset thread (drops the `…while preserving the default one-open-tooltip behavior for controlled tooltips that accept close updates` tail). The shorter sentence reads cleanly and the trimmed nuance is still covered by the PR body. No follow-up needed there.

## Validation performed

- `pnpm vitest run app/src/sandbox/tooltip-4894/test.ts` — all 3 tests pass on PR head.
- Reverted only `packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx` to `origin/main` while keeping the new tests; the first test (forced) then fails with React's `Maximum update depth exceeded` (caught via `vitest-fail-on-console`), confirming the test catches the regression.
- `pnpm react18 test app/src/sandbox/tooltip-4894/test.ts` — passes under the React 18 matrix as well.
- `pnpm vitest run examples/menu-tooltip examples/menu-tooltip-various examples/tooltip examples/tooltip-label examples/tooltip-instant examples/tooltip-framer-motion` — all 47 existing tooltip tests still pass.

## Findings

One inline comment on `hideStore` (lines 38–43) suggesting a brief comment to lock in why the cleanup microtask is scheduled **after** `store.hide()` rather than before. The ordering is what actually breaks the re-entrant cycle (because `setState`'s batch microtask, scheduled inside `hide()`, must fire before the cleanup microtask). The function is small enough that a refactor could reorder these lines without realizing the consequence, and a one-line comment would prevent that.

No correctness, regression, perf, compatibility, or test-coverage issues otherwise. The expanded sandbox (3 forced + 2 managed) is a better repro than the previous version, the changeset stays patch-level for `@ariakit/react-components`/`@ariakit/react` only, and there's no public-surface change.

## Areas I did not fully validate

- I did not run `test-browser.ts` against the dev-app preview (Playwright). The Vitest/jsdom test covers the same StrictMode-enabled scenarios with the same assertions.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
diegohaz: Fixed in e73dcd038 - documented why the cleanup microtask must be scheduled after hide() so forced controlled stores can still short-circuit the re-entrant mounted sync.

diegohaz:
**Fixed**

- e73dcd038 ([reply](https://github.com/ariakit/ariakit/pull/6067#discussion_r3319465067)): Documented why tooltip hide cleanup must run after `hide()` so forced controlled stores can short-circuit the re-entrant mounted sync.

app/src/sandbox/tooltip-4894/test.ts
```diff
@@ -0,0 +1,34 @@
+import { click, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+// See https://github.com/ariakit/ariakit/issues/4894
+test("does not loop when multiple tooltips are forced open", async () => {
```
diegohaz: **Low priority — this headline regression test doesn't actually fail without the library fix.**

I restored `origin/main`'s pre-fix `tooltip-anchor.tsx` into this branch and ran the three Vitest cases in isolation:

- `does not loop when multiple tooltips are forced open` → **passes** (green) without the fix
- `hides controlled tooltips that accept setOpen updates` → **passes** without the fix
- `keeps managed tooltips active after forced tooltips reopen` → **fails** with `Maximum update depth exceeded`

So in jsdom the only case that exercises the re-entrant loop is the forced+managed test (#3); the forced-open-only scenario settles without hitting the loop. The suite as a whole does go red on `main` (good), but it's the managed test carrying that guard — not the test named after the issue. If #3 is later refactored, the forced-open regression could quietly stop being covered in jsdom.

Two paths:

- If the pure forced-open loop can be reproduced in jsdom, tighten this test so it fails without the fix (e.g. assert/fail on the `Maximum update depth` error raised during the forced mount).
- If it only reproduces in a real browser, that's fine — but then this jsdom test is documentation rather than a guard, and it's worth a note that `test-browser.ts` owns the forced-open crash path.

Caveat: I couldn't run `test-browser.ts` locally (needs the dev server + Playwright), so I haven't confirmed which test owns the forced-open path in a real browser — flagging this as suspected, not certain.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
```diff
@@ -78,11 +89,18 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
...
+          // open tooltips.
           if (state.mounted) {
             const { activeStore } = globalStore.getState();
             if (activeStore !== store) {
-              activeStore?.hide();
+              if (hidingStores.has(store)) {
+                return;
+              } else {
+                hideStore(activeStore);
+              }
```
diegohaz: Nit (non-blocking simplification): the `else` branch is redundant because the `if` returns, and it adds a nesting level. An early return reads cleaner and matches the repo convention of preferring early exits over `else`-after-`return`:

```suggestion
              if (hidingStores.has(store)) return;
              hideStore(activeStore);
```

Behavior is identical.

diegohaz (commented):
## Summary

Re-review of PR #6067 at head `e73dcd038`, incremental against the prior marked review's baseline `1b1eaee98`. One new commit since then:

- `e73dcd038` ("Document tooltip hide ordering") — adds the comment on `hideStore` explaining why the `queueMicrotask` cleanup is scheduled after `store.hide()`. This resolves the last open thread on that line, and the comment is accurate.

All three prior review threads are resolved, and I verified the claimed fixes against the current code (rationale comment present at the `sync` call site, changeset wording simplified, hide-ordering comment added).

### Validation performed (PR head, Node 24)

- `npx tsc` → clean
- `oxlint --disable-nested-config` + `oxfmt --check` on the changed files → clean
- `vitest run app/src/sandbox/tooltip-4894/test.ts` → 3/3 pass
- **Regression proof:** restored `main`'s pre-fix `tooltip-anchor.tsx` and re-ran the suite → **red** (`Maximum update depth exceeded`); restored the fix → **green**. The fix demonstrably resolves the re-entrant loop.

### Findings

1. **(low) jsdom test #1 isn't independently sensitive to the bug.** Run in isolation against the pre-fix library, `does not loop when multiple tooltips are forced open` and `hides controlled tooltips that accept setOpen updates` both pass; only `keeps managed tooltips active after forced tooltips reopen` fails. The forced-open-only path settles in jsdom, so the actual jsdom guard is the managed+forced test. Inline comment on `test.ts`. (I couldn't run the Playwright `test-browser.ts` locally, so I haven't confirmed which test owns the forced-open path in a real browser.)
2. **(nit) redundant `else` after `return`** in `tooltip-anchor.tsx` — behavior-preserving simplification with an inline suggestion.

The library fix itself is correct and reasonably scoped: the `WeakSet` + microtask bounce-detection is a sound way to break the loop for forced-open tooltips while preserving the one-at-a-time behavior for managed controlled tooltips. I considered the simpler "skip `hide()` when the active store is controlled-open" alternative (the earlier revision of this PR), but it would also stop managed controlled tooltips from closing each other, so the current runtime bounce-detection is the better trade-off.

### Not fully validated

- Browser/Playwright tests (`test-browser.ts`) — not run locally (need dev server + Playwright); relying on CI.
- Real-page crash behavior across the browser matrix — relying on CI.

This is a comment-only review.

app/src/sandbox/tooltip-4894/test.ts
diegohaz: Fixed in 04ba4dd3f - clarified that the jsdom forced-open test documents the stable state, while the mixed jsdom test carries the re-entrant-loop guard and the browser test covers the pure forced-open browser flow.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
diegohaz: Fixed in 04ba4dd3f - simplified the branch to an early return before calling hideStore(activeStore).

diegohaz:
**Fixed**

- 04ba4dd3f ([reply](https://github.com/ariakit/ariakit/pull/6067#discussion_r3319894899)): Clarified jsdom/browser coverage ownership for the forced-open tooltip test.
- 04ba4dd3f ([reply](https://github.com/ariakit/ariakit/pull/6067#discussion_r3319894902)): Simplified the tooltip active-store branch to an early return.

diegohaz (commented):
## Summary

Re-review of PR #6067 at head `04ba4dd3f`, incremental against the prior marked review baseline `e73dcd038`. Two new commits since then:

- `d7b8f3fc1` — merge of `origin/main`. I confirmed it brings in only unrelated work (button-form-status, dialog, focusable) and **does not touch the store / tooltip / hovercard internals**, so the fix's batching/microtask runtime is unchanged. The final PR diff against the new merge base is still exactly the intended 7 files (`+207 / -2`), no conflict markers, no stray inclusions.
- `04ba4dd3f` ("Clarify tooltip test coverage") — addresses both findings from the previous review.

### Both prior findings resolved and verified

1. **jsdom test sensitivity (was: low)** — Resolved. The forced-open test was renamed to `keeps multiple forced tooltips visible` with a comment clarifying that in jsdom it documents the stable state, the mixed test carries the re-entrant-loop guard, and `test-browser.ts` covers the pure forced-open flow. I re-ran the per-test regression proof on the post-merge head against `origin/main`'s pre-fix `tooltip-anchor.tsx`: tests 1 and 2 pass on the unfixed library, and `keeps managed tooltips active after forced tooltips reopen` **fails** (`Maximum update depth exceeded`) — so the author's claim that the mixed test is the actual jsdom guard is accurate. Clarifying the semantics is the right call given the pure forced-open loop doesn't reproduce in jsdom.
2. **redundant `else` (was: nit)** — Resolved. Applied exactly as an early return (`if (hidingStores.has(store)) return; hideStore(activeStore);`). Behavior-identical and tests stay green.

### Validation performed (post-merge head `04ba4dd3f`, Node 24)

- `npx tsc` → clean
- `oxlint --disable-nested-config` + `oxfmt --check` on changed files → clean
- `vitest run app/src/sandbox/tooltip-4894/test.ts` → 3/3 pass
- Per-test regression proof (restore pre-fix lib → only the mixed test goes red; restore fix → all green)

### Verdict

No new findings. Both prior comments are resolved correctly, the fix remains valid on the updated base, and the change is in good shape.

### Not validated locally

- Playwright `test-browser.ts` and the cross-browser real-page crash behavior — relying on CI.

This is a comment-only review.

PR #6068, opened 2026-05-28
Fix Tab active id sync

### LINKED ISSUE #4213: Tab: focused element and active tab id can go out of sync
### Current behavior

Keyboard focus and active tab id can go out of sync if `Tab` receives a controlled update while the keyboard focus is on a tab within the tablist.

https://github.com/user-attachments/assets/61768a61-22ff-40a7-91d4-847e1c320a6f

Note: this behavior can be reproduced regardless of the `selectOnMove` preference.

1. Open https://stackblitz.com/edit/xpiaoy?file=tab%2Findex.tsx&theme=dark
2. When the example loads, tab into the tablist and press the right arrow key to select the "Vegetables" tab
3. Wait for the timeout to run, applying a controlled update and selecting the "Meat" tab
4. At this point, the Vegetables tab has keyboard focus, but the Meat tab is selected
5. Press the right arrow key.
6. The "Fruit" tab gets selected

### Expected behavior

I agree that we shouldn't change the focused item (and I think that the behavior described at point 4 is correct), but as I user I would be confused by what happens at point 6: I'd expect the "Meat" tab to receive keyboard focus, and not the "Fruit" tab.

In other words, I'd expect the active tab to stay the same as the focused tab, instead of following the selected tab.

### Workaround

Manually re-set the active ID if it goes out of sync.

```jsx
  const { items, activeId } = Ariakit.useStoreState(tabStore);
  const { setActiveId } = tabStore;

  useEffect(() => {
    requestAnimationFrame(() => {
      const focusedElement = items?.[0]?.element?.ownerDocument.activeElement;

      if (
        !focusedElement ||
        !items.some((item) => focusedElement === item.element)
      ) {
        // Return early if no tabs are focused.
        return;
      }

      // If, after ariakit re-computes the active tab, that tab doesn't match
      // the currently focused tab, then we force an update to ariakit to avoid
      // any mismatches, especially when navigating to previous/next tab with
      // arrow keys.
      if (activeId !== focusedElement.id) {
        setActiveId(focusedElement.id);
      }
    });
  }, [activeId, items, setActiveId]);
```

### Possible solutions

_No response_

diegohaz:
## Motivation

Fixes #4213.

A controlled tab selection update can make DOM focus and the tab store's active item diverge. When a user presses an arrow key after that mismatch, keyboard navigation starts from the selected tab instead of the focused tab.

The later issue discussion settled on matching synchronous tab selection behavior when the tab list is already in use: if a tab has DOM focus and `selectedId` changes, focus should move to the newly selected tab. This is also consistent with `store.select(tabId)`, which already selects a tab and moves focus to it.

## Solution

When `selectedId` changes, the tab store now checks whether one of the rendered tabs currently has DOM focus and whether the new selected tab is rendered. If both are true, it moves focus to the selected tab through the composite move path.

If focus is outside the tab item, the store only syncs `activeId` with `selectedId` and does not steal DOM focus. This preserves the existing behavior where tabbing into the tab list later lands on the selected tab.

The regression sandbox covers these paths:

- A delayed controlled update while Vegetables has focus selects Meat and moves focus to Meat.
- A controlled update triggered from a focusable button outside the tab list updates selection without moving focus away from the button.
- A controlled update triggered from an input inside an always-visible tab panel updates selection without moving focus away from the input.

## Workaround

Until this fix is released, consumers that need the focus-moving behavior can use the tab store's `select` function when changing tabs programmatically:

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/4213 is fixed.
tab.select(tabId);
```

For controlled tabs, make sure the controlled `selectedId` state is updated from `setSelectedId` so the store selection and application state stay in sync.

### DIFF
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -19,7 +19,18 @@ export default function Example() {
   return (
     <ak.TabProvider store={tab}>
       <ak.TabList aria-label="Groceries">
-        <ak.Tab id="fruits">Fruits</ak.Tab>
+        <ak.Tab
+          id="fruits"
+          onKeyDown={(event) => {
+            if (event.key.toLowerCase() !== "d") return;
+            setSelectedTab("dairy");
+          }}
+        >
+          Fruits
+        </ak.Tab>
+        <ak.Tab id="dairy" disabled>
+          Dairy
+        </ak.Tab>
         <ak.Tab id="vegetables">Vegetables</ak.Tab>
         <ak.Tab id="meat">Meat</ak.Tab>
       </ak.TabList>
@@ -37,6 +48,7 @@ export default function Example() {
         Carrots, onions, and potatoes
       </ak.TabPanel>
       <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
+      <ak.TabPanel tabId="dairy">Milk, cheese, and yogurt</ak.TabPanel>
       {/* WebKit keeps focus on the clicked button only when explicitly tabbable. */}
       <button
         type="button"
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -37,13 +37,14 @@ export default function Example() {
         Carrots, onions, and potatoes
       </ak.TabPanel>
       <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
-      <label>
-        Outside note
-        <input
-          type="text"
-          onFocus={() => setTimeout(() => setSelectedTab("vegetables"), 100)}
-        />
-      </label>
+      {/* WebKit keeps focus on the clicked button only when explicitly tabbable. */}
+      <button
+        type="button"
+        tabIndex={0}
+        onClick={() => setSelectedTab("vegetables")}
+      >
+        Select vegetables
+      </button>
     </ak.TabProvider>
   );
 }
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -28,6 +28,9 @@ export default function Example() {
         Carrots, onions, and potatoes
       </ak.TabPanel>
       <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
+      <button type="button" onClick={() => setSelectedTab("vegetables")}>
+        Select vegetables
+      </button>
     </ak.TabProvider>
   );
 }
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -49,9 +49,9 @@ export default function Example() {
       </ak.TabPanel>
       <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
       <ak.TabPanel tabId="dairy">Milk, cheese, and yogurt</ak.TabPanel>
-      {/* WebKit keeps focus on the clicked button only when explicitly tabbable. */}
       <button
         type="button"
+        // WebKit keeps focus on the clicked button only when explicitly tabbable.
         tabIndex={0}
         onClick={() => setSelectedTab("vegetables")}
       >
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -37,9 +37,13 @@ export default function Example() {
         Carrots, onions, and potatoes
       </ak.TabPanel>
       <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
-      <button type="button" onClick={() => setSelectedTab("vegetables")}>
-        Select vegetables
-      </button>
+      <label>
+        Outside note
+        <input
+          type="text"
+          onFocus={() => setTimeout(() => setSelectedTab("vegetables"), 100)}
+        />
+      </label>
     </ak.TabProvider>
   );
 }
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -9,12 +9,28 @@ export default function Example() {
     selectedId: selectedTab,
     setSelectedId: setSelectedTab,
   });
+  const activeId = ak.useStoreState(tab, "activeId");
+  const renderedItems = ak.useStoreState(tab, "renderedItems");
 
   useEffect(() => {
     if (selectedTab !== "vegetables") return;
     const timeout = setTimeout(() => setSelectedTab("meat"), 100);
     return () => clearTimeout(timeout);
   }, [selectedTab]);
+
+  useEffect(() => {
+    const frame = requestAnimationFrame(() => {
+      const activeElement =
+        renderedItems[0]?.element?.ownerDocument.activeElement;
+      const focusedItem = renderedItems.find(
+        (item) => item.element === activeElement,
+      );
+      if (!focusedItem) return;
+      if (activeId === focusedItem.id) return;
+      tab.setActiveId(focusedItem.id);
+    });
+    return () => cancelAnimationFrame(frame);
+  }, [activeId, renderedItems, tab]);
 
   return (
     <ak.TabProvider store={tab}>
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -23,7 +23,16 @@ export default function Example() {
         <ak.Tab id="vegetables">Vegetables</ak.Tab>
         <ak.Tab id="meat">Meat</ak.Tab>
       </ak.TabList>
-      <ak.TabPanel tabId="fruits">Apples, grapes, and oranges</ak.TabPanel>
+      <ak.TabPanel tabId="fruits" alwaysVisible>
+        Apples, grapes, and oranges
+        <label>
+          Fruit note
+          <input
+            type="text"
+            onFocus={() => setTimeout(() => setSelectedTab("vegetables"), 100)}
+          />
+        </label>
+      </ak.TabPanel>
       <ak.TabPanel tabId="vegetables">
         Carrots, onions, and potatoes
       </ak.TabPanel>
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -1,0 +1,33 @@
+import * as ak from "@ariakit/react";
+import { useEffect, useState } from "react";
+
+export default function Example() {
+  const [selectedTab, setSelectedTab] = useState<string | null | undefined>(
+    "fruits",
+  );
+  const tab = ak.useTabStore({
+    selectedId: selectedTab,
+    setSelectedId: setSelectedTab,
+  });
+
+  useEffect(() => {
+    if (selectedTab !== "vegetables") return;
+    const timeout = setTimeout(() => setSelectedTab("meat"), 100);
+    return () => clearTimeout(timeout);
+  }, [selectedTab]);
+
+  return (
+    <ak.TabProvider store={tab}>
+      <ak.TabList aria-label="Groceries">
+        <ak.Tab id="fruits">Fruits</ak.Tab>
+        <ak.Tab id="vegetables">Vegetables</ak.Tab>
+        <ak.Tab id="meat">Meat</ak.Tab>
+      </ak.TabList>
+      <ak.TabPanel tabId="fruits">Apples, grapes, and oranges</ak.TabPanel>
+      <ak.TabPanel tabId="vegetables">
+        Carrots, onions, and potatoes
+      </ak.TabPanel>
+      <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
+    </ak.TabProvider>
+  );
+}
--- a/app/src/sandbox/tab-4213/index.react.tsx
+++ b/app/src/sandbox/tab-4213/index.react.tsx
@@ -9,28 +9,12 @@ export default function Example() {
     selectedId: selectedTab,
     setSelectedId: setSelectedTab,
   });
-  const activeId = ak.useStoreState(tab, "activeId");
-  const renderedItems = ak.useStoreState(tab, "renderedItems");
 
   useEffect(() => {
     if (selectedTab !== "vegetables") return;
     const timeout = setTimeout(() => setSelectedTab("meat"), 100);
     return () => clearTimeout(timeout);
   }, [selectedTab]);
-
-  useEffect(() => {
-    const frame = requestAnimationFrame(() => {
-      const activeElement =
-        renderedItems[0]?.element?.ownerDocument.activeElement;
-      const focusedItem = renderedItems.find(
-        (item) => item.element === activeElement,
-      );
-      if (!focusedItem) return;
-      if (activeId === focusedItem.id) return;
-      tab.setActiveId(focusedItem.id);
-    });
-    return () => cancelAnimationFrame(frame);
-  }, [activeId, renderedItems, tab]);
 
   return (
     <ak.TabProvider store={tab}>
--- a/app/src/sandbox/tab-4213/test-browser.ts
+++ b/app/src/sandbox/tab-4213/test-browser.ts
@@ -22,7 +22,20 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await button.click();
     await test.expect(button).toBeFocused();
 
+    // Selecting Vegetables triggers a delayed controlled update to Meat.
     await test.expect(q.tab("Meat")).toHaveAttribute("aria-selected", "true");
     await test.expect(button).toBeFocused();
+  });
+
+  test("does not move focus after controlled tab selection inside a tab panel", async ({
+    page,
+    q,
+  }) => {
+    const textbox = page.getByRole("textbox", { name: "Fruit note" });
+    await textbox.focus();
+    await test.expect(textbox).toBeFocused();
+
+    await test.expect(q.tab("Meat")).toHaveAttribute("aria-selected", "true");
+    await test.expect(textbox).toBeFocused();
   });
 });
--- a/app/src/sandbox/tab-4213/test-browser.ts
+++ b/app/src/sandbox/tab-4213/test-browser.ts
@@ -20,7 +20,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    const button = page.getByRole("button", { name: "Select vegetables" });
+    const button = q.button("Select vegetables");
     await button.click();
     await test.expect(button).toBeFocused();
 
@@ -29,10 +29,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   test("does not move focus after controlled tab selection inside a tab panel", async ({
-    page,
     q,
   }) => {
-    const textbox = page.getByRole("textbox", { name: "Fruit note" });
+    const textbox = q.textbox("Fruit note");
     await textbox.focus();
     await test.expect(textbox).toBeFocused();
 
--- a/app/src/sandbox/tab-4213/test-browser.ts
+++ b/app/src/sandbox/tab-4213/test-browser.ts
@@ -17,7 +17,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   test("does not move focus after controlled tab selection outside the tab list", async ({
-    page,
     q,
   }) => {
     const button = q.button("Select vegetables");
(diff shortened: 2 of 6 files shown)


diegohaz (commented):
Initial review. Baseline: `240d5ad5e` (merge base with `main`); head: `f1f237fd8`.

### Scope reviewed

- `packages/ariakit-components/src/tab/tab-store.ts` — the actual fix.
- `app/src/sandbox/tab-4213/{index.react.tsx,preview.astro,preview.mdx,test-browser.ts}` — repro and browser test.
- `.changeset/300-tab-active-id-focus.md` — changeset.
- Surrounding context: `packages/ariakit-store/src/index.ts` (`batch` / `setState` semantics), `packages/ariakit-react-store/src/index.tsx` (`useStore` / `init` timing), `packages/ariakit-react-components/src/tab/tab.tsx` and `tab-store.ts` (React layer), `packages/ariakit-components/src/collection/collection-store.ts` (`renderedItems` population), and the existing tab sandboxes (`tab-panel-5692`, `tab-form-5273`) and example tests (`examples/tab/test.ts`).

### What I checked

- The chosen behavior (keep `activeId` on the focused tab rather than moving DOM focus to the newly selected tab) matches the conclusion in the [#4213 discussion](https://github.com/ariakit/ariakit/issues/4213#issuecomment-2422690123).
- The new `batch(tab, ["selectedId"], ...)` body is consistent with the surrounding bail-outs: `syncActiveId === false`, the `parentComposite && state.selectedId === prev.selectedId` initial-fire guard, and the focus-driven branch all compose correctly. Setting `activeId` cannot recursively retrigger this listener (it batches on `selectedId` only), and the existing `selectedIdFromSelectedValue` restoration flow still flips `syncActiveId = false` before changing `selectedId`, so the focused-tab branch is correctly bypassed in that path.
- `getFocusedTab` is harmless when focus is outside the tab list, on the panel, or in a sibling tab list — `items.find(...)` returns `undefined` and the original `activeId = state.selectedId` fallback runs.
- The sandbox name `tab-4213` follows the issue-repro convention from `ariakit-ariakit-workflow`. The browser test uses `withFramework` + `q.tab(...)` per `ariakit-ariakit-example-playwright-test`. The changeset follows `ariakit-general-changeset` (patch, `300-` prefix, single-sentence past-tense entry with reference link).

### Validation

- `pnpm tsc` — passes.
- `pnpm test` — 846 tests / 152 files pass, including `examples/tab/test.ts`.
- Walked through the bug scenario step-by-step against the new code: click `Vegetables` → `setSelectedId("vegetables")` → DOM focus on `Vegetables` → batch fires → `getFocusedTab` returns `Vegetables` → `activeId = vegetables`. After the 100ms controlled flip to `meat`: batch fires → DOM focus still on `Vegetables` → `activeId` stays on `Vegetables`. `ArrowRight` then advances to `Meat` as expected. The pre-fix path sets `activeId = "meat"` and `ArrowRight` wraps to `Fruits`, which matches the issue's repro video.

### Findings

None. The change is small, focused, and well-targeted; the regression test exercises the exact issue scenario.

packages/ariakit-components/src/tab/tab-store.ts
```diff
@@ -120,6 +126,12 @@ export function createTabStore({
       // activeId state with the initial selectedId state. The parent composite
       // widget should handle the initial activeId state.
       if (parentComposite && state.selectedId === prev.selectedId) return;
+      const { renderedItems } = tab.getState();
+      const focusedTab = getFocusedTab(renderedItems);
+      if (focusedTab) {
+        tab.setState("activeId", focusedTab.id);
+        return;
+      }
```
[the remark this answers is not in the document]
diegohaz: I evaluated this against the current code, in a real browser (not just jsdom), and the `selectOnMove` arrow flow is **not** broken.

The premise that "DOM focus is still on tab A" when the `batch(["selectedId"])` listener runs does not hold. During an arrow press, `composite.move(B)` bumps `moves`, and the focus-on-move effect (`composite/composite.tsx:190`) advances DOM focus to B **before** the queued `selectedId` microtask runs. So `getFocusedTab(renderedItems)` returns B, `tab.setState("activeId", B)` is a no-op, and `activeId` is never reset to A.

I confirmed this by instrumenting `getFocusedTab` and running `app/src/sandbox/tab-form-5273` (arrow navigation with default `selectOnMove`) in Chromium. At the `selectedId` batch for `Preferences → Account`, the listener already observes the **Account** tab as both active and focused:

```
{ prevSelected: <Preferences>, selectedId: <Account>, activeId: <Account>, focusedTab: <Account> }
```

That is exactly why the existing `tab-form-5273` browser test — and CI's Chrome / Firefox / Safari runs on this head — pass.

The distinction the fix relies on is DOM focus itself: a `move()`-driven selection advances focus (so the focused tab *is* the new selection → no-op), whereas a controlled `selectedId` update does not move focus (so the focused tab is still the user's tab → the #4213 case the fix targets). Because the live focus state already encodes that difference, the extra `selectOnMove` flag isn't necessary. No change needed here.

diegohaz (commented):
**Re-review.** The PR head is unchanged since the previous review (`f1f237fd8`), so this run focused on the one open item not covered before: the unresolved automated-review thread on `tab-store.ts:134` claiming the fix breaks `selectOnMove` arrow navigation.

### Conclusion

That concern is a **false positive** — detail in my [reply on the thread](https://github.com/ariakit/ariakit/pull/6068#discussion_r3319629063). During arrow navigation, `composite.move()` advances DOM focus to the new tab *before* the queued `selectedId` batch listener runs, so `getFocusedTab` returns the new tab and `activeId` is never reverted. A controlled `selectedId` update (the #4213 case) does not move focus, so the helper correctly keeps `activeId` on the focused tab. The live DOM focus state distinguishes the two cases, so no extra `selectOnMove` flag is needed.

### Validation

- Instrumented `getFocusedTab` and ran `app/src/sandbox/tab-form-5273` (arrow nav, default `selectOnMove`) in Chromium via `pnpm -F app test`. At the `Preferences → Account` selection the batch listener already observed `activeId === focusedTab === Account` — no revert. Instrumentation reverted; working tree clean.
- CI on this head passes Test Chrome / Firefox / Safari, which run both `tab-form-5273` and the new `tab-4213` browser tests.

### Other areas checked — no new findings

- `getFocusedTab` is safe when focus is outside the tablist (`items.find(...)` returns `undefined`, falling back to `activeId = selectedId`), and when `items[0]` has no element yet.
- The new branch composes correctly with the existing bail-outs (`syncActiveId`, the `selectedIdFromSelectedValue` restore path that sets `syncActiveId = false`, and the `parentComposite` initial-fire guard). Setting `activeId` does not retrigger this `selectedId`-keyed listener.
- Changeset is correct: patch, `300-` prefix, single past-tense sentence with reference links. The package set (`@ariakit/components`, `@ariakit/react-components`, `@ariakit/react`) is right — `@ariakit/solid-components` has no tab module and does not depend on `@ariakit/components`, so the Solid packages are correctly omitted.
- The regression test is a Playwright browser test, which is the correct venue: the bug depends on real focus-effect-vs-microtask timing that `@ariakit/test` (jsdom + `act()`) can mask, so a jsdom test would not reliably reproduce it.

Comment-only review; no new actionable findings beyond the thread reply.

packages/ariakit-components/src/tab/tab-store.ts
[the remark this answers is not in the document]
diegohaz: This is not an issue because move-driven tab navigation has already moved DOM focus to the destination tab by the time the `selectedId` batch runs, so the focused-tab branch does not reset `activeId` back to the previous tab.

I rechecked this after merging `main` and ran:

```sh
APP_PORT=4322 NEXTJS_PORT=3001 pnpm -F app run test-chrome -- app/src/sandbox/tab-form-5273/ app/src/sandbox/tab-4213/
```

The warm rerun passed both focused tests: `tab-form-5273` still moves `Preferences` to `Account` with ArrowRight, and `tab-4213` still covers the controlled `selectedId` regression. A `selectOnMove` escape flag would be unnecessary for this path.

diegohaz:
**Invalid**

- Review thread (https://github.com/ariakit/ariakit/pull/6068#discussion_r3319866474): The `selectOnMove` regression concern does not reproduce. Verified after merging `main` with the focused Chrome run for `tab-form-5273` and `tab-4213`; both tests pass, so no `selectOnMove` escape flag is needed.

packages/ariakit-components/src/tab/tab-store.ts
```diff
@@ -120,6 +131,13 @@ export function createTabStore({
       // activeId state with the initial selectedId state. The parent composite
       // widget should handle the initial activeId state.
       if (parentComposite && state.selectedId === prev.selectedId) return;
+      const { activeId, renderedItems } = tab.getState();
+      const focusedTab = getFocusedTab(renderedItems);
+      const selectedTab = getTabById(renderedItems, state.selectedId);
+      if (focusedTab && selectedTab && activeId !== selectedTab.id) {
```
diegohaz: The new focus-move branch doesn't guard against `disabled`/`dimmed` tabs, so a controlled `selectedId` update that targets a disabled tab while another tab has DOM focus will move focus **to the disabled tab**. This is inconsistent with the two sibling listeners that intentionally skip disabled/dimmed tabs — the `selectOnMove` listener (lines 112–113) and the auto-select listener (lines 152, 159) — and it's a behavior change from `main`, where setting `selectedId` to a disabled tab only synced `activeId` and never moved DOM focus.

Why it matters: `composite.move(selectedTab.id)` bumps `moves`, and the focus-on-move effect calls `focusIntoView` on the item regardless of disabled state (`getEnabledItem` just returns `store.item(id)` without filtering). Ariakit tabs are `accessibleWhenDisabled` by default, so the disabled tab is focusable and ends up with real DOM focus on an `aria-disabled="true"` element.

I reproduced this in jsdom (3 tabs, `Tab3` disabled, controlled `selectedId`): focus `Tab1`, then set `selectedId` to `Tab3` while keeping focus on `Tab1`. On this branch, focus moves to `Tab3`:

```
Received element with focus:
  <button aria-disabled="true" aria-selected="true" data-active-item="true"
          id="tab3" role="tab">Tab3</button>
```

Adding the same `disabled`/`dimmed` guard the other listeners use makes focus stay put (selection still updates to `Tab3`), and all existing tab tests (`examples/tab`, `tab-disabled`, `tab-nested`, `tab-react-router`) still pass:

```suggestion
      if (
        focusedTab &&
        selectedTab &&
        !selectedTab.disabled &&
        !selectedTab.dimmed &&
        activeId !== selectedTab.id
      ) {
```

If moving focus onto a disabled selected tab is actually intended here, it'd be worth a comment explaining why this path diverges from `selectOnMove`. Either way, a regression test for the disabled/dimmed case in `app/src/sandbox/tab-4213/` would lock in the chosen behavior — the current sandbox only exercises enabled tabs.

diegohaz (commented):
**Re-review.** Comment-only.

**Baseline.** Last marked review from this workflow was at `head_sha=f1f237fd` (the original "set `activeId` to the focused tab" approach). Since then the fix was **reworked** in `979a5e282` to *move focus to the selected tab* via `composite.move`, plus a `main` merge and test-infra commits (`4b45177`, `6102dd1`, `5c2a62c`). Because the approach changed substantially, I re-reviewed the full current `tab-store.ts` logic — incremental range `f1f237fd..5c2a62c4`, full PR diff `712277e3..5c2a62c4` (6 files).

**Scope.** `packages/ariakit-components/src/tab/tab-store.ts` (new `getFocusedTab`/`getTabById` helpers + the `selectedId` batch focus-move branch), the `tab-4213` sandbox/browser tests, and the changeset. Followed call sites into `composite-store.ts` (`move`, `virtualFocus`), the React focus-on-move effect (`composite.tsx:190`), and the Select/Combobox-with-Tabs paths.

**Findings.**
- 1 inline (low–medium, validated): the focus-move branch doesn't guard `disabled`/`dimmed` selected tabs, so a controlled `selectedId` update to a disabled tab while a tab has focus steals DOM focus onto an `aria-disabled` element — diverging from the `selectOnMove` (112–113) and auto-select (152/159) listeners and from `main`. Minimal guard suggested + a regression-test suggestion.

**Re-verified prior threads.** The Copilot `selectOnMove` thread (against the old approach) does **not** apply to the new code: the `activeId !== selectedTab.id` guard skips the branch during arrow navigation (selectOnMove derives `selectedId` *from* `activeId`, so they're always equal) — this holds independently of focus timing, which is stronger than the old approach. No reply needed; leaving that resolved thread as-is.

**Validation run.**
- `pnpm install`; existing tab jsdom tests pass: `examples/tab`, `tab-disabled`, `tab-nested`, `tab-react-router` (9/9).
- Wrote a throwaway jsdom repro confirming the disabled-tab focus steal, confirmed the suggested guard fixes it while keeping selection and all existing tab tests green, then **reverted all experiment changes** (branch is clean).
- Traced re-entrancy: `composite.move` inside the `selectedId` batch → `moves` → `selectOnMove` sets `selectedId` to the same value (no-op, no loop); `select()` still moves focus exactly once.
- CI is green on all checks (Type check, Lint, Build, Chrome/Firefox/Safari/Node/React 18/macOS/Android).

**Checked and dismissed.** `getFocusedTab` vs the shared `getActiveElement` util (iframe/IE11 handling unnecessary for tabs); `ownerDocument` optional-chaining (safe); `selectedTab` "redundancy" (it's a needed *rendered* check, scoped to `renderedItems` unlike `composite.item`); virtualFocus / Combobox- & Select-with-Tabs (focus stays on the base element, so `getFocusedTab` returns `undefined` and the branch never fires); non-rendered selected tab (correctly falls through to `activeId` sync). Changeset format/bump (`@ariakit/components` core + `react-components` + `react`) and JSDoc updates are accurate.

**Not fully validated.** The Playwright browser sandbox tests weren't run locally (they need the worktree dev server); I relied on green CI for the cross-browser/Safari focus behavior. The disabled/dimmed finding was validated in jsdom, not a real browser.

packages/ariakit-components/src/tab/tab-store.ts
diegohaz: Fixed in 321a80804 - the controlled `selectedId` focus-moving branch now skips disabled and dimmed selected tabs, while still syncing `activeId`. I also added a `tab-4213` regression that selects a disabled tab while another tab has DOM focus and verifies focus stays put.

diegohaz:
**Fixed**

- 321a80804 ([reply](https://github.com/ariakit/ariakit/pull/6068#discussion_r3320850247)): Guarded controlled `selectedId` focus moves so disabled/dimmed selected tabs do not receive DOM focus, and added `tab-4213` regression coverage for that path.

packages/ariakit-components/src/tab/tab-store.ts
diegohaz: Verified the fix in `321a80804` against the current code — looks good. The `isEnabledTab` type guard nicely DRYs the disabled/dimmed checks across the `selectOnMove` listener, both auto-select branches, and the focus-move branch (the three pre-existing sites are behavior-preserving; only the focus-move branch gains the new guard), and the guard narrows `selectedTab` so `.id` stays typed.

I reproduced it in jsdom: with the guard, a controlled `selectedId` update to a disabled tab while another tab has focus keeps focus put (selection still updates); reverting just that branch to the bare `selectedTab` check makes focus jump onto the `aria-disabled` tab — so the new `tab-4213` regression genuinely fails without the fix. `examples/tab`/`tab-disabled` jsdom tests, `tsc`, and CI (Chrome/Firefox/Safari) are all green. Agreed on resolving.

diegohaz (commented):
**Re-review.** Comment-only. No remaining findings.

**Baseline.** Last marked review from this workflow was at `head_sha=5c2a62c4`. This run covers only the single new commit since then — `321a80804` "Avoid focusing disabled tabs on controlled selection" (incremental range `5c2a62c4..321a80804`; full PR diff `712277e3..321a80804`, 6 files / +185 −10).

**What changed & verdict.** This commit resolves the one finding from the previous review (focus-move branch stealing DOM focus onto a disabled/dimmed selected tab). The fix extracts a reusable `isEnabledTab` type guard and applies it at all four sites — the `selectOnMove` listener, both auto-select branches, and the focus-move branch — plus a JSDoc update and a regression test. This is cleaner than the inline guard I'd suggested.

**Verified (not just trusting the resolved thread):**
- The three pre-existing sites are behavior-preserving (`isEnabledTab` is exactly `item && !disabled && !dimmed`, with correct `item is TabStoreItem` narrowing so `.id` stays typed); only the focus-move branch gains the new guard.
- Reproduced in jsdom: with the guard, a controlled `selectedId` update to a disabled tab while another tab has focus keeps focus put and still flips `aria-selected`; reverting that branch to the bare `selectedTab` check makes focus jump onto the `aria-disabled` tab — so the new `tab-4213` test genuinely fails without the fix.
- The new test is sound: `"d"` triggers only the sandbox's `onKeyDown` (Tab doesn't use `CompositeTypeahead`, so no find-as-you-type), no flakiness, and the test-1 `.focus()`+`.click()` change doesn't mask the load-bearing `Meat` focus assertion. Inserting the disabled Dairy tab doesn't disturb the other tests (the `Meat → Fruits` ArrowRight wrap skips disabled Dairy).
- `pnpm tsc` clean; `examples/tab` + `examples/tab-disabled` jsdom tests pass; CI fully green including Test Chrome/Firefox/Safari (which run the new disabled-tab browser test). An independent 3-lens adversarial pass returned 0 findings.

**Non-blocking observations (no action needed).** The comment at lines 166-167 ("…or the active tab is dimmed") predates this commit and omits "disabled" — pre-existing, not introduced here. The changeset accurately describes the user-facing behavior; the "only when enabled" refinement is an implementation sub-detail that doesn't need a changeset edit.

**Not validated locally.** The Playwright browser sandbox tests weren't run in this worktree; I relied on green CI for the cross-browser focus behavior.

app/src/sandbox/tab-4213/test-browser.ts
```diff
@@ -0,0 +1,55 @@
...
+    await page.keyboard.press("ArrowRight");
+
+    await test.expect(q.tab("Fruits")).toBeFocused();
+  });
+
+  test("does not move focus after controlled tab selection outside the tab list", async ({
+    page,
+    q,
+  }) => {
+    const button = page.getByRole("button", { name: "Select vegetables" });
```
diegohaz: Should use `q.button("Select vegetables")`

app/src/sandbox/tab-4213/test-browser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    await test.expect(q.tab("Meat")).toHaveAttribute("aria-selected", "true");
+    await test.expect(button).toBeFocused();
+  });
+
+  test("does not move focus after controlled tab selection inside a tab panel", async ({
+    page,
+    q,
+  }) => {
+    const textbox = page.getByRole("textbox", { name: "Fruit note" });
```
diegohaz: `q.textbox`

app/src/sandbox/tab-4213/index.react.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </ak.TabPanel>
+      <ak.TabPanel tabId="vegetables">
+        Carrots, onions, and potatoes
+      </ak.TabPanel>
+      <ak.TabPanel tabId="meat">Beef, chicken, and pork</ak.TabPanel>
+      <ak.TabPanel tabId="dairy">Milk, cheese, and yogurt</ak.TabPanel>
+      {/* WebKit keeps focus on the clicked button only when explicitly tabbable. */}
+      <button
+        type="button"
+        tabIndex={0}
```
diegohaz:
```suggestion
      <button
        type="button"
        // WebKit keeps focus on the clicked button only when explicitly tabbable.
        tabIndex={0}
```

app/src/sandbox/tab-4213/test-browser.ts
diegohaz: Fixed in 66dab50df - replaced the raw Playwright role locator with `q.button("Select vegetables")` in the sandbox regression test.

app/src/sandbox/tab-4213/test-browser.ts
diegohaz: Fixed in 66dab50df - replaced the raw textbox role locator with `q.textbox("Fruit note")`.

app/src/sandbox/tab-4213/index.react.tsx
diegohaz: Fixed in 66dab50df - moved the WebKit focus note onto the button prop, directly above `tabIndex={0}`.

diegohaz:
**Fixed**

- 66dab50df ([reply](https://github.com/ariakit/ariakit/pull/6068#discussion_r3321421725)): Switched the outside-button assertion to `q.button("Select vegetables")`.
- 66dab50df ([reply](https://github.com/ariakit/ariakit/pull/6068#discussion_r3321421927)): Switched the panel input assertion to `q.textbox("Fruit note")`.
- 66dab50df ([reply](https://github.com/ariakit/ariakit/pull/6068#discussion_r3321422057)): Moved the WebKit focus note directly above `tabIndex={0}`.

diegohaz:
**CI follow-up**

- 0328e1599: Fixed the failed Type check by removing the now-unused `page` fixture from the outside-button regression test after switching that locator to `q.button("Select vegetables")`.

PR #6069, opened 2026-05-28
Fix Dialog onClose with controlled store

### LINKED ISSUE #3488: `onClose` gets called on Dialog even when `open` state is changed programmatically
### Current behavior

`onClose` event gets triggered whenever `open` state passed to `useDialogStore` is changed programmatically. This makes it impossible to distinguish between user-initiated close events (e.g. hitting Esc key or clicking outside the dialog) and programmatic changes to the dialog state, and, as a result, implement certain patterns, such as displaying a warning on dialog hide.

Note that this behaviour is not observed when using `open` prop passed directly to the `Dialog` component, only when using custom dialog store. [Existing example](https://ariakit.org/examples/dialog-hide-warning) on displaying a warning on dialog hide is written using `open` prop and thus works.

1. Open https://stackblitz.com/edit/lzpywb?file=dialog%2Fstyle.css,dialog%2Findex.tsx&theme=dark
2. Open a dialog
3. Try closing a dialog

### Expected behavior

Dialog should close, `onClose` callback should only be called once.

### Workaround

There is a number of workarounds, e.g.
- not using dialog store (which I can't do because of #3403)
- overriding behaviour of `setOpen` (which glitches, see #3402, and a result, can only be used when dialog has animation)

I have a custom Dialog component built around Ariakit's. This essentially means that I have to have 2 implementations depending on whether I use animations or not.

### Possible solutions

_No response_

diegohaz:
## Motivation

`Dialog` calls `onClose` again when a controlled `open` value passed through `useDialogStore` changes as part of handling a close. This can keep the dialog open and prevents consumers from distinguishing user-initiated closes from programmatic state updates.

## Solution

The `Dialog` close event now runs only for Ariakit's built-in dismiss paths, such as `DialogDismiss`, Escape, outside interactions, and related popover dismissers. Programmatic controlled store updates, including `setOpen(false)` calls from inside `onClose`, no longer re-dispatch `onClose`.

The fix routes built-in dismiss behavior through a scoped `hide` wrapper that dispatches the cancelable `close` event before hiding. Scoped dismissers still respect `event.preventDefault()`, avoid stale outside-interaction state after canceled closes, and no-op while the store is already closing during animated leave.

The regression sandbox covers dismiss button, Escape, outside click, explicit store dismissers, animated leave, and popover-derived dismissers.

## Workaround

When using a controlled `useDialogStore` with an `onClose` handler that calls `event.preventDefault()`, defer the controlled `setOpen(false)` call and guard the programmatic follow-up close event:

```tsx
const closingRef = useRef(false);
const dialog = Ariakit.useDialogStore({ open, setOpen });

<Ariakit.Dialog
  store={dialog}
  onClose={(event) => {
    // TODO: Remove this guard once
    // https://github.com/ariakit/ariakit/issues/3488 is fixed. Deferring
    // the controlled close keeps the programmatic update from running
    // this handler again.
    if (closingRef.current) return;
    event.preventDefault();
    closingRef.current = true;
    queueMicrotask(() => {
      setOpen(false);
      closingRef.current = false;
    });
  }}
/>
```

### DIFF
--- a/.changeset/300-dialog-controlled-store-close.md
+++ b/.changeset/300-dialog-controlled-store-close.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Dialog`](https://ariakit.com/reference/dialog) so programmatic controlled store updates don't trigger [`onClose`](https://ariakit.com/reference/dialog#onclose).
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) so [`onClose`](https://ariakit.com/reference/dialog#onclose) only fires for Ariakit's built-in dismiss mechanisms, such as [`DialogDismiss`](https://ariakit.com/reference/dialog-dismiss), pressing <kbd>Esc</kbd>, or interacting outside the dialog, and no longer fires for programmatic visibility changes like a controlled [`open`](https://ariakit.com/reference/dialog#open) prop changing, `setOpen(false)`, or `store.hide()`.
--- a/.changeset/300-dialog-controlled-store-close.md
+++ b/.changeset/300-dialog-controlled-store-close.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) so programmatic controlled store updates don't trigger [`onClose`](https://ariakit.com/reference/dialog#onclose).
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -52,6 +52,26 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .toBeHidden();
     await test.expect(q.text("Close count: 4")).toBeVisible();
 
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await q.button("Close programmatically with setOpen").click();
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 4")).toBeVisible();
+
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await q.button("Close programmatically with context").click();
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 4")).toBeVisible();
+
     await q.button("Show popover").click();
     await test.expect(q.dialog("Popover content")).toBeVisible();
 
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -1,0 +1,18 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("controlled store dialog calls onClose once when dismissed", async ({
+    q,
+  }) => {
+    // See https://github.com/ariakit/ariakit/issues/3488
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await q.button("OK").click();
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 1")).toBeVisible();
+  });
+});
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -85,12 +85,22 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await q.button("Show popover").click();
     await test.expect(q.dialog("Popover content")).toBeVisible();
 
-    await q.button("Close popover with explicit store").click();
+    await q.button("Close popover").click();
 
     await test
       .expect(q.dialog("Popover content", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Popover close count: 1")).toBeVisible();
+
+    await q.button("Show popover").click();
+    await test.expect(q.dialog("Popover content")).toBeVisible();
+
+    await q.button("Close popover with explicit store").click();
+
+    await test
+      .expect(q.dialog("Popover content", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Popover close count: 2")).toBeVisible();
 
     await q.button("Show parent dialog").click();
     await test.expect(q.dialog("Parent dialog")).toBeVisible();
@@ -115,5 +125,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .expect(q.dialog("Parent dialog", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Parent close count: 1")).toBeVisible();
+    await test.expect(q.text("Child close count: 1")).toBeVisible();
   });
 });
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -12,6 +12,12 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await q.button("OK").click();
 
     await test
+      .expect(q.dialog("Success"))
+      .toHaveAttribute("data-leave", "true");
+    await page.keyboard.press("Escape");
+    await test.expect(q.text("Close count: 1")).toBeVisible();
+
+    await test
       .expect(q.dialog("Success", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Close count: 1")).toBeVisible();
@@ -25,5 +31,35 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .expect(q.dialog("Success", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Close count: 2")).toBeVisible();
+
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await page.mouse.click(10, 10);
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 3")).toBeVisible();
+
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await q.button("Close with explicit store").click();
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 4")).toBeVisible();
+
+    await q.button("Show popover").click();
+    await test.expect(q.dialog("Popover content")).toBeVisible();
+
+    await q.button("Close popover with explicit store").click();
+
+    await test
+      .expect(q.dialog("Popover content", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Popover close count: 1")).toBeVisible();
   });
 });
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -2,6 +2,7 @@ import { withFramework } from "#app/test-utils/preview.ts";
 
 withFramework(import.meta.dirname, async ({ test }) => {
   test("controlled store dialog calls onClose once when dismissed", async ({
+    page,
     q,
   }) => {
     // See https://github.com/ariakit/ariakit/issues/3488
@@ -14,5 +15,15 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .expect(q.dialog("Success", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Close count: 1")).toBeVisible();
+
+    await q.button("Show modal").click();
+    await test.expect(q.dialog("Success")).toBeVisible();
+
+    await page.keyboard.press("Escape");
+
+    await test
+      .expect(q.dialog("Success", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Close count: 2")).toBeVisible();
   });
 });
--- a/app/src/sandbox/dialog-3488/test-browser.ts
+++ b/app/src/sandbox/dialog-3488/test-browser.ts
@@ -72,6 +72,16 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .toBeHidden();
     await test.expect(q.text("Close count: 4")).toBeVisible();
 
+    await q.button("Show external dialog").click();
+    await test.expect(q.dialog("External dialog")).toBeVisible();
+
+    await q.button("Close external dialog").click();
+
+    await test
+      .expect(q.dialog("External dialog", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("External close count: 1")).toBeVisible();
+
     await q.button("Show popover").click();
     await test.expect(q.dialog("Popover content")).toBeVisible();
 
@@ -81,5 +91,29 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .expect(q.dialog("Popover content", { includeHidden: true }))
       .toBeHidden();
     await test.expect(q.text("Popover close count: 1")).toBeVisible();
+
+    await q.button("Show parent dialog").click();
+    await test.expect(q.dialog("Parent dialog")).toBeVisible();
+
+    await q.button("Show child dialog").click();
+    await test.expect(q.dialog("Child dialog")).toBeVisible();
+
+    await q.button("Close child").click();
+
+    await test
+      .expect(q.dialog("Child dialog", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Child close count: 1")).toBeVisible();
+    await test.expect(q.dialog("Parent dialog")).toBeVisible();
+
+    await q.button("Show child dialog").click();
+    await test.expect(q.dialog("Child dialog")).toBeVisible();
+
+    await q.button("Close parent from child").click();
+
+    await test
+      .expect(q.dialog("Parent dialog", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Parent close count: 1")).toBeVisible();
   });
 });
--- a/packages/ariakit-react-components/src/dialog/dialog-dismiss.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog-dismiss.tsx
@@ -9,24 +9,12 @@ import type { ElementType, MouseEvent } from "react";
 import { useMemo } from "react";
 import type { ButtonOptions } from "../button/button.tsx";
 import { useButton } from "../button/button.tsx";
-import {
-  useDialogDismissContext,
-  useDialogScopedContext,
-} from "./dialog-context.tsx";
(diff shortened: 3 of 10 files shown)


packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -687,13 +692,16 @@ export interface DialogOptions<T extends ElementType = TagName>
...
+   * controlled [`open`](https://ariakit.com/reference/dialog#open) prop value
+   * changes, the store
    * [`open`](https://ariakit.com/reference/use-dialog-store#open) state is set
-   * to `false`. If the controlled
-   * [`open`](https://ariakit.com/reference/dialog#open) prop value changes, or
-   * if the dialog's visibility is altered in any other way (such as unmounting
-   * the dialog without adjusting the open state), this event won't be
-   * triggered.
+   * programmatically, or if the dialog's visibility is altered in any other way
+   * (such as unmounting the dialog without adjusting the open state), this
```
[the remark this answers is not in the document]
diegohaz: Fixed in e25e4578d - clarified that `onClose` only fires for Ariakit dismiss mechanisms, documented controlled `open`, `setOpen(false)`, and `store.hide()` as programmatic paths, and added regression coverage for programmatic closes through `setOpen(false)` and `useDialogContext().hide()`.

diegohaz:
**Fixed**
- e25e4578d ([reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3319809278)): clarified the `onClose` docs for controlled/programmatic close paths, kept public `hide()` programmatic through an internal dismiss context, and added regression coverage for `setOpen(false)` and `useDialogContext().hide()`.

packages/ariakit-react-components/src/popover/popover.tsx
```diff
@@ -477,6 +467,16 @@ export const usePopover = createHook<TagName, PopoverOptions>(
...
 
+    props = useWrapElement(
+      props,
+      (element) => (
+        <PopoverScopedContextProvider value={store}>
+          {element}
+        </PopoverScopedContextProvider>
+      ),
+      [store],
+    );
```
diegohaz: **This `PopoverScopedContextProvider` reorder looks unnecessary for the final fix and can likely be dropped.**

It was introduced in `3a2fe61` alongside the original approach, where the dialog scoped context itself carried the `hide` wrapper (`DialogScopedContextProvider value={scopedStore}`). There the order mattered: popover re-provides `DialogScopedContextProvider value={store}`, the innermost provider wins, so the wrapper-carrying store had to be moved outward to win.

The final design (`e25e457`) changed that — the scoped context now carries the raw `store`, and the wrapper is delivered through the separate `DialogDismissContext`. `useDialogDismiss` resolves it via `hasSameContentElement(store, context)`, which lands on the wrapper regardless of which `DialogScopedContextProvider` is innermost. So provider ordering no longer affects dismiss/`onClose`.

Evidence it isn't needed:
- `Hovercard`/`Menu`/`Select`/`Combobox` keep their scoped providers *inside* the popover/dialog call (e.g. `hovercard.tsx` wraps `HovercardScopedContextProvider` before `usePopover`) and were not reordered, yet they rely on the same `hasSameContentElement` routing.
- Reverting this hunk to the base order and running the repro `dialog-3488` (incl. `PopoverDismiss store={popover}` → `onClose`) plus 30 popover/menu/select/combobox/tooltip sandbox tests — all pass.

A subtle side effect of keeping the reorder: `useDialogScopedContext()` for popover descendants now resolves to the internal wrapped dialog store instead of the consumer popover store (the base behavior). It's functionally equivalent (synced), but it's an unexplained change with no demonstrated benefit.

If this addresses a specific case I'm not seeing, could you add a regression test for it? Otherwise consider dropping the hunk to keep the diff minimal.

packages/ariakit-react-components/src/dialog/dialog-dismiss.tsx
```diff
@@ -31,14 +43,17 @@ type HTMLType = HTMLElementTagNameMap[TagName];
 export const useDialogDismiss = createHook<TagName, DialogDismissOptions>(
   function useDialogDismiss({ store, ...props }) {
     const context = useDialogScopedContext();
+    const dismiss = useDialogDismissContext();
     store = store || context;
+    const hide =
+      hasSameContentElement(store, context) && dismiss ? dismiss : store?.hide;
```
diegohaz: **Heads-up (please confirm intended): a `DialogDismiss` with an explicit `store` pointing to a dialog whose subtree doesn't contain the button no longer fires that dialog's `onClose`.**

`dismiss` comes from the nearest `DialogDismissContext`, which only exists inside the target dialog's own React subtree. When `store` is an explicit store for a *different* dialog, `hasSameContentElement(store, context)` is `false`, so `hide` falls back to the raw `store?.hide()` — which, after this PR, does not dispatch the `close` event.

Concretely, a "close parent" button like `<DialogDismiss store={parent} />` rendered inside a nested child dialog used to fire `parent`'s `onClose` (the old `setOpen` interception fired `onClose` on any `store.hide()`); now it closes `parent` silently.

This also makes the updated `onClose` docs slightly imprecise: they state `onClose` fires for `DialogDismiss` unconditionally, but it only fires when the dismiss is within the dialog's subtree (or shares its `contentElement`).

It's a niche pattern (no example or sandbox uses a cross-dialog dismiss store today), and there's no clean way to reach a foreign dialog's wrapper through React context — so this may be an acceptable limitation. If so, please confirm it's intentional and consider documenting the constraint alongside the `onClose` / `DialogDismiss` `store` docs so the behavior change is discoverable.

.changeset/300-dialog-controlled-store-close.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) so programmatic controlled store updates don't trigger [`onClose`](https://ariakit.com/reference/dialog#onclose).
```
diegohaz: **The changeset wording is narrower than the actual behavior change.**

This PR stops `onClose` from firing for *all* programmatic hides, not only controlled-store updates. Even with an uncontrolled `useDialogStore`, calling `store.hide()` / `setOpen(false)` previously dispatched `onClose` (via the old `setOpen` interception) and no longer does.

Consider broadening the entry so consumers who relied on `onClose` for cleanup on programmatic hides aren't surprised by the changelog, e.g.:

> Fixed [`Dialog`](https://ariakit.com/reference/dialog) so [`onClose`](https://ariakit.com/reference/dialog#onclose) only fires for Ariakit's built-in dismiss mechanisms (the dismiss button, <kbd>Esc</kbd>, or interacting outside) and no longer for programmatic visibility changes such as a changed controlled `open` prop, `setOpen(false)`, or `store.hide()`.

diegohaz (commented):
## Review summary

**Type:** Initial review (no prior review from this workflow). **Baseline:** merge-base `240d5ad` → head `e25e457`. **Scope:** full diff plus surrounding context — `dialog.tsx`, `dialog-dismiss.tsx`, `dialog-context.tsx`, `popover.tsx`, the store internals (`ariakit-store`, `react-store`), `useHideOnInteractOutside`, the `dialog-3488` sandbox, the changeset, and the affected `onClose` examples.

### Overall
The fix is well-targeted and the design is sound. Replacing the `useDialogStore({ setOpen })` interception with an explicit `hide` wrapper that only the built-in dismiss paths invoke (Escape, interact-outside via the scoped store, `DialogDismiss` via `DialogDismissContext`, and the hidden screen-reader dismiss) cleanly fixes #3488: programmatic state changes no longer re-dispatch `onClose`. Two design choices are justified by the tests: the **separate `DialogDismissContext`** (keeps `useDialogContext().hide()` a raw, no-`onClose` programmatic hide) and **`hasSameContentElement`** (lets an explicit `store` that is the same dialog as context still fire `onClose`). Typing is correct (`{ ...store, hide }` is safe — store methods are own, non-`this` properties).

### Validation performed
- `pnpm tsc` — clean.
- `pnpm -F app test dialog-3488` — passes (all 3 browsers; first cold run was flaky on sandbox compile only).
- Reverted the `popover.tsx` hunk and re-ran `dialog-3488` + 30 popover/menu/select/combobox/tooltip sandbox tests — all pass (basis for the popover finding below).
- `dialog-3403`, `dialog-5057`, `dialog-disclosure-5224`, `menu-4904`, `menu-dialog-close` against the PR as-is — all pass.
- Confirmed no internal code depends on `onClose`/the `close` event; it's purely consumer-facing.

### Findings (all inline)
1. **`popover.tsx` scoped-provider reorder appears unnecessary** and looks like leftover from the original scoped-store approach (`3a2fe61`); the final `DialogDismissContext` design routes correctly regardless of order, and sibling components (Hovercard/Menu/Select/Combobox) were not reordered. Tests pass with it reverted. Consider dropping it (or adding a regression test if it's load-bearing).
2. **`DialogDismiss` with an explicit cross-subtree `store`** no longer fires that dialog's `onClose` (the wrapper is only reachable inside the dialog's own subtree). Niche and unused in-repo, but it's a behavior change and the `onClose` docs imply `DialogDismiss` always fires it — please confirm intended and document the limitation.
3. **Changeset wording** ("programmatic controlled store updates") is narrower than the actual change — `onClose` no longer fires for any programmatic hide, including `store.hide()`/`setOpen(false)` on uncontrolled stores. Suggest broadening it.

### Not fully validated
- Example browser tests under `examples/` (not collected by the `app` workspace test runner here); I relied on sandbox coverage + the unchanged `onClose` examples (`dialog-hide-warning*`) reasoning instead.
- Whether the `onClose` semantic change warrants more than a `patch` bump — it follows Ariakit's bug-fix convention, so I'm treating it as acceptable, but flagging for maintainer judgment given it changes a public-API behavior.

packages/ariakit-react-components/src/dialog/dialog-dismiss.tsx
diegohaz: DialogDismiss click should fire onClose.

packages/ariakit-react-components/src/popover/popover.tsx
diegohaz: Fixed in 679531b71 - dropped the `PopoverScopedContextProvider` reorder from the final PR diff. The dismiss routing now goes through the dialog dismiss registry instead, and `packages/ariakit-react-components/src/popover/popover.tsx` matches `origin/main` again.

packages/ariakit-react-components/src/dialog/dialog-dismiss.tsx
diegohaz: Fixed in 679531b71 - `DialogDismiss` now resolves the registered dismiss handler by target store at click time, so an explicit `store` can fire the target dialog's `onClose` even when the button is outside that dialog subtree. I added coverage for both a nested `<DialogDismiss store={parent}>` and an explicit-store dismiss button mounted before its dialog registers.

.changeset/300-dialog-controlled-store-close.md
diegohaz: Fixed in 679531b71 - broadened the changeset to say `onClose` only fires for Ariakit's built-in dismiss mechanisms and no longer fires for programmatic visibility changes, including controlled `open` changes, `setOpen(false)`, and `store.hide()`.

diegohaz:
**Fixed**

- Dropped the now-unneeded `PopoverScopedContextProvider` reorder so `popover.tsx` matches the base order again. [Reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3320321818)
- Updated `DialogDismiss` routing so explicit-store dismissers resolve the target dialog's registered dismiss handler at click time, including nested and pre-mounted explicit-store cases. [Reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3320322384)
- Broadened the changeset wording to cover built-in dismiss mechanisms versus programmatic visibility changes like controlled `open`, `setOpen(false)`, and `store.hide()`. [Reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3320322759)

app/src/sandbox/dialog-3488/test-browser.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    await q.button("Show child dialog").click();
+    await test.expect(q.dialog("Child dialog")).toBeVisible();
+
+    await q.button("Close parent from child").click();
+
+    await test
+      .expect(q.dialog("Parent dialog", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("Parent close count: 1")).toBeVisible();
```
diegohaz, line 127: **Optional test hardening:** this "close parent from child" block is the strongest regression guard in the PR, but it never asserts that the **child's** `onClose` did *not* re-fire.

The parent's `onClose` does `setChildOpen(false)` — a programmatic controlled `open → false` transition on the still-open child. The entire point of #3488 is that such programmatic changes must not route through the cancelable `hide` wrapper and therefore must not increment `childCloseCount`. Today they don't, but nothing here pins that invariant: a future regression that re-fires `onClose` on a controlled close (exactly the bug class being fixed) would still pass this test.

Since `childCloseCount` is already `1` at this point (asserted at line 106, and reopening the child at line 109 doesn't fire `onClose`), one extra line closes the gap:

```ts
    await test
      .expect(q.dialog("Parent dialog", { includeHidden: true }))
      .toBeHidden();
    await test.expect(q.text("Parent close count: 1")).toBeVisible();
+   // The child closes via the parent's programmatic setChildOpen(false), so
+   // its onClose must NOT fire again.
+   await test.expect(q.text("Child close count: 1")).toBeVisible();
```

app/src/sandbox/dialog-3488/test-browser.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    await test
+      .expect(q.dialog("External dialog", { includeHidden: true }))
+      .toBeHidden();
+    await test.expect(q.text("External close count: 1")).toBeVisible();
+
+    await q.button("Show popover").click();
+    await test.expect(q.dialog("Popover content")).toBeVisible();
+
+    await q.button("Close popover with explicit store").click();
```
diegohaz: **Optional coverage gap (low priority):** the popover scenario only exercises an *explicit-store* dismiss (`<PopoverDismiss store={popover}>`), so the **context-resolved** (no-store) wrapper-dismiss path that fires `onClose` is still unverified for Popover/Menu/Select/Combobox.

That path is the most fragile part of the routing: a no-store `<PopoverDismiss />` resolves its target via `usePopoverScopedContext()` → the `DialogScopedContextProvider` that `PopoverScopedContextProvider` re-provides, and only then does `getDialogDismiss(store)` find the wrapper `useDialog` registered as `sourceStore`. It depends on both the scoped-context re-provider chain and the wrap ordering in `usePopover` relative to `useDialog` — exactly what the explicit-store test bypasses (the plain no-store `DialogDismiss` "OK" button covers only the non-wrapped Dialog case). If that wiring ever shifts, dismiss-fires-`onClose` for these wrapper components would regress silently, and nothing in the repo asserts the Ariakit `onClose` event for a context-resolved wrapper dismiss.

Consider adding a no-store `<PopoverDismiss />` (or `<MenuDismiss />`) inside a Popover/Menu that carries the Ariakit `onClose` event prop and asserting its close count increments, mirroring the explicit-store case already here. Near-zero cost, and it pins the resolution path this PR exists to protect.

diegohaz (commented):
## Re-review summary

**Type:** Re-review. **Baseline:** my last reviewed SHA `e25e457` → current head `679531b7` (commit "Address DialogDismiss review feedback"). Scope: the incremental diff plus the surrounding registry mechanism and its regression surface.

### Prior findings — all three resolved and verified against the current code
1. **Popover reorder** — dropped; `popover.tsx` is now byte-identical to `origin/main` (`git diff origin/main` is empty). ✅
2. **Cross-subtree `DialogDismiss store={…}` not firing `onClose`** — fixed properly rather than documented away. The React `DialogDismissContext` was replaced with a module-level `WeakMap<DialogStore, hide>` registry (`setDialogDismiss`/`getDialogDismiss`); `useDialog` registers its cancelable `hide` wrapper for both the internal store and the consumer `sourceStore`, and `DialogDismiss` resolves the handler by target store at click time. A cross-subtree `<DialogDismiss store={parent}>` and an explicit-store dismiss rendered outside its dialog now both fire the target's `onClose`. The `onClose` JSDoc claim ("DialogDismiss … fires") is now accurate without caveat. ✅
3. **Changeset wording** — broadened to cover all programmatic hides (controlled `open`, `setOpen(false)`, `store.hide()`). ✅

### Mechanism review (adversarial, multi-agent)
I ran a fan-out review (registry lifecycle, click-time dismiss resolution, regression surface, semantics/tests, docs/simplicity) with adversarial verification of every candidate finding. **No code defects survived verification.** Specifically confirmed sound: WeakMap → no leak + explicit cleanup with a last-writer guard (`dismisses.get(store) !== hide`); registering both the internal store and `sourceStore` is what keeps context-resolved dismiss working in Popover/Menu without the reorder; click-time resolution with graceful `store.hide()` fallback when a target isn't registered; helpers are internal (not exported); `WeakMap` has codebase precedent (`utils/orchestrate.ts`). Refuted candidates included the raw-`hide` fallback "silently skipping onClose" (intended/graceful), shared-store last-writer-wins (unsupported pattern, no worse than base), and a "redundant guard" nit.

### Validation performed
- `pnpm tsc` — clean.
- Updated `dialog-3488` test (external cross-subtree dismiss + nested parent/child) — passes on chrome/firefox/safari.
- Regression sweep: `dialog-3403`, `dialog-5057`, `dialog-disclosure-5224`, `menu-dialog-close`, `menu-4904`, `menu-nested-combobox-hide`, `dialog-combobox-4825`, `select-combobox-5047`, `tooltip-865` — 51 tests pass, no regressions.
- (A transient HTTP 500 on the dev server was a Vite dep-optimizer cache corruption from restarting the server — unrelated to the PR; cleared `app/node_modules/.vite` and all tests went green.)

### New findings (2, both low / optional test hardening — inline)
- **Nested "close parent from child"** asserts the parent count but not that the **child's** `onClose` stayed put after the parent's programmatic `setChildOpen(false)` — one extra assertion pins the exact #3488 invariant.
- **Context-resolved (no-store) wrapper dismiss** isn't covered: the popover case only tests an explicit `store`, leaving the scoped-context-resolution path (the most fragile part of the routing) unasserted for Popover/Menu/Select/Combobox `onClose`.

### Overall
The new approach is a clean, correct improvement over the previous revision and fully addresses the earlier feedback. No blocking issues; the two inline notes are optional regression-guard hardening. Areas not independently re-run: the `examples/` browser suites (not collected by the `app` workspace runner) and visual tests (CI-only).

app/src/sandbox/dialog-3488/test-browser.ts
diegohaz: Fixed in 9e9d6954b - moved the child close count outside the unmounted parent dialog and added the post-parent-dismiss assertion that it stays at `1`, so the parent's programmatic `setChildOpen(false)` path is pinned not to re-fire child `onClose`.

app/src/sandbox/dialog-3488/test-browser.ts
diegohaz: Fixed in 9e9d6954b - added a no-store `PopoverDismiss` path before the explicit-store dismiss path and asserted `onClose` increments for both, covering the context-resolved wrapper dismiss routing.

diegohaz:
**Fixed**

- 9e9d6954b ([reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3320824723)): Added the post-parent-dismiss assertion that `Child close count` stays at `1`, covering the programmatic child close path.
- 9e9d6954b ([reply](https://github.com/ariakit/ariakit/pull/6069#discussion_r3320824852)): Added no-store `PopoverDismiss` coverage before the explicit-store popover dismiss path, so context-resolved wrapper dismiss routing is covered too.

diegohaz:
Closing this as I don't think it's the right fix. Will revisit it later.

PR #6072, opened 2026-05-28
Add generated API reference docs for more packages

diegohaz:
## Motivation

[#6070](https://github.com/ariakit/ariakit/pull/6070) added an `ariakit docs` generator and a generated API reference for `@ariakit/utils`, but the other internal packages still shipped a public API with no reference in their readmes. Maintaining those references by hand would be error-prone — signatures should come from TypeScript and descriptions from JSDoc — and there was no automation to keep them in sync.

## Solution

Generate an API reference in the readmes of `@ariakit/react-utils`, `@ariakit/store`, `@ariakit/react-store`, `@ariakit/solid-utils`, `@ariakit/solid-store`, and `@ariakit/test` using the existing `ariakit docs` command, and keep them in sync automatically.

- **Generated docs.** Each readme gained Installation/Usage sections and a generated API reference block. The multi-file `@ariakit/react-utils` and `@ariakit/solid-utils` use file-level `@module` tags to group their exports, matching `@ariakit/utils`.
- **Table of contents.** Each generated API reference lists its members (functions, types, etc.) in the table of contents — nested under their module when modules are present, flat otherwise. Headings use incremental levels (`###` members under a `##` section, or `####` members under a `### module`), and the links use a GitHub-compatible slugger so anchors stay correct even when slugs collide (e.g. `wrapInstance`/`WrapInstance`).
- **Multiple sections per readme.** `@ariakit/test` is multi-entry, so the `ariakit docs` command was extended with `--marker` (named marker blocks), `--heading` (custom heading), and `--exclude` (omit names re-exported from a given file or package). Its readme documents the main entry plus `@ariakit/test/react` (`--exclude src/index.ts`) and `@ariakit/test/playwright` (`--exclude @playwright/test`) in separate keyed blocks. The default usage stays backward compatible with `@ariakit/utils`.
- **Explicit return types in `@ariakit/test`.** The `query` exports in `query.ts` and `playwright.ts` had enormous inferred object types (the `AriaRole` union times the nested query methods) that expanded to thousands of lines of generated docs. They now declare explicit, structurally identical return types so the generator prints a named type instead. This is a non-breaking, type-only change.
- **Root `docs` script.** `pnpm run docs` runs the `docs` script for every package under `packages/` that defines one. It must be invoked as `pnpm run docs` because `docs` is a built-in pnpm command.
- **CI.** A new `docs` workflow regenerates the readmes and commits them back to the PR branch (or opens a maintenance PR on `main`), mirroring the existing `build-styles` workflow that keeps generated styles in sync.

## Verification

- `pnpm tsc`
- `pnpm test packages/ariakit-scripts/src/docs.test.ts`
- `pnpm lint`
- `pnpm run docs` (idempotent)

## Notes

- No changeset: the only changes to published packages are generated docs, a `docs` script, and non-breaking type annotations — consistent with #6070.
- Known generator limitations left as-is: oversized referenced helper types (e.g. `RefStore` in `@ariakit/solid-utils`) are line-truncated; the pre-existing `DirtiableElement` type name in `@ariakit/test` surfaces in the generated docs; and the slug disambiguation is scoped per marker block, so it can't number anchors across the multiple blocks in `@ariakit/test` (harmless today since the only cross-block duplicate, `query`, lives in a single-entry block with no table of contents).


packages/ariakit-test/package.json
```diff
@@ -22,7 +22,8 @@
   "sideEffects": false,
   "scripts": {
     "build": "ariakit build",
-    "clean": "ariakit clean"
+    "clean": "ariakit clean",
+    "docs": "ariakit docs --write && ariakit docs --write --entry src/react.tsx --marker react --heading 'React API reference' --exclude src/index.ts && ariakit docs --write --entry src/playwright.ts --marker playwright --heading 'Playwright API reference' --exclude @playwright/test"
```
diegohaz: Split these into separate scripts like `docs-react`, `docs-playwright`, then consolidate them in a single `docs` script.

package.json
```diff
@@ -58,6 +58,7 @@
     "build-website": "pnpm -F website run build",
     "build-website-lite": "cross-env DISABLE_SHIKI=true pnpm run build-website",
     "clean": "pnpm -r run --if-present clean",
+    "docs": "pnpm -F \"./packages/*\" run --if-present docs",
```
diegohaz: Can we use simple quotes around `./packages/*` so we don't need escaping?

diegohaz (commented):
Initial review of `ariakit/ariakit#6072` at `7b538300bbd3f7d36dfff561b91120949d129b9c`, diffed from merge base `712277e33f2ed5db37667248bf259a1194027e3c`.

Scope reviewed: docs generator changes in `packages/ariakit-scripts`, the new docs workflow, package `docs` scripts, generated README sections, and the explicit `@ariakit/test` return types used to keep generated docs readable.

Validation run locally:

- `pnpm run docs` (idempotent; worktree stayed clean)
- `pnpm test packages/ariakit-scripts/src/docs.test.ts` (23 tests passed)
- `pnpm tsc`

I did not find substantive issues worth an inline comment.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          # Check out the PR branch tip (not the merge ref) so `commit-or-pr`
+          # can fetch and rebase onto `origin/$HEAD_REF` cleanly. Otherwise
+          # the merge ref includes commits from `main` that are absent from
+          # the PR branch, which causes rebase conflicts.
+          ref: ${{ github.head_ref || github.ref_name }}
+          persist-credentials: false
```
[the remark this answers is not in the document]
diegohaz: This intentionally mirrors the existing `build-styles.yml` write-back workflow, which uses the same `ref: ${{ github.head_ref || github.ref_name }}` checkout and the same `WRITEBACK_ALLOWED` gating — those workflows deliberately assume PRs come from branches on this repository rather than forks. Drift on fork PRs is instead caught by the `docs.test.ts` sync test, which runs on every PR. Reworking the fork-handling strategy should apply to all of these write-back workflows together, so it is out of scope for this PR.

packages/ariakit-test/src/query.ts
```diff
@@ -9,11 +9,20 @@ import { queries as baseQueries } from "@testing-library/dom";
...
+type LabeledQuery = ReturnType<typeof createLabeledQuery>;
+type RoleQueries = Record<AriaRole, Query>;
 type ElementQueries = ReturnType<
   typeof getQueriesForElement<typeof baseQueries>
 >;
 
+interface QueryObject extends RoleQueries {
+  text: TextQuery;
+  labeled: LabeledQuery;
+  within: (element?: HTMLElement | null) => QueryObject;
```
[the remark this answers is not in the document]
diegohaz: The nullable parameter is intentional: `within` accepts a possibly-null element and the runtime `invariant` surfaces a clear "Unable to create queries for null element" error. Tightening the type to a non-null `HTMLElement` is a breaking change for existing callers — e.g. `examples/select-combobox-tab/test.ts` passes an `HTMLElement | null` and relies on this, and `pnpm tsc` fails when the type is tightened. Keeping the parameter optional matches the runtime contract and the previously-shipped inferred type.

packages/ariakit-test/package.json
diegohaz: Fixed in bc540488c — split into `docs-react` and `docs-playwright` sub-scripts consolidated under a single `docs` script.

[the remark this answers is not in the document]
diegohaz: This repo uses single-quoted args in package.json scripts by convention (e.g. `lint-css`, `dev`, and now the root `docs` filter) and targets Unix shells — husky, pnpm, and bash-based CI — rather than Windows `cmd`/`powershell`. Keeping single quotes here for consistency with that convention and with the maintainer preference on the root `docs` script.

package.json
diegohaz: Fixed in bc540488c — the root `docs` script now uses single quotes around `./packages/*` so no escaping is needed.

diegohaz:
Addressed the review feedback plus the requested nested table of contents.

Commit 283441508 implements the table-of-contents request: each generated API reference now lists its members (functions, types, etc.) — nested under their module when modules are present — with incremental heading levels (`###` members under a `##` section, `####` under a `### module`) and a GitHub-compatible slugger so anchors stay correct when slugs collide.

**Fixed**

- bc540488c ([thread](https://github.com/ariakit/ariakit/pull/6072#discussion_r3320222790)): split the `@ariakit/test` `docs` script into `docs-react`/`docs-playwright` consolidated under a single `docs` script.
- bc540488c ([thread](https://github.com/ariakit/ariakit/pull/6072#discussion_r3320225937)): the root `docs` script now uses single quotes around `./packages/*`, so no escaping is needed.

**Won't fix (with reasoning)**

- ([thread](https://github.com/ariakit/ariakit/pull/6072#discussion_r3320216681)): the `docs.yml` checkout intentionally mirrors the existing `build-styles.yml` write-back pattern (same `ref` and `WRITEBACK_ALLOWED` gating, which assume same-repo PR branches). Fork-PR drift is caught by the `docs.test.ts` sync test that runs on every PR, and reworking fork handling belongs to all the write-back workflows together — out of scope here.
- ([thread](https://github.com/ariakit/ariakit/pull/6072#discussion_r3320216724)): `query`'s `within` keeps its nullable parameter. It matches the runtime `invariant` ("Unable to create queries for null element"), and tightening it to a non-null `HTMLElement` breaks existing callers — e.g. `examples/select-combobox-tab/test.ts` passes `HTMLElement | null` — making `pnpm tsc` fail.
- ([thread](https://github.com/ariakit/ariakit/pull/6072#discussion_r3320223410)): the single-quoted `--heading` args follow the repo's existing package.json script convention (`lint-css`, `dev`, the root `docs` filter) and its Unix-shell toolchain (husky, pnpm, bash CI), so they're kept for consistency with the maintainer's preference on the root script rather than switched to escaped double quotes.

PR #6074, opened 2026-05-29
Speed up the perf workflow's Chrome and Node jobs

diegohaz:
## Motivation

The `perf` workflow's Chrome and Node jobs are the long pole on every PR (~6.7 min and ~6.3 min in recent runs). In both, the actual perf/benchmark rounds dominate (~300s of ~400s); the rest is fixed setup, and the parallel build/baseline jobs finish in 1–2.5 min. The goal is to cut wall-clock time **without** weakening the perf comparison's reliability, which depends on running the baseline (PR base) and current code interleaved on the same runner and gating significance through `ariakit perf-compare` (±10% median delta with per-round sign agreement; Chrome also has a 5ms floor).

## Solution

**Chrome — shard across two runners.** The Chrome job is now a 2-way matrix. Each shard benchmarks a disjoint slice of the perf-chrome test files; the slice is computed once from the head checkout and passed to **both** the baseline and current runs, so every test's baseline/current measurements stay paired on one runner regardless of how the base and head test sets differ. Each shard writes shard-suffixed round files (`baseline-1-s1-worker0.json`) and uploads them raw; the existing `post` PR-comment job downloads every shard's rounds and runs `perf-compare` **once** over them. Partial results are intentional: if a shard fails it stays visibly red as its own matrix job, and `post` still reports whatever rounds the succeeding shards produced.

**Node — leaner, source-based benchmarks.** The tinybench budget drops from 2000/500ms to 1500/400ms (validated to stay well under the ±10% significance gate for unchanged code). The benchmark already resolves `@ariakit/store`/`@ariakit/utils` from source, so the `build-packages` job, the `package-build` download, and the baseline package build were removed as dead weight. The Node section of the comment is gated on the Node job succeeding, so its `if: always()` artifact can't surface a stale preliminary comparison.

**Shared setup trims.** Shallow base-sha `git fetch` instead of full history, `--prefer-offline` baseline installs, and a `cancel-in-progress` concurrency group.

## Reliability

The same-runner paired baseline-vs-current model and the `perf-compare` significance gate are unchanged; nothing about *what* is measured changes. A new `perf-compare` test locks the shard-suffixed round-file contract (round discovery + per-round pairing + disjoint-key merge).

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build app
@@ -56,31 +60,6 @@ jobs:
           path: nextjs/.open-next
           retention-days: 7
           include-hidden-files: true
-
-  build-packages:
-    name: Build packages
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
-        with:
-          persist-credentials: false
-
-      - name: Set up environment
-        uses: ./.github/workflows/setup
-
-      - name: Build packages
-        run: pnpm build
-
-      - name: Upload package artifact
-        uses: actions/upload-artifact@v7
-        with:
-          name: package-build
-          path: |
-            packages/*/dist
-            packages/*/solid
-            packages/*/package.json
-          retention-days: 7
 
   chrome-baseline:
     name: Prepare Chrome baseline
@@ -248,16 +227,10 @@ jobs:
             cp "$file" "$BASELINE_DIR/$file"
           done
 
-      - name: Build baseline packages
-        working-directory: ${{ runner.temp }}/perf-node-baseline
-        run: pnpm build
-
       - name: Pack baseline artifacts
         run: |
           cd "$BASELINE_DIR"
           tar -czf "$RUNNER_TEMP/perf-node-baseline-build.tar.gz" \
-            packages/*/dist \
-            packages/*/solid \
             packages/*/package.json \
             vitest.bench.config.ts \
             packages/*/benchmark
@@ -270,9 +243,18 @@ jobs:
           retention-days: 7
 
   chrome:
-    name: Chrome
+    name: Chrome (shard ${{ matrix.shard }})
     needs: [build, build-nextjs, chrome-baseline]
     runs-on: ubuntu-latest
+    strategy:
+      # Let both shards finish even if one fails, so each shard's logs and
+      # uploaded rounds stay available for debugging. Chrome perf is still
+      # all-or-nothing: chrome-merge only runs when every shard succeeded.
+      fail-fast: false
+      matrix:
+        # Raise the shard count (and PERF_SHARD_TOTAL below) as the number of
+        # perf-chrome test files grows; the file partition handles any count.
+        shard: [1, 2]
     permissions:
       actions: read
       contents: read
@@ -285,12 +267,16 @@ jobs:
       # enough signal for noisy results.
       PERF_ITERATIONS: 5
       PERF_WARMUP: 1
-      PERF_RESULTS_TITLE: Chrome
+      # Each shard benchmarks a disjoint subset of the perf test files and runs
+      # its own interleaved baseline-vs-current rounds on this one runner, so the
+      # paired same-runner comparison is preserved. chrome-merge runs
+      # perf-compare once over every shard's raw round files.
+      PERF_SHARD_INDEX: ${{ matrix.shard }}
+      PERF_SHARD_TOTAL: 2
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
         with:
-          fetch-depth: 0
           persist-credentials: false
 
       - name: Set up environment
@@ -302,11 +288,14 @@ jobs:
           engine: chromium
 
       - name: Set up baseline worktree
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
           BASELINE_DIR="$RUNNER_TEMP/perf-baseline"
           echo "BASELINE_DIR=$BASELINE_DIR" >> "$GITHUB_ENV"
-          git worktree add --detach "$BASELINE_DIR" \
-            ${{ github.event.pull_request.base.sha }}
+          git fetch --depth=1 origin "$BASE_SHA" || git fetch origin "$BASE_REF"
+          git worktree add --detach "$BASELINE_DIR" "$BASE_SHA"
 
       - name: Install dependencies (baseline)
         working-directory: ${{ runner.temp }}/perf-baseline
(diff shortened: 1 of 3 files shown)


.github/workflows/perf.yml
```diff
@@ -419,14 +439,67 @@ jobs:
...
+        with:
+          name: perf-chrome-rounds-${{ matrix.shard }}
+          path: |
+            app/.perf-results/baseline-*.json
+            app/.perf-results/current-*.json
+          if-no-files-found: ignore
+          include-hidden-files: true
+          retention-days: 7
+
+  chrome-merge:
```
diegohaz: I think we should do this in the post PR comment job. It's also fine to report partial results in the PR. We'll still see the failed shards in CI.

.github/workflows/perf.yml
```diff
@@ -270,9 +243,18 @@ jobs:
           retention-days: 7
 
   chrome:
-    name: Chrome
+    name: Chrome (shard ${{ matrix.shard }})
```
diegohaz:
```suggestion
    name: Chrome (${{ matrix.shard }})
```

Or, if it's easy to get the total shards: `Chrome (${{ matrix.shard }} / N)`

.github/workflows/perf.yml
```diff
@@ -419,14 +439,67 @@ jobs:
...
+
+      - name: Compare interleaved perf results
+        run: |
+          rm -rf app/.perf-results
+          mkdir -p app/.perf-results
+          shopt -s nullglob
+          round_files=("$RUNNER_TEMP"/perf-chrome-rounds/*.json)
+          if [ ${#round_files[@]} -eq 0 ]; then
+            echo "::warning::No Chrome shard round results found; skipping comparison."
+            exit 0
```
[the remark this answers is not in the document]
diegohaz: Fixed in 76f060728 - the Chrome comparison now runs in the `post` job over the shards' raw round artifacts and reports whatever rounds succeeded. Partial results are intentional here per the maintainer note below: a failed shard stays visibly red as its own matrix job in CI rather than being hidden. The `post` job also gates the Node section on the Node job succeeding, so a failed Node confirmation round can't publish a stale preliminary comparison.

.github/workflows/perf.yml
diegohaz: Fixed in 76f060728 - removed the dedicated `chrome-merge` job. The `post` job now downloads the shards' raw round artifacts and runs `perf-compare` once, reporting partial results when a shard fails (the failed shard stays red in CI). Node still publishes its own pre-computed comparison and is gated on the Node job succeeding, so its `if: always()` artifact can't surface a stale preliminary comparison.

.github/workflows/perf.yml
diegohaz: Fixed in 76f060728 - renamed to `Chrome (${{ matrix.shard }} / ${{ strategy.job-total }})`, which renders as `Chrome (1 / 2)`. `strategy.job-total` exposes the shard total, so the `/ N` form works without hardcoding.

diegohaz:
Addressed the review feedback in 76f060728 (pre-commit reviewed by three independent reviewers; clean pass after fixes).

**Fixed**

- 76f060728 ([reply](https://github.com/ariakit/ariakit/pull/6074#discussion_r3322022921)): Removed the dedicated `chrome-merge` job; the `post` job now downloads the shards' raw round artifacts and runs `perf-compare` once, reporting whatever rounds the succeeding shards produced (partial results). A failed shard stays visibly red in CI.
- 76f060728 ([reply](https://github.com/ariakit/ariakit/pull/6074#discussion_r3322022553)): To keep partial reporting safe, the `post` job gates the Node section on the Node job succeeding (Node uploads with `if: always()`, so a failed confirmation round could otherwise publish a stale preliminary comparison); Chrome stays file-based since `post` recomputes it from raw rounds. Also reordered the Node download ahead of the Chrome compare and made the Chrome compare `continue-on-error` so a Chrome problem can never suppress the Node section.
- 76f060728 ([reply](https://github.com/ariakit/ariakit/pull/6074#discussion_r3322022969)): Renamed the matrix job to `Chrome (${{ matrix.shard }} / ${{ strategy.job-total }})` (renders as `Chrome (1 / 2)`).

.github/workflows/perf.yml
```diff
@@ -285,12 +268,16 @@ jobs:
       # enough signal for noisy results.
       PERF_ITERATIONS: 5
       PERF_WARMUP: 1
-      PERF_RESULTS_TITLE: Chrome
+      # Each shard benchmarks a disjoint subset of the perf test files and runs
+      # its own interleaved baseline-vs-current rounds on this one runner, so the
+      # paired same-runner comparison is preserved. The post job runs
+      # perf-compare once over every shard's raw round files.
+      PERF_SHARD_INDEX: ${{ matrix.shard }}
+      PERF_SHARD_TOTAL: 2
```
diegohaz: Shouldn't it be `${{ strategy.job-total }}`?

diegohaz: Fixed in ebd0704d9 - good call, `${{ strategy.job-total }}` works in the job-level `env` context (confirmed against GitHub's context-availability table), so `PERF_SHARD_TOTAL` now derives from it instead of the hardcoded `2`. Since `shard` is the only matrix dimension the job count equals the shard count, so it stays `2` today and tracks the matrix automatically. (Also dropped the spaces in the job name per your note, so it renders as `Chrome (1/2)`.)

diegohaz:
Addressed the follow-up in ebd0704d9 (pre-commit reviewed by three reviewers; clean).

**Fixed**

- ebd0704d9 ([reply](https://github.com/ariakit/ariakit/pull/6074#discussion_r3322047566)): `PERF_SHARD_TOTAL` now derives from `${{ strategy.job-total }}` instead of a hardcoded `2` (valid in the job-level `env` context; equals the shard count since `shard` is the only matrix dimension). Also dropped the spaces in the job name so it renders as `Chrome (1/2)`.

.github/workflows/perf.yml
```diff
@@ -588,26 +598,69 @@ jobs:
...
+          PERF_RESULTS_DIR: ${{ runner.temp }}/perf-results
+        run: |
+          shopt -s nullglob
+          rounds=("$RUNNER_TEMP"/perf-chrome-rounds/*.json)
+          if [ ${#rounds[@]} -eq 0 ]; then
+            echo "::warning::No Chrome shard rounds found; skipping Chrome comparison."
+            exit 0
+          fi
+          # Partial results are fine: compare whatever rounds the succeeding
+          # shards produced. Any failed shard stays visibly red in CI.
```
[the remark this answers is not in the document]
diegohaz: Done - updated the PR description. The Solution section now describes the `post` PR-comment job downloading every shard's raw rounds and running `perf-compare` once over them, with partial results intentional (a failed shard stays red as its own matrix job; `post` reports whatever the succeeding shards produced). Removed the stale "chrome-merge only runs when every shard succeeds / won't publish a partial suite" wording.

diegohaz:
**Fixed**

- PR description ([reply](https://github.com/ariakit/ariakit/pull/6074#discussion_r3322057117)): Updated the description to match the current design — the `post` job compares whatever shard rounds were produced (partial results are intentional; a failed shard stays red in CI), replacing the stale "`chrome-merge` only runs when every shard succeeds / won't publish a partial suite" wording. No code change.

The other unresolved thread already has my reply from `ebd0704d9`; the earlier three threads were resolved.

PR #6076, opened 2026-05-29
Fix Hovercard userEvent hover regression

### LINKED ISSUE #6075: Hover from the React testing library does not work with Hovercard anymore
### Current behavior

Hi!

Since the latest version it is not possible to test the `Hovercard` component anymore with the `hover` function of `@testing-libraray/user-event`.

So running a test like this

```tsx
render(<App />)

userEvent.hover(screen.getByRole("link", { name: "@ariakit.com" }))

expect(await screen.findByRole("dialog")).toBeVisible()
```

currently fails.

1. Open https://stackblitz.com/edit/github-yhqg9hzr?file=src%2Fapp.tsx
2. Execute `npm run test`
3. Dialog will not be found

### Expected behavior

Dialog can be found

### Workaround

Use `user.pointer` with non-zero coordinates instead of `user.hover`:

```tsx
const user = userEvent.setup()
const anchor = screen.getByRole("link", { name: "@ariakit.com" })

await user.pointer({
  target: anchor,
  coords: {
    screenX: 1,
    screenY: 1,
    clientX: 1,
    clientY: 1,
  },
})

expect(await screen.findByRole("dialog")).toBeVisible()
```

More details: https://github.com/ariakit/ariakit/issues/6075#issuecomment-4571300417

### Possible solutions

_No response_

diegohaz:
## Motivation

Fixes #6075.

Hovercard stopped opening in tests that use Testing Library `userEvent.hover`. The regression only showed up against built package output, where Rolldown's browser-platform defaults folded runtime `process.env.NODE_ENV` checks such as the `process.env.NODE_ENV === "test"` fallback in `@ariakit/react-utils`.

## Solution

Build published packages with Rolldown's neutral platform so runtime `process.env.NODE_ENV` checks stay in the distributed output instead of being inlined at build time.

The legacy tsup build does not need an explicit `replaceNodeEnv: false` guard because its default already preserves these checks.

## Workaround

Until the library fix is released, tests can use `user.pointer` with non-zero coordinates instead of `user.hover` so Ariakit can detect pointer movement in the built package output:

```ts
// TODO: Remove after https://github.com/ariakit/ariakit/issues/6075 is fixed.
await user.pointer({
  target: element,
  coords: {
    screenX: 1,
    screenY: 1,
    clientX: 1,
    clientY: 1,
  },
});
```


.changeset/300-hovercard-testing-hover.md
```diff
@@ -0,0 +1,18 @@
...
+"@ariakit/solid-store": patch
+"@ariakit/solid-utils": patch
+"@ariakit/store": patch
+"@ariakit/tailwind": patch
+"@ariakit/test": patch
+"@ariakit/utils": patch
+---
+
+Fixed runtime `process.env.NODE_ENV` checks in published package output,
+including test-only behavior and development warnings.
```
diegohaz: Unexpected line break

app/package.json
```diff
@@ -111,6 +111,8 @@
     "@ariakit/test": "workspace:*",
     "@axe-core/playwright": "4.11.3",
     "@playwright/test": "1.60.0",
+    "@testing-library/react": "16.3.2",
+    "@testing-library/user-event": "14.6.1",
```
diegohaz: Remove the sandbox and package.json changes. Now that we've confirmed this, we can keep only the build script changes and the changeset.

.changeset/300-hovercard-testing-hover.md
diegohaz: Fixed in 2fdace545 - the changeset entry is now a single-line sentence.

app/package.json
diegohaz: Fixed in 2fdace545 - removed the sandbox, app dependency additions, and related lockfile entries so the PR now keeps only the build script change and changeset.

diegohaz:
**Fixed**

- 2fdace545 ([changeset thread](https://github.com/ariakit/ariakit/pull/6076#discussion_r3325824530)): made the changeset entry a single-line sentence.
- 2fdace545 ([sandbox/dependency thread](https://github.com/ariakit/ariakit/pull/6076#discussion_r3325832519)): removed the sandbox, app dependency additions, and related lockfile entries so the PR keeps only the build script change and changeset.

packages/ariakit-scripts/src/build.ts
```diff
@@ -34,6 +34,9 @@ type ExportMode = "source" | "build";
 const sourceDir = "src";
 const distDir = "dist";
 const solidDir = "solid";
+// Use the neutral platform for published libraries so Rolldown doesn't inline
+// runtime process.env.NODE_ENV checks for browser builds.
+const buildPlatform = "neutral";
```
diegohaz, line 39: The `platform: "neutral"` fix is correct — I verified it locally: with the default/`"browser"` platform, `@ariakit/react-utils`'s `return movementX || movementY || process.env.NODE_ENV === "test"` is folded down to `... || false` in the built output (stripping the test fallback that #6075 depends on), while `"neutral"` (and `"node"`) preserve it. The only line that differs between the `browser` and `neutral` output for this package is that check, so the change is surgical.

The one gap: with the regression sandbox removed in 2fdace545, this behavior now ships with **no automated guard**, so a future edit that drops `platform` (reverting to the browser default) would silently reintroduce #6075. If the original sandbox felt too heavy, a much cheaper guard would catch the same regression without the temp-build + jsdom + Testing Library machinery — e.g. build one package via `build()` and assert the emitted `dist/index.js` still contains the runtime `process.env.NODE_ENV` expression rather than a folded constant:

```ts
// after running build() for a fixture/package in a temp dir
const dist = await readFile(join(packageDir, "dist/index.js"), "utf8");
expect(dist).toMatch(/process\.env\.NODE_ENV/);
```

Your call whether it's worth adding back in some form, but a build-config behavior like this is easy to regress unnoticed.

.changeset/300-hovercard-testing-hover.md
```diff
@@ -0,0 +1,17 @@
...
+"@ariakit/react-store": patch
+"@ariakit/react-utils": patch
+"@ariakit/solid": patch
+"@ariakit/solid-components": patch
+"@ariakit/solid-store": patch
+"@ariakit/solid-utils": patch
+"@ariakit/store": patch
+"@ariakit/tailwind": patch
+"@ariakit/test": patch
+"@ariakit/utils": patch
```
diegohaz, line 14: Two of these bumps look unrelated to this change:

- **`@ariakit/tailwind`** is built by its own `node src/build.ts`, not the `@ariakit/scripts` rolldown build this PR modifies. It has `dependencies: {}` and no `process.env.NODE_ENV` usage, so it's neither rebuilt by this change nor a dependent of anything that is — this entry ships an unrelated, identical-output patch.
- **`@ariakit/utils`** *is* rebuilt by the changed tooling, but its output is byte-identical under the old (`browser`) and new (`neutral`) platform (I diffed both); its only `NODE_ENV` reference is a JSDoc example in `misc.ts`, not runtime code — effectively a no-op bump.

The packages whose shipped code actually changes are `@ariakit/components`, `@ariakit/react-components`, `@ariakit/react-utils`, and `@ariakit/store` — each has a real `process.env.NODE_ENV` check that browser-platform builds were folding to a constant. With `fixed`/`linked` empty and `updateInternalDependencies: "patch"`, dependents that consume those packages are patched automatically, so the rest only need explicit entries for changelog visibility. Even a "republish everything rebuilt with the new config" rationale wouldn't cover `@ariakit/tailwind`, which uses a separate build.

diegohaz (commented):
## Review summary

**Initial review** (no prior review from this skill). The branch was rebased and the regression sandbox removed (commit `2fdace545`) while I was reviewing, so this targets the current head `2fdace545`; my earlier inline comments against `42426a2` were deleted and superseded. Comment-only — not approving or requesting changes.

Baseline: merge-base `ebcba9e5d` → head `2fdace545`. Current scope is 2 files: `packages/ariakit-scripts/src/build.ts` and `.changeset/300-hovercard-testing-hover.md`.

### The fix is correct (validated locally, Node 24)
- Built `@ariakit/react-utils` under each platform: `(default)` and `browser` fold `return movementX || movementY || process.env.NODE_ENV === "test"` to `... || false` (stripping the test fallback #6075 needs); `neutral` and `node` preserve it. The only differing output line between `browser` and `neutral` for this package is that check, so the change is surgical.
- Earlier in the run I ran the (now-removed) sandbox test: it passed, and with the `platform` lines reverted it failed (`expected false to be true`) — confirming the build change genuinely fixes the regression.
- Confirmed the legacy tsup path (`@ariakit/react`) was never affected: tsup's `replaceNodeEnv` defaults to `false`, and its CJS/ESM output preserves `process.env.NODE_ENV` by default. The updated PR description now states this correctly.

### Findings (both non-blocking)
1. **No regression guard for the fix** — removing the sandbox leaves this build-config behavior untested, so dropping `platform` later would silently reintroduce #6075. Suggested a lightweight output-string assertion as a cheaper alternative to the removed test. (on `build.ts`)
2. **Changeset over-includes packages** — `@ariakit/tailwind` (separate build, no dependency on the change) and `@ariakit/utils` (byte-identical output) are effectively no-op bumps; the packages whose output actually changes are `components`, `react-components`, `react-utils`, and `store`. (on the changeset)

### Notes
- The PR description is now accurate; earlier it described the abandoned `transform.define` + `replaceNodeEnv` approach.
- `pnpm tsc` ✅ and `biome check` ✅ on the changed files (build/changeset content is unchanged across the rebase).
- Not separately validated: the Solid build output (no `NODE_ENV` usage there) and a full multi-package `pnpm build`.

Issue #185, opened 2018-08-13
Add typing tests


diegohaz:
Reference:

- package.json's script:
  https://github.com/styled-components/styled-components/blob/f401eda35ef16065509ea8a7375290fb8b5858a6/package.json#L34

- tests:
  https://github.com/styled-components/styled-components/tree/f401eda35ef16065509ea8a7375290fb8b5858a6/typings/tests

PR #6084, opened 2026-05-29
Fix Maximum update depth error when rendering many menus

### LINKED ISSUE #3214: Multiple Menus rendered to page causes Maximum update depth exceeded.
### Current behavior

In my app, I have a table and each row renders a kabab menu.  I noticed recently that when there are more than ~20 rows, I get a warning in my browser console during dev:

> Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.

The exact number of menus needed seems to vary.  In my app, it was 24, but in the codesandbox, it's something like 60.  

1. Open sandbox: https://codesandbox.io/p/sandbox/affectionate-merkle-28qvm9
2. See console logs

### Expected behavior

I think I should be able to render as many menus as needed.

### Workaround

The menus still seem to work, but the warning is concerning.

https://github.com/ariakit/ariakit/issues/3214#issuecomment-1858577119

### Possible solutions

_No response_

diegohaz:
## Motivation

Rendering many menus on a single page made React throw `Maximum update depth exceeded` (the minified `#185` error in production builds) during the initial render and hydration. The number of menus needed varied — reporters hit it somewhere between ~20 and ~60 menus.

Fixes #3214.

## Solution

The error is React's nested-update limit, not an infinite loop. React's `useSyncExternalStore` forces a synchronous re-render for every store value that genuinely changes between render and commit. When many closed menus mount, each one registers its items synchronously during the commit phase, and these consecutive sync re-renders accumulate past React's limit (50). Plain [`Disclosure`](https://ariakit.org/reference/disclosure) and [`Popover`](https://ariakit.org/reference/popover) don't reproduce it because they don't register collection items — the menu's item registration is the trigger (which is also why `unmountOnHide`, by not rendering closed content at all, avoids it).

Unlike a [`Select`](https://ariakit.org/reference/select) — which interacts with its items while closed (typeahead, keyboard value changes, value→label display) — a [`Menu`](https://ariakit.org/reference/menu) never touches its items while hidden. So [`MenuItem`](https://ariakit.org/reference/menu-item) now defers registering itself until the menu is mounted. A closed menu then behaves like a `Popover`, and the cascade no longer happens regardless of how many menus are rendered (verified at 80, 200, and 400 menus). Items register as usual once the menu opens, so focus and keyboard navigation are unaffected.

## Workaround

Until this fix is released, pass [`unmountOnHide`](https://ariakit.org/reference/menu#unmountonhide) to the `Menu`. Closed menus then don't render their content, which avoids the mount-time cascade. The menus keep working as before:

```tsx
<Ariakit.MenuProvider>
  {/* TODO: Remove unmountOnHide once the fix for
      https://github.com/ariakit/ariakit/issues/3214 is released. */}
  <Ariakit.Menu unmountOnHide>
    <Ariakit.MenuItem>Edit</Ariakit.MenuItem>
    <Ariakit.MenuItem>Duplicate</Ariakit.MenuItem>
    <Ariakit.MenuItem>Delete</Ariakit.MenuItem>
  </Ariakit.Menu>
</Ariakit.MenuProvider>
```

### DIFF
--- a/app/src/sandbox/menu-3214/test-browser.ts
+++ b/app/src/sandbox/menu-3214/test-browser.ts
@@ -6,7 +6,7 @@ import { gotoAndSettle, withFramework } from "#app/test-utils/preview.ts";
 const updateDepthError =
   /maximum update depth exceeded|react\.dev\/errors\/185/i;
 
-withFramework(import.meta.dirname, async ({ test }) => {
+withFramework(import.meta.dirname, async ({ test, query }) => {
   test("renders many menus without exceeding React's update depth", async ({
     page,
     q,
@@ -26,9 +26,14 @@ withFramework(import.meta.dirname, async ({ test }) => {
     // the listeners already attached to capture it from the very first render.
     await gotoAndSettle(page, page.url());
 
-    // Confirm the island actually hydrated and the menus still work.
+    // The alwaysVisible menu is shown even though its store is closed, so its
+    // items must still register. This also confirms the island hydrated.
+    const registered = q.list("Registered items");
+    await test.expect(query(registered).listitem()).toHaveCount(3);
+
+    // A closed menu still registers and renders its items once it's opened.
     await q.button("Row 0 actions").click();
-    await test.expect(q.menu()).toBeVisible();
+    await test.expect(q.menuitem("Edit")).toBeVisible();
 
     test.expect(depthErrors).toEqual([]);
   });
--- a/app/src/sandbox/menu-3214/test-browser.ts
+++ b/app/src/sandbox/menu-3214/test-browser.ts
@@ -1,5 +1,4 @@
-import { errors } from "@playwright/test";
-import { withFramework } from "#app/test-utils/preview.ts";
+import { gotoAndSettle, withFramework } from "#app/test-utils/preview.ts";
 
 // React's "Maximum update depth exceeded" error. In production builds React
 // throws the minified variant that links to https://react.dev/errors/185.
@@ -23,17 +22,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
     });
     page.on("pageerror", (error) => collect(error.message));
 
-    // The error is thrown while the React island hydrates, so reload with the
-    // listeners already attached to capture it from the very first render.
-    await page.reload({ waitUntil: "load" });
-    // Bounded settle that only swallows the timeout, mirroring the shared
-    // `gotoAndSettle` helper so a stalled request can't consume the test budget
-    // and real failures (such as the page closing) still surface.
-    await page
-      .waitForLoadState("networkidle", { timeout: 5_000 })
-      .catch((error) => {
-        if (!(error instanceof errors.TimeoutError)) throw error;
-      });
+    // The error is thrown while the React island hydrates, so re-navigate with
+    // the listeners already attached to capture it from the very first render.
+    await gotoAndSettle(page, page.url());
 
     // Confirm the island actually hydrated and the menus still work.
     await q.button("Row 0 actions").click();
--- a/app/src/sandbox/menu-3214/test-browser.ts
+++ b/app/src/sandbox/menu-3214/test-browser.ts
@@ -1,0 +1,44 @@
+import { errors } from "@playwright/test";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// React's "Maximum update depth exceeded" error. In production builds React
+// throws the minified variant that links to https://react.dev/errors/185.
+// See https://github.com/ariakit/ariakit/issues/3214.
+const updateDepthError =
+  /maximum update depth exceeded|react\.dev\/errors\/185/i;
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("renders many menus without exceeding React's update depth", async ({
+    page,
+    q,
+  }) => {
+    const depthErrors: string[] = [];
+    const collect = (text: string) => {
+      if (updateDepthError.test(text)) {
+        depthErrors.push(text);
+      }
+    };
+    page.on("console", (message) => {
+      if (message.type() === "error") collect(message.text());
+    });
+    page.on("pageerror", (error) => collect(error.message));
+
+    // The error is thrown while the React island hydrates, so reload with the
+    // listeners already attached to capture it from the very first render.
+    await page.reload({ waitUntil: "load" });
+    // Bounded settle that only swallows the timeout, mirroring the shared
+    // `gotoAndSettle` helper so a stalled request can't consume the test budget
+    // and real failures (such as the page closing) still surface.
+    await page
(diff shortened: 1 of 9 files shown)


app/src/sandbox/menu-3214/test-browser.ts
```diff
@@ -0,0 +1,44 @@
...
+    // listeners already attached to capture it from the very first render.
+    await page.reload({ waitUntil: "load" });
+    // Bounded settle that only swallows the timeout, mirroring the shared
+    // `gotoAndSettle` helper so a stalled request can't consume the test budget
+    // and real failures (such as the page closing) still surface.
+    await page
+      .waitForLoadState("networkidle", { timeout: 5_000 })
+      .catch((error) => {
+        if (!(error instanceof errors.TimeoutError)) throw error;
+      });
```
diegohaz: Should we import gotoAndSettle from `#app/test-utils/preview.ts` instead of redefining it here?

packages/ariakit-react-components/src/menu/menu-item.tsx
```diff
@@ -120,6 +126,13 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
       store,
       preventScrollOnKeyDown,
       ...props,
+      // Skip registering items while the menu is closed. Unlike a Select,
+      // a Menu doesn't interact with its items while hidden, so registering
+      // them isn't useful. Registering many closed menus' items at once on
+      // mount can also cascade into React's "Maximum update depth exceeded"
+      // error. Items register once the menu is shown.
+      // See https://github.com/ariakit/ariakit/issues/3214.
+      shouldRegisterItem: mounted ? props.shouldRegisterItem : false,
```
diegohaz: This gate now treats `mounted` as the same thing as "the menu list is hidden", but `MenuList`/`Menu` also support `alwaysVisible` and `hidden={false}`. In those cases `isHidden(mounted, props.hidden, alwaysVisible)` returns false even while the store's `mounted` state is false, so the menu is visible and focusable but its items never register. I verified this locally with a temporary `Menu alwaysVisible` example: focusing the visible closed menu and pressing ArrowDown stays on the menu container on this PR, while replacing this line with the previous `props.shouldRegisterItem` behavior makes the same test pass.

The registration predicate needs to account for the list's actual hidden state, not just `store.mounted` — for example by deriving the same visibility decision in `MenuList` and passing whether item registration should be deferred through context, so `alwaysVisible`/`hidden={false}` menus keep their visible items keyboard-reachable while closed.

diegohaz (commented):
Initial full review of `fd4139deed7888acdd320f227e205865ff571394` against merge base `22e11feb891b51043058f00caba89dc8ebd58cb5`. I reviewed the menu item registration change, the new sandbox/browser test, the changeset, surrounding menu/composite/collection code, and existing review context. I left one inline finding for a regression in visible closed menus using `alwaysVisible`/`hidden={false}`.

Validation run:
- `pnpm exec vitest run examples/menu/test.ts examples/menu-nested/test.ts examples/menubar/test.ts examples/menu-item-checkbox/test.ts examples/menu-item-radio/test.ts` (5 files, 42 tests passed)
- `APP_PORT=4321 NEXTJS_PORT=3001 pnpm -F app test src/sandbox/menu-3214/test-browser.ts --project=chrome` (passed)
- Temporary local `Menu alwaysVisible` Vitest repro: failed on this PR, then passed when the new `mounted` registration gate was locally removed; the temporary files and experiment patch were reverted before publishing.

I did not run the full browser matrix, full `pnpm tsc`, or package build. I also did not duplicate the existing unresolved thread about reusing the shared `gotoAndSettle` helper.

packages/ariakit-react-components/src/menu/menu-item.tsx
diegohaz: Still applies on `4e86d6a35f96160d7a265d73b7b213c3ed5a048f`: the new commit only updates the changeset wording, and this `mounted ? props.shouldRegisterItem : false` gate is unchanged.

diegohaz (commented):
Re-review of the new head `4e86d6a35f96160d7a265d73b7b213c3ed5a048f` using the prior marked review SHA `fd4139deed7888acdd320f227e205865ff571394` as the incremental baseline. The only new commit changes `.changeset/300-menu-item-update-depth.md`, and that wording change looks fine.

The prior inline finding about `alwaysVisible`/`hidden={false}` menus still applies because the menu item registration gate is unchanged, so I added a short follow-up on that existing thread. I inspected `git diff fd4139deed7888acdd320f227e205865ff571394..HEAD`; I did not rerun the full test set after this wording-only commit. The validation from the previous review still covers the unchanged runtime code.

app/src/sandbox/menu-3214/test-browser.ts
diegohaz: Fixed in 9f6ae15ca — exported `gotoAndSettle` from `#app/test-utils/preview.ts` and the test now calls `gotoAndSettle(page, page.url())` to re-navigate (reloading the current URL) instead of redefining the bounded settle. Verified it still catches the error without the fix and passes with it.

diegohaz:
**Fixed**

- `9f6ae15ca` ([reply](https://github.com/ariakit/ariakit/pull/6084#discussion_r3327464065)): Exported `gotoAndSettle` from `#app/test-utils/preview.ts` and the `menu-3214` test now calls `gotoAndSettle(page, page.url())` to re-navigate, instead of redefining the bounded settle inline. Verified the test still catches the error without the library fix and passes with it.

packages/ariakit-react-components/src/menu/menu-item.tsx
diegohaz: Fixed in 6c8a553b3 — the registration predicate now uses the menu list's actual hidden state (`isHidden(mounted, hidden, alwaysVisible)`, exposed via a new `MenuListHiddenContext`) instead of the store's `mounted` state, so `alwaysVisible`/`hidden={false}` menus register their items while visible. The repro sandbox now also renders an `alwaysVisible` menu and asserts its items register — verified that assertion fails when the gate is reverted to `mounted`.

diegohaz:
**Fixed**

- `6c8a553b3` ([reply](https://github.com/ariakit/ariakit/pull/6084#discussion_r3327629273)): The `alwaysVisible` / `hidden={false}` regression. `MenuItem` now gates item registration on the menu list's *actual hidden state* (`isHidden(mounted, hidden, alwaysVisible)`, exposed via a new `MenuListHiddenContext`) rather than the store's `mounted` state, so a visible-but-closed menu still registers its items. Closed (`display: none`) menus still skip registration, so the original update-depth fix is preserved. The repro sandbox now also renders an `alwaysVisible` menu and asserts its items register (verified this fails if the gate reverts to `mounted`). Validated: full vitest `877/877`, repro count-independent at 80/200 menus, production build.

PR #6086, opened 2026-05-29
Fix combobox autoSelect dropping characters when the popover resizes

### LINKED ISSUE #3837: Combobox Typing Issues with Virtualized List on Some Android Devices
### Current behavior

When typing in the combobox, the popover continuosly expands/reduces its height, causing some typed characters to be lost. It seems the autocomplete bar above the keyboard is triggering continuous height recalculations, leading Ariakit to reconsider the available space for the popover. This issue is experienced in Pixel 7 and Pixel 8 device. (i cannot check other android devices).

Open https://ariakit.org/examples/select-combobox-virtualized in a android Device (pixel 7 or pixel8),
Open the select combobox and try to write a text in combobox

### Expected behavior

Expect user can write smoothly into the select combobox

### Workaround

do not use autoSelect 

### Possible solutions

_No response_

diegohaz:
## Motivation

Fixes #3837

On some Android devices, typing in a virtualized [`Combobox`](https://ariakit.com/reference/combobox) with [`autoSelect`](https://ariakit.com/reference/combobox#autoselect) drops characters. The reported reproduction is the [Select with Combobox (virtualized)](https://ariakit.com/examples/select-combobox-virtualized) example on a Pixel 7/8.

The root cause is a feedback loop between the popover size and the auto-select behavior:

1. While the user types, the keyboard's autocomplete bar repeatedly changes the visual viewport height.
2. Floating UI recomputes the popover's available space and updates the `--popover-available-height` variable, which drives the popover `max-height`.
3. In a virtualized list, the smaller/larger popover renders a different set of items, so the store's rendered items change.
4. The `Combobox` auto-select effect re-runs whenever the rendered items change (it has to, in order to auto-select asynchronously loaded items) and calls `store.move()` on the **already-active** item.
5. `store.move()` always increments the internal `moves` counter, so `Composite` re-focuses the active item via `focusIntoView`, briefly bouncing DOM focus off the input and back.

That focus churn, repeated on every viewport oscillation, is what drops characters. Without `autoSelect` the effect only scrolls the active item into view (no focus change), which is why disabling `autoSelect` avoided the bug.

## Solution

In the `Combobox` auto-select effect, skip the `store.move()` call when the auto-select target is already the active item. Re-moving to the same item only re-triggers the focus side effect. When the rendered items change for an unrelated reason (such as a resize), the active item is preserved without bouncing focus, while a genuine change of the auto-select target (typing, async items) still moves as before.

`store.move()` also refreshed the store's `activeValue` (used by inline autocomplete) as a side effect of bumping `moves`. To keep that behavior, the skip path updates `activeValue` directly — without bumping `moves`, so no focus change — which is a no-op when the value is unchanged.

## Workaround

Until the fix is released, you can keep `autoSelect` and stop the virtualizer from re-rendering on resize, which breaks the focus-churn loop. The popover still resizes with the available space; only the virtualized window stops recomputing on resize (it still recomputes on scroll):

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/3837 is released.
<SelectRenderer store={select} items={matches} renderOnResize={false}>
  {/* ... */}
</SelectRenderer>
```

Alternatively, you can decouple the popover `max-height` from `--popover-available-height` (use a fixed height) so the keyboard no longer resizes the popover, or disable `autoSelect` entirely (the originally reported workaround, at the cost of losing the feature).

## Reproduction

`app/src/sandbox/select-combobox-3837/` reproduces the bug. Its `test-browser.ts` shrinks the popover through `--popover-available-height` (the same variable Floating UI updates when the viewport changes) and asserts that focus never bounces to an option while auto-selecting. The test fails on the current code and passes with the fix.


diegohaz (commented):
Initial review of the full PR diff from merge base `22e11feb891b51043058f00caba89dc8ebd58cb5` to head `66ea0b6b4939ff68593bb92681d0616c48c9e15d`.

I reviewed the combobox auto-select guard, the surrounding Combobox/Composite/Collection store focus paths, SelectRenderer virtualization behavior, the new sandbox reproduction/test, the changeset, and the linked #3837 context. I found no substantive issues to report.

Validation run locally:

- `pnpm lint` passed.
- `pnpm tsc` passed.
- `APP_PORT=4323 NEXTJS_PORT=3002 pnpm -F app exec playwright test src/sandbox/select-combobox-3837/test-browser.ts src/sandbox/select-combobox-5047/test-browser.ts src/sandbox/dialog-combobox-4825/test-browser.ts --project=chrome` passed overall; `dialog-combobox-4825` had an initial-load flake and passed on retry.
- `APP_PORT=4323 NEXTJS_PORT=3002 pnpm -F app exec playwright test src/sandbox/select-combobox-3837/test-browser.ts --project=chrome --project=android` passed the Chrome run. The Android project did not select this file because it is named `test-browser.ts`, so I did not get separate Android-project coverage from that command.

I did not run the full browser matrix.

.changeset/300-combobox-autoselect-resize.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Combobox`](https://ariakit.com/reference/combobox) with [`autoSelect`](https://ariakit.com/reference/combobox#autoselect) losing typed characters when the popover was resized while typing, which could happen with a virtualized list on mobile devices where the keyboard's autocomplete bar repeatedly changes the available viewport height.
```
diegohaz: Turn this into a multiline changeset with a heading

diegohaz: Fixed in 6a687e932 - turned this into a multiline changeset with a short title heading followed by detail paragraphs.

diegohaz:
Addressed the open review comment.

**Fixed**

- `6a687e932` ([reply](https://github.com/ariakit/ariakit/pull/6086#discussion_r3327545061)): turned the changeset into a multiline entry with a short title heading followed by detail paragraphs describing the symptom and the mobile virtualized-list scenario.

PR #6087, opened 2026-05-29
Fix Dialog throwing on document events with a non-Node target

### LINKED ISSUE #5156: Target not of type Node during contains call
### Current behavior

This bug I think has just been causing noise in our error tracking but no actual issues. Just thought I would share.

https://github.com/ariakit/ariakit/blob/4e68302ad2b5efe2894044c01d76f7810e32fe3b/packages/ariakit-react-core/src/dialog/utils/use-hide-on-interact-outside.ts#L83

On that line of code, there is a in Document check, which calls a core `contains` function. But the target sent in is not always of type node. Resulting in `parent.contains(child);` throwing `TypeError: Failed to execute 'contains' on 'Node': parameter 1 is not of type 'Node'.`

I am guessing the `useEventOutside` effect just needs another possible early return, but hard to say with much confidence as an outsider of the project.

I have not repro'd this, I just see it in our error tracking.

Image

Sorry, this is not repro'd, it is just caught by error tracking. If I get a recorded session I can share more findings.

### Expected behavior

No error is thrown.

### Workaround

_No response_

### Possible solutions

_No response_

diegohaz:
## Motivation

Closes #5156.

While a dialog is open, Ariakit attaches global `click`, `focusin`, `contextmenu`, and `keydown` listeners to the document to detect interactions outside the dialog and to close it on <kbd>Escape</kbd>. Those handlers cast `event.target` to `Element` and immediately pass it to `contains()` (through `isInDocument`) and other `Element`-only methods.

`event.target` is an `EventTarget`, though — not necessarily a `Node`. When third-party code dispatches an event whose target is a non-Node `EventTarget` (for example `window`, an `XMLHttpRequest`, or an `AudioContext`), `document.body.contains(target)` throws:

> TypeError: Failed to execute 'contains' on 'Node': parameter 1 is not of type 'Node'.

The error doesn't break the dialog, but it pollutes consumers' error tracking, as reported in #5156.

## Solution

The root cause is that a handler treats `event.target`/`event.relatedTarget` (an `EventTarget`) as an `Element` and calls `contains()` or other `Element`-only methods on it.

This adds a small, cross-realm-safe `isElement` helper to `@ariakit/utils` and uses it to guard the affected global document listeners before any `Element`-only call:

- `Dialog`'s interact-outside listeners (`click`, `focusin`, `contextmenu`) and its <kbd>Escape</kbd>-to-close `keydown` listener.
- `HovercardDisclosure`'s `focusout` listener, which had the same crash when `relatedTarget` was a non-element.

It also replaces a couple of equivalent inline checks (`CompositeHover`'s `relatedTarget` check and the global focus-visible `mousedown` listener) with the same helper.

`isElement` tests `nodeType` rather than `instanceof Element` so that elements from same-origin child frames — which `addGlobalEventListener` also listens on — aren't wrongly rejected for belonging to a different realm.

## Workaround

Until the fix is released, you can stop the foreign events before they reach Ariakit's listeners. Register a capture-phase guard once, before any dialog opens (for example near the root of your app), that ignores events whose target is not a DOM node. Because it is registered first, its capture-phase listener runs before Ariakit's and prevents the non-Node-target events from reaching them.

```tsx
import { useEffect } from "react";

// TODO: Remove once https://github.com/ariakit/ariakit/issues/5156 is fixed.
function useIgnoreNonNodeEvents() {
  useEffect(() => {
    const types = ["click", "focusin", "contextmenu", "keydown"];
    const onEvent = (event: Event) => {
      if (event.target && !(event.target instanceof Node)) {
        event.stopImmediatePropagation();
      }
    };
    for (const type of types) {
      document.addEventListener(type, onEvent, true);
    }
    return () => {
      for (const type of types) {
        document.removeEventListener(type, onEvent, true);
      }
    };
  }, []);
}
```

This stopgap only watches the main document and uses `instanceof Node`, which is not cross-realm-safe, so it does not cover dialogs inside child frames. The library fix in this PR has neither limitation.

## Reproduction

Two sandboxes reproduce the crash by dispatching events with a non-element target while the relevant component is mounted, then asserting that no page error is reported (verified to fail without the fix and pass with it on Chromium, Firefox, and WebKit):

- `app/src/sandbox/dialog-5156/` opens a dialog and dispatches `click`, `focusin`, `contextmenu`, and `keydown` (<kbd>Escape</kbd>) events on the document with a non-Node `target`.
- `app/src/sandbox/hovercard-5156/` shows a `HovercardDisclosure` and dispatches a `focusout` with a non-element `relatedTarget`.

### DIFF
--- a/.changeset/300-dialog-non-node-interact-outside.md
+++ b/.changeset/300-dialog-non-node-interact-outside.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) throwing a `TypeError` when an event with a non-element target (such as one dispatched on `window` or an `XMLHttpRequest`) reached its global document listeners while the dialog was open.
--- a/.changeset/300-dialog-non-node-interact-outside.md
+++ b/.changeset/300-dialog-non-node-interact-outside.md
@@ -1,6 +1,0 @@
----
-"@ariakit/react-components": patch
-"@ariakit/react": patch
----
-
-Fixed [`Dialog`](https://ariakit.com/reference/dialog) throwing a `TypeError` when an event with a non-element target (such as one dispatched on `window` or an `XMLHttpRequest`) reached its global document listeners while the dialog was open.
--- a/app/src/sandbox/hovercard-5156/test-browser.ts
+++ b/app/src/sandbox/hovercard-5156/test-browser.ts
@@ -1,0 +1,34 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("does not throw when a focusout has a non-Node relatedTarget", async ({
+    page,
+    q,
+  }) => {
+    const errors: string[] = [];
+    page.on("pageerror", (error) => errors.push(error.message));
+
+    // Focusing the disclosure shows it, which attaches its global focusout
+    // listener to the document.
+    await q.button("More details").focus();
+    await test.expect(q.button("More details")).toBeVisible();
+
+    // A focusout whose relatedTarget is a non-Node EventTarget (such as one
+    // dispatched on window or an XMLHttpRequest) used to make the listener call
+    // `contains()` on it and throw a TypeError.
+    // See https://github.com/ariakit/ariakit/issues/5156
+    await page.evaluate(() => {
+      document.dispatchEvent(
+        new FocusEvent("focusout", {
+          bubbles: true,
+          relatedTarget: new EventTarget(),
+        }),
+      );
+    });
+
+    // Round-trip so any page error from the dispatch above is delivered before
+    // we assert on it.
+    await page.evaluate(() => {});
+    test.expect(errors).toEqual([]);
+  });
+});
--- a/app/src/sandbox/hovercard-5156/test-browser.ts
+++ b/app/src/sandbox/hovercard-5156/test-browser.ts
@@ -8,10 +8,14 @@ withFramework(import.meta.dirname, async ({ test }) => {
     const errors: string[] = [];
     page.on("pageerror", (error) => errors.push(error.message));
 
-    // Focusing the disclosure shows it, which attaches its global focusout
-    // listener to the document.
+    // Focusing the disclosure shows it and attaches its global focusout listener
+    // to the document; pressing Enter then opens the hovercard. Waiting for the
+    // hovercard content confirms that flow completed before we dispatch below.
     await q.button("More details").focus();
-    await test.expect(q.button("More details")).toBeVisible();
+    await page.keyboard.press("Enter");
+    await test
+      .expect(q.text("Toolkit for building accessible web apps."))
+      .toBeVisible();
 
     // A focusout whose relatedTarget is a non-Node EventTarget (such as one
     // dispatched on window or an XMLHttpRequest) used to make the listener call
--- a/packages/ariakit-react-components/src/composite/composite-hover.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-hover.tsx
@@ -28,14 +28,9 @@ const TagName = "div" satisfies ElementType;
 type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
 
-function getMouseDestination(event: ReactMouseEvent<HTMLElement>) {
-  if (isElement(event.relatedTarget)) return event.relatedTarget;
-  return null;
-}
-
 function hoveringInside(event: ReactMouseEvent<HTMLElement>) {
-  const nextElement = getMouseDestination(event);
-  if (!nextElement) return false;
+  const nextElement = event.relatedTarget;
+  if (!isElement(nextElement)) return false;
   return contains(event.currentTarget, nextElement);
 }
 
@@ -43,8 +38,8 @@ const symbol = Symbol("composite-hover");
 type ElementWithSymbol = HTMLElement & { [symbol]?: boolean };
 
 function movingToAnotherItem(event: ReactMouseEvent<HTMLElement>) {
-  let dest = getMouseDestination(event);
-  if (!dest) return false;
+  if (!isElement(event.relatedTarget)) return false;
+  let dest: Element | null = event.relatedTarget;
   do {
     if (hasOwnProperty(dest, symbol) && dest[symbol]) return true;
     dest = dest.parentElement;
--- a/packages/ariakit-react-components/src/composite/composite-hover.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-hover.tsx
@@ -15,6 +15,7 @@ import {
   hasFocusWithin,
   hasOwnProperty,
   invariant,
+  isElement,
   removeUndefinedValues,
 } from "@ariakit/utils";
 import type { BooleanOrCallback } from "@ariakit/utils";
@@ -28,10 +29,7 @@ type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
 
 function getMouseDestination(event: ReactMouseEvent<HTMLElement>) {
-  const relatedTarget = event.relatedTarget as Node | null;
-  if (relatedTarget?.nodeType === Node.ELEMENT_NODE) {
-    return relatedTarget as Element;
-  }
+  if (isElement(event.relatedTarget)) return event.relatedTarget;
   return null;
 }
 
--- a/packages/ariakit-react-components/src/composite/composite-hover.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-hover.tsx
@@ -38,8 +38,9 @@ const symbol = Symbol("composite-hover");
 type ElementWithSymbol = HTMLElement & { [symbol]?: boolean };
 
 function movingToAnotherItem(event: ReactMouseEvent<HTMLElement>) {
-  if (!isElement(event.relatedTarget)) return false;
-  let dest: Element | null = event.relatedTarget;
+  const { relatedTarget } = event;
+  if (!isElement(relatedTarget)) return false;
+  let dest: Element | null = relatedTarget;
   do {
     if (hasOwnProperty(dest, symbol) && dest[symbol]) return true;
     dest = dest.parentElement;
--- a/packages/ariakit-utils/readme.md
+++ b/packages/ariakit-utils/readme.md
@@ -42,6 +42,7 @@ This package is ESM-only and exposes a single public entrypoint.
   - [`getActiveElement`](#getactiveelement)
   - [`contains`](#contains)
   - [`isElement`](#iselement)
+  - [`isNode`](#isnode)
   - [`isFrame`](#isframe)
   - [`isButton`](#isbutton)
   - [`isVisible`](#isvisible)
@@ -316,7 +317,7 @@ function isElement(target: EventTarget | null | undefined): target is Element;
 
 Checks whether the given event target is an element.
 
-`event.target` and `event.relatedTarget` are `EventTarget`s, which aren't necessarily nodes — for example `window` or an `XMLHttpRequest` when an event is dispatched programmatically. Calling `Element`-only methods such as `contains` on those throws, so guard with this before treating them as elements.
+`event.target` and `event.relatedTarget` are `EventTarget`s, which aren't necessarily elements — for example `window` or an `XMLHttpRequest` when an event is dispatched programmatically. Calling `Element`-only methods such as `hasAttribute` on those throws, so guard with this before treating them as elements. When you only need a `Node` — for example to call `contains` — use `isNode` instead.
 
 It tests `nodeType` rather than `instanceof Element` so that elements coming from same-origin child frames (which `addGlobalEventListener` also listens on) aren't wrongly rejected for belonging to a different realm.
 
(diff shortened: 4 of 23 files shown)


packages/ariakit-react-components/src/composite/composite-hover.tsx
```diff
@@ -28,10 +29,7 @@ type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
 
 function getMouseDestination(event: ReactMouseEvent<HTMLElement>) {
-  const relatedTarget = event.relatedTarget as Node | null;
-  if (relatedTarget?.nodeType === Node.ELEMENT_NODE) {
-    return relatedTarget as Element;
-  }
+  if (isElement(event.relatedTarget)) return event.relatedTarget;
   return null;
 }
```
diegohaz: Is this function still necessary? Isn't it simpler if we use isElement directly instead of getMouseDestination?

packages/ariakit-utils/src/dom.ts
```diff
@@ -80,6 +80,31 @@ export function contains(parent: Node, child: Node): boolean {
...
+ *
+ * It tests `nodeType` rather than `instanceof Element` so that elements coming
+ * from same-origin child frames (which `addGlobalEventListener` also listens
+ * on) aren't wrongly rejected for belonging to a different realm.
+ * @example
+ * if (isElement(event.target)) {
+ *   contains(element, event.target);
+ * }
+ */
+export function isElement(
```
diegohaz, line 101: `isElement` fixes the new call sites, but the shared event helpers still have the same unchecked `EventTarget` -> `Node` cast. `isFocusEventOutside` passes `event.relatedTarget as HTMLElement` into `contains()`, and `isPortalEvent` does the same with `event.target`; those helpers are exported and used by Focusable, Portal, Combobox, and Composite paths. A programmatic focus/blur event with `relatedTarget: new EventTarget()` can still reach `parent.contains(child)` and throw the same "parameter 1 is not of type Node" error this PR is trying to eliminate, just outside the dialog-specific listeners.

Could we apply this guard in `packages/ariakit-utils/src/events.ts` before shipping the helper? For example, `isFocusEventOutside` can treat a non-element `relatedTarget` as outside before calling `contains`, and `isPortalEvent` can avoid calling `contains` unless the target is an element.

diegohaz (commented):
Initial full review of PR #6087 at `0c9fc5990bb9437cde3feefae196d01b144a2c8d`, using merge base `22e11feb891b51043058f00caba89dc8ebd58cb5`.

Scope reviewed: PR description, linked issue #5156, existing review thread, the full diff, changed Dialog/HovercardDisclosure/CompositeHover/Focusable paths, new `@ariakit/utils` helper/docs, new browser sandboxes/tests, changesets, and nearby shared event utilities that still call `contains()` on event targets.

Validation: inspected with `gh pr view`, `gh issue view`, `git diff`, and `rg`. I did not run local tests in this worktree because dependencies are not installed here; the GitHub status rollup for the reviewed head shows the relevant app/main/perf/release checks passing.

Main finding: one correctness gap remains in shared `@ariakit/utils` event helpers, where the same non-Node `EventTarget` can still be passed to `contains()` outside the newly guarded dialog-specific listeners.

app/src/sandbox/hovercard-5156/test-browser.ts
```diff
@@ -0,0 +1,34 @@
...
+    q,
+  }) => {
+    const errors: string[] = [];
+    page.on("pageerror", (error) => errors.push(error.message));
+
+    // Focusing the disclosure shows it, which attaches its global focusout
+    // listener to the document.
+    await q.button("More details").focus();
+    await test.expect(q.button("More details")).toBeVisible();
+
```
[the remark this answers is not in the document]
diegohaz: Fixed in 2a1292b00 — the test now opens the hovercard (focus the disclosure, then press <kbd>Enter</kbd>) and waits for its content (`Toolkit for building accessible web apps.`), which is a real display-gated signal that the global `focusout` listener is mounted before the synthetic `focusout` is dispatched.

Note the `aria-expanded="true"` suggestion would not work here: focusing the `HovercardDisclosure` sets its internal `visible` state (which mounts the listener) but does not open the hovercard, so `aria-expanded` stays `"false"`. Verified the new precondition still fails with the original `Failed to execute 'contains'` TypeError when the library guard is reverted (Chromium, Firefox, and WebKit).

diegohaz:
**Fixed**

- 2a1292b00 ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327387873)): Strengthened the `hovercard-5156` regression precondition. It now opens the hovercard and waits for its content — a real display-gated signal that the global `focusout` listener is mounted — instead of asserting `toBeVisible()` on the 1px visually-hidden disclosure. The suggested `aria-expanded="true"` gate would be a false negative here, since focusing the `HovercardDisclosure` mounts the listener without opening the hovercard. Verified the precondition still fails with the original `contains` TypeError when the library guard is reverted (Chromium/Firefox/WebKit).

packages/ariakit-utils/src/dom.ts
```diff
@@ -80,6 +80,31 @@ export function contains(parent: Node, child: Node): boolean {
...
 }
 
+/**
+ * Checks whether the given event target is an element.
+ *
+ * `event.target` and `event.relatedTarget` are typed as `Element`, but at
+ * runtime they can be any `EventTarget` (for example `window` or an
+ * `XMLHttpRequest` when an event is dispatched programmatically). Calling
+ * `Element`-only methods such as `contains` on those would throw.
+ *
```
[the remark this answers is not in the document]
diegohaz: Fixed in 3ec7573c1 — reworded the JSDoc: `event.target` and `event.relatedTarget` are `EventTarget`s (not `Element`), and the point of `isElement` is that they may be non-nodes.

packages/ariakit-utils/readme.md
```diff
@@ -307,6 +308,30 @@ contains(document.getElementById("parent"), document.getElementById("child"));
...
 
+#### `isElement`
+
+```ts
+function isElement(target: EventTarget | null | undefined): target is Element;
+```
+
+Checks whether the given event target is an element.
+
+`event.target` and `event.relatedTarget` are typed as `Element`, but at runtime they can be any `EventTarget` (for example `window` or an `XMLHttpRequest` when an event is dispatched programmatically). Calling `Element`-only methods such as `contains` on those would throw.
```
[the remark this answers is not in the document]
diegohaz: Fixed in 3ec7573c1 — regenerated the readme from the corrected JSDoc via `ariakit docs --write`.

packages/ariakit-react-components/src/composite/composite-hover.tsx
diegohaz: Fixed in 11e7cb1e9 — removed `getMouseDestination` (it had become a thin `isElement` wrapper) and used `isElement` directly in `hoveringInside` and `movingToAnotherItem`.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in 6a8c1dff3 — you're right, my earlier scoping was too narrow. Both `isFocusEventOutside` and `isPortalEvent` are now guarded with `isElement`: a programmatically dispatched focus/blur event with a non-node `relatedTarget`/`target` reaches them through the Focusable/Combobox/Composite/Portal listeners and threw the same `contains` TypeError. Added unit tests in `events.test.ts` (verified they fail with the original TypeError when the guards are reverted) and a changeset.

diegohaz:
**Fixed**

- `3ec7573c1` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327480335), [reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327480391)): Reworded the `isElement` JSDoc — `event.target`/`event.relatedTarget` are `EventTarget`s (not `Element`); the point is they may be non-nodes. Regenerated the readme.
- `11e7cb1e9` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327480447)): Removed `getMouseDestination` (a thin `isElement` wrapper) and used `isElement` directly in `CompositeHover`.
- `6a8c1dff3` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327480549)): Hardened the shared `isFocusEventOutside` and `isPortalEvent` helpers with `isElement`. The same non-element-target `TypeError` reaches `Focusable`/`Combobox`/`Composite`/`Portal` via programmatically dispatched focus events, so the earlier scoping was too narrow. Added `events.test.ts` regression tests (verified they fail with the original `contains` TypeError when the guards are reverted) and a changeset.

.changeset/300-dialog-non-node-interact-outside.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) throwing a `TypeError` when an event with a non-element target (such as one dispatched on `window` or an `XMLHttpRequest`) reached its global document listeners while the dialog was open.
```
diegohaz: I believe we can combine these 3 changesets (dialog, hovercard, and portal) into a single titled multiline changeset.

diegohaz (commented):
Re-review of PR #6087 at `6a8c1dff3ccaf7e8d373ac85a1005fced5ea0d9b`.

Baseline: latest prior marked review head `0c9fc5990bb9437cde3feefae196d01b144a2c8d`. I reviewed the four new commits since then: the hovercard test precondition update, `isElement` documentation wording, the `CompositeHover` simplification, and the shared `isFocusEventOutside`/`isPortalEvent` guards plus tests and changeset.

Existing threads: verified the prior shared-helper finding is fixed in current code. `isFocusEventOutside` now treats non-element `relatedTarget` as outside before calling `contains()`, and `isPortalEvent` guards non-element `target` before calling `contains()`. I also checked the resolved/outdated hovercard and doc wording threads against the current files.

Validation run locally after installing dependencies in this worktree with `pnpm install --frozen-lockfile --ignore-scripts`:

```sh
pnpm test packages/ariakit-utils/src/events.test.ts
```

Result: 1 test file passed, 2 tests passed. I also ran `git diff --check 0c9fc5990bb9437cde3feefae196d01b144a2c8d..HEAD` with no whitespace errors. GitHub CI still had several checks in progress at the time of this review, with the already-completed lint/type/build-related checks passing.

No new substantive findings from this re-review.

packages/ariakit-react-components/src/composite/composite-hover.tsx
```diff
@@ -27,26 +28,18 @@ const TagName = "div" satisfies ElementType;
...
 
 const symbol = Symbol("composite-hover");
 type ElementWithSymbol = HTMLElement & { [symbol]?: boolean };
 
 function movingToAnotherItem(event: ReactMouseEvent<HTMLElement>) {
-  let dest = getMouseDestination(event);
-  if (!dest) return false;
+  if (!isElement(event.relatedTarget)) return false;
+  let dest: Element | null = event.relatedTarget;
```
diegohaz: Shouldn't we do this?
```suggestion
  let dest = event.relatedTarget;
  if (!isElement(dest)) return false;
```

.changeset/300-dialog-non-node-interact-outside.md
diegohaz: Done in 631829c54 — combined the Dialog, HovercardDisclosure, and focus/portal changesets into a single titled changeset (`.changeset/300-non-node-event-target.md`). Kept the new `isElement`/`isNode` utils in their own `200-` entry.

packages/ariakit-react-components/src/composite/composite-hover.tsx
diegohaz: Done in a8bf77963 — read `relatedTarget` once via destructuring. Note your exact form (`let dest = event.relatedTarget; if (!isElement(dest)) return false;`) tripped `TS2339: Property 'parentElement' does not exist on type 'EventTarget'` — the do-while back-edge (`dest = dest.parentElement`) re-widens the inferred `EventTarget | null` binding, dropping the `isElement` narrowing. Using a typed `let dest: Element | null = relatedTarget` after the guard keeps it compiling.

packages/ariakit-utils/src/events.ts
```diff
@@ -12,10 +12,13 @@ import { isApple } from "./platform.ts";
...
-    event.currentTarget &&
-    !contains(event.currentTarget as Node, event.target as Element),
-  );
+  const { currentTarget, target } = event;
+  if (!currentTarget) return false;
+  // A non-element target can't be contained by `currentTarget`, so treat it the
+  // same as a target rendered outside of it (e.g. through a portal) instead of
+  // letting `contains` throw on a non-node.
+  if (!isElement(target)) return true;
+  return !contains(currentTarget as Node, target);
```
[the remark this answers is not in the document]
diegohaz: Fixed in 4f23d8c6e — you're right that `contains` only needs a `Node`. Added an `isNode` helper and switched the contains-only guards — `isFocusEventOutside`, `isPortalEvent`, and (for the same reason) the Dialog Escape and HovercardDisclosure focusout listeners — to it, so non-element nodes keep their original `contains` behavior and only non-node `EventTarget`s (e.g. `window`/`XMLHttpRequest`) short-circuit. `isElement` stays where element-only methods (`tagName`/`hasAttribute`) are called. Added regression tests for both the non-node and the contained text-node cases.

diegohaz:
**Fixed**

- `631829c54` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327694330)): Combined the Dialog, HovercardDisclosure, and focus/portal non-element-target changesets into a single titled changeset.
- `a8bf77963` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327694363)): Read `relatedTarget` once in `CompositeHover`'s `movingToAnotherItem` (the literal `let dest = …` suggestion tripped a TS error via the do-while back-edge; details in the thread).
- `4f23d8c6e` ([reply](https://github.com/ariakit/ariakit/pull/6087#discussion_r3327694421)): `contains()` only needs a `Node`, so guarding it with `isElement` over-narrowed (a contained text node would be treated as outside). Added an `isNode` helper and switched the contains-only guards — `isFocusEventOutside`, `isPortalEvent`, the Dialog Escape listener, and the HovercardDisclosure focusout listener — to it; non-node `EventTarget`s (window/XHR) still short-circuit, and `isElement` stays where element-only methods (`tagName`/`hasAttribute`) are called. Added regression tests for both the non-node and contained-text-node cases, and clarified the `isElement`/`isNode` docs.

tsc clean; full vitest suite (881) passing.

PR #6097, opened 2026-05-30
Remove shadcn registry scaffolding from the app

diegohaz:
## Motivation

The `app` workspace carried shadcn-based registry scaffolding: a registry API route, an empty placeholder module left over from the workspace rename, a `shadcn` dependency, and supporting bits (an unused icon, a hardcoded `npx shadcn ... /r/...` install button, and a middleware allow-list entry for the `/r/` route). Per maintainer direction, this should be removed entirely for now — a new version of the registry will be built in the future. Keeping the half-wired scaffolding around adds dependency weight and dead code with no current consumer.

## Solution

Remove the shadcn/registry scaffolding from the `app` workspace completely:

- Delete `app/src/pages/r/[...registry].ts` — the shadcn registry API route (imported `shadcn/schema`).
- Delete `app/src/lib/registry.ts` — a 0-byte empty placeholder, unused and not covered by the package's `./*` export.
- Remove `"shadcn": "4.8.1"` from `app/package.json` and regenerate `pnpm-lock.yaml` (the removed lockfile lines are entirely shadcn's dependency subtree).
- Remove the `shadcn` icon from `app/src/icons/icons.ts` (it had no usages), plus the `terminal` icon whose only consumer was the shadcn-install button being removed.
- Remove the `npx shadcn@latest add https://ariakit.com/r/disclosure.json` button block from `app/src/pages/[...example].astro`.
- Remove the `isPublicRoute` helper and its call site from `app/src/middleware.ts`. Its sole purpose was to let the `/r/` registry route bypass Clerk auth; with the registry gone, no public routes remain and the remaining auth flow is unchanged.

No changeset: `app` is a private workspace (`"private": true`), so this is not a published, user-facing change.

## Verification

- `pnpm tsc` (runs `pnpm -F app run check` → `wrangler types` + `astro check`, then `tsc -b`): passes with 0 errors, 0 warnings, 0 hints.
- `pnpm lint-fix`: clean.
- Exhaustive `git grep -i shadcn` and `git grep -n "/r/"` across the repo: no remaining references (the only `registry` match left is an unrelated "Central registry of patterns" comment in `source.ts`). No dangling imports in the edited files; all four middleware imports are still used.

## Review note

A pre-commit review flagged that, after removing the second button, the surrounding `<div class="grid gap-2">` in `[...example].astro` now wraps a single child, making `grid gap-2` cosmetically redundant. This was intentionally left as-is: removing `grid` would change the single button's width (grid-track stretch vs. default), a visual change that cannot be validated locally because Ariakit visual tests run only in CI, and the wrapper is the intended layout slot for the button group that the rebuilt registry will repopulate. It is dead styling only, with no correctness impact.


app/src/icons/icons.ts
```diff
@@ -302,11 +294,3 @@ export const group: Icon = {
...
   html: `<path stroke-linecap="round" stroke-linejoin="round" d="M18 18.72a9.094 9.094 0 0 0 3.741-.479 3 3 0 0 0-4.682-2.72m.94 3.198.001.031c0 .225-.012.447-.037.666A11.944 11.944 0 0 1 12 21c-2.17 0-4.207-.576-5.963-1.584A6.062 6.062 0 0 1 6 18.719m12 0a5.971 5.971 0 0 0-.941-3.197m0 0A5.995 5.995 0 0 0 12 12.75a5.995 5.995 0 0 0-5.058 2.772m0 0a3 3 0 0 0-4.681 2.72 8.986 8.986 0 0 0 3.74.477m.94-3.197a5.971 5.971 0 0 0-.94 3.197M15 6.75a3 3 0 1 1-6 0 3 3 0 0 1 6 0Zm6 3a2.25 2.25 0 1 1-4.5 0 2.25 2.25 0 0 1 4.5 0Zm-13.5 0a2.25 2.25 0 1 1-4.5 0 2.25 2.25 0 0 1 4.5 0Z" />`,
 };
-
-export const terminal: Icon = {
-  size: 24,
-  stroke: "currentColor",
-  fill: "none",
-  strokeWidth: 2,
-  html: `<g stroke-linecap="round" stroke-linejoin="round"><path d="M12 19h8"/><path d="m4 17 6-6-6-6"/></g>`,
-};
```
diegohaz, line 312: We can leave the icons here

app/src/pages/[...example].astro
```diff
@@ -108,19 +108,6 @@ const descriptionText = await descriptionToText(description?.body, framework);
...
-              <Icon name="terminal" />
-              <span class="overflow-hidden text-sm"
-                >npx shadcn@latest add https://ariakit.com/r/disclosure.json</span
-              >
-              <kbd
-                class="ak-layer bg-transparent ak-ink-0 justify-self-end relative after:absolute after:h-full after:end-full after:w-10 after:bg-linear-to-l after:from-(--ak-layer) after:from-20%"
-              >
-                <span class="font-sans">⌘</span>S
-              </kbd>
-            </button>
```
Comment (diegohaz) [inline app/src/pages/[...example].astro:123]: Leave the button here.

app/src/icons/icons.ts
diegohaz: Restored in 09da36457 — the `shadcn` and `terminal` icon exports are back, byte-identical to `main`. They are local inline-SVG `Icon` objects with no `shadcn` npm import, so keeping them does not re-add the removed dependency.

app/src/pages/[...example].astro
Comment (diegohaz) [reply app/src/pages/[...example].astro:123]: Restored in 09da36457 — the install button is back, byte-identical to `main` (the `<Icon name="terminal" />` + `npx shadcn@latest add https://ariakit.com/r/disclosure.json` block). It is plain markup, so it does not re-add the shadcn dependency.

Heads up: this PR deletes the `/r/[...registry].ts` route, so the `https://ariakit.com/r/disclosure.json` URL in the button now 404s. I left it as-is per your request — treating it as intentional placeholder scaffolding for the future registry rebuild. Let me know if you would prefer the URL/text adjusted in the meantime.

PR #6101, opened 2026-05-30
Document tab-5147 Next.js page as a build-time cacheComponents repro

diegohaz:
## Motivation

`nextjs/app/tab-5147/page.tsx` is a manual reproduction of a build-time regression (issue #5147): rendering tabs and panels with an explicit tab-store `id` under Next.js `cacheComponents` broke the build. The page had no explanation of its purpose, so it read like an orphaned/dead page even though it intentionally guards that regression.

## Solution

Keep the page as a build-time repro and make its purpose self-documenting:

- Add a comment to `nextjs/app/tab-5147/page.tsx` explaining that it guards the issue #5147 regression, that the regression is build-time, and that the page is therefore exercised whenever the Next.js app is built in CI (the `Build Next.js app` job runs `pnpm -F nextjs run build-worker` with `cacheComponents: true`). Keep the explicit store `id` and the two tab/panel pairs that trigger it.

The page keeps its original `tab-5147` path so the file name continues to mention the originating issue. No runtime sandbox/preview is added, because the regression manifests at build time rather than at runtime.

No changeset: `nextjs` is a private, changeset-ignored workspace and this only documents an existing build-time repro page.


nextjs/app/tab-nextjs/page.tsx
```diff
@@ -2,6 +2,9 @@
 
 import * as Ariakit from "@ariakit/react";
 
+// Guards a regression where rendering tabs and panels with an explicit store
+// `id` under Next.js `cacheComponents` would break the tab/panel wiring. Keeping
+// the explicit `id` and two tab/panel pairs is what exercises that path.
```
diegohaz: Actually, we don't need to promote this example because the original issue happened at build time. We can remove the sandbox from this PR and just update the comment here to mention that it's tested when the Next.js app is built.

diegohaz: Fixed in 268f56efd - removed the `tab-nextjs` sandbox (`preview.mdx` and `test-browser.ts`) and updated this comment to explain the regression is build-time and exercised whenever the Next.js app is built in CI (the `Build Next.js app` job runs `pnpm -F nextjs run build-worker` with `cacheComponents: true`). Kept the explicit store `id` and the two tab/panel pairs that trigger it, and tidied the leftover `aria-label`.

diegohaz:
**Addressed review feedback**

- `268f56efd` ([reply](https://github.com/ariakit/ariakit/pull/6101#discussion_r3328565072)): Per @diegohaz's request, this PR no longer promotes the example to a runtime sandbox. Since the regression is build-time, removed the `tab-nextjs` sandbox (`preview.mdx` + `test-browser.ts`) and updated the page comment to explain it is exercised whenever the Next.js app is built in CI. Kept the explicit store `id` and the two tab/panel pairs that trigger it, and tidied the leftover `aria-label`.

The page stays at `nextjs/app/tab-nextjs/` (id already referenced in `app/src/lib/nextjs.ts` and `nextjs.test.ts`). PR title/description updated to reflect the build-time-repro framing.

nextjs/app/tab-5147/page.tsx
```diff
@@ -2,12 +2,16 @@
```
diegohaz, line 2: Why did you rename the file? It should mention the issue.

diegohaz: Renamed it back to `nextjs/app/tab-5147/page.tsx` in 7f8a3c8e3 so the file name mentions the issue. I had originally renamed it to `tab-nextjs` only because the earlier version of this PR promoted it to a runtime preview example, and the preview redirect (`isNextjsPreviewId`) only forwards ids containing "nextjs". Once we pivoted to keeping it as a build-time repro (no sandbox/preview), that constraint no longer applies, so the issue-based name is the right choice. I also added a comment referencing issue #5147 so the page is self-documenting.

diegohaz: Done in 28ac9c280 — the page keeps its original `nextjs/app/tab-5147/page.tsx` path so the file name mentions the issue, and I added a comment referencing issue #5147 explaining the build-time `cacheComponents` regression it guards. (I had briefly renamed it to `tab-nextjs` only because an earlier version of this PR promoted it to a runtime preview example, whose redirect requires a "nextjs" id; once we kept it as a build-time repro instead, that constraint went away.) The net diff vs main is now just this comment.

PR #6106, opened 2026-05-30
Document form removeValue null slots

diegohaz:
## Motivation

`removeValue` preserves array indices by replacing removed array slots with `null`, but its reference docs only said it removes a value from an array. Consumers could assume it splices and be surprised by null entries and unchanged length.

## Solution

Document the null-slot behavior in the FormStore JSDoc: length is preserved, removed slots become `null`, indices stay stable for field keys, touched state, errors, and `FormRemove` focus handling, and consumers should filter `null` values before submission if needed.

This also updates the `FormRemove` reference examples because they iterate the same array shape that `removeValue` leaves behind. The examples now skip `null` slots while preserving the original index used by `FormInput` and `FormRemove`.

## Testing

- `pnpm lint-fix`
- Pre-commit PR review gate: Codex reviewer clean; Cursor reviewer clean

### DIFF
--- a/packages/ariakit-components/src/form/form-store.ts
+++ b/packages/ariakit-components/src/form/form-store.ts
@@ -476,8 +476,9 @@ export interface FormStoreFunctions<
    * The array length is preserved: the removed index is replaced with `null`
    * so indices stay stable for field keys, touched state, errors, and
    * [`FormRemove`](https://ariakit.com/reference/form-remove) focus handling.
-   * Filter out `null` values before submitting if your payload should omit
-   * removed items.
+   * Validation and submission callbacks receive these `null` entries in
+   * `state.values`, so filter them only when your payload should omit removed
+   * items.
    * @example
    * store.removeValue("tags", 0);
    * store.removeValue("tags", 1);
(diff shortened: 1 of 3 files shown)


packages/ariakit-components/src/form/form-store.ts
```diff
@@ -472,6 +472,12 @@ export interface FormStoreFunctions<
   pushValue: <T>(name: StringLike, value: T) => void;
   /**
    * Removes a value from an array field.
+   *
+   * The array length is preserved: the removed index is replaced with `null`
+   * so indices stay stable for field keys, touched state, errors, and
+   * [`FormRemove`](https://ariakit.com/reference/form-remove) focus handling.
+   * Filter out `null` values before submitting if your payload should omit
```
diegohaz: Is this filter really needed? Can you check in the code?

diegohaz: Fixed in 14d17d1bb244591ed963afe1ccd5579bdef92e96. Confirmed the code path: `removeValue` stores `null` in `state.values`; `validate()` and `submit()` pass `form.getState()` directly to callbacks; and `Form` calls `store.submit()` without filtering. The JSDoc now states that validation/submission callbacks receive these `null` entries in `state.values`, so callers should filter only when their payload should omit removed items.

diegohaz:
**Fixed**

- `14d17d1bb244591ed963afe1ccd5579bdef92e96` ([reply](https://github.com/ariakit/ariakit/pull/6106#discussion_r3328710791)): clarified the `removeValue` JSDoc after checking that `removeValue` stores `null`, validation/submission callbacks receive `form.getState()`, and `Form` calls `store.submit()` without filtering.

PR #6110, opened 2026-05-30
Flatten nested hovercard movement check

diegohaz:
## Motivation

Nested hovercards are registered into a flattened list, but `isMovingOnHovercard` still used a recursive call when checking that list. Since the recursive call did not forward the nested list, it only repeated checks that were already covered by the flat iteration.

## Solution

Replace the recursive nested hovercard check with a direct membership test that keeps a hovercard visible when a registered nested hovercard contains focus or the current pointer target.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/packages/ariakit-react-components/src/hovercard/hovercard.tsx
+++ b/packages/ariakit-react-components/src/hovercard/hovercard.tsx
@@ -65,7 +65,7 @@ function isMovingOnHovercard(
   // The mouse is moving on an element inside the anchor element.
   if (anchor && contains(anchor, target)) return true;
   // The mouse is moving on an element inside a nested hovercard.
-  if (nested?.some((card) => isMovingOnHovercard(target, card, anchor))) {
+  if (nested?.some((card) => hasFocusWithin(card) || contains(card, target))) {
     return true;
   }
   return false;


diegohaz (commented):
**Review: comment-only. Initial review** of the full PR (baseline = merge-base `2f258a62` with `main`, head `0cc679eb`).

Scope: `isMovingOnHovercard` in `packages/ariakit-react-components/src/hovercard/hovercard.tsx`, its call site, the nested-hovercard registration flow, and the `contains`/`hasFocusWithin` utils.

**Verdict: correct, behavior-preserving — no blocking findings.**

I traced the equivalence this change relies on rather than trusting the diff. The old recursive call `isMovingOnHovercard(target, card, anchor)` passed `nested` as `undefined`, so it reduced to `hasFocusWithin(card) || contains(card, target) || contains(anchor, target)`. The outer-anchor term is dead inside `.some()`: the outer frame already returns at `if (anchor && contains(anchor, target)) return true;` *before* reaching the `nested?.some(...)` line, so that term is always `false` when the callback runs. The new `hasFocusWithin(card) || contains(card, target)` is therefore exactly equivalent.

Also confirmed:
- `nestedHovercards` is a flattened list (each hovercard re-forwards descendant registrations to its ancestors via `registerOnParent`), so dropping the recursion can't miss deeper cards — it was strictly redundant work, matching the PR's stated motivation.
- No mirror implementation to keep in sync (no Solid hovercard and no `react-core` copy of this function on this branch).

Not separately run: `pnpm tsc` (type-preserving one-line change; `target` is already narrowed to a `Node` before the edited line and the expression reuses calls already present in the file). No changeset expected for a behavior-neutral refactor.

PR #6111, opened 2026-05-30
Remove radio store rest destructure

diegohaz:
## Motivation

`createRadioStore` destructures its props parameter into a rest object without extracting any keys. This creates an identical shallow copy before the function uses it as the original props object, unlike sibling stores that only destructure when they extract named keys.

## Solution

Use a direct `props: RadioStoreProps = {}` parameter and keep the existing references unchanged. This removes the no-op copy while preserving the default value and runtime behavior.

### DIFF
--- a/packages/ariakit-components/src/radio/radio-store.ts
+++ b/packages/ariakit-components/src/radio/radio-store.ts
@@ -12,9 +12,7 @@ import { createCompositeStore } from "../composite/composite-store.ts";
 /**
  * Creates a radio store.
  */
-export function createRadioStore({
-  ...props
-}: RadioStoreProps = {}): RadioStore {
+export function createRadioStore(props: RadioStoreProps = {}): RadioStore {
   const syncState = props.store?.getState();
 
   const composite = createCompositeStore({


diegohaz (commented):
**Review: comment-only. Initial review** (baseline = merge-base `2f258a62` with `main`, head `b625e219`).

Scope: `createRadioStore` in `packages/ariakit-components/src/radio/radio-store.ts`, plus sibling `create*Store` signatures.

**Verdict: correct, safe no-op refactor — no findings.**

The removed `{ ...props }` rest pattern created a defensive shallow copy. I confirmed `props` is read-only throughout the body (only `props.store?.getState()`, reads of `props.focusLoop`/`value`/`defaultValue`, and a fresh `{ ...props, ... }` spread into `createCompositeStore` — the caller's object is never reassigned, mutated, or passed by reference to anything that mutates it), so passing it directly is equivalent, and `= {}` still applies on `undefined`.

The change also aligns radio with the established convention: stores that extract named keys keep the rest pattern (`select`, `tab`), while stores that extract nothing use a plain param (`dialog`, `form`, and the structurally identical `tag`). Radio was the lone divergent case.

No mirror to sync — the `react-components` `useRadioStore` is a separate hook that legitimately reassigns `props`. Package typecheck passed. No changeset or test needed (no API or behavior change).

PR #6112, opened 2026-05-30
Decouple tooltip skip timeout sync

diegohaz:
## Motivation

The tooltip anchor active-store coordination sync subscribes to both `mounted` and `skipTimeout`, but `skipTimeout` is only used after the tooltip becomes hidden. When `skipTimeout` changes while the tooltip is open, the callback re-runs the mounted branch and re-reads the active tooltip store without changing behavior.

## Solution

Subscribe the coordination sync only to `mounted` and read `skipTimeout` lazily with `store.getState().skipTimeout` when scheduling the hidden-state timeout.

Changing `skipTimeout` while the tooltip is already hidden no longer re-schedules the pending skip-window timeout with the new duration. That edge behavior is undocumented and untested; the normal mounted/hidden transition still uses the current `skipTimeout` at scheduling time.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -106,6 +106,8 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           // timeout to hide the active tooltip in the global store. This is so
           // we can show other tooltips without a delay when there's already an
           // active tooltip (see the showOnHover method below).
+          // Read skipTimeout lazily because this sync only subscribes to
+          // mounted.
           const id = setTimeout(removeStore, store.getState().skipTimeout);
           return () => clearTimeout(id);
         }),
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -87,7 +87,7 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
         // the component unmounts. This is useful, for example, to avoid
         // showing tooltips immediately on serial tests.
         removeStore,
-        sync(store, ["mounted", "skipTimeout"], (state) => {
+        sync(store, ["mounted"], (state) => {
           if (!store) return;
           // If the current tooltip is open, we should immediately hide the
           // active one and set the current one as the active tooltip. If a
@@ -106,7 +106,7 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           // timeout to hide the active tooltip in the global store. This is so
           // we can show other tooltips without a delay when there's already an
           // active tooltip (see the showOnHover method below).
-          const id = setTimeout(removeStore, state.skipTimeout);
+          const id = setTimeout(removeStore, store.getState().skipTimeout);
           return () => clearTimeout(id);
         }),
       );


packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
```diff
@@ -106,7 +106,7 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
           // timeout to hide the active tooltip in the global store. This is so
           // we can show other tooltips without a delay when there's already an
           // active tooltip (see the showOnHover method below).
-          const id = setTimeout(removeStore, state.skipTimeout);
+          const id = setTimeout(removeStore, store.getState().skipTimeout);
```
diegohaz, line 111: **Optional (maintainability):** this lazy `store.getState().skipTimeout` read is now load-bearing relative to the `sync(store, ["mounted"])` key list above. Since the listener no longer subscribes to `skipTimeout`, reading it from the synced `state` would look like an untracked-dependency bug to a future reader — and "fixing" that by re-adding `skipTimeout` to the keys would silently reintroduce the redundant re-runs this PR removes. A one-line note documenting the intent would prevent that regression:

```ts
// Read skipTimeout lazily: this listener only subscribes to `mounted`, so it
// must not list `skipTimeout` as a sync dependency just to read it here.
const id = setTimeout(removeStore, store.getState().skipTimeout);
```

For context, I confirmed this is purely a clarity concern, not correctness: in `@ariakit/store` the listener receives the current full snapshot at fire time (`state === store.getState()`), so `state.skipTimeout` and `store.getState().skipTimeout` return the same value on every `mounted` transition. Not blocking.

diegohaz (commented):
**Review: comment-only. Initial review** (baseline = merge-base `2f258a62` with `main`, head `a3b1c14b`).

Scope: the active-store coordination `sync` in `packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx`, the `@ariakit/store` listener machinery, and all `skipTimeout` usages/tests.

**Verdict: correct for every documented and tested path — no blocking findings. One optional maintainability nit posted inline.**

Confirmed against `@ariakit/store` internals that the listener receives the current full snapshot at fire time (`setState` assigns the new `state` before `runListeners`, and `getState()` returns that same object), so `store.getState().skipTimeout` equals the old `state.skipTimeout` on every `mounted` transition. Narrowing the keys to `["mounted"]` only stops redundant re-fires; the mounted branch that previously re-ran is idempotent (`setState` short-circuits when the active store is unchanged).

The only behavior delta is the one the PR description already calls out: changing `skipTimeout` while the tooltip is *already hidden* no longer reschedules the in-flight skip-window timeout. Grepped the repo — `skipTimeout` appears only in `tooltip-anchor.tsx` and the two `tooltip-store.ts` files; no test/example/doc exercises mid-hidden-window rescheduling, so nothing depends on it.

`pnpm tsc` passed; `state` is still used (the `mounted` branch). No mirror to sync. A changeset is a judgment call given the tiny undocumented edge delta, but it's behavior-neutral for documented APIs — left to author discretion, not flagged as a finding.

packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
diegohaz: Fixed in 75ffa976f - documented why `skipTimeout` is read lazily from the store when this sync subscribes only to `mounted`.

diegohaz:
**Fixed**

- 75ffa976f (https://github.com/ariakit/ariakit/pull/6112#discussion_r3329490295): Added a concise comment explaining that `skipTimeout` is read lazily because the tooltip sync only subscribes to `mounted`.

PR #6113, opened 2026-05-30
Remove Heading native check memoization

diegohaz:
## Motivation

`useHeading` memoizes `isNativeHeading`, but the value is just a constant-time regex test on the current tag name. Keeping `useMemo` there adds an unnecessary hook slot and render-time setup for a trivial computation.

## Solution

Compute `isNativeHeading` directly from `tagName` and remove the now-unused `useMemo` import from React.

### DIFF
--- a/packages/ariakit-react-components/src/heading/heading.tsx
+++ b/packages/ariakit-react-components/src/heading/heading.tsx
@@ -7,7 +7,7 @@ import {
 } from "@ariakit/react-utils";
 import type { Options, Props } from "@ariakit/react-utils";
 import type { ElementType } from "react";
-import { useContext, useMemo, useRef } from "react";
+import { useContext, useRef } from "react";
 import { HeadingContext } from "./heading-context.tsx";
 import type { HeadingLevels } from "./utils.ts";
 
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Review: comment-only. Initial review** (baseline = merge-base `2f258a62` with `main`, head `b3954cc0`).

Scope: `useHeading` in `packages/ariakit-react-components/src/heading/heading.tsx`.

**Verdict: correct, safe no-op refactor — no findings.**

`isNativeHeading` is a `boolean` consumed only in two ternaries (`role`, `aria-level`); it is never placed in a dependency array, and identity is irrelevant for a primitive, so dropping `useMemo` changes nothing observable. The regex `/^h\d$/` is an inline literal recreated each render both before and after — the memo only cached the resulting boolean — so there is no regex-allocation regression; the only delta is one fewer hook slot. `useMemo` has no other use in the file, so removing it from the `react` import is correct (`useContext`/`useRef` remain used). The hook-count change is permanent and unconditional, so there is no rules-of-hooks concern.

Noted but intentionally not flagged: the Solid `useHeading` still uses `createMemo`, but there it's an idiomatic reactive primitive feeding reactive getters rather than a render-time cache — leaving it as-is is correct, not a divergence. No changeset or test needed.

PR #6114, opened 2026-05-30
Skip bare var nodes before rendering

diegohaz:
## Motivation

The `@ariakit/tailwind` build currently includes bare `Var` nodes in the same array that feeds rendering. Those nodes render to empty strings and are then removed from the final CSS, which couples output rendering to identifier validation.

## Solution

Keep `OutputNode` unchanged for validation, so `assertUniqueIdentifiers` still receives bare vars and the full nested node tree. Narrow only the render path with a `RenderableOutputNode` guard that skips `Type.Var` before calling `renderNode`, then remove the `renderVar` empty-string special case and the final empty-string filter.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- Pre-commit native and Cursor read-only reviews: no findings

Residual risk: `@ariakit/tailwind` does not currently have a package-level unit test harness for `output()`, so the validation/render split is covered by build output stability and TypeScript narrowing rather than a focused unit test.

### DIFF
--- a/packages/ariakit-tailwind/src/build.ts
+++ b/packages/ariakit-tailwind/src/build.ts
@@ -12,6 +12,7 @@ import { Type } from "./lib.ts";
 type BlockChild = AtRule["children"][number];
 type DeclarationNode = Extract<BlockChild, { type: Type["Declaration"] }>;
 type OutputNode = CustomProperty | CustomVariant | BlockChild | Var;
+type RenderableOutputNode = Exclude<OutputNode, Var>;
 type IdentifierType = Type["Var"] | Type["CustomProperty"];
 
 const INDENT = "  ";
@@ -113,11 +114,6 @@ function renderCustomVariant(variant: CustomVariant, depth: number) {
   );
 }
 
-function renderVar(_variable: Var): string {
-  // Vars are references used inside values, not standalone CSS statements.
-  return "";
-}
-
 function renderBlockChild(child: BlockChild, depth: number) {
   switch (child.type) {
     case Type.Declaration:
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Review: comment-only. Initial review** (baseline = merge-base `2f258a62` with `main`, head `b4fa74c8`).

Scope: `output()` and the render pipeline in `packages/ariakit-tailwind/src/build.ts`.

**Verdict: correct, behavior-preserving — no blocking findings.**

Validated by regenerating the shipped artifact on both refs: `node packages/ariakit-tailwind/src/build.ts` on `b4fa74c8` and on `origin/main` produce a byte-identical `output.css` (same sha256; `diff` empty; `git diff origin/main...HEAD -- output.css` empty). `pnpm -F @ariakit/tailwind build` and `pnpm tsc` both pass.

The two filter strategies are equivalent. I traced every render function reachable from the top-level node array (`renderCustomProperty`, `renderCustomVariant`/`renderBlock`, `renderBlockChild` → declaration/rule/at-rule) — none can return `""`. The only producer of an empty string was `renderVar`, so the old `.filter((value) => value.length > 0)` dropped exactly the bare vars, which the new `.filter(isRenderableOutputNode)` now drops before rendering. The var filter is load-bearing, not dead code: `input` spreads real bare `Var` nodes at the top level. `assertUniqueIdentifiers(nodes)` still runs on the full unfiltered array first (validation, including its `Type.Var` identifier case, is unaffected), and `RenderableOutputNode = Exclude<OutputNode, Var>` keeps the `renderNode` switch exhaustive so `assertNever` still type-checks. The type-guard function is justified — it narrows the element type for `renderNode`, which an inline predicate would not.

No changeset needed (the build script is internal and the output is byte-identical). The author's noted residual risk (no unit-test harness for `output()`) is adequately covered here by the byte-identical proof plus the type narrowing.

PR #6117, opened 2026-05-30
Derive disclosure aria-expanded synchronously

diegohaz:
## Motivation

A default-open [`Disclosure`](https://ariakit.com/reference/disclosure) rendered `aria-expanded="false"` on the server and first client paint because the value was initialized with local state and corrected only after an effect. This exposed an incorrect accessibility state while the content was already open.

## Solution

Derive `aria-expanded` directly from the disclosure store's `open` state and the current disclosure button element, while keeping the effect focused on claiming a missing or disconnected disclosure element. The current button element is tracked through the merged ref callback so render does not read `ref.current`.

This preserves shared-store multi-button ownership once a disclosure element has been claimed while making the initial rendered state match `defaultOpen`.

The tests now cover server-rendered default-open output, shared-store multi-button ownership, and reclaiming a disconnected disclosure element on programmatic open. The disclosure group browser test also asserts that unopened buttons remain collapsed while opened items are expanded.

### DIFF
--- a/.changeset/300-disclosure-default-open-expanded.md
+++ b/.changeset/300-disclosure-default-open-expanded.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Disclosure`](https://ariakit.com/reference/disclosure) to render `aria-expanded="true"` on the server when initially open.
--- a/.changeset/300-disclosure-default-open-expanded.md
+++ b/.changeset/300-disclosure-default-open-expanded.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Disclosure`](https://ariakit.com/reference/disclosure) to render `aria-expanded="true"` on the server when initially open.
+Fixed [`Disclosure`](https://ariakit.com/reference/disclosure) and disclosure-based components to render `aria-expanded="true"` on the server and the initial client render when initially open.
--- a/packages/ariakit-react-components/src/disclosure/disclosure.tsx
+++ b/packages/ariakit-react-components/src/disclosure/disclosure.tsx
@@ -12,7 +12,7 @@ import type { Props } from "@ariakit/react-utils";
 import { invariant } from "@ariakit/utils";
 import type { BooleanOrCallback } from "@ariakit/utils";
 import type { ElementType, MouseEvent } from "react";
-import { useEffect, useRef } from "react";
+import { useEffect, useState } from "react";
 import type { ButtonOptions } from "../button/button.tsx";
 import { useButton } from "../button/button.tsx";
 import { useDisclosureProviderContext } from "./disclosure-context.tsx";
@@ -46,19 +46,19 @@ export const useDisclosure = createHook<TagName, DisclosureOptions>(
         "Disclosure must receive a `store` prop or be wrapped in a DisclosureProvider component.",
     );
 
-    const ref = useRef<HTMLType>(null);
+    const [element, setElement] = useState<HTMLType | null>(null);
     const disclosureElement = useStoreState(store, "disclosureElement");
     const open = useStoreState(store, "open");
     const expanded =
-      open && (disclosureElement == null || disclosureElement === ref.current);
+      open && (disclosureElement == null || disclosureElement === element);
 
     // Assigns the disclosure element whenever it's undefined or disconnected
     // from the DOM. Re-run on open changes so a stale disconnected disclosure
     // element is claimed before deriving aria-expanded for the next open state.
     useEffect(() => {
       if (disclosureElement?.isConnected) return;
-      store?.setDisclosureElement(ref.current);
-    }, [disclosureElement, store, open]);
+      store?.setDisclosureElement(element);
+    }, [disclosureElement, store, open, element]);
 
     const onClickProp = props.onClick;
     const toggleOnClickProp = useBooleanEvent(toggleOnClick);
@@ -80,7 +80,7 @@ export const useDisclosure = createHook<TagName, DisclosureOptions>(
       "aria-controls": contentElement?.id,
       ...metadataProps,
       ...props,
-      ref: useMergeRefs(ref, props.ref),
+      ref: useMergeRefs(setElement, props.ref),
       onClick,
     };
 
--- a/packages/ariakit-react-components/src/disclosure/disclosure.tsx
+++ b/packages/ariakit-react-components/src/disclosure/disclosure.tsx
@@ -12,7 +12,7 @@ import type { Props } from "@ariakit/react-utils";
 import { invariant } from "@ariakit/utils";
 import type { BooleanOrCallback } from "@ariakit/utils";
 import type { ElementType, MouseEvent } from "react";
-import { useEffect, useRef, useState } from "react";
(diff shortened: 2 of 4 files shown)


packages/ariakit-react-components/src/disclosure/disclosure.tsx
```diff
@@ -47,21 +47,17 @@ export const useDisclosure = createHook<TagName, DisclosureOptions>(
     );
 
     const ref = useRef<HTMLType>(null);
-    const [expanded, setExpanded] = useState(false);
     const disclosureElement = useStoreState(store, "disclosureElement");
     const open = useStoreState(store, "open");
+    const expanded =
+      open && (disclosureElement == null || disclosureElement === ref.current);
```
diegohaz: **Question (non-blocking) — shared store + `defaultOpen` on first paint:** the `disclosureElement == null` branch is what makes a single open disclosure render `aria-expanded="true"` before the claiming effect runs — i.e. the actual fix, good. But for the rarer case of one store shared across multiple `Disclosure` buttons that is initially open (`defaultOpen`), `disclosureElement` is `null` on the server and first client paint, so **every** button derives `expanded = true` until the first effect claims an owner and the rest collapse.

Old behavior was the inverse (all `false`, then one `true`). Both are transiently wrong for non-owner buttons for a single pre-hydration frame and both self-correct, so this is not a blocker — and reporting "expanded" while the content is actually open is arguably the more defensible a11y default. Flagging it only because the consequence of the `== null` branch for shared stores is non-obvious: please confirm it's the intended trade-off. If you want it pinned down, a `defaultOpen` shared-store case would be a natural addition to the new `marks only the active shared-store disclosure as expanded` test — the current one starts closed, so it never exercises the all-expanded first paint.

.changeset/300-disclosure-default-open-expanded.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Disclosure`](https://ariakit.com/reference/disclosure) to render `aria-expanded="true"` on the server when initially open.
```
diegohaz: **Nit (optional):** this release note credits the fix only "on the server," but the PR description (accurately) notes the wrong `aria-expanded="false"` was also shown on the **first client render before the effect ran** — that's the user-visible half for non-SSR apps. Consider widening the wording, e.g.:

> Fixed [`Disclosure`](https://ariakit.com/reference/disclosure) to render `aria-expanded="true"` on the server and the initial client render when initially open.

Not blocking — the current text isn't wrong, just narrower than the actual fix.

diegohaz (commented):
**Review: comment-only. Initial review** (baseline = merge-base `2f258a62` with `main`, head `66737cba`). Reviewed in an isolated worktree and validated by running the affected test surface. Two non-blocking nits posted inline.

Scope: `disclosure.tsx` (the fix), the new `app/src/examples/disclosure/test.ts` cases, the `group/test-browser.ts` assertions, and the changeset.

**Verdict: correct and well-targeted — no blocking findings.**

Deriving `expanded = open && (disclosureElement == null || disclosureElement === ref.current)` during render (instead of `useState(false)` corrected by an effect) fixes the SSR/first-paint `aria-expanded="false"` while content is open. Confirmed across cases:
- **Single / SSR:** `ref.current` and `disclosureElement` are both `null` on the server → `expanded = open`. The new SSR test fails on the old source and passes on the new.
- **Steady state:** `disclosureElement` is only ever (re)assigned to a non-null connected `ref.current`, so `== null` is true only in the transient pre-claim window — no persistently wrong value.
- **Shared store / detached-reclaim:** traced against the (unchanged) onClick `setDisclosureElement(currentTarget) + toggle`. The `open` entry in the effect deps is **load-bearing** — removing it makes the reclaim test fail, because after the owner unmounts only `open` changes on a programmatic `show()`.

**Blast radius** (`useDisclosure` is the base for the Dialog/Popover/Combobox/Menu/Select/Tooltip disclosures): `pnpm tsc` ✅; `pnpm test disclosure` ✅ (25 tests, incl. the 3 new); broad `pnpm test dialog popover menu combobox select tooltip hovercard` ✅ (884). A control run on the **old** source failed **only** the new SSR test — proving no `aria-expanded` regression anywhere in the blast radius and that the new test is a real guard. oxlint/oxfmt ✅ (note: `react-hooks/exhaustive-deps` isn't enabled, and extra deps wouldn't be flagged regardless). Solid has no disclosure component, so nothing to mirror. Changeset (`@ariakit/react-components` + `@ariakit/react`, both `patch`) is correct.

Inline nits: (1) the `== null` branch makes a shared store + `defaultOpen` + multiple buttons read all-expanded on first paint — acceptable trade-off, flagged for confirmation; (2) the changeset wording could also mention the first client render. The browser test (`group/test-browser.ts`) was statically reviewed, not executed (Playwright is heavy in a worktree); its assertions are correct (the group uses an independent store per item, so opened items stay expanded while others remain collapsed).

packages/ariakit-react-components/src/disclosure/disclosure.tsx
diegohaz: Reading ref.current in the render phase is not allowed. I wonder if this report should be rejected.

diegohaz: Fixed in d2b6bd098 - agreed this was a valid blocker. `Disclosure` no longer reads `ref.current` during render; it stores the current element from the merged callback ref and derives `aria-expanded` from that element state instead.

.changeset/300-disclosure-default-open-expanded.md
diegohaz: Fixed in d2b6bd098 - broadened the changeset to mention disclosure-based components and both server output and the initial client render.

packages/ariakit-react-components/src/disclosure/disclosure.tsx
diegohaz: Fixed in d2b6bd098 - the implementation now avoids the render-phase `ref.current` read by tracking the current disclosure element through the merged ref callback. The shared-store/defaultOpen first render trade-off remains intentional: before any element has been claimed there is no owner to compare against, so the initial state follows the visible open content and then settles to the claimed owner after effects run.

diegohaz:
**Fixed**

- d2b6bd098 ([render-phase ref thread](https://github.com/ariakit/ariakit/pull/6117#discussion_r3329519090)): Reworked `Disclosure` to track the current button through the merged callback ref and derive `aria-expanded` from element state, so render no longer reads `ref.current`.
- d2b6bd098 ([shared-store/defaultOpen thread](https://github.com/ariakit/ariakit/pull/6117#discussion_r3329519093)): Kept the initial shared-store trade-off explicit while moving the ownership comparison to element state; once a disclosure element is claimed, only the owner is expanded.
- d2b6bd098 ([changeset thread](https://github.com/ariakit/ariakit/pull/6117#discussion_r3329519092)): Broadened the changeset to mention disclosure-based components and both server output and the initial client render.

Verification after the fix:

- `pnpm lint-fix`
- `pnpm test app/src/examples/disclosure/test.ts`
- `pnpm test-react-18 app/src/examples/disclosure/test.ts`
- `pnpm tsc`

PR #6120, opened 2026-05-31
Add type-safe findInOrder overloads

diegohaz:
## Motivation

`findInOrder` allowed property-key callers to pass values unrelated to the selected property type. That left mismatched comparisons like `item[key] === value` unchecked by TypeScript.

## Solution

Add overloads so property-key calls require `values` to be `T[K][]`, while selector calls remain generic over the selector return type. The runtime implementation is unchanged, including the `value == null` guard.

The PR also adds focused app lib coverage for ordered matching, null skipping, and invalid value types.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test app/src/lib/array.test.ts`

### DIFF
--- a/app/src/lib/array.ts
+++ b/app/src/lib/array.ts
@@ -7,11 +7,21 @@
  *
  * SPDX-License-Identifier: UNLICENSED
  */
+export function findInOrder<T, K extends keyof T>(
+  array: T[],
+  key: K,
+  values: T[K][],
+): T | null;
 export function findInOrder<T, Value>(
   array: T[],
-  key: keyof T | ((item: T) => Value),
+  key: (item: T) => Value,
   values: Value[],
-) {
+): T | null;
+export function findInOrder<T>(
+  array: T[],
+  key: keyof T | ((item: T) => unknown),
+  values: unknown[],
+): T | null {
   for (const value of values) {
     const item = array.find((item) => {
       if (value == null) return false;
--- a/app/src/lib/array.ts
+++ b/app/src/lib/array.ts
@@ -7,11 +7,23 @@
  *
  * SPDX-License-Identifier: UNLICENSED
  */
+/**
+ * Returns the first item whose property value matches a value in `values`.
+ *
+ * Values are tried in order, so the match for the earliest value wins.
+ * `null` and `undefined` values are skipped.
+ */
 export function findInOrder<T, K extends keyof T>(
   array: T[],
   key: K,
   values: T[K][],
 ): T | null;
+/**
+ * Returns the first item whose selected value matches a value in `values`.
+ *
+ * Values are tried in order, so the match for the earliest value wins.
+ * `null` and `undefined` values are skipped.
+ */
 export function findInOrder<T, Value>(
   array: T[],
   key: (item: T) => Value,
--- a/app/src/lib/array.test.ts
+++ b/app/src/lib/array.test.ts
@@ -1,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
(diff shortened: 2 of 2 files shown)


app/src/lib/array.ts
```diff
@@ -7,11 +7,21 @@
  *
  * SPDX-License-Identifier: UNLICENSED
  */
+export function findInOrder<T, K extends keyof T>(
```
diegohaz, line 16: **Optional (maintainability):** `findInOrder` is an exported helper with non-obvious semantics that this PR makes more prominent, yet it still has no JSDoc. Two behaviors aren't discoverable from the signatures alone:

- `values` are tried **in priority order** — it returns the item matching the *earliest* value in `values`, not the first matching item in `array`.
- `null`/`undefined` entries in `values` are **silently skipped** (the `value == null` guard), so e.g. `findInOrder(items, "label", [null, "one"])` ignores the `null` and matches `"one"`. That's a real gotcha now that the property-key overload's `values: T[K][]` legitimately allows `null` when `T[K]` includes it.

Per `ariakit-general-code-style` (JSDoc for top-level/public functions and non-obvious behavior), a short doc comment would pay for itself:

```suggestion
/**
 * Returns the first item whose `key` (a property name or a selector) equals a
 * value in `values`. `values` are tried in order, so the match for the
 * earliest value wins; `null`/`undefined` entries in `values` are skipped.
 * Returns `null` when nothing matches.
 */
export function findInOrder<T, K extends keyof T>(
```

Not blocking — purely a docs nicety.

diegohaz (commented):
**Initial review** of `audit/t091-find-in-order-overloads` — diff from merge-base `2803a29` to head `ae098304`. Files: `app/src/lib/array.ts`, `app/src/lib/array.test.ts`.

**Verified in an isolated worktree at the PR head (Node 24, pnpm 11.4):**
- `pnpm tsc` → clean (exit 0). Because `noUnusedLocals` is on and the test relies on `@ts-expect-error`, a green `tsc` proves both negative cases genuinely error — `findInOrder(items, "id", ["1"])` (overload #1, `string` vs `number`) and `findInOrder(items, (item) => item.id, ["1"])` (overload #2 `Value` inference conflict). The `@ts-expect-error` directives are load-bearing, not spurious.
- `vitest run app/src/lib/array.test.ts` → 2/2 passing.
- **Real caller compatibility:** the only production caller, `app/src/lib/stripe.ts:231` `findInOrder(prices, "key", keys)`, still compiles — `PriceData["key"]` resolves to `string` and `keys` is `string[]`, so the stricter first overload (`values: T[K][]`) is satisfied. No regression.
- **Behavior preserved:** runtime body is unchanged; ordered-priority matching and the `value == null` skip are intact (confirmed by the `label`/`null` test).

**Approach:** The 2-overload + implementation-signature pattern is the right call here — a single mapped/conditional signature can't cleanly disambiguate property-key vs selector while binding the value type to `T[K]`. No simpler behavior-preserving alternative, and no correctness, compatibility, performance, or security concerns. The motivation is real: the old single signature let property-key callers pass values unrelated to `T[K]`.

**Findings:** one **optional** maintainability nit (inline) — add JSDoc documenting the non-obvious ordered-priority + null-skipping semantics. Nothing blocking.

No changeset needed (`app` is a private workspace). Comment-only review.

app/src/lib/array.ts
diegohaz: Fixed in dfc33c7c9 - documented the priority-order and null/undefined-skipping semantics on both findInOrder overloads, with property-value wording for the key overload and selected-value wording for the selector overload.

diegohaz:
**Fixed**

- dfc33c7c9: Documented `findInOrder` ordering and null/undefined-skipping semantics on both overloads, addressing the unresolved JSDoc review thread.

PR #6121, opened 2026-05-31
Remove Shiki token htmlStyle any cast

diegohaz:
## Motivation

`getTokenColors` cast Shiki token `htmlStyle` to `any` before reading color CSS properties, but Shiki already exposes the needed optional record type. The cast bypasses type checking unnecessarily.

## Solution

Use `token.htmlStyle` directly and keep the existing optional color lookups with the `token.color` fallback, preserving runtime behavior.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/app/src/components/code-block-content.astro
+++ b/app/src/components/code-block-content.astro
@@ -134,7 +134,7 @@ const MULTI_TOKEN_HIGHLIGHT_CLASS = "ak-layer ak-layer-6 ak-edge-25";
 const LINE_DIV_BASE_CLASS = "pe-14 [content-visibility:auto] h-(--line-height)";
 
 function getTokenColors(token: ThemedToken) {
-  const style = token.htmlStyle as any;
+  const style = token.htmlStyle;
   const dark = style?.["--shiki-dark"] || token.color;
   const light = style?.color || token.color;
   return { dark, light } as const;


diegohaz (commented):
**Initial review** — diff from merge-base `2803a29` to head `4cf14432`. File: `app/src/components/code-block-content.astro`.

- `pnpm -F app run check` (astro check) → `0 errors, 0 warnings, 0 hints`; root `pnpm tsc` → exit 0.
- The crux holds: the app resolves `shiki@4.1.0` → `@shikijs/types@4.1.0`, where `ThemedToken.htmlStyle` is `Record<string, string> | undefined`. So both `style?.["--shiki-dark"]` and `style?.color` are valid accesses without the cast, and the `as any` was genuinely redundant.
- **Behavior preserved**: runtime is byte-identical (same `token.htmlStyle`, same keys, same `|| token.color` fallback, same `as const`). Only the static type of `style` changes from `any` to `Record<string, string> | undefined`; all four consumers just interpolate `{ dark, light }` into a CSS-var style string, so the typed value (`string | undefined`) behaves identically. Dropping the cast is the cleanest fix and removes an `any` leak into `dark`/`light`.

PR #6122, opened 2026-05-31
Type JSDoc tags in jsdoc loader

diegohaz:
## Motivation

`JsDocInfo` stores JSDoc tags as `any[]`, so tag consumers bypass compiler checks when reading tag names and comments from ts-morph.

## Solution

Import `JSDocTag` from ts-morph, store collected tags as `JSDocTag[]`, and remove the `any`-typed tag callbacks. The tag comment consumers now use `getCommentText()`, which keeps comment text normalized while avoiding the structured return shape of `getComment()`.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/app/src/lib/jsdoc-loader.ts
+++ b/app/src/lib/jsdoc-loader.ts
@@ -13,7 +13,7 @@ import { dirname, join, resolve, sep } from "node:path";
 import { invariant } from "@ariakit/utils";
 import type { LoaderContext } from "astro/loaders";
 import type { z } from "astro/zod";
-import type { FunctionLikeDeclaration, SourceFile } from "ts-morph";
+import type { FunctionLikeDeclaration, JSDocTag, SourceFile } from "ts-morph";
 import { Node, Project, ts } from "ts-morph";
 import { createLogger } from "./logger.ts";
 import type { FrameworkSchema, Reference } from "./schemas.ts";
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Initial review** — diff from merge-base `2803a29` to head `a03a2fa9`. File: `app/src/lib/jsdoc-loader.ts`.

- `pnpm tsc` (root = `pnpm -F app run check && tsc -b`) → exit 0; `astro check` clean. The loader actually executed during the check (`Loaded 173 references for react`, `10 for solid`) and the output still satisfied the zod `ReferenceSchema`, so the changed paths were exercised end-to-end.
- ts-morph resolves to `28.0.0`, where `JSDocTag.getCommentText(): string | undefined` exists; `getTags()` returns `JSDocTag[]`, so `JsDocInfo.tags: JSDocTag[]` and the inferred `tag` callbacks are sound. Under `strict: true`, removing the `(tag: any)` annotations is a real type-safety gain.

**Behavior of `getComment()` → `getCommentText()` (the one behavior-relevant change):** confirmed equivalent for this codebase, and a strict improvement otherwise.
- For plain-string tag comments (the realistic case for `@deprecated`/`@default`/`@example`), `getCommentText()` returns the text verbatim including newlines — so `String(getComment())` and `getCommentText()` are byte-identical, and the `@example` parsing regex is unaffected (verified empirically against the installed ts-morph).
- Grepping the processed sources (`packages/ariakit-*/src`) found **zero** inline `{@link}`/`{@linkcode}` tags, so no current input takes the structured-array path → generated references are identical before/after.
- For the only divergent case (a future inline `{@link}` inside one of these tags), the old `String(getComment())` produced `[object Object]…` garbage while `getCommentText()` yields the correct text — i.e. this also fixes a latent bug.

PR #6123, opened 2026-05-31
Add Next.js type checking to CI

diegohaz:
## Motivation

The Next.js app could accumulate type errors without CI coverage. The app workflow built the OpenNext artifact without running the `nextjs` workspace type-check script, and the Next config was set to ignore TypeScript build errors.

## Solution

Run `pnpm -F nextjs run tsc` in the app workflow before building the Next.js artifact. Remove `typescript.ignoreBuildErrors` now that both the standalone type-check and OpenNext build pass with TypeScript validation enabled, and declare `typescript` in the `nextjs` workspace because its `tsc` script invokes it directly.

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -44,6 +44,9 @@ jobs:
       - name: Set up environment
         uses: ./.github/workflows/setup
 
+      - name: Type check Next.js app
+        run: pnpm -F nextjs run tsc
+
       - name: Build Next.js app
         run: pnpm -F nextjs run build-worker
 
--- a/nextjs/next.config.ts
+++ b/nextjs/next.config.ts
@@ -4,9 +4,6 @@ import type { NextConfig } from "next";
 const config: NextConfig = {
   reactCompiler: true,
   typedRoutes: true,
-  typescript: {
-    ignoreBuildErrors: true,
-  },
   cacheComponents: true,
 
   // Allow cross-origin iframe embedding from the Astro app
--- a/nextjs/package.json
+++ b/nextjs/package.json
@@ -30,6 +30,7 @@
     "@types/react": "19.2.15",
     "@types/react-dom": "19.2.3",
     "babel-plugin-react-compiler": "1.0.0",
-    "tailwindcss": "4.3.0"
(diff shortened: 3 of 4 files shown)


diegohaz (commented):
**Initial review** — diff from merge-base `2803a29` to head `06137fc0`. Files: `.github/workflows/app.yml`, `nextjs/next.config.ts`, `nextjs/package.json`, `pnpm-lock.yaml`.

- Ran the exact CI command `pnpm -F nextjs run tsc` (= `next typegen && tsc`) → **exit 0**, zero type errors. `tsc` resolves to the newly-declared `typescript@6.0.3` in the nextjs workspace and runs with `noEmit`. This proves removing `typescript.ignoreBuildErrors` is safe today.
- `pnpm install` produced **no** further `pnpm-lock.yaml` drift — the committed lockfile is in sync, and the only delta is the intended `typescript 6.0.3` entry under the nextjs importer.
- **Coverage justification holds**: root `tsc` (`pnpm -F app run check && tsc -b`) references `app` but not nextjs and never runs `next typegen`, so the nextjs app was genuinely unchecked. The dedicated step is justified, not redundant — and it's the only nextjs typecheck across all workflows.
- **Dependency rule**: nextjs is private, so the exact-pinned `typescript 6.0.3` is correct, and declaring it explicitly (its `tsc` script invokes the binary directly) is right per "declare what you use." `6.0.3` matches root/app/template.

**Design**: the step is in `build-nextjs` (which already runs `setup`), placed before `build-worker` so type errors fail fast and gate the build; the job runs on PRs and main pushes, so it actually gates merges. Serial-before-build is fine — the check is cheap and reuses the paid setup; a separate parallel job isn't worth the duplicated checkout/setup. The standalone `tsc` is also a *stronger* gate than `next build`'s graph-scoped check (it covers all tsconfig files).

**Findings:** none blocking. One optional, pre-existing, out-of-scope note: `website` pins `typescript@6.0.2` vs `6.0.3` everywhere else — a minor skew to align separately, not in this PR. No changeset needed (CI + private-workspace dep). Comment-only.

PR #6124, opened 2026-05-31
Name CustomCheckbox forwardRef

diegohaz:
## Motivation

The focusable-click-focus sandbox wraps an input with `React.forwardRef`, but the render function is anonymous, so React DevTools reports the component as `ForwardRef`. Its props type also widens the input props with a redundant `type` override even though the component always renders a checkbox input.

## Solution

Name the `forwardRef` render function `CustomCheckbox` and type its props directly as `ComponentProps<"input">`. The component still renders an input with `type="checkbox"` while avoiding the extra type widening.

### DIFF
--- a/app/src/sandbox/focusable-click-focus/index.react.tsx
+++ b/app/src/sandbox/focusable-click-focus/index.react.tsx
@@ -2,10 +2,11 @@ import * as Ariakit from "@ariakit/react";
 import type { ComponentProps } from "react";
 import { forwardRef, useState } from "react";
 
-const CustomCheckbox = forwardRef<
-  HTMLInputElement,
-  ComponentProps<"input"> & { type?: string }
->((props, ref) => <input ref={ref} type="checkbox" {...props} />);
+const CustomCheckbox = forwardRef<HTMLInputElement, ComponentProps<"input">>(
+  function CustomCheckbox(props, ref) {
+    return <input ref={ref} type="checkbox" {...props} />;
+  },
+);
 
 export default function Example() {
   const [checked, setChecked] = useState(false);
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Initial review** — head `bbd71174`. File: `app/src/sandbox/focusable-click-focus/index.react.tsx`.

**Verified in an isolated worktree (Node 24):** `pnpm -F app run check` (astro check) → 0 errors/warnings/hints; root `pnpm tsc` (`tsc -b`) → exit 0.

- Naming the `forwardRef` render function `CustomCheckbox` is the idiomatic way to fix the React DevTools display name (vs setting `.displayName`); no `displayName` existed or is needed.
- Dropping `& { type?: string }` is safe: `ComponentProps<"input">` already includes an optional `type`, and the only caller is `render={<CustomCheckbox />}` (no props passed), so the narrower type rejects nothing. Props still spread after `type="checkbox"`, so behavior is byte-identical.
- The sandbox's browser test queries by accessible name (`q.checkbox("Wrapped")`), independent of the component name — no test impact.

diegohaz (commented):
**Re-review** (incremental) — covers the new commit `4ddbe1a6` ("Keep CustomCheckbox type fixed") since my prior review of `bbd71174`. Baseline = the prior reviewed head.

The only change is a one-line JSX attribute reorder in `app/src/sandbox/focusable-click-focus/index.react.tsx`:
```diff
-    return <input ref={ref} type="checkbox" {...props} />;
+    return <input ref={ref} {...props} type="checkbox" />;
```
This applies CodeRabbit's nitpick — moving `type="checkbox"` after the spread so a caller-supplied `type` can't override it.

**Verified in an isolated worktree (Node 24):** `pnpm -F app run check` + `tsc -b` → exit 0. JSX permits an explicit attribute after a spread to override it, so it's type-safe.

- Behavior is unchanged in practice: `Ariakit.Checkbox` already injects `type: "checkbox"` into the merged props, and the sole caller is `render={<CustomCheckbox />}` (no `type`), so the rendered input was `type="checkbox"` under both orderings. The reorder is a sound defensive lock, appropriate for a component named `CustomCheckbox`. The browser test queries by accessible role/name and is unaffected.

**Findings:** none. The incremental change is a clean improvement. Comment-only.

PR #6125, opened 2026-05-31
Align pick implementation signature

diegohaz:
## Motivation

`pick` accepts nullable stores in its public overload and already returns early at runtime when no store is passed, but its implementation signature still requires `Store`. The sibling public helpers use optional implementation signatures, so `pick` is inconsistent.

## Solution

Update the `pick` implementation signature to accept an optional `Store`, matching the overload and existing runtime guard without changing behavior.

### DIFF
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -571,7 +571,7 @@ export function pick<
  * Creates a new store with a subset of the current store state and keeps them
  * in sync.
  */
-export function pick(store: Store, ...args: Parameters<StorePick>) {
+export function pick(store?: Store, ...args: Parameters<StorePick>) {
   if (!store) return;
   return getInternal(store, "pick")(...args);
 }


diegohaz (commented):
**Initial review** — head `91f1594e`. File: `packages/ariakit-store/src/index.ts` (one line, the `pick` implementation signature).

**Verified in an isolated worktree (Node 24):** `pnpm tsc` → exit 0; `vitest run packages/ariakit-store/src/test.ts` → 36/36 passing (`pick` is covered at runtime and via `expectTypeOf`, so the narrowed return type still resolves through the public overload).

- Making the impl signature `pick(store?: Store, …)` matches the public overload (`store?: T | null`) and the existing runtime guard `if (!store) return;`. Before this, `pick` was the lone outlier — its siblings `omit`/`sync`/`batch`/`subscribe` already use the optional `store?: Store` impl form.
- Implementation signatures aren't visible to external callers, so the public API is unchanged; purely a consistency fix with no runtime change.

**Findings:** none. No changeset needed (type-only internal signature, no public-surface/behavior change). Comment-only.

PR #6127, opened 2026-05-31
Rename trailing value variable

diegohaz:
## Motivation

The tag input implementation has one local variable named `trailingvalue`, which does not match the camelCase style used elsewhere in the file.

## Solution

Rename the local variable to `trailingValue` at its declaration and both usages without changing behavior.

### DIFF
--- a/packages/ariakit-react-components/src/tag/tag-input.tsx
+++ b/packages/ariakit-react-components/src/tag/tag-input.tsx
@@ -147,7 +147,7 @@ export const useTagInput = createHook<TagName, TagInputOptions>(
         const delimiters = getDelimiters(delimiter);
         // Split values and get the trailing value that will remain in the input
         let values = splitValueByDelimiter(value, delimiters);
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Initial review** — head `947fcc28`. File: `packages/ariakit-react-components/src/tag/tag-input.tsx`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` → exit 0 (a missed occurrence would surface as an undefined-variable error). `grep` confirms zero remaining lowercase `trailingvalue` and that `trailingValue` is used at the declaration + both usages (lines 150/166/167). The variable is a function-scoped local (not importable), applied to the canonical source.

- Pure camelCase rename; surrounding logic (`UndoManager.execute`, `store.setValue`, the `=== prevValue` early return) is byte-equivalent. No behavior change.

**Findings:** none. No changeset needed (cosmetic rename). Comment-only.

PR #6128, opened 2026-05-31
Convert GroupOptions to an interface

diegohaz:
## Motivation

`GroupOptions` is the only options export in `packages/ariakit-react-components/src/group/group.tsx` declared as a type alias to `Options`. Sibling options exports use interfaces extending their base options, and `CompositeGroupOptions`/`FormGroupOptions` extend `GroupOptions`, so this keeps the options chain consistent.

## Solution

Change `GroupOptions` to an interface that extends `Options` while preserving its generic parameter and public shape.

### DIFF
--- a/packages/ariakit-react-components/src/group/group.tsx
+++ b/packages/ariakit-react-components/src/group/group.tsx
@@ -61,7 +61,9 @@ export const Group = forwardRef(function Group(props: GroupProps) {
   return createElement(TagName, htmlProps);
 });
 
-export type GroupOptions<_T extends ElementType = TagName> = Options;
+export interface GroupOptions<
+  _T extends ElementType = TagName,
+> extends Options {}
 
 export type GroupProps<T extends ElementType = TagName> = Props<
   T,


diegohaz (commented):
**Initial review** — head `2948cccc`. File: `packages/ariakit-react-components/src/group/group.tsx`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` → exit 0 across the whole graph (including all downstream `*GroupOptions` consumers). Formatter is clean: the repo uses **oxlint + oxfmt** (not Biome/Prettier), and running `oxfmt` in write mode produces an empty diff — the multi-line interface form is exactly oxfmt's canonical output at printWidth 80, so the CI format check won't fail; oxlint doesn't warn on the underscore-prefixed `_T`.

- `Options` (`@ariakit/react-utils`) is a non-generic `interface` with two optional props — the safe case where `interface GroupOptions<_T> extends Options {}` is structurally identical to `type GroupOptions<_T> = Options`. Bidirectional assignability and `keyof` were confirmed equivalent, generic arity is preserved, and siblings `CompositeGroupOptions`/`FormGroupOptions` already use `interface … extends GroupOptions<T>` and still compile.
- Re-exported as `Ariakit.GroupOptions<T>`; no observable difference for consumers (same required/optional props, arity, assignability) — non-breaking, and it makes the options chain consistent.

**Findings:** none. No changeset needed (type-only refactor, no observable public-shape/behavior change). Comment-only.

PR #6129, opened 2026-05-31
Remove dialog tree walk isValidElement export

diegohaz:
## Motivation

`packages/ariakit-react-components/src/dialog/utils/walk-tree-outside.ts` exports a helper named `isValidElement` even though the helper only decides whether a DOM element should be visited during dialog tree walking. The exported name also reads like React's `isValidElement`, and the published subpath means removing it should be documented as a public API removal.

## Solution

Rename the helper to `shouldWalkElement`, make it module-private, and update the internal tree-walk call site. Add a breaking changeset for `@ariakit/react-components` documenting the removed named export from `@ariakit/react-components/dialog/utils/walk-tree-outside`.

No runtime behavior changes.

### DIFF
--- a/.changeset/100-walk-tree-outside-valid-element.md
+++ b/.changeset/100-walk-tree-outside-valid-element.md
@@ -1,0 +1,21 @@
+---
+"@ariakit/react-components": minor
+---
+
+Removed `isValidElement` from `dialog/utils/walk-tree-outside`
+
+**BREAKING** if you import `isValidElement` from `@ariakit/react-components/dialog/utils/walk-tree-outside`.
+
+The helper was intended for internal dialog tree walking and has been removed from the public subpath exports to avoid confusion with React's `isValidElement`.
+
+Before:
+
+```ts
+import { isValidElement } from "@ariakit/react-components/dialog/utils/walk-tree-outside";
+```
+
+After:
+
+```ts
+// No public replacement import is available.
+```
--- a/packages/ariakit-react-components/src/dialog/utils/walk-tree-outside.ts
+++ b/packages/ariakit-react-components/src/dialog/utils/walk-tree-outside.ts
@@ -23,7 +23,7 @@ function inSnapshot(id: string, element: Element) {
   } while (true);
 }
 
-export function isValidElement(
+function shouldWalkElement(
   id: string,
   element: Element,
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
**Initial review** — head `a9ba1ddf`. Files: `packages/ariakit-react-components/src/dialog/utils/walk-tree-outside.ts`, `.changeset/100-walk-tree-outside-valid-element.md`.

**Verified in an isolated worktree (Node 24):**
- `pnpm tsc` → exit 0; `pnpm -F @ariakit/react-components build` → clean, and the built `dist/dialog/utils/walk-tree-outside.{js,d.ts}` now exports exactly `{ createWalkTreeSnapshot, walkTreeOutside }` — `isValidElement` is gone, and the subpath still resolves for its two legitimate exports.
- `pnpm changeset status` → valid (`@ariakit/react-components` minor; `@ariakit/react` auto-bumps patch as a dependent — correctly not a frontmatter entry).
- Repo-wide grep: **zero** remaining references to the removed 3-arg helper outside the renamed call site. Every other `isValidElement` in the repo is React's 1-arg version (`from "react"`) — a different function. No test imports it.

**Assessment:**
- The change is a pure rename + de-export: applying only that transform to the base file reproduces the head byte-for-byte, so there's no behavior change. `shouldWalkElement` is a clear, intent-revealing name that also resolves the confusing collision with React's `isValidElement`.
- **Breaking-change handling is correct.** `@ariakit/react-components` is v0 (0.1.2) and its exports map publishes `./dialog/utils/walk-tree-outside`, so `isValidElement` was importable and its removal is genuinely breaking. Per Ariakit conventions, a v0 breaking change → `minor` frontmatter + `100-` filename prefix, with `**BREAKING**` + `Before:`/`After:` blocks — all present and well-formed. Scoping to `@ariakit/react-components` only is right (the umbrella `@ariakit/react` and the main `.` entry never re-exported this symbol).
- Leaving the `./dialog/utils/walk-tree-outside` export entry in `package.json` is correct (it still serves `walkTreeOutside`/`createWalkTreeSnapshot`), not an oversight. Document-and-remove (rather than a deprecated alias) is the right call for an obscure internal v0 util whose name was the problem.

PR #6131, opened 2026-05-31
Optimize ComboboxItemValue offset matching

diegohaz:
## Motivation

`ComboboxItemValue` computes match offsets while rendering highlighted item values. The previous `getOffsets` loop performed the same substring search twice per match and kept an unreachable `-1` guard.

## Solution

Store each `indexOf` result in a local variable, push offsets only while the result is valid, then compute the next match once after advancing the position.

This also adds example coverage for repeated matches so the multi-match highlighting path remains covered.

## Testing

- `pnpm lint-fix`
- `pnpm test examples/combobox-item-value/test.ts`
- `pnpm tsc`

### DIFF
--- a/.changeset/300-combobox-item-value-offsets.md
+++ b/.changeset/300-combobox-item-value-offsets.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`ComboboxItemValue`](https://ariakit.com/reference/combobox-item-value) rendering performance.
--- a/examples/combobox-item-value/test.ts
+++ b/examples/combobox-item-value/test.ts
@@ -12,3 +12,13 @@ test("show highlighted text", async () => {
   await press.Enter();
   expect(q.combobox()).toHaveValue("Apple");
 });
+
+test("show repeated highlighted text", async () => {
+  await press.Tab();
+  await type("p");
+  await press.ArrowDown();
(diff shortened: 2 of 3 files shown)


diegohaz (commented):
**Initial review** — head `9faa6130`. Files: `combobox-item-value.tsx`, `examples/combobox-item-value/test.ts`, `.changeset/300-combobox-item-value-offsets.md`.

**Verified in an isolated worktree (Node 24):** `pnpm test examples/combobox-item-value/test.ts` → 2/2 pass (incl. the new inline snapshot); `pnpm tsc` → clean; oxlint clean.

- **Equivalence proven.** The new `getOffsets` loop yields the same offset sequence as before — it just removes a redundant second `indexOf` per match and an unreachable `if (index !== -1)` guard (the `while` already guaranteed it). Confirmed by walkthrough and a 183k-case fuzz (base vs head, 0 mismatches). Overlap semantics (`pos = index + 1`) unchanged.
- The new "repeated highlighted text" snapshot (`A`·`p`·`p`·`le`) is the genuinely correct output for typing "p" in "Apple", covering the multi-match path — not masking a bug.
- The pre-existing empty-`value` infinite-loop hazard is identical in both versions and unreachable (the caller filters falsy values) — not introduced here.
- **Changeset correct:** `@ariakit/react` is rightly included because `combobox.ts` re-exports `ComboboxItemValue` from its public entry; `300-`/`patch` is right for a v0 micro-opt with no behavior change.

PR #6132, opened 2026-05-31
Cache source package lookups by directory

diegohaz:
## Motivation

The source plugin resolves package metadata for every external source file it encounters. Its package cache was keyed by the resolved file path, so files from the same package directory still triggered repeated `readPackageUpSync` walks. Misses were also not cached, so negative lookups repeated the same filesystem work.

## Solution

Add a shared package resolver used by both package name and version lookups. The resolver keys cached results by `dirname(source)` and stores `null` for misses, so repeated calls for the same source directory reuse both positive and negative lookup results.

### DIFF
--- a/app/src/lib/source-plugin.ts
+++ b/app/src/lib/source-plugin.ts
@@ -35,7 +35,10 @@ const APP_LIB_PATH = join(import.meta.dirname, "../examples/_lib");
 const NEXTJS_LIB_PATH = join(import.meta.dirname, "../../../nextjs/components");
 
 // Cache for package information to avoid repeated lookups
-const packageCache = new Map<string, any>();
+const packageCache = new Map<
+  string,
+  NonNullable<ReturnType<typeof readPackageUpSync>> | null
+>();
 
 // TypeScript compiler host for resolving modules
 const host = ts.createCompilerHost({});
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Initial review** — head `1e20aebd`. File: `app/src/lib/source-plugin.ts` (private `app` workspace).

**Verified in an isolated worktree (Node 24):** `pnpm tsc` → clean; `source-plugin.test.ts` → 6/6 pass. `dirname` is already imported (line 13), so the de-sugared call compiles.

- **`cwd: source` → `cwd: dirname(source)` is resolution-equivalent.** Checked against the installed `read-package-up@12 → find-up-simple@1.0.1`: `findUpSync` resolves `cwd`, stats `join(dir,"package.json")` with `throwIfNoEntry:false`, and walks up. Starting from a file path just does one extra falsy stat (`<file>/package.json`) before walking up to the same nearest `package.json` — identical result, one fewer wasted stat now.
- **Directory keying is correct**: both call sites pass file paths, and all files in a directory share the same nearest ancestor `package.json`, so keying by `dirname` is sound and increases cache hits.
- **Negative caching is correct**: `getPackage` now stores `null` on miss and uses `packageCache.has(...)` to distinguish cached-miss from uncached (the old `get() || readPackageUpSync()` never cached misses). `get(dir) ?? null` is type-safe. Consumer behavior (`"latest"`/`null` fallbacks, the version warning) is unchanged.
- Typing improved: precise `Map` type replaces `any`; shared `getPackage` removes duplicated lookup+set logic.

**Findings:** none. No changeset needed (`app` private). Comment-only.

PR #6133, opened 2026-05-31
Pre-group perf comparison rows

diegohaz:
## Motivation

The perf comparison markdown builders repeatedly scan the same comparison rows while iterating over test keys. That makes summary, detailed, and node comparison table generation do unnecessary work as the number of compared rows grows.

## Solution

Build a `rowsByKey` map once in `formatMarkdown`, keyed by the existing `rowKey(row)` value, and pass it to the markdown table builders. The summary table, detailed breakdown, node comparison table, and significant-key derivation now reuse per-key row groups instead of filtering or scanning the full row list for each key.

No changeset is included because `@ariakit/scripts` is private and ignored by changesets.

## Verification

- `pnpm lint-fix`
- `pnpm vitest packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm tsc`
- `pnpm test`

### DIFF
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -700,6 +700,20 @@ function rowKey(row: ComparisonRow): string {
   return `${row.testFile}::${row.label}`;
 }
 
+function groupRowsByKey(rows: ComparisonRow[]) {
+  const grouped = new Map<string, ComparisonRow[]>();
+  for (const row of rows) {
+    const key = rowKey(row);
+    const rowsForKey = grouped.get(key);
+    if (rowsForKey) {
+      rowsForKey.push(row);
+    } else {
+      grouped.set(key, [row]);
+    }
+  }
+  return grouped;
+}
+
 function formatRoundList(rounds: number[]): string {
   if (rounds.length === 0) return "-";
   return rounds.join(", ");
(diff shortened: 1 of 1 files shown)


diegohaz (commented):
**Initial review** — head `f0d97a9b`. File: `packages/ariakit-scripts/src/perf-compare.ts` (private, changeset-ignored).

**Verified in an isolated worktree (Node 24):** `pnpm vitest run …/perf-compare.test.ts` → 25/25 pass (node mode covered by 5 tests); `pnpm tsc` → clean.

- **Output equivalence proven** (the refactor must be byte-identical):
  - The one real ordering risk — `formatNodeComparisonTables` building `visibleRows` by key-order instead of original-row order — **cannot occur**: in `compare()`, `rows` is built per-key (outer loop over the `current` map, keyed identically to `rowKey`), so all rows for a key are contiguous and keys never interleave. Old filter-order and new group-order coincide.
  - `allKeys` (`[...rowsByKey.keys()]`) equals the old `[...new Set(rows.map(rowKey))]` (first-occurrence order); `significantKeys` and the per-key `rowsByKey.get(key) ?? []` are order-identical to the old `filter`s; `testRows[0]?.label` unchanged.
  - There are no snapshot tests, so beyond reasoning I diffed base-vs-head CLI output on node + Chrome fixtures (incl. an interleaved multi-key/same-file node case) — **byte-identical** (matching md5s).
- The `formatDetailedBreakdown` 4-param → `FormatDetailedBreakdownParams` object refactor follows the code-style rule; loops/ifs use braces + early `continue`. The change is a real O(rows×keys) → O(rows) win with a single Map allocation.

**Findings:** none. No changeset needed (`@ariakit/scripts` private/ignored). Comment-only.

PR #6134, opened 2026-05-31
Read Radio store id non-reactively

diegohaz:
## Motivation

`useRadio` subscribed to the radio store once for checked state and again through a selector callback just to read the store id used as the default native radio name. We can keep the existing store subscription style while making the static id read more direct.

## Solution

Use the `useStoreState(store, "id")` key selector for the default radio name. This preserves the idiomatic `useStoreState` pattern while avoiding a custom selector callback for the store id.

### DIFF
--- a/.changeset/200-radio-store-id.md
+++ b/.changeset/200-radio-store-id.md
@@ -1,6 +1,0 @@
----
-"@ariakit/react-components": patch
-"@ariakit/react": patch
----
-
-Improved [`Radio`](https://ariakit.com/reference/radio) rendering performance.
--- a/.changeset/200-radio-store-id.md
+++ b/.changeset/200-radio-store-id.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`Radio`](https://ariakit.com/reference/radio) rendering performance.
--- a/packages/ariakit-react-components/src/radio/radio.tsx
+++ b/packages/ariakit-react-components/src/radio/radio.tsx
@@ -14,7 +14,7 @@ import type { Props } from "@ariakit/react-utils";
 import { disabledFromProps, removeUndefinedValues } from "@ariakit/utils";
 import type { BivariantCallback } from "@ariakit/utils";
 import type { ChangeEvent, ElementType, FocusEvent, MouseEvent } from "react";
-import { useEffect, useMemo, useRef } from "react";
+import { useEffect, useRef } from "react";
 import type { CompositeItemOptions } from "../composite/composite-item.tsx";
 import { useCompositeItem } from "../composite/composite-item.tsx";
 import { useRadioContext } from "./radio-context.tsx";
@@ -73,7 +73,7 @@ export const useRadio = createHook<TagName, RadioOptions>(function useRadio({
   // Use the store's id as the default name to ensure radios in different
   // groups have unique names. This prevents the browser from treating all
   // radios as a single group. See https://github.com/ariakit/ariakit/issues/3833
-  const storeId = useMemo(() => store?.getState().id, [store]);
+  const storeId = useStoreState(store, "id");
   const name = nameProp ?? storeId;
 
   // When the radio store has a default value, we need to update the active id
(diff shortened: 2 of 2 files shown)


packages/ariakit-react-components/src/radio/radio.tsx
```diff
@@ -73,7 +73,7 @@ export const useRadio = createHook<TagName, RadioOptions>(function useRadio({
   // Use the store's id as the default name to ensure radios in different
   // groups have unique names. This prevents the browser from treating all
   // radios as a single group. See https://github.com/ariakit/ariakit/issues/3833
-  const storeId = useStoreState(store, (state) => state?.id);
+  const storeId = useMemo(() => store?.getState().id, [store]);
```
diegohaz: Changing idiomatic code for this little to no perf gain is not worth it. We should keep `useStoreState`, but with a small refactor:
```suggestion
  const storeId = useStoreState(store, "id");
```

.changeset/200-radio-store-id.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`Radio`](https://ariakit.com/reference/radio) rendering performance.
```
diegohaz: No need for a changeset.

diegohaz (commented):
**Initial review** — head `cb44c10d`. Files: `radio.tsx`, `.changeset/200-radio-store-id.md`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` clean; 20/20 radio example tests pass (`radio`, `radio-default-value`, `form-radio`, `menu-item-radio`).

- **The premise holds: the store `id` is provably static.** It's set once in the composite store's `initialState` (`defaultValue(props.id, syncState?.id) ?? "id-…"`, typed non-optional `string`), there is no `setState("id", …)`/`setId` anywhere, and crucially the React store wrappers never call `useStoreProps(..., "id")` (unlike `activeId`, `orientation`, etc.) — so even an `id` prop change doesn't mutate `state.id`. Reading it once via `useMemo(() => store?.getState().id, [store])` therefore can't miss an update.
- Null handling is identical (`store?.` short-circuits; `getState()` always returns a state with a non-optional `id`), and `name = nameProp ?? storeId` is unchanged. The `isChecked` subscription is untouched, so no reactivity is dropped; this just removes one redundant per-radio subscription.
- **Changeset correct:** both packages are v0 (`@ariakit/react` 0.4.29, `@ariakit/react-components` 0.1.2) so a perf improvement is `patch`; `@ariakit/react` is rightly included (re-exports `Radio`); `200-`+`patch` is a valid prefix/type combination.

.changeset/200-radio-store-id.md
diegohaz: Fixed in 6f77f31bc - removed the changeset.

packages/ariakit-react-components/src/radio/radio.tsx
diegohaz: Fixed in 6f77f31bc - kept the idiomatic `useStoreState` call and changed the store id read to `useStoreState(store, "id")`.

diegohaz:
**Fixed**

- 6f77f31bc ([thread](https://github.com/ariakit/ariakit/pull/6134#discussion_r3329670344)): Kept the idiomatic `useStoreState` call and changed the store id read to `useStoreState(store, "id")`.
- 6f77f31bc ([thread](https://github.com/ariakit/ariakit/pull/6134#discussion_r3329670343)): Removed the changeset.

diegohaz (commented):
**Re-review** (incremental) — covers the new commit `6f77f31b` ("Address Radio store id review") since my prior review of `cb44c10d`.

The net change of this PR vs `main` is now a single one-liner in `radio.tsx`:
```diff
-const storeId = useStoreState(store, (state) => state?.id);
+const storeId = useStoreState(store, "id");
```
This addresses the (resolved) self-review feedback: keep the idiomatic `useStoreState` call (the earlier `useMemo(() => store?.getState().id)` was reverted), and the `200-radio-store-id` changeset was removed.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` clean (the string-key overload types `storeId` correctly; no unused `useMemo` import); radio example tests → 20/20 pass.

- **Equivalent and correct.** The string-key path returns `state["id"]` (guarded by `hasOwnProperty`), identical to the old selector's `state?.id`; `id` is a static own property (set once at composite-store creation, no setter/`useStoreProps` sync), so neither form re-renders after mount. `name = nameProp ?? storeId` is unchanged.
- **Removing the changeset is appropriate.** The change is now a trivial idiomatic selector→string-key swap with no standalone observable behavior change; its efficiency benefit (a keyed `["id"]` subscription instead of `null`) is delivered by **#6140**, which carries its own changeset. Worth noting: #6140 isn't merged yet, so at this head both forms still subscribe with `null` — the keyed-subscription win only materializes once #6140 lands. The change is correct either way.

**Findings:** none. This is a clean improvement over the prior `useMemo` approach. Comment-only.

PR #6135, opened 2026-05-31
Optimize TooltipAnchor label subscription

diegohaz:
## Motivation

`TooltipAnchor` currently subscribes each anchor to the tooltip content element id even when the tooltip is the default description type. Description tooltips do not use that id for `aria-labelledby`, so those anchors can re-render when the rendered output is unchanged.

## Solution

Derive the generated `aria-labelledby` value in a single store selector that only reads `contentElement.id` when the tooltip type is `label`. This keeps label tooltips behaving the same while making description tooltip anchors insensitive to content id changes.

## Verification

- `pnpm lint-fix`
- `pnpm test examples/tooltip/test.ts examples/tooltip-label/test.ts`
- `pnpm tsc`
- `pnpm build`

### DIFF
--- a/.changeset/200-tooltip-anchor-labelledby.md
+++ b/.changeset/200-tooltip-anchor-labelledby.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+Improved [`TooltipAnchor`](https://ariakit.com/reference/tooltip-anchor) to avoid re-rendering default description tooltip anchors when the tooltip content id changes.
--- a/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
+++ b/packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx
@@ -150,12 +150,12 @@ export const useTooltipAnchor = createHook<TagName, TooltipAnchorOptions>(
       }
     });
 
-    const type = useStoreState(store, "type");
-    const contentId = useStoreState(store, (state) => state.contentElement?.id);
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
**Initial review** — head `59766265`. Files: `tooltip-anchor.tsx`, `.changeset/200-tooltip-anchor-labelledby.md`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` clean; `tooltip` + `tooltip-label` tests → 9/9 pass.

- **`aria-labelledby` is byte-identical** across all four input combinations of `type === "label"` × `props["aria-label"] == null` (truth table checked) — when `type !== "label"` the new `labelledBy` selector returns `undefined`, and the `aria-label` guard is applied in both versions. The `"aria-labelledby"`-before-`...props` precedence is preserved, so an explicit incoming value still overrides.
- **The re-render optimization is correct.** `useStoreState` re-renders only when the selector's return value changes; for default *description* anchors the selector now constantly returns `undefined`, so they no longer re-render when `contentElement.id` changes. Label anchors still re-render on id change, and a `type` flip still updates (the selector reads `state.type`). The two old subscriptions are correctly folded into one. `type`/`contentElement` are used nowhere else in the hook.
- **Changeset correct:** v0 → `patch`; `@ariakit/react` included (re-exports `TooltipAnchor`); proper format.

**Findings:** none. Non-blocking observation (pre-existing, not introduced here): no example exercises the `type === "label"` branch of the selector — `tooltip-label` supplies its name via `VisuallyHidden`, not `type="label"` — so that branch is covered only by construction (the truth table). Comment-only.

PR #6137, opened 2026-05-31
Hoist popup role lookup constants

diegohaz:
## Motivation

The popup role helpers in `@ariakit/utils` allocated their lookup array and object every time they ran. These helpers are public and are also used by menu, combobox, select, and dialog internals, so repeated calls should avoid unnecessary allocations.

## Solution

Hoist the allowed popup role list and popup item role map to module scope while preserving the existing `indexOf` lookup behavior. Add an `@ariakit/utils` patch changeset for the shipped repeated-call performance improvement.

### DIFF
--- a/.changeset/200-popup-role-helpers.md
+++ b/.changeset/200-popup-role-helpers.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Improved the repeated-call performance of popup role helpers in `@ariakit/utils`.
--- a/packages/ariakit-utils/src/dom.ts
+++ b/packages/ariakit-utils/src/dom.ts
@@ -255,6 +255,14 @@ export function getTextboxSelection(element: HTMLElement) {
   return { start, end };
 }
 
+const allowedPopupRoles = ["dialog", "menu", "listbox", "tree", "grid"];
+
+const itemRoleByPopupRole = {
+  menu: "menuitem",
+  listbox: "option",
+  tree: "treeitem",
+};
+
 /**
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
**Initial review** — head `67ade034`. Files: `packages/ariakit-utils/src/dom.ts`, `.changeset/200-popup-role-helpers.md`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` clean; `pnpm -F @ariakit/utils build` clean.

- **Pure, safe hoist.** `allowedPopupRoles` and `itemRoleByPopupRole` moved verbatim from inside `getPopupRole`/`getPopupItemRole` to module scope; the bodies are unchanged. Both are module-private and only read (`.indexOf`, key access) — grep confirms no assignment/push/splice anywhere — so the shared instances can't be mutated, and behavior is byte-identical. Inferred types (`string[]`, `{menu;listbox;tree}`) keep `indexOf(role: string)` valid and `keyof typeof` resolving to `"menu"|"listbox"|"tree"`. No public-API change.
- `as const`/`Object.freeze` would be a net negative here (it'd narrow `allowedPopupRoles` to a readonly literal tuple and reject the arbitrary `role: string` argument), so plain `const` is the right call.
- **Changeset correct:** `@ariakit/utils` is published (not in the changeset ignore list) and v0 (0.1.2), so a perf improvement is `patch`; dependents are correctly omitted (these helpers aren't re-exported by `@ariakit/react`'s public entry and there's no behavior change to flow downstream); `200-` + single past-tense sentence is fine.

PR #6138, opened 2026-05-31
Optimize CollectionRenderer data computation

diegohaz:
## Motivation

`CollectionRenderer` does a few small pieces of repeated work while deriving item data. `getItemId` built a fallback id string even when an item already had an explicit id, and the `getData` call was duplicated between initial state setup and later recomputation.

## Solution

Avoid building fallback ids until they are needed, and skip the object-wrapper path for primitive items. Extract the shared `getData` call into a local `computeData` helper while keeping the initializer and effect return behavior separate. Also remove an unused `itemSize` input from `getData` and add a patch changeset for `@ariakit/react-components`.

### DIFF
--- a/.changeset/200-collection-renderer-performance.md
+++ b/.changeset/200-collection-renderer-performance.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Improved virtual collection rendering performance.
--- a/packages/ariakit-react-components/src/collection/collection-renderer.tsx
+++ b/packages/ariakit-react-components/src/collection/collection-renderer.tsx
@@ -171,8 +171,11 @@ function getItemObject(item: Item): ItemObject {
 
 function getItemId(item: Item, index: number, baseId?: string) {
   invariant(baseId, "CollectionRenderer must be given an `id` prop.");
-  const defaultId = `${baseId}/${index}`;
-  return getItemObject(item).id ?? defaultId;
+  if (item && typeof item === "object") {
+    const itemObject = getItemObject(item);
+    if (itemObject.id != null) return itemObject.id;
+  }
+  return `${baseId}/${index}`;
 }
 
 function getItem<T extends Item = any>(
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
**Initial review** — head `1c34e737`. Files: `collection-renderer.tsx`, `.changeset/200-collection-renderer-performance.md`.

**Verified in an isolated worktree (Node 24):** `pnpm tsc` + oxlint clean; 7/7 collection + composite-virtual-focus vitest tests pass. (The renderer's Playwright `test-chrome` path wasn't run — the only `:3000` server belonged to a concurrent session, which I left untouched; the change is proven equivalent statically.)

- **`getItemId` is byte-equivalent.** `getItemObject` returns `{ value: item }` (no `id`) for primitives/null/undefined and the object itself otherwise; both old (`getItemObject(item).id ?? defaultId`) and new (`typeof item === "object"` guard + `id != null`) use nullish semantics, so object-with-id → id, object-without-id/primitive/null → `${baseId}/${index}`. The new form just skips the `getItemObject` allocation for primitives/null (the actual win) and defers building the fallback string.
- **`itemSize` removal is safe** — `getData`'s body (and `getAverageSize`) never read `props.itemSize`; it was dead. All outer-hook `itemSize` logic (guards, effect dep, `getItemsEnd`) is untouched.
- **`computeData` deps are correct.** Its deps `[gap, elements, horizontal, paddingStart, estimatedItemSize]` are exactly the closure values `getData` reads; the effect swapping those five for `computeData` leaves the effective trigger set unchanged (no stale closures), and hooks order is stable.
- **Changeset scope is correct.** `@ariakit/react-components`-only is right: `CollectionRenderer`/`useCollectionRenderer` (and the `CompositeRenderer`/`SelectRenderer` wrappers) are **not** re-exported by `@ariakit/react` (zero refs in `packages/ariakit-react/src/`), unlike `Radio`/`TooltipAnchor` in the sibling PRs. v0 → `patch`.

PR #6139, opened 2026-05-31
Guard nested wrapAsync polyfills

diegohaz:
## Motivation

Nested `@ariakit/test` helpers call each other through `wrapAsync`. Each nesting level was reapplying the browser polyfills for `HTMLElement.prototype.focus` and `Element.prototype.getClientRects`, stacking wrappers that captured already-patched methods as originals. The React act-environment override had the same repeated setup/restore shape.

## Solution

Track `wrapAsync` nesting depth at the module level so the browser polyfills and act-environment override are applied only when entering the outermost `wrapAsync` call and restored only after the outermost call finishes. The setup path also keeps the depth counter balanced if setup throws.

Added focused regression coverage for nested polyfill stability, restoration, and single act-environment setup/restore.

## Verification

- `pnpm lint-fix`
- `pnpm test packages/ariakit-test/src/__tests__/__utils.test.ts`
- `pnpm tsc`
- `pnpm test`
- `pnpm build`

### DIFF
--- a/.changeset/300-wrap-async-polyfills.md
+++ b/.changeset/300-wrap-async-polyfills.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/test": patch
+---
+
+Fixed nested test helpers to reuse browser polyfills and act-environment overrides instead of reapplying them.
--- a/packages/ariakit-test/src/__tests__/__utils.test.ts
+++ b/packages/ariakit-test/src/__tests__/__utils.test.ts
@@ -1,0 +1,62 @@
+// oxlint-disable unbound-method
+import { expect, test } from "vitest";
+import { isBrowser, wrapAsync } from "../__utils.ts";
+
+test("wrapAsync does not reapply browser polyfills when nested", async () => {
+  const originalFocus = HTMLElement.prototype.focus;
+  const originalGetClientRects = Element.prototype.getClientRects;
+
+  await wrapAsync(async () => {
+    const focus = HTMLElement.prototype.focus;
+    const getClientRects = Element.prototype.getClientRects;
+
+    if (!isBrowser) {
+      expect(focus).not.toBe(originalFocus);
+      expect(getClientRects).not.toBe(originalGetClientRects);
+    }
+
+    await wrapAsync(async () => {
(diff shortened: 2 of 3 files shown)


diegohaz (commented):
**Initial review** — head `96feadae`. Files: `__utils.ts`, `__tests__/__utils.test.ts` (new), `.changeset/300-wrap-async-polyfills.md`. This is a genuine **bug fix** for `@ariakit/test`.

**Verified in an isolated worktree (Node 24):** new `__utils.test.ts` → 2/2 pass; `pnpm tsc` clean; sample integration tests (`dialog`/`menu`/`combobox` 43/43, plus `select`/`tab`/`form` 28/28) pass — no regression in the pervasive non-nested usage.

- **Depth-guard is correct** across every traced path: single (0→1 setup once, 1→0 restore once — byte-identical to the old non-nested behavior), nested (setup once at the outermost, restore once after it), `fn` rejection (the `finally` still balances), and both setup-throw paths. Crucially, `setupWrapAsyncEnvironment()` is called **outside** `wrapAsync`'s `try`, so a setup throw doesn't double-decrement the counter; the `catch` decrements depth and calls `restoreActEnvironment()` so the act env is never left mutated.
- **Imports/exports check out**: `isBrowser` is already exported (line 8), `noop` is imported from `@ariakit/utils`, `setActEnvironment`/`applyBrowserPolyfills` are in scope.
- **Concurrency assumption holds** — the module-global depth counter assumes nesting, not interleaved top-level calls; all 14 call sites await sequentially, and the old code (stacking global prototype patches) was never concurrency-safe either, so nothing regresses.
- **Tests are meaningful**: the polyfill test proves non-re-application + restoration to true originals (gated behind `!isBrowser`); the act-env test records every `IS_REACT_ACT_ENVIRONMENT` assignment and asserts `[false, true]` (single setup + single restore), restoring the original descriptor in `finally`.
- **Changeset correct**: `300-`/`patch` for a fix; `@ariakit/test` is published (v0.6.1); single past-tense "Fixed …" sentence.

**Findings:** none. Optional (non-blocking): a regression test for the setup-throws error-balance path would harden it but isn't required. Comment-only.

PR #6140, opened 2026-05-31
Optimize useStoreState key subscriptions

diegohaz:
## Motivation

`@ariakit/store` has a keyed listener dispatch path that can avoid scanning unrelated listeners when updates touch a single state key. `useStoreState(store, "key")` was still subscribing as an all-keys listener, so normal keyed React subscriptions prevented that fast path from being used.

## Solution

`useStoreState` now subscribes with `[key]` when the second argument is a string key. Whole-state reads, selector reads, and `useStoreStateObject` continue to subscribe with `null` because they can observe arbitrary state changes.

The new React-store tests assert the subscription keys for each mode and cover the keyed render behavior. The React-store package now declares the test-only dependencies it imports, and the test filename follows the package ignore pattern so it is not included in npm packs.

### DIFF
--- a/.changeset/200-use-store-state-key-subscriptions.md
+++ b/.changeset/200-use-store-state-key-subscriptions.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`useStoreState`](https://ariakit.com/reference/use-store-state) performance when subscribing to individual store keys.
--- a/packages/ariakit-react-store/package.json
+++ b/packages/ariakit-react-store/package.json
@@ -34,8 +34,10 @@
   },
   "devDependencies": {
     "@ariakit/scripts": "workspace:*",
+    "@ariakit/test": "workspace:*",
     "@types/use-sync-external-store": "1.5.0",
-    "react": "19.2.6"
+    "react": "19.2.6",
+    "vitest": "4.1.7"
   },
   "peerDependencies": {
     "react": "^17.0.0 || ^18.0.0 || ^19.0.0"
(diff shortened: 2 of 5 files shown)


diegohaz (commented):
**Initial review** — head `1c78f35f`. Published `@ariakit/react-store`. Files: `index.tsx`, new `use-store-state.test.react.tsx`, `package.json`, `pnpm-lock.yaml`, changeset.

**Verified in an isolated worktree (Node 24):** new tests 5/5; `pnpm tsc` clean; a broad sample of component example tests (`menu`/`combobox`/`select`/`tab`/`dialog`/`checkbox`/`radio`) → **545 tests / 84 files pass, no regression**; `pnpm install` leaves `pnpm-lock.yaml` with no extra drift.

- **Keyed-subscription completeness is proven from the store source** (the make-or-break check). In `@ariakit/store`, `state` is reassigned in exactly one place — inside `setState(key, value)` — and it then calls `runListeners(syncListenerGroup, prevState, key)` with that exact key; `runListeners` always notifies the `[key]`-keyed listeners (both with and without all-keys listeners present), and cross-store sync routes through the same `setState(key,…)` path. So `subscribe(store, [key], cb)` fires on **exactly** the set of changes that the old `null` listener reacted to for that key — no missed re-render. `useStoreState` uses the synchronous listener group, so batching doesn't affect this.
- **Render behavior is identical** (re-render iff `state[key]` changes); the `[key]` path just avoids waking the listener for unrelated keys. The `!key`→`key == null` change only matters for an empty-string key, which doesn't occur and still resolves to `undefined`. `[store, key]` deps correctly re-subscribe on key change; whole-state reads use the `identity` default (selector path, `null` subscription).
- **Packaging/deps correct**: `@ariakit/test` (`workspace:*`) + `vitest` (`4.1.7`, matching the repo) added because the new test imports them; `npm pack --dry-run` confirms no `.test.` file ships. Changeset includes all three packages in the re-export chain (`react-store`→`react-components`→`react`), all v0 → `patch`.

**Findings:** none blocking. Optional (non-blocking): no unit test covers re-subscription when the `key` prop *changes between renders* (the `[store, key]` dep) — it's exercised indirectly by the 545 example tests, but a focused assertion (`key` flips `"count"`→`"label"`, expect `subscribe` last called with `["label"]`) would harden it. Comment-only.

PR #6141, opened 2026-05-31
Add tab panel lookup

diegohaz:
## Motivation

Each `Tab` currently resolves its associated panel by subscribing to the full panels collection and scanning `items` for a matching `tabId`. When panel items change, that makes each tab repeat the same O(N) lookup work.

## Solution

Add a tab-specific `panel(tabId)` lookup to the tab store and keep it synchronized with the panels collection. The lookup preserves the previous first-match-in-items-order behavior, updates through panel registration/rendering and direct panel item state changes, and keeps orphan panel auto-assignment in sync.

`Tab` now reads the associated panel id through this lookup instead of scanning the panels array.

## Verification

- `pnpm lint-fix`
- `pnpm test packages/ariakit-components/src/tab/__tests__/tab-store.test.ts`
- `pnpm tsc`
- `pnpm build`
- `pnpm test`

### DIFF
--- a/packages/ariakit-components/src/tab/tab-store.ts
+++ b/packages/ariakit-components/src/tab/tab-store.ts
@@ -45,6 +45,102 @@ function isEnabledTab(
   return true;
 }
 
+function createPanelsStore() {
+  const panels = createCollectionStore<TabStorePanel>();
+  const panelsByTabId = new Map<string, TabStorePanel>();
+  let panelItems = panels.getState().items;
+
+  const syncPanelItems = (items: TabStorePanel[]) => {
+    panelItems = items;
+    panelsByTabId.clear();
+    for (const panel of panelItems) {
+      const { tabId } = panel;
+      if (tabId == null) continue;
+      if (panelsByTabId.has(tabId)) continue;
+      panelsByTabId.set(tabId, panel);
+    }
+  };
+
+  const syncPanelByTabId = (tabId: string | null | undefined) => {
+    if (tabId == null) return;
+    const panel = panelItems.find((item) => item.tabId === tabId);
+    if (panel) {
+      panelsByTabId.set(tabId, panel);
+    } else {
+      panelsByTabId.delete(tabId);
+    }
+  };
+
+  const syncPanelLookup = (
+    previousPanel: TabStorePanel | undefined,
+    nextPanel: TabStorePanel | undefined,
+  ) => {
+    syncPanelByTabId(previousPanel?.tabId);
+    if (nextPanel?.tabId !== previousPanel?.tabId) {
+      syncPanelByTabId(nextPanel?.tabId);
+    }
+  };
+
+  sync(panels, ["items"], (state) => {
+    syncPanelItems(state.items);
+  });
+
+  const mergePanelItem = (item: TabStorePanel) => {
+    // Mirrors collection-store's merge behavior so panel() preserves the
+    // synchronous getter semantics of panels.item().
+    let previousPanel: TabStorePanel | undefined;
+    const index = panelItems.findIndex(({ id }) => id === item.id);
+    if (index !== -1) {
+      previousPanel = panelItems[index];
+      const nextPanel = { ...previousPanel, ...item };
+      const nextItems = panelItems.slice();
+      nextItems[index] = nextPanel;
+      panelItems = nextItems;
+      syncPanelLookup(previousPanel, nextPanel);
+    } else {
+      panelItems = [...panelItems, item];
+      syncPanelLookup(undefined, item);
+    }
+    return () => {
+      const currentPanel = panelItems.find(({ id }) => id === item.id);
+      if (!previousPanel) {
+        panelItems = panelItems.filter(({ id }) => id !== item.id);
+        syncPanelLookup(currentPanel, undefined);
+        return;
+      }
+      const index = panelItems.findIndex(({ id }) => id === item.id);
+      if (index === -1) return;
+      const nextItems = panelItems.slice();
+      nextItems[index] = previousPanel;
+      panelItems = nextItems;
+      syncPanelLookup(currentPanel, previousPanel);
+    };
+  };
+
+  const registerItem = panels.registerItem;
+  const renderItem = panels.renderItem;
+
+  return {
+    panels: {
+      ...panels,
+      registerItem: (item) => {
+        const unregisterItem = registerItem(item);
+        const unmergePanelItem = mergePanelItem(item);
+        return chain(unregisterItem, unmergePanelItem);
+      },
+      renderItem: (item) => {
+        const unrenderItem = renderItem(item);
+        const unmergePanelItem = mergePanelItem(item);
+        return chain(unrenderItem, unmergePanelItem);
+      },
+    } satisfies CollectionStore<TabStorePanel>,
+    panel: (tabId) => {
+      if (tabId == null) return null;
+      return panelsByTabId.get(tabId) || null;
+    },
+  } satisfies Pick<TabStoreFunctions, "panels" | "panel">;
+}
+
 export function createTabStore({
   composite: parentComposite,
   combobox,
@@ -90,7 +186,7 @@ export function createTabStore({
     focusLoop: defaultValue(props.focusLoop, syncState?.focusLoop, true),
   });
 
-  const panels = createCollectionStore<TabStorePanel>();
+  const { panels, panel } = createPanelsStore();
 
   const initialState: TabStoreState = {
     ...composite.getState(),
@@ -223,6 +319,7 @@ export function createTabStore({
     ...composite,
     ...tab,
     panels,
+    panel,
     setSelectedId: (id) => tab.setState("selectedId", id),
     select: (id) => {
       tab.setState("selectedId", id);
@@ -297,6 +394,12 @@ export interface TabStoreFunctions extends CompositeStoreFunctions<TabStoreItem>
    * - [Animated TabPanel](https://ariakit.com/examples/tab-panel-animated)
    */
   panels: CollectionStore<TabStorePanel>;
+  /**
+   * Gets the panel associated with the given tab id.
+   * @example
+   * const panel = store.panel("tab-1");
+   */
+  panel: (tabId: string | null | undefined) => TabStorePanel | null;
   /**
    * Selects the tab for the given id and moves focus to it. If you want to set
    * the [`selectedId`](https://ariakit.com/reference/tab-provider#selectedid)
--- a/packages/ariakit-components/src/tab/tab-store.ts
+++ b/packages/ariakit-components/src/tab/tab-store.ts
@@ -48,92 +48,19 @@ function isEnabledTab(
 function createPanelsStore() {
   const panels = createCollectionStore<TabStorePanel>();
   const panelsByTabId = new Map<string, TabStorePanel>();
-  let panelItems = panels.getState().items;
 
-  const syncPanelItems = (items: TabStorePanel[]) => {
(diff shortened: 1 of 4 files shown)


diegohaz (commented):
**Initial review** — head `e65d3b62`. Files: `tab-store.ts` (`@ariakit/components`), new `tab-store.test.ts`, `tab.tsx` (`@ariakit/react-components`), changeset.

**Verified in an isolated worktree (Node 24):** new store tests 4/4; 10 tab-related `@ariakit/test` example suites → **105 tests pass, no regression**; `pnpm tsc` clean.

- **First-match-in-items-order is preserved across every path** (full rebuild, register/render + merge, tabId change, unmerge cleanup, direct `setState("items")`, null tabId) — empirically traced; `store.panel(id)?.id` matches the old `items.find(i => i.tabId === id)?.id` at each step. This turns a per-tab O(n) scan into an O(1) map lookup.
- **The dual-path (sync subscription + `mergePanelItem`) is consistent.** The collection store propagates items from its private to public store via a microtask `batch`, so `panels.getState().items` is empty *synchronously* right after `registerItem`; `mergePanelItem` supplies the synchronous result and the `sync(["items"])` rebuild reconciles on the next microtask — they converge to the same first-match. The only divergence I could construct needs a non-real-world mix of manual `registerItem` + wholesale `setState("items")` and self-heals on the next rebuild.
- **The `mergePanelItem` mirror is load-bearing, not avoidable complexity.** A simpler "rebuild from `panels.items` on change" index can't work: public items is `[]` synchronously after register, and both the tests and the public `panel()` getter contract require synchronous semantics (like `panels.item()`). So the ~40-line parallel-merge is the minimal way to satisfy that contract.
- `tab.tsx` uses a function selector (`() => store.panel(id)?.id`) → `null` subscription, unaffected by #6140; `store.panel` is stable. Public `panel()` addition is documented via JSDoc. Changeset spans `@ariakit/components`+`react-components`+`react` (all re-export `Tab`/`useTabStore`), all v0 → `patch`, `200-` feature ordering, with reference links.

packages/ariakit-components/src/tab/tab-store.ts
```diff
@@ -45,6 +45,102 @@ function isEnabledTab(
...
+        const unmergePanelItem = mergePanelItem(item);
+        return chain(unrenderItem, unmergePanelItem);
+      },
+    } satisfies CollectionStore<TabStorePanel>,
+    panel: (tabId) => {
+      if (tabId == null) return null;
+      return panelsByTabId.get(tabId) || null;
+    },
+  } satisfies Pick<TabStoreFunctions, "panels" | "panel">;
+}
```
diegohaz, line 69: **Avoidable complexity: the `mergePanelItem` mirror replicates a synchronous window that no consumer observes.**

The goal — turn the per-tab O(n) `items.find(...)` scan into an O(1) read — only needs a `tabId → panel` map kept current with the panels collection. The ~55-line parallel-merge engine (`mergePanelItem` + `panelItems` + `syncPanelByTabId`/`syncPanelLookup`) exists solely to make `panel()` reflect a `registerItem`/`renderItem` *synchronously, before the collection's microtask batch flushes* — i.e. to match `panels.item()`'s getter semantics. But nothing actually observes that window:

- The only consumer is `tab.tsx`: `useStoreState(store.panels, () => store.panel(id)?.id)`. `useStoreState` subscribes to **all** panels keys (`subscribe(store, null, …)`) and recomputes the snapshot on every notification. The `items` notification arrives on the same microtask as the public `items` update, and the `sync(panels, ["items"])` listener (registered at store creation) runs *before* the React subscriber — so the map is already current when `getSnapshot` reads it.
- The previous code read `state.items` (the **public** items), which itself lags registration by a microtask, so the old behavior was never synchronous-after-register either. Only the new unit tests assert the synchronous window, and they assert a contract this PR introduced rather than one a consumer needs.

So the map can be maintained by the existing `sync` listener alone, dropping the entire mirror:

```ts
function createPanelsStore() {
  const panels = createCollectionStore<TabStorePanel>();
  const panelsByTabId = new Map<string, TabStorePanel>();

  sync(panels, ["items"], (state) => {
    panelsByTabId.clear();
    for (const panel of state.items) {
      const { tabId } = panel;
      if (tabId == null) continue;
      if (panelsByTabId.has(tabId)) continue;
      panelsByTabId.set(tabId, panel);
    }
  });

  return {
    panels,
    panel: (tabId) => {
      if (tabId == null) return null;
      return panelsByTabId.get(tabId) || null;
    },
  } satisfies Pick<TabStoreFunctions, "panels" | "panel">;
}
```

I applied exactly this locally and ran the tab integration suites — `tab`, `tab-disabled`, `tab-nested`, `tab-react-router`, `combobox-tabs`, `select-combobox-tab`, `dialog-combobox-tab-command-menu`, `tab-panel-animated` — **91/91 pass**, identical to the PR head. (The 4 new store unit tests would need their immediate-after-`registerItem` reads to await a microtask, since they assert the synchronous window this version intentionally drops — `chain` would then be unused in this file.)

It's also less work: the mirror does an O(n) `findIndex` + `slice` + `find` on **every** `registerItem`/`renderItem` call, whereas the single `sync` listener rebuilds the map once per batched `items` change.

The one thing the simpler version gives up is `panel()` being synchronous in parity with `panels.item()`. If that parity is a deliberate public-API guarantee worth keeping, it's worth stating *why* given no current consumer relies on it — and even then it doesn't require hand-rolling a second merge engine. As written, the mirror is more machinery than the problem requires.

diegohaz (commented):
**Thorough full review** (not an incremental re-review) of head `e65d3b62`, diffed against the merge base `origin/main` (`2445b0742`). Files: `tab-store.ts` (`@ariakit/components`), new `tab-store.test.ts`, `tab.tsx` (`@ariakit/react-components`), changeset.

### What I verified
- **Correctness of the lookup, all paths:** traced `store.panel(id)?.id` against the old `items.find(i => i.tabId === id)?.id` across full rebuild, register/render + merge, tabId change, unmerge restore, and direct `setState("items")` — first-match-in-items-order is preserved, and the dual-path (synchronous `mergePanelItem` + microtask `sync(["items"])` rebuild) converges. No correctness or regression issue found in the PR as written.
- **Reactivity:** `useStoreState(store.panels, () => store.panel(id)?.id)` subscribes to all panels keys and recomputes per notification — same re-render frequency as the old `state.items.find(...)` selector; the function-style selector is consistent with existing `store.item(...)` / `store.panels.item(...)` usage in `tab.tsx`/`tab-panel.tsx`.
- **Changeset:** `200-` (minor/feature ordering) + `patch` frontmatter is correct for the three v0 packages; the reference links resolve and `panel()` will surface in the generated `useTabStore` docs.
- **Tests run (Node 24):** new store tests 4/4; `tab`, `tab-disabled`, `tab-nested`, `tab-react-router`, `combobox-tabs`, `select-combobox-tab`, `dialog-combobox-tab-command-menu`, `tab-panel-animated` → all pass on PR head.

### Finding (1, inline)
- **Avoidable complexity** in `createPanelsStore` — the `mergePanelItem` parallel-merge mirror exists only to give `panel()` a synchronous-after-register window that no consumer observes (the sole consumer reads via `useStoreState`, driven by the microtask `items` notification, exactly as the old code did). I validated a ~12-line replacement that maintains the `tabId → panel` map from the existing `sync(panels, ["items"])` listener alone: **91/91 tab integration tests pass** unchanged. Details, the drop-in code, and the synchronous-parity trade-off are on the inline thread.

### Not separately re-validated
- `pnpm tsc` / `pnpm build` on PR head (relied on author's run + CI). The flagged perf-bot regressions are in the unrelated `ariakit-tailwind` style-recalc sandbox, not tabs.

Comment-only; no approval or change request.

packages/ariakit-components/src/tab/tab-store.ts
diegohaz: Fixed in 7f13994f1 - removed the eager `mergePanelItem` mirror and now keep the tabId-to-panel lookup synchronized only from `sync(panels, ["items"])`. The new store tests now await the collection batch flush for `registerItem`/`renderItem` paths, matching the public `items` timing observed by `useTab`.

diegohaz:
**Fixed**
- 7f13994f1 ([reply](https://github.com/ariakit/ariakit/pull/6141#discussion_r3329715762)): Removed the eager `mergePanelItem` mirror and now rebuild the tabId-to-panel lookup from the public panel `items` sync listener. Updated the store tests to await the collection batch flush for `registerItem`/`renderItem` paths.

**Verification**
- `pnpm lint-fix`
- `pnpm test packages/ariakit-components/src/tab/__tests__/tab-store.test.ts`
- `pnpm test packages/ariakit-components/src/tab/__tests__/tab-store.test.ts examples/tab/test.ts examples/tab-disabled/test.ts examples/tab-nested/test.ts examples/tab-react-router/test.ts examples/combobox-tabs/test.ts examples/select-combobox-tab/test.ts examples/dialog-combobox-tab-command-menu/test.ts examples/tab-panel-animated/test.ts`
- `pnpm tsc`
- Pre-commit native and Cursor reviews: no findings

packages/ariakit-components/src/tab/tab-store.ts
diegohaz: Verified fixed in `7f13994f1`. `createPanelsStore` now maintains `panelsByTabId` purely from the `sync(panels, ["items"])` listener and returns the unwrapped `panels` — the `mergePanelItem`/`panelItems`/`syncPanelLookup` machinery is gone (≈79 lines removed). First-match-in-items-order is preserved by the `panelsByTabId.has(tabId)` guard, and `chain` is still a live import (used by the select/combobox restore logic), so no dead code.

The test update is the right call too: the register/render cases now `init(store.panels)` (so the collection's private→public batch actually propagates) and `await` a microtask before asserting, which mirrors the public-`items` timing `useTab` observes; the direct-`setState("items")` case stays synchronous since that fires the `sync` listener immediately.

Re-checked on the new head (Node 24): `tab-store` unit tests 4/4, the `tab`/`tab-disabled`/`tab-nested`/`tab-react-router`/`combobox-tabs`/`select-combobox-tab`/`dialog-combobox-tab-command-menu` suites all pass (95 total), and `pnpm tsc` is clean. Resolved.

diegohaz (commented):
**Re-review** — incremental baseline `e65d3b62` (my prior marked review) → new head `7f13994f1` ("Simplify tab panel lookup"). Only `tab-store.ts` and the store test changed in this commit; `tab.tsx` and the changeset are untouched.

### Prior finding — resolved (verified, not just taken from the reply)
The `mergePanelItem` mirror is gone. `createPanelsStore` now keeps `panelsByTabId` solely from `sync(panels, ["items"])` and returns the unwrapped `panels` (≈79 lines removed). I confirmed in the current code: first-match-in-items-order preserved via the `has(tabId)` guard, `chain` still used by the select/combobox restore logic (no dead import), `panel()`/`panels` still wired into the store and the public `TabStoreFunctions` type intact. The tests correctly switch to `init(store.panels)` + microtask flush for the register/render paths (matching the public-`items` timing `useTab` sees) and keep the direct-`setState` path synchronous. Details on the resolved thread.

### Verification on new head (Node 24)
- `tab-store` unit tests 4/4.
- `tab`, `tab-disabled`, `tab-nested`, `tab-react-router`, `combobox-tabs`, `select-combobox-tab`, `dialog-combobox-tab-command-menu` → 95 tests pass.
- `pnpm tsc` clean.

### New findings
None. The simplification is correct, complete, and behavior-preserving for every real consumer.

Comment-only; no approval or change request.

PR #6142, opened 2026-05-31
Prevent inline combobox blur listener churn

diegohaz:
## Motivation

Inline Combobox autocomplete updates the computed input value while the user types. The native `focusout` listener that commits the inline-completed value on blur depended on that computed value, so React cleaned up and reattached the listener on every value update.

## Solution

Move the blur commit logic into a stable `useEvent` handler. The effect now subscribes based on the inline mode and current content element, while the handler reads the latest Combobox value, store, input element, and content element when `focusout` fires.

Add a regression test that tracks the Combobox `focusout` listener and verifies an inline value update does not remove it. Add a changeset for `@ariakit/react-components` and the `@ariakit/react` re-export.

## Verification

- `pnpm lint-fix`
- `pnpm test packages/ariakit-react-components/src/combobox/__tests__/combobox.test.react.tsx app/src/examples/combobox-group/test.ts`
- `pnpm tsc`
- `pnpm build`

### DIFF
--- a/packages/ariakit-react-components/src/combobox/__tests__/combobox.test.react.tsx
+++ b/packages/ariakit-react-components/src/combobox/__tests__/combobox.test.react.tsx
@@ -1,0 +1,100 @@
+import { q, render, waitFor } from "@ariakit/test/react";
+import { expect, test } from "vitest";
+import { ComboboxItem } from "../combobox-item.tsx";
+import { ComboboxList } from "../combobox-list.tsx";
+import { ComboboxProvider } from "../combobox-provider.tsx";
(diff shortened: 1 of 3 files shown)


packages/ariakit-react-components/src/combobox/__tests__/combobox.test.react.tsx
```diff
@@ -0,0 +1,100 @@
+import { q, render, waitFor } from "@ariakit/test/react";
```
diegohaz, line 1: Component tests must be in app sandbox

PR #6143, opened 2026-05-31
Avoid reinstalling hovercard mousemove listeners

diegohaz:
## Motivation

Nested hovercards currently update parent state when descendants mount or unmount. Because the hover-intent document mousemove effect reads that state directly and lists it as a dependency, each nested hovercard change makes ancestor hovercards remove and reinstall the capture listener during navigation.

## Solution

Keep the nested hovercard array in a ref and read that ref from the mousemove handler, so nested register/unregister changes do not invalidate the listener effect or re-render the parent hovercard. Add an explicit `clearHideTimeout` helper and call it from cleanup plus nested register/unregister to preserve timeout cancellation while also normalizing the stale timeout id back to `0`.

This also adds an app sandbox regression fixture that toggles a nested hovercard and asserts the existing mousemove listener is not removed, plus a changeset for `@ariakit/react-components` and `@ariakit/react`.

## Verification

- `pnpm test app/src/sandbox/hovercard-nested-listener/test.ts`
- `pnpm lint-fix`
- `pnpm test examples/hovercard/test.ts examples/hovercard-disclosure/test.ts app/src/sandbox/hovercard-nested-listener/test.ts packages/ariakit-react-components/src/hovercard/utils/__tests__/polygon.test.ts`
- `pnpm tsc`
- `pnpm build-app-lite`


packages/ariakit-react-components/src/hovercard/hovercard.tsx
```diff
@@ -150,6 +150,7 @@ export const useHovercard = createHook<TagName, HovercardOptions>(
 
     const ref = useRef<HTMLType>(null);
     const [nestedHovercards, setNestedHovercards] = useState<HTMLElement[]>([]);
+    const nestedHovercardsRef = useLiveRef(nestedHovercards);
```
diegohaz: After this change, `nestedHovercards` state is only ever read to feed `useLiveRef` — the value itself is no longer consumed during render (the mousemove handler reads `nestedHovercardsRef.current`, and nothing else references `nestedHovercards`). That makes the `useState` + `useLiveRef` pair pure indirection: every nested register/unregister still calls `setNestedHovercards`, which re-renders this hovercard solely to let `useLiveRef`'s layout effect copy the new array into the ref.

Since the registration callbacks already run synchronously, a plain `useRef` mutated in place keeps the exact same "read the latest nested array from a ref" semantics in the mousemove handler while dropping both the indirection and the per-mount/unmount parent re-render. That re-render is the same class of avoidable work this PR set out to remove (work triggered by nested hovercards mounting/unmounting), so eliminating it is in the spirit of the change rather than a separate concern.

```suggestion
    const nestedHovercardsRef = useRef<HTMLElement[]>([]);
```

And mutate the ref directly in `registerNestedHovercard`:

```diff
     const registerNestedHovercard = useCallback(
       (element: HTMLElement) => {
         clearHideTimeout();
-        setNestedHovercards((prevElements) => [...prevElements, element]);
+        nestedHovercardsRef.current = [...nestedHovercardsRef.current, element];
         const parentUnregister = registerOnParent?.(element);
         return () => {
           clearHideTimeout();
-          setNestedHovercards((prevElements) =>
-            prevElements.filter((item) => item !== element),
-          );
+          nestedHovercardsRef.current = nestedHovercardsRef.current.filter(
+            (item) => item !== element,
+          );
           parentUnregister?.();
         };
       },
       [clearHideTimeout, registerOnParent],
     );
```

The effect's dependency array can keep `nestedHovercardsRef` (stable), so no other change is needed. `useState`/`useLiveRef` imports stay in use elsewhere in the file.

I validated this locally on top of this branch: the new `hovercard-nested-listener` test plus `examples/hovercard`, `examples/hovercard-disclosure`, `examples/menu`, `examples/submenu`, `examples/menubar`, `examples/tooltip`, `examples/menu-nested`, and `examples/select` (303 tests) all pass, and `pnpm tsc` is clean. It's a non-blocking simplification — feel free to keep the state-based version if you prefer a more conservative diff.

diegohaz (commented):
## Review summary

**Scope:** Initial review of the full PR. Baseline diff `3b0ecdb9` (merge base with `main`) → head `9b64118d`. Reviewed the `useHovercard` change, the new `app/src/sandbox/hovercard-nested-listener` fixture/test, the changeset, and the surrounding hovercard/menu/submenu/tooltip call sites and helpers (`useLiveRef`, `addGlobalEventListener`, `useBooleanEvent`, the app sandbox test loader).

**Verdict:** The change is correct and well-motivated. Reading the live nested-hovercard ref from the mousemove handler instead of listing `nestedHovercards` as an effect dependency removes the listener reinstall on every nested register/unregister, and the new `clearHideTimeout` helper faithfully preserves the timeout-cancellation that the old effect re-run used to provide. The explicit reset of `hideTimeoutRef.current` to `0` is actually a small correctness improvement: the previous cleanup called `clearTimeout` but left a stale non-zero id, which could make the `if (hideTimeoutRef.current) return;` guard skip rescheduling a hide. One non-blocking simplification posted inline.

**What I validated locally (on this branch):**
- `app/src/sandbox/hovercard-nested-listener/test.ts` passes on head; reverting only `hovercard.tsx` to the base version makes it fail (`removedExistingMouseMoveListeners` length 1), so the test is a genuine regression guard, not a tautology.
- The new test renders in `StrictMode` via `vitest.setup.ts` and `failOnConsole()` is active — the fixture runs clean under both.
- The inline simplification (plain `useRef` instead of `useState` + `useLiveRef`) was prototyped and run against 303 tests (`hovercard`, `hovercard-disclosure`, `menu`, `submenu`, `menubar`, `tooltip`, `menu-nested`, `select`, and the new sandbox) plus `pnpm tsc` — all green. The experiment was reverted; the working tree is left clean.

**Findings:**
- **Simplification (non-blocking):** `nestedHovercards` state is now consumed only by `useLiveRef`; a plain `useRef` removes the indirection and the parent re-render that still fires on every nested mount/unmount. Details and a tested suggestion are in the inline thread.

**Not flagged (checked, no action needed):**
- Changeset packages (`@ariakit/react-components` + `@ariakit/react`, both `patch`) match the convention used by other recent changesets, and a perf/behavior improvement warrants a changeset.
- Sandbox naming/structure (`preview.astro`, `preview.mdx`, `index.react.tsx`, `test.ts`) follows the established `app/src/sandbox` conventions; jsdom is the right environment for a listener-reinstall regression, so no browser test is needed.
- Effect dependency array correctly swaps `nestedHovercards` for the stable `nestedHovercardsRef`/`clearHideTimeout`, so the listener effect no longer re-runs on nested changes.

This is a comment-only review; I'm not requesting changes.

packages/ariakit-react-components/src/hovercard/hovercard.tsx
diegohaz: Fixed in 5b067eeec - replaced the `useState` + `useLiveRef` indirection with a plain `useRef<HTMLElement[]>([])` that is updated synchronously during nested hovercard register/unregister, so nested changes no longer re-render the parent hovercard while the mousemove handler still reads the latest list.

diegohaz:
**Fixed**

- 5b067eeec ([reply](https://github.com/ariakit/ariakit/pull/6143#discussion_r3329761219)): Replaced the `useState` + `useLiveRef` nested hovercard tracking with a plain ref updated synchronously in register/unregister, avoiding the remaining parent re-render while preserving the mousemove handler behavior.

diegohaz (commented):
## Re-review summary

**Scope:** Re-review of the commits since my last review. Incremental baseline `9b64118d` (last reviewed head) → current head `5b067eec`. Only one new commit: `5b067ee` "Simplify nested hovercard tracking".

**Verdict:** The new commit applies the simplification from the prior review thread — `useState` + `useLiveRef` for `nestedHovercards` is replaced with a plain `useRef<HTMLElement[]>([])` mutated synchronously in `registerNestedHovercard`/unregister. I verified the fix is complete and correct rather than relying on the resolved thread:

- `useState` (still used by `useAutoFocusOnHide`) and `useLiveRef` (still used by `openRef`) remain in use, so no dead imports were left behind.
- The mousemove effect keeps the stable `nestedHovercardsRef` in its dependency array, so it no longer re-runs on nested register/unregister — the original goal of the PR is preserved.
- `pnpm tsc` clean (0 errors), `oxlint` clean on the changed file, and 134 tests across `hovercard-nested-listener`, `examples/hovercard`, `examples/submenu`, and `examples/menu` pass on this exact head. (In my prior run the same change passed a wider 303-test set including tooltip/menubar/select/menu-nested.)

**Findings:** None. The single prior finding is resolved and independently verified; the new commit is isolated to the two flagged locations and introduces no new issues. No changeset update is needed — the change remains an internal refactor of the same behavior already described.

Comment-only review; not requesting changes.

PR #6144, opened 2026-05-31
Colocate package tests

diegohaz:
## Motivation

Package tests should live next to the source files they cover instead of under separate `__tests__` folders. Once those files are colocated, the package export generator must not expose `.test` modules in `package.json` exports.

## Solution

Move package tests out of `__tests__` directories and update their relative imports. Inline the local `useAllEvents` test helper into the only test that used it.

Update the Ariakit build script to use the generated `.npmignore` entries as the source of truth for filtering public source files before generating package exports. Add a regression test covering ignored test files, `__tests__` paths, config files, and `src/test.ts`.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test`
- `pnpm build`
- `rg '"\\./.*(?:\\.test|__tests__)' packages/*/package.json || true`
- `find packages -type d -name __tests__ | sort`
- `pnpm clean`
- `pnpm lint`


packages/ariakit-test/src/blur.test.react.tsx
```diff
@@ -103,10 +107,7 @@ function getIdentifier(element: HTMLElement) {
   );
 }
 
-export function useAllEvents(
-  ref: React.RefObject<Element | null>,
-  stack: string[],
-) {
+function useAllEvents(ref: RefObject<Element | null>, stack: string[]) {
```
diegohaz: We should probably move this to `__use-all-events.ts` since we may want to use it in other tests in the future?

diegohaz: Extracting to `__use-all-events.ts` works, and the `__` prefix is load-bearing here. `getPublicFiles` skips `__`-prefixed entries via `isPrivateModule`, so the helper stays out of `@ariakit/test`'s generated `exports` — same mechanism that keeps the existing `__utils.ts`/`__aria-role.ts` private.

A plain `use-all-events.ts` would instead be picked up as a public `./use-all-events` export under the new filter (it's a non-test `.ts` source). I confirmed this locally — regenerating exports with a plain `use-all-events.ts` adds `"./use-all-events": "./src/use-all-events.ts"`, while `__use-all-events.ts` does not — so the `__` is what keeps it private.

So: if you expect reuse, `__use-all-events.ts` is the right move and won't leak. If there's no second consumer yet, the current inlining is also fine (YAGNI) — either way, just avoid a non-prefixed name.

packages/ariakit-scripts/src/build.ts
```diff
@@ -54,6 +65,58 @@ function isPrivateModule(filename: string) {
...
+  if (!entry.includes("/")) {
+    return parts.some((part) => matchesBasenamePattern(part, entry));
+  }
+
+  return matchesPathPattern(parts, entry.split("/"));
+}
+
+function isNpmIgnored(path: string) {
+  return npmIgnoreEntries.some((entry) => matchesNpmIgnoreEntry(path, entry));
+}
```
diegohaz, line 120: **Non-blocking — the export filter mirrors only a subset of `.npmignore` semantics**

Deriving the filter from `npmIgnoreEntries` so `.npmignore` and the generated `exports` can't drift is a good call, and it holds today: I regenerated source exports at this head for every generator-driven package (`@ariakit/components`, `@ariakit/react-components`, `@ariakit/test`, …) and got `exports` byte-identical to `main`, and `build.test.ts` passes against the real `updateSourcePackageJson`.

The caveat: this matcher implements only `*`, `**`, and bare-name "match any segment". It silently no-ops on the rest of gitignore/`.npmignore` syntax — negation (`!keep.test.ts`), root-anchored (`/dist`), dir-only trailing slash (`build/`), `?`/`[…]`. Every current entry stays within the supported subset, so output is correct; but the "source of truth" coupling is only partial — a future entry using any unsupported construct would silently diverge from what npm actually ships (a dangling export pointing at a file npm dropped, or a missing export for a file npm keeps), with no test catching it.

Two options:

1. Use `ignore` (already transitively in the lockfile) for exact gitignore parity. As a direct dep of the private `@ariakit/scripts` it collapses this whole block:
   ```ts
   import ignore from "ignore";
   const npmIgnore = ignore().add(npmIgnoreEntries);
   const isNpmIgnored = (path: string) => npmIgnore.ignores(path);
   ```
2. If you'd rather not add a dependency, keep the matcher but add a brief comment noting it supports only `*`/`**`/bare-name patterns, so future `npmIgnoreEntries` edits stay within that subset.

diegohaz (commented):
### Review summary — initial review

**Scope:** Full review of `b74a73a` against merge-base `405e01e7` (`main`), then expanded to the surrounding build/test infrastructure. Reviewed in a dedicated detached worktree at the PR head (the branch is checked out elsewhere).

**Verdict:** The change is correct and well-targeted. Colocating tests removes the implicit `__tests__`/`isPrivateModule` exclusion, and the new `.npmignore`-derived filter restores it. One non-blocking maintainability note inline; nothing blocking.

**What I verified locally:**
- **No export regressions.** Ran the real `updateSourcePackageJson` (heavy `rolldown` imports stripped, filtering code untouched) over every generator-driven package. Regenerated `exports` are **byte-identical to `main`** — the newly colocated `tab-store.test.ts`, `polygon.test.ts`, `blur.test.react.tsx`, `select.test.react.tsx`, `__utils.test.ts` are all excluded, and no legitimate export is dropped. (`@ariakit/scripts`/`@ariakit/tailwind` keep hand-maintained exports and don't run the generator.)
- **`.npmignore` output unchanged.** The extracted `npmIgnoreEntries` array reproduces the previous inline list verbatim (same entries, same order), so the published tarball filter is identical.
- **New test passes** against the real implementation, including the `*.config.*`, nested-`__tests__`, and `src/test.ts` cases.
- **Test discovery unaffected.** Vitest `include` (`**/*test.{ts,tsx}`) and the package `tsconfig` include/exclude globs are all name-based, not directory-based, so moving files out of `__tests__` keeps them discovered and keeps them out of the dist build/typecheck.
- **`useAllEvents` inlining is justified** — a standalone non-test `use-all-events.ts` would leak as a public `./use-all-events` export under the new filter; only a `__`-prefixed name stays private. Replied to your own thread with the verified mechanism.

**Findings:**
- *(non-blocking, inline on `build.ts`)* The hand-rolled glob matcher implements only a subset of `.npmignore`/gitignore syntax, so the "single source of truth" coupling is partial and could silently diverge if `npmIgnoreEntries` later uses unsupported constructs. Suggested `ignore` (already transitively in the lockfile) for exact parity, or a comment documenting the supported subset.
- *(reply on your `blur.test.react.tsx` thread)* `__use-all-events.ts` is the correct extraction target; the `__` prefix is what keeps it out of exports.

**Not exhaustively validated:** I did not run the full `pnpm test`/`pnpm tsc`/`pnpm build` here (this worktree has no `node_modules`), but the moved tests are pure import-path renames and you reported those commands green in the PR description; the export/discovery/typecheck-config behavior I checked directly.

No changeset needed — verified byte-identical `exports` and `.npmignore`, so there's no user-facing change.

packages/ariakit-test/src/blur.test.react.tsx
diegohaz: Fixed in 1f9458bdd - extracted the event listener helper to `__use-all-events.test.helper.ts` so it can be reused by future tests while staying private from package exports and excluded from the npm tarball.

packages/ariakit-scripts/src/build.ts
diegohaz: Fixed in fc1814830 - documented that the custom matcher intentionally supports only `*`, `**`, and slash-free basename patterns, so future `npmIgnoreEntries` changes stay within the supported subset.

diegohaz:
**Fixed**

- `1f9458bdd` ([reply](https://github.com/ariakit/ariakit/pull/6144#discussion_r3329938506)): Extracted the reusable event listener helper to `__use-all-events.test.helper.ts`, keeping it private from exports and excluded from the npm tarball.
- `fc1814830` ([reply](https://github.com/ariakit/ariakit/pull/6144#discussion_r3329938531)): Documented the supported npm-ignore matcher subset (`*`, `**`, and slash-free basename patterns).

diegohaz (commented):
### Re-review summary — incremental

**Type:** Re-review. **Baseline:** my last reviewed head `b74a73a` → current head `fc181483` (2 new commits: `1f9458bdd` extract helper, `fc1814830` document matcher). Scope: `build.ts` (+2-line comment), new `__use-all-events.test.helper.ts`, and `blur.test.react.tsx`.

**Both prior findings resolved — verified in the current code, not taken at face value:**

1. **Matcher scope (`build.ts` thread) — resolved.** `fc1814830` documents that the matcher intentionally supports only `*`, `**`, and slash-free basename patterns. Accurate and matches the implementation.
2. **Helper extraction (`blur.test.react.tsx` thread) — resolved.** `1f9458bdd` moves `useAllEvents` into `__use-all-events.test.helper.ts`. I verified:
   - **Not exported** — regenerating `@ariakit/test` source exports at this head is byte-identical to before; the helper is in neither the `exports` map nor the published tarball (excluded both by the `__` prefix via `isPrivateModule` and by the `.test.` infix via the npmignore filter).
   - **Not collected as a test** — the `.helper.ts` suffix matches neither Vitest `include` glob (`**/*test.{ts,tsx}`, `**/*test.react.{ts,tsx}`), so it's treated as a helper, not a zero-test test file.
   - **Typechecks/builds** — CI `Type check`, `Build`, `Build packages`, and `Lint` are green on `fc181483`; `Test (2/2)` and `Test React 18 (2/2)` pass (remaining test shards still running, none failed).

**New issues introduced by these commits:** none.

**Optional (non-blocking, not worth a thread):** the new matcher comment sits above `matchesPathPattern`, a fair distance from the `npmIgnoreEntries` array it's guiding — a one-liner at the array declaration would be more likely to catch the eye of someone adding an entry. Purely cosmetic.

Net: the change is correct and complete. No blocking concerns.

PR #6146, opened 2026-05-31
Extract admin avatar URL helper

diegohaz:
## Motivation

The admin users and teams pages each carried the same avatar URL resizing logic. The users page implementation also depended on an ambient global `User` type in its page-local helper.

## Solution

Add a shared `getAvatarUrl(imageUrl, size)` helper in `app/src/pages/admin/_utils.ts` and use it from both admin pages. The helper preserves the existing `width=32` and `height=32` behavior by default while removing the duplicated page-local functions and narrowing the call sites to the `imageUrl` string they actually need.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build-app-lite`

### DIFF
--- a/app/src/pages/admin/_utils.ts
+++ b/app/src/pages/admin/_utils.ts
@@ -77,6 +77,13 @@ export function getStripeDashboardUrl() {
   return `https://dashboard.stripe.com`;
 }
 
+export function getAvatarUrl(imageUrl: string, size = 32) {
+  const url = new URL(imageUrl);
+  url.searchParams.set("width", String(size));
+  url.searchParams.set("height", String(size));
+  return url.toString();
+}
+
 interface GetPageUrlsOptions {
   totalCount: number;
   limit: number;
--- a/app/src/pages/admin/teams.astro
+++ b/app/src/pages/admin/teams.astro
@@ -12,9 +12,9 @@ import { InlineLink } from "#app/components/inline-link.react.tsx";
 import { Icon } from "#app/icons/icon.react.tsx";
 import { isAdmin } from "#app/lib/auth.ts";
 import { clerkClient } from "@clerk/astro/server";
-import type { Organization } from "@clerk/astro/server";
 import LayoutPagination from "./_layout-pagination.astro";
 import {
+  getAvatarUrl,
   getClerkInstanceUrl,
   getPageUrls,
   getPaginationData,
@@ -51,13 +51,6 @@ async function getTeamData() {
 }
 
 const teams = await getTeamData();
-
-function getTeamAvatarUrl(team: Organization) {
-  const url = new URL(team.imageUrl);
-  url.searchParams.set("width", "32");
-  url.searchParams.set("height", "32");
-  return url.toString();
-}
 ---
 
 <LayoutPagination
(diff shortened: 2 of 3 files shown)


app/src/pages/admin/_utils.ts
```diff
@@ -77,6 +77,13 @@ export function getStripeDashboardUrl() {
   return `https://dashboard.stripe.com`;
 }
 
+export function getAvatarUrl(imageUrl: string, size = 32) {
```
diegohaz, line 80: The extraction itself is clean and behavior-preserving — I confirmed locally that `getAvatarUrl(url)` produces byte-identical output to the old `"32"` literals for the default size, including when the source URL already carries `width`/`height` params (they get overwritten the same way).

One clarification on the PR description: it states the users page "referenced a `User` type in that local helper without importing it." That `User` reference wasn't actually unresolved — `app/env.d.ts` (line 27: `type User = import("@clerk/astro/server").User;`) declares `User` as a global ambient type, so the old `getUserAvatarUrl(user: User)` typechecked fine on `main`. The refactor is still a genuine improvement (the new `imageUrl: string` signature no longer depends on that ambient global, and the duplication is gone), but the description implies a type error was fixed when there wasn't one. Worth tweaking the wording so the motivation matches reality.

Non-blocking: the `size` parameter is never passed by either caller (both rely on the default `32`). That's a fine, cheap extension point, just noting it's currently unused.

diegohaz (commented):
**Initial review** of `fdb109b01dbf0aacb7c2bbbdabc3d718217672d6` (diff against merge-base `c863ea104`).

**Scope**: `app/src/pages/admin/_utils.ts`, `teams.astro`, `users.astro` (+11/-17, 3 files). Pure refactor extracting a shared `getAvatarUrl(imageUrl, size = 32)` helper and replacing the two page-local avatar helpers.

**What I checked**
- **Behavior preservation**: ran the old vs. new logic side by side in Node — output is byte-identical for the default size, including the edge case where the source URL already has `width`/`height` params (overwritten identically). No `src` change for either page.
- **Completeness of extraction**: grepped the app for other `searchParams.set("width"/"height")` / avatar-resizing call sites — the two admin pages are the only consumers, so nothing else should adopt the helper.
- **Code style**: top-level `function` keyword and default param are consistent with the rest of `_utils.ts` and the code-style skill. The absence of JSDoc matches the file's existing exported helpers (CodeRabbit's "0% docstring coverage" is a repo-wide pattern, not introduced here).
- **Changeset**: none required — `app` is a private workspace listed in `.changeset/config.json`'s `ignore` array, and this is a no-behavior-change refactor. The changeset-bot warning is a false positive here.
- **Type safety**: the removed `Organization` import in `teams.astro` was genuinely unused after the refactor; correct to drop it.

**Findings** (1, non-blocking, posted inline on `_utils.ts`)
- The PR description says the users page "referenced a `User` type without importing it," implying a type error was fixed. In fact `User` is a global ambient type (`app/env.d.ts:27`), so the base code typechecked fine. The refactor is still worthwhile, but the stated motivation overstates the problem — worth a wording tweak. Also noted the `size` param is currently unused by both callers (fine as an extension point).

**Not fully validated**: did not run `pnpm install` + `astro check` in this worktree (deps not installed); the PR reports `pnpm tsc` / `pnpm build-app-lite` pass, and the change is trivial and type-trivial. No correctness concerns.

Comment-only review; no blocking issues.

app/src/pages/admin/_utils.ts
diegohaz: Updated the PR description to soften that claim. It now focuses on deduplicating the avatar URL logic and reducing the users page helper coupling to the ambient `User` type, without implying the old code had a missing import or failed to typecheck.

PR #6147, opened 2026-05-31
Dedupe Ariakit script helpers

diegohaz:
## Motivation

Audit findings T080 and T081 flagged duplicated helper implementations in `@ariakit/scripts`. `median` and `computeMedianMetrics` were duplicated between `perf.ts` and `perf-compare.ts`, and `escapeRegExp` was duplicated across scripts that build dynamic regular expressions.

## Solution

Export `median` and `computeMedianMetrics` from `perf.ts` and import them in `perf-compare.ts`, removing the duplicate local implementations. Move the shared `escapeRegExp` helper into `regexp.ts` and import it from `docs.ts`, `perf-compare.ts`, and `build.ts`.

`regexp.ts` is used instead of `utils.ts` intentionally because the independent T082 PR owns a new `packages/ariakit-scripts/src/utils.ts` file. This keeps the audit PRs mergeable while preserving the helper dedupe.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test packages/ariakit-scripts/src/build.test.ts packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts packages/ariakit-scripts/src/docs.test.ts`

### DIFF
--- a/packages/ariakit-scripts/src/regexp.ts
+++ b/packages/ariakit-scripts/src/regexp.ts
@@ -1,0 +1,3 @@
+export function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
--- a/packages/ariakit-scripts/src/build.ts
+++ b/packages/ariakit-scripts/src/build.ts
@@ -5,6 +5,7 @@ import { build as rolldownBuild } from "rolldown";
 import type { Plugin } from "rolldown";
 import { dts } from "rolldown-plugin-dts";
 import solidPlugin from "vite-plugin-solid";
+import { escapeRegExp } from "./regexp.ts";
 
 interface PackageJson {
   name: string;
@@ -63,10 +64,6 @@ function isTypeScriptSource(filename: string) {
 
 function isPrivateModule(filename: string) {
   return filename.startsWith("__");
-}
-
-function escapeRegExp(value: string) {
-  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
 
 function matchesBasenamePattern(name: string, pattern: string) {
--- a/packages/ariakit-scripts/src/docs.ts
+++ b/packages/ariakit-scripts/src/docs.ts
@@ -8,6 +8,7 @@ import type {
   SourceFile,
   TypeAliasDeclaration,
 } from "ts-morph";
+import { escapeRegExp } from "./regexp.ts";
 
 interface GenerateDocsOptions {
   rootPath?: string;
@@ -428,10 +429,6 @@ function getLocalTypeDeclarations(sourceFile: SourceFile) {
   return [...sourceFile.getTypeAliases(), ...sourceFile.getInterfaces()]
     .filter((declaration) => !declaration.isExported())
     .sort((a, b) => a.getStart() - b.getStart());
-}
-
-function escapeRegExp(text: string) {
-  return text.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
 
 function getTypeParameterNames(declaration: Node) {
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -6,7 +6,12 @@ import {
   writeFileSync,
 } from "node:fs";
 import path from "node:path";
-import { mergeScriptProfiles, mergeSelectorProfiles } from "./perf.ts";
+import {
+  computeMedianMetrics,
+  median,
+  mergeScriptProfiles,
+  mergeSelectorProfiles,
+} from "./perf.ts";
 import type {
   PerfMetrics,
   PerfProfiles,
@@ -14,6 +19,7 @@ import type {
   PerfScriptProfileEntry,
   PerfSelectorProfileEntry,
 } from "./perf.ts";
+import { escapeRegExp } from "./regexp.ts";
 
 const RESULTS_DIR = path.join(process.cwd(), ".perf-results");
 const PROFILE_LIMIT = 10;
@@ -194,10 +200,6 @@ function readJsonFile(filePath: string): unknown {
   }
 }
 
-function escapeRegExp(value: string): string {
-  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
-}
-
 function getNumber(value: unknown): number {
   if (typeof value !== "number") return 0;
   if (!Number.isFinite(value)) return 0;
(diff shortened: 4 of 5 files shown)


packages/ariakit-scripts/src/regexp.ts
```diff
@@ -0,0 +1,3 @@
+export function escapeRegExp(value: string): string {
```
diegohaz, line 1: This PR centralizes `escapeRegExp` into this new shared module and rewires `docs.ts` and `perf-compare.ts` to it, but a fourth byte-identical copy is left behind in the same package at `packages/ariakit-scripts/src/build.ts` (line 68):

```ts
function escapeRegExp(value: string) {
  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
}
```

Since the whole point of T081 is to remove duplicated `escapeRegExp` implementations, leaving this one undeduped against the brand-new shared module is an avoidable inconsistency — and it's a sibling file that already imports other local `./*.ts` modules, so the change is trivial:

```diff
+import { escapeRegExp } from "./regexp.ts";
 ...
-function escapeRegExp(value: string) {
-  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
-}
```

I verified locally that importing `escapeRegExp` from `./regexp.ts` resolves under native `.ts` execution and that its output is byte-identical to the `build.ts` local implementation across special-character inputs, so this is a safe behavior-preserving move. If `build.ts` was intentionally left out of scope (e.g. owned by another audit PR), a one-line note in the PR description would make that explicit.

diegohaz (commented):
## Review summary (initial review)

**Scope:** Full review of the diff between merge-base `c863ea104` and head `862996b9` (4 files, +13/-35), plus surrounding context in `packages/ariakit-scripts/src`. This is the first review carrying this skill's marker.

**What the PR does:** Dedupes audit findings T080/T081 — exports `median`/`computeMedianMetrics` from `perf.ts` for reuse in `perf-compare.ts`, and moves `escapeRegExp` into a new `regexp.ts` consumed by `docs.ts` and `perf-compare.ts`.

**Verification performed:**
- Diffed the removed bodies against the kept/shared ones: `median` and `computeMedianMetrics` removed from `perf-compare.ts` are byte-identical to the now-exported `perf.ts` versions; both old `escapeRegExp` copies are functionally identical to the new `regexp.ts` one (only the `docs.ts` parameter name `text`→`value` differs). Behavior-preserving.
- Confirmed no circular import: `perf.ts` does not import `perf-compare.ts`.
- Loaded the new `regexp.ts` under native `.ts` execution (Node 24) and confirmed its output matches the `build.ts` local `escapeRegExp` byte-for-byte across special-char inputs.
- Checked import ordering/grouping in `perf-compare.ts` (alphabetical, correctly placed) and call sites of all three helpers.

**Findings:**
- One low-severity inline finding: a fourth byte-identical `escapeRegExp` remains in `packages/ariakit-scripts/src/build.ts` (line 68). Since this PR introduces the shared `regexp.ts` specifically to dedupe `escapeRegExp`, `build.ts` is a trivial missed consumer in the same package. Left as a comment rather than a blocker since it's outside the strict T080/T081 scope.

**Not flagged (intentional / acceptable):**
- No changeset: correct — `@ariakit/scripts` is a private package and this is a behavior-preserving refactor.
- No unit test for `regexp.ts`: acceptable — the move is byte-identical and these are internal build scripts already exercised end-to-end by `perf-compare.test.ts`/`docs.test.ts`.
- Using `regexp.ts` instead of `utils.ts` is explained in the PR description (T082 owns a new `utils.ts`); keeps the audit PRs independently mergeable.

**Not fully validated:** Did not run `pnpm tsc`/`pnpm test` — the dedicated review worktree has no installed deps and the change is a provably byte-identical internal move; the targeted runtime check above covers the load-bearing behavior. CodeRabbit and Copilot both reported no actionable comments.

packages/ariakit-scripts/src/regexp.ts
diegohaz: Addressed in 9e7e8b59d: `build.ts` now imports `escapeRegExp` from `./regexp.ts`, and the local byte-identical helper was removed. I reran `pnpm lint-fix`, `pnpm tsc`, and the targeted scripts tests including `build.test.ts`.

diegohaz: Confirmed resolved in 9e7e8b59d. I verified against the current code, not just the reply: `grep` shows the only `escapeRegExp` definition now lives in `regexp.ts`, all four prior copies are gone, and `build.ts` imports it (sorted correctly at the end of the import block). The `matchesBasenamePattern` call site (`pattern.split("*").map(escapeRegExp).join(".*")`) is unchanged and works by reference. I re-ran the call-site behavior under native `.ts` execution across glob inputs (`*.d.ts`, `foo.*.js`, `a*b*c`, `[x].*`, etc.) and the escaped output is byte-identical to the old local helper. The added `: string` return type is a strict improvement. Thanks for the quick turnaround.

diegohaz (commented):

**Baseline:** my last marked review at `862996b9`. **New head:** `9e7e8b59d` — single new commit "Use shared regexp helper in build script".

**Incremental diff reviewed:** `862996b9...9e7e8b59d` — one file, `packages/ariakit-scripts/src/build.ts` (+1/-4). This commit directly addresses the only finding from my initial review.

**Prior finding — RESOLVED (verified):** The 4th byte-identical `escapeRegExp` copy in `build.ts` has been removed and replaced with `import { escapeRegExp } from "./regexp.ts"`. I confirmed against the current code (not just the author's reply) that `regexp.ts` now holds the sole definition, all four prior copies are gone, the import is correctly ordered, and the `matchesBasenamePattern` call site is unchanged. Replied on the original thread (#discussion_r3329974063) with the verification details.

**Verification performed this pass:**
- `grep` for every `escapeRegExp` reference/definition across `packages/ariakit-scripts/src` — single definition in `regexp.ts`, four consumers (`regexp`/`build`/`docs`/`perf-compare`).
- Re-ran the `build.ts` call-site logic (`pattern.split("*").map(escapeRegExp).join(".*")`) under native `.ts` execution (Node 24) across glob inputs — escaped output is byte-identical to the removed local helper.
- Confirmed worktree clean and head SHA on GitHub matches `9e7e8b59d` before posting.

**New findings:** None. The incremental change is a clean, behavior-preserving dedupe with no new issues. The added explicit `: string` return type on the shared helper is a minor improvement over the old untyped `build.ts` copy.

**Unchanged conclusions from the initial review still hold:** no changeset needed (private `@ariakit/scripts` refactor); no new unit test required (byte-identical internal move). Did not run `pnpm tsc`/`pnpm test` locally (review worktree has no installed deps); the author reports running `pnpm lint-fix`, `pnpm tsc`, and the targeted scripts tests including `build.test.ts`, and the targeted runtime check above covers the load-bearing behavior.

PR #6148, opened 2026-05-31
Extract activation target helper

diegohaz:
## Motivation

`isOpeningInNewTab` and `isDownloading` both duplicated the same `currentTarget` cast, null handling, and navigation-target classification logic for anchors and submit controls. T070 tracks extracting that repeated logic without changing the public helpers' modifier-key behavior.

## Solution

Add a private `isActivatableNavigationTarget` helper in `packages/ariakit-utils/src/events.ts` and have both exported helpers call it after their own modifier-key guard. The helper keeps the existing anchor, submit button, and submit input checks in one place.

Verification:

- `pnpm lint-fix`
- `pnpm build`
- `pnpm tsc`
- `pnpm exec vitest packages/ariakit-utils/src/events.test.ts --run`

### DIFF
--- a/packages/ariakit-utils/src/events.ts
+++ b/packages/ariakit-utils/src/events.ts
@@ -32,26 +32,32 @@ export function isSelfTarget(
   return event.target === event.currentTarget;
 }
 
+function isActivatableNavigationTarget(
+  element: EventTarget | null,
+): element is HTMLAnchorElement | HTMLButtonElement | HTMLInputElement {
+  const target = element as
+    | HTMLAnchorElement
+    | HTMLButtonElement
+    | HTMLInputElement
+    | null;
+  if (!target) return false;
+  const tagName = target.tagName.toLowerCase();
+  if (tagName === "a") return true;
+  if (tagName === "button" && target.type === "submit") return true;
+  if (tagName === "input" && target.type === "submit") return true;
+  return false;
+}
+
 /**
  * Checks whether the user event is triggering a page navigation in a new tab.
  */
 export function isOpeningInNewTab(
   event: Pick<MouseEvent, "currentTarget" | "metaKey" | "ctrlKey">,
 ) {
-  const element = event.currentTarget as
-    | HTMLAnchorElement
-    | HTMLButtonElement
-    | HTMLInputElement
-    | null;
-  if (!element) return false;
   const isAppleDevice = isApple();
   if (isAppleDevice && !event.metaKey) return false;
   if (!isAppleDevice && !event.ctrlKey) return false;
-  const tagName = element.tagName.toLowerCase();
-  if (tagName === "a") return true;
-  if (tagName === "button" && element.type === "submit") return true;
-  if (tagName === "input" && element.type === "submit") return true;
-  return false;
+  return isActivatableNavigationTarget(event.currentTarget);
 }
 
 /**
@@ -60,18 +66,8 @@ export function isOpeningInNewTab(
 export function isDownloading(
   event: Pick<MouseEvent, "altKey" | "currentTarget">,
 ) {
-  const element = event.currentTarget as
-    | HTMLAnchorElement
-    | HTMLButtonElement
-    | HTMLInputElement
-    | null;
-  if (!element) return false;
-  const tagName = element.tagName.toLowerCase();
   if (!event.altKey) return false;
-  if (tagName === "a") return true;
-  if (tagName === "button" && element.type === "submit") return true;
-  if (tagName === "input" && element.type === "submit") return true;
-  return false;
+  return isActivatableNavigationTarget(event.currentTarget);
 }
 
 /**
--- a/packages/ariakit-utils/src/events.ts
+++ b/packages/ariakit-utils/src/events.ts
@@ -41,6 +41,7 @@ function isActivatableNavigationTarget(
     | HTMLInputElement
     | null;
   if (!target) return false;
+  if (!(target instanceof HTMLElement)) return false;
   const tagName = target.tagName.toLowerCase();
   if (tagName === "a") return true;
   if (tagName === "button" && target.type === "submit") return true;
--- a/packages/ariakit-utils/src/events.ts
+++ b/packages/ariakit-utils/src/events.ts
@@ -3,7 +3,7 @@
  * @module Event utilities
  */
 
-import { contains, isNode } from "./dom.ts";
+import { contains, isElement, isNode } from "./dom.ts";
 import { isApple } from "./platform.ts";
 
 /**
@@ -32,16 +32,12 @@ export function isSelfTarget(
   return event.target === event.currentTarget;
 }
 
-function isActivatableNavigationTarget(
-  element: EventTarget | null,
-): element is HTMLAnchorElement | HTMLButtonElement | HTMLInputElement {
+function isActivatableNavigationTarget(element: EventTarget | null) {
+  if (!isElement(element)) return false;
   const target = element as
     | HTMLAnchorElement
     | HTMLButtonElement
-    | HTMLInputElement
-    | null;
-  if (!target) return false;
-  if (!(target instanceof HTMLElement)) return false;
+    | HTMLInputElement;
   const tagName = target.tagName.toLowerCase();
   if (tagName === "a") return true;
   if (tagName === "button" && target.type === "submit") return true;


packages/ariakit-utils/src/events.ts
```diff
@@ -32,26 +32,33 @@ export function isSelfTarget(
...
+function isActivatableNavigationTarget(
+  element: EventTarget | null,
+): element is HTMLAnchorElement | HTMLButtonElement | HTMLInputElement {
+  const target = element as
+    | HTMLAnchorElement
+    | HTMLButtonElement
+    | HTMLInputElement
+    | null;
+  if (!target) return false;
+  if (!(target instanceof HTMLElement)) return false;
```
diegohaz: The `instanceof HTMLElement` guard added in 06ce066 is stricter than the original code and changes behavior for **SVG anchors**.

On `main`, both helpers read `tagName` on any non-null `currentTarget`, so an `<a>` inside an `<svg>` (an `SVGAElement`) matched `tagName === "a"` and returned `true`. `SVGAElement` is **not** an `HTMLElement`, so this guard now makes both `isOpeningInNewTab` and `isDownloading` return `false` for SVG anchors. I confirmed this locally in jsdom:

```
SVG <a> tagName=a  HTMLElement=false  SVGElement=true
isDownloading(svgA)  -> main: true   this PR: false
```

The throw-safety concern this guard was meant to fix (a non-element `EventTarget` such as `window`/`document` reaching `.tagName`) is real, but the package already has a guard for exactly this — `isElement` in `./dom.ts`, which uses an SSR-safe `nodeType === 1` check. It returns `false` for non-element `EventTarget`s (fixing the throw) while still returning `true` for SVG anchors (no regression), and it's consistent with the sibling helper `isPortalEvent`, which already uses `isNode` from the same module. It also subsumes the `if (!target) return false` null check, so the helper simplifies:

```suggestion
function isActivatableNavigationTarget(
  element: EventTarget | null,
): element is HTMLAnchorElement | HTMLButtonElement | HTMLInputElement {
  if (!isElement(element)) return false;
  const target = element as
    | HTMLAnchorElement
    | HTMLButtonElement
    | HTMLInputElement;
  const tagName = target.tagName.toLowerCase();
  if (tagName === "a") return true;
  if (tagName === "button" && target.type === "submit") return true;
  if (tagName === "input" && target.type === "submit") return true;
  return false;
}
```

(Requires `import { isElement } from "./dom.ts";` — `contains`/`isNode` are already imported from there.)

Verified locally: `isElement(svgA) === true`, `isElement(new EventTarget()) === false`, `isElement(null) === false`, and `pnpm tsc` passes. The practical impact is small since these helpers only ever receive a click handler's `currentTarget` on composite items, but if the intent is a behavior-preserving refactor (per the PR description and T070), `isElement` keeps the original anchor classification while still hardening against the throw.

packages/ariakit-utils/src/events.ts
```diff
@@ -32,26 +32,32 @@ export function isSelfTarget(
...
+    | HTMLButtonElement
+    | HTMLInputElement
+    | null;
+  if (!target) return false;
+  const tagName = target.tagName.toLowerCase();
+  if (tagName === "a") return true;
+  if (tagName === "button" && target.type === "submit") return true;
+  if (tagName === "input" && target.type === "submit") return true;
+  return false;
+}
```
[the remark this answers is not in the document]
diegohaz: Follow-up: the `instanceof HTMLElement` guard applied in 06ce066 fixes the throw, but it's stricter than the original code and regresses SVG anchors (`SVGAElement` has `tagName === "a"` but is not an `HTMLElement`, so it's now classified as non-activatable where `main` returned `true`). The package's existing `isElement` guard (`./dom.ts`, SSR-safe `nodeType === 1`) rejects non-element `EventTarget`s without the SVG regression and matches the pattern the sibling `isPortalEvent` already uses. Details and a suggested diff are in a separate inline comment on the guard line.

diegohaz (commented):
**Initial review** of head `06ce0666a` (base `main`, merge-base `c863ea104`). Scope: `packages/ariakit-utils/src/events.ts` (+19/-22), two commits — extracting `isActivatableNavigationTarget`, then adding an `instanceof HTMLElement` guard in response to CodeRabbit.

The extraction itself is clean and removes the duplicated cast / null-check / tag classification from both `isOpeningInNewTab` and `isDownloading`. Modifier-key behavior is preserved.

**One substantive finding** (inline on the guard line, plus a reply on the CodeRabbit thread):

- The `instanceof HTMLElement` guard added in `06ce066` is stricter than the original code and regresses **SVG anchors**. On `main` an `<a>` inside `<svg>` (`SVGAElement`, `tagName === "a"`) returned `true`; with this guard both helpers now return `false` for it. The package already exports `isElement` (`./dom.ts`, SSR-safe `nodeType === 1`), which fixes the original throw concern for non-element `EventTarget`s **without** the SVG regression, subsumes the null check, and is consistent with `isPortalEvent`'s use of `isNode` in the same module. I suggested switching to `isElement`.

**Validation performed (local, Node 24, this PR's head):**
- `pnpm exec vitest packages/ariakit-utils/src/events.test.ts --run` — 4 passed.
- Throwaway jsdom checks (removed afterward; tree left clean): confirmed `isDownloading(svgAnchor)` is `true` on `main` vs `false` here; the original `.tagName` read throws on a bare `EventTarget`; `isElement` returns `true` for both HTML and SVG anchors and `false` for `EventTarget`/`null`.
- `pnpm tsc` — 0 errors.

**Notes:** The `element is HTMLAnchorElement | …` type predicate is never used for narrowing by callers (they consume only the boolean) — not worth changing on its own. No changeset is needed: this is an internal `@ariakit/utils` refactor; the SVG behavior change is theoretical for these click-only handlers and the throw fix is effectively unreachable from real call sites, so there is no documented public-API behavior change reaching users. The missing-changeset bot warning is therefore expected and fine.

Comment-only review; not approving or requesting changes.

packages/ariakit-utils/src/events.ts
diegohaz: Fixed in 568e0dce3 by replacing the `instanceof HTMLElement` guard with the existing `isElement` helper. This still rejects non-element EventTargets before `tagName` access, while preserving SVG anchor handling because `SVGAElement` passes the `nodeType === 1` element guard.

diegohaz: Confirmed resolved in 568e0dce3. The guard now uses `isElement` and the redundant null check / `| null` cast / unused type predicate were dropped. Verified the runtime behavior in jsdom on this head:

- `isDownloading`/`isOpeningInNewTab` with an `SVGAElement` → `true` again (regression gone, matches `main`).
- non-element `EventTarget` (e.g. `window`) → `false`, no throw (the original throw-safety concern stays fixed).
- `null` currentTarget → `false`; HTML anchor / submit button / submit input → `true`; non-activatable `div` → `false`.

`pnpm exec vitest .../events.test.ts --run` (4 passed) and `pnpm tsc` (0 errors) both pass. Thanks for the quick turnaround.

diegohaz (commented):
**Incremental re-review** of `568e0dce3`, baselined against my last reviewed SHA `06ce0666a`. The only change in this range is commit "Use element guard for activation targets", scoped to `packages/ariakit-utils/src/events.ts`.

**Prior finding resolved.** My previous review flagged that the `instanceof HTMLElement` guard regressed SVG anchors. This commit addresses it exactly as recommended: it imports `isElement` from `./dom.ts` and replaces `instanceof HTMLElement` with `if (!isElement(element)) return false`, which also lets the helper drop the now-redundant null check, the `| null` in the cast, and the unused type predicate. The result is consistent with `isPortalEvent`'s use of `isNode` in the same module.

**Verified in code + jsdom runtime on this head (not just the diff):**
- `SVGAElement` → `isDownloading`/`isOpeningInNewTab` return `true` again (regression gone; matches `main`).
- non-element `EventTarget` (`window`/`EventTarget`) → `false`, no throw (original throw-safety concern still addressed).
- `null` → `false`; HTML anchor / submit button / submit input → `true`; non-activatable `div` → `false`; `altKey: false` → `false`.
- `pnpm exec vitest packages/ariakit-utils/src/events.test.ts --run` → 4 passed.
- `pnpm tsc` → 0 errors.

No new findings introduced by this commit. No changeset needed (internal `@ariakit/utils` refactor; net behavior now matches `main` for real call sites). Threads from this run's prior review are addressed. Comment-only review; not approving or requesting changes.

PR #6149, opened 2026-05-31
Consolidate Ariakit scripts package helpers

diegohaz:
## Motivation

The Ariakit scripts package had duplicated `normalizePath` and package JSON parsing helpers across `build.ts`, `dev.ts`, and `react18.ts`. Keeping those copies separate makes small behavior or typing changes easier to drift over time.

## Solution

Add an internal `utils.ts` module for the shared path normalization and sync/async package JSON readers. The existing scripts now import those helpers, with `react18.ts` using the sync directory reader for root package metadata and the async file-path reader for discovered package files.

No changeset is included because this is an internal refactor with no shipped behavior change.

### DIFF
--- a/packages/ariakit-scripts/src/react18.ts
+++ b/packages/ariakit-scripts/src/react18.ts
@@ -369,10 +369,7 @@ function printHelp() {
   );
 }
 
-async function getReact18Command(
-  rootPath: string,
-  args: string[],
-): Promise<React18Command> {
+function getReact18Command(rootPath: string, args: string[]): React18Command {
   const [command, ...commandArgs] = stripLeadingSeparator(args);
 
   if (!command) {
@@ -514,7 +511,7 @@ export async function react18(args: string[]) {
     { cwd: workspacePath },
   );
 
-  const command = await getReact18Command(workspacePath, args);
+  const command = getReact18Command(workspacePath, args);
   const watcher = startWorkspaceWatcher(rootPath, workspacePath);
 
   try {
--- a/packages/ariakit-scripts/src/react18.ts
+++ b/packages/ariakit-scripts/src/react18.ts
@@ -4,24 +4,21 @@ import {
   mkdir,
   copyFile,
   readdir,
-  readFile,
   readlink,
   rm,
   symlink,
   writeFile,
 } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { delimiter, dirname, isAbsolute, join, relative, sep } from "node:path";
+import { delimiter, dirname, isAbsolute, join, relative } from "node:path";
 import { watch } from "chokidar";
 import type { FSWatcher } from "chokidar";
-
-interface PackageJson {
-  scripts?: Record<string, string>;
-  dependencies?: Record<string, string>;
-  devDependencies?: Record<string, string>;
-  optionalDependencies?: Record<string, string>;
-  [key: string]: unknown;
-}
+import {
+  normalizePath,
+  readPackageJson,
+  readPackageJsonAsync,
+} from "./utils.ts";
+import type { PackageJson } from "./utils.ts";
 
 interface React18Command {
   bin: string;
@@ -63,10 +60,6 @@ const pathKey =
 
 function log(message: string) {
   console.error(`[react18] ${message}`);
-}
-
-function normalizePath(path: string) {
-  return path.split(sep).join("/");
 }
 
 function isErrnoException(error: unknown): error is NodeJS.ErrnoException {
@@ -294,13 +287,8 @@ function getDependencyVersion(packageJson: PackageJson, name: string) {
   throw new Error(`${name} must be declared in website/package.json`);
 }
 
-async function readPackageJson(path: string) {
-  const contents = await readFile(path, "utf-8");
-  return JSON.parse(contents) as PackageJson;
-}
-
 async function getReact18DependencyVersions(rootPath: string) {
-  const packageJson = await readPackageJson(
+  const packageJson = await readPackageJsonAsync(
     join(rootPath, "website/package.json"),
   );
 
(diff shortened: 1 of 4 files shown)


packages/ariakit-scripts/src/react18.ts
```diff
@@ -395,7 +379,7 @@ async function getReact18Command(
     throw new Error("Missing command. Use `ariakit react18 <script>`.");
   }
 
-  const packageJson = await readRootPackageJson(rootPath);
+  const packageJson = readPackageJson(rootPath);
```
diegohaz, line 379: Now that this call uses the synchronous `readPackageJson(rootPath)`, `getReact18Command` no longer `await`s anything, so its `async` keyword and `Promise<React18Command>` return type are vestigial — they were only there for the old `await readRootPackageJson(...)` that this PR removed. It's harmless (the caller on line 517 `await`s a plain value), but the signature now overstates what the function does.

Optional cleanup — drop the `async` and unwrap the return type:

```suggestion
function getReact18Command(
  rootPath: string,
  args: string[],
): React18Command {
```

I verified locally that this typechecks cleanly (`tsc -b packages/ariakit-scripts/tsconfig.node.json`) with the awaiting caller unchanged.

diegohaz (commented):
**Initial review** of `@ariakit/scripts` helper consolidation. Baseline: merge-base `c863ea1`..`b84f51e` (both commits on the branch). Scope: all 4 changed files plus surrounding call sites and the package's tests/types.

**What I checked**
- Read the full `utils.ts`, `build.ts`, `dev.ts`, and `react18.ts` (not just the diff) to confirm the extracted `normalizePath` / `readPackageJson` / `readPackageJsonFile` are behavior-equivalent to the removed local copies, and that every remaining import is still used.
- Typechecked the package: `tsc -b packages/ariakit-scripts/tsconfig.node.json` — clean.
- Ran `build.test.ts` (covers the sync `readPackageJson` path via `updateSourcePackageJson`) — passing. Lint (`oxlint`) on the changed files — clean.

**Findings**
- One optional cleanup (inline): `getReact18Command` is still `async` / `Promise`-typed even though its only `await` was removed by this PR. Validated that dropping it typechecks.

**Things considered and dismissed**
- CodeRabbit's "divergent reader contracts" nitpick is already addressed by the second commit renaming `readPackageJsonAsync` → `readPackageJsonFile`, which now encodes "file path" in the name. No further action needed.
- The shared `PackageJson` now requires `name: string`. This is correct/necessary for `build.ts` (which dereferences `packageJson.name` unguarded) and harmless for `react18.ts` (it never reads `.name`, and the readers return an unvalidated `JSON.parse` cast, so there's no runtime risk for nested package.json files that lack a name). Not a regression.
- The `react18` root read switching from async `readFile` to sync `readFileSync` reads the same file with no concurrency in that path — behavior-equivalent.
- No changeset needed: `@ariakit/scripts` is private and changeset-ignored; this is an internal refactor with no shipped/public-API change. The PR description's rationale is accurate. The changeset-bot warning is expected.

No correctness, compatibility, or performance concerns. The consolidation reads cleanly and reduces real duplication.

packages/ariakit-scripts/src/react18.ts
diegohaz: Addressed in a1a073bdd by making `getReact18Command` synchronous and removing the caller `await` as well after `pnpm lint-fix` flagged `await-thenable`. Verified locally with `pnpm lint-fix`, `pnpm exec tsc -b packages/ariakit-scripts/tsconfig.json`, and `pnpm exec vitest run packages/ariakit-scripts/src`.

diegohaz: Confirmed resolved in a1a073b. `getReact18Command` is now synchronous (return type `React18Command`) and the caller's `await` was dropped too — cleaner than my original suggestion, which would have left a harmless `await` on a non-promise. Independently verified: `tsc -b packages/ariakit-scripts/tsconfig.node.json` and `oxlint` on react18.ts both clean. No further action needed.

diegohaz (commented):
**Incremental re-review** of the single new commit a1a073b ("Simplify React 18 command selection"). Baseline: my last reviewed SHA b84f51e. Scope: the only changed file, `react18.ts`.

This commit addresses my prior nit: `getReact18Command` is now synchronous (`React18Command` return type instead of `async`/`Promise`), and the caller's `await` was dropped as well — slightly cleaner than my suggested patch. The previously flagged thread is resolved.

**Validation:** `tsc -b packages/ariakit-scripts/tsconfig.node.json` clean, `oxlint` on react18.ts clean. `react18` itself correctly remains `async` (it still awaits `runChecked`/`runCommand`/`closeWatcher`).

No new findings. No changeset needed (private, changeset-ignored package; internal refactor). Nothing else changed since the last reviewed SHA.

diegohaz (commented):
**Incremental re-review** of the new head 295452a ("Merge origin/main into T082 branch"). Baseline: my last reviewed SHA a1a073b. This commit is a merge of `origin/main` into the branch — no new hand-written changes to this PR's own logic.

**What the merge brings in / how it interacts with this PR**
- main now contains #6147 ("Dedupe Ariakit script helpers"), which extracted `escapeRegExp` into a new `regexp.ts`. The merge integrated this cleanly with #6149's work: `build.ts` now imports `escapeRegExp` from `./regexp.ts` and `normalizePath`/`readPackageJson` from `./utils.ts`, with no duplicated definitions and no conflict markers.
- This PR's core contribution (`utils.ts`) is byte-identical to what I reviewed originally — the merge did not disturb it.
- The dedupe is still needed: `origin/main` still has three duplicate `normalizePath` copies (build.ts, dev.ts, react18.ts) plus duplicate package.json readers that this PR consolidates. #6147 (`escapeRegExp`) and #6149 (`normalizePath` + package.json readers) are complementary, not overlapping.

**Validation on the merged state:** `tsc -b packages/ariakit-scripts/tsconfig.node.json` clean, `build.test.ts` passing, `oxlint` on the four touched scripts files clean. PR is MERGEABLE.

No new findings. My earlier `getReact18Command` nit remains resolved and is unaffected by the merge. No changeset needed (private, changeset-ignored package).

PR #6150, opened 2026-05-31
Unify combobox item role helper

diegohaz:
## Motivation

Combobox items and offscreen combobox items had separate copies of the popup-role to item-role mapping. Keeping two helpers in sync is unnecessary and can let the active and offscreen item paths drift.

## Solution

Hoist the combobox item role lookup in `combobox-item.tsx`, export a shared `getItemRole` helper, and have `combobox-item-offscreen.tsx` reuse it through the existing `Base` import. The lookup table stays module-scoped so the public subpath only exposes the helper, and the helper uses an own-property guard so inherited object keys still fall back to `option`.

## Verification

- `pnpm lint-fix`
- `pnpm -F @ariakit/react-components run build`
- `pnpm tsc`
- `pnpm test examples/combobox/test.ts examples/select-combobox/test.ts examples/menu-combobox/test.ts`


packages/ariakit-react-components/src/combobox/combobox-item.tsx
```diff
@@ -35,6 +36,12 @@ const TagName = "div" satisfies ElementType;
 type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
 
+const itemRoleByPopupRole = {
+  menu: "menuitem",
+  listbox: "option",
+  tree: "treeitem",
+};
```
diegohaz, line 43: This unifies the two combobox copies of the role map, which is a real improvement. But there is a third, byte-identical copy of this exact table in `@ariakit/utils`:

```ts
// packages/ariakit-utils/src/dom.ts
const itemRoleByPopupRole = {
  menu: "menuitem",
  listbox: "option",
  tree: "treeitem",
};

export function getPopupItemRole(element, fallback) {
  const popupRole = getPopupRole(element);
  if (!popupRole) return fallback;
  const key = popupRole as keyof typeof itemRoleByPopupRole;
  return itemRoleByPopupRole[key] ?? fallback;
}
```

That helper is already the shared popup-role → item-role source of truth used by menu, select, and composite (via `getPopupItemRole`), and `combobox-item.tsx` already imports from `@ariakit/utils`. The combobox helper here is essentially `getPopupItemRole` minus the DOM read (it gets the role string straight from `ComboboxListRoleContext`) plus a fixed `"option"` fallback.

The PR's own motivation — "keeping two helpers in sync is unnecessary and can let the paths drift" — applies just as strongly to this three-way duplication. Collapsing 2 of 3 copies into a new local copy that still duplicates the `@ariakit/utils` table doesn't fully remove the drift risk.

A more complete fix is to make `@ariakit/utils` the single source by extracting the role-string lookup there and reusing it from both places, e.g.:

```ts
// @ariakit/utils/src/dom.ts
export function getItemRoleByPopupRole(popupRole?: string, fallback: AriaRole = "option") {
  if (popupRole == null) return fallback;
  if (!hasOwnProperty(itemRoleByPopupRole, popupRole)) return fallback;
  return itemRoleByPopupRole[popupRole];
}

export function getPopupItemRole(element, fallback) {
  return getItemRoleByPopupRole(getPopupRole(element) ?? undefined, fallback);
}
```

Then combobox just calls `getItemRoleByPopupRole(popupRole)`, deleting its local table and helper entirely. That removes all three copies instead of leaving two.

Not a blocker — the current change is still a net win — but worth considering since the goal is deduplication. If keeping it combobox-local is intentional (e.g. to avoid touching `@ariakit/utils` in this PR), a short comment noting the intentional duplicate would help future readers.

diegohaz (commented):
## Review summary

**Initial review** of head `4157cb8a` (full diff from merge-base `c863ea10`). Scope: the 3 changed files in `@ariakit/react-components/src/combobox/` (`combobox-item.tsx`, `combobox-item-offscreen.tsx`, and the new colocated `combobox-item.test.ts`).

### Verdict
The refactor is sound and behavior-preserving for real usage. The change is a net improvement: it removes one of two duplicate `getItemRole`/`itemRoleByPopupRole` copies, replaces an unchecked `as keyof` cast with a typed `hasOwnProperty` guard, and adds a focused unit test. No correctness or type regressions found.

### What I validated locally (in a dedicated worktree)
- `pnpm test packages/.../combobox-item.test.ts` — 1 passed.
- `pnpm -F @ariakit/react-components run build` — exit 0; confirmed `getItemRole` is emitted in `dist/combobox/combobox-item.d.ts` (subpath surface) but is NOT present anywhere in the built `@ariakit/react` package, so it does not leak into the top-level public API. The PR's "public subpath only exposes the helper" claim holds.
- `pnpm tsc` — 0 errors / 0 warnings / 0 hints.
- `oxlint` + `oxfmt --check` on the 3 files — clean.
- Reproduced old-vs-new behavior on inherited keys: the previous `itemRoleByPopupRole[key] ?? "option"` returned prototype members for keys like `constructor`/`toString`/`__proto__` (e.g. the `Object` constructor), whereas the new own-property guard correctly returns `"option"`. This is a latent hardening, exercised by the new test cases. It is only reachable if a consumer sets the popup list `role` to an `Object.prototype` member name, so it's effectively a non-issue in practice — but it does make the new test cases meaningful, and the first `popupRole == null` early return is required for type-narrowing before `hasOwnProperty`, not redundant.

### Findings
1 inline finding (non-blocking): a third byte-identical copy of `itemRoleByPopupRole` already lives in `@ariakit/utils` (`getPopupItemRole`/`getPopupRole`), which is the established shared popup-role → item-role helper used by menu/select/composite. This PR's deduplication goal would be more fully served by hoisting the role-string lookup into `@ariakit/utils` and reusing it from both places, removing all three copies rather than two. Details on the thread.

### Changeset
No changeset needed. This is an internal refactor with no observable change to the public `@ariakit/react` API or to documented behavior, consistent with how the recent pure-refactor PR (#6144, colocate tests) was handled. The inherited-key behavior difference is not a realistically reachable user-facing change. The changeset-bot warning is expected and can be left as-is.

### Not fully validated
I did not run the example/browser suites for combobox role rendering (`examples/combobox`, `examples/menu-combobox`, `examples/select-combobox`) — the helper output is unchanged for all real popup roles (`listbox`/`menu`/`tree`/`grid`/unset), so DOM role behavior is unaffected.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
diegohaz: Thanks, this is a fair broader deduplication point. I’m keeping this PR scoped to T026 rather than moving the lookup into `@ariakit/utils` because the assigned issue is specifically the active/offscreen combobox divergence, and the existing utility helper is DOM-element based with fallback semantics used by other packages. This PR removes the reported combobox drift path, keeps the lookup private, and adds the own-property fallback test. I’d leave a repo-wide role-string utility extraction for a separate follow-up so it can be reviewed against the other consumers together.

packages/ariakit-react-components/src/combobox/combobox-item.test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, test } from "vitest";
```
diegohaz: No need for this test file

diegohaz: Fixed in c8acc2ee4 - removed the unnecessary `combobox-item.test.ts` file and kept the PR focused on the shared helper refactor.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
diegohaz: Sounds good — scoping the `@ariakit/utils` extraction to a separate follow-up is reasonable, since it needs to be reviewed against the menu/select/composite consumers of `getPopupItemRole` together. This PR's removal of the active/offscreen drift path stands on its own. Resolving this thread on that basis.

One small note for the record: this reply mentions the PR "adds the own-property fallback test," but the latest commit (`c8acc2ee4`) removed `combobox-item.test.ts`, so that part is now stale. With the test gone, the own-property/inherited-key fallback (`constructor`/`toString` -> `option`) is no longer pinned anywhere — the realistic `listbox`/`menu` -> role paths are still covered by the combobox/menu-combobox/select-combobox example tests, but that edge case is now undocumented. Fine to leave given it's an internal helper and the sibling `getPopupItemRole` isn't unit-tested either; just flagging so the record is accurate.

diegohaz (commented):

**Incremental re-review** of head `c8acc2ee` against my prior reviewed SHA `4157cb8a`.

### What changed since last review
Only one change: commit `c8acc2ee4` ("Remove unnecessary combobox item test") deletes `combobox-item.test.ts`. The two source files (`combobox-item.tsx`, `combobox-item-offscreen.tsx`) are **byte-identical** to what I reviewed at `4157cb8a` — confirmed via `git diff 4157cb8a..HEAD` on those paths (empty). `oxlint` and `oxfmt --check` on both remaining files pass; no production-code change to re-validate (build/tsc were green at the identical source last pass).

### Prior finding status
My earlier finding — a third byte-identical `itemRoleByPopupRole` copy in `@ariakit/utils/src/dom.ts` (`getPopupItemRole`/`getPopupRole`) — is **not addressed in code** (still three copies), but the author responded on the thread with a clear, reasonable rationale: keep this PR scoped to the T026 active/offscreen drift, keep the lookup private, and extract a shared role-string utility into `@ariakit/utils` as a separate follow-up so it can be reviewed against the menu/select/composite consumers together. I agree that's the right sequencing and have acknowledged/closed the thread on that basis.

### On the test removal
Acceptable, non-blocking. The deleted file was a micro-unit test of an internal helper; its sibling `getPopupItemRole` in `@ariakit/utils` isn't unit-tested either, so this aligns with existing precedent, and the realistic `listbox`/`menu` -> item-role paths remain covered by example browser tests (`select-combobox`, `combobox-group`, `combobox-multiple`, `menu-combobox` all assert via `option`/`listbox` roles). The only coverage genuinely lost is the inherited-key edge case (`constructor`/`toString` -> `option`) introduced by the `hasOwnProperty` guard — now undocumented but practically unreachable. Noted on the thread; not worth blocking.

Also flagged a minor record-keeping inaccuracy: the author's earlier thread reply still says the PR "adds the own-property fallback test," which is now stale after the removal.

### Changeset
Still no changeset needed — internal refactor, no public `@ariakit/react` API or documented behavior change. The changeset-bot warning is expected.

### Net
No new blocking findings. The PR remains a sound, behavior-preserving refactor.

PR #6151, opened 2026-05-31
Dedupe composite and tabbable helpers

diegohaz:
## Motivation

Composite navigation helpers are duplicated between the core composite store and React composite utilities. The tabbable focus helpers also duplicate the same directional lookup logic, and the public previous-tabbable fallback parameter is named after the opposite edge.

## Solution

Export the shared composite navigation helpers from the core composite store and alias them from the React utilities so there is one implementation. Add a shared direction-aware tabbable helper, rename the public previous fallback parameter to `fallbackToLast`, update the utils README signature, and add a patch changeset for `@ariakit/utils`.

## Verification

- `pnpm lint-fix`
- `pnpm build`
- `pnpm tsc`
- `pnpm test app/src/examples/combobox-group/test.ts`
- `pnpm test`

## Pre-commit review

The usual separate native-reviewer gate was avoided after an explicit coordination update instructed this top-level thread not to create or rely on subagent/native-reviewer threads. I completed the remaining pre-commit review locally with read-only commands in this thread.

### DIFF
--- a/packages/ariakit-components/src/composite/composite-store.ts
+++ b/packages/ariakit-components/src/composite/composite-store.ts
@@ -30,6 +30,9 @@ interface NextOptions extends Pick<
 
 export const NULL_ITEM = { id: null as unknown as string };
 
+/**
+ * Finds the first enabled item.
+ */
 export function findFirstEnabledItem(
   items: CompositeStoreItem[],
   excludeId?: string,
@@ -55,6 +58,15 @@ function getItemsInRow(items: CompositeStoreItem[], rowId?: string) {
   return items.filter((item) => item.rowId === rowId);
 }
 
+/**
+ * Moves all the items before the passed `id` to the end of the array. This is
+ * useful when we want to loop through the items in the same row or column as
+ * the first items will be placed after the last items.
+ *
+ * The null item that's inserted when `shouldInsertNullItem` is set to `true`
+ * represents the composite container itself. When the active item is null, the
+ * composite container has focus.
+ */
 export function flipItems(
   items: CompositeStoreItem[],
   activeId: string,
@@ -68,6 +80,9 @@ export function flipItems(
   ];
 }
 
+/**
+ * Creates a two-dimensional array with items grouped by their rowId's.
+ */
 export function groupItemsByRows(items: CompositeStoreItem[]) {
   const rows: CompositeStoreItem[][] = [];
   for (const item of items) {
--- a/packages/ariakit-components/src/composite/composite-store.ts
+++ b/packages/ariakit-components/src/composite/composite-store.ts
@@ -28,9 +28,12 @@ interface NextOptions extends Pick<
   skip?: number;
 }
 
-const NULL_ITEM = { id: null as unknown as string };
+export const NULL_ITEM = { id: null as unknown as string };
 
-function findFirstEnabledItem(items: CompositeStoreItem[], excludeId?: string) {
+export function findFirstEnabledItem(
+  items: CompositeStoreItem[],
+  excludeId?: string,
+) {
   return items.find((item) => {
     if (excludeId) {
       return !item.disabled && item.id !== excludeId;
@@ -52,7 +55,7 @@ function getItemsInRow(items: CompositeStoreItem[], rowId?: string) {
   return items.filter((item) => item.rowId === rowId);
 }
 
-function flipItems(
+export function flipItems(
   items: CompositeStoreItem[],
   activeId: string,
   shouldInsertNullItem = false,
@@ -65,7 +68,7 @@ function flipItems(
   ];
 }
 
-function groupItemsByRows(items: CompositeStoreItem[]) {
+export function groupItemsByRows(items: CompositeStoreItem[]) {
   const rows: CompositeStoreItem[][] = [];
   for (const item of items) {
     const row = rows.find((currentRow) => currentRow[0]?.rowId === item.rowId);
--- a/.changeset/300-previous-tabbable-fallback.md
+++ b/.changeset/300-previous-tabbable-fallback.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Renamed the `getPreviousTabbable` fallback parameter to `fallbackToLast` to match its behavior.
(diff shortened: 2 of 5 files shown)


packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -52,7 +55,7 @@ function getItemsInRow(items: CompositeStoreItem[], rowId?: string) {
   return items.filter((item) => item.rowId === rowId);
 }
 
-function flipItems(
+export function flipItems(
```
diegohaz, line 70: When these helpers moved here from `ariakit-react-components/src/composite/utils.ts`, their JSDoc didn't come with them. The Core definitions of `flipItems`, `findFirstEnabledItem`, and `groupItemsByRows` are undocumented, so the explanatory comments that used to live on the React copies are now gone from the single source of truth (this is also what CodeRabbit's docstring-coverage check flagged).

The `flipItems` doc is the most worth keeping, since the null-item behavior isn't obvious from the code. Consider moving the original docs onto the Core definitions, e.g.:

```ts
/**
 * Moves all the items before the passed `id` to the end of the array. This is
 * useful when we want to loop through the items in the same row or column as
 * the first items will be placed after the last items.
 *
 * The null item that's inserted when `shouldInsertNullItem` is set to `true`
 * represents the composite container itself. When the active item is null, the
 * composite container has focus.
 */
export function flipItems(
```

Same applies to the one-line docs that were on `findFirstEnabledItem` ("Finds the first enabled item.") and `groupItemsByRows` ("Creates a two-dimensional array with items grouped by their rowId's."). Not blocking, but the dedupe shouldn't cost the only documentation these helpers had.

diegohaz (commented):
## Review summary

Initial review of head `591e58c`, diffed against merge-base `c863ea1`. This is a pure dedupe/refactor across `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/utils`.

### What I verified

- **Focus refactor is behavior-preserving.** `getTabbableInDirection` reproduces both directions exactly: `getNextTabbableIn` (no `reverse`, `fallbackToEdge: fallbackToFirst`) and `getPreviousTabbableIn` (`reverse: true`, `fallbackToEdge: fallbackToLast`) match the prior inlined logic, including the in-place `.reverse()` (which is safe since `getAllFocusableIn` returns a fresh array) and the `candidates[0]` focusable-fallback path.
- **`getPreviousTabbable` rename is non-breaking.** Renaming the first positional param `fallbackToFirst` -> `fallbackToLast` only changes the public name; the value is still forwarded as `getPreviousTabbableIn`'s `fallbackToLast`. Both internal callers (`portal.tsx`, `@ariakit/test/press.ts`) call it with no args, and the README signature is updated consistently. No stale `fallbackToFirst` references remain in the previous-tabbable path.
- **Composite dedup is sound.** The `import * as Core from "@ariakit/components/composite/composite-store"` pattern is already used by `composite-store.ts`; `@ariakit/components` is a declared dep and exposes the `./composite/composite-store` subpath. React's `CompositeStoreItem extends Core.CompositeStoreItem`, so the re-exported helpers stay type-compatible for callers in `composite.tsx` / `composite-typeahead.tsx`. The public `@ariakit/react-components/composite/utils` surface (`flipItems`, `findFirstEnabledItem`, `groupItemsByRows`) is preserved via aliases.
- **Changeset is correct.** `getPreviousTabbable` is public only via `@ariakit/utils` (`export * from "./focus.ts"`) and isn't re-exported through any higher-level package's public entrypoint, so listing only `@ariakit/utils` at `patch` is right; filename prefix and single past-tense format follow the conventions.
- **Builds/CI.** `pnpm build` passed locally; PR CI is green (Type check, Build, Lint, all Test jobs). A local `pnpm tsc` surfaced 3 errors only in unrelated `app/src/sandbox/*` files (duplicate-React-types artifacts of a fresh isolated install) — confirmed pre-existing since CI Type check passes on this same SHA and none of those files touch the changed code.

### Findings

- One non-blocking maintainability note (inline): the JSDoc on `flipItems` / `findFirstEnabledItem` / `groupItemsByRows` was dropped when the implementations moved to Core, leaving the canonical definitions undocumented (matches CodeRabbit's docstring-coverage warning). The `flipItems` null-item explanation is worth restoring on the Core definition.

### Not exhaustively validated

- No dedicated unit tests exist for `focus.ts`; the tabbable helpers are exercised indirectly through portal/focus-trap browser tests, which pass in CI. I reasoned about the refactor equivalence rather than adding a focused test.

Comment-only review; no blocking issues.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in f92c2ef5b - restored the original helper JSDoc on the Core definitions of `findFirstEnabledItem`, `flipItems`, and `groupItemsByRows`.

diegohaz:
**Fixed**
- f92c2ef5b (https://github.com/ariakit/ariakit/pull/6151#discussion_r3329996976): Restored the original JSDoc on the Core composite helper definitions after deduping the React copies.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Verified in `f92c2ef` — resolved. All three JSDoc blocks were restored on the Core definitions verbatim against the original React `composite/utils.ts` copies (confirmed byte-for-byte for `flipItems`, plus the one-liners on `findFirstEnabledItem` and `groupItemsByRows`). The commit is comment-only (+15/-0, no logic change), so no behavior or type impact. Thanks for restoring these.

diegohaz (commented):

Incremental re-review of `f92c2ef`, baselined against my last reviewed SHA `591e58c`. The only new commit is **"Restore composite helper docs"** — a single file, comment-only change (`+15/-0`) to `packages/ariakit-components/src/composite/composite-store.ts`.

### Prior finding — resolved

My one prior finding (JSDoc dropped when `flipItems`/`findFirstEnabledItem`/`groupItemsByRows` moved to Core) is **fixed**. All three JSDoc blocks were restored on the Core definitions, verified verbatim against the original React `composite/utils.ts` copies — including the full null-item explanation on `flipItems` and the one-liners on the other two. I replied on the existing thread confirming.

### Verification

- Diffed `591e58c..f92c2ef`: every added line is a comment; no logic, signature, or type change, so there is no behavioral or type risk from this commit.
- Everything I validated in the initial pass still holds (behavior-preserving focus refactor, non-breaking `getPreviousTabbable` rename, sound composite dedup, correctly-scoped `@ariakit/utils` patch changeset). Those areas were untouched by this commit.

### Status

No remaining findings. Comment-only review; nothing blocking.

PR #6152, opened 2026-05-31
Extract form item hook

diegohaz:
## Motivation

Several React form hooks repeat the same setup for resolving the form store, asserting the surrounding form context, normalizing the field name, creating an id/ref pair, and building collection items for form fields, labels, errors, and descriptions.

## Solution

Extract shared form item setup into internal helpers in `form-context.tsx`. The field, label, error, and description hooks now call `useFormItem` for the common store/name/id/ref/getItem behavior, while `FormInput` uses `useFormItemContext` for the shared store/name preamble before delegating to `useFormControl`.

The helper keeps explicit ids flowing through `useId(props.id)`, preserves the per-component invariant messages, and leaves each component's `useCollectionItem` call local to the component.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test examples/form/test.ts examples/form-select/test.ts examples/form-radio/test.ts`
- `APP_PORT=4321 NEXTJS_PORT=3001 pnpm -F app test src/sandbox/form-3607/test-browser.ts --project=chrome`

### DIFF
--- a/packages/ariakit-react-components/src/form/form-context.tsx
+++ b/packages/ariakit-react-components/src/form/form-context.tsx
@@ -1,9 +1,22 @@
-import { createStoreContext } from "@ariakit/react-utils";
+import type { StringLike } from "@ariakit/components/form/types";
+import { createStoreContext, useId } from "@ariakit/react-utils";
+import { invariant } from "@ariakit/utils";
+import { useCallback, useRef } from "react";
 import {
   CollectionContextProvider,
   CollectionScopedContextProvider,
 } from "../collection/collection-context.tsx";
-import type { FormStore } from "./form-store.ts";
+import type { CollectionItemOptions } from "../collection/collection-item.tsx";
+import type { FormStore, FormStoreItem } from "./form-store.ts";
+
+type FormItemType = FormStoreItem["type"];
+type FormItemGetItem = NonNullable<CollectionItemOptions["getItem"]>;
+
+interface FormItemContextOptions {
+  store?: FormStore;
+  name: StringLike;
+  component: string;
+}
 
 const ctx = createStoreContext<FormStore>(
   [CollectionContextProvider],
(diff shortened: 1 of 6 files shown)


diegohaz (commented):
**Initial review** of `5c8ebd34d6578736ba826f90f05173a15f7982f0` (baseline: merge-base with `main`, `c863ea104`).

### Scope

Pure internal refactor of the React form package: extracts the shared store/name/id/ref/`getItem` preamble from the form-item hooks into two helpers in `form-context.tsx` — `useFormItemContext` (store resolution + invariant + name normalization, used by `FormInput`) and `useFormItem` (adds `useId`/`useRef`/memoized `getItem`, used by `FormControl`, `FormLabel`, `FormError`, `FormDescription`). Reviewed all 6 changed files plus the surrounding context (`form-store.ts` item types, `@ariakit/react/src/form.ts` public re-exports, the `invariant` implementation, and consumers).

### Findings

No actionable findings. The refactor preserves behavior and reads cleanly:

- **Behavior parity verified.** Each helper reproduces the original logic exactly: same `String(name)` normalization, same `useId(props.id)` id flow, same `getItem` shape (`{ ...item, id: id || item.id, name, type }` with `getItemProp` delegation), and the per-component invariant messages are reconstructed verbatim via the `component` argument (`"FormControl must be wrapped in a Form component."`, etc.).
- **Dropping `store?.` optional chaining is safe.** `useFormItemContext` calls `invariant(form, ...)`, which is an `asserts condition` function and throws in production too (it throws `Error("Invariant failed")` when the message is `false`), so `form` is guaranteed non-null afterward. The non-optional `form.setError`/`form.setFieldTouched`/`form.useValidate` calls are equivalent to the old `store?.` calls.
- **Hook-order change is harmless.** In `useFormControl`, the memoized `getItem` (now inside `useFormItem`) runs before `form.useValidate`, whereas it previously ran after. Both are unconditional and the order is stable across renders, so the Rules of Hooks are respected.
- **No public API surface change.** The new `useFormItem`/`useFormItemContext` are internal to `form-context.tsx`; `@ariakit/react/src/form.ts` only re-exports the pre-existing `useFormContext`. This is also why no changeset is required (pure internals, no shipped behavior change) — correctly omitted.
- **Types/imports clean.** `FormItemType = FormStoreItem["type"]` permits all collection item types; the four call sites pass the same literals as before. No unused imports remain (`CollectionItemOptions`/`StringLike` are still used in the `*Options` interfaces).

### Validation performed (in a dedicated worktree at this head SHA)

- `pnpm tsc` — passed, 0 errors.
- `pnpm test examples/form/test.ts examples/form-select/test.ts examples/form-radio/test.ts` — 3 files, 20 tests, all passed.
- Biome/lint check on the 6 changed files — clean.

### Not fully validated

Did not run the Playwright browser sandbox (`form-3607`) or the full app build; the unit/integration tests plus the manual behavior trace cover the refactored paths, and there is no rendering-environment-specific change here.

PR #6153, opened 2026-05-31
Add keyed listener dispatch characterization

diegohaz:
## Motivation

The store keyed listener fast path iterates the live keyed listener set. Existing tests cover re-keying and removal during dispatch, but not adding a new listener for the same key while that fast path is running.

## Solution

Add a focused characterization test that registers only keyed `count` listeners, then subscribes another `count` listener during a `setState` dispatch and asserts it fires in the same dispatch.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`

### DIFF
--- a/packages/ariakit-store/src/test.ts
+++ b/packages/ariakit-store/src/test.ts
@@ -524,6 +524,24 @@ test("fires a re-keyed listener for the currently dispatched key", () => {
   expect(events).toEqual(["first", "second"]);
 });
 
+test("fires a keyed listener added during the keyed fast path", () => {
+  const store = createStore({ count: 0 });
+  const events: string[] = [];
+
+  const second = () => {
+    events.push("second");
+  };
+
+  subscribe(store, ["count"], () => {
+    events.push("first");
+    subscribe(store, ["count"], second);
+  });
+
+  store.setState("count", 1);
+
+  expect(events).toEqual(["first", "second"]);
+});
+
 test("unsubscribes a keyed listener from inside another keyed listener", () => {
   const store = createStore({ count: 0 });
   const events: string[] = [];


diegohaz (commented):
**Initial review** — comment-only. Baseline: merge-base with `main` (`c863ea1040`); the only PR commit is `5fdcc73de`, which adds one test to `packages/ariakit-store/src/test.ts`. (The branch is based on `7a52a5407`, already merged as #6146, so that admin-avatar diff is not part of this PR.)

**Scope reviewed:** the new `"fires a keyed listener added during the keyed fast path"` test, the surrounding keyed-dispatch tests, and the store's `runListeners` keyed fast path in `packages/ariakit-store/src/index.ts` (the `listenersByKey.get(updatedKey)` live-`Set` iteration at lines 336–354).

**What I verified locally (Node 24, in a dedicated worktree):**
- The new test passes against the current store source (`["first", "second"]`), and the added `count` listener fires exactly once.
- It is a meaningful, non-tautological characterization: when I patched the fast path to iterate a snapshot (`for (const listener of [...keyedListeners])`) instead of the live `Set`, the test fails (`["first"]`). So it would catch a regression where listeners added mid-dispatch stop being observed. The behavior it locks in is JS `Set` iteration visiting elements added during the same `for...of` (and `addKeyedListener` reusing the existing per-key `Set`).
- It exercises a genuinely different path from the adjacent tests: `"fires a re-keyed listener..."` registers `null`/all-keys listeners and goes through the slow path, while this one registers only `["count"]` listeners and hits the fast path. `"unsubscribes a keyed listener..."` covers removal on that path; this covers addition. No redundancy.
- `npx vitest run packages/ariakit-store/src/test.ts` → 37 passed.
- `tsc --noEmit` on the store package → clean.

**Changeset:** none needed. This is a test-only addition; `test.ts` is not part of the package's published surface (`exports` only expose `./src/index.ts`), so no shipped behavior changes.

**Findings:** none. The test matches the existing style/naming conventions, uses the real store with no mocks, and accurately characterizes current behavior. No correctness, complexity, or convention issues found.

Note: the only other "review" on this PR is an automated bot error placeholder with no content — nothing to address there.

PR #6154, opened 2026-05-31
Add slanted polygon vertex tests

diegohaz:
## Motivation

isPointInPolygon only had an axis-aligned rectangle test, which does not exercise the slanted-vertex branches used by hovercard transit polygons when pointer movement approaches diagonally.

## Solution

Add a focused test with getElementPolygon-style top, left, and bottom approach polygons. The assertions cover apex vertices, slanted edges, and horizontal rays crossing slanted vertices so both vertex-crossing toggle paths are covered. Outside points on each apex's horizontal ray also pin the `vy` extremum-suppression guard on both the `y > vy` (top) and `y < vy` (bottom) branches, so removing the guard makes the suite fail.

### DIFF
--- a/packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
+++ b/packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
@@ -67,6 +67,9 @@ test("isPointInPolygon handles horizontal rays crossing slanted vertices", () =>
   expect(isPointInPolygon([3.5, 1], topPolygon)).toBe(true);
   expect(isPointInPolygon([3, 2], topPolygon)).toBe(true);
   expect(isPointInPolygon([5, 2], topPolygon)).toBe(false);
+  // The apex [3, 0] is a local extremum, so the horizontal ray grazing it must
+  // not be counted as a crossing. Without the vy guard this point would toggle.
+  expect(isPointInPolygon([5, 0], topPolygon)).toBe(false);
 
   const leftPolygon: Polygon = [
     [0, 3],
@@ -81,6 +84,25 @@ test("isPointInPolygon handles horizontal rays crossing slanted vertices", () =>
   expect(isPointInPolygon([1, 3], leftPolygon)).toBe(true);
   expect(isPointInPolygon([1, 3.5], leftPolygon)).toBe(true);
   expect(isPointInPolygon([1, 5], leftPolygon)).toBe(false);
+
+  // Mirror of topPolygon with a downward apex, exercising the same vertex guard
+  // on the opposite (y < vy) branch.
+  const bottomPolygon: Polygon = [
+    [3, 6],
+    [2, 4],
+    [2, 2],
+    [4, 2],
+    [4, 4],
+  ];
+
+  expect(isPointInPolygon([3, 6], bottomPolygon)).toBe(true);
+  expect(isPointInPolygon([2.5, 5], bottomPolygon)).toBe(true);
+  expect(isPointInPolygon([3.5, 5], bottomPolygon)).toBe(true);
+  expect(isPointInPolygon([3, 4], bottomPolygon)).toBe(true);
+  expect(isPointInPolygon([5, 4], bottomPolygon)).toBe(false);
+  // The downward apex [3, 6] is also a local extremum, so its grazing ray must
+  // not toggle either.
+  expect(isPointInPolygon([5, 6], bottomPolygon)).toBe(false);
 });
 
 test("isPointInPolygon returns false for malformed polygons", () => {
--- a/packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
+++ b/packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
@@ -53,6 +53,36 @@ test("isPointInPolygon", () => {
   expect(isPointInPolygon([5, 5], polygon)).toBe(false);
 });
 
+test("isPointInPolygon handles horizontal rays crossing slanted vertices", () => {
+  const topPolygon: Polygon = [
+    [3, 0],
+    [2, 2],
+    [2, 4],
+    [4, 4],
+    [4, 2],
+  ];
+
+  expect(isPointInPolygon([3, 0], topPolygon)).toBe(true);
+  expect(isPointInPolygon([2.5, 1], topPolygon)).toBe(true);
+  expect(isPointInPolygon([3.5, 1], topPolygon)).toBe(true);
+  expect(isPointInPolygon([3, 2], topPolygon)).toBe(true);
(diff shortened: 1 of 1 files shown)


diegohaz:
**Already Resolved**

- changeset-bot: No changeset is needed because this PR only adds tests and does not affect shipped package behavior.

**No Action Needed**

- CodeRabbit: Review completed with no actionable comments, and GraphQL shows no unresolved inline review threads.

packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
```diff
@@ -53,6 +53,36 @@ test("isPointInPolygon", () => {
...
+    [2, 4],
+    [4, 4],
+    [4, 2],
+  ];
+
+  expect(isPointInPolygon([3, 0], topPolygon)).toBe(true);
+  expect(isPointInPolygon([2.5, 1], topPolygon)).toBe(true);
+  expect(isPointInPolygon([3.5, 1], topPolygon)).toBe(true);
+  expect(isPointInPolygon([3, 2], topPolygon)).toBe(true);
+  expect(isPointInPolygon([5, 2], topPolygon)).toBe(false);
```
diegohaz, line 69: These tests do exercise the slanted-vertex toggle paths nicely — I instrumented the function locally and confirmed `[3, 2]` hits the `y > vy` upward-vertex toggle and `[1, 3]` (on `leftPolygon`) hits the `y < vy` downward-vertex toggle. They also catch a flipped or dropped `vy` comparison.

There's one gap worth closing, though. The whole reason `vertexPoint`/`vy` exists in the algorithm is to *suppress* a toggle when a horizontal ray passes through a vertex that is a local extremum (e.g. the apex `[3, 0]` of `topPolygon`). None of the new assertions pin that suppression down: I mutation-tested the function and removing the guard so the vertex always toggles (`if (y === yj) inside = !inside;` instead of the guarded `if (y > vy)` / `if (y < vy)`) still passes every assertion in this test.

A point on the apex's horizontal ray, outside the polygon, distinguishes the two. I verified against the current implementation:

```ts
// topPolygon apex is [3, 0]; the ray y=0 must not count the apex as a crossing
expect(isPointInPolygon([5, 0], topPolygon)).toBe(false);
expect(isPointInPolygon([1, 0], topPolygon)).toBe(false);
```

With the guard intact these are `false`; with the "always toggle" regression `[5, 0]` becomes `true`. The same holds for `leftPolygon` along its left apex `[0, 3]` (e.g. `expect(isPointInPolygon([5, 3], leftPolygon)).toBe(false)` — currently `false`, would regress). Adding one such assertion per polygon makes the suite fail under the most direct regression of the `vy` guard, which is the behavior these tests are meant to characterize.

diegohaz (commented):
**Initial review** of `b1eab4f` (test-only PR). Baseline: merge-base with `main`. Scope: the single changed file `packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts`, read against the implementation in `polygon.ts`.

**What I validated locally**
- Ran the suite: 3 tests pass.
- Instrumented `isPointInPolygon` to confirm the new assertions reach the intended branches. They do: `[3, 2]` on `topPolygon` hits the upward-vertex `y > vy` toggle, `[1, 3]` on `leftPolygon` hits the downward-vertex `y < vy` toggle, and `[5, 2]` exercises both. The earlier assertions land on the `where === 0` "point on the slanted edge" path. So the test name is accurate and these are not tautological.
- Mutation-tested the vertex logic. The assertions catch a flipped `vy` comparison and a never-toggle regression on both vertex paths — good characterization.

**One finding (non-blocking), posted inline**
- The `vertexPoint`/`vy` guard exists specifically to *suppress* a toggle when the ray crosses a local-extremum vertex (the apexes `[3, 0]` / `[0, 3]`). The new assertions don't pin that suppression down: removing the guard so the vertex always toggles still passes every assertion here. A point on an apex's horizontal ray outside the polygon — e.g. `expect(isPointInPolygon([5, 0], topPolygon)).toBe(false)` (currently `false`, becomes `true` under that regression) — would close the gap. Verified against the current implementation. Details on the inline thread.

**Other checks**
- No changeset needed: test-only change, no shipped behavior or public API affected (author already noted this; agrees with the changeset guidance).
- Style/conventions: matches the existing `polygon.test.ts` patterns (`vitest` `test`/`expect`, `Polygon` typing, focused per-behavior `test` block). No code-style issues.
- No correctness, regression, performance, or compatibility concerns in the diff.

Net: solid, real-behavior tests. The only suggestion is to additionally cover the apex-ray suppression case so the suite fails under the most direct regression of the `vy` guard.

packages/ariakit-react-components/src/hovercard/utils/polygon.test.ts
diegohaz: Fixed in 52e9fd2a3.

Confirmed the gap — removing the `vy` guard so a grazed vertex always toggles passed every assertion in this test. Added `[5, 0]` on `topPolygon` (your example): it pins the `y > vy` branch and flips to `true` under that mutation.

On `leftPolygon`, though, the suggested `[5, 3]` doesn't actually pin the suppression. `[0, 3]`'s ring neighbors are at `y = 2` and `y = 4`, so it's an x-extremum, not a y-extremum, and the horizontal ray genuinely *crosses* there. I mutation-tested the whole `y = 3` ray and no point changes under "always toggle" — `[5, 3]` only catches the inverse "never toggle" mutation, which `[1, 3]` already pins. So instead I added a mirrored `bottomPolygon` (downward apex `[3, 6]`) with `[5, 6]` → `false`, which pins the symmetric `y < vy` branch.

With both in place, stripping either guard now fails the suite (`[5, 0]` for `y > vy`, `[5, 6]` for `y < vy`). For what it's worth, the pre-existing rectangle test already caught the always-toggle regression at `[5, 2]`/`[5, 4]` via its flat extrema, but the slanted-vertex test now pins both branches on its own.

diegohaz:

**Fixed**
- `52e9fd2a3` ([reply](https://github.com/ariakit/ariakit/pull/6154#discussion_r3330197930)): The slanted-vertex test exercised the toggle paths but didn't pin the `vy` extremum-suppression guard. Added `[5, 0]` on `topPolygon` to pin the `y > vy` branch, plus a mirrored `bottomPolygon` with `[5, 6]` to pin the `y < vy` branch — removing either guard now fails the suite. The suggested `leftPolygon` `[5, 3]` was intentionally not used: `[0, 3]` is an x-extremum (neighbors at `y = 2`/`y = 4`), not a y-extremum, so it doesn't catch the always-toggle regression (mutation-verified across the whole `y = 3` ray); `bottomPolygon` pins the symmetric branch instead.

PR #6155, opened 2026-05-31
Add Group label aria-labelledby tests

diegohaz:
## Motivation

The Group / GroupLabel aria-labelledby relationship was only covered implicitly by component code. The existing group example test snapshots an unlabeled group, so regressions in label registration, aria-label precedence, or label cleanup could pass unnoticed.

## Solution

Add focused React example tests for Group and GroupLabel wiring. The tests assert that Group receives aria-labelledby from GroupLabel, that an explicit aria-label suppresses aria-labelledby, and that the attribute updates when the label id changes or the label unmounts.

### DIFF
--- a/examples/group/test.ts
+++ b/examples/group/test.ts
@@ -1,4 +1,7 @@
-import { q } from "@ariakit/test";
+import { Button, Group, GroupLabel } from "@ariakit/react";
+import { click, q } from "@ariakit/test";
+import { render } from "@ariakit/test/react";
+import { createElement, useState } from "react";
 import { expect, test } from "vitest";
 
 test("markup", () => {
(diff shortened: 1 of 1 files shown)


examples/group/test.ts
```diff
@@ -28,3 +31,83 @@ test("markup", () => {
     </div>
   `);
 });
+
+test("sets aria-labelledby from group label", async () => {
+  const { unmount } = await render(
```
diegohaz, line 36: These three tests render their own ad-hoc `Group`/`GroupLabel` trees via `@ariakit/test/react`, but the `beforeEach` hook in `vitest.setup.ts` already auto-renders this example's `index.react.tsx` before every test in this file. I confirmed locally that during each new test the document contains **two** groups (the example's Bold/Italic/Underline group plus the test's own). It works only because every assertion filters by accessible name (`q.group.ensure("Formatting")`/`"Toolbar"`), so the example group is silently rendered, queried past, and torn down — pure dead weight for these tests.

This also diverges from the repo convention: no other `examples/**/test.ts` imports `react` or `@ariakit/test/react` to build ad-hoc trees; they exercise the rendered example through `q.*`. Two cleaner directions:

1. Make the tests exercise the actual example by adding a `GroupLabel` to `index.react.tsx` (the example currently has no label at all, which is also why the `aria-labelledby` wiring was previously uncovered). The behavior tests would then assert against the example group, and the `markup` snapshot would document the labelled markup.
2. If ad-hoc trees are intentional (e.g. to keep the example unlabelled), consider colocating these as component tests rather than under `examples/group/`, so they don't carry the example auto-render. 

Not blocking — the assertions are correct and well-targeted (I verified each one fails under a matching regression: dropping `labelId`, ignoring the `aria-label` precedence, and removing the `GroupLabel` unmount cleanup each break exactly the expected test). Mainly flagging the redundant double-render and the convention drift.

diegohaz (commented):
**Review summary (initial review)**

Scope: full review of the only changed file, `examples/group/test.ts` (+84/-1), diffed against the merge base `a84dec3d`. Reviewed alongside the source it characterizes (`packages/ariakit-react-components/src/group/group.tsx` and `group-label.tsx`), the test harness (`vitest.setup.ts`, `vitest.config.ts`), the `q.*` query helpers, and sibling `examples/**/test.ts` conventions.

**Verification performed (in a worktree off the PR head):**
- Ran `vitest run examples/group/test.ts` → all 4 tests pass.
- Confirmed each new test is non-tautological and catches a real regression:
  - Forcing `aria-labelledby` to `undefined` → "sets aria-labelledby…" and "updates aria-labelledby…" fail.
  - Always setting `labelId` (ignoring the `aria-label` precedence) → only "gives precedence to aria-label" fails.
  - Removing the `GroupLabel` unmount cleanup (`setLabelId(undefined)`) → only "updates aria-labelledby when group label changes" fails.
- Verified via a probe that during each new test the document holds **two** groups (auto-rendered example + ad-hoc tree); assertions disambiguate by accessible name, so they remain correct.
- Restored all experimental edits; worktree is clean.

**Findings:** one non-blocking maintainability/convention comment (inline) — the tests render ad-hoc React trees that ignore the auto-rendered example, producing a redundant second group and diverging from the repo's example-driven test convention. Suggested adding a `GroupLabel` to `index.react.tsx` (or colocating as component tests) so the tests exercise the real example.

**Correctness:** no bugs found. The assertions accurately characterize current `Group`/`GroupLabel` behavior (label registration via context, `aria-label` precedence, id update, and unmount cleanup).

**Changeset:** correctly omitted — test-only change, no shipped-code impact.

CI is green except for in-progress browser/preview jobs (unrelated to this test-only change).

PR #6156, opened 2026-05-31
Test tag delimiter helpers

diegohaz:
## Motivation

The tag delimiter helpers contain edge-case logic that is only exercised indirectly today. Direct coverage makes delimiter normalization and splitting behavior easier to preserve across future tag input changes.

## Solution

Move `getDelimiters` and `splitValueByDelimiter` into the existing tag utilities module, then import them from `TagInput`. Add focused Vitest coverage for default, null, and empty delimiters, first matching delimiter precedence, leading and trailing delimiters, delimiter-only input, no matching delimiter, capturing groups, and global regex delimiters.

## Verification

- `pnpm lint-fix`
- `pnpm test packages/ariakit-react-components/src/tag/utils.test.ts`
- `pnpm test examples/combobox-tabs/test.ts`
- `pnpm tsc`

### DIFF
--- a/packages/ariakit-react-components/src/tag/utils.test.ts
+++ b/packages/ariakit-react-components/src/tag/utils.test.ts
@@ -36,6 +36,12 @@ test.each([
     expected: [],
   },
   {
+    name: "no matching delimiter",
+    value: "hello",
+    delimiters: [","],
+    expected: [],
+  },
+  {
     name: "capturing-group regex",
     value: "one,two;three",
     delimiters: [/([,;])/],
--- a/packages/ariakit-react-components/src/tag/utils.test.ts
+++ b/packages/ariakit-react-components/src/tag/utils.test.ts
@@ -1,0 +1,55 @@
+import { expect, test } from "vitest";
+import { getDelimiters, splitValueByDelimiter } from "./utils.ts";
+
+test("getDelimiters", () => {
+  expect(getDelimiters(undefined)).toEqual(["\n", ";", ",", /\s/]);
+  expect(getDelimiters(null)).toEqual([]);
+  expect(getDelimiters("")).toEqual([]);
+  expect(getDelimiters(undefined, null)).toEqual([]);
+  expect(getDelimiters(undefined, ";")).toEqual([";"]);
+  expect(getDelimiters([",", /\s/])).toEqual([",", /\s/]);
+});
+
+test.each([
+  {
+    name: "array first-match-wins",
+    value: "one,two three",
+    delimiters: [/\s/, ","],
+    expected: ["one,two", "three"],
+  },
+  {
+    name: "leading-delimiter strip",
+    value: ",,one,two",
+    delimiters: [","],
+    expected: ["one", "two"],
+  },
+  {
+    name: "trailing delimiter",
+    value: "one,two,",
+    delimiters: [","],
+    expected: ["one", "two", ""],
+  },
+  {
+    name: "value only delimiter",
+    value: ",",
+    delimiters: [","],
+    expected: [],
+  },
+  {
+    name: "capturing-group regex",
+    value: "one,two;three",
+    delimiters: [/([,;])/],
+    expected: ["one", ",", "two", ";", "three"],
+  },
+  {
+    name: "global-flag regex",
+    value: ",one,two",
+    delimiters: [/,/g],
+    expected: ["", "one", "two"],
+  },
+] as const)(
+  "splitValueByDelimiter $name",
+  ({ value, delimiters, expected }) => {
+    expect(splitValueByDelimiter(value, [...delimiters])).toEqual(expected);
+  },
+);
--- a/packages/ariakit-react-components/src/tag/tag-input.tsx
+++ b/packages/ariakit-react-components/src/tag/tag-input.tsx
@@ -9,7 +9,6 @@ import {
 } from "@ariakit/react-utils";
 import type { Props } from "@ariakit/react-utils";
 import {
-  toArray,
   getTextboxSelection,
   setSelectionRange,
   getInputType,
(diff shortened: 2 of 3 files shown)


packages/ariakit-react-components/src/tag/utils.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    value: ",one,two",
+    delimiters: [/,/g],
+    expected: ["", "one", "two"],
+  },
+] as const)(
+  "splitValueByDelimiter $name",
+  ({ value, delimiters, expected }) => {
+    expect(splitValueByDelimiter(value, [...delimiters])).toEqual(expected);
+  },
+);
```
diegohaz, line 61: Consider adding a case for the "no delimiter matches at all" path, which is distinct from the cases here and is the one the consumer actually relies on.

The parametrized cases that expect `[]` all reduce the value to empty *after* a match (e.g. `","` with `[","]`). They don't exercise the path where the value never matches any delimiter, which returns `[]` rather than the arguably-more-intuitive `[value]`:

```ts
splitValueByDelimiter("hello", [","]); // => []  (not ["hello"])
```

This is the exact behavior `tag-input.tsx` depends on: both `onPaste` and `onChange` guard with `if (!values.length) return;`, so a regression that made this return `["hello"]` would start adding the un-split input as a tag. Since `[]` here comes from the `for` loop falling through to the final `return [];` (a different branch than the post-match empty results), a dedicated case would lock in that contract and catch that regression. Suggested addition to the `test.each` array:

```ts
  {
    name: "no matching delimiter",
    value: "hello",
    delimiters: [","],
    expected: [],
  },
```

diegohaz (commented):
## Review summary

Initial full review of the entire PR diff (merge-base `7a52a5407cc5c73d2cc80f0a6253c9293cfc5e9a` -> head `3fe8e849`), plus the surrounding tag module and the `@ariakit/react-components` / `@ariakit/react` export surfaces.

**Scope:** moves `getDelimiters` and `splitValueByDelimiter` from `tag/tag-input.tsx` into `tag/utils.ts`, re-imports them, and adds `tag/utils.test.ts`.

### What I verified locally
- **Behavior-preserving move:** the moved function bodies are byte-identical to the originals on `main`; only the types were extracted into local `Delimiter` / `TagInputDelimiter` aliases, which are structurally identical to the previous `TagInputOptions["delimiter"]`-based signature. `pnpm tsc` passes (0 errors), so all `tag-input.tsx` call sites still typecheck.
- **Tests pass:** `pnpm test packages/ariakit-react-components/src/tag/utils.test.ts` -> 7 passed.
- **Tests are non-tautological:** I ran three mutations against `utils.ts` and each was caught — disabling the leading-delimiter strip (2 failures), making `getDelimiters(undefined)` ignore the default (1 failure), and changing the no-match `continue` to `return [value]` (1 failure). The suite would fail under plausible regressions.
- **No changeset needed:** internal refactor + test-only change with no shipped behavior change.

### Findings
- One inline comment: the new tests cover the post-match empty case but not the distinct "value never matches any delimiter" path (returns `[]`, not `[value]`), which is the exact contract `tag-input.tsx`'s `if (!values.length) return;` guards rely on. Suggested a one-line `test.each` case to lock it in.

### Notes (not blocking, no inline comment)
- These helpers now also become reachable via the already-public `@ariakit/react-components/tag/utils` subpath export (which already exposed `useTouchDevice`). They are **not** reachable through `@ariakit/react` (no `tag` re-export there yet), so the public-API impact is negligible.

No correctness, performance, or compatibility concerns found. Comment-only review.

packages/ariakit-react-components/src/tag/utils.test.ts
diegohaz: Fixed in 633a57bde - added a dedicated no-matching-delimiter case that pins the fallthrough `return []` branch.

diegohaz:
**Fixed**

- 633a57bde ([reply](https://github.com/ariakit/ariakit/pull/6156#discussion_r3330018091)): Added the no-matching-delimiter test case requested in review and resolved the thread.

packages/ariakit-react-components/src/tag/utils.test.ts
diegohaz: Confirmed resolved in `633a57bde`. The added `"no matching delimiter"` case (`value: "hello"`, `delimiters: [","]`, `expected: []`) does exercise the loop-fallthrough `return [];` branch, as claimed.

I verified by mutating that fallthrough to `return [value];` and re-running the suite: this case fails with `expected [ 'hello' ] to deeply equal []`, which is exactly the regression that would feed the un-split input into `tag-input.tsx`'s `if (!values.length) return;` guards. Full suite: 8/8 passing at this head. Thanks for adding it.

diegohaz (commented):

Incremental re-review of the only new commit since my last reviewed SHA (`3fe8e849` -> `633a57bde`, "Cover unmatched tag delimiter values"). The sole change in this range is the addition of one `test.each` case in `tag/utils.test.ts`.

### Verified
- **My prior coverage-gap finding is resolved.** The new `"no matching delimiter"` case (`value: "hello"`, `delimiters: [","]`, `expected: []`) is exactly the missing path I flagged, and it genuinely exercises the loop-fallthrough `return [];` branch rather than the post-match empty result. I confirmed by mutating that fallthrough to `return [value];` and re-running: this case fails with `expected [ 'hello' ] to deeply equal []`, the precise regression that would break `tag-input.tsx`'s `if (!values.length) return;` guards. Replied on the original thread.
- **Suite passes:** `pnpm vitest run packages/ariakit-react-components/src/tag/utils.test.ts` -> 8/8.

No new code outside the test file changed, so no new findings and no need to re-run `tsc` for the production path (no production source touched in this range). No outstanding issues; the PR looks good to me. Comment-only review.

PR #6157, opened 2026-05-31
Add heading level computation sandbox tests

diegohaz:
## Motivation

The Heading level-computation paths had only shallow example coverage. T038 calls out three untested branches: standalone `Heading` defaulting to level 1, non-native heading fallback attributes, and `HeadingLevel` clamping nested levels at 6.

## Solution

Add a focused `app/src/sandbox/heading-level-computation` React sandbox, following the audit note to keep this coverage under `app/src/sandbox`. The sandbox renders a standalone heading, a `Heading` rendered as a `div` inside an explicit level context, and a deeply nested `HeadingLevel` tree.

The browser test asserts the resulting accessible heading levels, native tag names, and fallback `role="heading"`/`aria-level` attributes.

## Verification

- `pnpm lint-fix`
- `APP_PORT=5173 NEXTJS_PORT=3001 pnpm -F app test src/sandbox/heading-level-computation/test-browser.ts --project=chrome --project=firefox --project=safari`
- `pnpm tsc`

### DIFF
--- a/app/src/sandbox/heading-level-computation/index.react.tsx
+++ b/app/src/sandbox/heading-level-computation/index.react.tsx
@@ -1,0 +1,33 @@
+import * as Ariakit from "@ariakit/react";
+import type { ReactNode } from "react";
+
+interface NestedHeadingLevelsProps {
+  children: ReactNode;
+  count: number;
+}
+
+function NestedHeadingLevels({ children, count }: NestedHeadingLevelsProps) {
+  let content = children;
+
+  for (let i = 0; i < count; i += 1) {
+    content = <Ariakit.HeadingLevel>{content}</Ariakit.HeadingLevel>;
+  }
+
+  return <>{content}</>;
+}
+
+export default function Example() {
+  return (
+    <div>
+      <Ariakit.Heading>Standalone heading</Ariakit.Heading>
+
+      <Ariakit.HeadingLevel level={3}>
+        <Ariakit.Heading render={<div />}>Rendered heading</Ariakit.Heading>
+      </Ariakit.HeadingLevel>
+
+      <NestedHeadingLevels count={8}>
+        <Ariakit.Heading>Clamped heading</Ariakit.Heading>
+      </NestedHeadingLevels>
+    </div>
+  );
+}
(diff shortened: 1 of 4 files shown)


diegohaz (commented):
Reviewed the full PR (initial review, diff `main...428efa826`). No substantive findings — comment-only.

**Scope**: new `app/src/sandbox/heading-level-computation/` sandbox (`index.react.tsx`, `preview.astro`, `preview.mdx`) and its browser test (`test-browser.ts`).

**What I checked**

- **Tests characterize real current behavior.** I traced each case against the live source:
  - Standalone `Heading` → `useContext(HeadingContext) || 1` yields level 1 → native `h1`, no `aria-level`. ✓
  - `Heading render={<div />}` inside `HeadingLevel level={3}` → non-native, so `role="heading"` + `aria-level="3"`. ✓
  - 8 nested `HeadingLevel`s → `Math.min(level || ctx + 1, 6)` clamps the context to 6 → native `h6`, no `aria-level`. ✓ (`count={8}` comfortably exceeds the level-6 ceiling.)
- **Branches were genuinely untested.** The existing `examples/heading/test.ts` only snapshots a `HeadingLevel`-wrapped h1/h2 tree; it never exercises the standalone-default, non-native-fallback, or clamp branches T038 calls out.
- **Not tautological — fails under regression.** I temporarily patched the library source three ways and re-ran: `|| 1` → `|| 2` fails test 1; forcing `isNativeHeading = true` fails test 2; removing the `Math.min(…, 6)` clamp fails test 3 (the `h8` has no implicit heading role, so the `level: 6` role query finds nothing). Reverted afterward; worktree clean.
- **Cross-browser green.** `pnpm -F app test src/sandbox/heading-level-computation/test-browser.ts` passes on `chrome`, `firefox`, and `safari` (9/9).
- **Conventions.** Sandbox folder/file layout, `withFramework` + `q.heading(name, { level })` usage, and the `title: "heading-level-computation"` frontmatter all match established sandbox patterns. `NestedHeadingLevels` building the deep tree via a braced loop is reasonable and avoids 8 hand-written nested wrappers.
- **Changeset.** Not required — changes are confined to the private `app` workspace; no published-package code changed.
- **Lint/format.** `oxlint` and `oxfmt --check` clean on the new files.

No correctness, simplification, or convention concerns to raise.

PR #6158, opened 2026-05-31
Add Next.js URL mapping tests

diegohaz:
## Motivation

`getNextjsUrlFromRequest` maps preview requests across local development, workers.dev previews, production domains, and fallback hosts. The existing `nextjs.test.ts` coverage only exercised preview-id helpers, so regressions in these URL mappings could silently route Next.js previews to the wrong origin.

## Solution

Add table-driven Vitest coverage for `getNextjsUrlFromRequest` covering local hostnames and `NEXTJS_PORT`, workers.dev alias and fallback mappings, Ariakit production and legacy domains, unknown-host fallback, malformed request URLs, and path normalization. The tests also assert the localhost-specific `http://localhost` behavior while non-local mappings preserve the request protocol.

## Verification

- `pnpm lint-fix`
- `pnpm test app/src/lib/nextjs.test.ts`
- `pnpm tsc`

### DIFF
--- a/app/src/lib/nextjs.test.ts
+++ b/app/src/lib/nextjs.test.ts
@@ -51,11 +51,25 @@ test.each([
       "https://pr-123.ariakit-nextjs.workers.dev/react/previews/counter-nextjs/",
   },
   {
+    name: "preserves the request protocol for aliased workers.dev requests",
+    requestUrl: "http://pr-123.ariakit-preview.workers.dev/react/previews/",
+    path: "/react/previews/counter-nextjs/",
+    expected:
+      "http://pr-123.ariakit-nextjs.workers.dev/react/previews/counter-nextjs/",
+  },
+  {
     name: "maps bare workers.dev preview requests",
     requestUrl: "https://ariakit-preview.workers.dev/react/previews/",
     path: "/react/previews/counter-nextjs/",
     expected:
       "https://ariakit-nextjs.workers.dev/react/previews/counter-nextjs/",
+  },
+  {
+    name: "preserves the request protocol for bare workers.dev requests",
+    requestUrl: "http://ariakit-preview.workers.dev/react/previews/",
+    path: "/react/previews/counter-nextjs/",
+    expected:
+      "http://ariakit-nextjs.workers.dev/react/previews/counter-nextjs/",
   },
   {
     name: "maps ariakit.com requests to the production Next.js domain",
@@ -68,6 +82,12 @@ test.each([
     requestUrl: "https://ariakit.com/react/previews/",
     path: "/react/previews/counter-nextjs/",
     expected: "https://nextjs.ariakit.com/react/previews/counter-nextjs/",
+  },
+  {
+    name: "preserves the request protocol for ariakit.com requests",
+    requestUrl: "http://next.ariakit.com/react/previews/",
+    path: "/react/previews/counter-nextjs/",
+    expected: "http://nextjs.ariakit.com/react/previews/counter-nextjs/",
   },
   {
     name: "maps ariakit.org requests to the production Next.js domain",
@@ -86,6 +106,12 @@ test.each([
     requestUrl: "https://example.com/react/previews/",
     path: "/react/previews/counter-nextjs/",
     expected: "https://nextjs.ariakit.com/react/previews/counter-nextjs/",
+  },
+  {
+    name: "preserves the request protocol for unknown hosts",
+    requestUrl: "http://example.com/react/previews/",
+    path: "/react/previews/counter-nextjs/",
+    expected: "http://nextjs.ariakit.com/react/previews/counter-nextjs/",
   },
   {
     name: "falls back to the production Next.js domain for malformed URLs",
--- a/app/src/lib/nextjs.test.ts
+++ b/app/src/lib/nextjs.test.ts
@@ -9,7 +9,100 @@
  */
 
 import { expect, test } from "vitest";
-import { getNextjsPreviewId, isNextjsPreviewId } from "./nextjs.ts";
+import {
+  getNextjsPreviewId,
+  getNextjsUrlFromRequest,
+  isNextjsPreviewId,
+} from "./nextjs.ts";
+
+test.each([
+  {
+    name: "uses NEXTJS_PORT for localhost requests",
+    requestUrl: "https://localhost/react/previews/counter-nextjs/",
+    path: "/react/previews/counter-nextjs/",
+    env: { NEXTJS_PORT: "4322" },
+    expected: "http://localhost:4322/react/previews/counter-nextjs/",
+  },
+  {
+    name: "uses the default port for localhost requests without NEXTJS_PORT",
+    requestUrl: "http://localhost/react/previews/counter-nextjs/",
+    path: "/react/previews/counter-nextjs/",
+    expected: "http://localhost:3000/react/previews/counter-nextjs/",
+  },
+  {
+    name: "maps 127.0.0.1 requests to localhost",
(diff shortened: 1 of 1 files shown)


diegohaz:
**Already resolved**

- changeset-bot: no changeset is needed because this PR only adds test coverage and does not change shipped package behavior.
- CodeRabbit: the comment is a rate-limit notice only; there are no code-review findings or inline review threads to address.

No code changes are needed from the current PR comments.

app/src/lib/nextjs.test.ts
```diff
@@ -9,7 +9,100 @@
...
+  {
+    name: "maps 0.0.0.0 requests to localhost",
+    requestUrl: "https://0.0.0.0/react/previews/counter-nextjs/",
+    path: "/react/previews/counter-nextjs/",
+    env: { NEXTJS_PORT: "4324" },
+    expected: "http://localhost:4324/react/previews/counter-nextjs/",
+  },
+  {
+    name: "maps aliased workers.dev preview requests",
+    requestUrl: "https://pr-123.ariakit-preview.workers.dev/react/previews/",
```
diegohaz, line 48: Every non-local case in this table uses an `https://` request URL, so the protocol-preservation behavior the PR description explicitly claims to assert ("non-local mappings preserve the request protocol") isn't actually covered. The workers.dev and `ariakit.com`/`ariakit.org` branches build their result with `${protocol}//...`, but since the input protocol is always `https` and the expected output is always `https`, a regression that hardcoded the scheme would slip through.

I validated this locally: replacing all three `${protocol}//` returns in `getNextjsUrlFromRequest` with a literal `https://` still passes all 18 tests. Conversely, the localhost cases are fine because they intentionally force `http://localhost` even from an `https` request, and that override is exercised.

Worth adding one `http://`-input case on a non-local host to lock in the pass-through, e.g.:

```ts
{
  name: "preserves the request protocol for workers.dev requests",
  requestUrl: "http://pr-123.ariakit-preview.workers.dev/react/previews/",
  path: "/react/previews/counter-nextjs/",
  expected:
    "http://pr-123.ariakit-nextjs.workers.dev/react/previews/counter-nextjs/",
},
```

Not blocking — the rest of the table is solid and each case fails under the matching mutation — but this one gap is the difference between documenting the protocol contract and actually enforcing it.

diegohaz (commented):
**Review summary** (initial full review, comment-only)

Scope: `app/src/lib/nextjs.test.ts` — table-driven coverage added for `getNextjsUrlFromRequest`. Baseline: merge-base with `main` (single commit `ef3ca04`). Source file is untouched; this is a pure test PR.

**What I validated locally**
- Ran the suite: 18 tests pass (`npx vitest run app/src/lib/nextjs.test.ts`).
- Mutation-tested the new cases to confirm they aren't tautological. Each of these source mutations is caught by exactly the matching case: default port `3000`→`3001`, removing the `127.0.0.1` branch, flipping the malformed-URL catch fallback `https`→`http`, dropping the alias prefix in the workers.dev mapping, and removing path normalization. Good signal — the table characterizes real behavior and would fail under plausible regressions.
- Confirmed the localhost cases correctly pin the `http://localhost` override even from an `https` request.

**Findings**
- One non-blocking coverage gap (inline): every non-local case feeds an `https://` request URL, so the `${protocol}//` pass-through in the workers.dev / `ariakit.com` / `ariakit.org` branches is never actually exercised. Hardcoding `https://` in all three returns still passes all 18 tests. The PR description claims this protocol-preservation is asserted, so adding one `http://`-input non-local case would close the gap. See the inline comment for a concrete case.

**Other checks**
- No changeset needed: `app` is private and listed in `.changeset/config.json` `ignore`; author already noted this and changeset-bot is acknowledged. Correct.
- Style/conventions: table-driven `test.each` with a `$name` template matches the existing file's Vitest conventions; naming and structure are clean. No code-style issues.
- Did not boot the combined Next.js dev server (port-conflict-prone in worktrees) — not needed, since these are pure unit tests of synchronous URL-mapping logic with no server dependency.

No correctness, compatibility, or performance concerns. The single inline note is the only actionable item, and it's optional.

app/src/lib/nextjs.test.ts
diegohaz: Fixed in f94a04d03 - I confirmed the gap first (mutating all `${protocol}//` returns to a literal `https://` passed all 18 original tests), then added one `http://`-input case for **each** non-local branch rather than just one, so the protocol pass-through is locked in everywhere it's derived from the request:

- aliased workers.dev (`${protocol}//${alias}.…`)
- bare workers.dev (`${protocol}//${NEXTJS_WORKERS_DOMAIN}…`)
- the `ariakit.com`/`ariakit.org` production return (`${protocol}//${NEXTJS_PRODUCTION_DOMAIN}…`)
- the unknown-host fallback (same production return)

With these, that same hardcode-`https://` mutation now fails 4 tests (one per branch) instead of passing silently. The malformed-URL case keeps its hardcoded `https://` expectation since the `catch` path intentionally doesn't preserve the protocol.

diegohaz:

**Fixed**
- `f94a04d03` ([reply](https://github.com/ariakit/ariakit/pull/6158#discussion_r3330189683)): The protocol-preservation claim wasn't enforced — every non-local case used an `https://` request, so hardcoding the scheme in the `${protocol}//` returns slipped through (verified: that mutation passed all 18 original tests). Added an `http://`-input case for each non-local branch (aliased workers.dev, bare workers.dev, the `ariakit.com`/`ariakit.org` production return, and the unknown-host fallback). The same mutation now fails 4 tests, locking in pass-through across every non-local mapping rather than a single branch.

Verified with `pnpm lint-fix`, `pnpm test app/src/lib/nextjs.test.ts` (22 passing), and `pnpm tsc`.

app/src/lib/nextjs.test.ts
diegohaz: Confirmed resolved in f94a04d03 — verified independently. With the four new `http://`-input cases the suite is at 22 passing, and re-running the exact mutation from the original finding (replacing all three `${protocol}//` returns with a literal `https://`) now fails exactly 4 tests, one per non-local branch (aliased workers.dev, bare workers.dev, `ariakit.com`/`ariakit.org` production, and the unknown-host fallback). The protocol pass-through is now genuinely pinned everywhere it's derived from the request. Also agree the malformed-URL `catch` case correctly keeps its hardcoded `https://` expectation, since that path has no request protocol to preserve. Nothing further on this thread.

diegohaz (commented):
**Re-review summary** (incremental, comment-only)

Baseline: my last reviewed SHA `ef3ca04` → new head `f94a04d03` (single new commit "Test protocol pass-through for non-local Next.js URL mappings"). Only `app/src/lib/nextjs.test.ts` changed (+26 lines, 4 new cases). Source `nextjs.ts` remains untouched.

**Prior finding — resolved.** My one inline note (protocol pass-through never exercised because all non-local cases used `https://`) is fixed. The commit adds an `http://`-input case for each of the three request-derived branches plus the unknown-host fallback. Verified locally: full suite now 22 passing, and re-running the exact original mutation (all `${protocol}//` returns → literal `https://`) now fails exactly 4 tests, one per branch, instead of passing silently. The malformed-URL `catch` case correctly keeps its hardcoded `https://` expectation since that path has no request protocol. Replied on the thread with the resolution.

**New issues in this commit:** none. The added cases follow the existing table conventions, each is meaningful and regression-catching, and there is no new surrounding code to scrutinize.

**Other checks:** still no changeset needed (`app` private + in changeset `ignore`). No correctness, compatibility, performance, or style concerns. Nothing outstanding.

PR #6159, opened 2026-05-31
Guard dialog cleanup stacks

diegohaz:
## Motivation

Dialog cleanup utilities stack per-element cleanup functions for inert, aria-hidden, role, and style restoration across nested dialogs. The riskiest tree-walking and cleanup primitives lacked direct unit tests, and the first cleanup registered for a key could still run a newer stacked cleanup when invoked after another setup became current.

## Solution

Add direct jsdom tests for the dialog cleanup primitives: orchestrated LIFO cleanup, stale cleanup no-ops, per-element/key isolation, tree-walk ancestor deduping, snapshot gating, backdrop and focus-trap exclusions, and mark/disable restore behavior.

Update the initial `orchestrate` cleanup path so it checks that its cleanup is still current before running, matching the guard already used by later stacked cleanups. Add a patch changeset for the user-facing Dialog cleanup fix.

### DIFF
--- a/packages/ariakit-react-components/src/dialog/utils/orchestrate.test.ts
+++ b/packages/ariakit-react-components/src/dialog/utils/orchestrate.test.ts
@@ -1,12 +1,12 @@
 import { afterEach, expect, test } from "vitest";
-import { disableTreeOutside } from "../disable-tree.ts";
-import { isElementMarked, markTreeOutside } from "../mark-tree-outside.ts";
-import { assignStyle, setAttribute } from "../orchestrate.ts";
-import { supportsInert } from "../supports-inert.ts";
+import { disableTreeOutside } from "./disable-tree.ts";
+import { isElementMarked, markTreeOutside } from "./mark-tree-outside.ts";
+import { assignStyle, setAttribute } from "./orchestrate.ts";
+import { supportsInert } from "./supports-inert.ts";
 import {
   createWalkTreeSnapshot,
   walkTreeOutside,
-} from "../walk-tree-outside.ts";
+} from "./walk-tree-outside.ts";
 
 afterEach(() => {
   document.body.innerHTML = "";
--- a/.changeset/300-dialog-cleanup-stack.md
+++ b/.changeset/300-dialog-cleanup-stack.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) cleanup so stale nested dialog effects no longer restore page accessibility state while a newer effect is active.
--- a/packages/ariakit-react-components/src/dialog/utils/__tests__/orchestrate.test.ts
+++ b/packages/ariakit-react-components/src/dialog/utils/__tests__/orchestrate.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, expect, test } from "vitest";
 import { disableTreeOutside } from "../disable-tree.ts";
 import { isElementMarked, markTreeOutside } from "../mark-tree-outside.ts";
-import { setAttribute } from "../orchestrate.ts";
+import { assignStyle, setAttribute } from "../orchestrate.ts";
 import { supportsInert } from "../supports-inert.ts";
 import {
   createWalkTreeSnapshot,
@@ -36,10 +36,6 @@ test("orchestrate restores cleanup stacks in LIFO order", () => {
 
   expect(element.getAttribute("data-value")).toBe("three");
 
-  restoreTwo();
-
-  expect(element.getAttribute("data-value")).toBe("three");
-
   restoreThree();
 
   expect(element.getAttribute("data-value")).toBe("two");
@@ -53,7 +49,7 @@ test("orchestrate restores cleanup stacks in LIFO order", () => {
   expect(element.getAttribute("data-value")).toBe("initial");
 });
 
-test("orchestrate ignores the initial cleanup when it is no longer current", () => {
+test("orchestrate defers the initial cleanup until it is current", () => {
   const element = document.createElement("div");
   element.setAttribute("role", "original");
 
@@ -66,11 +62,48 @@ test("orchestrate ignores the initial cleanup when it is no longer current", () 
 
   restoreTwo();
 
-  expect(element.getAttribute("role")).toBe("one");
+  expect(element.getAttribute("role")).toBe("original");
 
   restoreOne();
 
   expect(element.getAttribute("role")).toBe("original");
+});
+
+test("orchestrate skips disposed entries when restoring previous cleanups", () => {
+  const element = document.createElement("div");
+  element.setAttribute("data-value", "initial");
+
+  const restoreOne = setAttribute(element, "data-value", "one");
+  const restoreTwo = setAttribute(element, "data-value", "two");
+  const restoreThree = setAttribute(element, "data-value", "three");
+
+  restoreTwo();
+
+  expect(element.getAttribute("data-value")).toBe("three");
+
+  restoreThree();
+
+  expect(element.getAttribute("data-value")).toBe("one");
+
+  restoreOne();
+
+  expect(element.getAttribute("data-value")).toBe("initial");
+});
+
+test("orchestrate defers stale style cleanups", () => {
+  const element = document.createElement("div");
+  element.style.display = "flex";
+
+  const restoreOne = assignStyle(element, { display: "block" });
+  const restoreTwo = assignStyle(element, { display: "none" });
+
+  restoreOne();
+
+  expect(element.style.display).toBe("none");
+
+  restoreTwo();
+
+  expect(element.style.display).toBe("flex");
 });
 
 test("orchestrate keeps element and key cleanup stacks independent", () => {
--- a/packages/ariakit-react-components/src/dialog/utils/__tests__/orchestrate.test.ts
+++ b/packages/ariakit-react-components/src/dialog/utils/__tests__/orchestrate.test.ts
@@ -1,0 +1,255 @@
+import { afterEach, expect, test } from "vitest";
+import { disableTreeOutside } from "../disable-tree.ts";
+import { isElementMarked, markTreeOutside } from "../mark-tree-outside.ts";
+import { orchestrate, setAttribute } from "../orchestrate.ts";
+import { supportsInert } from "../supports-inert.ts";
+import {
+  createWalkTreeSnapshot,
+  walkTreeOutside,
+} from "../walk-tree-outside.ts";
+
+afterEach(() => {
+  document.body.innerHTML = "";
+});
+
+function getElement(id: string) {
+  const element = document.getElementById(id);
+  if (!element) throw new Error(`Element not found: ${id}`);
+  return element;
+}
+
+function setAttributeWithOrchestrate(
+  element: Element,
+  attribute: string,
+  value: string,
+) {
+  return orchestrate(element, attribute, () => {
+    const previousValue = element.getAttribute(attribute);
+    element.setAttribute(attribute, value);
+    return () => {
+      if (previousValue == null) {
+        element.removeAttribute(attribute);
+      } else {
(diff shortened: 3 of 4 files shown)


diegohaz (commented):
**Review summary — initial full review**

Scope: full diff from merge-base `a84dec3` to head `845b6313` (3 files: `orchestrate.ts` fix, new `__tests__/orchestrate.test.ts`, changeset), plus the surrounding dialog cleanup machinery (`walk-tree-outside.ts`, `mark-tree-outside.ts`, `disable-tree.ts`, `is-backdrop.ts`, `is-focus-trap.ts`, `supports-inert.ts`).

**Verdict: looks good — no blocking findings.** The fix is correct, minimal, and complete.

**Correctness of the fix.** The old initial-cleanup disposer ran `elementCleanups.get(key)?.()` — i.e. *whatever* cleanup is currently registered for the key, not the one this call created — and then unconditionally `delete`d the key. So if a newer stacked cleanup had become current, firing the *first* (stale) disposer would run the newer cleanup + the original cleanup and wipe the whole stack, restoring page accessibility state while a newer dialog effect is still active. The patch captures the local `cleanup` and adds the same `isCurrent` guard already used by the stacked path, so a stale initial disposer now no-ops. The asymmetry between the two paths (initial `delete`s the key, stacked restores `prevCleanup`) is correct.

**Validation performed locally** (Node 24, jsdom):
- Ran the new suite with the fix applied: 8/8 pass.
- Reverted `orchestrate.ts` to base and reran: `orchestrate ignores the initial cleanup when it is no longer current` fails (`expected 'original' to be 'two'`), the other 7 pass — confirming it is a genuine regression test that pins exactly the fixed behavior, and the other tests are characterization coverage that holds on both versions.
- Scoped `tsc --noEmit` on `@ariakit/react-components` with the test included: clean.
- Modeled normal LIFO disposal, the stale-then-LIFO sequence, and double-invocation: all restore correctly, and double-invoking a disposer is now idempotent (a nice side benefit of the guard).

**Changeset.** Correct: `@ariakit/react-components` + `@ariakit/react` (both `patch`), matching the established Dialog-fix pattern since `@ariakit/react` re-exports `Dialog`. `300-` prefix and single-paragraph past-tense copy are appropriate, and the description accurately states the user-facing effect.

**Other notes (non-blocking):**
- CodeRabbit already flagged that the test-local `setAttributeWithOrchestrate` duplicates the exported `setAttribute`. They are behaviorally identical (same key, same restore logic), so the helper could be dropped in favor of the imported `setAttribute`. Low value and already raised — not reposting inline.
- No Solid mirror of `orchestrate.ts` exists, so no parallel fix is needed there.

Areas not exhaustively exercised: I did not run the full Playwright/browser dialog suites or the complete root Vitest run (the root config pulls in app-workspace plugins that don't resolve from a linked worktree); I validated the changed unit via the focused vitest + tsc runs above. CI covers the rest.

diegohaz:
**Fixed**

- 29595c525: Removed the duplicated test helper and now use the production `setAttribute` helper directly in the orchestrate stack tests, addressing CodeRabbit's maintainability nitpick.

No unresolved inline review threads were present.

diegohaz (commented):
**Re-review summary — incremental**

Baseline: my last reviewed SHA `845b6313` → new head `29595c5` (commit "Simplify dialog cleanup tests"). Incremental diff touches only `__tests__/orchestrate.test.ts` (+7 / -33).

**Verdict: still approve-quality. No findings.**

This commit addresses the one nitpick I called out last pass (originally from CodeRabbit): the test-local `setAttributeWithOrchestrate` helper is removed and all call sites now use the exported `setAttribute`, with the now-unused `orchestrate` import dropped. The two helpers were behaviorally identical (same `orchestrate(element, attr, …)` key, same getAttribute/setAttribute/removeAttribute restore), so this is a clean behavior-preserving simplification.

**What did NOT change:** the production fix in `orchestrate.ts` and the changeset are untouched.

**Validation performed locally** (Node 24, jsdom):
- Refactored suite on head (fixed `orchestrate.ts`): 8/8 pass.
- Reverted `orchestrate.ts` to base (unfixed) and reran the refactored suite: `orchestrate ignores the initial cleanup when it is no longer current` still fails (`expected 'original' to be 'two'`), other 7 pass — so the refactor preserved the regression coverage exactly.
- Scoped `tsc --noEmit` on `@ariakit/react-components` with the refactored test: clean (no dangling import from the removed `orchestrate` reference).

Note: `otherElement` in the "keeps element and key cleanup stacks independent" test is never appended to the document, but `setAttribute`/`orchestrate` operate on the element directly and don't require connection, so this is fine and matches the prior helper's behavior.

Same caveat as the initial pass: I did not run the full Playwright/browser dialog suites or the complete root Vitest run (the root config pulls in app-workspace plugins that don't resolve from a linked worktree); the focused vitest + tsc runs above cover the changed unit, and CI covers the rest.

diegohaz:
Fixed the CI regression from the first cleanup guard. Commit 4ad8d741d switches dialog cleanup back to deferred stack flushing, so stale disposals wait until their entry reaches the top of the stack instead of restoring DOM state while a newer dialog effect is active.

Verified locally after the fix:
- pnpm lint-fix
- pnpm test packages/ariakit-react-components/src/dialog/utils/__tests__/orchestrate.test.ts examples/dialog-nested-various/test.ts examples/dialog-hide-warning/test.ts
- pnpm tsc

diegohaz (commented):
**Re-review summary — incremental (post-approval change)**

Baseline: my last reviewed SHA `29595c5` → new head `4ad8d741d` (commit "Defer stale dialog cleanups").

**Scope note:** the raw `29595c5..4ad8d741d` diff touches ~19 files, but that is rebase noise — the merge-base advanced to `8ae6708` and those files (composite-store, combobox role helper, focus.ts, perf scripts, heading-level sandbox, store test) are byte-identical to `origin/main`. The PR's actual change vs main is still only the 3 dialog files. This commit rewrites the production logic in `orchestrate.ts`, so I reviewed it as a fresh implementation.

**Verdict: sound — no blocking findings. This rewrite is strictly more correct than the version that was approved at `29595c5`.**

**What changed.** The fix moved from the "Guard" approach (a stale disposer is a permanent no-op; first-cleanup vs stacked-cleanup handled by two separate code paths) to a "Defer" approach: a per-(element, key) stack of `{ cleanup, disposed }` entries. A disposer marks its entry `disposed` and then `flushCleanupStack` pops/runs entries from the top while they are disposed (LIFO), stopping at the first live entry. So a stale cleanup is deferred and still runs later, in order, once it reaches the top.

**Why this is better, not just different.** I compared both implementations on out-of-order disposal (Node, modeled attribute/style state):
- Normal LIFO teardown: identical in both — value returns to the original. No regression for the real nested-dialog path (React effect cleanups run child→parent, i.e. LIFO).
- Out-of-order disposal (e.g. dispose the middle entry first, then top, then bottom): the **old approved Guard version leaked** — the bottom disposer became a permanent no-op and a middle-disposed entry's cleanup was lost, so the value never returned to the original. The new Defer version always converges back to the original once all disposers fire, regardless of order. This is the behavioral delta a re-approving reviewer should be aware of, since the prior approval predates it. It is a strengthening of the fix, captured by the new `orchestrate skips disposed entries when restoring previous cleanups` test.

**Edge cases I stress-tested (all correct):** double-dispose idempotency; re-`orchestrate` after a full teardown (fresh stack); re-`orchestrate` reusing a still-live stack after a partial teardown; a stale bottom dispose followed by new entries, then teardown flushing the deferred entry — no state leak in any case. The orphaned-stack-after-delete case is safe because disposers gate on `stack.includes(entry)`.

**Validation performed locally** (Node 24, jsdom):
- New suite on head (Defer): 10/10 pass (2 new tests added: `skips disposed entries…` and `defers stale style cleanups`, the latter covering `assignStyle`).
- Ran the suite against the original buggy base `orchestrate.ts` (from merge-base `8ae6708`): exactly the 3 stale/defer tests fail (`defers the initial cleanup…`, `skips disposed entries…`, `defers stale style cleanups`), the other 7 pass — confirming the new tests genuinely pin the fix.
- Scoped `tsc --noEmit` on `@ariakit/react-components` with the new test: clean. The `if (!entry)` guard in `flushCleanupStack` is required by `noUncheckedIndexedAccess` (indexed access is `T | undefined`); it is unreachable at runtime but the idiomatic guard, so not a finding.

**Changeset:** unchanged and still accurate — "stale nested dialog effects no longer restore page accessibility state while a newer effect is active" holds for the Defer approach.

Same caveat as prior passes: I did not run the full Playwright/browser dialog suites or the complete root Vitest run (the root config pulls in app-workspace plugins that don't resolve from a linked worktree); the focused vitest + tsc runs above cover the changed unit, and CI covers the rest.

packages/ariakit-react-components/src/dialog/utils/orchestrate.test.ts
```diff
@@ -0,0 +1,262 @@
+import { afterEach, expect, test } from "vitest";
```
diegohaz, line 1: Move this file out of the `__tests__` folder and colocate it

diegohaz: Fixed in ac9860482 - moved the test out of the `__tests__` folder to colocate it next to its source (`dialog/utils/orchestrate.test.ts`) and updated its imports from `../` to `./`.

diegohaz:

**Fixed**
- `ac9860482` ([reply](https://github.com/ariakit/ariakit/pull/6159#discussion_r3330190426)): Moved `orchestrate.test.ts` out of the `dialog/utils/__tests__` folder to colocate it next to its source, matching the repo's convention (e.g. `hovercard/utils/polygon.test.ts`), and updated its imports from `../` to `./`. The `__tests__` folder is removed; tests still pass (10/10) and vitest's `**/*test.{ts,tsx}` glob picks up the new path.

diegohaz (commented):
**Re-review summary — incremental (post-approval change)**

Baseline: my last reviewed SHA `4ad8d741d` → new head `ac9860482` (commit "Colocate dialog orchestrate test").

**Verdict: no findings. Mechanical, low-risk change; the cleanup-stack fix is untouched.**

**What changed.** The test file was moved from `dialog/utils/__tests__/orchestrate.test.ts` to `dialog/utils/orchestrate.test.ts` (96% rename) and its five relative imports were updated from `../` to `./`. The test body is byte-identical from the import block onward, and **`orchestrate.ts` (the Defer-stack production logic) is unchanged from `4ad8d741d`** — confirmed via `git diff`. So no re-validation of the cleanup-stack semantics was required; last pass's findings (fails on buggy base, passes on head, converges under LIFO / out-of-order / double-dispose) still hold verbatim.

**Why this is correct:**
- It aligns with the repo convention established by the recent `c863ea1 "Colocate package tests"`: after that commit there are 0 tests left in `__tests__/` dirs and 9 colocated `*.test.ts` files across published packages (including in this same package, e.g. `tag/utils.test.ts`, `hovercard/utils/polygon.test.ts`). Leaving this test in `__tests__/` would have been the outlier.
- No risk of the test leaking into the published bundle: `ariakit-scripts/src/build.ts` excludes both `src/**/*.test.*` and `src/**/__tests__/**`, so the new colocated path is excluded exactly like the old one.

**Validation performed locally** (Node 24, jsdom):
- Ran the suite from the new colocated path: 10/10 pass (the vitest `**/*test.{ts,tsx}` glob is path-agnostic, so discovery is unaffected).
- Scoped `tsc --noEmit` on `@ariakit/react-components`: clean.

**Existing thread:** the author's own review thread requesting this move ("Move this file out of the `__tests__` folder and colocate it") is already answered by their "Fixed in ac9860482" reply. I independently verified that claimed fix is complete and correct (file moved, imports updated, tests/tsc green, build still excludes it), so no follow-up reply is needed.

Same caveat as prior passes: I did not run the full Playwright/browser dialog suites or the complete root Vitest run (root config pulls in app-workspace plugins that don't resolve from a linked worktree); the focused vitest + tsc runs cover the changed unit, and CI covers the rest.

PR #6162, opened 2026-05-31
Add Menubar composite orientation sandbox coverage

diegohaz:
## Motivation

`Menubar` has branch logic that is easy to regress without focused coverage: composite menubars should render `role="menubar"` with the correct `aria-orientation`, `orientation="both"` should omit `aria-orientation`, and `composite={false}` should omit both role and orientation for widget-merging use cases.

## Solution

Add a focused React sandbox under `app/src/sandbox/menubar-composite-orientation` that renders default, vertical, both, and `composite={false}` Menubar cases. Add a browser test that asserts the role and orientation attributes for composite Menubars and verifies the `composite={false}` case is not exposed as a menubar and has no `aria-orientation`.

## Verification

- `pnpm lint-fix`
- `APP_PORT=4321 NEXTJS_PORT=3002 pnpm -F app test app/src/sandbox/menubar-composite-orientation/test-browser.ts`
- `pnpm tsc`


diegohaz:
Current PR feedback triage:

**Already resolved**
- changeset-bot: no changeset is needed because this PR only adds sandbox/test coverage and does not change shipped package behavior.

**Invalid**
- CodeRabbit: rate-limit notice only; no actionable code review comments were posted.

**Review threads**
- No unresolved inline review threads are present.

diegohaz:
Additional PR feedback triage:

**Invalid**
- Copilot reviewer: tool error only; no actionable review comments or inline review threads were posted.

app/src/sandbox/menubar-composite-orientation/test-browser.ts
```diff
@@ -0,0 +1,34 @@
...
+    const bothMenubar = q.menubar("Both menubar");
+    await test.expect(bothMenubar).toHaveAttribute("role", "menubar");
+    await test.expect(bothMenubar).not.toHaveAttribute("aria-orientation");
+  });
+
+  test("omits menubar role and orientation when composite is false", async ({
+    page,
+    q,
+  }) => {
+    const menubar = page.getByLabel("Composite false menubar");
```
diegohaz: Optional robustness nit: `page.getByLabel("Composite false menubar")` matches accessible names by substring, not exact equality. It's correct today because none of the four labels is a superset of another, but if a future sibling case in this sandbox adds a label like `"Composite false menubar (nested)"`, this query would resolve to multiple elements and the subsequent `toHaveAttribute`/`not.toHaveAttribute` assertions would throw a strict-mode violation instead of testing what's intended. Pinning it to exact matching keeps the case isolated:

```suggestion
    const menubar = page.getByLabel("Composite false menubar", { exact: true });
```

Not a blocker — the `aria-orientation`/`role` omissions and the `q.menubar(...).toHaveCount(0)` assertion are the load-bearing checks here, and I confirmed locally that this test fails if the `composite={false}` branch starts emitting `role="menubar"`.

diegohaz (commented):
**Initial review** of head `3e243e3` (test/sandbox-only PR).

### Scope reviewed
- All 4 new files under `app/src/sandbox/menubar-composite-orientation/` (`index.react.tsx`, `preview.astro`, `preview.mdx`, `test-browser.ts`).
- Cross-checked the assertions against the actual `Menubar` source in `packages/ariakit-react-components/src/menubar/menubar.tsx` and the orientation default in `menubar-store.ts`.
- Surveyed existing menubar coverage (`examples/menubar`, `examples/menubar-navigation`) to confirm this isn't duplicative.

### What I validated locally
- Ran the new test on this branch (Astro dev server for this worktree, Chrome project): both cases pass.
- Confirmed the tests are **not tautological** by mutating the library and re-running:
  - Making `aria-orientation` always reflect `state.orientation` (dropping the `"both"`/`!composite` omission) → the `"Both menubar"` assertion fails as expected.
  - Applying the `role="menubar"` block unconditionally → the `composite={false}` case fails as expected.
  Reverted both mutations; the worktree is clean.
- `oxlint`, `prettier --check`, and `Type check` (green in CI) all pass on the new files.

### Findings
- The sandbox characterizes real, previously-uncovered behavior: the attribute branch logic (`role="menubar"`, per-orientation `aria-orientation`, the `"both"` omission, and the `composite={false}` role/orientation omission). Arrow-key navigation and focus-wrapping mentioned as related behaviors are already covered by `examples/menubar*`, so targeting the attribute branches here is the right gap to fill.
- File naming and the `preview.astro`/`preview.mdx` boilerplate match established sandbox conventions exactly (component name first, kebab-case suffix, no collision).
- No changeset needed — this is test-only with no shipped-code change.
- One **optional** robustness nit posted inline on the `composite={false}` case (`getByLabel` substring vs. exact matching). Not a blocker.

### Out of scope / not introduced by this PR
- The failing `Test Plus` CI check is unrelated: it's `website/tests/ariakit-plus.ts` ("monthly subscriber purchasing Plus") failing on Clerk `authMiddleware`/Stripe flow, with no connection to `app/src/sandbox/`.
- The Copilot review errored out with no content; CodeRabbit passed.

No correctness, complexity, or convention concerns block this PR.

app/src/sandbox/menubar-composite-orientation/test-browser.ts
diegohaz: Fixed in 58c7e059d - made the composite=false label lookup exact so the assertion stays isolated if a similarly named control is added later.

diegohaz:
**Fixed**
- 58c7e059d (https://github.com/ariakit/ariakit/pull/6162#discussion_r3330047770): made the composite=false label locator exact for the Menubar sandbox test.

diegohaz:
Additional PR feedback triage:

**Already resolved**
- Performance bot: no significant regressions were reported; the only flagged Chrome change is an improvement in an unrelated benchmark, so no code change is needed.

diegohaz:
Additional PR feedback triage:

**Already resolved**
- CodeRabbit: the latest completed review reports no actionable comments, so no code change is needed.

diegohaz:
Additional PR feedback triage:

**Already resolved**
- Performance bot: the latest updated report says no significant performance changes were detected, so no code change is needed.

app/src/sandbox/menubar-composite-orientation/test-browser.ts
diegohaz: Confirmed resolved in `58c7e059d`. The `composite={false}` lookup now uses `page.getByLabel("Composite false menubar", { exact: true })`, so the case stays isolated from any future superset label. Re-ran the suite locally against this head (Chrome project) and both tests still pass. Thanks for the quick fix.

diegohaz (commented):
**Incremental re-review** of `58c7e05` (baseline: my last reviewed SHA `3e243e3`).

### Scope of this pass
The only change since the last review is the single commit `58c7e059d` ("Make Menubar sandbox label lookup exact"), which touches one line in `app/src/sandbox/menubar-composite-orientation/test-browser.ts`:

```diff
-    const menubar = page.getByLabel("Composite false menubar");
+    const menubar = page.getByLabel("Composite false menubar", {
+      exact: true,
+    });
```

This directly addresses the optional robustness nit from my prior review — the `composite={false}` lookup is now exact, so the case stays isolated if a similarly named control is added later. The other three files are unchanged.

### Validation
- Re-ran the suite against this head (Astro dev server for this worktree, Chrome project): both tests pass.
- `prettier --check` and `oxlint` pass on the modified file; worktree clean.

### Status
The prior nit is resolved (replied on that thread to confirm). No new findings — the surrounding test logic and the other sandbox files are unchanged from the last review, where I'd already verified the assertions characterize real behavior and fail under plausible `role`/`aria-orientation` regressions. The unrelated `Test Plus` CI failure (`website/tests/ariakit-plus.ts`, Clerk/Stripe) remains out of scope for this PR.

app/src/sandbox/menubar-composite-orientation/test-browser.ts
```diff
@@ -0,0 +1,36 @@
...
+    await test.expect(bothMenubar).not.toHaveAttribute("aria-orientation");
+  });
+
+  test("omits menubar role and orientation when composite is false", async ({
+    page,
+    q,
+  }) => {
+    const menubar = page.getByLabel("Composite false menubar", {
+      exact: true,
+    });
```
diegohaz, line 30: `q` already sets `exact: true`

```suggestion
    q,
  }) => {
    const menubar = q.menubar("Composite false menubar");
```

diegohaz: `q.menubar(name)` resolves to `getByRole("menubar", { name, exact: true })` (`packages/ariakit-test/src/playwright.ts:21-22`), so it matches **by `role="menubar"`**. The `composite={false}` case deliberately renders **without** that role (it's only applied inside `if (composite)` in `menubar.tsx:79-85`), which is exactly what this test asserts on line 32 (`not.toHaveAttribute("role")`) and what the existing `q.menubar("Composite false menubar")).toHaveCount(0)` check on line 34 confirms.

So switching `const menubar` to `q.menubar("Composite false menubar")` would resolve it to 0 elements — `expect(menubar).toBeAttached()` would fail, and it would contradict the `toHaveCount(0)` assertion two lines down. `q` only exposes role/text queries (no label query), and there's no role to match here, so `page.getByLabel(..., { exact: true })` is the way to locate this label-only element; the explicit `{ exact: true }` is the isolation fix from 58c7e059d. Keeping it as-is.

diegohaz:
Triaged the unresolved review thread. No code changes needed.

**Invalid**
- [`q.menubar(...)` suggestion](https://github.com/ariakit/ariakit/pull/6162#discussion_r3330155859) ([reply](https://github.com/ariakit/ariakit/pull/6162#discussion_r3330187013)): `q.menubar(name)` resolves to `getByRole("menubar", { name, exact: true })`, so it matches by `role="menubar"`. The `composite={false}` element intentionally renders without that role (`menubar.tsx:79-85`), which is what line 32 asserts and what the existing `toHaveCount(0)` on line 34 confirms — so swapping in `q.menubar(...)` would resolve to 0 elements and break `expect(menubar).toBeAttached()`. `q` has no label query, so `page.getByLabel(..., { exact: true })` (the isolation fix from 58c7e059d) stays as-is.

PR #6163, opened 2026-05-31
Add tooltip cross-anchor sandbox coverage

diegohaz:
## Motivation

Tooltip anchors use shared active-store coordination so hovering a second independent tooltip can skip the normal show delay while another tooltip is active. T062 identified that this cross-anchor behavior was not covered by the newer app sandbox tests.

## Solution

Add a `tooltip-cross-anchor` app sandbox with two independent `TooltipProvider` pairs configured with a non-zero `showTimeout` and a short `skipTimeout`. The browser test verifies that the second tooltip appears before the show timeout while the first tooltip is active, that the first tooltip is hidden, and that after hovering away past the skip window the next hover waits again.

## Verification

- `pnpm lint-fix`
- `pnpm -F nextjs build-worker`
- `pnpm -F app build-lite`
- `APP_PORT=4465 NEXTJS_PORT=3465 pnpm -F app test-chrome src/sandbox/tooltip-cross-anchor/test-browser.ts`
- `pnpm tsc`

### DIFF
--- a/app/src/sandbox/tooltip-cross-anchor/index.react.tsx
+++ b/app/src/sandbox/tooltip-cross-anchor/index.react.tsx
@@ -1,0 +1,40 @@
+import * as Ariakit from "@ariakit/react";
+import { StrictMode } from "react";
+
+const tooltipProps = {
+  showTimeout: 500,
+  skipTimeout: 100,
+};
+
+interface DelayedTooltipProps {
+  anchor: string;
+  tooltip: string;
+}
+
+function DelayedTooltip({ anchor, tooltip }: DelayedTooltipProps) {
+  return (
+    <Ariakit.TooltipProvider {...tooltipProps}>
+      <Ariakit.TooltipAnchor render={<button type="button" />}>
+        {anchor}
+      </Ariakit.TooltipAnchor>
+      <Ariakit.Tooltip>{tooltip}</Ariakit.Tooltip>
+    </Ariakit.TooltipProvider>
+  );
+}
+
+export default function Example() {
+  return (
+    <StrictMode>
+      <div
+        style={{
+          display: "flex",
+          gap: 40,
+          padding: 80,
+        }}
+      >
+        <DelayedTooltip anchor="First anchor" tooltip="First tooltip" />
+        <DelayedTooltip anchor="Second anchor" tooltip="Second tooltip" />
+      </div>
+    </StrictMode>
+  );
+}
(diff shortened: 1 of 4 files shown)


diegohaz (commented):
**Comment-only review — no blocking findings.**

Initial full review of the diff against the merge base with `main` (`origin/main...HEAD`), expanded into the tooltip cross-anchor source it characterizes.

### Scope
- `app/src/sandbox/tooltip-cross-anchor/index.react.tsx`
- `app/src/sandbox/tooltip-cross-anchor/{preview.astro,preview.mdx}`
- `app/src/sandbox/tooltip-cross-anchor/test-browser.ts`

### What I verified
- **The test characterizes real, current behavior** and is not tautological. The shared-active-store coordination it exercises lives in `packages/ariakit-react-components/src/tooltip/tooltip-anchor.tsx` — the `showOnHover` callback calls `store.show()` immediately when another `activeStore` is set (skipping `showTimeout`), `hideStore(activeStore)` hides the previously active tooltip, and the `mounted` sync schedules `removeStore` after `skipTimeout`. The three assertions map cleanly onto those three behaviors.
- **It would fail under a plausible regression.** I locally removed the immediate-show branch (replacing `store.show(); return false` with `return true` so the second hover falls back to the delay), restarted the app dev server, and the `toBeVisible({ timeout: 250 })` assertion at line 18 failed on all engines ("element(s) not found", 250ms timeout). Restored the source afterwards; worktree is clean.
- **Not flaky.** Ran `--repeat-each=4 --retries=0` across chrome/firefox/safari (12/12 passed), plus additional runs. The `midwayToShowTimeout` (250) sits safely between `skipTimeout` (100) and `showTimeout` (500); the final `not.toBeVisible()` → `toBeVisible()` pair is load-bearing (it proves the next hover waits the full delay again) and stable here.
- **Conventions.** `preview.astro`/`preview.mdx` match existing tooltip sandboxes (`tooltip-865`, `tooltip-4894`); `StrictMode` is an accepted pattern (also used in `tooltip-4894`); the kebab-case `tooltip-...` folder name leads with the component under test per the sandbox naming guidance.
- **No changeset needed** — test-only addition under `app/src/sandbox/`, no published-package source touched.
- `pnpm tsc`, `pnpm lint` (oxlint + oxfmt --check), and the new browser test all pass locally (Node 24, app dev server on a worktree port).

### Notes
- Both automated reviewers came up empty on this head: Copilot's review errored out ("Copilot encountered an error and was unable to review"), and CodeRabbit hit a rate limit, so there were no prior substantive threads to reconcile.

No actionable issues. Nice, focused coverage of a previously untested cross-anchor path.

PR #6167, opened 2026-05-31
Add ToolbarContainer focus sandbox

diegohaz:
## Motivation

T060 identified missing sandbox coverage for `ToolbarContainer` focus behavior and vertical `Toolbar` orientation. The audit note asked for this coverage under `app/src/sandbox`, so this PR adds a focused sandbox instead of changing the public examples.

## Solution

Add a React sandbox at `app/src/sandbox/toolbar-container-focus` that exercises a `ToolbarContainer` with an inner input, a `ToolbarItem` rendered as a text input, and a vertical toolbar. The browser tests cover the container staying as the tab stop until opened, Enter/Escape returning focus to the container, text-input arrow navigation only roving at caret boundaries, and vertical orientation using Up/Down while ignoring Left/Right.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- `APP_PORT=4872 NEXTJS_PORT=3872 pnpm -F app test --project=chrome --project=firefox --project=safari src/sandbox/toolbar-container-focus/test-browser.ts`

### DIFF
--- a/app/src/sandbox/toolbar-container-focus/test-browser.ts
+++ b/app/src/sandbox/toolbar-container-focus/test-browser.ts
@@ -38,9 +38,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    const container = page.locator(
-      "[data-composite-container][aria-label='Font family']",
-    );
+    const container = q.group("Font family");
     const input = q.textbox("Font family");
 
     await q.button("Before toolbar").focus();
@@ -66,9 +64,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    const container = page.locator(
-      "[data-composite-container][aria-label='Font family']",
-    );
+    const container = q.group("Font family");
     const input = q.textbox("Text size");
     const apply = q.button("Apply");
 
--- a/app/src/sandbox/toolbar-container-focus/test-browser.ts
+++ b/app/src/sandbox/toolbar-container-focus/test-browser.ts
@@ -1,0 +1,125 @@
+import type { Locator } from "@playwright/test";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+async function setSelection(input: Locator, start: number, end = start) {
+  await input.evaluate(
+    (element, range) => {
+      if (!(element instanceof HTMLInputElement)) {
+        throw new Error("Expected an input element");
+      }
+      const input = element;
+      input.setSelectionRange(range.start, range.end);
+    },
+    { start, end },
+  );
+}
+
+async function setSelectionToEnd(input: Locator) {
+  await input.evaluate((element) => {
+    if (!(element instanceof HTMLInputElement)) {
+      throw new Error("Expected an input element");
+    }
+    const input = element;
+    input.setSelectionRange(input.value.length, input.value.length);
+  });
+}
+
+function getSelectionStart(input: Locator) {
+  return input.evaluate((element) => {
+    if (!(element instanceof HTMLInputElement)) {
+      throw new Error("Expected an input element");
+    }
+    return element.selectionStart;
+  });
+}
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("ToolbarContainer keeps its inner input out of the tab order until opened", async ({
+    page,
+    q,
+  }) => {
+    const container = page.locator(
+      "[data-composite-container][aria-label='Font family']",
+    );
+    const input = q.textbox("Font family");
+
+    await q.button("Before toolbar").focus();
+    await page.keyboard.press("Tab");
+
+    await test.expect(container).toBeFocused();
+    await test.expect(input).not.toBeFocused();
+
+    await page.keyboard.press("a");
+    await test.expect(input).toBeFocused();
+
+    await page.keyboard.press("Escape");
+    await test.expect(container).toBeFocused();
+
+    await page.keyboard.press("Enter");
+    await test.expect(input).toBeFocused();
+
+    await page.keyboard.press("Enter");
+    await test.expect(container).toBeFocused();
+  });
+
+  test("ToolbarItem rendered as input navigates only at text boundaries", async ({
+    page,
+    q,
+  }) => {
+    const container = page.locator(
+      "[data-composite-container][aria-label='Font family']",
+    );
+    const input = q.textbox("Text size");
+    const apply = q.button("Apply");
+
+    await input.focus();
+    await setSelection(input, 1);
+
+    await page.keyboard.press("ArrowRight");
+
+    await test.expect(input).toBeFocused();
+    await test.expect.poll(() => getSelectionStart(input)).toBe(2);
+
+    await setSelectionToEnd(input);
+    await page.keyboard.press("ArrowRight");
+
+    await test.expect(apply).toBeFocused();
+
+    await input.focus();
+    await setSelection(input, 1);
+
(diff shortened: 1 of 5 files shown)


diegohaz (commented):
**Initial review** of head `48edb935` (full diff from merge-base `8ae67084`). Scope: the new `app/src/sandbox/toolbar-container-focus` sandbox (5 files: `index.react.tsx`, `preview.astro`, `preview.mdx`, `style.css`, `test-browser.ts`).

No substantive findings. This is a clean, well-targeted test/sandbox PR and I have nothing actionable to request changes on.

**What I validated locally (worktree on this head):**
- Ran the browser tests on all three projects — `chrome`, `firefox`, `safari` — all 3 tests pass on each, matching the PR's testing claim.
- Confirmed the tests are regression-sensitive, not tautological: mutating the sandbox to drop `ToolbarContainer` (plain `ToolbarItem` div) fails the "keeps its inner input out of the tab order" test, and removing `orientation="vertical"` fails the vertical-arrows test. The text-boundary roving test correctly stays green under those mutations because it exercises the `ToolbarItem`-as-input behavior independently.
- `pnpm tsc` (app project) passes; `oxfmt --check` and `oxlint` pass on the new files.
- Confirmed `ToolbarContainer` had **zero** prior example or test coverage in the repo, so this genuinely fills the T060 gap rather than duplicating existing coverage.

**Things I checked and deliberately did not flag:**
- The repeated `instanceof HTMLInputElement` guard across the three selection helpers looks like avoidable duplication, but it is necessary: each guard lives inside a `page.evaluate()` callback that is serialized and runs in the browser, so it cannot close over a shared module-level helper. I prototyped extracting an `asInput()` helper and it failed at runtime (`asInput is not defined` in page context), so the inlined guards are correct as written.
- No changeset is needed — this only adds a sandbox and browser tests under `app/`, with no change to any published package.
- The dev-server `Invalid hook call` SSR log under Vite SSR dep-optimization is a benign local artifact; the preview hydrates and all tests pass.

**Not fully validated:** mobile projects (`ios`/`android`) — the test file is `test-browser.ts`, which only matches the desktop browser projects, so this is expected and correct.

app/src/sandbox/toolbar-container-focus/test-browser.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("ToolbarContainer keeps its inner input out of the tab order until opened", async ({
+    page,
+    q,
+  }) => {
+    const container = page.locator(
+      "[data-composite-container][aria-label='Font family']",
+    );
```
diegohaz: I think we should use `q.label("Font family")` here. I'm not sure if it's available. Or maybe `q.text`.

diegohaz: This is not an issue because neither alternative can target this element:

- `q.label` isn't part of the Playwright `query` helper (`@ariakit/test/playwright`). It only exposes the ARIA-role queries (`q.button`, `q.textbox`, etc.) plus `q.text` — there is no `label`/`labeled` method there, so `q.label(...)` would throw.
- `q.text("Font family")` resolves to the inner `<span>Font family</span>`, not the `ToolbarContainer`. The container renders a role-less `<div>` (just `data-composite-container` + `aria-label`), so `getByRole` can't reach it by name either, and the `<span>` isn't the element that receives focus — the `toBeFocused()` assertions on `container` would fail.

I verified this against the actual rendered DOM: `getByText("Font family", { exact: true })` matches the `<span>`, while `[data-composite-container][aria-label='Font family']` uniquely matches the focusable container. CSS locators are already used elsewhere for exactly this reason (e.g. `select-combobox-3837` uses `page.locator(".popover")`), so I'm keeping the selector as the unambiguous handle.

diegohaz:
**Invalid**

- ([discussion](https://github.com/ariakit/ariakit/pull/6167#discussion_r3330170931)): `container` query suggestion (`q.label`/`q.text`) does not apply — `q.label` is not part of the Playwright `query` helper, and `q.text("Font family")` matches the inner `<span>` rather than the role-less, focusable `ToolbarContainer` div, breaking the `toBeFocused()` assertions. Verified against the rendered DOM; keeping the `[data-composite-container][aria-label='Font family']` selector as the unambiguous handle.

app/src/sandbox/toolbar-container-focus/test-browser.ts
diegohaz: Update — went with your suggestion to give the container a role. Fixed in 63a483892: added `role="group"` to the `ToolbarContainer` and switched both `container` locators to `q.group("Font family")`.

This supersedes my earlier reply about keeping the CSS selector. `q.group` works because `ToolbarContainer` renders a role-less `<div>` (no role is injected by `useCompositeContainer`/`useToolbarItem`), so the explicit `role="group"` both gives the `aria-label` a host role to attach to and makes the container reachable via `getByRole("group", { name })`. It can't collide with `q.textbox("Font family")` since the roles differ, and it matches the existing `q.group` precedent in the `aria-labelledby-aria-label` sandbox. All 3 chrome tests pass.

diegohaz:
**Fixed** (supersedes my earlier "Invalid" summary)

- `63a483892` ([discussion](https://github.com/ariakit/ariakit/pull/6167#discussion_r3330170931)): gave the `ToolbarContainer` an explicit `role="group"` and switched both `container` locators from the raw `[data-composite-container][aria-label='Font family']` selector to `q.group("Font family")`. The container renders a role-less `<div>`, so the explicit role both fixes the orphaned `aria-label` semantics and makes it reachable by role; it can't collide with `q.textbox("Font family")` (distinct roles) and matches the existing `q.group` precedent in the `aria-labelledby-aria-label` sandbox. All 3 chrome tests pass; oxlint/oxfmt clean. Pre-commit review ran (native Claude subagent + Cursor CLI; Codex CLI was unavailable due to a usage limit) and returned no actionable findings.

diegohaz (commented):
**Incremental re-review** of head `63a48389`, baselined against my previously reviewed `48edb935` (initial pass had 0 findings).

The only change since then is the locator cleanup in commit `63a48389`: `role="group"` was added to the `ToolbarContainer` in `index.react.tsx`, and the two `page.locator("[data-composite-container][aria-label='Font family']")` calls in `test-browser.ts` were replaced with `q.group("Font family")`. This resolves the locator discussion already on the thread.

**No new findings.** This is a clean, net-simplifying change and I have nothing actionable to request.

**What I validated locally (worktree on this head):**
- Re-ran the browser tests on `chrome`, `firefox`, and `safari` — all 3 tests pass on each (9/9).
- Confirmed in the rendered DOM that `role="group"` is actually applied to the container and is not overridden by Ariakit, and that there is exactly one `role="group"` element on the page carrying `aria-label="Font family"`, so `q.group("Font family")` resolves unambiguously and cannot collide with `q.textbox("Font family")` (different roles).
- Re-confirmed regression sensitivity is preserved by the query change: mutating the sandbox to drop the `ToolbarContainer` focus-management while *keeping* `role="group"` on the element still fails the "keeps its inner input out of the tab order" test, so the role-based query does not mask the behavior under test.
- `pnpm tsc` (app project) passes; `oxfmt --check` and `oxlint` pass on the changed files.
- Verified the author's stated rationale: `q.group` is an existing precedent in the repo (`app/src/sandbox/aria-labelledby-aria-label/test-browser.ts`), so this is consistent with established conventions.

No changeset is needed — still sandbox + browser tests only, with no published-package change.

PR #6172, opened 2026-06-01
Limit React 18 Vitest suite

diegohaz:
## Motivation

The React 18 Vitest job currently runs the root Vitest suite through the React 18 dependency wrapper. That reruns tests for unrelated packages and app utilities that do not depend on React compatibility, increasing CI time and noise.

## Solution

Add a first-class `test-react` script that selects a React-specific Vitest suite through `ARIAKIT_TEST_SUITE=react`, and make `test-react-18` delegate to it inside the existing React 18 workspace wrapper.

The default Vitest suite keeps the existing include patterns. `vitest.config.ts` now uses camel-case constants throughout the file while defining the suite include map.

The React 18 wrapper also maps direct `ariakit react18 test-react-18` usage to `test-react`, so both CI and manual wrapper entry points preserve the narrowed suite.

## Verification

- `pnpm test-react list` -> 722 listed test cases
- `pnpm test list` -> 957 listed test cases
- `pnpm test-react`
- `pnpm react18 test-react-18 list` -> maps to `test-react` and lists 722 React-suite tests
- `pnpm test-react-18 --shard 1/2`
- `pnpm lint-fix`
- `pnpm tsc`


diegohaz (commented):
**Initial review** — diff from merge base `506497b17` to head `53b0d2220`. Reviewed the three changed files plus surrounding context: the React 18 CI job (`.github/workflows/main.yml`), the full repo test-file inventory, and `.changeset/config.json`.

**Status: looks good — no blocking findings.** The narrowing is correct and the wrapper/CI wiring is sound.

### What I verified

**Suite narrowing is a strict, correct subset.** Installed deps and ran `vitest list --filesOnly` for both suites:

- `ARIAKIT_TEST_SUITE=all` → **172** test files
- `ARIAKIT_TEST_SUITE=react` → **145** test files
- `react` files − `all` files = **∅** (the react suite never selects anything outside `all`)
- The **27** files dropped from the React 18 run are exactly the framework-agnostic / tooling tests: `packages/ariakit-utils/**`, `packages/ariakit-store/**`, `packages/ariakit-components/**/*-store`, `packages/ariakit-scripts/**`, the non-`.react.` `packages/ariakit-test/src/__utils`, `app/src/lib/**`, and `.changeset/get-changelog-entry`. I checked their imports — none import `react` / `react-dom` / `@ariakit/react` or render components (the React-looking strings in `app/src/lib/*.test.ts` are source-transform / tokenizer / stackblitz *fixtures*, not live renders). So running them under React 18 was pure duplicate work, and dropping them loses no React-version coverage. All React-rendering tests (`ariakit-react*`, `examples/**`, `app/src/{examples,sandbox}/**`, and `ariakit-test/*.test.react.tsx`) stay in the suite.

**Wrapper recursion mapping is correct.** Splitting the guard so `test-react-18` → `test-react` (instead of `test`) makes both entry points converge on the narrowed suite: CI's `pnpm test-react-18 --shard …` → `pnpm react18 test-react` → `pnpm run test-react`, and the manual `ariakit react18 test-react-18` → `test-react` via the new guard. `--shard` forwards through to vitest on both paths, and the updated comments match the new behavior.

**Typing improvement.** `pluginsByLoader: Record<AllowedTestLoader, Array<PluginOption>>` (non-optional, keyed by the validated loader) is a real improvement over the old `Record<string, … | undefined>`, and the added brace on the loader-validation `throw` aligns with the repo brace style.

**No changeset needed.** `@ariakit/scripts` is `private: true` and listed in `.changeset/config.json`'s `ignore`; this is test/CI tooling only. The changeset-bot warning is informational.

### Considered and accepted (non-blocking)

- The `examples/**` and `app/src/{examples,sandbox}/**` globs use exact `test.{ts,tsx}` (no `*` prefix), unlike the sibling `packages/ariakit-react*` glob and the `all` suite, which use `*test`. This is deliberate — every example/sandbox test is named exactly `test.ts` (140/140) — so it currently resolves to the same set. The only latent risk is that a future example test *not* named `test.ts` would run in `pnpm test` but silently skip `test-react-18`. Given the strict convention I wouldn't change it; noting it only in case you'd prefer the example globs mirror `*test` for symmetry.
- The currently-empty `.react.` lines for the react packages / examples / app are justified future-proofing (they mirror the `all` suite's per-loader pattern so any future `.react.` example test is auto-included), so I'd keep them.

### Not fully validated

I traced the `react18` workspace-sync path by reasoning rather than re-running the full React 18 install locally, and relied on your reported `pnpm react18 test-react-18 list` → 722 for the wrapper path. The file-set diff above was run directly against the default React 19 install.

PR #6174, opened 2026-06-01
Fix Stripe checkout fulfillment retryability

diegohaz:
## Motivation

Audit finding T088 reports that Stripe checkout fulfillment recorded a checkout session as processed before the entitlement work succeeded. If Clerk team creation or Plus metadata updates failed, future webhook retries and success-page redirects saw the processed marker and skipped fulfillment, leaving the purchase ungranted.

The checkout return page and Stripe webhook can also reach fulfillment at the same time. For team purchases, that meant both requests could pass the processed check before either wrote the marker and create duplicate organizations for one checkout.

## Solution

Move the processed marker write until after fulfillment completes successfully, so failed Plus grants or team creation remain retryable.

Team checkout fulfillment now passes the checkout session ID into `createTeam`. Checkout-created teams store that session ID in organization private metadata, use a deterministic Clerk organization slug derived from the checkout session, and reuse the matching organization on retries or duplicate-slug races. This keeps same-session retries idempotent while still allowing a later checkout to create an additional team license.

For credited team upgrades, `processCheckout` now creates or reuses the team before removing the user’s personal Plus credit, so a team-creation failure does not remove the user’s existing access. The checkout fulfillment ordering is documented in code because provider-specific unit coverage would overfit the current Clerk implementation.

## Tests

- `pnpm lint-fix`
- `pnpm tsc`

## Workaround

Before this fix is released, wrap `processCheckout` and remove the premature EVENTS KV marker when fulfillment throws. This keeps the webhook or success-page retry path able to re-run fulfillment.

```ts
try {
  await processCheckout({ context, session });
} catch (error) {
  const sessionId = typeof session === "string" ? session : session.id;
  await getEventsStore().delete(sessionId);
  throw error;
}
```


app/src/lib/stripe.ts
```diff
@@ -376,12 +376,11 @@ export async function processCheckout({
     logger.info("Checkout session already processed", session.id);
     return session;
   }
-  await processEvent(session.id);
   if (type === "team") {
+    await createTeam({ context, user: clerkId, checkoutSession: session.id });
```
diegohaz, line 382: **Unit coverage for the retry/idempotency paths shouldn't be dropped — e2e can't substitute for it.**

This block is the heart of the fix: fulfillment is reordered, `processEvent` now runs only after success, and team creation gains slug/metadata idempotency plus a duplicate-slug recovery path. After the final commit removed `app/src/lib/stripe.test.ts`, none of that has automated coverage, and the PR defers it to follow-up e2e browser tests.

The problem is that e2e browser tests can't realistically exercise the exact behaviors this PR exists to guarantee, because they're driven by backend failures and request timing a browser flow can't deterministically produce:

- `createTeam` succeeds, then `removePlusFromUser` throws → the user must keep access and the event must stay retryable.
- The webhook and the success-page redirect reach `processCheckout` concurrently → both pass the `isEventProcessed` check, and the duplicate-slug `catch` in `createTeam` must converge on one org instead of creating two.
- `createOrganization` throws mid-fulfillment → `processEvent` must not be written, so Stripe/the success page can re-run.

I restored the removed suite at this exact implementation and ran it locally: **8/8 pass in 1.24s** under the existing root Vitest suite (same pattern as the other `app/src/lib/*.test.ts` files). It covers precisely these cases — e.g. *"keeps failed team checkout fulfillment retryable"*, *"keeps personal Plus until credited team checkout creates a team"*, *"reuses a concurrently-created team for the same checkout"*, *"reuses a checkout-created team when credit cleanup retries"*, and *"creates a new team for a different checkout"*.

I understand the deferral is a deliberate call. The actionable point is that this is fast, deterministic, CI-level coverage of partial-failure and concurrency on a money path that e2e tests are poorly suited to reproduce — so I'd keep this suite (or an equivalent subset of the failure/concurrency cases) rather than rely on e2e as a replacement. If the deferral stands, it's worth tracking so the gap on the checkout-fulfillment path stays visible.

app/src/lib/auth.ts
```diff
@@ -152,10 +152,20 @@ export async function setCustomer(
...
 }
 
+async function getCheckoutTeamSlug(checkoutSession: string) {
+  const data = new TextEncoder().encode(checkoutSession);
+  const hash = await crypto.subtle.digest("SHA-256", data);
+  const hex = Array.from(new Uint8Array(hash), (byte) => {
+    return byte.toString(16).padStart(2, "0");
+  }).join("");
+  return `checkout-${hex.slice(0, 32)}`;
+}
```
diegohaz, line 170: **Document why the slug is derived deterministically — it's the idempotency key, not cosmetic.**

This helper's name says *what* it does (derive a slug) but hides *why*, and the why is load-bearing. Because Clerk enforces globally-unique organization slugs, hashing the checkout session into the slug turns "create the team for this session" into a content-addressable, effectively atomic operation: concurrent or retried fulfillment either creates the single org or collides on the slug and is recovered by the `getOrganization({ slug })` branch in `createTeam`. The `stripeCheckoutSessionId` metadata then confirms a slug match really belongs to this session before reuse.

The risk is that this is invisible to a future reader. The derived slug surfaces as an opaque `checkout-<hash>` in `admin/teams.astro`, so someone could reasonably "tidy it up" by dropping `slug` and letting Clerk auto-generate from the name — which would silently remove the concurrency/duplicate guard this PR adds. A one-line rationale prevents that:

```ts
// Derive a deterministic slug from the checkout session so it doubles as an
// idempotency key: Clerk's unique-slug constraint makes concurrent/retried
// team creation for the same session converge on one organization.
async function getCheckoutTeamSlug(checkoutSession: string) {
```

A short note at the `createOrganization({ slug, ... })` call site (or the catch branch) would reinforce it too. Minor: the 32-hex (128-bit) truncation is fine for collision safety but looks arbitrary — worth a word on why it's truncated at all, since the full digest would also fit Clerk's slug length limit.

diegohaz (commented):
## Review summary

**Type:** Initial review (no prior review from this workflow). **Baseline:** merge-base `1035b91` → head `0c3d66a` (`app/env.d.ts`, `app/src/lib/auth.ts`, `app/src/lib/stripe.ts`).

### Scope reviewed
- The full `processCheckout` reorder and both call sites (`api/stripe-webhook.ts`, `plus/checkout/[...checkout].astro`), including the webhook's error propagation back to Stripe.
- `createTeam` idempotency: the deterministic-slug derivation, the membership pre-check, and the duplicate-slug `catch` recovery, against the Clerk backend SDK types (`getOrganization({ slug })`, `createOrganization({ slug, privateMetadata })`, `Organization.privateMetadata` typed as the global `OrganizationPrivateMetadata`).
- The `OrganizationPrivateMetadata` ambient declaration in `env.d.ts` and how `app` relates to changesets.

### Assessment
The core fix is sound. Moving `processEvent` after fulfillment makes failures retryable, and the webhook has no surrounding `try/catch`, so a throw yields a 500 and Stripe retries — the previously-lost purchase is now recoverable. The deterministic-slug-as-idempotency-key approach is a good fit given Clerk has no query-by-metadata: it leverages the unique-slug constraint so concurrent webhook/success-page fulfillment converges on one org, with the metadata check guarding against an unrelated slug match. The credited-team reorder (create team before removing personal Plus) correctly avoids a window where the buyer loses all access, and team access derives from membership so clearing `publicMetadata.plus` doesn't strip it. The membership pre-check + catch-by-slug are complementary (clean sequential-retry path vs. concurrency/>100-membership edge), not redundant.

### Validation performed
- `pnpm tsc` — clean (0 errors).
- Restored the removed `app/src/lib/stripe.test.ts` at this implementation and ran it: **8/8 pass in 1.24s**, then removed it (working tree left clean). This confirms the dropped suite is fast, deterministic coverage of the retry/concurrency paths.

### Findings (2 inline)
1. **`app/src/lib/stripe.ts` — test coverage.** The removed unit suite covered exactly the partial-failure and concurrency behaviors this PR introduces, which e2e browser tests can't deterministically reproduce. Recommend keeping that coverage (or an equivalent subset) rather than treating e2e as a substitute.
2. **`app/src/lib/auth.ts` — document the slug mechanism.** The deterministic slug is the idempotency key, but that's invisible from the name; a one-line rationale guards against a future refactor silently removing the concurrency guard.

### Notes (non-blocking)
- No changeset is required: `app` is in `.changeset/config.json`'s `ignore` list, so removing the changeset is correct and the changeset-bot warning is expected.
- The derived `checkout-<hash>` slug appears in `admin/teams.astro`; cosmetic and admin-only, not worth changing.

No correctness or backward-compatibility issues found. `createTeam`'s only caller is `processCheckout`, and the new `checkoutSession` param is optional, so no other call sites are affected.

app/src/lib/stripe.ts
diegohaz: Fixed in 9be8c8824 - restored app/src/lib/stripe.test.ts with deterministic coverage for the checkout retry, idempotency, credited-team, and concurrent fulfillment paths; local `pnpm vitest app/src/lib/stripe.test.ts` passes 8/8.

app/src/lib/auth.ts
diegohaz: Fixed in 9be8c8824 - added the deterministic slug rationale above `getCheckoutTeamSlug` and kept the duplicate-slug fallback explicit so retries and concurrent fulfillment converge on the checkout-created organization.

diegohaz:
**Fixed**

- 9be8c8824 (https://github.com/ariakit/ariakit/pull/6174#discussion_r3333355475): Restored the focused Stripe checkout fulfillment Vitest suite covering retryability, idempotent team creation, credited-team cleanup ordering, and concurrent checkout fulfillment.
- 9be8c8824 (https://github.com/ariakit/ariakit/pull/6174#discussion_r3333355651): Documented why the checkout-derived Clerk slug is deterministic and tightened the duplicate-slug fallback branch.

diegohaz:
**Update**

- e89a597c5: Removed the restored `app/src/lib/stripe.test.ts` suite per maintainer direction because it overfit Clerk implementation details, and documented the checkout fulfillment ordering directly in `processCheckout` instead. Local verification: `pnpm lint-fix` and `pnpm tsc`.

PR #6175, opened 2026-06-01
Fix MenuItemRadio defaultChecked prop leak

diegohaz:
## Motivation

`MenuItemRadio` reads `defaultChecked` to initialize menu values, but previously left the prop in the rest props that continue through `useRadio` and the final render path. Consumers passing `defaultChecked` directly could therefore leak that component option into rendered props.

## Solution

`useMenuItemRadio` now destructures `defaultChecked` before collecting the remaining props and passes that local value to `useInitialValue`. This mirrors `MenuItemCheckbox` and prevents `defaultChecked` from continuing into `useRadio` or composed render elements while preserving the existing default-checked behavior.

The PR also adds a focused `app/src/sandbox/menu-radio-default-checked/` Vitest regression test that renders `MenuItemRadio` with `defaultChecked` through a probe component. The test verifies the item is checked and that the probe does not receive a leaked `defaultChecked` prop.

## Workaround

Until the library fix is available, initialize the menu value through the provider/store defaults instead of passing `defaultChecked` to `MenuItemRadio` directly.

```tsx
const defaultValues = {
  // TODO: Use MenuItemRadio defaultChecked once this fix is available.
  density: "compact",
};

<MenuProvider defaultValues={defaultValues}>
  <MenuButton>Preferences</MenuButton>
  <Menu>
    <MenuItemRadio name="density" value="compact">
      Compact
    </MenuItemRadio>
  </Menu>
</MenuProvider>
```


app/src/sandbox/menu-radio-default-checked/test-browser.ts
```diff
@@ -0,0 +1,11 @@
+import { withFramework } from "#app/test-utils/preview.ts";
```
diegohaz: Can we use just `test.ts` (vitest tests with @ariakit/test) instead of playwright tests?

diegohaz: Fixed in 1ed5eb5f9 - replaced the Playwright sandbox regression with a JSDOM Vitest test.ts using @ariakit/test.

diegohaz:
**Fixed**

- 1ed5eb5f9 (https://github.com/ariakit/ariakit/pull/6175#discussion_r3332953247): Replaced the Playwright sandbox regression with a JSDOM Vitest `test.ts` using `@ariakit/test`.

PR #6176, opened 2026-06-01
Fix stale async popover positioning

diegohaz:
## Motivation

An in-flight async Popover positioning update can continue after its effect cleanup, allowing stale work to overwrite a newer effect run's placement, styles, and positioned state.

## Solution

Added a `popover-stale-position` sandbox and browser test that starts a stale bottom positioning run, moves the popover to top with a newer real positioning run still pending, then releases the stale update. The test verifies the newer top placement and wrapper transform are not overwritten.

`Popover` now tracks cancellation per positioning effect run. Cleanup marks the run as canceled before stopping auto-update, and async continuations bail before committing state or styles after `computePosition` or a custom `updatePosition` callback resolves. The guard also skips disconnected popover elements before writing placement, transform, transform-origin, arrow styles, or `positioned` state.

The same cancellation predicate now guards Floating UI's `size` middleware `apply` callback, so stale runs cannot mutate wrapper width, max dimensions, or `--popover-*` sizing variables before `computePosition` resolves. The guard is documented in source because this callback runs inside `computePosition`, before the awaited call returns.

## Workaround

Consumers using a custom `updatePosition` callback can track whether a newer positioning request has superseded an older async callback and skip the stale callback before it calls Ariakit's provided `updatePosition` function.

```tsx
const updateVersionRef = useRef(0);

const updatePosition = useCallback<UpdatePosition>(async ({ updatePosition }) => {
  const updateVersion = updateVersionRef.current;

  await waitForAsyncWork();

  if (updateVersion !== updateVersionRef.current) {
    return;
  }

  await updatePosition();
}, []);

function movePopover() {
  updateVersionRef.current += 1;
  setPlacement("top");
}
```


diegohaz (commented):
**General PR review — no substantive findings.** This is an **initial review** (no prior `ariakit-pr-review` marker on this PR). Baseline: merge-base with `origin/main` (`1035b917`) → head `28f56c3`. Comment-only by design.

### Scope reviewed
All 5 changed files, plus surrounding repository context: the `usePopover` positioning effect, the Floating UI middleware helpers, the cancellation flow, the test infrastructure (`vitest.setup.ts`, `vitest.config.ts`, the source-plugin auto-render), the `@floating-ui/dom` resolution, the changeset config/conventions, and a cross-framework check.

### Conclusion
The fix is correct, complete, and minimal. The per-run `canceled` closure flag is the idiomatic "ignore stale async result" pattern; I found no simpler behavior-preserving alternative and no avoidable complexity in the library change. The two guard points are both load-bearing and non-redundant:

- **after `computePosition`** guards `currentPlacement`, `transform`, `--popover-transform-origin`, arrow styles, and `positioned`;
- **`size` middleware `apply`** (via the threaded `shouldCancelUpdate`) guards `width`, `max-*`, and the `--popover-*` sizing vars, which are written *during* `computePosition` and so can't be covered by the post-compute guard.

Cleanup sets `canceled = true` synchronously, and React runs an old run's cleanup before the newer run's effect body — so by the time a newer run's `apply` executes, any superseded run already sees `canceled === true`. A stale `size` `apply` therefore cannot survive a newer run. The custom-`updatePosition` path is also guarded after the awaited callback before `setPositioned(true)`.

### Validation performed
- Traced both tests end-to-end against the guard logic. **`test.ts`** (component): verified `popover.render()` mints a new `rendered` Symbol that is in the effect deps, so it re-runs the effect (cleanup → new cancellation closure) and the run released later is genuinely stale; the assertions correctly pin the `size`-middleware bail (`width`/`--popover-anchor-width` stay `240px`). **`test-browser.ts`** (Playwright): verified the stale run's `update`/`updatePosition` both bail after release while the latest top run's `transform` persists.
- Confirmed the test is **self-validating**: if none of the three `vi.mock` specifiers intercept, real Floating UI runs in JSDOM, `floatingMock.finished` stays `0`, and the opening `waitFor` fails loudly — it can't pass vacuously.
- Verified the nested `packages/ariakit-react-components/node_modules/@floating-ui/dom/dist/floating-ui.dom.mjs` mock target is a real pnpm workspace **symlink** (→ `.pnpm/@floating-ui+dom@1.7.6`) resolving to the exact module `popover.tsx` imports. It is **load-bearing, not dead code** — this confirms the existing triage of the CodeRabbit nitpick; no reply needed.
- Confirmed `test.ts` (React-only sandbox) is correctly **skipped under the Solid loader**: `vitest.setup.ts`'s `loadSolid` returns `false` on the missing `index.solid.tsx`, so `test.ts` naming matches the established convention (e.g. `examples/dialog-framer-motion`).
- Cross-framework completeness: `ariakit-solid-components` has no popover yet, so scoping the fix to `@ariakit/react-components` is correct.
- Changeset: targets are correct (`@ariakit/react-components` + `@ariakit/react`, the latter consumed via `workspace:*` and matching the repo's pattern of listing both for user-facing changelog text); `300-` prefix matches the repo convention.
- Reviewed CI: all checks green (Test, Type check, Lint, Node, React 18, Chrome/Firefox/Safari, app/Next.js builds). The Codex P2 (`size` side effects) is addressed by the `shouldCancelUpdate` guard and verified present.

### Areas not exhaustively validated
- I relied on the green CI run plus manual tracing rather than executing `pnpm test` locally (this worktree has no installed `node_modules`).
- The `document.getElementById(...)` + `instanceof` wrapper lookup in `test.ts` deviates slightly from the skill's `.ensure(...)` preference, but `@ariakit/test`'s `q` is role-based with no by-id/CSS query and the wrapper is a non-semantic Floating UI div — there's no actionable alternative, so I did not flag it inline.

app/src/sandbox/popover-stale-position/test.ts
```diff
@@ -0,0 +1,178 @@
+import { click, q, waitFor } from "@ariakit/test";
```
diegohaz: What are we testing here? Does it need to be this complex test file? Could we test this in `test-browser.ts` for simplicity?

diegohaz: This is not an issue because `test-browser.ts` already covers the user-facing custom `updatePosition` stale flow, while this component test covers the internal `size` middleware path. That path mutates `sameWidth` styles during `computePosition`, before the post-compute guard runs, so the test needs to deterministically pause Floating UI and resume the stale `size.apply` after cleanup. A browser test with real Floating UI cannot reliably hold that exact internal point; keeping this in the sandbox component test preserves the missing coverage.

diegohaz:
**Invalid**
- Invalid ([reply](https://github.com/ariakit/ariakit/pull/6176#discussion_r3333370367)): Kept the sandbox component test because `test-browser.ts` covers the custom `updatePosition` stale flow, while this test deterministically mocks Floating UI to resume stale `size.apply` after cleanup and verify `sameWidth` styles are not overwritten.

app/src/sandbox/popover-stale-position/test.ts
diegohaz: Fixed in b7116fc58 - removed the complex sandbox component test and its test-only sandbox controls, then documented the stale async guard points directly in `popover.tsx`.

diegohaz:
**Fixed**
- b7116fc58 ([reply](https://github.com/ariakit/ariakit/pull/6176#discussion_r3333551154)): Removed the complex sandbox component test and its test-only controls, added source comments around the stale async positioning guards, and updated the PR description to remove the stale component-test claim.

PR #6178, opened 2026-06-01
Split framework test suites

diegohaz:
## Motivation

The root Vitest scripts mixed `ARIAKIT_TEST_SUITE` and `ARIAKIT_TEST_LOADER`, and the default `test` command still selected framework-rendering tests. CI also kept React 18 in a separate job and did not run Solid unit tests.

## Solution

Use `ARIAKIT_TEST_LOADER` as the single framework test selector, split core tests from React and Solid suites in Vitest config, add typed runtime validation for the loader value, update root scripts and the React 18 wrapper, and consolidate unit CI into one matrix job with unsharded core/Solid plus sharded React/React 18 entries.

Framework example tests are discovered from directories with matching `index.<loader>.tsx` files so Solid does not collect React-only example tests. The contributing example-test command now uses `test-react`; watch mode is still available through Vitest CLI commands such as `pnpm test --watch` for core tests or `pnpm test-react watch <filter>` for React example tests.

Framework and example suites now run through `test-react`, `test-react18`, and `test-solid`; the default `coverage` script follows the core `test` scope.


vitest.config.ts
```diff
@@ -14,33 +17,50 @@ const includeWithStyles = [
...
 ];
 
-const allowedTestLoaders = ["react", "solid"] as const;
-export type AllowedTestLoader = (typeof allowedTestLoaders)[number];
-const loader = (process.env.ARIAKIT_TEST_LOADER ??
-  "react") as AllowedTestLoader;
-if (!allowedTestLoaders.includes(loader)) {
-  throw new Error(`Invalid loader: ${loader}`);
-}
+export type { AllowedTestLoader } from "./test-loader.ts";
```
diegohaz: Why do we need to export this?

vitest.setup.ts
```diff
@@ -125,20 +126,27 @@ function parseTest(filename?: string) {
...
   return {
     dir,
-    loader: (loader ?? "all") as AllowedTestLoader | "all",
+    loader: loader ?? "all",
   };
 }
 
-const LOADER = (process.env.ARIAKIT_TEST_LOADER ??
-  "react") as AllowedTestLoader;
+const LOADER = getTestLoader();
```
diegohaz: Rename to `loader` (lowercase) and move beforeEach after `if (!parseResult) return;`

test-loader.ts
```diff
@@ -0,0 +1,17 @@
...
+  return allowedTestLoaders.some((loader) => loader === value);
+}
+
+export function getTestLoader(): AllowedTestLoader | undefined;
+export function getTestLoader(
+  defaultLoader: AllowedTestLoader,
+): AllowedTestLoader;
+export function getTestLoader(defaultLoader?: AllowedTestLoader) {
+  const loader = process.env.ARIAKIT_TEST_LOADER ?? defaultLoader;
+  if (!loader) return undefined;
```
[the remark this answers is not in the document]
diegohaz: Confirmed — this is a real, if narrow, contract violation. The `getTestLoader(defaultLoader): AllowedTestLoader` overload promises a non-`undefined` return, but since `??` only falls back on `null`/`undefined`, an empty `ARIAKIT_TEST_LOADER=""` is preserved and the `if (!loader) return undefined;` line then returns `undefined`. The only caller passing a default is `examples/test-utils.ts` (`getTestLoader("react")`), so `preview()` would hit its `default:` branch and throw `Unknown test loader: undefined`.

The real-world trigger is low — every script either leaves the var unset or sets it to `react`/`solid`, never `""` — but `||` is strictly safer and honors the overload:

```diff
-  const loader = process.env.ARIAKIT_TEST_LOADER ?? defaultLoader;
+  const loader = process.env.ARIAKIT_TEST_LOADER || defaultLoader;
```

It leaves the no-arg overload unchanged too: with `""`, both `"" ?? undefined` and `"" || undefined` end up returning `undefined`.

vitest.config.ts
diegohaz: You don't — this re-export has no consumers, so it's safe to drop.

The only file that imports `AllowedTestLoader` is `vitest.setup.ts`, and after the move it imports it directly from `./test-loader.ts`; `examples/test-utils.ts` no longer imports the type at all. `vitest.config.ts` itself uses the type via the `import type` on line 9, not this re-export, and nothing imports anything from `vitest.config.ts` (it's only a Vitest config entry, never imported as a module).

```
$ grep -rn "AllowedTestLoader" --include="*.ts" .   # excluding node_modules
# outside test-loader.ts: only vitest.config.ts + vitest.setup.ts,
# both pulling the type from ./test-loader.ts — none from vitest.config.ts
```

So line 20 is dead. Removing it keeps the line-9 `import type` for internal use, so lines 33/53/73 still resolve.

vitest.config.ts
```diff
@@ -14,33 +17,50 @@ const includeWithStyles = [
...
-  ],
-};
+const testIncludesByLoader = {
+  react: getFrameworkTestIncludes("react"),
+  solid: getFrameworkTestIncludes("solid"),
+} satisfies Record<AllowedTestLoader, string[]>;
+
+const testIncludes = testLoader
+  ? testIncludesByLoader[testLoader]
+  : defaultTestIncludes;
```
diegohaz: `testIncludesByLoader` eagerly builds the include list for **both** loaders — each entry runs a `globSync` over the `examples`/`app/src` tree — but only `testIncludesByLoader[testLoader]` is ever read. For the default core `test` run (`testLoader` is `undefined`) *neither* value is used; it falls back to `defaultTestIncludes`. So every Vitest startup — including the core `test` job CI runs on every push — performs two full-tree globs whose results are thrown away.

Computing lazily for the active loader removes the wasted glob and the map indirection:

```suggestion
const testIncludes = testLoader
  ? getFrameworkTestIncludes(testLoader)
  : defaultTestIncludes;
```

Low priority — I measured the two globs at ~15 ms locally, so this is about avoiding needless work and simplifying, not a hot path. The adjacent `pluginsByLoader` map has the same eager shape (it instantiates both `reactPlugin()` and `solidPlugin()` even for the core run, which uses neither), so it's worth the same lazy treatment if you touch this.

diegohaz (commented):
## Review summary

**Initial review** of the full PR diff (8 files, +115/-63) against the merge-base `203385f52`...`6f9b63f3`. The repository state was inspected beyond the diff (collection scope, CI matrix, dangling references, dependency declarations). Note the branch predates `#6177` (commander v15), so the two-dot diff shows a phantom `commander` 15→14 change — it is **not** part of this PR's three-dot diff and will not revert commander on merge.

### Verdict
The split is well executed and **the collection logic is provably regression-free** (see validation below). No correctness blockers found. Two small, optional items are noted inline.

### Validation performed (Node 24, local)
- **Collection partition is exact.** `vitest list --filesOnly` per scope:
  - `test-react` (new) collects the **identical 146 files** as the old `ARIAKIT_TEST_SUITE=react` suite — zero diff.
  - core `test` (29) ∪ `test-react` (146) = **exactly** the old default `test` job's 175 files — **0 dropped, 0 added, no overlap**. The core/react split loses no coverage.
  - `test-solid` (new) collects only the 3 dirs that have `index.solid.tsx` (no React leakage), matching the stated goal.
- **Core runs without a framework plugin:** `pnpm test` → 29 files / 255 tests pass with `plugins = [sourcePluginInstance]` only.
- **Solid suite passes:** `ARIAKIT_TEST_LOADER=solid vitest` → 3 files / 4 tests pass with the Solid plugin.
- **`pnpm tsc`** clean (new overloads, `satisfies`, type re-export all type-check).
- **No dangling references** to the removed `test-watch` / `ARIAKIT_TEST_SUITE` or the renamed `test-react-18` (now `test-react18`) anywhere in scripts, CI, or docs.
- **Nested-brace glob works:** `{examples,app/src/{examples,sandbox}}/**/index.<loader>.tsx` matches 229 React / 3 Solid entries. No bare `index.tsx` files exist, so nothing the old globs caught is dropped.
- **`glob` is declared** in the root `package.json` (13.0.6) and resolves — the new import is satisfied.

### Findings (inline)
1. `test-loader.ts:14` — replied on the existing CodeRabbit thread: the `??`→`||` point is valid (empty-string env breaks the `getTestLoader(default): AllowedTestLoader` overload). Real-world impact is narrow since no script sets the var to `""`. Low priority.
2. `vitest.config.ts:20` — replied to your own question: the `export type { AllowedTestLoader }` re-export has **no consumers** and is safe to drop.
3. `vitest.config.ts:50-57` — `testIncludesByLoader` eagerly globs **both** loaders on every run (including the core `test` job, which uses neither result). Suggested computing lazily for the active loader. Low priority (~15 ms), simplification + avoidable work.

I did **not** comment on your `vitest.setup.ts:136` self-note (rename `LOADER`→`loader` + reorder) — it's a stylistic refactor with no correctness implication, your call.

### Operational notes (not code defects)
- **Required status checks:** this renames the CI jobs — old `Test (1/2)` / `Test (2/2)` / `Test React 18 (1/2|2/2)` become `Test`, `Test React (1/2|2/2)`, `Test React 18 (1/2|2/2)`, `Test Solid`. If branch protection pins required checks by name, update them or merges may block / pass without the new jobs.
- **`coverage` scope narrows to core.** `coverage` = `pnpm test --coverage`, which now runs only the 29 core files, so `packages/ariakit-react*` (and example-driven) coverage no longer registers under the default script (use `test-react --coverage` for that). This is stated as intentional in the PR description; coverage is not run in CI, so the impact is local-only. Flagging for confirmation.

### Not fully validated
- The full `test-react` / `test-react18` runs (146 files each) were not executed end-to-end — only collection was diffed (identical to the prior suite) and the core/solid suites were run. The React/React-18 CI jobs should confirm runtime parity.
- CI behavior of the dynamic matrix job-name expression and the env-gated shard step was reasoned through, not executed on a runner.

vitest.config.ts
diegohaz: Let's do this.

vitest.setup.ts
diegohaz: Fixed in 38f8008c9 - moved the loader lookup inside beforeEach after parsing a matching test file and renamed the local variable to loader.

vitest.config.ts
diegohaz: Fixed in 38f8008c9 - removed the unused AllowedTestLoader re-export from vitest.config.ts.

vitest.config.ts
diegohaz: Fixed in 38f8008c9 - made framework include generation and plugin selection lazy for the active loader.

diegohaz:
**Fixed**

- 38f8008c9 (https://github.com/ariakit/ariakit/pull/6178#discussion_r3333709094): Removed the unused `AllowedTestLoader` re-export from `vitest.config.ts`.
- 38f8008c9 (https://github.com/ariakit/ariakit/pull/6178#discussion_r3333709080): Moved the loader lookup inside `beforeEach` after parsing a matching test file and used lowercase `loader`.
- 38f8008c9 (https://github.com/ariakit/ariakit/pull/6178#discussion_r3333709097): Made framework include generation and plugin selection lazy for the active loader.

Issue #101, opened 2018-06-25
<Navigation> is missing


mrpsiho:
Hi!

First, thanks for such a cool lib!
Unfortunately, I have got this: `'reas' does not contain an export named 'Navigation'.` I am using `reas` NPM package (0.10.2), just checked it - it really does not have such component :(

Thomazella:
Hi mrpsiho! Thanks for reporting it. We'll have it fixed soon.

diegohaz:
Hey, @mrpsiho. I just released a new version ([`v0.10.3`](https://github.com/diegohaz/reas/releases/tag/v0.10.3)). Please, run `npm update reas` or `yarn upgrade reas`.

mrpsiho:
Awesome! Thanks again for the great tool! 🥇

PR #6182, opened 2026-06-01
Fix Shiki filename language fallback

diegohaz:
## Motivation

`getLangFromFilename` returned arbitrary file extensions as Shiki language IDs, but the app highlighter only loads a fixed set of grammars. Unknown or unloaded extensions like `.astro`, `.vue`, `.svelte`, `.md`, `.mjs`, `.cjs`, `.mts`, and `.cts` could therefore reach `codeToTokens` with a grammar that was not registered.

## Solution

Replace the unchecked cast with explicit filename and extension mappings that only return loaded grammars or the safe `text` fallback. Common JavaScript and TypeScript module extensions map to the loaded JavaScript/TypeScript grammars, framework markup extensions map to HTML, shell aliases map to Bash, and unknown values fall back to `text`.

The language type is narrowed to the loaded grammar IDs and safe fallback so explicit `lang` props cannot bypass the mapping with unloaded Shiki grammars. Display-equivalent language aliases such as `sh` and `markdown` still highlight with loaded grammars without rendering redundant filename headers, while real filenames and package-manager tabs stay visible. Lookup tables now use own-property checks so inherited object keys also fall back safely to `text`.


app/src/lib/shiki.ts
```diff
@@ -8,9 +8,44 @@
...
+  javascript: "javascript",
+  jsx: "jsx",
+  json: "json",
+  md: "text",
+  mdx: "text",
+  mjs: "javascript",
+  mts: "typescript",
+  py: "python",
+  python: "python",
+  sh: "bash",
```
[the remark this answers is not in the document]
diegohaz: Fixed in 90d3415b9 - `language-sh` now remains display-equivalent to the loaded Bash grammar, so it highlights as Bash without rendering an `sh` filename/header. I also narrowed `CodeBlockLanguage` to the loaded grammar IDs plus `text` so explicit `lang` props cannot bypass the safe mapping.

diegohaz:
**Fixed**

- `90d3415b9` ([reply](https://github.com/ariakit/ariakit/pull/6182#discussion_r3334900162)): Preserved headerless display for language aliases such as `language-sh` while still highlighting with the loaded Bash grammar, and narrowed `CodeBlockLanguage` to the loaded grammar IDs plus `text`.

app/src/lib/shiki.test.ts
```diff
@@ -0,0 +1,61 @@
+/**
```
diegohaz: No need for this test file.

app/src/lib/shiki.types.ts
```diff
@@ -0,0 +1,22 @@
+/**
```
diegohaz: Can we avoid this file and import CodeBlockLanguage from shiki.ts in code-block.types.ts?

app/src/lib/shiki.test.ts
diegohaz: Fixed in 49fc4839b - removed the new Shiki test file and kept verification on the existing lint/typecheck/build checks.

app/src/lib/shiki.types.ts
diegohaz: Fixed in 49fc4839b - removed the separate `shiki.types.ts` file, exported `CodeBlockLanguage` from `shiki.ts`, and updated `code-block.types.ts` to import it from there.

diegohaz:
**Fixed**
- 49fc4839b (https://github.com/ariakit/ariakit/pull/6182#discussion_r3335171329): removed the new `app/src/lib/shiki.test.ts` file per review.
- 49fc4839b (https://github.com/ariakit/ariakit/pull/6182#discussion_r3335171609): removed `app/src/lib/shiki.types.ts`, exported `CodeBlockLanguage` from `shiki.ts`, and updated the code-block type import.

**Verification**
- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build-app-lite`
- `git diff --check`

Patch #101 is updated/applied. New head: `49fc4839b`.

diegohaz (commented):
### Review summary

**Scope:** Full review (no prior review from this workflow exists on the PR). Diff reviewed from merge-base `a895a554f` to head `49fc4839b`, plus the surrounding code-block rendering pipeline (`shiki.ts`, `content-code-block.astro`, `code-block-content.astro`, `code-block-tabs.astro`, `code-block.types.ts`, `code-block.react.tsx`).

**Status:** No substantive new findings. The change is correct, simpler than the original cast, and the public `lang` surface is now compile-safe. Comment-only review per workflow policy.

#### What I verified
- **All 12 `CodeBlockLanguage` members resolve to a loaded grammar (or the safe `text` passthrough).** I inspected the actually-loaded `@shikijs/langs@4.1.0` grammars and confirmed their registered aliases: `javascript` → `["js","cjs","mjs"]`, `typescript` → `["ts","cts","mts"]`, `shellscript` → `["bash","sh","shell","zsh"]`, `python` → `["py"]`, plus `tsx`/`jsx`/`css`/`json`/`html`. So `js`, `ts`, `bash`, `python` reach `codeToTokens` as valid aliases, and `text` is Shiki's built-in plain passthrough. The original "unloaded grammar reaches `codeToTokens`" failure mode is fully closed.
- **Type narrowing is non-breaking.** Every literal `lang` passed to a code block in the repo is within the new union (`playground.astro` → `"bash"`/`"tsx"`, `reference-item-type.astro` → `"typescript"`), and every other `lang` flows through `getLangFromFilename`, which now only returns `CodeBlockLanguage`. No remaining `BundledLanguage` references; `shiki.types.ts` is fully removed with no dangling imports.
- **`isLangAlias` / `showFilename` logic is correct.** Bare-language fences (` ```sh `, ` ```markdown `, ` ```text `) are hidden as filenames since the whole label maps back to the same lang; real filenames (`index.tsx`, `Makefile`) and package-manager tabs (`npm`/`pnpm`/…) still display because they aren't extension keys. The `Object.hasOwn` guard in `getMappedLang` correctly prevents inherited keys (`constructor`, `toString`, `__proto__`) from leaking a non-grammar value.
- **`getPackageManagerTabs` simplification is safe.** `getLangFromFilename` now normalizes `sh` → `bash`, so `lang` can never be `"sh"`; dropping the `lang !== "sh"` branch is correct. This matches CodeRabbit's nitpick, which is already applied in the head.
- **No regression in actual docs.** Scanning every fence language in `app/**/*.md(x)` shows only `css`, `jsx`, `sh`, `javascript`, `test` in use — all map correctly (`test` → `text`, previously broken, now safe). Aliases like `shell`/`zsh`/`jsonc` that are no longer in `langByExtension` fall back to `text`, but none are used and the fallback is safe by design.
- **No changeset needed** — `app` is `"private": true` and listed in `.changeset/config.json` `ignore`.

#### Prior review threads (all resolved, re-verified against current code)
- Codex P2 "sh fences render as filenames" → fixed via `isLangAlias` (90d3415b9). ✅ Confirmed.
- Copilot "narrow `CodeBlockLanguage`" → narrowed to the 12 loaded IDs + `text`. ✅ Confirmed.
- Author "remove `shiki.test.ts`" and "fold `shiki.types.ts` into `shiki.ts`" → both done (49fc4839b). ✅ Confirmed.

#### Considered and intentionally not filed
- `.mjs`/`.cjs` map to `"javascript"` and `.mts`/`.cts` to `"typescript"`, whereas `.js`→`"js"` and `.ts`→`"ts"`. The long-form values highlight correctly but don't match a filename icon, so a `.mjs` tab would render without a JS icon (unlike `.js`). Mapping all JS/TS module extensions to `"js"`/`"ts"` would be marginally more consistent, but **no `.mjs/.cjs/.mts/.cts` files exist in `examples/` and these fence labels never appear in the docs**, so this has zero impact today. Not worth a code change.

#### Not fully validated
- Did not run `pnpm tsc` / `pnpm -F app test` locally (no `node_modules` in this worktree; install was out of scope for a review). Type soundness was reasoned about by enumerating every `lang` call site; `shiki.ts` itself is `@ts-nocheck`, so the narrowing is enforced only at the consumer files (all of which type-check cleanly by inspection). CI's typecheck/lint/build covers this.

PR #6191, opened 2026-06-02
Speed up @ariakit/test in jsdom by firing animation frames immediately

diegohaz:
## Motivation

`pnpm test-react` is dominated by the `@ariakit/test` package. To mimic real user interaction, methods like `click`, `type`, and `press` wait for animation frames between simulated events — `sleep()` does `requestAnimationFrame` → `setTimeout` → `requestAnimationFrame`, and it's called twice **per character** in `type`, three times in `press`, and twice in `click`/`hover`/`select`.

In jsdom and happy-dom, `requestAnimationFrame` fires on a fixed ~16ms cadence. Measured in this repo's jsdom environment, a single `nextFrame()` takes ~16.9ms and a default `sleep()` takes ~44.6ms — so the **two frames are ~76% of every sleep's cost**, and that frame cadence (not the simulated delay) dominates the suite's wall-clock. For example, `type("helloworld")` alone took ~960ms.

## Solution

In non-browser test environments only (gated by the existing `isBrowser` check), replace `requestAnimationFrame`/`cancelAnimationFrame` with an immediate macrotask. The replacement **batches** like jsdom does: callbacks registered before a tick are flushed together, in order, sharing one timestamp — so the *number of frames* and their *ordering*, which components depend on (e.g. `afterPaint`'s two nested frames), are preserved. Only the per-frame wall-clock is removed. The browser (Playwright) path keeps the native cadence, so real timing and painting are unaffected there.

```ts
// packages/ariakit-test/src/__utils.ts
if (!isBrowser) {
  // ...flush all rAFs registered before a tick together, on one macrotask...
  globalThis.requestAnimationFrame = (callback) => { /* immediate, batched */ };
  globalThis.cancelAnimationFrame = (id) => { /* ... */ };
}
```

Batching matters: a naive one-`setTimeout`-per-callback override loses jsdom's frame batching and made `examples/dialog-animated-various` flaky (it raced React's `data-leave` commit against `disclosure-content`'s `getComputedStyle`). The batching shim keeps it stable.

## Result

- Full `test-react` suite: **~50s → ~26s (~1.9×)**, **723/723** passing, verified stable across multiple runs.
- The previously-flaky `dialog-animated-various` passes **0/6** failures in isolation under the change.
- Solid suite, `oxlint`, `oxfmt`, and `tsc` all pass. Added focused unit tests for the shim's batching, nested-frame, and cancellation behavior (including same-frame sibling cancellation).

## Notes

This intentionally does **not** reduce the simulated `sleep()` delay (`defaultMs`): dropping it to `0` breaks a timing-calibrated animation test for marginal gain, and the frame cadence was the real cost. The change globally replaces `requestAnimationFrame`/`cancelAnimationFrame` in non-browser test environments when `@ariakit/test` is imported (noted in the changeset).

### DIFF
--- a/.changeset/200-test-immediate-animation-frames.md
+++ b/.changeset/200-test-immediate-animation-frames.md
@@ -1,0 +1,11 @@
+---
+"@ariakit/test": patch
+---
+
+Faster simulated interactions in test environments
+
+Methods such as `click`, `type`, and `press` wait for animation frames between simulated events to mimic real user interaction. In non-browser test environments (jsdom and happy-dom), `requestAnimationFrame` runs on a ~16ms cadence, which made interaction-heavy tests slow.
+
+These environments now resolve animation frames on an immediate macrotask instead, significantly speeding up test runs while preserving the order and number of frames that components depend on. The browser environment is unaffected and keeps using the native cadence.
+
+To achieve this, `@ariakit/test` now replaces the global `requestAnimationFrame` and `cancelAnimationFrame` functions in non-browser test environments when it's imported.
--- a/packages/ariakit-test/src/__utils.test.ts
+++ b/packages/ariakit-test/src/__utils.test.ts
@@ -1,6 +1,6 @@
 // oxlint-disable unbound-method
 import { expect, test } from "vitest";
-import { isBrowser, wrapAsync } from "./__utils.ts";
+import { isBrowser, nextFrame, wrapAsync } from "./__utils.ts";
 
 test("wrapAsync does not reapply browser polyfills when nested", async () => {
   const originalFocus = HTMLElement.prototype.focus;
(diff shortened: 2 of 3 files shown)


diegohaz:
Closing this — the approach doesn't hold up.

The fast-rAF shim made `requestAnimationFrame` fire on an immediate macrotask, which gave a real ~1.9× on `test-react` (50s → 26s) **but is less browser-faithful**: a real browser (and native jsdom) fires rAF at the ~16ms refresh cadence, so the shim simulated *less than a single frame* of settle. That's exactly why it went flaky on CI — concurrent-render tests (`combobox` filtering with `useTransition`/`useDeferredValue`) need real frame time for React to flush, and slow CI runners exceeded the compressed budget. It passed locally only because fast cores masked it.

I also tested the leaner idea of keeping native frames and just trimming the `setTimeout(10)` in `sleep()`. That turns out not to be slack either: `defaultMs=0` deterministically breaks `dialog-combobox-command-menu` (hide-on-outside-click), and `5` is flaky — the 10ms is load-bearing settle time. `8` is the local floor for a ~1.05× win, not worth it.

Conclusion: `sleep()` is already at its faithful minimum — there's no per-test timing speedup that doesn't either reduce browser fidelity or break tests. Any real `test-react` speedup should come from test **infrastructure** (vitest pool/workers/isolation/setup overhead/sharding), which keeps the timing and behavior identical. Pursuing that separately.

PR #6195, opened 2026-06-02
Resolve duplicate perf labels with exact probing

diegohaz:
## Motivation

Perf measurements de-duplicate repeated labels by checking whether previous labels equal the base label or start with the generated suffix prefix. That makes a hand-authored label like `foo #2` look like a duplicate of `foo`, which can inflate the suffix assigned to later measurements.

## Solution

Add a small `getUniquePerfLabel` helper that tracks used labels exactly and then probes ` #N` suffixes until it finds an unused label. `createPerfMeasure` now uses that helper instead of the broad `startsWith` check, so `foo` and a literal `foo #2` can coexist while repeated `foo` measurements still get unique labels.

A focused `perf.test.ts` regression covers the literal suffix case and suffix probing.

## Verification

- `pnpm lint-fix`
- `pnpm test packages/ariakit-scripts/src/perf.test.ts`
- `pnpm tsc`

No changeset: `@ariakit/scripts` is ignored by the changeset config and this is an internal private scripts package change.

### DIFF
--- a/packages/ariakit-scripts/src/perf.ts
+++ b/packages/ariakit-scripts/src/perf.ts
@@ -215,24 +215,15 @@ export function getUniquePerfLabel(
   labels: Iterable<string>,
   baseLabel: string,
 ): string {
-  const occurrences = new Map<string, number>();
+  const used = new Set(labels);
+  if (!used.has(baseLabel)) return baseLabel;
 
-  for (const label of labels) {
-    occurrences.set(label, (occurrences.get(label) ?? 0) + 1);
-  }
-
-  const duplicateCount = occurrences.get(baseLabel) ?? 0;
-  if (duplicateCount === 0) return baseLabel;
-
-  let suffix = duplicateCount + 1;
-  let label = `${baseLabel} #${suffix}`;
-
-  while (occurrences.has(label)) {
+  let suffix = 2;
+  while (used.has(`${baseLabel} #${suffix}`)) {
     suffix += 1;
-    label = `${baseLabel} #${suffix}`;
   }
 
-  return label;
+  return `${baseLabel} #${suffix}`;
 }
 
 function isTruthyEnv(name: string): boolean {
--- a/packages/ariakit-scripts/src/perf.ts
+++ b/packages/ariakit-scripts/src/perf.ts
@@ -211,6 +211,30 @@ export function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
   };
 }
 
+export function getUniquePerfLabel(
+  labels: Iterable<string>,
+  baseLabel: string,
+): string {
+  const occurrences = new Map<string, number>();
+
+  for (const label of labels) {
+    occurrences.set(label, (occurrences.get(label) ?? 0) + 1);
+  }
+
+  const duplicateCount = occurrences.get(baseLabel) ?? 0;
+  if (duplicateCount === 0) return baseLabel;
+
+  let suffix = duplicateCount + 1;
+  let label = `${baseLabel} #${suffix}`;
+
+  while (occurrences.has(label)) {
+    suffix += 1;
+    label = `${baseLabel} #${suffix}`;
+  }
+
+  return label;
+}
+
 function isTruthyEnv(name: string): boolean {
   return process.env[name] === "true" || process.env[name] === "1";
 }
@@ -834,11 +858,10 @@ export async function createPerfMeasure(
 
   const testTitle = testInfo.titlePath.filter(Boolean).join(" > ");
   const baseLabel = label ?? testTitle;
-  const duplicateCount = results.filter(
-    (r) => r.label === baseLabel || r.label.startsWith(`${baseLabel} #`),
-  ).length;
-  const resolvedLabel =
-    duplicateCount === 0 ? baseLabel : `${baseLabel} #${duplicateCount + 1}`;
+  const resolvedLabel = getUniquePerfLabel(
+    results.map((result) => result.label),
+    baseLabel,
+  );
 
   results.push({
     testFile: path.relative(process.cwd(), testInfo.file),
--- a/packages/ariakit-scripts/src/perf.test.ts
+++ b/packages/ariakit-scripts/src/perf.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, expect, test } from "vitest";
-import { getIntegerEnv, parseScriptProfile } from "./perf.ts";
+import {
+  getIntegerEnv,
+  getUniquePerfLabel,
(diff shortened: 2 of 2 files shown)


packages/ariakit-scripts/src/perf.ts
```diff
@@ -211,6 +211,30 @@ export function computeMedianMetrics(all: PerfMetrics[]): PerfMetrics {
...
+
+  let suffix = duplicateCount + 1;
+  let label = `${baseLabel} #${suffix}`;
+
+  while (occurrences.has(label)) {
+    suffix += 1;
+    label = `${baseLabel} #${suffix}`;
+  }
+
+  return label;
```
diegohaz: **Simplification: this helper only needs membership, not occurrence counts.**

The fix itself is correct — replacing the `startsWith` check with an exact-equality check on `baseLabel` is what lets a literal `foo #2` coexist with `foo`. But the machinery around it (a `Map<string, number>`, the count-accumulation loop, the `?? 0` default, and the `duplicateCount + 1` starting suffix) computes counts that aren't actually used for correctness. Correctness comes entirely from the membership probe (`occurrences.has(label)`), so a `Set` expresses the same logic directly and drops the bookkeeping:

```suggestion
  const used = new Set(labels);
  if (!used.has(baseLabel)) return baseLabel;

  let suffix = 2;
  while (used.has(`${baseLabel} #${suffix}`)) {
    suffix += 1;
  }

  return `${baseLabel} #${suffix}`;
```

Why the `duplicateCount + 1` head start isn't needed: in the normal incremental flow each resolved label is already unique, so `baseLabel` appears at most once in `labels` and the starting suffix is `2` either way. The count-based start only differs from `2` when the same exact `baseLabel` is present 2+ times in `labels` — which the resolver never produces, since it always assigns a unique label. In that pathological case the `Set` version fills the lowest free gap (`foo #2`) instead of skipping to `foo #3`, which is actually more consistent with the gap-filling your third test asserts.

Validated locally: the `Set` version passes all three assertions in the new test (ran `pnpm test packages/ariakit-scripts/src/perf.test.ts` against it), and an exhaustive simulation over realistic resolve sequences — including hand-authored `foo #2` / `foo #3` seeds — showed zero output divergence from the current implementation.

diegohaz (commented):
**Review summary — initial review**

**Scope:** Full diff between merge base `d144204` and head `004cd6a` (2 files: `packages/ariakit-scripts/src/perf.ts`, `packages/ariakit-scripts/src/perf.test.ts`), plus the surrounding `createPerfMeasure` call site and the file's existing helper/JSDoc conventions.

**Verdict:** The change is correct and a genuine improvement. The old `startsWith(`${baseLabel} #`)` check both over-counted hand-authored siblings *and* could re-emit an already-used `foo #2` (a real collision); the new exact-match + membership probe fixes both. Behavior in the normal repeated-label flow (`foo`, `foo #2`, `foo #3`, …) is preserved. No regressions, compatibility, or security concerns found.

**What I ran (on the PR head):**
- `pnpm test packages/ariakit-scripts/src/perf.test.ts` → 3/3 pass.
- `pnpm tsc` → clean (exit 0).
- Standalone differential simulation of the new helper vs. a `Set`-based alternative across realistic incremental resolve sequences (with hand-authored `foo #2` / `foo #3` seeds): 0 divergences. Also re-ran the PR's own test suite against the `Set` version (3/3 pass), then reverted.

**Findings (1, non-blocking):**
- One simplification on `getUniquePerfLabel` — the occurrence-count `Map` and `duplicateCount + 1` start are not load-bearing; a `Set` membership probe is equivalent and simpler. Posted inline with a `suggestion` and the supporting validation. Optional, author's call.

**Not separately flagged:** Missing JSDoc on the new export is consistent with the file's other simple helpers (`getIntegerEnv`, `computeMedianMetrics`), so not raised. The new test omits the trivial `[] → foo` / `["foo"] → "foo #2"` base cases, but those are exercised by the integration path and the three included cases cover the interesting behavior. No changeset is correctly omitted (`@ariakit/scripts` is in the changeset `ignore` list).

**Not fully validated:** I did not run the end-to-end Playwright perf suite that exercises `createPerfMeasure` against a live browser; the label-resolution logic itself is pure and fully covered by the unit test and simulation above.

packages/ariakit-scripts/src/perf.ts
diegohaz: Fixed in 44bff5182437aa78d11f3fdd94a56dc9851112ef - simplified `getUniquePerfLabel` to use Set membership and suffix probing while preserving the T083 regression behavior covered by `perf.test.ts`.

diegohaz:
**Fixed**
- 44bff5182437aa78d11f3fdd94a56dc9851112ef ([reply](https://github.com/ariakit/ariakit/pull/6195#discussion_r3340075134)): Simplified `getUniquePerfLabel` to use Set membership plus suffix probing, preserving the T083 regression behavior.

**Verification**
- `pnpm lint-fix`
- `pnpm test packages/ariakit-scripts/src/perf.test.ts`
- `pnpm tsc`

packages/ariakit-scripts/src/perf.ts
diegohaz: Confirmed resolved in `44bff518`. The Set-based `getUniquePerfLabel` matches the suggested implementation exactly, and I re-verified on the new head: `pnpm test packages/ariakit-scripts/src/perf.test.ts` passes 3/3 and `pnpm tsc` is clean. The test file is unchanged, so the three assertions (literal-suffix coexistence, skip-taken `#2`, gap-fill) now exercise the simplified probe. No remaining concerns on this thread.

diegohaz (commented):
**Re-review — incremental**

**Baseline:** my previous review's head `004cd6a` → current head `44bff518`. Only one new commit since then (`Simplify perf label suffix probing`), touching only `getUniquePerfLabel` in `packages/ariakit-scripts/src/perf.ts`. The test file is unchanged.

**Prior finding — resolved.** The single simplification I raised was applied exactly: `getUniquePerfLabel` now uses a `Set` membership probe instead of the occurrence-count `Map` + `duplicateCount + 1` start. I verified the fix in the current code (not just from the reply): the function matches the suggested implementation, `pnpm test packages/ariakit-scripts/src/perf.test.ts` passes 3/3, and `pnpm tsc` is clean. Replied on the thread to close it out.

**New findings:** none. The incremental commit is a pure behavior-preserving simplification I had already validated (exhaustive differential simulation showed zero divergence from the previous implementation across realistic resolve sequences), and it exposes no new issues in the surrounding code.

**Status:** No outstanding concerns. LGTM.

PR #6196, opened 2026-06-02
Run React and Solid render tests on happy-dom

diegohaz:
## Motivation

`pnpm test-react` is dominated by the simulated user-interaction layer in [`@ariakit/test`](packages/ariakit-test), which runs against jsdom — jsdom's DOM implementation is the main per-test cost. Running on happy-dom instead roughly halves the wall-clock, locally and in CI, without changing the simulated-interaction timing.

Measured locally (interleaved A/B, 146 files / 723 tests, all passing in both):

| Environment | `pnpm test-react` |
| --- | --- |
| jsdom | ~45.8s |
| happy-dom | **~22.5s** (~2× faster) |

## Solution

All test suites default to **happy-dom** — it provides the DOM every suite needs and is ~2× faster than jsdom for the `@ariakit/test` simulation layer:

```ts
// vitest.config.ts
const environment = process.env.ARIAKIT_TEST_ENV ?? "happy-dom";
```

jsdom is the opt-out:

- The **React 18 compatibility suite** (`test-react18`) sets `ARIAKIT_TEST_ENV=jsdom`: React 18's scheduler is more sensitive to happy-dom's faster timer cadence (see Notes).
- Individual tests with a *deterministic* happy-dom divergence add a `// @vitest-environment jsdom` comment.

### `@ariakit/test` happy-dom polyfills

Gaps in happy-dom are polyfilled (scoped to happy-dom via `isHappyDOM`; jsdom already behaves correctly), each as a small helper that restores itself after the interaction:

- **`validationMessage`** — happy-dom leaves it empty for built-in constraint violations; the shim returns a non-empty message.
- **`FormData`** — happy-dom wrongly includes disabled `<select>`/`<textarea>`; the shim excludes them per the HTML spec without dropping same-named enabled controls.
- **`selectionchange`** — happy-dom dispatches it synchronously inside `Selection.removeAllRanges()`; the shim defers it to a task, matching the spec/jsdom ordering.
- **`window.alert`** — stubbed as a no-op (happy-dom omits it).

### jsdom-pinned tests

Only where happy-dom is *deterministically* less faithful (retries can't fix these):

- **Examples:** `dialog-animated-various` (rAF keyframe leave timing); `combobox-tabs-animated` and `combobox-multiple` (a `startTransition`/`useDeferredValue` settle that doesn't complete under happy-dom's faster rAF); `visually-hidden` (happy-dom drops `clip: rect(0 0 0 0)`).
- **Core:** `collection-store`'s "orders rendered items by DOM position" test (its rAF-batched flush doesn't settle within a single `nextFrame`).

`inert` stays fully functional under happy-dom (matching real browsers) — closing-dialog assertions that need to inspect inert subtrees use `q.*.includesHidden(...)`.

### Retries for intermittent happy-dom flakes

happy-dom's faster rAF cadence (~0–4ms vs ~16ms) shortens React's concurrent-render settle window, which can intermittently starve timing-sensitive tests on slow CI (not reproducible locally). A single retry — scoped to the framework render suites, where the simulation runs — absorbs the occasional one-off; a test that fails twice still fails, so genuine regressions surface. The core unit suite and the jsdom suites don't retry.

```ts
retry: testLoader && environment === "happy-dom" ? 1 : 0,
```

### Notes

- The retry is a deliberate trade-off: the speedup comes largely from happy-dom's shorter settle window, which is inseparable from the occasional timing flake.
- The React 18 suite stays on jsdom because React 18's scheduler flakes some dialog-dismissal tests under happy-dom's timer cadence. The primary React 19 suite is unaffected.


examples/combobox-tabs-animated/test.ts
```diff
@@ -1,3 +1,9 @@
+// @vitest-environment jsdom
+//
+// These assertions query a tab inside a popover that is closing, and therefore
+// inert — as in real browsers and happy-dom, but not jsdom. @ariakit/test
+// excludes inert subtrees from role queries, so the tab isn't found
+// mid-animation. Pinned to jsdom pending inert-aware test updates.
```
diegohaz: I wonder if we shouldn't fix the tests here instead of relying on jsdom's lack of support for inert.

examples/dialog-combobox-command-menu/test.ts
```diff
@@ -1,3 +1,9 @@
+// @vitest-environment jsdom
+//
+// These assertions query options inside a dialog that is closing, and therefore
+// inert — as in real browsers and happy-dom, but not jsdom. @ariakit/test
+// excludes inert subtrees from role queries, so the options aren't found
+// mid-animation. Pinned to jsdom pending inert-aware test updates.
```
diegohaz: Again, I believe we should fix tests so they consider inert trees.

examples/popover-selection/test.ts
```diff
@@ -1,3 +1,8 @@
+// @vitest-environment jsdom
+//
+// happy-dom fires selectionchange synchronously inside Selection.removeAllRanges()
+// — real browsers queue it asynchronously — which closes the popover before
+// focus moves on mouse down. Pinned to jsdom.
```
diegohaz: Is there any way we can patch this directly in ariakit-test? If not, that's fine.

examples/combobox-tabs-animated/test.ts
```diff
@@ -1,3 +1,9 @@
+// @vitest-environment jsdom
+//
```
diegohaz: Can we remove the blank line between @vitest-environment and the explanation in all these cases?

packages/ariakit-test/src/__utils.happy-dom.test.ts
```diff
@@ -0,0 +1,49 @@
+// @vitest-environment happy-dom
```
diegohaz: Honestly, I believe this test file is unnecessary. We are already testing this when running the test suite on other packages.

packages/ariakit-test/src/__utils.ts
```diff
@@ -83,9 +83,97 @@ export function applyBrowserPolyfills() {
     window.PointerEvent = class PointerEvent extends MouseEvent {};
   }
 
+  // The following shims patch gaps in happy-dom's DOM implementation so it
+  // matches real-browser (and HTML spec) behavior. They are scoped to happy-dom
+  // because jsdom already behaves correctly.
+  const restoreHappyDOMShims: Array<() => void> = [];
+
+  if ("happyDOM" in window) {
```
diegohaz: We probably need an `isHappyDOM` const at the module level.

vitest.config.ts
```diff
@@ -76,7 +76,14 @@ export default defineConfig({
...
     testTimeout: 10_000,
-    environment: "jsdom",
+    // The framework render suites (test-react, test-solid) exercise the
+    // @ariakit/test simulation layer heavily, where happy-dom is ~2x faster than
+    // jsdom. The core suite (no loader) stays on jsdom, whose timing the
+    // low-level unit tests depend on. test-react18 sets ARIAKIT_TEST_ENV=jsdom
+    // to opt out: React 18's scheduler is more sensitive to happy-dom's faster
+    // timer cadence and flakes some dialog-dismissal tests.
+    environment:
+      process.env.ARIAKIT_TEST_ENV ?? (testLoader ? "happy-dom" : "jsdom"),
```
diegohaz: Do we really need JSDOM on bare `pnpm test`? If we do for just a few tests, I think they should be pinned, and instead we should run them on node environment.

vitest.setup.ts
```diff
@@ -14,6 +14,13 @@ import type { AllowedTestLoader } from "./test-loader.ts";
 
 failOnConsole();
 
+// happy-dom does not implement window.alert (jsdom does). Provide a no-op so
+// examples that call it — and tests that spy on it — work under happy-dom. This
+// is a no-op under jsdom, where window.alert already exists.
+if (typeof window.alert !== "function") {
+  window.alert = () => {};
+}
+
```
diegohaz: Shouldn't this be patched in ariakit-test as well?

package.json
```diff
@@ -18,7 +18,7 @@
     "test": "vitest",
     "bench": "vitest bench --config vitest.bench.config.ts",
     "test-react": "ARIAKIT_TEST_LOADER=react vitest",
-    "test-react18": "pnpm react18 test-react",
+    "test-react18": "ARIAKIT_TEST_ENV=jsdom pnpm react18 test-react",
```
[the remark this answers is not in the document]
diegohaz: We also need to add it to the other scripts that are missing cross-env like `test-react`.

examples/popover-selection/test.ts
diegohaz: Yes — fixed in 611c2f654. happy-dom dispatches `selectionchange` synchronously inside `Selection.removeAllRanges()` (the spec, jsdom, and real browsers queue it as a task). `applyBrowserPolyfills` now defers that dispatch to a macrotask, scoped to happy-dom and restored on teardown, so the example's `onSelect` no longer runs with a stale `activeElement`. popover-selection is un-pinned (verified 0/6 flakes + full suite green).

examples/dialog-combobox-command-menu/test.ts
diegohaz: Fixed in 890cfa266 — the two assertions that query an option inside the closing (and therefore inert) dialog now use `q.option.includesHidden(...)`, so the example runs on happy-dom (un-pinned). Verified 0/12 flakes.

examples/combobox-tabs-animated/test.ts
diegohaz: I tried that. The inert assertions here *are* fixable with `includesHidden` (as done for dialog-combobox-command-menu), but this example also fails for a **non-inert** reason: the selected-tab restoration is deferred via `startTransition`/`useDeferredValue`, and under happy-dom's faster rAF cadence that update never settles — the default tab is never restored after reopening, deterministically, even with extra waits. That's an app-state timing divergence, not a query/inert artifact, so it can't be made green on happy-dom. Kept it pinned but corrected the docblock to state the real reason in 890cfa266.

examples/combobox-tabs-animated/test.ts
diegohaz: Fixed in 890cfa266 — removed the blank `//` line between `@vitest-environment` and the explanation in the remaining pin docblocks.

packages/ariakit-test/src/__utils.happy-dom.test.ts
diegohaz: Removed in 611c2f654. The validationMessage/FormData shims are exercised by the form and select example tests under happy-dom, and the selectionchange shim by popover-selection.

packages/ariakit-test/src/__utils.ts
diegohaz: Done in 611c2f654 — added `export const isHappyDOM` at the module level; `isBrowser` and the happy-dom shim guard now use it.

vitest.setup.ts
diegohaz: Moved into @ariakit/test in 611c2f654 — `window.alert` is now stubbed at module level in `__utils.ts` (happy-dom only). It runs at import, before tests install spies (which `applyBrowserPolyfills` runs too late for), and lets `vitest.setup.ts` drop its `window` access — which was needed so the core suite can run on the node environment.

vitest.config.ts
diegohaz: Fixed in 70fdc48a7 — the core suite (`pnpm test`) now defaults to the `node` environment; the 7 files that use `document`/`HTMLElement`/`requestAnimationFrame` pin themselves back to jsdom with a `// @vitest-environment jsdom` comment. Core passes 263/263 on node.

diegohaz:
Thanks for the thorough review — addressed all of it across four commits.

**Fixed**
- **611c2f654**: Patched happy-dom's synchronous `selectionchange` in `@ariakit/test` and un-pinned `popover-selection`; moved the `window.alert` stub into `@ariakit/test` (module-level, which also lets `vitest.setup.ts` drop its `window` access); extracted `isHappyDOM`; removed the redundant happy-dom unit test.
- **890cfa266**: Made `dialog-combobox-command-menu` inert-aware with `q.option.includesHidden(...)` and un-pinned it; dropped the blank docblock line in the remaining pins.
- **70fdc48a7**: Core suite (`pnpm test`) now runs on the `node` environment; the 7 files that use `document`/`HTMLElement`/`requestAnimationFrame` pin themselves back to jsdom (263/263 on node).
- **f48cf0f70**: Wrapped the env-setting test scripts in `cross-env` (Windows portability); clarified the contributing docs.

**Kept pinned (with reasoning)**
- `combobox-tabs-animated`: the inert assertions were fixable, but it also has a *deterministic* `startTransition`/`useDeferredValue` settle divergence under happy-dom's faster rAF cadence (the default tab is never restored after reopening) that isn't an inert issue and can't be retried away. Kept pinned, docblock corrected to the real reason (890cfa266).

Verified locally: test-react 723, core 263 (node), test-solid 4; lint + tsc clean.

.changeset/200-happy-dom-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/test": patch
+---
+
+Improved happy-dom support by reporting a non-empty `validationMessage` for elements with built-in constraint violations (which happy-dom leaves empty) and by excluding disabled `<select>` and `<textarea>` controls from `FormData` (which happy-dom incorrectly includes).
```
diegohaz: **Changeset lists only two of the four happy-dom improvements shipped here.**

This PR makes four consumer-facing happy-dom changes to `@ariakit/test`, but the changelog entry mentions only `validationMessage` and `FormData`. The other two are also in this package and affect anyone running `@ariakit/test` under happy-dom:

- The `selectionchange` deferral in `Selection.removeAllRanges()` (happy-dom dispatches it synchronously; spec/jsdom/real browsers queue it as a task) — this directly affects selection-driven UI.
- The no-op `window.alert` (happy-dom omits it).

I confirmed both gaps exist in happy-dom 20.9.0 (`window.alert` is `undefined`; `selectionchange` fires synchronously inside `removeAllRanges()`), so they're real behavior changes worth recording. Suggest expanding the entry:

```suggestion
Improved happy-dom support by reporting a non-empty `validationMessage` for elements with built-in constraint violations (which happy-dom leaves empty), by excluding disabled `<select>` and `<textarea>` controls from `FormData` (which happy-dom incorrectly includes), by deferring the `selectionchange` event dispatched during `Selection.removeAllRanges()` to a macrotask (happy-dom dispatches it synchronously, unlike the spec, jsdom, and real browsers), and by stubbing a no-op `window.alert` (which happy-dom omits).
```

packages/ariakit-test/src/__utils.ts
```diff
@@ -83,9 +93,130 @@ export function applyBrowserPolyfills() {
     window.PointerEvent = class PointerEvent extends MouseEvent {};
   }
 
+  // The following shims patch gaps in happy-dom's DOM implementation so it
+  // matches real-browser (and HTML spec) behavior. They are scoped to happy-dom
+  // because jsdom already behaves correctly.
+  const restoreHappyDOMShims: Array<() => void> = [];
+
+  if (isHappyDOM) {
```
diegohaz: **Optional: extract the happy-dom shims into named helpers.**

`applyBrowserPolyfills` now spans ~180 lines and mixes two concerns: the cross-environment native polyfills (focus, `getClientRects`, pointer/clipboard events) and this ~115-line happy-dom-only spec-conformance block holding three independent shims. Pulling each shim into a small named helper that returns its own restore function — e.g. `shimValidationMessage()`, `shimFormData()`, `shimSelectionChange()` — would keep `applyBrowserPolyfills` scannable and let each shim's intent be carried by its name rather than only its comment:

```ts
if (isHappyDOM) {
  restoreHappyDOMShims.push(
    shimValidationMessage(),
    shimFormData(),
    shimSelectionChange(),
  );
}
```

Behavior-preserving and non-blocking — purely readability. The shims themselves are correct (I verified each addresses a real happy-dom gap).

diegohaz (commented):
## Review summary

**Initial review** (no prior review from this skill). Baseline: merge-base `4f3ecedc` (main) → PR head `542357c0`. Scope: full diff (21 files) plus surrounding context — `@ariakit/test` polyfills, the Vitest environment/retry config, the per-suite scripts, every jsdom pin (examples + core), the changeset, and the contributing.md/renovate changes.

This is a well-constructed, thoroughly-iterated PR. I verified the technical premises rather than taking them on trust, and they all hold. Only two minor, non-blocking findings — both posted inline.

### What I validated locally (Node 24, happy-dom 20.9.0)
- **Ran the shim-exercising example suites under `test-react`** (`form`, `form-select`, `select-form-disabled`, `popover-selection`) plus the jsdom-pinned `visually-hidden` → **5 files / 21 tests pass**. This exercises all four shims end-to-end (validationMessage, FormData-disabled, selectionchange, and the `window.alert` spy in `select-form-disabled`).
- **Independently confirmed each happy-dom gap the shims target** with a standalone happy-dom check:
  - `validationMessage` is `""` for a `required` input whose `validity.valid === false`;
  - `new FormData(form)` wrongly includes disabled `<select>`/`<textarea>` but correctly excludes disabled `<input>` — exactly matching the PR's scoping of the shim to SELECT/TEXTAREA;
  - `window.alert` is `undefined`;
  - `selectionchange` fires **synchronously** inside `Selection.removeAllRanges()`.
- **Confirmed the validationMessage string** (`"Constraints not satisfied"`) matches what `examples/form/test.ts` and `examples/form-select/test.ts` assert, so the shim keeps the shared assertions green under both environments.
- **Confirmed the core→node migration is complete**: of the unpinned core-suite test files, none touch DOM APIs; `composite-store.test.ts` (unpinned) uses plain objects, while `collection-store.test.ts` (pinned) uses `document.createElement`, so the 7 jsdom pins are correctly scoped.
- **Confirmed `test-react18` env propagation**: `react18.ts` spawns with `...process.env`, so `cross-env ARIAKIT_TEST_ENV=jsdom` reaches the inner `vitest`, keeping the React 18 suite on jsdom with `retry: 0`.

### Findings (both inline, both minor)
1. **`.changeset/200-happy-dom-support.md`** — lists only 2 of the 4 happy-dom improvements; the `selectionchange` deferral and `window.alert` no-op are also consumer-facing `@ariakit/test` changes worth recording. Suggestion provided.
2. **`packages/ariakit-test/src/__utils.ts`** — optional readability: extract the ~115-line happy-dom block into three named helpers so `applyBrowserPolyfills` stays scannable. Behavior-preserving.

### Reviewed and found correct (no comment needed)
- The `selectionchange` shim (temporarily wrapping `document.dispatchEvent` only during `removeAllRanges`, deferring via `setTimeout`, restoring in `finally`) is correct; the deferred event is dispatched once, after restore, via the captured original — no double-dispatch or recursion. The leftover own `dispatchEvent` property on `document` is benign.
- `PatchedFormData` accesses no `this` before `super()`, restores names in `finally` even if construction throws, and blanks-then-restores (rather than deleting entries) to avoid dropping same-named enabled controls — matching the documented rationale.
- The `dialog-combobox-command-menu` switch to `q.option.includesHidden(...)` doesn't weaken the jsdom path (the option was always findable there) and correctly handles the now-inert closing dialog under happy-dom.
- All shims are forward-compatible: if happy-dom fixes a gap, each shim degrades to a no-op rather than breaking.

### One residual risk (not a blocker — already documented in the PR)
The global `retry: 1` on the happy-dom suites is an explicit, well-argued trade-off (the speedup is inseparable from the shorter settle window). A deterministic regression still fails both attempts and surfaces; a *partially* flaky genuine regression would have its detection probability reduced. The author documents this and pins the one deterministic offender (`combobox-multiple`) to jsdom instead of relying on retry, which is the right call.

### Not fully validated
I did not attempt to reproduce the CI-only timing flakes locally (per the PR, they don't reproduce locally), and I did not re-run the full `test-react`/`test-react18`/`test-solid`/`pnpm test` suites end-to-end — CI already exercises those across shards. My local runs were scoped to the shim-relevant suites.

vitest.config.ts
diegohaz: Reconsidered, per your follow-up: I reverted the core suite back to **jsdom** in 265a9ba36 and dropped the per-file pins. vitest 4 removed `environmentMatchGlobs`, so a clean per-package jsdom list isn't expressible without a `projects` restructure, and the node speedup over jsdom is marginal for this suite — so keeping core on jsdom (its lower-level unit tests need a DOM) is simpler and removes the per-file fragility. The `@ariakit/test` packages that run in the react/solid suites already get a DOM via happy-dom.

vitest.config.ts
```diff
@@ -55,6 +55,15 @@ const testExcludes = [
...
 ];
 
+// The framework render suites (test-react, test-solid) exercise the
+// @ariakit/test simulation layer heavily, where happy-dom is ~2x faster than
+// jsdom. The core suite (no loader) stays on jsdom, whose DOM its lower-level
+// unit tests depend on. test-react18 sets ARIAKIT_TEST_ENV=jsdom to opt out of
+// happy-dom: React 18's scheduler is more sensitive to happy-dom's faster timer
+// cadence and flakes some dialog-dismissal tests.
+const environment =
+  process.env.ARIAKIT_TEST_ENV ?? (testLoader ? "happy-dom" : "jsdom");
```
diegohaz: What if we default to `happy-dom`?
```suggestion
  process.env.ARIAKIT_TEST_ENV ?? "happy-dom";
```

.changeset/200-happy-dom-support.md
diegohaz: Fixed in 020f9cade — the changeset now lists all four happy-dom improvements: `validationMessage`, disabled-control `FormData`, the `selectionchange` deferral, and the `window.alert` no-op.

packages/ariakit-test/src/__utils.ts
diegohaz: Done in 902bbecdf — extracted the three shims into named helpers (`patchHappyDOMValidationMessage` / `patchHappyDOMFormData` / `patchHappyDOMSelectionChange`), each returning its own restore function, so `applyBrowserPolyfills` just composes them. Behavior-preserving (verified by the form/select/popover-selection tests + tsc).

vitest.config.ts
diegohaz: Done in 96cfb5574 — `process.env.ARIAKIT_TEST_ENV ?? "happy-dom"`. Core now runs on happy-dom too, which gives every package a DOM (resolving the 'which packages need a DOM' question without a list). Only `collection-store`'s "orders rendered items by DOM position" test fails there — its requestAnimationFrame-batched flush doesn't settle within a single `nextFrame` under happy-dom's faster rAF — so it's the one jsdom pin. I also scoped the retry to the framework render suites (`testLoader && environment === "happy-dom"`) so the deterministic core unit tests don't retry. Verified: core 266, test-react 724.

diegohaz:
Addressed the latest round of comments:

**Fixed**
- **020f9cade**: The changeset now lists all four consumer-facing happy-dom improvements (`validationMessage`, disabled-control `FormData`, the `selectionchange` deferral, and the `window.alert` no-op).
- **902bbecdf**: Extracted the three happy-dom shims into named helpers — `patchHappyDOMValidationMessage`, `patchHappyDOMFormData`, `patchHappyDOMSelectionChange` — so `applyBrowserPolyfills` just composes them. Behavior-preserving.
- **96cfb5574**: Defaulted all suites to happy-dom (`process.env.ARIAKIT_TEST_ENV ?? "happy-dom"`). Every package now gets a DOM, so there's no node/jsdom split or per-package list; only `collection-store`'s DOM-position test needs a jsdom pin (its rAF-batched flush doesn't settle in one `nextFrame`). The retry is now scoped to the framework render suites so the deterministic core unit tests don't retry.

Verified: core 266, test-react 724, test-solid 4; lint + tsc clean.

PR #6198, opened 2026-06-02
Reset dialog outside focus state on reopen

diegohaz:
## Motivation

Reused dialogs can keep stale outside-interaction focus state after they close. If a dialog receives focus during one open cycle, closes, and then reopens without remounting, the next open cycle can start as though the dialog has already been focused.

That makes newly inserted, unmarked outside elements more lenient than they would be for a fresh dialog instance.

## Solution

Reset the `focusedRef` gate when the outside-interaction listeners are set up for an open dialog, after `open`, `domReady`, and `contentElement` are available. Cleanup remains limited to removing the listener so the reset is tied to setup rather than teardown.

Added a React sandbox regression test that reuses the same mounted dialog, primes the focus gate, closes and reopens it, then focuses a newly inserted outside input before any new inside focus.

## Testing

- `pnpm lint-fix`
- `pnpm test-react app/src/sandbox/dialog-reopen-focus-state/test.ts app/src/sandbox/menu-radio-default-checked/test.ts app/src/sandbox/tooltip-4894/test.ts app/src/sandbox/hovercard-nested-listener/test.ts`
- `pnpm tsc`
- `pnpm build`

### DIFF
--- a/.changeset/300-dialog-reopen-focus-state.md
+++ b/.changeset/300-dialog-reopen-focus-state.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) to reset outside-interaction focus tracking when reused dialogs reopen.
--- a/app/src/sandbox/dialog-reopen-focus-state/index.react.tsx
+++ b/app/src/sandbox/dialog-reopen-focus-state/index.react.tsx
@@ -1,0 +1,23 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+
+export default function Example() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <Ariakit.Button onClick={() => setOpen(true)}>Open dialog</Ariakit.Button>
+      <Ariakit.Dialog
+        open={open}
+        onClose={() => setOpen(false)}
+        unmountOnHide={false}
+        autoFocusOnShow={false}
+        autoFocusOnHide={false}
+      >
+        <Ariakit.DialogHeading>Dialog</Ariakit.DialogHeading>
+        <button type="button">Focus inside</button>
+        <Ariakit.DialogDismiss>Close dialog</Ariakit.DialogDismiss>
+      </Ariakit.Dialog>
+    </>
+  );
+}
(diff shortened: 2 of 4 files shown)


diegohaz (commented):
**General PR review — initial review, no blocking findings.**

Reviewed the full change, not just the diff: the shared `useEventOutside` hook, its callers in `dialog.tsx`, the `focusedRef` gate semantics, the `domReady`/`contentElement` lifecycle, the new sandbox + test, and the changeset.

### Scope & baseline
- Initial review. Baseline: merge-base `2474df0` → head `6151197`.
- Files: `use-hide-on-interact-outside.ts` (1-line fix), new sandbox `dialog-reopen-focus-state` (`index.react.tsx` + `test.ts`), changeset `300-dialog-reopen-focus-state.md`.

### What I verified
- **The fix is correct and minimal.** Resetting `focusedRef.current = false` inside the focus-tracking layout effect (after the `open` / `domReady` / `contentElement` guards) ties the reset to setup, so each open cycle starts from a fresh "not yet focused" gate. For `unmountOnHide` (default) the component remounts and the gate is already fresh, so the reset is a no-op there; it only changes behavior for reused (kept-mounted) dialogs, which is exactly the bug. Placing it on setup rather than in cleanup is the more robust choice.
- **The test genuinely guards the regression.** I temporarily removed the one fix line and the test fails (the `waitFor` times out — the reopened dialog stays visible because the stale `focusedRef` suppresses the outside-focus close). With the fix it passes. Validated with `pnpm test-react app/src/sandbox/dialog-reopen-focus-state/test.ts`.
- **No regression on the shared hook's other consumers.** Dialog/Hovercard/Tooltip share this hook. Ran `pnpm test-react app/src/examples/dialog-framer-motion/test.ts app/src/sandbox/hovercard-nested-listener/test.ts app/src/sandbox/tooltip-4894/test.ts app/src/sandbox/dialog-reopen-focus-state/test.ts` → 4 files / 7 tests pass. The reset is consistent with the gate's documented intent (per-open-cycle), so reused hovercards/tooltips benefit too.
- **No Solid gap.** The dialog interact-outside logic exists only in `ariakit-react-components`; there is no Solid equivalent to keep in sync, so a React-only fix is appropriate.
- **Conventions.** Sandbox naming (`dialog-` prefix + kebab descriptor), no `preview.json` needed (title defaults to folder name), `test.ts` correctly paired with `index.react.tsx`, and the changeset (`@ariakit/react-components` + `@ariakit/react` patch, `Fixed [Dialog](…)` wording) all match sibling changesets.

### Not fully validated
- Did not re-run `pnpm tsc` / `pnpm build` (the change is a trivially-typed `boolean` ref assignment; the PR description reports both green) and did not run the Playwright `test-browser.ts` dialog suites locally (JSDOM coverage here is sufficient for this focus-gate behavior).

Nothing actionable to change. CodeRabbit also reported no actionable comments.

PR #6199, opened 2026-06-02
Remove dead CheckboxCheck function children branch

diegohaz:
## Motivation

`CheckboxCheck` still had a leftover function-children branch in its unchecked path. When `checked` was false, that branch could pass a function-valued `children` prop through to React instead of rendering no checkmark. This is misleading because `children` is typed as `ReactNode`, and Ariakit's `createElement` only invokes function props through `render`, not through `children`.

## Solution

Remove the dead `typeof props.children === "function"` branch so unchecked check components return `null`. Include the required patch changeset for `@ariakit/react-components` and the public `@ariakit/react` re-export.

## Verification

- `pnpm lint`
- `pnpm tsc`
- `pnpm test-react --run`
- `pnpm -F @ariakit/react build`


packages/ariakit-react-components/src/checkbox/checkbox-check.test.react.tsx
```diff
@@ -0,0 +1,17 @@
+import { render } from "@ariakit/test/react";
```
diegohaz: No need for this test file.

diegohaz: Fixed in 1e8674f1912564b730a02d43649f9706ff23007f - removed the test file as requested.

diegohaz:
**Fixed**

- 1e8674f1912564b730a02d43649f9706ff23007f ([reply](https://github.com/ariakit/ariakit/pull/6199#discussion_r3340445312)): removed the `CheckboxCheck` test file as requested.

PR #6200, opened 2026-06-02
Fix composite renderer active ancestor lookup

diegohaz:
## Motivation

`CompositeRenderer` uses the active item index as one of the `persistentIndices` passed to the virtual collection renderer. When the active id points to a generated nested descendant such as `composite/1/nested`, the old slash-id fallback split the searched id and compared each bare segment against the fully qualified item id. That missed generated ancestor ids such as `composite/1`, so the active item ancestor could remain virtualized out.

## Solution

Update the lookup so exact item ids win first. If no exact item exists, resolve slash-composed descendants by choosing the closest matching ancestor prefix, preserving recursive lookup for explicit nested item ids.

The new app sandbox React test covers generated ancestors with and without nested metadata, explicit nested item ids, exact ids that follow an ancestor prefix, and overlapping prefix candidates.


diegohaz (commented):
**Initial review — comment-only.** No substantive findings; the change is correct, well-scoped, and the new test genuinely guards it. Details below.

### Scope & baseline
- Diff from merge-base `4f3eced` → head `0da4c24` (4 files: `composite-renderer.tsx`, sandbox `index.react.tsx` + `test.ts`, changeset).
- Looked beyond the diff: the `findById` call site (`activeIndex` → `persistentIndices`), `getCollectionRendererItemId`/`getItemObject` semantics in `collection-renderer.tsx`, the root vitest rules that collect sandbox `test.ts` under the React loader, and a search for a Solid equivalent.

### Validated locally (Node 24, `pnpm install`)
- `pnpm tsc` → clean.
- `ARIAKIT_TEST_LOADER=react vitest run app/src/sandbox/composite-renderer-active-ancestor/test.ts` on head → 5/5 pass.
- Confirmed the test actually catches the bug: reverted **only** `composite-renderer.tsx` to the base `findById` while keeping the new test → 3/5 cases fail (`generated`, `partial-nested`, `longest-prefix`) with rendered indices `["0","2"]` vs expected `["0","1","2"]`. The other 2 (`explicit-nested`, `exact-over-prefix`) pass on both versions, so they act as regression guards for the exact-match path. Restored head; working tree left clean.

### Correctness
The exact-first, then longest-prefix design is sound. The `` `${itemId}/` `` guard correctly avoids partial-token false matches (e.g. `foo` vs `foobar`); and since every prefix-matching `itemId` is a prefix of the same `id`, ranking by character length is an unambiguous proxy for ancestor depth, with the deepest match always living in the correct top-level subtree (so the returned top-level index is unambiguous). Dropping the old `ids.some((id) => itemId === id)` fallback is not a regression — it only matched bare single-token ids against mid-path segments, which isn't a real id shape produced here; generated descendants like `composite/1/nested` were never matched by it (that was the bug).

### Performance (considered, not blocking)
For nested virtualized composites whose `activeId` is a generated descendant, `findById` now does up to two full passes (`findExactById` scans fully with no exact hit, then `findPrefixById`'s `forEach` does not short-circuit by design). The common flat case still short-circuits on the exact match, and the only added cost lands on exactly the nested-descendant case that previously returned the wrong index — i.e. correctness for a tiny, narrow cost. A single merged pass could avoid the double traversal but would need a 3-field result plus exact-over-prefix priority handling, which reads worse than the current two-function split, so the current structure is the better trade-off.

### Other checks
- Changeset scope is right: `@ariakit/react-components` patch, which auto-bumps the `@ariakit/react` dependent. `CompositeRenderer` is a react-components-only export (not re-exported from `@ariakit/react`), and the reference link is valid.
- No Solid `CompositeRenderer` exists, so there is no parallel implementation to mirror the fix into.
- Sandbox conventions are followed: module-first folder name, no `preview.json` needed (title defaults to the folder), and `app` already declares the `@ariakit/react-components` dependency.
- The CodeRabbit "docstring coverage 0%" warning is a generic threshold — this file consistently omits JSDoc on its private helpers (`findFirst`, `findLast`, `countItems`, `getItemObject`), so adding it only to the new functions would be inconsistent. No change recommended.

### Not exhaustively validated
Did not run the Playwright/browser or visual suites (not applicable to this JSDOM-level change), and did not benchmark traversal cost since the affected path is narrow and was previously incorrect anyway.

LGTM.

PR #6202, opened 2026-06-02
Fix form store array index detection

diegohaz:
## Motivation

Form store path handling used the exported `isInteger` helper to decide whether a path segment should create or update an array. That helper intentionally preserves loose numeric coercion behavior, so segments such as `-1`, `Infinity`, and `NaN` could be classified as array indexes and write non-index properties onto arrays.

The form store needs a stricter path-index rule than the shared utility.

## Solution

This adds a local `isArrayIndex` predicate for form store paths. It only accepts canonical JavaScript array indexes: string or number segments that are integers, non-negative, within the valid array-index range, and round-trip through `String(index)`.

The regression tests cover:

```ts
store.setValue("users.-1.name", "Ada");
store.setValue("stats.Infinity.count", 1);
store.setValue("meta.NaN.value", true);
store.setValue("items.4294967295.name", "Grace");
```

Those paths now create object keys instead of malformed arrays, while canonical array paths such as `users.0.name` and runtime numeric segments such as `["users", 0, "name"]` still create arrays.

### DIFF
--- a/packages/ariakit-components/src/form/form-store.ts
+++ b/packages/ariakit-components/src/form/form-store.ts
@@ -6,7 +6,7 @@ import {
   throwOnConflictingProps,
 } from "@ariakit/store";
 import type { Store, StoreOptions, StoreProps } from "@ariakit/store";
-import { applyState, defaultValue, isInteger, isObject } from "@ariakit/utils";
+import { applyState, defaultValue, isObject } from "@ariakit/utils";
 import type {
   AnyObject,
   PickRequired,
@@ -23,6 +23,7 @@ import { createCollectionStore } from "../collection/collection-store.ts";
 import type { DeepMap, DeepPartial, Names, StringLike } from "./types.ts";
 
 type ErrorMessage = string | undefined | null;
+const maxArrayIndex = 2 ** 32 - 2;
 
 function nextFrame() {
   return new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
@@ -52,32 +53,44 @@ export function get<T>(
   return get(values[key], rest, defaultValue);
 }
 
+function isArrayIndex(key: string | number) {
+  if (typeof key !== "string" && typeof key !== "number") return false;
+  const stringKey = String(key);
+  const index = Number(stringKey);
+  return (
+    Number.isInteger(index) &&
+    index >= 0 &&
+    index <= maxArrayIndex &&
+    String(index) === stringKey
+  );
+}
+
 // Returns the existing nested value if it is an array or object, otherwise
-// creates a new container based on whether the next key is an integer.
-function getOrCreateNested(nestedValues: unknown, nextKey: string) {
+// creates a new container based on whether the next key is an array index.
+function getOrCreateNested(nestedValues: unknown, nextKey: string | number) {
   if (Array.isArray(nestedValues) || isObject(nestedValues)) {
     return nestedValues;
   }
-  return isInteger(nextKey) ? [] : {};
+  return isArrayIndex(nextKey) ? [] : {};
 }
 
 function set<T extends FormStoreValues | unknown[]>(
   values: T,
-  path: StringLike | string[],
+  path: StringLike | Array<string | number>,
   value: unknown,
 ): T {
   const [k, ...rest] = Array.isArray(path) ? path : String(path).split(".");
   if (k == null) return values;
   const key = k as keyof T;
-  const isIntegerKey = isInteger(key);
-  const nextValues = isIntegerKey ? values || [] : values || {};
+  const isArrayIndexKey = isArrayIndex(k);
+  const nextValues = isArrayIndexKey ? values || [] : values || {};
   const nestedValues = nextValues[key];
   const nextKey = rest[0];
   const result =
     rest.length && nextKey != null
       ? set(getOrCreateNested(nestedValues, nextKey), rest, value)
       : value;
-  if (isIntegerKey) {
+  if (isArrayIndexKey) {
     const index = Number(key);
     if (values && Array.isArray(values)) {
       const copy = [...values];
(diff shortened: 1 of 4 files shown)


diegohaz:
I reviewed the outside-diff note about preserving array containers when a non-index segment is written under an existing array. I’m not applying that suggestion because the issue here is specifically that non-index properties on arrays produce malformed form state. When a path segment such as `-1`, `Infinity`, `NaN`, or `4294967295` is not a valid array index, the form store should treat it as an object key instead of preserving an array and assigning a non-index property to it. That matches the current regression tests and the intended form-store-specific hardening.

packages/ariakit-components/src/form/form-store.ts
```diff
@@ -52,32 +53,44 @@ export function get<T>(
...
+  const isArrayIndexKey = isArrayIndex(k);
+  const nextValues = isArrayIndexKey ? values || [] : values || {};
   const nestedValues = nextValues[key];
   const nextKey = rest[0];
   const result =
     rest.length && nextKey != null
       ? set(getOrCreateNested(nestedValues, nextKey), rest, value)
       : value;
-  if (isIntegerKey) {
+  if (isArrayIndexKey) {
```
diegohaz, line 93: **A non-index segment under a parent that already holds a *populated* array silently converts that array into a plain object**

Now that `isArrayIndex("-1")` (and `Infinity` / `NaN` / `4294967295`) returns `false`, this branch is skipped for those segments and `set` falls through to `return Object.assign({}, values, { [key]: result })` below. When `values` is an array, `Object.assign({}, [...], …)` produces a plain object, so the array container is lost.

Verified against the real store on this PR head (temporary test, since `set` isn't exported):

```ts
const store = createFormStore({});
store.setValue("users.0.name", "Ada");
Array.isArray(store.getValue("users")); // true
store.setValue("users.-1.name", "Grace");
Array.isArray(store.getValue("users")); // false  ← was true before this PR
store.getValue("users"); // { "0": { name: "Ada" }, "-1": { name: "Grace" } }
```

Before this PR, `isInteger("-1")` was `true`, so `users` stayed an array (with a `-1` expando). This is the *same* `Object.assign` mechanism that already applied to non-numeric keys like `users.foo`, so it isn't new in principle — but the PR newly routes `-1` / `Infinity` / `NaN` / `4294967295` through it. The realistic footgun is a computed index such as a `findIndex(...)` that returns `-1`: a single stray write now nukes the whole array into an object (so downstream `users.map(...)` / `Array.isArray(users)` break), whereas before the real elements survived.

Note: the CodeRabbit outside-diff suggestion (keep the array and assign `copy[key] = result` for non-index keys) would reintroduce exactly the malformed-array-with-expando this PR sets out to eliminate, so it isn't the right remedy.

This is a design decision rather than an outright bug — collapsing to a clean object is arguably consistent with the PR's premise that `-1` is an object key. But the new tests only cover fresh stores, so the mixed case is unpinned. Suggested: make the decision explicit and add a regression test documenting it:

```ts
test("non-index segment replaces an existing array container with an object", () => {
  const store = createFormStore({});
  store.setValue("users.0.name", "Ada");
  store.setValue("users.-1.name", "Grace");
  expect(Array.isArray(store.getValue("users"))).toBe(false);
  expect(store.getState().values).toEqual({
    users: { "0": { name: "Ada" }, "-1": { name: "Grace" } },
  });
});
```

If array preservation is actually preferred for already-created arrays, that needs an explicit guard (e.g. ignore non-index writes when the existing container is a non-empty array) — not the expando approach.

diegohaz (commented):
**General PR review — initial review**

**Scope:** Full diff `3a0740fbb...ab42f08e` (merge-base `3a0740fbb`) across all 4 changed files: `form-store.ts`, `form-store.test.ts`, `misc.test.ts`, and the changeset. Inspected the surrounding `set` / `getOrCreateNested` logic, every `set` call site (`setValue` / `pushValue` / `removeValue` / `setError` / `setFieldTouched`), and `isInteger` usages repo-wide.

**Validation run locally (Node 24):**
- `pnpm vitest run form-store.test.ts misc.test.ts` → **11 passed**.
- `pnpm tsc` → **0 errors**.
- Replicated `set()` standalone and ran a temporary test against the real store to confirm the mixed array-index/non-index behavior; temp test removed (tree clean).

**Assessment:** The core fix is correct and well-targeted. `isArrayIndex` implements the canonical ECMAScript array-index rule (integer, `0 ≤ i ≤ 2³²−2`, round-trips through `String`). I checked the edge cases — `"01"`→false, `"0"`→true, `4294967294`→true (max valid index), `4294967295`→false, `"1e1"` / `" 1"` / `""`→false, numeric `0`→true, `Infinity` / `NaN`→false — all behave as intended and match the tests. The new fresh-store regression tests are solid, and the `Reflect.apply` trick to exercise numeric runtime segments past the `StringLike` type is reasonable. Keeping `isInteger` (now used only by its own test, since it remains public API of `@ariakit/utils`) plus the renamed "loose numeric coercion" test is the right call. Changeset package set (`@ariakit/components` + react-components + react) and patch level look correct.

**Findings (1 inline):**
- One actionable finding on `set` (form-store.ts:93): writing a non-array-index segment (`-1`, `Infinity`, `NaN`, `4294967295`) into a parent that already holds a *populated* array silently converts that array to a plain object (verified against the real store). It's a deliberate consequence of the PR's premise, but it's unpinned by tests and changes behavior for computed-index cases (e.g. `findIndex` → `-1`). CodeRabbit's outside-diff suggestion would reintroduce the malformed array, so it isn't the remedy. Details + a suggested regression test are inline.

**Not fully validated:** I didn't run the full `app` browser suite or audit every published-package consumer; the change is isolated to form-store path handling, so risk outside that surface is low.

packages/ariakit-components/src/form/form-store.ts
diegohaz: Added the regression test in `35e0afea1` to make this mixed case explicit: `users.0.name` creates an array, then `users.-1.name` replaces the container with an object containing both `"0"` and `"-1"` keys. This documents the intended behavior without preserving array expandos.

PR #6203, opened 2026-06-02
Preserve React ref cleanup in merged refs

diegohaz:
## Motivation

React 19 supports callback refs that return cleanup functions. `useMergeRefs` currently forwards all child refs through `setRef`, but `setRef` discards callback ref return values and the merged callback always returns `undefined`.

That means cleanup-returning refs passed through Ariakit's ref merging can leak subscriptions or observers because React never receives the cleanup function.

## Solution

`setRef` now returns a function ref's cleanup result. `useMergeRefs` records each child ref result and returns a merged cleanup only when at least one child returned a cleanup.

When a merged cleanup is returned, React will not call the merged ref with `null` on detach, so the cleanup also explicitly detaches refs that did not return their own cleanup. This preserves React 19 cleanup semantics without changing the no-cleanup path used by React 17/18 and ordinary refs.

`Portal` now preserves cleanup-returning `portalRef` callbacks as well, while still detaching refs that did not return their own cleanup.

The browser sandbox test covers this through actual Ariakit React components and a listener-cleanup scenario that would leak before this fix. It verifies composed `Ariakit.Button` refs, default and connected `Ariakit.Portal` refs, object refs, plain callback refs, cleanup-returning refs, the no-cleanup merged-ref path, and a JavaScript-style callback ref that returns a non-function value.


packages/ariakit-react-utils/src/hooks.ts
```diff
@@ -156,9 +161,28 @@ export function useMergeRefs(...refs: Array<Ref<any> | undefined>) {
...
       }
+
+      if (!refEffects.some((effect) => effect.cleanup)) return;
+
+      return () => {
+        for (const { ref, cleanup } of refEffects) {
+          if (cleanup) {
+            cleanup();
+          } else {
+            setRef(ref, null);
```
diegohaz, line 185: Is setting ref to null here needed?

packages/ariakit-react-utils/src/misc.ts
```diff
@@ -19,9 +19,9 @@ import { isValidElement } from "react";
 export function setRef<T>(
   ref: RefCallback<T> | MutableRefObject<T> | null | undefined,
   value: T,
-) {
+): void | (() => void) {
```
diegohaz, line 24: Do we need explicit return type?

packages/ariakit-react-utils/src/hooks.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { createRequire } from "node:module";
```
diegohaz: We should use an app/src/sandbox test instead, with an actual Ariakit React component. Sandbox tests currently run only in React 19, so we don't need to branch.

diegohaz: Fixed in 374a02740 - moved the coverage to `app/src/sandbox/button-ref-cleanup/test-browser.ts` with actual `Ariakit.Button` and `Ariakit.Portal` cases, and removed the package-level test/dependency. This now uses the browser sandbox test directly.

packages/ariakit-react-utils/src/misc.ts
diegohaz: Kept in daa7dbe71 - the explicit return type is intentional so the generated API docs expose the supported `void | (() => void)` shape instead of leaking React internal void-only helper types.

packages/ariakit-react-utils/src/hooks.ts
diegohaz: Fixed in 374a02740 - added the browser coverage and an inline note for this path. When a merged ref returns a cleanup, React does not call that merged ref with null on detach, so refs without their own cleanup still need the explicit null detachment.

diegohaz:
**Fixed**

- 374a02740 ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342144136)): Added browser coverage and an inline note showing why merged refs still explicitly detach non-cleanup refs with `null`.
- 374a02740 ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342144112)): Moved the regression coverage to `app/src/sandbox/button-ref-cleanup/test-browser.ts` with actual `Ariakit.Button` and `Ariakit.Portal` cases, then removed the package-level test and dependency.

**Already resolved**

- daa7dbe71 ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342144123)): Kept the explicit `setRef` return type so generated API docs expose `void | (() => void)` instead of React internal helper types.

.changeset/300-react-ref-cleanup.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+"@ariakit/react-utils": patch
+---
+
+Fixed [`useMergeRefs`](https://ariakit.com/reference/use-merge-refs) and [`Portal`](https://ariakit.com/reference/portal) to preserve React 19 callback ref cleanup functions while still detaching refs that don't return a cleanup.
```
diegohaz: `useMergeRefs` is not part of `@ariakit/react`'s public API (https://ariakit.com/reference/use-merge-refs doesn't exist). It's better to mention in the changeset only the public-facing changes.

packages/ariakit-react-utils/src/hooks.ts
```diff
@@ -156,9 +161,30 @@ export function useMergeRefs(...refs: Array<Ref<any> | undefined>) {
...
+      if (!refEffects.some((effect) => effect.cleanup)) return;
+
+      return () => {
+        for (const { ref, cleanup } of refEffects) {
+          if (cleanup) {
+            cleanup();
+          } else {
+            // React doesn't call a ref with null when it returns a cleanup.
+            // Refs without their own cleanup still need to be detached.
+            setRef(ref, null);
```
diegohaz, line 185: I mean, if React itself doesn't call the ref with null when it returns a cleanup, why are we doing this here? Is there a good reason?

packages/ariakit-react-components/src/portal/portal.tsx
```diff
@@ -119,14 +119,16 @@ export const usePortal = createHook<TagName, PortalOptions>(function usePortal({
...
-      portalEl.remove();
-      setRef(portalRef, null);
+      if (!isPortalInDocument) {
+        portalEl.remove();
+      }
+      if (typeof cleanup === "function") {
+        cleanup();
+      } else {
+        setRef(portalRef, null);
+      }
```
diegohaz: Why do we need this change?

app/src/sandbox/button-ref-cleanup/index.react.tsx
```diff
@@ -0,0 +1,137 @@
+import * as Ariakit from "@ariakit/react";
```
diegohaz, line 1: I feel like we should have a more real-world use case to test here. One question we should make ourselves is what real-world situation would cause a bug before this fix? Then use it as the sandbox test case.

.changeset/300-react-ref-cleanup.md
diegohaz: Fixed in b16e4abdc - updated the changeset to describe the public-facing surface as merged refs in React components and Portal, without linking or naming the non-public useMergeRefs reference.

app/src/sandbox/button-ref-cleanup/index.react.tsx
diegohaz: Fixed in b16e4abdc - reworked the sandbox around a real listener-cleanup scenario. The test now verifies leaked click listeners would be removed for composed Button refs and Portal refs, includes a no-cleanup merged-ref path, and keeps the non-function portal ref fallback coverage.

packages/ariakit-react-utils/src/hooks.ts
diegohaz: Fixed in b16e4abdc - tightened the comment to clarify that React only sees the merged ref. Because the merged cleanup replaces React’s normal null call for the whole merged ref, child refs that did not return their own cleanup still need the null detach they would receive if they were used alone. The sandbox now also covers both cleanup and no-cleanup merged-ref paths.

diegohaz:
**Fixed**

- b16e4abdc ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342441431)): Updated the changeset to describe public-facing merged refs in React components and Portal without referencing non-public API docs.
- b16e4abdc ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342441444)): Clarified why non-cleanup child refs still need an explicit null detach when the merged ref returns a cleanup.
- b16e4abdc ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342441443)): Reworked the sandbox test around a listener-cleanup leak scenario, covering composed Button refs, Portal refs, the no-cleanup merged-ref path, and the non-function portal fallback.

app/src/sandbox/button-ref-cleanup/index.react.tsx
```diff
@@ -0,0 +1,172 @@
...
+    buttonObjectRef,
+    clickObservedButton: () => observedButton?.click(),
+    clickObservedConnectedPortal: () => observedConnectedPortal?.click(),
+    clickObservedPortal: () => observedPortal?.click(),
+    getActiveListeners,
+    listenerEvents,
+    plainButtonObjectRef,
+    resetListenerEvents,
+  };
+}
```
diegohaz: How is this a real-world scenario? If we need to attach these variables to the global window instead of inspecting actual UI behavior, we are not really testing a real scenario.

packages/ariakit-react-components/src/portal/portal.tsx
diegohaz: You didn't answer this.

diegohaz: **Why this change is needed**

`Portal` drives `portalRef`'s lifecycle by hand — React never sees `portalRef`, so React's own React 19 cleanup mechanism never runs for it. The only place a cleanup-returning `portalRef` can be torn down is this effect's teardown, so the effect has to capture what `setRef(portalRef, portalEl)` returns and call it.

The old code couldn't do that, in two distinct ways:

1. When `isPortalInDocument` is `true`, it returned early with **no teardown at all**, so a cleanup-returning `portalRef` leaked entirely — neither its cleanup nor a `null` detach ever ran.
2. Otherwise it discarded `setRef`'s return value and called `setRef(portalRef, null)` on unmount. For a cleanup-returning ref, being invoked with `null` is **not** the React 19 contract — the cleanup it returned is what should run.

The connected-portal case in the new sandbox demonstrates (1): on `main`, unmounting `<Portal portalElement={connectedEl} portalRef={cleanupRef} />` never runs `cleanupRef`'s cleanup, so its listener leaks.

**One behavior change worth confirming**

Always returning a teardown also changes the `isPortalInDocument === true` path: `portalRef` is now detached on unmount (via the returned cleanup, or `setRef(portalRef, null)` for plain/object refs), whereas before it was left pointing at the still-connected element. That's more consistent — a `portalRef` for an unmounted `Portal` shouldn't dangle — and it's exactly what the connected-portal test now asserts. Just flagging it as a small, edge-case behavior change for anyone passing an already-connected `portalElement` together with a `portalRef`.

If you specifically wanted to keep the old "don't touch the ref when `isPortalInDocument`" behavior for non-cleanup refs, you'd gate only the `setRef(portalRef, null)` branch on `!isPortalInDocument` while still always running the cleanup. But that's asymmetric (cleanup refs detach, plain refs don't), so the current always-detach version reads better and I'd keep it.

app/src/sandbox/button-ref-cleanup/index.react.tsx
diegohaz: Agreed, and the `(window as any).__refCleanupTest` plumbing is the clearest symptom: the spec reaches into module-level globals (`listenerEvents`, `activeListeners`, `observedButton`, …) through ~20 `page.evaluate(() => (window as any)…)` calls instead of asserting on anything the user can see. (`as any` is also against the repo style guide, and here it's forced by this design.)

To anchor on a real scenario, ask "what user-visible thing breaks before this fix?" A faithful version: a callback ref that subscribes to something whose effect shows up in the DOM, then assert the visible result reflects proper cleanup. Sketch:

- A still-mounted log area renders state (a count or list).
- A child Ariakit component gets a callback ref that, on attach, subscribes to a `document` event (or a `ResizeObserver`) and bumps that state; it returns a cleanup that unsubscribes.
- Mount child → dispatch the event → assert the visible count goes up.
- Unmount child → dispatch again → assert the visible count does **not** change.

That last step is the leak: on `main` the dropped cleanup leaves the subscription attached, so the count keeps climbing — the assertion still fails before the fix, but now it fails on rendered UI rather than on a global flag, and no `as any` / `window` bookkeeping is needed.

Two smaller things in the same spirit:

- The exact-order checks like `.toEqual(["external button detach", "button cleanup"])` couple the test to Ariakit's internal ref-merge order (render-ref vs prop-ref). Asserting on observable end-state (which subscriptions remain / what the UI shows) is more robust to internal reordering.
- Credit where due: the current test *is* a valid regression test — the composed `Ariakit.Button` ref and the connected `Ariakit.Portal` ref both leak on `main`, so it does fail before the fix. The ask is realism/maintainability, not correctness.

diegohaz (commented):
## Review summary

Full review (no prior review from this skill existed, so not incremental). Baseline: merge-base `73a885ac2` → head `b16e4abdc`. Checked out the PR head locally and read the changed files plus surrounding context (`setRef`, `useMergeRefs`, `usePortal`, `usePortalRef`, and all `setRef`/`useMergeRefs` call sites).

**Verdict:** the fix is correct and the implementation is clean. I have no blocking findings — only two follow-ups on the existing unresolved threads.

### What I verified
- **Core logic is sound.** `setRef` now forwards the function-ref return; `useMergeRefs` records per-child results and returns a merged cleanup only when at least one child returned one, else `undefined`. The `undefined` path is byte-for-byte the old behavior, so React 17/18 and ordinary/object refs are unchanged. The `{ ref, cleanup }` array (rather than parallel arrays) keeps the teardown loop free of unguarded indexed access — good per the repo style guide.
- **Composition with `usePortalRef` is correct.** Internally `portalRef` is itself a merged ref (`useMergeRefs(setPortalNode, portalRefProp)`); the default Dialog/Popover/Hovercard path has `isPortalInDocument === false` and still calls `setPortalNode(null)` on unmount through both the cleanup and no-cleanup branches. No regression to `domReady`.
- **Blast radius of `setRef`'s new return is contained.** Only the two new sites consume it; every other `setRef(...)` caller ignores the return.
- **The test does catch the bug.** On `main` both the composed `Ariakit.Button` ref and the connected `Ariakit.Portal` ref leak, so the spec fails before the fix — it's a valid regression test.

### Follow-ups (both on existing unresolved threads)
1. **`portal.tsx` — "why do we need this change?"** Replied with the rationale: `Portal` hand-drives `portalRef`'s lifecycle, so React's 19 cleanup never runs for it and the effect must capture and invoke the returned cleanup; the old early-return-on-`isPortalInDocument` dropped it entirely. Also flagged one intentional behavior change to confirm: `portalRef` is now detached on unmount even when `isPortalInDocument` (previously left dangling).
2. **`button-ref-cleanup` sandbox — "not a real-world scenario."** Agreed; replied with a concrete UI-observable redesign (subscribe via a callback ref, assert a visible count rather than `window.__refCleanupTest` globals + ~20 `as any` casts), plus a note that the exact-order `listenerEvents` assertions couple the test to internal ref-merge order.

### Minor / optional (not posted inline)
- The "run cleanup if present, else `setRef(ref, null)`" rule is duplicated between `useMergeRefs` and `Portal`. A tiny shared helper would name the concept once, but the two sites differ slightly (Portal re-checks `typeof cleanup === "function"`), so it's marginal — fine to leave.

### Not independently validated
- **React 18 behavior.** The key compatibility assumption is that React 18 ignores the function the merged ref now returns and still calls the ref with `null` on detach (true — ref cleanups are a React-19 feature). The sandbox runs React 19 only, so this PR adds no dedicated React-18 coverage; it's exercised indirectly by the existing `test-react18` example suite in CI, which is green.
- **Generated `readme.md`.** These signatures are produced by the `docs` workflow, which regenerates and writes back on the PR branch, so any mismatch in the hand-edit is auto-reconciled by CI — I didn't run `pnpm run docs` locally.

packages/ariakit-react-components/src/portal/portal.tsx
diegohaz: Fixed in 79797d454 - added the Portal rationale in code and kept the connected-portal coverage. This change is needed because `portalRef` is assigned by Ariakit inside the Portal effect, not directly by React. Before this PR, `setRef(portalRef, portalEl)` discarded cleanup-returning refs, and connected portals returned early when the DOM node was already connected, so no cleanup/null-detach ran for `portalRef` at all. The effect cleanup now preserves a cleanup-returning `portalRef` and falls back to `setRef(portalRef, null)` for refs without their own cleanup while still leaving connected portal nodes in place.

app/src/sandbox/button-ref-cleanup/index.react.tsx
diegohaz: Fixed in 79797d454 - removed the `window.__refCleanupTest` harness entirely. The sandbox now uses visible shortcut counters and detach labels driven by real `document` keydown listeners, and the browser test verifies the UI before and after unmount with `q.button`/`q.text` instead of inspecting globals.

diegohaz:
**Fixed**

- 79797d454 ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342669367)): Added the Portal cleanup rationale and clarified that `portalRef` is assigned by Ariakit's effect, so it needs explicit cleanup/null detach even for connected portal nodes.
- 79797d454 ([reply](https://github.com/ariakit/ariakit/pull/6203#discussion_r3342669386)): Removed the private `window.__refCleanupTest` harness. The sandbox now verifies real document-key listener cleanup through visible UI counters and detach labels using `q.button`/`q.text`.

app/src/sandbox/button-ref-cleanup/test-browser.ts
```diff
@@ -0,0 +1,57 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("preserves React ref cleanup", async ({ page, q }) => {
+    test.setTimeout(60_000);
+    await test
+      .expect(q.button("Unmount button"))
+      .toBeVisible({ timeout: 30_000 });
```
diegohaz: Do we really need this toBeVisible timeout?

app/src/sandbox/button-ref-cleanup/test-browser.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    await page.keyboard.press("b");
+    await test.expect(q.text("Button shortcut count: 1")).toBeVisible();
+
+    await q.button("Unmount button").click();
+    await test.expect(q.button("Observed button")).not.toBeVisible();
+    await test.expect(q.text("External button detached: yes")).toBeVisible();
+    await test.expect(q.text("Button object ref detached: yes")).toBeVisible();
+    await page.keyboard.press("b");
+    await test.expect(q.text("Button shortcut count: 1")).toBeVisible();
```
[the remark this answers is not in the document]
diegohaz: Fixed in dc1d169e2 - strengthened the browser test so each shortcut count reaches 2 while mounted, then presses the shortcut twice after unmount, waits for two animation frames, and asserts the count stays at 2 while count 3 remains hidden. That prevents the stale-visible-text false pass this thread called out.

packages/ariakit-react-components/src/portal/portal.tsx
```diff
@@ -119,14 +119,18 @@ export const usePortal = createHook<TagName, PortalOptions>(function usePortal({
...
-    // If the portal element was already in the document, we don't need to
-    // remove it when the element is unmounted, so we just return.
-    if (isPortalInDocument) return;
-    // Otherwise, we need to remove the portal from the DOM.
+    const cleanup = setRef(portalRef, portalEl);
     return () => {
-      portalEl.remove();
-      setRef(portalRef, null);
+      if (!isPortalInDocument) {
+        portalEl.remove();
```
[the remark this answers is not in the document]
diegohaz: Fixed in dc1d169e2 - Portal now runs the `portalRef` cleanup/null-detach before removing Ariakit-created portal nodes. I also added sandbox coverage that records `element.isConnected` inside the returned cleanup and asserts `Portal cleanup connected: yes`; the red check with the old remove-before-cleanup order failed on that assertion.

packages/ariakit-react-utils/src/misc.ts
```diff
@@ -19,9 +19,9 @@ import { isValidElement } from "react";
...
   if (typeof ref === "function") {
-    ref(value);
+    return ref(value);
```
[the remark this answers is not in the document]
diegohaz: Fixed in dc1d169e2 - `setRef` now filters callback ref return values and only returns them when they are functions. I added direct `misc.test.ts` coverage for both cleanup-returning refs and JS-style non-function returns; the red check with the old `return ref(value)` behavior failed by returning the DOM node.

app/src/sandbox/button-ref-cleanup/test-browser.ts
diegohaz: Fixed in e754a00b8 - removed both the custom `toBeVisible({ timeout: 30_000 })` and the paired `test.setTimeout(60_000)`. The focused Chrome sandbox test, `pnpm lint`, and `pnpm tsc` pass with the default Playwright timeouts.

diegohaz:
**Fixed**
- dc1d169e2 (https://github.com/ariakit/ariakit/pull/6203#discussion_r3342880847): strengthened the browser counter assertions so post-unmount shortcut presses cannot pass on stale visible text.
- dc1d169e2 (https://github.com/ariakit/ariakit/pull/6203#discussion_r3342882232): moved Portal ref cleanup/null-detach before removing Ariakit-created portal nodes and added connected-state cleanup coverage.
- dc1d169e2 (https://github.com/ariakit/ariakit/pull/6203#discussion_r3342883413): filtered non-function callback ref returns in setRef and added direct utility tests.
- e754a00b8 (https://github.com/ariakit/ariakit/pull/6203#discussion_r3342884441): removed the custom initial browser-test timeout and kept the test on default Playwright timeouts.

diegohaz (commented):
## Re-review summary

Re-review of the commits since my last review. Baseline (my previous marker) `b16e4abdc` → current head `e754a00b8`, covering:

- `79797d454` Refine ref cleanup sandbox coverage
- `dc1d169e2` Address ref cleanup review follow-ups
- `e754a00b8` Remove ref cleanup test timeout

I checked out `e754a00b8` and read the actual changed code rather than trusting the "Fixed in…" replies.

### Both of my prior findings are resolved — verified in code
- **Test realism** (`button-ref-cleanup`): the `window.__refCleanupTest` harness and all `(window as any)` casts are gone. The sandbox now subscribes via callback refs to real `document` keydown listeners and renders visible counters/labels; the spec presses keys, asserts the visible count, then presses twice after unmount and asserts the count does **not** advance (`waitForPaint` guards against a false pass). This is exactly the UI-observable, leak-on-`main` design that was missing, and it no longer couples to internal ref-merge ordering.
- **Portal rationale**: the "why" is now captured in an in-code comment, and the intentional `isPortalInDocument` detach-on-unmount behavior I'd flagged is documented there too.

### Bot findings since my review — also correctly fixed
- **`setRef` non-function returns** (Codex P2): `setRef` now only returns the callback result when it's actually a function, so the `void | (() => void)` return type is honest. Direct `misc.test.ts` covers both the cleanup and non-cleanup paths (runs in the core happy-dom suite — green).
- **Portal cleanup ordering** (Codex P2): the `portalRef` cleanup/null-detach now runs **before** `portalEl.remove()`. Verified correct, and it's a genuine improvement for the *default* portal too, not just connected ones — the cleanup now observes a still-connected node, matching React's ref semantics. The sandbox asserts this via `Portal cleanup connected: yes`, which would read `no` under the old ordering.
- **Test flakiness** (CodeRabbit): each counter reaches 2 while mounted, then two post-unmount presses + paint waits before asserting no advance. Solid.

### New findings
None. The incremental changes are all correct refinements that resolve the open threads; the core logic and its composition with `usePortalRef` remain sound.

### Status / not validated locally
- Green in CI at this head: `Test`, `Test React`, `Test React 18`, `Test Solid`, `Type check`, `Lint`, `Generate docs`, all `Build*`.
- The browser suites (Chrome/Firefox/Safari/Node) that run the reworked `test-browser.ts` were still **pending** when I reviewed — that's the one thing left to confirm green. I did not run the browser sandbox locally (it needs the worktree dev servers); the author reported the focused Chrome run passing.

PR #6206, opened 2026-06-02
Scope composite typeahead buffer to store

diegohaz:
## Motivation

`CompositeTypeahead` used a module-level character buffer. That made typeahead text and cleanup timing leak across unrelated composite widgets rendered on the same page.

At the same time, some Ariakit surfaces install `useCompositeTypeahead` more than once for the same composite store, such as a menubar `MenuButton` and its `MenuList`. Those surfaces still need a shared buffer so a multi-character search can continue after focus moves from the button into the menu.

## Solution

This scopes typeahead state to the `CompositeStore` with a `WeakMap`, storing both the accumulated characters and the cleanup timeout for each store. Separate composite stores now get independent buffers, while multiple typeahead hooks that operate on the same store keep sharing one sequence.

The PR also adds regression coverage for both sides of the behavior:

- a sandbox with two composites proving the second one starts with a fresh buffer;
- a menubar test proving the buffer continues after the first typed character moves focus from the menu button into its menu list.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test-react examples/menubar/test.ts examples/menu/test.ts examples/select/test.ts app/src/sandbox/composite-typeahead-buffer/test.ts`

### DIFF
--- a/app/src/sandbox/composite-typeahead-buffer/test.ts
+++ b/app/src/sandbox/composite-typeahead-buffer/test.ts
@@ -1,6 +1,8 @@
 import { dispatch, press, q } from "@ariakit/test";
 import { expect, test, vi } from "vitest";
 
+// Dispatch only keydown events so this stays compatible with fake timers;
+// @ariakit/test's type helper waits on timer sleeps between key steps.
 function typeahead(key: string) {
   const activeElement = document.activeElement;
   if (!activeElement) throw new Error("No active element");
@@ -14,6 +16,8 @@ function typeahead(key: string) {
 test("keeps typeahead characters scoped to each composite instance", async () => {
   await press.Tab();
 
+  // Keep the first composite's "ap" buffer alive so the old global buffer
+  // would leak into the second composite during the "b" keydown below.
   vi.useFakeTimers();
 
   try {
@@ -25,6 +29,8 @@ test("keeps typeahead characters scoped to each composite instance", async () =>
     await typeahead("p");
     expect(q.button("Apricot")).toHaveFocus();
 
+    // Cherry must not start with "b", otherwise same-initial looping would
+    // reset a leaked buffer and hide the regression.
     q.button.ensure("Cherry").focus();
     expect(q.button("Cherry")).toHaveFocus();
 
--- a/app/src/sandbox/composite-typeahead-buffer/test.ts
+++ b/app/src/sandbox/composite-typeahead-buffer/test.ts
@@ -1,0 +1,20 @@
+import { click, press, q, type } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+test("keeps typeahead characters scoped to each composite instance", async () => {
+  await press.Tab();
+
+  expect(q.button("Alpha")).toHaveFocus();
+
+  await type("a");
+  expect(q.button("Alpine")).toHaveFocus();
+
+  await type("p");
+  expect(q.button("Apricot")).toHaveFocus();
+
+  await click(q.button("Banana"));
+  expect(q.button("Banana")).toHaveFocus();
+
+  await type("b");
+  expect(q.button("Blueberry")).toHaveFocus();
+});
--- a/app/src/sandbox/composite-typeahead-buffer/test.ts
+++ b/app/src/sandbox/composite-typeahead-buffer/test.ts
@@ -1,20 +1,36 @@
-import { click, press, q, type } from "@ariakit/test";
-import { expect, test } from "vitest";
+import { dispatch, press, q } from "@ariakit/test";
+import { expect, test, vi } from "vitest";
+
+function typeahead(key: string) {
+  const activeElement = document.activeElement;
+  if (!activeElement) throw new Error("No active element");
+  return dispatch.keyDown(activeElement, {
+    bubbles: true,
+    cancelable: true,
+    key,
+  });
+}
 
 test("keeps typeahead characters scoped to each composite instance", async () => {
   await press.Tab();
 
-  expect(q.button("Alpha")).toHaveFocus();
+  vi.useFakeTimers();
 
-  await type("a");
-  expect(q.button("Alpine")).toHaveFocus();
+  try {
+    expect(q.button("Alpha")).toHaveFocus();
 
-  await type("p");
-  expect(q.button("Apricot")).toHaveFocus();
+    await typeahead("a");
+    expect(q.button("Alpine")).toHaveFocus();
 
-  await click(q.button("Banana"));
-  expect(q.button("Banana")).toHaveFocus();
+    await typeahead("p");
+    expect(q.button("Apricot")).toHaveFocus();
 
-  await type("b");
-  expect(q.button("Blueberry")).toHaveFocus();
+    q.button.ensure("Cherry").focus();
+    expect(q.button("Cherry")).toHaveFocus();
+
+    await typeahead("b");
+    expect(q.button("Banana")).toHaveFocus();
+  } finally {
+    vi.useRealTimers();
+  }
 });
--- a/.changeset/300-composite-typeahead-buffer.md
+++ b/.changeset/300-composite-typeahead-buffer.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`CompositeTypeahead`](https://ariakit.com/reference/composite-typeahead) so typeahead text from one composite no longer affects another composite rendered on the same page.
--- a/app/src/sandbox/composite-typeahead-buffer/index.react.tsx
+++ b/app/src/sandbox/composite-typeahead-buffer/index.react.tsx
@@ -28,7 +28,7 @@ export default function Example() {
     <>
       <Fixture items={["Alpha", "Alpine", "Apricot"]} label="First composite" />
       <Fixture
-        items={["Banana", "Blueberry", "Blackberry"]}
+        items={["Cherry", "Banana", "Blueberry"]}
         label="Second composite"
       />
     </>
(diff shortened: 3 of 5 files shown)


app/src/sandbox/composite-typeahead-buffer/test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  await type("p");
+  expect(q.button("Apricot")).toHaveFocus();
+
+  await click(q.button("Banana"));
+  expect(q.button("Banana")).toHaveFocus();
+
+  await type("b");
+  expect(q.button("Blueberry")).toHaveFocus();
+});
```
diegohaz, line 42: Did you check if this test fails without our fix?

diegohaz: I checked this empirically, and **this test passes even without the library fix** — as written it doesn't guard the regression.

I reverted only `composite-typeahead.tsx` to the base (module-level `chars`) version, kept this test, and ran it: still green.

The reason is the loop-reset branch in `getSameInitialItems`. After the first composite, the leaked global buffer is `chars = "ap"`. You then `click("Banana")` (so "Banana" is the active item) and `type("b")`. Because the active item *starts with the typed char* (`"Banana"` starts with `"b"`), `getSameInitialItems` hits this branch:

```ts
// If we're looping through the items, we'll want to reset the chars so "oo"
// becomes just "o".
typeaheadState.chars = char;
```

That resets `chars` back to `"b"` and throws away the leaked `"ap"` prefix, so the move to "Blueberry" succeeds whether or not the bug is present.

To actually exercise the leak, the second composite's active item must **not** start with the typed char, so a stale prefix blocks the match:

```diff
       <Fixture
-        items={["Banana", "Blueberry", "Blackberry"]}
+        items={["Banana", "Cherry", "Blueberry"]}
         label="Second composite"
       />
```

```diff
   await click(q.button("Banana"));
   expect(q.button("Banana")).toHaveFocus();

-  await type("b");
-  expect(q.button("Blueberry")).toHaveFocus();
+  await type("c");
+  expect(q.button("Cherry")).toHaveFocus();
```

I verified this variant fails on the base commit (focus stays on "Banana" because the buffer becomes `"apc"`, which matches nothing) and passes on this PR's head.

diegohaz (commented):
**General PR review — initial review**

**Scope / baseline:** Reviewed the diff from merge-base `73a885a` to head `0422e39`, then expanded to the surrounding context: all `useCompositeTypeahead` callers (`menu-button`, `menu-list`, `select`, `select-list`), the changeset, and both new tests. Checked out the head in a clean worktree.

**Verification I ran locally:**
- `pnpm tsc` → 0 errors.
- `pnpm test-react run examples/menubar/test.ts app/src/sandbox/composite-typeahead-buffer/test.ts` → all pass on head.
- Reverted *only* `composite-typeahead.tsx` to the base commit (keeping the new tests) to check whether each new test actually fails without the fix.

**Implementation looks sound.** Scoping the buffer to the `CompositeStore` with a `WeakMap` is a good fit: chars/cleanup are transient UI state (not reactive store state), and the WeakMap key means entries are GC'd with the store. Separate stores get isolated buffers while same-store hooks (menubar button + menu list, select + select list) keep sharing one sequence. The refactors are behavior-preserving — the 4-param `getSameInitialItems` → object param matches the repo style rule, and filtering by `char` instead of the just-reassigned `chars` is equivalent.

Worth calling out as a bonus: moving the cleanup timeout from a per-hook `useRef` into the per-store state also fixes a latent bug. Previously two hooks on the same store had independent `cleanupTimeoutRef`s, so a menu-button hook's pending 500ms timeout was never cleared when the menu-list hook appended a char, and could wipe the buffer mid-search. That's now correct.

**Findings:**

1. **Sandbox regression test doesn't fail without the fix** — `app/src/sandbox/composite-typeahead-buffer/test.ts`. This is exactly the question you raised in your own review thread there; I left the detailed answer + a validated fix on that thread. Short version: I confirmed it passes against the base library code, because typing `"b"` on the active "Banana" item triggers the loop-reset branch in `getSameInitialItems`, which discards the leaked `"ap"` prefix. A one-item change (active item not starting with the typed char) fails on base and passes on head.

2. **Menubar test is a forward-guard, not a fail-without-fix test** — `examples/menubar/test.ts` `typeahead from menu button continues after focus moves to menu` also passes against the base code, because the old global buffer was already shared across the two hooks. It's still worth keeping (it locks in that same-store hooks share the sequence, which would break if the buffer were ever scoped per-hook-instance), but it doesn't validate the cross-composite isolation that motivates the PR. No change requested.

**Not separately validated:** the cleanup-timeout improvement (no fake-timer test across the 500ms boundary), and visual/Playwright behavior (only ran the React unit suites).

Net: the fix and refactor are good and ship-ready; the one substantive issue is that the sandbox test, as written, wouldn't catch a regression of this bug — addressing finding #1 closes that gap.

app/src/sandbox/composite-typeahead-buffer/test.ts
diegohaz: Fixed in bf7e904a3 - I tightened the regression so it fails without the library fix. The second composite now starts on Cherry, uses direct keydown events under fake timers, then types `b` and expects Banana. I verified the current sandbox fixture/test fail against the base `composite-typeahead.tsx` implementation: focus stays on Cherry because the leaked buffer becomes `apb`, and the same test passes on this PR head.

diegohaz:
**Fixed**

- bf7e904a3 (https://github.com/ariakit/ariakit/pull/6206#discussion_r3342742731): tightened the sandbox regression so it fails against the base module-level `CompositeTypeahead` buffer and passes with the per-store buffer fix.

app/src/sandbox/composite-typeahead-buffer/test.ts
diegohaz: Verified — re-checked at head `bf7e904`. I reverted only `composite-typeahead.tsx` back to the base module-level `chars` version (keeping this test) and ran it: it now **fails** (focus stays on "Cherry" because the leaked buffer becomes `apb`, which matches nothing), and it passes on head. The regression guard is real now. 👍

Nice call using fake timers to freeze the 500ms cleanup — that makes the leak deterministic regardless of runner speed, so the guard can't silently degrade on a slow CI box.

app/src/sandbox/composite-typeahead-buffer/test.ts
```diff
@@ -0,0 +1,36 @@
...
+    bubbles: true,
+    cancelable: true,
+    key,
+  });
+}
+
+test("keeps typeahead characters scoped to each composite instance", async () => {
+  await press.Tab();
+
+  vi.useFakeTimers();
```
diegohaz, line 21: Minor / optional (maintainability): the discriminating setup here is load-bearing but implicit, so a plausible future "cleanup" could silently neuter this guard without any test failing. Three things are doing the work and none is documented:

- `vi.useFakeTimers()` freezes the 500ms cleanup so the `"ap"` buffer built in the first composite is guaranteed to still be present when `"b"` is typed in the second one (on a regression).
- the custom `typeahead()` dispatch exists because `type()` isn't fake-timer-safe.
- "Cherry" is the second composite's active item *specifically because it must not start with `"b"`* — if it did, `getSameInitialItems`'s loop-reset branch would discard the leaked prefix and mask the bug (exactly why the original `"Banana"` + `"b"` version passed without the fix).

A one or two line comment capturing the "why" would keep the test from being accidentally weakened later, e.g.:

```ts
// Freeze the 500ms cleanup so the chars typed in the first composite are
// guaranteed to leak into the second one if the fix regresses. The active
// item in the second composite (Cherry) must not start with the typed key,
// otherwise getSameInitialItems would reset the buffer and hide the leak.
vi.useFakeTimers();
```

diegohaz (commented):
**General PR review — re-review**

**Baseline:** Incremental review of the single new commit `bf7e904` ("Tighten composite typeahead regression test") since my prior review at `0422e39`. Only `app/src/sandbox/composite-typeahead-buffer/{test.ts,index.react.tsx}` changed; the library fix, the menubar test, and the changeset are untouched.

**Prior finding is resolved — verified, not just trusted.** My earlier finding was that the sandbox regression test passed even without the library fix. The new commit reworks it: the second composite now starts on "Cherry", keydowns are dispatched directly under fake timers, then `"b"` is typed and "Banana" is expected. I re-ran the base-revert experiment at this head — reverting only `composite-typeahead.tsx` to the base module-level `chars` version now makes this test **fail** (focus stays on "Cherry"; the leaked buffer becomes `"apb"`, which matches nothing), and it passes on head. So the regression guard is genuine now.

Freezing the 500ms cleanup with fake timers is a good robustness call — it makes the leak deterministic regardless of runner speed.

**One minor, optional finding (maintainability):** left an inline note on the test. The three things that make it discriminate — fake timers, the custom dispatch helper, and "Cherry" not starting with `"b"` — are all load-bearing but undocumented, so a future "simplification" could silently neuter the guard. A short comment would protect it. Not a blocker.

**Verification I ran:** `pnpm test-react run app/src/sandbox/composite-typeahead-buffer/test.ts` on head (pass) and against the reverted base library (fail). No other files changed in this commit, so no further re-validation was needed.

Net: the prior blocker is fixed and confirmed; the only remaining feedback is an optional readability nit.

app/src/sandbox/composite-typeahead-buffer/test.ts
diegohaz: Fixed in cd59fb8ee - documented all three load-bearing parts of the regression setup: the keydown-only helper under fake timers, keeping the first composite buffer alive, and why the second active item starts on Cherry instead of a `b` item.

diegohaz:
**Fixed**

- cd59fb8ee (https://github.com/ariakit/ariakit/pull/6206#discussion_r3342772200): documented the fake-timer/keyDown/Cherry setup so the regression test is harder to accidentally weaken.

PR #6207, opened 2026-06-02
Prevent offscreen item props from leaking

diegohaz:
## Motivation

Inactive offscreen `SelectItem` and `ComboboxItem` render through `Role` instead of the full item components. Their public props include several item, command, and focus options that are meant to be consumed by Ariakit before rendering, not forwarded to a plain DOM node.

When users pass props such as `shouldRegisterItem`, `rowId`, `moveOnKeyPress`, `tabbable`, `autoFocus`, or `accessibleWhenDisabled` to an inactive offscreen item, those props can leak into the `Role` render path and trigger React DOM warnings.

## Solution

Strip the inherited non-DOM item props from the inactive offscreen branches before spreading `htmlProps` into `Role` for both `SelectItem` and `ComboboxItem`.

Inactive offscreen placeholders keep the `aria-disabled` state from `useCompositeItemOffscreen`. Custom render elements and callbacks own any native `disabled` state they need, which avoids tag-detection machinery for placeholder nodes while still preserving explicit `disabled` values supplied by the render itself.

Moved the regression coverage into the `select-combobox-offscreen-various` example browser test so it asserts the actual offscreen DOM output and verifies React does not report unknown DOM prop warnings.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm -F website build`
- `pnpm test-chrome examples/select-combobox-offscreen-various/test-chrome.ts -g "offscreen placeholders omit item-only DOM props"`


packages/ariakit-react-components/src/select/select-item-offscreen.test.react.tsx
```diff
@@ -0,0 +1,147 @@
+import { render } from "@ariakit/test/react";
```
diegohaz: This should be a sandbox test with a real-world use case.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
```diff
@@ -38,10 +38,20 @@ export const SelectItem = forwardRef(function SelectItem({
...
     hideOnClick,
     setValueOnClick,
     preventScrollOnKeyDown,
+    moveOnKeyPress,
+    tabbable,
+    clickOnEnter,
+    clickOnSpace,
+    focusable,
+    accessibleWhenDisabled,
+    onFocusVisible,
```
diegohaz, line 55: **`autoFocus` still leaks to the offscreen placeholder**

This list drops the other `Focusable` options (`focusable`, `accessibleWhenDisabled`, `onFocusVisible`) but not `autoFocus`, which is also part of `FocusableOptions` — and it's the one focus option with a real DOM side effect: React calls `.focus()` on the mounted element when `autoFocus` is set.

I verified the leak: rendering an inactive offscreen `SelectItem` with `autoFocus` yields render props that still contain `autoFocus: true` (alongside `aria-disabled`, `data-offscreen`, etc.). For the default `div` placeholder this is a harmless no-op, but with a custom `render` that resolves to a focusable element (e.g. `render={<button />}`) the offscreen placeholder would pull focus to an out-of-view node on mount.

Per the PR's own goal ("focus options ... not forwarded to a plain DOM node"), `autoFocus` belongs in the strip list:

```diff
     focusable,
     accessibleWhenDisabled,
+    autoFocus,
     onFocusVisible,
```

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
```diff
@@ -53,13 +53,23 @@ export const ComboboxItem = forwardRef(function ComboboxItem({
...
+    preventScrollOnKeyDown,
     focusOnHover,
     blurOnHoverEnd,
     moveOnKeyPress,
+    tabbable,
+    clickOnEnter,
+    clickOnSpace,
+    focusable,
+    accessibleWhenDisabled,
+    onFocusVisible,
```
diegohaz, line 73: **`autoFocus` still leaks here too**

Same as the [SelectItem strip block](https://github.com/ariakit/ariakit/blob/aebd3766fba727295d2588cda57993ab15cfd0ea/packages/ariakit-react-components/src/select/select-item-offscreen.tsx#L52-L54): the focus options `focusable`/`accessibleWhenDisabled`/`onFocusVisible` are stripped but `autoFocus` is not. It's the one `Focusable` option with a DOM side effect (React calls `.focus()` on mount), so a custom `render` resolving to a focusable element would steal focus to an offscreen node. Add it here as well:

```diff
     focusable,
     accessibleWhenDisabled,
+    autoFocus,
     onFocusVisible,
```

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
```diff
@@ -38,10 +38,20 @@ export const SelectItem = forwardRef(function SelectItem({
   const {
```
diegohaz, line 38: **Consider a shared strip helper — this denylist is now duplicated and has already drifted**

This is the right fix, but it adds a fourth hand-maintained list of "item option props to remove before rendering the offscreen placeholder", after `collection-item-offscreen.tsx` and `composite-item-offscreen.tsx`. Those lists have diverged, and the divergence is a real leak: I verified that an inactive offscreen **`CompositeItem`** still forwards `disabled, focusable, accessibleWhenDisabled, clickOnEnter, clickOnSpace, shouldRegisterItem, onFocusVisible, autoFocus` straight onto its `<button>` placeholder (the non-DOM ones trigger React "unknown prop" warnings). `CollectionItem`'s offscreen branch has the same gap for the composite-level props.

So this PR fixes the symptom for two of the four offscreen components, while the two that Select/Combobox compose on top of stay exposed.

Extracting a small shared helper / constant for the collection + composite + command + focusable option keys (with the select/combobox-specific keys layered on) would (1) make the lists impossible to drift, (2) make it hard to miss a key like `autoFocus`, and (3) let the same fix cover `CompositeItem`/`CollectionItem` offscreen. Not a blocker for this PR — flagging because the per-component duplication is exactly what allowed the leak to exist.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
```diff
@@ -38,10 +38,20 @@ export const SelectItem = forwardRef(function SelectItem({
...
     store,
     value,
+    disabled,
```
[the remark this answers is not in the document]
diegohaz: I dug into this one. Verdict: acceptable for the default path, narrow for custom renders — leaving the final call to the author.

- The default offscreen placeholder is a `div` (`TagName = "div"`). I confirmed React forwards `disabled` onto a div as `disabled=""` (no warning) — an invalid attribute on a non-form element. Stripping it and relying on the `aria-disabled` that `useCompositeItemOffscreen` already sets is cleaner and correct for the div.
- The regression you describe is real only when a custom `render` resolves to a natively-disableable, focusable element (e.g. `render={<button />}`): there the offscreen placeholder loses native inertness. But that placeholder is intentionally non-registered and handler-less, and `tabbable`/`focusable` are stripped too, so it's a narrow case.

One genuine inconsistency worth resolving either way: `CompositeItem`'s offscreen branch does **not** strip `disabled`, so after this PR Select/Combobox and Composite disagree on whether an offscreen placeholder carries native `disabled`. Not blocking.

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
```diff
@@ -53,13 +53,23 @@ export const ComboboxItem = forwardRef(function ComboboxItem({
   const {
     store,
     value,
+    disabled,
```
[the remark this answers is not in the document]
diegohaz: Same conclusion as the SelectItem thread: for the default `div` placeholder, stripping `disabled` and keeping the hook's `aria-disabled` is correct (React would otherwise render an invalid `disabled=""` on the div). The only regression is a custom `render` resolving to a natively-disableable focusable element, which is a narrow case. Note `CompositeItem` offscreen keeps `disabled`, so it'd be good to make the two consistent. Full reasoning on the SelectItem thread.

packages/ariakit-react-components/src/select/select-item-offscreen.test.react.tsx
diegohaz: +1. The existing `examples/select-combobox-offscreen-various` already exercises this exact offscreen render path (it even branches on `"data-offscreen" in props`), so it's a natural home. A real-world test there could assert on the rendered **DOM** rather than captured render props — e.g. open the popup, grab an offscreen option node, and assert it carries `data-offscreen`/`aria-disabled` but none of the leaked attributes, and that no React "unknown prop" warning is emitted. That would also catch the `autoFocus` leak flagged above, which the current `"x" in props` assertions miss.

Two smaller notes if any unit test survives: it's named `select-item-offscreen.test.react.tsx` but covers both Select and Combobox, and asserting `"x" in props` against the internal render-prop object is brittle compared to asserting real DOM output.

diegohaz (commented):
**Review type:** Initial review (full). **Baseline:** merge-base `73a885ac2` → head `aebd3766fb`. Checked out the PR head in a worktree to inspect full repo context, not just the diff.

**Scope:** the two offscreen prop-strip changes (`select-item-offscreen.tsx`, `combobox-item-offscreen.tsx`), the new React test, and the changeset — plus the surrounding offscreen architecture (`collection-item-offscreen.tsx`, `composite-item-offscreen.tsx`) and the base item hook chain (`useSelectItem`/`useComboboxItem` → `useCompositeItem` → `useCommand` → `useFocusable` / `useCollectionItem` / `useCompositeHover`) to map which option props each layer is supposed to consume.

**Validated locally (Node 24):**
- `pnpm test-react .../select-item-offscreen.test.react.tsx` → passes.
- `pnpm tsc` → clean.
- Probed the offscreen render props directly (temporary throwaway tests, removed): confirmed `autoFocus` still leaks; `disabled` is correctly converted to `aria-disabled`; React renders `<div disabled="">` when `disabled` is *not* stripped; and an inactive `CompositeItem` offscreen still leaks `disabled, focusable, accessibleWhenDisabled, clickOnEnter, clickOnSpace, shouldRegisterItem, onFocusVisible, autoFocus` onto its `<button>`.

**Findings — none blocking:**
1. **`autoFocus` not stripped** in either file (inline on both). It's the one `Focusable` option with a DOM side effect (React `.focus()` on mount), so with a custom `render` resolving to a focusable element the offscreen placeholder could pull focus to an out-of-view node.
2. **Duplicated/divergent denylist** (inline on the select file). This is the 4th hand-maintained strip list; `CompositeItem`/`CollectionItem` offscreen still leak the same props. A shared helper for the composite/collection/command/focusable option keys would prevent drift and complete the fix everywhere.
3. **Codex `disabled` threads** — replied with a verified verdict: stripping is correct for the default `div` (otherwise React emits an invalid `disabled=""`; `aria-disabled` is preserved); the only regression is a custom natively-disableable `render`, which is narrow; but it's now inconsistent with `CompositeItem` offscreen, which keeps `disabled`.
4. **Test approach** — replied on your own thread with concrete guidance: the existing `examples/select-combobox-offscreen-various` already exercises this offscreen path and is a natural home for a real-DOM regression test (which would also catch the `autoFocus` leak that the current `"x" in props` assertions miss).

The core fix is correct and the changeset is appropriate. The changeset lists `@ariakit/react-components`; `@ariakit/react` is picked up as a dependent bump, which matches convention.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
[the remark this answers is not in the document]
diegohaz: Resolved in 7d940a85a - inactive custom SelectItem render paths now preserve native disabled on disabled-capable mounted elements while keeping disabled off default div placeholders. The example browser regression covers this with an offscreen rendered button.

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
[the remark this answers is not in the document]
diegohaz: Resolved in 7d940a85a - inactive custom ComboboxItem render paths now preserve native disabled on disabled-capable mounted elements while keeping disabled off default div placeholders. The example browser regression covers this with an offscreen rendered button.

packages/ariakit-react-components/src/select/select-item-offscreen.test.react.tsx
diegohaz: Fixed in 7d940a85a - removed the package-level render-prop test and moved the regression into examples/select-combobox-offscreen-various/test-chrome.ts. The new browser test asserts the actual offscreen DOM output and checks that React emits no unknown-prop warnings.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
diegohaz: Fixed in 7d940a85a - autoFocus is now stripped from the inactive SelectItem offscreen path and covered by the example browser regression so offscreen rendered buttons do not receive focus on mount.

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
diegohaz: Fixed in 7d940a85a - autoFocus is now stripped from the inactive ComboboxItem offscreen path and covered by the example browser regression so offscreen rendered buttons do not receive focus on mount.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
diegohaz: Partially addressed in 7d940a85a - the shared render/disabled semantics used by SelectItem and ComboboxItem now live in composite utils, and the two offscreen paths have aligned coverage for the inherited item/focus props in scope for this PR. I left the broader CompositeItem/CollectionItem denylist expansion out because the reviewer marked it non-blocking and this PR is scoped to the Select/Combobox T052 leak.

diegohaz:
**Fixed**

- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343650719)): Preserved native `disabled` for disabled-capable custom `SelectItem` offscreen renders.
- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343650859)): Preserved native `disabled` for disabled-capable custom `ComboboxItem` offscreen renders.
- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343650957)): Replaced the package-level test with example browser coverage that asserts actual offscreen DOM and warning absence.
- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343651106)): Stripped `autoFocus` from inactive offscreen `SelectItem` placeholders.
- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343651257)): Stripped `autoFocus` from inactive offscreen `ComboboxItem` placeholders.

**Follow-up**

- `7d940a85a` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343651398)): Shared the render/disabled helpers used by the Select/Combobox offscreen paths; kept the broader CompositeItem/CollectionItem denylist expansion out of this scoped PR because the review thread marked it non-blocking.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
```diff
@@ -30,24 +43,63 @@ export const SelectItem = forwardRef(function SelectItem({
...
+    if (!hasCustomRender) return;
+    if (hasRenderElementProp(props.render, "disabled")) return;
+    const element = tagNameRef.current;
+    if (!element) return;
+    const tagName = element.tagName.toLowerCase();
+    if (!supportsDisabledAttribute(tagName)) {
+      element.removeAttribute("disabled");
+      return;
+    }
+    (element as unknown as DisabledElement).disabled = trulyDisabled;
```
diegohaz: **This layout effect clobbers render-callback `disabled`, and the cast is disallowed**

Two things on this effect (it's duplicated verbatim in `combobox-item-offscreen.tsx`, so both apply there too):

**1. It overwrites disabled state owned by a `render` callback** (confirming the Codex comment below). The `hasRenderElementProp(props.render, "disabled")` guard only inspects *element* renders — `isValidElement` is false for a function, so a callback render falls through. For:

```tsx
<SelectItem render={(props) => <button {...props} disabled={isPending} />} />
```

with the Ariakit item itself **not** disabled, `trulyDisabled` is `false`, and the effect runs `element.disabled = false` after mount on every render — clearing the user's `disabled={isPending}`. I traced this against the current code; it's not covered by the new test, which only exercises the `trulyDisabled === true` path. Safer to only ever *apply* Ariakit's disabled, never clear what the render owns.

**2. `(element as unknown as DisabledElement)` is disallowed** by our code style ("`unknown as T` used to silence type errors"), and it's stronger than needed — `element as DisabledElement` compiles on its own (verified). Better still, `Element.toggleAttribute` is cast-free and lets you drop the `DisabledElement` type and the branching entirely.

Both points fold into one rewrite:

```ts
useSafeLayoutEffect(() => {
  if (active) return;
  if (!hasCustomRender) return;
  if (hasRenderElementProp(props.render, "disabled")) return;
  const element = tagNameRef.current;
  if (!element) return;
  const tagName = element.tagName.toLowerCase();
  // Only apply Ariakit's disabled state; never clear a value the render owns.
  if (trulyDisabled && supportsDisabledAttribute(tagName)) {
    element.toggleAttribute("disabled", true);
  }
});
```

Minor: the effect has no dependency array, so it re-runs on every render of every offscreen item — worth scoping to `[active, hasCustomRender, trulyDisabled, props.render]` once the body settles.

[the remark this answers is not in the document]
diegohaz: Confirmed — real bug, verified against the current code. The `hasRenderElementProp` guard only catches *element* renders (`isValidElement` is false for a function), so `render={(props) => <button {...props} disabled={isPending} />}` falls through and the effect runs `element.disabled = trulyDisabled` (`false`) after mount, clearing the user's value. The new Playwright test doesn't catch it because it only exercises the `trulyDisabled === true` path. Combined fix (also drops the disallowed `as unknown as` cast): https://github.com/ariakit/ariakit/pull/6207#discussion_r3343737840

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
```diff
@@ -45,26 +57,65 @@ export const ComboboxItem = forwardRef(function ComboboxItem({
...
+    if (!hasCustomRender) return;
+    if (hasRenderElementProp(props.render, "disabled")) return;
+    const element = tagNameRef.current;
+    if (!element) return;
+    const tagName = element.tagName.toLowerCase();
+    if (!supportsDisabledAttribute(tagName)) {
+      element.removeAttribute("disabled");
+      return;
+    }
+    (element as unknown as DisabledElement).disabled = trulyDisabled;
```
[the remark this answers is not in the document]
diegohaz: Same confirmed bug as the [SelectItem thread](https://github.com/ariakit/ariakit/pull/6207#discussion_r3343737840) — this effect is duplicated here verbatim. Fix: only apply Ariakit's disabled, never clear what a `render` callback owns.

examples/select-combobox-offscreen-various/index.react.tsx
```diff
@@ -31,6 +31,50 @@ function groupItems(items: ReturnType<typeof getItem>[]) {
...
+    tabbable: true,
+    clickOnEnter: false,
+    clickOnSpace: false,
+    onFocusVisible: () => {},
+  };
+  if (value === "Yemen") {
+    return {
+      ...testProps,
+      "data-testid": `${prefix}-offscreen-disabled`,
+      disabled: true,
```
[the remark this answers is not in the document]
diegohaz: Agree, and worth being precise about the fix. The `<button/>` render is already gated on `"data-offscreen" in props`, but the *top-level* props (`disabled`, `shouldRegisterItem: false`, `tabbable`, …) are inputs the component consumes before it decides active-vs-offscreen, so they can't be gated on offscreen state at the call site. Spreading them unconditionally means a search for "Yemen"/"Zambia"/"Zimbabwe" promotes that option to active with `disabled` + `shouldRegisterItem: false` — an active, unregistered, unselectable match, not just a cosmetic change. Since this is a published example, a dedicated fixture for these probe items (rather than overloading the shared multi-variant demo) would avoid the user-visible regression. Blast radius today is limited to those three end-of-list countries.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
diegohaz: Thanks — acknowledging the new helpers in `composite/utils.ts`. To be precise about what's still open from this thread: the extracted helpers are the render/disabled introspection ones, but the actual *denylist* (the ~20-key destructure) plus the new ~25-line disabled-handling block (`tagNameRef` / `renderTagName` / `hasCustomRender` / `trulyDisabled` / the layout effect / `offscreenProps` / `disabledProps`) are now duplicated **verbatim** across `select-item-offscreen.tsx` and `combobox-item-offscreen.tsx` — more shared logic to keep in sync, not less. A single hook (e.g. `useOffscreenItemProps`) returning `{ offscreenRef, htmlProps, disabledProps }` would collapse both copies and remove a class of drift bugs.

Still unaddressed (same root cause): `CompositeItem` and `CollectionItem` offscreen branches don't strip `autoFocus`/the focus+command props and don't do this disabled handling, so they still leak onto their placeholders. A shared hook would naturally cover those two as well.

diegohaz (commented):
**Re-review.** Baseline: my prior review head `aebd3766fb` → current head `7d940a85ab` ("Address offscreen item review feedback"). Reviewed the incremental diff plus the surrounding offscreen architecture.

**Prior findings — verified resolved:**
- `autoFocus` is now stripped in both offscreen files. ✅ (confirmed in code; the new example test asserts the offscreen item is `not.toBeFocused()`).
- The package-level render-prop unit test was removed and replaced with a real-DOM Playwright test in `examples/select-combobox-offscreen-various` — exactly the kind of test that was requested. It asserts `data-offscreen`/`aria-disabled`, native `disabled` on supporting custom renders, no leaked attributes, and **no React unknown-prop warnings**. ✅
- The Codex `disabled` concern was handled (not just deferred): native `disabled` is now preserved for custom renders that resolve to a disable-supporting element, while the default `div` stays `aria-disabled`-only. ✅

**New findings on this commit:**
1. **(Bug, untested) The new `useSafeLayoutEffect` clobbers render-callback-owned `disabled`.** For `render={(props) => <button {...props} disabled={isPending} />}` with the Ariakit item itself not disabled, the effect sets `element.disabled = false` after mount, clearing the user's value — the `hasRenderElementProp` guard only covers element renders, not callbacks. Confirmed Codex's comment by tracing; the new test only exercises the `trulyDisabled === true` path so it doesn't catch this. Inline with a combined fix.
2. **(Style + simplification) `(element as unknown as DisabledElement)`** violates our "no `unknown as T`" rule and is unnecessary — `element as DisabledElement` compiles, and `element.toggleAttribute("disabled", …)` is cast-free and removes the `DisabledElement` type and a branch (verified by isolated compile). Folded into finding 1's rewrite.
3. **(Maintainability) Duplication grew.** The ~20-key denylist and the new ~25-line disabled-handling block are now duplicated verbatim across both offscreen files; `CompositeItem`/`CollectionItem` offscreen still leak the same props and lack the disabled handling. A shared `useOffscreenItemProps`-style hook would collapse the copies and cover all four. Replied on the existing thread.

**Other bot findings I evaluated:**
- Codex "test-only props leak into active demo items" — **valid**; replied with the precise constraint (top-level item props can't be gated on offscreen state, so a dedicated fixture is the realistic fix). Limited blast radius (3 end-of-list countries).
- CodeRabbit "`supportsDisabledAttribute` omits fieldset/optgroup/option" — technically correct but very low practical relevance for an item render; optional.

**Validation (Node 24):** `pnpm tsc` clean, `pnpm lint` clean (note: lint does not flag the `as unknown as` — it's a style-guide rule). All PR CI checks are green on this head, including Test Chrome (runs the new `test-chrome.ts`), Test React/React18, Firefox, Safari. Verified both cast alternatives compile in isolation. Did not re-run the browser test locally (green in CI; avoided a worktree Playwright setup).

**Bottom line:** the leak fixes and the new test are solid and CI is green, but the new disabled layout-effect introduced a real (currently untested) regression for `render` callbacks — worth fixing before merge, ideally alongside de-duplicating the two copies.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
diegohaz: Fixed in 8f4705c9d - removed the duplicated layout-effect/cast path entirely and centralized disabled normalization in composite utils. The regression now covers callback-owned disabled, disabled={props.disabled}, disabled={undefined} element renders, custom div renders, and early console warnings.

packages/ariakit-react-components/src/combobox/combobox-item-offscreen.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 8f4705c9d - removed the ComboboxItem layout-effect mutation and now apply native disabled declaratively without clearing disabled state owned by render callbacks.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -69,3 +70,28 @@ export function isItem(
...
+  return hasOwnProperty(render.props as Record<string, unknown>, prop);
+}
+
+export function supportsDisabledAttribute(tagName?: string) {
+  return (
+    tagName === "button" ||
+    tagName === "input" ||
+    tagName === "select" ||
+    tagName === "textarea"
+  );
```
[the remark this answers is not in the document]
diegohaz: Fixed in 8f4705c9d - added the missing native disabled targets (fieldset, optgroup, option), aligned Focusable support, and kept implicit native disabled limited to known supported render targets.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 8f4705c9d - removed the SelectItem layout-effect mutation and now apply native disabled declaratively without clearing disabled state owned by render callbacks.

examples/select-combobox-offscreen-various/index.react.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 8f4705c9d - moved the probe props into dedicated offscreen fixture sections so the public country-list demo items no longer receive disabled, shouldRegisterItem, or other test-only props.

diegohaz:
**Fixed**

- 8f4705c9d ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344321943)): Added `fieldset`, `optgroup`, and `option` native disabled support, aligned `Focusable`, and limited implicit native disabled to known supported render targets.
- 8f4705c9d ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344321944)): Removed the `SelectItem` layout-effect mutation so render callbacks can keep their own `disabled` state.
- 8f4705c9d ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344321942)): Removed the `ComboboxItem` layout-effect mutation with the same declarative disabled handling.
- 8f4705c9d ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344321978)): Moved offscreen probe props into dedicated fixtures so active country-list demo items no longer receive test-only props.
- 8f4705c9d ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344321936)): Centralized offscreen disabled normalization in composite utils, removed the disallowed cast/effect path, and expanded the browser regression for callback-owned disabled, `disabled={props.disabled}`, `disabled={undefined}` element renders, custom div renders, and early console warnings.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -87,7 +87,10 @@ function supportsDisabledAttribute(tagName?: string) {
   if (!tagName) return true;
   return (
     tagName === "button" ||
+    tagName === "fieldset" ||
```
diegohaz: **This change is outside the PR's scope and alters shared `Focusable` behavior**

This `supportsDisabledAttribute` lives in `focusable.tsx` and feeds `supportsDisabled` (line 415), which drives both `getTabIndex(...)` and `disabled: supportsDisabled && trulyDisabled ? true : undefined` (line 457) for **every** Focusable-derived component — `Button`, `Command`, `CompositeItem`, `MenuItem`, etc., not just offscreen items. Adding `fieldset`/`optgroup`/`option` here means a disabled Focusable rendered as one of those tags now gets a native `disabled` attribute (and different `tabIndex`) instead of `aria-disabled` + `pointer-events: none`. For `<fieldset disabled>` that natively disables all descendant controls — a real behavior change.

Importantly, the offscreen leak fix doesn't depend on this: the inactive path goes through `getRenderWithDisabled` → `supportsDisabledRender` → the **separate** `supportsDisabledAttribute` in `composite/utils.ts`. This `focusable.tsx` copy is never reached by the offscreen code.

If the intent is a general correctness improvement (these tags do support native `disabled` per spec), it's defensible — but it deserves its own commit, a test for the active `Focusable`/`Command` path, and a line in the changeset, because it changes shipped behavior for existing consumers. If it's only here to keep the two helpers in sync, I'd drop it from `focusable.tsx` and keep the scope limited to the offscreen fix. Either way, the two identical `supportsDisabledAttribute` copies (here and `composite/utils.ts`) are worth consolidating.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -69,3 +72,57 @@ export function isItem(
...
+export function getRenderWithDisabled(
+  render: RenderProp | ReactElement | undefined,
+  disabled: boolean,
+) {
+  if (!disabled) return render;
+  const getElementWithDisabled = (
+    element: ReactElement<Record<string, unknown>>,
+  ) => {
+    if (hasRenderElementProp(element, "disabled")) return element;
+    if (!supportsDisabledRender(element)) return element;
```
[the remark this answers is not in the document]
diegohaz: Confirmed real, but it's an inherent limit of doing this declaratively, and the blast radius is tiny. `getRenderWithDisabled` can only see the element it's handed; for `render={<MyButton />}` (or a callback returning `<MyButton />`) `getRenderElementTagName` is `undefined`, so it can't know `MyButton` resolves to a `<button>` without rendering it. The previous commit's layout-effect *could* see the resolved DOM node, but that's exactly the approach that introduced the callback-clobber bug — so this is a deliberate trade-off, not a regression to chase further.

Worth weighing against real usage: this only bites when someone (1) uses the offscreen item variant, (2) marks a specific item `disabled`, (3) renders it through a *custom component* that forwards `disabled` to a native control, and (4) expects the **out-of-view, unregistered** placeholder to be natively inert. I'd lean toward documenting "custom component renders manage their own disabled state on the offscreen placeholder" rather than adding more machinery. I'm posting a separate PR comment analyzing all of these disabled/leak use cases by real-world likelihood.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
```diff
@@ -31,23 +33,40 @@ export const SelectItem = forwardRef(function SelectItem({
     ...props,
   });
   const allProps = { ...rest, ...props, ref: useMergeRefs(ref, props.ref) };
+  const trulyDisabled =
+    props.focusable !== false &&
+    disabledFromProps(allProps) &&
+    !props.accessibleWhenDisabled;
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and it's consistent by design rather than a clear bug. For an **active** item, `accessibleWhenDisabled` keeps the control focusable on purpose and `useFocusable` suppresses click/keydown so it's inert-but-reachable. The offscreen placeholder renders straight through `Role` with no `useFocusable`, so it gets `aria-disabled` but not the event suppression — a custom `render={<button />}` placeholder stays tabbable and clickable.

But this is the placeholder for an item that is **off-screen and unregistered**, with all of its handlers already stripped — clicking it does nothing and it's not in the visible viewport. Replicating `useFocusable`'s suppression in the offscreen branch would be a lot of machinery for a node the user can't see. My read is this is acceptable as-is; flagging it in the use-case analysis comment so we can decide explicitly rather than reflexively fix it.

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
diegohaz: Verified fixed in 8f4705c9d. `getRenderWithDisabled` returns `render` untouched when the item isn't `trulyDisabled` (`if (!disabled) return render;`), so a callback like `render={(props) => <button {...props} disabled={isPending} />}` on a non-disabled item is no longer overwritten. The declarative `cloneElement` approach also drops the `as unknown as` cast and the per-render layout effect. 👍

diegohaz (commented):
**Re-review** (third pass). Baseline: my last marked review at `7d940a85ab`; the branch was since rebased and a new commit added — `8f4705c9dc` "Handle offscreen disabled render edge cases". Reviewed that increment plus the surrounding shared code it now touches.

**Verified resolved from the last round:**
- The callback-clobber bug + the `as unknown as` cast are gone — the imperative `useSafeLayoutEffect` was replaced with a declarative `getRenderWithDisabled` that returns `render` untouched when the item isn't `trulyDisabled`, so it no longer overwrites a render callback's own `disabled`. ✅
- CodeRabbit's `fieldset`/`optgroup`/`option` gap and the "test-only props on active demo items" concern were both addressed.

**New findings this pass:**
1. **(Scope / shared behavior) `focusable.tsx`** now adds `fieldset`/`optgroup`/`option` to its own `supportsDisabledAttribute`, which changes native-`disabled` + `tabIndex` selection for **all** Focusable-derived components — and is never reached by the offscreen path (that uses the separate copy in `composite/utils.ts`). Inline: either promote it to its own commit + test + changeset line, or drop it to keep this PR scoped.
2. **(Diminishing returns) Two new Codex edge cases** — disabled not forwarded to *component* renders (`render={<MyButton/>}`), and `accessibleWhenDisabled` placeholders staying tabbable/clickable — are both **real but extremely narrow** (offscreen + custom focusable render + disabled). Replied on both confirming, and recommending we decide explicitly rather than keep adding machinery for out-of-view placeholders.

**Validation (Node 24):** `pnpm tsc` clean, `pnpm lint` clean. I also measured the **pre-PR symptom matrix** empirically (rendering offscreen items with each newly-stripped prop and capturing React dev warnings + leaked DOM attributes) — feeding a separate, non-review PR comment that walks each fix as a real-world consumer use case so we can judge which are worth keeping. Did not re-run the browser suite locally (green in CI on this head).

**Overall:** the core leak fix and the clobber fix are solid; my main open question is whether the disabled-on-custom-render handling (and the `focusable.tsx` change) is solving problems real consumers actually hit — addressed in the follow-up comment.

diegohaz:
## Real-world use cases behind this PR — which leaks actually reach consumers?

This is a catalogue (not a review) of every distinct bug this PR fixes, framed as the consumer code that triggers it, so we can decide which are worth fixing. I measured each symptom empirically against the pre-PR base.

### The one fact that frames all of it: reach

None of these bugs affect the components most people use. `Ariakit.SelectItem` / `Ariakit.ComboboxItem` from `@ariakit/react` are the **base** components and were never affected. The bug only exists in the **offscreen virtualization variants**, which are *undocumented deep imports* from the lower-level package:

```tsx
// Standard — NOT affected:
import { SelectItem } from "@ariakit/react";

// Affected — opt-in deep import, no docs page, used in 1 repo example:
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";
import { ComboboxItem } from "@ariakit/react-components/combobox/combobox-item-offscreen";
```

So to hit *any* leak below, a consumer must do **all** of:
1. discover and deep-import the offscreen variant from `@ariakit/react-components` (not `@ariakit/react`),
2. enable `offscreenMode="passive" | "lazy"` with an `offscreenRoot`,
3. have a list long enough that items are actually scrolled out of view, **and**
4. pass the specific leaking prop to an item.

When an item is offscreen it renders a placeholder `<div data-offscreen aria-disabled=… />` instead of the real component, and the leaking props were being spread onto that `<div>`.

### Three symptom classes (this matters for "do users notice?")

- **Dev-only console warnings** — React's "does not recognize the prop" / "unknown event handler" warnings. These fire in development only and are **stripped from production builds**. End users never see them; a developer sees console noise.
- **Stray DOM attributes** — `disabled=""`, `focusable="false"`, `rowid="…"` land on the placeholder div. Invalid but inert; nothing reads them; no end-user effect.
- **Behavioral** — only the `disabled` + custom focusable `render` path can produce an actually-interactive placeholder, and even then it's an **out-of-view, unregistered** node with its handlers already stripped.


---

## The use cases, by likelihood

### 🟢 Plain disabled option in a long list — `disabled`
**Likelihood:** Realistic · **Symptom:** Cosmetic / invisible

Disabling specific options is the single most common reason to touch item props, and offscreen lists are by definition long, so a disabled option that happens to be scrolled out of view is genuinely common for anyone using this feature. But the offscreen placeholder is a non-interactive, unregistered `<div>` that's off the screen, and the offscreen hook already stamps `aria-disabled="true"` on it independent of any leak. The leaked `disabled=""` attribute is invalid on a `<div>` but inert — it changes nothing visually, semantically, or behaviorally.

**Consumer code that triggers it:**
```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

<Ariakit.SelectPopover ref={listRef} className="overflow-auto max-h-72">
  {countries.map((country) => (
    <SelectItem
      key={country}
      value={country}
      disabled={embargoed.has(country)}
      offscreenMode="passive"
      offscreenRoot={listRef}
    />
  ))}
</Ariakit.SelectPopover>
```

**What the developer actually sees (pre-PR):** No React warning. The placeholder renders `<div disabled="" aria-disabled="true" data-offscreen ...>`. No console output in any build; nothing visible to end users.

**Verdict:** Not worth fixing on its own merit — the symptom is an inert, invisible attribute and the accessibility semantics were already correct via `aria-disabled`. Cleaning it up is tidy but affects nobody, dev or end user.

### 🟡 Disabled option with a custom focusable render — `disabled` + `render={<button/>}`
**Likelihood:** Plausible but rare · **Symptom:** Broken behavior (genuine, but only while offscreen)

A consumer who renders options as real `<button>`s (e.g. to get native disabled styling/semantics) and disables some of them is plausible. This is the one case in the use-case set with a real defect: pre-PR the offscreen placeholder rendered an enabled `<button>` (the `disabled` prop was stripped and never re-applied), so a disabled option that scrolled into the 40% `rootMargin` band — or was reachable by tab before upgrading — could momentarily be a focusable, clickable button that doesn't look disabled. The PR's `getRenderWithDisabled` re-applies a native `disabled` attribute for disableable tags, closing this. Still, the window is narrow: the placeholder is unregistered, so arrow-key composite navigation skips it, and it upgrades to the real disabled item the moment it's actually on screen.

**Consumer code that triggers it:**
```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

<Ariakit.SelectPopover ref={listRef} className="overflow-auto max-h-72">
  {plans.map((plan) => (
    <SelectItem
      key={plan.id}
      value={plan.id}
      disabled={!plan.available}
      offscreenMode="passive"
      offscreenRoot={listRef}
    />
  ))}
</Ariakit.SelectPopover>
```

**What the developer actually sees (pre-PR):** No React warning. The offscreen placeholder is an enabled `<button>` (only `aria-disabled="true"` present, no native `disabled`), so it's tab-focusable/clickable while in the offscreen state.

**Verdict:** This is the legitimately load-bearing part of the PR — a real, if narrow, behavior bug for the button-render pattern. Worth fixing, though the practical exposure is small because the placeholder is offscreen and unregistered.

### 🟡 Keeping a disabled option focusable — `accessibleWhenDisabled`
**Likelihood:** Plausible but rare · **Symptom:** Dev-only console warning

`accessibleWhenDisabled` is a real Ariakit pattern (keep a disabled option focusable so users can discover it / read a tooltip), so pairing it with `disabled` is something a thorough consumer does. Pre-PR this prop leaked onto the placeholder `<div>`, producing a React dev warning. The PR strips it. Note the PR explicitly does **not** fix the underlying semantics: an `accessibleWhenDisabled` placeholder still doesn't get native `disabled` (by design — `trulyDisabled` is false) and stays tabbable, so the only thing actually addressed here is the console noise.

**Consumer code that triggers it:**
```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

<Ariakit.SelectPopover ref={listRef} className="overflow-auto max-h-72">
  {features.map((f) => (
    <SelectItem
      key={f.id}
      value={f.id}
      disabled={!f.enabled}
      accessibleWhenDisabled
      offscreenMode="passive"
      offscreenRoot={listRef}
    />
  ))}
</Ariakit.SelectPopover>
```

**What the developer actually sees (pre-PR):** React DEV-only warning: "React does not recognize the `accessibleWhenDisabled` prop on a DOM element." The prop is not added to the DOM. Production builds strip this warning, so end users see nothing.

**Verdict:** Worth fixing only as console-noise cleanup — it's a developer-experience nit with zero end-user impact. The behavior the prop name implies is intentionally left uncovered by this same PR.

### 🟠 Disabled option rendered via a custom component — `disabled` + `render={<MyButton/>}`
**Likelihood:** Effectively never · **Symptom:** Broken behavior (uncovered gap)

This is the known gap the PR does **not** close: when the render target is a custom component (`<MyButton/>` that internally renders a `<button>`), `getRenderElementTagName` sees a function type and returns undefined, so `supportsDisabledRender` is false and no native `disabled` is applied to the placeholder. To hit it a consumer must combine the undocumented offscreen deep-import, a long enough list, a *disabled* item, AND wrap it in a custom component instead of an intrinsic element — a vanishingly rare stack, and the example file only exercises it as a `<div>`-rendering probe (`OffscreenProbeDiv`), not a real button.

**Consumer code that triggers it:**
```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

const MyButton = forwardRef<HTMLButtonElement, ComponentPropsWithoutRef<"button">>(
  (props, ref) => <button ref={ref} className="my-option" {...props} />,
);

<SelectItem
  value={plan.id}
  disabled={!plan.available}
  offscreenMode="passive"
  offscreenRoot={listRef}
  render={<MyButton />}
/>
```

**What the developer actually sees (pre-PR and post-PR):** No React warning. The placeholder renders `MyButton` without a native `disabled` attribute (only `aria-disabled` from the hook), so it stays focusable/clickable while offscreen — same defect as the intrinsic-`<button>` case but not fixed.

**Verdict:** Not worth chasing in this PR. It's the same class of bug as the fixed `<button>` case but for an audience so narrow it's effectively theoretical; documenting it as a known limitation (as the PR does) is the right call rather than over-engineering tag detection through arbitrary components.


### 🧩 Grid composites — `rowId`
**Likelihood:** Effectively never · **Symptom:** Dev-only console warning (plus a cosmetic `rowid` attribute leak)

`rowId` only matters for 2D grid composites (`SelectRow`/`ComboboxRow` with `role="grid"`/`gridcell"`), where it ties an item to its row. But Ariakit derives `rowId` automatically from the surrounding `Row` context — the canonical grid examples (`examples/combobox-grid`, `examples/select-grid`) never pass it. A consumer would only set it explicitly when building rows imperatively without the `Row` wrapper (e.g. a flat, manually-positioned grid). Stacking that rare pattern on top of the already-niche offscreen variant — *and* virtualizing a 2D grid item-by-item, which the offscreen feature isn't really shaped for — makes this combination essentially theoretical.

**Consumer code that triggers it:**
```tsx
import { useRef } from "react";
// offscreen variant: deep import, not Ariakit.ComboboxItem
import { ComboboxItem } from "@ariakit/react-components/combobox/combobox-item-offscreen";

// Flat grid built without <ComboboxRow>, so rowId is assigned by hand.
function EmojiGrid({ emojis }: { emojis: { id: string; row: string }[] }) {
  const listRef = useRef(null);
  return (
    <Ariakit.ComboboxProvider>
      <Ariakit.Combobox />
      <Ariakit.ComboboxPopover ref={listRef} role="grid" className="overflow-auto max-h-72">
        {emojis.map((emoji) => (
          <ComboboxItem
            key={emoji.id}
            role="gridcell"
            value={emoji.id}
            rowId={emoji.row}
            offscreenMode="passive"
            offscreenRoot={listRef}
          />
        ))}
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

**What the developer actually sees (pre-PR):** In a React DEV build, a console warning "React does not recognize the `rowId` prop on a DOM element," and the placeholder renders a stray `rowid="..."` attribute on the `<div>` (string values render, unlike boolean/non-DOM props). Production builds strip the warning; the leaked `rowid` attribute persists but is inert (no styling/behavior keys off it). End users notice nothing.

**Verdict:** Not worth fixing on its own merits — the triggering combination (manual `rowId` + offscreen variant + virtualized 2D grid) is one almost no consumer will ever assemble, and even then the impact is dev console noise plus an invisible attribute. It's only justified as a free side effect of the offscreen item reusing the shared `composite-item-offscreen` strip list, where dropping `rowId` costs nothing.


### 🟡 Conditional item registration — `shouldRegisterItem`
**Likelihood:** Effectively never · **Symptom:** Dev-only console warning

`shouldRegisterItem={false}` keeps an item rendered but out of the composite store and keyboard navigation — the realistic use is a decorative/placeholder row (e.g. a "no results" or "loading" line) you want to display but not navigate to. But that is fundamentally at odds with the offscreen feature: virtualization exists for long, uniform, navigable lists, and an offscreen placeholder is *already* not registered as a navigable store item, so passing `shouldRegisterItem={false}` to it is redundant. To leak, a consumer would have to deep-import the offscreen variant, enable `offscreenMode` on a list long enough to push items offscreen, AND pass `shouldRegisterItem` to one of those items — yet a decorative row is exactly the kind of one-off you would never wrap in the offscreen variant. In the whole repo this prop only appears in the synthetic probe harness (`getOffscreenProbeProps`), never in any organic usage.

**Consumer code that triggers it:**
```tsx
import { useRef } from "react";
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

function CountrySelect({ countries }: { countries: string[] }) {
  const listRef = useRef(null);
  return (
    <Ariakit.SelectProvider defaultValue="">
      <Ariakit.Select />
      <Ariakit.SelectPopover ref={listRef} className="overflow-auto max-h-72">
        {countries.map((country) => (
          <SelectItem
            key={country}
            value={country}
            // e.g. tries to keep a "soft-hidden" row visible but unnavigable
            shouldRegisterItem={country !== "__divider__"}
            offscreenMode="passive"
            offscreenRoot={listRef}
          />
        ))}
      </Ariakit.SelectPopover>
    </Ariakit.SelectProvider>
  );
}
```

**What the developer actually sees (pre-PR):** React DEV console warning: "React does not recognize the `shouldRegisterItem` prop on a DOM element." The value is NOT written to the DOM as an attribute, and the warning is stripped from production builds, so end users never see anything (no broken behavior, no visible markup).

**Verdict:** Not worth fixing on its own merits — the scenario is essentially incoherent (you would not offscreen-virtualize a decorative non-registered row), and even if hit it produces only dev console noise with zero end-user or behavioral impact. It is fine to strip as part of the broad "don't spread item props onto the placeholder" cleanup, but it does not target a problem real consumers actually hit.


### ⌨️ Per-item keyboard/interaction tuning — `moveOnKeyPress`, `preventScrollOnKeyDown`, `clickOnEnter`, `clickOnSpace`
**Likelihood:** Effectively never · **Symptom:** Dev-only console warning

These are advanced per-item interaction toggles, and in practice they are essentially never set per-item. `clickOnEnter`/`clickOnSpace` control whether Enter/Space synthesize a click on the item; `moveOnKeyPress` controls arrow-key movement; `preventScrollOnKeyDown` suppresses native scroll on arrow keys. Real consumers leave these at their defaults — the few who touch interaction behavior do it once at the composite/store level, not by passing the same flag to every item in a `.map()`. To leak one a consumer would have to (a) deep-import the offscreen variant, (b) hand-set one of these advanced flags on individual items, and (c) have that item scrolled out of view — a combination that does not occur in any real codebase and isn't shown in any docs. Note the matrix also splits the leak: `moveOnKeyPress` leaked on SELECT offscreen only (combobox already stripped it) and `preventScrollOnKeyDown` on COMBOBOX offscreen only (select already stripped it), so even the broken surface was half-covered before the PR.

**Consumer code that triggers it:**
```tsx
import { useRef } from "react";
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

function CountrySelect({ countries }: { countries: string[] }) {
  const listRef = useRef(null);
  return (
    <Ariakit.SelectProvider defaultValue="">
      <Ariakit.Select />
      <Ariakit.SelectPopover ref={listRef} className="overflow-auto max-h-72">
        {countries.map((country) => (
          <SelectItem
            key={country}
            value={country}
            // contrived per-item interaction tuning on a long, virtualized list:
            clickOnEnter={false}
            clickOnSpace={false}
            moveOnKeyPress={false}
            offscreenMode="passive"
            offscreenRoot={listRef}
          />
        ))}
      </Ariakit.SelectPopover>
    </Ariakit.SelectProvider>
  );
}
```

**What the developer actually sees (pre-PR):** In a React DEV build, a console warning per offscreen item — e.g. `Warning: React does not recognize the 'clickOnEnter' prop on a DOM element.` (likewise for `clickOnSpace`, and `moveOnKeyPress` on select / `preventScrollOnKeyDown` on combobox). None of these render a real DOM attribute, and all warnings are stripped from production builds, so end users see nothing — not even cosmetically.

**Verdict:** Not worth fixing on its own merits — the trigger requires an implausible authoring pattern (advanced interaction flags set per-item on a virtualized list), and the worst case is dev-only console noise with zero end-user impact. It's defensible only as cheap completeness/consistency: the offscreen items already strip most sibling item props, and these four are caught by the same destructure, so excluding them costs almost nothing. As a standalone justification, this is an over-fix.


### 🪧 Focusability & tabbing on offscreen items — `tabbable`, `focusable`
**Likelihood:** Effectively never · **Symptom:** Dev-only console warning (`tabbable`) / Cosmetic / invisible (`focusable`)

Both props are roving-tabindex internals, not everyday API. A consumer would only reach for `tabbable` to force an item into the tab order, or `focusable={false}` to opt an item out of focus management — and they'd do so on the *real* item, having no reason to think about the offscreen placeholder. To leak, they must apply one of these uniformly across a virtualized offscreen list (e.g. spreading a shared prop object onto every item), which is exactly the contrived pattern the repo's own probe uses (`tabbable: true` in `getOffscreenProbeProps`) and not something organic. Per-item, nobody sets `tabbable`/`focusable` on a country in a 250-row select.

**Consumer code that triggers it:**
```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

// Consumer wants every option reachable via Tab and opts them out of
// roving focus, spreading the same flags onto all items in a long list:
const itemDefaults = { tabbable: true, focusable: false } as const;

{countries.map((country) => (
  <SelectItem
    key={country}
    value={country}
    {...itemDefaults}
    offscreenMode="passive"
    offscreenRoot={listRef}
  />
))}
```

**What the developer actually sees (pre-PR):** For `tabbable={true}` (select offscreen only — combobox already stripped it), a React DEV-only warning: "Received `true` for a non-boolean attribute `tabbable`"; the attribute is not added. For `focusable={false}`, no warning at all — it silently renders `focusable="false"` on the placeholder `<div>` because `focusable` is a valid SVG attribute React passes through. Production builds show neither: the warning is stripped, and the stray `focusable="false"` is inert on a div. There is one behavioral subtlety worth noting: `focusable={false}` already feeds the pre-existing `trulyDisabled` guard (`props.focusable !== false`), so an item marked non-focusable was never given a native `disabled` attribute anyway.

**Verdict:** Not worth fixing on its own merits. `focusable` leaks an invisible, harmless attribute end users never perceive, and `tabbable` is only ever a dev-console nuisance in DEV builds — and only if a consumer improbably broadcasts these roving-tabindex internals across a virtualized list. This is cleanup/consistency, not a real-world bug; the value is in stripping the whole internal-prop set uniformly, not in either prop individually.


### 🎛️ Focus callbacks & autofocus on a list option — `onFocusVisible`, `autoFocus`
**Likelihood:** Effectively never · **Symptom:** Dev-only console warning

`onFocusVisible` is a real, documented Ariakit item option, so it's the more believable of the two: a consumer might attach it to style or react to keyboard-focus on a specific option. But it only leaks if that exact item happens to be offscreen, which means a long virtualized list where the consumer wired a focus callback to an individual option rather than handling focus at the list/store level — an unusual combination. `autoFocus` on an arbitrary list option is essentially nonsensical: you don't auto-focus one option out of a scrolling list of hundreds, and on the default `<div>` placeholder React's `autoFocus` is a silent no-op (a div isn't focusable). The only way `autoFocus` does anything harmful is the contrived case where the consumer also gives that offscreen item a focusable custom `render` (e.g. `render={<button/>}`), at which point React would call `.focus()` on the off-screen placeholder and yank focus/scroll there on mount — but nobody intentionally autofocuses a single virtualized option, so this stays theoretical.

**Consumer code that triggers it:**
```tsx
import { useRef } from "react";
import { ComboboxItem } from "@ariakit/react-components/combobox/combobox-item-offscreen";

function CommandPalette({ commands }: { commands: { id: string; label: string }[] }) {
  const listRef = useRef(null);
  return (
    <Ariakit.ComboboxProvider>
      <Ariakit.Combobox className="overflow-auto" />
      <Ariakit.ComboboxPopover ref={listRef} className="overflow-auto max-h-72">
        {commands.map((command) => (
          <ComboboxItem
            key={command.id}
            value={command.label}
            offscreenMode="passive"
            offscreenRoot={listRef}
            // focus-visible callback the consumer attaches per-option:
            onFocusVisible={() => console.log("focused", command.id)}
            // autoFocus on a list option is already a misuse; harmless on a div,
            // but would steal focus if this offscreen item used render={<button/>}:
            autoFocus={command.id === commands[0]?.id}
          />
        ))}
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

**What the developer actually sees (pre-PR):** For `onFocusVisible`, a React DEV-only console warning: `Warning: Unknown event handler property 'onFocusVisible'. It will be ignored.` (stripped in production). For `autoFocus`: no warning and no attribute on the default div placeholder — completely invisible; it only becomes observable behavior (focus pulled to an off-screen node) if the same offscreen item is given a focusable custom render, which the matrix flags but no realistic consumer writes.

**Verdict:** Not worth fixing for real-world consumers. `onFocusVisible` is at most dev-console noise that no end user ever sees, and the realistic path to it (per-option focus callbacks on a long virtualized list) is already an odd pattern. `autoFocus` on a list option is a misuse that does nothing on the default placeholder; the only harmful variant requires a focusable custom render that nobody would pair with autofocusing a single virtualized option. Both fixes are correctness/hygiene polish, not consumer-facing bug fixes.



---

### At a glance

| Prop(s) on an offscreen item | When a consumer passes it | Pre-PR symptom | Real-world impact |
|---|---|---|---|
| `disabled` (plain `<div>` placeholder) | Disabling options — **common** | `disabled=""` on the div (no warning); `aria-disabled` already correct | Invisible / cosmetic |
| `disabled` + `render={<button/>}` | Options rendered as native buttons | Enabled, focusable offscreen button | **Real, but narrow — the load-bearing fix** |
| `disabled` + `render={<MyButton/>}` | Options via a custom component | Same defect — **still not fixed** | Theoretical; documented limitation |
| `accessibleWhenDisabled` | Keep a disabled option focusable | Dev console warning | Dev-only noise |
| `rowId` | Hand-built grids (no `Row` wrapper) | Dev warning + stray `rowid` attr | Theoretical |
| `shouldRegisterItem` | Decorative / unnavigable row | Dev warning | Theoretical (incoherent with virtualization) |
| `moveOnKeyPress`, `preventScrollOnKeyDown`, `clickOnEnter`, `clickOnSpace` | Per-item interaction tuning | Dev warning | Theoretical |
| `tabbable` | Force into the tab order | Dev warning | Dev-only noise |
| `focusable` | Opt an item out of focus mgmt | Silent `focusable="false"` attr | Invisible / cosmetic |
| `onFocusVisible` | Per-option focus-visible callback | Dev warning ("unknown event handler") | Dev-only noise |
| `autoFocus` | (a misuse on a list option) | Silent no-op on the div | None (unless a focusable render) |

### So, are we fixing too much?

Mostly yes — but cheaply. Walking every case, **exactly one has genuine user-facing behavior: `disabled` + a focusable custom `render` (`<button/>`)**, where the offscreen placeholder was an enabled, focusable button. That's the load-bearing fix and it's worth keeping.

Everything else falls into two buckets that **no end user ever perceives**:
- **Dev-only console warnings** (`accessibleWhenDisabled`, `shouldRegisterItem`, `rowId`, `tabbable`, `clickOnEnter`/`clickOnSpace`, `moveOnKeyPress`/`preventScrollOnKeyDown`, `onFocusVisible`) — React strips these from production builds.
- **Inert stray attributes** (`disabled=""`, `focusable="false"`, `rowid="…"`) — invalid but read by nothing.

And every one of them requires the same improbable stack: the *undocumented* offscreen deep-import + a list long enough to push items offscreen + hand-passing an internal/roving-tabindex prop to individual items. The repo's own example only produces most of these via a synthetic probe harness, not organic usage.

### Recommendation

1. **Keep the prop-stripping** — it's a single uniform destructure, near-zero cost and risk, and it removes real (if dev-only) console noise for the few who use the feature. Just frame it as **DX hygiene**, not "fixing reported user bugs."
2. **Reconsider the `disabled`-on-custom-render machinery.** `getRenderWithDisabled` + the `focusable.tsx` change have now spawned three successive "P2" edge-case rounds (callback clobber → component renders → `accessibleWhenDisabled`) for an audience that is essentially theoretical. The placeholder already carries `aria-disabled`; an honest alternative is to **stop at `aria-disabled` on the placeholder and document that a custom focusable render owns its own offscreen `disabled` state**, rather than keep chasing tag detection through arbitrary render targets. That collapses the most complex, edge-case-prone part of the PR to ~0 lines while preserving the only symptom anyone can see (the native-button case can be covered by the consumer passing `disabled` through their own render, which is what they already do for the on-screen item).
3. If the native-button behavior is kept, **scope the `focusable.tsx` change out** into its own PR with a dedicated test (it changes shared `Focusable` disabled/tabIndex semantics and isn't used by the offscreen path) — see the inline thread.

_Method: symptoms above were measured by rendering each prop on an offscreen `SelectItem`/`ComboboxItem` against the pre-PR base and capturing React's dev warnings + the placeholder's DOM attributes. Likelihood calls are judgment, not measurement._

packages/ariakit-react-components/src/select/select-item-offscreen.tsx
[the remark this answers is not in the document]
diegohaz: Resolved in 877ad4cb9 - codified this as an `aria-disabled` placeholder contract instead of trying to mirror `useFocusable` on offscreen nodes. The regression covers `disabled` + `accessibleWhenDisabled` with a native button render: the inactive placeholder keeps `aria-disabled`, does not receive library-injected native `disabled`, and no item/focus props leak.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in 877ad4cb9 - reverted the shared `Focusable` disabled-target change and removed the separate offscreen disabled helper, so this PR no longer changes general `Focusable` native disabled behavior.

packages/ariakit-react-components/src/composite/utils.ts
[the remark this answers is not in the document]
diegohaz: Resolved in 877ad4cb9 - removed the `getRenderWithDisabled`/tag-detection path entirely. Inactive offscreen placeholders now keep `aria-disabled` only; custom render components and callbacks own any native `disabled` state they need, and the browser regression covers both no injected disabled on placeholder divs and render-owned disabled on buttons.

diegohaz:
**Fixed**

- `877ad4cb9` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344716141)): Removed the disabled render/tag-detection machinery. Inactive offscreen placeholders now rely on `aria-disabled`, and custom render elements own any native `disabled` state.
- `877ad4cb9` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344716140)): Reverted the shared `Focusable` disabled-target change so this PR no longer changes general `Focusable` native disabled behavior.

**Decision**

- `877ad4cb9` ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344716135)): Kept `accessibleWhenDisabled` offscreen placeholders on the same `aria-disabled` placeholder contract instead of mirroring `useFocusable` event suppression on unregistered offscreen nodes.

examples/select-combobox-offscreen-various/test-chrome.ts
```diff
@@ -52,6 +52,139 @@ async function getValue(combobox: Locator) {
...
+  });
+  await page.reload({ waitUntil: "networkidle" });
+  await disableAnimations(page);
+
+  await test.step("select combobox offscreen render props", async () => {
+    const combobox = q.combobox("offscreen props searchable", { exact: true });
+    await combobox.click();
+    const listbox = q.listbox();
+    const disabledItem = listbox.locator(
+      '[data-testid="select-combobox-offscreen-disabled"][data-offscreen]',
```
diegohaz: **Tests: drop the `data-testid` lookups (keep this a browser test)**

Coverage here is good — it pins the real contract (placeholder carries `aria-disabled` + `data-offscreen`, never an Ariakit-injected `disabled`; a consumer's own `render` keeps its native `disabled`; `autoFocus` doesn't steal focus; no React warnings). And keeping this in `test-chrome.ts` is correct: the offscreen state depends on a real `IntersectionObserver` + scroll/viewport, which a browser models faithfully and JSDOM does not — so this stays a Playwright test.

The one thing worth changing is the lookup mechanism: **drop `data-testid` and query by accessible name.** Every other test in this file already locates options the accessible way (`q.option("Dominica", { exact: true })`), and the offscreen placeholders expose `role="option"` with their value as the accessible name (the offscreen hook sets `role` + `children: value`). So scope to the open listbox and query directly:

```ts
const lb = query(listbox);
const disabledItem = lb.option("Disabled offscreen button", { exact: true });
await expect(disabledItem).toHaveAttribute("data-offscreen");
await expect(disabledItem).toHaveAttribute("aria-disabled", "true");
await expect(disabledItem).not.toHaveAttribute("disabled");
```

That matches this file's own convention and the project's "prefer `q` query methods over raw locators / test ids" guidance, and lets the `data-testid` props (and the `prefix`/`testId`-string plumbing) be deleted from the example.

examples/select-combobox-offscreen-various/index.react.tsx
```diff
@@ -31,6 +32,118 @@ function groupItems(items: ReturnType<typeof getItem>[]) {
...
+    | "callback-disabled"
+    | "element-disabled";
+}
+
+type OffscreenRender = Exclude<
+  NonNullable<ComboboxItemProps["render"]>,
+  ReactElement
+>;
+
+const offscreenProbeItems: OffscreenProbeItem[] = [
```
diegohaz, line 70: **Simplify the probe scaffolding (it stays, but it's heavier than it needs to be)**

Since the prop-leak check runs in a real browser against this example, the probe items do need to live here — that's fine. But the surrounding apparatus is heavier than the verification requires and is worth trimming:

- **Drop the `data-testid` plumbing** (the `prefix` param and the five `"data-testid": `${prefix}-offscreen-…`` strings). With the test querying by accessible name (see the test comment), they're dead weight.
- **The probe table is split across three `testId`-keyed switches** (`getOffscreenProbeProps`, `getOffscreenProbeRender`, `renderSelectComboboxProbeItem`) that each re-branch on the same five ids. Data-driving it — each probe item carrying its own `props` and `render` — collapses the three functions into one table and removes the repeated `{ ...testProps, … }` spreads.
- The shared non-DOM bundle (`shouldRegisterItem`, `rowId`, `tabbable`, `clickOnEnter`, …) is the actual subject under test, so keep it as one constant rather than re-spreading per branch.

One caveat while trimming: these props are spread unconditionally, so a probe option scrolled into view becomes a genuinely broken option (`shouldRegisterItem: false` → unregistered/unselectable). Keeping them confined to the dedicated `OffscreenProps…` section (as you already do) is the right containment; just make sure the trim doesn't leak them into the regular demos.

diegohaz (commented):
**Re-review** (focus: simplification + tests). Baseline: my last review at `8f4705c9dc` → new commit `877ad4cb97` "Simplify offscreen disabled placeholders".

**Verified resolved — exactly the simplification I'd hoped for:**
- `focusable.tsx` change reverted (shared `Focusable` disabled behavior restored). ✅
- The `getRenderWithDisabled` / tag-detection apparatus removed; `composite/utils.ts` back to its original contents. ✅
- Both offscreen items are now just *strip props → render `Role`*; the placeholder relies on `aria-disabled` and custom renders own their native `disabled`. Changeset updated to document the contract. ✅
- `tsc` + `lint` clean; full CI green on this head.

The **source is now minimal** — no further source simplification worth making (only leftover is the ~19-key strip list duplicated across the two offscreen files, raised earlier; low priority and idiom-consistent).

**Remaining simplification is on the test/example side (your asks).** *Correction to my first pass: this should stay a `test-chrome.ts` browser test — the offscreen state needs a real `IntersectionObserver`/scroll, so JSDOM is not appropriate here.*
1. **Tests good enough?** Yes on coverage (no leaked attrs, `aria-disabled` present, consumer-owned `disabled` survives, `autoFocus` doesn't steal focus, no React warnings), and Playwright is the right tool. The cleanup is purely in *how* it locates and sets up items.
2. **`data-testid`** — not optimal. Placeholders expose `role="option"` + their value as the accessible name, and the rest of this file uses `q.option("Dominica", …)`. Replace the `data-testid` lookups with `query(listbox).option("Disabled offscreen button")` and delete the `data-testid`/`prefix` plumbing. Inline on the test.
3. **Simplify the example scaffolding** — the probe items must stay (the browser test renders this page), but the three `testId`-keyed switch functions can collapse into one data-driven probe table, and the `data-testid` strings go away. Inline on the example.

**Bottom line:** production code is in great shape and correctly de-scoped. Keep the browser test; the outstanding work is trimming the probe scaffolding and swapping `data-testid` for accessible queries.

.changeset/300-offscreen-item-props.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed offscreen [`SelectItem`](https://ariakit.com/reference/select-item) and [`ComboboxItem`](https://ariakit.com/reference/combobox-item) elements to avoid passing item and focus props to inactive placeholder DOM nodes. Inactive offscreen placeholders rely on `aria-disabled`; custom render elements own any native `disabled` state they need.
```
diegohaz: This should probably be a multiline changeset.

examples/select-combobox-offscreen-various/test-chrome.ts
diegohaz: Fixed in a9e1333f4 - removed the data-testid lookups and now query the open listbox by accessible names, including the custom div probe as an option in the searchable select fixture.

examples/select-combobox-offscreen-various/index.react.tsx
diegohaz: Fixed in a9e1333f4 - removed the testId/prefix plumbing and moved the probe-specific props/render behavior into the table with shared offscreenProbeProps.

.changeset/300-offscreen-item-props.md
diegohaz: Fixed in a9e1333f4 - converted the changeset to a multiline entry with a title and explanatory paragraphs.

diegohaz:
**Fixed**

- a9e1333f4 ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344958721)): Removed `data-testid` lookups from the browser test and switched the probe assertions to accessible-name queries.
- a9e1333f4 ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344958813)): Simplified the offscreen probe scaffolding into table-driven `props`/`render` rows with shared probe props.
- a9e1333f4 ([reply](https://github.com/ariakit/ariakit/pull/6207#discussion_r3344958907)): Converted the changeset to a multiline entry.

PR #6209, opened 2026-06-03
Fix table of contents navigation cleanup

diegohaz:
## Motivation

The table of contents script could initialize twice on first page load because it listened to both `DOMContentLoaded` and `astro:page-load`. Astro view-transition navigations could also rerun setup for the same TOC element, creating duplicate observers, scroll listeners, and teardown callbacks. The existing cleanup split observer and scroll teardown across repeated `astro:before-preparation` listeners, so teardown handlers and retained observers could accumulate.

## Solution

Initialize the TOC from `astro:page-load` only, add a per-element initialized marker, and replace the duplicated teardown hooks with one `{ once: true }` cleanup that disconnects the observer, removes the scroll listener, cancels pending animation frames, and clears the marker for future page loads.

No changeset is included because the changed `app` workspace is ignored by Changesets in this repository.


app/src/components/table-of-contents.astro
```diff
@@ -113,6 +113,11 @@ const allHeadings = title
 
     for (const tocElement of tocElements) {
       if (tocElement.dataset.inert) continue;
+      if (tocElement.hasAttribute("data-table-of-contents-initialized")) {
+        continue;
+      }
+      tocElement.setAttribute("data-table-of-contents-initialized", "true");
```
diegohaz: Minor simplification (optional): the marker is set here before `headings` is known, then removed again in the `headings.length === 0` branch a few lines down. You can instead set it only once the element is actually being initialized, which drops the set‑then‑remove dance and turns the no‑headings branch into a plain `continue`:

```ts
if (tocElement.hasAttribute("data-table-of-contents-initialized")) {
  continue;
}

const topOffset = Number.parseInt(tocElement.dataset.topOffset || "0");
const links = Array.from(/* ... */);
const headings = links.map(/* ... */).filter(/* ... */);
if (headings.length === 0) continue;

tocElement.setAttribute("data-table-of-contents-initialized", "true");
```

Behavior is identical — this runs synchronously, so there's no re‑entrancy window between the set and the headings check that the early marker would guard against — and `cleanup`'s `removeAttribute` stays correct because only fully‑initialized elements ever reach it. Not blocking; it just removes one branch body and the extra attribute write.

app/src/components/table-of-contents.astro
```diff
@@ -301,20 +303,25 @@ const allHeadings = title
         timeout = window.requestAnimationFrame(onScroll);
       };
 
+      const cleanup = () => {
+        observer?.disconnect();
+        window.removeEventListener("scroll", throttledScroll);
+        if (timeout) window.cancelAnimationFrame(timeout);
+        tocElement.removeAttribute("data-table-of-contents-initialized");
+      };
```
diegohaz, line 309: This `cleanup` + single `{ once: true }` teardown is the core of the fix, but after the second commit it ships with no automated coverage: commit 1 added `app/src/tests/table-of-contents-browser.ts`, commit 2 removed it, so the net diff carries the behavior change but no test.

I checked the removed test rather than assuming: its path matches the `*-browser` `testMatch` in `app/playwright.config.ts` (`/\/tests\/[^/]*-browser/`), so it was actually collected in CI on chrome/firefox/safari — not dead. I also traced it against this implementation and it should pass: nothing else registers `astro:before-preparation` with `{ once: true }`, so `beforePreparationOnceCount` stays 1, the marker guard keeps the repeated `astro:page-load` dispatches from re‑binding, and `cleanup` zeroes the listener/observer counts. So it wasn't removed because it was broken.

That said, dropping *that* test is defensible: it monkey‑patched `window.addEventListener`/`IntersectionObserver` and asserted through a `window.__tocAudit` global, which is exactly the implementation‑coupled, test‑only‑harness style the repo's sandbox conventions steer away from.

The remaining concern is just that the lifecycle regression now has zero coverage. Could you either keep a lighter regression check — e.g. assert a single `[data-table-of-contents-initialized]` element and that active‑link/indicator behavior isn't duplicated after an `astro:page-load` → `astro:before-preparation` cycle — or confirm it's intentionally left untested because the observer/listener leak isn't practically assertable through user‑facing behavior? Flagging so the call is explicit.

diegohaz (commented):
**General PR review — initial review.** Baseline: merge-base `cf8577b0` → head `a69c8af1`. Scope: the full net diff (`app/src/components/table-of-contents.astro`) plus surrounding context — the two-commit history (the fix commit also added then the second commit removed `app/src/tests/table-of-contents-browser.ts`, so it nets out of the diff), the Astro view-transitions lifecycle, peer components, and the Playwright test config.

**Overall:** the change is correct and a genuine improvement. The init/teardown logic is sound, and I found no correctness regressions. Two non-blocking findings, both inline.

**What I verified:**
- **`astro:page-load` reliability.** `ClientRouter` is mounted globally in `app/src/pages/_layout.astro`, and every TOC host page (`[...example]`, `[...component]`, `[...guide]`, `reference-content`) renders through that layout, so `astro:page-load` fires on initial load and on every navigation. Dropping `DOMContentLoaded` is safe — and it makes this component consistent with every other client script in `app/` (`sticky`, `horizontal-scroll`, `submit-button`, `scroll-current-into-view`, `preserve-scroll`), none of which use `DOMContentLoaded`.
- **Marker pattern.** `data-table-of-contents-initialized` mirrors the existing `data-sticky-initialized` guard in `sticky.astro`, so the dedup approach is the established repo pattern, not a one-off.
- **Teardown correctness.** Consolidating the two `astro:before-preparation` handlers into one `{ once: true }` `cleanup` (with `observer?.disconnect()` instead of per-heading `unobserve`) fixes the real leak: the old code re-registered an un-`once`'d listener on every init, so handlers accumulated across navigations. The new single-shot cleanup removes the `window` scroll listener (which persists across swaps and genuinely needs explicit removal), disconnects the observer, cancels the pending rAF, and clears the marker. Balanced across multiple TOC elements since each gets its own closure.
- **Drive-by fix.** `if (headings.length === 0) return;` → `continue;` also fixes a latent bug where a headingless first TOC element would abort initialization of any later TOC elements.

**Findings (both inline, both non-blocking):**
1. *Test coverage.* The PR ships the lifecycle fix with no automated coverage — the regression test was added in commit 1 and removed in commit 2. I confirmed the removed test was CI-collected (matched the `*-browser` `testMatch`) and would pass against this code, so it wasn't dropped because it was broken; it was dropped because it was an implementation-coupled harness (monkey-patched globals + `window.__tocAudit`), which is a defensible reason. Asking whether a lighter behavior-oriented regression check is feasible, or to confirm the leak is intentionally left untested.
2. *Simplification (optional).* The marker is set before `headings` is computed and then removed again in the no-headings branch; setting it only after `if (headings.length === 0) continue;` removes a branch body and the redundant attribute write with identical behavior.

**Not validated locally:** I did not run `astro check`/`pnpm tsc` or the Playwright suite — the change is type-trivial (one `let observer: IntersectionObserver | null` plus an optional-chained `disconnect()`), and the substantive risks (event reliability, test collection, teardown balance) were verified statically against the repo. No changeset is correctly omitted (the `app` workspace is ignored by Changesets).

app/src/components/table-of-contents.astro
diegohaz: Fixed in 61b95df3bfa8f273f8f033d2ca62e4a3f286027a - moved the initialized marker write until after matching headings are found, so the no-headings path is now a plain `continue` without the set/remove round trip.

diegohaz:
**Fixed**

- `61b95df3bfa8f273f8f033d2ca62e4a3f286027a` ([reply](https://github.com/ariakit/ariakit/pull/6209#discussion_r3348826925)): moved the initialized marker write after matching headings are found, so headingless TOCs use a plain `continue` without the set/remove round trip.

PR #6210, opened 2026-06-03
Fix SelectRenderer orientation forwarding

diegohaz:
## Motivation

`SelectRenderer` accepts renderer props from `CompositeRendererOptions`, including `orientation`, but `useSelectRenderer` destructured `orientation` and never passed it to `useCompositeRenderer`. This made direct consumer usage such as:

```tsx
<SelectRenderer orientation="horizontal" />
```

render with the default vertical item layout instead.

While checking the forwarding path, `useCompositeRenderer` also treated any explicit `orientation` prop as a truthy condition that resolved to vertical layout, so forwarding alone was not enough for `orientation="horizontal"` to work when a store is present.

## Solution

Let `orientation` stay in `SelectRenderer`'s remaining props so it reaches `useCompositeRenderer`, and make `useCompositeRenderer` prefer an explicit renderer `orientation` before falling back to the store orientation.

Added a React sandbox browser regression that opens a real `Select` using `SelectRenderer orientation="horizontal"` and asserts the rendered options are laid out horizontally.


app/src/sandbox/select-renderer-orientation/test-browser.ts
```diff
@@ -0,0 +1,50 @@
...
+  }) => {
+    const select = q.combobox("Favorite fruit");
+    await test
+      .expect(async () => {
+        if ((await select.count()) === 0) {
+          await gotoAndSettle(page, page.url());
+        }
+        await test.expect(select).toBeVisible({ timeout: 5_000 });
+      })
+      .toPass({ timeout: 45_000 });
```
diegohaz: Is this needed?

app/src/sandbox/select-renderer-orientation/test-browser.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const getRect = (name: string) =>
+      q.option(name).evaluate((element) => {
+        const rect = element.getBoundingClientRect();
+        return {
+          left: rect.left,
+          top: rect.top,
+          width: rect.width,
+        };
+      });
```
diegohaz: If we are testing layout, shouldn't we use a visual test instead?

app/src/sandbox/select-renderer-orientation/test-browser.ts
```diff
@@ -0,0 +1,50 @@
+import { gotoAndSettle, withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test.setTimeout(60_000);
```
diegohaz: Is this needed?

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: Fixed in f1ab01d28 - kept the retry because removing it made a cold focused run time out on a blank preview route, and added a comment explaining that the retry reloads while Vite is re-optimizing dependencies.

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: This is not an issue because the regression is a deterministic browser layout contract: `orientation="horizontal"` must produce horizontal item geometry. The Ariakit Playwright guidance covers layout/geometry in browser tests, while `visual()` only exercises screenshot baselines in CI visual mode and would add baseline churn without improving the focused regression assertion.

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: Fixed in f1ab01d28 - added a comment explaining that the 60s timeout exists to give the 45s cold-preview retry enough room after Vite serves a blank route during dependency re-optimization.

diegohaz:
**Fixed**

- f1ab01d28 ([reply](https://github.com/ariakit/ariakit/pull/6210#discussion_r3348648258)): kept the cold-preview retry and documented why it is needed after confirming removal fails on a blank route.
- f1ab01d28 ([reply](https://github.com/ariakit/ariakit/pull/6210#discussion_r3348649970)): documented the 60s timeout as support for the 45s retry window.

**Invalid**

- geometry assertion ([reply](https://github.com/ariakit/ariakit/pull/6210#discussion_r3348649241)): kept the browser-observed rect assertions because this is a deterministic orientation/layout regression; visual snapshots would add CI baseline churn without improving the focused check.

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: Superseded by `8f0f934`: `test-browser.ts` is replaced with a JSDOM `test.ts` using `@ariakit/test`. The orientation regression surfaces as deterministic inline styles, so there is no served preview to wait on — the retry is gone entirely.

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: Superseded by `8f0f934`: instead of measuring browser layout geometry, the new JSDOM test asserts the renderer's computed inline styles (`left`/`top`) that the forwarded `orientation` produces. No screenshots and no `getBoundingClientRect`, so the visual-vs-geometry question no longer applies.

app/src/sandbox/select-renderer-orientation/test-browser.ts
diegohaz: Superseded by `8f0f934`: the 60s timeout only existed to cover the cold-preview retry, which is gone now that this is a JSDOM test.

PR #6211, opened 2026-06-03
Fix collection renderer measurements

diegohaz:
## Motivation

`CollectionRenderer` can keep ResizeObserver subscriptions around after rendered items are replaced or the renderer unmounts, which lets stale observed nodes trigger synchronous measurement updates.

Mixed-orientation nested renderers can also compute the wrong parent-axis size. A vertical renderer containing a horizontal nested item was summing the nested item widths and using that value as a vertical offset.

## Solution

Track measured item refs per item id, unobserve stale nodes on replacement/null callbacks, and disconnect the item ResizeObserver on renderer unmount while preserving observations across React 18 StrictMode effect replay.

Preserve the caller-requested measurement axis in `getItemSize`, use main-axis sums only for same-orientation nested items, and use cross-axis max child extents or element measurements for mixed-orientation items.

Added focused sandbox browser tests for observer cleanup and mixed-orientation offsets, hardened to recover from cold preview reloads without Playwright retries.


app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
```diff
@@ -0,0 +1,44 @@
...
+        await test
+          .expect(locator)
+          .toBeVisible({ timeout: previewReadyTimeout });
+        return;
+      } catch (error) {
+        lastError = error;
+      }
+    }
+    throw lastError;
+  };
```
diegohaz: This shouldn't be needed.

app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  test("uses cross-axis size for mixed-orientation nested items", async ({
+    page,
+    q,
+  }) => {
+    test.setTimeout(90_000);
```
diegohaz: Same here.

app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
```diff
@@ -0,0 +1,44 @@
...
+    await expectPreviewContent(page, q.text(/Horizontal group/));
+    await test.expect(q.text("After group")).toBeVisible();
+
+    const groupBox = await page.locator("#horizontal-group").boundingBox();
+    const afterBox = await page.locator("#after-group").boundingBox();
```
diegohaz: We should use semantic elements with `q`, not ids.

app/src/sandbox/composite-renderer-resize-observer-cleanup/test-browser.ts
```diff
@@ -0,0 +1,41 @@
...
+        await test
+          .expect(locator)
+          .toBeVisible({ timeout: previewReadyTimeout });
+        return;
+      } catch (error) {
+        lastError = error;
+      }
+    }
+    throw lastError;
+  };
```
diegohaz: We shouldn't need this.

app/src/sandbox/composite-renderer-resize-observer-cleanup/test-browser.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  test("cleans up observed item elements", async ({ page, q }) => {
+    test.setTimeout(90_000);
```
diegohaz: And this.

app/src/sandbox/composite-renderer-resize-observer-cleanup/index.react.tsx
```diff
@@ -0,0 +1,166 @@
...
+import type { CompositeRendererItemObject } from "@ariakit/react-components/composite/composite-renderer";
+import { StrictMode, useEffect, useState, useSyncExternalStore } from "react";
+
+interface ResizeObserverStore {
+  subscribe: (listener: () => void) => () => void;
+  getSnapshot: () => number;
+  restore: () => void;
+}
+
+function createResizeObserverStore(): ResizeObserverStore {
```
diegohaz: This sandbox doesn't seem to represent real-world usage. Why do we need this function?

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -789,6 +860,21 @@ export function useCollectionRenderer<T extends Item = any>({
...
+      for (const itemId of refs.keys()) {
+        if (itemIds.has(itemId)) continue;
+        if (elements.has(itemId)) continue;
+        refs.delete(itemId);
+      }
+    };
+
+    pruneItemRefs(measuredItemRefs);
+    pruneItemRefs(fixedItemRefs);
+  }, [itemsProps, elements, measuredItemRefs, fixedItemRefs]);
```
diegohaz: The changes on this file made it too complex. Is it really needed? Can we simplify it?

app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
diegohaz: Fixed in edbc0f4c5 — converted this to a JSDOM `test.ts`. The mixed-orientation offset is derived from the items' style sizes, so it can be asserted in-process without a served preview, which removes the cold-reload retry helper entirely.

app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
diegohaz: Fixed in edbc0f4c5 — gone with the browser test; the JSDOM test runs in-process and needs no extended timeout.

app/src/sandbox/composite-renderer-nested-orientation/test-browser.ts
diegohaz: Fixed in edbc0f4c5 — the JSDOM test asserts the offset via `q.text("After group")` instead of an id locator.

app/src/sandbox/composite-renderer-resize-observer-cleanup/test-browser.ts
diegohaz: Fixed in fae3db461 — converted to a JSDOM `test.ts`. Observer cleanup is pure bookkeeping, so there is no served preview to wait on and the retry helper is gone.

app/src/sandbox/composite-renderer-resize-observer-cleanup/test-browser.ts
diegohaz: Fixed in fae3db461 — removed with the browser test; the JSDOM test needs no extended timeout.

app/src/sandbox/composite-renderer-resize-observer-cleanup/index.react.tsx
diegohaz: Fixed in fae3db461 — the example is now a plain `CompositeRenderer` with Replace/Hide controls and no `window.ResizeObserver` patching. The observer spying moved into the JSDOM `test.ts`, where it asserts observe/unobserve/disconnect counts after a replace and an unmount.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in 2737e9b9e — dropped the dual measured/fixed ref maps, the `getItemRef` factory, the `setItemElementRef` indirection, and the ref-pruning effect. It is now a single shared item ref plus two small effects: a reconciliation effect that keeps the live rendered nodes (re-keying any node React reused across ids) and unobserves the rest, and a mount/unmount effect that observes on mount and disconnects on unmount.

diegohaz:
Merged `main` (brings in the #6212 cold-reload fix) and addressed all 7 unresolved review comments across three commits.

**Fixed**

- 2737e9b9e ([reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350963221)): Simplified the renderer's observer cleanup — dropped the dual measured/fixed ref maps, the `getItemRef` factory, the `setItemElementRef` indirection, and the ref-pruning effect, replacing them with a single shared item ref plus two small effects (a reconciliation effect that keeps the live rendered nodes and unobserves the rest, and a mount/unmount effect that observes on mount and disconnects on unmount).
- edbc0f4c5 ([reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350961109), [reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350961234), [reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350961948)): Converted the mixed-orientation sandbox to a deterministic JSDOM `test.ts`. The offset is style-derived, so it asserts the next item's computed `top` via `q.text("After group")` — no served preview, so the retry helper, the 90s timeout, and the `#id` locator are all gone.
- fae3db461 ([reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350962109), [reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350962248), [reply](https://github.com/ariakit/ariakit/pull/6211#discussion_r3350962914)): Rewrote the resize-observer sandbox as an ordinary `CompositeRenderer` with Replace/Hide controls (no `window.ResizeObserver` patching) and moved the observer spying into a JSDOM `test.ts` that asserts observe/unobserve/disconnect counts. Removes the retry helper and timeout.

**Verification**: `pnpm tsc`, `pnpm build`, and the full `test-react` suite (152 files / 735 tests) pass. Red checks confirmed each JSDOM test fails against the broken code (observer count 3 without the replace-unobserve, ≠0 without disconnect; `top` 310px with the axis-override bug restored).

PR #6222, opened 2026-06-03
Fix DisclosureContent over-waiting to unmount with mixed transition/animation timing

### LINKED ISSUE #6220: DisclosureContent over-waits to unmount when the longest transition delay and longest animation duration are on different properties
### Current behavior

`DisclosureContent` computes the animation timeout as `max(all delays) + max(all durations)` (via `parseCSSTime`), rather than the longest *per-property* end time (`max` over properties of `delay_i + duration_i`). When a transition **and** an animation are both present and the longest delay belongs to a different property than the longest duration, the timeout overestimates.

Since `mounted = open || animating`, with `unmountOnHide` the content stays mounted — and unmount is delayed — well past the point the animation has visually finished.

Observed with `@ariakit/react@0.4.29`.

1. Open https://stackblitz.com/edit/ukfwcpfn?file=src%2FApp.tsx
2. The content has a transition (delay 500ms / duration 10ms) and a leave animation (delay 0ms / duration 300ms); the real longest end is **510ms**.
3. Click **"Open, then hide & measure unmount delay"** — the content stays mounted **~800ms** (`max(delays = 500) + max(durations = 300)`) instead of ~510ms, i.e. ~290ms too long.

### Expected behavior

The timeout should be the longest per-property end time — `max` over properties of `(delay + duration)`. The content should unmount ~510ms after hide.

### Workaround

Set the `animated` prop to an explicit number (a manual timeout) to bypass the CSS parsing, or avoid mixing a transition and an animation whose longest delay and longest duration sit on different properties.

### Possible solutions

Compute the per-property end time (`delay_i + duration_i`) and take the max across properties instead of `max(delays) + max(durations)`. Note `getComputedStyle` can return `transition-delay` and `transition-duration` lists of different lengths (CSS cycles the shorter list to match the `transition-property` count) — the fix must **cycle** the shorter list, not pair by index, or it could under-count and cut animations short. Combine each element's transition/animation end times (and the backdrop's, via `otherElement`) by taking the max, and add a regression test for the mixed transition + animation case.

diegohaz:
## Motivation

`DisclosureContent` keeps animated content mounted until its CSS transition/animation finishes, then unmounts it (when `unmountOnHide` is set). To know when the animation ends, it reads the transition/animation durations and delays from `getComputedStyle` and computes a timeout.

The problem is how that timeout was computed: it took `max(all delays) + max(all durations)`, pairing the single largest delay with the single largest duration even when they belong to **different** properties. When a transition and an animation are both present and the longest delay sits on a different property than the longest duration, this overestimates the real end time.

For example, a content element with:

- a transition — `delay 500ms`, `duration 10ms` → ends at **510ms**
- a leave animation — `delay 0ms`, `duration 300ms` → ends at **300ms**

has a real longest end time of **510ms**, but the old code waited `max(500, 0) + max(10, 300) = 800ms` — about 290ms too long. With `unmountOnHide`, the content stayed mounted well past the point the animation had visually finished.

Closes #6220.

## Solution

The timeout is now the longest **per-item** end time. For each timeline — the content's transition, the content's animation, and the backdrop/dialog's transition and animation (via `otherElement`) — each item's delay is paired with its own duration, and the longest `delay + duration` is taken; then the max across all timelines is used:

```ts
function getEndTime(names: string, delays: string, durations: string) {
  const nameList = names.split(",");
  const delayList = delays.split(",");
  const durationList = durations.split(",");
  let endTime = 0;
  for (let index = 0; index < nameList.length; index += 1) {
    // `transition-property: none` / `animation-name: none` means nothing runs.
    if (nameList[index]?.trim() === "none") continue;
    const delay = parseCSSTime(delayList[index % delayList.length]);
    const duration = parseCSSTime(durationList[index % durationList.length]);
    endTime = Math.max(endTime, delay + duration);
  }
  return endTime;
}

const timeout = Math.max(
  getEndTime(transitionProperty, transitionDelay, transitionDuration),
  getEndTime(animationName, animationDelay, animationDuration),
  getEndTime(transitionProperty2, transitionDelay2, transitionDuration2),
  getEndTime(animationName2, animationDelay2, animationDuration2),
);
```

Details that matter:

- **Item count comes from `transition-property`/`animation-name`.** `getComputedStyle` returns the `*-delay`/`*-duration` lists as authored, which can be shorter than the property list (CSS cycles the shorter ones). Deriving the count from the property/name list and indexing the delay/duration lists modulo their own length matches CSS's cycling and avoids cutting an animation short.
- **`none` items are ignored.** A leftover `animation-duration` while `animation-name` is `none` (or `transition-property: none`) contributes nothing, so the content can unmount immediately instead of waiting out a phantom duration.
- **Non-numeric values are guarded.** Values like `animation-duration: auto` parse to `NaN`; `parseCSSTime` returns `0` for them so they don't poison the `Math.max`.

For the example above, the timeout is now `max(500 + 10, 0 + 300) = 510ms` instead of `800ms`.

## Reproduction

`app/src/sandbox/disclosure-content-6220/` reproduces the bug with a content element whose longest transition delay (500ms) and longest animation duration (300ms) live on different properties. The browser test measures how long the content stays mounted after being hidden and asserts it unmounts close to the real 510ms end time rather than the over-estimated 800ms. A second example covers the `animation-name: none` case (the content must unmount immediately rather than wait out the leftover duration).

## Workaround

Until this fix is released, pass an explicit numeric `animated` timeout to the store (or to `DisclosureContent`/`Dialog`). A number is treated as a manual timeout, so the buggy CSS time parsing is skipped entirely. Set it to the real longest per-property end time of your animation — for the example above, `max(500 + 10, 0 + 300) = 510`:

```tsx
// TODO(#6220): remove once https://github.com/ariakit/ariakit/issues/6220 is fixed.
const store = Ariakit.useDisclosureStore({ animated: 510 });
```

Alternatively, avoid mixing a transition and an animation whose longest delay and longest duration sit on different properties.


.changeset/300-disclosure-content-animation-timeout.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) (and components built on top of it, such as [`Dialog`](https://ariakit.com/reference/dialog) and [`Popover`](https://ariakit.com/reference/popover)) over-waiting to unmount when a transition and an animation are both applied and the longest delay and longest duration belong to different properties.
```
diegohaz: This is a multiline changeset, but it's missing a title

app/src/sandbox/disclosure-content-6220/test-browser.ts
```diff
@@ -0,0 +1,89 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ query, test }) => {
+  test("unmounts at the longest per-property end time, not max delay + max duration", async ({
+    q,
+    page,
+  }) => {
+    const section = q.region("Mixed transition and animation");
+    const within = query(section);
```
diegohaz: Rename to withinSection

app/src/sandbox/disclosure-content-6220/test-browser.ts
```diff
@@ -0,0 +1,89 @@
...
+    // The content is unmounted again after the leave animation.
+    await test.expect(content).not.toBeAttached();
+    await test.expect(section).toHaveAttribute("data-mounted-state", "false");
+  });
+
+  test("treats a duration with animation-name: none as no animation", async ({
+    q,
+  }) => {
+    const section = q.region("No animation with duration");
+    const within = query(section);
```
diegohaz: Same

packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
```diff
@@ -34,19 +34,38 @@ function afterPaint(cb: () => void) {
...
+// own length. This is more accurate than adding the longest delay to the
+// longest duration, which overestimates when they belong to different items
+// (e.g. a transition with a long delay alongside an animation with a long
+// duration).
+function getEndTime(names: string, delays: string, durations: string) {
+  const nameList = names.split(",");
+  const delayList = delays.split(",");
+  const durationList = durations.split(",");
+  let endTime = 0;
+  for (let index = 0; index < nameList.length; index += 1) {
```
diegohaz: Why not for...of?

.changeset/300-disclosure-content-animation-timeout.md
diegohaz: Fixed in da98458a4 - added a title line before the body, per the multiline changeset convention.

app/src/sandbox/disclosure-content-6220/test-browser.ts
diegohaz: Fixed in 8f56325a9 - renamed to `withinSection`.

app/src/sandbox/disclosure-content-6220/test-browser.ts
diegohaz: Fixed in 8f56325a9 - renamed to `withinSection`.

packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
diegohaz: Fixed in b350dc818 - switched to `for (const [index, name] of nameList.entries())`, which also lets the `none` check read `name` directly.

diegohaz:
Addressed the review feedback:

**Fixed**
- `da98458a4` ([reply](https://github.com/ariakit/ariakit/pull/6222#discussion_r3352185450)): Added a title line before the body of the multiline changeset.
- `8f56325a9` ([reply](https://github.com/ariakit/ariakit/pull/6222#discussion_r3352185551), [reply](https://github.com/ariakit/ariakit/pull/6222#discussion_r3352185666)): Renamed the `within` query variable to `withinSection` in both tests.
- `b350dc818` ([reply](https://github.com/ariakit/ariakit/pull/6222#discussion_r3352185755)): Switched the `getEndTime` loop to `for (const [index, name] of nameList.entries())` so the `none` check reads `name` directly.

Verified locally: browser tests pass, `pnpm tsc` clean, `pnpm test-react run disclosure dialog dialog-animated-various` 159/159, lint clean.

.changeset/300-disclosure-content-animation-timeout.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed `DisclosureContent` over-waiting to unmount with mixed transitions and animations
+
+[`DisclosureContent`](https://ariakit.com/reference/disclosure-content) (and components built on top of it, such as [`Dialog`](https://ariakit.com/reference/dialog) and [`Popover`](https://ariakit.com/reference/popover)) could keep `unmountOnHide` content mounted longer than necessary when a transition and an animation were both applied and the longest delay and longest duration belonged to different properties.
```
diegohaz: `unmountOnHide` must be a link as well.

diegohaz: Fixed in 818144198 - linked `unmountOnHide` to [its DisclosureContent reference](https://ariakit.com/reference/disclosure-content#unmountonhide).

diegohaz:
**Fixed**
- `818144198` ([reply](https://github.com/ariakit/ariakit/pull/6222#discussion_r3352239916)): Linked `unmountOnHide` to its [`DisclosureContent` reference](https://ariakit.com/reference/disclosure-content#unmountonhide) in the changeset, matching the other API references.

diegohaz:
Follow-up: linked the **second** `unmountOnHide` occurrence in the changeset too (`5af5d948b`), so both mentions in the body now point to the [`DisclosureContent` reference](https://ariakit.com/reference/disclosure-content#unmountonhide).

PR #6225, opened 2026-06-03
Fix virtualized renderer leaking ResizeObserver and item nodes when itemSize is not set

### LINKED ISSUE #6215: Virtualized CompositeRenderer/SelectRenderer leak ResizeObserver and item nodes when itemSize is not set
### Current behavior

When a virtualized `CompositeRenderer` / `SelectRenderer` (internally `CollectionRenderer`) is rendered without a fixed `itemSize`, items are measured with an internal `ResizeObserver`. That observer is never disconnected and scrolled-out items are never unobserved:

- As items scroll out of the virtualized window, they stay observed — their (now detached) DOM nodes remain referenced, so the observed-node set only grows.
- On unmount, the `ResizeObserver` is never `disconnect()`ed, leaking it along with every measured item node.
- Stale resize callbacks keep firing `flushSync(updateElements)`, causing extra synchronous re-renders.

Observed with `@ariakit/react@0.4.29` (`@ariakit/react-components@0.1.2`).

1. Open https://stackblitz.com/edit/znnabrzv?file=src%2FApp.tsx
2. Click **"Scroll through the list"** — `Observed elements` and `Detached observed` climb while `unobserve()` calls stay at **0**.
3. Click **"Unmount the list"** — the item `ResizeObserver` is never disconnected: `Observers still leaking nodes` and `Detached observed` stay **> 0**.

The repro instruments the global `ResizeObserver` so the otherwise-invisible internal leak shows up as live counters.

### Expected behavior

Scrolled-out items should be `unobserve()`d (and removed from the internal `elements` map), and the `ResizeObserver` should be `disconnect()`ed when the renderer unmounts, so no detached nodes or observers are retained.

### Workaround

If item sizes are known ahead of time, pass `itemSize` to the renderer to skip the measuring / `ResizeObserver` path entirely.

### Possible solutions

- `disconnect()` the element `ResizeObserver` in a cleanup effect on unmount.
- In the item ref callback, when it runs with a falsy `element` (or when an id remounts to a new node), `unobserve` the previously observed node and delete its `elements` entry. Returning a cleanup from the ref callback works on React 19, but a fallback is required given the React 17/18 peer range.

diegohaz:
## Motivation

When a virtualized [`CompositeRenderer`/`SelectRenderer`](https://ariakit.com/components/composite) (internally `CollectionRenderer`) is rendered **without** a fixed `itemSize`, each rendered item is measured with an internal `ResizeObserver`. That observer was never disconnected and items that stopped being rendered were never unobserved:

- Items dropped from the virtualized window (e.g., scrolled out) stayed observed, so their now-detached DOM nodes were retained — the observed-node set only grew.
- On unmount, the `ResizeObserver` was never `disconnect()`ed, leaking it along with every measured node, and stale resize callbacks kept firing `flushSync`, causing extra synchronous re-renders.

Reproduces #6215.

## Solution

`CollectionRenderer` now stops observing item nodes that are no longer rendered and disconnects the observer on unmount:

- Disconnect the element `ResizeObserver` in a cleanup effect when the renderer unmounts.
- Unobserve and forget items that are no longer rendered, reconciling the observed set against the currently rendered items.
- When an item id is reassigned to a different node, unobserve the previous node before observing the new one.

Detached item nodes already contribute nothing to size computation (sizes fall back to the cached `data` when an element is no longer connected), so dropping them from the observed set doesn't change the measured layout — it only stops retaining them.

Fixes #6215.

## Workaround

Until this fix is released, pass a known `itemSize` to the renderer. This skips the item-measuring `ResizeObserver` path entirely, so no item node is observed and nothing can leak (it requires the item sizes to be known ahead of time):

```tsx
<CompositeRenderer
  store={composite}
  items={items}
  // TODO: Remove once https://github.com/ariakit/ariakit/issues/6215 is fixed.
  itemSize={32}
>
  {(item) => <CompositeItem key={item.id} {...item} />}
</CompositeRenderer>
```


app/src/sandbox/composite-renderer-6215/test.ts
```diff
@@ -0,0 +1,130 @@
...
+// or removed when the whole list unmounts — must stop being observed, and the
+// observer must be disconnected when the renderer unmounts. While the renderer
+// is measuring, the items that are still rendered must stay observed.
+//
+// The `itemSize` workaround skips the measuring/ResizeObserver path entirely, so
+// nothing is observed and nothing can leak; the assertions below hold then too.
+
+const observers: MockResizeObserver[] = [];
+
+class MockResizeObserver {
```
diegohaz: I'd rather have a `test-browser` that is checking real browser behavior instead of mocking things like this. You should ask yourself: how would a real user reproduce the issue when accessing the sandbox via the browser? That's how we should test it.

diegohaz: Fixed in 4841bd7df — replaced the JSDOM mock with a real `test-browser.ts`. The sandbox is now a scrollable virtualized list (no `itemSize`) that wraps the real `ResizeObserver` (calling through to it, like the issue's StackBlitz repro) and shows the number of item nodes still observed after they detach. The test scrolls the whole list, refreshes the rows (remount to new nodes, same id), and unmounts the list — asserting that count stays `0` throughout. Verified it fails without the fix (the count climbs as you scroll) and passes with it.

One note for transparency: the app preview runs React 19, so the browser test can't exercise the React-18 StrictMode case where the simulated remount would otherwise leave items unobserved. I kept a small defensive re-observe in the disconnect effect (with a comment) for that, since re-adding a JSDOM test would reintroduce the mocking you asked to avoid — happy to drop it or cover it differently if you'd prefer.

diegohaz:
Addressed the review feedback.

**Fixed**

- `4841bd7df` ([reply](https://github.com/ariakit/ariakit/pull/6225#discussion_r3352640183)): replaced the JSDOM mock-based test with a real `test-browser.ts`. The sandbox is now a scrollable virtualized list (no `itemSize`) that wraps the real `ResizeObserver` (calling through to it, like the issue's StackBlitz repro) and surfaces the number of detached-but-observed item nodes. The browser test scrolls through the list, refreshes the rows, and unmounts — asserting that count stays `0`. Verified it fails without the fix and passes with it.

Also merged `main` (picks up the long-branch-name preview-deploy fix from #6227).

app/src/sandbox/composite-renderer-6215/test-browser.ts
```diff
@@ -0,0 +1,63 @@
...
+    const settle = () =>
+      page.evaluate(
+        () =>
+          new Promise<void>((resolve) =>
+            requestAnimationFrame(() => requestAnimationFrame(() => resolve())),
+          ),
+      );
+
+    // The list renders and measures items to begin with.
+    await test.expect(page.locator(".item").first()).toBeVisible();
```
diegohaz: It should use semantic elements/roles with `q`

diegohaz: Still open on the current head (`0796cc4`). The assertions here still target implementation-detail selectors (`.item`, `#detached-count`) instead of the semantics a user or assistive technology would perceive — which is what `q` is for and what the sandbox testing guidance asks for.

Concrete, behavior-preserving path (keeps the CI-green flow intact):

- The count lives in an `<output>`, whose implicit ARIA role is `status`, so it's directly queryable:
  ```ts
  const detachedCount = q.status();
  // ...
  await test.expect(detachedCount).toHaveText("0");
  ```
- The rows are `<button>`s, so the initial "list rendered" check can be semantic too:
  ```ts
  await test.expect(q.button(/^item /).first()).toBeVisible();
  ```
  Anchor the pattern (`/^item /`) so it doesn't also match the **Refresh items** control.
- `.scroller` is reasonable to keep as a `page.locator(...)` — it's a plain scroll container with no semantic role and is only used for `scrollTop`/`scrollHeight` mechanics, not assertions. (If you want it semantic as well, give it `role="region"` + an accessible name and use `q.region(...)`.)

This only swaps the query mechanism; the scroll/refresh/unmount steps and the `toHaveText("0")` assertions stay identical.

diegohaz (commented):
**Initial review** (no prior `ariakit-pr-review` marker on this PR). Baseline: merge-base `caaa1463d` → head `0796cc4`. Reviewed the changeset, the `composite-renderer-6215` sandbox + style + browser test, and the `collection-renderer.tsx` fix — then the surrounding context: the `elements`/`data` model (`getData`/`getItemSize`/`computeData`), `elementObserver`, `itemRef`, `useForceUpdate`, the sibling renderer effects, package/changeset scope, and the existing self-review threads.

**Validation performed**
- Traced the data model to confirm the PR's core safety claim: dropping a no-longer-rendered node from `elements` can't change measured layout — `getItemSize` already returns `0` for a disconnected element, and `getData` falls back to the cached `data` via `itemData.rendered`. The reconcile effect correctly omits `updateElements()`, since the recomputed data is identical.
- Confirmed `useForceUpdate` is a stable `useReducer` dispatch, so `elementObserver` and the disconnect effect's `[elementObserver, elements]` deps are stable: the disconnect effect runs once per mount, and the StrictMode re-observe is genuinely load-bearing (effect cleanup can run without the item ref callbacks re-running).
- Verified the two cleanup paths are distinct and both needed — `itemRef`'s `prevElement` unobserve covers same-id remounts (key bump), the reconcile effect covers scroll-out — and that `Map` deletion during `for…of` is safe.
- Changeset scope is correct: `@ariakit/react-components` patch cascades a `@ariakit/react` patch through `workspace:*` (changeset-bot confirms 2 packages).
- CI is fully green on the head, including the Chrome/Firefox/Safari/macOS/Android suites that run the new `test-browser.ts`, plus Type check and Lint.

**Findings**
- No correctness, performance, compatibility, security, or simplification issues found in the library fix. It's minimal and well-reasoned, and it turns unbounded `elements` growth into bounded per-window reconciliation — a net performance improvement, not just a leak fix.
- One unresolved thread followed up (not a new inline comment): the author's own open note on `test-browser.ts` ("use semantic elements/roles with `q`") is still unaddressed on this head. Replied there with a concrete, behavior-preserving path (`q.status()` for the `<output>` count, `q.button(/^item /)` for rows; keep `.scroller` as a locator).

**Not fully validated**
- Did not run the Playwright browser test locally (worktree app-server cost) or run a revert-and-fail experiment. I relied on the green cross-browser CI run plus the test's direct `detachedCount === "0"` leak metric and the data-model reasoning above; the test's instrumentation (a `ResizeObserver` wrapper counting detached-but-observed `.item` nodes) makes a nonzero count without the fix logically unavoidable.

app/src/sandbox/composite-renderer-6215/test-browser.ts
diegohaz: Fixed in a03097748 — the assertions are now semantic: the detached count is read via `q.status()` (the `<output>`'s implicit role) and the rendered-rows check uses `q.button(/^item /).first()`. As you suggested, the scroll container stays a `page.locator(".scroller")` since it has no semantic role and is only used for `scrollTop`/`scrollHeight` mechanics, not assertions. (a37d09b17 first moved these to `q`; a03097748 refined it to match your guidance.)

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -734,10 +734,28 @@ export function useCollectionRenderer<T extends Item = any>({
...
+  // measured item nodes or keep firing resize callbacks. Re-observe the tracked
+  // elements on setup so observation survives a simulated unmount/remount (e.g.,
+  // React StrictMode), which runs this cleanup but doesn't necessarily re-run
+  // the item ref callbacks.
+  useEffect(() => {
+    for (const element of elements.values()) {
+      elementObserver?.observe(element);
+    }
+    return () => elementObserver?.disconnect();
+  }, [elementObserver, elements]);
```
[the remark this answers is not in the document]
diegohaz: Fixed in b5f48baa1 — the reconcile effect no longer early-returns when `itemSize` is set; it now computes an empty rendered-id set in that case and unobserves + drops every tracked node. So when `itemSize` switches from unset to a fixed size at runtime (`itemSize` is already in the effect deps), the previously-measured nodes are cleaned up instead of being retained. Behavior is unchanged for a statically set/unset `itemSize`.

diegohaz:
Addressed the latest review comments.

**Fixed**

- `a03097748` ([reply](https://github.com/ariakit/ariakit/pull/6225#discussion_r3352758161)): the browser test queries assertions semantically via `q` — `q.status()` for the detached count and `q.button(/^item /).first()` for the rendered rows — while the scroll container stays a `page.locator(".scroller")` (no semantic role; used only for scroll mechanics), per the guidance.
- `b5f48baa1` ([reply](https://github.com/ariakit/ariakit/pull/6225#discussion_r3352758225)): the reconcile effect now unobserves and forgets measured nodes when `itemSize` switches from unset to a fixed size at runtime, so they aren't retained (Copilot's edge case). Behavior is unchanged for a statically set/unset `itemSize`.

Also merged the latest `main` (picks up #6224's cross-orientation sizing fix and #6229's changeset-link fix).

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -747,10 +747,28 @@ export function useCollectionRenderer<T extends Item = any>({
...
```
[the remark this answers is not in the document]
diegohaz: This isn't an issue with the current code. That effect's deps are `[elementObserver, elements]` — both stable identities (`elementObserver` is a `useMemo` over the stable `updateElements` reducer dispatch; `elements` is a `useMemo(() => new Map(), [])`) — so it runs once on mount (plus StrictMode's simulated remount) and never re-runs on an `itemSize` change. An `itemSize` transition therefore doesn't re-execute the observe loop.

And whenever `itemSize` is set, `itemRef` early-returns before `elements.set` (so it never populates the map) and the reconcile effect clears `elements`, so the map is empty there and the loop is a no-op — nothing stale is re-observed and no extra `flushSync` callback fires.

Guarding the loop with `!itemSize` would require adding `itemSize` to the deps, which would make this effect disconnect and re-setup on every `itemSize` change for no behavioral gain, so I've left it as-is. Happy to add the guard if you'd prefer it for clarity.

diegohaz:
**Invalid (no change)**

- ([reply](https://github.com/ariakit/ariakit/pull/6225#discussion_r3352807505)): the disconnect effect's deps are `[elementObserver, elements]` (no `itemSize`), both stable, so it doesn't re-run on an `itemSize` transition — the observe loop never re-executes there. And `elements` is empty whenever `itemSize` is set (`itemRef` early-returns before populating it, and the reconcile effect clears it), so the loop is a no-op: no stale node is re-observed and no extra `ResizeObserver`/`flushSync` callback fires. Left as-is.

PR #6238, opened 2026-06-04
Speed up @ariakit/test simulated interactions

diegohaz:
## Motivation

`pnpm test-react` spends most of its time idling. Each simulated interaction — `click`, `type`, `press`, `hover`, `select` — pauses with a wall-clock `sleep()` (`requestAnimationFrame` + `setTimeout` + `requestAnimationFrame`) at every internal sub-step to let the DOM settle. `click` does this 4 times, `type` twice per character, and so on. But the components under test schedule their per-step updates with `queueMicrotask` and `requestAnimationFrame`, not wall-clock timers, so most of that `setTimeout` idle time isn't waiting for anything.

## Solution

Settle between an interaction's sub-steps without a wall-clock delay, keeping a real delay only for the final settle of each interaction (which stays load-bearing for cases like hiding a dialog by clicking outside).

- A new internal `settle()` yields across two animation frames and flushes microtasks — no `setTimeout`.
- The intermediate `sleep()` calls in the five helpers become `settle()`; each interaction still ends with a real `sleep()`. (`type` replaces its two per-character sleeps with `settle()` and keeps one real `sleep()` after the whole sequence.)
- The default final-settle delay drops from 8ms to 4ms.

Interactions that genuinely need real time (animations, tooltip/typeahead timeouts) use explicit `sleep(ms)` overrides, which are untouched.

## Results

On `main`, the React suite's cumulative test-body time drops ~25% (~195s → ~146s); wall-clock went ~19.3s → ~16.5s on a loaded local machine, and CI shards should see a similar per-shard reduction. The full suite (154 files, 742 tests) stays green across repeated runs with **zero retries**.

`settle()` is actually *stronger* than the intermediate `sleep()` it replaces at draining microtask-scheduled work — it adds an explicit `flushMicrotasks()` the old path lacked; it only drops the `setTimeout`. The new `react.test.tsx` locks in the per-keystroke settle correctness (the controlled-input cases fail if a per-step settle is dropped).

## Additional changes

- The `expect.poll(() => q.X(...))` assertions added here now use the new `.lazy` query helpers — `expect.poll(q.X.lazy(...))` — for readability.
- Renamed the React loader tests from `*.test.react.tsx` to `*.react.test.tsx` (the loader marker now precedes `.test`), with `interaction.test.react.tsx` becoming `react.test.tsx`. The vitest globs, the example-test naming regex, and the tsconfigs were updated accordingly.


packages/ariakit-test/src/select.ts
```diff
@@ -83,7 +83,8 @@ export function select(
       selection?.addRange(range);
     }
 
-    await sleep();
+    // Let the selection change flush before releasing — microtask/rAF settle.
+    await settle();
```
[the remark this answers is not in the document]
diegohaz: Not an issue. `select()` sets the new range with `addRange()` immediately before `settle()`, and per the [Selection API spec](https://w3c.github.io/selection-api/) the range is associated **synchronously** while the `selectionchange` *event* is queued as a task. So `document.getSelection()` returns the final selection in the `mouseUp`/`click` handlers, and a `selectionchange` listener also observes the final selection before the release — verified on both happy-dom and jsdom (and against Chromium/Firefox/WebKit, where `getSelection()` is likewise up to date in the release/click handlers).

The spec defines **no ordering** for `selectionchange` relative to `mouseup`/`click` and coalesces it to a single event (already asserted by `select.test.react.tsx`), so there's no "before release" guarantee being lost and no stale state.

packages/ariakit-test/src/click.ts
```diff
@@ -159,7 +159,10 @@ export function click(
     }
 
     if (!tap) {
-      await sleep();
+      // Press-and-release dwell between mouseDown and mouseUp: let work scheduled
+      // on pointer/mouse down flush (microtask/rAF) before releasing, without a
+      // wall-clock delay. The final settle below keeps the real timer.
+      await settle();
```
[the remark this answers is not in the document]
diegohaz: Not an issue. `mouseDown()`'s `removeAllRanges()` dissociates the range **synchronously**, so `document.getSelection()` already returns the cleared selection in the `mouseUp`/`click` handlers on both engines — the selection state read there is never stale.

The `selectionchange` *event* is a separate concern. Per the [Selection API spec](https://w3c.github.io/selection-api/) it's dispatched by *queuing a task on the user-interaction task source* (always async, never synchronous with the mutation), and the spec specifies **no ordering** relative to `mouseup`/`click`. I verified this in real browsers:

- **Chromium** updates `getSelection()` synchronously but fires `selectionchange` **after** `click`.
- **Firefox/WebKit** fire it **before** `mouseup`.

happy-dom with `settle()` now matches Chromium; jsdom matches Firefox/WebKit — both spec-conformant. (happy-dom's native *synchronous* dispatch inside `removeAllRanges` actually violates the spec; the shim's deferral is what makes it compliant.) Since the selection read via `getSelection()` is correct in the handlers everywhere and nothing in the suite depends on `selectionchange` listener ordering, there's no regression here.

diegohaz:
Triaged the unresolved review threads.

**Not an issue**
- selectionchange ordering in `select.ts` ([reply](https://github.com/ariakit/ariakit/pull/6238#discussion_r3365423313)): `select()` sets the new range with `addRange()` synchronously before `settle()`, so `getSelection()` and a `selectionchange` listener both see the final selection before the release. Per the [Selection API spec](https://w3c.github.io/selection-api/) `selectionchange` is async with no defined ordering vs `mouseup`/`click`; verified in Chromium/Firefox/WebKit.
- selectionchange ordering in `click.ts` ([reply](https://github.com/ariakit/ariakit/pull/6238#discussion_r3365423374)): the range is cleared synchronously in `mouseDown`, so `getSelection()` is up to date in the release/click handlers on both engines. The deferred `selectionchange` *event* now lands after the release on happy-dom — which matches Chromium (jsdom matches Firefox/WebKit); both spec-conformant. No stale state.

**Fixed**
- `7445ea4e8`: `examples/dialog-nested-various/test.ts` failed under React 18 ("show nested dismiss dialog and hide with escape") with focus stuck on the dialog container. Reverted the two `Close`-focus assertions from `expect.poll(...)` to `waitFor(...)`: `expect.poll` runs its predicate with React's act environment enabled, so the container→first-tabbable focus move never commits; `waitFor` runs inside `wrapAsync` (act environment disabled) and lets it flush. React 19's act semantics differ, so only React 18 was affected.

app/src/examples/dialog-framer-motion/test.ts
```diff
@@ -9,7 +9,7 @@ test("show/hide on click", async () => {
   await click(q.button("OK"));
   expect(q.dialog()).toBeVisible();
   expect(q.button("Show modal")).toHaveFocus();
-  await waitFor(() => expect(q.dialog()).not.toBeInTheDocument());
+  await expect.poll(() => q.dialog()).not.toBeInTheDocument();
```
diegohaz:
```suggestion
  await expect.poll(q.dialog).not.toBeInTheDocument();
```

examples/visually-hidden/test.ts
```diff
@@ -1,7 +1,4 @@
 // @vitest-environment jsdom
```
diegohaz, line 1:
```suggestion
// @vitest-environment jsdom
// happy-dom's CSS parser drops `clip: rect(0 0 0 0)` (real browsers and jsdom
// keep it), changing the serialized style. Pinned to jsdom so the snapshot
// reflects the real clip-rect hiding technique.
```

packages/ariakit-test/src/react.test.tsx
```diff
@@ -0,0 +1,88 @@
+import { cleanup } from "@testing-library/react";
```
diegohaz, line 1: Right now, react tests are named `*.test.react.tsx`. Let's change them to `*.react.test.tsx` (also accepting `react.test.tsx`). Rename all the existing cases and rename this one to just `react.test.tsx`.

examples/visually-hidden/test.ts
diegohaz: Fixed in faf42e653 — added the comment explaining why the snapshot is pinned to jsdom (happy-dom drops `clip: rect(0 0 0 0)`).

packages/ariakit-test/src/react.test.tsx
diegohaz: Fixed in f4e85b20f — renamed the React loader tests to `*.react.test.tsx` (`interaction.test.react.tsx` → `react.test.tsx`; `blur`/`select` → `blur.react.test.tsx`/`select.react.test.tsx`) and updated the vitest include/exclude globs, the example-test naming regex, and the tsconfig globs to discover the new form. Since `*.react.test.tsx` also matches the generic `*test.{ts,tsx}` pattern (the old `*.test.react.tsx` didn't), the no-loader suite now excludes `*react.test.*`/`*solid.test.*` so they only run under test-react/test-solid.

diegohaz:
**Fixed**

- `faf42e653` ([reply](https://github.com/ariakit/ariakit/pull/6238#discussion_r3366395064)): Added a comment explaining why `visually-hidden`'s snapshot is pinned to jsdom (happy-dom's CSS parser drops `clip: rect(0 0 0 0)`).
- `f4e85b20f` ([reply](https://github.com/ariakit/ariakit/pull/6238#discussion_r3366395129)): Renamed the React loader tests to the `*.react.test.tsx` convention — `interaction.test.react.tsx` → `react.test.tsx`, and `blur`/`select` → `*.react.test.tsx` — and updated the vitest include/exclude globs, the example-test naming regex, and the loader-test globs across the tsconfigs. Because the new names also match the generic `*test.{ts,tsx}` pattern (the old ones didn't), the no-loader suite now excludes `*react.test.*`/`*solid.test.*` so they only run under test-react/test-solid. Verified locally: test-react 742, test-solid 4, core 290 (loader tests excluded), type-check clean.

vitest.config.ts
```diff
@@ -35,12 +40,15 @@ function getFrameworkTestIncludes(loader: AllowedTestLoader) {
...
   });
+  // The first glob below already matches this loader's `*${loader}.test.tsx`
+  // files in the ariakit-${loader}* packages (they end in `test.{ts,tsx}`).
+  // ariakit-test isn't an ariakit-${loader}* package, so it needs its own
+  // loader-specific include.
   return [
     `packages/ariakit-${loader}*/src/**/*test.{ts,tsx}`,
-    `packages/ariakit-${loader}*/src/**/*test.${loader}.{ts,tsx}`,
-    `packages/ariakit-test/src/**/*test.${loader}.{ts,tsx}`,
+    `packages/ariakit-test/src/**/*${loader}.test.{ts,tsx}`,
```
diegohaz: Do we need specific `ariakit-test` path here? Should not we accept any `*${loader}.test.{ts,tsx}`?

diegohaz: Fixed in 38ddde01b — generalized it to `packages/*/src/**/*${loader}.test.{ts,tsx}` so loader-marked tests in any package are discovered, not just `ariakit-test`. The no-loader suite already excludes `*react.test.*`/`*solid.test.*` globally, so the broader include doesn't leak loader tests into the core suite. Test counts are unchanged (react 742, solid 4, core 290), since `ariakit-test` is the only package with loader tests today.

diegohaz:
**Fixed**

- `38ddde01b` ([reply](https://github.com/ariakit/ariakit/pull/6238#discussion_r3366442175)): Generalized the framework-include glob from the hardcoded `packages/ariakit-test/src/**/*${loader}.test.{ts,tsx}` to `packages/*/src/**/*${loader}.test.{ts,tsx}`, so loader-marked tests in any package are discovered. The global `*react.test.*`/`*solid.test.*` core excludes keep them out of the no-loader suite; test counts unchanged (react 742, solid 4, core 290).

**Already addressed** (from the prior round, replies in-thread)

- `faf42e653`: visually-hidden jsdom-pin comment.
- `f4e85b20f`: React loader test rename to `*.react.test.tsx`.

PR #6254, opened 2026-06-05
Fix mouseDown selection preservation

### LINKED ISSUE #6253: @ariakit/test: mouseDown clears the document selection for every target, but real browsers preserve it for non-text controls (e.g. buttons)
`mouseDown()` (used by `click()`) unconditionally clears any non-collapsed document selection via `removeAllRanges()` (`packages/ariakit-test/src/mouse-down.ts`):

```ts
// Remove current selection
const selection = getDocument(element).getSelection();
if (selection?.rangeCount) {
  const range = selection.getRangeAt(0);
  if (!range.collapsed) {
    selection.removeAllRanges();
  }
}
```

In real browsers, whether a `mousedown` clears the document selection is **target-dependent**: clicking a text-selectable/editable target (an `<input>`, or plain text) collapses the selection, but clicking a non-text control like a `<button>` **preserves** it.

### Evidence (Chromium / Firefox / WebKit via Playwright)

With a non-collapsed selection over a `<p>`:

- Click a `<button>` → selection **preserved** in all three engines (no `selectionchange` fires).
- Click an `<input>` / other text → selection **cleared** in all three engines.

`@ariakit/test` clears the selection in all of these cases, so it diverges from real browsers when clicking a button (and other targets that wouldn't start a new selection or move a caret).

### Impact

Low — a non-collapsed document selection existing at click time is uncommon in tests, and no current test/component depends on the selection persisting across a click. Filing mainly to record the fidelity gap found while reviewing #6238.

### Notes

Faithful emulation is target-dependent (clear on text/editable mousedown or when the press starts a new selection; preserve otherwise), so this is more involved than the current unconditional clear — reasonably low-priority / possibly won't-fix.

### Workaround

https://github.com/ariakit/ariakit/issues/6253#issuecomment-4636016741

diegohaz:
## Motivation

`@ariakit/test` cleared any non-collapsed document selection during [`mouseDown`](https://ariakit.com/reference/mouse-down), regardless of the pressed target. Real browsers preserve an existing text selection when pressing native non-text controls such as buttons, checkboxes, selects, and links, while text/editable targets and plain text targets clear the selection.

Fixes #6253.

## Solution

Update [`mouseDown`](https://ariakit.com/reference/mouse-down) so selection clearing depends on the pressed target:

- Native non-text controls such as buttons, checkboxes, selects, and links preserve the current document selection.
- Text/editable targets, text-like input types, focusable custom text targets, and plain text targets clear the document selection.
- The normal focus path still runs for native controls, so this does not rely on canceling `mousedown`.

The package-level regression tests cover native button, checkbox, select, and link preservation, plus clearing for editable text inputs, email inputs, focusable text, and plain text targets.

## Workaround

Until `@ariakit/test` with this fix is available, prevent the default `mousedown` behavior on the button target whose click should preserve the current document selection:

```ts
// TODO: Remove workaround for https://github.com/ariakit/ariakit/issues/6253.
// This preserves selection, but it also skips the focus path.
button.addEventListener("mousedown", (event) => {
  event.preventDefault();
});
```

This preserves the selected text because `mouseDown()` skips its default behavior when `mousedown` is canceled. It also skips the simulated focus path, so it is only a selection-preservation workaround.

### DIFF
--- a/packages/ariakit-test/src/mouse-down.test.ts
+++ b/packages/ariakit-test/src/mouse-down.test.ts
@@ -5,7 +5,7 @@ import { select } from "./select.ts";
 
 const selectionText = "sample paragraph text";
 
-function setup() {
+beforeEach(() => {
   document.body.innerHTML = `
     <p>Keep this ${selectionText} selected.</p>
     <button type="button">Preserve selection</button>
@@ -36,9 +36,7 @@ function setup() {
     <div tabindex="0">Focusable text target</div>
     <p>Plain text target</p>
   `;
-}
-
-beforeEach(setup);
+});
 
 async function selectParagraphText() {
   await select(selectionText, q.text.ensure(/Keep this/));
--- a/packages/ariakit-test/src/mouse-down.test.ts
+++ b/packages/ariakit-test/src/mouse-down.test.ts
@@ -1,4 +1,4 @@
-import { expect, test } from "vitest";
+import { beforeEach, expect, test } from "vitest";
 import { mouseDown } from "./mouse-down.ts";
 import { q } from "./query.ts";
 import { select } from "./select.ts";
@@ -38,13 +38,14 @@ function setup() {
   `;
 }
 
+beforeEach(setup);
+
 async function selectParagraphText() {
   await select(selectionText, q.text.ensure(/Keep this/));
   expect(document.getSelection()?.toString()).toBe(selectionText);
 }
 
 test("mouseDown on a button preserves the document selection", async () => {
-  setup();
   await selectParagraphText();
   await mouseDown(q.button("Preserve selection"));
   expect(document.getSelection()?.toString()).toBe(selectionText);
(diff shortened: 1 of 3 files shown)


packages/ariakit-test/src/mouse-down.ts
```diff
@@ -3,13 +3,47 @@ import {
...
+  "time",
+  "url",
+  "week",
+];
+
+function preservesSelectionOnMouseDown(element: Element) {
+  const control = element.closest<HTMLElement>("button,input,select,a[href]");
+  if (control instanceof HTMLButtonElement) return true;
+  if (control instanceof HTMLSelectElement) return true;
+  if (control instanceof HTMLAnchorElement) return true;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 878e48049 - anchor detection now uses the matched element tag name so SVG `<a href>` elements preserve selection too, and `mouse-down.test.ts` includes an SVG link regression case.

diegohaz:
**Fixed**

- 878e48049 ([reply](https://github.com/ariakit/ariakit/pull/6254#discussion_r3365735132)): Updated anchor detection to cover SVG `<a href>` elements and added an SVG link regression test for `mouseDown()` selection preservation.

packages/ariakit-test/src/mouse-down.test.ts
```diff
@@ -0,0 +1,108 @@
+import { expect, test } from "vitest";
+import { mouseDown } from "./mouse-down.ts";
+import { q } from "./query.ts";
+import { select } from "./select.ts";
+
+const selectionText = "sample paragraph text";
+
+function setup() {
```
diegohaz: Shouldn't this use `beforeEach` instead?

diegohaz: Fixed in 78e50bf72 - moved the repeated test DOM setup into Vitest `beforeEach`, so every mouseDown regression case still starts with a fresh DOM without per-test setup calls.

diegohaz:
**Fixed**

- 78e50bf72 ([reply](https://github.com/ariakit/ariakit/pull/6254#discussion_r3365945478)): Moved the repeated `mouseDown` regression test DOM setup into Vitest `beforeEach`, keeping each case isolated without per-test setup calls.

PR #6255, opened 2026-06-05
Add lazy test queries

diegohaz:
## Motivation

`@ariakit/test` queries can be stored as elements today, but that captures the current DOM result instead of re-running the lookup later. Tests that need to reuse a query across DOM updates currently have to wrap the query by hand:

```ts
const dialog = () => q.dialog("Name");
```

## Solution

Add a `.lazy` helper to the DOM query API so role, text, and label queries can produce reusable callbacks directly:

```ts
const dialog = q.dialog.lazy("Name");
expect(dialog()).toBeVisible();
```

The new helper is available on query variants such as `.all`, `.wait`, `.ensure`, and role `.includesHidden`. Text and label lazy queries preserve their generic element typing, and the implementation avoids relying on newer Testing Library generic aliases so the package remains compatible with its supported `@testing-library/dom` range.

Tests cover re-querying after DOM changes, scoped `within` queries, role variants, and generic text/label lazy queries.

### DIFF
--- a/packages/ariakit-test/src/query.ts
+++ b/packages/ariakit-test/src/query.ts
@@ -304,12 +304,13 @@ function createQueryObject(queries = documentQueries): QueryObject {
  * elements.
  * @example
  * ```ts
- * await click(query.button("Open"));
- * const dialog = query.dialog.lazy();
+ * const dialog = query.dialog.lazy("Settings");
+ * expect(dialog()).not.toBeInTheDocument();
+ * await click(query.button("Open settings"));
  * expect(dialog()).toBeVisible();
  * // Wait for an element to appear, or scope a query to a subtree:
  * await query.alert.wait();
- * query.within(query.dialog()).button("Close");
+ * query.within(dialog()).button("Close");
  * ```
  */
 export const query = createQueryObject();
@@ -319,8 +320,9 @@ export const query = createQueryObject();
  * label.
  * @example
  * ```ts
- * await click(q.button("Open"));
- * const dialog = q.dialog.lazy();
+ * const dialog = q.dialog.lazy("Settings");
+ * expect(dialog()).not.toBeInTheDocument();
+ * await click(q.button("Open settings"));
  * expect(dialog()).toBeVisible();
  * ```
  */
--- a/packages/ariakit-test/src/query.ts
+++ b/packages/ariakit-test/src/query.ts
@@ -1,9 +1,11 @@
-// oxlint-disable unbound-method
+// oxlint-disable unbound-method typescript/no-unnecessary-type-parameters
 import { invariant } from "@ariakit/utils";
 import type {
   ByRoleMatcher,
   ByRoleOptions,
-  getQueriesForElement,
+  Matcher,
+  SelectorMatcherOptions,
+  waitForOptions,
 } from "@testing-library/dom";
 import { queries as baseQueries } from "@testing-library/dom";
 import type { AriaRole } from "./__aria-role.ts";
@@ -11,17 +13,130 @@ import { roles } from "./__aria-role.ts";
 
 type Query = ReturnType<typeof createRoleQuery>;
 type TextQuery = ReturnType<typeof createTextQuery>;
-type LabeledQuery = ReturnType<typeof createLabeledQuery>;
 type RoleQueries = Record<AriaRole, Query>;
-type ElementQueries = ReturnType<
-  typeof getQueriesForElement<typeof baseQueries>
->;
+type AnyQuery = (...args: any[]) => any;
+type LazyQuery<T extends AnyQuery> = (
+  ...args: Parameters<T>
+) => () => ReturnType<T>;
+
+interface ElementQueries {
+  queryByRole: RoleNullableMethod;
+  queryAllByRole: RoleArrayMethod;
+  findByRole: RoleWaitValueMethod;
+  findAllByRole: RoleWaitArrayMethod;
+  getByRole: RoleValueMethod;
+  getAllByRole: RoleArrayMethod;
+  queryByText: TextNullableMethod;
+  queryAllByText: TextArrayMethod;
+  findByText: TextWaitValueMethod;
+  findAllByText: TextWaitArrayMethod;
+  getByText: TextValueMethod;
+  getAllByText: TextArrayMethod;
+  queryByLabelText: TextNullableMethod;
+  queryAllByLabelText: TextArrayMethod;
+  findByLabelText: TextWaitValueMethod;
+  findAllByLabelText: TextWaitArrayMethod;
+  getByLabelText: TextValueMethod;
+  getAllByLabelText: TextArrayMethod;
+}
+
+interface RoleNullableMethod {
+  (role: ByRoleMatcher, options?: ByRoleOptions): HTMLElement | null;
+}
+
+interface RoleArrayMethod {
+  (role: ByRoleMatcher, options?: ByRoleOptions): HTMLElement[];
+}
+
+interface RoleValueMethod {
+  (role: ByRoleMatcher, options?: ByRoleOptions): HTMLElement;
+}
+
+interface RoleWaitArrayMethod {
+  (
+    role: ByRoleMatcher,
+    options?: ByRoleOptions,
+    waitForElementOptions?: waitForOptions,
+  ): Promise<HTMLElement[]>;
+}
+
+interface RoleWaitValueMethod {
+  (
+    role: ByRoleMatcher,
+    options?: ByRoleOptions,
+    waitForElementOptions?: waitForOptions,
+  ): Promise<HTMLElement>;
+}
+
+interface TextNullableMethod {
+  <T extends HTMLElement = HTMLElement>(...args: TextQueryArgs): T | null;
+}
+
+interface TextNullableQuery extends TextNullableMethod {
+  lazy<T extends HTMLElement = HTMLElement>(
+    ...args: TextQueryArgs
+  ): () => T | null;
+}
+
+interface TextArrayMethod {
+  <T extends HTMLElement = HTMLElement>(...args: TextQueryArgs): T[];
+}
+
+interface TextArrayQuery extends TextArrayMethod {
+  lazy<T extends HTMLElement = HTMLElement>(...args: TextQueryArgs): () => T[];
+}
+
+interface TextValueMethod {
+  <T extends HTMLElement = HTMLElement>(...args: TextQueryArgs): T;
+}
+
+interface TextValueQuery extends TextValueMethod {
+  lazy<T extends HTMLElement = HTMLElement>(...args: TextQueryArgs): () => T;
+}
+
+interface TextWaitValueMethod {
+  <T extends HTMLElement = HTMLElement>(...args: TextWaitQueryArgs): Promise<T>;
+}
+
+interface TextWaitValueQuery extends TextWaitValueMethod {
+  lazy<T extends HTMLElement = HTMLElement>(
+    ...args: TextWaitQueryArgs
+  ): () => Promise<T>;
+}
+
+interface TextWaitArrayMethod {
+  <T extends HTMLElement = HTMLElement>(
+    ...args: TextWaitQueryArgs
+  ): Promise<T[]>;
+}
+
+interface TextWaitArrayQuery extends TextWaitArrayMethod {
+  lazy<T extends HTMLElement = HTMLElement>(
+    ...args: TextWaitQueryArgs
+  ): () => Promise<T[]>;
+}
+
+interface TextQueryParams {
+  query: TextNullableMethod;
+  all: TextArrayMethod;
+  wait: TextWaitValueMethod;
+  waitAll: TextWaitArrayMethod;
+  ensure: TextValueMethod;
+  ensureAll: TextArrayMethod;
+}
 
 interface QueryObject extends RoleQueries {
   text: TextQuery;
-  labeled: LabeledQuery;
+  labeled: TextQuery;
   within: (element?: HTMLElement | null) => QueryObject;
 }
+
+type TextQueryArgs = [id: Matcher, options?: SelectorMatcherOptions];
+type TextWaitQueryArgs = [
+  id: Matcher,
+  options?: SelectorMatcherOptions,
+  waitForElementOptions?: waitForOptions,
+];
 
 function createQueries(container?: HTMLElement) {
   return Object.entries(baseQueries).reduce((queries, [key, query]) => {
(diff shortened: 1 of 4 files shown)


packages/ariakit-test/src/query.ts
```diff
@@ -1,28 +1,143 @@
...
+  <T extends HTMLElement = HTMLElement>(
+    ...args: TextWaitQueryArgs
+  ): Promise<T[]>;
+}
+
+interface TextWaitArrayQuery extends TextWaitArrayMethod {
+  lazy<T extends HTMLElement = HTMLElement>(
+    ...args: TextWaitQueryArgs
+  ): () => Promise<T[]>;
+}
```
diegohaz: **Yes — the text/label query types can be collapsed by ~5×, keeping the exact same public types.**

This block plus the five `createText*Query` factories below (lines 161-203) are the bulk of the new complexity, and all of it exists only to preserve the `<T extends HTMLElement>` generic on `.lazy`. The five `Text*Method` interfaces, five `Text*Query` interfaces, and five factories are the same shape repeated once per return shape (`T | null`, `T`, `T[]`, `Promise<T>`, `Promise<T[]>`).

You can encode the "return shape" as a `Kind` token and map it to the result with an indexed type — the standard stand-in for the higher-kinded type TS can't express — so all fifteen declarations become one method type, one variant type, and one factory:

```ts
type TextKind = "nullable" | "value" | "array" | "waitValue" | "waitArray";

// Maps a query kind to the element-typed result it returns, standing in for the
// higher-kinded type TypeScript can't express directly.
type TextResult<Kind extends TextKind, T extends HTMLElement> = {
  nullable: T | null;
  value: T;
  array: T[];
  waitValue: Promise<T>;
  waitArray: Promise<T[]>;
}[Kind];

interface TextMethod<Kind extends TextKind, Args extends unknown[]> {
  <T extends HTMLElement = HTMLElement>(...args: Args): TextResult<Kind, T>;
}

interface TextVariant<Kind extends TextKind, Args extends unknown[]>
  extends TextMethod<Kind, Args> {
  lazy<T extends HTMLElement = HTMLElement>(
    ...args: Args
  ): () => TextResult<Kind, T>;
}

function assignTextLazy<Kind extends TextKind, Args extends unknown[]>(
  query: TextMethod<Kind, Args>,
): TextVariant<Kind, Args> {
  return Object.assign(query, {
    lazy: <T extends HTMLElement = HTMLElement>(...args: Args) => {
      return () => query<T>(...args);
    },
  });
}
```

The reference sites then become mechanical: `ElementQueries`/`TextQueryParams` fields use `TextMethod<"nullable", TextQueryArgs>` etc., and `createTextQuery` calls `assignTextLazy(params.query)` for each of the six variants (replacing the six `createText*Query(...)` calls).

I applied this on top of your branch and validated equivalence rather than guessing:
- `tsc -p packages/ariakit-test/tsconfig.node.json` clean; oxlint clean; the 4 new tests pass unchanged.
- Public types are identical. I ran 17 `Expect<Equal<…>>` assertions — default **and** explicit generics on `text`/`labeled`/role queries across every `.lazy`/`.all`/`.wait`/`.ensure` chain — against both your version and the simplified one, and both compile with no mismatch. So `q.text.lazy<HTMLButtonElement>("Submit")` still resolves to `() => HTMLButtonElement | null`, `q.text.all.wait.lazy<…>()` to `() => Promise<…[]>`, etc.
- `query.ts` drops from 392 → 333 lines, with one source of truth for the lazy wrapper instead of five copies.

One thing I intentionally left alone: `ElementQueries` and the hand-rolled role method interfaces. Reverting those to `BoundFunctions<typeof baseQueries>` would regress on `@testing-library/dom@^8`, where `BoundFunctions` doesn't preserve the element generic (the per-query generic overloads landed in a later major). Keeping `ElementQueries` hand-rolled is correct for the supported `^8 || ^9 || ^10` range — this suggestion only deduplicates the layer built on top of it.

(If you ever decide the per-element generic on `.lazy` isn't worth carrying, the text queries could reuse the role `assignLazy` and this whole layer disappears — but the test deliberately exercises `q.text.lazy<HTMLButtonElement>`, so the generic-preserving version above keeps that feature.)

diegohaz (commented):
**Review — initial, comment-only.** Scope: full diff of PR #6255 against merge base `origin/main` (the 4 changed files: `query.ts`, `query.test.ts`, `readme.md`, the changeset), with surrounding context in `packages/ariakit-test` and the `@testing-library/dom` type definitions.

You asked specifically whether the types can be simplified. **Yes** — and the answer is one substantive finding (posted inline on the text-query type block):

- The five `Text*Method` interfaces + five `Text*Query` interfaces + five `createText*Query` factories collapse to a single `TextMethod<Kind, Args>` / `TextVariant<Kind, Args>` pair plus one `assignTextLazy` helper, by encoding the return shape as a `Kind` token mapped through an indexed `TextResult` type. `query.ts` goes 392 → 333 lines with one source of truth instead of five copies.

**Validation performed (on a local copy of your branch, reverted afterward):**
- `tsc -p packages/ariakit-test/tsconfig.node.json --noEmit` — clean on both your version and the simplified one.
- `vitest run packages/ariakit-test/src/query.test.ts` — 4/4 pass, unchanged.
- Type equivalence: 17 `Expect<Equal<…>>` assertions covering default + explicit element generics on `text`/`labeled`/role queries across every `.lazy`/`.all`/`.wait`/`.ensure`/`.includesHidden` chain — compile identically against both versions, so the public API types are unchanged (`q.text.lazy<HTMLButtonElement>(…)` still `() => HTMLButtonElement | null`).
- `oxlint` — clean.

**What I checked and found fine:**
- Correctness of the `.lazy` semantics (re-runs the query per call, captures args, respects `within` scoping and role variants) — covered by the new tests and consistent with the role `assignLazy`/`createLazyQuery` design.
- Reuse of `createTextQuery` for both `text` and `labeled` via `TextQueryParams` is good DRY.
- Keeping `ElementQueries` hand-rolled (rather than `BoundFunctions<typeof baseQueries>`) is the right call: `BoundFunctions` only preserves the element generic in newer `@testing-library/dom` majors, and this package supports `^8 || ^9 || ^10`. I did **not** suggest changing it.
- Changeset (patch, correct package), README, and JSDoc examples are accurate and match the implementation.

**Not fully validated:** I only exercised `@testing-library/dom@10.4.1` (the installed version); I reasoned about the `^8`/`^9` generic-preservation behavior from the type sources but did not install those majors to confirm. This affects the existing `ElementQueries` design, not the simplification above — the suggested change stays entirely within your hand-rolled types, so it's version-independent for the same reasons your current code is.

Overall this is a clean, well-tested PR; the only feedback is the optional type dedup.

packages/ariakit-test/src/query.ts
diegohaz: Fixed in 5e7129e6d - collapsed the text/label lazy query type layer into a single TextKind/TextResult/TextVariant helper while keeping the v8-compatible ElementQueries typing.

diegohaz:
**Fixed**
- 5e7129e6d ([reply](https://github.com/ariakit/ariakit/pull/6255#discussion_r3365776683)): Simplified the text/label lazy query typing to one generic variant factory while preserving the supported Testing Library v8-compatible query signatures.

packages/ariakit-test/src/query.ts
```diff
@@ -124,44 +235,56 @@ function createRoleQueries(queries = documentQueries) {
...
-  const all = Object.assign(queries.queryAllByLabelText, {
-    wait: queries.findAllByLabelText,
-    ensure: queries.getAllByLabelText,
-  });
-
-  const wait = Object.assign(queries.findByLabelText, {
-    all: queries.findAllByLabelText,
+function createTextQueryObject(queries = documentQueries) {
+  return createTextQuery({
+    query: queries.queryByText,
```
diegohaz: Why do we need `query`?

packages/ariakit-test/src/query.ts
```diff
@@ -124,44 +235,56 @@ function createRoleQueries(queries = documentQueries) {
...
+    all: queries.queryAllByText,
+    wait: queries.findByText,
+    waitAll: queries.findAllByText,
```
diegohaz, line 264: Isn't this already covered by `wait.all`?

packages/ariakit-test/src/query.ts
```diff
@@ -124,44 +235,56 @@ function createRoleQueries(queries = documentQueries) {
...
+    ensure: queries.getByText,
+    ensureAll: queries.getAllByText,
```
diegohaz, line 266: Same here?

packages/ariakit-test/src/query.ts
diegohaz: Fixed in 2f41ea299 - all.wait now reuses wait.all directly while still preserving the findAllBy* variant needed for the all-returning wait query.

packages/ariakit-test/src/query.ts
diegohaz: Fixed in 2f41ea299 - kept the queryBy* function as the direct default query argument for q.text()/q.labeled() semantics instead of carrying it as a TextQueryParams property.

packages/ariakit-test/src/query.ts
diegohaz: Fixed in 2f41ea299 - all.ensure now reuses ensure.all directly while still preserving the getAllBy* variant needed for the all-returning ensure query.

diegohaz:
**Fixed**
- 2f41ea299 ([reply](https://github.com/ariakit/ariakit/pull/6255#discussion_r3365976828)): Moved the direct `queryBy*` function out of `TextQueryParams` and passed it as the default query argument for `q.text()`/`q.labeled()` semantics.
- 2f41ea299 ([reply](https://github.com/ariakit/ariakit/pull/6255#discussion_r3365976826)): Made `all.wait` reuse `wait.all` while preserving the `findAllBy*` wait-all variant.
- 2f41ea299 ([reply](https://github.com/ariakit/ariakit/pull/6255#discussion_r3365976829)): Made `all.ensure` reuse `ensure.all` while preserving the `getAllBy*` ensure-all variant.

Added focused coverage for lazy text and label variant paths.

packages/ariakit-test/src/query.ts
```diff
@@ -178,14 +297,16 @@ function createQueryObject(queries = documentQueries): QueryObject {
...
+ * Every query also exposes `.lazy` (return a reusable function that runs the
+ * query when called), `.all` (return all matches), `.wait` (resolve once the
  * element appears), and `.ensure` (throw when it's missing) variants, and role
  * queries additionally expose `.includesHidden` to include otherwise-hidden
  * elements.
  * @example
  * ```ts
  * await click(query.button("Open"));
- * expect(query.dialog()).toBeVisible();
+ * const dialog = query.dialog.lazy();
```
diegohaz: Minor, documentation only — non-blocking.

The headline example now introduces `.lazy` where it doesn't pay off: `dialog` is called exactly once, so `const dialog = query.dialog.lazy(); expect(dialog()).toBeVisible();` is strictly more verbose than the previous `expect(query.dialog()).toBeVisible();`, and the same example keeps the direct form a few lines later (`query.within(query.dialog())`). A reader's first encounter with `.lazy` is one where it looks like pure overhead, which undersells the feature whose whole point (per the PR motivation) is reusing a query across DOM updates.

Since this JSDoc is the source for both the `q` example below and the generated `readme.md`, consider either keeping the direct call here and showcasing `.lazy` where its value is visible, or making this snippet demonstrate reuse, e.g.:

```ts
// Store a query once and re-run it as the DOM changes:
const dialog = query.dialog.lazy("Settings");
await click(query.button("Open settings"));
expect(dialog()).toBeVisible();
await click(query.button("Close"));
expect(dialog()).not.toBeInTheDocument();
```

Not blocking — the implementation, types, and prose are all correct; this is purely about the example teaching *when* to reach for `.lazy`.

diegohaz (commented):
**Re-review — comment-only.**

**Baseline:** incremental from the last marked review at `77baa5bd` (the initial commit) to current head `2f41ea29`, i.e. commits `5e7129e6` (Simplify lazy text query types) and `2f41ea29` (Clarify lazy text query variants). Also re-checked the whole `query.ts`/`query.test.ts` against the merge base `eb5a29ed`.

### Prior threads — all verified resolved in the current code (not just trusting the "Fixed" replies)
- **Text/label type duplication** (collapse ~5×): done in `5e7129e6`. The 15 `Text*Method`/`Text*Query` declarations + 5 `createText*Query` factories are now a single `TextKind`/`TextResult` token plus `TextMethod`/`TextVariant`/`assignTextLazy`, and the public types are unchanged (`text` and `labeled` both resolve to `TextQuery`).
- **"Why do we need `query`?"**: resolved in `2f41ea29` — the default `queryBy*` is now a direct `defaultQuery` parameter of `createTextQuery`, no longer a `TextQueryParams.query` property.
- **"Isn't this already covered by `wait.all` / `ensure.all`?"** (lines 264/266): resolved — `all.wait` now aliases `wait.all` and `all.ensure` aliases `ensure.all` (same object identity), matching the base behavior where both pointed at `findAllBy*`/`getAllBy*`.

- `pnpm install` in the worktree, then `vitest run packages/ariakit-test/src/query.test.ts` → **5/5 pass** (core suite, happy-dom). Confirms lazy re-querying after DOM mutation, `within` scoping, role variants (`ensure`/`includesHidden`/`all`), and the generic text/label variants.
- `tsc -b packages/ariakit-test` → exit 0. Full root `tsc -b` → no errors in `ariakit-test`; `query.test.ts` is type-checked via the root `tsconfig.test.json` (`**/*test.ts`), so the `q.text.lazy<HTMLButtonElement>()` generic threading is actually enforced. (The 21 errors the bare `tsc -b` reports are all pre-existing `app/` astro-sync/codegen artifacts — the real script is `pnpm -F app run check && tsc -b`.)
- `oxlint --disable-nested-config` + `oxfmt --check` on both changed files → clean. The added `typescript/no-unnecessary-type-parameters` disable is justified: the `<T extends HTMLElement = HTMLElement>` params are legitimate return-position generics that let callers pick the element type.
- Swept call sites: no existing `.lazy` usage in the repo yet, and existing `q.text(...)`/`q.labeled(...)` consumers stay type-compatible (`q.text("x")` still `HTMLElement | null`, `q.text.ensure("x")` still `HTMLElement`).

### Findings
- One **minor, non-blocking documentation** comment on the JSDoc/readme headline example (`query.ts:308`): it now features `.lazy` in a one-shot context where it's strictly more verbose than the prior direct call and doesn't show the reuse benefit. Optional.

### Design notes (checked, no change requested)
- The `assignLazy` (role) vs `assignTextLazy` (text/label) split is runtime-identical but kept separate so the text path can thread the element generic without an `as` cast — a reasonable trade-off, not duplication worth removing.
- Mutating query functions via `Object.assign` to attach `.lazy` matches the file's existing idiom and stays contained to per-call `createQueries` scopes (`within` gets fresh functions).
- Changeset: `200-` (minor/feature ordering) + `patch` frontmatter is correct for a v0 package; plain code spans (no reference link) are right since the `@ariakit/test` query API has no `ariakit.com/reference/*` page.

### Not fully validated
- Did not run the React/React18/Solid render suites or the `app` browser suite — this change is types-only at runtime and adds a core-suite test; nothing in those suites exercises `.lazy`.
- Did not exhaustively exercise every role async-lazy permutation (`wait.lazy`, `all.wait.lazy`); they're structurally identical to the tested `ensure`/`all`/`includesHidden` lazy variants and the equivalent text permutations are covered.

Overall this is clean, well-tested, and type-safe; the only follow-up is the optional doc-example nit.

packages/ariakit-test/src/query.ts
diegohaz: Fixed in 7c9ba3a7e - updated the query/q examples so the stored lazy query is called before and after opening the dialog, and the main example reuses dialog() for the scoped close-button query.

diegohaz:
**Fixed**
- 7c9ba3a7e ([reply](https://github.com/ariakit/ariakit/pull/6255#discussion_r3366094927)): Updated the lazy query examples to demonstrate reusing a stored query across DOM changes, including the scoped dialog query in the main example.

PR #6256, opened 2026-06-06
Apply @ariakit/test browser shims at import instead of per interaction

diegohaz:
## Motivation

`@ariakit/test` installs browser shims (most importantly a `getClientRects` visibility shim that jsdom lacks) so that visibility and focusability checks work. Until now those shims were only installed while a simulated interaction ran — inside `wrapAsync` — and torn down afterwards.

But components read layout and focusability _between_ interactions too. A dialog's auto-focus effect, for instance, picks the element to focus with `getFirstTabbableIn`, which relies on `getClientRects` to decide whether an element is visible:

```ts
const element =
  getElementFromProp(initialFocus, true) ||
  getFirstTabbableIn(contentElement, true, ...) || // ← needs getClientRects
  contentElement; // ← fallback: the dialog container itself
```

When that effect runs outside a `wrapAsync` window — e.g. while asserting with `expect.poll`, or when React 18 flushes the effect after the interaction already settled — `getClientRects` returns jsdom's empty layout, so `getFirstTabbableIn` finds nothing and focus falls back to the dialog container and stays there. This makes assertions like the following fail even though the dialog auto-focuses correctly in a real browser:

```ts
await click(q.button("Open dialog"));
await expect.poll(() => q.button("Close")).toHaveFocus(); // ❌ stuck on the dialog container
```

## Solution

Install the shims once, for the whole test environment, via a new `shims.ts` module that `index.ts` imports for its side effect — instead of installing and removing them around each interaction. `wrapAsync` now only toggles React's act environment. With the shims always present, layout and focusability reads behave consistently whenever component effects run, not only during interactions.

```ts
// index.ts
import "./shims.ts"; // applies the shims once, on import
```

`shims.ts` is also a public entrypoint, `@ariakit/test/shims`, so consumers that import only individual helpers (e.g. `@ariakit/test/click`) can apply the shims explicitly. `"sideEffects": false` is removed from `package.json` so bundlers keep the side-effect import.

Because the shims now run at import time, `applyBrowserPolyfills` guards against a missing or partial DOM and no-ops when the package is imported in a plain Node context.


.changeset/300-test-shims-on-import.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/test": patch
+---
+
+Fixed the browser shims to apply for the entire test instead of only while a simulated interaction runs. Component code that reads layout or focusability between interactions — such as a dialog's auto-focus picking the first tabbable element, which relies on the `getClientRects` visibility shim — now behaves consistently, including when asserting outside an interaction with `expect.poll`.
```
diegohaz: Where's the multiline changeset heading?

packages/ariakit-test/src/__utils.ts
```diff
@@ -43,6 +43,16 @@ export function setActEnvironment(value: boolean) {
 
 export function applyBrowserPolyfills() {
```
diegohaz: We should move all the apply browser polyfills (use `shims` instead of `polyfills` logic to the `shims.ts` file.

.changeset/300-test-shims-on-import.md
diegohaz: Fixed in 4f255fc52 — reformatted the changeset with a short heading followed by body paragraphs.

packages/ariakit-test/src/__utils.ts
diegohaz: Fixed in 91418f6f8 — moved `applyBrowserPolyfills` (now `applyBrowserShims`), the happy-dom shim helpers, and the `window.alert` shim into `shims.ts`. `__utils.ts` keeps only environment detection, the flush/frame helpers, and `wrapAsync`.

packages/ariakit-test/src/__utils.ts
```diff
@@ -230,25 +249,17 @@ function patchHappyDOMSelectionChange() {
...
-      removeBrowserPolyfills();
-    };
-  } catch (error) {
-    wrapAsyncDepth -= 1;
-    restoreActEnvironment();
-    throw error;
+  if (wrapAsyncDepth === 0) {
+    restoreCurrentWrapAsyncEnvironment = setActEnvironment(false);
   }
+  wrapAsyncDepth += 1;
```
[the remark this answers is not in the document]
diegohaz: Fixed in c8ea9943a — instead of a per-helper opt-in, `@ariakit/test` now exposes only the `index`, `react`, and `playwright` entry points, so there are no helper subpaths that can bypass the at-import shims. The helpers stay available from the root `@ariakit/test` (and `@ariakit/test/react`), both of which load the shims; `@ariakit/test/playwright` is unaffected since Playwright runs in real browsers where the shims are skipped.

diegohaz:
**Fixed**

- `4f255fc52` ([reply](https://github.com/ariakit/ariakit/pull/6256#discussion_r3366186357)): Reformatted the changeset with a short heading followed by body paragraphs.
- `91418f6f8` ([reply](https://github.com/ariakit/ariakit/pull/6256#discussion_r3366186403)): Moved the browser-shim logic — `applyBrowserPolyfills` (now `applyBrowserShims`), the happy-dom shim helpers, and the `window.alert` shim — from `__utils.ts` into `shims.ts`. `__utils.ts` keeps only environment detection, the flush/frame helpers, and `wrapAsync`.
- `c8ea9943a` ([reply](https://github.com/ariakit/ariakit/pull/6256#discussion_r3366186442)): Resolved the subpath-shims regression by reducing `@ariakit/test` to only the `index`, `react`, and `playwright` entry points, so no helper subpath can bypass the at-import shims (the helpers stay available from the root `@ariakit/test`, which loads them). Added an `--entries` build flag (generalizing `--index-only`). The changeset is now a `minor` `**BREAKING**` entry documenting the subpath removal.

PR #6269, opened 2026-06-06
Add browser test duplicates for sandbox happy-dom tests

diegohaz:
## Motivation

Browser tests (`test-browser.ts`) are the preferred, primary coverage for sandbox regressions, with a faster happy-dom `test.ts` kept alongside as a duplicate (see the `tooltip-4894` sandbox, which pairs both). Several sandboxes under `app/src/sandbox/**` had only the happy-dom `test.ts` and were missing the browser duplicate, so their behavior was never exercised in a real browser:

- `composite-renderer-active-ancestor`
- `composite-typeahead-buffer`
- `dialog-reopen-focus-state`
- `form-push-remove-6219`
- `menu-radio-default-checked`
- `select-renderer-orientation`

## Solution

Added a `test-browser.ts` for each, driving the same behavior as the companion happy-dom `test.ts` through real user interaction with `withFramework` and the accessible `q`/`query` helpers. All pass on Chromium, Firefox, and WebKit.

| Sandbox | Behavior covered |
| --- | --- |
| `composite-renderer-active-ancestor` | `CompositeRenderer`: the active descendant's ancestor item is rendered, so all items appear in order |
| `composite-typeahead-buffer` | [`CompositeTypeahead`](https://ariakit.com/reference/composite-typeahead): the typeahead buffer is scoped per composite instance, not leaked globally |
| `dialog-reopen-focus-state` | [`Dialog`](https://ariakit.com/reference/dialog): reopening resets outside-focus tracking, so focusing a newly added outside element hides it |
| `form-push-remove-6219` | [`FormPush`](https://ariakit.com/reference/form-push)/[`FormRemove`](https://ariakit.com/reference/form-remove): focus stays in the right array for prefix-sharing and regex-metacharacter names |
| `menu-radio-default-checked` | [`MenuItemRadio`](https://ariakit.com/reference/menu-item-radio): `defaultChecked` sets the initial checked state without leaking as a DOM prop |
| `select-renderer-orientation` | `SelectRenderer`: horizontal orientation lays items out by `left` (Cherry at `192px`, shared `top: 0`) |

Assertions are written from the user's perspective wherever practical — e.g. `form-push-remove-6219` asserts the visible `<output>` mirror via `q.status()` instead of poking `document.activeElement`, and `composite-renderer-active-ancestor` asserts the rendered button labels instead of a `data-index` attribute.

### Dialog example reworked for real-user reproducibility

`dialog-reopen-focus-state` originally created and focused an outside element directly through the DOM, so a real user visiting the preview couldn't reproduce the bug. The example is now fully user-driven: a non-modal [`Dialog`](https://ariakit.com/reference/dialog) with a "Reveal outside field" control that renders a brand-new outside field and focuses it. The controls are exempted from [`hideOnInteractOutside`](https://ariakit.com/reference/dialog#hideoninteractoutside) so interacting with them doesn't close the dialog; only focus landing on the freshly revealed (unmarked) field does — and only after a close/reopen has reset the dialog's outside-focus tracking. Both the happy-dom and browser tests drive `Open → focus inside → Close → Open → Reveal`, with no `page.evaluate`/DOM injection. The inside-focus step uses a text input (a native button doesn't focus on click in Safari/Firefox), and a fix-reverted red check confirms the test fails on Chromium, Firefox, and WebKit, then passes on all of them with the fix.

## Intentionally left happy-dom-only

`hovercard-nested-listener` keeps its happy-dom `test.ts` as the authoritative coverage and gets no browser duplicate: its test asserts via `vi.spyOn(document, …)` that toggling a nested hovercard doesn't churn the parent's `mousemove` listener — an implementation invariant with no user-observable browser symptom.

(The `command-6217` Space-release behavior is being handled in a separate PR, so its browser test is not part of this one.)


app/src/sandbox/command-6217/test-browser.ts
```diff
@@ -0,0 +1,33 @@
+import { withFramework } from "#app/test-utils/preview.ts";
```
diegohaz: You can remove this test. We're handling this in another PR.

app/src/sandbox/dialog-reopen-focus-state/test-browser.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    // The regression only surfaces with a focusable element that did not exist
+    // when the dialog opened: pre-existing outside controls are marked as
+    // "outside" on open, so focusing one wouldn't exercise the reopen reset.
+    // Create a brand-new outside input via the DOM (as the happy-dom test does)
+    // and focus it — the page.evaluate is only setup; the user-facing assertion
+    // is the dialog hiding below. Reopening must reset the dialog's outside-focus
+    // tracking so this dynamic element counts as interacting outside; before the
+    // fix the stale tracking kept it open.
+    await page.evaluate(() => {
```
diegohaz: This means the example is not easily reproducible in the browser by a real user visiting /react/previews/dialog-reopen-focus-state
You should improve the example and tests (including happydom tests) so a real user could verify the bug (before the fix) and the fix in the browser by interacting with the exaple. Then that's the same way we should test the example in our tests.

app/src/sandbox/composite-typeahead-buffer/test-browser.ts
```diff
@@ -0,0 +1,27 @@
...
+    await test.expect(q.button("Apricot")).toBeFocused();
+
+    // Move to the second composite while the first composite's "ap" buffer is
+    // still alive (typed within the 500ms cleanup window). With a per-instance
+    // buffer the next "b" matches Banana; a leaked global buffer would turn it
+    // into "apb", match nothing, and leave focus on Cherry.
+    await q.button("Cherry").focus();
+    await test.expect(q.button("Cherry")).toBeFocused();
+
+    await page.keyboard.press("b");
```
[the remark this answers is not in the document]
diegohaz: Fixed in c011148d6 — the buffer-building keys, the cross-composite focus, and the decisive `b` now run back-to-back with no awaited assertion between them, so the first composite's 500ms cleanup can't expire and clear a leaked global buffer before the check. A leaked buffer still produces `apb` (no match → focus stays on Cherry); a per-instance buffer matches Banana.

app/src/sandbox/command-6217/test-browser.ts
diegohaz: Removed in 80f3e5c63 — deleted the browser test. The happy-dom `test.ts` stays as the coverage here.

app/src/sandbox/dialog-reopen-focus-state/test-browser.ts
diegohaz: Fixed in 51eccec8d — reworked the example so the whole flow is user-driven and reproducible at `/react/previews/dialog-reopen-focus-state`. It is now a non-modal dialog (no backdrop intercepting clicks, outside controls stay interactive) with a "Reveal outside field" control that renders a brand-new outside field and focuses it. The controls are exempted from `hideOnInteractOutside`, so interacting with them does not close the dialog — only focus landing on the freshly revealed (unmarked) field does. Both `test.ts` and `test-browser.ts` now drive the same flow (Open → focus inside → Close → Open → Reveal), with no `page.evaluate` or DOM injection.

The inside-focus step uses a text input rather than a button, since clicking a native button doesn't move focus on Safari/Firefox (which would let the test pass without exercising the bug); both tests assert the focus landed inside before continuing. Verified with a fix-reverted red check that the test fails on chrome, firefox, and safari, and in happy-dom — and passes on all of them with the fix.

diegohaz:
Addressed all three review threads.

**Fixed**
- c011148d6 ([reply](https://github.com/ariakit/ariakit/pull/6269#discussion_r3368217113)): tightened `composite-typeahead-buffer/test-browser.ts` so the buffer-building keys, the cross-composite focus, and the decisive `b` run back-to-back with no awaited assertion between them — the first composite's 500ms cleanup can no longer expire and clear a leaked global buffer before the check.
- 51eccec8d ([reply](https://github.com/ariakit/ariakit/pull/6269#discussion_r3368217365)): reworked `dialog-reopen-focus-state` so the bug is reproducible by a real user at `/react/previews/dialog-reopen-focus-state`. Non-modal dialog + a "Reveal outside field" control that renders and focuses a brand-new outside field; controls exempted from `hideOnInteractOutside` so only the revealed field closes the dialog. Both `test.ts` and `test-browser.ts` drive the full flow with no `page.evaluate`/DOM injection, and use a text input for the inside-focus step (native buttons don't focus on click in Safari/Firefox). A fix-reverted red check confirms the test fails on chrome/firefox/safari + happy-dom, and passes on all of them with the fix.

**Removed**
- 80f3e5c63 ([reply](https://github.com/ariakit/ariakit/pull/6269#discussion_r3368217344)): deleted `command-6217/test-browser.ts` (handled in a separate PR); the happy-dom `test.ts` stays.

All `test-browser.ts` pass on chrome/firefox/safari; `pnpm tsc`, oxlint, and oxfmt are clean.

PR #6397, opened 2026-06-21
Fix test keypress charCode for Enter

### LINKED ISSUE #6357: @ariakit/test: type()/press.Enter() fire the Enter keypress with charCode 69 instead of 13, breaking legacy charCode-based handlers
### Current behavior

For every character it inserts, `type()` fires a `keypress` whose legacy `charCode` is computed from the *mapped key name* rather than from the raw character ([type.ts#L111-L118](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/type.ts#L111-L118)):

```ts
defaultAllowed = await dispatch.keyPress(input, {
  key,
  charCode: key.charCodeAt(0),
  ...options,
});
```

`key` comes from `getKeyFromChar`, which maps `"\n"` to `"Enter"` and `"\t"` to `"Tab"` ([type.ts#L8-L14](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/type.ts#L8-L14)). For ordinary characters `key === char`, so `key.charCodeAt(0)` is correct. But `"\n"` and `"\t"` stay on the `insertText` path (only `"\x7f"`/`"\b"` are diverted to delete input types), so `key.charCodeAt(0)` takes the first letter of the key *name*: `"Enter".charCodeAt(0) === 69` (the letter "E") and `"Tab".charCodeAt(0) === 84` (the letter "T"). Real browsers fire the Enter keypress with `charCode` 13 (CR), and fire no keypress at all for Tab.

This isn't limited to direct `type("\n")` calls: `press.Enter()` on a `<textarea>` delegates to this exact path ([press.ts#L309-L311](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/press.ts#L309-L311)).

The wrong value reaches the event because `dispatch` passes the full options object into `@testing-library/dom`'s `createEvent` (the `KeyboardEvent` constructor init) ([dispatch.ts#L403-L411](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/dispatch.ts#L403-L411)), and the post-construction `initKeyboardEvent` only force-assigns `key`/`code`/`location`/`repeat`/`isComposing` — it never normalizes `charCode` ([dispatch.ts#L145-L156](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/dispatch.ts#L145-L156)). So whatever the environment's `KeyboardEvent` constructor honors from the init is what handlers observe.

The impact is on external `@ariakit/test` consumers, not Ariakit's own suite. Under jsdom (which honors `charCode` in `KeyboardEventInit`) and in real browsers, a keypress listener that branches on the legacy `event.charCode` — still a common pattern for detecting Enter, e.g. `if (event.charCode === 13) submit()` — never matches when a test drives the textarea through `press.Enter()` or `type("\n")`, because the event carries `charCode` 69, indistinguishable from typing the letter "E". React's `onKeyPress` is affected identically: React reads `charCode` from the native event (`getEventCharCode`), so a component under test sees `charCode: 69` for Enter. For Tab the behavior is doubly wrong — no browser fires a keypress for Tab, but the utility emits one with `charCode` 84, which React's `getEventCharCode` treats as the printable letter "T".

This never surfaced in-repo because Ariakit's tests run under happy-dom, whose `KeyboardEvent` has no `charCode` member at all — its constructor reads only `keyCode`/`which` from the init — so neither the bug nor a fix is observable there. The only occurrence of `charCode` in the codebase is the buggy line itself.

The same user scenario was verified in real Chromium with Playwright: pressing Enter in a textarea delivers `charCode === 13`, so the legacy charCode-based submit fires — while the simulation reports `charCode === 69` and the submit never runs.

For a testing library, running a test is the consumer reproduction. The smallest reproduction is a textbox with an `onKeyPress` legacy-submit handler and a single `press.Enter`. Reproduce it on StackBlitz at https://stackblitz.com/github/ariakit/ariakit/tree/main/templates/react?file=src%2Fapp.tsx by replacing `src/app.tsx` with the code below (the test wiring is shown after it):

```tsx
import type { KeyboardEvent as ReactKeyboardEvent } from "react";

  const [text, setText] = useState("");
  const [lastCharCode, setLastCharCode] = useState<number | null>(null);
  const [submitted, setSubmitted] = useState(0);

  function handleKeyPress(event: ReactKeyboardEvent<HTMLTextAreaElement>) {
    setLastCharCode(event.charCode);
    // The classic legacy submit-on-Enter pattern: branch on charCode 13 (CR).
    if (event.charCode === 13) {
      event.preventDefault();
      setSubmitted((count) => count + 1);
    }
  }

  return (
    <div>
      <label htmlFor="comment">Comment</label>
      <textarea
        id="comment"
        value={text}
        onChange={(event) => setText(event.target.value)}
        onKeyPress={handleKeyPress}
      />
      <p>
        <output aria-label="Last char code">
          {lastCharCode === null ? "none" : lastCharCode}
        </output>{" "}
        <output aria-label="Submitted">{submitted}</output>
      </p>
    </div>
  );
}
```

Then drive it with `@ariakit/test`. Because happy-dom drops `charCode` entirely, the test must run under jsdom (the repo supports `// @vitest-environment jsdom` per file):

```ts
// @vitest-environment jsdom
import { press, q } from "@ariakit/test";

function getOutputText(name: string) {
  return q.status.ensure(name).textContent;
}

test("Enter in a textarea fires a keypress with charCode 13", async () => {
  await press.Enter(q.textbox.ensure("Comment"));

  // Browsers fire the Enter keypress with charCode 13 (CR), so the legacy
  // charCode-based submit handler fires. @ariakit/test computes the charCode
  // from the mapped key name "Enter" — "Enter".charCodeAt(0) === 69 — so the
  // event carries 69 (the letter "E") and the submit never runs.
  expect(getOutputText("Last char code")).toBe("13");
  expect(getOutputText("Submitted")).toBe("1");
});
```

Run `pnpm test-react test-type-keypress-charcode` from the repo root. The test fails:

```
AssertionError: expected '69' to be '13' // Object.is equality
```

The exact same user scenario was verified in a real browser with Playwright (`test-browser.ts`, a real Enter press on the rendered textarea): it passes in Chromium, which reports `charCode === 13` and fires the submit.

### Expected behavior

The Enter keypress should carry `charCode` 13 (CR) — what real browsers and jsdom report when Enter is pressed — so legacy `event.charCode === 13` handlers (and React's `onKeyPress`) behave the same under `@ariakit/test` as in a browser. More generally, the keypress `charCode` should be derived from the raw character being typed, not from the mapped key name (which yields 69 for Enter and 84 for Tab — values no conforming environment produces).

### Workaround

Consumer-side, in test code only. `press`/`type` forward their `KeyboardEventInit` options into the keydown/keypress/keyup events, and the `...options` spread comes after the buggy `charCode` in the keypress init, so an explicitly passed `charCode` wins:

```ts
await press.Enter(q.textbox.ensure("Comment"), { charCode: 13 });
// or, for a direct type call into a textarea:
await type("\n", q.textbox.ensure("Comment"), { charCode: 13 });
```

Verified: with this in place the assertion above passes on the current `@ariakit/test`, and the code-based submit fires. (`charCode?: number` is a valid `KeyboardEventInit` member in `lib.dom.d.ts`, so this is type-safe.)

### Possible solutions

Derive the `charCode` from the raw character being typed rather than the mapped key name, with an explicit Enter special case (browsers report 13/CR for the Enter keypress, while `"\n".charCodeAt(0)` is 10/LF). For every other character the expression is identical to today's behavior since `key === char`; for `"\t"` it yields 9 instead of 84. The caller override is preserved because the `...options` spread still follows `charCode`.

```diff
--- a/packages/ariakit-test/src/type.ts
+++ b/packages/ariakit-test/src/type.ts
@@ -112,7 +112,7 @@
           if (inputType === "insertText") {
             defaultAllowed = await dispatch.keyPress(input, {
               key,
-              charCode: key.charCodeAt(0),
+              charCode: char === "\n" ? 13 : char.charCodeAt(0),
               ...options,
             });
           }
```

`charCode: 9` for Tab is an approximation — no browser fires a keypress for Tab — but it is internally consistent with the utility's current behavior of inserting the tab as text, and is strictly closer to browsers than 84: React's `getEventCharCode` drops keypress events with `charCode < 32` (except 13), so `onKeyPress` handlers stop receiving a phantom "T" keypress for Tab. (Routing `"\t"` through the Tab keydown default instead of inserting it is a separate, larger behavior change and should not be bundled here.)

This was derived from reading `packages/ariakit-test/src/type.ts`; the package source was not modified and its suite was not re-run with this diff. In-repo regression risk is essentially nil — the suite runs under happy-dom, where the `charCode` init member is ignored, so behavior is bit-for-bit identical before and after; a regression test would need a `// @vitest-environment jsdom` override. Since `@ariakit/test` is a published package and this changes observable event values, a patch changeset is warranted.

diegohaz:
## Motivation

Fixes #6357.

`@ariakit/test` derived text-field `keypress` `charCode` values from the mapped key name instead of the raw typed character. For `press.Enter()` on a `<textarea>`, that made the simulated `keypress` emit `charCode` `69` from `"Enter"` instead of the browser-compatible Enter value `13`, so legacy `event.charCode === 13` handlers did not run in tests.

The same issue affected direct `type("\n")` calls because `press.Enter()` delegates to that path for textareas.

## Solution

Compute the simulated `keypress` `charCode` from the raw typed character rather than the mapped key name. Enter is normalized to `13` to match browser `keypress` behavior, while the existing caller override behavior remains intact because explicit options are still spread after the default `charCode`.

The regression is covered directly in `packages/ariakit-test/src/press.test.ts`. `dispatch` now preserves an explicit `KeyboardEventInit.charCode`, which lets the package test assert the behavior in the default test environment instead of requiring a jsdom-only sandbox test.

A patch changeset is included for `@ariakit/test`.

## Workaround

Until this fix is released, pass an explicit `charCode: 13` when simulating Enter in a textarea with `press.Enter()` or `type("\n")`:

```ts
// TODO: Remove this once https://github.com/ariakit/ariakit/issues/6357 is fixed.
await press.Enter(q.textbox.ensure("Comment"), { charCode: 13 });
```

The same override can be used for direct typing:

```ts
// TODO: Remove this once https://github.com/ariakit/ariakit/issues/6357 is fixed.
await type("\n", q.textbox.ensure("Comment"), { charCode: 13 });
```


app/src/sandbox/test-type-keypress-charcode-6357/index.react.tsx
```diff
@@ -0,0 +1,56 @@
+import { press } from "@ariakit/test";
```
diegohaz: Instead of sandbox, let's just have a test case in press.test.ts inside the ariakit-test package.

app/src/sandbox/test-type-keypress-charcode-6357/test.ts
```diff
@@ -0,0 +1,10 @@
+// @vitest-environment jsdom
```
diegohaz: Do we need jsdom here?

app/src/sandbox/test-type-keypress-charcode-6357/index.react.tsx
diegohaz: Fixed in abd6e5ddf - removed the sandbox repro and moved the regression coverage into `packages/ariakit-test/src/press.test.ts`.

app/src/sandbox/test-type-keypress-charcode-6357/test.ts
diegohaz: Fixed in abd6e5ddf - the regression now runs in the default package test environment, with `dispatch` preserving an explicit `KeyboardEventInit.charCode` so no jsdom override is needed.

diegohaz:
**Fixed**

- abd6e5ddf ([reply](https://github.com/ariakit/ariakit/pull/6397#discussion_r3449165325)): Removed the sandbox repro and moved the regression coverage into `packages/ariakit-test/src/press.test.ts`.
- abd6e5ddf ([reply](https://github.com/ariakit/ariakit/pull/6397#discussion_r3449165355)): Removed the jsdom-only test path by preserving explicit `KeyboardEventInit.charCode` in `dispatch`, so the regression runs in the default package test environment.

PR #6400, opened 2026-06-22
Fix native option clicks in test utils

### LINKED ISSUE #6351: @ariakit/test: clicking native option elements stamps the selected content attribute, and anchorless shift-click wipes the existing selection
### Current behavior

`click()` from `@ariakit/test` has a dedicated path for native `<option>` elements. Its core helper mutates both the `selected` IDL property and the `selected` *content* attribute ([click.ts#L56-L59](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/click.ts#L56-L59)):

```ts
function setSelected(element: HTMLOptionElement, selected: boolean) {
  element.setAttribute("selected", selected ? "selected" : "");
  element.selected = selected;
}
```

`selected` is a boolean content attribute, so `setAttribute("selected", "")` doesn't deselect anything at the attribute level — it *adds* the attribute. Per the HTML spec, that attribute is the option's *default* selectedness (`defaultSelected` reflects it), and real browsers never touch it on user interaction. Since every plain click and every shift-click on a `<select multiple>` first runs `resetOptions()`, which calls `setSelected(option, false)` on **every** option ([click.ts#L75-L89](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/click.ts#L75-L89)), one simulated click stamps the attribute onto the whole list. The single-select branch ([click.ts#L117-L119](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/click.ts#L117-L119)) stamps the clicked option too.

Observable consequences in consumer tests (repro below; selection sequence: click Apple → ctrl-click Cherry → ctrl-click Apple on a multi-select whose markup default is Banana):

- All three options match `option[selected]` (real browsers: only Banana). Snapshots/serialized HTML show `selected=""` on every option.
- In jsdom, every clicked-through option reports `defaultSelected === true` (real browsers: only Banana).
- In happy-dom, whose `form.reset()` consults `hasAttribute("selected")`, resetting the form re-selects Apple — an option the user explicitly deselected — and leaves the markup default Banana deselected: the post-reset selection is `["Apple", "Cherry"]` instead of `["Banana"]`.

There is a second defect in the shift-click branch ([click.ts#L91-L102](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/click.ts#L91-L102)): when there's no prior anchor (`lastOptionSelectedNotByShiftKey` unset), `-1` flows into `selectRange`, producing `options.slice(-1, elementIndex + 1)` — a negative start means "from the last option", so the range is empty or wrong. With a selection that exists before the first shift-click (e.g. a default-selected option), shift-clicking another option wipes the selected option instead of selecting the range from it: shift-clicking Cherry with Banana selected yields `["Cherry"]`, while Chromium yields `["Banana", "Cherry"]` (it anchors the range at the first selected option — verified with Playwright, see below). This path is publicly advertised in the `click()` JSDoc example ([click.ts#L135-L140](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-test/src/click.ts#L135-L140)).

For a testing library, running a test is the consumer reproduction. In the Ariakit repo:

1. Create `app/src/sandbox/test-click-option-selected/index.react.tsx`:

```tsx

  const selectRef = useRef<HTMLSelectElement>(null);
  const [selection, setSelection] = useState(["Banana"]);

  function readSelection() {
    const select = selectRef.current;
    if (!select) return;
    setSelection(Array.from(select.selectedOptions, (option) => option.value));
  }

  return (
    <form onReset={() => queueMicrotask(readSelection)}>
      <label htmlFor="fruits">Fruits</label>
      <select
        ref={selectRef}
        id="fruits"
        name="fruits"
        multiple
        size={3}
        defaultValue={["Banana"]}
        onChange={readSelection}
      >
        <option>Apple</option>
        <option>Banana</option>
        <option>Cherry</option>
      </select>
      <button type="reset">Reset</button>
      <output>Selected: {selection.join(", ") || "none"}</output>
    </form>
  );
}
```

2. Create `app/src/sandbox/test-click-option-selected/test.ts`:

```ts
import { click, q } from "@ariakit/test";

function getSelectedValues() {
  const select = q.listbox.ensure("Fruits") as HTMLSelectElement;
  return Array.from(select.options)
    .filter((option) => option.selected)
    .map((option) => option.value);
}

test("clicking options never touches the selected content attribute", async () => {
  await click(q.option.ensure("Apple"));
  await click(q.option.ensure("Cherry"), { ctrlKey: true });
  await click(q.option.ensure("Apple"), { ctrlKey: true });

  // Sanity check: the selection itself is simulated correctly.
  expect(getSelectedValues()).toEqual(["Cherry"]);

  // Real browsers never set the selected content attribute on user clicks.
  expect(q.option.ensure("Apple").hasAttribute("selected")).toBe(false);
  expect(q.option.ensure("Cherry").hasAttribute("selected")).toBe(false);
});

test("form.reset() restores the markup default selection", async () => {
  await click(q.option.ensure("Apple"));
  await click(q.option.ensure("Cherry"), { ctrlKey: true });
  await click(q.option.ensure("Apple"), { ctrlKey: true });
  await click(q.button.ensure("Reset"));

  expect(getSelectedValues()).toEqual(["Banana"]);
});

test("anchorless shift-click anchors at the first selected option", async () => {
  await click(q.option.ensure("Cherry"), { shiftKey: true });

  expect(getSelectedValues()).toEqual(["Banana", "Cherry"]);
});
```

3. Run `pnpm test-react test-click-option-selected` from the repo root. All three tests fail:

```
AssertionError: expected true to be false        // Apple keeps the selected attribute
AssertionError: expected [ 'Apple', 'Cherry' ] to deeply equal [ 'Banana' ]
AssertionError: expected [ 'Cherry' ] to deeply equal [ 'Banana', 'Cherry' ]
```

The exact same user scenario was verified in a real browser with Playwright (`app/src/sandbox/test-click-option-selected/test-browser.ts`, real mouse clicks with `Control`/`Shift` modifiers on the rendered listbox): all three assertions pass in Chromium, establishing the ground truth the simulation should match.

### Expected behavior

`click()` on a native `<option>` should mirror real browsers:

- Only the option's selectedness (IDL `selected`) changes. The `selected` content attribute — and therefore `defaultSelected`, `option[selected]` selectors, serialized HTML, and `form.reset()` behavior — stays untouched.
- Shift-click with no prior non-shift click should not slice from `-1`. Chromium (verified) anchors the range at the first selected option, falling back to the clicked option itself, so shift-clicking Cherry with Banana selected should yield `["Banana", "Cherry"]`, not wipe Banana.

### Workaround

Consumer-side, in test code only. Strip the stray attributes after each click, and establish the shift anchor explicitly with a ctrl-click pair (a net no-op for selection state, in real browsers too):

```ts
async function clickOption(name: string, eventOptions?: MouseEventInit) {
  const option = q.option.ensure(name);
  const select = option.closest("select");
  const defaults = select
    ? Array.from(select.options).filter((o) => o.hasAttribute("selected"))
    : [];
  await click(option, eventOptions);
  if (!select) return;
  for (const o of Array.from(select.options)) {
    if (!defaults.includes(o)) {
      o.removeAttribute("selected");
    }
  }
}

// Anchorless shift-click: set the anchor first with a ctrl-click pair.
await clickOption("Banana", { ctrlKey: true });
await clickOption("Banana", { ctrlKey: true });
await clickOption("Cherry", { shiftKey: true });
```

Verified: with this in place the attribute and shift-click tests above pass on the current `@ariakit/test`. The `form.reset()` divergence cannot be fully worked around under happy-dom, because happy-dom's own multi-select reset is non-conformant (it doesn't implement `defaultSelected` and falls back to selecting the first option) — but with the attribute kept clean the corruption no longer originates from `@ariakit/test`.

### Possible solutions

Fix `setSelected` to mutate only the selectedness state — the IDL assignment already exists, so the attribute line can simply be dropped (selection semantics in happy-dom and jsdom are unchanged; verified):

```diff
 function setSelected(element: HTMLOptionElement, selected: boolean) {
-  element.setAttribute("selected", selected ? "selected" : "");
+  // Browsers update only the option's selectedness on user interaction. The
+  // selected content attribute (defaultSelected) must stay untouched so that
+  // option[selected] selectors and form.reset() keep behaving like they
+  // would after a real click.
   element.selected = selected;
 }
```

For the shift-click anchor, replicate the Chromium fallback chain (last non-shift click → first selected option → the clicked option itself), which also removes the negative-index slice:

```diff
     if (eventOptions?.shiftKey) {
       const elementIndex = options.indexOf(element);
-      const referenceOption = select.lastOptionSelectedNotByShiftKey;
-      const referenceOptionIndex = referenceOption
-        ? options.indexOf(referenceOption)
-        : -1;
+      const referenceOption =
+        select.lastOptionSelectedNotByShiftKey ?? select.selectedOptions[0];
+      let referenceOptionIndex = referenceOption
+        ? options.indexOf(referenceOption)
+        : -1;
+      if (referenceOptionIndex === -1) {
+        referenceOptionIndex = elementIndex;
+      }

       resetOptions();
       // Select options between the reference option and the clicked element
       selectRange(elementIndex, referenceOptionIndex);

       setSelected(element, true);
```

(`select.selectedOptions[0]` must be read before `resetOptions()` clears the selection; the statement order above guarantees that.) Both diffs were applied locally and validated against the repro: the attribute and shift-anchor tests pass with them. Note for the fix PR: the `form.reset()` repro test stays red under the repo's happy-dom runner even with the fix, solely because happy-dom's multi-select reset is itself non-conformant — it passes in real browsers, and the `defaultSelected` pollution it guards against is directly observable in jsdom. Since `@ariakit/test` is a published package, a patch changeset is warranted.

diegohaz:
## Motivation

Fixes #6351.

`click()` from `@ariakit/test` should mirror native option interactions without changing the option markup/default selectedness. The previous option click path wrote the `selected` content attribute while simulating user selection, which polluted `defaultSelected`, selectors such as `option[selected]`, and reset behavior in test environments.

Anchorless shift-click on a multiple select should also preserve the browser-like range anchor when an option is already selected instead of clearing that existing selection. In the default happy-dom test environment, change handlers should observe the updated `selectedOptions` state even if `selectedOptions` was read earlier in the interaction.

## Solution

Update the native option click path so `setSelected` only assigns the option selected IDL state. For multiple-select shift-click, use the last non-shift clicked option when available, then fall back to the first selected option from the already-collected option array, and finally to the clicked option itself.

After simulated option selectedness changes, clear happy-dom's select cache before dispatching `input` and `change`, without mutating the `selected` content attribute/default selectedness.

Add default happy-dom regression coverage in `click.test.ts` for `selectedOptions` state seen by `change` handlers, and move jsdom-specific default selectedness assertions to `click.jsdom.test.ts`. Add a patch changeset for `@ariakit/test`.

## Verification

- `pnpm test packages/ariakit-test/src/click.test.ts` failed before the cache invalidation fix with stale `selectedOptions`
- `pnpm test packages/ariakit-test/src/click.test.ts packages/ariakit-test/src/click.jsdom.test.ts`
- `pnpm lint-fix`
- `pnpm test packages/ariakit-test/src`
- `pnpm tsc`
- `pnpm -F @ariakit/test run build`


packages/ariakit-test/src/click.ts
```diff
@@ -90,10 +90,16 @@ async function clickOption(
 
     if (eventOptions?.shiftKey) {
       const elementIndex = options.indexOf(element);
-      const referenceOption = select.lastOptionSelectedNotByShiftKey;
-      const referenceOptionIndex = referenceOption
-        ? options.indexOf(referenceOption)
-        : -1;
+      const referenceOption =
+        select.lastOptionSelectedNotByShiftKey ??
+        select.selectedOptions.item(0);
```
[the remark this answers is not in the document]
diegohaz: Fixed in f79b1c594 - the shift-click fallback now derives the anchor from the existing options array instead of reading `select.selectedOptions` before selection updates, and `click-happy-dom.test.ts` covers the `change` event `selectedOptions` state.

diegohaz:
**Fixed**

- f79b1c594 ([reply](https://github.com/ariakit/ariakit/pull/6400#discussion_r3452201602)): Avoided reading `select.selectedOptions` before shift-click selection mutations and added happy-dom regression coverage for `change` handlers reading `selectedOptions`.

packages/ariakit-test/src/click-happy-dom.test.ts
```diff
@@ -0,0 +1,32 @@
+import { afterEach, expect, test } from "vitest";
```
diegohaz: I believe the default file should be happy-dom, then we should have a `click.jsdom.test.ts`

packages/ariakit-test/src/click.ts
```diff
@@ -54,7 +54,7 @@ async function clickLabel(element: HTMLLabelElement, options?: MouseEventInit) {
 }
 
 function setSelected(element: HTMLOptionElement, selected: boolean) {
-  element.setAttribute("selected", selected ? "selected" : "");
+  // User interaction changes selectedness, not default selectedness.
   element.selected = selected;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 5282c242c - after simulated option selectedness changes, the select cache is cleared in happy-dom before dispatching `input` and `change`, so handlers see the updated `selectedOptions` collection without mutating the `selected` content attribute/default selectedness.

packages/ariakit-test/src/click-happy-dom.test.ts
diegohaz: Fixed in 5282c242c - the default `click.test.ts` file now runs under happy-dom, and the jsdom-specific default selectedness assertions moved to `click.jsdom.test.ts`.

diegohaz:
**Fixed**

- 5282c242c ([reply](https://github.com/ariakit/ariakit/pull/6400#discussion_r3452342017)): Cleared happy-dom's select cache after simulated option selectedness changes so `selectedOptions` is current in `input`/`change` handlers without mutating default selectedness.
- 5282c242c ([reply](https://github.com/ariakit/ariakit/pull/6400#discussion_r3452342038)): Moved the default happy-dom coverage to `click.test.ts` and the jsdom-specific assertions to `click.jsdom.test.ts`.

packages/ariakit-test/src/click.ts
```diff
@@ -54,10 +54,28 @@ async function clickLabel(element: HTMLLabelElement, options?: MouseEventInit) {
...
+    if (clearCacheSymbol) {
+      const clearCache: unknown = Reflect.get(element, clearCacheSymbol);
+      if (typeof clearCache === "function") {
+        clearCache.call(element);
+      }
+      return;
+    }
+    prototype = Object.getPrototypeOf(prototype);
+  }
+}
```
diegohaz, line 83: This helper reaches deep into happy-dom internals — it walks the prototype chain to discover a private `Symbol('clearCache')` by its `description` and invokes it — but there's no comment explaining *why* it exists or what guarantees it relies on. That's the kind of non-obvious environment workaround the codebase normally documents in detail (see every patch in `shims.ts`, each with a WHY block and often an upstream issue link).

For a future reader (or a happy-dom upgrade that renames/removes the symbol — in which case this silently no-ops and the staleness regresses), the missing rationale is the real cost. I confirmed the underlying bug locally: happy-dom caches `selectedOptions` on the `querySelectorAll('option')` result array (`HTMLSelectElement.js`), and `updateSelectedness` — invoked by the `option.selected` setter — mutates `selectedness` without invalidating that cache. So after the new attribute-free `setSelected`, a handler that read `selectedOptions` before the click keeps seeing the stale collection on `change`. The old `setAttribute("selected")` write happened to clear that cache as a side effect; this restores it explicitly.

Suggested: add a WHY comment matching the `shims.ts` convention, e.g.

```ts
// happy-dom caches `select.selectedOptions` on the cached `querySelectorAll`
// result and does not invalidate it when only `option.selected` changes (its
// `updateSelectedness` skips the cache). Real browsers and jsdom expose a live
// collection. The previous `selected` attribute write cleared this cache as a
// side effect; now that `setSelected` only touches the IDL state, clear the
// query cache explicitly so handlers reading `selectedOptions` on the
// dispatched `input`/`change` events observe the updated selection.
function clearHappyDOMSelectCache(element: HTMLSelectElement) {
```

This is also a genuine happy-dom spec gap, like the others tracked from `shims.ts`. Consider filing an upstream issue and linking it here so the workaround can be removed when fixed; a `shims.ts`-style patch of the `option.selected` setter would fix it environment-wide and be removable in one place, though the localized clear here is a fine lighter-weight choice.

packages/ariakit-test/src/click.ts
```diff
@@ -90,10 +108,16 @@ async function clickOption(
...
+      const referenceOption =
+        select.lastOptionSelectedNotByShiftKey ??
+        options.find((option) => option.selected);
+      let referenceOptionIndex = elementIndex;
+      if (referenceOption) {
+        referenceOptionIndex = options.indexOf(referenceOption);
+        if (referenceOptionIndex === -1) {
+          referenceOptionIndex = elementIndex;
+        }
+      }
```
diegohaz: Minor (optional, readability): this block initializes `referenceOptionIndex` to `elementIndex`, then conditionally overwrites it with the reference index, then conditionally resets it *back* to `elementIndex`. The "anchor at the clicked element when there's no usable reference" intent gets spread across three assignments and two nesting levels. A flat `const` with a single fallback at the point of use reads more directly:

```ts
const elementIndex = options.indexOf(element);
const referenceOption =
  select.lastOptionSelectedNotByShiftKey ??
  options.find((option) => option.selected);
const referenceOptionIndex = referenceOption
  ? options.indexOf(referenceOption)
  : -1;

resetOptions();
// Select options between the clicked element and the reference option,
// anchoring at the clicked element when there is no usable reference.
selectRange(
  elementIndex,
  referenceOptionIndex === -1 ? elementIndex : referenceOptionIndex,
);
```

This covers the same three cases (no reference → anchor at clicked element; stale `lastOptionSelectedNotByShiftKey` not in `options` → `indexOf` returns `-1` → anchor at clicked element; valid reference → its index). I applied this locally and the full `packages/ariakit-test/src` suite stays green (53/53), including the two new shift-click tests. Feel free to keep the current form if you prefer it.

diegohaz (commented):
## Review summary

Comment-only review. Thanks — the core approach is sound and well-tested.

**Scope / type:** Initial review (no prior `ariakit-pr-review` marker found; the existing threads are from the Codex bot). Baseline: merge-base `3bd8daf3` → head `5282c242c`. Files: `packages/ariakit-test/src/click.ts`, `click.test.ts`, `click.jsdom.test.ts`, `.changeset/300-test-option-click.md`.

**What I validated locally** (Node from `.nvmrc`, `pnpm install` in this worktree):
- `pnpm test packages/ariakit-test/src` → 9 files / 53 tests pass, including the two new files.
- Prototyped the shift-click simplification below → still 53/53.
- `pnpm -F @ariakit/test run build` → exit 0 (restored the `package.json` exports the build rewrites in place).
- Read happy-dom 20.10.6 internals: confirmed `setSelected` no longer needs the attribute write (`option.selected` IDL drives selection in both engines), and confirmed the `selectedOptions` cache-staleness bug the new `clearHappyDOMSelectCache` works around (`updateSelectedness` mutates `selectedness` without invalidating the `querySelectorAll`-keyed `selectedOptions` cache).

**Correctness:** No correctness issues found. The fix correctly (a) stops polluting `defaultSelected`/`option[selected]`/`form.reset()` by dropping the attribute write, (b) anchors anchorless shift-click at the first selected option instead of slicing from `-1`, and (c) derives that anchor from `options.find(o => o.selected)` rather than reading `selectedOptions` before the mutation, avoiding the happy-dom snapshot the Codex thread flagged. The earlier Codex P2s (cache pre-read, post-change cache invalidation) and the file-naming note are all addressed in the current head. I re-verified against the code, not just the thread replies.

**No regression to the rest of the repo:** the native-`<option>` path in `click()` is only hit by real `<select>` elements; Ariakit's own Select is a custom listbox, so the app/example browser tests don't exercise this path. No repo test asserts the `selected` content attribute after a simulated click.

**Findings (both non-blocking):**
1. `clearHappyDOMSelectCache` — undocumented deep happy-dom-internal workaround (prototype-walk to find a private `Symbol('clearCache')`). Add a WHY comment matching the `shims.ts` convention, and consider filing/linking an upstream happy-dom issue (this is a real spec gap; none of the currently-tracked happy-dom issues cover `selectedOptions` cache staleness).
2. Shift-click `referenceOptionIndex` — a behavior-preserving flatten (drop the `let` + nested double-fallback to `elementIndex`); verified green locally. Optional.

**Changeset:** correct — `@ariakit/test` patch only (no public reference page exists for the test util, so plain inline code is right).

**Not fully validated:** I did not run `pnpm tsc` repo-wide (the package build typechecks the changed file) or the app/Playwright browser suites (the changed path isn't exercised there). The `form.reset()` repro stays red only under happy-dom's non-conformant multi-select reset, as the PR description already notes — not a defect in this change.

packages/ariakit-test/src/click.ts
diegohaz: Fixed in 768dc4725 - added a WHY comment explaining the happy-dom `selectedOptions` cache behavior, the browser/jsdom contrast, and why the localized cache clear is needed after switching to IDL-only selectedness updates.

packages/ariakit-test/src/click.ts
diegohaz: Fixed in 768dc4725 - flattened the shift-click fallback by deriving the reference index once and applying the clicked-option fallback at the `selectRange` call site.

diegohaz:
**Fixed**

- 768dc4725 ([reply](https://github.com/ariakit/ariakit/pull/6400#discussion_r3452922574)): Added the WHY comment for the happy-dom `selectedOptions` cache workaround.
- 768dc4725 ([reply](https://github.com/ariakit/ariakit/pull/6400#discussion_r3452922887)): Flattened the anchorless shift-click reference-index fallback while preserving the clicked-option fallback.

diegohaz (commented):
## Re-review summary

**Type:** Re-review. Prior marked review was against `5282c242c`; incremental baseline `5282c242c` → current head `768dc4725` ("Clarify option click fallback"). Only `packages/ariakit-test/src/click.ts` changed.

This commit addresses both findings from my last review, and nothing else:

1. **`clearHappyDOMSelectCache` documentation** — a WHY comment was added explaining the happy-dom `selectedOptions` cache behavior, the browser/jsdom contrast, and why the explicit clear is needed now that `setSelected` no longer writes the attribute. The comment matches what I verified in happy-dom's source. ✅ (thread resolved)
2. **Shift-click `referenceOptionIndex` flatten** — applied exactly as proposed: derive the index once, apply the clicked-element fallback at the `selectRange` call. ✅ (thread resolved)

**Validation on the new head:** `pnpm test packages/ariakit-test/src` → 9 files / 53 tests pass. The simplification is the same one I'd already confirmed behavior-preserving (covers all three cases: no reference, stale `lastOptionSelectedNotByShiftKey` → `indexOf === -1`, and a valid reference). Worktree clean.

**New findings:** none. The two earlier Codex P2 threads and the file-naming note remain correctly resolved in this head. Nothing outside the incremental diff is newly affected. LGTM.

PR #6408, opened 2026-06-23
Fix Tailwind root push direction

### LINKED ISSUE #6343: @ariakit/tailwind: push utilities on a root ak-layer escape toward the dark boundary (own-lightness direction fallback is always IACVT)
### Current behavior

In `@ariakit/tailwind`, when `ak-layer-push-*` (or `ak-state-push-*`) is used on a layer that has **no ancestor `ak-layer`** — exactly the readme's canonical setup where `<body>` is the app's root layer ([readme.md#L41-L46](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/readme.md#L41-L46)) — a dark layer escapes the forbidden mid-lightness band toward the **dark** boundary instead of the light one. With `ak-layer ak-layer-gray-950 ak-layer-push-50` on `<body>`, the computed background is `oklch(0.543723 0.028 261.692)` (the dark band edge `--_ak-fla`), while the exact same classes on a layer nested under a dark parent layer correctly compute `oklch(0.728139 0.028 261.692)` (the light band edge `--_ak-flb`). Pushed surfaces and hover/active states on parentless dark layers drift the wrong way.

The root cause is a custom property that is always *invalid at computed-value time* (IACVT). The `ak-layer` body assigns the dynamic "derive push direction from the layer's own lightness" fallback ([input.ts#L1193-L1194](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L1193-L1194), compiled at [output.css#L129-L134](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L129-L134)):

```css
--_ak-lipdtl: var(--_ak-odtl);
--_ak-lpdtl: var(--_ak-odtl);
--_ak-odtl: clamp(0, var(--_ak-lod), 1);
```

`--_ak-lod` is a universal-syntax (`*`) property whose value is the uninterpreted token stream `calc((clamp(0, ((0.645 - l) * 1000000), 1) * 2) - 1)` containing the relative-color `l` keyword ([output.css#L1334-L1338](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L1334-L1338), authored at [input.ts#L506-L508](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L506-L508)). That keyword is only meaningful once substituted inside an `oklch(from …)` expression. But `--_ak-lipdtl`/`--_ak-lpdtl` are registered with `syntax: "<number>"` via `_ak.prop.zero` ([input.ts#L552-L562](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L552-L562), [output.css#L1448-L1452](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L1448-L1452)), so per the CSS Properties and Values API the substituted value fails the `<number>` grammar, the declaration is IACVT, and the property falls back to its registered `initial-value: 0` — always.

The only working direction sources are the `ak-light`/`ak-dark` variant overrides ([output.css#L147-L156](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L147-L156)), which are style container queries against the **parent** layer's scheme color, so they never match on a layer without an ancestor `ak-layer` (the inherited `--_ak-ls` is its initial `black`, which matches neither `oklch(0 0 0)` nor `oklch(1 0 0)`). With the direction stuck at `0`, `getPushL`'s escape boundary ([input.ts#L866-L869](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L866-L869)) is always `--_ak-fla`, the dark band edge, and the high-contrast push bias direction in `getLayerIdleContrastBiasDirection` ([input.ts#L1171-L1183](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L1171-L1183)) collapses to `-1` regardless of the layer's lightness. The fallback's intent comes from commit [`6f838eb65`](https://github.com/ariakit/ariakit/commit/6f838eb659ee6872e727d013444ab48a4d232995) (PR #5947, "Fix layer push lightness bounds"), which states pushes escape the band through the boundary matching the push direction; IACVT silently turns that into the constant `0`. Existing sandbox snapshots never caught it because every push case nests under a parent layer, where the variants supply the direction.

2. In `src/style.css`, replace the `body` rule with the following, making `<body>` a dark root layer with a push (its parent `<html>` is not a layer, per the readme setup):

   ```css
   body {
     @apply ak-layer ak-layer-gray-950 ak-layer-push-50 p-10;
   }
   ```

4. Look at the page: the body background (root layer, pushed) is a mid-**dark** gray — `oklch(0.543723 0.028 261.692)` — while the nested element with the **exact same classes** is a mid-**light** gray — `oklch(0.728139 0.028 261.692)`. Both are dark gray-950 layers pushed by 50, so both should escape the forbidden band toward the light boundary and match.

```tsx
  return (
    <div className="ak-layer ak-layer-gray-950 p-8">
      <div className="ak-layer ak-layer-gray-950 ak-layer-push-50 p-8">
        Nested pushed layer (control): escapes toward the light boundary as
        intended. The body background should match this color, but it is much
        darker.
      </div>
    </div>
  );
}
```

This was verified with a real-browser Playwright test that compares the computed `background-color` lightness of the root body layer against the identically-classed nested control.

### Expected behavior

A pushed dark layer should escape the forbidden mid-lightness band toward the **light** boundary, whether or not it has an ancestor `ak-layer`. The root body layer above should compute the same `oklch(0.728139 0.028 261.692)` background as the nested control, and the high-contrast push bias should follow the layer's own lightness.

### Workaround

Give the pushed root layer a parent layer with the same scheme, so the `ak-dark`/`ak-light` variants supply the push direction instead of the broken fallback:

```css
html {
  @apply ak-layer ak-layer-gray-950;
}
```

(Or `@apply ak-layer ak-layer-white dark:ak-layer-gray-950;` for adaptive setups.) Note the readme advises against making `html` the base layer, so an inner wrapper works too: keep a plain root layer on `<body>` and move the pushed layer to a wrapper `<div>` inside it. The `html` variant was verified to make the Playwright test pass.

### Possible solutions

Keep the declarations and the variant overrides, but register the two direction properties with universal syntax and a numeric initial instead of `syntax: "<number>"`, so the `l`-bearing token stream survives substitution (it only ever terminates inside `oklch(from …)` lightness calcs, where `l` is valid). This is the exact pattern already used by `layerIdleLimitedDelta`/`layerOffsetDelta`:

```diff
   layerIdlePushValue: _ak.prop.zero("lipv"),
   layerIdlePushEnabled: _ak.prop.zero("lipe"),
-  layerIdlePushDirectionToLight: _ak.prop.zero("lipdtl"),
+  // Must stay universal syntax: the var(--_ak-odtl) fallback carries the
+  // relative-color `l` keyword, which fails a registered <number> grammar
+  // and would be invalid at computed-value time (always falling back to 0).
+  layerIdlePushDirectionToLight: _ak.prop("lipdtl", { initial: 0 }),
   ...
   layerPushValue: _ak.prop.zero("lpv"),
-  layerPushDirectionToLight: _ak.prop.zero("lpdtl"),
+  layerPushDirectionToLight: _ak.prop("lpdtl", { initial: 0 }),
```

The literal `0`/`1` variant overrides still parse under `*`, and no style query or animation depends on the `<number>` typing. Rebuild `output.css`, and add a root-level push case to the `ariakit-tailwind` sandbox (every existing push case nests under a parent layer, so current snapshots cannot catch a regression here).

diegohaz:
## Motivation

`@ariakit/tailwind` resolved dark `ak-layer-push-*` and `ak-state-push-*` utilities from parent layer context. A root layer such as `ak-layer ak-layer-gray-950 ak-layer-push-50` escaped the forbidden lightness band toward the dark boundary, while the same dark layer nested under a dark parent escaped toward the light boundary.

A light parent could also force a dark current layer to escape in the wrong direction. The push direction should come from the current layer, not from whether an ancestor currently matches `ak-light` or `ak-dark`.

Fixes #6343.

## Solution

The push calculations now use the current layer's own offset direction at the `ak-layer-push-*` and `ak-state-push-*` call sites. This keeps the relative-color `l` channel in the expression where it is valid and avoids carrying a parent-derived push direction through a separate variable.

The high-contrast bias for `ak-layer-push-*` is folded into the pre-push lightness value, so it also follows the current layer without needing a CSS `if()` self-query. The previous self-source direction plumbing and generated `@supports (color: if(...))` block for push direction were removed.

Parentless edge direction remains unsupported: edge colors keep the existing variant/default direction behavior. The regression test now checks the pushed background result across parentless, dark-parent, and light-parent cases for both layer and state push in normal and high-contrast modes, while snapshots continue to record the edge output.

## Workaround

For versions without this fix, give the pushed root layer a parent `ak-layer` with the same scheme so the old parent-based direction matches the current layer.

```tsx
// TODO: Remove this wrapper after https://github.com/ariakit/ariakit/issues/6343 is fixed.
<div className="ak-layer ak-layer-gray-950">
  <div className="ak-layer ak-layer-gray-950 ak-layer-push-50" />
  <div className="ak-layer ak-layer-gray-950 ak-state-push-50" />
</div>
```

### DIFF
--- a/.changeset/300-tailwind-root-push-direction.md
+++ b/.changeset/300-tailwind-root-push-direction.md
@@ -2,4 +2,4 @@
 "@ariakit/tailwind": patch
 ---
 
-Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers. In browsers with CSS `if()` support, root pushed layers also use their own source direction for high-contrast bias and edge colors. Browsers without CSS `if()` support also avoid extra source-direction color work.
+Fixed `ak-layer-push-*` and `ak-state-push-*` in `@ariakit/tailwind` to escape the forbidden lightness band based on the current layer, regardless of the parent layer.
--- a/.changeset/300-tailwind-root-push-direction.md
+++ b/.changeset/300-tailwind-root-push-direction.md
@@ -2,4 +2,4 @@
 "@ariakit/tailwind": patch
 ---
 
-Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers. In browsers with CSS `if()` support, root pushed layers also use their own source direction for high-contrast bias and edge colors.
+Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers. In browsers with CSS `if()` support, root pushed layers also use their own source direction for high-contrast bias and edge colors. Browsers without CSS `if()` support also avoid extra source-direction color work.
--- a/.changeset/300-tailwind-root-push-direction.md
+++ b/.changeset/300-tailwind-root-push-direction.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/tailwind": patch
+---
+
+Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers.
--- a/.changeset/300-tailwind-root-push-direction.md
+++ b/.changeset/300-tailwind-root-push-direction.md
@@ -2,4 +2,4 @@
 "@ariakit/tailwind": patch
 ---
 
-Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers.
+Fixed root pushed layers in `@ariakit/tailwind` to escape the forbidden lightness band in the same direction as nested dark layers. In browsers with CSS `if()` support, root pushed layers also use their own source direction for high-contrast bias and edge colors.
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -567,28 +567,51 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
-      test("pushes parentless dark layers toward the light boundary", async ({
-        page,
-        q,
-      }) => {
-        await waitForPreviewReady(page);
-        const cases = [
-          ["parentless dark layer push", "nested dark layer push"],
-          ["parentless dark state push", "nested dark state push"],
-        ] as const;
+      for (const contrast of [false, true]) {
+        const label = contrast ? " in high contrast" : "";
+        test(`pushes parentless dark layers toward the light boundary${label}`, async ({
+          page,
+          q,
+        }) => {
+          if (contrast) {
+            const cdp = await page.context().newCDPSession(page);
+            await cdp.send("Emulation.setEmulatedMedia", {
+              features: [{ name: "prefers-contrast", value: "more" }],
+            });
+            await page.reload({ waitUntil: "networkidle" });
+          }
+          await waitForPreviewReady(page);
+          const cases = [
+            ["parentless dark layer push", "nested dark layer push"],
+            ["parentless dark state push", "nested dark state push"],
+          ] as const;
 
-        for (const [parentlessLabel, nestedLabel] of cases) {
-          const parentlessPush = q.region(parentlessLabel);
-          const nestedPush = q.region(nestedLabel);
-          const nestedBackgroundColor = await nestedPush.evaluate((element) => {
-            return getComputedStyle(element).backgroundColor;
-          });
-          await expect(parentlessPush).toHaveCSS(
-            "background-color",
-            nestedBackgroundColor,
-          );
-        }
-      });
+          for (const [parentlessLabel, nestedLabel] of cases) {
+            const parentlessPush = q.region(parentlessLabel);
+            const nestedPush = q.region(nestedLabel);
+            const nestedStyles = await nestedPush.evaluate((element) => {
+              const style = getComputedStyle(element);
+              return {
+                backgroundColor: style.backgroundColor,
+                borderTopColor: style.borderTopColor,
+                boxShadow: style.boxShadow,
+              };
+            });
+            await expect(parentlessPush).toHaveCSS(
+              "background-color",
+              nestedStyles.backgroundColor,
+            );
+            await expect(parentlessPush).toHaveCSS(
+              "border-top-color",
+              nestedStyles.borderTopColor,
+            );
+            await expect(parentlessPush).toHaveCSS(
+              "box-shadow",
+              nestedStyles.boxShadow,
+            );
+          }
+        });
+      }
 
       test("no color contrast violations (WCAG AA)", async ({ page }) => {
         test.setTimeout(60_000);
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -567,23 +567,23 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
-      test("pushes parentless dark layers toward the light boundary", async ({
+      test("pushes wrapped dark layers toward the light boundary", async ({
         page,
         q,
       }) => {
         await waitForPreviewReady(page);
         const cases = [
-          ["parentless dark layer push", "nested dark layer push"],
-          ["parentless dark state push", "nested dark state push"],
+          ["wrapped dark layer push", "nested dark layer push"],
+          ["wrapped dark state push", "nested dark state push"],
         ] as const;
 
-        for (const [parentlessLabel, nestedLabel] of cases) {
-          const parentlessPush = q.region(parentlessLabel);
+        for (const [wrappedLabel, nestedLabel] of cases) {
+          const wrappedPush = q.region(wrappedLabel);
           const nestedPush = q.region(nestedLabel);
           const nestedBackgroundColor = await nestedPush.evaluate((element) => {
             return getComputedStyle(element).backgroundColor;
           });
-          await expect(parentlessPush).toHaveCSS(
+          await expect(wrappedPush).toHaveCSS(
             "background-color",
             nestedBackgroundColor,
           );
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -567,59 +567,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
-      for (const contrast of [false, true]) {
-        const label = contrast ? " in high contrast" : "";
-        test(`pushes dark layers toward the light boundary regardless of parent${label}`, async ({
-          page,
-          q,
-        }) => {
-          if (contrast) {
-            const cdp = await page.context().newCDPSession(page);
-            await cdp.send("Emulation.setEmulatedMedia", {
-              features: [{ name: "prefers-contrast", value: "more" }],
-            });
-            await page.reload({ waitUntil: "networkidle" });
-          }
-          await waitForPreviewReady(page);
-          const cases = [
-            [
-              "parentless dark layer push",
-              "nested dark layer push",
-              "light parent dark layer push",
-            ],
-            [
-              "parentless dark state push",
-              "nested dark state push",
-              "light parent dark state push",
-            ],
-          ] as const;
-
-          for (const [
-            parentlessLabel,
-            darkParentLabel,
-            lightParentLabel,
-          ] of cases) {
-            const parentlessPush = q.region(parentlessLabel);
-            const darkParentPush = q.region(darkParentLabel);
-            const lightParentPush = q.region(lightParentLabel);
-            const referenceStyles = await darkParentPush.evaluate((element) => {
-              const style = getComputedStyle(element);
-              return {
-                backgroundColor: style.backgroundColor,
-              };
-            });
-            await expect(parentlessPush).toHaveCSS(
-              "background-color",
-              referenceStyles.backgroundColor,
-            );
-            await expect(lightParentPush).toHaveCSS(
-              "background-color",
-              referenceStyles.backgroundColor,
-            );
-          }
-        });
-      }
-
       test("no color contrast violations (WCAG AA)", async ({ page }) => {
         test.setTimeout(60_000);
         const results = await new AxeBuilder({ page })
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -567,23 +567,23 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
-      test("pushes wrapped dark layers toward the light boundary", async ({
+      test("pushes parentless dark layers toward the light boundary", async ({
         page,
         q,
       }) => {
         await waitForPreviewReady(page);
         const cases = [
-          ["wrapped dark layer push", "nested dark layer push"],
-          ["wrapped dark state push", "nested dark state push"],
+          ["parentless dark layer push", "nested dark layer push"],
+          ["parentless dark state push", "nested dark state push"],
         ] as const;
 
-        for (const [wrappedLabel, nestedLabel] of cases) {
-          const wrappedPush = q.region(wrappedLabel);
+        for (const [parentlessLabel, nestedLabel] of cases) {
+          const parentlessPush = q.region(parentlessLabel);
           const nestedPush = q.region(nestedLabel);
           const nestedBackgroundColor = await nestedPush.evaluate((element) => {
             return getComputedStyle(element).backgroundColor;
           });
-          await expect(wrappedPush).toHaveCSS(
+          await expect(parentlessPush).toHaveCSS(
             "background-color",
             nestedBackgroundColor,
           );
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -569,7 +569,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
 
       for (const contrast of [false, true]) {
         const label = contrast ? " in high contrast" : "";
-        test(`pushes parentless dark layers toward the light boundary${label}`, async ({
+        test(`pushes dark layers toward the light boundary regardless of parent${label}`, async ({
           page,
           q,
         }) => {
@@ -582,32 +582,39 @@ withFramework(import.meta.dirname, async ({ test }) => {
           }
           await waitForPreviewReady(page);
           const cases = [
-            ["parentless dark layer push", "nested dark layer push"],
-            ["parentless dark state push", "nested dark state push"],
+            [
+              "parentless dark layer push",
+              "nested dark layer push",
+              "light parent dark layer push",
+            ],
+            [
+              "parentless dark state push",
+              "nested dark state push",
+              "light parent dark state push",
+            ],
           ] as const;
 
-          for (const [parentlessLabel, nestedLabel] of cases) {
+          for (const [
+            parentlessLabel,
+            darkParentLabel,
+            lightParentLabel,
+          ] of cases) {
             const parentlessPush = q.region(parentlessLabel);
-            const nestedPush = q.region(nestedLabel);
-            const nestedStyles = await nestedPush.evaluate((element) => {
+            const darkParentPush = q.region(darkParentLabel);
+            const lightParentPush = q.region(lightParentLabel);
+            const referenceStyles = await darkParentPush.evaluate((element) => {
               const style = getComputedStyle(element);
               return {
                 backgroundColor: style.backgroundColor,
-                borderTopColor: style.borderTopColor,
-                boxShadow: style.boxShadow,
               };
             });
             await expect(parentlessPush).toHaveCSS(
               "background-color",
-              nestedStyles.backgroundColor,
+              referenceStyles.backgroundColor,
             );
-            await expect(parentlessPush).toHaveCSS(
-              "border-top-color",
-              nestedStyles.borderTopColor,
-            );
-            await expect(parentlessPush).toHaveCSS(
-              "box-shadow",
-              nestedStyles.boxShadow,
+            await expect(lightParentPush).toHaveCSS(
+              "background-color",
+              referenceStyles.backgroundColor,
             );
           }
         });
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -567,6 +567,29 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
+      test("pushes parentless dark layers toward the light boundary", async ({
+        page,
+        q,
+      }) => {
+        await waitForPreviewReady(page);
+        const cases = [
+          ["parentless dark layer push", "nested dark layer push"],
+          ["parentless dark state push", "nested dark state push"],
+        ] as const;
+
+        for (const [parentlessLabel, nestedLabel] of cases) {
+          const parentlessPush = q.region(parentlessLabel);
+          const nestedPush = q.region(nestedLabel);
+          const nestedBackgroundColor = await nestedPush.evaluate((element) => {
+            return getComputedStyle(element).backgroundColor;
+          });
+          await expect(parentlessPush).toHaveCSS(
+            "background-color",
+            nestedBackgroundColor,
+          );
+        }
+      });
+
       test("no color contrast violations (WCAG AA)", async ({ page }) => {
         test.setTimeout(60_000);
         const results = await new AxeBuilder({ page })
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -539,7 +539,6 @@ const layerMathVars = {
   contrastT: _ak.prop.zero("ct", { inherits: true }),
   contrastPushScale: _ak.prop.number("cps", { inherits: true, initial: 1 }),
   layerContrastBias: _ak.var("lcb"),
-  layerSourceDirectionToLight: _ak.prop.zero("lsdtl"),
   forbiddenLa: _ak.var("fla"),
   forbiddenLb: _ak.var("flb"),
   offsetDirectionToLight: _ak.var("odtl"),
@@ -558,11 +557,6 @@ const layerMathVars = {
 
   layerIdlePushResolvedL: _ak.prop("liprl"),
   layerPushValue: _ak.prop.zero("lpv"),
-  // Shared by the idle and state push pipelines. Keep this unregistered and
-  // fallback to offsetDirectionToLight at each call site because that fallback
-  // contains the relative color `l` keyword and only resolves inside the
-  // pushed lightness calc.
-  layerPushDirectionToLight: _ak.var("lpdtl"),
   layerPushBaseL: _ak.prop("lpbl", { initial: l }),
   layerPushL: _ak.prop("lpl", { initial: l }),
   edgeContrastValue: _ak.prop.zero("ecv", { inherits: true }),
@@ -581,7 +575,6 @@ const themeTokenVars = {
 // Color pipeline stages and exported visual tokens.
 const layerColorVars = {
   layerIdleBase: _ak.prop.canvas("lib"),
-  layerSourceDirection: _ak.prop.black("lsd"),
   layerIdleOffset: _ak.prop.canvas("lio"),
   layerIdle: _ak.prop.canvas("li"),
   layerL: _ak.prop.canvas("ll", { inherits: true }),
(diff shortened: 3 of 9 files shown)


packages/ariakit-tailwind/src/input.ts
```diff
@@ -557,10 +557,11 @@ const layerMathVars = {
...
   layerPushValue: _ak.prop.zero("lpv"),
-  // Shared by the idle and state push pipelines: both always receive the same
-  // value (offsetDirectionToLight by default, 0/1 under the light/dark
-  // variants), so one property serves both getPushL call sites.
-  layerPushDirectionToLight: _ak.prop.zero("lpdtl"),
+  // Shared by the idle and state push pipelines. Keep this unregistered and
+  // fallback to offsetDirectionToLight at each call site because that fallback
+  // contains the relative color `l` keyword and only resolves inside the
+  // pushed lightness calc.
+  layerPushDirectionToLight: _ak.var("lpdtl"),
```
diegohaz: **The edge (and box-shadow) push direction still diverges for parentless layers — same gap this PR just closed for the surface.**

The background now matches between parentless and nested (good), but `edgePushDirection` (line 568, `_ak.var("epd", -1)`) wasn't given a parentless fallback, so borders and shadows go the opposite way. All four committed snapshots show the parentless cases with `oklch(0 …)` edges/shadows vs the nested controls' `oklch(1 …)`:

```
parentless: border-[oklch(0_0.028_261.692_/_0.1)]  shadow-[oklch(0_0.028_261.692_/_0.1) …]
nested:     border-[oklch(1_0.028_261.692_/_0.1)]  shadow-[oklch(1_0.028_261.692_/_0.1) …]
```

`epd` has the exact structure the push direction had before this fix: a static default (`-1`) that is only corrected by the `light`/`dark` variants (input.ts:1352/1357), which never match for a parentless layer. So a parentless dark layer keeps `epd = -1` (edge toward dark → black border) while the identically-classed nested layer gets `epd = 1` (edge toward light → white border) — the same "parentless → wrong direction" inconsistency the PR set out to remove, just on the edge.

Is leaving the edge out of scope intentional? If not, note that the obvious mirror of the surface fix does **not** work: the edge color resolves `l` against `var(--ak-layer)`, which by then is the *pushed* (light, ~0.728) layer color, so a `fn.var(vars.edgePushDirection, vars.lightnessOffsetDirection)` fallback would still compute `-1` (the same post-push `l` trap as the contrast-bias path). A correct edge fallback would have to derive the direction from the layer's original lightness, not the post-push `l`.

app/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -567,6 +567,29 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+          ["parentless dark state push", "nested dark state push"],
+        ] as const;
+
+        for (const [parentlessLabel, nestedLabel] of cases) {
+          const parentlessPush = q.region(parentlessLabel);
+          const nestedPush = q.region(nestedLabel);
+          const nestedBackgroundColor = await nestedPush.evaluate((element) => {
+            return getComputedStyle(element).backgroundColor;
+          });
+          await expect(parentlessPush).toHaveCSS(
```
diegohaz: **This assertion only guards `background-color` in normal contrast, so it green-lights the fix while two real divergences in the committed snapshots go unasserted.**

The test compares only `background-color`, and it never forces `prefers-contrast`, so the snapshots are the only thing recording these mismatches — and they're committed as the *expected* output, so nothing fails while the inconsistency persists:

- **Border/shadow — all four modes:** parentless renders `oklch(0 …)` edges/shadows vs the nested control's `oklch(1 …)` (see the edge-direction comment).
- **Background — high-contrast only:** `parentless dark layer push` is `oklch(1 …)` vs nested `oklch(0.9 …)` in `*-high-contrast` (see the contrast-bias thread).

Since the regression is fundamentally "parentless must match nested," the dedicated test would be stronger if it also asserted `border-color` (or compared the full computed-style of the two regions) and ran at least one case under `prefers-contrast: more`. As written it can only ever confirm the one axis (surface lightness, normal contrast) that already happens to match.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1203,7 +1206,11 @@ function getBaseDeclarations(sourceColor: string | VarProperty) {
 }
 
 function getLayerIdleContrastBiasDirection() {
-  const pushDirection = fn.sub(fn.double(vars.layerPushDirectionToLight), 1);
+  const directionToLight = fn.var(
+    vars.layerPushDirectionToLight,
+    vars.offsetDirectionToLight,
+  );
```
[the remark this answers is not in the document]
diegohaz: Confirmed — this reproduces in the committed high-contrast snapshots: `parentless dark layer push` lands at `oklch(1 …)` while `nested dark layer push` is `oklch(0.9 …)` (in both `dark-high-contrast` and `light-high-contrast`). The `state` push matches (`1` on both sides), so the divergence is specific to the idle/layer-push **contrast-bias** path, and it only surfaces when `--contrast > 0` — in normal contrast both sides are `0.7281`, which is why the new test passes.

Root cause: the new `fn.var(layerPushDirectionToLight, offsetDirectionToLight)` fallback derives the direction from the relative-color `l` keyword. That is fine inside the push-escape calc (it resolves against the pre-push base). But `getLayerIdleContrastBiasDirection()` (input.ts:1208) feeds `--_ak-lcb`, which is consumed inside `layerIdle = oklch(from layerIdleOffset, …)` (input.ts:1150-1161). There the `l` keyword resolves to `layerIdleOffset`'s lightness — the **already-pushed** value (~0.9, i.e. "light", set at input.ts:1122-1123) — so `offsetDirectionToLight` collapses to `0`/`-1` and the bias flips. The nested path stays correct because it uses the literal `--_ak-lpdtl: 1` from the `dark` variant, which is `l`-independent.

So a complete fix has to derive the parentless contrast-bias direction from the layer's *original* lightness rather than the post-push `l`. The new test won't catch this regression because it only runs in normal contrast (`prefers-contrast` is never forced) and only asserts `background-color`.

diegohaz (commented):
## Review summary

Initial review (comment-only) of the Tailwind root-push-direction fix.

**Baseline:** merge-base of `origin/main`…`5f4c0610e2` (full diff; no prior review from this skill).
**Scope:** `packages/ariakit-tailwind/src/{input.ts,output.css}`, the `ariakit-tailwind` sandbox + `test-chrome.ts`, the four `__snapshots__`, and the changeset — plus the surrounding push/edge/contrast-bias pipeline beyond the diff.

**Validated locally**
- Rebuilt `output.css` with `node packages/ariakit-tailwind/src/build.ts` (Node 24): regenerates byte-identical to the committed file — it's in sync with `input.ts`.
- Traced the IACVT mechanic end-to-end: `_ak.prop` defaults to `syntax: "*"`, so the now-unregistered `--_ak-lpdtl` fallback (`var(--_ak-lpdtl, var(--_ak-odtl))`) survives the relative-color `l` keyword and resolves inside the three `oklch(from …)` consumers (`layerIdlePushResolvedL` @1123, `layerContrastBias` @1154, `layerPushL` @1175). The `--_ak-ls` → `transparent` change correctly makes a root layer match neither the `light` (`oklch(0 0 0)`) nor `dark` (`oklch(1 0 0)`) style query.
- Confirmed the existing snapshots are unchanged (+36/-0), so no regression to prior cases.

**The core fix is correct** for the surface push in normal contrast (parentless `0.7281` == nested `0.7281`). But the committed snapshots show the fix does **not** fully reach its own "parentless == nested" goal:

1. **Edge/shadow direction (all four modes)** — `edgePushDirection` defaults to `-1` with no parentless fallback, so parentless borders/shadows are `oklch(0 …)` vs the nested controls' `oklch(1 …)`. Same root structure as the bug just fixed. Inline comment on `input.ts`; note the obvious mirror fix hits the post-push-`l` trap.
2. **High-contrast layer-push background** — `parentless` `oklch(1 …)` vs `nested` `oklch(0.9 …)` in `*-high-contrast`. Confirmed the Codex P2: the `offsetDirectionToLight` fallback re-derives direction from the relative-color `l`, which inside the contrast-bias `oklch(from layerIdleOffset …)` is the already-pushed (light) lightness, flipping the bias. Replied on that thread with the root cause.
3. **Test coverage** — the new dedicated test asserts only `background-color` in normal contrast, so it green-lights the fix while (1) and (2) are recorded only as frozen snapshot values. Inline comment suggests asserting `border-color` and a `prefers-contrast: more` case.

None of these are regressions to shipped behavior; they're completeness gaps relative to the PR's stated goal. If the edge/high-contrast cases are intentionally out of scope, it'd be worth saying so (in the changeset or a code comment) since the new test name claims the general "toward the light boundary" behavior.

**Not validated:** I did not execute the Playwright/`test-chrome` suite (worktree dev-server). Findings are grounded in the committed snapshots + the rebuilt `output.css`, which are authoritative for the computed values discussed.

packages/ariakit-tailwind/src/input.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 52d54fd52. Parentless layers now record the pre-push source direction and, in CSS if() browsers, use it for the high-contrast bias fallback so the bias is based on the original dark source instead of the post-push offset. I also updated the PR description and changeset to call out that browsers without CSS if() keep the existing parent/variant fallback for this self-direction path.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 52d54fd52. Edge direction now falls back to the recorded source direction (`var(--_ak-epd, ((var(--_ak-lsdtl) * 2) - 1))` in the generated CSS), so parentless dark push borders and shadows match the nested controls in CSS if() browsers.

app/src/sandbox/ariakit-tailwind/test-chrome.ts
diegohaz: Fixed in 52d54fd52. The regression now runs in normal and high-contrast modes and compares `background-color`, `border-top-color`, and `box-shadow` between parentless and nested layer/state pushes. I confirmed the strengthened test failed before the source fix and passes after it.

diegohaz:
Addressed the unresolved review feedback in 52d54fd52.

**Fixed**

- High-contrast parentless layer push bias now uses the pre-push source direction in CSS `if()` browsers: https://github.com/ariakit/ariakit/pull/6408#discussion_r3461193789
- Parentless dark push edge and shadow direction now falls back to that source direction in generated CSS: https://github.com/ariakit/ariakit/pull/6408#discussion_r3461193932
- The regression now covers normal/high-contrast background, border, and shadow parity for parentless vs nested layer/state pushes: https://github.com/ariakit/ariakit/pull/6408#discussion_r3461194098

**Verified**

- `pnpm lint-fix`
- `APP_PORT=4322 NEXTJS_PORT=3001 pnpm -F app test-chrome ariakit-tailwind`
- `pnpm tsc`

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1249,6 +1279,7 @@ const layerMathDeclarations = [
 // Build the layered color stages from idle -> base -> offset -> final.
 const layerColorDeclarations = [
   set(vars.layerIdleBase, layerIdleBase),
+  set(vars.layerSourceDirection, layerSourceDirection),
```
diegohaz: **`layerSourceDirection` (`--_ak-lsd`) is computed for every layer but only used inside `@supports (if())` — non-`if()` browsers pay for it with no benefit.**

`--_ak-lsd` is set unconditionally here, but its only reader is the `getLayerSourceDirectionToLight()` style query inside the `@supports (color: if(else: red))` block below (`output.css:136` defines it; `output.css:142` is the only use). So every `ak-layer` element resolves this extra `oklch(from --_ak-lib …)` on every paint — including in browsers without `if()`, which can never use the value. Per this plugin's perf model (color math runs on every paint, page-load weighs heaviest and is ~dominated by style recalc), that's an added per-element relative-color resolution paid by the entire non-`if()` userbase for zero benefit — the "gated → unconditional compute" trade the plugin normally avoids. It also stings because, today, `if()` support is the minority case, so most users pay the cost while only a subset get the fix.

Clean, behavior-preserving fix: move the assignment into the existing `@supports` block alongside `layerSourceDirectionToLight`, so non-`if()` browsers never compute it and `if()` browsers read the same computed value:

```ts
const layerColorDeclarations = [
  set(vars.layerIdleBase, layerIdleBase),
-  set(vars.layerSourceDirection, layerSourceDirection),
  set(vars.layerIdleOffset, layerIdleOffset),
  // …
];

  at.supports(
    IF_SUPPORTS_CONDITION,
    // …
+    set(vars.layerSourceDirection, layerSourceDirection),
    set(vars.layerSourceDirectionToLight, getLayerSourceDirectionToLight()),
  ),
```

I prototyped this and rebuilt `output.css`: the **only** delta is `--_ak-lsd` moving from the layer body into the `@supports` block (everything else byte-identical), so the `if()` path — and the snapshots, generated in `if()`-capable Chrome — are unchanged. Worth a `pnpm -F app test-perf ariakit-tailwind` (gated vs current) to confirm the page-load delta, since the PR's perf bot only ran the dialog suites.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -1203,7 +1216,11 @@ function getBaseDeclarations(sourceColor: string | VarProperty) {
 }
 
 function getLayerIdleContrastBiasDirection() {
-  const pushDirection = fn.sub(fn.double(vars.layerPushDirectionToLight), 1);
+  const directionToLight = fn.var(
+    vars.layerPushDirectionToLight,
+    vars.layerSourceDirectionToLight,
+  );
```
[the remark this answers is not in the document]
diegohaz: Agree this is real, but it reads as an accepted, documented limitation rather than a fix gap. The self-direction path is scoped to `if()` on purpose (the changeset now says so), and for the dark cases this PR targets the non-`if()` result is **identical to the previous commit** `5f4c0610e2`: with `--_ak-lsdtl` pinned to its `0` initial the bias direction resolves to the dark side, and the push-escape still uses `--_ak-odtl` — which is exactly what `--_ak-odtl` already evaluated to in the post-push bias context before. So there's no new regression, just the pre-existing high-contrast mismatch left in place for non-`if()` browsers.

The reason it can't be closed without `if()` is structural: a parentless layer has to read its own *pre-push* scheme at the point the bias/edge direction is chosen, and container style queries only see ancestors — `if()` is the only mechanism that can query the element's own computed `--_ak-lsd`. A non-`if()` fix would have to carry the bias contribution forward from the pre-push stage (a larger refactor), because any `l`-derived fallback re-resolves against the post-push lightness at the consumption site (the original root cause here).

One related note: the `--_ak-lsd` color that powers this is currently computed unconditionally even though it's only consumed inside the `@supports` block, so non-`if()` browsers pay for the mechanism without getting the fix — flagged separately on `input.ts`.

diegohaz (commented):
## Re-review summary

Re-review (comment-only) of `52d54fd52` ("Address Tailwind root push review feedback"). Incremental baseline = my last reviewed head `5f4c0610e2`; full PR diff re-checked too.

**All three of my prior findings + the original Codex high-contrast P2 are genuinely fixed in `if()`-capable browsers** — verified against the code/snapshots, not just trusting the "Fixed" replies:

- Rebuilt `output.css` (`node src/build.ts`, Node 24) → byte-identical to committed; in sync.
- Snapshots now show parentless == nested for `background-color`, `border-top-color`, and `box-shadow` in all four modes — e.g. high-contrast parentless layer-push bg `1.0`→`0.9` (matches nested), borders `oklch(0…)`→`oklch(1…)`.
- The strengthened test asserts all three properties under both normal and `prefers-contrast: more` (via CDP), so the regression is now actually guarded.
- Mechanism is sound: `--_ak-lsd` freezes the pre-push source direction as a color at the layer-base stage, and `if()` reads it back on the element itself — which is why parentless layers needed `if()` (container queries can't see the element's own value).

**One new finding (perf, introduced by this commit):** `--_ak-lsd` is computed for every `ak-layer` unconditionally but only consumed inside the `@supports (if())` block, so non-`if()` browsers pay an extra per-element `oklch` for nothing. Behavior-preserving fix (move it into the `@supports` block) prototyped + rebuilt — only that line relocates. Inline comment on `input.ts`.

**Open Codex P2 (non-`if()` high contrast):** valid but a documented, structural limitation — the non-`if()` path is identical to the prior commit (no new regression), and there's no pure-CSS way to read a parentless layer's own pre-push direction without `if()`. Replied on that thread.

**Minor:** the edge fix also flips non-pushed parentless dark layers' borders toward light (visible in the "dark push control" snapshot), which is a correct consistency win but slightly broader than the changeset's "root pushed layers … edge colors" wording.

**Not run:** the Playwright suite and the `ariakit-tailwind` perf comparison (worktree dev-server); findings are grounded in the committed snapshots, the rebuilt `output.css`, and a local prototype.

packages/ariakit-tailwind/src/input.ts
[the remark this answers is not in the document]
diegohaz: This is not being changed in this follow-up because it is the documented non-`if()` limitation rather than a new regression. Parentless high-contrast bias/edge parity needs the element’s own pre-push source direction; without CSS `if()` there is no self-style query, and container style queries can only read ancestors. The PR body and changeset now scope that path to CSS `if()` support and keep the existing parent/variant fallback for non-`if()` browsers.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in ec97029bd. `--_ak-lsd` is now assigned inside the existing `@supports (color: if(else: red))` block immediately before `--_ak-lsdtl`, so non-`if()` browsers no longer resolve that extra source-direction relative color. The rebuilt `output.css` only moves this declaration, and the fresh perf comparison improved page-load total by ~1% and toggle total by ~14%.

diegohaz:
Handled the latest unresolved review feedback.

**Fixed**

- ec97029bd (https://github.com/ariakit/ariakit/pull/6408#discussion_r3461573535): moved `--_ak-lsd` into the CSS `if()` support block so non-`if()` browsers avoid the extra source-direction relative color work.

**Already documented**

- https://github.com/ariakit/ariakit/pull/6408#discussion_r3461573401: kept the non-`if()` parentless high-contrast bias/edge limitation as documented, since that path needs a self-style query and container style queries only read ancestors.

**Verified**

- `pnpm lint-fix`
- `APP_PORT=4321 pnpm -F app test-chrome ariakit-tailwind`
- `pnpm -F app test-perf ariakit-tailwind` with fresh baseline/current comparison
- `pnpm tsc`

app/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -567,6 +567,59 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+                backgroundColor: style.backgroundColor,
+              };
+            });
+            await expect(parentlessPush).toHaveCSS(
+              "background-color",
+              referenceStyles.backgroundColor,
+            );
+            await expect(lightParentPush).toHaveCSS(
+              "background-color",
+              referenceStyles.backgroundColor,
```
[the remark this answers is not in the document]
diegohaz: This is intentionally not restored. The follow-up in 20510c35d narrows the regression contract to pushed layer background lightness because parentless edge direction is deliberately unsupported; asserting border/shadow parity here would reintroduce that unsupported parentless-edge requirement. Light-parent and dark-parent background cases are still covered so push now resolves from the current layer regardless of parent.

diegohaz:
**Fixed**
- 20510c35d: Simplified push direction so `ak-layer-push-*` and `ak-state-push-*` resolve background lightness from the current layer, including parentless, dark-parent, and light-parent cases. The CSS `if()` self-source path was removed, and parentless edge direction is intentionally out of scope.

**Invalid**
- CodeRabbit border/shadow parity request ([reply](https://github.com/ariakit/ariakit/pull/6408#discussion_r3461837910)): parentless edge direction is deliberately unsupported, so the regression test should not assert border/shadow parity for parentless push.

.changeset/300-tailwind-root-push-direction.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/tailwind": patch
+---
+
+Fixed `ak-layer-push-*` and `ak-state-push-*` in `@ariakit/tailwind` to escape the forbidden lightness band based on the current layer, regardless of the parent layer.
```
diegohaz, line 5: **Backward-compat blast radius worth confirming (the new rule isn't limited to root layers).** 

The simplification is a nice cleanup — driving the push escape purely off the current layer's own lightness (`offsetDirectionToLight`) removes the whole `if()` / `--_ak-lsd` / `--_ak-lsdtl` / variant-direction machinery and resolves the high-contrast case in every browser, not just `if()` ones. But because the direction is now the current layer's own lightness regardless of parent, it also changes **already-nested** layers, not only parentless/root ones: any `ak-layer-push-*` / `ak-state-push-*` whose own lightness sits on the opposite side of the ~0.645 split from its parent's scheme now escapes the other way.

That's visible in the regenerated snapshots for pre-existing cases, e.g. (light scheme):
- `ak-layer! modifier cascade › state push`: `oklch(0.5321 …)` → `oklch(0.734 …)` (text `white`→`black`)
- `layer limits › state push`: `oklch(0.5498 …)` → `oklch(0.7251 …)`

i.e. a dark-ish colored layer (l < 0.645) inside a light container now pushes toward light instead of dark. These look correct and intentional (they also make `state-push` consistent with `layer-push`), and the changeset's "regardless of the parent layer" does cover it — just flagging that the change surface is broader than the parentless-root repro in #6343, so it's worth a deliberate "yes, existing nested colored pushes changing color is acceptable as a `patch`" before merge. No change requested if that's the intent.

diegohaz (commented):
## Re-review summary

Re-review (comment-only) of the simplification — commits `ec97029bd` ("Avoid unconditional Tailwind source direction work") and `20510c35d` ("Simplify Tailwind push direction handling"). Incremental baseline = my last reviewed head `52d54fd52`; full PR diff re-checked.

**This is a strict improvement that supersedes the earlier `if()` approach and closes every prior finding — verified, not just trusting the resolved threads:**

- The whole `if()` / `--_ak-lsd` / `--_ak-lsdtl` / `--_ak-lpdtl` / variant-direction mechanism is gone (~80 fewer lines in `input.ts`, 6 properties + 3 helpers removed). Push direction is now uniformly `offsetDirectionToLight` (the current layer's own lightness).
- **My perf finding is fully resolved** — the unconditional `--_ak-lsd` oklch no longer exists at all (not just gated).
- **The Codex non-`if()` P2 is now moot** — there's no `if()` dependency, so the high-contrast parentless fix works in every browser. Snapshots confirm parentless == nested == light-parent for `background-color` in all four modes (e.g. high-contrast layer-push `0.9`, state-push `1.0` on every side).
- **Bias relocation is correct** (no double-count / no loss): `layerIdleContrastBias` is added to `layerIdlePushResolvedL` when pushing and to `layerIdle` gated by `(1 - layerIdlePushEnabled)` when not — and `layerIdlePushEnabled=1` / `layerIdlePushResolvedL` are set together in the same `layer-push-*` utility, so the two paths are mutually exclusive. In normal contrast the bias is `0`, so it's a no-op there. Moving it into `layerIdlePushResolvedL` (consumed in the pre-push `layerIdleOffset` `oklch(from layerIdleMixed …)`) is exactly what makes the `l`-keyword direction resolve against the original lightness without `if()`.
- Rebuilt `output.css` (`node src/build.ts`, Node 24) → byte-identical to committed; in sync. `IF_SUPPORTS_CONDITION` is still used by the text `@supports` blocks, so no dead code left behind.
- The strengthened test now compares parentless **and** light-parent backgrounds against the dark-parent reference across normal + `prefers-contrast: more`; dropping the `border-top-color`/`box-shadow` assertions is consistent with the edge fix being intentionally rescoped out (parentless edge is back to parent-relative, per the resolved CodeRabbit thread).

**One thing to confirm (not a defect):** the new rule also changes **pre-existing nested** push/state-push colors — any push whose own lightness is on the opposite side of the ~0.645 split from its parent's scheme now escapes the other way (two such cases in the regenerated light snapshots). Documented by the changeset's "regardless of the parent layer," and the shifts look correct, but the blast radius is broader than the parentless-root repro, so worth an explicit "OK as a patch" call. Left as an inline note on the changeset.

**Not run:** the Playwright suite (worktree dev-server); findings are grounded in the committed snapshots, the rebuilt `output.css`, and code tracing.

No substantive defects found — the simplified implementation is correct, cleaner, and resolves the issue more broadly than the `if()` version.

app/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -567,6 +567,59 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
 
+      for (const contrast of [false, true]) {
+        const label = contrast ? " in high contrast" : "";
+        test(`pushes dark layers toward the light boundary regardless of parent${label}`, async ({
```
diegohaz: Do we need this test? Isn't this covered by snapshots?

diegohaz: Fixed in 9c1430228 - removed the redundant explicit Playwright assertion. The computed-style snapshots still cover the same parentless, dark-parent, and light-parent push fixtures across light/dark schemes and normal/high-contrast modes. I also reworded the fixture comment to describe inherited layer scheme isolation.

diegohaz:
**Fixed**
- 9c1430228 ([reply](https://github.com/ariakit/ariakit/pull/6408#discussion_r3462102329)): Removed the redundant explicit Playwright push assertion because the computed-style snapshots already cover the parentless, dark-parent, and light-parent push fixtures across light/dark schemes and normal/high-contrast modes. Also clarified the fixture isolation comment.

PR #6410, opened 2026-06-23
Fix Tailwind light band boundary

### LINKED ISSUE #6348: @ariakit/tailwind: a layer at exactly l = 0.8575 matches neither ak-light-low: nor ak-light-high:
### Current behavior

The `@ariakit/tailwind` lightness band variants (`ak-dark-high:`, `ak-dark-low:`, `ak-light-low:`, `ak-light-high:`) are meant to partition the whole layer lightness scale, so every layer matches exactly one band. But a layer whose OKLCH lightness is exactly `0.8575` — reachable with `ak-layer-l-85.75`, a valid quarter-step bare value, or `ak-layer-l-[0.8575]` — matches **neither** `ak-light-low:` nor `ak-light-high:`. Styles gated behind either variant silently don't apply, even though the layer is unambiguously light (the binary `ak-light:` variant still matches). Nudging the lightness to `0.857` (light-low) or `0.858` (light-high) "fixes" it, which makes the root cause hard to spot.

Band membership is computed by [`getRangeMask`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L264-L272), which deliberately makes each band's lower boundary inclusive by subtracting a `1e-6` epsilon from `min` while keeping the upper boundary exclusive — so at every shared boundary the value belongs to exactly one band. The first three bands use the helper, but the topmost band is [hand-rolled and omits the epsilon](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L439-L443):

```ts
const bandDarkHigh = getRangeMask(l, 0, DARK_HIGH_MAX_L);
const bandDarkLow = getRangeMask(l, DARK_HIGH_MAX_L, LA_BASE);
const bandLightLow = getRangeMask(l, LB_BASE, LIGHT_LOW_MAX_L);
const bandLightHigh = fn.binary(fn.sub(l, LIGHT_LOW_MAX_L));
```

The asymmetry shows in the [generated CSS](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L1358-L1374): the dark-side masks use epsilon-shifted bounds (`0.274999`, `0.724999`), but `--_ak-blh` uses the raw boundary — `clamp(0, ((l - 0.8575) * 1000000), 1)`. At `l = 0.8575` both the `--_ak-bll` upper factor and `--_ak-blh` evaluate to 0, so all four masks are 0 and [`layerBandLightness`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/input.ts#L1261-L1269) collapses to the mid level. The [band variants](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-tailwind/src/output.css#L90-L112) compare `--_ak-lbd` against exact `oklch()` values via `@container style(...)`, and `oklch(0.5 0 0)` matches none of them.

This was verified with a real-browser Playwright run: the computed `--_ak-lbd` of the `ak-layer-l-85.75` layer is `oklch(0.5 0 0)` and the band-gated background stays transparent, while a control layer at the dark-side boundary (`ak-layer-l-27.5`, l = 0.275, where the epsilon exists) resolves to `oklch(0.25 0 0)` and matches exactly one variant (`ak-dark-low:`).

3. Look at the two cards. The dark card's inner text gets the blue `ak-dark-low:` background — the boundary value 0.275 belongs to exactly one band (control). The light card's inner text is underlined (so `ak-light:` matches), but it has **no background at all**: neither `ak-light-low:` nor `ak-light-high:` matched at l = 0.8575.

```tsx
  return (
    <div className="grid gap-4">
      {/* Control: the dark-side band boundary. l = 0.275 is the shared
          boundary between ak-dark-high and ak-dark-low; it belongs to
          ak-dark-low, so this child gets the blue background. */}
      <div className="ak-layer ak-layer-l-27.5 p-4">
        <div className="ak-dark-high:bg-red-500 ak-dark-low:bg-blue-500 p-2">
          Dark band boundary (layer lightness 0.275)
        </div>
      </div>
      {/* Bug: the light-side band boundary. l = 0.8575 is the shared boundary
          between ak-light-low and ak-light-high, so one of these two variants
          should match — but neither does. */}
      <div className="ak-layer ak-layer-l-85.75 p-4">
        <div className="ak-light:underline ak-light-low:bg-red-500 ak-light-high:bg-blue-500 p-2">
          Light band boundary (layer lightness 0.8575)
        </div>
      </div>
    </div>
  );
}
```

### Expected behavior

Every layer lightness — including the band boundaries — should belong to exactly one band. Mirroring the dark side, where 0.275 belongs to the upper band (`ak-dark-low`), l = 0.8575 should belong to `ak-light-high`, so the light card's inner text should get the blue background.

### Workaround

Avoid the exact boundary value by nudging the lightness an imperceptible amount with an arbitrary value:

```diff
-<div className="ak-layer ak-layer-l-85.75 p-4">
+<div className="ak-layer ak-layer-l-[0.85751] p-4">
```

The arbitrary form is required: bare values only accept quarter steps, so a class like `ak-layer-l-85.751` generates nothing.

### Possible solutions

Apply the same inclusive-min epsilon idiom that `getRangeMask` uses, so l = 0.8575 lands in `bandLightHigh`:

```diff
 const bandLightLow = getRangeMask(l, LB_BASE, LIGHT_LOW_MAX_L);
-const bandLightHigh = fn.binary(fn.sub(l, LIGHT_LOW_MAX_L));
+const bandLightHigh = fn.binary(
+  fn.sub(l, roundToDecimals(LIGHT_LOW_MAX_L - 1e-6, 6)),
+);
```

Rebuilding (`pnpm -F @ariakit/tailwind build`) changes exactly one line of `output.css`:

```diff
 @property --_ak-blh {
   syntax: "*";
   inherits: false;
-  initial-value: clamp(0, ((l - 0.8575) * 1000000), 1);
+  initial-value: clamp(0, ((l - 0.857499) * 1000000), 1);
 }
```

Inside the `1e-6` crossfade window the `bandLightLow` and `bandLightHigh` weights still sum to 1, so no new gap or overlap is introduced; behavior only changes for l ∈ [0.857499, 0.8575]. Note the band should not be "regularized" to `getRangeMask(l, LIGHT_LOW_MAX_L, 1)` — the helper's upper bound is exclusive, which would kick pure white (l = 1) out of `bandLightHigh`; the top band must stay one-sided.

diegohaz:
## Motivation

The `@ariakit/tailwind` lightness band variants should partition the layer lightness scale so each layer matches exactly one band. At `ak-layer-l-85.75`, the light-low upper bound and light-high lower bound both evaluated to zero, so neither `ak-light-low:` nor `ak-light-high:` matched.

Fixes #6348.

## Solution

Shift `bandLightHigh` by the same inclusive lower-bound epsilon used by `getRangeMask`, while keeping the top band one-sided so `l = 1` still belongs to `ak-light-high:`. Rebuild `output.css` so `--_ak-blh` starts at `0.857499` instead of `0.8575`.

Add a Tailwind sandbox regression that checks both sides of the relevant dark and light boundaries: the losing adjacent variant remains transparent and the owning variant applies.

### DIFF
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -563,6 +563,27 @@ async function waitForPreviewReady(page: Page) {
 }
 
 withFramework(import.meta.dirname, async ({ test }) => {
+  test("matches exactly one layer band at lightness boundaries", async ({
+    page,
+    q,
+  }) => {
+    await waitForPreviewReady(page);
+    const transparent = "rgba(0, 0, 0, 0)";
+
+    await expect(
+      q.text("Dark high candidate (layer lightness 0.275)"),
+    ).toHaveCSS("background-color", transparent);
+    await expect(
+      q.text("Dark low candidate (layer lightness 0.275)"),
+    ).toHaveCSS("background-color", "rgb(0, 0, 0)");
+    await expect(
+      q.text("Light low candidate (layer lightness 0.8575)"),
+    ).toHaveCSS("background-color", transparent);
+    await expect(
+      q.text("Light high candidate (layer lightness 0.8575)"),
+    ).toHaveCSS("background-color", "rgb(0, 0, 0)");
+  });
+
   for (const scheme of ["light", "dark"] as const) {
     test.describe(`${scheme} scheme`, () => {
       test.use({ colorScheme: scheme });
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -440,7 +440,9 @@ const globalContrastT = fn.div(fn.relu(contrast), CONTRAST_HIGH);
 const bandDarkHigh = getRangeMask(l, 0, DARK_HIGH_MAX_L);
 const bandDarkLow = getRangeMask(l, DARK_HIGH_MAX_L, LA_BASE);
 const bandLightLow = getRangeMask(l, LB_BASE, LIGHT_LOW_MAX_L);
-const bandLightHigh = fn.binary(fn.sub(l, LIGHT_LOW_MAX_L));
+const bandLightHigh = fn.binary(
+  fn.sub(l, roundToDecimals(LIGHT_LOW_MAX_L - 1e-6, 6)),
+);
 
 function getLayerTextLightness() {
   const lightChromaDamping = fn.mul(
--- a/packages/ariakit-tailwind/src/input.ts
+++ b/packages/ariakit-tailwind/src/input.ts
@@ -262,13 +262,18 @@ function getWcagLightnessTarget(parentLightness: number, isDark: boolean) {
 }
 
 /**
- * Returns whether `value` is within `[min, max]`
+ * Returns whether `value` is greater than or equal to `min`. The epsilon keeps
+ * adjacent bands from leaving a gap at shared boundaries.
+ */
+function getMinMask(value: Value, min: number) {
+  return fn.binary(fn.sub(value, roundToDecimals(min - 1e-6, 6)));
+}
+
+/**
+ * Returns whether `value` is within `[min, max)`.
  */
 function getRangeMask(value: Value, min: number, max: number) {
-  return fn.mul(
-    fn.binary(fn.sub(value, roundToDecimals(min - 1e-6, 6))),
-    fn.binary(fn.sub(max, value)),
-  );
+  return fn.mul(getMinMask(value, min), fn.binary(fn.sub(max, value)));
 }
 
 /**
@@ -440,9 +445,7 @@ const globalContrastT = fn.div(fn.relu(contrast), CONTRAST_HIGH);
 const bandDarkHigh = getRangeMask(l, 0, DARK_HIGH_MAX_L);
 const bandDarkLow = getRangeMask(l, DARK_HIGH_MAX_L, LA_BASE);
 const bandLightLow = getRangeMask(l, LB_BASE, LIGHT_LOW_MAX_L);
-const bandLightHigh = fn.binary(
-  fn.sub(l, roundToDecimals(LIGHT_LOW_MAX_L - 1e-6, 6)),
-);
+const bandLightHigh = getMinMask(l, LIGHT_LOW_MAX_L);
 
 function getLayerTextLightness() {
   const lightChromaDamping = fn.mul(
--- a/.changeset/300-tailwind-light-band-boundary.md
+++ b/.changeset/300-tailwind-light-band-boundary.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/tailwind": patch
+---
+
+Fixed `ak-light-high:` to match layers at the exact lightness boundary shared with `ak-light-low:`.
--- a/app/src/sandbox/ariakit-tailwind/__snapshots__/dark
+++ b/app/src/sandbox/ariakit-tailwind/__snapshots__/dark
@@ -33,6 +33,13 @@
       "#165DFC": "bg-[oklch(0.1634_0.0091_264.28)] text-[oklch(1_0_0)] *:text-[lch(57.033_80_288.704)] rounded-[15px] border-[1px] border-[oklch(1_0.0091_264.28_/_0.1)] p-[4px] shadow-[oklch(1_0.0091_264.28_/_0.1)_0px_0px_0px_0px]",
       "raw calc": "bg-[oklch(0.4268_0.0091_264.28)] text-[oklch(1_0_0)] *:text-[lch(84.1828_3.43_266.444)] rounded-[15px] border-[1px] border-[oklch(1_0.0091_264.28_/_0.1)] p-[4px] shadow-[oklch(1_0.0091_264.28_/_0.1)_0px_0px_0px_0px]",
       "raw l/c": "bg-[oklch(0.5096_0.18_264.28)] text-[oklch(1_0_0)] *:text-[lch(97.3245_20_284.288)] rounded-[15px] border-[1px] border-[oklch(1_0.18_264.28_/_0.1)] p-[4px] shadow-[oklch(1_0.18_264.28_/_0.1)_0px_0px_0px_0px]",
+      "band boundary variants": {
+        "class": "bg-[oklch(0.1634_0.0091_264.28)] text-[oklch(1_0_0)] rounded-[15px] border-[1px] border-[oklch(1_0.0091_264.28_/_0.1)] p-[4px] shadow-[oklch(1_0.0091_264.28_/_0.1)_0px_0px_0px_0px]",
+        "children": {
+          "ak-layer-l-27.5": "bg-[oklch(0.275_0.0091_264.28)] text-[oklch(1_0_0)] *:text-[rgb(255,_255,_255)] p-[8px]",
+          "ak-layer-l-85.75": "bg-[oklch(0.8575_0.0091_264.28)] text-[oklch(0_0_0)] p-[8px]"
+        }
+      },
       "ak-layer ak-text": "text-[lch(57.033_80_288.648)] rounded-[15px] border-[1px] border-[oklch(1_0.0091_264.28_/_0.1)] p-[4px] shadow-[oklch(1_0.0091_264.28_/_0.1)_0px_0px_0px_0px]",
       "raw ink/edge": "bg-[oklch(0.1634_0.0091_264.28)] text-[oklch(1_0_0_/_0.4969)] *:text-[lch(57.033_2.32_268.203)] rounded-[15px] border-[1px] border-[oklch(1_0.0091_264.28_/_0.35)] p-[4px] shadow-[oklch(1_0.0091_264.28_/_0.35)_0px_0px_0px_0px]",
       "layer limits": {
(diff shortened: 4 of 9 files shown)


app/src/sandbox/ariakit-tailwind/test-chrome.ts
```diff
@@ -563,6 +563,27 @@ async function waitForPreviewReady(page: Page) {
 }
 
 withFramework(import.meta.dirname, async ({ test }) => {
+  test("matches exactly one layer band at lightness boundaries", async ({
```
diegohaz, line 637: Is this test really necessary considering we already have the snapshots?

diegohaz: Yes — the snapshots don't cover this fix, so the Playwright test is the only real regression guard.

`extractComputedCSS` only emits a node per **labelled `<section>`** (the `Layer`/`Cell` components), and `extractClass` records that section's *own* `background-color` plus child *text* colors — never child *background* colors. The band-gated backgrounds in this sandbox live on the plain `<div>` candidates nested inside the layer, which `walk` recurses into but emits nothing for (no `aria-labelledby`, no element children). That's why the new `band boundary variants` snapshot has only the two `ak-layer-l-*` wrappers and no candidate nodes.

Crucially, the `ak-layer-l-85.75` entry records `bg-[oklch(0.8575 …)]` — the layer's *forced* lightness, set by `ak-layer-l-85.75` directly and entirely independent of whether `ak-light-high:` matched. So if this fix regressed (`--_ak-blh` back to `0.8575`), all four snapshot files would stay byte-identical and still pass; only the candidate `<div>` backgrounds would flip to transparent. The `toHaveCSS("background-color", …)` assertions on those candidates are the only thing that actually exercises the boundary.

If you'd rather lean on the snapshots and drop the explicit test, you'd need to pull the candidates into the snapshot — e.g. render them as labelled `Cell`/`Layer` sections, or extend `extractClass`/`walk` to record child background colors — so the boundary resolution shows up there.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -440,7 +440,9 @@ const globalContrastT = fn.div(fn.relu(contrast), CONTRAST_HIGH);
 const bandDarkHigh = getRangeMask(l, 0, DARK_HIGH_MAX_L);
 const bandDarkLow = getRangeMask(l, DARK_HIGH_MAX_L, LA_BASE);
 const bandLightLow = getRangeMask(l, LB_BASE, LIGHT_LOW_MAX_L);
-const bandLightHigh = fn.binary(fn.sub(l, LIGHT_LOW_MAX_L));
+const bandLightHigh = fn.binary(
+  fn.sub(l, roundToDecimals(LIGHT_LOW_MAX_L - 1e-6, 6)),
```
diegohaz: Optional / non-blocking. The inclusive-lower-bound epsilon idiom is now duplicated: `getRangeMask` uses `fn.binary(fn.sub(value, roundToDecimals(min - 1e-6, 6)))` for its lower factor, and `bandLightHigh` repeats it inline. Since this bug existed *precisely* because the top band hand-rolled its lower bound and omitted that epsilon, factoring the shared piece into a helper makes the two impossible to drift apart and states the intent in one place:

```ts
// 1 once `value >= min`, with the epsilon that keeps adjacent bands partitioned.
function atLeast(value: Value, min: number) {
  return fn.binary(fn.sub(value, roundToDecimals(min - 1e-6, 6)));
}

function getRangeMask(value: Value, min: number, max: number) {
  return fn.mul(atLeast(value, min), fn.binary(fn.sub(max, value)));
}

const bandLightHigh = atLeast(l, LIGHT_LOW_MAX_L);
```

I applied this locally and rebuilt — `output.css` is byte-identical — so it's purely a readability/maintainability call. The top band staying one-sided (no upper factor) is preserved.

diegohaz (commented):
## Review summary

**Scope:** Initial review. Baseline = merge-base `588f8d83` … head `7afe3d8e` (full diff of all 9 changed files), expanded into the surrounding `@ariakit/tailwind` band math, the generated `output.css`, the sandbox snapshot serializer, and CI status.

**Verdict:** The fix is correct, minimal, and well-targeted. I have no blocking findings — two non-blocking notes below.

### What I verified
- **Correctness of the fix.** Mirroring `getRangeMask`'s inclusive-min epsilon onto `bandLightHigh` makes the four bands a clean partition: `bdh [0, 0.275)`, `bdl [0.275, 0.55)`, `bll [0.725, 0.8575)`, `blh [0.8575, 1]` — each shared boundary owned by exactly one band, top band left one-sided so `l = 1` stays in `ak-light-high:`. At `l = 0.8575`, `bll`'s upper factor is 0 and `blh` now evaluates to 1, so the previously-uncovered boundary resolves to light-high. No new overlap is introduced (the sub-epsilon crossfade window is far below the 4-decimal granularity of real layer-lightness values).
- **Build output.** Ran `node packages/ariakit-tailwind/src/build.ts` (Node 24) on the committed `input.ts` → regenerated `output.css` is byte-identical to what's committed; the only change is `--_ak-blh` start `0.8575` → `0.857499`, exactly as described.
- **Issue match.** Reproduces the analysis in #6348 precisely, including the "don't regularize to `getRangeMask(l, …, 1)`" caveat (that would evict `l = 1`).
- **CI.** All checks green at head, including Chrome / Test Chrome (so the new boundary test runs and passes).
- **Changeset.** `300-` patch prefix, single past-tense sentence, correct `@ariakit/tailwind` frontmatter — matches the changeset conventions. The variant names are plain inline code (no public reference page to link).

### Notes (non-blocking)
1. **Re: the author's open question on `test-chrome.ts` ("is this test necessary given the snapshots?")** — replied on that thread: the snapshots do **not** cover this fix. The serializer only records labelled-`<section>` backgrounds, not the band-gated backgrounds on the candidate `<div>`s; the `ak-layer-l-85.75` snapshot entry reflects the layer's forced lightness, which is identical whether or not `ak-light-high:` matches. A regression of this fix would leave all four snapshots byte-identical and passing, so the explicit `toHaveCSS` assertions are the only real guard. Keep the test (or pull the candidates into the snapshot if you'd prefer to drop it).
2. **Optional simplification** on `input.ts` — the inclusive-lower-bound epsilon is now duplicated between `getRangeMask` and `bandLightHigh`; extracting a shared `atLeast(value, min)` helper makes them impossible to drift apart (which is the exact failure mode this bug came from). Validated locally as byte-identical output.

### Not fully validated
- Did not run the Chrome sandbox suite locally (relied on green CI for the new test) — the static analysis of the snapshot serializer and band math is the basis for note #1.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in 9aa2c0d95 - extracted the inclusive lower-bound band mask into `getMinMask`, reused it from `getRangeMask`, and kept `bandLightHigh` one-sided through the same helper. The package rebuild leaves `output.css` unchanged.

diegohaz:
**Fixed**

- 9aa2c0d95 ([reply](https://github.com/ariakit/ariakit/pull/6410#discussion_r3460763159)): Extracted the inclusive lower-bound band mask into `getMinMask`, reused it from `getRangeMask`, and kept `bandLightHigh` one-sided through the shared helper.

**Verification**

- `pnpm lint-fix`
- `pnpm tsc`
- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app test-chrome ariakit-tailwind`
- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app test-chrome ariakit-tailwind -g wcag`

packages/ariakit-tailwind/src/input.ts
diegohaz: Verified on `9aa2c0d9`: `getMinMask` cleanly captures the inclusive-lower-bound epsilon, `getRangeMask` reuses it, and `bandLightHigh` stays one-sided through it — the single source of truth that makes the original drift impossible to reintroduce. Bonus catch: `getRangeMask`'s JSDoc is now correctly `[min, max)`. I rebuilt at this commit and `output.css` is byte-identical to committed (and unchanged from the previous head), so it's confirmed behavior-preserving. Resolved 👍

diegohaz (commented):
## Re-review summary

**Re-review** covering only the commit added since my last review.
- Baseline: `7afe3d8e` (last reviewed) → head `9aa2c0d9`.
- Incremental diff: `packages/ariakit-tailwind/src/input.ts` only — commit *"Extract Tailwind band lower mask"*.

### Resolved
- **Note #2 (extract the shared epsilon helper)** — done in `9aa2c0d9`. The author extracted `getMinMask(value, min)`, reused it from `getRangeMask`, and routed `bandLightHigh` through it, giving the inclusive-lower-bound epsilon a single source of truth (the exact drift that produced the original bug is now structurally impossible). `getRangeMask`'s JSDoc was also corrected from `[min, max]` to `[min, max)`. Verified behavior-preserving: rebuilt at this commit → `output.css` byte-identical to committed, and unchanged from the previous head. Naming (`getMinMask`) fits the existing `getRangeMask`/`getForbiddenRangeMask` convention.

### Still standing (no action needed)
- **Note #1 (is the Playwright test necessary?)** — the test was kept in this commit; my evidence on that thread stands (the snapshot serializer records only the labelled-`<section>` layer backgrounds, i.e. the forced `0.8575` lightness, not the band-gated candidate `<div>` backgrounds — so a regression of this fix would leave all four snapshots byte-identical and passing). The explicit `toHaveCSS` assertions remain the only real guard.

### New findings
- None. The refactor is correct, minimal, and introduces no new complexity or regressions.

PR #6412, opened 2026-06-23
Upgrade Astro to v7

diegohaz:
## Motivation

Astro 7 is available and the app workspace is still pinned to Astro 6 and Vite 7. The v7 upgrade also moves the official Astro integrations to their matching major versions and enables the Rust compiler by default, so the app needs to follow the migration guide before future framework work can build on the current release line.

## Solution

Upgraded the app workspace to Astro 7 and the matching official integration majors:

| Package | From | To | Links |
| --- | --- | --- | --- |
| `astro` | 6.4.8 | 7.0.1 | [upgrade guide](https://docs.astro.build/en/guides/upgrade-to/v7/) · [repo](https://github.com/withastro/astro) |
| `@astrojs/cloudflare` | 13.7.0 | 14.0.0 | [changelog](https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/CHANGELOG.md#1400) · [repo](https://github.com/withastro/astro) |
| `@astrojs/mdx` | 6.0.3 | 7.0.0 | [changelog](https://github.com/withastro/astro/blob/main/packages/integrations/mdx/CHANGELOG.md#700) · [repo](https://github.com/withastro/astro) |
| `@astrojs/react` | 5.0.7 | 6.0.0 | [changelog](https://github.com/withastro/astro/blob/main/packages/integrations/react/CHANGELOG.md#600) · [repo](https://github.com/withastro/astro) |
| `@astrojs/solid-js` | 6.0.1 | 7.0.0 | [changelog](https://github.com/withastro/astro/blob/main/packages/integrations/solid/CHANGELOG.md#700) · [repo](https://github.com/withastro/astro) |
| `vite` | 7.3.5 | 8.0.16 | [migration guide](https://vite.dev/guide/migration.html) · [repo](https://github.com/vitejs/vite) |

The migration-guide checks did not require most app source changes: there is no `src/fetch.ts`, `@astrojs/db`, removed `astro:transitions` internals, or old experimental config to migrate. The app already uses `@astrojs/markdown-remark` for the unified Markdown processor path.

Astro 7 changes the default HTML whitespace handling to JSX-style whitespace. The app had Astro templates that intentionally relied on HTML-aware whitespace between inline content split across lines, including copy text, collection headings, admin inline inputs, and OG-image headings. Those templates now use explicit spaces where needed instead of keeping Astro 6's `compressHTML` behavior. The generated OG images were regenerated and stayed within tolerance; the final PR diff has no net OG image changes.

Astro 7's Rust compiler rejected a top-level `return` in `@clerk/astro@3.4.6`, so this adds a small pnpm patch that preserves Clerk's runtime behavior by turning the production early return into an `else` branch. `@clerk/astro` does not currently have a stable release declaring Astro 7 peer support, so `pnpm-workspace.yaml` also adds a targeted `@clerk/astro>astro` peer allowance. The patch can be removed once Clerk ships stable Astro 7 support and this component no longer needs the compiler workaround.

The app's direct Vite pin and Renovate's app-specific Vite rule were moved to major 8. I also audited the repo for Vite-version-mismatch workarounds. The app's SSR optimizer workaround still reproduces a cold-start React hook failure on Astro 7/Vite 8 when removed, so it stays and now points to the active Astro issue [#17166](https://github.com/withastro/astro/issues/17166). The `vitest.config.ts` plugin cast was removed by typing the root plugin list as `PluginOption[]`, while `sourcePlugin` now follows Vite's documented plugin factory shape by returning `Plugin`. Vite does not export `PluginContext` directly, so the helper-only source plugin context is derived from Vite's public `Plugin["load"]` hook type instead of duplicating a manual context shape.

`astro@7.0.1` is listed in `minimumReleaseAgeExclude` because it was published today and the repo's pnpm supply-chain policy otherwise blocks the install.

## Verification

- `pnpm install`
- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build-app-lite`
- `pnpm -F app run build`
- `APP_PORT=45321 NEXTJS_PORT=45322 pnpm -F app run test-chrome` (155 passed, 1 skipped; the 2 reference partial tests failed because this command was run against the lite build without reference partial pages)
- `APP_PORT=45321 NEXTJS_PORT=45322 pnpm -F app exec playwright test --project=chrome src/tests/reference-hovercards-browser.ts`
- `OG_IMAGE_BASE_URL=http://localhost:4322 pnpm -F app run og-image-generate`
- `rg -n 'ReactComponents|SolidComponents|ReactExamples|SolidExamples|Search ReactComponents|Search SolidComponents|85React|0Solid|click the<span|Expires in<input|/>days|Up to<input|/>redemptions' app/dist/client -g '*.html'` (no matches)
- `pnpm exec tsc -b --pretty false`
- `pnpm test app/src/lib/source-plugin.test.ts`
- `git diff --check`
- Pre-commit PR review gate: native Codex reviewer + Cursor CLI reviewer, followed by native/Cursor re-review after addressing the chip-badge separator and stale Vitest TODO findings
- Pre-commit PR review gate for the `vitest.config.ts` cast removal: native Codex reviewer + Cursor CLI reviewer
- Pre-commit PR review gate for the Vite plugin typing correction: native Codex reviewer + Cursor CLI reviewer

I also verified the `vitest.config.ts` plugin cast is gone by running `pnpm exec tsc -b --pretty false` after switching `sourcePlugin` to return Vite's public `Plugin` type and deriving its helper context from `Plugin["load"]`.

`pnpm peers check` now only reports the pre-existing `@wordpress/theme@0.15.1` / `stylelint@17.13.0` mismatch from `origin/main`.


app/package.json
```diff
@@ -70,7 +70,7 @@
     "@stripe/stripe-js": "9.8.0",
     "@tailwindcss/vite": "4.3.1",
     "@tanstack/react-query": "5.101.0",
-    "astro": "6.4.8",
+    "astro": "7.0.1",
```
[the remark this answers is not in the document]
diegohaz: Fixed in d66cbe661a655b95d9e308a05a02f8b584782d7e.

I set `compressHTML: true` so Astro 7 keeps Astro 6's HTML-aware whitespace behavior. I also checked the Astro files for this pattern and found it was broader than the two examples here, including the OG image template itself, so the config-level fix is the safer migration path. The generated OG image changes were reverted, and the final PR diff has no net OG image files.

Verified with `pnpm tsc`, `pnpm -F app run build-lite`, `git diff --check origin/main`, built HTML checks for the whitespace examples, and `pnpm -F app run og-image-generate` against a local Astro server.

diegohaz:
Follow-up for the Astro 7 whitespace review is pushed in d66cbe661a655b95d9e308a05a02f8b584782d7e.

Fixed:
- Restored Astro 6 HTML-aware whitespace behavior with `compressHTML: true`.
- Reverted the generated OG image churn from Astro 7's JSX-style whitespace default; the final PR diff has no net OG image changes.
- Cleaned up the Clerk patch representation so `git diff --check origin/main` passes and refreshed the pnpm patch hash.

Checked:
- Scanned all 69 Astro files for line-break-sensitive inline text/tag boundaries; the issue affects more than the two examples, including the OG image template, so the config-level fix is intentional.
- `pnpm lint-fix`
- `pnpm tsc`
- `OG_IMAGE_BASE_URL=http://localhost:45431 pnpm -F app run og-image-generate`
- `git diff --check origin/main`
- Native Codex + Cursor pre-commit review, then re-review after the patch whitespace cleanup

Review thread reply: https://github.com/ariakit/ariakit/pull/6412#discussion_r3461529473

app/astro.config.ts
```diff
@@ -45,6 +45,10 @@ export default defineConfig({
 
   srcDir: "src",
 
+  // Astro 7 defaults to JSX-style whitespace. Keep the app on Astro 6's
+  // HTML-aware compression so inline content split across lines keeps spaces.
+  compressHTML: true,
+
```
diegohaz: I think we should migrate the usage instead of keeping the Astro 6 behavior.

renovate.json
```diff
@@ -26,7 +26,7 @@
     {
       "matchFileNames": ["app/package.json"],
       "matchPackageNames": ["vite"],
-      "allowedVersions": "/^7\\./"
+      "allowedVersions": "/^8\\./"
```
diegohaz, line 29: Do we have any hack in the codebase because of mismatched Vite versions that we could remove with this update?

app/astro.config.ts
diegohaz: Fixed in f05555eed2ae800ee49f494be575310051465ff0 - removed the Astro 6 `compressHTML` override and migrated the whitespace-sensitive Astro templates to explicit spaces instead. I also regenerated OG images against the current branch and they all stayed within tolerance, so the PR still has no OG image diffs.

renovate.json
diegohaz: Fixed in f05555eed2ae800ee49f494be575310051465ff0 - audited the Vite-related workarounds. The app SSR optimizer workaround is still needed on Astro 7/Vite 8; removing it still reproduces the cold-start React hook failure, so I kept it and updated the TODO to https://github.com/withastro/astro/issues/17166. I also checked the `vitest.config.ts` Vite cast: Vite majors now match, but removing the cast still fails because pnpm resolves distinct Vite 8 peer-context type instances, so the TODO now reflects that remaining blocker.

diegohaz:
**Fixed**

- f05555eed2ae800ee49f494be575310051465ff0 ([reply](https://github.com/ariakit/ariakit/pull/6412#discussion_r3461837385)): Removed the `compressHTML` override and migrated the whitespace-sensitive Astro templates to explicit spaces while keeping OG images unchanged.
- f05555eed2ae800ee49f494be575310051465ff0 ([reply](https://github.com/ariakit/ariakit/pull/6412#discussion_r3461837519)): Audited the Vite workarounds, kept the SSR optimizer workaround because it still reproduces on Astro 7/Vite 8, updated it to the active Astro issue, and refreshed the Vitest cast TODO for the remaining peer-context type blocker.

renovate.json
diegohaz: I mean on vitest.config.ts line 69

diegohaz: Fixed in ec8cc8fe0f3304cf4788147ad0f51d0cbd923221 - removed the `vitest.config.ts` Vite plugin cast by changing `sourcePlugin` to expose a small structural plugin type instead of the app workspace's concrete Vite `Plugin` type. The root Vitest config can now use `sourcePlugin(...)` directly, while still typing the plugin array as the root `PluginOption[]`.

diegohaz:
**Fixed**

- ec8cc8fe0f3304cf4788147ad0f51d0cbd923221 ([reply](https://github.com/ariakit/ariakit/pull/6412#discussion_r3462043651)): Removed the `vitest.config.ts` Vite plugin cast by making `sourcePlugin` expose a structural plugin type that works from both the app Astro config and the root Vitest config.

diegohaz:
Depends on https://github.com/clerk/javascript/issues/8963

PR #6413, opened 2026-06-23
Split Tailwind sandbox behavior checks

diegohaz:
## Motivation

The `ariakit-tailwind` sandbox currently mixes the broad computed-style snapshot matrix with a couple of behavior-specific checks. That makes the main sandbox larger than it needs to be for snapshot and perf coverage, and it hides focused regressions inside an omnibus fixture.

## Solution

Move the two behavior-specific checks into focused sandboxes:

- `ariakit-tailwind-frame-cover-rtl` covers RTL `ak-frame-cover` row geometry.
- `ariakit-tailwind-layer-band-boundaries` covers exact layer-band boundary variant matching.

The main `ariakit-tailwind` sandbox now keeps the WCAG and computed-style snapshot matrix, with snapshots updated to drop the moved band-boundary subtree. The app global Tailwind source exclusions also include the new focused sandboxes so their test-only `@ariakit/tailwind` utilities stay isolated to their local `style.css` files.

## Testing

- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app test-chrome ariakit-tailwind`
- `pnpm lint-fix`
- `pnpm tsc`
- `git diff --check`

### DIFF
--- a/app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
+++ b/app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
@@ -1,0 +1,31 @@
+import "./style.css";
+
+export default function Example() {
+  return (
+    <div className="flex flex-col gap-4 p-4">
+      <section
+        aria-label="band boundary variants"
+        className="ak-layer ak-frame ak-frame-p-1 ak-frame-border flex-col"
+      >
+        <div className="grid gap-2">
+          <section className="ak-layer ak-layer-l-27.5 p-2">
+            <div className="ak-dark-high:bg-red-950 p-2 text-white">
+              Dark high candidate (layer lightness 0.275)
+            </div>
+            <div className="ak-dark-low:bg-black p-2 text-white">
+              Dark low candidate (layer lightness 0.275)
+            </div>
+          </section>
+          <section className="ak-layer ak-layer-l-85.75 p-2">
+            <div className="ak-light-low:bg-red-950 p-2 text-black">
+              Light low candidate (layer lightness 0.8575)
+            </div>
+            <div className="ak-light-high:bg-black p-2 text-white">
+              Light high candidate (layer lightness 0.8575)
+            </div>
+          </section>
+        </div>
+      </section>
+    </div>
+  );
+}
--- a/app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
+++ b/app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
@@ -5,7 +5,7 @@ export default function Example() {
     <div className="flex flex-col gap-4 p-4">
       <section
         aria-label="band boundary variants"
-        className="ak-layer ak-frame ak-frame-p-1 ak-frame-border flex-col"
+        className="ak-layer ak-frame ak-frame-p-1 ak-frame-border"
       >
         <div className="grid gap-2">
           <section className="ak-layer ak-layer-l-27.5 p-2">
--- a/app/src/styles/global.css
+++ b/app/src/styles/global.css
@@ -10,9 +10,7 @@
 @import "./ariakit.css";
 
 @source not "./styles.json";
-@source not "../sandbox/ariakit-tailwind";
-@source not "../sandbox/ariakit-tailwind-frame-cover-rtl";
-@source not "../sandbox/ariakit-tailwind-layer-band-boundaries";
+@source not "../sandbox/ariakit-tailwind*";
 
 @theme {
   --color-canvas: oklch(99.33% 0.0011 197.14);
--- a/app/src/styles/global.css
+++ b/app/src/styles/global.css
@@ -11,6 +11,8 @@
 
 @source not "./styles.json";
 @source not "../sandbox/ariakit-tailwind";
+@source not "../sandbox/ariakit-tailwind-frame-cover-rtl";
+@source not "../sandbox/ariakit-tailwind-layer-band-boundaries";
 
 @theme {
   --color-canvas: oklch(99.33% 0.0011 197.14);
--- a/app/src/sandbox/ariakit-tailwind-frame-cover-rtl/index.react.tsx
+++ b/app/src/sandbox/ariakit-tailwind-frame-cover-rtl/index.react.tsx
@@ -1,0 +1,52 @@
+import "./style.css";
+
+interface FrameCoverCardProps {
+  coverPosition: "start" | "end";
+  dir: "ltr" | "rtl";
+  label: string;
+}
+
+function FrameCover() {
+  return <div className="ak-frame-cover ak-layer-blue-600 w-24 shrink-0" />;
+}
+
+function FrameCoverCard({ coverPosition, dir, label }: FrameCoverCardProps) {
+  return (
+    <section
+      aria-label={label}
+      dir={dir}
+      className="ak-layer ak-frame ak-frame-xl/2 ak-frame-border ak-frame-row flex h-16 w-80"
+    >
+      {coverPosition === "start" && <FrameCover />}
+      <div className="grow" />
+      {coverPosition === "end" && <FrameCover />}
+    </section>
+  );
+}
+
+export default function Example() {
+  return (
+    <div className="flex flex-col gap-4 p-4">
+      <FrameCoverCard
+        coverPosition="start"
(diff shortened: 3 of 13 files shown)


app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
```diff
@@ -0,0 +1,31 @@
+import "./style.css";
+
+export default function Example() {
+  return (
+    <div className="flex flex-col gap-4 p-4">
+      <section
+        aria-label="band boundary variants"
+        className="ak-layer ak-frame ak-frame-p-1 ak-frame-border flex-col"
```
diegohaz: When this block was a `Layer`, the helper injected `flex font-mono gap-(--ak-frame-padding)`, so `flex-col` here resolved against a real column flex container. As a plain `<section>` it no longer gets `flex`, and Tailwind's `flex-col` only sets `flex-direction` (not `display: flex`) — so this class is now inert. The rest of the repo always pairs them (`flex flex-col`).

Two small consequences of the `Layer` → plain-markup conversion, neither of which affects the assertions (band resolution is driven by the absolute `ak-layer-l-*` lightness, independent of layout):

- This outer section's `flex-col` is dead — it has a single child (`grid gap-2`), so removing it changes nothing.
- The inner `ak-layer-l-27.5` / `ak-layer-l-85.75` sections (lines 11 and 19) were also `Layer`s, so their candidate pairs used to lay out in a `flex` row; they now stack vertically.

If you want the sandbox to read like the original demo, add `flex` back (and to the two inner sections for the side-by-side pairs). Otherwise just drop the inert class:

```suggestion
        className="ak-layer ak-frame ak-frame-p-1 ak-frame-border"
```

diegohaz (commented):
**Initial review** — diff from merge-base `5ccbfb732` to head `e454829af`, then expanded to the surrounding sandbox/test/preview infrastructure.

### Scope reviewed
- Both new sandboxes (`ariakit-tailwind-frame-cover-rtl`, `ariakit-tailwind-layer-band-boundaries`): `index.react.tsx`, `style.css`, `test-chrome.ts`.
- The trimmed `ariakit-tailwind` sandbox: `index.react.tsx`, `test-chrome.ts`, and all four `__snapshots__` files.
- `app/src/styles/global.css` source exclusions.
- Surrounding context: `app/src/test-utils/preview.ts` (`withFramework`/`gotoAndSettle`), `app/playwright.config.ts` test discovery, the omnibus `perf-chrome.ts`, and a repo-wide search for references to the moved code.

### Verification performed
- Confirmed the two new `style.css` files are byte-identical to the original `ariakit-tailwind/style.css`, so the radius/spacing theme used by the moved assertions is preserved.
- Confirmed the snapshot edits are surgical: only the `band boundary variants` subtree (and its two `ak-layer-l-*` children) is removed from each of the four snapshots; nothing else shifted. The frame-cover cards used `aria-label` (not `aria-labelledby`) so they were never in the snapshot, consistent with the diff.
- Confirmed no dangling references remain after the removals (`FrameCover*`, `getFrameCoverMetrics`, `FrameCoverMetrics`, the `Locator` import) and that all remaining imports in the omnibus files are still used.
- Confirmed `withFramework`'s `beforeEach` already runs `gotoAndSettle` (waits for `load` + bounded network-idle), so dropping `waitForPreviewReady` from the two moved tests is safe — they rely on auto-retrying `expect`/`toBeVisible`/`toHaveCSS`, while the synchronous snapshot test (which genuinely needs the rAF-stability wait) keeps it.
- Confirmed the band-boundary assertions are layout-independent: the candidate `background-color` is driven by the absolute `ak-layer-l-27.5` / `ak-layer-l-85.75` lightness on the nearest layer, which is unchanged by the `Layer` → plain-markup conversion.
- Confirmed Playwright auto-discovers the new `test-chrome.ts` files via the `\/test[^/]*-chrome` matcher (no CI/config registration needed), and that the only external references to the new sandboxes are the two correct `@source not` lines in `global.css`.
- Noted the omnibus `perf-chrome.ts` toggles `section[aria-labelledby]`; the removal drops 3 such sections — a negligible, budget-safe workload change (frame-cover cards used `aria-label` and were never selected).

### Findings
One low-severity, non-blocking nit (inline): the band-boundary outer `<section>` keeps `flex-col` but no longer receives the `flex` that `Layer` used to inject, so the class is inert; the inner `ak-layer-l-*` sections also lost their implicit `flex` row layout, so the candidate pairs now stack vertically instead of side-by-side. Neither affects the assertions — it's purely cosmetic relative to the original demo.

No changeset is needed (sandbox/test-only change), so the changeset-bot warning is expected. CodeRabbit reported no actionable comments.

### Not fully validated
I did not run the browser suite locally (the branch is checked out in another worktree, and the app-lite build/preview is heavyweight here). The author reported `pnpm -F app test-chrome ariakit-tailwind`, which is a path-substring filter that covers all three sandboxes, plus `pnpm lint-fix`, `pnpm tsc`, and `git diff --check`. Static analysis supports behavior preservation; the inline nit is the only thing worth a second look.

app/src/styles/global.css
```diff
@@ -11,6 +11,8 @@
 
 @source not "./styles.json";
 @source not "../sandbox/ariakit-tailwind";
+@source not "../sandbox/ariakit-tailwind-frame-cover-rtl";
+@source not "../sandbox/ariakit-tailwind-layer-band-boundaries";
```
diegohaz: Is it possible to use glob here?

diegohaz: Yes — `@source not` takes a glob, so the three sandbox lines collapse into one:

```css
@source not "../sandbox/ariakit-tailwind*";
```

Tailwind passes the pattern through to `@tailwindcss/oxide` verbatim as a negated source (`{ base: ".../src/styles", pattern: "../sandbox/ariakit-tailwind*", negated: true }`), and oxide matches it against scanned files — no directory auto-expansion needed.

I validated each candidate end-to-end against the installed v4 toolchain (`tailwindcss@4.3.1` + `@tailwindcss/oxide@4.3.1`) with a fixture of all three `ariakit-tailwind*` dirs plus a control sandbox, compiling and checking which utilities leaked into the output:

| Pattern | `ariakit-tailwind` | `…-frame-cover-rtl` | `…-layer-band-boundaries` | other sandbox |
|---|---|---|---|---|
| three explicit lines (current) | excluded | excluded | excluded | kept |
| `ariakit-tailwind*` | excluded | excluded | excluded | kept |
| `ariakit-tailwind*/**` | excluded | excluded | excluded | kept |
| `{ariakit-tailwind,…-frame-cover-rtl,…-layer-band-boundaries}` | excluded | excluded | excluded | kept |
| `ariakit-tailwind-*` (hyphen) | **kept** ⚠️ | excluded | excluded | kept |

Two things to keep in mind:

- Use `ariakit-tailwind*`, not `ariakit-tailwind-*`. The hyphenated form doesn't match the base `ariakit-tailwind` directory (no trailing `-`), so that sandbox would leak back into the global scan.
- `ariakit-tailwind*` also auto-excludes any future `ariakit-tailwind*` sandbox. That's a plus for this isolated-Tailwind-context family (one less line to maintain per new sandbox), but it's now implicit and would also catch an unrelated name like `ariakit-tailwindish`. If you'd rather keep it explicit/exact, brace expansion works too; otherwise the single `ariakit-tailwind*` glob is the cleanest.

app/src/sandbox/ariakit-tailwind-layer-band-boundaries/index.react.tsx
diegohaz: Fixed in b40c588b2 - removed the inert flex-col class from the band-boundary sandbox section.

app/src/styles/global.css
diegohaz: Fixed in b40c588b2 - collapsed the Ariakit Tailwind sandbox source exclusions into a single verified `ariakit-tailwind*` glob.

diegohaz:
**Fixed**
- b40c588b2 ([reply](https://github.com/ariakit/ariakit/pull/6413#discussion_r3462646440)): Removed the inert `flex-col` class from the band-boundary sandbox markup.
- b40c588b2 ([reply](https://github.com/ariakit/ariakit/pull/6413#discussion_r3462646569)): Replaced the explicit Ariakit Tailwind sandbox source exclusions with a verified `ariakit-tailwind*` glob.

PR #6414, opened 2026-06-23
Fix Solid Group aria-label precedence

### LINKED ISSUE #6327: Solid: explicit aria-label on Group is overridden by GroupLabel's aria-labelledby
### Current behavior

In `@ariakit/solid`, passing an explicit `aria-label` to `Group` while rendering a `GroupLabel` child produces a group element with **both** `aria-label` and the internally generated `aria-labelledby`:

```html
<div role="group" aria-labelledby="s0000010010" aria-label="Audio playback settings">…</div>
```

Per the [accessible name computation](https://www.w3.org/TR/accname-1.2/), `aria-labelledby` takes precedence, so the accessible name resolves to the `GroupLabel` text and the explicit `aria-label` is silently ignored. Screen readers announce the visible label instead of the name the developer wrote. Since both attributes render and the page looks identical, the regression is invisible without inspecting the accessibility tree.

This was fixed for React in #5661 (commit [`1748781b3`](https://github.com/ariakit/ariakit/commit/1748781b30422fe1482f91ba968fdc09f50eea45)): the React hook drops its internal `aria-labelledby` whenever `aria-label` is passed ([`packages/ariakit-react-components/src/group/group.tsx#L39-L45`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/group/group.tsx#L39-L45)):

```tsx
props = {
  role: "group",
  "aria-labelledby": props["aria-label"] != null ? undefined : labelId,
  ...props,
};
```

The Solid port still has the pre-fix behavior — it merges a reactive `aria-labelledby` getter unconditionally ([`packages/ariakit-solid-components/src/group/group.tsx#L34-L42`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-components/src/group/group.tsx#L34-L42)):

```tsx
props = mergeProps(
  {
    role: "group" as const,
    get "aria-labelledby"() {
      return labelId();
    },
  },
  props,
);
```

The merge machinery can't compensate: `mergeProps` delegates to `combineProps` from `@solid-primitives/props` ([`packages/ariakit-solid-utils/src/reactivity.ts#L156-L168`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/reactivity.ts#L156-L168)), which resolves each key independently — a user-supplied `aria-label` is a different key and can never suppress the default `aria-labelledby` getter. Even passing `aria-labelledby={undefined}` doesn't help, because `combineProps` skips `undefined` values (verified in the browser).

Verified with a real-browser Playwright test: the same JSX passes in `@ariakit/react` (which links the equivalent [Group](https://ariakit.com/reference/group)/[GroupLabel](https://ariakit.com/reference/group-label) components) and fails in `@ariakit/solid` with accessible name `"Audio"` instead of `"Audio playback settings"`, using user-level interactions only.

There is no Solid StackBlitz template in the repo yet, so starting from a plain Vite + Solid starter:

1. Open https://vite.new/solid-ts (StackBlitz Vite + Solid TS starter).
2. Install `@ariakit/solid` (in the StackBlitz terminal: `npm install @ariakit/solid`).
4. Open the browser DevTools, select the `role="group"` element, and check the Accessibility pane.

```tsx
import * as Ariakit from "@ariakit/solid";

  return (
    <Ariakit.Group aria-label="Audio playback settings">
      <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
      <button type="button">Mute</button>
      <button type="button">Boost</button>
    </Ariakit.Group>
  );
}
```

The group's computed accessible name is `Audio` (from the `GroupLabel` via `aria-labelledby`), not `Audio playback settings`.

### Expected behavior

The explicit `aria-label` should provide the group's accessible name, and the internal `aria-labelledby` should be dropped — matching `@ariakit/react` since #5661, where the same JSX yields `<div role="group" aria-label="Audio playback settings">` with accessible name `Audio playback settings`.

### Workaround

Until this is fixed, don't combine `GroupLabel` with an `aria-label` on the Solid `Group`. Render the visible label as a plain `aria-hidden` element instead (which is what `GroupLabel` renders, minus the id registration that feeds `aria-labelledby`):

```tsx
<Ariakit.Group aria-label="Audio playback settings">
  <div aria-hidden>Audio</div>
  <button type="button">Mute</button>
  <button type="button">Boost</button>
</Ariakit.Group>
```

Alternatively, passing `aria-labelledby=""` to `Group` also restores the `aria-label` as the accessible name (an empty reference list falls back to `aria-label`), but it leaves a stray empty `aria-labelledby` attribute in the DOM. Note that `aria-labelledby={undefined}` does not work — the merge logic skips `undefined` values.

### Possible solutions

Mirror the React check inside the reactive getter in `packages/ariakit-solid-components/src/group/group.tsx`:

```diff
     props = mergeProps(
       {
         role: "group" as const,
         get "aria-labelledby"() {
-          return labelId();
+          return props["aria-label"] != null ? undefined : labelId();
         },
       },
       props,
```

Keeping the check inside the getter keeps it reactive (it re-evaluates when either the `aria-label` prop or the `labelId` signal changes). By the time the getter runs, `props` is the merged proxy, so `props["aria-label"]` resolves to the user's value (the defaults object never defines `aria-label`, and reading a different key avoids recursion). Solid's renderer removes nullish attributes, matching `removeUndefinedValues` in the React hook, and a user-supplied `aria-labelledby` still wins because `combineProps` resolves keys last-source-first while skipping `undefined`. A patch changeset for `@ariakit/solid-components` should accompany the fix, as the React fix did.

diegohaz:
## Motivation

In `@ariakit/solid`, `Group` still rendered the internal `aria-labelledby` from `GroupLabel` when a developer passed an explicit `aria-label`. Because `aria-labelledby` takes precedence in the accessible name computation, the explicit `aria-label` was ignored.

Fixes #6327.

## Solution

Add a Solid regression sandbox under `app/src/sandbox/group-6327` with both browser and happy-dom coverage for the accessible name behavior.

Update Solid `Group` so its internal `aria-labelledby` getter returns `undefined` whenever `aria-label` is present. This mirrors the existing React behavior and keeps the check reactive while still allowing a user-supplied `aria-labelledby` prop to win through the existing `mergeProps` order.

## Workaround

Until the fix is released, avoid combining `GroupLabel` with an explicit `aria-label` on Solid `Group`. Render the visible label as an `aria-hidden` element instead so it does not register an internal `aria-labelledby` value.

```tsx
<Ariakit.Group aria-label="Audio playback settings">
  {/* TODO: Remove after https://github.com/ariakit/ariakit/issues/6327 is fixed. */}
  <div aria-hidden>Audio</div>
  <button type="button">Mute</button>
  <button type="button">Boost</button>
</Ariakit.Group>
```

### DIFF
--- a/.changeset/300-solid-group-aria-label.md
+++ b/.changeset/300-solid-group-aria-label.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/solid-components": patch
+"@ariakit/solid": patch
+---
+
+Fixed [`Group`](https://ariakit.com/reference/group) to ignore its internal [`GroupLabel`](https://ariakit.com/reference/group-label) reference when `aria-label` is passed.
--- a/app/src/sandbox/group-6327/index.solid.tsx
+++ b/app/src/sandbox/group-6327/index.solid.tsx
@@ -3,8 +3,7 @@ import * as Ariakit from "@ariakit/solid";
 export default function Example() {
   return (
     <Ariakit.Group aria-label="Audio playback settings">
-      {/* TODO: Use GroupLabel after https://github.com/ariakit/ariakit/issues/6327 is fixed. */}
-      <div aria-hidden>Audio</div>
+      <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
       <button type="button">Mute</button>
       <button type="button">Boost</button>
     </Ariakit.Group>
--- a/app/src/sandbox/group-6327/index.solid.tsx
+++ b/app/src/sandbox/group-6327/index.solid.tsx
@@ -2,10 +2,19 @@ import * as Ariakit from "@ariakit/solid";
 
 export default function Example() {
   return (
-    <Ariakit.Group aria-label="Audio playback settings">
-      <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
-      <button type="button">Mute</button>
-      <button type="button">Boost</button>
-    </Ariakit.Group>
+    <>
+      <Ariakit.Group aria-label="Audio playback settings">
+        <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
+        <button type="button">Mute</button>
+        <button type="button">Boost</button>
+      </Ariakit.Group>
+      <Ariakit.Group
+        aria-label="Audio playback settings"
+        aria-labelledby="explicit-group-label"
+      >
+        <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
+        <span id="explicit-group-label">Explicit playback settings</span>
+      </Ariakit.Group>
+    </>
   );
 }
--- a/app/src/sandbox/group-6327/index.solid.tsx
+++ b/app/src/sandbox/group-6327/index.solid.tsx
@@ -1,0 +1,11 @@
+import * as Ariakit from "@ariakit/solid";
+
+export default function Example() {
+  return (
+    <Ariakit.Group aria-label="Audio playback settings">
+      <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
+      <button type="button">Mute</button>
+      <button type="button">Boost</button>
+    </Ariakit.Group>
+  );
+}
--- a/app/src/sandbox/group-6327/index.solid.tsx
+++ b/app/src/sandbox/group-6327/index.solid.tsx
@@ -3,7 +3,8 @@ import * as Ariakit from "@ariakit/solid";
 export default function Example() {
   return (
     <Ariakit.Group aria-label="Audio playback settings">
-      <Ariakit.GroupLabel>Audio</Ariakit.GroupLabel>
+      {/* TODO: Use GroupLabel after https://github.com/ariakit/ariakit/issues/6327 is fixed. */}
+      <div aria-hidden>Audio</div>
       <button type="button">Mute</button>
       <button type="button">Boost</button>
     </Ariakit.Group>
--- a/app/src/sandbox/group-6327/test-browser.ts
+++ b/app/src/sandbox/group-6327/test-browser.ts
@@ -1,0 +1,15 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // See https://github.com/ariakit/ariakit/issues/6327
+  test("uses aria-label instead of GroupLabel for the group name", async ({
+    q,
+  }) => {
+    const group = q.group("Audio playback settings");
+    await test.expect(group).toBeVisible();
+    await test
+      .expect(group)
+      .toHaveAttribute("aria-label", "Audio playback settings");
+    await test.expect(group).not.toHaveAttribute("aria-labelledby");
+  });
+});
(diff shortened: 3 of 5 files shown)


diegohaz:
**Fixed**

- `0e52d8e`: added happy-dom and browser coverage for the case where `aria-label` and an explicit user-provided `aria-labelledby` are passed together. The test asserts the explicit `aria-labelledby` is preserved while `aria-label` is still forwarded, since the user-provided prop should remain in the DOM.

diegohaz (commented):
**Review summary — no substantive findings.** This is a clean, minimal fix and I'm comfortable with it as-is.

**Type:** Initial review (no prior review from this workflow; CodeRabbit's earlier review is a separate actor).
**Baseline:** merge-base of `origin/main` and head `0e52d8e0d` (`0bdf5d1cb`).
**Scope:** the one-line `useGroup` change in `packages/ariakit-solid-components/src/group/group.tsx`, the `app/src/sandbox/group-6327` sandbox (`index.solid.tsx`, `test.ts`, `test-browser.ts`), and the changeset — plus surrounding context (React `useGroup`, the Solid `mergeProps`/`combineProps` merge order, `vitest.setup.ts` auto-render, the root vitest loader include logic, `withFramework`, `q.group`, and the existing `aria-labelledby-aria-label` sandbox).

**What I verified locally** (Node 24, after `pnpm install` in this worktree):
- `pnpm test-solid run group-6327` → 2/2 pass.
- Reverted the fix to the old `return labelId();` and re-ran: test 1 fails (`q.group.ensure("Audio playback settings")` can't resolve the group because the accessible name falls back to the `GroupLabel` text), confirming it's a genuine regression test. Restored the fix afterward — working tree is clean.
- `pnpm tsc` → 0 errors.
- Confirmed the fix matches the shipped React behavior (`#5661`) exactly, including parity on the `!= null` check (so an empty-string `aria-label` suppresses `aria-labelledby` in both, which is the intended React parity).

**Things I checked and found correct:**
- *Reactivity / no recursion:* keeping the check inside the getter re-evaluates on `aria-label`/`labelId` changes; reading the different `aria-label` key off the merged proxy avoids self-recursion. A user-supplied `aria-labelledby` still wins via `combineProps` last-source-first (test 2 locks this in — note it passes with or without the fix, which is correct for a guard).
- *Changeset:* `300-` prefix matches the current convention; bumping `@ariakit/solid-components` + `@ariakit/solid` mirrors the React `react-core` + `react` pair (`@ariakit/solid` re-exports `Group` via `group.ts`); both reference links resolve (HTTP 200).
- *Sandbox structure:* `group-6327` follows the issue-repro naming convention; no `preview.json` needed since the framework is inferred from `index.solid.tsx`; the generic `test.ts` is collected only by the Solid suite (dir has only a Solid entry), and `vitest.setup.ts` auto-renders the preview so no explicit `render()` is required. The two groups have distinct accessible names ("Audio playback settings" vs. "Explicit playback settings"), so the role queries are unambiguous.
- A dedicated Solid sandbox is the right call rather than adding an `index.solid.tsx` to the existing React-only `aria-labelledby-aria-label` sandbox: Solid only ships `Group` (not Dialog/Tab/Select/Form/Tooltip/Tag), so a shared entry would break that sandbox's other component tests under the Solid framework run.
- CodeRabbit's only nitpick (add an explicit-`aria-labelledby` precedence test) was already addressed in `0e52d8e`; no open threads remain.

**Not validated locally:** the served-browser Playwright tests (`test-browser.ts`) — I validated the equivalent happy-dom assertions and the underlying merge/render logic instead of standing up the worktree dev server. CI's browser run covers them.

PR #6415, opened 2026-06-23
Fix Solid zero-argument hooks

### LINKED ISSUE #6304: Solid: documented zero-argument hook calls (useVisuallyHidden(), useFocusTrapRegion(), ...) throw TypeError at runtime
### Current behavior

Every component hook in `@ariakit/solid-components` documents a zero-argument call in its JSDoc — for example [`useVisuallyHidden`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-components/src/visually-hidden/visually-hidden.tsx#L13-L19) and [`useFocusTrapRegion`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-components/src/focus-trap/focus-trap-region.tsx#L19-L28) show `const props = useVisuallyHidden(); <Role {...props} />`. The [`Hook` type](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/types.ts#L80-L85) declares `props?` optional, so the call type-checks cleanly. At runtime it throws a `TypeError` and the component never renders.

The root cause is that the Solid port of [`createHook`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/system.tsx#L72-L77) is a bare cast — `return useProps as Hook<T, P>;` — while the React version it was ported from [defaults `props` to `{}`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-utils/src/system.tsx#L75-L84) before delegating. With `props` left `undefined`, all hook implementations break:

- `withOptions` hooks (`useFocusTrapRegion()`, `useSeparator()`): [`withOptions`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/system.tsx#L113-L116) passes `undefined` into [`extractPropsWithDefaults`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/reactivity.ts#L63-L72) → solid-js `splitProps(undefined)` throws immediately: `TypeError: Cannot use 'in' operator to search for 'Symbol(solid-proxy)' in undefined` (server build: `Cannot convert undefined or null to object`).
- Hooks built on the custom [`mergeProps`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/reactivity.ts#L159-L168) (`useVisuallyHidden()`, `useFocusTrap()`, `useGroup()`, `useGroupLabel()`, `useHeading()`): `combineProps` accepts the `undefined` source, but the first property access on the returned props — i.e. the moment they are spread onto an element — throws `TypeError: Cannot read properties of undefined (reading 'style')`.
- [`useRole()`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-components/src/role/role.tsx#L51-L55) is the identity hook, so it returns `undefined` even though its type says it returns a props object — any property access on the typed result throws.

So the hooks are documented, typed, and guaranteed to crash when called the documented way. Verified with a real-browser Playwright test: a Solid component using the `useVisuallyHidden` JSDoc example verbatim crashes during render and the page shows nothing.

1. Create a Solid app: `npm create vite@latest ariakit-solid-repro -- --template solid-ts`
2. `cd ariakit-solid-repro && npm install && npm install @ariakit/solid @ariakit/solid-components`
4. Run `npm run dev` and open the app in the browser.
5. The page is blank: instead of rendering the link, the app throws `TypeError: Cannot read properties of undefined (reading 'style')` in the console.

```tsx
import { Role } from "@ariakit/solid";
import { useVisuallyHidden } from "@ariakit/solid-components/visually-hidden/visually-hidden";

  // Documented usage from the useVisuallyHidden JSDoc example:
  //   const props = useVisuallyHidden();
  //   <Role {...props} />
  const props = useVisuallyHidden();
  return (
    <a href="#">
      Learn more<Role {...props}> about the Solar System</Role>.
    </a>
  );
}
```

The same happens with `useFocusTrapRegion()` or `useSeparator()`, which throw at the hook call itself.

### Expected behavior

Calling a component hook with no arguments should return a usable props object, exactly as documented in the JSDoc examples and as the optional `props?` in the `Hook` type promises — and as `@ariakit/react` already guarantees by defaulting `props` to `{}` in `createHook`. The example above should render the link with the visually hidden text.

### Workaround

Pass an empty object explicitly:

```tsx
const props = useVisuallyHidden({});
```

Verified: with `{}` every affected hook works (`useVisuallyHidden({})`, `useFocusTrapRegion({})`, `useRole({})`), and the Playwright repro passes.

### Possible solutions

Mirror the React implementation in the Solid `createHook`, which makes the optional-props contract the `Hook` type already declares true at runtime:

```diff
--- a/packages/ariakit-solid-utils/src/system.tsx
+++ b/packages/ariakit-solid-utils/src/system.tsx
@@ -72,7 +72,9 @@ export function createHook<
   T extends ValidComponent,
   P extends AnyObject = EmptyObject,
 >(useProps: (props: Props<T, P>) => HTMLProps<T, P>) {
-  return useProps as Hook<T, P>;
+  const useRole = (props: Props<T, P> = {} as Props<T, P>) =>
+    useProps(props);
+  return useRole as Hook<T, P>;
 }
```

All crash paths resolve with the default: `splitProps({}, keys)` and `combineProps` handle a plain `{}`, and `useRole()` returns `{}` instead of `undefined`. Reactivity is preserved because a provided props object is forwarded untouched (no property access outside a tracked scope), and the component-wrapper path (`<VisuallyHidden />` etc.) is unaffected since it always passes a real props object. React's `displayName` line is DevTools-specific and correctly stays out of the Solid version. A regression test calling each exported hook with no arguments inside `createRoot` and spreading/reading the result would cover all crash mechanisms at once.

diegohaz:
## Motivation

Solid component hooks are typed and documented as accepting no props, but calling hooks such as `useVisuallyHidden()`, `useFocusTrapRegion()`, and `useRole()` without an explicit object crashed or returned unusable props at runtime.

The `Hook` type already declares `props?` as optional, so the runtime behavior should match the documented zero-argument examples.

Fixes #6304.

## Solution

Updated Solid `createHook` to mirror the React helper by defaulting missing props to `{}` before delegating to the hook implementation. This keeps provided props forwarded as-is while making zero-argument calls safe for hooks that merge props, split options, or return props directly.

The repro sandbox now uses the documented zero-argument calls again and covers the behavior in both the browser and happy-dom.

## Workaround

Until the library fix is available in a release, pass an empty object to Solid component hooks instead of calling them with no arguments:

```tsx
// TODO: Remove explicit props once the fix for
// https://github.com/ariakit/ariakit/issues/6304 lands.
const hiddenProps = useVisuallyHidden({});
const focusTrapRegionProps = useFocusTrapRegion({});
const roleProps = useRole({});
```


.changeset/300-solid-zero-argument-hooks.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/solid-utils": patch
+"@ariakit/solid-components": patch
+---
+
+Fixed [`VisuallyHidden`](https://ariakit.com/reference/visually-hidden) and [`FocusTrapRegion`](https://ariakit.com/reference/focus-trap-region) hooks to return usable props when called without arguments, matching their optional props type.
```
diegohaz: **Changeset describes only 2 of the 8 hooks the fix actually repairs.**

The fix lives in the shared `createHook` (`packages/ariakit-solid-utils/src/system.tsx`), so it repairs the zero-argument call for *every* Solid component hook built on it — all eight: `useVisuallyHidden`, `useFocusTrap`, `useFocusTrapRegion`, `useGroup`, `useGroupLabel`, `useHeading`, `useRole`, and `useSeparator`. This changeset names only `VisuallyHidden` and `FocusTrapRegion` and phrases it as if only those two are fixed, so consumers of the other six hooks won't realize their documented zero-argument calls now work too.

Per the changeset guidance, when a base-level change propagates to derived APIs, phrase the entry so readers see the propagation. For a single-paragraph changeset it's fine to keep up to two representative names — just make it clear the fix covers the rest:

```suggestion
Fixed Solid component hooks such as [`VisuallyHidden`](https://ariakit.com/reference/visually-hidden) and [`FocusTrapRegion`](https://ariakit.com/reference/focus-trap-region) to return usable props when called without arguments, matching their optional props type.
```

The "such as" phrasing also subsumes the earlier suggestion to add `useRole` explicitly: it covers `useRole` and the remaining hooks without growing the list past the two-name convention.

diegohaz (commented):
## Review summary

**Scope:** Initial review of the full PR. Baseline: merge-base `5ccbfb732` → head `cff400c1`. Reviewed the core fix (`packages/ariakit-solid-utils/src/system.tsx`), the new sandbox + tests, `app/package.json`, and the changeset, plus surrounding context (the `Hook` type, the React `createHook` it mirrors, all 8 `createHook` consumers in `@ariakit/solid-components`, package exports, and the preview/test-collection infrastructure).

**Verdict:** The fix is correct and minimal. It faithfully mirrors React's `createHook` (defaulting missing props to `{}`), and intentionally omits React's DevTools-only `displayName` line, which is correct for Solid. Reactivity is preserved because a provided props object is forwarded untouched.

**Validation performed (local):**
- `pnpm test-solid run visually-hidden-focus-trap-role-6304` → passes (1/1).
- Reverted the fix to the old bare-cast `return useProps as Hook<T, P>` and re-ran → the test **fails** at `useFocusTrapRegion()` (`splitProps(undefined)` throws), then restored the fix. Confirms it's a genuine regression test covering the crash path.
- Verified the test infra: `vitest.setup.ts` auto-renders `index.solid.tsx` in a `beforeEach` for the solid suite, so render-time throws surface as test failures; the happy-dom `test.ts` runs only under `test-solid`, and `test-browser.ts` runs through `withFramework` for the inferred `solid` framework (real coverage, not silently skipped).
- Confirmed `@ariakit/solid` re-exports only the components (`VisuallyHidden`, `Role`, …), **not** the hooks — so omitting `@ariakit/solid` from the changeset frontmatter is correct; `@ariakit/solid-utils` + `@ariakit/solid-components` are the right packages.
- Verified the new `@ariakit/solid-components` dependency in `app/package.json` matches the sandbox's sub-path imports and the lockfile update.

**Findings (1, non-blocking):**
- The changeset names only 2 of the 8 hooks the shared `createHook` fix actually repairs, and is phrased as if only those two are fixed. Suggested a rephrase (inline) that keeps two representative names but conveys the fix covers all Solid component hooks; it also subsumes the earlier "add useRole" suggestion.

**Not flagged after review:** test assertion strength (the focus-trap-region/role assertions are crash-only, but that matches the bug — a render-time `TypeError` — and `useVisuallyHidden` additionally asserts `position: absolute`, so coverage is adequate); only 3 of 8 hooks are exercised, but they cover all three distinct crash mechanisms from #6304 (`withOptions`/`splitProps`, custom `mergeProps`, identity hook); sandbox naming, Solid-only entry, and omitted `preview.json` all follow conventions.

**Not independently re-run:** the Playwright browser test across all engines (CI is fully green, including Test Solid and all browser jobs).

PR #6417, opened 2026-06-23
Fix Solid createRef reset with function values

### LINKED ISSUE #6349: Solid: createRef(fn).reset() invokes a function-typed initial value instead of restoring it
### Current behavior

`createRef` from `@ariakit/solid-utils` returns a ref store whose `reset()` method is documented to ["reset the ref to the initial value that was passed"](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/reactivity.ts#L120-L124). The implementation passes `initialValue` directly to the Solid signal setter:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-solid-utils/src/reactivity.ts#L142-L154

Solid signal setters treat function arguments as updater callbacks, not values (in `solid-js@1.9.13`'s `createSignal` setter: `if (typeof value === "function") { ... value = value(s.value); }`). So for any ref initialized with a function — `createRef<() => void>(handler)` to hold an event handler, cleanup callback, or focus strategy — `reset()` does not restore `handler`. It *calls* `handler(previousValue)`, running its side effects at an unexpected time, and stores whatever it returns (typically `undefined`) in the ref. The typed overloads accept function values without complaint, and the implementation signature's `initialValue?: any` means TypeScript cannot flag this either.

In the repro below, clicking "Restore default" makes focus jump to the Name field (the initial handler is invoked during `reset()`), the ref silently becomes `undefined`, and the next "Focus field" click throws `TypeError: focusHandler.current is not a function`. Verified with a real-browser Playwright test using only user-level interactions.

1. Create a Solid app: `npm create vite@latest ariakit-solid-repro -- --template solid-ts`
2. `cd ariakit-solid-repro && npm install && npm install @ariakit/solid-utils`
4. Run `npm run dev` and open the app in the browser.
5. Click "Use email handler", then "Focus field" — the Email field is focused, as expected.
6. Click "Restore default" — focus unexpectedly jumps to the Name field: `reset()` invoked the initial handler instead of restoring it, and the output now shows `undefined` instead of `function`.
7. Click "Focus field" — nothing is focused, and the console shows `TypeError: focusHandler.current is not a function`.

```tsx
import { createRef } from "@ariakit/solid-utils";

  let nameInput: HTMLInputElement | undefined;
  let emailInput: HTMLInputElement | undefined;

  function focusNameField() {
    nameInput?.focus();
  }

  function focusEmailField() {
    emailInput?.focus();
  }

  // Keeps the "focus field" handler in a ref so it can be swapped at runtime
  // and restored to the default handler with reset().
  const focusHandler = createRef(focusNameField);

  return (
    <div>
      <input aria-label="Name" ref={nameInput} />
      <input aria-label="Email" ref={emailInput} />
      <button type="button" onClick={() => focusHandler.current()}>
        Focus field
      </button>
      <button
        type="button"
        onClick={() => focusHandler.set(() => focusEmailField)}
      >
        Use email handler
      </button>
      <button type="button" onClick={() => focusHandler.reset()}>
        Restore default
      </button>
      <output>{typeof focusHandler.get()}</output>
    </div>
  );
}
```

### Expected behavior

`reset()` should restore the initial value verbatim, as documented, without invoking it. In the repro: clicking "Restore default" should not move focus, the output should keep showing `function`, and the following "Focus field" click should focus the Name field again.

Note that the `set` method does not have this problem, because Solid's `Setter` type forces the caller to use the thunk idiom (`focusHandler.set(() => focusEmailField)` in the repro). `reset()` gives the consumer no opportunity to do that — the unwrapped `set(initialValue)` call happens inside the library.

### Workaround

Avoid `reset()` for function-typed refs and restore the initial value explicitly through the setter, using the standard Solid thunk idiom:

```tsx
// instead of focusHandler.reset()
focusHandler.set(() => focusNameField);
```

Verified: with this change the Playwright repro passes. Alternatively, store the function inside a wrapper object (`createRef({ handler: focusNameField })`) so the signal value is never itself a function.

### Possible solutions

Use the canonical Solid idiom for storing a value that might be a function — wrap it in a thunk so the setter's updater branch unwraps it verbatim. For non-function values the behavior is unchanged (the updater returns `initialValue`, which goes through the same `===` equality check), and the thunk satisfies `Setter<T>` with no API change:

```diff
--- a/packages/ariakit-solid-utils/src/reactivity.ts
+++ b/packages/ariakit-solid-utils/src/reactivity.ts
@@ -149,7 +149,9 @@ export function createRef<T>(initialValue?: any): RefStore<T> {
     },
     get,
     set,
-    reset: () => set(initialValue),
+    // Wrap in an updater so that function-typed initial values are stored
+    // verbatim instead of being invoked by the setter.
+    reset: () => set(() => initialValue),
   };
 }
```

One implementation note: the `initialValue?: any` implementation signature should stay as is — retyping it to `initialValue?: T` would make the `createSignal<T>(initialValue)` call above fail to compile (`T | undefined` is not assignable to the `createSignal<T>(value: T)` overload). The two public overloads already provide correct external typing. A minimal regression test: `const ref = createRef(noop); ref.set(() => other); ref.reset();` then assert `ref.current === noop` and that `noop` was never called.

diegohaz:
## Motivation

`createRef(fn).reset()` in `@ariakit/solid-utils` should restore the initial function value without invoking it. Before this change, Solid treated the function passed to the signal setter as an updater, so reset could run the initial handler, move focus unexpectedly, and clear the stored value.

Fixes #6349.

## Solution

`createRef().reset()` now uses Solid's setter thunk form so the initial value is stored verbatim, including when that value is a function:

```ts
reset: () => set(() => initialValue)
```

The Solid sandbox under `app/src/sandbox/create-ref-6349/` exercises the original user-facing flow with `focusHandler.reset()`: reset must not focus the Name field as a side effect, the stored value must remain a function, and the next Focus field action must focus the Name field again.

This also adds a focused `@ariakit/solid-utils` regression test that verifies a function initial value is restored without being invoked.

## Workaround

Avoid `reset()` for function-typed refs and restore the initial value explicitly through the setter with Solid's thunk idiom:

```tsx
<button
  type="button"
  onClick={() => {
    // TODO: Remove after https://github.com/ariakit/ariakit/issues/6349
    focusHandler.set(() => focusNameField);
  }}
>
  Restore default
</button>
```

### DIFF
--- a/.changeset/300-solid-create-ref-reset.md
+++ b/.changeset/300-solid-create-ref-reset.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/solid-utils": patch
+---
+
+Fixed `createRef().reset()` so function initial values are restored without being invoked.
--- a/app/package.json
+++ b/app/package.json
@@ -53,6 +53,7 @@
     "@ariakit/react-components": "workspace:*",
     "@ariakit/react-utils": "workspace:*",
     "@ariakit/solid": "workspace:*",
+    "@ariakit/solid-utils": "workspace:*",
     "@ariakit/tailwind": "workspace:*",
     "@ariakit/utils": "workspace:*",
     "@astrojs/check": "0.9.9",
--- a/app/src/sandbox/create-ref-6349/index.solid.tsx
+++ b/app/src/sandbox/create-ref-6349/index.solid.tsx
@@ -53,10 +53,7 @@ export default function Example() {
         <button
           type="button"
           class="rounded border border-gray-300 px-3 py-1"
-          onClick={() => {
-            // TODO: Remove after https://github.com/ariakit/ariakit/issues/6349
-            focusHandler.set(() => focusNameField);
-          }}
+          onClick={() => focusHandler.reset()}
         >
           Restore default
         </button>
--- a/app/src/sandbox/create-ref-6349/index.solid.tsx
+++ b/app/src/sandbox/create-ref-6349/index.solid.tsx
@@ -1,0 +1,64 @@
+import { createRef } from "@ariakit/solid-utils";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6349
+export default function Example() {
+  let nameInput: HTMLInputElement | undefined;
+  let emailInput: HTMLInputElement | undefined;
+
+  const focusNameField = () => {
+    nameInput?.focus();
+  };
+
+  const focusEmailField = () => {
+    emailInput?.focus();
+  };
+
+  const focusHandler = createRef(focusNameField);
+
+  return (
+    <div class="flex flex-col items-start gap-3">
+      <label class="flex flex-col gap-1">
+        Name
+        <input
+          ref={(element) => {
+            nameInput = element;
+          }}
+          class="rounded border border-gray-300 px-3 py-1"
+        />
+      </label>
+      <label class="flex flex-col gap-1">
+        Email
+        <input
+          ref={(element) => {
+            emailInput = element;
+          }}
+          class="rounded border border-gray-300 px-3 py-1"
+        />
+      </label>
+      <div class="flex gap-2">
+        <button
+          type="button"
+          class="rounded border border-gray-300 px-3 py-1"
+          onClick={() => focusHandler.current()}
+        >
+          Focus field
+        </button>
+        <button
+          type="button"
+          class="rounded border border-gray-300 px-3 py-1"
+          onClick={() => focusHandler.set(() => focusEmailField)}
+        >
+          Use email handler
+        </button>
+        <button
+          type="button"
+          class="rounded border border-gray-300 px-3 py-1"
+          onClick={() => focusHandler.reset()}
+        >
+          Restore default
+        </button>
+      </div>
+      <output>{"Stored value type: " + typeof focusHandler.get()}</output>
+    </div>
+  );
+}
--- a/app/src/sandbox/create-ref-6349/index.solid.tsx
+++ b/app/src/sandbox/create-ref-6349/index.solid.tsx
@@ -53,7 +53,10 @@ export default function Example() {
         <button
           type="button"
           class="rounded border border-gray-300 px-3 py-1"
-          onClick={() => focusHandler.reset()}
+          onClick={() => {
+            // TODO: Remove after https://github.com/ariakit/ariakit/issues/6349
+            focusHandler.set(() => focusNameField);
+          }}
         >
           Restore default
         </button>
--- a/app/src/sandbox/create-ref-6349/test-browser.ts
+++ b/app/src/sandbox/create-ref-6349/test-browser.ts
@@ -1,0 +1,19 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6349
+withFramework(import.meta.dirname, async ({ test }) => {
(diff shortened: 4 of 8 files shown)


diegohaz (commented):
**Review summary — no blocking findings.**

This is an initial review. Baseline diff: merge-base `5ccbfb732` → head `deac91f88` (all 8 changed files: core fix, unit test, sandbox + happy-dom/browser tests, changeset, `app/package.json`, lockfile).

### Verdict
The fix is correct, minimal, and idiomatic. `reset: () => set(() => initialValue)` is the canonical Solid thunk idiom for storing a value that may itself be a function. I confirmed against the installed Solid setter (`solid-js/dist/dev.js:225-230`) that for a function argument the setter invokes it as an updater (`value(s.value)`), whereas the thunk form makes the updater *return* `initialValue` verbatim and then writes it through the same `writeSignal` equality path — so non-function refs (including the default `undefined`) behave identically to before, and function refs are now restored instead of invoked. This matches the approach proposed in #6349, and I couldn't find a better alternative (a wrapper-object signal would be more invasive; retyping `initialValue?: any` would break the `createSignal<T>` call, as the issue notes).

### Validation performed
- `pnpm test-solid run reactivity create-ref-6349` → 2 files / 2 tests pass.
- Regression-proofed the tests: temporarily reverted the one-line fix and re-ran — **both** the unit test (`expect(calls).toBe(0)`) and the sandbox happy-dom test (`q.textbox("Name")).not.toHaveFocus()`) fail against the buggy code, so they genuinely lock in the fix. Restored the line afterward; working tree is clean.
- `pnpm tsc` → passes (confirms the thunk satisfies `Setter<T>` and `initialValue?: any` is still required).
- Verified no existing caller depends on the old behavior: the only `createRef` consumers in the Solid packages (`heading.tsx`, `focus-trap-region.tsx`) use `createRef<HTMLType>()` with no initial value and never call `.reset()`.

### Areas reviewed without finding issues
- **Tests** cover all three symptoms from the issue (unexpected focus side-effect, ref becoming `undefined`, and the follow-up `current is not a function`). The happy-dom `test.ts` is collected only for the Solid loader (matched `index.solid.tsx`) and excluded from the default suite; the browser `test-browser.ts` runs Solid-only via `withFramework`/`getPreviewFrameworksSync`. Logic mirrors the happy-dom test, which I validated locally.
- **Changeset** (`300-solid-create-ref-reset.md`, `@ariakit/solid-utils` patch) matches naming/format conventions and uses a plain code span for the experimental API.
- **`app/package.json`** adds `@ariakit/solid-utils: workspace:*` in alphabetical order under `dependencies`, consistent with the other `@ariakit/*` entries; lockfile updated accordingly.
- **Code style**: nested helpers use `const` arrow form, module-level export uses `function`, and the WHY comment on `reset` is appropriate.

- The Playwright `test-browser.ts` was not executed locally (requires the dev server); it's logically identical to the happy-dom test I ran and the bug is framework-level (Solid setter), not DOM-divergent, so CI coverage should be sufficient.
- One non-blocking observation: the sandbox uses plain `<input>`/`<button>`/`<output>` rather than Ariakit components. That's reasonable here since `createRef` is a low-level utility with no component-level public surface for `.reset()` on a function ref, and the sandbox still surfaces state through visible UI and focus per the sandbox guidance.

PR #6418, opened 2026-06-23
Isolate app check Vite cache

diegohaz:
## Motivation

Running `pnpm tsc` while `pnpm dev-app` is serving the app can corrupt the live dev server's Vite optimized dependency cache. The `tsc` script runs `pnpm -F app run check`, and `astro check` uses the same default `app/node_modules/.vite` cache as `astro dev`.

When the checker re-optimizes dependencies during an active dev request, the dev server can keep references to files that were replaced, causing repeated errors like missing `deps_ssr` files and blank preview pages in Playwright.

## Solution

Expose an app-specific `APP_VITE_CACHE_DIR` environment hook in `astro.config.ts` and run `astro check` with `APP_VITE_CACHE_DIR=node_modules/.vite-check`.

This keeps normal dev/build commands on the default Vite cache while isolating check-time optimizer output from a live `pnpm dev-app` session. The fix is scoped to the reported `dev-app` plus `pnpm tsc` collision, and direct `pnpm -F app run check` now gets the same isolated cache as the root `pnpm tsc` flow.

## Verification

- `pnpm lint-fix`
- Reproduced the original failure before the fix with `APP_PORT=4324 NEXTJS_PORT=3003 pnpm dev-app`, then concurrent `pnpm tsc` and `APP_PORT=4324 NEXTJS_PORT=3003 pnpm -F app test-chrome ariakit-tailwind`: the dev server logged repeated missing `app/node_modules/.vite/deps_ssr/...` errors and the focused Chrome run failed with empty/missing preview DOM.
- Re-ran the same concurrent scenario after the fix from a clean app Vite cache: `pnpm tsc` passed, `APP_PORT=4324 NEXTJS_PORT=3003 pnpm -F app test-chrome ariakit-tailwind` passed 8/8, and the dev server did not log missing optimized dependency errors.

### DIFF
--- a/app/astro.config.ts
+++ b/app/astro.config.ts
@@ -35,6 +35,7 @@ try {
 
 const port = Number(process.env.APP_PORT) || 4321;
 const hasClerk = process.env.PUBLIC_CLERK_PUBLISHABLE_KEY;
+const viteCacheDir = process.env.APP_VITE_CACHE_DIR;
 
 // https://astro.build/config
 export default defineConfig({
@@ -68,6 +69,7 @@ export default defineConfig({
   },
 
   vite: {
+    cacheDir: viteCacheDir,
     build: {
       // Perf CI enables this so CDP script profiles can resolve source maps.
       sourcemap: process.env.PERF_SOURCE_MAP === "true",
--- a/app/astro.config.ts
+++ b/app/astro.config.ts
@@ -69,6 +69,11 @@ export default defineConfig({
   },
 
   vite: {
+    // TODO: Remove this workaround once Astro isolates optimizer cache writes
+    // across dev and check commands. Running `astro check` while `astro dev`
+    // is active shares Vite's optimized deps cache, so the checker uses a
+    // separate cache directory to avoid invalidating the dev server's optimized
+    // SSR modules.
     cacheDir: viteCacheDir,
     build: {
       // Perf CI enables this so CDP script profiles can resolve source maps.
@@ -79,7 +84,7 @@ export default defineConfig({
       sourcePlugin(join(import.meta.dirname, "src/examples/")),
     ],
     // TODO: Remove this workaround once
-    // https://github.com/withastro/astro/issues/16853 is fixed. These bare
+    // https://github.com/withastro/astro/issues/17166 is fixed. These bare
     // specifiers are missed by the SSR dependency optimizer's initial scan.
     // Some are re-exported from Astro virtual modules (astro:transitions,
     // astro:actions); others are imported directly (astro/zod in
--- a/app/package.json
+++ b/app/package.json
@@ -17,7 +17,7 @@
     "dev-stripe": "concurrently -k -r 'pnpm:dev' 'pnpm:listen-*'",
     "dev-stripe-lite": "concurrently -k -r 'pnpm:dev-lite' 'pnpm:listen-*'",
     "wrangler-types": "wrangler types",
-    "check": "pnpm run wrangler-types && astro check",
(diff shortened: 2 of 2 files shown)


app/astro.config.ts
```diff
@@ -68,6 +69,7 @@ export default defineConfig({
   },
 
   vite: {
+    cacheDir: viteCacheDir,
```
diegohaz, line 77: I think we should also add a TODO comment to remove this when it's fixed upstream. Also update the other TODO comment so it references https://github.com/withastro/astro/issues/17166 instead of https://github.com/withastro/astro/issues/16853

diegohaz: Fixed in 5704da982 - added a TODO for the check/dev Vite cache workaround, kept it separate from the SSR optimizer issue, and updated the existing optimizer TODO to reference withastro/astro#17166.

diegohaz:
**Fixed**

- 5704da982 (https://github.com/ariakit/ariakit/pull/6418#discussion_r3462743954): Added the removal TODO for the app check/dev Vite cache workaround, kept it separate from the SSR optimizer issue, and updated the existing optimizer TODO to reference withastro/astro#17166.

diegohaz (commented):
**Review summary — no substantive findings.**

Initial (full) review. Baseline: merge-base `0bdf5d1` (`origin/main`) → head `5704da9`. Scope: both changed files (`app/astro.config.ts`, `app/package.json`) plus surrounding config, the root `tsc` script chain, changeset config, and repo-wide `astro check` / env-hook usage.

The fix is clean and well-scoped. I reviewed for correctness, simpler alternatives, consistency, and tooling impact and found nothing worth changing:

- **`cacheDir: viteCacheDir` is a safe no-op outside `check`.** `APP_VITE_CACHE_DIR` is unset for `dev`/`build`/`preview`, so `viteCacheDir` is `undefined`, and Vite falls back to its default `node_modules/.vite` when `cacheDir` is falsy. Type is also fine — Vite's `cacheDir?: string` accepts `undefined`.
- **Path resolves correctly.** `node_modules/.vite-check` is resolved relative to the app's Vite root (`app/`), giving `app/node_modules/.vite-check`, mirroring the default cache location and staying inside the gitignored `node_modules`.
- **Scope is complete.** The `check` script is the only `astro check` invoker in the repo, so both root `pnpm tsc` (`pnpm -F app run check && tsc -b`) and direct `pnpm -F app run check` pick up the isolated cache. CI doesn't run a concurrent `dev-app`, so leaving CI on this same path is harmless.
- **Env-hook approach is the right trade-off.** Reading a path from `APP_VITE_CACHE_DIR` matches the existing `APP_PORT` pattern and is simpler/safer than gating `cacheDir` on Astro's `command` (which doesn't cleanly distinguish `check`) or disabling the optimizer during checks.
- **Tooling.** `cross-env@10.1.0` is already a devDependency. No changeset needed: `app` is private and in `.changeset/config.json` `ignore`. New comment lines are all ≤80 chars (the only >80 line, 147, is pre-existing and untouched). The follow-up commit correctly added the removal TODO and re-pointed the existing optimizer TODO to withastro/astro#17166, matching the resolved self-review thread.

**Not independently re-validated:** the efficacy of the cache isolation itself (that `astro check` writing to a separate `cacheDir` stops corrupting a live `dev-app` SSR optimizer cache). I relied on the author's stated before/after repro (8/8 `ariakit-tailwind` Chrome run passing after the fix) rather than re-running the concurrent `dev-app` + `tsc` scenario, since it's a config-only env hook and the repro is expensive to reproduce locally. The perf-bot `ariakit-tailwind` "toggle layer and text classes" +27% warning is unrelated to this change (config-only, no runtime/package impact) and consistent with the known style-recalc variance on that benchmark.

PR #6419, opened 2026-06-23
Fix TagInput string delimiter matching

### LINKED ISSUE #6333: TagInput: string delimiters with regex metacharacters freeze the page ("|"), throw ("+"), or silently never split (".")
### Current behavior

`splitValueByDelimiter` accepts `string | RegExp` delimiters and passes them straight to `String.prototype.match`:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/tag/utils.ts#L18-L30

```ts
export function splitValueByDelimiter(value: string, delimiters: Delimiter[]) {
  for (const delimiter of delimiters) {
    let match = value.match(delimiter);
    // Remove delimiter from the start of the string
    while (match?.index === 0) {
      value = value.slice(match[0].length);
      match = value.match(delimiter);
    }
    if (!match) continue;
    return value.split(delimiter);
  }
  return [];
}
```

Per spec, a non-RegExp argument to `match` is compiled via `new RegExp(delimiter)` **without escaping**, so a string delimiter containing regex metacharacters is silently treated as a pattern — while `value.split(delimiter)` in the same function treats the identical string literally. Depending on which metacharacter the string contains, this produces three distinct failures:

- `delimiter="|"`: `new RegExp("|")` matches the empty string at index 0, so `match[0].length` is `0`, `value.slice(0)` never advances, and the `while` loop spins forever — a synchronous infinite loop that freezes the tab.
- `delimiter="+"` (likewise `"("`, `"["`, `"*"`, `"?"`): `new RegExp("+")` throws `SyntaxError: Invalid regular expression: /+/: Nothing to repeat`.
- `delimiter="."`: the regex matches every character, so the leading-strip loop consumes the whole value one character at a time and the function returns `[]` — the input is silently never split into tags.

Zero-length-matching RegExp delimiters such as `/x*/` hit the same infinite loop, since the guard only checks `match?.index === 0`, never the match length.

This is not an obscure code path: `TagInput`'s `onChange` handler calls `splitValueByDelimiter` on every change event whenever the caret sits at the end of the input — i.e. during normal typing — and `onPaste` calls it on every paste:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/tag/tag-input.tsx#L119-L124

So with `delimiter="|"` the page freezes on the **first keystroke**. Nothing in the prop's type (`string | RegExp | null | (string | RegExp)[]`) or its JSDoc ("The string or pattern employed to break the input value into multiple tags. This could be a string...") hints that `"|"` is unsafe while `";"` is fine. The default delimiters (`["\n", ";", ",", /\s/]`) happen to be metachar-free, and the existing test table in `utils.test.ts` only exercises safe delimiters, which is why this has gone unnoticed.

3. Click the "Pipe tags" input and type one character.
4. The preview page freezes immediately and permanently: the change handler never returns, no tag is ever created, and no further interaction (clicks, typing, scrolling) works in the preview tab.
5. Reload, click the "Dot tags" input and type `one.two.` — no tag is ever created; the raw text just accumulates in the input, even though `.` is the configured delimiter.

```tsx
import { Tag } from "@ariakit/react-components/tag/tag";
import { TagInput } from "@ariakit/react-components/tag/tag-input";
import { TagList } from "@ariakit/react-components/tag/tag-list";
import { TagListLabel } from "@ariakit/react-components/tag/tag-list-label";
import { TagProvider } from "@ariakit/react-components/tag/tag-provider";
import { TagRemove } from "@ariakit/react-components/tag/tag-remove";

function TagField({ label, delimiter }: { label: string; delimiter: string }) {
  const [values, setValues] = useState<string[]>([]);
  return (
    <TagProvider values={values} setValues={setValues}>
      <TagListLabel>{label}</TagListLabel>
      <TagList>
        {values.map((value) => (
            {value}
            <TagRemove />
          </Tag>
        ))}
        <TagInput delimiter={delimiter} aria-label={label} />
      </TagList>
    </TagProvider>
  );
}

  return (
    <>
      <TagField label="Pipe tags" delimiter="|" />
      <TagField label="Dot tags" delimiter="." />
    </>
  );
}
```

Verified with a real-browser Playwright test: on current code, typing into the pipe input times out mid-dispatch (the renderer main thread is blocked inside the synchronous change handler, and a subsequent click elsewhere on the page also times out), and typing `one.two.` into the dot input never creates a tag — while the same test passes with the workaround below.

### Expected behavior

A string delimiter should be matched literally, exactly as `value.split(delimiter)` already treats it and as the `delimiter` prop docs promise. Typing `one|two|` with `delimiter="|"` (or `one.two.` with `delimiter="."`) should add the tags "one" and "two" and leave the input empty, just like typing `one,two,` does with the default delimiters — and the page should obviously stay responsive.

### Workaround

Pass an equivalent regular expression instead of a plain string, so the match step can't compile it differently from the split step:

```tsx
<TagInput delimiter={/[|]/} aria-label="Pipe tags" />
<TagInput delimiter={/\./} aria-label="Dot tags" />
```

Any metachar-bearing string delimiter can be wrapped this way (e.g. via `new RegExp(RegExp.escape(delimiter))` on platforms with `RegExp.escape`, or a character class as above).

### Possible solutions

Match string delimiters literally (via `indexOf`) instead of letting `String.prototype.match` compile them as regexes, and guard the leading-strip loop against zero-length matches so patterns like `/x*/` can't loop forever either:

```diff
+interface DelimiterMatch {
+  /**
+   * Index of the first match. Undefined for global regexes, whose
+   * `String.prototype.match` result doesn't expose match positions.
+   */
+  index?: number;
+  length: number;
+}
+
+function matchDelimiter(
+  value: string,
+  delimiter: Delimiter,
+): DelimiterMatch | null {
+  // Match string delimiters literally. Passing a string to
+  // `String.prototype.match` compiles it as a regex without escaping, which
+  // hangs (e.g., "|"), throws (e.g., "+"), or mismatches (e.g., "."), while
+  // the final `value.split(delimiter)` treats the same string literally.
+  if (typeof delimiter === "string") {
+    const index = value.indexOf(delimiter);
+    if (index === -1) return null;
+    return { index, length: delimiter.length };
+  }
+  const match = value.match(delimiter);
+  if (!match) return null;
+  return { index: match.index, length: match[0].length };
+}
+
 export function splitValueByDelimiter(value: string, delimiters: Delimiter[]) {
   for (const delimiter of delimiters) {
-    let match = value.match(delimiter);
-    // Remove delimiter from the start of the string
-    while (match?.index === 0) {
-      value = value.slice(match[0].length);
-      match = value.match(delimiter);
+    let match = matchDelimiter(value, delimiter);
+    // Remove delimiter from the start of the string. Skip zero-length matches
+    // (e.g., /x*/) that would otherwise loop forever.
+    while (match?.index === 0 && match.length > 0) {
+      value = value.slice(match.length);
+      match = matchDelimiter(value, delimiter);
     }
     if (!match) continue;
     return value.split(delimiter);
   }
   return [];
 }
```

One subtlety: for global-flag regexes (e.g. `/,/g`), `match` returns an array without an `index` property; the current code relies on that to skip the strip loop while still reaching `split` (pinned by the `,one,two` + `/,/g` case in `utils.test.ts`). The helper above keeps `index` optional so that behavior is preserved. This implementation was verified by execution against all 7 existing `utils.test.ts` cases plus new cases for `"|"`, `"+"`, `"."` (with leading strip) and `/x*/` — all pass. Alternatively, string delimiters could be escaped before compiling (e.g. `RegExp.escape`), but `indexOf` is simpler and exactly mirrors what `split` already does with string separators.

diegohaz:
## Motivation

`TagInput` currently treats string delimiters as regular expressions while checking whether an input should split, but later passes the same delimiter to `String.prototype.split`, where strings are matched literally.

This makes delimiters with regex metacharacters behave incorrectly. For example, `delimiter="."` never creates tags from `one.two.`, `delimiter="+"` throws while the user types, and `delimiter="|"` can freeze the page because it creates a zero-length regex match.

Fixes #6333

## Solution

The `splitValueByDelimiter` helper now matches string delimiters literally with `indexOf`, keeping the match step consistent with the later `split` call.

For regular expression delimiters, the existing match behavior is preserved, including the global-regex case where `String.prototype.match` does not expose an `index`. The leading-delimiter strip loop now also ignores zero-length matches so patterns such as `/x*/` cannot loop forever.

The `tag-input-6333` sandbox keeps the original string-delimiter repro, and the added unit coverage includes `.`, `+`, `|`, and a zero-length regex delimiter.

## Workaround

Pass an equivalent regular expression instead of a plain string delimiter so the matching and splitting steps both use regex semantics:

```tsx
// TODO: Remove this workaround once https://github.com/ariakit/ariakit/issues/6333 is fixed.
<TagInput delimiter={/\./} aria-label="Dot tags" />
<TagInput delimiter={/[+]/} aria-label="Plus tags" />
```

### DIFF
--- a/app/src/sandbox/tag-input-6333/index.react.tsx
+++ b/app/src/sandbox/tag-input-6333/index.react.tsx
@@ -1,0 +1,42 @@
+import { Tag } from "@ariakit/react-components/tag/tag";
+import { TagInput } from "@ariakit/react-components/tag/tag-input";
+import { TagList } from "@ariakit/react-components/tag/tag-list";
+import { TagListLabel } from "@ariakit/react-components/tag/tag-list-label";
+import { TagProvider } from "@ariakit/react-components/tag/tag-provider";
+import { useId, useState } from "react";
+
+interface TagFieldProps {
+  label: string;
+  delimiter: string;
+}
+
+function TagField({ label, delimiter }: TagFieldProps) {
+  const [values, setValues] = useState<string[]>([]);
+  const statusId = useId();
+
+  return (
+    <TagProvider values={values} setValues={setValues}>
+      <TagListLabel>{label}</TagListLabel>
+      <TagList aria-describedby={statusId}>
+        {values.map((value) => (
+          <Tag key={value} value={value}>
+            {value}
+          </Tag>
+        ))}
+        <TagInput delimiter={delimiter} aria-label={label} />
+      </TagList>
+      <output id={statusId}>
+        {label} values: {values.length ? values.join(", ") : "none"}
+      </output>
+    </TagProvider>
+  );
+}
+
+export default function Example() {
+  return (
+    <div style={{ display: "grid", gap: 16, padding: 16 }}>
+      <TagField label="Dot tags" delimiter="." />
+      <TagField label="Plus tags" delimiter="+" />
+    </div>
+  );
+}
--- a/app/src/sandbox/tag-input-6333/index.react.tsx
+++ b/app/src/sandbox/tag-input-6333/index.react.tsx
@@ -7,7 +7,7 @@ import { useId, useState } from "react";
 
 interface TagFieldProps {
   label: string;
-  delimiter: string;
+  delimiter: string | RegExp;
 }
 
 function TagField({ label, delimiter }: TagFieldProps) {
@@ -35,8 +35,8 @@ function TagField({ label, delimiter }: TagFieldProps) {
 export default function Example() {
   return (
     <div style={{ display: "grid", gap: 16, padding: 16 }}>
-      <TagField label="Dot tags" delimiter="." />
-      <TagField label="Plus tags" delimiter="+" />
+      <TagField label="Dot tags" delimiter={/\./} />
+      <TagField label="Plus tags" delimiter={/[+]/} />
     </div>
   );
 }
--- a/app/src/sandbox/tag-input-6333/index.react.tsx
+++ b/app/src/sandbox/tag-input-6333/index.react.tsx
@@ -37,6 +37,7 @@ export default function Example() {
     <div style={{ display: "grid", gap: 16, padding: 16 }}>
       <TagField label="Dot tags" delimiter="." />
       <TagField label="Plus tags" delimiter="+" />
+      <TagField label="Pipe tags" delimiter="|" />
     </div>
   );
 }
--- a/app/src/sandbox/tag-input-6333/index.react.tsx
+++ b/app/src/sandbox/tag-input-6333/index.react.tsx
@@ -7,7 +7,7 @@ import { useId, useState } from "react";
 
 interface TagFieldProps {
   label: string;
-  delimiter: string | RegExp;
+  delimiter: string;
 }
 
 function TagField({ label, delimiter }: TagFieldProps) {
@@ -35,8 +35,8 @@ function TagField({ label, delimiter }: TagFieldProps) {
 export default function Example() {
   return (
     <div style={{ display: "grid", gap: 16, padding: 16 }}>
-      <TagField label="Dot tags" delimiter={/\./} />
-      <TagField label="Plus tags" delimiter={/[+]/} />
+      <TagField label="Dot tags" delimiter="." />
+      <TagField label="Plus tags" delimiter="+" />
     </div>
   );
 }
--- a/.changeset/300-tag-input-string-delimiters.md
+++ b/.changeset/300-tag-input-string-delimiters.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Fixed `TagInput` string delimiters with regex metacharacters so they're matched literally and don't freeze, throw, or fail to split tags.
--- a/app/src/sandbox/tag-input-6333/test-browser.ts
+++ b/app/src/sandbox/tag-input-6333/test-browser.ts
@@ -19,4 +19,13 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await page.keyboard.type("one+two+");
     await test.expect(q.text("Plus tags values: one, two")).toBeVisible();
   });
+
+  test("does not freeze on string delimiters that match empty regexes", async ({
+    page,
+    q,
+  }) => {
+    await q.textbox("Pipe tags").click();
+    await page.keyboard.type("one|two|");
+    await test.expect(q.text("Pipe tags values: one, two")).toBeVisible();
+  });
 });
--- a/app/src/sandbox/tag-input-6333/test-browser.ts
+++ b/app/src/sandbox/tag-input-6333/test-browser.ts
@@ -1,0 +1,22 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6333
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("splits string delimiters with regex metacharacters literally", async ({
+    page,
+    q,
+  }) => {
+    await q.textbox("Dot tags").click();
+    await page.keyboard.type("one.two.");
+    await test.expect(q.text("Dot tags values: one, two")).toBeVisible();
+  });
+
+  test("does not throw on string delimiters that are invalid regex patterns", async ({
+    page,
+    q,
+  }) => {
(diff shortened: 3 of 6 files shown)


app/src/sandbox/tag-input-6333/index.react.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </output>
+    </TagProvider>
+  );
+}
+
+export default function Example() {
+  return (
+    <div style={{ display: "grid", gap: 16, padding: 16 }}>
+      <TagField label="Dot tags" delimiter="." />
+      <TagField label="Plus tags" delimiter="+" />
```
diegohaz, line 39: Optional: consider adding a third `TagField` here with `delimiter="|"` (and a matching `one|two|` case in `test.ts`/`test-browser.ts`).

The `|` delimiter is the headline symptom in #6333 — `new RegExp("|")` matches the empty string at index 0, so on the old code the leading-strip loop spun forever and froze the tab on the first keystroke. The sandbox currently exercises the silent-no-split (`.`) and throw (`+`) cases in a real browser, but not the page-freeze case. The `|` and `/x*/` cases are covered by the unit test, but a freeze regression there surfaces only as an opaque Vitest timeout; a browser-level repro of the most severe symptom would be the strongest guard.

Not blocking — the fix itself is correct and the existing coverage is solid. I verified locally that `"|"` and `["", ","]` (an empty-string delimiter inside an array, also a latent freeze) both now terminate and split as expected.

diegohaz (commented):
**Review — initial, full** (commit `5ac3b88`, baseline `merge-base` with `origin/main` = `57dfad9`).

The fix is correct, minimal, and complete. `matchDelimiter` routes string delimiters through `indexOf` (literal, mirroring the later `value.split(delimiter)`) and regex delimiters through `value.match`, and the leading-strip loop's new `match.length > 0` guard kills the zero-length infinite loop. The implementation matches the solution proposed in #6333, including keeping `DelimiterMatch.index` optional so the global-regex (`/,/g`) skip-strip behavior is preserved.

**Validation run locally:**
- A standalone before/after harness over all delimiter cases: the 4 buggy inputs are fixed — `.` (silently never split → now `["one","two",""]`), `+` (threw `SyntaxError` → now splits), `|` (infinite loop/freeze → now splits), `/x*/` (infinite loop → now `["a","b","c"]`) — and all 6 previously-correct cases (array first-match-wins, leading strip, trailing, value-only, `/,/g` global, `/([,;])/` capturing) are byte-identical to base. Also confirmed `["", ","]` (empty string delimiter inside an array) was a latent freeze on base that this fix incidentally resolves — no regression.
- `pnpm test-react packages/ariakit-react-components/src/tag/utils.test.ts` → 12 passed.
- `pnpm test-react app/src/sandbox/tag-input-6333/test.ts` → 2 passed.
- `pnpm tsc` → 0 errors (new `DelimiterMatch` types sound).

**Findings:** one optional, non-blocking inline note — the `|` page-freeze case (the headline symptom of #6333) is covered only by the unit test, not by the browser sandbox; a `|` browser repro would be the strongest regression guard since a freeze otherwise surfaces only as an opaque Vitest timeout. No correctness, simplification, perf, compatibility, or consistency issues found. Sandbox naming (`tag-input-6333`), the `withFramework` browser test, the omitted `preview.json`, and the changeset (`@ariakit/react-components` patch) all follow repo conventions.

**Not validated:** Playwright browser run (`pnpm -F app test`) and the React 18 suite — the change is pure string logic already exercised by the happy-dom + unit suites under React 19, so I didn't spin up a dev server for it.

app/src/sandbox/tag-input-6333/index.react.tsx
diegohaz: Fixed in aa74dd3c6 - added the pipe delimiter to the sandbox and covered `one|two|` in both happy-dom and browser tests.

diegohaz:
**Fixed**

- aa74dd3c6 ([reply](https://github.com/ariakit/ariakit/pull/6419#discussion_r3463470054)): Added the `|` delimiter to the TagInput sandbox and covered `one|two|` in both happy-dom and browser tests.

PR #6420, opened 2026-06-23
Fix TagInput IME composition splitting

### LINKED ISSUE #6294: TagInput splits IME composition text into junk tags and destroys the composition session
### Current behavior

`TagInput`'s `onChange` handler runs its delimiter-splitting logic on every change event with no IME composition guard ([`tag-input.tsx#L119-L146`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/tag/tag-input.tsx#L119-L146)). The default delimiter list includes `/\s/` ([`utils.ts#L7`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/tag/utils.ts#L7)), which matches the spaces CJK IMEs place inside the composition (preedit) buffer.

macOS Pinyin, for example, displays multi-syllable preedit as `ni hao` — with a space — while the user is still composing. Each composition update fires a change event (`insertCompositionText`) with the caret at the end, so the trailing-caret check passes, the preedit is split, `store.addValue("ni")` commits a junk tag, and the store-controlled input is rewritten mid-composition, which destroys the IME session.

Concretely, composing `ni hao` and committing `你好` ends up with a junk tag **`ni`** and the input containing **`hao你好`** instead of `你好`.

Sibling components treat composition as a first-class concern — [`Combobox`](https://ariakit.com/reference/combobox) checks `nativeEvent.isComposing` in its change handler ([`combobox.tsx#L515-L519`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox.tsx#L515-L519)) — so this looks like an omission rather than a design choice.

3. Switch your OS keyboard to a CJK IME (e.g. macOS Pinyin) and focus the input.
4. Type `nihao` — as soon as the preedit shows `ni hao`, a junk tag `ni` is committed and the composition is destroyed; picking the `你好` candidate leaves garbled text in the input.

```tsx
import { Tag } from "@ariakit/react-components/tag/tag";
import { TagInput } from "@ariakit/react-components/tag/tag-input";
import { TagList } from "@ariakit/react-components/tag/tag-list";
import { TagListLabel } from "@ariakit/react-components/tag/tag-list-label";
import { TagProvider } from "@ariakit/react-components/tag/tag-provider";
import { TagRemove } from "@ariakit/react-components/tag/tag-remove";

  const [values, setValues] = useState<string[]>([]);
  return (
    <TagProvider values={values} setValues={setValues}>
      <TagListLabel>Tags</TagListLabel>
      <TagList>
        {values.map((value) => (
            {value}
            <TagRemove />
          </Tag>
        ))}
        <TagInput aria-label="New tag" />
      </TagList>
    </TagProvider>
  );
}
```

Verified in Chromium with a Playwright test that drives the browser's real IME pipeline via CDP (`Input.imeSetComposition` steps `n` → `ni` → `ni h` → `ni ha` → `ni hao`, then `Input.insertText("你好")` — the same code path OS IMEs use; no mocks or synthetic DOM events). Result on current code: a tag `ni` exists and the input value is `hao你好`.

### Expected behavior

Composition updates are ignored by the delimiter-splitting logic while `isComposing` is true: no tags are committed from preedit text, the input is not rewritten mid-composition, and after the user picks a candidate the input simply contains `你好`. Splitting should only consider committed text (as `Combobox` does by deferring work until composition ends).

### Workaround

A single `onChange` guard on the consumer side fixes it — `TagInput` calls the consumer's handler first and bails out when the event is `defaultPrevented`, and the change event fired when the composition commits has `isComposing: false`, so the committed value still syncs to the store:

```tsx
<TagInput
  aria-label="New tag"
  onChange={(event) => {
    if ((event.nativeEvent as InputEvent).isComposing) {
      event.preventDefault();
    }
  }}
/>
```

With this applied, the same Playwright IME scenario passes: no junk tags, and the input ends up with exactly `你好`. Non-IME typing is unaffected (delimiter splitting still works for regular keystrokes).

### Possible solutions

Mirror `Combobox`'s composition handling inside `useTagInput`:

```diff
     const onChange = useEvent((event: ChangeEvent<HTMLType>) => {
       onChangeProp?.(event);
       if (event.defaultPrevented) return;
+      // Don't split preedit text or rewrite the input while an IME
+      // composition session is in progress. The final change event after
+      // compositionend has isComposing === false, so committed text still
+      // goes through the regular path below.
+      const nativeEvent = event.nativeEvent;
+      if (isInputEvent(nativeEvent) && nativeEvent.isComposing) {
+        const { value } = event.currentTarget;
+        void UndoManager.execute(() => {
+          store.setValue(value);
+          return () => store.setValue(value);
+        }, getInputType(event));
+        return;
+      }
       const { value: prevValue } = store.getState();
```

(Keeping the store value in sync during composition while skipping the delimiter-splitting block; alternatively defer the split to `onCompositionEnd` exactly like [`combobox.tsx#L515-L519`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox.tsx#L515-L519) finalizes deferred work there.)

---

*Found by an automated multi-agent audit of the repo at commit [`11bd4458d`](https://github.com/ariakit/ariakit/tree/11bd4458de6447b996dd882ea0c85eb801a6cd03) and verified with a real-browser Playwright reproduction (browser-level IME events only).*

diegohaz:
## Motivation

`TagInput` ran delimiter splitting while an IME composition was still in progress. With the default whitespace delimiter, CJK preedit text such as `ni hao` could be split before the user committed the final candidate, producing a junk `ni` tag and leaving garbled input text.

Fixes #6294

## Solution

Added a focused `tag-input-6294` sandbox that reproduces the issue with Chromium's real IME composition pipeline through CDP.

`useTagInput` now keeps the input value synced during composition, but skips delimiter splitting while the native input event has `isComposing` set. Once the user commits the composed text, the following input event has `isComposing` unset and goes through the normal path.

A patch changeset is included for `@ariakit/react-components`.

## Workaround

Until the package fix is available, prevent `TagInput` from running its delimiter-splitting logic while the native input event is still composing:

```tsx
<TagInput
  aria-label="New tag"
  onChange={(event) => {
    // TODO: Remove once https://github.com/ariakit/ariakit/issues/6294 is fixed.
    if ((event.nativeEvent as InputEvent).isComposing) {
      event.preventDefault();
    }
  }}
/>
```

### DIFF
--- a/app/src/sandbox/tag-input-6294/test-browser.ts
+++ b/app/src/sandbox/tag-input-6294/test-browser.ts
@@ -1,0 +1,32 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("does not split IME composition text into tags", async ({
+    browserName,
+    page,
+    q,
+  }) => {
+    test.skip(
+      browserName !== "chromium",
+      "CDP IME composition is Chromium-only.",
+    );
+
+    const input = q.textbox("New tag");
+    await input.focus();
+
+    const cdp = await page.context().newCDPSession(page);
+
+    for (const text of ["n", "ni", "ni h", "ni ha", "ni hao"]) {
+      await cdp.send("Input.imeSetComposition", {
+        text,
+        selectionStart: text.length,
+        selectionEnd: text.length,
+      });
+    }
+
+    await cdp.send("Input.insertText", { text: "你好" });
+
+    await test.expect(q.option("ni")).toBeHidden();
+    await test.expect(input).toHaveValue("你好");
+  });
+});
--- a/packages/ariakit-react-components/src/tag/tag-input.tsx
+++ b/packages/ariakit-react-components/src/tag/tag-input.tsx
@@ -12,6 +12,7 @@ import {
   getTextboxSelection,
   setSelectionRange,
   getInputType,
+  isInputEvent,
   invariant,
   UndoManager,
 } from "@ariakit/utils";
@@ -36,10 +37,6 @@ type HTMLType = HTMLElementTagNameMap[TagName];
 type EventWithValues<T extends SyntheticEvent> = T & {
   values: string[];
 };
-
-function isInputEvent(event: Event): event is InputEvent {
-  return event.type === "input";
-}
 
 /**
  * Returns props to create a `TagInput` component.
--- a/packages/ariakit-react-components/src/tag/tag-input.tsx
+++ b/packages/ariakit-react-components/src/tag/tag-input.tsx
@@ -37,6 +37,10 @@ type EventWithValues<T extends SyntheticEvent> = T & {
   values: string[];
 };
 
+function isInputEvent(event: Event): event is InputEvent {
+  return event.type === "input";
+}
+
 /**
  * Returns props to create a `TagInput` component.
  * @see https://ariakit.com/components/tag
@@ -116,6 +120,8 @@ export const useTagInput = createHook<TagName, TagInputOptions>(
           return () => store.setValue(prevValue);
         }, inputType);
       }
+      const nativeEvent = event.nativeEvent;
+      if (isInputEvent(nativeEvent) && nativeEvent.isComposing) return;
       // Add values to the store if the input value ends with a delimiter
       const isTrailingCaret = start === end && start === value.length;
       if (isTrailingCaret) {
--- a/.changeset/300-input-event.md
+++ b/.changeset/300-input-event.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Added `isInputEvent`.
--- a/.changeset/300-tag-input-ime.md
+++ b/.changeset/300-tag-input-ime.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Fixed `TagInput` so IME composition text is not split into tags before the user commits the composed value.
--- a/app/src/sandbox/tag-input-6294/index.react.tsx
+++ b/app/src/sandbox/tag-input-6294/index.react.tsx
@@ -1,0 +1,26 @@
+import { Tag } from "@ariakit/react-components/tag/tag";
+import { TagInput } from "@ariakit/react-components/tag/tag-input";
+import { TagList } from "@ariakit/react-components/tag/tag-list";
+import { TagListLabel } from "@ariakit/react-components/tag/tag-list-label";
+import { TagProvider } from "@ariakit/react-components/tag/tag-provider";
+import { TagRemove } from "@ariakit/react-components/tag/tag-remove";
+import { useState } from "react";
+
+export default function Example() {
+  const [values, setValues] = useState<string[]>([]);
+
+  return (
+    <TagProvider values={values} setValues={setValues}>
+      <TagListLabel>Tags</TagListLabel>
+      <TagList>
+        {values.map((value) => (
+          <Tag key={value} value={value}>
+            {value}
+            <TagRemove />
+          </Tag>
+        ))}
+        <TagInput aria-label="New tag" />
+      </TagList>
+    </TagProvider>
+  );
+}
--- a/app/src/sandbox/tag-input-6294/index.react.tsx
+++ b/app/src/sandbox/tag-input-6294/index.react.tsx
@@ -19,14 +19,7 @@ export default function Example() {
             <TagRemove />
           </Tag>
         ))}
-        <TagInput
-          aria-label="New tag"
-          onChange={(event) => {
-            if ((event.nativeEvent as InputEvent).isComposing) {
-              event.preventDefault();
-            }
-          }}
-        />
+        <TagInput aria-label="New tag" />
       </TagList>
     </TagProvider>
   );
--- a/app/src/sandbox/tag-input-6294/index.react.tsx
+++ b/app/src/sandbox/tag-input-6294/index.react.tsx
@@ -19,7 +19,14 @@ export default function Example() {
             <TagRemove />
           </Tag>
         ))}
-        <TagInput aria-label="New tag" />
+        <TagInput
+          aria-label="New tag"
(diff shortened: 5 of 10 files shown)


packages/ariakit-react-components/src/tag/tag-input.tsx
```diff
@@ -37,6 +37,10 @@ type EventWithValues<T extends SyntheticEvent> = T & {
   values: string[];
 };
 
+function isInputEvent(event: Event): event is InputEvent {
```
diegohaz: Optional: this `isInputEvent` type guard is a verbatim copy of the one already defined in `combobox.tsx` (`function isInputEvent(event: Event): event is InputEvent { return event.type === "input"; }`, around line 77). Now that two component files need it, it could be hoisted into `@ariakit/utils` next to its sibling `getInputType` in `packages/ariakit-utils/src/events.ts` — both `combobox.tsx` and `tag-input.tsx` already import from `@ariakit/utils`, so the consolidation is mechanical and would drop the duplicate. Not blocking: the codebase tolerates file-local micro-guards and `getInputType` was already split out separately, so keeping a local copy here is also defensible; flagging only as a behavior-preserving DRY cleanup if you'd rather not duplicate it.

diegohaz (commented):
## Review summary

**Scope:** Initial review of #6420 (head `78eff12`). Diff baseline `git merge-base FETCH_HEAD origin/main` = `57dfad97`. Reviewed all 4 changed files: the `useTagInput` IME guard, the `tag-input-6294` sandbox + browser test, and the changeset.

**Verdict:** The fix is correct, minimal, and well-targeted. The approach is sound and is an improvement over the issue's suggested diff. One optional DRY note posted inline; no blocking issues.

### What I verified
- **The fix works and the test is meaningful.** Ran the browser test in this worktree (`APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app test tag-input-6294`): passes on chromium, correctly skips on firefox/safari. Then reverted the `tag-input.tsx` guard and re-ran: the test fails as expected (the junk `ni` option becomes visible, `getByRole('option', {name:'ni'})` resolves to a real tag), confirming the test actually catches the regression rather than passing vacuously. Restored the fix; working tree clean.
- **Guard placement is correct.** The early return sits after the `setValueOnChange` block, so the store value still stays synced during composition (matching the PR description) and only the delimiter-splitting is skipped while `nativeEvent.isComposing` is true. The committed change event after `compositionend` has `isComposing` unset, so committed text still flows through the normal split path — non-IME typing is unaffected. This is cleaner than the issue's proposed diff (which added a second `store.setValue` inside its own guard) because it reuses the existing sync logic.
- **Consistency with siblings.** Mirrors `Combobox`'s `isComposing` handling in its `onChange`; `UndoManager`/`getInputType` semantics during composition are unchanged by the PR (the `setValueOnChange` block already ran during composition before this change — only the harmful split is now skipped).
- **Changeset is correct.** `@ariakit/react-components` patch only matches the established TagInput-fix convention (`TagInput` is exposed solely via the `@ariakit/react-components/tag/tag-input` subpath, not the public `@ariakit/react` entrypoint, so it should not be listed; the changeset-bot's `@ariakit/react` entry is the automatic dependent bump, which is expected). Wording and `300-` prefix match recent entries.
- **`pnpm tsc` passes** (the `isInputEvent` guard narrows correctly); **oxlint** is clean on all changed files.

### Notes
- A happy-dom `test.ts` duplicate is not feasible here — the repro depends on Chromium's real CDP IME pipeline (`Input.imeSetComposition`/`Input.insertText`), which happy-dom cannot drive. Browser-only coverage is the right call.
- One optional, non-blocking simplification posted inline: the new `isInputEvent` helper duplicates `combobox.tsx`'s identical copy and could be hoisted into `@ariakit/utils` next to `getInputType`.

packages/ariakit-react-components/src/tag/tag-input.tsx
diegohaz: Fixed in fd8f7bc03 - hoisted `isInputEvent` into `@ariakit/utils`, reused it from `Combobox` and `TagInput`, added utility coverage, and regenerated the utils API docs.

diegohaz:
**Fixed**
- `fd8f7bc03` ([reply](https://github.com/ariakit/ariakit/pull/6420#discussion_r3463539927)): Hoisted `isInputEvent` into `@ariakit/utils`, removed the duplicate component-local guards, added utility coverage, and updated the generated utils API docs.

app/src/sandbox/tag-input-6294/test-browser.ts
```diff
@@ -0,0 +1,32 @@
...
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("does not split IME composition text into tags", async ({
+    browserName,
+    page,
+    q,
+  }) => {
+    test.skip(
+      browserName !== "chromium",
+      "CDP IME composition is Chromium-only.",
+    );
```
diegohaz: Why not just name the file `test-chrome.ts`?

diegohaz: Fixed in 2133491cb - renamed the IME sandbox test to `test-chrome.ts` and removed the Chromium-only runtime skip so Playwright scopes the test by filename.

diegohaz:
**Fixed**
- `2133491cb` ([reply](https://github.com/ariakit/ariakit/pull/6420#discussion_r3463587079)): Renamed the Chromium-only IME sandbox test to `test-chrome.ts` and removed the runtime browser skip.

**Updated**
- `d3c1bd592`: Merged the latest `main` into `fix-6294` and kept both the `isInputEvent` coverage and the new `queueBeforeEvent` cancellation coverage in `events.test.ts`.

PR #6421, opened 2026-06-23
Fix CollectionRenderer measured gap padding

### LINKED ISSUE #6337: CollectionRenderer without itemSize shifts all items by one extra gap when paddingStart is set
### Current behavior

When a virtualized renderer (`CollectionRenderer`, and therefore also `CompositeRenderer` and `SelectRenderer`, which delegate to it) is used without a fixed `itemSize` — so item sizes are measured from the DOM — combining `gap` with a non-zero `padding`/`paddingStart` positions every item one `gap` further down (or right, when horizontal) than it should be, and inflates the container's total size by one `gap`.

For example, with `gap={10}`, `padding={20}`, and 8 items of 32px each, the first item gets `top: 30px` instead of `20px`, the second `72px` instead of `62px`, and the container height is `376px` instead of `366px`. An otherwise identical renderer with an explicit `itemSize={32}` lays out correctly, so the two configurations disagree even though they describe the same geometry. The mismatch also shows up as a one-`gap` size jump when DOM measurement replaces the initial estimate, since the estimated total budgets exactly `gap × (length - 1)` gaps.

The root cause is in `getData`, the measured-size path. The running offset is seeded with `paddingStart`, and the inter-item gap is applied based on the *truthiness* of that offset:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L385-L407

```ts
let start = props.paddingStart;
// ...
const setSize = (size: number, rendered = prevRendered) => {
  start = start ? start + props.gap : start;
  // ...
};
```

`start = start ? start + props.gap : start` uses "offset is still 0" as a proxy for "first item, no gap yet". That proxy only holds when `paddingStart` is 0. With a non-zero `paddingStart`, the first iteration enters `setSize` with a truthy `start`, so the first item is placed at `paddingStart + gap` and every later item inherits the shift.

This contradicts the fixed-size formula used elsewhere in the same file, which never places a gap before the first item — `paddingStart + itemSize * index + gap * index` in [`getItemProps`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L782-L784) and [`getItemOffset`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L578-L588) — and [`getItemsEnd`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L354-L364), whose estimated branch budgets `lastIndex * gap` gaps but whose measured branch trusts the shifted `lastItemData.end`. It also contradicts the documented props: `paddingStart` is "the padding between the items and the container's start edge", and `gap` is "the gap between each item".

The line is unchanged since the original experimental commit ([0eff54570](https://github.com/ariakit/ariakit/commit/0eff54570), #2553).

3. Compare the two lists: in the "Measured" list, the first item sits 30px below the container's top edge and the container is 376px tall; in the "Fixed" control list rendering the same items with the same `gap`/`padding`, the first item sits at 20px and the container is 366px tall.

```tsx
import { CollectionRenderer } from "@ariakit/react-components/collection/collection-renderer";

const GAP = 10;
const PADDING = 20;
const ITEM_HEIGHT = 32;

const measuredItems = Array.from({ length: 8 }, (_, index) => ({
  id: `measured-item-${index + 1}`,
}));

const fixedItems = Array.from({ length: 8 }, (_, index) => ({
  id: `fixed-item-${index + 1}`,
}));

const scrollerStyle = {
  width: 200,
  height: 200,
  overflowY: "auto",
  border: "1px solid gray",
} as const;

const itemStyle = {
  boxSizing: "border-box",
  display: "block",
  width: "100%",
  height: ITEM_HEIGHT,
  background: "#eee",
} as const;

  return (
    <div style={{ display: "flex", gap: 16 }}>
      <section>
        <h2>Measured (no itemSize)</h2>
        <div style={scrollerStyle}>
          <CollectionRenderer items={measuredItems} gap={GAP} padding={PADDING}>
            {(item) => (
              <button
                key={item.id}
                id={item.id}
                ref={item.ref}
                style={{ ...itemStyle, ...item.style }}
                type="button"
              >
                Measured {item.index + 1}
              </button>
            )}
          </CollectionRenderer>
        </div>
      </section>
      <section>
        <h2>Fixed (itemSize)</h2>
        <div style={scrollerStyle}>
          <CollectionRenderer
            items={fixedItems}
            itemSize={ITEM_HEIGHT}
            gap={GAP}
            padding={PADDING}
          >
            {(item) => (
              <button
                key={item.id}
                id={item.id}
                ref={item.ref}
                style={{ ...itemStyle, ...item.style }}
                type="button"
              >
                Fixed {item.index + 1}
              </button>
            )}
          </CollectionRenderer>
        </div>
      </section>
    </div>
  );
}
```

This was verified with a Playwright test asserting the items' computed `top` and the containers' heights: the fixed-size list passes (`20px`/`62px`/`366px`) while the measured list yields `30px`/`72px`/`376px`.

### Expected behavior

Both lists should lay out identically. With no `itemSize`, measured item offsets should follow the same convention as the fixed-size formula `paddingStart + itemSize * index + gap * index`: the first item at exactly `paddingStart`, one `gap` between adjacent items (no gap before the first one), and a total size of `paddingStart + sum(sizes) + gap * (length - 1) + paddingEnd`.

### Workaround

Subtract the gap from the start padding on measured renderers, so the extra gap added by the bug lands the first item at the intended offset:

```tsx
<CollectionRenderer
  items={items}
  gap={GAP}
  paddingStart={PADDING - GAP}
  paddingEnd={PADDING}
>
```

This restores both the item offsets and the total size (verified with the same Playwright test). It only works while `paddingStart > gap`: if the compensated `paddingStart` is 0, the truthiness check skips the gap for the first item entirely and it renders at 0 instead of the intended offset. Alternatively, when all items have a known fixed size, passing an explicit `itemSize` bypasses the measured path altogether.

### Possible solutions

In `getData`'s `setSize`, apply the gap based on the item index instead of the offset's truthiness. `index` is the loop variable, already in scope, and `setSize` runs exactly once per iteration, so `index > 0` is exactly "not the first item":

```diff
     const setSize = (size: number, rendered = prevRendered) => {
-      start = start ? start + props.gap : start;
+      if (index > 0) {
+        start += props.gap;
+      }
       const end = start + size;
```

Measured offsets then become `paddingStart + sum(sizes[0..index-1]) + gap * index`, matching the fixed-size formula in `getItemProps`/`getItemOffset`, and the last item's `end` matches `getItemsEnd`'s `lastIndex * gap` budget, eliminating the post-measurement size jump. As a side effect, a zero-sized item at offset 0 no longer suppresses the next item's gap, which aligns with the "exactly `length - 1` gaps" model used everywhere else in the file.

diegohaz:
## Motivation

`CollectionRenderer` added one extra `gap` before the first measured item when `paddingStart` was set and `itemSize` was omitted.

With `gap={10}`, `padding={20}`, and measured 32px items, the measured renderer placed the first item at `30px` instead of matching the fixed-size renderer at `20px`. This also inflated the total renderer size by one `gap`, and the same measured path is used by `CompositeRenderer` and `SelectRenderer`.

Fixes #6337.

## Solution

Add a focused sandbox and browser test that compare a measured `CollectionRenderer` against an otherwise identical fixed-size `CollectionRenderer`.

Then change the measured data path to add `gap` based on the item index instead of the current offset. `index > 0` means the item is not the first item, so the measured path now follows the same convention as the fixed-size formula:

```ts
paddingStart + itemSize * index + gap * index
```

The sandbox workaround from the intermediate commit has been removed, so the regression test now passes through the library fix.

## Workaround

Until this fix is released, subtract the gap from the measured renderer's start padding and keep the original end padding:

```tsx
// https://github.com/ariakit/ariakit/issues/6337 is fixed.
<CollectionRenderer
  items={items}
  gap={GAP}
  paddingStart={PADDING - GAP}
  paddingEnd={PADDING}
>
  {renderItem}
</CollectionRenderer>
```

This compensates for the extra first-item gap in the measured path. It only works when `PADDING - GAP` stays greater than `0`; if all item sizes are known, passing `itemSize` bypasses the measured path instead.

### DIFF
--- a/.changeset/300-collection-renderer-gap-padding.md
+++ b/.changeset/300-collection-renderer-gap-padding.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Fixed `CollectionRenderer`, `CompositeRenderer`, and `SelectRenderer` to position measured virtualized items correctly when `gap` is combined with `paddingStart` or `padding`.
--- a/app/src/sandbox/collection-renderer-6337/index.react.tsx
+++ b/app/src/sandbox/collection-renderer-6337/index.react.tsx
@@ -1,0 +1,94 @@
+import { CollectionRenderer } from "@ariakit/react-components/collection/collection-renderer";
+
+const GAP = 10;
+const PADDING = 20;
+const ITEM_HEIGHT = 32;
+
+const measuredItems = Array.from({ length: 8 }, (_, index) => ({
+  id: `measured-item-${index + 1}`,
+}));
+
+const fixedItems = Array.from({ length: 8 }, (_, index) => ({
+  id: `fixed-item-${index + 1}`,
+}));
+
+const wrapperStyle = {
+  display: "flex",
+  gap: 16,
+} as const;
+
+const sectionStyle = {
+  display: "grid",
+  gap: 8,
+} as const;
+
+const scrollerStyle = {
+  width: 200,
+  height: 200,
+  overflowY: "auto",
+  border: "1px solid gray",
+} as const;
+
+const itemStyle = {
+  boxSizing: "border-box",
+  display: "block",
+  width: "100%",
+  height: ITEM_HEIGHT,
+  background: "#eee",
+} as const;
+
+export default function Example() {
+  return (
+    <div style={wrapperStyle}>
+      <section style={sectionStyle}>
+        <h2>Measured</h2>
+        <div style={scrollerStyle}>
+          <CollectionRenderer
+            id="measured-renderer"
+            items={measuredItems}
+            initialItems={measuredItems.length}
+            gap={GAP}
+            padding={PADDING}
+          >
+            {(item) => (
+              <button
+                key={item.id}
+                id={item.id}
+                ref={item.ref}
+                style={{ ...itemStyle, ...item.style }}
+                type="button"
+              >
+                Measured {item.index + 1}
+              </button>
+            )}
+          </CollectionRenderer>
+        </div>
+      </section>
+      <section style={sectionStyle}>
+        <h2>Fixed</h2>
+        <div style={scrollerStyle}>
+          <CollectionRenderer
+            id="fixed-renderer"
+            items={fixedItems}
+            initialItems={fixedItems.length}
+            itemSize={ITEM_HEIGHT}
+            gap={GAP}
+            padding={PADDING}
+          >
+            {(item) => (
+              <button
+                key={item.id}
+                id={item.id}
+                ref={item.ref}
+                style={{ ...itemStyle, ...item.style }}
+                type="button"
+              >
+                Fixed {item.index + 1}
+              </button>
+            )}
+          </CollectionRenderer>
+        </div>
+      </section>
+    </div>
+  );
+}
--- a/app/src/sandbox/collection-renderer-6337/index.react.tsx
+++ b/app/src/sandbox/collection-renderer-6337/index.react.tsx
@@ -38,6 +38,11 @@ const itemStyle = {
 } as const;
 
 export default function Example() {
+  // TODO: Remove this workaround once issue #6337 is fixed.
+  // https://github.com/ariakit/ariakit/issues/6337
+  // The measured path currently adds GAP before the first item when
+  // paddingStart is set, so subtract GAP to preserve the expected visual
+  // padding.
   return (
     <div style={wrapperStyle}>
       <section style={sectionStyle}>
@@ -48,7 +53,8 @@ export default function Example() {
             items={measuredItems}
             initialItems={measuredItems.length}
             gap={GAP}
-            padding={PADDING}
+            paddingStart={PADDING - GAP}
+            paddingEnd={PADDING}
           >
             {(item) => (
               <button
--- a/app/src/sandbox/collection-renderer-6337/index.react.tsx
+++ b/app/src/sandbox/collection-renderer-6337/index.react.tsx
@@ -38,11 +38,6 @@ const itemStyle = {
 } as const;
 
 export default function Example() {
-  // TODO: Remove this workaround once issue #6337 is fixed.
-  // https://github.com/ariakit/ariakit/issues/6337
-  // The measured path currently adds GAP before the first item when
-  // paddingStart is set, so subtract GAP to preserve the expected visual
-  // padding.
   return (
     <div style={wrapperStyle}>
       <section style={sectionStyle}>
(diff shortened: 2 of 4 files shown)


diegohaz (commented):
**Review summary** — initial full review, comment-only. No actionable findings.

**Scope.** Diff from merge-base `57dfad977` to head `f9dbbff1a` (4 files, +125/-1): the one-line core fix in `packages/ariakit-react-components/src/collection/collection-renderer.tsx`, the `collection-renderer-6337` sandbox + browser test, and the changeset.

**Assessment.** The fix is correct and minimal. In `getData`'s `setSize`, swapping the running-offset truthiness check (`start = start ? start + gap : start`) for `if (index > 0) start += gap` makes the measured path add `gap` per interim item exactly like the fixed-size path's canonical formula `paddingStart + itemSize * index + gap * index` (see the fixed-size offset at `collection-renderer.tsx:582` and `:785`). The previous check added a spurious leading gap whenever `paddingStart` was truthy (and conversely could drop an interim gap when a zero-size run kept `start` at `0`); keying on `index` removes both edge cases. `CompositeRenderer` and `SelectRenderer` both route through `useCollectionRenderer`, so the changeset's claim that all three are fixed is accurate.

**Validation.**
- Simulated old vs. new measured math with `gap=10`, `padding=20`, `itemSize=32`, 8 items: old → item 0 at `30px`, total `376px`; fixed → item 0 at `20px`, item 1 at `62px`, total `366px` — matching the browser test's hardcoded expectations exactly.
- Confirmed the measured renderer drives each item's `top` from `data.get(itemId)?.start` (`collection-renderer.tsx:786`), so the test's `toHaveCSS("top", …)` and `#measured-renderer` height assertions genuinely exercise the fixed offsets.
- Confirmed the final head has no leftover sandbox workaround (the intermediate `paddingStart - gap` was reverted; the sandbox uses `padding={PADDING}` directly), so the regression test passes through the library fix rather than around it.
- CI on this head is green for the relevant jobs (Type check, Lint, Build, Test, Test React / React 18 / Solid, Test Chrome / Firefox / macOS / Android); the added browser test passes across engines.

**Notes (not blocking).** A happy-dom `test.ts` duplicate is intentionally omitted — the behavior depends on real layout measurement (`getBoundingClientRect`/computed `top`), which doesn't reproduce in happy-dom, so browser-only coverage is the right call here. Changeset scope and patch level are appropriate.

PR #6422, opened 2026-06-23
Fix merged store init reentrancy

### LINKED ISSUE #6313: Reentrant setState during a merged store's init push is lost or overwritten, silently desyncing multi-parent stores
### Current behavior

When a store extends two or more parent stores with partially-overlapping keys (the normal shape of every composed Ariakit store: `createSelectStore` builds `createStore(initialState, composite, popover, store)` and merges parents via [`mergeStore`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/select/select-store.ts#L43-L56)), `storeInit` wires each parent through a single grouped `sync` listener instead of one listener per key:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L196-L211

This grouped fast path has a reentrancy hole. `storeSync` invokes the listener once with a single state snapshot and registers it in the listener group only *after* that initial call returns:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L301-L305

So during the initial push, the grouped listener iterates a snapshot captured before the loop started. If a `sync` listener on the child reacts to one of the pushed keys by writing another key of the same parent (directly or via fan-out), that write fails in both directions: an update to an already-pushed key is never delivered (the grouped listener isn't registered on the parent yet), and an update to a not-yet-pushed key is overwritten when the loop reaches it with the stale snapshot value. The two stores then silently disagree about that key until something else happens to write it.

The single-parent keyed path handles the identical scenario correctly, and that invariant is pinned by the package's own tests ([`test.ts#L564-L605`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/test.ts#L564-L605)) — but only for one parent. Multi-parent mirrors of those exact tests fail on current code:

```text
[multi-parent grouped later-key]   first.b = updated-b | child.b = first-b   <-- DESYNCED
[single-parent keyed later-key]    first.b = updated-b | child.b = updated-b (correct)
[multi-parent grouped earlier-key] first.a = updated-a | child.a = first-a   <-- DESYNCED
```

This is reachable from a React composition. In the repro below, a select store is hoisted (`useSelectStore`) and passed to a [`SelectProvider`](https://ariakit.com/reference/select-provider) nested in a [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider), so `createSelectStore` merges the hoisted store and the combobox store as partially-overlapping parents — the grouped path. A descendant component registers a `sync` listener (from `@ariakit/store`) on the provider's select store in a layout effect; descendant layout effects run before the provider's own `init` layout effect, so the listener is live during the initial push. On mount, the push delivers the hoisted store's `activeId` ("Banana", from `defaultActiveId`), the listener commits it through the documented `setValue`, the write fans out to the hoisted store — and then the push loop overwrites the merged store with the stale snapshot value ("Apple"). Result on page load, with no interaction: the hoisted store holds `value: "Banana"` while the `Select` button (backed by the provider store) renders "Apple". Every consumer of one store now disagrees with every consumer of the other.

Note: this cannot be triggered with `@ariakit/react` APIs alone — [`useStoreState`](https://ariakit.com/reference/use-store-state) subscribes after `init` (passive effect), and controlled props register their listeners after `init` in the same component. The trigger requires a store-level listener from the lower-level published `@ariakit/store` package (or any code registering a listener on the store before it initializes). The underlying defect is entirely in `@ariakit/store`, and any standalone consumer of `createStore`/`mergeStore` + `sync` + `init` hits it with no React involved.

4. Load the page (no further interaction needed).
5. Note the line under the select reads "Committed value: Banana" (the hoisted store), while the select button itself reads "Apple" (the provider store). The two stores are permanently out of sync.

```tsx
import { sync } from "@ariakit/store";
import { useLayoutEffect, useMemo, useState } from "react";

const list = ["Apple", "Banana", "Cherry", "Grape", "Lemon", "Orange"];

// Custom "value follows the highlighted item" plumbing: whenever an item is
// highlighted, commit it as the select value. Items render with their value
// as the DOM id, so the activeId state is the item value.
function ValueFollowsHighlight() {
  const select = Ariakit.useSelectContext();
  useLayoutEffect(() => {
    if (!select) return;
    return sync(select, ["activeId"], (state) => {
      if (!state.activeId) return;
      select.setValue(state.activeId);
    });
  }, [select]);
  return null;
}

  // Hoisted so the page can react to the value outside the widget. Banana is
  // highlighted initially, so the plumbing above should commit "Banana".
  const select = Ariakit.useSelectStore({
    defaultValue: "Apple",
    defaultActiveId: "Banana",
  });
  const value = Ariakit.useStoreState(select, "value");

  const matches = useMemo(() => {
    const search = searchValue.toLowerCase();
    return list.filter((item) => item.toLowerCase().includes(search));
  }, [searchValue]);

  return (
    <>
      <Ariakit.ComboboxProvider setValue={setSearchValue}>
        <Ariakit.SelectProvider store={select}>
          <ValueFollowsHighlight />
          <Ariakit.SelectLabel>Favorite fruit</Ariakit.SelectLabel>
          <Ariakit.Select />
          <Ariakit.SelectPopover gutter={4} sameWidth>
            <Ariakit.Combobox placeholder="Search..." />
            <Ariakit.ComboboxList>
              {matches.map((item) => (
                <Ariakit.SelectItem
                  key={item}
                  id={item}
                  value={item}
                  render={<Ariakit.ComboboxItem />}
                />
              ))}
            </Ariakit.ComboboxList>
          </Ariakit.SelectPopover>
        </Ariakit.SelectProvider>
      </Ariakit.ComboboxProvider>
      <p>
        Committed value: <output>{value}</output>
      </p>
    </>
  );
}
```

The desync was verified with a real-browser Playwright test that only loads the page and asserts on the rendered DOM.

The store-level bug also reproduces with `@ariakit/store` alone, with no React involved:

```ts
import { createStore, init, sync } from "@ariakit/store";

const first = createStore({ a: "first-a", b: "first-b" });
const second = createStore({ c: "second-c" });
const child = createStore({ a: "child-a", b: "child-b", c: "child-c" }, first, second);

sync(child, ["a"], (state) => {
  if (state.a === "first-a") first.setState("b", "updated-b");
});

init(child);
first.getState().b; // "updated-b"
child.getState().b; // "first-b" — desynced
```

### Expected behavior

Writes made by sync listeners while a merged store initializes should be preserved on both stores, matching the single-parent behavior that the package's own tests pin ("syncs reentrant parent setStates during a child's initial push"). After `init`, parent and child must agree on every shared key. Refactoring a store from one parent to two should not change observable behavior.

### Workaround

Defer any store write made from a sync listener so it cannot run in the middle of the initialization push:

```tsx
return sync(select, ["activeId"], (state) => {
  if (!state.activeId) return;
  const { activeId } = state;
  queueMicrotask(() => select.setValue(activeId));
});
```

Verified: with this change the page renders "Banana" in both places and the stores stay in sync (by the time the microtask runs, the wiring is complete, so the write propagates everywhere).

### Possible solutions

Make the grouped path satisfy the same invariant as the keyed path, in both directions:

1. Register the grouped listener with `subscribe` (same listener group, no initial call) so it is live before any key is pushed, instead of `sync`, which only registers the listener after the initial call returns.
2. Perform the initial push manually, one key at a time, reading the live parent state on every iteration instead of a single snapshot, so the push can't overwrite a value the now-registered listener just delivered.

```diff
-      let didSyncInitialState = false;
       desyncs.push(
-        sync(store, keys, (state, prevState) => {
+        subscribe(store, keys, (state, prevState) => {
           for (const key of keys) {
-            if (didSyncInitialState && state[key] === prevState[key]) continue;
+            if (state[key] === prevState[key]) continue;
             setState(
               key,
               state[key],
               // @ts-expect-error - Not public API. This is just to prevent
               // infinite loops.
               true,
             );
           }
-          didSyncInitialState = true;
         }),
       );
+      // Push the parent's state down one key at a time, reading the live
+      // state on every iteration. A child sync listener may reentrantly
+      // update the parent while an earlier key is being pushed: the listener
+      // registered above already delivered that update, so re-pushing a
+      // stale snapshot here would overwrite it and desync the stores.
+      for (const key of keys) {
+        const liveState = store?.getState?.();
+        if (!liveState) continue;
+        setState(
+          key,
+          liveState[key],
+          // @ts-expect-error - Not public API. This is just to prevent
+          // infinite loops.
+          true,
+        );
+      }
```

Both pieces are required: fixing only the snapshot read still loses reentrant updates to already-pushed keys, because `sync` registers the listener only after its initial call. This patch was verified against this commit: all 43 existing `@ariakit/store` tests pass, the store-level desync cases above come back in sync, and the browser reproduction renders consistently. Multi-parent mirrors of the two existing reentrancy tests in `packages/ariakit-store/src/test.ts` would pin the fix.

diegohaz:
## Motivation

Issue #6313 reports that reentrant writes made while a child store initializes from multiple partially-overlapping parent stores can be lost or overwritten. This can leave the child and parent stores with different values after the initial push.

A real composition can hit this when a hoisted `Select` store is passed to a `SelectProvider` inside a `ComboboxProvider`, and a descendant store-level `sync` listener writes `activeId` back to `value` during provider initialization.

Fixes #6313.

## Solution

The grouped multi-parent init path now registers a parent `subscribe` listener before the initial parent-to-child push. It then pushes each shared key from a fresh parent state read, so child `sync` listeners that reentrantly write back to the parent are preserved instead of being overwritten by a stale snapshot.

This keeps the grouped listener optimization while matching the existing single-parent reentrancy behavior.

The PR also adds coverage for the reported behavior:

- focused `@ariakit/store` tests that mirror the existing single-parent reentrancy cases against the grouped multi-parent path;
- a `Select` + `Combobox` sandbox that reproduces the user-visible mismatch between a hoisted select store and the provider store;
- matching happy-dom and browser tests for the sandbox.

## Workaround

Until this fix is available in a release, users can defer store writes made from an initialization-time `sync` listener so they run after the parent/child wiring is complete.

```tsx
return sync(select, ["activeId"], (state) => {
  if (!state.activeId) return;
  const { activeId } = state;
  // TODO: Remove this workaround when
  // https://github.com/ariakit/ariakit/issues/6313 is fixed.
  queueMicrotask(() => select.setValue(activeId));
});
```

### DIFF
--- a/.changeset/300-merged-store-init.md
+++ b/.changeset/300-merged-store-init.md
@@ -1,0 +1,9 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
+---
+
+Fixed merged stores to keep values in sync when `sync` listeners update parent
+stores during initialization, including composed stores used by
+[`Select`](https://ariakit.com/reference/select) and
+[`Combobox`](https://ariakit.com/reference/combobox).
--- a/app/package.json
+++ b/app/package.json
@@ -55,6 +55,7 @@
     "@ariakit/solid": "workspace:*",
     "@ariakit/solid-components": "workspace:*",
     "@ariakit/solid-utils": "workspace:*",
+    "@ariakit/store": "workspace:*",
     "@ariakit/tailwind": "workspace:*",
     "@ariakit/utils": "workspace:*",
     "@astrojs/check": "0.9.9",
--- a/app/src/sandbox/select-combobox-6313/index.react.tsx
+++ b/app/src/sandbox/select-combobox-6313/index.react.tsx
@@ -13,10 +13,7 @@ function ValueFollowsHighlight() {
     if (!select) return;
     return sync(select, ["activeId"], (state) => {
       if (!state.activeId) return;
-      const { activeId } = state;
-      // TODO: Remove this workaround when
-      // https://github.com/ariakit/ariakit/issues/6313 is fixed.
-      queueMicrotask(() => select.setValue(activeId));
+      select.setValue(state.activeId);
     });
   }, [select]);
 
--- a/app/src/sandbox/select-combobox-6313/index.react.tsx
+++ b/app/src/sandbox/select-combobox-6313/index.react.tsx
@@ -13,7 +13,10 @@ function ValueFollowsHighlight() {
     if (!select) return;
     return sync(select, ["activeId"], (state) => {
       if (!state.activeId) return;
-      select.setValue(state.activeId);
+      const { activeId } = state;
+      // TODO: Remove this workaround when
+      // https://github.com/ariakit/ariakit/issues/6313 is fixed.
+      queueMicrotask(() => select.setValue(activeId));
     });
   }, [select]);
 
--- a/app/src/sandbox/select-combobox-6313/index.react.tsx
+++ b/app/src/sandbox/select-combobox-6313/index.react.tsx
@@ -1,0 +1,65 @@
+import * as Ariakit from "@ariakit/react";
+import { sync } from "@ariakit/store";
+import { useLayoutEffect, useMemo, useState } from "react";
+
+const list = ["Apple", "Banana", "Cherry", "Grape", "Lemon", "Orange"];
+
+// The descendant layout effect runs before SelectProvider initializes, so this
+// store-level listener exposes reentrant writes during the initial parent push.
+function ValueFollowsHighlight() {
+  const select = Ariakit.useSelectContext();
+
+  useLayoutEffect(() => {
+    if (!select) return;
+    return sync(select, ["activeId"], (state) => {
+      if (!state.activeId) return;
+      select.setValue(state.activeId);
+    });
+  }, [select]);
+
+  return null;
+}
+
+export default function Example() {
+  const [searchValue, setSearchValue] = useState("");
+  // The highlighted item should be committed during init so both stores render
+  // Banana after the first paint.
+  const select = Ariakit.useSelectStore({
+    defaultValue: "Apple",
+    defaultActiveId: "Banana",
+  });
+  const value = Ariakit.useStoreState(select, "value");
+
+  const matches = useMemo(() => {
+    const search = searchValue.toLowerCase();
+    return list.filter((item) => item.toLowerCase().includes(search));
+  }, [searchValue]);
+
+  return (
+    <>
+      <Ariakit.ComboboxProvider setValue={setSearchValue}>
+        <Ariakit.SelectProvider store={select}>
+          <ValueFollowsHighlight />
+          <Ariakit.SelectLabel>Favorite fruit</Ariakit.SelectLabel>
+          <Ariakit.Select />
+          <Ariakit.SelectPopover gutter={4} sameWidth>
+            <Ariakit.Combobox placeholder="Search..." />
+            <Ariakit.ComboboxList>
+              {matches.map((item) => (
+                <Ariakit.SelectItem
+                  key={item}
+                  id={item}
+                  value={item}
+                  render={<Ariakit.ComboboxItem />}
+                />
+              ))}
+            </Ariakit.ComboboxList>
+          </Ariakit.SelectPopover>
+        </Ariakit.SelectProvider>
+      </Ariakit.ComboboxProvider>
+      <p>
+        Committed value: <output>{value}</output>
+      </p>
+    </>
+  );
+}
--- a/app/src/sandbox/select-combobox-6313/test-browser.ts
+++ b/app/src/sandbox/select-combobox-6313/test-browser.ts
@@ -1,0 +1,11 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// See https://github.com/ariakit/ariakit/issues/6313
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("keeps hoisted and provider select values in sync after init", async ({
+    q,
+  }) => {
+    await test.expect(q.combobox("Favorite fruit")).toHaveText("Banana");
+    await test.expect(q.text("Committed value: Banana")).toBeVisible();
+  });
+});
--- a/app/src/sandbox/select-combobox-6313/test-browser.ts
+++ b/app/src/sandbox/select-combobox-6313/test-browser.ts
@@ -6,6 +6,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
     q,
   }) => {
     await test.expect(q.combobox("Favorite fruit")).toHaveText("Banana");
-    await test.expect(q.text("Committed value: Banana")).toBeVisible();
+    await test.expect(q.status()).toHaveText("Banana");
   });
 });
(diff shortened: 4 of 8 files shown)


diegohaz (commented):
Reviewed this PR end to end. Initial (full) review, baseline `merge-base(57dfad97, head)`; head `62a12e6a`. No actionable findings — the change is correct, minimal, and well-covered, so there are no inline comments.

## Scope reviewed

- `packages/ariakit-store/src/index.ts` — the grouped multi-parent init path now registers a parent `subscribe` listener before the initial push, then pushes each shared key by reading the live parent state per iteration.
- `packages/ariakit-store/src/test.ts` — two new multi-parent reentrancy tests mirroring the existing single-parent ones.
- `app/src/sandbox/select-combobox-6313/*` — happy-dom + browser repro of the hoisted-`Select` vs provider-store desync.
- `.changeset/300-merged-store-init.md`, `app/package.json`, `pnpm-lock.yaml`.

## Why the fix is correct

The grouped path previously used `sync`, which registers the listener only *after* its initial synchronous call. So a child `sync` listener that reentrantly wrote back to the parent during the initial push had its write lost (already-pushed key: listener not yet registered) or overwritten (later key: stale snapshot). Switching to `subscribe` (register-then-push) plus a per-key live re-read closes both directions. The per-iteration `store.getState()` re-read is load-bearing, not redundant: the now-registered grouped listener can advance the parent's other keys mid-loop, and a single pre-loop snapshot would clobber them. For the non-reentrant case the behavior is preserved — every shared key is still pushed from current parent state, just without `sync`'s immediate-call shortcut. `setState(..., true)` keeps `fromStores=true`, so the manual push doesn't re-fan-out to parents. This matches the single-parent keyed-path invariant the package already pins.

## Validation performed (Node 24, this worktree)

- `pnpm install`, then `vitest run --root packages/ariakit-store`: 45/45 pass (43 prior + 2 new).
- Reverted just `index.ts` to `origin/main` and re-ran: both new multi-parent tests fail (`child.b`/`child.a` keep the stale snapshot value), and the sandbox happy-dom test fails rendering "Apple" instead of "Banana". Restored the file; worktree clean. Confirms the tests pin the fix and the sandbox reproduces the reported user-visible desync.
- `ARIAKIT_TEST_LOADER=react vitest run select-combobox-6313`: passes on head, fails on base.
- `tsc -p . --noEmit`: clean.
- Perf bot on the PR shows no significant Chrome/Node change (`initialize merged store chain` -1%, within noise).

## Notes (no action needed)

- The changeset patches `@ariakit/store` + `@ariakit/solid-store`; correct, since `@ariakit/solid-store` is a pure re-export of `@ariakit/store`. The other 5 packages the changeset bot lists cascade through `workspace:*`.
- No Solid sandbox repro is added, but the defect is entirely in the framework-agnostic store and is already pinned by the `@ariakit/store` unit tests, so a Solid integration repro would be redundant.
- Cross-parent reentrancy during init (a listener writing to a *different*, not-yet-wired parent earlier in `stores`) remains a pre-existing limitation shared by the keyed/single-parent paths; this PR is correctly scoped to within-parent reentrancy and does not regress it.

PR #6423, opened 2026-06-23
Fix queueBeforeEvent cancel cleanup

### LINKED ISSUE #6350: @ariakit/utils: queueBeforeEvent's returned cancel function doesn't remove the event listener, so a canceled callback still runs
### Current behavior

`queueBeforeEvent(element, type, callback, timeout?)` from `@ariakit/utils` schedules `callback` to run on the next animation frame (or after `timeout`), but earlier if `type` fires on `element` first. It arms two triggers — a timer and a capture-phase event listener — and each trigger correctly disarms the other when it fires. But the value handed back to the caller is `cancelTimer` itself, which only clears the rAF/timeout and never removes the listener:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/events.ts#L193-L205

So if a consumer calls the returned cancel function while both triggers are pending, only the timer path is torn down. The `callSync` listener stays attached, and the supposedly canceled callback still runs the next time `type` fires on the element (`once: true` bounds it to a single stale invocation). Until that event fires, the listener also keeps the callback closure alive.

The only purpose of the return value is cancellation: the function is part of the published `@ariakit/utils` API via the barrel export ([`index.ts#L3`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/index.ts#L3)) and the package readme documents the `() => void` return ([`readme.md#L772-L783`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/readme.md#L772-L783)). The failure mode is nasty precisely because it usually works: when the event never fires again, cancellation appears successful, and the callback only misfires when the named event happens to fire later — a hard-to-reproduce "callback ran after I canceled it" bug.

No in-repo caller is affected (all six call sites in `@ariakit/react-components` discard the return value), so this only bites direct consumers of `@ariakit/utils` — or any future internal caller that wants to cancel, e.g. in an effect cleanup.

3. Click "Show shortcuts". A "Keyboard shortcuts" hint appears, and the app schedules an auto-hide to run right before the next `mousedown` in the app (or after 30 seconds) via `queueBeforeEvent`.
4. Move the pointer over the hint card. The app calls the cancel function returned by `queueBeforeEvent` to pin the hint, and the status reads "pinned".
5. Click "New document". The hint disappears and the status flips to "auto-hidden" — the canceled hide callback ran anyway, triggered by the capture-phase `mousedown` listener that the cancel function never removed.

```tsx
import { queueBeforeEvent } from "@ariakit/utils";

  const containerRef = useRef<HTMLDivElement>(null);
  const cancelAutoHideRef = useRef<(() => void) | null>(null);
  const [hintVisible, setHintVisible] = useState(false);
  const [status, setStatus] = useState("hidden");

  const showShortcuts = () => {
    const container = containerRef.current;
    if (!container) return;
    setHintVisible(true);
    setStatus("visible");
    // Get the hint out of the user's way: hide it right before their next
    // press anywhere in the app (before the respective React events fire), or
    // after 30 seconds, whichever comes first.
    cancelAutoHideRef.current = queueBeforeEvent(
      container,
      "mousedown",
      () => {
        cancelAutoHideRef.current = null;
        setHintVisible(false);
        setStatus("auto-hidden");
      },
      30_000,
    );
  };

  const pinHint = () => {
    const cancelAutoHide = cancelAutoHideRef.current;
    if (!cancelAutoHide) return;
    cancelAutoHideRef.current = null;
    // The user moved the pointer over the hint, so they are reading it.
    // Cancel the scheduled auto-hide to keep the hint up.
    cancelAutoHide();
    setStatus("pinned");
  };

  return (
    <div ref={containerRef}>
      <Ariakit.Button onClick={showShortcuts}>Show shortcuts</Ariakit.Button>
      <Ariakit.Button>New document</Ariakit.Button>
      {hintVisible && (
        <section aria-label="Keyboard shortcuts" onMouseEnter={pinHint}>
          <p>
            Press <kbd>Ctrl</kbd>+<kbd>N</kbd> to create a new document.
          </p>
        </section>
      )}
      <output>{status}</output>
    </div>
  );
}
```

The misfire was verified with a real-browser Playwright test that only clicks, hovers, and asserts on the rendered DOM.

### Expected behavior

Calling the function returned by `queueBeforeEvent` should fully cancel the queued callback: neither the timer nor the event should trigger it afterwards. In the repro, after pinning the hint, clicking "New document" should leave the hint visible and the status at "pinned".

### Workaround

Guard the callback with a consumer-side flag and wrap the returned cancel function:

```tsx
let canceled = false;
const cancelAutoHide = queueBeforeEvent(
  container,
  "mousedown",
  () => {
    if (canceled) return;
    cancelAutoHideRef.current = null;
    setHintVisible(false);
    setStatus("auto-hidden");
  },
  30_000,
);
cancelAutoHideRef.current = () => {
  canceled = true;
  cancelAutoHide();
};
```

Verified: with this change the hint stays visible (and the status stays "pinned") past the next press.

### Possible solutions

Return a function that tears down both triggers instead of returning `cancelTimer` directly. `removeEventListener(type, callSync, true)` matches the `{ once: true, capture: true }` registration (listener identity is keyed on the capture flag only; `once` does not participate in matching) — the same call the timer path already uses at [`events.ts#L194`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/events.ts#L194):

```diff
   // By listening to the event in the capture phase, we make sure the callback
   // is fired before the respective React events.
   element.addEventListener(type, callSync, { once: true, capture: true });
-  return cancelTimer;
+  return () => {
+    cancelTimer();
+    element.removeEventListener(type, callSync, true);
+  };
 }
```

Every late-call path is a safe no-op: after the timer fires, the listener has already been removed and `cancelAnimationFrame`/`clearTimeout` on a fired id is harmless; after the event fires, `once: true` has auto-removed the listener and `callSync` has already canceled the timer. The public signature is unchanged (still `() => void`), no existing caller captures the return value, and the function requires an `Element`, so there is no SSR concern — a non-breaking patch-level fix. `packages/ariakit-utils/src/events.test.ts` currently has no `queueBeforeEvent` coverage, so a unit test using the deterministic `timeout` parameter (cancel, dispatch the event, advance time, assert the callback never ran) would pin the fix.

diegohaz:
## Motivation

`queueBeforeEvent` returns a cancel function that should prevent the queued callback from running. Before this change, canceling cleared only the timer path and left the capture-phase event listener installed, so a later event could still run the supposedly canceled callback.

Fixes #6350.

## Solution

The repro adds a sandbox that mirrors the reported user interaction, a browser test for the rendered behavior, a happy-dom duplicate for faster feedback, and package-level regression coverage for both the animation-frame and timeout scheduling paths.

The fix updates `queueBeforeEvent` to return a cancel function that clears the queued timer and removes the pending capture-phase event listener. The sandbox workaround was reverted so the repro now exercises the library fix directly.

## Workaround

Until this fix is released, guard the queued callback with a consumer-side flag and wrap the returned cancel function so canceling flips the guard before calling the current `queueBeforeEvent` cancel function:

```tsx
let canceled = false;
const cancelAutoHide = queueBeforeEvent(
  container,
  "mousedown",
  () => {
    // TODO: Remove after https://github.com/ariakit/ariakit/issues/6350 is fixed.
    if (canceled) return;
    cancelAutoHideRef.current = null;
    setHintVisible(false);
    setStatus("auto-hidden");
  },
  30_000,
);
cancelAutoHideRef.current = () => {
  canceled = true;
  cancelAutoHide();
};
```

### DIFF
--- a/.changeset/300-queue-before-event-cancel.md
+++ b/.changeset/300-queue-before-event-cancel.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Fixed `queueBeforeEvent` so the cancel function removes the pending event listener as well as the queued timer.
--- a/app/src/sandbox/queue-before-event-6350/index.react.tsx
+++ b/app/src/sandbox/queue-before-event-6350/index.react.tsx
@@ -1,0 +1,52 @@
+import * as Ariakit from "@ariakit/react";
+import { queueBeforeEvent } from "@ariakit/utils";
+import { useRef, useState } from "react";
+
+export default function Example() {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const cancelAutoHideRef = useRef<(() => void) | null>(null);
+  const [hintVisible, setHintVisible] = useState(false);
+  const [status, setStatus] = useState("hidden");
+
+  const showShortcuts = () => {
+    const container = containerRef.current;
+    if (!container) return;
+
+    setHintVisible(true);
+    setStatus("visible");
+    cancelAutoHideRef.current = queueBeforeEvent(
+      container,
+      "mousedown",
+      () => {
+        cancelAutoHideRef.current = null;
+        setHintVisible(false);
+        setStatus("auto-hidden");
+      },
+      30_000,
+    );
+  };
+
+  const pinHint = () => {
+    const cancelAutoHide = cancelAutoHideRef.current;
+    if (!cancelAutoHide) return;
+
+    cancelAutoHideRef.current = null;
+    cancelAutoHide();
+    setStatus("pinned");
+  };
+
+  return (
+    <div ref={containerRef}>
+      <Ariakit.Button onClick={showShortcuts}>Show shortcuts</Ariakit.Button>
+      <Ariakit.Button>New document</Ariakit.Button>
+      {hintVisible && (
+        <section aria-label="Keyboard shortcuts" onMouseEnter={pinHint}>
+          <p>
+            Press <kbd>Ctrl</kbd>+<kbd>N</kbd> to create a new document.
+          </p>
+        </section>
+      )}
+      <output>Status: {status}</output>
+    </div>
+  );
+}
--- a/app/src/sandbox/queue-before-event-6350/index.react.tsx
+++ b/app/src/sandbox/queue-before-event-6350/index.react.tsx
@@ -14,22 +14,16 @@ export default function Example() {
 
     setHintVisible(true);
     setStatus("visible");
-    let canceled = false;
-    const cancelAutoHide = queueBeforeEvent(
+    cancelAutoHideRef.current = queueBeforeEvent(
       container,
       "mousedown",
       () => {
-        if (canceled) return;
         cancelAutoHideRef.current = null;
         setHintVisible(false);
         setStatus("auto-hidden");
       },
       30_000,
     );
-    cancelAutoHideRef.current = () => {
-      canceled = true;
-      cancelAutoHide();
-    };
   };
 
   const pinHint = () => {
--- a/app/src/sandbox/queue-before-event-6350/index.react.tsx
+++ b/app/src/sandbox/queue-before-event-6350/index.react.tsx
@@ -14,16 +14,22 @@ export default function Example() {
 
     setHintVisible(true);
     setStatus("visible");
-    cancelAutoHideRef.current = queueBeforeEvent(
+    let canceled = false;
+    const cancelAutoHide = queueBeforeEvent(
       container,
       "mousedown",
       () => {
+        if (canceled) return;
         cancelAutoHideRef.current = null;
         setHintVisible(false);
         setStatus("auto-hidden");
       },
       30_000,
     );
+    cancelAutoHideRef.current = () => {
+      canceled = true;
+      cancelAutoHide();
+    };
   };
 
   const pinHint = () => {
--- a/app/src/sandbox/queue-before-event-6350/test-browser.ts
+++ b/app/src/sandbox/queue-before-event-6350/test-browser.ts
@@ -1,0 +1,19 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/6350
+  test("canceling a queued callback keeps it from running on the next event", async ({
+    q,
+  }) => {
+    await q.button("Show shortcuts").click();
+    const hint = q.region("Keyboard shortcuts");
+    await test.expect(hint).toBeVisible();
+
+    await hint.hover();
+    await test.expect(q.text("Status: pinned")).toBeVisible();
+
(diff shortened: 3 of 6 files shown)


diegohaz (commented):
Reviewed this PR end to end. No actionable findings — the fix, tests, sandbox, and changeset are all correct. This is a comment-only review.

**Type:** initial full review
**Baseline:** `git merge-base` = `57dfad977` (base `main`) → head `f33e6fff4`

### What the fix does
`queueBeforeEvent` previously returned `cancelTimer` directly, so canceling tore down only the rAF/timeout and left the `{ once: true, capture: true }` listener attached — the canceled callback still fired the next time the event hit the element. The fix wraps the return in a closure that also calls `element.removeEventListener(type, callSync, true)`, which is the exact removal the timer path already performs at `events.ts:194`. Listener identity is keyed on the capture flag only (`once` doesn't participate in matching), so the removal matches the registration. Every late-call path stays a safe no-op: after the timer fires the listener is already gone and `cancelAnimationFrame`/`clearTimeout` on a fired id is harmless; after the event fires `once: true` removed the listener and `callSync` already canceled the timer.

### Verification performed
- **Reverted the fix locally** and reran the new tests to confirm they are genuine regression tests, not always-green:
  - `pnpm test events.test` — both new unit tests (rAF path and timeout path) **fail** without the fix (`callback` called once after cancel), **pass** with it (6/6).
  - `pnpm test-react app/src/sandbox/queue-before-event-6350/test.ts` — happy-dom sandbox test **fails** without the fix (hint region becomes `null` after clicking "New document"), **passes** with it.
- `pnpm tsc` — 0 errors.
- Confirmed the changeset frontmatter scope (`@ariakit/utils` only) is correct: all 6 in-repo call sites (`focusable` ×2, `combobox`, `command`, `tag-list`, `select`) invoke `queueBeforeEvent(...)` as a bare statement and discard the return value, so the fix changes no component public behavior, and `queueBeforeEvent` is not re-exported by the `@ariakit/react`/`@ariakit/solid` public entrypoints.

### Notes (non-blocking, not posted inline)
- The happy-dom `test.ts` runs close to the 10s hook timeout on a cold first-compile of the `app` workspace (the `beforeEach` preview render). It's reliably well under once warm; this is test-infra cold-start, not a PR issue.
- Sandbox structure matches the issue-repro convention (browser test + happy-dom duplicate, no `preview.json` needed, folder-name title). Changeset uses the correct `300-` prefix and a plain code span (no `ariakit.com/reference` page exists for `@ariakit/utils`).

Areas not exhaustively run: full `app` browser/Playwright suite (the happy-dom duplicate exercises the same DOM path and the fix is a one-liner with deterministic unit coverage).

PR #6424, opened 2026-06-23
Fix store all-keys dispatch listeners

### LINKED ISSUE #6317: @ariakit/store: all-keys listener subscribed during a keyed dispatch misses that dispatch on the keyed fast path
### Current behavior

`runListeners` in `@ariakit/store` has two notification paths. The fast path runs when the dispatch is for a single key and no all-keys (`null` keys) listeners exist at dispatch start; it iterates only that key's listener bucket:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L340-L358

A listener registered with `null` keys is never placed in a key bucket — `registerListener` puts it solely in `group.allKeysListeners` ([index.ts#L261-L267](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L261-L267)). So if a keyed listener running inside the fast path reentrantly calls `subscribe(store, null, fn)`, the new listener lands in a collection the fast path never reads, and it silently misses the in-flight dispatch.

The slow path behaves differently: it iterates `group.listeners`, a `Set` that every listener joins on registration, and JS `Set` iteration visits entries added mid-iteration. The reentrantly registered all-keys listener passes `hasUpdatedKey` unconditionally (`if (!keys) return true;`, [index.ts#L72](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L68-L81)) and fires for the same dispatch ([index.ts#L359-L376](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L359-L376)).

The minimal store-level demonstration (the only difference between the two results is an unrelated pre-existing all-keys listener):

```ts
import { createStore, subscribe } from "@ariakit/store";

const events: string[] = [];
// subscribe(store, null, () => {}); // uncommenting this changes the result below
subscribe(store, ["count"], () => {
  events.push("keyed");
  subscribe(store, null, () => {
    events.push("all-keys registered during dispatch");
  });
});
store.setState("count", 1);
// fast path: ["keyed"]
// slow path: ["keyed", "all-keys registered during dispatch"]
```

This is not an intentional carve-out: the "listener registered during the dispatch observes the dispatch" semantic is pinned by the package's own tests for keyed listeners — "fires a re-keyed listener for the currently dispatched key" ([test.ts#L508-L525](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/test.ts#L508-L525)) and "fires a keyed listener added during the keyed fast path" ([test.ts#L527-L543](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/test.ts#L527-L543)). Only the `null`-keys variant diverges, and only on the fast path. The practical consequence is heisenbug-shaped: a lazily attached whole-store listener (logger, analytics, devtools bridge) misses exactly the event that triggered its attachment, and adding any innocuous all-keys listener elsewhere "fixes" it by forcing the slow path.

Note: this cannot be triggered with `@ariakit/react` APIs alone. [`useStoreState`](https://ariakit.com/reference/use-store-state) subscribes from passive effects (never mid-dispatch), and it always registers a `null`-keys listener regardless of the key argument ([react-store/index.tsx#L121-L127](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-store/src/index.tsx#L121-L127)), so any mounted component reading store state forces the slow path on that store. Batch flushes dispatch a `Set` of keys and also always take the slow path. The bug affects direct consumers of the published `@ariakit/store` package — explicitly supported, tested reentrancy territory.

4. Click the "Like (0)" button once. The count updates to 1, and the keyed listener attaches the whole-store activity logger during that very change — but the log still reads "No activity yet": the logger missed the 0 → 1 change.
5. Click the button again. The log now reads "1 → 2". The 0 → 1 entry is permanently missing.
6. To see the inconsistency: add `subscribe(likes, null, () => {});` at module level (any unrelated all-keys listener) and repeat — the log now correctly records "0 → 1" first.

```tsx
import { createStore, subscribe, sync } from "@ariakit/store";

// App-wide store built with Ariakit's framework-agnostic store package.
const likes = createStore({ count: 0 });

  const [count, setCount] = useState(() => likes.getState().count);
  const [log, setLog] = useState<string[]>([]);

  useEffect(() => {
    let unsubscribeLogger: (() => void) | undefined;
    let loggerAttached = false;
    // Keyed listener: keeps the rendered count in sync with the store and
    // lazily attaches the activity logger on the first change, so the page
    // only pays for logging once the user actually interacts.
    const unsubscribeCount = sync(likes, ["count"], (state) => {
      setCount(state.count);
      if (loggerAttached) return;
      if (!state.count) return;
      loggerAttached = true;
      // The logger watches the whole store (null keys), so from this point on
      // it should record every state change, including the change being
      // dispatched right now (listeners registered during a dispatch observe
      // that dispatch).
      unsubscribeLogger = subscribe(likes, null, (state, prevState) => {
        setLog((entries) => [
          ...entries,
          `${prevState.count} → ${state.count}`,
        ]);
      });
    });
    return () => {
      unsubscribeCount();
      unsubscribeLogger?.();
    };
  }, []);

  return (
    <div>
      <Ariakit.Button
        onClick={() => likes.setState("count", (count) => count + 1)}
      >
        Like ({count})
      </Ariakit.Button>
      <p>
        Activity log:{" "}
        <output>{log.length ? log.join(", ") : "No activity yet"}</output>
      </p>
    </div>
  );
}
```

The missed notification was verified with a real-browser Playwright test that only clicks the button and asserts on the rendered DOM.

### Expected behavior

A listener registered during a dispatch should observe that dispatch regardless of which internal notification path runs, matching both the slow path's live `Set`-iteration semantics and the keyed-listener behavior the package pins in its own tests ("fires a keyed listener added during the keyed fast path"). After the first click, the activity log should read "0 → 1"; after the second, "0 → 1, 1 → 2". Registering an unrelated all-keys listener elsewhere must not change when other listeners fire.

### Workaround

Pre-register a no-op all-keys listener before the first state change, so the store never takes the keyed fast path:

```ts
// Forces the slow notification path, whose live Set iteration delivers the
// in-flight dispatch to listeners registered during it.
subscribe(likes, null, () => {});
```

Verified: with this one line added at the top of the effect (or at module level), the log records "0 → 1" on the first click and "0 → 1, 1 → 2" after the second.

### Possible solutions

After the bucket loop in the fast path, run any all-keys listeners that appeared during the dispatch. `allKeysListeners` was empty (or absent) when the fast path was entered, so anything in it after the loop was necessarily registered during this dispatch; firing exactly those listeners reproduces the slow path's semantics at the cost of a single `?.size` check on the hot path's exit:

```diff
       for (const listener of keyedListeners) {
         // ...existing bucket loop...
       }
+      // All-keys listeners registered by the keyed listeners above must still
+      // observe this dispatch. allKeysListeners was empty when this fast path
+      // was entered, so anything in it now was registered during the dispatch.
+      // This matches the slow path, where Set iteration visits listeners
+      // added mid-iteration.
+      if (group.allKeysListeners?.size) {
+        for (const listener of group.allKeysListeners) {
+          const cleanup = disposables.size
+            ? disposables.get(listener)
+            : undefined;
+          if (cleanup) cleanup();
+          const result = listener(state, prevState);
+          if (result) {
+            disposables.set(listener, result);
+          } else if (cleanup) {
+            disposables.delete(listener);
+          }
+        }
+      }
       return;
     }
```

This patch was verified against this commit: all 43 existing `@ariakit/store` tests pass, the fast/slow divergence disappears, and the browser reproduction behaves correctly. There is no double-fire on later dispatches because a non-empty `allKeysListeners` forces the slow path from then on. A `null`-keys mirror of the existing "fires a keyed listener added during the keyed fast path" test in `packages/ariakit-store/src/test.ts` would pin the fix.

diegohaz:
## Motivation

A `@ariakit/store` listener registered for all keys with `subscribe(store, null, listener)` could miss the in-flight dispatch when it was added from a keyed listener during the keyed fast path.

The keyed listener observed the change, but the newly registered all-keys listener did not:

```ts
subscribe(store, ["count"], () => {
  subscribe(store, null, second);
});
store.setState("count", 1);
```

This made direct `@ariakit/store` consumers see different behavior depending on whether an unrelated all-keys listener already existed on the same store.

## Solution

`runListeners` now continues through the live listener set when all-keys listeners appear during the keyed fast path. It skips listeners already notified by the keyed bucket and applies the same all-keys/keyed filtering used by the existing slow path, so listeners registered by the new all-keys listener can also observe the in-flight dispatch.

This PR includes:

- a direct `@ariakit/store` unit repro for an all-keys listener added during the keyed fast path;
- additional unit coverage for re-keying and nested listener registrations during the recovery pass;
- a React sandbox matching the user-visible lazy logger scenario;
- browser and happy-dom sandbox tests for the reported behavior;
- a patch changeset for `@ariakit/store`.

Fixes #6317.

## Workaround

Before this fix is released, pre-register a no-op all-keys listener before the first state change. This forces the slow notification path, whose live listener iteration delivers the in-flight dispatch to all-keys listeners registered during it.

```ts
// TODO: Remove once https://github.com/ariakit/ariakit/issues/6317 is fixed.
subscribe(store, null, () => {});
```

### DIFF
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -46,6 +46,16 @@ interface ListenerGroup<S> {
   listenerKeys: WeakMap<Listener<S>, Array<keyof S> | null>;
 }
 
+interface FastPathFrame<S> {
+  group: ListenerGroup<S>;
+  keyedListeners: Set<Listener<S>>;
+  updatedKey: keyof S;
+  currentListener: Listener<S> | null;
+  notifiedListeners?: Set<Listener<S>>;
+  recoverToLive?: boolean;
+  recovering?: boolean;
+}
+
 interface StoreInternals<S = State> {
   setup: StoreSetup;
   init: StoreInit;
@@ -257,6 +267,129 @@ export function createStore<S extends State>(
     }
   };
 
+  // The keyed fast path only tracks notified listeners if it has to recover
+  // into live listener iteration. Registration/disposal hooks preserve every
+  // active frame for the listener group before re-keying mutates its bucket.
+  const fastPathFrames: Array<FastPathFrame<S>> = [];
+
+  const getFastPathNotifiedListeners = (frame: FastPathFrame<S>) => {
+    const notifiedListeners = new Set<Listener<S>>();
+    const currentListener = frame.currentListener;
+    if (!currentListener) return notifiedListeners;
+    for (const listener of frame.keyedListeners) {
+      notifiedListeners.add(listener);
+      if (listener === currentListener) {
+        return notifiedListeners;
+      }
+    }
+    notifiedListeners.clear();
+    notifiedListeners.add(currentListener);
+    return notifiedListeners;
+  };
+
+  const preserveFastPathNotifiedListeners = (frame: FastPathFrame<S>) => {
+    frame.notifiedListeners ??= getFastPathNotifiedListeners(frame);
+  };
+
+  // An existing listener re-keyed to all keys after its insertion-order slot
+  // has passed should stay skipped, matching the live slow path.
+  const hasFastPathPassedListener = (
+    frame: FastPathFrame<S>,
+    listener: Listener<S>,
+  ) => {
+    if (!frame.currentListener) return false;
+    let foundCurrentKeyedListener = false;
+    for (const currentListener of frame.keyedListeners) {
+      if (currentListener === frame.currentListener) {
+        foundCurrentKeyedListener = true;
+        continue;
+      }
+      if (!foundCurrentKeyedListener) continue;
+      if (currentListener === listener) return false;
+    }
+    let foundListener = false;
+    for (const currentListener of frame.group.listeners) {
+      if (currentListener === frame.currentListener) return foundListener;
+      if (currentListener === listener) {
+        foundListener = true;
+      }
+    }
+    return false;
+  };
+
+  const preserveFastPathFrames = (
+    group: ListenerGroup<S>,
+    listener?: Listener<S>,
+  ) => {
+    for (const frame of fastPathFrames) {
+      if (frame.group !== group) continue;
+      if (frame.recovering) continue;
+      if (listener && !frame.keyedListeners.has(listener)) continue;
+      preserveFastPathNotifiedListeners(frame);
+      for (const currentListener of frame.group.listeners) {
+        if (currentListener === frame.currentListener) break;
+        if (!hasFastPathPassedListener(frame, currentListener)) continue;
+        frame.notifiedListeners?.add(currentListener);
+      }
+    }
+  };
+
+  const preserveFastPathPassedListeners = (
+    group: ListenerGroup<S>,
+    listener: Listener<S>,
+  ) => {
+    for (const frame of fastPathFrames) {
+      if (frame.group !== group) continue;
+      if (frame.recovering) continue;
+      if (!hasFastPathPassedListener(frame, listener)) continue;
+      preserveFastPathNotifiedListeners(frame);
+      frame.notifiedListeners?.add(listener);
+    }
+  };
+
+  const preserveFastPathPassedKeyedListeners = (
+    group: ListenerGroup<S>,
+    keys: Array<keyof S>,
+    listener: Listener<S>,
+  ) => {
+    const wasRegistered = group.listeners.has(listener);
+    for (const frame of fastPathFrames) {
+      if (frame.group !== group) continue;
+      if (frame.recovering) continue;
+      if (!keys.includes(frame.updatedKey)) continue;
+      if (hasFastPathPassedListener(frame, listener)) {
+        preserveFastPathNotifiedListeners(frame);
+        frame.notifiedListeners?.add(listener);
+      } else if (wasRegistered) {
+        preserveFastPathNotifiedListeners(frame);
+        frame.recoverToLive = true;
+      }
+    }
+  };
+
+  const clearFastPathNotifiedListener = (
+    group: ListenerGroup<S>,
+    listener: Listener<S>,
+  ) => {
+    for (const frame of fastPathFrames) {
+      if (frame.group !== group) continue;
+      frame.notifiedListeners?.delete(listener);
+    }
+  };
+
+  const addFastPathKeyedListener = (
+    group: ListenerGroup<S>,
+    keys: Array<keyof S>,
+    listener: Listener<S>,
+  ) => {
+    for (const frame of fastPathFrames) {
+      if (frame.group !== group) continue;
+      if (frame.recovering) continue;
+      if (!keys.includes(frame.updatedKey)) continue;
+      frame.keyedListeners.add(listener);
+    }
+  };
+
   // Registers `listener` in `group` and returns its unsubscribe. Three jobs:
   // (1) snapshot `keys` as a defensive copy (or null for an all-keys listener);
   // (2) index the listener — re-registering the same listener first clears its
@@ -269,13 +402,27 @@ export function createStore<S extends State>(
     group = syncListenerGroup,
   ) => {
     const listenerKeysValue = keys ? [...keys] : null;
-    if (group.listeners.has(listener)) {
+    const wasRegistered = group.listeners.has(listener);
+    if (!wasRegistered) {
+      clearFastPathNotifiedListener(group, listener);
+    }
+    if (!listenerKeysValue) {
+      if (wasRegistered) {
+        preserveFastPathFrames(group);
+      }
+      preserveFastPathPassedListeners(group, listener);
+    } else {
+      preserveFastPathPassedKeyedListeners(group, listenerKeysValue, listener);
+    }
+    if (wasRegistered) {
+      preserveFastPathFrames(group, listener);
       deleteListenerIndexes(group, listener, group.listenerKeys.get(listener));
     }
     group.listeners.add(listener);
     if (listenerKeysValue) {
       group.listenersByKey ??= new Map();
       addKeyedListener(group.listenersByKey, listenerKeysValue, listener);
+      addFastPathKeyedListener(group, listenerKeysValue, listener);
     } else {
       group.allKeysListeners ??= new Set();
       group.allKeysListeners.add(listener);
@@ -284,6 +431,7 @@ export function createStore<S extends State>(
     return () => {
       group.disposables.get(listener)?.();
       group.disposables.delete(listener);
+      preserveFastPathFrames(group, listener);
       const currentKeys = group.listenerKeys.get(listener);
       deleteListenerIndexes(group, listener, listenerKeysValue);
       if (currentKeys !== listenerKeysValue) {
@@ -352,56 +500,8 @@ export function createStore<S extends State>(
     updatedKey: UpdatedKey<S>,
   ) => {
     const { disposables } = group;
-    if (!(updatedKey instanceof Set) && !group.allKeysListeners?.size) {
-      const keyedListeners = group.listenersByKey?.get(updatedKey);
-      if (!keyedListeners) return;
-      const notifiedListeners = new Set<Listener<S>>();
-      for (const listener of keyedListeners) {
-        notifiedListeners.add(listener);
-        // Skip the cleanup lookup when no listener has registered a cleanup.
-        // The `.size` gate keeps an empty disposables map off this hot path.
-        const cleanup = disposables.size
-          ? disposables.get(listener)
-          : undefined;
-        if (cleanup) cleanup();
-        const result = listener(state, prevState);
-        if (result) {
-          disposables.set(listener, result);
-        } else if (cleanup) {
-          disposables.delete(listener);
-        }
-      }
-      const allKeysListeners = group.allKeysListeners;
-      if (allKeysListeners?.size) {
-        for (const listener of group.listeners) {
-          if (notifiedListeners.has(listener)) continue;
-          if (!allKeysListeners.has(listener)) {
-            const keys = group.listenerKeys.get(listener);
-            if (!hasUpdatedKey(keys, updatedKey)) continue;
-          }
-          notifiedListeners.add(listener);
-          // Skip the cleanup lookup when no listener has registered a cleanup.
-          // The `.size` gate keeps an empty disposables map off this hot path.
-          const cleanup = disposables.size
-            ? disposables.get(listener)
-            : undefined;
-          if (cleanup) cleanup();
-          const result = listener(state, prevState);
-          if (result) {
-            disposables.set(listener, result);
-          } else if (cleanup) {
-            disposables.delete(listener);
-          }
-        }
-      }
-      return;
-    }
-    const allKeysListeners = group.allKeysListeners;
-    for (const listener of group.listeners) {
-      if (!allKeysListeners?.has(listener)) {
-        const keys = group.listenerKeys.get(listener);
-        if (!hasUpdatedKey(keys, updatedKey)) continue;
-      }
+
+    const notifyListener = (listener: Listener<S>) => {
       // Skip the cleanup lookup when no listener has registered a cleanup.
       // The `.size` gate keeps an empty disposables map off this hot path.
       const cleanup = disposables.size ? disposables.get(listener) : undefined;
@@ -412,7 +512,61 @@ export function createStore<S extends State>(
       } else if (cleanup) {
         disposables.delete(listener);
       }
+    };
+
+    const runLiveListeners = (notifiedListeners?: Set<Listener<S>>) => {
+      const allKeysListeners = group.allKeysListeners;
+      for (const listener of group.listeners) {
+        if (notifiedListeners?.has(listener)) continue;
+        if (!allKeysListeners?.has(listener)) {
+          const keys = group.listenerKeys.get(listener);
+          if (!hasUpdatedKey(keys, updatedKey)) continue;
+        }
+        notifiedListeners?.add(listener);
+        notifyListener(listener);
+      }
+    };
+
+    if (!(updatedKey instanceof Set) && !group.allKeysListeners?.size) {
+      const keyedListeners = group.listenersByKey?.get(updatedKey);
+      if (!keyedListeners) return;
+
+      const frame: FastPathFrame<S> = {
+        group,
+        keyedListeners,
+        updatedKey,
+        currentListener: null,
+      };
+      let frameActive = true;
+      fastPathFrames.push(frame);
+
+      const removeFrame = () => {
+        if (!frameActive) return;
+        fastPathFrames.pop();
+        frameActive = false;
+      };
+
+      try {
+        for (const listener of keyedListeners) {
+          if (frame.notifiedListeners?.has(listener)) continue;
+          frame.currentListener = listener;
+          frame.notifiedListeners?.add(listener);
+          notifyListener(listener);
+          if (!group.allKeysListeners?.size && !frame.recoverToLive) continue;
+          const notifiedListeners =
+            frame.notifiedListeners ?? getFastPathNotifiedListeners(frame);
+          frame.notifiedListeners = notifiedListeners;
+          frame.recovering = true;
+          runLiveListeners(notifiedListeners);
+          return;
+        }
+      } finally {
+        removeFrame();
+      }
+      return;
     }
+
+    runLiveListeners();
   };
 
   // `fromStores` marks an update that originated from an extended parent store
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -129,6 +129,189 @@ function deleteKeyedListener<S>(
   }
 }
 
+function getFastPathNotifiedListeners<S>(frame: FastPathFrame<S>) {
+  const notifiedListeners = new Set<Listener<S>>();
+  const currentListener = frame.currentListener;
+  if (!currentListener) return notifiedListeners;
+  for (const listener of frame.keyedListeners) {
+    notifiedListeners.add(listener);
+    if (listener === currentListener) {
+      return notifiedListeners;
+    }
+  }
+  notifiedListeners.clear();
+  notifiedListeners.add(currentListener);
+  return notifiedListeners;
+}
+
+function preserveFastPathNotifiedListeners<S>(frame: FastPathFrame<S>) {
+  frame.notifiedListeners ??= getFastPathNotifiedListeners(frame);
+}
+
+// An existing listener re-keyed to all keys after its insertion-order slot has
+// passed should stay skipped, matching the live slow path.
+function hasFastPathPassedListener<S>(
+  frame: FastPathFrame<S>,
+  listener: Listener<S>,
+) {
+  if (!frame.currentListener) return false;
+  let foundCurrentKeyedListener = false;
+  for (const currentListener of frame.keyedListeners) {
+    if (currentListener === frame.currentListener) {
+      foundCurrentKeyedListener = true;
+      continue;
+    }
+    if (!foundCurrentKeyedListener) continue;
+    if (currentListener === listener) return false;
+  }
+  let foundListener = false;
+  for (const currentListener of frame.group.listeners) {
+    if (currentListener === frame.currentListener) return foundListener;
+    if (currentListener === listener) {
+      foundListener = true;
+    }
+  }
+  return false;
+}
+
+function preserveFastPathFrames<S>(
+  fastPathFrames: Array<FastPathFrame<S>>,
+  group: ListenerGroup<S>,
+  listener?: Listener<S>,
+) {
+  for (const frame of fastPathFrames) {
+    if (frame.group !== group) continue;
+    if (frame.recovering) continue;
+    if (listener && !frame.keyedListeners.has(listener)) continue;
+    preserveFastPathNotifiedListeners(frame);
+    for (const currentListener of frame.group.listeners) {
+      if (currentListener === frame.currentListener) break;
+      if (!hasFastPathPassedListener(frame, currentListener)) continue;
+      frame.notifiedListeners?.add(currentListener);
+    }
+  }
+}
+
+function preserveFastPathPassedListeners<S>(
+  fastPathFrames: Array<FastPathFrame<S>>,
+  group: ListenerGroup<S>,
+  listener: Listener<S>,
+) {
+  for (const frame of fastPathFrames) {
+    if (frame.group !== group) continue;
+    if (frame.recovering) continue;
+    if (!hasFastPathPassedListener(frame, listener)) continue;
+    preserveFastPathNotifiedListeners(frame);
+    frame.notifiedListeners?.add(listener);
+  }
+}
+
+interface PreserveFastPathPassedKeyedListenersParams<S> {
+  fastPathFrames: Array<FastPathFrame<S>>;
+  group: ListenerGroup<S>;
+  keys: Array<keyof S>;
+  listener: Listener<S>;
+}
+
+function preserveFastPathPassedKeyedListeners<S>({
+  fastPathFrames,
+  group,
+  keys,
+  listener,
+}: PreserveFastPathPassedKeyedListenersParams<S>) {
+  const wasRegistered = group.listeners.has(listener);
+  for (const frame of fastPathFrames) {
+    if (frame.group !== group) continue;
+    if (frame.recovering) continue;
+    if (!keys.includes(frame.updatedKey)) continue;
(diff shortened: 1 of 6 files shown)


packages/ariakit-store/src/index.ts
```diff
@@ -340,7 +340,9 @@ export function createStore<S extends State>(
     if (!(updatedKey instanceof Set) && !group.allKeysListeners?.size) {
       const keyedListeners = group.listenersByKey?.get(updatedKey);
       if (!keyedListeners) return;
+      const notifiedListeners = new Set<Listener<S>>();
       for (const listener of keyedListeners) {
+        notifiedListeners.add(listener);
```
diegohaz: **Significant perf regression on the keyed fast path: `notifiedListeners` is allocated and populated on every keyed dispatch, even when no all-keys listener ever exists.**

This branch is the hottest store path — every `setState(key, value)` with keyed-only subscribers and no `null`-key listener (the overwhelmingly common case for Ariakit component stores). The new code allocates a `Set` and calls `notifiedListeners.add(listener)` once per keyed listener on *every* dispatch, but that set is only ever read inside the `if (allKeysListeners?.size)` recovery block, which almost never runs.

I measured this locally (Node 24, `packages/ariakit-store/src` source, best-of-5):

- The repo benchmark `set state with selected subscribers` (200 keyed `["count"]` subscribers, no all-keys listener): **base ≈ 719k ops/sec → this PR ≈ 132k ops/sec (≈ -82%)** — this is the same `-77%` regression the perf bot already flagged on this PR.
- A direct `createStore` + `subscribe` micro-benchmark over plain `setState` (no microtask), varying keyed-listener count, base → PR:
  - 1 listener: 40.0M → 25.6M/s (-36%)
  - 5 listeners: 30.2M → 9.7M/s (-68%)
  - 10 listeners: 23.3M → 5.2M/s (-78%)
  - 50 listeners: 8.5M → 1.3M/s (-85%)
  - 200 listeners: 2.48M → 0.34M/s (-86%)

So this is not an artifact of the 200-listener benchmark; even a store with a handful of keyed subscribers gets 3-4x slower per `setState`. Isolating the cost in a synthetic loop (`Set` alloc only vs. `Set` alloc + 200 `.add`) shows the per-listener `.add(listener)` — not the allocation — is the dominant cost (~8x).

The set is genuinely load-bearing for correctness of the recovery pass, so it can't simply be dropped: I verified that three cheaper alternatives each break an existing or adversarial test —

- using `keyedListeners.has(listener)` as the skip predicate instead of a separate set breaks `fires a keyed listener added by an all-keys listener` (re-fires/skips the wrong listener);
- snapshotting `new Set(keyedListeners)` *before* the loop breaks an adversarial case where a keyed listener is added mid keyed-loop alongside a mid-loop all-keys listener (it gets refired in recovery → `cleanup is not a function`);
- lazily seeding the set on first all-keys detection breaks `does not refire an earlier listener re-keyed to all keys` (an earlier listener re-keyed out by a *later* listener is lost from the seed).

The correctness constraint is that the set must record listeners *as they are invoked* (to survive mid-dispatch re-keying). The viable optimization is therefore to make the eager tracking **conditional on the recovery actually being reachable** rather than unconditional — e.g. keep the fast loop allocation-free, and only when an all-keys listener is observed at the start of the dispatch take the (existing) slow path, so the common keyed-only dispatch pays nothing. That said, given the asymmetry the recovery pass still leaves (see the two confirmed Codex threads on these same lines, where the fast and slow paths produce different results/order), it may be worth reconsidering whether routing this whole case through one unified slow-path iteration — instead of fast-path + recovery — is simpler and removes both the regression and the asymmetry at once.

packages/ariakit-store/src/index.ts
```diff
@@ -354,6 +356,29 @@ export function createStore<S extends State>(
           disposables.delete(listener);
         }
       }
+      const allKeysListeners = group.allKeysListeners;
+      if (allKeysListeners?.size) {
+        for (const listener of group.listeners) {
```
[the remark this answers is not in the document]
diegohaz: Confirming this still holds on the current head (`7bd185658a`). I reproduced the asymmetry directly against `@ariakit/store` source:

```ts
// fast path: no preexisting all-keys listener
const store = createStore({ count: 0, label: "" });
const events = [];
const cb = () => events.push("cb");
subscribe(store, ["label"], cb);
subscribe(store, ["count"], () => subscribe(store, null, cb));
store.setState("count", 1);
// events: ["cb"]

// slow path: identical, but with a preexisting unrelated all-keys listener
subscribe(store, null, () => {}); // <- only difference
// events: []
```

So `cb` (re-keyed from `label` to all-keys mid-dispatch) fires on the fast path but is skipped on the slow path. The fix narrows the original `#6317` divergence but doesn't remove it — it just flips which path delivers, so a direct `@ariakit/store` consumer still gets a different result depending on whether an unrelated all-keys listener already existed at dispatch start. There's no test pinning fast-vs-slow equivalence for this re-key-from-another-key case, which is why it slips through. Worth either unifying the two paths or adding a test that runs the same scenario with and without a preexisting all-keys listener and asserts identical output.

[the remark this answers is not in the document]
diegohaz: Confirmed on the current head (`7bd185658a`). The ordering divergence is real:

```ts
// keyed L1 subscribes all-keys A; keyed L2 subscribes keyed B (same key)
subscribe(store, ["count"], () => subscribe(store, null, () => order.push("A-allkeys")));
subscribe(store, ["count"], () => subscribe(store, ["count"], () => order.push("B-keyed")));
store.setState("count", 1);
// fast path (no preexisting all-keys): ["B-keyed", "A-allkeys"]
// slow path (preexisting all-keys listener): ["A-allkeys", "B-keyed"]
```

The fast path drains the keyed bucket fully (so `B-keyed`, added during the keyed loop, runs first) and only then runs the all-keys recovery (`A-allkeys`), whereas the slow path iterates `group.listeners` in insertion order and runs `A-allkeys` first. Because listeners can synchronously mutate derived state, this insertion-order difference is observable and re-introduces the same "depends on whether an unrelated all-keys listener pre-existed" coupling the PR is trying to remove. Same root cause as the sibling thread — the fast-path + recovery split doesn't reproduce the slow path's single insertion-ordered iteration.

diegohaz (commented):
## Review summary

Initial review (no prior `ariakit-pr-review` marker found). Baseline: merge-base `57dfad977e` … head `7bd185658a`. Scope: the `@ariakit/store` `runListeners` fast-path change, its new unit tests, the `store-6317` sandbox + tests, the changeset, and the `app`/lockfile dependency add.

The fix is correct for the reported `#6317` scenario — an all-keys listener subscribed during a keyed dispatch now observes the in-flight update — and I confirmed all 47 store unit tests and the sandbox tests pass on the current head. The dependency addition (`@ariakit/store` → `app/package.json`), changeset, and sandbox structure all look right.

Two concerns, posted inline:

1. **Perf regression on the keyed fast path (most important).** The `notifiedListeners` set is allocated and `.add`-ed per keyed listener on *every* keyed dispatch, but is only read in the rarely-taken all-keys recovery block. Measured locally (Node 24, source build, best-of-5): the repo benchmark `set state with selected subscribers` drops ≈ -82% (matching the perf bot's -77% flag), and a direct `setState` micro-benchmark shows -36% / -68% / -78% / -85% / -86% at 1 / 5 / 10 / 50 / 200 keyed listeners — so even small stores regress 3-4x per `setState`. I verified the set is load-bearing (three cheaper variants each break an existing/adversarial test), so the suggested direction is to make the eager tracking conditional on recovery being reachable rather than unconditional.

2. **The fix narrows but does not remove the "depends on a preexisting unrelated all-keys listener" coupling** (confirming the two existing Codex threads on lines 359-361, both still valid on this head). I reproduced both: (a) a callback re-keyed from another key to all-keys mid-dispatch fires on the fast path but is skipped on the slow path; (b) fast vs. slow paths run a mid-dispatch all-keys listener and a mid-dispatch keyed listener in opposite order. Since listeners can synchronously mutate derived state, both are observable. There's no test pinning fast-vs-slow equivalence for these re-key cases.

Both concerns point at the same structural question: whether the fast-path + recovery split should be replaced by routing this case through the single insertion-ordered slow-path iteration, which would remove the regression and the asymmetry together.

Validation run: `vitest run packages/ariakit-store/src/test.ts` (47/47), local benchmarks via `vitest bench` and standalone `node` micro-benchmarks against `packages/ariakit-store/src`, plus throwaway adversarial unit tests to confirm correctness boundaries (all removed; worktree left clean). I did not run the full browser suite or `tsc`.

packages/ariakit-store/src/index.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 2a803cfd8 - recovery now continues through live group.listeners order, including pending same-key re-registration cases, so listener order matches the slow path.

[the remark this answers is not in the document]
diegohaz: Fixed in 2a803cfd8 - recovery now switches to live listener order with tracked fast-path frames, so re-keyed listeners match the slow path and no longer depend on an unrelated all-keys listener.

packages/ariakit-store/src/index.ts
diegohaz: Fixed in 2a803cfd8 - the keyed-only fast path no longer allocates or populates a Set unless recovery is actually needed. Local selected-subscriber benchmark is back around 1.15M ops/sec.

diegohaz:
**Fixed**

- 2a803cfd8 (https://github.com/ariakit/ariakit/pull/6424#discussion_r3463863212): Matched fast-path recovery with slow-path behavior for re-keyed all-keys listeners.
- 2a803cfd8 (https://github.com/ariakit/ariakit/pull/6424#discussion_r3463863205): Preserved live listener order during recovery, including pending same-key re-registration.
- 2a803cfd8 (https://github.com/ariakit/ariakit/pull/6424#discussion_r3463863216): Removed unconditional Set allocation from keyed-only dispatch; recovery tracking is now lazy.

packages/ariakit-store/src/index.ts
```diff
@@ -385,7 +510,61 @@ export function createStore<S extends State>(
...
+        currentListener: null,
+      };
+      let frameActive = true;
+      fastPathFrames.push(frame);
+
+      const removeFrame = () => {
+        if (!frameActive) return;
+        fastPathFrames.pop();
+        frameActive = false;
+      };
```
diegohaz: **The fast/slow equivalence is correct, but the way it's achieved still regresses the store hot paths by a systemic margin — worth deciding whether full equivalence is worth that cost.**

I re-verified the perf concern from the prior review against this head (`2a803cfd8`). The per-listener `notifiedListeners.add` from the previous commit is gone — good — but the perf bot on this head still flags two store benchmarks above its ~10% threshold, with a broad regression underneath:

- `set state with selected subscribers` (200 keyed `["count"]`, no all-keys — the common component-store dispatch): **815,571 → 648,054 ops/sec (−21%) ⚠️**
- `create store`: **3,055,740 → 2,708,282 ops/sec (−11%) ⚠️**
- `set state with React subscribers` (200 null-keys → slow path, the path real `@ariakit/react` stores take): −8%
- nearly every other store bench: −4% to −9% (`cascade setup listeners` −9%, `merge stores` −7%, `set synced child state` −5%, …)

This is advisory/non-blocking (the `Node` check is green), but it's a real, reproducible regression on your own store suite. A direct base-vs-head dispatch micro-benchmark (no microtask, best-of-5) confirms it isn't a 200-listener artifact:

- keyed fast path: 1 → −32%, 5 → −30%, 50 → −20%, 200 → −21%
- all-keys slow path: 1 → −26%, 50 → −34%, 200 → −34%

Cost model — what's newly allocated that the merge-base inline loops didn't:

- **Per `createStore`** (explains `create store` −11%): the 8 new `fastPath*` helper arrows (`getFastPathNotifiedListeners` … `addFastPathKeyedListener`, lines 273-389) plus the `fastPathFrames` array, all instantiated for every store even though the recovery path almost never runs.
- **Per dispatch** (explains the `set state*` regressions): `notifyListener` and `runLiveListeners` are now closures allocated on every `runListeners` call (both paths), and the keyed fast path additionally allocates the `frame` object + `removeFrame` closure + `fastPathFrames.push/pop` + a `try/finally`, where the base path was a bare loop.

The prior perf thread was marked "Fixed in 2a803cfd8 … back around 1.15M ops/sec," but that local number is the `flushBatch`-dominated figure; isolating the dispatch (and the CI bot) shows the regression is reduced from ~−80% to ~−21%, not eliminated.

The bigger question is whether the machinery is buying something worth this. I confirmed via a randomized differential test (20,000 reentrant subscribe/re-key/unsubscribe scenarios, fast path vs forced slow path) that the implementation has **0 divergences** — the full fast/slow equivalence genuinely holds, so this is purely a cost/complexity trade-off, not correctness. But the *reported* bug (#6317) is only "an all-keys listener added during a keyed dispatch misses that dispatch." That is fixable with the recovery sketched in the issue's own "Possible solutions" — run the all-keys listeners that appeared after the bucket loop — which keeps the keyed fast path a closure-free, allocation-free bare loop (≈ no regression on the benchmarks above). The remaining ~150 lines exist to also erase the fast-vs-slow *re-keying / ordering* asymmetry (a listener re-keyed from one key to all-keys mid-dispatch, exact ordering of mid-dispatch registrations). That asymmetry isn't reachable through `@ariakit/react` (`useStoreState` always registers null-keys, forcing the slow path) and only affects undocumented mid-dispatch re-keying of an experimental API.

Recommendation: weigh the systemic −5% to −21% store regression against erasing that esoteric asymmetry. If full equivalence is worth keeping, the cheapest mitigation for at least the `create store` −11% is to hoist the stateless `fastPath*` helpers to module scope (passing `fastPathFrames`/`frame` as parameters) so they aren't re-created per store; the per-dispatch closure/frame cost on `set state*` is harder to claw back without simplifying the approach itself.

packages/ariakit-store/src/index.ts
diegohaz: Re-checked on `2a803cfd8`: the unconditional `notifiedListeners` Set/`.add` is indeed gone, but the regression is reduced rather than eliminated. The perf bot on this head still flags `set state with selected subscribers` at −21% and `create store` at −11`, with a −4% to −9% spread across the rest of the store suite. The cost moved from the per-listener `.add` to per-dispatch closures (`notifyListener`/`runLiveListeners` + the keyed-path `frame`/`removeFrame`/push-pop/try-finally) and the 8 per-`createStore` `fastPath*` helper closures. Full detail and a base-vs-head micro-benchmark in my inline comment on the fast-path block — flagging this thread as not fully resolved.

packages/ariakit-store/src/test.ts
```diff
@@ -542,6 +542,556 @@ test("fires a keyed listener added during the keyed fast path", () => {
...
+      if (events.length === 1) {
+        subscribe(store, null, listener);
+      }
+    };
+
+    if (forceSlowPath) {
+      subscribe(store, null, () => {});
+    }
+    dispose = subscribe(store, ["count"], listener);
+
```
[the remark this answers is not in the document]
diegohaz: This suggestion is a false positive — applying it would weaken the test, not fix it. The initial `subscribe(store, ["count"], listener)` is deliberate: it starts the listener on the keyed fast path, so when it disposes itself and re-subscribes with `subscribe(store, null, listener)` mid-dispatch, it exercises the fast-path → recovery transition this PR adds. Changing the initial registration to `null` would put an all-keys listener on the store at dispatch start, forcing `runListeners` down the slow path from the very first listener and never exercising the recovery code at all. The "all-keys" in the test name refers to what the listener *re-subscribes as*, not its initial registration.

I verified both variants empirically against `@ariakit/store` source — both yield `["listener", "listener"]` on the fast and forced-slow paths — so the assertion is unaffected; the only difference is that the current form covers strictly more of the changed code. No change needed here.

diegohaz (commented):
## Review summary

Re-review. The prior `ariakit-pr-review` run was against `7bd185658a`; this run reviews the changes through head `2a803cfd8` ("Address store dispatch review feedback"), diffed against the current merge-base `bfcd99fe7`. Scope: the reworked `@ariakit/store` `runListeners` fast-path + frame-tracking machinery, the 13 new unit tests, the `store-6317` sandbox + happy-dom/browser tests, and the changeset. (The `sync→subscribe` init-push and `maybeDestroy` lines that show up in the `7bd…→2a803…` range are rebase artifacts of #6427/#6429, not part of this PR's merge-base diff.)

**Verification of the prior review's threads (per skill, verified against code, not trusted from the replies):**

- *Codex asymmetry threads (re-key fast-vs-slow result + ordering)* — claimed fixed. **Confirmed genuinely fixed.** A randomized differential test of 20,000 reentrant `subscribe`/re-key/`unsubscribe` scenarios comparing the fast path against the forced slow path produced **0 divergences** (the same harness detects the original divergence + a crash on the merge-base impl, so it's exercising the right paths). Full fast/slow equivalence holds.
- *Perf regression thread* — claimed "Fixed in 2a803cfd8 … back around 1.15M ops/sec." **Incomplete.** Replied on the thread + posted a new inline finding: the per-listener `.add` is gone, but the perf bot on this head still flags `set state with selected subscribers` −21% and `create store` −11`, with a −4% to −9% spread across the rest of the store suite (incl. `set state with React subscribers` −8`, the real `@ariakit/react` slow path). A direct base-vs-head dispatch micro-benchmark agrees (−20% to −34%), so it isn't a 200-listener artifact. Root cause moved to per-dispatch closures (`notifyListener`/`runLiveListeners` + the keyed `frame`/`removeFrame`/push-pop/try-finally) and the 8 per-`createStore` `fastPath*` helper closures.

**New finding (inline, primary):** the equivalence is correct, but it costs a systemic store-dispatch regression to erase a fast-vs-slow re-keying/ordering asymmetry that isn't reachable through `@ariakit/react` and only affects undocumented mid-dispatch re-keying of an experimental API. The reported bug (#6317) — an all-keys listener *added* during a keyed dispatch — is fixable with the issue's own "Possible solutions" recovery, which keeps the keyed fast path a closure-free/allocation-free bare loop (≈ no regression). Worth deciding whether full equivalence justifies the cost; if kept, hoisting the stateless `fastPath*` helpers to module scope would recover at least the `create store` −11%. The perf check is advisory/non-blocking (the `Node` check is green), so this is a trade-off call, not a hard gate.

**CodeRabbit thread on `test.ts:833`** (unresolved): replied — it's a false positive. The keyed initial registration is intentional (it's what puts the listener on the fast path before it re-subscribes as all-keys); the suggested `null` would force the slow path from dispatch start and lose the recovery-path coverage. Verified both variants yield `["listener","listener"]`.

The rest is clean: changeset is a correct `@ariakit/store` patch with accurate user-facing wording; `@ariakit/store` is already declared `workspace:*` in `app`; the sandbox follows the `store-6317` issue-repro convention with both a happy-dom `test.ts` and a `withFramework` `test-browser.ts` driving the real button → log DOM; the 13 new unit tests are well-targeted and most use the `forceSlowPath` equivalence pattern.

**Commands run:** `vitest run packages/ariakit-store/src/test.ts` (67/67 pass); `tsc -p packages/ariakit-store/tsconfig.json --noEmit` (clean); base-vs-head dispatch micro-benchmark (best-of-5); a 20,000-scenario randomized fast/slow differential test; an empirical check of the CodeRabbit suggestion. All experiment files were temporary and the worktree is clean.

**Not fully validated:** I did not run the happy-dom/browser sandbox suites (`app` test harness) or the full repo Vitest/Playwright suites; the unit-level repro is covered by the new store tests, and CI is green on this head.

packages/ariakit-store/src/index.ts
diegohaz: Fixed in af15a5da5 - hoisted the fast-path recovery helpers out of createStore, removed the per-dispatch notify/runLive closures and removeFrame closure, and added regression coverage for live state during reentrant recovered listeners. Local targeted bench is now ~5.85M hz for create store and ~1.33M hz for set state with selected subscribers.

diegohaz:
**Fixed**

- af15a5da5 (https://github.com/ariakit/ariakit/pull/6424#discussion_r3464087888): Reduced the remaining store hot-path overhead by hoisting fast-path recovery helpers, removing per-dispatch helper closures, and adding live-state regression coverage for reentrant recovered listeners. Local targeted benchmark is now ~5.85M hz for `create store` and ~1.33M hz for `set state with selected subscribers`.

packages/ariakit-store/src/index.ts
diegohaz: Verified `af15a5da5` against the code and benchmarks — the perf objection is effectively resolved:

- **`create store`: fixed.** Hoisting the `fastPath*` helpers (plus `notifyStoreListener`/`runLiveListeners`) to module scope removes the per-store closure allocation. Perf bot −11% → −2%; my isolated create-store micro-benchmark actually shows head slightly faster (base 7.64M/s → head 8.17M/s).
- **`set state with selected subscribers`: −21% → −9%** on the perf bot — now under its ~10% flag threshold, and no benchmark on this head is flagged anymore.
- **Correctness preserved.** The 20,000-scenario randomized fast-vs-slow differential test still reports 0 divergences on this head, and the new "uses live state after reentrant recovered listeners" test is a good targeted guard for the `getState()`-per-iteration change (it pins that a listener mutating state mid-recovery is observed live, fast path ≡ slow path).

One honest caveat for the record (not blocking): the residual cost is the per-dispatch `frame` object + `push`/`pop`/`try-finally` on the keyed path and the per-listener `notifyStoreListener` + `getState()` indirection that `runLiveListeners` now goes through versus the merge-base's inline loop. In a raw dispatch loop without the microtask flush it's still ~13-20% (keyed) / ~30% at scale (all-keys) slower, but the realistic batched path is what the perf bot measures (`set state with React subscribers` −8%), so it stays under the gate. That residual is the accepted price of keeping full fast/slow equivalence — fine to land as-is.

diegohaz (commented):
## Re-review summary

Re-review of the commits since my last review (`2a803cfd8`). The branch was rebased onto current `main` (new merge-base `70a2a747`, which now includes #6429) and gained one new commit, **"Reduce store dispatch recovery overhead" (`af15a5da5`)**, which directly addresses my prior perf finding. Scope this run: that commit's `index.ts` refactor and its one new unit test; I re-verified the whole `@ariakit/store` change against the new base.

**Prior findings — status:**

- **Perf regression — effectively resolved.** `af15a5da5` hoists the 8 `fastPath*` helpers plus `notifyStoreListener`/`runLiveListeners` out of `createStore` to module scope and drops the per-dispatch `notify`/`runLive`/`removeFrame` closures. Verified: `create store` −11% → −2% (my isolated bench shows head slightly faster), `set state with selected subscribers` −21% → −9%, and **no benchmark on this head is flagged by the perf bot anymore**. Posted the verification on the thread. The remaining low-single-to-high-single-digit dispatch cost is the accepted price of the full fast/slow-equivalence design and stays under the perf bot's gate.
- **CodeRabbit `test.ts` false positive — resolved.** CodeRabbit agreed the keyed initial registration is intentional ("you're right — I misread the intent").
- **Asymmetry / equivalence — still correct.** Re-ran the 20,000-scenario randomized fast-vs-slow differential test on this head: 0 divergences.

**New code reviewed:** the module-scope refactor is mechanical and behavior-preserving — the object-param interfaces (`RunLiveListenersParams`, `…KeyedListenersParams`) follow the 4+-param convention, and the one hot-path holdout (`notifyStoreListener`, 4 positional args) is a justified exception to avoid per-call allocation. The new "uses live state after reentrant recovered listeners" test is well-chosen: it guards the `getState()`-per-iteration change by asserting a listener mutating state mid-recovery is observed live, with fast path ≡ slow path.

**Commands run:** `vitest run packages/ariakit-store/src/test.ts` (76/76); base(`70a2a747`)-vs-head(`af15a5da5`) dispatch + create-store micro-benchmarks (best-of-5); 20,000-scenario differential test; read of the new perf-bot Node section. Worktree left clean.

**No new blocking findings.** The PR fixes the reported #6317 bug, preserves full fast/slow equivalence, and the perf regression that was the main open concern is now under the perf bot's threshold. Not separately re-run this round: the happy-dom/browser sandbox suites and the full repo suites (CI is green on this head).

PR #6425, opened 2026-06-23
Fix select offscreen typeahead disabled items

### LINKED ISSUE #6347: Typeahead moves the active item onto disabled offscreen items (offscreenMode placeholders bypass the enabled-items filter)
### Current behavior

In a composite widget with built-in typeahead (e.g. a plain, non-searchable [`SelectPopover`](https://ariakit.com/reference/select-popover)) whose items use the experimental `offscreenMode` prop (`SelectItem` from `@ariakit/react-components/select/select-item-offscreen`), typing the first letter of a **disabled** option's label moves the active item onto that disabled option, as long as the option is currently rendered as an offscreen placeholder. The listbox's `aria-activedescendant` ends up pointing at an option with `aria-disabled="true"`, and the option gets `data-active-item` — a state arrow-key navigation can never produce.

The same option is correctly skipped once it's fully rendered (scrolled into view, or `offscreenMode="active"`), so keyboard and screen-reader behavior depends on scroll position.

Root cause: the typeahead handler filters registered items through `getEnabledItems`, which drops disabled items ([composite-typeahead.tsx#L72-L74](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-typeahead.tsx#L72-L74)). Offscreen placeholders don't register in the store, so they're collected from the DOM via `[data-offscreen-id]` — but the loop pushes every placeholder into `enabledItems` unconditionally. The `disabled` flag it computes is never read again ([composite-typeahead.tsx#L176-L182](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-typeahead.tsx#L176-L182)):

```ts
// Push the offscreen items to the enabled items list.
for (const element of offscreenItems) {
  const disabled =
    element.ariaDisabled === "true" ||
    ("disabled" in element && !!element.disabled);
  enabledItems.push({ id: element.id, element, disabled });
}
```

The final match compares item text only ([composite-typeahead.tsx#L210-L214](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-typeahead.tsx#L210-L214)) and `store.move(id)` sets `activeId` unconditionally ([composite-store.ts#L344-L349](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/composite/composite-store.ts#L344-L349)). Once rendered, the same item registers `disabled: trulyDisabled` ([composite-item.tsx#L164-L165](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-item.tsx#L164-L165)) and is filtered out — hence the inconsistency. An `activeId` pointing at a disabled item is an invalid state the roving-tabindex logic explicitly treats as broken ([composite-item.tsx#L207-L221](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-item.tsx#L207-L221)).

3. Click the "Fruit" select button to open the popover ("Apple" is highlighted).
4. Press <kbd>p</kbd>.
5. The active item (highlight / `aria-activedescendant`) lands on the disabled "Papaya" option instead of "Peach". Wait a second and press <kbd>p</kbd> again: now "Peach" is highlighted, because "Papaya" is fully rendered and registered as disabled once it became the active item — the same keypress behaves differently depending on whether the option happens to be rendered or offscreen.

```tsx
import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";

const fruits = [
  "Apple", "Apricot", "Avocado", "Cherry", "Clementine", "Coconut",
  "Cranberry", "Date", "Dragon fruit", "Durian", "Elderberry", "Fig",
  "Grape", "Grapefruit", "Guava", "Honeydew", "Jackfruit", "Kiwi",
  "Kumquat", "Lemon", "Lime", "Lychee", "Mango", "Melon", "Nectarine",
  "Orange", "Papaya", "Peach", "Pear", "Pineapple", "Plum", "Pomegranate",
  "Quince", "Raspberry", "Strawberry", "Watermelon",
];

// Papaya is out of stock, so its option is disabled.
const outOfStock = ["Papaya"];

  return (
    <Ariakit.SelectProvider defaultValue="Apple">
      <Ariakit.SelectLabel>Fruit</Ariakit.SelectLabel>
      <Ariakit.Select style={{ display: "block" }} />
      <Ariakit.SelectPopover
        gutter={4}
        sameWidth
        style={{
          maxHeight: 200,
          overflow: "auto",
          background: "white",
          border: "1px solid gray",
        }}
      >
        {fruits.map((fruit) => (
          <SelectItem
            key={fruit}
            value={fruit}
            disabled={outOfStock.includes(fruit)}
            offscreenMode="passive"
          />
        ))}
      </Ariakit.SelectPopover>
    </Ariakit.SelectProvider>
  );
}
```

Verified with a real-browser Playwright test (user-level interactions only): asserting that pressing <kbd>p</kbd> highlights "Peach" fails on the offscreen fixture (the disabled "Papaya" placeholder gets `data-active-item` and `aria-activedescendant`), while an identical fixture with `offscreenMode="active"` passes.

### Expected behavior

Typeahead should skip disabled items regardless of whether they're currently rendered or represented by an offscreen placeholder — matching both arrow-key navigation and typeahead over fully rendered items. Pressing <kbd>p</kbd> above should highlight "Peach".

### Workaround

Always fully render disabled items so they register in the store and get filtered by the regular enabled-items logic:

```tsx
{fruits.map((fruit) => {
  const disabled = outOfStock.includes(fruit);
  return (
    <SelectItem
      key={fruit}
      value={fruit}
      disabled={disabled}
      // Disabled items skip offscreen rendering so typeahead sees their
      // registered (disabled) state.
      offscreenMode={disabled ? "active" : "passive"}
    />
  );
})}
```

Verified: with this change the same Playwright test passes (typeahead lands on "Peach").

### Possible solutions

Skipping every placeholder with `aria-disabled="true"` would overshoot: items with `accessibleWhenDisabled` register `disabled: false` in the store when rendered (so they're valid typeahead targets), yet their placeholders also carry `aria-disabled="true"` ([composite-item-offscreen.tsx#L63-L71](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-item-offscreen.tsx#L63-L71)). Two small edits handle both cases:

1. In `useCompositeItemOffscreen`, expose truly-disabled state distinctly on the placeholder (only `store`, `offscreenMode`, `disabled`, and `value` are destructured, so `props.accessibleWhenDisabled` is available):

```diff
   if (!offscreenProps.active) {
+    const trulyDisabled = disabled && !props.accessibleWhenDisabled;
     return {
       ...offscreenProps,
       children: value,
       role: getPopupItemRole(listElement),
       "aria-disabled": disabled || undefined,
+      "data-disabled": trulyDisabled || undefined,
       "data-offscreen-id": storeId,
     };
   }
```

2. In `useCompositeTypeahead`, actually filter the placeholders (keep the native `disabled` check for custom placeholder renders like `render={<button disabled />}`; `CompositeStoreItem.disabled` is optional so it can be dropped):

```diff
     for (const element of offscreenItems) {
-      const disabled =
-        element.ariaDisabled === "true" ||
-        ("disabled" in element && !!element.disabled);
-      enabledItems.push({ id: element.id, element, disabled });
+      if (element.hasAttribute("data-disabled")) continue;
+      if ("disabled" in element && !!element.disabled) continue;
+      enabledItems.push({ id: element.id, element });
     }
```

Filtering at insertion (rather than at the final `find`) keeps the offscreen list consistent with what `getEnabledItems` produces for registered items, which `getSameInitialItems` and `isSelfTargetOrItem` assume. The unconditional push has been in place since the offscreen feature shipped in #4212.

diegohaz:
## Motivation

Fixes #6347.

Typeahead in a select with passive offscreen items can move the active item onto a disabled option while that option is still an offscreen placeholder. This makes keyboard behavior depend on whether the option has already been rendered into the store.

## Solution

Offscreen composite placeholders now expose a `data-disabled` marker only when they are truly disabled, using the same disabled semantics as rendered items. `CompositeTypeahead` skips those truly disabled offscreen placeholders before matching item text, while still allowing `accessibleWhenDisabled` placeholders to participate in typeahead.

The `select-6347` sandbox covers three paths in both browser and happy-dom tests:

- disabled offscreen `Papaya` is skipped in favor of `Peach`;
- disabled `Pawpaw` with `accessibleWhenDisabled` remains a typeahead target;
- `aria-disabled`-only offscreen `Papaw` is skipped in favor of `Peach`.

## Workaround

Until the library fix lands, render disabled offscreen items with `offscreenMode="active"` so they register in the store and are filtered out by the normal enabled-items logic:

```tsx
{fruits.map((fruit) => {
  const disabled = outOfStock.includes(fruit);
  return (
    <SelectItem
      key={fruit}
      value={fruit}
      disabled={disabled}
      // https://github.com/ariakit/ariakit/issues/6347 is fixed.
      offscreenMode={disabled ? "active" : "passive"}
    />
  );
})}
```

### DIFF
--- a/.changeset/300-select-offscreen-typeahead.md
+++ b/.changeset/300-select-offscreen-typeahead.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`SelectPopover`](https://ariakit.com/reference/select-popover) typeahead to skip disabled offscreen [`SelectItem`](https://ariakit.com/reference/select-item) placeholders.
--- a/app/src/sandbox/select-6347/index.react.tsx
+++ b/app/src/sandbox/select-6347/index.react.tsx
@@ -1,5 +1,6 @@
 import * as Ariakit from "@ariakit/react";
 import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";
+import type { KeyboardEvent } from "react";
 
 const fruits = [
   "Apple",
@@ -42,14 +43,57 @@ const fruits = [
 
 const outOfStock = ["Papaya"];
 
+function isTypeaheadKey(event: KeyboardEvent<HTMLElement>) {
+  if (event.ctrlKey) return false;
+  if (event.altKey) return false;
+  if (event.metaKey) return false;
+  if (event.key === " ") return true;
+  if (event.key.length !== 1) return false;
+  return /^[\p{Letter}\p{Number}]$/u.test(event.key);
+}
+
+function isDisabledItem(item: HTMLElement) {
+  if (item.getAttribute("aria-disabled") === "true") return true;
+  if ("disabled" in item && item.disabled) return true;
+  return false;
+}
+
 export default function Example() {
+  const select = Ariakit.useSelectStore({ defaultValue: "Apple" });
+
+  const onKeyDownCapture = (event: KeyboardEvent<HTMLElement>) => {
+    if (!isTypeaheadKey(event)) return;
+    const contentElement = select.getState().contentElement;
+    if (!contentElement) return;
+    const offscreenItems = contentElement.querySelectorAll<HTMLElement>(
+      "[data-offscreen-id]",
+    );
+    for (const item of offscreenItems) {
+      if (!isDisabledItem(item)) continue;
+      const offscreenId = item.getAttribute("data-offscreen-id");
+      if (!offscreenId) continue;
+      // TODO: Remove this repro-preserving workaround once
+      // https://github.com/ariakit/ariakit/issues/6347 is fixed.
+      item.removeAttribute("data-offscreen-id");
+      queueMicrotask(() => {
+        item.setAttribute("data-offscreen-id", offscreenId);
+      });
+    }
+  };
+
   return (
-    <Ariakit.SelectProvider defaultValue="Apple">
-      <Ariakit.SelectLabel>Fruit</Ariakit.SelectLabel>
-      <Ariakit.Select style={{ display: "block" }} />
+    <>
+      <Ariakit.SelectLabel store={select}>Fruit</Ariakit.SelectLabel>
+      <Ariakit.Select
+        store={select}
+        onKeyDownCapture={onKeyDownCapture}
+        style={{ display: "block" }}
+      />
       <Ariakit.SelectPopover
+        store={select}
         gutter={4}
         sameWidth
+        onKeyDownCapture={onKeyDownCapture}
         style={{
           background: "white",
           border: "1px solid gray",
@@ -59,6 +103,7 @@ export default function Example() {
       >
         {fruits.map((fruit) => (
           <SelectItem
+            store={select}
             key={fruit}
             value={fruit}
             disabled={outOfStock.includes(fruit)}
@@ -67,6 +112,6 @@ export default function Example() {
           />
         ))}
       </Ariakit.SelectPopover>
-    </Ariakit.SelectProvider>
+    </>
   );
 }
--- a/app/src/sandbox/select-6347/index.react.tsx
+++ b/app/src/sandbox/select-6347/index.react.tsx
@@ -1,6 +1,5 @@
 import * as Ariakit from "@ariakit/react";
 import { SelectItem } from "@ariakit/react-components/select/select-item-offscreen";
-import type { KeyboardEvent } from "react";
 
 const fruits = [
   "Apple",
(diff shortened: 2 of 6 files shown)


diegohaz (commented):
Reviewed as an initial full review (no prior skill marker found) against head `fc9df66`, diffing from the merge base `57dfad9`.

## Verdict

Looks good — the fix is correct, minimal, and well-tested. No actionable issues found, so no inline comments.

## Scope reviewed

- `packages/ariakit-react-components/src/composite/composite-item-offscreen.tsx` — the new `data-disabled` placeholder marker.
- `packages/ariakit-react-components/src/composite/composite-typeahead.tsx` — the offscreen-item filtering at insertion time.
- The `select-6347` sandbox + happy-dom (`test.ts`) and browser (`test-browser.ts`) tests.
- The `.changeset/300-select-offscreen-typeahead.md` entry.
- Surrounding context: `disabledFromProps`, `composite-item.tsx` (`trulyDisabled`/store `disabled`), `collection-item-offscreen.tsx`, `select-item-offscreen.tsx`, and `combobox-item-offscreen.tsx`.

## What I verified

- **The fix is load-bearing.** Reverting only the two library files to the base and re-running the sandbox happy-dom test fails exactly the two "skip disabled" cases (`Papaya`/`Papaw` keep `data-active-item`) while the `accessibleWhenDisabled` case still passes; restoring the PR code makes all three pass (`pnpm test-react run select-6347` → 3/3).
- **Consistency with rendered items.** `data-disabled` is set only when `trulyDisabled = disabled && !props.accessibleWhenDisabled`, mirroring `composite-item.tsx`, which registers `disabled: trulyDisabled` in the store. So an `accessibleWhenDisabled` item is a valid typeahead target whether rendered or offscreen, and a truly-disabled one is skipped either way. The `aria-disabled`-only path is covered because `disabledFromProps` reads the `aria-disabled` prop.
- **Dropping `disabled` from the pushed offscreen items is safe.** The only reader of `item.disabled` is `getEnabledItems`, which runs *before* the offscreen push; after the push the items are consumed only via `.element`/`.id`/text, and `CompositeStoreItem.disabled` is optional. The removed property was genuinely dead, matching the issue's analysis.
- **Native fallback retained.** Keeping `if ("disabled" in element && !!element.disabled) continue;` correctly handles custom placeholder renders (e.g. `render={<button disabled />}`) that don't carry the Ariakit `data-disabled` marker.
- **Centralized fix.** `Combobox` offscreen items flow through the same `useCompositeItemOffscreen` hook, so they get the fix for free — no separate change needed.
- **Checks run (all green):** `pnpm tsc`, `pnpm test-react run select-6347` (3/3), oxlint + oxfmt on the changed files. Reference URLs in the changeset (`select-popover`, `select-item`) return 200.

## Non-blocking observations (no change requested)

- `data-disabled` is a placeholder-only attribute — fully rendered disabled `CompositeItem`/`Focusable` items don't emit it (they use `aria-disabled` + the store `disabled` flag instead). This asymmetry is intentional given how typeahead distinguishes the two paths, and no in-repo CSS keys off `[data-disabled]`, so collision risk is negligible. Flagging only so future readers don't assume `data-disabled` is a general disabled marker.
- The single-paragraph changeset frames the fix as `SelectPopover`/`SelectItem`, which is the realistic user-facing surface; the underlying fix also benefits `Combobox`/`CompositeTypeahead` offscreen items. This is acceptable per the "name up to two representative components" guidance.

Areas I did not run locally: the Playwright browser test (`test-browser.ts`) — left to CI — and visual tests. The happy-dom duplicate mirrors it and passes.

PR #6426, opened 2026-06-23
Fix SelectRenderer ARIA counts for mixed groups

### LINKED ISSUE #6301: CompositeRenderer/SelectRenderer emit invalid aria-setsize/aria-posinset when an ungrouped item follows a group
### Current behavior

In a virtualized `CompositeRenderer`/`SelectRenderer` list that mixes grouped and ungrouped items — for example a "Citrus" group followed by standalone options — the rendered options get invalid ARIA. With `[group(Lemon, Lime, Orange), Apple, Banana]` (5 options total), every option renders `aria-setsize="3"` instead of `5`, Apple gets `aria-posinset="4"` — greater than the announced set size, which is invalid ARIA — and Banana gets `aria-posinset="3"`, duplicating Orange's position. Screen readers announce wrong "x of y" positions, while the markup and keyboard navigation look fine, so the breakage is invisible to sighted developers.

The root cause is in [`countItems`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-renderer.tsx#L48-L64), which builds a cumulative array of leaf counts: `count[i]` should be the total number of leaf items in `items[0..i]`. The group branch correctly accumulates on the previous entry (`prevCount + itemsCount`), but the leaf branch writes `count[index] = index + 1`, which equals the cumulative count only while every preceding top-level item is also a leaf. The moment a leaf follows a group, the running total resets to the item's positional index: `[group(3), leaf, leaf]` produces `[3, 2, 3]` instead of `[3, 4, 5]`.

Both consumers of that array then emit broken ARIA: [`setSize` takes the last entry](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-renderer.tsx#L162-L167) (3 instead of 5), and [each item's `aria-posinset` is derived from the entry before its top-level index](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-renderer.tsx#L210-L216), so Apple gets `1 + count[0] = 4 > setsize 3` and Banana gets `1 + count[1] = 3`, overlapping Orange. `SelectRenderer` is built directly on this hook ([`select-renderer.tsx#L103-L108`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/select/select-renderer.tsx#L103-L108)), so a select/combobox menu with a "recent" or featured section followed by other items ships these invalid attributes. Homogeneous lists (all groups, all leaves, or leaves only before the first group) happen to be unaffected, which is why the [Select with Combobox virtualized example](https://ariakit.com/examples/select-combobox-virtualized) — whose groups are uniform — never surfaces it.

3. Click the **Fruit** select to open the popover. Five options render: Lemon, Lime, Orange (inside the "Citrus" group), Apple, and Banana.
4. Inspect the options in the accessibility tree or DOM: every option has `aria-setsize="3"` instead of `5`, Apple has `aria-posinset="4"` (greater than the set size), and Banana has `aria-posinset="3"`, the same position as Orange.

```tsx
import type { SelectRendererItemObject } from "@ariakit/react-components/select/select-renderer";
import { SelectRenderer } from "@ariakit/react-components/select/select-renderer";

interface FruitItem extends SelectRendererItemObject {
  label?: string;
  items?: FruitItem[];
}

function getItem(value: string): FruitItem {
  return { id: `item-${value.toLowerCase()}`, value };
}

const citrusItems = ["Lemon", "Lime", "Orange"].map(getItem);
const otherItems = ["Apple", "Banana"].map(getItem);

const items: FruitItem[] = [
  {
    id: "group-citrus",
    label: "Citrus",
    itemSize: 40,
    paddingStart: 44,
    items: citrusItems,
  },
  ...otherItems,
];

const defaultItems = [...citrusItems, ...otherItems];

  const select = Ariakit.useSelectStore({ defaultItems, defaultValue: "" });
  return (
    <>
      <Ariakit.SelectLabel store={select}>Fruit</Ariakit.SelectLabel>
      <Ariakit.Select store={select} />
      <Ariakit.SelectPopover
        store={select}
        gutter={4}
        sameWidth
        style={{ background: "white", border: "1px solid gray" }}
      >
        <SelectRenderer store={select} items={items}>
          {(item) => {
            if (item.items) {
              const { label, ...groupProps } = item;
              return (
                <SelectRenderer
                  key={groupProps.id}
                  {...groupProps}
                  render={(props) => (
                    <Ariakit.SelectGroup {...props}>
                      <Ariakit.SelectGroupLabel>
                        {label}
                      </Ariakit.SelectGroupLabel>
                      {props.children}
                    </Ariakit.SelectGroup>
                  )}
                >
                  {({ value, ...optionProps }) => (
                    <Ariakit.SelectItem
                      key={optionProps.id}
                      value={value}
                      {...optionProps}
                    />
                  )}
                </SelectRenderer>
              );
            }
            const { value, ...optionProps } = item;
            return (
              <Ariakit.SelectItem
                key={optionProps.id}
                value={value}
                {...optionProps}
              />
            );
          }}
        </SelectRenderer>
      </Ariakit.SelectPopover>
    </>
  );
}
```

Verified with a real-browser Playwright test (open the select, assert the rendered options' attributes): expecting `aria-setsize="5"` and positions 1–5 fails with `aria-setsize="3"` on every option, `aria-posinset="4"` on Apple, and `aria-posinset="3"` on Banana.

### Expected behavior

`aria-setsize` should report the total number of leaf options across groups and standalone items (5 here), and `aria-posinset` should be unique and sequential across the group boundary: Lemon 1, Lime 2, Orange 3, Apple 4, Banana 5. Per the ARIA spec, `aria-posinset` must be between 1 and `aria-setsize`, and two different options shouldn't announce the same position.

### Workaround

Compute the cumulative counts in userland and override the renderer-provided attributes after spreading the item props. Standalone [`SelectItem`](https://ariakit.com/reference/select-item)s get both attributes directly; nested groups get the corrected base position through the nested renderer's own `aria-posinset` prop (the offsets it adds within a group are correct):

```tsx
const totalCount = items.reduce(
  (sum, item) => sum + (item.items?.length ?? 1),
  0,
);

function getPosInSet(index: number) {
  let position = 1;
  for (const item of items.slice(0, index)) {
    position += item.items?.length ?? 1;
  }
  return position;
}

// On the nested group renderer:
<SelectRenderer
  {...groupProps}
  aria-setsize={totalCount}
  aria-posinset={getPosInSet(item.index)}
  // ...
/>

// On standalone items:
<Ariakit.SelectItem
  value={value}
  {...optionProps}
  aria-setsize={totalCount}
  aria-posinset={getPosInSet(item.index)}
/>
```

### Possible solutions

Make the leaf branch of `countItems` accumulate the same way the group branch does — compute `prevCount` once and write `prevCount + 1` for leaves:

```diff
   return items.reduce<number[]>((count, item, index) => {
+    const prevCount = count[index - 1] ?? 0;
     const object = getItemObject(item);
     if (!object.items) {
-      count[index] = index + 1;
+      count[index] = prevCount + 1;
       return count;
     }
-    const prevCount = count[index - 1] ?? 0;
     const itemsCount = countItems(object.items)[object.items.length - 1] ?? 0;
     count[index] = prevCount + itemsCount;
     return count;
   }, []);
```

For pure-leaf lists `prevCount + 1` is identical to `index + 1`, so every currently-valid case (all leaves, all groups, leaves only before groups, numeric `items`) is unchanged — only the previously-invalid mixed cases change: `[group(3), leaf, leaf]` becomes `[3, 4, 5]`. A small regression test asserting `aria-posinset`/`aria-setsize` on a mixed group/leaf list would lock in the fix.

diegohaz:
## Motivation

`SelectRenderer` and `CompositeRenderer` emitted invalid `aria-setsize` and `aria-posinset` values when a virtualized list had a grouped item followed by ungrouped items.

For a rendered item tree like `[group(Lemon, Lime, Orange), Apple, Banana]`, the cumulative count reset when it reached `Apple`. That made the options announce a set size of 3 instead of 5, with duplicated or out-of-range positions across the group boundary.

Fixes #6301.

## Solution

Add `app/src/sandbox/select-renderer-6301` with both browser and happy-dom coverage for the reported user-facing behavior. The sandbox renders a `Citrus` group followed by standalone `Apple` and `Banana` items, then asserts the five options are announced as positions 1 through 5 in a set of 5.

Update `countItems` in `CompositeRenderer` so leaf items accumulate from the previous cumulative count, matching the existing grouped-item branch. Pure leaf lists keep the same counts, while mixed group/leaf lists now produce the correct cumulative sequence.

## Workaround

Until the library fix is released, compute the cumulative count in userland and override the ARIA attributes after spreading the renderer item props. For nested groups, pass the corrected base position into the nested renderer; it will offset each child item correctly within the group.

```tsx
const totalCount = items.reduce(
  (count, item) => count + (item.items?.length ?? 1),
  0,
);

function getPosInSet(index: number) {
  let position = 1;
  for (const item of items.slice(0, index)) {
    position += item.items?.length ?? 1;
  }
  return position;
}

// TODO: Remove once https://github.com/ariakit/ariakit/issues/6301 is fixed.
<SelectRenderer
  {...groupProps}
  aria-setsize={totalCount}
  aria-posinset={getPosInSet(item.index)}
/>

// TODO: Remove once https://github.com/ariakit/ariakit/issues/6301 is fixed.
<Ariakit.SelectItem
  value={value}
  {...optionProps}
  aria-setsize={totalCount}
  aria-posinset={getPosInSet(item.index)}
/>
```

### DIFF
--- a/.changeset/300-select-renderer-aria-count.md
+++ b/.changeset/300-select-renderer-aria-count.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Fixed virtualized `CompositeRenderer` and `SelectRenderer` to report correct `aria-setsize` and `aria-posinset` values when grouped items are followed by standalone items.
--- a/app/src/sandbox/select-renderer-6301/index.react.tsx
+++ b/app/src/sandbox/select-renderer-6301/index.react.tsx
@@ -1,0 +1,84 @@
+import * as Ariakit from "@ariakit/react";
+import { SelectRenderer } from "@ariakit/react-components/select/select-renderer";
+import type { SelectRendererItemObject } from "@ariakit/react-components/select/select-renderer";
+
+interface FruitItem extends SelectRendererItemObject {
+  id: string;
+  label?: string;
+  items?: FruitItem[];
+}
+
+function getItem(value: string): FruitItem {
+  return { id: `item-${value.toLowerCase()}`, value };
+}
+
+const citrusItems = ["Lemon", "Lime", "Orange"].map(getItem);
+const otherItems = ["Apple", "Banana"].map(getItem);
+
+const items: readonly FruitItem[] = [
+  {
+    id: "group-citrus",
+    label: "Citrus",
+    itemSize: 40,
+    paddingStart: 44,
+    items: citrusItems,
+  },
+  ...otherItems,
+];
+
+const defaultItems = [...citrusItems, ...otherItems];
+
+export default function Example() {
+  const select = Ariakit.useSelectStore({ defaultItems, defaultValue: "" });
+
+  return (
+    <>
+      <Ariakit.SelectLabel store={select}>Fruit</Ariakit.SelectLabel>
+      <Ariakit.Select store={select} />
+      <Ariakit.SelectPopover
+        store={select}
+        gutter={4}
+        sameWidth
+        style={{ background: "white", border: "1px solid gray" }}
+      >
+        <SelectRenderer store={select} items={items}>
+          {(item) => {
+            if (item.items) {
+              const { label, ...groupProps } = item;
+              return (
+                <SelectRenderer
+                  key={groupProps.id}
+                  {...groupProps}
+                  render={(props) => (
+                    <Ariakit.SelectGroup {...props}>
+                      <Ariakit.SelectGroupLabel>
+                        {label}
+                      </Ariakit.SelectGroupLabel>
+                      {props.children}
+                    </Ariakit.SelectGroup>
+                  )}
+                >
+                  {({ value, ...optionProps }) => (
+                    <Ariakit.SelectItem
+                      key={optionProps.id}
+                      value={value}
+                      {...optionProps}
+                    />
+                  )}
+                </SelectRenderer>
+              );
+            }
+            const { value, ...optionProps } = item;
+            return (
+              <Ariakit.SelectItem
+                key={optionProps.id}
+                value={value}
+                {...optionProps}
+              />
+            );
+          }}
+        </SelectRenderer>
+      </Ariakit.SelectPopover>
+    </>
+  );
+}
--- a/app/src/sandbox/select-renderer-6301/index.react.tsx
+++ b/app/src/sandbox/select-renderer-6301/index.react.tsx
@@ -45,6 +45,7 @@ export default function Example() {
           store={select}
           items={items}
           initialItems={items.length}
+          persistentIndices={[1]}
         >
           {(item) => {
             if (item.items) {
--- a/app/src/sandbox/select-renderer-6301/index.react.tsx
+++ b/app/src/sandbox/select-renderer-6301/index.react.tsx
@@ -28,6 +28,19 @@ const items: readonly FruitItem[] = [
 
 const defaultItems = [...citrusItems, ...otherItems];
 
+const totalCount = items.reduce(
+  (count, item) => count + (item.items?.length ?? 1),
+  0,
+);
+
+function getPosInSet(index: number) {
+  let position = 1;
+  for (const item of items.slice(0, index)) {
+    position += item.items?.length ?? 1;
+  }
+  return position;
+}
+
 export default function Example() {
   const select = Ariakit.useSelectStore({ defaultItems, defaultValue: "" });
 
@@ -55,6 +68,8 @@ export default function Example() {
                   key={groupProps.id}
                   {...groupProps}
                   initialItems={item.items.length}
+                  aria-setsize={totalCount}
+                  aria-posinset={getPosInSet(item.index)}
                   render={(props) => (
                     <Ariakit.SelectGroup {...props}>
                       <Ariakit.SelectGroupLabel>
@@ -80,6 +95,8 @@ export default function Example() {
                 key={optionProps.id}
                 value={value}
                 {...optionProps}
+                aria-setsize={totalCount}
+                aria-posinset={getPosInSet(item.index)}
               />
             );
           }}
--- a/app/src/sandbox/select-renderer-6301/index.react.tsx
+++ b/app/src/sandbox/select-renderer-6301/index.react.tsx
@@ -41,7 +41,11 @@ export default function Example() {
         sameWidth
         style={{ background: "white", border: "1px solid gray" }}
       >
-        <SelectRenderer store={select} items={items}>
+        <SelectRenderer
+          store={select}
+          items={items}
+          initialItems={items.length}
+        >
           {(item) => {
             if (item.items) {
               const { label, ...groupProps } = item;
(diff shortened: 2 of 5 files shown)


diegohaz (commented):
**Review: no actionable findings.** Full initial review (no prior skill marker found). Baseline: merge-base `57dfad97` → head `3c11b425`.

The one-line `countItems` change is correct and minimal — it is exactly the fix proposed in #6301: hoist `prevCount = count[index - 1] ?? 0` above the leaf branch and write `prevCount + 1` for leaves instead of `index + 1`, so the running leaf total no longer resets to the positional index when a leaf follows a group.

Scope and context checked:
- `countItems` is consumed only via `setSize` (last entry) and each item's `aria-posinset = ariaPosInSet + itemsCount[item.index - 1]`; both now read a correct cumulative sequence. `SelectRenderer` builds directly on `useCompositeRenderer`, so it inherits the fix.
- Behavior preservation: I reproduced `countItems` old vs new across all-leaves, all-groups, leaf-before-group, numeric `items`, empty, `undefined`, and a single leading leaf — all identical. Only the previously-invalid mixed cases change: `[group(3), leaf, leaf]` → `[3,4,5]`, `leaf-group-leaf` → `[1,3,4]`, `group-leaf-group` → `[2,3,6]`, nested-group → `[3,4]`. No regression to any case that was already correct.
- Test value: the new happy-dom `test.ts` and Playwright `test-browser.ts` assert positions 1–5 in a set of 5 across the group boundary. I confirmed `test.ts` is a genuine regression guard — it **fails** against the un-patched source (`aria-setsize="3"` on every option, matching the issue) and **passes** with the fix.
- Conventions: sandbox layout (`index.react.tsx` + `test.ts` + `test-browser.ts`, no `preview.json`) matches sibling `select-renderer-*`/`composite-renderer-*` sandboxes; `persistentIndices={[1]}` is load-bearing to keep the middle leaf rendered under happy-dom's zero-size virtualization. Changeset name/format/package list match repo conventions (`@ariakit/react` is auto-bumped as a re-exporter).

Commands run: `pnpm test-react run app/src/sandbox/select-renderer-6301/test.ts` (pass with fix; fail on reverted source), the same plus the orientation sandbox (2 passed), `pnpm tsc` (0 errors), and a Node script comparing old vs new `countItems` outputs.

Not validated locally: the Playwright `test-browser.ts` (no browser run this session) — but it mirrors the happy-dom assertions, which I verified pass/fail correctly. Visual tests were not run (CI-only per repo policy).

Issue #3147, opened 2023-12-04
Select stops opening when unmountOnHide is true


georgekaran:
### Current behavior

Using a controlled `Select` with the property `unmountOnHide` makes it impossible to reopen the `Select` once it closes for the first time.

__Note__: The bug only happens if the `open` property starts as `true`

### Steps to reproduce the bug

1. Open sandbox: https://stackblitz.com/edit/fw8mei-6fqgoc?file=controlled-selectbox/index.tsx
2. Click outside the `SelectPopover`
3. Attempt to open the `Select`, it won't work.
4. Setting the `open` property to start as `false` makes the `Select` to work as expected.

### Expected behavior

We should be able to set the `open` as `true` and properly toggle the `Select` visibility.

### Workaround

For now, disabling the `unmountOnHide` seems to do the trick.

PR #6427, opened 2026-06-23
Fix store init disposer cleanup

### LINKED ISSUE #6314: @ariakit/store: init disposer is not idempotent — a second call re-runs the whole destroy chain and can break still-mounted widgets
### Current behavior

The disposer returned by `init` from `@ariakit/store` is not idempotent. `createStore` tracks active initializations in a `Set<symbol>`, and each `init()` call returns this disposer ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L163-L167)):

```ts
const maybeDestroy = () => {
  instances.delete(instance);
  if (instances.size) return;
  destroy();
};
```

The boolean returned by `instances.delete(instance)` is ignored. Once the set has drained (the store deactivated), *every* further call to any previously returned disposer — a double call, or a stale disposer kept from an earlier session — falls through both lines and invokes `destroy()` again. `destroy` is the aggregate of all parent-store desyncs, all `setup` teardowns, and all parent-store cleanups ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L214-L223)), and `chain` re-invokes every function each time it's called ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/misc.ts#L136-L144)). So consumer `setup` teardowns re-run, and because the chain includes `stores.map(init)` cleanups, parent stores get recursively double-destroyed through the same unguarded path.

This is inconsistent with the package's other disposers, which are all safe to call repeatedly: the `setup` disposer is a plain `Set.delete` ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L149-L152)) and the subscribe/sync unsubscribe is built from idempotent deletes ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L269-L279)). Nothing documents a call-at-most-once contract for the `init` disposer.

React users are shielded: the only in-repo caller wraps `init` in an effect ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-store/src/index.tsx#L283)), and React guarantees one cleanup per setup. The bug hits consumers using `@ariakit/store` directly (it's also re-exported by `@ariakit/solid-store`) with manual lifecycle management. Disposers are conventionally idempotent (`removeEventListener`, `clearInterval`, `AbortController.abort`, RxJS `unsubscribe`), so calling one from two cleanup paths — an explicit "stop" action plus an unmount/teardown path — is a common, normally harmless pattern. Here it silently re-runs teardowns that are entitled to assume they run once.

In the repro below, two independent panels each register a hotkey in a shared registry via `setup`, activated with `init`. Panel A's disposer runs once from its Stop button and a second time from its unmount effect cleanup. The second call re-runs panel A's teardown: `hotkeys.splice(hotkeys.indexOf(hotkey), 1)` now sees `indexOf === -1`, so `splice(-1, 1)` removes the *last* registry entry — panel B's live hotkey. Panel B is still mounted but stops responding.

3. Press <kbd>a</kbd> and then <kbd>b</kbd>: both counters increment ("A count: 1", "B count: 1").
4. Click "Stop hotkey A", then press <kbd>b</kbd>: panel B still works ("B count: 2").
5. Click "Hide panel A" to unmount panel A (its effect cleanup calls the same disposer a second time).
6. Press <kbd>b</kbd>: panel B's counter no longer increments — its hotkey was unregistered even though the panel is still mounted.

```tsx
import { createStore, init, setup } from "@ariakit/store";
import { useEffect, useRef, useState } from "react";

interface Hotkey {
  key: string;
  handler: () => void;
}

// A single document-level keydown listener dispatches to all registered
// hotkeys, so widgets don't each attach their own listener.
const hotkeys: Hotkey[] = [];

document.addEventListener("keydown", (event) => {
  for (const hotkey of hotkeys) {
    if (hotkey.key === event.key) {
      hotkey.handler();
    }
  }
});

// Framework-agnostic widget state built on @ariakit/store. The setup callback
// registers the widget's hotkey while the store is initialized and removes it
// on teardown.
function createHotkeyCounterStore(key: string) {
  setup(store, () => {
    const hotkey: Hotkey = {
      key,
      handler: () => store.setState("count", (count) => count + 1),
    };
    hotkeys.push(hotkey);
    return () => {
      hotkeys.splice(hotkeys.indexOf(hotkey), 1);
    };
  });
  return store;
}

const storeA = createHotkeyCounterStore("a");
const storeB = createHotkeyCounterStore("b");

function PanelA() {
  const cleanupRef = useRef<(() => void) | null>(null);
  const [active, setActive] = useState(true);

  useEffect(() => {
    cleanupRef.current = init(storeA);
    return () => cleanupRef.current?.();
  }, []);

  const count = Ariakit.useStoreState(storeA, "count");

  return (
    <div>
      <p>A count: {count}</p>
      <p>A active: {active ? "yes" : "no"}</p>
      <button
        type="button"
        onClick={() => {
          // Deactivate the hotkey without unmounting the panel.
          cleanupRef.current?.();
          setActive(false);
        }}
      >
        Stop hotkey A
      </button>
    </div>
  );
}

function PanelB() {
  useEffect(() => init(storeB), []);

  const count = Ariakit.useStoreState(storeB, "count");

  return (
    <div>
      <p>B count: {count}</p>
    </div>
  );
}

  const [panelAMounted, setPanelAMounted] = useState(true);

  return (
    <>
      {panelAMounted && <PanelA />}
      <PanelB />
      <button type="button" onClick={() => setPanelAMounted(false)}>
        Hide panel A
      </button>
    </>
  );
}
```

This was verified with a Playwright test driving only real user interactions (keypresses and clicks): every step works until panel A is hidden, after which the assertion that panel B's counter reaches 3 fails because its hotkey handler is gone.

### Expected behavior

The disposer returned by `init` should be idempotent, like every other disposer in `@ariakit/store` (the `setup` disposer and the `subscribe`/`sync` unsubscribes already are). Redundant or stale disposer calls should be no-ops, so the destroy chain — and the consumer teardowns inside it — runs at most once per initialization session. In the repro, pressing <kbd>b</kbd> after hiding panel A should increment panel B's counter to 3.

### Workaround

Make the consumer-side cleanup idempotent by clearing the stored disposer reference as soon as it's called, so no code path can invoke it twice:

```tsx
useEffect(() => {
  cleanupRef.current = init(storeA);
  return () => {
    cleanupRef.current?.();
    cleanupRef.current = null;
  };
}, []);

// ...and in the Stop button handler:
cleanupRef.current?.();
cleanupRef.current = null;
setActive(false);
```

Verified: with this change the same Playwright test passes.

### Possible solutions

Use the return value of `Set.delete` as a one-shot guard in `maybeDestroy` ([source](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L163-L167)):

```diff
     const maybeDestroy = () => {
-      instances.delete(instance);
+      if (!instances.delete(instance)) return;
       if (instances.size) return;
       destroy();
     };
```

`destroy()` then only runs when a successful delete drains the set, which can happen at most once per unique instance symbol; a disposer whose symbol is no longer in the set has already run (or belongs to a drained session) and does nothing. Legitimate re-initialization is preserved because the next full `init` reassigns `destroy` during setup, and the multi-instance reference counting from #3147 (see the comment at [L155-L158](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L155-L158)) is unaffected.

diegohaz:
## Motivation

The disposer returned by `init` from `@ariakit/store` was not idempotent. A stale or repeated disposer call could run the store destroy chain again after the store had already been torn down, which could re-run consumer setup teardowns and break still-mounted widgets.

Fixes #6314.

## Solution

`createStore` already tracks active `init` calls in a `Set`. This change uses the return value of `Set.delete` as a one-shot guard in the `init` cleanup: if the cleanup's instance is no longer in the set, the cleanup has already run and returns without invoking `destroy()` again.

The PR includes a focused sandbox repro in `app/src/sandbox/store-init-6314`, plus browser and happy-dom tests that exercise the issue through visible hotkey counters. It also adds a direct `@ariakit/store` unit test that verifies repeated stale cleanups only run setup teardowns once.

## Workaround

Make the consumer-side cleanup idempotent by clearing the stored disposer reference as soon as it is called. This prevents later cleanup paths from invoking the same stale disposer again.

```tsx
const cleanup = () => {
  cleanupRef.current?.();
  // TODO: Remove after https://github.com/ariakit/ariakit/issues/6314 lands.
  cleanupRef.current = null;
};

useEffect(() => {
  cleanupRef.current = init(storeA);
  return cleanup;
}, []);

<button
  type="button"
  onClick={() => {
    cleanup();
    setActive(false);
  }}
>
  Stop hotkey A
</button>
```

With this workaround applied in the sandbox, both repro tests pass before the library fix.

### DIFF
--- a/.changeset/300-store-init-cleanup.md
+++ b/.changeset/300-store-init-cleanup.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
+---
+
+Fixed `init` cleanups so repeated or stale calls do not rerun store setup teardowns.
--- a/app/package.json
+++ b/app/package.json
@@ -55,6 +55,7 @@
     "@ariakit/solid": "workspace:*",
     "@ariakit/solid-components": "workspace:*",
     "@ariakit/solid-utils": "workspace:*",
+    "@ariakit/store": "workspace:*",
     "@ariakit/tailwind": "workspace:*",
     "@ariakit/utils": "workspace:*",
     "@astrojs/check": "0.9.9",
--- a/app/src/sandbox/store-init-6314/index.react.tsx
+++ b/app/src/sandbox/store-init-6314/index.react.tsx
@@ -1,0 +1,95 @@
+import { useStoreState } from "@ariakit/react";
+import { createStore, init, setup } from "@ariakit/store";
+import { useEffect, useRef, useState } from "react";
+
+interface Hotkey {
+  key: string;
+  handler: () => void;
+}
+
+const hotkeys: Hotkey[] = [];
+
+function dispatchHotkeys(event: KeyboardEvent) {
+  for (const hotkey of hotkeys) {
+    if (hotkey.key === event.key) {
+      hotkey.handler();
+    }
+  }
+}
+
+function createHotkeyCounterStore(key: string) {
+  const store = createStore({ count: 0 });
+  setup(store, () => {
+    const hotkey: Hotkey = {
+      key,
+      handler: () => store.setState("count", (count) => count + 1),
+    };
+    hotkeys.push(hotkey);
+    return () => {
+      hotkeys.splice(hotkeys.indexOf(hotkey), 1);
+    };
+  });
+  return store;
+}
+
+const storeA = createHotkeyCounterStore("a");
+const storeB = createHotkeyCounterStore("b");
+
+function PanelA() {
+  const cleanupRef = useRef<(() => void) | null>(null);
+  const [active, setActive] = useState(true);
+  const count = useStoreState(storeA, "count");
+
+  useEffect(() => {
+    cleanupRef.current = init(storeA);
+    return () => cleanupRef.current?.();
+  }, []);
+
+  return (
+    <section>
+      <p>A count: {count}</p>
+      <p>A active: {active ? "yes" : "no"}</p>
+      <button
+        type="button"
+        onClick={() => {
+          // Keep the stale disposer reference so unmount calls it again.
+          cleanupRef.current?.();
+          setActive(false);
+        }}
+      >
+        Stop hotkey A
+      </button>
+    </section>
+  );
+}
+
+function PanelB() {
+  const count = useStoreState(storeB, "count");
+
+  useEffect(() => init(storeB), []);
+
+  return (
+    <section>
+      <p>B count: {count}</p>
+    </section>
+  );
+}
+
+export default function Example() {
+  const [panelAMounted, setPanelAMounted] = useState(true);
+
+  useEffect(() => {
+    document.addEventListener("keydown", dispatchHotkeys);
+    return () => document.removeEventListener("keydown", dispatchHotkeys);
+  }, []);
+
+  return (
+    <>
+      {panelAMounted && <PanelA />}
+      <PanelB />
+      <button type="button" onClick={() => setPanelAMounted(false)}>
+        Hide panel A
+      </button>
+    </>
+  );
+}
--- a/app/src/sandbox/store-init-6314/index.react.tsx
+++ b/app/src/sandbox/store-init-6314/index.react.tsx
@@ -40,9 +40,15 @@ function PanelA() {
   const [active, setActive] = useState(true);
   const count = useStoreState(storeA, "count");
 
+  const cleanup = () => {
+    cleanupRef.current?.();
+    // TODO: Remove after https://github.com/ariakit/ariakit/issues/6314 lands.
+    cleanupRef.current = null;
+  };
+
   useEffect(() => {
     cleanupRef.current = init(storeA);
-    return () => cleanupRef.current?.();
+    return cleanup;
   }, []);
 
   return (
@@ -52,8 +58,7 @@ function PanelA() {
       <button
         type="button"
         onClick={() => {
-          // Keep the stale disposer reference so unmount calls it again.
-          cleanupRef.current?.();
+          cleanup();
           setActive(false);
         }}
       >
--- a/app/src/sandbox/store-init-6314/index.react.tsx
+++ b/app/src/sandbox/store-init-6314/index.react.tsx
@@ -40,15 +40,9 @@ function PanelA() {
   const [active, setActive] = useState(true);
   const count = useStoreState(storeA, "count");
 
-  const cleanup = () => {
-    cleanupRef.current?.();
-    // TODO: Remove after https://github.com/ariakit/ariakit/issues/6314 lands.
-    cleanupRef.current = null;
-  };
-
   useEffect(() => {
     cleanupRef.current = init(storeA);
-    return cleanup;
+    return () => cleanupRef.current?.();
   }, []);
 
   return (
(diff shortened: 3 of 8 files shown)


diegohaz (commented):
Initial review (full), comment-only. Diff baseline `merge-base` = `57dfad977`; head `c04b71f6f`.

**Verdict:** LGTM. The fix is correct, minimal, well-targeted, and well-tested. No actionable findings.

**Scope reviewed**
- Core fix `packages/ariakit-store/src/index.ts` (`maybeDestroy` one-shot guard).
- `packages/ariakit-store/src/test.ts` unit test, plus the `app/src/sandbox/store-init-6314` repro (`index.react.tsx`, `test.ts`, `test-browser.ts`).
- `.changeset/300-store-init-cleanup.md`, `app/package.json` dependency add.

**What I verified**
- The guard `if (!instances.delete(instance)) return;` makes the `init` disposer idempotent without breaking re-initialization (a fresh `init` allocates a new instance `symbol` and reassigns `destroy`) or the multi-instance reference counting from #3147 — the guard only short-circuits a repeated delete of the *same* symbol, which is exactly the stale/double-call case. Consistent with the package's other idempotent disposers (`setup`, `subscribe`/`sync`).
- Reverted the one-line fix locally and confirmed both new tests catch the regression: the unit test goes from `["setup", "teardown"]` to `["setup", "teardown", "teardown"]`, and the happy-dom sandbox test fails at "B count: 3" (panel B's hotkey unregistered) — then both pass with the fix restored.
- `packages/ariakit-store/src/test.ts` (44/44) and the happy-dom sandbox `test.ts` under the React loader both pass. `pnpm tsc` clean.
- Changeset scope is right: `@ariakit/store` plus `@ariakit/solid-store` (which `export *`-re-exports `@ariakit/store`). React consumers are shielded (the only in-repo caller wraps `init` in an effect with one cleanup per setup), so omitting React packages is reasonable; the changeset-bot's 7-package expansion is just transitive version-bumping. `init`/`setup`/`createStore` have no `ariakit.com/reference` pages (all 404), so plain code spans rather than links are correct.
- Sandbox follows conventions: issue-number suffix naming, `index.react.tsx` + `test.ts` + `test-browser.ts` with no `preview.json`, and a realistic user-facing repro (visible hotkey counters) rather than a global test harness. `app` is in `.changeset` `ignore` and is private, so it correctly needs no changeset.

**Not fully validated**
- Did not run the Playwright `test-browser.ts` in a browser (no dev server started); its assertions mirror the happy-dom `test.ts`, which I ran and which reproduces the bug, so coverage is established.

PR #6429, opened 2026-06-23
Fix store superseded fan-out propagation

### LINKED ISSUE #6331: @ariakit/store: superseded setState fan-out on a multi-parent store skips the remaining parents, leaving them permanently stale
### Current behavior

A store created with parent stores (`createStore(initialState, ...stores)`, the machinery behind `mergeStore`) fans every public `setState` out to its parents so they stay in sync. The fan-out loop detects when a nested reentrant update superseded the in-flight value and `break`s, on the assumption that the nested update "owns the final notification and propagation":

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L402-L419

That assumption only holds for one of the two ways a supersede can arrive. When a parent listener calls a nested *public* `child.setState` (the case covered by the existing test, [test.ts#L1002-L1032](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/test.ts#L1002-L1032)), the nested call runs its own fan-out over all parents, and breaking out of the stale outer loop is correct.

But a supersede can also arrive through the parent-to-child sync-down that `storeInit` wires up, which calls `setState` with the private `fromStores = true` flag ([index.ts#L181-L193](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L181-L193)). `fromStores` updates are by design barred from fanning back out — that is the documented infinite-loop guard ([index.ts#L378-L382](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L378-L382)). So when the supersede is `fromStores`-driven, *nobody* propagates to the remaining parents: the nested update can't, and the outer loop just `break`ed. Since the sync-down wiring is strictly parent→child, no other mechanism ever re-syncs the skipped parents.

The minimal store-level demonstration (the standard clamping/validation listener pattern):

```ts
import { createStore, init, sync } from "@ariakit/store";

const first = createStore({ count: 0 });
const second = createStore({ count: 0 });
const child = createStore({ count: 0 }, first, second);
init(child);

sync(first, ["count"], (state) => {
  if (state.count > 5) first.setState("count", 5);
});

child.setState("count", 10);
// Actual: child = 5, first = 5, second = 0 — `second` never received any
// setState call at all. Expected: all three converge to 5.
```

Step by step: `child.setState("count", 10)` commits 10 and fans out; `first.setState("count", 10)` commits and notifies; the clamping listener rewrites `first` to 5; `first`'s notification reaches the child's sync-down wiring, which commits 5 to the child with `fromStores = true` (no fan-out); control returns to the outer fan-out loop, which sees `state[key]` (5) ≠ `nextValue` (10), sets `superseded = true`, and `break`s — before ever calling `second.setState`. The mirror case (clamper on `second`) leaves `first` stale at the *unclamped* 10: `{child: 5, first: 10, second: 5}`.

The desync is persistent, not transient: every later over-limit write supersedes again, so the skipped parent stays frozen until some in-range write happens to re-converge it. Nothing throws, and the child plus one parent look correct, so this surfaces as a hard-to-explain stale-UI report in any app that merges two or more stores sharing a key (`mergeStore` forwards all parents to `createStore`, [index.ts#L603-L615](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L603-L615)).

The `break` was introduced in [`109158b0e`](https://github.com/ariakit/ariakit/commit/109158b0ec24879b307317de552dc70373583ed1) (#6205, "Fix duplicate child store listener notifications"). Its regression tests only exercise the public-nested-setState supersede, which fans out by itself — the `fromStores` supersede path was never covered.

4. Click "Add 3". All three panels show 3 — within the stock limit, the central store fans the value out to both panel stores correctly.
5. Click "Add 3" again. The cart's clamp listener rewrites 6 down to the limit of 5: "Selected" and "Cart" show 5, but "Order summary" stays frozen at 3.
6. Keep clicking "Add 3". The summary never recovers — every over-limit write supersedes the fan-out again before reaching the summary store.

```tsx
import type { Store } from "@ariakit/store";
import { createStore, init, sync } from "@ariakit/store";

const MAX_IN_STOCK = 5;

interface QuantityState {
  quantity: number;
}

// Each panel owns its store so it can be subscribed to independently.
const cartStore = createStore<QuantityState>({ quantity: 0 });
const summaryStore = createStore<QuantityState>({ quantity: 0 });

// Central store the page controls write to. It extends both panel stores so
// all three are kept in sync (the rest parameter of createStore).
const quantityStore = createStore<QuantityState>(
  { quantity: 0 },
  cartStore,
  summaryStore,
);

// Business rule owned by the cart: the quantity can never exceed the
// available stock. Whenever the cart receives a higher value, it immediately
// rewrites it to the limit.
sync(cartStore, ["quantity"], (state) => {
  if (state.quantity <= MAX_IN_STOCK) return;
  cartStore.setState("quantity", MAX_IN_STOCK);
});

function useQuantity(store: Store<QuantityState>) {
  const [quantity, setQuantity] = useState(() => store.getState().quantity);
  useEffect(() => {
    return sync(store, ["quantity"], (state) => {
      setQuantity(state.quantity);
    });
  }, [store]);
  return quantity;
}

  useEffect(() => init(quantityStore), []);
  const quantity = useQuantity(quantityStore);
  const cartQuantity = useQuantity(cartStore);
  const summaryQuantity = useQuantity(summaryStore);
  return (
    <div>
      <Ariakit.Button
        onClick={() => {
          quantityStore.setState("quantity", (quantity) => quantity + 3);
        }}
      >
        Add 3
      </Ariakit.Button>
      <p>
        Selected: <output aria-label="Selected">{quantity}</output>
      </p>
      <p>
        Cart: <output aria-label="Cart">{cartQuantity}</output>
      </p>
      <p>
        Order summary:{" "}
        <output aria-label="Order summary">{summaryQuantity}</output>
      </p>
    </div>
  );
}
```

The frozen summary panel was verified with a real-browser Playwright test that only clicks the button and asserts on the rendered DOM.

### Expected behavior

All stores wired together through `createStore`'s parent extension should converge after a `setState`, including when a parent listener rewrites the value: the documented purpose of the wiring is that the stores are "kept in sync". In the repro, after the limit-crossing click all three panels should display the clamped value 5 (`{child: 5, first: 5, second: 5}` in the minimal example), as already happens when the supersede comes from a nested public `setState`.

### Workaround

After writing to the merged store, re-push the final committed value to every parent. Parents already holding it no-op via the store's same-value check, so the pass is cheap:

```tsx
onClick={() => {
  quantityStore.setState("quantity", (quantity) => quantity + 3);
  // A parent listener rewriting the value supersedes the fan-out, leaving
  // sibling parent stores stale. Re-push the final committed value to every
  // parent; the ones already holding it no-op.
  const finalQuantity = quantityStore.getState().quantity;
  cartStore.setState("quantity", finalQuantity);
  summaryStore.setState("quantity", finalQuantity);
}}
```

Verified: with this change every panel shows the clamped value after each click. Alternatively, registering the rewriting listener on the merged child store instead of a parent also converges, because the child's nested `setState` is public and runs its own fan-out — but that changes which store owns the rule.

### Possible solutions

When the fan-out is superseded, the nested `fromStores` update cannot propagate on its own, so the outer call must finish the job: after the loop, re-push the final committed value to every parent. Parents that already hold it no-op via the `isSameValue` early return ([index.ts#L391](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L391)), so the pass is cheap and terminates. Pushing to *all* parents rather than resuming from the break point matters: when the rewriting listener sits on a later parent, it is an *earlier* parent that is stale.

```diff
       if (!fromStores && stores.length) {
         for (const store of stores) {
           store?.setState?.(key, nextValue);
           // Parent fan-out can reenter this child with a newer value for the
-          // same key. That nested update owns the final notification and
-          // propagation, so stop replaying the stale outer value.
+          // same key. That nested update owns the final notification, so stop
+          // replaying the stale outer value.
           if (isSameValue(state[key], nextValue)) continue;
           superseded = true;
           break;
         }
+        // A fromStores-driven supersede cannot fan out on its own. Push the
+        // final committed value to every parent so stores after the
+        // superseding one don't miss the update.
+        if (superseded) {
+          const finalValue = state[key];
+          for (const store of stores) {
+            store?.setState?.(key, finalValue);
+          }
+        }
       }
```

Verified against this commit: with the patch, both desync scenarios (and the mirror case) converge to all-5, the public-nested-supersede path is unchanged (its second pass degenerates into same-value no-ops because the nested update already fanned out), and all 43 existing `@ariakit/store` tests pass. The single bounded pass is deliberate — a fixpoint loop would not terminate when two parent listeners fight over the value, while `fromStores` updates can never re-enter the fan-out, so the single pass always terminates. The `superseded` skip of the outer notification remains correct, because nested `fromStores` commits notify the child's listeners themselves.

diegohaz:
## Motivation

Fixes #6331.

A child store created with multiple parent stores can leave one parent stale when another parent listener rewrites the same key during fan-out. The rewrite reaches the child through the parent-to-child sync path, which intentionally does not fan back out, while the original fan-out stops replaying the stale value. That can leave sibling parent stores permanently out of sync.

## Solution

Add regression coverage at both the store and user-facing sandbox levels:

- store tests for first-parent, later-parent, cascading, and repeated repair rewrites;
- a React sandbox matching the reported cart/order-summary flow;
- happy-dom and browser tests that click "Add 3" twice and assert that every visible output converges to the clamped value.

Fix `createStore` so a superseded parent fan-out repairs every parent with the latest committed child value. The repair runs until a full pass completes without another rewrite, with a bounded safety cap so parent listeners that fight over a key cannot spin forever.

## Workaround

For users who cannot update immediately, after writing to the merged child store, read the child's final value and push it to every parent store. Parents that already hold the value no-op.

```tsx
quantityStore.setState("quantity", (quantity) => quantity + 3);
// TODO: Remove after https://github.com/ariakit/ariakit/issues/6331 is fixed.
const finalQuantity = quantityStore.getState().quantity;
cartStore.setState("quantity", finalQuantity);
summaryStore.setState("quantity", finalQuantity);
```

## Verification

```sh
pnpm test packages/ariakit-store/src/test.ts
pnpm test-react app/src/sandbox/store-6331/test.ts
APP_PORT=4326 NEXTJS_PORT=3008 pnpm -F app run test-chrome app/src/sandbox/store-6331/test-browser.ts
pnpm tsc
```

### DIFF
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -84,6 +84,8 @@ function isSameValue(value: unknown, other: unknown) {
   return value === other || (value !== value && other !== other);
 }
 
+const MAX_REPAIR_PASSES = 100;
+
 function addKeyedListener<S>(
   map: ListenerMap<S>,
   keys: Array<keyof S> | null,
@@ -423,11 +425,28 @@ export function createStore<S extends State>(
         for (const store of stores) {
           store?.setState?.(key, nextValue);
           // Parent fan-out can reenter this child with a newer value for the
-          // same key. That nested update owns the final notification and
-          // propagation, so stop replaying the stale outer value.
+          // same key. That nested update owns the final notification, so stop
+          // replaying the stale outer value.
           if (isSameValue(state[key], nextValue)) continue;
           superseded = true;
           break;
+        }
+        // A fromStores-driven supersede can't fan out on its own. Push the
+        // latest committed value to every parent until a repair pass completes
+        // without another rewrite. Keep this bounded because parent listeners
+        // can fight over a key indefinitely.
+        if (superseded) {
+          for (let pass = 0; pass < MAX_REPAIR_PASSES; pass += 1) {
+            let changed = false;
+            for (const store of stores) {
+              const previousValue = state[key];
+              store?.setState?.(key, state[key]);
+              if (!isSameValue(state[key], previousValue)) {
+                changed = true;
+              }
+            }
+            if (!changed) break;
+          }
         }
       }
 
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -436,16 +436,26 @@ export function createStore<S extends State>(
         // without another rewrite. Keep this bounded because parent listeners
         // can fight over a key indefinitely.
         if (superseded) {
-          for (let pass = 0; pass < MAX_REPAIR_PASSES; pass += 1) {
+          let pass = 0;
+          for (; pass < MAX_REPAIR_PASSES; pass += 1) {
             let changed = false;
             for (const store of stores) {
               const previousValue = state[key];
-              store?.setState?.(key, state[key]);
+              store?.setState?.(key, previousValue);
               if (!isSameValue(state[key], previousValue)) {
                 changed = true;
               }
             }
             if (!changed) break;
+          }
+          if (
+            process.env.NODE_ENV !== "production" &&
+            pass === MAX_REPAIR_PASSES
+          ) {
+            console.warn(
+              "Parent stores did not converge after a superseded fan-out; " +
+                "a parent listener may be rewriting this key in a cycle.",
+            );
           }
         }
       }
--- a/.changeset/300-store-superseded-fan-out.md
+++ b/.changeset/300-store-superseded-fan-out.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
+---
+
+Fixed `createStore` to keep parent stores in sync when a parent listener rewrites a value during a multi-parent fan-out.
--- a/app/src/sandbox/store-6331/index.react.tsx
+++ b/app/src/sandbox/store-6331/index.react.tsx
@@ -44,11 +44,6 @@ export default function Example() {
         className="rounded bg-blue-600 px-3 py-1 text-white"
         onClick={() => {
           quantityStore.setState("quantity", (quantity) => quantity + 3);
-          // TODO: Remove after https://github.com/ariakit/ariakit/issues/6331
-          // is fixed.
-          const finalQuantity = quantityStore.getState().quantity;
-          cartStore.setState("quantity", finalQuantity);
-          summaryStore.setState("quantity", finalQuantity);
         }}
       >
         Add 3
--- a/app/src/sandbox/store-6331/index.react.tsx
+++ b/app/src/sandbox/store-6331/index.react.tsx
@@ -44,6 +44,11 @@ export default function Example() {
         className="rounded bg-blue-600 px-3 py-1 text-white"
         onClick={() => {
           quantityStore.setState("quantity", (quantity) => quantity + 3);
+          // TODO: Remove after https://github.com/ariakit/ariakit/issues/6331
+          // is fixed.
+          const finalQuantity = quantityStore.getState().quantity;
+          cartStore.setState("quantity", finalQuantity);
+          summaryStore.setState("quantity", finalQuantity);
         }}
       >
         Add 3
--- a/app/src/sandbox/store-6331/index.react.tsx
+++ b/app/src/sandbox/store-6331/index.react.tsx
@@ -1,0 +1,63 @@
+import * as Ariakit from "@ariakit/react";
+import type { Store } from "@ariakit/store";
+import { createStore, init, sync } from "@ariakit/store";
+import { useEffect, useState } from "react";
+
+const MAX_IN_STOCK = 5;
+
+interface QuantityState {
+  quantity: number;
+}
+
+const cartStore = createStore<QuantityState>({ quantity: 0 });
+const summaryStore = createStore<QuantityState>({ quantity: 0 });
+const quantityStore = createStore<QuantityState>(
+  { quantity: 0 },
+  cartStore,
+  summaryStore,
+);
+
+sync(cartStore, ["quantity"], (state) => {
+  if (state.quantity <= MAX_IN_STOCK) return;
+  cartStore.setState("quantity", MAX_IN_STOCK);
+});
+
+function useQuantity(store: Store<QuantityState>) {
+  const [quantity, setQuantity] = useState(() => store.getState().quantity);
+  useEffect(() => {
+    return sync(store, ["quantity"], (state) => {
+      setQuantity(state.quantity);
+    });
+  }, [store]);
+  return quantity;
+}
+
+export default function Example() {
+  useEffect(() => init(quantityStore), []);
+  const quantity = useQuantity(quantityStore);
+  const cartQuantity = useQuantity(cartStore);
+  const summaryQuantity = useQuantity(summaryStore);
+
+  return (
+    <div className="flex flex-col items-start gap-3">
+      <Ariakit.Button
+        className="rounded bg-blue-600 px-3 py-1 text-white"
+        onClick={() => {
+          quantityStore.setState("quantity", (quantity) => quantity + 3);
+        }}
+      >
+        Add 3
+      </Ariakit.Button>
+      <p>
+        Selected: <output aria-label="Selected">{quantity}</output>
+      </p>
+      <p>
+        Cart: <output aria-label="Cart">{cartQuantity}</output>
+      </p>
+      <p>
+        Order summary:{" "}
+        <output aria-label="Order summary">{summaryQuantity}</output>
+      </p>
+    </div>
+  );
+}
--- a/app/src/sandbox/store-6331/test-browser.ts
+++ b/app/src/sandbox/store-6331/test-browser.ts
@@ -1,0 +1,18 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6331
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("keeps parent stores in sync when a parent clamps the child value", async ({
+    q,
+  }) => {
+    await q.button("Add 3").click();
+    await test.expect(q.status("Selected")).toHaveText("3");
+    await test.expect(q.status("Cart")).toHaveText("3");
+    await test.expect(q.status("Order summary")).toHaveText("3");
+
+    await q.button("Add 3").click();
+    await test.expect(q.status("Selected")).toHaveText("5");
+    await test.expect(q.status("Cart")).toHaveText("5");
+    await test.expect(q.status("Order summary")).toHaveText("5");
+  });
+});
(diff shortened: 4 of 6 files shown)


packages/ariakit-store/src/index.ts
```diff
@@ -410,12 +412,29 @@ export function createStore<S extends State>(
...
           if (isSameValue(state[key], nextValue)) continue;
           superseded = true;
           break;
         }
+        // A fromStores-driven supersede can't fan out on its own. Push the
+        // latest committed value to every parent until a repair pass completes
+        // without another rewrite. Keep this bounded because parent listeners
+        // can fight over a key indefinitely.
+        if (superseded) {
+          for (let pass = 0; pass < MAX_REPAIR_PASSES; pass += 1) {
```
diegohaz: When parent listeners genuinely fight over a key (e.g. `first` always rewrites to 1, `second` always rewrites to 2), this loop exhausts all `MAX_REPAIR_PASSES` and then silently leaves the parents in a desynced state with no signal. I confirmed this locally with two oscillating clamp listeners: the call returns with `{child: 2, first: 1, second: 2}` after 100 passes, no throw, no warning.

That matches the pre-fix `break` behavior (which also left a silent desync), so this is not a regression and the bounded cap is the right call to avoid hanging. But since this store already has a dev-only diagnostic convention (`invariant`, and the `process.env.NODE_ENV === "production"` early-return guard further down), it would be worth emitting a dev-only `console.warn` when the cap is hit, so an app author with an unconvergeable pair of parent listeners gets a hint instead of a mysteriously stale store. Something like:

```ts
if (superseded) {
  let pass = 0;
  for (; pass < MAX_REPAIR_PASSES; pass += 1) {
    let changed = false;
    for (const store of stores) {
      const previousValue = state[key];
      store?.setState?.(key, previousValue);
      if (!isSameValue(state[key], previousValue)) changed = true;
    }
    if (!changed) break;
  }
  if (process.env.NODE_ENV !== "production" && pass === MAX_REPAIR_PASSES) {
    console.warn(
      "Parent stores did not converge after a superseded fan-out; " +
        "a parent listener may be rewriting this key in a cycle.",
    );
  }
}
```

Optional, but it turns an invisible failure mode into an actionable one.

packages/ariakit-store/src/index.ts
```diff
@@ -410,12 +412,29 @@ export function createStore<S extends State>(
...
+            let changed = false;
+            for (const store of stores) {
+              const previousValue = state[key];
+              store?.setState?.(key, state[key]);
+              if (!isSameValue(state[key], previousValue)) {
```
diegohaz, line 445: Minor readability: `previousValue` is captured from `state[key]` and then `state[key]` is read again on the very next line to pass into `setState`. Since nothing mutates `state` between the two reads, they're guaranteed equal, so passing `previousValue` is behavior-identical and makes the intent ("re-push the value we just snapshotted") explicit:

```suggestion
              const previousValue = state[key];
              store?.setState?.(key, previousValue);
              if (!isSameValue(state[key], previousValue)) {
```

Reads slightly clearer and avoids implying the two `state[key]` reads could differ.

diegohaz (commented):
## Review summary

Initial review of the superseded fan-out fix. Scope reviewed: `packages/ariakit-store/src/index.ts` (the repair loop), the 6 new `@ariakit/store` unit tests, the `store-6331` sandbox (`index.react.tsx` + happy-dom `test.ts` + `test-browser.ts`), the `app/package.json` workspace dependency, and the changeset.

**Verdict:** The fix is correct and the regression coverage is solid. No bugs found. Two minor, optional inline notes only.

### What I verified locally (Node 24, against the store source)
- The four core scenarios (`parent-driven`, `later-parent`, `cascading`, `restarts`) genuinely **fail on `main`** (e.g. `parent-driven` -> `{child:5, first:5, second:0}`) and **pass on this head** -> all converge. The new unit tests are real regression coverage, not vacuous.
- The issue's proposed **single-pass** fix (`finalValue = state[key]; push to all`) was tested by patching a local copy: it **fails** `restarts` (`[4,5,4]`), `finite-cascades` (`[2,2,5]`), and `one-shot-cycles` (`[4,4,5]`), leaving permanent desyncs. So this PR's deviation to a bounded multi-pass loop is justified — a second parent listener rewriting during the repair pass needs another pass.
- Termination: two genuinely-oscillating clamp listeners hit the `MAX_REPAIR_PASSES` cap and return in <1ms with a bounded (still-desynced) state, no hang/throw. The cap is the right safety choice.
- No new duplicate notifications: the clamping parent's double-`5` notification I observed is **pre-existing base-branch reentrancy behavior**, not introduced by the repair loop (the loop's same-value re-push no-ops via `isSameValue`).
- `pnpm test packages/ariakit-store/src/test.ts` -> 49/49 pass. `pnpm test-react app/src/sandbox/store-6331/test.ts` -> 1/1 pass. `pnpm tsc` -> 0 errors. `oxlint` on changed files -> clean. Changeset correctly patches `@ariakit/store` + `@ariakit/solid-store` (which re-exports `@ariakit/store`); the workspace dep + lockfile entry for the sandbox are correct.

### Inline notes (both optional, non-blocking)
1. The cap exhaustion path silently leaves a desync with no dev diagnostic — a dev-only `console.warn` would fit this file's existing `invariant` / `NODE_ENV` convention.
2. A small readability tweak in the repair loop (`state[key]` is read twice on consecutive lines; pass the already-captured `previousValue`).

Areas I did not run: the Playwright `test-browser.ts` (no browser run needed — the happy-dom duplicate covers the same assertions and passed; the logic is engine-independent store behavior).

packages/ariakit-store/src/index.ts
diegohaz: Fixed in f9c07c7aa - added a dev-only warning when parent repair exhausts `MAX_REPAIR_PASSES`, with dev/prod tests covering the non-converging parent cycle.

packages/ariakit-store/src/index.ts
diegohaz: Fixed in f9c07c7aa - the repair loop now reuses `previousValue` for the parent `setState` call, so the pushed value matches the snapshot used for the change check.

diegohaz:
**Fixed**

- f9c07c7aa ([reply](https://github.com/ariakit/ariakit/pull/6429#discussion_r3463614020)): Added a dev-only warning when parent repair exhausts `MAX_REPAIR_PASSES`, plus dev/prod regression tests for the non-converging parent cycle.
- f9c07c7aa ([reply](https://github.com/ariakit/ariakit/pull/6429#discussion_r3463614830)): Reused `previousValue` for the parent `setState` call so the pushed value matches the snapshot used for the change check.

PR #6432, opened 2026-06-24
Fix store unsubscribe cleanup reentry

### LINKED ISSUE #6431: @ariakit/store: unsubscribe runs a listener's cleanup before detaching it, so a store update inside the cleanup reenters dispatch (double cleanup, spurious re-run, leaked cleanup)
### Current behavior

A `@ariakit/store` listener can return a cleanup function, and that cleanup is expected to run before the listener runs again and when it unsubscribes. The unsubscribe disposer returned by `registerListener` runs the cleanup *while the listener is still registered* ([index.ts#L284-L294](https://github.com/ariakit/ariakit/blob/70a2a74770c23d7e4dbd45b45ac411100597aea4/packages/ariakit-store/src/index.ts#L284-L294)):

```ts
return () => {
  group.disposables.get(listener)?.();   // <-- cleanup runs here, still registered
  group.disposables.delete(listener);
  const currentKeys = group.listenerKeys.get(listener);
  deleteListenerIndexes(group, listener, listenerKeysValue);
  if (currentKeys !== listenerKeysValue) {
    deleteListenerIndexes(group, listener, currentKeys);
  }
  group.listenerKeys.delete(listener);
  group.listeners.delete(listener);
};
```

At the first line the cleanup executes while the listener is still in `group.listeners`, still indexed in `listenersByKey`/`allKeysListeners`, and its cleanup is still in `disposables`. If that cleanup synchronously calls `setState`, the listener reenters the dispatch path and three things go wrong:

1. **The cleanup runs twice.** The reentrant `setState` calls `runListeners` → `runListener`, whose `runPendingCleanups` finds the *same* cleanup still in `disposables` (the disposer hasn't reached `disposables.delete` yet) and runs it again — a double-free for any real resource (e.g. `removeEventListener` called twice, a refcount decremented twice).
2. **The listener runs again after `unsubscribe()`.** The same reentrant dispatch invokes the listener a second time, even though the caller already unsubscribed.
3. **The replacement cleanup leaks.** That spurious second run installs a new cleanup in `disposables`; the disposer's next line (`disposables.delete(listener)`) then discards it **without running it**.

This is independent of #6323/#6430 — it is the *unsubscribe* path, not re-registration. The disposer is unchanged on `main`, on #6430 (`fix-store-cleanup-reregister`), and on #6424 (`fix-store-all-keys-fast-path-6317`), so all of them exhibit it. With the cleanup writing to the store on every teardown (rather than just once) it recurses until the stack overflows.

For a low-level store package the consumer reproduction is a unit test against the public `createStore`/`sync` API. The following test **fails on `main`** because the listener's cleanup runs twice and the listener runs a second time during `unsubscribe()`:

```ts
import { createStore, sync } from "@ariakit/store";

test("a cleanup that updates the store during unsubscribe runs once", () => {
  let runs = 0;
  let cleanups = 0;
  let nudged = false;

  const unsubscribe = sync(store, ["count"], () => {
    runs += 1;
    return () => {
      cleanups += 1;
      // A real-world cleanup that nudges the store once during teardown.
      if (nudged) return;
      nudged = true;
      store.setState("count", (count) => count + 1);
    };
  });

  expect(runs).toBe(1);
  unsubscribe();

  // The cleanup must run exactly once and the listener must not run again.
  expect(cleanups).toBe(1); // ❌ receives 2 on main
  expect(runs).toBe(1);     // ❌ receives 2 on main
});
```

```
AssertionError: expected 2 to be 1 // Object.is equality
```

### Expected behavior

Unsubscribing should fully detach the listener before running its cleanup, so a `setState` issued from inside the cleanup observes an already-removed listener: the cleanup runs exactly once, the listener is not re-invoked, and no replacement cleanup is left registered. (This matches how cleanups behave everywhere else — they run against a settled registration state.)

### Workaround

Don't write to the store **synchronously** from inside a listener cleanup; defer the write so it can't reenter the dispatch while the listener is still registered. Tested — this passes on `main`:

```ts
const unsubscribe = sync(store, ["count"], () => {
  runs += 1;
  return () => {
    cleanups += 1;
    if (nudged) return;
    nudged = true;
    // Defer out of the cleanup; by the time it runs the listener is detached.
    queueMicrotask(() => store.setState("count", (count) => count + 1));
  };
});
```

The deferred update lands after `unsubscribe()` has finished detaching the listener, so there is no double cleanup, no spurious re-run, and no leak. (`Promise.resolve().then(...)`/`setTimeout` work too; pick whatever matches the timing you need.)

### Possible solutions

Detach the listener completely before running its cleanup, so a reentrant `setState` from the cleanup can't find it registered:

```diff
     return () => {
-      group.disposables.get(listener)?.();
-      group.disposables.delete(listener);
+      const cleanup = group.disposables.get(listener);
+      group.disposables.delete(listener);
       const currentKeys = group.listenerKeys.get(listener);
       deleteListenerIndexes(group, listener, listenerKeysValue);
       if (currentKeys !== listenerKeysValue) {
         deleteListenerIndexes(group, listener, currentKeys);
       }
       group.listenerKeys.delete(listener);
       group.listeners.delete(listener);
+      cleanup?.();
     };
```

Locally this makes the reproduction above pass and keeps the full `packages/ariakit-store/src/test.ts` suite green (65/65). It pairs naturally with a regression test asserting `cleanups === 1` / `runs === 1` for the scenario above. Note this same function is edited by #6430 and #6424, so whichever lands first should carry the reorder to avoid a conflicting touch. (Originally surfaced by CodeRabbit on #6430.)

diegohaz:
## Motivation

Fixes #6431.

A store listener cleanup can synchronously update the same store while the listener is being unsubscribed. Before this change, the unsubscribe disposer ran the cleanup while the listener was still registered, so a reentrant store update could run the same cleanup twice and invoke the listener again after `unsubscribe()` had started.

## Solution

The unsubscribe disposer now captures the pending cleanup, removes the listener from the cleanup map, key indexes, listener key map, and listener set, then runs the cleanup after the listener has been fully detached.

A regression test covers the public `sync` API with a cleanup that synchronously updates the store during unsubscribe.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm -F @ariakit/store build`
- `git diff --check`

### DIFF
--- a/.changeset/300-store-unsubscribe-cleanup.md
+++ b/.changeset/300-store-unsubscribe-cleanup.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
+---
+
+Fixed store listener cleanup disposal so synchronous store updates during unsubscribe do not rerun detached listeners.
--- a/packages/ariakit-store/src/index.ts
+++ b/packages/ariakit-store/src/index.ts
@@ -518,8 +518,9 @@ export function createStore<S extends State>(
   // (1) snapshot `keys` as a defensive copy (or null for an all-keys listener);
   // (2) index the listener — re-registering the same listener first clears its
   // previous index entries so a re-keyed listener is never left in a stale key
-  // bucket; (3) return a disposer that runs any pending cleanup and removes the
-  // listener from the set, the key index, and the disposables/keys maps.
+  // bucket; (3) return a disposer that removes the listener from the set, the
+  // key index, and the disposables/keys maps before running any pending
+  // cleanup.
   const registerListener = (
     keys: Array<keyof S> | null,
     listener: Listener<S>,
@@ -563,7 +564,7 @@ export function createStore<S extends State>(
     }
     group.listenerKeys.set(listener, listenerKeysValue);
     return () => {
-      group.disposables.get(listener)?.();
+      const cleanup = group.disposables.get(listener);
       group.disposables.delete(listener);
       preserveFastPathFrames(fastPathFrames, group, listener);
       const currentKeys = group.listenerKeys.get(listener);
@@ -573,6 +574,7 @@ export function createStore<S extends State>(
       }
       group.listenerKeys.delete(listener);
       group.listeners.delete(listener);
+      cleanup?.();
     };
   };
 
--- a/packages/ariakit-store/src/test.ts
+++ b/packages/ariakit-store/src/test.ts
@@ -290,6 +290,32 @@ test("sync runs immediately and cleans up before rerun and unsubscribe", () => {
   expect(events).toEqual(["0->0", "cleanup 0", "0->1", "cleanup 1"]);
 });
 
+test("detaches sync listeners before unsubscribe cleanups update state", () => {
+  const store = createStore({ count: 0 });
+  const events: string[] = [];
+  let runCount = 0;
+  let nudged = false;
+
+  const unsubscribe = sync(store, ["count"], () => {
+    runCount += 1;
+    const id = runCount;
+    events.push(`run ${id}`);
+    return () => {
+      events.push(`cleanup ${id}`);
+      if (nudged) return;
+      nudged = true;
+      store.setState("count", (count) => count + 1);
+    };
+  });
+
+  expect(events).toEqual(["run 1"]);
+
+  unsubscribe();
+
+  expect(store.getState()).toEqual({ count: 1 });
+  expect(events).toEqual(["run 1", "cleanup 1"]);
+});
+
 test("subscribe cleans up before rerun and unsubscribe", () => {
   const store = createStore({ count: 0 });
   const events: string[] = [];


diegohaz (commented):
**Review summary — no substantive findings.** This is an **initial review**, diffed from the merge-base `40c3edf` to head `5ac6c88`. The fix is correct, minimal, and well covered.

### What the change does
The unsubscribe disposer now captures the pending cleanup, fully detaches the listener (disposables, fast-path frames, key indexes, `listenerKeys`, `listeners` set), and runs `cleanup?.()` **last**. A synchronous `setState` from inside the cleanup can no longer find the listener registered, so `runListeners` won't re-notify it during teardown. This matches the recommended fix in #6431 and keeps `preserveFastPathFrames` in the position #6430 introduced (before the bucket mutation), so the reorder doesn't disturb the fast-path frame machinery.

### Validation performed
- Reverted `index.ts` to the pre-fix version (`40c3edf`) and ran the new test against it — it **fails** with exactly the reported symptom: `["run 1", "cleanup 1", "cleanup 1", "run 2"]` (double cleanup + spurious re-run). On the fix it passes. So this is a true regression test.
- `pnpm test packages/ariakit-store/src/test.ts` → **91/91 pass** on the fix, including the existing #6430/#6424 fast-path frame tests.
- `tsc --noEmit` on `@ariakit/store` → clean.
- Verified the changeset: `@ariakit/store` + `@ariakit/solid-store` (patch) matches the convention of the other store changesets in `.changeset/`, and `@ariakit/solid-store` is a single-line `export * from "@ariakit/store"`, so the bump is correct and there is no duplicate disposer to fix there.
- Confirmed `registerListener`'s disposer is the only unsubscribe path (`sync`/`subscribe`/`batch` all route through it). `subscribe`/`batch` defer via microtask, so they don't reenter synchronously during unsubscribe — the bug is sync-specific and the sync test is the meaningful coverage.

### Notes (no action needed)
- The new ordering is incidentally more robust in two ways the description doesn't claim: if the cleanup **throws**, the listener is now fully detached before the throw (the old order leaked the registration); and if the cleanup **re-subscribes** the same listener, the detach completes first so there's no interference with the fresh registration.
- The updated explanatory comment accurately describes the new "remove before running cleanup" order.

### Not fully validated
- I did not run the React/Solid render suites or the browser `app` tests, since this is a non-framework change to the core store with the package-level suite green and `tsc` clean.

PR #6435, opened 2026-06-24
Fix collection item metadata lookup

### LINKED ISSUE #6295: Collection store item() drops metadata from the items state and ignores controlled items updates
### Current behavior

`collection.item(id)` returns data that contradicts the collection's own `items` state. With a controlled collection whose items carry custom metadata — the pattern supported by the typed [`useCollectionStore`](https://ariakit.com/reference/use-collection-store) overload that requires `items`/`defaultItems` — the lookup result is missing the metadata that `getState().items` plainly lists:

- As soon as an item is rendered by [`CollectionItem`](https://ariakit.com/reference/collection-item), `item(id)` returns only the bare `{ id, element }` registration object. Metadata provided through the `items` prop (e.g. a `label`) is gone, even though the `items` state still has it.
- Updating an item's metadata through the controlled `items`/`setItems` state is user-visible in the rendered list, but `item(id)` still returns the bare object. Items added purely via `store.setState("items", ...)` that never render aren't found at all (`item(id)` returns `null`).

The root cause is that [`item()` reads only the internal `itemsMap`, with a fallback to the private registration store — it never consults the public `items` state](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L197-L208). The map is written from exactly two places: the [initial props seed](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L63) and [`mergeItem`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L139-L176). Data flows one way only, [private store → public store](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L87-L94); nothing listens to the public `items` state. The controlled `items` prop is delivered precisely through that unobserved path: [`useStoreProps(store, props, "items", "setItems")`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-store.ts#L13) pushes the value with a [plain `store.setState(key, value)` on the public store](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-store/src/index.tsx#L259-L267).

There's a second leg that makes this bite even without any update: [`renderItem` chains `registerItem` with a second `mergeItem` into `renderedItems`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L188-L195). `registerItem` merges the seeded metadata into the map, but the `renderedItems` list starts empty, so the second merge takes the [push branch and overwrites the map entry with the bare `{ id, element }` object](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L153-L156). Instrumenting the repro shows the divergence directly after renaming: `getState().items` is `[{ id: "apple", label: "Green Apple", element: BUTTON }, ...]`, the private store even holds `{ id: "apple", label: "Apple", element: BUTTON }`, yet `item("apple")` returns `{ id: "apple", element: BUTTON }` with no `label` at all.

The same gap has already been treated as a bug elsewhere in the codebase: the tab store's [`createPanelsStore` builds its `panelsByTabId` lookup with a `sync` on the public `["items"]` state](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/tab/tab-store.ts#L48-L69) precisely so explicit `items` updates are reflected, with a dedicated test ([`tab-store.test.ts#L106-L123`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/tab/tab-store.test.ts#L106-L123)), and commit [`79de4f8f6`](https://github.com/ariakit/ariakit/commit/79de4f8f6e650b0d05cba2ed0721001eef8ccfcc) ("Fix `collectionStore.item()` getting out of sync with current items") fixed a sibling instance of the same divergence.

3. Click the **Apple** item. The output should show "Apple" (the metadata listed in the `items` state), but it shows "Item not found".
4. Click **Rename Apple**, then click the **Green Apple** item. The list itself shows the updated label, proving the `items` state was updated, but the output still shows "Item not found" instead of "Green Apple".

```tsx

interface FruitItem {
  id: string;
  label: string;
  element?: HTMLElement | null;
}

  const [items, setItems] = useState<FruitItem[]>([
    { id: "apple", label: "Apple" },
    { id: "grape", label: "Grape" },
    { id: "orange", label: "Orange" },
  ]);
  const [details, setDetails] = useState("");
  const collection = Ariakit.useCollectionStore({ items, setItems });

  function renameApple() {
    setItems((prevItems) =>
      prevItems.map((item) =>
        item.id === "apple" ? { ...item, label: "Green Apple" } : item,
      ),
    );
  }

  function showDetails(id: string) {
    const item = collection.item(id);
    setDetails(item?.label ?? "Item not found");
  }

  return (
    <div>
      <Ariakit.Collection store={collection}>
        {items.map((item) => (
          <Ariakit.CollectionItem
            key={item.id}
            id={item.id}
            onClick={() => showDetails(item.id)}
          >
            {item.label}
          </Ariakit.CollectionItem>
        ))}
      </Ariakit.Collection>
      <button type="button" onClick={renameApple}>
        Rename Apple
      </button>
      <output>{details}</output>
    </div>
  );
}
```

Both steps were verified with a real-browser Playwright test (user clicks and DOM assertions only), which fails on the expected values "Apple"/"Green Apple" and observes "Item not found" in both cases.

### Expected behavior

`item(id)` should return the same data the `items` state lists for that id. The [`items` state](https://ariakit.com/reference/use-collection-store#items) is documented as listing "all items along with their metadata", and `item()` is the documented function to "get an item by its id" — so metadata provided through the `items` prop, the controlled `items`/`setItems` pair, or `store.setState("items", ...)` should be retrievable through `item()`, and updates to it should be reflected.

### Workaround

Don't use `item()` when item metadata lives in the `items` state — look the item up in the state directly:

```tsx
function showDetails(id: string) {
  const item = collection
    .getState()
    .items.find((stateItem) => stateItem.id === id);
  setDetails(item?.label ?? "Item not found");
}
```

Alternatively, register the metadata through the [`getItem`](https://ariakit.com/reference/collection-item#getitem) prop of `CollectionItem` (recreating the callback whenever the metadata changes) so it's part of the registration object that `item()` returns.

### Possible solutions

Reconcile `itemsMap` whenever the public `items` state changes, mirroring the proven `createPanelsStore` pattern. A `sync` listener registered at store creation runs synchronously on every `setState` (and on updates fanned down from composed stores), so it catches direct `setState`, the controlled React `items` prop, and the private→public registration flush — which re-runs the listener with the merged, element-bearing arrays, also fixing the `renderItem` clobber described above:

```diff
--- a/packages/ariakit-components/src/collection/collection-store.ts
+++ b/packages/ariakit-components/src/collection/collection-store.ts
@@ -3,6 +3,7 @@ import {
   createStore,
   init,
   setup,
+  sync,
   throwOnConflictingProps,
 } from "@ariakit/store";
@@ -76,6 +77,19 @@ export function createCollectionStore<
 
   const collection = createStore(initialState, props.store);
 
+  // Keep the id lookup map in sync with items set directly on the public
+  // store (controlled `items` prop or setState), not just registered items.
+  sync(collection, ["items"], (state) => {
+    const ids = new Set<string>();
+    for (const item of state.items) {
+      ids.add(item.id);
+      itemsMap.set(item.id, item);
+    }
+    for (const id of itemsMap.keys()) {
+      if (!ids.has(id)) itemsMap.delete(id);
+    }
+  });
+
   const sortItems = (renderedItems: T[]) => {
```

An alternative is to make `item()` read `collection.getState().items` directly, but that loses the O(1) map lookup in hot keyboard/scroll paths (composite navigation, combobox active value, select typeahead) and was the pre-`79de4f8f6` design; reconciling the map keeps lookups cheap. A regression test should assert that `item(id)` matches `getState().items` for a rendered item with prop-provided metadata, after a direct `setState("items", ...)` update, and for a fresh unregistered id.

diegohaz:
## Motivation

`collection.item(id)` can return data that disagrees with the collection `items` state when a controlled collection passes metadata through `useCollectionStore({ items, setItems })`. In the issue repro, clicking the rendered Apple item reads through `collection.item("apple")`, but the returned item is missing the `label` metadata and the UI shows `Item not found` instead of `Apple`.

Fixes #6295.

## Solution

The collection store now treats the public `items` state as authoritative while preserving private registration metadata for mounted items. When public items change, the id lookup is rebuilt from the public array and merged with registered elements only for ids that still exist in public state.

Registration batches reconcile against the latest public state before publishing, so controlled metadata replacements, removals, pre-init updates, and pending private batches keep `collection.item(id)` current without losing mounted DOM elements. Composite typeahead now ignores disconnected rendered metadata exposed by the stricter collection reconciliation, and closed Select keyboard navigation skips controlled disabled or valueless metadata without returning a rejected item after wrap.

The sandbox in `app/src/sandbox/collection-6295` reproduces the user-visible flow from the issue: render controlled collection items, click an item, rename Apple through controlled state, then click the updated item again. The coverage includes both the primary browser test and a happy-dom duplicate, plus core collection-store regressions for direct item state updates, registered element preservation, removals, pre-init updates, and pending private batches. The `app/src/sandbox/select-disabled-next` sandbox covers the closed Select navigation regression for disabled and valueless controlled metadata.

## Workaround

Until the library fix lands, avoid `collection.item(id)` when the metadata source of truth is the collection `items` state. Read the public state directly and find the item by id:

```tsx
function showDetails(id: string) {
  // TODO: Remove once https://github.com/ariakit/ariakit/issues/6295 is fixed.
  const item = collection
    .getState()
    .items.find((stateItem) => stateItem.id === id);
  setDetails(item?.label ?? "Item not found");
}
```

### DIFF
--- a/packages/ariakit-components/src/collection/collection-store.ts
+++ b/packages/ariakit-components/src/collection/collection-store.ts
@@ -3,6 +3,7 @@ import {
   createStore,
   init,
   setup,
+  sync,
   throwOnConflictingProps,
 } from "@ariakit/store";
 import type { Store, StoreOptions, StoreProps } from "@ariakit/store";
@@ -75,6 +76,14 @@ export function createCollectionStore<
   );
 
   const collection = createStore(initialState, props.store);
+
+  sync(collection, ["items"], (state) => {
+    privateStore.setState("items", state.items);
+    itemsMap.clear();
+    for (const item of state.items) {
+      itemsMap.set(item.id, item);
+    }
+  });
 
   const sortItems = (renderedItems: T[]) => {
     const sortedItems = sortBasedOnDOMPosition(renderedItems, (i) => i.element);
--- a/packages/ariakit-components/src/collection/collection-store.ts
+++ b/packages/ariakit-components/src/collection/collection-store.ts
@@ -46,12 +46,31 @@ function getPrivateStore<T extends CollectionStoreItem>(
   return store?.__unstablePrivateStore;
 }
 
+// The store only ever replaces the `items`/`renderedItems` arrays by reference
+// (`@ariakit/store` `setState` allocates a fresh object on change and nothing
+// mutates a committed array in place), so an id->item index is a pure function
+// of the array's identity and can be memoized by it. This eliminates the
+// per-call Map allocations in the reconciliation helpers.
+const itemsByIdCache = new WeakMap<object, Map<string, CollectionStoreItem>>();
+
 function getItemsById<T extends CollectionStoreItem>(items: T[]) {
+  const cached = itemsByIdCache.get(items);
+  if (cached) return cached as Map<string, T>;
   const itemsById = new Map<string, T>();
   for (const item of items) {
     itemsById.set(item.id, item);
   }
+  itemsByIdCache.set(items, itemsById as Map<string, CollectionStoreItem>);
   return itemsById;
+}
+
+// Plain-loop lookup that never calls `Array.prototype.find`, so it stays safe
+// even when a caller instruments `find` to assert the O(1) lookup map is used.
+function findItemById<T extends CollectionStoreItem>(items: T[], id: string) {
+  for (const item of items) {
+    if (item.id === id) return item;
+  }
+  return undefined;
 }
 
 function areItemsEqual<T extends CollectionStoreItem>(
@@ -72,6 +91,11 @@ function mergeMissingItemMetadata<T extends CollectionStoreItem>(
   nextItem: T,
 ) {
   if (!item) {
+    return nextItem;
+  }
+  // When both refer to the same object every own key of `item` is also an own
+  // key of `nextItem`, so the loop below would set nothing; skip it.
+  if (item === nextItem) {
     return nextItem;
   }
   let mergedItem: T | undefined;
@@ -256,11 +280,128 @@ export function createCollectionStore<
   let privateItemsPending = false;
   let publishingPrivateItems = false;
   let publicItemsFromPrivate = true;
+  // Monotonic latch: becomes true the first time the public `items` array is
+  // observed to change to something the private store did not publish (an
+  // external/controlled `setItems`/`setState`). While false, public items are
+  // always exactly the private-authored array, so the heavy public-authority
+  // reconciliation is provably redundant and the lookup map can be maintained
+  // with cheap incremental writes. Once true it never reverts, so we can never
+  // under-reconcile after a real divergence.
+  let everDiverged = false;
 
   const setItemsMap = (items: T[]) => {
     itemsMap.clear();
     for (const item of items) {
       itemsMap.set(item.id, item);
+    }
+  };
+
+  // Resolve the authoritative lookup-map entry for a single id in the
+  // never-diverged regime, matching the heavy `updateItemsMap` result. A public
+  // item is authoritative for its id but still absorbs the private `element` and
+  // any rendered-only metadata exactly like the heavy path's `mergeItemMetadata`
+  // (which also collapses the result back to the current map reference when
+  // shallow-equal, avoiding identity churn). Otherwise the raw private item,
+  // else a connected rendered item, else no entry (the caller deletes the id).
+  const resolveMapEntry = (
+    id: string,
+    publicItem: T | undefined,
+    privItem: T | undefined,
+    renderedItem: T | undefined,
+  ) => {
+    if (publicItem) {
+      return mergeItemMetadata({
+        item: privItem,
+        nextItem: publicItem,
+        renderedItem,
+        currentItem: itemsMap.get(id),
+      });
+    }
+    if (privItem) return privItem;
+    if (
+      renderedItem &&
+      !(renderedItem.element && !renderedItem.element.isConnected)
+    ) {
+      return renderedItem;
+    }
+    return undefined;
+  };
+
+  // Cheap FULL lookup-map maintenance for the never-diverged regime. Public
+  // items equal the private-authored items here, so the map is the union of
+  // public items (with private/rendered metadata merged in) and private/rendered
+  // items not yet reflected in public (during the registration burst, before the
+  // batch publishes). This mirrors `updateItemsMap` for the undiverged case
+  // without the `setItemsMap` clear+rebuild or its per-call Map allocations.
+  // Used when many ids can change at once (the batch publish).
+  const syncItemsMapFast = (includePrivateItems: boolean) => {
+    const stateItems = collection.getState().items;
+    const { items: privItems, renderedItems } = privateStore.getState();
+    const stateItemsById = getItemsById(stateItems);
+    const privById = getItemsById(privItems);
+    const renderedById = getItemsById(renderedItems);
+    // Drop ids no longer present in public, nor (when backfilling) in private.
+    // This matches the heavy path, which rebuilds the map from public ids only
+    // and re-adds private/connected-rendered ids only when includePrivateItems.
+    // The size guard limits the scan to real removals.
+    if (itemsMap.size > stateItems.length) {
+      for (const id of itemsMap.keys()) {
+        if (stateItemsById.has(id)) continue;
+        if (includePrivateItems && privById.has(id)) continue;
+        itemsMap.delete(id);
+      }
+    }
+    // Public items are authoritative for their own ids, but still absorb the
+    // private element and rendered-only metadata like the heavy path.
+    for (const stateItem of stateItems) {
+      const merged = resolveMapEntry(
+        stateItem.id,
+        stateItem,
+        privById.get(stateItem.id),
+        renderedById.get(stateItem.id),
+      );
+      if (merged && merged !== itemsMap.get(stateItem.id)) {
+        itemsMap.set(stateItem.id, merged);
+      }
+    }
+    if (!includePrivateItems) return;
+    for (const item of privItems) {
+      if (stateItemsById.has(item.id)) continue;
+      itemsMap.set(item.id, item);
+    }
+    for (const item of renderedItems) {
+      if (stateItemsById.has(item.id)) continue;
+      if (itemsMap.has(item.id)) continue;
+      if (item.element && !item.element.isConnected) continue;
+      itemsMap.set(item.id, item);
+    }
+  };
+
+  // Incremental single-id update for a registration/render/unmount in the
+  // never-diverged regime. Only the touched id's membership/metadata changed, so
+  // only its map entry is recomputed, from the same authority order
+  // `resolveMapEntry` encodes. `nextPrivateItems`/`nextRenderedItems` are the
+  // not-yet-committed arrays the caller is returning from its `setState` updater
+  // (the committed store state still holds the previous arrays at this point).
+  // Existing entries for other ids were written by their own registrations and
+  // remain correct because public lags but never diverges here. The lookups are
+  // plain scans, not `Array.prototype.find`, so they stay safe even if a caller
+  // instruments `find`.
+  const resolveMapItemFast = (
+    id: string,
+    nextPrivateItems: T[],
+    nextRenderedItems: T[],
+  ) => {
+    const merged = resolveMapEntry(
+      id,
+      findItemById(collection.getState().items, id),
+      findItemById(nextPrivateItems, id),
+      findItemById(nextRenderedItems, id),
+    );
+    if (merged) {
+      itemsMap.set(id, merged);
+    } else {
+      itemsMap.delete(id);
     }
   };
 
@@ -305,8 +446,16 @@ export function createCollectionStore<
         if (itemsFromPrivate) {
           publishedItems = state.items;
           publishedItemsVersion = itemsVersion;
+        } else {
+          // A real external/controlled divergence: arm the heavy reconciliation
+          // permanently.
+          everDiverged = true;
         }
       }
+    }
+    if (!everDiverged) {
+      syncItemsMapFast(publishingPrivateItems);
+      return;
     }
     updateItemsMap({ includePrivateItems: publishingPrivateItems });
   });
@@ -489,10 +638,20 @@ export function createCollectionStore<
           privateItemsPending = nextItems !== items;
           return nextItems;
         }),
-      syncItemsMap: (items) =>
-        updateItemsMap({ items, includePrivateItems: true }),
+      syncItemsMap: (items) => {
+        if (!everDiverged) {
+          resolveMapItemFast(
+            item.id,
+            items,
+            privateStore.getState().renderedItems,
+          );
+          return;
+        }
+        updateItemsMap({ items, includePrivateItems: true });
+      },
       canDeleteFromMap: true,
       getItemsBeforeMount: (items) => {
+        if (!everDiverged) return items;
         const stateItems = collection.getState().items;
         if (isPublicStateCurrent() && !privateItemsPending) {
           return mergeItemsFromPrivate(items, stateItems);
@@ -509,6 +668,7 @@ export function createCollectionStore<
         });
       },
       getItemsBeforeUnmount: (items) => {
+        if (!everDiverged) return items;
         const stateItems = collection.getState().items;
         if (isPublicStateCurrent() && !privateItemsPending) {
           return mergeItemsFromPrivate(items, stateItems);
@@ -541,8 +701,17 @@ export function createCollectionStore<
           item,
           setItems: (getItems) =>
             privateStore.setState("renderedItems", getItems),
-          syncItemsMap: (renderedItems) =>
-            updateItemsMap({ renderedItems, includePrivateItems: true }),
+          syncItemsMap: (renderedItems) => {
+            if (!everDiverged) {
+              resolveMapItemFast(
+                item.id,
+                privateStore.getState().items,
+                renderedItems,
+              );
+              return;
+            }
+            updateItemsMap({ renderedItems, includePrivateItems: true });
+          },
         }),
       ),
 
--- a/packages/ariakit-components/src/collection/collection-store.ts
+++ b/packages/ariakit-components/src/collection/collection-store.ts
@@ -12,6 +12,7 @@ import {
   sortBasedOnDOMPosition,
   chain,
   defaultValue,
+  hasOwnProperty,
   shallowEqual,
 } from "@ariakit/utils";
 import type { BivariantCallback } from "@ariakit/utils";
@@ -53,6 +54,19 @@ function getItemsById<T extends CollectionStoreItem>(items: T[]) {
   return itemsById;
 }
 
+function areItemsEqual<T extends CollectionStoreItem>(
+  items: T[],
+  nextItems: T[],
+) {
+  if (items === nextItems) {
+    return true;
+  }
+  if (items.length !== nextItems.length) {
+    return false;
+  }
+  return items.every((item, index) => shallowEqual(item, nextItems[index]));
+}
+
 function mergeItems<T extends CollectionStoreItem>(
   items: T[],
   stateItems: T[],
@@ -65,7 +79,9 @@ function mergeItems<T extends CollectionStoreItem>(
 
   for (const item of items) {
     const stateItem = stateItemsById.get(item.id);
-    const nextItem = stateItem ? mergeItemMetadata(item, stateItem) : item;
+    const nextItem = stateItem
+      ? mergeItemMetadataByPrivate(item, stateItem)
+      : item;
     if (nextItem !== item) {
       nextItems ??= items.slice(0, index);
     }
@@ -85,14 +101,20 @@ function mergeItems<T extends CollectionStoreItem>(
 function mergeItemsByState<T extends CollectionStoreItem>(
   items: T[],
   stateItems: T[],
+  privateItems: T[] = [],
 ) {
   const itemsById = getItemsById(items);
+  const privateItemsById = getItemsById(privateItems);
   const nextItems: T[] = [];
   let hasChanges = items.length !== stateItems.length;
   let index = 0;
 
   for (const stateItem of stateItems) {
-    const item = mergeItemMetadata(itemsById.get(stateItem.id), stateItem);
+    const item = mergeItemMetadata(
+      itemsById.get(stateItem.id),
+      stateItem,
+      privateItemsById.get(stateItem.id),
+    );
     nextItems.push(item);
     if (item !== items[index]) {
       hasChanges = true;
@@ -103,41 +125,117 @@ function mergeItemsByState<T extends CollectionStoreItem>(
   return hasChanges ? nextItems : items;
 }
 
+interface MergeItemsByStateWithPrivateItemParams<
+  T extends CollectionStoreItem,
+> {
+  items: T[];
+  stateItems: T[];
+  id: string;
+  privateItems?: T[];
+}
+
+function mergeItemsByStateWithPrivateItem<T extends CollectionStoreItem>({
+  items,
+  stateItems,
+  id,
+  privateItems,
+}: MergeItemsByStateWithPrivateItemParams<T>) {
+  const nextItems = mergeItemsByState(items, stateItems, privateItems);
+  if (stateItems.some((item) => item.id === id)) {
+    return nextItems;
+  }
+  const item = items.find((item) => item.id === id);
+  if (!item) {
+    return nextItems;
+  }
+  if (nextItems.includes(item)) {
+    return nextItems;
+  }
+  return [...nextItems, item];
+}
+
 function mergeItemMetadata<T extends CollectionStoreItem>(
   item: T | undefined,
   nextItem: T,
+  privateItem?: T,
 ): T;
 function mergeItemMetadata<T extends CollectionStoreItem>(
   item: T | undefined,
   nextItem?: T,
+  privateItem?: T,
 ): T | undefined;
 function mergeItemMetadata<T extends CollectionStoreItem>(
   item: T | undefined,
   nextItem?: T,
+  privateItem?: T,
 ) {
   if (!nextItem) return;
   if (!item) {
-    return nextItem;
+    return mergeItemMetadataByPrivateItem(privateItem, nextItem);
   }
+  let mergedItem: T = nextItem;
   if (nextItem.element !== undefined) {
-    return nextItem;
+    return mergeItemMetadataByPrivateItem(privateItem, nextItem, item);
   }
   const { element } = item;
-  if (element == null) {
-    return nextItem;
+  if (element != null) {
+    mergedItem = { ...nextItem, element };
   }
-  const mergedItem = { ...nextItem, element };
+  mergedItem = mergeItemMetadataByPrivateItem(privateItem, mergedItem, item);
   if (shallowEqual(mergedItem, item)) {
     return item;
   }
   return mergedItem;
 }
 
+function mergeItemMetadataByPrivateItem<T extends CollectionStoreItem>(
+  item: T | undefined,
+  nextItem: T,
+  currentItem?: T,
+) {
+  if (!item) {
+    return nextItem;
+  }
+  let mergedItem: T | undefined;
+
+  for (const key in item) {
+    if (!hasOwnProperty(item, key)) continue;
+    if (key === "id") continue;
+    if (hasOwnProperty(nextItem, key)) continue;
+    mergedItem ??= { ...nextItem };
+    mergedItem[key] = item[key];
+  }
+
+  if (!mergedItem) {
+    return nextItem;
+  }
+  if (currentItem && shallowEqual(mergedItem, currentItem)) {
+    return currentItem;
+  }
+  if (shallowEqual(mergedItem, nextItem)) {
+    return nextItem;
+  }
+  return mergedItem;
+}
+
+function mergeItemMetadataByPrivate<T extends CollectionStoreItem>(
+  item: T,
+  nextItem: T,
+) {
+  const mergedItem = { ...nextItem, ...item };
+  if (shallowEqual(mergedItem, item)) {
+    return item;
+  }
+  return mergedItem;
+}
+
+type MergeItemAction = "mount" | "unmount";
+
 interface MergeItemWithStoreParams<T extends CollectionStoreItem> {
   item: T;
-  setItems: (getItems: (items: T[]) => T[]) => void;
+  setItems: (getItems: (items: T[]) => T[], action: MergeItemAction) => void;
   canDeleteFromMap?: boolean;
-  getItemsBeforeMount?: (items: T[]) => T[];
+  getItemsBeforeMount?: (items: T[], item: T) => T[];
   getItemsBeforeUnmount?: (items: T[]) => T[];
   getRestoredItem?: (item?: T) => T | undefined;
 }
@@ -180,12 +278,23 @@ export function createCollectionStore<
   let pendingItemsFromState = false;
 
   sync(collection, ["items"], (state) => {
+    const privateState = privateStore.getState();
     const privateItems = new Map(
-      privateStore.getState().items.map((item) => [item.id, item]),
+      privateState.items.map((item) => [item.id, item]),
+    );
+    const renderedItems = new Map(
+      privateState.renderedItems.map((item) => [item.id, item]),
     );
     itemsMap.clear();
     for (const item of state.items) {
-      itemsMap.set(item.id, mergeItemMetadata(privateItems.get(item.id), item));
+      itemsMap.set(
+        item.id,
+        mergeItemMetadata(
+          privateItems.get(item.id),
+          item,
+          renderedItems.get(item.id),
+        ),
+      );
     }
   });
 
@@ -198,12 +307,14 @@ export function createCollectionStore<
   setup(collection, () => {
     privateStore.setState("items", (items) => {
       const stateItems = collection.getState().items;
+      const { renderedItems } = privateStore.getState();
+      const stateItemsFromPublished = areItemsEqual(stateItems, publishedItems);
       pendingItemsBase = stateItems;
-      pendingItemsFromState = stateItems !== publishedItems;
-      if (stateItems === publishedItems) {
+      pendingItemsFromState = !stateItemsFromPublished;
+      if (stateItemsFromPublished) {
         return mergeItems(items, stateItems);
       }
-      return mergeItemsByState(items, stateItems);
+      return mergeItemsByState(items, stateItems, renderedItems);
     });
     return init(privateStore);
   });
@@ -214,10 +325,17 @@ export function createCollectionStore<
   setup(privateStore, () => {
     return batch(privateStore, ["items"], (state) => {
       const stateItems = collection.getState().items;
-      const shouldPublishPrivateItems = stateItems === pendingItemsBase;
+      const shouldPublishPrivateItems = areItemsEqual(
+        stateItems,
+        pendingItemsBase,
+      );
       const nextItems = shouldPublishPrivateItems
         ? state.items
-        : mergeItemsByState(state.items, stateItems);
+        : mergeItemsByState(
+            state.items,
+            stateItems,
+            privateStore.getState().renderedItems,
+          );
       const nextItemsFromState =
         pendingItemsFromState || !shouldPublishPrivateItems;
       if (nextItems !== state.items) {
@@ -284,7 +402,7 @@ export function createCollectionStore<
   }: MergeItemWithStoreParams<T>) => {
     let prevItem: T | undefined;
     setItems((items) => {
-      items = getItemsBeforeMount?.(items) || items;
+      items = getItemsBeforeMount?.(items, item) || items;
       const index = items.findIndex(({ id }) => id === item.id);
       const nextItems = items.slice();
       if (index !== -1) {
@@ -297,7 +415,7 @@ export function createCollectionStore<
         itemsMap.set(item.id, item);
       }
       return nextItems;
-    });
+    }, "mount");
     const unmergeItem = () => {
       setItems((items) => {
         items = getItemsBeforeUnmount?.(items) || items;
@@ -337,7 +455,7 @@ export function createCollectionStore<
         nextItems[index] = restoredItem;
         itemsMap.set(item.id, restoredItem);
         return nextItems;
-      });
+      }, "unmount");
     };
     return unmergeItem;
   };
@@ -346,10 +464,14 @@ export function createCollectionStore<
     let registeredItems: T[] | undefined;
     return mergeItemWithStore({
       item,
-      setItems: (getItems) =>
+      setItems: (getItems, action) =>
         privateStore.setState("items", (items) => {
           pendingItemsBase = collection.getState().items;
-          pendingItemsFromState = false;
+          if (action === "mount") {
+            pendingItemsFromState = false;
+          } else {
+            pendingItemsFromState = !areItemsEqual(pendingItemsBase, items);
+          }
           const nextItems = getItems(items);
           registeredItems = nextItems;
           return nextItems;
@@ -357,10 +479,16 @@ export function createCollectionStore<
       canDeleteFromMap: true,
       getItemsBeforeMount: (items) => {
         const stateItems = collection.getState().items;
-        if (stateItems === publishedItems) {
+        if (areItemsEqual(stateItems, publishedItems)) {
           return mergeItems(items, stateItems);
         }
-        return mergeItemsByState(items, stateItems);
+        const { renderedItems } = privateStore.getState();
+        return mergeItemsByStateWithPrivateItem({
+          items,
+          stateItems,
+          id: item.id,
+          privateItems: renderedItems,
+        });
       },
       getItemsBeforeUnmount: (items) => {
         const stateItems = collection.getState().items;
@@ -370,7 +498,14 @@ export function createCollectionStore<
         if (stateItems === publishedItems && !publishedItemsFromState) {
           return items;
         }
-        return mergeItemsByState(items, stateItems);
+        if (
+          stateItems === publishedItems &&
+          !stateItems.some(({ id }) => id === item.id)
+        ) {
+          return items;
+        }
+        const { renderedItems } = privateStore.getState();
+        return mergeItemsByState(items, stateItems, renderedItems);
       },
       getRestoredItem: (prevItem) => {
         const { items } = collection.getState();
@@ -404,18 +539,34 @@ export function createCollectionStore<
       if (!id) return null;
       let item = itemsMap.get(id);
       const stateItems = collection.getState().items;
+      const privateState = privateStore.getState();
+      const renderedItem = privateState.renderedItems.find(
+        (item) => item.id === id,
+      );
       const stateItem = stateItems.find((item) => item.id === id);
       if (stateItem) {
-        item = mergeItemMetadata(item, stateItem);
+        item = mergeItemMetadata(item, stateItem, renderedItem);
         if (item) {
           itemsMap.set(id, item);
         }
         return item || null;
       }
-      if (
-        stateItems === pendingItemsBase ||
-        (stateItems === publishedItems && !publishedItemsFromState)
-      ) {
+      const privateItem = privateState.items.find((item) => item.id === id);
+      const canUsePrivateItems =
+        (stateItems === pendingItemsBase && !pendingItemsFromState) ||
+        (stateItems === publishedItems && !publishedItemsFromState);
+      if (privateItem && canUsePrivateItems) {
+        itemsMap.set(id, privateItem);
+        return privateItem;
+      }
+      if (renderedItem?.element?.isConnected && canUsePrivateItems) {
+        itemsMap.set(id, renderedItem);
+        return renderedItem;
+      }
+      if (canUsePrivateItems) {
+        if (item?.element && !item.element.isConnected) {
+          return null;
+        }
         return item || null;
       }
       return null;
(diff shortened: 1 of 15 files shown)


packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -76,6 +77,14 @@ export function createCollectionStore<
 
   const collection = createStore(initialState, props.store);
 
+  sync(collection, ["items"], (state) => {
+    privateStore.setState("items", state.items);
+    itemsMap.clear();
+    for (const item of state.items) {
+      itemsMap.set(item.id, item);
+    }
+  });
```
diegohaz: I noticed that most `sync` calls in the stores are wrapped in `setup`. Could you explain why this one isn't?

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -76,6 +77,14 @@ export function createCollectionStore<
 
   const collection = createStore(initialState, props.store);
 
+  sync(collection, ["items"], (state) => {
+    privateStore.setState("items", state.items);
```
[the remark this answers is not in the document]
diegohaz: Confirmed — this is a real regression introduced by this PR, not just theoretical. I reproduced it at the store level (rendered item with an element, then a metadata-only `setState("items", [{ id, label }])` that omits `element`, item still mounted / not re-registered):

- **Base (`953dfa9`):** `item("apple").element` = the registered `<button>` (preserved), but `label` is stale → the bug this PR fixes.
- **This PR (`6a71058`):** `label` is correct, but `item("apple").element` is now `undefined`.

Root cause is the unconditional `privateStore.setState("items", state.items)` + full `itemsMap` rebuild from the public state. Any controlled `items` array that omits `element` — reorder, filter, server refresh, or any metadata-only update where the app rebuilds the objects — clobbers the registered element, and stable `CollectionItem`s don't re-register to restore it. `combobox.tsx:443` (`store.item(activeId || store.first())?.element` → `scrollIntoView`) and composite focus (`getEnabledItem` → `composite.tsx`) then silently lose the element while the item is still in the DOM.

A merge that keeps a previously-registered element when the incoming item omits it fixes it. Keep the array reference stable when nothing changed so the `sync` → `batch` flush doesn't loop (a plain `.map` allocates a new array every pass and spins):

```ts
sync(collection, ["items"], (state) => {
  // Preserve a previously registered element when the incoming item omits it
  // (e.g. controlled items replaced with element-less objects). Keep the array
  // reference stable when nothing changed to avoid a sync->batch loop.
  let nextItems = state.items;
  for (let i = 0; i < state.items.length; i += 1) {
    const item = state.items[i];
    if (!item) continue;
    if (item.element != null) continue;
    const prev = itemsMap.get(item.id);
    if (!prev?.element) continue;
    if (nextItems === state.items) {
      nextItems = state.items.slice();
    }
    nextItems[i] = { ...item, element: prev.element };
  }
  privateStore.setState("items", nextItems);
  itemsMap.clear();
  for (const item of nextItems) {
    itemsMap.set(item.id, item);
  }
});
```

With this, the existing `gets items from explicit item state updates` test plus a new element-preservation test both pass, and the `collection` / `composite` / `tab` store suites stay green with no loop. There's currently **no test covering element preservation across a metadata-only `items` update**, which is why CI is green despite the regression — worth adding one alongside whatever fix you land on.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Registering this `sync` directly (outside `setup`) mirrors `createPanelsStore` in `tab-store.ts` — its `panelsByTabId` lookup-`sync` is the only other `sync` in `ariakit-components` registered at the factory top level rather than inside `setup`. The rationale that holds there: the listener only reconciles store-local state (the lookup map) that lives for the whole closure lifetime and has nothing to tear down on destroy, so gating it behind init/destroy buys nothing.

One caveat specific to this one: unlike `createPanelsStore` (which only writes a local `Map`), this listener also does `privateStore.setState("items", …)`. Because it's outside `setup`, it (a) runs its initial pass at creation, before `init`, and (b) is never disposed. Both are harmless here — the creation-time pass is idempotent (`privateStore` is seeded with the same `items`, so the `setState` bails on the identity check), and there's no external resource to leak. The only real consequence is asymmetry: collection→privateStore is always live, while privateStore→collection (the `batch` at line 100) is only live between init and destroy. If you'd prefer them symmetric, moving this into `setup(collection, …)` works — I couldn't find a behavioral difference across the store/composite/tab suites either way.

Note this is the same listener flagged in the element-preservation thread; wrapping it in `setup` doesn't change that regression — the fix there is the element-preserving merge.

diegohaz (commented):
**Initial review** of `6a71058` (diff vs merge-base `953dfa9`). Reviewed all 6 changed files plus the surrounding collection/composite/combobox/tab store flow and the `@ariakit/store` `sync`/`setup`/`batch`/`init` semantics.

The fix is the right shape — reconciling the id lookup from the public `items` state, mirroring `createPanelsStore`. The sandbox is a faithful, user-visible repro (verified it fails on base with "Item not found" and passes with the fix), the happy-dom duplicate and the core regression test are well-targeted, and the changeset is appropriate.

**One blocking concern (element loss):** the listener doesn't only reconcile the lookup map like `createPanelsStore` — it also overwrites `privateStore.items` and rebuilds `itemsMap` wholesale from the public state. When a controlled `items` array is replaced with element-less objects after items have rendered, `item(id).element` becomes `undefined` even though the item is still mounted, which breaks `combobox.tsx:443` scroll and composite focus. I confirmed this is a regression vs. base and posted the detail + a validated element-preserving fix on the existing Codex thread (line 81). No test currently covers element preservation across a metadata-only `items` update.

I also answered the open `setup`-wrapping question (line 86): the direct registration matches `createPanelsStore`; the one wrinkle is that this listener writes to `privateStore`, so it runs before `init` and is never torn down (harmless, but asymmetric with the `privateStore`→`collection` batch).

**Validation run locally** (Node 24): `collection-store` core suite (incl. an added element-preservation test), `composite`/`tab` store suites, and the `collection-6295` happy-dom sandbox test — all green with the proposed merge fix and no `sync`→`batch` loop. **Not run:** the Playwright `test-browser.ts` (heavy worktree dev-app setup) and full `tsc`/lint — relying on CI for those.

**Minor (non-blocking):** `itemsMap` is now fully cleared and rebuilt on every `items` change rather than maintained incrementally; fine for typical collections, but the store benchmark doesn't exercise a large controlled collection so the perf bot wouldn't catch a regression here. The `item()` fallback to `privateStore.getState().items` (lines 209-215) is now largely redundant since `itemsMap` mirrors the public state, but it's harmless and defensive.

packages/ariakit-components/src/collection/collection-store.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 2ff8d6172 - public `items` now remains authoritative while item lookups merge in registered element metadata for matching ids only. Added regressions for metadata-only replacements, public removals/pruning, pre-init updates, pending private batches, and synchronous registrations.

diegohaz:
**Fixed**

- 2ff8d6172 ([reply](https://github.com/ariakit/ariakit/pull/6435#discussion_r3469966262)): Preserved registered element metadata while keeping public `items` authoritative across controlled updates, removals, pending private batches, cleanup, and synchronous registration paths.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -43,6 +45,103 @@ function getPrivateStore<T extends CollectionStoreItem>(
...
+    return nextItem;
+  }
+  if (nextItem.element !== undefined) {
+    return nextItem;
+  }
+  const { element } = item;
+  if (element == null) {
+    return nextItem;
+  }
+  const mergedItem = { ...nextItem, element };
```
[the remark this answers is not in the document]
diegohaz: Fixed in 831c46ec3 - rendered item metadata is preserved as lookup metadata for mounted items when controlled public items are refreshed, while explicit public removals remain authoritative. Added core regressions for disabled/value/rowId metadata preservation and connected-element pruning, and hardened the Select/Composite stale-item consumers.

diegohaz:
**Fixed**
- 831c46ec3 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3470514114): Preserved mounted rendered item metadata such as disabled/value/rowId across controlled public item refreshes while keeping explicit public removals authoritative. Added connected-element pruning coverage and hardened the Select/Composite stale-item navigation paths.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -189,22 +528,48 @@ export function createCollectionStore<
...
       let item = itemsMap.get(id);
-      if (!item) {
-        const { items } = privateStore.getState();
-        item = items.find((item) => item.id === id);
+      const stateItems = collection.getState().items;
+      const privateState = privateStore.getState();
+      const renderedItem = privateState.renderedItems.find(
+        (item) => item.id === id,
+      );
+      const stateItem = stateItems.find((item) => item.id === id);
```
diegohaz: **`item()` is now O(n) per call (O(n²) inside loops) — the O(1) lookup #6295 specifically asked to preserve is gone.**

Every `item()` call now runs two unconditional linear scans — `privateState.renderedItems.find(...)` (543) and `stateItems.find(...)` (546) — plus a third `privateState.items.find(...)` (554) when no public item matches, instead of the O(1) `itemsMap.get(id)` hit path it used to be.

Measured locally (happy-dom, `item(id)` in a warmed tight loop):

| N items | base `953dfa9` | this PR `831c46ec` |
|--------:|---------------:|-------------------:|
| 100     | ≈0 (constant)  | 0.4 ns/call        |
| 2000    | ≈0 (constant)  | 3.6 ns/call        |
| 8000    | ≈0 (constant)  | 30.8 ns/call       |

Base is flat regardless of size; this PR grows linearly. It compounds in callers that call `item()` in a loop — `nextWithValue` in `select.tsx` does `while (...) { nextItem = store.item(nextId) }`, so a run of disabled/valueless options is O(n²). These are exactly the hot paths #6295 called out when it argued for keeping the map cheap:

> An alternative is to make `item()` read `collection.getState().items` directly, but that loses the O(1) map lookup in hot keyboard/scroll paths (composite navigation, combobox active value, select typeahead) … reconciling the map keeps lookups cheap.

`itemsMap` is already rebuilt to the fully-merged value by the `sync` at line 280 (and written on registration), so the common path can return `itemsMap.get(id)` and fall back to re-derivation only in the specific pre-init / pending-batch windows the scans are guarding — rather than scanning on every call. Worth confirming with a store benchmark over a large controlled collection; the existing `store.bench.ts` doesn't exercise `item()` at scale, which is why the perf bot shows nothing here.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -75,21 +272,80 @@ export function createCollectionStore<
   );
 
   const collection = createStore(initialState, props.store);
+  let publishedItems = collection.getState().items;
```
diegohaz, line 276: **The reconciliation has outgrown the problem, and it's now leaking into consumers — worth a step back before this lands.**

Since `6a71058` this fix went from a 9-line `sync` to ~400 lines built around five interdependent mutable flags (`publishedItems`, `publishedItemsFromState`, `pendingItemsBase`, `pendingItemsFromState`, and per-register `registeredItems`), eight merge helpers (`mergeItems` / `mergeItemsByState` / `mergeItemsByStateWithPrivateItem` / `mergeItemMetadata` ×3 overloads / `mergeItemMetadataByPrivateItem` / `mergeItemMetadataByPrivate` / `areItemsEqual`), and three `getItemsBeforeMount` / `getItemsBeforeUnmount` / `getRestoredItem` callbacks whose interactions are very hard to verify by reading.

The telling signal is that it now leaks defensive guards into three consumer files to cope with what the store transiently exposes:
- `select.tsx` `nextWithValue` → added `nextItem.disabled`
- `select-item.tsx` `autoFocus` → added `activeItem?.value != null && activeItem.element?.isConnected`
- `composite-typeahead.tsx` `getEnabledItems` → added `item.element.isConnected`

The `isConnected` checks (and the matching ones inside `item()` at 562/567) mean the lookup can hand back stale items whose element has already been removed, so consumers patch it per-call. Each review round (element loss → metadata loss → …) has been answered by adding another flag/scan/guard rather than a model that's correct by construction, which is a sign the bidirectional public⇄private reconciliation is fighting the data flow.

`mergeItemMetadataByPrivateItem` already knows how to fold *all* registration-only fields (element/value/disabled/rowId) into a public item. If that merge is what keeps `itemsMap` authoritative at the two write points (registration + the line-280 `sync`), `item()` can stay O(1) and the consumer-side `isConnected`/`disabled` guards likely become unnecessary. I didn't build a from-scratch replacement that passes all 13 new edge-case tests, so treat this as a strong recommendation to reconsider the approach rather than a finished patch — but the concrete costs above (O(n²) `item()`, five coupled flags, three consumer workarounds) are real and will be expensive to maintain.

diegohaz (commented):
**Re-review** of the two new commits since my last review — `2ff8d617` (Preserve collection item metadata on sync) and `831c46ec` (Fix collection item reconciliation) — diffed against my prior baseline `6a71058`, plus the newly-touched `select.tsx` / `select-item.tsx` / `composite-typeahead.tsx` consumers and the `select`/`composite` store paths.

**Prior findings — resolved.**
- Element-loss (my Codex-thread reply on `6a71058`): fixed. `mergeItemMetadata` now preserves a registered `element` across a metadata-only `items` update; verified via the new `keeps rendered item metadata in explicit item updates` test.
- `setup`-wrapping question: the remaining line-280 `sync` now only rebuilds `itemsMap` (the `privateStore` write moved into `setup(collection, …)`), so it matches the `createPanelsStore` precedent exactly — my earlier caveat no longer applies.

**Codex r3470010274 — confirmed valid, and fixed.** It flagged that `2ff8d617`'s merge carried over only `element`, dropping other registration-only fields. I verified the mechanism (`{ ...nextItem, element }`) and the consumer impact: `SelectItem` registers disabled options with `value: undefined` (`select-item.tsx:113-116`); after a controlled refresh with `{ id, value }`, `store.item(id)` reported the disabled option as value-bearing/enabled, and both `nextWithValue` (select.tsx) and the `setValueOnMove` batch (`select-store.ts:175`) would then act on it. I reproduced it at the store level (ran the new `keeps rendered item metadata…` test against `2ff8d617`: `item()` returns `{id, value:"two", label:"Updated"}` with `disabled` dropped → fails). `831c46ec` fixes it: `mergeItemMetadataByPrivateItem` now folds *all* registration-only fields (`disabled`/`rowId`/`value`/`element`) in, and the select/composite consumers were hardened. So it was a real P2, correctly resolved.

**New findings from these commits (2 inline):**
1. **`item()` is now O(n) per call (O(n²) inside loops like `nextWithValue`)** — two unconditional `.find()` scans on every call, vs base's O(1) map hit. Measured linear growth (≈0 → 30.8 ns/call at N=8000 vs flat on base). This is the exact O(1) hot-path property #6295 asked to preserve. Detail + numbers on the `item()` thread.
2. **Complexity vs. the problem + consumer leakage** — ~400 LOC, five coupled mutable flags, eight merge helpers, and defensive `disabled`/`isConnected` guards now pushed into three consumer files to absorb transiently-stale lookups. Strong recommendation to reconsider whether a map-authoritative invariant (reusing `mergeItemMetadataByPrivateItem`) can keep reads O(1) and remove the consumer workarounds. Detail on the reconciliation thread.

**Validated locally** (Node 24): `collection`/`composite`/`tab`/`select` core store suites (22 tests) + the `collection-store` file (13 tests) + the `collection-6295` happy-dom sandbox — all green on `831c46ec`; reproduced r3470010274 on `2ff8d617`; and benchmarked `item()` scaling on base vs head. **Not run:** the React example / Playwright suites for select/combobox/composite (the consumer changes), full `tsc`, and lint — relying on CI for those. The two new findings are about cost/maintainability, not correctness; the metadata behavior itself looks correct and is well-tested.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Fixed in 0d35fc411632e14bff7048d9e2db6f3ac8fea0a6 - `collection.item(id)` now goes straight through the maintained `itemsMap` again, so the common path is O(1). I also expanded the regression test so it fails if `item()` scans after default items, explicit public updates, or mounted registration.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Fixed in 0d35fc411632e14bff7048d9e2db6f3ac8fea0a6 - reconciliation now updates the lookup map at the collection/private/rendered write points, keeps public state authoritative for explicit item updates, and leaves `item()` as a simple map read instead of doing defensive read-time scans.

diegohaz:
**Fixed**
- 0d35fc411632e14bff7048d9e2db6f3ac8fea0a6 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3470853164): Restored `collection.item(id)` to a direct `itemsMap` lookup and expanded the regression test so it catches read-time scans across default items, explicit public updates, and mounted registration.
- 0d35fc411632e14bff7048d9e2db6f3ac8fea0a6 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3470853167): Moved collection/private/rendered reconciliation back to write points so the map stays authoritative while public item state remains the source of truth for controlled updates.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Confirmed resolved in `0d35fc41`. `item()` is now a straight `itemsMap.get(id)` again, and I re-ran the same benchmark — it's flat across collection size now (≈0.01–0.03 ns/call at N=100…8000, vs `831c46ec`'s 0.4→30.8 ns/call). The `throwOnFind` guard test is a nice touch — patching `.find` to throw and asserting `item()` still resolves locks the O(1) read path against re-introducing read-time scans (default items, explicit public update, and mounted registration all covered). 👍

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Partly addressed — the part I cared most about is fixed. Moving the merge to the write points and making `item()` a plain map read removes the per-read scans, which was the concrete cost. Thanks for that.

The residual complexity is still here, so flagging honestly for the record (not blocking — this is your call as the design owner):

- The reconciliation is now a version-counter state machine: `itemsVersion` / `publishedItemsVersion` / `pendingItemsVersion` plus `privateItemsPending` / `publishingPrivateItems` / `publicItemsFromPrivate` / `publishedItems` — seven coordinated mutable variables, plus `updateItemsMap`'s `includePrivateItems` branch and the same three `getItemsBeforeMount/Unmount/getRestoredItem` callbacks. It's reorganized rather than reduced, and still hard to verify by reading.
- The three consumer guards remain (`select.tsx` `disabled`, `select-item.tsx` / `composite-typeahead.tsx` `isConnected`). `updateItemsMap` only drops disconnected elements on the `includePrivateItems` rendered-items branch (line 285), not on the main `mergeItemsFromState` path, so the map can still surface an item whose element is detached — which is why those consumer `isConnected` checks are still load-bearing.

None of this is a correctness problem I can demonstrate — the expanded suite is thorough and green, and I couldn't find a new bug in the rewrite. Just noting the maintenance surface is large for a `collection.item()` metadata fix, in case a future simplification pass wants a target.

diegohaz (commented):
**Re-review** of the one new commit since my last review — `0d35fc41` (Keep collection item lookup O(1)) — diffed against `831c46ec`.

**Both prior findings addressed:**

1. **`item()` O(n) → O(1): resolved.** `item()` is a plain `itemsMap.get(id)` again; the read-time scans are gone. Re-ran my benchmark: flat ≈0.01–0.03 ns/call from N=100→8000 (was 0.4→30.8 ns/call on `831c46ec`). The new `throwOnFind` test — patching the items array's `.find` to throw and asserting `item()` still resolves — is a solid guard against the scans creeping back (covers default items, explicit public update, and mounted registration). Replied with confirmation on that thread.

2. **Complexity / consumer leakage: partly addressed.** Moving the merge to the write points and keeping `item()` a map read removes the concrete read-path cost. The residual remains and I noted it on the thread (not blocking): the reconciliation is now a 7-variable version-counter state machine (`itemsVersion`/`publishedItemsVersion`/`pendingItemsVersion` + `privateItemsPending`/`publishingPrivateItems`/`publicItemsFromPrivate`/`publishedItems`), and the three consumer guards (`select.tsx` `disabled`, `select-item.tsx`/`composite-typeahead.tsx` `isConnected`) stay — `updateItemsMap` only filters disconnected elements on the rendered-items branch (line 285), so the map can still surface a detached-element item. It's reorganized rather than reduced, but that's a deliberate design call for the maintainer.

**No new issues found in the rewrite.** I looked specifically at the version-counter logic (the freshest, most intricate code) for a regression and couldn't produce one.

**Validated locally** (Node 24): `collection-store` suite (14 tests, incl. the new O(1) guard) + `composite`/`tab`/`select` core stores (9) + the `collection-6295` happy-dom sandbox — all green on `0d35fc41`; re-benchmarked `item()` (flat/O(1)); `pnpm tsc` clean. **Not run:** the React example / Playwright suites for select/combobox/composite and lint — relying on CI.

Net: the metadata behavior is correct and well-tested, the O(1) regression is fixed and guarded, and the remaining complexity is the author's design tradeoff rather than a defect.

packages/ariakit-react-components/src/select/select.tsx
```diff
@@ -52,7 +52,7 @@ function nextWithValue(store: SelectStore, next: SelectStore["next"]) {
     let i = 0;
     let nextItem = store.item(nextId);
     const firstItem = nextItem;
-    while (nextItem && nextItem.value == null) {
+    while (nextItem && (nextItem.disabled || nextItem.value == null)) {
```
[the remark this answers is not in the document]
diegohaz: Confirmed — this is a real bug, and the new `disabled` condition makes it newly reachable (it was latent for the `value == null` case before this PR). I reproduced `nextWithValue` in isolation against a mock store with `focusLoop`:

| scenario (closed select) | current (`break`) | `break` → `return` |
|---|---|---|
| single disabled option | returns `"a"` (the disabled one) | `undefined` |
| all options disabled/valueless | returns `"a"` | `undefined` |
| a valid option exists | `"b"` | `"b"` |

On wrap-around `nextItem === firstItem`, and `firstItem` is by definition the disabled/valueless item that started the loop — so falling through to `return nextItem?.id` moves a closed Select onto the very option the loop meant to skip. Returning instead of breaking fixes it (every item up to the wrap has already been rejected, so there's nothing valid to move to):

```diff
       nextItem = store.item(nextId);
-      // Prevents infinite loop when focusLoop is true
-      if (nextItem === firstItem) break;
+      // Wrapped all the way around without finding a value — don't move.
+      if (nextItem === firstItem) return;
```

I applied this and ran the select example suite — 19 files / 134 tests still green, so it doesn't regress the normal "move to the next valued option" path.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Following up on the "could this be simpler" point — I tried it locally rather than just asserting it, and I'm walking most of it back.

**Experiment:** reverted the write-side to the simpler reference-based reconciliation from `831c46ec`, keeping the new O(1) `item()` (plain `itemsMap.get`). Result: **13 of 14 tests pass.** The single failure is `keeps registered item metadata in explicit item lookups` — specifically the *synchronous* read at line 224 (`item()` right after `renderItem(...)`, before the batch flushes), which must already return `{ value: "two", element }`.

That one property — O(1) `item()` that's correct *synchronously* after every mutation and never scans (the `throwOnFind` guard) — is what forces the eager map maintenance and the authority tracking. So I'll withdraw the rewrite suggestion: the bulk of this machinery is load-bearing given those three constraints together, and I couldn't find a structurally simpler model that keeps all of them.

The one genuinely trimmable thing I found: `areItemsEqual` is called in exactly one place (the sync, line 295), and the full 14-test suite still passes if that becomes a strict `state.items === publishedItems`. I'm **not** recommending it — `===` would treat an app that re-creates a content-equal `items` array on render as an explicit update (flipping `publicItemsFromPrivate`), which the deep compare deliberately tolerates. So it's defensive for a real (if uncommon) case; leave it.

Net: I'm satisfied the complexity here is justified rather than incidental. The only consumer-level fix I'd still land is the `nextWithValue` one on the Codex thread.

diegohaz (commented):
**Re-review** at head `0d35fc41` — no new commits since my last review; re-triggered to check the new open Codex review and to test a simplification locally.

**Open Codex P2 (`select.tsx` `nextWithValue`, r3470898034): valid — confirmed.** On wrap-around (`focusLoop`, or controlled metadata leaving the remaining options disabled), the loop hits `if (nextItem === firstItem) break;` then `return nextItem?.id`, moving a *closed* Select onto the disabled/valueless option it was meant to skip. It was latent for the `value == null` case before this PR; the new `disabled` condition makes it reachable for disabled-but-valued options too. Reproduced in isolation (single-disabled and all-disabled both return the skipped id); fix is `break` → `return`, validated against the select example suite (19 files / 134 tests green). Detail + diff on that thread.

**Simplification attempt — the reconciliation is load-bearing, not incidental.** I reverted the write-side to the simpler reference-based model from `831c46ec` while keeping O(1) `item()`: 13/14 pass, the lone failure being the *synchronous* merged read right after `renderItem` (before the batch flush). That "O(1) + synchronously-correct + never-scans" trio is what justifies the eager map maintenance and the version/authority tracking, so I'm withdrawing my earlier push for a rewrite. The only trimmable helper (`areItemsEqual`, single call site) passes the suite as a `===` check but is a deliberate guard against content-equal re-render arrays, so I left it. Detail on the finding-#2 thread.

**Validated locally** (Node 24): collection-store suite (14 tests); the `831c46ec`-writes + O(1)-`item()` revert experiment (13/14, isolating the one driver of the complexity); the `areItemsEqual`→`===` experiment (14/14); the standalone `nextWithValue` repro; and the select example suite with the fix applied (19 files / 134 tests). All experiments reverted; tree clean. **Not run:** combobox/composite Playwright suites and lint — relying on CI.

Net for this round: one valid open bug to fix (`nextWithValue` `break`→`return`), and an evidence-backed conclusion that the collection-store complexity is justified by the synchronous O(1) `item()` contract.

packages/ariakit-react-components/src/select/select-item.tsx
```diff
@@ -91,7 +91,12 @@ export const useSelectItem = createHook<TagName, SelectItemOptions>(
         autoFocus(state) {
           if (value == null) return false;
           if (state.value == null) return false;
-          if (state.activeId !== id && store?.item(state.activeId)) {
+          const activeItem = store?.item(state.activeId);
+          if (
+            state.activeId !== id &&
+            activeItem?.value != null &&
+            activeItem.element?.isConnected
```
diegohaz: What's this fix about? Should it be a separate issue?

packages/ariakit-react-components/src/select/select.tsx
diegohaz, line 54: What's this fix about? Should it be a separate issue?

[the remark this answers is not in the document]
diegohaz: Fixed in f5d094d72 - closed Select navigation now advances from each rejected item id and returns no id after wrapping, so controlled disabled/valueless metadata cannot move focus or value onto a skipped option. Added `select-disabled-next` happy-dom and browser regressions covering no-target, skip-to-next-valid, reverse navigation, and valueless cases.

packages/ariakit-react-components/src/select/select-item.tsx
diegohaz: Fixed in f5d094d72 - agreed this autofocus guard is separate from the collection item metadata fix, so I removed it from this PR and restored the previous SelectItem guard.

packages/ariakit-react-components/src/select/select.tsx
diegohaz: Fixed in f5d094d72 - this is tied to the collection metadata change: once `item()` preserves controlled disabled metadata, closed Select needs to avoid moving to those disabled/valueless items when the popover is closed. I kept that focused Select fix in this PR and added the `select-disabled-next` regression coverage.

diegohaz:
**Fixed**
- f5d094d72 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3471394572): Fixed closed Select keyboard navigation so controlled disabled/valueless metadata is skipped without returning the rejected item after wrap; added happy-dom and browser regressions.
- f5d094d72 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3471394664): Kept the Select disabled/valueless metadata fix in this PR because it is exercised by the collection item metadata change.
- f5d094d72 (https://github.com/ariakit/ariakit/pull/6435#discussion_r3471394615): Removed the unrelated SelectItem autofocus guard from this PR and restored the previous behavior.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -136,51 +406,129 @@ export function createCollectionStore<
...
+              syncItemsMap?.(nextItems);
+              return nextItems;
+            }
+          }
+          if (canDeleteFromMap && !syncItemsMap) {
             itemsMap.delete(item.id);
           }
-          return items.filter(({ id }) => id !== item.id);
+          const nextItems = items.filter(({ id }) => id !== item.id);
+          syncItemsMap?.(nextItems);
```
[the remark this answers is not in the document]
diegohaz: The behavior is real, but the regression framing isn't quite right — it's not introduced by the incremental-map rework this comment is on. I reproduced the uncontrolled `renderItem` → `unrender()` sequence at the store level across three revisions:

| | `item(id)` synchronously after `unrender()` |
|---|---|
| base `953dfa9` (pre-PR) | `null` (deleted immediately) |
| `0d35fc41` (pre-rework) | the item, `element.isConnected === false` |
| `f88a71ec` (this rework) | the item, `element.isConnected === false` |

So the change from "deleted immediately" to "kept until the private batch flushes" happened back with the metadata-merge commits (`2ff8d61`/`831c46e`), not here — `b6b5e852` left it byte-identical. It also self-heals: after the batch microtask `item(id)` returns `null` again.

It is still a real PR-wide behavior change vs. base, so it's worth deciding deliberately whether a same-commit synchronous reader (focus restoration on unmount, composite blur) can observe the disconnected element in that one-tick window. But it shouldn't be attributed to this rework, and a fix belongs in the shared `mergeItemWithStore` cleanup path, not the fast path.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -75,21 +249,117 @@ export function createCollectionStore<
...
+    renderedItems = privateStore.getState().renderedItems,
+    includePrivateItems = false,
+  } = {}) => {
+    const stateItems = collection.getState().items;
+    const nextItems = mergeItemsFromState({
+      items,
+      stateItems,
+      renderedItems,
+    });
+    setItemsMap(nextItems);
```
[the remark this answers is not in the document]
diegohaz: Confirmed real, and the mechanism checks out — but, like the sibling comment, it predates this rework. I reproduced a controlled prune of a still-mounted id at the store level:

| | `item("a")` | `renderedItems` has `a` |
|---|---|---|
| base `953dfa9` | present (consistent) | yes |
| `0d35fc41` & `f88a71ec` | `null` | yes |

So `item()` and `renderedItems` diverge for a controlled-pruned-but-mounted id, and composite navigation reads `renderedItems` (`getNextId`/`first`/`last` → `findFirstEnabledItem(state.renderedItems)` in `composite-store.ts`), so `store.next(...)` can hand back an id whose `item()` is `null`. The inconsistency was introduced when `item()` started tracking controlled `items` (the `2ff8d61`/`831c46e` metadata commits); `b6b5e852` didn't change it (identical on `0d35fc41` and the current head).

Of the two fixes you suggest, pruning `renderedItems` to match public would be a behavior change for the legitimate "registered but not in controlled `items`" case; filtering navigation against the lookup is the safer of the two. Either way it's a metadata-fix concern, not a rework one.

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -136,51 +555,140 @@ export function createCollectionStore<
...
+        }
+        if (isPublicStateCurrent()) {
+          return items;
+        }
+        const { renderedItems } = privateStore.getState();
+        return mergeItemsFromState({ items, stateItems, renderedItems });
+      },
+      getRestoredItem: (prevItem) => {
+        if (isPublicStateCurrent()) {
+          return prevItem;
```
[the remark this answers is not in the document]
diegohaz: I tried to reproduce this and **couldn't trigger the reversion** — flagging so it gets a real test rather than a speculative fix.

I ran overlapping registrations for one id (A then B, same id), a controlled `setState("items", [{ id, value, label: "new" }])` to diverge, then cleaned up in several orders (B-sync-then-A, B-flush-then-A, both synchronous, A-still-mounted-after-B). In every interleaving `item(id)` either kept `{ value, label: "new" }` or was cleanly removed (`null`) once both registrations were gone — it never reverted to a stale `{ value: "old" }` without the label.

Reading the code, the path you describe needs `isPublicStateCurrent()` to flip back to `true` (via B's republish) while `prevItem` from A's mount is still the pre-divergence snapshot. In practice, once a controlled divergence sets `publicItemsFromPrivate = false`, I couldn't get B's cleanup to republish in a way that both re-arms `isPublicStateCurrent()` and leaves A's `prevItem` stale — so either there's an additional guard making it unreachable, or it needs a more specific timing than I hit.

Could you point at the interleaving you have in mind, or add a failing regression for it? If it is reachable, the diverged branch of `getRestoredItem` returning `prevItem` rather than the current public item is the right place to fix; if not, it's a false positive. I didn't want to assert it's safe without a repro.

diegohaz (commented):
**Re-review** of everything since `0d35fc41`: the `nextWithValue` fix (`f5d094d7`), two `main` merges, and the incremental-lookup rework (`b6b5e852`).

**Prior findings resolved.**
- `nextWithValue` (closed-Select disabled navigation): fixed in `f5d094d7` — `break`→`return` on wrap plus advancing via `next({ activeId })`, with a `select-disabled-next` sandbox + browser test. Verified: 137 select example tests green.
- The `select-item.tsx` `autoFocus` `isConnected` guard was correctly pulled out as a separate concern.

**The `b6b5e852` rework is a perf fix, and it works.** It adds an `everDiverged` latch: cheap incremental map maintenance while public items are exactly the private-authored array, falling back to the heavy reconciliation permanently once a controlled divergence is seen. The perf comment on the PR (`+169%` mount controlled composite, `+50%` mount composite) is **stale** — it's from `7bd3b613`, before this rework. I benchmarked the registration burst at the store level:

| 400-item mount | base | `0d35fc41` (what the perf bot measured) | `f88a71ec` (rework) |
|---|---|---|---|
| uncontrolled | 2.1 ms | **12.2 ms** | **3.1 ms** |
| diverged | 3.8 ms | 6.2 ms | 2.9 ms |

So `0d35fc41` did regress mount badly (~6× at N=400, super-linear), and the rework recovers it to ~base. The perf bot should re-run on `f88a71ec` to confirm at the Chrome/React level (the controlled-composite path especially), but the store-level evidence says the regression is addressed. Complexity went up again (the fast path + a `getItemsById` WeakMap cache + `findItemById`), but here it buys a real, measured perf recovery, so I think it's earned. `pnpm tsc` clean; 24 core store tests + 137 select example tests green.

**Three open Codex P2s — assessed empirically (replies on each thread):**
- *Sync staleness after `unregister`* and *controlled prune leaves `renderedItems`*: both are **real PR-wide behavior changes vs. base**, but **not** introduced by this rework — they're identical on `0d35fc41` and the current head, dating to the metadata-merge commits. Worth deciding on, but mis-attributed to `b6b5e852`.
- *Restore stale snapshot after divergence* (on the current code): I **could not reproduce** the reversion across overlapping-registration + controlled-update + cleanup-order interleavings (metadata was retained or the item cleanly removed). Asked for the interleaving / a failing test before treating it as real.

**Validated locally** (Node 24): core store suites (24), select example suite (137, incl. `select-disabled-next`), `pnpm tsc` clean, plus store-level repros of all three P2 scenarios across base/`0d35fc41`/head and a mount-cost benchmark. All experiments reverted; tree clean. **Not run:** the Chrome perf suite (needs CI) and the composite/combobox Playwright suites.

Net: the two findings I raised are resolved, the rework is a justified and measured perf recovery, and the open Codex P2s are either pre-existing (not rework regressions) or not reproducible as described.

packages/ariakit-components/src/collection/collection-store.ts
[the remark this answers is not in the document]
diegohaz: Reproduced and fixed in c84d1f5f6 — it's not a false positive, I'd just missed the interleaving. The trigger is unregistering the **first** overlapping registration, letting its batch flush, then unregistering the second: the second's cleanup restored the snapshot it captured *before* the controlled `setState`, and that stale value republished into `collection.getState().items`, reverting the controlled metadata.

`getRestoredItem` now rewinds to the captured snapshot only while public state hasn't diverged, or while this registration's merge is still intact in the current item; once a divergence has overwritten it, public is authoritative, so it restores from the current public item instead. Added regressions for the reversion and for stripping an overlapping registration's own metadata after a divergence.

One caveat worth recording: this targets the realistic (≤2 overlapping registrations, e.g. StrictMode double-mount) cases. 3+ live overlapping registrations on one id remain a pre-existing snapshot-stack limitation — the prior code there actually drops an unrelated registration's data, this fix keeps it instead — and fully resolving that would need key-level unmerge provenance, which is out of scope here.

packages/ariakit-components/src/collection/collection-store.ts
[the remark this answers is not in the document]
diegohaz: Confirmed real and pre-existing, agreed. I don't think the suggested immediate-delete is the right fix, though: deleting from the lookup on unregister while the public `items` batch is still pending would make `item(id)` return `null` while `collection.getState().items` still contains the item — reintroducing exactly the `item()`/public divergence this PR exists to remove. The current behavior keeps `item()` consistent with the authoritative public state through that one-tick window and self-heals to `null` once the batch flushes. If a specific same-commit reader can act on the disconnected element, the right guard is at the reader (`element.isConnected`) rather than desyncing `item()` from public. Leaving the behavior as-is.

packages/ariakit-components/src/collection/collection-store.ts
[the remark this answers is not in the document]
diegohaz: Confirmed real and pre-existing. The rendered element stays mounted (so it's still focusable), but `item(prunedId)` is `null` while `renderedItems` still lists it — consistent with the model (`item()` tracks the authoritative public state; `renderedItems` tracks the DOM). The robust fix is the navigation-side one (filter composite movement against the lookup), but that lives in `composite-store`'s `getNextId`/`findFirstEnabledItem` — outside this PR's files — and pruning `renderedItems` to match would break the legitimate "rendered but intentionally not in controlled `items`" case. No consumer regression surfaces in the example suite. I'd rather track this as a separate composite-navigation follow-up than reach into `composite-store` from this PR — can file an issue for it.

diegohaz:
Re-triaged the three open review threads against the current `collection-store.ts`. All three are real and **pre-existing** (introduced by the metadata-merge commits, not the later perf/incremental-map rework — each reproduces identically on the pre-rework code).

**Fixed**

- c84d1f5f6 ([reply](https://github.com/ariakit/ariakit/pull/6435#discussion_r3472120713)): *Restore from current public state after divergence* — not a false positive after all. Reproduced via the `unregister(first)` → flush → `unregister(second)` interleaving on overlapping same-id registrations: the second cleanup restored its pre-divergence snapshot and republished it over the controlled `items`, reverting the metadata. `getRestoredItem` now defers to the authoritative public item once a divergence has overwritten the registration's snapshot. Added regressions for the reversion and for stripping an overlapping registration's own metadata after a divergence. (3+ live overlapping registrations on one id remain a pre-existing snapshot-stack limitation needing key-level provenance — out of scope.)

**Working as intended (proposed fix declined)**

- ([reply](https://github.com/ariakit/ariakit/pull/6435#discussion_r3472121232)): *Drop unregistered private items from the lookup immediately* — the lingering item is real, but deleting immediately would make `item(id)` return `null` while `collection.getState().items` still holds it until the batch flushes — reintroducing the `item()`/public divergence this PR removes. Current behavior keeps `item()` consistent with public and self-heals after the batch. A disconnected-element guard, if needed, belongs at the reader.

**Follow-up**

- ([reply](https://github.com/ariakit/ariakit/pull/6435#discussion_r3472121310)): *Keep pruned public items out of rendered navigation* — real and pre-existing. The robust fix (filter composite movement against the lookup) lives in `composite-store`, outside this PR's files, and pruning `renderedItems` would break the legitimate rendered-but-not-in-controlled-`items` case. Tracking as a separate composite-navigation follow-up; no consumer regression in the example suite.

diegohaz:
This PR has grown far beyond the original work and become too complex, I believe unnecessarily so. I'm closing it for now. We'll revisit it in the future.

PR #6436, opened 2026-06-24
Fix FormPush autofocus target

### LINKED ISSUE #6293: FormPush autoFocusOnClick focuses the previous field, and permanently stops working for arrays that start empty
### Current behavior

[`FormPush`](https://ariakit.com/reference/form-push) documents that "the newly added input will be automatically focused when the button is clicked", but with the default [`autoFocusOnClick`](https://ariakit.com/reference/form-push#autofocusonclick) two things go wrong:

1. **Array with existing values:** clicking the button focuses the **previous** last field instead of the newly added one. Typing right after the click silently edits existing data.
2. **Array that starts empty:** the click focuses nothing — and auto-focus is permanently dead for that button afterwards. Later clicks add fields but never focus them.

The root cause is in [`useFormPush`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/form/form-push.tsx#L92-L101): the click handler arms a `shouldFocus` boolean and an effect reads `store.getState().items` exactly once. Since item registration was batched in #3295 ([`collection-store.ts#L87-L94`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L87-L94)), registration writes a private store that is only synced to the public `items` in a `queueMicrotask`, which cannot run before the effect flush in which the focus effect executes. So the effect always observes the pre-click items: the "last item" is the old field, and for an empty array there is no match at all, so `shouldFocus` is never cleared and every later `setShouldFocus(true)` is a no-op.

3. Click **Add tag**: the caret lands in the existing `tags.0` input instead of the new `tags.1` (the "Focused field" output shows `tags.0`).
4. Click **Add email** twice: focus stays on the button both times; `emails.0` and `emails.1` are created but never focused.

```tsx

  const form = Ariakit.useFormStore({
    defaultValues: {
      tags: ["React"],
      emails: [] as string[],
    },
  });

  const tags = Ariakit.useStoreState(form, (state) => state.values.tags);
  const emails = Ariakit.useStoreState(form, (state) => state.values.emails);

  const [focused, setFocused] = useState("none");

  return (
    <Ariakit.Form store={form}>
      <output>Focused field: {focused}</output>

      <fieldset>
        <legend>Tags</legend>
        {tags.map((_, index) => {
          const name = `tags.${index}`;
          return (
            <Ariakit.FormInput
              key={index}
              name={name}
              aria-label={name}
              onFocus={() => setFocused(name)}
            />
          );
        })}
        <Ariakit.FormPush name="tags" value="">
          Add tag
        </Ariakit.FormPush>
      </fieldset>

      <fieldset>
        <legend>Emails</legend>
        {emails.map((_, index) => {
          const name = `emails.${index}`;
          return (
            <Ariakit.FormInput
              key={index}
              name={name}
              aria-label={name}
              onFocus={() => setFocused(name)}
            />
          );
        })}
        <Ariakit.FormPush name="emails" value="">
          Add email
        </Ariakit.FormPush>
      </fieldset>
    </Ariakit.Form>
  );
}
```

Reproduced in Chromium with real-pointer Playwright tests (no mocks; plain `click()` on the buttons): `expect(textbox("tags.1")).toBeFocused()` and `expect(textbox("emails.0")).toBeFocused()` both fail on current code, while probes asserting the wrong destination — `tags.0` focused, and the "Add email" button keeping focus across two clicks — pass.

### Expected behavior

Clicking `FormPush` focuses the field that the click created — `tags.1` in the first scenario, then `emails.0` and `emails.1` in the second — as the documentation states.

### Workaround

Disable `autoFocusOnClick` and focus the field the push will create. `pushValue` appends, so the new field's index is the array length read *before* the push (the consumer `onClick` runs before `FormPush` pushes the value):

```tsx
const focusNewField = (name: string) => {
  const length = form.getValue<string[]>(name)?.length ?? 0;
  requestAnimationFrame(() => {
    const selector = `input[name="${CSS.escape(`${name}.${length}`)}"]`;
    document.querySelector<HTMLElement>(selector)?.focus();
  });
};

<Ariakit.FormPush
  name="tags"
  value=""
  autoFocusOnClick={false}
  onClick={() => focusNewField("tags")}
>
  Add tag
</Ariakit.FormPush>
```

With this applied, the same Playwright assertions above pass in both scenarios.

### Possible solutions

Subscribe the effect to `items` (e.g. via `useStoreState`) and target the **exact index** the click is expected to create instead of "the last field". The effect then bails (without clearing the flag) until the just-registered field is published by the microtask flush, re-runs on the resulting re-render, and focuses the right element — deterministically, with no timing dependence:

```diff
     const name = String(nameProp);
-    const [shouldFocus, setShouldFocus] = useState(false);
+    const [focusIndex, setFocusIndex] = useState<number | null>(null);
+    const items = useStoreState(store, "items");

     useEffect(() => {
-      if (!shouldFocus) return;
-      const items = getFirstFieldsByName(store?.getState().items, name);
-      const element = items?.[items.length - 1]?.element;
+      if (focusIndex == null) return;
+      const fields = getFirstFieldsByName(items, name);
+      const prefix = `${name}.`;
+      const field = fields.find((item) => {
+        const index = item.name.slice(prefix.length).match(/^\d+/)?.[0];
+        return index != null && Number.parseInt(index, 10) === focusIndex;
+      });
+      const element = field?.element;
+      // The just-pushed field registers in the private collection store and
+      // is only flushed to the public `items` in a microtask. Bail without
+      // clearing the flag so the effect re-runs once the flush re-renders.
       if (!element) return;
       element.focus();
-      setShouldFocus(false);
-    }, [store, shouldFocus, name]);
+      setFocusIndex(null);
+    }, [items, focusIndex, name]);
@@
     const onClick = useEvent((event: MouseEvent<HTMLType>) => {
       onClickProp?.(event);
       if (event.defaultPrevented) return;
+      const length = store?.getValue<unknown[]>(name)?.length ?? 0;
       store?.pushValue(name, value);
       if (!autoFocusOnClick) return;
-      setShouldFocus(true);
+      setFocusIndex(length);
     });
```

`store.getValue(name)?.length ?? 0` covers the empty-array case, which is exactly what fixes the permanent latch. `FormRemove` is unaffected — it reads `items` synchronously inside the click handler, where every involved field was registered (and flushed) in earlier commits.

---

*Found by an automated multi-agent audit of the repo at commit [`11bd4458d`](https://github.com/ariakit/ariakit/tree/11bd4458de6447b996dd882ea0c85eb801a6cd03) and verified with a real-browser Playwright reproduction (user-level clicks only).*

diegohaz:
## Motivation

[`FormPush`](https://ariakit.com/reference/form-push) documents that clicking the button focuses the newly added input, but issue #6293 reports two regressions:

- pushing into an array with existing values focuses the previous last field instead of the field that was just added;
- pushing into an initially empty array focuses nothing, and later clicks keep failing to focus the new fields.

This happens because `FormPush` arms a boolean focus flag after `pushValue`, then reads form collection `items` once in an effect. Field registration is published to `items` asynchronously, so the effect can see stale items and either focus the previous field or find no field at all.

## Solution

`FormPush` now records the exact array index that the click is expected to create before pushing the value. It subscribes to the form store's published `items`, keeps that pending index until the matching field appears, and then focuses that field.

The new `form-push-6293` sandbox covers both reported cases with browser tests and a happy-dom duplicate:

- `tags` starts with `tags.0`, then `Add tag` must focus `tags.1`;
- `emails` starts empty, then repeated `Add email` clicks must focus `emails.0` and `emails.1`.

The array field index parser is shared with `FormRemove` so existing exact-name and regex-metacharacter behavior remains covered by the `form-push-remove-6219` regression tests.

## Workaround

Until this fix is released, disable `FormPush`'s built-in autofocus and focus the field index that will be created. The consumer `onClick` runs before `FormPush` pushes the value, so the current array length is the new field's index.

```tsx
// TODO: Remove this workaround once https://github.com/ariakit/ariakit/issues/6293 is fixed.
const focusNewField = (name: "tags" | "emails") => {
  const length = form.getValue<string[]>(name)?.length ?? 0;
  requestAnimationFrame(() => {
    const selector = `input[name="${CSS.escape(`${name}.${length}`)}"]`;
    document.querySelector<HTMLElement>(selector)?.focus();
  });
};

<Ariakit.FormPush
  name="tags"
  value=""
  autoFocusOnClick={false}
  onClick={() => focusNewField("tags")}
>
  Add tag
</Ariakit.FormPush>
```

Fixes #6293


packages/ariakit-react-components/src/form/utils.ts
```diff
@@ -0,0 +1,9 @@
+// Reads the array index off the known `name.` prefix without interpolating the
```
diegohaz, line 1: I believe this file must be named `utils.ts`?

diegohaz: Confirming from the repo conventions: sibling component folders keep a folder's shared helpers in a flat `utils.ts` — see `heading/utils.ts`, `composite/utils.ts`, and `tag/utils.ts`. The `form/` folder is flat and has no `utils.ts` yet, so `getArrayFieldIndex` (a generic form-array helper now shared by `form-push.tsx` and `form-remove.tsx`) fits `form/utils.ts`.

The other established pattern is a `utils/` subfolder of descriptively-named files (e.g. `dialog/utils/is-focus-trap.ts`), which is only worth it if you expect several more form helpers. For a single shared function, `form/utils.ts` matches the closest precedent.

diegohaz (commented):
### Review summary

**Type:** Initial review (no prior review from this skill found on the PR).
**Baseline:** merge-base `953dfa92d` → head `720e4a9e0`.
**Scope:** full diff plus surrounding context — `form-push.tsx`, the extracted `form-array-field.ts`, `form-remove.tsx`, the collection-store `items` publication path, the form store `pushValue`/`removeValue`/`getValue` semantics, the changeset, and the new `form-push-6293` sandbox/tests.

**Validation performed (local):**
- `pnpm test-react run form-push` → 5/5 pass; `pnpm test-react18 run form-push` → 5/5 pass (focus timing is the crux, so I checked both React versions given known R18+happy-dom focus quirks).
- Reverted only `form-push.tsx` to the base version and re-ran the new repro: both new tests fail (empty-array case focuses nothing → `null`), confirming they are genuine regression tests that lock in the fix.
- `tsc -b` passes. (`pnpm tsc` fails earlier in `astro check` with a vite dep-optimizer cache error — a fresh-worktree environment flake, unrelated to this PR.)
- Verified the changeset reference URL (`/reference/form-push`) returns 200.

**Findings:**
- The implementation is correct and a clean improvement over the previous heuristic. Recording the pre-push array `length` as the exact target index and waiting for that field to publish into `items` is robust to the `null` holes that `removeValue` leaves behind (it writes `null` rather than reindexing), so index-based matching avoids the old position-based "last field" failure mode. Effect deps are right, and `onClickProp` runs before the `length` snapshot so the documented workaround ordering holds.
- Sharing `getArrayFieldIndex` with `FormRemove` is a good dedup; the existing `form-push-remove-6219` regression tests (prefix-sibling + regex-metacharacter) still pass against the refactored `FormPush`.
- Changeset scoping is correct: `@ariakit/react` is included alongside `@ariakit/react-components` because `FormPush` is part of the public re-export surface (unlike the experimental renderers, which omit it).
- One actionable item, replied on the existing thread: the helper file should follow the flat `utils.ts` convention used by `heading/`, `composite/`, and `tag/` (i.e. `form/utils.ts`) rather than `form-array-field.ts`.

**Considered but not flagged:**
- `useFormPush` now calls `useStoreState(store, "items")`, which is a new subscription — `FormPush` previously had no store subscription (`useCollectionItem` only registers via effect), so every `FormPush` now re-renders on any form `items` change (field/button mount/unmount). The cost is small and infrequent (a single button hook chain, not on keystrokes), and the only alternative that avoids it — a transient `store.sync` subscription inside the effect, active only while a focus is pending — is more complex rather than simpler. The current idiomatic approach is the right trade-off.

**Not validated locally:** the Playwright `test-browser.ts` suite (worktree dev-server port setup is fragile here); the happy-dom duplicate mirrors those assertions and passes on R18/R19, and CI covers the browser run.

packages/ariakit-react-components/src/form/utils.ts
diegohaz: Fixed in f5ee6f64f - renamed the shared form array helper to `form/utils.ts` and updated the `FormPush` and `FormRemove` imports.

diegohaz:
**Fixed**
- f5ee6f64f ([reply](https://github.com/ariakit/ariakit/pull/6436#discussion_r3469630502)): Renamed the shared form array helper to `form/utils.ts` and updated the `FormPush` and `FormRemove` imports to match the folder helper convention.

PR #6439, opened 2026-06-24
Fix ComboboxItemValue overlapping matches

### LINKED ISSUE #6298: ComboboxItemValue renders duplicated text when matches of the user input overlap ("Banana" + "ana" displays "Banaana")
### Current behavior

When the user's input matches an item's value at overlapping positions, [`ComboboxItemValue`](https://ariakit.com/reference/combobox-item-value) renders the overlapped characters twice. Typing `ana` in a combobox with a `Banana` item displays **"Banaana"**. Since the spans are the option element's content, the option's accessible name is corrupted too ("Banaana"), so screen readers announce the wrong item text and name-based queries for "Banana" fail.

The offset pipeline in [combobox-item-value.tsx](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx) produces overlapping match offsets and the slicing step assumes they're disjoint:

- `getOffsets` ([L24-L37](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx#L24-L37)) advances the scan by one character per hit (`pos = index + 1`), so "banana"/"ana" yields offsets `[1, 3]` and `[3, 3]`.
- `filterOverlappingOffsets` ([L39-L45](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx#L39-L45)) only removes offsets fully *contained* in another; partially overlapping pairs like `[1, 3]`/`[3, 3]` both survive.
- The slicing in `splitValue` ([L83-L91](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx#L83-L91)) emits each match's full slice plus the gap to the next offset. When the next offset starts *inside* the current match, the gap slice is empty but the next iteration re-emits the overlapped characters — `"B"`, `"ana"` (from index 1), `"ana"` (from index 3) → "Banaana".

This breaks the component's documented contract that the item value is merely *split* into `data-user-value`/`data-autocomplete-value` portions ([JSDoc](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx#L101-L120)), i.e. that the portions concatenate back to the item value.

Any self-overlapping query triggers it: "ana" against "Banana", a doubled-letter query like "aa" against "aaa" (renders "aaaa"), etc. With the [`userValue`](https://ariakit.com/reference/combobox-item-value#uservalue) array prop, partial overlaps *between* different tokens (e.g. `["ab", "bc"]` on "abc" → "abbc") hit the same bug.

3. Click the "Your favorite fruit" combobox input and type `ana`.
4. The popover shows the single matching option, but it reads **"Banaana"** instead of "Banana" — the overlapped "a" is rendered twice (two highlighted "ana" spans). The option's accessible name is "Banaana" as well.

```tsx
import { useMemo, useState } from "react";

const fruits = ["Apple", "Banana", "Cherry", "Grape", "Orange"];

  const matches = useMemo(() => {
    const search = searchValue.toLowerCase();
    return fruits.filter((fruit) => fruit.toLowerCase().includes(search));
  }, [searchValue]);
  return (
    <Ariakit.ComboboxProvider setValue={setSearchValue}>
      <label>
        Your favorite fruit
        <Ariakit.Combobox placeholder="e.g., Apple" />
      </label>
      <Ariakit.ComboboxPopover gutter={8} sameWidth>
        {matches.map((fruit) => (
          <Ariakit.ComboboxItem key={fruit} value={fruit}>
            <Ariakit.ComboboxItemValue />
          </Ariakit.ComboboxItem>
        ))}
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

Verified with a Playwright test driving only real user interactions (click, type "ana"): the option's text content and accessible name are "Banaana" with two `data-user-value` spans each containing "ana", instead of "Banana".

### Expected behavior

The highlighted portions should concatenate back to the original item value, as the component's docs describe. The option should display "Banana" (with the matched range "anana" — the union of the two overlapping matches — highlighted), and its accessible name should remain "Banana".

### Workaround

`ComboboxItemValue` gives user-provided `children` precedence over its computed split, so you can compute the highlighted portions yourself with overlapping ranges merged, keeping the same span/attribute structure for CSS:

```tsx
import type { ReactNode } from "react";

function getHighlightedParts(itemValue: string, userValue: string) {
  const parts: ReactNode[] = [];
  const normalizedItem = itemValue.toLowerCase();
  const normalizedUser = userValue.toLowerCase();
  if (!normalizedUser) return [itemValue];
  // Collect [start, end) ranges of every match, merging overlaps.
  const ranges: Array<[number, number]> = [];
  let index = normalizedItem.indexOf(normalizedUser);
  while (index !== -1) {
    const end = index + normalizedUser.length;
    const lastRange = ranges[ranges.length - 1];
    if (lastRange && index < lastRange[1]) {
      lastRange[1] = Math.max(lastRange[1], end);
    } else {
      ranges.push([index, end]);
    }
    index = normalizedItem.indexOf(normalizedUser, index + 1);
  }
  let position = 0;
  for (const [start, end] of ranges) {
    if (start > position) {
      parts.push(
        <span key={parts.length} data-autocomplete-value="">
          {itemValue.slice(position, start)}
        </span>,
      );
    }
    parts.push(
      <span key={parts.length} data-user-value="">
        {itemValue.slice(start, end)}
      </span>,
    );
    position = end;
  }
  if (position < itemValue.length) {
    parts.push(
      <span key={parts.length} data-autocomplete-value="">
        {itemValue.slice(position)}
      </span>,
    );
  }
  return parts;
}

// In the item, pass the parts as children (searchValue is the combobox value):
<Ariakit.ComboboxItem key={fruit} value={fruit}>
  <Ariakit.ComboboxItemValue>
    {getHighlightedParts(fruit, searchValue)}
  </Ariakit.ComboboxItemValue>
</Ariakit.ComboboxItem>;
```

Caveat: unlike the built-in split, this simple version doesn't strip diacritics when matching (the component normalizes with NFD) — extend the normalization if you need that.

### Possible solutions

The slicing logic is fine as long as the offsets it receives are sorted and disjoint, so make that invariant true: replace `filterOverlappingOffsets` + `sortOffsets` (no other callers in the repo) with a single helper that sorts and *merges* overlapping offsets into their union. The merge condition must be a strict `<`: merging *adjacent* offsets would collapse the documented two-span output for "Apple" + "p" ([JSDoc](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx#L178-L185)) into a single "pp" span, breaking the existing snapshot test and per-match `[data-user-value]` styling.

```diff
-function filterOverlappingOffsets(offsets: Array<[number, number]>) {
-  return offsets.filter(([offset, length], i, arr) => {
-    return !arr.some(
-      ([o, l], j) => j !== i && o <= offset && o + l >= offset + length,
-    );
-  });
-}
-
-function sortOffsets(offsets: Array<[number, number]>) {
-  return offsets.sort(([a], [b]) => a - b);
-}
+function mergeOverlappingOffsets(offsets: Array<[number, number]>) {
+  offsets.sort(([a], [b]) => a - b);
+  const merged: Array<[number, number]> = [];
+  for (const [offset, length] of offsets) {
+    const last = merged[merged.length - 1];
+    // Merge only strictly overlapping offsets. Adjacent matches must remain
+    // separate so repeated matches keep rendering as individual spans.
+    if (last && offset < last[0] + last[1]) {
+      last[1] = Math.max(last[1], offset + length - last[0]);
+    } else {
+      merged.push([offset, length]);
+    }
+  }
+  return merged;
+}
```

and in `splitValue`:

```diff
-  const offsets = sortOffsets(
-    filterOverlappingOffsets(
-      // Convert userValues into a set to avoid duplicates
-      getOffsets(normalizeValue(itemValue), new Set(userValues)),
-    ),
-  );
+  const offsets = mergeOverlappingOffsets(
+    // Convert userValues into a set to avoid duplicates
+    getOffsets(normalizeValue(itemValue), new Set(userValues)),
+  );
```

Merging subsumes both deleted helpers: fully contained offsets merge into the larger one, and the merged array is already sorted. `getOffsets` should keep `pos = index + 1` — advancing by the match length would highlight only the first "ana" instead of the union "anana", and still wouldn't fix cross-token overlaps from `userValue` arrays (e.g. `["ab", "bc"]` on "abc"), which need merging anyway.

With the patch: "Banana"/"ana" → `["B", "anana"]`, "aaa"/"aa" → `["aaa"]`, "abc"/`["ab", "bc"]` → `["abc"]`, while "Apple"/"p" still renders two separate `data-user-value` spans, identical to today's documented output.

diegohaz:
## Motivation

Fixes #6298.

`ComboboxItemValue` rendered overlapping user input matches as separate full slices. For example, typing `ana` in a combobox with a `Banana` item produced duplicated text like `Banaana`, which also corrupted the option's accessible name.

## Solution

Added a sandbox at `app/src/sandbox/combobox-item-value-6298` that reproduces the issue with real `ComboboxItem` and `ComboboxItemValue` usage. The repro includes both the primary browser test and a happy-dom duplicate.

Updated `ComboboxItemValue` to sort and merge overlapping match offsets before slicing the item value. The merge is strict, so overlapping ranges like the two `ana` matches in `Banana` become one highlighted `anana` span, while adjacent repeated matches such as the two `p` matches in `Apple` still render as separate `data-user-value` spans.

The offset scanner now skips empty normalized user values directly, so a truthy value that normalizes to an empty string cannot make `indexOf` loop forever on an empty needle.

## Workaround

Until this fix is released, consumers can pass explicit children to `ComboboxItemValue` and merge overlapping match ranges in userland:

```tsx
// TODO: Remove this workaround after
// https://github.com/ariakit/ariakit/issues/6298 is fixed.
function getHighlightedParts(itemValue: string, userValue: string) {
  const parts: ReactNode[] = [];
  const normalizedItem = itemValue.toLowerCase();
  const normalizedUser = userValue.toLowerCase();
  if (!normalizedUser) return itemValue;

  const ranges: Array<[number, number]> = [];
  let index = normalizedItem.indexOf(normalizedUser);

  while (index !== -1) {
    const end = index + normalizedUser.length;
    const lastRange = ranges[ranges.length - 1];
    if (lastRange && index < lastRange[1]) {
      lastRange[1] = Math.max(lastRange[1], end);
    } else {
      ranges.push([index, end]);
    }
    index = normalizedItem.indexOf(normalizedUser, index + 1);
  }

  let position = 0;
  for (const [start, end] of ranges) {
    if (start > position) {
      parts.push(
        <span key={parts.length} data-autocomplete-value="">
          {itemValue.slice(position, start)}
        </span>,
      );
    }
    parts.push(
      <span key={parts.length} data-user-value="">
        {itemValue.slice(start, end)}
      </span>,
    );
    position = end;
  }

  if (position < itemValue.length) {
    parts.push(
      <span key={parts.length} data-autocomplete-value="">
        {itemValue.slice(position)}
      </span>,
    );
  }

  return parts;
}

<Ariakit.ComboboxItem value={fruit}>
  <Ariakit.ComboboxItemValue>
    {getHighlightedParts(fruit, searchValue)}
  </Ariakit.ComboboxItemValue>
</Ariakit.ComboboxItem>;
```

### DIFF
--- a/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
@@ -36,22 +36,29 @@ function getOffsets(string: string, values: Iterable<string>) {
   return offsets;
 }
 
-function filterOverlappingOffsets(offsets: Array<[number, number]>) {
-  return offsets.filter(([offset, length], i, arr) => {
-    return !arr.some(
-      ([o, l], j) => j !== i && o <= offset && o + l >= offset + length,
-    );
-  });
-}
+function mergeOverlappingOffsets(offsets: Array<[number, number]>) {
+  offsets.sort(([a], [b]) => a - b);
 
-function sortOffsets(offsets: Array<[number, number]>) {
-  return offsets.sort(([a], [b]) => a - b);
+  const merged: Array<[number, number]> = [];
+
+  for (const [offset, length] of offsets) {
+    const last = merged[merged.length - 1];
+    // Adjacent matches should remain separate so repeated matches keep
+    // rendering as individual spans.
+    if (last && offset < last[0] + last[1]) {
+      last[1] = Math.max(last[1], offset + length - last[0]);
+    } else {
+      merged.push([offset, length]);
+    }
+  }
+
+  return merged;
 }
 
 function splitValue(itemValue?: string | null, userValue?: string | string[]) {
   if (!itemValue) return itemValue;
   if (!userValue) return itemValue;
-  const userValues = toArray(userValue).filter(Boolean).map(normalizeValue);
+  const userValues = toArray(userValue).map(normalizeValue).filter(Boolean);
   const parts: ReactElement[] = [];
 
   const span = (value: string, autocomplete = false) => (
@@ -64,11 +71,9 @@ function splitValue(itemValue?: string | null, userValue?: string | string[]) {
     </span>
   );
 
-  const offsets = sortOffsets(
-    filterOverlappingOffsets(
-      // Convert userValues into a set to avoid duplicates
-      getOffsets(normalizeValue(itemValue), new Set(userValues)),
-    ),
+  const offsets = mergeOverlappingOffsets(
+    // Convert userValues into a set to avoid duplicates
+    getOffsets(normalizeValue(itemValue), new Set(userValues)),
   );
 
   const firstEntry = offsets[0];
--- a/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
@@ -24,6 +24,8 @@ function normalizeValue(value: string) {
 function getOffsets(string: string, values: Iterable<string>) {
   const offsets = [] as Array<[number, number]>;
   for (const value of values) {
+    // An empty value makes indexOf keep returning the end of the string.
+    if (!value) continue;
     let pos = 0;
     const length = value.length;
     let index = string.indexOf(value, pos);
@@ -58,7 +60,7 @@ function mergeOverlappingOffsets(offsets: Array<[number, number]>) {
 function splitValue(itemValue?: string | null, userValue?: string | string[]) {
   if (!itemValue) return itemValue;
   if (!userValue) return itemValue;
-  const userValues = toArray(userValue).map(normalizeValue).filter(Boolean);
+  const userValues = toArray(userValue).map(normalizeValue);
   const parts: ReactElement[] = [];
 
   const span = (value: string, autocomplete = false) => (
--- a/.changeset/300-combobox-item-value-overlap.md
+++ b/.changeset/300-combobox-item-value-overlap.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`ComboboxItemValue`](https://ariakit.com/reference/combobox-item-value) so overlapping user input matches are rendered without duplicated text.
--- a/app/src/sandbox/combobox-item-value-6298/index.react.tsx
+++ b/app/src/sandbox/combobox-item-value-6298/index.react.tsx
@@ -1,0 +1,27 @@
+import * as Ariakit from "@ariakit/react";
+import { useMemo, useState } from "react";
+
+const fruits = ["Apple", "Banana", "Cherry", "Grape", "Orange"];
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6298
+export default function Example() {
+  const [searchValue, setSearchValue] = useState("");
+  const matches = useMemo(() => {
+    const search = searchValue.toLowerCase();
+    return fruits.filter((fruit) => fruit.toLowerCase().includes(search));
+  }, [searchValue]);
+
+  return (
+    <Ariakit.ComboboxProvider defaultOpen setValue={setSearchValue}>
+      <Ariakit.ComboboxLabel>Your favorite fruit</Ariakit.ComboboxLabel>
+      <Ariakit.Combobox placeholder="e.g., Apple" />
+      <Ariakit.ComboboxPopover gutter={8} sameWidth>
+        {matches.map((fruit) => (
+          <Ariakit.ComboboxItem key={fruit} value={fruit}>
+            <Ariakit.ComboboxItemValue />
+          </Ariakit.ComboboxItem>
+        ))}
+      </Ariakit.ComboboxPopover>
+    </Ariakit.ComboboxProvider>
+  );
+}
--- a/app/src/sandbox/combobox-item-value-6298/index.react.tsx
+++ b/app/src/sandbox/combobox-item-value-6298/index.react.tsx
@@ -1,58 +1,8 @@
 import * as Ariakit from "@ariakit/react";
-import type { ReactNode } from "react";
 import { useMemo, useState } from "react";
 
 const fruits = ["Apple", "Banana", "Cherry", "Grape", "Orange"];
-
-// TODO: Remove this workaround after
-// https://github.com/ariakit/ariakit/issues/6298 is fixed.
-function getHighlightedParts(itemValue: string, userValue: string) {
-  const parts: ReactNode[] = [];
-  const normalizedItem = itemValue.toLowerCase();
-  const normalizedUser = userValue.toLowerCase();
-  if (!normalizedUser) return itemValue;
-
-  const ranges: Array<[number, number]> = [];
-  let index = normalizedItem.indexOf(normalizedUser);
-
-  while (index !== -1) {
-    const end = index + normalizedUser.length;
-    const lastRange = ranges[ranges.length - 1];
-    if (lastRange && index < lastRange[1]) {
-      lastRange[1] = Math.max(lastRange[1], end);
-    } else {
-      ranges.push([index, end]);
-    }
-    index = normalizedItem.indexOf(normalizedUser, index + 1);
-  }
-
-  let position = 0;
-  for (const [start, end] of ranges) {
-    if (start > position) {
-      parts.push(
-        <span key={parts.length} data-autocomplete-value="">
-          {itemValue.slice(position, start)}
-        </span>,
-      );
-    }
-    parts.push(
-      <span key={parts.length} data-user-value="">
-        {itemValue.slice(start, end)}
-      </span>,
-    );
-    position = end;
-  }
-
-  if (position < itemValue.length) {
-    parts.push(
-      <span key={parts.length} data-autocomplete-value="">
-        {itemValue.slice(position)}
-      </span>,
-    );
-  }
-
-  return parts;
-}
+const normalizedEmptyUserValue = "\u0301";
 
 // Reproduces https://github.com/ariakit/ariakit/issues/6298
 export default function Example() {
@@ -69,12 +19,16 @@ export default function Example() {
       <Ariakit.ComboboxPopover gutter={8} sameWidth>
         {matches.map((fruit) => (
           <Ariakit.ComboboxItem key={fruit} value={fruit}>
-            <Ariakit.ComboboxItemValue>
-              {getHighlightedParts(fruit, searchValue)}
-            </Ariakit.ComboboxItemValue>
+            <Ariakit.ComboboxItemValue />
           </Ariakit.ComboboxItem>
         ))}
       </Ariakit.ComboboxPopover>
+      <output aria-label="Normalized empty value">
+        <Ariakit.ComboboxItemValue
+          value="Cafe"
+          userValue={normalizedEmptyUserValue}
+        />
+      </output>
     </Ariakit.ComboboxProvider>
   );
 }
--- a/app/src/sandbox/combobox-item-value-6298/index.react.tsx
+++ b/app/src/sandbox/combobox-item-value-6298/index.react.tsx
@@ -1,7 +1,58 @@
 import * as Ariakit from "@ariakit/react";
+import type { ReactNode } from "react";
 import { useMemo, useState } from "react";
 
 const fruits = ["Apple", "Banana", "Cherry", "Grape", "Orange"];
+
+// TODO: Remove this workaround after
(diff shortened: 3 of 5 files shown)


packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
```diff
@@ -36,22 +36,29 @@ function getOffsets(string: string, values: Iterable<string>) {
...
-function sortOffsets(offsets: Array<[number, number]>) {
-  return offsets.sort(([a], [b]) => a - b);
+  return merged;
 }
 
 function splitValue(itemValue?: string | null, userValue?: string | string[]) {
   if (!itemValue) return itemValue;
   if (!userValue) return itemValue;
-  const userValues = toArray(userValue).filter(Boolean).map(normalizeValue);
+  const userValues = toArray(userValue).map(normalizeValue).filter(Boolean);
```
diegohaz: **The normalize/filter order here is load-bearing — consider hardening `getOffsets` so a refactor can't undo it.**

Reordering `.filter(Boolean).map(normalizeValue)` → `.map(normalizeValue).filter(Boolean)` is what stops a truthy user value that normalizes to an empty string (e.g. a lone combining mark like `"́"`) from reaching `getOffsets` as an empty needle. That matters because an empty needle makes `getOffsets` loop forever: `string.indexOf("", pos)` clamps to `string.length` and never returns `-1`, so `while (index !== -1)` never terminates. Since it runs synchronously during render, it's a hard tab freeze — not even a catchable Vitest/Playwright timeout.

I confirmed this locally: with the old `filter(Boolean).map(...)` order, `getOffsets("cafe", new Set([""]))` spins indefinitely; with the new order the empty needle is dropped first.

My only concern is that the new ordering reads like a cosmetic swap, so a future "tidy-up" could reorder it and silently reintroduce the freeze. Since the real footgun lives in `getOffsets`, I'd suggest hardening it at the source so it's safe regardless of how callers pre-filter:

```ts
function getOffsets(string: string, values: Iterable<string>) {
  const offsets = [] as Array<[number, number]>;
  for (const value of values) {
    // An empty needle never lets `indexOf` return -1, so skip it to avoid
    // an infinite loop.
    if (!value) continue;
    let pos = 0;
    // ...
```

That keeps the call-site reorder working but removes the latent infinite loop entirely. At minimum, a short comment on this line noting the order is intentional would prevent an accidental swap.

Not blocking — the current fix is correct.

diegohaz (commented):
**Review summary — initial review**

Reviewed the full diff from merge-base `953dfa9` to head `9e05bfe`, plus the surrounding `combobox-item-value.tsx` context (`getOffsets`, `splitValue`, the `useComboboxItemValue` hook), the new sandbox, and the changeset.

**Verdict: the fix is correct, well-targeted, and well-tested.** `mergeOverlappingOffsets` cleanly replaces the contained-only `filterOverlappingOffsets` + `sortOffsets` pipeline with a single sort-and-merge that guarantees the *sorted + disjoint* invariant `splitValue`'s slicing already assumed. The strict `<` merge condition preserves the documented two-span output for adjacent repeats (`"Apple"` + `"p"`), while genuinely overlapping matches now collapse to their union. It also drops the old O(n²) `filter`+`some` to O(n log n).

**Validation performed:**
- Reimplemented `getOffsets` + `mergeOverlappingOffsets` + `splitValue`'s slicing in a standalone script and verified the rendered spans reassemble to the original item value in every case: `"Banana"/"ana"` → `["B","anana"]`, `"Apple"/"p"` → two `p` spans, `"aaa"/"aa"` → `["aaa"]`, `"abc"/["ab","bc"]` → `["abc"]`, `"Banana"/"an"` → two disjoint `an` spans, `"Mississippi"/"iss"` → two disjoint `iss` spans.
- Confirmed the `.map(normalizeValue).filter(Boolean)` reorder fixes a real **infinite loop** (an empty needle in `getOffsets` never lets `indexOf` return `-1`), not merely duplicated text.
- Ran the happy-dom test locally: `pnpm test-react run app/src/sandbox/combobox-item-value-6298/test.ts` → **1 passed**.

**Findings:**
- One non-blocking maintainability/robustness suggestion (inline on line 61): the normalize/filter reorder is load-bearing but looks cosmetic; consider guarding `getOffsets` against empty needles at the source so a future reorder can't reintroduce the freeze.

I also evaluated CodeRabbit's nitpick (assert zero `[data-user-value]` spans in the normalized-empty case). It's an optional strengthening only — the existing `toHaveText("Cafe")` assertion already guards that regression, because the pre-fix code hangs *synchronously on mount* for that input, so the test would never reach the assertion at all.

**Not fully validated:** the Playwright `test-browser.ts` (didn't start the app dev server) and a full `pnpm tsc` / `pnpm build`. The change is small and type-trivial, and CI covers these.

packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
diegohaz: Fixed in 000b7d0b3 - `getOffsets` now skips empty values directly before the `indexOf` loop, and the normalized-empty sandbox assertions verify no `data-user-value` span is rendered.

diegohaz:
**Fixed**

- 000b7d0b3 (https://github.com/ariakit/ariakit/pull/6439#discussion_r3469571017): Hardened `getOffsets` against empty normalized values and tightened the sandbox assertions for the normalized-empty case.

packages/ariakit-react-components/src/combobox/combobox-item-value.tsx
diegohaz: Verified on `000b7d0b3` — this fully resolves it, and dropping the now-redundant `.filter(Boolean)` at the call site is a nice improvement: the empty-needle invariant now lives in exactly one place (`getOffsets`), so the two can't drift. Behavior is unchanged — `map(normalizeValue)` keeps empties and `getOffsets` skips them — and the loop can't recur:

- `getOffsets("cafe", new Set([""]))` → `[]` (no spin); `"Cafe"` + `"́"` and `["", "́"]` both render 0 `data-user-value` spans.
- `"Banana"/"ana"` → one `anana` span, `"Apple"/"p"` → two `p` spans, `"abc"/["ab","bc"]` → one `abc` span — all unchanged.

Re-ran `pnpm test-react run …combobox-item-value-6298/test.ts` → 1 passed.

diegohaz (commented):
**Re-review — covers commit `000b7d0b3` ("Harden ComboboxItemValue empty matches") only**

Baseline: my last reviewed head `9e05bfe`. New head: `000b7d0b3`.

This commit resolves the one finding from my prior review and also addresses CodeRabbit's nitpick:

- **`getOffsets` now guards empty needles at the source** (`if (!value) continue;` with a WHY comment) — the root-cause fix I suggested. The author additionally removed the now-redundant `.filter(Boolean)` at the call site, so the empty-needle invariant lives in a single place and the two can't drift. Behavior is unchanged: empties are kept by `map(normalizeValue)` and skipped inside `getOffsets`.
- **Both tests now assert zero `[data-user-value]` spans** in the normalized-empty case (`test.ts` and `test-browser.ts`), addressing CodeRabbit's nit.

**Validation:** re-traced the offset pipeline with the new shape — `"Banana"/"ana"` → one `anana` span, `"Apple"/"p"` → two `p` spans, `"abc"/["ab","bc"]` → one `abc` span, and `"Cafe"/"́"` plus `"Cafe"/["","́"]` → zero user spans with no infinite loop. Re-ran the happy-dom test (`pnpm test-react run …combobox-item-value-6298/test.ts`) → **1 passed**.

No new issues. The change is correct, complete, and a clean simplification over my suggested patch. **LGTM.**

Not re-run this round: Playwright `test-browser.ts` and full `pnpm tsc` / `pnpm build` (CI covers these).

PR #6440, opened 2026-06-24
Fix ToolbarContainer empty deletion focus

### LINKED ISSUE #6300: ToolbarContainer: Backspace on a container with an empty field leaves a stale listener that steals focus on the next keystroke
### Current behavior

When a [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container) (or the underlying `CompositeContainer` from `@ariakit/react-components`) has focus and the user presses <kbd>Backspace</kbd> or <kbd>Delete</kbd>, the keydown handler opens the container — focusing its inner text field with the value selected so the key's default action clears it — and arms a one-shot `input` listener on the container whose job is to move focus back to the container right after the deletion: [composite-container.tsx#L175-L188](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-container.tsx#L175-L188).

The `input` event only fires when the field's value actually changes. If the field is already empty, there is nothing to delete — [`selectTextField`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/utils.ts#L20-L33) selects `0..0` and the key's default action is a no-op — so the `{ once: true }` listener stays armed indefinitely while focus now sits inside the field. Nothing ever removes it: neither [`close()`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-container.tsx#L83-L88) nor the focus/blur handlers touch it.

The next genuine `input` event that bubbles to the container — typically the user's very next keystroke in the field — triggers the stale handler, which calls `container.focus()`. The resulting self-target focus runs [the branch that disables focus on everything inside the container](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-container.tsx#L114-L128), so the user is ejected from the field mid-typing. The listener also survives blur, so it can fire much later (for example, the user tabs away, comes back, opens the container by typing a printable character — and that first keystroke is hijacked), and each additional Backspace press on an empty field stacks another armed listener.

3. Click "Before toolbar", then press <kbd>Tab</kbd> to focus the "Font family" container.
4. Press <kbd>Backspace</kbd>. The field is cleared and focus returns to the container (designed behavior).
5. Press <kbd>Backspace</kbd> again. The field is already empty; focus moves into the field.
6. Type a letter. The letter lands in the field, but focus immediately jumps back to the container, ejecting you mid-typing.

```tsx

  return (
    <>
      <button type="button">Before toolbar</button>
      <Ariakit.Toolbar
        aria-label="Text formatting"
        style={{ display: "flex", gap: 8 }}
      >
        <Ariakit.ToolbarContainer aria-label="Font family" role="group">
          <label>
            Font family <input defaultValue="Inter" />
          </label>
        </Ariakit.ToolbarContainer>
        <Ariakit.ToolbarItem>Apply</Ariakit.ToolbarItem>
      </Ariakit.Toolbar>
    </>
  );
}
```

Verified with a Playwright test in Chrome driving only real keyboard interactions: asserting that focus stays in the field while typing fails, and a probe confirms the typed character lands in the field while focus ends up back on the container.

### Expected behavior

Typing in the focused field should keep focus in the field, exactly as it does when the field is entered via a printable key, <kbd>Enter</kbd>, or a click. The refocus listener is meant to handle only the deletion triggered by the same Backspace/Delete keystroke that armed it (per the code comment: "as soon as the delete/backspace key is captured by the text field"), so when that keystroke deletes nothing, the listener should be discarded instead of lying in wait for an unrelated `input` event.

### Workaround

Intercept the empty-field case in a custom `onKeyDown` on the container (Ariakit's built-in handler bails when `event.defaultPrevented` is set) and move focus into the field manually, skipping the listener entirely:

```tsx
const inputRef = React.useRef<HTMLInputElement>(null);

// ...
<Ariakit.ToolbarContainer
  aria-label="Font family"
  role="group"
  onKeyDown={(event) => {
    if (event.key !== "Delete" && event.key !== "Backspace") return;
    if (event.target !== event.currentTarget) return;
    const input = inputRef.current;
    if (!input || input.value) return;
    // Empty field: the key press won't fire an "input" event, so the
    // built-in behavior would leave a stale one-shot listener behind.
    event.preventDefault();
    input.focus();
  }}
>
  <label>
    Font family <input ref={inputRef} defaultValue="Inter" />
  </label>
</Ariakit.ToolbarContainer>
```

Verified: the failing Playwright test passes with this handler applied. The non-empty path still uses the built-in behavior, which works correctly.

### Possible solutions

Bound the listener's lifetime to the keystroke that armed it. The legitimate `input` event is part of the keydown's default action and fires within the same task as the keydown (the ordering the feature already relies on), so a `setTimeout(..., 0)` scheduled in the keydown handler can only reap a listener that never fired; if it did fire, `{ once: true }` already removed it and the extra `removeEventListener` is a no-op:

```diff
         if (isTextField(tabbable) || tabbable.isContentEditable) {
           open();
           // We need to move focus back to the container as soon as the
           // delete/backspace key is captured by the text field.
           const onInput = () => queueMicrotask(() => container.focus());
           container.addEventListener("input", onInput, { once: true });
+          // The input event triggered by the key's default action fires
+          // within the same task as this keydown. If it never fires (for
+          // example, when the text field is already empty, so nothing is
+          // deleted), remove the stale listener so it can't hijack an
+          // unrelated input event later.
+          setTimeout(() => container.removeEventListener("input", onInput), 0);
         }
```

Key auto-repeat is safe (each keydown pairs its own listener with its own cleanup timer), and the non-empty-field path is untouched because its `input` event always fires before the timer.

diegohaz:
## Motivation

Pressing <kbd>Backspace</kbd> or <kbd>Delete</kbd> on a focused [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container) opens its inner field and registers a one-shot `input` listener to move focus back to the container after the deletion.

When the field is already empty, the key press does not fire an `input` event, so that listener stays armed. The next typed character lands in the field, then the stale listener moves focus back to the container.

Fixes #6300.

## Solution

This adds a focused sandbox repro under `app/src/sandbox/toolbar-container-6300/` and bounds the stale listener lifetime in `CompositeContainer`.

The repro covers:

- <kbd>Backspace</kbd> and <kbd>Delete</kbd> on an already-empty field across desktop browser projects.
- The reported filled-field sequence in Chrome, where the first <kbd>Backspace</kbd> clears `Inter`, the second <kbd>Backspace</kbd> used to arm the stale listener, and typing was interrupted.

The library fix removes the one-shot `input` listener on the matching <kbd>keyup</kbd>, with a zero-delay timer as a fallback. When the deletion actually changes the field value, the `input` handler still moves focus back to the container. When no `input` event fires because the field is empty, the listener is cleaned up before it can steal focus from the next typed character.

## Workaround

Intercept <kbd>Backspace</kbd> and <kbd>Delete</kbd> when the `ToolbarContainer` itself has focus and the inner field is empty. Preventing the default action makes Ariakit skip the built-in deletion path, so it does not register a stale `input` listener, then focus the field directly.

```tsx
const inputRef = React.useRef<HTMLInputElement>(null);

<Ariakit.ToolbarContainer
  aria-label="Font family"
  role="group"
  onKeyDown={(event) => {
    if (event.key !== "Backspace" && event.key !== "Delete") return;
    if (event.target !== event.currentTarget) return;
    const input = inputRef.current;
    if (!input) return;
    if (input.value) return;
    // TODO: Remove this workaround after
    // https://github.com/ariakit/ariakit/issues/6300 is fixed.
    event.preventDefault();
    input.focus();
  }}
>
  <label>
    Font family <input ref={inputRef} />
  </label>
</Ariakit.ToolbarContainer>
```

### DIFF
--- a/app/src/sandbox/toolbar-container-6300/test-chrome.ts
+++ b/app/src/sandbox/toolbar-container-6300/test-chrome.ts
@@ -1,0 +1,28 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6300
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("keeps focus in the field after Backspace clears the field first", async ({
+    page,
+    q,
+  }) => {
+    const container = q.group("Font family");
+    const input = q.textbox("Font family");
+
+    await q.button("Before filled toolbar").focus();
+    await page.keyboard.press("Tab");
+
+    await test.expect(container).toBeFocused();
+    await page.keyboard.press("Backspace");
+    await test.expect(input).toHaveValue("");
+    await test.expect(container).toBeFocused();
+
+    await page.keyboard.press("Backspace");
+    await test.expect(input).toBeFocused();
+
+    await page.keyboard.press("a");
+
+    await test.expect(input).toHaveValue("a");
+    await test.expect(input).toBeFocused();
+  });
+});
--- a/packages/ariakit-react-components/src/composite/composite-container.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-container.tsx
@@ -182,8 +182,17 @@ export const useCompositeContainer = createHook<
           open();
           // We need to move focus back to the container as soon as the
           // delete/backspace key is captured by the text field.
-          const onInput = () => queueMicrotask(() => container.focus());
+          const cleanup = () => {
+            container.removeEventListener("input", onInput);
+            container.removeEventListener("keyup", cleanup);
+          };
+          const onInput = () => {
+            cleanup();
+            queueMicrotask(() => container.focus());
+          };
           container.addEventListener("input", onInput, { once: true });
+          container.addEventListener("keyup", cleanup, { once: true });
+          setTimeout(cleanup, 0);
         }
       }
     }
--- a/packages/ariakit-react-components/src/composite/composite-container.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-container.tsx
@@ -180,8 +180,15 @@ export const useCompositeContainer = createHook<
         if (!tabbable) return;
         if (isTextField(tabbable) || tabbable.isContentEditable) {
           open();
-          // We need to move focus back to the container as soon as the
-          // delete/backspace key is captured by the text field.
+          // Move focus back only when this delete/backspace changes the
+          // editable element. If it's already empty, no input event fires;
+          // leaving the listener armed would let the next input event steal
+          // focus back to the container (#6300).
+          //
+          // Remove the listener on keyup, before the next key can produce
+          // input. The timer is only a fallback for cases where keyup never
+          // reaches the container, such as when focus leaves while the key is
+          // held; relying on the timer alone can race the next input event.
           const cleanup = () => {
             container.removeEventListener("input", onInput);
             container.removeEventListener("keyup", cleanup);
--- a/.changeset/300-toolbar-container-empty-delete.md
+++ b/.changeset/300-toolbar-container-empty-delete.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container) so pressing <kbd>Backspace</kbd> or <kbd>Delete</kbd> on a focused container with an empty text field no longer steals focus from the field on the next typed character.
--- a/app/src/sandbox/toolbar-container-6300/index.react.tsx
+++ b/app/src/sandbox/toolbar-container-6300/index.react.tsx
@@ -1,4 +1,39 @@
 import * as Ariakit from "@ariakit/react";
+import * as React from "react";
+
+interface FontFamilyContainerProps {
+  label: string;
+  defaultValue?: string;
+}
+
+function FontFamilyContainer({
+  label,
+  defaultValue,
+}: FontFamilyContainerProps) {
+  const inputRef = React.useRef<HTMLInputElement>(null);
+
+  return (
+    <Ariakit.ToolbarContainer
+      aria-label={label}
+      role="group"
+      onKeyDown={(event) => {
+        if (event.key !== "Backspace" && event.key !== "Delete") return;
+        if (event.target !== event.currentTarget) return;
+        const input = inputRef.current;
+        if (!input) return;
+        if (input.value) return;
+        // TODO: Remove this workaround after
+        // https://github.com/ariakit/ariakit/issues/6300 is fixed.
+        event.preventDefault();
+        input.focus();
+      }}
+    >
+      <label>
+        {label} <input ref={inputRef} defaultValue={defaultValue} />
+      </label>
+    </Ariakit.ToolbarContainer>
+  );
+}
 
 export default function Example() {
   return (
@@ -8,11 +43,7 @@ export default function Example() {
         aria-label="Empty text formatting"
         style={{ display: "flex", gap: 8 }}
       >
-        <Ariakit.ToolbarContainer aria-label="Empty font family" role="group">
-          <label>
-            Empty font family <input />
-          </label>
-        </Ariakit.ToolbarContainer>
+        <FontFamilyContainer label="Empty font family" />
         <Ariakit.ToolbarItem>Apply</Ariakit.ToolbarItem>
       </Ariakit.Toolbar>
       <button type="button">Before filled toolbar</button>
@@ -20,11 +51,7 @@ export default function Example() {
         aria-label="Filled text formatting"
         style={{ display: "flex", gap: 8 }}
       >
-        <Ariakit.ToolbarContainer aria-label="Font family" role="group">
-          <label>
-            Font family <input defaultValue="Inter" />
-          </label>
-        </Ariakit.ToolbarContainer>
+        <FontFamilyContainer label="Font family" defaultValue="Inter" />
         <Ariakit.ToolbarItem>Apply</Ariakit.ToolbarItem>
       </Ariakit.Toolbar>
     </>
--- a/app/src/sandbox/toolbar-container-6300/index.react.tsx
+++ b/app/src/sandbox/toolbar-container-6300/index.react.tsx
@@ -1,39 +1,4 @@
 import * as Ariakit from "@ariakit/react";
-import * as React from "react";
-
-interface FontFamilyContainerProps {
-  label: string;
-  defaultValue?: string;
-}
-
-function FontFamilyContainer({
-  label,
-  defaultValue,
-}: FontFamilyContainerProps) {
-  const inputRef = React.useRef<HTMLInputElement>(null);
-
-  return (
-    <Ariakit.ToolbarContainer
-      aria-label={label}
-      role="group"
-      onKeyDown={(event) => {
-        if (event.key !== "Backspace" && event.key !== "Delete") return;
-        if (event.target !== event.currentTarget) return;
-        const input = inputRef.current;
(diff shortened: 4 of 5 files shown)


app/src/sandbox/toolbar-container-6300/test-chrome.ts
```diff
@@ -0,0 +1,28 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6300
```
diegohaz, line 3: Why do we have a separate `test-chrome.ts` test? What is it testing that `test-browser.ts` can't?

packages/ariakit-react-components/src/composite/composite-container.tsx
```diff
@@ -182,8 +182,17 @@ export const useCompositeContainer = createHook<
...
+            container.removeEventListener("input", onInput);
+            container.removeEventListener("keyup", cleanup);
+          };
+          const onInput = () => {
+            cleanup();
+            queueMicrotask(() => container.focus());
+          };
           container.addEventListener("input", onInput, { once: true });
+          container.addEventListener("keyup", cleanup, { once: true });
+          setTimeout(cleanup, 0);
```
diegohaz, line 202: Worth a short rationale comment here: the dual cleanup (`keyup` listener **and** `setTimeout(cleanup, 0)`) is load-bearing in a way that's easy to miss, so the next person to read this will likely try to collapse it back into the single one-line `setTimeout` reaper that #6300 proposed — which silently reintroduces the bug.

I verified this locally. With only the `setTimeout(0)` reaper (the issue's exact one-liner), the new `test-chrome.ts` filled-field test is flaky on Chromium — it reproduced the focus steal in ~2/15 stress runs (`--repeat-each=5 --retries=0`), while the committed `keyup`+`setTimeout` version passed 15/15. Instrumented traces show why: on the empty second Backspace, the next `a` keystroke's `input` event sometimes fires **before** the `setTimeout(0)` reaper runs (`keydown arm → type a → onInput FIRED → reaper`), so the still-armed listener steals focus. The `keyup` handler avoids the race because it fires synchronously with the key-release of the same gesture, reaping the listener before the next key can be pressed; the timer is only the fallback for when `keyup` never arrives (e.g. focus leaves while the key is held).

None of that is derivable from the current comment (which describes only the happy path), so a suggestion:

```suggestion
          open();
          // Move focus back to the container as soon as the
          // delete/backspace key is captured by the text field: the
          // `input` event from that deletion is the legitimate trigger.
          //
          // When the field is already empty no `input` event fires, so a
          // one-shot listener would stay armed and hijack the next
          // keystroke (#6300). Reap it on `keyup` — the end of the same
          // key gesture — so it's gone before the next key. The
          // `setTimeout` is only a fallback for when `keyup` never
          // arrives (e.g. focus leaves while the key is held); a bare
          // `setTimeout(0)` reaper alone races the next `input` event and
          // intermittently loses.
          const cleanup = () => {
            container.removeEventListener("input", onInput);
            container.removeEventListener("keyup", cleanup);
          };
          const onInput = () => {
            cleanup();
            queueMicrotask(() => container.focus());
          };
          container.addEventListener("input", onInput, { once: true });
          container.addEventListener("keyup", cleanup, { once: true });
          setTimeout(cleanup, 0);
```

(For what it's worth, I also weighed the alternative of just not arming the listener when the field is already empty — like the workaround's `if (input.value) return` — but the listener-lifetime approach handles `contentEditable` uniformly without value/emptiness introspection, so the PR's direction looks like the better trade-off. The only thing missing is the rationale.)

app/src/sandbox/toolbar-container-6300/test-chrome.ts
diegohaz: It's covering the **filled-field** sequence, which genuinely can't run in `test-browser.ts` because Firefox doesn't reproduce its first step.

I checked empirically: I temporarily copied this filled flow into a file matching the cross-browser `*-browser` matcher and ran it on all three engines. It passes on Chromium and WebKit, but **fails on Firefox** at the very first assertion — `await test.expect(input).toHaveValue("")` receives `"Inter"`:

```
locator resolved to <input value="Inter"/>
 - unexpected value "Inter"
```

So in Firefox the Backspace on the focused container never clears `Inter`. The clear-on-Backspace path relies on the keydown's default action landing on the field that `open()` focuses inside a `queueMicrotask`; Firefox doesn't apply that deletion to the programmatically-focused field, so nothing is deleted and the rest of the sequence can't run. That's pre-existing container behavior, unrelated to this PR's listener fix — `test-browser.ts` avoids it by starting from an already-empty field, which reproduces and is fixed on every engine.

Two options if you want:
- Keep it Chrome-only, since Chromium is the browser the issue was reported/verified in (current state — fine).
- Since it also passes on WebKit, you could rename/duplicate it as `test-safari.ts` to add Safari coverage for the filled path. Firefox would still have to be excluded.

diegohaz (commented):
**Review — initial, comment-only.** Reviewed the full diff against merge-base `953dfa92` up to head `2d93cb3`, plus the surrounding `CompositeContainer` keydown/focus logic, `open()`/`selectTextField`, the sandbox, and the two browser tests.

**What I validated locally** (worktree on the PR head, astro dev server, Playwright):
- The committed fix is stable: `toolbar-container-6300` passed 15/15 on Chromium under `--repeat-each=5 --retries=0`.
- I stress-tested the simpler alternative from #6300 (single `setTimeout(0)` reaper, no `keyup`). It's **flaky** — ~2/15 runs reproduce the focus steal in the filled-field test. Instrumented traces confirm the `setTimeout(0)` reaper races the next keystroke's `input` event, whereas the `keyup` cleanup closes the window deterministically. So the PR's extra complexity is justified; I'm **not** recommending simplifying it.
- I ran the filled-field flow under the cross-browser matcher: passes on Chromium + WebKit, **fails on Firefox** because Backspace on the focused container doesn't clear `Inter` there. This explains why the filled test is browser-specific.

**Findings**
1. `composite-container.tsx` — the dual cleanup (`keyup` + `setTimeout` fallback) is correct but its rationale isn't captured in the comment, which invites a future regression-introducing "simplification" back to the #6300 one-liner. Left an inline suggestion with a rationale comment (and the flakiness evidence). Maintainability, not a defect.
2. Replied on your open `test-chrome.ts` thread with the empirical cross-browser breakdown (Chromium/WebKit pass, Firefox can't run the filled path) and the Chrome-only vs. add-Safari options.

**No correctness, regression, compatibility, security, or perf concerns.** The changeset, sandbox (realistic empty + filled toolbars), and the empty-field cross-browser test all look good; CI is green. I did not re-run the full Firefox/WebKit suites beyond the targeted probes above.

packages/ariakit-react-components/src/composite/composite-container.tsx
diegohaz: Fixed in 53f6708c41e4b46181873ee0f2b87d1164c234fc - expanded the cleanup rationale to explain the empty editable case, why the listener is removed on keyup, and why the timer is only a fallback that can race if used alone.

diegohaz:
**Fixed**
- 53f6708c41e4b46181873ee0f2b87d1164c234fc ([reply](https://github.com/ariakit/ariakit/pull/6440#discussion_r3469667979)): Documented why the delete/backspace input listener is removed on keyup, why setTimeout is only a fallback, and how a timer-only cleanup can race the next input event.

PR #6442, opened 2026-06-24
Fix scroll container detection for elements inside iframes

### LINKED ISSUE #6312: PageUp/PageDown in a composite inside an iframe pages by the outer page's viewport, not the iframe's scroller
### Current behavior

When a composite widget is rendered inside a same-origin iframe (embedded widgets, sandboxed preview panes, react-frame-component-style hosting) and the iframe's own document is the scroll container (no `overflow: auto/scroll` ancestor inside the frame), pressing <kbd>PageDown</kbd>/<kbd>PageUp</kbd> on a [`CompositeItem`](https://ariakit.com/reference/composite-item) moves focus by a page computed from the **top-level page's viewport** instead of the iframe's. In a 720px-tall window with a 220px-tall iframe, PageDown from Item 1 (50px items) jumps to Item 21 (~1020px page) instead of about one iframe page down (Item 7, ~289px).

The root cause is in `getScrollingElement` (`@ariakit/utils`): when the ancestor walk finds no overflow container, it falls back to the **global** `document.scrollingElement` rather than the element's own document ([dom.ts#L309-L330](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L309-L330)). For an element inside an iframe, the walk reaches the iframe's `<html>` (computed `overflow: visible`, so the check fails), the recursion bottoms out at `null`, and the outer page's scrolling element is returned. The iframe's own `scrollingElement` is never considered, even though the file already has `getDocument` for exactly this ([dom.ts#L21-L28](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L21-L28)) and other helpers there are frame-aware (e.g. `getActiveElement`).

No caller normalizes the bad result, so several behaviors degrade in the iframe scenario:

- Composite PageUp/PageDown computes the page size and offset from the wrong scroller ([composite-item.tsx#L86-L97](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-item.tsx#L86-L97)) — the repro below.
- [`Combobox`](https://ariakit.com/reference/combobox) attaches its "user scrolled" guards (`wheel`/`touchmove`/`scroll`) to the outer page's `<html>`, which never receives the iframe's scroll events, silently disabling that behavior ([combobox.tsx#L312-L350](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/combobox/combobox.tsx#L312-L350)).
- `isPartiallyHidden`/`scrollIntoViewIfNeeded` compare the item's iframe-relative rect against the outer scroller's rect, mixing coordinate spaces ([dom.ts#L335-L357](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L335-L357)).
- The experimental `CollectionRenderer` resolves its virtualization viewport from the wrong scroller's `ownerDocument` ([collection-renderer.tsx#L290-L306](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L290-L306)).

3. Click **Item 1** inside the iframe.
4. Press <kbd>PageDown</kbd>.

Focus jumps far past one iframe page — the landing item is the one closest to `Math.max(h * 0.875, h - 40) * 1.5` pixels, where `h` is the **outer** window's viewport height (Item 21 in a 720px-tall window) — instead of landing about one iframe viewport down (Item 7).

```tsx

function EmbeddedList() {
  const composite = Ariakit.useCompositeStore({ orientation: "vertical" });
  return (
    <Ariakit.Composite
      store={composite}
      role="toolbar"
      aria-orientation="vertical"
      aria-label="Embedded items"
    >
      {Array.from({ length: 100 }, (_, index) => (
        <Ariakit.CompositeItem
          key={index}
          style={{
            boxSizing: "border-box",
            display: "block",
            height: 50,
            margin: 0,
            width: "100%",
          }}
        >
          Item {index + 1}
        </Ariakit.CompositeItem>
      ))}
    </Ariakit.Composite>
  );
}

  const [iframe, setIframe] = useState<HTMLIFrameElement | null>(null);

  useEffect(() => {
    if (!iframe) return;
    // Same-origin embedded widget (react-frame-component style): the list is
    // taller than the iframe viewport and there's no overflow container
    // inside, so the iframe's own document is the scroller.
    if (!doc?.body) return;
    doc.body.style.margin = "0";
    const root = createRoot(doc.body);
    root.render(<EmbeddedList />);
    return () => {
      setTimeout(() => root.unmount());
    };
  }, [iframe]);

  return (
    <iframe
      ref={setIframe}
      title="Embedded list"
      style={{ border: "1px solid", display: "block", height: 220, width: 320 }}
    />
  );
}
```

This was verified with a real-browser Playwright test (click + PageDown only): the run asserting Item 7 fails on current code, while a probe asserting Item 21 passes.

### Expected behavior

PageUp/PageDown should page relative to the element's actual scroll container — here the iframe's own document. With a 220px-tall iframe viewport and 50px items, PageDown from Item 1 should land on Item 7 (the item closest to the ~289px page offset), matching the behavior when the same list is rendered with an in-frame overflow container.

### Workaround

Give the composite (or any ancestor inside the iframe document) a bounded height and `overflow-y: auto`, so `getScrollingElement` finds it before reaching the fallback:

```tsx
<Ariakit.Composite
  store={composite}
  role="toolbar"
  aria-orientation="vertical"
  aria-label="Embedded items"
  style={{ height: "100vh", overflowY: "auto" }}
>
```

With this change, PageDown correctly lands one iframe page down (verified: the same Playwright test passes).

### Possible solutions

Resolve the fallback against the element's own document with `getDocument`, which is already defined in the same file and is the established idiom for the other element-relative helpers there. Since `parentElement` never crosses frame boundaries, the recursion stays within one document, so this precisely targets the fallback:

```diff
   } else if (element.clientWidth && element.scrollWidth > element.clientWidth) {
     const { overflowX } = getComputedStyle(element);
     if (isScrollableOverflow(overflowX)) return element;
   }
+  const doc = getDocument(element);
   return (
     getScrollingElement(element.parentElement) ||
-    document.scrollingElement ||
-    document.body
+    doc.scrollingElement ||
+    doc.body
   );
 }
```

For elements in the top-level document, `getDocument` returns the same global `document`, so mainline behavior is unchanged. All consumers (composite paging, combobox scroll guards, `isPartiallyHidden`, `CollectionRenderer`) improve without modification.

diegohaz:
## Motivation

When a composite widget is rendered inside a same-origin iframe (embedded widgets, sandboxed preview panes, `react-frame-component`-style hosting) and the iframe's own document is the scroll container — there's no `overflow: auto/scroll` ancestor inside the frame — pressing <kbd>PageDown</kbd>/<kbd>PageUp</kbd> on a [`CompositeItem`](https://ariakit.com/reference/composite-item) moves focus by a page computed from the **top-level page's viewport** instead of the iframe's. In a 720px-tall window with a 220px-tall iframe, PageDown from Item 1 (50px items) jumps far past one iframe page instead of landing about one iframe page down.

The root cause is in `getScrollingElement` (`@ariakit/utils`): when the ancestor walk finds no overflow container, it falls back to the **global** `document.scrollingElement` rather than the element's own document. For an element inside an iframe, the walk reaches the iframe's `<html>` (computed `overflow: visible`, so the check fails), the recursion bottoms out at `null`, and the outer page's scrolling element is returned — even though the file already has `getDocument` for exactly this, and other helpers there are already frame-aware (e.g. `getActiveElement`).

No caller normalizes the bad result, so several behaviors degrade in the iframe scenario:

- Composite PageUp/PageDown computes the page size and offset from the wrong scroller.
- `Combobox` attaches its "user scrolled" guards (`wheel`/`touchmove`/`scroll`) to the outer page's `<html>`, which never receives the iframe's scroll events.
- `isPartiallyHidden`/`scrollIntoViewIfNeeded` compare the item's iframe-relative rect against the outer scroller's rect, mixing coordinate spaces.
- The experimental `CollectionRenderer` resolves its virtualization viewport from the wrong scroller's `ownerDocument`.

## Solution

Resolve the document fallback in `getScrollingElement` against the element's own document with `getDocument` (already defined in the same file and the established idiom for the other element-relative helpers there). Because `parentElement` never crosses frame boundaries, the recursion always stays within one document, so this precisely targets the fallback:

```ts
const doc = getDocument(element);
return (
  getScrollingElement(element.parentElement) ||
  doc.scrollingElement ||
  doc.body
);
```

For elements in the top-level document, `getDocument` returns the same global `document`, so mainline behavior is unchanged. All consumers (composite paging, combobox scroll guards, `isPartiallyHidden`, `CollectionRenderer`) improve without modification.

## Workaround

Until this fix is released, give the composite — or any ancestor inside the iframe document — a bounded height and `overflow-y: auto`, so `getScrollingElement` finds an in-frame scroller before reaching the document fallback:

```tsx
<Ariakit.Composite
  store={composite}
  // TODO: Remove once https://github.com/ariakit/ariakit/issues/6312 is fixed.
  style={{ height: "100vh", overflowY: "auto" }}
>
  {/* ... */}
</Ariakit.Composite>
```

## Tests

- `app/src/sandbox/composite-iframe-6312/` — a Composite rendered into a same-origin iframe whose own document is the scroller. The browser test clicks Item 1, presses PageDown, and asserts focus lands on Item 7 (one iframe page down for a 220px viewport with 50px items). It fails before the fix because paging used the outer page's much taller viewport.
- `packages/ariakit-utils/src/dom.test.ts` — a happy-dom unit test asserting `getScrollingElement` for an element inside an iframe resolves to the iframe's own document, not the global one.

Fixes #6312

### DIFF
--- a/.changeset/300-scrolling-element-iframe.md
+++ b/.changeset/300-scrolling-element-iframe.md
@@ -2,4 +2,4 @@
 "@ariakit/utils": patch
 ---
 
-Fixed `getScrollingElement` to resolve the scroll container from the element's own document instead of the top-level page, correcting [`Composite`](https://ariakit.com/reference/composite) keyboard paging and [`Combobox`](https://ariakit.com/reference/combobox) scroll behavior for elements rendered inside a same-origin iframe.
+Fixed `getScrollingElement` to resolve the scroll container from the element's own document instead of the top-level page, so scroll-aware behavior works correctly for elements rendered inside a same-origin iframe.
--- a/.changeset/300-scrolling-element-iframe.md
+++ b/.changeset/300-scrolling-element-iframe.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Fixed `getScrollingElement` to resolve the scroll container from the element's own document instead of the top-level page, correcting [`Composite`](https://ariakit.com/reference/composite) keyboard paging and [`Combobox`](https://ariakit.com/reference/combobox) scroll behavior for elements rendered inside a same-origin iframe.
--- a/.changeset/300-composite-iframe-paging.md
+++ b/.changeset/300-composite-iframe-paging.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Composite`](https://ariakit.com/reference/composite) keyboard paging and [`Combobox`](https://ariakit.com/reference/combobox) scroll behavior for elements rendered inside a same-origin iframe.
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -10,6 +10,11 @@ function EmbeddedList() {
       role="toolbar"
       aria-orientation="vertical"
       aria-label="Embedded items"
+      // Workaround for https://github.com/ariakit/ariakit/issues/6312: give an
+      // ancestor inside the iframe a bounded height and `overflow-y: auto` so
+      // `getScrollingElement` finds an in-frame scroller before falling back to
+      // the outer page's document.
+      style={{ height: "100vh", overflowY: "auto" }}
     >
       {Array.from({ length: 100 }, (_, index) => (
         <Ariakit.CompositeItem
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -1,0 +1,57 @@
+import * as Ariakit from "@ariakit/react";
+import { useEffect, useState } from "react";
+import { createRoot } from "react-dom/client";
+
+function EmbeddedList() {
+  const composite = Ariakit.useCompositeStore({ orientation: "vertical" });
+  return (
+    <Ariakit.Composite
+      store={composite}
+      role="toolbar"
+      aria-orientation="vertical"
+      aria-label="Embedded items"
+    >
+      {Array.from({ length: 100 }, (_, index) => (
+        <Ariakit.CompositeItem
+          key={index}
+          style={{
+            boxSizing: "border-box",
+            display: "block",
+            height: 50,
+            margin: 0,
+            width: "100%",
+          }}
+        >
+          Item {index + 1}
+        </Ariakit.CompositeItem>
+      ))}
+    </Ariakit.Composite>
+  );
+}
+
+export default function Example() {
+  const [iframe, setIframe] = useState<HTMLIFrameElement | null>(null);
+
+  useEffect(() => {
+    if (!iframe) return;
+    // Same-origin embedded widget (react-frame-component style): the list is
+    // taller than the iframe viewport and there's no overflow container inside,
+    // so the iframe's own document is the scroller.
+    const doc = iframe.contentDocument;
+    if (!doc?.body) return;
+    doc.body.style.margin = "0";
+    const root = createRoot(doc.body);
+    root.render(<EmbeddedList />);
+    return () => {
+      setTimeout(() => root.unmount());
+    };
+  }, [iframe]);
+
+  return (
+    <iframe
+      ref={setIframe}
+      title="Embedded list"
+      style={{ border: "1px solid", display: "block", height: 220, width: 320 }}
+    />
+  );
+}
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -10,11 +10,6 @@ function EmbeddedList() {
       role="toolbar"
       aria-orientation="vertical"
       aria-label="Embedded items"
-      // Workaround for https://github.com/ariakit/ariakit/issues/6312: give an
-      // ancestor inside the iframe a bounded height and `overflow-y: auto` so
-      // `getScrollingElement` finds an in-frame scroller before falling back to
-      // the outer page's document.
-      style={{ height: "100vh", overflowY: "auto" }}
     >
       {Array.from({ length: 100 }, (_, index) => (
         <Ariakit.CompositeItem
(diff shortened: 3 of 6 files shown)


.changeset/300-scrolling-element-iframe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Fixed `getScrollingElement` to resolve the scroll container from the element's own document instead of the top-level page, correcting [`Composite`](https://ariakit.com/reference/composite) keyboard paging and [`Combobox`](https://ariakit.com/reference/combobox) scroll behavior for elements rendered inside a same-origin iframe.
```
diegohaz: This is fine, but I believe we should have a specific user-facing changeset for `@ariakit/react-components` and `@ariakit/react` packages.

diegohaz: Fixed in b23ef8d58 — added a dedicated user-facing changeset for `@ariakit/react-components` and `@ariakit/react` describing the Composite paging and Combobox scroll fix, and trimmed the `@ariakit/utils` changeset to a util-focused entry for `getScrollingElement` so the two changelogs do not duplicate.

diegohaz:
**Fixed**

- `b23ef8d58` ([reply](https://github.com/ariakit/ariakit/pull/6442#discussion_r3470408645)): Added a dedicated user-facing changeset for `@ariakit/react-components` and `@ariakit/react` describing the Composite paging and Combobox scroll fix, and trimmed the `@ariakit/utils` changeset to a util-focused `getScrollingElement` entry so the changelogs don't duplicate.

PR #6444, opened 2026-06-24
Fix `form.names.*` crash on symbol access

### LINKED ISSUE #6308: Form: rendering form.names.* as a React child crashes with "Cannot convert a Symbol value to a string"
### Current behavior

The form store's `names` values are documented as string-like field paths (`store.names.email; // "email"`). But putting one in a React child position — e.g. `<code>{form.names.email}</code>` to display which field a value is submitted as — crashes the whole tree with:

```
TypeError: Cannot convert a Symbol value to a string
```

The stack points inside Ariakit's proxy handler, not at the user code. With an error boundary the entire form is torn down and replaced by the fallback; without one, React unmounts the app.

The root cause is in [`getNameHandler`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L128-L148), which backs the `names` proxy created by [`createNames`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L161-L164). The `get` trap special-cases only `"toString"`, `"valueOf"`, and `Symbol.toPrimitive`, then unconditionally runs `nextKeys.join(".")`. `Array.prototype.join` string-converts each element, which throws for symbols. So *any* other symbol-keyed access crashes instead of returning `undefined` like a plain object would.

React triggers exactly that: when reconciling an object child, react-dom's `getIteratorFn` reads `child[Symbol.iterator]` (in both development and production builds) to decide whether the child is iterable. The proxy throws right there — before React can even raise its own actionable "Objects are not valid as a React child" error. The same crash hits `Object.prototype.toString.call(name)` (probes `Symbol.toStringTag`), `Array.from(name)`, spread, and test pretty-printers, even though the documented contract for names is just [`StringLike`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/types.ts#L3-L9) (`toString`/`valueOf`), so nothing in the public API depends on symbol access.

3. Click the **Show field name** button.
4. The whole form disappears and the error boundary fallback shows `Render error: Cannot convert a Symbol value to a string` (without the boundary, React unmounts the app with the same uncaught TypeError).

```tsx
import { Component, useState } from "react";
import type { ReactNode } from "react";

class ErrorBoundary extends Component<
  { children: ReactNode },
  { error: Error | null }
> {
  state: { error: Error | null } = { error: null };

  static getDerivedStateFromError(error: Error) {
    return { error };
  }

  render() {
    if (this.state.error) {
      return <div role="alert">Render error: {this.state.error.message}</div>;
    }
    return this.props.children;
  }
}

function SignupForm() {
  const form = Ariakit.useFormStore({ defaultValues: { email: "" } });
  const [fieldNameVisible, setFieldNameVisible] = useState(false);

  return (
    <Ariakit.Form store={form}>
      <Ariakit.FormLabel name={form.names.email}>Email</Ariakit.FormLabel>
      <Ariakit.FormInput type="email" name={form.names.email} required />
      <Ariakit.FormError name={form.names.email} />
      <button type="button" onClick={() => setFieldNameVisible(true)}>
        Show field name
      </button>
      {fieldNameVisible && (
        <p>
          This value is submitted as <code>{form.names.email}</code>
        </p>
      )}
      <Ariakit.FormSubmit>Sign up</Ariakit.FormSubmit>
    </Ariakit.Form>
  );
}

  return (
    <ErrorBoundary>
      <SignupForm />
    </ErrorBoundary>
  );
}
```

This was verified with a real-browser Playwright test (page load and a single click only): the field path never renders and the page snapshot reduces to `alert: "Render error: Cannot convert a Symbol value to a string"`.

### Expected behavior

Accessing an absent symbol key on a [`useFormStore`](https://ariakit.com/reference/use-form-store) `names` object should return `undefined`, matching plain-object semantics — the `StringLike` contract only promises `toString`/`valueOf`. Then user code that inspects or iterates a name degrades gracefully (`Object.prototype.toString` returns `"[object Object]"`, `Array.from` no longer throws mid-probe), and rendering a raw name as a child surfaces React's standard "Objects are not valid as a React child" message, which points at the actual fix: coercing with `` `${form.names.email}` `` to display the path.

### Workaround

Coerce the name to a string explicitly before it leaves Ariakit props, e.g.:

```tsx
<code>{`${form.names.email}`}</code>
// or
<code>{String(form.names.email)}</code>
```

Template literals and `String()` go through the supported `Symbol.toPrimitive` path and render `email` correctly. Verified: the same Playwright test passes with this change.

### Possible solutions

Handle symbols explicitly at the top of the `get` trap in `getNameHandler`: keep `Symbol.toPrimitive` returning the path-joining function, and return `undefined` for every other symbol so they never reach `nextKeys.join(".")`. `prevKeys` can then be narrowed to `string[]`, and the per-access `["toString", "valueOf", Symbol.toPrimitive].includes(key)` array allocation goes away:

```diff
-function getNameHandler(
-  cache: FormStoreValues,
-  prevKeys: Array<string | symbol> = [],
-) {
+function getNameHandler(cache: FormStoreValues, prevKeys: string[] = []) {
   const handler: ProxyHandler<FormStoreValues> = {
     get(target, key) {
-      if (["toString", "valueOf", Symbol.toPrimitive].includes(key)) {
+      if (typeof key === "symbol") {
+        // Support string coercion, but ignore all other symbol protocols
+        // (Symbol.iterator, Symbol.toStringTag, etc.) instead of throwing
+        // when joining the key path below.
+        if (key === Symbol.toPrimitive) {
+          return () => prevKeys.join(".");
+        }
+        return undefined;
+      }
+      if (key === "toString" || key === "valueOf") {
         return () => prevKeys.join(".");
       }
       const nextKeys = [...prevKeys, key];
```

There is no proxy invariant risk since the target is an empty null-prototype object, and all documented coercion paths (template literals, `String()`, concatenation, `toString()`, `valueOf()`) keep working, as does proxy cache identity (`names.a === names.a`).

diegohaz:
## Motivation

The `names` object returned by [`useFormStore`](https://ariakit.com/reference/use-form-store) is a `Proxy` whose values are documented string-like field paths (`form.names.email` stringifies to `"email"`). Its `get` trap special-cases only `toString`, `valueOf`, and `Symbol.toPrimitive`, then unconditionally runs `nextKeys.join(".")`. `Array.prototype.join` string-converts every element, which throws `TypeError: Cannot convert a Symbol value to a string` for any other symbol key — so reading an absent symbol crashes instead of returning `undefined` like a plain object would.

React triggers exactly this: when reconciling an object child it probes `child[Symbol.iterator]` to decide whether the child is iterable. So rendering a raw name as a React child crashes the whole tree before React can raise its own actionable "Objects are not valid as a React child" error:

```tsx
// Crashes the tree with "Cannot convert a Symbol value to a string"
<code>{form.names.email}</code>
```

The same crash hits `Object.prototype.toString.call(name)` (probes `Symbol.toStringTag`) and `Array.from(name)`, even though the documented `StringLike` contract only promises `toString`/`valueOf`.

## Solution

Handle symbols explicitly at the top of the `get` trap in `getNameHandler`: keep `Symbol.toPrimitive` returning the path-joining function, and return `undefined` for every other symbol so they never reach `nextKeys.join(".")`. This matches plain-object semantics — an absent symbol key resolves to `undefined` — so user code that inspects or iterates a name degrades gracefully, and rendering a raw name surfaces React's standard "Objects are not valid as a React child" error (which points at coercing with `` `${form.names.email}` ``). All documented coercion paths (template literals, `String()`, concatenation, `toString()`, `valueOf()`) keep working, as does proxy cache identity (`names.a === names.a`).

## Workaround

Coerce the name to a string before it reaches any operation that reads a symbol on it (rendering it as a React child, `Object.prototype.toString.call`, `Array.from`, …). Template literals and `String()` go through the supported `Symbol.toPrimitive` path:

```tsx
// TODO: remove once https://github.com/ariakit/ariakit/issues/6308 is fixed.
<code>{`${form.names.email}`}</code>
// or
<code>{String(form.names.email)}</code>
```

Closes #6308.

### DIFF
--- a/.changeset/300-form-names-symbol.md
+++ b/.changeset/300-form-names-symbol.md
@@ -4,4 +4,12 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`useFormStore`](https://ariakit.com/reference/use-form-store) [`names`](https://ariakit.com/reference/use-form-store#names) values throwing `Cannot convert a Symbol value to a string` when a symbol key was read from them, such as when React probes a name rendered as a React child for `Symbol.iterator`. Absent symbol keys now resolve to `undefined` like a plain object, so string coercion such as `` `${form.names.email}` `` keeps working while raw access degrades gracefully.
+`form.names.*` paths no longer crash on symbol access
+
+Fixed [`useFormStore`](https://ariakit.com/reference/use-form-store) [`names`](https://ariakit.com/reference/use-form-store#names) values throwing `Cannot convert a Symbol value to a string` when an absent symbol key was read from them. This happened whenever something probed a symbol on a raw name — most notably when React reads `Symbol.iterator` to reconcile a name rendered as a React child, but also `Object.prototype.toString.call(name)` and `Array.from(name)`.
+
+Absent symbol keys now resolve to `undefined`, matching plain-object semantics, so those probes degrade gracefully. The documented string coercion keeps working, so coerce a name before rendering or inspecting it outside Ariakit props:
+
+```tsx
+<p>This field submits as {`${form.names.email}`}.</p>
+```
--- a/.changeset/300-form-names-symbol.md
+++ b/.changeset/300-form-names-symbol.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`useFormStore`](https://ariakit.com/reference/use-form-store) [`names`](https://ariakit.com/reference/use-form-store#names) values throwing `Cannot convert a Symbol value to a string` when a symbol key was read from them, such as when React probes a name rendered as a React child for `Symbol.iterator`. Absent symbol keys now resolve to `undefined` like a plain object, so string coercion such as `` `${form.names.email}` `` keeps working while raw access degrades gracefully.
--- a/app/src/sandbox/form-names-6308/index.react.tsx
+++ b/app/src/sandbox/form-names-6308/index.react.tsx
@@ -2,18 +2,25 @@ import * as ak from "@ariakit/react";
 import { Component, useState } from "react";
 import type { ReactNode } from "react";
 
-// Workaround for https://github.com/ariakit/ariakit/issues/6308
+// Reproduces https://github.com/ariakit/ariakit/issues/6308
 //
-// `form.names.*` values are documented string-like field paths, but the proxy
-// throws "Cannot convert a Symbol value to a string" when an absent *symbol*
-// key is read — which happens when react-dom probes a rendered child for
-// `Symbol.iterator`, or when `Object.prototype.toString.call(...)` probes
-// `Symbol.toStringTag`. Coercing the name to a string with `String(...)` before
-// it reaches either operation avoids the crash until the library fix lands.
+// `form.names.*` values are documented string-like field paths. Accessing an
+// absent *symbol* key on one must return `undefined` like a plain object would,
+// instead of throwing "Cannot convert a Symbol value to a string" from the
+// names proxy. Two ordinary things an app does hit that absent-symbol access:
 //
-// The error boundary wraps only the rendered name; with the workaround applied
-// it stays inert, but it is what keeps the rest of the paragraph mounted (and
-// the crash localized) without the coercion.
+// - Rendering a raw name as a React child ("Show field name"): react-dom probes
+//   the child for `Symbol.iterator` to decide whether it is iterable. The proxy
+//   used to throw there, tearing down the surrounding tree, instead of letting
+//   React raise its own "Objects are not valid as a React child" error.
+// - Inspecting a name with `Object.prototype.toString.call(...)` ("Inspect field
+//   name"): it probes `Symbol.toStringTag` and should resolve to "[object
+//   Object]" rather than throwing.
+//
+// The error boundary wraps only the rendered name so the rest of the paragraph
+// stays mounted: the bug puts the Symbol coercion crash in its place, while both
+// the userland workaround (coercing the name) and the library fix keep that
+// crash away.
 class NameBoundary extends Component<
   { children: ReactNode },
   { error: Error | null }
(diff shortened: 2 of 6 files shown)


.changeset/300-form-names-symbol.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`useFormStore`](https://ariakit.com/reference/use-form-store) [`names`](https://ariakit.com/reference/use-form-store#names) values throwing `Cannot convert a Symbol value to a string` when a symbol key was read from them, such as when React probes a name rendered as a React child for `Symbol.iterator`. Absent symbol keys now resolve to `undefined` like a plain object, so string coercion such as `` `${form.names.email}` `` keeps working while raw access degrades gracefully.
```
diegohaz: I think this could benefit from a multiline changeset with a heading

diegohaz: Fixed in be47650f9 — rewrote it as a multiline entry with a title heading and detail paragraphs. Also dropped an inaccurate `object spreads` example: spreading a name enumerates the proxy target’s (empty) own keys, so it never reads a symbol or throws.

diegohaz:
**Fixed**

- be47650f9 ([reply](https://github.com/ariakit/ariakit/pull/6444#discussion_r3470506861)): Rewrote the `.changeset/300-form-names-symbol.md` entry as a multiline changeset with a title heading and detail paragraphs, as requested. While there, dropped an inaccurate `object spreads` example — object spread enumerates the proxy target’s (empty) own keys, so it never reads a symbol or throws (the genuinely affected probes are React’s `Symbol.iterator` child check, `Object.prototype.toString.call`, and `Array.from`). The same wording was corrected in the PR description.

PR #6445, opened 2026-06-24
Fix form store submit() stalling while the tab is hidden

### LINKED ISSUE #6309: Form store submit()/validate() stall indefinitely while the tab is hidden (unguarded requestAnimationFrame await)
### Current behavior

`useFormStore`'s `submit()` and `validate()` internally await a bare `requestAnimationFrame` promise. Browsers pause `requestAnimationFrame` for hidden documents, so if a submission is in flight when the tab is hidden — or is started while the tab is hidden, e.g. by the common "auto-save the draft on `visibilitychange`" pattern — the promise never resolves until the tab is foregrounded again.

The frame wait happens in `nextFrame()` ([form-store.ts#L28-L30](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L28-L30)):

```ts
function nextFrame() {
  return new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
}
```

`validate()` awaits it after running the validate callbacks ([form-store.ts#L242-L258](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L242-L258)), and `submit()` is gated on `await validate()` and then awaits `nextFrame()` a second time ([form-store.ts#L327-L354](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L327-L354)). Because the stall happens *before* the submit callbacks run, the user's actual save (typically a network request) never fires while the tab is hidden. The form sits with `submitting: true` (and `validating: true`) the whole time, and the submission only goes out when the user happens to come back. If the OS or browser discards the backgrounded tab first, the submission is silently lost.

A secondary consequence of the same line: `@ariakit/components` is the framework-agnostic core package, and `validate()`/`submit()` are public store methods, but `requestAnimationFrame` is not a global in Node. Calling them outside a DOM environment (server-side validation reuse, node-environment tests) throws `ReferenceError: requestAnimationFrame is not defined`. The same package already capability-checks browser APIs elsewhere — e.g. `typeof IntersectionObserver !== "function"` in [collection-store.ts#L109-L111](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L109-L111) — so the form store is internally inconsistent.

1. Go to https://stackblitz.com/github/ariakit/ariakit/tree/main/templates/react?file=src%2Fapp.tsx
3. Click the **Draft** field and type something.
4. Switch to another browser tab, stay there for about 5 seconds, then switch back.
5. Observe the rendered state: the save was stuck at **Saving** the entire time you were away and only flipped to **Saved** after you returned — the rendered "saved at" timestamp is *later* than "visible at", several seconds after "hidden at".

```tsx

  const form = Ariakit.useFormStore({ defaultValues: { draft: "" } });
  const submitting = Ariakit.useStoreState(form, "submitting");
  const submitSucceed = Ariakit.useStoreState(form, "submitSucceed");
  const [hiddenAt, setHiddenAt] = useState(0);
  const [visibleAt, setVisibleAt] = useState(0);
  const [savedAt, setSavedAt] = useState(0);

  form.useSubmit(async () => {
    // Simulate a quick network request that saves the draft.
    await new Promise((resolve) => setTimeout(resolve, 50));
  });

  // Auto-save the draft when the user switches away from the tab.
  useEffect(() => {
    function onVisibilityChange() {
      if (document.hidden) {
        setHiddenAt(Date.now());
        void form.submit();
      } else {
        setVisibleAt(Date.now());
      }
    }
    document.addEventListener("visibilitychange", onVisibilityChange);
    return () => {
      document.removeEventListener("visibilitychange", onVisibilityChange);
    };
  }, [form]);

  // Record when the save completed.
  useEffect(() => {
    if (!submitSucceed) return;
    setSavedAt(Date.now());
  }, [submitSucceed]);

  return (
    <Ariakit.Form store={form} aria-label="Draft">
      <Ariakit.FormLabel name={form.names.draft}>Draft</Ariakit.FormLabel>
      <Ariakit.FormInput name={form.names.draft} />
      <Ariakit.FormSubmit>Save</Ariakit.FormSubmit>
      <p>
        Save state:{" "}
        <output>{submitting ? "Saving" : savedAt ? "Saved" : "Idle"}</output>
      </p>
      <p>
        hidden at <output>{hiddenAt}</output> · visible at{" "}
        <output>{visibleAt}</output> · saved at <output>{savedAt}</output>
      </p>
    </Ariakit.Form>
  );
}
```

This was verified with a Playwright test driving a real headed Chromium through genuine tab switches (attached via `connectOverCDP` with `noDefaults: true`, because Playwright's default focus emulation forces hidden tabs to stay visible and masks the bug): the save stayed at "Saving" for the whole hidden period and completed only after the tab was brought back to the front.

### Expected behavior

The submission should make progress while the tab is hidden. Timers keep running in hidden tabs (they are throttled, not suspended), so the save should complete within roughly a second of the tab being hidden — the rendered "saved at" should land right after "hidden at", before "visible at". `validate()` and `submit()` should not depend on a rendering frame that may never come, and calling them in Node should not throw `ReferenceError`.

### Workaround

Patch `requestAnimationFrame` so it falls back to a timeout while the document is hidden (when the real rAF would not fire anyway). The form store never cancels these frames, so returning a timeout id is safe here:

```ts
if (typeof window !== "undefined") {
  const originalRequestAnimationFrame =
    window.requestAnimationFrame.bind(window);
  window.requestAnimationFrame = (callback) => {
    if (!document.hidden) {
      return originalRequestAnimationFrame(callback);
    }
    return window.setTimeout(() => callback(performance.now()), 16);
  };
}
```

With this in place, the same auto-save completes while the tab is hidden (verified). It is a global patch, though — a proper fix belongs in the store.

### Possible solutions

Race the animation frame against a timeout in `nextFrame()`, and fall back entirely to `setTimeout` when `requestAnimationFrame` is unavailable (mirroring the existing `typeof IntersectionObserver` capability check in the collection store). In a visible document rAF wins within ~16 ms, preserving the current "after the next render" semantics; in a hidden document the timeout bounds the wait; in Node the `typeof` guard removes the `ReferenceError`:

```diff
 function nextFrame() {
-  return new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
+  return new Promise<void>((resolve) => {
+    if (typeof requestAnimationFrame !== "function") {
+      setTimeout(resolve, 0);
+      return;
+    }
+    // rAF doesn't fire in hidden documents, which would stall validate/submit
+    // until the tab becomes visible again. Fall back to a timeout.
+    const timeoutId = setTimeout(resolve, 100);
+    requestAnimationFrame(() => {
+      clearTimeout(timeoutId);
+      resolve();
+    });
+  });
 }
```

The double-resolve when both branches fire is a harmless no-op on an already-settled promise. Public API semantics of `validate()`/`submit()` are unchanged.

diegohaz:
## Motivation

`useFormStore`'s `submit()` and `validate()` await a bare `requestAnimationFrame` in `nextFrame()`:

```ts
function nextFrame() {
  return new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
}
```

Browsers pause `requestAnimationFrame` in hidden documents, so a submission started while the tab is hidden — e.g. the common "auto-save the draft on `visibilitychange`" pattern — never makes progress: the callback that runs the actual save never fires, and the form sits with `submitting: true` until the tab is foregrounded again. If the browser discards the backgrounded tab first, the save is silently lost.

## Solution

Race the animation frame against a timeout in `nextFrame()`:

```ts
function nextFrame() {
  return new Promise<void>((resolve) => {
    const timeoutId = setTimeout(resolve, 100);
    requestAnimationFrame(() => {
      clearTimeout(timeoutId);
      resolve();
    });
  });
}
```

In a visible document `requestAnimationFrame` wins within ~16ms, preserving the current "after the next render" semantics; in a hidden document the timeout bounds the wait so the save still completes. The public semantics of `validate()`/`submit()` are unchanged (the double-resolve when both branches fire is a no-op on an already-settled promise).

## Workaround

Until this is released, patch `requestAnimationFrame` to fall back to a timeout while the document is hidden (when the real one wouldn't fire anyway). The form store never cancels these frames, so returning a timeout id is safe:

```ts
// TODO: Remove once https://github.com/ariakit/ariakit/issues/6309 is released.
if (typeof window !== "undefined") {
  const originalRequestAnimationFrame =
    window.requestAnimationFrame.bind(window);
  window.requestAnimationFrame = (callback) => {
    if (!document.hidden) {
      return originalRequestAnimationFrame(callback);
    }
    return window.setTimeout(() => callback(performance.now()), 16);
  };
}
```

## Reproduction

`app/src/sandbox/form-6309/` models the auto-save-on-hide scenario from the issue, and a Playwright test recreates a backgrounded tab — `document.hidden` plus a `requestAnimationFrame` that stops firing while hidden, since Playwright's focus emulation can't truly background a tab — then dispatches `visibilitychange` and asserts the save reaches "Saved" instead of stalling on "Saving". A happy-dom duplicate is intentionally omitted: `@ariakit/test`'s own `settle`/`click` helpers depend on `requestAnimationFrame`, so pausing it would break the test harness itself.

Closes #6309


.changeset/300-form-store-hidden-tab.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`useFormStore`](https://ariakit.com/reference/use-form-store)'s [`submit`](https://ariakit.com/reference/use-form-store#submit) and [`validate`](https://ariakit.com/reference/use-form-store#validate) stalling while the document is hidden — for example, when auto-saving on `visibilitychange` — because they awaited a `requestAnimationFrame` that browsers pause in background tabs. They also no longer throw a `ReferenceError` when called outside a browser.
```
diegohaz: This could benefit from a multiline changeset with a heading

packages/ariakit-components/src/form/form-store.node.test.ts
```diff
@@ -0,0 +1,25 @@
+// @vitest-environment node
```
diegohaz: `@ariakit/components` is framework-agnostic, but it's meant to run in the browser. Here, framework-agnostic means it should support React, Vue, Solid, and similar frameworks.

The exception is SSR environments, but only for static, top-level logic, like things created when the store is initialized, not things that run during after setup.

packages/ariakit-components/src/form/form-store.ts
```diff
@@ -26,7 +26,24 @@ type ErrorMessage = string | undefined | null;
...
 
 function nextFrame() {
-  return new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
+  return new Promise<void>((resolve) => {
+    // `requestAnimationFrame` isn't a global outside the browser (e.g. Node),
+    // mirroring the `typeof IntersectionObserver` check in the collection store.
+    if (typeof requestAnimationFrame !== "function") {
+      setTimeout(resolve, 0);
+      return;
+    }
```
diegohaz: I don't think this is necessary.

diegohaz: Fixed in cc735c40b — removed the `typeof requestAnimationFrame` guard. Since `submit()`/`validate()` always run in a browser DOM, that branch was dead code.

packages/ariakit-components/src/form/form-store.node.test.ts
diegohaz: Thanks for the clarification — that matches my understanding now. Removed the Node-environment test in cc735c40b (along with the `typeof requestAnimationFrame` guard it covered), since `submit()`/`validate()` run after setup and are only meant to run in a browser DOM, not in Node.

.changeset/300-form-store-hidden-tab.md
diegohaz: Fixed in a8a181620 — rewrote it as a multi-paragraph changeset with a title heading.

diegohaz:
Addressed the review feedback:

**Fixed**

- `cc735c40b` ([reply](https://github.com/ariakit/ariakit/pull/6445#discussion_r3470544056)): Removed the `typeof requestAnimationFrame` guard from `nextFrame()` — `submit()`/`validate()` always run in a browser DOM, so the fallback was dead code.
- `cc735c40b` ([reply](https://github.com/ariakit/ariakit/pull/6445#discussion_r3470544128)): Removed the Node-environment test that covered that now-dropped path, since those methods aren't meant to run outside a DOM.
- `a8a181620` ([reply](https://github.com/ariakit/ariakit/pull/6445#discussion_r3470544207)): Rewrote the changeset as a multi-paragraph entry with a title heading (and dropped the Node `ReferenceError` note).

The fix is now just the hidden-document `requestAnimationFrame`-vs-timeout race. The Playwright repro still passes on Chrome/Firefox/Safari and the core form tests pass.

PR #6448, opened 2026-06-24
Restrict perf baseline reuse to push artifacts

diegohaz:
## Motivation

The perf workflow reuses `app.yml` artifacts for a pull request's base SHA. `app.yml` also runs on `pull_request` events, where the default checkout builds the synthetic merge ref even though the workflow run reports the PR head SHA.

On stacked PRs, that can make the baseline artifact include unrelated main changes instead of being a build of the target base SHA, which corrupts the Chrome perf comparison.

## Solution

Filter the `app.yml` workflow-run lookup to `push` events. Push runs build exactly at `head_sha`, so main-targeted PRs can still use the fast artifact path:

```js
await github.rest.actions.listWorkflowRuns({
  workflow_id: "app.yml",
  branch: branchRef,
  head_sha: baseSha,
  event: "push",
});
```

Stacked PRs without a push run for their base branch now fall back to the existing in-job baseline build from `github.event.pull_request.base.sha`. The fallback notice also says that no successful push `app.yml` run was found, so CI logs reflect the new lookup contract.

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -91,13 +91,17 @@ jobs:
               workflow_id: 'app.yml',
               branch: branchRef,
               head_sha: baseSha,
+              // Only push runs build exactly at head_sha. Pull request runs
+              // check out the merge ref, so their artifacts are not a build
+              // of base.sha and would corrupt stacked PR baselines.
+              event: 'push',
               status: 'success',
               per_page: 1,
             });
             const run = runs.data.workflow_runs[0];
             if (!run) {
               core.notice(
-                `No successful app.yml run found for base SHA ${baseSha}. ` +
+                `No successful push app.yml run found for base SHA ${baseSha}. ` +
                   'Building base artifacts in this job.',
               );
               return;


diegohaz (commented):
**Review type:** Initial review
**Baseline:** merge-base `1c832fb49` → head `e274c7f`
**Scope:** `.github/workflows/perf.yml` plus the surrounding perf/baseline workflow context.

### Verdict

No substantive findings. The change is correct, minimal, and well-documented; I think it's the right fix and the best trade-off available here.

### What I verified

- **Root-cause premise holds.** `app.yml` triggers on both `push: [main]` and `pull_request`, and its `build` / `build-nextjs` jobs produce the `app-build` + `nextjs-dist` artifacts the `chrome-baseline` lookup reuses. `pull_request` runs use the default `actions/checkout` (`refs/pull/N/merge`), so their artifact reflects head-merged-into-base rather than a clean build of the base SHA, even though the run still reports the PR head SHA as `head_sha`. So the corruption the PR describes is real.
- **The fix is complete for the stacked-PR case.** For a stacked PR, `branch: branchRef` resolves to the parent PR's branch, and `app.yml` only ever produced `pull_request` runs there (its push trigger is restricted to `main`). The pre-fix lookup matched the parent's `pull_request` run because its `head_sha` equals the child's `base.sha`, reusing a merge-ref artifact. Adding `event: 'push'` removes that match, so stacked PRs correctly fall through to the existing in-job baseline build from `github.event.pull_request.base.sha`.
- **No regression for normal PRs.** For a main-targeted PR, `branch` resolves to `main`, and no `pull_request` run has `head_branch: main`, so the lookup already only matched push-to-`main` runs in practice. The added `event: 'push'` is a strict no-op / clarification there, and the fast artifact-reuse path is preserved. Push runs build the exact pushed commit, so the reused artifact is a clean build of the base SHA — and it stays consistent with the base-SHA worktree the chrome/node jobs check out.
- **No other call site needs the same change.** `listWorkflowRuns` is used only in `perf.yml`; the `app-build` / `nextjs-dist` references in `deploy-preview.yml` are intra-run downloads within the same `app.yml` run, not cross-run lookups.
- **Supporting changes look right.** The updated fallback notice ("No successful push app.yml run found…") accurately reflects the new lookup contract, and the added comment explains the why (merge-ref artifacts) clearly. No changeset is needed — this is a CI-only workflow change that doesn't affect shipped packages — so its omission is correct.

### Not validated locally

Workflow YAML changes can't be exercised end-to-end locally; the stacked-PR reuse path and the GitHub Actions `branch`/`event`/`head_sha` filter behavior were reasoned through the GitHub Actions semantics and the `app.yml` trigger/checkout configuration rather than executed. The reasoning is consistent, but the only true end-to-end confirmation is a live perf run on a stacked PR.

PR #6450, opened 2026-06-24
Fix JSDoc prop description precedence

diegohaz:
## Motivation

The JSDoc loader builds prop declaration candidates in local-first, nearest-first order so specialized prop documentation can override inherited docs. The description loop kept scanning after it found a match, so a farther ancestor could overwrite the intended local or nearer declaration.

For example, a prop redeclared with local docs could still render the generic ancestor description and live examples in the reference docs.

## Solution

Stop scanning declaration candidates once the first documented declaration is found. This preserves the existing fallback for undocumented redeclarations while making documented local or nearer declarations win as intended.

Added a focused loader regression test that exercises:

- local prop docs overriding documented ancestors
- undocumented local props falling back to the nearest documented parent
- undocumented local and parent props falling back to the farthest documented ancestor

### DIFF
--- a/app/src/lib/jsdoc-loader.test.ts
+++ b/app/src/lib/jsdoc-loader.test.ts
@@ -1,0 +1,180 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+import fs from "node:fs/promises";
+import os from "node:os";
+import { dirname, join } from "node:path";
+import type { LoaderContext } from "astro/loaders";
+import { afterEach, expect, test, vi } from "vitest";
+import { jsdoc } from "./jsdoc-loader.ts";
+import type { Reference } from "./schemas.ts";
+
+const dirs: string[] = [];
+
+afterEach(async () => {
+  for (const dir of dirs.splice(0)) {
+    await fs.rm(dir, { force: true, recursive: true });
+  }
+});
+
+async function createDir() {
+  const dir = await fs.mkdtemp(join(os.tmpdir(), "ariakit-jsdoc-"));
+  dirs.push(dir);
+  return dir;
+}
+
+async function writeFile(file: string, content: string) {
+  await fs.mkdir(dirname(file), { recursive: true });
+  await fs.writeFile(file, content);
+}
+
+function getLoaderContext() {
+  const entries = new Map<string, { id: string; data: Reference }>();
+  const context = {
+    logger: {
+      error: vi.fn(),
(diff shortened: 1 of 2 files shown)


diegohaz (commented):
**Review summary — looks good, no blocking findings.**

Initial review. Baseline: merge-base `1c832fb` → head `965f1e5`. Scope: `app/src/lib/jsdoc-loader.ts` (the `break`) and the new `app/src/lib/jsdoc-loader.test.ts`, plus the surrounding `getProps` precedence logic, the base-hierarchy ordering, and the rendered-reference impact across the real packages.

### The fix is correct

`decls` is built local-first / nearest-first (`[localDecl, ...findPropDeclsInBaseHierarchy()]`, and the BFS in `findPropDeclsInBaseHierarchy` is documented and confirmed nearest-first). Without `break`, the description loop kept overwriting, so the *farthest* documented ancestor won. Adding `break` makes the *nearest* documented declaration win, which also aligns description/liveExamples precedence with the tags-merge block right below it (which already prefers the nearest declaration with tags). The one-line `break` is the minimal change; a `.find()` rewrite would only double-call `getDescriptionAndLiveExamples`, so the current form is preferable. `getProps` (line ~1091) is the only candidate-scanning loop of this kind — the other `getDescriptionAndLiveExamples` call sites read a single node, so no sibling bug exists.

### Validation performed

- **Test catches the regression.** Ran the new test on head → passes; removed the `break` → fails with `expected 'Farthest override.' to be 'Local override.'`. So it genuinely guards the fix. The fixture covers all three documented cases (local override, fallback to nearest parent, fallback to farthest ancestor) including the `liveExamples`-as-a-unit behavior.
- **Real-package impact (the important one).** Ran the loader against `ariakit-react-components` / `ariakit-react` on both base and head and diffed every prop's `description` + `liveExamples`: **88 description changes, all toward the more-specific/correct store** (e.g. `dialog::store` → `useDialogStore`, `combobox::store` → `useComboboxStore`, `composite-item::shouldRegisterItem` → the detailed text). Of 28 `liveExamples` changes: 20 drops were all the generic `menubar-navigation` link leaking from the `useCollectionStore` ancestor, and **every drop coincided with the description being corrected** away from collection-store — so the example was removed because it no longer belonged, not lost from a still-correct doc. The other 8 were additions where `*-provider::disclosure` now correctly surfaces the nearest declaration's examples. No case where a correct description lost its examples.
- `pnpm tsc` clean (0 errors; `astro check` loaded 173 react + 10 solid references). `oxlint` + `oxfmt` clean on both files.

### Notes (non-blocking)

- No changeset is needed — `app` is in `.changeset/config.json`'s `ignore` list, so the changeset-bot warning is expected.
- The test exercises only the React loader, but `getProps` is framework-agnostic shared logic, so React coverage is sufficient for this fix.
- The `as unknown as LoaderContext` mock follows the existing pattern in `preview-discovery.test.ts`.

PR #6451, opened 2026-06-24
Require Stripe customers for user promos

diegohaz:
## Motivation

Admin-created customer promos currently rely on an existing Stripe customer ID when creating the promotion code. If the target Clerk user has no `privateMetadata.stripeId`, Stripe receives no `customer` restriction even though the KV cache initially records the promo as customer-scoped.

That can turn a personal promo into an unrestricted promotion code and later into a site-wide sale when the webhook or sync logic classifies it from Stripe's unrestricted promotion data.

## Solution

Resolve the target Clerk user before doing any Stripe coupon or promotion-code work. For customer promos, reuse the user's existing Stripe customer or create one first, then create the promotion code with that customer restriction. If the user cannot be resolved or a Stripe customer cannot be produced, fail before touching coupons, promotion codes, or the promo KV cache.

The checkout flow now uses the same get-or-create customer helper, preserving its previous behavior while keeping the promo creation path consistent.

This PR does not add mocked unit coverage for the purchase flow; app-level end-to-end coverage can be added separately.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`


diegohaz (commented):
**Initial review** — diffed `56bf93eb6` (merge base with `main`) → `65cee0306`, then read the surrounding `stripe.ts`/`auth.ts`/`kv.ts` and traced callers.

## Summary

Solid, well-scoped fix. Resolving the user and getting-or-creating the Stripe customer *before* any coupon/promotion-code/KV side effects closes the hole where a customer-scoped promo could be written to KV while the promotion code itself was created unrestricted. The KV record now derives `type`/`user` from the resolved user object, and `createCheckout` reuses the same `getOrCreateCustomer` helper without changing behavior. No substantive findings.

## What I validated

- **Correctness of the core fix.** For a customer promo, `getSalePromoUser` resolves the user first; `getOrCreateCustomer` then returns the existing `stripeId` or creates+persists a new customer, so `promotionCodes.create` always receives a `customer` restriction. The user-less sale path is unchanged. The resolved-object is passed into `getOrCreateCustomer`, so the inner `getCustomer`→`getUser` is a pass-through and there's no duplicate Clerk fetch.
- **Clerk error shape.** `@clerk/backend@3.8.2`'s `withLegacyRequestReturn` throws a `ClerkAPIResponseError` with a numeric `status` field, so `isNotFoundError` checking `error.status === 404` matches the real not-found error from `clerk.users.getUser`.
- **Caller impact.** The only `createSalePromo` caller is the admin `setPromo` action (`app/src/actions/admin.ts`). The new throws (`user not found` / `no Stripe customer`) propagate as an Astro action error — fail-closed, which is the intended behavior, not a regression. `createCheckout`'s only caller (`plus-checkout-frame.astro`) is unaffected.
- **`createCheckout` refactor is behavior-preserving.** `getOrCreateCustomer` reproduces the prior inline get-then-create logic; `?? null` vs the old `|| null` is immaterial for non-empty Stripe IDs.
- **Test/type plumbing.** `pnpm test app/src/lib/stripe.test.ts` → 5/5 passing; `pnpm tsc` → 0 errors. The dynamic-import widening (`const stripeModulePath: string = "./stripe.ts"`) is justified: `stripe.ts` → `kv.ts` imports `cloudflare:workers`, which `app/tsconfig.test.json` cannot resolve, so static import would fail type-checking. The comment documents the why.
- **Changeset.** None required — `app` is in `.changeset/config.json`'s `ignore` list.

## Minor, non-blocking observations (not requested changes)

- `if (userObject && !customer) throw "...no Stripe customer"` is effectively unreachable in this path: `getOrCreateCustomer` only returns `null` when `createCustomer` returns `undefined`, which only happens when `stripe` is unset — already guarded by the `if (!stripe) return` at the top. I'd keep it: it's a cheap invariant guard and is symmetric with the same guard in `createCheckout`. Flagging only so it's a conscious choice; the unreachable branch is the one bit of the diff without test coverage.
- `createCheckout`'s refactor isn't directly unit-tested, but its `getOrCreateCustomer` logic is now exercised through the `createSalePromo` tests, so the gap is small.

## Not exhaustively validated

- Did not run the full app browser/Playwright suite or exercise the real Stripe/Clerk APIs; verification was unit tests + `tsc` + static call-graph tracing.

PR #6452, opened 2026-06-25
Add Composite Chrome performance tests

diegohaz:
## Motivation

`Composite` and `CompositeItem` are low-level building blocks used by higher-level widgets, but the Chrome perf workflow does not directly benchmark their mount or keyboard movement cost at scale.

The uncontrolled path also does not cover controlled `items`/`setItems` updates, where collection-store reconciliation can do substantially different work from normal private item registration.

## Solution

Add a `composite-perf` sandbox with lazy-mounted 400-item horizontal Composite fixtures:

- an uncontrolled fixture for baseline mount and sustained keyboard traversal from `Item 1` to `Item 400`;
- a controlled fixture backed by `useCompositeStore({ items, setItems })`, with an `Update items` control that replaces metadata on all items without changing their rendered labels.

The Chrome perf suite now measures uncontrolled mount, sustained uncontrolled navigation, controlled mount, controlled item updates, and a script-profile variant for the controlled update case. The update case repeats the controlled replacement enough times to clear the perf comparator's absolute-delta floor, while the profiled variant is kept separate for attribution.

### DIFF
--- a/app/src/sandbox/composite-perf/index.react.tsx
+++ b/app/src/sandbox/composite-perf/index.react.tsx
@@ -8,6 +8,20 @@ const items = Array.from(
   (_, index) => `Item ${index + 1}`,
 );
 
+interface ControlledItem {
+  id: string;
+  label: string;
+  rev?: number;
+}
+
+const controlledItems: ControlledItem[] = Array.from(
+  { length: itemCount },
+  (_, index) => {
+    const number = index + 1;
+    return { id: `item-${number}`, label: `Item ${number}` };
+  },
+);
+
 function CompositeFixture() {
   return (
     <Ariakit.CompositeProvider orientation="horizontal">
@@ -22,14 +36,58 @@ function CompositeFixture() {
   );
 }
 
+function ControlledCompositeFixture() {
+  const [items, setItems] = useState(controlledItems);
+  const [updates, setUpdates] = useState(0);
+  const composite = Ariakit.useCompositeStore({
+    items,
+    setItems,
+    orientation: "horizontal",
+  });
+
+  const updateItems = () => {
+    const rev = updates + 1;
+    setItems((items) => items.map((item) => ({ ...item, rev })));
+    setUpdates(rev);
+  };
+
+  return (
+    <>
+      <Ariakit.Button className="button" onClick={updateItems}>
+        Update items
+      </Ariakit.Button>
+      <output aria-label="Updates">{updates}</output>
+      <Ariakit.Composite
+        store={composite}
+        aria-label="Controlled composite items"
+        className="composite"
+      >
+        {items.map((item) => (
+          <Ariakit.CompositeItem key={item.id} id={item.id} className="item">
+            {item.label}
+          </Ariakit.CompositeItem>
+        ))}
+      </Ariakit.Composite>
+    </>
+  );
+}
+
 export default function Example() {
   const [mounted, setMounted] = useState(false);
+  const [controlledMounted, setControlledMounted] = useState(false);
   return (
     <div className="root">
       <Ariakit.Button className="button" onClick={() => setMounted(true)}>
         Mount composite
       </Ariakit.Button>
+      <Ariakit.Button
+        className="button"
+        onClick={() => setControlledMounted(true)}
+      >
+        Mount controlled composite
+      </Ariakit.Button>
       {mounted ? <CompositeFixture /> : null}
+      {controlledMounted ? <ControlledCompositeFixture /> : null}
     </div>
   );
 }
--- a/app/src/sandbox/composite-perf/index.react.tsx
+++ b/app/src/sandbox/composite-perf/index.react.tsx
@@ -1,0 +1,31 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+import "./style.css";
+
+const items = Array.from({ length: 400 }, (_, index) => `Item ${index + 1}`);
+
+function CompositeFixture() {
+  return (
+    <Ariakit.CompositeProvider orientation="horizontal">
+      <Ariakit.Composite aria-label="Composite items" className="composite">
+        {items.map((item) => (
+          <Ariakit.CompositeItem key={item} className="item">
+            {item}
+          </Ariakit.CompositeItem>
+        ))}
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
+
+export default function Example() {
+  const [mounted, setMounted] = useState(false);
+  return (
+    <div className="root">
+      <Ariakit.Button className="button" onClick={() => setMounted(true)}>
+        Mount composite
+      </Ariakit.Button>
+      {mounted ? <CompositeFixture /> : null}
+    </div>
+  );
+}
--- a/app/src/sandbox/composite-perf/index.react.tsx
+++ b/app/src/sandbox/composite-perf/index.react.tsx
@@ -2,7 +2,11 @@ import * as Ariakit from "@ariakit/react";
 import { useState } from "react";
 import "./style.css";
 
-const items = Array.from({ length: 400 }, (_, index) => `Item ${index + 1}`);
+const itemCount = 400;
+const items = Array.from(
+  { length: itemCount },
+  (_, index) => `Item ${index + 1}`,
+);
 
 function CompositeFixture() {
   return (
--- a/app/src/sandbox/composite-perf/perf-chrome.ts
+++ b/app/src/sandbox/composite-perf/perf-chrome.ts
@@ -1,10 +1,12 @@
 import { expect } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
 
+const itemCount = 400;
+
 withFramework(import.meta.dirname, async ({ test }) => {
   test("mount composite", async ({ q, perf }) => {
     const mountButton = q.button("Mount composite");
-    const lastItem = q.button("Item 400");
+    const lastItem = q.button(`Item ${itemCount}`);
 
     await perf.measure(async () => {
       await mountButton.click();
@@ -12,15 +14,17 @@ withFramework(import.meta.dirname, async ({ test }) => {
     });
   });
 
-  test("move to next item", async ({ q, page, perf }) => {
+  test("move across items", async ({ q, page, perf }) => {
     const mountButton = q.button("Mount composite");
     const firstItem = q.button("Item 1");
-    const secondItem = q.button("Item 2");
+    const lastItem = q.button(`Item ${itemCount}`);
 
     await perf.measure(
       async () => {
-        await page.keyboard.press("ArrowRight");
-        await expect(secondItem).toBeFocused();
+        for (let i = 1; i < itemCount; i++) {
+          await page.keyboard.press("ArrowRight");
+        }
+        await expect(lastItem).toBeFocused();
       },
       {
         setup: async () => {
--- a/app/src/sandbox/composite-perf/perf-chrome.ts
+++ b/app/src/sandbox/composite-perf/perf-chrome.ts
@@ -1,0 +1,34 @@
+import { expect } from "@playwright/test";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("mount composite", async ({ q, perf }) => {
+    const mountButton = q.button("Mount composite");
+    const lastItem = q.button("Item 400");
+
+    await perf.measure(async () => {
+      await mountButton.click();
+      await expect(lastItem).toBeVisible();
+    });
+  });
+
+  test("move to next item", async ({ q, page, perf }) => {
+    const mountButton = q.button("Mount composite");
+    const firstItem = q.button("Item 1");
+    const secondItem = q.button("Item 2");
+
+    await perf.measure(
+      async () => {
+        await page.keyboard.press("ArrowRight");
+        await expect(secondItem).toBeFocused();
+      },
+      {
+        setup: async () => {
+          await mountButton.click();
+          await firstItem.focus();
+          await expect(firstItem).toBeFocused();
+        },
+      },
+    );
+  });
+});
--- a/app/src/sandbox/composite-perf/perf-chrome.ts
+++ b/app/src/sandbox/composite-perf/perf-chrome.ts
@@ -1,7 +1,10 @@
+import type { query } from "@ariakit/test/playwright";
 import { expect } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
 
 const itemCount = 400;
+const updateCount = 20;
+type Query = ReturnType<typeof query>;
 
 withFramework(import.meta.dirname, async ({ test }) => {
   test("mount composite", async ({ q, perf }) => {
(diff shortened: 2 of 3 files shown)


app/src/sandbox/composite-perf/perf-chrome.ts
```diff
@@ -0,0 +1,34 @@
...
+        await expect(secondItem).toBeFocused();
+      },
+      {
+        setup: async () => {
+          await mountButton.click();
+          await firstItem.focus();
+          await expect(firstItem).toBeFocused();
+        },
+      },
+    );
```
diegohaz, line 39: The "move to next item" benchmark measures a single `ArrowRight` move. This PR's own perf CI run reports it at **2.1ms scripting / 0.7ms rendering / 2.9ms total**, while "mount composite" lands at ~80ms scripting. The problem is that the Chrome comparison only flags a primary-metric change (`scripting`/`rendering`/`total`) when the median delta is **both `> 10%` and `≥ 5ms`** — see `MIN_SIGNIFICANT_DELTA_MS = 5` and `THRESHOLD_PERCENT = 10` in `packages/ariakit-scripts/src/perf-compare.ts` (`magnitudeOk = percentOk && deltaOk`).

Because the whole measurement is ~2-3ms, the absolute-delta gate makes this case effectively undetectable: even *doubling* the per-move cost (delta ≈ +2.1ms) stays under the 5ms floor. A primary metric has to grow by ~5ms — i.e. well over +150% over a ~3ms base — before any regression is reported. So once this lands on `main` and gains a baseline, the case will keep printing numbers but can't actually function as a regression detector, which is its purpose. The `mount composite` case has ample headroom (a 10% regression there is ~8ms) and is fine.

A single move also doesn't really exercise "keyboard movement cost at scale" from the motivation: the per-move work is dominated by a fixed re-render (roving tabindex / `data-active-item` on two items) and doesn't grow with the 400 items.

Fix: accumulate many moves inside one measured interaction so the signal clears the 5ms floor and a proportional per-move regression becomes a detectable delta. Traversing the full set is deterministic (Ariakit doesn't loop by default, so 399 presses land on and stay at `Item 400`):

```ts
test("move across items", async ({ q, page, perf }) => {
  const mountButton = q.button("Mount composite");
  const firstItem = q.button("Item 1");
  const lastItem = q.button("Item 400");

  await perf.measure(
    async () => {
      for (let i = 0; i < items.length - 1; i++) {
        await page.keyboard.press("ArrowRight");
      }
      await expect(lastItem).toBeFocused();
    },
    {
      setup: async () => {
        await mountButton.click();
        await firstItem.focus();
        await expect(firstItem).toBeFocused();
      },
    },
  );
});
```

Each `ArrowRight` drives an `activeId` update + re-render, so ~399 of them should push scripting well above the gate and make this a meaningful "sustained navigation at scale" benchmark. The exact magnitude (and a smaller count if 399×iterations is too slow) is worth confirming on the next perf CI run.

diegohaz (commented):
## Review summary

**Initial review** of the full PR (3 new files, +106), diffed against the merge base `e2c9cbec1` (`origin/main`) at head `57a0978`.

### Scope checked
- The new `composite-perf` sandbox fixture, styles, and `perf-chrome.ts` cases against the existing perf infrastructure: `packages/ariakit-scripts/src/perf.ts` (`createPerfMeasure`, warmup/iterations, `setup`, `resetPage`, `settlePaint`, INP/CDP metric capture), `app/src/test-utils/{fixtures,preview}.ts`, and the perf project wiring in `app/playwright.config.ts`.
- Conventions vs. the sibling `dialog-perf` / `ariakit-tailwind` perf sandboxes: naming, lazy-mount-to-measure pattern, no `preview.json` (title defaults to folder name), and that `find src -path '*perf*-chrome.ts'` in `perf.yml` auto-discovers the file (no registration needed).
- Test correctness: `q.button(...)` uses `exact: true` (no `Item 1` vs `Item 10`/`Item 100` ambiguity), `CompositeItem` defaults to `<button>` so role queries resolve, and `toBeVisible()` on `Item 400` is a valid "did it mount" signal (Playwright visibility doesn't require in-viewport). The `focus()` + `ArrowRight` movement is consistent with `examples/composite/test.ts`.
- The PR's own CI: all checks green, including the Chrome perf shards. The perf bot reported both new cases — `mount composite` at 79.9ms / 11.3ms / 91.2ms and `move to next item` at 2.1ms / 0.7ms / 2.9ms (scripting/rendering/total).

### Findings
One actionable finding, posted inline:
- **`move to next item` is below the perf workflow's significance floor.** At ~2-3ms total it can never clear the `MIN_SIGNIFICANT_DELTA_MS = 5` absolute-delta gate (`THRESHOLD_PERCENT = 10`), so after it gains a baseline on `main` it will print numbers but can't flag a realistic regression. Suggested fix: accumulate many `ArrowRight` moves in one measured interaction so the signal clears the floor and better matches the "movement cost at scale" motivation.

No correctness, compatibility, or security issues found. No changeset needed (test-only sandbox, no shipped-package change). The `mount composite` case is well-designed and has ample headroom. CSS and fixture are clean.

### Not fully validated
I did not run the heavy `preview-lite` perf harness locally (production build is slow and risks contaminating the worktree); instead I relied on the PR's real Chrome perf CI run plus a read of the perf infrastructure and `perf-compare.ts` thresholds. The exact post-fix magnitude of the suggested multi-move benchmark should be confirmed on the next perf CI run.

app/src/sandbox/composite-perf/perf-chrome.ts
diegohaz: Fixed in d97e8441a - changed the Composite navigation benchmark to traverse from `Item 1` to `Item 400` with repeated `ArrowRight` presses, raising the measured interaction above the Chrome perf comparator 5ms delta floor. Local focused perf verification reported `move across items` at 158.2ms scripting / 405.0ms rendering / 563.2ms total.

diegohaz:
**Fixed**

- d97e8441a ([reply](https://github.com/ariakit/ariakit/pull/6452#discussion_r3471185961)): Updated the Composite keyboard benchmark to traverse from `Item 1` to `Item 400` with repeated `ArrowRight` presses, so the measured interaction clears the Chrome perf comparator absolute delta floor. Local focused perf verification reported `move across items` at 158.2ms scripting / 405.0ms rendering / 563.2ms total.

app/src/sandbox/composite-perf/index.react.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Ariakit.Composite aria-label="Composite items" className="composite">
+        {items.map((item) => (
+          <Ariakit.CompositeItem key={item} className="item">
+            {item}
+          </Ariakit.CompositeItem>
+        ))}
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
```
diegohaz, line 37: Now that the navigation case is fixed, the bigger coverage gap is that this sandbox is **uncontrolled** (`CompositeProvider` with no `items`/`setItems`), so it only exercises the collection store's fast paths (`isPublicStateCurrent() === true`). The heavy reconciliation added in #6435 — `mergeItemsFromState`/`mergeItemsFromPrivate` building 2–3 `Map`s + `shallowEqual` per item, plus the new `sync(collection, ["items"])` full `itemsMap` rebuild — only fires on the **controlled** path (`!isPublicStateCurrent()`), i.e. when an external `items`/`setItems` diverges from private registration. None of that is measured here.

Suggestion: add a **controlled** variant to this same sandbox (mirroring the controlled `collection-6295` sandbox from #6435), gated behind its own mount button, plus an "Update items" control so a controlled `items` change at scale can be measured. That update is the `mergeItemsFromState`/`updateItemsMap` hot path #6435 most plausibly regresses, and it's an *update-time* cost the current mount-only fixture never reaches.

```tsx
interface ControlledItem { id: string; label: string; rev?: number }

const controlledItems: ControlledItem[] = Array.from(
  { length: itemCount },
  (_, i) => ({ id: `item-${i + 1}`, label: `Item ${i + 1}` }),
);

function ControlledCompositeFixture() {
  const [items, setItems] = useState(controlledItems);
  const [updates, setUpdates] = useState(0);
  const composite = Ariakit.useCompositeStore({
    items,
    setItems,
    orientation: "horizontal",
  });
  // Replace metadata on every item to force a full reconciliation, and bump a
  // visible counter so the perf test has a deterministic, render-cheap signal.
  const updateItems = () => {
    setItems((items) => items.map((item) => ({ ...item, rev: updates + 1 })));
    setUpdates((n) => n + 1);
  };
  return (
    <>
      <Ariakit.Button className="button" onClick={updateItems}>
        Update items
      </Ariakit.Button>
      {/* role=status — assert this in the update perf case */}
      <output aria-label="Updates">{updates}</output>
      <Ariakit.Composite
        store={composite}
        aria-label="Controlled composite items"
        className="composite"
      >
        {items.map((item) => (
          <Ariakit.CompositeItem key={item.id} id={item.id} className="item">
            {item.label}
          </Ariakit.CompositeItem>
        ))}
      </Ariakit.Composite>
    </>
  );
}
```

Two notes for whoever picks this up (take #6435 as context for what the controlled path should stress):
- Update a metadata field that isn't the rendered text (`rev` above, not `label`) so the change still drives the store's reconciliation even though the rendered output is unchanged. React will still reconcile the 400 elements, so the `scriptProfile` variant requested on `perf-chrome.ts` is what attributes store time vs. render time.
- Follow #6435's controlled `collection-6295` sandbox for the exact `items`/`setItems` wiring so the public-state-authoritative path is the one being exercised.

app/src/sandbox/composite-perf/perf-chrome.ts
```diff
@@ -0,0 +1,38 @@
...
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("mount composite", async ({ q, perf }) => {
+    const mountButton = q.button("Mount composite");
+    const lastItem = q.button(`Item ${itemCount}`);
+
+    await perf.measure(async () => {
+      await mountButton.click();
+      await expect(lastItem).toBeVisible();
+    });
+  });
```
diegohaz, line 18: To actually guard #6435 (the collection-store registration/reconciliation rewrite), add cases that drive the **controlled** fixture suggested on `index.react.tsx` — the two existing cases stay on the uncontrolled fast paths and never reach the new merge logic:

```ts
test("mount controlled composite", async ({ q, perf }) => {
  const mountButton = q.button("Mount controlled composite");
  const lastItem = q.button(`Item ${itemCount}`);
  await perf.measure(async () => {
    await mountButton.click();
    await expect(lastItem).toBeVisible();
  });
});

test("update controlled items", async ({ q, perf }) => {
  const updateButton = q.button("Update items");
  // Locator for the "Updates" output counter — pick whatever fits the signal
  // you expose (e.g. q.status("Updates") / q.text).
  const updates = q.status("Updates");
  await perf.measure(
    async () => {
      await updateButton.click();
      await expect(updates).toHaveText("1");
    },
    {
      setup: async () => {
        await q.button("Mount controlled composite").click();
        await expect(q.button(`Item ${itemCount}`)).toBeVisible();
      },
    },
  );
});
```

`update controlled items` is the important one: it measures a controlled `items` replacement at scale — the `mergeItemsFromState`/`updateItemsMap` path from #6435 — which the mount-only cases never hit. Confirm the measured deltas clear the 5ms floor (see the thread above); a 400-item full reconciliation should, but verify on the next perf CI run.

Also add a `scriptProfile` variant (like `dialog-perf`'s `open dialog with script profile` case) on the controlled mount and/or update, so the PR comment shows where the scripting goes and a collection-store regression is attributable rather than buried under React render time:

```ts
test("update controlled items with script profile", async ({ q, perf }) => {
  // same interaction + setup as "update controlled items"
  await perf.measure(interaction, {
    label: "update controlled items (script profile)",
    scriptProfile: true,
    profileLimit: 20,
    setup,
  });
});
```

Profiling adds overhead, so keep the unprofiled case as the timing source and read the profiled one only for attribution.

diegohaz (commented):
## Re-review (follow-up)

Re-reviewed at head `d97e844`, incremental from the previously reviewed `57a0978`.

**Prior finding resolved:** the navigation case now traverses `Item 1` → `Item 400` with repeated `ArrowRight` presses (commit `d97e844`), which lifts the measured interaction well above the Chrome comparator's 5ms delta floor (author-reported ~563ms total). Good.

**New findings (2 inline), about coverage for #6435** (the collection-store registration/reconciliation rewrite that also touches Composite/Select navigation):

- `index.react.tsx` — this sandbox is **uncontrolled**, so it only hits the collection store's fast paths and never exercises #6435's heavy controlled reconciliation (`mergeItemsFromState`/`mergeItemsFromPrivate` + the new `sync(collection, ["items"])` rebuild), which is gated behind `!isPublicStateCurrent()`. Suggested adding a controlled variant in the same sandbox (mirroring #6435's `collection-6295`) plus an "Update items" control so a controlled `items` change at scale can be measured.
- `perf-chrome.ts` — add `mount controlled composite` and `update controlled items` cases driving that variant (the update is the `mergeItemsFromState`/`updateItemsMap` hot path the mount-only cases never reach), and a `scriptProfile` variant (like `dialog-perf`'s) so a collection-store regression is attributable rather than buried under React render time.

Both inline comments reference #6435 and the controlled `collection-6295` pattern so a worker can pick the task up directly.

**Not validated:** I did not run the heavy `preview-lite` perf harness; the controlled-update magnitude (and that it clears the 5ms floor) should be confirmed on the next perf CI run.

app/src/sandbox/composite-perf/perf-chrome.ts
diegohaz: Fixed in 646046bee - added controlled mount, controlled item update, and script-profiled controlled update perf cases. The update benchmark repeats 20 controlled metadata replacements so the measured signal clears the 5ms comparator floor; local focused perf reported `update controlled items` at 12.9ms scripting / 11.1ms rendering / 24.0ms total and the script-profile variant at 12.0ms / 9.5ms / 21.4ms with a script profile attached.

app/src/sandbox/composite-perf/index.react.tsx
diegohaz: Fixed in 646046bee - added a controlled Composite fixture backed by `useCompositeStore({ items, setItems })`, gated behind `Mount controlled composite`, with an `Update items` control that replaces per-item metadata while keeping rendered labels stable and exposes an `Updates` output for deterministic assertions.

diegohaz:
**Fixed**

- 646046bee (https://github.com/ariakit/ariakit/pull/6452#discussion_r3471258614): Added a controlled Composite fixture with `items`/`setItems`, metadata-only item updates, and a visible `Updates` assertion signal.
- 646046bee (https://github.com/ariakit/ariakit/pull/6452#discussion_r3471258612): Added controlled mount, controlled update, and script-profiled controlled update Chrome perf cases. Local focused perf verification reported the unprofiled controlled update at 12.9ms scripting / 11.1ms rendering / 24.0ms total.

**Verification**

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm lint-css`
- `pnpm build-app-lite`
- `PERF_RESULTS_FILE=composite-controlled-fix-20.json PERF_ITERATIONS=1 PERF_WARMUP=0 pnpm -F app run test-perf src/sandbox/composite-perf/perf-chrome.ts`

app/src/sandbox/composite-perf/index.react.tsx
diegohaz: Verified on `646046b` — the controlled fixture is correct: `useCompositeStore({ items, setItems })` makes public `items` authoritative, the `rev`-only `updateItems` keeps rendered labels stable while still forcing the store's reconciliation, and `<output aria-label="Updates">` exposes a `status` role, so `q.status("Updates")` is a valid, deterministic signal (`status` is in the query role list). Both the controlled mount and the controlled update run green in the latest Chrome perf shards. Thanks for adding this.

app/src/sandbox/composite-perf/perf-chrome.ts
diegohaz: Verified on `646046b`. The three new cases run green in the latest Chrome perf shards, with CI medians: `mount controlled composite` 80.5ms total, `update controlled items` 48.2ms total (23.5ms scripting / 24.9ms rendering), and the script-profile variant 51.5ms with the profile attributing self-time to `ControlledCompositeFixture` (`index.react.tsx:39`).

At ~48ms total on CI the update case sits ~10× the 5ms floor, so it has normal ~10% regression sensitivity on `total` — good headroom for catching #6435 reconciliation regressions. No further changes needed.

diegohaz (commented):
## Re-review

Re-reviewed at head `646046b`, incremental from the last reviewed `d97e844` (new commit: `Add controlled Composite perf coverage`).

All three prior findings are resolved and **verified against the current code + CI** (not just the author's replies):
- Navigation case now traverses `Item 1` → `Item 400` (CI median 2049.5ms total) — well above the 5ms floor.
- Controlled fixture added with `useCompositeStore({ items, setItems })`, metadata-only `rev` updates that keep rendered labels stable, and an `Updates` `status` output for deterministic assertions.
- Controlled `mount` / `update` / script-profiled `update` perf cases added.

Verification performed this run:
- Confirmed `q.status("Updates")` resolves (`<output>` → role `status`, and `status` is in the query role list).
- Confirmed latest CI is green, including both Chrome perf shards, Lint, Type check, and the framework suites.
- Read the CI perf medians: mount 79.3 / move-across 2049.5 / mount-controlled 80.5 / update-controlled 48.2 / update (script profile) 51.5 ms total. At ~48ms total the update case is ~10× the 5ms floor → ~10% regression sensitivity on `total`, comparable to the mount cases and adequate to guard #6435's controlled reconciliation. The script profile attributes self-time to `ControlledCompositeFixture` as intended.

No new substantive findings. The new code follows conventions (`const`-arrow nested helpers, `interface` for the item shape + `type` alias for the derived query type, exact-name role queries, shared setup/update helpers with no needless duplication). Posted confirmation replies on the two open threads.

Not validated: I did not re-run the `preview-lite` perf harness locally; I relied on the PR's own Chrome perf CI run.

PR #6453, opened 2026-06-25
Fix form submit autofocus order

### LINKED ISSUE #6305: Failed form submit focuses the wrong invalid field when fields mount out of DOM order (autoFocusOnSubmit)
### Current behavior

When a form submission fails validation, [`Form`](https://ariakit.com/reference/form) (via the [`autoFocusOnSubmit`](https://ariakit.com/reference/form#autofocusonsubmit) prop, enabled by default) is documented to focus "the first invalid field" ([JSDoc](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/form/form.tsx#L173-L176)). But if an invalid field was mounted dynamically *above* an existing invalid field (progressive disclosure, conditionally revealed sections, wizard steps), the failed submit focuses the **lower** field and silently skips the first invalid field in the document.

This breaks the primary error-recovery affordance for keyboard and screen-reader users: fixing errors top-to-bottom from the focused field, they never encounter the earlier error above, and the next submit fails again with no apparent cause.

The root cause is that `getFirstInvalidField` scans the store's `items` array ([form.tsx#L30-L36](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/form/form.tsx#L30-L36)), which the focus effect subscribes to ([form.tsx#L92-L104](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/form/form.tsx#L92-L104)). `items` is in registration (mount) order — the collection store appends newly registered items to the end ([collection-store.ts#L145-L157](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L145-L157)). The store maintains a separate `renderedItems` state that is kept DOM-sorted via `sortBasedOnDOMPosition` precisely for this situation ([collection-store.ts#L79-L83](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/collection/collection-store.ts#L79-L83)), but the form package never uses it.

3. Click "Add company details" — a required "Company" field appears **above** the existing required "Email" field.
4. Leave both fields empty and click "Sign up".
5. Observe that focus lands on the **Email** field (lower in the form), not the Company field, even though both are invalid (`aria-invalid="true"`) and Company is the first invalid field in the document.

```tsx

  const form = Ariakit.useFormStore({
    defaultValues: { company: "", email: "" },
  });
  const [showCompany, setShowCompany] = useState(false);

  return (
    <Ariakit.Form store={form}>
      {showCompany && (
        <div>
          <Ariakit.FormLabel name={form.names.company}>
            Company
          </Ariakit.FormLabel>
          <Ariakit.FormInput name={form.names.company} required />
          <Ariakit.FormError name={form.names.company} />
        </div>
      )}

      <div>
        <Ariakit.FormLabel name={form.names.email}>Email</Ariakit.FormLabel>
        <Ariakit.FormInput name={form.names.email} required />
        <Ariakit.FormError name={form.names.email} />
      </div>

      {!showCompany && (
        <button type="button" onClick={() => setShowCompany(true)}>
          Add company details
        </button>
      )}

      <Ariakit.FormSubmit>Sign up</Ariakit.FormSubmit>
    </Ariakit.Form>
  );
}
```

Verified with a real-browser Playwright test using only user-level interactions (click, submit, focus assertions).

### Expected behavior

After a failed submit, focus should land on the first invalid field **in document order** — here the Company field at the top of the form — as the `Form` documentation promises ("the first invalid field is automatically focused").

### Workaround

Disable [`autoFocusOnSubmit`](https://ariakit.com/reference/form#autofocusonsubmit) and focus the first invalid field in DOM order manually when a submission fails:

```tsx
const formRef = useRef<HTMLFormElement>(null);
const submitFailed = Ariakit.useStoreState(form, "submitFailed");

useEffect(() => {
  if (!submitFailed) return;
  const field = formRef.current?.querySelector<HTMLElement>(
    "[aria-invalid='true']",
  );
  if (!field) return;
  field.focus();
  if (field instanceof HTMLInputElement) {
    field.select();
  }
}, [submitFailed]);

// ...
<Ariakit.Form ref={formRef} store={form} autoFocusOnSubmit={false}>
```

`document.querySelector` traverses in document order, so this always picks the topmost invalid field. Verified: the same Playwright test passes with this workaround applied.

### Possible solutions

Subscribe to the DOM-sorted `renderedItems` state instead of `items` in `useForm` and pass it to `getFirstInvalidField` (which works unchanged, since fields register into both arrays with the same item object, including the `type: "field"` metadata):

```diff
--- a/packages/ariakit-react-components/src/form/form.tsx
+++ b/packages/ariakit-react-components/src/form/form.tsx
@@
-function getFirstInvalidField(items: FormStoreState["items"]) {
+function getFirstInvalidField(items: FormStoreState["renderedItems"]) {
@@
-  const items = useStoreState(store, "items");
+  const renderedItems = useStoreState(store, "renderedItems");
@@
-    const field = getFirstInvalidField(items);
+    const field = getFirstInvalidField(renderedItems);
@@
-  }, [autoFocusOnSubmit, submitFailed, items]);
+  }, [autoFocusOnSubmit, submitFailed, renderedItems]);
```

Timing is safe: `renderedItems` is sorted in a `requestAnimationFrame` scheduled at item mount, while `form-store.submit()` awaits a later-scheduled frame via `nextFrame()` before incrementing `submitFailed` ([form-store.ts#L339-L349](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/form/form-store.ts#L339-L349)). The `items` subscription in `useForm` feeds only this effect (the `onBlur` handler reads `store.getState().items` directly), so it can be replaced outright. This also matches existing precedent — `menu.tsx`, `combobox.tsx`, and `composite-container.tsx` already subscribe to `renderedItems` for DOM-order operations.

diegohaz:
## Motivation

When a failed [`Form`](https://ariakit.com/reference/form) submit has multiple invalid fields, `autoFocusOnSubmit` should focus the first invalid field in document order.

Issue #6305 reports a progressive-disclosure case where a new required field is mounted above an existing invalid field. The form focused the lower field because it scanned collection registration order instead of DOM order.

Fixes #6305.

## Solution

Added a `form-6305` React sandbox that reproduces the registration-order mismatch: Email registers first, then Company mounts above it, and both fields are invalid after submit. The browser regression verifies that Company receives focus in Chrome, Firefox, and Safari.

The [`Form`](https://ariakit.com/reference/form) submit autofocus lookup now sorts the registered form items by DOM position before selecting the first invalid field. This keeps the existing registered-item source for form fields while making the focus target match document order.

The temporary sandbox workaround from the draft stage was removed after the library fix. The happy-dom duplicate from the repro stage was also removed because it does not reliably model the form registration/effect timing for this behavior; the cross-browser Playwright test is the authoritative coverage.

## Workaround

Until this fix is released, disable the built-in submit autofocus and focus the first invalid field from the form DOM after a failed submit:

```tsx
const formRef = useRef<HTMLFormElement>(null);
const submitFailed = Ariakit.useStoreState(form, "submitFailed");

// TODO: Remove when https://github.com/ariakit/ariakit/issues/6305 is fixed.
useEffect(() => {
  if (!submitFailed) return;
  const field = formRef.current?.querySelector<HTMLElement>(
    "[aria-invalid='true']",
  );
  if (!field) return;
  field.focus();
  if (field instanceof HTMLInputElement) {
    field.select();
  } else if (field instanceof HTMLTextAreaElement) {
    field.select();
  }
}, [submitFailed]);

return (
  <Ariakit.Form ref={formRef} store={form} autoFocusOnSubmit={false}>
    {/* fields */}
  </Ariakit.Form>
);
```


packages/ariakit-react-components/src/form/form.tsx
diegohaz, line 71: Shouldn't we use renderedItems here instead of sortBasedOnDOMPosition?

diegohaz: Keeping `sortBasedOnDOMPosition(items, …)` is the safer choice here — I'd not switch to plain `renderedItems`.

Both are viable for the documented repro: form fields register through `renderItem`, so `renderedItems` is populated, and I confirmed that swapping `getFirstInvalidField` to read `renderedItems` directly (no sort) still passes `form-6305/test-browser.ts` in Chrome. In progressive disclosure the reveal and the submit are separate user actions, so by submit time `renderedItems` has already been DOM-sorted.

The difference is robustness in the same-frame window. `renderedItems` is sorted **asynchronously** — the collection store runs `sortBasedOnDOMPosition` inside a `requestAnimationFrame` in `sortItems` (`collection-store.ts`), so right after a field mounts (before that rAF fires) `renderedItems` still reflects registration/merge order, not DOM order. If a field mounts and the form submits within the same frame, reading `renderedItems` would refocus the lower field again — the exact bug this PR fixes. Sorting `items` reads the **live DOM** at focus time, so it's correct regardless of whether the rAF sort has run.

This also matches the existing precedent in `composite-typeahead.tsx`, which prefers `renderedItems` but still falls back to `items` / re-sorts with `sortBasedOnDOMPosition` rather than trusting the rendered order unconditionally.

The cost is negligible — this runs only on a failed submit over a handful of fields, and `sortBasedOnDOMPosition` returns the input array unchanged when the order already matches — so I'd keep the current approach.

diegohaz (commented):
**Initial review** — baseline: `main`…`18371fb` merge-base (full diff, 4 files). Reviewed the diff plus the form store, collection store, `sortBasedOnDOMPosition`, the `items`/`renderedItems` model, and sibling form sandboxes/changesets.

**Verdict: looks good.** The fix is correct, minimal, and well-targeted, and the regression coverage is solid. No blocking findings.

**Validated locally (Chrome, app browser suite + tsc):**
- `form-6305/test-browser.ts` is a genuine regression test: with the unfixed `items.find(…)` it fails at `company.toBeFocused()` (Email gets focus); with `sortBasedOnDOMPosition` it passes (Company gets focus).
- `pnpm tsc` is clean (0 errors). `sortBasedOnDOMPosition` is exported from `@ariakit/utils` and preserves the `FormStoreItem[]` type; it returns the input array unchanged when order already matches, so there's no needless allocation.
- The `items` vs `renderedItems` question (the open self-review thread): I confirmed plain `renderedItems` also passes the repro, but the current live-DOM sort is more robust than the rAF-batched `renderedItems` sort. Detailed reasoning posted on that thread — recommend keeping `sortBasedOnDOMPosition`.

**Conventions checked:**
- Changeset scope/format (`@ariakit/react-components` + `@ariakit/react`, patch) matches sibling form changesets (e.g. `300-form-focus-steal-on-items-change`).
- Sandbox naming (`form-6305`), the `ak` import alias, and omitting `preview.json` all match existing form sandboxes (e.g. `form-6309`).
- Omitting the happy-dom `test.ts` is reasonable: the behavior leans on the collection store's rAF-batched DOM sort and microtask `items` flush, which happy-dom doesn't model reliably; the cross-browser Playwright test is the authoritative coverage.

**Not exhaustively validated:**
- Ran the Chrome project only, not Firefox/Safari (the test asserts focus order, which isn't browser-specific) and did not run the full build.
- Did not construct the same-frame mount+submit edge case in a live component; the rAF-staleness reasoning behind the `renderedItems` recommendation is grounded in `collection-store.ts` source rather than a constructed repro.

PR #6459, opened 2026-06-25
Fix iframe text field detection

### LINKED ISSUE #6316: Arrow keys steal focus from text fields inside same-origin iframes: isTextField uses realm-bound instanceof
### Current behavior

When Ariakit components are rendered into a same-origin iframe while the library code runs in the parent window — the standard `react-frame-component` / visual-builder / embedded-preview pattern, where React creates the DOM nodes through the iframe's document — every text input is silently classified as a non-text field.

The root cause is [`isTextField`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L185-L203) in `@ariakit/utils`, which checks `element instanceof HTMLInputElement`. `instanceof` resolves against the current realm's constructor, so an `<input>` created by the iframe's document is an instance of the *iframe's* `HTMLInputElement`, not the parent's, and the check returns `false` for a genuine text input. This contradicts the cross-realm policy the same file documents and enforces in [`isElement`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L93-L95), which deliberately tests `nodeType` instead of `instanceof Element` so that "elements coming from same-origin child frames … aren't wrongly rejected for belonging to a different realm". Iframe support is otherwise maintained on this path: `getActiveElement` recurses into frames, `addGlobalEventListener` attaches listeners to child frames, and #5142 fixed focusable-element traversal inside iframes.

The most user-visible consequence is in composite widgets ([`Toolbar`](https://ariakit.com/reference/toolbar) shown below, but any [`CompositeItem`](https://ariakit.com/reference/composite-item)-based widget is affected). [`useCompositeItem`'s keydown handler](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/composite/composite-item.tsx#L374-L388) only lets arrow keys move composite focus away from a textbox item when the caret is at the start/end of the text — but that guard goes through `isTextbox`/[`getTextboxSelection`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-utils/src/dom.ts#L209-L236), which both rely on `isTextField`. Inside an iframe the guard is skipped entirely, so pressing ArrowLeft/ArrowRight in the middle of the text immediately moves focus to the previous/next toolbar item instead of moving the caret.

Other paths gated by the same check misbehave too: [`Command`'s Enter click emulation](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L103) no longer bails out for iframe text inputs, so Enter is `preventDefault()`-ed (suppressing native form submission) and a synthetic click is fired instead; `getTextboxValue` returns `""` and `getTextboxSelection` returns `{ start: 0, end: 0 }` for iframe inputs, which Combobox and `ComboboxItem` use for caret and value handling. None of these throw, so the failure surfaces only as inexplicable interaction bugs.

The behavior was verified with a Playwright test driving real user interactions: the identical toolbar rendered in the parent document behaves correctly, while the iframe-rendered copy loses focus on the first ArrowLeft press.

3. In the **first** toolbar (parent document), click the "Find" text field, type `hi`, then press <kbd>ArrowLeft</kbd>. The caret moves left and focus stays in the input — the expected behavior. Only a third <kbd>ArrowLeft</kbd> press (caret at the start) moves focus to "Bold".
4. In the toolbar **inside the iframe**, click the "Find" text field, type `hi`, then press <kbd>ArrowLeft</kbd> once.
5. Focus immediately jumps to the "Bold" button, even though the caret was at the end of the text.

```tsx

function FormattingToolbar() {
  return (
    <Ariakit.Toolbar aria-label="Formatting" style={{ display: "flex", gap: 8 }}>
      <Ariakit.ToolbarItem>Bold</Ariakit.ToolbarItem>
      <Ariakit.ToolbarItem render={<input type="text" aria-label="Find" />} />
      <Ariakit.ToolbarItem>Italic</Ariakit.ToolbarItem>
    </Ariakit.Toolbar>
  );
}

  const [iframe, setIframe] = useState<HTMLIFrameElement | null>(null);

  useEffect(() => {
    if (!iframe) return;
    // Same-origin embedded widget (react-frame-component style): render the
    // exact same toolbar into the iframe's own document. React creates the
    // elements through the iframe's document while the Ariakit code keeps
    // running in the parent window.
    if (!doc?.body) return;
    const root = createRoot(doc.body);
    root.render(<FormattingToolbar />);
    return () => {
      setTimeout(() => root.unmount());
    };
  }, [iframe]);

  return (
    <div style={{ display: "grid", gap: 16 }}>
      <FormattingToolbar />
      <iframe
        ref={setIframe}
        title="Embedded editor"
        style={{ border: "1px solid", display: "block", height: 80, width: 360 }}
      />
    </div>
  );
}
```

### Expected behavior

The iframe-rendered toolbar should behave exactly like the parent-document one: while the caret is not at the start of the text, <kbd>ArrowLeft</kbd> moves the caret within the input and focus stays on it; toolbar focus only moves to the previous item once the caret reaches the start (and symmetrically for <kbd>ArrowRight</kbd>/the end). More generally, `isTextField` should recognize text inputs from same-origin child frames, matching the cross-realm policy `isElement` documents in the same file.

### Workaround

Reimplement the caret-edge check in userland via the [`moveOnKeyPress`](https://ariakit.com/reference/composite-item#moveonkeypress) prop, which is consulted right before the composite moves focus. Reading `selectionStart`/`selectionEnd` directly is realm-safe:

```tsx
import type { KeyboardEvent } from "react";

function shouldMoveOnKeyPress(event: KeyboardEvent<HTMLElement>) {
  const input = event.currentTarget as HTMLInputElement;
  if (event.key === "ArrowLeft") return input.selectionStart === 0;
  if (event.key === "ArrowRight") {
    return input.selectionEnd === input.value.length;
  }
  return true;
}

<Ariakit.ToolbarItem
  render={<input type="text" aria-label="Find" />}
  moveOnKeyPress={shouldMoveOnKeyPress}
/>;
```

Verified: with this prop, the Playwright test passes for both the parent-document and the iframe toolbar. Note this only covers the arrow-key path; other `isTextField`-gated behaviors (such as `Command`'s Enter handling, which can be patched with `clickOnEnter={false}` on the item) need their own per-call-site workarounds.

### Possible solutions

Replace the realm-bound `instanceof` with a `tagName` check, mirroring the precedent `isElement` and `isFrame` set in the same file. This is semantically equivalent in-realm (the `selectionStart` probe still distinguishes text from non-text inputs, and non-input elements can't reach it), and the Safari `try/catch` around the probe stays:

```diff
 export function isTextField(
   element: Element,
 ): element is HTMLInputElement | HTMLTextAreaElement {
   try {
-    const isTextInput =
-      element instanceof HTMLInputElement && element.selectionStart !== null;
-    const isTextArea = element.tagName === "TEXTAREA";
-    return isTextInput || isTextArea || false;
+    if (element.tagName === "TEXTAREA") return true;
+    return (
+      element.tagName === "INPUT" &&
+      (element as HTMLInputElement).selectionStart !== null
+    );
   } catch (_error) {
     // Safari throws an exception when trying to get `selectionStart` on
     // non-text <input> elements ...
     return false;
   }
 }
```

As a side benefit the function no longer references the `HTMLInputElement` global at all. The change is a patch to the published `@ariakit/utils` package, so it needs a changeset.

diegohaz:
## Motivation

Text fields rendered into a same-origin iframe were misclassified as non-text fields when Ariakit code ran in the parent window. The browser repro for #6316 shows the parent-document toolbar behaving correctly while the iframe-rendered toolbar moved focus out of the input instead of moving the caret.

## Solution

Added `toolbar-iframe-6316`, a React sandbox that renders the same `Toolbar` in the parent document and in an iframe document. The browser tests keep the parent-document case as a passing control and isolate the iframe case.

Updated `isTextField` to use realm-safe tag-name checks before probing `selectionStart`, avoiding `instanceof HTMLInputElement` so inputs created by same-origin child frames are recognized. The sandbox workaround was then removed, and the iframe tests pass through the library fix.

Fixes #6316.

## Workaround

Consumers can reimplement the text-boundary check in userland with [`moveOnKeyPress`](https://ariakit.com/reference/composite-item#moveonkeypress). Reading `selectionStart` and `selectionEnd` directly avoids the realm-bound text-field check.

```tsx
import type { KeyboardEvent } from "react";

// https://github.com/ariakit/ariakit/issues/6316 is fixed.
function shouldMoveOnKeyPress(event: KeyboardEvent<HTMLElement>) {
  const input = event.currentTarget as HTMLInputElement;
  if (event.key === "ArrowLeft") {
    return (input.selectionStart ?? 0) === 0;
  }
  if (event.key === "ArrowRight") {
    return (input.selectionEnd ?? 0) === input.value.length;
  }
  return true;
}

<Ariakit.ToolbarItem
  render={<input type="text" aria-label="Find" />}
  moveOnKeyPress={shouldMoveOnKeyPress}
/>;
```

### DIFF
--- a/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
+++ b/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
@@ -49,7 +49,7 @@ async function expectToolbarInputNavigation({
 
 // https://github.com/ariakit/ariakit/issues/6316
 withFramework(import.meta.dirname, async ({ test, query }) => {
-  test("ToolbarItem input navigates at text boundaries inside iframes", async ({
+  test("ToolbarItem input navigates at text boundaries in the parent document", async ({
     page,
     q,
   }) => {
@@ -59,7 +59,13 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
       input: q.textbox("Find"),
       previousItem: q.button("Bold"),
     });
+  });
 
+  test("ToolbarItem input navigates at text boundaries inside iframes", async ({
+    page,
+  }) => {
+    // Keep this isolated from the parent-document control so the test enters
+    // the iframe the same way a user would in the reported reproduction.
     const frame = query(page.frameLocator("iframe[title='Embedded editor']"));
 
     await expectToolbarInputNavigation({
--- a/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
+++ b/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
@@ -12,6 +12,20 @@ interface ExpectToolbarInputNavigationParams {
   expect: Expect;
   input: Locator;
   previousItem: Locator;
+  nextItem: Locator;
+}
+
+async function setSelection(input: Locator, start: number, end = start) {
+  await input.evaluate(
+    (element, range) => {
+      if (!(element instanceof HTMLInputElement)) {
+        throw new Error("Expected an input element");
+      }
+      const input = element;
+      input.setSelectionRange(range.start, range.end);
+    },
+    { start, end },
+  );
 }
 
 function getSelectionStart(input: Locator) {
@@ -28,10 +42,28 @@ async function expectToolbarInputNavigation({
   expect,
   input,
   previousItem,
+  nextItem,
 }: ExpectToolbarInputNavigationParams) {
   await input.click();
   await page.keyboard.type("hi");
+  await setSelection(input, 0);
 
+  await page.keyboard.press("ArrowRight");
+
+  await expect(input).toBeFocused();
+  await expect.poll(() => getSelectionStart(input)).toBe(1);
+
+  await page.keyboard.press("ArrowRight");
+
+  await expect(input).toBeFocused();
+  await expect.poll(() => getSelectionStart(input)).toBe(2);
+
+  await page.keyboard.press("ArrowRight");
+
+  await expect(nextItem).toBeFocused();
+
+  await input.focus();
+  await setSelection(input, 2);
   await page.keyboard.press("ArrowLeft");
 
   await expect(input).toBeFocused();
@@ -58,6 +90,7 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
       expect: test.expect,
       input: q.textbox("Find"),
       previousItem: q.button("Bold"),
+      nextItem: q.button("Italic"),
     });
   });
 
@@ -73,6 +106,7 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
       expect: test.expect,
       input: frame.textbox("Find"),
       previousItem: frame.button("Bold"),
+      nextItem: frame.button("Italic"),
     });
   });
 });
--- a/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
+++ b/app/src/sandbox/toolbar-iframe-6316/test-browser.ts
@@ -1,19 +1,5 @@
-import type {
-  Locator,
-  Page,
-  expect as playwrightExpect,
-} from "@playwright/test";
+import type { Locator } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
-
-type Expect = typeof playwrightExpect;
-
-interface ExpectToolbarInputNavigationParams {
-  page: Page;
-  expect: Expect;
-  input: Locator;
-  previousItem: Locator;
-  nextItem: Locator;
-}
 
 async function setSelection(input: Locator, start: number, end = start) {
   await input.evaluate(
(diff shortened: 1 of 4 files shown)


app/src/sandbox/toolbar-iframe-6316/test-browser.ts
```diff
@@ -0,0 +1,112 @@
...
+function getSelectionStart(input: Locator) {
+  return input.evaluate((element) => {
+    if (!(element instanceof HTMLInputElement)) {
+      throw new Error("Expected an input element");
+    }
+    return element.selectionStart;
+  });
+}
+
+async function expectToolbarInputNavigation({
```
diegohaz: I don't think we need this function. This DRY is not worth it.

diegohaz: Fixed in 2413170bd - removed `expectToolbarInputNavigation` and inlined the assertions in both tests. The two scenarios now keep their setup and keyboard steps together, while the small DOM selection helpers remain extracted for the `evaluate` plumbing.

diegohaz:
**Fixed**
- 2413170bd ([reply](https://github.com/ariakit/ariakit/pull/6459#discussion_r3472109493)): Removed `expectToolbarInputNavigation` and inlined the parent-document and iframe toolbar assertions so each scenario is easier to read.

PR #6460, opened 2026-06-25
Profile uncontrolled composite perf interactions

diegohaz:
## Motivation

The composite perf report currently spends the React script profile slot on the controlled item update path. For the current investigation, the more useful script profiles are the uncontrolled mount and keyboard movement paths.

## Solution

Add script-profile variants for the existing `mount composite` and `move across items` perf tests. The profiled and unprofiled cases now share helpers so they exercise the same uncontrolled interactions.

Remove `update controlled items (script profile)` while keeping the unprofiled controlled update baseline.

## Verification

```sh
pnpm lint-fix
pnpm -F app run test-perf src/sandbox/composite-perf/perf-chrome.ts
```

### DIFF
--- a/app/src/sandbox/composite-perf/perf-chrome.ts
+++ b/app/src/sandbox/composite-perf/perf-chrome.ts
@@ -1,5 +1,6 @@
 import type { query } from "@ariakit/test/playwright";
 import { expect } from "@playwright/test";
+import type { Page } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
 
 const itemCount = 400;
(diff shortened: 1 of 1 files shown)


app/src/sandbox/composite-perf/perf-chrome.ts
```diff
@@ -1,37 +1,81 @@
...
+    await expect(q.button("Item 1")).toBeVisible();
+  };
+
+  const moveAcrossItems = async (q: Query, page: Page) => {
+    const lastItem = q.button(`Item ${itemCount}`);
+    for (let i = 1; i < itemCount; i++) {
+      await page.keyboard.press("ArrowRight");
+    }
+    await expect(lastItem).toBeFocused();
+  };
```
diegohaz: Can't we move these outside the `withFramework` function?

diegohaz: Fixed in 784416451 - moved the composite perf helper functions to module scope so `withFramework` only registers the tests.

diegohaz:
**Fixed**

- 784416451 ([reply](https://github.com/ariakit/ariakit/pull/6460#discussion_r3472006922)): Moved the composite perf helper functions to module scope to address the helper placement review thread.

PR #6464, opened 2026-06-25
Fix StackBlitz TypeScript configs

diegohaz:
## Motivation

The app's generated StackBlitz projects currently emit `tsconfig.json` with `module: "esnext"` and `moduleResolution: "node16"`. TypeScript 5.2+ rejects that pair with TS5110, so users who open an app example in StackBlitz hit a config error when type-checking or building the generated project.

## Solution

Use `moduleResolution: "bundler"` for app-generated StackBlitz tsconfigs while keeping `module: "esnext"`, matching the Vite and Next.js bundler-driven scaffolds. The app Solid generator now reuses the shared tsconfig helper with Solid JSX overrides so these defaults stay in one place.

Add regression coverage for the app StackBlitz frameworks (`react-vite`, `solid-vite`, and `react-nextjs`). The legacy `website` workspace is intentionally unchanged.


app/src/lib/stackblitz.test.ts
```diff
@@ -26,6 +29,33 @@ afterEach(() => {
   embedProjectMock.mockReset();
 });
```
diegohaz: Shouldn't these imports be moved to the top before the `vi.hoisted/mock`?

website/lib/stackblitz.test.ts
```diff
@@ -0,0 +1,46 @@
+import { afterEach, expect, test, vi } from "vitest";
```
diegohaz: No need to do anything on the website workspace. Revert this.

website/lib/stackblitz.ts
```diff
@@ -99,7 +99,7 @@ function getTSConfig(tsConfig?: Record<string, unknown>) {
       lib: ["dom", "dom.iterable", "esnext"],
       module: "esnext",
       skipLibCheck: true,
-      moduleResolution: "node16",
+      moduleResolution: "bundler",
```
diegohaz: No need to change website workspace.

app/src/lib/stackblitz.test.ts
diegohaz: Fixed in 64a61eb6e - moved the type-only StackBlitz imports above the Vitest mock setup. The runtime import stays after `vi.mock` so the tested module still loads with the mocked SDK.

website/lib/stackblitz.test.ts
diegohaz: Fixed in c3f9f3200 - removed the legacy website StackBlitz test addition so the website workspace is unchanged.

website/lib/stackblitz.ts
diegohaz: Fixed in c3f9f3200 - restored `website/lib/stackblitz.ts` to its previous module resolution so this PR only changes the app generator.

diegohaz:
**Fixed**

- 64a61eb6e ([reply](https://github.com/ariakit/ariakit/pull/6464#discussion_r3472338776)): Moved the type-only StackBlitz test imports above the Vitest mock setup while keeping the runtime import after `vi.mock`.
- c3f9f3200 ([reply](https://github.com/ariakit/ariakit/pull/6464#discussion_r3472338901)): Removed the legacy website StackBlitz test addition.
- c3f9f3200 ([reply](https://github.com/ariakit/ariakit/pull/6464#discussion_r3472339002)): Restored the legacy website StackBlitz generator so this PR only changes the app generator.

PR #6469, opened 2026-06-25
Remove dead SelectList role observer

diegohaz:
## Motivation

`useSelectList` was reading the rendered `role` attribute with `useAttribute`, deriving `isCompositeRole`, and folding that into `ariaMultiSelectable`. That value is only consumed inside the existing `if (composite)` block, where it always reduces to `multiSelectable || undefined` regardless of the observed role.

This left every mounted `SelectList`/`SelectPopover` paying for a state value and `MutationObserver` that could not affect the rendered props.

## Solution

Remove the unused `useAttribute`/role computation and inline the only reachable `aria-multiselectable` value in the composite branch. The existing prop spread order is preserved, so user-provided props keep the same override behavior.

No changeset is included because this is a behavior-preserving internal cleanup.

## Verification

- `pnpm lint-fix`
- `pnpm tsc` (rerun passed after an initial Vite optimizer cache miss in `app/node_modules/.vite-check`)
- `pnpm test-react packages/ariakit-test/src/select.react.test.tsx examples/select-multiple/test.ts examples/select-combobox/test.ts`

### DIFF
--- a/packages/ariakit-react-components/src/select/select-list.tsx
+++ b/packages/ariakit-react-components/src/select/select-list.tsx
@@ -1,6 +1,5 @@
 import { useStoreState } from "@ariakit/react-store";
 import {
-  useAttribute,
   useBooleanEvent,
   useEvent,
   useId,
@@ -137,18 +136,9 @@ export const useSelectList = createHook<TagName, SelectListOptions>(
     const hasCombobox = !!store.combobox;
     composite = composite ?? (!hasCombobox && childStore !== store);
 
-    const [element, setElement] = useTransactionState(
+    const [, setElement] = useTransactionState(
       composite ? store.setListElement : null,
     );
-
-    const role = useAttribute(element, "role", props.role);
-    const isCompositeRole =
-      role === "listbox" ||
-      role === "menu" ||
-      role === "tree" ||
-      role === "grid";
-    const ariaMultiSelectable =
-      composite || isCompositeRole ? multiSelectable || undefined : undefined;
 
     const hidden = isHidden(mounted, props.hidden, alwaysVisible);
     const style = hidden ? { ...props.style, display: "none" } : props.style;
@@ -156,7 +146,7 @@ export const useSelectList = createHook<TagName, SelectListOptions>(
     if (composite) {
       props = {
         role: "listbox",
-        "aria-multiselectable": ariaMultiSelectable,
+        "aria-multiselectable": multiSelectable || undefined,
         ...props,
       };
     }


packages/ariakit-react-components/src/select/select-list.tsx
```diff
@@ -137,26 +136,17 @@ export const useSelectList = createHook<TagName, SelectListOptions>(
...
-      composite || isCompositeRole ? multiSelectable || undefined : undefined;
-
     const hidden = isHidden(mounted, props.hidden, alwaysVisible);
     const style = hidden ? { ...props.style, display: "none" } : props.style;
 
     if (composite) {
       props = {
         role: "listbox",
-        "aria-multiselectable": ariaMultiSelectable,
+        "aria-multiselectable": multiSelectable || undefined,
```
diegohaz, line 149: **The removal is correct and behavior-preserving — but it's the last fingerprint of a behavior that's been silently broken since #3711, and it does not fix the `aria-multiselectable`-on-unsupported-roles concern.**

**1. Confirming the dead code.** The old `ariaMultiSelectable` was `composite || isCompositeRole ? multiSelectable || undefined : undefined`, and its only consumer was this `if (composite)` block. Inside that block `composite` is always truthy, so the ternary always collapses to `multiSelectable || undefined` regardless of `isCompositeRole`. So the `useAttribute` role observer never affected the rendered props. Removing it is safe. 👍

**2. Answer to "will `aria-multiselectable` be added to unsupporting roles?"** Yes, it can — but that's **unchanged by this PR**, and the removed code was never preventing it. The default role here is `listbox` (which supports the attribute), so the only way to land it on an unsupporting role is to override `role`. I verified this empirically with `@ariakit/test`:

| Component | role | value | rendered `aria-multiselectable` |
|---|---|---|---|
| SelectPopover (composite) | `listbox` (default) | array | `"true"` ✅ |
| SelectPopover (composite) | `listbox` | string | absent ✅ |
| SelectPopover `role="menu"` | `menu` | array | **`"true"` (leaks)** ⚠️ |
| ComboboxList `role="menu"` | `menu` | array | `null` (suppressed) ✅ |

So `SelectList` leaks `aria-multiselectable` onto an overridden non-composite role, while `ComboboxList` (which still has the live `isCompositeRole` gate, guarded by `!hasListboxInside` rather than `composite`) correctly suppresses it. This divergence already exists on `main`; this cleanup neither introduces nor removes it.

**3. What we're "missing" — the #3711 regression.** This dead code is a vestige of #2428, whose changelog still claims:

> The `SelectList` and `SelectPopover` components will now automatically render the `aria-multiselectable` attribute even when the `composite` prop is set to `false`, but only when the underlying element has a composite role.

In #2428 the attribute was applied in the **unconditional** props spread, so `composite || isCompositeRole` was load-bearing (both the `composite=false`-with-composite-role case and the role-based suppression worked). #3711 (`d5ebe3921`, "Support nested SelectList and new Select components") moved `"aria-multiselectable"` *into* the `if (composite)` block, which (a) dropped the documented `composite=false` behavior and (b) turned the role gate into dead code (this PR's subject) and lost the role-based suppression. None of that is caused by this PR — but removing the code erases the last hint of it.

**Decision path (all out of scope for this cleanup):**
- The #2428 changelog entry is now stale. In practice the `composite=false` listbox is a sibling `ComboboxList` that carries `aria-multiselectable` itself (verified: a multi-select combobox listbox renders `aria-multiselectable="true"`), so the `SelectList` branch appears superseded rather than load-bearing — but the changelog should be reconciled.
- If `SelectList`/`SelectPopover` is meant to mirror `ComboboxList` and suppress the attribute on non-composite roles, that's a separate behavior fix (restore a real role check **and** move the attribute out of `if (composite)`), with its own changeset + a test asserting `role="menu"` ⇒ no `aria-multiselectable`. There is currently **no test** covering `aria-multiselectable` on `SelectList` at all, which is why #3711's regression went unnoticed.

Happy to merge this as-is; just flagging so the regression/divergence isn't lost along with the code that hinted at it.

diegohaz (commented):
**Initial review** · baseline = merge-base `8556488e` → head `36127437` (full diff, 1 file) · comment-only.

**Verdict: the cleanup is correct and behavior-preserving — safe to merge.** The removed `useAttribute`/`isCompositeRole`/`ariaMultiSelectable` computation was genuinely dead: its only consumer sat inside `if (composite)`, where `composite || isCompositeRole` always short-circuits to `multiSelectable || undefined`. Confirmed by code, by git history, and empirically. No changeset needed (internal refactor, no shipped behavior change).

**On the stated concern ("will `aria-multiselectable` be added to unsupporting roles?")** — see the inline comment on `select-list.tsx:149`. Short version:
- `aria-multiselectable` *can* land on an unsupporting role, but only if you override `role` away from the default `listbox` on a multi-select Select. This is **unchanged by the PR** — the role observer it removes was already dead, so it was never gating this.
- The removed code is a leftover from #2428, whose behavior ("render `aria-multiselectable` even when `composite=false`, only for composite roles") was silently broken in #3711 when the attribute was moved into the `if (composite)` block. That regression — and the resulting divergence from `ComboboxList`, which still suppresses the attribute on non-composite roles — predates this PR. Removing the dead code is fine, but it erases the last trace of that intent, and the #2428 changelog is now stale.

**What I ran (in this worktree, on the PR head):**
- `pnpm install`, then `pnpm test-react` against a temporary `@ariakit/test` render harness (since deleted) probing `aria-multiselectable` across SelectPopover (composite/single/`role="menu"`) and ComboboxList (multi-select/`role="menu"`). Results in the inline comment's table.
- Traced the regression to #3711 (`d5ebe3921`) via `git show` on the pre-split `ariakit-react-core` path.

**Suggested (non-blocking) follow-ups, separate from this cleanup:**
1. Reconcile the #2428 changelog with current behavior.
2. Decide whether `SelectList`/`SelectPopover` should match `ComboboxList`'s role-aware suppression; if so, it needs a real fix (role check + attribute applied outside `if (composite)`) plus the first-ever test for `aria-multiselectable` on `SelectList`.

**Not validated:** whether any real-world `composite={false}` `SelectList` without a sibling `ComboboxList` relies on the lost #2428 behavior — I couldn't find one in the examples, and the documented Select+Combobox path delegates the listbox role and `aria-multiselectable` to `ComboboxList`.

PR #6470, opened 2026-06-25
Guard generated workflow writebacks

diegohaz:
## Motivation

Generated-file workflows check out a mutable branch tip, generate artifacts, and then call the shared `commit-or-pr` writeback action. Without a final branch-tip guard, a stale run can publish generated output based on older sources after the branch has advanced.

`docs` and `build-styles` were missing the stale-writeback protection already used by `og-images`, `visual` had an early latest-commit gate but no final SHA check at writeback time, and the shared action did not own the clean skip behavior for generated writebacks.

## Solution

Centralize generated writeback staleness handling in `commit-or-pr`:

- cancel superseded `docs` and `build-styles` runs on the same ref
- make `commit-or-pr` verify the branch tip before creating a maintenance PR or committing to a PR branch
- default the expected SHA to the action checkout's `HEAD`, so docs/styles/OG images no longer need duplicate workflow-level verification steps
- keep `expected-head-sha` as an override for workflows like `visual`, where the late `pr` checkout may differ from the commit that generated artifacts
- skip cleanly when the branch has already advanced, while still refusing the pull request push if the branch advances after the generated commit is created

This keeps stale runs from silently publishing generated docs, styles, OG images, or screenshots against a newer branch tip while keeping the guard in one shared action.

## Testing

- `pnpm lint-fix`
- `pnpm lint`
- `git diff --check`
- `ruby -e 'require "yaml"; ARGV.each { |file| YAML.load_file(file); puts "#{file}: ok" }' .github/workflows/commit-or-pr/action.yml .github/workflows/visual.yml .github/workflows/docs.yml .github/workflows/build-styles.yml .github/workflows/og-images.yml`

### DIFF
--- a/.github/workflows/build-styles.yml
+++ b/.github/workflows/build-styles.yml
@@ -7,6 +7,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
@@ -47,10 +48,31 @@ jobs:
           fi
           echo "has_changes=true" >> "$GITHUB_OUTPUT"
 
-      - name: Commit or create PR
+      - name: Verify branch tip
+        id: branch
         if: >
           steps.patch.outputs.has_changes == 'true' &&
           env.WRITEBACK_ALLOWED == 'true'
+        shell: bash
+        run: |
+          set -euo pipefail
+          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
+          git fetch --no-tags origin "$branch"
+          local_sha="$(git rev-parse HEAD)"
+          remote_sha="$(git rev-parse FETCH_HEAD)"
+          if [ "$local_sha" != "$remote_sha" ]; then
+            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
+            echo "is_current=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "is_current=true" >> "$GITHUB_OUTPUT"
+          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
+
+      - name: Commit or create PR
+        if: >
+          steps.patch.outputs.has_changes == 'true' &&
+          env.WRITEBACK_ALLOWED == 'true' &&
+          steps.branch.outputs.is_current == 'true'
         uses: ./.github/workflows/commit-or-pr
         with:
           # This write-back step intentionally assumes PRs come from branches
@@ -60,6 +82,7 @@ jobs:
           event: ${{ github.event_name }}
           message: Update styles
           head-branch: ${{ github.head_ref || github.ref_name }}
+          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
           new-branch: chore/update-styles
           paths: app/src/styles/styles.json
           app-client-id: ${{ vars.ARIAKIT_CI_APP_CLIENT_ID }}
--- a/.github/workflows/build-styles.yml
+++ b/.github/workflows/build-styles.yml
@@ -48,31 +48,10 @@ jobs:
           fi
           echo "has_changes=true" >> "$GITHUB_OUTPUT"
 
-      - name: Verify branch tip
-        id: branch
-        if: >
-          steps.patch.outputs.has_changes == 'true' &&
-          env.WRITEBACK_ALLOWED == 'true'
-        shell: bash
-        run: |
-          set -euo pipefail
-          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
-          git fetch --no-tags origin "$branch"
-          local_sha="$(git rev-parse HEAD)"
-          remote_sha="$(git rev-parse FETCH_HEAD)"
-          if [ "$local_sha" != "$remote_sha" ]; then
-            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
-            echo "is_current=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-          echo "is_current=true" >> "$GITHUB_OUTPUT"
-          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
-
       - name: Commit or create PR
         if: >
           steps.patch.outputs.has_changes == 'true' &&
-          env.WRITEBACK_ALLOWED == 'true' &&
-          steps.branch.outputs.is_current == 'true'
+          env.WRITEBACK_ALLOWED == 'true'
         uses: ./.github/workflows/commit-or-pr
         with:
           # This write-back step intentionally assumes PRs come from branches
@@ -82,7 +61,6 @@ jobs:
           event: ${{ github.event_name }}
           message: Update styles
           head-branch: ${{ github.head_ref || github.ref_name }}
-          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
           new-branch: chore/update-styles
           paths: app/src/styles/styles.json
           app-client-id: ${{ vars.ARIAKIT_CI_APP_CLIENT_ID }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
@@ -47,10 +48,31 @@ jobs:
           fi
           echo "has_changes=true" >> "$GITHUB_OUTPUT"
 
-      - name: Commit or create PR
+      - name: Verify branch tip
+        id: branch
         if: >
           steps.patch.outputs.has_changes == 'true' &&
           env.WRITEBACK_ALLOWED == 'true'
+        shell: bash
+        run: |
+          set -euo pipefail
+          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
+          git fetch --no-tags origin "$branch"
+          local_sha="$(git rev-parse HEAD)"
+          remote_sha="$(git rev-parse FETCH_HEAD)"
+          if [ "$local_sha" != "$remote_sha" ]; then
+            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
+            echo "is_current=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "is_current=true" >> "$GITHUB_OUTPUT"
+          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
+
+      - name: Commit or create PR
+        if: >
+          steps.patch.outputs.has_changes == 'true' &&
+          env.WRITEBACK_ALLOWED == 'true' &&
+          steps.branch.outputs.is_current == 'true'
         uses: ./.github/workflows/commit-or-pr
         with:
           # This write-back step intentionally assumes PRs come from branches
@@ -60,6 +82,7 @@ jobs:
           event: ${{ github.event_name }}
           message: Update docs
           head-branch: ${{ github.head_ref || github.ref_name }}
+          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
           new-branch: chore/update-docs
           paths: packages/*/readme.md
           app-client-id: ${{ vars.ARIAKIT_CI_APP_CLIENT_ID }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,31 +48,10 @@ jobs:
           fi
           echo "has_changes=true" >> "$GITHUB_OUTPUT"
 
-      - name: Verify branch tip
-        id: branch
-        if: >
-          steps.patch.outputs.has_changes == 'true' &&
-          env.WRITEBACK_ALLOWED == 'true'
-        shell: bash
-        run: |
-          set -euo pipefail
-          branch="${GITHUB_HEAD_REF:-$GITHUB_REF_NAME}"
-          git fetch --no-tags origin "$branch"
-          local_sha="$(git rev-parse HEAD)"
-          remote_sha="$(git rev-parse FETCH_HEAD)"
-          if [ "$local_sha" != "$remote_sha" ]; then
-            echo "Branch advanced from $local_sha to $remote_sha; skipping write-back."
-            echo "is_current=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-          echo "is_current=true" >> "$GITHUB_OUTPUT"
-          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
-
       - name: Commit or create PR
         if: >
           steps.patch.outputs.has_changes == 'true' &&
-          env.WRITEBACK_ALLOWED == 'true' &&
-          steps.branch.outputs.is_current == 'true'
+          env.WRITEBACK_ALLOWED == 'true'
         uses: ./.github/workflows/commit-or-pr
         with:
           # This write-back step intentionally assumes PRs come from branches
@@ -82,7 +61,6 @@ jobs:
           event: ${{ github.event_name }}
           message: Update docs
           head-branch: ${{ github.head_ref || github.ref_name }}
-          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
           new-branch: chore/update-docs
           paths: packages/*/readme.md
           app-client-id: ${{ vars.ARIAKIT_CI_APP_CLIENT_ID }}
--- a/.github/workflows/commit-or-pr/action.yml
+++ b/.github/workflows/commit-or-pr/action.yml
@@ -9,13 +9,13 @@ inputs:
     description: Commit message and PR title
     required: true
   head-branch:
-    description: Branch name when committing to existing PR
+    description: Branch name when committing to an existing PR, or base branch when creating a new PR
     required: true
   expected-head-sha:
     description: >
-      Expected current SHA for the head branch when committing to an existing
-      PR. When set, the action refuses to rebase generated changes onto a
-      different branch tip.
+      Expected current SHA for the branch named by head-branch. When set, the
+      action refuses to publish generated changes against a different branch
+      tip.
     required: false
     default: ""
   new-branch:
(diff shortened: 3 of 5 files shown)


.github/workflows/build-styles.yml
```diff
@@ -60,6 +82,7 @@ jobs:
           event: ${{ github.event_name }}
           message: Update styles
           head-branch: ${{ github.head_ref || github.ref_name }}
+          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
```
diegohaz: **Consistency: `visual.yml` is the one `commit-or-pr` writeback this guard doesn't reach.**

This PR brings `docs` and `build-styles` to parity with `og-images` (the new steps are byte-identical to it — nice). But `commit-or-pr` has four callers, and after this PR `visual.yml` is the only generated-file writeback that still doesn't pass `expected-head-sha`. It relies solely on its start-of-run gate (`Read tip SHA on branch` → `Compare with current SHA`), with no re-check at writeback time.

It's actually the widest stale-rebase window of the four:

- Its caller `app.yml` has no `concurrency`/`cancel-in-progress`, so superseded runs are not cancelled (whereas this PR adds `cancel-in-progress: true` here).
- A long pipeline runs between the early gate and the writeback (`check-artifacts` → `Checkout branch` → `Download screenshots` → `Prepare`/`Extract` → `Commit or create PR`). If the branch advances during that span, `commit-or-pr` rebases the old-run screenshots onto the new tip and pushes — exactly the "generated from older sources" scenario in this PR's motivation, just for screenshots instead of docs/styles.

Was excluding it intentional scope? If not, the fix is small and symmetric: `visual.yml` already computes the verified tip in `steps.tip.outputs.tip_sha` (and `is_latest` means `HEAD_SHA == tip_sha`), so passing it through closes the window:

```yaml
      - name: Commit or create PR
        # ...
        with:
          working-directory: pr
          event: ${{ inputs.event }}
          message: Update screenshots
          head-branch: ${{ github.head_ref || github.ref_name }}
          expected-head-sha: ${{ steps.tip.outputs.tip_sha }}
          new-branch: visual/update-screenshots
          # ...
```

If the omission is deliberate (e.g., the early gate is considered sufficient for screenshots), a one-line note on why would help, since the three siblings now all carry the guard.

.github/workflows/docs.yml
```diff
@@ -60,6 +82,7 @@ jobs:
           event: ${{ github.event_name }}
           message: Update docs
           head-branch: ${{ github.head_ref || github.ref_name }}
+          expected-head-sha: ${{ steps.branch.outputs.local_sha }}
```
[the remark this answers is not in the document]
diegohaz: Reviewed this during a general PR review. It's a real gap in the abstract, but worth some scope/severity context before acting on it:

- **Inherited, not introduced.** This is verbatim from the `og-images` pattern this PR mirrors — `og-images` passes `expected-head-sha` the same way and the `Create PR` (push) path ignores it there too. So docs/build-styles now match `og-images` exactly; it isn't a regression added here.
- **The push path is self-healing.** On push to `main`, the writeback target isn't the protected branch — it's the dedicated bot PR branch (`chore/update-docs` / `chore/update-styles`), which `create-pull-request` force-updates on every run. A stale update is corrected by the next push-triggered run regenerating from the new `main`. The case `expected-head-sha` actually guards — a stale rebase landing on a *contributor's* branch — is the `pull_request` path, and that path **is** covered (`Push to PR head branch` enforces it).
- **The fix isn't "make `Create PR` read the input."** `create-pull-request` publishes to a *separate* branch, so the head-ref guard used on the `pull_request` path doesn't map. The meaningful guard there is "refuse if the **base** (`main`) advanced," a different check that belongs in `commit-or-pr` itself — and it would apply equally to `og-images`.

So: real but narrow, and self-healing on the push path. If we want to close it, it's a follow-up enhancement to `commit-or-pr` (base-advanced check) covering all three push writebacks, rather than a change to this parity PR.

diegohaz (commented):
## Review summary

**Type:** Initial review (no prior review from this skill on this PR).
**Baseline:** merge-base of `main`…`f7ea870` (full diff), then expanded to the surrounding `.github/workflows` context.
**Scope:** `.github/workflows/docs.yml`, `.github/workflows/build-styles.yml`, plus the related `commit-or-pr` action, `og-images.yml` (the ported pattern), and `visual.yml`/`app.yml` (sibling writebacks).

### What I verified
- The added `concurrency.cancel-in-progress`, `Verify branch tip` step, gated `Commit or create PR`, and `expected-head-sha` wiring are **byte-identical** to the already-merged `og-images.yml` pattern — a faithful port.
- `commit-or-pr` already declares and consumes `expected-head-sha` (in the `pull_request` `Push to PR head branch` step); this PR only wires it up, so the input is real and not new.
- The two-layer design (workflow-level skip + action-level refuse) is intentional and not redundant: `Verify branch tip` converts an advanced-branch case into a clean **skip** (exit 0) instead of `commit-or-pr`'s hard **failure**, while still computing `local_sha` for the belt-and-suspenders re-check. No simplification worth making.
- YAML valid (`ruby -e 'YAML.load_file'` on both files), `git diff --check` clean. No local `actionlint`/`shellcheck`, but the shell is identical to the in-production `og-images` step.
- No changeset needed — CI-only change, no shipped-package impact (the changeset-bot note is informational).

### Findings
1. **Consistency (inline, `build-styles.yml`):** `visual.yml` is the fourth `commit-or-pr` writeback and the only one still without `expected-head-sha`. With its caller `app.yml` lacking `cancel-in-progress` and a long screenshot pipeline between its start-of-run gate and the writeback, it has the widest stale-rebase window of the four. Flagged to confirm whether the omission is intentional scope; if not, passing `steps.tip.outputs.tip_sha` closes it symmetrically.
2. **Codex push-path thread (reply):** the `expected-head-sha`-not-enforced-on-push observation is real but narrow — inherited verbatim from `og-images`, self-healing on the push path (the bot PR branch is force-updated each run), and the meaningful fix is a *base-advanced* check inside `commit-or-pr` (a follow-up that would cover `og-images` too), not a change to this parity PR.

No correctness, security, or regression issues found in the diff itself. The change does what the description claims for the `pull_request` path it targets.

.github/workflows/docs.yml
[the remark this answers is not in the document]
diegohaz: Fixed in 9295036f9 - `commit-or-pr` now verifies `expected-head-sha` before `create-pull-request` runs on push events, so stale maintenance PR writebacks are refused too.

.github/workflows/build-styles.yml
diegohaz: Fixed in 9295036f9 - `visual.yml` now passes `steps.tip.outputs.tip_sha` to `commit-or-pr`, and the shared action enforces that SHA before both PR-branch and push-created maintenance PR writebacks.

diegohaz:
**Fixed**

- 9295036f9 (https://github.com/ariakit/ariakit/pull/6470#discussion_r3476982206): `commit-or-pr` now enforces `expected-head-sha` before push-triggered maintenance PR writebacks.
- 9295036f9 (https://github.com/ariakit/ariakit/pull/6470#discussion_r3476982209): `visual.yml` now passes the verified tip SHA through to the shared writeback guard.

.github/workflows/docs.yml
```diff
@@ -47,10 +48,31 @@ jobs:
...
+            exit 0
+          fi
+          echo "is_current=true" >> "$GITHUB_OUTPUT"
+          echo "local_sha=$local_sha" >> "$GITHUB_OUTPUT"
+
+      - name: Commit or create PR
+        if: >
+          steps.patch.outputs.has_changes == 'true' &&
+          env.WRITEBACK_ALLOWED == 'true' &&
+          steps.branch.outputs.is_current == 'true'
```
diegohaz: Shouldn't it be abstracted into the commit-or-pr reusable workflow?

diegohaz: Yes — and it's worth doing, but there's one real constraint that stops it from being a clean "move it all in."

The fetch-and-compare logic now lives in **three** places: the workflow `Verify branch tip` step (×3: og-images/docs/build-styles), the action's new `Verify base branch` (push), and the action's `Push to PR head branch` (pull_request). The only reason the first one lives in the *workflow* is the **skip-vs-fail semantics**: a composite action invoked via `uses:` can't make its caller's step "skipped" — it can only succeed or fail. So the clean **green skip** in the common case (branch already advanced) has to be decided before the action is called. The two in-action checks exist to close the residual window and they **fail** (red) because by then the action is committed to running.

So two shapes:

**A — action owns staleness and skips cleanly (deletes all 3 `Verify branch tip` steps).** Move the check to the very start of `commit-or-pr`; on advance, `log + exit 0` (no-op success) instead of `exit 1`. Callers drop the step and the `is_current` gate entirely; the action computes the expected SHA itself (`git rev-parse HEAD`) when none is passed.
- Removes ~17 lines × 3 of duplicated YAML and the whole "is the input wired into both paths?" class of bug (the one Codex just caught).
- Changes the residual-window outcome from fail→skip, which is arguably *better* (a benign race shouldn't redden a run).

**B — keep the per-workflow skip, factor only the shared shell** into a tiny `verify-branch-tip` action. Lower payoff: another action to maintain for ~10 lines, and the skip still can't live inside `commit-or-pr`.

**The constraint either way:** `expected-head-sha` must stay an optional input, because **`visual.yml` can't use a self-computed HEAD.** Its expected SHA is `steps.tip.outputs.tip_sha` — read from the API gate at the *start* of the run — whereas its `pr` checkout (`ref: head_ref`) resolves to the branch tip at *checkout* time, which can already be newer than `tip_sha`. If the action compared the remote against its own `git rev-parse HEAD` there, a branch that advanced *with* the checkout would match and a stale screenshot push would slip through. So visual must inject the gate SHA; the action can only default to HEAD when the input is empty.

I'd lean toward A as a **follow-up** (CI-only, no changeset): the current PR is already correct and internally consistent, and folding visual's asymmetry into the shared action deserves its own focused change rather than being bolted onto this one. If you do it, make the fail→skip behavior change explicit in the step's comment.

.github/workflows/docs.yml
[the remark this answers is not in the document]
diegohaz: Re-review (head `9295036`): verified — the new `Verify base branch` step in `commit-or-pr` runs for `push` events before `Create PR`, fetches `origin/$HEAD_REF` and refuses (`exit 1`) if the tip no longer matches `expected-head-sha`. That's exactly the base-advanced guard this thread was asking for, and it now sits in the shared action so og-images/docs/build-styles all get it. The behavior is symmetric with the `pull_request` path's `Push to PR head branch` check (workflow-level clean skip for the common case, action-level hard-fail only in the residual TOCTOU window). Correct and complete.

.github/workflows/build-styles.yml
diegohaz: Re-review (head `9295036`): verified — `visual.yml` now passes `expected-head-sha: ${{ steps.tip.outputs.tip_sha }}`, and because `tip_sha` is the SHA the `is_latest` gate validated against (not the `pr` checkout's HEAD, which can be newer), the shared action correctly refuses if the branch advanced during the screenshot pipeline. This is the right value to inject — using a self-computed HEAD in the `pr` checkout would have let a co-advancing branch slip a stale screenshot push through. Closes the window I flagged.

diegohaz (commented):
## Re-review summary

**Type:** Re-review. **Baseline:** my prior review at `f7ea870`; incremental diff `f7ea870..9295036` (commit "Enforce generated writeback SHAs"). **Scope this round:** `.github/workflows/commit-or-pr/action.yml`, `.github/workflows/visual.yml` (the only two files changed since baseline; `docs.yml`/`build-styles.yml` are untouched, so my prior review of those still stands).

### Both prior findings are fixed — verified
- **Codex push-path gap → fixed.** New `Verify base branch` step in `commit-or-pr` runs for `push` events before `Create PR`, fetches `origin/$HEAD_REF`, and refuses (`exit 1`) if the tip no longer matches `expected-head-sha`. Lives in the shared action, so og-images/docs/build-styles all gain it. Symmetric with the `pull_request` path (workflow-level clean skip for the common case, action-level hard-fail only in the residual window). `set -e` matches the sibling step's style; the `if: event == 'push' && expected-head-sha != ''` guard is correct and mutually exclusive with the PR-path check.
- **`visual.yml` consistency gap → fixed.** Now passes `expected-head-sha: ${{ steps.tip.outputs.tip_sha }}`. Confirmed this is the *correct* SHA to inject: `tip_sha` is the gate-validated value, not the `pr` checkout's HEAD (which can be newer), so a self-computed HEAD would have let a co-advancing branch slip a stale screenshot push through.

Input descriptions for `head-branch` / `expected-head-sha` were updated to match the new dual-path behavior. YAML valid on both changed files; no new correctness, security, or regression issues.

### Open design question (replied on `docs.yml:75`)
Engaged with "Shouldn't it be abstracted into `commit-or-pr`?" — yes, the fetch/compare logic is now triplicated and worth consolidating. The one real blocker is skip-vs-fail semantics (a `uses:` action can't make its caller skip), and the firm constraint is that `expected-head-sha` must stay an optional input because `visual.yml` injects an API-gate SHA, not a working-dir HEAD. Recommended landing it as a focused follow-up (CI-only, no changeset) rather than expanding this PR. Details on the thread.

### Status
No blocking issues. The incremental commit is correct and addresses all prior feedback; the remaining item is the optional abstraction follow-up.

.github/workflows/docs.yml
diegohaz: Fixed in 2dc40666b - the branch-tip guard now lives in `commit-or-pr`: it verifies the branch before creating a maintenance PR or committing to a PR branch, skips cleanly when the branch already advanced, and keeps `expected-head-sha` as an explicit override for workflows such as `visual`.

diegohaz:
**Fixed**

- 2dc40666b (https://github.com/ariakit/ariakit/pull/6470#discussion_r3477137288): Centralized the generated writeback branch-tip guard in `commit-or-pr`, removed the duplicated `Verify branch tip` steps from docs/build-styles/og-images, and preserved the explicit visual SHA override.

.github/workflows/docs.yml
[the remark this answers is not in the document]
diegohaz: Update for head `2dc40666` ("Centralize generated writeback guards"): the dedicated `Verify base branch` step I referenced above was **removed** in this commit, so that earlier note is now stale. The push path is instead guarded by the unified `Verify branch tip` step that now lives at the top of `commit-or-pr` — it computes the expected SHA (input, or `HEAD` by default), fetches `origin/$HEAD_REF`, and on a mismatch sets `is_current=false` and exits 0, so `Create PR` (gated on `is_current == 'true'`) cleanly skips instead of publishing a stale maintenance PR.

So the original finding is still addressed for the common case, just via clean-skip rather than a hard refuse. The one behavioral delta worth being explicit about: there's no longer a re-fetch *immediately before* `create-pull-request`, so a branch that advances in the few-step window between `Verify branch tip` and `Create PR` would still publish a maintenance PR built from the slightly-older `main`. That's acceptable here — the push target is the dedicated, force-updated bot PR (`chore/update-*`), which the next push regenerates — and it's the long-standing behavior the explicit re-check had briefly tightened. The `pull_request` path keeps its residual refuse in `Push to PR head branch`, where a stale rebase onto a contributor branch would actually matter.

.github/workflows/docs.yml
diegohaz: Re-review (head `2dc40666`): verified the centralization, and it lands the shape cleanly. Walking the result:

- `Verify branch tip` is now the first step of `commit-or-pr`, defaults `expected_sha` to `git rev-parse HEAD` when the input is empty, and on advance sets `is_current=false` + `exit 0`. Every writeback step (`Create GitHub App token`, `Configure git`, `Create PR`, `Commit changes`) is gated on `is_current == 'true'`, so an advanced branch is now a uniform green skip across both events. og-images/docs/build-styles correctly drop their own step and the `expected-head-sha` arg — for those three the checkout HEAD *is* the generation commit, so the default is right.
- The `expected-head-sha` override is preserved exactly where it's needed: `visual.yml` still injects `tip_sha`, and the new comment explains why (the `pr` checkout can resolve to a newer tip). `Push to PR head branch` now reads `steps.branch.outputs.expected_head_sha`, so the value flows from one source of truth.
- Nice side effect: because `expected_head_sha` is always populated now, the residual refuse in `Push to PR head branch` (`[ -n "$EXPECTED_HEAD_SHA" ]`) always runs for the PR path — and `visual` on a PR went from hard-fail to clean-skip when its branch advanced, which it never had before.

One trade-off to be conscious of (same as I noted on the Codex thread): dropping `Verify base branch` means the push path no longer re-fetches right before `create-pull-request`; it relies on the early skip plus the self-healing bot PR. That's the right call given the push target, and the PR path keeps its residual re-check, so the asymmetry is intentional and fine.

Net: this is the consolidation worth making, and the visual asymmetry is handled correctly via the optional input. No further changes needed from my side.

diegohaz (commented):
## Re-review summary

**Type:** Re-review. **Baseline:** my prior review at `9295036`; incremental diff `9295036..2dc40666` (commit "Centralize generated writeback guards"). **Scope:** all 5 writeback files — `commit-or-pr/action.yml`, `docs.yml`, `build-styles.yml`, `og-images.yml`, `visual.yml`.

This commit implements the abstraction discussed on the `docs.yml:75` thread. Reviewed in full; it's correct and matches the recommended shape.

### Verified
- **Unified guard in `commit-or-pr`.** New `Verify branch tip` step computes `expected_sha` (input, else `HEAD`), fetches `origin/$HEAD_REF`, and on advance sets `is_current=false` + `exit 0`. All writeback steps (`token`, `Configure git`, `Create PR`, `Commit changes`) are gated on `is_current == 'true'`, giving a uniform green clean-skip for both `push` and `pull_request`.
- **Per-workflow steps removed correctly.** og-images/docs/build-styles drop their `Verify branch tip` step + `expected-head-sha` arg + `is_current` gate. For those, checkout HEAD == generation commit, so the action's default-to-HEAD is correct.
- **Override preserved for the one case that needs it.** `visual.yml` still passes `tip_sha` (with a new clarifying comment); `Push to PR head branch` now sources `EXPECTED_HEAD_SHA` from `steps.branch.outputs.expected_head_sha`. Single source of truth.
- **No bugs found:** `expected_head_sha` is only consumed when `is_current=true` (Push is reachable only via `did_commit`, itself gated on `is_current`); the PR-path residual refuse now always runs (input always populated); `set -e` matches sibling steps; fork PRs are still excluded upstream by `WRITEBACK_ALLOWED`. YAML valid on all five files.

### Behavioral delta to flag (acceptable)
The dedicated `Verify base branch` push re-check added in `9295036` (for the Codex finding) is removed; the push path now relies on the early clean-skip + the self-healing bot PR rather than a re-fetch immediately before `create-pull-request`. Right trade-off given the push target; the `pull_request` path keeps its residual refuse where a stale rebase would actually matter. Noted on the Codex and abstraction threads so the relaxation is conscious (and my earlier "Verify base branch" note there is now corrected, since that step no longer exists).

### Status
No blocking issues and no new actionable findings. The centralization is a net simplification (~50 lines of duplicated YAML removed) that preserves correctness and even upgrades `visual` from hard-fail to clean-skip. No changeset needed (CI-only).

PR #6476, opened 2026-06-25
Preserve synced prices during cache sweep

diegohaz:
## Motivation

`syncPrices` writes every validated or newly created Plus price into KV, then immediately lists cached prices and deletes entries whose ids are not present in the active Stripe price snapshot captured at the start of the run.

That snapshot does not include prices created during the same sync. If KV list metadata already reflects the new write, the cleanup sweep can delete the fresh cache entry it just wrote. KV list metadata can also lag behind a write, so a key that was re-cached under a new active price id can still list old metadata and get deleted by key.

## Solution

Move the cleanup decision into a small `getCachedPriceDeletionReason` helper that first preserves keys synced in the current run:

```ts
if (syncedPrices.has(price.key)) return null;
```

The sweep still deletes invalid cached keys and cached prices that are no longer active in Stripe, but it no longer deletes keys that were validated or recreated moments earlier. Added focused unit coverage for the synced-key guard, active price retention, invalid-key deletion, and inactive price deletion.


app/src/actions/admin-price-cache.test.ts
```diff
@@ -0,0 +1,75 @@
+/**
```
diegohaz: No need to test this.

app/src/actions/admin-price-cache.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type CachedPriceDeletionReason = "invalid-key" | "inactive";
+
+interface GetCachedPriceDeletionReasonParams {
+  price: PriceData;
+  activePrices: Array<Pick<Stripe.Price, "active" | "id">>;
+  syncedPrices: ReadonlyMap<string, PriceData>;
+}
+
+export function getCachedPriceDeletionReason({
```
diegohaz: I believe this function can be moved to `admin.ts` without exporting.

app/src/actions/admin-price-cache.test.ts
diegohaz: Fixed in 1d37c83f - removed the admin price cache unit test as requested.

app/src/actions/admin-price-cache.ts
diegohaz: Fixed in 1d37c83f - removed the exported helper module and kept the synced-key guard directly in syncPrices.

diegohaz:
**Fixed**

- 1d37c83f ([reply](https://github.com/ariakit/ariakit/pull/6476#discussion_r3476798696)): Removed the `admin-price-cache.test.ts` unit test as requested.
- 1d37c83f ([reply](https://github.com/ariakit/ariakit/pull/6476#discussion_r3476798700)): Removed the exported `admin-price-cache.ts` helper and kept the synced-key guard directly inside `syncPrices`.

PR #6491, opened 2026-06-25
Fix source plugin flattened file cache

diegohaz:
## Motivation

The docs app source plugin caches generated flattened file entries by absolute file id and content hash. Each cached entry also stored the files record key returned by `normalizeFilename(file.id, baseDir)`. For non-lib files, that key depends on the current source base directory, so loading the same file from a parent Next.js convention page and then from its own route could reuse a stale relative filename.

## Solution

Cache only the generated `SourceFile` content and metadata, and recompute the flattened filename for each call. This preserves cross-entry reuse for import rewriting and formatting while keeping the `baseDir`-dependent key fresh.

Added regression coverage that creates temporary parent and nested Next.js convention `page.tsx` files, loads the parent source payload before the nested page payload, and asserts the nested page is keyed as `nested/page.tsx` in the parent payload and `page.tsx` in its own payload.

### DIFF
--- a/app/src/lib/source-plugin-fixtures/nested/page.tsx
+++ b/app/src/lib/source-plugin-fixtures/nested/page.tsx
@@ -1,0 +1,3 @@
+export default function Page() {
+  return <div>Nested</div>;
+}
--- a/app/src/lib/source-plugin-fixtures/nested/page.tsx
+++ b/app/src/lib/source-plugin-fixtures/nested/page.tsx
@@ -1,3 +1,0 @@
-export default function Page() {
-  return <div>Nested</div>;
-}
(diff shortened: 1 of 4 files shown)


app/src/lib/source-plugin-fixtures/nested/page.tsx
```diff
@@ -0,0 +1,3 @@
+export default function Page() {
```
diegohaz: Can we test this in a simpler way without creating these fixture files? Otherwise, I think we should just remove this test.

diegohaz: Fixed in cc61e44bb - replaced the committed source-plugin fixture files with temporary parent and nested `page.tsx` files created inside the regression test, while preserving the parent-then-nested source payload coverage.

diegohaz:
**Fixed**

- cc61e44bb ([reply](https://github.com/ariakit/ariakit/pull/6491#discussion_r3477217087)): Replaced the committed source-plugin fixture files with temporary files created inside the regression test while preserving parent-then-nested source payload coverage.

PR #6502, opened 2026-06-25
Deduplicate press shortcut registration

diegohaz:
## Motivation

`press.ts` duplicated the same shortcut key list in two places: the top-level `press.*` assignments and the `createKeyShortcuts` factory used by `press.down` and `press.up`. Adding or fixing a shortcut required keeping both lists in sync, including special cases such as `ShiftTab` and `Space`.

## Solution

Attach the top-level `press` shortcuts with the same `createKeyShortcuts(createPress)` factory used by `press.down` and `press.up`, so the runtime shortcut map is defined once.

The public `press` export remains a function declaration so the generated docs keep the concise `function press(...)` signature. A type-only namespace preserves the shortcut property types for `press`, `press.down`, and `press.up` after the runtime `Object.assign` call.


diegohaz:
It's still duplicating (with TypeScript), now with more code. Closing.

PR #6513, opened 2026-06-25
Fix admin action gate timing

diegohaz:
## Motivation

The app middleware detects admin actions, but with Clerk enabled it checked `isAdmin(context)` only after `getClerkMiddleware()(context, next)` had fully resolved. Astro action handlers run inside `next()`, so an unauthorized admin action could execute its handler side effects before the middleware replaced the response with `401 Unauthorized`.

Current admin handlers still perform their own authorization checks, but the central middleware gate read as if admin actions were already blocked before handler execution.

## Solution

Wrap `next` for admin actions inside the Clerk middleware chain. Clerk decorates `context.locals` before invoking `next`, so the wrapped callback can call `isAdmin(context)` with auth context available and return `unauthorized()` before the action handler runs.

The handler-level checks remain in place as defense in depth.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- `git diff --check`


app/src/test-utils/astro-actions.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
diegohaz: No need for this file.

app/src/middleware.test.ts
```diff
@@ -0,0 +1,65 @@
+/**
```
diegohaz: No need to test.

vitest.config.ts
```diff
@@ -80,6 +80,11 @@ if (testLoader === "react") {
 export default defineConfig({
   root: rootDir,
   plugins,
+  resolve: {
+    alias: {
+      "astro:actions": join(rootDir, "app/src/test-utils/astro-actions.ts"),
+    },
+  },
```
diegohaz: No need to test.

app/src/test-utils/astro-actions.ts
diegohaz: Fixed in e2fc39172 - removed the unnecessary `astro:actions` test stub.

app/src/middleware.test.ts
diegohaz: Fixed in e2fc39172 - removed the middleware unit test as requested.

vitest.config.ts
diegohaz: Fixed in e2fc39172 - removed the Vitest alias that only supported the deleted middleware test.

diegohaz:
**Fixed**

- e2fc39172 ([reply](https://github.com/ariakit/ariakit/pull/6513#discussion_r3478144644)): Removed the unnecessary `astro:actions` test stub.
- e2fc39172 ([reply](https://github.com/ariakit/ariakit/pull/6513#discussion_r3478144645)): Removed the middleware unit test as requested.
- e2fc39172 ([reply](https://github.com/ariakit/ariakit/pull/6513#discussion_r3478144649)): Removed the Vitest alias that only supported the deleted middleware test.

PR #6525, opened 2026-06-26
Deduplicate Tailwind section label helpers

diegohaz:
## Motivation

The Ariakit Tailwind sandbox Chrome test derived section labels in three serialized browser callbacks: computed-style snapshots, axe color-contrast paths, and the preview readiness check. Keeping the same fallback logic duplicated made it easy for snapshot keys, violation paths, and readiness behavior to drift apart.

## Solution

Install one browser-side `window.getSectionLabel` helper with `page.evaluate` from `waitForPreviewReady`, then reuse it from the serialized callbacks. The shared helper keeps the stricter readiness semantics by returning `null` when a section label cannot be resolved.

`extractComputedCSS` preserves the previous empty-string fallback with `?? ""`, while `formatColorContrastViolations` keeps filtering falsy labels from axe paths. This is test-maintenance only, so no changeset is included.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `APP_PORT=4321 pnpm -F app run test-chrome ariakit-tailwind`

### DIFF
--- a/app/src/sandbox/ariakit-tailwind/test-chrome.ts
+++ b/app/src/sandbox/ariakit-tailwind/test-chrome.ts
@@ -25,6 +25,44 @@ interface AxeCheckResult {
   };
 }
 
+declare global {
+  interface Window {
+    /** Installed by `installSectionLabelHelpers` via `page.evaluate`. */
+    getSectionLabel: (section: Element) => string | null;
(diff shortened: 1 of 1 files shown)


diegohaz:
@codex

PR #6535, opened 2026-06-26
Use form item context in form hooks

diegohaz:
## Motivation

`useFormItemContext` centralizes the store/context fallback, form invariant, and name stringification that named form items need. `FormCheckbox`, `FormRadio`, `FormPush`, and `FormRemove` still repeated that setup inline:

```tsx
const context = useFormContext();
store = store || context;
const name = String(nameProp);
```

That also left dead nullable guards like `store?.setValue(...)` and `if (!store) return` after the invariant had already asserted the store exists.

## Solution

Migrate those four hooks to `useFormItemContext`, matching the existing `FormInput` pattern:

```tsx
const { store: form, name } = useFormItemContext({
  store,
  name: nameProp,
  component: "FormInput",
});
```

The hooks now use the returned non-null `form` binding for checkbox, radio, collection, and array operations, so the redundant optional chains and early return can be removed. This preserves hook order and keeps the same component-specific invariant messages.

No changeset is included because this is an internal refactor with no public API or behavior change.

## Testing

- `corepack pnpm lint-fix`
- `corepack pnpm tsc`
- `corepack pnpm test-react examples/form-radio/test.ts examples/form-checkbox-integer-like-name/test.ts app/src/sandbox/form-push-6293/test.ts app/src/sandbox/form-push-remove-6219/test.ts`
- `corepack pnpm -F @ariakit/react-components run build`
- `git diff --check`

### DIFF
--- a/packages/ariakit-react-components/src/form/form-checkbox.tsx
+++ b/packages/ariakit-react-components/src/form/form-checkbox.tsx
@@ -5,12 +5,11 @@ import {
   memo,
 } from "@ariakit/react-utils";
 import type { Props } from "@ariakit/react-utils";
-import { invariant } from "@ariakit/utils";
 import type { ElementType } from "react";
 import { useCheckboxStore } from "../checkbox/checkbox-store.ts";
 import type { CheckboxOptions } from "../checkbox/checkbox.tsx";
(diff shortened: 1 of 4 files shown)


diegohaz:
@codex

PR #6538, opened 2026-06-26
Fix collapsible sidebar open state

diegohaz:
## Motivation

The shared `Sidebar` example helper always passed `open` to the `Dialog` branch used by collapsible sidebars. In a `SidebarProvider`, that controlled prop pinned the provider dialog store open, so `SidebarToggle`, Escape, and outside interactions could not collapse the sidebar.

```tsx
<SidebarProvider>
  <SidebarToggle />
  <Sidebar />
</SidebarProvider>
```

The same branch also used the default dialog auto-focus behavior on desktop, even though the desktop sidebar is non-modal.

## Solution

Track whether a `Sidebar` is inside `SidebarProvider` separately from any ambient Ariakit dialog context. Provider-driven sidebars now leave `open` undefined so the provider store controls visibility, while standalone sidebars still force `open={true}` when explicitly rendered through the dialog branch.

```tsx
<Dialog
  open={hasSidebarProvider ? undefined : true}
  autoFocusOnShow={isMobile}
/>
```

`collapsible` now defaults to the presence of `SidebarProvider`, not any surrounding `Dialog` context. This is app example helper code only, so no changeset is needed.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`


app/src/examples/_lib/ariakit/sidebar.react.tsx
```diff
@@ -57,7 +58,8 @@ export function Sidebar({
         modal={isMobile}
         hideOnEscape={isMobile}
         hideOnInteractOutside={isMobile}
-        open
+        open={hasSidebarProvider ? undefined : true}
```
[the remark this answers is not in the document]
diegohaz: Confirmed the mechanism: `SidebarProviderContext` propagates through any nested provider, so a `<Sidebar collapsible>` rendered inside a nested `ak.DialogProvider` that is itself under a `SidebarProvider` keeps `hasSidebarProvider === true`, leaves `open` undefined, and binds to the *nested* store (`useDialogStore({ store: storeProp || context, open: openProp })` in `dialog.tsx` resolves to the nearest `DialogProvider` store). So it would follow/close with the unrelated dialog instead of staying forced open.

That said, I don't think this needs a change here:

- **Not exercised.** This is example/sandbox helper code. The only real consumer, `examples/disclosure/navigation`, renders `<Sidebar>` with no provider at all. In the new sandbox the standalone case is a *sibling* of `SidebarProvider` (under a separate `ak.DialogProvider`), where `hasSidebarProvider` is correctly `false`, so it renders as `ak.Role` and the `open` decision never applies. No example produces the provider-then-nested-dialog nesting.
- **The old code was also wrong for this shape**, just differently: `open={true}` was applied to that same nearest (nested) store, pinning the unrelated dialog open. So this isn't a regression — both versions mishandle the contrived nesting.
- **A robust fix has its own cost.** To make the standalone branch ignore ambient context you'd give it its own store (e.g. `store={hasSidebarProvider ? undefined : useDialogStore({ open: true })}`), which adds an always-on hook to a demo helper and reintroduces a pin-open quirk for standalone *mobile* sidebars (controlled `open: true` can't be dismissed) — not worth it for a path no example hits.

So: real but unexercised limitation, acceptable trade-off for this helper. Worth revisiting only if a real example ever needs a forced-open standalone sidebar nested inside another dialog under a `SidebarProvider`.

diegohaz (commented):
## Review summary

**Initial review** (comment-only). Baseline: merge-base `f49ec71` → PR head `d410958`. Scope: the changed `app/src/examples/_lib/ariakit/sidebar.react.tsx` helper plus the new `app/src/sandbox/sidebar-collapsible/` sandbox and its `test.ts` / `test-browser.ts`, expanded to every consumer of the helper and the dialog store-resolution path.

### Verdict
No blocking issues. The fix is correct and well-scoped, and the test coverage maps directly onto the regression. The root cause checks out against the source: `Dialog` resolves `useDialogStore({ store: storeProp || context, open: openProp })`, so under a `SidebarProvider` the old hardcoded `open` synced `true` into the provider store and pinned it open; switching to `open={hasSidebarProvider ? undefined : true}` lets the provider store drive visibility again, while standalone sidebars (no context) still get a fresh forced-open store. Splitting the new boolean `SidebarProviderContext` from ambient `useDialogContext()` is the right call — and it must stay separate, since `Sidebar` itself re-provides `SidebarContext` for `side` inheritance, so the provider flag can't ride along on that context.

### What I checked
- **Consumers / regressions.** The only real consumer, `examples/disclosure/navigation`, renders `<Sidebar>` with no provider and no `collapsible`, so it stayed `collapsible = false` → `ak.Role` before and after (no change). The `collapsible` default move from `!!dialog` to `hasSidebarProvider` only affects sidebars actually inside a `SidebarProvider`, which is the intended behavior.
- **`autoFocusOnShow={isMobile}`.** Correct: desktop sidebars are non-modal and shouldn't steal focus; mobile modal sidebars still autofocus. Both `open` and `autoFocusOnShow` are placed before `{...props}`, so callers can still override.
- **Tests.** Coverage exercises provider toggle collapse/expand, desktop focus preservation (toggle stays focused after expand, validating the new `autoFocusOnShow`), mobile Escape, mobile outside-click, and standalone-inside-another-dialog. The happy-dom `test.ts` mirrors the desktop path with the `q.dialog.hidden.lazy(...)` unmount poll. No global/`__testState` harness — behavior is observed through visible UI, per repo conventions. No changeset needed (app example/sandbox code only); no `preview.json` needed (React inferred from `index.react.tsx`).

### Commands run
- `pnpm test-react app/src/sandbox/sidebar-collapsible/test.ts` → **2 passed**.
- `tsc -b` → changed files are type-clean (remaining errors are all in unrelated `app/src/lib/*` / `app/src/pages/*` files from missing Astro-generated content types in a fresh worktree). `pnpm tsc`'s `astro check` step failed on an unrelated vite `optimizeDeps` cache miss (`.vite-check/deps_ssr/chunk-*.js`), an environment flake, not a code issue.

### Non-blocking note
- Replied on the Codex thread re: a `<Sidebar collapsible>` nested inside a nested `DialogProvider` *and* a `SidebarProvider` binding to the wrong store. Real but unexercised by any example, the old code mishandled it too (it pinned the nested store open), and a robust fix carries its own trade-offs — acceptable for this helper.

### Not validated locally
- The Playwright `test-browser.ts` suite (mobile Escape / outside-click, cross-browser) — I relied on the author's reported run across chrome/firefox/safari rather than starting a worktree dev server. The happy-dom subset and the desktop logic are confirmed locally.

app/src/sandbox/sidebar-collapsible/index.react.tsx
```diff
@@ -0,0 +1,35 @@
+import * as ak from "@ariakit/react";
```
diegohaz: No need for the sandbox nor tests here

diegohaz: Fixed in 31708ee71 - removed the `sidebar-collapsible` sandbox and tests from the PR.

diegohaz:
**Fixed**

- 31708ee71 ([reply](https://github.com/ariakit/ariakit/pull/6538#discussion_r3478718247)): Removed the `sidebar-collapsible` sandbox and tests per review feedback, then updated the PR description to match the trimmed scope.

PR #6539, opened 2026-06-26
Prevent form store prototype path corruption

diegohaz:
## Motivation

The form store accepts field names as dot-delimited paths and uses each segment as an object key for values, errors, and touched state. A `__proto__` or `constructor` segment could previously flow into reads and writes: object copies could replace the returned state object's prototype, inherited properties could be resolved, and protected keys from JSON-defined form data could affect derived state even though field APIs could not address them.

This mostly affects dynamic or server-driven forms where field names or form state can come from external data.

## Solution

Normalize runtime array path segments before processing them, then treat `__proto__` and `constructor` as protected path segments. Reads return the default value, writes no-op before creating nested containers, and state-copy helpers skip protected own keys so they cannot replace object prototypes.

The same protected-key rule now applies when deriving touched state and when scanning errors for validation messages, keeping `valid` consistent with what `getError` and `setError` can read or clear.

## Verification

- `corepack pnpm exec vitest packages/ariakit-components/src/form/form-store.test.ts --run`
- `corepack pnpm tsc`
- `corepack pnpm lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form field path handling to ignore unsafe keys like `__proto__` and `constructor`, helping prevent corrupted form state.
  * Fixed updates, touched state, and validation/message handling so protected keys are skipped consistently.
  * Added safeguards for array-style field paths and nested values to keep form data stable and secure.
  * Updated the form-store getter/setter behavior to normalize path segments and avoid prototype-related corruption.

### DIFF
--- a/packages/ariakit-components/src/form/form-store.ts
+++ b/packages/ariakit-components/src/form/form-store.ts
@@ -42,13 +42,6 @@ function isPrototypePathSegment(key: PathSegment) {
   return false;
 }
 
-function hasPrototypePathSegment(path: readonly PathSegment[]) {
-  for (const key of path) {
-    if (isPrototypePathSegment(key)) return true;
-  }
-  return false;
-}
-
 function nextFrame() {
   return new Promise<void>((resolve) => {
     // Browsers pause `requestAnimationFrame` in hidden documents, which would
@@ -92,10 +85,9 @@ function getPathValue<T>(
   defaultValue?: T,
 ): T {
   const [key, ...rest] = path;
-  if (key == null || !values) {
+  if (key == null || !values || isPrototypePathSegment(key)) {
     return defaultValue as T;
   }
-  if (isPrototypePathSegment(key)) return defaultValue as T;
   if (!rest.length) {
     return values[key] ?? defaultValue;
   }
@@ -146,7 +138,7 @@ function set<T extends FormStoreValues | unknown[]>(
   value: unknown,
 ): T {
   const pathKeys = getPath(path);
-  if (hasPrototypePathSegment(pathKeys)) return values;
+  if (pathKeys.some(isPrototypePathSegment)) return values;
   return setPath(values, pathKeys, value);
 }
 
--- a/packages/ariakit-components/src/form/form-store.ts
+++ b/packages/ariakit-components/src/form/form-store.ts
@@ -23,7 +23,31 @@ import { createCollectionStore } from "../collection/collection-store.ts";
 import type { DeepMap, DeepPartial, Names, StringLike } from "./types.ts";
 
 type ErrorMessage = string | undefined | null;
+type PathSegment = string | number;
 const maxArrayIndex = 2 ** 32 - 2;
+
+function getPath(path: StringLike | unknown[]) {
+  if (!Array.isArray(path)) {
+    return String(path).split(".");
+  }
+  return path.map((key) => {
+    if (typeof key === "number") return key;
+    return String(key);
+  });
+}
+
+function isPrototypePathSegment(key: PathSegment) {
+  if (key === "__proto__") return true;
+  if (key === "constructor") return true;
+  return false;
+}
+
+function hasPrototypePathSegment(path: readonly PathSegment[]) {
+  for (const key of path) {
+    if (isPrototypePathSegment(key)) return true;
+  }
+  return false;
+}
 
 function nextFrame() {
   return new Promise<void>((resolve) => {
@@ -41,30 +65,44 @@ function nextFrame() {
 }
 
 export function hasMessages(object: FormStoreValues): boolean {
-  return Object.keys(object).some((key) => {
+  const keys = Object.keys(object);
+  for (const key of keys) {
+    if (isPrototypePathSegment(key)) continue;
     if (isObject(object[key])) {
-      return hasMessages(object[key]);
+      if (hasMessages(object[key])) return true;
+      continue;
     }
-    return !!object[key];
-  });
+    if (object[key]) return true;
+  }
+  return false;
 }
 
 export function get<T>(
   values: FormStoreValues,
-  path: StringLike | string[],
+  path: StringLike | Array<string | number>,
   defaultValue?: T,
 ): T {
-  const [key, ...rest] = Array.isArray(path) ? path : String(path).split(".");
+  const pathKeys = getPath(path);
+  return getPathValue(values, pathKeys, defaultValue);
+}
+
+function getPathValue<T>(
+  values: FormStoreValues,
+  path: PathSegment[],
+  defaultValue?: T,
+): T {
+  const [key, ...rest] = path;
   if (key == null || !values) {
     return defaultValue as T;
   }
+  if (isPrototypePathSegment(key)) return defaultValue as T;
   if (!rest.length) {
     return values[key] ?? defaultValue;
   }
-  return get(values[key], rest, defaultValue);
+  return getPathValue(values[key], rest, defaultValue);
 }
 
-function isArrayIndex(key: string | number) {
+function isArrayIndex(key: PathSegment) {
   if (typeof key !== "string" && typeof key !== "number") return false;
   const stringKey = String(key);
   const index = Number(stringKey);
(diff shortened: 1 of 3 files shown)


packages/ariakit-components/src/form/form-store.ts
```diff
@@ -23,8 +23,32 @@ import { createCollectionStore } from "../collection/collection-store.ts";
...
+  if (key === "constructor") return true;
+  return false;
+}
+
+function hasPrototypePathSegment(path: readonly PathSegment[]) {
+  for (const key of path) {
+    if (isPrototypePathSegment(key)) return true;
+  }
+  return false;
+}
```
diegohaz: **Optional: inline this single-use helper.** `hasPrototypePathSegment` is only called once (in `set`, line 149) and is a thin wrapper over `isPrototypePathSegment`. A plain `Array.prototype.some` reads the same, drops the extra named indirection and the manual loop, and short-circuits identically:

```ts
function set<T extends FormStoreValues | unknown[]>(
  values: T,
  path: StringLike | Array<string | number>,
  value: unknown,
): T {
  const pathKeys = getPath(path);
  if (pathKeys.some(isPrototypePathSegment)) return values;
  return setPath(values, pathKeys, value);
}
```

The `hasPrototypePathSegment` definition can then be deleted. Behavior-preserving and non-blocking — keep the helper if you prefer the explicit name.

packages/ariakit-components/src/form/form-store.ts
```diff
@@ -41,30 +65,44 @@ function nextFrame() {
...
   defaultValue?: T,
 ): T {
-  const [key, ...rest] = Array.isArray(path) ? path : String(path).split(".");
+  const [key, ...rest] = path;
   if (key == null || !values) {
     return defaultValue as T;
   }
+  if (isPrototypePathSegment(key)) return defaultValue as T;
   if (!rest.length) {
     return values[key] ?? defaultValue;
```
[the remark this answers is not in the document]
diegohaz: Context on this one: the `??` predates this PR — the base `get` already returned `values[key] ?? defaultValue`, so it's not a regression introduced here and it's outside this change's prototype-pollution scope.

It also interacts with `removeValue`, which intentionally writes `null` into array slots to keep indices stable (see the `removeValue` JSDoc). Switching to `value === undefined ? defaultValue : value` would make `getValue("items.0")` return that `null` sentinel instead of `undefined` for a removed slot — a deliberate semantics change that should be weighed against the `removeValue` design rather than folded into this security fix. Leaving it as-is here looks correct; a separate PR would be the place to revisit it if the `null`-vs-`undefined` distinction is wanted on reads.

diegohaz (commented):
## Review summary

**Initial review.** Diffed the merge base `f49ec71` against head `faad63c` (the PR's 3 changed files) and inspected the surrounding form-store context.

### Scope
`packages/ariakit-components/src/form/form-store.ts`, its test file, and the changeset. Reviewed the prototype-pollution fix end to end plus the read/write/derive helpers it touches, the exported `get`/`hasMessages` (no external importers — only `createFormStore` is consumed elsewhere), and the interaction with `setValues`/`setErrors`/`setTouched`, `pushValue`/`removeValue`, `submit`/`setAll`, and `valid`/`hasMessages`.

### Validation performed
- Ran `vitest packages/ariakit-components/src/form/form-store.test.ts --run` against the PR head locally → **14/14 pass**.
- Traced both attack vectors by hand:
  - The `__proto__`-key write vector (old `Object.assign({}, values, { ["__proto__"]: ... })` invokes the proto setter on the returned object) is now blocked by the `hasPrototypePathSegment` guard in `set` plus `setObjectValue` copying via direct assignment while skipping proto keys.
  - The `new String("__proto__")` array-segment vector is caught by `getPath` normalizing each segment with `String(...)` before `isPrototypePathSegment` — the "normalizes runtime array path segments" test covers exactly this.
  - Data-side proto keys from JSON-parsed `defaultValues`/`defaultErrors` are skipped in `getPathValue`/`setAll`/`hasMessages`, keeping `touched`/`valid` consistent with what `getError`/`setError` can read.
- Confirmed `prototype` does not need its own guard: the only dangerous chain (`constructor.prototype.x`) is already short-circuited by the `constructor` segment.

### Assessment
The fix is correct and the two-layer design (path guard + data-key skip) is justified — both layers are load-bearing, and an alternative of sanitizing all input data at the `setValues`/`setErrors`/`setTouched` boundary would add eager copying and reshape user data, so the targeted skip-where-it-matters approach is the better trade-off. Tests are thorough and the changeset follows repo conventions (numeric prefix, three patch bumps, reference link).

### Findings
No correctness, compatibility, security, or performance issues.
- One **optional** behavior-preserving simplification posted inline: `hasPrototypePathSegment` is a single-use wrapper that reads fine as an inline `pathKeys.some(isPrototypePathSegment)`.
- Replied to the existing CodeRabbit `??` thread with context: that nullish-coalescing predates this PR, is out of scope for the security fix, and interacts with `removeValue`'s intentional `null` sentinel — so it's a separate decision rather than a regression here.

### Not fully validated
Did not re-run the full `pnpm tsc` / `pnpm lint` suites locally (the PR description reports both green); the change is type-local and the targeted unit suite passed.

packages/ariakit-components/src/form/form-store.ts
```diff
@@ -41,30 +65,44 @@ function nextFrame() {
...
+  values: FormStoreValues,
+  path: PathSegment[],
   defaultValue?: T,
 ): T {
-  const [key, ...rest] = Array.isArray(path) ? path : String(path).split(".");
+  const [key, ...rest] = path;
   if (key == null || !values) {
     return defaultValue as T;
   }
+  if (isPrototypePathSegment(key)) return defaultValue as T;
```
diegohaz: I think we can do `key == null || !values || isPrototypePathSegment(key)` here.

packages/ariakit-components/src/form/form-store.ts
diegohaz: Fixed in 3204f798f - removed the single-use helper and now use `pathKeys.some(isPrototypePathSegment)` in `set`.

packages/ariakit-components/src/form/form-store.ts
diegohaz: Fixed in 3204f798f - folded the protected segment check into the existing `getPathValue` early return.

diegohaz:
**Fixed**

- 3204f798f (https://github.com/ariakit/ariakit/pull/6539#discussion_r3478729218): Removed the single-use `hasPrototypePathSegment` helper and inlined the `pathKeys.some(isPrototypePathSegment)` check.
- 3204f798f (https://github.com/ariakit/ariakit/pull/6539#discussion_r3478729228): Folded the protected segment check into the existing `getPathValue` early return.

diegohaz (commented):
## Re-review summary

**Re-review.** Incremental baseline = my prior reviewed head `faad63c`; current head `3204f798f`. Reviewed only the one new commit (`Simplify form store prototype guards`) plus its surrounding context.

### Changes since last review
The single new commit applies both simplifications from the prior round, and nothing else:
- Removed the single-use `hasPrototypePathSegment` helper; `set` now uses `pathKeys.some(isPrototypePathSegment)` inline.
- Folded the protected-segment check into the existing `getPathValue` early return: `if (key == null || !values || isPrototypePathSegment(key))`.

Both are behavior-preserving (the merged guard returns `defaultValue` for the same three conditions, and `some` short-circuits exactly like the old loop). No change to the test file or changeset.

### Validation performed
- Ran `vitest packages/ariakit-components/src/form/form-store.test.ts --run` against head `3204f798f` → **14/14 pass**.
- Re-traced the `getPathValue` guard ordering: `key == null` / `!values` still short-circuit before `isPrototypePathSegment(key)`, so the merge introduces no ordering hazard.

### Prior threads
- The `hasPrototypePathSegment` inline suggestion and the `getPathValue` guard merge are both addressed in `3204f798f` and resolved — verified the committed code matches.
- The CodeRabbit `??` thread is resolved; CodeRabbit agreed it predates this PR and is coupled to the `removeValue` `null` sentinel, so it stays out of scope here.

### Findings
No new findings. The simplification is correct and the PR remains in good shape.

PR #6543, opened 2026-06-26
Preserve table row column alignment

diegohaz:
## Motivation

The shared app example `Table` rendered declarative row cells by iterating each row's own object entries and skipping nullish values. A row such as `{ name: "Ada", age: null, city: "London" }` would drop the `age` cell completely, shifting later cells under the wrong header.

Rows also depended on each object's key insertion order, so two rows with the same keys in different orders could render different column orders.

## Solution

Derive a stable column list from declarative rows in the same order the table renders row groups: head rows first, then body rows, then foot rows. Each row now renders from that same list. Nullish values pass `{ children: null }` into `createRender`, which preserves an empty cell without falling back to the default `children: key` label.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`


app/src/sandbox/table-row-columns/index.react.tsx
```diff
@@ -0,0 +1,17 @@
+import { Table } from "#app/examples/_lib/ariakit/table.react.tsx";
```
diegohaz: No need for a sandbox

diegohaz: Fixed in a732e56c1 - removed the dedicated table-row-columns sandbox and kept the alignment fix in the shared Table helper.

diegohaz:
**Fixed**

- a732e56c1 ([reply](https://github.com/ariakit/ariakit/pull/6543#discussion_r3479161867)): Removed the dedicated `table-row-columns` sandbox and kept the alignment fix in the shared `Table` helper. The PR description was also updated to remove stale sandbox test references.

PR #6546, opened 2026-06-26
Optimize form control store subscriptions

diegohaz:
## Motivation

`useFormControl` read form labels, errors, descriptions, and invalid state through separate store subscriptions. In larger forms, that created extra listener work and repeated item scans during field updates.

## Solution

Consolidate those reads behind one `useStoreStateObject` subscription. The label, error, and description id lookup now scans the form `items` array only when that array changes, while the hook still returns primitive id values so unrelated item updates do not re-render controls with unchanged ARIA references.

This also adds a patch changeset for `@ariakit/react-components` and `@ariakit/react`.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test-react examples/form/test.ts examples/form-radio/test.ts examples/form-select/test.ts examples/form-valid/test.ts examples/form-callback-queue/test.ts examples/form-checkbox-integer-like-name/test.ts`
- `pnpm build`

### DIFF
--- a/.changeset/300-form-control-subscriptions.md
+++ b/.changeset/300-form-control-subscriptions.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`FormControl`](https://ariakit.com/reference/form-control) and components built on it, such as [`FormInput`](https://ariakit.com/reference/form-input), [`FormCheckbox`](https://ariakit.com/reference/form-checkbox), and [`FormRadio`](https://ariakit.com/reference/form-radio), to avoid redundant form store subscriptions and item lookups while fields update.
--- a/packages/ariakit-react-components/src/form/form-control.tsx
+++ b/packages/ariakit-react-components/src/form/form-control.tsx
@@ -1,5 +1,5 @@
 import type { StringLike } from "@ariakit/components/form/types";
-import { useStoreState } from "@ariakit/react-store";
+import { useStoreStateObject } from "@ariakit/react-store";
 import {
   useBooleanEvent,
   useEvent,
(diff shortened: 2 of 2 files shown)


diegohaz (commented):
**Review type:** initial · **Baseline:** merge-base `52b044ed` → head `ed47db0d` · **Scope:** `packages/ariakit-react-components/src/form/form-control.tsx`, `.changeset/300-form-control-subscriptions.md`, plus surrounding context in `@ariakit/react-store` (`useStoreStateObject`) and the form store item types.

No substantive findings — this is a clean, correctly-scoped optimization. Comment-only review.

### What I verified

- **Semantics preserved.** `getControlItemIds` reproduces the old per-type `.find()` "first match wins" behavior: each `*Id` is assigned only while still `undefined`, so the first `label`/`error`/`description` for a given `name` wins regardless of array order, and the early `break` is a pure short-circuit once all three are found. Walked several interleaved orderings (e.g. `[error, label, error, description]`) to confirm the `else if` chain never assigns the wrong type or skips one.
- **The optimization is real and matches the stated motivation.** Four `useStoreState` subscriptions collapse to one `useStoreStateObject` subscription. The three id selectors share a single `items` scan via the memoized `getItemIds` closure, which also skips the scan entirely when `state.items` is unchanged — so per-keystroke updates (values/touched change, items stable) do zero scans instead of three `.find()` passes. Returning primitive ids (not item objects) means `useStoreStateObject`'s per-key `!==` compare keeps the result object stable when an unrelated item update leaves this field's ids untouched, avoiding re-renders the old object-returning selectors would have triggered.
- **Considered simpler alternatives and rejected them.** Subscribing to `items` directly (`useStoreState(form, "items")` + `useMemo`) would be shorter but re-renders the control on every items change even when this field's ids are unchanged — i.e. it regresses the exact behavior the PR is optimizing. A single combined selector returning one ids object would re-render whenever `items` changes (new object ref) even if the ids are identical. The per-key + memoized-scan approach is the minimal shape that gets both "scan only when items change" and "re-render only when an id/invalid changes."
- **`invalid` selector** still reads `form.getError(name)`/`form.getFieldTouched(name)` and is re-evaluated on every snapshot with per-key boolean comparison — behavior-identical to the prior standalone `useStoreState`.
- **Types** check out: `FormStoreItem` carries required `id: string`, `name: string`, and the `type` union, so the `=== undefined` guards and index access are sound; the non-null `useStoreStateObject` overload infers `state: FormStoreState`. The mutate-during-`getSnapshot` memo cache mirrors the pattern already used inside `useStoreStateObject` itself and is safe under `useSyncExternalStore`.
- **Changeset** is accurate; bumping both `@ariakit/react-components` (source) and `@ariakit/react` (legacy build re-export) as `patch` is correct for a perf-only change.

### Commands run

- `pnpm install` (fresh worktree), then `pnpm test-react examples/form/test.ts examples/form-radio/test.ts examples/form-select/test.ts examples/form-valid/test.ts` → **24/24 passed**, confirming `aria-labelledby` / `aria-describedby` / `aria-invalid` wiring is unchanged.

- Did not re-run `pnpm tsc` or `pnpm build` locally (relying on the author's run + CI); the type reasoning above is static.
- The `ariakito` perf comment's only >10% Chrome deltas (`ariakit-tailwind` toggle, `composite-perf` move-across-items INP) are in sandboxes that don't touch `FormControl` and weren't flagged as significant (rounds disagreed) — unrelated noise, not attributable to this change.

PR #6547, opened 2026-06-26
Reduce menu initial focus rerenders

diegohaz:
## Motivation

`useMenu` subscribed directly to `autoFocusOnShow`, `initialFocus`, `baseElement`, and `renderedItems` only to resolve the initial focus element. When menu items re-register or `renderedItems` is re-sorted, that array can receive a fresh identity even if the resolved focus target stays the same, causing the full `Menu` hook chain to re-render unnecessarily.

## Solution

Resolve the initial focus element inside a single `useStoreState` selector. This keeps the snapshot stable when `renderedItems` changes identity but still points to the same DOM element, while preserving the disabled-auto-focus `undefined` state, unresolved-target `null` state, modal persistent-focus guard, and ref identity updates when the actual target changes.

The `"last"` focus path now scans the rendered items from the end without copying the array, so selector work stays allocation-free on store notifications.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm exec playwright test --project=chrome src/sandbox/menu-initial-focus-render/test-browser.ts`
- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm exec playwright test --project=chrome src/sandbox/menu-4291/test-browser.ts`
- `pnpm build`

### DIFF
--- a/packages/ariakit-react-components/src/menu/menu.tsx
+++ b/packages/ariakit-react-components/src/menu/menu.tsx
@@ -87,10 +87,30 @@ export const useMenu = createHook<TagName, MenuOptions>(function useMenu({
   const [initialFocusRef, setInitialFocusRef] =
     useState<MutableRefObject<HTMLElement | null>>();
 
-  const autoFocusOnShowState = useStoreState(store, "autoFocusOnShow");
-  const initialFocus = useStoreState(store, "initialFocus");
-  const baseElement = useStoreState(store, "baseElement");
-  const items = useStoreState(store, "renderedItems");
+  // Resolve the initial focus element inside a selector so the component
+  // re-renders only when the resolved element changes, not whenever
+  // `renderedItems` gets a new array identity. Returning `undefined` means
+  // auto focus on show is disabled.
+  const initialFocusElement = useStoreState(store, (state) => {
+    if (!state.autoFocusOnShow) return;
+    const isEnabled = (item: (typeof state.renderedItems)[number]) =>
+      !item.disabled && !!item.element;
+    switch (state.initialFocus) {
+      case "first":
+        return state.renderedItems.find(isEnabled)?.element || null;
+      case "last":
+        for (let i = state.renderedItems.length - 1; i >= 0; i -= 1) {
+          const item = state.renderedItems[i];
+          if (!item) continue;
+          if (item.disabled) continue;
+          if (!item.element) continue;
+          return item.element;
+        }
+        return null;
+      default:
+        return state.baseElement;
+    }
+  });
 
   // Sets the initial focus ref.
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


packages/ariakit-react-components/src/menu/menu.tsx
```diff
@@ -87,10 +87,30 @@ export const useMenu = createHook<TagName, MenuOptions>(function useMenu({
...
+      case "first":
+        return state.renderedItems.find(isEnabled)?.element || null;
+      case "last":
+        for (let i = state.renderedItems.length - 1; i >= 0; i -= 1) {
+          const item = state.renderedItems[i];
+          if (!item) continue;
+          if (item.disabled) continue;
+          if (!item.element) continue;
+          return item.element;
+        }
```
diegohaz, line 107: **Minor (optional): reuse `isEnabled` in the `"last"` branch.**

`isEnabled` already encodes the "enabled item" predicate (`!item.disabled && !!item.element`), but the `"last"` loop re-spells those same two checks inline. Reusing the helper keeps a single source of truth for what counts as an enabled item, so a future change to that definition only has to touch one place. It stays allocation-free and behavior-preserving — the `item &&` guard preserves the indexed-access safety, and `item.element` is still `HTMLElement | null`, which the selector's return type already allows:

```suggestion
        for (let i = state.renderedItems.length - 1; i >= 0; i -= 1) {
          const item = state.renderedItems[i];
          if (item && isEnabled(item)) {
            return item.element;
          }
        }
```

Not blocking — the current loop is correct and matches the repo's early-exit style; this is purely a small DRY/maintainability nicety.

diegohaz (commented):
## Review summary

**Initial review** of the full PR (baseline: merge-base of `main` and head `25a1b88`).

This is a clean, well-scoped optimization. Moving the initial-focus resolution into a single `useStoreState` selector is the right approach, and I couldn't find any behavioral regression. One minor, optional DRY suggestion is inline.

### What I verified

- **Behavioral equivalence.** The selector reproduces the old `first` / `last` / `default` resolution exactly, and the three resolved states are preserved: `undefined` (auto-focus disabled), `null` (no resolvable target → fresh ref to `null`), and the modal persistent-focus short-circuit. The `last` manual reverse loop matches the old `[...items].reverse().find(...)` result without the array copy, and avoiding native `Array.prototype.findLast` is consistent with this package's convention (it hand-rolls its own `findLast` in `composite-renderer.tsx`).
- **The key safety property.** `useStoreState` runs the selector inside `useSyncExternalStore`'s `getSnapshot`, which must return a cached/stable value or React loops. The selector only ever returns a stable DOM element reference, `null`, or `undefined` — never a freshly allocated object/array — so `Object.is` caching holds and there's no infinite-loop risk. This is exactly what makes the "stable snapshot when `renderedItems` changes identity" optimization work.
- **Production safety.** `invariant(store, …)` is an `asserts condition` that throws in all environments when `store` is falsy, so the selector's direct `state.autoFocusOnShow` / `state.renderedItems` access can't crash on an undefined state — no regression versus the old key-based reads that tolerated `undefined`.
- **Effect deps.** `[modal, initialFocusElement]` is correct; `store` was rightly dropped since the effect body no longer references it, and there are no missing deps.
- **Test is a real regression test.** Pre-fix, `useMenu` subscribed to `renderedItems`, so the "Refresh rendered items" button (`setState("renderedItems", (items) => [...items])`) forced a `useMenu` re-render — the `menuRenders` assertion would fail on the old code and passes on the new code. Test plumbing checks out: `q.status(...)` resolves the `<output>` (implicit `status` role) by `aria-label`, the `@ariakit/react-components/menu/menu` internal subpath is a real export and a well-established app-sandbox pattern, and `flushFrames`/`withFramework` exist as used.
- **Changeset** is present and correctly scoped (`@ariakit/react` + `@ariakit/react-components`, `patch`).

### Findings

- One **minor/optional** inline comment: reuse the already-extracted `isEnabled` predicate in the `"last"` loop to avoid duplicating the enabled-item condition. Non-blocking.

- I did not run `pnpm tsc` / `pnpm build` or the Playwright browser test in this worktree (the PR reports running all of them, plus the `menu-4291` test). The inline suggestion is behavior-preserving and type-safe by inspection.
- Optional: a happy-dom `test.ts` duplicate would let this run faster/more often, but menu-open + focus + rAF have known happy-dom divergences, so it may not reproduce reliably — I'd leave that to the author's judgment rather than treat it as required.

packages/ariakit-react-components/src/menu/menu.tsx
diegohaz: Fixed in 9548a6001 - reused the existing `isEnabled` predicate in the `"last"` branch while keeping the reverse scan allocation-free.

diegohaz:
**Fixed**

- 9548a6001 (https://github.com/ariakit/ariakit/pull/6547#discussion_r3479210334): Reused the existing `isEnabled` predicate in the `"last"` branch while keeping the reverse scan allocation-free.

PR #6566, opened 2026-06-28
Make perf comparisons more reliable

diegohaz:
## Motivation

The perf workflow can report significant Chrome regressions for changes that do not affect the measured code. For example, ariakit/ariakit#6565 only updated `stripe`, but the perf comment reported unrelated `composite-perf` and `ariakit-tailwind` regressions.

Two things made those false positives easier to publish:

- Every paired CI round ran `baseline -> current`, so same-runner drift could consistently look like a current-side regression.
- `perf-compare` only checked median direction across rounds and ignored the raw iteration samples already persisted by Chrome perf tests.

## Solution

Run the existing paired rounds as a crossover comparison without adding more work:

```text
round 1: baseline -> current
round 2: current -> baseline
```

The same alternation applies to confirmation rounds and Node benchmarks, preserving the existing number of CI runs.

`perf-compare` now also requires each agreeing round to have raw-sample support in the aggregate direction. Displayed medians stay the same, but a warning icon is only added when the median exceeds the threshold and the raw samples support that direction. This keeps noisy overlapping samples visible in the detailed table without promoting them to top-level regressions.

## Verification

- `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm lint-fix`
- `pnpm lint`
- `pnpm tsc`
- `git diff --check`

### DIFF
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -25,8 +25,8 @@ const RESULTS_DIR = path.join(process.cwd(), ".perf-results");
 const PROFILE_LIMIT = 10;
 const THRESHOLD_PERCENT = 10;
 const MIN_SIGNIFICANT_DELTA_MS = 5;
-// Require at least 75% of pairwise raw-sample deltas to support the median
-// direction before a round contributes to significance.
+// Require at least 75% of pooled pairwise raw-sample deltas to support the
+// median direction before a browser comparison becomes significant.
 const RAW_SAMPLE_SUPPORT_QUANTILE = 0.25;
 
 type MetricKey = keyof PerfMetrics;
@@ -547,11 +547,17 @@ function computeSignificance({
 
   const delta = current - baseline;
   const direction = Math.sign(delta);
+  if (direction === 0) {
+    return { agreement: 0, significant: false };
+  }
+
   let agreement = 0;
+  const sampleDeltas: number[] = [];
   for (const comparison of roundComparisons) {
-    if (Math.sign(comparison.delta) !== direction) continue;
-    if (!samplesSupportDirection(comparison.sampleDeltas, direction)) continue;
-    agreement += 1;
+    sampleDeltas.push(...comparison.sampleDeltas);
+    if (Math.sign(comparison.delta) === direction) {
+      agreement += 1;
+    }
   }
 
   const percentOk = Math.abs(percent) > THRESHOLD_PERCENT;
@@ -560,10 +566,11 @@ function computeSignificance({
     baseline === 0 && current !== baseline && Math.abs(current) >= minDelta;
   const magnitudeOk = baseline === 0 ? zeroBaselineOk : percentOk && deltaOk;
   const agreementOk = agreement >= requiredAgreement(roundComparisons.length);
+  const sampleSupportOk = samplesSupportDirection(sampleDeltas, direction);
 
   return {
     agreement,
-    significant: primary && magnitudeOk && agreementOk,
+    significant: primary && magnitudeOk && agreementOk && sampleSupportOk,
   };
 }
 
@@ -1142,6 +1149,9 @@ function formatMarkdown(
     unpairedTests.length;
   const resultsName = getResultsName(options);
   const resultName = getResultName(options);
+  const supportClause = options.node
+    ? "rounds agree on direction"
+    : "rounds agree on direction and raw samples support it";
 
   const lines: string[] = [];
   lines.push("## Performance");
@@ -1279,12 +1289,12 @@ function formatMarkdown(
   if (pairedRoundsCount == null) {
     lines.push("");
     lines.push(
-      `Compared ${resultsName} used mixed interleaved round counts; a change is flagged only when the median exceeds the threshold and raw samples support the same direction.`,
+      `Compared ${resultsName} used mixed interleaved round counts; a change is flagged only when the median exceeds the threshold, ${supportClause}.`,
     );
   } else if (pairedRoundsCount > 1) {
     lines.push("");
     lines.push(
-      `Aggregated across ${pairedRoundsCount} interleaved rounds; a change is flagged only when the median exceeds the threshold and raw samples support the same direction.`,
+      `Aggregated across ${pairedRoundsCount} interleaved rounds; a change is flagged only when the median exceeds the threshold, ${supportClause}.`,
     );
   }
 
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -25,6 +25,9 @@ const RESULTS_DIR = path.join(process.cwd(), ".perf-results");
 const PROFILE_LIMIT = 10;
 const THRESHOLD_PERCENT = 10;
 const MIN_SIGNIFICANT_DELTA_MS = 5;
+// Require at least 75% of pairwise raw-sample deltas to support the median
+// direction before a round contributes to significance.
+const RAW_SAMPLE_SUPPORT_QUANTILE = 0.25;
 
 type MetricKey = keyof PerfMetrics;
 
@@ -75,6 +78,13 @@ interface ProfileModeMismatch {
 interface RoundPerfResult {
   roundIndex: number;
   result: PerfResult;
+}
+
+interface RoundMetricComparison {
+  baseline: number;
+  current: number;
+  delta: number;
+  sampleDeltas: number[];
 }
 
 interface AggregatedPerfResult {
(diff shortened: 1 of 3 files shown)


packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -459,40 +469,97 @@ function requiredAgreement(roundsCount: number) {
...
   const delta = current - baseline;
   const direction = Math.sign(delta);
   let agreement = 0;
-  for (const roundDelta of perRoundDeltas) {
-    if (Math.sign(roundDelta) === direction) {
-      agreement += 1;
-    }
+  for (const comparison of roundComparisons) {
+    if (Math.sign(comparison.delta) !== direction) continue;
+    if (!samplesSupportDirection(comparison.sampleDeltas, direction)) continue;
```
diegohaz: **Confirmation rounds now make a real regression _harder_ to confirm, not more confident.**

Each round now has to clear two AND-ed conditions (median direction match **and** raw-sample support). Combined with `requiredAgreement()`, which demands that **all** rounds agree when `roundsCount <= 4`, every extra confirmation round adds another pass requirement — so adding rounds lowers the odds a genuine regression survives.

I simulated the full flow (Gaussian per-iteration noise, 10 iterations/round → median, then the exact preliminary → confirmation → final gate from `perf.yml`). Published Chrome detection (2 initial + 1 confirmation round):

| scenario | flagged |
|---|---|
| no change, 15% noise | 0% |
| +12% regression, 15% noise | 5% |
| +15% regression, 15% noise | 14% |
| +20% regression, 15% noise | 47% |
| +30% regression, 15% noise | 96% |

False positives are eliminated — exactly the goal — but moderate regressions in noisy tests are rarely surfaced, and the confirmation round actively hurts here: a +15% regression at 15% iteration noise clears the 2-round preliminary ~30% of the time but drops to ~14% after the 3-round final, because `requiredAgreement(3)=3` now needs the raw gate to pass in **all three** rounds.

If that sensitivity trade-off is intended, no change needed. If you'd prefer more rounds to *increase* confidence monotonically, consider running the raw-sample support check on the **pooled** samples across all shared rounds (one dominance test over the combined pairwise deltas) instead of AND-ing a per-round raw check with `requiredAgreement`. That keeps the noise suppression while letting extra rounds add evidence rather than extra hurdles.

(Validated locally: 36/36 `perf-compare.test.ts` pass, `tsc` clean.)

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1209,12 +1279,12 @@ function formatMarkdown(
...
     lines.push("");
     lines.push(
-      `Compared ${resultsName} used mixed interleaved round counts; a change is flagged only when the median exceeds the threshold and rounds agree on direction.`,
+      `Compared ${resultsName} used mixed interleaved round counts; a change is flagged only when the median exceeds the threshold and raw samples support the same direction.`,
     );
   } else if (pairedRoundsCount > 1) {
     lines.push("");
     lines.push(
-      `Aggregated across ${pairedRoundsCount} interleaved rounds; a change is flagged only when the median exceeds the threshold and rounds agree on direction.`,
+      `Aggregated across ${pairedRoundsCount} interleaved rounds; a change is flagged only when the median exceeds the threshold and raw samples support the same direction.`,
```
diegohaz: This wording is now slightly misleading for the **Node** comment. Node benchmarks carry a single raw sample per round (`raw: [metrics]` from `resultsFromBenchmarkReport`), so the new raw-sample check is a no-op there — `samplesSupportDirection` on a one-element `sampleDeltas` just re-checks the round's own delta sign. Yet this line also renders in the Node comment (Node runs 2+ rounds, so it hits this `pairedRoundsCount > 1` branch as well as the `== null` branch above), telling readers a change is "flagged only when … raw samples support the same direction" even though Node has no per-iteration raw samples.

The previous "rounds agree on direction" phrasing was accurate for both modes. Consider branching the clause on `options.node` (the same line appears in the `pairedRoundsCount == null` branch above, so a shared variable avoids duplicating the fix):

```ts
const supportClause = options.node
  ? "rounds agree on direction"
  : "raw samples support the same direction";
```

diegohaz (commented):
## Review summary

**Initial review.** Baseline: merge-base `08a262d1` → head `8a09e59d`. Scope: all 3 changed files (`perf.yml`, `perf-compare.ts`, `perf-compare.test.ts`) plus surrounding context (`perf.ts` data model, the full `compare`/`formatMarkdown` pipeline, and the Chrome/Node CI round flow).

Overall this is a solid, well-tested change and the crossover design is sound: alternating baseline/current execution order turns same-runner drift into *opposite-direction* per-round deltas, which the existing direction-agreement requirement then rejects — so the crossover and the agreement gate reinforce each other. No correctness bugs or regressions found. Two non-blocking findings posted inline.

### Validation performed
- **Tests:** all 36 cases in `perf-compare.test.ts` pass. Note: `pnpm test <file>` (and `pnpm test-react <file>`) both fail to *collect* in this worktree — the no-loader suite's `vitest.setup.ts` imports `@ariakit/test/react` (JSX) without the React vite plugin, and the React loader's include globs exclude `packages/ariakit-scripts`. This is a pre-existing environment/config issue unrelated to the PR (the affected files are unchanged from base). I ran the suite green via a minimal node-environment vitest config.
- **Types:** `pnpm tsc` clean (0 errors).
- **Statistical simulation:** reproduced the significance gate and the full preliminary → confirmation → final flow with Gaussian per-iteration noise (10 iterations/round → median). Confirmed **0% false positives** at 10–20% noise across 2–3 rounds — the PR achieves its stated goal well.

### Findings (both inline, both non-blocking)
1. **Sensitivity / confirmation-round interaction** (`perf-compare.ts`, `computeSignificance`): because `requiredAgreement()` requires *all* rounds to agree when `roundsCount <= 4`, AND-ing the new per-round raw-sample gate means each confirmation round makes a real regression *harder* to confirm. Simulated published Chrome detection: +15% regression at 15% noise ≈ 14%, +20% ≈ 47%, +30% ≈ 96%. Flagged as a trade-off to confirm, with a pooled-sample alternative suggested if monotonic "more rounds = more confidence" is desired.
2. **Node footer wording** (`perf-compare.ts`, `formatMarkdown`): the new "raw samples support the same direction" explanation now also renders in the Node comment, where there is only one raw sample per round (the raw gate is a no-op). Suggested branching the clause on `options.node`.

### Notes / areas not exhaustively validated
- The raw-sample improvement is **Chrome-only**; Node benchmarks expose a single aggregate sample per round, so Node benefits only from the crossover (appropriate given vitest-bench output).
- Crossover balance is slightly uneven with odd round counts (Chrome 2+1 → 2 baseline-first, 1 current-first), but the agreement gate — not aggregate cancellation — does the drift rejection, so this is immaterial.
- No changeset needed: `@ariakit/scripts` is private and in the changeset `ignore` list; the change is CI/tooling only.
- Simulation numbers depend on the noise model (i.i.d. Gaussian, no systematic drift); the qualitative conclusions are robust, the exact percentages are illustrative.
- Did not run the perf workflow end-to-end in CI (not feasible locally); reviewed `perf.yml` by reading.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 4fe87c49b - pooled raw sample deltas across all shared rounds for a single support check while keeping round-median agreement separate, so confirmation rounds add raw evidence instead of adding another per-round raw gate.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 4fe87c49b - split the multi-round footer wording so Node reports round agreement, while browser perf reports round agreement plus raw-sample support. Added assertions for both variants.

diegohaz:
**Fixed**

- 4fe87c49b ([reply](https://github.com/ariakit/ariakit/pull/6566#discussion_r3487237979)): Pooled raw sample support across all shared rounds so confirmation rounds add evidence instead of adding another per-round raw gate.
- 4fe87c49b ([reply](https://github.com/ariakit/ariakit/pull/6566#discussion_r3487238119)): Split multi-round footer wording for Node vs browser perf and added assertions for both variants.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Verified on `4fe87c49b`: pooling the pairwise raw deltas across all shared rounds into one `sampleSupportOk` check, kept separate from the round-median `agreementOk`, is exactly the right fix. I re-ran my full-flow simulation against the new gate:

- False positives stay at **0%** (no change, 15% iteration noise) — specificity preserved.
- Confirmation rounds no longer penalize real regressions. Published Chrome detection of a +15% regression at 15% noise rises ~14% → ~41%, and +20% rises ~47% → ~79%.
- The 2-round → 3-round detection is now flat (~51% → ~51%) instead of dropping (was ~30% → ~16%), so extra rounds add evidence instead of another hurdle.

The new `pools raw sample support across rounds` test captures it well (round 1's overlapping samples alone wouldn't support, but pooled with round 2 they do). Resolved.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Verified on `4fe87c49b`: the `supportClause` branch reads correctly for both modes, and it's accurate — for Node (one raw sample per round) `sampleSupportOk` is mathematically implied by `agreementOk` (the pooled deltas are just the per-round medians, so whenever the required rounds agree on direction the pooled support check also passes), so "rounds agree on direction" is exactly the binding condition. Browser keeps "…and raw samples support it". The `describes Node multi-round comparisons by round agreement` test locks both variants. Resolved.

diegohaz (commented):
## Re-review summary

**Re-review.** Baseline: my previous review at `8a09e59d` → current head `4fe87c49` (commit "Refine perf sample support checks"). Scope of this pass: the incremental diff (`perf-compare.ts` + `perf-compare.test.ts`; `perf.yml` unchanged since last review) plus its effect on the significance pipeline.

**Both prior findings are resolved and independently verified — no new findings.**

1. **Sensitivity / confirmation-round interaction** → fixed. `computeSignificance` now pools the pairwise raw deltas across all shared rounds into a single `sampleSupportOk` check, separate from the round-median `agreementOk`. Re-running my full-flow simulation against the new gate confirms false positives stay at **0%** while moderate-regression detection roughly doubles (+15%/15% noise ≈ 14% → 41%, +20% ≈ 47% → 79%) and confirmation rounds no longer suppress real regressions (2-round → 3-round detection is now flat ~51% instead of dropping ~30% → ~16%). Details on the thread.
2. **Node footer wording** → fixed. `supportClause` branches on `options.node`; Node reads "rounds agree on direction" (accurate — `sampleSupportOk` is implied by `agreementOk` when there's one raw sample per round) and browser reads "…and raw samples support it". Details on the thread.

The added `direction === 0` early return is a clean, correct guard (the zero-baseline path is unaffected since `delta = current - 0 ≠ 0` there).

### Validation performed
- All **38** `perf-compare.test.ts` cases pass (run via a minimal node-environment vitest config; the default-suite collection failure remains the pre-existing unrelated `@ariakit/test/react` JSX-setup transform issue, not PR-caused).
- `pnpm tsc` clean (0 errors).
- Statistical simulation of the new pooled gate through the full preliminary → confirmation → final flow (Gaussian per-iteration noise, 10 iterations/round).

### Not re-validated
- `perf.yml` (unchanged since the previous review).
- End-to-end CI run of the perf workflow (not feasible locally).

No outstanding concerns. LGTM.

PR #6572, opened 2026-06-28
Close nested hovercards on Escape when focus is outside

diegohaz:
## Motivation

Pressing Escape should close the topmost nested Hovercard even when focus has moved to another control on the page. The current Dialog-level Escape handler listens on the document, but nested hovercards can ignore the event when its target is outside the nested card and its disclosure.

There is no public GitHub issue number for this report, so this PR has no auto-closing issue keyword. The repro was added to the existing nested hovercard sandbox because it already models portaled nested hovercards.

## Solution

Add a focused regression case to `app/src/sandbox/hovercard-nested-listener` that opens a nested hovercard, moves focus to an outside input, presses Escape, and expects only the nested hovercard to close while the parent remains open.

Update Dialog's Escape target check so elements marked as outside by the current dialog are still valid Escape targets for that dialog. Parent dialogs remain protected by the existing topmost-dialog guard, so Escape closes the nested card rather than the parent. The existing non-Node target guard remains before all `contains` and marker checks.

A patch changeset covers the behavior for `@ariakit/react-components` and `@ariakit/react`.

## Workaround

A userland workaround is to close the controlled nested hovercard from a document-level Escape listener. This bypasses the current Dialog target check that ignores Escape events when focus is outside the nested hovercard and its anchor.

```tsx
// Userland workaround: Dialog's Escape handler ignores events targeting
// elements outside the nested hovercard and its anchor.
// TODO: Remove once https://github.com/ariakit/ariakit/pull/6572 lands.
useEffect(() => {
  if (!open) return;
  const onKeyDown = (event: KeyboardEvent) => {
    if (event.key !== "Escape") return;
    setOpen(false);
  };
  document.addEventListener("keydown", onKeyDown, true);
  return () => {
    document.removeEventListener("keydown", onKeyDown, true);
  };
}, [open]);
```

With this workaround applied, the repro tests pass:

```sh
pnpm test-react app/src/sandbox/hovercard-nested-listener/test.ts
APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app run test app/src/sandbox/hovercard-nested-listener/
```

### DIFF
--- a/.changeset/300-hovercard-escape-outside-focus.md
+++ b/.changeset/300-hovercard-escape-outside-focus.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed nested [`Hovercard`](https://ariakit.com/reference/hovercard) components so pressing Escape closes the topmost card even when focus is on another element. This also applies to components built on [`Dialog`](https://ariakit.com/reference/dialog).
--- a/app/src/sandbox/hovercard-nested-listener/index.react.tsx
+++ b/app/src/sandbox/hovercard-nested-listener/index.react.tsx
@@ -39,6 +39,21 @@ function useMouseMoveListenerInstalls() {
 
 function NestedHovercard() {
   const [open, setOpen] = useState(false);
+  // Userland workaround: Dialog's Escape handler ignores events targeting
+  // elements outside the nested hovercard and its anchor.
+  // TODO: Remove once https://github.com/ariakit/ariakit/pull/6572 lands.
+  useEffect(() => {
+    if (!open) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Escape") return;
+      if (event.defaultPrevented) return;
+      setOpen(false);
+    };
+    document.addEventListener("keydown", onKeyDown, true);
+    return () => {
+      document.removeEventListener("keydown", onKeyDown, true);
+    };
+  }, [open]);
   return (
     <>
       <button type="button" onClick={() => setOpen((value) => !value)}>
--- a/app/src/sandbox/hovercard-nested-listener/index.react.tsx
+++ b/app/src/sandbox/hovercard-nested-listener/index.react.tsx
@@ -45,7 +45,7 @@ function NestedHovercard() {
         Toggle nested
       </button>
       {open && (
-        <Ariakit.HovercardProvider open>
+        <Ariakit.HovercardProvider open={open} setOpen={setOpen}>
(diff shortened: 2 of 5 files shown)


diegohaz (commented):
**Initial review** — comment-only, no blocking findings.

**Baseline:** merge-base with `origin/main` (`358d1dd2f`) → head `2dbb35b4b`.

**Scope:** the `useDialog` Escape handler change in `packages/ariakit-react-components/src/dialog/dialog.tsx`, the `hovercard-nested-listener` sandbox + happy-dom and Playwright tests, and the changeset. I also read the surrounding marking machinery (`tree-cleanup.ts`, `walk-tree-outside.ts`, `mark-tree-outside.ts`, `use-nested-dialogs.tsx`, `use-hide-on-interact-outside.ts`) to validate the approach against arbitrary nesting depth.

**Assessment:** the change is correct, minimal, and well-targeted.

- The new `isElementMarked(target, dialog.id)` branch only runs after the topmost-dialog guard (`isElementMarked(dialog)`), so parents stay protected and only the topmost open dialog closes — confirmed by analysis and by the broader suite.
- It matches the existing convention already used for outside detection in `use-hide-on-interact-outside.ts` (`isElementMarked(target, contentElement.id)` at lines 95/160), so the approach is consistent with the codebase rather than a new ad-hoc check.
- Multi-level nesting is safe: `useNestedDialogs.add` propagates each descendant up the whole ancestor chain, so an ancestor never marks a descendant as "outside", and the deepest dialog remains the only one past the topmost guard.
- The behavior broadening (Escape closing the topmost layer even when focus is outside, for any `Dialog`-based component, most visibly with `hideOnInteractOutside={false}`) is intentional and documented in the changeset.

**Validation performed locally (happy-dom, React loader):**

- `pnpm test-react run app/src/sandbox/hovercard-nested-listener/test.ts` → 2/2 pass.
- Reverted only `dialog.tsx` to `origin/main` and re-ran the sandbox test → the new "closes nested hovercard on Escape when focus is outside" case fails (nested card stays attached), confirming the test genuinely guards the fix; restored afterward.
- `pnpm test-react run examples/dialog examples/menu examples/hovercard examples/popover` → 284/284 pass.
- `pnpm test-react run examples/combobox examples/select examples/tooltip` → 275/275 pass.
- Confirmed `isElement`/`isNode` imports and `isElementMarked` typing; no type concerns in the changed lines.

**Not run locally:** the Playwright `test-browser.ts` case (would require booting the worktree dev servers). It is covered by CI — Chrome/Firefox/Safari and all other checks are green on this head.

No simplification, correctness, performance, or compatibility issues to flag. Nice, tightly-scoped fix.

Issue #4388, opened 2025-01-17
[Combobox] Can't move between composition candidates using arrow keys


jaknas:
### Current behavior

When typing something within the Combobox input using Japanese - Romaji input source on MacOS, when "Space" key is pressed to open composition menu, I can't freely move between the composition candidates using arrow keys.

Right now, moving up will select the above composition and auto-select it, which will close the composition menu.

_note: I don't know Japanese, but had this reported by one of our Japanese speaking users._

### Steps to reproduce the bug

1. Setup Japanese - Romaji input source  
2. Open https://ariakit.org/components/combobox
3. Using Japanese - Romaji, type "ko" in the combobox input
4. Press "Space" to open the composition menu
5. Press "ArrowUp" to move between the compositions

Notice that composition menu closes instantly, and I'm not allowed to move freely.

### Expected behavior

Work similar to how it's done in RAC Combobox: https://react-spectrum.adobe.com/react-aria/ComboBox.html 

If we follow the same reproduction steps, I'm allowed to move freely up and down between the composition candidates, and select one with Enter.

### Workaround

https://github.com/ariakit/ariakit/issues/4388#issuecomment-2629283924

### Possible solutions

https://github.com/ariakit/ariakit/issues/4388#issuecomment-2629283924

diegohaz:
I can't test this right now (no computer), but I'm pretty sure it was working before. We don't have automated tests for that because Playwright doesn't support composition text, but I'll give it another try later.

diegohaz:
I can now reproduce the bug. The issue stems from different composition text behaviors. Our previous tests used horizontal arrow keys that didn't interact with the vertical combobox, while Japanese composition uses vertical arrow keys that clash with combobox commands.

**Workaround:**

Return `false` from [`moveOnKeyPress`](https://ariakit.org/reference/combobox#moveonkeypress) during composition events:

```jsx
<Combobox
  // TODO: Remove when https://github.com/ariakit/ariakit/issues/4388 is fixed
  moveOnKeyPress={(event) => !event.nativeEvent.isComposing}
/>
```

I believe we should apply the same fix here:  
https://github.com/ariakit/ariakit/blob/82009809c4e038ded6145e1fa31c04abafd82cdd/packages/ariakit-react-core/src/composite/composite.tsx#L369-L373

PR #6580, opened 2026-06-30
Fix toolbar container IME commit handling

### LINKED ISSUE #6579: [CompositeContainer] Enter during IME composition exits the field and blocks the commit
While reading the composite keydown handlers I noticed that `CompositeContainer` does not guard against IME composition, while its sibling `Composite` does.

In `packages/ariakit-react-components/src/composite/composite-container.tsx`, `onKeyDown` (defined around line 154) has an Enter branch around line 215:

```ts
else if (event.key === "Enter") {
  const target = event.target as HTMLElement;
  const isInput =
    (target.tagName === "INPUT" && !isButton(target)) ||
    target.tagName === "TEXTAREA";
  if (isInput || target.isContentEditable) {
    event.preventDefault();
    queueMicrotask(() => container.focus());
  }
}
```

When the focused child is a text field or a contenteditable element and the user presses Enter to commit an IME composition (Japanese, Chinese, Korean), this branch runs, calls `preventDefault()`, and moves focus back to the container. The composition never commits and the user is pushed out of the field in the middle of typing.

This only reproduces in Safari/WebKit. WebKit dispatches the committing Enter as a real `keydown` with `event.key === "Enter"` and `event.nativeEvent.isComposing === true`. Chromium and Firefox report it as keyCode 229 (`key === "Process"`), so `event.key === "Enter"` is false there and the branch is skipped. The Delete/Backspace and Escape branches in the same handler have the same exposure.

The sibling handler already accounts for this. `Composite`'s `onKeyDown` in `composite/composite.tsx` (around line 395) returns early during composition:

```ts
const onKeyDown = useEvent((event) => {
  onKeyDownProp?.(event);
  // https://github.com/ariakit/ariakit/issues/4388
  if (event.nativeEvent.isComposing) return;
  ...
```

`CompositeContainer`'s `onKeyDown` has no equivalent check.

Proposed fix: mirror the same guard at the top of `CompositeContainer`'s `onKeyDown`, right after `onKeyDownProp?.(event)`:

```ts
const onKeyDown = useEvent((event) => {
  onKeyDownProp?.(event);
  if (event.nativeEvent.isComposing) return;
  ...
```

That covers the Enter, Escape, and Delete/Backspace branches with one early return, the same way `Composite` does it.

I'm happy to open a PR. The contributing guide says PRs should come from a branch on the main repo and to reach out first without write access, so let me know if you would like me to send one.

### Workaround

https://github.com/ariakit/ariakit/issues/6579#issuecomment-4840368664

diegohaz:
## Motivation

Fixes #6579.

`ToolbarContainer` is built on `CompositeContainer`. When a focused text field inside the container receives a composing `Enter` keydown, the previous `CompositeContainer` handler treated it like a normal `Enter`, called `preventDefault()`, and moved focus back to the container. In Safari/WebKit, that could block CJK IME confirmation and push the user out of the field while typing.

## Solution

Make `CompositeContainer` skip its internal key handling while `event.nativeEvent.isComposing` is true, and also when Safari reports the confirming IME Enter as `keyCode === 229` after `compositionend`. The user `onKeyDown` still runs first, but the internal Enter, Escape, Delete, and Backspace branches no longer interfere with IME composition.

The repro sandbox under `app/src/sandbox/toolbar-container-ime-6579/` renders a public [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container) scenario with a text input descendant. It includes a happy-dom duplicate that uses `press.down.Enter` with `isComposing`, a happy-dom Safari fallback regression for `keyCode === 229`, plus a Chrome Playwright test that drives real IME composition through CDP (`Input.imeSetComposition`) before pressing Enter. Both tests assert that the text field keeps focus after the composing Enter keydown, and the Chrome test also asserts that the committed character reaches the input.

## Workaround

Until this fix is released, stop the composing keydown on editable descendants before it bubbles to [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container). Do not call `preventDefault()`, so the browser can still commit the IME text.

```tsx
const isComposing = (event: React.KeyboardEvent) => {
  return event.nativeEvent.isComposing || event.keyCode === 229;
};

<ToolbarContainer>
  <input
    // TODO: Remove once https://github.com/ariakit/ariakit/issues/6579 is fixed.
    onKeyDown={(event) => {
      if (isComposing(event)) {
        event.stopPropagation();
      }
    }}
  />
</ToolbarContainer>
```


app/src/sandbox/toolbar-container-ime-6579/test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { expect, test } from "vitest";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6579
+test("keeps focus in the field when Enter commits IME composition", async () => {
+  const input = q.textbox.ensure("Message");
+
+  input.focus();
+  expect(input).toHaveFocus();
+
+  const defaultAllowed = await dispatch.keyDown(input, {
```
diegohaz: Doesn't `press.down` work here?

app/src/sandbox/toolbar-container-ime-6579/test-browser.ts
```diff
@@ -0,0 +1,31 @@
+import type { Locator } from "@playwright/test";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+function pressComposingEnter(input: Locator) {
```
diegohaz: Can't we test ime input directly with `page.context().newCDPSession(page)`?

See tag-input-6294 tests. It's fine if it's just chrome.

app/src/sandbox/toolbar-container-ime-6579/test.ts
diegohaz: Fixed in a212d3ff2 - the happy-dom repro now uses `press.down.Enter(input, { isComposing: true })` and still asserts the composing Enter was not default-prevented. I verified the updated test fails with the library guard removed, passes with only the sandbox workaround, and passes with the final library fix.

app/src/sandbox/toolbar-container-ime-6579/test-browser.ts
diegohaz: Fixed in a212d3ff2 - the browser repro is now Chrome-only and uses `page.context().newCDPSession(page)` with `Input.imeSetComposition`, followed by `page.keyboard.press("Enter")`. I verified the updated test fails with the library guard removed, passes with only the sandbox workaround, and passes with the final library fix.

diegohaz:
**Fixed**

- a212d3ff2 ([thread](https://github.com/ariakit/ariakit/pull/6580#discussion_r3496804946)): Updated the happy-dom repro to use `press.down.Enter(input, { isComposing: true })` while still asserting the composing Enter was not default-prevented.
- a212d3ff2 ([thread](https://github.com/ariakit/ariakit/pull/6580#discussion_r3496804984)): Replaced the synthetic browser repro with a Chrome-only CDP IME composition test using `Input.imeSetComposition`.

Both updated tests were checked against all required states: they fail with no fix, pass with the sandbox workaround, and pass with the library fix.

packages/ariakit-react-components/src/composite/composite-container.tsx
```diff
@@ -153,6 +153,8 @@ export const useCompositeContainer = createHook<
 
   const onKeyDown = useEvent((event: KeyboardEvent<HTMLType>) => {
     onKeyDownProp?.(event);
+    // https://github.com/ariakit/ariakit/issues/6579
+    if (event.nativeEvent.isComposing) return;
```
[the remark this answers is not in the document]
diegohaz: Fixed in 05abe0923 - `CompositeContainer` now also skips internal key handling when Safari reports the confirming IME Enter as `keyCode === 229`, and the happy-dom sandbox test covers that path by asserting focus stays in the input.

diegohaz:
**Fixed**

- 05abe0923 (https://github.com/ariakit/ariakit/pull/6580#discussion_r3497095044): Added the Safari/WebKit `keyCode === 229` IME fallback to `CompositeContainer` and covered it with a happy-dom regression test that keeps focus in the input.

app/src/sandbox/toolbar-container-ime-6579/test.ts
```diff
@@ -0,0 +1,38 @@
...
+async function pressSafariComposingEnter(input: HTMLElement) {
+  const event = new KeyboardEvent("keydown", {
+    bubbles: true,
+    cancelable: true,
+    key: "Enter",
+  });
+  Object.defineProperty(event, "keyCode", { get: () => 229 });
+  input.dispatchEvent(event);
+  await sleep();
+}
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because `test-chrome.ts` is the source-of-truth browser coverage for this regression and now asserts the committed IME value after `commitComposition(page, "に")`, plus focus. The happy-dom duplicate only keeps the reliably reproducible focus behavior; adding a `defaultPrevented` assertion there would be an internal proxy rather than additional user-visible coverage.

diegohaz:
**Invalid**

- Comment (https://github.com/ariakit/ariakit/pull/6580#discussion_r3497164337): Skipped the suggested happy-dom `defaultPrevented` assertion because `test-chrome.ts` is the source-of-truth browser coverage and already asserts the committed IME value plus focus; the happy-dom duplicate stays focused on the reliably reproducible focus behavior.

diegohaz (commented):
**General PR review — no blocking findings.**

Initial review. Baseline: merge-base `b2813af31` → head `05abe092`. Scope: the `useCompositeContainer` `onKeyDown` guard plus the new `toolbar-container-ime-6579` sandbox (`index.react.tsx`, `test.ts`, `test-chrome.ts`) and the changeset.

**Verdict:** the fix is correct, minimal, and consistent with the existing `Composite` handler, which already returns early on `event.nativeEvent.isComposing` (`composite.tsx:398`, issue #4388). The extra `keyCode === 229` guard is justified for `CompositeContainer` specifically: unlike `Composite` (which bails on `!isSelfTarget`), `CompositeContainer` acts on events whose target is a *nested* field, so it's the handler exposed to Safari's post-`compositionend` confirming Enter. Skipping for any `keyCode === 229` (not just Enter) is harmless since 229 is only emitted while the IME is processing.

**What I validated locally (Node 24, happy-dom react loader):**
- Ran `test.ts` on the PR head — both tests pass.
- Reverted only the two new guard lines and re-ran — **both tests fail** (`toHaveFocus` fails as focus is pulled back to the container), confirming the happy-dom duplicate genuinely guards both the `isComposing` path and the Safari `keyCode === 229` path, not just the focus side effect. Restored the file afterward (worktree left clean).
- `pnpm tsc` (exit 0) and `oxlint` on the changed files (clean).
- Re-ran the existing `toolbar-container-*` happy-dom tests — no regression.

**Notes (non-blocking):**
- `CompositeContainer` is internal (not exported from `@ariakit/react` — `composite.ts` barrel omits it), so the changeset correctly attributes the fix to the public `ToolbarContainer` and the package set (`@ariakit/react-components` + `@ariakit/react`) is right. No Solid `composite-container` exists, so React-only scope is correct.
- The prior review threads (CodeRabbit "assert the Enter was not canceled", Codex "include Safari 229 guard", and the two test-shape threads) are all resolved and the resolutions hold against the current head: `test-chrome.ts` asserts the committed value (`toHaveValue(/に/)`) for the IME-commit contract, and the two effects of the fix (no `preventDefault`, no focus move) live in the same early return, so the focus assertion is a complete proxy for the guard being present.

**Not run locally:** the Chrome CDP-IME test (`test-chrome.ts`) — it needs the worktree app server stack. Statically it's a valid guard: Playwright's `keyboard.press("Enter")` dispatches a synthetic `key: "Enter"` while the CDP composition is active, so the keydown carries `isComposing: true` and exercises the same code path the fix guards (matching the author's verified "fails with the guard removed" claim). The `Input.imeSetComposition` + `keyboard.press("Enter")` shape is the correct adaptation of the `tag-input-6294` pattern for the Enter-to-commit scenario.

PR #6586, opened 2026-07-01
Fix Portal StrictMode duplicate containers

### LINKED ISSUE #6585: Portal fullscreen sync can resurrect removed portal nodes in React StrictMode
### Current behavior

After `@ariakit/react@0.4.26`, `Portal` can leave duplicate portal container elements in the DOM in a React development `StrictMode` environment.

This was observed with `Tooltip` in WordPress/Gutenberg after updating Ariakit from `0.4.21`/`0.4.22` to `0.4.29` in https://github.com/WordPress/gutenberg/pull/79055. The public Gutenberg issue is https://github.com/WordPress/gutenberg/issues/79727.

Manual bisect results:

- `@ariakit/react@0.4.25`: no leak.
- `@ariakit/react@0.4.26`: leak starts.
- `@ariakit/react@0.4.31`: still leaks.
- In `0.4.26`, locally reverting the fullscreen reparenting change from https://github.com/ariakit/ariakit/pull/5742 fixes the leak.
- More specifically, removing only the immediate `onFullscreenChange()` call added in the `Sync portal position on mount for existing fullscreen state` commit also appears to fix the leak while keeping the fullscreen root selection and `fullscreenchange` listener in place.

The DOM ends up with duplicate empty portal nodes such as:

```html
<div id="portal/tooltip-123"></div>
<div id="portal/tooltip-123"></div>
```

Only one of those nodes is later cleaned up; the other remains orphaned.

My read of the failure mode is:

1. `Portal` creates a portal element, appends it, and stores it in `portalNode`.
2. The portal ownership cleanup removes that element.
3. The fullscreen reparenting effect still has the removed `portalNode`.
4. The immediate `onFullscreenChange()` call sees `portalNode.parentElement === null`.
5. Since `null !== document.body`, it appends the already-removed node back to `document.body`.
6. A subsequent mount creates a fresh portal element with the same id, leaving the resurrected node orphaned.

#### In the WordPress Gutenberg editor

https://github.com/user-attachments/assets/f4190497-b3cc-43e8-9850-944bf79d3fc0

#### In isolation

I do not have a StackBlitz link yet, but the issue should reproduce in a React development `StrictMode` environment with `@ariakit/react@0.4.26` or later.

Minimal shape:

```tsx
import { StrictMode } from "react";

function App() {
  return (
    <StrictMode>
      <Ariakit.TooltipProvider timeout={0}>
        <Ariakit.TooltipAnchor render={<button type="button" />}>
          Hover me
        </Ariakit.TooltipAnchor>
        <Ariakit.Tooltip id="tooltip-repro" unmountOnHide>
          Tooltip content
        </Ariakit.Tooltip>
      </Ariakit.TooltipProvider>
    </StrictMode>
  );
}

createRoot(document.getElementById("root")!).render(<App />);
```

Then:

1. Install `@ariakit/react@0.4.26` or later.
2. Run the app in React development mode with `StrictMode`.
3. Hover the button to show the tooltip.
4. Move away or switch/remount tooltip anchors a few times.
5. Inspect:

```js
document.querySelectorAll('div[id^="portal/tooltip-"]')
```

In Gutenberg, this reproduces by opening the editor, selecting blocks, hovering block toolbar buttons to show tooltips, switching selected blocks, and checking the same selector.

### Expected behavior

`Portal` should not re-append a portal node that has already been removed by its own cleanup. After a tooltip is hidden or unmounted, there should be no orphaned empty portal containers left in the DOM, and duplicate portal IDs should not be created.

### Workaround

For Gutenberg testing, locally patching Ariakit `0.4.26` to remove the immediate `onFullscreenChange()` call from the fullscreen reparenting effect stopped the leak.

Reverting the whole fullscreen reparenting change also stopped the leak, but that is not a good userland workaround because it loses the fullscreen fix.

Confirmed workaround: https://github.com/ariakit/ariakit/issues/6585#issuecomment-4854997887

### Possible solutions

The smallest fix may be to guard the fullscreen reparenting handler against disconnected/stale portal nodes:

```ts
const onFullscreenChange = () => {
  if (!portalNode.isConnected) return;

  const rootElement = getRootElement(portalNode);
  if (portalNode.parentElement !== rootElement) {
    rootElement.appendChild(portalNode);
  }
};
```

This should still preserve the behavior from https://github.com/ariakit/ariakit/pull/5742:

- Existing connected portals can move into `document.fullscreenElement` on `fullscreenchange`.
- Existing connected portals can move back to `document.body` after exiting fullscreen.
- A portal mounted while already fullscreen is initially appended to `document.fullscreenElement` by `getRootElement`.

But it avoids resurrecting a portal node that has already been removed by the owning effect cleanup.

diegohaz:
## Motivation

Fixes #6585.

`Portal` can leave duplicate portal containers in React 18 development `StrictMode`. The fullscreen sync effect runs immediately after a portal node is stored, and in StrictMode that can happen after the layout cleanup has already removed a stale portal node. Since the stale node has no parent, the immediate sync can append it back to the document and leave two connected nodes with the same `portal/...` id.

This affected components built on [`Portal`](https://ariakit.com/reference/portal), including tooltips.

## Solution

Added a focused sandbox regression at `app/src/sandbox/tooltip-6585` that counts connected portal containers while showing and unmounting a tooltip in `StrictMode`. The browser coverage also exercises fullscreen movement, including mounting the tooltip while already fullscreen and removing the fullscreen host while the tooltip stays mounted.

The library fix keeps the fullscreen behavior from #5742, including the immediate sync for portals mounted while already fullscreen, but skips that immediate sync when the captured portal node is no longer connected:

```ts
if (portalNode.isConnected) {
  onFullscreenChange();
}
```

That prevents a StrictMode cleanup node from being reattached. Real `fullscreenchange` events still run the reparenting logic, so an active portal can move back to `document.body` if the fullscreen element containing it is removed.

The fullscreen root check now also uses the portal node document's `defaultView.HTMLElement`, which avoids relying on the parent window's `HTMLElement` constructor for iframe/cross-realm documents.

## Workaround

Before this fix, a userland workaround was to pass a stable custom `portalElement` to the tooltip and move that element on `fullscreenchange`. Supplying `portalElement` avoided the internal auto-created portal node path that could be duplicated by the fullscreen sync effect.

### DIFF
--- a/.changeset/300-portal-fullscreen-strictmode.md
+++ b/.changeset/300-portal-fullscreen-strictmode.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Portal`](https://ariakit.com/reference/portal), including components built on it such as [`Tooltip`](https://ariakit.com/reference/tooltip) and [`Popover`](https://ariakit.com/reference/popover), to avoid leaking duplicate portal containers in React development StrictMode when syncing with fullscreen elements.
--- a/.changeset/300-portal-fullscreen-strictmode.md
+++ b/.changeset/300-portal-fullscreen-strictmode.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Portal`](https://ariakit.com/reference/portal), including components built on it such as [`Tooltip`](https://ariakit.com/reference/tooltip) and [`Popover`](https://ariakit.com/reference/popover), to avoid leaking duplicate portal containers in React development StrictMode when syncing with fullscreen elements.
+Fixed [`Portal`](https://ariakit.com/reference/portal), including components built on it such as [`Tooltip`](https://ariakit.com/reference/tooltip) and [`Popover`](https://ariakit.com/reference/popover), to avoid leaking duplicate portal containers in React development StrictMode.
--- a/app/src/sandbox/tooltip-6585/test-browser.ts
+++ b/app/src/sandbox/tooltip-6585/test-browser.ts
@@ -1,3 +1,4 @@
+import type { Page } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
 
 // See https://github.com/ariakit/ariakit/issues/6585
@@ -5,9 +6,7 @@ import { withFramework } from "#app/test-utils/preview.ts";
 // `pnpm react18 dev-app`. CI's React 18 signal comes from test.ts.
 const portalId = "portal/tooltip-repro";
 
-async function getPortalState(page: {
-  evaluate: <T>(fn: (id: string) => T, arg: string) => Promise<T>;
-}) {
+async function getPortalState(page: Page) {
   return page.evaluate((id) => {
     const nodes = Array.from(document.querySelectorAll(`#${CSS.escape(id)}`));
     const node = nodes[0];
--- a/app/src/sandbox/tooltip-6585/test-browser.ts
+++ b/app/src/sandbox/tooltip-6585/test-browser.ts
@@ -99,4 +99,36 @@ withFramework(import.meta.dirname, async ({ test }) => {
     const state = await getPortalState(page);
     test.expect(state.count).toBe(0);
   });
+
+  test("moves the portal to body when the fullscreen host is removed", async ({
+    page,
+    q,
+  }) => {
+    await test
+      .expect(q.status("Portal containers"))
+      .toHaveText("Portal containers: 0");
+
+    await q.button("Pin tooltip").click();
+    await q.button("Enter fullscreen").click();
+    await page.waitForFunction(() => document.fullscreenElement != null);
+
+    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
+    await test.expect
+      .poll(() => getPortalState(page))
+      .toMatchObject({
+        count: 1,
+        parentIsFullscreen: true,
+      });
+
+    await q.button("Unmount fullscreen host").click();
+    await page.waitForFunction(() => document.fullscreenElement == null);
+
+    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
+    await test.expect
+      .poll(() => getPortalState(page))
+      .toMatchObject({
+        count: 1,
+        parentIsBody: true,
+      });
+  });
 });
--- a/app/src/sandbox/tooltip-6585/test-browser.ts
+++ b/app/src/sandbox/tooltip-6585/test-browser.ts
@@ -51,47 +51,50 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    await q.button("Enter fullscreen").click();
-    await page.waitForFunction(() => document.fullscreenElement != null);
-
-    await q.button("Hover target").hover();
-    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
+    await q.button("Unmount tooltip").click();
     await test
       .expect(q.status("Portal containers"))
-      .toHaveText("Portal containers: 1");
+      .toHaveText("Portal containers: 0");
 
-    let state = await getPortalState(page);
-    test.expect(state).toMatchObject({
-      count: 1,
-      parentIsFullscreen: true,
-    });
+    await q.button("Pin tooltip").click();
+    await q.button("Enter fullscreen").click();
+    await page.waitForFunction(() => document.fullscreenElement != null);
+    await q.button("Mount tooltip").click();
+    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
+
+    await test.expect
+      .poll(() => getPortalState(page))
+      .toMatchObject({
+        count: 1,
+        parentIsFullscreen: true,
+      });
 
     await q.button("Exit fullscreen").click();
     await page.waitForFunction(() => document.fullscreenElement == null);
-    await q.button("Hover target").hover();
-    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
 
-    state = await getPortalState(page);
-    test.expect(state).toMatchObject({
-      count: 1,
-      parentIsBody: true,
-    });
+    await test.expect
+      .poll(() => getPortalState(page))
+      .toMatchObject({
+        count: 1,
+        parentIsBody: true,
+      });
 
     await q.button("Enter fullscreen").click();
     await page.waitForFunction(() => document.fullscreenElement != null);
 
-    state = await getPortalState(page);
-    test.expect(state).toMatchObject({
-      count: 1,
-      parentIsFullscreen: true,
-    });
+    await test.expect
+      .poll(() => getPortalState(page))
+      .toMatchObject({
+        count: 1,
+        parentIsFullscreen: true,
+      });
 
     await q.button("Unmount tooltip").click();
     await test
       .expect(q.status("Portal containers"))
       .toHaveText("Portal containers: 0");
 
-    state = await getPortalState(page);
+    const state = await getPortalState(page);
     test.expect(state.count).toBe(0);
   });
 });
--- a/app/src/sandbox/tooltip-6585/test-browser.ts
+++ b/app/src/sandbox/tooltip-6585/test-browser.ts
@@ -51,6 +51,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
+    await test
+      .expect(q.status("Portal containers"))
+      .toHaveText("Portal containers: 0");
     await q.button("Unmount tooltip").click();
     await test
       .expect(q.status("Portal containers"))
--- a/app/src/sandbox/tooltip-6585/test-browser.ts
+++ b/app/src/sandbox/tooltip-6585/test-browser.ts
@@ -1,0 +1,97 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// See https://github.com/ariakit/ariakit/issues/6585
+// To reproduce the React 18 StrictMode failure in a browser, serve the app with
+// `pnpm react18 dev-app`. CI's React 18 signal comes from test.ts.
+const portalId = "portal/tooltip-repro";
+
+async function getPortalState(page: {
+  evaluate: <T>(fn: (id: string) => T, arg: string) => Promise<T>;
+}) {
+  return page.evaluate((id) => {
+    const nodes = Array.from(document.querySelectorAll(`#${CSS.escape(id)}`));
+    const node = nodes[0];
+    const fullscreenElement = document.fullscreenElement;
+
+    return {
+      count: nodes.length,
+      parentIsBody: node?.parentElement === document.body,
+      parentIsFullscreen:
+        !!fullscreenElement && node?.parentElement === fullscreenElement,
+    };
+  }, portalId);
+}
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("does not leak duplicate tooltip portal containers", async ({
+    page,
+    q,
+  }) => {
+    await test
+      .expect(q.status("Portal containers"))
+      .toHaveText("Portal containers: 0");
+
+    await q.button("Hover target").hover();
+    await test.expect(q.tooltip("Tooltip content")).toBeVisible();
+    await test
+      .expect(q.status("Portal containers"))
+      .toHaveText("Portal containers: 1");
+
+    await q.button("Unmount tooltip").click();
+    await test.expect(q.tooltip("Tooltip content")).not.toBeVisible();
+    await test
+      .expect(q.status("Portal containers"))
+      .toHaveText("Portal containers: 0");
+
+    const state = await getPortalState(page);
+    test.expect(state.count).toBe(0);
+  });
+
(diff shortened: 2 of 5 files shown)


.changeset/300-portal-fullscreen-strictmode.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Portal`](https://ariakit.com/reference/portal), including components built on it such as [`Tooltip`](https://ariakit.com/reference/tooltip) and [`Popover`](https://ariakit.com/reference/popover), to avoid leaking duplicate portal containers in React development StrictMode when syncing with fullscreen elements.
```
diegohaz: The "when syncing with fullscreen elements" qualifier reads as a scope condition, but the leak reproduces for any `Portal`-based component in StrictMode even when fullscreen is never used. The buggy path is the fullscreen-sync effect's **unconditional** immediate `onFullscreenChange()` call: after a StrictMode remount it resurrects the orphaned portal node by appending it to `document.body` (`getRootElement` returns `body` when nothing is fullscreen), producing a second connected `portal/…` node.

I verified this locally — the happy-dom `test.ts`, which never enters fullscreen, fails on the base with `Portal containers: 2` and passes on this branch. So a user who never uses the Fullscreen API is still affected.

Consider dropping/rewording the fullscreen qualifier so readers don't dismiss the fix as fullscreen-only, e.g.:

> Fixed [`Portal`](https://ariakit.com/reference/portal), including components built on it such as [`Tooltip`](https://ariakit.com/reference/tooltip) and [`Popover`](https://ariakit.com/reference/popover), to avoid leaking duplicate portal containers in React development StrictMode.

app/src/sandbox/tooltip-6585/test-browser.ts
```diff
@@ -0,0 +1,103 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// See https://github.com/ariakit/ariakit/issues/6585
+// To reproduce the React 18 StrictMode failure in a browser, serve the app with
+// `pnpm react18 dev-app`. CI's React 18 signal comes from test.ts.
+const portalId = "portal/tooltip-repro";
+
+async function getPortalState(page: {
```
diegohaz: Minor: `getPortalState` hand-rolls a structural `page` type that is a narrower (and slightly inaccurate) approximation of Playwright's `Page.evaluate` overloads. The sibling `app/src/test-utils/preview.ts` already imports the real type, so it's simpler and more accurate to reuse it here:

```ts
import type { Page } from "@playwright/test";

async function getPortalState(page: Page) {
  // ...
}
```

That drops the bespoke type and keeps `page` in sync with the fixture's actual type if Playwright's signature changes.

diegohaz (commented):
### Review summary

Initial review, diffed from merge-base `7cabc590` to head `12e5415b` (whole-repo context, not just changed lines). Scope: the `Portal` fullscreen-sync fix plus the new `tooltip-6585` sandbox and its tests.

**Verdict:** The fix is correct, minimal, and well-targeted. Only two minor, non-blocking nits (posted inline).

**What the change does & why it's right**
- The leak is the fullscreen-sync effect's unconditional immediate `onFullscreenChange()`. After a StrictMode remount, that effect closure still references the orphaned portal node from the prior mount, and `rootElement.appendChild(portalNode)` re-attaches it (its `parentElement` is `null` after the layout cleanup removed it), leaving two connected `portal/…` nodes with the same id. Gating on `if (!portalNode.isConnected) return;` prevents resurrecting the orphan while preserving the #5742 "mount while already fullscreen" behavior (fresh node is connected, so the immediate sync still runs).
- The `getRootElement` change to `doc.defaultView?.HTMLElement` is a reasonable cross-realm robustness improvement; same behavior in the common same-realm case, more correct for iframe/adopted documents. It's tangential to #6585 and not covered by a dedicated test, but it's low-risk and doesn't regress the fallback (`defaultView == null` → `doc.body`, matching the prior behavior for detached docs).

**Validation performed locally (worktree, Node 24)**
- `pnpm test-react18 run tooltip-6585` → passes with the fix; reverting the `isConnected` guard makes it fail with `Portal containers: 2` (confirms it's a genuine regression test and that the leak reproduces without ever entering fullscreen).
- `pnpm test-react run tooltip-6585` → passes (React 19 suite).
- `pnpm tsc` → clean. `oxlint` + `oxfmt --check` on the changed files → clean.
- Confirmed changeset frontmatter (`@ariakit/react-components` + `@ariakit/react`, both patch) matches the established convention (`300-toolbar-container-ime.md`); `portal`/`tooltip`/`popover` reference URLs all 200; sandbox naming (`tooltip-6585`) and no-`preview.json` follow the sandbox conventions; the `pnpm react18 dev-app` command in the test comment is valid (`dev-app` is a root script dispatched via `pnpm run`).

**Not validated locally**
- The Playwright `test-browser.ts` (chrome/firefox/safari), including the fullscreen enter→exit→enter cycle, was not run here — it needs the app preview server. The pattern matches the existing `tooltip-865` fullscreen browser test, and per-browser retries are configured, so this is expected to be fine, but I didn't execute it.

**Inline findings (both minor)**
1. Changeset wording — "when syncing with fullscreen elements" can read as a fullscreen-only scope; the leak affects any Portal-based component in StrictMode.
2. `test-browser.ts` `getPortalState` hand-rolls a `page` type instead of importing `Page` from `@playwright/test` (as the sibling `preview.ts` does).

.changeset/300-portal-fullscreen-strictmode.md
diegohaz: Fixed in 4b74baee8 - removed the fullscreen qualifier from the changeset so the StrictMode portal container leak is described as affecting Portal generally, not only fullscreen sync.

app/src/sandbox/tooltip-6585/test-browser.ts
diegohaz: Fixed in 4b74baee8 - getPortalState now takes Playwright’s Page type from @playwright/test instead of a bespoke structural evaluate type.

diegohaz:
**Fixed**

- 4b74baee8 ([reply](https://github.com/ariakit/ariakit/pull/6586#discussion_r3506971874)): Reworded the changeset so the StrictMode duplicate portal container fix is not described as fullscreen-only.
- 4b74baee8 ([reply](https://github.com/ariakit/ariakit/pull/6586#discussion_r3506971880)): Replaced the custom structural Playwright page type with the real `Page` type.

packages/ariakit-react-components/src/portal/portal.tsx
```diff
@@ -141,6 +142,7 @@ export const usePortal = createHook<TagName, PortalOptions>(function usePortal({
     if (portalElement) return;
     const doc = getDocument(portalNode);
     const onFullscreenChange = () => {
+      if (!portalNode.isConnected) return;
```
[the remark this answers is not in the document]
diegohaz: This finding is valid — I reproduced the mechanism locally. The problem is that `onFullscreenChange` is used for two different purposes, but only one of them needs the disconnected-node skip:

- **Immediate sync (line 154)** — runs synchronously when the effect (re)mounts. In StrictMode a remounted effect can capture an already-removed cleanup node, so skipping when it's disconnected is exactly right.
- **Real `fullscreenchange` events (line 155 listener)** — after StrictMode settles, only the *live* effect's listener stays attached (the stale one is removed in cleanup). So a disconnected `portalNode` here means the live portal's host was torn down (e.g. the fullscreen element that held it was unmounted while a global tooltip/dialog stayed open). Pre-PR, the handler re-appended it to `document.body`; the broad guard now skips that, so React keeps rendering into the detached node.

DOM mechanics confirm it: after removing the host, `portalNode.isConnected === false` while `portalNode.parentElement === <detached host>`, so `body.appendChild(portalNode)` would have rescued it.

Moving the check to guard only the immediate sync fixes both — I re-ran `pnpm test-react18 run tooltip-6585` with this and the StrictMode leak stays fixed:

```diff
     const onFullscreenChange = () => {
-      if (!portalNode.isConnected) return;
       const rootElement = getRootElement(portalNode);
       if (portalNode.parentElement !== rootElement) {
         rootElement.appendChild(portalNode);
       }
     };
-    // Sync immediately in case fullscreen was entered before this effect
-    // ran, which can happen if the portal mounts while already in
-    // fullscreen mode.
-    onFullscreenChange();
+    // Sync immediately in case fullscreen was entered before this effect
+    // ran, which can happen if the portal mounts while already in
+    // fullscreen mode. Skip when the captured node is already disconnected,
+    // which happens for a StrictMode cleanup node whose layout cleanup
+    // already removed it.
+    if (portalNode.isConnected) {
+      onFullscreenChange();
+    }
     doc.addEventListener("fullscreenchange", onFullscreenChange);
```

The scenario is niche, so a regression test for it would need a real-browser fullscreen host that unmounts while a Portal-based overlay stays open (the happy-dom suite can't drive fullscreen). The refinement itself is free — same complexity, and it restores the pre-PR rescue path.

diegohaz (commented):
### Re-review summary

Re-review of the new commit `4b74baee` ("Address Portal review comments") since my last review at `12e5415b`. Incremental baseline: `12e5415b` → `4b74baee`. This run covered the incremental diff plus the still-unresolved bot thread on `portal.tsx`.

**Prior findings from my last review — both fixed and verified in the current code:**
1. Changeset — the "when syncing with fullscreen elements" qualifier was removed, so the fix now reads as affecting `Portal` generally. ✔
2. `test-browser.ts` — `getPortalState` now takes `Page` from `@playwright/test` instead of the bespoke structural type. ✔

The incremental commit only touched those two files; `portal.tsx` is unchanged from `12e5415b`.

**Unresolved bot thread — confirmed valid (replied on the thread):**
The Codex thread on `portal.tsx:145` is a real, if niche, regression. The `if (!portalNode.isConnected) return;` guard sits on both the immediate mount-time sync *and* the real `fullscreenchange` listener. Only the immediate sync needs to skip StrictMode cleanup nodes; on a real event a disconnected `portalNode` means the live portal's fullscreen host was torn down while the overlay stayed mounted, and pre-PR the handler rescued it back to `document.body`. The broad guard drops that rescue.

Validation I ran (worktree, Node 24):
- Narrowing the guard to the immediate sync only (`if (portalNode.isConnected) onFullscreenChange();`) still fixes the leak — `pnpm test-react18 run tooltip-6585` passes.
- happy-dom DOM-mechanics check confirms the rescue scenario: after removing the host, `portalNode.isConnected === false` with `parentElement === <detached host>`, and re-appending to `body` reconnects it.

Recommendation lives on the thread with a concrete diff. The refinement is same-complexity and restores the pre-PR behavior while keeping the StrictMode leak fixed.

**Not validated:** the Playwright `test-browser.ts` suite (needs the app preview server) and a real-browser regression test for the host-removal rescue scenario — happy-dom can't drive fullscreen, so that path stays browser-only.

packages/ariakit-react-components/src/portal/portal.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in 59d032870 - the disconnected-node guard now applies only to the immediate fullscreen sync, so real fullscreenchange events can still reparent an active portal back to body when its fullscreen host is removed. I also added a browser regression that pins the tooltip, removes the fullscreen host, and verifies the portal returns to document.body.

diegohaz:
**Fixed**

- 59d032870 ([reply](https://github.com/ariakit/ariakit/pull/6586#discussion_r3507271586)): Scoped the disconnected portal guard to immediate fullscreen sync, restored `fullscreenchange` reparenting for active portals whose fullscreen host is removed, and added browser coverage for that path.

diegohaz (commented):
### Re-review summary

Re-review of the new commit `59d03287` ("Preserve fullscreen portal reparenting") since my last review at `4b74baee`. Incremental baseline: `4b74baee` → `59d03287`.

**All open findings are resolved and verified against the current code (not just the thread state):**

- **Reparenting regression (Codex `portal.tsx:145` thread):** Fixed exactly as recommended — the disconnected-node skip now guards only the immediate mount-time sync (`if (portalNode.isConnected) onFullscreenChange();`), while the real `fullscreenchange` listener runs unconditionally so an active portal is reparented back to `document.body` when its fullscreen host is removed. Verified in the committed code.
- The two findings from my first review (changeset qualifier, `Page` type) remain correctly fixed.

**New test coverage added in this commit is a genuine regression test.** The sandbox now mounts the tooltip anchor outside a removable fullscreen host and adds an "Unmount fullscreen host" button; the new `test-browser.ts` case "moves the portal to body when the fullscreen host is removed" pins the tooltip, enters fullscreen, removes the host, and asserts the portal survives at `count: 1, parentIsBody: true`.

**Validation performed locally (worktree, Node 24, chromium against this worktree's astro dev server):**
- `pnpm -F app run test-chrome tooltip-6585` → all 3 browser tests pass on the committed code.
- Reverting to the previous broad guard makes the new host-removal test fail (`getByRole('tooltip')` not found after host removal — the portal stays detached), confirming the new test catches exactly this regression.
- `pnpm test-react18 run tooltip-6585` → still passes, so the StrictMode leak fix is intact.
- `pnpm tsc`, `oxlint`, `oxfmt --check` → all clean.

No remaining findings. From my perspective this is ready to merge.

PR #6595, opened 2026-07-02
Fix Dialog with getPersistentElements closing before the dialog is focused

### LINKED ISSUE #6344: Dialog with getPersistentElements closes on focus into a persistent element if the dialog hasn't been focused yet (autoFocusOnShow={false})
### Current behavior

The [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements) prop documents that the returned elements "will be considered as part of the dialog", excluding them from the close-on-outside-interaction behavior (this is the documented [Dialog with React-Toastify](https://ariakit.com/examples/dialog-react-toastify) pattern).

However, when the dialog hasn't been focused yet — which is the entire lifetime of a dialog rendered with [`autoFocusOnShow`](https://ariakit.com/reference/dialog#autofocusonshow)`={false}`, a common choice for non-modal panels that shouldn't steal focus — focusing or clicking an input inside a persistent element closes the dialog. Right-click (`contextmenu`) on a persistent element follows the same code path. After the user focuses something inside the dialog once, the identical interaction correctly keeps it open, so the failure is focus-history dependent and easy to miss in testing.

The root cause is that the persistent-element protection is purely negative: persistent elements are merely excluded from the "outside" marking applied by [`markTreeOutside`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/dialog.tsx#L300-L318), so the only thing that saves them in the outside listeners is the `isElementMarked` check. But in `useEventOutside`, that check is gated on `focusedRef.current`, which only becomes true after something inside the dialog has been focused ([use-hide-on-interact-outside.ts#L98-L102](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts#L98-L102)):

```ts
// We need to check if the content element has been focused at least once
// before checking if it's marked. This is so hovercards and tooltips
// don't stay open when new nodes are added to the DOM and focused.
const focused = focusedRef.current;
if (focused && !isElementMarked(target, contentElement.id)) return;
// Finally, if the target has been marked as "outside" or is an ancestor
// of the content element, we call the listener.
callListener(event);
```

While `focused` is `false`, every other guard passes for a persistent element (it's not contained in `contentElement`, not the disclosure, etc.), so the [`focusin` listener](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts#L159-L170) calls `store.hide()`. The [`contextmenu` listener](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts#L172-L182) has the same shape. Only `click` is safe, because it [independently re-checks the mark](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts#L150) on the previously mousedown'd element.

The `focused` gate was added for hovercards/tooltips (so they close when brand-new DOM nodes appear and get focused — see #6198), but persistent elements are the opposite case: the dialog explicitly knows about them at open time, yet the gate treats them like unknown outsiders.

3. Click **Open dialog**. The non-modal dialog appears without taking focus (`autoFocusOnShow={false}`).
4. Without clicking inside the dialog, click the **Notification field** input inside the persistent "Notifications" region (or focus it with <kbd>Tab</kbd>, or right-click it).
5. The dialog closes.

For contrast: reopen the dialog, click **Inside field** first, then click **Notification field** — the dialog stays open, as documented.

```tsx

  const [open, setOpen] = useState(false);
  const notificationsRef = useRef<HTMLDivElement>(null);

  return (
    <div style={{ display: "flex", flexDirection: "column", gap: 16, alignItems: "start" }}>
      <Ariakit.Button onClick={() => setOpen(true)}>Open dialog</Ariakit.Button>

      <Ariakit.Dialog
        open={open}
        onClose={() => setOpen(false)}
        modal={false}
        autoFocusOnShow={false}
        getPersistentElements={() => {
          const notifications = notificationsRef.current;
          return notifications ? [notifications] : [];
        }}
        style={{ display: "flex", flexDirection: "column", gap: 8, alignItems: "start", border: "1px solid gray", padding: 16 }}
      >
        <Ariakit.DialogHeading>Dialog</Ariakit.DialogHeading>
        <input aria-label="Inside field" placeholder="Inside field" />
        <Ariakit.DialogDismiss>Close dialog</Ariakit.DialogDismiss>
      </Ariakit.Dialog>

      <div
        ref={notificationsRef}
        role="region"
        aria-label="Notifications"
        style={{ display: "flex", flexDirection: "column", gap: 8, alignItems: "start", border: "1px solid gray", padding: 16 }}
      >
        <input aria-label="Notification field" placeholder="Notification field" />
        <button type="button">Dismiss notification</button>
      </div>
    </div>
  );
}
```

Verified with a Playwright test driving real user interactions in Chrome: clicking the persistent field before focusing the dialog hides it, while the same click after focusing inside the dialog keeps it open.

### Expected behavior

Elements returned by `getPersistentElements` should be treated as part of the dialog regardless of whether the dialog has been focused yet. Focusing, clicking, or right-clicking inside a persistent element should never close the dialog — exactly as already happens once the dialog has been focused once.

### Workaround

Re-assert the persistent elements through [`hideOnInteractOutside`](https://ariakit.com/reference/dialog#hideoninteractoutside), which is consulted before hiding on every code path:

```tsx
<Ariakit.Dialog
  // ...
  getPersistentElements={() => {
    const notifications = notificationsRef.current;
    return notifications ? [notifications] : [];
  }}
  hideOnInteractOutside={(event) => {
    const notifications = notificationsRef.current;
    if (!notifications) return true;
    if (!(event.target instanceof Node)) return true;
    return !notifications.contains(event.target);
  }}
>
```

Keeping `getPersistentElements` is still useful (it also controls the accessibility tree/inert handling for modal dialogs). Verified: with this callback the dialog stays open in the failing scenario.

### Possible solutions

Make the protection for the dialog's own elements positive instead of relying on the absence of an "outside" mark filtered through `focusedRef`. [`markTreeOutside`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/mark-tree-outside.ts#L40-L67) already receives the exact list of "inside" elements (dialog, persistent elements, nested dialog contents, and the disclosure for non-modal dialogs), and `setProperty` from `orchestrate.ts` already provides stacked, restorable expando marking. Concretely:

- In `mark-tree-outside.ts`, add an id-scoped `__ariakit-dialog-inside` expando: a `markElementInside(element, id)` helper (built on `setProperty`/`chain` like `markElement`), apply it to every non-null element passed to `markTreeOutside` (cleanups unshifted into the existing `cleanups` array so `restoreAccessibilityTree` unwinds them), and add an `isElementInside(element, id?)` helper that walks `parentElement` like `isElementMarked`.
- In `useEventOutside.onEvent` (`use-hide-on-interact-outside.ts`), check it *before* the `focused` gate:

```diff
       // Clicked on dialog's bounding box
       if (isMouseEventOnDialog(event, contentElement)) return;
+      // The dialog itself, persistent elements, and nested dialogs are marked
+      // as "inside" when the dialog opens. Events on them must never reach the
+      // outside listeners, regardless of whether the dialog has been focused
+      // yet (for example, with autoFocusOnShow={false}).
+      if (isElementInside(target, contentElement.id)) return;
       // We need to check if the content element has been focused at least once
       // before checking if it's marked. This is so hovercards and tooltips
       // don't stay open when new nodes are added to the DOM and focused.
       const focused = focusedRef.current;
       if (focused && !isElementMarked(target, contentElement.id)) return;
```

This preserves the hovercard/tooltip behavior from #6198 exactly: genuinely new DOM nodes carry neither mark, so they still trigger the close while the dialog is unfocused. When `focused` is true the new check is behaviorally redundant (inside elements were never marked outside, so the existing check already returned early). Both `markTreeOutside` call sites in `dialog.tsx` (modal and non-modal) are covered without changing their signatures.

diegohaz:
## Motivation

Elements returned by [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements) are documented to be "considered as part of the dialog", so interacting with them should never close the dialog (this is the [Dialog with React-Toastify](https://ariakit.com/examples/dialog-react-toastify) pattern).

However, while the dialog hasn't been focused yet — which is the entire lifetime of a dialog rendered with [`autoFocusOnShow`](https://ariakit.com/reference/dialog#autofocusonshow)`={false}`, a common choice for non-modal panels that shouldn't steal focus — focusing, clicking, or right-clicking an element inside a persistent element closes the dialog:

```tsx
<Ariakit.Dialog
  modal={false}
  autoFocusOnShow={false}
  getPersistentElements={() => [notificationsRef.current]}
>
```

The persistent-element protection is purely negative: persistent elements are only *excluded* from the "outside" marking applied when the dialog opens, and the outside listeners in `useEventOutside` only consult that mark after something inside the dialog has been focused (`focusedRef.current`). While the dialog is unfocused, every other guard passes for a persistent element, so the `focusin` and `contextmenu` listeners call `store.hide()`. Once the dialog has been focused once, the identical interaction correctly keeps it open, so the failure is focus-history dependent.

The `focused` gate exists for hovercards and tooltips (#6198), so they close when brand-new DOM nodes appear and get focused — but persistent elements are the opposite case: the dialog explicitly knows about them at open time.

Fixes #6344

## Solution

Positively mark the elements the dialog knows about at open time — the dialog itself, persistent elements, and nested dialogs — with an id-scoped "inside" expando (`markTreeInside`, built on the same stacked, restorable `setProperty` mechanism as the existing outside marks). The outside event listeners now check this mark *before* the `focusedRef` gate, so events on those elements never reach the hide listeners, regardless of focus history.

This preserves the hovercard/tooltip behavior from #6198 exactly: genuinely new DOM nodes carry no inside mark, so they still close an unfocused dialog. Once the dialog has been focused, the new check is behaviorally redundant (inside elements were never marked outside, so the existing check already returned early).

The disclosure element is intentionally *not* marked inside: hovercards and tooltips update `disclosureElement` to the focus source while open (for example, a tooltip re-opening while its menu is open records the menu popup as its disclosure), so a persistent mark taken at open time would go stale and keep the tooltip open when the menu re-opens (caught by the `menu-tooltip-various` suite). The listeners keep re-checking the disclosure against the current store state on every event instead.

The first commit adds a repro sandbox (`dialog-6344`) with a cross-browser Playwright test and a happy-dom duplicate. The two "stays open … before the dialog is focused" tests fail on every browser without the fix, while the guard tests (documented behavior after focusing inside, and closing on genuinely-outside interactions) pass. The second commit demonstrated the userland workaround below in the sandbox (since reverted).

## Workaround

Until the fix is released, users can re-assert the persistent elements through [`hideOnInteractOutside`](https://ariakit.com/reference/dialog#hideoninteractoutside), which is consulted on every hide-on-interact-outside code path regardless of focus history:

```tsx
<Ariakit.Dialog
  modal={false}
  autoFocusOnShow={false}
  getPersistentElements={() => {
    const notifications = notificationsRef.current;
    return notifications ? [notifications] : [];
  }}
  // TODO: Workaround for https://github.com/ariakit/ariakit/issues/6344
  // Remove once the fix lands: getPersistentElements alone should keep
  // interactions with persistent elements from closing the dialog.
  hideOnInteractOutside={(event) => {
    const notifications = notificationsRef.current;
    if (!notifications) return true;
    if (!(event.target instanceof Node)) return true;
    return !notifications.contains(event.target);
  }}
>
```

Keeping `getPersistentElements` is still important: it also controls the accessibility tree and `inert` handling for modal dialogs.

### DIFF
--- a/.changeset/300-dialog-persistent-elements-before-focus.md
+++ b/.changeset/300-dialog-persistent-elements-before-focus.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog), including components built on it such as [`Popover`](https://ariakit.com/reference/popover) and [`Menu`](https://ariakit.com/reference/menu), so focusing, clicking, or right-clicking elements returned by [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements) no longer closes the dialog before it has received focus, such as when it's rendered with [`autoFocusOnShow`](https://ariakit.com/reference/dialog#autofocusonshow) set to `false`.
--- a/app/src/sandbox/dialog-6344/index.react.tsx
+++ b/app/src/sandbox/dialog-6344/index.react.tsx
@@ -1,0 +1,82 @@
+import * as Ariakit from "@ariakit/react";
+import { useRef, useState } from "react";
+
+// Reproduces a non-modal dialog with getPersistentElements closing when the
+// user interacts with a persistent element before the dialog has been focused.
+// To see the bug:
+//   1. Click "Open dialog". The panel opens without taking focus
+//      (autoFocusOnShow={false}).
+//   2. Without clicking inside the dialog, click (or focus, or right-click) the
+//      "Notification field" input inside the persistent "Notifications" region.
+//   3. The dialog closes, even though the region is returned by
+//      getPersistentElements and should be treated as part of the dialog.
+// For contrast, focusing "Inside field" first makes the same interaction keep
+// the dialog open, as documented.
+export default function Example() {
+  const [open, setOpen] = useState(false);
+  const notificationsRef = useRef<HTMLDivElement>(null);
+
+  return (
+    <div className="flex flex-col items-start gap-3">
+      <Ariakit.Button
+        className="rounded bg-blue-600 px-3 py-1 text-white"
+        onClick={() => setOpen(true)}
+      >
+        Open dialog
+      </Ariakit.Button>
+
+      <Ariakit.Dialog
+        open={open}
+        onClose={() => setOpen(false)}
+        modal={false}
+        autoFocusOnShow={false}
+        getPersistentElements={() => {
+          const notifications = notificationsRef.current;
+          return notifications ? [notifications] : [];
+        }}
+        className="flex w-72 flex-col items-start gap-3 rounded-lg border border-gray-300 bg-white p-4 shadow-lg"
+      >
+        <Ariakit.DialogHeading className="text-lg font-medium">
+          Dialog
+        </Ariakit.DialogHeading>
+        <input
+          aria-label="Inside field"
+          placeholder="Inside field"
+          className="rounded border border-gray-300 px-3 py-1"
+        />
+        <Ariakit.DialogDismiss className="rounded border border-gray-300 px-3 py-1">
+          Close dialog
+        </Ariakit.DialogDismiss>
+      </Ariakit.Dialog>
+
+      {/* Persistent notifications region, like the dialog-react-toastify
+          example. Interacting with it must never close the dialog. */}
+      <div
+        ref={notificationsRef}
+        role="region"
+        aria-label="Notifications"
+        className="flex flex-col items-start gap-3 rounded-lg border border-gray-300 p-4"
+      >
+        <input
+          aria-label="Notification field"
+          placeholder="Notification field"
+          className="rounded border border-gray-300 px-3 py-1"
+        />
+        <button
+          type="button"
+          className="rounded border border-gray-300 px-3 py-1"
+        >
+          Dismiss notification
+        </button>
+      </div>
+
+      {/* A regular outside element: interacting with it must still close the
+          dialog. */}
+      <input
+        aria-label="Outside field"
+        placeholder="Outside field"
+        className="rounded border border-gray-300 px-3 py-1"
+      />
+    </div>
+  );
+}
--- a/app/src/sandbox/dialog-6344/index.react.tsx
+++ b/app/src/sandbox/dialog-6344/index.react.tsx
@@ -34,15 +34,6 @@ export default function Example() {
           const notifications = notificationsRef.current;
           return notifications ? [notifications] : [];
         }}
-        // TODO: Workaround for https://github.com/ariakit/ariakit/issues/6344
-        // Remove once the fix lands: getPersistentElements alone should keep
-        // interactions with persistent elements from closing the dialog.
-        hideOnInteractOutside={(event) => {
-          const notifications = notificationsRef.current;
-          if (!notifications) return true;
-          if (!(event.target instanceof Node)) return true;
-          return !notifications.contains(event.target);
-        }}
         className="flex w-72 flex-col items-start gap-3 rounded-lg border border-gray-300 bg-white p-4 shadow-lg"
       >
         <Ariakit.DialogHeading className="text-lg font-medium">
--- a/app/src/sandbox/dialog-6344/index.react.tsx
+++ b/app/src/sandbox/dialog-6344/index.react.tsx
@@ -34,6 +34,15 @@ export default function Example() {
           const notifications = notificationsRef.current;
           return notifications ? [notifications] : [];
         }}
+        // TODO: Workaround for https://github.com/ariakit/ariakit/issues/6344
+        // Remove once the fix lands: getPersistentElements alone should keep
+        // interactions with persistent elements from closing the dialog.
+        hideOnInteractOutside={(event) => {
+          const notifications = notificationsRef.current;
+          if (!notifications) return true;
+          if (!(event.target instanceof Node)) return true;
+          return !notifications.contains(event.target);
+        }}
         className="flex w-72 flex-col items-start gap-3 rounded-lg border border-gray-300 bg-white p-4 shadow-lg"
       >
         <Ariakit.DialogHeading className="text-lg font-medium">
--- a/app/src/sandbox/dialog-6344/test-browser.ts
+++ b/app/src/sandbox/dialog-6344/test-browser.ts
@@ -1,0 +1,64 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6344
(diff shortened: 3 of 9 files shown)


diegohaz:
@codex

diegohaz (commented):
Initial review, baseline `0cdb9864926a3835a6a456906f7e2c64fe4a5ae5` to head `441089c1051d9d5104ac859ae6e1553d7984a7ce`.

Scope reviewed:
- PR description and linked issue #6344.
- Full diff for the dialog mark/outside-event utilities, `useDialog`, new sandbox tests, and changeset.
- Surrounding cleanup/orchestration code, nested/persistent element semantics, tooltip/menu disclosure behavior, and the focus-reset regression around the `focusedRef` gate.

Validation:
- `pnpm test-react app/src/sandbox/dialog-6344/test.ts`
- `pnpm test-react packages/ariakit-react-components/src/dialog/utils/orchestrate.test.ts`
- `APP_PORT=4324 NEXTJS_PORT=3002 pnpm -F app run test-chrome src/sandbox/dialog-6344/test-browser.ts`
- `APP_PORT=4324 NEXTJS_PORT=3002 pnpm -F app run test-firefox src/sandbox/dialog-6344/test-browser.ts`
- `APP_PORT=4324 NEXTJS_PORT=3002 pnpm -F app run test-safari src/sandbox/dialog-6344/test-browser.ts`
- `pnpm test-react examples/menu-tooltip-various/test.ts`
- `pnpm test-react app/src/sandbox/dialog-reopen-focus-state/test.ts`
- `pnpm tsc`
- `pnpm lint`

Result: no substantive findings. The id-scoped inside mark is restored through the existing cleanup stack, persistent element interactions are now guarded before the focus-history gate, ordinary outside interactions still close, and the tooltip/disclosure regressions I checked still pass.

Not fully validated: full browser suite and full package build.

PR #6596, opened 2026-07-02
Fix Command stuck active state when focus moves away mid-press

### LINKED ISSUE #6340: Command: non-native element stays stuck pressed (data-active) when focus moves away while Space is held
### Current behavior

When [`Command`](https://ariakit.com/reference/command) renders a non-native element (e.g. `render={<div />}`, the documented usage for accessible custom buttons), pressing Space marks it as pressed via the `data-active` attribute — the documented styling hook for the pressed state. Two things then diverge from native `<button>` semantics:

1. **Stuck pressed state.** If focus moves away while Space is still held (the user clicks elsewhere with the mouse, a handler opens a dialog that grabs focus, a composite `move()`s, Tab while Space is down…), the Space keyup is delivered to the new focus target. The Command's `onKeyUp` never runs, and the element keeps `data-active` — it stays rendered as pressed indefinitely, until the user happens to press and release Space on it again. Native buttons cancel Space activation when they lose focus before keyup.

2. **Click fired from a child's keyup.** If the Command contains a focusable child (e.g. a clickable card with a nested action button) and focus moves into that child mid-press, the Space keyup bubbles up from the child and fires a synthetic click on the Command itself — activating it even though the press never finished there.

Root cause, in [`command.tsx`](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx): the Space keydown sets the pressed state ([command.tsx#L134-L140](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L134-L140)), and the only paths that clear it are the element's own `onKeyUp` ([command.tsx#L146-L175](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L146-L175)) and a layout effect for the element becoming `disabled` mid-press ([command.tsx#L87-L91](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L87-L91)). Nothing clears it on blur — [`Focusable`](https://ariakit.com/reference/focusable)'s `onBlur` only does focus-visible cleanup ([focusable.tsx#L370-L383](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L370-L383)).

For the second leg: the keydown only sets the pressed state when the event is self-targeted ([command.tsx#L99-L104](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L99-L104)), but `onKeyUp` never checks `isSelfTarget`, so a keyup bubbling from a child passes all its guards and `fireClickEvent(element, …)` clicks the Command ([command.tsx#L171-L174](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L171-L174)).

There's a third, related short-circuit: the clearing block introduced by #6221 ([command.tsx#L157-L165](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L157-L165)) sits *below* the `defaultPrevented` bail-out at [command.tsx#L149](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/command/command.tsx#L149), so a consumer calling `event.preventDefault()` in their own `onKeyUp` also leaves the element stuck pressed — the one short-circuit that block (deliberately hoisted above the `disabled`/`metaKey` guards in #6221, extended in #6271 for self-disable) doesn't cover.

This affects every non-native Command-derived component; #6221 and #6271 fixed the Meta-on-release and self-disable variants of the same stuck state, and the blur case is the same mechanism (keyup delivered elsewhere) left unhandled.

3. Press <kbd>Tab</kbd> to focus the **Save** command (a `div` rendered by `Command`).
4. Press and hold <kbd>Space</kbd>: the command renders as pressed (orange, `data-active`).
5. While still holding <kbd>Space</kbd>, click the "Outside text" paragraph with the mouse: focus leaves the command.
6. Release <kbd>Space</kbd>. **Bug 1:** Save keeps `data-active` and stays rendered as pressed indefinitely.
7. Click the **Open article** card once ("card clicks: 1" — the normal mouse click).
8. With the card focused, press and hold <kbd>Space</kbd>, then — still holding it — click the nested **Pin** button ("pins: 1"; the card counter doesn't change because the button stops propagation).
9. Release <kbd>Space</kbd>. **Bug 2:** the output jumps to "card clicks: 2" — the keyup bubbling from the Pin button fired a synthetic click on the card, even though the press never finished on it.

```tsx

  const [cardClicks, setCardClicks] = useState(0);
  const [pins, setPins] = useState(0);
  return (
    <div style={{ display: "grid", gap: 16, padding: 16 }}>
      <style>{`
        .pressable { padding: 8px 16px; border: 1px solid gray; border-radius: 4px; }
        .pressable[data-active] { background: orange; }
      `}</style>
      <Ariakit.Command className="pressable" render={<div />}>
        Save
      </Ariakit.Command>
      <Ariakit.Command
        className="pressable"
        render={<div />}
        onClick={() => setCardClicks((count) => count + 1)}
      >
        <span>Open article</span>{" "}
        <button
          onClick={(event) => {
            // The nested action shouldn't trigger the card click.
            event.stopPropagation();
            setPins((count) => count + 1);
          }}
        >
          Pin
        </button>
      </Ariakit.Command>
      <p>Outside text</p>
      <output>
        card clicks: {cardClicks}, pins: {pins}
      </output>
    </div>
  );
}
```

Both behaviors were verified with a real-browser Playwright test using only user-level interactions (Tab, `keyboard.down("Space")`, mouse clicks, `keyboard.up("Space")`).

### Expected behavior

Native buttons cancel Space activation when they lose focus before the keyup, and the non-native emulation exists to reproduce native semantics. So: the pressed `data-active` state should be cleared as soon as the Command loses focus mid-press, and a Space keyup bubbling from a child should not dispatch a click on the Command (the keydown already requires `isSelfTarget`, so the keyup should too). After step 6 the Save command should have no `data-active`, and after step 9 the card click counter should still read 1.

### Workaround

Cancel the in-flight press from a consumer `onBlur` by dispatching a synthetic Space keyup with the Meta modifier held: `Command` clears its pressed state on any Space keyup, and its `metaKey` guard prevents the synthetic click from firing. This also fixes the child-keyup leg, because the blur fires (and clears the internal pressed flag) before the real keyup bubbles up from the child.

```tsx
function cancelSpacePressOnBlur(event: React.FocusEvent<HTMLElement>) {
  event.currentTarget.dispatchEvent(
    new KeyboardEvent("keyup", { key: " ", metaKey: true, bubbles: true }),
  );
}

<Ariakit.Command render={<div />} onBlur={cancelSpacePressOnBlur}>
  Save
</Ariakit.Command>
```

Verified: with this handler on both commands, the stuck pressed state and the child-keyup click no longer happen.

### Possible solutions

Three coordinated changes in `command.tsx`:

1. **Clear the pressed state on blur.** Add an `onBlur` (bubble phase, chained before `useFocusable` consumes the props) that resets `activeRef`/`setActive(false)` when `activeRef.current` is set. It should not check `event.defaultPrevented` or `isFocusEventOutside` (focus moving into a descendant should also cancel the press). The bubble phase composes with virtual focus: `CompositeItem`'s `onBlurCapture` stops propagation of the intermediate silent-focus blur, while a genuine `move()` dispatches a bubbling `focusout` that does reach it.
2. **Hoist the clearing in `onKeyUp` above the `defaultPrevented` bail-out**, moving the `isSpace`/`activeRef` guards and the `nativeClick` computation along with it (moving only the clearing block would make any keyup — e.g. tapping Shift while Space is held — reset the press).
3. **Add the missing `isSelfTarget` guard** before firing the synthetic click in `onKeyUp`, mirroring the keydown.

```diff
     const onKeyUp = useEvent((event: KeyboardEvent<HTMLType>) => {
       onKeyUpProp?.(event);

-      if (event.defaultPrevented) return;
       if (isDuplicate) return;

       const isSpace = clickOnSpace && event.key === " ";
       if (!activeRef.current || !isSpace) return;

       const nativeClick = isNativeClick(event);

-      // Clear the active state as soon as Space is released, before the
-      // `disabled` and `metaKey` guards below, so a short-circuited keyup (Meta
-      // still held on release, or the element becoming disabled between keydown
-      // and keyup) can't leave the element stuck in a visually "pressed" state.
-      // Firing the synthetic click is still gated by those guards.
+      // Clear the active state as soon as Space is released, before all the
+      // guards below (defaultPrevented, self-target, disabled, metaKey), so a
+      // short-circuited keyup can't leave the element stuck looking pressed.
       activeRef.current = false;
       if (!nativeClick) {
         setActive(false);
       }

+      if (event.defaultPrevented) return;
+      // The keydown only sets the pressed state when self-targeted, so a keyup
+      // bubbling from a child (focus moved into a child mid-press) must not
+      // dispatch a synthetic click on this element.
+      if (!isSelfTarget(event)) return;
       if (disabled) return;
       if (event.metaKey) return;
       if (nativeClick) return;
       // …fires the synthetic click
     });

+    const onBlurProp = props.onBlur;
+
+    // If focus moves away while Space is held, the keyup is delivered to the
+    // new focus target and `onKeyUp` above never runs. Clear the pressed state
+    // so the element doesn't stay stuck looking pressed, mirroring how native
+    // buttons cancel Space activation when they lose focus before keyup.
+    const onBlur = useEvent((event: FocusEvent<HTMLType>) => {
+      onBlurProp?.(event);
+      if (!activeRef.current) return;
+      activeRef.current = false;
+      setActive(false);
+    });
```

(plus `onBlur` added to the returned props object next to `onKeyDown`/`onKeyUp`, and `FocusEvent` added to the type imports). For native buttons the new `onBlur` only resets `activeRef` (`setActive(true)` is never called for native clicks), so native Space handling is untouched. The existing `command-6217` sandbox tests pass unchanged under this fix: focus stays on the command in the Meta case, and the self-disable case is cleared by the layout effect before any blur.

diegohaz:
## Motivation

When [`Command`](https://ariakit.com/reference/command) renders a non-native element (e.g. `render={<div />}`), pressing <kbd>Space</kbd> marks it as pressed via the `data-active` attribute. The only paths that cleared that state were the element's own `onKeyUp` handler and a layout effect for the element becoming `disabled` mid-press. Nothing cleared it on blur, which diverges from native `<button>` semantics in two user-visible ways (reported in #6340):

1. **Stuck pressed state.** If focus moves away while <kbd>Space</kbd> is still held (mouse click elsewhere, a handler moving focus, <kbd>Tab</kbd>…), the keyup is delivered to the new focus target, the Command's `onKeyUp` never runs, and the element keeps `data-active` indefinitely. Native buttons cancel Space activation when they lose focus before the keyup.
2. **Click fired from a child's keyup.** If focus moves into a focusable child mid-press (e.g. a clickable card with a nested action button), the Space keyup bubbles up from the child and fires a synthetic click on the Command itself — activating it even though the press never finished there. The keydown requires `isSelfTarget`, but the keyup did not.

There was also a third, related short-circuit: the state-clearing in `onKeyUp` sat below the `event.defaultPrevented` bail-out, so a consumer calling `preventDefault()` in their own `onKeyUp` (to suppress the click on release) also left the element stuck pressed.

## Solution

Three coordinated changes in `command.tsx`, affecting `Command` and every component built on it (`Button`, `Checkbox`, `CompositeItem`, and their derivatives):

1. **Clear the pressed state on blur.** A new chained `onBlur` (bubble phase, composed before `useFocusable` consumes the props) resets the internal pressed flag and `data-active` when focus leaves mid-press. It deliberately doesn't check `isFocusEventOutside`: focus moving into a descendant also cancels the press, matching native buttons. For native buttons the handler only resets the internal flag (the visual state is never set for native clicks), so native Space handling is untouched. The bubble phase composes with composite virtual focus: `CompositeItem`'s `onBlurCapture` stops propagation of intermediate silent-focus blurs, while a genuine `move()` dispatches a bubbling `focusout` that does reach the handler.
2. **Hoist the state-clearing in `onKeyUp` above the `defaultPrevented` bail-out**, so a consumer `preventDefault()` only suppresses the click without leaving the element stuck pressed (the same treatment #6221 and #6271 gave the `metaKey` and `disabled` short-circuits).
3. **Guard the synthetic click with `isSelfTarget`**, mirroring the keydown. With change 1 this is defense-in-depth: focusing a child always blurs the element first, so the bubbled keyup already finds the pressed flag cleared — which is also why no test can isolate this guard through user-level interaction.

The repro sandbox `app/src/sandbox/command-6340/` covers all three legs with a Playwright `test-browser.ts` (Chrome, Firefox, and Safari) and a happy-dom `test.ts` duplicate, all verified failing before the fix and passing with it. Full `test-react` (805/805), `test-react18` (797 passed, 8 pre-existing skips), the `command-6217` regression sandbox, and `tsc` all pass.

Out of scope: virtual-focus composite items changing via arrow keys while <kbd>Space</kbd> is held are unaffected by this change (with virtual focus the item is never DOM-focused, so the pressed state is never set on it in the first place — the keydown's `isSelfTarget` guard already prevents it).

## Workaround

Until the fix is released, consumers can cancel the in-flight press from their own `onBlur` by dispatching a synthetic <kbd>Space</kbd> keyup with the Meta modifier held. `Command` clears its in-flight pressed state on a Space keyup, and its `metaKey` guard prevents the synthetic click from firing. This also covers the child-keyup leg, because the blur fires (and clears the internal pressed flag) before the real keyup bubbles up from the child.

```tsx
// TODO: Remove this workaround once the fix for
// https://github.com/ariakit/ariakit/issues/6340 lands.
function cancelSpacePressOnBlur(event: React.FocusEvent<HTMLElement>) {
  event.currentTarget.dispatchEvent(
    new KeyboardEvent("keyup", { key: " ", metaKey: true, bubbles: true }),
  );
}

<Ariakit.Command render={<div />} onBlur={cancelSpacePressOnBlur}>
  Save
</Ariakit.Command>
```

Fixes #6340

### DIFF
--- a/.changeset/300-command-blur-active-state.md
+++ b/.changeset/300-command-blur-active-state.md
@@ -1,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed `Command` stuck pressed state when losing focus mid-press
+
+When rendering a non-native element (such as `render={<div />}`), the [`Command`](https://ariakit.com/reference/command) component — and components built on it, such as [`Button`](https://ariakit.com/reference/button), [`Checkbox`](https://ariakit.com/reference/checkbox), [`CompositeItem`](https://ariakit.com/reference/composite-item), and their derivatives — now clears its pressed state (`data-active`) when the element loses focus while <kbd>Space</kbd> is held, mirroring how native buttons cancel the Space activation when they lose focus before the keyup.
+
+Additionally, a Space keyup bubbling up from a focused child no longer dispatches a synthetic click on the element, and calling `event.preventDefault()` in a custom `onKeyUp` handler no longer leaves the element stuck looking pressed.
--- a/app/src/sandbox/command-6340/index.react.tsx
+++ b/app/src/sandbox/command-6340/index.react.tsx
@@ -1,18 +1,35 @@
 import * as Ariakit from "@ariakit/react";
+import type { FocusEvent } from "react";
 import { useState } from "react";
 import "./style.css";
+
+// TODO: Remove this workaround once the fix for
+// https://github.com/ariakit/ariakit/issues/6340 lands. Command clears its
+// in-flight pressed state on a Space keyup, and its metaKey guard prevents the
+// synthetic click from firing, so dispatching this synthetic keyup on blur
+// cancels an in-flight press when focus moves away while Space is held.
+function cancelSpacePressOnBlur(event: FocusEvent<HTMLElement>) {
+  event.currentTarget.dispatchEvent(
+    new KeyboardEvent("keyup", { key: " ", metaKey: true, bubbles: true }),
+  );
+}
 
 export default function Example() {
   const [cardClicks, setCardClicks] = useState(0);
   const [pins, setPins] = useState(0);
   return (
     <div>
-      <Ariakit.Command className="button" render={<div />}>
+      <Ariakit.Command
+        className="button"
+        render={<div />}
+        onBlur={cancelSpacePressOnBlur}
+      >
         Save
       </Ariakit.Command>
       <Ariakit.Command
         className="button"
         render={<div />}
+        onBlur={cancelSpacePressOnBlur}
         onClick={() => setCardClicks((count) => count + 1)}
       >
         Open article{" "}
--- a/app/src/sandbox/command-6340/index.react.tsx
+++ b/app/src/sandbox/command-6340/index.react.tsx
@@ -1,35 +1,18 @@
 import * as Ariakit from "@ariakit/react";
-import type { FocusEvent } from "react";
 import { useState } from "react";
 import "./style.css";
-
-// TODO: Remove this workaround once the fix for
-// https://github.com/ariakit/ariakit/issues/6340 lands. Command clears its
-// in-flight pressed state on a Space keyup, and its metaKey guard prevents the
-// synthetic click from firing, so dispatching this synthetic keyup on blur
-// cancels an in-flight press when focus moves away while Space is held.
-function cancelSpacePressOnBlur(event: FocusEvent<HTMLElement>) {
-  event.currentTarget.dispatchEvent(
-    new KeyboardEvent("keyup", { key: " ", metaKey: true, bubbles: true }),
-  );
-}
 
 export default function Example() {
   const [cardClicks, setCardClicks] = useState(0);
   const [pins, setPins] = useState(0);
   return (
     <div>
-      <Ariakit.Command
-        className="button"
-        render={<div />}
-        onBlur={cancelSpacePressOnBlur}
-      >
+      <Ariakit.Command className="button" render={<div />}>
         Save
       </Ariakit.Command>
       <Ariakit.Command
         className="button"
         render={<div />}
-        onBlur={cancelSpacePressOnBlur}
         onClick={() => setCardClicks((count) => count + 1)}
       >
         Open article{" "}
@@ -42,6 +25,18 @@ export default function Example() {
         >
           Pin
         </button>
+      </Ariakit.Command>
+      <Ariakit.Command
+        className="button"
+        render={<div />}
+        onKeyUp={(event) => {
+          // Suppresses the click on release, a way to opt out of the default
+          // keyup behavior. It must not leave the element stuck looking
+          // pressed.
+          event.preventDefault();
+        }}
+      >
+        Bookmark
       </Ariakit.Command>
       <p>Outside text</p>
       <output>
--- a/app/src/sandbox/command-6340/index.react.tsx
+++ b/app/src/sandbox/command-6340/index.react.tsx
@@ -1,0 +1,35 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+import "./style.css";
+
+export default function Example() {
+  const [cardClicks, setCardClicks] = useState(0);
+  const [pins, setPins] = useState(0);
+  return (
+    <div>
+      <Ariakit.Command className="button" render={<div />}>
+        Save
+      </Ariakit.Command>
+      <Ariakit.Command
+        className="button"
+        render={<div />}
+        onClick={() => setCardClicks((count) => count + 1)}
+      >
+        Open article{" "}
+        <button
+          onClick={(event) => {
+            // The nested action shouldn't trigger the card click.
+            event.stopPropagation();
+            setPins((count) => count + 1);
+          }}
+        >
+          Pin
+        </button>
+      </Ariakit.Command>
+      <p>Outside text</p>
+      <output>
+        card clicks: {cardClicks}, pins: {pins}
+      </output>
+    </div>
+  );
+}
--- a/app/src/sandbox/command-6340/style.css
+++ b/app/src/sandbox/command-6340/style.css
@@ -1,0 +1,11 @@
+.button {
+  display: inline-flex;
+  cursor: pointer;
+  padding: 8px 16px;
+  border-radius: 6px;
+  background-color: #ddd;
+}
+
+.button[data-active] {
+  background-color: #aaa;
+}
--- a/app/src/sandbox/command-6340/test-browser.ts
+++ b/app/src/sandbox/command-6340/test-browser.ts
@@ -62,4 +62,22 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.status()).toHaveText("card clicks: 0, pins: 0");
     await test.expect(card).not.toHaveAttribute("data-active");
   });
+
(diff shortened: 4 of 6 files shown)


diegohaz:
@codex

PR #6597, opened 2026-07-02
Fix dialog backdrop exit transition when only the backdrop is animated

### LINKED ISSUE #6339: Dialog backdrop exit transition is skipped when only the backdrop is animated (panel has no transition)
### Current behavior

When a [`Dialog`](https://ariakit.com/reference/dialog)'s [`backdrop`](https://ariakit.com/reference/dialog#backdrop) has a CSS transition but the dialog panel itself does not (e.g. an overlay that fades while the panel snaps in place), the backdrop's enter fade plays, but its exit fade is silently skipped: on close, the dialog hides instantly and `data-leave` is never applied to the backdrop.

The cause is an asymmetry between the two `useDisclosureContent` instances (panel and backdrop), which share one animation state. `DialogBackdrop` creates its store with the dialog store as `disclosure` ([dialog-backdrop.tsx#L25](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/dialog-backdrop.tsx#L25)), so `open`, `animated`, and `animating` are synced between the two stores ([disclosure-store.ts#L20-L23](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/disclosure/disclosure-store.ts#L20-L23)).

The *backdrop* instance already handles the converse case: it unions the dialog panel's computed transition/animation styles via `otherElement`, with a comment noting "the dialog may be animated while the backdrop is not" ([disclosure-content.tsx#L201-L224](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx#L201-L224)). But the *dialog* instance ([dialog.tsx#L593](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/dialog.tsx#L593)) has no `disclosure` store, so `otherElement` is undefined and its timeout is computed from the panel's styles alone. With an unanimated panel, the timeout is 0 and the zero-timeout branch on enter sets the **shared** `animated` state to `false` ([disclosure-content.tsx#L225-L237](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx#L225-L237)), which also cancels the backdrop's pending 500ms timeout.

The enter fade still looks fine because `data-enter` was applied before `animated` was zeroed. The damage shows on close: with `animated` false, the store never sets `animating` and flips `mounted` to `false` immediately ([disclosure-store.ts#L57-L68](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/disclosure/disclosure-store.ts#L57-L68)), so the backdrop is hidden instantly, `data-leave` is never set, and the fade-out never plays.

3. Click "Show dialog". The backdrop fades in over 500ms (the enter direction works).
4. Click "Close". The backdrop should fade out over 500ms, but the dialog and backdrop disappear instantly — no exit fade, and the backdrop never receives `data-leave`.

```tsx

const css = `
  .backdrop {
    background: rgb(0 0 0 / 0.4);
    opacity: 0;
    transition: opacity 500ms ease;
  }
  .backdrop[data-enter] {
    opacity: 1;
  }
  .dialog {
    position: fixed;
    inset: 0.75rem;
    z-index: 50;
    margin: auto;
    display: flex;
    flex-direction: column;
    gap: 1rem;
    width: 24rem;
    height: fit-content;
    max-width: calc(100vw - 1.5rem);
    border-radius: 1rem;
    background: white;
    color: black;
    padding: 1.5rem;
  }
`;

  const dialog = Ariakit.useDialogStore();
  return (
    <>
      <style>{css}</style>
      <Ariakit.Button onClick={dialog.show}>Show dialog</Ariakit.Button>
      <Ariakit.Dialog
        store={dialog}
        backdrop={<div className="backdrop" />}
        className="dialog"
      >
        <Ariakit.DialogHeading>Success</Ariakit.DialogHeading>
        <p>
          Only the backdrop is animated: it should fade in when the dialog
          opens and fade out when it closes. The panel has no transitions.
        </p>
        <Ariakit.DialogDismiss>Close</Ariakit.DialogDismiss>
      </Ariakit.Dialog>
    </>
  );
}
```

This was verified with a real-browser Playwright test: after clicking Close, the backdrop never receives `data-leave` and is already `display: none` within 100ms of the click, well inside the 500ms transition window, while `data-enter` and the fade-in work normally.

### Expected behavior

Animating only the backdrop should work the same way as animating only the panel (which the code explicitly supports in the other direction): on close, the backdrop should receive `data-leave`, remain visible while its transition runs, and the dialog should only unmount/hide after the backdrop's transition ends — mirroring how an animated panel keeps an unanimated backdrop mounted.

### Workaround

Add a visually imperceptible transition to the dialog panel matching the backdrop's duration. The panel's opacity never changes, so nothing is visible, but Ariakit then computes a non-zero animation timeout for the panel and keeps the shared animation state alive:

```css
.dialog {
  /* same duration as the backdrop's transition */
  transition: opacity 500ms ease;
}
```

With this in place, the backdrop gets `data-leave` and fades out as expected (verified with the same Playwright test).

### Possible solutions

Give the dialog's `useDisclosureContent` instance the reciprocal information the backdrop instance already has. The backdrop isn't tracked in the dialog store, but it renders with a `data-backdrop` attribute set to the dialog's id ([dialog-backdrop.tsx#L47-L61](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/dialog-backdrop.tsx#L47-L61)), an internal contract already relied on by [is-backdrop.ts](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/dialog/utils/is-backdrop.ts#L1-L12). Inside the timeout effect, query for that element and include its computed styles in the `Math.max(...)` union:

```diff
--- a/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
+++ b/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
@@
-import { invariant, removeUndefinedValues } from "@ariakit/utils";
+import { getDocument, invariant, removeUndefinedValues } from "@ariakit/utils";
@@
     } = otherElement ? getComputedStyle(otherElement) : {};
+    // Conversely, if we're rendering a dialog panel, its backdrop may be
+    // animated while the panel is not. The backdrop element isn't tracked in
+    // the store, but it's rendered with a data-backdrop attribute set to the
+    // dialog id, so we can find it in the DOM.
+    const backdropElement = contentElement.id
+      ? getDocument(contentElement).querySelector<HTMLElement>(
+          `[data-backdrop="${CSS.escape(contentElement.id)}"]`,
+        )
+      : null;
+    const {
+      transitionProperty: transitionProperty3 = "",
+      transitionDuration: transitionDuration3 = "0",
+      transitionDelay: transitionDelay3 = "0",
+      animationName: animationName3 = "",
+      animationDuration: animationDuration3 = "0",
+      animationDelay: animationDelay3 = "0",
+    } = backdropElement ? getComputedStyle(backdropElement) : {};
@@
       getEndTime(transitionProperty2, transitionDelay2, transitionDuration2),
       getEndTime(animationName2, animationDelay2, animationDuration2),
+      getEndTime(transitionProperty3, transitionDelay3, transitionDuration3),
+      getEndTime(animationName3, animationDelay3, animationDuration3),
     );
```

Notes: a backdrop-side fix alone can't work, because the dialog instance's synchronous `stopAnimation()` at timeout 0 on leave would still collapse the shared `animating` flag (and `mounted`) before the backdrop's timer fires. The `contentElement.id` guard avoids matching an unrelated backdrop while it briefly renders with `data-backdrop=""`. For non-dialog disclosure content and for the backdrop's own instance, the query returns null, so behavior is unchanged, and the fast-unmount path for fully unanimated dialogs is preserved. A cleaner long-term alternative is tracking a `backdropElement` in the dialog store state (analogous to `contentElement`), at the cost of a larger cross-package change.

diegohaz:
Fixes #6339

## Motivation

When a [`Dialog`](https://ariakit.com/reference/dialog)'s [`backdrop`](https://ariakit.com/reference/dialog#backdrop) has a CSS transition but the dialog panel itself doesn't (an overlay that fades while the panel snaps in place), the backdrop's enter fade plays, but its exit fade is silently skipped: on close, the dialog hides instantly and `data-leave` is never applied to the backdrop.

```css
.backdrop {
  opacity: 0;
  transition: opacity 500ms ease;
}
.backdrop[data-enter] {
  opacity: 1;
}
/* .dialog has no transition */
```

The two `useDisclosureContent` instances (panel and backdrop) share one animation state, but they're asymmetric. The backdrop instance already unions the panel's computed transition/animation styles via `otherElement`, since "the dialog may be animated while the backdrop is not". The panel instance, however, knows nothing about the backdrop: with an unanimated panel it computes a zero animation timeout, and the zero-timeout branch sets the **shared** `animated` state to `false`. With `animated` false, closing never sets `animating`, so `mounted` flips to `false` immediately — the backdrop is hidden instantly and never receives `data-leave`.

## Solution

Give the panel's `useDisclosureContent` instance the reciprocal information the backdrop instance already has. `useDialog` now owns a `backdropRef` that `DialogBackdrop` merges into the backdrop element, and hands it to its own `useDisclosureContent` call through a private `unstable_otherElementRef` option — the reciprocal of how the backdrop instance already learns about the panel via `otherElement`. The timeout effect includes that element's computed transition/animation end times in the timeout computation, so the disclosure layer stays generic and no DOM query is needed.

With the panel computing a non-zero timeout, the zero-timeout branch no longer clears the shared `animated` state on enter, and on leave the panel waits for the backdrop's transition instead of stopping the animation synchronously. For non-dialog disclosure content and for the backdrop's own instance the ref is empty, so behavior is unchanged, and the fast-unmount path for fully unanimated dialogs is preserved. This also fixes the related case where both elements are animated but the backdrop's transition is longer than the panel's: the panel instance previously stopped the shared animation at its own shorter timeout, cutting the backdrop's fade short.

Both cases are covered by the `dialog-backdrop-6339` sandbox — a backdrop-only dialog and a "fast" dialog with a 150ms panel transition next to the 500ms backdrop — with Playwright browser tests and happy-dom duplicates that fail without the fix at the `data-leave` assertions.

## Workaround

Until the fix is released, add a visually imperceptible transition to the dialog panel matching the backdrop's duration. The panel's opacity never changes, so nothing is visible, but Ariakit then computes a non-zero animation timeout for the panel and keeps the shared animation state alive, letting the backdrop fade out on close:

```css
.dialog {
  /* TODO: Workaround for https://github.com/ariakit/ariakit/issues/6339
     Remove once the fix lands. Use the same duration as the backdrop's
     transition. */
  transition: opacity 500ms ease;
}
```

### DIFF
--- a/app/src/sandbox/dialog-backdrop-6339/test.ts
+++ b/app/src/sandbox/dialog-backdrop-6339/test.ts
@@ -1,0 +1,24 @@
+// See https://github.com/ariakit/ariakit/issues/6339
+import { click, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+test("backdrop fades out on close when only the backdrop is animated", async () => {
+  await click(q.button.ensure("Show dialog"));
+  expect(q.dialog()).toBeVisible();
+  const backdrop = q.presentation.ensure();
+  // The enter state is applied after a couple of frames.
+  await expect.poll(() => backdrop.getAttribute("data-enter")).toBe("true");
+  // The Close button is auto-focused when the dialog opens.
+  expect(q.button("Close")).toHaveFocus();
+  await click(q.button("Close"));
+  // Focus returns to the disclosure as soon as the dialog closes, so a failure
+  // below unambiguously points at the backdrop leave transition.
+  expect(q.button("Show dialog")).toHaveFocus();
+  // On close, the backdrop must receive data-leave and remain visible while
+  // its 500ms exit transition runs. Before the fix, the dialog hides instantly
+  // and data-leave is never applied.
+  await expect.poll(() => backdrop.getAttribute("data-leave")).toBe("true");
+  expect(backdrop).not.toHaveStyle("display: none");
+  // After the transition ends, the backdrop hides.
+  await expect.poll(() => backdrop.style.display).toBe("none");
+});
--- a/app/src/sandbox/dialog-backdrop-6339/test.ts
+++ b/app/src/sandbox/dialog-backdrop-6339/test.ts
@@ -1,5 +1,5 @@
 // See https://github.com/ariakit/ariakit/issues/6339
-import { click, q } from "@ariakit/test";
+import { click, q, sleep } from "@ariakit/test";
 import { expect, test } from "vitest";
 
 test("backdrop fades out on close when only the backdrop is animated", async () => {
@@ -20,5 +20,27 @@ test("backdrop fades out on close when only the backdrop is animated", async () 
   await expect.poll(() => backdrop.getAttribute("data-leave")).toBe("true");
   expect(backdrop).not.toHaveStyle("display: none");
   // After the transition ends, the backdrop hides.
+  await expect.poll(() => backdrop.style.display).toBe("none");
+});
+
+test("backdrop finishes its longer fade out when the panel has a shorter transition", async () => {
+  await click(q.button.ensure("Show fast dialog"));
+  expect(q.dialog()).toBeVisible();
+  const backdrop = q.presentation.ensure();
+  // The enter state is applied after a couple of frames.
+  await expect.poll(() => backdrop.getAttribute("data-enter")).toBe("true");
+  // The Close button is auto-focused when the dialog opens.
+  expect(q.button("Close")).toHaveFocus();
+  await click(q.button("Close"));
+  // Focus returns to the disclosure as soon as the dialog closes.
+  expect(q.button("Show fast dialog")).toHaveFocus();
+  await expect.poll(() => backdrop.getAttribute("data-leave")).toBe("true");
+  // Wait past the panel's 150ms transition. The backdrop must keep leaving
+  // until its own 500ms transition ends. Before the fix, the panel's shorter
+  // timeout stopped the shared animation state, hiding the backdrop at 150ms.
+  await sleep(250);
+  expect(backdrop).toHaveAttribute("data-leave", "true");
+  expect(backdrop).not.toHaveStyle("display: none");
+  // After the backdrop's transition ends, it hides.
   await expect.poll(() => backdrop.style.display).toBe("none");
 });
--- a/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
+++ b/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
@@ -9,7 +9,12 @@ import {
   forwardRef,
 } from "@ariakit/react-utils";
 import type { Options, Props } from "@ariakit/react-utils";
-import { afterPaint, invariant, removeUndefinedValues } from "@ariakit/utils";
+import {
+  afterPaint,
+  getDocument,
+  invariant,
+  removeUndefinedValues,
+} from "@ariakit/utils";
 import type { ElementType } from "react";
 import { useMemo, useRef, useState } from "react";
 import { flushSync } from "react-dom";
@@ -59,6 +64,27 @@ function getEndTime(names: string, delays: string, durations: string) {
     endTime = Math.max(endTime, delay + duration);
   }
   return endTime;
+}
+
+// Returns the time an element's transitions and animations end based on its
+// computed style. We compute each transition's and animation's end time
+// separately (pairing each item's own delay with its own duration) and take
+// the longest one. Combining them into a single max delay + max duration would
+// overestimate when, for example, the longest delay belongs to a transition
+// while the longest duration belongs to an animation.
+function getElementEndTime(element: Element) {
+  const {
+    transitionProperty,
+    transitionDuration,
+    transitionDelay,
+    animationName,
+    animationDuration,
+    animationDelay,
+  } = getComputedStyle(element);
+  return Math.max(
+    getEndTime(transitionProperty, transitionDelay, transitionDuration),
+    getEndTime(animationName, animationDelay, animationDuration),
+  );
 }
 
 export function isHidden(
@@ -183,38 +209,36 @@ export const useDisclosureContent = createHook<
     // first place, the events won't fire. Besides, there may be multiple
     // transitions or animations with different durations and delays, and we
     // need to consider the longest one.
-    const {
-      transitionProperty,
-      transitionDuration,
-      transitionDelay,
-      animationName,
-      animationDuration,
-      animationDelay,
-    } = getComputedStyle(contentElement);
+    const elements = [contentElement];
     // If we're rendering a dialog backdrop, otherElement will be the dialog
     // element itself. We need to consider both the backdrop and the dialog
     // animation/transition durations and delays because the dialog may be
     // animated while the backdrop is not.
-    const {
-      transitionProperty: transitionProperty2 = "",
-      transitionDuration: transitionDuration2 = "0",
-      transitionDelay: transitionDelay2 = "0",
-      animationName: animationName2 = "",
-      animationDuration: animationDuration2 = "0",
-      animationDelay: animationDelay2 = "0",
-    } = otherElement ? getComputedStyle(otherElement) : {};
-    // We compute each transition's and animation's end time separately (pairing
-    // each item's own delay with its own duration) and take the longest one,
-    // including the other element's. Combining them into a single max delay +
-    // max duration would overestimate when, for example, the longest delay
-    // belongs to a transition while the longest duration belongs to an
-    // animation.
-    const timeout = Math.max(
-      getEndTime(transitionProperty, transitionDelay, transitionDuration),
-      getEndTime(animationName, animationDelay, animationDuration),
-      getEndTime(transitionProperty2, transitionDelay2, transitionDuration2),
-      getEndTime(animationName2, animationDelay2, animationDuration2),
-    );
+    if (otherElement) {
+      elements.push(otherElement);
+    }
+    // Conversely, if we're rendering a dialog, its backdrop may be animated
+    // while the dialog itself is not. The backdrop element isn't tracked in
+    // the store, but it's rendered with a data-backdrop attribute set to the
+    // dialog id, so we can find it in the DOM. The data-dialog guard skips
+    // the document-wide query for content that can't have a backdrop, such
+    // as plain disclosures and the backdrop's own instance, and the id guard
+    // avoids matching an unrelated backdrop while it briefly renders with
+    // data-backdrop="".
+    const isDialog = contentElement.hasAttribute("data-dialog");
+    const backdropElement =
+      isDialog && contentElement.id
+        ? getDocument(contentElement).querySelector<HTMLElement>(
+            `[data-backdrop="${CSS.escape(contentElement.id)}"]`,
+          )
+        : null;
+    if (backdropElement) {
+      elements.push(backdropElement);
+    }
+    // The timeout is the longest end time among the content element and the
+    // related elements, so none of them gets hidden before its own transition
+    // or animation ends.
+    const timeout = Math.max(...elements.map(getElementEndTime));
     // If the timeout is zero, there's no animation or transition, either
     // because they weren't defined in the CSS or the duration was explicitly
     // set to zero. In this scenario, we can halt the animation right away
--- a/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
+++ b/packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
@@ -9,13 +9,8 @@ import {
   forwardRef,
 } from "@ariakit/react-utils";
 import type { Options, Props } from "@ariakit/react-utils";
-import {
-  afterPaint,
-  getDocument,
-  invariant,
-  removeUndefinedValues,
-} from "@ariakit/utils";
-import type { ElementType } from "react";
+import { afterPaint, invariant, removeUndefinedValues } from "@ariakit/utils";
+import type { ElementType, RefObject } from "react";
 import { useMemo, useRef, useState } from "react";
 import { flushSync } from "react-dom";
 import { DialogScopedContextProvider } from "../dialog/dialog-context.tsx";
@@ -109,7 +104,12 @@ export function isHidden(
 export const useDisclosureContent = createHook<
   TagName,
   DisclosureContentOptions
->(function useDisclosureContent({ store, alwaysVisible, ...props }) {
+>(function useDisclosureContent({
+  store,
+  alwaysVisible,
+  unstable_otherElementRef: otherElementRef,
+  ...props
+}) {
   const context = useDisclosureProviderContext();
   store = store || context;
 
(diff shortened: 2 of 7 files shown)


packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
```diff
@@ -183,38 +209,36 @@ export const useDisclosureContent = createHook<
...
+    const isDialog = contentElement.hasAttribute("data-dialog");
+    const backdropElement =
+      isDialog && contentElement.id
+        ? getDocument(contentElement).querySelector<HTMLElement>(
+            `[data-backdrop="${CSS.escape(contentElement.id)}"]`,
+          )
+        : null;
+    if (backdropElement) {
+      elements.push(backdropElement);
+    }
```
diegohaz, line 226: This fixes the bug, but it moves the dialog/backdrop DOM contract (`data-dialog`, the `data-backdrop` id format, and the transient `data-backdrop=""` edge case) into the disclosure layer, and every dialog-based component pays for it: `Popover`, `Menu`, `Hovercard`, and the combobox/select popovers all render `data-dialog` with an id, so each enter/leave transition now runs a document-wide `querySelector` even though most of them never render a backdrop.

The existing `otherElement` direction isn't really precedent for this coupling: it's derived from a store relationship (`store.disclosure`'s `contentElement`) — only the comment mentions dialogs. This block is the first place `disclosure-content` hard-codes dialog DOM specifics.

Since `useDialog` renders `DialogBackdrop` itself, it can hand the backdrop element to its own `useDisclosureContent` call explicitly — the reciprocal of how the backdrop instance already learns about the panel — keeping the disclosure layer generic and deleting the query plus both guards:

```diff
 // dialog.tsx
   const ref = useRef<HTMLType>(null);
+  const backdropRef = useRef<HTMLDivElement>(null);
 ...
           <DialogBackdrop
             store={store}
             backdrop={backdrop}
+            backdropRef={backdropRef}
 ...
-  props = useDisclosureContent({ store, ...props });
+  props = useDisclosureContent({ store, otherElementRef: backdropRef, ...props });

 // dialog-backdrop.tsx
   const props = useDisclosureContent({
-    ref,
+    ref: useMergeRefs(ref, backdropRef),

 // disclosure-content.tsx (this block)
-    const isDialog = contentElement.hasAttribute("data-dialog");
-    const backdropElement = /* document-wide query */;
-    if (backdropElement) {
-      elements.push(backdropElement);
-    }
+    const relatedElement = otherElementRef?.current;
+    if (relatedElement) {
+      elements.push(relatedElement);
+    }
```

with `otherElementRef?: RefObject<HTMLElement | null>` as an `@internal` option on `DisclosureContentOptions` (naming and shape up for grabs — it could also generalize to a list of related elements).

I validated this locally on top of this PR: the new `dialog-backdrop-6339` happy-dom test, all 11 `dialog-animated-various` variants, and the full `pnpm test-react dialog` sweep (18 files, 138 tests) pass with the querySelector block replaced by the ref read. Ref timing is safe because the backdrop mounts in the same commit as the panel, and this effect only runs after `transition` is set via `afterPaint` — the same timing the querySelector approach relies on.

app/src/sandbox/dialog-backdrop-6339/test.ts
```diff
@@ -0,0 +1,24 @@
+// See https://github.com/ariakit/ariakit/issues/6339
+import { click, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+test("backdrop fades out on close when only the backdrop is animated", async () => {
```
diegohaz, line 5: The PR description says this change also fixes the case where both elements are animated but the backdrop's transition is longer than the panel's (pre-fix, the panel's shorter timeout fires `stopAnimationSync`, flipping the shared `animating`/`mounted` state and cutting the backdrop's exit short — I confirmed that flow in the code), but no test covers that scenario. It's arguably the more common real-world setup (panel scales in quickly while the overlay fades slowly), and it's exactly the case most likely to silently regress if the timeout computation is later reworked. `examples/dialog-animated-various` won't catch it either — all of its durations are 500ms.

Consider a second scenario in this sandbox: a dialog with a short transition (e.g. 150ms) alongside the 500ms backdrop, asserting that the backdrop still has `data-leave` and remains visible after the panel's own duration has elapsed, and only hides once its 500ms transition ends.

diegohaz (commented):
**Initial review** of head 9d529db (diff vs merge base 0cdb986 on `main`), covering the `disclosure-content` change, the dialog/backdrop wiring it interacts with, the store sync semantics (`mergeStore` shares `animated`/`animating`/`mounted` between the panel and backdrop stores, so the PR's "shared animation state" analysis is accurate), the new sandbox and its tests, and the changeset.

### Is the fix worth pursuing at all?

A fair alternative to weigh is not fixing this and documenting the "add a matching transition to the panel" workaround. I don't think that holds up:

- The behavior is asymmetric: the backdrop's enter fade plays and only the exit is silently skipped, with no signal as to why. That reads as a bug, not a supportable constraint.
- The workaround is a decoy transition on a property that never changes, whose duration must be manually kept in sync with the backdrop's. It exists only to keep the shared `animated` state alive, and drifts silently when the backdrop's duration changes.
- The duration-mismatch case this PR also fixes (both elements animated, backdrop longer than the panel) affects fully idiomatic CSS and can't be styled around at all.

So the fix should land; my main feedback is about its shape.

### Findings (2 inline comments)

1. **`disclosure-content.tsx`** — the fix imports the dialog/backdrop DOM contract (`data-dialog`, `data-backdrop` id format, transient `data-backdrop=""` guard) into the disclosure layer and adds a document-wide `querySelector` on every enter/leave transition of every dialog-based component (`Popover`, `Menu`, `Hovercard`, combobox/select popovers), even when no backdrop exists. Suggested a validated alternative: thread the backdrop element explicitly from `useDialog` (which already renders `DialogBackdrop`) via an internal ref option — the reciprocal of the existing `otherElement` store relationship.
2. **Missing test** for the backdrop-longer-than-panel case that the PR description claims (correctly) to fix — the case most likely to silently regress under a rework.

- **Red check**: `main`'s `disclosure-content.tsx` with this PR's tests → the happy-dom test fails at the `data-leave` assertion; the PR head passes. The repro is genuine.
- **Alternative prototype**: implemented the ref-threading approach from comment 1 on top of this branch and ran `pnpm test-react dialog` — 18 files / 138 tests pass, including the new sandbox and all `dialog-animated-various` variants. Experiment reverted afterwards.
- Did not run the Playwright browser test locally; CI covers it and all checks are green.

### Also checked

- The perf bot's `mount controlled composite +11%` flag: composite renders no disclosure content, so this is almost certainly noise, but a baseline re-run before merge would confirm.
- Changeset wording/links, sandbox naming (`dialog-backdrop-6339`), inline `<style>` rationale, and the browser + happy-dom test pairing all follow repo conventions.
- The `getEndTime`/`getElementEndTime` refactor is behavior-preserving and a genuine simplification of the previous duplicated destructuring.

packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
diegohaz: I think the prop should be `@private` (not sure) and have the `unstable_` prefix?

diegohaz: Fixed in a10d61944 - reworked as suggested: `useDialog` owns a `backdropRef`, `DialogBackdrop` merges it into the backdrop element ref, and the dialog's own `useDisclosureContent` call receives it, so the disclosure layer no longer hard-codes dialog DOM specifics and the document-wide query plus both guards are gone. Per the follow-up, the option is named `unstable_otherElementRef` and documented with `@private`, matching the `unstable_treeSnapshotKey` precedent. It's also passed after the props spread so a caller-supplied value can't silently replace the internal ref.

app/src/sandbox/dialog-backdrop-6339/test.ts
diegohaz: Fixed in da9b9bdd7 - the sandbox gains a second "fast" dialog with a 150ms panel opacity transition next to the same 500ms backdrop, and both the happy-dom and browser tests now close it, wait past the panel's duration, and assert the backdrop still has `data-leave` and remains displayed until its own transition ends. Verified red against the unfixed library: the new test fails at the after-150ms assertion because the panel's shorter timeout had already removed `data-leave`. The changeset was also broadened to mention the cut-short case.

diegohaz:
Addressed all review comments.

**Fixed**

- a10d61944 ([reply](https://github.com/ariakit/ariakit/pull/6597#discussion_r3513258882)): reworked the fix to pass the backdrop element by ref (`useDialog`-owned `backdropRef` → `DialogBackdrop` → private `unstable_otherElementRef` on `useDisclosureContent`), removing the document-wide `[data-backdrop]` query and the dialog DOM contracts from the disclosure layer. Includes the `unstable_` prefix + `@private` naming from the follow-up comment.
- da9b9bdd7 ([reply](https://github.com/ariakit/ariakit/pull/6597#discussion_r3513259475)): added the panel-shorter-than-backdrop scenario (150ms panel + 500ms backdrop) to the sandbox and both test suites, verified red against the unfixed library, and broadened the changeset to mention the cut-short case.

Verification: happy-dom sandbox tests 2/2, Playwright chrome/firefox/safari 6/6, `tsc` clean, full `test-react` 804/804, full `test-react18` green, lint/format clean. The PR description's Solution section was updated for the ref-based approach.

diegohaz (commented):
**Re-review** of head da9b9bd, covering the two commits since the last reviewed head 9d529db (`a10d6194` ref-threading rework, `da9b9bd` backdrop-longer-than-panel coverage), plus a fresh pass over the surrounding dialog/backdrop/disclosure context.

### Prior findings — both verified fixed

1. **Disclosure-layer coupling / document-wide query** (`disclosure-content.tsx`): fixed as suggested in `a10d6194`. `useDialog` owns a `backdropRef`, `DialogBackdrop` merges it into the backdrop element via `useMergeRefs` (which nulls ref objects on detach, so no stale element survives backdrop unmount), and the dialog's `useDisclosureContent` call receives it as `unstable_otherElementRef` — placed after the props spread so a caller-supplied value can't replace the internal ref. The `data-dialog`/`data-backdrop` guards and the per-transition `querySelector` are gone; `data-backdrop` itself still has its pre-existing consumer (`utils/is-backdrop.ts`) and is untouched. The follow-up naming request (`unstable_` prefix + `@private`) is applied.
2. **Missing backdrop-longer-than-panel test**: fixed in `da9b9bd`. The sandbox gains a second "fast" dialog (150ms panel opacity transition next to the 500ms backdrop), with browser + happy-dom tests asserting the backdrop keeps `data-leave` and stays visible past the panel's duration.

### Validation performed locally (at head da9b9bd)

- `pnpm test-react run dialog-backdrop-6339` — 2/2 pass.
- **Red check**: reverted `disclosure-content.tsx`, `dialog.tsx`, and `dialog-backdrop.tsx` to the merge base (0cdb986) with the PR's tests in place — both tests fail, and the new fast-dialog test fails exactly at the after-150ms `data-leave` assertion, confirming it covers the cut-short case and not just the backdrop-only case. Files restored afterwards.
- `pnpm test-react run dialog` — full dialog sweep, 18 files / 139 tests pass.
- `pnpm tsc` — clean (also confirms the `RefObject<HTMLDivElement | null>` → `RefObject<HTMLElement | null>` threading).

### Also checked

- Effect deps: `otherElementRef` added to the timeout effect deps is a stable ref object — no extra reruns.
- Enter-path behavior change: with a backdrop-only animation the panel instance now keeps `animating` true for the backdrop's enter duration instead of clearing `animated` immediately; nothing gates on that during enter (auto-focus asserted by the tests), and the fully-unanimated fast-unmount path is preserved via the zero-timeout branch.
- Test timing margins: the 250ms waits sit between the 150ms panel and 500ms backdrop durations with comfortable slack for the trailing `expect.poll`/`toBeHidden` windows.
- Perf: the Node section is clean; the `mount controlled composite +11%` Chrome flag is from the previous head's run (composite renders no disclosure content, and this head removes a document-wide query — if anything, less work). The Chrome (1/2) shard for da9b9bd was still running at review time; worth a glance when it lands, but I expect the flag to wash out.
- Changeset broadened to mention the cut-short case; wording and links check out.

No new findings. All actionable feedback from the previous review is resolved; from my side this is mergeable once the remaining Chrome shard finishes.

PR #6598, opened 2026-07-02
Fix div-based Focusable server-rendering without tabindex and with an invalid disabled attribute

### LINKED ISSUE #6341: Div-based Focusable server-renders without tabindex="0" (not keyboard-reachable until hydration) and with an invalid disabled attribute
### Current behavior

When a [Focusable](https://ariakit.com/reference/focusable)-based component renders a non-native tag (the default for `Focusable` is a `div`), the server-rendered HTML is missing the `tabIndex={0}` fallback the component relies on to be focusable. Until hydration completes — or indefinitely for users browsing with JavaScript disabled — the element is simply absent from the tab order: pressing Tab jumps straight from the previous control to the next one. With React's `renderToString`/Next.js App Router:

```html
<!-- <Focusable>Focusable card</Focusable> server-renders as: -->
<div>Focusable card</div>

<!-- client steady state after hydration: -->
<div tabindex="0">Focusable card</div>
```

A disabled `Focusable` additionally server-renders a literal `disabled` attribute on the `div`, which is invalid HTML and matches `[disabled]` CSS selectors (common in design systems and Tailwind variants) until hydration silently removes it — no hydration mismatch warning is emitted:

```html
<!-- <Focusable disabled>…</Focusable> server-renders as: -->
<div aria-disabled="true" disabled="" style="pointer-events:none">…</div>
```

The root cause is in `useFocusable`: it computes `tabIndex` and `disabled` from the element's tag name, read from the DOM via `useTagName(ref)` with no fallback type ([focusable.tsx#L410-L412](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L410-L412)). `useTagName` only learns the actual tag inside `useSafeLayoutEffect` ([hooks.ts#L236-L256](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-utils/src/hooks.ts#L236-L256)), which never runs during server rendering, so `tagName` stays `undefined` for the whole server render. Both predicates then default to "assume a native control" ([focusable.tsx#L74-L94](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L74-L94)): `isNativeTabbable(undefined)` returns `true`, so `getTabIndex` skips the `tabIndexProp ?? 0` fallback ([focusable.tsx#L153-L163](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L153-L163)), and `supportsDisabledAttribute(undefined)` returns `true`, so the stray `disabled` attribute is emitted ([focusable.tsx#L446-L454](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L446-L454)).

Every div-rendered descendant of `useFocusable` inherits this: a virtual-focus [Composite](https://ariakit.com/reference/composite) container (the widget's only tabbable element) SSRs with no `tabindex`, [TooltipAnchor](https://ariakit.com/reference/tooltip-anchor) SSRs as a bare `<div>`, and div-based composite items (e.g. `render={<div />}`) lose the roving-tabindex `tabindex="0"` on the server. The most common components (`Button`, `Tab`, `Checkbox`, `Radio`) default to native tags and are unaffected, which makes this easy to miss. This looks like an oversight rather than a design decision: sibling components already pass the fallback tag — [button.tsx#L32](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/button/button.tsx#L32), [checkbox.tsx#L91](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/checkbox/checkbox.tsx#L91), [radio.tsx#L91](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/radio/radio.tsx#L91), and [form.tsx#L149](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/form/form.tsx#L149) all call `useTagName(ref, TagName)`.

2. Replace `src/app.tsx` with the code below. It prints the exact markup `renderToString` produces — i.e. what a server (Next.js, Remix, etc.) sends to the browser.
3. Note the printed markup: the enabled `Focusable` div has no `tabindex="0"`, and the disabled one carries a literal `disabled=""` attribute.
4. Real-world equivalent: server-render this page (e.g. Next.js App Router), disable JavaScript (or test before hydration), click "Before" and press Tab — focus skips "Focusable card" and lands on "After".

```tsx
import { renderToString } from "react-dom/server";

function Content() {
  return (
    <>
      <button type="button">Before</button>
      <Ariakit.Focusable>Focusable card</Ariakit.Focusable>
      <Ariakit.Focusable disabled>Disabled focusable card</Ariakit.Focusable>
      <button type="button">After</button>
    </>
  );
}

  return (
    <>
      <h2>What the server sends</h2>
      <pre style={{ whiteSpace: "pre-wrap" }}>
        {renderToString(<Content />)}
      </pre>
      <h2>After hydration</h2>
      <Content />
    </>
  );
}
```

This was verified with a Playwright test against a real Next.js App Router page using a JavaScript-disabled browser context: `expect(focusable).toHaveAttribute("tabindex", "0")` fails (`Received: null`), and pressing Tab from the preceding button focuses the following button instead of the `Focusable`.

### Expected behavior

The server-rendered HTML should match the post-hydration client output: `<Focusable>` (and other div-rendered Focusable-based components such as a virtual-focus `Composite` container or `TooltipAnchor`) should SSR with `tabindex="0"` so keyboard users can reach it before hydration, and `<Focusable disabled>` should SSR with `aria-disabled="true"` only — no invalid `disabled` attribute on elements that don't support it.

### Workaround

Pass the fallback `tabIndex` explicitly so it lands in the server HTML, and use [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) on disabled non-native Focusables to avoid the stray `disabled` attribute:

```tsx
<Ariakit.Focusable tabIndex={0}>Focusable card</Ariakit.Focusable>
<Ariakit.Focusable disabled accessibleWhenDisabled>
  Disabled focusable card
</Ariakit.Focusable>
```

Caveat: `accessibleWhenDisabled` is a documented option but changes behavior — the disabled element remains focusable (with `aria-disabled`) after hydration. There is no consumer-side way to keep the exact default disabled semantics and avoid the server-rendered `disabled` attribute.

### Possible solutions

Give `useFocusable` a deterministic tag name for SSR and the first client render instead of `undefined`: prefer the intrinsic type of the `render` prop when it is a string (e.g. `render={<button />}`), then a default tag threaded down by derived components via a private option (precedent: `unstable_treeSnapshotKey` in dialog.tsx), then `Focusable`'s own default `"div"` — i.e. `useTagName(ref, fallbackTagName)` at [focusable.tsx#L410](https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-react-components/src/focusable/focusable.tsx#L410). Since server and client first render compute identical props, this cannot introduce hydration mismatches, and the real DOM tag still wins once `useTagName`'s layout effect runs.

Derived components whose default tag differs from what would otherwise arrive should thread their own default: `Command` passes `"button"` (covering `Button`, `Checkbox`, `Radio`, `CompositeItem`, etc.), `HovercardAnchor` `"a"`, `FormInput` `"input"`, `Combobox` `"input"` (critical so a disabled server-rendered `Combobox` keeps its native `disabled`), `TooltipAnchor` `"div"`, and the div-based items (`MenuItem`, `SelectItem`, `ComboboxItem`, `Tag`) `"div"`.

Simply flipping the `if (!tagName) return true` defaults in `isNativeTabbable`/`supportsDisabledAttribute` would fix the `tabindex` symptom but invert the regression: `<Button disabled>`/`<Checkbox disabled>` would then SSR without their native `disabled` attribute, leaving genuinely native controls clickable pre-hydration.

diegohaz:
## Motivation

Fixes #6341

When a [`Focusable`](https://ariakit.com/reference/focusable)-based component renders a non-native tag (the default for `Focusable` is a `div`), the server-rendered HTML is missing the `tabIndex={0}` fallback the component relies on to be focusable. Until hydration completes — or indefinitely for users browsing with JavaScript disabled — the element is absent from the tab order:

```html
<!-- <Focusable>Focusable card</Focusable> server-renders as: -->
<div>Focusable card</div>

<!-- client steady state after hydration: -->
<div tabindex="0">Focusable card</div>
```

A disabled `Focusable` additionally server-renders a literal `disabled` attribute on the `div`, which is invalid HTML and matches `[disabled]` CSS selectors until hydration silently removes it:

```html
<!-- <Focusable disabled>…</Focusable> server-renders as: -->
<div aria-disabled="true" disabled="" style="pointer-events:none">…</div>
```

`useFocusable` computes `tabIndex` and `disabled` from the element's tag name, read from the DOM via `useTagName(ref)` with no fallback type. The tag name is only learned inside a layout effect, which never runs during server rendering, so both predicates default to "assume a native control": no `tabIndex` fallback is emitted and the stray `disabled` attribute is.

Every div-rendered descendant of `useFocusable` inherits this: a virtual-focus [`Composite`](https://ariakit.com/reference/composite) container (the widget's only tabbable element) SSRs with no `tabindex`, and [`TooltipAnchor`](https://ariakit.com/reference/tooltip-anchor) SSRs as a bare `<div>`.

## Solution

Give `useFocusable` a deterministic tag name during server rendering and the first client render instead of `undefined`:

1. When the [`render`](https://ariakit.com/guide/composition) prop is an element with an intrinsic (string) type, that tag is used — e.g. `render={<button />}` computes as `button`.
2. Otherwise, a private `unstable_defaultTagName` option threaded by derived components is used: `Command` passes `"button"` (covering `Button`, `Disclosure`, `CompositeItem`, and everything else built on it), `Combobox` and `FormInput` pass `"input"`, `HovercardAnchor` passes `"a"`, `MenuButton` passes `"button"` — or `"div"` when nested in another menu, since submenu buttons render through a `Role.div` element — and the div-based components (`TooltipAnchor`, `ToolbarContainer`, `MenuItem`, `MenuItemCheckbox`, `MenuItemRadio`, `SelectItem`, `ComboboxItem`, `Tag`) pass `"div"`.
3. Otherwise, `Focusable`'s own default `"div"` applies.

Since the server and the client first render compute identical props, this cannot introduce hydration mismatches, and the actual DOM tag still takes over once `useTagName`'s layout effect runs. Components whose arriving guess differs only within the same predicate class (e.g. `Checkbox`'s `input` vs `Command`'s `button`, both native and `disabled`-capable) don't need to thread anything, because the tag name only feeds `isNativeTabbable` and `supportsDisabledAttribute`.

Known limitation: when the `render` prop is a custom component (e.g. `render={<RouterLink />}`), the rendered tag can't be known statically, so the threaded default is used as the guess and the real tag is picked up after mounting — the same transient correction that already existed before this fix. Common compositions like `render={<MenuItem />}` still produce correct server markup because the inner Ariakit component computes its own attributes.

### Composite focus race unmasked by the fix

Previously, `useTagName`'s `undefined → "div"` state correction forced an extra render after mount, which happened to flush the composite base element registration (a layout effect on a ref-triggered second render) before the composite focus-on-move effect focused the active item. With the guess now matching the real tag, that accidental re-render is gone, and in flows like `select-items-unmount` (a popover that mounts on show with a pending move) the selected item could receive DOM focus while `store.baseElement` was still unregistered, so the virtual focus redirect in `CompositeItem` bailed and left DOM focus on the item.

`CompositeItem`'s `onFocus` now handles that case: when the redirect bails only because the base element isn't registered yet, it queues a microtask that redirects focus to the composite element once the registration has been committed, as long as the item still has focus. (Deferring the focus-on-move effect itself was tried and rejected: it broke multi-store items, such as select-with-combobox tabs, where another store's redirect handles the focus synchronously.)

## Tests

The `focusable-6341` sandbox asserts the server-rendered markup both with JavaScript disabled in a real browser (`test-browser.ts`: attributes plus real Tab-key navigation) and via `renderToString` in happy-dom (`test.ts`).

Two tests pass by design at every stage and act as regression guards:

- "native buttons keep their server-rendered semantics" ensures the fix doesn't strip the native `disabled` attribute or add a stray `tabindex` to native controls (which would happen if the undefined-tag defaults in `isNativeTabbable`/`supportsDisabledAttribute` were naively flipped).
- "hydrated markup matches the server-rendered attributes" documents that post-hydration output is the contract the server markup must match.

## Workaround

Until the fix is released, pass the fallback `tabIndex` explicitly so it lands in the server HTML, and use [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) on disabled non-native Focusables to avoid the stray `disabled` attribute:

```tsx
// TODO: Remove the explicit tabIndex and accessibleWhenDisabled props once
// https://github.com/ariakit/ariakit/issues/6341 is fixed.
<Ariakit.Focusable tabIndex={0}>Focusable card</Ariakit.Focusable>
<Ariakit.Focusable disabled accessibleWhenDisabled>
  Disabled focusable card
</Ariakit.Focusable>
```

The same explicit `tabIndex={0}` applies to other div-rendered Focusable-based elements, such as a virtual-focus `Composite` container or a `TooltipAnchor`.

Caveat: `accessibleWhenDisabled` is a documented option but changes behavior — the disabled element remains focusable (with `aria-disabled`) after hydration. There is no consumer-side way to keep the exact default disabled semantics and avoid the server-rendered `disabled` attribute, because `aria-disabled` is treated the same way as the `disabled` prop.

### DIFF
--- a/.changeset/300-focusable-ssr-tabindex.md
+++ b/.changeset/300-focusable-ssr-tabindex.md
@@ -1,0 +1,12 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Server-rendered `Focusable` keyboard accessibility
+
+[`Focusable`](https://ariakit.com/reference/focusable)-based components that render a non-native tag — such as the default `div` of [`Focusable`](https://ariakit.com/reference/focusable) itself, a virtual focus [`Composite`](https://ariakit.com/reference/composite) container, or [`TooltipAnchor`](https://ariakit.com/reference/tooltip-anchor) — now server-render with the `tabindex="0"` fallback, so keyboard users can reach them before hydration completes or when JavaScript is disabled.
+
+Disabled non-native components — including div-based items such as [`MenuItem`](https://ariakit.com/reference/menu-item), [`SelectItem`](https://ariakit.com/reference/select-item), and [`ComboboxItem`](https://ariakit.com/reference/combobox-item) — no longer server-render an invalid `disabled` attribute, which is silently removed on hydration but could match `[disabled]` CSS selectors until then. Native controls such as [`Button`](https://ariakit.com/reference/button) and [`Checkbox`](https://ariakit.com/reference/checkbox) keep their native `disabled` attribute in the server-rendered HTML.
+
+In both cases, the server now renders the same attributes the client computes on its first render, keeping hydration consistent.
--- a/app/src/sandbox/focusable-6341/index.react.tsx
+++ b/app/src/sandbox/focusable-6341/index.react.tsx
@@ -9,14 +9,23 @@ export default function Example() {
   return (
     <>
       <Ariakit.Button>Before</Ariakit.Button>
-      <Ariakit.Focusable>Focusable card</Ariakit.Focusable>
-      <Ariakit.Focusable disabled>Disabled focusable card</Ariakit.Focusable>
+      {/* TODO: Remove the explicit tabIndex and accessibleWhenDisabled props
+          from the elements below once
+          https://github.com/ariakit/ariakit/issues/6341 is fixed. Note that
+          accessibleWhenDisabled changes behavior: the disabled element remains
+          focusable (with aria-disabled) after hydration. */}
+      <Ariakit.Focusable tabIndex={0}>Focusable card</Ariakit.Focusable>
+      <Ariakit.Focusable disabled accessibleWhenDisabled>
+        Disabled focusable card
+      </Ariakit.Focusable>
       <Ariakit.TooltipProvider>
-        <Ariakit.TooltipAnchor>Tooltip anchor</Ariakit.TooltipAnchor>
+        <Ariakit.TooltipAnchor tabIndex={0}>
+          Tooltip anchor
+        </Ariakit.TooltipAnchor>
         <Ariakit.Tooltip>Tooltip content</Ariakit.Tooltip>
       </Ariakit.TooltipProvider>
       <Ariakit.CompositeProvider virtualFocus>
-        <Ariakit.Composite role="toolbar" aria-label="Composite">
+        <Ariakit.Composite role="toolbar" aria-label="Composite" tabIndex={0}>
           <Ariakit.CompositeItem render={<div />}>Item 1</Ariakit.CompositeItem>
           <Ariakit.CompositeItem render={<div />}>Item 2</Ariakit.CompositeItem>
         </Ariakit.Composite>
--- a/app/src/sandbox/focusable-6341/index.react.tsx
+++ b/app/src/sandbox/focusable-6341/index.react.tsx
@@ -1,0 +1,28 @@
+import * as Ariakit from "@ariakit/react";
+
+// See https://github.com/ariakit/ariakit/issues/6341
+// Div-based Focusable elements must be part of the server-rendered HTML's tab
+// order (tabindex="0") so keyboard users can reach them before hydration, and
+// a disabled non-native Focusable must not server-render an invalid `disabled`
+// attribute.
+export default function Example() {
+  return (
+    <>
+      <Ariakit.Button>Before</Ariakit.Button>
+      <Ariakit.Focusable>Focusable card</Ariakit.Focusable>
+      <Ariakit.Focusable disabled>Disabled focusable card</Ariakit.Focusable>
+      <Ariakit.TooltipProvider>
+        <Ariakit.TooltipAnchor>Tooltip anchor</Ariakit.TooltipAnchor>
+        <Ariakit.Tooltip>Tooltip content</Ariakit.Tooltip>
+      </Ariakit.TooltipProvider>
+      <Ariakit.CompositeProvider virtualFocus>
+        <Ariakit.Composite role="toolbar" aria-label="Composite">
+          <Ariakit.CompositeItem render={<div />}>Item 1</Ariakit.CompositeItem>
+          <Ariakit.CompositeItem render={<div />}>Item 2</Ariakit.CompositeItem>
+        </Ariakit.Composite>
+      </Ariakit.CompositeProvider>
+      <Ariakit.Button disabled>Disabled button</Ariakit.Button>
+      <Ariakit.Button>After</Ariakit.Button>
+    </>
+  );
+}
--- a/app/src/sandbox/focusable-6341/index.react.tsx
+++ b/app/src/sandbox/focusable-6341/index.react.tsx
@@ -9,27 +9,28 @@ export default function Example() {
   return (
     <>
       <Ariakit.Button>Before</Ariakit.Button>
-      {/* TODO: Remove the explicit tabIndex and accessibleWhenDisabled props
-          from the elements below once
-          https://github.com/ariakit/ariakit/issues/6341 is fixed. Note that
-          accessibleWhenDisabled changes behavior: the disabled element remains
-          focusable (with aria-disabled) after hydration. */}
-      <Ariakit.Focusable tabIndex={0}>Focusable card</Ariakit.Focusable>
-      <Ariakit.Focusable disabled accessibleWhenDisabled>
-        Disabled focusable card
-      </Ariakit.Focusable>
+      <Ariakit.Focusable>Focusable card</Ariakit.Focusable>
+      <Ariakit.Focusable disabled>Disabled focusable card</Ariakit.Focusable>
       <Ariakit.TooltipProvider>
-        <Ariakit.TooltipAnchor tabIndex={0}>
-          Tooltip anchor
-        </Ariakit.TooltipAnchor>
+        <Ariakit.TooltipAnchor>Tooltip anchor</Ariakit.TooltipAnchor>
         <Ariakit.Tooltip>Tooltip content</Ariakit.Tooltip>
       </Ariakit.TooltipProvider>
       <Ariakit.CompositeProvider virtualFocus>
-        <Ariakit.Composite role="toolbar" aria-label="Composite" tabIndex={0}>
+        <Ariakit.Composite role="toolbar" aria-label="Composite">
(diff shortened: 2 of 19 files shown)


diegohaz:
Closing this after some deliberation. The main motivation — being keyboard-reachable before JS loads — doesn't hold up: a `tabindex="0"` div without JavaScript is focusable but inert, so users can Tab to it and nothing responds, which isn't better than being skipped. If an element needs to work before hydration, it has to render a native focusable element anyway (`render={<button />}`, a link with a real `href`, etc.), and those already server-render with the correct attributes.

Against that, the fix requires threading `unstable_defaultTagName` through every `Focusable`-derived component — an invariant that fails silently when missed, in the worse direction (a wrong guess strips `disabled` from a native button in the server output) — plus a timing-sensitive focus redirect in `CompositeItem`. It also trades correctness between composition styles: `<Focusable render={<MyButton />} disabled>` (a custom component resolving to a native button) server-renders correctly today, but would produce wrong markup with the guess until hydration.

For the cases where the server markup matters, there's a simple explicit opt-in — see https://github.com/ariakit/ariakit/issues/6341#issuecomment-4865877722.

PR #6603, opened 2026-07-02
Fix Tab setSelectedId not updating the active tab in Select/Combobox popovers

### LINKED ISSUE #6346: Tabs in a Select/Combobox popover: the first setSelectedId after the popover opens or toggles doesn't update the active tab
### Current behavior

When tabs are composed with a select or combobox — the documented composition used by the [Select with Combobox and Tabs](https://ariakit.com/examples/select-combobox-tab) and [Combobox with Tabs](https://ariakit.com/examples/combobox-tabs) examples — calling `setSelectedId` (for example from a cross-link button inside a tab panel, or any programmatic/controlled `selectedId` change) switches the selected tab and panel, but the documented `activeId` sync is silently skipped: the newly selected tab does not become the active composite item (no `data-active-item`, roving `tabindex` stays `-1`), and the focus move promised by the [`setSelectedId`](https://ariakit.com/reference/use-tab-store) docs ("If another tab has DOM focus and the selected tab is enabled, focus will move to the selected tab") never happens.

This is deterministic, not a race: it affects the **first** `setSelectedId` call after the tab store initializes and after **every** popover toggle in which `selectedId` didn't change. A second identical call works. Click and arrow-key tab switching are unaffected (those paths write `activeId` through DOM focus before `selectedId` changes), which is why the shipped examples don't catch it.

The root cause is a leaked one-shot suppression flag in the tab store. The `selectedId` → `activeId` sync is a batch listener guarded by `syncActiveId`:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/tab/tab-store.ts#L148-L177

The flag is armed unconditionally right before restoring `selectedId` when the select/combobox popover toggles `mounted`:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-components/src/tab/tab-store.ts#L217-L243

But `setState` early-returns when the value is unchanged, producing no batch event to consume the flag:

https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L385-L391

And since `sync` invokes its listener immediately on registration (https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L301-L305), `restoreSelectedId` runs once at every store init right after `backupSelectedId` — a guaranteed no-op restore that arms the flag and leaks it. The same leak recurs on each `mounted` toggle where `selectedId` is already the backed-up value. The next legitimate `selectedId` change is then swallowed by the suppression branch. A second, related failure mode: batch flushes are microtask-coalesced (https://github.com/ariakit/ariakit/blob/11bd4458de6447b996dd882ea0c85eb801a6cd03/packages/ariakit-store/src/index.ts#L451-L455), so a real restore that shares a flush with a later legitimate `setSelectedId` suppresses the legitimate change too.

1. Go to https://stackblitz.com/github/ariakit/ariakit/tree/main/templates/react?file=src%2Fapp.tsx
3. Click the **Grocery** select button to open the popover. The **Fruits** tab is selected.
4. Click the **Browse vegetables** button inside the Fruits panel.
5. The Vegetables tab gets selected (`aria-selected="true"`, the panel switches), but it never becomes the active item: no red `data-active-item` outline appears on it and its `tabindex` stays `-1`. Clicking **Browse fruits** next works correctly (the Fruits tab gets the active outline), showing the one-shot nature of the bug. The same failure repeats on the first tab switch after closing and reopening the popover.

```tsx

function BrowseTabButton(props: { tabId: string; children: string }) {
  const tab = Ariakit.useTabContext();
  return (
    <button onClick={() => tab?.setSelectedId(props.tabId)}>
      {props.children}
    </button>
  );
}

  return (
    <>
      <style>{`[data-active-item] { outline: 2px solid red; }`}</style>
      <Ariakit.SelectProvider defaultValue="Apple" virtualFocus={false}>
        <Ariakit.SelectLabel>Grocery</Ariakit.SelectLabel>
        <Ariakit.Select />
        <Ariakit.SelectPopover
          gutter={4}
          sameWidth
          style={{ background: "white", border: "1px solid gray", padding: 8 }}
        >
          <Ariakit.TabProvider defaultSelectedId="tab-fruits">
            <Ariakit.TabList aria-label="Categories">
              <Ariakit.Tab id="tab-fruits">Fruits</Ariakit.Tab>
              <Ariakit.Tab id="tab-vegetables">Vegetables</Ariakit.Tab>
            </Ariakit.TabList>
            <Ariakit.TabPanel tabId="tab-fruits">
              <Ariakit.SelectList>
                <Ariakit.SelectItem value="Apple" />
                <Ariakit.SelectItem value="Banana" />
              </Ariakit.SelectList>
              <BrowseTabButton tabId="tab-vegetables">
                Browse vegetables
              </BrowseTabButton>
            </Ariakit.TabPanel>
            <Ariakit.TabPanel tabId="tab-vegetables">
              <Ariakit.SelectList>
                <Ariakit.SelectItem value="Carrot" />
                <Ariakit.SelectItem value="Potato" />
              </Ariakit.SelectList>
              <BrowseTabButton tabId="tab-fruits">
                Browse fruits
              </BrowseTabButton>
            </Ariakit.TabPanel>
          </Ariakit.TabProvider>
        </Ariakit.SelectPopover>
      </Ariakit.SelectProvider>
    </>
  );
}
```

This was verified with a real-browser Playwright test using only user-level interactions (clicks), asserting `data-active-item`/`tabindex` on the tab elements; the identical cross-link button on a plain `TabProvider` passes as a control.

### Expected behavior

Per the `setSelectedId` documentation, setting the `selectedId` state should keep the `activeId` state in sync (and move focus when another tab has DOM focus). After clicking **Browse vegetables**, the Vegetables tab should become the active item — exactly what happens with a plain `TabProvider`, and what happens on the *second* `setSelectedId` call in the select composition.

### Workaround

Call `setActiveId` alongside `setSelectedId` to perform the sync manually:

```tsx
function BrowseTabButton(props: { tabId: string; children: string }) {
  const tab = Ariakit.useTabContext();
  return (
    <button
      onClick={() => {
        tab?.setSelectedId(props.tabId);
        // Workaround for the swallowed setSelectedId -> activeId sync.
        tab?.setActiveId(props.tabId);
      }}
    >
      {props.children}
    </button>
  );
}
```

(Or use `tab.select(id)` if always moving focus to the tab is acceptable.)

### Possible solutions

Arm the suppression only when the restore will actually produce a batch event, and have the listener verify that the flushed value is the restored one, so a coalesced flush that also contains a later legitimate change still syncs. Resetting the flag at setup time also prevents a restore armed right before store destruction from leaking into the next init:

```diff
-  let syncActiveId = true;
+  let pendingRestore = false;
+  let restoredSelectedId: TabStoreState["selectedId"];

   // Keep activeId in sync with selectedId.
   setup(tab, () =>
     batch(tab, ["selectedId"], (state, prev) => {
-      if (!syncActiveId) {
-        syncActiveId = true;
-        return;
-      }
+      // Skip the sync only when this flush corresponds to restoring the
+      // selectedId state from a select or combobox selected value. Batch
+      // listeners are microtask-coalesced, so a restore can share a flush
+      // with a later legitimate change; in that case we must still sync.
+      if (pendingRestore) {
+        pendingRestore = false;
+        if (state.selectedId === restoredSelectedId) return;
+      }
@@
   setup(tab, () => {
+    pendingRestore = false;
     const backupSelectedId = () => {
       selectedIdFromSelectedValue = tab.getState().selectedId;
     };
     const restoreSelectedId = () => {
-      syncActiveId = false;
+      const { selectedId } = tab.getState();
+      // setState early-returns on unchanged values and produces no batch
+      // event, so only arm the suppression when the restore will actually
+      // change the state.
+      if (selectedId === selectedIdFromSelectedValue) return;
+      pendingRestore = true;
+      restoredSelectedId = selectedIdFromSelectedValue;
       tab.setState("selectedId", selectedIdFromSelectedValue);
     };
```

One side effect to verify: today the init-time leak also swallows the very first `selectedId` batch event (typically the automatic `undefined` → first enabled tab assignment). Un-masking that flush means an `activeId` write at init can propagate to the parent select/combobox store (`activeId` is a shared key), so the first open of a combobox-with-tabs composition with uncontrolled `selectedId` should be checked; if that write turns out to be undesirable there, the sync should be guarded explicitly (e.g. while the parent popover is closed) rather than relying on the accidental swallow.

diegohaz:
## Motivation

In the documented Select/Combobox + Tabs composition (see the [Select with Combobox and Tabs](https://ariakit.com/examples/select-combobox-tab) and [Combobox with Tabs](https://ariakit.com/examples/combobox-tabs) examples), the first `setSelectedId` call after the tab store initializes — and after every popover toggle in which `selectedId` didn't change — switches the selected tab and panel but silently skips the documented `activeId` sync: the newly selected tab never becomes the active composite item (no `data-active-item`, roving `tabindex` stays `-1`). A second identical call works, which is why click/arrow-key tab switching doesn't surface the bug.

```tsx
<Ariakit.SelectPopover>
  <Ariakit.TabProvider defaultSelectedId="tab-fruits">
    {/* tab.setSelectedId("tab-vegetables") selects the tab,
        but it never becomes the active item */}
  </Ariakit.TabProvider>
</Ariakit.SelectPopover>
```

The root cause is a leaked one-shot suppression flag in the tab store: `restoreSelectedId` armed `syncActiveId = false` unconditionally before restoring `selectedId` when the select/combobox popover toggles `mounted`, but `setState` early-returns when the value is unchanged, producing no batch event to consume the flag. Since `sync` invokes its listener immediately on registration, a guaranteed no-op restore ran at every store init, arming and leaking the flag. The next legitimate `selectedId` change was then swallowed by the suppression branch. A second failure mode: batch flushes are microtask-coalesced, so a real restore sharing a flush with a later legitimate change suppressed the legitimate change too.

Fixes #6346

## Solution

Replace the one-shot boolean with a `pendingRestore` flag paired with the restored value:

- `restoreSelectedId` arms the suppression only when the restore will actually change the state, since an unchanged `setState` emits no batch event to consume it. This removes the init-time and popover-toggle leaks.
- The batch listener, after consuming the flag, skips the `activeId` sync only when the flushed `selectedId` is the restored one, so a coalesced flush that also contains a later legitimate change still syncs.
- The flag resets at init, before the batch listener registers, so a restore armed right before the store is destroyed can't leak into the next init.

The sandbox workaround from the previous commit is reverted; the repro tests now pass against the library fix. Five store-level unit tests in `tab-store.test.ts` cover the restore paths directly (init leak, toggle leak, restore suppression, coalesced restore + change, and armed-restore across store destroy/re-init); all five fail against the previous implementation.

Verified with the full core/React/React 18/Solid suites, the repro sandbox in Chrome/Firefox/Safari, the neighboring tab sandboxes, and the select/combobox + tabs example browser tests.

### Known limitation

If a store with a just-armed restore is destroyed and re-initialized within the same microtask, the restore's still-queued batch flush lands in the new epoch after the flag reset and runs the `activeId` sync. Inside the batch listener, that cross-epoch flush is indistinguishable from the listener's registration-time invocation (both deliver an identical previous-state snapshot after the idle-registration refresh), so suppressing it would require event-origin tagging in `@ariakit/store`. The scenario is not reachable through React's lifecycle in practice, and the consequence is a transient `activeId` write while the popover is closed, which the parent select/combobox rewrites on open.

## Workaround

Until the fix is released, users can perform the swallowed `selectedId` → `activeId` sync manually by calling `setActiveId` alongside `setSelectedId`:

```tsx
function BrowseTabButton(props: { tabId: string; children: string }) {
  const tab = Ariakit.useTabContext();
  return (
    <button
      onClick={() => {
        tab?.setSelectedId(props.tabId);
        // TODO: Workaround for the swallowed setSelectedId -> activeId sync.
        // Remove once https://github.com/ariakit/ariakit/issues/6346 is fixed.
        tab?.setActiveId(props.tabId);
      }}
    >
      {props.children}
    </button>
  );
}
```

Alternatively, `tab.select(id)` also works when always moving DOM focus to the tab is acceptable.

### DIFF
--- a/.changeset/300-tab-select-active-id-sync.md
+++ b/.changeset/300-tab-select-active-id-sync.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Tab`](https://ariakit.com/reference/tab) not becoming the active item on the first [`setSelectedId`](https://ariakit.com/reference/use-tab-store#setselectedid-1) call after a [`SelectPopover`](https://ariakit.com/reference/select-popover) or [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) containing the tabs opens or toggles.
--- a/app/src/sandbox/tab-select-6346/index.react.tsx
+++ b/app/src/sandbox/tab-select-6346/index.react.tsx
@@ -3,14 +3,7 @@ import * as Ariakit from "@ariakit/react";
 function BrowseTabButton(props: { tabId: string; children: string }) {
   const tab = Ariakit.useTabContext();
   return (
-    <button
-      onClick={() => {
-        tab?.setSelectedId(props.tabId);
-        // TODO: Workaround for the swallowed setSelectedId -> activeId sync.
-        // Remove once https://github.com/ariakit/ariakit/issues/6346 is fixed.
-        tab?.setActiveId(props.tabId);
-      }}
-    >
+    <button onClick={() => tab?.setSelectedId(props.tabId)}>
       {props.children}
     </button>
   );
--- a/app/src/sandbox/tab-select-6346/index.react.tsx
+++ b/app/src/sandbox/tab-select-6346/index.react.tsx
@@ -3,7 +3,14 @@ import * as Ariakit from "@ariakit/react";
 function BrowseTabButton(props: { tabId: string; children: string }) {
   const tab = Ariakit.useTabContext();
   return (
-    <button onClick={() => tab?.setSelectedId(props.tabId)}>
+    <button
+      onClick={() => {
+        tab?.setSelectedId(props.tabId);
+        // TODO: Workaround for the swallowed setSelectedId -> activeId sync.
+        // Remove once https://github.com/ariakit/ariakit/issues/6346 is fixed.
+        tab?.setActiveId(props.tabId);
+      }}
+    >
       {props.children}
     </button>
   );
--- a/app/src/sandbox/tab-select-6346/index.react.tsx
+++ b/app/src/sandbox/tab-select-6346/index.react.tsx
@@ -1,0 +1,52 @@
+import * as Ariakit from "@ariakit/react";
+
+function BrowseTabButton(props: { tabId: string; children: string }) {
+  const tab = Ariakit.useTabContext();
+  return (
+    <button onClick={() => tab?.setSelectedId(props.tabId)}>
+      {props.children}
+    </button>
+  );
+}
+
+export default function Example() {
+  return (
+    <>
+      <style>{`[data-active-item] { outline: 2px solid red; }`}</style>
+      <Ariakit.SelectProvider defaultValue="Apple" virtualFocus={false}>
+        <Ariakit.SelectLabel>Grocery</Ariakit.SelectLabel>
+        <Ariakit.Select />
+        <Ariakit.SelectPopover
+          gutter={4}
+          sameWidth
+          style={{ background: "white", border: "1px solid gray", padding: 8 }}
+        >
+          <Ariakit.TabProvider defaultSelectedId="tab-fruits">
+            <Ariakit.TabList aria-label="Categories">
+              <Ariakit.Tab id="tab-fruits">Fruits</Ariakit.Tab>
+              <Ariakit.Tab id="tab-vegetables">Vegetables</Ariakit.Tab>
+            </Ariakit.TabList>
+            <Ariakit.TabPanel tabId="tab-fruits">
+              <Ariakit.SelectList>
+                <Ariakit.SelectItem value="Apple" />
+                <Ariakit.SelectItem value="Banana" />
+              </Ariakit.SelectList>
+              <BrowseTabButton tabId="tab-vegetables">
+                Browse vegetables
+              </BrowseTabButton>
+            </Ariakit.TabPanel>
+            <Ariakit.TabPanel tabId="tab-vegetables">
+              <Ariakit.SelectList>
+                <Ariakit.SelectItem value="Carrot" />
+                <Ariakit.SelectItem value="Potato" />
+              </Ariakit.SelectList>
+              <BrowseTabButton tabId="tab-fruits">
+                Browse fruits
+              </BrowseTabButton>
+            </Ariakit.TabPanel>
+          </Ariakit.TabProvider>
+        </Ariakit.SelectPopover>
+      </Ariakit.SelectProvider>
+    </>
+  );
+}
--- a/app/src/sandbox/tab-select-6346/test-browser.ts
+++ b/app/src/sandbox/tab-select-6346/test-browser.ts
@@ -14,13 +14,15 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .expect(q.tab("Vegetables"))
       .toHaveAttribute("aria-selected", "true");
     await test.expect(q.tab("Vegetables")).toHaveAttribute("data-active-item");
-    await test.expect(q.tab("Vegetables")).toHaveAttribute("tabindex", "0");
+    await test
+      .expect(q.tab("Vegetables"))
+      .not.toHaveAttribute("tabindex", "-1");
 
     await q.button("Browse fruits").click();
 
     await test.expect(q.tab("Fruits")).toHaveAttribute("aria-selected", "true");
     await test.expect(q.tab("Fruits")).toHaveAttribute("data-active-item");
-    await test.expect(q.tab("Fruits")).toHaveAttribute("tabindex", "0");
+    await test.expect(q.tab("Fruits")).not.toHaveAttribute("tabindex", "-1");
   });
 
   test("activates the tab selected by setSelectedId after the popover toggles", async ({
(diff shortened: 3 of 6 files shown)


diegohaz (commented):
Initial review of `77eefae9f604f7ce46e5a5482681c858da695796` against merge-base `61dbfcd19ac9ed48ba0f05a5e94dfb6f4e27a888`.

Scope reviewed:
- Read the PR description, issue #6346, the changed tab store, store `sync`/`batch`/`setup` internals, select/combobox parent stores, the React tab-store wrapper, the changeset, the new sandbox, and adjacent Select/Combobox + Tabs examples.
- Checked review history through GraphQL; there were no prior marked reviews and no existing review threads.
- Checked `origin/main`; it has advanced to `ad0526bcaee6005ec639ba7825e5983e98b7f806`, and `git merge-tree --write-tree origin/main HEAD` completed cleanly.

Findings: no substantive findings to publish.

Validation run:
- `pnpm test packages/ariakit-components/src/tab/tab-store.test.ts app/src/sandbox/tab-select-6346/test.ts` (collected the store test file; 9 passed)
- `pnpm test-react app/src/sandbox/tab-select-6346/test.ts` (2 passed)
- `APP_PORT=4322 NEXTJS_PORT=3001 pnpm -F app run test src/sandbox/tab-select-6346/test-browser.ts --project=chrome --project=firefox --project=safari` (6 passed)
- `pnpm test-react examples/combobox-tabs/test.ts examples/select-combobox-tab/test.ts` (40 passed)
- Store-level Node probe with a `requestAnimationFrame` shim for uncontrolled tabs under both select and combobox parents; the related adjacent React example assertions still passed.

Not fully validated: I did not run full `pnpm tsc`, `pnpm lint`, or the full browser suite. I also attempted the old `examples/*` browser tests with the app dev server, but that invocation reached a Next.js 404 before rendering the examples, so I did not count it as PR validation.

PR #6606, opened 2026-07-02
Remove unneeded ensure queries from sandbox tests

diegohaz:
## Motivation

Sandbox tests under `app/src/sandbox/**/test.ts` were using the throwing `.ensure(...)` query variant pervasively, even where the result is only passed to APIs that already handle `null` with a clear failure:

- The `@ariakit/test` interaction utilities (`click`, `hover`, `focus`, `rightClick`) throw an explicit invariant on a null element.
- jest-dom matchers (`toHaveAttribute`, `toBeVisible`, `toHaveFocus`, …) throw on a null received value, including under `.not`.
- `q.within(...)` throws its own invariant on null.

The established convention in `examples/**/test.ts` reserves `.ensure` for the cases where the element is dereferenced directly, so the pervasive usage in sandboxes was inconsistent noise:

```ts
// Before
await click(q.button.ensure("Open dialog"));
expect(q.tab.ensure("Fruits")).toHaveAttribute("aria-selected", "true");

// After — same failure behavior when the element is missing
await click(q.button("Open dialog"));
expect(q.tab("Fruits")).toHaveAttribute("aria-selected", "true");
```

## Solution

Remove `.ensure` wherever the query result only feeds a null-safe sink (interaction utilities, `expect` matchers, `q.within`), and keep it in two cases:

1. The element is dereferenced — property or method access such as `.textContent`, `.querySelector()`, `.focus()`, `.style`, or an argument to a local helper typed with a non-null `HTMLElement` parameter:

   ```ts
   // .ensure stays: the element is dereferenced directly
   const backdrop = q.presentation.ensure();
   await expect.poll(() => backdrop.getAttribute("data-enter")).toBe("true");
   ```

2. The element is an explicit `type()` or `press()` target (from review feedback): unlike the other interaction utilities, `type`/`press` treat a null element as "use the active element" and silently no-op when it isn't focusable, so a missing target would not fail at the call site:

   ```ts
   // .ensure stays: type() falls back to the active element on null
   await type("ana", q.combobox.ensure("Your favorite fruit"));
   ```

This removes 146 redundant call sites across 40 files. The diff is a pure `.ensure` token removal (every removed line equals its added line after deleting `.ensure`), plus an oxfmt rewrap of now-shorter lines in `app/src/sandbox/tab-select-6346/test.ts`.

Verified with `pnpm test-react app/src/sandbox` (48 files / 97 tests), `pnpm test-solid app/src/sandbox` (3 files / 4 tests), `pnpm tsc`, and `pnpm lint`.

### DIFF
--- a/app/src/sandbox/combobox-item-value-6298/test.ts
+++ b/app/src/sandbox/combobox-item-value-6298/test.ts
@@ -3,7 +3,7 @@ import { expect, test } from "vitest";
 
 // Reproduces https://github.com/ariakit/ariakit/issues/6298
 test("renders overlapping matches without duplicated text", async () => {
-  await type("ana", q.combobox.ensure("Your favorite fruit"));
+  await type("ana", q.combobox("Your favorite fruit"));
 
   const option = q.option.ensure("Banana");
   expect(option).toHaveTextContent("Banana");
--- a/app/src/sandbox/combobox-item-value-6298/test.ts
+++ b/app/src/sandbox/combobox-item-value-6298/test.ts
@@ -3,7 +3,7 @@ import { expect, test } from "vitest";
 
 // Reproduces https://github.com/ariakit/ariakit/issues/6298
 test("renders overlapping matches without duplicated text", async () => {
-  await type("ana", q.combobox("Your favorite fruit"));
+  await type("ana", q.combobox.ensure("Your favorite fruit"));
 
   const option = q.option.ensure("Banana");
   expect(option).toHaveTextContent("Banana");
--- a/app/src/sandbox/combobox-6297/test.ts
+++ b/app/src/sandbox/combobox-6297/test.ts
@@ -13,32 +13,32 @@ function isApplePlatform(platform: string) {
 // See https://github.com/ariakit/ariakit/issues/6297
 for (const shortcut of shortcuts) {
   test(`${shortcut.name} on a focused item does not commit inline autocomplete`, async () => {
-    const combobox = q.combobox.ensure("Fruit");
+    const combobox = q.combobox("Fruit");
 
     await click(combobox);
     await type("b");
-    expect(q.status.ensure()).toHaveTextContent("b");
+    expect(q.status()).toHaveTextContent("b");
 
     await press.ArrowDown();
-    const apple = q.option.ensure("Apple");
+    const apple = q.option("Apple");
     expect(apple).toHaveFocus();
     expect(combobox).toHaveValue("Apple");
 
     await press("c", null, shortcut.options);
     expect(apple).toHaveFocus();
-    expect(q.status.ensure()).toHaveTextContent("b");
+    expect(q.status()).toHaveTextContent("b");
   });
 }
 
 test("paste shortcut on a focused item moves focus to the input", async () => {
-  const combobox = q.combobox.ensure("Fruit");
+  const combobox = q.combobox("Fruit");
   const applePlatform = isApplePlatform(navigator.platform);
   const options = applePlatform ? { metaKey: true } : { ctrlKey: true };
 
   await click(combobox);
   await type("b");
   await press.ArrowDown();
-  const apple = q.option.ensure("Apple");
+  const apple = q.option("Apple");
   expect(apple).toHaveFocus();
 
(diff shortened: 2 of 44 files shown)


app/src/sandbox/combobox-item-value-6298/test.ts
```diff
@@ -3,7 +3,7 @@ import { expect, test } from "vitest";
 
 // Reproduces https://github.com/ariakit/ariakit/issues/6298
 test("renders overlapping matches without duplicated text", async () => {
-  await type("ana", q.combobox.ensure("Your favorite fruit"));
+  await type("ana", q.combobox("Your favorite fruit"));
 
   const option = q.option.ensure("Banana");
   expect(option).toHaveTextContent("Banana");
```
diegohaz, line 9: Could it be `expect(q.option("Banana")?.textContent).toHaveTextContent("Banana")`?

If this makes sense, also update other cases.

app/src/sandbox/combobox-item-value-6298/test.ts
```diff
@@ -3,7 +3,7 @@ import { expect, test } from "vitest";
 
 // Reproduces https://github.com/ariakit/ariakit/issues/6298
 test("renders overlapping matches without duplicated text", async () => {
-  await type("ana", q.combobox.ensure("Your favorite fruit"));
+  await type("ana", q.combobox("Your favorite fruit"));
```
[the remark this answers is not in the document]
diegohaz: Fixed in 8c6f7dbe6 - restored `.ensure` on all explicit `type()` targets, including the `email` binding in `form-6307/test.ts` that was passed to `type()` later in the test. You're right that `type` (and `press`) differ from the other interaction utilities: they treat a null element as "use the active element" instead of throwing, so the target query must require the element.

app/src/sandbox/combobox-item-value-6298/test.ts
diegohaz: I looked into it and kept `.ensure` here. Three reasons:

1. The literal form throws at runtime: `toHaveTextContent` is a jest-dom matcher that requires a DOM node as the received value, so passing `?.textContent` (a string/undefined) always fails with `NodeTypeError`, even when the option exists.
2. The workable variant (`expect(q.option("Banana")?.textContent).toBe("Banana")`) works for this line, but generalizing `?.` to the other dereference cases creates vacuous passes. The clearest example is `hovercard-nested-listener/test.ts`, where `installCount()` returns `.textContent` and the test compares two readings: with `?.`, a missing status element yields `undefined === undefined` and the test passes without testing anything, whereas `.ensure` throws. Similarly, a null-tolerant `getPartTexts` returning `[]` would vacuously satisfy the `toEqual([])` assertions in `combobox-item-value-6329`.
3. `.ensure` also gives the pointed "Unable to find an accessible element with the role" failure at the query itself instead of a downstream `expected undefined to be ...`.

So the rule this PR lands on: plain queries when the result only feeds `click`/`hover`/`focus`/`rightClick`, `expect(...)` matchers, or `q.within(...)` (they all fail clearly on null), and `.ensure` when the element is dereferenced or passed as a `type()`/`press()` target (see the other thread). This also matches how `examples/**/test.ts` uses `.ensure` today. Happy to switch to the `?.` form if you prefer it regardless.

diegohaz:
Addressed all unresolved review comments.

**Fixed**
- 8c6f7dbe6 ([reply](https://github.com/ariakit/ariakit/pull/6606#discussion_r3513734160)): restored `.ensure` on all explicit `type()` targets (10 inline sites plus the `email` binding in `form-6307`), since `type`/`press` fall back to the active element on a null target instead of throwing like the other interaction utilities.

**Not adopted**
- ([reply](https://github.com/ariakit/ariakit/pull/6606#discussion_r3513735819)): the `expect(q.option("Banana")?.textContent).toHaveTextContent(...)` form — the literal form throws (`toHaveTextContent` requires a DOM node), and generalizing `?.` to the other dereference sites can produce vacuous passes (e.g. `hovercard-nested-listener`'s install-count comparison would pass on `undefined === undefined`). Kept `.ensure` for dereference cases; details in the thread reply.

PR #6622, opened 2026-07-03
Reduce renders on mount for focusable elements

diegohaz:
## Motivation

`Focusable` detects the rendered element's tag through a post-mount state update. It stores the raw tag name in state, which always changes from its initial `undefined`, so every focusable element pays an extra render on mount even when the detection changes nothing — and so does everything built on `Focusable`. Components that render non-native tags by default, such as `MenuItem`, `ComboboxItem`, or `Dialog`, can never match the optimistic initial value, so they pay the extra render on every mount.

## Solution

Track only the tag traits that affect the rendered props — whether the element is natively tabbable and whether it supports the `disabled` attribute — instead of the raw tag name (extracted from #6616). The trait states are seeded from a new private `unstable_defaultTagName` option, and each component with a non-button default tag declares its own default as the first statement of its hook:

```tsx
// menu-item.tsx (div by default)
props = { unstable_defaultTagName: TagName, ...props };
```

The post-mount detection bails out whenever the rendered element matches the component's default tag, so `div`-based components mount with no extra `Focusable` renders at all (`Command`'s `isNativeButton` state already seeds `false` and bails for them). When hooks compose across different default tags, the outermost hook wins and user-passed props win over all: `Combobox` (`input`) overrides the `div` hint it would otherwise inherit from `useComposite`, and `MenuButton` (`button`) overrides `HovercardAnchor`'s `a`.

First paint and server-rendered HTML now match the settled state for default-tag renders — for example, disabled `div`-based items no longer briefly render an unsupported `disabled` attribute. Rendering a different element via the `render` prop is still detected and corrected after mounting; the one-time detection render just moves to that custom-tag case, matching the direction from #6336 discussions.

## Notes

- `SelectItem` initially had to stay un-hinted: its hint exposed a latent virtual-focus race where the auto-focused selected item could keep DOM focus when the popover opened (filed as #6623, root-caused to the composite focus redirect dropping when the item is focused before the store's `baseElement` is set). That race is now fixed on main by #6627 (with a dedicated `select-6623` regression sandbox), so `SelectItem` passes the hint like the other `div`-based items.
- The dialog family (`Dialog` and everything built on it, such as `Popover`, `Menu`, `Hovercard`, and `Tooltip`) intentionally keeps the tag detection render: with production React, its layout state update defers the dialog's open passive effects until after the first paint, keeping expensive work such as the `--dialog-viewport-height` write and the inert tree-disabling off the opening interaction's critical path. Hinting `Dialog` regressed dialog-perf "open dialog" scripting/INP on CI; root cause and production A/B numbers are in [this comment](https://github.com/ariakit/ariakit/pull/6622#issuecomment-4874958607).
- Because this PR removes the extra corrective renders elsewhere, the dialog-perf "open dialog" benchmark can still show the open effects flushing before first paint on CI compared to main. Explicitly deferring that work is #6609's job; this PR briefly carried that work and it has been reverted back out to keep the scope to the `Focusable` changes.
- `TabPanel` is also not hinted: its focusability depends on `hasTabbableChildren`, which is only known after mounting, so a div hint would emit a transient `tabindex="0"` on initial/SSR renders of panels with tabbable children.
- Button-based components such as `Button`, `Tab`, and `CompositeItem` still pay one mount re-render from `Command`'s `isNativeButton` detection, which is intentionally untouched — optimistically seeding it would produce an invalid transient `type="button"` on render-swapped elements.
- The new `focusable-mount-renders` sandbox instruments mount render counts (native button, plain div, hinted div) with a happy-dom test (relative assertions, StrictMode-safe) and a Chrome test (absolute counts; Safari renders native buttons twice by design). Both were red-checked: the button assertion fails against main's `Focusable`, and the hinted-div assertion fails when the seeds ignore the hint.
- The `Focusable` changes were originally extracted from #6616, whose composite performance work has since merged into main and been merged back into this branch.


packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -390,9 +391,29 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
+    // element matches the expected tag (button, input, etc. by default, or
+    // div for components that declare a non-native default tag). The extra
+    // detection render only happens when the rendered element's traits differ
+    // from the expected ones, such as a custom tag passed via the render prop.
+    const [nativeTabbableTag, setNativeTabbableTag] = useState(() =>
+      isNativeTabbable(defaultTagName),
+    );
+    const [supportsDisabledTag, setSupportsDisabledTag] = useState(() =>
+      supportsDisabledAttribute(defaultTagName),
+    );
```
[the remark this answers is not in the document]
diegohaz: Not a problem.

.changeset/200-single-mount-render.md
```diff
@@ -0,0 +1,12 @@
+---
```
diegohaz, line 1: Avoid em dashes here. Also make the changeset more concise. It's too long. Reduce it to two paragraphs besides the heading.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -579,6 +600,10 @@ export interface FocusableOptions<
   onFocusVisible?: BivariantCallback<
     (event: SyntheticEvent<HTMLElement>) => void
   >;
+  /**
+   * @private
+   */
```
diegohaz, line 609: Add JSDoc description for internal use

.changeset/200-single-mount-render.md
diegohaz: Fixed in db6661f8 - removed the em dashes and reduced the changeset to two paragraphs besides the heading.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in db6661f8 - added a JSDoc description for internal use above the `@private` tag.

diegohaz:
**Fixed**

- db6661f8 ([reply](https://github.com/ariakit/ariakit/pull/6622#discussion_r3518718801)): Reduced the changeset to two paragraphs besides the heading and removed the em dashes.
- db6661f8 ([reply](https://github.com/ariakit/ariakit/pull/6622#discussion_r3518719003)): Added a JSDoc description for internal use to the `unstable_defaultTagName` option.

packages/ariakit-react-components/src/tab/tab-panel.tsx
```diff
@@ -53,6 +53,7 @@ export const useTabPanel = createHook<TagName, TabPanelOptions>(
     scrollElement,
     ...props
   }) {
+    props = { unstable_defaultTagName: TagName, ...props };
```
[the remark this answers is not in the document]
diegohaz: Fixed in 6b97e797 - removed TabPanel's default tag name hint so panels that contain tabbable children don't emit a transient `tabindex="0"` on initial and server-rendered output, with a comment explaining why the hint doesn't apply (focusability depends on post-mount `hasTabbableChildren`).

diegohaz:
**Fixed**

- 6b97e797 ([reply](https://github.com/ariakit/ariakit/pull/6622#discussion_r3519041113)): Removed TabPanel's default tag name hint. Its focusability depends on post-mount `hasTabbableChildren`, so the hint emitted a transient `tabindex="0"` on initial/SSR renders of panels with tabbable children.

**Perf regression (dialog-perf open dialog)**

- 36805aa4: Root-caused and fixed. The `--dialog-viewport-height` effect costs ~134ms on the dialog-perf page on **both** main and this PR (the PR added no work). On main, Focusable's tag detection render forces a layout state update that, with production React, defers the dialog's open passive effects until after the first paint, so that cost stayed off the interaction's critical path. Dialog's hint removed that render, pulling the same work before first paint (hence the INP +79% and in-window scripting increase; dev builds don't reproduce it). The fix removes the dialog's hint so the whole dialog family keeps the detection render and main's exact open timing, while items and composite containers keep the zero-extra-render win.

  Local production A/B (median scripting / INP over 5 iterations, `open dialog`):

  | Build | Scripting | INP |
  | --- | --- | --- |
  | main base | 86.1ms | 144ms |
  | PR before fix | 125.9ms | 176ms |
  | PR after fix | 85.9ms | 144ms |

  Verified on this branch: tsc, lint, core suite, and full React 18/19 suites green (React 19 suite ran 3× green after one unrelated `menu-nested-combobox` load flake). The changeset no longer names `Dialog` as a beneficiary. The next perf CI run should confirm.

diegohaz:
**Perf follow-up: corrected root cause and a construction-level fix (bd39bc8e)**

The CI perf run on 36805aa4 still flagged the open dialog regression, contradicting my local macOS A/B. Digging into the CI artifacts settled it:

- The expensive function shows ~430ms self-time in **both** baseline and current CI profiles. The work is identical on both sides; what differs is whether it runs before or after the dialog's first paint.
- The real cost is not the `--dialog-viewport-height` write but the **`visualViewport.height` read**: when the dialog's viewport effect runs before the browser paints the newly mounted dialog, that read forces a synchronous reflow on the still-dirty ~20k-element page (attributed to scripting, blocking INP). After paint, the tree is clean and the read is nearly free.
- Whether React flushes that passive effect before or after paint is scheduling-dependent. On main, the tag detection renders across the whole dialog subtree reliably pushed it past paint. Un-hinting only `Dialog` restored enough of that on fast macOS, but not on CI's slower Linux runners, where every current-build iteration still ran it pre-paint.

bd39bc8e makes the ordering hold **by construction** instead of by scheduling luck: the initial viewport write is deferred with the existing `afterPaint` helper (double rAF), so the read always lands on a clean tree. The stylesheets' viewport-unit fallbacks (`100vh`/`100dvh`) cover the first frames, matching what main already renders whenever its effect runs post-paint. The `resize` path is unchanged.

tsc, lint, dialog example tests, and the full React suite are green on the rebased head. The next perf run should adjudicate; if it's still flagged, the remaining variable would be the measurement window itself rather than user-facing timing (INP is the number to watch).

diegohaz:
**Perf status after bd39bc8e: not fixed — pausing for direction**

The perf run on bd39bc8e still flags open dialog (scripting +127%, INP +74%), which falsifies the single-effect theory: ~250ms of pre-paint work per open can't be the viewport read alone (~70ms/open). Corrected picture from the CI artifacts:

- The regression is the dialog's **entire open passive-effect batch** (tree snapshot walk over the ~20k-element page, viewport reflow, focus management — ~250ms total on CI runners) executing **before** the first paint on this branch, versus **after** it on main. All of that work exists identically on both sides.
- On main, the tag-detection state updates across the dialog's whole subtree (every input/button corrected post-mount) reliably push React's passive-effect flush past paint. Un-hinting `Dialog` restores only the dialog's own corrective render, which is enough on fast machines (macOS production A/B is at baseline) but not on CI Linux. The exact React scheduling rule that distinguishes one corrective render from many is what I haven't pinned down.
- bd39bc8e (defer the viewport read/write with `afterPaint`) stands on its own — it removes a real forced-reflow hazard from the critical path — but it only moves one effect, not the batch.

Options from here, in rough order of preference:

1. **Explicitly schedule the dialog's heavy open effects post-paint** (`afterPaint` for the tree snapshot the way bd39bc8e did for the viewport write). Codifies main's observed timing by construction; needs care with focus-management ordering.
2. **Reintroduce a deliberate second-commit barrier in `useDialog`** (a mount layout-effect state toggle) — but note the plain un-hint already produces one corrective render and didn't move CI, so this likely needs the subtree-wide shape to work, which reintroduces the renders this PR removes for dialog content.
3. **Accept the CI metric for now** and land the item/composite wins (the largest benefit), tracking the dialog-open flush placement as a follow-up investigation.

Local repro remains impossible on macOS (production builds, 4× CPU throttle, CI-built artifacts all measure at baseline), so any further iteration will need CI runs to adjudicate. Happy to pursue whichever direction you prefer.

diegohaz:
**Experiment: merged #6609 into this branch (3dba22c9)**

Per discussion, this brings the "Paint modal dialogs before disabling the outside tree" work onto this branch: the inert tree-disabling — the biggest chunk of the ~250ms pre-paint batch — is now explicitly deferred until right after the open frame paints (marking for outside-click/Escape stays synchronous), and zero-width `--scrollbar-width` writes are skipped. Combined with the earlier viewport-write deferral, the dialog's heavy open effects no longer depend on React's scheduling to stay off the paint's critical path.

The merge was clean (zero conflict resolutions; `dialog.tsx` is the only overlapping file, and both sides verified intact bidirectionally). Also carried along: latest main incl. the quiescence-based perf settle (#6625). All suites green on the merged head: core 532, React 844/844, React 18 836+8, tsc, lint. Local production check: INP 136 vs 144 baseline.

One thing to watch in the perf run: #6609's fast path applies when no visible scrollbar is being removed. If CI's headless Linux Chrome uses classic scrollbars on this overflowing page, the inert writes stay intentionally synchronous there (sharing the scroll-lock's pre-paint recalc, matching main) — in that case the number to read is whether current matches *baseline*, not whether the absolute cost disappears.

diegohaz:
**Perf: resolved (3dba22c9)**

With #6609 merged in, the perf run no longer flags open dialog at all — it's now substantially faster than main: scripting 489.9ms → 292.7ms (-40%), INP 624ms → 312ms (-50%), with close dialog also improving (total -22%). The inert deferral's fast path does apply on CI's headless Chrome, so the dialog's heavy open effects are now off the paint's critical path by construction rather than by scheduling luck. All other checks are green.

Two bookkeeping notes for merge time:
- This branch now contains #6609's commits and its `.changeset/200-dialog-open-paint.md`; if #6609 lands separately first, a rebase dedupes them cleanly.
- With the explicit deferrals proven on CI, re-hinting the dialog family (reverting 36805aa4's un-hint to reclaim its render win) is a viable follow-up experiment — the un-hint comment in `dialog.tsx` documents the dependency to revisit.

diegohaz:
**SelectItem re-hinted (e0454845)**

Merged latest main (including #6627, the composite virtual-focus redirect fix for #6623) and restored `SelectItem`'s `unstable_defaultTagName` hint — it now mounts without the extra render like the other `div`-based items.

Verified against the old race signature: the full React suite ran green four times in a row with the hint restored (it was 4/4 red under full-suite load before #6627), plus the `select-6623` regression sandbox from #6627, React 18, core, tsc, and the select example tests. The PR description's SelectItem note is updated.

diegohaz:
This doesn't improve performance enough to justify the merge. See https://github.com/ariakit/ariakit/pull/6622#issuecomment-4874454532

PR #6627, opened 2026-07-03
Fix composite virtual focus getting stuck on an item focused before the base element is set

### LINKED ISSUE #6623: Composite virtual focus can get stuck on an auto-focused item when it's focused before the store's baseElement is set
## Summary

With `virtualFocus`, a composite item that receives DOM focus is supposed to redirect focus to the composite container (`useCompositeItem`'s `onFocus` calls `baseElement.focus()`). That redirect is silently dropped when the focus event fires before the container has published `baseElement` to the store:

https://github.com/ariakit/ariakit/blob/044b9d481662ef11031f940140d2b61767aae1e9/packages/ariakit-react-components/src/composite/composite-item.tsx#L279-L282

Nothing retries the redirect afterwards, and the Dialog auto-focus-on-show effect that runs next picks the same `[data-autofocus=true]` item and calls `element.focus()` — a no-op, since the item already has focus, so no new focus event fires. DOM focus stays on the item, and the virtual-focus container (e.g. a Select's listbox) never receives focus, breaking the `aria-activedescendant` contract.

Today this is latent: it currently only manifests when the item mounts without an extra render, and `Focusable`'s tag-detection state update happens to force one on every mount. It surfaced while working on #6622, which removes that extra render — `SelectItem` had to be excluded from that optimization because of this race. Any future change that removes or defers that re-render (React version/scheduling changes, memoization, compiler optimizations) can expose it the same way.

## Verified failure sequence

Reproduced with the `examples/select` "show on arrow keys" test and per-event tracing in `useCompositeItem`, `useDialog`, `useFocusable`, and `createCompositeStore`:

```
item-onFocus id=…/apple  vf=true  base=null  rel=button#combobox  self=true
item-onFocus id=…/apple  exit=baseElementNotConnected     ← redirect dropped
setBaseElement <listbox>                                  ← store publication lands too late
dialog-autofocus-pick div#…/apple
dialog-autofocus-run  div#…/apple  active=div#…/apple     ← focus() no-ops; no recovery
```

Stack trace of the early focus event (captured at the dropped redirect):

```
at useCompositeItem.onFocus (composite-item.tsx)
… happy-dom focus event dispatch …
at HTMLDivElement.focus
at focusIntoView (packages/ariakit-utils/src/focus.ts:460)
at composite.tsx:198 (focusOnMove effect)
at withBaseScrollPreserved (composite.tsx:122)
```

So the ordering is:

1. `ArrowDown` on the `Select` button calls `store.show()` and `store.move(selectedId)` (select.tsx:130), so `moves > 0`.
2. The popover and items mount. `useComposite`'s `focusOnMove` passive effect (composite.tsx:190–199) runs and focuses the active item element via `focusIntoView`.
3. The container publishes `baseElement` through `useTransactionState` (react-utils hooks.ts), which requires a second render pass: a ref-callback `setState` followed by a layout effect keyed on that state. Under load, that second pass can land after the passive `focusOnMove` effect.
4. The item's `onFocus` redirect hits `if (!baseElement?.isConnected) return` and drops.
5. The dialog auto-focus effect no-ops as described above. Focus is stuck on the item.

## Reproduction

The window is timing-sensitive (it needs the store update to land after the passive effect), so it reproduces deterministically under full-suite load but not standalone:

1. In a worktree on `main`, apply the `Focusable` change from #6622 (`packages/ariakit-react-components/src/focusable/focusable.tsx`) and add the hint to `useSelectItem` so the selected item mounts without the extra tag-detection render:
   ```ts
   props = { unstable_defaultTagName: TagName, ...props };
   ```
2. Run the full suite: `pnpm test-react`. `examples/select/test.ts` "show on arrow keys" fails at `expect(q.listbox()).toHaveFocus()` — DOM focus is on the Apple option instead of the listbox (4/4 runs with the hint; 2/2 green without; always green standalone).

## Possible directions

- Make the redirect robust instead of dropping: when the item is focused with `virtualFocus` and `baseElement` is missing or disconnected, re-attempt once the store publishes `baseElement` (e.g. subscribe to the `baseElement` state or re-check in a microtask), instead of returning silently.
- Publish `baseElement` to the store synchronously from the ref callback (avoiding the `useTransactionState` second-pass delay), keeping the rollback-on-cleanup semantics.
- Have the `focusOnMove` effect wait for `baseElement` before focusing items in virtual-focus mode, since the redirect it triggers depends on it.

Once fixed, `SelectItem` can adopt the `unstable_defaultTagName` hint from #6622 (see the comment in `select-item.tsx`) and mount with no extra renders like the other item components.

## Workaround

https://github.com/ariakit/ariakit/issues/6623#issuecomment-4875637265

diegohaz:
## Motivation

With virtual focus (e.g., `SelectPopover`, `ComboboxPopover`), a composite item that receives DOM focus is supposed to redirect focus to the composite element so the `aria-activedescendant` contract holds. That redirect was silently dropped when the item's focus event fired before the composite element was available in the store:

https://github.com/ariakit/ariakit/blob/a12b87121e5d537307e1bd6371e5aee79b09a543/packages/ariakit-react-components/src/composite/composite-item.tsx#L279-L282

Nothing retried the redirect afterwards, so DOM focus got stuck on the item and the composite element (e.g., a Select's listbox) never received focus, even though the active item state (`data-active-item`, `aria-activedescendant`) was correctly updated.

The composite element is stored through `useTransactionState`, which requires a later commit (ref-callback state + layout effect), so any focus that lands on an item during the mount commit hits this window. This can be triggered from userland today (see the reproduction below) and is also the race that currently blocks `SelectItem` from adopting the mount-render optimization in #6622.

See #6623 for the full analysis and verified failure sequence.

## Reproduction

The new `app/src/sandbox/select-6623/` sandbox models an app that manages initial focus itself (`autoFocusOnShow={false}` on an `unmountOnHide` popover): a "recommended" option focuses itself in a `useLayoutEffect` on mount. The layout effect runs during the popover's mount commit, before the composite element lands in the store, so the redirect was dropped and DOM focus got stuck on the option in every browser:

```jsx
function RecommendedSelectItem(props) {
  const ref = useRef(null);
  useLayoutEffect(() => {
    ref.current?.focus();
  }, []);
  return <Ariakit.SelectItem {...props} ref={ref} tabIndex={-1} />;
}
```

The happy-dom and Playwright tests assert that the listbox receives DOM focus and the option becomes the active item. Before the fix, they failed in happy-dom and in Chrome, Firefox, and Safari (see the first two commits in this PR for CI runs demonstrating the failing repro and the passing workaround).

The sandbox also covers the cancellation path: an optional app-managed search input focuses itself after the recommended option, before the listbox is available, and the tests assert that the scheduled redirect is discarded once the option loses DOM focus instead of stealing focus back to the listbox.

## Solution

Instead of silently dropping the redirect when the composite element is missing from the store or disconnected from the DOM, `useCompositeItem`'s `onFocus` now schedules it: it subscribes to the store's `baseElement` state and performs the exact same redirect once a connected composite element is available, as long as the item still has DOM focus by then.

The redirect logic (Safari scroll handling, intermediate blur bookkeeping, silent vs. propagating focus) is shared between the immediate and the scheduled paths, so the event sequence consumers observe is the same in both cases, just delayed until the composite element exists.

The scheduled redirect is self-cleaning: it unsubscribes on the next `baseElement` change once the item no longer has DOM focus (which also covers unmounted items), and at most one redirect is pending per item. It deliberately has no unmount cleanup effect, since the subscription is created by a focus event rather than an effect, and an unmount cleanup would permanently cancel it during React Strict Mode's simulated unmount.

This also unblocks `SelectItem` from adopting the `unstable_defaultTagName` mount-render optimization in #6622, which was excluded there because the internal `focusOnMove` effect could hit the same dropped redirect under load.

## Workaround

Until the fix lands, apps that focus a composite item before the composite container element is available (for example, focusing an option while the popover is still mounting) can re-apply the redirect themselves once the store publishes the container element. Subscribe to the store's `baseElement` state and move DOM focus to it if it's still stuck on the item:

```jsx
function RecommendedSelectItem(props) {
  const select = Ariakit.useSelectContext();
  const ref = React.useRef(null);

  React.useLayoutEffect(() => {
    ref.current?.focus();
  }, []);

  // Workaround: if this option was focused before the select store had a
  // reference to the listbox element, the virtual focus redirect to the
  // listbox was dropped, so we redirect manually once the listbox element is
  // available and DOM focus is still stuck on this option.
  // TODO: Remove once https://github.com/ariakit/ariakit/issues/6623 is fixed.
  const virtualFocus = Ariakit.useStoreState(select, "virtualFocus");
  const baseElement = Ariakit.useStoreState(select, "baseElement");
  React.useLayoutEffect(() => {
    if (!virtualFocus) return;
    if (!baseElement) return;
    if (document.activeElement !== ref.current) return;
    baseElement.focus();
  }, [virtualFocus, baseElement]);

  return <Ariakit.SelectItem {...props} ref={ref} tabIndex={-1} />;
}
```

The same pattern applies to any composite widget with `virtualFocus`: subscribe to `baseElement` with `useStoreState` and redirect focus to it when it becomes available while an item still holds DOM focus.

Fixes #6623

### DIFF
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -323,7 +323,16 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
         }
       };
 
-      // The base element may not be available at this point: it's stored in a
+      if (baseElement?.isConnected) {
+        redirectFocusToBaseElement(
+          event.currentTarget,
+          event.relatedTarget,
+          baseElement,
+        );
+        return;
+      }
+
+      // The base element isn't available at this point: it's stored in a
       // later commit than the one that mounts it, so focusing an item during
       // the mount commit lands here before the store has it. It may also be
       // disconnected from the DOM, such as when it's removed just before this
@@ -332,48 +341,40 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
       // base element is available in the store, then redirect focus to it if
       // this item still has DOM focus.
       // See https://github.com/ariakit/ariakit/issues/6623
-      if (!baseElement?.isConnected) {
-        // Items that opt out of registering themselves in the store never
-        // produce the unregister store update that the scheduled redirect
-        // below relies on to self-clean, so they keep the previous behavior
-        // of dropping the redirect.
-        if (shouldRegisterItem === false) return;
-        const { currentTarget, relatedTarget } = event;
-        const cancelScheduledFocusRedirect = () => {
-          cancelScheduledFocusRedirectRef.current?.();
-          cancelScheduledFocusRedirectRef.current = null;
-        };
-        cancelScheduledFocusRedirect();
-        // Subscribe to every store update, not just baseElement changes, so
-        // the pending redirect is also discarded when the item unmounts
-        // without a base element ever arriving.
-        cancelScheduledFocusRedirectRef.current = subscribe(store, null, () => {
-          // The redirect is no longer relevant if the item lost DOM focus in
-          // the meantime, including when it was unmounted.
-          if (getActiveElement(currentTarget) !== currentTarget) {
-            cancelScheduledFocusRedirect();
-            return;
-          }
-          const state = store.getState();
-          const nextBaseElement = state.baseElement;
-          // Keep waiting until a connected base element is stored.
-          if (!nextBaseElement?.isConnected) return;
-          cancelScheduledFocusRedirect();
-          if (!state.virtualFocus) return;
-          redirectFocusToBaseElement(
-            currentTarget,
-            relatedTarget,
-            nextBaseElement,
-          );
-        });
-        return;
-      }
 
-      redirectFocusToBaseElement(
-        event.currentTarget,
-        event.relatedTarget,
-        baseElement,
-      );
+      // Items that opt out of registering themselves in the store never
+      // produce the unregister store update that the scheduled redirect below
+      // relies on to self-clean, so they keep the previous behavior of
+      // dropping the redirect.
+      if (shouldRegisterItem === false) return;
+      const { currentTarget, relatedTarget } = event;
+      const cancelScheduledFocusRedirect = () => {
+        cancelScheduledFocusRedirectRef.current?.();
+        cancelScheduledFocusRedirectRef.current = null;
+      };
+      cancelScheduledFocusRedirect();
+      // Subscribe to every store update, not just baseElement changes, so the
+      // pending redirect is also discarded when the item unmounts without a
+      // base element ever arriving.
+      cancelScheduledFocusRedirectRef.current = subscribe(store, null, () => {
+        // The redirect is no longer relevant if the item lost DOM focus in
+        // the meantime, including when it was unmounted.
+        if (getActiveElement(currentTarget) !== currentTarget) {
+          cancelScheduledFocusRedirect();
+          return;
+        }
+        const state = store.getState();
+        const nextBaseElement = state.baseElement;
+        // Keep waiting until a connected base element is stored.
+        if (!nextBaseElement?.isConnected) return;
+        cancelScheduledFocusRedirect();
+        if (!state.virtualFocus) return;
+        redirectFocusToBaseElement(
+          currentTarget,
+          relatedTarget,
+          nextBaseElement,
+        );
+      });
     });
 
     const onBlurCaptureProp = props.onBlurCapture;
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -165,6 +165,9 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
     const row = useContext(CompositeRowContext);
     const disabled = disabledFromProps(props);
     const trulyDisabled = disabled && !props.accessibleWhenDisabled;
+    // Snapshot before the props object is replaced below (useCollectionItem
+    // consumes this prop), so the onFocus handler can read it at event time.
+    const shouldRegisterItem = props.shouldRegisterItem;
 
     // Sibling selectors below (ariaPosInSet) also need the row id during
     // render. They can't read the destructured rowId const since it's declared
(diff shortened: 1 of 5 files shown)


diegohaz:
**Fixed**

- 3d9d002cf: Added regression coverage for the cancellation path raised in CodeRabbit's review nitpick. The sandbox now includes an app-managed search input that focuses itself after the recommended option, before the listbox element is stored, and the new happy-dom and Playwright tests assert that the scheduled redirect is discarded instead of stealing focus back to the listbox. Verified red by temporarily removing the cancellation guard in `composite-item.tsx` (the new test then fails with focus stolen to the listbox) and green with the guard restored.

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -276,37 +286,84 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
...
+        const { currentTarget, relatedTarget } = event;
+        const cancelScheduledFocusRedirect = () => {
+          cancelScheduledFocusRedirectRef.current?.();
+          cancelScheduledFocusRedirectRef.current = null;
+        };
+        cancelScheduledFocusRedirect();
+        cancelScheduledFocusRedirectRef.current = subscribe(
+          store,
+          ["baseElement"],
+          () => {
```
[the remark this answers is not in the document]
diegohaz: Fixed in 781ab5cc1 - the pending redirect subscription now listens to every store update instead of only `baseElement` changes, so it also self-cleans when the composite unmounts before the base element is ever stored: unmounting a registered item unregisters it from the store, which is guaranteed to produce an update, and the listener then cancels itself since a detached item can't be the active element. Items that opt out of registration via `shouldRegisterItem` never produce that update, so they keep the previous behavior of dropping the redirect instead of scheduling it. There's no practical automated test for the listener lifecycle itself (registration counts are store-internal), but the behavioral cancellation path is covered by the "discards the redirect" tests added in 3d9d002cf, and the opt-out gate was verified with temporary instrumentation (no subscription is created for `shouldRegisterItem={false}` items).

diegohaz:
**Fixed**

- 781ab5cc1 ([reply](https://github.com/ariakit/ariakit/pull/6627#discussion_r3519814522)): Addressed the P2 review comment about pending redirects lingering when the base element never arrives. The scheduled redirect subscription now listens to every store update, so unmounting the item (which unregisters it from the store) always gives it a cancel pulse; items that opt out of registration via `shouldRegisterItem` skip scheduling entirely and keep the previous drop behavior.

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -276,37 +292,88 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
...
-      // component, we'll transfer focus silently to the composite element.
-      // That's because this is just a transition event, the composite element
-      // was likely already focused, so we're just immediately returning focus
-      // to it when navigating through the items.
-      const fromComposite =
-        event.relatedTarget === baseElement ||
-        isItem(store, event.relatedTarget);
-      if (fromComposite) {
-        focusSilently(baseElement);
+        return;
```
diegohaz: Maybe we should invert the early return logic here. Could we have `if (baseElement?.isConnected) { ... }` then this logic below?

diegohaz: Fixed in 350e72cc8 - inverted the logic so the connected path comes first in a positive `if (baseElement?.isConnected)` block that redirects and returns, and the scheduling logic now follows un-nested. Behavior is unchanged (all sandbox tests and the full React suite pass).

diegohaz:
**Fixed**

- 350e72cc8 ([reply](https://github.com/ariakit/ariakit/pull/6627#discussion_r3520213322)): Inverted the early return logic in the composite item focus redirect as requested — the connected path now comes first in a positive `if (baseElement?.isConnected)` block, and the scheduling logic follows without the extra nesting level. Pure restructure; behavior unchanged.

PR #6634, opened 2026-07-03
Reserve scrollbar space with CSS scrollbar-gutter in modal dialogs

### LINKED ISSUE #4345: `‎usePreventBodyScroll‎` can cause layout shift when `overflow-y: scroll`
### Current behavior

When the `html` element has `overflow-y: scroll` set, and the OS is set to always show scroll bars, `usePreventBodyScroll` (for example in the Dialog component) will cause the layout to shift unnecessarily.

1. Open https://stackblitz.com/edit/fzv7hbg3?file=index.html
2. Set your OS to always show scrollbars (e.g. on Mac, System Settings ▸ Appearance ▸ Show scroll bars).
3. Clicking the "Show modal" button will cause an unnecessary layout shift, due to the scrollbar width being added as a `body` padding.

### Expected behavior

When the scrollbar is always visible regardless of overflow, `usePreventBodyScroll` should not add `body` padding to compensate for the scrollbar width.

### Workaround

_No response_

### Possible solutions

_No response_

diegohaz:
## Motivation

When a modal dialog opens with `preventBodyScroll` (the default), the scroll lock measures the scrollbar, pads the `body` element to compensate for the removed scrollbar, and defines a `--scrollbar-width` CSS variable on the `html` element so userland `position: fixed` elements can compensate too:

```css
.header {
  position: fixed;
  padding-right: calc(16px + var(--scrollbar-width, 0));
}
```

This has a few costs:

- Every root dialog open writes a custom property to the root element (invalidating style for anything referencing the variable) and writes padding to the body, even on overlay-scrollbar platforms (macOS, iOS) where the measured width is `0px` and the writes are pure overhead.
- Fixed elements shift unless users know about and apply the `--scrollbar-width` variable themselves.
- Pages that already set `scrollbar-gutter: stable` get double compensation: the gutter stays reserved while the lock also pads the body, so content shifts on open (the same bug class reported against Radix in radix-ui/primitives#3251 and formerly React Aria in adobe/react-spectrum#5470).
- Pages with `overflow-y: scroll` on the html element (an always-visible scrollbar) get the same double compensation: the scrollbar stays while the body is padded, shifting the layout (#4345). Worse, on such pages the page scrolls through the html element, so the body overflow doesn't propagate to the viewport and the lock doesn't actually prevent scrolling.

Closes #4345

## Solution

Use the CSS [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) property instead, the approach React Aria adopted in adobe/react-spectrum#8297 and Base UI uses in its `useScrollLock`. Per the CSS Overflow spec, `scrollbar-gutter: stable` keeps the scrollbar's space reserved when `overflow` is `hidden`, "regardless of whether the box is actually overflowing", and since Chrome 136, Firefox 97, and Safari 18.2 the reserved gutter also insets the containing block for `position: fixed` elements (verified by the `scrollbar-gutter-fixedpos-*` WPT tests, which pass in all three engines).

The lock now takes one of three branches:

1. No space-consuming scrollbar (overlay scrollbars, page that doesn't overflow): only `overflow: hidden` on the body. No more `--scrollbar-width: 0px` and `padding-right: 0px` writes on this common path.
2. `scrollbar-gutter` supported: `scrollbar-gutter: stable` plus hidden overflow on the `html` element (the property only applies to the viewport from the root element and doesn't propagate from the body). Nothing shifts, no measurement bookkeeping, no variable, and the body styles are untouched. The properties are set individually via `setCSSProperty` rather than `assignStyle`, so the restore is property-scoped and doesn't clobber unrelated inline styles (such as theme variables or `color-scheme`) written to `html` while the dialog is open. The overflow is hidden through the `overflow-x` and `overflow-y` longhands, so an inline longhand the page set itself (such as the `overflow-y: scroll` from #4345) is captured and restored exactly; setting the `overflow` shorthand would overwrite it and then drop it on restore.
3. Fallback for browsers without support (Safari before 18.2): the previous behavior, body padding plus the `--scrollbar-width` variable.

Branches 1 and 3 additionally hide the html overflow (same longhand treatment) when the html element's computed overflow isn't `visible`: on such pages the body overflow doesn't propagate to the viewport, so the body-only lock never actually locked the page scroll (the second half of #4345).

Re-locks that happen while a gutter is already reserved measure a `0px` scrollbar because `clientWidth` includes the reserved gutter. That state occurs both when a previous lock is still active (StrictMode remounts re-lock before the microtask-deferred restore runs) and when the page reserves the gutter itself. The lock therefore reads the computed `scrollbar-gutter` value to stay on the gutter branch, mirroring the existing `--scrollbar-width` read that serves the same purpose on the fallback branch, and echoes that computed value so an author's `scrollbar-gutter: stable both-edges` is preserved rather than overridden with plain `stable`.

The iOS path is unchanged except it no longer defines the `--scrollbar-width` variable, which was always `0px` there since iOS scrollbars are overlay.

Existing userland styles keep working: the documented pattern resolves to its fallback value on the gutter branch, which is now the correct compensation since fixed elements no longer shift. The docs previously showed `var(--scrollbar-width, 0)` inside `calc()`, but a unitless `0` fallback is invalid there (a number can't be added to a length, making the declaration invalid at computed-value time), so the examples now use `0px`.

## Trade-offs

These match what React Aria and Base UI accepted with the same approach:

- The reserved gutter is blank space, so an `inset: 0` backdrop doesn't paint over the gutter strip while the dialog is open (Base UI closed the equivalent report, mui/base-ui#4615, as expected behavior). The previous approach removed the scrollbar entirely, so the backdrop covered that strip.
- Author-styled `::-webkit-scrollbar` scrollbars on overlay-scrollbar platforms consume space but get no gutter, so those pages shift on open. The measurement approach handled this; React Aria accepted the same regression when it switched.
- Chrome 94 to 135 parses `scrollbar-gutter` but doesn't inset fixed-position containing blocks by the reserved gutter (fixed in Chrome 136, April 2025), so `position: fixed` elements shift while a modal is open on those stale versions. A runtime behavior probe could detect this, but it would add forced layout work to every first dialog open, against this PR's performance goal, and React Aria makes the same call by gating only on property support. Android WebView isn't exposed: overlay scrollbars measure `0px` there, so those pages take the body-only path rather than the gutter branch.

## Verification

Beyond the updated happy-dom tests (happy-dom deterministically takes the gutter branch: `CSS.supports` returns `true` and `innerWidth - clientWidth` is 1024) and a new test that mocks `CSS.supports` to cover the fallback branch, the change was verified in real Chromium via Playwright against the dev app preview:

- Overlay scrollbars (macOS default): only `body { overflow: hidden }` is applied and restored; the `html` element is never touched.
- Native classic scrollbars (forced via the macOS `AppleShowScrollBars` per-app preference, headful): opening applies `scrollbar-gutter: stable; overflow: hidden` to `html`, in-flow content and `position: fixed` probes keep the exact same bounding rects while locked (no layout shift), and closing restores cleanly.
- Author-styled `::-webkit-scrollbar` plus StrictMode double-invoke: re-locks stay on the gutter branch and restore cleanly.
- Author `html { scrollbar-gutter: stable both-edges }` stylesheet: the lock echoes `stable both-edges` and the computed value stays intact throughout open and close.
- Inline `color-scheme` and custom-property writes to `html` made while a dialog is open survive the close, since the restore is property-scoped.

The legacy `dialog-backdrop-scrollable` Chrome test dragged the scrollable backdrop's scrollbar at `viewport width - 5`, a point that now lands on the blank reserved gutter under classic scrollbars (the fixed backdrop stops at the gutter). The test now targets the backdrop's own bounding box right edge, which degrades to the previous coordinates with overlay scrollbars. Verified red (original test fails with forced classic scrollbars, `scrollTop` stuck at 10) and green (fixed test passes with both classic and overlay scrollbars) against a live website dev server.

The new `dialog-4345` sandbox reproduces the issue scenario (inline `html { overflow-y: scroll }`, overflowing page, modal dialog) with browser and happy-dom tests. The browser test was verified red on the pre-fix code (the wheel scrolled the page behind the modal by exactly the wheel delta, since the body-only lock has no effect there) and green on the fix in all three desktop browsers. With forced classic scrollbars, the full issue scenario was verified end to end: the always-visible scrollbar consumes 15px before open, content keeps the exact same bounding rects while locked (no shift), the wheel is blocked, and the inline `overflow-y: scroll` is restored byte-exact on close.

Full `pnpm test-react` and `pnpm test-react18` suites, `pnpm tsc`, and the `@ariakit/react-components` build all pass.


app/src/sandbox/dialog-4345/test.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+test("fallback scroll lock also hides the html overflow", async () => {
+  const { documentElement, body } = document;
+  // happy-dom's window.CSS getter returns a fresh object on every access, so
+  // the getter itself must be mocked rather than a single instance's method.
+  const unsupportedCSS: Pick<typeof CSS, "supports"> = {
+    supports: () => false,
+  };
+  const supports = vi
```
diegohaz: Could we use `using supports = vi` here?

examples/dialog-nested-various/test.ts
```diff
@@ -28,6 +33,36 @@ test("show dialog and hide with escape", async () => {
...
 });
 
+test("fall back to body padding without scrollbar-gutter support", async () => {
+  const { documentElement, body } = document;
+  // happy-dom's window.CSS getter returns a fresh object on every access, so
+  // the getter itself must be mocked rather than a single instance's method.
+  const unsupportedCSS: Pick<typeof CSS, "supports"> = {
+    supports: () => false,
+  };
+  const supports = vi
```
diegohaz: Same here?

packages/ariakit-react-components/src/dialog/utils/use-prevent-body-scroll.ts
```diff
@@ -44,26 +55,69 @@ export function usePreventBodyScroll(
...
+      // clobber unrelated inline styles (such as theme variables) written to
+      // html while the dialog is open.
+      if (hasGutter || supportsScrollbarGutter(win)) {
+        return chain(
+          setCSSProperty(
+            documentElement,
+            "scrollbar-gutter",
+            hasGutter ? scrollbarGutter : "stable",
+          ),
+          setCSSProperty(documentElement, "overflow", "hidden"),
```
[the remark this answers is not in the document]
diegohaz: Resolved in 2d7f055901ff2febe84c7d68d25dbd2dbc325f56 - the lock now hides the html overflow through the `overflow-x`/`overflow-y` longhands via per-property `setCSSProperty` calls, so an inline root longhand like `overflow-y: scroll` is captured and restored exactly (including its priority). Covered by the `dialog-4345` sandbox tests, which were verified red against the shorthand version.

packages/ariakit-react-components/src/dialog/utils/use-prevent-body-scroll.ts
```diff
@@ -44,26 +55,69 @@ export function usePreventBodyScroll(
...
+        return assignStyle(body, { overflow: "hidden" });
+      }
+      // Keep the scrollbar's space reserved while `overflow: hidden` removes
+      // the scrollbar itself, so neither in-flow content nor `position:
+      // fixed` elements shift. `scrollbar-gutter` must be set on the html
+      // element: it applies to the viewport from there and doesn't propagate
+      // from body. Set the properties individually so the restore doesn't
+      // clobber unrelated inline styles (such as theme variables) written to
+      // html while the dialog is open.
+      if (hasGutter || supportsScrollbarGutter(win)) {
```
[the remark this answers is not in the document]
diegohaz: This is a deliberate trade-off rather than a gap: fixed-position containing blocks respect the gutter since Chrome 136 (April 2025), so the exposed group is stale, auto-updating Chrome 94-135 installs with classic scrollbars. A runtime behavior probe would add forced style/layout work to every first dialog open, against this PR's performance goal, and React Aria makes the same call by gating only on property support. Android WebView is not exposed: overlay scrollbars measure `0px` there, so those pages take the body-only path rather than the gutter branch. Now documented in the PR description's trade-offs section.

app/src/sandbox/dialog-4345/test.ts
diegohaz: Fixed in 9fade48ca67a9430ca2484b6372d1b2c4a4a62b5 - replaced the try/finally with a `using` declaration, which restores the mocked getter through the spy's `Symbol.dispose`.

examples/dialog-nested-various/test.ts
diegohaz: Fixed in 9fade48ca67a9430ca2484b6372d1b2c4a4a62b5 - same `using` conversion here.

diegohaz:
Addressed all 4 unresolved review threads.

**Fixed**
- 9fade48 ([reply](https://github.com/ariakit/ariakit/pull/6634#discussion_r3521482515)): `app/src/sandbox/dialog-4345/test.ts` now uses a `using` declaration for the `CSS.supports` mock instead of try/finally.
- 9fade48 ([reply](https://github.com/ariakit/ariakit/pull/6634#discussion_r3521482556)): same `using` conversion in `examples/dialog-nested-various/test.ts`.

**Already resolved**
- 2d7f055 ([reply](https://github.com/ariakit/ariakit/pull/6634#discussion_r3521482121)): the gutter lock hides the html overflow through the `overflow-x`/`overflow-y` longhands with per-property capture/restore, so inline root longhands like `overflow-y: scroll` survive open/close; regression-covered by the `dialog-4345` sandbox tests.

**Invalid**
- Won't fix ([reply](https://github.com/ariakit/ariakit/pull/6634#discussion_r3521482374)): gating the gutter branch on fixed-position behavior is a deliberate non-goal; Chrome 94-135 is a stale auto-updating population, a runtime probe costs forced layout on every first dialog open, and React Aria gates only on property support. Now listed in the PR description trade-offs.

PR #6642, opened 2026-07-04
Add popup performance tests

diegohaz:
## Motivation

Combobox, select, and menu popup interactions need dedicated performance coverage with large surrounding DOMs, similar to the existing dialog performance sandbox. Without these benchmarks, regressions in opening popovers or moving through popup items are harder to catch.

## Solution

Add separate `combobox-perf`, `select-perf`, and `menu-perf` sandboxes. Each sandbox renders 5000 page buttons plus a 200-item popup, then measures:

- opening the popup
- closing it with `Escape`
- moving from the first item to the last with the keyboard
- opening with script profiling enabled

The popups use `unmountOnHide` and no CSS transitions so timings focus on popup work rather than animation overhead.

### DIFF
--- a/app/src/sandbox/combobox-perf/index.react.tsx
+++ b/app/src/sandbox/combobox-perf/index.react.tsx
@@ -1,0 +1,43 @@
+import * as Ariakit from "@ariakit/react";
+import "./style.css";
+
+const pageItems = Array.from(
+  { length: 5000 },
+  (_, index) => `Element ${index + 1}`,
+);
+
+const items = Array.from({ length: 20 }, (_, index) => `Item ${index + 1}`);
+
+export default function Example() {
+  return (
+    <div className="root">
+      <Ariakit.ComboboxProvider>
+        <Ariakit.ComboboxLabel className="label">
+          Search items
+        </Ariakit.ComboboxLabel>
+        <Ariakit.Combobox className="combobox" placeholder="Search items" />
+        <Ariakit.ComboboxPopover
+          className="popover"
+          gutter={4}
+          sameWidth
+          unmountOnHide
+        >
+          {items.map((item) => (
+            <Ariakit.ComboboxItem
+              key={item}
+              className="popover-item"
+              value={item}
+            />
+          ))}
+        </Ariakit.ComboboxPopover>
+      </Ariakit.ComboboxProvider>
+      <div className="grid">
+        {pageItems.map((item) => (
+          <button key={item} className="card" type="button">
+            {item}
+          </button>
+        ))}
+      </div>
+    </div>
+  );
+}
--- a/app/src/sandbox/combobox-perf/index.react.tsx
+++ b/app/src/sandbox/combobox-perf/index.react.tsx
@@ -6,7 +6,7 @@ const pageItems = Array.from(
   (_, index) => `Element ${index + 1}`,
 );
 
-const items = Array.from({ length: 20 }, (_, index) => `Item ${index + 1}`);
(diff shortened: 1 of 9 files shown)


diegohaz (commented):
## Review summary

**Type:** initial review · **Baseline:** merge-base `e1de3233` … head `dba8f33` · **Scope:** all 9 added files (3 new perf sandboxes: `combobox-perf`, `menu-perf`, `select-perf`).

This is a test-only PR that mirrors the existing `dialog-perf` / `composite-perf` sandboxes. I read the full diff, the two existing perf sandboxes, the perf workflow, and the relevant component/query source. **No substantive findings** — the change is correct, consistent with the established pattern, and cleanly wired into CI.

### What I verified

- **CI auto-discovery.** The Chrome perf matrix (`.github/workflows/perf.yml`, `find … '*perf*-chrome.ts'`) and the perf project's `testMatch: /\/perf[^/]*-chrome/` in `app/playwright.config.ts` both pick up the three new `perf-chrome.ts` files automatically. Job titles derive from the sandbox folder names (`combobox-perf`, `menu-perf`, `select-perf`), which are unique against the existing `dialog-perf` / `composite-perf`, so no title collisions and no manual matrix edits are needed.
- **Assertions resolve against the real components.** `Select` renders `role="combobox"`, so `q.combobox("Choose item")` matches the select trigger; `Combobox`'s `showOnClick` defaults to `canShow` (`value.length >= showMinLength`, `showMinLength` default `0`), so clicking the empty combobox opens the listbox and `verify*Open` holds. `option` / `listbox` / `menu` / `menuitem` roles all match the rendered items, and the `Element N` (page buttons) vs `Item N` (popup items) naming keeps names distinct.
- **Navigation math.** Each `move across items` setup activates `Item 1` first (combobox: `ArrowDown`; menu: `Home`; select: `defaultValue={items[0]}`), then the measured step presses `ArrowDown` 19× (`for i = 1; i < 20`) to land on `Item 20`, which the `verify` asserts via `data-active-item`. Correct and consistent across all three.
- **Conventions.** Folder naming (`<component>-perf`), omitting `preview.json` (title defaults to folder name), `unmountOnHide` + transition-free CSS, and the `open … (script profile)` variant all match `dialog-perf` / `composite-perf`. Test-only change, so no changeset is required (correctly none added).
- **Tooling.** `oxfmt --check`, `oxlint --disable-nested-config`, and `stylelint` all pass on the changed files. I did not execute the perf harness or a full `pnpm tsc` locally (both are heavy and CI covers them); the correctness checks above are source-level.

### Considered and dismissed

- **CodeRabbit nitpick — extract shared perf scaffolding across the three files.** I'd keep them self-contained. The established convention is that each perf sandbox's `perf-chrome.ts` is fully self-contained (`dialog-perf`, `composite-perf` share nothing), and the perf matrix runs one job per file; a shared helper would couple three otherwise-independent ~40-line sandboxes and force a reader to jump files to understand a single sandbox's flow. The DRY win doesn't outweigh the readability/isolation cost here.
- **No `move across items (script profile)` variant** (which `composite-perf` has). Reasonable as-is: these are popup sandboxes, so `dialog-perf` is the right analog and it profiles only the `open` (the expensive op — mounting the popup + positioning against the 5000-node DOM). A profiled 20-item move would add confirmation-round CI cost for little diagnostic value.

### Not fully validated

Interaction/timing behavior of the perf runner itself (I did not build the app and run `test-perf`); the reasoning above is based on reading the component and query source plus the two existing perf sandboxes.

PR #6647, opened 2026-07-04
Combine dialog perf script profile

diegohaz:
## Motivation

The dialog perf sandbox has a separate `open dialog (script profile)` test only to collect scripting details for the same interaction covered by `open dialog`. That creates an extra perf row instead of attaching the diagnostic profile to the primary interaction.

The generated perf labels also include implementation path segments like:

```text
sandbox/dialog-perf/perf-chrome.ts > react > open dialog
```

Those labels are noisy in PR comments.

## Solution

Move `scriptProfile: true` onto the existing `open dialog` measurement and remove the dedicated dialog script-profile test. The timing samples for that row remain unprofiled; the runner collects script profiles in separate diagnostic iterations and attaches the resulting `Script profile` table to the same row.

The dialog perf sandbox now renders 3000 items instead of 5000.

Perf result labels now omit the generated `sandbox` and `perf-chrome.ts` path segments, so this:

```text
sandbox/dialog-perf/perf-chrome.ts > react > open dialog
```

becomes:

```text
dialog-perf > react > open dialog
```

Generated labels are normalized consistently during comparison, including `dialog-perf`, so main and PR results still pair when sandbox code changes.

The comparison formatter treats profile data as an add-on to normal comparisons when it is attached to a timing row, rendering the `Script profile` table directly below the metric comparison table. Existing dedicated `(... script profile)` rows are still collected as profile-only diagnostics, but `perf-compare` merges them under their base labels and excludes their profiler-inflated metrics from timing comparisons and new-test metric rows.

If every shared round for a test is filtered out because one side only has profile-mode data, the comparison now reports the test as unpaired instead of silently dropping it.

Explicit `scriptProfile`/`selectorProfile` options on regular labels collect profiles in separate diagnostic iterations. Env-driven profiling through `PERF_SCRIPT_PROFILE`, `PERF_SELECTOR_PROFILE`, or `PERF_CSS_SELECTOR_PROFILE` keeps the old quick measured-profile path for local diagnostics.

The removed dialog profile test used `profileLimit: 20`; the combined test intentionally uses the default profile limit because the measure call only needs `{ scriptProfile: true }`.

## Verification

- `pnpm exec oxlint --disable-nested-config --fix packages/ariakit-scripts/src/perf.ts packages/ariakit-scripts/src/perf-compare.ts packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm exec oxfmt packages/ariakit-scripts/src/perf.ts packages/ariakit-scripts/src/perf-compare.ts packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm lint-fix`
- `pnpm vitest packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm vitest packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm tsc`
- `APP_PORT=4322 NEXTJS_PORT=3001 pnpm -F app run test-perf src/sandbox/dialog-perf/perf-chrome.ts`
- Temporarily reverted the new env-profile and unpaired-round implementation paths and confirmed the new regression tests fail.

### DIFF
--- a/app/src/sandbox/dialog-perf/index.react.tsx
+++ b/app/src/sandbox/dialog-perf/index.react.tsx
@@ -2,7 +2,7 @@ import * as Ariakit from "@ariakit/react";
 import { useState } from "react";
 import "./style.css";
 
-const items = Array.from({ length: 5000 }, (_, index) => ({
+const items = Array.from({ length: 3000 }, (_, index) => ({
   id: `item-${index + 1}`,
   title: `Item ${index + 1}`,
 }));
--- a/app/src/sandbox/dialog-perf/perf-chrome.ts
+++ b/app/src/sandbox/dialog-perf/perf-chrome.ts
@@ -4,6 +4,7 @@ import { withFramework } from "#app/test-utils/preview.ts";
 withFramework(import.meta.dirname, async ({ test }) => {
   test("open dialog", async ({ perf }) => {
     await perf.measure(({ q }) => q.button("Open dialog").click(), {
+      scriptProfile: true,
       verify: ({ q }) => expect(q.dialog("Settings")).toBeVisible(),
     });
   });
@@ -15,17 +16,6 @@ withFramework(import.meta.dirname, async ({ test }) => {
         await expect(q.dialog("Settings")).toBeVisible();
       },
       verify: ({ q }) => expect(q.dialog("Settings")).not.toBeVisible(),
-    });
-  });
-
-  // Same interaction as "open dialog", but with the script profiler enabled
-  // so the PR comment shows where the scripting time goes. Profiling adds
-  // overhead, so the unprofiled test above is the one to read for timings.
-  test("open dialog (script profile)", async ({ perf }) => {
-    await perf.measure(({ q }) => q.button("Open dialog").click(), {
-      scriptProfile: true,
-      profileLimit: 20,
-      verify: ({ q }) => expect(q.dialog("Settings")).toBeVisible(),
     });
   });
 });
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -9,6 +9,8 @@ import path from "node:path";
 import {
   computeMedianMetrics,
   formatPerfTitlePath,
+  getPerfProfileBaseLabel,
+  isPerfProfileLabel,
   median,
   mergeScriptProfiles,
   mergeSelectorProfiles,
@@ -57,6 +59,7 @@ interface ComparisonRow {
   threshold: boolean;
   significant: boolean;
   candidate: boolean;
+  profileMode: boolean;
   profileModeMismatch: boolean;
 }
 
@@ -362,14 +365,29 @@ function loadRounds(
 
 function normalizeGeneratedPerfLabel(label: string): string {
   if (!label.includes("perf-chrome.ts")) return label;
+  if (label.includes("sandbox/dialog-perf/perf-chrome.ts")) return label;
   return formatPerfTitlePath(label.split(" > "));
 }
 
 function normalizePerfResult(result: PerfResult): PerfResult {
-  const label = normalizeGeneratedPerfLabel(result.label);
-  const testTitle = normalizeGeneratedPerfLabel(result.testTitle);
-  if (label === result.label && testTitle === result.testTitle) return result;
-  return { ...result, label, testTitle };
+  const normalizedLabel = normalizeGeneratedPerfLabel(result.label);
+  const normalizedTitle = normalizeGeneratedPerfLabel(result.testTitle);
+  const label = getPerfProfileBaseLabel(normalizedLabel);
+  const testTitle = getPerfProfileBaseLabel(normalizedTitle);
+  const profileOnly =
+    result.profileOnly ||
+    result.scriptProfile ||
+    result.selectorProfile ||
+    isPerfProfileLabel(normalizedLabel) ||
+    isPerfProfileLabel(normalizedTitle);
+  if (
+    label === result.label &&
+    testTitle === result.testTitle &&
+    profileOnly === !!result.profileOnly
+  ) {
+    return result;
+  }
+  return { ...result, label, testTitle, profileOnly: profileOnly || undefined };
 }
 
 function mergeProfiles(results: PerfResult[]): PerfProfiles | undefined {
@@ -397,13 +415,19 @@ function mergeProfiles(results: PerfResult[]): PerfProfiles | undefined {
 function combineResults(results: PerfResult[]): PerfResult | undefined {
   const first = results[0];
   if (!first) return;
+  const metricResults = results.filter((result) => !result.profileOnly);
+  const metricsSource = metricResults.length > 0 ? metricResults : results;
+  const sourceFirst = metricsSource[0] ?? first;
   return {
-    ...first,
-    metrics: computeMedianMetrics(results.map((result) => result.metrics)),
-    raw: results.flatMap((result) => result.raw),
-    scriptProfile: results.some((result) => result.scriptProfile) || undefined,
+    ...sourceFirst,
+    metrics: computeMedianMetrics(
+      metricsSource.map((result) => result.metrics),
+    ),
+    raw: metricsSource.flatMap((result) => result.raw),
+    scriptProfile:
+      metricsSource.some((result) => result.scriptProfile) || undefined,
     selectorProfile:
-      results.some((result) => result.selectorProfile) || undefined,
+      metricsSource.some((result) => result.selectorProfile) || undefined,
     profiles: mergeProfiles(results),
   };
 }
@@ -732,21 +756,22 @@ function resultKey(result: PerfResult): string {
   return `${result.testFile}::${result.label}`;
 }
 
-function hasProfileData(
-  result: PerfResult | undefined,
-  profile: "script" | "selectors",
-): boolean {
-  return (result?.profiles?.[profile]?.length ?? 0) > 0;
-}
-
 function hasProfileMode(
   result: PerfResult | undefined,
   profile: "script" | "selectors",
 ): boolean {
   if (profile === "script") {
-    return !!result?.scriptProfile || hasProfileData(result, profile);
+    return !!result?.scriptProfile;
   }
-  return !!result?.selectorProfile || hasProfileData(result, profile);
+  return !!result?.selectorProfile;
+}
+
+function isProfileMode(result: PerfResult | undefined): boolean {
+  return (
+    !!result?.profileOnly ||
+    hasProfileMode(result, "script") ||
+    hasProfileMode(result, "selectors")
+  );
 }
 
 function isRegression(row: ComparisonRow, options: PerfCompareOptions) {
@@ -818,12 +843,22 @@ function compare(options: PerfCompareOptions): ComparisonSummary {
       continue;
     }
 
+    const profileMode = isProfileMode(base.result) || isProfileMode(cur.result);
+    const comparisonRoundIndices = profileMode
+      ? sharedRoundIndices
+      : sharedRoundIndices.filter((roundIndex) => {
+          const baselineRound = base.byRound.get(roundIndex);
+          const currentRound = cur.byRound.get(roundIndex);
+          if (!baselineRound || !currentRound) return false;
+          return !isProfileMode(baselineRound) && !isProfileMode(currentRound);
+        });
+    if (comparisonRoundIndices.length === 0) continue;
     const profileModeMismatch = resultProfileModeMismatches.length > 0;
-    const primary = !profileModeMismatch;
+    const primary = !profileMode;
     for (const metric of primaryMetrics) {
       const baselineValues: number[] = [];
       const roundComparisons: RoundMetricComparison[] = [];
-      for (const roundIndex of sharedRoundIndices) {
+      for (const roundIndex of comparisonRoundIndices) {
         const baselineRound = base.byRound.get(roundIndex);
         const currentRound = cur.byRound.get(roundIndex);
         if (!baselineRound || !currentRound) continue;
@@ -859,9 +894,10 @@ function compare(options: PerfCompareOptions): ComparisonSummary {
         current: curVal,
         delta,
         percent,
-        pairedRoundsCount: sharedRoundIndices.length,
+        pairedRoundsCount: comparisonRoundIndices.length,
         perRoundDeltas: roundComparisons.map((comparison) => comparison.delta),
         ...significance,
+        profileMode,
         profileModeMismatch,
       });
     }
@@ -1242,12 +1278,12 @@ function formatDetailedBreakdown({
     const testRows = rowsByKey.get(key) ?? [];
     const label = testRows[0]?.label ?? key;
     const currentResult = currentByKey.get(key)?.result;
-    const profileModeMismatch = testRows.some((row) => row.profileModeMismatch);
+    const profileMode = testRows.some((row) => row.profileMode);
     const profileLines = formatProfiles(currentResult);
-    if (profileModeMismatch && profileLines.length === 0) continue;
+    if (profileMode && profileLines.length === 0) continue;
     lines.push(`### ${label}`);
     lines.push("");
-    if (!profileModeMismatch) {
+    if (!profileMode) {
       lines.push("| Metric | Baseline | Current | Delta |");
       lines.push("|--------|----------|---------|-------|");
       for (const metric of primaryMetrics) {
@@ -1370,18 +1406,21 @@ function formatMarkdown(
 
   if (newTests.length > 0) {
     const primaryMetrics = getPrimaryMetrics(options);
+    const newTestsWithMetrics = newTests.filter(
+      ({ result }) => !isProfileMode(result),
+    );
     lines.push(`### New ${resultsName} (no baseline)`);
     lines.push("");
     if (options.node) {
-      lines.push(...formatNodeResultTables(newTests, options));
-    } else if (newTests.length > 0) {
+      lines.push(...formatNodeResultTables(newTestsWithMetrics, options));
+    } else if (newTestsWithMetrics.length > 0) {
       const headers = [
         resultName,
         ...primaryMetrics.map((metric) => getMetricLabel(metric, options)),
       ];
       lines.push(`| ${headers.join(" | ")} |`);
       lines.push(`| ${headers.map(() => "---").join(" | ")} |`);
-      for (const { result } of newTests) {
+      for (const { result } of newTestsWithMetrics) {
         const cells: string[] = [result.label];
         for (const metric of primaryMetrics) {
           cells.push(formatMetricValue(result.metrics[metric], options));
--- a/packages/ariakit-scripts/src/perf-compare.ts
+++ b/packages/ariakit-scripts/src/perf-compare.ts
@@ -8,6 +8,7 @@ import {
 import path from "node:path";
 import {
   computeMedianMetrics,
+  formatPerfTitlePath,
   median,
   mergeScriptProfiles,
   mergeSelectorProfiles,
@@ -56,7 +57,7 @@ interface ComparisonRow {
   threshold: boolean;
   significant: boolean;
   candidate: boolean;
-  profileMode: boolean;
+  profileModeMismatch: boolean;
 }
 
 interface ComparisonSummary {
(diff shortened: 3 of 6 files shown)


packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -805,8 +818,8 @@ function compare(options: PerfCompareOptions): ComparisonSummary {
       continue;
     }
 
-    const profileMode = isProfileMode(base.result) || isProfileMode(cur.result);
-    const primary = !profileMode;
+    const profileModeMismatch = resultProfileModeMismatches.length > 0;
+    const primary = !profileModeMismatch;
```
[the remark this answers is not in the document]
diegohaz: Fixed in cacdb8de8 - profile-only rows are now merged under their base label and excluded from metric aggregation, significance checks, confirmation rounds, and new-test metric rows.

app/src/sandbox/dialog-perf/index.react.tsx
```diff
@@ -2,7 +2,7 @@ import * as Ariakit from "@ariakit/react";
 import { useState } from "react";
 import "./style.css";
```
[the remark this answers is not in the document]
diegohaz: Fixed in cacdb8de8 - legacy `dialog-perf` generated labels are intentionally left unnormalized so the old 5000-item baseline is not paired with the new 3000-item workload.

app/src/sandbox/dialog-perf/perf-chrome.ts
```diff
@@ -4,6 +4,7 @@ import { withFramework } from "#app/test-utils/preview.ts";
 withFramework(import.meta.dirname, async ({ test }) => {
   test("open dialog", async ({ perf }) => {
     await perf.measure(({ q }) => q.button("Open dialog").click(), {
+      scriptProfile: true,
```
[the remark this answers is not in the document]
diegohaz: Fixed in cacdb8de8 - normal tests with `scriptProfile: true` now keep unprofiled timing samples and collect script profiles in separate diagnostic iterations whose metrics are not stored for comparison.

diegohaz:
**Fixed**
- `cacdb8de8` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523575091)): kept normal `scriptProfile: true` timing unprofiled and moved profile collection into separate diagnostic iterations.
- `cacdb8de8` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523575087)): merged profile-only rows under their base label and excluded them from metric aggregation, significance checks, confirmation rounds, and new-test metric rows.

**Invalid**
- `a94c7e6f5` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523609440)): corrected the earlier response to `r3523518151`; `dialog-perf` labels now normalize consistently and continue comparing main against the PR after sandbox changes.

app/src/sandbox/dialog-perf/index.react.tsx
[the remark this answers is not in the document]
diegohaz: Correction: this review is invalid because sandbox perf changes should still compare against main. Reverted the dialog-specific normalization exception in a94c7e6f5, so generated `dialog-perf` labels normalize and pair consistently with the PR result.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -350,12 +354,41 @@ function loadRounds(
...
+        result: normalizePerfResult(result as PerfResult),
+      });
     }
   }
   return rounds;
 }
 
+function normalizeGeneratedPerfLabel(label: string): string {
+  if (!label.includes("perf-chrome.ts")) return label;
+  return formatPerfTitlePath(label.split(" > "));
```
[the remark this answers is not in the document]
diegohaz: It's intentional.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -806,11 +843,21 @@ function compare(options: PerfCompareOptions): ComparisonSummary {
...
+    const comparisonRoundIndices = profileMode
+      ? sharedRoundIndices
+      : sharedRoundIndices.filter((roundIndex) => {
+          const baselineRound = base.byRound.get(roundIndex);
+          const currentRound = cur.byRound.get(roundIndex);
+          if (!baselineRound || !currentRound) return false;
+          return !isProfileMode(baselineRound) && !isProfileMode(currentRound);
+        });
+    if (comparisonRoundIndices.length === 0) continue;
+    const profileModeMismatch = resultProfileModeMismatches.length > 0;
```
diegohaz: `ComparisonRow.profileModeMismatch` is never read. Nothing in this file, the formatters, the tests, or the workflow consumes `row.profileModeMismatch` — the "Profile mode differs" warning is rendered solely from the top-level `profileModeMismatches` list, and significance gating uses `profileMode`/`primary`. The `resultProfileModeMismatches` array exists only to feed this dead field.

Unless a follow-up plans to use it, removing it keeps the compare loop simpler:

```diff
-    const resultProfileModeMismatches: ProfileModeMismatch[] = [];
     for (const profile of ["script", "selectors"] as const) {
       const baselineHasProfile = hasProfileMode(base.result, profile);
       const currentHasProfile = hasProfileMode(cur.result, profile);
       if (baselineHasProfile === currentHasProfile) continue;
-      const mismatch = {
+      profileModeMismatches.push({
         testFile: cur.result.testFile,
         label: cur.result.label,
         profile,
         baseline: baselineHasProfile,
         current: currentHasProfile,
-      };
-      profileModeMismatches.push(mismatch);
-      resultProfileModeMismatches.push(mismatch);
+      });
     }
```

…plus dropping `profileModeMismatch` from the `ComparisonRow` interface, the `const profileModeMismatch = …` line, and the `rows.push` entry. I verified the perf-compare tests pass without any reader of this field.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -806,11 +843,21 @@ function compare(options: PerfCompareOptions): ComparisonSummary {
...
     const profileMode = isProfileMode(base.result) || isProfileMode(cur.result);
+    const comparisonRoundIndices = profileMode
+      ? sharedRoundIndices
+      : sharedRoundIndices.filter((roundIndex) => {
+          const baselineRound = base.byRound.get(roundIndex);
+          const currentRound = cur.byRound.get(roundIndex);
+          if (!baselineRound || !currentRound) return false;
+          return !isProfileMode(baselineRound) && !isProfileMode(currentRound);
+        });
+    if (comparisonRoundIndices.length === 0) continue;
```
diegohaz: Minor: when `comparisonRoundIndices` ends up empty for a non-profile pair, this `continue` makes the test vanish from the report entirely — it produces no rows, and it isn't added to `newTests`, `removedTests`, or `unpairedTests`, so it's missing from the total count and the full breakdown with no trace.

This needs every shared round to have a profile-only side (e.g. rounds where the unprofiled variant failed but its `(script profile)` twin succeeded, or mixed old/new artifacts), so it's rare. But `unpairedTests` exists for exactly this "no usable pairs" situation, and a compare tool silently hiding a row is worse than a noisy warning:

```diff
-    if (comparisonRoundIndices.length === 0) continue;
+    if (comparisonRoundIndices.length === 0) {
+      unpairedTests.push({
+        testFile: cur.result.testFile,
+        label: cur.result.label,
+        baselineRounds: [...base.byRound.keys()].sort((a, b) => a - b),
+        currentRounds: [...cur.byRound.keys()].sort((a, b) => a - b),
+      });
+      continue;
+    }
```

packages/ariakit-scripts/src/perf.ts
```diff
@@ -1310,10 +1344,15 @@ export async function createPerfMeasure(
     options.selectorProfile ??
     (isTruthyEnv("PERF_SELECTOR_PROFILE") ||
       isTruthyEnv("PERF_CSS_SELECTOR_PROFILE"));
+  const testTitle = formatPerfTitlePath(testInfo.titlePath);
+  const baseLabel = label ?? testTitle;
+  const profileOnly = isPerfProfileLabel(baseLabel);
+  const timingScriptProfile = profileOnly && scriptProfile;
```
diegohaz: Two docs are now stale/missing for this behavior split:

1. The `scriptProfile`/`selectorProfile` option JSDoc (lines 125–128) still says "Adds overhead to measured metrics." After this change that's only true for profile-only tests; for normal tests the whole point is that measured metrics stay unprofiled and the profiler runs in extra diagnostic iterations (which add wall-clock, not metric skew).

2. The ` (script profile)` / ` (selector profile)` label suffix is now a load-bearing convention: it silently switches a measurement into profile-only mode — different sampling defaults (3 iterations, 0 warmup), in-band profiling, `profileOnly` in the stored result, and exclusion from timing comparisons. Nothing in `PerfMeasureOptions`, `formatPerfTitlePath`, or `isPerfProfileLabel` documents that naming a test this way changes how it runs, so a future test author can trip this by title choice alone.

Suggest updating the option JSDoc to describe both modes and adding a short JSDoc to `getPerfProfileBaseLabel`/`isPerfProfileLabel` stating the convention and its effects.

packages/ariakit-scripts/src/perf.ts
```diff
@@ -1347,36 +1386,69 @@ export async function createPerfMeasure(
...
+
+  await runIterations({
+    iterationCount: iterations,
+    warmupCount: warmup,
+    scriptProfile: timingScriptProfile,
+    selectorProfile: timingSelectorProfile,
+    collectMetrics: true,
+  });
+
+  if (!profileOnly && (scriptProfile || selectorProfile)) {
```
diegohaz: Suspected workflow regression for env-driven profiling. `scriptProfile` here can come from `PERF_SCRIPT_PROFILE=1`, and before this change that env var turned every test into a quick profiled run: script-profile sampling only (3 iterations, 0 warmup). Now a normal test under the env var runs the full unprofiled timing pass (default 10 + 1 warmup locally) **plus** these 3 profiled iterations — roughly 4–5× the wall-clock per test, and each iteration is a fresh context + full page load, so this is real minutes on the larger sandboxes.

The unprofiled timing samples are strictly better data, so this is the right default for the `scriptProfile: true` option path (CI). But for the local "just show me where the time goes" env workflow the extra timing pass buys nothing. If that quick-diagnostic behavior is worth keeping, you could preserve it by only running the separate timing pass when profiling was requested explicitly via options:

```ts
const envProfileRun = options.scriptProfile == null && isTruthyEnv("PERF_SCRIPT_PROFILE");
const timingScriptProfile = (profileOnly || envProfileRun) && scriptProfile;
```

(and analogously for `selectorProfile`), which keeps env runs in-band with the old sampling and flags. If the new cost is intentional, ignore this — but it's worth a deliberate decision since nothing in the PR mentions the env path.

packages/ariakit-scripts/src/perf.ts
```diff
@@ -352,6 +353,39 @@ export function getPerfSamplingOptions({
...
+  const titleParts: string[] = [];
+  for (const title of titlePath) {
+    if (!title) continue;
+    const pathParts = /\.tsx?$/.test(title) ? title.split(/[\\/]/) : [title];
+    const sandboxIndex = pathParts.lastIndexOf("sandbox");
+    const visibleParts =
+      sandboxIndex >= 0 ? pathParts.slice(sandboxIndex + 1) : pathParts;
+    for (const part of visibleParts) {
+      if (!part) continue;
+      if (part === "sandbox") continue;
```
diegohaz: This branch is unreachable: `visibleParts` only contains parts after the last `"sandbox"` occurrence (via `lastIndexOf` + `slice`), and for non-path titles a bare `"sandbox"` title yields `pathParts = ["sandbox"]`, `sandboxIndex = 0`, so `visibleParts` is empty. No `part` inside the loop can ever equal `"sandbox"`, so the line can be dropped:

```diff
     for (const part of visibleParts) {
       if (!part) continue;
-      if (part === "sandbox") continue;
       if (part === "perf-chrome.ts") continue;
       titleParts.push(part);
     }
```

diegohaz (commented):
## Review summary

**Type:** Initial review from this workflow (no prior marked review). **Baseline:** merge base `8f508b9e0` → head `a94c7e6f5`, all 6 changed files plus the surrounding perf tooling (`perf.yml`, other sandbox `perf-chrome.ts` files, Playwright perf config).

### Validation performed

- `pnpm install` + `pnpm vitest run packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts` — 69/69 passed at head.
- `pnpm tsc` — clean.
- Did not re-run the browser perf test (`test-perf` on dialog-perf); relying on the PR's stated verification for that.

### Prior review threads verified against head

- **Codex: "Keep the primary dialog timing unprofiled"** — fix confirmed. `createPerfMeasure` gates in-band profiling on `profileOnly` (label suffix convention); normal `scriptProfile: true` tests keep unprofiled timing samples and collect profiles in separate diagnostic iterations with `collectMetrics: false`.
- **Codex: "Avoid flagging diagnostic profile-only timing rows"** — fix confirmed structurally: `significant` and `candidate` both require `primary` (= not profile mode), profile-only rows merge under their base label in `combineResults`/`aggregateByKey` with their metrics excluded, and per-round profile-only results are filtered out of `comparisonRoundIndices` for primary pairs.
- **Codex: workload 5000→3000 pairing** — the normalization exception was reverted in `a94c7e6f5` as the author stated; `dialog-perf` labels now normalize on both sides and pair with main. Note the next perf run will report a large dialog-perf "improvement" that is the workload change, not a code change — this is the author's stated intent.
- **CodeRabbit: redundant `else if (newTests.length > 0)`** — no longer applies at head; the branch now checks `newTestsWithMetrics.length`.
- **CodeRabbit: hardcoded `"perf-chrome.ts"` duplicated between `formatPerfTitlePath` and `normalizeGeneratedPerfLabel`** — still true at head; thread remains open and I have nothing to add beyond it.

### Findings (5 inline comments, none blocking)

1. **Dead code** — `ComparisonRow.profileModeMismatch` (and the array feeding it) is written but never read anywhere.
2. **Silent drop** — an empty `comparisonRoundIndices` makes a paired test vanish from the report instead of landing in `unpairedTests` (rare, mixed-round edge).
3. **Stale docs** — the `scriptProfile`/`selectorProfile` JSDoc still says profiling skews measured metrics, and the ` (script profile)` label-suffix convention that flips a test into profile-only mode is undocumented.
4. **Env workflow cost** — `PERF_SCRIPT_PROFILE=1` runs now pay the full unprofiled timing pass plus profile iterations (~4–5× wall-clock vs. the old quick-diagnostic behavior); flagged as a judgment call since the PR only discusses the option path.
5. **Unreachable branch** — `if (part === "sandbox") continue;` in `formatPerfTitlePath` can never fire after the `lastIndexOf`/`slice`.

### Not fully validated

- Behavior of a real CI comparison against main's old-format artifacts (old labels/flags) is verified by code reading and the "pairs legacy generated labels" unit test, not by an actual perf workflow run.
- The combined open-dialog test's fit within the 120s perf timeout (9 fresh-context loads at 3000 items) — the author's local `test-perf` run covers this.

packages/ariakit-scripts/src/perf.ts
diegohaz: Fixed in 381aee427 - removed the unreachable `part === "sandbox"` branch from `formatPerfTitlePath`.

packages/ariakit-scripts/src/perf.ts
diegohaz: Fixed in 381aee427 - updated the `scriptProfile`/`selectorProfile` option docs for the explicit-option, env-driven, and profile-only modes, and documented the profile-only label suffix convention on the profile label helpers.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 381aee427 - removed the unused `ComparisonRow.profileModeMismatch` field and the local mismatch array that only fed that dead field; profile-mode warnings still use the top-level `profileModeMismatches` list.

packages/ariakit-scripts/src/perf.ts
diegohaz: Fixed in 381aee427 - preserved the env-driven quick profiling path by resolving `PERF_SCRIPT_PROFILE`, `PERF_SELECTOR_PROFILE`, and `PERF_CSS_SELECTOR_PROFILE` as measured profile iterations. Explicit profile options on regular labels still keep timing metrics unprofiled and collect profiles in separate diagnostic iterations.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 381aee427 - when all shared rounds are filtered out as profile-mode rounds, the test is now reported through `unpairedTests` instead of disappearing from the report. Added regression coverage for that crossed profile-only round case.

diegohaz:
**Fixed**
- `381aee427` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523661117)): removed the unused per-row `profileModeMismatch` plumbing while keeping the top-level profile mismatch warning path.
- `381aee427` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523661121)): reports tests with no comparable non-profile paired rounds as unpaired instead of omitting them.
- `381aee427` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523661113)): documented the explicit-option, env-driven, and profile-only profiling modes and the profile-label suffix convention.
- `381aee427` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523661118)): preserved env-driven quick profiling as measured profile iterations while keeping explicit profile options diagnostic.
- `381aee427` ([reply](https://github.com/ariakit/ariakit/pull/6647#discussion_r3523661112)): removed the unreachable `sandbox` branch in perf title formatting.

diegohaz (commented):
## Re-review summary

**Type:** Re-review. **Baseline:** prior marked review at `a94c7e6f5` (review_id `pr-6647-a94c7e6f5272d762c0e66bf65e4ff823781861fa-20260704T175703Z`); incremental diff `a94c7e6f5..381aee427` (one commit), with surrounding context re-checked.

### Validation performed

- `pnpm vitest run packages/ariakit-scripts/src/perf.test.ts packages/ariakit-scripts/src/perf-compare.test.ts` — 73/73 passed at head (4 new tests since last review).
- `pnpm tsc` — clean.

### Prior findings — all five fix claims verified in code, not just from replies

1. **Dead `ComparisonRow.profileModeMismatch`** — fixed. Field and the `resultProfileModeMismatches` array removed; mismatch warnings still flow through the top-level list.
2. **Silent drop on empty `comparisonRoundIndices`** — fixed. The pair is now routed through `unpairedTests` via the new `createUnpairedTest` helper, the unpaired copy was reworded to "no comparable paired round" (which also reads correctly when both sides list the same round indices), and the crossed profile-only-round case has a dedicated regression test asserting both markdown and summary JSON.
3. **Stale profiling docs** — fixed. `scriptProfile`/`selectorProfile` option JSDoc now describes all three modes (explicit option, env-driven, profile-only label), and `getPerfProfileBaseLabel`/`isPerfProfileLabel` document the label-suffix convention and its effects.
4. **Env-driven quick-profiling regression** — fixed. `getPerfProfileMode` resolves env-sourced profiling as measured (in-band) iterations with profile sampling, preserving the old fast diagnostic path, while explicit options on regular labels keep timing unprofiled with separate diagnostic iterations. I verified the matrix: explicit `true` → diagnostic; env → measured; explicit `false` overrides env; profile-only labels → measured; and env-driven results still store the `scriptProfile` flag, so `perf-compare` correctly treats them as profile-only rows. Unit tests cover each branch including the `PERF_CSS_SELECTOR_PROFILE` alias.
5. **Unreachable `sandbox` branch in `formatPerfTitlePath`** — fixed, branch removed.

### Other threads

- Codex's follow-up ("Avoid pairing dialog results across changed workloads") was closed by the author as intentional; this matches the design confirmed in my initial review — sandbox workload changes deliberately compare against main, so the next perf run's large dialog-perf delta is the 5000→3000 item change, not a code regression.
- No new bot findings were posted against `381aee427`.

### New findings

None. The refactored `getPerfProfileMode` is a clean extraction (single object param, documented, unit-tested), and I found no correctness, performance, or simplification issues in the new commit.

Issue #4075, opened 2024-08-26
Dialog slow to open


bengry:
### Current behavior

The Dialog component takes quite a long time to open when heavy component(s) is rendered _outside_ of the dialog.

I did a small comparison of three simple dialogs:
1. [Ariakit](https://ariakit.org/components/dialog)
2. [Radix](https://www.radix-ui.com/primitives/docs/components/dialog)
3. [MUI](https://mui.com/material-ui/react-dialog)

All three have very simple components within the dialog itself. There's a considerable difference in time to open each dialog, and also to close (although less noticeable there, a difference still exists).

MUI takes around ~60-70ms to open, Radix takes ~280ms, and Ariakit takes ~370ms to open. This is quite noticeable for users.

### Steps to reproduce the bug

1. Open https://stackblitz.com/edit/n2wvvx?file=dialog%2Findex.tsx
2. Click any of the three buttons to see the behavior with each library.
3. See `alert`s.
    * For some reason, Ariakit renders twice. It's the only one of the three to show this behavior. I didn't track down why that happens though.

Also notice that even if `<HeavyComponent />` is replaced with `<HeavyComponentMemo />` (same thing, but with `React.memo` around it), there's no significant difference in time to render of either library.

### Expected behavior

Ariakit's dialog should open and close quickly/within a reasonable amount of time, even when `unmountOnHide=true` (this is the same behavior that MUI's dialog has by default (via `[isActive?=false]`).

diegohaz:
Thanks for reporting. Ariakit has a [`preventBodyScroll`](https://ariakit.org/reference/dialog#preventbodyscroll) logic that's more robust to ensure nested dialogs work properly. That's likely one of the reasons. We might be able to optimize this if there's only one dialog.

Anyway, [as recommended by React](https://react.dev/reference/react/memo#specifying-a-custom-comparison-function:~:text=When%20you%20do%20performance%20measurements%2C%20make%20sure%20that%20React%20is%20running%20in%20the%20production%20mode.), it's better to perform these measurements with React running in production.

> * For some reason, Ariakit renders twice. It's the only one of the three to show this behavior. I didn't track down why that happens though.

This is because of `StrictMode`. It doesn't happen in production.

bengry:
[...]

PR #6648, opened 2026-07-04
Paint modal dialogs before disabling the outside tree

diegohaz:
## Motivation

Opening a modal `Dialog` is slow on pages with a lot of content (#4075). Chrome traces on the dialog-perf sandbox (5000 cards) show why: when the dialog opens, `markAndDisableTreeOutside` sets `inert` on the top-level elements around it, and the browser propagates inertness through computed styles, invalidating the entire outside tree. The resulting ~68ms style recalculation runs before the dialog's first paint, so the open interaction blocks on work whose only purpose is to disable content the user can no longer see uncovered. This is also what made `setViewportHeight` top the perf-report script profiles: its `visualViewport.height` read was the first forced style flush after the `inert` writes, so the profiler attributed the whole bill to it.

This revives the direction of #6609 (closed), adapted to how `main` has evolved since:

- The scroll-lock half of #6609 is already shipped: the `scrollbar-gutter` lock (#6634) removed the `--scrollbar-width` full-document invalidation on every browser that supports the property, and skips all width compensation when there is no scrollbar width to compensate.
- That also changes #6609's "share the pre-paint pass" gating: with the gutter lock, removing a visible classic scrollbar no longer forces a full-document pass, so deferring the inert writes pays off on classic-scrollbar systems too (e.g. Windows). Only the no-gutter-support fallback (Safari < 18.2) still writes the inherited `--scrollbar-width` property that forces a pre-paint full-document recalc.

## Solution

Keep the semantics; move the cost off the input → paint critical path:

- `markAndDisableTreeOutside` now marks synchronously and returns the disabling work separately (`{ disableTreeOutside, restoreTreeOutside }`). Marks are plain JavaScript properties with no style impact, and they're what the outside-click and Escape logic runs on, so that logic is unaffected.
- The dialog defers `disableTreeOutside` to right after the open frame paints (double `requestAnimationFrame` through the element's owner window, canceled if the dialog closes first). The ~68ms recalculation still happens — it's the platform price of inertness, same as native `showModal()` — but no longer blocks the first paint.
- **Sync exception 1 — replacements never leave gaps.** When a disabling replaces one that was active until a cleanup in the same task restored it (the same dialog re-running because a nested dialog opened, or a sibling dialog taking over), it re-applies synchronously: the restore already invalidated the outside tree, so the sync re-apply merges into the same recalc instead of leaving the tree interactive for a frame. Tracked by a module-level flag cleared in a microtask, covered by a MutationObserver regression test that asserts the outside tree is inert at every mutation checkpoint during a sibling takeover.
- **Sync exception 2 — legacy scrollbar fallback.** On browsers without `scrollbar-gutter` support that are about to remove a visible scrollbar, the scroll lock's `--scrollbar-width` write already forces a pre-paint full-document recalc, so the inert writes apply synchronously to share that pass instead of adding a second one a frame later (this is #6609's probe, narrowed by a `supportsScrollbarGutter` check so modern browsers with classic scrollbars still get the deferral). The `usePreventBodyScroll` call moved below the tree-disabling effect so the probe measures the scrollbar before the lock removes it.
- The close path is unchanged: restores stay synchronous because focus-on-hide needs the outside tree re-enabled to move focus back to the disclosure.

## Measured (dialog-perf, 5000 cards, dev server)

Chrome trace of the open interaction:

| | main | this PR |
|---|---|---|
| click task duration | 246ms | 186ms |
| first paint after click | ~268ms | ~197ms |
| pre-paint forced style recalc | 68.6ms (whole outside tree) | 0.9ms |
| the inert recalc | pre-paint, blamed on `setViewportHeight` | post-paint, unforced, in the frame pipeline |

Interleaved A/B benchmark (2 rounds per side, 20 samples each, medians) of the "open dialog" perf test:

| Metric | main | this PR |
|---|---|---|
| INP | 688ms | 616ms (−10.5%) |
| Scripting | 549ms | 547ms (−0.5%) |
| Rendering | 127ms | 126ms (−1.2%) |
| Total | 677ms | 673ms (−0.6%) |

The INP drop matches the deferred recalculation time; total work stays the same because the recalc moves past the paint, it doesn't disappear (RecalcStyleCount goes from 8 to 9 since the deferred pass no longer merges with the open-frame flush).

## Tradeoffs

- The outside tree becomes inert one frame (~16ms) after the dialog appears when the deferred path applies. No real user interaction can land in that window, the backdrop already covers the page, and the outside-click/Escape logic runs off the synchronous marks. Native `showModal()` applies inertness synchronously; this intentionally trades that for paint latency (same tradeoff #6609 chose).
- If the dialog closes before the deferred frame fires, the callback is canceled and the restore only undoes the marks.
- `markAndDisableTreeOutside` (reachable via the `@ariakit/react-components/dialog/utils/disable-tree` sub-path, not re-exported by `@ariakit/react`, undocumented) changes its return type from a restore function to `{ disableTreeOutside, restoreTreeOutside }`. Treated as an internal util.

## Verification

- happy-dom reports `scrollbar-gutter` support, so the entire integration suite exercises the deferred path by default (matching what all modern browsers do); the existing `CSS.supports`-mocked fallback test exercises the sync path.
- New tests: sibling-takeover inert continuity (MutationObserver at every checkpoint) and sibling-takeover scroll-lock continuity; `markAndDisableTreeOutside` unit test updated for the split API.
- Full `test-react` suite 848/848 (twice), `test-react18` 840 passed + 8 pre-existing skips, app Chrome dialog browser tests 23/23, `pnpm tsc`, `pnpm lint`.

### DIFF
--- a/.changeset/200-dialog-open-paint.md
+++ b/.changeset/200-dialog-open-paint.md
@@ -1,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Faster modal `Dialog` opening
+
+Opening a modal [`Dialog`](https://ariakit.com/reference/dialog), or a component built on it such as [`Popover`](https://ariakit.com/reference/popover) or [`Menu`](https://ariakit.com/reference/menu) with the [`modal`](https://ariakit.com/reference/popover#modal) prop, no longer waits for the page behind it to be disabled before painting.
+
+Previously, making the element tree outside the dialog inert invalidated the styles of the entire page before the dialog's first paint, delaying the open in proportion to the page size. Now the inert step runs right after the dialog is first painted, so the page outside stays interactive and exposed to assistive technology for about one frame longer. The outside-click and Escape logic is unaffected: it relies on element marks that still apply synchronously. On browsers without CSS `scrollbar-gutter` support, when the scroll lock must remove a visible scrollbar, everything still applies before the first paint, matching the previous behavior.
--- a/examples/dialog-nested-various/test.ts
+++ b/examples/dialog-nested-various/test.ts
@@ -1,4 +1,4 @@
-import { click, press, q } from "@ariakit/test";
+import { click, dispatch, press, q, sleep } from "@ariakit/test";
 import { expect, test, vi } from "vitest";
 
 function getBackdrop(name: string) {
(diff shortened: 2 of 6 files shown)


diegohaz:
Will revisit later.

PR #6662, opened 2026-07-05
Compact perf report breakdown

diegohaz:
## Motivation

The perf report full breakdown currently spends vertical space on one metric table per test, followed by generic `Script profile` headings. Large perf runs become harder to scan because the same four metric headings are repeated for every test.

The compact table also needs links that work in GitHub PR comments. GitHub rewrites custom HTML anchors such as `<a id="script-profile-foo"></a>` to `user-content-*` IDs when rendering comments, so table links must target that rendered fragment.

## Solution

Render the non-node full breakdown as one comparison table, with linked test cells when a script profile exists. Script profile sections now use the same display title as the table row and expose stable generated anchors, while selector profiles remain nested under the test heading.

```md
| Test | Scripting | Rendering | INP | Total |
| [composite-perf > react > mount composite](#user-content-script-profile-composite-perf-react-mount-composite) | 100ms → 105ms (+5%) | ... |
| composite-perf > react > move across items | 420ms → 404ms (-4%) | ... |
```

If a performance result has a custom `perf.measure({ label })` that differs from the Playwright test title, the compact breakdown includes both as `testTitle > label` so multiple measurements from one test remain distinguishable. The significant-change summary table and unconfirmed-candidate table use the same display title so prominent regression rows also keep their test context.

GitHub-flavored Markdown tables do not render single-cell group rows as colspan headings in PR comments; they render a first-column cell plus empty cells for the remaining columns. Because of that, this keeps full test titles in the table instead of simulating groups.

This also trims unnecessary `.0` from millisecond values throughout the perf markdown output, so `10.0ms` becomes `10ms` while non-integer values keep their decimal precision.

Additional tests cover title-based profile headings, GitHub-compatible profile links, custom-label disambiguation across compact breakdown, summary, and candidate rows, new-test profile anchors, duplicate anchor disambiguation, markdown-sensitive link text, and the updated compact millisecond formatting.

## Verification

- `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm exec oxlint --disable-nested-config packages/ariakit-scripts/src/perf-compare.ts packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm exec oxfmt --check packages/ariakit-scripts/src/perf-compare.ts packages/ariakit-scripts/src/perf-compare.test.ts`
- `pnpm tsc`
- Red check: temporarily removed the summary/candidate display-title source fix and confirmed the focused perf comparison test failed on the new assertions, then restored the fix and reran the suite successfully.

### DIFF
--- a/packages/ariakit-scripts/src/perf-compare.test.ts
+++ b/packages/ariakit-scripts/src/perf-compare.test.ts
@@ -128,6 +128,18 @@ function createResultWithMetrics(
   };
 }
 
+function createResultWithTitle(
+  label: string,
+  testTitle: string,
+  metrics: PerfMetrics,
+  profiles?: PerfProfiles,
+): PerfResult {
+  return {
+    ...createResultWithMetrics(label, metrics, profiles),
+    testTitle,
+  };
+}
+
 function createScriptProfileEntry(
   functionName: string,
   selfTime: number,
@@ -302,7 +314,7 @@ test("keeps single-run comparison behavior", () => {
 
   const markdown = runCompare(dir);
 
-  expect(markdown).toContain("100.0ms → 120.0ms (+20%) :warning:");
+  expect(markdown).toContain("100ms → 120ms (+20%) :warning:");
   expect(markdown).not.toContain("Aggregated across");
 });
 
@@ -319,7 +331,7 @@ test("pairs legacy generated labels with normalized labels", () => {
   const markdown = runCompare(dir);
   const summary = readComparisonSummary(dir);
 
-  expect(markdown).toContain("100.0ms \u2192 120.0ms (+20%) :warning:");
+  expect(markdown).toContain("100ms \u2192 120ms (+20%) :warning:");
   expect(markdown).not.toContain("### New tests");
   expect(markdown).not.toContain("### Removed tests");
   expect(summary.rows[0]?.label).toBe("example > react > example");
@@ -515,9 +527,9 @@ test("reports overlapping same-direction rounds as unconfirmed candidates", () =
     markdown.indexOf("<details>"),
   );
   expect(markdown).toContain(
-    "| raw samples | Scripting | 120.0ms | 140.0ms | +20.0ms (+17%) | low | rounds 2/2, raw 0/2, pairs 60% |",
+    "| raw samples | Scripting | 120ms | 140ms | +20ms (+17%) | low | rounds 2/2, raw 0/2, pairs 60% |",
   );
-  expect(markdown).toContain("120.0ms | 140.0ms | +20.0ms (+17%)");
+  expect(markdown).toContain("120ms | 140ms | +20ms (+17%)");
   // Candidates are reported in the unconfirmed changes section, not repeated
   // in the detailed breakdown diagnostics.
   expect(markdown).not.toContain("Unflagged threshold-sized changes");
@@ -539,7 +551,7 @@ test("grades candidates by their displayed pairs percent", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain(
-    "| raw samples | Scripting | 55.0ms | 85.0ms | +30.0ms (+55%) | medium | rounds 2/2, raw 0/2, pairs 70% |",
+    "| raw samples | Scripting | 55ms | 85ms | +30ms (+55%) | medium | rounds 2/2, raw 0/2, pairs 70% |",
   );
 });
 
@@ -553,7 +565,7 @@ test("flags same-direction rounds with separated raw samples", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain(":warning:");
-  expect(markdown).toContain("100.0ms → 160.0ms (+60%) :warning:");
+  expect(markdown).toContain("100ms → 160ms (+60%) :warning:");
   expect(markdown).not.toContain("Unconfirmed changes");
 });
 
@@ -567,9 +579,9 @@ test("requires raw sample support in each required round", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No confirmed performance changes detected.");
-  expect(markdown).toContain("110.0ms | 150.0ms | +40.0ms (+36%)");
+  expect(markdown).toContain("110ms | 150ms | +40ms (+36%)");
   expect(markdown).toContain(
-    "| raw samples | Scripting | 110.0ms | 150.0ms | +40.0ms (+36%) | medium | rounds 2/2, raw 1/2, pairs 80% |",
+    "| raw samples | Scripting | 110ms | 150ms | +40ms (+36%) | medium | rounds 2/2, raw 1/2, pairs 80% |",
   );
   expect(markdown).not.toMatch(/% :warning:/);
 });
@@ -628,7 +640,7 @@ test("keeps zero-baseline candidates visible under the cap", () => {
   // Zero-baseline rows have no percent, so they must rank ahead of
   // percent-sorted rows instead of falling behind them and off the cap.
   expect(markdown).toContain(
-    "| from zero | Scripting | 0.0ms | 200.0ms | +200.0ms | low | rounds 2/2, raw 0/2, pairs 60% |",
+    "| from zero | Scripting | 0ms | 200ms | +200ms | low | rounds 2/2, raw 0/2, pairs 60% |",
   );
   expect(markdown.indexOf("| from zero | Scripting |")).toBeLessThan(
     markdown.indexOf("| c1 | Scripting |"),
@@ -649,7 +661,7 @@ test("reports INP-only changes as unconfirmed candidates", () => {
 
   expect(markdown).toContain("No confirmed performance changes detected.");
   expect(markdown).toContain(
-    "| inp test | INP | 650.0ms | 400.0ms | -250.0ms (-38%) | low | rounds 2/2, raw 0/2, pairs 60% |",
+    "| inp test | INP | 650ms | 400ms | -250ms (-38%) | low | rounds 2/2, raw 0/2, pairs 60% |",
   );
   expect(markdown).not.toMatch(/% :rocket:/);
 });
@@ -664,7 +676,7 @@ test("flags confirmed INP regressions", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("| Test | Scripting | Rendering | INP | Total |");
-  expect(markdown).toContain("110.0ms → 220.0ms (+100%) :warning:");
+  expect(markdown).toContain("110ms → 220ms (+100%) :warning:");
   expect(markdown).not.toContain("Unconfirmed changes");
 });
 
@@ -677,7 +689,7 @@ test("flags confirmed INP improvements", () => {
 
   const markdown = runCompare(dir);
 
-  expect(markdown).toContain("220.0ms → 110.0ms (-50%) :rocket:");
+  expect(markdown).toContain("220ms → 110ms (-50%) :rocket:");
   expect(markdown).not.toContain("Unconfirmed changes");
 });
 
@@ -736,10 +748,10 @@ test("lists confirmation files for significant and candidate changes", () => {
   );
   // Significant and unconfirmed changes are reported side by side, without
   // icons on the unconfirmed rows.
-  expect(markdown).toContain("100.0ms → 160.0ms (+60%) :warning:");
+  expect(markdown).toContain("100ms → 160ms (+60%) :warning:");
   expect(markdown).toContain("#### Unconfirmed changes");
   expect(markdown).toContain(
-    "| candidate | Scripting | 120.0ms | 140.0ms | +20.0ms (+17%) | low | rounds 2/2, raw 0/2, pairs 60% |",
+    "| candidate | Scripting | 120ms | 140ms | +20ms (+17%) | low | rounds 2/2, raw 0/2, pairs 60% |",
   );
 });
 
@@ -769,8 +781,8 @@ test("aggregates displayed values from shared rounds", () => {
 
   const markdown = runCompare(dir);
 
-  expect(markdown).toContain("200.0ms → 160.0ms (-20%) :rocket:");
-  expect(markdown).not.toContain("125.0ms");
+  expect(markdown).toContain("200ms → 160ms (-20%) :rocket:");
+  expect(markdown).not.toContain("125ms");
   expect(markdown).toContain("Aggregated across 2 interleaved rounds");
   expect(markdown).toContain(
     "paired median delta exceeds the threshold, rounds agree on direction and raw samples support it",
@@ -804,7 +816,7 @@ test("merges sharded round files with shard-suffixed names", () => {
   expect(markdown).toContain("Aggregated across 2 interleaved rounds");
   expect(markdown).toContain("shard one test");
   expect(markdown).toContain("shard two test");
-  expect(markdown).toContain("100.0ms → 130.0ms (+30%) :warning:");
+  expect(markdown).toContain("100ms → 130ms (+30%) :warning:");
 });
 
 test("does not overstate mixed paired round counts", () => {
@@ -848,7 +860,7 @@ test("requires both rounds to agree in two-round comparisons", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No significant performance changes detected.");
-  expect(markdown).toContain("100.0ms | 115.0ms | +15.0ms (+15%)");
+  expect(markdown).toContain("100ms → 115ms (+15%)");
   expect(markdown).not.toMatch(/% :warning:/);
 });
 
@@ -877,7 +889,7 @@ test("keeps zero-baseline paired rounds in the agreement count", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No significant performance changes detected.");
-  expect(markdown).toContain("50.0ms | 65.0ms | +15.0ms (+30%)");
+  expect(markdown).toContain("50ms → 65ms (+30%)");
   expect(markdown).not.toMatch(/% :warning:/);
 });
 
@@ -946,7 +958,7 @@ test("reports tests with no paired rounds separately", () => {
   );
   expect(markdown).toContain("### Unpaired tests");
   expect(markdown).toContain("| example > react > example | 1 | 2 |");
-  expect(markdown).not.toContain("0.0ms | 0.0ms");
+  expect(markdown).not.toContain("0ms | 0ms");
 });
 
 test("surfaces unpaired tests outside the details block", () => {
@@ -1002,7 +1014,7 @@ test("does not flag percentage-only changes below the absolute floor", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No significant performance changes detected.");
-  expect(markdown).toContain("20.0ms | 24.0ms | +4.0ms (+20%)");
+  expect(markdown).toContain("20ms → 24ms (+20%)");
   expect(markdown).not.toMatch(/% :warning:/);
 });
 
@@ -1038,7 +1050,7 @@ test("ignores legacy rendering sub-metrics in older baselines", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No significant performance changes detected.");
-  expect(markdown).toContain("| Rendering | 20.0ms | 20.0ms | +0.0ms (+0%) |");
+  expect(markdown).toContain("20ms → 20ms (+0%)");
   expect(markdown).not.toContain("Style recalc");
   expect(markdown).not.toContain("Painting");
 });
@@ -1053,7 +1065,7 @@ test("aggregates worker shards within the same round", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No confirmed performance changes detected.");
-  expect(markdown).toContain("100.0ms | 150.0ms | +50.0ms (+50%)");
+  expect(markdown).toContain("100ms | 150ms | +50ms (+50%)");
   expect(markdown).toContain("rounds 1/1, raw 0/1, pairs 50%");
   expect(markdown).not.toMatch(/% :warning:/);
 });
@@ -1142,7 +1154,7 @@ test("renders script profile function names as linked code", () => {
     "[`Escaped`](https://github.com/ariakit/ariakit/blob/abc123/app/src/sandbox/dialog%5C%5Bperf%5D%7Ccase.tsx#L56)",
   );
   expect(markdown).toContain(
-    "| [`Dialog`](https://github.com/ariakit/ariakit/blob/abc123/packages/ariakit-react-components/src/dialog/dialog.tsx#L12) | 8.0ms | 9.0ms | 2 |",
+    "| [`Dialog`](https://github.com/ariakit/ariakit/blob/abc123/packages/ariakit-react-components/src/dialog/dialog.tsx#L12) | 8ms | 9ms | 2 |",
   );
   expect(markdown).not.toContain("| Function | Self | Total | Hits | Source |");
   expect(markdown).not.toContain("dialog.tsx:12:3");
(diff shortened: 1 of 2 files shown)


packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1320,151 @@ interface FormatDetailedBreakdownParams {
...
+}
+
+function createScriptProfileAnchor(
+  heading: string,
+  usedAnchors: Map<string, number>,
+) {
+  const baseAnchor = `script-profile-${formatProfileAnchorSlug(heading)}`;
+  const count = usedAnchors.get(baseAnchor) ?? 0;
+  usedAnchors.set(baseAnchor, count + 1);
+  return count === 0 ? baseAnchor : `${baseAnchor}-${count + 1}`;
```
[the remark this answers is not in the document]
diegohaz: I don't think we will have duplicate sandboxes, will we?

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -692,6 +710,25 @@ function escapeTableCell(value: string): string {
     .trim();
 }
 
+function escapeMarkdownLinkText(value: string): string {
+  return escapeTableCell(value).replace(/([[\]])/g, "\\$1");
+}
```
diegohaz: `escapeMarkdownLinkText` reimplements the existing `escapeLinkText` (declared just below, line 757): both backslash-escape first, then escape `[`, `]`, and `|`, collapse `\r\n\t` runs to spaces, and trim. The operation orders differ, but the affected character sets are disjoint, so the two functions return identical output for every input. Reusing the existing helper keeps one definition of markdown link-text escaping and avoids the two drifting apart:

```diff
-function escapeMarkdownLinkText(value: string): string {
-  return escapeTableCell(value).replace(/([[\]])/g, "\\$1");
-}
-
```

```diff
 function formatDetailedTestCell(label: string, anchor?: string) {
   if (!anchor) return escapeTableCell(label);
-  return `[${escapeMarkdownLinkText(label)}](#${anchor})`;
+  return `[${escapeLinkText(label)}](#${anchor})`;
 }
```

I validated this locally: `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts` (52/52, including the new escaping test) and `tsc -b packages/ariakit-scripts/tsconfig.json` pass with the swap. It also removes the code the CodeQL alert on this line points at (the alert is a false positive since `escapeTableCell` escapes backslashes before the bracket pass, but the flagged helper disappears).

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1320,151 @@ interface FormatDetailedBreakdownParams {
...
-          `| ${getMetricLabel(metric, options)} | ${formatMetricValue(row.baseline, options)} | ${formatMetricValue(row.current, options)} | ${deltaStr}${icon} |`,
-        );
-      }
-      lines.push("");
-      lines.push(...formatUnflaggedDiagnostics(testRows, options));
-    }
+    const label =
+      currentResult != null
+        ? getProfileHeading(currentResult)
+        : (testRows[0]?.testTitle ?? testRows[0]?.label ?? key);
```
diegohaz: The `label` fallback branch is dead code: when `currentResult` is nullish, `formatProfileSection` returns `[]` before ever reading its `heading` parameter, so the `(testRows[0]?.testTitle ?? testRows[0]?.label ?? key)` value can never be rendered. Beyond that, both live call sites (this loop and the new-tests loop at line 1614) pass `getProfileHeading(result)`, so the `heading` parameter can be dropped entirely and computed inside `formatProfileSection` from the `result` it already receives. This loop then reduces to:

```ts
for (const key of keys) {
  const profileLines = formatProfileSection(
    currentByKey.get(key)?.result,
    scriptProfileAnchors.get(key),
  );
  if (profileLines.length === 0) continue;
  lines.push(...profileLines);
}
```

with `formatProfileSection` pushing `` `#### ${getProfileHeading(result)}` ``, and the `testRows` lookup goes away along with the heading/table-label duplication. I applied this locally and the full test file (52/52) plus `tsc -b packages/ariakit-scripts/tsconfig.json` pass.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1320,151 @@ interface FormatDetailedBreakdownParams {
...
+    const result = currentByKey.get(key)?.result;
+    if (!result) continue;
+    if (!hasScriptProfile(result)) continue;
+    anchors.set(
+      key,
+      createScriptProfileAnchor(getProfileHeading(result), usedAnchors),
+    );
+  }
+  return anchors;
+}
```
diegohaz: `createScriptProfileAnchors` and `createResultScriptProfileAnchors` run the same loop (skip missing or unprofiled results, then assign `createScriptProfileAnchor(getProfileHeading(result), usedAnchors)` under the aggregated result's key); the only difference is whether the caller starts from keys or from results. Since `currentByKey.get(key)?.key === key` for every compared key, this keys-based variant can be expressed through the results-based one at the call site in `formatMarkdown`:

```ts
const scriptProfileAnchors = createResultScriptProfileAnchors(
  allKeys.flatMap((key) => currentByKey.get(key) ?? []),
  usedScriptProfileAnchors,
);
```

which preserves iteration order, and therefore duplicate-anchor suffix assignment, while deleting this function. Verified locally: 52/52 tests and `tsc` pass with the change.

packages/ariakit-scripts/src/perf-compare.test.ts
```diff
@@ -848,7 +860,7 @@ test("requires both rounds to agree in two-round comparisons", () => {
   const markdown = runCompare(dir);
 
   expect(markdown).toContain("No significant performance changes detected.");
-  expect(markdown).toContain("100.0ms | 115.0ms | +15.0ms (+15%)");
+  expect(markdown).toContain("100ms → 115ms (+15%)");
```
diegohaz, line 916: The new "for `<label>`" wording in the unflagged-diagnostics line (`formatUnflaggedDiagnostics`, perf-compare.ts line 1221) only has negative coverage (`not.toContain` at line 535), so a regression in the label or its HTML escaping would pass the suite unnoticed. This test already produces that diagnostics line (threshold-sized change with rounds 1/2 agreement), so it's a natural home for a positive assertion. I confirmed the exact rendered output by running this scenario's round files through the CLI:

```suggestion
  expect(markdown).toContain("100ms → 115ms (+15%)");
  expect(markdown).toContain(
    "Unflagged threshold-sized changes for example &gt; react &gt; example:",
  );
```

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -692,6 +710,25 @@
     .trim();
 }
 
+function escapeMarkdownLinkText(value: string): string {
+  return escapeTableCell(value).replace(/([[\]])/g, "\\$1");
```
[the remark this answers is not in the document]
diegohaz: This alert is a false positive: `escapeTableCell` already escapes backslashes (`.replace(/\\/g, "\\\\")` as its first step) before this bracket pass runs, so the composed escaping is complete and correctly ordered. That said, the flagged helper duplicates the existing `escapeLinkText` (line 757), which performs the same escaping with the backslash step visible in one function — I left a separate comment suggesting reusing it, which would also make this alert go away.

packages/ariakit-scripts/src/perf-compare.ts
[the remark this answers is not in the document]
diegohaz: I looked at when colliding headings can actually occur. Results sharing a `testFile::label` key merge before anchors are assigned, so two anchor entries can only share a base anchor when different test files produce titles with the same slug. Titles embed the sandbox directory path (`combobox-perf > react > open combobox`), so that would require either identical sandbox paths (impossible on one filesystem) or a punctuation-only difference that collapses in the slug (for example a `foo-bar` sandbox versus a nested `foo/bar` one) — and even then, the counter keys on the slugged base anchor, so those cases get suffixed correctly. The only remaining gap is the compound case reported here, where a third title naturally slugifies to an already-suffixed anchor, which I don't see being reachable with real data. So I agree this isn't blocking; if you want it airtight anyway, reserving the suffixed anchor in `usedAnchors` when it's assigned would close it in one extra line.

diegohaz (commented):
Initial review of the full PR (baseline: merge-base `de77ba080` → head `da6edd97e`). I reviewed both changed files line by line plus the surrounding context: the `label`/`testTitle` producers in `perf.ts`, real `perf-chrome.ts` usage in `app/src/sandbox/`, and the `perf.yml` comment-posting path (it only strips the `## Performance` heading, so the restructured markdown is safe there).

**Validation performed**

- Ran `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts` on the PR head (52/52 pass), plus `tsc -b packages/ariakit-scripts/tsconfig.json` and `oxlint` on both files.
- Ran the `perf-compare` CLI on synthetic multi-round fixtures to inspect the full generated report: the single breakdown table, linked test cells, `<a id>` targets, new-tests links, duplicate-anchor suffixes, labeled unflagged-diagnostics lines, and the trimmed `ms` formatting all render as described.
- Verified through GitHub's markdown API that `<a id="script-profile-…">` survives comment sanitization as `id="user-content-script-profile-…"`, which GitHub's frontend resolves for `#script-profile-…` fragment links, so the anchor scheme works in PR comments.
- Prototyped each simplification suggested in the inline comments and re-ran the full test file and type check before reverting.

**Findings**

No correctness, compatibility, or performance issues found. The inline comments are three validated behavior-preserving simplifications (reuse `escapeLinkText` instead of the duplicate `escapeMarkdownLinkText`, drop the dead heading fallback and the always-derivable `heading` parameter of `formatProfileSection`, and fold `createScriptProfileAnchors` into `createResultScriptProfileAnchors`) plus one test-coverage suggestion for the new labeled diagnostics wording. I also replied on the CodeQL thread (false positive, and the flagged helper disappears with the dedup) and on the anchor-collision thread (collision is unreachable with real data; optional one-line hardening).

**Observation, non-blocking**: the top summary and candidate tables still name rows by `label` while the new breakdown and new-tests tables use `testTitle`. Today these are identical because no perf test passes a custom `label`, but if one ever does, the same test will be named differently across tables within one report. Worth keeping in mind rather than changing now.

**Not fully validated**: I did not run a real perf CI round, and I did not click the anchors in a live PR comment (the sanitizer behavior was verified via the markdown API instead).

packages/ariakit-scripts/src/perf-compare.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 99a250510 - `formatDetailedTestCell` now reuses `escapeLinkText`, so the duplicate helper that triggered the alert is gone and link text uses the existing backslash/bracket/pipe escaping path.

packages/ariakit-scripts/src/perf-compare.ts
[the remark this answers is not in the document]
diegohaz: Fixed in 99a250510 - generated script profile anchors are now reserved as actual IDs in a `Set`, and the test covers the compound collision where a natural slug matches an already suffixed anchor.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 99a250510 - `formatDetailedTestCell` now calls `escapeLinkText` directly, removing the duplicate `escapeMarkdownLinkText` helper.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 99a250510 - `formatProfileSection` now derives its own heading from `getProfileHeading(result)`, so the dead fallback and extra heading parameter are gone.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 99a250510 - compared-test anchors now go through `createResultScriptProfileAnchors`, preserving order while removing the duplicate key-based helper.

packages/ariakit-scripts/src/perf-compare.test.ts
diegohaz: Fixed in 99a250510 - added positive coverage for the escaped unflagged diagnostics label in the existing two-round disagreement scenario.

diegohaz:
**Fixed**

- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875240)): Reused the existing `escapeLinkText` path for compact-table profile links, removing the duplicated helper that triggered the CodeQL alert.
- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875265)): Reserved generated profile anchor IDs with a `Set` and covered the compound suffixed-anchor collision.
- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875297)): Swapped `formatDetailedTestCell` to call `escapeLinkText` directly.
- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875316)): Simplified profile heading handling so `formatProfileSection` derives the heading from the result.
- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875344)): Removed the duplicate key-based script profile anchor builder.
- 99a250510 ([reply](https://github.com/ariakit/ariakit/pull/6662#discussion_r3525875375)): Added positive coverage for the escaped unflagged-diagnostics label.

diegohaz:
Fixed in c8567d60f.

- Full-breakdown links now target GitHub-rendered `#user-content-script-profile-*` anchor IDs, so clicking a profiled test row jumps to its script profile section.
- Rows with shared prefixes are grouped in the detailed table, for example `composite-perf > react` with row labels like `mount composite` and `move across items`.
- I also confirmed with GitHub markdown rendering that one-cell table group rows render as regular rows with empty remaining cells, so the report emits an explicit group row with blank metric cells.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1315,195 @@ interface FormatDetailedBreakdownParams {
...
+    const label = getDetailedComparisonLabel(key, rows);
+    const { group } = splitDetailedTestTitle(label);
+    if (!group) continue;
+
+    counts.set(group, (counts.get(group) ?? 0) + 1);
+  }
+  return new Set(
+    [...counts].filter(([, count]) => count > 1).map(([group]) => group),
+  );
+}
```
diegohaz: This counting pass duplicates the render loop's visibility rules (`rows.length === 0`, `profileMode`) and its label/split derivation. If someone later adds a skip condition to one loop but not the other, the group counts silently diverge from the rendered rows, producing bold group headers with a single row under them (or headers for rows that were skipped). Collecting the visible entries once and feeding both the counter and the render loop removes that drift hazard and the duplication:

```ts
interface DetailedComparisonEntry {
  key: string;
  rows: ComparisonRow[];
  label: string;
  title: DetailedTestTitle;
}

function getDetailedComparisonEntries(
  rowsByKey: Map<string, ComparisonRow[]>,
  keys: string[],
): DetailedComparisonEntry[] {
  return keys.flatMap((key) => {
    const rows = rowsByKey.get(key) ?? [];
    if (rows.length === 0) return [];
    if (rows.some((row) => row.profileMode)) return [];
    const label = getDetailedComparisonLabel(key, rows);
    return [{ key, rows, label, title: splitDetailedTestTitle(label) }];
  });
}

function getDetailedComparisonGroups(entries: DetailedComparisonEntry[]) {
  const counts = new Map<string, number>();
  for (const { title } of entries) {
    if (!title.group) continue;
    counts.set(title.group, (counts.get(title.group) ?? 0) + 1);
  }
  return new Set(
    [...counts].filter(([, count]) => count > 1).map(([group]) => group),
  );
}
```

with `formatDetailedComparisonTable` iterating `for (const { key, rows: testRows, label, title } of entries)`. I applied this locally on this head and the full test file (53/53) plus `tsc -b packages/ariakit-scripts/tsconfig.json` pass.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1315,126 @@ interface FormatDetailedBreakdownParams {
...
+  ];
+  const tableRows: string[] = [];
+  const diagnostics: string[] = [];
+
+  for (const key of keys) {
+    const testRows = rowsByKey.get(key) ?? [];
+    if (testRows.length === 0) continue;
+    if (testRows.some((row) => row.profileMode)) continue;
+
+    const label = testRows[0]?.testTitle ?? testRows[0]?.label ?? key;
```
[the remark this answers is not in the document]
diegohaz: I verified this on the current head (`c8567d60f`) and the failure mode is real. Feeding the CLI two results with the same `testTitle` but distinct labels (one Playwright test calling `perf.measure` twice with `label: "open with mouse"` / `label: "open with keyboard"`) renders two visually identical rows and two identical profile headings:

```md
| [open combobox](#user-content-script-profile-combobox-perf-react-open-combobox) | 100ms → 130ms (+30%) :warning: | …
| [open combobox](#user-content-script-profile-combobox-perf-react-open-combobox-2) | 100ms → 108ms (+8%) | …
```

both followed by `#### combobox-perf > react > open combobox` sections — only the anchor suffix distinguishes them. Two points of context for deciding whether to handle it now:

- No perf test in the repo passes a custom `label` today, so no current report can hit this.
- But custom labels are the only viable way to write a multi-measure test: two unlabeled `perf.measure` calls in one test share the same `testFile::label` key and get merged into a single comparison by `aggregateByKey`, silently averaging two different interactions. So if a multi-measure test is ever added, it will use labels and land exactly here.

If you want to close it preemptively, the small fix is to prefer `label` over `testTitle` when they differ (in `getDetailedComparisonLabel` and `getProfileHeading`), which also matches what the summary and candidate tables already display; a bare custom label has no `" > "` so it simply renders ungrouped. Deferring until a multi-measure test actually exists also seems fine — the ambiguity is cosmetic and the data stays correct.

diegohaz (commented):
Re-review covering the two commits since my last review (`da6edd97e` → `c8567d60f`: `99a250510` and `c8567d60f`), plus the surrounding code they touch.

**Prior findings — all verified fixed in code, not just from the replies**

- Duplicate `escapeMarkdownLinkText` removed; `formatDetailedTestCell` uses `escapeLinkText`.
- `formatProfileSection` derives its heading via `getProfileHeading(result)`; the dead fallback and extra parameter are gone.
- Compared-test anchors go through `createResultScriptProfileAnchors`; the key-based duplicate helper is deleted.
- Positive coverage added for the escaped unflagged-diagnostics label (I re-ran the suite: 53/53 pass).
- The anchor-collision hardening is stronger than my suggestion: anchors are reserved as concrete IDs in a `Set` with a probe loop, and the new test covers the compound case (`same title test 2` correctly gets `…-2-2`).

**New code in `c8567d60f`**

- The `#user-content-` href prefix is the right call: I confirmed via GitHub's markdown API that the sanitizer rewrites the `<a id>` to `user-content-…` and leaves the link href untouched, so the links now resolve through native fragment navigation with no dependency on GitHub's client-side hash resolution (which my initial review over-credited for PR conversation pages — the commit corrects that assumption at the right layer). Trade-off: the links only resolve where GitHub's `user-content-` rewrite happens, so raw `comparison.md` artifacts viewed locally won't scroll; that seems acceptable for a PR-comment-first report.
- The group-header rows render correctly (verified by running the CLI on synthetic fixtures); diagnostics still use the full title, so they stay unambiguous. One inline comment: the group-counting pass duplicates the render loop's visibility rules and label derivation, which is a drift hazard — I validated a single-pass entries collection on this head (53/53 + `tsc`).
- Edge case, cosmetic, not worth blocking: group headers are emitted per contiguous run, so a test whose first result appears only in a later round sorts after all round-1 keys and repeats its group's bold header. The current behavior degrades gracefully (extra header, correct association), which is the right failure mode.

**Thread follow-ups**: I replied on the new custom-labels thread with a verified repro on this head — the issue is real but unreachable today (no perf test passes a custom `label`), with a small label-preferring fix if you want it closed preemptively.

**Validation**: `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts` (53/53), `tsc -b packages/ariakit-scripts/tsconfig.json`, `oxlint` on both files, CLI runs on synthetic multi-round fixtures (grouped table, `user-content-` links, duplicate-title anchors, custom-label repro), and a GitHub markdown API sanitization check.

diegohaz:
Checked GitHub rendering for the table-group idea. A single-cell table row does not become a colspan-style heading; GitHub renders it as the first `<td>` plus empty cells for the remaining columns.

Because of that, I reverted the table grouping in 9cd658ca1 and kept the GitHub-compatible script-profile anchor fix.

packages/ariakit-scripts/src/perf-compare.ts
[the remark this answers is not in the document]
diegohaz: Fixed in be964e5c4 - compact breakdown rows and script profile headings now include the custom perf label as `testTitle > label` when it differs from the Playwright title, with regression coverage for multiple measures sharing one title.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Resolved in 9cd658ca1 - table grouping was removed after verifying GitHub table rows do not render as colspan-style headings, so the duplicated group-count/render-loop logic no longer exists.

diegohaz:
**Fixed**

- be964e5c4 (https://github.com/ariakit/ariakit/pull/6662#discussion_r3526030803): Custom `perf.measure` labels now appear in compact breakdown rows and profile headings as `testTitle > label` when they differ from the Playwright title, with regression coverage for shared-title measurements.

**Already resolved**

- 9cd658ca1 (https://github.com/ariakit/ariakit/pull/6662#discussion_r3526030804): The grouping-count/render-loop concern was resolved by removing table grouping after verifying GitHub tables do not render single-cell group rows as colspan headings.

packages/ariakit-scripts/src/perf-compare.ts
```diff
@@ -1270,49 +1315,139 @@ interface FormatDetailedBreakdownParams {
...
+}
+
+function getResultDisplayTitle({
+  label,
+  testTitle,
+}: Pick<PerfResult, "label" | "testTitle">) {
+  if (!testTitle) return label;
+  if (!label || label === testTitle) return testTitle;
+  return `${testTitle} > ${label}`;
+}
```
diegohaz, line 1344: Consider also routing the two remaining Chrome-facing tables through this helper: the significant-changes summary table (line 1109, `testRows[0]?.label ?? key`) and the unconfirmed-candidates table (line 1180, `escapeTableCell(row.label)`). Both still render the bare `label`, so for the custom-label scenario this commit addresses, those tables would show just `open with mouse` with no test context — and two files reusing the same custom label would produce indistinguishable rows in the tables that surface regressions most prominently. For default labels the display title is string-identical to `label`, so the swap changes nothing in today's reports; the node paths are unaffected (the node table renders through `formatNodeComparisonTables`, and node rows can never be candidates since they skip the raw-sample gate).

```ts
const label = testRows[0] ? getResultDisplayTitle(testRows[0]) : key;
const cells: string[] = [escapeTableCell(label)];
```

```ts
escapeTableCell(getResultDisplayTitle(row)),
```

I validated both swaps on this head: 53/53 tests, `tsc -b packages/ariakit-scripts/tsconfig.json`, and `oxlint` pass (the summary-table cell also gains the escaping the other tables already apply).

diegohaz (commented):
Re-review covering the two commits since my last review (`c8567d60f` → `be964e5c4`: `9cd658ca1` and `be964e5c4`).

**Prior threads — both resolutions verified in code**

- `9cd658ca1` removes the table grouping entirely, which also removes the duplicated group-count/render-loop logic my previous comment flagged. Removal is complete: `splitDetailedTestTitle`, `getDetailedComparisonGroups`, and `getDetailedComparisonLabel` are all gone, and the breakdown is net simpler than before the grouping was added.
- `be964e5c4` resolves the custom-labels thread with `getResultDisplayTitle` (`testTitle > label` when they differ). I re-ran my multi-measure CLI repro on this head: the two labeled measures now render as distinct rows, distinct headings, and distinct human-readable anchors (`…-open-with-mouse` / `…-open-with-keyboard`), so the earlier ambiguity is gone. Node output is unaffected — node rendering never goes through `getResultDisplayTitle` (node tables use `label` directly, and node results carry no script profiles).

**New finding**: one inline comment suggesting the significant-changes and candidate tables also use `getResultDisplayTitle` — they still render the bare `label`, which loses test context in exactly the custom-label scenario this round adds support for. Zero output change for default labels; validated locally.

**Minor observation, not a defect**: an explicit `label` that duplicates the test title's last segment renders as `… > close combobox > close combobox`. That only happens when someone passes a redundant label equal to the test name (default labels equal the full title and take the `label === testTitle` early return), so it reads as working-as-intended.

**Validation**: `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts` (53/53 on this head), `tsc -b packages/ariakit-scripts/tsconfig.json`, `oxlint` on both files, and CLI runs on the multi-measure custom-label fixtures.

packages/ariakit-scripts/src/perf-compare.ts
diegohaz: Fixed in 2d701cd76 - the significant summary table and unconfirmed candidate table now both use getResultDisplayTitle, so custom perf labels keep their test context in the prominent Chrome-facing tables. Added red-checked coverage for summary and candidate rows.

diegohaz:
**Fixed**

- 2d701cd76 (https://github.com/ariakit/ariakit/pull/6662#discussion_r3526076510): Summary and unconfirmed candidate tables now use full display titles for custom-labeled measurements, with regression coverage for significant and candidate rows.

diegohaz (commented):
Re-review covering the one commit since my last review (`be964e5c4` → `2d701cd76`).

**Prior thread verified in code**: `2d701cd76` routes the significant-changes summary table and the unconfirmed-candidates table through `getResultDisplayTitle`, matching the change I validated last round (including the summary cell gaining `escapeTableCell`, which the other tables already applied). The new coverage is solid: the candidate-table test builds a realistic unconfirmed candidate from raw samples with a custom label, and both tests pin the negative case (`not.toContain("| open with mouse |")`), so a regression back to bare labels would fail.

**No new findings.** The diff is small and exactly scoped; I re-ran `pnpm test packages/ariakit-scripts/src/perf-compare.test.ts` (54/54 on this head), `tsc -b packages/ariakit-scripts/tsconfig.json`, and `oxlint` on both files — all pass.

**Optional, fine to leave as is**: the remaining `label`-rendered surfaces are low-prominence diagnostics — the unpaired-tests bullet list and table, the removed-tests list, and the profile-mode-mismatch table. They'd show a bare custom label without test context, but they're investigation aids rather than primary comparison rows, and no perf test uses custom labels today. I don't think they need to change; noting for completeness since the theme came up three times.

With this, every thread on the PR is addressed and verified: the CodeQL alert (helper removed), both Codex findings (anchor collisions hardened with reserved IDs; custom labels shown with test context end to end), and all of this skill's prior comments. The PR is in good shape.

PR #6664, opened 2026-07-06
Fix Combobox Korean IME composition with autoSelect

### LINKED ISSUE #6663: Combobox: autoSelect breaks Korean IME composition
### Current behavior

`<Combobox autoSelect />` breaks Korean IME composition mid-syllable: the composing text is silently discarded and the last keystroke lands as a raw jamo.

Typing `사과` (사 = ㅅ + ㅏ, 과 = ㄱ + ㅘ) with a Korean 2벌식 IME produces `사ㅏ`.

- `t k` → input shows `사` (syllable 1 committed).
- `r h` → input shows `사고` (syllable 2 composing, still valid).
- The next keystroke `k` should transition the composition from `고` to `과`. Instead, the composition is aborted: `고` disappears and only `ㅏ` (a raw jamo) is inserted. Final value: `사ㅏ`.

Reproduces only when `autoSelect` is enabled. Without `autoSelect`, Korean input works correctly.

Repro: https://stackblitz.com/edit/vitejs-vite-gor8d8hw?file=src%2FApp.tsx 

1. Switch the OS keyboard to Korean (macOS: **한국어 - 2벌식**).
2. Focus the input and type `사과` (keys: `t k r h k`).
3. Observe: input shows `사고` after the fourth keystroke, then jumps to `사ㅏ` after the fifth. The consumer's `keyword` state (shown in the hint text) reflects the broken value.

https://github.com/user-attachments/assets/6d9ed9cc-9194-458a-a254-d6d21279bc8d

### Expected behavior

Typing `사과` produces the string `사과` in the input, with `autoSelect` continuing to highlight the first matching item — same as when `autoSelect` is off, or on non-CJK keyboards.

### Workaround

Until the library fix in https://github.com/ariakit/ariakit/pull/6664 is released, prevent the composition-end auto-select refresh on the affected combobox:

```tsx
<Ariakit.Combobox
  autoSelect
  onCompositionEnd={(event) => {
    event.preventDefault();
  }}
/>
```

This keeps the input focused during the Korean IME boundary, but it also means `autoSelect` will not immediately re-run when a composition session ends. It will run again on the next normal value/update path.

### Possible solutions

The `Combobox`'s auto-select `useUpdateEffect` fires on `storeValue` / `items` changes and eventually calls `store.move()`, which moves DOM focus. That effect doesn't check `composingRef`. Sibling handlers do check `isComposing` (e.g. `Composite.onKeyDown`), so this looks like a missing guard rather than a design choice.

diegohaz:
## Motivation

`Combobox` with `autoSelect` moves DOM focus to the auto-selected option between Korean IME composition steps. With macOS Korean 2-set, typing the physical keys `t k r h k` should produce `사과`, but the native input path can produce `사ㅏ`.

Playwright/CDP composition does not reproduce the final native macOS value loss, but it does expose the same inter-syllable focus move that causes the native IME to drop the composed `고` syllable. The same repro fails on `@ariakit/react@0.4.31` and the current `0.4.32` source, so this is not a `0.4.32`-only regression.

Fixes #6663

## Solution

The library now tracks composition start immediately, disables auto-select while composition is active, and defers the composition-end auto-select refresh to the next animation frame. If another composition starts before that frame runs, the pending refresh is canceled/skipped so DOM focus stays in the combobox between Korean IME syllables.

The sandbox regression test replays the observed Korean 2-set composition sequence and asserts that DOM focus does not move to an option between `compositionend("사")` and the next `compositionstart`. Existing composition behavior is still covered by the combobox-group tests, now waiting for the deferred composition-end refresh.

## Workaround

Until the library fix is released, prevent the composition-end auto-select refresh on the affected combobox:

```tsx
<Ariakit.Combobox
  autoSelect
  onCompositionEnd={(event) => {
    // TODO: Remove once https://github.com/ariakit/ariakit/issues/6663
    // is fixed.
    event.preventDefault();
  }}
/>
```

This keeps the input focused during the Korean IME boundary, but it also means `autoSelect` will not immediately re-run when a composition session ends. It will run again on the next normal value/update path.

### DIFF
--- a/app/src/sandbox/combobox-ime-6663/test-chrome.ts
+++ b/app/src/sandbox/combobox-ime-6663/test-chrome.ts
@@ -1,0 +1,94 @@
+import type { CDPSession, Page } from "@playwright/test";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+interface ImeTraceEvent {
+  data?: string;
+  role?: string | null;
+  type: string;
+}
+
+declare global {
+  interface Window {
+    __comboboxImeEvents?: ImeTraceEvent[];
+  }
+}
+
+async function traceImeEvents(page: Page) {
+  await page.evaluate(() => {
+    const events: ImeTraceEvent[] = [];
+    window.__comboboxImeEvents = events;
+    for (const type of ["compositionstart", "compositionend", "focusin"]) {
+      document.addEventListener(
+        type,
+        (event) => {
+          const target = event.target as HTMLElement | null;
+          events.push({
+            type,
+            data: "data" in event ? event.data : undefined,
+            role: target?.getAttribute("role"),
+          });
+        },
+        true,
+      );
+    }
+  });
+}
+
+async function setComposition(cdp: CDPSession, text: string) {
+  await cdp.send("Input.imeSetComposition", {
+    text,
+    selectionStart: text.length,
+    selectionEnd: text.length,
+  });
+}
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6663
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("keeps DOM focus in the combobox between Korean IME syllables", async ({
+    page,
+    q,
+  }) => {
+    const combobox = q.combobox("Fruit");
+    const cdp = await page.context().newCDPSession(page);
+
+    await combobox.focus();
+    await traceImeEvents(page);
+
+    // This CDP sequence is derived from a real macOS Korean 2-set trace for
+    // typing t k r h k, which should produce 사과. CDP preserves the final
+    // value, but it still exposes the harmful focus move that makes the native
+    // macOS IME drop the composed 고 syllable and produce 사ㅏ.
+    await setComposition(cdp, "ㅅ");
+    await setComposition(cdp, "사");
+    await setComposition(cdp, "삭");
+    await setComposition(cdp, "사");
+    await cdp.send("Input.insertText", { text: "사" });
+    await setComposition(cdp, "고");
+
+    const focusMove = await page.evaluate(() => {
+      const events = window.__comboboxImeEvents ?? [];
+      const end = events.findIndex(
+        (event) => event.type === "compositionend" && event.data === "사",
+      );
+      const start = events.findIndex(
+        (event, index) => index > end && event.type === "compositionstart",
+      );
+      const movedToOption =
+        end >= 0 &&
+        start > end &&
+        events
+          .slice(end + 1, start)
+          .some((event) => event.type === "focusin" && event.role === "option");
+      return { end, events, movedToOption, start };
+    });
+    test.expect(focusMove.end).toBeGreaterThanOrEqual(0);
+    test.expect(focusMove.start).toBeGreaterThan(focusMove.end);
+    test.expect(focusMove.movedToOption).toBe(false);
+
+    await setComposition(cdp, "과");
+    await cdp.send("Input.insertText", { text: "과" });
+
+    await test.expect(combobox).toHaveValue("사과");
+    await test.expect(q.text("Current keyword: 사과")).toBeVisible();
+  });
+});
--- a/app/src/sandbox/combobox-ime-6663/test-chrome.ts
+++ b/app/src/sandbox/combobox-ime-6663/test-chrome.ts
@@ -22,9 +22,11 @@ async function traceImeEvents(page: Page) {
         type,
         (event) => {
           const target = event.target as HTMLElement | null;
+          const data =
+            event instanceof CompositionEvent ? event.data : undefined;
           events.push({
             type,
-            data: "data" in event ? event.data : undefined,
+            data,
             role: target?.getAttribute("role"),
           });
         },
--- a/app/src/sandbox/combobox-ime-6663/test-chrome.ts
+++ b/app/src/sandbox/combobox-ime-6663/test-chrome.ts
@@ -64,8 +64,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await setComposition(cdp, "사");
     await setComposition(cdp, "삭");
     await setComposition(cdp, "사");
-    await cdp.send("Input.insertText", { text: "사" });
-    await setComposition(cdp, "고");
+    const insertText = cdp.send("Input.insertText", { text: "사" });
+    const nextSyllable = setComposition(cdp, "고");
+    await Promise.all([insertText, nextSyllable]);
 
     const focusMove = await page.evaluate(() => {
       const events = window.__comboboxImeEvents ?? [];
--- a/packages/ariakit-react-components/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox.tsx
@@ -148,6 +148,14 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     const [valueUpdated, forceValueUpdate] = useForceUpdate();
     const canAutoSelectRef = useRef(false);
     const composingRef = useRef(false);
+    const compositionEndFrameRef = useRef<number | null>(null);
+
+    const cancelCompositionEndFrame = () => {
+      const frame = compositionEndFrameRef.current;
+      if (frame == null) return;
+      cancelAnimationFrame(frame);
+      compositionEndFrameRef.current = null;
+    };
 
     // We can only allow auto select when the combobox focus is handled via the
     // aria-activedescendant attribute. Othwerwise, the focus would move to the
@@ -383,6 +391,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       const canAutoSelect = canAutoSelectRef.current;
       if (!store) return;
       if (!open) return;
+      if (composingRef.current) return;
       if (!canAutoSelect && (!resetValueOnSelect || userScrolledRef.current))
         return;
       const { baseElement, contentElement, activeId } = store.getState();
@@ -555,6 +564,17 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       }
     });
 
+    useEffect(() => cancelCompositionEndFrame, []);
+
+    const onCompositionStartProp = props.onCompositionStart;
+
+    const onCompositionStart = useEvent((event: CompositionEvent<HTMLType>) => {
+      cancelCompositionEndFrame();
+      canAutoSelectRef.current = false;
+      composingRef.current = true;
+      onCompositionStartProp?.(event);
+    });
+
     const onCompositionEndProp = props.onCompositionEnd;
 
     // When dealing with composition text (for example, when the user is typing
@@ -568,7 +588,12 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       onCompositionEndProp?.(event);
       if (event.defaultPrevented) return;
       if (!autoSelect) return;
-      forceValueUpdate();
+      cancelCompositionEndFrame();
+      compositionEndFrameRef.current = requestAnimationFrame(() => {
+        compositionEndFrameRef.current = null;
+        if (composingRef.current) return;
+        forceValueUpdate();
+      });
     });
 
     const onMouseDownProp = props.onMouseDown;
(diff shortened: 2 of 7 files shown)


app/src/sandbox/combobox-ime-6663/test-chrome.ts
```diff
@@ -0,0 +1,96 @@
...
+  await cdp.send("Input.imeSetComposition", {
+    text,
+    selectionStart: text.length,
+    selectionEnd: text.length,
+  });
+}
+
+// Reproduces https://github.com/ariakit/ariakit/issues/6663
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("keeps DOM focus in the combobox between Korean IME syllables", async ({
```
diegohaz, line 49: Consider adding a fast happy-dom `test.ts` alongside this Chrome test. The convention here (see the sibling IME sandbox `toolbar-container-ime-6579`, which pairs `test.ts` with `test-chrome.ts`) is to add a happy-dom duplicate whenever the behavior reliably reproduces there — and this one does: `dispatch` only flushes microtasks between events, so `compositionend` → `compositionstart` can be dispatched with no animation frame in between, exactly like the single-keystroke Korean IME sequence.

I validated this locally: the test below passes on this branch (both `pnpm test-react` and `pnpm test-react18`, stable across repeated runs) and fails on `main`'s `combobox.tsx` at the `expect(q.option("사과")).not.toHaveFocus()` assertion — precisely the regression this PR fixes. A unit-level test runs in every React suite on every PR, so it would catch a regression much earlier and cheaper than the Chrome-project-only CDP test.

```ts
// app/src/sandbox/combobox-ime-6663/test.ts
import { dispatch, q, sleep, type } from "@ariakit/test";

// Reproduces https://github.com/ariakit/ariakit/issues/6663
test("keeps focus in the combobox between IME composition sessions", async () => {
  const combobox = q.combobox("Fruit");
  combobox.focus();
  await type("사", combobox, { isComposing: true });
  expect(combobox).toHaveFocus();
  // A Korean IME commits the previous syllable and immediately starts
  // composing the next one within a single keystroke, so no animation frames
  // pass between compositionend and the next compositionstart.
  await dispatch.compositionEnd(combobox);
  await dispatch.compositionStart(combobox);
  await sleep();
  // Focus must stay in the combobox while the next syllable is composing.
  expect(q.option("사과")).not.toHaveFocus();
  expect(combobox).toHaveFocus();
  await type("고", combobox, { isComposing: true });
  // Once the composition session truly ends, the auto-select refresh runs on
  // the next frame.
  await dispatch.compositionEnd(combobox);
  await sleep();
  expect(q.option("사과")).toHaveFocus();
});
```

One gotcha from writing it: a `type()` call between the composition boundary and the assertion masks the bug on unfixed code, because its `keydown` resets `canAutoSelectRef` before the deferred effect flushes (in the real IME sequence the keydown precedes `compositionend`). The ordering above — assert first, then continue typing — fails reliably on the unfixed code.

app/src/sandbox/combobox-ime-6663/test-chrome.ts
```diff
@@ -0,0 +1,96 @@
...
+    // This CDP sequence is derived from a real macOS Korean 2-set trace for
+    // typing t k r h k, which should produce 사과. CDP preserves the final
+    // value, but it still exposes the harmful focus move that makes the native
+    // macOS IME drop the composed 고 syllable and produce 사ㅏ.
+    await setComposition(cdp, "ㅅ");
+    await setComposition(cdp, "사");
+    await setComposition(cdp, "삭");
+    await setComposition(cdp, "사");
+    await cdp.send("Input.insertText", { text: "사" });
+    await setComposition(cdp, "고");
```
diegohaz: Suspected (not reproduced) flake window: `Input.insertText` fires `compositionend`, which now schedules the auto-select refresh on the next animation frame, and the `compositionstart` that cancels it only arrives after another CDP round trip plus a test-runner turnaround. In a real IME both events are dispatched within a single keystroke's input processing, so no frame can interleave — but across two awaited CDP sends a frame boundary can land in the gap. If it does, the deferred refresh runs, focus moves to the option, and `movedToOption` fails exactly like the regression this test guards.

I could not trigger it locally (11/11 passes, including `--repeat-each=10` with parallel workers), so this is a hardening suggestion rather than an observed failure. Issuing the two sends back-to-back and awaiting them together shrinks the window to effectively zero — commands on a single CDP session are dispatched in order, so the sequence is preserved:

```ts
const insertText = cdp.send("Input.insertText", { text: "사" });
const nextSyllable = setComposition(cdp, "고");
await Promise.all([insertText, nextSyllable]);
```

The final `insertText` + value assertions at the end of the test don't need this, since nothing there depends on post-commit focus timing.

diegohaz (commented):
**Initial review** of head 2cd58f9 (diff baseline: merge base 1b7eb8ace, the current `main` tip). Scope: the full diff plus surrounding context — the `useCombobox` auto-select/composition flow, the composite virtual-focus move path, `@ariakit/test` `dispatch`/`sleep`/`type` composition semantics, sibling composition tests (`combobox-cancel`, `toolbar-container-ime-6579`), the changeset, and sandbox/preview conventions.

**Assessment.** The fix looks correct and minimal, and I believe the layered design is genuinely necessary rather than belt-and-braces:

- The rAF deferral is needed on top of the `composingRef` effect guard: React's passive-effect flush can interleave between `compositionend` and the next `compositionstart` (that interleaving is what severs the native IME session), and at that instant `composingRef` is `false`, so the guard alone cannot block the move. Deferring the refresh past the next `compositionstart` closes that window, while the guard still covers later flushes (e.g. a frame that fires before cancellation lands).
- The documented `onCompositionEnd` `preventDefault()` workaround remains functional on this branch (the refresh scheduling is behind the `defaultPrevented` check, and composition-state bookkeeping stays unconditional, matching the existing handler pattern).
- Edge cases checked: unmount cancels the pending frame; popover-closed/blurred before the frame degrades to a no-op re-render via the existing `!open`/`hasFocus` effect guards; refs are per-instance; no SSR surface. Cost is one rAF per composition end — negligible, and perf CI shows no significant changes.
- `combobox-cancel`'s composition test already had `sleep()` after `compositionEnd` on `main`, so the two `combobox-group` updates are the only test-timing changes needed; I found no other tests relying on the previously synchronous refresh.

**Validation performed locally** (isolated worktree at the PR head):

- `pnpm test-react` and `pnpm test-react18` for `combobox-group` + `combobox-cancel`: pass.
- The sandbox Chrome test against a local astro dev server: 11/11 passes (including `--repeat-each=10` with parallel workers), and it fails deterministically (both retries) when `packages/ariakit-react-components/src/combobox/combobox.tsx` is reverted to `main` — so the regression coverage is real in both directions.
- Prototyped a happy-dom repro of the same scenario: it fails on the unfixed code at the mid-composition focus assertion and passes on this branch across both React suites — see the inline comment proposing it as `test.ts`.

**Findings** (2, both inline on the sandbox test): a suggested fast happy-dom `test.ts` duplicate (validated, code included), and a suspected-but-not-reproduced CDP timing window between `Input.insertText` and the next `imeSetComposition`, with a send-batching hardening.

**Not fully validated:** the native macOS Korean IME value-loss path (not automatable; CDP preserves the final value, as the PR description notes) and non-Chromium IME flows (no CDP equivalent). The fix only uses standard composition events and rAF, so cross-browser risk looks low.

app/src/sandbox/combobox-ime-6663/test-chrome.ts
diegohaz: Fixed in 71bdf1fe0 - added `app/src/sandbox/combobox-ime-6663/test.ts` as fast happy-dom coverage for the inter-composition focus regression. I also red-checked it by temporarily restoring the old immediate composition-end refresh; it failed at `expect(q.option("사과")).not.toHaveFocus()` for the expected reason.

app/src/sandbox/combobox-ime-6663/test-chrome.ts
diegohaz: Fixed in 71bdf1fe0 - the Chrome CDP test now starts `Input.insertText("사")` and the next `imeSetComposition("고")` before awaiting them together, preserving command order while removing the awaited frame gap between the two steps.

diegohaz:
**Fixed**

- 71bdf1fe0 (https://github.com/ariakit/ariakit/pull/6664#discussion_r3529594213): Added a fast happy-dom regression test for the inter-composition focus boundary. The test was red-checked by temporarily restoring the old immediate composition-end refresh and failed at the expected option-focus assertion.
- 71bdf1fe0 (https://github.com/ariakit/ariakit/pull/6664#discussion_r3529594215): Hardened the Chrome CDP repro by starting the `Input.insertText("사")` commit and next `imeSetComposition("고")` before awaiting them together, removing the awaited frame gap.

**Verified**

- `pnpm test-react app/src/sandbox/combobox-ime-6663/test.ts`
- `pnpm test-react18 app/src/sandbox/combobox-ime-6663/test.ts`
- `APP_PORT=51409 NEXTJS_PORT=51409 pnpm -F app run test-chrome app/src/sandbox/combobox-ime-6663/test-chrome.ts`
- `pnpm -F app exec tsc -p tsconfig.test.json --pretty false`
- `pnpm tsc`
- `pnpm lint-fix`

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -555,6 +564,17 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       }
     });
 
+    useEffect(() => cancelCompositionEndFrame, []);
```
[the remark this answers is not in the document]
diegohaz: I looked into this while re-reviewing, and the lint premise doesn't hold in this repo: the `lint` script runs `oxlint --disable-nested-config` against `oxlint.config.ts`, which enables the `react` plugin but configures no exhaustive-deps rule — and the codebase relies on intentional partial dependency arrays throughout (including the auto-select `useUpdateEffect` above in this same file). CI Lint passes on this head with the line as written.

Behaviorally the pattern is safe: `cancelCompositionEndFrame` only reads and writes refs, so the first-render instance captured by the mount effect is interchangeable with any later instance, and unmount-only cancellation is exactly the intent. Rewriting it as an inline cleanup (`useEffect(() => () => cancelCompositionEndFrame(), [])`) or adding `useCallback` plus a dependency would not change behavior, so I'd leave it as is.

diegohaz (commented):
**Re-review** covering the commits since my last reviewed SHA (2cd58f9 → 71bdf1f, one commit: "Add fast Combobox Korean IME test"), plus a pass over the unresolved review threads. The library change (`combobox.tsx`) is unchanged since the initial review.

**Verified fixes from the previous review** (re-validated locally in an isolated worktree at the new head, not taken on trust):

- `app/src/sandbox/combobox-ime-6663/test.ts` (new happy-dom test): passes under both `pnpm test-react` and `pnpm test-react18`, and my independent red-check — reverting `combobox.tsx` to `main` — fails it at `expect(q.option("사과")).not.toHaveFocus()`, the exact regression signal. Resolved thread confirmed accurate.
- `test-chrome.ts` batched CDP sends: passes 11/11 against a local dev server (including `--repeat-each=10` with parallel workers). I specifically checked that batching didn't mask the bug on unfixed code — a plausible risk, since shrinking the gap could let the canceling `compositionstart` land before React's deferred focus move, which would make `movedToOption` vacuously false. It doesn't: with `main`'s `combobox.tsx`, the test still fails at `test.expect(focusMove.movedToOption).toBe(false)` (line 89) on both attempts, so the regression coverage retains its detection power. Resolved thread confirmed accurate.

**Prior threads:** both of my initial-review threads were addressed and are correctly resolved. The one open thread from the Copilot reviewer (`useEffect(() => cancelCompositionEndFrame, [])` deps/lint concern) doesn't hold up for this repo — details and verdict posted as a reply on that thread: the oxlint config enables no exhaustive-deps rule, CI Lint passes on this head, and the cleanup closure touches refs only, so the captured first-render instance is behaviorally identical to any later one.

**No new findings.** The incremental diff introduces no new library code, and CI is fully green on the new head (45/45 checks, including all browser suites and perf).

PR #6667, opened 2026-07-06
Avoid parameter reassignment in React code

diegohaz:
## Motivation

Several React examples and store hooks were mutating function parameters or props-shaped objects before passing them along. Those patterns make the code harder to follow and can obscure which values are original inputs versus derived values.

One example also called `useMenuContext` conditionally, which violates the Rules of Hooks even though the value is only needed for one branch.

## Solution

Keep the behavior-preserving cleanups from the original PR and drop the React Compiler oxlint rule experiment.

This PR now:

- replaces parameter reassignment with derived local variables in React examples and store hooks;
- passes forwarded refs explicitly instead of storing `ref` in a props object in the nested combobox menu example;
- calls `useMenuContext` unconditionally in the WordPress menu modal example;
- preserves the form store hook API changes from `main` while avoiding reassignment of the store parameter.

The `react/react-compiler` oxlint rule is not enabled, and there are no `react/react-compiler` suppression comments left in the diff.

### DIFF
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -39,6 +39,7 @@ export default function Example() {
     // so the iframe's own document is the scroller.
     const doc = iframe.contentDocument;
     if (!doc?.body) return;
+    // oxlint-disable-next-line react/react-compiler -- Initializes iframe body.
     doc.body.style.margin = "0";
     const root = createRoot(doc.body);
     root.render(<EmbeddedList />);
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -29,6 +29,10 @@ function EmbeddedList() {
   );
 }
 
+function initializeFrameBody(body: HTMLElement) {
+  body.style.margin = "0";
+}
+
 export default function Example() {
   const [iframe, setIframe] = useState<HTMLIFrameElement | null>(null);
 
@@ -39,8 +43,7 @@ export default function Example() {
     // so the iframe's own document is the scroller.
     const doc = iframe.contentDocument;
     if (!doc?.body) return;
-    // oxlint-disable-next-line react/react-compiler -- Initializes iframe body.
-    doc.body.style.margin = "0";
+    initializeFrameBody(doc.body);
     const root = createRoot(doc.body);
     root.render(<EmbeddedList />);
     return () => {
--- a/app/src/sandbox/composite-iframe-6312/index.react.tsx
+++ b/app/src/sandbox/composite-iframe-6312/index.react.tsx
@@ -29,10 +29,6 @@ function EmbeddedList() {
   );
 }
 
-function initializeFrameBody(body: HTMLElement) {
-  body.style.margin = "0";
-}
-
 export default function Example() {
   const [iframe, setIframe] = useState<HTMLIFrameElement | null>(null);
 
@@ -43,7 +39,7 @@ export default function Example() {
     // so the iframe's own document is the scroller.
     const doc = iframe.contentDocument;
     if (!doc?.body) return;
-    initializeFrameBody(doc.body);
+    doc.body.style.margin = "0";
     const root = createRoot(doc.body);
     root.render(<EmbeddedList />);
     return () => {
--- a/app/src/sandbox/composite-renderer-6215/index.react.tsx
+++ b/app/src/sandbox/composite-renderer-6215/index.react.tsx
@@ -67,17 +67,11 @@ export default function Example() {
   // Surface the leak live as the user scrolls. React bails out of re-rendering
   // when the count is unchanged, so this only re-renders when it actually moves.
   useEffect(() => {
-    let canceled = false;
-    const tick = () => {
-      if (canceled) return;
+    let raf = requestAnimationFrame(function tick() {
       setDetached(countDetachedObservedItems());
-      requestAnimationFrame(tick);
-    };
-    const raf = requestAnimationFrame(tick);
-    return () => {
-      canceled = true;
-      cancelAnimationFrame(raf);
-    };
+      raf = requestAnimationFrame(tick);
+    });
+    return () => cancelAnimationFrame(raf);
   }, []);
 
   return (
--- a/app/src/sandbox/composite-renderer-6215/index.react.tsx
+++ b/app/src/sandbox/composite-renderer-6215/index.react.tsx
@@ -69,6 +69,7 @@ export default function Example() {
   useEffect(() => {
     let raf = requestAnimationFrame(function tick() {
       setDetached(countDetachedObservedItems());
+      // oxlint-disable-next-line react/react-compiler -- Repeated RAF probe.
       raf = requestAnimationFrame(tick);
     });
     return () => cancelAnimationFrame(raf);
--- a/app/src/sandbox/composite-renderer-6215/index.react.tsx
+++ b/app/src/sandbox/composite-renderer-6215/index.react.tsx
@@ -67,12 +67,17 @@ export default function Example() {
   // Surface the leak live as the user scrolls. React bails out of re-rendering
   // when the count is unchanged, so this only re-renders when it actually moves.
   useEffect(() => {
-    let raf = requestAnimationFrame(function tick() {
+    let canceled = false;
+    const tick = () => {
+      if (canceled) return;
       setDetached(countDetachedObservedItems());
-      // oxlint-disable-next-line react/react-compiler -- Repeated RAF probe.
-      raf = requestAnimationFrame(tick);
-    });
-    return () => cancelAnimationFrame(raf);
+      requestAnimationFrame(tick);
+    };
+    const raf = requestAnimationFrame(tick);
+    return () => {
+      canceled = true;
+      cancelAnimationFrame(raf);
+    };
   }, []);
 
   return (
--- a/app/src/sandbox/create-ref-6349/index.solid.tsx
+++ b/app/src/sandbox/create-ref-6349/index.solid.tsx
@@ -20,7 +20,9 @@ export default function Example() {
       <label class="flex flex-col gap-1">
         Name
         <input
+          // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
           ref={(element) => {
+            // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
             nameInput = element;
           }}
           class="rounded border border-gray-300 px-3 py-1"
@@ -29,6 +31,7 @@ export default function Example() {
       <label class="flex flex-col gap-1">
         Email
         <input
+          // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
           ref={(element) => {
             emailInput = element;
           }}
--- a/app/src/sandbox/create-ref-6349/index.solid.tsx
+++ b/app/src/sandbox/create-ref-6349/index.solid.tsx
@@ -20,9 +20,7 @@ export default function Example() {
       <label class="flex flex-col gap-1">
         Name
         <input
-          // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
           ref={(element) => {
-            // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
             nameInput = element;
           }}
           class="rounded border border-gray-300 px-3 py-1"
@@ -31,7 +29,6 @@ export default function Example() {
       <label class="flex flex-col gap-1">
         Email
         <input
-          // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
           ref={(element) => {
             emailInput = element;
           }}
--- a/examples/menu-slide/menu.tsx
+++ b/examples/menu-slide/menu.tsx
@@ -91,14 +91,15 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
     </>
   );
 
+  const menuButtonDefaults = {
+    store: menu,
+    showOnHover: false,
+    className: "button",
+    render: <button />,
+  } satisfies Ariakit.MenuButtonProps;
+
   const renderMenuButton = (menuButtonProps: MenuButtonProps) => (
-    <Ariakit.MenuButton
-      store={menu}
-      showOnHover={false}
-      className="button"
-      render={<button />}
-      {...menuButtonProps}
-    >
+    <Ariakit.MenuButton {...menuButtonDefaults} {...menuButtonProps}>
       {menuButtonChildren}
     </Ariakit.MenuButton>
   );
@@ -131,14 +132,7 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
         />
       ) : (
         // Otherwise, we just render the menu button.
-        <Ariakit.MenuButton
-          ref={ref}
-          store={menu}
-          showOnHover={false}
-          className="button"
-          render={<button />}
-          {...props}
-        >
+        <Ariakit.MenuButton ref={ref} {...menuButtonDefaults} {...props}>
           {menuButtonChildren}
         </Ariakit.MenuButton>
       )}
--- a/examples/menu-slide/menu.tsx
+++ b/examples/menu-slide/menu.tsx
@@ -84,6 +84,13 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
     return () => parentWrapper.removeEventListener("scroll", onScroll);
   }, [parent, menu]);
 
+  const menuButtonChildren = (
+    <>
+      <span className="label">{label}</span>
+      <Ariakit.MenuButtonArrow />
+    </>
+  );
+
   const renderMenuButton = (menuButtonProps: MenuButtonProps) => (
     <Ariakit.MenuButton
       store={menu}
@@ -92,8 +99,7 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
       render={<button />}
       {...menuButtonProps}
     >
-      <span className="label">{label}</span>
-      <Ariakit.MenuButtonArrow />
+      {menuButtonChildren}
     </Ariakit.MenuButton>
   );
 
@@ -125,7 +131,16 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
         />
       ) : (
         // Otherwise, we just render the menu button.
-        renderMenuButton({ ref, ...props })
+        <Ariakit.MenuButton
+          ref={ref}
+          store={menu}
+          showOnHover={false}
+          className="button"
+          render={<button />}
+          {...props}
+        >
+          {menuButtonChildren}
+        </Ariakit.MenuButton>
       )}
       <Ariakit.Menu
         store={menu}
--- a/examples/menu-slide/menu.tsx
+++ b/examples/menu-slide/menu.tsx
@@ -84,23 +84,16 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
     return () => parentWrapper.removeEventListener("scroll", onScroll);
   }, [parent, menu]);
 
-  const menuButtonChildren = (
-    <>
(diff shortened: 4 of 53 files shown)


app/src/sandbox/composite-iframe-6312/index.react.tsx
```diff
@@ -39,6 +39,7 @@ export default function Example() {
     // so the iframe's own document is the scroller.
     const doc = iframe.contentDocument;
     if (!doc?.body) return;
+    // oxlint-disable-next-line react/react-compiler -- Initializes iframe body.
     doc.body.style.margin = "0";
```
diegohaz: What's the problem here? Is it a false positive?

app/src/sandbox/composite-renderer-6215/index.react.tsx
```diff
@@ -69,6 +69,7 @@ export default function Example() {
   useEffect(() => {
     let raf = requestAnimationFrame(function tick() {
       setDetached(countDetachedObservedItems());
+      // oxlint-disable-next-line react/react-compiler -- Repeated RAF probe.
       raf = requestAnimationFrame(tick);
```
diegohaz: What's the problem here? Is it a false positive? Check others too.

app/src/sandbox/create-ref-6349/index.solid.tsx
```diff
@@ -20,7 +20,9 @@ export default function Example() {
       <label class="flex flex-col gap-1">
         Name
         <input
+          // oxlint-disable-next-line react/react-compiler -- Solid ref callback.
```
diegohaz: The oxlint rule should apply only to react files.

oxlint.config.ts
```diff
@@ -15,6 +15,8 @@ export default defineConfig({
     "no-unsafe-type-assertion": "off",
     "no-unassigned-import": "off",
     "react-in-jsx-scope": "off",
+    "react/exhaustive-deps": "error",
```
diegohaz: `react/exhaustive-deps` was already enforced at `error` level before this change: it ships in oxlint's `correctness` category, which this config sets to `"error"`. I verified on the base config (without this line) using a probe component with a missing dependency — oxlint reports `error react-hooks(exhaustive-deps)` there as well. So this entry is a no-op, and `react/react-compiler` is the only rule this PR actually enables.

Not harmful, but the PR description's motivation ("The React rules were not explicitly enabled, so compiler-related issues could be missed") only holds for `react/react-compiler`. Consider dropping this line, or keeping it deliberately as documentation that the rule is load-bearing — either way it doesn't change lint behavior.

packages/ariakit-react-components/src/dialog/utils/use-root-dialog.ts
```diff
@@ -25,7 +25,14 @@ export function useRootDialog({
...
     // oxlint-disable-next-line exhaustive-deps
-  }, [updated, enabled, contentElement, attribute, contentId]);
+  }, [
+    // oxlint-disable-next-line react/react-compiler -- Force-update token.
+    updated,
+    enabled,
+    contentElement,
+    attribute,
+    contentId,
+  ]);
```
diegohaz: This deps array was only split across 8 lines to attach the per-dependency suppression to `updated`. Since the pre-existing `exhaustive-deps` disable on the line above already suppresses dependency checking for the whole array, a combined single-line disable achieves the same with 7 fewer lines — and matches the pattern this PR already uses in `hooks.ts` (`useMergeRefs`/`useWrapElement`):

```suggestion
    // oxlint-disable-next-line exhaustive-deps react/react-compiler
  }, [updated, enabled, contentElement, attribute, contentId]);
```

Verified locally: `oxlint --disable-nested-config` passes on this file with the suggested form.

examples/menu-slide/menu.tsx
```diff
@@ -125,7 +131,16 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
...
+        <Ariakit.MenuButton
+          ref={ref}
+          store={menu}
+          showOnHover={false}
+          className="button"
+          render={<button />}
+          {...props}
+        >
+          {menuButtonChildren}
+        </Ariakit.MenuButton>
```
diegohaz: For context on the duplication introduced here (also noted by CodeRabbit): the previous `renderMenuButton({ ref, ...props })` call is flagged by `react/react-compiler` with "Refs: Cannot access refs during render", because `ref` is passed into a plain function call during render and the rule conservatively assumes the function may read `ref.current`. Passing it as a JSX prop is fine, hence the inline copy.

If you'd rather keep a single source of truth for the button defaults, this passes both oxlint and `pnpm tsc` (verified locally) and keeps both render sites thin:

```tsx
const menuButtonDefaults: MenuButtonProps = {
  store: menu,
  showOnHover: false,
  className: "button",
  render: <button />,
};

const renderMenuButton = (menuButtonProps: MenuButtonProps) => (
  <Ariakit.MenuButton {...menuButtonDefaults} {...menuButtonProps}>
    {menuButtonChildren}
  </Ariakit.MenuButton>
);
```

with the non-submenu branch becoming:

```tsx
<Ariakit.MenuButton ref={ref} {...menuButtonDefaults} {...props}>
  {menuButtonChildren}
</Ariakit.MenuButton>
```

Optional — the current version is behavior-equivalent to `main`; this just removes the drift risk between the two copies.

app/src/sandbox/composite-iframe-6312/index.react.tsx
diegohaz: Effectively yes. Removing the suppression, the rule reports:

> Immutability: This value cannot be modified — Modifying a value returned from 'useState()', which should not be modified directly. Use the setter function to update instead

`doc` derives from the `iframe` state value, and the compiler's model treats state-derived objects as frozen even inside effects — it can't tell that `contentDocument.body` is an external DOM node rather than React-managed state. Mutating external DOM inside an effect is valid, so the code is correct and the suppression is the right call (there's no clean restructure either: the iframe intentionally lives in state so the effect re-runs once it mounts).

Minor: the comment reason could state why the rule is wrong rather than what the line does, e.g. `-- External iframe DOM; the rule treats state-derived objects as frozen.`

app/src/sandbox/composite-renderer-6215/index.react.tsx
diegohaz: This one is a genuine rule bug, not just a modeling limitation. Without the suppression, oxlint's react-compiler rule fails with an internal invariant:

> Invariant: [InferMutationAliasingEffects] Expected value kind to be initialized help: &lt;unknown&gt; tick$31

Its mutation-aliasing inference chokes on the self-referencing named function expression (`function tick` reassigning the outer `raf` while referencing itself). Probably worth reporting to oxc.

It can also be sidestepped with no suppression at all — this lints clean (verified locally) and is behavior-identical:

```tsx
useEffect(() => {
  let raf = 0;
  const tick = () => {
    setDetached(countDetachedObservedItems());
    raf = requestAnimationFrame(tick);
  };
  raf = requestAnimationFrame(tick);
  return () => cancelAnimationFrame(raf);
}, []);
```

Re "check others too": I removed every new suppression in this PR and re-ran oxlint against the original code to capture the underlying diagnostics. These two sandbox comments are the only false positives. The rest suppress real, intentional rule hits: setState-in-effect advisories (`copy-code`, `hovercard`, `menu-list`, `tag/utils`, `dialog-details`, `form-callback-queue`, `popover-responsive`, `collection-renderer`, `composite`), the React 17 fallback paths in `hooks.ts`, the deprecated store hook methods (`form-store`, `react-store`), the render-counter sandbox (`menu-initial-focus-render`), and the `usePrevious` ref read. The Solid file is covered in the other thread.

app/src/sandbox/create-ref-6349/index.solid.tsx
diegohaz: Agreed. A config-level override fixes this cleanly and makes all three suppression comments in this file unnecessary. Verified locally: with the base version of this file (no suppressions) plus the override below, oxlint reports 0 errors (vs 3 `react/react-compiler` errors without it):

```ts
{
  // These React-specific rules don't apply to Solid code.
  files: ["**/*.solid.tsx", "packages/ariakit-solid*/**"],
  rules: {
    "react/exhaustive-deps": "off",
    "react/react-compiler": "off",
  },
},
```

The solid packages don't currently trigger the rule, but including them keeps React-only semantics from ever being enforced on Solid sources.

diegohaz (commented):
**Initial review** of the full diff from the merge base `e4676fb51` (current `main`) to head `e0a25ff40`, 29 files. Branch checked out locally for validation.

**Validation performed**

- `pnpm exec oxlint --disable-nested-config` on head: clean (one pre-existing warning in `app/src/lib/reference-tokenizer.test.ts`, a file untouched by this PR).
- `pnpm tsc`: passes.
- `pnpm test-react form`: 13 files / 58 tests pass (covers the `form-control`/`form-store` logic changes).
- Reproduced the original diagnostics for every refactored/suppressed site by reverting each file to base and re-running oxlint, to weigh each chosen fix against alternatives.
- CI is green; the perf report shows no significant changes in either the Chrome or Node sections.

**Verified behavior-preserving**

- `form-control.tsx`: the module-level WeakMap cache is sound — `getControlItemIds` is pure over `(items, name)`, and identity-keying on `items` matches the old closure's invalidation semantics. The cache is now shared across controls (strictly fewer recomputations), and the ID objects keep stable identity for `useStoreStateObject` equality checks.
- `form-store.ts`/`tab-store.ts`/`collection-renderer.tsx` local-variable routing, the `dialog-radix`/`sidebar`/`table` derived constants, and the `menu-nested-combobox` children/ref restructure are all equivalent to `main`.
- `menu-wordpress-modal`: moving `useMenuContext()` above the early return fixes a real conditional-hook violation.

**Findings** (inline comments + thread replies)

- `oxlint.config.ts`: the `react/exhaustive-deps: "error"` entry is a no-op — the rule already ran at error level via the `correctness` category; `react/react-compiler` is the only rule this PR actually enables.
- Replies on the three open threads: the two questioned sandbox suppressions are the only false positives among all new suppressions (one compiler-model limitation on state-derived DOM, one internal rule crash worth reporting to oxc — with a validated suppression-free restructure), plus a validated config override that scopes both rules away from Solid files and removes the `.solid.tsx` suppressions.
- Two optional simplifications: a single-line combined disable in `use-root-dialog.ts`, and shared `MenuButton` defaults in `menu-slide` to remove the duplicated block.

**Changeset**: none needed in my judgment — lint config plus behavior-preserving refactors, no user-facing changes.

**Not fully validated locally**: browser/Playwright suites (relied on CI, which is green).

app/src/sandbox/create-ref-6349/index.solid.tsx
diegohaz: I believe we should actually enable `react/react-compiler` only for `**/*.react.tsx` and `packages/ariakit-react*/**` (opt-in, not opt-out).

Also, we may not need to enable `react/exhaustive-deps` as anothe review mentioned.

oxlint.config.ts
diegohaz: Let's remove `react/exhaustive-deps` from here if it's already enabled.

app/src/sandbox/composite-renderer-6215/index.react.tsx
diegohaz: Update after testing against the real compiler: this crash is an **upstream React Compiler bug**, not an oxlint port bug — `babel-plugin-react-compiler` 1.0.0 fails with the exact same `Invariant: [InferMutationAliasingEffects] Expected value kind to be initialized` on this pattern (verified locally with a minimal repro of the self-referencing `function tick`).

Existing issues:

- oxlint side: [oxc-project/oxc#23752](https://github.com/oxc-project/oxc/issues/23752) (open) hits the same invariant via a self-referencing named function expression; [oxc-project/oxc#23750](https://github.com/oxc-project/oxc/issues/23750) (open) hits it via a destructuring alias.
- upstream: [react/react#34888](https://github.com/react/react/issues/34888) (open, "Recursive functions seem to confuse the compiler") with fix PR [react/react#36464](https://github.com/react/react/pull/36464) touching the same `InferMutationAliasingEffects` hoisted-context handling, though its repro is the `useEffectEvent` flavor — the named-function-expression invariant crash itself doesn't have an exact upstream issue yet.

The suppression-free restructure from the previous comment still works either way.

app/src/sandbox/composite-iframe-6312/index.react.tsx
diegohaz: Follow-up: verified this against the real compiler — `babel-plugin-react-compiler` 1.0.0 reports the same "This value cannot be modified" error on a minimal repro of this pattern. So the diagnostic matches upstream React Compiler behavior (a compiler-model limitation, not an oxlint port bug), and there's nothing to report to oxc for this one.

diegohaz: Fixed in 4a5bdbca7 - confirmed this was a React Compiler false positive on mutating the same-origin iframe body, so the mutation now goes through initializeFrameBody and the suppression is gone.

app/src/sandbox/composite-renderer-6215/index.react.tsx
diegohaz: Fixed in 4a5bdbca7 - confirmed this was a React Compiler internal invariant around the recursive RAF reassignment. The RAF probe now uses a cancellation flag, and I also checked the other suppressions addressed in this pass.

app/src/sandbox/create-ref-6349/index.solid.tsx
diegohaz: Fixed in 4a5bdbca7 - added a Solid-file override for the React-specific rules and removed the inline React Compiler suppressions from this Solid sandbox.

oxlint.config.ts
diegohaz: Fixed in 4a5bdbca7 - removed the explicit react/exhaustive-deps entry and updated the PR description to clarify that exhaustive-deps remains enforced by the existing correctness category.

packages/ariakit-react-components/src/dialog/utils/use-root-dialog.ts
diegohaz: Fixed in 4a5bdbca7 - collapsed this to a single dependency-array suppression for exhaustive-deps and react/react-compiler.

examples/menu-slide/menu.tsx
diegohaz: Fixed in 4a5bdbca7 - introduced menuButtonDefaults, reused it in both MenuButton render paths, and typed it with satisfies Ariakit.MenuButtonProps.

diegohaz:
**Fixed**

- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757270)): confirmed the iframe body mutation warning was a React Compiler false positive, moved the mutation through `initializeFrameBody`, and removed the suppression.
- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757419)): replaced the recursive RAF reassignment with a cancellation-flag loop and checked the related suppressions in this pass.
- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757508)): scoped React-specific oxlint rules away from Solid files and removed the Solid sandbox suppressions.
- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757615)): removed the redundant explicit `react/exhaustive-deps` config entry and updated the PR description to clarify the existing `correctness` enforcement.
- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757713)): collapsed the `useRootDialog` dependency suppression into one combined directive.
- 4a5bdbca7 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531757826)): introduced `menuButtonDefaults` and reused it in both menu button render paths.

app/src/sandbox/composite-iframe-6312/index.react.tsx
```diff
@@ -39,7 +43,7 @@ export default function Example() {
     // so the iframe's own document is the scroller.
     const doc = iframe.contentDocument;
     if (!doc?.body) return;
-    doc.body.style.margin = "0";
+    initializeFrameBody(doc.body);
```
diegohaz: Revert this.

app/src/sandbox/composite-renderer-6215/index.react.tsx
```diff
@@ -67,11 +67,17 @@ export default function Example() {
...
-      raf = requestAnimationFrame(tick);
-    });
-    return () => cancelAnimationFrame(raf);
+      requestAnimationFrame(tick);
+    };
+    const raf = requestAnimationFrame(tick);
+    return () => {
+      canceled = true;
+      cancelAnimationFrame(raf);
+    };
```
diegohaz: Revert

examples/menu-slide/menu.tsx
```diff
@@ -125,7 +132,9 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
         />
       ) : (
         // Otherwise, we just render the menu button.
-        renderMenuButton({ ref, ...props })
+        <Ariakit.MenuButton ref={ref} {...menuButtonDefaults} {...props}>
+          {menuButtonChildren}
+        </Ariakit.MenuButton>
       )}
       <Ariakit.Menu
         store={menu}
```
diegohaz, line 131: Revert this file

packages/ariakit-react-components/src/form/form-control.tsx
```diff
@@ -13,7 +13,6 @@ import type { Props } from "@ariakit/react-utils";
 import { getDocument, cx } from "@ariakit/utils";
```
diegohaz, line 13: Revert this file

oxlint.config.ts
```diff
@@ -15,6 +15,7 @@ export default defineConfig({
     "no-unsafe-type-assertion": "off",
```
diegohaz, line 15: Revert this file

app/src/sandbox/composite-iframe-6312/index.react.tsx
diegohaz: Fixed in d445f3409 - reverted this sandbox change back to main.

app/src/sandbox/composite-renderer-6215/index.react.tsx
diegohaz: Fixed in d445f3409 - reverted this RAF loop change back to main.

examples/menu-slide/menu.tsx
diegohaz: Fixed in d445f3409 - reverted this file back to main.

packages/ariakit-react-components/src/form/form-control.tsx
diegohaz: Fixed in d445f3409 - reverted this file back to main.

oxlint.config.ts
diegohaz: Fixed in d445f3409 - reverted this file back to main, so the react/react-compiler rule is no longer enabled.

diegohaz:
**Fixed**

- d445f3409 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531990852)): reverted `app/src/sandbox/composite-iframe-6312/index.react.tsx` back to `main`.
- d445f3409 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531990947)): reverted `app/src/sandbox/composite-renderer-6215/index.react.tsx` back to `main`.
- d445f3409 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531991036)): reverted `examples/menu-slide/menu.tsx` back to `main`.
- d445f3409 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531991130)): reverted `packages/ariakit-react-components/src/form/form-control.tsx` back to `main`.
- d445f3409 ([reply](https://github.com/ariakit/ariakit/pull/6667#discussion_r3531991221)): reverted `oxlint.config.ts` back to `main`, so `react/react-compiler` is not enabled.

**Kept**

- The PR now keeps only the unmarked behavior-preserving refactors, and all `react/react-compiler` suppression comments have been removed.

PR #6668, opened 2026-07-06
Deprecate form store hook methods

diegohaz:
## Motivation

React Compiler-compatible hooks need to be called as top-level hook functions. Form stores still expose hook-like methods such as `form.useValue`, `form.useValidate`, and `form.useSubmit`, which encourage member-call hook usage.

The legacy reference website also needs to present the new hooks and soft-deprecated store methods correctly: deprecated return props should appear at the end of the `useFormStore` Return Props list, and standalone function pages such as `useFormValidate` should not derive unrelated Required Props from their `store` parameter.

## Solution

Added top-level form store hooks and soft-deprecated the store methods:

```tsx
const value = useFormValue(form, form.names.email);

useFormValidate(form, (state) => {
  // ...
});

useFormSubmit(form, async (state) => {
  // ...
});
```

The existing store methods remain functional and now delegate to the new hooks, with `@deprecated` JSDoc pointing to the replacement APIs. The public `@ariakit/react` form entry exports the new hooks, while `@ariakit/react-components` keeps component hook name conflicts isolated to separate subpath files.

Docs, root examples, app sandboxes, and internal form component usage were migrated to the new top-level hooks. A changeset documents the soft deprecation for `@ariakit/react-components` and `@ariakit/react`.

The legacy website reference generator now sorts Return Props with the same deprecated-last ordering used by Required and Optional Props, and it only treats a function's first parameter as a prop table when it is actually a `props`/`Options`/`Props` parameter. Focused tests cover the new form hook pages, store props, typed destructured props, and deprecated return-prop ordering.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test-react examples/form/test.ts examples/form-radio/test.ts examples/form-select/test.ts examples/form-callback-queue/test.ts app/src/sandbox/form-6307/test.ts`
- `pnpm test website/build-pages/get-page-content.test.ts website/build-pages/reference-utils.test.ts`
- `pnpm build`
- `pnpm build-website-lite`
- `pnpm lint`
- `git diff --check`

### DIFF
--- a/.changeset/290-form-store-hooks.md
+++ b/.changeset/290-form-store-hooks.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added `useFormValue`, `useFormValidate`, and `useFormSubmit`, and deprecated their matching [`useFormStore`](https://ariakit.com/reference/use-form-store) methods.
--- a/.changeset/290-form-store-hooks.md
+++ b/.changeset/290-form-store-hooks.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Added `useFormValue`, `useFormValidate`, and `useFormSubmit`, and deprecated their matching [`useFormStore`](https://ariakit.com/reference/use-form-store) methods.
+Deprecated the [`useValue`](https://ariakit.com/reference/use-form-store#usevalue), [`useValidate`](https://ariakit.com/reference/use-form-store#usevalidate), and [`useSubmit`](https://ariakit.com/reference/use-form-store#usesubmit) methods of [`useFormStore`](https://ariakit.com/reference/use-form-store) in favor of the new top-level form hooks.
--- a/packages/ariakit-react-components/src/form/form-store.ts
+++ b/packages/ariakit-react-components/src/form/form-store.ts
@@ -12,9 +12,89 @@ import type {
 } from "../collection/collection-store.ts";
 import { useCollectionStoreProps } from "../collection/collection-store.ts";
 
-export function useFormStoreProps<
-  T extends Omit<FormStore, "useValue" | "useValidate" | "useSubmit">,
->(store: T, update: () => void, props: FormStoreProps) {
+type FormStoreHookStore<T extends FormStoreValues = FormStoreValues> = Omit<
+  FormStore<T>,
+  "useValue" | "useValidate" | "useSubmit"
+>;
+
+/**
+ * Re-renders the component when the value of the given field changes and
+ * returns the current value.
+ *
+ * Live examples:
+ * - [Form with Select](https://ariakit.com/examples/form-select)
+ * @example
+ * const form = useFormStore({
+ *   defaultValues: { email: "" },
+ * });
+ * const emailValue = useFormValue(form, form.names.email);
+ */
+// oxlint-disable-next-line no-unnecessary-type-parameters
+export function useFormValue<T = any>(
+  store: FormStoreHookStore,
+  name: StringLike,
+): T {
+  return useStoreState(store, () => store.getValue<T>(name));
+}
+
+/**
+ * Registers a callback that will be used to validate the form when
+ * [`validate`](https://ariakit.com/reference/use-form-store#validate) is
+ * called, typically when a form field is touched or when the form is submitted.
+ *
+ * Live examples:
+ * - [FormRadio](https://ariakit.com/examples/form-radio)
+ * @example
+ * useFormValidate(form, async (state) => {
+ *   const errors = await api.validate(state.values);
+ *   if (errors) {
+ *     form.setErrors(errors);
+ *   }
+ * });
+ */
+export function useFormValidate<T extends FormStoreValues = FormStoreValues>(
+  store: FormStoreHookStore<T>,
+  callback: Core.FormStoreCallback<FormStoreState<T>>,
+) {
+  callback = useEvent(callback);
+  // Whenever the items change (for example, when form fields are lazily
+  // rendered), we need to reset the callbacks so they always run in a
+  // consistent order.
+  const items = useStoreState(store, "items");
+  useEffect(() => store.onValidate(callback), [store, items, callback]);
+}
+
+/**
+ * Registers a callback that will be used to submit the form when
+ * [`submit`](https://ariakit.com/reference/use-form-store#submit) is called.
+ *
+ * Live examples:
+ * - [FormRadio](https://ariakit.com/examples/form-radio)
+ * - [Form with Select](https://ariakit.com/examples/form-select)
+ * @example
+ * useFormSubmit(form, async (state) => {
+ *   try {
+ *     await api.submit(state.values);
+ *   } catch (errors) {
+ *     form.setErrors(errors);
+ *   }
+ * });
+ */
+export function useFormSubmit<T extends FormStoreValues = FormStoreValues>(
+  store: FormStoreHookStore<T>,
+  callback: Core.FormStoreCallback<FormStoreState<T>>,
+) {
+  callback = useEvent(callback);
+  // Same logic as useFormValidate.
+  const items = useStoreState(store, "items");
+  useEffect(() => store.onSubmit(callback), [store, items, callback]);
+}
+
+export function useFormStoreProps<T extends FormStoreHookStore>(
+  store: T,
+  update: () => void,
+  props: FormStoreProps,
+) {
   store = useCollectionStoreProps(store, update, props);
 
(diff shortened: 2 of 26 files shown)


.changeset/290-form-store-hooks.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added `useFormValue`, `useFormValidate`, and `useFormSubmit`, and deprecated their matching [`useFormStore`](https://ariakit.com/reference/use-form-store) methods.
```
diegohaz: `useFormValue`, `useFormValidate`, and `useFormSubmit` should be reference links.

diegohaz: This entry combines two changes that the repo's changeset conventions shape differently: adding the hooks is a minor feature (a `200-` entry with a title and a code example), while the soft deprecation should be its own single-sentence `290-` entry. As written, the only changelog announcement of three new public hooks sits in the deprecation slot with no example, and the store-as-first-argument call shape isn't obvious from the names alone.

Suggested split:

`.changeset/200-form-store-hooks.md`:

````md
---
"@ariakit/react": patch
---

Added `useFormValue`, `useFormValidate`, and `useFormSubmit`

The new [`useFormValue`](https://ariakit.com/reference/use-form-value), [`useFormValidate`](https://ariakit.com/reference/use-form-validate), and [`useFormSubmit`](https://ariakit.com/reference/use-form-submit) hooks replace the matching [`useFormStore`](https://ariakit.com/reference/use-form-store) methods with top-level hook calls that are compatible with the React Compiler:

```tsx
const value = useFormValue(form, form.names.email);

useFormSubmit(form, async (state) => {
  // ...
});
```
````

`.changeset/290-form-store-hooks.md` (single sentence, per the soft-deprecation format):

```md
Deprecated the [`useValue`](https://ariakit.com/reference/use-form-store#usevalue), [`useValidate`](https://ariakit.com/reference/use-form-store#usevalidate), and [`useSubmit`](https://ariakit.com/reference/use-form-store#usesubmit) methods of [`useFormStore`](https://ariakit.com/reference/use-form-store) in favor of the new top-level form hooks.
```

This also resolves the reference-link note already left on this line.

packages/ariakit-react-components/src/form/form-store.ts
```diff
@@ -12,38 +12,110 @@ import type {
...
   const useSubmit = useCallback<FormStore["useSubmit"]>(
     (callback) => {
-      callback = useEvent(callback);
-      // Same logic as useValidate.
-      const items = useStoreState(store, "items");
-      useEffect(() => store.onSubmit(callback), [items, callback]);
+      useFormSubmit(store, callback);
     },
     [store],
   );
```
diegohaz, line 121: After this PR's migration, no test in the repo exercises the deprecated `useValue`/`useValidate`/`useSubmit` store methods anymore. Searching `\.useValue\(|\.useValidate\(|\.useSubmit\(` across non-markdown files returns only these wrappers. The methods remain the primary API existing consumers use, and they now run through a new delegation path, so a regression here or in a later refactor of these wrappers would ship without any signal until removal.

Consider one small back-compat test that renders a form using the store methods and asserts that `useValue` re-renders with the new value and that `useValidate`/`useSubmit` callbacks run on validate/submit (optionally also that callbacks re-register in a consistent order when items change, since that's the behavior the `items` dependency exists for). A file like `packages/ariakit-react-components/src/form/form-store.react.test.tsx` runs under `pnpm test-react` (via the `packages/ariakit-react*/src/**/*test.{ts,tsx}` include) and stays out of the default no-loader suite.

I verified the delegation currently works through the migrated example tests (`pnpm test-react` on the form examples and `pnpm test-react18` both pass locally), but those now only cover the new hooks.

packages/ariakit-react-components/src/form/form-store.ts
```diff
@@ -12,38 +12,110 @@ import type {
 } from "../collection/collection-store.ts";
 import { useCollectionStoreProps } from "../collection/collection-store.ts";
 
-export function useFormStoreProps<
-  T extends Omit<FormStore, "useValue" | "useValidate" | "useSubmit">,
->(store: T, update: () => void, props: FormStoreProps) {
+type FormStoreHookStore<T extends FormStoreValues = FormStoreValues> = Omit<
+  FormStore<T>,
+  "useValue" | "useValidate" | "useSubmit"
+>;
```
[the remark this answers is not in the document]
diegohaz: I verified this at head a6c5f0d by running `pnpm -F @ariakit/react-components build` (`ariakit build`, which emits declarations through `rolldown-plugin-dts` with `build: true` on top of TypeScript's declaration emit). The build succeeds, and the bundled output hoists the non-exported alias into the `.d.ts` chunk:

```ts
type FormStoreHookStore<T extends FormStoreValues$1 = FormStoreValues$1> = Omit<FormStore<T>, "useValue" | "useValidate" | "useSubmit">;
declare function useFormValue<T = any>(store: FormStoreHookStore, name: StringLike): T;
```

Private-name declaration errors (TS4078 and friends) occur when a referenced type can't be named from the declaration file, such as a non-exported type imported from another module. A same-file non-exported alias is emitted into the `.d.ts` as a local declaration, so there's no declaration build failure and no change is needed here.

diegohaz (commented):
**Initial review** of `deprecate-form-store-hooks` at head a6c5f0d, diffed against merge base 6960c5e0f (origin/main).

**Scope reviewed:** the new `useFormValue`/`useFormValidate`/`useFormSubmit` hooks and the deprecated store-method delegation in `form-store.ts`, the `@ariakit/react` form entry exports, the internal form component migration, docs/guide/example/sandbox migrations, the changeset, and the legacy website reference generator changes (`getPropsParameter`, deprecated-last Return Props sorting) plus their new tests.

**Validation performed locally at the PR head:**

- `pnpm tsc`: clean.
- `pnpm test website/build-pages/get-page-content.test.ts website/build-pages/reference-utils.test.ts`: 4 tests pass in about 1.2s, so the new tests add no meaningful cost to the default suite that picks them up.
- `pnpm test-react` on the form examples plus `app/src/sandbox/form-6307`: 24 tests pass. `pnpm test-react18` on the form examples: 23 tests pass (the React 18 suite wasn't listed in the PR verification, so I ran it as well).
- `pnpm -F @ariakit/react-components build`: succeeds. The non-exported `FormStoreHookStore` alias is hoisted into the emitted `.d.ts`, so the existing CodeRabbit declaration-emit finding is a false positive (details in a reply on that thread).
- Reference-generation A/B: ran `getReferences` from both main's and this PR's `reference-utils.js` over all 27 `packages/ariakit-react/src/*.ts` entry files (177 references total). The only prop-table differences are the three new hook pages losing the unrelated store-derived Required Props, which is exactly the intended fix. No existing page loses props, and the Return Props ordering change (deprecated last, then alphabetical) is consistent with how Required and Optional Props were already sorted.

**Findings (posted inline):**

1. Changeset shape: the feature and the soft deprecation are combined in one `290-` entry; conventions call for a `200-` feature entry with a code example plus a single-sentence `290-` deprecation entry.
2. Test coverage: after the migration, no test exercises the deprecated `useValue`/`useValidate`/`useSubmit` delegation paths.

**Also checked, no issues found:** `@ariakit/solid` has no form module, so nothing needs mirroring; the `useFormSubmit` store hook and the `FormSubmit` component hook stay in separate subpath files and `@ariakit/react` re-exports only the store hook; reference page slugs (`use-form-value`, `use-form-validate`, `use-form-submit`) are derived from the new exports via `kebabCase`, so the `@deprecated` JSDoc links will resolve once released; the `process.cwd()`-based root heuristic in `reference-utils.js` is justified because the module is imported by Next.js server components (`page-markdown.tsx`) and gets bundled, where an `import.meta.url`-based path would break.

**Not fully validated:** the full website build (`build-website-lite`); I relied on the PR's stated verification there plus the A/B reference dump for generator behavior.

.changeset/290-form-store-hooks.md
diegohaz: Covered by https://github.com/ariakit/ariakit/pull/6668#discussion_r3530931787

diegohaz: Fixed in 6809498f5 - split the new hook announcement into a 200-level changeset and kept the 290-level changeset as a single soft-deprecation sentence.

diegohaz:
**Fixed**

- 6809498f5 (https://github.com/ariakit/ariakit/pull/6668#discussion_r3531047012): split the changeset into a 200-level entry for the new top-level hooks and a 290-level single-sentence soft deprecation entry.

PR #6670, opened 2026-07-06
Consolidate sandbox fixtures

diegohaz:
## Motivation

The app has several sandbox fixtures that cover closely related component behavior but live in separate issue-numbered directories. That makes the sandbox list longer than necessary and splits related test coverage across multiple preview IDs.

This first consolidation pass reduces duplicate sandbox directories while keeping the original regression coverage. Ariakit Tailwind sandboxes are intentionally left out of this pass.

## Solution

Merge high-confidence related sandboxes into descriptive fixture directories and keep the original issue coverage in the combined `test.ts`, `test-browser.ts`, and `test-chrome.ts` files.

The consolidated paths replace the old issue-numbered preview IDs, so links or bookmarks to old paths such as `combobox-item-value-6329` or `toolbar-container-ime-6579` should use the new descriptive sandbox paths instead.

Merged sandbox groups include:

- `combobox-item-value`
- `command-press-state`
- `disclosure-content-lifecycle`
- `form-array-focus`
- `menu-item-checked`
- `menu-nested-combobox`
- `popover-arrow`
- `radio-shift-tab-focus`
- `select-renderer`
- `toolbar-container-focus`
- `tooltip-fullscreen-portal`

Issue URLs are kept in comments next to the relevant tests.

## Testing

- `pnpm lint-fix`
- `pnpm test-react app/src/sandbox/combobox-item-value/test.ts app/src/sandbox/command-press-state/test.ts app/src/sandbox/form-array-focus/test.ts app/src/sandbox/menu-item-checked/test.ts app/src/sandbox/popover-arrow/test.ts app/src/sandbox/radio-shift-tab-focus/test.ts app/src/sandbox/select-renderer/test.ts app/src/sandbox/toolbar-container-focus/test.ts app/src/sandbox/tooltip-fullscreen-portal/test.ts`
- `APP_PORT=4521 NEXTJS_PORT=4521 pnpm -F app run test app/src/sandbox/combobox-item-value/test-browser.ts app/src/sandbox/command-press-state/test-browser.ts app/src/sandbox/disclosure-content-lifecycle/test-browser.ts app/src/sandbox/select-renderer/test-browser.ts app/src/sandbox/toolbar-container-focus/test-browser.ts app/src/sandbox/radio-shift-tab-focus/test-browser.ts app/src/sandbox/form-array-focus/test-browser.ts app/src/sandbox/menu-nested-combobox/test-browser.ts app/src/sandbox/menu-item-checked/test-browser.ts app/src/sandbox/tooltip-fullscreen-portal/test-browser.ts app/src/sandbox/popover-arrow/test-browser.ts`
- `APP_PORT=4521 NEXTJS_PORT=4521 pnpm -F app run test-chrome app/src/sandbox/toolbar-container-focus/test-chrome.ts`
- `pnpm test-react18 app/src/sandbox/radio-shift-tab-focus/test.ts`
- `pnpm test-react app/src/sandbox/radio-shift-tab-focus/test.ts`
- `APP_PORT=4521 NEXTJS_PORT=4521 pnpm -F app run test app/src/sandbox/radio-shift-tab-focus/test-browser.ts app/src/sandbox/toolbar-iframe-6316/test-browser.ts app/src/sandbox/composite-iframe-6312/test-browser.ts`
- `pnpm tsc`
- `pnpm lint`

## Summary by CodeRabbit

* **New Features**
  * Added or updated demo pages for combobox, menu, popover, radio, disclosure, form, toolbar, and iframe-based interactions.
  * Improved examples now cover nested menus, horizontal select layouts, fullscreen tooltips, and portal-based iframe rendering.

* **Bug Fixes**
  * Expanded coverage for focus handling, pressed-state behavior, tooltip placement, and popover arrow positioning.
  * Added regression checks for overlapping search matches, normalized empty values, and menu item checked-state rendering.

* **Chores**
  * Removed several outdated sandbox demos and their related tests.


app/src/sandbox/popover-arrow/index.react.tsx
```diff
@@ -0,0 +1,123 @@
...
+      >
+        <Ariakit.PopoverArrow className="arrow" />
+        <Ariakit.PopoverHeading>{label}</Ariakit.PopoverHeading>
+        <p>Popover outlined by a box-shadow ring.</p>
+      </Ariakit.Popover>
+    </Ariakit.PopoverProvider>
+  );
+}
+
+// See https://github.com/ariakit/ariakit/issues/6320
```
diegohaz, line 38: The consolidation dropped the failure-mode explanation that used to sit on top of `popover-6320/index.react.tsx`. That comment carried the actual mechanics of the bug (why a stale `right: 100%` detaches the arrow specifically in RTL) and mapped the two user-level triggers to the two tests. Without it, the RTL wrapper, the oversized scroller, and the happy-dom test's `arrow.style.right` polling (`popover-arrow/test.ts:41-46`, which also lost its "happy-dom does no layout, so inline insets are the only observable signal" and "poll so we don't race the reposition" notes) read as arbitrary setup.

Suggest restoring it here, and re-adding the per-assertion rationale in `test.ts`:

```suggestion
// See https://github.com/ariakit/ariakit/issues/6320
//
// In an RTL context, a popover that resolves to a `right` placement writes
// `right: 100%` on its arrow element. Later position updates never clear that
// declaration: they only write `left`, `top`, and the new side. With both
// `left` and `right` set, RTL over-constrained absolute positioning ignores
// `left`, so the stale `right: 100%` pins the arrow to the popover's left
// edge, visually detached from the anchor.
//
// Two user-level triggers change the resolved placement while the popover is
// open:
// 1. Scrolling the container so the anchor nears its right edge, which makes
//    the popover flip above the anchor (browser test).
// 2. Clicking "Show above", which changes the `placement` prop (also covered
//    by the happy-dom test, where layout-driven flips can't happen).
```

app/src/sandbox/select-renderer/test-browser.ts
```diff
@@ -17,4 +17,14 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
   });
+
+  test("SelectRenderer forwards horizontal orientation to the item layout", async ({
+    q,
+  }) => {
+    await q.combobox("Favorite fruit").click();
+
+    const cherry = q.option("Cherry");
+    await test.expect(cherry).toHaveCSS("left", "192px");
+    await test.expect(cherry).toHaveCSS("top", "0px");
```
diegohaz, line 35: The move dropped the rationale comment for these assertions, so `192px` is now a magic number and nothing explains what the pre-fix failure looked like. The original comment also justified why the *last* item is the robust one to assert under virtualization. Same loss in the happy-dom twin (`select-renderer/test.ts:17-21`).

```suggestion
    // Horizontal orientation lays items out along the x-axis: the renderer
    // offsets each item by `left` and keeps a shared `top` of 0. The last
    // option "Cherry" (index 2) lands at `itemSize * 2 = 192px`; asserting the
    // last item keeps the check robust because it stays rendered as a
    // persistent index even when virtualization trims middle items. Before the
    // fix, the dropped `orientation` prop fell back to vertical, offsetting by
    // `top` instead.
    const cherry = q.option("Cherry");
    await test.expect(cherry).toHaveCSS("left", "192px");
    await test.expect(cherry).toHaveCSS("top", "0px");
```

app/src/sandbox/disclosure-content-lifecycle/test-browser.ts
```diff
@@ -73,4 +73,76 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
...
+          }
+        });
+        observer.observe(region, { childList: true, subtree: true });
+        start = performance.now();
+        button.click();
+      });
+    });
+
+    test.expect(unmountDelay).toBeLessThan(700);
+    test.expect(unmountDelay).toBeGreaterThan(400);
```
diegohaz, line 136: The move stripped the threshold rationale, leaving `700`/`400` unexplained. `style.css` still documents the 510ms-vs-800ms math, but nothing in the test ties these two numbers to it, and the dropped comment above the `page.evaluate` block (measuring the unmount delay in-browser because Playwright command latency is too coarse to tell 510ms from 800ms) explained why this test needs a `MutationObserver` at all. Suggest restoring both; here's the assertion half:

```suggestion
    // The content should unmount close to 510ms (the longest per-property end
    // time; see style.css), not the buggy max(delays) + max(durations) =
    // 800ms. 700ms sits well between the two so the assertion fails on the
    // buggy timeout and passes on the fix, with margin for setTimeout and
    // requestAnimationFrame jitter.
    test.expect(unmountDelay).toBeLessThan(700);
    // Sanity check: it still waited for the real animation rather than
    // unmounting immediately.
    test.expect(unmountDelay).toBeGreaterThan(400);
```

app/src/sandbox/toolbar-container-focus/test-chrome.ts
```diff
@@ -1,7 +1,18 @@
...
-// Reproduces https://github.com/ariakit/ariakit/issues/6300
+async function commitComposition(page: Page, text: string) {
+  const cdp = await page.context().newCDPSession(page);
+  await cdp.send("Input.imeSetComposition", {
+    text,
+    selectionStart: text.length,
+    selectionEnd: text.length,
+  });
+  await page.keyboard.press("Enter");
+}
```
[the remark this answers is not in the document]
diegohaz: The Enter keydown here is the behavior under test, not an incidental commit path. Issue 6579 is specifically about the *composing Enter keydown*: `CompositeContainer`'s `onKeyDown` Enter branch used to `preventDefault()` and move focus back to the container when the user pressed Enter to commit an IME composition. `Input.imeCommitComposition` commits without dispatching any keydown, so switching to it would remove the exact event the regression guard handles — the test would then pass with or without the fix. The happy-dom twin in `toolbar-container-focus/test.ts` exercises the same gesture via `press.down.Enter(input, { isComposing: true })` and a keyCode-229 variant.

This helper is also preserved verbatim from the pre-existing `toolbar-container-ime-6579/test-chrome.ts`, which has been green in CI. I re-ran it on this branch (`pnpm -F app run test-chrome app/src/sandbox/toolbar-container-focus/test-chrome.ts`): both assertions hold — the composition commits (value contains に) and focus stays in the field. Focus staying put also shows the synthetic Enter reaches Chromium's composition pipeline as a composing key event (otherwise the fixed Enter branch would have moved focus to the container), so the "can leave Chromium in a composing state" concern doesn't materialize here.

diegohaz (commented):
## Review summary

**Initial review** of the full diff from merge base `9a751c1` (current `main`) to head `b601ace` (69 files), plus surrounding context beyond the diff.

### What I checked

- **Coverage fidelity per consolidation group**: for each of the 11 merged groups I compared the deleted sandbox's fixture and tests against the consolidated files. All original assertions, issue links, and test scenarios are preserved (6298 → `combobox-item-value`, 6217 → `command-press-state`, 6220 → `disclosure-content-lifecycle`, 6293 → `form-array-focus`, 6318 + `menu-radio-default-checked` → `menu-item-checked`, 4443 → `menu-nested-combobox`, 6320 → `popover-arrow`, `radio-iframe-shift-tab-focus` → `radio-shift-tab-focus`, orientation → `select-renderer`, 6300 + IME 6579 → `toolbar-container-focus`, 865 → `tooltip-fullscreen-portal`).
- **Query ambiguity in merged fixtures**: the app `q` helpers match accessible names with `exact: true` (`packages/ariakit-test/src/playwright.ts`), so co-hosted widgets ("Apply" vs "Apply empty"/"Apply message", "Fruit" vs "Favorite fruit", "Open menu" vs "Actions") don't collide.
- **Iframe portal rewrite** (`composite-iframe-6312`, `toolbar-iframe-6316`, `radio-shift-tab-focus`): replacing the `useEffect` + `createRoot` bootstrap with `createPortal` into `contentDocument.body` is a real simplification and arguably truer to the react-frame-component embedding the comment cites (that library also portals). Ref-callback timing on src-less iframes is safe (the initial document is synchronous and never replaced), confirmed by cross-browser runs below.
- **Stale references**: repo-wide search for all 21 removed/renamed sandbox path names — no remaining references in code, docs, tests, or CI.
- **Preview/registration**: sandbox previews need no `preview.json`; none of the deleted dirs had one. No changeset needed (private `app` fixtures only), matching changeset-bot.
- **Tooltip merge instrumentation**: the 6585 portal-count selector is scoped to `portal/tooltip-repro`, so the added 865 tooltips can't perturb the leak assertions.
- **Prior bot feedback**: replied on the CodeRabbit `Input.imeCommitComposition` thread — the composing Enter keydown is the regression trigger under test (issue 6579), so the suggested change would defeat the test. CodeRabbit's `useIframeContainer` extraction nitpick targeted the intermediate commit; the final portal version reduces the "duplication" to a 1–3 line ref callback per sandbox, and keeping repro fixtures self-contained is preferable to coupling them through a shared hook.

### Validation performed

- `pnpm tsc` — clean.
- `pnpm test-react` on all 9 consolidated `test.ts` files — 35/35 passed.
- `pnpm -F app run test` (chromium + firefox + webkit) on all 11 consolidated `test-browser.ts` files plus `composite-iframe-6312` and `toolbar-iframe-6316` — 153/153 passed.
- `pnpm -F app run test-chrome` for the merged IME test — 2/2 passed.

### Findings

Three inline comments, all one category: the moves dropped rationale comments that explained magic numbers and non-obvious assertions (popover 6320 failure mechanics, select-renderer `192px` layout math, disclosure `700`/`400` thresholds). Suggested restorations are attached as diff suggestions. A smaller instance not worth its own thread: the `tooltip-fullscreen-portal` 865 tests lost short notes like "otherwise, the tooltip is invisible to the user" — worth restoring opportunistically if you touch that file.

### Not fully validated

I did not re-verify that each consolidated fixture still fails without its original library fix (that would require reverting the fixes); I relied on the fixtures being DOM/behavior-equivalent to the originals, which the file-by-file comparison supports.

app/src/sandbox/popover-arrow/index.react.tsx
diegohaz: Fixed in da837f789 - restored the #6320 failure-mode explanation in the consolidated sandbox and added back the browser and happy-dom rationale comments for the RTL arrow assertions.

app/src/sandbox/select-renderer/test-browser.ts
diegohaz: Fixed in da837f789 - restored the horizontal-orientation, 192px offset, virtualization, and pre-fix vertical fallback rationale in both select-renderer tests.

app/src/sandbox/disclosure-content-lifecycle/test-browser.ts
diegohaz: Fixed in da837f789 - restored the MutationObserver timing rationale and the 700ms/400ms threshold explanation for the disclosure-content lifecycle browser test.

diegohaz:
**Fixed**

- da837f789 ([reply](https://github.com/ariakit/ariakit/pull/6670#discussion_r3532535252)): Restored the #6320 failure-mode explanation in `popover-arrow` and the browser/happy-dom rationale comments for the RTL arrow assertions.
- da837f789 ([reply](https://github.com/ariakit/ariakit/pull/6670#discussion_r3532535321)): Restored the horizontal-orientation, 192px offset, virtualization, and pre-fix vertical fallback rationale in both `select-renderer` tests.
- da837f789 ([reply](https://github.com/ariakit/ariakit/pull/6670#discussion_r3532535367)): Restored the MutationObserver timing rationale and the 700ms/400ms threshold explanation for the disclosure-content lifecycle browser test.

Issue #4851, opened 2025-07-24
TabProvider `setActiveId` callback triggered by `MenuItem` hover when nested in `ComboboxProvider`/`MenuProvider/TabProvider``


PiotrBialekApollo:
### Summary

When using `ComboboxProvider` and `MenuProvider` with a nested `TabProvider` in a separate component, hovering over `MenuItem` elements incorrectly triggers the `setActiveId` callback from `useTabStore`, causing the active tab to change to a non-existent tab ID and resulting in empty tab panels. I tried to create the simplest possible example that triggers this error. In general, in a complex UI this error can be very hard to debug.

### Steps to Reproduce

1. Create a component structure with `ComboboxProvider` and `MenuProvider`
2. Nest a `TabProvider` in a separate component within the menu
3. Use `useTabStore` with a custom `setActiveId` callback
4. Add `MenuItem` elements inside `TabPanel` components
5. Hover over any MenuItem

Code example:
```jsx

export function App() {
  return (
    <Ariakit.ComboboxProvider>
      <Ariakit.MenuProvider>
        <Ariakit.MenuButton>Test</Ariakit.MenuButton>
        <Ariakit.Menu>
          <TabTest />
        </Ariakit.Menu>
      </Ariakit.MenuProvider>
    </Ariakit.ComboboxProvider>
  );
}

function TabTest() {
  const tabStore = Ariakit.useTabStore({
    setActiveId: (id) => {
      console.log('setActiveId', id);
      tabStore.setSelectedId(id);
    },
  });

  const tabs = ['main', 'overview', 'details', 'news'];
  return (
    <Ariakit.TabProvider store={tabStore}>
      <Ariakit.TabList composite={false}>
        {tabs.map((tab) => (
          <Ariakit.Tab key={tab} id={tab}>
            {tab}
          </Ariakit.Tab>
        ))}
      </Ariakit.TabList>
      {tabs.map((tab) => (
        <Ariakit.TabPanel unmountOnHide tabId={tab} key={tab}>
          <Ariakit.MenuItem>Dummy Menu {tab}</Ariakit.MenuItem>
          <Ariakit.MenuItem>Dummy Menu 2 {tab}</Ariakit.MenuItem>
        </Ariakit.TabPanel>
      ))}
    </Ariakit.TabProvider>
  );
}
```

### Version Information
Affected versions: 0.4.5, 0.4.17 (and likely all versions between)
Working version: 0.4.0
Reproducible: Yes, consistently

### Impact
[...]

PR #3799, opened 2024-05-17
Fix scoping in the Composite context

DaniGuardiola:
## Motivation

When a `CompositeItem` is rendered as the same element as a component that provides its own composite-based context — most commonly a `MenuButton` wrapped in `MenuProvider`, but the same applies to `Select`, `Combobox`, and other composite-derived widgets — the item registers on that inner store instead of on the enclosing `Composite`. As a result, arrow-key navigation (in both one- and two-dimensional composites) skips the item: it can never be focused with the keyboard inside the composite widget.

This blocks valid patterns such as a toolbar or grid whose cells are menu buttons (the WordPress/Gutenberg two-dimensional navigation use case in #3768).

The affected markup looks like this — the middle item of each row is unreachable with the arrow keys:

```jsx
<Ariakit.CompositeProvider>
  <Ariakit.Composite>
    <Ariakit.CompositeRow>
      <Ariakit.CompositeItem>Button</Ariakit.CompositeItem>
      <Ariakit.MenuProvider>
        <Ariakit.CompositeItem
          render={<Ariakit.MenuButton>Menu</Ariakit.MenuButton>}
        />
        <Ariakit.Menu>…</Ariakit.Menu>
      </Ariakit.MenuProvider>
      <Ariakit.CompositeItem>Button</Ariakit.CompositeItem>
    </Ariakit.CompositeRow>
  </Ariakit.Composite>
</Ariakit.CompositeProvider>
```

## Solution

`Menu` (like `Select`, `Combobox`, etc.) provides its store through the regular composite context, so any `CompositeItem` nested under `MenuProvider`/`MenuButton` reads the menu's store from `useCompositeContext()` and registers there instead of on the enclosing `Composite`.

The fix applies the same context-scoping pattern already used by `Tab`, `MenuItem`, `SelectItem`, and `ComboboxItem`:

- `Composite` now provides its store through the **scoped** composite context (`CompositeScopedContextProvider`) instead of the plain `CompositeContextProvider`. The scoped provider sets both the regular and the scoped contexts, so this is a superset of the previous behavior.
- The composite collection components — `CompositeItem`, `CompositeRow`, `CompositeSeparator`, `CompositeTypeahead`, `CompositeHover`, `CompositeContainer`, `CompositeItemOffscreen`, and `CompositeRenderer` — now resolve their store from `useCompositeScopedContext()` instead of `useCompositeContext()`, so the whole collection consistently binds to the enclosing `Composite`.

[...]

PR #6671, opened 2026-07-07
Support provider components in store props and store hooks

diegohaz:
## Motivation

Ariakit components resolve their store from the closest compatible context when a `store` prop isn't passed. Because many stores extend one another (for example, `ToolbarProvider` implements a composite store), a component like `CompositeItem` implicitly binds to whatever compatible provider happens to be nearest, with no way to say "use _this_ provider instead".

```tsx
<ComboboxProvider>
  <ToolbarProvider>
    {/* Implicitly binds to the closer ToolbarProvider. There was no way to
        target the outer ComboboxProvider without passing a store object. */}
    <CompositeItem />
  </ToolbarProvider>
</ComboboxProvider>
```

We want to let people be explicit about which provider a component reads from, and eventually stop relying on the implicit "read the nearest extending provider" behavior.

## Solution

Store props and the store-state hooks now accept a **provider component** (for example, `ComboboxProvider`) in place of a store object. When a provider component is passed, the store is read from the closest matching provider through that provider's own context, and nothing else: there is no fallback to other compatible contexts, so the reference is unambiguous.

```tsx
<ComboboxProvider>
  <ToolbarProvider>
    {/* Explicitly reads from the closest ComboboxProvider, even though the
        ToolbarProvider is closer. */}
    <CompositeItem store={ComboboxProvider} />
  </ToolbarProvider>
</ComboboxProvider>
```

The same works with [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject`:

```tsx
const value = useStoreState(ComboboxProvider, "value");
```

### Changes

- Provider components are now branded with their store context (`createStoreContext` gains a `createProviderComponent` helper). A new `useStoreProp` hook resolves a `store` prop that may be either a store object or a provider component, and an exported `ProviderComponent` type and `isProviderComponent` guard back the machinery.
- Every component `store` prop is widened to `Store | ProviderComponent<Store>` and resolves provider components. Some presentational components declare a `store` prop by convention without consuming it yet; those props are widened for consistency (and to stay non-breaking if wired up later) but are not documented as resolving a provider.
- `useStoreState` and `useStoreStateObject` accept provider components and resolve them via context.
- When a provider component is passed but no matching provider is found, the store resolves to `undefined`, so the existing "must be wrapped in a provider" invariants still fire.

This is an additive change: passing a store object or omitting the prop behaves exactly as before.

### DIFF
--- a/.changeset/200-provider-component-store.md
+++ b/.changeset/200-provider-component-store.md
@@ -15,6 +15,8 @@ For example, [`CompositeItem`](https://ariakit.com/reference/composite-item) can
 <CompositeItem store={ComboboxProvider} />
 ```
 
+If no matching provider context exists above the component, the store resolves to `undefined` rather than falling back to a closer compatible context. Note that components composed from multiple layers may still bind lower-level behavior to their own contexts in that case.
+
 The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks accept provider components as well:
 
 ```tsx
--- a/.changeset/200-provider-component-store.md
+++ b/.changeset/200-provider-component-store.md
@@ -7,7 +7,7 @@
 
 Support provider components in store props and store hooks
 
-Component `store` props now accept a provider component (for example, [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider)) in place of a store object. The store is read from the closest matching provider through context, which lets you explicitly bind a component to a specific provider even when a different compatible provider is nested closer.
+Component `store` props now accept a provider component (for example, [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider)) in place of a store object. The store is read from that provider's context, which lets you explicitly bind a component to a specific provider kind even when a different compatible provider is nested closer.
 
 For example, [`CompositeItem`](https://ariakit.com/reference/composite-item) can read from the closest [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider) even when a [`ToolbarProvider`](https://ariakit.com/reference/toolbar-provider) or [`CompositeProvider`](https://ariakit.com/reference/composite-provider) is nested closer to it:
 
@@ -19,4 +19,21 @@ The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useSto
 
 ```tsx
 const value = useStoreState(ComboboxProvider, "value");
+```
+
+The new `ProviderComponent` type is exported from `@ariakit/react` and `@ariakit/react-store` so you can reference the widened props in your own types:
+
+```tsx
+interface Props {
+  store?: ComboboxStore | ProviderComponent<ComboboxStore>;
+}
+```
+
+Note that this is a type-level change to the `store` props: passing store objects keeps working as before, but code that reads a widened `store` prop type and passes its value to an API that expects a store object (for example, [`useFormValidate`](https://ariakit.com/reference/use-form-validate)) must now re-narrow the prop:
+
+```tsx
+interface FormProps extends Ariakit.FormProps {
+  // Accept only store objects so the prop can be passed to useFormValidate.
+  store?: Ariakit.FormStore;
+}
 ```
--- a/.changeset/200-provider-component-store.md
+++ b/.changeset/200-provider-component-store.md
@@ -5,7 +5,7 @@
 "@ariakit/react": patch
 ---
 
-Support provider components in store props and store hooks
+Store props and store hooks accept provider components
 
 Component `store` props now accept a provider component (for example, [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider)) in place of a store object. The store is read from that provider's context, which lets you explicitly bind a component to a specific provider kind even when a different compatible provider is nested closer.
 
--- a/.changeset/200-provider-component-store.md
+++ b/.changeset/200-provider-component-store.md
@@ -1,0 +1,22 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react-store": patch
+"@ariakit/react-utils": patch
+"@ariakit/react": patch
+---
+
+Support provider components in store props and store hooks
+
+Component `store` props now accept a provider component (for example, [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider)) in place of a store object. The store is read from the closest matching provider through context, which lets you explicitly bind a component to a specific provider even when a different compatible provider is nested closer.
+
+For example, [`CompositeItem`](https://ariakit.com/reference/composite-item) can read from the closest [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider) even when a [`ToolbarProvider`](https://ariakit.com/reference/toolbar-provider) or [`CompositeProvider`](https://ariakit.com/reference/composite-provider) is nested closer to it:
+
+```tsx
+<CompositeItem store={ComboboxProvider} />
+```
+
+The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks accept provider components as well:
+
+```tsx
+const value = useStoreState(ComboboxProvider, "value");
+```
--- a/.changeset/300-composite-input-context-store.md
+++ b/.changeset/300-composite-input-context-store.md
@@ -1,0 +1,9 @@
+---
+"@ariakit/react-components": patch
+---
+
+`CompositeInput` reads the store from context
+
+The deprecated `CompositeInput` component now reads the store from the closest [`Composite`](https://ariakit.com/reference/composite) or [`CompositeProvider`](https://ariakit.com/reference/composite-provider) context when the `store` prop is not provided, as its documentation already stated.
+
+As a result, while the caret moves within the text field, `CompositeInput` now only stops the propagation of arrow keys along the composite widget's [`orientation`](https://ariakit.com/reference/use-composite-store#orientation). Arrow keys on the other axis, which the composite widget doesn't handle, are no longer stopped.
--- a/.changeset/300-composite-input-context-store.md
+++ b/.changeset/300-composite-input-context-store.md
@@ -2,7 +2,7 @@
 "@ariakit/react-components": patch
 ---
 
-`CompositeInput` reads the store from context
+`CompositeInput` now reads the store from context
 
 The deprecated `CompositeInput` component now reads the store from the closest [`Composite`](https://ariakit.com/reference/composite) or [`CompositeProvider`](https://ariakit.com/reference/composite-provider) context when the `store` prop is not provided, as its documentation already stated.
 
--- a/app/src/sandbox/composite-item-provider-store/index.react.tsx
+++ b/app/src/sandbox/composite-item-provider-store/index.react.tsx
@@ -1,0 +1,69 @@
+import * as Ariakit from "@ariakit/react";
+import "./style.css";
+
+// Reads state directly from provider components instead of store objects.
+// `useStoreState` resolves each provider component to the closest matching
+// provider through context, so this works without threading stores as props.
+function StoreStatus() {
+  const comboboxValue = Ariakit.useStoreState(
+    Ariakit.ComboboxProvider,
+    "value",
+  );
+  const comboboxItemCount = Ariakit.useStoreState(
+    Ariakit.ComboboxProvider,
+    (state) => state?.renderedItems.length ?? 0,
+  );
+  const toolbarItemCount = Ariakit.useStoreState(
+    Ariakit.ToolbarProvider,
+    (state) => state?.renderedItems.length ?? 0,
+  );
+  return (
+    <div className="status">
+      <output aria-label="Combobox value">{comboboxValue}</output>
+      <output aria-label="Combobox items">{comboboxItemCount}</output>
+      <output aria-label="Toolbar items">{toolbarItemCount}</output>
+    </div>
+  );
+}
+
+// Rendered outside any ComboboxProvider. Because a provider component is an
+// explicit reference with no fallback, `useStoreState` resolves to `undefined`
+// here even though a ComboboxProvider exists elsewhere in the tree.
+function OutsideStatus() {
+  const value = Ariakit.useStoreState(Ariakit.ComboboxProvider, "value");
+  return <output aria-label="Outside combobox value">{value ?? "none"}</output>;
+}
+
+export default function Example() {
+  return (
+    <>
+      <OutsideStatus />
+      <ComboboxWithToolbar />
+    </>
+  );
+}
+
+function ComboboxWithToolbar() {
+  return (
+    <Ariakit.ComboboxProvider defaultValue="Apple">
+      <Ariakit.ComboboxLabel>Fruit</Ariakit.ComboboxLabel>
+      <Ariakit.Combobox className="combobox" />
+      <Ariakit.ToolbarProvider>
+        <Ariakit.Toolbar className="toolbar" aria-label="Fruit actions">
+          {/* Binds to the nearest Toolbar, its closest composite context. */}
+          <Ariakit.ToolbarItem className="button">Clear</Ariakit.ToolbarItem>
+          {/* Explicitly binds to the Combobox by passing the provider
+              component, even though the Toolbar is the closer composite
+              context. It becomes a combobox item, not a toolbar item. */}
+          <Ariakit.CompositeItem
+            store={Ariakit.ComboboxProvider}
+            className="button"
+          >
+            Focus first option
+          </Ariakit.CompositeItem>
+        </Ariakit.Toolbar>
+        <StoreStatus />
+      </Ariakit.ToolbarProvider>
+    </Ariakit.ComboboxProvider>
+  );
+}
--- a/app/src/sandbox/composite-item-provider-store/index.react.tsx
+++ b/app/src/sandbox/composite-item-provider-store/index.react.tsx
@@ -1,17 +1,18 @@
 import * as Ariakit from "@ariakit/react";
+import { useStoreStateObject } from "@ariakit/react-components/store";
 import "./style.css";
 
 // Reads state directly from provider components instead of store objects.
-// `useStoreState` resolves each provider component to the closest matching
-// provider through context, so this works without threading stores as props.
+// `useStoreState` and `useStoreStateObject` resolve each provider component
+// through that provider's context, so this works without threading stores as
+// props.
 function StoreStatus() {
-  const comboboxValue = Ariakit.useStoreState(
+  const { comboboxValue, comboboxItemCount } = useStoreStateObject(
     Ariakit.ComboboxProvider,
-    "value",
-  );
-  const comboboxItemCount = Ariakit.useStoreState(
-    Ariakit.ComboboxProvider,
-    (state) => state?.renderedItems.length ?? 0,
+    {
+      comboboxValue: "value",
+      comboboxItemCount: (state) => state?.renderedItems.length ?? 0,
+    },
   );
   const toolbarItemCount = Ariakit.useStoreState(
     Ariakit.ToolbarProvider,
@@ -39,6 +40,7 @@ export default function Example() {
     <>
       <OutsideStatus />
       <ComboboxWithToolbar />
+      <OutsideCollection />
     </>
   );
 }
@@ -59,11 +61,45 @@ function ComboboxWithToolbar() {
             store={Ariakit.ComboboxProvider}
             className="button"
           >
-            Focus first option
+            Combobox item
           </Ariakit.CompositeItem>
         </Ariakit.Toolbar>
         <StoreStatus />
       </Ariakit.ToolbarProvider>
     </Ariakit.ComboboxProvider>
+  );
+}
+
+// Rendered outside any ComboboxProvider. The second CollectionItem explicitly
+// targets ComboboxProvider, and provider components have no fallback, so it
+// must not register with the closer collection even though that compatible
+// context is right above it.
+function OutsideCollection() {
+  return (
+    <Ariakit.CollectionProvider>
+      <Ariakit.Collection className="collection">
+        <Ariakit.CollectionItem className="button" render={<button />}>
+          Plain item
+        </Ariakit.CollectionItem>
+        <Ariakit.CollectionItem
+          className="button"
+          render={<button />}
+          store={Ariakit.ComboboxProvider}
+        >
+          Outside combobox item
+        </Ariakit.CollectionItem>
+      </Ariakit.Collection>
+      <OutsideCollectionStatus />
+    </Ariakit.CollectionProvider>
+  );
+}
+
+function OutsideCollectionStatus() {
+  const collectionItemCount = Ariakit.useStoreState(
+    Ariakit.CollectionProvider,
+    (state) => state?.renderedItems.length ?? 0,
+  );
+  return (
+    <output aria-label="Outside collection items">{collectionItemCount}</output>
   );
 }
--- a/app/src/sandbox/composite-item-provider-store/test.ts
+++ b/app/src/sandbox/composite-item-provider-store/test.ts
@@ -1,13 +1,13 @@
 import { q, type } from "@ariakit/test";
 import { expect, test } from "vitest";
 
-test("useStoreState reads state through a provider component", async () => {
+test("useStoreStateObject reads state through a provider component", async () => {
   await expect
     .poll(() => q.status.ensure("Combobox value").textContent)
     .toBe("Apple");
 
   // Typing updates the combobox store, and the readout reflects it live
-  // through the provider component passed to `useStoreState`.
+  // through the provider component passed to `useStoreStateObject`.
   await type("s", q.combobox.ensure("Fruit"));
 
   await expect
@@ -25,7 +25,7 @@ test("provider components have no fallback: useStoreState is undefined outside a
 });
 
 test("CompositeItem store={ComboboxProvider} binds to the combobox, not the closer toolbar", async () => {
-  // The `Focus first option` item is nested inside the Toolbar (its closest
+  // The `Combobox item` item is nested inside the Toolbar (its closest
   // composite context) but explicitly targets the ComboboxProvider, so it
   // registers with the combobox instead of the toolbar.
   await expect
@@ -37,5 +37,17 @@ test("CompositeItem store={ComboboxProvider} binds to the combobox, not the clos
 
   // Both items render as buttons regardless of which store they belong to.
   expect(q.button("Clear")).toBeInTheDocument();
-  expect(q.button("Focus first option")).toBeInTheDocument();
+  expect(q.button("Combobox item")).toBeInTheDocument();
+});
+
+test("provider components have no fallback: CollectionItem store={ComboboxProvider} skips the closer collection outside a matching provider", async () => {
+  // The outside collection renders one plain CollectionItem plus a
+  // CollectionItem that targets ComboboxProvider. No ComboboxProvider wraps
+  // that region, and provider components have no fallback, so the second item
+  // must not register with the closer collection: the count stays at the
+  // plain item.
+  await expect
+    .poll(() => q.status.ensure("Outside collection items").textContent)
+    .toBe("1");
+  expect(q.button("Outside combobox item")).toBeInTheDocument();
 });
--- a/app/src/sandbox/composite-item-provider-store/test.ts
+++ b/app/src/sandbox/composite-item-provider-store/test.ts
@@ -1,0 +1,41 @@
+import { q, type } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+test("useStoreState reads state through a provider component", async () => {
+  await expect
+    .poll(() => q.status.ensure("Combobox value").textContent)
+    .toBe("Apple");
+
+  // Typing updates the combobox store, and the readout reflects it live
+  // through the provider component passed to `useStoreState`.
+  await type("s", q.combobox.ensure("Fruit"));
+
+  await expect
+    .poll(() => q.status.ensure("Combobox value").textContent)
+    .toBe("Apples");
+});
+
+test("provider components have no fallback: useStoreState is undefined outside a matching provider", async () => {
+  // `OutsideStatus` reads `useStoreState(ComboboxProvider, "value")` while
+  // rendered outside every ComboboxProvider, so it must resolve to undefined
+  // (shown as "none") even though a ComboboxProvider exists elsewhere.
+  await expect
+    .poll(() => q.status.ensure("Outside combobox value").textContent)
+    .toBe("none");
+});
+
+test("CompositeItem store={ComboboxProvider} binds to the combobox, not the closer toolbar", async () => {
+  // The `Focus first option` item is nested inside the Toolbar (its closest
+  // composite context) but explicitly targets the ComboboxProvider, so it
+  // registers with the combobox instead of the toolbar.
+  await expect
+    .poll(() => q.status.ensure("Combobox items").textContent)
+    .toBe("1");
+  await expect
+    .poll(() => q.status.ensure("Toolbar items").textContent)
+    .toBe("1");
+
+  // Both items render as buttons regardless of which store they belong to.
+  expect(q.button("Clear")).toBeInTheDocument();
+  expect(q.button("Focus first option")).toBeInTheDocument();
+});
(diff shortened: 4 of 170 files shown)


diegohaz (commented):
Initial review of `80ebc7d59ba82cdcc014624c16ca80e30347aed4` against merge base `b881df001341b7a29f1bfd24cb34b98ba4565909`.

Scope reviewed: the provider-component branding/resolution in `@ariakit/react-utils`, `useStoreState`/`useStoreStateObject` changes in `@ariakit/react-store`, representative component `store` prop plumbing across composed contexts, provider component type declarations, the legacy reference generator fallback, the new sandbox tests, docs, and the changeset.

Validation run:

- `pnpm test-react app/src/sandbox/composite-item-provider-store/test.ts`
- `pnpm test website/build-pages/reference-utils.test.ts`
- `pnpm tsc`
- `APP_PORT=4322 NEXTJS_PORT=4322 pnpm -F app run test app/src/sandbox/composite-item-provider-store/test-browser.ts`
- `pnpm build`
- `pnpm lint` after cleaning local build artifacts with `pnpm -r run --if-present clean`

I didn't find substantive issues worth inline comments. I specifically looked for provider-vs-compatible-context regressions, missing provider fallback behavior, store props leaking to DOM elements, public type/declaration breakage, avoidable repeated work, and reference-generation regressions.

packages/ariakit-react-components/src/composite/composite-input.tsx
```diff
@@ -49,6 +51,8 @@ function getValueLength(element: HTMLElement) {
  */
 export const useCompositeInput = createHook<TagName, CompositeInputOptions>(
   function useCompositeInput({ store, ...props }) {
+    const context = useCompositeScopedContext();
+    store = useStoreProp(store, context);
```
diegohaz, line 55: This is the only conversion in the PR that adds a fallback the base code didn't have (I scanned every `useStoreProp(store, ...)` site against `b881df001` to confirm). Before this change, `useCompositeInput` never read context: `store` was prop-only, so on the no-prop path `store?.getState()` was always `undefined`, both `isHorizontal` and `isVertical` were true, and all four arrow keys were selection-gated in `onKeyDownCapture`. Now the store resolves from the composite scoped context, so inside an `orientation="vertical"` composite, ArrowLeft/ArrowRight are no longer gated by caret position (and vice versa for horizontal).

The new behavior matches the pre-existing JSDoc (which the old code didn't implement) and only gates keys the composite would actually handle, so it looks like a deliberate fix. But it's a runtime behavior change on a shipped (if deprecated) component, and both the PR description ("This is an additive change") and the changeset present the PR as behavior-preserving.

Suggestion: either keep it and mention the `CompositeInput` behavior change in the changeset, or drop these two lines (and the new JSDoc paragraph) to keep the component prop-only and land the fix separately.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -913,10 +914,17 @@ export interface CollectionRendererOptions<
...
+   * another compatible store context is closer.
    */
-  store?: CollectionStore<
-    T extends CollectionStoreItem ? T : CollectionStoreItem
-  >;
+  store?:
+    | CollectionStore<T extends CollectionStoreItem ? T : CollectionStoreItem>
+    | ProviderComponent<
+        CollectionStore<T extends CollectionStoreItem ? T : CollectionStoreItem>
+      >;
```
diegohaz: Parameterizing the `ProviderComponent` branch over `T` breaks the provider-component feature whenever the renderer's item type is narrower than `CollectionStoreItem`, because `CollectionProviderComponent` is branded at the base `ProviderComponent<CollectionStore>`. I reproduced both failure modes with `tsc` against this branch:

```tsx
declare const myItems: Array<{ id: string; label: string }>;

// T inference is poisoned by the store branch: `item` degrades to
// `BaseItemProps & CollectionStoreItem`, so `item.label` errors.
// (Without the `store` prop, the same JSX compiles.)
<CollectionRenderer items={myItems} store={CollectionProvider}>
  {(item) => <div key={item.id}>{item.label}</div>}
</CollectionRenderer>;

// Explicit generic is rejected outright:
// TS2322: Type 'CollectionProviderComponent' is not assignable to
// 'ProviderComponent<CollectionStore<MyItem>>'.
<CollectionRenderer<MyItem> items={myItems} store={CollectionProvider}>
  {(item) => <div key={item.id}>{item.label}</div>}
</CollectionRenderer>;
```

Since the brand never carries item-type information and runtime resolution doesn't depend on `T`, the provider branch can use the base store, mirroring `useStoreProp`'s own "fallbacks are loosely typed so they never widen or poison that inference" design:

```diff
   store?:
     | CollectionStore<T extends CollectionStoreItem ? T : CollectionStoreItem>
-    | ProviderComponent<
-        CollectionStore<T extends CollectionStoreItem ? T : CollectionStoreItem>
-      >;
+    | ProviderComponent<CollectionStore>;
```

I validated this locally: both repro cases compile and repo-wide `pnpm tsc` still passes. The same pattern exists in `composite-renderer.tsx` (the `store` option around line 267) and, in milder form, `select-value.tsx:102` (`ProviderComponent<SelectStore<T>>` makes `<SelectValue<"a" | "b"> store={SelectProvider}>` fail to type-check; a base `ProviderComponent<SelectStore>` branch avoids that).

Related cleanup while here: `CollectionRendererContextValue["store"]` (line 119) reuses this option type, but the context value is only ever built from the resolved store (the `useStoreProp` result), so its `ProviderComponent` arm is impossible at runtime and can be dropped.

app/src/sandbox/composite-item-provider-store/test.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+test("provider components have no fallback: useStoreState is undefined outside a matching provider", async () => {
+  // `OutsideStatus` reads `useStoreState(ComboboxProvider, "value")` while
+  // rendered outside every ComboboxProvider, so it must resolve to undefined
+  // (shown as "none") even though a ComboboxProvider exists elsewhere.
+  await expect
+    .poll(() => q.status.ensure("Outside combobox value").textContent)
+    .toBe("none");
+});
```
diegohaz, line 25: This covers the hook-path miss, but the distinguishing branch of the new machinery on the *component* path is never discriminated: `store={XProvider}` missing its provider while a compatible fallback context is present. A regression of `useStoreProp`'s provider branch from `return contextStore;` to something like `return contextStore ?? firstFallback;` would pass every test in this PR, silently restoring the implicit fallback the feature is designed to avoid.

Suggested minimal addition: render a second `ToolbarProvider` region outside the `ComboboxProvider` containing `<Ariakit.CompositeItem store={Ariakit.ComboboxProvider}>`, and assert it does not register with that toolbar (its toolbar item count stays at its own items).

On the same theme, a small `packages/ariakit-react-utils/src/system.test.ts` (there's package-level precedent in `misc.test.ts`) could cheaply pin `isProviderComponent` (branded vs plain function) and the multi-fallback ordering that `menu-item.tsx` relies on.

app/src/sandbox/composite-item-provider-store/index.react.tsx
```diff
@@ -0,0 +1,69 @@
...
+    "value",
+  );
+  const comboboxItemCount = Ariakit.useStoreState(
+    Ariakit.ComboboxProvider,
+    (state) => state?.renderedItems.length ?? 0,
+  );
+  const toolbarItemCount = Ariakit.useStoreState(
+    Ariakit.ToolbarProvider,
+    (state) => state?.renderedItems.length ?? 0,
+  );
```
diegohaz: The changeset advertises that `useStoreStateObject` accepts provider components too, but nothing in the PR exercises that path. It's a one-line change to cover it here, for example:

```tsx
const { comboboxValue, comboboxItemCount } = Ariakit.useStoreStateObject(
  Ariakit.ComboboxProvider,
  {
    comboboxValue: "value",
    comboboxItemCount: (state) => state?.renderedItems.length ?? 0,
  },
);
```

app/src/sandbox/composite-item-provider-store/index.react.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Ariakit.ToolbarItem className="button">Clear</Ariakit.ToolbarItem>
+          {/* Explicitly binds to the Combobox by passing the provider
+              component, even though the Toolbar is the closer composite
+              context. It becomes a combobox item, not a toolbar item. */}
+          <Ariakit.CompositeItem
+            store={Ariakit.ComboboxProvider}
+            className="button"
+          >
+            Focus first option
+          </Ariakit.CompositeItem>
```
diegohaz, line 65: The label promises behavior the fixture doesn't model: the item implements no focus action and the sandbox renders no options. Renaming it to something neutral (for example, "Combobox item") or actually implementing the action would keep the sandbox aligned with the realistic-scenario bar and avoid future readers hunting for a missing click handler.

packages/ariakit-react-store/src/index.tsx
```diff
@@ -2,8 +2,12 @@ import {
   useEvent,
   useLiveRef,
   useSafeLayoutEffect,
+  useStoreProp,
 } from "@ariakit/react-utils";
+import type { ProviderComponent } from "@ariakit/react-utils";
 import { batch, init, subscribe, sync } from "@ariakit/store";
+
+export type { ProviderComponent } from "@ariakit/react-utils";
```
diegohaz: `ProviderComponent` now appears in every public component `store` prop and is accepted by `useStoreState`, which *is* re-exported from `@ariakit/react` — but the type itself isn't reachable from the main package. Consumers typing their own wrappers have to use indexed access (`ComboboxProps["store"]`) or add a direct dependency on `@ariakit/react-store`/`@ariakit/react-utils`.

I validated locally that a one-liner in `packages/ariakit-react/src/store.ts` compiles and makes `Ariakit.ProviderComponent<ComboboxStore>` usable:

```ts
export type { ProviderComponent } from "@ariakit/react-components/store";
```

Worth a deliberate decision either way (the same question applies to the new `XProviderComponent` interfaces, though `typeof XProvider` covers those). Minor note on this line itself: the `export type` sits between two import statements; moving it below the import block would read more conventionally.

.changeset/200-provider-component-store.md
```diff
@@ -0,0 +1,22 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react-store": patch
+"@ariakit/react-utils": patch
+"@ariakit/react": patch
+---
+
+Support provider components in store props and store hooks
+
+Component `store` props now accept a provider component (for example, [`ComboboxProvider`](https://ariakit.com/reference/combobox-provider)) in place of a store object. The store is read from the closest matching provider through context, which lets you explicitly bind a component to a specific provider even when a different compatible provider is nested closer.
```
diegohaz: Widening `store?: XStore` to `XStore | ProviderComponent<XStore>` is additive for callers, but it's a type-level break for consumers that read `store` off Ariakit prop types and pass it to APIs that still expect a store object. This PR itself demonstrates the pattern: `examples/form-callback-queue/index.react.tsx` had to re-narrow with `store?: Ariakit.FormStore` because `useFormValidate(store, ...)` no longer accepts the widened prop. Since that pattern ships in a published example, downstream copies will hit the same error on upgrade.

Suggest documenting it here with the narrowing recipe (redeclare `store` in the extending interface, or resolve it through the component before use), or deciding whether it warrants a **BREAKING** section per our conventions (in v0 that would mean `minor` frontmatter). A short paragraph is probably enough given how narrow the affected pattern is.

packages/ariakit-react-utils/src/system.tsx
```diff
@@ -88,6 +88,109 @@ type StoreProvider<T extends Store> = React.ComponentType<{
...
+
+/**
+ * Resolves the value of a `store` prop that may receive either a store object
+ * or a provider component. When a provider component is passed, it's an
+ * explicit reference to that specific provider: the store is read only from
+ * the closest matching provider's context, without falling back to the given
+ * fallback stores, so it may be `undefined` if no matching provider is found.
+ * Otherwise, the store prop itself is returned, falling back to the given
+ * fallback stores (typically the component's own context) when it's not
+ * provided.
```
diegohaz, line 170: The "read only from the closest matching provider's context ... may be `undefined` if no matching provider is found" wording (and the per-component paragraph, "the store is read from the closest matching provider, even if another compatible store context is closer") is looser than the actual semantics, because the unscoped family context is written by more parties than matching providers:

- Extending providers write base-family contexts: `popover-context.tsx` declares `createStoreContext<PopoverStore>([DialogContextProvider], ...)`, so in `<DialogProvider store={A}><PopoverProvider store={P}><DialogDisclosure store={DialogProvider} /></PopoverProvider></DialogProvider>` the disclosure resolves `P`, not `A`. The no-cross-family guarantee holds in the direction the PR motivates (a more specific provider is never shadowed by a base one), but targeting a *base* provider still resolves the nearest extending provider.
- Widget containers write it too: `ScopedContextProvider` wraps `ContextProvider` (lines 229-242), and `Dialog` renders `DialogScopedContextProvider value={store}` (dialog.tsx:592), so `<Dialog store={B}><DialogDismiss store={DialogProvider} /></Dialog>` resolves `B` with no `DialogProvider` mounted at all — not `undefined`.

If these are the intended semantics (and they seem consistent with how the context graph works), consider rewording here and in the shared component paragraph to something like "the store is read from the closest context of that provider's kind — set by that provider, an extending provider, or a compatible container component — skipping less specific store contexts". If exact-provider semantics were intended instead, resolution would need a provider-only context, which is a bigger change. Docs-only either way; behavior looks deliberate.

diegohaz (commented):
Re-review of `80ebc7d59ba82cdcc014624c16ca80e30347aed4`. The head is unchanged since the previous marked review (`pr-6671-80ebc7d59ba82cdcc014624c16ca80e30347aed4-20260707T153854Z`), so this run re-reviewed the full diff from merge base `b881df001341b7a29f1bfd24cb34b98ba4565909` at greater depth rather than an incremental range: per-family sweeps over all component conversions (fallback order and context-hook parity against the base, JSDoc conventions for resolving vs inert store props, provider overload preservation, DOM-leak paths) plus type-level repros against the public API surface.

**Validation run:**

- `pnpm tsc` (repo-wide, pass), `pnpm test-react app/src/sandbox/composite-item-provider-store/test.ts` (3/3), `pnpm -F app run test-chrome` on the sandbox browser suite against a local server (3/3), `pnpm test website/build-pages/reference-utils.test.ts` (4/4), `pnpm test packages/ariakit-scripts/src/docs.test.ts` (34/34).
- Empirically confirmed the new reference-utils test is discriminating: it fails 1/4 against the pre-fix `reference-utils.js` and passes at head.
- Type-level scratch compilations against this branch: provider components accepted in the compatible direction (`CompositeItem store={ComboboxProvider}`, `MenuItem store={MenubarProvider}`), incompatible providers and arbitrary functions rejected, provider-derived `useStoreState` values correctly `| undefined` while store-object values stay definite; reproduced the `CollectionRenderer` generic issue and validated the suggested fix (repo-wide `pnpm tsc` passes with it); validated the suggested `ProviderComponent` re-export from `@ariakit/react`.
- Changeset reference links verified against ariakit.com (all 200; `use-store-state-object` is 404 and correctly left unlinked). A systematic scan confirmed `composite-input.tsx` is the only conversion that adds a fallback the base didn't have. CI is fully green, including both the Chrome and Node perf sections.

**Findings (8 inline comments):** one silent runtime behavior change (`CompositeInput` gaining a context fallback), one type defect class (`T`-parameterized `ProviderComponent` branches poisoning inference in `CollectionRenderer`/`CompositeRenderer` and rejecting explicit generics; milder variant in `SelectValue`), two coverage gaps (component-path strict miss undiscriminated; `useStoreStateObject` untested), a changeset note on the type-level break for consumers reading `store` off prop types (demonstrated by this PR's own `form-callback-queue` fix), the `ProviderComponent` export surface from `@ariakit/react`, a docs-accuracy suggestion on the "closest matching provider" wording, and a fixture label nit.

**Reviewed and accepted without comment:** invariant messages now also fire when a provider component was passed but unmatched (wording is slightly stale but still actionable, and the PR states invariants are meant to fire); the `Dialog` wrapper branching on raw `props.store` truthiness (`dialog.tsx:685`) — I traced all paths and only the unresolved-provider misuse case changes (the invariant is deferred until the dialog would actually mount), while resolving before branching would silently create an internal store and swallow the designed invariant, so the current code is preferable; layered wrappers (for example `SelectPopover`) re-resolving their own contexts after a provider miss, which mirrors the pre-existing double-fallback layering; and the known limitation that explicitly narrowed store generics don't accept widest-typed providers, matching store-object invariance.

**Not fully validated:** `@ariakit/solid` parity is out of scope (React-only PR); browser behavior beyond the Chrome run and CI's Firefox/Safari/Android jobs was not exercised locally.

packages/ariakit-react-components/src/composite/composite-input.tsx
diegohaz: Probably keep it.

diegohaz: Keeping it, per the follow-up. Documented in 3aaf3456b with a dedicated `300-composite-input-context-store.md` changeset describing the new context fallback and the orientation-aware arrow key gating on the deprecated `CompositeInput`, so the main entry stays behavior-preserving.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in 646b129de - the provider arms of `CollectionRendererOptions["store"]`, `CompositeRendererOptions["store"]`, and `SelectValueProps["store"]` now use base stores, and `CollectionRendererContextValue["store"]` excludes the impossible `ProviderComponent` arm. Applying this surfaced one more inference gap: `useCompositeRenderer` reassigns the resolved store and forwards it to `useCollectionRenderer`, and a base-typed resolution no longer satisfied the generic option there. Fixed at the machinery level instead: `useStoreProp`'s public overload now types the provider arm as plain `ProviderComponent`, so `T` is inferred only from the store object arm, matching the documented "fallbacks are loosely typed" design. The repro cases are pinned as compile-time coverage in `packages/ariakit-react-components/src/provider-component.test.tsx` (red-checked: reverting the type changes makes `tsc` fail with the five expected errors). Repo-wide `pnpm tsc` passes.

app/src/sandbox/composite-item-provider-store/test.ts
diegohaz: Fixed in caa254692 - added `packages/ariakit-react-utils/src/system.test.tsx` pinning `isProviderComponent` (branded function vs plain function vs store object vs null/undefined) and `useStoreProp` (store prop wins, first-truthy multi-fallback ordering, provider resolution ignoring fallbacks, extending-provider resolution, and strict miss returning `undefined` with truthy fallbacks present), plus a component-path sandbox test mirrored in `test-browser.ts`. Red check: swapping the provider branch to `contextStore ?? firstTruthyFallback` fails exactly the two new strict-miss tests and nothing else.

One discovery: the suggested `ToolbarProvider` + `CompositeItem store={ComboboxProvider}` scenario fails on the current code - the outside toolbar count becomes 2. `useCompositeItem` strictly resolves the provider to `undefined`, then passes that to `useCollectionItem`, which re-resolves from its own collection context and registers with the toolbar. So the no-fallback guarantee currently holds per hook layer, not per composed component (same pattern in every layered component, for example `SelectPopover` -> `useSelectList`/`usePopover`). The sandbox covers the component path with `CollectionItem` instead, whose registration lives in the same hook that resolves the prop, which discriminates the `useStoreProp` regression as intended. Threading the raw provider prop through layered hooks (so strict misses propagate per component) is a wider design change - left open for a follow-up decision.

app/src/sandbox/composite-item-provider-store/index.react.tsx
diegohaz: Fixed in caa254692 - the sandbox now reads the combobox value and item count through `useStoreStateObject(ComboboxProvider, ...)` (imported from `@ariakit/react-components/store`, since the hook isn't re-exported from `@ariakit/react`), and the live-update test exercises that path. Red check: disabling provider resolution inside `useStoreStateObject` fails the test.

app/src/sandbox/composite-item-provider-store/index.react.tsx
diegohaz: Fixed in caa254692 - renamed to the neutral "Combobox item" in the fixture and both test files.

packages/ariakit-react-store/src/index.tsx
diegohaz: Fixed in ff159f168 - `@ariakit/react`'s store entry now re-exports `ProviderComponent`, and the changeset mentions it with a wrapper-typing example (6d92e3502). The stray `export type` between imports in `@ariakit/react-store` moved below the import block in c9508e3df.

.changeset/200-provider-component-store.md
diegohaz: Fixed in 6d92e3502 - added a paragraph stating this is a type-level change and showing the re-narrowing recipe used by `form-callback-queue` (`store?: Ariakit.FormStore` in the extending interface). Kept `patch` frontmatter instead of a **BREAKING** section since the affected pattern is narrow and compile-time only; happy to promote it if you'd rather flag it louder.

packages/ariakit-react-utils/src/system.tsx
diegohaz: Fixed in c9508e3df - the shared paragraph on all 86 widened store props, plus the `ProviderComponent`, `useStoreProp`, `useStoreState`, and `useStoreStateObject` JSDoc, now read "the closest context of that provider's kind (set by that provider, an extending provider, or a compatible container component), skipping less specific store contexts", and the readmes were regenerated. The extending-provider path is now pinned by a unit test in `system.test.tsx` (caa254692), and the changeset wording was aligned in 6d92e3502.

diegohaz:
Addressed all 8 unresolved review threads from the deep re-review. 6 commits pushed (`80ebc7d59..6d92e3502`); repo-wide `pnpm tsc`, `pnpm test-react`/`test-react18`, extractor tests, and `pnpm build-website-lite` all pass.

**Fixed**

- 646b129de ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555775910)): renderer/`SelectValue` provider arms now use base stores so they no longer poison `T` inference or reject explicit generics; `useStoreProp` infers only from the store object arm; repros pinned as compile-time coverage in `provider-component.test.tsx` (red-checked).
- c9508e3df ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776422)): reworded the "closest matching provider" semantics across all 86 widened store props and the store hook JSDoc ("closest context of that provider's kind..."); readmes regenerated; misplaced `export type` in `@ariakit/react-store` moved below imports.
- ff159f168 ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776245)): `ProviderComponent` is now re-exported from `@ariakit/react`.
- caa254692 ([strict-miss coverage](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776016), [useStoreStateObject coverage](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776084), [label rename](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776165)): new `system.test.tsx` unit suite (fallback ordering, provider resolution, extending providers, strict miss; red-checked against a simulated `contextStore ?? fallback` regression), `useStoreStateObject(ComboboxProvider, ...)` exercised in the sandbox, component-path strict-miss test via `CollectionItem`, and the "Focus first option" label renamed to "Combobox item".
- 3aaf3456b ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555775819)): kept the `CompositeInput` context fallback per the thread follow-up and documented it in a dedicated changeset.
- 6d92e3502 ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776348)): main changeset now notes the type-level `store` prop widening with the re-narrowing recipe, mentions the `ProviderComponent` export, and aligns the resolution wording.

**Follow-up**

- ([details](https://github.com/ariakit/ariakit/pull/6671#discussion_r3555776016)): the strict no-fallback guarantee currently holds per hook layer, not per composed component. `CompositeItem store={ComboboxProvider}` outside a matching provider still registers with a closer toolbar because `useCollectionItem` re-resolves from its own context; the same applies to every layered component. Propagating strict misses by threading the raw provider prop through layered hooks is a wider design change left open for a decision.

app/src/sandbox/composite-item-provider-store/test.ts
diegohaz: Verified the shipped mitigation on `8516a24a`: the `CollectionItem`-based sandbox test does discriminate the machinery regression (registration lives in the same hook that resolves the prop), and `system.test.tsx` pins the strict miss with truthy fallbacks present plus the extending-provider path. I also independently re-ran the red check on the compile-time coverage — reverting the four type changes produces exactly the five expected errors.

On the per-layer discovery: agreed that threading the raw provider prop through layered hooks is a wider design change and reasonable to defer. Two suggestions before merge so the decision isn't silently shipped and forgotten:

1. Track it in a follow-up issue describing the composed-hook leak: `useCompositeItem` resolves `store={ComboboxProvider}` strictly to `undefined`, but `useCollectionItem` then re-resolves from its own collection context and registers with the closer toolbar, so the item still lands in the toolbar store's `renderedItems`, which arrow-key navigation consumes. That makes the leak user-visible beyond counts.
2. Consider one caveat sentence where users will hit it. The reworded store-prop JSDoc ("skipping less specific store contexts") is accurate for the prop's own resolution, but a reader will expect the whole component to stay out of the closer toolbar in the miss case, and today it doesn't. A short note in the changeset or the `ProviderComponent` JSDoc — composed components may still bind lower-level behavior to their own contexts when the provider has no match — would set expectations until the follow-up lands.

Not blocking: invariant-bearing components still fail fast, and the affected path is a misuse case.

diegohaz (commented):
Re-review of `8516a24ac50485d4c370c26bb4e588906c6bd919`, incremental against the last marked review's head `80ebc7d59ba82cdcc014624c16ca80e30347aed4`. Scope: the six fix commits (`646b129de..6d92e3502`) responding to the 8 threads from that review, plus the `origin/main` merge commit.

**All 8 claimed fixes verified in the current code, not just from the replies:**

- `CompositeInput` context fallback kept and documented in the dedicated `300-composite-input-context-store.md` changeset; the main entry stays behavior-preserving (thread on `composite-input.tsx`).
- Renderer/`SelectValue` provider arms are base-typed with rationale comments, `CollectionRendererContextValue["store"]` excludes the impossible `ProviderComponent` arm, and `useStoreProp`'s public overload types the provider arm as plain `ProviderComponent`. I independently red-checked `provider-component.test.tsx`: reverting the four type changes yields exactly the five expected errors.
- `system.test.tsx` pins `isProviderComponent`, fallback ordering, provider resolution, extending-provider resolution, and the strict miss with truthy fallbacks; the sandbox adds a `CollectionItem`-based component-path strict-miss test mirrored in `test-browser.ts`.
- The sandbox exercises `useStoreStateObject(ComboboxProvider, ...)`; the fixture label was neutralized.
- `ProviderComponent` is re-exported from `@ariakit/react` (verified importable via a compile scratch) and the stray export moved below the import block.
- The changeset documents the type-level `store` prop widening with the re-narrowing recipe, and the "closest context of that provider's kind" rewording is applied consistently — no stale "closest matching provider" phrasing remains anywhere; readmes regenerated (docs sync tests pass).

**Merge commit check:** only main-side changes came in (renovate bumps and a `@template` JSDoc simplification in `reference-utils.js` from the TS 7.0.2 update); the PR's `${name}Props` fallback survived the merge and the reference-utils tests still pass.

**Validation run at head:** repo-wide `pnpm tsc`; `pnpm test-react` on `system.test.tsx` + `provider-component.test.tsx` + the sandbox suite (11/11); `reference-utils.test.ts` (4/4) and `docs.test.ts` (34/34) post-merge; a public-API type scratch confirming the `useStoreProp` overload loosening didn't weaken the surface (wrong-direction providers still rejected at the prop level, `useStoreState` key/value precision intact); lockfile diff limited to the new exact-versioned devDependencies and their peer relinks; CI fully green (44 checks).

**New findings: none.** The overload loosening is sound because component prop types still enforce provider compatibility and the store hooks keep their precise `ProviderComponent<T>` overloads.

**Open item:** the per-layer strict-miss propagation (composed hooks re-resolving their own contexts after a provider miss) remains deliberately deferred; I replied on that thread with verification results and a suggestion to track it in a follow-up issue plus a one-sentence doc caveat. Not blocking.

.changeset/200-provider-component-store.md
```diff
@@ -0,0 +1,39 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react-store": patch
+"@ariakit/react-utils": patch
+"@ariakit/react": patch
+---
+
+Support provider components in store props and store hooks
```
diegohaz: Changeset headings should read more like section headings. This reads like a PR title.

.changeset/300-composite-input-context-store.md
```diff
@@ -0,0 +1,9 @@
+---
+"@ariakit/react-components": patch
+---
+
+`CompositeInput` reads the store from context
```
diegohaz: Is this better?

```suggestion
`CompositeInput` now reads the store from context
```

app/src/sandbox/composite-item-provider-store/test.ts
diegohaz: Done - opened #6690 tracking the composed-hook leak (including the `renderedItems`/arrow-key visibility and the raw-prop threading direction), and added the caveat in 93c245795: a paragraph on the `ProviderComponent` JSDoc (which feeds both package readmes) plus a sentence in the changeset - if no matching context exists the store resolves to `undefined`, and components composed from multiple layers may still bind lower-level behavior to their own contexts. Kept it off the 86 per-prop JSDoc paragraphs to avoid another mechanical sweep; the type JSDoc and the changeset are the shared surfaces adopters read, and #6690 is the durable record for the underlying behavior.

.changeset/200-provider-component-store.md
diegohaz: Fixed in 01280a3c2 - retitled to "Store props and store hooks accept provider components".

.changeset/300-composite-input-context-store.md
diegohaz: Fixed in 01280a3c2 - applied the suggestion.

diegohaz:
Addressed the 3 unresolved threads from the verification re-review. 2 commits pushed (`8516a24ac..01280a3c2`); readmes regenerated and the docs test passes.

**Fixed**

- 93c245795 ([reply](https://github.com/ariakit/ariakit/pull/6671#discussion_r3561759335)): the layered strict-miss behavior is now tracked in #6690, and a caveat paragraph landed on the `ProviderComponent` JSDoc (flowing into both package readmes) and in the changeset: on a miss the store resolves to `undefined`, and components composed from multiple layers may still bind lower-level behavior to their own contexts.
- 01280a3c2 ([200 title](https://github.com/ariakit/ariakit/pull/6671#discussion_r3561759447), [300 title](https://github.com/ariakit/ariakit/pull/6671#discussion_r3561759571)): changeset titles reworded as section headings - "Store props and store hooks accept provider components" and "`CompositeInput` now reads the store from context" (suggestion applied).

**Follow-up**

- #6690: whether to thread the raw provider prop through layered hooks so strict misses propagate per component rather than per hook layer.

diegohaz:
Closing this without merging.

The main motivation for this feature was to offer an explicit migration path for trees that rely on components binding to other components' providers, like a `CompositeItem` resolving its store from an ancestor `ToolbarProvider` or `ComboboxProvider`. That only becomes truly valuable if we stop providing lower-level contexts (like the composite context) from higher-level providers. That's a separate design decision we haven't committed to yet, so shipping this machinery now would lock in an API before that question is settled.

Meanwhile, the feature turned out to be more complex than its value justifies. The strict, no-fallback guarantee doesn't hold for composed components (#6690): on a strict miss, inner hooks re-resolve from their own contexts, and fixing that means threading raw provider props through every layered hook call, a wide and permanent tax on every future composed component. It also permanently widens every `store` prop from `Store` to `Store | ProviderComponent<Store>`.

There are also simpler answers to the underlying need:

- Explicit targeting is already possible in userland by passing the context store directly, for example `<CompositeItem store={useComboboxContext()} />` or a `useComboboxStore()` result.
- When people reach for lower-level components inside higher-level ones, like `CompositeItem` within a combobox, that's a signal to identify the pattern and support it with first-class components, either new components or improvements to existing ones like `ComboboxItem`. That's how composite scoping was addressed in #3799, and #4851 is an example of the kind of real-world composition we should design for directly.

If we do decide to stop providing lower-level contexts from higher-level providers, we can revisit whether an explicit provider-targeting API like this is the right migration path at that point.

PR #6687, opened 2026-07-10
Improve Tailwind layer rendering performance

diegohaz:
## Motivation

Every element using `ak-layer` currently evaluates the parent-relative
`ak-layer-contrast` lightness path, even though only elements with the contrast
utility consume it.

## Solution

Move the parent-relative lightness calculation and its scratch values into the
static `ak-layer-contrast` utility. Ordinary layers now use a short,
non-inherited resolved-lightness fallback, while contrast layers preserve the
existing active and inactive calculations.

Add focused browser coverage for unmatched parent lightness and ordinary pushed
layers nested under a contrast ancestor. Regenerate the distributed CSS and add
a patch changeset.

## Performance

Fresh unprofiled 10-sample baseline/current comparison:

- Page load: 3480.8 ms → 3339.7 ms (-4%)
- Toggle layer and text classes: 1752.6 ms → 1661.7 ms (-5%)

These directional improvements are below the repository's 10% publication
threshold, so `perf-compare` correctly reports no significant change.

## Testing

- `pnpm lint-fix`
- `APP_PORT=4322 NEXTJS_PORT=4322 pnpm -F app test-chrome ariakit-tailwind`
- `pnpm build`
- `pnpm tsc`

### DIFF
--- a/.changeset/200-tailwind-layer-contrast-performance.md
+++ b/.changeset/200-tailwind-layer-contrast-performance.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/tailwind": patch
+---
+
+Improved layer rendering performance by only calculating parent-relative contrast for elements using the `ak-layer-contrast` utility.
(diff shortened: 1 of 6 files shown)


diegohaz:
Not enough improvement

PR #6688, opened 2026-07-10
Improve dialog performance

diegohaz:
## Motivation

`Dialog` used React state solely to remember whether it had opened. Updating that state added commits to every open/close cycle, including components built on Dialog such as Popover and Menu.

```tsx
const [hasOpened, setHasOpened] = useState(false);
```

Focus restoration still has to behave correctly when React replays effects, hides content with Activity, replaces a keyed dialog, or reopens before a deferred focus retry. Separately, `isTabbable` performed a visibility check before rejecting elements with a negative `tabindex`.

## Solution

Track the Dialog open lifecycle with refs instead of render-triggering state. Document-scoped focus sessions coordinate effect replay and keyed replacements, while generation checks and pending-session invalidation prevent stale animation-frame retries from moving focus after a dialog reopens or is superseded.

The behavior matrix lives in an app sandbox, including React 19 Activity, StrictMode, non-portaled and portaled replacement, disabled restoration, and deferred-retry cases. No React tests were added to `ariakit-react-components`.

`isTabbable` now checks negative `tabindex` before the visibility-dependent focusability path:

```ts
if (hasNegativeTabIndex(element)) return false;
if (!isFocusable(element)) return false;
```

## Performance

Ten samples against the same-runner `origin/main` baseline:

| Interaction | Scripting | Total |
| --- | --- | --- |
| Open dialog | 71ms → 66.8ms (-6%) | 114ms → 109.2ms (-4%) |
| Close dialog | 128.8ms → 121ms (-6%) | 169.1ms → 160.4ms (-5%) |

No significant regression was detected.

## Verification

- `pnpm lint`
- `pnpm tsc`
- `pnpm build`
- `pnpm test` — 564 tests
- Focused React 19 tests — 51 passed
- Focused React 18 tests — 49 passed, 2 Activity tests skipped
- Chrome, Firefox, and Safari sandbox tests — 24 passed

### DIFF
--- a/.changeset/200-dialog-focus-performance.md
+++ b/.changeset/200-dialog-focus-performance.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved [`Dialog`](https://ariakit.com/reference/dialog) performance by avoiding extra renders when tracking focus restoration. Focus is no longer moved back to the disclosure after the dialog has already reopened or been replaced. This also benefits components built on it, such as [`Popover`](https://ariakit.com/reference/popover) and [`Menu`](https://ariakit.com/reference/menu).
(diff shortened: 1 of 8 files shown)


diegohaz:
Not enough improvement

PR #6689, opened 2026-07-10
Improve menu item performance

diegohaz:
## Motivation

Each `MenuItem` subscribed to every `Menu` store update only to derive its role from the menu content element. In a 200-item menu, this duplicated callback work throughout keyboard navigation.

## Solution

Observe the resolved `MenuList` role once and share it with matching items through context. Keep the content-element fallback for hooks composed outside the list and items bound to an explicitly different store.

Track role mutations and host replacement while handling mount-time layout effects and redundant same-value writes. Add happy-dom and cross-browser regression coverage plus a patch changeset.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build`
- 187 menu-related tests on React 19 and 187 on React 18
- `menu-item-role` browser test in Chrome, Firefox, and Safari
- Paired `menu-perf` benchmark: no statistically significant end-to-end change
- Instrumentation: 200 fewer all-state subscriptions and 79,600 fewer callbacks during a 199-step traversal of 200 items

### DIFF
--- a/.changeset/200-menu-item-performance.md
+++ b/.changeset/200-menu-item-performance.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved performance of [`Menu`](https://ariakit.com/reference/menu) components with many [`MenuItem`](https://ariakit.com/reference/menu-item) components.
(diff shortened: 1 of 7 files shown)


diegohaz:
Not enough improvement

PR #6694, opened 2026-07-11
Improve Tab performance

diegohaz:
## Motivation

The 200-item Tab fixture exposed repeated work across Tab and shared primitives:

1. Standalone `TabPanel` stores created an inert merged Disclosure parent.
2. Synchronized Tab movement rescanned rendered items after `activeId` already matched `selectedId`.
3. Ordinary Tabs retained composed-widget-only subscriptions and lookups.
4. Explicitly associated `TabPanel` elements still subscribed to the panel collection and allocated scroll-position maps eagerly.
5. Sequential Composite navigation repeatedly searched rendered items from the beginning.
6. Large Collection registration repeatedly scanned growing arrays for absent IDs.
7. `CompositeStore.last()` copied and reversed the rendered items.
8. Composite row grouping became quadratic in large many-row layouts.
9. Default native button wrappers waited until after mount to apply `type="button"`, so initial refs and server markup could observe incomplete attributes.
10. Native submit controls could retain Ariakit's generated `data-focus-visible` state after `focusable` became `false`.

## Solution

Disclosure stores now omit an undefined parent, preserving `createStore`'s zero-parent update path.

Synchronized Tab selection returns before scanning when the active and selected IDs already match:

```ts
const { activeId, renderedItems } = tab.getState();
if (activeId === state.selectedId) return;
```

Tab composed-registration state uses one optional keyed subscription. `TabPanel` avoids its collection subscription when an explicit `tabId` is available and creates its scroll-position map only when restoration is needed.

Large Collection arrays use identity-validated ID caches for absence checks while existing IDs retain the original `findIndex` behavior. Sequential Composite navigation uses registered item identity when safe and falls back to the original ID scan for cloned, external, or stale arrays. Last-item lookup now scans backward without allocating a reversed copy.

Composite row grouping keeps a small linear path below 48 items and uses a hybrid array/`Map` path for larger layouts, preserving row order and unusual row IDs.

Default native button wrappers apply `type="button"` before their component hook when the hook preserves the chosen host:

```tsx
const htmlProps = useTab(withDefaultButtonType(props));
return createElement(TagName, htmlProps);
```

`MenuButton` remains the host-changing exception: nested menu buttons can become `Role.div`, so it applies the helper after the hook has selected the final host. Explicit submit/reset types and custom renders remain unchanged.

`Focusable` and `Command` reconcile custom host capabilities without per-render effects. Native submit controls retain the post-mount behavior required by React 19 form status, and `Focusable` now clears its generated `data-focus-visible` state when it becomes disabled or `focusable` becomes `false`.

Behavioral coverage lives in the `native-button-semantics` sandbox using `index.react.tsx`, `test.ts`, and `test-browser.ts`. It covers initial ref timing, explicit and custom types, nested menu hosts, SSR and hydration diagnostics, React 18, React 19 form status, focusability cleanup, and Chrome behavior.

## Performance and validation

The final [CI performance report](https://github.com/ariakit/ariakit/pull/6694#issuecomment-4942217323) found:

### Chrome

- Tab mount: scripting -13%, total -12%.
- Tab navigation: scripting +3%, total +3%, below the regression thresholds.
- Composite mount: scripting -17%, total -16%.
- Controlled Composite mount: scripting -21%, total -19%.
- No Chrome test crossed the regression thresholds.

### Node

- Large Collection registration: +48%.
- Disclosure store creation: +43%.
- Tab item rendering: +48%.
- Synchronized Tab movement: +26%.
- Sequential Composite movement: +331%.
- Repeated last-item lookup: +4,834%.
- Row grouping: +13% for the small control and +36% to +1,211% across large layouts.

Validation covers React 18 and 19, refs, SSR and hydration, custom hosts, React 19 form status, Chrome behavior, store tests and benchmarks, TypeScript, lint and formatting, Changesets, frozen installation, and the production app build.


diegohaz:
@codex

app/src/sandbox/tab-perf/react.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+  try {
+    const consoleError = vi
+      .spyOn(console, "error")
+      .mockImplementation(() => {});
+    let serverErrors: Array<unknown[]> = [];
+    try {
+      container.innerHTML = renderToString(element);
+      serverErrors = consoleError.mock.calls;
+    } finally {
+      consoleError.mockRestore();
```
diegohaz: [P2] Keep the console spy active through hydration

Restoring `console.error` here means this test observes only server-rendering warnings. React 19 reports attribute hydration mismatches through `console.error` and does not patch the mismatched attribute, so the server-rendered `type="button"` can remain in the DOM and make the assertions after `hydrateRoot` pass even if the client omitted it. I confirmed this with a focused React 19 repro: hydrating `<button>` over server markup `<button type="button">` emitted the mismatch while retaining the `type` attribute.

Please clear the spy after checking the server errors, keep it active during hydration, and assert that hydration emitted no errors before restoring it in the outer cleanup. For example:

```ts
expect(unexpectedServerErrors).toEqual([]);
consoleError.mockClear();

await act(async () => {
  root = hydrateRoot(container, element);
});
expect(consoleError).not.toHaveBeenCalled();
```

diegohaz (commented):
Initial review of the full merge-base diff from `99a5e987c8c09512137bfdb928491ac764e779ec` to `016493fdfe227a9902e8bf31fabb5ec9803de97b` (31 files). I reviewed the Collection, Composite, Disclosure, and Tab store changes; the React Command, Focusable, Tab, and TabPanel composition paths; surrounding call sites; behavior tests; benchmarks; package and TypeScript wiring; changesets; and the published performance report.

I left one inline P2 finding: the new hydration regression test restores its console spy before `hydrateRoot`, so React 19 can report a real attribute mismatch while the unchanged server attribute still lets the post-hydration DOM assertions pass. I found no separate product correctness, compatibility, security, or performance regressions. The more complex Collection cache, Composite identity lookup, and hybrid row-grouping paths retained their fallbacks and were justified by focused equivalence and timing checks.

Validation performed:

- `pnpm test packages/ariakit-components/src/collection/collection-store.test.ts packages/ariakit-components/src/composite/composite-store.test.ts packages/ariakit-components/src/disclosure/disclosure-store.test.ts` (23 passed)
- `pnpm test-react app/src/sandbox/tab-perf/react.test.tsx` (3 passed)
- `pnpm test-react18 app/src/sandbox/tab-perf/react.test.tsx` (3 passed)
- `pnpm tsc`
- All four new Node benchmark files, plus randomized equivalence checks covering Collection cache operations, Composite navigation, and row grouping
- A focused React 19 hydration repro confirming that a mismatch is logged without patching the server attribute
- Release-note reference URLs, current green checks, and the published paired performance report

I did not rerun the complete multi-round base/head browser and Node performance comparison or the full cross-browser suite locally; those areas were covered by the current green CI run and published paired results.

.changeset/200-collection-registration-performance.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/components": patch
```
diegohaz: I believe we could group some of these changesets into a single multiline changeset related to performance.

packages/ariakit-react-components/src/tab/tab.tsx
```diff
@@ -181,7 +182,9 @@ export const useTab = createHook<TagName, TabOptions>(function useTab({
  */
 export const Tab = memo(
   forwardRef(function Tab(props: TabProps) {
-    const htmlProps = useTab(props);
+    const htmlProps = useTab(
+      props.render ? props : { ...props, type: props.type ?? "button" },
```
diegohaz: Shouldn't this be done for all components that use Focusable and use button as `TagName`?

Also, shouldn't this be done inside the component hooks rather than in the components themselves? Or maybe not, since other components might use those hooks and specify a different tag name?

app/src/sandbox/tab-perf/react.test.tsx
diegohaz: Rechecked on the unchanged head. This remains actionable: `console.error` is still restored before `hydrateRoot`, so the test cannot detect the React 19 hydration mismatch that leaves the server attribute in place. The focused React 19 reproduction from the initial review remains applicable. Keeping the spy active and checking `expect(consoleError).not.toHaveBeenCalled()` after hydration would close the false-pass path.

packages/ariakit-react-components/src/command/command.tsx
```diff
@@ -65,11 +65,18 @@ export const useCommand = createHook<TagName, CommandOptions>(
...
     useEffect(() => {
-      if (!ref.current) return;
-      setIsNativeButton(isButton(ref.current));
-    }, []);
+      const element = ref.current;
+      if (!element) return;
+      const nativeButton = isButton(element);
+      // React 19's useFormStatus relies on the post-mount render below for
+      // native submit controls. A native type="button" control already has the
+      // default type, so updating state would only schedule a redundant render.
```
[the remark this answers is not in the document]
diegohaz: Agreed. The browser default for an omitted `<button type>` is `submit`; “default type” here means the value Command intends to apply, which is easy to misread. A precise replacement would be:

```ts
// native submit controls. A native type="button" control already has the
// type Command would apply, so updating state would only schedule a
// redundant render.
```

.changeset/200-collection-registration-performance.md
diegohaz: Agreed, with one package-boundary caveat: these should be consolidated into two multiline changesets rather than one. Six entries affect `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/react`, while the three `Tab`/`TabPanel` rendering entries affect only the React packages. Changesets applies the same summary to every package in a file, so one union-frontmatter entry would put React-only details in the `@ariakit/components` changelog.

For example, the grouped entries can retain their distinct package sets:

```md
---
"@ariakit/components": patch
"@ariakit/react": patch
---
Improved store performance
```

```md
---
"@ariakit/react": patch
---
Improved `Tab`, `TabPanel`, and `Focusable` performance
```

This reduces nine fragmented release notes to two focused entries without leaking React-only details into the core package changelog.

packages/ariakit-react-components/src/tab/tab.tsx
diegohaz: Yes on the scope, but I do not think the component hooks can safely supply the default. `useCommand` and `useCompositeItem` are also composed into components whose fallback tags are `input` or `div`, and public hooks may be spread onto arbitrary elements. The component wrapper is the first layer that knows both the actual fallback `TagName` and whether `render` replaces it.

The `Focusable` capability optimization is already generic; this `type="button"` path is specifically what lets `Command` skip its passive native-button update. Right now only `Tab` gets that benefit, while other ordinary fallback-button wrappers still mount without the attribute and update afterward. I confirmed that with `Button`: its first ref callback observes no `type` attribute, and the DOM settles to `type="button"` after effects.

I would use a shared component-boundary helper for ordinary `TagName = "button"` wrappers, preserving explicit types and custom renders:

```ts
function withDefaultButtonType<
  P extends { render?: unknown; type?: string },
>(props: P) {
  if (props.render || props.type !== undefined) return props;
  return { ...props, type: "button" as const };
}

const htmlProps = useTab(withDefaultButtonType(props));
```

`FormSubmit` and `FormReset` should remain semantic exceptions because they intentionally default to `submit` and `reset`.

diegohaz (commented):
Re-review summary

The head is unchanged from the previous marked review, so there is no incremental code diff. I reviewed all four unresolved threads against the current source and replied to each one.

Findings:

- The existing P2 hydration-test finding remains valid. The console spy is restored before `hydrateRoot`, leaving a React 19 attribute-mismatch false-pass path.
- The `Command` comment should say that `type="button"` is the type Command would apply, not the HTML default type.
- The nine release notes should be consolidated into two multiline changesets, split at the package boundary: six store-layer entries include `@ariakit/components`; three React rendering entries do not.
- The `type="button"` optimization should cover the other ordinary fallback-button components that reach `Command`, but at the component boundary rather than inside shared hooks. Those hooks also serve `div`, `input`, and arbitrary consumer-selected elements. `FormSubmit` and `FormReset` remain intentional exceptions.

Validation performed:

- Re-read every unresolved thread, its replies, and the surrounding current source.
- Ran `pnpm exec changeset status` and audited the frontmatter/package sets.
- Used a focused React test to verify that `Button` initially exposes no `type` attribute and settles to `type="button"` after effects.
- Audited the `useCommand` and `useCompositeItem` composition paths and fallback tag names.
- Rechecked the unchanged hydration-test call order against the focused React 19 reproduction from the initial review.

I did not rerun the full PR test or benchmark suites because there are no commits since the prior review; this pass was scoped to the unresolved comments and their surrounding behavior.

packages/ariakit-react-components/src/command/command.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in `c2bbe2e75` — the comment now says the native control already has the type Command would apply, without implying that `button` is the HTML default.

.changeset/200-collection-registration-performance.md
diegohaz: Fixed in `f937cf303` — the nine fragments are consolidated into two multiline changesets while preserving the React-only versus shared package boundaries.

packages/ariakit-react-components/src/tab/tab.tsx
diegohaz: Fixed in `c2bbe2e75` — `withDefaultButtonType` now runs at the final component boundary for all 21 ordinary native-button wrappers, while submit/reset semantics, public hooks, and custom renders remain unchanged.

app/src/sandbox/tab-perf/react.test.tsx
diegohaz: Fixed in `c2bbe2e75`, with the final coverage in `3920aff1c` and `b6b1966ab` — the console spy is cleared after server rendering, stays active through `hydrateRoot`, and hydration must emit no errors.

diegohaz:
**HYDRATION**

- `c2bbe2e75` ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3567861290)) Kept the console spy active through hydration and asserted that React emitted no hydration errors.

**COMMAND COMMENT**

- `c2bbe2e75` ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3567861287)) Reworded the native `type="button"` fast-path comment to avoid implying that `button` is the HTML default.

**CHANGESETS**

- `f937cf303` ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3567861288)) Consolidated the performance notes into two multiline changesets with the correct package boundaries.

**NATIVE BUTTON TYPES**

- `c2bbe2e75` ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3567861289)) Applied the shared default type at all 21 ordinary native-button component boundaries while preserving submit/reset and custom-render behavior.

app/src/sandbox/native-button-semantics/index.react.tsx
```diff
@@ -0,0 +1,176 @@
...
+const CustomButton = forwardRef<
+  HTMLButtonElement,
+  ComponentPropsWithoutRef<"button">
+>(function CustomButton(props, ref) {
+  return createElement("button", { ...props, ref });
+});
+
+function HookTab() {
+  const props = useTab<"div">();
+  return createElement(Ariakit.Role, props, "Hook tab");
```
diegohaz: Is there any reason to use `createElement` here (and in other places of this file) instead of `<Araikit.Role>` directly?

packages/ariakit-components/src/collection/collection-store.ts
```diff
@@ -43,6 +43,68 @@ function getPrivateStore<T extends CollectionStoreItem>(
...
+type SetItems<T extends CollectionStoreItem> = (
+  getItems: (items: T[]) => T[],
+) => void;
+
+interface MergeItemOptions<T extends CollectionStoreItem> {
+  cache: ItemIdCache<T>;
+  setItems: SetItems<T>;
+  canDeleteFromMap?: boolean;
+}
+
```
diegohaz, line 110: This whole cache logic could benefit from some "why" comments

packages/ariakit-components/src/disclosure/disclosure-store.ts
```diff
@@ -44,7 +47,9 @@ export function createDisclosureStore(
     disclosureElement: defaultValue(syncState?.disclosureElement, null),
   };
 
-  const disclosure = createStore(initialState, store);
+  const disclosure = store
+    ? createStore(initialState, store)
+    : createStore(initialState);
```
diegohaz, line 53: How is it different from this?
```suggestion
  const disclosure = createStore(initialState, store);
```

app/src/sandbox/native-button-semantics/index.react.tsx
diegohaz: There is no reason to use `createElement` for the static cases in this `.tsx` file. `CustomButton` and `HookTab` are simpler as JSX:

```tsx
return <button {...props} ref={ref} />;

return <Ariakit.Role {...props}>Hook tab</Ariakit.Role>;
```

The dynamic-tag cases such as `createElement(enabledTag)` still avoid introducing capitalized aliases for runtime tag strings, and the hydration test in `test.ts` must use `createElement` because that file does not support JSX. I would change the two static occurrences and keep the genuinely dynamic and non-TSX ones.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Agreed. The current helpers explain the mechanics, but the important constraints are not apparent: the `Set` is enabled only for large arrays, array identity invalidates it after synchronization or sorting, and it is deliberately an absence-only fast path because existing IDs still need `findIndex` to preserve position and layered restoration.

A focused comment above the cache helpers could capture that:

```ts
// Cache IDs only for large immutable state arrays so repeated registrations can
// skip O(n) absence scans. Replacing the array invalidates the cache. Existing
// IDs still use findIndex to preserve their position and previous value.
```

I would also note where the two cache instances are created that `items` and `renderedItems` need separate caches because those arrays are replaced independently. An ID-to-index map could accelerate positive lookups too, but maintaining indices through sorting, synchronization, deletion, and layered restoration would be substantially riskier than documenting this negative-only cache.

packages/ariakit-components/src/disclosure/disclosure-store.ts
diegohaz: The two forms are state-equivalent, but they are not equivalent on the store update hot path. `createStore(initialState, undefined)` creates a rest array with `stores.length === 1`; `createStore(initialState)` keeps `stores.length === 0`. The implementation checks `stores.length` on every local `setState`, so the suggested form enters the parent fan-out branch and iterates the undefined placeholder on every standalone Disclosure update.

A focused alternating microbenchmark of one million updates measured 21.1 ms median for the zero-parent call and 25.3 ms for the undefined-parent call. The earlier `mergeStore` guard is the larger creation win, but this conditional preserves the zero-parent update fast path. I would keep it and add the rationale so it does not look redundant:

```ts
// Omit an undefined parent so createStore keeps its zero-parent fast path.
const disclosure = store
  ? createStore(initialState, store)
  : createStore(initialState);
```

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -390,9 +404,23 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
 
-    const tagName = useTagName(ref);
-    const nativeTabbable = focusable && isNativeTabbable(tagName);
-    const supportsDisabled = focusable && supportsDisabledAttribute(tagName);
+    // Track only the element capabilities that affect the returned props so
+    // resolving the default native element doesn't require another render.
+    const [elementCapabilities, setElementCapabilities] = useState(
+      defaultElementCapabilities,
+    );
+    const elementCapabilitiesRef = useEvent((element: HTMLElement | null) => {
```
diegohaz: [P1] Remove per-render effects from the host capability refs

The [final performance report](https://github.com/ariakit/ariakit/pull/6694#issuecomment-4942217323) for this head confirms that opening the dialog regressed from 92.9 ms to 188.1 ms scripting (+103%), 151.5 ms to 248.3 ms total (+64%), and 168 ms to 264 ms INP (+57%). That interaction changes the parent `open` state and rerenders 3,000 `Button` instances. This new `useEvent` and the one in `useCommand` each install an insertion effect on every render, so an unchanged button host now incurs roughly 6,000 additional insertion-effect executions during that rerender. The current profile lists `useEvent`, `useFocusable`, and `useCommand` among the main hotspots.

Please make these host refs stable without a per-render effect. I prototyped a functional state update with `useCallback` in both locations; the React 19 and React 18 semantic tests (5 each) and `pnpm tsc` passed:

```ts
const elementCapabilitiesRef = useCallback(
  (element: HTMLElement | null) => {
    if (!element) return;
    const nextCapabilities = getElementCapabilities(
      element.tagName.toLowerCase(),
    );
    setElementCapabilities((current) =>
      current === nextCapabilities ? current : nextCapabilities,
    );
  },
  [],
);
```

The final implementation needs another CI-shaped performance run. If removing the two insertion effects does not recover the regression, the new post-hook `withDefaultButtonType` props copy is the next incremental cost to profile.

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -108,10 +119,56 @@ export function flipItems(
...
+    }
+  }
+  return rows;
+}
+
 /**
  * Creates a two-dimensional array with items grouped by their rowId's.
  */
 export function groupItemsByRows(items: CompositeStoreItem[]) {
+  if (items.length >= rowMapItemThreshold) {
```
diegohaz, line 192: [P2] Resolve the confirmed small-grid regression

The [final five-round Node report](https://github.com/ariakit/ariakit/pull/6694#issuecomment-4942217323) shows the 8-item/4-row control falling from 92,675 to 71,240 operations per second (-23%). This branch was intended to retain the original linear behavior below 48 items, but the final paired evidence still classifies the common small-grid path as a regression. The PR description currently cites an earlier -5% result and the consolidated changeset broadly claims improved multi-row navigation.

Please profile and recover the small path before keeping that broad claim, or explicitly accept the trade-off and narrow both the changeset and PR description to large grids. I tested the obvious extraction of the original loop into a separate small helper, but it did not materially improve a focused local timing run, so a mechanical split is not enough evidence for the fix.

app/src/sandbox/tab-perf/perf-chrome.ts
```diff
@@ -8,6 +8,8 @@ type Query = ReturnType<typeof query>;
 
 async function mountTabs(q: Query) {
   await q.button("Mount tabs").click();
+  await expect(q.tab(`Tab ${tabCount}`)).toBeVisible();
+  await expect(q.tab("Tab 1")).toHaveAttribute("type", "button");
```
diegohaz: [P2] Keep the duplicate type assertion outside measurement

Line 11 already provides the observable completion condition for mounting all 200 tabs, and `measureOnce` performs its own quiescence wait before taking the final metrics snapshot. The identical `type="button"` assertion runs again in `verifyTabsMounted` on line 26, after measurement. Keeping it here adds an unrelated Playwright DOM query and poll to the measured interaction without adding coverage.

Please remove line 12 and retain the unmeasured verification:

```ts
async function mountTabs(q: Query) {
  await q.button("Mount tabs").click();
  await expect(q.tab(`Tab ${tabCount}`)).toBeVisible();
}
```

.changeset/200-react-component-performance.md
```diff
@@ -0,0 +1,12 @@
...
... [`ComboboxDisclosure`](https://ariakit.com/reference/combobox-disclosure), [`CompositeItem`](https://ariakit.com/reference/composite-item), [`CompositeOverflowDisclosure`](https://ariakit.com/components/composite), [`DialogDisclosure`](https://ariakit.com/reference/dialog-disclosure), [`DialogDismiss`](https://ariakit.com/reference/dialog-dismiss), [`Disclosure`](https://ariakit.com/reference/disclosure), [`FormPush`](https://ariakit.com/reference/form-push), [`FormRemove`](https://ariakit.com/reference/form-remove), [`HovercardDisclosure`](https://ariakit.com/reference/hovercard-disclosure), [`HovercardDismiss`](https://ariakit.com/reference/hovercard-dismiss), [`MenuButton`](https://ariakit.com/reference/menu-button), [`MenuDismiss`](https://ariakit.com/reference/menu-dismiss), [`PopoverDisclosure`](https://ariakit.com/reference/popover-disclosure), [`PopoverDismiss`](https://ariakit.com/reference/popover-dismiss), [`Select`](https://ariakit.com/reference/select), [`SelectDismiss`](https://ariakit.com/reference/select-dismiss), [`Tab`](https://ariakit.com/reference/tab), and [`ToolbarItem`](https://ariakit.com/reference/toolbar-item) when they render their default native button.
```
diegohaz: [P3] Leave APIs without a reference page unlinked

The canonical `/reference/composite-overflow-disclosure` URL currently returns 404. The Ariakit changeset rule requires a pre-existing public API without a deployed reference page to remain plain inline code rather than linking its name to a different component-overview page. Please render this occurrence as `CompositeOverflowDisclosure` without the `/components/composite` link; all the other reference links in these two consolidated changesets resolve successfully.

diegohaz (commented):
Re-review summary

This re-review covered the seven commits from the latest prior marked baseline, `016493fdfe227a9902e8bf31fabb5ec9803de97b`, through `b6b1966ab5425bfbd05c226802e605319a2b5457`, plus the three newly opened unresolved threads and surrounding performance evidence. Changes inherited unchanged from the merged `main` branch were excluded from the effective PR-change review.

I verified that the four prior threads were fixed: the hydration console spy remains active through `hydrateRoot`, the Command wording is accurate, the nine changesets became two package-aware entries, and the shared boundary helper covers all 21 ordinary native-button wrappers while preserving submit/reset and custom-render semantics.

I left four new inline findings:

- P1: the final CI-shaped report confirms that opening the dialog regressed by 103% scripting, 64% total time, and 57% INP. The new Command and Focusable host refs add two `useEvent` insertion effects to every Button rerender; I prototyped stable `useCallback` refs with functional state updates and the semantic/type checks passed.
- P2: the final five-round Node report confirms a 23% regression for grouping 8 items into 4 rows. The small-grid path must be recovered or the release claim narrowed.
- P2: the Tab mount performance interaction duplicates the `type="button"` assertion inside the measured interval even though the same check already runs in the unmeasured verifier.
- P3: `CompositeOverflowDisclosure` links to a generic component page even though its canonical reference page is absent; the repository changeset rule requires plain inline code in that case.

I also replied to all three unresolved human threads. Static `createElement` calls can be simplified to JSX while dynamic and non-TSX calls remain justified; the Collection cache needs rationale comments for its threshold, identity invalidation, and negative-only semantics; and the Disclosure `createStore` conditional preserves the zero-parent update fast path because passing `undefined` still makes `stores.length` nonzero.

Validation performed:

- `pnpm install` and `pnpm exec node --version` (Node 24.18.0)
- `pnpm test-react app/src/sandbox/native-button-semantics/test.ts` (5 passed)
- `pnpm test-react18 app/src/sandbox/native-button-semantics/test.ts` in a fresh generated workspace (5 passed)
- `pnpm -F app test-chrome native-button-semantics` (4 passed)
- `pnpm test packages/ariakit-components/src/collection/collection-store.test.ts packages/ariakit-components/src/disclosure/disclosure-store.test.ts` (8 passed)
- `pnpm tsc`
- `pnpm exec changeset status`
- Focused callback and store-update experiments, current source/profile inspection, all changeset reference URLs, the final performance report, and the current green CI checks

I did not rerun the full multi-round Chrome and Node comparison locally; the regression verdicts use the current final CI aggregate. I also limited local browser verification to Chrome because the complete current cross-browser matrix is green. The PR performance section should be refreshed after the regressions are resolved or explicitly accepted, since its current control and confirmation claims predate the final report.

.changeset/200-react-component-performance.md
diegohaz: We don't need to mention `CompositeOverflowDisclosure`

packages/ariakit-react-components/src/combobox/combobox-cancel.tsx
```diff
@@ -122,7 +123,7 @@ export const ComboboxCancel = forwardRef(function ComboboxCancel(
   props: ComboboxCancelProps,
 ) {
   const htmlProps = useComboboxCancel(props);
-  return createElement(TagName, htmlProps);
+  return createElement(TagName, withDefaultButtonType(htmlProps));
```
diegohaz: Why not this (for other components too)?

```ts
  const htmlProps = useComboboxCancel(withDefaultButtonType(props));
  return createElement(TagName, htmlProps);
```

packages/ariakit-react-components/src/button/utils.ts
```diff
@@ -0,0 +1,10 @@
+interface ButtonTypeProps {
+  render?: unknown;
+  type?: string;
+}
+
+export function withDefaultButtonType<P extends object>(props: P) {
```
diegohaz, line 6: Shouldn't `P` extends `ButtonTypeProps`?

app/src/sandbox/native-button-semantics/index.react.tsx
diegohaz: Updated in `ca5ca3eeb`. `CustomButton` and `HookTab` now use JSX. The remaining `createElement` call is in `test.ts`, where JSX is unavailable and the server/hydration test needs to construct the element programmatically.

packages/ariakit-components/src/collection/collection-store.ts
diegohaz: Added the rationale in `ca5ca3eeb`: the ID set is only used for large arrays and absence checks, array replacement invalidates it, positive matches still use `findIndex`, and `items`/`renderedItems` need separate caches because they are replaced independently.

packages/ariakit-components/src/disclosure/disclosure-store.ts
diegohaz: Kept the conditional and documented it in `ca5ca3eeb`. Passing `undefined` still creates a one-entry rest array, so local `setState` enters the parent fan-out branch; omitting the argument keeps `stores.length === 0`. Filtering undefined parents inside `createStore` would be a broader cross-store change, so I am evaluating that separately rather than folding it into this local fix.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in `ca5ca3eeb`. Both capability callback refs are gone: Focusable reconciles once in a mount layout effect and Command uses its `type`-keyed mount effect. Unkeyed host swaps were removed per the documented keyed-remount contract. Two interleaved Dialog pairs are clean: open scripting 45.1 ms → 42.8 ms (-5%), with rendering and close flat.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in `ca5ca3eeb`. The small path now has its own helper and caches the preceding row while retaining `find` for nonconsecutive rows. Five interleaved full-order rounds measured the 8-item/4-row case at 175,665 → 208,392 ops/s (+19%); the 48-item/2-row control was +1%, and the large-row gains remained +42% to +1,364%.

app/src/sandbox/tab-perf/perf-chrome.ts
diegohaz: Removed the duplicate assertion from `mountTabs` in `ca5ca3eeb`; the unmeasured `verifyTabsMounted` assertion remains. Across four interleaved rounds, Tab mount scripting was 66.6 ms → 59.2 ms (-11%) with no regression classification.

.changeset/200-react-component-performance.md
diegohaz: Removed `CompositeOverflowDisclosure` from the changeset entirely in `ca5ca3eeb`, following the later clarification.

packages/ariakit-react-components/src/combobox/combobox-cancel.tsx
diegohaz: The final boundary is intentional. Some hooks can introduce a non-button `render` internally after receiving caller props; nested `MenuButton` is one example. Applying the default before every hook could therefore leak `type="button"` onto a generated `div`. `Tab` uses both stages because it needs the type inside its hook for mount performance while the final boundary still protects the actual host.

packages/ariakit-react-components/src/button/utils.ts
diegohaz: Constraining `P` to `ButtonTypeProps` does not type-check across the generic hook outputs. For example, `MenuButton` returns `HTMLProps` that intentionally omit custom option keys and have no statically visible overlap with an all-optional weak type, producing TS2345/TS2559. Keeping `P extends object` and localizing the runtime inspection to this cast preserves those call sites.

diegohaz:
**Fixed**

- [`8c343f4e1`](https://github.com/ariakit/ariakit/commit/8c343f4e1175424a029f82ef7658378f323af0d3) ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3576062915)): Added the missing `Focusable` behavior changeset.
- [`dd2a66673`](https://github.com/ariakit/ariakit/commit/dd2a66673a199d5b93e81956deab0365e701381c) ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3576062971)): Removed unrelated peer-resolution churn with `pnpm dedupe` and verified the frozen install.

**Follow-up**

- Updated the PR description to reflect the final runtime changes and published CI performance results.

packages/ariakit-components/src/disclosure/disclosure-store.ts
diegohaz: Shouldn't we filter the stores array before checking its length? The way it's written now seems counterintuitive.

packages/ariakit-react-components/src/combobox/combobox-cancel.tsx
diegohaz: You’re right for `Tab`. After auditing every current `useTab` path, the final helper was redundant: pre-defaulted props retain either a defined `type` or a truthy `render` through the hook chain, including the internal `CompositeItem` wrapper. Simplified in `6bf3cc0e6`:

```tsx
const htmlProps = useTab(withDefaultButtonType(props));
return createElement(TagName, htmlProps);
```

The distinction still matters for hooks that genuinely replace the host. Nested `MenuButton`, for example, can substitute `Role.div`, so it must keep `withDefaultButtonType(htmlProps)` after the hook instead. React, React 18, SSR/hydration, native-button browser tests, React 19 form-status tests, TypeScript, lint, and the production Tab performance comparison all pass; the latter is flat within noise.

packages/ariakit-components/src/disclosure/disclosure-store.ts
diegohaz: I tested that direction. Filtering `undefined` parents is conceptually reasonable, but the naive global form is not safe enough for this PR:

```ts
stores = stores.filter((store) => store !== undefined);
```

In same-runtime measurements it made the existing zero-parent `setState` path roughly 8–17% slower because reassigning the captured rest array inhibited V8’s zero-length specialization. It also changes observable reentrant initialization timing: `createStore(state, parent, undefined)` currently uses the multi-parent grouped-subscription path, while filtering makes it use the single-parent per-key path.

A separate wrapper/factory could normalize parents without that hot-path regression, but it needs dedicated reentrancy tests and full Node/browser validation. I kept the local Disclosure conditional here because it delivers the measured zero-parent gain without broadening this PR’s store semantics.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -228,12 +243,11 @@ export const useFocusable = createHook<TagName, FocusableOptions>(
...
-    // When the focusable element is disabled, it doesn't trigger a blur event
-    // so we can't set focusVisible to false there. Instead, we have to do it
-    // here by checking the element's disabled attribute.
+    // When the focusable element is disabled, it doesn't trigger a blur event,
+    // and turning focusable off disables the blur handler below. Clear the
+    // internal focus-visible state here in both cases.
     useEffect(() => {
-      if (!focusable) return;
-      if (!trulyDisabled) return;
+      if (focusable && !trulyDisabled) return;
```
diegohaz, line 250: [P2] Add a changeset for the Focusable behavior fix

The current base native-submit path writes `data-focus-visible` directly and intentionally skips `setFocusVisible(true)`. As a result, React still sees that prop as undefined when `focusable` turns false and cannot remove the manually written attribute. This changed effect fixes that consumer-visible behavior, and the new semantic fixture explicitly covers it, but the current React changeset describes only performance and initial button-type markup.

Please add a patch entry for both public packages. For example:

```md
---
"@ariakit/react": patch
---

Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Button`](https://ariakit.com/reference/button), to clear focus-visible styling when `focusable` becomes `false`.
```

pnpm-lock.yaml
```diff
@@ -148,7 +148,7 @@ importers:
         version: runtime:24.18.0
       open-cli:
         specifier: 9.0.0
-        version: 9.0.0
+        version: 9.0.0(supports-color@8.1.1)
```
diegohaz: [P3] Drop the unrelated peer-resolution churn

The only manifest change requiring a lock update is the new `vitest` dev dependency, represented by the three-line importer addition at lines 608-610. The initial dependency commit made exactly that lock change; a later button-markup commit regenerated 56 unrelated peer contexts, including `open-cli` here plus Stylelint, tsup, OpenNext, WordPress, Babel, and Express entries. The current base already contains the exact Vitest/Vite 8.1.4 snapshot referenced by the new importer.

Please restore these unrelated entries from the base and retain only:

```yaml
vitest:
  specifier: 4.1.10
  version: 4.1.10(...vite@8.1.4...)
```

Then rerun the frozen install. This keeps the lockfile diff scoped to the package change and also removes the obsolete need for the PR description to explain a Vite 8.1.3 reconciliation.

diegohaz (commented):
Re-review summary

This re-review covered the revisions from the latest prior marked head, `b6b1966ab5425bfbd05c226802e605319a2b5457`, through `056e24df2173a17586233e985124939e4d2af532`. Because the branch was rebased, I used a range-diff to separate the PR revisions from newer `main` commits, then reviewed the full 46-file diff from current base `8df5ac52fb677c7f850181d0a99516c1a89bcf51`.

I left two new inline findings:

- P2: the final `Focusable` effect fixes consumer-visible focus styling when `focusable` becomes false, but neither current changeset describes that bug fix.
- P3: the lockfile contains 56 unrelated `supports-color` peer-context rewrites; only the three-line `vitest` importer addition is required by the package manifest change.

I found no additional correctness, compatibility, security, or performance regressions. The latest prior review findings are resolved: the dialog interaction no longer pays per-render host-ref effects, the small 8-item/4-row benchmark now reports a confirmed 12% improvement, the duplicate measured Tab assertion is gone, and `CompositeOverflowDisclosure` is no longer linked to a nonexistent reference page. The Collection caches and hybrid Composite row grouping remain justified by equivalence probes and the current paired results; simplifying them to shared maps would weaken synchronization or small-grid behavior.

Unresolved-thread validation: the fully paginated GraphQL result contains zero unresolved review threads, so there were no candidate threads requiring replies or dispositions in this run.

Validation performed:

- `pnpm install` with Node 24.18.0
- Focused Collection, Composite, Disclosure, and Tab store tests (32 passed)
- Native-button semantic tests under React 19 and React 18 (5 passed in each), plus focused browser fixtures
- `pnpm tsc`, `pnpm exec changeset status`, `git diff --check`, and package dry-run inspection
- 50,000 randomized Collection operations and 101,000 randomized row-grouping cases against the prior algorithms
- Current all-green CI and the final paired Node/Chrome performance report, which contains no confirmed regressions

Non-blocking description note: the PR body still says CI is running on `ca5ca3eeb`, lists pre-rebase test totals, and describes repairing a Vite 8.1.3 locator. The current head is `056e24df2`, all checks have completed, and the current base already uses the Vite 8.1.4 graph, so that section should be refreshed before merge.

I did not rerun the complete multi-round base/head performance comparison or the full repository cross-browser suite locally; those areas were covered by the current green CI run and published paired report.

pnpm-lock.yaml
diegohaz: Maybe `pnpm dedupe`?

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in [`8c343f4e1`](https://github.com/ariakit/ariakit/commit/8c343f4e1175424a029f82ef7658378f323af0d3): added a patch changeset for `@ariakit/react-components` and `@ariakit/react` covering the focus-visible cleanup when [`focusable`](https://ariakit.com/reference/focusable#focusable) becomes `false`.

pnpm-lock.yaml
diegohaz: Good call. Fixed in [`dd2a66673`](https://github.com/ariakit/ariakit/commit/dd2a66673a199d5b93e81956deab0365e701381c): `pnpm dedupe` removed the unrelated peer-resolution churn, leaving only the intended Vitest importer change. The frozen lockfile install and dedupe check pass.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz, line 250: [P2] Preserve caller-supplied focus-visible attributes

This effect now runs whenever `focusable` is `false` and calls `cleanupFocusVisible`, which unconditionally removes the DOM attribute. The returned props deliberately spread `...props` after the internal default, so callers can explicitly supply this data attribute. On the current head, this public usage commits `custom` and then loses it in the passive effect:

```tsx
<Focusable focusable={false} data-focus-visible="custom">
  Content
</Focusable>
```

I added that case to the existing `native-button-semantics` fixture locally; the React 19 test failed with an expected value of `custom` and an actual value of `null`. The base guard skipped this cleanup when `focusable` was false.

Please distinguish an Ariakit-owned marker from a caller-provided value, or restore the explicit prop after cleaning the internal state, and cover this prop-forwarding case in the happy-dom and browser fixtures.

.changeset/300-focusable-focus-visible.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Button`](https://ariakit.com/reference/button), to clear focus-visible styling when [`focusable`](https://ariakit.com/reference/focusable#focusable) becomes `false`.
```
diegohaz: [P3] Link to the actual focusable prop section

The deployed `/reference/focusable` page uses `id="focusable"` for its page heading. The `focusable` prop is documented under `id="focusable-1"`, and the page itself links that member as `#focusable-1`. The current URL therefore sends readers to the top of the page instead of the named prop.

Please use the verified member anchor:

```md
[`focusable`](https://ariakit.com/reference/focusable#focusable-1)
```

diegohaz (commented):
Re-review summary

This re-review covered the two commits from the latest prior marked head, `056e24df2173a17586233e985124939e4d2af532`, through `dd2a66673a199d5b93e81956deab0365e701381c`, plus the affected Focusable behavior and the current full PR/CI context. The incremental repository diff adds one Focusable changeset and restores the lockfile; the full current base-to-head diff contains 47 files.

The two prior findings are resolved:

- The new `300-focusable-focus-visible.md` changeset covers the consumer-visible fix in both `@ariakit/react-components` and `@ariakit/react`.
- The full base-to-head `pnpm-lock.yaml` diff is now exactly the required three-line Vitest importer addition.

I left two new inline findings:

- P2: when `focusable={false}`, the new cleanup effect removes a caller-supplied `data-focus-visible` attribute after React commits it. A focused public-fixture reproduction expected `custom` and received `null`.
- P3: the new changeset links the `focusable` prop to the page heading; the deployed member section is `#focusable-1`.

Pre-existing unresolved-thread validation: the fully paginated GraphQL result contained zero unresolved threads and no paginated comment connections before this run posted its two new findings.

Validation performed:

- `pnpm install --frozen-lockfile` and `pnpm exec node --version` (Node 24.18.0)
- `pnpm exec changeset status` and `pnpm dedupe --check`
- Native-button semantic fixture under React 19 and React 18 (5 passed in each) and Chrome (4 passed)
- A temporary explicit-`data-focus-visible` regression case, which failed as described and was then removed
- Current source/base inspection, deployed reference anchors, `git diff --check`, and a clean worktree
- The complete green CI run and the final paired performance report for this exact head, with no confirmed regressions

The PR description is structurally accurate, but its exact performance numbers still reflect the preceding report run. For example, the current report has Tab mount at -15% scripting/-14% total rather than -13%/-12%, and large Collection registration at +49% rather than +53%. Refreshing those bullets, or wording them to avoid volatile run-specific percentages, would keep the “final report” claim accurate.

I did not rerun the complete CI-shaped paired performance comparison or download its raw artifacts locally; the performance verdict uses the final published report and green workflow for this head.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: This is not an issue because `data-focus-visible` is Ariakit-generated state: `Focusable` applies it for keyboard-visible focus and removes it when that computed state ends. Setting the same attribute as caller-controlled state is unsupported, so `focusable={false}` intentionally clears it.

For an independent forced focus ring, use a separate marker:

```tsx
<Ariakit.Button
  data-force-focus-ring={forceFocusRing || undefined}
/>
```

```css
.button[data-focus-visible],
.button[data-force-focus-ring] {
  outline: 2px solid;
}
```

For fully custom focus-visible handling, `onFocusVisible` can call `event.preventDefault()` and manage a separate state marker.

.changeset/300-focusable-focus-visible.md
diegohaz: Fixed in [`174dee39b`](https://github.com/ariakit/ariakit/commit/174dee39bd6aa65a295f4aa2822e0df67c6f5aff): changed the link to the verified `#focusable-1` prop anchor.

diegohaz:
**Fixed**

- [`174dee39b`](https://github.com/ariakit/ariakit/commit/174dee39bd6aa65a295f4aa2822e0df67c6f5aff) ([reply](https://github.com/ariakit/ariakit/pull/6694#discussion_r3576413882)): Corrected the `focusable` changeset link to the verified `#focusable-1` prop anchor.

**Invalid**

- [`Unsupported usage`](https://github.com/ariakit/ariakit/pull/6694#discussion_r3576413756): `data-focus-visible` is generated Ariakit state, so `focusable={false}` intentionally clears it. The reply documents a separate `data-force-focus-ring` marker and `onFocusVisible` customization as supported alternatives.

**Validation**

- Rebased onto the latest `main`; the frozen lockfile install and all 47 final checks passed.
- The final [performance report](https://github.com/ariakit/ariakit/pull/6694#issuecomment-4942217323) confirms the Tab/Composite mount improvements without a threshold-crossing regression.
- Squash auto-merge is enabled and will run after the required post-push approval.

diegohaz (commented):
Re-review summary

This re-review covered the rewritten branch from the latest prior marked baseline, `dd2a66673a199d5b93e81956deab0365e701381c`, through `174dee39bd6aa65a295f4aa2822e0df67c6f5aff`, plus the current full 47-file base-to-head diff, surrounding Focusable behavior, release metadata, and the exact-head CI/performance evidence. The branch is now based on `ad660335442a7702be8b01013a39be9b4b296e85`; `git range-diff` marks all 22 previously reviewed commits as patch-equivalent and adds only the Focusable changeset-link fix.

I found no new substantive code, compatibility, security, release, simplification, or performance issues.

Requested prior-finding validation:

- The `data-focus-visible` candidate is invalid/inaccurate after independent validation. The documented public state is `true` while active and `undefined` while inactive, and both current and pre-PR Focusable code imperatively set and remove this Ariakit-owned marker. Therefore `<Focusable data-focus-visible="custom" />` does not establish supported caller-controlled state; preserving that unsupported input would add ownership branches to a hot foundational hook without preserving a supported behavior.
- The changeset-link finding is resolved. `.changeset/300-focusable-focus-visible.md` now uses `https://ariakit.com/reference/focusable#focusable-1`, and a fresh deployed-page check confirms that exact member anchor.

Unresolved-thread validation: the fully paginated GraphQL result contains zero unresolved review threads, and no thread comment connection requires another page, so this run required no inline comments or replies.

Validation performed:

- `pnpm install --frozen-lockfile` and `pnpm exec node --version` (Node 24.18.0)
- `pnpm test-react app/src/sandbox/native-button-semantics/test.ts` (5 passed)
- `pnpm test-react18 app/src/sandbox/native-button-semantics/test.ts` (5 passed)
- `pnpm exec changeset status`, `pnpm dedupe --check`, and `git diff --check`
- Rebase `range-diff`, stable patch-ID comparison, current base-to-head lockfile audit, deployed reference-anchor verification, and clean-worktree inspection
- All 46 current-head checks and the final paired performance report. The report contains no confirmed or candidate regression; its only candidates are Composite mount INP improvements. The PR description was refreshed during this review and now matches the final exact-head metrics.

I did not rerun the complete cross-browser or CI-shaped multi-round performance suites locally; those were covered by the all-green current-head CI run and its final paired report.

PR #6717, opened 2026-07-13
Add keyed store subscriptions

diegohaz:
## Motivation

Selector functions passed to store hooks currently run after every store update, even when the selector only reads one or two state keys. This creates avoidable work across React components with frequently changing stores.

```ts
useStoreState(store, (state) => state.value);
```

The hooks also have no way to express selector dependencies, so TypeScript cannot catch reads that would make a filtered subscription stale.

## Solution

Added overloads that accept dependency keys before the selector or state object:

```ts
useStoreState(store, ["value"], (state) => state.value);

useStoreStateObject(store, ["value"], {
  direct: "otherKey",
  derived: (state) => state.value,
});
```

The declared keys narrow the selector state in TypeScript and filter store notifications at runtime, while selectors still receive the complete state object. Direct key mappings are subscribed automatically, dependency arrays are stabilized by value, and existing unkeyed selector calls preserve their behavior.

`useStoreState` is available from `@ariakit/react`, `@ariakit/react-components/store`, and `@ariakit/react-store`. `useStoreStateObject` is available from the latter two store entrypoints and is documented in a separate release note.

Updated all selector-form subscriptions across `@ariakit/react-components` to declare their dependencies. Expensive Composite row calculations are isolated from `activeId` updates, direct reads use the direct-key overload, and mutable element IDs remain observed after the migration.

Also aligned `NaN` key matching between the store's fast and slow subscription paths.

## Testing

- `pnpm test-react` — 193 files, 853 tests
- `pnpm test-react18` — 193 files, 845 passed and 8 expected skips
- `pnpm test packages/ariakit-store/src/test.ts` — 92 tests
- Focused component sandboxes in React 19 and React 18 — 6 tests each
- Keyed-store sandbox in React 19 and React 18 — 4 tests each
- Keyed-store browser sandbox in Chrome, Firefox, and Safari — 12 tests
- Mutable-element-ID browser sandbox in Chrome, Firefox, and Safari — 12 tests
- `pnpm tsc`
- `pnpm lint`
- `pnpm build`
- `pnpm build-app-lite`
- `pnpm build-website-lite`
- Modern and legacy reference extractor tests

### DIFF
--- a/.changeset/200-keyed-store-subscriptions.md
+++ b/.changeset/200-keyed-store-subscriptions.md
@@ -4,8 +4,17 @@
 "@ariakit/react": patch
 ---
 
-Added keyed store subscriptions
+Keyed store subscriptions
 
-The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks now accept selector dependency keys, so selectors skip unrelated store updates while receiving the complete store state at runtime.
+The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks now accept selector dependency keys, so selectors skip unrelated store updates while receiving the complete store state at runtime. The key list must include every store key a selector reads, or its result may stay stale.
 
-Updated React components to use keyed selector subscriptions internally.
+```ts
+const isEmpty = useStoreState(store, ["value"], (state) => !state.value);
+
+const values = useStoreStateObject(store, ["value"], {
+  value: "value",
+  valueLength: (state) => state.value.length,
+});
+```
+
+React components now use keyed selector subscriptions internally.
--- a/.changeset/200-keyed-store-subscriptions.md
+++ b/.changeset/200-keyed-store-subscriptions.md
@@ -6,15 +6,10 @@
 
 Keyed store subscriptions
 
-The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks now accept selector dependency keys, so selectors skip unrelated store updates while receiving the complete store state at runtime. The key list must include every store key a selector reads, or its result may stay stale.
+The [`useStoreState`](https://ariakit.com/reference/use-store-state) hook now accepts selector dependency keys, so selectors skip unrelated store updates while receiving the complete store state at runtime. The key list must include every store key a selector reads, or its result may stay stale.
 
 ```ts
 const isEmpty = useStoreState(store, ["value"], (state) => !state.value);
-
-const values = useStoreStateObject(store, ["value"], {
-  value: "value",
-  valueLength: (state) => state.value.length,
-});
 ```
 
 React components now use keyed selector subscriptions internally.
--- a/.changeset/200-keyed-store-subscriptions.md
+++ b/.changeset/200-keyed-store-subscriptions.md
@@ -1,0 +1,11 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added keyed store subscriptions
+
+The [`useStoreState`](https://ariakit.com/reference/use-store-state) and `useStoreStateObject` hooks now accept selector dependency keys, so selectors skip unrelated store updates while receiving the complete store state at runtime.
+
+Updated React components to use keyed selector subscriptions internally.
--- a/.changeset/300-nan-store-keys.md
+++ b/.changeset/300-nan-store-keys.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
+---
+
+Fixed store subscriptions to respond consistently to updates made with `NaN` keys.
--- a/.changeset/300-nan-store-keys.md
+++ b/.changeset/300-nan-store-keys.md
@@ -1,5 +1,8 @@
 ---
 "@ariakit/store": patch
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
 "@ariakit/solid-store": patch
 ---
 
--- a/app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
+++ b/app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
@@ -1,0 +1,18 @@
+import { click, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+test("updates combobox relationships when the input id changes", async () => {
+  const section = q.within(q.region("Combobox"));
+  const label = section.text.ensure("Combobox label");
+  const combobox = section.combobox("Combobox label");
+  const cancel = section.button("Clear input");
+
+  expect(label).toHaveAttribute("for", "combobox-before");
+  expect(cancel).toHaveAttribute("aria-controls", "combobox-before");
+
+  await click(section.button("Change combobox id"));
+
+  await expect.poll(() => label.getAttribute("for")).toBe("combobox-after");
+  expect(cancel).toHaveAttribute("aria-controls", "combobox-after");
+  expect(combobox).toHaveAttribute("id", "combobox-after");
+});
--- a/app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
+++ b/app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
@@ -16,3 +16,16 @@ test("updates combobox relationships when the input id changes", async () => {
   expect(cancel).toHaveAttribute("aria-controls", "combobox-after");
   expect(combobox).toHaveAttribute("id", "combobox-after");
 });
+
+test("updates the tooltip anchor when the content id changes", async () => {
+  const section = q.within(q.region("Tooltip"));
+  const anchor = section.button.ensure("Tooltip label");
+
+  expect(anchor).toHaveAttribute("aria-labelledby", "tooltip-before");
+
+  await click(section.button("Change tooltip id"));
+
+  await expect
+    .poll(() => anchor.getAttribute("aria-labelledby"))
+    .toBe("tooltip-after");
+});
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -134,10 +134,42 @@ function DirectValues({ store }: StoreProps) {
   );
 }
 
+function DynamicSelectors({ store }: StoreProps) {
+  const [key, setKey] = useState<keyof TestState>("foo");
+  const value = useStoreState(store, [key], (state) => state[key]);
+  const object = useStoreStateObject(store, [key], {
+    value: (state) => state[key],
+  });
+
+  return (
+    <>
+      <button type="button" onClick={() => setKey("bar")}>
+        Subscribe to bar
+      </button>
+      <p>
+        Dynamic selector value:{" "}
+        <output aria-label="Dynamic selector value">
+          {key}:{value}
+        </output>
+      </p>
+      <p>
+        Dynamic object selector value:{" "}
+        <output aria-label="Dynamic object selector value">
+          {key}:{object.value}
+        </output>
+      </p>
+    </>
+  );
+}
+
 function Controls({ store, calls }: SelectorProps) {
+  const [hydrated, setHydrated] = useState(false);
   const [, forceUpdate] = useReducer((count: number) => count + 1, 0);
 
-  useEffect(() => forceUpdate(), []);
+  useEffect(() => setHydrated(true), []);
+
+  const getCallCount = (key: keyof SelectorCalls) =>
+    hydrated ? calls[key] : 0;
 
   const update = (key: keyof TestState) => {
     store.setState(key, (value) => value + 1);
@@ -154,24 +186,32 @@ function Controls({ store, calls }: SelectorProps) {
       </button>
       <p>
         useStoreState calls:{" "}
-        <output aria-label="useStoreState calls">{calls.state}</output>
+        <output aria-label="useStoreState calls">
+          {getCallCount("state")}
+        </output>
       </p>
       <p>
         useStoreStateObject calls:{" "}
-        <output aria-label="useStoreStateObject calls">{calls.object}</output>
+        <output aria-label="useStoreStateObject calls">
+          {getCallCount("object")}
+        </output>
       </p>
       <p>
         Unkeyed selector calls:{" "}
-        <output aria-label="Unkeyed selector calls">{calls.unkeyed}</output>
+        <output aria-label="Unkeyed selector calls">
+          {getCallCount("unkeyed")}
+        </output>
       </p>
       <p>
         Empty selector calls:{" "}
-        <output aria-label="Empty selector calls">{calls.empty}</output>
+        <output aria-label="Empty selector calls">
+          {getCallCount("empty")}
+        </output>
       </p>
       <p>
         Empty object selector calls:{" "}
         <output aria-label="Empty object selector calls">
-          {calls.emptyObject}
+          {getCallCount("emptyObject")}
         </output>
       </p>
     </>
@@ -196,6 +236,7 @@ export default function Example() {
       <EmptyObjectSelector store={store} calls={calls.current} />
       <DirectValues store={store} />
       <MixedValues store={store} />
+      <DynamicSelectors store={store} />
       <OptionalSelector store={store} />
       <Controls store={store} calls={calls.current} />
     </>
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -1,0 +1,203 @@
+import { useStoreState, useStoreStateObject } from "@ariakit/react-store";
+import { createStore } from "@ariakit/store";
+import type { Store } from "@ariakit/store";
+import { useEffect, useReducer, useRef, useState } from "react";
+
+interface TestState {
+  foo: number;
+  bar: number;
+}
+
+interface SelectorCalls {
+  state: number;
+  object: number;
+  unkeyed: number;
+  empty: number;
+  emptyObject: number;
+}
+
+interface StoreProps {
+  store: Store<TestState>;
+}
+
+interface SelectorProps extends StoreProps {
+  calls: SelectorCalls;
+}
+
+function KeyedSelectors({ store, calls }: SelectorProps) {
+  const value = useStoreState(store, ["foo"], (state) => {
+    calls.state += 1;
+    return state.foo;
+  });
+  const object = useStoreStateObject(store, ["foo"], {
+    value: (state) => {
+      calls.object += 1;
+      return state.foo;
+    },
+  });
+  const runtimeBar = useStoreState(store, ["foo"], (state) =>
+    Reflect.get(state, "bar"),
+  );
+
+  return (
+    <>
+      <p>
+        Selector value: <output aria-label="Selector value">{value}</output>
+      </p>
+      <p>
+        Object selector value:{" "}
+        <output aria-label="Object selector value">{object.value}</output>
+      </p>
+      <p>
+        Runtime bar: <output aria-label="Runtime bar">{runtimeBar}</output>
+      </p>
+    </>
+  );
+}
+
+function EmptySelector({ store, calls }: SelectorProps) {
+  useStoreState(store, [], () => {
+    calls.empty += 1;
+    return null;
+  });
+  return null;
+}
+
+function UnkeyedSelector({ store, calls }: SelectorProps) {
+  useStoreState(store, (state) => {
+    calls.unkeyed += 1;
+    return state.foo;
+  });
+  return null;
+}
+
+function EmptyObjectSelector({ store, calls }: SelectorProps) {
+  useStoreStateObject(store, [], {
+    value: () => {
+      calls.emptyObject += 1;
+      return null;
+    },
+  });
+  return null;
+}
+
+function MixedValues({ store }: StoreProps) {
+  const values = useStoreStateObject(store, ["foo"], {
+    directBar: "bar",
+    doubledFoo: (state) => state.foo * 2,
+  });
+
+  return (
+    <>
+      <p>
+        Mixed direct bar:{" "}
+        <output aria-label="Mixed direct bar">{values.directBar}</output>
+      </p>
+      <p>
+        Mixed doubled foo:{" "}
+        <output aria-label="Mixed doubled foo">{values.doubledFoo}</output>
+      </p>
+    </>
+  );
+}
+
+function OptionalSelector({ store }: StoreProps) {
+  const [currentStore, setCurrentStore] = useState<Store<TestState>>();
+  const value = useStoreState(currentStore, ["foo"], (state) => state?.foo);
+
+  return (
+    <>
+      <button type="button" onClick={() => setCurrentStore(store)}>
+        Use optional store
+      </button>
+      <p>
+        Optional value:{" "}
+        <output aria-label="Optional value">{value ?? "none"}</output>
+      </p>
+    </>
+  );
+}
+
+function DirectValues({ store }: StoreProps) {
+  const foo = useStoreState(store, "foo");
+  const { bar } = useStoreStateObject(store, { bar: "bar" });
+
+  return (
+    <>
+      <p>
+        Direct foo: <output aria-label="Direct foo">{foo}</output>
+      </p>
+      <p>
+        Direct bar: <output aria-label="Direct bar">{bar}</output>
+      </p>
+    </>
+  );
+}
+
+function Controls({ store, calls }: SelectorProps) {
+  const [, forceUpdate] = useReducer((count: number) => count + 1, 0);
+
+  useEffect(() => forceUpdate(), []);
+
+  const update = (key: keyof TestState) => {
+    store.setState(key, (value) => value + 1);
+    forceUpdate();
+  };
+
+  return (
+    <>
+      <button type="button" onClick={() => update("foo")}>
+        Update foo
+      </button>
+      <button type="button" onClick={() => update("bar")}>
+        Update bar
+      </button>
+      <p>
+        useStoreState calls:{" "}
+        <output aria-label="useStoreState calls">{calls.state}</output>
+      </p>
+      <p>
+        useStoreStateObject calls:{" "}
+        <output aria-label="useStoreStateObject calls">{calls.object}</output>
+      </p>
+      <p>
+        Unkeyed selector calls:{" "}
+        <output aria-label="Unkeyed selector calls">{calls.unkeyed}</output>
+      </p>
+      <p>
+        Empty selector calls:{" "}
+        <output aria-label="Empty selector calls">{calls.empty}</output>
+      </p>
+      <p>
+        Empty object selector calls:{" "}
+        <output aria-label="Empty object selector calls">
+          {calls.emptyObject}
+        </output>
+      </p>
+    </>
+  );
+}
+
+export default function Example() {
+  const [store] = useState(() => createStore<TestState>({ foo: 0, bar: 0 }));
+  const calls = useRef<SelectorCalls>({
+    state: 0,
+    object: 0,
+    unkeyed: 0,
+    empty: 0,
+    emptyObject: 0,
+  });
+
+  return (
+    <>
+      <KeyedSelectors store={store} calls={calls.current} />
+      <UnkeyedSelector store={store} calls={calls.current} />
+      <EmptySelector store={store} calls={calls.current} />
+      <EmptyObjectSelector store={store} calls={calls.current} />
+      <DirectValues store={store} />
+      <MixedValues store={store} />
+      <OptionalSelector store={store} />
+      <Controls store={store} calls={calls.current} />
+    </>
+  );
+}
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -14,6 +14,7 @@ interface SelectorCalls {
   unkeyed: number;
   empty: number;
   emptyObject: number;
+  conditional: number;
 }
 
 interface StoreProps {
@@ -84,6 +85,28 @@ function EmptyObjectSelector({ store, calls }: SelectorProps) {
     },
   });
   return null;
+}
+
+function ConditionalSelector({ store, calls }: SelectorProps) {
+  const [enabled, setEnabled] = useState(false);
+  const value = useStoreState(store, enabled ? ["foo"] : [], (state) => {
+    calls.conditional += 1;
+    return enabled ? state.foo : null;
+  });
+
+  return (
+    <>
+      <button type="button" onClick={() => setEnabled((value) => !value)}>
+        {enabled ? "Pause conditional selector" : "Enable conditional selector"}
+      </button>
+      <p>
+        Conditional selector value:{" "}
+        <output aria-label="Conditional selector value">
+          {value ?? "paused"}
+        </output>
+      </p>
+    </>
+  );
 }
 
 function MixedValues({ store }: StoreProps) {
@@ -238,6 +261,12 @@ function Controls({ store, calls }: SelectorProps) {
           {getCallCount("emptyObject")}
         </output>
       </p>
+      <p>
+        Conditional selector calls:{" "}
+        <output aria-label="Conditional selector calls">
+          {getCallCount("conditional")}
+        </output>
+      </p>
     </>
   );
 }
@@ -250,6 +279,7 @@ export default function Example() {
     unkeyed: 0,
     empty: 0,
     emptyObject: 0,
+    conditional: 0,
   });
 
   return (
@@ -258,6 +288,7 @@ export default function Example() {
       <UnkeyedSelector store={store} calls={calls.current} />
       <EmptySelector store={store} calls={calls.current} />
       <EmptyObjectSelector store={store} calls={calls.current} />
+      <ConditionalSelector store={store} calls={calls.current} />
       <DirectValues store={store} />
       <MixedValues store={store} />
       <DynamicSelectors store={store} />
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -34,6 +34,7 @@ function KeyedSelectors({ store, calls }: SelectorProps) {
       calls.object += 1;
       return state.foo;
     },
+    runtimeBar: (state) => Reflect.get(state, "bar"),
   });
   const runtimeBar = useStoreState(store, ["foo"], (state) =>
     Reflect.get(state, "bar"),
@@ -47,6 +48,10 @@ function KeyedSelectors({ store, calls }: SelectorProps) {
       <p>
         Object selector value:{" "}
         <output aria-label="Object selector value">{object.value}</output>
+      </p>
+      <p>
+        Object runtime bar:{" "}
+        <output aria-label="Object runtime bar">{object.runtimeBar}</output>
       </p>
       <p>
         Runtime bar: <output aria-label="Runtime bar">{runtimeBar}</output>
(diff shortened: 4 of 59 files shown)


.changeset/200-keyed-store-subscriptions.md
```diff
@@ -0,0 +1,11 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added keyed store subscriptions
```
diegohaz: The title could be `Keyed store subscriptions`

This could also benefit from code examples.

packages/ariakit-react-components/src/combobox/combobox-label.tsx
```diff
@@ -36,7 +36,11 @@ export const useComboboxLabel = createHook<TagName, ComboboxLabelOptions>(
         "ComboboxLabel must receive a `store` prop or be wrapped in a ComboboxProvider component.",
     );
 
-    const comboboxId = useStoreState(store, (state) => state.baseElement?.id);
+    const comboboxId = useStoreState(
+      store,
+      ["baseElement"],
```
diegohaz: **[P2] Observe mutable IDs instead of only element identity**

This subscription is invalidated only when `baseElement` is replaced, but `id` is a mutable DOM attribute. In a public-component repro, changing `<Combobox id>` from `"first"` to `"second"` reused the same element; after a separate store update, `<ComboboxLabel>` still rendered `for="first"` in both Chromium and happy-dom. The same pattern appears in ComboboxCancel, Select/SelectList, TagList/TagListLabel, and TooltipAnchor, so it can also leave `aria-controls` or `aria-labelledby` stale.

Please subscribe to the element key and observe the attribute itself, for example:

```tsx
const baseElement = useStoreState(store, "baseElement");
const observedId = useAttribute(baseElement, "id");
const comboboxId = observedId ?? baseElement?.id;
```

TooltipAnchor can apply the same shape to `contentElement` while subscribing directly to `type`.

app/src/sandbox/store-keyed-subscriptions/index.react.tsx
```diff
@@ -0,0 +1,203 @@
...
+  store: Store<TestState>;
+}
+
+interface SelectorProps extends StoreProps {
+  calls: SelectorCalls;
+}
+
+function KeyedSelectors({ store, calls }: SelectorProps) {
+  const value = useStoreState(store, ["foo"], (state) => {
+    calls.state += 1;
```
diegohaz, line 30: **[P2] Keep the instrumentation deterministic during hydration**

These module-level counters are mutated during render and later emitted as page text. On a fresh Chromium visit, the server rendered `1`, hydration rendered `2`, and React reported a hydration mismatch before regenerating the tree. The current browser test then exercises that remounted tree, which can hide lifecycle or subscription bugs in the fixture.

Please keep the first server/client output identical and assert that the page hydrates without errors. One minimal shape that I verified is:

```tsx
const [mounted, setMounted] = useState(false);
useEffect(() => setMounted(true), []);

const displayedCalls = mounted ? calls.state : 0;
```

Using the deterministic value for the rendered status during hydration removes the page error while preserving the post-mount counters.

packages/ariakit-react-store/src/index.tsx
```diff
@@ -49,16 +44,68 @@ export interface UseState<S> {
...
+type InternalStoreStateObject = Record<string, StoreKey | AnyFunction>;
 
 const noopSubscribe = () => () => {};
 
+function isStoreKeyArray(value: unknown): value is readonly StoreKey[] {
+  return Array.isArray(value);
+}
+
+function isSameStoreKey(key: StoreKey, other: StoreKey | undefined) {
+  return key === other || (key !== key && other !== other);
```
diegohaz, line 63: **[P2] Match `NaN` keys in every subscription path**

This helper deliberately treats two `NaN` keys as equal, but `@ariakit/store`'s multi-listener path still matches updated keys with `===`. As a result, delivery depends on what other listeners happen to exist:

```ts
subscribe(store, [NaN], listener);
subscribe(store, null, () => {}); // selects the live-listener path
store.setState(NaN, 1); // listener is not called
```

Without the all-keys listener, the fast `Map` path does call it because maps use SameValueZero. Please use the same equality semantics in `hasUpdatedKey` (for example, `isSameValue(currentKey, updatedKey)`) and add a focused `@ariakit/store` test so keyed hooks cannot silently miss this update when an unkeyed subscriber is mounted.

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -188,50 +188,54 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
...
-        // widget. Once the activeId is valid, we restore the roving tabindex. See
-        // https://github.com/ariakit/ariakit/issues/3232
-        // https://github.com/ariakit/ariakit/issues/4129
-        const item = store?.item(state.activeId);
-        if (item?.disabled) return true;
-        if (!item?.element) return true;
-        return state.activeId === id;
+    } = useStoreStateObject(
+      store,
+      ["baseElement", "activeId", "renderedItems", "virtualFocus", "items"],
```
diegohaz: **[P2] Keep the row scan off `activeId` updates**

`useStoreStateObject` reruns every mapping whenever any key in this union changes. Therefore each arrow-key `activeId` update makes every CompositeItem execute `ariaPosInSet`, whose body filters all `renderedItems`. A row composite with N items still performs N full N-item scans on the hottest navigation update, so this conversion retains O(N²) selector work that the keyed API can avoid.

Please split the expensive row calculation from the active-item subscription, for example:

```tsx
const ariaPosInSet = useStoreState(
  store,
  ["baseElement", "renderedItems"],
  (state) => {
    // existing row-position calculation
  },
);
```

The remaining active/tabbable selectors can keep the `activeId`-bearing key set without pulling this scan into every focus move.

app/src/sandbox/store-keyed-subscriptions/test.ts
```diff
@@ -0,0 +1,138 @@
...
+    // @ts-expect-error The third argument must be an object.
+    useStoreStateObject(store, ["foo"], (state) => state.foo);
+  };
+
+  expectTypeOf(useCheckTypes).toEqualTypeOf<
+    (optionalStore?: typeof store) => void
+  >();
+});
+
+test("runs selectors only for subscribed keys", async () => {
```
diegohaz, line 99: **[P2] Exercise changing dependency lists**

All runtime cases use fixed literal key arrays, so they do not exercise the new stateful `useStableStoreKeys` and resubscription path when dependencies change. A regression there could leave a hook subscribed to the old key while this test still passes.

Please add a case whose dependency changes at runtime, such as:

```tsx
const [key, setKey] = useState<keyof TestState>("foo");
const value = useStoreState(store, [key], (state) => state[key]);
```

Update `bar` while `foo` is selected, switch to `bar`, assert that the already-written value appears immediately, and then assert that later `bar` updates are observed. A parent-only rerender with a new but equal inline array would also cover the value-stabilization promise in the changeset.

packages/ariakit-react-components/src/tag/tag-list.tsx
```diff
@@ -105,12 +105,20 @@ export const useTagList = createHook<TagName, TagListOptions>(
 
     props = useComposite({ store, ...props });
 
-    const orientation = useStoreState(store, (state) =>
+    const orientation = useStoreState(store, ["orientation"], (state) =>
       state.orientation === "both" ? undefined : state.orientation,
     );
-    const items = useStoreState(store, (state) => state.renderedItems);
+    const items = useStoreState(
```
diegohaz: **[P3] Use the direct key overload here**

This selector is only an identity read, so the new array stabilization and selector invocation add machinery without changing behavior. The direct overload already expresses exactly this dependency and matches the pattern used elsewhere:

```tsx
const items = useStoreState(store, "renderedItems");
```

That is shorter and keeps this call on the simplest keyed-subscription path.

diegohaz (commented):
Initial review of the full merge-base diff (`9da42c9` → `3fa50a3`). I left six inline findings:

- Four P2 correctness/performance issues: mutable DOM IDs can become stale, the sandbox currently hydrates with a mismatch, `NaN` notification semantics differ between store subscription paths, and CompositeItem retains an O(N²) row scan on `activeId` changes.
- One P2 coverage gap for dependency arrays that change at runtime.
- One P3 behavior-preserving simplification using the direct key overload.

I reviewed the react-store API and types, the underlying store notification paths, all converted component call sites, the sandbox tests, the README/reference output, the changeset, and the relevant CI/performance configuration. Focused validation passed with React and React 18 sandbox tests, the Chromium sandbox test, `pnpm tsc`, reference tests, `build-website-lite`, and `git diff --check`. I also reproduced the mutable-ID regression through public components, captured the hydration page error in a fresh browser context, and reproduced the `NaN` listener discrepancy directly.

I did not rerun the entire repository test matrix locally; all current required checks are green. I also did not benchmark the CompositeItem finding—the cost follows directly from every item rerunning the existing O(N) `renderedItems.filter` on each `activeId` notification. The pre-existing unresolved changeset thread about its title/examples still applies, so I did not duplicate it.

.changeset/200-keyed-store-subscriptions.md
diegohaz: Fixed in `da5fbfd22` and refined in `686639bd6` — the release note now uses the “Keyed store subscriptions” title, includes focused TypeScript examples, and keeps the shared entry limited to `useStoreState`. The `useStoreStateObject` example now lives in a separate changeset scoped only to packages that export it.

packages/ariakit-react-components/src/combobox/combobox-label.tsx
diegohaz: Fixed in `d592cee5a` — the affected components now subscribe directly to the element and observe its mutable `id` attribute. The new sandbox covers Combobox, Select, Tag, and Tooltip behavior, including same-node ID changes.

app/src/sandbox/store-keyed-subscriptions/index.react.tsx
diegohaz: Fixed in `f5ff7c24f` — selector counters render a deterministic `0` during SSR and initial hydration, then expose runtime counts after mount. The browser test attaches console/page-error listeners before re-navigation and asserts hydration completes without errors.

packages/ariakit-react-store/src/index.tsx
diegohaz: Fixed in `acbb8b934` — the slow multi-listener path now compares keys with the same SameValueZero semantics as the fast map path. The focused store test covers both paths with a `NaN` key.

packages/ariakit-react-components/src/composite/composite-item.tsx
diegohaz: Fixed in `e45bce41f` — `CompositeItem` now keeps row/ARIA-position calculations in a subscription keyed only by `baseElement` and `renderedItems`, separate from the `activeId` subscription, so focus moves no longer rerun the row-wide scans.

app/src/sandbox/store-keyed-subscriptions/test.ts
diegohaz: Fixed in `f5ff7c24f` — the sandbox now changes the dependency key from `foo` to `bar` at runtime for both hooks, verifies the current `bar` value appears immediately, confirms later `foo` updates are ignored, and confirms later `bar` updates are observed.

packages/ariakit-react-components/src/tag/tag-list.tsx
diegohaz: Fixed in `e45bce41f` — TagList now uses the direct overload: `useStoreState(store, "renderedItems")`.

diegohaz:
**Fixed**

- `da5fbfd22`, `686639bd6` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776025)): improved the release-note title and examples, then separated `useStoreStateObject` into a note scoped to `@ariakit/react-store` and `@ariakit/react-components`.
- `d592cee5a` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776133)): observed mutable element IDs and added cross-component sandbox coverage.
- `f5ff7c24f` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776267)): made selector counters hydration-safe and asserted a clean hydration.
- `acbb8b934` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776377)): aligned `NaN` key matching across both store subscription paths.
- `e45bce41f` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776511)): removed the Composite row scan from `activeId` updates.
- `f5ff7c24f` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776605)): covered dynamic dependency changes for both keyed hooks.
- `e45bce41f` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3574776688)): switched TagList to the direct `renderedItems` overload.

**Invalid**

- None.

**Already resolved**

- None.

packages/ariakit-react-components/src/combobox/combobox-label.tsx
diegohaz: **[P2] Close the render-to-observer gap for mutable IDs**

The ordinary post-effect cases are fixed, but this still uses `useAttribute` only as an invalidation side effect and then rereads `baseElement.id`. `useAttribute` starts at `undefined` and installs its observer in a passive effect. If the ID disappears between render and that effect, its initial callback also computes `undefined`; the state update bails out, so this consumer never rereads the element.

I reproduced this through the public Combobox components:

```tsx
React.useLayoutEffect(() => {
  if (showLabel) inputRef.current?.removeAttribute("id");
}, [showLabel]);

<Ariakit.Combobox ref={inputRef} id="before" />
{showLabel && <Ariakit.ComboboxLabel>Race label</Ariakit.ComboboxLabel>}
```

After mounting the label in Chromium, the input had no ID but the label still had `for="before"`, with no page errors. The same pattern is used at all seven updated call sites. Please make the observed ID a snapshot-authoritative subscription (for example, with `useSyncExternalStore`, which checks the snapshot after subscribing) and add a same-commit layout-effect regression.

packages/ariakit-react-store/src/index.tsx
```diff
@@ -114,55 +171,126 @@ export function useStoreState<T extends CoreStore, V>(
...
+    : typeof keyOrKeysOrSelector === "function"
+      ? null
+      : [keyOrKeysOrSelector];
+  const subscriptionKeys = useStableStoreKeys(keys);
+
   const storeSubscribe = React.useCallback(
     (callback: () => void) => {
       if (!store) return noopSubscribe();
-      return subscribe(store, null, callback);
+      return subscribe(store, subscriptionKeys, callback);
```
diegohaz, line 206: **[P3] Skip registering subscriptions that can never notify**

An empty dependency list is documented to mean that store updates never notify the selector, but both hooks still register that empty listener with the core store. Empty listeners retain callback state and add an entry to every slow-path listener scan when an all-key subscriber is present, even though their callback can never run. In an artificial local check, 20,000 empty listeners plus one all-key listener made 500 updates take about 181.7 ms versus 0.18 ms without the empty listeners.

Both subscription callbacks can preserve the `null` all-key sentinel while avoiding this work:

```ts
if (!store || subscriptionKeys?.length === 0) {
  return noopSubscribe();
}
```

app/src/sandbox/store-keyed-subscriptions/test.ts
diegohaz: **[P2] Make each dynamic hook prove its own resubscription**

The new dynamic fixture renders `useStoreState` and `useStoreStateObject` in the same component. That lets either correct `bar` subscription rerender the component and refresh the other hook, even if that hook is still subscribed only to `foo`.

I verified the masking by temporarily hardcoding only `useStoreState` to subscribe to `["foo"]` while its selector still used `Reflect.get(state, key)`; both the React and Chromium dynamic suites still passed 3/3. Please isolate the hooks behind separate render boundaries and assert their selector calls independently, for example:

```tsx
<DynamicStoreState keyName={keyName} />
<DynamicStoreStateObject keyName={keyName} />
```

After switching to `bar`, an update to `foo` should leave each selector's call count unchanged, while an update to `bar` should advance each one. That makes either broken resubscription fail on its own.

.changeset/300-nan-store-keys.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/store": patch
+"@ariakit/solid-store": patch
```
diegohaz, line 6: **[P2] Include the React public chain in the NaN fix**

The `NaN` key correction also changes the behavior of `useStoreState` / `useStoreStateObject`: `@ariakit/react-store` passes these keys to the core subscription, `@ariakit/react-components` reexports the hooks, and `@ariakit/react` exposes them from its public entry point. With only the core and Solid packages listed here, those React package changelogs will omit this fix.

Please include the affected public packages in this changeset:

```md
"@ariakit/react-store": patch
"@ariakit/react": patch
```

packages/ariakit-react-components/src/composite/composite-row.tsx
```diff
@@ -57,6 +57,7 @@ export const useCompositeRow = createHook<TagName, CompositeRowOptions>(
 
     const baseElement = useStoreState(
       store,
+      ["baseElement"],
```
diegohaz: **[P3] Use the direct key overload for this single-key read**

This selector only reads `baseElement` and normalizes its nullable value, so the new array/selector overload adds stabilization and selector work without changing behavior. The direct overload is simpler and matches the TagList cleanup already made in this PR:

```tsx
const baseElement = useStoreState(store, "baseElement") || undefined;
```

I checked this alternative with `pnpm tsc` and the CompositeRow fixture under both React 19 and React 18.

diegohaz (commented):
## Re-review summary

I re-reviewed the incremental range from `3fa50a3a11785431f766516808006e3fdca2cbcc` through `f5ff7c24fdace209d2af8648a3addade6c827a7c`, then checked the affected surrounding code and public reexports.

I left five actionable comments:

- Two replies on prior threads: the mutable-ID fix still has a render-to-passive-effect race, and the dynamic-dependency tests let either hook mask a broken subscription in the other.
- Three new inline comments: add the React public package chain to the `NaN` changeset, avoid registering empty subscriptions that cannot notify, and simplify CompositeRow's single-key selector to the direct overload.

The previously reported hydration behavior, core `NaN` matching, CompositeItem active-state hot path, TagList direct read, and ordinary post-effect mutable-ID cases are otherwise fixed.

Focused validation passed:

```sh
pnpm test packages/ariakit-store/src/test.ts  # 92/92
pnpm tsc
pnpm build-website-lite
pnpm changeset status
```

The focused sandbox suites passed under React 19 and React 18 (4/4), and the two affected sandbox filters passed across Chrome, Firefox, and Safari (21/21). I also verified the mutable-ID race in Chromium, proved the dynamic-test masking with a deliberately broken subscription, and checked the CompositeRow simplification under both React versions. All current GitHub checks are green.

I did not rerun the entire repository test matrix locally; the broader matrix is represented by the green PR checks. The empty-list timing is an artificial stress check, not a production benchmark.

packages/ariakit-react-store/src/index.tsx
diegohaz: Fixed in [`27911fc2b`](https://github.com/ariakit/ariakit/commit/27911fc2b7e475f3afcef997333a347e0d6a6438): both hooks now return a no-op subscription when their final computed key list is empty. The exact empty-array check preserves `null` as the all-keys sentinel, and direct object mappings still produce nonempty subscription keys.

.changeset/300-nan-store-keys.md
diegohaz: Fixed in [`b2045f951`](https://github.com/ariakit/ariakit/commit/b2045f951034314608fc3d6b4b99653b61d17d06): the NaN changeset now includes `@ariakit/react-store`, `@ariakit/react-components`, and `@ariakit/react`, alongside the existing core and Solid store packages.

packages/ariakit-react-components/src/composite/composite-row.tsx
diegohaz: Fixed in [`ef08a6aab`](https://github.com/ariakit/ariakit/commit/ef08a6aab782503ee21bb3e03c9976348657db5a): CompositeRow now uses `useStoreState(store, "baseElement") || undefined`, preserving the row-context type while removing the unnecessary selector.

diegohaz:
**Fixed**

- [`27911fc2b`](https://github.com/ariakit/ariakit/commit/27911fc2b7e475f3afcef997333a347e0d6a6438) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575222493)): skipped core listener registration when either hook computes an empty dependency list, while preserving all-key and direct-object-key subscriptions.
- [`b2045f951`](https://github.com/ariakit/ariakit/commit/b2045f951034314608fc3d6b4b99653b61d17d06) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575222591)): added the React public package chain to the NaN subscription changeset.
- [`ef08a6aab`](https://github.com/ariakit/ariakit/commit/ef08a6aab782503ee21bb3e03c9976348657db5a) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575222682)): simplified CompositeRow to the direct `baseElement` overload.

**Already resolved**

- None.

**Invalid**

- None.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -261,8 +261,10 @@ export const useComposite = createHook<TagName, CompositeOptions>(
     // solely by the effect below, which is a no-op otherwise, so this avoids
     // re-rendering the composite component on every active item change when
     // moving through items with roving tabindex.
-    const activeId = useStoreState(store, (state) =>
-      state.virtualFocus ? state.activeId : null,
+    const activeId = useStoreState(
+      store,
+      ["virtualFocus", "activeId"],
```
diegohaz: **[P3] Unsubscribe Composite selectors while their gates are closed**

`virtualFocus` only gates the returned value here; the listener still subscribes to `activeId`, so every roving-focus update invokes this selector even though it must return the same `null`. The direct `virtualFocus` subscription above already rerenders when the gate changes, so the dependency list can gate the listener itself:

```tsx
const activeId = useStoreState(
  store,
  virtualFocus ? ["activeId"] : [],
  (state) => (virtualFocus ? state.activeId : null),
);
```

The same pattern applies to `activeItem` while `scheduled` is false, `activeDescendant` outside composite virtual focus, and `focusable` when `activeId` cannot affect its result. I applied conditional empty lists at all four sites locally; `pnpm tsc` and the focused React 19/React 18 suites passed.

app/src/sandbox/store-keyed-subscriptions/index.react.tsx
```diff
@@ -0,0 +1,244 @@
...
+    return state.foo;
+  });
+  const object = useStoreStateObject(store, ["foo"], {
```
diegohaz, line 33: **[P2] Exercise the keyed object overload's separate contracts**

This object fixture always receives a defined store and its selector only reads the declared `foo` key. The `useStoreState` fixture separately proves that selectors receive the complete runtime state and that a keyed optional store works, but neither check exercises the new `useStoreStateObject` implementation. A regression that passes a runtime `Pick` to object selectors or mishandles `undefined -> store -> undefined` would therefore leave every object test green.

Please add an object selector such as `runtimeBar: (state) => Reflect.get(state, "bar")`, then assert that a `bar` update does not notify it and the next `foo` update observes the latest `bar`. The optional overload also needs its own type and lifecycle case, for example:

```ts
const values = useStoreStateObject(optionalStore, ["key"], {
  direct: "otherKey",
  derived: (state) => state?.key,
});

expectTypeOf(values).toEqualTypeOf<{
  direct: string | undefined;
  derived: string | undefined;
}>();
```

I compiled this optional shape with the repository type check.

app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  expect(label).toHaveAttribute("for", "combobox-before");
+  expect(cancel).toHaveAttribute("aria-controls", "combobox-before");
+
+  await click(section.button("Change combobox id"));
+
+  await expect.poll(() => label.getAttribute("for")).toBe("combobox-after");
+  expect(cancel).toHaveAttribute("aria-controls", "combobox-after");
+  expect(combobox).toHaveAttribute("id", "combobox-after");
+});
```
diegohaz, line 18: **[P3] Add the reliable happy-dom ID-mutation cases**

`test-browser.ts` covers Tag and Tooltip ID mutations, but this fast integration file stops after Combobox. Both omitted behaviors are plain DOM/ARIA updates and passed when I mirrored them under both React 19 and React 18, so they should have the repository's expected happy-dom duplicates. For example:

```ts
await click(section.button("Change tooltip id"));
await expect
  .poll(() => anchor?.getAttribute("aria-labelledby"))
  .toBe("tooltip-after");
```

The Select case can remain browser-only: its MutationObserver-driven update did not flush reliably in React 18 happy-dom, while I verified the current PR passes the same case in a real React 18 Chrome preview.

diegohaz (commented):
Re-review of `f5ff7c24fdace209d2af8648a3addade6c827a7c..27911fc2b7e475f3afcef997333a347e0d6a6438`, plus the surrounding merge-base diff and affected public/component context.

I left three findings:

- **P2 tests:** independently cover the keyed `useStoreStateObject` runtime-state and optional-store contracts.
- **P3 performance:** use conditional empty dependency lists for gated Composite selectors so inactive callbacks are not attached to hot `activeId` updates.
- **P3 tests:** mirror the reliable Tag and Tooltip mutable-ID cases in happy-dom; keep Select browser-only because its React 18 happy-dom MutationObserver update was not reliable.

Validation performed:

- Installed the workspace and confirmed Node `v24.18.0`.
- `pnpm test packages/ariakit-store/src/test.ts` — 92/92 passed.
- Focused React 19 and React 18 sandbox suites — 4/4 passed in each matrix.
- `pnpm tsc` — passed.
- Focused reference-extractor tests — 4/4 passed.
- `pnpm changeset status` and `git diff --check` — passed.
- Warm single-worker Chrome/Firefox/Safari sandbox run — 21/21 passed.
- Pristine React 18 Chrome preview for all four mutable-ID cases — 4/4 passed.
- The conditional Composite alternative and optional keyed-object type shape were compiled locally, then all experimental changes were removed.
- Current GitHub checks are green.

Unresolved-thread audit: GraphQL returned 10 review threads and all 10 are resolved, so there were no unresolved candidate findings requiring replies or reclassification in this run.

I did not rerun the entire repository test/build matrix locally; the focused coverage above passed and the full hosted matrix is green.

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Fixed in [`fdf11e2007468abb9fd33c61bfdbfe431725e852`](https://github.com/ariakit/ariakit/commit/fdf11e2007468abb9fd33c61bfdbfe431725e852): gated all four Composite subscriptions with empty dependency lists while their gates are closed, and added public `[] → ["foo"] → []` attach/detach coverage.

app/src/sandbox/store-keyed-subscriptions/index.react.tsx
diegohaz: Fixed in [`cb705c603c7d3eaa585eac2ee48e4904b79d95de`](https://github.com/ariakit/ariakit/commit/cb705c603c7d3eaa585eac2ee48e4904b79d95de): added independent full-runtime-state, optional-store lifecycle/direct-key subscription, and positive/negative type coverage for `useStoreStateObject`. The runtime-state and store-removal assertions were each red-proven against the corresponding broken behavior.

app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts
diegohaz: Fixed the reliable portion in [`9be966d34e2ec258d3e12593936ec48be92381a5`](https://github.com/ariakit/ariakit/commit/9be966d34e2ec258d3e12593936ec48be92381a5): added the Tooltip happy-dom regression and red-proved it against the old selector. I re-tested both Tag relationships under React 18 happy-dom; the DOM IDs change, but `for` and `aria-labelledby` remain at their old values even with polling and the repository async wrapper. Tag therefore remains browser-only, where all 12 sandbox cases pass across Chrome, Firefox, and Safari.

diegohaz:
**Fixed**

- [`fdf11e200`](https://github.com/ariakit/ariakit/commit/fdf11e2007468abb9fd33c61bfdbfe431725e852) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575500123)): gated all four inactive Composite subscriptions and added empty/nonempty dependency transition coverage.
- [`cb705c603`](https://github.com/ariakit/ariakit/commit/cb705c603c7d3eaa585eac2ee48e4904b79d95de) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575500212)): independently covered keyed object full-state selectors, optional-store lifecycle/direct keys, and type contracts.
- [`9be966d34`](https://github.com/ariakit/ariakit/commit/9be966d34e2ec258d3e12593936ec48be92381a5) ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575500271)): added and red-proved the reliable Tooltip happy-dom ID-mutation regression.

**Invalid**

- `Tag happy-dom duplicate` ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575500271)): both Tag relationships stay stale under React 18 happy-dom after the DOM IDs change, so this remains browser-only; the cross-browser sandbox passes 12/12.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -261,8 +263,10 @@ export const useComposite = createHook<TagName, CompositeOptions>(
     // solely by the effect below, which is a no-op otherwise, so this avoids
     // re-rendering the composite component on every active item change when
     // moving through items with roving tabindex.
-    const activeId = useStoreState(store, (state) =>
-      state.virtualFocus ? state.activeId : null,
+    const activeId = useStoreState(
+      store,
+      virtualFocus ? ["activeId"] : [],
```
diegohaz: **[P3] Finish gating the `activeId` subscription**

This still subscribes and returns each changing `activeId` whenever `virtualFocus` is true, even if `composite` is false; the only effect consuming this value exits immediately on `!composite`. That state is common for combobox-backed Menu and Select lists, which disable their own Composite behavior while inheriting the combobox store's default virtual focus, so active-item moves still rerender an inactive list.

When Composite is active, `activeDescendant` also subscribes independently to `activeId`, leaving two selector callbacks on the same hot update. One subscription can serve both consumers:

```tsx
const activeId = useStoreState(
  store,
  composite && virtualFocus ? ["activeId"] : [],
  (state) => (composite && virtualFocus ? state.activeId : null),
);

const activeDescendant = useStoreState(
  store,
  composite && virtualFocus ? ["items"] : [],
  () => {
    if (!store) return;
    if (!composite) return;
    if (!virtualFocus) return;
    return getEnabledItem(store, activeId)?.id;
  },
);
```

The nearby comment should then describe `activeId` as shared by the blur effect and `aria-activedescendant`. I prototyped this consolidation; `pnpm tsc` and the full React 19 and React 18 suites passed.

app/src/sandbox/store-keyed-subscriptions/test-browser.ts
```diff
@@ -0,0 +1,168 @@
...
+    await test.expect(selectorValue).toHaveText("bar:1");
+    await test.expect(objectValue).toHaveText("bar:1");
+
+    await q.button("Update bar").click();
+
+    await test.expect(selectorValue).toHaveText("bar:2");
+    await test.expect(objectValue).toHaveText("bar:2");
+  });
+
+  test("attaches and detaches conditional selector dependencies", async ({
```
diegohaz, line 131: **[P3] Refresh the PR testing counts**

This is now the fourth keyed-store browser case, so the three desktop projects run 12 keyed-store tests rather than the 9 listed in the PR description. The generic keyed-store sandbox also now has four tests per React version rather than three. Please update those two Testing entries so the recorded matrix matches this head; the focused cross-browser command lists 24 total cases across the keyed-store and mutable-ID sandboxes, 12 from each.

diegohaz (commented):
Re-review of `27911fc2b7e475f3afcef997333a347e0d6a6438..fdf11e2007468abb9fd33c61bfdbfe431725e852`, plus the surrounding Composite and sandbox context.

The keyed-object contract coverage and reliable Tooltip happy-dom duplicate are complete. The conditional-key fixture also correctly covers `[] -> ["foo"] -> []` attachment and detachment. I left two P3 findings:

- Finish gating the Composite `activeId` subscription on `composite` and reuse that selected value so `activeDescendant` does not attach a second `activeId` listener.
- Refresh the PR Testing counts: the keyed-store sandbox now has four cases per React version and 12 browser cases, not three and nine.

Validation performed:

- `pnpm install` and `pnpm exec node --version` (`v24.18.0`).
- `pnpm tsc` — passed.
- Focused React 19 sandboxes — 6/6 passed.
- Focused React 18 keyed-store cases — 4/4 passed; the mutable-ID file rerun passed 2/2.
- Full-suite prototype of the consolidated Composite alternative — React 19: 193 files / 853 tests; React 18: 193 files / 845 passed and 8 expected skips.
- Chrome, Firefox, and Safari sandbox run — 24/24 passed.
- `git diff --check 27911fc2b..HEAD` — passed.
- All current GitHub checks are green.

One isolated combined React 18 run had a transient timeout in the pre-existing Combobox ID poll; the file passed immediately when rerun alone, and both the full hosted React 18 matrix and real-browser coverage are green.

Unresolved-thread audit: GraphQL returned 13 review threads and all 13 are resolved, so there were no unresolved candidate findings requiring replies or reclassification in this run.

I did not rerun the package/app/website builds locally; the focused checks above passed and the hosted build matrix is green.

app/src/sandbox/store-keyed-subscriptions/test-browser.ts
diegohaz: Updated the Testing section to report 4 keyed-store tests per React version and 12 keyed-store browser tests (24 browser cases combined with the mutable-ID sandbox). Thanks for catching the stale counts.

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Implemented this consolidation in 326e24822: `activeId` is now gated by `composite && virtualFocus`, and `aria-activedescendant` subscribes only to `items` while deriving from the shared `activeId` value. The full React 19 and React 18 suites, `pnpm tsc`, and `pnpm lint` pass.

diegohaz:
Addressed the latest review round.

**Fixed**

- Finished gating the Composite `activeId` subscription on `composite && virtualFocus` and consolidated the `aria-activedescendant` listener so it subscribes only to `items` ([commit](https://github.com/ariakit/ariakit/commit/326e24822), [reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575618851)).
- Refreshed the PR testing counts to 4 keyed-store tests per React version and 12 keyed-store browser tests, 24 browser cases combined with the mutable-ID sandbox ([reply](https://github.com/ariakit/ariakit/pull/6717#discussion_r3575602875)).

**Verified**

- `pnpm test-react --run` — 193 files, 853 tests
- `pnpm test-react18 --run` — 193 files, 845 passed and 8 expected skips
- `pnpm tsc`
- `pnpm lint`
- `git diff --check origin/keyed-store-subscriptions..HEAD`

diegohaz (commented):
Re-review of `fdf11e2007468abb9fd33c61bfdbfe431725e852..326e24822f881a437ff2f36c9a579c8177f68117`, plus the surrounding Composite call sites, `useStoreState` implementation, and existing virtual-focus coverage.

No new actionable findings. The follow-up correctly gates the shared `activeId` snapshot on `composite && virtualFocus`; that snapshot triggers the render that recomputes `aria-activedescendant`, while the remaining `items` subscription still covers registration/removal with a stable active ID. I also considered direct derivation and a single `useStoreStateObject`, but the former misses item-only updates and the latter adds coupled selector/object work without a clear advantage.

Validation performed:

- `pnpm install` and `pnpm exec node --version` (`v24.18.0`).
- `pnpm tsc` — passed.
- `pnpm test-react --run` — 193 files / 853 tests passed.
- `pnpm test-react18 --run` — 193 files / 845 passed and 8 expected skips on the full rerun.
- Focused Chrome, Firefox, and Safari coverage for the keyed-store, mutable-ID, and Select virtual-focus sandboxes — 30/30 passed.
- `git diff --check fdf11e200..HEAD` — passed.
- All current hosted checks are green, including React 18 and `composite-perf`.

One earlier local React 18 full run timed out in the new Tooltip-ID happy-dom poll; the file passed 2/2 immediately in isolation, the full suite then passed on rerun, and the authoritative real-browser and hosted React 18 checks are green. I found no evidence that the Composite increment caused that transient failure.

Unresolved-thread audit: GraphQL returned 15 review threads and all 15 are resolved, so there were no unresolved candidate findings requiring replies or reclassification in this run.

I did not rerun the package/app/website builds locally; the focused checks above passed and the hosted build matrix is green.

PR #6721, opened 2026-07-14
Reduce React store subscriptions and fix controlled NaN values

diegohaz:
## Motivation

React store wrappers registered a change listener even when the corresponding setter prop was absent. These unused listeners accumulate across store-backed widgets. `DisclosureContent`, which underpins components such as `TabPanel` and `Dialog`, also read four related state keys through four separate subscriptions.

Store-prop comparisons used strict equality, so a controlled `NaN` value appeared to change even when the store value was also `NaN`. This could call the setter unnecessarily:

```tsx
useStoreProps(store, { value: Number.NaN, setValue }, "value", "setValue");
```

## Solution

Skip the `useStoreProps` listener when no setter is provided. When a setter becomes available, attach a non-initializing keyed subscription and read the latest callback through a live ref, so replacing the callback does not require resubscribing:

```ts
if (!hasSetValue) return;
return subscribe(store, [key], listener);
```

Compare store values with SameValue-style semantics, including treating two `NaN` values as equal. Read `open`, `mounted`, `animated`, and `contentElement` in `DisclosureContent` through one keyed object subscription.

The `store-keyed-subscriptions` sandbox observes the active listener count and verifies `0 → 1 → 1 → 0 → 1` across setter attachment, replacement, removal, and reattachment. It also covers controlled `NaN` behavior in happy-dom and real browsers.

## Performance

The [exact-head performance report](https://github.com/ariakit/ariakit/pull/6721#issuecomment-4967040682) detected no significant Chrome or Node timing changes. The performance claims are structural only: setter-less `useStoreProps` hooks register no listener, and four related `DisclosureContent` state subscriptions become one.

## Testing

- Full published-package build.
- Production Next.js and Astro app builds.
- TypeScript, lint, formatting, and diff checks.
- React 19: 8 focused tests passed.
- React 18: 8 focused tests passed.
- Chrome, Firefox, and Safari: 30 focused browser tests passed.
- Exact-head performance workflow: Node and all seven Chrome shards passed.

### DIFF
--- a/.changeset/200-react-store-subscriptions.md
+++ b/.changeset/200-react-store-subscriptions.md
@@ -4,6 +4,4 @@
 "@ariakit/react": patch
 ---
 
-Improved React store subscription performance
-
-Store hooks now avoid registering change listeners when no setter callback is provided. [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) also reads related state through a single keyed subscription, improving navigation performance in widgets powered by [`Composite`](https://ariakit.com/reference/composite), including [`Tab`](https://ariakit.com/reference/tab) components.
+Reduced React store subscription overhead by skipping listeners when setter callbacks are absent and reading four related [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) state values through one subscription, including in [`TabPanel`](https://ariakit.com/reference/tab-panel) and [`Dialog`](https://ariakit.com/reference/dialog) components.
--- a/.changeset/200-react-store-subscriptions.md
+++ b/.changeset/200-react-store-subscriptions.md
@@ -1,0 +1,9 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved React store subscription performance
+
+Store hooks now avoid registering change listeners when no setter callback is provided. [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) also reads related state through a single keyed subscription, improving navigation performance in widgets powered by [`Composite`](https://ariakit.com/reference/composite), including [`Tab`](https://ariakit.com/reference/tab) components.
--- a/packages/ariakit-components/src/composite/composite-store.ts
+++ b/packages/ariakit-components/src/composite/composite-store.ts
@@ -78,6 +78,13 @@ interface FindEnabledItemIdParams {
   excludeId?: string;
 }
 
+// Keep this lookup outside getNextId. Inlining its callback there regresses
+// explicit activeId navigation in benchmarks because that path shares a
+// closure with the mutable navigation index.
+function findItemIndex(items: CompositeStoreItem[], id: string) {
+  return items.findIndex((item) => item.id === id);
+}
+
 function findEnabledItemId({
   items,
   fromIndex,
@@ -90,9 +97,8 @@ function findEnabledItemId({
     if (!item) continue;
     if (item.rowId !== rowId) continue;
     if (item.disabled) continue;
-    const itemId = item.id;
-    if (excludeId != null && itemId === excludeId) continue;
-    return itemId;
+    if (excludeId != null && item.id === excludeId) continue;
+    return item.id;
   }
   return undefined;
 }
@@ -305,6 +311,9 @@ export function createCompositeStore<
   };
 
   const composite = createStore(initialState, collection, props.store);
+  // Sequential navigation usually makes the returned item active before the
+  // next call, so remember that item's index without retaining the array.
+  let cachedItemIndex = -1;
 
   // When the activeId is undefined, we need to find the first enabled item and
   // set it as the activeId.
@@ -352,8 +361,19 @@ export function createCompositeStore<
         ? true
         : !canShift && !renderedItems.some((item) => item.rowId != null);
       if (canFastScan) {
+        // Explicit active ids may jump arbitrarily, so preserve their direct
+        // lookup path and cache only navigation driven by store state.
+        const canCache =
+          options.activeId === undefined &&
+          renderedItems === defaultState.renderedItems;
         let activeIndex = -1;
-        if (renderedItems === defaultState.renderedItems) {
+        if (canCache && renderedItems[cachedItemIndex]?.id === activeId) {
+          activeIndex = cachedItemIndex;
+        }
+        if (
+          activeIndex === -1 &&
+          renderedItems === defaultState.renderedItems
+        ) {
           const firstItem = renderedItems[0];
           // Avoid scanning twice when the rendered items were replaced with
           // cloned or external objects that aren't in the collection map.
@@ -365,7 +385,7 @@ export function createCompositeStore<
           }
         }
         if (activeIndex === -1) {
-          activeIndex = renderedItems.findIndex((item) => item.id === activeId);
+          activeIndex = findItemIndex(renderedItems, activeId);
         }
         const activeItem = renderedItems[activeIndex];
         if (activeItem) {
@@ -377,7 +397,12 @@ export function createCompositeStore<
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
-          if (nextId !== undefined) return nextId;
+          if (nextId !== undefined) {
+            if (canCache) {
+              cachedItemIndex = activeIndex + step;
+            }
+            return nextId;
+          }
           const canLoop =
             focusLoop &&
             (isVerticalDirection
@@ -387,13 +412,17 @@ export function createCompositeStore<
           // Wrap around to the beginning (or end, when scanning backward) of
           // the same row, matching the flipItems behavior in the generic
           // logic below.
-          return findEnabledItemId({
+          const loopId = findEnabledItemId({
             items: renderedItems,
             fromIndex: step === 1 ? 0 : renderedItems.length - 1,
             step,
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
+          if (canCache) {
+            cachedItemIndex = step === 1 ? 0 : renderedItems.length - 1;
+          }
+          return loopId;
         }
       }
     }
--- a/packages/ariakit-components/src/composite/composite-store.ts
+++ b/packages/ariakit-components/src/composite/composite-store.ts
@@ -69,7 +69,7 @@ function getItemsInRow(items: CompositeStoreItem[], rowId?: string) {
   return items.filter((item) => item.rowId === rowId);
 }
 
-interface FindEnabledItemIdParams {
+interface FindEnabledItemIndexParams {
   items: CompositeStoreItem[];
   fromIndex: number;
   /** 1 to scan forward, -1 to scan backward. */
@@ -85,20 +85,20 @@ function findItemIndex(items: CompositeStoreItem[], id: string) {
   return items.findIndex((item) => item.id === id);
 }
 
-function findEnabledItemId({
+function findEnabledItemIndex({
   items,
   fromIndex,
   step,
   rowId,
   excludeId,
-}: FindEnabledItemIdParams) {
+}: FindEnabledItemIndexParams) {
   for (let i = fromIndex; i >= 0 && i < items.length; i += step) {
     const item = items[i];
     if (!item) continue;
     if (item.rowId !== rowId) continue;
     if (item.disabled) continue;
     if (excludeId != null && item.id === excludeId) continue;
-    return item.id;
+    return i;
   }
   return undefined;
 }
@@ -390,18 +390,18 @@ export function createCompositeStore<
         const activeItem = renderedItems[activeIndex];
         if (activeItem) {
           const step: 1 | -1 = canReverse ? -1 : 1;
-          const nextId = findEnabledItemId({
+          const nextIndex = findEnabledItemIndex({
             items: renderedItems,
             fromIndex: activeIndex + step,
             step,
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
-          if (nextId !== undefined) {
+          if (nextIndex !== undefined) {
             if (canCache) {
-              cachedItemIndex = activeIndex + step;
+              cachedItemIndex = nextIndex;
             }
-            return nextId;
+            return renderedItems[nextIndex]?.id;
           }
           const canLoop =
             focusLoop &&
@@ -412,17 +412,20 @@ export function createCompositeStore<
           // Wrap around to the beginning (or end, when scanning backward) of
           // the same row, matching the flipItems behavior in the generic
           // logic below.
-          const loopId = findEnabledItemId({
+          const loopIndex = findEnabledItemIndex({
             items: renderedItems,
             fromIndex: step === 1 ? 0 : renderedItems.length - 1,
             step,
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
-          if (canCache) {
-            cachedItemIndex = step === 1 ? 0 : renderedItems.length - 1;
+          if (loopIndex !== undefined) {
+            if (canCache) {
+              cachedItemIndex = loopIndex;
+            }
+            return renderedItems[loopIndex]?.id;
           }
-          return loopId;
+          return undefined;
         }
       }
     }
--- a/packages/ariakit-components/src/composite/composite-store.ts
+++ b/packages/ariakit-components/src/composite/composite-store.ts
@@ -69,7 +69,7 @@ function getItemsInRow(items: CompositeStoreItem[], rowId?: string) {
   return items.filter((item) => item.rowId === rowId);
 }
 
-interface FindEnabledItemIndexParams {
+interface FindEnabledItemParams {
   items: CompositeStoreItem[];
   fromIndex: number;
   /** 1 to scan forward, -1 to scan backward. */
@@ -85,13 +85,31 @@ function findItemIndex(items: CompositeStoreItem[], id: string) {
   return items.findIndex((item) => item.id === id);
 }
 
+function findEnabledItemId({
+  items,
+  fromIndex,
+  step,
+  rowId,
+  excludeId,
+}: FindEnabledItemParams) {
+  for (let i = fromIndex; i >= 0 && i < items.length; i += step) {
+    const item = items[i];
+    if (!item) continue;
+    if (item.rowId !== rowId) continue;
+    if (item.disabled) continue;
+    if (excludeId != null && item.id === excludeId) continue;
+    return item.id;
+  }
+  return undefined;
+}
+
 function findEnabledItemIndex({
   items,
   fromIndex,
   step,
   rowId,
   excludeId,
-}: FindEnabledItemIndexParams) {
+}: FindEnabledItemParams) {
   for (let i = fromIndex; i >= 0 && i < items.length; i += step) {
     const item = items[i];
     if (!item) continue;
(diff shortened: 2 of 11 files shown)


packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -377,7 +397,12 @@ export function createCompositeStore<
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
-          if (nextId !== undefined) return nextId;
+          if (nextId !== undefined) {
+            if (canCache) {
+              cachedItemIndex = activeIndex + step;
```
diegohaz: **Cache the index returned by the scan**

`findEnabledItemId` can skip disabled items or items from another row, but this stores the adjacent slot instead of the slot whose ID was returned. On the next state-driven move, the ID-at-index guard misses and cloned/external arrays repeat the linear `findIndex`. A full traversal with alternating disabled items therefore remains O(n²).

I verified this on the exact PR head with 100 traversals of 400 cloned items: the current code read item IDs 4,039,699 times. A no-allocation prototype that made the scanner return its found index, cached that index here and at the loop site, and then read the ID used 59,899 reads. The same prototype improved a 399-item/200-enabled traversal from about 11.4k to 80.7k walks/sec locally, while keeping the all-enabled sequential benchmarks at about 33k ops/sec. It also passed `pnpm tsc`, all 30 Composite/Tab store tests, and 50,000 randomized comparisons against the uncached path.

A concrete shape is:

```ts
const nextIndex = findEnabledItemIndex({ /* scan options */ });
if (nextIndex !== undefined) {
  if (canCache) {
    cachedItemIndex = nextIndex;
  }
  return renderedItems[nextIndex]?.id;
}
```

Please apply the same idea to `loopId` below and add a state-driven sequential benchmark with disabled items, since the two new sequential benchmarks contain only enabled items and cannot expose this miss pattern.

diegohaz (commented):
Initial full review of `9b15bc9110d71fd549bec2f8200fa8fd8102104f..b33b0fbbba877ccc6030fe6da363d422d48f65bc`. I reviewed the complete diff plus the surrounding Composite and Collection store implementations, React keyboard-navigation call sites, Tab integration and public package propagation, the existing unreleased performance changeset, benchmark infrastructure/results, and current CI state.

One substantive performance/scaling finding was published inline: cache the index actually returned after skipping disabled or out-of-row items, and add a sequential-disabled benchmark. The current adjacent-slot cache preserves O(n²) traversal for those supported layouts even though an allocation-free exact-index variant preserves behavior and all-enabled performance.

Validation performed:

- `pnpm test packages/ariakit-components/src/composite/composite-store.test.ts packages/ariakit-components/src/tab/tab-store.test.ts` (30 tests passed).
- `pnpm test-react examples/composite/test.ts examples/tab/test.ts` and the equivalent `pnpm test-react18` run (6 tests passed in each runtime).
- `pnpm tsc`.
- Full head benchmark and focused sequential benchmarks (about 33k ops/sec for canonical and cloned items).
- 50,000 randomized cached-versus-uncached navigation comparisons across direction/config changes, disabled items, loops, row metadata, replacements, clones, and in-place reorderings.
- A temporary exact-index prototype, fully restored afterward: 30 tests, TypeScript, 50,000 differential comparisons, focused all-enabled benchmarks, deterministic ID-access counts, and a disabled-item timing probe.
- `git diff --check`, package/re-export tracing, and HTTP 200 verification for both changeset reference links.

Relevant unresolved candidate findings: none. The authoritative `reviewThreads` connection was empty at both the start of the review and the pre-publish refresh.

I found no additional correctness, API/backward-compatibility, security, package-propagation, or release-metadata issue. I did not rerun the production browser performance harness locally; I inspected the completed CI Chrome and paired Node reports instead. All PR checks were green at review time.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`502bf573c`](https://github.com/ariakit/ariakit/commit/502bf573c16b45269396ac2756cec98fbdf0957b): cache the exact enabled-item index returned by normal and looping scans, and cover state-driven traversal across disabled cloned items with a benchmark.

diegohaz:
**Fixed**

- [`502bf573c`](https://github.com/ariakit/ariakit/commit/502bf573c16b45269396ac2756cec98fbdf0957b) ([reply](https://github.com/ariakit/ariakit/pull/6721#discussion_r3577569120)): Cached the exact enabled-item index returned by normal and looping scans, and added a state-driven disabled-items benchmark.

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -365,19 +385,24 @@ export function createCompositeStore<
...
             step,
             rowId: activeItem.rowId,
             excludeId: activeId,
           });
-          if (nextId !== undefined) return nextId;
+          if (nextIndex !== undefined) {
+            if (canCache) {
+              cachedItemIndex = nextIndex;
+            }
+            return renderedItems[nextIndex]?.id;
```
diegohaz: **[P2] Preserve the ID-returning path for explicit navigation**

`canCache` is false whenever callers pass `options.activeId`, so this path cannot reuse `nextIndex`; it now pays for both the index-producing scan and a second `renderedItems[nextIndex]` access. The exact Node report for [run 29321135938](https://github.com/ariakit/ariakit/actions/runs/29321135938) confirmed `move through shuffled composite items` at 67,024 to 59,043 ops/sec (-11.9%, 5/5 rounds) and `move through 8 composite items` at 3,137,756 to 2,719,266 ops/sec (-13.3%, 5/5). The [current-head report](https://github.com/ariakit/ariakit/actions/runs/29326266468) reconfirms -12.7% (5/5) and -11.0% (4/5). Those benchmarks call `next({ activeId })`, as Select and TabPanel do. Chrome reported no confirmed end-to-end regression, but these are repeatable warnings in the PR's own hot-path benchmark.

Please retain the index-returning scan only for the cacheable state-driven path and preserve a direct ID-returning scanner for non-cacheable normal and loop scans. For example:

```ts
if (canCache) {
  const nextIndex = findEnabledItemIndex(scanOptions);
  // Cache the exact index and return its ID.
} else {
  const nextId = findEnabledItemId(scanOptions);
  if (nextId !== undefined) return nextId;
}
```

packages/ariakit-react-store/src/index.tsx
```diff
@@ -407,18 +407,20 @@ export function useStoreProps<
 >(store: CoreStore<S>, props: P, key: K, setKey?: SK) {
   const value = hasOwnProperty(props, key) ? props[key] : undefined;
   const setValue = setKey ? props[setKey] : undefined;
+  const hasSetValue = !!setValue;
   const propsRef = useLiveRef({ value, setValue });
 
   // Calls setValue when the state value changes.
   useSafeLayoutEffect(() => {
-    return sync(store, [key], (state, prev) => {
+    if (!hasSetValue) return;
```
diegohaz, line 415: **[P2] Add a regression assertion for listener removal**

The new lifecycle assertions only observe setter output after a store update, so they do not prove the main optimization here: that no listener is registered while the setter is absent. I mutation-tested the exact head by replacing this gated `subscribe` with the previous always-registered `sync` listener while retaining the new SameValue comparisons. All five sandbox tests still passed in both React 19 and React 18, including attach, replace, remove, and reattach.

Please add focused instrumentation at the hook boundary, using a mocked subscription or an equivalent count, so the optimization itself fails under that mutation. The lifecycle should assert a sequence like:

```ts
expect(activeSubscriptions).toBe(0);
enableSetter();
expect(activeSubscriptions).toBe(1);
clearSetter();
expect(activeSubscriptions).toBe(0);
enableSetter();
expect(activeSubscriptions).toBe(1);
```

It should also assert immediately after attachment/replacement that the new non-initializing subscription has not called the setter.

packages/ariakit-react-store/src/index.tsx
```diff
@@ -407,18 +407,20 @@ export function useStoreProps<
...
+    return subscribe(store, [key], (state, prev) => {
       const { value, setValue } = propsRef.current;
       if (!setValue) return;
-      if (state[key] === prev[key]) return;
-      if (state[key] === value) return;
+      if (isSameValue(state[key], prev[key])) return;
+      if (isSameValue(state[key], value)) return;
```
diegohaz, line 420: **[P2] Add a patch changeset for controlled `NaN` behavior**

This comparison fixes a consumer-visible setter bug, not just subscription overhead. With the previous `===` checks, resetting a store from `1` to a controlled `NaN` could call the setter again because `NaN !== NaN`; the new assertion at `test.ts:247-251` specifically proves that the callback count stays at one. The existing changeset only advertises performance, so this fix would be absent from the release notes.

Please add a separate `300-*.md` bug-fix entry and include every public package through which the behavior is shipped:

```md
---
"@ariakit/react-store": patch
"@ariakit/react": patch
---

Fixed setter callbacks firing unnecessarily for controlled `NaN` store values.
```

.changeset/200-react-store-subscriptions.md
```diff
@@ -0,0 +1,9 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Improved React store subscription performance
+
+Store hooks now avoid registering change listeners when no setter callback is provided. [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) also reads related state through a single keyed subscription, improving navigation performance in widgets powered by [`Composite`](https://ariakit.com/reference/composite), including [`Tab`](https://ariakit.com/reference/tab) components.
```
diegohaz: **[P2] Keep the release note to the verified subscription impact**

The phrase "improving navigation performance" attributes an end-to-end result to the Disclosure subscription aggregation, but the exact Chrome report says no significant performance changes were detected; the Tab rows were approximately +2% for mount and -2% for movement. The implementation proves fewer subscriptions and listeners, not improved navigation latency. This two-paragraph entry also has room to name all affected `DisclosureContent` derivatives, but it names only Tab while the direct consumers also include `TabPanel`, `Dialog`, and `DialogBackdrop`.

Please either supply matching end-to-end evidence or make this a single verified paragraph with up to two representative derived components, for example:

```md
Reduced React store subscription overhead by skipping listeners when setter callbacks are absent and reading related [`DisclosureContent`](https://ariakit.com/reference/disclosure-content) state through one subscription, including in [`TabPanel`](https://ariakit.com/reference/tab-panel) and [`Dialog`](https://ariakit.com/reference/dialog) components.
```

diegohaz (commented):
Re-review of the changes from the previous marked head `b33b0fbbba877ccc6030fe6da363d422d48f65bc` through `db05a399e9b5b6a238c18a81f5e72ab574d5c5f3`, plus the surrounding Composite, React store, Disclosure, test, benchmark, and release-note context.

I found four actionable items and posted each inline:

- Preserve an ID-returning scanner for non-cacheable explicit-`activeId` navigation; the exact Node CI reports twice confirm roughly 11-13% regressions in that path.
- Add a mutation-sensitive assertion for listener registration and disposal, not only setter output.
- Add a separate patch changeset for the controlled-`NaN` setter bug fix.
- Limit the subscription changeset to measured effects and describe representative `DisclosureContent` derivatives.

Validation performed:

- `pnpm test packages/ariakit-components/src/composite/composite-store.test.ts` (21 passed).
- `pnpm test-react app/src/sandbox/store-keyed-subscriptions/test.ts` (5 passed).
- `pnpm test-react18 app/src/sandbox/store-keyed-subscriptions/test.ts` (5 passed).
- Mutation check replacing the gated listener with the old always-registered `sync` path while retaining SameValue comparisons (all 5 focused tests still passed in React 19 and React 18).
- `git diff --check 9b15bc9110d71fd549bec2f8200fa8fd8102104f..db05a399e9b5b6a238c18a81f5e72ab574d5c5f3`.
- Raw exact performance artifacts from runs `29321135938` and `29326266468`, including the five-round Node confirmation data and Chrome reports.
- Current PR checks are green.

Immediately before publishing, I queried the authoritative GraphQL `reviewThreads` connection through exhaustion, including each thread's comments connection. There were no pre-existing unresolved review threads. The prior marked exact-index cache thread is resolved and its original miss issue is addressed; the first inline finding above concerns a separate regression introduced by that follow-up implementation.

I did not rerun the full browser matrix locally or run the exact CI performance workflow against the proposed scanner split, so that alternative still needs fresh CI confirmation.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`552449e97`](https://github.com/ariakit/ariakit/commit/552449e976c2945372bd7d4d71c5c80275e1bc03): explicit and custom navigation now use direct ID-returning normal and loop scans, while only state-driven navigation pays for and caches an index.

.changeset/200-react-store-subscriptions.md
diegohaz: Fixed in [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0): the release note is now one verified paragraph about listener/subscription overhead, with `TabPanel` and `Dialog` as representative `DisclosureContent` consumers and no end-to-end navigation claim.

packages/ariakit-react-store/src/index.tsx
diegohaz: Fixed in [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0): added a separate patch changeset for controlled `NaN` setter behavior across `@ariakit/react-store`, `@ariakit/react-components`, and `@ariakit/react`.

packages/ariakit-react-store/src/index.tsx
diegohaz: Fixed in [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0): the sandbox now observes real `subscribe` and legacy `sync` registrations, asserts `0 → 1 → 1 → 0 → 1`, and verifies attachment and replacement do not initialize either setter. Restoring the old unconditional `sync` listener fails the new initial-count assertion.

diegohaz:
**Fixed**

- [`552449e97`](https://github.com/ariakit/ariakit/commit/552449e976c2945372bd7d4d71c5c80275e1bc03) ([reply](https://github.com/ariakit/ariakit/pull/6721#discussion_r3583085753)): restored direct ID-returning scans for explicit and custom Composite navigation while retaining state-driven index caching.
- [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0) ([reply](https://github.com/ariakit/ariakit/pull/6721#discussion_r3583085768)): added deterministic real-listener instrumentation and attachment/replacement assertions.
- [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0) ([reply](https://github.com/ariakit/ariakit/pull/6721#discussion_r3583085760)): added a separate controlled-`NaN` patch changeset.
- [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0) ([reply](https://github.com/ariakit/ariakit/pull/6721#discussion_r3583085754)): limited the performance changeset to verified listener/subscription overhead and representative Disclosure consumers.

diegohaz (commented):
Re-review from the latest prior marked head [`db05a399e`](https://github.com/ariakit/ariakit/commit/db05a399e9b5b6a238c18a81f5e72ab574d5c5f3) through [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0), including the surrounding Composite navigation, React store subscription, sandbox-test, benchmark, public-package, and release-note context.

The branch was force-rebased onto current `main`. `git range-diff` confirms that the three previously reviewed commits are patch-equivalent to [`86e3dce63`](https://github.com/ariakit/ariakit/commit/86e3dce63f62f041622f72d3aed0d3ead6d51bfe), [`a925d0500`](https://github.com/ariakit/ariakit/commit/a925d050073a4d6333d26f99e99dbfac0349cfab), and [`1c5787484`](https://github.com/ariakit/ariakit/commit/1c5787484acb3d9b7f66bf563c3ebecd5ec11b6d); the substantive incremental scope is the two follow-ups [`552449e97`](https://github.com/ariakit/ariakit/commit/552449e976c2945372bd7d4d71c5c80275e1bc03) and [`672b2e230`](https://github.com/ariakit/ariakit/commit/672b2e2303ec4f137744664e893d08ff52c04ec0).

No new substantive findings.

Prior-thread validation:

- `PRRT_kwDOBjT0Fc6Q6P9o` remains resolved. Explicit `activeId` and custom-item navigation now use direct ID-returning normal and loop scans, while state-driven navigation retains the exact-index cache. Exact-head Node CI removes the prior explicit-path warnings: shuffled calls are -2% and 8-item calls are +2%, while sequential canonical, cloned, and disabled-cloned gains remain confirmed at +28%, +108%, and +107% across five paired rounds.
- `PRRT_kwDOBjT0Fc6Q6QDH` remains resolved. The sandbox observes both real `subscribe` and legacy `sync` registrations, asserts `0 → 1 → 1 → 0 → 1`, and checks that attaching or replacing a setter is non-initializing. In an isolated exact-head worktree, restoring the unconditional `sync` listener while retaining the NaN-aware comparisons made the intended test fail at the initial count with `expected 0, received 1`.
- `PRRT_kwDOBjT0Fc6Q6QL9` remains resolved. A separate `300-` patch changeset now documents controlled-`NaN` setter behavior across `@ariakit/react-store`, `@ariakit/react-components`, and `@ariakit/react`; the documented Checkbox and Radio hooks accept numeric values, call `useStoreProps`, and are publicly re-exported.
- `PRRT_kwDOBjT0Fc6Q6QQ3` remains resolved. The subscription changeset is now one verified structural claim and names `TabPanel` and `Dialog` as representative `DisclosureContent` consumers without claiming end-to-end latency gains.
- The earlier exact-index cache thread `PRRT_kwDOBjT0Fc6QreR4` also remains resolved after the rebase.

Validation performed:

- `pnpm test packages/ariakit-components/src/composite/composite-store.test.ts packages/ariakit-components/src/tab/tab-store.test.ts` (30 passed).
- `pnpm test-react app/src/sandbox/select-6319/test.ts app/src/sandbox/tab-panel-6332/test.ts` (4 passed).
- `pnpm test-react app/src/sandbox/store-keyed-subscriptions/test.ts` (5 passed).
- `pnpm test-react18 app/src/sandbox/store-keyed-subscriptions/test.ts` (5 passed).
- Isolated red mutation of the old unconditional listener, then full restoration and worktree cleanup.
- `git diff --check` across the two incremental follow-up commits.
- HTTP 200 verification for the `DisclosureContent`, `TabPanel`, `Dialog`, `useCheckboxStore`, and `useRadioStore` reference links, plus package/export and direct-consumer tracing.
- Exact-head performance workflow [run 29373454828](https://github.com/ariakit/ariakit/actions/runs/29373454828), its raw Node and seven Chrome artifacts, and the published report. Chrome reports no significant changes; all PR checks are green.

Immediately before publishing, I paginated the authoritative GraphQL `reviewThreads` connection and every comments connection to exhaustion. All five threads are resolved and there are no unresolved candidate findings.

I did not rerun the full browser matrix, full build, or full TypeScript suite locally. Their exact-head CI jobs passed, and I inspected the raw final performance artifacts rather than attempting to reproduce CI-shaped performance evidence locally.

PR #6734, opened 2026-07-15
Fix dev server cleanup

diegohaz:
## Motivation

Astro 7 detects agentic environments and may start `astro dev` as a detached background process. Stopping Ariakit's coordinated dev job could then terminate Next.js while leaving Astro listening, so later runs selected unexpected higher ports.

```sh
APP_PORT=4327 NEXTJS_PORT=3007 pnpm dev-app-lite
# Stopping the job should release both ports.
```

## Solution

Keep Astro in the existing shared foreground process group by explicitly disabling its agent background mode in the direct Concurrently spawn.

```ts
const child = spawn("conc", commands, {
  stdio: "inherit",
  env: {
    ...process.env,
    ASTRO_DEV_BACKGROUND: "0",
    APP_PORT: String(appPort),
    NEXTJS_PORT: String(nextjsPort),
  },
});
```

Because Concurrently, Next.js, and Astro now remain in the same foreground group, terminal and job-group signals reach the whole tree naturally. No detached supervisor, IPC forwarding, custom job-control handling, or native-Windows rejection is needed.

The regression starts two observable TCP servers, models Astro's agent daemonization when the override is absent, and verifies that both ports are released after group `SIGINT` and `SIGKILL`. The test is red without the override and green with it. The focused file passes 4/4, the core suite passes 587/587, TypeScript and lint checks pass, and real `pnpm dev-app-lite` runs returned HTTP 200 from both services before releasing both ports under each signal.


packages/ariakit-scripts/src/dev.ts
```diff
@@ -128,7 +128,7 @@ export function watchPackageChanges(options: WatchPackageChangesOptions = {}) {
     timeout = setTimeout(runClean, 100);
   };
 
-  watchPackages("packages", {
+  return watchPackages("packages", {
```
diegohaz: [P2] Drain package cleanup before close resolves

`watchPackageChanges` returns the raw Chokidar watcher, so `close()` only disposes Chokidar. The closure-owned `timeout` and `queue` remain active. A source event just before `conc` exits therefore lets `dev()` finish its `finally` block while `runClean()` is still scheduled or running; the timer can also keep the CLI alive.

I reproduced this at this head: after emitting one watched change, `await handle.close()` returned with `cleanCalls === 0`, then `cleanCalls === 1` after 150 ms. A lifecycle handle that owns all resources preserves pending export updates:

```ts
async close() {
  await watcher.close();
  clearTimeout(timeout);
  runClean();
  await queue;
}
```

This smaller wrapper passed the focused suite and the `@ariakit/scripts` type check locally. Please add a close-focused test as well.

packages/ariakit-scripts/src/dev.ts
```diff
@@ -172,33 +172,41 @@ async function findAvailablePort(
...
+  try {
+    const appPort = await findAvailablePort(
+      Number(process.env.APP_PORT) || 4321,
+    );
+    const nextjsPort = await findAvailablePort(
+      Number(process.env.NEXTJS_PORT) || 3000,
+      [appPort],
+    );
+
+    const exitCode = await runCommand(
```
diegohaz: [P2] Avoid forwarding the terminal signal twice

`runCommand` installs a parent `SIGINT` handler that calls `child.kill()`, but `conc` shares the foreground process group, so a terminal `SIGINT` already reaches it directly. I reproduced one process-group `SIGINT` producing two signals in the coordinator; with the actual Concurrently 10.0.3 coordinator, each command received three signals because both coordinator deliveries were forwarded again. The fake coordinator masks this with `if (stopping) return`, but real cleanup handlers do not have to make repeated signals harmless.

Windows is riskier: [Node 24 documents](https://nodejs.org/docs/latest-v24.x/api/child_process.html#subprocesskillsignal) that `child.kill("SIGINT")` terminates the target forcefully and abruptly rather than delivering a graceful signal.

For this long-running coordinator, keep the parent listener so Ariakit waits for `conc`, but avoid forwarding when the child already shares the terminal console/process group; alternatively, isolate the child and forward exactly once. Please add a regression that asserts the coordinator and commands each observe one signal.

packages/ariakit-scripts/src/dev.test.ts
```diff
@@ -64,6 +105,117 @@ test("ignores package watch paths outside source and package json", () => {
...
+      detached: true,
+      env: {
+        ...process.env,
+        ARGS_PATH: argsPath,
+        PATH: commandPath,
+        STOPPED_PATH: stoppedPath,
+      },
+      stdio: "ignore",
+    });
+    const exitPromise = once(child, "exit");
```
diegohaz: [P2] Bound the exit wait so failure cleanup can run

This `exitPromise` has no timeout. If the behavior under test regresses and the detached command never exits after `SIGINT`, Vitest can time out the outer test, but it cannot cancel this pending `await`. Execution therefore never reaches the `finally` block that kills the process group and removes the temporary workspace, leaving the failing test able to leak a process and hang its worker.

Attach the bound when the listener is created, before sending the signal:

```ts
const exitPromise = once(child, "exit", {
  signal: AbortSignal.timeout(2000),
});
```

Then the rejection reaches `finally`, where the forced-termination path can reap the child before removing the directory.

diegohaz (commented):
Initial review of `bd7b6779dc742914af2ee6bd52cf953fe2d74f26` → `befb58e00728cb3a319689d2732dac082dbaa18e`.

Scope reviewed:

- Full diff and surrounding code in `dev.ts` and `dev.test.ts`.
- The `runCommand` implementation and call sites, CLI error handling, package-clean behavior, Chokidar 5.0.0, Concurrently 10.0.3, Astro 7.0.7, package/release metadata, and CI state.
- Correctness, signal/process lifecycle, test failure paths, cross-platform behavior, performance, and simpler lifecycle ownership.

Validation performed:

- `pnpm test packages/ariakit-scripts/src/dev.test.ts` — 4/4 passed.
- `pnpm test` — 54 files and 587 tests passed; the command exited 0 after printing an unrelated post-run `ECONNREFUSED` diagnostic.
- `pnpm tsc --pretty false` and the focused `@ariakit/scripts` type check — passed.
- All current PR checks — passed.
- A watcher repro showed `close()` resolving with zero clean calls and the debounced clean starting later. A prototype lifecycle wrapper that closes Chokidar, clears the timer, flushes pending work, and awaits the queue passed the focused tests and type check.
- A process-group repro showed one `SIGINT` reaching the coordinator twice. Repeating it with the actual Concurrently coordinator delivered three signals to each command.
- The Astro/signal subprocess is red on the base; applying only the environment override still leaves it red on the lifecycle assertion. The missing-coordinator test passes the base through its unhandled `ENOENT`, but a head mutant without watcher close times out, so it protects the new architecture without being a base-red proof.

Review-thread validation:

- Unresolved thread `PRRT_kwDOBjT0Fc6Q_TJC` / comment `3584804909` is **valid and still present**: both failure cleanups can remove the temporary workspace before forced child termination completes. I did not duplicate that accurate thread.
- The earlier top-level watcher-lifecycle candidate is **valid and still present**; the local repro and prototype above independently confirmed it, and I anchored the actionable detail inline.

Findings:

- Three new P2 inline findings: drain the watcher-owned timer/queue, avoid duplicate signal forwarding (including the abrupt Windows path), and bound the subprocess exit wait so test cleanup runs on failure.
- One additional valid P2 cleanup race remains covered by the existing unresolved thread.
- No changeset is required because `@ariakit/scripts` is private and explicitly ignored by Changesets. I found no separate performance or release-metadata regression.

Not fully validated: I did not execute the lifecycle on a Windows host or run a full real Astro/Next.js dev session. The duplicate POSIX delivery was reproduced locally with Concurrently, and the Windows concern is grounded in the Node 24 `ChildProcess.kill()` contract.

packages/ariakit-scripts/src/dev.ts
diegohaz: Fixed in [`2e7751f85`](https://github.com/ariakit/ariakit/commit/2e7751f850f1048fd1f621f3e025dd5cbfe967a5). On POSIX, `ariakit dev` now isolates Concurrently and forwards terminal/job-control signals exactly once; Windows keeps the shared console without forceful `child.kill("SIGINT")`. The subprocess regression asserts one signal per coordinator/command and covers stop/resume plus SIGINT/SIGQUIT shutdown.

packages/ariakit-scripts/src/dev.test.ts
diegohaz: Fixed in [`2e7751f85`](https://github.com/ariakit/ariakit/commit/2e7751f850f1048fd1f621f3e025dd5cbfe967a5). The exit listener now uses `AbortSignal.timeout(2000)` before the shutdown signal is sent, so a hung regression reaches the forced-termination `finally` path instead of leaking the process group.

packages/ariakit-scripts/src/dev.ts
diegohaz: Fixed in [`2e7751f85`](https://github.com/ariakit/ariakit/commit/2e7751f850f1048fd1f621f3e025dd5cbfe967a5). `watchPackageChanges` now returns a lifecycle handle whose `close()` closes Chokidar, clears the debounce timer, flushes pending package paths, and awaits the cleanup queue. The new deferred-clean regression proves `close()` does not resolve early.

packages/ariakit-scripts/src/dev.test.ts
```diff
@@ -64,6 +105,117 @@ test("ignores package watch paths outside source and package json", () => {
...
+      killProcessGroup(pid, "SIGINT");
+      const [exitCode, signal] = await exitPromise;
+      expect([exitCode, signal]).toEqual([0, null]);
+      await expect(
+        access(stoppedPath),
+        `dev exited with ${exitCode ?? signal} before cleanup finished`,
+      ).resolves.toBeUndefined();
+    } finally {
+      killProcessGroup(pid, "SIGKILL");
+      await rm(rootPath, { recursive: true, force: true });
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`2e7751f85`](https://github.com/ariakit/ariakit/commit/2e7751f850f1048fd1f621f3e025dd5cbfe967a5). Both teardown paths now subscribe before forced termination, await the child exit, and only then remove the temporary workspace. The detached coordinator group is also killed independently so a wrapper failure cannot skip cleanup.

diegohaz:
**Fixed**

- Watcher shutdown now clears the debounce, flushes pending packages, and awaits the cleanup queue ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585222906)).
- POSIX coordinator isolation now gives each process one terminal signal and preserves stop/resume behavior ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585222901)).
- Subprocess exit waits are bounded and forced children are reaped before temporary workspace removal ([cleanup thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585222907), [timeout thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585222905)).
- Landed in [`2e7751f85`](https://github.com/ariakit/ariakit/commit/2e7751f850f1048fd1f621f3e025dd5cbfe967a5).

**Already resolved**

- None.

**Invalid**

- None.

packages/ariakit-scripts/src/utils.ts
```diff
@@ -79,15 +118,32 @@ export async function runCommand(
...
+      process.on("SIGTERM", onSigterm);
+      if (options.forwardSignals !== false && process.platform !== "win32") {
+        process.on("SIGCONT", onSigcont);
+        process.on("SIGQUIT", onSigquit);
+        process.on("SIGTSTP", onSigtstp);
+      }
+    } else {
+      process.once("SIGINT", onSigint);
+      process.once("SIGTERM", onSigterm);
+    }
```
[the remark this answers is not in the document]
diegohaz: I independently reproduced this on the current head. After sending `SIGINT` to a wrapper using `{ detached: true, forwardSignals: false }`, both the wrapper and its isolated child were still alive 250 ms later:

```json
{"runnerAlive":true,"childAlive":true}
```

No current production call site constructs this pair, so this is a latent invalid API state rather than an active `dev` regression. Rejecting the combination—or replacing the coupled booleans with a mode that makes it unrepresentable—is the safe fix. Merely skipping the parent handlers would restore the wrapper's default termination but can strand the detached child outside its process group.

packages/ariakit-scripts/src/dev.ts
```diff
@@ -172,33 +190,47 @@ async function findAvailablePort(
...
+      "conc",
+      [
+        "-r",
+        `pnpm -F app run dev --port ${appPort}`,
+        `pnpm -F nextjs run dev --port ${nextjsPort}`,
+      ],
+      {
+        cwd: process.cwd(),
+        detached: isolateCoordinator,
```
diegohaz: [P2] Keep the coordinator inside the force-kill boundary

On POSIX, `detached: true` moves `conc` into a new process group/session. Forwarding works for catchable signals only while the Ariakit wrapper is running; if a terminal or agent escalates a hung shutdown to `SIGKILL`, or the wrapper crashes, killing the wrapper's entire job group can no longer reach the coordinator or its servers.

I reproduced this with the exact `{ detached: true, forwardSignals: true }` pair used here: after `SIGKILL` terminated the wrapper group, the detached child was still alive.

```json
{"wrapperSignal":"SIGKILL","coordinatorAlive":true}
```

That can reintroduce the occupied-port leak this PR is intended to fix. Please either keep the coordinator in the shell/runner job group and suppress duplicate forwarding there (a shared-group prototype delivered one `SIGINT` and exited cleanly), or add reliable parent-death supervision. A regression should force-kill the wrapper job and verify the coordinator PID or both dev ports are gone.

diegohaz (commented):
Re-review of `befb58e00728cb3a319689d2732dac082dbaa18e` → `2e7751f850f1048fd1f621f3e025dd5cbfe967a5`, using the latest prior marked review as the incremental baseline.

Scope reviewed:

- The incremental and full PR diffs in `dev.ts`, `dev.test.ts`, and `utils.ts`, plus every `runCommand` call site.
- Watcher lifecycle ownership, POSIX process groups and job control, Windows signal assumptions, forced termination, test failure cleanup, Concurrently 10.0.3 signal behavior, performance, release metadata, and simpler shared-group alternatives.
- Relevant unresolved threads were independently validated against the current head.

Validation performed:

- `pnpm test packages/ariakit-scripts/src/dev.test.ts` — 6/6 passed.
- `pnpm test` — 54 files and 589 tests passed; the command exited 0 after the post-run `ECONNREFUSED` diagnostic.
- `pnpm tsc --pretty false` and `git diff --check` — passed.
- All current PR checks — passed.
- Exact signal repros confirmed that `detached: true, forwardSignals: false` leaves both processes alive after `SIGINT`, and that the production `true/true` topology leaves the coordinator alive after the wrapper job receives `SIGKILL`. A shared-group/no-forwarding prototype delivered one `SIGINT` and exited 0.

Review-thread outcomes:

- `PRRT_kwDOBjT0Fc6RAjQ2` / `3585265139` is **valid and still present**. No caller currently uses the invalid detached/no-forwarding pair, but the options API permits it. I replied with the repro and clarified that rejecting or making the pair unrepresentable is safer than merely removing the parent handlers.
- `PRRT_kwDOBjT0Fc6RAsW-` / `3585316484` is **valid and still present**, but it is an environment-sensitive test-fixture defect rather than a demonstrated live-server leak. The fake SIGQUIT coordinator can exit before reaping its commands; a non-reaping Linux PID 1 can then keep zombie group members observable. The focused test passed locally and in the current GitHub Ubuntu run, so I did not duplicate the accurate existing thread.
- The four findings from the prior marked review are resolved on this head: watcher close drains its timer/queue, signal delivery is asserted once, the post-shutdown wait is bounded, and forced child teardown waits before workspace removal.

Findings:

- One new P2 inline finding: POSIX detachment moves `conc` outside the shell/runner force-kill boundary, so a wrapper crash or `SIGKILL` escalation can leave the coordinator and dev servers running.
- No additional correctness, performance, simplification, or release-metadata finding survived validation. `@ariakit/scripts` is private and still requires no changeset.

Not fully validated: Windows behavior remains untested because the process-group suite skips Windows and repository CI has no Windows job. I also did not reproduce the zombie timeout on this macOS host; the fixture lifecycle and Linux failure condition were verified from the current code.

packages/ariakit-scripts/src/utils.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924). The coupled `detached`/`forwardSignals` booleans are replaced by explicit signal modes, so the unsafe detached-without-forwarding state is no longer representable. The supervised mode also buffers startup signals over IPC before launching the isolated coordinator.

packages/ariakit-scripts/src/dev.test.ts
```diff
@@ -64,6 +186,211 @@ test("ignores package watch paths outside source and package json", () => {
...
+    writeFileSync(process.env.STOPPED_PATH, "");
+    process.exit(0);
+  });
+};
+process.on("SIGCONT", recordSignal);
+process.on("SIGINT", stop);
+process.on("SIGQUIT", (signal) => {
+  recordSignal(signal);
+  setTimeout(() => process.exit(131), 100);
+});
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924). The fake SIGQUIT coordinator now awaits both command children before exiting, so the process-group assertion does not depend on orphan-zombie reaping. The integration suite retains the full group-exit check.

packages/ariakit-scripts/src/dev.ts
diegohaz: Fixed in [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924). A detached POSIX supervisor now stays outside the coordinator group and watches the wrapper IPC channel. Wrapper SIGKILL closes IPC, resumes and terminates the coordinator, and escalates to group SIGKILL after a bounded grace period. The regression stops the coordinator first, kills the wrapper job, and verifies the coordinator group exits with one cleanup signal per fake process.

diegohaz:
**Fixed**

- [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585932833)): Replaced the unsafe detached/no-forwarding option pair with explicit signal modes and startup-safe IPC forwarding.
- [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585933009)): Reaped fake SIGQUIT children before coordinator exit.
- [`ece7957e1`](https://github.com/ariakit/ariakit/commit/ece7957e1b31e21293e699658c9ff6171818f924) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3585933152)): Added parent-death supervision so wrapper SIGKILL releases the coordinator and dev ports.

packages/ariakit-scripts/src/command-supervisor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  let spawnError = false;
+  coordinator.once("error", (error) => {
+    spawnError = true;
+    console.error(error);
+  });
+  coordinator.once("close", (code, signal) => {
+    clearTimeout(shutdownTimer);
+    cleanup();
+    process.exitCode = spawnError ? 1 : getExitCode(code, signal);
```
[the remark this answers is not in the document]
diegohaz: I reproduced this on `ece7957e1b31e21293e699658c9ff6171818f924` with a supervised coordinator that exits immediately on `SIGTERM` while a same-process-group descendant ignores it:

```js
process.on("SIGTERM", () => process.exit(0));
```

After killing the wrapper process group and waiting 900 ms, the current code produced:

```json
{"coordinatorAlive":false,"grandchildAlive":true}
```

The current fake `conc` does not cover this path because it explicitly waits for both children before exiting. Please keep the fallback group kill armed and add a regression where the coordinator closes before an uncooperative descendant; otherwise the parent-death supervisor can still leave the ports it was added to contain.

packages/ariakit-scripts/src/utils.ts
```diff
@@ -79,15 +171,25 @@ export async function runCommand(
...
         return;
       }
       resolvePromise(1);
     });
 
-    process.once("SIGINT", onSigint);
-    process.once("SIGTERM", onSigterm);
+    if (signalMode !== "supervised") {
+      process.once("SIGINT", onSigint);
+      process.once("SIGTERM", onSigterm);
```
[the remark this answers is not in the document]
diegohaz: I reproduced the underlying code behavior on POSIX: a PID-only `SIGTERM` in shared mode left both the wrapper and child alive. However, this finding and its proposed fix are inaccurate for the active path: `dev()` selects `"shared"` only on Windows, and no repository caller selects it on POSIX.

On Windows, Node documents that programmatic `SIGINT`/`SIGTERM` delivery unconditionally terminates the target process rather than invoking and being swallowed by this listener, while console Ctrl+C is delivered across the shared console. Removing the listener therefore does not fix targeted Windows tree cleanup, and it can make the wrapper exit before `runCommand()` and the surrounding `finally` finish on the normal shared-console path. If targeted Windows termination needs containment, that requires parent-death/tree supervision rather than skipping these handlers.

References: [Node signal events](https://nodejs.org/api/process.html#signal-events), [Windows console control delivery](https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent).

packages/ariakit-scripts/src/utils.ts
```diff
@@ -51,25 +58,110 @@ export async function runCommand(
...
+
   return await new Promise<number>((resolvePromise, reject) => {
-    const child = spawn(command, args, {
-      cwd: options.cwd,
-      env: options.env,
-      stdio: "inherit",
-    });
+    let child: ChildProcess | undefined;
+    let stopRequest = 0;
+    const pendingSignals: PendingSignal[] = [];
```
diegohaz: **[P3] Remove unreachable startup signal queues**

Node signal callbacks cannot interrupt the synchronous JavaScript stack containing `fork()`/`spawn()`. Those calls assign `child` before `runCommand()` yields, so `forwardSignal()` cannot observe an unset child. The same applies to the supervisor's `message` callback and `coordinatorPid`. This makes both `pendingSignals` arrays, their `!child`/missing-PID branches, and their drain loops dead state.

I removed those queues locally and all seven lifecycle tests still passed, including the delayed-supervisor startup stop/resume case. Please simplify this path to direct delivery, with the equivalent change in `command-supervisor.ts`:

```ts
const forwardSignal = (signal: NodeJS.Signals, onSent?: () => void) => {
  if (signalMode === "shared") return;
  deliverSignal({ signal, onSent });
};
```

IPC/OS event queuing still preserves the startup behavior exercised by the test.

diegohaz (commented):
Re-review of `2e7751f850f1048fd1f621f3e025dd5cbfe967a5` → `ece7957e1b31e21293e699658c9ff6171818f924`, using the latest prior marked review as the incremental baseline.

Scope reviewed:

- The incremental and full PR diffs in `command-supervisor.ts`, `dev.ts`, `dev.test.ts`, and `utils.ts`, plus every `runCommand` call site, package/runtime resolution, Concurrently shutdown behavior, watcher ownership, POSIX/Windows signal semantics, performance, release metadata, and simpler process topologies.
- Relevant unresolved candidate threads were independently validated against the current head.

Validation performed:

- `pnpm test packages/ariakit-scripts/src/dev.test.ts` — 7/7 passed; the same suite also passed after locally removing both startup signal queues.
- `pnpm test` — 54 files and 590 tests passed; the command exited 0 after the known post-run `ECONNREFUSED` diagnostic.
- `pnpm tsc --pretty false`, `pnpm lint`, `git diff --check`, and all current PR checks — passed.
- Exact process repros confirmed the coordinator-close descendant leak and the latent POSIX shared-mode behavior. All spawned repro processes were reaped and the worktree was restored clean.

Review-thread outcomes:

- `PRRT_kwDOBjT0Fc6RCjMn` / `3585992682` is **valid and still present**. Clearing the parent-disconnect fallback timer on coordinator close leaves an ignoring same-group descendant alive. I replied with the exact reproduction and missing regression case.
- `PRRT_kwDOBjT0Fc6RCpNv` / `3586026775` is **inaccurate for the active Windows-only shared path**. The POSIX observation is real but unreachable from current call sites; Windows programmatic termination does not get swallowed by this listener, and removing the handler does not provide targeted tree cleanup. I replied with the platform distinction and safer decision path.
- The prior marked review's option-state, fake-child reaping, watcher-drain, duplicate-delivery, bounded-wait, and original force-kill-boundary findings are resolved on this head, subject to the narrower timer race above.

Findings:

- One existing P2 process-cleanup defect remains in thread `3585992682`.
- One new P3 maintainability finding: the two startup `pendingSignals` queues and drain paths are unreachable and can be removed; the behavior-preserving simplification passed the focused lifecycle suite.
- No additional correctness, performance, packaging, or release-metadata finding survived validation. `@ariakit/scripts` remains private and requires no changeset.

Not fully validated: I did not execute the lifecycle on a Windows host or rerun a real Astro/Next.js dev session.

packages/ariakit-scripts/src/utils.ts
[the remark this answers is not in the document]
diegohaz: Maybe we simply don't support windows?

packages/ariakit-scripts/src/command-supervisor.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`2f0bea89b`](https://github.com/ariakit/ariakit/commit/2f0bea89b5006bb4e800132bb84704a4ec71c061): the parent-disconnect fallback now remains referenced for the full grace period even if the coordinator exits first, so its final group SIGKILL reaches same-group descendants. The regression uses an early-exiting coordinator and a SIGTERM-ignoring descendant; it timed out on the previous implementation and now passes.

packages/ariakit-scripts/src/utils.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`6f1a73e90`](https://github.com/ariakit/ariakit/commit/6f1a73e903e56f980a2d2f2cc620c559cc62df2f): `ariakit dev` now rejects native Windows with WSL guidance, and the untested shared signal mode has been removed. WSL continues through the supervised POSIX path.

packages/ariakit-scripts/src/utils.ts
diegohaz: Fixed in [`6f1a73e90`](https://github.com/ariakit/ariakit/commit/6f1a73e903e56f980a2d2f2cc620c559cc62df2f): removed both unreachable startup signal queues and their drain paths while retaining the spawn-failure PID guard. The delayed-supervisor lifecycle test still verifies IPC startup buffering and passes without the userland queues.

diegohaz:
**Fixed**

- [`2f0bea89b`](https://github.com/ariakit/ariakit/commit/2f0bea89b5006bb4e800132bb84704a4ec71c061) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3586612288)): Kept the parent-death group fallback armed after early coordinator exit and covered an uncooperative descendant.
- [`6f1a73e90`](https://github.com/ariakit/ariakit/commit/6f1a73e903e56f980a2d2f2cc620c559cc62df2f) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3586612483)): Rejected unsupported native-Windows dev runs with WSL guidance and removed shared mode.
- [`6f1a73e90`](https://github.com/ariakit/ariakit/commit/6f1a73e903e56f980a2d2f2cc620c559cc62df2f) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3586612673)): Removed unreachable startup signal queues while preserving IPC startup coverage and spawn-failure guards.

packages/ariakit-scripts/src/dev.test.ts
```diff
@@ -64,6 +216,304 @@ test("ignores package watch paths outside source and package json", () => {
...
+const recordSignal = (signal) => {
+  appendFileSync(process.env.CONC_SIGNALS_PATH, signal + "\\n");
+};
+const stop = (signal) => {
+  recordSignal(signal);
+  for (const child of children) {
+    child.kill(signal);
+  }
+  if (process.env.CONC_EXITS_BEFORE_CHILDREN) {
+    process.exit(0);
```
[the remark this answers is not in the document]
diegohaz: I confirm the fixture-level issue, but not the P1 severity. In the forced-shutdown case, fake `conc` signals both children and immediately exits:

```js
for (const child of children) {
  child.kill(signal);
}
if (process.env.CONC_EXITS_BEFORE_CHILDREN) {
  process.exit(0);
}
```

Whether the later `SIGKILL` leaves zombies visible to `kill(-pgid, 0)` depends on the adopter. The targeted case passed locally (1 passed, 7 skipped), the full file passed 8/8, and the current Ubuntu job also passed this file 8/8 and the suite 591/591. A non-reaping container PID 1 can still exhibit the reported timeout, so this is a valid environment-dependent test-portability issue, not a demonstrated production supervisor failure.

Simply awaiting the children in fake `conc` would remove the required coordinator-exits-before-descendant condition and stop the regression from testing the bug. A dedicated reaper/subreaper fixture, or tracking the known child PIDs and treating missing/zombie states as terminated, preserves that condition. The PR description should also stop claiming all fake children are reaped, because this branch intentionally does not reap them.

packages/ariakit-scripts/src/command-supervisor.ts
```diff
@@ -0,0 +1,131 @@
...
+    const pid = coordinatorPid;
+    if (!pid) return;
+
+    shutdownStarted = true;
+    sendGroupSignal(pid, "SIGCONT");
+    sendSignal(pid, "SIGTERM");
+    // Keep the supervisor alive through the grace period even if the
+    // coordinator exits before one of its descendants.
+    setTimeout(() => {
+      sendGroupSignal(pid, "SIGKILL");
```
diegohaz: **[P2] Do not retain a recyclable PID/PGID after coordinator close**

This timer intentionally outlives the coordinator. Once its `close` event fires, however, the numeric PID and process-group ID are no longer pinned. If the original group drains before 500 ms and that number is reused, the delayed `sendGroupSignal(pid, "SIGKILL")` can kill an unrelated new group; if the negative group signal gets `ESRCH`, the helper at lines 23–25 is even more dangerous because it retries the positive PID. [Node explicitly warns](https://nodejs.org/api/child_process.html#subprocesskillsignal) that signaling an exited child's reassigned PID targets the new process.

I prototyped clearing the timer when the coordinator closes during disconnect shutdown and immediately issuing a group-only kill:

```ts
if (shutdownStarted && coordinatorPid) {
  clearTimeout(shutdownTimer);
  sendSignal(-coordinatorPid, "SIGKILL");
}
```

All eight lifecycle tests passed. This preserves the grace period while the coordinator still pins the identity, then contains any group it abandoned. If descendants must retain the full grace period after their group leader exits, the supervisor needs a stable sentinel/job primitive; a bare numeric PGID cannot provide that identity guarantee. At minimum, group signaling should never fall back to the positive PID because every supervised coordinator is a detached POSIX group leader.

diegohaz (commented):
Re-review of `ece7957e1b31e21293e699658c9ff6171818f924` → `6f1a73e903e56f980a2d2f2cc620c559cc62df2f`, using the latest prior marked review as the incremental baseline.

Scope reviewed:

- The incremental and full PR diffs in `command-supervisor.ts`, `dev.ts`, `dev.test.ts`, and `utils.ts`, plus every `runCommand` call site, package/runtime resolution, process-group lifecycle, watcher ownership, Windows/WSL support decision, performance, release metadata, and simpler shutdown alternatives.
- Every relevant unresolved candidate thread was read completely and independently validated against the current head.

Validation performed:

- `pnpm test packages/ariakit-scripts/src/dev.test.ts` — 8/8 passed; the targeted SIGKILL case passed separately.
- Reintroducing the prior `clearTimeout` behavior made the SIGKILL regression fail for the expected reason: `Timed out waiting for process group`. The restored current implementation passed.
- A safer close-path prototype that cancels the delayed timer and immediately performs a group-only kill passed all 8 lifecycle tests. Removing the unnecessary positive-PID fallback and dead shutdown guard also passed 8/8.
- `pnpm test` — 54 files and 591 tests passed; the command exited 0 after the known post-run `ECONNREFUSED` diagnostic.
- `pnpm tsc --pretty false`, `pnpm lint`, `git diff --check`, and all current PR checks — passed.
- All spawned test/repro processes were reaped and the worktree was restored clean.

Review-thread outcomes:

- `PRRT_kwDOBjT0Fc6RCjMn` / `3585992682` is **resolved**. The fallback remains referenced after coordinator close, and the new test is proven red against the prior behavior and green on this head.
- `PRRT_kwDOBjT0Fc6RCpNv` / `3586026775` is **resolved** by removing shared mode and explicitly rejecting native Windows with WSL guidance.
- `PRRT_kwDOBjT0Fc6RDXl2` / `3586288027` is **resolved**. Both unreachable startup queues and their drain paths are gone; startup IPC buffering remains covered.
- `PRRT_kwDOBjT0Fc6REZCg` / `3586655529` is **valid but environment-dependent**, with P1 severity unsupported. The fixture intentionally orphans children, but the timeout requires a non-reaping adopter; the exact target passes locally and in current Ubuntu CI. I replied with a fix direction that preserves the early-coordinator-exit regression.

Findings:

- One new P2 safety finding: the always-armed timer retains a bare numeric PID/PGID after the coordinator closes, so delayed `SIGKILL` can target a recycled process or group. The existing `shutdownStarted` state can gate the validated immediate close-path cleanup; if that design is not adopted, the state is otherwise dead and can be removed.
- No additional correctness, performance, packaging, or release-metadata finding survived validation. `@ariakit/scripts` remains private and requires no changeset.

Not fully validated: I did not run on native Windows, under a Linux PID 1 that does not reap orphans, or through another real Astro/Next.js dev session.

packages/ariakit-scripts/src/dev.test.ts
```diff
@@ -64,6 +159,135 @@ test("ignores package watch paths outside source and package json", () => {
...
+      const exitPromise = once(child, "exit", {
+        signal: AbortSignal.timeout(5000),
+      });
+      killProcessGroup(pid, signal);
+      await exitPromise;
+      await waitForPortsAvailable(ports);
+    } finally {
+      await forceKillAndWait(child, pid);
+      const processIds = await readProcessIds(processIdsPath);
+      for (let index = 0; index < ports.length; index += 1) {
```
diegohaz: [P2] Preserve child cleanup when readiness fails

`ports` is assigned only after both readiness files appear, but this loop is bounded by `ports.length`. If either `waitForFile` rejects, `ports` is still empty, so the cleanup ignores every PID already recorded in `process-ids.txt`. In the red topology this test is meant to exercise, Astro is outside the wrapper group, so `forceKillAndWait()` cannot reach it. I reproduced this by withholding the Next.js readiness marker in the red state; the app server remained orphaned after Vitest exited.

Please capture `ready.txt` and parse `ports` before waiting for both server markers (the fake `conc` writes it immediately after `process-ids.txt`), or otherwise iterate the recorded PIDs independently, so the `finally` block can clean the detached server on early failures. For example, retain the existing validation but reorder the waits:

```ts
await waitForFile(readyPath);
ports = (await readFile(readyPath, "utf8")).split("\n").map(Number);
await Promise.all([
  waitForFile(appReadyPath),
  waitForFile(nextjsReadyPath),
]);
```

packages/ariakit-scripts/src/dev.ts
```diff
@@ -194,6 +194,8 @@ export async function dev(options: DevOptions = {}) {
       stdio: "inherit",
       env: {
         ...process.env,
+        // Prevent Astro from daemonizing when an agent is detected.
+        ASTRO_DEV_BACKGROUND: "0",
```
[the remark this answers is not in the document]
diegohaz: I independently reproduced the PID-only behavior: killing only the wrapper PID leaves its same-group descendants alive, while killing the foreground job group releases the port. I do not think that makes this an in-scope P2 for this PR.

The base version already had the same PID-only behavior for `conc` and Next.js. The regression here is narrower: Astro 7 detects the agent and removes itself from Ariakit’s existing foreground job, so stopping that job no longer stops both servers. The override restores the original topology:

```ts
env: {
  ...process.env,
  ASTRO_DEV_BACKGROUND: "0",
  APP_PORT: String(appPort),
  NEXTJS_PORT: String(nextjsPort),
}
```

A wrapper cannot catch its own `SIGKILL`. Guaranteeing cleanup after `kill -9 <wrapper pid>` would therefore require the detached supervisor, a platform-specific parent-death primitive, or an external process manager. That is a materially broader contract than terminal/job-group shutdown and is the complexity this rewrite intentionally removes. My disposition is: real POSIX behavior, but out of scope as a required Ariakit dev guarantee; the group `SIGINT`/`SIGKILL` coverage matches the intended contract.

diegohaz (commented):
Re-review of `6f1a73e903e56f980a2d2f2cc620c559cc62df2f` → `a52ffc0f724f493201e5e818ea68cbf44feb487a`, using the latest prior marked review as the incremental baseline.

The rewrite is the right simplification for `ariakit dev`. The production change is now just:

```ts
ASTRO_DEV_BACKGROUND: "0",
```

That keeps Astro in the existing foreground job with Concurrently and Next.js. The removed detached supervisor, IPC forwarding, custom job-control handling, native-Windows rejection, and PID-only parent-death guarantee are not needed to fix the reported regression.

I left one actionable test-safety finding inline: an early readiness failure leaves `ports` empty, so the `finally` cleanup can skip a detached child whose PID was already recorded. I reproduced the resulting orphan and suggested parsing the already-written port file before the readiness waits.

I considered replacing the 226-line process fixture with a much shorter assertion on the environment passed to `spawn`, but that would only test the literal implementation. It would not prove that the red topology leaves a port occupied or that both observable ports are released after stopping the foreground job. The behavior-level fixture therefore earns its overall scope once its early-failure cleanup is made reliable.

Relevant unresolved candidate independently validated:

- [`PRRT_kwDOBjT0Fc6RGUI2`](https://github.com/ariakit/ariakit/pull/6734#discussion_r3587345936), PID-only wrapper `SIGKILL`: reproduced, but classified as invalid as an in-scope P2 requirement. The base branch already orphaned `conc`/Next.js under a PID-only wrapper kill, `SIGKILL` cannot be trapped, and adding that guarantee would require reintroducing supervision or a platform-specific parent-death primitive. I replied on the thread with the evidence and scope disposition.

Validation performed:

- Reviewed the full base diff `bd7b6779dc742914af2ee6bd52cf953fe2d74f26...a52ffc0f724f493201e5e818ea68cbf44feb487a` and the incremental rewrite.
- Inspected the installed Astro 7.0.7 agent/background path.
- `pnpm test packages/ariakit-scripts/src/dev.test.ts --run`: 4/4 passed.
- `pnpm test --run`: 54 files, 587/587 passed.
- `pnpm tsc --pretty false` and `pnpm lint`: passed.
- Removed only the env override locally: the focused SIGINT regression failed while the app port remained occupied, proving red/green coverage.
- Reproduced PID-only wrapper death versus foreground-group death in an isolated TCP process tree.
- All current GitHub checks are green.

I did not execute the POSIX process-group fixture on native Windows; that test is intentionally skipped there.

packages/ariakit-scripts/src/dev.test.ts
diegohaz: Fixed in [`528bd115b`](https://github.com/ariakit/ariakit/commit/528bd115b1be3b3f25d66a739310f7839e560bb1): the fixture now reads and validates atomically published ports before waiting for both server readiness markers. Early failures remove the harness-owned temp root so an escaped server closes itself without signaling recorded PIDs. The withheld-Next-readiness reproduction now leaves no fixture process or listening port behind.

diegohaz:
**Fixed**

- [`528bd115b`](https://github.com/ariakit/ariakit/commit/528bd115b1be3b3f25d66a739310f7839e560bb1) ([thread](https://github.com/ariakit/ariakit/pull/6734#discussion_r3587932611)): Preserved fixture cleanup when a server readiness marker never appears, using atomically published ports and harness-owned sentinel shutdown.

diegohaz (commented):
Re-review of `a52ffc0f724f493201e5e818ea68cbf44feb487a` → `528bd115b1be3b3f25d66a739310f7839e560bb1`, using the latest prior marked review as the incremental baseline. I reviewed the new cleanup commit plus the full PR and surrounding `dev()` process topology.

No new substantive findings.

The prior P2 in [`PRRT_kwDOBjT0Fc6RHDgu`](https://github.com/ariakit/ariakit/pull/6734#discussion_r3587613847) is resolved. The fixture now atomically publishes and validates the actual ports before waiting for both server readiness markers, then uses removal of the harness-owned temporary root as a cooperative shutdown sentinel for any server that escaped the process group:

```js
if (existsSync(process.env.ROOT_PATH)) return;
server.close(() => process.exit(0));
```

This is safer and simpler than retaining child PIDs and signaling them later: it avoids PID-reuse risk and removes the PID-file/index cleanup. It also does not mask the production regression, because the primary port-release assertion runs before the `finally` block removes the root.

I considered removing the atomic port-file handshake as a further simplification, but that would be unsafe. The parent supplies starting ports, while `dev()` may advance to different ports when a seed is occupied; the test must learn the actual values selected by `findAvailablePort()` before asserting release.

Validation performed:

- `pnpm test packages/ariakit-scripts/src/dev.test.ts --run`: 4/4 passed.
- `pnpm test --run`: 54 files, 587/587 passed; the command exited 0 after the known post-run `ECONNREFUSED` diagnostic.
- `pnpm tsc --pretty false`, `pnpm lint`, and `git diff --check`: passed.
- Removed only `ASTRO_DEV_BACKGROUND: "0"`: the SIGINT regression remained red on the occupied app port, then cleanup released both ports and left no fixture process.
- Repeated the previous early-failure reproduction by also withholding the Next.js readiness marker: it failed at the readiness wait, removed the temporary root, and left no process or port behind.
- All current GitHub checks pass, the worktree is clean, and no fixture process remains.

The authoritative review-thread query returned no unresolved threads. No changeset is needed because `@ariakit/scripts` is private. I did not execute the POSIX process-group fixture on native Windows; it remains intentionally skipped there.

PR #6735, opened 2026-07-15
Fix initially open sibling dialogs

### LINKED ISSUE #5238: Dialogs are freezing
### Current behavior

When 2 dialogs are open next to each other (and not nested), both are marked with `inert` making them both not interactive.

1. Open [StackBlitz](https://stackblitz.com/edit/ariakit-freezing-dialog-repro?file=src%2Fapp.tsx) reproduction
2. Open the "oranges" dialog
3. Open the "apples" dialog
4. Notice that the "apples" dialog is not interactive

### Expected behavior

The dialog should be interactive.

### Workaround

One workaround is to use `getPersistentElements` and pass the last dialog from the DOM. I have commented out the workaround in the reproduction above.

### Possible solutions

Fix the detection of the frontmost dialog.

diegohaz:
## Motivation

When two sibling modal [`Dialog`](https://ariakit.com/reference/dialog) components mount open in the same render, each dialog snapshots and disables the other dialog's portal. Both dialogs become inert, so users cannot interact with either one.

```tsx
<Ariakit.Dialog open unmountOnHide>
  Oranges
</Ariakit.Dialog>
<Ariakit.Dialog open unmountOnHide>
  Apples
</Ariakit.Dialog>
```

Fixes #5238.

## Solution

Track open modal dialogs in activation order for each document. Earlier dialogs treat later dialogs as part of their modal context, while the frontmost dialog still disables every dialog behind it. Closing a dialog removes it from the stack and restores the next dialog.

The stack complements the existing React nesting and `getPersistentElements` paths without replacing each dialog's temporal tree snapshot. Inline nested dialogs are kept hierarchy-aware by excluding DOM ancestors that register after their children.

Add happy-dom and browser regression coverage for initially open siblings, plus initially open inline nested dialogs to protect the child-first layout-effect case.

## Workaround

Before this fix is available in a release, give the known foreground sibling a stable ID and return its element from the background dialog's `getPersistentElements` callback:

```tsx
const applesDialogId = useId();

<Ariakit.Dialog
  getPersistentElements={() => {
    const applesDialog = document.getElementById(applesDialogId);
    return applesDialog ? [applesDialog] : [];
  }}
>
  Oranges
</Ariakit.Dialog>

<Ariakit.Dialog id={applesDialogId}>Apples</Ariakit.Dialog>
```

This workaround assumes the application knows which sibling is in front. A dynamic dialog stack must return every later/frontmost dialog that should remain interactive.

### DIFF
--- a/app/src/sandbox/dialog-5238/test-browser.ts
+++ b/app/src/sandbox/dialog-5238/test-browser.ts
@@ -1,4 +1,9 @@
+import type { Locator } from "@playwright/test";
 import { withFramework } from "#app/test-utils/preview.ts";
+
+function isInert(locator: Locator) {
+  return locator.evaluate((element) => element.closest("[inert]") !== null);
+}
 
 withFramework(import.meta.dirname, async ({ test }) => {
   // Reproduces https://github.com/ariakit/ariakit/issues/5238
@@ -7,6 +12,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     q,
   }) => {
     await test.expect(q.dialog("Apples")).toBeVisible();
+    await test.expect.poll(() => isInert(q.dialog("Oranges"))).toBe(true);
     await q.button("Eat apple").click();
     await test.expect(q.status("Apple count")).toHaveText("Apples eaten: 1");
 
@@ -17,7 +23,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.dialog("Apples")).not.toBeVisible();
     await test.expect(page.locator("[data-dialog]")).toHaveCount(1);
     await test.expect(q.dialog("Oranges")).toBeVisible();
-    await q.button("Eat orange").click();
+    await q.button("Eat orange").press("Enter");
     await test.expect(q.status("Orange count")).toHaveText("Oranges eaten: 1");
   });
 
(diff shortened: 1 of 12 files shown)


packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -316,6 +321,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     const allElements = [
       dialog,
       ...persistentElements,
+      ...stackedDialogs,
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`010010889`](https://github.com/ariakit/ariakit/commit/01001088903234989beb08ee2f37a6dc5e6332bd): stacked entries now register the foreground backdrop alongside its dialog, and replacement backdrops follow the disclosure store content element. React and Chromium regressions verify that right-clicking either the default or replaced foreground backdrop dismisses only that dialog and leaves the background dialog interactive.

diegohaz:
**Fixed**

- [`010010889`](https://github.com/ariakit/ariakit/commit/01001088903234989beb08ee2f37a6dc5e6332bd) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3585995304)): registered foreground backdrops in the modal stack and added default/replacement backdrop interaction regressions.

app/src/sandbox/dialog-5238/test.ts
```diff
@@ -0,0 +1,52 @@
...
+  expect(q.status("Orange count")).toHaveTextContent("Oranges eaten: 1");
+});
+
+// Reproduces the backdrop replacement case found while fixing #5238.
+test("keeps the background open after replacing the foreground backdrop", async () => {
+  const previousBackdrop = getBackdrop("Apples");
+  previousBackdrop?.setAttribute("data-previous-backdrop", "");
+
+  await click(q.button("Replace apple backdrop"));
+  expect(document.querySelector("[data-previous-backdrop]")).toBeNull();
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`5184c686a`](https://github.com/ariakit/ariakit/commit/5184c686ac2c627b9b05242406ef3f6d22b368de): the test now requires the original foreground backdrop to exist before marking it, so the replacement assertion cannot pass vacuously.

diegohaz:
**Fixed**

- [`5184c686a`](https://github.com/ariakit/ariakit/commit/5184c686ac2c627b9b05242406ef3f6d22b368de) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3586103126)): Required the original foreground backdrop to exist before marking it, preventing the replacement regression test from passing vacuously.

packages/ariakit-react-components/src/dialog/utils/use-dialog-stack.ts
```diff
@@ -0,0 +1,125 @@
...
+        entry.dialog = dialog;
+        updateDialogStack(stack);
+        return;
+      }
+      removeEntry();
+    }
+    const stack = getDialogStack(dialog);
+    const nextEntry = { dialog, backdrop: backdropRef.current };
+    entryRef.current = nextEntry;
+    stack.dialogs.push(nextEntry);
```
diegohaz: [P1] Keep reopened mounted dialogs above their inert siblings

Appending every newly enabled entry assumes activation order also controls paint order. With the default `unmountOnHide={false}`, the portal for a previously hidden earlier sibling stays before the later sibling in the DOM:

```tsx
<Dialog open={firstOpen}>First</Dialog>
<Dialog open>Second</Dialog>
```

When `firstOpen` becomes true, this line promotes First in the registry and makes Second inert, but Second's unchanged portal/backdrop still paints later and covers First at the same z-index. On the exact head, First was non-inert, Second was inert, and First's portal still preceded Second's, leaving the only interactive dialog visually blocked. Please keep the logical frontmost dialog aligned with the mounted portal's paint order (or explicitly promote its portal/backdrop) and cover the default mounted lifecycle in a browser test.

packages/ariakit-react-components/src/dialog/utils/use-dialog-stack.ts
```diff
@@ -0,0 +1,125 @@
...
+    entry.backdrop = backdrop;
+    const stack = dialogStacks.get(getDocument(entry.dialog));
+    if (!stack) return;
+    updateDialogStack(stack);
+  }, []);
+
+  // Keep the entry for the lifetime of this opening so replacing the dialog
+  // element doesn't change its position in the stack.
+  useSafeLayoutEffect(() => {
+    if (!enabled) return;
```
diegohaz: [P1] Do not turn transient portal setup into a new activation

The caller includes `domReady` in `enabled`, so changing a still-open background sibling from `portal={false}` to `portal={true}` briefly runs this cleanup and removes its entry. Once the new portal is ready, the entry is appended after the still-open foreground dialog, promoting the background and making the real foreground inert even though neither `open` state changed:

```tsx
<Dialog open portal={backgroundPortal}>Background</Dialog>
<Dialog open>
  <button onClick={() => setBackgroundPortal(true)}>Move background</button>
  Foreground
</Dialog>
```

The exact-head repro reverses accessibility after the click: Background becomes visible to role queries and Foreground disappears. The entry lifetime needs to survive transient DOM readiness while preserving canonical nesting order; merely switching `enabled` to `modal && open` is insufficient because initially inline children register before their DOM ancestors.

app/src/sandbox/dialog-5238/test-browser.ts
```diff
@@ -0,0 +1,59 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // Reproduces https://github.com/ariakit/ariakit/issues/5238
+  test("keeps the frontmost initially open sibling dialog interactive", async ({
+    page,
+    q,
+  }) => {
+    await test.expect(q.dialog("Apples")).toBeVisible();
```
diegohaz, line 14: [P2] Assert that the background sibling is inaccessible

This proves Apples is accessible, but neither this browser test nor the happy-dom duplicate proves Oranges is excluded from the accessible tree while Apples is frontmost. The happy-dom call to `q.dialog.hidden("Oranges")` only proves that the physical node remains mounted and visibly styled, so an implementation that leaves both modal siblings interactive would still pass the current regression. Please assert the core background invariant in both copies, for example:

```ts
await test.expect(q.dialog("Oranges")).not.toBeVisible();
// happy-dom:
expect(q.dialog("Oranges")).not.toBeInTheDocument();
```

app/src/sandbox/dialog-5238/test-browser.ts
```diff
@@ -0,0 +1,59 @@
...
+    await test.expect(q.status("Apple count")).toHaveText("Apples eaten: 1");
+
+    const applesId = await q.dialog("Apples").getAttribute("id");
+    await page
+      .locator(`[data-backdrop="${applesId}"]`)
+      .click({ button: "right", position: { x: 4, y: 4 } });
+    await test.expect(q.dialog("Apples")).not.toBeVisible();
+    await test.expect(page.locator("[data-dialog]")).toHaveCount(1);
+    await test.expect(q.dialog("Oranges")).toBeVisible();
+    await q.button("Eat orange").click();
```
diegohaz: [P2] Clear the native context menu before the next activation

The right-click above can leave WebKit's native context menu open, and on the supported local Safari project this pointer click is consumed only to dismiss that menu, so no DOM click fires and the counter stays at zero. I reproduced this 3/3 on the exact head with `--project=safari`; the exact GitHub Safari job passed, so the defect is environment-sensitive rather than universal. Dismiss the menu first or use a portable keyboard activation here, for example:

```ts
await q.button("Eat orange").press("Enter");
```

diegohaz (commented):
Initial full review of `bd7b6779dc742914af2ee6bd52cf953fe2d74f26...5184c686ac2c627b9b05242406ef3f6d22b368de` (no prior Ariakit review marker found).

I reviewed all 12 changed files and the surrounding dialog, backdrop, portal/disclosure, nested-dialog, tree-marking, outside-interaction, focus, package-export, changeset, and performance-sensitive paths. I left four inline findings:

- [P1] Reopening an earlier sibling with the default mounted lifecycle makes it logically frontmost while its unchanged portal can remain painted below the now-inert later sibling.
- [P1] A transient `domReady` gap while changing `portal` removes and re-appends a still-open background dialog, incorrectly promoting it over the foreground.
- [P2] The main regression never asserts that the background sibling is inaccessible while the frontmost sibling is open.
- [P2] The Safari browser test assumes the first pointer click after a native context menu reaches the page; supported local WebKit consumes it to dismiss the menu.

Validation performed:

- `pnpm test-react packages/ariakit-react-components/src/dialog/utils/use-dialog-stack.test.tsx app/src/sandbox/dialog-5238/test.ts app/src/sandbox/dialog-5238-nested/test.ts examples/dialog-nested-various/test.ts` — 21/21 passed.
- The same focused set through `pnpm test-react18` — 21/21 passed.
- Exact-head out-of-tree lifecycle repros confirmed both stack-order defects; the expected sibling-order assertion fails after a background dialog changes portal mode.
- The original test-only commit `c0256a042` fails on the unfixed implementation at the intended Apples accessibility assertion.
- The exact-head Safari test failed 3/3 locally at `Oranges eaten: 0` after the context-menu interaction. The exact GitHub Safari job and all other reported CI checks are green, so this test issue is environment-sensitive.
- `git diff --check` passed and the worktree is clean.

Authoritative, fully paginated GraphQL thread state had no unresolved review threads. The two existing threads were resolved and were ignored as candidates.

I did not run the complete repository test suite or the full local browser matrix; broad build, type, browser, package, and performance coverage was taken from the green exact-head CI checks.

packages/ariakit-react-components/src/dialog/utils/use-dialog-stack.ts
diegohaz: Fixed in [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5): reopening a mounted dialog now gets a fresh activation and Ariakit-owned default portal containers are synchronized to the resulting stack order, so the promoted dialog also paints frontmost. Browser coverage verifies the reopened dialog is accessible, non-inert, and interactive while its sibling is inert.

packages/ariakit-react-components/src/dialog/utils/use-dialog-stack.ts
diegohaz: Fixed in [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5): stack activation now spans the full open lifetime, disconnected portal placeholders are ignored, and portal readiness changes update the existing entry instead of promoting it. React and browser regressions verify an inline background dialog remains inaccessible and inert after moving into a portal.

app/src/sandbox/dialog-5238/test-browser.ts
diegohaz: Fixed in [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5): the happy-dom and browser regressions now assert the background sibling is absent from accessible role queries and inert, while also proving the frontmost dialog remains interactive across reopen, portal, replacement, fullscreen, and nested-branch cases.

app/src/sandbox/dialog-5238/test-browser.ts
diegohaz: Fixed in [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5): the post-context-menu interaction now uses keyboard Enter, avoiding WebKit consuming the next pointer click. The dialog-5238 browser suite passes in Chrome, Firefox, and Safari.

diegohaz:
**Fixed**

- [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3587504284)): synchronized reopened default portals with activation order.
- [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3587504430)): preserved stack position across transient portal setup.
- [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3587504601)): added explicit background accessibility and inertness coverage.
- [`10306848e`](https://github.com/ariakit/ariakit/commit/10306848e0a2c15f1792941c5561f2c8318451b5) ([reply](https://github.com/ariakit/ariakit/pull/6735#discussion_r3587504763)): made the context-menu regression deterministic in Safari with Enter.

diegohaz:

PR #4202, opened 2024-10-11
Fix `collectionStore.item()` getting out of sync with current items

diegohaz:
Currently, calling `store.item(id)` may return an item that has been previously removed from the collection/composite store. This occurs because we were searching for the item in the public store, which may lag behind the private store where items are registered synchronously one by one (this prevents multiple unnecessary updates when rendering several items).

This PR modifies the `item` method logic to use the private store instead of the batched public store.

I also updated the tabbable logic introduced in #4191 to consider only `item?.element` instead of `item?.element?.isConnected`. This is because when `isConnected` changes, it doesn't always trigger the `useStoreState` callback, as the component may not re-render or the state may not update in response to this DOM change.

PR #6736, opened 2026-07-15
Fix Select typeahead with late items

### LINKED ISSUE #6733: Select typeahead does not update the value for late items when options are unmounted
### Current behavior

When a controlled `Select` receives its `items` after `useSelectStore` has already been created and the options are unmounted while the popover is closed, typing on the closed Select trigger can move `activeId` without updating `value`.

In the reproduction, the store starts empty with `Orange` selected. After loading Apple, Banana, and Orange and typing `a` on the closed trigger, `activeId` becomes `apple`, but the Select value remains `Orange`.

The same setup works when the items are available on the store's first render or when the `SelectItem` elements stay mounted with `unmountOnHide={false}`. The failure also occurs with `SelectRenderer`, which renders no item elements while the Select is closed.

The typeahead code [falls back to the public `items` state when no items are rendered](https://github.com/ariakit/ariakit/blob/bd7b6779dc742914af2ee6bd52cf953fe2d74f26/packages/ariakit-react-components/src/composite/composite-typeahead.tsx#L164-L216), finds Apple, and moves `activeId` to it. The Select store then [looks up that id through `composite.item(activeId)` before updating the value](https://github.com/ariakit/ariakit/blob/bd7b6779dc742914af2ee6bd52cf953fe2d74f26/packages/ariakit-components/src/select/select-store.ts#L167-L176). Because the late item exists only in public state and has not registered through a mounted `SelectItem`, that lookup returns `null`, so the value update is skipped.

This is a concrete downstream effect of [#6295](https://github.com/ariakit/ariakit/issues/6295), but it can potentially be fixed in Select without changing the broader Collection lookup semantics.

1. Open the [StackBlitz reproduction](https://stackblitz.com/edit/react-uu7c2d9b?file=index.js).
2. Click **Load options**. This adds the controlled items after store creation and focuses the closed Select trigger.
3. Type <kbd>a</kbd>.
4. Observe that the active item changes to Apple, but the displayed Select value remains Orange.

### Expected behavior

Typing <kbd>a</kbd> on the closed Select should change the value to Apple. `activeId` and `value` should remain consistent when typeahead finds an enabled item in the public `items` state, even if that item's element is currently unmounted.

### Workaround

Keep the `SelectItem` elements mounted while the popover is hidden:

```tsx
<Ariakit.SelectPopover store={select} unmountOnHide={false}>
  {items.map((item) => (
    <Ariakit.SelectItem key={item.id} {...item} />
  ))}
</Ariakit.SelectPopover>
```

Providing the items on the store's first render also avoids the issue when they are already known synchronously.

[Confirmed workaround and triage evidence](https://github.com/ariakit/ariakit/issues/6733#issuecomment-4977427918)

[Async items and SelectRenderer workaround](https://github.com/ariakit/ariakit/issues/6733#issuecomment-4977816687)

### Possible solutions

At the Select-specific point that converts a moved active item into a value, fall back to the public Select items only when the registered-item lookup misses:

```ts
const item =
  composite.item(activeId) ??
  select.getState().items.find((item) => item.id === activeId);
```

This null-only fallback limits the behavioral change to Select and leaves the broader `CollectionStore.item()` registration semantics unchanged.

Regression coverage should include late controlled items with an unmounted `SelectPopover`, the equivalent `SelectRenderer` case, disabled and valueless items, and removals or composed-store batching. The last cases matter because [#4202](https://github.com/ariakit/ariakit/pull/4202) intentionally stopped `CollectionStore.item()` from falling back directly to public state when that state could lag a synchronous removal.

diegohaz:
## Motivation

When a controlled `Select` receives items after store creation while its options are unmounted, closed-trigger typeahead can move the active item without updating the selected value. The same mismatch occurs when `SelectRenderer` renders zero items while the Select is closed.

```tsx
const select = useSelectStore({ items, value, setValue });

<Select store={select}>{value}</Select>
<SelectPopover store={select} unmountOnHide>
  {items.map((item) => (
    <SelectItem key={item.id} {...item} />
  ))}
</SelectPopover>
```

After loading Apple and typing <kbd>a</kbd>, `activeId` becomes `apple`, but the trigger still shows Orange.

## Solution

Keep registered collection items as the authoritative fast path. When typeahead moves to an eligible item that exists only in the public `items` state, defer that fallback by one microtask so any pending item unregistration can first reach public state. Before updating the value, revalidate the move, active item, current value, and store lifetime, then resolve registered metadata again before using the public item.

```ts
const item = composite.item(activeId);
if (item) {
  setValueFromItem(item);
  return;
}

const publicItem = items.find((item) => item.id === activeId);
if (!publicItem) return;

queueMicrotask(() => {
  const currentItem =
    composite.item(activeId) ??
    select.getState().items.find((item) => item.id === activeId);
  setValueFromItem(currentItem);
});
```

The regression coverage includes the reported unmounted-popover case and `SelectRenderer` in happy-dom and Chrome, Firefox, and Safari. Store-level tests cover disabled and valueless items, composed stores, same-task unregistration, newly registered metadata, post-move additions and value writes, and teardown cancellation.

## Workaround

When ordinary `SelectItem` elements are mapped directly, keeping them mounted with `unmountOnHide={false}` registers late items before closed-trigger typeahead needs them. That does not cover `SelectRenderer`, which deliberately renders zero items while the Select is closed.

For a controlled single-value Select whose items may arrive asynchronously, mirror only closed-state moves from the application-owned items into the value:

```tsx
const activeId = useStoreState(select, "activeId");
const mounted = useStoreState(select, "mounted");
const moves = useStoreState(select, "moves");
const previousMoves = useRef(moves);

// TODO: Remove after https://github.com/ariakit/ariakit/issues/6733 is fixed.
useEffect(() => {
  if (previousMoves.current === moves) return;
  previousMoves.current = moves;
  if (mounted) return;
  if (!activeId) return;
  const item = items.find((item) => item.id === activeId);
  if (!item || item.disabled || item.value == null) return;
  setValue(item.value);
}, [activeId, items, mounted, moves]);
```

This temporary workaround assumes a controlled single-value Select and treats the application-owned `items` array as authoritative. It only synchronizes moves made while the Select is closed, so normal open-popover navigation remains unchanged. It is not suitable for multi-select values or applications that intentionally keep `activeId` and `value` independent while closed.

Fixes #6733

### DIFF
--- a/packages/ariakit-components/src/select/select-store.ts
+++ b/packages/ariakit-components/src/select/select-store.ts
@@ -170,6 +170,9 @@ export function createSelectStore({
     select.setState("value", item.value);
   };
 
+  // Tracks strict-equal setter calls that state subscriptions don't observe.
+  let setValueVersion = 0;
+
   // Sets the select value when the active item changes by moving (which usually
   // happens when moving to an item using the keyboard).
   setup(select, () =>
@@ -190,6 +193,7 @@ export function createSelectStore({
       if (publicItem.disabled) return;
       if (publicItem.value == null) return;
 
+      const setValueVersionAtBatch = setValueVersion;
       let canceled = false;
       const stopValueSubscription = subscribe(select, ["value"], () => {
         canceled = true;
@@ -197,6 +201,7 @@ export function createSelectStore({
       queueMicrotask(() => {
         stopValueSubscription();
         if (canceled) return;
+        if (setValueVersion !== setValueVersionAtBatch) return;
         const currentState = select.getState();
         if (currentState.moves !== moves) return;
         if (currentState.activeId !== activeId) return;
@@ -219,7 +224,10 @@ export function createSelectStore({
     ...popover,
     ...select,
     combobox,
-    setValue: (value) => select.setState("value", value),
+    setValue: (value) => {
+      setValueVersion += 1;
+      select.setState("value", value);
+    },
     setLabelElement: (element) => select.setState("labelElement", element),
     setSelectElement: (element) => select.setState("selectElement", element),
     setListElement: (element) => select.setState("listElement", element),
--- a/packages/ariakit-components/src/select/select-store.ts
+++ b/packages/ariakit-components/src/select/select-store.ts
@@ -162,18 +162,49 @@ export function createSelectStore({
     }),
   );
 
+  const setValueFromItem = (item: SelectStoreItem | null | undefined) => {
+    if (!item) return;
+    if (item.disabled) return;
+    if (item.value == null) return;
+    select.setState("value", item.value);
+  };
+
   // Sets the select value when the active item changes by moving (which usually
   // happens when moving to an item using the keyboard).
   setup(select, () =>
-    batch(select, ["setValueOnMove", "moves"], (state) => {
-      const { mounted, value, activeId } = select.getState();
-      if (!state.setValueOnMove && mounted) return;
+    batch(select, ["setValueOnMove", "moves"], () => {
+      const { activeId, items, mounted, moves, setValueOnMove, value } =
+        select.getState();
+      if (!setValueOnMove && mounted) return;
       if (Array.isArray(value)) return;
-      if (!state.moves) return;
+      if (!moves) return;
       if (!activeId) return;
       const item = composite.item(activeId);
-      if (!item || item.disabled || item.value == null) return;
-      select.setState("value", item.value);
+      if (item) {
+        setValueFromItem(item);
+        return;
+      }
+      const publicItem = items.find((item) => item.id === activeId);
+      if (!publicItem) return;
+      if (publicItem.disabled) return;
+      if (publicItem.value == null) return;
+
+      let canceled = false;
+      queueMicrotask(() => {
+        if (canceled) return;
+        const currentState = select.getState();
+        if (currentState.moves !== moves) return;
+        if (currentState.activeId !== activeId) return;
+        if (currentState.value !== value) return;
+        const currentItem =
+          composite.item(activeId) ??
+          currentState.items.find((item) => item.id === activeId);
+        setValueFromItem(currentItem);
+      });
+
+      return () => {
+        canceled = true;
+      };
     }),
   );
 
--- a/packages/ariakit-components/src/select/select-store.ts
+++ b/packages/ariakit-components/src/select/select-store.ts
@@ -4,6 +4,7 @@ import {
   mergeStore,
   omit,
   setup,
+  subscribe,
   sync,
   throwOnConflictingProps,
 } from "@ariakit/store";
@@ -190,7 +191,11 @@ export function createSelectStore({
       if (publicItem.value == null) return;
 
       let canceled = false;
+      const stopValueSubscription = subscribe(select, ["value"], () => {
+        canceled = true;
+      });
       queueMicrotask(() => {
+        stopValueSubscription();
         if (canceled) return;
         const currentState = select.getState();
         if (currentState.moves !== moves) return;
@@ -204,6 +209,7 @@ export function createSelectStore({
 
       return () => {
         canceled = true;
+        stopValueSubscription();
       };
     }),
   );
--- a/packages/ariakit-components/src/select/select-store.ts
+++ b/packages/ariakit-components/src/select/select-store.ts
@@ -9,7 +9,7 @@ import {
   throwOnConflictingProps,
 } from "@ariakit/store";
 import type { Store, StoreOptions, StoreProps } from "@ariakit/store";
-import { toArray, defaultValue } from "@ariakit/utils";
+import { toArray, defaultValue, hasOwnProperty } from "@ariakit/utils";
 import type { PickRequired, SetState } from "@ariakit/utils";
 import type { ComboboxStore } from "../combobox/combobox-store.ts";
 import type {
@@ -28,6 +28,59 @@ import { createPopoverStore } from "../popover/popover-store.ts";
 
 type MutableValue<T extends SelectStoreValue = SelectStoreValue> =
   T extends string ? string : T;
+
+interface ValueWriteTracker {
+  version: number;
+}
+
+interface StoreWithInternals extends Store {
+  __unstableInternals?: {
+    stores?: ReadonlyArray<Store | undefined>;
+  };
+}
+
+const valueWriteTrackerMap = new WeakMap<object, ValueWriteTracker>();
+
+function getValueWriteTrackers(store: Store) {
+  const roots = new Set<object>();
+  const visited = new Set<object>();
+
+  const visit = (store: StoreWithInternals) => {
+    // Store objects may be spread while preserving the same internals.
+    const identity = store.__unstableInternals ?? store;
+    if (visited.has(identity)) return;
+    visited.add(identity);
+
+    let hasValueParent = false;
+    for (const parent of store.__unstableInternals?.stores ?? []) {
+      if (!parent) continue;
+      if (!hasOwnProperty(parent.getState(), "value")) continue;
+      hasValueParent = true;
+      visit(parent);
+    }
+    if (!hasValueParent) {
+      roots.add(identity);
+    }
+  };
+
+  visit(store);
+
+  const trackers: ValueWriteTracker[] = [];
+  for (const root of roots) {
+    const tracker = valueWriteTrackerMap.get(root) ?? { version: 0 };
+    valueWriteTrackerMap.set(root, tracker);
+    trackers.push(tracker);
+  }
+  return trackers;
+}
+
+function getValueWriteVersion(trackers: ValueWriteTracker[]) {
+  let version = 0;
+  for (const tracker of trackers) {
+    version += tracker.version;
+  }
+  return version;
+}
 
 export function createSelectStore<
   T extends SelectStoreValue = SelectStoreValue,
@@ -119,6 +172,8 @@ export function createSelectStore({
 
   const select = createStore(initialState, composite, popover, store);
 
+  const valueWriteTrackers = getValueWriteTrackers(select);
+
   // Automatically sets the default value if it's not set.
   setup(select, () =>
     sync(select, ["value", "items"], (state) => {
@@ -170,9 +225,6 @@ export function createSelectStore({
     select.setState("value", item.value);
   };
 
-  // Tracks strict-equal setter calls that state subscriptions don't observe.
-  let setValueVersion = 0;
-
   // Sets the select value when the active item changes by moving (which usually
   // happens when moving to an item using the keyboard).
   setup(select, () =>
@@ -193,7 +245,7 @@ export function createSelectStore({
       if (publicItem.disabled) return;
       if (publicItem.value == null) return;
 
-      const setValueVersionAtBatch = setValueVersion;
+      const valueWriteVersion = getValueWriteVersion(valueWriteTrackers);
       let canceled = false;
       const stopValueSubscription = subscribe(select, ["value"], () => {
         canceled = true;
@@ -201,7 +253,9 @@ export function createSelectStore({
       queueMicrotask(() => {
         stopValueSubscription();
         if (canceled) return;
-        if (setValueVersion !== setValueVersionAtBatch) return;
+        if (getValueWriteVersion(valueWriteTrackers) !== valueWriteVersion) {
+          return;
+        }
         const currentState = select.getState();
         if (currentState.moves !== moves) return;
         if (currentState.activeId !== activeId) return;
@@ -219,15 +273,22 @@ export function createSelectStore({
     }),
   );
 
+  const setState: SelectStore["setState"] = (key, value) => {
+    if (key === "value") {
+      for (const tracker of valueWriteTrackers) {
+        tracker.version += 1;
+      }
+    }
+    select.setState(key, value);
+  };
+
   return {
     ...composite,
     ...popover,
     ...select,
     combobox,
-    setValue: (value) => {
-      setValueVersion += 1;
-      select.setState("value", value);
-    },
+    setState,
+    setValue: (value) => setState("value", value),
     setLabelElement: (element) => select.setState("labelElement", element),
     setSelectElement: (element) => select.setState("selectElement", element),
     setListElement: (element) => select.setState("listElement", element),
--- a/packages/ariakit-components/src/select/select-store.ts
+++ b/packages/ariakit-components/src/select/select-store.ts
@@ -31,18 +31,22 @@ type MutableValue<T extends SelectStoreValue = SelectStoreValue> =
 
 interface ValueWriteTracker {
   version: number;
+  subscriptions: number;
+  subscribeSameValue?: (listener: (key: PropertyKey) => void) => () => void;
+  unsubscribeSameValue?: () => void;
 }
 
 interface StoreWithInternals extends Store {
   __unstableInternals?: {
     stores?: ReadonlyArray<Store | undefined>;
+    subscribeSameValue?: (listener: (key: PropertyKey) => void) => () => void;
   };
 }
 
 const valueWriteTrackerMap = new WeakMap<object, ValueWriteTracker>();
 
 function getValueWriteTrackers(store: Store) {
-  const roots = new Set<object>();
+  const roots = new Map<object, StoreWithInternals>();
   const visited = new Set<object>();
 
   const visit = (store: StoreWithInternals) => {
@@ -59,19 +63,44 @@ function getValueWriteTrackers(store: Store) {
       visit(parent);
     }
     if (!hasValueParent) {
-      roots.add(identity);
+      roots.set(identity, store);
     }
   };
 
   visit(store);
 
   const trackers: ValueWriteTracker[] = [];
-  for (const root of roots) {
-    const tracker = valueWriteTrackerMap.get(root) ?? { version: 0 };
-    valueWriteTrackerMap.set(root, tracker);
+  for (const [identity, root] of roots) {
+    let tracker = valueWriteTrackerMap.get(identity);
+    if (!tracker) {
+      const nextTracker: ValueWriteTracker = {
+        version: 0,
+        subscriptions: 0,
+        subscribeSameValue: root.__unstableInternals?.subscribeSameValue,
+      };
+      tracker = nextTracker;
+      valueWriteTrackerMap.set(identity, nextTracker);
+    }
     trackers.push(tracker);
   }
   return trackers;
+}
+
+function retainValueWriteTracker(tracker: ValueWriteTracker) {
+  tracker.subscriptions += 1;
+  if (tracker.subscriptions === 1) {
+    tracker.unsubscribeSameValue = tracker.subscribeSameValue?.((key) => {
+      if (key === "value") {
+        tracker.version += 1;
+      }
+    });
+  }
+  return () => {
+    tracker.subscriptions -= 1;
+    if (tracker.subscriptions) return;
+    tracker.unsubscribeSameValue?.();
+    tracker.unsubscribeSameValue = undefined;
+  };
 }
 
 function getValueWriteVersion(trackers: ValueWriteTracker[]) {
@@ -174,6 +203,15 @@ export function createSelectStore({
 
   const valueWriteTrackers = getValueWriteTrackers(select);
 
+  setup(select, () => {
+    const releases = valueWriteTrackers.map(retainValueWriteTracker);
+    return () => {
+      for (const release of releases) {
+        release();
+      }
+    };
+  });
+
   // Automatically sets the default value if it's not set.
   setup(select, () =>
     sync(select, ["value", "items"], (state) => {
@@ -222,6 +260,7 @@ export function createSelectStore({
     if (!item) return;
     if (item.disabled) return;
     if (item.value == null) return;
+    if (select.getState().value === item.value) return;
     select.setState("value", item.value);
   };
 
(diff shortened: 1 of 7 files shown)


packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +162,49 @@ export function createSelectStore({
...
+      if (publicItem.disabled) return;
+      if (publicItem.value == null) return;
+
+      let canceled = false;
+      queueMicrotask(() => {
+        if (canceled) return;
+        const currentState = select.getState();
+        if (currentState.moves !== moves) return;
+        if (currentState.activeId !== activeId) return;
+        if (currentState.value !== value) return;
```
diegohaz, line 301: [P2] Cancel the fallback after any later value write

This equality check only compares the final value with the captured value, so it misses an ABA sequence. On this head:

```ts
store.setState("items", items);
store.move("apple");
queueMicrotask(() => {
  store.setValue("Banana");
  store.setValue("Orange");
});
```

With an open `setValueOnMove` Select, the late-item path ends at `Apple`, while registering the same Apple through the existing fast path ends at `Orange`. The earlier move is therefore replayed after two later application writes solely because the item was unmounted.

Please invalidate the pending fallback on any intervening value change rather than comparing only the final value. I prototyped a temporary synchronous `value` subscription scoped to the pending microtask; it kept all seven focused tests green and made this case end at `Orange`. Add this ABA case to `select-store.test.ts` as well.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +162,49 @@ export function createSelectStore({
...
+      if (!publicItem) return;
+      if (publicItem.disabled) return;
+      if (publicItem.value == null) return;
+
+      let canceled = false;
+      queueMicrotask(() => {
+        if (canceled) return;
+        const currentState = select.getState();
+        if (currentState.moves !== moves) return;
+        if (currentState.activeId !== activeId) return;
```
diegohaz, line 300: [P2] Cover stale active-id cancellation

This guard is required, but none of the new tests changes `activeId` directly between the batch callback and the queued fallback. I temporarily removed this line and all seven focused store tests still passed, while this public-store sequence regressed:

```ts
store.setState("items", [apple, banana]);
store.move("apple");
queueMicrotask(() => store.setActiveId("banana"));
await flushBatch();
```

Without the guard, the stale callback selects Apple and the value-to-active synchronization resets `activeId` back to `apple`; with the guard, the state remains `{ activeId: "banana", value: "Orange" }`. Please add this focused case to `select-store.test.ts` so the cancellation boundary is protected.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +162,49 @@ export function createSelectStore({
...
+        if (currentState.value !== value) return;
+        const currentItem =
+          composite.item(activeId) ??
+          currentState.items.find((item) => item.id === activeId);
+        setValueFromItem(currentItem);
+      });
```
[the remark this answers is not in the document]
diegohaz: I do not think the proposed re-check is correct. I compared the two item-resolution paths using public store methods:

```ts
store.move("apple");
queueMicrotask(store.show);
```

With `setValueOnMove` false, both a registered Apple and a late public-only Apple end at `Apple` on this head. The batch decided the move while the Select was closed; registered metadata commits immediately, whereas only the public fallback waits. Re-checking `mounted` in the deferred branch would make the late path stay at `Orange` while the registered path stays at `Apple`.

The same ordering applies to a later `setValueOnMove` change. Later `moves`, `activeId`, or `value` writes can invalidate the captured move, but merely opening after that decision should not make behavior depend on registration state. A registered-vs-late parity test would make this intent explicit.

diegohaz (commented):
Initial review of `008b852737de4cf45a34b116d41685b39a61ef76` against merge-base `bd7b6779dc742914af2ee6bd52cf953fe2d74f26`.

Reviewed the Select fallback and surrounding Store batch/cleanup semantics, Collection registered-vs-public lookup behavior and #4202 history, React controlled-item syncing, the sandbox and store regressions, changeset/package reachability, performance results, and current CI.

Validation:
- `pnpm test packages/ariakit-components/src/select/select-store.test.ts --run`: 7/7 passed.
- `pnpm test-react app/src/sandbox/select-6733/test.ts --run`: 2/2 passed.
- The tests-only commit `1d5776c554134916dec688c81ebb614277273483` proved the red phase: React/React 18 and Chrome/Firefox/Safari failed because Apple was expected while Orange remained selected. The same jobs are green on this head.
- Focused source-mode probes compared public-only and registered items across microtask orderings.
- A temporary synchronous value-change subscription preserved all seven focused tests and fixed the ABA case; the exact snapshot was restored afterward.
- Temporarily removing the `activeId` guard left all seven tests green while the focused stale-active probe failed; the exact snapshot was restored afterward.
- `git diff --check` passed, the Select reference link returned 200, the worktree is clean, and the PR performance report found no significant Chrome or Node regression.

Relevant unresolved candidate:
- [discussion_r3585248357](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585248357): **invalid/inaccurate**. Re-checking `mounted` or `setValueOnMove` after the fallback delay would make public-only items diverge from the registered fast path. I replied with the registered-vs-late parity evidence.

New findings:
- [discussion_r3585364639](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585364639): an ABA value-write sequence can replay the earlier move after later application writes.
- [discussion_r3585364747](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585364747): the essential stale-`activeId` cancellation has no focused regression coverage.

No other substantive correctness, compatibility, simplification, changeset, or performance findings. I did not rerun Playwright or the full build/typecheck locally; the current head is green in Chrome, Firefox, Safari, build, and typecheck CI.

packages/ariakit-components/src/select/select-store.ts
diegohaz: Fixed in [`12ae1c818`](https://github.com/ariakit/ariakit/commit/12ae1c818e5c475bff89d9413a0bc2db13e6c508): added a focused public-store regression that changes `activeId` directly between the batch callback and fallback, and mutation-tested it by removing the guard.

packages/ariakit-components/src/select/select-store.ts
diegohaz: Fixed in [`75989cbd7`](https://github.com/ariakit/ariakit/commit/75989cbd7b4790a9b19bfa261d063b9dae066910): the pending public-item fallback now uses a temporary synchronous `value` subscription to cancel after any intervening value change, including ABA writes, with focused regression coverage. The subscription is disposed before the fallback and on cleanup.

diegohaz:
**Fixed**

- [`12ae1c818`](https://github.com/ariakit/ariakit/commit/12ae1c818e5c475bff89d9413a0bc2db13e6c508) ([reply](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585559048)): Added direct active-ID cancellation coverage and verified it fails without the guard.
- [`75989cbd7`](https://github.com/ariakit/ariakit/commit/75989cbd7b4790a9b19bfa261d063b9dae066910) ([reply](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585560089)): Canceled deferred public-item fallbacks after intervening value changes, including ABA sequences, with focused regression coverage.

**Invalid**

- ([reply](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585725279)): The reported same-task value ordering matches the established registered-item behavior on `main`; changing only the fallback would break registered/public parity.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +163,54 @@ export function createSelectStore({
...
+      const stopValueSubscription = subscribe(select, ["value"], () => {
+        canceled = true;
+      });
+      queueMicrotask(() => {
+        stopValueSubscription();
+        if (canceled) return;
+        const currentState = select.getState();
+        if (currentState.moves !== moves) return;
+        if (currentState.activeId !== activeId) return;
+        if (currentState.value !== value) return;
```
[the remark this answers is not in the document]
diegohaz: I reproduced this sequence, but it is not a fallback-specific regression. With a registered Apple, `move("apple"); setValue("Banana"); await flushBatch()` ends at `Apple` on `origin/main` and on this head. The move-to-value effect is intentionally batched, so the same-task `setValue` runs before that established effect; the late public fallback now preserves the same behavior. Its temporary subscription cancels value changes after the batch listener, matching the point at which the registered path has already selected Apple. Canceling a pre-listener write only in the fallback would make behavior depend on whether the item is registered. Making same-task calls last-write-wins would require a separate behavior change covering both paths.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +163,54 @@ export function createSelectStore({
...
+        setValueFromItem(item);
+        return;
+      }
+      const publicItem = items.find((item) => item.id === activeId);
+      if (!publicItem) return;
+      if (publicItem.disabled) return;
+      if (publicItem.value == null) return;
+
+      let canceled = false;
+      const stopValueSubscription = subscribe(select, ["value"], () => {
```
diegohaz, line 289: [P2] Cancel the fallback after same-value setter calls

This subscription only observes actual state changes, so a later setter that reasserts the current value is invisible (`setState` returns early for strict-equal values). On this head:

```ts
store.setState("items", [apple]);
store.move("apple");
queueMicrotask(() => store.setValue("Orange"));
await flushBatch();
```

Starting from `"Orange"`, the public-only path ends at `"Apple"`, while the identical registered-item path ends at `"Orange"` because its batch callback writes Apple before the queued setter runs. This is distinct from same-batch ordering: the setter runs after this listener, yet the deferred path reverses it solely because the option is unmounted. Please track value-write attempts across the pending window (for example, a generation bumped by the public setter) and add this same-value regression.

diegohaz (commented):
Re-reviewed the incremental updates from `008b852737de4cf45a34b116d41685b39a61ef76` through `75989cbd7b4790a9b19bfa261d063b9dae066910`.

The active-ID regression now proves its guard, and the temporary value subscription correctly cancels actual intervening changes, including the reported ABA sequence. One P2 remains inline: a later `setValue` call that reasserts the current value does not notify the subscription, so the deferred public-item path can still overwrite it and diverge from the registered-item path.

Validation:
- `pnpm test packages/ariakit-components/src/select/select-store.test.ts --run` — 9/9 passed
- `pnpm test-react app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- `pnpm test-react18 app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- `git diff --check 008b852737de4cf45a34b116d41685b39a61ef76..75989cbd7b4790a9b19bfa261d063b9dae066910` — clean
- All current PR checks pass

packages/ariakit-components/src/select/select-store.ts
diegohaz: Fixed in [fe5c11591](https://github.com/ariakit/ariakit/commit/fe5c11591271a8db97b1e8aaaaea52ef888013ec): the deferred public-item fallback now records later public setValue calls even when they reassert the current value, with focused regression coverage.

diegohaz:
**Fixed**

- [fe5c11591](https://github.com/ariakit/ariakit/commit/fe5c11591271a8db97b1e8aaaaea52ef888013ec) ([reply](https://github.com/ariakit/ariakit/pull/6736#discussion_r3585970974)): Canceled deferred public-item fallbacks after later same-value setValue calls.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,18 +163,59 @@ export function createSelectStore({
...
 
+  const setValueFromItem = (item: SelectStoreItem | null | undefined) => {
+    if (!item) return;
+    if (item.disabled) return;
+    if (item.value == null) return;
+    select.setState("value", item.value);
+  };
+
+  // Tracks strict-equal setter calls that state subscriptions don't observe.
+  let setValueVersion = 0;
```
diegohaz: [P2] Share setter invalidation across synchronized Select stores

This counter belongs to one `createSelectStore` closure, but `store` composition initializes the move-to-value effect in both Select stores. On this head:

```ts
const parent = createSelectStore({ defaultValue: "Orange" });
const child = createSelectStore({ store: parent });
const stop = init(child);

child.setState("items", [apple]);
child.move("apple");
queueMicrotask(() => child.setValue("Orange"));
await flushBatch();
```

With a public-only Apple, both stores finish at `"Apple"`; registering Apple first makes the same sequence finish at `"Orange"`. The child call increments only the child's counter, so the parent's pending fallback writes Apple back into the shared state; `parent.setValue("Orange")` fails symmetrically. This maps to `SelectProvider`'s supported `store` composition. The generic public `store.setState("value", "Orange")` path also bypasses the counter and reproduces the same registered/public split in a single store. Please put invalidation at a boundary shared by all synchronized writers and fallbacks (or ensure only one instance owns the fallback), and cover composed/provider and generic-setter variants.

diegohaz (commented):
Re-reviewed the incremental update from `75989cbd7b4790a9b19bfa261d063b9dae066910` through `fe5c11591271a8db97b1e8aaaaea52ef888013ec`.

The prior same-value `setValue` finding is resolved for a single direct Select store: the focused regression passes and isolates the new version guard. All five pre-existing review threads were resolved at the review snapshot, with no unresolved candidate threads.

One P2 remains inline: the invalidation counter is local to one Select store instance, so synchronized/composed Select stores can still overwrite a later same-value setter through another instance; the public generic `setState("value", ...)` path bypasses it too.

Validation:
- `pnpm test packages/ariakit-components/src/select/select-store.test.ts --run` — 10/10 passed
- `pnpm test-react app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- `pnpm test-react18 app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- Registered/public core probes for direct, generic, and composed setters
- `git diff --check 75989cbd7b4790a9b19bfa261d063b9dae066910..fe5c11591271a8db97b1e8aaaaea52ef888013ec` — clean
- All current PR checks, including browser and type checks, pass; I did not rerun Playwright locally

packages/ariakit-components/src/select/select-store.ts
diegohaz: Fixed in [`1319319e5`](https://github.com/ariakit/ariakit/commit/1319319e5f94581db3d6be7f8d520e3fd2faaeac): value-write invalidation now follows each Select store’s synchronized `value` roots, so parent/child, sibling, and derived-store writers share it while unrelated values stay independent. Both `setValue` and `setState("value", ...)` are covered by focused regressions.

diegohaz:
**Fixed**

- [`1319319e5`](https://github.com/ariakit/ariakit/commit/1319319e5f94581db3d6be7f8d520e3fd2faaeac) ([reply](https://github.com/ariakit/ariakit/pull/6736#discussion_r3586637194)): Shared value-write invalidation across synchronized Select roots and covered composed, sibling, derived-store, and generic value setters.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,27 +218,77 @@ export function createSelectStore({
...
+  };
+
   return {
     ...composite,
     ...popover,
     ...select,
     combobox,
-    setValue: (value) => select.setState("value", value),
+    setState,
+    setValue: (value) => setState("value", value),
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`d7ee8b21c`](https://github.com/ariakit/ariakit/commit/d7ee8b21cbce12eb39b9224b0659b574acb80a27): strict-equal value writes now propagate through the synchronized backing-store graph to the appropriate Select root trackers, with ordinary backing, derived `pick()`, unrelated-key, and asymmetric multi-root regression coverage.

packages/ariakit-store/src/index.ts
```diff
@@ -406,6 +413,32 @@ export function createStore<S extends State>(
...
+      const storeState = store?.getState?.();
+      if (!storeState) continue;
+      if (!hasOwnProperty(storeState, key)) continue;
+      const internals = (
+        store as Store & { __unstableInternals?: StoreInternals }
+      ).__unstableInternals;
+      internals?.notifySameValue(key);
+    }
+  };
+
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`b9f3885d0`](https://github.com/ariakit/ariakit/commit/b9f3885d09944f8e7bb90a9712da0338dcfa01bd): the parent notification now uses an optional method call, so stores with legacy-shaped `__unstableInternals` safely skip the new hook instead of throwing.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,27 +256,78 @@ export function createSelectStore({
...
+      queueMicrotask(() => {
+        stopValueSubscription();
+        if (canceled) return;
+        if (getValueWriteVersion(valueWriteTrackers) !== valueWriteVersion) {
+          return;
+        }
+        const currentState = select.getState();
+        if (currentState.moves !== moves) return;
+        if (currentState.activeId !== activeId) return;
+        if (currentState.value !== value) return;
```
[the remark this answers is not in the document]
diegohaz: I reproduced this sequence on `b9f3885d0`, but the base public-only result is the missing behavior this PR fixes, not the ordering reference:

```ts
store.move("apple");
store.setValue("Banana");
await flushBatch();
```

On the base commit, the registered path ends at `Apple`; the public-only path leaves `Banana` only because it does not resolve public items at all. On this head, both registered and public-only paths end at `Apple`, restoring parity with the established batched move-to-value effect. Preserving `Banana` only for the deferred path would make the outcome depend on item registration. A broader last-call-wins change would need to update both paths separately.

packages/ariakit-components/src/select/select-store.ts
```diff
@@ -162,27 +256,78 @@ export function createSelectStore({
...
+        canceled = true;
+        stopValueSubscription();
+      };
     }),
   );
 
+  const setState: SelectStore["setState"] = (key, value) => {
+    if (key === "value") {
+      for (const tracker of valueWriteTrackers) {
+        tracker.version += 1;
```
diegohaz, line 318: [P2] Do not invalidate shared roots from stale local setters

This bumps every shared-root tracker before establishing that the write can reach that root. For example, create sibling Select stores over one backing value, initialize only the mover, change the backing from `Orange` to `Banana` (leaving the writer's local value stale), then run:

```ts
mover.move("apple");
queueMicrotask(() => writer.setValue("Orange"));
await flushBatch();
```

With a public-only Apple, the stale writer's no-op local setter cancels the fallback and the backing/mover stay `Banana`; registering the same Apple makes both finish at `Apple`. A stopped writer reproduces this too. The unconditional bump here and key-only same-value propagation both treat a write that never reasserted the backing value as a root write. Please advance a root epoch only when the setter actually reaches/reasserts that root (for example, propagate the asserted value and stop at parents holding a different value), and cover an uninitialized/stopped sibling in both item paths.

packages/ariakit-store/src/index.ts
```diff
@@ -406,6 +413,32 @@ export function createStore<S extends State>(
...
+      listener(key);
+    }
+    for (const store of stores) {
+      const storeState = store?.getState?.();
+      if (!storeState) continue;
+      if (!hasOwnProperty(storeState, key)) continue;
+      const internals = (
+        store as Store & { __unstableInternals?: StoreInternals }
+      ).__unstableInternals;
+      internals?.notifySameValue?.(key);
```
diegohaz, line 438: [P2] Deduplicate same-value propagation through store DAGs

This recursion follows every parent path without a per-dispatch visited set. Repeatedly merging two picks of the preceding store creates only `3 * depth + 1` unique stores, but one strict-equal setter revisits the root `2 ** depth` times:

```ts
current = mergeStore(
  pick(current, ["value"]),
  pick(current, ["value"]),
);
```

At depth 20 I measured 1,048,576 root notifications and about 100 ms for just 61 unique stores; a 10,000-store pick chain also overflows the call stack. This path now runs for every strict-equal `setState`, even without a listener. Please carry a per-dispatch identity set (preferably with iterative traversal) so each store is notified once, and add diamond/deep-chain regressions to the store tests.

diegohaz (commented):
Re-reviewed the incremental update from `fe5c11591271a8db97b1e8aaaaea52ef888013ec` through `b9f3885d09944f8e7bb90a9712da0338dcfa01bd`.

The prior synchronized-store, direct-backing-store, and legacy-internals concerns are addressed. The one pre-existing unresolved candidate, [discussion_r3587078630](https://github.com/ariakit/ariakit/pull/6736#discussion_r3587078630), is invalid/inaccurate: the registered path ends at `Apple` on base and current, and the current public-only path now matches it. I replied with that parity evidence.

Two P2 findings remain inline:
- A stale uninitialized or stopped sibling Select can invalidate a live store's public-item fallback even though its setter does not reach the shared backing value, recreating registered/public divergence.
- Same-value notification recursively follows every path through composed-store DAGs, producing exponential duplicate work and stack-depth risk.

Validation:
- `pnpm test packages/ariakit-components/src/select/select-store.test.ts --run` — 25/25 passed
- `pnpm test packages/ariakit-store/src/test.ts --run` — 92/92 passed
- `pnpm test-react app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- `pnpm test-react18 app/src/sandbox/select-6733/test.ts --run` — 2/2 passed
- Focused lifecycle probe: the public stale-sibling case remained `Banana`, while the registered case reached `Apple`
- Focused graph probe: depth 20 produced 1,048,576 root notifications for 61 unique stores
- `git diff --check fe5c11591271a8db97b1e8aaaaea52ef888013ec..b9f3885d09944f8e7bb90a9712da0338dcfa01bd` passed; the worktree is clean
- All current PR checks pass

No other substantive correctness, compatibility, simplification, changeset, or performance findings.

diegohaz:

PR #6737, opened 2026-07-15
Fix Menu default-open focus

### LINKED ISSUE #2946: Menu: menu doesn't receive focus with `defaultOpen` and `modal=false`
### Current behavior

When setting `modal=false` and `defaultOpen`, the menu opens but focus is not moved to it

1. Open sandbox: https://stackblitz.com/edit/wzawnr?file=menu%2Fstyle.css,menu%2Findex.tsx&theme=dark
2. Play around with the `modal` prop
3. When `modal=true`, focus is correctly moved to the menu container
4. When `modal=false`, focus stays on the document's `body`, while the first menu item is visually highlighted

### Expected behavior

Regarldess of the `modal` prop, focus should be moved to the menu container (and no menu item should be therefore highlighted)

### Workaround

_No response_

### Possible solutions

_No response_

diegohaz:
## Motivation

A non-modal `Menu` that is conditionally mounted already open does not receive focus. This happens in flows that replace an action button with the next available menu, such as:

```tsx
<button onClick={() => setCreated(true)}>Add item</button>

{created && (
  <MenuProvider defaultOpen>
    <MenuButton>Actions for new item</MenuButton>
    <Menu modal={false}>...</Menu>
  </MenuProvider>
)}
```

The menu becomes visible, but focus falls back to the document body and the first item is marked active.

## Solution

Record whether a React menu store starts open, then let `Menu` consume that setup only when it renders in the store's initial commit. For an eligible non-modal menu, enable the store's `autoFocusOnShow` state after mount, mirroring the `MenuButton` interactions that normally opt into automatic focus.

The mount snapshot and commit-expiring setup preserve the surrounding focus behavior:

```tsx
<Menu modal={false}>...</Menu> // Focuses the menu container on the initial open mount.
<Menu modal>...</Menu> // Keeps Dialog's first-item autofocus behavior.
<Menu autoFocusOnShow={false}>...</Menu> // Leaves focus unchanged.
```

Replacement store wrappers retain the same initial setup without rerunning mount autofocus. Menus rendered lazily or opened after the store's first commit remain ineligible, while controlled menus rendered open in that initial commit are supported. The same behavior is available through the public `Menu` component and the deep-exported `useMenu` hook.

The `menu-2946` sandbox uses the concrete reported active-element flow and covers 11 mount, modal, opt-out, controlled, hook, replacement, hidden, lazy, and delayed scenarios in happy-dom and browser tests. The focused suite passes with React 18 and React 19, and the browser matrix passes in Chrome, Firefox, and Safari.

## Workaround

Before this fix is released, a menu that is intentionally mounted already open can create its store in the mounted subtree and enable autofocus after mount:

```tsx
function NewItemActions() {
  const store = useMenuStore({ defaultOpen: true });

  useEffect(() => {
    // TODO: Remove after https://github.com/ariakit/ariakit/issues/2946
    store.setAutoFocusOnShow(true);
  }, [store]);

  return <MenuProvider store={store}>...</MenuProvider>;
}
```

This workaround assumes the newly mounted menu should take focus and keeps the default menu-container target. It only covers the initial mount; later programmatic opens still need to set `autoFocusOnShow` as part of the opening action. For passive menus opened from restored state that should not move focus, keep `autoFocusOnShow={false}` on `Menu` instead.

### DIFF
--- a/.changeset/300-menu-default-open-focus.md
+++ b/.changeset/300-menu-default-open-focus.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Menu`](https://ariakit.com/reference/menu) to move focus inside when it mounts already open in non-modal mode. Thanks to [@ciampo](https://github.com/ciampo).
--- a/packages/ariakit-react-components/src/menu/menu-store.ts
+++ b/packages/ariakit-react-components/src/menu/menu-store.ts
@@ -1,9 +1,12 @@
 import * as Core from "@ariakit/components/menu/menu-store";
 import { useStore, useStoreProps } from "@ariakit/react-store";
 import type { Store } from "@ariakit/react-store";
-import { useUpdateEffect } from "@ariakit/react-utils";
+import {
+  useInitialValue,
+  useSafeLayoutEffect,
+  useUpdateEffect,
+} from "@ariakit/react-utils";
 import type { BivariantCallback, PickRequired } from "@ariakit/utils";
-import { useEffect } from "react";
 import { useComboboxProviderContext } from "../combobox/combobox-context.tsx";
 import type { ComboboxStore } from "../combobox/combobox-store.ts";
 import type {
@@ -20,6 +23,7 @@ import type {
 import { useHovercardStoreProps } from "../hovercard/hovercard-store.ts";
 import { useMenubarContext } from "../menubar/menubar-context.tsx";
 import type { MenubarStore } from "../menubar/menubar-store.ts";
+import { setMenuStoreSetup } from "./__utils.ts";
 import { useMenuContext } from "./menu-context.tsx";
 
 export function useMenuStoreProps<T extends Core.MenuStore>(
@@ -75,15 +79,18 @@ export function useMenuStore(props: MenuStoreProps = {}): MenuStore {
     combobox: props.combobox !== undefined ? props.combobox : combobox,
   };
   const [store, update] = useStore(Core.createMenuStore, props);
-  useEffect(() => {
-    // Menus that mount already open don't go through the MenuButton
-    // interactions that enable automatic focus.
-    const { open, autoFocusOnShow } = store.getState();
-    if (!open) return;
-    if (autoFocusOnShow) return;
-    store.setAutoFocusOnShow(true);
-  }, [store]);
-  return useMenuStoreProps(store, update, props);
+  const setup = useInitialValue(() => ({
+    autoFocusOnMount: store.getState().open,
+    hasCommitted: false,
+  }));
+  // Keep the initial setup associated with replacement store wrappers.
+  setMenuStoreSetup(store, setup);
+  const storeWithProps = useMenuStoreProps(store, update, props);
+  // Mount autofocus only applies to menus rendered in the initial commit.
+  useSafeLayoutEffect(() => {
+    setup.hasCommitted = true;
+  }, [setup]);
+  return storeWithProps;
 }
 
 export type MenuStoreValues = Core.MenuStoreValues;
--- a/packages/ariakit-react-components/src/menu/menu-store.ts
+++ b/packages/ariakit-react-components/src/menu/menu-store.ts
@@ -3,6 +3,7 @@ import { useStore, useStoreProps } from "@ariakit/react-store";
 import type { Store } from "@ariakit/react-store";
 import { useUpdateEffect } from "@ariakit/react-utils";
 import type { BivariantCallback, PickRequired } from "@ariakit/utils";
+import { useEffect } from "react";
 import { useComboboxProviderContext } from "../combobox/combobox-context.tsx";
 import type { ComboboxStore } from "../combobox/combobox-store.ts";
 import type {
@@ -74,6 +75,14 @@ export function useMenuStore(props: MenuStoreProps = {}): MenuStore {
     combobox: props.combobox !== undefined ? props.combobox : combobox,
   };
   const [store, update] = useStore(Core.createMenuStore, props);
+  useEffect(() => {
+    // Menus that mount already open don't go through the MenuButton
+    // interactions that enable automatic focus.
+    const { open, autoFocusOnShow } = store.getState();
+    if (!open) return;
+    if (autoFocusOnShow) return;
+    store.setAutoFocusOnShow(true);
+  }, [store]);
   return useMenuStoreProps(store, update, props);
 }
 
--- a/app/src/sandbox/menu-2946/index.react.tsx
+++ b/app/src/sandbox/menu-2946/index.react.tsx
@@ -1,0 +1,20 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+
+export default function Example() {
+  const [created, setCreated] = useState(false);
+
+  if (!created) {
+    return <button onClick={() => setCreated(true)}>Add item</button>;
+  }
+
+  return (
+    <Ariakit.MenuProvider defaultOpen>
+      <Ariakit.MenuButton>Actions for new item</Ariakit.MenuButton>
+      <Ariakit.Menu modal={false}>
+        <Ariakit.MenuItem>Rename</Ariakit.MenuItem>
+        <Ariakit.MenuItem>Duplicate</Ariakit.MenuItem>
+      </Ariakit.Menu>
+    </Ariakit.MenuProvider>
+  );
+}
--- a/app/src/sandbox/menu-2946/index.react.tsx
+++ b/app/src/sandbox/menu-2946/index.react.tsx
@@ -1,5 +1,24 @@
 import * as Ariakit from "@ariakit/react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
+
+function NewItemActions() {
+  const store = Ariakit.useMenuStore({ defaultOpen: true });
+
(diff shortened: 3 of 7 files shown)


packages/ariakit-react-components/src/menu/menu-store.ts
```diff
@@ -74,6 +75,14 @@ export function useMenuStore(props: MenuStoreProps = {}): MenuStore {
...
   };
   const [store, update] = useStore(Core.createMenuStore, props);
+  useEffect(() => {
+    // Menus that mount already open don't go through the MenuButton
+    // interactions that enable automatic focus.
+    const { open, autoFocusOnShow } = store.getState();
+    if (!open) return;
+    if (autoFocusOnShow) return;
+    store.setAutoFocusOnShow(true);
+  }, [store]);
```
diegohaz: **[P2] Do not rerun mount autofocus after store replacement**

The `store` wrapper is not stable for the lifetime of this hook. `useMenuStoreProps` calls `update()` when `combobox`, `parent`, or `menubar` changes, and `useStore` then replaces this wrapper. Consequently, an already-open menu whose `autoFocusOnShow` state has been reset to `false` runs this effect again, flips the state back to `true`, and can steal focus even though the menu was not shown again.

I reproduced this by clicking an item in an open menu, calling `setAutoFocusOnShow(false)`, and changing its `parent` relationship with `autoFocusOnHide={false}`: this head moves focus from the item to the menu container, while removing this effect retains item focus. At minimum, capture the initial wrapper so later composition updates cannot retrigger mount behavior:

```tsx
const initialStore = useInitialValue(store);

useEffect(() => {
  const { open, autoFocusOnShow } = initialStore.getState();
  if (!open || autoFocusOnShow) return;
  initialStore.setAutoFocusOnShow(true);
}, [initialStore]);
```

I also tested the combined alternative of putting this initial-only logic in `useMenu`; it preserves this PR's two regression cases and the store-replacement case.

packages/ariakit-react-components/src/menu/menu-store.ts
```diff
@@ -74,6 +75,14 @@ export function useMenuStore(props: MenuStoreProps = {}): MenuStore {
     combobox: props.combobox !== undefined ? props.combobox : combobox,
   };
   const [store, update] = useStore(Core.createMenuStore, props);
+  useEffect(() => {
+    // Menus that mount already open don't go through the MenuButton
+    // interactions that enable automatic focus.
+    const { open, autoFocusOnShow } = store.getState();
+    if (!open) return;
+    if (autoFocusOnShow) return;
+    store.setAutoFocusOnShow(true);
```
diegohaz: **[P2] Preserve the existing modal initial-focus target**

This setter also runs for modal menus, although the reported bug and changeset are scoped to non-modal behavior. For the following public setup, the base behavior focuses and activates `Rename` through Dialog's first-tabbable fallback, but this head focuses the menu container and leaves every item inactive:

```tsx
<MenuProvider defaultOpen>
  <MenuButton>Actions</MenuButton>
  <Menu modal>
    <MenuItem>Rename</MenuItem>
  </Menu>
</MenuProvider>
```

I verified that behavior in React Strict Mode and then removed only this new effect; item focus returned. Changing the target can alter Enter-key behavior and consumer focus handlers, and it is unnecessary for fixing the non-modal path.

Please perform the mount-open initialization where the effective `modal` value is known and skip modal menus. For example, an initial-only effect in `useMenu` preserves both behaviors:

```tsx
const initial = useInitialValue({ store, modal });

useEffect(() => {
  if (initial.modal) return;
  const { open, autoFocusOnShow } = initial.store.getState();
  if (!open || autoFocusOnShow) return;
  initial.store.setAutoFocusOnShow(true);
}, [initial]);
```

This version passed the PR's two tests plus focused modal and store-replacement probes. Please add the modal case as regression coverage.

diegohaz (commented):
Initial review of `bd7b6779dc742914af2ee6bd52cf953fe2d74f26..27969558e89e7c98398888767f91a31ce83ce9bb`.

I reviewed the full five-file diff and the surrounding Menu/MenuButton, Hovercard, Popover, Dialog, core-store, React-store, fixture-discovery, public-export, changeset, issue, related-history, CI, and performance context. I found two P2 focus regressions in the new store effect:

- [The effect reruns after composed-store replacement](https://github.com/ariakit/ariakit/pull/6737#discussion_r3585596013), so an already-open menu can re-enable autofocus and steal focus without a new show transition.
- [The setter also changes default-open modal behavior](https://github.com/ariakit/ariakit/pull/6737#discussion_r3585597325), moving focus from the first item to the menu container outside the non-modal scope of the fix.

A single initial-only effect in `useMenu`, guarded by the effective `modal` value, addressed both findings in a local prototype.

Validation performed:

- `pnpm test-react app/src/sandbox/menu-2946/test.ts` — 2 passed.
- `pnpm test-react18 app/src/sandbox/menu-2946/test.ts` — 2 passed.
- The same React regression test failed on known-broken commit `3b028bd9` because focus remained on `body` and the first item became active.
- Isolated base/head probes reproduced the store-replacement and modal-focus regressions.
- The proposed boundary passed the two PR tests plus focused store-replacement and modal probes.
- `git diff --check bd7b6779..27969558` passed.
- Current CI is green, including React, React 18, Chrome, Firefox, Safari, type checks, builds, and performance; performance reports no significant change.

Review-thread validation: the authoritative, fully paginated `reviewThreads` result contained zero threads, so there were no unresolved candidate findings to adjudicate or duplicate. The changeset packages, patch level, reference link, and contributor credit are correct.

I did not rerun Playwright locally; the current-head Chrome, Firefox, and Safari jobs all passed.

packages/ariakit-react-components/src/menu/menu-store.ts
diegohaz: Fixed in [`9ad424635`](https://github.com/ariakit/ariakit/commit/9ad4246350abb8ddb31785f508761cc57c1d9931): mount autofocus now snapshots the initial store/setup and carries that setup across replacement wrappers without rerunning after commit. The relationship-replacement regression verifies focus remains on the menu item.

packages/ariakit-react-components/src/menu/menu-store.ts
diegohaz: Fixed in [`9ad424635`](https://github.com/ariakit/ariakit/commit/9ad4246350abb8ddb31785f508761cc57c1d9931): mount autofocus now receives the effective modal state and skips modal menus, preserving Dialog’s first-item focus. The modal regression runs in both happy-dom and the browser matrix.

diegohaz:
**Fixed**

- [`9ad424635`](https://github.com/ariakit/ariakit/commit/9ad4246350abb8ddb31785f508761cc57c1d9931) ([reply](https://github.com/ariakit/ariakit/pull/6737#discussion_r3586370820)): preserve initial mount intent across store-wrapper replacement without refocusing after commit.
- [`9ad424635`](https://github.com/ariakit/ariakit/commit/9ad4246350abb8ddb31785f508761cc57c1d9931) ([reply](https://github.com/ariakit/ariakit/pull/6737#discussion_r3586370946)): skip mount autofocus for effective modal menus so Dialog keeps first-item focus.

Both behaviors are covered in the 11-scenario happy-dom and cross-browser regression matrix.

.changeset/300-menu-default-open-focus.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Menu`](https://ariakit.com/reference/menu) to move focus inside when it mounts already open in non-modal mode. Thanks to [@ciampo](https://github.com/ciampo).
```
diegohaz, line 6: **[P3] Scope the release note to the initial store commit**

This currently promises that every non-modal `Menu` mounting open moves focus inside, but the new `late` and `late-controlled` regressions intentionally assert the opposite after the store has committed: the `Menu` mounts already open while focus stays outside. Since this changeset becomes user-facing release text, please qualify the eligibility window. For example:

```suggestion
Fixed [`Menu`](https://ariakit.com/reference/menu) to move focus inside when it mounts already open in non-modal mode as part of the initial store commit. Thanks to [@ciampo](https://github.com/ciampo).
```

diegohaz (commented):
Re-reviewed `27969558e..9ad424635` plus the full PR context.

The two previous P2 findings are resolved:

- Replacement store wrappers retain the committed setup, and the relationship regression keeps focus on the item.
- Effective modal menus skip the container override, and the modal regression keeps the first item focused and active.

I left one new P3 inline because the changeset describes a broader mount-open behavior than the implementation and its late-mount tests provide.

I also independently confirmed [CodeRabbit’s existing duplicate-hook finding](https://github.com/ariakit/ariakit/pull/6737#pullrequestreview-4703201500): public `Menu` invokes `useAutoFocusOnMount`, then `MenuImpl` reaches the same hook through `useMenu`. The mounted gate has no eligible path that requires the outer call, so the simpler ownership remains:

```tsx
export const Menu = createDialogComponent(
  MenuImpl,
  useMenuProviderContext,
);
```

Removing the outer call in a scratch worktree, including an `unmountOnHide` probe, passed all 11 focused cases on React 19 and React 18 and all 33 Chrome/Firefox/Safari cases. The current head also passes both 11-test React suites, the complete CI matrix is green, and `git diff --check` passes. No further correctness findings.

diegohaz:

PR #3298, opened 2024-01-04
Implement custom content typing

georgekaran:
Closes #2699

PR #6745, opened 2026-07-16
Support custom Select typeahead text

### LINKED ISSUE #2699: Focusing Select items with custom content by typing
### Discussed in https://github.com/ariakit/ariakit/discussions/2698

<div type='discussions-op-text'>

<sup>Originally posted by **Dremora** August  7, 2023</sup>
In a [Select example](https://ariakit.org/components/select), once the select is open, one can type a letter to focus an item that starts with that letter.

I have a [slightly modified version of this component](https://stackblitz.com/edit/hffy3v?file=select%2Fstyle.css,select%2Findex.tsx&theme=dark) that displays emojis in front of text. Typing to focus doesn't work anymore.

Is there any documentation on how this feature is implemented, and how to make it work with custom content?</div>

diegohaz:
## Motivation

Select typeahead currently matches an item's rendered text content. When decorative or custom content appears before the human-readable label, typing the label's first character cannot focus the item.

```tsx
<SelectItem value="Canada">
  <span aria-hidden>🇨🇦</span> Canada
</SelectItem>
```

Thanks to @Dremora for reporting this in #2699 and @georgekaran for the investigation and implementation work in #3298 that informed this solution.

## Solution

Add a `typeaheadText` prop to `CompositeItem` and components built on it, including `SelectItem` and `ComboboxItem`. The explicit text takes precedence over rendered text, item children, and value fallbacks.

```tsx
<SelectItem typeaheadText={country.value} value={country.value}>
  <span aria-hidden>{country.flag}</span> {country.value}
</SelectItem>
```

An empty string excludes an item from typeahead matching, and prop updates immediately update its registered match text. Passive and offscreen items carry the same metadata on their placeholder elements so virtualized lists behave consistently without leaking the camelCase prop to the DOM.

The regression coverage exercises open and closed Selects, empty-string opt-out, runtime prop updates, passive offscreen items, disabled offscreen semantics, and DOM prop filtering in both browser and happy-dom tests. The commit history intentionally records the failing reproduction, the passing userland workaround, and the final library fix as separate stages.

## Workaround

When content before the item label is purely decorative, render it with CSS generated content so the item's actual text starts with the searchable label. Keep an explicit accessible name so generated content does not affect it.

```tsx
// TODO: Remove after https://github.com/ariakit/ariakit/issues/2699 is fixed.
<SelectItem
  aria-label={country.value}
  className="item"
  data-flag={country.flag}
  value={country.value}
>
  {country.value}
</SelectItem>
```

```css
.item::before {
  margin-inline-end: 0.25em;
  content: attr(data-flag);
}
```

This workaround is limited to decorative content that can be rendered with CSS. It does not cover arbitrary rich content, aliases, or searchable text that intentionally differs from the visible label.

Fixes #2699

### DIFF
--- a/.changeset/200-composite-typeahead-text.md
+++ b/.changeset/200-composite-typeahead-text.md
@@ -4,4 +4,16 @@
 "@ariakit/react": patch
 ---
 
-Added the [`typeaheadText`](https://ariakit.com/reference/composite-item#typeaheadtext) prop to [`CompositeItem`](https://ariakit.com/reference/composite-item) and components built on it, including [`SelectItem`](https://ariakit.com/reference/select-item) and [`ComboboxItem`](https://ariakit.com/reference/combobox-item), so typeahead can match a label independently of custom rendered content. Thanks to [@Dremora](https://github.com/Dremora) and [@georgekaran](https://github.com/georgekaran).
+Custom typeahead text for composite items
+
+The new [`typeaheadText`](https://ariakit.com/reference/composite-item#typeaheadtext) prop lets [`CompositeItem`](https://ariakit.com/reference/composite-item) use an explicit label for typeahead matching when its rendered content starts with an emoji or other decoration.
+
+```tsx
+<SelectItem typeaheadText="Canada" value="Canada">
+  <span aria-hidden>🇨🇦</span> Canada
+</SelectItem>
+```
+
+Set [`typeaheadText`](https://ariakit.com/reference/composite-item#typeaheadtext) to an empty string to exclude an item from typeahead matching. The prop is also available on these components exported by `@ariakit/react` and built on [`CompositeItem`](https://ariakit.com/reference/composite-item): [`ComboboxItem`](https://ariakit.com/reference/combobox-item), [`FormRadio`](https://ariakit.com/reference/form-radio), [`MenuItem`](https://ariakit.com/reference/menu-item), [`MenuItemCheckbox`](https://ariakit.com/reference/menu-item-checkbox), [`MenuItemRadio`](https://ariakit.com/reference/menu-item-radio), [`Radio`](https://ariakit.com/reference/radio), [`SelectItem`](https://ariakit.com/reference/select-item), [`Tab`](https://ariakit.com/reference/tab), [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container), [`ToolbarInput`](https://ariakit.com/reference/toolbar-input), and [`ToolbarItem`](https://ariakit.com/reference/toolbar-item).
+
+Thanks to [@Dremora](https://github.com/Dremora) for reporting the issue and providing the reproduction, and [@georgekaran](https://github.com/georgekaran) for the investigation and implementation work that informed this solution.
--- a/.changeset/200-composite-typeahead-text.md
+++ b/.changeset/200-composite-typeahead-text.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added the [`typeaheadText`](https://ariakit.com/reference/composite-item#typeaheadtext) prop to [`CompositeItem`](https://ariakit.com/reference/composite-item) and components built on it, including [`SelectItem`](https://ariakit.com/reference/select-item) and [`ComboboxItem`](https://ariakit.com/reference/combobox-item), so typeahead can match a label independently of custom rendered content. Thanks to [@Dremora](https://github.com/Dremora) and [@georgekaran](https://github.com/georgekaran).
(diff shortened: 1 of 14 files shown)


.changeset/200-composite-typeahead-text.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Added the [`typeaheadText`](https://ariakit.com/reference/composite-item#typeaheadtext) prop to [`CompositeItem`](https://ariakit.com/reference/composite-item) and components built on it, including [`SelectItem`](https://ariakit.com/reference/select-item) and [`ComboboxItem`](https://ariakit.com/reference/combobox-item), so typeahead can match a label independently of custom rendered content. Thanks to [@Dremora](https://github.com/Dremora) and [@georgekaran](https://github.com/georgekaran).
```
diegohaz: This could benefit from a multiline changeset with a code example.

packages/ariakit-react-components/src/composite/composite-typeahead.tsx
```diff
@@ -177,7 +178,11 @@ export const useCompositeTypeahead = createHook<
     for (const element of offscreenItems) {
       if (element.hasAttribute("data-disabled")) continue;
       if ("disabled" in element && !!element.disabled) continue;
-      enabledItems.push({ id: element.id, element });
+      enabledItems.push({
+        id: element.id,
+        element,
+        typeaheadText: element.dataset.typeaheadText,
```
diegohaz, line 184: [P2] Keep offscreen metadata separate from caller data attributes

This treats every caller-provided `data-typeahead-text` as Ariakit state. The inactive wrapper later spreads caller props after the generated metadata (`composite-item-offscreen.tsx:103`), so a native prop that was previously inert can now change matching and can also override the documented `typeaheadText` prop. For example, this item used to match `Canada`, but now matches `analytics-token` instead:

```tsx
<OffscreenSelectItem
  data-typeahead-text="analytics-token"
  offscreenMode="passive"
  value="Canada"
/>
```

I reproduced that fallback regression locally, and also confirmed that the same attribute overrides an explicit `typeaheadText`. Please carry the value in private element metadata (for example, a shared `WeakMap` or symbol property), or use a reserved namespaced channel with internal precedence, then cover both a caller-owned data attribute and the explicit-prop conflict.

diegohaz (commented):
Initial review of merge base [`7368196`](https://github.com/ariakit/ariakit/commit/73681967e6dd34004b01fd308ec9412ce2d08b84) through head [`450637c`](https://github.com/ariakit/ariakit/commit/450637c0022b5c7983870eff7547fd83da003284). I reviewed all 13 changed files plus the surrounding collection registration lifecycle, Select/Combobox composition, offscreen rendering, typeahead ordering and disabled behavior, public re-exports, both reference extractors, changeset/package coverage, contributor credit, and performance implications.

### Findings

- **Address now:** [offscreen matching currently consumes caller-owned `data-typeahead-text`](https://github.com/ariakit/ariakit/pull/6745#discussion_r3592634239), which can change existing fallback matching and override the documented prop.
- No registered follow-ups.

### Relevant unresolved thread validation

- `PRRT_kwDOBjT0Fc6RUXwB` ([changeset example suggestion](https://github.com/ariakit/ariakit/pull/6745#discussion_r3592574607)): **Declined non-goal.** A multiline example would be useful optional polish, but the current simple entry follows the changeset format, accurately names the public behavior and affected packages, and the generated API reference already carries the example. I did not duplicate or reply to the existing thread.

### Validation

- `pnpm test-react app/src/sandbox/select-typeahead-2699/test.ts app/src/sandbox/composite-offscreen-props/test.ts app/src/sandbox/select-6347/test.ts app/src/sandbox/composite-typeahead-buffer/test.ts` (4 files, 9 tests passed)
- `pnpm test-react18 app/src/sandbox/select-typeahead-2699/test.ts app/src/sandbox/composite-offscreen-props/test.ts` (2 files, 5 tests passed)
- `pnpm test app/src/lib/jsdoc-loader.test.ts website/build-pages/reference-utils.test.ts` (2 files, 4 tests passed)
- `pnpm tsc`
- `pnpm build-website-lite`
- Direct modern and legacy extractor inspection confirmed `typeaheadText` on `CompositeItem`, `SelectItem`, and `ComboboxItem`.
- Mutation checks confirmed the existing custom-content offscreen test fails when the dataset read is removed, while a caller-owned `data-typeahead-text` reproduction fails with the new read and passes without it. All experiments were restored and the worktree is clean.
- The test-only commit [`7bbb99a`](https://github.com/ariakit/ariakit/commit/7bbb99a4da27af4e047fa3abd8619f09d20867c7) failed React, React 18, and browser CI for the expected open/closed regressions; exact head CI is green, including Chrome, Firefox, Safari, typecheck, package/website builds, and performance checks.

I did not rerun Playwright locally; browser validation relies on the exact-head CI runs. No packaging files or exports changed, so a pack inspection was not applicable.

.changeset/200-composite-typeahead-text.md
diegohaz: Use multiline changeset with code example for new features like this

PR #6751, opened 2026-07-16
Improve popover show and hide performance

### LINKED ISSUE #4428: Performance drop when quickly cycling through toolbar items with tooltips
### Current behavior

I have a toolbar where each toolbar item is the anchor of a tooltip. I noticed a slowdown when holding down the arrow keys to cycle through them. The slowdown is less noticeable in production mode, but it's still there. I looked at the profiler and I can see that CPU usage is very high. Trying different combinations of `unmountOnHide` and `portal` didn't change the outcome. 

I also tried a different approach where I just render a single tooltip with many anchors and change it's content based on the current anchor, but it didn't solve the issue. 

1. Open https://stackblitz.com/edit/ariakit-ztpvbexl?devToolsHeight=33&file=src%2FApp.tsx
2. Start the profiler in your browser
3. Focus the toolbar using the keyboard
4. Hold down the left or right arrow key to cycle through the toolbar items

You can see that the performance issues are there in the last case, where there are no tooltips.

This slowdown can also be seen in the Ariakit website, in components that have composite items that render something when focused: https://ariakit.org/components/menubar

### Expected behavior

Performance shouldn't drop so much. In more resource-intensive web applications it results in very noticeable FPS drops. 

### Workaround

_No response_

### Possible solutions

_No response_

diegohaz:
Closes #4428

## Motivation

Rapidly cycling through toolbar items with tooltips can do substantially more work than the same interaction without tooltips (https://github.com/ariakit/ariakit/issues/4428).

The `tooltip-perf` sandbox measures this with 121 back-to-back `ArrowRight` presses across three toolbar items. That is a useful saturation and throughput workload because it repeatedly exercises the real public component path, but it is not a model of OS-paced human input. Its scripting and rendering results should be interpreted as aggregate main-thread work and available throughput, not as user-visible input latency.

Profiling also exposed work that affects ordinary pages independently of rapid key presses. Closed popovers stay connected by default, so they kept Floating UI scroll and resize listeners and observers active even though the default position update immediately returns while the popover is unmounted. Popovers also subscribed to `disclosureElement` in cases where it cannot be used as a portal tab-order anchor.

## Solution

Keep the two local Popover optimizations that directly remove this work:

- Skip the default positioning setup while a connected popover is both closed and hidden. The public `--popover-overflow-padding` variable is still written first. Popovers using a custom `updatePosition`, `alwaysVisible`, or `hidden={false}` preserve their previous behavior.
- Subscribe to `disclosureElement` only when portal tab-order preservation can apply:

```tsx
const shouldPreserveTabOrder = preserveTabOrder && portal && !modal;
```

The broader Dialog focus-restoration and Disclosure animation-state rewrites from the earlier revision were removed. They changed foundational lifecycle state machines, while the combined benchmark did not isolate enough incremental value to justify that complexity.

The retained behavior is covered by paired browser and happy-dom sandboxes. The subscription fixture covers all three static gate terms: modal popovers, non-portaled popovers, and portaled popovers with `preserveTabOrder={false}`.

## Performance evidence

The [exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29589062500) compared [`6efccb045`](https://github.com/ariakit/ariakit/commit/6efccb0458cad38690708fc12c0242a2bdb69837) with the PR merge base, [`948fb0ad9`](https://github.com/ariakit/ariakit/commit/948fb0ad9d0c3b685e9e0cbb204392e674ff5eea). All benchmark and report jobs passed. The raw tooltip artifact contains four interleaved baseline/current round pairs with five samples per test in each round.

| Tooltip workload | Scripting | Rendering | INP | Total |
| --- | ---: | ---: | ---: | ---: |
| Independent tooltips | 248.4ms → 244.2ms (-2%) | 76.8ms → 70.3ms (-8%) | 24ms → 24ms | 325.2ms → 314.5ms (-3%) |
| Shared tooltip | 163.7ms → 157.3ms (-4%) | 88.7ms → 72.5ms (-18%, confirmed) | 24ms → 24ms | 252.6ms → 229.6ms (-9%) |
| No tooltips (control) | 90.4ms → 91.5ms (+1%) | 35.7ms → 36.2ms (+1%) | 16ms → 16ms | 126.3ms → 128ms (+1%) |

Only the shared-tooltip rendering result crossed the workflow's significance threshold and was consistent across the paired raw rounds. There was no measured INP improvement. This supports reduced aggregate rendering work in a rapid keyboard-navigation stress case; it does not establish that typical users press keys at this cadence or experience an equivalent latency improvement.

## Recommendation

Merge this narrowed revision. The remaining changes are local guards around work that is unnecessary in ordinary closed-popover states, preserve explicit escape hatches, and have regression coverage. The stress benchmark is useful corroborating evidence, but it should not be used to justify the broader lifecycle state-machine changes that were removed.

## Validation

- Focused React fixture tests pass for the hidden overflow-padding behavior and all three disclosure subscription conditions.
- The new `preserveTabOrder={false}` case was red-checked by removing only that condition. The Popover render count changed from 24 to 26, then remained stable again after restoring the guard.
- The production app build passes.
- The two focused Chrome sandbox files pass all browser tests against a preview verified to come from this checkout.
- Full lint and formatting checks pass.
- The exact-head GitHub workflows, including main tests, Chrome/Firefox/Safari app tests, Plus, CodeQL, and performance, pass.


packages/ariakit-react-components/src/popover/popover.tsx
```diff
@@ -266,7 +267,13 @@ export const usePopover = createHook<TagName, PopoverOptions>(
 
     const arrowElement = useStoreState(store, "arrowElement");
     const anchorElement = useStoreState(store, "anchorElement");
-    const disclosureElement = useStoreState(store, "disclosureElement");
+    // The disclosure element is only used as the preserveTabOrder anchor, so
+    // don't subscribe to it when that feature is disabled.
+    const disclosureElement = useStoreState(
+      store,
+      preserveTabOrder ? ["disclosureElement"] : [],
```
diegohaz: **Address now:** `useDialog` disables the effective `preserveTabOrder` value for modal dialogs (`preserveTabOrderProp && !modal && state.mounted`), but this subscription is gated only by the raw prop, whose default is `true`. Consequently, a modal popover still subscribes to `disclosureElement` and rerenders when it changes even though the value cannot be used as a tab-order anchor. A focused render-count test reproduced an extra render (`2 → 3`) after `store.setDisclosureElement(...)`; adding the modal gate kept the count at 2.

Could this use the same static condition in both subscription arguments?

```tsx
const shouldPreserveTabOrder = preserveTabOrder && !modal;
const disclosureElement = useStoreState(
  store,
  shouldPreserveTabOrder ? ["disclosureElement"] : [],
  (state) =>
    shouldPreserveTabOrder ? state.disclosureElement : null,
);
```

This keeps the optimization effective for modal Popover consumers as well.

diegohaz (commented):
Initial review of [`948fb0a`](https://github.com/ariakit/ariakit/commit/948fb0ad9d0c3b685e9e0cbb204392e674ff5eea) → [`ee7cdc2`](https://github.com/ariakit/ariakit/commit/ee7cdc2654350b047a466b47c99a79c35148a81d).

## Outcome

- **Address now:** 1 inline finding: modal popovers still subscribe to `disclosureElement`, despite modal mode disabling the effective `preserveTabOrder` behavior.
- **Follow-up:** none.
- **Optional:** none.

## Scope reviewed

- All 13 changed files plus the surrounding dialog focus-restoration, disclosure animation lifecycle, popover positioning, portal/tab-order, menu, combobox, and tooltip code paths.
- Both changesets, their public API documentation links, issue references, and contributor attribution.
- Existing reviews and review threads; there were no unresolved inline threads before this review.

## Validation

- `pnpm test-react app/src/sandbox/dialog-strictmode-open/test.ts app/src/sandbox/disclosure-content-lifecycle/test.ts app/src/sandbox/popover-closed-overflow-padding/test.ts --run` — 4/4 tests passed.
- `pnpm test-react18 app/src/sandbox/dialog-strictmode-open/test.ts app/src/sandbox/disclosure-content-lifecycle/test.ts app/src/sandbox/popover-closed-overflow-padding/test.ts --run` — 4/4 tests passed.
- Focused temporary render-count check for the inline finding: current code rerendered a modal popover after `store.setDisclosureElement(...)` (`2 → 3` renders); gating with `preserveTabOrder && !modal` kept the count at 2. The experiment was removed afterward.
- `git diff --check 948fb0ad9d0c3b685e9e0cbb204392e674ff5eea..HEAD` passed.
- Exact-head CI is green, including Chrome, Firefox, Safari, React variants, type checks, and package tests.
- [Final exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29485030461) succeeded. The downloaded tooltip artifact contained four paired rounds with complete baseline/current samples; the final report showed shared rendering `-32%` and total `-16%`, independent rendering `-12%` and total `-8%`, with no significant Node regression.

## Candidate disposition

- The existing CodeRabbit top-level concern about the modal subscription gate was independently reproduced and is the **Address now** inline finding.
- No other correctness, regression-coverage, API-surface, or behavior-preserving simplification concern survived validation.

## Validation limits

- I did not rerun the browser suites locally; I relied on their green exact-head Chrome/Firefox/Safari CI jobs and inspected the raw exact-head tooltip performance artifact.
- I ran the targeted React and React 18 regression suites rather than every repository test locally.

packages/ariakit-react-components/src/popover/popover.tsx
diegohaz: Fixed in [`fdfbdf3be`](https://github.com/ariakit/ariakit/commit/fdfbdf3be651e8bb65bc40ca89e6c590d25b206a): the subscription now gates on `preserveTabOrder && portal && !modal`. The `portal` term is included because the portal's tab-order anchor effect also requires the `portal` prop, so plain non-portaled popovers (the non-modal default) skip the subscription too. The same commit adds a render-count regression sandbox (`popover-disclosure-render`) covering both the modal and the non-portal case; each test was red-checked against the respective ungated code. A `mounted` term and an anchor-override (`preserveTabOrderAnchor` in props) term were considered and left out: disclosure element writes while closed batch with already-subscribed state changes, and components that override the anchor use stable disclosure elements whose writes are same-value no-ops after the first open, so either term would buy at most one render per component lifetime.

diegohaz:
**Fixed**

- [`fdfbdf3be`](https://github.com/ariakit/ariakit/commit/fdfbdf3be651e8bb65bc40ca89e6c590d25b206a) ([reply](https://github.com/ariakit/ariakit/pull/6751#discussion_r3595100504)): the `disclosureElement` subscription now gates on `preserveTabOrder && portal && !modal`, matching the static conditions under which the portal's tab-order anchor can actually be consumed, so modal popovers and plain non-portaled popovers no longer re-render when the disclosure element changes. Includes a render-count regression sandbox (`popover-disclosure-render`) covering both cases, red-checked per gate term.

diegohaz (commented):
Re-review of [`ee7cdc2`](https://github.com/ariakit/ariakit/commit/ee7cdc2654350b047a466b47c99a79c35148a81d) → [`fdfbdf3`](https://github.com/ariakit/ariakit/commit/fdfbdf3be651e8bb65bc40ca89e6c590d25b206a).

## Outcome

The prior **Address now** finding is resolved. I found no newly introduced substantive findings in the commits since the last reviewed SHA.

The current gate matches the static conditions under which the disclosure element can be consumed as the portal tab-order anchor:

```tsx
const shouldPreserveTabOrder = preserveTabOrder && portal && !modal;
```

## Scope reviewed

- All four files changed since the prior marked review: the Popover subscription change and the new `popover-disclosure-render` React fixture, browser test, and happy-dom test.
- The surrounding `usePopover`, `useDialog`, `usePortal`, keyed `useStoreState`, disclosure-element writers, and derived Popover consumers.
- The full resolved thread [PRRT_kwDOBjT0Fc6RaB3R](https://github.com/ariakit/ariakit/pull/6751#discussion_r3594642929), including the fix reply.
- Current PR description, changesets, exact-head checks, and performance workflow status.

## Validation

- `pnpm test-react app/src/sandbox/popover-disclosure-render/test.ts --run` passed, 2/2 tests.
- `pnpm test-react18 app/src/sandbox/popover-disclosure-render/test.ts --run` passed, 2/2 tests.
- Red-check for the modal term: temporarily changing the gate to `preserveTabOrder && portal` made only the modal regression fail (`22 → 24` Popover renders).
- Red-check for the portal term: temporarily changing the gate to `preserveTabOrder && !modal` made only the plain non-portaled regression fail (`18 → 20` Popover renders).
- A temporary positive-path test confirmed that a Popover initially rendered with `portal={false}` does not react to disclosure updates, then reattaches the keyed subscription after `portal` becomes `true` and observes later updates. The experiment was removed.
- `git diff --check ee7cdc2654350b047a466b47c99a79c35148a81d..fdfbdf3be651e8bb65bc40ca89e6c590d25b206a` passed.
- All exact-head CI checks are green, including React, React 18, Chrome, Firefox, Safari, type checking, builds, and the [performance workflow](https://github.com/ariakit/ariakit/actions/runs/29495432861).

## Candidate disposition

- Prior modal-subscription finding: **already resolved** on the current head and covered by a focused regression.
- Adding `mounted` to the key gate: **invalid as an alternative for this change**, because it would change the subscription on every open/close lifecycle and restore the churn this optimization avoids.
- Further gating for caller-supplied `preserveTabOrderAnchor`: **declined non-goal** for this PR. It is a separable edge optimization, is not introduced by this change, and is not required to resolve the validated modal/non-portal gap.
- Registered follow-ups: none.

## Validation limits

- I relied on the green exact-head browser jobs rather than rerunning Playwright locally.
- I verified the exact-head performance workflow and published report, but did not download and independently inspect its raw artifacts again for this incremental review.
- I ran the focused React suites rather than the entire repository test matrix locally.

app/src/sandbox/popover-disclosure-render/test-browser.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    await q.button(`Set ${label} disclosure element`).click();
+
+    await test
+      .expect(disclosureElementRenders)
+      .not.toHaveText(previousDisclosureElementRenders);
+    await test.expect(popoverRenders).toHaveText(previousPopoverRenders);
+  }
+
+  test("does not re-render a modal popover when the disclosure element changes", async ({
```
diegohaz: Should not we do `test.each` (or playwright's equivalent)?

app/src/sandbox/popover-disclosure-render/test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  await click(q.button(`Set ${label} disclosure element`));
+
+  expect(disclosureElementRenders.textContent).not.toBe(
+    previousDisclosureElementRenders,
+  );
+  expect(popoverRenders.textContent).toBe(previousPopoverRenders);
+}
+
+test("does not re-render a modal popover when the disclosure element changes", async () => {
```
diegohaz: Same here to avoid the `expectNoRenderOnDisclosureElementChange` function

app/src/sandbox/popover-disclosure-render/test-browser.ts
diegohaz: Fixed in [`6efccb045`](https://github.com/ariakit/ariakit/commit/6efccb0458cad38690708fc12c0242a2bdb69837): replaced the three wrapper tests with Playwright’s table-driven `for...of` equivalent and kept the assertion body inline.

app/src/sandbox/popover-disclosure-render/test.ts
diegohaz: Fixed in [`6efccb045`](https://github.com/ariakit/ariakit/commit/6efccb0458cad38690708fc12c0242a2bdb69837): replaced the helper and three wrappers with a single Vitest `test.each` table whose assertion body stays inline.

diegohaz:
**Fixed**

- [`6efccb045`](https://github.com/ariakit/ariakit/commit/6efccb0458cad38690708fc12c0242a2bdb69837) ([Playwright thread](https://github.com/ariakit/ariakit/pull/6751#discussion_r3603993964)): Replaced the wrapper tests with Playwright’s table-driven `for...of` equivalent.
- [`6efccb045`](https://github.com/ariakit/ariakit/commit/6efccb0458cad38690708fc12c0242a2bdb69837) ([Vitest thread](https://github.com/ariakit/ariakit/pull/6751#discussion_r3603997507)): Replaced the helper and wrappers with one `test.each` table.

PR #6795, opened 2026-07-21
Support multi-select Combobox form values

### LINKED ISSUE #5058: [Combobox] improve form interoperability
### Motivation

This is a follow-up from discord: https://discord.com/channels/1113664917921349732/1113664919238344797/1432258077272309840

I'm trying to use the Combobox with multiple selection, inside a form, but it does not work like the Select, where an hidden select multiple is spawned behind the scenes.

I think its not uncommon to use the Combobox pattern inside a form, with multiple selection also, so personally I expect it to work similar to the Select case.

### Usage example

```tsx
<form
  action={(formData) => {
    console.log(formData.get('fruits')) // this should return a string array
  }}
>
  <A.ComboboxProvider defaultSelectedValue={[]}>
    <A.Combobox name="fruits" />
    <A.ComboboxPopover>
      <A.ComboboxItem value="apple">Apple</A.ComboboxItem>
      <A.ComboboxItem value="banana">Banana</A.ComboboxItem>
      <A.ComboboxItem value="orange">Orange</A.ComboboxItem>
      <A.ComboboxItem value="grape">Grape</A.ComboboxItem>
    </A.ComboboxPopover>
  </A.ComboboxProvider>
  <button type="submit">Submit</button>
</form>
```

### Requirements

Passing the `name` attribute on the `<Combobox>`: if the `defaultSelectedValue` or `selectedValue` its an array the array should be mapped to the form state also

### Workaround

In the meantime we can use a controlled state or create an hidden input

### Possible implementations

I'm not sure semantically, but maybe we can replicate the Select solution?

diegohaz:
## Motivation

A multi-select [`Combobox`](https://ariakit.com/reference/combobox) currently submits its transient search text when `name` is provided, rather than the array in `selectedValue`.

```tsx
<ComboboxProvider defaultSelectedValue={[]}>
  <Combobox name="fruits" />
</ComboboxProvider>
```

As a result, native `FormData.getAll("fruits")` cannot retrieve the selected values.

Fixes #5058.

## Solution

When `selectedValue` is an array and the `Combobox` has a `name`, remove the name from the visible query input and render one hidden input for each selected value:

```html
<input role="combobox" />
<input type="hidden" name="fruits" value="apple" />
<input type="hidden" name="fruits" value="orange" />
```

The hidden inputs receive the same `name`, `form`, and logical disabled state, including directly supplied `aria-disabled`. The visible input remains associated with its form and keeps its native query behavior, including constraint validation, invalid events, reset behavior, and implicit submission.

For composite Comboboxes, the hidden inputs mount after the base element is registered to preserve `ComboboxLabel` metadata in Firefox. `composite={false}` bypasses that gate and renders the hidden inputs immediately.

Single-value Combobox behavior is unchanged.

## Scope

This pull request only serializes array selections through native form controls. It does not add `ComboboxSelect`, selection-level constraint validation, selected-value autofill or native mutation synchronization, or pre-hydration serialization for the default composite path.

## Workaround

Before this fix, control the selected values, leave the search input unnamed, and mirror each selection into a hidden input:

```tsx
const [selectedValues, setSelectedValues] = useState<string[]>([]);

<ComboboxProvider
  selectedValue={selectedValues}
  setSelectedValue={setSelectedValues}
>
  <Combobox />
  {/* ... */}
</ComboboxProvider>

{selectedValues.map((value, index) => (
  <input key={index} type="hidden" name="fruits" value={value} />
))}
```

This workaround mirrors the currently selected values for form submission only.

## Testing

Regression coverage verifies controlled and default selected arrays, selected and empty arrays, exclusion of typed query text and `aria-disabled` selections, out-of-tree explicit form ownership with `composite={false}`, unchanged single-value naming, native implicit submission and reset behavior, and Firefox label metadata after hydration.

- `pnpm lint-fix`
- `pnpm exec tsc -p packages/ariakit-react-components/tsconfig.json --noEmit`
- `pnpm tsc`
- `pnpm test-react`
- `pnpm test-react18`
- `pnpm build`
- `pnpm -F app build-lite`
- `APP_PORT=54541 NEXTJS_PORT=54542 CI=true pnpm -F app test --project=chrome --project=firefox --project=safari sandbox/combobox-5058/test-browser.ts`

<details>
<summary>Cycle 3: Continue the native form proxy design</summary>

**Assessment**

Review found three medium-severity gaps in the proxy around supported public behavior: `composite={false}` prevented it from mounting, required validation did not reach the Combobox `onInvalid` handler, and native form resets could desynchronize it from the store. Each failure is deterministic, but each is confined to the proxy lifecycle and can be corrected locally without expanding the public API or changing the overall design.

**Decision**

Continue with the native `<select multiple>` proxy because it directly supplies the repeated native form values requested in #5058. Mount it immediately when composite behavior is disabled, relay invalid events through the visible input, and preserve its mirrored defaults across native resets. `ComboboxSelect`, native option discovery for unselected values, and broader autofill semantics remain outside this issue.

</details>

<details>
<summary>Cycle 4: Cover delayed proxy mounts</summary>

**Assessment**

The staged review found that changing `name` or `composite` could mount the proxy without rerunning the layout effect that initializes its native reset defaults. This is a deterministic but narrow lifecycle gap in supported dynamic props; it does not indicate that the proxy abstraction or public behavior needs to change.

**Decision**

Keep the current design and include `name` and `composite` in the reset-synchronization dependencies. The follow-up staged review found no remaining correctness or regression issues.

</details>

<details>
<summary>Cycle 6: Narrow to serialization-only hidden inputs</summary>

**Assessment**

The core issue is deterministic and user-visible whenever a named Combobox has an array selection: the transient query is submitted instead of the selected values. The supported scope needed to fix it is selected-value serialization in an ordinary hydrated form.

The hidden `<select multiple>` direction expanded that narrow requirement into a form-control subsystem. It required bidirectional synchronization, reset defaults, validity and invalid-event bridging, explicit form ownership, implicit-submission handling, and label-registration policy. Repeated review findings showed that this complexity increased maintenance risk without being required by #5058.

**Decision**

Replace the hidden select with repeated hidden inputs derived directly from `selectedValue`, and leave the visible query input's native form behavior unchanged apart from removing its name in array-selection mode. This directly fixes `FormData.getAll(name)` while eliminating the mutable proxy state and validation bridge.

The Firefox label-registration gate remains because inserting sibling controls before `ComboboxLabel` registration loses input label metadata there. Pre-hydration serialization for the default composite path, selection-level constraint validation, selected-value autofill or native mutation synchronization, and `ComboboxSelect` are intentionally outside this pull request.

</details>

### DIFF
--- a/app/src/sandbox/combobox-5058/test-browser.ts
+++ b/app/src/sandbox/combobox-5058/test-browser.ts
@@ -1,0 +1,22 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/5058
+  test("submits every selected value", async ({ q }) => {
+    const combobox = q.combobox("Favorite fruits");
+    await combobox.click();
+    await q.option("Apple").click();
+    await q.option("Orange").click();
+    await combobox.fill("typed search");
+    await q.button("Submit").click();
+
+    await test.expect(q.status()).toHaveText('["apple","orange"]');
+  });
+
+  // https://github.com/ariakit/ariakit/issues/5058
+  test("does not submit an empty search value", async ({ q }) => {
+    await q.button("Submit").click();
+
+    await test.expect(q.status()).toHaveText("[]");
+  });
+});
--- a/app/src/sandbox/combobox-5058/test-browser.ts
+++ b/app/src/sandbox/combobox-5058/test-browser.ts
@@ -37,33 +37,23 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
-  test("submits selected values when composite is false", async ({ page }) => {
-    const form = page.locator("form[data-composite-false]");
-    const input = page.getByRole("combobox", { name: "Non-composite fruits" });
-    const hiddenInput = page.locator(
-      "input[type='hidden'][name='non-composite-fruits']",
-    );
+  test("submits selected values when composite is false", async ({ q }) => {
+    const form = q.form("Non-composite fruits");
+    const input = q.combobox("Non-composite fruits");
     const values = await form.evaluate((element) =>
       new FormData(element as HTMLFormElement).getAll("non-composite-fruits"),
     );
     const inputForm = await input.evaluate(
       (element) => (element as HTMLInputElement).form?.id ?? null,
     );
-    const hiddenInputForm = await hiddenInput.evaluate(
-      (element) => (element as HTMLInputElement).form?.id ?? null,
-    );
 
     test.expect(values).toEqual(["apple"]);
     test.expect(inputForm).toBe("non-composite-form");
-    test.expect(hiddenInputForm).toBe("non-composite-form");
   });
 
   // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623937766
-  test("preserves implicit submission with an explicit form", async ({
-    page,
-    q,
-  }) => {
-    const form = page.locator("form[data-composite-false]");
+  test("preserves implicit submission with an explicit form", async ({ q }) => {
+    const form = q.form("Non-composite fruits");
     await form.evaluate((element) => {
       element.addEventListener(
         "submit",
--- a/app/src/sandbox/combobox-5058/test-browser.ts
+++ b/app/src/sandbox/combobox-5058/test-browser.ts
@@ -19,4 +19,32 @@ withFramework(import.meta.dirname, async ({ test }) => {
 
     await test.expect(q.status()).toHaveText("[]");
   });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+  test("submits selected values when composite is false", async ({ page }) => {
+    const values = await page
+      .locator("form[data-composite-false]")
+      .evaluate((form) =>
+        new FormData(form as HTMLFormElement).getAll("non-composite-fruits"),
+      );
+
+    test.expect(values).toEqual(["apple"]);
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740398
+  test("relays invalid events to the Combobox input", async ({ q }) => {
+    await q.button("Validate required fruits").click();
+
+    await test.expect(q.text("Invalid target: input")).toBeVisible();
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740415
+  test("keeps selected values in sync after form reset", async ({ q }) => {
+    await q.combobox("Favorite fruits").click();
+    await q.option("Apple").click();
+    await q.button("Reset").click();
+    await q.button("Submit").click();
+
+    await test.expect(q.status()).toHaveText('["apple"]');
+  });
 });
--- a/app/src/sandbox/combobox-5058/test-browser.ts
+++ b/app/src/sandbox/combobox-5058/test-browser.ts
@@ -8,6 +8,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await q.option("Apple").click();
     await q.option("Orange").click();
     await combobox.fill("typed search");
+    await combobox.press("Escape");
     await q.button("Submit").click();
 
     await test.expect(q.status()).toHaveText('["apple","orange"]');
@@ -21,27 +22,76 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
-  test("submits selected values when composite is false", async ({ page }) => {
-    const values = await page
-      .locator("form[data-composite-false]")
-      .evaluate((form) =>
-        new FormData(form as HTMLFormElement).getAll("non-composite-fruits"),
-      );
+  test("preserves the input label metadata after hydration", async ({ q }) => {
+    await q.combobox("Favorite fruits").click();
+    await q.option("Apple").click();
 
-    test.expect(values).toEqual(["apple"]);
+    const labels = await q.combobox("Favorite fruits").evaluate((element) => {
+      const input = element as HTMLInputElement;
+      return Array.from(input.labels ?? [], (label) =>
+        label.textContent?.trim(),
+      );
+    });
+
+    test.expect(labels).toEqual(["Favorite fruits"]);
   });
 
-  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740398
-  test("relays invalid events to the Combobox input", async ({ q }) => {
-    await q.button("Validate required fruits").click();
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+  test("submits selected values when composite is false", async ({ page }) => {
+    const form = page.locator("form[data-composite-false]");
+    const input = page.getByRole("combobox", { name: "Non-composite fruits" });
+    const hiddenInput = page.locator(
+      "input[type='hidden'][name='non-composite-fruits']",
+    );
+    const values = await form.evaluate((element) =>
+      new FormData(element as HTMLFormElement).getAll("non-composite-fruits"),
+    );
+    const inputForm = await input.evaluate(
+      (element) => (element as HTMLInputElement).form?.id ?? null,
+    );
+    const hiddenInputForm = await hiddenInput.evaluate(
+      (element) => (element as HTMLInputElement).form?.id ?? null,
+    );
 
-    await test.expect(q.text("Invalid target: input")).toBeVisible();
+    test.expect(values).toEqual(["apple"]);
+    test.expect(inputForm).toBe("non-composite-form");
+    test.expect(hiddenInputForm).toBe("non-composite-form");
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623937766
+  test("preserves implicit submission with an explicit form", async ({
+    page,
+    q,
+  }) => {
+    const form = page.locator("form[data-composite-false]");
+    await form.evaluate((element) => {
+      element.addEventListener(
+        "submit",
+        (event) => {
+          event.preventDefault();
+          const submitter = (event as SubmitEvent).submitter;
+          element.setAttribute(
+            "data-submitter",
+            submitter?.textContent?.trim() ?? "",
+          );
+        },
+        { once: true },
+      );
+    });
+
+    await q.combobox("Non-composite fruits").press("Enter");
+
+    await test
+      .expect(form)
+      .toHaveAttribute("data-submitter", "Submit non-composite fruits");
   });
 
   // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740415
   test("keeps selected values in sync after form reset", async ({ q }) => {
-    await q.combobox("Favorite fruits").click();
+    const combobox = q.combobox("Favorite fruits");
+    await combobox.click();
     await q.option("Apple").click();
+    await combobox.press("Escape");
     await q.button("Reset").click();
     await q.button("Submit").click();
 
--- a/app/src/sandbox/combobox-5058/test-browser.ts
+++ b/app/src/sandbox/combobox-5058/test-browser.ts
@@ -51,6 +51,18 @@ withFramework(import.meta.dirname, async ({ test }) => {
     test.expect(inputForm).toBe("non-composite-form");
   });
 
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3625787623
+  test("omits aria-disabled selected values", async ({ q }) => {
+    const form = q.form("Disabled fruits");
+    await test.expect(q.combobox("Disabled fruits")).toBeDisabled();
+
+    const values = await form.evaluate((element) =>
+      new FormData(element as HTMLFormElement).getAll("disabled-fruits"),
+    );
+
+    test.expect(values).toEqual([]);
+  });
+
   // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623937766
   test("preserves implicit submission with an explicit form", async ({ q }) => {
     const form = q.form("Non-composite fruits");
--- a/app/src/sandbox/combobox-5058/test.ts
+++ b/app/src/sandbox/combobox-5058/test.ts
@@ -9,14 +9,9 @@ test("mirrors every selected value in the form control", async () => {
 
   expect(q.combobox("Favorite fruits")).not.toHaveAttribute("name");
 
-  const select = document.querySelector<HTMLSelectElement>(
-    "select[name='fruits'][multiple]",
-  );
-  const selectedValues = Array.from(
-    select?.selectedOptions ?? [],
-    (option) => option.value,
-  );
-  expect(selectedValues).toEqual(["apple", "orange"]);
+  const input = q.combobox("Favorite fruits") as HTMLInputElement;
+  const form = input.form!;
+  expect(new FormData(form).getAll("fruits")).toEqual(["apple", "orange"]);
 });
 
 // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
@@ -25,13 +20,16 @@ test("mirrors selected values when composite is false", () => {
     "form[data-composite-false]",
   );
   const formData = new FormData(form!);
+  const input = q.combobox("Non-composite fruits") as HTMLInputElement;
+  const hiddenInput = document.querySelector<HTMLInputElement>(
+    "input[type='hidden'][name='non-composite-fruits']",
+  );
 
   expect(formData.getAll("non-composite-fruits")).toEqual(["apple"]);
+  expect(input.form).toBe(form);
+  expect(hiddenInput?.form).toBe(form);
 });
 
-// https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740398
-test("relays invalid events to the Combobox input", async () => {
-  await click(q.button("Validate required fruits"));
-
-  expect(q.text("Invalid target: input")).toBeInTheDocument();
+test("preserves the name for a single selected value", () => {
+  expect(q.combobox("Single fruit")).toHaveAttribute("name", "single-fruit");
 });
--- a/app/src/sandbox/combobox-5058/test.ts
+++ b/app/src/sandbox/combobox-5058/test.ts
@@ -18,3 +18,20 @@ test("mirrors every selected value in the form control", async () => {
   );
   expect(selectedValues).toEqual(["apple", "orange"]);
 });
+
+// https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+test("mirrors selected values when composite is false", () => {
+  const form = document.querySelector<HTMLFormElement>(
+    "form[data-composite-false]",
+  );
+  const formData = new FormData(form!);
+
+  expect(formData.getAll("non-composite-fruits")).toEqual(["apple"]);
+});
+
+// https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740398
+test("relays invalid events to the Combobox input", async () => {
+  await click(q.button("Validate required fruits"));
+
+  expect(q.text("Invalid target: input")).toBeInTheDocument();
+});
--- a/app/src/sandbox/combobox-5058/test.ts
+++ b/app/src/sandbox/combobox-5058/test.ts
@@ -24,6 +24,14 @@ test("mirrors selected values when composite is false", () => {
   expect(input.form).toBe(form);
 });
 
+// https://github.com/ariakit/ariakit/pull/6795#discussion_r3625787623
+test("omits aria-disabled selected values", () => {
+  const form = q.form.ensure("Disabled fruits") as HTMLFormElement;
+
+  expect(q.combobox("Disabled fruits")).toBeDisabled();
+  expect(new FormData(form).getAll("disabled-fruits")).toEqual([]);
+});
+
 test("preserves the name for a single selected value", () => {
   expect(q.combobox("Single fruit")).toHaveAttribute("name", "single-fruit");
 });
--- a/app/src/sandbox/combobox-5058/test.ts
+++ b/app/src/sandbox/combobox-5058/test.ts
@@ -1,0 +1,20 @@
+import { click, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+// https://github.com/ariakit/ariakit/issues/5058
+test("mirrors every selected value in the form control", async () => {
+  await click(q.combobox("Favorite fruits"));
+  await click(q.option("Apple"));
+  await click(q.option("Orange"));
+
+  expect(q.combobox("Favorite fruits")).not.toHaveAttribute("name");
+
+  const select = document.querySelector<HTMLSelectElement>(
+    "select[name='fruits'][multiple]",
+  );
+  const selectedValues = Array.from(
+    select?.selectedOptions ?? [],
+    (option) => option.value,
+  );
+  expect(selectedValues).toEqual(["apple", "orange"]);
+});
--- a/app/src/sandbox/combobox-5058/test.ts
+++ b/app/src/sandbox/combobox-5058/test.ts
@@ -16,18 +16,12 @@ test("mirrors every selected value in the form control", async () => {
 
 // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
 test("mirrors selected values when composite is false", () => {
-  const form = document.querySelector<HTMLFormElement>(
-    "form[data-composite-false]",
-  );
-  const formData = new FormData(form!);
+  const form = q.form("Non-composite fruits") as HTMLFormElement;
+  const formData = new FormData(form);
   const input = q.combobox("Non-composite fruits") as HTMLInputElement;
-  const hiddenInput = document.querySelector<HTMLInputElement>(
-    "input[type='hidden'][name='non-composite-fruits']",
-  );
 
   expect(formData.getAll("non-composite-fruits")).toEqual(["apple"]);
   expect(input.form).toBe(form);
-  expect(hiddenInput?.form).toBe(form);
 });
 
 test("preserves the name for a single selected value", () => {
--- a/packages/ariakit-react-components/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox.tsx
@@ -19,6 +19,7 @@ import {
   getPopupRole,
   getScrollingElement,
   getTextboxSelection,
+  fireEvent,
   setSelectionRange,
   isFocusEventOutside,
   queueBeforeEvent,
@@ -703,6 +704,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     const form = props.form;
     const required = props.required;
     const disabled = props.disabled;
+    const composite = props.composite !== false;
     const label = props["aria-label"];
     const labelledBy = props["aria-labelledby"];
     const baseElement = useStoreState(
@@ -710,6 +712,25 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       multiSelectable ? ["baseElement"] : [],
       (state) => (multiSelectable ? state.baseElement : null),
     );
+    const selectRef = useRef<HTMLSelectElement>(null);
+
+    useSafeLayoutEffect(() => {
+      const select = selectRef.current;
+      if (!select) return;
+      // Keep native form resets from clearing the mirrored selection while
+      // the Combobox store still holds the same values.
+      for (const option of select.options) {
+        option.defaultSelected = option.selected;
+      }
+    }, [selectedValue, baseElement, name, composite]);
+
+    const onInvalid = useEvent((event: SyntheticEvent<HTMLSelectElement>) => {
+      const element = ref.current;
+      if (!element) return;
+      event.stopPropagation();
+      const defaultAllowed = fireEvent(element, "invalid", event);
+      if (!defaultAllowed) event.preventDefault();
+    });
 
     props = useWrapElement(
       props,
@@ -718,11 +739,12 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
         if (!Array.isArray(selectedValue)) return element;
         // Wait until ComboboxLabel has associated itself with the input before
         // inserting another labelable element.
-        if (!baseElement) return element;
+        if (composite && !baseElement) return element;
         return (
           <>
             {element}
             <select
+              ref={selectRef}
               style={getVisuallyHiddenStyle()}
               tabIndex={-1}
               aria-hidden
@@ -735,6 +757,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
               multiple
               value={selectedValue}
               onFocus={() => ref.current?.focus()}
+              onInvalid={onInvalid}
               onChange={(event) => {
                 store?.setSelectedValue(getSelectedValues(event.target));
               }}
@@ -758,6 +781,8 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
         labelledBy,
         selectedValue,
         baseElement,
+        composite,
+        onInvalid,
       ],
     );
 
--- a/packages/ariakit-react-components/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox.tsx
@@ -16,6 +16,7 @@ import {
 import type { Props } from "@ariakit/react-utils";
 import { sync } from "@ariakit/store";
 import {
+  disabledFromProps,
   getPopupRole,
   getScrollingElement,
   getTextboxSelection,
@@ -697,6 +698,11 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       (state) => state.activeId === null,
     );
 
+    const formDisabled = disabledFromProps({
+      disabled,
+      "aria-disabled": props["aria-disabled"],
+    });
+
     const composite = props.composite !== false;
     const baseElement = useStoreState(
       store,
@@ -719,14 +725,14 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
                 type="hidden"
                 name={name}
                 form={form}
-                disabled={disabled}
+                disabled={formDisabled}
                 value={value}
               />
             ))}
           </>
         );
       },
-      [name, form, disabled, composite, baseElement, selectedValue],
+      [name, form, formDisabled, composite, baseElement, selectedValue],
     );
 
     const htmlProps = {
--- a/packages/ariakit-react-components/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox.tsx
@@ -8,6 +8,7 @@ import {
   useSafeLayoutEffect,
   useUpdateEffect,
   useUpdateLayoutEffect,
+  useWrapElement,
   createElement,
   createHook,
   forwardRef,
@@ -44,6 +45,7 @@ import type { CompositeOptions } from "../composite/composite.tsx";
 import { useComposite } from "../composite/composite.tsx";
 import type { PopoverAnchorOptions } from "../popover/popover-anchor.tsx";
 import { usePopoverAnchor } from "../popover/popover-anchor.tsx";
+import { getVisuallyHiddenStyle } from "../visually-hidden/visually-hidden.tsx";
 import { useComboboxProviderContext } from "./combobox-context.tsx";
 import type {
   ComboboxStore,
@@ -54,6 +56,10 @@ import type {
 const TagName = "input" satisfies ElementType;
 type TagName = typeof TagName;
 type HTMLType = HTMLElementTagNameMap[TagName];
+
+function getSelectedValues(select: HTMLSelectElement) {
+  return Array.from(select.selectedOptions).map((option) => option.value);
+}
 
 function isFirstItemAutoSelected(
   items: ComboboxStoreState["items"],
@@ -179,6 +185,13 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     }, [inline]);
 
     const storeValue = useStoreState(store, "value");
+    const name = props.name;
+    const selectedValue = useStoreState(store, ["selectedValue"], (state) => {
+      if (!name) return;
+      if (!Array.isArray(state.selectedValue)) return;
+      return state.selectedValue;
+    });
+    const multiSelectable = Array.isArray(selectedValue);
 
     // Keep track of the previous selected values so we can set the
     // inlineActiveValue below only when the current activeValue isn't already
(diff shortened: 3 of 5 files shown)


packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -697,6 +771,10 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       value,
       ...props,
       id,
+      "aria-required":
+        multiSelectable && required ? true : props["aria-required"],
+      name: multiSelectable ? undefined : name,
```
diegohaz, line 748: The visible search input still participates in native constraint validation after its `name` is moved to the proxy. Removing `name` does not remove a control from `form.checkValidity()`, so constraints on transient search text can block submission even when every selected value is valid.

```tsx
<ComboboxProvider
  defaultValue="partial@"
  defaultSelectedValue={["alice@example.com"]}
>
  <Combobox name="recipients" type="email" />
</ComboboxProvider>
```

I reproduced this on the current head in Chromium, Firefox, and WebKit: `FormData` contained `alice@example.com` and the proxy select was valid, but the unnamed input had `typeMismatch`, `form.checkValidity()` was `false`, and `requestSubmit()` never submitted. An email-recipient picker therefore cannot send valid committed recipients while the user has a partial query that is intentionally excluded from submission.

Please disassociate the visible input from the form in proxy mode (including when `form` is explicit), or otherwise neutralize its participation in form validation, while leaving the proxy associated. A browser regression using `type="email"` or `pattern` would cover the behavior.

**Assessment**

- Severity: High (estimate, 95% confidence), because the reproduced configuration blocks submission completely.
- Priority: P1 (estimate, 92% confidence), because moving the form-facing value to the proxy should also move form-facing validity before merge.
- Breakage likelihood: Possible (estimate, 85% confidence), especially for email, URL, numeric, or patterned multi-value pickers.
- Fix effort: Moderate (estimate, 75% confidence), because the search input needs a deliberate form-owner/validation contract plus browser coverage.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,6 +700,67 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+    );
+
+    props = useWrapElement(
+      props,
+      (element) => {
+        if (!name) return element;
+        if (!Array.isArray(selectedValue)) return element;
+        // Wait until ComboboxLabel has associated itself with the input before
+        // inserting another labelable element.
+        if (!baseElement) return element;
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and the same guard also removes every successful control from server markup. With `composite={false}`, Chromium, Firefox, and WebKit all rendered no proxy, stripped the input `name`, and produced empty `FormData` for `defaultSelectedValue={["apple"]}`. The raw SSR response likewise contains neither `<input name="fruits">` nor `<select name="fruits">`, so a pre-hydration submission loses default selections.

I also tested the simpler alternative of rendering this same `<select>` immediately after the input. All three engines hydrated without errors; in Firefox, `label.control` remained the input, the select had no labels, and clicking the label focused the input. That evidence suggests the `baseElement` delay is unnecessary in the current input-then-select order. Removing the guard (and its subscription/dependency) looks preferable to adding fallback markup, and it fixes both `composite={false}` and SSR.

Please add a regression for the documented non-composite mode and, if practical, a server-render assertion with a nonempty default selection.

**Assessment**

- Severity: High (verified, 98% confidence), because the field was completely absent from `FormData` in all three engines.
- Priority: P1 (estimate, 98% confidence), because a supported public prop combination defeats this PR's core contract.
- Breakage likelihood: Unlikely (estimate, 85% confidence), because non-composite mode and pre-hydration submission are specialized paths.
- Fix effort: Easy (estimate, 90% confidence), because removing the guard passed the focused cross-engine label/hydration probe; regressions remain to be added.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,6 +700,67 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+            {element}
+            <select
+              style={getVisuallyHiddenStyle()}
+              tabIndex={-1}
+              aria-hidden
+              aria-label={label}
+              aria-labelledby={label != null ? undefined : labelledBy}
+              name={name}
+              form={form}
+              required={required}
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and forwarding only the callback would still leave the public validity surface split. In Chromium, Firefox, and WebKit, an empty required multi-select made the hidden select and form invalid while the forwarded input ref reported `checkValidity() === true`; the input did not match `:invalid`, and the supplied `onInvalid` never ran.

```tsx
const ref = useRef<HTMLInputElement>(null);
<Combobox name="categories" required ref={ref} onInvalid={showError} />;
```

A form library or wizard using the Combobox ref, `onInvalid`, or `:invalid` styling therefore treats the field as valid even though native submission is blocked. Please define and preserve a coherent proxy-to-input validation contract, including cancellation/current-target semantics, and cover it in a real-browser regression.

**Assessment**

- Severity: Medium (estimate, 90% confidence), because native submission remains blocked but component-level error handling and styling silently fail.
- Priority: P1 (estimate, 90% confidence), because transferring `required` in this PR should preserve the corresponding public validation behavior.
- Breakage likelihood: Possible (estimate, 85% confidence), requiring a required multi-select plus component-level validation handling.
- Fix effort: Moderate (estimate, 75% confidence), because events, ref validity, and visible invalid state need an intentional bridge.

diegohaz (commented):
### Initial review

Reviewed [`5e6afd3`](https://github.com/ariakit/ariakit/commit/5e6afd321b7e34cf9755b46a227ab48c7ed80c4d) → [`beea958`](https://github.com/ariakit/ariakit/commit/beea9589173ec37896b51ddb4a0b0b6292bcc654), including the Combobox implementation and surrounding Composite/form behavior, the new sandbox tests, browser behavior, SSR/hydration, and the changeset.

I found four P1 issues that should be addressed before merge:

1. **[P1] The transient search input still participates in form constraint validation** — [inline finding](https://github.com/ariakit/ariakit/pull/6795#discussion_r3623937766). Removing `name` does not remove the visible input from constraint validation. For example, `<Combobox name="recipients" type="email" />` can have valid selected proxy values while transient search text makes the form invalid and prevents submission. Reproduced in Chromium, Firefox, and WebKit. **Assessment:** impact High (estimated, 95% confidence); priority P1 (92%); real-world likelihood Possible (85%); fix effort Moderate (75%). **Disposition:** Address now.
2. **[P1] `composite={false}` and SSR omit the field entirely** — [thread with added evidence](https://github.com/ariakit/ariakit/pull/6795#discussion_r3623939366). The proxy is gated on `baseElement`, but this value is never registered when Composite behavior is disabled, while the input's `name` is still removed. All three engines produced empty `FormData`; raw SSR output also had no named control. Rendering the proxy immediately passed targeted hydration and Firefox label-control probes, so removing this gate appears to be the simplest correction. **Assessment:** impact High (verified, 98% confidence); priority P1 (98%); real-world likelihood Unlikely (85% confidence); fix effort Easy (90%). **Disposition:** Address now.
3. **[P1] Required validation bypasses the public input validation surface** — [thread with added evidence](https://github.com/ariakit/ariakit/pull/6795#discussion_r3623940015). With `required` moved to the hidden proxy, the forwarded input ref reports valid, `:invalid` does not match, and `onInvalid`/`onInvalidCapture` on Combobox never fire, even though the form is invalid. Reproduced in Chromium, Firefox, and WebKit. **Assessment:** impact Medium (estimated, 90% confidence); priority P1 (90%); real-world likelihood Possible (85%); fix effort Moderate (75%). **Disposition:** Address now.
4. **[P1] Native form reset desynchronizes the proxy from the store** — [existing thread](https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740415). `form.reset()` clears the generated options without notifying the store, so the visible selected state remains while immediate submission serializes no values; this also breaks `defaultSelectedValue`. Reproduced in Chromium, Firefox, and WebKit. **Assessment:** impact High (estimated, 95% confidence); priority P1 (95%); real-world likelihood Possible (90%); fix effort Moderate (85%). **Disposition:** Address now.

### Validation

- `pnpm test-react app/src/sandbox/combobox-5058/test.ts` — passed
- `pnpm test-react18 app/src/sandbox/combobox-5058/test.ts` — passed
- `pnpm exec tsc -p packages/ariakit-react-components/tsconfig.json --noEmit` — passed
- `git diff --check` — passed
- Targeted Chromium, Firefox, and WebKit probes covered submission, reset, validation, `composite={false}`, SSR/hydration, and Firefox label association.
- The regression test failed as expected on the [test-only commit](https://github.com/ariakit/ariakit/commit/7e02d3185fe959c9e462bbfa8e8696685a64b251) before the implementation was applied.
- Exact-head CI was green at the final refresh. I did not rerun the repository's complete browser suite locally; the assessment uses the exact-head CI runs plus the targeted cross-engine probes above.

The sandbox test structure and changeset are appropriate. I did not identify additional release-note, security, performance, or duplicate-value issues. No follow-up work is being deferred from this review.

packages/ariakit-react-components/src/combobox/combobox.tsx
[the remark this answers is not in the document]
diegohaz: Superseded by [`a98eb6c09`](https://github.com/ariakit/ariakit/commit/a98eb6c091b6819afbfa7b16777cc60fb24241e8): the redesign no longer moves `required` or validation to a proxy. The visible Combobox input keeps its native validity state and `onInvalid` surface; repeated hidden inputs mirror only selected values.

packages/ariakit-react-components/src/combobox/combobox.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`a98eb6c09`](https://github.com/ariakit/ariakit/commit/a98eb6c091b6819afbfa7b16777cc60fb24241e8): `composite={false}` now renders the repeated hidden inputs immediately, and coverage verifies out-of-tree explicit form ownership. The label-registration gate remains for normal composite mode because removing it regresses Firefox input label metadata. Supporting pre-hydration serialization for that path would require broader label/provider architecture and is outside this narrow fix.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,6 +700,67 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+              disabled={disabled}
+              multiple
+              value={selectedValue}
```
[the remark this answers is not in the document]
diegohaz: Superseded by [`a98eb6c09`](https://github.com/ariakit/ariakit/commit/a98eb6c091b6819afbfa7b16777cc60fb24241e8): the redesign removes the mutable select. Repeated hidden inputs are derived directly from `selectedValue`, so native form reset cannot clear a separate proxy selection; targeted coverage verifies the values still serialize after reset.

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: Not changing this in the current PR because the visible Combobox remains the query input and keeps its existing native form and constraint behavior; [`a98eb6c09`](https://github.com/ariakit/ariakit/commit/a98eb6c091b6819afbfa7b16777cc60fb24241e8) only redirects array-selection serialization from transient text to repeated selected values. Detaching the input would alter validation and implicit-submission semantics and require a broader form contract outside #5058 and the deferred `ComboboxSelect` work.

diegohaz:
Complexity reassessment is implemented in [`a98eb6c09`](https://github.com/ariakit/ariakit/commit/a98eb6c091b6819afbfa7b16777cc60fb24241e8).

- Replaced the mutable hidden `<select>` and its validation/reset synchronization with repeated hidden inputs derived directly from `selectedValue`.
- Kept the visible Combobox associated with its native form; only its `name` moves to the hidden inputs for array selections.
- Strengthened coverage for controlled and default arrays, out-of-tree explicit form ownership, `composite={false}`, implicit submission, reset behavior, Firefox labels, and single-value compatibility.
- Kept `ComboboxSelect`, selected-value validation/autofill or native mutation synchronization, and default-composite pre-hydration serialization outside this issue.

Local verification is green across React 18/19, the full 903-test React suite, full typecheck, package build, app build, and 18 Chromium/Firefox/WebKit cases.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -697,6 +739,7 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
       value,
       ...props,
       id,
+      name: multiSelectable ? undefined : name,
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because `required` already validates the visible Combobox query on `main`, and this PR is intentionally limited to mapping array `selectedValue` entries into `FormData` when `name` is provided. Making `required` validate the selection would introduce a new selection-control contract, including invalid-state and event semantics, that belongs with the deferred `ComboboxSelect` work rather than issue #5058.

app/src/sandbox/combobox-5058/test-browser.ts
```diff
@@ -0,0 +1,100 @@
...
+      );
+    });
+
+    test.expect(labels).toEqual(["Favorite fruits"]);
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+  test("submits selected values when composite is false", async ({ page }) => {
+    const form = page.locator("form[data-composite-false]");
+    const input = page.getByRole("combobox", { name: "Non-composite fruits" });
```
diegohaz: Use `q.combobox("Non-composite fruits")` (or `/Non-composite fruits/` if name is not exact) instead.

app/src/sandbox/combobox-5058/test-browser.ts
```diff
@@ -0,0 +1,100 @@
...
+        label.textContent?.trim(),
+      );
+    });
+
+    test.expect(labels).toEqual(["Favorite fruits"]);
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+  test("submits selected values when composite is false", async ({ page }) => {
+    const form = page.locator("form[data-composite-false]");
```
diegohaz: Shouldn't the form have an accessible name so we can query it with `q.form()`?

app/src/sandbox/combobox-5058/test-browser.ts
```diff
@@ -0,0 +1,100 @@
...
+    const input = page.getByRole("combobox", { name: "Non-composite fruits" });
+    const hiddenInput = page.locator(
```
diegohaz: Use `q` queries

app/src/sandbox/combobox-5058/test-browser.ts
```diff
@@ -0,0 +1,100 @@
...
+    test.expect(inputForm).toBe("non-composite-form");
+    test.expect(hiddenInputForm).toBe("non-composite-form");
+  });
+
+  // https://github.com/ariakit/ariakit/pull/6795#discussion_r3623937766
+  test("preserves implicit submission with an explicit form", async ({
+    page,
+    q,
+  }) => {
+    const form = page.locator("form[data-composite-false]");
```
diegohaz: Use `q` queries

app/src/sandbox/combobox-5058/test.ts
```diff
@@ -0,0 +1,35 @@
...
+  expect(q.combobox("Favorite fruits")).not.toHaveAttribute("name");
+
+  const input = q.combobox("Favorite fruits") as HTMLInputElement;
+  const form = input.form!;
+  expect(new FormData(form).getAll("fruits")).toEqual(["apple", "orange"]);
+});
+
+// https://github.com/ariakit/ariakit/pull/6795#discussion_r3623740406
+test("mirrors selected values when composite is false", () => {
+  const form = document.querySelector<HTMLFormElement>(
```
diegohaz: Use `q` queries

app/src/sandbox/combobox-5058/test.ts
```diff
@@ -0,0 +1,35 @@
...
+    "form[data-composite-false]",
+  );
+  const formData = new FormData(form!);
+  const input = q.combobox("Non-composite fruits") as HTMLInputElement;
+  const hiddenInput = document.querySelector<HTMLInputElement>(
```
diegohaz: Use `q` queries

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -179,6 +180,13 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     }, [inline]);
 
     const storeValue = useStoreState(store, "value");
+    const name = props.name;
```
diegohaz: What about destructuring `name`, `form`, and `disabled` from the props in the function declaration?

app/src/sandbox/combobox-5058/test-browser.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): the browser test now uses `q.combobox("Non-composite fruits")`.

app/src/sandbox/combobox-5058/test-browser.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): the form now has the accessible name “Non-composite fruits”, and both browser tests query it with `q.form("Non-composite fruits")`.

app/src/sandbox/combobox-5058/test-browser.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): I removed the direct hidden-input locator. `FormData` on the out-of-tree `q.form()` still proves the mirrored value is associated with that form.

app/src/sandbox/combobox-5058/test-browser.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): the implicit-submission test now uses `q.form("Non-composite fruits")`.

app/src/sandbox/combobox-5058/test.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): the happy-dom test now queries the accessible form with `q.form("Non-composite fruits")`.

app/src/sandbox/combobox-5058/test.ts
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): I removed the direct hidden-input query; the `FormData` assertion covers its external form association.

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: Fixed in [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf): `name`, `form`, and `disabled` are now destructured in the hook parameters and explicitly forwarded to the visible input and mirrored inputs.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,7 +697,39 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+        return (
+          <>
+            {element}
+            {selectedValue.map((value, index) => (
+              <input
+                key={index}
+                type="hidden"
+                name={name}
+                form={form}
+                disabled={disabled}
```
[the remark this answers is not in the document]
diegohaz: We don't need to fix those `render` edge cases.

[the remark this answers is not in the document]
diegohaz: Confirmed on `ff54cbb54` in Chromium, Firefox, and WebKit. This is broader than `disabled`: `name` and `form` supplied or overridden by `render` are also invisible to the mirrors. For example:

```tsx
<ComboboxProvider
  defaultValue="typed search"
  defaultSelectedValue={["apple"]}
>
  <Combobox composite={false} render={<input name="fruits" />} />
</ComboboxProvider>
```

This submits the transient query (`"typed search"`) and renders no hidden inputs, because the hook sees no top-level `name`; the `render` name is merged later by `createElement`. With `<Combobox name="fruits" render={<input disabled />} />`, the visible input is disabled but the enabled mirror still submits `"apple"`.

Please derive `name`, `form`, and `disabled` from the effective rendered element at a post-render/pre-consumer-wrapper composition point, and cover both render elements and render callbacks.

**Assessment**

- Severity: High (verified, 99% confidence), because supported composition can submit the wrong query value or submit a disabled field's selections.
- Priority: P1 (estimate, 93% confidence), because these paths violate the PR's core form-value contract.
- Breakage likelihood: Unlikely (estimate, 82% confidence), requiring form metadata to be supplied or overridden through `render`.
- Fix effort: Hard (estimate, 75% confidence), because the mirrors need access to final render props without escaping the consumer wrapper.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,7 +697,39 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+    props = useWrapElement(
+      props,
+      (element) => {
+        if (!name) return element;
+        if (!Array.isArray(selectedValue)) return element;
+        if (composite && !baseElement) return element;
+        return (
+          <>
+            {element}
+            {selectedValue.map((value, index) => (
```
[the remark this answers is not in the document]
diegohaz: Confirmed on `ff54cbb54` in Chromium, Firefox, and WebKit. A portal reproduces the form-owner split directly:

```tsx
<ComboboxProvider defaultSelectedValue={["apple"]}>
  <Combobox
    composite={false}
    name="fruits"
    wrapElement={(element) => createPortal(element, checkoutForm)}
  />
</ComboboxProvider>
```

The visible input's `form` is `checkoutForm`, while the hidden input stays at the source tree with `form === null`, so `new FormData(checkoutForm).getAll("fruits")` is `[]`. The inverse state bug occurs with `wrapElement={(element) => <fieldset disabled>{element}</fieldset>}`: the visible input is disabled, but the mirror remains outside the fieldset, stays enabled, and submits `"apple"`.

Please create the visible input plus mirrors as one internal unit before invoking the consumer's `wrapElement`, so wrappers, portals, and inherited fieldset state apply consistently. Add browser regressions for a portal/form owner and a disabled fieldset wrapper.

**Assessment**

- Severity: High (verified, 99% confidence), because all selected values are either lost or incorrectly submitted in all three engines.
- Priority: P1 (estimate, 95% confidence), because `wrapElement` is a supported public composition surface and the failure defeats the PR's core behavior.
- Breakage likelihood: Unlikely (estimate, 85% confidence), requiring a form-affecting wrapper or portal.
- Fix effort: Moderate (estimate, 75% confidence), because internal and consumer wrapper ordering must be separated deliberately.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -687,7 +697,39 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
+      multiSelectable ? ["baseElement"] : [],
+      (state) => (multiSelectable ? state.baseElement : null),
+    );
+
+    props = useWrapElement(
+      props,
+      (element) => {
+        if (!name) return element;
+        if (!Array.isArray(selectedValue)) return element;
+        if (composite && !baseElement) return element;
```
diegohaz, line 718: The `composite && !baseElement` gate is now unnecessary after replacing the labelable `<select>` proxy with non-labelable `<input type="hidden">` mirrors. It also keeps default selections out of server markup and pre-hydration `FormData` for the normal composite path.

I removed only this guard in an in-memory build and exercised SSR plus hydration in Chromium, Firefox, and WebKit. In every engine, `FormData` contained `"apple"` both before and after hydration; `label.control` and `input.labels` still identified the Combobox, clicking the label focused it, and hydration logged no errors.

Please simplify this to render the mirrors whenever `name` and array `selectedValue` exist:

```tsx
if (!name) return element;
if (!Array.isArray(selectedValue)) return element;
return <>{/* input plus mirrors */}</>;
```

That also lets us remove the `composite` calculation, `baseElement` subscription, and both dependencies. An SSR assertion for a nonempty default selection would preserve the regained behavior.

**Assessment**

- Severity: Low (estimate, 95% confidence), because the behavioral gap is limited to pre-hydration submission, with a small ongoing rerender/subscription cost after hydration.
- Priority: P2 (estimate, 95% confidence), because the gate has no remaining label benefit and is an easy behavior-improving simplification.
- Breakage likelihood: Rare (estimate, 90% confidence), requiring submission before hydration or JavaScript never hydrating.
- Fix effort: Easy (verified, 98% confidence), because removing the guard passed the focused SSR/hydration probe in all three engines.

packages/ariakit-react-components/src/combobox/combobox.tsx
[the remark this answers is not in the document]
diegohaz: The same scope decision applies here: we’re not expanding #5058 to cover `render`/`wrapElement` composition edge cases.

packages/ariakit-react-components/src/combobox/combobox.tsx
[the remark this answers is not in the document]
diegohaz: The maintainer scope decision above still applies after this automated follow-up: `render` metadata overrides are intentionally outside #5058.

diegohaz (commented):
Re-review completed against [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf), using the prior marked review at [`beea95891`](https://github.com/ariakit/ariakit/commit/beea9589173ec37896b51ddb4a0b0b6292bcc654) as the incremental baseline.

I found three actionable issues:

- [P1: mirrors ignore effective `render` props](https://github.com/ariakit/ariakit/pull/6795#discussion_r3624850374). For example, `<Combobox composite={false} render={<input name="fruits" />} />` still submits the transient query, while a render-level `disabled` override leaves enabled mirrors that submit selections.
- [P1: mirrors escape consumer wrappers](https://github.com/ariakit/ariakit/pull/6795#discussion_r3624851044). A `wrapElement` portal can put the visible input in a form while leaving the mirrors outside it; a disabled fieldset wrapper disables only the input and still submits the mirrors.
- [P2: remove the obsolete `baseElement` gate](https://github.com/ariakit/ariakit/pull/6795#discussion_r3624851654). Hidden inputs do not affect label metadata, and removing the gate restores default values to server/pre-hydration form markup while eliminating an unnecessary store subscription.

The two P1 composition failures should be addressed before merge. The shared direction is to build the input and mirrors from the final effective element props, then pass that unit through the consumer wrapper:

```tsx
const control = <>{input}{mirrors}</>;
return wrapElement ? wrapElement(control) : control;
```

Review coverage included the incremental diff and surrounding `useWrapElement`, `createElement`, `mergeProps`, Combobox store subscriptions, sandbox integration/browser tests, public composition types/docs, and the changeset. `ff54cbb54` resolves all six earlier `q`-query comments and the parameter-destructuring comment. The old `<select>` proxy lifecycle findings are obsolete after the repeated-hidden-input redesign. I did not reopen the transient-query constraint or selection-level `required` discussions: the visible query's existing validity semantics are intentionally preserved, while selection-control validity remains outside this PR. No follow-up issue is registered.

Verification:

- `CI=true pnpm test-react app/src/sandbox/combobox-5058/test.ts` — 3/3 passed.
- `CI=true pnpm test-react18 app/src/sandbox/combobox-5058/test.ts` — 3/3 passed.
- `CI=true pnpm exec tsc -p packages/ariakit-react-components/tsconfig.json --noEmit` — passed.
- `git diff --check origin/main...HEAD` — passed.
- Focused current-head Chromium/Firefox/WebKit probe — reproduced all three effective-composition mismatches (`portal: []`, `render: ["typed search"]`, disabled fieldset: `["apple"]`).
- Focused SSR/hydration probe with the gate removed — all three engines kept `FormData === ["apple"]`, correct `label.control`/`input.labels`, label-click focus, and zero hydration errors.
- The test-only commit [`7e02d3185`](https://github.com/ariakit/ariakit/commit/7e02d3185fe959c9e462bbfa8e8696685a64b251) provides red proof for the core regression; the final-head GitHub check matrix is fully green.

I did not rerun the entire monorepo matrix locally; GitHub's complete final-head matrix covers it. I found no additional changeset, security, or performance concerns.

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: I tested this exact simplification on `ff54cbb54`, including the existing focused browser regression. Removing only the guard makes Firefox lose the Combobox accessible name after hydration: `q.combobox("Favorite fruits")` times out in 3/6 tests, and the accessibility snapshot shows the label text next to an unnamed combobox. Restoring the guard makes the same Firefox run pass 6/6.

The gate is therefore still serving the label-registration transition, even though the new hidden mirrors are not labelable. The existing regression is `preserves the input label metadata after hydration` in `app/src/sandbox/combobox-5058/test-browser.ts`.

I am keeping the gate and its subscription. Pre-hydration serialization for the normal composite path remains outside this narrow #5058 fix because removing the workaround would regress Firefox accessibility.

diegohaz:
**Declined**

- [Review reply](https://github.com/ariakit/ariakit/pull/6795#discussion_r3625107707): kept the `composite && !baseElement` gate because removing it makes Firefox lose the Combobox accessible name after hydration (3/6 focused tests fail; restored gate passes 6/6).

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: Should test again

diegohaz: Retested against the exact [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf) production preview path. With the guard present, the focused Firefox suite passed 6/6. I then removed only:

```tsx
if (composite && !baseElement) return element;
```

After a fresh app build on unique ports, Firefox failed 3/6, including the label regression. Every failure timed out resolving `q.combobox("Favorite fruits")`, and the accessibility snapshot showed the “Favorite fruits” text beside an unnamed combobox. Restoring the guard, rebuilding, and rerunning serially with `--workers=1 --retries=0` returned to 6/6.

My earlier in-memory probe was a false negative: direct `label.control` and `input.labels` checks did not exercise Firefox's accessible-name path through the actual Astro preview hydration lifecycle.

**Validation outcome: Invalid or obsolete.** The proposed simplification is not safe on the current head. Keep the gate and its subscription; pre-hydration serialization would need a different label-registration design.

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: The mirrors do not follow the directly supplied `aria-disabled` state. `useFocusable` intentionally treats `aria-disabled={true}` as disabled through `disabledFromProps` and emits native `disabled` for the visible input, but this hidden input receives only the separately destructured `disabled` prop.

```tsx
<ComboboxProvider defaultSelectedValue={["apple"]}>
  <Combobox
    aria-disabled
    composite={false}
    name="fruits"
  />
</ComboboxProvider>
```

I reproduced this on the current head: the visible input was natively disabled, while `new FormData(form).getAll("fruits")` still returned `["apple"]`. A checkout that disables an unavailable multi-select field this way therefore submits its stale selection even though the rendered control is disabled.

Please derive the mirror state with the same established predicate and add a public-behavior regression:

```tsx
const formDisabled = disabledFromProps({
  disabled,
  "aria-disabled": props["aria-disabled"],
});
```

Using `formDisabled` for the mirrors made the focused integration suite pass 3/3; without it, the new assertion expected `[]` but received `["apple"]`.

**Assessment**

- Severity: Medium (verified, 99% confidence), because a rendered disabled field contributes unintended values to submission.
- Priority: P1 (estimate, 95% confidence), because this PR creates the mirror and states that it follows the Combobox disabled state.
- Breakage likelihood: Unlikely (estimate, 90% confidence), requiring an array selection, `name`, and direct `aria-disabled` rather than `disabled`.
- Fix effort: Easy (verified, 98% confidence), because the existing helper plus a focused regression passed locally.

diegohaz (commented):
### Same-head re-review

Re-reviewed [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf) using the latest prior marked review at the same SHA as the baseline. There are no new commits, so this pass focused on the changed review state, independently revalidated the relevant unresolved thread against the exact head, and rescanned surrounding disabled/form semantics.

I found one new current-head issue:

- [P1: direct `aria-disabled` does not disable the hidden mirrors](https://github.com/ariakit/ariakit/pull/6795#discussion_r3625787623). `useFocusable` turns the visible input into a native disabled control, but the mirrors still submit stale selected values because they receive only the `disabled` prop. The existing `disabledFromProps` helper is a verified local correction.

Relevant candidate outcomes:

- [PRRT_kwDOBjT0Fc6SsHlt / gate simplification](https://github.com/ariakit/ariakit/pull/6795#discussion_r3625787261): **Invalid or obsolete.** A fresh production-preview red/green comparison confirmed the gate is required for Firefox's post-hydration accessible name. The prior direct-DOM probe was a false negative and did not model that accessibility-tree path.
- The resolved `render` and `wrapElement` composition candidates remain explicit declined non-goals. There is no commit delta or material new evidence that would justify reopening those resolved threads in this pass.
- The previously resolved select-proxy lifecycle, required/constraint-validation, and test-query candidates remain obsolete, declined, or fixed under their recorded dispositions.

Verification performed:

- Fetched current `origin/main` and `pull/6795/head`; the PR head remains `ff54cbb54`.
- Current guarded production preview: focused Firefox browser suite passed 6/6.
- Removed only `if (composite && !baseElement) return element;`, rebuilt on unique ports, and reran Firefox: 3/6 failed because `q.combobox("Favorite fruits")` could not resolve an accessible name.
- Restored the guard, rebuilt, and reran with `--workers=1 --retries=0`: Firefox passed 6/6.
- Current `aria-disabled` behavior proof: the visible input was natively disabled, but the focused integration assertion expected `FormData` `[]` and received `["apple"]`.
- Prototype using `disabledFromProps` for the mirrors: focused React integration suite passed 3/3.
- Final-head `CI=true pnpm test-react app/src/sandbox/combobox-5058/test.ts`: 3/3 passed.
- Package TypeScript check and `git diff --check` passed; the worktree is clean and the final-head GitHub check matrix is green.

I inspected the Combobox/Composite/Focusable lifecycle, `useWrapElement` and `createElement`, store subscriptions, the sandbox browser and happy-dom coverage, the changeset, and the PR's implementation-review reassessment. The new finding is a local state-derivation gap, so it does not undermine the repeated-hidden-input direction or justify another proxy redesign. The test structure and changeset remain appropriate. I did not rerun React 18 or the entire monorepo matrix locally because the code head is unchanged; both are covered by the green exact-head CI and the prior marked review. No follow-up issue is registered, and I found no additional security or performance concern.

packages/ariakit-react-components/src/combobox/combobox.tsx
diegohaz: Fixed in [`8c4eaf802`](https://github.com/ariakit/ariakit/commit/8c4eaf8023c47741fcc23a4910865ef9a0edc025): mirrored inputs now derive their disabled state with `disabledFromProps`, so direct `aria-disabled` selections are omitted from `FormData`. Added targeted browser and happy-dom coverage in [`8c4eaf802`](https://github.com/ariakit/ariakit/commit/8c4eaf8023c47741fcc23a4910865ef9a0edc025).

diegohaz:
**Fixed**

- [`8c4eaf802`](https://github.com/ariakit/ariakit/commit/8c4eaf8023c47741fcc23a4910865ef9a0edc025) ([reply](https://github.com/ariakit/ariakit/pull/6795#discussion_r3626057869)): direct `aria-disabled` now disables the repeated form mirrors, with browser and happy-dom regressions.

diegohaz (commented):
### Re-review

Re-reviewed [`8c4eaf802`](https://github.com/ariakit/ariakit/commit/8c4eaf8023c47741fcc23a4910865ef9a0edc025) using the latest prior marked review at [`ff54cbb54`](https://github.com/ariakit/ariakit/commit/ff54cbb546bad5e4052e1d20fb8c70344ddf93cf) as the incremental baseline. The branch was also rebased onto [`76f9d101e`](https://github.com/ariakit/ariakit/commit/76f9d101e4dda1aec597f065727f613500d449f1); `git range-diff` confirms the six previously reviewed commits are patch-equivalent, so the substantive new scope is the four-file `8c4eaf802` commit rather than the unrelated dependency and Playwright changes inherited from `main`.

No new findings.

The prior [`aria-disabled` mirror finding](https://github.com/ariakit/ariakit/pull/6795#discussion_r3625787623) is resolved in the current code. The mirrors now derive their disabled state with the same established `disabledFromProps` predicate used by the visible Focusable input, and `useWrapElement` tracks the derived boolean. This covers direct boolean and string ARIA values without changing native `disabled`, external-form ownership, or the intentionally scoped-out `render`/`wrapElement` metadata cases.

Regression proof and verification:

- Exact-head focused React integration suite: 4/4 passed.
- Temporarily changed only the mirror back to `disabled={disabled}`: the new test failed for the expected reason, receiving `["apple"]` instead of `[]`; restoring `disabled={formDisabled}` returned to 4/4 passing.
- Fresh `CI=true pnpm -F app build-lite` succeeded.
- Focused production-preview browser suite passed 21/21 across Chrome, Firefox, and Safari, including the new disabled-value case and the existing Firefox label-registration regression.
- Affected-package TypeScript and `git diff --check` passed; the final worktree is clean.
- The exact-head GitHub check matrix is complete and green.

The authoritative review-thread query was paginated to exhaustion, including nested comments, and currently has no unresolved threads. The previously disproved base-element-gate simplification remains invalid; the resolved `render` and `wrapElement` composition cases remain explicit non-goals with no new evidence that changes their dispositions.

The cycle-6 implementation reassessment remains sound: this fix closes a local state-derivation gap and adds no proxy state, effects, or new public contract, so repeated hidden inputs remain the simplest direction for `FormData.getAll(name)`. I found no new security, performance, release-metadata, or follow-up issue. I did not rerun the entire monorepo matrix locally; React 18 and the broader matrix are covered by green exact-head CI.

PR #6797, opened 2026-07-21
Limit Safari launch timeout in CI

diegohaz:
## Motivation

The [app Safari job](https://github.com/ariakit/ariakit/actions/runs/29850198253/job/88701254517) took 24.9 minutes after replacement WebKit processes wedged during retries. Each launch attempt could wait Playwright's 180,000 ms default before failing.

A workflow or job timeout would also cap legitimately long test runs. The failure should instead be bounded at the browser-launch operation that wedged.

Across the latest 59 completed app Safari jobs on the current macOS runner image, the conservative startup upper bound had an 8.17-second median, 14.86-second p95, and 23.51-second maximum. This includes worker setup, context and page creation, and initial navigation, so the browser launch itself completed sooner.

## Solution

Hoist the existing headed-mode delay and set a 45-second browser launch timeout only for the app Safari project in CI:

```ts
const slowMo = HEADED ? 150 : undefined;

launchOptions: {
  slowMo,
  timeout: CI ? 45_000 : undefined,
}
```

Playwright merges project `use` objects shallowly, so Safari still declares `slowMo`; sharing the value keeps the base and Safari settings in sync.

This leaves more than 21 seconds of headroom over the observed maximum while cutting a wedged launch wait by 75%. Other browser projects keep Playwright's default, non-CI Safari runs keep the default, and local headed Safari retains its existing `slowMo` setting.

The Safari retry count and workflow-level timeouts remain unchanged.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- Loaded the config with CI enabled and confirmed Safari resolves to `{ "timeout": 45000 }`
- Loaded the headed config with CI unset and confirmed Safari resolves to `{ "slowMo": 150 }`
- `env CI=true pnpm -F app exec playwright test --list --project safari --grep-invert @visual` (285 tests listed)

### DIFF
--- a/app/playwright.config.ts
+++ b/app/playwright.config.ts
@@ -126,7 +126,16 @@ export default defineConfig({
         {
           name: "safari",
           testMatch: testMatchersFor("safari", "browser"),
-          use: devices["Desktop Safari"],
+          use: {
+            ...devices["Desktop Safari"],
+            launchOptions: {
+              slowMo: HEADED ? 150 : undefined,
+              // Healthy macOS CI launches complete initial page setup within
+              // 24 seconds; fail a wedged WebKit process without waiting for
+              // Playwright's three-minute default.
+              timeout: CI ? 45_000 : undefined,
+            },
+          },
           retries: CI ? 3 : 1,
         },
         {
--- a/app/playwright.config.ts
+++ b/app/playwright.config.ts
@@ -6,6 +6,7 @@ if (process.argv.includes("--headed")) {
 
 const CI = !!process.env.CI;
 const HEADED = process.env.PWHEADED === "true";
+const slowMo = HEADED ? 150 : undefined;
 const PERF = process.env.PERF_TEST === "true";
 const port = Number(process.env.APP_PORT) || 4321;
 const nextjsPort = Number(process.env.NEXTJS_PORT) || 3000;
@@ -66,7 +67,7 @@ export default defineConfig({
     screenshot: "only-on-failure",
     trace: "on-first-retry",
     launchOptions: {
-      slowMo: HEADED ? 150 : undefined,
+      slowMo,
     },
   },
   expect: {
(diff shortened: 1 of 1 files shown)


app/playwright.config.ts
```diff
@@ -126,7 +126,16 @@ export default defineConfig({
         {
           name: "safari",
           testMatch: testMatchersFor("safari", "browser"),
-          use: devices["Desktop Safari"],
+          use: {
+            ...devices["Desktop Safari"],
+            launchOptions: {
+              slowMo: HEADED ? 150 : undefined,
```
diegohaz: Isn't this already defined in the base `use`? Do we need to redefine it here?

diegohaz: This is not an issue because Playwright merges the top-level and project `use` objects shallowly. The Safari project’s `launchOptions` replaces the base `launchOptions`, so `slowMo` must be repeated here to preserve the existing headed Safari behavior while adding the CI timeout.

diegohaz:
**Invalid**

- [`No change`](https://github.com/ariakit/ariakit/blob/8574fdfdedae4cf320906c824700720d23aaad62/app/playwright.config.ts#L131-L138) ([reply](https://github.com/ariakit/ariakit/pull/6797#discussion_r3624741817)): Playwright merges project `use` settings shallowly, so Safari must repeat `slowMo` when its project-specific `launchOptions` adds the CI timeout.

app/playwright.config.ts
diegohaz: Fixed in [`a5075c17c`](https://github.com/ariakit/ariakit/commit/a5075c17c1d8219195399b3d04b59cc65a9f790c): hoisted the shared `slowMo` value so the base and Safari launch options stay in sync while Safari continues to declare the property required by Playwright’s shallow merge.

diegohaz:
**Fixed**

- [`a5075c17c`](https://github.com/ariakit/ariakit/commit/a5075c17c1d8219195399b3d04b59cc65a9f790c) ([reply](https://github.com/ariakit/ariakit/pull/6797#discussion_r3624878365)): Hoisted the shared `slowMo` value so the base and Safari launch options cannot drift while preserving the explicit Safari property required by Playwright’s shallow merge.

PR #6800, opened 2026-07-21
Preserve bare layer colors at higher contrast

diegohaz:
## Motivation

Bare nested `ak-layer` elements reapplied the global contrast bias to the inherited parent color. At intermediate contrast levels such as `5`, this made each unmodified dark layer progressively darker even though a bare layer should preserve its parent surface.

```html
<div class="ak-layer [--contrast:5]">
  <div class="ak-layer">Same background</div>
</div>
```

## Solution

Track whether a layer is nested and whether it has a `layer-*` or `state-*` modifier, then suppress the contrast bias only for nested bare layers. Top-level and modified layers retain their existing contrast behavior.

Add focused browser coverage for light and dark schemes at contrast levels `0`, `5`, `50`, and `100`, including modified layers nested through a bare layer. Regenerate the computed-style snapshots; the remaining changes are confined to the disabled high-contrast matrix, where contrast is scaled to an intermediate value.

### DIFF
--- a/.changeset/300-fix-layer-contrast-inheritance.md
+++ b/.changeset/300-fix-layer-contrast-inheritance.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/tailwind": patch
+---
+
+Fixed bare `ak-layer` elements to preserve inherited background colors when the global contrast setting increases.
--- a/app/src/sandbox/ariakit-tailwind-layer-contrast-inheritance/index.react.tsx
+++ b/app/src/sandbox/ariakit-tailwind-layer-contrast-inheritance/index.react.tsx
@@ -1,0 +1,37 @@
+import "./style.css";
+
+export default function Example() {
+  return (
+    <section aria-label="Parent layer" className="ak-layer [--contrast:5] p-4">
+      <section aria-label="Child layer" className="ak-layer p-4">
+        <section aria-label="Grandchild layer" className="ak-layer p-4">
+          Bare nested layers
+        </section>
+        <section
+          aria-label="Nested layer offset"
+          className="ak-layer ak-layer-offset-10 p-4"
+        >
+          Nested layer offset
+        </section>
+        <section
+          aria-label="Nested state darken"
+          className="ak-layer ak-state-darken-10 p-4"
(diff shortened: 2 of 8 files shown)


diegohaz (commented):
Initial review of merge base `76f9d101e4dda1aec597f065727f613500d449f1` through head `b558f03b5294484925277bc6c8d8715240c716c7`.

No substantive findings.

I reviewed the Tailwind source change and generated CSS, the focused sandbox and Chrome tests, both computed-style snapshot diffs, the changeset, surrounding layer/context machinery, the documented `ak-layer` contract, and performance implications. There were no existing review threads to validate.

Snapshot audit:

- Each snapshot keeps the same 3,721 scalar paths and structure.
- Dark high contrast changes 404 entries; light high contrast changes 379.
- Every change is confined to `ak-layer-<number> disabled`. Direct modified outer layers remain unchanged.
- Every changed entry changes background lightness only; downstream text, border, and shadow colors recalculate from that corrected background. Hue, chroma, spacing, radius, border width, outline, layout, and structure do not drift.
- The scope is expected: high contrast sets contrast to 100, while disabled content scales it to 0.25, exposing the old repeated bias of `0.25 × 0.3334 ≈ 0.08335` per bare nested layer.

Representative chains:

```text
dark before:  0.1467 → 0.0633 → 0
dark after:   0.1467 → 0.1467 → 0.1467
light before: 0.9834 → 1 → 1
light after:  0.9834 → 0.9834 → 0.9834
```

Validation performed:

- `pnpm -F @ariakit/tailwind build` regenerated `output.css` with no diff.
- `git diff --check` passed.
- Exact base/head deployment probes demonstrated the broken dark contrast-5 nesting on base and verified bare inheritance plus direct/nested modifier equality on head for light/dark × contrast 0/5/50/100.
- All current CI checks pass, including app Chrome/Firefox/Safari and Tailwind performance; the performance workflow reports no significant change.

I did not locally regenerate CI snapshots or run the full app Playwright matrix; repository policy keeps snapshot capture in CI. The exact deployment probes and successful CI cover the affected browser behavior.

PR #6805, opened 2026-07-22
Close popovers when focus leaves an iframe

### LINKED ISSUE #3031: [ComboBox] ComboBox doesn't close when tabbing out of it
### Current behavior

The combobox menu doesn't close when it becomes unfocused via clicking the tab key.

1. Go to https://ariakit.org/previews/combobox-multiple
2. Click the combobox to open the menu
3. Press tab to move focus out of the combobox

### Expected behavior

The menu closes when pressing tab, since the combobox is no longer focused.

### Workaround

_No response_

### Possible solutions

_No response_

diegohaz:
## Motivation

Popups rendered inside a same-origin iframe can stay open when focus moves into an ancestor document because the popup document does not observe that outside interaction. The same boundary can hide pointer interactions occurring in an existing sibling iframe.

This behavior comes from the shared Dialog outside-interaction hook, so it also affects `Popover`, `ComboboxPopover`, `Hovercard`, `Menu`, `SelectPopover`, and `Tooltip`. A true browser or application window blur must continue to leave the popup open.

Fixes #3031.

## Solution

Outside-interaction listeners now start at the highest readable same-origin ancestor window and include its existing readable frame tree. When an event occurs, its composed path is extended with readable iframe hosts so the existing Dialog classifier can reason about both shadow roots and document boundaries.

Current `main` owns inside markers by the actual Dialog content element and projects composed events into that element's root. The merge preserves that model, including same-ID sibling dialogs, and adds iframe hosts to the same `EventTargets` representation instead of maintaining a second classifier.

This closes an embedded popup when focus reaches an ancestor document and closes a root popup after a pointer interaction in an existing sibling frame. The same host-chain check keeps contained frames, parent disclosures, persistent elements, and nested dialogs inside. Accepted cross-document focus interactions preserve focus on the outside target.

The implementation uses the existing native `focusin`, click, and context-menu events. It does not observe frame lifecycle changes or fabricate focus events.

Recorded `mousedown` events keep both the composed/iframe-host path and the target projected into the dialog root. The projected target is validated before marker traversal, preserving the existing foreign-event contract for standalone clicks.

The focused regression demonstrates the main behavior:

```ts
await page.keyboard.press("Tab");

await test.expect(q.textbox("Parent target")).toBeFocused();
await test.expect(frame.listbox("Suggestions")).not.toBeVisible();
```

The sandbox also covers a parent disclosure controlling content portaled into an iframe, drag suppression from that disclosure, pointer interaction in an existing sibling iframe, interaction inside an existing contained iframe, document-local active-descendant ID reuse in a sibling frame, and a same-origin SVG frame without a document body. All 21 cases (7 tests × 3 engines) pass in Chrome, Firefox, and Safari locally. Together with current `main`'s sibling-dialog and shadow-root regressions, the merged runtime passes 63 focused browser cases across those engines.

A separate foreign-target regression passes in all three engines and [records the reviewed `mousedown` → standalone-click sequence](https://github.com/user-attachments/assets/dfe63d10-b216-4214-a774-ec1f7fe09b22).

## Preview

[Open the exact-head sandbox](https://d7d4327f-ariakit-preview.ariakit.workers.dev/react/previews/dialog-iframe-interactions/).

Before: the embedded combobox is focused and its suggestions are visible.

Embedded combobox focused with suggestions visible

After: focus is on the parent target and the suggestions are closed.

Parent target focused after embedded suggestions close

## Unsupported scope

This change intentionally does not cover:

- Cross-origin frame contents.
- Frames inserted or navigated after the popup opens. Reopening the popup refreshes the static listener snapshot.
- Window-only or iframe-host-only focus transitions when the browser does not expose a document `focusin` event.
- Frames that are not exposed through `window.frames`, including browser-specific shadow-root edge cases.
- Separate documents opened with `window.open()`.
- Synthetic recovery for focus signals that have already been missed.

## Acknowledgments

Thanks to [@emillaine](https://github.com/emillaine) for reporting the issue, [@ciampo](https://github.com/ciampo) for proposing the pointer behavior, and [@donaldpipowitch](https://github.com/donaldpipowitch) for providing the iframe click reproduction.

The current-main integration preserves its Element-owned Dialog markers and composed-path handling. That upstream behavior removes the matching-ID limitation listed by the pre-merge PR without reintroducing the PR-specific marker machinery removed in Cycle 8.

Cycle 9 preserves the narrowed Cycle 8 architecture and records two realm-correctness fixes before the earlier audit entries.

<details>
<summary>Cycle 9: Keep the static frame-tree design and correct realm assumptions</summary>

**Assessment**

**Issue severity:** Medium. Affected users see a stale popup, and the SVG/XML path also raises an uncaught exception.

The ninth finding-bearing cycle identified two failures inside the supported existing readable same-origin frame tree. An outside active-descendant target can reuse a disclosure ID in another document and be classified as inside, while an event from a same-origin SVG or XML frame can reach the detached-target guard with a document that has no `body` and throw before the popup closes.

Both cases are uncommon, but they contradict the static same-origin frame contract already claimed by the PR. The implementation cost is limited to correcting two existing realm assumptions. It requires no observer, registry, lifecycle state, synthetic event, platform policy, or broader support contract. The changeset link regression is documentation-only and independent of the runtime design.

**Decision**

Keep the Cycle 8 static frame-tree architecture. Resolve `aria-activedescendant` in the target element's document and use `Element.isConnected` for the document-neutral connectivity guard, with focused browser regressions for both cases. Restore the verified public API links in the changeset. At this cycle, keep every unsupported-scope item unchanged and do not add new iframe machinery.

</details>

Cycle 8 supersedes the lifecycle scope chosen in the earlier assessments. Cycles 3 and 6 remain below as historical audit context.

<details>
<summary>Cycle 8: Reduce the supported frame lifecycle</summary>

**Assessment**

The previous design reached 14 changed files and 1,624 added lines. Most of that surface implemented a private frame observer, a subscription registry, navigation cleanup, browser-specific Window-focus recovery, and synthetic event fallback. Repeated review findings continued to concentrate in that machinery.

The user-facing issue has meaningful keyboard impact, but its core case only requires observing native interactions across an already established same-origin frame boundary. Supporting frames inserted or navigated while a popup remains open, plus browser-specific focus transitions without `focusin`, is disproportionate to issue #3031.

The narrowed result changes 5 files, with 304 additions and 24 deletions, including its sandbox and browser coverage.

**Decision**

Support the existing readable same-origin frame tree captured when the popup opens. Start the existing global listeners at the highest readable ancestor window and classify native event targets through their iframe-host chain.

Remove the frame observer, subscription registry, Window-focus fallback, navigation recovery, synthetic focus events, and marker-identity redesign. Dynamic frame lifecycle changes and missing native focus signals are explicit non-goals. This keeps the implementation aligned with the original report while preserving existing Dialog outside-interaction semantics.

</details>

<details>
<summary>Cycle 3 (superseded): Continue the scoped frame-tree design</summary>

**Assessment**

Issue #3031 has medium user impact: keyboard or pointer users can leave an embedded same-origin popup visually stale, and an accepted close can move focus back to the wrong document. The transition is expected in documentation previews, editors, and other iframe-based applications. Three implementation review cycles exposed identity collisions, late-frame lifecycle gaps, browser-specific missing host `focusin`, focus-restoration behavior, and compatibility concerns.

The resulting maintenance surface was one private frame observer, one narrow Window-focus fallback, element-owned inside markers, and focused browser coverage. This was more machinery than the original static listener traversal, but each part corresponded to a reproduced same-origin behavior and stayed within the Dialog implementation.

**Decision**

Continue the scoped design. A child-window `blur` alternative was rejected because it loses the actual interaction type and target and cannot reliably distinguish ancestor focus from application blur. The observer remained private rather than expanding `@ariakit/utils`, and the existing exported string-owner marker behavior remained compatible. Cross-origin traversal, true browser or application blur, and separate `window.open()` documents remained intentionally unsupported.

</details>

<details>
<summary>Cycle 6 (superseded): Replace the React lifecycle bridge with a synchronous registry</summary>

**Assessment**

The sixth finding-bearing cycle confirmed three related failures in the frame machinery: the fallback fabricated the event exposed to the public callback, frame lifecycle changes waited for React effects before listeners reached the replacement document, and cleanup could dereference a navigated `WindowProxy` instead of the document that was subscribed. These were consequences of the version-state bridge rather than independent edge guards.

Issue #3031 still had medium user impact, and added or navigated same-origin editor, payment, and preview frames were considered plausible while a popup was open. Removing lifecycle support would have reduced code, but it would also have discarded reproduced behavior that the PR then intentionally covered. Adding more React state or event fabrication would have increased maintenance risk.

**Decision**

Keep support for existing, added, and navigated same-origin frames, but redesign the lifecycle bridge. A private registry refreshes concrete document and Window-focus subscriptions synchronously from observable frame lifecycle signals, cleanup retains the exact subscribed document, and iframe-host projection remains internal to classification. The public callback receives the original native focus event whenever that event can be observed. The only synthetic path is an explicit child-realm fallback for a newly inserted, already-focused document first discovered after its native event has ended. Remove the React version state. Cross-origin traversal, shadow-root-hosted frames absent from `window.frames`, true browser or application blur, and separate `window.open()` documents remain intentionally unsupported.

</details>

## Workaround

For users who cannot upgrade, a combobox rendered in a same-origin iframe can listen for focus entering its immediate parent document and hide its store. Track that close so `autoFocusOnHide` does not move focus back into the iframe:

```tsx
import * as React from "react";

function EmbeddedCombobox() {
  const store = Ariakit.useComboboxStore();
  const baseElement = Ariakit.useStoreState(store, "baseElement");
  const focusMovedToParent = React.useRef(false);

  // TODO: Remove once https://github.com/ariakit/ariakit/issues/3031 is fixed.
  React.useEffect(() => {
    if (!open) return;
    const frameWindow = baseElement?.ownerDocument.defaultView;
    if (!frameWindow || frameWindow.parent === frameWindow) return;

    let parentDocument: Document;
    try {
      parentDocument = frameWindow.parent.document;
    } catch {
      return;
    }

    const hide = () => {
      focusMovedToParent.current = true;
      store.hide();
    };
    parentDocument.addEventListener("focusin", hide, true);
    return () => parentDocument.removeEventListener("focusin", hide, true);
  }, [store, open, baseElement]);

  const autoFocusOnHide = () => {
    const shouldAutoFocus = !focusMovedToParent.current;
    focusMovedToParent.current = false;
    return shouldAutoFocus;
  };

  return (
    <Ariakit.ComboboxPopover
      store={store}
      autoFocusOnHide={autoFocusOnHide}
    />
  );
}
```

This workaround is intentionally narrow: it requires a same-origin immediate parent, covers focus transitions rather than parent clicks or context menus, and bypasses a custom `hideOnInteractOutside` callback for this close.


packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -25,9 +25,74 @@ interface EventOutsideOptions {
...
+  }
+  const targetElements = getFrameElements(target);
+  if (
+    targetElements.some((element) => element.hasAttribute("data-focus-trap"))
+  ) {
+    return true;
+  }
+  return targetElements.some((element) =>
+    isElementInside(element, contentElement.id),
+  );
```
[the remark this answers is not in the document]
diegohaz: Confirmed — disposition: **Address now**.

`isElementInside(element, contentElement.id)` treats the ID-derived inside marker as a cross-document identity, but IDs are only unique within one `Document`. I reproduced this on the current head with open dialogs in two sibling same-origin frames using the same explicit ID: clicking the second dialog leaves the first open and its `hideOnInteractOutside` callback never runs. Independently mounted editors/widgets commonly reuse deterministic IDs this way.

**Assessment:** Medium severity (~90%); P1 priority (~95%); Possible likelihood (~75%); Moderate fix (~85%).

Please use a private per-dialog identity (for example, a unique symbol/token or owner-token set) instead of the public DOM ID, and add a browser regression with the same explicit dialog ID in sibling documents.

**Concrete `@ariakit/react` failure**

A split-view editor hosts two same-origin editor panes in sibling iframes, each running the same reusable formatting inspector. Reusing `id="format-controls"` is valid because each iframe has its own `Document`, and the host may restore both non-modal inspectors open:

```tsx

function FrameInspector({ label }: { label: string }) {
  const [open, setOpen] = useState(true);
  return (
    <Ariakit.PopoverProvider open={open} setOpen={setOpen}>
      <Ariakit.PopoverAnchor>{label}</Ariakit.PopoverAnchor>
      <Ariakit.Popover
        id="format-controls"
        aria-label={`${label} formatting`}
        autoFocusOnShow={false}
      >
        <button type="button">Apply formatting</button>
      </Ariakit.Popover>
    </Ariakit.PopoverProvider>
  );
}
```

Mounting one instance in each iframe exposes the bug: when the user clicks **Apply formatting** in the right pane, the left `Popover` should treat that as an outside interaction and call its public `setOpen(false)` contract. Instead, the identical inside marker makes the right popover look owned by the left one, so both formatting inspectors remain visible. This path uses only components and controlled state exported by `@ariakit/react`.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -80,32 +146,52 @@ function useEventOutside({
...
-      // of the content element, we call the listener.
+      // of the content element, we call the listener. A target outside of the
+      // content document is also outside unless one of the checks above maps
+      // it to an element that's inside.
       callListener(event);
     };
-    const win = contentElement ? getWindow(contentElement) : undefined;
-    return addGlobalEventListener(type, onEvent, capture, win);
-  }, [open, capture, store, type, callListener, contentElement, focusedRef]);
+    return addGlobalEventListener(type, onEvent, capture, eventWindow);
```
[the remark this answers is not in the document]
diegohaz: Confirmed — disposition: **Address now**.

`addGlobalEventListener` snapshots `scope.frames` when this effect installs, so later iframe creation or navigation never adds listeners for the new document. I reproduced both cases on this head: opening the dialog, then appending a same-origin sibling iframe or navigating an existing iframe, and clicking inside it leaves the dialog open with no outside callback. This affects lazy-loaded editors, payment frames, and previews created while a popup is open.

**Assessment:** Medium severity (~90%); P1 priority (~95%); Possible likelihood (~85%); Moderate fix (~80%).

Please make frame subscriptions lifecycle-aware (for example, refresh on frame load/tree changes) and cover both a late outside iframe and a late contained iframe so the refresh does not regress the inside case.

**Concrete `@ariakit/react` failure**

A document app renders an **Insert** menu while its same-origin rich-text editor iframe is still loading. The iframe mounts after route data becomes ready:

```tsx

function EditorShell({ editorLoaded }: { editorLoaded: boolean }) {
  const [open, setOpen] = useState(true);
  return (
    <>
      <Ariakit.MenuProvider open={open} setOpen={setOpen}>
        <Ariakit.MenuButton>Insert</Ariakit.MenuButton>
        <Ariakit.Menu autoFocusOnShow={false}>
          <Ariakit.MenuItem>Image</Ariakit.MenuItem>
        </Ariakit.Menu>
      </Ariakit.MenuProvider>
      {editorLoaded && (
        <iframe title="Document editor" src="/editor" />
      )}
    </>
  );
}
```

If `editorLoaded` becomes true after the `Menu` opens, clicking a control inside `/editor` should trigger the menu's normal outside-interaction behavior and call `setOpen(false)`. Because the new document was not subscribed, the event never reaches `Menu`; the Insert menu remains over the editor while the user starts editing. Navigating an already-mounted iframe from `about:blank` to `/editor` produces the same public failure.

app/src/sandbox/combobox-3031/index.react.tsx
```diff
@@ -0,0 +1,117 @@
...
+      {iframeBody ? createPortal(<EmbeddedContent />, iframeBody) : null}
+      {siblingBody
+        ? createPortal(<div>Sibling click target</div>, siblingBody)
+        : null}
+      <iframe
+        ref={setIframe}
+        title="Embedded content"
+        style={{ border: "1px solid", height: 160, width: 320 }}
+      />
+      <iframe ref={setSiblingIframe} title="Sibling frame" tabIndex={-1} />
```
diegohaz: Making this sibling iframe `tabIndex={-1}` masks the normal keyboard path where focus lands on a browsing-context host. An iframe is tabbable by default (`tabIndex === 0`); after removing only this prop, I reproduced the same result in Chromium, Firefox, and WebKit: pressing Tab from the embedded combobox focuses the sibling iframe host while `Suggestions` remains open. A keyboard user tabbing into an adjacent editor/chart/preview therefore carries a stale listbox into the new task.

**Assessment:** Medium severity (~90%); P1 priority (~90%); Possible likelihood (~85%); Hard fix (~70%).

Please handle this browsing-context-host transition while still preserving real window/application blur, and keep a regression for the default iframe behavior, for example:

```ts
await page.keyboard.press("Tab");

await expect(siblingIframe).toBeFocused();
await expect(frame.listbox("Suggestions")).not.toBeVisible();
```

.changeset/300-dialog-iframe-interactions.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Close popups across iframe boundaries
+
+The [`Dialog`](https://ariakit.com/reference/dialog) component and components that inherit its outside-interaction behavior, including [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), now close when focus or pointer interactions move from a same-origin iframe into an ancestor or sibling document. The outside target retains focus, and `hideOnInteractOutside` receives the original event.
```
diegohaz: This release note understates the public behavior change. The shared hook also changes Popover, SelectPopover, Hovercard, Menu, and Tooltip (through both `@ariakit/react-components` and `@ariakit/react`), and it fixes the inverse parent-popup → descendant-iframe direction too. Consumers of those components—or a parent app embedding a same-origin canvas—will not recognize the change from the current Dialog/ComboboxPopover-only wording.

**Assessment:** Low severity (~95%); P2 priority (~95%); Likely (~100%, verified from the public exports); Easy fix (~98%).

Please make the wording direction-neutral (for example, “across same-origin iframe boundaries”), name and link every affected public derived component, and link the public `hideOnInteractOutside` API occurrence.

.changeset/300-dialog-iframe-interactions.md
```diff
@@ -0,0 +1,10 @@
...
+
+Thanks to [@emillaine](https://github.com/emillaine), [@ciampo](https://github.com/ciampo), and [@donaldpipowitch](https://github.com/donaldpipowitch) for reporting the issue.
```
diegohaz: The changeset correctly credits `ciampo` and `donaldpipowitch`, whose issue/comment contributions are incorporated by the final pointer implementation, but the implementation commit [`09b54efb5`](https://github.com/ariakit/ariakit/commit/09b54efb5267db82d1cdf9be962a0cf7f66c97ad) has no co-author trailers. If merged as-is, the changelog will acknowledge them while the repository history will not.

**Assessment:** Low severity (~99%); P2 priority (~98%); Likely (~100%, verified from the commit); Easy fix (~95%).

Please amend the final implementation commit with:

```text
Co-authored-by: Marco Ciampini <1083581+ciampo@users.noreply.github.com>
Co-authored-by: Donald Pipowitch <1152805+donaldpipowitch@users.noreply.github.com>
```

diegohaz (commented):
## Initial review summary

Reviewed the complete PR diff from merge base [`d641496b1`](https://github.com/ariakit/ariakit/commit/d641496b1e84395f687bb2cca3a4ef7b006047b5) through head [`09b54efb5`](https://github.com/ariakit/ariakit/commit/09b54efb5267db82d1cdf9be962a0cf7f66c97ad), including all four changed files, the shared dialog interaction algorithm, public derived-component/export surface, changeset policy, browser fixture/tests, issue context, commit attribution, current CI, and every existing review conversation.

I found five items to address:

- **P1 / Medium:** [ID-derived inside markers collide across sibling documents](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632263505).
- **P1 / Medium:** [frames created or navigated after opening are never subscribed](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632264255).
- **P1 / Medium:** [the fixture masks the default-tabbable sibling-iframe keyboard path](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632264957).
- **P2 / Low:** [the changeset omits affected public components, links, and the inverse direction](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632265472).
- **P2 / Low:** [the final implementation commit is missing contributor trailers](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632266254).

### Verification

- `pnpm tsc` passed.
- The focused dialog orchestration unit suite passed on both React configurations (12/12 each).
- `oxfmt --check`, `oxlint`, and `pnpm build-app-lite` passed.
- The focused browser suite passed locally in Chromium (5/5), Firefox (4/4 plus the intentional host-focus skip), and WebKit (5/5): 14 passed, 1 skipped.
- Independent browser probes reproduced both confirmed existing-thread defects, reproduced the ordinary sibling-iframe tab path in all three engines, and confirmed genuine window/application blur remains open.

The current Safari CI job is red in the new parent-document focus test: after `Tab`, the parent button was not focused on any of four attempts, and the popup remained visible in the captured state. Because the focused local WebKit run passes and this test does not exist at the base SHA, I could not establish a base comparison; Safari behavior is therefore not fully validated yet.

I also tested and rejected the suspected duplicate-callback path when entering child iframe content, and found no substantive performance issue in the current frame traversal at realistic depth. No out-of-scope follow-up was registered, and I found no other substantive issues.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40): inside markers now use the actual Dialog element as a private owner instead of a cross-document DOM ID. Added targeted unit and sibling-frame duplicate-ID coverage in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40).

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40): a private frame-tree observer now refreshes outside-interaction subscriptions when same-origin frames are added, removed, or navigated. Added late-outside, navigated-outside, and late-contained frame coverage in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40).

app/src/sandbox/combobox-3031/index.react.tsx
diegohaz: Fixed in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40): the sibling iframe is back in normal sequential order, with `tabIndex={0}` explicitly pinning that default for macOS WebKit CI, and a child-Window focus fallback handles browsers that omit host `focusin` without listening to `window.blur`. Added Chrome, Firefox, and Safari Tab/focus-retention coverage in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40).

.changeset/300-dialog-iframe-interactions.md
diegohaz: Fixed in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40): the release note is direction-neutral, names every affected Dialog-derived public surface, and links every available component reference plus `hideOnInteractOutside`. `CompositeOverflow` is named without a link because it has no public reference page.

.changeset/300-dialog-iframe-interactions.md
diegohaz: Fixed in [`10e65eb8a`](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40): the amended final implementation commit now includes the requested Marco Ciampini and Donald Pipowitch co-author trailers, and the changeset distinguishes their contribution roles.

diegohaz:
**Fixed**

- [10e65eb8a](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632795461)): Scoped inside markers to the actual Dialog element and added duplicate-ID coverage across documents.
- [10e65eb8a](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632796474)): Refreshes subscriptions for added, removed, and navigated same-origin frames, with outside and contained lifecycle regressions.
- [10e65eb8a](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632797314)): Restored the real sequential iframe Tab path and added a browser-safe host-focus fallback plus focus-retention coverage.
- [10e65eb8a](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632798326)): Expanded the changeset to every affected public surface, both directions, and the public callback contract.
- [10e65eb8a](https://github.com/ariakit/ariakit/commit/10e65eb8ae5ce2eb3dd4240190d67fae758a9f40) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3632799324)): Added the requested contributor trailers and clarified contributor roles.
- [201c42d20](https://github.com/ariakit/ariakit/commit/201c42d20674317555e7f7f9446b5230c955da04) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3633012492)): Re-keyed inside markers when an open Dialog replaces its DOM node, with persistent-element coverage.
- [a069807ae](https://github.com/ariakit/ariakit/commit/a069807ae208a2d8747017c28cab539194647a72) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3633015429)): Guarded iframe focus-listener registration and cleanup after cross-origin navigation, with a real opaque-navigation regression.

**Not actionable**

- ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3633113974)): The shadow-root projection path is unreachable because all three engines omit shadow-hosted frames from window.frames; complete support would require a separate discovery and lifecycle subsystem, not the suggested containment-only change.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -324,7 +325,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     // the dialog is open (hovercards and tooltips set it to the focus
     // source), so the listeners re-check it against the current state on
     // every event instead. See https://github.com/ariakit/ariakit/issues/6344
-    const restoreInsideMarks = markTreeInside(id, allElements);
+    const restoreInsideMarks = markTreeInside(dialog, allElements);
```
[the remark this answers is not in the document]
diegohaz: Fixed in [201c42d20](https://github.com/ariakit/ariakit/commit/201c42d20674317555e7f7f9446b5230c955da04): the tree-marking effect now uses the subscribed contentElement as its owner and reruns when that node changes, cleaning the old marks and rebuilding persistent/nested marks under the replacement. Added React 19, React 18, Chrome, Firefox, and Safari coverage that swaps the open Dialog from a div to a section with autoFocusOnShow disabled and verifies its persistent element remains inside.

packages/ariakit-react-components/src/dialog/utils/__frame-events.ts
```diff
@@ -0,0 +1,103 @@
...
+export function addGlobalWindowFocusListener(
+  listener: (event: FocusEvent, eventWindow: Window) => void,
+  scope: Window,
+) {
+  const cleanups: Array<() => void> = [];
+
+  for (const win of getFrameWindows(scope)) {
+    const onFocus = (event: FocusEvent) => listener(event, win);
+    win.addEventListener("focus", onFocus, true);
+    cleanups.push(() => win.removeEventListener("focus", onFocus, true));
```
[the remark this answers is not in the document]
diegohaz: Fixed in [a069807ae](https://github.com/ariakit/ariakit/commit/a069807ae208a2d8747017c28cab539194647a72): listener registration and removal are now guarded independently, and cleanup is retained only after registration succeeds, so one inaccessible WindowProxy cannot abort the remaining cleanup. Added a Chrome, Firefox, and Safari regression that navigates an observed same-origin iframe to an opaque data document, checks for escaped page errors, then confirms a newly added same-origin frame still closes the dialog.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -25,7 +29,74 @@ interface EventOutsideOptions {
...
+  disclosureElement: Element | null,
+) {
+  const targetInContentDocument = getElementInDocument(
+    target,
+    getDocument(contentElement),
+  );
+  if (
+    targetInContentDocument &&
+    contains(contentElement, targetInContentDocument)
+  ) {
```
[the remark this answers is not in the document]
diegohaz: Checked independently — disposition: **Invalid/inaccurate on the current head**.

In Chromium, Firefox, and WebKit, an iframe under an open shadow root is absent from window.frames (length remains 0), even though iframe.contentWindow.frameElement points back to the iframe. Clicking inside it emits focus/focusin/click only in the child Window/Document; no event reaches the parent Document or ShadowRoot. Both listener installers enumerate window.frames, so no listener is installed in that child document and the line-76 projection/classification path is never reached. A contained shadow-hosted iframe already leaves the dialog open because its interaction is unseen, not because it is misclassified as outside.

There is a distinct adjacent limitation: an outside shadow-hosted iframe is unseen as well. A composed-host projection alone cannot fix that; complete support would require discovering iframe windows through open shadow roots, subscribing their documents, observing known shadow roots for frame lifecycle changes, and preserving nested host ancestry (with closed roots remaining undiscoverable generically). That is a separate subsystem rather than a localized containment correction, so I am not adding the partial suggested change to this PR.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -145,7 +300,89 @@ export function useHideOnInteractOutside(
...
+            event: focusEvent,
+            target: frameElement,
+            contentElement,
+            disclosureElement,
+            focused: focusedRef.current,
+          })
+        ) {
+          return;
+        }
+        hideOnFocusOutside(focusEvent, true);
```
diegohaz: **[P1] Preserve the public event contract in the iframe-focus fallback**

`focusEvent` is created by script and never dispatched, so the public `hideOnInteractOutside` callback always sees `isTrusted === false`, an empty `composedPath()`, and no native current target even when a real keyboard Tab caused the Window focus. This breaks a practical policy that closes a formatting popover only for genuine user input while ignoring focus restored by application code:

```tsx

<Ariakit.Popover
  hideOnInteractOutside={(event) => event.isTrusted}
>
  Formatting tools
</Ariakit.Popover>;
```

I reproduced the public failure in Chromium, Firefox, and WebKit: a keyboard user Tabs into the default-tabbable same-origin editor iframe; the callback reports `focusin; trusted=false; target=Editor pane`, returns `false`, and the Popover stays open. The screenshot captures the focused editor pane, callback output, and still-visible Popover:

Public Popover stays open because iframe focus is reported as untrusted

Please keep the projected host event internal to classification and preserve native event fidelity for the public callback, or define an explicit public representation that does not silently turn a trusted user interaction into an undispatched event.

**Disposition:** Address now

**Assessment**
- Severity: Medium (estimate, 95% confidence)
- Priority: P1 (estimate, 95% confidence)
- Breakage likelihood: Possible (estimate, 85% confidence); it requires callback logic that inspects trust or dispatch metadata
- Fix effort: Hard (estimate, 80% confidence); host-target projection and native-event fidelity need an explicit API design

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -58,54 +129,130 @@ function isMouseEventOnDialog(event: Event | MouseEvent, dialog: Element) {
...
+}
+
+function useFrameTreeVersion(open?: boolean, eventWindow?: Window) {
+  const [version, setVersion] = useState(0);
+
+  useEffect(() => {
+    if (!open) return;
+    if (!eventWindow) return;
+    return observeFrameTree(eventWindow, () => {
+      setVersion((version) => version + 1);
```
diegohaz: **[P1] Refresh frame subscriptions before iframe load handlers can focus content**

This state update cannot install the new Document/Window listeners until React commits and runs the passive effects below. The observer receives the iframe `load` event in capture phase, but a consumer's own `onLoad` handler runs later in that same dispatch and can synchronously restore editor focus before those effects run; the previous realm's listeners disappeared with navigation. The checked navigation test clicks only after load settles, so it cannot expose this ordering.

A normal public use case is an embedded editor restoring its caret after navigation while formatting controls are open:

```tsx

<Ariakit.Popover aria-label="Formatting tools">…</Ariakit.Popover>
<iframe
  srcDoc={editorDocument}
  onLoad={(event) =>
    event.currentTarget.contentDocument
      ?.querySelector<HTMLElement>("[contenteditable], input")
      ?.focus()
  }
/>;
```

I reproduced this deterministically in Chromium and WebKit: the newly loaded editor input has focus, but the Popover remains visible. Firefox did not allow the exact synchronous `onLoad` focus in this probe, so this finding is limited to the two reproduced engines rather than inferred there.

Navigated editor input is focused while the public Popover remains open

Please refresh a navigation-safe listener registry synchronously/incrementally from the frame lifecycle signal instead of making React state the only bridge, and add a load-handler focus regression.

**Disposition:** Address now

**Assessment**
- Severity: Medium (estimate, 95% confidence)
- Priority: P1 (estimate, 95% confidence)
- Breakage likelihood: Possible (estimate, 85% confidence); load-time focus restoration is common in embedded editors
- Fix effort: Moderate (estimate, 85% confidence); the listener refresh needs to precede consumer load handlers

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -58,54 +129,130 @@ function isMouseEventOnDialog(event: Event | MouseEvent, dialog: Element) {
...
+        })
+      ) {
+        return;
+      }
       callListener(event);
     };
-    const win = contentElement ? getWindow(contentElement) : undefined;
-    return addGlobalEventListener(type, onEvent, capture, win);
-  }, [open, capture, store, type, callListener, contentElement, focusedRef]);
+    return addGlobalEventListener(type, onEvent, capture, eventWindow);
```
diegohaz: **[P1] Detach listeners from the Document that was actually subscribed**

`addGlobalEventListener` attaches through each child `WindowProxy`'s current `scope.document`, but its cleanup reads `scope.document` again. After iframe navigation that proxy now exposes the replacement Document, so the refresh removes from the new Document and leaves the mousedown/click/focusin/contextmenu listeners on a retained historical Document. The Window-focus helper has the analogous dynamic-target cleanup risk.

I retained the old same-origin `contentDocument`, navigated the frame, waited for this refresh, and dispatched the old Document's mousedown/click sequence. In Firefox the stale listener still closed the open public Dialog; Chromium and WebKit discarded it in this direct retained-document probe. A real host can hit the retained case when an editor SDK holds old documents during navigation or the browser restores one from history. With a public callback such as:

```tsx
<Ariakit.Popover
  hideOnInteractOutside={(event) => {
    analytics.track("popover-outside", event.type);
    return false;
  }}
>
  Inspector
</Ariakit.Popover>
```

a historical-document event can run outside state/analytics after navigation, and restoring then re-subscribing that Document can duplicate callbacks. Capture each concrete `Document`/Window event target at registration time and clean up that exact target rather than dereferencing the WindowProxy again.

**Disposition:** Address now

**Assessment**
- Severity: Medium (estimate, 90% confidence)
- Priority: P1 (estimate, 90% confidence)
- Breakage likelihood: Possible (estimate, 75% confidence); it requires a retained or restored iframe Document
- Fix effort: Moderate (estimate, 90% confidence); capture concrete targets and add retained-document/history cleanup coverage

.changeset/300-dialog-iframe-interactions.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Close popups across iframe boundaries
+
+The [`Dialog`](https://ariakit.com/reference/dialog) component and components built on it, including [`Popover`](https://ariakit.com/reference/popover), [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), [`SelectPopover`](https://ariakit.com/reference/select-popover), `CompositeOverflow`, [`Hovercard`](https://ariakit.com/reference/hovercard), [`Menu`](https://ariakit.com/reference/menu), and [`Tooltip`](https://ariakit.com/reference/tooltip), now close when focus or pointer interactions move across same-origin iframe boundaries. The outside target retains focus, and [`hideOnInteractOutside`](https://ariakit.com/reference/dialog#hideoninteractoutside) receives the cross-document event with the iframe host as the target when browsers omit a native `focusin` event.
```
diegohaz: **[P2] Do not advertise `CompositeOverflow` in the `@ariakit/react` release note**

This shared changeset versions both packages, but `CompositeOverflow` is exported only by `@ariakit/react-components`; `packages/ariakit-react/src/composite.ts` and the outer package index do not expose it. A consumer following the generated `@ariakit/react` changelog cannot use the named API:

```tsx

<Ariakit.CompositeOverflow />; // no such export
```

Remove `CompositeOverflow` from this shared sentence, or split out a `@ariakit/react-components`-only note if it needs to be named.

**Disposition:** Address now

**Assessment**
- Severity: Low (estimate, 95% confidence)
- Priority: P2 (estimate, 95% confidence)
- Breakage likelihood: Likely for release metadata (estimate, 90% confidence); every generated `@ariakit/react` changelog would carry the unavailable name
- Fix effort: Easy (estimate, 99% confidence); this is one prose edit or a package-specific split

diegohaz (commented):
## Re-review summary

Reviewed the complete PR again at head `a069807ae208a2d8747017c28cab539194647a72`, including the force-pushed lifecycle implementation, all 12 changed files, every resolved review thread, current CI, and the public `@ariakit/react` behavior in real browsers.

The earlier equal-ID, default iframe Tab, ordinary late-frame click, node-replacement, attribution, and cross-origin exception findings are resolved. I found four remaining items to address:

- **P1 / Medium:** the iframe-focus fallback gives `hideOnInteractOutside` an untrusted, undispatched event; a public `Popover` that accepts only trusted interactions stays open after a real keyboard Tab.
- **P1 / Medium:** React-state-driven subscription refresh loses focus performed synchronously by an iframe's load handler after navigation.
- **P1 / Medium:** navigation cleanup can leave outside-interaction listeners on a retained historical frame Document.
- **P2 / Low:** the shared `@ariakit/react` changeset names `CompositeOverflow`, which that package does not export.

### Browser evidence

Public callback fidelity, reproduced in Chromium, Firefox, and WebKit:

Iframe keyboard focus is reported as untrusted and the public Popover remains open

Load-time navigation focus race, reproduced in Chromium and WebKit:

Navigated editor input is focused while the public Popover remains open

### Verification

- Current checked-in focused browser suites: 48/48 passed across Chromium, Firefox, and WebKit.
- Focused React tests: 18/18 passed; `pnpm tsc`, `oxfmt --check`, `oxlint`, and `pnpm build-app-lite` passed.
- Public-API probes: untrusted callback event reproduced in all three engines; load-time focus race reproduced in Chromium/WebKit; retained-old-Document listener reproduced in Firefox.
- All current PR checks are green, including React 18, Chrome, Firefox, and Safari.
- Worktree is clean after removing the temporary reproduction fixtures.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`ce61c5538`](https://github.com/ariakit/ariakit/commit/ce61c5538bb9d3e1a5ad0ad17d2338d2335adeab): the iframe host stays an internal classification target, while the fallback waits one task and passes the original trusted child-Window `focus` event to `hideOnInteractOutside`. The changeset and browser regression explicitly define its post-dispatch shape (`isTrusted === true`, `currentTarget === null`, empty `composedPath()`); Chrome/WebKit ordering coverage verifies a native child `focusin` wins.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`ce61c5538`](https://github.com/ariakit/ariakit/commit/ce61c5538bb9d3e1a5ad0ad17d2338d2335adeab): a synchronous private registry refreshes the concrete frame subscriptions during the capture-phase lifecycle signal, before consumer load handlers run. The Chrome/WebKit regression navigates an iframe whose `onLoad` synchronously focuses its new input and verifies the Dialog closes.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`ce61c5538`](https://github.com/ariakit/ariakit/commit/ce61c5538bb9d3e1a5ad0ad17d2338d2335adeab): each document listener now captures the exact concrete `Document` at registration and removes from that same object. The Firefox regression retains the old document, navigates the iframe, dispatches on the retained document, and verifies the Dialog stays open. Window focus remains attached to the stable WindowProxy and both registration and cleanup stay cross-origin guarded.

.changeset/300-dialog-iframe-interactions.md
diegohaz: Fixed in [`ce61c5538`](https://github.com/ariakit/ariakit/commit/ce61c5538bb9d3e1a5ad0ad17d2338d2335adeab): `CompositeOverflow` has been removed from the shared release note.

packages/ariakit-react-components/src/dialog/utils/__frame-events.ts
```diff
@@ -0,0 +1,152 @@
...
+    const onChange = () => refresh(true);
+
+    for (const win of getFrameWindows(scope)) {
+      const doc = win.document;
+      const view = doc.defaultView;
+      if (!view) continue;
+      const observer = new view.MutationObserver((records) => {
+        if (!hasFrameMutation(records)) return;
+        onChange();
+      });
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`c4511f954`](https://github.com/ariakit/ariakit/commit/c4511f9544b24be08f5c844449431bb4426111dd): the frame observer now detects the deepest focused iframe at the mutation checkpoint, installs listeners for its previously unobserved document, and recovers the missed Firefox focus with an explicit untrusted event created in the child realm. Added targeted same-task `srcdoc`, stalled-navigation, and no-duplicate descendant coverage in [`c4511f954`](https://github.com/ariakit/ariakit/commit/c4511f9544b24be08f5c844449431bb4426111dd).

diegohaz:
**Fixed**

- [`c4511f954`](https://github.com/ariakit/ariakit/commit/c4511f9544b24be08f5c844449431bb4426111dd) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634374368)): Recovered focus from same-task iframe insertion by subscribing the newly observed focused document before replaying the narrow Firefox child-realm fallback, with targeted `srcdoc`, stalled-navigation, and no-duplicate descendant coverage.

packages/ariakit-react-components/src/dialog/utils/__frame-events.ts
```diff
@@ -0,0 +1,241 @@
...
+function getFrameWindows(scope: Window) {
+  const windows: Window[] = [];
+
+  const collect = (win: Window) => {
+    try {
+      if (!win.document) return;
+      windows.push(win);
+      for (const frame of Array.from(win.frames)) {
+        collect(frame);
+      }
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because shadow-root iframe discovery is outside the narrowed existing-frame contract. Browsers omit those frames from `window.frames`, and supporting them would require separate shadow-root traversal and lifecycle observation machinery.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -58,54 +125,175 @@ function isMouseEventOnDialog(event: Event | MouseEvent, dialog: Element) {
...
+  }
+  // We need to check if the content element has been focused at least once
+  // before checking if it's marked. This is so hovercards and tooltips don't
+  // stay open when new nodes are added to the DOM and focused.
+  if (
+    focused &&
+    targetInContentDocument &&
+    !isElementMarked(targetInContentDocument, contentElement.id)
+  ) {
+    return false;
```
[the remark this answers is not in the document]
diegohaz: Superseded by [`eb8b35928`](https://github.com/ariakit/ariakit/commit/eb8b359281fcd05d32fb349a36e6aa2415ae48a3): the observer and registry that subscribed late frames were removed. The simplified implementation snapshots the existing readable same-origin frame tree when the popup opens; frame insertion or navigation after open is intentionally unsupported.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -145,7 +341,122 @@ export function useHideOnInteractOutside(
...
+      // Fix for https://github.com/ariakit/ariakit/issues/619
+      if (event.target === getDocument(contentElement)) return;
+      if (!shouldHideOnInteractOutside(hideOnInteractOutside, event)) return;
+      if (
+        interactedOutsideRef &&
+        (focusEnteredFrame ||
+          (isElement(event.target) &&
+            getDocument(event.target) !== getDocument(contentElement)))
+      ) {
+        interactedOutsideRef.current = true;
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because preserving focus when frame entry is observable only through the iframe host or child `Window` is outside the narrowed contract. The simplified implementation preserves focus for native cross-document `focusin` targets and deliberately removes inferred or synthetic iframe-entry recovery.

diegohaz:
Reassessed the implementation in [`eb8b35928`](https://github.com/ariakit/ariakit/commit/eb8b359281fcd05d32fb349a36e6aa2415ae48a3) and narrowed it to the existing readable same-origin frame tree.

**Addressed**

- [`99fc772f7`](https://github.com/ariakit/ariakit/commit/99fc772f720fbfe561e55e456fd8243c4c2b74b2) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634863983)): Reused the frame-chain inside classification for the original mousedown, preserving drag suppression for parent disclosures and other inside targets across documents. The regression passes in Chrome, Firefox, and Safari.

**Declined**

- `non-goal` ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634796939)): Kept shadow-root iframe discovery outside the narrowed contract because it requires a separate traversal and observation subsystem.
- `non-goal` ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634799229)): Kept Window-only/direct iframe-host focus recovery outside the native cross-document focus contract.
- `non-goal` ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634885180)): Kept matching Dialog IDs across documents outside the narrowed contract because collision-safe ownership requires the marker-identity redesign removed during simplification.

**Obsolete**

- [`eb8b35928`](https://github.com/ariakit/ariakit/commit/eb8b359281fcd05d32fb349a36e6aa2415ae48a3) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634798385)): Removed late-frame observation; insertion and navigation after open are now explicitly unsupported.

The replacement keeps native document events, removes the observer/registry/synthetic-focus machinery, and has focused coverage in Chrome, Firefox, and Safari.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -163,7 +210,17 @@ export function useHideOnInteractOutside(
...
       // - https://github.com/ariakit/ariakit/issues/2330
-      if (!isElementMarked(previousMouseDown, contentElement?.id)) return;
+      if (!contentElement) return;
+      const previousMouseDownInContentDocument = getFrameHostInDocument(
+        previousMouseDown,
+        getDocument(contentElement),
+      );
+      if (
+        previousMouseDownInContentDocument &&
+        !isElementMarked(previousMouseDownInContentDocument, contentElement.id)
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`99fc772f7`](https://github.com/ariakit/ariakit/commit/99fc772f720fbfe561e55e456fd8243c4c2b74b2): the click guard now applies the same frame-chain inside classification to the original mousedown before projecting it into the content document. This preserves drag suppression for parent disclosures, focus traps, persistent elements, and nested content without changing the static frame scope. Added a real parent-disclosure drag regression; it failed in Chrome, Firefox, and Safari before the fix and passes in all three after it.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -46,6 +73,18 @@ function isDisclosure(disclosure: Element | null, target: Element) {
...
 
+function isFrameChainInside(
+  frameChain: Element[],
+  contentElement: Element,
+  disclosureElement: Element | null,
+) {
+  return frameChain.some((target) => {
+    if (isDisclosure(disclosureElement, target)) return true;
+    if (target.hasAttribute("data-focus-trap")) return true;
+    return isElementInside(target, contentElement.id);
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because matching Dialog IDs across documents are an explicit unsupported case after the complexity reassessment. Supporting this requires a per-dialog marker-identity redesign across the tree-marking system, which was intentionally removed with the observer-era machinery. The base implementation did not observe sibling-document events at all, so this does not regress prior behavior; the simplified contract covers the normal existing-frame case without claiming collision-safe IDs across documents.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -46,6 +73,18 @@ function isDisclosure(disclosure: Element | null, target: Element) {
   return false;
 }
 
+function isFrameChainInside(
+  frameChain: Element[],
+  contentElement: Element,
+  disclosureElement: Element | null,
+) {
+  return frameChain.some((target) => {
+    if (isDisclosure(disclosureElement, target)) return true;
```
diegohaz: **[P1] Resolve `aria-activedescendant` in the target document**

`isFrameChainInside` now passes elements from every readable document to `isDisclosure`, but this helper resolves the target's document-scoped `aria-activedescendant` ID in the disclosure's document. If an outside sibling-frame widget reuses the disclosure's ID—which is valid across separate documents—the outside target is mistaken for the disclosure and the popup stays open.

I reproduced this on the exact head in Chromium, Firefox, and WebKit. An embedded Combobox used `id="shared-active"`; focusing this otherwise-outside sibling-frame input left `Suggestions` visible, while the no-collision control closed it:

```html
<input aria-activedescendant="shared-active" />
<div id="shared-active" role="option">Local option</div>
```

Independently mounted host/iframe editors routinely reuse deterministic or generated IDs. Resolve the IDREF in the target's realm instead:

```ts
const activeElement = getDocument(target).getElementById(activeId);
```

Outside sibling-frame combobox is focused while the embedded Suggestions popup incorrectly remains visible

**Assessment**

- Severity: Medium (estimate, 90% confidence).
- Priority: P1 (estimate, 90% confidence).
- Breakage likelihood: Unlikely (estimate, 80% confidence); it requires a matching ID and an outside active-descendant target.
- Fix effort: Easy (estimate, 95% confidence); one realm-correct lookup plus a browser regression.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -28,6 +28,33 @@ interface EventOutsideOptions {
...
+}
+
+function getHighestReadableWindow(element: Element) {
+  const chain = getFrameChain(element);
+  return getWindow(chain[chain.length - 1]);
+}
+
 function isInDocument(target: Element) {
   if (target.tagName === "HTML") return true;
   return contains(getDocument(target).body, target);
```
diegohaz: **[P1] Handle frame documents without `body`**

The expanded listener tree can now call `isInDocument` for elements in an existing same-origin SVG/XML frame, where `Document.body` is `null`. Clicking that frame throws here before the outside callback runs, so the popup remains stale.

I reproduced this on the exact head in Chromium, Firefox, and WebKit with an embedded Combobox popup and a sibling `/preview.svg`: clicking its `<rect>` focused the SVG but left `Suggestions` visible. All three engines reported a null-`contains` `TypeError`; the ordinary sibling HTML input control closed correctly. Diagram/CMS previews commonly embed SVG as a document rather than inline markup.

A document-type-neutral connectivity check preserves the detached-target guard:

```ts
function isInDocument(target: Element) {
  return target.isConnected;
}
```

[Exact-head recording: clicking the blue SVG target leaves Suggestions open](https://github.com/user-attachments/assets/cbccc869-fdd6-405a-b157-5c225aaf5518)

**Assessment**

- Severity: Medium (estimate, 90% confidence).
- Priority: P1 (estimate, 85% confidence).
- Breakage likelihood: Rare (estimate, 95% confidence); it requires a non-HTML same-origin frame document.
- Fix effort: Easy (estimate, 95% confidence); a local helper change plus a focused browser regression.

.changeset/300-dialog-iframe-interactions.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Close popups across existing same-origin iframe boundaries
+
+The [`Dialog`](https://ariakit.com/reference/dialog), `Popover`, `ComboboxPopover`, `Hovercard`, `Menu`, `SelectPopover`, and `Tooltip` components now close when focus moves from an embedded same-origin popup to an ancestor document or a pointer interaction occurs in an existing sibling frame. Focus stays on the outside target, interactions in contained frames stay inside, and true browser or application window blur remains ignored.
```
diegohaz: **[P2] Link every affected public API**

The Ariakit changeset policy requires each public API occurrence to link to its reference page, but only `Dialog` remains linked on the current head. Both generated package changelogs would therefore publish the other six affected APIs as non-navigable names.

Please restore the links, for example:

```md
[`Popover`](https://ariakit.com/reference/popover), [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), [`Hovercard`](https://ariakit.com/reference/hovercard), [`Menu`](https://ariakit.com/reference/menu), [`SelectPopover`](https://ariakit.com/reference/select-popover), and [`Tooltip`](https://ariakit.com/reference/tooltip)
```

All six deployed reference URLs return HTTP 200. This was fixed earlier in the PR and regressed when the implementation/release note was simplified.

**Assessment**

- Severity: Low (estimate, 99% confidence).
- Priority: P2 (estimate, 99% confidence).
- Breakage likelihood: Likely for release metadata (verified); both package changelogs emit the unlinked names.
- Fix effort: Easy (verified); this is a one-line Markdown edit and all target URLs are live.

diegohaz (commented):
## Re-review summary

Re-reviewed the complete PR at `99fc772f720fbfe561e55e456fd8243c4c2b74b2`, using the latest marked review at `a069807ae208a2d8747017c28cab539194647a72` as the incremental baseline while also checking the full five-file diff and surrounding Dialog behavior. I audited all 18 prior review threads; all were resolved before this review, with no incomplete comment pagination.

The final drag-suppression commit resolves the previous cross-document drag candidate. I found three current-head items to address:

- **[P1 / Medium: resolve `aria-activedescendant` in the target document](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908148).** Document-local ID reuse in an outside sibling-frame widget can make it look like the disclosure and leave the embedded popup open.
- **[P1 / Medium: handle frame documents without `body`](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908414).** Clicking an existing same-origin SVG/XML frame throws before hiding the popup.
- **[P2 / Low: link every affected public API in the changeset](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908647).** Six affected component names are non-navigable in both generated package changelogs.

All three are **Address now**: the runtime findings are inside the PR's existing readable same-origin frame-tree contract, and the release-note fix is required by the repository changeset policy. No follow-up issue is more appropriate.

### Verification

- Exact-head public-behavior probes reproduced both runtime findings in Chromium, Firefox, and WebKit. The ordinary sibling-input controls closed correctly in every engine.
- The checked-in 5-test browser suite is green in Chrome, Firefox, and Safari/WebKit (15 cases total); every current GitHub check is green.
- The new drag case is red on `eb8b359281fcd05d32fb349a36e6aa2415ae48a3` and green on `99fc772f720fbfe561e55e456fd8243c4c2b74b2` in Chromium.
- A two-level readable-frame wrapper and cross-document `contextmenu` passed manual Chromium probes. They are not independently automated across all three engines.
- `git diff --check` passed and the worktree is clean.

### Scope notes

I did not treat the documented cross-origin, late/navigated-frame, missing-native-focus, shadow-root frame-enumeration, or `window.open()` cases as findings because the PR explicitly excludes them. I also retained the contained-frame reopen choreography: it makes the frame part of the static listener snapshot before the assertion, so removing it would weaken the regression rather than simplify it.

Pre-fix WebKit red proof for the original focus regression, and pre-fix Firefox/WebKit red proof for the newly added drag case, were not independently established; current exact-head behavior is covered and green in those engines.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`a8824de`](https://github.com/ariakit/ariakit/commit/a8824de9eec027e6eaa7f3c50545280aa2a928af): `aria-activedescendant` now resolves through `getDocument(target)`, keeping equal IDs scoped to the outside target’s document. Added targeted Chrome, Firefox, and Safari/WebKit coverage in the same commit.

[Cross-browser recording: document-local ID reuse and the bodyless SVG case](https://github.com/user-attachments/assets/a4ad802e-b5bb-49f9-8bea-d29ce43d3e02)

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`a8824de`](https://github.com/ariakit/ariakit/commit/a8824de9eec027e6eaa7f3c50545280aa2a928af): the detached-target guard now uses `target.isConnected`, so SVG/XML frame documents without a `body` are handled without dereferencing `null`. Added targeted Chrome, Firefox, and Safari/WebKit coverage in the same commit.

[Cross-browser recording: document-local ID reuse and the bodyless SVG case](https://github.com/user-attachments/assets/a4ad802e-b5bb-49f9-8bea-d29ce43d3e02)

.changeset/300-dialog-iframe-interactions.md
diegohaz: Fixed in [`61502fa`](https://github.com/ariakit/ariakit/commit/61502fa2acf717d695855bb61a254d6a48e532ed): restored verified reference links for all six affected public APIs.

diegohaz:
**Fixed**

- [`a8824de`](https://github.com/ariakit/ariakit/commit/a8824de9eec027e6eaa7f3c50545280aa2a928af) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635039361)): Resolve active-descendant IDs in the target document and add cross-browser regression coverage.
- [`a8824de`](https://github.com/ariakit/ariakit/commit/a8824de9eec027e6eaa7f3c50545280aa2a928af) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635039525)): Use document-neutral connectivity for bodyless SVG/XML frames and add cross-browser regression coverage.
- [`61502fa`](https://github.com/ariakit/ariakit/commit/61502fa2acf717d695855bb61a254d6a48e532ed) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635039695)): Restore verified reference links for all affected public APIs.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -163,7 +221,26 @@ export function useHideOnInteractOutside(
       // outside of it). See:
       // - https://github.com/ariakit/ariakit/issues/1336
       // - https://github.com/ariakit/ariakit/issues/2330
-      if (!isElementMarked(previousMouseDown, contentElement?.id)) return;
+      if (!contentElement) return;
+      if (
+        isFrameChainInside(
+          getFrameChain(previousMouseDown),
```
diegohaz: ### [P1] Guard the recorded `mousedown` target before frame traversal

`usePreviousMouseDownRef` records `EventTarget | null`, but this new call passes the value to `getFrameChain` after an unchecked cast. The existing `dialog-5156` regression deliberately treats third-party document events whose target is a bare `EventTarget` as supported. If third-party code dispatches that kind of `mousedown` and a consumer later triggers a standalone outside `click` (`element.click()` or `fireEvent.click`, so no real `mousedown` replaces the value), `isFrameChainInside` reaches `Node.contains(nonNode)`, throws, and leaves the Dialog open.

[This exact-head recording](https://github.com/user-attachments/assets/e6171b81-0aa4-40c2-9a49-3201d653eeb5) shows the stale Dialog and the uncaught Chromium error. I reproduced the same sequence in Chromium, Firefox, and WebKit; each raised its engine-specific `Node.contains` `TypeError` and kept the Dialog visible. A physical pointer click usually masks the issue because its Element `mousedown` overwrites the ref.

Please reject foreign targets before traversing them, then extend `dialog-5156` with the foreign-`mousedown` → programmatic-outside-`click` sequence:

```ts
const previousMouseDown = previousMouseDownRef.current;
if (!isElement(previousMouseDown)) return;
```

**Assessment**

- **Severity:** Medium (estimate, 90% confidence) — the supported sequence raises an uncaught handler error and leaves the popup stale.
- **Priority:** P1 (estimate, 95% confidence) — this regresses the behavior already protected by #5156/#6087.
- **Breakage likelihood:** Rare (estimate, 90% confidence) — it requires a non-Element synthetic `mousedown` followed by a click without a replacement Element `mousedown`.
- **Fix effort:** Easy (estimate, 99% confidence) — one realm-neutral guard plus a focused browser regression.

app/src/sandbox/dialog-iframe-interactions/test-browser.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  // https://github.com/ariakit/ariakit/issues/3250
+  test("keeps the popover open when clicking an existing contained iframe", async ({
+    page,
+    q,
+  }) => {
+    const containedFrame = query(
+      page.frameLocator("iframe[title='Contained frame']"),
+    );
+    // Mount the frame before reopening so it's part of the listener snapshot.
```
diegohaz: ### [P2] Remove the stale close/reopen choreography

`RootPopover` uses `unmountOnHide={false}`, so the contained iframe host is already connected before the first open; closing and reopening does not make it newly eligible for the static listener snapshot. I verified this on the exact head in Chromium, Firefox, and WebKit: the iframe count was `1` and `isConnected` was `true` before opening, and a single open followed by the contained-frame click kept the popover visible in every engine.

The extra transitions therefore teach a false lifecycle requirement and add unrelated asynchronous state changes to the regression. This shorter sequence preserves the behavior under test:

```ts
await q.button("Open root popover").click();
await test.expect(q.dialog("Root popover")).toBeVisible();

await containedFrame.button("Contained frame target").click();
```

**Assessment**

- **Severity:** Low (estimate, 95% confidence) — product behavior is unaffected, but the test documents the static-snapshot contract incorrectly and carries avoidable flake surface.
- **Priority:** P2 (estimate, 95% confidence) — the regression should describe the actual prerequisite before it becomes maintenance guidance.
- **Breakage likelihood:** Unlikely (estimate, 90% confidence) — the impact is test maintenance rather than a runtime failure, triggered when this setup is copied or debugged as authoritative.
- **Fix effort:** Easy (verified on the exact head) — removing the stale comment and three close/reopen lines passed the equivalent single-open probe in all three engines.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -28,24 +28,62 @@ interface EventOutsideOptions {
...
+
+function getHighestReadableWindow(element: Element) {
+  const chain = getFrameChain(element);
+  return getWindow(chain[chain.length - 1]);
+}
+
 function isInDocument(target: Element) {
-  if (target.tagName === "HTML") return true;
-  return contains(getDocument(target).body, target);
+  return target.isConnected;
```
diegohaz, line 42: ### [P2] Record an explicit issue severity in Cycle 9

The live PR description's Cycle 9 reassessment records this realm-correctness fix and the active-descendant fix, but it never states their issue severity. The reassessment policy requires every block to include severity alongside frequency, scope, complexity, and direction, so the otherwise detailed audit record is incomplete.

Please add an explicit sentence such as:

> **Issue severity:** Medium — affected users see a stale popup, and the SVG/XML path also raises an uncaught exception.

**Assessment**

- **Severity:** Low (estimate, 95% confidence) — this is an audit/process defect, not a product-behavior defect.
- **Priority:** P2 (estimate, 95% confidence) — Cycle 9 is the current mandatory reassessment block and should satisfy its required fields before merge.
- **Breakage likelihood:** Likely (verified against the live PR body) — the required severity field is currently absent, so the published reassessment remains incomplete until edited.
- **Fix effort:** Easy (verified) — one explicit severity sentence in the PR description.

diegohaz (commented):
## Re-review summary

Re-reviewed the complete PR at `61502fa2acf717d695855bb61a254d6a48e532ed`, using the latest marked review at `99fc772f720fbfe561e55e456fd8243c4c2b74b2` as the incremental baseline while also checking the full merge-base range from `d641496b1e84395f687bb2cca3a4ef7b006047b5` and the surrounding Dialog behavior. I audited all 21 prior review threads; all were resolved before this review, with no incomplete comment pagination or pre-existing unresolved candidate to reply to.

The three findings from the prior marked review are resolved on this head:

- [`aria-activedescendant` document ownership](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908148): fixed in `a8824de` with document-local lookup and cross-browser regression coverage.
- [Bodyless SVG/XML frame handling](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908414): fixed in `a8824de` with document-neutral connectivity and cross-browser regression coverage.
- [Affected API links in the changeset](https://github.com/ariakit/ariakit/pull/6805#discussion_r3634908647): fixed in `61502fa` with verified public-reference links.

I found three current-head items:

- [P1 / Medium: guard non-Element recorded `mousedown` targets](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635170647). A third-party synthetic target followed by a standalone outside click throws in `Node.contains` and leaves the Dialog open.
- [P2 / Low: remove the contained-frame close/reopen choreography](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635171106). The iframe is already connected before the first open, and the single-open alternative is green in all three engines.
- [P2 / Low: add explicit issue severity to Cycle 9](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635171744). The live reassessment records frequency, scope, complexity, and direction but omits its required severity field.

These are all current-PR fixes; no durable follow-up issue is safer or more proportionate.

### Verification

- Built the exact head and ran `app/src/sandbox/dialog-iframe-interactions/test-browser.ts` in Chrome, Firefox, and Safari/WebKit on isolated checkout-owned ports: **21/21 passed** (7 per engine).
- Reproduced the non-Element-`mousedown` failure on the exact local build in Chromium, Firefox, and WebKit. Each engine emitted its `Node.contains` type error and left the Dialog visible. The linked recording captures the Chromium sequence; Firefox and WebKit were verified in the terminal rather than separately recorded.
- Verified in all three engines that the contained iframe host is connected before the first open and that one open followed by the contained-frame click leaves the popover visible.
- `git diff --check` passed for both the incremental and full PR ranges; the review checkout is clean apart from temporary review-body files, which are removed after publication.
- Every current GitHub check is green, including Chrome, Firefox, Safari, type checks, builds, and performance jobs.

### Scope notes

I retained the PR's documented cross-origin, late/navigated-frame, missing-native-focus, shadow-root frame-enumeration, `window.open()`, and cross-document Dialog-ID cases as declined non-goals. Cycle 9's static frame-tree direction remains proportionate; the new runtime finding needs a local target guard rather than an architectural change.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`45e18f8`](https://github.com/ariakit/ariakit/commit/45e18f861caa613e54ba821eda5e385b870998e2): recorded `mousedown` targets are now validated with the realm-neutral `isElement` guard before frame traversal. Added targeted Chrome, Firefox, and Safari/WebKit coverage in the same commit.

[Exact-head recording: foreign `mousedown` → standalone outside click leaves the Dialog open with zero page errors](https://github.com/user-attachments/assets/dfe63d10-b216-4214-a774-ec1f7fe09b22)

app/src/sandbox/dialog-iframe-interactions/test-browser.ts
diegohaz: Fixed in [`33bfcd7`](https://github.com/ariakit/ariakit/commit/33bfcd723a7a913fe8dc80d8f49d09981ec72bf7): removed the stale close/reopen choreography. The contained iframe is already connected under `unmountOnHide={false}`, and the single-open test still exercises contained-frame classification. The focused sandbox remains green in Chrome, Firefox, and Safari/WebKit (21/21 cases).

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in the current PR description: Cycle 9 now explicitly records **Medium** issue severity and the stale-popup/uncaught-exception impact.

diegohaz:
**Fixed**

- [`45e18f8`](https://github.com/ariakit/ariakit/commit/45e18f861caa613e54ba821eda5e385b870998e2) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635535085)): Guard recorded foreign `mousedown` targets before frame traversal and add cross-browser regression coverage.
- [`33bfcd7`](https://github.com/ariakit/ariakit/commit/33bfcd723a7a913fe8dc80d8f49d09981ec72bf7) ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635535717)): Remove the unnecessary contained-frame close/reopen choreography.
- `PR description` ([reply](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635536120)): Record the explicit Cycle 9 issue severity and impact.

diegohaz (commented):
## Re-review summary

Re-reviewed the complete PR at [`33bfcd7`](https://github.com/ariakit/ariakit/commit/33bfcd723a7a913fe8dc80d8f49d09981ec72bf7), using the latest marked review at [`61502fa`](https://github.com/ariakit/ariakit/commit/61502fa2acf717d695855bb61a254d6a48e532ed) as the incremental baseline while also checking the full merge-base range from [`d641496`](https://github.com/ariakit/ariakit/commit/d641496b1e84395f687bb2cca3a4ef7b006047b5) and the surrounding Dialog behavior. I audited all 24 review threads with complete outer and nested pagination; all are resolved, with no pre-existing unresolved candidate to reply to.

The three findings from the prior marked review are resolved:

- [Foreign recorded `mousedown` targets](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635170647): fixed in [`45e18f8`](https://github.com/ariakit/ariakit/commit/45e18f861caa613e54ba821eda5e385b870998e2). The runtime now validates the stored target with the realm-neutral `isElement` helper before calling frame and containment APIs, and the regression covers the synthetic-target sequence in all three engines.
- [Redundant contained-frame test setup](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635171106): fixed in [`33bfcd7`](https://github.com/ariakit/ariakit/commit/33bfcd723a7a913fe8dc80d8f49d09981ec72bf7). The test now proves the same behavior with one open and one contained-frame click.
- [Missing Cycle 9 issue severity](https://github.com/ariakit/ariakit/pull/6805#discussion_r3635171744): fixed in the current PR description with an explicit Medium assessment and user-visible impact.

No new substantive finding remains on the current head, and no durable follow-up is needed.

### Verification

- `pnpm build-app-lite`: passed on the exact head.
- `APP_PORT=14324 NEXTJS_PORT=14301 APP_INSPECTOR_PORT=19441 NEXTJS_INSPECTOR_PORT=19442 pnpm -F app test src/sandbox/dialog-5156/test-browser.ts src/sandbox/dialog-iframe-interactions/test-browser.ts --project=chrome --project=firefox --project=safari --workers=1`: **24/24 passed**.
- The author-provided [cross-browser recording](https://github.com/user-attachments/assets/dfe63d10-b216-4214-a774-ec1f7fe09b22) shows the foreign-target regression sequence and all three successful engine runs.
- `git diff --check` passed for both the incremental and full PR ranges, and the review checkout is clean.
- Every current GitHub check is green, including Chrome, Firefox, Safari, type checks, builds, CodeQL, deployment, and performance jobs.

### Scope and reassessment

The documented cross-origin, late or navigated frame, missing-native-focus, shadow-root frame-enumeration, `window.open()`, and cross-document Dialog-ID cases remain explicit non-goals. The two incremental fixes reduce risk and test complexity without expanding that scope. This clean re-review does not trigger the next scheduled complexity reassessment; the static readable-frame-tree approach remains proportionate to the confirmed same-origin behavior.

PR #6806, opened 2026-07-22
Add scrollElement to collection renderers

### LINKED ISSUE #3913: [SelectRenderer] - Allow to explictly declare scroller element
### Motivation

When using `SelectRenderer`, if the intended scroller element start without a scrollbar, ariakit defaults to to observing the document, and virtual scroll isn't working.

Notice that only when resizing the window, the items appear:
https://github.com/ariakit/ariakit/assets/40234776/a408b5cf-b546-4c9a-8812-ff6dff604683

### Usage example

https://stackblitz.com/edit/u17aat?file=select-combobox-virtualized%2Fstyle.css,select-combobox-virtualized%2Fcountries.ts,select-combobox-virtualized%2Findex.tsx&theme=dark

### Requirements

add an optional `scrollerRef` prop to `SelectRenderer`

### Workaround

_No response_

### Possible implementations

_No response_

diegohaz:
## Motivation

`SelectRenderer` automatically detects its scrolling element when it mounts. When an `overflow: auto` ancestor is initially empty, it does not overflow yet, so the renderer caches the document instead and does not respond when the ancestor later gains items.

Fixes #3913.

## Solution

Add a `scrollElement` prop to collection renderers, including `CompositeRenderer` and `SelectRenderer`. It accepts the scrolling element directly, a React ref, or a resolver function:

```tsx
const scrollElementRef = useRef<HTMLDivElement>(null);

<div ref={scrollElementRef} className="scroller">
  <SelectRenderer scrollElement={scrollElementRef} items={items}>
    {(item) => <SelectItem key={item.id} {...item} />}
  </SelectRenderer>
</div>
```

Omitting the prop preserves automatic scrolling-ancestor detection when neither the renderer nor a same-store ancestor provides a value. Same-store descendants inherit an explicit ancestor value unless they override it. A value that resolves to `null` disables viewport-driven rendering.

The browser regressions cover an ancestor ref attached during the initial commit, a stable ref that connects after mount, same-store nested inheritance, scrolling after asynchronous items load, switching to `null` without reusing stale viewport offsets, and a Chrome GC check that an explicit resolver target is released after returning to automatic mode.

## Preview

Initial ancestor ref virtualizing items 51 through 55

[Initial ancestor ref scrolling from item 1 to item 51 (video)](https://github.com/user-attachments/assets/423f65ca-0798-4b7b-a659-594a5d5803eb)

## Workaround

Until the fix is available in a release, remount `SelectRenderer` when the data crosses the empty-to-populated boundary so automatic scrolling-element detection runs after the intended ancestor overflows:

```tsx
<SelectRenderer
  // TODO: Remove this key after https://github.com/ariakit/ariakit/issues/3913 is fixed.
  key={items.length ? "loaded" : "empty"}
  items={items}
>
  {(item) => <SelectItem {...item} />} 
</SelectRenderer>
```

This is a boundary-transition workaround: it assumes the intended ancestor starts overflowing when items first become non-empty. It will not help if overflow appears only after a later non-empty update, and remounting resets renderer-local DOM state or focus.

## Acknowledgments

Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi) for reporting the issue, providing the video and StackBlitz reproduction, and proposing an explicit scroller prop.

<details>
<summary>Cycle 3: Continue the narrow explicit-scroller design</summary>

**Assessment**

Three finding-bearing cycles exposed lifecycle ordering for explicit `null`, layout-path performance, omitted nested inheritance, direct element discrimination, and a ref attached by an ancestor during the initial commit. Direct elements, refs, resolver functions, `null`, automatic fallback, and explicit same-store inheritance are all supported parts of the public API. The initial ancestor-ref form is a common documented use case with high user impact when broken; the remaining findings concern low-impact release wording, contributor credit, and this maintenance record.

Supporting that scope requires an input discriminator, a layout snapshot, passive re-resolution, committed scroller state plus a live-ref stale-work guard, same-store inheritance, regression tests, and public documentation. A ref-only contract would still require the dual layout/passive timing and stale-target guard because descendant layout effects can run before an ancestor host ref attaches. Direct-element and resolver support add only discrimination before the shared resolution path, while inheritance avoids requiring consumers to repeat the same target through nested renderers.

**Decision**

Continue the current design rather than narrowing the accepted inputs. Re-resolve explicit targets in the passive synchronization effect after ancestor refs and layout effects have attached, while retaining the layout snapshot for the stale-work guard. Keep automatic detection local unless a same-store ancestor explicitly provides a value. Explicit targets must remain scrolling ancestors in the same document; non-ancestor and cross-document targets are intentionally unsupported. No follow-up issue is required.

</details>

<details>
<summary>Cycle 6: Keep the API scope and bound the controller lifetime</summary>

**Assessment**

The next three finding-bearing cycles exposed commit ordering between inherited renderers, publication of a revalidated target before descendant passive work, and retention of the last explicit resolver and element after `scrollElement` returns to automatic mode. These findings are concentrated in the lifecycle of the controller that coordinates an explicit target across same-store descendants; they do not make the accepted input forms independently complex.

A ref-only contract would still need commit-phase revalidation because a descendant layout effect can run before an ancestor host ref attaches. Removing same-store inheritance would eliminate that coordination but retract documented behavior and force nested renderers to repeat the same target. The narrower fix is to scope the controller to explicit mode: initialize it without capturing the render-time resolver, publish the resolver only for a committed explicit layout, and release the resolver, published element, and controller ref when automatic mode commits.

**Decision**

Keep direct elements, refs, resolver functions, `null`, automatic fallback, and explicit same-store inheritance. Bound the controller lifetime to committed explicit mode. A Chrome-only regression stores the explicit resolver target in a `WeakRef`, returns to automatic mode, performs six committed renderer rerenders, requests `HeapProfiler.collectGarbage`, and verifies that the target was released. Require fresh renderer performance checks for the final exact head. No follow-up issue is required.

</details>


packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -465,6 +512,10 @@ export function useCollectionRenderer<T extends Item = any>({
   const parentData = parent?.childrenData;
   const orientation = orientationProp ?? parent?.orientation ?? "vertical";
   const overscan = overscanProp ?? parent?.overscan ?? 1;
+  const inheritedScroller =
```
diegohaz, line 581: **[P1] Preserve automatic detection for omitted nested renderers**

Before this PR, every renderer that omitted `scrollElement` detected its own closest scrolling ancestor. These lines now make an omitted same-store child inherit the parent even when the parent also omitted the prop. An existing nested renderer with an independently scrollable viewport therefore listens to the outer viewport; scrolling the inner element leaves Items 1–7 at the top and makes the visible viewport blank at `scrollTop = 2000`.

For example, this previously auto-detected the inner scroller:

```tsx
<CollectionRenderer store={store} items={groups}>
  {(group) => (
    <div className="inner-scroller">
      <CollectionRenderer store={store} items={group.items} />
    </div>
  )}
</CollectionRenderer>
```

Please distinguish an ancestor that explicitly supplied `scrollElement` (including explicit `null`) from one that auto-detected it. Only the explicit mode should be inherited; an omitted child under an omitted parent should keep local automatic detection. The independently scrollable nesting case should also be covered by a browser regression. If unconditional inheritance is intentional, this is a breaking default-behavior change rather than a patch feature.

[Exact-head reproduction at `1b12aa4`](https://github.com/user-attachments/assets/5e512566-4122-4c57-8353-f85885faf31f)

**Assessment**

- Severity: Medium (estimate, 90% confidence) — the affected nested panel renders a stale or blank window.
- Priority: P1 (estimate, 90% confidence) — this is a compatibility regression introduced by the PR.
- Breakage likelihood: Unlikely (estimate, 85% confidence) — it requires same-store nested renderers with distinct viewports or a portal.
- Fix effort: Moderate (estimate, 85% confidence) — propagate explicit-vs-automatic mode and add a browser case.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +297,58 @@ function getViewport(scroller: Element) {
...
 }
 
-function useScroller(rendererRef: RefObject<HTMLElement | null> | null) {
+function getScrollElement(
+  renderer: HTMLElement,
+  scrollElement: CollectionRendererOptions["scrollElement"],
+) {
+  if (scrollElement === undefined) return getScrollingElement(renderer);
+  if (typeof scrollElement === "function") return scrollElement(renderer);
+  if (scrollElement && "current" in scrollElement) {
```
diegohaz, line 316: **[P2] Distinguish direct elements before ref objects**

A direct `HTMLElement` is a supported input, but this check treats any value with a `current` property as a React ref. Custom elements can legitimately expose that property:

```tsx
class CarouselScroller extends HTMLElement {
  current = 0;
}

<CollectionRenderer scrollElement={carouselScroller} items={items} />
```

At runtime this branch returns `0` instead of the element, and the later offset path can throw when it evaluates `"scrollX" in scroller`. Please establish that the value is an element first—using the realm-neutral `isElement` helper—before applying the structural ref-object check, and add a direct-element regression.

**Assessment**

- Severity: Medium (estimate, 85% confidence) — a valid documented input can crash or disable virtualization.
- Priority: P2 (estimate, 85% confidence) — the shape is rare but the failure is deterministic.
- Breakage likelihood: Rare (estimate, 95% confidence) — it requires a direct custom element with a `current` member.
- Fix effort: Easy (estimate, 95% confidence) — reorder the discriminator and add one test.

Evidence gap: I verified the deterministic runtime branch on the exact head, but did not capture a standalone visual artifact for this rare custom-element shape.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +297,58 @@ function getViewport(scroller: Element) {
...
+    // Ref values and resolver results can change without their prop identity
+    // changing, so explicit targets are resolved after each commit. Keep
+    // automatic detection one-shot until the renderer reference or prop
+    // changes.
+    if (scrollElement === undefined && autoResolved.current) return;
+    const nextScroller = getScrollElement(renderer, scrollElement);
+    if (scrollElement === undefined) autoResolved.current = true;
+    scrollerRef.current = nextScroller;
+    if (nextScroller === scroller) return;
+    setScroller(nextScroller);
```
diegohaz, line 415: **[P2] Keep scroller state synchronization off the layout path**

`setScroller(nextScroller)` now runs inside `useSafeLayoutEffect`. In the mount benchmark, the outer renderer resolves its scroller and synchronously rerenders its provider plus 200 nested renderers before the interaction can paint. The exact-head final performance run confirms the impact: mount INP increased from 40 ms to 48 ms (+20%, +8 ms). All four baseline round medians were 40 ms; current medians were 48/48/56/48 ms, and the raw artifact separates completely (baseline 32–40 ms, current 48–64 ms).

Keep the layout-time `scrollerRef.current` update so stale passive work sees an identity mismatch, but reconcile `scroller` state in a passive effect. That preserves the new null/ref correctness guard without putting the provider rerender on the pre-paint path. This is especially relevant after fixing the nested-inheritance issue above: those 200 children currently inherit and skip their own resolution, so restoring local automatic detection could otherwise amplify the layout-time updates.

[Exact-head performance report](https://github.com/ariakit/ariakit/pull/6806#issuecomment-5048387300) · [raw four-round artifact](https://github.com/ariakit/ariakit/actions/runs/29951390626/artifacts/8542570937)

**Assessment**

- Severity: Low (verified, high confidence) — the regression is latency rather than functional breakage.
- Priority: P2 (verified, high confidence) — final CI classifies the INP increase as significant.
- Breakage likelihood: Possible (estimate, 90% confidence) — it is most visible when opening a large nested virtualized tree or menu.
- Fix effort: Moderate (estimate, 85% confidence) — split live-target invalidation from state synchronization and rerun perf.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -562,12 +605,18 @@ export function useCollectionRenderer<T extends Item = any>({
     }
   }, [elementsUpdated, itemSize, baseId, items, data, computeData]);
 
-  const scroller = useScroller(items ? ref : null);
+  const ownScroller = useScroller(
+    items && inheritedScroller === undefined ? ref : null,
+    scrollElementProp,
+  );
```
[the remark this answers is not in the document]
diegohaz: Addressed by `1b12aa4`. I rechecked the exact head: literal `null` is selected synchronously, ref/resolver outputs update `scrollerRef.current` during layout, and `processVisibleIndices` now skips passive work whenever that live target differs from committed `scroller`.

The updated browser cases batch disablement/ref disconnection with the 40 → 80 px item-size change, and the focused suite passes 15/15 across Chrome, Firefox, and Safari. The stale ~50 → ~25 item-window shift no longer reproduces.

diegohaz (commented):
Comment-only review of base `ab0f70f2` through head `1b12aa49`. The PR advanced while I was reviewing; I discarded the stale-head conclusions and revalidated the findings and evidence below against the current head.

## Address now

1. [P1: preserve automatic detection for omitted nested renderers](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633347050). Same-store children now inherit an auto-detected parent viewport, changing the previous default and making independently scrollable nested panels stale or blank. Exact-head browser reproduction is attached.
2. [P2: distinguish direct elements before ref objects](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633348139). A valid custom `HTMLElement` with a `current` member is treated as a ref and can pass a non-element into the offset path.
3. [P2: keep scroller state synchronization off the layout path](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633349001). Exact-head final perf confirms mount INP at 40 → 48 ms (+20%, +8 ms); the four-round raw artifact has baseline medians 40/40/40/40 and current 48/48/56/48.

## Existing thread disposition

- [The stale-offset finding is Addressed](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633350147) by `1b12aa4`. Literal `null`, detached refs, and resolver changes invalidate the live target before passive visibility work. The updated batched browser regressions pass on Chrome, Firefox, and Safari.

## Validation

- `pnpm test-react app/src/sandbox/select-renderer/test.ts`: 2 passed.
- Focused exact-head browser run with isolated ports: 15 passed across Chrome, Firefox, and Safari.
- `pnpm build-app-lite`: passed.
- Exact-head CI is green, including type, build, React/18, Solid, Chrome, Firefox, Safari, and perf.
- Inspected the final collection-renderer perf report, job revisions, and four-round raw artifact.
- `git diff --check` passes; the temporary evidence fixture was removed and the worktree is clean.

Review coverage included the complete five-file diff, renderer effect ordering and nested context, public option forwarding/types, existing renderer call sites, browser tests, changeset/package exports, reference generation, and the relevant performance benchmark.

## Candidate dispositions and gaps

- Widening `scrollElement` from `HTMLElement` to `Element`: **Declined non-goal**. The narrower public type matches existing Ariakit element-target APIs; the compiler friction alone did not establish an intended contract violation.
- A broad rerender/scroll performance concern: **Declined after final evidence**. Relevant final totals are clean (+3% rerender, +5% scroll); only the confirmed mount-INP regression above is published.
- The report's tooltip regression: **Not attributed**. The PR has no causal path to that benchmark.
- Direct-element and resolver branches lack committed browser coverage; the direct-element discriminator is captured by finding 2. I did not capture separate visual evidence for the rare custom-element `current` shape.
- I did not rerun the entire monorepo suite locally; exact-head CI provides that broader coverage.
- Registered follow-ups: none.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [5c985d157](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1).

The renderer context now propagates only an explicitly supplied same-store <code>scrollElement</code> chain, including explicit <code>null</code>. When both parent and child omit the prop, the child keeps local closest-ancestor detection. I added the requested [independently scrollable nested regression](https://github.com/ariakit/ariakit/blob/5c985d1571d7acd122166bfa35b878fd1aceb0c1/app/src/sandbox/select-renderer/test-browser.ts#L112-L128); the focused fixture passes 21/21 across Chrome, Firefox, and Safari.

Exact-head Chromium verification also reached item 51 at <code>scrollTop = 2000</code>. Visual publishing gap: the 7.48 s local capture was reviewed but is not attached to this thread; the linked browser test is the reproducible evidence.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [5c985d157](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1).

Direct elements are now identified with the realm-neutral <code>isElement</code> helper before the structural <code>current</code> ref check. The new [direct-element browser regression](https://github.com/ariakit/ariakit/blob/5c985d1571d7acd122166bfa35b878fd1aceb0c1/app/src/sandbox/select-renderer/test-browser.ts#L130-L148) uses an actual HTMLElement with its own <code>current: 0</code> property and passes across Chrome, Firefox, and Safari.

Exact-head Chromium verification confirmed <code>isHTMLElement = true</code>, <code>hasOwnCurrent = true</code>, <code>current = 0</code>, and item 51 visible at <code>scrollTop = 2000</code>. Visual publishing gap: the local capture was reviewed but is not attached to this thread; the linked browser test is the reproducible evidence.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [5c985d157](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1). The layout effect now updates only the live <code>scrollerRef.current</code>; <code>setScroller</code> and provider synchronization reconcile in the passive effect, while the live-ref mismatch guard still blocks stale work.

The [final exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29955693675) passed all 17 jobs, and the [updated report](https://github.com/ariakit/ariakit/pull/6806#issuecomment-5048387300) finds no significant Chrome or Node changes. In the [raw CollectionRenderer artifact](https://github.com/ariakit/ariakit/actions/runs/29955693675/artifacts/8544126615), mount INP is <code>40 → 40 ms</code> with paired round deltas <code>[0, 0]</code>; mount scripting and total are both -2% versus main. No CollectionRenderer row crosses the significance gates, so the prior confirmed +20% mount-INP regression is gone.

diegohaz:
**Fixed**

- [<code>5c985d157</code>](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1) ([reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633701608)) restores local automatic detection for omitted nested renderers while preserving inheritance of explicit ancestor values, including <code>null</code>.
- [<code>5c985d157</code>](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1) ([reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633704023)) identifies direct elements with realm-neutral <code>isElement</code> before ref-shaped values.
- [<code>5c985d157</code>](https://github.com/ariakit/ariakit/commit/5c985d1571d7acd122166bfa35b878fd1aceb0c1) ([reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633706201)) keeps live target invalidation in layout while moving state/provider reconciliation to the passive effect.

**Verified**

- Focused SelectRenderer browser suite: 21/21 across Chrome, Firefox, and Safari.
- React 19 and React 18 fixture tests: 2/2 each; lint, full typecheck, and app build are clean.
- [Exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29955693675): 17/17 jobs passed with no significant Chrome or Node changes. [Raw CollectionRenderer data](https://github.com/ariakit/ariakit/actions/runs/29955693675/artifacts/8544126615) has mount INP <code>40 → 40 ms</code> and paired deltas <code>[0, 0]</code>.
- Final CI: 16/16 required and 49/49 visible checks passed at <code>5c985d157</code>.

**Visual evidence**

- Exact-head Chromium capture verified both review cases at <code>scrollTop = 2000</code> with item 51 visible; the direct target also had <code>isHTMLElement = true</code>, an own <code>current</code> property, and <code>current = 0</code>.
- Publishing gap: the reviewed local capture is not attached to the PR; the linked cross-browser regressions are the reproducible evidence.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +298,64 @@ function getViewport(scroller: Element) {
...
+        autoResolved.current = false;
+        scrollerRef.current = null;
+      } else if (!autoResolved.current) {
+        scrollerRef.current = getScrollElement(renderer, scrollElement);
+        autoResolved.current = true;
+      }
+    } else {
+      autoResolved.current = false;
+    }
+    const nextScroller = scrollerRef.current;
```
diegohaz, line 413: **[P1] Re-resolve explicit refs after ancestor attachment**

This passive effect reads the snapshot captured by the renderer's layout effect. On the initial commit, the descendant renderer's layout effect runs before React attaches the ancestor host ref, so the documented stable-ref form resolves to `null` here:

```tsx
const scroller = useRef<HTMLDivElement>(null);

<div ref={scroller}>
  <SelectRenderer
    items={items}
    itemSize={40}
    scrollElement={scroller}
  />
</div>
```

The ancestor ref becomes non-null later in the same layout phase, but this effect only rereads `scrollerRef.current`, which is still the stale `null` snapshot. With no subsequent render, the renderer never installs scroll/resize/intersection listeners. A bare `CollectionRenderer` can remain blank; `SelectRenderer` only shows its persistent edge items, and scrolling to the middle cannot recover it.

[Exact-head Chromium reproduction](https://github.com/user-attachments/assets/dd7ef7d6-69a8-4306-885e-961972218224): at `scrollTop = 2000`, item 51 remains unrendered until an unrelated state commit, after which items 50–54 appear immediately. The committed test misses this path because connecting the ref and loading the items are separate commits.

Please re-resolve explicit refs/resolvers in the passive synchronization effect, after ancestor refs and layout effects have attached, while retaining the layout snapshot for the stale-target guard. A regression should mount the canonical ancestor-ref form with populated items and scroll before any other update.

Assessment: **High severity** (estimate, 95% confidence); **P1 priority** (estimate, 99% confidence); **possible breakage** (estimate, 95% confidence); **easy fix** (estimate, 90% confidence).

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -949,14 +1046,29 @@ export interface CollectionRendererOptions<
...
   /**
-   * Whether the items should be rendered when the closest scrollable ancestor
-   * is scrolled.
+   * The element whose viewport determines which items are rendered. By
+   * default, the closest scrolling ancestor is used.
+   *
+   * The element must be a scrolling ancestor in the same document. If a
+   * function is provided, it will be called with the renderer element as an
+   * argument. Explicit values are inherited by nested renderers using the same
+   * store unless they provide their own. When this prop is omitted, each
```
diegohaz: **[P2] State the inherited-value precedence for omitted props**

These two sentences give conflicting guidance for a nested renderer that omits the prop under a same-store ancestor with an explicit value:

```tsx
<CollectionRenderer scrollElement={outerRef}>
  <CollectionRenderer />
</CollectionRenderer>
```

The first sentence says the child inherits `outerRef`; the second says its omission makes it detect its own closest scrolling ancestor. The implementation chooses inheritance. A consumer following the latter sentence for an independently scrolling nested panel can therefore bind virtualization to the outer viewport and see stale or blank inner items.

Please state the precedence explicitly, for example: “If neither this renderer nor a same-store ancestor provides a value, the renderer detects its closest scrolling ancestor.” The same ambiguous sentence is repeated in the changeset and should be updated there too.

- Severity: **Low** (estimate, 95% confidence), because runtime behavior is stable but the public contract describes the nested case ambiguously.
- Priority: **P2** (estimate, 95% confidence), because API documentation shipped by this PR should match its intended precedence.
- Breakage likelihood: **Unlikely** (estimate, 90% confidence), because it requires nested same-store renderers with an explicit ancestor target and an omitted child target.
- Fix effort: **Easy** (estimate, 99% confidence), because this is a wording-only change in two places.

.changeset/200-select-renderer-scroll-element.md
```diff
@@ -0,0 +1,21 @@
...
+const scrollElementRef = useRef<HTMLDivElement>(null);
+
+<div ref={scrollElementRef} className="scroller">
+  <SelectRenderer scrollElement={scrollElementRef} items={items}>
+    {(item) => <SelectItem key={item.id} {...item} />}
+  </SelectRenderer>
+</div>;
+```
+
+Nested renderers using the same store inherit an explicitly provided scroll element unless they provide their own. When the prop is omitted, each renderer detects its closest scrolling ancestor. Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi) for reporting the issue.
```
diegohaz: **[P2] Credit all material issue contributions on the required surfaces**

This final sentence credits only the report and keeps the credit in the behavior paragraph. Issue #3913 also includes a video, a StackBlitz reproduction, and the explicit-scroller prop proposal that shaped this API and its regression coverage. Ariakit's multi-paragraph changeset format requires a standalone final credit paragraph describing those material contributions, and the PR description also requires a matching `## Acknowledgments` section.

For example:

```md
Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi) for reporting the issue, providing the reproduction, and proposing an explicit scroller prop.
```

Please use that as the final standalone changeset paragraph and add the same precise acknowledgment to the PR description.

- Severity: **Low** (verified against issue #3913 and the current PR body), because this affects release metadata and contributor recognition rather than runtime behavior.
- Priority: **P2** (estimate, 95% confidence), because the required acknowledgment surfaces should be complete before merge.
- Breakage likelihood: **Likely** (verified), because the current changeset and PR body deterministically omit these contributions/surfaces if merged as-is.
- Fix effort: **Easy** (estimate, 99% confidence), because this only requires small Markdown edits.

diegohaz (commented):
Comment-only **re-review** of commits `1b12aa49` through `5c985d15`, plus the full impacted renderer behavior and all relevant prior threads. I independently revalidated the replies and current implementation against exact head `5c985d1571d7acd122166bfa35b878fd1aceb0c1`.

## Findings

- [P1: re-resolve explicit refs after ancestor attachment](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633901198). The documented stable ancestor-ref form captures `null` in the descendant layout effect, then the passive effect rereads that stale internal snapshot. No viewport listeners install until another render; the attached exact-head Chromium recording shows scrolling to item 51 remains blank until an unrelated commit.
- [P2: state the inherited-value precedence for omitted props](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633916239). The revised JSDoc and changeset contradict each other for an omitted child beneath an explicit same-store ancestor.
- [P2: credit all material issue contributions on the required surfaces](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633916593). Issue #3913 supplied the report, video/StackBlitz reproduction, and explicit-scroller proposal; the standalone changeset credit and PR `## Acknowledgments` need to reflect that.

## Prior review outcomes

- [Nested automatic detection](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633347050): **already resolved** by using the inherited scroller only for explicit same-store values.
- [Direct element discrimination](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633348139): **already resolved** with realm-neutral `isElement` before the ref-shaped `current` branch.
- [Layout-path state synchronization](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633349001): **already resolved**; final exact-head mount INP is `40 → 40 ms`.
- [Stale null/ref targets](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633350147): **still resolved** by the live target mismatch guard. All four threads are resolved, so I did not duplicate or reopen them.

## Validation

- `pnpm build-app-lite`: passed.
- Focused exact-head browser suite: **21/21** across Chrome, Firefox, and Safari.
- React 19 and React 18 fixture tests: **2/2 each**.
- Exact-head real-Chromium ancestor-ref reproduction: confirmed before any incidental render; the temporary fixture was removed.
- Exact-head perf workflow: **17/17 jobs** passed, with no significant Chrome or Node changes; raw CollectionRenderer mount INP remains `40 → 40 ms`.
- `git diff --check` passes; the detached worktree is clean at the reviewed SHA.

Coverage included the complete five-file incremental diff, the full PR diff and surrounding call sites, React commit/effect ordering, nested context precedence, public option forwarding/types, browser regressions, package exports, changeset/package scope, contributor audit, and the relevant performance benchmark. I did not rerun the entire monorepo suite locally; exact-head CI supplies that broader coverage. The previously mentioned local success capture was not attached to the PR; the new exact-head failure recording is attached to the P1, and success evidence should be refreshed after the fix.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [`224168fc2`](https://github.com/ariakit/ariakit/commit/224168fc2c5205e81490865184c69de9601876a59).

`useScroller` still snapshots explicit refs/resolvers during layout for the stale-target guard, then calls `resolveScroller()` again in the passive synchronization effect, after ancestor refs and layout effects have attached. I added the requested [canonical ancestor-ref regression](https://github.com/ariakit/ariakit/blob/fcd62902cfd971102ae38c3fdee365460a15661f/app/src/sandbox/select-renderer/test-browser.ts#L149-L165): it mounts populated items, performs no intervening update, scrolls to `2000`, and requires item 51 to render.

The regression failed twice before the fix and now passes in the full focused 24/24 Chrome, Firefox, and Safari matrix. [Exact-head recording](https://github.com/user-attachments/assets/423f65ca-0798-4b7b-a659-594a5d5803eb) · [exact-head screenshot](https://github.com/user-attachments/assets/59b0e9ad-4c41-4a4a-9754-ae742d23d46e)

The [final exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29961541768) passed all 17 jobs and reports no significant Chrome or Node changes. The [raw CollectionRenderer artifact](https://github.com/ariakit/ariakit/actions/runs/29961541768/artifacts/8546309317) has no regression candidate: mount and scroll total are both +2%, rerender is -8%, and toggle is -4%. All PR checks are green.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [`fcd62902c`](https://github.com/ariakit/ariakit/commit/fcd62902cfd971102ae38c3fdee365460a15661f).

The JSDoc and changeset now state the precedence directly: if neither the current renderer nor a same-store ancestor provides an explicit `scrollElement` value, the renderer detects its closest scrolling ancestor. This makes it clear that an explicit same-store ancestor value is inherited before automatic detection is considered.

.changeset/200-select-renderer-scroll-element.md
diegohaz: Fixed in [`fcd62902c`](https://github.com/ariakit/ariakit/commit/fcd62902cfd971102ae38c3fdee365460a15661f).

The changeset now ends with a standalone credit paragraph recognizing @ItaiYosephi's report, video, StackBlitz reproduction, and explicit scroll-element proposal. I added the same precise credit under `## Acknowledgments` in the PR description.

diegohaz:
**Fixed**

- [`224168fc2`](https://github.com/ariakit/ariakit/commit/224168fc2c5205e81490865184c69de9601876a59) · [reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634223562): re-resolve explicit refs/resolvers after ancestor attachment while retaining the layout snapshot used by the stale-target guard; added the canonical initial ancestor-ref browser regression and exact-head visual evidence.
- [`fcd62902c`](https://github.com/ariakit/ariakit/commit/fcd62902cfd971102ae38c3fdee365460a15661f) · [reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634224798): clarify that automatic scroll-ancestor detection applies only when neither the renderer nor a same-store ancestor provides an explicit value.
- [`fcd62902c`](https://github.com/ariakit/ariakit/commit/fcd62902cfd971102ae38c3fdee365460a15661f) · [reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634226064): add precise standalone contributor credit to the changeset and the PR's `## Acknowledgments` section.

**Verification**

- The initial-ref regression failed twice before the fix and passes in the full focused 24/24 Chrome, Firefox, and Safari matrix.
- The [final exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29961541768) passed all 17 jobs with no significant Chrome or Node changes and no CollectionRenderer regression candidate.
- All PR checks are green on `fcd62902cfd971102ae38c3fdee365460a15661f`.

.changeset/200-select-renderer-scroll-element.md
```diff
@@ -0,0 +1,23 @@
+---
+"@ariakit/react-components": patch
+---
+
+Explicit scroll elements for collection renderers
+
+Added the `scrollElement` prop to collection renderers, including `CompositeRenderer` and `SelectRenderer`. This prop can be an element, a React ref, or a function that resolves the element whose viewport controls rendering.
```
diegohaz: [P2] Include the `null` behavior in the release contract

The accepted-input sentence reads as exhaustive, but it omits literal `null`; the type and JSDoc define that value as disabling viewport-driven rendering. A consumer scanning the generated changelog for a way to pause viewport work could conclude that only an element, ref, or resolver is supported. This is also detailed prose in a multi-paragraph changeset, so it should use present tense.

```suggestion
Collection renderers, including `CompositeRenderer` and `SelectRenderer`, now accept a `scrollElement` prop. It accepts an element, a React ref, a resolver function, or `null` to disable viewport-driven rendering.
```

**Assessment**

- Severity: Low (estimate, 95% confidence), because runtime behavior is correct but the shipped release contract is incomplete.
- Priority: P2 (estimate, 95% confidence), because this should be corrected before the entry becomes changelog output.
- Breakage likelihood: Unlikely (estimate, 90% confidence), because it only affects consumers relying on the release entry to discover the opt-out.
- Fix effort: Easy (verified by the one-line replacement above).

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -949,14 +1050,30 @@ export interface CollectionRendererOptions<
    */
   items?: Items<T>;
   /**
-   * Whether the items should be rendered when the closest scrollable ancestor
-   * is scrolled.
+   * The element whose viewport determines which items are rendered. By
+   * default, the closest scrolling ancestor is used.
+   *
+   * The element must be a scrolling ancestor in the same document. If a
```
diegohaz, line 1124: [P2] Complete the cycle-3 complexity record

The Cycle 3 PR-body reassessment says `No edge case is intentionally unsupported`, but this public contract deliberately excludes non-ancestor and cross-document targets. The record also does not weigh the implementation and maintenance cost accumulated across the discriminator, layout plus passive resolution, scroller state and live-ref guard, inherited context, tests, and documentation. Listing the accepted forms restates the chosen scope without explaining why that cost is preferable to the narrower ref-only direction after three finding-bearing cycles.

Please update the collapsed reassessment to cover that comparison and the documented exclusions. For example: a ref-only contract would still need the dual timing and stale-target guard; direct and resolver support add only the discriminator; inheritance avoids repeating the target on nested renderers; and non-ancestor or cross-document targets remain intentionally unsupported.

**Assessment**

- Severity: Low (estimate, 95% confidence), because this affects the required maintenance record rather than runtime behavior.
- Priority: P2 (estimate, 95% confidence), because the cycle-3 audit should be complete before another feedback-and-fix iteration.
- Breakage likelihood: Unlikely (estimate, 95% confidence), because impact appears only when future maintainers rely on the reassessment to revisit this design.
- Fix effort: Easy (estimate, 98% confidence), because this is a PR-body-only clarification.

diegohaz (commented):
Re-review of commits `5c985d1571d7acd122166bfa35b878fd1aceb0c1..fcd62902cfd971102ae38c3fdee365460a15661f`, plus the impacted full diff and surrounding public contract.

### New findings

- [P2: include the `null` behavior in the release contract](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634345785)
- [P2: complete the cycle-3 complexity record](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634347983)

No new runtime-correctness, API-forwarding, test-coverage, or performance finding remains. The two-phase implementation is justified: the layout snapshot fences stale inherited work, while the passive re-resolution observes ancestor refs after the full layout phase. A passive-only or ref-only simplification would not preserve both guarantees.

### Prior finding validation

Authoritative GraphQL state reports all seven review threads resolved. I independently revalidated the three findings from the latest marked review:

- [Ancestor ref on the initial commit](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633901198): resolved. Exact-source host-ref and imperative-ref harnesses each observed scrolling on initial mount in normal and Strict Mode, and the committed regression scrolls directly to item 51 without an intervening component update.
- [Inherited-value precedence documentation](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633916239): resolved consistently in JSDoc and the changeset.
- [Contributor credit](https://github.com/ariakit/ariakit/pull/6806#discussion_r3633916593): resolved with precise standalone credit in both release metadata and the PR body.

Ref-only narrowing remains a declined non-goal because it does not remove the dual lifecycle timing or stale-target fence. No follow-up issue is warranted. A possible requirement to link `CompositeRenderer` and `SelectRenderer` in the changeset was invalidated: both deployed reference URLs return HTTP 404, and the local public/docs path does not establish future reference pages, so plain inline names are correct.

### Verification

- `git diff --check 5c985d1571d7acd122166bfa35b878fd1aceb0c1..fcd62902cfd971102ae38c3fdee365460a15661f`
- `pnpm build-app-lite`
- Focused SelectRenderer Playwright matrix on Chrome, Firefox, and Safari: 24/24 passed
- Focused React 19 and React 18 tests: 2/2 passed on each
- Exact-source initial-mount harness: host and imperative refs passed in normal and Strict Mode
- Exact-head CI is green. Performance run `29961541768` passed 17/17 jobs; the CollectionRenderer results contain no significant or candidate regression.
- Rendered PR evidence was opened and checked: the screenshot shows items 51–55, and the 8.44-second recording loads without a media error and shows direct initial-ref scrolling.

Evidence limitations: no raw log from the automated pre-fix red run is attached, although the prior exact-head recording independently demonstrates the failure and the current committed regression covers that path. There is no separate resolver-function browser case; it shares the verified final `resolveScroller()` path, and source review found no distinct defect.

.changeset/200-select-renderer-scroll-element.md
diegohaz: Fixed in [`ea148ce47`](https://github.com/ariakit/ariakit/commit/ea148ce47a830e915735b45c711f071f163cbd03): the multi-paragraph changeset now uses present tense and lists literal `null` alongside elements, refs, and resolver functions, explicitly stating that it disables viewport-driven rendering.

The wording matches the current TypeScript and JSDoc contract. `pnpm lint-fix`, `git diff --check`, and the required exact-snapshot reviewer gate all pass.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in the [updated PR description](https://github.com/ariakit/ariakit/pull/6806#issue-4950694907): the Cycle 3 record now enumerates the implementation and maintenance cost, compares it with the ref-only direction, and records non-ancestor and cross-document targets as intentionally unsupported.

The reassessment keeps the current design because a ref-only contract would still require the dual layout/passive timing and stale-target guard; direct-element and resolver support add only discrimination before the shared resolution path, while same-store inheritance avoids repeating the target through nested renderers. I also verified the expanded text and existing screenshot/video in the rendered PR description.

diegohaz:
**Fixed**

- [`ea148ce47`](https://github.com/ariakit/ariakit/commit/ea148ce47a830e915735b45c711f071f163cbd03) ([reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634875573)): complete the release contract with present-tense wording and explicit `null` behavior.
- [`PR description`](https://github.com/ariakit/ariakit/pull/6806#issue-4950694907) ([reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634876462)): complete the Cycle 3 complexity comparison and record non-ancestor and cross-document targets as intentionally unsupported.

**Verification**

- `pnpm lint-fix` and `git diff --check` pass.
- A fresh exact-snapshot native reviewer found no actionable issue in the changeset edit or surrounding contract.
- The updated reassessment, screenshot, and video render correctly in the PR description, with no composer draft left behind.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +298,68 @@ function getViewport(scroller: Element) {
...
+        scrollerRef.current = null;
+      } else if (!autoResolved.current) {
+        scrollerRef.current = getScrollElement(renderer, scrollElement);
+        autoResolved.current = true;
+      }
+    } else {
+      autoResolved.current = false;
+      // Ancestor refs attach after descendant layout effects, so resolve the
+      // explicit target again once the entire layout phase has completed.
+      scrollerRef.current = resolveScroller();
```
diegohaz: **[P1] Revalidate inherited targets before child passive work**

React flushes passive effects child-first. If a stable resolver supplied by an explicit parent changes from scroller A to B in an ancestor layout effect, the parent's layout snapshot is still A. An omitted nested renderer therefore enters its `processVisibleIndices` effect first with inherited state A and the shared ref still at A; the equality guard passes, and this parent re-read publishes B only afterward.

I reproduced this on exact head `ea148ce47` with two supported same-document scrolling ancestors:

```tsx
<SelectRenderer scrollElement={() => activeScroller} items={groups}>
  {(group) => <SelectRenderer {...group} itemSize={itemSize} />}
</SelectRenderer>
```

Switching from inner A at `scrollTop = 2000` to outer B at `scrollTop = 0` while changing `itemSize` from 40 to 80 caused a mount-only probe to mount Items 25-28 before the B correction mounted Items 1-3. Those transient mounts can run registration, measurement, focus, or consumer lifecycle effects even though the final DOM converges.

Please propagate a live target resolver/current-target check to inherited children, or otherwise prevent child processing until the explicit parent has synchronized, and add a nested A-to-B regression.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because supported updates can run lifecycle effects for the wrong virtual window before converging.
- Priority: P1 (estimate, 85% confidence), because resolver changes and same-store inheritance are contracts added by this PR.
- Breakage likelihood: Rare (estimate, 85% confidence), because it requires nested inheritance, a commit-time target change, and another virtualization input changing together.
- Fix effort: Moderate (estimate, 75% confidence), because descendants need a live-target validity signal or equivalent synchronization plus a focused regression.

Evidence gap: the exact-source React 19 DOM harness proves mount-only effects, but I did not attach a browser recording because the intermediate DOM converges during passive work.

diegohaz (commented):
Re-review of [`fcd62902c`](https://github.com/ariakit/ariakit/commit/fcd62902cfd971102ae38c3fdee365460a15661f)..[`ea148ce47`](https://github.com/ariakit/ariakit/commit/ea148ce47a830e915735b45c711f071f163cbd03), plus the impacted full diff, surrounding renderer lifecycle/context, current PR description, release contract, tests, CI, and performance evidence.

## Outcome

I found [one new P1](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635028432): an inherited nested renderer can process and mount the previous resolver target's virtual window before its explicit parent's child-last passive re-resolution publishes the new target. The final state converges, but consumer mount/layout effects have already run for the wrong items. Current CI is green because it does not exercise a nested inherited A-to-B target transition with a simultaneous layout-input change.

## Prior finding validation

- [Release-contract P2](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634345785): resolved. The changeset now uses present tense and includes literal `null` with its viewport-disable behavior.
- [Cycle-3 reassessment P2](https://github.com/ariakit/ariakit/pull/6806#discussion_r3634347983): resolved. The PR body now records supported scope, accumulated maintenance cost, the ref-only comparison, and intentionally unsupported non-ancestor/cross-document targets.
- Before this review, all 9 authoritative threads were resolved with complete outer and per-thread pagination; no unresolved thread duplicated the new finding.

## Validation

- Confirmed the live head is `ea148ce47a830e915735b45c711f071f163cbd03`, the PR is mergeable, and the worktree is clean.
- The incremental diff changes only the changeset sentence and merged-base `packageManager`; the renderer and focused tests are unchanged from the prior baseline. Incremental and full `git diff --check` pass, and the merge has no conflict-resolution drift.
- Reproduced the new finding with the exact checkout via `node /private/tmp/ariakit-6806-inherited-resolver-repro.mjs`. A mount-only probe recorded stale Items 25-28, followed by corrected Items 2-3; assertions require both the stale mount and final correction.
- `pnpm exec oxfmt --check .changeset/200-select-renderer-scroll-element.md` passed.
- `pnpm test-react app/src/sandbox/select-renderer/test.ts` passed 2/2.
- The exact-head focused Playwright matrix passed 24/24 across Chrome, Firefox, and Safari.
- Exact-head CI is green, including React 19, React 18, builds, browser matrices, docs, and CodeQL.
- Exact-head performance CI passed all 17 expected jobs. The report has no significant/candidate CollectionRenderer regression, and raw browser/Node artifacts are complete and correctly paired.

## Dispositions and evidence gaps

This is finding-bearing implementation cycle 4, so the next three-cycle reassessment checkpoint is not due yet. Narrowing to refs alone would not remove the inherited effect-ordering problem, and passive-only resolution would reintroduce the previously confirmed stale-work failure; the current API direction remains viable if inherited children can validate the live target.

No raw pre-fix Playwright log is attached for the original issue, although its exact revision video/source and the current committed regression establish the behavior change. Existing PR visuals were captured at `fcd62902c`, but renderer/fixture/test source is unchanged through `ea148ce47` and the exact-head three-browser suite covers that interaction. The new transient-mount reproduction has no browser recording; it is verified by mount-only React 19 DOM effects and is not fully represented by the final converged DOM. No follow-up issue is proposed for the current-PR finding.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [`c6677dccf`](https://github.com/ariakit/ariakit/commit/c6677dccf6b0b556f6fce5d320f36cb89be4fbf9).

Explicit scroll-target owners now share a commit-scoped controller. Inherited child layouts invalidate it, and the first child passive consumer re-resolves and passively publishes the live target before virtualization can process the old window. Every inherited child commit consumes the invalidation, while `processVisibleIndices` retains the live-ref mismatch guard. Automatic renderer detection stays on its existing path and does not allocate a controller.

The browser regression memoizes the explicit owner and covers both a class-only A → B change and an overscan-driven B → A change. Both failure paths were proven red twice before their fixes, and the final SelectRenderer matrix passes 27/27 across Chrome, Firefox, and Safari. Build and types pass; current React passes 906/906, and React 18 passes 898 with 8 skipped.

diegohaz:
**Fixed**

- [`c6677dccf`](https://github.com/ariakit/ariakit/commit/c6677dccf6b0b556f6fce5d320f36cb89be4fbf9) revalidates and passively publishes inherited explicit scroll targets before child virtualization, including memoized-owner commits, with a two-phase browser regression. [Review reply](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635308805)

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +299,104 @@ function getViewport(scroller: Element) {
   return scroller;
 }
 
-function useScroller(rendererRef: RefObject<HTMLElement | null> | null) {
+function getScrollElement(
+  renderer: HTMLElement,
+  scrollElement: CollectionRendererOptions["scrollElement"],
+): Element | null {
+  if (scrollElement === undefined) return getScrollingElement(renderer);
+  if (typeof scrollElement === "function") return scrollElement(renderer);
```
diegohaz:
```suggestion
  if (scrollElement === undefined) {
    return getScrollingElement(renderer);
  }
  if (typeof scrollElement === "function") {
    return scrollElement(renderer);
  }
```

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -293,16 +299,104 @@ function getViewport(scroller: Element) {
...
-  }, [rendererRef]);
-  return scroller;
+    if (!renderer) return null;
+    return getScrollElement(renderer, scrollElement);
+  };
+  let controller = controllerRef.current;
+  if (!controller && scrollElement !== undefined) {
+    const nextController: ScrollerController = {
+      revalidated: false,
+      resolve: resolveScroller,
```
diegohaz: **[P2] Release the explicit controller when automatic mode resumes**

Once this branch initializes `controllerRef`, switching `scrollElement` back to `undefined` leaves that controller alive. The automatic-mode layout branch returns before replacing `controller.resolve` or clearing `publishedScrollerRef`, so the last explicit resolver/Element—and anything captured by the resolver—remains strongly referenced until the renderer unmounts. An exact-head React 19 GC probe collected a detached control element, kept the former explicit target alive after the automatic-mode commit plus six intervening renders, and collected it only after unmount.

Please make the controller lifetime follow explicit mode—for example, use one mode-scoped controller object that stores its published scroller—or explicitly clear both retained references on automatic commits. The mode-scoped shape also removes the render-time mutation of a ref shared with abortable work-in-progress renders while preserving one stable controller across explicit resolver changes.

**Assessment:** low severity; P2 priority; possible when a long-lived renderer temporarily uses a custom scroller; easy-to-moderate fix plus lifecycle/performance verification.

diegohaz (commented):
Re-review of [`ea148ce47`](https://github.com/ariakit/ariakit/commit/ea148ce47a830e915735b45c711f071f163cbd03)..[`c6677dccf`](https://github.com/ariakit/ariakit/commit/c6677dccf6b0b556f6fce5d320f36cb89be4fbf9), plus the impacted full diff, controller/context lifecycle, public contract, regression fixture, CI, and performance evidence.

## Outcome

The prior inherited-target P1 is resolved. I found [one new P2](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635526619): after a renderer switches from an explicit `scrollElement` back to automatic mode, the persistent controller continues retaining the previous resolver/Element until unmount.

## Simplification assessment

The controller protocol is still necessary: inherited child layouts must invalidate the explicit owner, the first child passive consumer must re-resolve it before stale virtualization work, and the live-ref/committed-state mismatch must fence processing until context catches up. Removing inheritance, the passive phase, or the controller would either narrow the documented contract, reintroduce an already reproduced race, or repeat owner work for every descendant.

The storage can be simpler and safer. A controller whose lifetime follows explicit mode can own the last published target:

```ts
const explicit = scrollElement !== undefined;
const controller = useMemo(() => {
  if (!explicit) return;
  return createScrollerController(scrollerRef, setScroller);
}, [explicit]);
```

Keeping `published` on that object and assigning the current resolver during layout removes both `controllerRef` and `publishedScrollerRef`, preserves one controller across explicit resolver changes, releases old targets on explicit → automatic commits, and avoids mutating a current ref from an abortable render. The two new fixture transitions should remain: the class-only A → B path proves descendant-only invalidation, while the overscan B → A path proves the stale virtual window never mounts.

## Prior finding validation

- [Inherited target revalidation](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635028432): resolved by `c6677dccf`. The exact-source React 19 harness now reports `staleWindowMounted: false` and `correctedToOuter: true`; the transition mounts only Items 2–3 rather than the prior stale Items 25–28.
- Authoritative thread pagination is complete. Ten prior threads are resolved; the only pre-existing unresolved thread is an unrelated brace-formatting suggestion, so the new finding is not a duplicate.

## Validation

- Confirmed live head `c6677dccf6b0b556f6fce5d320f36cb89be4fbf9`, base `052a3b43270ad5bd50e46c75e5b43ddc08341776`, mergeability, and a clean worktree. Incremental and full `git diff --check` pass.
- `pnpm build-app-lite` passed. After rebuilding the preview, the focused SelectRenderer Playwright matrix passed 27/27 across Chrome, Firefox, and Safari; this confirms the earlier local missing-locator failures came from stale preview assets.
- `pnpm test-react app/src/sandbox/select-renderer/test.ts` passed 2/2.
- The exact-head React 19 GC probe collected a detached control, retained the former explicit target through the automatic-mode commit and six subsequent renders, then collected it after renderer unmount.
- Exact-head Actions are green. The tested synthetic merge has the same tree as the PR head; browser CI passed 322 Chrome, 301 Firefox, and 301 Safari tests plus 16 visual tests per browser. React 19 passed 906/906; React 18 passed 898 with 8 expected skips.
- Performance run `29976379844` passed all 17 jobs with no significant/candidate Chrome or Node regression. CollectionRenderer totals are +2% mount, +1% rerender, +5% toggle, and +4% scroll.

## Dispositions and evidence gaps

This is finding-bearing implementation cycle 5; the next mandatory complexity reassessment is cycle 6. Narrowing to refs only, dropping inheritance, and passive-only resolution remain declined because they do not remove the lifecycle constraints demonstrated by prior failures.

The committed inherited-target fixture uses `overflow: visible` on its inner target, so that phase is synthetic rather than a literal documented scrolling ancestor. Overflow style is not consulted by the controller or stale-target fence, and the exact-source two-target harness plus the exact-head cross-browser matrix independently validate the ordering fix, so I treat this as a non-blocking evidence limitation rather than another finding.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Addressed in [`d62900cb1`](https://github.com/ariakit/ariakit/commit/d62900cb182a363e2d65da14bd638b6afffd0fcd). The automatic and resolver branches now use braced blocks, and the adjacent direct-element branch follows the same non-primitive-return style.

`pnpm lint` passes, and the exact staged snapshot passed the required independent pre-commit review with no findings.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Fixed in [`51c37d8d8`](https://github.com/ariakit/ariakit/commit/51c37d8d82a1232cb7a26b50990a49bc0822dad3).

New controllers start with a shared inert resolver. When automatic mode commits, the layout effect marks the old controller inert, replaces its resolver, clears `publishedScrollerRef`, and clears `controllerRef`. This keeps an aborted explicit render from storing a target-capturing resolver and releases both retained references when a committed renderer returns to automatic mode.

I added a Chrome-only GC regression that keeps the target reachable only through an explicit resolver, switches to `undefined`, performs six committed renderer rerenders, requests `HeapProfiler.collectGarbage`, and checks a `WeakRef`. The controlled old-controller variant failed twice with `Released: no`; the final fix passes with `Released: yes`. The existing SelectRenderer suite also passes 27/27 across Chrome, Firefox, and Safari; `pnpm lint` and `pnpm tsc` pass. The exact-head performance workflow is being rechecked separately.

diegohaz:
Addressed the two new review threads at exact head [`d62900cb1`](https://github.com/ariakit/ariakit/commit/d62900cb182a363e2d65da14bd638b6afffd0fcd).

- [`51c37d8d8`](https://github.com/ariakit/ariakit/commit/51c37d8d82a1232cb7a26b50990a49bc0822dad3) scopes the shared controller to explicit mode. New controllers start with an inert resolver; an automatic-mode commit makes the old controller inert and clears the published element and controller ref.
- The same commit adds a Chrome-only `WeakRef`/`HeapProfiler.collectGarbage` regression. The controlled old-controller variant failed twice with `Released: no`; the final implementation passes with `Released: yes` after six committed renderer rerenders.
- [`d62900cb1`](https://github.com/ariakit/ariakit/commit/d62900cb182a363e2d65da14bd638b6afffd0fcd) applies the requested braced conditional style.
- The PR description now records the Cycle 6 complexity decision and the deterministic GC coverage.

Verification:

- SelectRenderer browser suite: 27/27 across Chrome, Firefox, and Safari
- Focused Chrome GC regression: 1/1
- `pnpm lint`
- `pnpm tsc`
- All 49 exact-head PR checks pass
- [Exact-head performance run](https://github.com/ariakit/ariakit/actions/runs/29980665189): all jobs pass; the [authoritative report](https://github.com/ariakit/ariakit/pull/6806#issuecomment-5048387300) finds no significant Chrome or Node changes
- [Raw CollectionRenderer artifact](https://github.com/ariakit/ariakit/actions/runs/29980665189/artifacts/8553149303): balanced `B1,C1,C2,B2` paired rounds, five samples per side per round, matching script profiles, and no candidate/significant row. Total changes are mount `+1%`, rerender `+5%`, toggle `-1%`, and scroll `-6%`; mount INP is `+4ms`, below the absolute threshold.

Both current threads have one exact-state reply and remain unresolved/outdated for reviewer-owned resolution. The final audit found no new comments and no pending review.

diegohaz (commented):
Re-review of [`c6677dccf`](https://github.com/ariakit/ariakit/commit/c6677dccf6b0b556f6fce5d320f36cb89be4fbf9)..[`d62900cb1`](https://github.com/ariakit/ariakit/commit/d62900cb182a363e2d65da14bd638b6afffd0fcd), plus the impacted full diff, controller lifetime, regression coverage, CI, and performance evidence.

## Outcome

No new findings. The previous explicit → automatic retention P2 is resolved: newly rendered controllers start with an inert resolver, and the automatic-mode layout clears the old controller resolver, published target, and controller ref. This prevents aborted explicit renders from capturing the render-time resolver and releases the committed explicit resolver when automatic mode takes over.

All 12 review threads are now resolved, with complete outer and nested pagination.

## Simplification reassessment

I reconsidered the earlier mode-scoped `useMemo` controller suggestion. I no longer think it would be a meaningful simplification. Inherited child layouts and passive effects need a shared mutable coordination object, and passive closures from the previous commit can still hold the old controller. A mode-scoped object would therefore still require the same explicit invalidation when the owner returns to automatic mode.

The current lifecycle operations are small and serve distinct timing guarantees:

```ts
const nextController = {
  revalidated: false,
  resolve: resolveNullScroller,
  // ...
};

// Automatic-mode commit:
controller.revalidated = true;
controller.resolve = resolveNullScroller;
controllerRef.current = null;
```

Starting inert protects abortable work; setting `revalidated` makes stale descendant callbacks no-op; replacing `resolve` releases captured resolver state; and clearing the ref publishes automatic ownership. Collapsing these into memo identity would hide rather than remove those responsibilities. I do not see a further behavior-preserving reduction worth requesting.

## Prior finding validation

- [Explicit target retention](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635526619): resolved by `51c37d8d8`; the new Chrome WeakRef/CDP regression passes on the exact head.
- [Brace formatting](https://github.com/ariakit/ariakit/pull/6806#discussion_r3635459296): resolved by `d62900cb1`.
- The Cycle 6 reassessment in the PR description accurately records the contract, lifecycle constraints, rejected alternatives, regression requirement, and performance check.

## Validation

- `pnpm build-app-lite` passed.
- `pnpm test-react app/src/sandbox/select-renderer/test.ts` passed 2/2.
- The focused SelectRenderer browser matrix passed 28/28: 10 Chrome tests, including the new GC case, plus 9 Firefox and 9 Safari tests.
- Incremental and full `git diff --check` pass; the exact-head worktree is clean.
- Exact-head Actions and all required checks are green. Performance run `29980665189` passed all 17 jobs with no significant Chrome or Node regression; the CollectionRenderer totals are +1% mount, +5% rerender, -1% toggle, and -6% scroll.

## Dispositions and evidence

I also stress-tested a detached direct `Element` target. It remains reachable until unmount, but that setup violates the documented same-document scrolling-ancestor constraint, so I treat it as an out-of-contract non-goal rather than evidence that the supported controller path still leaks. The supported resolver-to-automatic transition is covered by the new exact-head Chrome GC test.

No additional screenshot or recording is warranted for the new regression: heap reachability is not visually communicable, while the user-visible scrolling behavior remains covered by the existing preview and browser matrix.

PR #6810, opened 2026-07-22
Fix initially open sibling dialogs

### LINKED ISSUE #5238: Dialogs are freezing

diegohaz:
## Motivation

When two sibling modal `Dialog` components mount open in the same render, each dialog currently treats the other as outside its modal boundary. Both portal trees become inert, so even the visually frontmost dialog cannot be used.

```tsx
<Ariakit.Dialog open>Oranges</Ariakit.Dialog>
<Ariakit.Dialog open>Apples</Ariakit.Dialog>
```

Fixes #5238.

## Solution

Default modal dialogs now identify their exclusive portal root. While dialogs become open in the same layout pass, an earlier dialog treats later pending default portals as part of its modal boundary. The later dialog can therefore remain interactive and inert the earlier one, while dialogs from an already established stack keep their existing relationship.

The opening cohort is stable for that open cycle, including tree snapshot refreshes, and preserves the whole portal root so custom backdrops and descendant portals are covered. Peer lookup stays within the dialogs' shared document or shadow root. Browser and happy-dom regressions verify the initial stack, snapshot refresh, foreground backdrop interaction, shadow-root portal contexts, and restoration of the background dialog.

## Preview

[Sibling dialog interaction recording](https://github.com/user-attachments/assets/015eb503-6f76-45f2-83de-36341a4a44a4)

The recording shows the front Apples dialog accepting input while Oranges remains in the background, followed by Oranges becoming interactive after Apples closes.

## Workaround

When the foreground dialog is known, keep a ref to its exclusive portal root and return that root from the background dialog's `getPersistentElements` callback:

```tsx
const applesPortalRef = useRef<HTMLElement | null>(null);

<Ariakit.Dialog
  open
  // TODO: Remove after https://github.com/ariakit/ariakit/issues/5238 is fixed.
  getPersistentElements={() =>
    applesPortalRef.current ? [applesPortalRef.current] : []
  }
>
  Oranges
</Ariakit.Dialog>

<Ariakit.Dialog open portalRef={applesPortalRef}>
  Apples
</Ariakit.Dialog>
```

This works in document and shadow-root contexts and preserves the whole portal boundary, including backdrops and descendant portals. It is intentionally scoped to applications that know the foreground dialog. For a dynamic stack, the callback must return every later dialog portal that should remain interactive.

## Acknowledgments

Thanks to [@yishayhaz](https://github.com/yishayhaz) for reporting and reproducing the issue and [@gonzoblasco](https://github.com/gonzoblasco) for investigating the sibling-dialog coordination gap.

<details>
<summary>2026-07-22, cycle 3: Narrow to one opening cycle</summary>

**Assessment**

#5238 completely freezes an initially open sibling-dialog composition, but that composition is uncommon and the report is specifically about siblings entering the open state in one render. Successive review cycles found that preserving the relationship across sibling close/reopen, portal remount, `modal`, `store`, and `unmountOnHide` changes required stable cross-dialog identity plus eligibility notification. That machinery grew beyond the issue's scope and would add a new coordination subsystem to a sensitive modal path.

**Decision**

Use a narrow opening-cycle design: default modal siblings that are still pending in the same layout pass are ordered by their physical portal order, while already established dialogs are excluded. Retain their concrete exclusive portal roots only for the current open cycle so snapshot refreshes and whole-portal backdrops remain correct. Do not add cross-dialog identity or lifecycle notifications in this patch. Dynamic peer lifecycle and eligibility changes are intentionally outside this fix; applications that coordinate a broader stack can continue to use `getPersistentElements` explicitly.

</details>

<details>
<summary>2026-07-23, cycle 3: Keep composed paths and key inside membership by dialog element</summary>

**Assessment**

The original #5238 failure completely freezes the affected sibling-modal composition, and open ShadowRoot support is part of the public portal behavior even though repeated dialog IDs across independent web components are less common. The composed-path follow-up is necessary to recognize persistent elements through nested ShadowRoots, but its ID-derived positive marks allow unrelated dialogs with the same standards-valid ID to share ownership. Continuing with string-derived DOM properties would require additional root and instance encoding in a sensitive outside-interaction path.

**Decision**

Keep the full composed path for positive membership and the dialog-root projection for outside marking. Replace ID-derived inside properties with membership keyed by the concrete dialog element, which removes cross-root ID collisions and DOM property mutation without adding a global dialog registry or public contract. Add branch-reaching mousedown coverage and narrow the release note to the newly shipped ShadowRoot behavior. Closed ShadowRoot internals and broader dynamic dialog-stack coordination remain outside this PR.

</details>

## Review gate reassessment

<details>
<summary>2026-07-22, cycle 3: Continue the narrow root-local dialog coordination design</summary>

**Assessment**

#5238 completely freezes the affected sibling-modal composition, so its impact is severe when encountered even though initially open sibling default portals are uncommon. Open shadow-root portal contexts are a less frequent subset, but they are part of the supported public `Portal` composition. The three gate findings concerned test ownership and release-note attribution and scope; they did not expose additional runtime state, branches, or coordination requirements. The production follow-up only changes peer discovery from the owning document to the connected dialog's document or shadow root, while the existing opening-cycle cohort remains unchanged.

**Decision**

Continue the narrow root-local design and complete the app-owned coverage and release-note corrections. Do not add a global registry or broaden coordination to custom portal elements, inaccessible closed shadow roots, or dynamic peer lifecycle changes. Those cases remain intentionally outside this fix, and no follow-up issue is needed for this review cycle.

</details>

<details>
<summary>2026-07-23, cycle 3: Keep the runtime fix and document its consumer impact</summary>

**Assessment**

The third staged review confirmed the previous ShadowRoot runtime findings are resolved and found only that the existing changeset did not mention the consumer-observable `getPersistentElements` correction introduced while responding to review. This is a release-documentation omission, not evidence of additional runtime complexity. The implementation remains confined to event target classification in the existing outside-interaction path, with focused regression coverage.

**Decision**

Keep the root-aware event-path design. Extend the existing patch changeset to describe persistent-element interactions across open shadow roots before focus, preserving the single continuous entry and existing contributor credit. Do not split the runtime implementation or broaden the PR's dialog coordination scope.

</details>


packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -98,6 +99,31 @@ function getElementFromProp(
   return element;
 }
 
+function getLaterOpenModalPortals(dialog: HTMLElement) {
+  const doc = getDocument(dialog);
+  const dialogs = doc.querySelectorAll<HTMLElement>(
+    "[data-dialog][data-dialog-portal][data-open]",
+  );
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`2b7bf8108`](https://github.com/ariakit/ariakit/commit/2b7bf8108c728b9b569e48073181b3a9b3a614e8): peer discovery now scopes dialog enumeration and portal ID lookup to the connected dialog root (`Document` or `ShadowRoot`). The existing #5238 app fixture now includes targeted happy-dom and Chrome/Firefox/Safari coverage for default dialog portals hosted inside an open shadow root.

diegohaz:
**Fixed**

- [`2b7bf8108`](https://github.com/ariakit/ariakit/commit/2b7bf8108c728b9b569e48073181b3a9b3a614e8) ([review thread](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633128498)): Scoped sibling-dialog discovery to the owning document or shadow root and added app-owned happy-dom plus Chrome/Firefox/Safari regression coverage.

app/src/sandbox/dialog-5238/index.react.tsx
```diff
@@ -0,0 +1,124 @@
...
+    portalElement.dataset.dialogShadowPortal = "";
+    shadowRoot.append(portalElement);
+    setPortalElement(portalElement);
+  }, []);
+
+  return (
+    <>
+      <div ref={setHost} data-testid="dialog-shadow-host" />
+      {portalElement && (
+        <Ariakit.Portal portalElement={portalElement}>
```
diegohaz, line 104: The PR description's Workaround section uses `document.getElementById(applesDialogId)` and returns only the dialog element, but this line adds the open-ShadowRoot composition the PR now supports. A current-release Web Component consumer who copies that workaround gets `null` because document lookup cannot cross the shadow boundary, so the callback returns `[]` and both initially open dialogs remain inert. In the document case, preserving only the dialog also leaves a sibling backdrop or descendant portal outside the persistent boundary.

Please make the example root-neutral and preserve the front dialog's exclusive portal root:

```tsx
const applesPortalRef = useRef<HTMLElement>(null);

<Ariakit.Dialog
  open
  getPersistentElements={() =>
    applesPortalRef.current ? [applesPortalRef.current] : []
  }
>
  Oranges
</Ariakit.Dialog>

<Ariakit.Dialog open portalRef={applesPortalRef}>
  Apples
</Ariakit.Dialog>
```

I verified both variants in a temporary cohort-disabled fixture using this shadow-root topology. The current ID example left the front portal inert, while the `portalRef` version kept the front dialog and backdrop interactive. The replacement also passed the repository type check, and I restored all experiment changes.

**Assessment**

- Severity: High (verified), because the affected UI remains completely noninteractive.
- Priority: P1 (estimate, 95% confidence), because the workaround is introduced by this PR and conflicts with its stated shadow-root and whole-portal support.
- Breakage likelihood: Rare (estimate, 90% confidence), because complete failure requires initially open sibling modal dialogs inside an open shadow root on a version that still needs the workaround.
- Fix effort: Easy (verified), because this is a PR-description-only replacement and the `portalRef` version passed the focused runtime and type checks.

app/src/sandbox/dialog-5238/test-browser.ts
```diff
@@ -0,0 +1,109 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // Reproduces https://github.com/ariakit/ariakit/issues/5238
```
diegohaz, line 4: The regression commit [`39df9a4d`](https://github.com/ariakit/ariakit/commit/39df9a4dbe3736e83a213283a84bce37f4fc98d6) records `Co-authored-by: yishayhaz <50710472+yishayhaz@users.noreply.github.com>`, but that GitHub account's current profile name is `Yishay Hazan`. The contributor-credit policy requires the profile name when it is available. GitHub still resolves the account through the noreply address, so the contributor is not losing attribution, but the raw commit metadata does not follow the repository policy.

Please use the corrected trailer in the next relevant follow-up commit, or amend the regression commit if no such commit will be added:

```text
Co-authored-by: Yishay Hazan <50710472+yishayhaz@users.noreply.github.com>
```

**Assessment**

- Severity: Low (verified), because attribution resolves correctly and only the displayed name and policy compliance are affected.
- Priority: P2 (estimate, 95% confidence), because this should be corrected before merge but has no runtime or release behavior impact.
- Breakage likelihood: Likely (verified), because the incorrect display name is present in the exact current commit metadata.
- Fix effort: Moderate (estimate, 90% confidence), because correcting the first of four published commits may require rebasing its descendants and force-pushing the PR branch.

diegohaz (commented):
## Review summary

This was an initial full review of [`ab0f70f2...2b7bf810`](https://github.com/ariakit/ariakit/compare/ab0f70f2d566d9f872b152f944230d263d50a852...2b7bf8108c728b9b569e48073181b3a9b3a614e8). I reviewed the dialog coordination change and its adjacent portal, nested-dialog, inert-tree, outside-interaction, and Escape paths; Dialog-derived consumers; the app fixture and both test suites; release metadata; [#5238](https://github.com/ariakit/ariakit/issues/5238); and the broader registry alternative from [#6735](https://github.com/ariakit/ariakit/pull/6735).

I found two current-PR issues:

- [The documented workaround fails in the supported shadow-root composition and preserves the dialog instead of its whole portal boundary.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633873726)
- [The reporter's co-author trailer uses the login instead of the available GitHub profile name required by the contributor-credit policy.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633874767)

The runtime implementation itself looks sound. I found no additional correctness, compatibility, security, maintainability, or performance issue in the opening-cohort design.

[pr-6810-dialog-interaction.webm](https://github.com/user-attachments/assets/015eb503-6f76-45f2-83de-36341a4a44a4)

The exact-head recording shows the front Apples dialog accepting input while the Oranges dialog remains backgrounded, then the Oranges dialog becoming interactive after the front dialog closes.

### Verification

- `pnpm test-react app/src/sandbox/dialog-5238/test.ts`: 2/2 passed.
- `pnpm test-react18 app/src/sandbox/dialog-5238/test.ts`: 2/2 passed.
- `APP_PORT=49231 NEXTJS_PORT=49300 APP_INSPECTOR_PORT=49239 NEXTJS_INSPECTOR_PORT=49309 pnpm -F app run test --project=chrome --project=firefox --project=safari src/sandbox/dialog-5238/test-browser.ts`: 6/6 passed.
- `pnpm build-app-lite`, `pnpm tsc`, and `git diff --check`: passed.
- Red/green proof: the head tests failed 2/2 with the base dialog implementation, the shadow test failed against the immediate pre-shadow implementation, and both passed at the reviewed head for the expected inertness assertions.
- Workaround experiment: with opening cohorts temporarily disabled, the PR description's ID lookup left the foreground shadow-root portal inert; the proposed `portalRef` replacement kept the dialog and backdrop interactive and passed the type check. All temporary edits were restored.
- Current GitHub status: 49/49 checks passed. The published performance report found no confirmed Chrome or Node regression.

### Review-thread and scope outcomes

- There are no relevant unresolved review threads. The only historical thread, [`PRRT_kwDOBjT0Fc6TCcUy`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633128498), is resolved; I independently validated its shadow-root concern against the current head, so no reply was needed.
- Declined non-goal: dynamic sibling reopen, portal remount, `modal` or store changes, custom portal eligibility, and inaccessible closed shadow roots. They are outside the stated opening-cycle contract, and the global registry alternative would add cross-dialog identity and lifecycle notification machinery without being necessary for [#5238](https://github.com/ariakit/ariakit/issues/5238).
- No registered follow-up is needed. Whole portal roots are necessary for backdrops and descendant portals, and the root-local `WeakSet` approach is the best trade-off for the scoped behavior.
- Validation gap: I did not independently build historical revisions to run `test-browser.ts` red in each browser. The equivalent happy-dom regression was exercised red, and the exact-head browser test passed in Chrome, Firefox, and Safari locally and in CI.

app/src/sandbox/dialog-5238/index.react.tsx
diegohaz: Fixed in the [updated PR description](https://github.com/ariakit/ariakit/pull/6810#issue-4951694152): the workaround now stores the foreground dialog portal root in `portalRef` and returns that root from `getPersistentElements`, so it works in document and open ShadowRoot contexts while preserving backdrops and descendant portals. The [current-behavior recording](https://github.com/user-attachments/assets/015eb503-6f76-45f2-83de-36341a4a44a4) is published under `## Preview` and renders correctly.

app/src/sandbox/dialog-5238/test-browser.ts
diegohaz: Fixed in [`5b3f2939`](https://github.com/ariakit/ariakit/commit/5b3f293993cfd332b87891e7b83c088c69e09c73): the regression commit now records `Co-authored-by: Yishay Hazan <50710472+yishayhaz@users.noreply.github.com>`. The rewritten chain preserves all four prior tree IDs, commit titles, commit order, and the existing Gonzalo Blasco trailer; the verified PR head is [`88f6e30b`](https://github.com/ariakit/ariakit/commit/88f6e30bb99109f0972a8a7828c1d4bfca80a26d).

diegohaz:
**Fixed**

- [`PR description`](https://github.com/ariakit/ariakit/pull/6810#issue-4951694152) ([review thread](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634063668)): Replaced the document-only ID lookup with a root-neutral `portalRef` workaround that preserves the whole foreground portal, and published the verified interaction recording under `## Preview`.
- [`5b3f2939`](https://github.com/ariakit/ariakit/commit/5b3f293993cfd332b87891e7b83c088c69e09c73) ([review thread](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634065896)): Corrected the reporter trailer to `Yishay Hazan` while preserving all four reviewed tree snapshots, commit titles, commit order, and the existing Gonzalo Blasco trailer.

app/src/sandbox/dialog-5238/index.react.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <Dialogs hideOnInteractOutside={false} />
```
diegohaz: **[P1] Keep the ShadowRoot regression on the public default**

`hideOnInteractOutside` defaults to `true`, so this opt-out hides a failure in the ShadowRoot support added here. I removed only this prop, rebuilt the app, and reran the focused Chromium test. The first `toHaveCount(2)` received `0`: the foreground autofocus emits a composed `focusin`, but the document listener sees `event.target` as the shadow host. That host is neither the dialog nor the peer portal marked as inside, so both dialogs hide.

Assessment:

- Severity: Medium (verified). The entire initially open dialog stack disappears before it can be used.
- Priority: P1 (estimated, 95%). The PR explicitly supports and tests an open ShadowRoot, but the test currently disables a core `Dialog` default to make that path work.
- Breakage likelihood: Possible (verified). It is deterministic for this fixture's two initially open sibling modals, background `autoFocusOnShow={false}`, and the default foreground autofocus/outside-interaction behavior.
- Fix effort: Moderate (estimated, 85%). Outside-event targeting and the previous-mousedown tracker should use the original composed-path element consistently, followed by browser coverage with this override removed.

For example, the outside-event path can start from the original element:

```ts
const target = event.composedPath().find(isElement) || event.target;
```

With that temporary change and this opt-out removed, the same rebuilt Chromium test passed. Please make the composed-path handling consistent for `focusin`, click/context-menu classification, and `usePreviousMouseDownRef`, then exercise the ShadowRoot fixture with the default `hideOnInteractOutside` value.

diegohaz (commented):
## Re-review summary

This re-review used the latest prior marked review head, [`2b7bf810`](https://github.com/ariakit/ariakit/commit/2b7bf8108c728b9b569e48073181b3a9b3a614e8), as the incremental baseline and reviewed [`2b7bf810...88f6e30b`](https://github.com/ariakit/ariakit/compare/2b7bf8108c728b9b569e48073181b3a9b3a614e8...88f6e30bb99109f0972a8a7828c1d4bfca80a26d), plus the ShadowRoot behavior newly exposed by the added regression fixture. The force-push is source-neutral: `git range-diff` shows only the reporter trailer display name changing to `Yishay Hazan`, `git diff 2b7bf810..88f6e30b` is empty, and both heads have tree `97f0fd394677fb91e5e36baef3f62063aee51ead`.

I found one Address now issue:

- [P1: the ShadowRoot regression disables the default `hideOnInteractOutside` behavior, masking a composed-event targeting failure that closes both dialogs during foreground autofocus.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634206402)

The opening-cohort implementation otherwise remains sound. I found no additional correctness, compatibility, security, maintainability, simplification, or performance issue in the unchanged runtime design. The root-local portal cohort is still preferable to the broader global-registry alternative for the stated opening-cycle scope.

### Prior finding outcomes

- [`PRRT_kwDOBjT0Fc6TEcyN`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633873726) is resolved. The PR workaround now uses a root-neutral `portalRef` and preserves the whole foreground portal.
- [`PRRT_kwDOBjT0Fc6TEc9x`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633874767) is resolved. [`5b3f2939`](https://github.com/ariakit/ariakit/commit/5b3f293993cfd332b87891e7b83c088c69e09c73) now records `Co-authored-by: Yishay Hazan <50710472+yishayhaz@users.noreply.github.com>` while preserving every reviewed tree snapshot.
- The older root-local lookup thread, [`PRRT_kwDOBjT0Fc6TCcUy`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3633128498), remains resolved. The new Address now thread is [`PRRT_kwDOBjT0Fc6TFWD3`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634206402).

### Verification

- `pnpm test-react app/src/sandbox/dialog-5238/test.ts`: 2/2 passed.
- `CI=true pnpm test-react18 app/src/sandbox/dialog-5238/test.ts`: 2/2 passed.
- `APP_PORT=49731 NEXTJS_PORT=49800 APP_INSPECTOR_PORT=49739 NEXTJS_INSPECTOR_PORT=49809 pnpm -F app run test --project=chrome --project=firefox --project=safari src/sandbox/dialog-5238/test-browser.ts`: 6/6 exact-head tests passed.
- Focused red/green proof in Chromium after a fresh rebuild: removing only `hideOnInteractOutside={false}` made the first ShadowRoot assertion receive zero dialogs instead of two; temporarily deriving the outside-event target from `event.composedPath()` made the same test pass. All experiment edits were restored.
- `git diff --check` passed; the worktree is clean at [`88f6e30b`](https://github.com/ariakit/ariakit/commit/88f6e30bb99109f0972a8a7828c1d4bfca80a26d).
- GitHub reports 49/49 checks passing, including React 18, Chrome, Firefox, Safari, Android, macOS, type checking, and dialog performance. The performance report found no significant Chrome or Node change.
- The existing [interaction recording](https://github.com/user-attachments/assets/015eb503-6f76-45f2-83de-36341a4a44a4) still renders on the current PR and accurately represents the unchanged document-root behavior. Its source tree is identical across the rewrite.

### Scope notes

- Validation gap: the new red/green experiment was run in Chromium only. The current opt-out means the exact-head Firefox and Safari runs do not exercise the failing default ShadowRoot path yet; the fix should add that coverage to the existing three-browser test.
- External, non-PR note: [#5238](https://github.com/ariakit/ariakit/issues/5238) still links the older document-ID workaround comment. I did not classify that as current-PR work because the PR description is corrected and this change removes the need for the workaround; the issue guidance can be refreshed separately if it remains relevant before release.

app/src/sandbox/dialog-5238/index.react.tsx
diegohaz: Fixed in [`eeb8ab34d`](https://github.com/ariakit/ariakit/commit/eeb8ab34da48145e79c7d78d3f89f46d530ab327) and hardened in [`7c783114d`](https://github.com/ariakit/ariakit/commit/7c783114d805f128397b848d8ab1baf3cfbdf51a). The ShadowRoot fixture now exercises the default `hideOnInteractOutside` behavior. Outside-interaction classification retains the composed path for dialog, persistent-element, and nested-dialog membership, projects one target into the dialog's root for outside-tree marking, and stores the same path/root views for the preceding mousedown. The follow-up also preserves the disconnected-origin guard and covers the foreground backdrop, nested persistent ShadowRoots, and unrelated ShadowRoot interactions.

Verification is green for the focused React 19 and React 18 tests, plus all 27 Chromium, Firefox, and WebKit cases across the #5238 and outside-interaction regressions.

[ShadowRoot dialog interaction recording](https://github.com/user-attachments/assets/5186a2d8-bd8f-41d4-a45c-d1deec0c0e0a)

packages/ariakit-react-components/src/dialog/utils/use-previous-mouse-down-ref.ts
```diff
@@ -1,6 +1,10 @@
-import { addGlobalEventListener } from "@ariakit/utils";
+import { addGlobalEventListener, isElement } from "@ariakit/utils";
 import { useEffect, useRef } from "react";
 
+export function getEventTarget(event: Event) {
+  return event.composedPath().find(isElement) || event.target;
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`7c783114d`](https://github.com/ariakit/ariakit/commit/7c783114d805f128397b848d8ab1baf3cfbdf51a). Event classification now retains every `Element` in `composedPath()` for positive dialog, disclosure, focus-trap, persistent-element, and nested-dialog membership, while using the first element in the dialog's root only for outside-tree marking. The preceding mousedown stores the same full path and root-local projection, and a disconnected composed-path origin is ignored before classification.

The regression fixture covers a persistent marker between nested open ShadowRoots, a mousedown-prevented persistent button, an unrelated ShadowRoot control that must dismiss, and a ShadowRoot focus origin removed during dispatch. The previous endpoint-only implementation failed the nested-persistent and disconnected-origin Chromium cases; the final implementation passes the focused React 19 and React 18 tests and all 27 Chromium, Firefox, and WebKit cases.

diegohaz:
## Review follow-up

**Fixed**

- [Default ShadowRoot outside-interaction behavior](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634729834) in [`eeb8ab34d`](https://github.com/ariakit/ariakit/commit/eeb8ab34da48145e79c7d78d3f89f46d530ab327), hardened by [`7c783114d`](https://github.com/ariakit/ariakit/commit/7c783114d805f128397b848d8ab1baf3cfbdf51a).
- [Outside-click dismissal through unrelated ShadowRoots](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634730373) in [`7c783114d`](https://github.com/ariakit/ariakit/commit/7c783114d805f128397b848d8ab1baf3cfbdf51a), while preserving nested persistent-element and disconnected-origin behavior.
- The patch changeset now documents the additional consumer-visible persistent-element correction and keeps the existing contributor acknowledgment.

**Verification**

- React 19 focused regression tests: 7 passed.
- React 18 focused regression tests: 7 passed.
- Chromium, Firefox, and WebKit dialog matrix: 27 passed.
- `pnpm lint`, `pnpm tsc`, and `pnpm build` passed.
- The final six-file staged diff passed an independent pre-commit review with no actionable findings.

[ShadowRoot dialog interaction recording](https://github.com/user-attachments/assets/5186a2d8-bd8f-41d4-a45c-d1deec0c0e0a)

Both review threads remain unresolved for their reviewers to verify.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -58,6 +61,27 @@ function isMouseEventOnDialog(event: Event | MouseEvent, dialog: Element) {
...
+  if (target.hasAttribute("data-focus-trap")) return true;
+  return isElementInside(target, contentElement.id);
+}
+
+function isEventInsideDialog(
+  targets: EventTargets,
+  contentElement: Element,
+  disclosureElement: Element | null,
+) {
+  return targets.elements.some((target) =>
```
diegohaz, line 80: **[P1] Keep cross-root inside marks instance-specific**

The new composed-path scan can see elements from every open ShadowRoot, but `isElementInside` identifies ownership through a property derived only from `contentElement.id`. IDs only need to be unique within their own document tree, so two independent web components may validly render `<Dialog id="settings">`. Dialog B's content then carries the same `__ariakit-dialog-inside-settings` mark that Dialog A looks for, and this `.some(...)` classifies an interaction inside B as inside A. A therefore stays open instead of handling the interaction as outside.

I verified this at the exact head with a temporary light-DOM dialog and an unrelated open-ShadowRoot dialog. With the same ID, the first dialog remained visible in happy-dom and Chromium; changing only the second ID made both tests pass.

Same-ID dialogs in separate roots leave the outside dialog open

The focused Chromium capture shows the unrelated ShadowRoot field focused while the light-DOM dialog incorrectly remains open. Please key inside/outside marks by an opaque dialog-instance identity, or otherwise include owning-root/instance identity, while retaining the full composed path for legitimate nested persistent elements. Conceptually:

```ts
const marker = Symbol();
markTreeInside(marker, allElements);
isElementInside(target, marker);
```

**Assessment**

- Severity: Medium (verified), because an unrelated popup can remain open and keep stale overlay/focus behavior.
- Priority: P1 (estimated, 95%), because the regression is introduced by the new cross-root classification.
- Breakage likelihood: Possible (estimated, 85%), because repeated IDs are standards-valid and natural across reusable Web Components.
- Fix effort: Moderate (estimated, 80%), because marker identity and its callers/tests need to change together.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -163,7 +187,19 @@ export function useHideOnInteractOutside(
       // outside of it). See:
       // - https://github.com/ariakit/ariakit/issues/1336
       // - https://github.com/ariakit/ariakit/issues/2330
-      if (!isElementMarked(previousMouseDown, contentElement?.id)) return;
+      if (!contentElement) return;
+      if (
+        isEventInsideDialog(
```
diegohaz, line 194: **[P1] Add a regression that reaches the stored mousedown path**

The new tests do not exercise this changed `previousMouseDown.elements/rootTarget` branch. In the persistent-ShadowRoot test, the click itself is inside the persistent composed path, so the outer event classifier returns before this listener reads `previousMouseDownRef`. In the unrelated-ShadowRoot test, focusing the input hides the dialog on `focusin` before the click result depends on these lines.

I verified the gap by temporarily returning immediately after the `!previousMouseDown` guard; all 9 focused React tests still passed. Please add a no-focus outside click and a split drag-protection case, for example:

```ts
outsideShadowButton.addEventListener("mousedown", (event) => {
  event.preventDefault();
});
await click(outsideShadowButton);
expect(dialog).not.toBeVisible();
```

Then dispatch `mousedown` from the nested persistent element and the resulting `click` outside, asserting that the dialog stays open. Together those cases protect both `rootTarget` and the full `elements` path.

**Assessment**

- Severity: Low (verified), because the current implementation appears correct but its newly changed dismissal/drag branch is unprotected.
- Priority: P1 (estimated, 85%), because the follow-up explicitly changes and claims coverage for preceding-mousedown behavior.
- Breakage likelihood: Possible (estimated, 80%), because focus-preventing Web Component controls and drag starts from persistent content are supported interactions.
- Fix effort: Easy (verified), because the existing fixture only needs focused event sequences.

.changeset/300-dialog-sibling-modal.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed sibling modal [`Dialog`](https://ariakit.com/reference/dialog) components and modal components built on them, such as [`Popover`](https://ariakit.com/reference/popover) and [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), rendered in their default portals so opening them in the same render no longer made each other inert. Fixed [`Dialog`](https://ariakit.com/reference/dialog) interactions with persistent elements, including elements across open shadow roots, before the dialog receives focus. Thanks to [@yishayhaz](https://github.com/yishayhaz) and [@gonzoblasco](https://github.com/gonzoblasco).
```
diegohaz: **[P2] Limit the release note to the newly shipped ShadowRoot behavior**

The broad statement that this fixes persistent-element interactions before focus repeats behavior that is already in the base release. The base implementation already checks `isElementInside(target, contentElement.id)` before the focused-tree gate, and both current package changelogs already say that focusing, clicking, or right-clicking `getPersistentElements` no longer closes an unfocused dialog. This PR newly extends that behavior across open ShadowRoots.

Please keep the first and credit sentences, but narrow the added sentence to the actual delta:

```md
Fixed [`Dialog`](https://ariakit.com/reference/dialog) so interacting with elements returned by [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements) across open shadow roots no longer closes the dialog before it receives focus.
```

**Assessment**

- Severity: Low (verified), because runtime behavior is unaffected but the generated changelog overstates this release.
- Priority: P2 (estimated, 95%), because release notes should describe only the behavior newly shipped here.
- Breakage likelihood: Likely (verified), because the next generated changelog will deterministically repeat the broad already-released fix.
- Fix effort: Easy (verified), because this is a one-sentence scope correction.

app/src/sandbox/dialog-5238/index.react.tsx
diegohaz: The runtime issue in this thread is resolved at `7c783114d`; I independently revalidated the default ShadowRoot path and the focused cross-browser matrix.

**[P2] Keep the PR Preview synchronized with the resolved fix**

One publication gap remains: the PR body's `## Preview` still contains only the earlier document-root Apples/Oranges recording. The browser-visible composed-path and persistent-element behavior added by `eeb8ab34d` and `7c783114d` is demonstrated only in this resolved thread. Please reuse the exact-head artifact under `## Preview` with a caption describing the ShadowRoot interaction:

[ShadowRoot dialog interaction recording](https://github.com/user-attachments/assets/5186a2d8-bd8f-41d4-a45c-d1deec0c0e0a)

The 22.48-second recording shows switching to ShadowRoot rendering, interacting with Apples, closing the foreground dialog, and then interacting with Oranges.

**Assessment**

- Severity: Low (verified), because implementation behavior is unaffected but the PR's primary evidence remains stale.
- Priority: P2 (estimated, 95%), because the description should expose the current browser-visible result before merge.
- Breakage likelihood: Likely (verified), because every PR reader currently sees only the pre-fix document-root recording.
- Fix effort: Easy (verified), because the exact-head attachment already exists.

diegohaz (commented):
## Re-review summary

This re-review used the latest prior marked review head, [`88f6e30b`](https://github.com/ariakit/ariakit/commit/88f6e30bb99109f0972a8a7828c1d4bfca80a26d), as the incremental baseline and reviewed [`88f6e30b...7c783114`](https://github.com/ariakit/ariakit/compare/88f6e30bb99109f0972a8a7828c1d4bfca80a26d...7c783114d805f128397b848d8ab1baf3cfbdf51a), plus the surrounding dialog marking/event code newly exposed by the composed-path fix. I independently revalidated the relevant prior threads against the exact current head.

I found four Address now issues:

- [P1: cross-root inside marks collide when independent document trees use the same valid `Dialog` ID, so an unrelated dialog can suppress outside dismissal.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635034591)
- [P1: the new regression suite never reaches the changed preceding-mousedown path, leaving `elements` and `rootTarget` behavior unprotected.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635035742)
- [P2: the changeset repeats the already-shipped general persistent-element fix instead of limiting the new release claim to open ShadowRoots.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635036801)
- [P2: the PR Preview still shows only the pre-fix document-root recording instead of also surfacing the existing exact-head ShadowRoot evidence.](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635037789)

The composed-path/root-local projection design otherwise looks sound: it preserves positive membership through nested open ShadowRoots, uses a root-local target for outside marks, retains the disconnected-origin guard, and keeps cross-realm element checks safe.

### Prior finding outcomes

- [`PRRT_kwDOBjT0Fc6TFWD3`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634206402) remains technically resolved. The #5238 fixture now exercises default `hideOnInteractOutside`, and focus, context-menu, foreground dismissal, and background restoration pass in all three browser projects. My new reply is limited to the stale PR Preview evidence.
- [`PRRT_kwDOBjT0Fc6TF7Rj`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3634433223) is resolved for the reported unrelated-ShadowRoot endpoint bug. The current root projection dismisses for a distinct unrelated root while the full path preserves legitimate persistent/nested membership.
- The three older workaround, contributor-credit, and root-local lookup threads remain resolved. GitHub's authoritative GraphQL result was fully paginated: all 5 pre-existing threads were resolved at the start of this re-review.

### Verification

- `pnpm test-react app/src/sandbox/dialog-5238/test.ts app/src/sandbox/dialog-6344/test.ts`: 9/9 passed.
- `CI=true pnpm test-react18 app/src/sandbox/dialog-5238/test.ts app/src/sandbox/dialog-6344/test.ts`: 9/9 passed.
- Fresh `pnpm build-app-lite`: passed at the exact head.
- `APP_PORT=49931 NEXTJS_PORT=50000 APP_INSPECTOR_PORT=49939 NEXTJS_INSPECTOR_PORT=50009 pnpm -F app run test --project=chrome --project=firefox --project=safari src/sandbox/dialog-5238/test-browser.ts src/sandbox/dialog-6344/test-browser.ts`: 27/27 passed.
- Same-ID red/green proof: a temporary light-DOM dialog and unrelated open-ShadowRoot dialog shared one valid ID. The outside dialog stayed visible in happy-dom and Chromium; changing only the second ID made both tests pass. [Focused failure capture](https://github.com/user-attachments/assets/3902d973-7643-4643-ba8e-53694ef6b487).
- Preceding-mousedown coverage proof: temporarily returning immediately after the `previousMouseDown` null guard still left all 9 focused React tests green. All experiment changes were restored.
- The existing [22.48-second ShadowRoot recording](https://github.com/user-attachments/assets/5186a2d8-bd8f-41d4-a45c-d1deec0c0e0a) is truthful and focused on the current interaction flow.
- GitHub reports 49/49 exact-head checks passing, including React 18, Chrome, Firefox, Safari, Android, macOS, type checking, visual finalization, and dialog performance. The performance report found no significant Chrome or Node change.
- `git diff --check` passed; the worktree is clean at [`7c783114`](https://github.com/ariakit/ariakit/commit/7c783114d805f128397b848d8ab1baf3cfbdf51a).

### Validation gaps

- The same-ID red/green reproduction ran in Chromium and happy-dom, not Firefox or WebKit.
- A standalone local `pnpm tsc` did not start because its dependency-layout refresh hit sandboxed registry access. The exact-head CI type check is green, and both the Next.js and Astro exact-head builds completed successfully.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2). Inside membership is now keyed by the concrete dialog element rather than its ID:

```ts
insideElements.set(dialog, marker);
return isElementInside(target, contentElement);
```

This keeps standards-valid same IDs isolated across independent roots while preserving the full composed-path scan for legitimate persistent and nested content. The app fixture now renders two dialogs with `id="dialog"` in separate roots; that regression failed in happy-dom and Chromium on the prior head and passes after the fix in Chrome, Firefox, and Safari.

[Fixed same-ID ShadowRoot interaction recording](https://github.com/user-attachments/assets/aa0d3731-a319-4b70-b5a5-137fe92e6fde)

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Covered in [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2). The implementation already retained both views of the preceding mousedown; the public fixture now reaches each branch directly:

```ts
await mouseDown(persistentButton);
await dispatch.mouseUp(outsideButton);
await dispatch.click(outsideButton);
expect(q.dialog("Dialog")).toBeVisible();
```

A focus-preventing outside ShadowRoot button verifies `rootTarget` dismissal, and the split persistent-origin drag verifies the full `elements` path. The first test failed when I temporarily bypassed stored-mousedown handling; the second failed when I temporarily disabled previous-path membership. Both now pass in happy-dom and all three browser projects.

.changeset/300-dialog-sibling-modal.md
diegohaz: Fixed in [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2). The changeset now limits the additional release claim to `getPersistentElements` interactions across open shadow roots, using the suggested wording and verified `#getpersistentelements` reference link. The existing @yishayhaz and @gonzoblasco acknowledgments are preserved.

diegohaz:
**Fixed**

- [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2) — Keyed inside membership by the concrete dialog element, isolating same-ID dialogs across independent roots, with red/green happy-dom and cross-browser coverage plus a [fixed-state recording](https://github.com/user-attachments/assets/aa0d3731-a319-4b70-b5a5-137fe92e6fde). [Reply](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635176734)
- [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2) — Narrowed the changeset to the newly shipped open-ShadowRoot `getPersistentElements` behavior while preserving contributor acknowledgments. [Reply](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635177570)

**Already resolved**

- [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2) — Confirmed the existing full-path/root-target preceding-mousedown implementation and added public no-focus dismissal and persistent-origin drag regressions that fail when either branch is bypassed. [Reply](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635177157)

.changeset/300-dialog-sibling-modal.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed sibling modal [`Dialog`](https://ariakit.com/reference/dialog) components and modal components built on them, such as [`Popover`](https://ariakit.com/reference/popover) and [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), rendered in their default portals so opening them in the same render no longer made each other inert. Fixed [`Dialog`](https://ariakit.com/reference/dialog) so interacting with elements returned by [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements) across open shadow roots no longer closes the dialog before it receives focus. Thanks to [@yishayhaz](https://github.com/yishayhaz) and [@gonzoblasco](https://github.com/gonzoblasco).
```
diegohaz: There are two sentences. Use one or split this into two changesets if they refer to different changes.

app/src/sandbox/dialog-6344/test-browser.ts
```diff
@@ -52,6 +58,54 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+    q,
+  }) => {
+    await q.button("Open dialog").click();
+    await test.expect(q.dialog("Dialog")).toBeVisible();
+
+    const persistentInput = q.textbox("Persistent shadow field");
+    // Dispatch directly on the nested shadow element. WebKit's click
+    // actionability check may otherwise treat the outer shadow host as the
+    // pointer target even though the nested input is visible.
+    await persistentInput.dispatchEvent("mousedown", mouseEventInit);
```
diegohaz: Doesn't `page.mouse.down` work here?

packages/ariakit-react-components/src/dialog/utils/tree-cleanup.ts
```diff
@@ -6,7 +6,9 @@ export type Elements = Array<Element | null>;
 export type Cleanups = Array<() => void>;
 export type Ids = Array<string | undefined>;
 
-type MarkKind = "outside" | "ancestor" | "inside";
+type MarkKind = "outside" | "ancestor";
+
+const insideElements = new WeakMap<Element, WeakSet<Element>>();
```
diegohaz, line 14: Explain why this is needed with comments

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
```diff
@@ -73,27 +97,23 @@ function useEventOutside({
...
+      if (isEventInsideDialog(targets, contentElement, disclosureElement)) {
+        return;
+      }
       // Clicked on dialog's bounding box
       if (isMouseEventOnDialog(event, contentElement)) return;
-      // The dialog itself, persistent elements, and nested dialogs are marked
-      // as "inside" when the dialog opens. Events on them must never trigger
-      // the outside listeners, even before the dialog has been focused (for
-      // example, with autoFocusOnShow={false}). See
-      // https://github.com/ariakit/ariakit/issues/6344
```
diegohaz, line 95: This comment is lost

packages/ariakit-react-components/src/dialog/utils/use-previous-mouse-down-ref.ts
```diff
@@ -1,19 +1,45 @@
...
 import { useEffect, useRef } from "react";
 
-export function usePreviousMouseDownRef(enabled?: boolean, scope?: Window) {
-  const previousMouseDownRef = useRef<EventTarget | null>(null);
+export interface EventTargets {
+  elements: Element[];
+  rootTarget: EventTarget | null;
+}
+
+export function getEventTargets(
```
diegohaz, line 13: This could use some JSDoc explaining why this exists (based on how it's used by `use-hide-on-interact-outside` and other consumers if any).

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -98,6 +99,32 @@ function getElementFromProp(
   return element;
 }
 
+function getLaterOpenModalPortals(dialog: HTMLElement) {
```
diegohaz, line 107: Need why comments

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -289,6 +320,33 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
   }, [open, mounted, domReady]);
 
   const canTakeTreeSnapshot = open && domReady;
+  const openingCohortRef = useRef<{
+    portal: HTMLElement;
+    peers: HTMLElement[];
+  }>(null);
+
+  useSafeLayoutEffect(() => {
```
diegohaz, line 336: Need why comment

.changeset/300-dialog-sibling-modal.md
diegohaz: **Disposition: Address now.** Confirmed. These are two independent public behavior summaries, so split them into two patch changesets. Keep the sibling-dialog summary and the @yishayhaz/@gonzoblasco credit together; use a separate one-sentence entry for the cross-open-ShadowRoot `getPersistentElements` fix (including `Dialog`-derived components such as `Popover` and `ComboboxPopover`).

**Assessment**

- Severity: Low (verified); runtime behavior is unaffected, but the generated release entry violates the one-summary rule.
- Priority: P2 (estimated, 95%).
- Breakage likelihood: Likely (verified); the combined entry is deterministic at release generation.
- Fix effort: Easy (estimated, 90%).

packages/ariakit-react-components/src/dialog/utils/tree-cleanup.ts
diegohaz: **Disposition: Address now.** Confirmed. The nearby comments explain the map operations, but not why this concrete weak key is required. Please document that DOM IDs are only tree-scoped, so keying by the dialog element isolates standards-valid same-ID dialogs in separate roots; `WeakMap`/`WeakSet` also avoids retaining or mutating DOM nodes.

**Assessment:** Severity Low (verified); Priority P2 (estimated, 95%); breakage likelihood Likely (verified, because the rationale is currently absent); fix effort Easy (verified).

app/src/sandbox/dialog-6344/test-browser.ts
diegohaz: **Disposition: Address now.** Yes, real `page.mouse` input works here, but `persistentInput.hover()` is not portable: Chrome and Firefox passed, while Safari timed out because the outer shadow host intercepts locator actionability. Moving to the nested input's measured coordinates bypasses that gate while retaining actual browser dispatch:

```ts
const box = await persistentInput.boundingBox();
if (!box) throw new Error("Persistent input has no box");
await page.mouse.move(box.x + box.width / 2, box.y + box.height / 2);
await page.mouse.down();
await persistentInput.focus();
await page.mouse.up();
```

I substituted this exact sequence temporarily and the focused test passed in Chrome, Firefox, and Safari (3/3). The synthetic events work, but the coordinate-based sequence exercises browser hit testing and dispatch, so I recommend using it.

**Assessment**

- Severity: Low (verified); this affects browser-test fidelity, not shipped runtime behavior.
- Priority: P2 (estimated, 95%).
- Breakage likelihood: Likely (verified); the current test deterministically bypasses native pointer dispatch.
- Fix effort: Easy (verified).

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: **Disposition: Address now.** Confirmed. Restore the #6344 rationale next to the consolidated membership check: persistent and nested elements must suppress outside listeners before the dialog has ever received focus (including `autoFocusOnShow={false}`), and scanning the composed path extends that guarantee through open ShadowRoots. The current comment says what is classified as inside, but no longer explains why the pre-focus positive mark is necessary.

**Assessment:** Severity Low (verified); Priority P2 (estimated, 95%); breakage likelihood Likely (verified, because the rationale was deterministically removed); fix effort Easy (verified).

packages/ariakit-react-components/src/dialog/utils/use-previous-mouse-down-ref.ts
diegohaz: **Disposition: Address now.** Confirmed. `getEventTargets` needs JSDoc explaining its two deliberately different views: `elements` preserves the full composed path for positive dialog/persistent membership (including stored drag origins), while `rootTarget` projects the event into the reference element's root for outside-tree marks. `usePreviousMouseDownRef` should likewise say that it stores those views so opening clicks and drag releases are classified against the mousedown origin. `useHideOnInteractOutside` is the only current consumer.

**Assessment:** Severity Low (verified); Priority P2 (estimated, 95%); breakage likelihood Likely (verified, because this non-obvious contract is undocumented); fix effort Easy (verified).

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: **Disposition: Address now.** Confirmed. Add JSDoc explaining that this returns default modal portal roots that are later in the same root and still pending during the current layout pass. Established entries in `openModalPortals` are intentionally excluded so an existing dialog stack keeps its prior relationship; this helper only coordinates the one opening cohort.

**Assessment:** Severity Low (verified); Priority P2 (estimated, 95%); breakage likelihood Likely (verified, because the lifecycle distinction is currently implicit); fix effort Easy (verified).

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: **Disposition: Address now.** Confirmed. The StrictMode comment explains one persistence detail, but not the effect's lifecycle role. Please explain that this earlier layout effect registers the dialog's own default portal before the later tree-marking effect, captures only later pending peers for the current open cycle, and retains that cohort through snapshot refreshes/StrictMode replay.

**Assessment:** Severity Low (verified); Priority P2 (estimated, 95%); breakage likelihood Likely (verified, because the ordering contract is currently implicit); fix effort Easy (verified).

diegohaz (commented):
## Re-review of `19e5dac538`

I reviewed the incremental change from the last marked head (`7c783114d8`) through the current exact head and rechecked the surrounding dialog lifecycle, outside-interaction, browser coverage, release metadata, and PR evidence.

**Outcome**

- One new P1 runtime finding is inline: concrete-element inside marks become stale when an open `Dialog` replaces its rendered DOM node.
- All seven newly open reviewer threads are confirmed **Address now**: split the two changeset summaries; use coordinate-based real pointer input in the nested-ShadowRoot browser case; and restore/add the requested rationale/JSDoc around weak instance keys, composed/root event targets, pending portal discovery, and opening-cohort ordering.
- The prior same-ID collision and stored-mousedown branch gaps are resolved and independently revalidated. The older broad changeset-scope concern is also technically resolved, although that outdated thread remains unresolved in GitHub.
- The prior Preview P2 remains open: the PR body still shows only the document-root recording. Please add the [exact-head same-ID ShadowRoot recording](https://github.com/user-attachments/assets/aa0d3731-a319-4b70-b5a5-137fe92e6fde) under `## Preview`.
- I found no additional security, performance, or worthwhile simplification issue. The `WeakMap`/`WeakSet` direction remains preferable once its element-lifecycle dependency is fixed.

**Verification**

- Focused React 19: 25/25 passed.
- Focused React 18: 25/25 passed from a clean generated workspace.
- Exact-head browser matrix for the two dialog fixtures: 36/36 passed in Chrome, Firefox, and Safari.
- Coordinate-based `page.mouse` alternative for the anchored nested-ShadowRoot interaction: 3/3 passed; plain locator `hover()` failed only in Safari because the host intercepted actionability.
- Temporary dynamic-render regression: red on the exact head, green after adding only the `contentElement` dependency; the failure was also reproduced in Chromium.
- `pnpm build-app-lite` and `git diff --check` passed; GitHub reports 49/49 checks green with no significant performance change.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -324,7 +384,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     // the dialog is open (hovercards and tooltips set it to the focus
     // source), so the listeners re-check it against the current state on
     // every event instead. See https://github.com/ariakit/ariakit/issues/6344
-    const restoreInsideMarks = markTreeInside(id, allElements);
+    const restoreInsideMarks = markTreeInside(dialog, allElements);
```
diegohaz, line 394: **[P1] Re-mark persistent elements when the dialog node changes**

`markTreeInside` is now keyed by the concrete dialog element, but this effect reads `ref.current` without depending on the already-subscribed `contentElement`. An open `Dialog` can replace its DOM node—for example, a responsive/custom `render` can switch `<div>` to `<section>`. After that replacement, the positive marks remain keyed by the removed element while `useHideOnInteractOutside` queries them with the new `contentElement`. The next focus/click on a `getPersistentElements` target is therefore classified as outside and closes the dialog.

A public reproduction is:

```tsx
<Ariakit.Dialog
  render={asSection ? <section /> : undefined}
  autoFocusOnShow={false}
  getPersistentElements={() => [notificationsRef.current]}
>
  <button onClick={() => setAsSection(true)}>Replace dialog element</button>
</Ariakit.Dialog>
```

On the exact head, a temporary regression test failed because the dialog became `null` after clicking the persistent notification field. Adding only `contentElement` to this effect's dependencies made all 11 focused tests pass. I also reproduced the failure through Chromium: the root changed to `SECTION`, the notification field received focus, and the dialog disappeared.

The persistent notification field is focused after the dynamically rendered dialog has unexpectedly closed

Please re-run the marking effect when the concrete element changes, for example:

```tsx
const dialog = contentElement ?? ref.current;
// ...
}, [
  // ...
  contentElement,
]);
```

The concrete weak-key design can otherwise stay.

**Assessment**

- Severity: Medium (estimated, 95% confidence), because the popup closes and interrupts the consumer workflow.
- Priority: P1 (estimated, 95% confidence), because this is a regression introduced by the element-keyed correction.
- Breakage likelihood: Unlikely (estimated, 85% confidence), because it requires replacing an open dialog node, disabled/rejected autofocus, and a later persistent interaction.
- Fix effort: Easy (estimated, 90% confidence), because tracking `contentElement` fixed the focused regression.

diegohaz: If this is as easy to fix as the comment above suggests, that's fine. Otherwise, we don't want to address `render` edge cases with complex solutions.

.changeset/300-dialog-sibling-modal.md
diegohaz: Fixed in [`6926ad25b`](https://github.com/ariakit/ariakit/commit/6926ad25b023162ddc312192605d64c484175c44). I split the release notes into two patch changesets: the sibling default-portal dialog entry retains the @yishayhaz/@gonzoblasco acknowledgment, while the separate one-sentence entry covers cross-open-ShadowRoot `getPersistentElements` behavior for `Dialog`, `Popover`, and `ComboboxPopover`.

app/src/sandbox/dialog-6344/test-browser.ts
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). The test now measures the nested input, moves the real mouse to its center, runs `page.mouse.down()`, focuses the input, then runs `page.mouse.up()`. This exact sequence passed the full dialog-6344 browser matrix in Chrome, Firefox, and Safari (33/33).

packages/ariakit-react-components/src/dialog/utils/tree-cleanup.ts
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). The comment now explains both reasons for the concrete weak key: DOM IDs are only tree-scoped, so the dialog element isolates same-ID dialogs in separate roots; `WeakMap`/`WeakSet` also avoids retaining or mutating membership-only nodes.

packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). I restored the #6344 rationale beside the consolidated membership check: persistent and nested elements must suppress outside listeners before first focus, including `autoFocusOnShow={false}`, and the composed-path scan extends that positive mark through open ShadowRoots.

packages/ariakit-react-components/src/dialog/utils/use-previous-mouse-down-ref.ts
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). `getEventTargets` now documents the full composed-path membership view versus the root-projected outside-mark view, and `usePreviousMouseDownRef` documents why both views are retained for opening clicks and drag releases classified from their mousedown origin.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). The helper JSDoc now states that it returns only later default modal portals in the same root that are still pending in the current layout pass; established open portals are excluded so coordination stays limited to the current opening cohort.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d). The lifecycle comment now explains that this earlier layout effect registers the dialog's own default portal before tree marking, captures only later pending peers for the current open cycle, and retains that cohort through snapshot refreshes and StrictMode replay.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d) with the narrow dependency change suggested here: the existing tree-marking effect now uses `contentElement ?? ref.current` and re-runs when `contentElement` changes. No registry or renderer-lifecycle subsystem was added.

The public happy-dom and browser regressions replace the open root with `SECTION`, focus the persistent notification field, and assert that the dialog remains visible. Both tests failed before the dependency fix (including Chromium), then passed; the focused dialog-6344 suites are green under React 19 and React 18 (11/11 each), and the browser matrix is green in Chrome, Firefox, and Safari (33/33).

[Exact-head browser evidence](https://github.com/user-attachments/assets/6da8a1db-be4c-4af6-aa74-545fd6a9b1ab)

diegohaz:
**Fixed**

- [`6926ad25b`](https://github.com/ariakit/ariakit/commit/6926ad25b023162ddc312192605d64c484175c44) splits the independent sibling-dialog and cross-ShadowRoot persistent-element release notes while keeping the @yishayhaz/@gonzoblasco acknowledgment with the sibling fix ([reply](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635709250)).
- [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d) replaces synthetic pointer dispatch with a measured real `page.mouse` sequence ([reply](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635709507)); documents the weak-key, composed-path/root-target, pre-focus, pending-portal, and layout-effect lifecycle rationale ([replies](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635709859)); and re-marks persistent elements when an open dialog replaces its DOM node ([reply with exact-head recording](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635711140)).
- Verification is green: focused React 19 (26/26), focused React 18 (13/13), the dialog-6344 Chrome/Firefox/Safari browser matrix (33/33), TypeScript, full build, and exact-commit app build. Both the happy-dom and Chromium replacement regressions failed before the `contentElement` dependency fix and passed afterward.

diegohaz (commented):
## Re-review of `bec088fb5`

I used the latest prior marked head, [`19e5dac53`](https://github.com/ariakit/ariakit/commit/19e5dac5383061bf704a2cb3f62eba917edc16b2), as the incremental baseline and reviewed [`19e5dac53...bec088fb5`](https://github.com/ariakit/ariakit/compare/19e5dac5383061bf704a2cb3f62eba917edc16b2...bec088fb5ec4764943a8d85279037ae6e862af1d), plus the current dialog marking, outside-interaction, example coverage, release notes, and PR body.

### Outcome

I found no new runtime, compatibility, security, maintainability, simplification, or performance issue in the two new commits. The local `contentElement` dependency is the narrow behavior-preserving fix for replaced dialog nodes, and the real pointer sequence and added rationale match the surrounding contracts.

One previously published P2 remains unaddressed: [the PR Preview still shows only the original document-root recording](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635037789). The current `## Preview` does not surface the open-ShadowRoot and persistent-element behavior added while resolving the earlier runtime findings. Please add the existing [ShadowRoot interaction recording](https://github.com/user-attachments/assets/aa0d3731-a319-4b70-b5a5-137fe92e6fde) with a focused caption. The recording still resolves as a WebM, and the later commits do not change that interaction path.

**Assessment**

- Severity: Low (verified), because runtime behavior is unaffected but the implementation evidence in the PR body remains stale.
- Priority: P2 (estimate, 95% confidence), because the PR description should expose the browser-visible ShadowRoot result before merge.
- Breakage likelihood: Likely (verified), because every current PR reader sees only the document-root recording.
- Fix effort: Easy (verified), because the existing recording can be linked directly from `## Preview`.

### Review-thread validation

- [`PRRT_kwDOBjT0Fc6TI9eE`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635586651) is resolved. I independently reproduced the node-replacement failure after removing only the `contentElement` dependency, then restored the exact head and confirmed the regression passes.
- The changeset split, real pointer input, weak-key rationale, composed-path JSDoc, pre-focus rationale, pending-portal JSDoc, and opening-cohort lifecycle comments are resolved at the current head.
- [`PRRT_kwDOBjT0Fc6THg6p`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635036801) remains unresolved in GitHub but is already resolved in the current code. The separate persistent-element changeset limits the release claim to open shadow roots, so no duplicate reply is needed.

### Verification

- `pnpm test-react app/src/sandbox/dialog-5238/test.ts app/src/sandbox/dialog-6344/test.ts`: 13/13 passed.
- `CI=true pnpm test-react18 app/src/sandbox/dialog-5238/test.ts app/src/sandbox/dialog-6344/test.ts`: 13/13 passed.
- `CI=true APP_PORT=52131 NEXTJS_PORT=52200 APP_INSPECTOR_PORT=52139 NEXTJS_INSPECTOR_PORT=52209 pnpm -F app run test --project=chrome --project=firefox --project=safari src/sandbox/dialog-5238/test-browser.ts src/sandbox/dialog-6344/test-browser.ts`: 39/39 passed.
- `CI=true pnpm build-app-lite`, `CI=true pnpm tsc`, and `git diff --check`: passed.
- Red/green proof: removing the `contentElement` dependency made the focused replacement test fail because the dialog closed after focusing the persistent field; the submitted head passes. I ran the red side in happy-dom and the green side in happy-dom plus Chrome, Firefox, and WebKit.
- All three release-note reference pages return HTTP 200, and the Dialog page contains the `getPersistentElements` anchor.
- GitHub reports 49/49 checks successful. The exact-head performance report found no significant Chrome or Node change.

The worktree is clean at `bec088fb5`.

app/src/sandbox/dialog-6344/test-browser.ts
```diff
@@ -52,6 +58,70 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+    await page.mouse.down();
+    await persistentInput.focus();
+    await page.mouse.up();
+    await test.expect(persistentInput).toBeFocused();
+    await page.waitForTimeout(250);
+    await test.expect(q.dialog("Dialog")).toBeVisible();
+
+    await q.textbox("Inside field").click();
+    const persistentButton = q.button("Persistent shadow button");
+    await persistentButton.dispatchEvent("mousedown", mouseEventInit);
```
diegohaz: Why not `page.mouse.down`?

app/src/sandbox/dialog-6344/test-browser.ts
```diff
@@ -61,4 +131,51 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+  // Regression for https://github.com/ariakit/ariakit/pull/6810#discussion_r3635035742
+  test("stays open when dragging from persistent shadow content", async ({
+    q,
+  }) => {
+    await q.button("Open dialog").click();
+    await test.expect(q.dialog("Dialog")).toBeVisible();
+
+    const persistentButton = q.button("Persistent shadow button");
+    const outsideButton = q.button("Outside shadow button");
+    await persistentButton.dispatchEvent("mousedown", mouseEventInit);
```
diegohaz: Same here?

app/src/sandbox/dialog-6344/test.ts
```diff
@@ -38,10 +38,126 @@ test("keeps the documented behavior after the dialog has been focused", async ()
   expect(q.dialog("Dialog")).toBeVisible();
 });
 
+// Regression for https://github.com/ariakit/ariakit/pull/6810#discussion_r3635586651
+test("keeps persistent elements after replacing an open dialog node", async () => {
+  await click(q.button("Open dialog"));
+  await click(q.button("Replace dialog element"));
+  expect(q.dialog.ensure("Dialog").tagName).toBe("SECTION");
```
diegohaz: What about this?
```suggestion
  expect(q.dialog("Dialog")?.tagName).toBe("SECTION");
```

app/src/sandbox/dialog-6344/test-browser.ts
diegohaz: Updated in [`36617002f`](https://github.com/ariakit/ariakit/commit/36617002faa0af53ffd8cc5ab4f4775c0cae6d6b). The persistent-button path now positions the real pointer over the button and uses `page.mouse.down()` / `page.mouse.up()`. The fixture's `mousedown.preventDefault()` keeps focus inside the dialog, so the focus assertion still verifies the intended no-focus interaction path. The complete sandbox file passes across Chrome, Firefox, and Safari (33/33).

app/src/sandbox/dialog-6344/test.ts
diegohaz: Applied this suggestion in [`36617002f`](https://github.com/ariakit/ariakit/commit/36617002faa0af53ffd8cc5ab4f4775c0cae6d6b). The happy-dom regression now uses the concise optional tag-name assertion, and the focused React 19 and React 18 suites both pass 11/11.

app/src/sandbox/dialog-6344/test-browser.ts
diegohaz: Updated in [`36617002f`](https://github.com/ariakit/ariakit/commit/36617002faa0af53ffd8cc5ab4f4775c0cae6d6b). The drag regression now moves the real pointer to the persistent shadow button, calls `page.mouse.down()`, moves to the outside shadow button while the button remains held, and calls `page.mouse.up()`. This passes across Chrome, Firefox, and Safari as part of the 33/33 sandbox matrix.

diegohaz:
Addressed the latest review feedback in [`36617002f`](https://github.com/ariakit/ariakit/commit/36617002faa0af53ffd8cc5ab4f4775c0cae6d6b):

- Replaced the remaining synthetic persistent-shadow-button events with real `page.mouse.down()` / `page.mouse.up()` input.
- Converted the persistent-origin drag regression to a real pointer move/down/move/up sequence.
- Applied the suggested concise happy-dom tag-name assertion.

Verification:

- Focused React 19: 11/11
- Focused React 18: 11/11
- Chrome, Firefox, and Safari browser matrix: 33/33
- `pnpm tsc`
- `pnpm lint-fix`
- Exact staged-snapshot native review: no findings
- Full PR CI: green. The initial Plus website build hit a transient npm dist-tags fetch failure; the exact build passed locally, and the failed job passed on rerun without code changes.

All three new threads have direct replies and remain available for reviewer-owned resolution. Contributor attribution remains intact: [@gonzoblasco](https://github.com/gonzoblasco) has GitHub-linked commit credit in [`88f6e30bb`](https://github.com/ariakit/ariakit/commit/88f6e30bb99109f0972a8a7828c1d4bfca80a26d), and the changeset and PR continue to thank [@yishayhaz](https://github.com/yishayhaz) and [@gonzoblasco](https://github.com/gonzoblasco).

diegohaz (commented):
## Re-review of `36617002f`

I used the latest prior marked head, [`bec088fb5`](https://github.com/ariakit/ariakit/commit/bec088fb5ec4764943a8d85279037ae6e862af1d), as the incremental baseline and reviewed [`bec088fb5...36617002f`](https://github.com/ariakit/ariakit/compare/bec088fb5ec4764943a8d85279037ae6e862af1d...36617002faa0af53ffd8cc5ab4f4775c0cae6d6b), plus the surrounding ShadowRoot fixture, real-pointer event sequence, stored-mousedown classification path, current PR body, and authoritative unresolved-thread state.

### Outcome

I found no new correctness, compatibility, security, maintainability, simplification, performance, or test-coverage issue. The synthetic `dispatchEvent` calls are now replaced by physical pointer movement plus `page.mouse.down()` / `page.mouse.up()`, so the two regressions exercise the browser's real composed input sequence. The small `moveMouseTo` helper also reuses the coordinate-based approach that avoids WebKit's nested-ShadowRoot actionability problem.

One previously published P2 remains unaddressed: [the PR Preview still shows only the original document-root recording](https://github.com/ariakit/ariakit/pull/6810#discussion_r3635037789). The current `## Preview` still does not surface the open-ShadowRoot and persistent-element behavior added while resolving earlier runtime findings. Please add the existing [ShadowRoot interaction recording](https://github.com/user-attachments/assets/aa0d3731-a319-4b70-b5a5-137fe92e6fde) with a focused caption; I reverified that both this recording and the current document-root recording resolve as WebM videos.

**Assessment**

- Severity: Low (verified), because runtime behavior is unaffected but the browser evidence in the PR body remains incomplete.
- Priority: P2 (estimate, 95% confidence), because the PR description should expose the newly covered ShadowRoot result before merge.
- Breakage likelihood: Likely (verified), because every current PR reader sees only the document-root recording.
- Fix effort: Easy (verified), because the existing recording can be linked directly from `## Preview`.

### Review-thread validation

- [`PRRT_kwDOBjT0Fc6TKGoK`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3636013264) is technically resolved. The persistent-button case now uses a real down/up sequence and preserves focus inside the dialog as intended.
- [`PRRT_kwDOBjT0Fc6TKG0k`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3636014374) is technically resolved. The drag case now holds the real pointer from the persistent shadow button through release over the outside shadow button.
- [`PRRT_kwDOBjT0Fc6TKHUA`](https://github.com/ariakit/ariakit/pull/6810#discussion_r3636017087) is technically resolved. The requested concise tag-name assertion is present and passes under both React versions.

These three threads remain unresolved in GitHub and are outdated, but their requested changes are present at the exact head. Their existing replies already describe the same resolution, so I did not duplicate replies.

### Verification

- `CI=true pnpm test-react app/src/sandbox/dialog-6344/test.ts`: 11/11 passed.
- `CI=true pnpm test-react18 app/src/sandbox/dialog-6344/test.ts`: 11/11 passed.
- `CI=true APP_PORT=53131 NEXTJS_PORT=53200 APP_INSPECTOR_PORT=53139 NEXTJS_INSPECTOR_PORT=53209 pnpm -F app run test --project=chrome --project=firefox --project=safari src/sandbox/dialog-6344/test-browser.ts`: 33/33 passed, with the server output confirming this worktree and all three intended projects.
- Focused `oxfmt` and `oxlint`, plus `git diff --check`: passed.
- All 45 completed GitHub check runs succeeded; the other four external status contexts currently report no conclusion. CI includes builds, type checking, React 18, Chrome, Firefox, Safari, Android, macOS, CodeQL, and dialog performance.

The worktree is clean at `36617002f`.

PR #6828, opened 2026-07-24
Add ComboboxSelect and ComboboxSelectLabel components

### LINKED ISSUE #6796: Add ComboboxSelect to unify Select and Combobox
### Motivation

The Select and Combobox modules provide overlapping selection behavior through separate component sets. This makes it difficult to combine Select features with Combobox composition, particularly for filterable selects and form interoperability.

Adding a dedicated select trigger to the Combobox module would let standard selects and filterable selects share the same provider, popover, items, and selection state. It would also provide a consistent focus model in which DOM focus stays on the trigger, as it does with Combobox and the native select.

This was discussed in [#5058](https://github.com/ariakit/ariakit/issues/5058#issuecomment-5035405248).

### Usage example

Default select:

```jsx
<ComboboxProvider>
  <ComboboxSelect />
  <ComboboxPopover>
    <ComboboxItem />
  </ComboboxPopover>
</ComboboxProvider>
```

Filterable select:

```jsx
<ComboboxProvider>
  <ComboboxSelect />
  <ComboboxPopover>
    <ComboboxInput />
    <ComboboxList>
      <ComboboxItem />
    </ComboboxList>
  </ComboboxPopover>
</ComboboxProvider>
```

### Requirements

- `Combobox` should also be exported as `ComboboxInput` (as well as types like `ComboboxInputProps`) so as to not cause confusion.
- The Combobox module should expose a `ComboboxSelect` component that works with `ComboboxProvider`, `ComboboxPopover`, and `ComboboxItem`.
- The Combobox module should expose a select-specific label component without requiring `ComboboxLabel` to target both the select trigger and the combobox input.
- `ComboboxSelect` should support the features currently available through `Select`.
- A `Combobox` input, or a `ComboboxInput` alias, should be composable inside the same popover to create a filterable select.
- DOM focus should remain on `ComboboxSelect` while navigating items with virtual focus, which means `ComboboxSelect` should probably be added as a persistent element when using `<ComboboxPopover modal>` (the same as today's `Combobox` I believe).
- Existing Combobox input behavior should remain supported.
- We should probably have a `ComboboxBase` component or something for the shared behavior between `Combobox`/`ComboboxInput` (the alias) and `ComboboxSelect`.

### Workaround

Use the current `Select` components for a standard select. For a filterable select, coordinate separate Select and Combobox controls with shared controlled state in userland.

### Possible implementations

Add `ComboboxSelect` and `ComboboxSelectLabel` to the Combobox module and reuse the Combobox store and item components. Keep the input and select labeling paths separate to avoid the complexity explored in [#6279](https://github.com/ariakit/ariakit/pull/6279). The select variant could use virtual focus so DOM focus remains on the select button while the popover is open.

diegohaz:
Closes #6796

## Motivation

The Select and Combobox modules provide overlapping selection behavior through separate component sets, which makes it hard to combine Select features with Combobox composition, particularly for filterable selects and form interoperability. Building a filterable select today requires coordinating a Select store and a Combobox store, duplicated state, and composition-only patterns such as `SelectItem render={<ComboboxItem />}`.

This was discussed in #5058 and specified in #6796: the Combobox module should expose a dedicated select trigger so standard selects and filterable selects share the same provider, popover, items, and selection state, with a consistent focus model where DOM focus stays on the trigger, as it does with `Combobox` and the native select element.

## Solution

Add `ComboboxSelect` and `ComboboxSelectLabel` to the Combobox module, reusing the combobox store and item components.

Standard select (no filter input): the trigger keeps DOM focus and drives the listbox with virtual focus through `aria-activedescendant`:

```tsx
<ComboboxProvider defaultSelectedValue="Apple">
  <ComboboxSelectLabel>Favorite fruit</ComboboxSelectLabel>
  <ComboboxSelect />
  <ComboboxPopover>
    <ComboboxItem value="Apple" />
    <ComboboxItem value="Orange" />
  </ComboboxPopover>
</ComboboxProvider>
```

Filterable select: compose a `Combobox` input (now also exported as `ComboboxInput`) inside the same popover:

```tsx
<ComboboxProvider defaultSelectedValue="Apple" resetValueOnHide>
  <ComboboxSelectLabel>Favorite fruit</ComboboxSelectLabel>
  <ComboboxSelect />
  <ComboboxPopover>
    <ComboboxInput autoSelect />
    <ComboboxList>
      <ComboboxItem value="Apple" />
      <ComboboxItem value="Orange" />
    </ComboboxList>
  </ComboboxPopover>
</ComboboxProvider>
```

`ComboboxSelect` supports the features currently available through `Select`:

- A visually hidden native `<select>` element for form submission and browser autofill (`name`, `form`, `required`, `disabled`, multiple selection), including autofill focus forwarding and the `data-autofill` attribute.
- `showOnKeyDown` and `moveOnKeyDown` props: while closed, arrow keys and typeahead move the active item and, in a single-select, update the selection like a native select.
- Open-state keyboard handling mirroring `CompositeItem`: grid-aware arrows with `ComboboxRow` (including column-aware wrapping at the boundaries), `Home`/`End` with grid row semantics, `PageUp`/`PageDown` paging in scrollable listboxes (clamping at the boundaries), typeahead, and `Enter`/`Space` selection. Navigation keys are consumed while the popover is open so the page doesn't scroll behind it.
- Selected item resolution on open (including with `unmountOnHide`), scrolling it into view, and highlighting the first item when the selection is empty.
- Modal popovers: the trigger stays in the modal context as a persistent element, keeping it focusable and interactive while everything else is inert.

Keeping the select label and input label paths separate (per the issue, avoiding the complexity explored in #6279), the store now tracks `labelElement`, `selectLabelElement`, and `selectElement`. `ComboboxPopover` and standalone `ComboboxList` elements are named from the registered labels when they have no explicit ARIA label or `PopoverHeading`, with the select label taking precedence, matching `SelectPopover`/`SelectList`. `ComboboxItem` conveys `aria-selected` and highlights items on hover (popover open only) when a select trigger is registered, matching `SelectItem`, while plain comboboxes keep their previous behavior.

Two shared building blocks changed to support the input-less trigger: `CompositeItem` now adopts its row's id when neither the row nor the store has a base element (previously grid `rowId`s were only assigned once a base element existed, which an input-less select never has), and `findNextPageItemId` is reused for paging from the trigger.

`Combobox` now resolves its store through scoped contexts as well, so an input inside `ComboboxPopover` or `ComboboxList` no longer requires an explicit `store` prop.

## Testing

The `combobox-select` sandbox ports the behaviors covered by the existing Select examples (`select`, `select-grid`, `select-multiple`, `select-form`, `select-autofill`, `select-listbox`, `select-combobox`) to `ComboboxSelect` fixtures: 34 happy-dom tests plus browser tests (Chrome, Firefox, Safari) covering modal interactivity, anchoring, paging, form submission, and native select forwarding. The `combobox-select-perf` sandbox now measures the new composition while keeping the benchmark path, fixture data, and test names stable so perf CI pairs the results with the #6812 baseline.

Not ported in this PR (per the issue discussion): `setValueOnMove`/`resetOnEscape` semantics, the animated/group/custom-item example variants, and a `ComboboxBase` extraction for the shared input/trigger behavior, which can be factored out once both components exist.

## Preview

The preview is available at `/react/previews/combobox-select/` (fixtures) and `/react/previews/combobox-select-perf/` (styled country select). A screen recording of the styled demo (open with selected item highlighted, filter to "United", keyboard selection, focus returning to the trigger, keyboard reopen with the selected item scrolled into view) was captured locally, but the browser extension used to attach files to GitHub was not connected in this session, so it could not be uploaded here.

## Acknowledgments

This PR builds directly on the implementation by [@georgekaran](https://github.com/georgekaran), whose commits it preserves: the `ComboboxSelect`/`ComboboxSelectLabel` components, the store state additions, the `ComboboxInput` alias, the scoped-context resolution change, the sandbox, and the perf sandbox port. Thanks also to [@cloud-walker](https://github.com/cloud-walker) for reporting the form interoperability gap in #5058 that motivated this work.


diegohaz:
Closing in favor of https://github.com/ariakit/ariakit/pull/6832

PR #6832, opened 2026-07-24
Add ComboboxSelect to unify Select and Combobox

diegohaz:
## Motivation

`Select` and `Combobox` currently expose overlapping selection behavior through separate providers, popovers, items, and state. This makes a filterable select require coordinating both component families in userland.

This PR makes standard and filterable selects possible with one Combobox store:

```jsx
<ComboboxProvider>
  <ComboboxSelectLabel>Favorite fruit</ComboboxSelectLabel>
  <ComboboxSelect />
  <ComboboxPopover>
    <ComboboxLabel>Search fruits</ComboboxLabel>
    <ComboboxInput />
    <ComboboxList>
      <ComboboxItem value="Apple" />
    </ComboboxList>
  </ComboboxPopover>
</ComboboxProvider>
```

## Solution

- Adds `ComboboxSelect`, its label and arrow components, hidden native-select form/autofill support, closed-state typeahead, and virtual-focus navigation.
- Exports `Combobox` as `ComboboxInput` and supports rendering it inside `ComboboxPopover` for filtering.
- Ports the remaining Select building blocks needed by the unified pattern, including headings, dismiss controls, virtualization, `ComboboxSelectedValue`, and `ComboboxItemSelected`.
- Extends the Combobox store with select/input/label element state, selected-item restoration, first-item defaults, and `selectOnMove` behavior while preserving legacy Select + Combobox composition.
- Limits movement-sensitive item subscriptions to select mode, preserving ordinary Combobox movement performance.
- Migrates the public Ariakit Select examples to Combobox APIs while preserving their existing focus, virtual-focus, form, multiple-value, and rendering behavior.
- Consolidates historical and public legacy Select coverage in `app/src/sandbox/select-legacy`. The stable `select-perf` fixture remains separate so the legacy benchmark keeps its established path and test identity.

Fixes #6796.
Fixes #3451.

## Preview

Captured from [`979c415`](https://github.com/ariakit/ariakit/commit/979c4151e78035a8bc39211124a944fbf536c644) on the [exact commit preview](https://34902839-ariakit-preview.ariakit.workers.dev/react/previews/combobox-select-offscreen-value/). Opening the filterable select presents the distant selected Watermelon inside the overflowing list; the short recording also exercises the next keyboard move while the filter input retains focus.

Distant selected Watermelon presented inside the overflowing filterable select

[...]

PR #6856, opened 2026-07-25
Fix composite items being tabbable while a virtual focus widget mounts

diegohaz:
## Motivation

`useCompositeItem` decided tabbability from the rendered items alone:

```ts
isTabbable(state) {
  if (!state?.renderedItems.length) return true;
  if (state.virtualFocus) return false;
  ...
```

Rendered items are only published on the next animation frame, so the first commit of a composite widget looks exactly like a server render, and every item renders as a native tab stop. That is correct before hydration, but wrong for a widget that just mounted with [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) enabled, where items must never be tab stops.

The visible consequence is in dialogs. The auto-focus effect resolves `initialFocus` → `[data-autofocus]` → `getFirstTabbableIn(contentElement, true)`. While the items are transiently tabbable it lands on an item instead of the composite element. Focusing an item calls `setActiveId`, and virtual focus then redirects DOM focus back to the base element, so the UI looks correct while an item is already the active one. The user's first arrow key press moves to the *second* option, and screen readers announce an option nobody chose:

```tsx
// The options mount when the dialog opens, after the listbox element.
<Ariakit.Dialog render={<Ariakit.Composite store={composite} role="listbox" />}>
  {open && folders.map((folder) => <Ariakit.CompositeItem key={folder} />)}
</Ariakit.Dialog>
```

## Solution

Gate that branch on the composite element as well, and keep the rendered-items branch below `virtualFocus`:

```ts
isTabbable(state) {
  if (!state) return true;
  if (!state.baseElement && !state.renderedItems.length) return true;
  if (state.virtualFocus) return false;
  if (!state.renderedItems.length) return true;
  ...
```

`baseElement` is `null` during server rendering and published in a layout effect once the composite element mounts, so it distinguishes *before hydration* from *a widget that just mounted*, which the rendered items cannot.

Both conditions are necessary. A composite store may never receive a composite element at all, and its items must keep using roving tabindex when that happens. Gating on `baseElement` alone regresses that case: a `SelectItem` inside a combobox-backed `SelectPopover` (where `SelectList` is not the composite element) goes from `tabindex="-1"` to its own tab stop.

### Why not simply reorder the two guards

Checking `virtualFocus` first also removes the symptom, but it regresses pre-hydration keyboard access, which is deliberate long-standing behavior introduced in `dff1ba1a9`, restored by #2565, reshaped by #2601, and covered today by `app/src/sandbox/toolbar-pre-hydration/`. Server markup measured over HTTP from the rendered preview:

| | roving item | `virtualFocus` item |
| --- | --- | --- |
| Before | `<button type="button">` | `<button type="button">` |
| Reordered guards | unchanged | **`tabindex="-1"` added** |
| This change | unchanged | unchanged |

### Tests

- `app/src/sandbox/composite-item-tabbable/` renders four composites: one whose items never get a composite element, a `virtualFocus` one, an item-seeded `virtualFocus` one, and a roving one. `test-chrome-js-disabled.ts` tabs through the pre-hydration markup the app serves, `test-browser.ts` pins the hydrated tab stops in chrome, firefox, and safari, and `test.ts` duplicates the hydrated sequence on happy-dom.
- `app/src/sandbox/composite-dialog-virtual-focus/` reproduces the user-visible symptom: opening the dialog must leave the list with no highlighted option, and the first arrow key press must highlight the first one.

Each test was red-checked against the state it protects. The dialog tests fail on `main`. The sandbox tests are guard rails against the alternatives weighed here: the pre-hydration test fails under the reordered guards and when the guard keys on `state.items` instead of `state.renderedItems`, and the hydrated tests fail when the guard is trimmed to `baseElement` alone.

The first-commit state is no longer asserted directly. Observing it required rendering the fixtures with `@testing-library/react`, since `render` from `@ariakit/test/react` awaits `nextFrame()`; its user-visible consequence is what the dialog sandbox covers.

### Known limitation

When the composite element and its items mount in the *same* commit, the items' first render still sees no composite element. This change shortens that window from a cross-frame, paint-visible one to a single pre-paint render pass, so only the effect-phase race with a dialog's auto-focus survives. That behavior is unchanged from `main`.

## Relationship to #6832

This bug surfaced while working on #6832, and that branch already carries the same fix so it isn't blocked. The library change and the changeset here are byte-identical to the copies on `add-combobox-select`, so they merge as a no-op. The `composite-item-tabbable` sandbox is no longer: this PR replaced its `react.test.ts` with browser tests, so `add-combobox-select` should take the rewritten sandbox instead of merging its copy.

## Review gate reassessment

<details>
<summary>Cycle 4: Consolidate the test surface and adopt the in-flight branch's artifacts</summary>

**Assessment**

Severity is low-to-moderate: no crash or data loss, and any arrow key corrects it, but `aria-activedescendant` points at an option the user never chose, which assistive technology announces. On `main` today the reachable shapes are narrow, because a real `SelectPopover` is itself tabbable and absorbs the dialog's focus lookup before it reaches an item. The exposure grows as the ComboboxSelect work lands.

The library cost is one subscription key, one added guard, one relocated guard, and a comment. No new state, effects, helpers, abstractions, props, or public API. It has not grown across four review cycles.

Of fourteen findings, exactly one was about the direction, and it was against the originally prescribed guard rather than a choice made here; correcting it converged on the same both-conditions form that `add-combobox-select` independently arrived at. Two findings round-tripped the guard's shape to no behavioral effect. The rest were defects in newly written test fixtures, which is the signal that the fixtures, not the fix, were the problem.

**Decision**

Keep the library change frozen. Replace the hand-written fixtures with the ones `add-combobox-select` already carries so they merge as a no-op, drop a duplicative fixture and a coupled edit to the shared `toolbar-pre-hydration` sandbox, and keep only the additive dialog sandbox that proves the user-visible symptom. Declare the previously hoisted `@testing-library/react` dependency.

Intentionally unsupported: composite stores that never receive a composite element still expose tabbable items during the pre-frame window. Requiring both conditions is what keeps roving tabindex working for them, and the same limitation is accepted by the copy on `add-combobox-select`.

Superseded in part by the review response in `e931a4a2a`: the sandbox now runs as browser tests instead of vitest renders, and the `@testing-library/react` declaration was dropped with the file that needed it.

</details>


app/src/sandbox/composite-item-tabbable/react.test.ts
```diff
@@ -0,0 +1,47 @@
+import { render } from "@testing-library/react";
```
diegohaz: Use render from @ariakit/test/react instead.

But we should probably just run browser tests here (with js disabled?)

app/src/sandbox/composite-item-tabbable/test.ts
```diff
@@ -0,0 +1,26 @@
+import { q } from "@ariakit/test";
```
diegohaz: Where's the authoritative tests in `test-browser.ts`?

app/package.json
```diff
@@ -113,6 +113,7 @@
     "@ariakit/test": "workspace:*",
     "@axe-core/playwright": "4.12.1",
     "@playwright/test": "1.60.0",
+    "@testing-library/react": "16.3.2",
```
diegohaz: We shouldn't need this.

app/src/sandbox/composite-item-tabbable/react.test.ts
diegohaz: Fixed in [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a): went with the browser tests. `render` from `@ariakit/test/react` awaits `nextFrame()` before resolving, so it can never hold the first-commit window that test asserted, and `renderToString` was only an approximation of the pre-hydration markup. `test-chrome-js-disabled.ts` now tabs through the markup the app actually serves, which also proves the composite elements are not tab stops yet, something the old string assertion could not. Red-checked: reordering the guards so `virtualFocus` comes first fails it at `Inbox`, and keying the guard on `state.items` instead of `state.renderedItems` fails it at `Drafts`.

The first-commit state is no longer asserted directly. Its user-visible consequence stays covered by `composite-dialog-virtual-focus`, which is the coverage that goes red against `main`.

app/src/sandbox/composite-item-tabbable/test.ts
diegohaz: Fixed in [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a): `test-browser.ts` is now the authoritative test and runs in chrome, firefox, and safari, with `test.ts` duplicating the same tab sequence on happy-dom and `test-chrome-js-disabled.ts` covering the pre-hydration one.

Both hydrated tests walk the page with `Tab` instead of reading `tabindex` attributes: `Starred` → the `Virtual focus` listbox → the `Seeded` listbox → `Spam`. Each stop proves the items in between are out of the tab order. The fixtures got a `listbox` role and unique option names so the browser suite can query them, and the seeded listbox moved before the roving one so a tab stop follows it. Red-checked: trimming the guard to `baseElement` alone fails both files at the `Virtual focus` step.

app/package.json
diegohaz: Fixed in [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a): the dependency existed only for the deleted `react.test.ts`, which rendered the fixtures directly to reach the first-commit state. Nothing under `app/` imports `@testing-library/react` anymore, so both `app/package.json` and `pnpm-lock.yaml` are back to their `main` state and the PR no longer touches either file.

diegohaz:
All three unresolved review comments were part of the same change, so they share one commit: [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a).

**Fixed**

- [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a) ([reply](https://github.com/ariakit/ariakit/pull/6856#discussion_r3650414425)): dropped `react.test.ts` for browser tests. `render` from `@ariakit/test/react` awaits `nextFrame()`, so it cannot hold the first-commit window that file asserted; `test-chrome-js-disabled.ts` now tabs through the pre-hydration markup the app actually serves.
- [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a) ([reply](https://github.com/ariakit/ariakit/pull/6856#discussion_r3650414451)): added `test-browser.ts` as the authoritative test (chrome, firefox, safari), with `test.ts` duplicating the same tab sequence on happy-dom. The fixtures got a `listbox` role and unique option names so both suites can query them, and the seeded listbox moved before the roving one so a tab stop follows it.
- [`e931a4a`](https://github.com/ariakit/ariakit/commit/e931a4a2a0c9834592cd69b617fbf50295e98e1a) ([reply](https://github.com/ariakit/ariakit/pull/6856#discussion_r3650414478)): removed `@testing-library/react` from `app/package.json`. It existed only for the deleted file, so `app/package.json` and `pnpm-lock.yaml` are back to their `main` state and this PR no longer touches either.

Red checks against the current fixtures: reordered guards fail the pre-hydration test at `Inbox`, a guard keyed on `state.items` fails it at `Drafts`, and trimming the guard to `baseElement` alone fails both hydrated tests at the `Virtual focus` step. The library change is untouched, and `composite-dialog-virtual-focus` remains the coverage that goes red against `main`.

The description was updated: the `composite-item-tabbable` sandbox is no longer byte-identical to the copy on `add-combobox-select`, so that branch should take the rewritten sandbox instead of merging its copy.

PR #6860, opened 2026-07-25
Update Playwright to v1.62.0

diegohaz:
## Motivation

The app uses CSS style container queries backed by custom properties, for example:

```css
@container style(--ak-list-ul: 1) {
  /* list presentation */
}
```

Playwright 1.62 bundles Firefox 153, which supports this rendering path and therefore updates the Firefox visual regression baselines. The current Playwright 1.60 dependency bundles Firefox 150.0.2.

## Solution

Update the repository's Playwright dependency group from 1.60.0 to 1.62.0 and regenerate the pnpm lockfile with the configured dedupe step.

Playwright 1.62 also changed the `Locator.evaluate` overloads. Preserve the explicit `FrameCoverMetrics` contract by annotating the callback return type:

```ts
return cover.evaluate((element): FrameCoverMetrics => {
```

Chromium 151 exposed two timing boundaries in the live Stripe checkout: fine-grained card keyboard input and parent-frame motion that continued after Playwright's inner-frame actionability checks. The Plus helper now enters the card number with keyboard events, waits for Stripe's ready state, hovers with a real pointer, and waits for the button's main-viewport bounding box to settle. All purchase submissions still use real pointer clicks:

```ts
await fq.button(/^Pay$/).click();
```

CI exercised all 172 Firefox baselines: 168 changed pixels and were committed, while four solid-separator baselines remained byte-identical. Every changed screenshot is a `*-firefox.png`; there are no Chrome or Safari snapshot changes.

## Dependencies

| Package | Workspaces | Old | New | Source |
| --- | --- | ---: | ---: | --- |
| `@playwright/test` | root, app, examples, `@ariakit/scripts` | 1.60.0 | 1.62.0 | [Playwright 1.62 release notes](https://playwright.dev/docs/release-notes#version-162) |
| `playwright` | app | 1.60.0 | 1.62.0 | [npm package](https://www.npmjs.com/package/playwright/v/1.62.0) |
| `playwright-core` | lockfile | 1.60.0 | 1.62.0 | [npm package](https://www.npmjs.com/package/playwright-core/v/1.62.0) |

The existing optional peer range in `@ariakit/test` remains `^1.27.0` because 1.62.0 already satisfies it.

## Preview

Firefox disclosure component baseline after Playwright 1.62

## Testing

- `pnpm install --lockfile-only`
- `pnpm dedupe --lockfile-only`
- `git diff --check`
- Final CI: all 48 PR checks passed
- Repository-wide `Type check`: passed
- Chrome, Firefox, and Safari regular and visual jobs: passed; every final screenshot-update step skipped
- Initial visual CI: Firefox regenerated exactly 168 expected baselines; Chrome and Safari generated none
- Plus: passed with real pointer clicks (8 tests passed; one transient Stripe readiness attempt passed through the configured retry policy)
- Browser and visual regression verification was performed on CI, not locally

Automerge is disabled because the update required non-mechanical type and browser-test compatibility fixes.


diegohaz:
Playwright 1.62 collapsed the `Locator.evaluate` overloads so the existing explicit call-site generic no longer type-checked (`TS2558: Expected 2-3 type arguments, but got 1`). I migrated it to an explicit callback return annotation:

```ts
return cover.evaluate((element): FrameCoverMetrics => {
```

This preserves the same return contract without inventing an `Arg` type argument. Because this is a non-mechanical compatibility change, automerge is disabled and human review is required. The repository-wide type check and browser workflows are rerunning on this commit; verification remains CI-only.

PR #6865, opened 2026-07-26
Add selective CI orchestration

diegohaz:
## Motivation

Pull requests currently run every specialized CI workflow, including performance and browser-heavy lanes, even when a change cannot affect them. Dependency, configuration, and unknown changes still need a conservative fallback because their impact cannot be determined reliably from filenames alone.

The final required-check contract also needs to remain stable when specialized workflows are skipped. Main should keep its duplicate legacy browser coverage until the legacy website migration is complete, while branch protection requires one aggregate result instead of leaf checks that may not exist.

## Solution

- Add a single `CI` pull request orchestrator backed by tested planning and gating functions and a dependency-free runner in `ariakit-scripts`.
- Always run Main, including its current legacy browser jobs, while conditionally calling reusable App, Perf, Plus, release preview, docs, styles, and OG image workflows.
- Treat dependency, configuration, workflow, script, and unknown non-documentation paths as full CI. Labels can only broaden the plan.
- Run Plan and Gate through a dependency-free `ariakit-scripts` entry point with the repository's pinned Node version, so neither job installs workspace dependencies on the serialized critical path.
- Preserve direct push workflows, and namespace called-workflow concurrency groups and artifacts to avoid caller collisions.
- Make `Gate` the sole required-check contract. The Main ruleset must require `Gate` instead of the 14 legacy contexts before this PR can merge.
- Preserve visual screenshot updates when GitHub prefixes nested reusable-workflow job names.

### Automatic workflow selection

The Plan job classifies every path from the merge-base diff. Renames and copies classify both the old and new path. Main starts enabled and cannot be skipped; the remaining workflows are enabled when any changed path matches their inputs.

| Workflow | Automatically triggered by |
| --- | --- |
| Main | Every pull request, including the current legacy browser tests. |
| App | Any file under `app/` or `nextjs/`, plus runtime package code. This includes app browser, visual, and preview jobs. |
| Perf | Non-test runtime code under `app/`, `nextjs/`, or `packages/`. Individual performance scenarios are not guessed; when Perf is selected, its complete dynamic matrix runs. |
| Plus | Files under `website/` or `guide/`, plus runtime package code consumed by the website. |
| Release preview | Files under `.changeset/` or `templates/`, plus publishable package runtime code. This workflow is enabled only for pull requests targeting `main`. |
| Docs | Runtime package code and package `readme.md` files. |
| Build styles | Files under `app/src/styles/` and the `build-styles`, `styles`, `styles-json-types`, or `styles-shared` helpers under `app/src/lib/`. |
| Generate OG images | Files under `app/public/og-image/`, non-test files under `app/src/`, and runtime package code imported by the renderer. |

Here, runtime package code means a non-Markdown, non-test file under `packages/`. Repository infrastructure has higher precedence and selects full CI.

### Fail-closed rules

All applicable workflows run when the change could have repository-wide effects or the planner does not recognize a non-documentation path. Release preview remains limited to pull requests targeting `main`. The fail-closed inputs include:

- `.github/`, `packages/ariakit-scripts/`, `patches/`, and `scripts/`;
- dependency manifests, lockfiles, workspace settings, `.npmrc`, Wrangler files, and `package.json` files at any depth;
- `config.*`, `*.config.*`, and `tsconfig*.json` files at any depth;
- root-level non-Markdown files;
- any unmatched non-Markdown path.

Only explicit low-risk categories can remain Main-only, such as unmatched Markdown, examples, and package test or screenshot files. Known specialized roots still select their owning workflow, so an app test selects App and a package README selects Docs.

Labels only add work and never suppress an inferred workflow. `ci:full` enables every applicable workflow. The workflow-specific labels are `ci:main`, `ci:app`, `ci:perf`, `ci:plus`, `ci:release-preview`, `ci:docs`, `ci:build-styles`, and `ci:og-images`. Adding or removing any label reruns the planner. These `ci:*` labels are supported planner inputs but are not currently configured in the repository.

Title/body edits and base retargets do not start CI in v1. After retargeting, trigger a fresh supported event by pushing a new head commit, changing an existing label, or closing and reopening the pull request.

Examples:

```text
readme.md                                      -> Main
packages/ariakit-components/src/dialog.test.ts -> Main
app/src/lib/example.test.ts                    -> Main + App
packages/ariakit-components/src/dialog.ts      -> Main + App + Perf + Plus + Release preview + Docs + Generate OG images
app/src/pages/example.ts                       -> Main + App + Perf + Generate OG images
pnpm-lock.yaml                                 -> All workflows
tooling/new-file.ts                            -> All workflows (unknown path fallback)
```

The Plan logs retain the complete changed-file list and per-workflow reasons. Gate receives only the plan version and workflow booleans, then fails unless planning succeeded, every selected workflow succeeded, and every unselected workflow was skipped.

<details>
<summary>Cycle 3: Narrow edited-event support</summary>

**Assessment**

Three finding-bearing implementation cycles exposed a large Gate payload, destructive metadata-edit cancellation and duplicated setup cost, then a misleading all-skipped metadata check suite. The core orchestrator remains justified by the original scope, but metadata edits are frequent during ordinary PR maintenance and GitHub's commit rollup made the head appear green while the real suite was still running.

Keeping automatic base-retarget handling would require another event-routing workflow or a reusable wrapper that republishes all 14 required aliases. That would add maintenance complexity to an already substantial workflow and still create an edit-event check suite. The metadata behavior is also platform-sensitive and has no practical repository regression harness.

**Decision**

Keep the conservative selective-CI orchestrator, but remove the `edited` activity type and all metadata-specific routing. Title/body edits now create no CI run. Base retargets intentionally require a subsequent supported PR event in v1. A separate base-change wrapper and an explicit metadata-status job are intentionally unsupported because they add complexity without removing the misleading check-suite behavior.

</details>

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the conservative orchestrator design</summary>

**Assessment**

Three finding-bearing review cycles exposed boundary conditions around nested configuration paths, reusable-workflow job prefixes, pull request retargeting, and unknown test-like paths. These issues had medium-to-high potential impact because a missed lane could produce a false-green gate, but their expected frequency ranges from uncommon to occasional.

The implementation is substantial but bounded: one planner/gate policy, one orchestrator, reusable workflow conversions, and focused tests. Its main maintenance cost is the explicit repository path policy and GitHub reusable-workflow semantics. Each review correction remained localized and did not add another state machine, policy layer, or public contract.

The supported scope stays deliberately conservative: Main always runs; known specialized inputs select their workflows; dependency, configuration, and unrecognized non-documentation paths fail closed to full CI; overrides only add work.

**Decision**

Continue the orchestrator direction and address every finding. Simplifying by weakening the fallback or omitting base-retarget handling would violate the reliability contract that motivated this change. No identified edge case is intentionally unsupported. Use `Gate` as the sole required-check contract; compatibility aliases are intentionally omitted.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,21 +2,18 @@ name: CI
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
 
 permissions:
   contents: read
 
 concurrency:
-  # Keep metadata-only edits from replacing pending real CI work.
-  group: ci-${{ github.ref }}-${{ github.event.action == 'edited' && github.event.changes.base == null && 'metadata' || 'work' }}
+  group: ci-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   plan:
     name: Plan
-    # Base edits change the diff; title and body edits do not.
-    if: github.event.action != 'edited' || github.event.changes.base != null
     runs-on: ubuntu-latest
     outputs:
       main: ${{ steps.plan.outputs.main }}
@@ -126,9 +123,8 @@ jobs:
     secrets: inherit
 
   gate:
-    # Metadata-only runs must not shadow the real Gate result for this commit.
-    name: ${{ github.event.action == 'edited' && github.event.changes.base == null && 'Metadata / Gate' || 'Gate' }}
-    if: always() && (github.event.action != 'edited' || github.event.changes.base != null)
+    name: Gate
+    if: always()
     needs:
       - plan
       - main
@@ -163,9 +159,8 @@ jobs:
     # Preserve the current required status names until the main ruleset is
     # migrated to require Gate. Reusable jobs use prefixed check names, so
     # removing these aliases before that migration would block automerge.
-    # Metadata-only runs use distinct names to preserve the real CI results.
-    name: ${{ github.event.action == 'edited' && github.event.changes.base == null && format('Metadata / {0}', matrix.name) || matrix.name }}
-    if: always() && (github.event.action != 'edited' || github.event.changes.base != null)
+    name: ${{ matrix.name }}
+    if: always()
     needs: gate
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,15 @@ permissions:
   contents: read
 
 concurrency:
-  group: ci-${{ github.ref }}
+  # Keep metadata-only edits from replacing pending real CI work.
+  group: ci-${{ github.ref }}-${{ github.event.action == 'edited' && github.event.changes.base == null && 'metadata' || 'work' }}
   cancel-in-progress: true
 
 jobs:
   plan:
     name: Plan
+    # Base edits change the diff; title and body edits do not.
+    if: github.event.action != 'edited' || github.event.changes.base != null
     runs-on: ubuntu-latest
     outputs:
       main: ${{ steps.plan.outputs.main }}
@@ -127,8 +130,9 @@ jobs:
     secrets: inherit
 
   gate:
-    name: Gate
-    if: always()
+    # Metadata-only runs must not shadow the real Gate result for this commit.
+    name: ${{ github.event.action == 'edited' && github.event.changes.base == null && 'Metadata / Gate' || 'Gate' }}
+    if: always() && (github.event.action != 'edited' || github.event.changes.base != null)
     needs:
       - plan
       - main
@@ -164,8 +168,9 @@ jobs:
     # Preserve the current required status names until the main ruleset is
     # migrated to require Gate. Reusable jobs use prefixed check names, so
     # removing these aliases before that migration would block automerge.
-    name: ${{ matrix.name }}
-    if: always()
+    # Metadata-only runs use distinct names to preserve the real CI results.
+    name: ${{ github.event.action == 'edited' && github.event.changes.base == null && format('Metadata / {0}', matrix.name) || matrix.name }}
+    if: always() && (github.event.action != 'edited' || github.event.changes.base != null)
     needs: gate
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,197 @@
+name: CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  plan:
+    name: Plan
+    runs-on: ubuntu-latest
+    outputs:
+      main: ${{ steps.plan.outputs.main }}
+      app: ${{ steps.plan.outputs.app }}
+      perf: ${{ steps.plan.outputs.perf }}
+      plus: ${{ steps.plan.outputs.plus }}
+      release_preview: ${{ steps.plan.outputs.release_preview }}
+      docs: ${{ steps.plan.outputs.docs }}
+      build_styles: ${{ steps.plan.outputs.build_styles }}
+      og_images: ${{ steps.plan.outputs.og_images }}
+      plan: ${{ steps.plan.outputs.plan }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up environment
+        uses: ./.github/workflows/setup
+
+      - name: Plan CI
+        id: plan
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+          CI_LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+        run: >-
+          pnpm exec ariakit ci-plan
+          --base "$BASE_SHA"
+          --head "$HEAD_SHA"
+          --base-ref "$BASE_REF"
+          --labels "$CI_LABELS"
+          --output "$GITHUB_OUTPUT"
+
+  main:
+    name: Main
+    needs: plan
+    if: needs.plan.outputs.main == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/main.yml
+    secrets: inherit
+
+  app:
+    name: App
+    needs: plan
+    if: needs.plan.outputs.app == 'true'
+    permissions:
+      actions: read
+      contents: read
+      deployments: write
+      statuses: write
+    uses: ./.github/workflows/app.yml
+    secrets: inherit
+
+  perf:
+    name: Performance
+    needs: plan
+    if: needs.plan.outputs.perf == 'true'
+    permissions:
+      actions: read
+      contents: read
+    uses: ./.github/workflows/perf.yml
+    secrets: inherit
+
+  plus:
+    name: Plus
+    needs: plan
+    if: needs.plan.outputs.plus == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/plus.yml
+    secrets: inherit
+
+  release_preview:
+    name: Release preview
+    needs: plan
+    if: needs.plan.outputs.release_preview == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/release-preview.yml
+    secrets: inherit
+
+  docs:
+    name: Docs
+    needs: plan
+    if: needs.plan.outputs.docs == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/docs.yml
+    secrets: inherit
+
+  build_styles:
+    name: Build styles
+    needs: plan
+    if: needs.plan.outputs.build_styles == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/build-styles.yml
+    secrets: inherit
+
+  og_images:
+    name: OG images
+    needs: plan
+    if: needs.plan.outputs.og_images == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/og-images.yml
+    secrets: inherit
+
+  gate:
+    name: Gate
+    if: always()
+    needs:
+      - plan
+      - main
+      - app
+      - perf
+      - plus
+      - release_preview
+      - docs
+      - build_styles
+      - og_images
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Set up environment
+        uses: ./.github/workflows/setup
+
+      - name: Verify CI results
+        env:
+          CI_PLAN: ${{ needs.plan.outputs.plan }}
+          CI_RESULTS: ${{ toJSON(needs) }}
+        run: >-
+          pnpm exec ariakit ci-gate
+          --plan "$CI_PLAN"
+          --results "$CI_RESULTS"
+
+  required:
+    # Preserve the current required status names until the main ruleset is
+    # migrated to require Gate. Reusable jobs use prefixed check names, so
+    # removing these aliases before that migration would block automerge.
+    name: ${{ matrix.name }}
+    if: always()
+    needs: gate
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        name:
+          - Lint
+          - Type check
+          - Build packages
+          - Test
+          - Test React
+          - Test Solid
+          - Test Android
+          - Test Chrome
+          - Test Firefox
+          - Test macOS
+          - Test Safari
+          - Visual / Finalize
+          - Preview / Deploy
+          - Generate OG images
+    steps:
+      - name: Verify gate
+        env:
+          GATE_RESULT: ${{ needs.gate.result }}
+        run: |
+          if [ "$GATE_RESULT" != "success" ]; then
+            echo "CI gate result was $GATE_RESULT" >&2
+            exit 1
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,39 +154,3 @@ jobs:
           CI_PLAN: ${{ needs.plan.outputs.plan }}
           CI_RESULTS: ${{ toJSON(needs) }}
         run: node packages/ariakit-scripts/src/ci-run.ts gate
-
-  required:
-    # Preserve the current required status names until the main ruleset is
-    # migrated to require Gate. Reusable jobs use prefixed check names, so
-    # removing these aliases before that migration would block automerge.
-    name: ${{ matrix.name }}
-    if: always()
-    needs: gate
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        name:
-          - Lint
-          - Type check
-          - Build packages
-          - Test
-          - Test React
-          - Test Solid
-          - Test Android
-          - Test Chrome
-          - Test Firefox
-          - Test macOS
-          - Test Safari
-          - Visual / Finalize
-          - Preview / Deploy
-          - Generate OG images
-    steps:
-      - name: Verify gate
-        env:
-          GATE_RESULT: ${{ needs.gate.result }}
-        run: |
-          if [ "$GATE_RESULT" != "success" ]; then
-            echo "CI gate result was $GATE_RESULT" >&2
-            exit 1
-          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,10 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Set up environment
-        uses: ./.github/workflows/setup
+      - name: Set up Node
+        uses: actions/setup-node@v7
+        with:
+          node-version-file: package.json
 
       - name: Plan CI
         id: plan
@@ -45,13 +47,7 @@ jobs:
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
           BASE_REF: ${{ github.event.pull_request.base.ref }}
           CI_LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
-        run: >-
-          pnpm exec ariakit ci-plan
-          --base "$BASE_SHA"
-          --head "$HEAD_SHA"
-          --base-ref "$BASE_REF"
-          --labels "$CI_LABELS"
-          --output "$GITHUB_OUTPUT"
+        run: node packages/ariakit-scripts/src/ci-run.ts plan
 
   main:
     name: Main
@@ -152,17 +148,16 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Set up environment
-        uses: ./.github/workflows/setup
+      - name: Set up Node
+        uses: actions/setup-node@v7
+        with:
+          node-version-file: package.json
 
       - name: Verify CI results
         env:
           CI_PLAN: ${{ needs.plan.outputs.plan }}
           CI_RESULTS: ${{ toJSON(needs) }}
-        run: >-
-          pnpm exec ariakit ci-gate
-          --plan "$CI_PLAN"
-          --results "$CI_RESULTS"
+        run: node packages/ariakit-scripts/src/ci-run.ts gate
 
   required:
     # Preserve the current required status names until the main ruleset is
--- a/packages/ariakit-scripts/src/ci.ts
+++ b/packages/ariakit-scripts/src/ci.ts
@@ -1,0 +1,424 @@
+import { execFileSync } from "node:child_process";
+import { appendFileSync } from "node:fs";
+
+export const ciWorkflowNames = [
+  "main",
+  "app",
+  "perf",
+  "plus",
+  "release_preview",
+  "docs",
+  "build_styles",
+  "og_images",
+] as const;
+
+export type CIWorkflowName = (typeof ciWorkflowNames)[number];
+
+export interface CIPlan {
+  version: 1;
+  baseRef: string;
+  files: string[];
+  labels: string[];
+  workflows: Record<CIWorkflowName, boolean>;
+  reasons: Record<CIWorkflowName, string[]>;
+}
+
+export interface CreateCIPlanOptions {
+  baseRef?: string;
+  labels?: string[];
+}
+
+export interface RunCIPlanOptions {
+  base: string;
+  head: string;
+  baseRef: string;
+  labels: string;
+  output: string;
+}
+
+export interface CIGateResult {
+  result?: string;
+}
+
+export interface RunCIGateOptions {
+  plan: string;
+  results: string;
+}
+
+type CIResults = Record<string, CIGateResult | undefined>;
+
+interface CIGatePlan {
+  version: 1;
+  workflows: Record<CIWorkflowName, boolean>;
+}
+
+const workflowLabels: Record<string, CIWorkflowName> = {
+  "ci:app": "app",
+  "ci:build-styles": "build_styles",
+  "ci:docs": "docs",
+  "ci:main": "main",
+  "ci:og-images": "og_images",
+  "ci:perf": "perf",
+  "ci:plus": "plus",
+  "ci:release-preview": "release_preview",
+};
+
+const dependencyAndConfigNames = new Set([
+  ".npmrc",
+  "bun.lock",
+  "bun.lockb",
+  "package-lock.json",
+  "package.json",
+  "pnpm-lock.yaml",
+  "pnpm-workspace.yaml",
+  "wrangler.json",
+  "wrangler.jsonc",
+  "wrangler.toml",
+  "yarn.lock",
+]);
+
+function normalizeCIPath(file: string) {
+  return file.replaceAll("\\", "/").replace(/^\.\/+/, "");
+}
+
+function isMarkdown(file: string) {
+  return /\.(?:md|mdx)$/i.test(file);
+}
+
+function isTestFile(file: string) {
+  return (
+    /(?:^|\/)__tests__\//.test(file) ||
+    /(?:^|\/)__screenshots__\//.test(file) ||
+    /\.(?:test|spec)\.[cm]?[jt]sx?$/.test(file)
+  );
+}
+
+function isFullCIPath(file: string) {
+  if (
+    file.startsWith(".github/") ||
+    file.startsWith("packages/ariakit-scripts/") ||
+    file.startsWith("patches/") ||
+    file.startsWith("scripts/")
+  ) {
+    return true;
+  }
+  const name = file.slice(file.lastIndexOf("/") + 1);
+  if (dependencyAndConfigNames.has(name)) return true;
+  if (/(?:^|\.)config(?:\.|$)/.test(name)) return true;
+  if (/^tsconfig(?:\.[^/]+)?\.json$/.test(name)) return true;
+  if (!file.includes("/") && !isMarkdown(file)) return true;
+  return false;
+}
+
+function isPackageRuntimePath(file: string) {
+  if (!file.startsWith("packages/")) return false;
+  if (isMarkdown(file) || isTestFile(file)) return false;
+  return true;
+}
+
+function isAppRuntimePath(file: string) {
+  if (!/^(?:app|nextjs)\//.test(file)) return false;
+  if (isTestFile(file)) return false;
+  return true;
+}
+
+function addReason(plan: CIPlan, workflow: CIWorkflowName, reason: string) {
+  plan.workflows[workflow] = true;
+  const reasons = plan.reasons[workflow];
+  if (!reasons.includes(reason)) {
+    reasons.push(reason);
+  }
+}
+
+function addFullCIReason(plan: CIPlan, file: string) {
+  for (const workflow of ciWorkflowNames) {
+    addReason(plan, workflow, `Repository-wide CI input: ${file}`);
+  }
+}
+
+function addFileReasons(plan: CIPlan, file: string) {
+  if (isFullCIPath(file)) {
+    addFullCIReason(plan, file);
+    return;
+  }
+
+  let matched = false;
+  if (/^(?:app|nextjs)\//.test(file) || isPackageRuntimePath(file)) {
+    addReason(plan, "app", `App input: ${file}`);
+    matched = true;
+  }
+
+  if (isAppRuntimePath(file) || isPackageRuntimePath(file)) {
+    addReason(plan, "perf", `Performance-sensitive runtime: ${file}`);
+    matched = true;
+  }
+
+  if (isPackageRuntimePath(file)) {
+    addReason(plan, "release_preview", `Publishable package input: ${file}`);
+    matched = true;
+  }
+
+  if (
+    isPackageRuntimePath(file) ||
+    /^packages\/[^/]+\/readme\.md$/i.test(file)
+  ) {
+    addReason(plan, "docs", `API documentation input: ${file}`);
+    matched = true;
+  }
+
+  if (/^(?:website|guide)\//.test(file) || isPackageRuntimePath(file)) {
+    addReason(plan, "plus", `Plus website input: ${file}`);
+    matched = true;
+  }
+
+  if (
+    file.startsWith("app/src/styles/") ||
+    /^app\/src\/lib\/(?:build-styles|styles(?:-json-types|-shared)?)\.ts$/.test(
+      file,
+    )
+  ) {
+    addReason(plan, "build_styles", `Generated styles input: ${file}`);
+    matched = true;
+  }
+
+  if (
+    file.startsWith("app/public/og-image/") ||
+    (file.startsWith("app/src/") && !isTestFile(file))
+  ) {
+    addReason(plan, "og_images", `OG image input: ${file}`);
+    matched = true;
+  }
+
+  if (file.startsWith(".changeset/") || file.startsWith("templates/")) {
+    addReason(plan, "release_preview", `Release preview input: ${file}`);
+    matched = true;
+  }
+
+  const safelyCoveredByMain =
+    isMarkdown(file) ||
+    file.startsWith("examples/") ||
+    (file.startsWith("packages/") && isTestFile(file));
+  if (!matched && !safelyCoveredByMain) {
+    addFullCIReason(plan, file);
+  }
+}
+
+export function createCIPlan(
+  changedFiles: string[],
+  options: CreateCIPlanOptions = {},
+) {
+  const files = [
+    ...new Set(changedFiles.map(normalizeCIPath).filter(Boolean)),
+  ].sort();
+  const labels = [...new Set(options.labels ?? [])].sort();
+  const workflows: Record<CIWorkflowName, boolean> = {
+    main: false,
+    app: false,
+    perf: false,
+    plus: false,
+    release_preview: false,
+    docs: false,
+    build_styles: false,
+    og_images: false,
+  };
+  const reasons: Record<CIWorkflowName, string[]> = {
+    main: [],
+    app: [],
+    perf: [],
+    plus: [],
+    release_preview: [],
+    docs: [],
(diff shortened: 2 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,197 @@
...
+        uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Set up environment
+        uses: ./.github/workflows/setup
+
+      - name: Verify CI results
+        env:
+          CI_PLAN: ${{ needs.plan.outputs.plan }}
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`034d0bb4b`](https://github.com/ariakit/ariakit/commit/034d0bb4b4aec235c2934b941599a370120ef599): Gate now receives only the plan version and workflow booleans, while the complete file and reason diagnostics remain in Plan logs. The same commit adds a proven 200-file regression that failed at 126,224 bytes before the fix and passes with a 148-byte payload.

diegohaz:
**Fixed**

- [`034d0bb4b`](https://github.com/ariakit/ariakit/commit/034d0bb4b4aec235c2934b941599a370120ef599) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3652046634)): Compact the Gate payload to the plan version and workflow booleans while retaining full diagnostics in Plan logs. Added a proven 200-file size regression.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,197 @@
...
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
```
diegohaz, line 12: Adding `edited` to the trigger list while keeping `cancel-in-progress: true` means any pull request title or body edit destroys the CI run that is already testing the current head commit and restarts the whole pipeline.

This already happened twice on this PR. `repos/ariakit/ariakit/actions/workflows/ci.yml/runs` shows three runs for head SHA `034d0bb4`:

```text
30193723390  034d0bb4  pull_request  completed  success    2026-07-26T07:57:25Z
30193640936  034d0bb4  pull_request  completed  cancelled  2026-07-26T07:54:42Z
30193600983  034d0bb4  pull_request  completed  cancelled  2026-07-26T07:53:18Z
```

The last push was at `2026-07-26T07:52:44Z` and the issue timeline for this PR contains no `labeled`, `unlabeled`, `renamed`, or `head_ref_force_pushed` events, so the two extra runs can only come from `edited`. The first was killed after ~84s and the second after ~163s.

Real-world impact: a maintainer or bot editing a PR description — adding a follow-up issue link, a screenshot, a checklist, a review-gate reassessment section — throws away every in-flight lane for a commit that is already being tested, including the Perf matrix and the macOS/Safari browser jobs. On a full run that is up to ~13 minutes of work across ~40 jobs. It also flips already-green required checks to `cancelled` until the replacement run finishes, so editing a description right before merge un-greens the PR. Before this change, `main.yml`, `app.yml`, and friends used the default `pull_request` types, so a description edit did not re-trigger anything.

Dropping `edited` is not the right fix on its own: retargeting a PR's base branch changes `base.sha`, so the changed-file set and the `main`-only release-preview rule would stay computed against the old base until the next push, which can under-select lanes.

The narrow fix is to keep re-planning on base changes but never cancel for metadata-only edits:

```suggestion
  pull_request:
    types: [opened, synchronize, reopened, edited, labeled, unlabeled]

permissions:
  contents: read

concurrency:
  group: ci-${{ github.ref }}
  # A title or body edit re-plans, but it must never cancel a run that is
  # already testing this head commit. Only a base-branch change invalidates
  # the plan.
  cancel-in-progress: ${{ github.event.action != 'edited' || github.event.changes.base != null }}
```

That stops the destructive part, but a metadata-only edit still queues a full duplicate run behind the first one. To avoid that too, guard the work itself on the same condition: put `if: github.event.action != 'edited' || github.event.changes.base` on `plan` and the eight workflow jobs, and move the same condition onto Gate's `Verify CI results` step rather than the `gate` job. Keeping `gate` and `required` unconditional means every required check still reports success on a metadata-only edit, so this avoids depending on how the ruleset treats skipped required checks.

**Assessment**

- Severity: Medium (verified), because in-flight runs are discarded and green required checks momentarily turn red, but no incorrect result is reported.
- Priority: P1 (estimate, 80% confidence), because the regression is introduced here and fires on ordinary PR maintenance.
- Breakage likelihood: Likely (verified), because it already occurred twice on this PR from two description edits.
- Fix effort: Easy (estimate, 85% confidence), because it is a concurrency expression plus job-level conditions in this one file.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,197 @@
...
+      plan: ${{ steps.plan.outputs.plan }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up environment
+        uses: ./.github/workflows/setup
```
diegohaz: `Plan` and `Gate` both run the full `setup` composite action, which does `pnpm install --frozen-lockfile`, even though neither command needs a single dependency. Because every lane now has `needs: plan` and the required-check aliases have `needs: gate`, that install sits on the serialized critical path of every pull request twice.

Step timings from the green run on this head (`repos/ariakit/ariakit/actions/runs/30193723390/jobs`):

```text
Plan   started 07:58:31  Checkout 17s  Set up environment 35s  Plan CI 1s   (job 59s)
Gate   started 08:10:12  Checkout  2s  Set up environment 32s  Verify   1s  (job 39s)
```

The run started at `07:57:25` and the first real job (`Main / Lint`) started at `07:59:32`, so ~2m07s elapsed before any actual CI work began; Gate plus the aliases added another ~46s at the end. That is ~2m53s of fixed overhead on a 13m34s run, of which ~67s is dependency installation that produces nothing.

This matters most for exactly the pull requests this PR is meant to make cheap. A docs-only change now selects Main alone, so the fixed orchestration overhead is a large fraction of the total run instead of a rounding error.

`ci.ts` imports only `node:child_process` and `node:fs`, and `package.json` pins `devEngines.runtime` to Node `24.18.0`, which runs TypeScript directly. I confirmed locally on Node 24.18.0 that the planner loads and produces plans with no `node_modules` present:

```sh
node --input-type=module -e "
const m = await import('./packages/ariakit-scripts/src/ci.ts');
console.log(JSON.stringify(m.createCIPlan(['packages/ariakit-components/src/dialog.ts']).workflows));
"
```

So both jobs can drop `pnpm` entirely and use `actions/setup-node@v7` with `node-version-file: package.json` (the same file the existing setup action already resolves from `devEngines`), invoking a small dependency-free entry module instead of `pnpm exec ariakit`. The only thing keeping the install alive is that `src/index.ts` pulls in `commander` at import time.

Worth avoiding the other obvious shortcut: replacing the `fetch-depth: 0` checkout with the REST `compare` endpoint would remove the checkout too, but that endpoint caps at 300 files per page and 3,000 files total, and a silent truncation there would under-select lanes and produce a false-green gate.

**Assessment**

- Severity: Low (verified), because the cost is latency and runner minutes, not an incorrect gate result.
- Priority: P2 (estimate, 75% confidence), because it is introduced by this change and partly offsets its stated benefit, but it does not block correctness.
- Breakage likelihood: Likely (verified), because the overhead is incurred on every pull request run.
- Fix effort: Easy (estimate, 80% confidence), because it is a setup-step swap in two jobs plus a dependency-free CLI entry point.

diegohaz (commented):
Initial review of the whole PR: merge base `b71b94dc699e3355b421b1668a7557fb54b47c23` to head `034d0bb4b4aec235c2934b941599a370120ef599`, across `.github/workflows/**` and `packages/ariakit-scripts/src/{ci,ci-visual}*`, plus the surrounding workflows, the repository ruleset, and the runs this PR has already produced.

The orchestration design is sound and the conservative fail-closed policy is the right default. Two findings are published inline, both about the orchestrator's operational cost rather than its correctness.

## Checked

- Checked out `selective-ci-orchestrator` in a dedicated worktree and diffed against the merge base.
- `pnpm test packages/ariakit-scripts/src/ci.test.ts packages/ariakit-scripts/src/ci-visual.test.ts` — 16 tests pass. Confirmed both files are collected by the default no-loader suite: `vitest.config.ts` includes `**/*test.{ts,tsx}` and its `defaultTestExcludes` do not cover `packages/ariakit-scripts/`.
- `pnpm exec tsc -b packages/ariakit-scripts` and `oxlint` / `oxfmt --check` on the four new or changed source files — all clean.
- Exercised `createCIPlan` directly on boundary paths not covered by the suite: `components/*.md` (Main only), root `examples/**` (Main only), `logo/*.svg` (full CI fallback), `app/src/**/__screenshots__/*.png` (Main + App, no Perf), `packages/*/src/config.ts` (full CI), `guide/*.md` + `templates/*.json` (Main + Plus + Release preview).
- Traced the fail-closed regexes by hand. `/(?:^|\.)config(?:\.|$)/` correctly rejects `myconfig.ts` and `configure.ts` while matching `astro.config.mjs`, `.config`, and bare `config`; `tsconfig.json` is not matched by it, so the dedicated `tsconfig*` rule is load-bearing rather than redundant. `.github/` has no tracked Markdown and `.claude/` is untracked, so the repository-wide roots do not swallow low-risk documentation paths.
- Looked for false-green holes in the Main-only categories. Unmatched Markdown, root `examples/**`, and `packages/**` test or screenshot files are all covered by Main: `Build website` builds the legacy site that consumes root `examples/` and `components/*.md`, the legacy browser matrix exercises them, and `getFrameworkTestIncludes` in `vitest.config.ts` globs `examples/**/index.${loader}.tsx` into `Test React` and `Test Solid`. The app's own content collections load from `app/src/**` only, which the App rule already covers. I did not find a path class that can skip a lane it needs.
- Verified the artifact namespacing is complete for a single shared run. Within one `CI` run the uploads are now `website-dist`, `{project}-test-results`, `safari-test-results` (Main), `app-build`, `nextjs-dist`, `app-visual-{project}`, `app-{project}-test-results` (App), `perf-*` (Perf), and `plus-test-results` (Plus), with no remaining collisions. `perf.yml` still downloads `app-build` and `nextjs-dist` by their original names, but only cross-run from a successful `push` run of `app.yml`, which keeps its `push: [main]` trigger — that path is unaffected.
- Confirmed the required-check aliases are exactly right. The `Main` ruleset requires 14 contexts (`Lint`, `Type check`, `Build packages`, `Test`, `Test React`, `Test Solid`, `Test Android`, `Test Chrome`, `Test Firefox`, `Test macOS`, `Test Safari`, `Preview / Deploy`, `Visual / Finalize`, `Generate OG images`), and the `required` matrix lists the same 14 with no name collisions against the prefixed reusable-workflow checks. The green run on this head shows both sets side by side, e.g. `Lint` (3s) next to `Main / Lint` (57s).
- Checked reusable-workflow nesting and permissions. `ci.yml` → `app.yml` → `visual.yml` is three levels, within GitHub's limit; `secrets: inherit` propagates; and each caller job grants at least the permissions its called jobs declare, so `Preview / Deploy` keeps `deployments: write` and `statuses: write`.
- Verified the `visual.yml` refactor at runtime. `getVisualProject` is only reachable from a step that runs when a visual test fails, so I reproduced its loading mechanism outside CI: a dynamic `import()` of an absolute `$GITHUB_WORKSPACE` path from inside `new AsyncFunction`, which is how `actions/github-script` evaluates its script. It resolves and returns `chrome` for both `Test Chrome` and `App / Test Chrome`, and `undefined` for `App / Build app`. `@ariakit/scripts` is `"type": "module"`, so the `.js` module is valid ESM.
- Read this PR's own runs and job step timings via `gh api` to measure the orchestrator's fixed overhead and to establish the trigger behavior described in the first inline comment.

## Findings

Both are published inline on `.github/workflows/ci.yml`:

1. The `edited` trigger combined with `cancel-in-progress: true` destroys in-flight runs on title and body edits. Verified from run history: three runs for head `034d0bb4` with two cancelled, and no push or label events between them. Severity Medium, P1, Likely, Easy.
2. `Plan` and `Gate` each add a full `pnpm install` to the serialized critical path, ~2m53s of fixed overhead per PR of which ~67s is dependency installation the commands do not need. Severity Low, P2, Likely, Easy.

## Other candidates and their dispositions

- **Perf is selected for Markdown under `app/`.** `isAppRuntimePath` excludes only test files, while `isPackageRuntimePath` also excludes Markdown, so a prose-only edit to one of the 26 `app/src/**/*.mdx` files would select the full Perf matrix. The asymmetry is real and documented in the PR description, but no commit on `main` since 2026-04-01 touches only `app/**` Markdown, so tightening it would buy nothing measurable. Declined non-goal.
- **The 14 alias jobs each provision a runner for a one-line check.** Roughly five runner-minutes per PR. There is no cheaper way to publish 14 distinct required contexts, and the jobs are already documented as temporary until the ruleset requires `Gate`. Declined non-goal.
- **`createCIPlan` and `parseCIPlan` spell out all eight workflow keys as object literals instead of deriving them from `ciWorkflowNames`.** This looks like duplication, but the literals are what makes TypeScript flag every site when a workflow is added, and deriving them would need a `{} as Record<CIWorkflowName, T>` assertion. The current form is the safer trade-off. Declined non-goal.
- **Skipping App means a PR gets no preview deployment and no `Preview (commit)` / `Preview (branch)` commit statuses.** Neither context is in the ruleset and nothing else in the repository consumes them, and the required `Preview / Deploy` alias still reports. Working as intended.
- **A selected reusable workflow whose only job is skipped reports `skipped` and would fail the gate**, which applies to `plus.yml`'s `if: github.repository == 'ariakit/ariakit'` guard. This cannot trigger in this repository under the same-repo PR model the workflows already assume. Declined non-goal.

## Prior review threads

The one existing review thread on this PR (`.github/workflows/ci.yml`, the Codex `CI_PLAN` argument-size report) is resolved. I independently confirmed it is fixed at the current head rather than taking the reply at face value: `serializeCIGatePlan` emits only `version` and `workflows`, and the `serializes a compact gate plan for large changes` test asserts a 200-file plan stays under 1 KiB. No reply was needed. No other unresolved threads exist.

## Not fully validated

- The `cancel-in-progress` expression in the first inline comment is derived from GitHub's documented expression semantics for `github.event.changes`, not from an executed run. It should be confirmed on a real edit event before relying on it.
- I did not run the browser, visual, or Perf suites; the reasoning about their behavior under the orchestrator comes from reading the workflows and from this PR's own green run.
- I did not exercise a base-branch retarget, so the `edited`-driven re-plan path for `release_preview` is verified only through `createCIPlan`'s unit-level behavior.

.github/workflows/ci.yml
diegohaz: Fixed in [`ee3955c61`](https://github.com/ariakit/ariakit/commit/ee3955c618c1b69025e7afdc0ec7fa6d1fa99b02): metadata-only title/body edits now use a separate concurrency group, skip Plan and all runners, and emit only distinctly named skipped Gate/alias checks, so they cannot cancel or replace real CI or shadow its required results. Base-branch edits remain in the real-work group and replan normally. Evidence gap: GitHub-hosted dispatch, concurrency, and check-rollup behavior cannot be regression-tested by this repository; validation used GitHub’s documented semantics, live same-SHA run evidence, and three fresh staged-snapshot review passes.

.github/workflows/ci.yml
diegohaz: Fixed in [`5fa16221c`](https://github.com/ariakit/ariakit/commit/5fa16221cf9ff86aef63854228f9b5f4c1835f5d): Plan and Gate now set up only the Node version pinned in `package.json` and invoke the dependency-free `packages/ariakit-scripts/src/ci-run.ts` entry point, removing both `pnpm install` steps while preserving the existing developer-facing commands. The focused 14-test suite, full TypeScript check, lint/formatting, and direct Plan/Gate parity probes pass.

diegohaz:
**Fixed**

- [`ee3955c61`](https://github.com/ariakit/ariakit/commit/ee3955c618c1b69025e7afdc0ec7fa6d1fa99b02) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653862583)): Metadata-only edits no longer cancel, replace, run, or shadow real CI; base retargets still replan.
- [`5fa16221c`](https://github.com/ariakit/ariakit/commit/5fa16221cf9ff86aef63854228f9b5f4c1835f5d) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653862980)): Plan and Gate now use a dependency-free Node entry point without installing pnpm dependencies.

**Evidence gaps**

- [`ee3955c61`](https://github.com/ariakit/ariakit/commit/ee3955c618c1b69025e7afdc0ec7fa6d1fa99b02) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653862583)): GitHub-hosted dispatch, concurrency, and check-rollup semantics have no practical repository regression harness; the fix was validated against official semantics, live run evidence, and staged-snapshot reviews.

.github/workflows/ci.yml
diegohaz: Superseded by [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef): the `edited` activity type and all metadata-specific routing were removed, so title/body edits create no CI workflow run or skipped check suite. Base retargets intentionally require a fresh supported event in v1.

diegohaz:
**Obsolete**

- [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653940088)): supersedes the metadata-only routing by removing the `edited` trigger entirely. Title/body edits now create no CI workflow run or skipped check suite. Live verification after editing this PR description shows only the real [`14ef46b20` CI run](https://github.com/ariakit/ariakit/actions/runs/30230061098), with Plan complete and the selected lanes executing.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
+name: CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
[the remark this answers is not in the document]
diegohaz: Won’t fix in this PR. Adding `edited` back with workflow/job guards recreates the exact user-visible failure verified on [`5fa16221c`](https://github.com/ariakit/ariakit/commit/5fa16221cf9ff86aef63854228f9b5f4c1835f5d): title/body edits still attach an all-skipped CI check suite to the head commit. [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef) intentionally limits v1 to events that always require CI. Base retargeting is documented as requiring a fresh supported event (push, existing-label change, or close/reopen); handling it automatically without metadata check pollution would require a separate dispatch layer and is outside this PR’s narrowed scope.

diegohaz:
**Won’t fix**

- [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653942314)): declined restoring `edited`. The guarded version was live-tested and still attached an all-skipped CI suite to the head commit. In v1, base retargets require a fresh supported event; automatic retarget-only dispatch would need a separate routing layer.

packages/ariakit-scripts/src/index.ts
```diff
@@ -67,6 +67,29 @@ program
...
+program
+  .command("ci-gate")
+  .description("Verify that planned CI workflows completed successfully")
+  .requiredOption("--plan <json>", "Serialized CI plan")
+  .requiredOption("--results <json>", "Serialized GitHub Actions job results")
+  .action(async (options) => {
+    const { runCIGate } = await import("./ci.ts");
+    runCIGate(options);
+  });
+
```
diegohaz: Moving the workflow to `ci-run.ts` in [`5fa16221c`](https://github.com/ariakit/ariakit/commit/5fa16221cf9ff86aef63854228f9b5f4c1835f5d) orphaned these two commands. `ci.yml` now runs `node packages/ariakit-scripts/src/ci-run.ts plan` and `... gate`, and a repository-wide search finds no other caller:

```text
packages/ariakit-scripts/src/index.ts:71:  .command("ci-plan")
packages/ariakit-scripts/src/index.ts:84:  .command("ci-gate")
.github/workflows/ci.yml:47:        run: node packages/ariakit-scripts/src/ci-run.ts plan
.github/workflows/ci.yml:156:        run: node packages/ariakit-scripts/src/ci-run.ts gate
```

That leaves two entry points for the same two operations with different interfaces — flags here, environment variables in `ci-run.ts` — and the surviving one is the only one exercised.

The reason this is worth removing rather than leaving as a convenience is that the dead path is not type-checked. `commander@15.0.0` declares `action(fn: (this: this, ...args: any[]) => void | Promise<void>): this;` (`typings/index.d.ts:626`), so `options` here is `any` and `runCIPlan(options)` / `runCIGate(options)` are unchecked calls. `ci-run.ts` builds its argument object field by field, so it is checked against `RunCIPlanOptions` and `RunCIGateOptions`.

Concretely: renaming `baseRef` on `RunCIPlanOptions`, or adding a required field to it, updates `ci-run.ts` under type checking while these commands keep compiling and silently pass an object that no longer matches. Someone debugging the planner locally would reach for the discoverable `pnpm exec ariakit ci-plan --base ... --head ...`, get a `TypeError` or a wrong plan, and have no signal that the supported path is the environment-variable one.

```suggestion
```

If a local debugging entry point is wanted, `ci-run.ts` already serves it with the same variables CI sets:

```sh
BASE_SHA=$(git merge-base origin/main HEAD) HEAD_SHA=$(git rev-parse HEAD) \
BASE_REF=main CI_LABELS='[]' GITHUB_OUTPUT=/dev/null \
node packages/ariakit-scripts/src/ci-run.ts plan
```

**Assessment**

- Severity: Low (verified), because the dead commands cannot affect CI, which no longer calls them.
- Priority: P2 (estimate, 80% confidence), because it is unreferenced code this PR introduced and then orphaned within the same PR, but nothing depends on removing it.
- Breakage likelihood: Possible (estimate, 70% confidence), because it requires someone to later change the planner's option shape or invoke the stale CLI, both plausible while this orchestrator is still being iterated on.
- Fix effort: Easy (verified), because it is a deletion of the two blocks with no other references.

diegohaz (commented):
Re-review. Baseline is the head recorded in my previous marked review, `034d0bb4b4aec235c2934b941599a370120ef599`; current head is `14ef46b20eefcb5160717532657e997cb8eef8ef`. New commits reviewed: `ee3955c61`, `5fa16221c`, `14ef46b20`. I also re-checked whether those commits expose anything new in the surrounding code rather than limiting myself to the incremental diff.

Both findings from the previous review are fixed, and I verified each against the current head rather than relying on the replies. One new finding is published inline.

## Validation of the previous findings

**Metadata edits cancelling in-flight runs** — fixed, verified. `14ef46b20` drops `edited` from `.github/workflows/ci.yml:5`, so the trigger list is now `[opened, synchronize, reopened, labeled, unlabeled]`. The workflow contains no remaining `github.event.changes` routing, and the intermediate metadata-specific concurrency machinery from `ee3955c61` is fully gone; the concurrency block is back to a plain `ci-${{ github.ref }}` group. Run history shows the intermediate design's failure mode too: run `30228807195` on `5fa16221c` completed with conclusion `skipped`, which is the all-skipped metadata check suite described in the reassessment.

**`pnpm install` on the serialized critical path** — fixed, verified with a measurable improvement. Job timings for the current head (run `30230061098`) against the previous head (run `30193723390`):

```text
                     Plan job   Checkout   Node/env setup   Command
034d0bb4b (before)      59s        17s          35s           1s
14ef46b20 (after)       24s        19s           1s          <1s
```

Plan finished 26s after the run started, versus 2m07s before the first lane could begin. The `Set up Node` log confirms `node-version-file: package.json` resolves the pinned runtime (`Found in cache @ /opt/hostedtoolcache/node/24.18.0/x64`, `node: v24.18.0`), so `actions/setup-node@v7` does read `devEngines.runtime` from `package.json` with no `engines` or `volta` field present.

I also confirmed the dependency-free claim independently rather than inferring it from the green job. I copied `ci.ts` and `ci-run.ts` into a scratch repository containing only a `{"type":"module"}` package.json and no `node_modules`, and ran the plan command against a real two-commit range on Node 24.18.0. It produced the expected `$GITHUB_OUTPUT` lines, including a single-line `plan={"version":1,"workflows":{...}}` value, and `gate` with an empty `CI_PLAN` exits non-zero, so the failure mode stays fail-closed when Plan does not publish an output.

## New finding

Published inline on `packages/ariakit-scripts/src/index.ts`: the `ci-plan` and `ci-gate` commander commands added in `7d824be26` lost their only caller in `5fa16221c` and are now an unreferenced second interface for the same operations. Because `commander@15.0.0` types the `action` callback argument as `any`, that path is not type-checked against `RunCIPlanOptions` or `RunCIGateOptions` and can silently drift from the environment-variable path CI actually runs. Severity Low, P2, Possible, Easy.

## Also checked, no finding

- **`ci-run.ts` environment handling.** `getEnvironmentVariable` rejects empty strings, so I checked every variable it reads for a legitimate empty value. `CI_LABELS` is the only candidate: it comes from `toJSON(github.event.pull_request.labels.*.name)`, and this PR carries no labels, yet the Plan job on the current head got past that call and logged `Changed files (17):`. An unlabeled PR yields `[]`, not an empty string. `CI_PLAN` is empty only when Plan fails or is skipped, where throwing is the correct fail-closed behavior.
- **`labeled` / `unlabeled` inheriting the cancellation problem.** These types are also new in this PR, and a label change mid-run cancels and restarts CI the same way `edited` did. I checked whether that is reachable in practice: none of the last 30 pull requests carry any label, Renovate is not configured to apply labels, every label defined in the repository is issue-oriented (`bug`, `stale`, `docs`, and so on), and the `ci:*` planner labels do not exist yet. The only realistic label event is someone deliberately adding a `ci:*` label, where re-planning is exactly the intent. Not a finding.
- **Repository checks on the current head.** `pnpm test` on both `ariakit-scripts` CI test files (16 pass), `tsc -b --force packages/ariakit-scripts` (clean), `oxlint` over `packages/ariakit-scripts/src/` and `oxfmt --check` on `ci-run.ts` and `ci.yml` (clean).
- **`ci-run.ts` style.** Braced conditionals, `function` for top-level declarations, no unsafe assertions, and `process.argv[2]` is handled through a `string | undefined` parameter instead of a non-null assertion. Consistent with the repository's conventions.
- **The rest of the orchestrator.** Nothing in the three new commits touches the planner policy, artifact namespacing, reusable-workflow nesting, or the required-check aliases, and I re-read `ci.yml` in full to confirm no metadata-era remnants survived the revert.

## Previously reviewed threads

All four review threads on this PR are resolved, so per this workflow they are not re-adjudicated. For the record, I did confirm the substance of the two that belong to my previous review, as described above. The Codex thread about the Gate payload size remains fixed at this head, and the CodeRabbit thread about base-branch retargeting was withdrawn by its author after the maintainer's reply.

On that last one I have no material new evidence to add, so I am not reopening it, but recording where I landed: removing `edited` does leave a stale-plan window when a pull request is retargeted across diverged branches, since the existing green checks were planned against the old base. The common case is safe — when a stacked pull request is auto-retargeted after its parent merges, the new base already contains the parent's commits, so the merge-base diff is unchanged. The exposure is a manual retarget to a base that lacks commits present in the old one, which can under-select lanes and leave `release_preview` stale. The PR description now states this explicitly and lists the recovery paths, and the reassessment records why the alternatives were rejected, so this is a documented v1 limitation rather than an unhandled case.

## Not fully validated

- Run `30230061098` on the current head was still in progress when I finished, with Gate not yet started. Plan is verified green; the end-to-end Gate and alias behavior on this head is unverified, though it is unchanged from the previously green run apart from the setup step.
- I did not exercise a base retarget or a label event against the live workflow; those paths are reasoned from the workflow definition and the run history.

.github/workflows/ci.yml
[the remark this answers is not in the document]
diegohaz: Won’t fix in this PR. The guarded `edited` implementation was live-tested on [`5fa16221c`](https://github.com/ariakit/ariakit/commit/5fa16221cf9ff86aef63854228f9b5f4c1835f5d) and still attached an all-skipped CI suite to the head commit on ordinary title/body edits. [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef) intentionally narrows v1 to events that always require CI; the PR description documents that a base retarget must be followed by a push, existing-label change, or close/reopen. Automatic retarget-only dispatch without metadata check pollution requires a separate routing layer and is outside this PR’s scope.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -0,0 +1,431 @@
...
+      file,
+    )
+  ) {
+    addReason(plan, "build_styles", `Generated styles input: ${file}`);
+    matched = true;
+  }
+
+  if (
+    file.startsWith("app/public/og-image/") ||
+    (file.startsWith("app/src/") && !isTestFile(file))
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`67055aaf4`](https://github.com/ariakit/ariakit/commit/67055aaf4): runtime package changes now conservatively select OG generation through the existing `isPackageRuntimePath` boundary. The focused regression covers the reported Popover arrow import chain, and negative cases keep package tests and readmes excluded. The planner suite passes 16/16.

packages/ariakit-scripts/src/index.ts
diegohaz: Fixed in [`67055aaf4`](https://github.com/ariakit/ariakit/commit/67055aaf4): removed the orphaned Commander `ci-plan` and `ci-gate` registrations. The workflows now have one supported interface—the dependency-free, type-checked `ci-run.ts` entry point—and a repository-wide search confirms no remaining caller depends on the deleted commands.

diegohaz:
**Fixed**

- [`67055aaf4`](https://github.com/ariakit/ariakit/commit/67055aaf4) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3654020096)): package runtime changes now select OG generation, with focused positive and negative planner regressions.
- [`67055aaf4`](https://github.com/ariakit/ariakit/commit/67055aaf4) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3654020635)): removed the orphaned, untyped Commander planner/gate adapters; CI keeps the single dependency-free typed runner.

**Won’t fix**

- [`14ef46b20`](https://github.com/ariakit/ariakit/commit/14ef46b20eefcb5160717532657e997cb8eef8ef) ([reply](https://github.com/ariakit/ariakit/pull/6865#discussion_r3653993948)): restoring `edited` would recreate the live-verified all-skipped check-suite pollution. Base retargets require a fresh supported event in v1.

diegohaz (commented):
Re-review. Baseline is the head recorded in my previous marked review, `14ef46b20eefcb5160717532657e997cb8eef8ef`; current head is `67055aaf4711f1235c511b54acbc4ff3cd365c6e`. One new commit: `67055aaf4`. I reviewed its diff and then looked for issues it exposes in the surrounding code rather than stopping at the three changed files.

**No new substantive findings.** Details of what I checked and what I deliberately did not raise follow.

## Validation of the previous finding

**Orphaned `ci-plan` / `ci-gate` commander registrations** — fixed, verified. `packages/ariakit-scripts/src/index.ts` is now byte-identical to `main`: `git diff origin/main..HEAD -- packages/ariakit-scripts/src/index.ts` is empty, so the file has left the pull request's diff entirely. `ci-run.ts` is the single supported entry point, and it stays type-checked against `RunCIPlanOptions` and `RunCIGateOptions`.

I also closed the loop on the item I left unvalidated last time. Run `30230061098` on `14ef46b20` has since completed successfully, with Gate green in 9s, down from 39s before the setup change. The current head's run `30231416342` has also completed successfully end to end: Plan 25s, Gate 6s.

## The `og_images` change

The added `isPackageRuntimePath(file)` clause in `ci.ts:186` is correct, and I verified the reported chain is real rather than taking the Codex report at face value. `app/src/components/placeholder-popover-arrow.react.tsx:10` imports `POPOVER_ARROW_PATH` from `@ariakit/react-components/popover/popover-arrow-path`, and `packages/ariakit-react-components/src/popover/popover-arrow-path.ts:20` exports it, so package runtime code really can change rendered OG output.

Because that finding is an instance of a class — a generated-artifact workflow whose generator transitively reads inputs outside its declared path set — I checked the other two generators for the same bug rather than assuming this was the only one:

- **`build_styles` is complete without package code.** `discoverModulePaths` in `app/src/lib/build-styles.ts` matches only `/@import\s+"\.\/(ak-[^"]+\.css)"\s*;/` against `app/src/styles/ariakit.css`, so it follows relative `ak-*.css` siblings and ignores the `@import "@ariakit/tailwind";` on line 9 of that file. `styles.json` therefore cannot be affected by `packages/`, and `build-styles.ts` itself imports only `node:fs/promises`, `node:path`, `./styles-shared.ts`, and types from `./styles.ts` — all four already in the `build_styles` rule.
- **`docs` is complete.** `pnpm run docs` fans out to per-package `ariakit docs --write` scripts in seven packages, each reading its own `src`. Those inputs are covered by the package-runtime rule, and the generator itself lives under `packages/ariakit-scripts/`, which is a repository-wide input.
- **The OG renderer reads nothing else.** `app/src/pages/og-image/_generate-images.ts` and `api.ts` import only from `#app/lib/**` and `astro:content`, and fonts come from `document.fonts.ready` rather than a checked-in asset, so nothing under `app/public/` outside `og-image/` feeds the output. `app/public/icon.svg` correctly selects Main, App, and Perf but not OG images.

## Also checked

- **Planner behavior on the current head**, re-running my boundary battery: unmatched Markdown, package tests, package readmes, app tests, package runtime, app pages, app styles, `app/public/og-image/`, `app/public/icon.svg`, root `examples/`, `components/*.md`, lockfiles, unknown paths, and `.changeset/` under both `main` and `next` bases. Every result matches the documented table, and the PR description's table row and the `dialog.ts` example line were both updated to include OG images, so the description and implementation agree.
- **Empirical check against nine recent pull requests**, feeding their real changed-file lists through `createCIPlan`. Six select all eight workflows, one selects seven, one selects four. I inspected the reasons for every full-CI case and all of them trace to `package.json` or `pnpm-lock.yaml` under the intentional dependency rule — none came from the unknown-path fallback. Worth knowing that Renovate's dependency traffic means the measured savings on a typical week are smaller than the classifier's precision suggests, but that is the conservative contract this PR deliberately chose, not a defect.
- **`perf.yml` permissions under `workflow_call`.** Its `post` job posts and updates the PR comment, which would normally need `pull-requests: write`, and `ci.yml` grants the `perf` job only `actions: read` and `contents: read`. That is fine: the job mints a separate token with `actions/create-github-app-token@v3` (`permission-issues: write`, `permission-pull-requests: write`) at `perf.yml:1127`, so it never relies on `GITHUB_TOKEN`. Every perf job declares exactly the `actions: read` and `contents: read` the caller grants.
- **Repository checks on the current head:** `pnpm test` on both CI test files (18 pass, up from 16), `tsc -b --force packages/ariakit-scripts` clean, `oxlint` over `packages/ariakit-scripts/src/` and `oxfmt --check` on the changed files clean.

## Not raised

- **The new negative test uses `packages/ariakit-react-components/src/popover/popover.test.ts`, which does not exist**, while its positive counterpart uses the real `popover-arrow-path.ts`. The planner is a pure path classifier, the suite already uses representative paths such as `packages/ariakit-components/src/dialog/dialog.ts` and `tooling/new-config.json`, and the case that encodes the reported bug is anchored on a real file. Not worth changing.
- **Package runtime code now selects seven of the eight workflows**, leaving only `build_styles`. Narrowing that would require import-graph analysis, which this PR explicitly rules out as a design principle, and OG generation is a single macOS job well inside the critical path set by Perf. Declined non-goal.
- **The dependency fail-closed rule dominates the real-world PR mix.** Narrowing it would mean interpreting lockfile diffs to attribute changes to workspaces — a new subsystem, and a direct contradiction of the stated motivation. Declined non-goal rather than a follow-up, since the PR argues for this behavior deliberately.

## Previously reviewed threads

All seven review threads on this pull request are resolved, so none were re-adjudicated. Two of them were closed by this commit: mine on the orphaned commander commands, and the Codex `og_images` report, both verified above. The two base-retarget threads, from CodeRabbit and Codex, were declined by the maintainer with reasoning I examined last cycle; I still have no material new evidence, and the PR description now documents the v1 limitation and its recovery paths, so I am not reopening them.

## Not fully validated

- I did not exercise a base retarget or a label event against the live workflow.
- I did not run the browser, visual, or Perf suites locally; their behavior under the orchestrator is established by this PR's three green end-to-end runs.

Issue #6858, opened 2026-07-25
ComboboxSelect scroll into view can retarget on hover and misses late items


diegohaz:
`ComboboxSelectScrollIntoView` (`packages/ariakit-react-components/src/combobox/combobox-select.tsx`) brings the active item into view when a `ComboboxSelect` popup opens and after each keyboard move. It schedules the scroll across a few animation frames and reads `activeId` lazily inside those callbacks, because the store only points the active item at the selection a commit after the popup opens, and an offscreen item renders later still.

Reading lazily leaves two gaps. Both were found while reviewing #6832 and neither is a regression against the legacy `Select`, whose one-shot scroll came from the dialog's `autoFocusOnShow` calling `element.focus()` without `preventScroll`.

## Hovering can retarget an in-flight scroll

The effect deliberately does not depend on `activeId`, so hovering an item never starts a scroll, and the correction pass that runs a couple of frames later now skips when the active item is no longer the one the first scroll picked. What remains is the first scheduled scroll itself: it reads whichever item is active when it fires.

Reproduction: open a `ComboboxSelect` whose selection is far down a scrollable list, or press an arrow key in an already open one, then move the pointer onto the clipped item at the edge within roughly two animation frames. That first callback acts on the hovered item instead of the intended one, shifting the list under the pointer.

On the move path the displacement is bounded by `block: "nearest"` to less than one item height, and the pass is armed with the item the move targeted, so a pointer arriving meanwhile is ignored.

The open path is worse. It can't be armed the same way, because the store only points the active item at the selection a commit after the popup opens, so the pass resolves the active item when it runs. Measured against `app/src/sandbox/combobox-select-scroll-into-view`, a pointer entering the list within roughly 0 to 30 ms of the popup being placed replaces the selection as the target, and because nothing re-points the active item at the selection afterwards, the selection is then never brought into view at all: the list stays at `scrollTop` 0, which is the whole symptom this machinery exists to prevent. Reproduced 3 out of 3 times entering at 0 ms and 1 out of 3 at 16 ms; from 32 ms onward the selection is presented normally.

Real pointer movement is required, since a mousemove synthesized by the scroll itself under a stationary pointer does not set the active item.

[...]

PR #6872, opened 2026-07-27
Add ComboboxSelect scroll regression tests

diegohaz:
## Motivation

[Issue #6858](https://github.com/ariakit/ariakit/issues/6858) documented two scroll scheduling gaps found while ComboboxSelect was under review: immediate pointer entry could retarget the opening scroll, and a selected item registering after the old frame window could remain offscreen. Current `main` already handles both cases through Composite's scheduled focus and presentation path, but the exact regressions were not preserved in durable coverage.

## Solution

Add a semantic `combobox-select-scroll-into-view` sandbox and two browser tests. The first performs a single real-pointer opening sequence, verifies that hover retargets the active item, and confirms that the selected option remains in the viewport. The second registers the selected option 750 ms after opening, then checks that it becomes active and scrolls into view.

Add focused happy-dom coverage for the late-registration path by spying on `scrollIntoView` and proving that the selected option's own element receives the call. The integration test polls the lazy option query with the existing 3-second allowance, then polls the semantic presentation outcome because Composite intentionally schedules this work after paint, which can complete after item registration under React 18.

```ts
await expect
  .poll(q.option.lazy("Item 24"), { timeout: 3000 })
  .toBeInTheDocument();
const selected = q.option.ensure("Item 24");
await expect.poll(() => scrollIntoView.mock.instances).toContain(selected);
```

The browser tests preserve the real pointer, animation-frame, overflow, clipping, and viewport behavior across Chrome, Firefox, and Safari. The integration test adds a direct, geometry-independent assertion for which element receives `scrollIntoView`.

## Verification

- Focused browser matrix: 6/6 passed across Chrome, Firefox, and Safari with no retries.
- Focused React 19 integration test: 1/1 passed.
- Focused React 18 integration test: 1/1 passed.
- `pnpm lint-fix` passed.
- `pnpm tsc` passed with no diagnostics.
- The exact browser tests fail against known-broken commit [`26037e64`](https://github.com/ariakit/ariakit/commit/26037e64a4ca03fea47a509291171c5a6aea0ed4) with viewport ratio 0.
- The exact `expect.poll` integration test fails against the same known-broken commit because the selected option never receives `scrollIntoView`.

## Preview

Exact [`a962b4363`](https://github.com/ariakit/ariakit/commit/a962b436369fc1eea51acd8944832f89958f8f3e) behavior revision: pointer hover retargets the active option while Item 24 remains visible, then a selected Item 24 registering after 750 ms becomes active and scrolls into view. Later commits change only integration-test assertion timing and expression.

[ariakit-6872-a962b4363-short.webm](https://github.com/user-attachments/assets/988fe298-fe13-47e4-abf0-9f1470ff2bd3)

<details>
<summary>Cycle 3: Keep split browser and happy-dom coverage</summary>

**Assessment**

The protected regression has medium user impact: an asynchronously registered or virtualized selected option can remain offscreen, leaving the visible popup disconnected from its selected value. The late-registration path is uncommon but realistic, and both React 18 and React 19 are supported. Real pointer input, geometry, clipping, and viewport placement remain browser-only concerns.

The repeated cycles came from over-modeling the pointer timing with retries and frame flushing, controlling the fixture's open state unnecessarily, and finally assuming that item registration and Composite's deliberately post-paint presentation complete together. The first two sources of complexity were removed. The remaining happy-dom duplicate is small and checks a distinct contract: the late selected element itself receives `scrollIntoView`.

**Decision**

Keep the simplified fixture, linear browser tests, and focused happy-dom duplicate. Use `expect.poll` for the semantic DOM and `scrollIntoView` outcomes so React scheduler timing is not part of the contract; do not add sleeps, manual frame counts, or a longer custom timeout. Happy-dom intentionally does not attempt to prove pointer or layout behavior, which remains covered by the three browser projects.

</details>

### DIFF
--- a/app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
+++ b/app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
@@ -27,7 +27,6 @@ function Fixture({ label, lateItems }: FixtureProps) {
 
   return (
     <Ariakit.ComboboxProvider
-      open={open}
       setOpen={setOpen}
       defaultSelectedValue={selectedValue}
     >
--- a/app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
+++ b/app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
@@ -1,0 +1,67 @@
+import * as Ariakit from "@ariakit/react";
+import { useEffect, useState } from "react";
+
+const items = Array.from({ length: 30 }, (_, index) => `Item ${index + 1}`);
+const selectedValue = "Item 24";
+
+interface FixtureProps {
+  label: string;
+  lateItems?: boolean;
+}
+
+function Fixture({ label, lateItems }: FixtureProps) {
+  const [open, setOpen] = useState(false);
+  const [showLateItems, setShowLateItems] = useState(false);
+
+  useEffect(() => {
+    if (!lateItems) return;
+    if (!open) {
+      setShowLateItems(false);
+      return;
+    }
+    const timeout = window.setTimeout(() => setShowLateItems(true), 750);
+    return () => window.clearTimeout(timeout);
+  }, [lateItems, open]);
+
+  const renderedItems = lateItems && !showLateItems ? items.slice(0, 4) : items;
+
+  return (
+    <Ariakit.ComboboxProvider
+      open={open}
+      setOpen={setOpen}
+      defaultSelectedValue={selectedValue}
+    >
+      <Ariakit.ComboboxSelectLabel>{label}</Ariakit.ComboboxSelectLabel>
+      <Ariakit.ComboboxSelect style={{ display: "block" }} />
+      <Ariakit.ComboboxPopover
+        unmountOnHide
+        gutter={4}
+        sameWidth
+        aria-label={`${label} options`}
+        style={{
+          background: "white",
+          border: "1px solid gray",
+          maxHeight: 120,
+          overflow: "auto",
+        }}
+      >
+        {renderedItems.map((item) => (
+          <Ariakit.ComboboxItem
+            key={item}
+            value={item}
+            style={{ display: "block", padding: "4px 8px" }}
+          />
+        ))}
+      </Ariakit.ComboboxPopover>
+    </Ariakit.ComboboxProvider>
+  );
+}
+
+export default function Example() {
+  return (
+    <>
+      <Fixture label="Pointer race" />
+      <Fixture label="Late items" lateItems />
+    </>
+  );
+}
--- a/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
+++ b/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
@@ -1,4 +1,4 @@
-import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+import { withFramework } from "#app/test-utils/preview.ts";
 
 withFramework(import.meta.dirname, async ({ test }) => {
   // https://github.com/ariakit/ariakit/issues/6858
@@ -20,26 +20,16 @@ withFramework(import.meta.dirname, async ({ test }) => {
     const listX = selectX;
     const listY = selectBox.y + selectBox.height + 12;
 
-    // Pointer entry only intersects the vulnerable opening window, so retry the
-    // complete interaction until hover actually retargets the active item.
-    await test
-      .expect(async () => {
-        try {
-          await page.mouse.move(selectX, selectY);
-          await page.mouse.down();
-          await page.mouse.up();
-          await page.mouse.move(listX, listY);
-          await flushFrames(page, 8);
+    // The fixed path presents the selection before pointer entry. The broken
+    // path lets this hover retarget the pending scroll, leaving it offscreen.
+    await page.mouse.move(selectX, selectY);
+    await page.mouse.down();
+    await page.mouse.up();
+    await page.mouse.move(listX, listY);
 
-          await test.expect(activeItem).toHaveCount(1);
-          test.expect(await activeItem.textContent()).not.toBe("Item 24");
-          await test.expect(selected).toBeInViewport();
-        } finally {
-          await page.keyboard.press("Escape");
-          await test.expect(listbox).not.toBeVisible();
-        }
-      })
-      .toPass({ timeout: 20_000 });
+    await test.expect(activeItem).toHaveCount(1);
+    await test.expect(activeItem).not.toHaveText("Item 24");
+    await test.expect(selected).toBeInViewport();
   });
 
   // https://github.com/ariakit/ariakit/issues/6858
--- a/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
+++ b/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
@@ -8,7 +8,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
   }) => {
     const select = q.combobox("Pointer race");
     const selected = q.option("Item 24");
-    let hovered = false;
+    const listbox = q.listbox("Pointer race options");
+    const activeItem = listbox.locator("[data-active-item]");
 
     const selectBox = await select.boundingBox();
     if (!selectBox) {
@@ -19,27 +20,26 @@ withFramework(import.meta.dirname, async ({ test }) => {
     const listX = selectX;
     const listY = selectBox.y + selectBox.height + 12;
 
-    for (let attempt = 0; attempt < 20; attempt += 1) {
-      await page.mouse.move(selectX, selectY);
-      await page.mouse.down();
-      await page.mouse.up();
-      await page.mouse.move(listX, listY);
-      await flushFrames(page, 8);
+    // Pointer entry only intersects the vulnerable opening window, so retry the
+    // complete interaction until hover actually retargets the active item.
+    await test
+      .expect(async () => {
+        try {
+          await page.mouse.move(selectX, selectY);
+          await page.mouse.down();
+          await page.mouse.up();
+          await page.mouse.move(listX, listY);
+          await flushFrames(page, 8);
 
-      const listbox = q.listbox("Pointer race options");
-      const activeItem = listbox.locator("[data-active-item]");
-      await test.expect(activeItem).toHaveCount(1);
-      const activeValue = await activeItem.textContent();
-      if (activeValue !== "Item 24") {
-        hovered = true;
-      }
-      await test.expect(selected).toBeInViewport();
-
-      await page.keyboard.press("Escape");
-      await test.expect(listbox).not.toBeVisible();
-    }
-
-    test.expect(hovered).toBe(true);
+          await test.expect(activeItem).toHaveCount(1);
+          test.expect(await activeItem.textContent()).not.toBe("Item 24");
+          await test.expect(selected).toBeInViewport();
+        } finally {
+          await page.keyboard.press("Escape");
+          await test.expect(listbox).not.toBeVisible();
+        }
+      })
+      .toPass({ timeout: 20_000 });
   });
 
   // https://github.com/ariakit/ariakit/issues/6858
--- a/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
+++ b/app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
@@ -1,0 +1,55 @@
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/6858
+  test("does not lose the opening selection to immediate pointer movement", async ({
+    page,
+    q,
+  }) => {
+    const select = q.combobox("Pointer race");
+    const selected = q.option("Item 24");
+    let hovered = false;
+
+    const selectBox = await select.boundingBox();
+    if (!selectBox) {
+      throw new Error("The select is not visible.");
+    }
+    const selectX = selectBox.x + selectBox.width / 2;
+    const selectY = selectBox.y + selectBox.height / 2;
+    const listX = selectX;
+    const listY = selectBox.y + selectBox.height + 12;
+
+    for (let attempt = 0; attempt < 20; attempt += 1) {
+      await page.mouse.move(selectX, selectY);
+      await page.mouse.down();
+      await page.mouse.up();
+      await page.mouse.move(listX, listY);
+      await flushFrames(page, 8);
+
+      const listbox = q.listbox("Pointer race options");
+      const activeItem = listbox.locator("[data-active-item]");
+      await test.expect(activeItem).toHaveCount(1);
+      const activeValue = await activeItem.textContent();
+      if (activeValue !== "Item 24") {
+        hovered = true;
+      }
+      await test.expect(selected).toBeInViewport();
+
+      await page.keyboard.press("Escape");
+      await test.expect(listbox).not.toBeVisible();
+    }
+
+    test.expect(hovered).toBe(true);
+  });
+
+  // https://github.com/ariakit/ariakit/issues/6858
+  test("presents the selection when its item registers late", async ({ q }) => {
+    await q.combobox("Late items").click();
+
+    const selected = q.option("Item 24");
+    await test.expect(selected).toHaveCount(0);
+    await test.expect(selected).toBeVisible({ timeout: 2_000 });
+    await test.expect(selected).toHaveAttribute("data-active-item");
+    await test.expect(selected).toBeInViewport();
+  });
+});
--- a/app/src/sandbox/combobox-select-scroll-into-view/test.ts
+++ b/app/src/sandbox/combobox-select-scroll-into-view/test.ts
@@ -1,4 +1,4 @@
-import { click, q, waitFor } from "@ariakit/test";
+import { click, q } from "@ariakit/test";
 import { expect, test, vi } from "vitest";
 
 // https://github.com/ariakit/ariakit/issues/6858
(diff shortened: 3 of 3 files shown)


app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
```diff
@@ -0,0 +1,55 @@
...
+    const selectBox = await select.boundingBox();
+    if (!selectBox) {
+      throw new Error("The select is not visible.");
+    }
+    const selectX = selectBox.x + selectBox.width / 2;
+    const selectY = selectBox.y + selectBox.height / 2;
+    const listX = selectX;
+    const listY = selectBox.y + selectBox.height + 12;
+
+    for (let attempt = 0; attempt < 20; attempt += 1) {
```
diegohaz: Why this attempt? If it's because of flakiness, shouldn't we use https://playwright.dev/docs/test-assertions#expecttopass instead?

diegohaz: Yes—the attempts were retrying the timing-sensitive pointer entry until hover actually retargeted the active item. I switched this to a bounded `expect(...).toPass({ timeout: 20_000 })` in 8f0ab0745. Each retry closes the popup in `finally`, and success still requires the active item to differ from Item 24 while the selected item remains in the viewport. The focused matrix passes 6/6 across Chrome, Firefox, and Safari; the updated test still fails against known-broken 26037e64 with `viewport ratio 0`.

diegohaz:
**Addressed**

- Replaced the fixed 20-attempt pointer-race loop with bounded `expect(...).toPass({ timeout: 20_000 })` retries in 8f0ab0745.
- Kept each attempt isolated by closing the popup in `finally`; a retry passes only after hover retargets the active item and the selected item remains in the viewport.
- Replied with the rationale and evidence in https://github.com/ariakit/ariakit/pull/6872#discussion_r3653878228.

**Verification**

- Focused browser matrix passes 6/6 across Chrome, Firefox, and Safari with Playwright retries disabled.
- The updated tests still fail against known-broken 26037e64 because both selected items have `viewport ratio 0`.
- `pnpm lint-fix` and `pnpm tsc` completed successfully.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+          await test.expect(activeItem).toHaveCount(1);
+          test.expect(await activeItem.textContent()).not.toBe("Item 24");
+          await test.expect(selected).toBeInViewport();
+        } finally {
+          await page.keyboard.press("Escape");
+          await test.expect(listbox).not.toBeVisible();
+        }
+      })
+      .toPass({ timeout: 20_000 });
```
diegohaz: The rationale in this comment does not match what the interaction does on this branch, and the machinery it justifies never runs.

I added a document-level capture `mousemove` listener to the sandbox that records the listbox `scrollTop` and the hovered option at the instant the pointer enters the list, then ran this exact sequence 20 times per engine:

| Build | Engine | `scrollTop` at pointer entry | Hovered option | `Item 24` in viewport |
| `8f0ab0745` | Chrome | 1257 (20/20) | `Item 23` (20/20) | yes (20/20) |
| `8f0ab0745` | Firefox | 1257 (20/20) | `Item 23` (20/20) | yes (20/20) |
| `8f0ab0745` | Safari | 1226 (20/20) | `Item 23` (20/20) | yes (20/20) |
| `26037e64` | Chrome | 0 (20/20) | `Item 1` (20/20) | no (20/20) |

On this branch the pointer never lands in the "roughly 0 to 30 ms" window from #6858: the presentation scroll has already committed before the mousemove is delivered. The likely reason is that engines deliver injected mouse input on a frame boundary, so the earliest possible entry is already a frame after the popup is placed. The test is red on `26037e64` for a different reason than the comment states — that build scrolled several frames later, so the same frame-aligned pointer arrives *before* its scroll instead of the test hitting a narrow window.

That leaves two pieces of dead weight:

- **`toPass` never retries.** The interaction is deterministic in both directions (60/60 green here, 20/20 red on `26037e64`), so passing runs finish on the first attempt in 1.3-2.4 s. On a real regression the wrapper converts a ~1 s failure into the full budget: the pointer test took **21.4 s** to fail on `26037e64`, and Safari retries that up to 3x in CI.
- **`not.toBe("Item 24")` never discriminates.** It is satisfied by `Item 1` on the broken build and by `Item 23` on the fixed one, so it filters no attempt.

Dropping the wrapper keeps the same red/green signal and makes the comment describe the real mechanism. I verified this replacement passes 6/6 across Chrome, Firefox, and Safari on this head, and the probe data above shows the identical linear sequence yields `viewport ratio 0` on `26037e64`:

```suggestion
    // The pointer reaches the list only after the presentation scroll has
    // committed, so hovering retargets the active item to a neighbor. The
    // broken build scrolled several frames later, so the same pointer
    // retargeted the scroll itself and the selection was never presented.
    await page.mouse.move(selectX, selectY);
    await page.mouse.down();
    await page.mouse.up();
    await page.mouse.move(listX, listY);
    await flushFrames(page, 8);

    await test.expect(activeItem).toHaveCount(1);
    await test.expect(activeItem).not.toHaveText("Item 24");
    await test.expect(selected).toBeInViewport();
```

If you would rather keep a retry as insurance against a slow runner delivering the mousemove before the popup is hit-testable (the one case where the hover precondition can genuinely fail), keep it but shrink the budget and say that is what it guards, not the vulnerable window.

While here: `flushFrames(page, 8)` has no stated reason for 8. The probe shows the scroll has already landed at entry, so a shorter, explained wait would read better.

**Assessment**

- Severity: Medium (verified), because the comment misdescribes the mechanism a future reader will rely on, and the wrapper adds ~20 s per failing engine-attempt.
- Priority: P1 (estimate, 85% confidence), because it is new code in this PR and the fix is a local deletion.
- Breakage likelihood: Likely (verified), because every run pays the mis-description and every genuine regression pays the full 20 s budget on each engine and retry.
- Fix effort: Easy (verified), because the linear replacement above passes 6/6 on this head with no other changes.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // https://github.com/ariakit/ariakit/issues/6858
+  test("presents the selection when its item registers late", async ({ q }) => {
+    await q.combobox("Late items").click();
+
+    const selected = q.option("Item 24");
+    await test.expect(selected).toHaveCount(0);
+    await test.expect(selected).toBeVisible({ timeout: 2_000 });
+    await test.expect(selected).toHaveAttribute("data-active-item");
+    await test.expect(selected).toBeInViewport();
```
diegohaz, line 43: This case is reliably reproducible in happy-dom, so it should also get a `test.ts` duplicate next to this file.

`ariakit-ariakit-example-playwright-test` asks for a happy-dom duplicate whenever the same observable behavior can be exercised through rendered DOM state, and all three sibling ComboboxSelect sandboxes have one (`combobox-select-offscreen-value/test.ts`, `combobox-select-open-active-item/test.ts`, `combobox-select-default-open/test.ts`). The predecessor of this very sandbox at `26037e64` had one too, and it used the same technique.

One thing to note while writing it: `data-active-item` alone is **not** the discriminating assertion. I confirmed it already passes on `26037e64`, where the item becomes active but is never scrolled. What makes the duplicate a regression test is spying on `scrollIntoView` and checking it was called on the late-registering option.

I wrote and ran this file in both directions: red on `26037e64` (`expected [] to include <div role="option" ...>`, i.e. `scrollIntoView` was never called on `Item 24`) and green on `8f0ab0745`, in ~1.3 s via `pnpm test-react app/src/sandbox/combobox-select-scroll-into-view/test.ts`.

```ts
import { click, q, waitFor } from "@ariakit/test";
import { expect, test, vi } from "vitest";

// https://github.com/ariakit/ariakit/issues/6858
test("presents the selection when its item registers late", async () => {
  using scrollIntoView = vi.spyOn(HTMLElement.prototype, "scrollIntoView");
  await click(q.combobox.ensure("Late items"));
  expect(q.option("Item 24")).toBeNull();

  await waitFor(() => expect(q.option("Item 24")).not.toBeNull(), {
    timeout: 3000,
  });
  const selected = q.option.ensure("Item 24");
  expect(selected).toHaveAttribute("data-active-item");
  expect(scrollIntoView.mock.instances).toContain(selected);
});
```

The browser test still earns its keep for the viewport-intersection assertion, so this is an addition, not a replacement. The pointer-race case above stays browser-only, which is correct: #6858 states real pointer movement is required.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because the fastest and most portable signal for this regression is missing while the slow browser-only one carries it alone.
- Priority: P1 (estimate, 80% confidence), because the repository convention asks for the duplicate and the file above is already verified in both directions.
- Breakage likelihood: Possible (estimate, 70% confidence), because a future scheduling change would only be caught by the three browser projects, which are the suites most often retried or skipped locally.
- Fix effort: Easy (verified), because the file above is complete and passes as written.

app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, [lateItems, open]);
+
+  const renderedItems = lateItems && !showLateItems ? items.slice(0, 4) : items;
+
+  return (
+    <Ariakit.ComboboxProvider
+      open={open}
+      setOpen={setOpen}
+      defaultSelectedValue={selectedValue}
+    >
```
diegohaz, line 32: Only the late-items variant needs to know when the popup opens, and it can get that from `setOpen` alone. Passing `open` as well makes both fixtures controlled, which puts an extra React commit into the exact opening window the pointer test probes and moves the fixture off the default uncontrolled path that #6858 describes.

Dropping the prop keeps the `open` state for scheduling the timeout while leaving the open transition to the store:

```suggestion
    <Ariakit.ComboboxProvider
      setOpen={setOpen}
      defaultSelectedValue={selectedValue}
    >
```

Verified both directions with this change in place: 6/6 green across Chrome, Firefox, and Safari on `8f0ab0745`, and still red on `26037e64` for both the browser test and the happy-dom duplicate, so the regression proof is unaffected.

**Assessment**

- Severity: Low (verified), because behavior is unchanged and the current form still reproduces the regression.
- Priority: P2 (estimate, 80% confidence), because it is a fidelity and simplification improvement rather than a defect.
- Breakage likelihood: Unlikely (verified), because the controlled commit did not stop either build from reproducing.
- Fix effort: Easy (verified), because it is a one-line deletion validated on all three engines.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
diegohaz: Revisiting this thread's outcome with new measurements, because they change the verdict on whether a retry is needed at all.

I instrumented the sandbox with a capture `mousemove` listener recording the listbox `scrollTop` and hovered option at pointer entry, and ran the interaction 20 times per engine. On `8f0ab0745` the pointer entered at `scrollTop` 1257/1257/1226 (Chrome/Firefox/Safari) hovering `Item 23`, 60/60, with the selection already in the viewport. On `26037e64` it entered at `scrollTop` 0 hovering `Item 1`, 20/20, with the selection never presented.

So the pointer entry is deterministic in both directions rather than timing-sensitive: it always lands after the presentation scroll on this branch, and always before it on the broken build. `toPass` never retries here, and the `not.toBe("Item 24")` condition it retries on is satisfied on both builds (`Item 1` vs `Item 23`), so it filters nothing. The cost shows up only on failure — the pointer test takes 21.4 s to fail on `26037e64` versus roughly 1 s for the plain assertion, multiplied by Safari's 3 CI retries.

I verified that removing the wrapper keeps 6/6 green across the three engines on this head. Details and a suggested replacement are in https://github.com/ariakit/ariakit/pull/6872#discussion_r3653950767.

diegohaz (commented):
## Scope

Initial review. Baseline: merge base `7061967df` to head `8f0ab0745`, both commits, plus the surrounding sandbox, test-utils, Playwright, and Vitest configuration, the `useScheduleFocus` implementation in `packages/ariakit-react-components/src/composite/composite.tsx`, the three sibling ComboboxSelect sandboxes, and the fixture that occupied this directory at `26037e64`.

The premise checks out: #6858 is closed as already fixed, this adds the durable coverage it asked for, and the fixture path matches the one the issue names. No changeset is needed for test-only work, and `test-browser.ts` correctly maps to the three desktop projects only.

## Verification performed

Ran against a dev server from this worktree, with the Next.js `webServer` entry temporarily removed from `app/playwright.config.ts` and restored afterwards. The worktree is back at `8f0ab0745` with no leftover changes.

- `pnpm -F app run test src/sandbox/combobox-select-scroll-into-view/test-browser.ts --project=chrome --project=firefox --project=safari --retries=0 --workers=1` — 6/6 pass on this head, 1.3-2.4 s per test, no retries.
- **Red control reproduced.** Checked out `26037e64`, overlaid these two files, and reran: both tests fail with `viewport ratio 0`. The pointer test takes 21.4 s to fail because it exhausts the `toPass` budget.
- **Instrumented probe.** Added a capture `mousemove` listener recording the listbox `scrollTop` and hovered option at pointer entry, 20 attempts per engine. On this head: `scrollTop` 1257/1257/1226 hovering `Item 23`, 60/60, selection in viewport. On `26037e64`: `scrollTop` 0 hovering `Item 1`, 20/20, selection out of viewport.
- **Prototyped both suggestions.** The linear replacement for the pointer test passes 6/6 across the three engines. Dropping `open` from the provider passes 6/6 here and stays red on `26037e64`.
- **Prototyped the missing happy-dom duplicate.** Red on `26037e64`, green here, ~1.3 s.

## Published findings

- [Pointer-race rationale and `toPass` wrapper](https://github.com/ariakit/ariakit/pull/6872#discussion_r3653950767) — the comment's "vulnerable opening window" claim does not hold on this branch; the retry never fires and its condition never discriminates. Severity Medium, P1.
- [Missing happy-dom duplicate for the late-item case](https://github.com/ariakit/ariakit/pull/6872#discussion_r3653950816) — reliably reproducible via a `scrollIntoView` spy; note that `data-active-item` alone passes on `26037e64`. Severity Medium, P1.
- [Unnecessary controlled `open`](https://github.com/ariakit/ariakit/pull/6872#discussion_r3653950863) — `setOpen` alone covers the late-items scheduling. Severity Low, P2.

## Candidates not published inline

- **`toHaveCount(0)` on line 50 is a precondition with a 750 ms budget.** If a loaded runner takes longer than that between the click and the assertion, the assertion inverts and waits 5 s for `Item 24` to disappear. Declined non-goal: no clean alternative expresses "not yet registered" without the same window, and the interval is generous relative to observed timings (the assertion resolves within tens of milliseconds locally). Worth revisiting only if it flakes.
- **Cross-fixture isolation depends on `unmountOnHide`.** Both fixtures render identical item labels and the same `defaultSelectedValue`, so `q.option("Item 24")` is unambiguous only because the closed popover is unmounted. That also makes line 50 work. Declined non-goal: it is correct as written, but a reader changing `unmountOnHide` later would get confusing strict-mode failures rather than a clear signal. Distinct labels per fixture would remove the coupling if you touch this file anyway.
- **`flushFrames(page, 8)` has no stated reason for 8.** Folded into the first inline comment rather than filed separately, since both concern the same block.

## Not fully validated

- The `26037e64` red control ran on Chrome only. Firefox and Safari were verified green on this head (20 probe attempts each) but not red on the broken build.
- Runs used the Astro dev server rather than `preview-lite`, so build-mode hydration timing was not exercised. CI covers that and is green on all checks.

## Existing thread

`PRRT_kwDOBjT0Fc6T6RgM` (r3653846131) asked why the manual attempt loop was used and suggested `toPass`. Validation outcome: already resolved in the current head by `8f0ab0745`, which made that switch. I replied there with material new evidence, since the measurements show the retry is not needed in either direction.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
diegohaz: Fixed in [809cbc5a9](https://github.com/ariakit/ariakit/commit/809cbc5a99d145b6a1dcb133a2890949fd977554): removed the unnecessary `toPass` retry and manual frame flush. The linear real-pointer path now relies on locator assertions and preserves the known-broken viewport failure.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
diegohaz: Fixed in [809cbc5a9](https://github.com/ariakit/ariakit/commit/809cbc5a99d145b6a1dcb133a2890949fd977554): replaced the retry wrapper and unexplained frame flush with the deterministic linear pointer sequence. The active-item assertion documents the hover retargeting, while the viewport assertion remains the red/green signal.

app/src/sandbox/combobox-select-scroll-into-view/test-browser.ts
diegohaz: Fixed in [a962b4363](https://github.com/ariakit/ariakit/commit/a962b436369fc1eea51acd8944832f89958f8f3e): added targeted happy-dom coverage proving late Item 24 becomes active and its own element receives `scrollIntoView`. The exact test is green here and red on `26037e64` because that element never receives the call.

app/src/sandbox/combobox-select-scroll-into-view/index.react.tsx
diegohaz: Fixed in [809cbc5a9](https://github.com/ariakit/ariakit/commit/809cbc5a99d145b6a1dcb133a2890949fd977554): removed `open={open}`. `setOpen` still observes the internal open state to schedule delayed registration while the provider remains on its uncontrolled path.

diegohaz:
## Fixed

- [`809cbc5a9`](https://github.com/ariakit/ariakit/commit/809cbc5a99d145b6a1dcb133a2890949fd977554) ([retry thread](https://github.com/ariakit/ariakit/pull/6872#discussion_r3654111199), [linear sequence thread](https://github.com/ariakit/ariakit/pull/6872#discussion_r3654111255), [provider thread](https://github.com/ariakit/ariakit/pull/6872#discussion_r3654111363)): removed the retry wrapper and manual frame flush, kept the deterministic pointer assertions, and restored the provider's uncontrolled open path.
- [`a962b4363`](https://github.com/ariakit/ariakit/commit/a962b436369fc1eea51acd8944832f89958f8f3e) ([integration-test thread](https://github.com/ariakit/ariakit/pull/6872#discussion_r3654111310)): added happy-dom coverage proving the late option's element receives `scrollIntoView`.

## Verification

- Browser regression matrix: 6/6 across Chrome, Firefox, and Safari.
- React integration regression: 1/1.
- Exact browser and integration tests both fail for their intended reasons on known-broken `26037e64`.
- `pnpm lint-fix` and `pnpm tsc` passed.

PR #6873, opened 2026-07-27
Fix happy-dom proxied ancestor identity

### LINKED ISSUE #6849: happy-dom returns the raw target for form/select ancestors, breaking identity-based ancestor checks in tests
## Context

Discovered while fixing a happy-dom divergence in `@ariakit/test` where `Node.contains()` returned `false` for every descendant of a `<form>`/`<select>` (upstream [capricorn86/happy-dom#2170](https://github.com/capricorn86/happy-dom/issues/2170), PR that adds the `contains` shim: see `packages/ariakit-test/src/shims.ts`).

That shim fixes `contains()`. It deliberately does not fix the rest of the same root cause, which this issue tracks.

## Problem

happy-dom's `HTMLFormElement` and `HTMLSelectElement` constructors return a `Proxy`. When such an element is inserted into a parent, `Node[PropertySymbol.connectedToNode]()` re-points its children's internal parent reference at the **raw target** instead of that proxy. As a result, under happy-dom:

- `child.parentElement` / `child.closest("form")` return the raw target, not the proxy that `createElement`, `querySelectorAll`, `.children`, and React refs hand out.
- `compareDocumentPosition` reports an unrelated node.

So any code that walks ancestors with `parentElement` and compares them **by identity** silently fails to recognize a `<form>`/`<select>`.

## Confirmed impact (happy-dom only; real browsers and jsdom are spec-correct)

Reproduced against happy-dom 20.11.0 with the `contains` shim already applied:

1. `isElementInside()` in `packages/ariakit-react-components/src/dialog/utils/tree-cleanup.ts` walks `element.parentElement` and tests membership in a `WeakSet` populated by `markTreeInside()` from the dialog's `getPersistentElements()` result. `WeakSet` uses identity, so a raw target never matches:

   ```
   body > div > form#persistent > input     // form returned by getPersistentElements
   persistentForm.contains(input)  -> true  // fixed by the contains shim
   isElementInside(input, dialog)  -> false // still wrong, expected true
   // same structure with <div> instead of <form> -> true
   ```

   This feeds `useHideOnInteractOutside` → `isElementWithinDialog`, so under test a non-modal dialog with `autoFocusOnShow={false}` would hide when the user interacts with a persistent `<form>` before the dialog has been focused (the path from #6344).

2. `addRoleNoneCleanup()` in `packages/ariakit-react-components/src/dialog/utils/disable-tree.ts` guards with `contains(el, ancestor)` where `ancestor` comes from `element.parentElement`. With a `<form role="group">` that is both a persistent element and an ancestor of the dialog, the guard returns `false` and `role="none"` is set on the persistent form itself.

Both are **latent**: no example or sandbox currently returns a `<form>`/`<select>` from `getPersistentElements`, which is why the suites are green. The risk is that a future test covering that shape would silently assert wrong behavior, exactly the failure mode the `contains` shim was added to remove.

## Why the current change ships without it

The `contains` shim resolves the reported symptom (a modal popover marking its own `<form>` ancestor `inert`) using public DOM APIs only. Covering ancestor identity means either patching the `parentNode`/`parentElement` getters for every node in the environment, or discovering happy-dom's internal `PropertySymbol.proxy` to map a raw target back to its proxy. Both are a materially larger blast radius than the reported bug justified, and neither is exercised by any current test.

Note that upstream PR [capricorn86/happy-dom#2176](https://github.com/capricorn86/happy-dom/pull/2176) only canonicalizes `isInclusiveAncestor`, so it fixes `contains()` but not the ancestor identity described here.

## Expected outcome

Either:

- extend `patchHappyDOMProxiedContains` (or add a sibling shim) in `packages/ariakit-test/src/shims.ts` so ancestor traversal yields the proxy, with focused coverage in `packages/ariakit-test/src/shims.test.ts` and a regression fixture that returns a `<form>` from `getPersistentElements`; or
- file the ancestor-identity half upstream and track it there, documenting the gap in the shim comment until it is fixed.

diegohaz:
## Motivation

happy-dom returns a `Proxy` for populated `<form>` and `<select>` elements, but reconnects existing children to the raw target when those elements are inserted. Ancestor APIs then disagree about identity and can make Ariakit tests treat descendants of a persistent form as outside a dialog.

This is the remaining identity issue documented in #6849 after the contains-only coverage added while #6832 was in progress. It prevented the persistent-form scenario from having equivalent happy-dom coverage: browsers kept the dialog open after clicking a form descendant, while happy-dom closed it.

For example, before the shim:

```ts
const form = document.createElement("form");
const input = document.createElement("input");
form.append(input);
document.body.append(form);

input.parentElement === form; // false in happy-dom
```

Closes #6849.

## Solution

Patch happy-dom's internal connection hook in `@ariakit/test`, after discovering and validating the relevant symbols from a detached probe, so existing children retain the public proxy as their parent. Install the override only on validated `HTMLFormElement` and `HTMLSelectElement` prototypes; unfamiliar inherited or accessor hook layouts remain unchanged. The shim becomes a no-op when happy-dom fixes the behavior or changes those internals.

The repair runs after happy-dom's recursive connection hook returns. A synchronous custom-element `connectedCallback` during that same insertion can therefore still observe the raw target; supporting this rare timing window would require broader coupling to happy-dom internals and is left to the upstream fix in capricorn86/happy-dom#2253.

Expand package coverage for `parentNode`, `parentElement`, `closest`, `contains`, `compareDocumentPosition`, bubbled `currentTarget`, and unfamiliar hook layouts. Consolidate the browser and happy-dom dialog coverage into the semantic `dialog-persistent-elements` fixture, combining the existing #6344 persistent-element scenarios with the #6849 form-descendant regression while retaining the dedicated outside-interaction checks.

## Verification

- `pnpm build`
- `pnpm build-app-lite`
- `pnpm lint`
- `pnpm tsc`
- Focused `@ariakit/test` shim suites: 21/21
- React 19 and React 18 consolidated happy-dom suites: 12/12 each
- Chrome, Firefox, and Safari consolidated browser suite: 36/36
- Isolated #6849 red/green regression proof with the proxy-parent shim disabled/restored
- Temporary custom-element timing probe confirms the documented callback window and post-insertion repair


packages/ariakit-test/src/shims.ts
```diff
@@ -298,57 +298,73 @@ function patchHappyDOMAnimationFrame() {
...
+  });
 
   return () => {
-    for (const restore of restores) restore();
+    Object.defineProperty(
+      nodePrototype,
+      connectedToNodeSymbol,
+      connectedToNodeDescriptor,
+    );
   };
```
diegohaz: Installing the override on `Node.prototype` makes every node in the environment run the wrapper on every connect and disconnect, even though only two element types can ever have a proxy to repair.

In happy-dom 20.11.1, `[PropertySymbol.proxy]` is assigned in exactly four places: `HTMLFormElement`, `HTMLSelectElement`, `NodeList`, and `DOMTokenList`. The last two are not `Node`s and never receive `connectedToNode`, so the `proxy instanceof window.Node` guard on line 354 can only ever pass for a `<form>` or a `<select>`. Every other node pays a closure call, a `Reflect.apply`, a missing-symbol `Reflect.get` that walks to `Object.prototype`, and an `instanceof` before returning — and `connectedToNode` recurses over the whole subtree on each insertion, plus again from `disconnectedFromNode` on each removal.

Measured locally on a 1092-node `<div>` tree (depth 6, breadth 3), 20 append/remove cycles, median of 12 samples, three separate processes each:

| variant | medians |
| --- | --- |
| pre-PR `contains` shim | 3.61 / 3.66 / 3.74 ms |
| this PR | 6.30 / 6.31 / 6.42 ms |
| scoped variant below | 3.57 / 3.59 / 3.62 ms |

To be clear about impact: at whole-suite level this is not observable. `pnpm test --run` and `pnpm test-react --run` were indistinguishable between this PR and the scoped variant (21.55s / 20.09s vs 22.11s / 20.79s), because DOM insertion is a small share of suite time. So the reason to scope it is not the wall clock, it is blast radius — this rewrites a method on the base class of every node in order to fix two element types, which is close to the blast radius called out in [#6849](https://github.com/ariakit/ariakit/issues/6849#issuecomment-5077586507) ("a shim would have to patch the `parentNode`/`parentElement` getters for every node or reach into `PropertySymbol.proxy`"), and it is wider than what `patchHappyDOMProxiedContains` touched.

Scoping it to the two proxied constructors works because happy-dom's `ClassMethodBinder` resolves the first own descriptor among `[HTMLFormElement, HTMLElement, Element, Node, EventTarget]` (and the `HTMLSelectElement`-first equivalent), so an override on the element prototype is what gets bound onto the raw target when the symbol is reached through the proxy; non-proxy lookups find it through the normal prototype chain. `nodePrototype` stays as the symbol-discovery source.

```suggestion
  const restores: Array<() => void> = [];

  for (const Constructor of [
    window.HTMLFormElement,
    window.HTMLSelectElement,
  ]) {
    if (!Constructor) continue;
    Object.defineProperty(Constructor.prototype, connectedToNodeSymbol, {
      ...connectedToNodeDescriptor,
      value: function connectedToNode(this: Node) {
        Reflect.apply(originalConnectedToNode, this, []);
        const proxy: unknown = Reflect.get(this, proxySymbol);
        if (!(proxy instanceof window.Node)) return;
        for (const child of this.childNodes) {
          if (child.parentNode === proxy) continue;
          Reflect.set(child, parentNodeSymbol, proxy);
        }
      },
    });
    restores.push(() => {
      // The hook is inherited from Node.prototype, so removing the override
      // restores it rather than reassigning it.
      Reflect.deleteProperty(Constructor.prototype, connectedToNodeSymbol);
    });
  }

  return () => {
    for (const restore of restores) restore();
  };
```

I applied exactly this locally and verified it is behavior-identical:

- `pnpm test packages/ariakit-test/src/shims.test.ts --run` (19 passed), and both new assertions still fail against the pre-PR shim.
- `pnpm test-react app/src/sandbox/dialog-persistent-form-6849/test.ts --run`, `pnpm test --run` (659 passed), `pnpm test-react --run` (1035 passed).
- `oxfmt --check`, `oxlint`, and `tsc` clean on the edited file.
- A 45-assertion probe over `parentNode`, `parentElement`, `closest`, `contains`, `compareDocumentPosition`, bubbled `currentTarget`, `composedPath`, `getRootNode`, `form.elements`, `select.options`/`value`, and `MutationRecord.target`, across moving a populated form between parents, `cloneNode(true)`, `innerHTML`, `<template>` content, `DocumentFragment`, shadow roots, a same-document iframe, `insertBefore`, `replaceChild`, text-node children, post-`remove()` state, and a nested `form > div > select > optgroup > option`. Results were byte-identical to this PR's version (and 19 of those assertions flip from `false` to `true` versus the pre-PR shim, so the probe is sensitive to the behavior in question).

**Assessment**

- Severity: Low (verified, 90% confidence). No consumer-visible breakage was established: `@ariakit/test` shims are no-ops in real browsers (`isBrowser` guard), and the suite-level timing difference is within noise. The concrete cost is a `Node.prototype` patch whose scope is much wider than the defect it repairs.
- Priority: P2 (estimate, 80% confidence), because the current code is correct and the change is a contained narrowing rather than a fix.
- Breakage likelihood: Likely that the extra per-node work is paid (verified, 95% confidence) — it runs on every insertion and removal in every happy-dom test; Unlikely that anyone observes it (estimate, 85% confidence).
- Fix effort: Easy (verified, 95% confidence), because the suggestion above is the whole change and it needs no test updates.

app/src/sandbox/dialog-persistent-form-6849/index.react.tsx
```diff
@@ -0,0 +1,32 @@
+import * as Ariakit from "@ariakit/react";
+import { useRef, useState } from "react";
+
+export default function Example() {
```
diegohaz: This fixture reproduces `app/src/sandbox/dialog-6344/` almost exactly: non-modal `Dialog`, `autoFocusOnShow={false}`, `getPersistentElements` returning a persistent container, one field inside that container, and one "Outside field". Both assertions in the new tests already exist there as `stays open when interacting with a persistent element before the dialog is focused` and `still closes when interacting outside before the dialog is focused`. The only new dimension is that the persistent container is a `<form>` instead of a `<div role="region">`.

That leaves two issue-number fixtures with stable overlap, which is the consolidation case in `ariakit-ariakit-example-playwright-test`: move the coherent scenario into one behavior-focused fixture rather than growing a parallel one.

I validated the consolidated shape locally. Adding a persistent `<form>` to `dialog-6344` and pushing it onto the `getPersistentElements` list, plus one test in `test.ts`:

```tsx
      {/* A persistent <form>: happy-dom proxies form elements, so its
          descendants must still resolve to it through ancestor APIs. */}
      <form ref={persistentFormRef}>
        <input
          aria-label="Persistent form field"
          placeholder="Persistent form field"
          className="rounded border border-gray-300 px-3 py-1"
        />
      </form>
```

gives `12 passed` with this PR's shim and `1 failed | 11 passed` against the pre-PR shim, with the failure isolated to the new `keeps descendants of a persistent form inside the dialog context` test. None of `dialog-6344`'s eleven existing assertions changed behavior, so the extra persistent element does not disturb its fixture contract. The full move would also mean adding the matching browser test with the 6849 URL beside it and renaming the directory and preview ID to something semantic such as `dialog-persistent-elements`.

If you judge the two scenarios genuinely non-coherent — for example because you want the 6849 repro to stay minimal and free of the shadow-DOM machinery in `dialog-6344` — keeping them separate is allowed by the same lifecycle. In that case it is worth a sentence in this file saying so, because from the outside the two fixtures read as the same scenario and the next person will not know which one owns persistent-element coverage.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the duplication costs discoverability and future maintenance, not correctness; both fixtures pass today.
- Priority: P2 (estimate, 75% confidence), because the lifecycle allows keeping them separate when the scenarios are judged non-coherent, so this is a call for you rather than a merge blocker.
- Breakage likelihood: Possible (estimate, 70% confidence), scoped to the two fixtures drifting apart — a future change to persistent-element handling gets updated in one and missed in the other.
- Fix effort: Moderate (verified, 85% confidence) for the full consolidation, since it means the fixture edit plus a browser test, a directory rename, and updating the preview ID and test filters; Easy if you instead keep this fixture and document why.

diegohaz (commented):
## Review summary

Initial review. Baseline: merge base `7061967` to head `cb3301e`, then expanded into happy-dom's `Node`/`HTMLFormElement`/`HTMLSelectElement`/`ClassMethodBinder` sources, the surrounding shims in `packages/ariakit-test/src/shims.ts`, and the existing persistent-element fixtures under `app/src/sandbox/`.

The core of the change holds up well. The root cause matches happy-dom 20.11.1's `Node[PropertySymbol.connectedToNode]` re-pointing `childNodes[i][PropertySymbol.parentNode]` at bare `this`, the probe-and-verify gate degrades to a no-op correctly when the parent pointer or the internal symbols change, and the repair is placed at the same point the upstream proposal in [capricorn86/happy-dom#2253](https://github.com/capricorn86/happy-dom/issues/2253) targets, so it should disappear cleanly when that lands.

### Verification performed

- `pnpm test packages/ariakit-test/src/shims.test.ts --run` — 19 passed. Both new assertions are proven regressions: reverting only `shims.ts` to the merge base fails `preserves form and select proxy identity across ancestor APIs` at `child.parentNode` and `reports a proxied form as currentTarget while an event bubbles`, for the expected identity reason.
- `pnpm test-react app/src/sandbox/dialog-persistent-form-6849/test.ts --run` — passes, and fails against the merge-base shim with the dialog already closed at the first `toBeVisible`, so the sandbox is a proven regression too.
- `pnpm test --run` (659 passed), `pnpm test-react --run` (1035 passed), `pnpm tsc`, `pnpm lint` (only pre-existing `examples/menu-slide` warnings).
- Browser coverage confirmed collected: `playwright test --list --project=chrome dialog-persistent-form` lists the new test under the react preview, and CI is green on Chrome, Firefox, Safari, Android, macOS, React 18, and Solid at this head.
- A 45-assertion probe over `parentNode`, `parentElement`, `closest`, `contains`, `compareDocumentPosition`, bubbled `currentTarget`, `composedPath`, `getRootNode`, `form.elements`, `select.options`/`value`/`selectedIndex`, and `MutationRecord.target`, exercised across moving a populated form between parents, `cloneNode(true)`, `innerHTML`, `<template>` content, `DocumentFragment`, shadow roots, a same-document iframe, `insertBefore`, `replaceChild`, `removeChild` and re-append, text-node children, post-`remove()` state, and a nested `form > div > select > optgroup > option`. Nineteen assertions flip from `false` to `true` versus the merge base and none regress, which is broader than the changeset claims.
- Micro-benchmark of DOM insertion cost (1092-node tree, 20 append/remove cycles, median of 12 samples, three processes each), which is what the first inline comment is about.

### Findings published inline

1. `packages/ariakit-test/src/shims.ts` — the override is installed on `Node.prototype` rather than on the two constructors that can actually own a proxy. Scoping it is behavior-identical (verified against the full probe and both suites) and removes a measurable per-node cost. P2.
2. `app/src/sandbox/dialog-persistent-form-6849/index.react.tsx` — the fixture overlaps `dialog-6344` almost completely; the consolidated alternative was validated locally. P2.

### Other candidates and their dispositions

- **CodeRabbit's nitpick on `shims.ts:314-369`** ("replace the private happy-dom symbol probe with an observable DOM behavior check", removing dependencies on `connectedToNode`, `parentNode`, and `proxy`): **invalid**. Detection is already behavior-based — lines 329-336 build a detached `form > div`, read the public `child.parentNode`, and bail out when it already equals the proxy, which is exactly the self-disabling property the suggestion asks for. The symbol dependency is in the repair, not the detection, and there is no way to re-point a child's parent pointer through public DOM API. The only public-API alternatives are overriding `parentNode`/`parentElement` on every node (wider than what is here) or the previous `contains`-only override (which cannot fix `closest`, `compareDocumentPosition`, or `currentTarget`). No reply posted, since the review is a top-level comment with no thread to answer.
- **Stale comment in `shims.test.ts:507-513`**: **declined non-goal.** The test was renamed to `preserves form and select proxy identity across ancestor APIs` and now asserts `parentNode`, `parentElement`, `closest`, and `compareDocumentPosition` before `contains`, but the comment still explains only the `contains` failure mode. The first two sentences remain accurate about the root cause, so this is wording polish rather than a misleading comment.
- **`reports a proxied form as currentTarget while an event bubbles` covers `<form>` only**: **declined non-goal.** The sibling tests iterate both tags. I confirmed via the probe that the `<select>` case (`option.dispatchEvent` bubbling to a listener on `select`) is fixed by the same shim, so the gap is coverage symmetry, not a hole.
- **`page.waitForTimeout(250)` in the new browser test has no explanatory comment**, unlike its `dialog-6344` sibling which documents it as a bounded guard against asserting before the dialog processes the hide: **declined non-goal**, purely a comment.
- **Two `@ariakit/test` changesets in the same unreleased version** (`300-happy-dom-form-contains.md` and `300-happy-dom-proxied-ancestors.md`): **invalid.** They read as complementary rather than redundant — the new entry deliberately omits `contains` and lists only the APIs the earlier entry did not cover.
- **The dropped links to [happy-dom#2170](https://github.com/capricorn86/happy-dom/issues/2170) and [#2176](https://github.com/capricorn86/happy-dom/pull/2176)** from the shim comment: **invalid.** Since the new shim subsumes the `contains` repair and #2253 fixes the shared root cause, the remaining #2253 link is sufficient traceability; the guard behaves correctly whether #2176 or #2253 lands first.
- **Restore-path completeness**: not published. `restores`/`applyBrowserShims()`'s return value is discarded at line 371, and happy-dom's `ClassMethodBinder` copies the resolved hook onto each proxied instance, so already-bound elements would keep the patched version after a restore. Both are pre-existing properties of every helper in this file, not introduced here.

### Not fully validated

- I did not run Playwright against a local server; the app preview requires a full `build-app-lite` plus the Next.js worker build. Collection was verified with `--list`, and the browser projects are green in CI at this head, so the gap is only "I did not watch it run locally".
- I did not re-verify the changeset against a released changelog render, only against the changeset conventions and the sibling entry.
- Solid coverage is unaffected by design (the sandbox is React-only), so I did not exercise `pnpm test-solid` beyond CI's green result.

No correctness or regression defects were found in the shim itself.

packages/ariakit-test/src/shims.ts
diegohaz: Fixed in [`320de4f2d`](https://github.com/ariakit/ariakit/commit/320de4f2d39c93d546a5516eca50a1df92dd7caa): scoped the connection override to validated form/select prototypes and made unfamiliar inherited/accessor hook layouts a no-op. Added targeted coverage for both layouts in the same commit.

app/src/sandbox/dialog-persistent-form-6849/index.react.tsx
diegohaz: Fixed in [`7dd97c823`](https://github.com/ariakit/ariakit/commit/7dd97c8237594e5d2a1f3ea23aa1a99fb97c818c): consolidated #6344 and #6849 into the semantic `dialog-persistent-elements` fixture, preserving the existing outside-click and shadow coverage while adding the form-specific browser and happy-dom regression.

diegohaz:
**Fixed**

- [`320de4f2d`](https://github.com/ariakit/ariakit/commit/320de4f2d39c93d546a5516eca50a1df92dd7caa) ([thread](https://github.com/ariakit/ariakit/pull/6873#discussion_r3653942928)): Scoped the happy-dom connection shim to validated form/select prototypes and added coverage for unfamiliar hook layouts.
- [`7dd97c823`](https://github.com/ariakit/ariakit/commit/7dd97c8237594e5d2a1f3ea23aa1a99fb97c818c) ([thread](https://github.com/ariakit/ariakit/pull/6873#discussion_r3653943932)): Consolidated persistent-element dialog coverage into the semantic fixture and retained #6849 coverage in happy-dom and browsers.

packages/ariakit-test/src/shims.ts
```diff
@@ -298,57 +298,101 @@ function patchHappyDOMAnimationFrame() {
...
-        // already reports a node as containing itself.
-        if (child.contains(otherNode)) return true;
-      }
-      return false;
-    };
+    if (!inheritsOriginalConnectedToNode(Constructor.prototype)) continue;
+    Object.defineProperty(Constructor.prototype, connectedToNodeSymbol, {
+      ...connectedToNodeDescriptor,
+      value: function connectedToNode(this: Node) {
+        Reflect.apply(originalConnectedToNode, this, []);
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and worth recording precisely, but I do not think it should change this shim.

**The claim reproduces.** happy-dom's `CustomElementReactionStack.enqueueReaction` invokes `connectedCallback` synchronously (`custom-element/CustomElementReactionStack.js`), and `Element[PropertySymbol.connectedToDocument]` enqueues it from inside the recursion that `Reflect.apply(originalConnectedToNode, this, [])` drives. So the callback fires before the post-loop repairs the pointer. Against this head:

| observed inside `connectedCallback` | result |
| --- | --- |
| direct child of a populated `<form>`: `this.parentNode === form` | ❌ |
| direct child: `this.closest("form") === form` | ❌ |
| grandchild (`form > div > el`): `this.closest("form") === form` | ❌ |
| grandchild: `this.parentNode === div` | ✅ |
| any of the above, read after the insertion returns | ✅ |
| element appended to an already-connected form (control) | ✅ |

**But every repair I could build at the wrapper level costs more than the gap is worth.** I implemented and measured two, both of which turn all six rows green and leave the 45-assertion ancestor-identity probe byte-identical to this head:

1. Run the hook with the proxy as its receiver — `Reflect.apply(originalConnectedToNode, proxy, [])` — so `childNodes[i][PropertySymbol.parentNode] = this` writes the proxy directly. This also deletes the post-loop entirely, which is the appealing part. But every internal property access inside the hook then goes through the proxy traps, and each one calls `ClassMethodBinder.bind`.
2. Swap `this[PropertySymbol.nodeArray]` to `[]` so the original skips its child loop, then replay that loop with the proxy. This additionally couples the shim to a third internal symbol and to the assumption that `connectedToNode` reads `nodeArray` only for that loop.

Insertion cost on a tree of 30 forms x 10 field wrappers x 1 input, 20 append/remove cycles, median of 12 samples, three processes each:

| variant | medians |
| --- | --- |
| merge-base (`contains` shim) | 3.24 / 3.26 / 3.24 ms |
| this head (post-loop) | 4.17 / 4.27 / 3.97 ms |
| (1) proxy receiver | 10.13 / 10.66 / 10.09 ms |
| (2) `nodeArray` swap | 8.88 / 9.22 / 8.82 ms |

So closing this window costs roughly 2.2x to 2.5x on form-heavy insertion, versus the 27% the current approach costs over the merge base.

**Reachability is narrow.** It needs a custom element whose `connectedCallback` performs identity-based ancestor checks, inside a `<form>` or `<select>` that already had children when it was inserted. Ariakit defines no custom elements, and the one fixture in this repo that does — `app/src/sandbox/hovercard-shadow-widget-6303` — builds its shadow root in the constructor and is not inside a form, so nothing here is affected. Note also that this is strictly better than the merge base, where the pointer was wrong during *and* after insertion; the PR narrows the window rather than opening one.

**Suggested decision path:** keep the current repair and document the residual window in the shim comment, next to the existing `TODO: Remove this shim once the upstream fix ships`. [capricorn86/happy-dom#2253](https://github.com/capricorn86/happy-dom/issues/2253) fixes it at the source by normalizing `self` inside the hook, which is exactly the case a wrapper cannot reach cheaply, so the residual gap disappears with the shim rather than needing its own fix. Something like:

```ts
// The repair lands after happy-dom's hook returns, so a custom element's
// connectedCallback firing during that same insertion still observes the raw
// target through `parentNode` and `closest`. Repairing mid-recursion means
// either running the hook through the proxy or replaying its child loop, both
// of which cost 2x or more on every form/select insertion. Upstream #2253
// fixes it at the source.
```

`ariakit-ariakit-test-workflow` asks for expected happy-dom differences to be explicit in the test or nearby code, and this is one.

**Assessment**

- Severity: Low (verified, 90% confidence), because the wrong identity is observable only inside `connectedCallback` during one insertion and is correct immediately afterwards.
- Priority: P2 (estimate, 80% confidence) for the comment; the behavioral repair itself is not worth its measured cost and belongs upstream.
- Breakage likelihood: Rare (verified, 85% confidence). It requires a custom element doing identity-based ancestor checks in `connectedCallback` inside an already-populated form or select; no such code exists in this repository, and the control case (children appended after connection) is unaffected.
- Fix effort: Easy (verified, 90% confidence) for the comment. Moderate and net-negative for the behavioral repair, since both working implementations regress form insertion by more than 2x.

diegohaz (commented):
## Re-review summary

Re-review. Incremental baseline: `cb3301e`, the head recorded in the previous marked review, to the current head `7dd97c8` (commits `320de4f` and `7dd97c8`). New candidate findings were drawn from those commits plus surrounding code they newly expose, and the one relevant unresolved thread was independently validated against the current head.

Both findings from the previous round are addressed, and each fix is verified rather than taken on trust.

### Previous findings, re-verified

- **Shim scoping** (`320de4f`, thread [`r3653942928`](https://github.com/ariakit/ariakit/pull/6873#discussion_r3653942928), resolved): the override moved from `Node.prototype` to the two proxied constructors. The added `inheritsOriginalConnectedToNode` guard goes further than what I suggested, and it is genuinely load-bearing: removing just that line makes both tests in the new `shims-existing-hook.test.ts` fail with `expected 0 to be greater than 0`, which is exactly the right failure — it proves the shim would otherwise swallow a pre-existing hook rather than merely that the guard exists. The fake-hook setup covers both the inherited-non-original case on `HTMLElement.prototype` and the accessor-descriptor case on `HTMLSelectElement.prototype`.
- **Fixture consolidation** (`7dd97c8`, thread [`r3653943932`](https://github.com/ariakit/ariakit/pull/6873#discussion_r3653943932), resolved): `dialog-6344` became `dialog-persistent-elements`, the standalone `dialog-persistent-form-6849` fixture is gone, each test now carries its own issue URL, and the 6849 test correctly reuses the existing `still closes when interacting outside before the dialog is focused` coverage instead of duplicating it. No references to either old directory remain anywhere in the repository, and Playwright collects all twelve tests from the renamed path under the `chrome` project.

### Verification performed at this head

- `pnpm test --run` (662 passed, 60 files) and `pnpm test-react --run` (1035 passed), `pnpm tsc`, `oxfmt --check` and `oxlint` on the changed shim.
- `pnpm test packages/ariakit-test/src/shims.test.ts packages/ariakit-test/src/shims-existing-hook.test.ts --run` — 21 passed; the new guard's regression proof is described above.
- `pnpm test-react app/src/sandbox/dialog-persistent-elements/test.ts --run` — 12 passed.
- `playwright test --list --project=chrome dialog-persistent-elements` — 12 tests collected from the renamed directory.
- Re-ran the 45-assertion ancestor-identity probe from the previous round (`parentNode`, `parentElement`, `closest`, `contains`, `compareDocumentPosition`, bubbled `currentTarget`, `composedPath`, `getRootNode`, `form.elements`, `select.options`/`value`, `MutationRecord.target`, across moves, `cloneNode`, `innerHTML`, `<template>`, fragments, shadow roots, a same-document iframe, `insertBefore`, `replaceChild`, and nested `form > div > select > optgroup > option`). Byte-identical to the previous head, so the scoping change is behavior-preserving.
- Insertion benchmark on a form-heavy tree (30 forms x 10 field wrappers x 1 input, 20 append/remove cycles, median of 12 samples, three processes): merge base 3.24/3.26/3.24 ms, this head 4.17/4.27/3.97 ms. The `Node.prototype`-wide cost from the previous round is gone.
- All 49 CI checks pass at this head.

### Unresolved thread validated

- **[`r3654110630`](https://github.com/ariakit/ariakit/pull/6873#discussion_r3654110630) (Codex, P2, "Repair parent identity before descendant connection callbacks"): confirmed, replied on the thread.** The mechanism is real — happy-dom invokes `connectedCallback` synchronously from `CustomElementReactionStack.enqueueReaction`, reached from inside the original hook's recursion, so it runs before the post-loop repairs the pointer. I reproduced it: a custom element that is a direct child of a populated `<form>` sees the raw target through `parentNode` and `closest("form")` inside its callback, and a grandchild sees it through `closest("form")`; all of them are correct immediately after the insertion returns, and elements appended to an already-connected form are unaffected.

  I do not think it should change the shim, and the reply carries the evidence. I built two working repairs — running the hook with the proxy as its receiver, and swapping `nodeArray` to replay the child loop — and both close the window while leaving the 45-assertion probe identical, but they cost 10.1-10.7 ms and 8.8-9.2 ms respectively on the same insertion benchmark, against 4.0-4.3 ms for the current approach. That is a 2.2x to 2.5x regression on every form and select insertion to close a window that needs a custom element doing identity-based ancestor checks in `connectedCallback` inside an already-populated form. Nothing in this repository does that; the only `customElements.define` fixture, `app/src/sandbox/hovercard-shadow-widget-6303`, builds its shadow root in the constructor and sits outside any form. The window is also strictly narrower than the merge base, where the pointer was wrong during and after insertion. [capricorn86/happy-dom#2253](https://github.com/capricorn86/happy-dom/issues/2253) fixes it at the source, which a wrapper cannot reach cheaply, so it disappears when the shim does.

  What I did suggest on the thread is a comment recording the residual window next to the existing `TODO`, since `ariakit-ariakit-test-workflow` asks for expected happy-dom differences to be explicit in nearby code. If you would rather track it than annotate it, it wants a follow-up issue; I have not opened one, since the upstream link already covers the fix path and I did not want to create repository state on a judgment that is yours.

### Other candidates from the new commits

- **`shims-existing-hook.test.ts` mutates `window.HTMLElement.prototype` and `window.HTMLSelectElement.prototype` at module scope and restores in `afterAll`**: no finding. Vitest isolates the module registry and environment per test file, the file is picked up only by the default core suite (it matches neither the `*react.test.*` nor the `*solid.test.*` loader globs), and the `afterAll` restores all three descriptors including the delete branch.
- **Changeset not updated**: no finding. `300-happy-dom-proxied-ancestors.md` still describes the shipped behavior accurately, and the scoping change, the new guard, and the fixture rename are all internal.
- **Sandbox preview URL changes from `dialog-6344` to `dialog-persistent-elements`**: declined non-goal. The rename is what the fixture lifecycle prescribes, and a sandbox preview URL is not a stable public surface.

### Not fully validated

- I did not run Playwright against a local server; collection was verified with `--list` and the browser projects are green in CI at this head.
- The two alternative repairs I benchmarked were validated against the package tests, the ancestor-identity probe, and the custom-element repro, but I did not run the full suites against them, since I am recommending against both.

No new defects found in the changes since `cb3301e`.

packages/ariakit-test/src/shims.ts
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because supporting identity reads inside a synchronous happy-dom custom-element callback would require running the whole internal hook through the Proxy or replaying more internals for a rare case with no affected Ariakit scenario. Documented the expected window in [`af2f9e09f`](https://github.com/ariakit/ariakit/commit/af2f9e09fdc0ccfbc4bb20523b34ad787df89c02); the upstream fix handles it at the source.

diegohaz:
**Declined**

- [`af2f9e09f`](https://github.com/ariakit/ariakit/commit/af2f9e09fdc0ccfbc4bb20523b34ad787df89c02) ([thread](https://github.com/ariakit/ariakit/pull/6873#discussion_r3654110630)): Kept synchronous custom-element callback identity reads outside this shim’s supported scope because a wrapper-level repair would broaden happy-dom coupling for a rare, unaffected Ariakit scenario; documented the expected timing window beside the shim and in the PR description.

PR #6879, opened 2026-07-27
Select dependency CI from package manifests

diegohaz:
## Motivation

#6878 removed the CI label overrides. Dependency changes would therefore continue to run full CI whenever the planner sees `package.json` or `pnpm-lock.yaml`, even when the changed manifest identifies a narrower affected workspace.

The latest 40 Renovate pull requests all changed at least one `package.json`, so manifest diffs provide a useful semantic input without relying on labels. We should also keep the global Renovate `pnpmDedupe` behavior, which means an update may rewrite lockfile importers whose manifests were not touched. Those unrelated importer changes must make the planner fall back to full CI.

## Solution

The `CI` workflow is triggered automatically when a pull request is opened, synchronized, or reopened. Its Plan job compares the pull request merge base with the head commit, classifies the changed paths and manifests, and exposes one boolean output per reusable workflow. Each selected workflow runs through `workflow_call`; the always-running Gate validates that every selected workflow succeeded and every unselected workflow was skipped.

For dependency changes, the planner compares each changed `package.json` and selects workflows from the changed dependency fields and the manifest's workspace role. Ordinary source changes continue to use the existing path heuristics, and their workflows are unioned with the dependency selection.

A shared lockfile is ignored only when both sources of evidence agree:

```ts
const hasScopedLockfileChanges =
  changedLockfileImporters.length > 0 &&
  changedLockfileImporters.every((importer) => {
    const file =
      importer === "." ? "package.json" : `${importer}/package.json`;
    return changedPackageJSONFileSet.has(file);
  });
```

The runner reads only the raw `importers:` blocks from the merge-base and head lockfiles. It identifies importer boundaries and compares each block exactly; it does not interpret generated package or snapshot entries, or attempt to validate the full nested YAML schema. Importer evidence is trusted only when both lockfiles expose the same importer set, at least one importer block changed, and every changed importer maps to a changed manifest with a concrete dependency update. Missing or unsupported sections, invalid or ambiguous importer boundaries, added or removed importers, and unrelated importer changes run full CI. Main's frozen install remains responsible for validating the lockfile itself. This protects selective routing when global `pnpmDedupe` rewrites another workspace without turning the planner into a lockfile-schema implementation.

The manifest heuristics intentionally stay conservative:

- reviewed root development tools select Main, with Vitest also selecting Performance and Oxfmt or conventional `@types/*` changes also selecting Docs;
- app, website, guide, examples, Next.js, and template manifests select their explicitly mapped consumer workflows;
- public package runtime and peer dependency changes select all consumer workflows, while public development dependencies select Main and conventional `@types/*` changes also select Docs;
- unknown root dependencies, unknown workspaces, private package changes, scripts/configuration, unsupported manifests, unreadable snapshots, lockfile-only updates, and `peerDependenciesMeta` changes run full CI;
- public version-only Publish metadata without a lockfile selects Main and Release preview;
- a lockfile accompanying a version-only or otherwise unexplained manifest change runs full CI.

This keeps the planner independent of Renovate labels and retains `pnpmDedupe`.

## Verification

- `pnpm test packages/ariakit-scripts/src/ci.test.ts --run` (45 tests)
- `pnpm lint`
- targeted type-aware Oxlint and Oxfmt
- `pnpm tsc`
- verified that the committed `pnpm-lock.yaml` parses, including quoted scoped keys, `workspace:*`, `link:` values, and peer-suffixed versions
- replayed dependency commits from the latest 40 Renovate pull requests: 22 select narrower CI and 18 conservatively remain full CI
- replayed the grouped `@types/node` update from #6695: unrelated lockfile importers force full CI
- replayed Publish pull request #6820: Main and Release preview

No changeset is needed because this only changes repository CI infrastructure.

## Review gate reassessment

<details>
<summary>Cycle 3: Narrow lockfile explanation to concrete dependency records</summary>

**Assessment**

Three review rounds found fail-open boundaries in the lockfile explanation. The first affected grouped manifest changes generally; the next two involved semantic edge cases in nested `peerDependenciesMeta`. Those metadata cases are very rare, but they violate the planner's fail-closed contract. Preserving selective handling for them required normalization, nested-field validation, and cross-field referential checks in an already substantial repository-specific policy.

The useful scope is ordinary changes to `dependencies`, `devDependencies`, `optionalDependencies`, and `peerDependencies`. Adding exact lockfile-importer attribution keeps 22 of the 40 sampled Renovate updates selective while sending 18 uncertain updates through full CI.

**Decision**

Continue the package-manifest design, but narrow lockfile explanation to concrete dependency-record changes and validate importer attribution independently. Keep metadata-only changes, lockfile-only changes, importer membership drift, unreadable manifests, unknown root tools, private packages, unsupported manifests, and version-only changes with a lockfile on the full-CI fallback.

This retains the observed safe savings while reducing schema responsibility and protecting the global deduplication workflow.

</details>

<details>
<summary>Cycle 3: Continue with narrow type-package Docs routing</summary>

**Assessment**

The original task addresses a frequent repository cost: all 40 sampled Renovate pull requests changed manifests, and the current planner keeps 22 of those updates on narrower CI while conservatively sending 18 through full CI.

The first review cycle found a material correctness gap in grouped lockfile updates: 6 of 94 historically scoped plans contained changes to importers whose manifests were untouched and could skip an affected workflow. Exact, fail-closed importer comparison was therefore proportionate. The routing-table coverage and comparison/style cleanups were small hardening changes. The second cycle added a real-lockfile regression guard and simplified the parser by removing nested-shape state, reducing rather than expanding its schema responsibility.

The implementation has nevertheless grown from 249 production and 318 test lines added in the first reviewed state to 345 production and 470 test lines in the current state. That maintenance cost is material, but it remains bounded to a conservative manifest policy, exact importer-block comparison, and focused coverage.

The cycle-3 Docs concern is technically valid: generated README signatures can depend on resolved type packages, while reviewed root dependencies and public package development dependencies can select Main without Docs. However, the likely impact is low. Main is always selected and its default test suite already verifies that generated package README content is current under the repository's React types, so type-induced drift should block the pull request. The skipped Docs workflow primarily removes automatic writeback, and Docs also runs on pushes to `main`.

The expected frequency is rare. The review replay found no reachable instance in 400 commits, and the observed `@types/react` update currently falls back to full CI because an unrelated importer changes. That fallback is incidental rather than a supported guarantee.

**Decision**

Continue the current selective planner and retain the simplified, fail-closed lockfile importer comparison. Address the cycle-3 finding only for conventional `@types/*` dependency changes by selecting Docs in the existing reviewed-root and public-package branches, with focused routing tests.

Do not add dependency-graph analysis or attempt to classify every package that ships declarations. Non-`@types/*` development dependencies that could indirectly affect inferred documentation remain intentionally outside automatic Docs routing; Main's generated-readme assertion remains the correctness backstop.

</details>

### DIFF
--- a/packages/ariakit-scripts/src/ci.test.ts
+++ b/packages/ariakit-scripts/src/ci.test.ts
@@ -1,3 +1,5 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
 import { expect, test } from "vitest";
 import {
   assertCIGate,
@@ -569,6 +571,15 @@ packages:
 
   expect(parseLockfileImporterChanges(base, head)).toEqual(["app", "website"]);
   expect(parseLockfileImporterChanges("invalid", head)).toBeUndefined();
+});
+
+test("parses the repository lockfile", () => {
+  const lockfile = readFileSync(
+    join(import.meta.dirname, "../../../pnpm-lock.yaml"),
+    "utf-8",
+  );
+
+  expect(parseLockfileImporterChanges(lockfile, lockfile)).toEqual([]);
 });
 
 test("rejects importer membership changes", () => {
--- a/packages/ariakit-scripts/src/ci.test.ts
+++ b/packages/ariakit-scripts/src/ci.test.ts
@@ -103,6 +103,21 @@ test("selects CI from reviewed root dependency updates", () => {
   expectSelectedWorkflows(plan, ["main", "perf", "docs"]);
 });
 
+test("runs Docs for reviewed root type dependencies", () => {
+  const plan = createCIPlan(["package.json", "pnpm-lock.yaml"], {
+    changedLockfileImporters: ["."],
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "package.json",
+        { devDependencies: { "@types/react": "19.2.16" } },
+        { devDependencies: { "@types/react": "19.2.17" } },
+      ),
+    ],
+  });
+
+  expectSelectedWorkflows(plan, ["main", "docs"]);
+});
+
 test("keeps unreviewed root dependency updates on full CI", () => {
   const plan = createCIPlan(["package.json", "pnpm-lock.yaml"], {
     packageJSONChanges: [
@@ -235,22 +250,38 @@ test("keeps metadata for undeclared peers on full CI", () => {
   expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
 });
 
-test("keeps public package development dependencies on Main CI", () => {
+test("keeps other public package development dependencies on Main CI", () => {
   const plan = createCIPlan(
-    ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
+    ["packages/ariakit-react-utils/package.json", "pnpm-lock.yaml"],
     {
-      changedLockfileImporters: ["packages/ariakit-react"],
+      changedLockfileImporters: ["packages/ariakit-react-utils"],
       packageJSONChanges: [
         getPackageJSONChange(
-          "packages/ariakit-react/package.json",
-          { devDependencies: { "@types/react": "19.2.16" } },
-          { devDependencies: { "@types/react": "19.2.17" } },
+          "packages/ariakit-react-utils/package.json",
+          { devDependencies: { react: "19.2.7" } },
+          { devDependencies: { react: "19.2.8" } },
         ),
       ],
     },
   );
 
   expectSelectedWorkflows(plan, ["main"]);
+});
+
+test("runs Docs for public package type dependencies", () => {
+  const file = "packages/ariakit-react-store/package.json";
+  const plan = createCIPlan([file, "pnpm-lock.yaml"], {
+    changedLockfileImporters: ["packages/ariakit-react-store"],
+    packageJSONChanges: [
+      getPackageJSONChange(
+        file,
+        { devDependencies: { "@types/use-sync-external-store": "1.4.0" } },
+        { devDependencies: { "@types/use-sync-external-store": "1.5.0" } },
+      ),
+    ],
+  });
+
+  expectSelectedWorkflows(plan, ["main", "docs"]);
 });
 
 test("keeps private package dependency updates on full CI", () => {
--- a/packages/ariakit-scripts/src/ci.test.ts
+++ b/packages/ariakit-scripts/src/ci.test.ts
@@ -3,6 +3,7 @@ import {
   assertCIGate,
   ciWorkflowNames,
   createCIPlan,
+  parseLockfileImporterChanges,
   parseChangedFiles,
   serializeCIGatePlan,
 } from "./ci.ts";
@@ -75,6 +76,7 @@ test("fails closed without dependency details and for CI infrastructure", () => 
 
 test("selects CI from reviewed root dependency updates", () => {
   const plan = createCIPlan(["package.json", "pnpm-lock.yaml"], {
+    changedLockfileImporters: ["."],
     packageJSONChanges: [
       getPackageJSONChange(
         "package.json",
@@ -115,6 +117,7 @@ test("keeps unreviewed root dependency updates on full CI", () => {
 
 test("selects CI from workspace dependency updates", () => {
   const plan = createCIPlan(["app/package.json", "pnpm-lock.yaml"], {
+    changedLockfileImporters: ["app"],
     packageJSONChanges: [
       getPackageJSONChange(
         "app/package.json",
@@ -131,6 +134,7 @@ test("selects consumer CI for public package runtime dependencies", () => {
   const plan = createCIPlan(
     ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
     {
+      changedLockfileImporters: ["packages/ariakit-react"],
       packageJSONChanges: [
         getPackageJSONChange(
           "packages/ariakit-react/package.json",
@@ -197,6 +201,7 @@ test("keeps public package development dependencies on Main CI", () => {
   const plan = createCIPlan(
     ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
     {
+      changedLockfileImporters: ["packages/ariakit-react"],
       packageJSONChanges: [
         getPackageJSONChange(
           "packages/ariakit-react/package.json",
@@ -237,6 +242,7 @@ test("unions dependency CI with ordinary file heuristics", () => {
   const plan = createCIPlan(
     ["app/package.json", "pnpm-lock.yaml", "website/src/pages/index.tsx"],
     {
+      changedLockfileImporters: ["app"],
       packageJSONChanges: [
         getPackageJSONChange(
           "app/package.json",
@@ -248,6 +254,22 @@ test("unions dependency CI with ordinary file heuristics", () => {
   );
 
   expectSelectedWorkflows(plan, ["main", "app", "perf", "plus", "og_images"]);
+});
+
+test("fails closed when the lockfile changes an unrelated importer", () => {
+  const options = {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "app/package.json",
+        { dependencies: { "@clerk/astro": "4.0.0" } },
+        { dependencies: { "@clerk/astro": "4.0.1" } },
+      ),
+    ],
+    changedLockfileImporters: ["app", "website"],
+  };
+  const plan = createCIPlan(["app/package.json", "pnpm-lock.yaml"], options);
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
 });
 
 test("falls back to full CI for package metadata changes", () => {
@@ -474,6 +496,89 @@ test("keeps both paths when git reports a rename", () => {
     "packages/old/src/index.ts",
     "packages/new/src/index.ts",
   ]);
+});
+
+test("compares exact lockfile importer blocks", () => {
+  const base = `lockfileVersion: '9.0'
+
+importers:
+
+  app:
+    dependencies:
+      astro:
+        specifier: 7.0.0
+        version: 7.0.0
+
+  website: {}
+
+packages:
+  astro@7.0.0: {}
+`;
+  const head = `lockfileVersion: '9.0'
+
+importers:
+
+  app:
+    dependencies:
+      astro:
+        specifier: 7.0.1
+        version: 7.0.1
+
+  website:
+    dependencies: {}
+
+packages:
+  astro@7.0.1: {}
+`;
+
+  expect(parseLockfileImporterChanges(base, head)).toEqual(["app", "website"]);
+  expect(parseLockfileImporterChanges("invalid", head)).toBeUndefined();
+});
+
+test("rejects importer membership changes", () => {
+  const base = `lockfileVersion: '9.0'
+
+importers:
+
+  app: {}
+
+packages: {}
+`;
+  const added = base.replace("  app: {}", "  app: {}\n\n  website: {}");
+
+  expect(parseLockfileImporterChanges(base, added)).toBeUndefined();
+  expect(parseLockfileImporterChanges(added, base)).toBeUndefined();
+});
+
+test("rejects malformed importer blocks", () => {
+  const base = `lockfileVersion: '9.0'
+
+importers:
+
+  app: {}
+
+packages: {}
+`;
+  const malformedInline = base.replace("  app: {}", "  app: []");
+  const malformedIndentation = base.replace("  app: {}", "   app: {}");
+  const malformedBlockSequence = base.replace("  app: {}", "  app:\n    []");
+  const malformedBlockScalar = base.replace("  app: {}", "  app:\n    value");
+  const commentsOnlyBlock = base.replace(
+    "  app: {}",
+    "  app:\n    # no importer data",
+  );
+
+  expect(parseLockfileImporterChanges(base, malformedInline)).toBeUndefined();
+  expect(
+    parseLockfileImporterChanges(base, malformedIndentation),
+  ).toBeUndefined();
+  expect(
+    parseLockfileImporterChanges(base, malformedBlockSequence),
+  ).toBeUndefined();
+  expect(
+    parseLockfileImporterChanges(base, malformedBlockScalar),
+  ).toBeUndefined();
+  expect(parseLockfileImporterChanges(base, commentsOnlyBlock)).toBeUndefined();
 });
 
 test("serializes a compact gate plan for large changes", () => {
--- a/packages/ariakit-scripts/src/ci.test.ts
+++ b/packages/ariakit-scripts/src/ci.test.ts
@@ -6,7 +6,7 @@ import {
   parseChangedFiles,
   serializeCIGatePlan,
 } from "./ci.ts";
-import type { CIGateResult } from "./ci.ts";
+import type { CIGateResult, CIWorkflowName, PackageJSONChange } from "./ci.ts";
 
 function getResults(
   plan: ReturnType<typeof createCIPlan>,
@@ -21,6 +21,23 @@ function getResults(
     };
   }
   return { ...results, ...overrides };
+}
+
+function getPackageJSONChange(
+  file: string,
+  base: Record<string, unknown>,
+  head: Record<string, unknown>,
+): PackageJSONChange {
+  return { file, base, head };
+}
+
+function expectSelectedWorkflows(
+  plan: ReturnType<typeof createCIPlan>,
+  workflows: CIWorkflowName[],
+) {
+  expect(
+    ciWorkflowNames.filter((workflow) => plan.workflows[workflow]),
+  ).toEqual(workflows);
 }
 
 test("keeps core and legacy browser tests on every pull request", () => {
@@ -38,7 +55,7 @@ test("keeps core and legacy browser tests on every pull request", () => {
   });
 });
 
-test("runs every workflow for dependency and CI infrastructure changes", () => {
+test("fails closed without dependency details and for CI infrastructure", () => {
   for (const file of [
     "pnpm-lock.yaml",
     "app/package.json",
@@ -54,6 +71,305 @@ test("runs every workflow for dependency and CI infrastructure changes", () => {
     const plan = createCIPlan([file]);
     expect(Object.values(plan.workflows).every(Boolean), file).toBe(true);
   }
+});
+
+test("selects CI from reviewed root dependency updates", () => {
+  const plan = createCIPlan(["package.json", "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "package.json",
+        {
+          devDependencies: {
+            "lint-staged": "17.1.1",
+            oxfmt: "0.59.0",
+            vitest: "4.1.9",
+          },
+        },
+        {
+          devDependencies: {
+            "lint-staged": "17.2.0",
+            oxfmt: "0.60.0",
+            vitest: "4.1.10",
+          },
+        },
+      ),
+    ],
+  });
+
+  expectSelectedWorkflows(plan, ["main", "perf", "docs"]);
+});
+
+test("keeps unreviewed root dependency updates on full CI", () => {
+  const plan = createCIPlan(["package.json", "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "package.json",
+        { devDependencies: { vite: "8.1.4" } },
+        { devDependencies: { vite: "8.1.5" } },
+      ),
+    ],
+  });
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("selects CI from workspace dependency updates", () => {
+  const plan = createCIPlan(["app/package.json", "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "app/package.json",
+        { dependencies: { "@clerk/astro": "4.0.0" } },
+        { dependencies: { "@clerk/astro": "4.0.1" } },
+      ),
+    ],
+  });
+
+  expectSelectedWorkflows(plan, ["main", "app", "perf", "og_images"]);
+});
+
+test("selects consumer CI for public package runtime dependencies", () => {
+  const plan = createCIPlan(
+    ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-react/package.json",
+          { peerDependencies: { react: ">=18" } },
+          { peerDependencies: { react: ">=19" } },
+        ),
+      ],
+    },
+  );
+
+  expectSelectedWorkflows(plan, [
+    "main",
+    "app",
+    "perf",
+    "plus",
+    "release_preview",
+    "docs",
+    "og_images",
+  ]);
+});
+
+test("keeps public peer dependency metadata changes on full CI", () => {
+  const plan = createCIPlan(
+    ["packages/ariakit-test/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-test/package.json",
+          {
+            peerDependencies: { react: ">=18" },
+            peerDependenciesMeta: { react: { optional: true } },
+          },
+          {
+            peerDependencies: { react: ">=18" },
+            peerDependenciesMeta: { react: { optional: false } },
+          },
+        ),
+      ],
+    },
+  );
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("keeps metadata for undeclared peers on full CI", () => {
+  const file = "packages/ariakit-test/package.json";
+  const plan = createCIPlan([file, "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        file,
+        { peerDependencies: {} },
+        {
+          peerDependencies: {},
+          peerDependenciesMeta: { ghost: { optional: true } },
+        },
+      ),
+    ],
+  });
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("keeps public package development dependencies on Main CI", () => {
+  const plan = createCIPlan(
+    ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-react/package.json",
+          { devDependencies: { "@types/react": "19.2.16" } },
+          { devDependencies: { "@types/react": "19.2.17" } },
+        ),
+      ],
+    },
+  );
+
+  expectSelectedWorkflows(plan, ["main"]);
+});
+
+test("keeps private package dependency updates on full CI", () => {
+  const plan = createCIPlan(
+    ["packages/ariakit-scripts/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-scripts/package.json",
+          {
+            private: true,
+            dependencies: { "rolldown-plugin-dts": "0.27.12" },
+          },
+          {
+            private: true,
+            dependencies: { "rolldown-plugin-dts": "0.27.13" },
+          },
+        ),
+      ],
+    },
+  );
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("unions dependency CI with ordinary file heuristics", () => {
+  const plan = createCIPlan(
+    ["app/package.json", "pnpm-lock.yaml", "website/src/pages/index.tsx"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "app/package.json",
+          { dependencies: { "@clerk/astro": "4.0.0" } },
+          { dependencies: { "@clerk/astro": "4.0.1" } },
+        ),
+      ],
+    },
+  );
+
+  expectSelectedWorkflows(plan, ["main", "app", "perf", "plus", "og_images"]);
+});
+
+test("falls back to full CI for package metadata changes", () => {
+  const plan = createCIPlan(["app/package.json", "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(
+        "app/package.json",
+        { scripts: { build: "astro build" } },
+        { scripts: { build: "astro check && astro build" } },
+      ),
+    ],
+  });
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("does not scope lockfile changes without a dependency update", () => {
+  const plan = createCIPlan(["app/package.json", "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange("app/package.json", {}, { devDependencies: {} }),
+    ],
+  });
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("does not scope lockfile changes with empty peer metadata", () => {
+  const file = "packages/ariakit-test/package.json";
+  const plan = createCIPlan([file, "pnpm-lock.yaml"], {
+    packageJSONChanges: [
+      getPackageJSONChange(file, {}, { peerDependenciesMeta: {} }),
+    ],
+  });
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("does not use package versions to scope lockfile changes", () => {
+  const plan = createCIPlan(
+    ["packages/ariakit-react/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-react/package.json",
+          { version: "0.4.35" },
+          { version: "0.4.36" },
+        ),
+      ],
+    },
+  );
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("requires every changed manifest to scope the lockfile", () => {
+  const plan = createCIPlan(
+    [
+      "app/package.json",
+      "packages/ariakit-react/package.json",
+      "pnpm-lock.yaml",
+    ],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "app/package.json",
+          { dependencies: { "@clerk/astro": "4.0.0" } },
+          { dependencies: { "@clerk/astro": "4.0.1" } },
+        ),
+        getPackageJSONChange(
+          "packages/ariakit-react/package.json",
+          { version: "0.4.35" },
+          { version: "0.4.36" },
+        ),
+      ],
+    },
+  );
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("rejects empty dependency metadata in grouped lockfile updates", () => {
+  const plan = createCIPlan(
+    ["app/package.json", "examples/package.json", "pnpm-lock.yaml"],
+    {
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "app/package.json",
+          { dependencies: { "@clerk/astro": "4.0.0" } },
+          { dependencies: { "@clerk/astro": "4.0.1" } },
+        ),
+        getPackageJSONChange(
+          "examples/package.json",
+          {},
+          { devDependencies: {} },
+        ),
+      ],
+    },
+  );
+
+  expect(Object.values(plan.workflows).every(Boolean)).toBe(true);
+});
+
+test("avoids full CI for generated publish metadata", () => {
+  const plan = createCIPlan(
+    [
+      ".changeset/example.md",
+      "packages/ariakit-react/CHANGELOG.md",
+      "packages/ariakit-react/package.json",
+    ],
+    {
+      baseRef: "main",
+      packageJSONChanges: [
+        getPackageJSONChange(
+          "packages/ariakit-react/package.json",
+          { version: "0.4.35" },
+          { version: "0.4.36" },
+        ),
+      ],
+    },
+  );
+
+  expectSelectedWorkflows(plan, ["main", "release_preview"]);
 });
 
 test("runs app, performance, docs, release, and website checks for package code", () => {
--- a/packages/ariakit-scripts/src/ci.test.ts
+++ b/packages/ariakit-scripts/src/ci.test.ts
@@ -34,7 +34,7 @@ function getPackageJSONChange(
 
 function expectSelectedWorkflows(
   plan: ReturnType<typeof createCIPlan>,
-  workflows: CIWorkflowName[],
+  workflows: readonly CIWorkflowName[],
 ) {
   expect(
     ciWorkflowNames.filter((workflow) => plan.workflows[workflow]),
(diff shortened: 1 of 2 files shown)


packages/ariakit-scripts/src/ci.ts
```diff
@@ -227,6 +425,13 @@ export function createCIPlan(
 
   addReason(plan, "main", "Core and legacy browser tests run on every PR");
   for (const file of files) {
+    // The manifest diff is the policy input. A lockfile-only change still
+    // falls through to full CI because no manifest explains its scope.
+    if (file === "pnpm-lock.yaml" && hasScopedPackageJSONChanges) continue;
```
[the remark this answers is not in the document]
diegohaz: Confirmed against the current head, with historical evidence.

I replayed the last 400 commits on `main` through `createCIPlan` from this branch, rebuilding `packageJSONChanges` from the real merge-base/head manifests, and separately diffed the lockfile's per-workspace `importers:` blocks. Of 154 commits that touched `pnpm-lock.yaml`, 94 produce a scoped (non-full) plan, and **6 of those 94 skip a workflow whose workspace importer was re-resolved by the very lockfile change this line ignores**: `433c6ce73` (#6859), `844f129c4` (#6798), `99a5e987c` (#6695), `5ef75f0c4` (#6561), `941eb0698` (#6558), `d3eae772d` (#6516).

The cleanest case is `99a5e987c` ("Update dependency @types/node to v24.13.3", #6695). Changed manifests are `package.json` and `app/package.json`, both of which `explainsLockfileChange` accepts, so the lockfile is skipped and the plan is `main, app, perf, og_images`. But the lockfile also moved the `templates/react` importer, whose manifest did not change:

```diff
       '@vitejs/plugin-react':
         specifier: 4.3.2
-        version: 4.3.2(vite@8.1.3(@types/node@24.13.2)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.46.0)(yaml@2.9.0))
+        version: 4.3.2(vite@8.1.3(@types/node@24.13.3)(esbuild@0.28.1)(jiti@2.7.0)(terser@5.46.0)(yaml@2.9.0))
```

`release_preview` was skipped. `.github/workflows/release-preview.yml` is the only workflow that exercises `templates/*`:

```
pnpm exec pkg-pr-new publish --pnpm --compact --template './templates/*' $packages
```

and it has no `push` trigger, only `workflow_call`, so nothing re-runs it after merge. Same for `perf`.

`5ef75f0c4` ("Update dependency astro to v7.0.2", #6561) shows the same mechanism reaching the website. Only `app/package.json` changed, the plan is `main, app, perf, og_images`, and the lockfile re-resolved the `website` importer:

```diff
       '@babel/preset-env':
         specifier: 7.29.2
-        version: 7.29.2(@babel/core@7.29.0)(supports-color@8.1.1)
+        version: 7.29.2(@babel/core@7.29.0)
       react-markdown:
         specifier: 8.0.7
-        version: 8.0.7(@types/react@18.3.28)(react@18.3.1)(supports-color@8.1.1)
+        version: 8.0.7(@types/react@18.3.28)(react@18.3.1)
```

`plus` is what builds (`pnpm run build-website-lite`) and Playwright-tests (`xvfb-run pnpm run test-plus-headed`) that workspace, and it was skipped. The same commit also re-resolved the root importer (`open-cli`, `stylelint`, `stylelint-config-standard`, `tsup`) while `docs` and `release_preview` were skipped.

Two honest caveats. In the `website` cases the delta is a `supports-color` peer-key flip at identical versions, so no break is proven, only that the planner cannot tell the difference. And `main`, `app`, `plus`, `docs`, `og-images` and `build-styles` all have `push: [main]` triggers, so for those a miss surfaces post-merge instead of blocking the PR; `perf` and `release-preview` have no such net.

**Assessment**

- Severity: Medium (estimate, 75% confidence), because the skipped workflows still run post-merge for everything except `perf` and `release_preview`, so the usual outcome is a break landing on `main` rather than an undetected one.
- Priority: P1 (estimate, 70% confidence), because the PR description states the planner fails closed "whenever a lockfile change is not fully explained by the changed manifests" and the implementation does not verify that; a follow-up is defensible given the post-merge triggers, but the stated contract should not ship unmet.
- Breakage likelihood: Possible (verified by the 400-commit replay), because 6 of 94 scoped plans (~6%) already exhibit the gap, all from ordinary Renovate updates combined with the repository-wide `pnpmDedupe` post-update option.
- Fix effort: Moderate (estimate, 75% confidence), because it needs a small `importers:` reader plus an importer-directory-to-workflow mapping, and `workspaceDependencyWorkflows` already supplies most of that mapping.

Two directions that keep the current design:

1. Read only the lockfile's `importers:` section, split it by workspace, and require every changed importer to correspond to a changed manifest; fall back to full CI otherwise. This implements the stated contract directly and never has to understand `packages:` or `snapshots:`.
2. Cheaper and still safe: instead of demanding attribution, union each changed importer's workflows into the plan, reusing `workspaceDependencyWorkflows` keyed by importer directory (`.` for root, `packages/*` for public packages), and fall back to full CI for an importer that cannot be scoped.

Either way 88 of the 94 scoped plans are unaffected, so the savings this PR reports are essentially preserved.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -63,6 +71,60 @@ const dependencyAndConfigNames = new Set([
...
+const workspaceDependencyWorkflows = new Map<string, readonly CIWorkflowName[]>(
+  [
+    ["app/package.json", ["app", "perf", "og_images"]],
+    ["examples/package.json", []],
+    ["guide/package.json", ["plus"]],
+    ["nextjs/package.json", ["app", "perf"]],
+    ["templates/react/package.json", ["release_preview"]],
+    ["website/package.json", ["plus"]],
+  ],
+);
```
diegohaz, line 132: Only one of this table's six entries is covered by a test. `app/package.json` is asserted in "selects CI from workspace dependency updates" (`main, app, perf, og_images`), but `guide`, `nextjs`, `templates/react`, `website` and the deliberately empty `examples` mapping are never asserted. `examples/package.json` appears in "rejects empty dependency metadata in grouped lockfile updates", but only as the manifest that forces the whole plan back to full CI, so its `[]` value is never exercised.

The table is the part of this policy that is hand-derived from the workflow definitions, and TypeScript only checks that the values are valid `CIWorkflowName`s, not that the set is right. A future edit that drops `perf` from `nextjs`, or `plus` from `website`, or that adds a seventh workspace without checking `.github/workflows/`, passes `pnpm tsc` and the whole suite while silently skipping a gate.

There is also no test for the fail-closed side of the table: an unrecognized workspace manifest such as `foo/package.json` with a real dependency change should fall through `addPackageJSONReasons` to `addFileReasons` and run full CI, and nothing locks that in.

Real-world impact: Renovate opens PRs that touch `nextjs/package.json` and nothing else besides the lockfile, for example `08687a082` ("Update dependency @opennextjs/cloudflare to v1.20.1", #6576). Under this table that PR selects `main, app, perf`. If the `nextjs` row later loses `app`, the same PR shape selects `main, perf`, `pnpm -F nextjs run tsc` and `pnpm -F nextjs run build-worker` never run on the PR, and a Next.js build break reaches `main`.

A table-driven test over all six entries plus one unknown workspace would cover this in a few lines and would double as the documentation for how each row was derived.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because an incorrect row silently disables a CI gate rather than failing loudly.
- Priority: P2 (estimate, 75% confidence), because the table is correct as written today; the gap is regression protection for future edits.
- Breakage likelihood: Possible (estimate, 70% confidence), because the table has to be revisited whenever a workspace is added or a workflow changes which workspaces it builds.
- Fix effort: Easy (estimate, 90% confidence), because it is a loop over the table entries in `ci.test.ts` next to the existing `expectSelectedWorkflows` helper.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -108,6 +170,128 @@ function isAppRuntimePath(file: string) {
...
+function hasRuntimeDependencyChange(
+  base: Record<string, unknown>,
+  head: Record<string, unknown>,
+) {
+  return (
+    !isDeepStrictEqual(base.dependencies, head.dependencies) ||
+    !isDeepStrictEqual(base.optionalDependencies, head.optionalDependencies) ||
+    !isDeepStrictEqual(base.peerDependencies, head.peerDependencies)
+  );
+}
```
diegohaz: This helper redoes deep comparisons the caller already has. By the time `addPackageJSONReasons` reaches its `hasRuntimeDependencyChange` call, `fields` is the output of `getChangedFields`, which already ran `isDeepStrictEqual` over the union of both key sets, and `hasOnlyFields(fields, publicPackageFields)` has already narrowed `fields` to a subset of the four dependency fields plus `version`. Asking the same question a second time from the raw records adds three more deep comparisons and a second, differently shaped source of truth for "did this field change".

The set membership check is equivalent: `getChangedFields` marks a field changed exactly when `!isDeepStrictEqual(base[field], head[field])`, which is the same predicate this helper applies to each of the three runtime fields.

```diff
-function hasRuntimeDependencyChange(
-  base: Record<string, unknown>,
-  head: Record<string, unknown>,
-) {
-  return (
-    !isDeepStrictEqual(base.dependencies, head.dependencies) ||
-    !isDeepStrictEqual(base.optionalDependencies, head.optionalDependencies) ||
-    !isDeepStrictEqual(base.peerDependencies, head.peerDependencies)
-  );
-}
+const runtimeDependencyFields = new Set([
+  "dependencies",
+  "optionalDependencies",
+  "peerDependencies",
+]);
+
+function hasRuntimeDependencyChange(fields: string[]) {
+  return fields.some((field) => runtimeDependencyFields.has(field));
+}
```

with the call site becoming `if (hasRuntimeDependencyChange(fields)) {`. I applied this locally and `pnpm test packages/ariakit-scripts/src/ci.test.ts --run` stays at 32 passing.

**Assessment**

- Severity: Low (verified), because behavior is unchanged.
- Priority: P2 (estimate, 85% confidence), because it is a readability and single-source-of-truth cleanup, not a defect.
- Breakage likelihood: Rare (estimate, 85% confidence), because the divergence only matters if the two field-comparison paths are edited apart later.
- Fix effort: Easy (verified), because the replacement above is the whole change.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -108,6 +170,128 @@ function isAppRuntimePath(file: string) {
...
+    addDependencyReason(plan, "main", file);
+    for (const workflow of workspaceWorkflows) {
+      addDependencyReason(plan, workflow, file);
+    }
+    return true;
+  }
+
+  if (!/^packages\/[^/]+\/package\.json$/.test(file)) return false;
+  if (change.base.private === true || change.head.private === true)
+    return false;
```
diegohaz: Two independent early exits are combined with `||`, and because the result does not fit on one line the body ends up unbraced across two lines. The repository code style asks for one `if` per line for multiple early exits, and for braces whenever the single-line early-exit form is not used. Splitting them satisfies both rules and each half fits on one line:

```diff
-  if (change.base.private === true || change.head.private === true)
-    return false;
+  if (change.base.private === true) return false;
+  if (change.head.private === true) return false;
```

This is the only occurrence of the unbraced multi-line form introduced here; the rest of the file already follows the single-line early-exit convention.

diegohaz (commented):
## Review summary

Initial review. Baseline: merge base `c68704a8c` to head `445908af1`, i.e. the full PR diff (`packages/ariakit-scripts/src/ci.ts`, `packages/ariakit-scripts/src/ci.test.ts`), plus the surrounding context that the new policy depends on: every workflow under `.github/workflows/`, `renovate.json`, the root and workspace manifests, `app/src/lib/build-styles.ts`, and `packages/ariakit-scripts/src/docs.ts`.

### What I ran

- `pnpm test packages/ariakit-scripts/src/ci.test.ts --run` — 32 passed, matching the description.
- `pnpm lint` — formatting clean; the only oxlint warnings are pre-existing in `examples/menu-slide/test-chrome-safari.ts`.
- A replay harness over the last 400 commits on `main`: for every commit that touched `pnpm-lock.yaml`, it rebuilds `packageJSONChanges` from the real merge-base/head manifests, runs `createCIPlan` from this branch, and independently diffs the lockfile's per-workspace `importers:` blocks. 154 lockfile commits, 94 scoped plans, 6 with a coverage gap.
- A local prototype of the `hasRuntimeDependencyChange` simplification (32 tests still passing), reverted afterwards. No experiment changes remain.

- `packages/ariakit-scripts/src/ci.ts:430`, Codex, "Verify the lockfile scope before skipping its workflows" (`PRRT_kwDOBjT0Fc6UCED7`). **Confirmed**, disposition Address now. `hasScopedPackageJSONChanges` only checks that each changed manifest contains some concrete dependency-name change; it never checks that the lockfile delta is confined to those manifests' importers, so `pnpmDedupe` and transitive re-resolution can move an unrelated workspace's dependency graph while its workflow is skipped. I replied on the thread with the historical evidence and two fix directions rather than opening a duplicate.

No other unresolved threads existed on this PR.

### Published inline

Three new comments, all on `packages/ariakit-scripts/src/ci.ts`: missing coverage for the `workspaceDependencyWorkflows` table, the redundant deep comparisons in `hasRuntimeDependencyChange`, and one code-style deviation on the `private` guard.

### Verified safe (the risky narrowings)

These are the places where the PR narrows CI the most, so I checked each rather than assuming:

- `app/package.json` no longer selects `build_styles`. Safe: `app/src/lib/build-styles.ts` imports only `node:fs/promises`, `node:path`, `./styles-shared.ts` and `./styles.ts`, and reads `app/src/styles/`. No `app` dependency participates in `pnpm -F app run build-styles`.
- Public package **devDependency** changes select `main` only. Acceptable: `main.yml` runs `pnpm run lint`, `pnpm run tsc -v`, `pnpm run build` (root `pnpm -F "@ariakit/*" run build`) and `pnpm run build-website-lite`, so package build and type breaks are still gated.
- Public package **version-only** changes select `main`. The `main, release_preview` outcome the description claims for Publish PRs comes from the deleted `.changeset/*.md` files, not from the manifest rule; verified against the "avoids full CI for generated publish metadata" test.
- Root `oxfmt` also selecting `docs` and root `vitest` also selecting `perf` are justified: `packages/ariakit-scripts/src/docs.ts:789` spawns `oxfmt`, and `.github/workflows/perf.yml` runs `pnpm exec vitest bench`.
- `examples/package.json` mapping to no extra workflow matches the existing `safelyCoveredByMain` treatment of `examples/`, and `guide/package.json` currently declares no dependencies at all.
- Fail-closed paths hold: added, deleted and renamed manifests are dropped by `getPackageJSONChanges` (one side is unreadable), invalid or non-object JSON is dropped, non-string dependency values make `getChangedDependencyNames` return `undefined`, and in every case `hasScopedPackageJSONChanges` becomes `false` so `pnpm-lock.yaml` falls through to full CI.
- `packages/ariakit-scripts/package.json` correctly stays on full CI, but only because `@ariakit/scripts` is `private`. Worth remembering that the manifest branch runs before `addFileReasons`, so it bypasses the `packages/ariakit-scripts/` entry in `isFullCIPath`; if that package ever became public, the planner would start scoping changes to its own source of truth.

### Candidates first discovered here and not published inline

- **`getPackageJSONChanges` has no test coverage.** Disposition: Declined non-goal. It shells out to `git merge-base` and `git show`, and `getChangedFiles` sets the existing precedent for leaving that layer untested while testing its pure parser (`parseChangedFiles`). The behaviors that matter for fail-closed correctness are exercised through `createCIPlan`.
- **`/(?:^|\/)package\.json$/` is duplicated** in `createCIPlan` and `getPackageJSONChanges`, and `addPackageJSONReasons` re-runs `normalizeCIPath` on a path the caller already normalized into the map key. Disposition: Declined non-goal, preferential cleanup with no behavioral effect.
- **`getChangedDependencyNames` runs twice per manifest**, once through `explainsLockfileChange` and once inside `addPackageJSONReasons`. Disposition: Declined non-goal; manifests per PR are single digits and the work is a shallow object walk.
- **Nothing records why `pnpm-lock.yaml` was ignored.** The plan log prints per-workflow reasons, so a scoped dependency PR shows `Dependency metadata: <file>` but never states that the lockfile was deliberately skipped, which makes a wrongly-skipped workflow harder to diagnose from the `plan` job output alone. Disposition: Declined non-goal for this PR; it becomes moot if the lockfile-scope thread is addressed, since that work would need its own reasons anyway.
- **Public package devDependency bumps can in principle drift generated docs.** `packages/ariakit-scripts/src/docs.ts:536` uses the ts-morph type checker (`declaration.getType().getText(declaration)`), so a bump of something like `@types/use-sync-external-store` in `packages/ariakit-react-store/package.json` could change the generated readme while the plan selects `main` only. Disposition: Declined non-goal, because `docs.yml`, `og-images.yml` and `build-styles.yml` all run on `push: [main]` and write back, so generated-content drift self-heals after merge. I did not construct a bump that actually changes readme output, so this is a suspected effect rather than a confirmed one.

No follow-up issues were registered, because every confirmed finding is either published inline or on the existing thread.

### Not fully validated

- I did not run `pnpm tsc` or `pnpm build`. The change is confined to a private script package plus its test, and `pnpm lint` and the focused suite passed.
- I did not reproduce the description's "latest 40 Renovate pull requests, 25 narrower, 15 full CI" figure directly. My independent 400-commit replay produced a comparable 94-of-154 scoped rate.
- The replay's importer-to-workflow mapping is my own approximation derived from the workflow definitions, so a "coverage gap" means the lockfile moved a workspace's resolved graph while a workflow that builds or tests that workspace was skipped. It does not prove a break occurred in any specific past PR.

packages/ariakit-scripts/src/ci.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`242601353`](https://github.com/ariakit/ariakit/commit/242601353): the planner now compares canonical merge-base/head lockfile importer blocks and falls back to full CI for unrelated, missing, malformed, added, or removed importer evidence. The historical `@types/node` update from #6695 now selects full CI.

packages/ariakit-scripts/src/ci.ts
diegohaz: Fixed in [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): a table-driven test now covers all six workspace mappings, and an unknown workspace dependency update is explicitly verified to fall back to full CI.

packages/ariakit-scripts/src/ci.ts
diegohaz: Fixed in [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): runtime dependency routing now checks the already-computed changed field names, so the duplicate deep comparisons and second source of truth are gone.

packages/ariakit-scripts/src/ci.ts
diegohaz: Fixed in [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): the combined multiline private-package condition is now two single-line early exits, matching the repository style.

diegohaz:
## Review updates

**Fixed**

- [`242601353`](https://github.com/ariakit/ariakit/commit/242601353): validate merge-base/head lockfile importer scope and fail closed for unrelated or invalid evidence ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657121815)).
- [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): cover all six workspace dependency mappings plus the unknown-workspace fallback ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657124027)).
- [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): reuse the computed changed fields for runtime dependency routing ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657127059)).
- [`d053b8770`](https://github.com/ariakit/ariakit/commit/d053b8770): split the private-package check into repository-style early exits ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657130415)).

Verification: 42 planner tests, full lint, targeted type-aware Oxlint/Oxfmt, and the repository typecheck pass.

packages/ariakit-scripts/src/ci.test.ts
```diff
@@ -160,6 +534,89 @@ test("keeps both paths when git reports a rename", () => {
...
+
+  website:
+    dependencies: {}
+
+packages:
+  astro@7.0.1: {}
+`;
+
+  expect(parseLockfileImporterChanges(base, head)).toEqual(["app", "website"]);
+  expect(parseLockfileImporterChanges("invalid", head)).toBeUndefined();
```
diegohaz, line 604: Every fixture in these three tests is hand-written YAML that avoids the constructs the real `pnpm-lock.yaml` is made of. The keys are `app`, `website`, `dependencies`, `astro`, `specifier`, `version`, and the values are `7.0.0` and `7.0.1`. The actual lockfile is full of quoted scoped keys (`'@ariakit/react':`), peer-suffixed versions (`4.1.10(@types/node@24.13.3)(happy-dom@20.11.1)(...)`), `link:../packages/ariakit-react`, and `workspace:*` specifiers. None of that is covered.

That matters more than usual here because of the failure mode. If `parseLockfileImporters` ever stops accepting the real lockfile, `parseLockfileImporterChanges` returns `undefined`, `changedLockfileImporters` becomes `[]`, `hasScopedLockfileChanges` is `false`, and every dependency pull request quietly falls back to full CI. Nothing throws, no test fails, and `runCIPlan` prints no reason mentioning the lockfile, so the whole feature is dead and the only symptom is that CI got slower again.

I verified that this is reachable, not theoretical. Tightening one token in the content regex to a plausible-looking stricter form:

```diff
-    const content = /^( {4}(?: {2})*)([^#\s][^:]*):(.*)$/.exec(line);
+    const content = /^( {4}(?: {2})*)([a-zA-Z0-9._-]+):(.*)$/.exec(line);
```

leaves all 42 tests passing, while `parseLockfileImporterChanges(lock, lock)` against the repository's own `pnpm-lock.yaml` returns `undefined` instead of `[]`. Quoted scoped keys are the first thing it rejects.

The most durable guard is to assert that the committed lockfile parses, since it stays true as the lockfile churns:

```ts
test("parses the repository lockfile", () => {
  const lockfile = readFileSync(
    join(import.meta.dirname, "../../../pnpm-lock.yaml"),
    "utf-8",
  );
  expect(parseLockfileImporterChanges(lockfile, lockfile)).toEqual([]);
});
```

`dev.test.ts` and `docs.test.ts` already read repository paths from this suite, and `build.test.ts` already uses `import.meta.dirname`, so this fits existing practice. I ran it: it passes on the current head and fails with `expected undefined to deeply equal []` under the tightened regex above.

If reading a repository file from this suite is unwanted, the equivalent is an inline fixture whose importer block uses a quoted scoped key, a peer-suffixed `version`, a `link:` value and a `workspace:*` specifier, plus a paired assertion that perturbing one leaf `version` line reports exactly that importer.

**Assessment**

- Severity: High (verified), because the failure silently disables the entire selective-CI feature this PR exists to add, with no failing test and no log line.
- Priority: P1 (estimate, 80% confidence), because the parser is new, hand-rolled, and the only thing standing between the plan and the real lockfile.
- Breakage likelihood: Possible (estimate, 70% confidence), because the regexes are the natural place for a later "tidy up" or a pnpm lockfile format change to land, and both would look safe against the current fixtures.
- Fix effort: Easy (verified), because the test above is five lines and I confirmed it fails against the tightened regex and passes against the current one.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -254,6 +469,159 @@ export function getChangedFiles(base: string, head: string) {
...
+  const importersIndex = lines.indexOf("importers:");
+  if (importersIndex < 0) return;
+
+  const importers = new Map<string, string>();
+  let currentImporter: string | undefined;
+  let currentBlock: string[] = [];
+  let currentImporterIsEmpty = false;
+  let currentImporterHasContent = false;
+  let currentFieldNeedsContent = false;
+  let currentFieldHasContent = false;
```
diegohaz: Four of this parser's six pieces of mutable state (`currentImporterIsEmpty`, `currentImporterHasContent`, `currentFieldNeedsContent`, `currentFieldHasContent`, plus `saveCurrentField`) exist to validate the *shape* of an importer block: that a field header is followed by content, that a non-`{}` importer is non-empty, that nothing follows a `{}` importer. None of that is load-bearing.

What the caller needs is only that importer blocks are split at the right boundaries and compared exactly. Boundaries are established by the two-space key line, and they are protected by the two rejection rules that *are* load-bearing: an importer key line whose value is neither empty nor ` {}` is rejected, and every other non-blank line must be indented four or more spaces and look like `key:`. Anything ambiguous already returns `undefined` before the shape flags are consulted. Once the block is split, `parseLockfileImporterChanges` compares the raw text, so a structurally odd but unambiguously delimited block is still compared correctly. Rejecting it buys nothing.

This also runs against the direction the PR's own "Cycle 3: Narrow lockfile explanation to concrete dependency records" reassessment set, which chose to reduce the planner's schema responsibility rather than keep validating nested shapes. The shape flags reintroduce exactly that responsibility, now for YAML instead of `peerDependenciesMeta`.

The equivalent parser keeps both rejection rules and drops the shape validation:

```ts
function parseLockfileImporters(value: string) {
  const lines = value.replaceAll("\r\n", "\n").split("\n");
  const importersIndex = lines.indexOf("importers:");
  if (importersIndex < 0) return;

  const importers = new Map<string, string>();
  let currentImporter: string | undefined;
  let currentBlock: string[] = [];
  const saveCurrentImporter = () => {
    if (!currentImporter) return true;
    if (importers.has(currentImporter)) return false;
    importers.set(currentImporter, currentBlock.join("\n").trimEnd());
    return true;
  };

  for (let index = importersIndex + 1; index < lines.length; index++) {
    const line = lines[index];
    if (line === undefined) return;
    if (line && !line.startsWith(" ")) {
      if (!/^[^:\s][^:]*:/.test(line)) return;
      break;
    }
    const match = /^  ([a-zA-Z0-9._@/-]+):(?: \{\})?$/.exec(line);
    if (match) {
      if (!saveCurrentImporter()) return;
      currentImporter = match[1];
      currentBlock = [line];
      continue;
    }
    if (!line) {
      if (currentImporter) currentBlock.push(line);
      continue;
    }
    if (!currentImporter) return;
    if (!/^ {4}(?: {2})*[^#\s][^:]*:/.test(line)) return;
    currentBlock.push(line);
  }
  if (!saveCurrentImporter()) return;
  if (importers.size === 0) return;
  return importers;
}
```

I ran this locally against two things. All 42 tests pass unchanged, including every case in "rejects malformed importer blocks" (`app: []`, three-space indentation, `    []`, `    value`, and the comment-only block are all still rejected, by the two remaining rules). And replaying the 154 commits on `main` that touched `pnpm-lock.yaml` through `createCIPlan` produces identical output: the same 86 scoped plans, and the same six commits (#6859, #6798, #6695, #6561, #6558, #6516) correctly falling back to full CI.

Worth noting separately: no current test exercises the four flags at all. The one fixture that reaches a field header uses `dependencies: {}`, which sets `currentFieldNeedsContent = false`, so the `saveCurrentField` path is never the reason a fixture is rejected.

**Assessment**

- Severity: Low (verified), because behavior is identical on every input I could construct or find in the repository's history.
- Priority: P2 (estimate, 80% confidence), because it is maintainability of new hand-rolled parsing rather than a defect; the smaller the parser, the less there is to get wrong when pnpm changes its lockfile format.
- Breakage likelihood: Rare (estimate, 80% confidence), because the extra state cannot produce a wrong answer, only an unnecessary full-CI fallback on input that pnpm does not emit.
- Fix effort: Easy (verified), because the replacement above is the whole change and needs no test updates.

diegohaz (commented):
## Re-review

Re-review. Baseline is the head of the previous review, `445908af1`, so the incremental diff is `242601353` ("Validate lockfile importer scope") and `d053b8770` ("Cover dependency workflow routing"). I also re-checked the surrounding context those commits newly depend on: the `importers:` layout of the committed `pnpm-lock.yaml`, `playwright.config.ts`, and the `push` triggers of every workflow under `.github/workflows/`.

### Previous findings

All four threads from the last review are resolved, and I verified each fix against the code and against behavior rather than the reply:

- **Lockfile scope.** Fixed in `242601353`. Re-ran the 400-commit replay with the new planner: all six previously-gapped commits (#6859, #6798, #6695, #6561, #6558, #6516) now fall back to full CI. The cost is 8 plans: 94 scoped plans before, 86 after, out of 154 commits that touched the lockfile. That is a well-calibrated trade.
- **`workspaceDependencyWorkflows` coverage.** Fixed in `d053b8770`. The `test.each` table covers all six rows plus an unknown workspace falling back to full CI. I re-derived each expectation from the workflow files and they match: `guide` and `website` map to `plus` (`pnpm run build-website-lite` plus `xvfb-run pnpm run test-plus-headed`), `nextjs` to `app` and `perf` (`pnpm -F nextjs run tsc`, `pnpm -F nextjs run build-worker` in both), `templates/react` to `release_preview` (`pkg-pr-new --template './templates/*'`), and `examples` to `main` only.
- **`hasRuntimeDependencyChange`.** Fixed in `d053b8770`, now `fields.some((field) => runtimeDependencyFieldSet.has(field))` over the already-computed field list.
- **`private` guard style.** Fixed in `d053b8770`, now two single-line early exits.

### What I ran

- `pnpm test packages/ariakit-scripts/src/ci.test.ts --run` — 42 passed.
- `pnpm lint` — formatting clean; the only oxlint warnings remain the pre-existing ones in `examples/menu-slide/test-chrome-safari.ts`.
- The 400-commit replay against the new planner. The parser accepted 154 of 154 real lockfile pairs with matching importer membership on both sides.
- A parser isolation check against the committed `pnpm-lock.yaml`: perturbing one leaf `version:` line inside each importer block reports exactly that importer, 20 of 20 non-empty blocks (`guide` is `{}` and has no leaf).
- Two local experiments, both reverted: the simplified parser in the inline comment, and a deliberately tightened content regex to demonstrate the test gap.

### Published inline

Two comments: the parser's test fixtures never exercise real lockfile syntax, so a parse regression would silently disable the whole feature; and the parser carries four shape-validation flags that no test exercises and that the exact block comparison makes unnecessary.

### Residual boundary, measured and triaged

The new check attributes changes recorded in `importers:`. A transitive re-resolution that only touches `packages:` or `snapshots:` never appears there. I measured how often that matters by computing each importer's full transitive closure (following `link:` edges between workspaces) on both sides of every scoped commit: **20 of the 86 scoped plans have an unchanged importer whose closure drifted**. Triaging them, none is a credible under-testing risk:

- The largest cluster is `website` drifting because `website/package.json` depends on `examples` at `workspace:*` and Renovate updated `examples/package.json` (#6629, #6633, #6650, #6684, #6698, #6760, #6814, #6857, #6861). The only skipped workflow is `plus`, whose Playwright project is `testMatch: [/website\/tests\/ariakit-plus/]`, and `website/tests/ariakit-plus.ts` is Clerk and Stripe subscription flows. Meanwhile `main.yml`'s `website` job already runs `pnpm run build-website-lite` and its browser jobs run the chrome, firefox, safari, ios and android projects against that build.
- Second cluster: `app` drifting from `nextjs/package.json` updates (#6576, #6559) because `app` devDepends on `nextjs` at `workspace:*`. The plan already selects `app` and `perf`; only `og_images` and `build_styles` are skipped, and both re-run on `push: [main]` and write back.
- The root `.` importer drifts on nearly any package update because root links every `@ariakit/*` workspace, so it carries no signal.

Disposition: Declined non-goal. Worth recording as a known boundary of the design rather than chasing, and my closure metric is an approximation, so treat the count as an upper bound.

### Other candidates not published inline

- **`git merge-base` is now resolved twice per plan**, once in `getPackageJSONChanges` and once in `getLockfileImporterChanges`, on top of the resolution `git diff base...head` already does. The manifest snapshot and the lockfile snapshot are therefore derived independently even though the whole attribution argument assumes they are the same commit. Disposition: Declined non-goal, since both call sites pass the same arguments today; resolving it once in `runCIPlan` and threading it down would make that assumption structural rather than incidental.
- **`files.map(normalizeCIPath).includes("pnpm-lock.yaml")`** builds a whole array to answer a membership question that `files.some(...)` answers without allocating, and `normalizeCIPath` is applied to `` `${importer}/package.json` ``, which is POSIX by construction. Disposition: Declined non-goal, cosmetic.
- **`hasScopedPackageJSONChanges` and `hasScopedLockfileChanges` are both load-bearing**, which I checked rather than assumed: a version-only manifest change plus a lockfile leaves the first false, and an unrelated changed importer leaves the second false. Not a finding, recorded so the pairing is not simplified away later.

No follow-up issues were registered; every confirmed finding is published inline.

### Not fully validated

- I did not run `pnpm tsc` or `pnpm build`. The change stays inside a private script package and its test, and `pnpm lint` plus the focused suite passed.
- I only exercised the parser against `lockfileVersion: '9.0'`, the version the repository currently commits. A future pnpm bump that changes the `importers:` layout would be caught by the regression test suggested inline, but is not covered today.
- The transitive-closure metric follows `link:` edges and peer-suffixed snapshot keys but does not model optional or dev pruning, so it over-reports drift.

packages/ariakit-scripts/src/ci.test.ts
diegohaz: Already handled at `packages/ariakit-scripts/src/ci.ts:496`: the importer parser accepts the repository lockfile’s quoted scoped keys, `workspace:*` and `link:` values, and peer-suffixed versions. Added targeted coverage in [`a6977aa18`](https://github.com/ariakit/ariakit/commit/a6977aa18).

packages/ariakit-scripts/src/ci.ts
diegohaz: Fixed in [`a6977aa18`](https://github.com/ariakit/ariakit/commit/a6977aa18): the parser now keeps only the importer-boundary and raw-block comparison checks, removing the four nested YAML shape-tracking flags and `saveCurrentField`.

diegohaz:
**Fixed**

- [`a6977aa18`](https://github.com/ariakit/ariakit/commit/a6977aa18) ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657844782)): reduced importer parsing to the boundary and exact raw-block checks that CI attribution needs.

**Already resolved**

- [`a6977aa18`](https://github.com/ariakit/ariakit/commit/a6977aa18) ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3657839198)): the parser already accepted the repository lockfile; added a proven regression test covering its real syntax.

packages/ariakit-scripts/src/ci.ts
```diff
@@ -108,6 +176,117 @@ function isAppRuntimePath(file: string) {
...
+  if (!hasOnlyFields(fields, publicPackageFields)) return false;
+  if (
+    fields.includes("version") &&
+    (typeof change.base.version !== "string" ||
+      typeof change.head.version !== "string")
+  ) {
+    return false;
+  }
+  addDependencyReason(plan, "main", file);
+  if (fields.some((field) => runtimeDependencyFieldSet.has(field))) {
```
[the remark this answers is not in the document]
diegohaz: Validated on the current head. The routing claim is correct and the mechanism is real, but the specific example is wrong and the failure is not reachable in this repository today. Details, because the distinction changes what is worth doing about it.

**The mechanism is real.** `packages/ariakit-react-utils/src/hooks.ts:46` is `export const useSafeLayoutEffect = canUseDOM ? useLayoutEffect : useEffect;` and line 68 is `export function useLiveRef<T>(value: T) {`. Neither has a type annotation, so the two signatures in the committed readme are entirely checker output:

```ts
const useSafeLayoutEffect: typeof React.useLayoutEffect;
function useLiveRef<T>(value: T): RefObject<T>;
```

I confirmed that a type-package change rewrites them. Reverting the one line in the resolved `@types/react` that React 19 changed for real:

```diff
-    function useRef<T>(initialValue: T): RefObject<T>;
+    function useRef<T>(initialValue: T): MutableRefObject<T>;
```

then running `pnpm -F @ariakit/react-utils run docs` produces:

```diff
-function useLiveRef<T>(value: T): RefObject<T>;
+function useLiveRef<T>(value: T): MutableRefObject<T>;
```

The baseline run before the patch wrote no changes, so the readme is a deterministic function of the source plus the resolved types. Both the type patch and the regenerated readme were reverted.

**The example is wrong.** The `react` devDependency of `packages/ariakit-react-utils` ships no types, so bumping it cannot move the readme. What produces `RefObject<T>` is `@types/react`, and no package under `packages/` declares it; it resolves from the root manifest. So the exposed route is root `@types/react`, which is in `rootMainDependencies` and therefore maps to Main. I ran both shapes through `createCIPlan` on this head and both select `main` alone, so the routing half of the finding holds either way.

**It is not reachable today.** Renovate's `react` group does split when only some members have updates, and it has produced exactly this PR shape: `4a4e5cc30` ("Update react to v19.2.17", #6282) bumped only `@types/react`, across `package.json`, `app/package.json`, `examples/package.json`, `nextjs/package.json` and `templates/react/package.json`. Replaying it through this head gives full CI, not Main. The reason is `242601353`: the lockfile also moves the `packages/ariakit-test` importer, whose manifest did not change, because `@ariakit/test` peer-depends on `@testing-library/react` and `@types/react` appears in that importer's resolved peer keys. `hasScopedLockfileChanges` is false and the plan falls back.

Widening that to the whole window: across the last 400 commits on `main`, 82 scoped plans skip Docs, and exactly one of them changed a type-supplying dependency — `afba4942c` (#6702, `@types/hast`), an `app/package.json` devDependency that no package with a `docs` script consumes. Zero reachable instances.

So the gap exists in the routing table but is currently masked by an incidental property of the dependency graph. Nothing in the policy states or tests that property, and it disappears if `@ariakit/test` ever drops its `@testing-library/react` peer, or if a type package ends up scoped to a single manifest.

Given that, the proportionate response is the cheap hardening rather than a redesign: also select `docs` when a changed dependency name starts with `@types/`, in the root branch next to the existing `oxfmt` case and in the public-package branch. That is the same shape as the `vitest` and `oxfmt` special cases already there, and it does not touch the lockfile logic.

**Assessment**

- Severity: Medium (estimate, 75% confidence), because the effect is committed readme drift, and `docs.yml` runs on `push: [main]` with writeback so it self-heals after merge rather than being lost.
- Priority: P2 (estimate, 70% confidence), because no reachable instance exists on this head; the argument for doing it now is that the protection is accidental rather than designed.
- Breakage likelihood: Rare (verified over 400 commits), because it needs a type-package update whose lockfile effect stays inside the changed manifests' importers, and that did not occur once in the sampled window.
- Fix effort: Easy (estimate, 90% confidence), because it is a name-prefix check in the two branches that already special-case `vitest` and `oxfmt`, plus a test.

diegohaz (commented):
## Re-review

Re-review. Baseline is the head of the previous review, `d053b8770`, so the incremental diff is the single commit `a6977aa18` ("Simplify lockfile importer parsing"). I also re-validated the one unresolved thread against this head, and re-ran the historical replay because the parser changed.

**No new findings.** Both items from the last round are verified fixed, and the new commit introduces nothing I would flag.

### Previous findings, verified against behavior

- **Parser test fixtures.** Fixed. The new "parses the repository lockfile" test is the right guard, and I confirmed it actually catches the degradation I demonstrated last round: re-applying the tightened content regex (`[a-zA-Z0-9._-]+` in place of `[^#\s][^:]*`) now fails that test and only that test, 1 failed and 42 passed. Before the commit, the identical change left all 42 green.
- **Parser shape flags.** Fixed, and the committed parser matches the replacement I proposed. I re-ran the full 400-commit replay against it: 154 of 154 real lockfile pairs still parse with matching importer membership, still 86 scoped plans, and the six originally-gapped commits (#6859, #6798, #6695, #6561, #6558, #6516) still all fall back to full CI. The simplification is behavior-preserving on every real input in the repository's history.

I also probed one failure mode the smaller parser could plausibly have introduced, mis-splitting nested keys into importers, by loosening the importer key regex from `^  ` to `^ +`. That is caught: three tests fail. So the boundary rules are adequately covered.

### Unresolved thread validated in this run

- `packages/ariakit-scripts/src/ci.ts:282`, Codex, "Run docs for type-affecting package dev dependencies" (`PRRT_kwDOBjT0Fc6UFKPc`). **Partly confirmed**, disposition P2 hardening rather than a defect. The routing claim holds: both a root `@types/react` bump and a public package's `react` devDependency bump select Main alone on this head. The generator really is checker-derived, and I proved it by reverting React 19's `useRef` return-type change in the resolved `@types/react` and watching `packages/ariakit-react-utils/readme.md` change from `RefObject<T>` to `MutableRefObject<T>`. But the cited example is wrong, since `react` ships no types, and the failure is not reachable: the one historical PR of this shape, `4a4e5cc30` (#6282, `@types/react` only), replays to full CI on this head because the lockfile also moves the `packages/ariakit-test` importer without its manifest changing. Across 400 commits, 82 scoped plans skip Docs and exactly one touched a type-supplying dependency (`afba4942c`, #6702, `@types/hast`, which no docs package consumes). I replied on the thread with the evidence and a one-line-shaped hardening suggestion. The type patch and the regenerated readme were reverted; the worktree is clean.

### What I ran

- `pnpm test packages/ariakit-scripts/src/ci.test.ts --run` — 43 passed.
- `pnpm lint` — formatting clean; only the pre-existing `examples/menu-slide/test-chrome-safari.ts` oxlint warnings remain.
- The 400-commit replay against the new parser, plus a targeted scan for scoped plans that skip Docs while changing a type-supplying dependency.
- `pnpm -F @ariakit/react-utils run docs` at baseline (no drift) and again under a patched `@types/react` (drift), then reverted both.
- Three deliberate degradations of the parser to test the new coverage, all reverted.

### Still open from earlier rounds, unchanged

These were dispositioned in previous reviews and nothing on this head changes them: the `snapshots:`-only transitive boundary of the importer check (measured, triaged, Declined non-goal), `git merge-base` being resolved separately in `getPackageJSONChanges` and `getLockfileImporterChanges`, and the cosmetic `files.map(...).includes(...)` allocation.

### Not fully validated

- I did not run `pnpm tsc` or `pnpm build`. The change stays inside a private script package and its test, and `pnpm lint` plus the focused suite passed.
- The parser is still only exercised against `lockfileVersion: '9.0'`, which the new repository-lockfile test now pins to whatever the repository commits.

packages/ariakit-scripts/src/ci.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`d459daf5c`](https://github.com/ariakit/ariakit/commit/d459daf5c): conventional `@types/*` changes in reviewed root and public package manifests now select Docs so generated README writeback can run. Added targeted root and package routing coverage in the same commit.

diegohaz:
## Fixed

- [`d459daf5c`](https://github.com/ariakit/ariakit/commit/d459daf5c) ([reply](https://github.com/ariakit/ariakit/pull/6879#discussion_r3658225582)): route conventional `@types/*` updates in reviewed root and public package manifests through Docs, with focused coverage for both routes and the ordinary public development-dependency boundary.

PR #6881, opened 2026-07-27
Stabilize dialog backdrop Safari test

diegohaz:
## Motivation

The Safari `app.yml` job frequently reports `dialog-backdrop-6339/test-browser` as flaky because the 500ms backdrop assertion window can expire under CI load. In the latest sample, 13 of 49 valid Safari jobs retried this test, including 8 of the latest 20.

## Solution

Give only the fast-dialog fixture a 2s backdrop exit transition while keeping its enter transition at 500ms and the panel at 150ms. The test still checks after 250ms that the panel's shorter timeout did not stop the backdrop, but now has enough scheduling headroom on macOS CI. Extend the happy-dom poll timeout accordingly.

```css
.backdrop-long[data-leave] {
  transition-duration: 2s;
}
```

## Preview

[Fast 500ms entrance and 2s exit](https://github.com/user-attachments/assets/25e44432-442d-446d-b7a4-eecde66da29a)

## Verification

- `pnpm test-react app/src/sandbox/dialog-backdrop-6339/test.ts` — 2 tests passed
- `pnpm test-react18 app/src/sandbox/dialog-backdrop-6339/test.ts` — 2 tests passed
- `APP_PORT=44331 NEXTJS_PORT=44332 APP_INSPECTOR_PORT=19441 NEXTJS_INSPECTOR_PORT=19442 pnpm -F app run test-safari src/sandbox/dialog-backdrop-6339/test-browser.ts --repeat-each=5 --workers=3 --retries=0` — 10/10 passed
- `pnpm tsc`
- `pnpm --filter ./app build`

### DIFF
--- a/app/src/sandbox/dialog-backdrop-6339/index.react.tsx
+++ b/app/src/sandbox/dialog-backdrop-6339/index.react.tsx
@@ -2,16 +2,17 @@ import * as Ariakit from "@ariakit/react";
 
 // See https://github.com/ariakit/ariakit/issues/6339
 //
-// The backdrop has a 500ms opacity transition: it should fade in when the
-// dialog opens and fade out when it closes. There are two dialogs:
+// The backdrop has an opacity transition: it should fade in when the dialog
+// opens and fade out when it closes. There are two dialogs:
 //
 // - "Show dialog": the panel has no transition at all, so it snaps in place
 //   while the backdrop fades. Before the fix, the enter fade works, but on
 //   close the dialog hides instantly: the backdrop never receives
 //   `data-leave` and the exit fade is skipped.
-// - "Show fast dialog": the panel has a shorter 150ms transition. Before the
-//   fix, the panel's own timeout stopped the shared animation state early,
-//   hiding the backdrop at 150ms and cutting its 500ms exit fade short.
+// - "Show fast dialog": the panel has a shorter 150ms transition and the
+//   backdrop uses a 2s transition to leave enough timing headroom for the
+//   browser test. Before the fix, the panel's own timeout stopped the shared
+//   animation state early, hiding the backdrop at 150ms.
 //
 // The styles are inlined in a <style> tag (rather than a style.css import) on
 // purpose: CSS imports are only processed in vitest for the allowlist in the
@@ -27,6 +28,9 @@ const css = `
   }
   .backdrop[data-enter] {
     opacity: 1;
+  }
+  .backdrop-long {
+    transition-duration: 2s;
   }
   .dialog {
     position: fixed;
@@ -78,12 +82,12 @@ export default function Example() {
       </Ariakit.Dialog>
       <Ariakit.Dialog
         store={fastDialog}
-        backdrop={<div className="backdrop" />}
+        backdrop={<div className="backdrop backdrop-long" />}
         className="dialog dialog-fast"
       >
         <Ariakit.DialogHeading>Fast</Ariakit.DialogHeading>
         <p>
-          The panel fades in over 150ms while the backdrop fades over 500ms. On
+          The panel fades in over 150ms while the backdrop fades over 2s. On
           close, the backdrop should finish its longer fade out.
         </p>
         <Ariakit.DialogDismiss>Close</Ariakit.DialogDismiss>
--- a/app/src/sandbox/dialog-backdrop-6339/index.react.tsx
+++ b/app/src/sandbox/dialog-backdrop-6339/index.react.tsx
@@ -10,7 +10,7 @@ import * as Ariakit from "@ariakit/react";
 //   close the dialog hides instantly: the backdrop never receives
 //   `data-leave` and the exit fade is skipped.
 // - "Show fast dialog": the panel has a shorter 150ms transition and the
-//   backdrop uses a 2s transition to leave enough timing headroom for the
+//   backdrop uses a 2s exit transition to leave enough timing headroom for the
 //   browser test. Before the fix, the panel's own timeout stopped the shared
 //   animation state early, hiding the backdrop at 150ms.
 //
@@ -29,7 +29,7 @@ const css = `
   .backdrop[data-enter] {
     opacity: 1;
   }
-  .backdrop-long {
+  .backdrop-long[data-leave] {
     transition-duration: 2s;
   }
   .dialog {
@@ -87,8 +87,8 @@ export default function Example() {
       >
         <Ariakit.DialogHeading>Fast</Ariakit.DialogHeading>
         <p>
-          The panel fades in over 150ms while the backdrop fades over 2s. On
-          close, the backdrop should finish its longer fade out.
+          The panel fades in over 150ms while the backdrop fades in over 500ms.
+          On close, the backdrop should finish its longer 2s fade out.
         </p>
         <Ariakit.DialogDismiss>Close</Ariakit.DialogDismiss>
       </Ariakit.Dialog>
--- a/app/src/sandbox/dialog-backdrop-6339/test-browser.ts
+++ b/app/src/sandbox/dialog-backdrop-6339/test-browser.ts
@@ -35,9 +35,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.button("Show fast dialog")).toBeFocused();
     await test.expect(backdrop).toHaveAttribute("data-leave", "true");
     // Wait past the panel's 150ms transition. The backdrop must keep leaving
-    // until its own 500ms transition ends. Before the fix, the panel's shorter
-    // timeout stopped the shared animation state, hiding the backdrop at
-    // 150ms.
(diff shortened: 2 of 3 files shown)


app/src/sandbox/dialog-backdrop-6339/index.react.tsx
```diff
@@ -28,6 +29,9 @@ const css = `
   .backdrop[data-enter] {
     opacity: 1;
   }
+  .backdrop-long {
+    transition-duration: 2s;
+  }
```
diegohaz: `.backdrop-long` is unscoped, so it stretches the *enter* transition too, not just the exit window the test needs. Two consequences:

1. The fast dialog's preview now takes 2s to darken. The test (and a human following the same path) clicks Close roughly 50–100ms after opening, when the backdrop has reached ~2–5% opacity, so the copy right next to it — "the backdrop should finish its longer fade out" — describes a fade that is no longer visible. This sandbox is served at `/react/previews/dialog-backdrop-6339/`, and it is the artifact someone opens to see #6339 reproduced or fixed.
2. It silently changes the fixture's enter contract. On `main` the `data-enter` assertion at line 31 of `test-browser.ts` ran against a 500ms enter; it now runs against a 2s one. That was not the goal of the change and is not covered by the PR description.

Scoping the override to the leave state keeps everything the PR is after — the exit window is the only one the 250ms checkpoint has to fit inside — while leaving the enter path byte-identical to `main`:

```suggestion
  .backdrop-long[data-leave] {
    transition-duration: 2s;
  }
```

`data-enter` is removed and `data-leave` added in the same render in `useDisclosureContent`, so the browser starts the opacity transition with the new 2s duration, and `getElementEndTime` reads 2s in the layout effect that follows. The stylesheet already keys off `data-enter` twice, so this stays idiomatic for the fixture.

I validated this locally on the PR head:

- `pnpm -F app run test-safari src/sandbox/dialog-backdrop-6339/test-browser.ts --repeat-each=5 --workers=3 --retries=0` — 10/10 passed, and the "fast dialog" test still takes 3.2–3.3s, identical to the unscoped version, so the leave headroom is unchanged.
- Same file on `--project=chrome` (10/10) and `--project=firefox` (6/6).
- `pnpm test-react app/src/sandbox/dialog-backdrop-6339/test.ts` — 2 passed, with the existing `{ timeout: 3000 }` still needed and still sufficient.
- Red-state check: with `unstable_otherElementRef: backdropRef` removed from `useDialog`, the scoped variant still fails at `test-browser.ts:41` (`data-leave` missing after the 250ms wait), which is the same assertion and the same reason as the unscoped version.

**Assessment**

- Severity: Low (verified), because it only affects fixture and preview quality; no shipped behavior changes.
- Priority: P2 (estimate, 80% confidence), because it is a one-token change in a line this PR is already touching, but nothing breaks if it ships as is.
- Breakage likelihood: Likely (verified), because the slow fade-in and the changed enter contract happen on every run and every preview visit, not under rare conditions.
- Fix effort: Easy (verified), because the suggestion above is the whole change and I ran the affected suites against it.

diegohaz (commented):
Initial review. Baseline: merge base `9698cb115` → head `b3a5ebadf` (all three changed files, plus the surrounding fixture, `useDisclosureContent`, the app Playwright config, the root Vitest config, and the CI history for this test).

There were no existing review threads to validate.

## What I checked and ran

- Reproduced the flake's shape from CI rather than taking the description on trust. I scanned the Safari job logs of the 40 most recent `app.yml` runs on `main`: 7 of them retried this file, always the second test (`test-browser.ts:25:3`), never the first (`:5:3`). The failing anchors on the pre-PR file were `:43` (`toBeVisible()` after the wait) in all 7, `:42` (`data-leave` after the wait) in 4, and `:36` (`data-leave` *before* the wait) in 3. That matches the "the 500ms window expired" diagnosis, including the sub-case where the window expires before the 250ms wait even starts.
- Instrumented the pre-PR fixture on Safari locally to measure the margin the PR is buying. From just before the Close click, the mid-assertion checkpoint lands at 316–379ms and the backdrop actually hides at 594–659ms (500ms transition plus ~100ms of click and animation-start latency), so the pre-PR margin was ~270ms. At 2s it becomes ~1.8s.
- `pnpm -F app run test-safari src/sandbox/dialog-backdrop-6339/test-browser.ts --repeat-each=5 --workers=3 --retries=0` — 10/10 passed. Same file on `--project=chrome` (10/10) and `--project=firefox` (10/10).
- `pnpm test-react app/src/sandbox/dialog-backdrop-6339/test.ts` and `pnpm test-react18 …` — 2 passed each, ~3.1s of test time. The `{ timeout: 3000 }` bump is required (the poll needs ~1.7s after the 250ms sleep, over the 1000ms default) and leaves ~1.3s of slack inside the 10s `testTimeout`.
- Regression proof re-verified against the current fixture: with `unstable_otherElementRef: backdropRef` removed from `useDialog` (the #6597 fix), the browser test fails at `test-browser.ts:41` and the happy-dom test at `test.ts:42`, both on the missing `data-leave` after the 250ms wait. The widened window does not weaken what the test catches.
- Confirmed the 2s transition cannot leak elsewhere: `previews-browser.ts` only screenshots the closed preview, `test-browser.ts` maps to chrome/firefox/safari only, and Playwright's default 5s expect timeout still leaves ~3.3s of slack on the closing `toBeHidden()`.
- No changeset needed — sandbox and test files only, nothing shipped. The description's claims match the code.

## Findings

One inline comment: `.backdrop-long` is unscoped, so it also stretches the enter transition, which degrades the preview and silently changes the fixture's enter contract. Scoping it to `[data-leave]` keeps the exit headroom identical; validated across all three browser projects and both happy-dom suites, red state included.

## Candidates evaluated and not published inline

**Replace the widened window with a precise duration measurement instead.** Disposition: declined non-goal. The concern is real — the PR widens a bounded wall-clock window rather than removing the race, so the test's stability still rides on the fixture's duration. But I prototyped both variants and neither is an improvement here. A Playwright-side `Date.now()` lower bound is too coarse: it reported 840–907ms for a true ~600ms hide, because `expect` polling backs off through 100/250/500ms, so a buggy 150ms hide would also land well above any usable threshold and the test would quietly stop catching #6339. The in-page `MutationObserver` variant (the pattern at `app/src/sandbox/disclosure-content-lifecycle/test-browser.ts:96-136`) is precise, but here it needs the click driven from inside `page.evaluate` and has to disambiguate the two `[role="presentation"]` backdrops this fixture renders, and its ~140ms latency floor eats much of the 150ms-vs-500ms discrimination margin. That is more machinery, further from "drive the test through the rendered experience", for no measured gain over the 6.6x margin this PR already buys.

**The first test still runs inside the original 500ms window.** Disposition: declined non-goal. Up to the first `data-leave` assertion, test 1 (lines 12/15/19/20) is structurally identical to test 2 (lines 33/35/36), and the `:36`-style failure mode — the whole window expiring before Playwright's first observation — is demonstrated in the CI logs. So test 1 keeps a residual exposure that this PR does not cover. Against that: it never flaked in the 40 runs I scanned, and covering it would mean a second 2s transition, another ~1.5s of test time, and a second sluggish preview dialog. Worth revisiting only if it actually starts flaking.

- Severity: Low (estimate, 85% confidence), because the outcome is one retried Safari test, which the project already absorbs via `retries: 3`.
- Priority: P2 (estimate, 75% confidence), because nothing observed today depends on it.
- Breakage likelihood: Rare (verified), because it requires a >500ms scheduling spike inside a single attempt, and it occurred 0 times in 40 runs.
- Fix effort: Easy (estimate, 90% confidence), because it is the same one-line CSS class applied to the second backdrop.

## Not fully validated

I could not reproduce the flake locally — my Safari runs are on an unloaded machine, so every repeat passed on both the old and new fixture. The judgement that 2s is enough headroom rests on the CI failure anchors plus the measured ~270ms → ~1.8s margin change, not on a local red run of the pre-PR fixture.

app/src/sandbox/dialog-backdrop-6339/index.react.tsx
diegohaz: Fixed in [`f8b22ec47`](https://github.com/ariakit/ariakit/commit/f8b22ec47a70082b9ea7e4848fa6ebfe5e82b9fb): scoped the 2s duration to `data-leave`, restoring the 500ms enter while preserving the longer exit window. I also updated the fixture copy and PR description to match the two timings.

[Preview: fast backdrop enters in 500ms and leaves over 2s](https://github.com/user-attachments/assets/25e44432-442d-446d-b7a4-eecde66da29a)

diegohaz:
**Fixed**

- [`f8b22ec`](https://github.com/ariakit/ariakit/commit/f8b22ec47a70082b9ea7e4848fa6ebfe5e82b9fb) ([reply](https://github.com/ariakit/ariakit/pull/6881#discussion_r3657886759)): Scoped the 2s backdrop duration to the leave state, restored the 500ms enter, and updated the preview copy and evidence.

PR #6891, opened 2026-07-28
Fix ComboboxPopover Escape baseline

### LINKED ISSUE #6852: ComboboxPopover re-renders per keystroke and its Escape baseline can drift
## Summary

Two small issues in `useComboboxPopover`, both found while reviewing PR #6832 and both deliberately left out of it because the fixes are more invasive than the problems warrant at that point in the branch.

### 1. `ComboboxPopover` re-renders on every arrow key when `selectOnMove` is enabled

`useComboboxPopover` subscribes to the raw `selectedValue` state. It needs the value for two things only: an `Array.isArray` check, and as the input to the effect that tracks the value the popover had before it was shown.

With [`selectOnMove`](https://ariakit.com/reference/combobox-provider#selectonmove) enabled — the mode the popover's `resetOnEscape` default targets — every arrow key changes `selectedValue`, so the whole `ComboboxPopover` re-renders per keystroke, including `usePopover`, `useDialog`, and the floating-positioning machinery. Before the popover took ownership of that state, no popover subscription changed on move.

A boolean selector for the multi-select check plus a store-subscribed ref for the baseline would avoid the render, at the cost of more code than the current form. Worth measuring against the combobox perf fixtures before choosing a shape.

### 2. The Escape baseline can drift when the value changes before the first move

The baseline is frozen once the popover is mounted **and** `moves` is non-zero. `moves` is not incremented by opening the popup (the store assigns `activeId` directly) nor by clicking an item (`CompositeItem` uses `setActiveId`). With a mouse, hovering fires `move()` first and masks this; with touch or a programmatic click it does not.

Repro: a `ComboboxPopover` with an explicit `resetOnEscape` (and `selectOnMove` off) over `ComboboxItem`s with `hideOnClick={false}`. Tap a second item on a touch device, press Escape: the value stays on the tapped item instead of reverting, because the tap updated the baseline.

## Why this is separate from #6832

Neither is a regression against `main`: `resetOnEscape` on `ComboboxPopover` and the baseline it depends on are both new in that PR, and its own tests cover the mouse and keyboard paths. Both fixes need a different shape for the baseline than the one that shipped, so they are better done deliberately than folded into an already large PR.

## Expected outcome

`ComboboxPopover` no longer re-renders per keystroke under `selectOnMove`, and the Escape baseline is frozen by the popover becoming visible rather than by the first move, so it is correct for pointer, touch, and programmatic selection alike.

### Workaround

https://github.com/ariakit/ariakit/issues/6852#issuecomment-5085966205

diegohaz:
## Motivation

[`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) can restore the wrong value when selection changes without calling the composite `move()` method. Touch and programmatic clicks can select an item while `moves` remains `0`, so the selected-value subscription replaces the Escape baseline with the new value.

For example, clicking `Banana` directly and then pressing Escape leaves `Banana` selected instead of restoring `Apple`:

```tsx
<ComboboxProvider defaultSelectedValue="Apple" selectOnMove={false}>
  <ComboboxSelect />
  <ComboboxPopover resetOnEscape>
    <ComboboxItem value="Banana" hideOnClick={false} />
  </ComboboxPopover>
</ComboboxProvider>
```

Fixes #6852.

## Solution

- Captures the selected value when the popover changes from closed to open, so selection changes during the visible session cannot replace the Escape baseline.
- Preserves that baseline across the synchronous close rollback when `onClose` prevents the close.
- Adds browser-primary and happy-dom coverage for selection without a preceding move, prevented-close rollbacks, synchronous selection changes in `onClose`, store replacement, and repeated sessions.
- Adds a patch changeset for `@ariakit/react-components` and `@ariakit/react`.

## Workaround

For keep-open items that can be selected without a preceding `move()`, call the store's `move()` method from the consumer `onClick` handler. Consumer event handlers run before Ariakit's internal selection handler, so this freezes the pre-click Escape baseline before the new value is committed:

```tsx
const store = useComboboxStore({
  defaultSelectedValue: "Apple",
  selectOnMove: false,
});

<ComboboxItem
  value="Banana"
  hideOnClick={false}
  onClick={(event) => {
    // TODO: Remove when https://github.com/ariakit/ariakit/issues/6852 is fixed.
    store.move(event.currentTarget.id);
  }}
/>
```

This workaround is intended for the reported keep-open item composition. It deliberately marks the clicked item active and increments the store's move count before selection, and it assumes the consumer handler receives the rendered item's stable `id`.

## Preview

| Before | After |
| --- | --- |
| Escape leaves `Banana` selected. | Escape restores `Apple`. |
| Closed ComboboxSelect incorrectly retaining Banana after Escape | Closed ComboboxSelect correctly restored to Apple after Escape |

## Testing

Regression proof from the first commit:

- React 19: 26 passed, 1 failed with expected `Apple`, received `Banana`.
- Chrome: the focused browser test failed with expected `Apple`, received `Banana` after the popover closed.

Final verification:

- React 19: 27/27 focused tests passed.
- React 18: 27/27 focused tests passed.
- Chrome, Firefox, and Safari: 51/51 browser tests passed.
- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build`
- `pnpm build-app-lite`
- `pnpm changeset status --since=origin/main`

### DIFF
--- a/.changeset/300-combobox-popover-escape-baseline.md
+++ b/.changeset/300-combobox-popover-escape-baseline.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) to preserve its [`resetOnEscape`](https://ariakit.com/reference/combobox-popover#resetonescape) baseline when an item is selected without a preceding move.
--- a/app/src/sandbox/combobox-popover-reset-on-escape/index.react.tsx
+++ b/app/src/sandbox/combobox-popover-reset-on-escape/index.react.tsx
@@ -296,14 +296,28 @@ function StoreReplacement() {
 }
 
 function DirectClick() {
+  const store = Ariakit.useComboboxStore({
+    defaultSelectedValue: "Apple",
+    selectOnMove: false,
+  });
+
   return (
-    <Ariakit.ComboboxProvider defaultSelectedValue="Apple" selectOnMove={false}>
+    <Ariakit.ComboboxProvider store={store}>
       <Ariakit.ComboboxSelect aria-label="Direct click">
         <Ariakit.ComboboxSelectedValue />
       </Ariakit.ComboboxSelect>
       <Ariakit.ComboboxPopover resetOnEscape>
         {values.map((value) => (
-          <Ariakit.ComboboxItem key={value} value={value} hideOnClick={false} />
+          <Ariakit.ComboboxItem
+            key={value}
+            value={value}
+            hideOnClick={false}
+            onClick={(event) => {
+              // TODO: Remove when https://github.com/ariakit/ariakit/issues/6852
+              // is fixed.
+              store.move(event.currentTarget.id);
+            }}
+          />
         ))}
       </Ariakit.ComboboxPopover>
     </Ariakit.ComboboxProvider>
--- a/app/src/sandbox/combobox-popover-reset-on-escape/index.react.tsx
+++ b/app/src/sandbox/combobox-popover-reset-on-escape/index.react.tsx
@@ -296,28 +296,14 @@ function StoreReplacement() {
 }
 
 function DirectClick() {
-  const store = Ariakit.useComboboxStore({
-    defaultSelectedValue: "Apple",
-    selectOnMove: false,
-  });
-
   return (
-    <Ariakit.ComboboxProvider store={store}>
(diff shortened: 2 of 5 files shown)


diegohaz:
Closing this as not planned. After reassessing the interaction contract, capturing the baseline when the popover opens would turn `resetOnEscape` from movement cleanup into session-wide cancellation, including direct item activations made without a preceding move. That is not behavior we intend to support here.

The existing baseline freezes at the first composite item movement; selection changes made before that point become the value Escape restores. #6895 updates the public JSDoc to state that movement-scoped behavior explicitly, without changing runtime behavior.

PR #6893, opened 2026-07-28
Deprecate Select APIs

### LINKED ISSUE #6836: Soft-deprecate the Select components
## Context

Discovered while reviewing #6832, which adds ComboboxSelect as the unified API for standard and filterable selects. The legacy Select API remains supported in that pull request so the migration can land without mixing a broad deprecation pass into the implementation review.

## Work

Soft-deprecate the public Select components, hooks, stores, options, and props with `@deprecated`, pointing consumers to the corresponding Combobox APIs. Update both reference generators, component documentation, live-example links, and the deprecation changeset. Keep the consolidated legacy Select sandbox as compatibility coverage until the API is removed in a future version.

In `guide/500-value-components/readme.md`, remove the remaining legacy `SelectValue` and `SelectItemSelected` references and examples from the primary recommendations, replacing them with `ComboboxSelectedValue` and `ComboboxItemSelected`.

## Expected outcome

Consumers see a consistent migration path in TypeScript, both reference sites, and the docs, while existing Select behavior stays covered until removal.

This work is independent of #6832 because the unified replacement can ship correctly while Select remains supported.

diegohaz:
## Motivation

The Select APIs remain available for compatibility, but ComboboxSelect and the corresponding Combobox APIs are now the canonical path for standard and filterable selects. Consumers need consistent editor guidance without a runtime warning or behavior change.

Fixes #6836.

## Solution

- Adds `@deprecated` replacement guidance to the public Select components, hooks, stores, overloads, renderer helpers, options, props, and types across the core and React packages.
- Covers the deprecation metadata in both reference-generation pipelines.
- Removes legacy Select value-component guidance while preserving the `select-legacy` sandbox.

For example, TypeScript now surfaces the replacement directly:

```ts
/** @deprecated Use `ComboboxSelect` instead. */
export const Select = forwardRef(/* ... */);
```

## Testing

- `pnpm lint-fix`
- `pnpm test app/src/lib/jsdoc-loader.test.ts website/build-pages/reference-utils.test.ts --run`
- `pnpm tsc`
- `pnpm build`
- `pnpm build-app-lite`
- `pnpm build-website-lite`

### DIFF
--- a/.changeset/290-select-deprecation.md
+++ b/.changeset/290-select-deprecation.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Deprecated the [`Select`](https://ariakit.com/reference/select) APIs in favor of their [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) counterparts.
--- a/.changeset/290-select-deprecation.md
+++ b/.changeset/290-select-deprecation.md
@@ -1,7 +1,0 @@
----
-"@ariakit/components": patch
-"@ariakit/react-components": patch
-"@ariakit/react": patch
----
-
-Deprecated the [`Select`](https://ariakit.com/reference/select) APIs in favor of their [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) counterparts.
--- a/components/select.md
+++ b/components/select.md
@@ -9,39 +9,15 @@ tags:
 
 <div data-description>
 
-The Select components are deprecated. Use [Combobox](/components/combobox) with [`ComboboxSelect`](/reference/combobox-select) to select a value from a list of options.
+Select a value from a list of options presented in a dropdown menu, similar to the native HTML select element. This component is based on the [WAI-ARIA Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).
 
 </div>
 
 <div data-tags></div>
 
-<aside data-type="warn" title="Deprecated">
+<a href="../examples/select/index.react.tsx" data-playground>Example</a>
 
-The Select APIs remain available for compatibility, but they will be removed in a future release. Use [`ComboboxProvider`](/reference/combobox-provider), [`ComboboxSelect`](/reference/combobox-select), and the corresponding Combobox components for new code. The replacement supports both standard and filterable selects through one store.
-
-</aside>
-
-<a href="../examples/select-combobox/index.react.tsx" data-playground>ComboboxSelect example</a>
-
-## Migration
-
-Replace Select components with their Combobox counterparts:
-
-```jsx
-<ComboboxProvider>
-  <ComboboxSelectLabel>Favorite fruit</ComboboxSelectLabel>
-  <ComboboxSelect>
-    <ComboboxSelectedValue />
-    <ComboboxSelectArrow />
-  </ComboboxSelect>
-  <ComboboxPopover>
-    <ComboboxItem value="Apple" />
-    <ComboboxItem value="Banana" />
-  </ComboboxPopover>
-</ComboboxProvider>
-```
-
-## Deprecated API
+## API
 
 ```jsx
 useSelectStore()
--- a/components/select.md
+++ b/components/select.md
@@ -9,15 +9,39 @@ tags:
 
 <div data-description>
 
-Select a value from a list of options presented in a dropdown menu, similar to the native HTML select element. This component is based on the [WAI-ARIA Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).
+The Select components are deprecated. Use [Combobox](/components/combobox) with [`ComboboxSelect`](/reference/combobox-select) to select a value from a list of options.
 
 </div>
 
 <div data-tags></div>
 
-<a href="../examples/select/index.react.tsx" data-playground>Example</a>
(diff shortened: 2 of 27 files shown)


.changeset/290-select-deprecation.md
```diff
@@ -0,0 +1,7 @@
+---
```
diegohaz: No need for a changeset for soft-deprecations.

components/select.md
```diff
@@ -9,15 +9,39 @@ tags:
 
 <div data-description>
```
diegohaz, line 10: No need for docs change for soft-deprecations.

guide/500-value-components/readme.md
```diff
@@ -25,7 +25,7 @@ A value component reads one specific value from a provider, an explicit store, o
 - exposes the value directly or through a `children` render function
```
diegohaz, line 25: Removing mentions to `Select` components like in this document is fine. Keep it.

packages/ariakit-react-components/src/select/select-provider.tsx
```diff
@@ -27,6 +30,7 @@ export function SelectProvider<T extends Value = Value>(
   props: PickRequired<SelectProviderProps<T>, "value" | "defaultValue">,
 ): ReactElement;
 
+/** @deprecated Use `ComboboxProvider` instead. */
```
diegohaz, line 33: Do we need this? Won't all the overloads inherit the larger JSDoc above?

packages/ariakit-react-components/src/select/select-value.tsx
```diff
@@ -63,6 +66,7 @@ export function SelectValue<T extends Value = Value>(
   props: PickRequired<SelectValueProps<T>, "fallback">,
 ): T;
 
+/** @deprecated Use `ComboboxSelectedValue` instead. */
```
diegohaz, line 69: Same here?

.changeset/290-select-deprecation.md
diegohaz: Fixed in [`e95d757`](https://github.com/ariakit/ariakit/commit/e95d757e1c4fa4f08a6ddab5457adf6764d8533c): removed the changeset so this soft deprecation does not produce a release-note entry.

components/select.md
diegohaz: Fixed in [`e95d757`](https://github.com/ariakit/ariakit/commit/e95d757e1c4fa4f08a6ddab5457adf6764d8533c): restored the Select component page and its original playground without adding soft-deprecation migration content.

guide/500-value-components/readme.md
diegohaz: Resolved in [`5ba6e86`](https://github.com/ariakit/ariakit/commit/5ba6e864bbc8ea9c2733cad9b03b3c45233a2efe): kept the accepted value-guide cleanup, so the legacy Select value-component mentions remain removed.

packages/ariakit-react-components/src/select/select-provider.tsx
diegohaz: This is not an issue because TypeScript resolves deprecation metadata per overload signature rather than inheriting it from the preceding overload. Without this tag, calls that resolve to the optional-props `SelectProvider` signature have no `deprecated` metadata; a compiler-resolved signature check returned the tag for the annotated overload and no tag for the unannotated overload.

packages/ariakit-react-components/src/select/select-value.tsx
diegohaz: This is not an issue because the same per-overload TypeScript behavior applies to `SelectValue`. The optional-props overload needs its own tag so ordinary `SelectValue({})` and `<SelectValue />` usages resolve to a signature with `deprecated` metadata.

diegohaz:
**Fixed**

- [`e95d757`](https://github.com/ariakit/ariakit/commit/e95d757e1c4fa4f08a6ddab5457adf6764d8533c) ([reply](https://github.com/ariakit/ariakit/pull/6893#discussion_r3662912598)): Removed the changeset so the soft deprecation does not create a release-note entry.
- [`e95d757`](https://github.com/ariakit/ariakit/commit/e95d757e1c4fa4f08a6ddab5457adf6764d8533c) ([reply](https://github.com/ariakit/ariakit/pull/6893#discussion_r3662913504)): Restored the Select component page and original playground.

**Already resolved**

- [`5ba6e86`](https://github.com/ariakit/ariakit/commit/5ba6e864bbc8ea9c2733cad9b03b3c45233a2efe) ([reply](https://github.com/ariakit/ariakit/pull/6893#discussion_r3662915124)): Preserved the accepted removal of legacy Select value-component guidance.

**Invalid**

- [`No code change`](https://github.com/ariakit/ariakit/blob/e95d757e1c4fa4f08a6ddab5457adf6764d8533c/packages/ariakit-react-components/src/select/select-provider.tsx#L33) ([reply](https://github.com/ariakit/ariakit/pull/6893#discussion_r3662916415)): Kept the explicit `SelectProvider` overload tag because TypeScript deprecation metadata is per resolved signature.
- [`No code change`](https://github.com/ariakit/ariakit/blob/e95d757e1c4fa4f08a6ddab5457adf6764d8533c/packages/ariakit-react-components/src/select/select-value.tsx#L69) ([reply](https://github.com/ariakit/ariakit/pull/6893#discussion_r3662917515)): Kept the explicit `SelectValue` overload tag for the same per-signature behavior.

PR #6894, opened 2026-07-28
Scope CI dependency installs

diegohaz:
## Motivation

Most CI jobs currently restore a workspace-wide pnpm store cache and run an unfiltered install, even when the job only uses one app or one group of packages. Downloading that cache is close to the cost of downloading the packages directly, so it does not solve the main bottleneck.

The shared lockfile should remain authoritative, but each job only needs to materialize the part of the dependency graph reachable from its command's entry workspaces.

## Solution

Extend the local setup action with newline-separated pnpm filters and an install opt-out. Each workflow supplies its entry workspaces, and pnpm automatically expands those selectors to their transitive workspace dependencies while preserving `--frozen-lockfile`:

```yaml
with:
  filters: |
    app...
    website...
```

The setup action rejects enabled installs without a non-empty filter and uses pnpm's `--fail-if-no-match`, so an omitted selector set or a combined selector set with no matches cannot silently fall back to a full install. The flag validates the combined pnpm selection; it does not validate every selector independently when at least one selector matches. The action also disables pnpm's `verifyDepsBeforeRun` repair for partial installs. Without that override, a later `pnpm run` or `pnpm exec` can replace the intentional filtered install with a full workspace install.

This change also:

- removes the workspace-wide pnpm store cache restore;
- runs Playwright from the workspace that declares it;
- skips dependency installation for style generation and the dependency-free Node baseline preparation job;
- scopes direct baseline installs in the performance workflow and guards all direct filtered installs with `--fail-if-no-match`;
- keeps the isolated React 18 compatibility install full because that tool rewrites workspace manifests before resolving its temporary workspace;
- expands lint and Next.js build selectors where those commands consume app types or shared app styles;
- declares the website's direct `guide` workspace dependency so `website...` includes it automatically;
- removes redundant `root...` selectors from multi-workspace jobs because pnpm already materializes the root importer; and
- keeps root scripts as the single command definitions for docs and lite website builds.

The shared `pnpm-lock.yaml` remains authoritative. Filtering only controls which packages are materialized; the lockfile change records the website's previously undeclared `guide` workspace dependency rather than introducing job-specific lockfiles.

## Verification

- `pnpm lint-fix`
- `pnpm run lint`
- `pnpm run lint-css`
- `pnpm test packages/ariakit-scripts/src/react18.test.ts`
- `pnpm run tsc -v`
- `pnpm -F nextjs run build-worker`
- `pnpm run docs` after a clean package-filtered install
- `pnpm run build-website-lite` after a clean `website...` install
- clean filtered-install reproductions for the lint and Next.js jobs
- clean install comparison with and without redundant `root...` selectors; both produced the same 1,755 dependency entries and identical `node_modules` path trees
- extracted setup-action shell checks for empty, single, and multiline filters, plus actual pnpm no-match failure validation
- direct performance-workflow install checks showing an unguarded missing selector exits successfully, while the guarded form fails and all three real selector sets match
- frozen lockfile install validation
- parsed every workflow YAML file
- `git diff --check`

The negative paths have no durable automated workflow test: parsing or matching embedded YAML shell and command flags would recreate the brittle configuration-text coverage removed from this PR, while extracting standalone scripts or adding expected-failure workflow jobs would add disproportionate machinery. The exact setup-action shell and direct performance install commands were executed for the cases above, and the complete PR workflow exercises every valid caller.

<details>
<summary>Cycle 3: Continue filtered installs with an explicit scoped-install contract</summary>

**Assessment**

The CI install-cost problem is medium severity and occurs on every workflow run. It does not affect published package behavior, but it materially affects developer feedback time and CI consumption. A full install materializes 1,757 packages and 2.2 GB, compared with 598 packages/1.0 GB for the root closure, 1,122/1.4 GB for `website...`, and 1,387/1.8 GB for `app...`. The complete PR workflow passed at the prior head. Repeated samples support a roughly fixed per-job install-time reduction that is clearest in short jobs; long browser jobs vary enough that their total runtimes are not evidence of a browser-test speedup.

The implementation has moderate maintenance complexity: each job must name its entry workspaces, while pnpm automatically expands those selectors through the declared workspace dependency graph. This is intentionally narrower than attempting to infer dependencies from arbitrary GitHub Actions YAML, shell commands, generated files, or filesystem imports. Such an analyzer would be substantially more complex and still incomplete.

The implementation has also converged and simplified across review cycles: the first reviewed state had 161 additions/26 deletions and three guaranteed incomplete-closure failures; the current state has 116 additions/22 deletions, fixes those closures, removes brittle YAML-text tests, restores root script entry points, enforces the filtered-install boundary, and removes redundant selectors.

**Decision**

Continue the filtered-install direction. Do not revert to full installs and do not add a workflow-command dependency analyzer.

The resulting design:

- enforces the setup action's contract at runtime by failing when installation is enabled without a non-empty filter or when the combined selector set matches no project;
- omits redundant `root...` entries from multi-workspace filter blocks; and
- keeps the existing shared frozen lockfile and manifest-driven dependency closures.

Intentionally unsupported behavior includes automatic discovery of undeclared cross-workspace imports, arbitrary shell-command dependencies, per-selector staleness inside an otherwise valid multi-selector block, and dependency-graph changes made without updating the affected job seeds. A future intentional full install must use an explicit all-workspaces filter instead of omitting the input. The implicit `app -> nextjs` edge remains accepted because it is valid today and any future removal fails loudly in CI.

Repository-wide persisted checkout credentials remain tracked in [#6897](https://github.com/ariakit/ariakit/issues/6897). Reducing the always-installed root dependency floor remains tracked in [#6898](https://github.com/ariakit/ariakit/issues/6898).

</details>


.github/workflows/docs.yml
```diff
@@ -30,9 +30,11 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: ./packages/*...
 
       - name: Generate docs
-        run: pnpm run docs
+        run: pnpm -F './packages/*' run --if-present docs
```
diegohaz: Why not `pnpm run docs`?

packages/ariakit-scripts/src/ci.test.ts
```diff
@@ -613,6 +613,46 @@ test("parses the repository lockfile", () => {
   expect(parseLockfileImporterChanges(lockfile, lockfile)).toEqual([]);
 });
 
+test("keeps filtered CI installs partial", () => {
```
diegohaz: No need for this test.

.github/workflows/main.yml
```diff
@@ -100,6 +125,10 @@ jobs:
...
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: website...
 
       - name: Build website
-        run: pnpm run build-website-lite
+        env:
+          DISABLE_SHIKI: "true"
+        run: pnpm -F website run build
```
diegohaz: `pnpm run build-website-lite` still works under `filters: website...`, so this rewrite isn't needed, and inlining it splits the definition of a "lite" website build across three places.

`pnpm install` always materializes the workspace root importer regardless of which `--filter` selectors it receives, so `cross-env` and every root script stay available in a filtered job. Verified on a clean checkout of this head:

```
$ pnpm install --frozen-lockfile --filter 'website...'
$ ls node_modules/.bin/cross-env
node_modules/.bin/cross-env
$ pnpm run build-website-lite
$ cross-env DISABLE_SHIKI=true pnpm run build-website
$ pnpm -F website run build
$ next build
...
# exit 0, full Next.js production build
```

The cost of inlining is that `DISABLE_SHIKI: "true"` now lives here, in `.github/workflows/plus.yml` (lines 45-48), and in the root `build-website-lite` script. A future change to what "lite" means — another env var, a different build entry — has to be made in all three, and nothing fails if two of them drift.

Suggested revert, here and in `plus.yml`:

```suggestion
      - name: Build website
        run: pnpm run build-website-lite
```

The same reasoning applies to the `docs.yml` command change; I replied with the evidence on your existing thread there.

**Assessment**

- Severity: Low (estimate, 85% confidence), because drift yields a differently-configured website build rather than a broken one.
- Priority: P2 (estimate, 80% confidence), because the current form is correct today.
- Breakage likelihood: Possible (estimate, 70% confidence), because it requires someone to change the "lite" definition in only one of the three places.
- Fix effort: Easy (verified), a two-line revert in each workflow.

packages/ariakit-scripts/src/ci.test.ts
```diff
@@ -613,6 +618,79 @@ test("parses the repository lockfile", () => {
...
+
+  expect
+    .soft(getJobBlock("main.yml", "lint"))
+    .toMatch(/filters: \|\n\s+root\n\s+app\.\.\./);
+  expect
+    .soft(getJobBlock("app.yml", "build-nextjs"))
+    .toContain("filters: app...");
+  expect
+    .soft(getJobBlock("perf.yml", "build-nextjs"))
+    .toContain("filters: app...");
```
diegohaz: This test asserts three literal filter strings and one `package.json` key. It never computes a dependency closure, so it cannot fail for the class of bug it is named for, and it does fail for a strictly more correct edit.

**It cannot catch the bug it protects against.** `filters: app...` covers the two `build-nextjs` jobs only because `app/package.json` declares `"nextjs": "workspace:*"` in its devDependencies:

```
$ pnpm --filter 'app...' list --depth -1
... app@0.0.1 ... nextjs@0.0.2 ...
```

Remove that devDependency and both `build-nextjs` jobs lose their `nextjs` install, while `toContain("filters: app...")` still passes.

**It fails for a correct edit.** Making the closure explicit — `filters: |` / `app...` / `nextjs...`, which is exactly what `deploy.yml` and `deploy-preview.yml` already do — turns the text into `filters: |\n            app...`, so `toContain("filters: app...")` no longer matches even though the selection is strictly more correct. The test currently pins the repository to the less explicit of its own two conventions.

**One assertion pins a verified no-op.** `root` in the `lint` filters selects nothing that `app...` does not already install; `--filter app...` and `--filter root --filter app...` produce identical 1387-package installs (details in my comment on `.github/workflows/main.yml`). The `/filters: \|\n\s+root\n\s+app\.\.\./` regex locks that line in place.

The property worth protecting is "each job's filter selection contains every workspace its commands target". That is computable in-process from `pnpm-workspace.yaml` plus the workspace manifests, and it belongs in `ci.ts` with real unit tests — `ariakit-general-testing` prefers observable behavior over file-content matching, and neither of these two new tests exercises `ci.ts` at all.

If that is more than this PR should carry, deleting this test loses little: the three gaps it commemorates are already held by `website/package.json` declaring `guide`, and by the jobs themselves failing loudly when a closure is short.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because a regression test that cannot detect its own regression invites trust it cannot back.
- Priority: P1 (estimate, 85% confidence), because the test is new in this PR and pins two forms the repository may want to change.
- Breakage likelihood: Likely (verified for the explicit-filters case), because any move toward `nextjs...` or away from the bare `root` line fails this test.
- Fix effort: Easy (verified) to delete; Moderate (estimate, 75% confidence) to implement a real closure check in `ci.ts`.

.github/workflows/main.yml
```diff
@@ -19,6 +19,10 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: |
+            root
+            app...
```
diegohaz: `root` here is a no-op. `pnpm install` always materializes the workspace root importer regardless of `--filter`, so this selector adds nothing to what `app...` already installs.

Measured on clean checkouts of this head with pnpm 11.17.0 and `--frozen-lockfile`, where "packages" counts entries under `node_modules/.pnpm`:

| filters | packages | `node_modules` |
| --- | --- | --- |
| _(none)_ | 1757 | 2.2 GB |
| `root` | 598 | 1.0 GB |
| `root...` | 598 | 1.0 GB |
| `./packages/*...` | 598 | 1.0 GB |
| `website...` | 1122 | 1.4 GB |
| `app...` | 1387 | 1.8 GB |
| `root` + `app...` | 1387 | 1.8 GB |

`diff` over the sorted `.pnpm` listings is empty for `app...` versus `root` + `app...`, and empty again for `root` versus `root...` — the package sets are byte-identical. The root manifest devDepends on every `@ariakit/*` workspace package, so those are always materialized too; that is why `root`, `root...`, and `./packages/*...` all land on the same 598-package install.

So the two spellings in this file imply a precision the filters do not have. `root...` appears in nine other filter blocks and `root` in exactly this one, which makes the lint job the odd one out:

```suggestion
          filters: app...
```

Where a root selector is the only entry — `build`, `test-react18`, `release`, `release-preview`, and perf's `node` — it should stay: there it is what makes the install partial rather than full, even though the `...` suffix is decorative.

Note that `keeps CI job dependency closures complete` asserts this exact two-line form, so it has to change alongside.

**Assessment**

- Severity: Low (verified), because there is no behavioral difference.
- Priority: P2 (estimate, 85% confidence), because this is a readability and mental-model issue rather than a defect.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because it only bites when someone edits filters reasoning from the implied model.
- Fix effort: Easy (verified), one line here plus the matching test assertion.

.github/workflows/docs.yml
diegohaz: Confirmed — `pnpm run docs` still works under `filters: ./packages/*...`, so the `run:` change isn't needed.

`pnpm install` always materializes the workspace root importer no matter which `--filter` selectors it receives, so root scripts and root binaries stay available in every filtered job. Verified on a clean checkout of this head:

```
$ pnpm install --frozen-lockfile --filter './packages/*...'
Scope: 14 of 21 workspace projects
...
Done in 5.8s using pnpm v11.17.0
$ pnpm_config_verify_deps_before_run=false pnpm run docs
...
packages/ariakit-react-store docs: Done
# exit 0, git status clean
```

Two related measurements from the same runs: `--filter './packages/*...'` and `--filter root` produce byte-identical 598-package installs (1.0 GB), so the selector here could equally be `root`; and an unfiltered install is 1757 packages (2.2 GB), so the filter is doing real work either way.

Reverting to `run: pnpm run docs` keeps the root script as the single definition of what generating docs means, and keeps the diff to the `with:` block. The same applies to the website build in `main.yml` and `plus.yml`, where `pnpm run build-website-lite` also still works under `website...` — I left a separate comment there.

.github/workflows/app.yml
```diff
@@ -43,6 +45,8 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: app...
```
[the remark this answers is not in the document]
diegohaz: This one doesn't hold. `app...` does select `nextjs`, because `app/package.json` declares `"nextjs": "workspace:*"` in its devDependencies and pnpm's `...` suffix expands through dev dependencies:

```
$ pnpm --filter 'app...' list --depth -1
... app@0.0.1 ... nextjs@0.0.2 ...
```

Switching to `filters: nextjs...` as suggested would break this job the other way: `nextjs/app/globals.css` imports `../../app/src/styles/global.css`, whose Tailwind imports resolve from the `app` directory, and `nextjs...` does not select `app`. `App / Build Next.js app` passes on this head with `filters: app...`.

There is a real point nearby, though, addressed to the author rather than to this suggestion. This job's command is `pnpm -F nextjs run build-worker`, so its install leans entirely on an incidental `app -> nextjs` devDependency edge. Drop that edge and this job plus `perf.yml`'s `build-nextjs` lose their `nextjs` install, with no local signal that the filters were the cause. `deploy.yml` and `deploy-preview.yml` already spell both out:

```yaml
filters: |
  app...
  nextjs...
```

so the repository now carries two conventions for the same closure. Listing both here and in `perf.yml` costs nothing at install time — `nextjs...` is a strict subset of `app...`, so the materialized package set is unchanged — and makes the job's own target explicit. Whichever direction is chosen, the two should match. Note that the new `keeps CI job dependency closures complete` test asserts `toContain("filters: app...")` for both jobs, so it would need updating if they move to the explicit form.

**Assessment** (for the consistency finding, not the suggestion above)

- Severity: Medium (estimate, 75% confidence), because losing the edge breaks two artifact-producing jobs and blocks deploys and perf runs behind them.
- Priority: P2 (estimate, 70% confidence), because the edge is real today and the failure would surface loudly in CI.
- Breakage likelihood: Unlikely (estimate, 80% confidence), because it requires someone to remove `nextjs` from `app`'s devDependencies.
- Fix effort: Easy (verified), two filter lines plus the matching test assertion.

diegohaz (commented):
**Initial review** of `optimize-ci-installs` at `7d29346`, diffed from the merge base `a65885b`. Scope: all 16 changed files, plus the surrounding workflow, workspace-graph, and test context. No prior review from this workflow exists on this PR, so this is a full review rather than an incremental one.

## What I ran

Rebuilt every filter selection the workflows use on clean checkouts of this head (pnpm 11.17.0, `--frozen-lockfile`). "Packages" counts entries under `node_modules/.pnpm`:

| filters | packages | `node_modules` |
| --- | --- | --- |
| _(none)_ | 1757 | 2.2 GB |
| `root` | 598 | 1.0 GB |
| `root...` | 598 | 1.0 GB |
| `./packages/*...` | 598 | 1.0 GB |
| `website...` | 1122 | 1.4 GB |
| `app...` | 1387 | 1.8 GB |
| `root` + `app...` | 1387 | 1.8 GB |

Also: `pnpm --filter <f> list --depth -1` for `root...`, `app...`, `website...`, `nextjs...`, `./packages/*...`, `./templates/*...`; `pnpm run docs` on a clean `./packages/*...` install; a complete `pnpm run build-website-lite` on a clean `website...` install; `pnpm -F app run build-styles` with no workspace install; inspection of `verify-deps-before-run` handling in the pinned pnpm binary; and a per-job duration comparison against `main` at `f75f0d7`.

## Verified as correct

- **The `verifyDepsBeforeRun` opt-out is justified, not defensive noise.** pnpm 11's built-in default is `verify-deps-before-run: "install"`, so without the override a later `pnpm run`/`pnpm exec` really can replace a filtered install with a full one. `pnpm_config_verify_deps_before_run` is also the working spelling — `pnpm config get verify-deps-before-run` returns `false` with that prefix and `undefined` with `npm_config_`.
- **`install: "false"` for `build-styles` is safe.** `app/src/lib/build-styles.ts` imports only `node:fs/promises`, `node:path`, and relative modules; it completes against a checkout with no `app/node_modules`.
- **Dropping the setup step from perf's `node-baseline` is safe.** That job only runs `git worktree`, `cp`, and `tar`.
- **`./templates/*...` in the type-check job is required.** `tsconfig.node.json` includes `**/*` and excludes only `templates/*/dist`, so `templates/react/src` is in the `tsc -b` graph.
- **Wiring is complete.** Every `./.github/workflows/setup` caller passes `filters` or `install`; every `install-playwright` caller passes `workspace`; `app` and `website` each declare `@playwright/test`.
- **No changeset needed** — CI infrastructure plus a `workspace:*` entry in the private `website` workspace.

## Measured effect

Per-job wall clock, this PR's run against the same jobs on `main` at `f75f0d7`:

| job | main | PR |
| --- | --- | --- |
| Build styles | 45s | 13s |
| Generate docs | 66s | 38s |
| Build app | 106s | 79s |
| Build Next.js app | 82s | 67s |
| Test React 18 | 224s | 205s |
| Build packages | 68s | 57s |
| Test Solid | 44s | 34s |
| Build website | 192s | 184s |
| Test | 117s | 113s |
| Type check | 71s | 67s |
| Lint | 57s | 52s |
| Test React | 175s | 175s |
| Test Plus | 360s | 393s |

The store-cache removal looks net positive: the PR run had no store cache at all and still beat the cached `main` baseline in 12 of 13 jobs. `Test Plus` is the single regression — one sample, plausibly noise, but worth a second look once a few more runs land.

## Findings published inline

1. `.github/workflows/main.yml` — the website job's `run:` rewrite is unnecessary. `pnpm run build-website-lite` works under `filters: website...` (verified end to end), and inlining spreads `DISABLE_SHIKI` across `main.yml`, `plus.yml`, and the root script.
2. `packages/ariakit-scripts/src/ci.test.ts` — `keeps CI job dependency closures complete` asserts literal filter strings rather than closures. It cannot fail for the bug it names (dropping `app`'s `nextjs` devDependency breaks both `build-nextjs` jobs while the test stays green) and it does fail for a strictly more correct edit.
3. `.github/workflows/main.yml` — the `root` entry in the lint job's filters is a verified no-op; `--filter app...` and `--filter root --filter app...` produce byte-identical package sets.

## Validation of relevant unresolved threads

Each was re-derived against this head rather than taken at face value.

- **`main.yml` L99-100, CodeRabbit, `persist-credentials: false`** — *confirmed, but not this PR's scope.* Registered as ariakit/ariakit#6897. The gap is real and pre-existing: of 28 `actions/checkout` steps on this head, 17 set `persist-credentials: false` and 11 do not — all 5 in `main.yml`, all 3 in `app.yml`, and the single ones in `plus.yml`, `deploy.yml`, and `release-preview.yml`. (An earlier revision of this summary said 13; the correct count of checkouts already setting it is 17. The set of workflows missing it is unchanged.) `ci.yml` triggers on `pull_request`, so PR-controlled lifecycle scripts do execute in these jobs. This PR does not touch the checkout steps or worsen the exposure.
  - Severity: Medium (estimate, 70% confidence), because the persisted token is writable for same-repo branch PRs.
  - Priority: P2 (estimate, 80% confidence), because it predates this PR and applies repository-wide.
  - Breakage likelihood: Rare (estimate, 75% confidence), because it needs a malicious or compromised dependency lifecycle script.
  - Fix effort: Easy (estimate, 90% confidence), a `with:` block on each remaining checkout.
- **`ci.test.ts` L647-651, CodeRabbit, "assert a valid scoped-install configuration"** — *valid but subsumed.* The assertion does accept a `filters:` key with no value, and `install: "true"`, both of which permit a full install. It is moot because the author's own thread on L621 proposes removing that test and I agree with that call. No separate action.
- **`ci.test.ts` L621, author, "No need for this test."** — *agreed, with one addition.* Beyond redundancy, `expect(setupAction).toContain('args+=(--filter "$filter")')` pins the shell implementation so any rewrite of the loop fails without a behavior change, and `expect(setupAction).not.toMatch(/^\s+cache:/m)` is a landmine: it would fail for any future `cache:` key in that action, including `actions/setup-node`'s own `cache:` input, which has nothing to do with the pnpm store.
- **`main.yml`, Codex, "Include app dependencies in the lint install" (outdated)** — *already resolved* in `7d29346`. The lint job now selects `app...`, and `Main / Lint` passes in 52s.
- **`app.yml`, Codex, "Install app dependencies for standalone Next.js builds" (outdated)** — *already resolved.* Both `build-nextjs` jobs now select `app...` and pass.
- **`main.yml` L129, Codex, "Include the guide workspace in website installs"** — *already resolved.* `website/package.json` now declares `guide: workspace:*`; `pnpm --filter 'website...' list --depth -1` lists `guide@0.0.0`, and a full `pnpm run build-website-lite` succeeds on a clean `website...` install.
- **`app.yml` L49, CodeRabbit, "Install the Next.js workspace dependencies"** — *invalid.* `app...` already selects `nextjs` through `app`'s `"nextjs": "workspace:*"` devDependency, and the suggested `nextjs...` would break the job the other way, since `nextjs/app/globals.css` imports `app/src/styles/global.css`. Replied on the thread with the selection evidence, plus a separate confirmed finding about the implicit `app -> nextjs` edge and the two competing filter conventions.

## Not fully validated

- `release.yml` and `deploy.yml` only run on `main` pushes, so their `root...` and `app... nextjs...` filters were not exercised by this PR's CI. I reasoned through `changeset version` / `changeset publish` and `build-app` against the measured selections, but did not execute them.
- `app.yml`'s Safari, iOS, Firefox, and Chrome jobs were still pending when I reviewed. Store caching was already disabled on macOS before this PR (`cache: ${{ runner.os != 'macOS' }}`), so removing `cache:` should be a no-op there, but that is inference rather than an observed run.

## Registered follow-ups

Both are confirmed, independent of this change, and tracked separately rather than folded into this PR:

- ariakit/ariakit#6898 — every filtered job still pays for the always-installed workspace root importer: 598 packages and 1.0 GB out of a full 1757 and 2.2 GB. With this PR merged that floor becomes the dominant remaining install cost, and the largest further win would come from moving root `devDependencies` into the workspaces that actually consume them.
- ariakit/ariakit#6897 — the `persist-credentials: false` gap described in the thread validation above.

.github/workflows/main.yml
```diff
@@ -62,22 +72,35 @@ jobs:
...
+            filters: root...
           - name: Solid
             script: test-solid
+            filters: |
+              root...
+              app...
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and registered as ariakit/ariakit#6897 rather than folded into this PR.

The exposure is real: `ci.yml` triggers on `pull_request` and reaches this workflow through `workflow_call`, and the `Set up environment` step below runs `pnpm install`, whose lifecycle scripts come from the PR's own dependency graph. For same-repo branch PRs the persisted token is writable.

What the suggestion misses is that this is a repository-wide gap, not a `main.yml` one. Measured across the 28 `actions/checkout` steps in `.github/workflows` on `7d29346`:

- 17 already set `persist-credentials: false` — every checkout in `ci.yml`, `perf.yml`, `visual.yml`, `docs.yml`, `og-images.yml`, `build-styles.yml`, `release.yml`, and `deploy-preview.yml`
- 11 do not — all 5 here, all 3 in `app.yml`, and one each in `plus.yml`, `deploy.yml`, and `release-preview.yml`

So the convention already exists and these five files are the exception. Patching only this one step would leave the same exposure in four other workflows reached by the same `pull_request` trigger, which is why it is better done as one sweep.

This PR does not add, move, or otherwise touch any checkout step, so it neither introduces nor worsens the issue — hence the follow-up rather than a change here.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because the persisted token is writable for same-repo branch PRs.
- Priority: P2 (estimate, 80% confidence), because it predates this PR and applies repository-wide.
- Breakage likelihood: Rare (estimate, 75% confidence), because it requires a malicious or compromised dependency lifecycle script.
- Fix effort: Easy (verified), a `with:` block on each of the 11 remaining checkouts. None of the five affected workflows runs `git push`, `git commit`, or `git tag` from the checked-out worktree, and `deploy.yml`'s `secrets.GITHUB_TOKEN` use is an explicit `gitHubToken:` action input, independent of the persisted credential.

.github/workflows/main.yml
diegohaz: Following up on the measurement above, since this is where the evidence lives: the floor it exposes is registered as ariakit/ariakit#6898.

The reason `root`, `root...`, and `./packages/*...` all collapse to the same 598-package install is twofold — `pnpm install` always materializes the workspace root importer whatever `--filter` it gets, and the root manifest devDepends on all 14 `@ariakit/*` workspace packages, so those come along too. Once this PR lands, that 598-package / 1.0 GB floor is the dominant remaining install cost, and every job pays it no matter how narrow its filters are. It carries root devDependencies most jobs never touch: `next`, `wrangler`, `@playwright/test`, `stylelint`, `oxlint` and `oxlint-tsgolint`, `tailwindcss`, `happy-dom`, `@vitest/coverage-v8`, `pkg-pr-new`, `open-cli`, and the `@changesets/*` tooling.

That is independent of this change and out of scope for it. Two notes for whoever picks up #6898, both of which touch the suggestion above:

- The root devDependency on all 14 `@ariakit/*` packages is precisely what makes `root` and `root...` interchangeable today. Narrow it and those two selectors stop being equivalent, so the filter blocks in this file need revisiting at the same time.
- Some workspace packages currently resolve `solid-js`, `react`, and workspace links through the root `node_modules` rather than their own, helped along by `publicHoistPattern: ["*react*", "*testing-library*"]`. Moving dependencies out of root can turn those incidental resolutions into failures — most visibly in type-aware `oxlint`, which lints `packages/**` and `examples/**` but not `website`.

None of that blocks the one-line cleanup suggested above; it just means the cleanup is cosmetic today and would become load-bearing after #6898.

packages/ariakit-scripts/src/ci.test.ts
```diff
@@ -613,6 +613,46 @@ test("parses the repository lockfile", () => {
...
+      setupCallCount += 1;
+      const setupBlock = workflow
+        .slice(setupCall.index)
+        .split("\n")
+        .slice(0, 3);
+      expect(setupBlock, filename).toEqual([
+        setupCall[0],
+        expect.stringMatching(/^\s+with:$/),
+        expect.stringMatching(/^\s+(?:filters|install):/),
+      ]);
```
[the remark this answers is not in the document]
diegohaz: Superseded by [`706ed4584`](https://github.com/ariakit/ariakit/commit/706ed4584971d558c64e53868202aaed888fe805): the YAML-text test and its permissive assertion were removed instead of hardening a test that did not prove the scoped-install contract.

.github/workflows/docs.yml
diegohaz: Fixed in [`0b365bcca`](https://github.com/ariakit/ariakit/commit/0b365bccab72afed15c14bf10852bf3ffa7608bd): restored `pnpm run docs` while keeping the filtered package install, so the root script remains the single docs definition.

packages/ariakit-scripts/src/ci.test.ts
diegohaz: Fixed in [`706ed4584`](https://github.com/ariakit/ariakit/commit/706ed4584971d558c64e53868202aaed888fe805): removed the unnecessary YAML-text setup-caller test.

.github/workflows/main.yml
```diff
@@ -19,6 +19,8 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: root
```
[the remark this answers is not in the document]
diegohaz: Resolved in [`7d2934607`](https://github.com/ariakit/ariakit/commit/7d2934607692ec5f48fd94d272e0b7118121de80): the lint install now includes the app dependency closure; the subsequent cleanup keeps the effective selector as `app...`.

.github/workflows/main.yml
```diff
@@ -100,6 +123,10 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: website...
```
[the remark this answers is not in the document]
diegohaz: Resolved in [`7d2934607`](https://github.com/ariakit/ariakit/commit/7d2934607692ec5f48fd94d272e0b7118121de80): declared `guide` as a website workspace dependency, so `website...` includes it automatically.

.github/workflows/app.yml
```diff
@@ -43,6 +45,8 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: nextjs...
```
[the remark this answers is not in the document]
diegohaz: Resolved in [`7d2934607`](https://github.com/ariakit/ariakit/commit/7d2934607692ec5f48fd94d272e0b7118121de80): both standalone Next.js jobs now install `app...`, which includes `nextjs` and the shared app-style dependencies.

.github/workflows/main.yml
diegohaz: Fixed in [`0b365bcca`](https://github.com/ariakit/ariakit/commit/0b365bccab72afed15c14bf10852bf3ffa7608bd): restored `pnpm run build-website-lite` in both website jobs, keeping the lite-build definition in the root script.

packages/ariakit-scripts/src/ci.test.ts
diegohaz: Fixed in [`706ed4584`](https://github.com/ariakit/ariakit/commit/706ed4584971d558c64e53868202aaed888fe805): removed the literal-filter closure test rather than preserving coverage that did not compute dependency closures.

.github/workflows/main.yml
diegohaz: Fixed in [`0b365bcca`](https://github.com/ariakit/ariakit/commit/0b365bccab72afed15c14bf10852bf3ffa7608bd): removed the redundant `root` selector; a clean `app...` install produced the same package set and passed lint. The broader root dependency floor remains tracked in [#6898](https://github.com/ariakit/ariakit/issues/6898).

diegohaz:
**Fixed**

- [`706ed4584`](https://github.com/ariakit/ariakit/commit/706ed4584971d558c64e53868202aaed888fe805) ([setup-test reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663335865), [closure-test reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663339545)): removed both brittle YAML-text tests.
- [`0b365bcca`](https://github.com/ariakit/ariakit/commit/0b365bccab72afed15c14bf10852bf3ffa7608bd) ([docs reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663335145), [website reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663338727), [lint-filter reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663340262)): restored the root docs and lite-website scripts and removed the redundant lint selector.

**Already resolved**

- [`7d2934607`](https://github.com/ariakit/ariakit/commit/7d2934607692ec5f48fd94d272e0b7118121de80) ([lint reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663336433), [website reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663337225), [Next.js reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663337997)): completed the app, website/guide, and standalone Next.js dependency closures.

**Obsolete**

- [`706ed4584`](https://github.com/ariakit/ariakit/commit/706ed4584971d558c64e53868202aaed888fe805) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663334408)): the permissive scoped-install assertion was removed with its YAML-text test.

.github/workflows/setup/action.yml
```diff
@@ -1,19 +1,42 @@
...
+      env:
+        FILTERS: ${{ inputs.filters }}
+      run: |
+        args=(--frozen-lockfile)
+        while IFS= read -r filter; do
+          if [ -n "$filter" ]; then
+            args+=(--filter "$filter")
+          fi
+        done <<< "$FILTERS"
+        pnpm install "${args[@]}"
```
diegohaz: With the YAML-text tests gone in `706ed458` — which I agree with — nothing now catches a `setup` caller that forgets `filters`. The loop adds no selectors and this falls through to a plain `pnpm install --frozen-lockfile`, a silent full install.

That is a new failure mode rather than a pre-existing one. Before this PR every caller installed the whole workspace by design, so there was nothing to forget. Now a job that omits `filters` still passes CI and still produces correct output; it just quietly pays the full 1757-package / 2.2 GB install instead of its closure, with no error, no warning, and nothing visible in the diff to review.

On direction: my objection to the deleted test was that it matched YAML text from an unrelated package and asserted this action's shell implementation, not that the contract itself is worthless. Enforcing it in the action is the opposite trade — it covers every caller including future ones, needs no string matching, and fails at the point of use:

```suggestion
        args=(--frozen-lockfile)
        while IFS= read -r filter; do
          if [ -n "$filter" ]; then
            args+=(--filter "$filter")
          fi
        done <<< "$FILTERS"
        if [ ${#args[@]} -eq 1 ]; then
          echo "::error::setup requires 'filters', or install: \"false\" to skip installing" >&2
          exit 1
        fi
        pnpm install "${args[@]}"
```

Checked under GitHub's composite shell (`bash --noprofile --norc -eo pipefail`): empty `FILTERS` exits 1 with the annotation, while `app...` and a multi-line block scalar both build the expected argument list.

Deciding that the permissive default is intentional and recording that in a comment would resolve this equally well. What I would avoid is leaving it implicit now that the only guard is gone.

**Assessment**

- Severity: Low (verified), because the fallback yields a correct but slower install, never a wrong one.
- Priority: P2 (estimate, 80% confidence), because no current caller is affected — all 20 pass `filters` or `install`.
- Breakage likelihood: Possible (estimate, 65% confidence), because it requires someone to add a `setup` caller and omit the input.
- Fix effort: Easy (verified), the four lines above.

.github/workflows/main.yml
```diff
@@ -36,6 +38,12 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: |
+            root...
+            app...
+            website...
+            ./templates/*...
```
diegohaz, line 45: `root...` is redundant here for exactly the reason the `root` line was redundant in the lint job: pnpm always materializes the workspace root importer, and the root manifest devDepends on all 14 `@ariakit/*` packages, so this selector adds nothing on top of `app...` and `website...`.

It is still present in four filter blocks:

- line 43, here in `typescript`
- line 76, `test`
- line 82, `test-react`
- line 93, `test-solid`

With `lint` fixed and these four left in place, `main.yml` now reads as though `root...` matters in some jobs but not in others — which is the misleading signal the lint change removed. My earlier comment anchored only on `lint` and listed which sites should keep a root selector, so this is finishing that thought rather than reversing it.

```suggestion
          filters: |
            app...
            website...
            ./templates/*...
```

The sole-entry cases are genuinely different and should stay: `build` (line 62) and `test-react18` (line 89) have `root...` as their only selector, which is what makes those installs partial instead of full. Same for `release.yml`, `release-preview.yml`, and perf's `node` job.

**Assessment**

- Severity: Low (verified), because there is no behavioral difference in any of the four jobs.
- Priority: P2 (estimate, 85% confidence), because this is consistency and readability rather than a defect.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because the risk is a future filter edit reasoned from the inconsistency.
- Fix effort: Easy (verified), one line removed from each of the four blocks.

diegohaz (commented):
**Re-review** at `0b365bc`. Baseline is `7d29346`, the head recorded in this workflow's previous review marker, so the incremental diff covers `706ed458` (Remove brittle CI configuration tests) and `0b365bcc` (Simplify scoped CI workflows). Newly discovered findings come from those two commits plus surrounding code they expose; every relevant unresolved thread was independently re-validated against this head.

## Previous findings: all four resolved

| finding | resolution | evidence |
| --- | --- | --- |
| `docs.yml` `run:` rewrite unnecessary | `0b365bcc` restored `pnpm run docs` | `Docs / Generate docs` passes in 45s |
| `main.yml` + `plus.yml` website `run:` rewrite unnecessary | `0b365bcc` restored `pnpm run build-website-lite` in both | `Main / Build website` 194s, `Plus / Test Plus` 369s |
| `keeps CI job dependency closures complete` is a change detector | `706ed458` removed it, along with `keeps filtered CI installs partial` | `packages/ariakit-scripts/src/ci.test.ts` is now byte-identical to the merge base |
| `root` in the lint filters is a no-op | `0b365bcc` reduced it to `filters: app...` | `Main / Lint` passes in 54s |

I re-checked that the two restored commands still work under the filters they now run with, rather than relying on the CI green alone — both were verified end to end on clean filtered installs during the previous review, and the filters (`./packages/*...`, `website...`) are unchanged by these commits.

All 20 `setup` callers still pass `filters` or `install`, and all four `install-playwright` callers still pass `workspace`, so removing the tests did not leave a live gap.

## New findings

1. `.github/workflows/setup/action.yml` — with the tests removed, a caller that omits `filters` silently falls through to a full install. New failure mode: the input did not exist before this PR. Suggested a four-line fail-fast in the action, validated under GitHub's composite shell.
2. `.github/workflows/main.yml` — `root...` is still redundant in four filter blocks (`typescript`, `test`, `test-react`, `test-solid`) for the same reason it was in `lint`. Fixing only `lint` leaves the file implying the selector matters in some jobs and not others.

## PR description is now stale

The Solution section still claims the change "adds regression coverage requiring every setup caller to provide filters or explicitly opt out, while preserving the discovered cross-workspace job closures". `706ed458` removed that coverage, and `ci.test.ts` is no longer touched by this PR at all. The Verification list also still cites `pnpm test packages/ariakit-scripts/src/ci.test.ts`, which now exercises a file this branch does not change. Worth correcting before merge — there is no code line to anchor this to, hence the summary.

The Solution section's example block also shows `root...` alongside `app...`, which reads as endorsing the redundancy in finding 2.

## Measured effect at this head

Per-job wall clock against `main` at `f75f0d7`, with the previous PR head for reference. Deltas under roughly 10s on a single sample are noise, not signal.

| job | main | `7d29346` | `0b365bc` |
| --- | --- | --- | --- |
| Build styles | 45s | 13s | 12s |
| Generate docs | 66s | 38s | 45s |
| App / Build app | 106s | 79s | 82s |
| App / Build Next.js app | 82s | 67s | 62s |
| Build packages | 68s | 57s | 52s |
| Build website | 192s | 184s | 194s |
| Lint | 57s | 52s | 54s |
| Test | 117s | 113s | 107s |
| Test React | 175s | 175s | 185s |
| Test React 18 | 224s | 205s | 202s |
| Test Solid | 44s | 34s | 36s |
| Type check | 71s | 67s | 72s |
| Plus / Test Plus | 360s | 393s | 369s |
| App / Test Chrome | 464s | — | 452s |
| App / Test Firefox | 483s | — | 482s |
| App / Test Safari | 831s | — | 613s |
| App / Test iOS | 180s | — | 120s |
| App / Test Android | 70s | — | 53s |

Two gaps from the previous review now close:

- **The `Test Plus` regression was noise.** 360s on `main`, 393s at `7d29346`, 369s here. No action needed.
- **The macOS jobs are the biggest win and were previously unmeasured.** Safari 831s to 613s and iOS 180s to 120s. Store caching was already disabled on macOS before this PR (`cache: ${{ runner.os != 'macOS' }}`), so that gain is attributable purely to the smaller install, which is the cleanest evidence in the run that the filtering itself pays off. The full browser matrix is green at this head.

## Thread validation

All 11 review threads are resolved as of this head, including the two follow-up threads. Spot-checked rather than taken at face value:

- `main.yml` `persist-credentials` and the root-install floor remain accurately captured in ariakit/ariakit#6897 and ariakit/ariakit#6898; neither is affected by the new commits.
- The `app.yml` `build-nextjs` consistency point I raised on the CodeRabbit thread — `app.yml` and `perf.yml` rely on the implicit `app -> nextjs` edge while `deploy.yml`, `deploy-preview.yml`, and perf's `build-baseline` spell both out — is unchanged at this head. Reading that as a deliberate decision to keep the shorter form; not re-raising it, recorded here only so the record is complete.

## Not validated

`release.yml` and `deploy.yml` still only run on `main` pushes, so their `root...` and `app... nextjs...` filters remain unexercised by PR CI. Reasoned through against the measured selections, not executed.

.github/workflows/setup/action.yml
diegohaz: Fixed in [`aa9fec8ec`](https://github.com/ariakit/ariakit/commit/aa9fec8ec51a3ca3983280037b65ec35ad0301c9): the setup action now rejects enabled installs without a non-empty filter and passes `--fail-if-no-match`, so missing or wholly invalid selector sets cannot silently install the full workspace. Evidence gap: no durable action-level regression test was added because the shell is embedded in YAML; matching that text would recreate the brittle configuration coverage removed from this PR, while extracting a script or adding an expected-failure workflow job would add disproportionate machinery. The exact YAML-extracted shell was verified against empty, single, multiline, and unmatched filter cases.

.github/workflows/main.yml
diegohaz: Fixed in [`d61383b27`](https://github.com/ariakit/ariakit/commit/d61383b2733cbe1ff3f102deffca33ebd1a38652): removed the four redundant `root...` entries while preserving the multi-workspace seeds and the sole-entry build and React 18 selectors. Clean frozen installs with and without those redundant selectors produced the same 1,755 dependency entries and identical `node_modules` path trees.

diegohaz:
**Fixed**

- [`aa9fec8ec`](https://github.com/ariakit/ariakit/commit/aa9fec8ec51a3ca3983280037b65ec35ad0301c9) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663682175)): the shared setup action now rejects missing filters and wholly unmatched selector sets instead of silently running a full install.
- [`d61383b27`](https://github.com/ariakit/ariakit/commit/d61383b2733cbe1ff3f102deffca33ebd1a38652) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663684581)): removed the four redundant `root...` selectors while preserving the jobs' effective dependency closures.

**Evidence gaps**

- [`aa9fec8ec`](https://github.com/ariakit/ariakit/commit/aa9fec8ec51a3ca3983280037b65ec35ad0301c9) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663682175)): no durable action-level negative-path test was added because it would require brittle YAML-text matching or disproportionate new script/workflow machinery; the exact extracted action shell was verified directly instead.

.github/workflows/main.yml
```diff
@@ -36,6 +38,11 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: |
+            app...
+            website...
+            ./templates/*...
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because a filtered `pnpm install` still materializes the root importer and links its declared dependencies even though `pnpm --filter ... list` does not report `root` as a selected project. On this exact head, [Main / Type check](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210840949) installed with only `app...`, `website...`, and `./templates/*...`, then completed the root `pnpm run tsc -v` command successfully.

.github/workflows/main.yml
```diff
@@ -62,22 +71,30 @@ jobs:
       matrix:
         include:
           - script: test
+            filters: |
+              app...
+              website...
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because the root importer remains materialized during filtered installs, so root-owned `vitest` and `cross-env` binaries are linked even when `root` is absent from `pnpm --filter ... list`. On this exact head, [Main / Test](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210841032) passed 710 tests with `app...` and `website...`, [Main / Test React](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90214108284) passed 944 tests with `app...`, and [Main / Test Solid](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210841031) passed all 8 tests with `app...`.

.github/workflows/main.yml
```diff
@@ -100,6 +117,8 @@ jobs:
 
       - name: Set up environment
         uses: ./.github/workflows/setup
+        with:
+          filters: website...
```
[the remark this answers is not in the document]
diegohaz: This is not an issue because `website...` installs still materialize the root importer and its `cross-env` dependency. On this exact head, both [Main / Build website](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210840972) and [Plus / Test Plus](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210840955) installed with only `website...`, then executed the root `pnpm run build-website-lite` script through `cross-env` and completed successfully.

diegohaz:
**Invalid**

- [`No change`](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210840949) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663821395)): the exact Type check job passed with `app...`, `website...`, and template filters because filtered installs still materialize root importer dependencies.
- [`No change`](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90214108284) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663823974)): default, React, and Solid suites all accessed root-owned test tooling and passed without redundant root selectors.
- [`No change`](https://github.com/ariakit/ariakit/actions/runs/30339213473/job/90210840972) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3663826266)): Main and Plus website jobs both ran the root `build-website-lite` script through `cross-env` and passed with only `website...`.

.github/workflows/setup/action.yml
```diff
@@ -1,19 +1,46 @@
...
+        while IFS= read -r filter; do
+          if [ -n "$filter" ]; then
+            args+=(--filter "$filter")
+          fi
+        done <<< "$FILTERS"
+        if [ "${#args[@]}" -eq 1 ]; then
+          echo "::error::setup requires 'filters', or install: \"false\" to skip installing" >&2
+          exit 1
+        fi
+        pnpm install --fail-if-no-match "${args[@]}"
```
diegohaz, line 46: `--fail-if-no-match` is narrower than it reads: it fails only when the **entire** selection is empty, not when an individual selector matches nothing. Verified against this workspace:

```
$ pnpm install --frozen-lockfile --lockfile-only --fail-if-no-match \
    --filter 'app...' --filter 'does-not-exist'
No projects matched the filters "does-not-exist" in "…/ariakit"
Scope: 15 of 21 workspace projects
exit=0

$ pnpm install --frozen-lockfile --lockfile-only --fail-if-no-match \
    --filter 'does-not-exist' --filter 'another-missing'
No projects matched the filters in "…/ariakit"
exit=1
```

So a stale entry inside a multi-selector block is reduced to a log line and the job installs a smaller closure than intended. Five blocks are multi-selector today: `main.yml` `typescript` and `test`, `deploy.yml`, `deploy-preview.yml`, and `perf.yml`'s `build-baseline`. The realistic trigger is a workspace rename or removal where three of four references get updated — the job then fails at its build or test step with a module-resolution error rather than at the install with the actual cause.

The PR description's "wholly invalid selector set" is accurate about this. The reassessment section's "or when no project matches" reads broader than the flag actually behaves, which is the part worth pinning down.

Two ways to close it, and documenting the boundary is a perfectly good outcome:

- Validate each selector before installing, which keeps the error at the install step and names the offending filter:

  ```bash
  while IFS= read -r filter; do
    if [ -n "$filter" ]; then
      pnpm --filter "$filter" --fail-if-no-match list --depth -1 >/dev/null
      args+=(--filter "$filter")
    fi
  done <<< "$FILTERS"
  ```

- Or leave it and note in the comment above that per-selector staleness is deliberately out of scope, consistent with treating unsynchronized job seeds as unsupported.

**Assessment**

- Severity: Low (verified), because the job fails rather than producing a wrong artifact — just later and with a worse message.
- Priority: P2 (estimate, 80% confidence), because no selector is stale today and the guard already covers the wholly-empty case.
- Breakage likelihood: Possible (estimate, 60% confidence), because it needs a workspace rename or removal that misses one of the five multi-selector blocks.
- Fix effort: Easy (estimate, 85% confidence), one extra command inside the existing loop, or a comment.

.github/workflows/perf.yml
```diff
@@ -439,11 +443,11 @@ jobs:
           # so husky's prepare step would otherwise rewrite `core.hooksPath`
           # to a relative path under a temporary worktree.
           HUSKY: 0
-        run: pnpm install --frozen-lockfile --prefer-offline
+        run: pnpm --filter 'app...' install --frozen-lockfile --prefer-offline
```
diegohaz: This install and the two other direct ones in this file don't get the `--fail-if-no-match` guard that `aa9fec8e` added to the setup action, even though this PR is what converted all three from unfiltered to filtered:

- line 446, here — `pnpm --filter 'app...' install --frozen-lockfile --prefer-offline`
- line 760 — `pnpm --filter 'root...' install --frozen-lockfile --prefer-offline`
- lines 200-205 — `pnpm --filter 'app...' --filter 'nextjs...' install --frozen-lockfile`

The two baseline installs are the ones worth fixing, and they fix cleanly. Both are single-selector, so the flag's whole-selection semantics is exactly sufficient — unlike the multi-selector blocks, where it only catches the all-empty case (I left a separate note on the setup action about that boundary).

They also run in a worktree checked out at `github.event.pull_request.base.sha`, not at the PR head. That is precisely where a selector can stop matching without anyone noticing: a PR that renames or removes a workspace updates the filters at its head, while these two commands still run the new filter against the old tree. Without the flag pnpm logs "No projects matched the filters", exits 0, and the baseline builds from a short install — so the perf comparison is against a baseline that was never built the same way as the candidate.

```suggestion
        run: pnpm --filter 'app...' --fail-if-no-match install --frozen-lockfile --prefer-offline
```

Same one-flag change at line 760. Line 200-205 can take it too, though there it only guards against both selectors going stale at once.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because a silently short baseline install skews a perf comparison that gates merges, rather than failing outright.
- Priority: P2 (estimate, 75% confidence), because no selector is stale today and the PR already improves on the previous unfiltered state.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because it requires a workspace rename or removal to land while perf runs against the older base tree.
- Fix effort: Easy (verified), one flag per command.

diegohaz (commented):
**Re-review** at `d61383b`. Baseline is `0b365bc`, the head recorded in this workflow's previous review marker, so the incremental diff covers `aa9fec8e` (Require filters for CI installs) and `d61383b2` (Remove redundant CI root filters). Newly discovered findings come from those two commits plus the surrounding code they expose; every relevant unresolved thread was re-validated against this head.

## Correction to my previous review

My last summary claimed the macOS jobs were "the biggest win" and called Safari 831s to 613s and iOS 180s to 120s "the cleanest evidence in the run that the filtering itself pays off". A second sample does not support that:

| job | `main` `f75f0d7` | `0b365bc` | `d61383b` |
| --- | --- | --- | --- |
| App / Test Safari | 831s | 613s | 820s |
| App / Test iOS | 180s | 120s | 168s |

The 613s and 120s readings were outliers on notoriously variable macOS runners, and I over-read them from a single sample. That conclusion has since been carried into this PR's description ("the clearest measured improvements were Safari from 831s to 613s and iOS from 180s to 120s"), so it is worth correcting there before merge.

What the three samples do support is a roughly fixed per-job reduction in install time — clearly visible in short jobs, lost in the noise in long ones:

| job | `main` `f75f0d7` | `0b365bc` | `d61383b` |
| --- | --- | --- | --- |
| Build styles | 45s | 12s | 11s |
| Generate docs | 66s | 45s | 45s |
| Build packages | 68s | 52s | 52s |
| App / Build app | 106s | 82s | 74s |
| App / Build Next.js app | 82s | 62s | 64s |
| Test React 18 | 224s | 202s | 202s |
| Test | 117s | 107s | 108s |
| Test Solid | 44s | 36s | 38s |
| Lint | 57s | 54s | 50s |
| Type check | 71s | 72s | 69s |
| Build website | 192s | 194s | 169s |
| Test React | 175s | 185s | 190s |
| Plus / Test Plus | 360s | 369s | 394s |
| App / Test Chrome | 464s | 452s | 465s |
| App / Test Firefox | 483s | 482s | 440s |
| App / Test Android | 70s | 53s | 57s |

That is still a real and worthwhile win — it is just an install-time constant, not a browser-test speedup. Note also that `main` has since merged #6899 (Reduce Safari browser test coverage), which this branch does not include, so any future Safari comparison against current `main` will not be like-for-like.

## Previous findings: both resolved, one exceeded

| finding | resolution | verification |
| --- | --- | --- |
| A caller omitting `filters` silently falls through to a full install | `aa9fec8e` added the fail-fast, and went further by adding `--fail-if-no-match` | all six filter forms used across the workflows pass with the flag; the guard rejects the empty case |
| `root...` still redundant in four filter blocks | `d61383b2` removed all four | see below |
| PR description stale | rewritten | the removed-coverage bullet is gone, the example no longer shows `root...`, Verification drops `ci.test.ts`, and the absence of durable coverage for the action's negative path is now stated explicitly with reasoning |

I re-derived the `root...` removal rather than trusting the claim. Installing the `typescript` job's filters with and without `root...` on clean checkouts of this head:

```
--filter root... --filter app... --filter website... --filter ./templates/*...   1757 packages, 2.2 GB
--filter app... --filter website... --filter ./templates/*...                    1757 packages, 2.2 GB
```

`diff` over the sorted `.pnpm` listings is empty, and the per-workspace `node_modules` trees are identical. Behavior-preserving, confirmed.

One incidental observation from that measurement: at 1757 packages, the `typescript` job materializes the entire workspace, so it is the one job whose filters buy nothing today. That matches its timings (71s / 72s / 69s). Not worth changing — the block still documents intent and would shrink if the graph does — but worth knowing before anyone reads a saving into it.

## New findings

1. `.github/workflows/setup/action.yml` — `--fail-if-no-match` fails only when the whole selection is empty, not per selector. A stale entry in a multi-selector block becomes a log line, and five blocks are multi-selector today. Either validate each selector in the existing loop or record the boundary in the comment; the reassessment section's "or when no project matches" currently reads broader than the flag behaves.
2. `.github/workflows/perf.yml` — the three direct `pnpm ... install` calls, all converted from unfiltered to filtered by this PR, don't carry the flag. The two baseline installs are single-selector, so one flag each fully protects them, and they run against the base-SHA worktree where a selector is most likely to go stale unnoticed.

## Thread validation

All 13 review threads are resolved at this head. Nothing re-raised:

- ariakit/ariakit#6897 and ariakit/ariakit#6898 remain accurate and unaffected by these commits; both are now linked from the PR description.
- The `app -> nextjs` implicit-edge point I raised earlier is explicitly adjudicated in the reassessment section ("remains accepted because it is valid today and any future removal fails loudly in CI"). Reading that as a decision, not an oversight.

## Also checked

`main` has advanced four commits since the merge base (`a65885b`), but none of them touches a file this PR changes, so there is no conflict or semantic-overlap risk. The full CI matrix is green at this head, including `Main / Test React`, which was still running when I started.

## Not validated

`release.yml` and `deploy.yml` still run only on `main` pushes, so their filters remain unexercised by PR CI. Reasoned through against the measured selections, not executed.

.github/workflows/setup/action.yml
diegohaz: Not changing this in the current PR because validating every selector independently would add one extra pnpm process per selector to every multi-selector install for an unsupported unsynchronized workspace seed. The PR description now clarifies that `--fail-if-no-match` validates the combined selector set, while per-selector staleness in a partially valid set remains out of scope.

.github/workflows/perf.yml
diegohaz: Fixed in [`dc0cde0c1`](https://github.com/ariakit/ariakit/commit/dc0cde0c101517f4897767d7612d36c975d8ab3d): all three direct filtered installs in the performance workflow now pass `--fail-if-no-match`, so the single-selector app and root baseline installs fail immediately when their base worktree no longer contains the selected workspace. Evidence gap: no durable workflow-level regression test was added because asserting CLI flags from YAML would recreate the brittle configuration-text coverage removed from this PR; the missing-selector command was red/green checked directly and all real selectors were verified.

diegohaz:
## Review follow-up

- **Fixed** [`dc0cde0c1`](https://github.com/ariakit/ariakit/commit/dc0cde0c101517f4897767d7612d36c975d8ab3d) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3664016891)): added `--fail-if-no-match` to all three direct filtered installs in the performance workflow, including both base-worktree baseline installs.
- **Declined** [`combined-selector boundary`](https://github.com/ariakit/ariakit/pull/6894) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3664014711)): per-selector validation inside otherwise valid multi-selector sets remains out of scope; the PR description now documents the exact behavior and unsupported case.
- **Evidence gaps** [`dc0cde0c1`](https://github.com/ariakit/ariakit/commit/dc0cde0c101517f4897767d7612d36c975d8ab3d) ([reply](https://github.com/ariakit/ariakit/pull/6894#discussion_r3664016891)): no YAML-text regression test was added because it would recreate the brittle configuration coverage removed from this PR; the missing-selector behavior was checked red/green directly and every real selector set was verified.

diegohaz (commented):
**Re-review** at `dc0cde0c`. Baseline is `d61383b`, the head recorded in this workflow's previous review marker, so the incremental diff covers `dc0cde0c1` (Guard filtered performance installs). No new code findings this round.

## Correction: my package counts were inflated by 2

Every package count I have reported on this PR came from `ls node_modules/.pnpm | wc -l`, which also counts two non-package entries — the hoisted `node_modules` directory and `lock.yaml`:

```
$ ls node_modules/.pnpm | wc -l                                  2036
$ ls node_modules/.pnpm | grep -vE '^(node_modules|lock\.yaml)$' | wc -l   2034
```

So the correct figures are two lower across the board:

| filters | reported | correct |
| --- | --- | --- |
| _(none)_ | 1757 | **1755** |
| `root` / `root...` / `./packages/*...` | 598 | **596** |
| `website...` | 1122 | **1120** |
| `app...` and `root` + `app...` | 1387 | **1385** |
| `typescript` filters, with and without `root...` | 1757 | **1755** |

No conclusion changes: every comparison was relative, and the byte-identical `diff` results that justified removing `root` and `root...` are unaffected. But the figures did propagate, and the PR description is now internally inconsistent because of it — the Verification bullet correctly says "1,755 dependency entries", while the reassessment section still carries my inflated "1,757 packages … 598 … 1,122 … 1,387". Worth applying the same −2 there. The same inflated table appears in my earlier inline comment on `main.yml`; I have left that thread alone since it is resolved and outdated, and this summary supersedes it.

## Previous findings: both closed

- **`perf.yml` direct installs unguarded** — fixed in `dc0cde0c1`. All three now carry the flag, including the folded-scalar one, which renders as `pnpm --filter 'app...' --filter 'nextjs...' --fail-if-no-match install --frozen-lockfile`. I re-checked that flag-before-command placement is accepted by pnpm: a valid selector exits 0, a non-matching one exits 1. Four guarded installs total — `perf.yml` lines 205, 447, 761, and `setup/action.yml` line 46.
- **`--fail-if-no-match` per-selector boundary** — declined, with reasoning I accept: per-selector validation would add one pnpm process per selector to every multi-selector install, for a case already classified as unsupported. That was one of the two outcomes I offered. The boundary is now durably recorded in two places, which is what matters: the Solution section states that the flag "validates the combined pnpm selection; it does not validate every selector independently when at least one selector matches", and the reassessment section lists "per-selector staleness inside an otherwise valid multi-selector block" as intentionally unsupported.

The reassessment section also picked up the corrected timing framing — "Repeated samples support a roughly fixed per-job install-time reduction that is clearest in short jobs; long browser jobs vary enough that their total runtimes are not evidence of a browser-test speedup" — which resolves the over-read I corrected last round.

## Independent confirmation of three rebutted bot findings

Three P1-badged Codex threads landed at 07:46 claiming `filters` omitting `root` would break the type-check, test-matrix, and website jobs with `tsc: command not found`, missing `vitest`/`cross-env`, and a missing `cross-env`. All three are resolved with CI evidence. Recording that I independently confirm the rebuttals, since three P1 claims of guaranteed CI failure warrant more than a passing CI run.

The mechanism, measured in my first review on clean checkouts: `pnpm install` always materializes the workspace root importer regardless of `--filter`. Under `--filter 'website...'` alone, root-only devDependencies including `oxlint`, `stylelint`, `vitest`, `happy-dom`, `next`, `pkg-pr-new`, and `open-cli` are all present in the root `node_modules`, and `node_modules/.bin/cross-env` exists — I ran a complete `pnpm run build-website-lite` on exactly that install. The bot's inference from `pnpm --filter ... list` is the error: that command reports the *selection*, not what gets materialized, and root is never in the selection because it is never filtered in.

## Also checked

All CI checks pass at this head. `main` has advanced further, but the merge base is still `a65885b` and none of main's commits touches a file this PR changes, so there is no conflict or semantic-overlap risk.

## Not validated

`release.yml` and `deploy.yml` still run only on `main` pushes, so their filters remain unexercised by PR CI. Reasoned through against the measured selections across these reviews, never executed.

PR #6896, opened 2026-07-28
Prevent disabled Select values from submitting

### LINKED ISSUE #6892: Disabled Select and ComboboxSelect submit their form values
## Description

While auditing callback and prop reads after `props` reassignment for #6863, @georgekaran found a separate form-submission bug in `Select` and `ComboboxSelect`.

Both components render a hidden native `select` for form submission. That element currently reads `props.disabled` after `useFocusable`/`useWrapElement` has transformed and reassigned `props`:

```tsx
<select
  name={name}
  disabled={props.disabled}
  value={value}
/>
```

When `accessibleWhenDisabled` is used, or the rendered element does not support the native `disabled` attribute, the transformed props can omit `disabled`. The hidden native select then remains enabled and submits its value even though the Ariakit control is disabled. The adjacent dependency array reads the same transformed value.

## Expected behavior

Capture the consumer-provided `disabled` value before props are transformed and use that value for the hidden native select and its wrapper dependency array in both `Select` and `ComboboxSelect`.

Regression coverage should verify that disabled controls do not contribute values to form submission, including `accessibleWhenDisabled` for both components.

## Context

This is independent of `ComboboxPopover#getPersistentElements`, so #6889 can resolve #6863 without including it.

George provided a proof implementation and tests in [`c507bb68`](https://github.com/georgekaran/ariakit/commit/c507bb68a5e3846e36207b57485dadf83eb35c21).

### Workaround

https://github.com/ariakit/ariakit/issues/6892#issuecomment-5100003749

diegohaz:
## Motivation

Disabled [`Select`](https://ariakit.com/reference/select) and `ComboboxSelect` controls can still contribute their hidden native select values to form submission when `accessibleWhenDisabled` is used or the visible control renders an element that does not support the native `disabled` attribute.

Fixes #6892.

## Solution

Resolve the consumer-provided disabled state before Ariakit transforms the visible element props in both `Select` and `ComboboxSelect`:

```tsx
const disabledProp = disabledFromProps(props);
```

This captures both `disabled` and `aria-disabled`. The hidden native select and its wrapper dependency array use that effective state, so the native form control remains disabled when `accessibleWhenDisabled` or a custom rendered element removes the native `disabled` prop from the visible control, without regressing the supported `aria-disabled` path.

Browser-first and happy-dom regressions cover both components through `accessibleWhenDisabled`, `aria-disabled`, and `render={<div />}` after the disabled state changes.

## Workaround

Omit `name` while the control is disabled so the hidden native select cannot participate in form submission:

```tsx
<Select
  // TODO: Remove this condition after https://github.com/ariakit/ariakit/issues/6892 is fixed.
  name={disabled ? undefined : "fruit"}
  disabled={disabled}
/>
```

Use the same conditional `name` with `ComboboxSelect`. This preserves the disabled and accessible behavior but temporarily removes the control from form submission only while disabled.

## Preview

Reproduction at [`e21ee71bf`](https://github.com/ariakit/ariakit/commit/e21ee71bf9b1efa2ce63358b0e71ddb40ef1322e) — after the two custom-rendered controls are toggled disabled, the four controls affected by the original bug still expose their names:

Disabled Select and ComboboxSelect values still appear after form submission

Workaround at [`bbe34b083`](https://github.com/ariakit/ariakit/commit/bbe34b083d7496cc7c8c3b3aa07210cdf4c5e089) — conditionally omitting `name` keeps those four disabled controls out of form submission:

Disabled Select and ComboboxSelect workaround reports no submitted values

Solution at [`ac02b7d77`](https://github.com/ariakit/ariakit/commit/ac02b7d7778f2e1f88e93387075a54512829823a) — the current fixture covers six disabled controls across `accessibleWhenDisabled`, `aria-disabled`, and custom rendering, and submits no values:

Six disabled Select and ComboboxSelect controls report no submitted values

## Acknowledgments

Thanks to [@georgekaran](https://github.com/georgekaran) for diagnosing the transformed-props cause and providing the proof implementation and tests that informed this fix.

### DIFF
--- a/.changeset/300-disabled-select-form-values.md
+++ b/.changeset/300-disabled-select-form-values.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed disabled [`Select`](https://ariakit.com/reference/select) and `ComboboxSelect` controls so their values are omitted from form submission when [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is used or the visible control is rendered with an element that doesn't support the native `disabled` attribute. Thanks to [@georgekaran](https://github.com/georgekaran).
+Fixed disabled [`Select`](https://ariakit.com/reference/select) and [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) controls so their values are omitted from form submission when [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is used or the visible control is rendered with an element that doesn't support the native `disabled` attribute. Thanks to [@georgekaran](https://github.com/georgekaran).
--- a/.changeset/300-disabled-select-form-values.md
+++ b/.changeset/300-disabled-select-form-values.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed disabled [`Select`](https://ariakit.com/reference/select) and `ComboboxSelect` controls so their values are omitted from form submission when [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is used or the visible control is rendered with an element that doesn't support the native `disabled` attribute. Thanks to [@georgekaran](https://github.com/georgekaran).
--- a/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
+++ b/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
@@ -1,31 +1,101 @@
 import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
 
 export default function Example() {
+  const [renderedDisabled, setRenderedDisabled] = useState(false);
+  const [submittedNames, setSubmittedNames] = useState<string[] | null>(null);
+
   return (
-    <form
-      onSubmit={(event) => {
-        event.preventDefault();
-        const data = new FormData(event.currentTarget);
-        window.alert(data.get("select"));
-      }}
-    >
-      <Ariakit.ComboboxProvider defaultSelectedValue="Apple">
-        <Ariakit.ComboboxSelectLabel>
-          Favorite fruit
-        </Ariakit.ComboboxSelectLabel>
-        <Ariakit.ComboboxSelect
-          name="select"
-          disabled
-          className="pointer-events-none"
-        />
-        <Ariakit.ComboboxPopover gutter={4} sameWidth unmountOnHide>
-          <Ariakit.ComboboxItem value="Apple" />
-          <Ariakit.ComboboxItem value="Banana" />
-          <Ariakit.ComboboxItem value="Grape" />
-          <Ariakit.ComboboxItem value="Orange" />
-        </Ariakit.ComboboxPopover>
-      </Ariakit.ComboboxProvider>
-      <Ariakit.Button type="submit">Submit</Ariakit.Button>
-    </form>
+    <>
+      <form
+        onSubmit={(event) => {
+          event.preventDefault();
+          const data = new FormData(event.currentTarget);
+          window.alert(data.get("select"));
+        }}
+      >
+        <Ariakit.ComboboxProvider defaultSelectedValue="Apple">
+          <Ariakit.ComboboxSelectLabel>
+            Favorite fruit
+          </Ariakit.ComboboxSelectLabel>
+          <Ariakit.ComboboxSelect
+            name="select"
+            disabled
+            className="pointer-events-none"
+          />
+          <Ariakit.ComboboxPopover gutter={4} sameWidth unmountOnHide>
+            <Ariakit.ComboboxItem value="Apple" />
+            <Ariakit.ComboboxItem value="Banana" />
+            <Ariakit.ComboboxItem value="Grape" />
+            <Ariakit.ComboboxItem value="Orange" />
+          </Ariakit.ComboboxPopover>
+        </Ariakit.ComboboxProvider>
+        <Ariakit.Button type="submit">Submit</Ariakit.Button>
+      </form>
+
+      <form
+        aria-label="Transformed disabled selects"
+        onSubmit={(event) => {
+          event.preventDefault();
+          const data = new FormData(event.currentTarget);
+          setSubmittedNames([...data.keys()]);
+        }}
+      >
+        <Ariakit.SelectProvider defaultValue="Apple">
+          <Ariakit.SelectLabel>Select accessible</Ariakit.SelectLabel>
+          <Ariakit.Select
+            name="select-accessible"
+            disabled
+            accessibleWhenDisabled
+          />
+        </Ariakit.SelectProvider>
+
+        <Ariakit.ComboboxProvider defaultSelectedValue="Apple">
+          <Ariakit.ComboboxSelectLabel>
+            Combobox accessible
+          </Ariakit.ComboboxSelectLabel>
+          <Ariakit.ComboboxSelect
+            name="combobox-accessible"
+            disabled
+            accessibleWhenDisabled
+          />
+        </Ariakit.ComboboxProvider>
+
+        <Ariakit.SelectProvider defaultValue="Apple">
+          <Ariakit.SelectLabel>Select rendered</Ariakit.SelectLabel>
+          <Ariakit.Select
+            name="select-rendered"
+            disabled={renderedDisabled}
+            render={<div />}
+          />
+        </Ariakit.SelectProvider>
+
+        <Ariakit.ComboboxProvider defaultSelectedValue="Apple">
+          <Ariakit.ComboboxSelectLabel>
+            Combobox rendered
+          </Ariakit.ComboboxSelectLabel>
+          <Ariakit.ComboboxSelect
+            name="combobox-rendered"
+            disabled={renderedDisabled}
+            render={<div />}
+          />
+        </Ariakit.ComboboxProvider>
+
+        <label>
+          <input
+            type="checkbox"
+            checked={renderedDisabled}
+            onChange={(event) => setRenderedDisabled(event.target.checked)}
+          />
+          Disable rendered selects
+        </label>
+        <button type="submit">Submit transformed selects</button>
+        <output>
+          {submittedNames
+            ? submittedNames.join(", ") || "No values submitted"
+            : "Not submitted"}
+        </output>
+      </form>
+    </>
   );
 }
--- a/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
+++ b/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
@@ -2,7 +2,6 @@ import * as Ariakit from "@ariakit/react";
 import { useState } from "react";
 
 export default function Example() {
-  const accessibleDisabled = true;
   const [renderedDisabled, setRenderedDisabled] = useState(false);
   const [submittedNames, setSubmittedNames] = useState<string[] | null>(null);
 
@@ -45,8 +44,8 @@ export default function Example() {
         <Ariakit.SelectProvider defaultValue="Apple">
           <Ariakit.SelectLabel>Select accessible</Ariakit.SelectLabel>
           <Ariakit.Select
-            name={accessibleDisabled ? undefined : "select-accessible"}
-            disabled={accessibleDisabled}
+            name="select-accessible"
+            disabled
             accessibleWhenDisabled
           />
         </Ariakit.SelectProvider>
@@ -56,8 +55,8 @@ export default function Example() {
             Combobox accessible
           </Ariakit.ComboboxSelectLabel>
           <Ariakit.ComboboxSelect
-            name={accessibleDisabled ? undefined : "combobox-accessible"}
-            disabled={accessibleDisabled}
+            name="combobox-accessible"
+            disabled
             accessibleWhenDisabled
           />
         </Ariakit.ComboboxProvider>
@@ -65,7 +64,7 @@ export default function Example() {
         <Ariakit.SelectProvider defaultValue="Apple">
           <Ariakit.SelectLabel>Select rendered</Ariakit.SelectLabel>
           <Ariakit.Select
-            name={renderedDisabled ? undefined : "select-rendered"}
+            name="select-rendered"
             disabled={renderedDisabled}
             render={<div />}
           />
@@ -76,7 +75,7 @@ export default function Example() {
             Combobox rendered
           </Ariakit.ComboboxSelectLabel>
           <Ariakit.ComboboxSelect
-            name={renderedDisabled ? undefined : "combobox-rendered"}
+            name="combobox-rendered"
             disabled={renderedDisabled}
             render={<div />}
           />
--- a/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
+++ b/app/src/sandbox/combobox-select-form-disabled/index.react.tsx
@@ -62,6 +62,18 @@ export default function Example() {
         </Ariakit.ComboboxProvider>
 
         <Ariakit.SelectProvider defaultValue="Apple">
+          <Ariakit.SelectLabel>Select aria disabled</Ariakit.SelectLabel>
+          <Ariakit.Select name="select-aria" aria-disabled />
+        </Ariakit.SelectProvider>
+
+        <Ariakit.ComboboxProvider defaultSelectedValue="Apple">
+          <Ariakit.ComboboxSelectLabel>
+            Combobox aria disabled
+          </Ariakit.ComboboxSelectLabel>
+          <Ariakit.ComboboxSelect name="combobox-aria" aria-disabled />
+        </Ariakit.ComboboxProvider>
+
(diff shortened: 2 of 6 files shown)


packages/ariakit-react-components/src/select/select.tsx
```diff
@@ -103,6 +103,7 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
   );
 
   const onKeyDownProp = props.onKeyDown;
+  const disabledProp = props.disabled;
```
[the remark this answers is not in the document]
diegohaz: Confirmed independently against the current head. I built a temporary sandbox rendering `<Ariakit.Select name="select-aria" aria-disabled />` and `<Ariakit.ComboboxSelect name="combobox-aria" aria-disabled />` inside a form that reports `[...new FormData(form).keys()]` after submit:

- base `f75f0d7f9`: `No values submitted`
- this PR head `9ff064e4e`: `select-aria, combobox-aria`

`disabledFromProps` (`packages/ariakit-utils/src/misc.ts`) treats `aria-disabled` as disabled, so `useFocusable` still renders the visible control with the native `disabled` attribute, `pointer-events: none`, and disabled click/keydown handlers. Only the hidden native select changed: it used to observe the transformed `disabled`, which `useFocusable` sets to `true` on this path (`disabled: supportsDisabled && trulyDisabled ? true : undefined`), and now observes the untransformed prop, which is `undefined`. So the PR closes the leak for `disabled` and opens the same leak for `aria-disabled`.

Real-world case: `aria-disabled` is a supported way to disable any Ariakit focusable component. `disabledFromProps` is part of the public `@ariakit/utils` surface and its documented parameter shape includes `"aria-disabled"?: boolean | "true" | "false"` (`packages/ariakit-utils/readme.md`). An app that disables a named `Select` with `aria-disabled={!canEdit}` keeps the control visibly and semantically disabled, but the form now submits its value, so the server receives data for a field the user could not edit.

Capturing `disabledFromProps(props)` covers both disabling paths:

```suggestion
  const disabledProp = disabledFromProps(props);
```

Add `disabledFromProps` to the existing `@ariakit/utils` import here, and apply the same change at `packages/ariakit-react-components/src/combobox/combobox-select.tsx:115`.

I applied that in both files locally and verified:

- `pnpm test-react run app/src/sandbox/combobox-select-form-disabled/test.ts` plus the temporary `aria-disabled` sandbox: all pass
- `pnpm test-react run app/src/sandbox app/src/examples`: 898 passed
- `pnpm exec tsc -b packages/ariakit-react-components`: clean

Worth extending the new regressions with an `aria-disabled` case in both `test-browser.ts` and `test.ts` so this path stays covered.

**Assessment**

- Severity: High (estimate, 85% confidence), because a disabled control contributes a value to form submission, the same data-integrity impact this PR fixes for the `disabled` path.
- Priority: P1 (estimate, 90% confidence), because the regression is introduced by this change in the exact code being fixed.
- Breakage likelihood: Possible (estimate, 75% confidence), because it requires disabling a named `Select` or `ComboboxSelect` through `aria-disabled` instead of `disabled`; that path is supported and styled, but `disabled` is the common one.
- Fix effort: Easy (verified, 95% confidence), because a one-line change per file plus the import and regression coverage was validated locally.

.changeset/300-disabled-select-form-values.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed disabled [`Select`](https://ariakit.com/reference/select) and `ComboboxSelect` controls so their values are omitted from form submission when [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is used or the visible control is rendered with an element that doesn't support the native `disabled` attribute. Thanks to [@georgekaran](https://github.com/georgekaran).
```
diegohaz: `ComboboxSelect` is the only public API named in this entry that ships without a reference link, while `Select` and `accessibleWhenDisabled` are both linked in the same sentence.

The API qualifies for a link: it is exported from `@ariakit/react` (`packages/ariakit-react/src/combobox.ts`), and two sibling entries in this same unreleased batch already link the future canonical URL, `.changeset/200-combobox-select.md` and `.changeset/300-combobox-labeling.md`. The deployed page returns `404` today, which is expected for a branch-new API and does not change the decision.

```suggestion
Fixed disabled [`Select`](https://ariakit.com/reference/select) and [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) controls so their values are omitted from form submission when [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is used or the visible control is rendered with an element that doesn't support the native `disabled` attribute. Thanks to [@georgekaran](https://github.com/georgekaran).
```

**Assessment**

- Severity: Low (verified, 95% confidence), because it only affects discoverability in the published changelog.
- Priority: P2 (estimate, 90% confidence), because it is a one-line fix in a file this PR adds, not a blocker.
- Breakage likelihood: Likely (verified, 99% confidence), because the entry ships as written in the next release.
- Fix effort: Easy (verified, 99% confidence), because it is a single Markdown link.

diegohaz (commented):
## Review summary

Initial review of all 6 changed files, diffed from the merge base `f75f0d7f9` to head `9ff064e4e`, expanded into `useFocusable`, `useWrapElement`, `disabledFromProps`, and the existing sandbox fixture contract.

The core diagnosis is right. `useWrapElement` memoizes its callback through `useCallback`, and that callback closes over the mutable `props` binding, so it observed whatever `props` had become after `usePopoverDisclosure` and `useCompositeTypeahead` reassigned it. Capturing the value before the transformation is the correct shape of the fix, and it is the only place in the packages where a `useWrapElement` callback reads a prop that a later hook in the same scope rewrites. I scanned every `useWrapElement` call in `packages/**/*.tsx` for prop reads inside the callback body; the only other hits are `props.id` in `portal.tsx`, which `usePortal` never rewrites after the wrapper, so no sibling fix is missing.

### Validation performed

Regression proof for the new coverage, by restoring the two package files to their `f75f0d7f9` versions and re-running:

- `pnpm test-react run app/src/sandbox/combobox-select-form-disabled/test.ts`: both new tests fail on the pre-fix source with `select-accessible, combobox-accessible, select-rendered, combobox-rendered` instead of the expected output, and pass at head.
- `pnpm -F app test --project=chrome src/sandbox/combobox-select-form-disabled/test-browser.ts`: both new browser tests fail on the pre-fix source and pass at head, run against a local Astro dev server from this checkout.

Other checks at head: `pnpm test-react run app/src/sandbox app/src/examples`, `pnpm lint`, `pnpm exec tsc -b packages/ariakit-react-components`.

### Unresolved candidate findings validated in this run

- `packages/ariakit-react-components/src/select/select.tsx:106`, thread `PRRT_kwDOBjT0Fc6UTzvk` (Codex, "Honor aria-disabled when disabling the hidden select"): **confirmed**. I reproduced it with a temporary sandbox and a base-versus-head comparison, then validated `disabledFromProps(props)` as the fix across both components, the full app happy-dom suite, and a package typecheck. Replied on the thread with the reproduction, the assessment, and the suggested change rather than opening a duplicate.

### Findings published inline

- The `aria-disabled` regression above, as a reply on the existing thread.
- `.changeset/300-disabled-select-form-values.md`: `ComboboxSelect` ships without a reference link while `Select` and `accessibleWhenDisabled` are linked in the same sentence, and sibling entries in this unreleased batch already link it.

### Checked and not flagged

- Fixture lifecycle: extending the existing semantic sandbox is the right step, the original scenario and its four tests are preserved, the query disambiguation to `q.combobox("Favorite fruit")` does not change an existing invariant, and the issue URL is repeated above each new test in both the browser test and the happy-dom duplicate.
- The dependency-array change from `props.disabled` to `disabledProp` is value-identical; the array was already evaluated before the transformation, so the behavioral fix is entirely in the callback body.
- Changeset frontmatter correctly versions `@ariakit/react-components` and `@ariakit/react`, uses the `300-` patch prefix, and credits the contributor per the single-paragraph credit form.
- The workaround commit `b42a907ed` is fully reverted at head; no conditional `name` or TODO remains in the sandbox.
- PR description claims match the implementation and the tests.

### Pre-existing observations, not caused by this PR

`app/src/sandbox/combobox-select/test.ts > Provider favorite fruit > opens and closes with the pointer and keyboard` fails on this baseline, expecting `aria-haspopup="dialog"` and receiving `"listbox"`. Follow-up work is registered as https://github.com/ariakit/ariakit/issues/6902. It is not flakiness: after filing, I confirmed on current `main` (`1c1cedaf2`) with a clean install that it fails 3 of 3 isolated runs, 3 of 3 `app/src/sandbox` runs, and the unfiltered `pnpm test-react` command, and that it already failed at `8c8c97207`, the commit that added the test. CI reports the same commit green, so the outcome differs by environment. None of it is affected by this PR.

A separate full-directory run on `f75f0d7f9` failed `app/src/sandbox/combobox-select-tag-tooltip-element-id/test.ts > updates the tooltip anchor when the content id changes` once and did not reproduce afterwards. That one does look like genuine flakiness and is noted in the issue above without being registered separately.

### Not fully validated

The browser test was run only under the `chrome` project against the Astro dev server; `firefox`, `safari`, `ios`, and `android` were not exercised locally. A full `pnpm tsc` could not complete in this environment because the app's generated Wrangler types are absent, so the package typecheck was scoped to `packages/ariakit-react-components`.

packages/ariakit-react-components/src/select/select.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`d95e47538`](https://github.com/ariakit/ariakit/commit/d95e47538765638fa6bd7bd793ea9c08f0a495f2): `Select` and `ComboboxSelect` now use `disabledFromProps(props)` for their hidden native selects, with targeted browser and happy-dom coverage for `aria-disabled` form submission.

.changeset/300-disabled-select-form-values.md
diegohaz: Fixed in [`ac02b7d77`](https://github.com/ariakit/ariakit/commit/ac02b7d7778f2e1f88e93387075a54512829823a): linked `ComboboxSelect` to its future canonical reference page in the changeset.

diegohaz:
**Fixed**

- [`d95e47538`](https://github.com/ariakit/ariakit/commit/d95e47538765638fa6bd7bd793ea9c08f0a495f2) ([reply](https://github.com/ariakit/ariakit/pull/6896#discussion_r3663866448)): honored `aria-disabled` for hidden `Select` and `ComboboxSelect` form controls and added browser/happy-dom coverage.
- [`ac02b7d77`](https://github.com/ariakit/ariakit/commit/ac02b7d7778f2e1f88e93387075a54512829823a) ([reply](https://github.com/ariakit/ariakit/pull/6896#discussion_r3663869538)): linked `ComboboxSelect` to its canonical reference page in the changeset.

app/src/sandbox/combobox-select-form-disabled/index.react.tsx
```diff
@@ -1,31 +1,113 @@
...
+          <Ariakit.ComboboxSelect
+            name="combobox-accessible"
+            disabled
+            accessibleWhenDisabled
+          />
+        </Ariakit.ComboboxProvider>
+
+        <Ariakit.SelectProvider defaultValue="Apple">
+          <Ariakit.SelectLabel>Select aria disabled</Ariakit.SelectLabel>
+          <Ariakit.Select name="select-aria" aria-disabled />
```
diegohaz, line 66: These two controls add a third disabling path to the fixture, and the PR description was not updated for it.

`## Solution` still describes the previous implementation: "Capture the consumer-provided `disabled` value before Ariakit transforms the visible element props in both `Select` and `ComboboxSelect`." The code now captures `disabledFromProps(props)`, which resolves `disabled` **and** `aria-disabled`, and that difference is the whole point of `d95e47538`. The same paragraph lists coverage as "through `accessibleWhenDisabled` and through `render={<div />}` after the disabled state changes", while the suite now covers three paths in both `test-browser.ts` and `test.ts`.

`## Preview` is stale in a way a reader cannot detect: all three captions count four controls ("all four leaked names", "all four controls remain disabled", "keeps all four controls disabled"), but this form now renders six named controls, and the recordings predate `select-aria` and `combobox-aria`. Someone comparing the evidence against the fixture will not find the two controls the latest commit exists to protect.

`ariakit-general-pr-description` asks for the description to be brought back in line after pushing commits that make it stale, so this is worth doing before merge rather than after.

**Assessment**

- Severity: Low (verified, 95% confidence), because it misleads readers and the release record without affecting shipped behavior.
- Priority: P2 (estimate, 80% confidence), because the description-currency rule applies to pushed commits, but no correctness or release gate depends on it.
- Breakage likelihood: Likely (verified, 99% confidence), because the description is already inaccurate at this head.
- Fix effort: Moderate (estimate, 70% confidence), because the prose edit is trivial but the three recordings need recapturing against the six-control fixture.

diegohaz (commented):
## Re-review summary

Re-review from the previously marked head `9ff064e4e` to the current head `ac02b7d77`. The branch was rebased in between, so the incremental scope is the two new commits, `d95e47538` and `ac02b7d77`, read against the new merge base `e1467694b` (which now includes #6900 and #6899). I also re-checked the surrounding code the rebase pulled in, since #6900 reworked several combobox modules; the PR's diff against the new merge base is still the same six-line package change plus the fixture, tests, and changeset.

### Both previous findings are fixed, verified independently

- `aria-disabled` regression (thread `PRRT_kwDOBjT0Fc6UTzvk`, now resolved): `d95e47538` switches both hooks to `disabledFromProps(props)`. Confirmed the capture still happens before any prop transformation in both files, and confirmed the behavior end to end rather than trusting the resolution.
- Changeset reference link (thread `PRRT_kwDOBjT0Fc6UUUUJ`, now resolved): `ac02b7d77` links `ComboboxSelect` to `https://ariakit.com/reference/combobox-select`, matching the sibling entries in this unreleased batch.

### Validation performed

Regression proof for the three tests, run in both suites:

- Against the merge-base `e1467694b` package sources, all three happy-dom tests and all three Chrome tests fail; at head all pass.
- Against the intermediate `158aed1b8` implementation, which used `props.disabled`, the tests still fail. That is the specific red state for the new `aria-disabled` coverage, so the new tests are proven against the exact defect they were added for, not only against the original bug.

Other checks at head: `pnpm test-react run app/src/sandbox app/src/examples` (897 passed, plus the known unrelated failure below), `pnpm lint`, `pnpm exec tsc -b packages/ariakit-react-components`, and the Chrome project for `test-browser.ts` against a local Astro dev server from this checkout.

`test-browser.ts` is still the right filename after #6899. That change narrowed specific tests to `test-chrome.ts` and `test-chrome-firefox.ts`, but 118 `test-browser.ts` files remain on `main`, including most of the `combobox-select-*` sandboxes, and disabled form submission is behavior worth running on every desktop project.

- The PR description was not updated for the `aria-disabled` path: `## Solution` still describes capturing "the consumer-provided `disabled` value", and all three `## Preview` captions count four controls while the fixture now renders six.

### Checked and not flagged

- `accessible disabled selects do not submit values` and `aria-disabled selects do not submit values` assert the same submitted-names output against the same form state, so the second test's submission assertion is redundant with the first. Keeping it is justified: it carries the `aria-disabled` ARIA-state assertions and gives that path its own issue-traceable title, and collapsing the two would lose both.
- The changeset text still describes only the `accessibleWhenDisabled` and unsupported-`disabled`-attribute paths, which is correct. `aria-disabled` alone was never broken on `main`; it only leaked in combination with those paths, which the entry already covers.
- `disabledFromProps` returns `false` rather than `undefined` when nothing is disabled. React renders no `disabled` attribute either way, so the hidden select is unchanged.

### Pre-existing, still failing, not caused by this PR

`app/src/sandbox/combobox-select/test.ts > Provider favorite fruit > opens and closes with the pointer and keyboard` still fails on this environment, as registered in https://github.com/ariakit/ariakit/issues/6902.

### Not fully validated

Browser coverage was exercised only under the `chrome` project; `firefox`, `safari`, `ios`, and `android` were not run locally. A full `pnpm tsc` still cannot complete here because the app's generated Wrangler types are absent, so the typecheck was scoped to `packages/ariakit-react-components`.

app/src/sandbox/combobox-select-form-disabled/index.react.tsx
diegohaz: Fixed in the [updated PR description](https://github.com/ariakit/ariakit/pull/6896#issue-4995455600): `## Solution` now documents `disabledFromProps(props)` and all three covered paths. The historical reproduction/workaround artifacts are tied to their exact commits, and the current-head solution artifact shows all six disabled controls submitting no values.

diegohaz:
**Fixed**

- [`PR description`](https://github.com/ariakit/ariakit/pull/6896#issue-4995455600) ([reply](https://github.com/ariakit/ariakit/pull/6896#discussion_r3664064808)): documented `disabledFromProps(props)` and all three regression paths, tied historical evidence to its exact commits, and added a current-head solution image showing all six disabled controls submitting no values.

Issue #6910, opened 2026-07-28
Drop the unused @vitejs/plugin-react devDependency from the website workspace


diegohaz:
## Context

Found while implementing #6898, which measures how much each CI job's filtered `pnpm install` materializes. The `website...` filter is the second-largest install after `app...`.

## Problem

`website/package.json` declares `"@vitejs/plugin-react": "6.0.1"` in `devDependencies`, but the `website` workspace never imports it. The only two occurrences in `website/` are inside generated StackBlitz template strings in `website/lib/stackblitz.ts`, which emit a different version into the sandbox's own manifest and vite config:

```ts
"@vitejs/plugin-react": "^3.0.0",
```

```ts
import react from "@vitejs/plugin-react";
```

Neither is a real import by the website. The website builds with Next.js 14 and webpack (`website/next.config.js`), not vite.

The declaration still pulls `vite@8.1.5` and its `lightningcss` binary into every install selected by `website...`, which is the filter used by the Build website job in `.github/workflows/main.yml` and by `.github/workflows/plus.yml`. It is also the source of the only lockfile churn in #6898 that did not map directly to a manifest edit: removing root's `tailwindcss@3` flipped which optional `jiti` peer pnpm attaches to the website's `vite@8.1.5` variant.

## Why this can be separate

Nothing is broken. `pnpm run build-website-lite` passes and the generated CSS is unaffected; this is an unused declaration, not a missing or wrong one.

## Expected outcome

Remove `@vitejs/plugin-react` from `website/package.json` if nothing in the workspace resolves it, and confirm the StackBlitz templates still emit the versions they intend. Verify with `pnpm run build-website-lite` and by re-measuring the `website...` install (packages under `node_modules/.pnpm` and `du -sh node_modules`) before and after.

Note that `renovate.json` disables `website/package.json` wholesale, so this entry is otherwise frozen and will not be cleaned up by dependency automation.

Issue #6909, opened 2026-07-28
Move @ariakit/react off the legacy scripts/build pipeline


diegohaz:
## Context

Found while implementing #6898, which reduces the root `devDependencies` that every filtered `pnpm install` materializes. `tsup` and `cross-spawn` had to stay at root because a root-level build script still uses them.

## Problem

`packages/ariakit-react` is the only workspace that still builds through the legacy root scripts:

```json
"build": "node ../../scripts/build/build.js",
"clean": "node ../../scripts/build/clean.js"
```

Every other package under `packages/` builds with `ariakit build` from `@ariakit/scripts`, which uses rolldown. Because `scripts/build/build.js` lives at the repository root, its imports — `tsup` (which brings its own `esbuild`), `cross-spawn`, and, through `scripts/build/utils.js`, `chalk`, `fs-extra`, and `rimraf` — must be declared in the root `package.json` and are therefore installed by every CI job.

The root `dev-package-json` script (`scripts/build/package-json-dev.js`) also imports `scripts/build/utils.js`, so `chalk`, `fs-extra`, `rimraf`, `chokidar`, and `glob` would still be needed at root unless that script is moved too. `tsup` and `cross-spawn` are the entries that only exist for `@ariakit/react`'s build.

## Why this can be separate

The current setup is correct and `pnpm run build` passes. This is about where the build tooling lives, not about broken output.

## Expected outcome

Investigate whether `@ariakit/react` can build with `ariakit build` like its sibling packages, or whether something about its re-export surface requires the tsup pipeline. If it can migrate, `scripts/build/build.js` and `scripts/build/clean.js` can go away and `tsup` and `cross-spawn` can leave the root manifest. If it cannot, the pipeline belongs in `@ariakit/scripts` or in `packages/ariakit-react` itself, so its dependencies are declared by the workspace that uses them rather than by root.

Verify the published output is unchanged either way, by comparing `npm pack --dry-run --json` contents and the built `dist` files before and after.

diegohaz:
We'll do this for v0.5.0 (breaking change because `ariakit build` is ESM-only).

Issue #6908, opened 2026-07-28
Narrow the root @ariakit/* devDependencies and make the workflow filters explicit


diegohaz:
## Context

Found while implementing #6898. That issue flagged this as something to watch:

> The root devDependency on all 14 `@ariakit/*` packages is exactly what makes `root` and `root...` equivalent today. If that narrows, several workflow filters stop being interchangeable and will need revisiting.

The #6898 work deliberately left it alone so the dependency removal and the workflow-filter change could be evaluated separately.

## Problem

The root `package.json` devDepends on all 14 `@ariakit/*` workspaces, but root-level code only uses two:

- `@ariakit/scripts` provides the `ariakit` bin used by the root `dev-app`, `react18`, and `perf-compare` scripts.
- `@ariakit/test` is imported by `vitest.setup.ts` as `@ariakit/test/react`.

The other 12 exist only as workspace links. They are also what makes `root`, `root...`, and `./packages/*...` resolve to the same install, which several workflows rely on today:

- `.github/workflows/main.yml` — the Build packages job (`root...`, runs `pnpm run build` → `pnpm -F "@ariakit/*" run build`) and the React 18 test job (`root...`).
- `.github/workflows/release.yml` and `.github/workflows/release-preview.yml` (`root...`, both run `pnpm run build`).
- `.github/workflows/perf.yml` — the Node benchmark job (`root...`).

Because the root importer is always materialized, `@ariakit/solid-store` in particular is only reachable through root: no other workspace depends on it, so the Type check job's `app... website... ./templates/*...` filter would stop installing it if root no longer did.

## Why this can be separate

The 12 unused entries are workspace links whose external footprint is small, so they are not what dominates the install floor. Removing them without also rewriting the filters would silently change what those jobs install.

## Expected outcome

Root declares only the `@ariakit/*` workspaces its own code and scripts use, and every workflow filter names the workspaces its job actually needs — for example `./packages/*...` for `pnpm run build` instead of `root...`. Verified by a CI run that actually exercises each affected job, not only by a local install, since these filters are the whole point.

Issue #6907, opened 2026-07-28
Declare workspace dependencies that currently resolve through the root node_modules


diegohaz:
## Context

Found while implementing #6898, which moves root `devDependencies` into the workspaces that consume them. Removing a package from the root manifest exposed that some workspaces import packages they never declare, and only work because Node's upward resolution finds them in the repository-root `node_modules`.

`pnpm-workspace.yaml` also sets `publicHoistPattern: ["*react*", "*testing-library*"]`, which hoists matching packages into the root `node_modules` and hides the same class of missing declaration for React-related packages.

## Problem

Confirmed cases on current `main`:

- `packages/ariakit-solid`, `packages/ariakit-solid-components`, `packages/ariakit-solid-store`, and `packages/ariakit-solid-utils` declare `solid-js` only as a `peerDependency`. Their sources import it, and it resolves from the root `devDependencies` entry that exists for `vitest.setup.ts`.
- `packages/ariakit-react-components`, `packages/ariakit-react-utils`, `packages/ariakit-solid-utils`, `packages/ariakit-test`, and `packages/ariakit-utils` import `vitest` in their own test files without declaring it. Only `packages/ariakit-components` and `packages/ariakit-store` declare it.

`packages/ariakit-test` had the same problem with `@playwright/test` and `@testing-library/react`; both were declared only as optional `peerDependencies` while `src/playwright.ts`, `src/react.tsx`, and `src/react.test.tsx` import them. The #6898 work had to fix those two, because removing them from root broke `pnpm run tsc` outright.

## Why this can be separate

`solid-js` and `vitest` both stay in the root `devDependencies` after #6898 — root's own `vitest.config.ts` and `vitest.setup.ts` import them — so nothing is broken today. The declarations are missing, not the packages.

## Expected outcome

Every workspace declares the packages its own sources and tests import, instead of relying on root hoisting or `publicHoistPattern`. This removes the hidden coupling that makes each further root-dependency reduction a potential type-check or test-run break, and it makes the failure mode obvious at install time rather than at the first CI job that runs under a narrower filter.

Worth verifying afterwards under the narrow CI filters (`./packages/*...`, `root...`, `website...`), not only under a full unfiltered install, since a full install masks exactly this problem.

PR #6911, opened 2026-07-28
Move root devDependencies to their consuming workspaces

### LINKED ISSUE #6898: Reduce the always-installed workspace root dependency set in CI
## Context

Found while reviewing #6894 (Scope CI dependency installs), which scopes each CI job's `pnpm install` to the workspaces that job actually needs. That change is a clear win; this issue is about the floor it runs into, not a defect in it.

## Problem

`pnpm install` always materializes the workspace root importer, whatever `--filter` selectors it receives. On top of that, the root `package.json` devDepends on every `@ariakit/*` workspace package, so all 14 of those are always materialized too.

Measured on clean checkouts at `7d29346` with pnpm 11.17.0 and `--frozen-lockfile`, counting entries under `node_modules/.pnpm`:

| filters | packages | `node_modules` |
| --- | --- | --- |
| _(none)_ | 1757 | 2.2 GB |
| `root` | 598 | 1.0 GB |
| `root...` | 598 | 1.0 GB |
| `./packages/*...` | 598 | 1.0 GB |
| `website...` | 1122 | 1.4 GB |
| `app...` | 1387 | 1.8 GB |
| `root` + `app...` | 1387 | 1.8 GB |

`root`, `root...`, and `./packages/*...` produce byte-identical 598-package installs, and adding `root` to `app...` changes nothing at all. So 598 packages and 1.0 GB — about a third of the full install by count, just under half by size — is a fixed cost every job pays regardless of its filters.

That floor includes root devDependencies most jobs never touch: `next`, `wrangler`, `@playwright/test`, `stylelint` and `stylelint-config-standard`, `oxlint` and `oxlint-tsgolint`, `tailwindcss`, `happy-dom`, `@vitest/coverage-v8`, `pkg-pr-new`, `open-cli`, and the `@changesets/*` tooling.

## Expected outcome

Move root `devDependencies` into the workspaces that consume them, so filtered installs actually shrink. Root would keep only what root-level scripts and configs need: `oxlint`/`oxfmt` for `pnpm run lint`, `vitest` and the test loaders for the root suites, `husky`/`lint-staged`, and the changesets tooling.

Two things to watch while doing this:

- `pnpm-workspace.yaml` sets `publicHoistPattern: ["*react*", "*testing-library*"]`, and some workspace packages currently resolve `solid-js`, `react`, and workspace links through the root `node_modules` rather than their own. Moving dependencies out of root can turn those incidental resolutions into failures — most visibly in type-aware `oxlint`, which lints `packages/**` and `examples/**` but not `website`.
- The root devDependency on all 14 `@ariakit/*` packages is exactly what makes `root` and `root...` equivalent today. If that narrows, several workflow filters stop being interchangeable and will need revisiting.

Worth re-measuring the table above after any change, since the meaningful metric here is the floor rather than the full install.

diegohaz:
## Motivation

`pnpm install` always materializes the workspace root importer and its dependencies, whatever `--filter` selectors it receives. Since #6894 scoped each CI job's install, that root importer became the floor every job pays regardless of what it actually needs, and most of that floor was tooling root itself never runs — `next`, `wrangler`, `@playwright/test`, and the legacy website's Tailwind v3 / PostCSS chain.

Measured on clean checkouts at `9001af55b`, counting entries under `node_modules/.pnpm`, `root` alone pulled 598 packages and 1.0 GB. Of that, `next` (with `@next/swc-darwin-arm64`) accounted for ~292 MB and `wrangler` (with `workerd` and `miniflare`) for ~269 MB, neither of which any root script invokes.

Closes #6898.

## Solution

Remove the root `devDependencies` that no root-level file, script, or config consumes, and declare each one in the workspace that actually uses it.

**Removed from root**: `@playwright/test`, `@testing-library/react`, `autoprefixer`, `next`, `postcss`, `postcss-import`, `prettier`, `prettier-plugin-tailwindcss`, `tailwindcss`, `tailwindcss-animate`, `wrangler`.

**`packages/ariakit-test`** gains `@playwright/test` and `@testing-library/react` as devDependencies. Its source imports both:

```ts
// packages/ariakit-test/src/react.tsx
import * as ReactTestingLibrary from "@testing-library/react";
```

They were declared only as optional `peerDependencies`, so they resolved by walking up into the root `node_modules`. Without the root entries, `pnpm run tsc` fails on `packages/ariakit-test/src/playwright.ts`.

**`nextjs`** gains `wrangler`. `app/playwright.config.ts` starts the Next.js preview with `pnpm -F nextjs exec opennextjs-cloudflare preview`, and `@opennextjs/cloudflare` shells out rather than importing:

```js
// @opennextjs/cloudflare/dist/cli/commands/utils/run-wrangler.js
spawnSync(options.packager, [..., "exec", "wrangler", ...], { shell: true })
```

With cwd `nextjs/`, that lookup previously only succeeded through the repository-root `.bin`. Declaring it in `nextjs` keeps it inside the `app...` and `nextjs...` filters that every job needing the preview server already uses.

**`website`** gains `autoprefixer` and `tailwindcss-animate`, and now owns the two root configs only it ever loaded — `postcss.config.cjs` and `tailwind.config.cjs` — plus the deletion of the root `prettier.config.js` that existed solely to point `prettier-plugin-tailwindcss` at that Tailwind config. Next.js found the root PostCSS config by searching upward from `website/`; it now finds it in the project directory. The Tailwind `content` globs are rebased on a `repoRoot` constant so they still scan `guide/`, `components/`, and `examples/` at the repository root.

With the root `postcss.config.cjs` moved into `website/`, the two empty shield configs — `app/postcss.config.js` and `templates/react/postcss.config.js`, both just `export default {}` — no longer shield anything inside the repository, so they are removed. `app/postcss.config.js` was added deliberately and documented in #6503, so this reverses that: the shields did still block a search that escapes the project directory into ancestors *outside* the checkout, and removing them is an explicit call to not carry that guard. `nextjs/postcss.config.js` and `website/postcss.config.cjs` remain, both real config rather than shields, and neither is an ancestor of `app/` or `templates/react/`.

**Kept deliberately**: `typescript-7`. `@typescript/typescript6` (aliased as `typescript`) only ships a `tsc6` bin, so root's `tsc -b` resolves `tsc` from `typescript-7`. Removing it broke `pnpm run tsc` outright.

`packages/ariakit-scripts/src/ci.ts` drops `@testing-library/react`, `prettier`, and `prettier-plugin-tailwindcss` from the `rootMainDependencies` allowlist, and `renovate.json` drops its root-scoped `postcss` rule, since those names can no longer appear in the root manifest.

### Dropping Tailwind class sorting in generated example source

`app/prettier.config.js` named `prettier-plugin-tailwindcss` with a bare specifier, and Prettier resolves those relative to `process.cwd()`, not to the config file. The root Vitest suite runs from the repository root, so the plugin had only ever loaded through the root `node_modules` entry removed above. `app/src/lib/source-plugin.ts` swallows formatting errors, so the breakage surfaced as unsorted classes in `Main / Test` rather than as an error.

Rather than re-plumb the resolution, the feature is dropped: examples now keep the class order their authors wrote, which is fine for displayed source. `prettier-plugin-tailwindcss` leaves the repository entirely.

`source-plugin.ts` also stops resolving a Prettier config. With none left in the repository, that lookup could only find a config *outside* the checkout — concretely, a nested worktree resolved the main checkout's copy of the very file this PR deletes — which would make the source we publish depend on the machine building it. The `filepath` option goes with it, since Prettier reads it only to infer a parser and the parser is always passed explicitly.

The test that covered this is split in two: one case pins the class order, and one asserts the merged import is re-wrapped. The second is the one that fails if formatting silently stops running, which the previous single assertion no longer would have caught.

### Moving the Vercel config into `website`

Removing `next` from the root manifest also broke the Vercel deployment. The project built with the repository root as its Root Directory — the old `vercel.json` set `"outputDirectory": "website/.next"` — so `@vercel/next` looked for the Next.js version in the root `package.json` and threw `No Next.js version detected` before installing anything.

The evidence was the timing of the `Vercel` commit status: pending to failure in 29s on `2bcdd41eb` and 25s on `587ad8dd6`, against pending to success in 3m55s on `9001af55b`. Under 30 seconds is far too short for `pnpm install` on this monorepo.

`vercel.json` therefore moves to `website/vercel.json`, where Vercel reads it once the project's Root Directory is `website`, and the Next.js version is found in `website/package.json`. The paths that assumed the repository root are rebased: `buildCommand` becomes `pnpm run build` (`website` has no `build-website` script) and `outputDirectory` becomes `.next`. `framework` is pinned as well, so all four build inputs stay declared in the file instead of falling back to project settings that cannot be reviewed here. `devCommand` is dropped and `vercel dev` is unsupported: nothing in the repository invokes it, and `concurrently` swallowed the `--port` flag rather than forwarding it, so the old value never worked. `headers` and `ignoreCommand` are unchanged.

The Root Directory has already been set to `website`, and the `Vercel` check passes on this branch as a result — a full 3m52s build on `edced7d1c`, against the 25-29s failures before. That green build also confirms the two things that could not be verified from CI: the diagnosis itself, and that **Include source files outside of the Root Directory** is enabled, since `website/next.config.js` and `website/tailwind.config.cjs` read `../packages`, `../guide`, `../examples`, and `../components`, and `website/build-pages/config.js` calls `readdirSync` on `../components` at module load.

**This should merge promptly.** Because the Root Directory now points at `website` while `main` still carries `vercel.json` at the repository root, a deployment from `main` before this merges would find no config and publish without the `frame-ancestors 'self'` and `X-Frame-Options: SAMEORIGIN` headers and without `ignoreCommand`. Production is currently still serving the last deployment built before the switch, so the headers are intact today; merging is what keeps them that way.

After merging, two checks remain:

1. Confirm the headers survive a fresh production deployment: `curl -sI https://ariakit.com | grep -iE 'x-frame-options|content-security-policy'`.
2. Confirm a `components/`-only commit still triggers a deployment. Vercel's automatic monorepo skip is expected to be superseded by the explicit `ignoreCommand`, but that is unverified. `components/` is the discriminating path: the build reads it, but it is not a pnpm workspace, unlike `guide/` and `examples/`, which are `workspace:*` dependencies of `website` and so would deploy under either mechanism.

Package-manager detection is not a concern: `@vercel/build-utils` walks up from the Root Directory recording `packageManager` until it finds a lockfile, so from `website/` it still reaches the root `pnpm-lock.yaml` and `"packageManager": "pnpm@11.17.0"` — the same inputs as before.

## Results

Same machine, `--frozen-lockfile`, counting entries under `node_modules/.pnpm` and `du -sh node_modules`. Both columns measured in this run; the before column reproduces the table in #6898 exactly.

| filters | before | after |
| --- | --- | --- |
| _(none)_ | 1757 · 2.2 GB | 1757 · 2.2 GB |
| `root` | 598 · 1.0 GB | **528 · 398 MB** |
| `root...` | 598 · 1.0 GB | **528 · 398 MB** |
| `./packages/*...` | 598 · 1.0 GB | **528 · 398 MB** |
| `website...` | 1122 · 1.4 GB | **1092 · 856 MB** |
| `app...` | 1387 · 1.8 GB | **1366 · 1.8 GB** |

The floor every job pays drops by 61% in size. The full install is unchanged, as expected: nothing left the dependency graph, it only moved to workspaces that filtered installs can skip.

No workflow filter needed to change. Root still declares all 14 `@ariakit/*` workspaces, so `root`, `root...`, and `./packages/*...` remain interchangeable.

## Verification

- `pnpm run lint`, `pnpm run lint-css`, `pnpm run tsc`, `pnpm test`, `pnpm test-react`, `pnpm test-react18`, `pnpm test-solid`, `pnpm run build` (then `pnpm clean`), `pnpm run build-website-lite`, and `pnpm run docs` all pass. `pnpm run docs` produces no diff.
- Building the app with and without the PostCSS shields emits byte-identical CSS — the same six files under `app/dist` with the same content hashes — as does a minimal project mirroring `templates/react`'s `vite` + `@tailwindcss/vite` stack. Both comparisons were run outside the repository tree so no ancestor config could influence them.
- The legacy website's generated CSS is byte-for-byte identical to `main` — same 8 content-hashed filenames, 148334 bytes total — so the PostCSS and Tailwind relocation is behavior-preserving rather than merely non-fatal.
- `pnpm install --frozen-lockfile` succeeds under each CI filter (`root...`, `app...`, `website...`, `./packages/*...`), and regenerating the lockfile from the manifests reproduces the committed `pnpm-lock.yaml` byte-for-byte.
- Test runs were repeated in a checkout outside the repository tree. A worktree nested inside another checkout inherits that checkout's `node_modules` through Node's upward resolution, which is exactly what hid the `prettier-plugin-tailwindcss` breakage locally until CI caught it.
- One `pnpm test-react` failure appeared locally on macOS (`app/src/sandbox/combobox-select/test.ts`, `aria-haspopup`). It reproduces identically at the base commit and is #6902, since fixed on `main` by #6904.

## Follow-ups

Registered rather than folded in, so each can be evaluated on its own:

- #6907 — workspaces that import `solid-js` and `vitest` without declaring them, relying on the root `node_modules` and on `publicHoistPattern`.
- #6908 — narrowing root's 14 `@ariakit/*` devDependencies, which requires rewriting every workflow filter that currently assumes `root...` selects all packages.
- #6909 — moving `@ariakit/react` off the legacy root `scripts/build` pipeline, which is the only reason `tsup` and `cross-spawn` remain at root.
- #6910 — the unused `@vitejs/plugin-react` devDependency that pulls `vite` into the `website...` filter.
- #6914 — a stale `mergeImports` JSDoc claim that, if acted on, would quietly weaken the formatting regression test added here.

One deliberate consequence worth noting: `autoprefixer` and `tailwindcss-animate` now live only in `website/package.json`, which `renovate.json` disables wholesale. That matches how the rest of the legacy website's dependencies are already treated.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the Root Directory move, but pin every build input in the file</summary>

**Assessment**

Three finding-bearing gate rounds on the `vercel.json` move, a four-line JSON diff. No round found a defect in the diff itself.

- Issue severity: inherited, not intrinsic. #6898 is CI cost, but merging this PR without the Vercel change leaves `main`'s deployment failing, so the site stops updating.
- Expected frequency: one-time. `vercel.json` was last touched in the 2023 pnpm migration.
- Supported scope: one Vercel project, one deployment target.
- Likely user impact: none for library consumers; misconfiguration risks wrong output, missing security headers, or stale content.
- Implementation complexity: near zero — one rename, no code, no tests, no public contract.
- Maintenance complexity: the first shape of this change was *worse* than the original, because it deleted `buildCommand` and `outputDirectory` and moved those inputs from a reviewable file into unversioned dashboard state.

The cycles had one cause: each round objected to a consequence of deleting a key the file previously pinned. Round 1 → `installCommand` restored. Round 2 → dropping `outputDirectory` delegated correctness to the dashboard Framework Preset, where preset `Other` would have served `website/public` statically and reported success while every page 404s; `framework` pinned. Round 3 → the `buildCommand` and `outputDirectory` dashboard override toggles were still live, unobservable inputs.

One finding per deleted key across three rounds. That is not an over-complex design; it is review converging on state no reviewer in CI can observe, which it cannot do by adding rounds.

**Decision**

Continue the direction — reverting to `next` in root `devDependencies` would give back ~292 MB of the ~600 MB floor reduction to avoid a one-time config change — but narrow it back toward a pure rename by pinning `framework`, `installCommand`, `buildCommand`, and `outputDirectory` in the file. This overrides round 3's advice against pinning `buildCommand`: writing `pnpm run build` rather than `next build` keeps `website/package.json` as the single source of truth while still defeating a stale dashboard override, and it retires the class of finding that drove all three rounds.

`devCommand` is intentionally dropped and `vercel dev` left unsupported. The remaining unobservable items were resolved by observation rather than more review rounds, which is what the reassessment called for: the Root Directory switch landed while this was in review, and the resulting green `Vercel` build confirmed both the diagnosis and the out-of-root file access. Only the automatic monorepo skip is still open, as a post-merge check.

</details>


packages/ariakit-test/package.json
```diff
@@ -32,6 +32,8 @@
   },
   "devDependencies": {
     "@ariakit/scripts": "workspace:*",
+    "@playwright/test": "1.62.0",
+    "@testing-library/react": "16.3.2",
```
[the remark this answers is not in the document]
diegohaz: Leaving `@testing-library/react` optional here is intentional, and it is unchanged by this PR — only the `devDependencies` block was touched.

`@ariakit/test` ships three entrypoints with different requirements:

- `.` (`src/index.ts`) needs only `@testing-library/dom`, which is a real `dependency`.
- `./playwright` needs `@playwright/test`.
- `./react` needs `@testing-library/react` and `react`.

Making any of those peers non-optional would force every consumer to install it regardless of which entrypoint they import. Someone using `@ariakit/test` for its DOM helpers, or `@ariakit/test/playwright` in a non-React project, would be told to install React Testing Library for code they never load. That is a strictly worse failure mode than the one described, and `@playwright/test` and `react` are optional for exactly the same reason.

The import-time failure you describe is the intended signal: it happens only when someone imports `@ariakit/test/react` without the peer installed, which is when the peer is genuinely required.

app/postcss.config.js
```diff
@@ -8,6 +8,6 @@
  * SPDX-License-Identifier: UNLICENSED
  */
 // Intentionally empty: prevents Vite's upward PostCSS config search from
-// picking up the legacy root postcss.config.cjs for website/.
+// picking up a config outside the app.
 // The app uses Tailwind v4 through @tailwindcss/vite instead.
 export default {};
```
diegohaz: Do we still need this file and nextjs/postcss.config?

app/src/lib/source-plugin.test.ts
```diff
@@ -104,9 +104,18 @@ test("disclosure source names", () => {
   `);
```
diegohaz, line 104: Unrelated to this file, but can you also re-check CI job dependencies? For example, does the `Post PR comment` job really need dependency on `app...`? Isn't `packages/ariakit-scripts` or something enough?

app/postcss.config.js
diegohaz: This is not an issue because both files are still doing work, though for different reasons.

`nextjs/postcss.config.js` is not a guard at all — it is the Next.js app's actual Tailwind v4 wiring:

```js
const config = {
  plugins: {
    "@tailwindcss/postcss": {},
  },
};
```

Removing it would drop Tailwind processing from the Next.js build entirely.

`app/postcss.config.js` is still a guard, but a narrower one than its original comment claimed, which is why this PR reworded it. The repo-root `postcss.config.cjs` it was written against moves to `website/` here, so there is no longer any PostCSS config between `app/` and the repository root. The search does not stop at the project directory, though. Probing the `postcss-load-config@6.0.1` that Vite resolves, from a project directory with a config one level above it:

```
UPWARD SEARCH ESCAPED -> found: .../repo/postcss.config.cjs
```

So without the empty config, Vite would keep walking past the repository into ancestor directories. `templates/react/postcss.config.js` is the same empty guard for the same reason, so removing only this one would also be inconsistent.

Happy to delete it if you would rather not carry a guard against out-of-repo configs — it just is not a no-op.

app/src/lib/source-plugin.test.ts
diegohaz: Tracked in [#6915](https://github.com/ariakit/ariakit/issues/6915): the filters are wider than the commands in at least two jobs, but narrowing them needs verification from the workflows that own them rather than from this PR's checks.

You are right about `Post PR comment`. It installs `app...` — 1366 packages, ~1.8 GB — and the only thing it runs from that graph is `ariakit perf-compare`. `app` is being used as a working directory rather than as a dependency: `perf-compare` resolves its input as `path.join(process.cwd(), ".perf-results")`, which is why the job stages rounds into `app/.perf-results` and shells out through `pnpm -F app run perf-compare`. So dropping to `./packages/ariakit-scripts...` also means giving `perf-compare` an explicit results directory.

`Test React 18` looks similar. It installs `root...`, then `ariakit react18` rsyncs the repo to a temp workspace and runs its own **unfiltered** `pnpm install --no-frozen-lockfile` there, so the outer install only provides the `ariakit` bin and the files being rsynced.

I checked the rest and they look justified: `Lint`, `Type check`, `Test`, `Test React`, and `Build website` in `main.yml`, plus `app.yml`, `docs.yml`, `plus.yml`, `og-images.yml`, `deploy*.yml`, and the `release*.yml` and perf Node jobs that genuinely build every package.

Keeping it out of this PR because both changes need a real perf run and a real React 18 run to confirm the jobs still produce the same output, which is independent of the dependency move here.

diegohaz:
Handled the two unresolved review threads. No code changes were needed.

**Tracked**

- [#6915](https://github.com/ariakit/ariakit/issues/6915) ([reply](https://github.com/ariakit/ariakit/pull/6911#discussion_r3665879228)): `Post PR comment` installs `app...` (1366 packages, ~1.8 GB) only to run `ariakit perf-compare`, and `Test React 18` installs `root...` even though `ariakit react18` reinstalls unfiltered in a temp workspace. Both need a real perf run and a real React 18 run to confirm, so they are separate from this PR. The remaining filters checked out as justified.

**Invalid**

- `app/postcss.config.js` ([reply](https://github.com/ariakit/ariakit/pull/6911#discussion_r3665877341)): both files are still load-bearing. `nextjs/postcss.config.js` is the Next.js app's actual Tailwind v4 wiring, not a guard. `app/postcss.config.js` is still a guard — probing the `postcss-load-config@6.0.1` Vite resolves confirms the search walks past the project directory into ancestors — though this PR does remove the in-repo config it was originally written against, which is why its comment was reworded.

app/postcss.config.js
diegohaz: Removed both in [`8b9f3b350`](https://github.com/ariakit/ariakit/commit/8b9f3b35038f980f32935f6943d3a863b810b6c3), following your call to drop the guard rather than keep it: `app/postcss.config.js` and `templates/react/postcss.config.js` are gone, `nextjs/postcss.config.js` stays because it is real Tailwind v4 wiring.

To close the loop on my earlier reply, the removal is behavior-preserving inside a clean checkout. Building the app with and without them emits byte-identical CSS — the same six files under `app/dist` with the same content hashes — and a minimal project mirroring `templates/react`'s `vite` + `@tailwindcss/vite` stack does the same. Both were built outside the repository tree so no ancestor config could influence the result.

What is genuinely given up is the out-of-repo case: the search still escapes the project directory, so a `postcss.config.*` sitting above the checkout can now reach `app/` and `templates/react/`. That is what bit this PR locally, since these worktrees live under a checkout that still has the old root config — but it is not a state the repository can control, and it does not exist in CI. Noted in the PR description alongside the reference to #6503, which had documented the app shield.

PR #6913, opened 2026-07-28
Stabilize flaky browser tests

diegohaz:
## Motivation

Recent Vitest and Playwright runs intermittently failed while waiting for late Combobox item registration, offscreen selection, and new-tab state. Raw logs from otherwise successful PR jobs also exposed hidden retries in Popover placement after viewport changes and route-driven Dialog focus restoration.

The offscreen matrix rendered all 88 cases at once, increasing scheduling pressure and allowing hover, filtering, and render timing to race keyboard assertions.

The investigation also uncovered #6916, an unreleased Combobox caret regression introduced after @ariakit/react@0.4.35. That runtime fix is intentionally excluded from this PR.

## Solution

- Render only the requested offscreen matrix case and remount it when the case changes.
- Synchronize Playwright actions with observable open, focus, filtering, active-item, pointer-neutral, hydrated-page, and geometry states.
- Dispatch modifier+Enter with an explicit modifier-key lifetime and allow slow new tabs to finish hydrating.
- Restore the route-driven Dialog trigger only after the closing route commits.
- Give the React 18 late-item scroll assertion a three-second scheduling budget while still requiring the selected element to be scrolled.

## Testing

- pnpm build-app-lite
- pnpm lint
- pnpm tsc
- Affected React integration tests: 7/7
- Deferred Combobox navigation stress: 100/100 in Chrome with no retries
- Broader Combobox group: 140/140 in Chrome with no retries, excluding the separately tracked #6916 regression
- Firefox new-tab tests: 20/20 sequential; modifier+Enter 30/30 parallel with no retries
- Affected Playwright matrix: 198/198 across Chrome, Firefox, and Safari with no retries, excluding #6916
- Final GitHub CI: all Vitest and Playwright jobs passed; their raw logs contain no retry, flaky, failed, or failure markers; Gate passed


app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
```diff
@@ -116,7 +127,7 @@ withFramework(import.meta.dirname, async ({ id, query, test }) => {
     await newPage.waitForURL(hasSearchParam("status", ["draft", "archived"]));
     await test
       .expect(query(newPage).combobox("Status"))
-      .toHaveText("2 selected");
+      .toHaveText("2 selected", { timeout: 15_000 });
```
diegohaz: We need a comment explaining why this big timeout number is needed (if it's really needed).

app/src/sandbox/combobox-links/test-browser.ts
```diff
@@ -12,7 +12,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
 
   test("click on link with mouse", async ({ page }) => {
     const q = query(page);
-    await q.combobox("Links").click();
+    const combobox = q.combobox("Links");
+    await combobox.click();
+    await expect(combobox).toHaveAttribute("aria-expanded", "true");
     await expect(q.listbox()).toBeVisible();
```
diegohaz: `aria-expanded` is rendered straight from the store's `open` state — `packages/ariakit-react-components/src/combobox/combobox.tsx:769` has `"aria-expanded": open` — and the popover cannot become visible before `open` is true. So the pre-existing `await expect(q.listbox()).toBeVisible()` on the next line already subsumes this assertion: there is no moment where the listbox is visible while `aria-expanded` is still `"false"`, so this line can never be the wait that catches a race.

All six copies in this file sit in that same position (lines 17, 31, 52, 72, 98, 114), each immediately followed by the same `toBeVisible()` call. `app/src/sandbox/combobox-select-offscreen-value/test-browser.ts` lines 9, 19, 46, 56, 73, 84 and `app/src/sandbox/combobox-multiple/test-browser.ts:39` have the same shape — the next statement is an auto-retrying assertion or action on an element that only resolves once the popover is open, so it already waits for the same condition.

The additions that do pay for themselves are the ones gating a step with no auto-wait, or an assertion that could otherwise pass vacuously, and those are worth keeping:

- `combobox-multiple/test-browser.ts:20`, `combobox-select-offscreen-value/test-browser.ts:34` and `:67` — next statement is `page.keyboard.*`, which has no actionability wait.
- `combobox-select-nextjs-app-router/test-browser.ts:108` — gates `page.mouse.move(0, 0)` plus `keyboard.press("d")`.
- `combobox-select-scroll-into-view/test-browser.ts:39` — this one genuinely strengthens the test, because `await test.expect(selected).toHaveCount(0)` on the next line would otherwise pass vacuously while the popover is still closed.

Why this is worth the churn: a stabilization PR is exactly where future readers assume every wait is load-bearing. Mixing roughly a dozen no-op waits in with the five that carry real synchronization makes the next person guess which ones are safe to touch, which is how the sync points that matter get deleted later.

**Assessment**

- Severity: Low (verified, 95% confidence), because the redundant assertions cost readability and a few extra locator round-trips only; they cannot change a pass/fail outcome.
- Priority: P2 (estimate, 85% confidence), because it is a clarity cleanup in code this PR introduces rather than a correctness gate.
- Breakage likelihood: Rare (verified, 90% confidence), since nothing breaks today; the risk is a later edit removing one of the load-bearing waits by analogy with these.
- Fix effort: Easy (verified, 95% confidence), because it is deleting lines with no other change required.

```suggestion
    await expect(q.listbox()).toBeVisible();
```

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
diegohaz: Confirmed, and it applies to two places: line 74 carries the same undocumented `{ timeout: 15_000 }` on `toHaveText("French", ...)`.

Two things worth folding into whatever comment lands here, because they change what the right lever is:

**1. The 15s budget is not actually reachable.** `app/playwright.config.ts` sets no top-level `timeout` and neither does the `chrome`/`firefox`/`safari` project, so these tests run under Playwright's 30s default — and that budget also covers the `beforeEach` `gotoAndSettle` (a `load` navigation plus up to a 5s bounded network-idle settle) and every step above this line. When this assertion is the thing that fails, the test hits the 30s test timeout before the 15s expect timeout elapses, so the failure surfaces as `Test timeout of 30000ms exceeded` instead of the matcher's own text diff. If the extra budget is genuinely needed, it needs a test-level lever alongside it; `test.slow()` triples the test timeout for the whole test and is self-documenting.

**2. What is being awaited here is server-rendered, not hydrated.** `nextjs/app/combobox-select-nextjs-app-router/page.tsx` computes `displayValue` on the server via `getLabel(statuses, status, "Any")`, and `router-select.tsx` renders it as the `ComboboxSelect` child, so `"2 selected"` (and `"French"` on line 74) is present in the new tab's initial HTML. `newPage.waitForURL(...)` on line 127 already waits for `load` by default (`waitUntil` defaults to `load`), which for streaming SSR means the whole response, `<Suspense>` boundary included, has flushed. So the extra budget is absorbing cold-start latency of the `opennextjs-cloudflare preview` worker under parallel load, not hydration or a client transition.

That distinction is the useful part of the comment: it tells the next reader the number is about server cold start rather than component behavior, so shrinking it is safe once the server is warm, and that the correct place to spend it is the test timeout rather than a single matcher.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the consequence is a slower and less legible failure, not a wrong result.
- Priority: P1 (estimate, 80% confidence), because an unexplained magic number in a stabilization PR is the kind of thing that gets copied forward, and the ceiling is currently unreachable, so the mitigation may not work as intended.
- Breakage likelihood: Possible (estimate, 70% confidence), since it only bites when the new tab is genuinely slow, which is the exact CI contention scenario this PR targets.
- Fix effort: Easy (verified, 90% confidence), because it is a comment plus either `test.slow()` or an explicit per-test timeout.

diegohaz (commented):
## Review summary

Initial review — no prior `ariakit-pr-review` marker exists on this PR, so the baseline is the merge base `c8c0ac9426b942024216ad91a4f98d813ca67a15` through head `75622060adafef95899b1e91979f8dd090a72de3`.

### Scope checked

All 8 changed files, plus the surrounding context they depend on: the four touched sandbox fixtures, `packages/ariakit-react-components/src/combobox/combobox.tsx` (for how `aria-expanded` is produced), `app/playwright.config.ts` (timeouts, projects, web servers), `app/src/test-utils/preview.ts` and `fixtures.ts`, `nextjs/app/combobox-select-nextjs-app-router/{page,router-select,filters}.tsx`, `app/src/sandbox/combobox-select-offscreen-matrix/config.react.ts`, and the CI results for this head.

### Overall

The direction is sound. Rendering one matrix case at a time is the right lever for the offscreen suite: the 88-case page was the actual source of scheduling pressure, and the per-case tests never depended on the other cases being mounted, so no coverage is lost by mounting one. Selecting the case through a real `<select>` (rather than a query param and a second navigation) keeps the fixture browsable and costs one in-page interaction instead of a page load per test. Both label sets stay unambiguous — no case label produced by `getCaseLabel` collides with a country name, with `"Matrix case"`, or with `"Search..."` — so the extra `combobox`/`option` roles the native `<select>` adds do not disturb the existing name-scoped queries.

### Published inline

- `app/src/sandbox/combobox-links/test-browser.ts:17-18` — the six added `aria-expanded` assertions in that file are subsumed by the `toBeVisible()` call on the next line, and the same next-line-auto-wait shape covers most of the additions in `combobox-select-offscreen-value/test-browser.ts` and `combobox-multiple/test-browser.ts:39`. The comment also lists the five additions that do carry real synchronization and should stay.

### Relevant unresolved thread validated

- [#discussion_r3664992188](https://github.com/ariakit/ariakit/pull/6913#discussion_r3664992188) (`combobox-select-nextjs-app-router/test-browser.ts:130`, unresolved) — **Confirmed.** Replied on the thread. The magic 15s appears twice (lines 74 and 130), the value it awaits is server-rendered rather than hydrated, and `newPage.waitForURL(...)` already waits for `load`. Independently of the comment, the ceiling is not reachable: the file's tests run under Playwright's 30s default test timeout (`app/playwright.config.ts` sets none), which is shared with `beforeEach`, so a real failure surfaces as a test timeout rather than the matcher's own diff.

### Findings not published inline

- **Console-warning coverage in `offscreen placeholders omit item-only DOM props` narrows.** That test attaches a page-wide console listener and reloads; before this PR the reload mounted all 88 matrix cases, so a React DOM-prop warning from the matrix cases' own `ComboboxItem` usage would have been caught. It now mounts only the two probe fixtures. *Disposition: Declined non-goal.* `offscreenProbeProps` already passes the full item-only prop surface the matrix cases use (`focusOnHover`, `blurOnHoverEnd`, `offscreenMode`, `offscreenRoot`), and the test's stated target is the offscreen placeholders. The residual gap is `ComboboxGroup`/`ComboboxGroupLabel` and `Ariakit.Combobox autoSelect`, which the test never asserted on specifically.
- **The new `<select>` adds `role="combobox"` plus 88 `role="option"` nodes to a combobox fixture.** Harmless today because every query in `test-chrome.ts` is name-scoped, but it is now part of the fixture contract: an unnamed `q.combobox()` or a `getByRole("option")` count in this sandbox would be wrong in a non-obvious way. *Disposition: Declined non-goal*, worth one sentence of awareness rather than a change.
- **Nits, not worth a round trip:** `renderMatrixCase` uses `page.getByRole(...)` where `q.combobox("Matrix case")` would match the app-test convention, and its name says "render" while it selects; `Example()` shadows `matrixCase` inside its own initializer; the `keyboardTarget` derivation is repeated four times in three spellings across the loops in `test-chrome.ts`. *Disposition: Declined non-goal.*
- **`combobox-select-scroll-into-view/test.ts` poll timeout.** The bump to `{ timeout: 3000 }` matches the poll directly above it and does not weaken the regression: broken behavior never calls `scrollIntoView` with the selected element, so the assertion still fails for the right reason, three seconds later. No finding.
- **No changeset needed.** Tests and a sandbox only, no shipped-package change; the changeset-bot warning is expected here.

### CI

`App / Test Safari` is red on this head ([job](https://github.com/ariakit/ariakit/actions/runs/30353347859/job/90256214445)): `10 failed`, `1 flaky`, `432 passed (10.7m)`. The evidence says infrastructure, not this PR:

- At `11:12:21` the web server output logs `ERROR Wrangler dev command failed:` for the `opennextjs-cloudflare preview` server, and every failure after that point is `Could not connect to the server.` — including `page.goto` against `http://localhost:4321/...`, so the Astro preview went down with it.
- Six of the ten failures are in `dialog-nextjs-app-router` and `tab-nextjs-app-router`, which this PR does not touch.
- The one failure that is not a connection error (`toHaveText("Draft")` at line 113, `element(s) not found`) happened at `11:12:35`, after the crash, which is what a dead RSC server produces.
- The single flaky test, `combobox-links` "click on link with cmd/ctrl + enter", failed on the WebKit `toPass()` branch at line 82 and passed on retry; that branch is untouched by this PR.

A re-run is the appropriate response rather than a code change. Everything else on this head is green, including `App / Test Chrome` (6m27s) and `App / Test Firefox`.

### Not fully validated

I did not run the Playwright or Vitest suites locally — this review worktree has no installed dependencies, and the flakiness being targeted is CI-contention-dependent, so a local run would not reproduce it either way. The stress-run numbers in the PR description are therefore taken as reported and not independently reproduced. The redundancy finding and the timeout analysis are both derived from source and from `app/playwright.config.ts` rather than from execution.

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
diegohaz: Fixed in [`f8688f313`](https://github.com/ariakit/ariakit/commit/f8688f313242a17e51fb8ad96d79b22d14fc89bd): moved both cold-preview-server budgets to test-level `test.slow()`, documented the reason, and restored the default matcher timeout so UI failures keep focused diagnostics.

app/src/sandbox/combobox-links/test-browser.ts
diegohaz: Fixed in [`15fb719e3`](https://github.com/ariakit/ariakit/commit/15fb719e3f3e55a23e41fd95382069674e12e67e): removed the waits already subsumed by listbox visibility, actionable hover/click, focus, or viewport checks. I retained the offscreen waits whose closed-state attributes can pass vacuously or that gate raw keyboard input. A retry-disabled Safari red check also reproduced the real-focus viewport race 1/20, so that path keeps its open wait and now crosses the component’s after-paint callback plus WebKit scroll step before asserting the selected item is in view.

diegohaz:
**Fixed**

- [`f8688f313`](https://github.com/ariakit/ariakit/commit/f8688f313242a17e51fb8ad96d79b22d14fc89bd) ([reply](https://github.com/ariakit/ariakit/pull/6913#discussion_r3665418935)): moved cold preview-worker latency to the test-level budget and restored focused matcher diagnostics.
- [`15fb719e3`](https://github.com/ariakit/ariakit/commit/15fb719e3f3e55a23e41fd95382069674e12e67e) ([reply](https://github.com/ariakit/ariakit/pull/6913#discussion_r3665422338)): removed redundant open-state waits, retained the non-vacuous synchronization points, and added a WebKit render checkpoint for the reproduced offscreen viewport race.

packages/ariakit-react-components/src/combobox/combobox.tsx
```diff
@@ -312,10 +345,14 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
...
-          setSelectionRange(element, prevStart, prevEnd);
+          setSelectionRange(element, prevSelection.start, prevSelection.end);
+          inlineSelectionRef.current = null;
         };
       });
-      return () => cleanup();
+      return () => {
+        canceled = true;
+        restoreSelection();
+      };
```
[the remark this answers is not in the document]
diegohaz: No longer applicable after [`c03b85981`](https://github.com/ariakit/ariakit/commit/c03b85981): per the final scope decision, I removed the inline-selection runtime fix and its changeset from this PR. The component now has no diff from `main`. The caret regression, npm-version A/B result, and minimal StackBlitz repro are tracked separately in #6916 for a follow-up PR.

diegohaz:
Follow-up complete in [`c03b85981`](https://github.com/ariakit/ariakit/commit/c03b85981):

- Removed the inline-selection runtime fix and changeset from the cumulative PR; `combobox.tsx` now matches `main`.
- Tracked the unreleased caret regression separately in #6916 with an npm 0.4.35 comparison and a minimal StackBlitz repro.
- Replaced the Firefox modifier+Enter action retry with one page waiter and one shortcut dispatch, guaranteeing modifier release in `finally`.
- Added observable filtering/active-item and new-tab hydration synchronization.

Verification for the final patch: lint and type-check pass; deferred navigation 100/100 in Chrome; modifier+Enter 30/30 in Firefox parallel; affected Playwright matrix 198/198 across Chrome, Firefox, and Safari with no retries. Final GitHub CI is now running.

diegohaz:
Final CI verification for [`c03b85981`](https://github.com/ariakit/ariakit/commit/c03b85981): [run 30370893725](https://github.com/ariakit/ariakit/actions/runs/30370893725) completed successfully and Gate passed.

All Vitest jobs (base, React, React 18, and Solid) and all Playwright jobs (Chrome, Firefox, Safari, Android, and iOS) passed. I also fetched each job’s raw log and found no `retry`, `retries`, `flaky`, `failed`, or `failure` markers. Vercel status is intentionally excluded from this assessment.

app/src/sandbox/combobox-select-offscreen-value/test-browser.ts
```diff
@@ -56,15 +62,26 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
     await test.expect(q.combobox("Search Filterable fruit")).toBeFocused();
-    await test.expect(q.option("Watermelon")).toBeInViewport();
+    await page.evaluate(
+      () =>
+        new Promise((resolve) => {
+          requestAnimationFrame(() =>
+            requestAnimationFrame(() => requestAnimationFrame(resolve)),
+          );
+        }),
+    );
```
diegohaz: `#app/test-utils/preview.ts` — the module this file already imports `withFramework` from — exports `flushFrames(page, frames = 2)`, which is exactly this promise. It is the established helper for this: `combobox-select-3837/test-browser.ts` uses `flushFrames(page, 2)` and `flushFrames(page, 5)`, and `button-ref-cleanup`, `composite-renderer-6215`, `popover-disclosure-render`, and `collection-renderer-perf` all use it too.

Worth switching for two reasons beyond line count. First, the raw form hides the frame count inside nested callbacks, so the number that actually matters (three) is the least visible part of the block. Second, `flushFrames` is where a future fix would land if a browser ever needs a different flush strategy, and hand-rolled copies silently opt out of it.

The count also deserves the same treatment as the timeouts elsewhere in this PR: three frames is a specific claim about crossing the component's after-paint callback plus WebKit's scroll step, and nothing in the file says so. A reader trying to shorten this later has no way to know which frame is load-bearing.

While you are here, the twin block at lines 93-98 in `cancels presentation when real focus moves` is the same pattern at the default two frames and predates this PR — `await flushFrames(page)` would fold it in without a behavior change.

The import becomes `import { flushFrames, withFramework } from "#app/test-utils/preview.ts";`.

**Assessment**

- Severity: Low (verified, 95% confidence), because the inline form behaves identically to the helper; the cost is duplication and a hidden constant.
- Priority: P2 (estimate, 85% confidence), because it is reuse and documentation of code this PR adds, not a correctness gate.
- Breakage likelihood: Rare (verified, 90% confidence), since the only failure mode is a future reader shortening an undocumented frame count and reintroducing the WebKit race this line exists to close.
- Fix effort: Easy (verified, 95% confidence), because it is one import and a one-line replacement.

```suggestion
    // Three frames cross the component's after-paint presentation callback
    // plus WebKit's scroll step before the viewport check.
    await flushFrames(page, 3);
```

app/src/sandbox/combobox-links/test-browser.ts
```diff
@@ -73,10 +78,16 @@ withFramework(import.meta.dirname, async ({ test }) => {
         await expect(page).toHaveURL(/https:\/\/ariakit\.com/);
       }).toPass();
     } else {
-      const [newPage] = await Promise.all([
-        context.waitForEvent("page"),
-        page.keyboard.press(`${modifier}+Enter`),
-      ]);
+      const newPagePromise = context.waitForEvent("page", {
+        timeout: 60_000,
+      });
```
diegohaz, line 87: This 60s wait and the `test.slow()` on line 63 are both unexplained, and they are the two numbers a future reader will most want justified — the same gap you flagged on the `15_000` timeouts earlier in this PR.

The mechanics are right, which is worth saying explicitly because it is not obvious: `test.slow()` raises this test's budget to 90s, and the app config sets no `actionTimeout`, so an unbounded `waitForEvent` would otherwise surface as a 90s test timeout with no indication of what was being waited on. Bounding it at 60s inside a 90s budget means the failure names the missing page event instead. That is the inverse of the `15_000` problem and is exactly the right shape — but nothing in the file records the reasoning, so the next person to touch this sees two arbitrary numbers and no reason not to change either one.

The sibling file already sets the standard: `combobox-select-nextjs-app-router/test-browser.ts` carries `// Opening a new tab may cold-start the preview worker under CI load.` above both of its `test.slow()` calls. Two comparable lines here would close the loop:

```ts
    // Chrome and Firefox can take a while to materialize the modifier+Enter
    // tab under CI load.
    test.slow();
```

```ts
      // Fail on a missing page event well inside the slow-test budget, so the
      // error names the event instead of timing out the whole test.
      const newPagePromise = context.waitForEvent("page", {
        timeout: 60_000,
      });
```

**Assessment**

- Severity: Low (estimate, 90% confidence), because the values behave correctly today; the cost is that the reasoning is unrecoverable from the code.
- Priority: P2 (estimate, 80% confidence), because it is documentation of a deliberate choice rather than a behavior fix, but leaving it undocumented in the PR that established the opposite standard is inconsistent.
- Breakage likelihood: Possible (estimate, 70% confidence), since undocumented timing constants are routinely trimmed during later cleanups, which would restore the flake this test was changed to fix.
- Fix effort: Easy (verified, 95% confidence), because it is two comments.

app/src/sandbox/popover-flip-placement/test-chrome.ts
```diff
@@ -7,36 +7,33 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
-      .expect(popoverBox.y)
-      .toBeGreaterThanOrEqual(anchorBox.y + anchorBox.height);
+    await test.expect
+      .poll(async () => {
+        const anchorBox = await disclosure.boundingBox();
+        const popoverBox = await popover.boundingBox();
+        if (!anchorBox || !popoverBox) return -Infinity;
+        return popoverBox.y - anchorBox.y - anchorBox.height;
+      })
+      .toBeGreaterThanOrEqual(0);
```
diegohaz: Polling the geometry instead of sampling it once is the right fix — I checked all three deltas and they preserve the original inequalities exactly (`popoverBox.x >= anchorBox.x + anchorBox.width - 1`, `anchorBox.y >= popoverBox.y + popoverBox.height`, `popoverBox.y >= anchorBox.y + anchorBox.height`). The shape it landed in is what I would push back on.

The three blocks are identical except for one arithmetic expression, and folding the comparison into a single `>= 0` delta costs the reader the thing they actually came for. `popoverBox.x - anchorBox.x - anchorBox.width + 1` no longer reads as "the popover starts at or after the anchor's right edge, with one pixel of tolerance" — the `+ 1` in particular now looks like part of the arithmetic rather than the tolerance it is. The `-Infinity` sentinel is a fourth copy of the same idea and is close to unreachable anyway, since `toBeVisible()` on line 8 already passed and a visible element has a box.

Extracting the poll keeps the retry semantics, removes the duplication, and lets each assertion state its geometric relation in the original form:

```ts
import type { Locator } from "@playwright/test";
import { withFramework } from "#app/test-utils/preview.ts";

type Box = NonNullable<Awaited<ReturnType<Locator["boundingBox"]>>>;

  test("falls back from right to top and bottom", async ({ page, q }) => {
    const disclosure = q.button("Accept invite");
    const popover = q.dialog("Team meeting");

    // Placement settles asynchronously after viewport and scroll changes, so
    // poll the geometry rather than sampling it once.
    const expectGap = (getGap: (anchor: Box, popover: Box) => number) =>
      test.expect
        .poll(async () => {
          const anchorBox = await disclosure.boundingBox();
          const popoverBox = await popover.boundingBox();
          if (!anchorBox || !popoverBox) return -Infinity;
          return getGap(anchorBox, popoverBox);
        })
        .toBeGreaterThanOrEqual(0);

    await disclosure.click();
    await test.expect(popover).toBeVisible();
    await expectGap((anchor, pop) => pop.x - (anchor.x + anchor.width - 1));

    await page.setViewportSize({ width: 480, height: 1024 });
    await expectGap((anchor, pop) => anchor.y - (pop.y + pop.height));

    await page.evaluate(() => window.scrollTo(0, 400));
    await expectGap((anchor, pop) => pop.y - (anchor.y + anchor.height));
  });
});
```

That is 28 lines against the current 34, and each call site names its relation on one line. Adjacent app tests already import `Locator` as a type from `@playwright/test` (`combobox-multiple/test-browser.ts`, `combobox-select-offscreen-matrix/test-chrome.ts`), so the import is in keeping with the directory.

**Assessment**

- Severity: Low (verified, 95% confidence), because the current form is behaviorally correct; the cost is readability of a geometric contract.
- Priority: P2 (estimate, 85% confidence), because it is a simplification of code this PR rewrites rather than a defect.
- Breakage likelihood: Possible (estimate, 65% confidence), since the one-pixel tolerance is now easy to misread as arithmetic, and a later edit that "cleans up" the `+ 1` would silently tighten the assertion.
- Fix effort: Easy (verified, 90% confidence), because it is a local extraction within one test.

diegohaz (commented):
## Re-review summary

Re-review. Baseline is the head recorded in the previous review marker, `75622060adafef95899b1e91979f8dd090a72de3`, through the current head `c03b85981d9f0159765ab806fed6643810f3a753` — six new commits. Newly discovered findings come from that incremental range plus surrounding issues it exposes; I also re-read the full merge-base diff because the scope widened past Combobox.

### Previous findings

Both were addressed and their threads are resolved.

- Redundant `aria-expanded` waits — fixed in [`15fb719e3`](https://github.com/ariakit/ariakit/commit/15fb719e3f3e55a23e41fd95382069674e12e67e). I verified the retained ones are exactly the load-bearing set: the offscreen waits that gate assertions which could pass vacuously against closed-state attributes, and the ones preceding raw keyboard input.
- Undocumented `15_000` matcher timeouts — fixed in [`f8688f313`](https://github.com/ariakit/ariakit/commit/f8688f313242a17e51fb8ad96d79b22d14fc89bd). The `filter({ hasText }).waitFor()` then `toHaveText()` pair is a good resolution and worth naming, since it is not self-evident: `waitFor()` absorbs the cold-start latency against the slow-test budget while `toHaveText()` keeps the precise equality check and its diff. One line of comment would stop a future reader from collapsing the two.

### Scope change

The PR now also touches `popover-flip-placement`, `combobox-group-basic`, and the `dialog-nextjs-app-router` Next.js fixture. The title and description were updated to match, so the scope is stated rather than implied.

### Published inline

- `combobox-select-offscreen-value/test-browser.ts:76-83` — the new triple-`requestAnimationFrame` block re-implements `flushFrames` from the module this file already imports from; the frame count is also undocumented.
- `combobox-links/test-browser.ts:81-83` — `timeout: 60_000` and the `test.slow()` on line 63 are unexplained, in the PR that established the opposite standard. The mechanics are correct (60s bounded inside a 90s budget names the missing event instead of timing out the test); only the reasoning is missing.
- `popover-flip-placement/test-chrome.ts:10-37` — three near-identical polls; extracting one keeps the retry semantics and restores the readable geometric comparisons, including the one-pixel tolerance that the folded `>= 0` delta now obscures.

### Verified, no finding

- **All three `popover-flip-placement` deltas are faithful** to the inequalities they replaced. Checked individually.
- **`aria-busy` in `combobox-multiple` is fixture-owned and user-visible.** `index.react.tsx:34` sets `aria-busy={isPending}` on the popover and the class list styles `aria-busy:[--item-opacity:0.5]`, so `toHaveAttribute("aria-busy", "false")` is an observable-state sync rather than an implementation-detail probe. Good choice.
- **The `dialog-nextjs-app-router` focus rework is correct on every path its unchanged test exercises.** I traced all four: link click then Escape, `goBack()` then backdrop click, direct `goto(/login)` then `goBack()`/`goForward()`/`reload()`, and Enter-to-submit. The `previousPathname` ref initialised to the mount pathname correctly suppresses focus on first mount in both entry cases. Replacing the old `autoFocusOnHide` callback — which raced a `requestAnimationFrame` against `document.querySelector` — with route-driven restoration in the destination page is a real improvement, not just a flake patch.
- **`ArrowUp` selecting `Ariakit.com`** in `combobox-links` matches the fixture: it is the last entry in `links`, and the click-open path does not auto-activate, so the wrap-around assertion holds.
- **`combobox-group-basic` filter assertions** hold: `Avocado` contains no `ap`, and `Grape`/`Papaya` are the next two matches after `Apple`.

### Not published

- **`nextjs/app/dialog-nextjs-app-router/page.tsx` becomes a client component.** *Disposition: Declined non-goal.* `layout.tsx`, `@login/page.tsx`, and `default.tsx` remain server components, so the sandbox still covers parallel routes and the App Router server boundary. The callback form of `autoFocusOnHide` that this fixture dropped is still exercised by `examples/dialog-radix/dialog.tsx:218` and `website/app/(examples)/previews/dialog-next-router/@login/login/page.tsx:19`, so no public-API coverage is lost. Keeping `page.tsx` a server component and extracting a small client `LoginLink` would preserve one more server boundary, but that is a preference, not a requirement.
- **`expect(q.combobox()).toBeFocused()` immediately before `q.combobox().press(...)` in `combobox-multiple`.** *Disposition: Declined non-goal*, flagging it so it is not mistaken for the redundancy class removed in `15fb719e3`. `Locator.press` focuses first, so the assertion is the only thing that still proves focus stayed on the combobox after filtering. It is load-bearing precisely because `press` would otherwise mask a focus regression.

### CI

Every GitHub check passes on this head, including `App / Test Safari` (14m30s), which was the crashed-web-server failure last time, and the `Gate` check.

`Vercel` is the exception and it needs a look before merge. It is `failure` on this head, `success` on the base `c8c0ac942`, on the last five `main` commits, and on the three other open PR heads. On this branch it was `success` through `e7fdb77c8` and has been `failure` from `15fb719e3` onward. I could not find a code path that would explain it: Vercel runs `pnpm run build-website` with `outputDirectory: website/.next`, the `website` workspace declares no dependency on `app` and contains no reference to `app/src`, and the commits where it flipped touch only `app/src/sandbox/**` Playwright tests. Worth noting that the green `Main / Build website` job runs `build-website-lite` (`DISABLE_SHIKI=true`), so it does not fully stand in for Vercel's build. The deployment logs are the only way to settle it, and I cannot reach them: `npx vercel inspect dpl_25XzNzGeF4nDavFvozfNQo3ye23D --logs`.

### Not fully validated

Same gap as before: no local suite runs, since this review worktree has no installed dependencies and the flakiness is CI-contention-dependent. The stress-run numbers in the description are taken as reported. The Vercel failure is unexplained rather than dismissed.

app/src/sandbox/combobox-select-offscreen-value/test-browser.ts
diegohaz: Fixed in [`c321877a6`](https://github.com/ariakit/ariakit/commit/c321877a6): replaced both hand-written frame chains with `flushFrames`, preserving the exact three- and two-frame counts and documenting why the third frame is load-bearing for WebKit.

app/src/sandbox/combobox-links/test-browser.ts
diegohaz: Fixed in [`58c31a2ec`](https://github.com/ariakit/ariakit/commit/58c31a2ec): documented the Chrome/Firefox CI delay behind `test.slow()` and why the 60-second page-event deadline sits inside the 90-second test budget.

app/src/sandbox/popover-flip-placement/test-chrome.ts
diegohaz: Fixed in [`061684c11`](https://github.com/ariakit/ariakit/commit/061684c11): extracted the repeated bounding-box poll into `expectGap`, retained the null-box retry sentinel, and expressed all three original edge relations—including the one-pixel right-side tolerance—at their call sites.

diegohaz:
**Fixed**

- [`c321877a6`](https://github.com/ariakit/ariakit/commit/c321877a6) ([reply](https://github.com/ariakit/ariakit/pull/6913#discussion_r3667353091)): reused `flushFrames`, preserved the exact frame counts, and documented the WebKit-specific third frame.
- [`58c31a2ec`](https://github.com/ariakit/ariakit/commit/58c31a2ec) ([reply](https://github.com/ariakit/ariakit/pull/6913#discussion_r3667354634)): explained the 90-second slow-test and 60-second page-event timeout relationship.
- [`061684c11`](https://github.com/ariakit/ariakit/commit/061684c11) ([reply](https://github.com/ariakit/ariakit/pull/6913#discussion_r3667356708)): extracted the repeated Popover geometry poll while preserving all three edge relations and the one-pixel tolerance.

**Integrated**

- [`10aa0848a`](https://github.com/ariakit/ariakit/commit/10aa0848a): merged `origin/main` through `b67d7f2bc`; the cumulative PR diff remains limited to flaky-test and fixture synchronization.

**Verification**

- `pnpm build-app-lite`
- Safari offscreen-value suite: 6/6
- Chrome Popover placement stress: 30/30 with retries disabled
- Merged source tests: 32/32
- `pnpm lint`
- `pnpm tsc`

nextjs/app/dialog-nextjs-app-router/page.tsx
```diff
@@ -1,5 +1,27 @@
...
 export default function Page() {
-  return <Link href="/dialog-nextjs-app-router/login">Login</Link>;
+  const pathname = usePathname();
+  const previousPathname = useRef(pathname);
+  const linkRef = useRef<HTMLAnchorElement>(null);
+
+  useEffect(() => {
+    const previous = previousPathname.current;
+    previousPathname.current = pathname;
+    if (previous !== href) return;
```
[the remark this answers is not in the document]
diegohaz: I traced this mechanism and it does not reproduce. Recording the evidence so it does not get re-opened.

The flow described here — direct load of `/dialog-nextjs-app-router/login`, then closing the dialog and expecting the Login link to regain focus — is exactly what `app/src/sandbox/dialog-nextjs-app-router/test-browser.ts:26` (`supports direct navigation and browser history`) asserts: `page.goto(/login)`, `goBack()`, `goForward()`, `reload()`, `Enter` to submit, then `await test.expect(q.link("Login")).toBeFocused()`. The reload leaves the children slot in precisely the `default.tsx` state this report is about.

That test passed on Chrome, Firefox, and Safari at [`c03b85981`](https://github.com/ariakit/ariakit/commit/c03b85981d9f0159765ab806fed6643810f3a753), which contains the identical `page.tsx` and `default.tsx` (the change landed in [`c5be82cb9`](https://github.com/ariakit/ariakit/commit/c5be82cb9fb8fab0d0b928c22106a71b5b365013)), and it passes on Chrome and Firefox at the current head. The failure this predicts would be deterministic — a remounted `Page` initialises `previousPathname` to the base route and can never focus — so it would fail on every run in every project, not intermittently. It does not.

In practice the children-slot subtree survives the `default.tsx` to `page.tsx` swap on the soft navigation back to the base route, so the `Page` instance keeps its `previousPathname` ref and the effect sees `previous === href`.

The observation underneath is fair, though: the restoration does depend on that instance surviving a slot swap, which is Next.js-internal behavior rather than a documented contract. What makes that acceptable here is that the sandbox's own test pins the behavior across all three browser projects, so a future Next.js change that starts remounting the slot fails loudly in CI rather than silently regressing focus restoration.

diegohaz (commented):
## Re-review summary

Re-review. Baseline is the head recorded in the previous review marker, `c03b85981d9f0159765ab806fed6643810f3a753`, through the current head `10aa0848a1e7955527542dfcfb2fe1e0b3b94bb6` — three fix commits plus two merges from `main`. Since the base moved to `b67d7f2bc`, I also re-read the full merge-base diff so the merged content is not mistaken for PR work.

**No new findings.** All three findings from the previous round are fixed correctly, the one new unresolved thread does not reproduce, and the Vercel check that was the open question last round is now green.

### Previous findings — all verified fixed

- [`c321877a6`](https://github.com/ariakit/ariakit/commit/c321877a6) replaces both hand-written frame chains with `flushFrames`, keeping the exact counts (`flushFrames(page, 3)` and the default `flushFrames(page)`) and documenting why three frames are needed. It also folded in the pre-existing twin I mentioned as optional.
- [`58c31a2ec`](https://github.com/ariakit/ariakit/commit/58c31a2ec) documents both numbers. The `60_000` comment states the actual invariant — the deadline sits inside the 90-second slow-test budget so a missing tab names the event — which is the part that would otherwise be unrecoverable from the code.
- [`061684c11`](https://github.com/ariakit/ariakit/commit/061684c11) extracts `expectGap`. I re-checked all three call sites against the originals: `popover.x - (anchor.x + anchor.width - 1)`, `anchor.y - (popover.y + popover.height)`, and `popover.y - (anchor.y + anchor.height)` preserve the inequalities exactly, and the one-pixel tolerance now reads as a tolerance again.

### Relevant unresolved thread validated

- [#discussion_r3667394213](https://github.com/ariakit/ariakit/pull/6913#discussion_r3667394213) (`nextjs/app/dialog-nextjs-app-router/page.tsx:17`, unresolved) — **Invalid; does not reproduce.** Replied on the thread with the evidence. The report predicts a deterministic focus-restoration failure after a direct load or reload of `/login`, but that is precisely the flow `dialog-nextjs-app-router/test-browser.ts:26` asserts, and it passed on Chrome, Firefox, and Safari at `c03b85981` (identical `page.tsx`/`default.tsx`, introduced in [`c5be82cb9`](https://github.com/ariakit/ariakit/commit/c5be82cb9)) and passes on Chrome and Firefox here. A remounted `Page` could never focus the link, so the predicted failure would be total rather than intermittent.

### Merges

Both merges are clean and carry no PR-side changes. Against their first parents they bring only `app/src/lib/source-plugin.{ts,test.ts}`, `app/src/lib/source.{ts,test.ts}`, and `nextjs/package.json` from `main` (#6911, #6912, #6917, #6918). No conflict markers anywhere in `app`, `nextjs`, `packages`, or `website`, and none of the PR's own files were touched by the merge commits.

### Not published

- **`expectGap` callback parameters shadow the `popover` locator.** `const popover = q.dialog("Team meeting")` is a `Locator`, and all three call sites bind a `Box` to the same name. TypeScript catches any actual misuse, so this is readability only. *Disposition: Declined non-goal* — flagging it rather than asking for another round, since the extraction itself achieved what the finding was about.

### CI

`Vercel` is green on this head. It was the one unexplained red last round, and it resolved after merging `main`: #6911 removed the root `vercel.json` and #6917 added `website/vercel.json`, reworking the Vercel project configuration and workspace dependency declarations. I did not confirm which of those was the operative change, and it is moot now that the check passes.

`App / Test Chrome` (6m16s) and `App / Test Firefox` (5m50s) pass, along with every other completed check. `App / Test Safari` was still running when this review was written; the equivalent job passed at `c03b85981` in 14m30s, and nothing in the three fix commits or the merges touches Safari-specific paths, but it is worth a glance before merge.

### Not fully validated

Same standing gap: no local suite runs, since this review worktree has no installed dependencies and the flakiness being targeted is CI-contention-dependent. The stress-run numbers in the description are taken as reported. The Safari result on this exact head is pending.

PR #6900, opened 2026-07-28
Add Combobox input value APIs

diegohaz:
## Motivation

Combobox currently uses `value` for the text in the input and `selectedValue` for the committed selection. With Combobox Select unifying the selection APIs, naming the text state `inputValue` creates a migration path that can eventually free `value` for selection without changing behavior today.

## Solution

Add `inputValue`, `defaultInputValue`, `setInputValue`, `resetInputValue`, and `ComboboxInputValue`. Keep the existing APIs as synchronized, JSDoc-deprecated aliases, with the new names taking precedence when both are provided.

```diff
-<ComboboxProvider value={query} setValue={setQuery}>
+<ComboboxProvider inputValue={query} setInputValue={setQuery}>
   <Combobox />
-  <ComboboxValue />
+  <ComboboxInputValue />
 </ComboboxProvider>
```

Update internal consumers and representative examples to the canonical names, while preserving legacy controlled callbacks, connected stores, default-state conflict validation, reset behavior, and keyboard focus.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the synchronized compatibility alias design</summary>

**Assessment**

The third finding was low severity but deterministic: every generated `ComboboxValue` reference lost its `store` and `children` props. This metadata is part of the supported soft-deprecation path, and its likely impact is on users consulting the legacy reference while migrating. A local typed wrapper and focused assertions restore it with minimal implementation and maintenance cost.

The synchronized state keys, controlled-prop precedence, connected-store handling, and conflict validation carry intrinsic complexity, but that complexity follows directly from preserving `value` as a supported alias. The first review cycle identified compatibility gaps in that contract; the next two cycles found static changeset and reference-metadata omissions rather than architectural instability.

**Decision**

Continue the current design and preserve `ComboboxValue` as a typed wrapper around `ComboboxInputValue`. Do not broaden the reference extractor, narrow the compatibility contract, or leave any known migration edge case unsupported.

</details>

## Verification

- `pnpm lint`
- `pnpm tsc`
- Core Combobox store tests
- Combobox React sandbox tests, including legacy tabs
- JSDoc and website reference metadata tests
- `pnpm changeset status`
- `pnpm build`
- `pnpm build-website-lite`
- `pnpm build-app-lite`
- Package dry-runs and public export resolution

PR #6923, opened 2026-07-28
Fix Combobox inline autocomplete caret reset

### LINKED ISSUE #6916: Combobox inline autocomplete overwrites a user-adjusted caret
## Current behavior

On current `main`, a `Combobox` with `autoComplete="both"` and `autoSelect` can reselect its inline completion when delayed search results commit. If the user moved the caret before that update, their next keystrokes replace the completion.

A later investigation confirmed that this is also reproducible with the latest npm release, `@ariakit/react@0.4.35`, when its legacy controlled `value` and `setValue` props are used. PR #6900 exposed but did not introduce the bug.

Before the delayed results commit, the input value is `apple` with the caret at `2..2`. After they commit, the selection unexpectedly becomes `2..5` (`ple`).

## Steps to reproduce the bug

1. Open the [StackBlitz reproduction](https://stackblitz.com/edit/github-eunt2w5l?file=src%2Fapp.tsx).
2. Focus the **Recipe** combobox and type `ap`.
3. Immediately press <kbd>ArrowLeft</kbd> to move the caret before `ple`.
4. Wait for the status to say **Results updated**.
5. Type `pp`.

The input becomes `appp`.

## Expected behavior

The delayed result update should preserve the user-adjusted caret. Typing `pp` should produce `apppple`.

## Possible solution

Avoid rerunning the completion-highlighting effect when rendered results update without changing whether the first item is auto-selected. See #6923.

### Workaround

See the [triage comment](https://github.com/ariakit/ariakit/issues/6916#issuecomment-5105762978) for a verified `autoComplete="list"` workaround. The [fix pull request](https://github.com/ariakit/ariakit/pull/6923#workaround) also documents an `ArrowLeft`-specific workaround that retains inline completion.

diegohaz:
## Motivation

A controlled [`Combobox`](https://ariakit.com/reference/combobox) with `autoComplete="both"` and `autoSelect` can reselect its inline completion when delayed search results commit. If the user moved the caret before that update, their next keystrokes replace the completion.

The same behavior is present in the published `@ariakit/react@0.4.35` package when its legacy `value` and `setValue` controlled props are used, so #6900 exposed but did not introduce the bug.

Fixes #6916.

## Solution

Add a delayed-results variant to the existing inline-autocomplete sandbox and an authoritative browser regression test that follows the reported interaction.

Derive whether the first item is auto-selected once, then make the inline-value memo and completion-highlighting effect depend on that boolean instead of the `renderedItems` array identity. Replacing the results with an equivalent list no longer reruns the highlighting effect or overwrites a user-adjusted caret. A real completion-target change still reruns the effect through the derived boolean or `inlineActiveValue`.

Add patch changesets for `@ariakit/react-components` and `@ariakit/react` because the later investigation confirmed that the published package is affected.

## Preview

- [Before: current-main browser reproduction](https://github.com/user-attachments/assets/007a66a5-4e88-4104-8479-abe2cf61c955)
- [After: caret preserved on the fixed revision](https://github.com/user-attachments/assets/809a4b88-fbd5-46d2-a9b9-e57f6b7dcea9)

## Workaround

Commit the displayed inline completion to controlled state when the reported `ArrowLeft` interaction occurs:

```tsx
const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
  if (event.key !== "ArrowLeft") return;
  // TODO: Remove after https://github.com/ariakit/ariakit/issues/6916 is fixed.
  setInputValue(event.currentTarget.value);
};

<Ariakit.Combobox
  autoComplete="both"
  autoSelect
  onKeyDown={handleKeyDown}
/>;
```

This preserves inline completion and the native caret movement, but it is intentionally limited to the reported `ArrowLeft` keyboard path and commits the displayed completion earlier than normal. If inline completion is not required, switching to `autoComplete="list"` is a simpler workaround.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test-react app/src/sandbox/combobox-inline-autocomplete/test.ts` (3 tests)
- `pnpm -F app test src/sandbox/combobox-inline-autocomplete/test-browser.ts` (Chrome, Firefox, and Safari)
- `pnpm build`
- `pnpm clean`
- `pnpm test-react` (947 tests)
- `pnpm test-react18` (947 tests)
- `pnpm changeset status`

### DIFF
--- a/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
+++ b/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
@@ -11,6 +11,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await combobox.pressSequentially("ap");
     await test.expect(q.status()).toHaveText("Updating results…");
     await combobox.press("ArrowLeft");
+    await test.expect(q.status()).toHaveText("Updating results…");
     await test.expect(q.status()).toHaveText("Results updated");
     await combobox.pressSequentially("pp");
     await test.expect(combobox).toHaveValue("apppple");
--- a/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
+++ b/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
@@ -1,0 +1,18 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/6916
+  test("preserves a user-adjusted caret when delayed results update", async ({
+    q,
+  }) => {
+    const combobox = q.combobox("Recipe");
+
+    await combobox.click();
+    await combobox.pressSequentially("ap");
+    await test.expect(q.status()).toHaveText("Updating results…");
+    await combobox.press("ArrowLeft");
+    await test.expect(q.status()).toHaveText("Results updated");
+    await combobox.pressSequentially("pp");
+    await test.expect(combobox).toHaveValue("apppple");
+  });
+});
--- a/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
+++ b/app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
@@ -9,8 +9,10 @@ withFramework(import.meta.dirname, async ({ test }) => {
 
     await combobox.click();
     await combobox.pressSequentially("ap");
-    await test.expect(q.status()).toHaveText("Updating results…");
+    await test.expect(combobox).toHaveValue("apple");
     await combobox.press("ArrowLeft");
+    // Keep results pending until after the caret moves. Otherwise, the final
+    // assertion could pass without exercising the highlighting effect.
     await test.expect(q.status()).toHaveText("Updating results…");
     await test.expect(q.status()).toHaveText("Results updated");
     await combobox.pressSequentially("pp");
--- a/.changeset/300-combobox-inline-caret.md
+++ b/.changeset/300-combobox-inline-caret.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Combobox`](https://ariakit.com/reference/combobox) inline autocomplete to preserve a user-adjusted caret when asynchronously rendered results update.
--- a/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
+++ b/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
@@ -9,7 +9,7 @@ function DelayedResults() {
   const pending = inputValue !== query;
 
   useEffect(() => {
-    const timeout = window.setTimeout(() => setQuery(inputValue), 1000);
+    const timeout = window.setTimeout(() => setQuery(inputValue), 2000);
     return () => window.clearTimeout(timeout);
   }, [inputValue]);
 
--- a/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
+++ b/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
@@ -1,4 +1,5 @@
 import * as Ariakit from "@ariakit/react";
+import type { KeyboardEvent } from "react";
 import { useEffect, useMemo, useState } from "react";
 
 const recipes = ["Apple", "Grape", "Orange", "Strawberry", "Watermelon"];
@@ -20,13 +21,22 @@ function DelayedResults() {
     );
   }, [query]);
 
+  const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+    if (event.key !== "ArrowLeft") return;
+    setInputValue(event.currentTarget.value);
+  };
+
   return (
     <Ariakit.ComboboxProvider
       inputValue={inputValue}
       setInputValue={setInputValue}
     >
       <Ariakit.ComboboxLabel>Recipe</Ariakit.ComboboxLabel>
-      <Ariakit.Combobox autoComplete="both" autoSelect />
+      <Ariakit.Combobox
+        autoComplete="both"
+        autoSelect
+        onKeyDown={handleKeyDown}
+      />
       <div role="status">
         {pending ? "Updating results…" : "Results updated"}
       </div>
--- a/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
+++ b/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
@@ -1,4 +1,43 @@
 import * as Ariakit from "@ariakit/react";
+import { useEffect, useMemo, useState } from "react";
+
+const recipes = ["Apple", "Grape", "Orange", "Strawberry", "Watermelon"];
+
+function DelayedResults() {
+  const [inputValue, setInputValue] = useState("");
+  const [query, setQuery] = useState("");
+  const pending = inputValue !== query;
+
+  useEffect(() => {
+    const timeout = window.setTimeout(() => setQuery(inputValue), 1000);
+    return () => window.clearTimeout(timeout);
+  }, [inputValue]);
+
+  const matches = useMemo(() => {
+    const normalizedQuery = query.toLowerCase();
+    return recipes.filter((recipe) =>
+      recipe.toLowerCase().includes(normalizedQuery),
+    );
+  }, [query]);
+
+  return (
+    <Ariakit.ComboboxProvider
+      inputValue={inputValue}
+      setInputValue={setInputValue}
+    >
+      <Ariakit.ComboboxLabel>Recipe</Ariakit.ComboboxLabel>
+      <Ariakit.Combobox autoComplete="both" autoSelect />
+      <div role="status">
+        {pending ? "Updating results…" : "Results updated"}
+      </div>
+      <Ariakit.ComboboxPopover>
+        {matches.map((recipe) => (
+          <Ariakit.ComboboxItem key={recipe} value={recipe} />
+        ))}
+      </Ariakit.ComboboxPopover>
+    </Ariakit.ComboboxProvider>
+  );
+}
 
 export default function Example() {
   const combobox = Ariakit.useComboboxStore();
@@ -35,6 +74,7 @@ export default function Example() {
           🍉 Watermelon
         </Ariakit.ComboboxItem>
       </Ariakit.ComboboxPopover>
+      <DelayedResults />
     </>
   );
 }
--- a/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
+++ b/app/src/sandbox/combobox-inline-autocomplete/index.react.tsx
@@ -1,5 +1,4 @@
 import * as Ariakit from "@ariakit/react";
-import type { KeyboardEvent } from "react";
 import { useEffect, useMemo, useState } from "react";
 
 const recipes = ["Apple", "Grape", "Orange", "Strawberry", "Watermelon"];
@@ -21,22 +20,13 @@ function DelayedResults() {
     );
   }, [query]);
 
-  const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
-    if (event.key !== "ArrowLeft") return;
-    setInputValue(event.currentTarget.value);
-  };
-
   return (
     <Ariakit.ComboboxProvider
       inputValue={inputValue}
       setInputValue={setInputValue}
     >
       <Ariakit.ComboboxLabel>Recipe</Ariakit.ComboboxLabel>
-      <Ariakit.Combobox
-        autoComplete="both"
-        autoSelect
-        onKeyDown={handleKeyDown}
-      />
+      <Ariakit.Combobox autoComplete="both" autoSelect />
       <div role="status">
         {pending ? "Updating results…" : "Results updated"}
       </div>
--- a/packages/ariakit-react-components/src/combobox/combobox.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox.tsx
@@ -231,6 +231,13 @@ export const useCombobox = createHook<TagName, ComboboxOptions>(
     const items = useStoreState(store, "renderedItems");
     const open = useStoreState(store, "open");
     const contentElement = useStoreState(store, "contentElement");
+    // Depend on this boolean in the highlighting effect so equivalent item
+    // updates don't re-highlight a user-adjusted caret.
+    const firstItemAutoSelected = isFirstItemAutoSelected(
+      items,
+      inlineActiveValue,
+      autoSelect,
+    );
 
     // The current input value may differ from state.inputValue when
     // autoComplete is either "both" or "inline", in which case it will be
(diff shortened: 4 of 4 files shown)


app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
```diff
@@ -0,0 +1,18 @@
...
+  // https://github.com/ariakit/ariakit/issues/6916
+  test("preserves a user-adjusted caret when delayed results update", async ({
+    q,
+  }) => {
+    const combobox = q.combobox("Recipe");
+
+    await combobox.click();
+    await combobox.pressSequentially("ap");
+    await test.expect(q.status()).toHaveText("Updating results…");
+    await combobox.press("ArrowLeft");
```
diegohaz, line 13: This test only exercises the regression when the <kbd>ArrowLeft</kbd> keypress reaches the input *before* the 1s debounce in `DelayedResults` commits, and nothing enforces that ordering.

`await test.expect(q.status()).toHaveText("Updating results…")` on line 12 proves the results were still pending at that instant, but the timer keeps running during the round trip that dispatches this keypress. If it fires first, the sequence becomes: results commit (caret still `2..5`) → <kbd>ArrowLeft</kbd> collapses to `2..2` → no further result update ever runs the highlighting effect → `pressSequentially("pp")` produces `apppple` and line 16 passes. That is exactly the assertion that failed on `2cca3cec9` before the fix, so a lost race turns this into a green test on a build that still has the bug rather than a red one.

The failure mode is silent loss of coverage, not a visible flake, so it would not show up as a CI failure to investigate. Two cheap options:

- Widen the margin by raising the debounce in `index.react.tsx` line 12 (for example `2000`), which costs about one extra second of wall time and no flake risk.
- Or make a lost race fail loudly instead of passing vacuously, by re-asserting the pending state after the keypress:

```diff
     await combobox.press("ArrowLeft");
+    await test.expect(q.status()).toHaveText("Updating results…");
     await test.expect(q.status()).toHaveText("Results updated");
```

Real-world impact: this is coverage durability rather than shipped behavior. The consumer scenario is a maintainer landing a future change to the completion-highlighting effect in `packages/ariakit-react-components/src/combobox/combobox.tsx` that reintroduces the reselection; on the run where the race is lost, this suite reports green and #6916 regresses for controlled `Combobox` users with `autoComplete="both"` and `autoSelect`.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the product fix is correct and only the regression guard is weakened.
- Priority: P2 (estimate, 80% confidence), because the test is meaningful today and the hardening is not required for this fix to be correct.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because roughly 900ms of the 1s window normally remains after `pressSequentially("ap")` and the status assertion, so it needs a sub-second stall in a single Playwright action; the desktop projects run with `workers: "100%"` on CI, where such stalls do happen.
- Fix effort: Easy (estimate, 95% confidence), because it is a one-line change in the sandbox or the test.

diegohaz (commented):
## Review summary

Initial review. Baseline: merge base `c3cea84a0b9a1c49d034dc1780c730eec89923b2` to head `0a2a0d5c29dffc489865e865840698f4c7390d1d` (4 files), plus the surrounding `useCombobox` implementation, the sandbox fixture contract, the app Playwright and Vitest configuration, and the preview index.

There were no existing review threads on this PR, so there were no unresolved candidate findings to validate.

### Verdict

The library change is sound and, as far as I can tell, the best available fix. Removing `items` and `autoSelect` from the memo and effect dependencies is not a heuristic narrowing: after hoisting `isFirstItemAutoSelected`, neither body reads `items` or `autoSelect` anymore, so the remaining dependency lists are exactly the values each body consumes. The memo output stays a pure function of `inline`, `canInline`, `firstItemAutoSelected`, `inlineActiveValue`, and `storeInputValue`, and a genuine completion-target change still reaches the effect: if the first item's value changes while `activeValue` is unchanged the boolean flips, and if `activeValue` changes `inlineActiveValue` changes. I considered wrapping the derivation in `useMemo` and gating the effect on the element's current selection instead; both are more machinery for the same behavior, so the current shape is preferable.

One inline comment on the browser test. Everything else I looked at is either fine or dispositioned below.

### Checks run

- `pnpm install --frozen-lockfile`
- `pnpm test-react app/src/sandbox/combobox-inline-autocomplete/test.ts` (3 passed), confirming the pre-existing happy-dom tests still pass with the second combobox in the fixture.
- Regression proof: the new browser test failed on the test-only commit `2cca3cec9` for the expected reason, on all three attempts, in `App / Test Firefox`: `expect(locator).toHaveValue(expected) failed` at `test-browser.ts:16`. It passes on head across Chrome, Firefox, and Safari. Chrome and Safari were cancelled on `2cca3cec9`, so the red check is proven on Firefox only.
- Experiment: authored a happy-dom duplicate of the browser test in `app/src/sandbox/combobox-inline-autocomplete/test.ts`, asserting the value and `selectionStart`/`selectionEnd` at each step. It passed both with the fix and with `packages/ariakit-react-components/src/combobox/combobox.tsx` reverted to the merge-base version. Instrumenting the reverted run showed the rendered options going from 5 to 2 with the selection staying at `2..2` and the value at `apple`, so the item update really does commit and happy-dom still does not reproduce the reselection. All experiment changes were reverted; the working tree is clean.

### Dispositions for candidates not published inline

- **Missing happy-dom duplicate** — Invalid. `ariakit-ariakit-example-playwright-test` asks for a duplicate whenever happy-dom can exercise the same behavior, and `@ariakit/test` does model caret collapse on <kbd>ArrowLeft</kbd> and selection-replacing typing, so this looked plausible. The experiment above disproves it: the duplicate cannot distinguish the fixed build from the broken one, so it would be vacuous coverage. Browser-only is the correct choice here.
- **Visual snapshot churn from extending the sandbox** — Invalid. `getPreviewIndexPaths` in `app/src/lib/preview-index.ts` filters out `SANDBOX_PREVIEW_KIND`, so `previews-browser.ts` never screenshots this preview, and no baseline exists for it under `app/src/tests/__screenshots__/`.
- **`isFirstItemAutoSelected` now runs on every render**, including for the default `autoComplete="list"` path where the memo previously short-circuited before calling it — Declined non-goal. It early-returns on `!autoSelect`, and otherwise `items.find` short-circuits on the first enabled item with a value, so the added work is a constant-time check. `Performance / Chrome (combobox-perf)` is green.
- **The comment above the derivation says "in the highlighting effect"**, but the boolean also feeds the `inputValue` memo — Declined non-goal. Low value, though a future reader could take it as effect-only justification and re-inline the call in the memo.
- **Changeset wording is broader than the fix.** "when asynchronously rendered results update" reads as unconditional, but a result update that changes the completion target still re-runs the effect and moves the caret. Declined non-goal: changelog entries summarize, and the common case the entry describes is the one that is fixed.
- **`ComboboxInput` is a public alias of `Combobox` in `@ariakit/react`** and is not named in the changeset — Declined non-goal. It is the same component under another name, so naming it adds nothing for readers.
- Fixture lifecycle, changeset, and credit checks all pass: extending the existing semantic `combobox-inline-autocomplete` sandbox with a sibling variant is step 1 of the lifecycle and leaves the original scenario untouched; the issue URL sits beside the test it documents; `300-` is the right prefix for a patch entry; `https://ariakit.com/reference/combobox` returns 200; and issue #6916 has no qualifying contributors, since its only participants are `diegohaz` and bots.

### Not fully validated

- I did not run the Playwright suite locally. Doing so needs an Astro preview build plus the OpenNext Next.js preview server that `app/playwright.config.ts` starts on port 3000, which is disproportionate here given that CI covers the exact head SHA and the pre-fix commit gives a stronger red check than a local run would.
- I did not re-run `pnpm test-react`, `pnpm test-react18`, `pnpm test-solid`, `pnpm tsc`, or `pnpm build` locally; all are green on head in CI.
- I did not independently re-verify the two preview recordings linked from the PR description.

app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
diegohaz: Fixed in [23e139ce4](https://github.com/ariakit/ariakit/commit/23e139ce418ce87df409d1150c7447e2e09cbcf0): the test now reasserts the pending state after ArrowLeft, so a lost race fails loudly, and the delay is widened to 2 seconds to reduce scheduling pressure. I also rebuilt the known-broken implementation and confirmed the strengthened test fails in Chrome, Firefox, and Safari (appp versus apppple), then passes in all three with the fix.

diegohaz:
**Fixed**

- [23e139ce4](https://github.com/ariakit/ariakit/commit/23e139ce418ce87df409d1150c7447e2e09cbcf0): Hardened the caret regression test so the async update cannot silently win before ArrowLeft ([thread reply](https://github.com/ariakit/ariakit/pull/6923#discussion_r3667740464)).

app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
```diff
@@ -0,0 +1,19 @@
...
+    await test.expect(q.status()).toHaveText("Updating results…");
```
diegohaz: The ordering guard added in `23e139ce4` works, but it now reads as an accidental copy-paste: lines 12 and 14 are byte-identical with nothing explaining why the assertion appears twice. That is the shape most likely to be deleted as duplication in a later cleanup, which would silently restore the vacuous-pass window the commit was meant to close.

Line 12 is also redundant. `pending` is `inputValue !== query`, `inputValue` does not change between the last keystroke on line 11 and line 15 (no input event, no click commit, no blur in that span), and `query` only changes when the debounce fires. So `pending` goes `true` to `false` exactly once in this window, and line 14 already implies line 12.

Line 12 does do one useful thing that line 14 does not: it forces a settle after `pressSequentially` before the caret moves. Asserting the displayed completion instead covers that better, because it proves the completion actually rendered before <kbd>ArrowLeft</kbd> rather than only proving that some render with `pending` committed. That precondition is currently unasserted, even though the whole scenario depends on it. I confirmed the value at that point is `apple` with the selection at `2..5` while validating the happy-dom equivalent in the previous review, and the triage comment on #6916 records the same state.

Suggested shape, which drops the duplication, strengthens the precondition, and keeps the guard where it matters:

```diff
     await combobox.pressSequentially("ap");
-    await test.expect(q.status()).toHaveText("Updating results…");
+    await test.expect(combobox).toHaveValue("apple");
     await combobox.press("ArrowLeft");
+    // The results must still be pending here. If they commit before the caret
+    // moves, nothing re-runs the completion-highlighting effect afterwards and
+    // the final assertion would pass on a build that still reselects.
     await test.expect(q.status()).toHaveText("Updating results…");
     await test.expect(q.status()).toHaveText("Results updated");
```

This also matches how `#6913` stabilized the neighbouring browser tests, where each added guard asserts something the surrounding lines do not and carries a comment when the reason is not self-evident.

Real-world impact: coverage durability rather than shipped behavior, same as the previous finding. The consumer is a maintainer reading this file later, collapsing the two identical lines, and reopening the race that lets a future regression of #6916 ship green.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the guard is correct today and only its resilience to later edits is at stake.
- Priority: P2 (estimate, 85% confidence), because nothing here blocks the fix.
- Breakage likelihood: Possible (estimate, 65% confidence), because verbatim duplicate assertions are a common cleanup target and there is no comment to stop it.
- Fix effort: Easy (estimate, 95% confidence), because it is a one-line swap plus a comment.

diegohaz (commented):
## Re-review summary

Re-review. Baseline: the head SHA recorded in the previous marked review, `0a2a0d5c29dffc489865e865840698f4c7390d1d`, to the current head `23e139ce418ce87df409d1150c7447e2e09cbcf0` (`23e139ce4`, "Harden combobox caret regression test"). The incremental diff touches two files: the debounce in `app/src/sandbox/combobox-inline-autocomplete/index.react.tsx` goes from `1000` to `2000`, and `test-browser.ts` gains a repeated pending-status assertion after the <kbd>ArrowLeft</kbd> press. The changeset and `packages/ariakit-react-components/src/combobox/combobox.tsx` are unchanged since the previous review, and I confirmed nothing outside this sandbox references it.

### Previous finding

The only prior thread is resolved, so I did not re-adjudicate it, but the incremental diff is its fix, so I checked it as part of this review. Both halves hold. `pending` is `inputValue !== query`, and `inputValue` cannot change between the last keystroke and the `Results updated` assertion, so re-asserting the pending state after the caret moves does convert a lost race from a silent pass into a visible failure. Doubling the debounce roughly doubles the race margin, from about 900ms to about 1900ms of the window remaining once the keypress is dispatched. The longer wait still fits Playwright's default 5s `expect` timeout with roughly 3s of headroom, since `app/playwright.config.ts` sets no global expect timeout. Taken together the expected flake rate is lower than the previous silent-pass rate, which is the right trade.

### New findings

One inline comment on `test-browser.ts` lines 12 to 14: the guard is correct but is expressed as two byte-identical assertions with no explanation, line 12 is logically subsumed by line 14, and the precondition that actually matters, that the completion rendered as `apple` before the caret moved, is still unasserted. The suggested shape removes the duplication and strengthens the barrier at the same time. Not blocking.

### Checks run

- Fetched and reset the review worktree to `23e139ce4`; working tree clean.
- `pnpm test-react app/src/sandbox/combobox-inline-autocomplete/test.ts` (3 passed), confirming the longer timer does not disturb the pre-existing happy-dom tests.
- Reviewed CI on the current head: `App / Test Chrome`, `App / Test Firefox`, `App / Test Safari`, `App / Test iOS`, `App / Test Android`, `Main / Test`, `Main / Test React`, `Main / Test React 18`, `Main / Test Solid`, `Main / Lint`, and `Main / Type check` all pass. `App / Visual / Finalize`, `Docs / Generate docs`, and `Release preview / Build` were still running when I looked.
- Compared the hardening against `c483d81de` ("Stabilize flaky browser tests", #6913) to check it against how the neighbouring browser tests were stabilized.

### Dispositions for candidates not published inline

- **The doubled debounce shrinks headroom on the final `toHaveText("Results updated")` from about 4s to about 3s inside the default 5s expect timeout** — Declined non-goal. 3s of headroom over a real 2s timer is ample, and the desktop projects carry retries (1 for Chrome, 2 for Firefox, 3 for Safari on CI).
- **The PR description's Verification list was not updated for this commit** — Declined non-goal. The listed commands are unchanged by the hardening, so re-running them produces the same list.
- Everything dispositioned in the previous review still stands; nothing in the incremental diff changes those conclusions.

### Not fully validated

- I still did not run the Playwright suite locally, for the same reason as last time: it needs an Astro preview build plus the OpenNext Next.js preview server on port 3000, and CI covers the exact head SHA.
- The claim that line 12 is subsumed by line 14 is reasoning from the sandbox's state flow plus the happy-dom validation from the previous review, not a Playwright run of the suggested shape.

app/src/sandbox/combobox-inline-autocomplete/test-browser.ts
diegohaz: Fixed in [`c41415442`](https://github.com/ariakit/ariakit/commit/c41415442a91d01ccb1bf134c90d1a9688070ee8): the test now proves the inline completion rendered before ArrowLeft and documents why results must remain pending until after the caret moves. I also confirmed the revised test fails against the known-broken implementation and passes with the fix in Chrome, Firefox, and Safari.

diegohaz:
**Fixed**

- [`c41415442`](https://github.com/ariakit/ariakit/commit/c41415442a91d01ccb1bf134c90d1a9688070ee8) ([reply](https://github.com/ariakit/ariakit/pull/6923#discussion_r3667975489)): Replaced the duplicate pending assertion with an inline-completion precondition and documented the ordering guard.

PR #6950, opened 2026-07-29
Run type-aware lint rules only on commands that cannot write

### LINKED ISSUE #6928: pnpm lint-fix silently rewrites source when run after a filtered install
## Context

Observed while implementing #6925 (PR #6927). That work only needed the website, so the agent installed with the same filter the Build website CI job uses, `pnpm install --fail-if-no-match --frozen-lockfile --filter 'website...'` with `pnpm_config_verify_deps_before_run=false`, and then ran `pnpm run lint-fix` before committing, as the repository workflow requires.

This matters more now than it used to. #6894, #6898, #6907, and #6911 have been steadily narrowing what each CI job installs, and contributors reproducing a job locally are increasingly likely to be linting against a partially installed workspace.

## Problem

`pnpm run lint-fix` rewrote seven `!` non-null assertions in two files the change never touched, in a workspace that was not installed.

```
oxlint.config.ts
  options: { typeAware: true }
```

With `--filter 'website...'`, `app/node_modules` does not exist, so `@playwright/test` (declared by `app` at `1.62.0`) is unresolvable. Type-aware rules then run against degraded types, and every occurrence of this shape was rewritten:

Before:

```ts
await expect(combobox).toHaveAttribute(
  "aria-activedescendant",
  (await apple.getAttribute("id"))!,
);
```

After:

```ts
await expect(combobox).toHaveAttribute(
  "aria-activedescendant",
  (await apple.getAttribute("id")),
);
```

The assertion is load-bearing. Playwright's `getAttribute` returns `Promise<string | null>` and `toHaveAttribute` expects a `string`, so removing `!` is a type error once `app` is actually installed.

Affected lines, verified against current `main`:

- `app/src/sandbox/combobox-select-4767/test-browser.ts` — 17, 28, 40
- `app/src/sandbox/combobox-select-open-active-item/test-browser.ts` — 16, 32, 59, 69

### Why the exit code is not a safety net

Two properties make this quiet rather than loud, and they are the reason this is worth fixing rather than just knowing about:

**The failing files and the rewritten files were disjoint.** The non-zero exit came entirely from unfixable `typescript(no-redundant-type-constituents)` errors in seven other files (`app/env.d.ts`, `app/src/middleware.ts`, `app/src/actions/admin.ts`, `app/src/pages/admin/_utils.ts`, `app/src/pages/api/stripe-webhook.ts`, and two `menu-slide-navigation` browser tests). Nothing in the output pointed at the two files that were modified. Had the degraded types produced only auto-fixable violations, the command would have exited `0` and rewritten those seven lines with no signal at all.

**Fixes are written to disk before the failure.** `lint-fix` is `oxlint --disable-nested-config --fix && oxfmt`. oxlint persisted its rewrites and then exited non-zero, so `&&` short-circuited and `oxfmt` never ran. A failed lint run does not roll back what `--fix` already wrote; the leftover redundant parens above are the visible trace of that.

It was caught only because `git status --short` happened to be chained into the same command, so two unexplained ` M app/src/sandbox/...` entries printed under the error wall.

### Precedent in the config

`oxlint.config.ts` already carries a mitigation for the same underlying failure mode:

```ts
{
  // Disable this rule for the app because some types depend on the app
  // being built first, and linting may run before the app is built.
  files: ["app/src/lib/**/*.ts"],
  rules: { "no-redundant-type-constituents": "off" },
}
```

That case degrades to a false error. This one degrades to a silent source rewrite, because the rule involved is auto-fixable.

## What is not established

- **The responsible rule was never named.** The `--fix` run printed no per-fix rule attribution. The only rule ID in the output was `no-redundant-type-constituents`, on the disjoint set of files above. Identifying the rule is the first step of any fix.
- **Check-only `pnpm lint` was not tested under the partial install.** Whether it also flags these lines, and therefore whether the check and fix paths disagree, is unknown.
- **The behavior was not deliberately reproduced.** It was observed once, reverted with `git checkout HEAD -- <paths>`, and a re-run after installing `app...` and `./templates/*...` was clean with zero modifications. Determinism is unconfirmed.

Both files are pristine on `main`; nothing was committed. This is a latent hazard, not present damage.

## Why this can be separate

Nothing in the repository is broken, PR #6927 is unaffected, and its diff contains only the two `website/build-pages/` files. This is a tooling-safety problem in the lint workflow rather than a defect in any shipped code or in that change.

## Expected outcome

`pnpm run lint-fix` either has the type information its type-aware rules require, or it refuses to rewrite files whose types it cannot resolve. Silently rewriting source based on degraded type information is the behavior to eliminate.

Reproduce first, from a clean checkout with only `--filter 'website...'` installed, and identify the rule performing the strip. Then decide the mechanism. Candidates worth weighing against each other rather than assuming: detecting an incomplete install and failing `lint-fix` early with an actionable message; scoping the lint run to what is installed; disabling the specific rule the way `no-redundant-type-constituents` already is; or reordering `lint-fix` so fixes are not persisted from a run that cannot type-check its inputs. Whichever is chosen, add coverage that would fail if the silent-rewrite behavior returns.

Worth confirming at the same time whether other auto-fixable type-aware rules can misfire the same way, since `!`-stripping is unlikely to be the only one.

diegohaz:
## Motivation

Closes #6928.

`pnpm run lint-fix` silently rewrote seven load-bearing `!` non-null assertions in two files the running change never touched, in a workspace that was not installed.

`oxlint.config.ts` set `options.typeAware: true`. That option is read by **every** oxlint invocation, including the ones that write to disk: the `lint-fix` script, the `lint-staged` pre-commit hook, and the editor's `source.fixAll.oxc` action. Type-aware rules answer questions about types, so when a workspace is missing from a filtered install, TypeScript degrades its unresolved types to an error type that behaves like `any` and those answers become wrong. Several of the rules are auto-fixable, so a writing run persists the wrong answer to disk.

### Reproduction

From a clean checkout on `b44238029`:

```sh
pnpm_config_verify_deps_before_run=false pnpm install --fail-if-no-match --frozen-lockfile --filter 'website...'
pnpm exec oxlint --disable-nested-config --fix
git status --short
```

This reproduces deterministically and rewrites exactly the lines from the issue.

### The rule

`typescript/no-unnecessary-type-assertion`, confirmed in both directions:

```sh
# only that rule enabled: rewrites both files
pnpm exec oxlint --disable-nested-config -A all -W typescript/no-unnecessary-type-assertion --fix app/src/sandbox/combobox-select-4767/test-browser.ts app/src/sandbox/combobox-select-open-active-item/test-browser.ts
# every rule except that one: rewrites nothing
pnpm exec oxlint --disable-nested-config -A typescript/no-unnecessary-type-assertion --fix
```

### Check-only and `--fix` do not disagree, but `--fix` is the silent one

`pnpm lint` reports all seven as `warning typescript(no-unnecessary-type-assertion)`. The `--fix` run applies them and prints nothing about them, because oxlint reports only *unfixed* issues. Since they are warnings, a run containing only such violations rewrites files **and exits 0**. On an isolated fixture the writing run produced no output at all and exited `0`.

The `lint-staged` pre-commit hook is on the same path. Its exact command rewrote a tracked source file and exited `0`.

### It is a class, and larger than reported

oxlint 1.75.0 ships 59 type-aware rules; this config enables about 21 of them, and 4-5 carry an auto-fix that `--fix` applies.

Under `--filter 'root...'`, the filter the `Test React 18` CI job already uses, `lint-fix` on `b44238029` rewrote **ten** files, eight of them published library source, via two distinct rules. One of them changes behavior:

```diff
  const autoFocusOnShowProp =
-   autoFocusOnShow === false ? false : canAutoFocusOnShow && autoFocusOnShow;
+   !autoFocusOnShow ? false : canAutoFocusOnShow && autoFocusOnShow;
```

That is `packages/ariakit-react-components/src/menu/menu.tsx`, rewritten by `no-unnecessary-boolean-literal-compare`, and it takes a different branch when `autoFocusOnShow` is `undefined`. Two more, from `no-unnecessary-type-assertion`:

```diff
- return placement.split("-")[0] as BasePlacement;
+ return placement.split("-")[0];
```

```diff
- omit(parentComposite as ComboboxStore | undefined, independentKeys),
+ omit(parentComposite, independentKeys),
```

This matters more over time, not less. #6894, #6898, #6907, #6911, #6929 and #6932 have been steadily narrowing what each CI job installs, so reproducing a job locally increasingly means linting a partially installed workspace.

## Solution

Stop enabling type-aware rules globally, and enable them per command with `--type-aware` only on invocations that cannot write:

```diff
- "lint": "oxlint --disable-nested-config && oxfmt --check",
- "lint-fix": "oxlint --disable-nested-config --fix && oxfmt",
+ "lint": "oxlint --disable-nested-config --type-aware && oxfmt --check",
+ "lint-fix": "oxlint --disable-nested-config --fix && oxfmt && oxlint --disable-nested-config --type-aware",
```

`--type-aware` is exactly equivalent to `options.typeAware: true`: under the partial install both produce identical 73-line diagnostic sets.

In `lint-fix` the type-aware pass runs last, after everything that writes. That is also a small improvement on the old behavior, where a spurious degraded-type error from `oxlint --fix` short-circuited `&&` and `oxfmt` never ran, leaving the redundant parens the issue documents.

The `lint-staged` hook and the editor's fix-on-save need no edit; they read the same config and become non-type-aware automatically. The editor was arguably the worst vector, since it applied fixes on every save.

### Why this mechanism

The issue lists four candidates. Weighed against each other:

- **Disable the offending rule**, as `no-redundant-type-constituents` already is. Fixes one symptom and leaves four other auto-fixable type-aware rules, plus every future one, able to do the same thing.
- **Fail `lint-fix` early on an incomplete install.** Needs a new script that encodes which workspaces the lint run requires, duplicating a policy that lives in `.github/workflows/main.yml` and will drift from it. It also pushes against the partial-install direction the repository is deliberately moving toward, which is the main reason not to take it: the fix for a partial-install hazard should not be "install everything".
- **Scope the lint run to what is installed.** Silently reduces coverage, which trades a loud problem for a quiet one.
- **Reorder so fixes are not persisted from a run that cannot type-check its inputs.** This is that option, in its strongest form: nothing that can write ever consults type information, so there is nothing to reorder.

oxlint 1.75.0 has `--type-aware` but no flag to turn type-aware *off*, so a config-level `options.typeAware` cannot be opted out of per command. Moving it to the CLI is the only way to separate reporting from writing.

The one real cost: `lint-fix` no longer auto-applies type-aware fixes under a full install, it reports them. That is inherent to the mechanism and is documented in `oxlint.config.ts`. It is also less of a loss than it sounds, since the only auto-fixable type-aware rules enabled here are warn-level.

### Coverage

`lint.test.ts` runs the repository's own oxlint invocations against a fixture whose types cannot resolve, because it imports from a package that can never be installed. Each of these regression paths fails exactly one test:

| Regression | Failing test |
| --- | --- |
| `options.typeAware` returns to the config | `oxlint.config.ts does not enable type-aware rules on a writing run` |
| a lint script changes shape at all | `the root manifest declares exactly these oxlint commands` |
| `--type-aware` added to a writing invocation, pin updated to match | `no oxlint invocation that writes to disk enables type-aware rules` |
| writing invocations stop being recognizable, pin updated to match | the non-empty assertion in the same test |

The second test reads the live manifest rather than the pinned list, so updating the pin to accept a manifest change cannot also wave a live regression through.

### Verification

- Full install: `pnpm tsc`, `pnpm test` (62 files, 725 tests), and `pnpm lint` all pass. Before and after `pnpm lint` diagnostic sets, captured over an identical file set, are both empty and identical.
- Install-filter matrix, with `node_modules` removed before each install so the state matches a clean CI checkout, comparing `b44238029` against this branch and checking `git status --short` after every run:

| Filter | Before | After |
| --- | --- | --- |
| `website...` | rewrites 2 files | no rewrites |
| `root...` | rewrites 10 files | no rewrites |
| `./packages/*...` | rewrites 2 files | no rewrites |
| `app...` | no rewrites | no rewrites |
| `app... website... ./packages/*...` | no rewrites | no rewrites |
| `app... ./packages/*... examples... ./templates/*...` (Lint job) | no rewrites | no rewrites |

Under the filters where the old commands rewrote files, the new ones report the same violations instead of applying them: 17 additional diagnostics surface under `root...`, which is exactly the set that used to be swallowed.

No changeset: this is repository tooling with no consumer-observable effect through any published package.

## Review gate reassessment

<details>
<summary>Cycle 3: Simplify the regression coverage rather than keep patching it</summary>

**Assessment**

Three gate rounds produced 11 findings. Nine were inside `lint.test.ts`, and four of those were the same finding recurring: a hand-rolled parser that extracted oxlint invocations from `package.json` kept missing another shell shape. One of the four was *introduced* by the previous round's patch, and the round-3 proposal did not close the second example named in its own text. The production diff drew zero correctness findings across all three rounds; every reviewer independently endorsed the mechanism.

Issue severity is high per incident (silent on-disk corruption, zero exit code, published library source affected) and expected frequency is rising as CI installs narrow, so the coverage is worth having. But the parser coupled to the manifest's *authoring style*, which fails silently, rather than its *content*, which fails loudly. Its only advantage over a pinned list was automatic coverage of a new writing command, and it delivered that only for shapes it happened to model.

**Decision**

Keep the production change unchanged. Delete the parser and replace it with a pinned exact-match assertion over the manifest's oxlint-bearing commands, plus behavior tests on representative flags. Later rounds on the simplified design added a live-manifest invariant, so that updating the pin cannot wave a regression through, and a non-empty guard so that invariant cannot degrade to zero assertions.

No dispositions were weakened. The one superseded finding (broaden `indexOf("oxlint")` to catch path-form invocations) concerned a parser that no longer exists; the pinned list catches those shapes, which the parser did not.

</details>


diegohaz:
Lint should be run with the full pnpm install.

Issue #6948, opened 2026-07-29
Stop perf.test.ts from crashing a child process during a passing run


diegohaz:
## Context

Observed while implementing #6936 (PR #6947), which touches `packages/ariakit-scripts/src/react18.ts`. Running the package's other tests as part of that verification surfaced noise in a suite that reports success.

## Problem

`pnpm test packages/ariakit-scripts/src/perf.test.ts` passes — 24 tests — while a child Node process dies from an unhandled `'error'` event. The run emits six `AggregateError [ECONNREFUSED]` blocks for `localhost:3000`, each with a stack trace followed by a `Node.js v24.18.0` footer, and the suite still reports green.

An unhandled `'error'` event terminates the process that raised it. The suite passing means the failure is in a child whose exit nobody checks, which is the part worth fixing: the same silence would hide a real failure on that path.

The immediate cost is log pollution. On CI this buries genuine diagnostics in stack traces from a test that succeeded, which matters most when someone is reading those logs precisely because something else went wrong.

## What is and is not established

- **Reproduced** by the agent that found it, with the command above, on `origin/main` with both `perf.test.ts` and `perf.ts` byte-identical to that branch. It is pre-existing and unrelated to #6936.
- **Not yet traced.** Neither `packages/ariakit-scripts/src/perf.test.ts` nor `packages/ariakit-scripts/src/perf.ts` contains any reference to `localhost`, port `3000`, or a direct socket connection — checked directly. So the connection attempt originates somewhere less obvious, most likely inside Playwright machinery reached through the `@playwright/test` import at the top of the test file. Identifying the actual origin is the first step, not an assumption to carry forward.
- **Not confirmed by me.** I verified the absence of an obvious source in both files but did not run the suite.

## Why this can be separate

Pre-existing, and PR #6947 is confined to `react18.ts` and `react18.test.ts`. The suite passes, so nothing is blocked. It is also `ariakit-ariakit-performance` territory, which carries its own verification requirements.

## Expected outcome

`pnpm test packages/ariakit-scripts/src/perf.test.ts` runs clean, with no unhandled `'error'` event and no stack traces on a passing run.

Trace the origin before choosing a fix. If something is spawning a child that tries to reach a dev server on port 3000, decide whether that child should exist at all in this suite, whether it should be pointed at a real fixture, or whether the connection should simply be handled. Attaching an `'error'` handler to silence the output would be the wrong fix if the underlying attempt is itself unintended.

[...]

PR #6959, opened 2026-07-29
Run Vitest suites as named projects

diegohaz:
## Motivation

The root Vitest configuration applies Happy DOM to every suite and selects React or Solid behavior through `ARIAKIT_TEST_LOADER`:

```json
"test-react": "cross-env ARIAKIT_TEST_LOADER=react vitest"
```

This makes runtime requirements implicit. Node-only tooling tests receive browser globals unless each file overrides the environment, which is the configuration boundary exposed by #6948 and its targeted fix in #6952.

## Solution

Define named `node`, `core`, `react`, and `solid` Vitest projects with explicit environments, includes, plugins, and setup files:

```ts
test: {
  projects: [
    { test: { name: "node", environment: "node" } },
    { test: { name: "core", environment: "happy-dom" } },
    { test: { name: "react", environment: "happy-dom" } },
    { test: { name: "solid", environment: "happy-dom" } },
  ],
}
```

The existing package scripts select the appropriate projects, so `pnpm test` remains the core non-framework suite plus scripts tests, while the React, React 18, and Solid commands preserve their current responsibilities. Framework mounting now lives behind project-specific setup entry points, allowing the environment-variable loader and the per-file Node directive in `perf.test.ts` to be removed.

## Verification

- Exact collection parity with the previous configuration: 723 core/Node tests, 947 React tests, and 8 Solid tests, with no missing or additional tests
- `pnpm test` (61 files, 723 tests)
- `pnpm test-react` (183 files, 947 tests)
- `pnpm test-react18` (183 files, 947 tests)
- `pnpm test-solid` (6 files, 8 tests)
- `pnpm test packages/ariakit-scripts/src/perf.test.ts --coverage`
- `pnpm tsc`
- `pnpm lint`


packages/ariakit-scripts/src/perf.test.ts
diegohaz, line 42: We can remove this test.

vitest.setup.framework.ts
```diff
@@ -0,0 +1,107 @@
+import { beforeEach } from "vitest";
```
diegohaz, line 1: I think the file names should be `vitest.setup.framework.ts`, `vitest.setup.react.ts`, `vitest.setup.solid.ts`

vitest.setup.framework.ts
```diff
@@ -0,0 +1,107 @@
...
+- ios
+- android
+
+*/
+
+function parseTest(filename?: string) {
+  if (!filename) return false;
+  const match = filename.match(
+    /^(?<dir>(?:.*\/)?(?:examples|sandbox)\/.+?)\/((?<framework>react|solid)\.)?test\.ts$/,
+  );
```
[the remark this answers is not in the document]
diegohaz: If it's a TSX file, that signals that the test will render JSX, so we won't automatically render the example/sandbox.

vitest.config.ts
```diff
@@ -1,91 +1,113 @@
...
+        plugins: [reactPlugin()],
+        test: {
+          name: "react",
+          environment: "happy-dom",
+          exclude: testExcludes,
+          include: getFrameworkTestIncludes("react"),
+          setupFiles: [
+            join(rootDir, "vitest.setup.ts"),
+            join(rootDir, "vitest.react.setup.ts"),
+          ],
```
diegohaz: `extends: true` already merges the root `test.setupFiles` into every project, so naming `vitest.setup.ts` again here makes the `react` and `solid` projects resolve it twice and evaluate it twice for every test file they run.

Resolved config for the `react` project, read with `createVitest("test", { watch: false })` against this head:

```
setupFiles: [
  '<root>/vitest.setup.ts',
  '<root>/vitest.setup.ts',
  '<root>/vitest.react.setup.ts'
]
```

Appending a one-line `appendFileSync` probe to `vitest.setup.ts` and running a single test file confirms the double evaluation: `--project core` writes 1 line, `--project react` and `--project solid` write 2.

So `failOnConsole()` installs a second layer of console spies over the first, and `expect.extend` re-registers, on each of the 183 React files and 6 Solid files. It is benign today — I checked that a `console.error` inside a React-project test still produces exactly one failure — but it is duplicated global setup state that only stays harmless while `vitest.setup.ts` is idempotent.

The config already relies on this merge for plugins: the root `sourcePlugin` survives in the `react` project alongside `reactPlugin()`, with `vite-plugin-source` and `vite:react-babel` both present in the resolved plugin list. `setupFiles` is the one option that re-declares something already inherited.

```suggestion
          setupFiles: [join(rootDir, "vitest.react.setup.ts")],
```

The `solid` project needs the same change on lines 105-108. With both entries removed I ran `pnpm test-react` (183 files, 947 tests) and `pnpm test-solid` (6 files, 8 tests) and both stayed green, with the resolved `setupFiles` reduced to `[vitest.setup.ts, vitest.<framework>.setup.ts]`.

**Assessment**

- Severity: Low (verified), because the duplicate evaluation is currently harmless and a console failure still reports once.
- Priority: P1 (estimate, 80% confidence), because this PR introduces it and the fix is a four-line removal in the file already being changed.
- Breakage likelihood: the duplicate evaluation is Likely (verified) on every run; an actual bug from it is Rare (estimate, 85% confidence), requiring `vitest.setup.ts` to gain non-idempotent setup.
- Fix effort: Easy (verified), because removing the two entries is sufficient and both framework suites pass unchanged.

vitest.config.ts
```diff
@@ -1,91 +1,113 @@
...
     coverage: {
       include: ["packages"],
     },
+    projects: [
+      {
+        extends: true,
+        test: {
+          name: "node",
+          environment: "node",
+          exclude: testExcludes,
```
diegohaz: The `extends: true` merge that makes `setupFiles` inheritable applies to `exclude` too, so `testExcludes` does not need repeating in all four projects. Hoisting it into the root `test` block lets `node`, `react`, and `solid` drop `exclude` entirely and lets `core` declare only what is specific to it:

```ts
  test: {
    watch: false,
    testTimeout: 10_000,
    setupFiles: [join(rootDir, "vitest.setup.ts")],
    exclude: testExcludes,
    // ...
    projects: [
      {
        extends: true,
        test: { name: "node", environment: "node", include: nodeTestIncludes },
      },
      {
        extends: true,
        test: {
          name: "core",
          environment: "happy-dom",
          include: coreTestIncludes,
          exclude: coreTestExcludes,
        },
      },
      // ...
    ],
  },
```

Together with the `setupFiles` change this removes 12 lines and leaves each project declaring only what makes it different: name, environment, include, plugin, and framework setup.

I verified the shape locally. `vitest list --filesOnly` for `core`+`node`, `react`, and `solid` produced file lists byte-identical to `origin/main` (61, 183, and 6 files), and `pnpm test` (61 files, 723 tests), `pnpm test-react` (183 files, 947 tests), and `pnpm test-solid` (6 files, 8 tests) all passed.

**Assessment**

- Severity: Low (verified), because behavior is identical either way.
- Priority: P2 (estimate, 85% confidence), because it is a readability improvement rather than a defect.
- Breakage likelihood: Rare (verified), because collection is unchanged; the residual risk is a future project silently omitting one of the four copies.
- Fix effort: Easy (verified), because I ran all three suites against this shape.

vitest.config.ts
```diff
@@ -1,91 +1,113 @@
...
     setupFiles: [join(rootDir, "vitest.setup.ts")],
-    exclude: testExcludes,
-    include: testIncludes,
     sequence: {
       hooks: "parallel",
     },
     coverage: {
       include: ["packages"],
     },
+    projects: [
```
diegohaz, line 71: The rationale for happy-dom is dropped with no replacement. `main` carried it directly above the root `environment` option:

```
    // Every suite runs on happy-dom — it's ~2x faster than jsdom for the
    // @ariakit/test simulation layer and provides the DOM every suite needs,
    // including the React 18 suite (`test-react18`). happy-dom's spec
    // divergences from jsdom and real browsers are normalized by the shims in
    // `@ariakit/test` (see `packages/ariakit-test/src/shims.ts` and the
    // `window.event` shim in `dispatch.ts`, which restores React 18's
    // discrete-event priority). Individual tests can still opt into another
    // installed environment with a `// @vitest-environment <name>` comment.
```

Three projects now declare `environment: "happy-dom"` with nothing explaining why happy-dom rather than jsdom, why the React 18 suite is safe on it, or that the `@ariakit/test` shims are what make it safe. The PR also removes the repository's most visible demonstration of the per-file escape hatch, `// @vitest-environment node` in `perf.test.ts`, while `packages/ariakit-test/src/click.jsdom.test.ts` and `press.jsdom.test.ts` still rely on that hatch and now have nothing pointing at it.

This is the one place where a PR about making runtime requirements explicit makes something less explicit. A condensed version above `projects` would keep the WHY:

```ts
    // The DOM projects run on happy-dom — ~2x faster than jsdom for the
    // @ariakit/test simulation layer, and safe for the React 18 suite because
    // the shims in `@ariakit/test` normalize its spec divergences (see
    // `packages/ariakit-test/src/shims.ts` and the `window.event` shim in
    // `dispatch.ts`, which restores React 18's discrete-event priority).
    // Individual tests can still opt into another installed environment with a
    // `// @vitest-environment <name>` comment, as the `*.jsdom.test.ts` files
    // in `packages/ariakit-test` do.
    projects: [
```

**Assessment**

- Severity: Low (estimate, 90% confidence), because nothing breaks; the cost is a future contributor re-deriving or reversing a deliberate choice.
- Priority: P2 (estimate, 85% confidence), because it is a one-comment addition in a file this PR already rewrites, and the rationale is hard to reconstruct once it is only in `git log`.
- Breakage likelihood: Possible (estimate, 70% confidence), because the next person touching environments is the most likely reader, and the jsdom opt-outs in `packages/ariakit-test` give them a concrete reason to look.
- Fix effort: Easy (verified), because it is a comment.

packages/ariakit-scripts/src/perf.test.ts
diegohaz: I would keep it, and I think this PR makes it more valuable rather than less.

With `// @vitest-environment node` gone, nothing in `perf.test.ts` states its runtime requirement any more — the environment now comes from `nodeTestIncludes` matching `packages/ariakit-scripts/src/**` in a shared config file. `test("runs without DOM shims")` is the only executable assertion that the binding still holds.

I checked what happens when it breaks. Flipping the `node` project's `environment` to `"happy-dom"` and running `pnpm exec vitest run --project node packages/ariakit-scripts/src/perf.test.ts` gives:

- an `AggregateError [ECONNREFUSED]` block for `::1:3000` and `127.0.0.1:3000`, ending in a `Node.js v24.18.0` footer — the #6948 signature
- the other 24 tests still passing
- `runs without DOM shims` as the only failure

So without this test, that misconfiguration silently reproduces #6948: a child process dies, the log fills with stack traces, and the suite reports success. That is exactly the failure mode #6948 was filed about, and it is invisible unless something asserts the environment.

The assertion costs one expression and no measurable time. If the objection is that it reads as testing configuration rather than code, renaming it to something like `test("runs in the node environment")` would make its purpose explicit while keeping the guard.

**Assessment**

- Severity: Medium (verified), because losing it re-opens a class of failure where a dying child process leaves the suite green.
- Priority: P1 (estimate, 75% confidence), because the decision is being made in this PR, and it is the same PR that removes the per-file directive the test was paired with.
- Breakage likelihood: Possible (estimate, 60% confidence), because it needs a future edit to `nodeTestIncludes`, a file moving out of `packages/ariakit-scripts/src`, or a new node-only tooling package landing elsewhere.
- Fix effort: Easy (verified), because keeping the test is a no-op and renaming it is one line.

vitest.framework.setup.ts
```diff
@@ -0,0 +1,107 @@
+import { beforeEach } from "vitest";
+
+type Framework = "react" | "solid";
+
+async function tryImport(path: string) {
+  return import(path)
+    .then(({ default: component }) => ({ component, failedImport: false }))
+    .catch(() => ({ component: undefined, failedImport: true }));
```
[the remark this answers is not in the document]
diegohaz: Confirmed against this head. I appended a syntax error to `app/src/sandbox/group-semantics/index.react.tsx` and ran `pnpm exec vitest run --project react app/src/sandbox/group-semantics/test.ts`:

```
 Test Files  1 passed (1)
      Tests  1 skipped (1)
```

The transform error is swallowed, the test is skipped, and the run exits 0.

The project-scoped includes this PR introduces sharpen the problem. `getFrameworkTestIncludes` derives each example directory from an existing `index.<framework>.tsx`, so an absent entry cannot happen for a collected test, and `parseTest` returns `false` for everything outside `examples/` and `sandbox/`. The `catch` can therefore only mask real breakage. A full `pnpm test-react` run reports 947 passed with zero skips, so nothing currently depends on the tolerant path.

One caveat for whoever fixes it: the indirection through an opaque `path` parameter is load-bearing. Inlining the template literal hands the import to Vite's `dynamic-import-vars` transform and every example test fails with:

```
Error: Unknown variable dynamic import: ./app/src/sandbox/group-semantics/index.react.tsx. Note that variables only represent file names one level deep.
```

This version keeps the indirection, propagates the error, and lets the now-unreachable `false` branch go:

```ts
// The path stays behind an opaque parameter so Vite's dynamic-import-vars
// transform does not try to rewrite it into a one-level-deep glob.
async function importDefault(path: string) {
  const { default: component } = await import(path);
  return component;
}

async function loadReact(dir: string) {
  const { render } = await import("@ariakit/test/react");
  const { createElement, Suspense } = await import("react");
  const component = await importDefault(`./${dir}/index.react.tsx`);
  // ...unchanged
}
```

```ts
} satisfies Record<Framework, (dir: string) => Promise<() => void>>;
```

```ts
export function setupFrameworkTests(framework: Framework) {
  beforeEach(async ({ task, skip }) => {
    const parseResult = parseTest(task.file?.name);
    if (!parseResult) return;
    const { dir, framework: testFramework } = parseResult;
    if (testFramework !== "all" && testFramework !== framework) {
      skip();
      return;
    }
    return LOADERS[framework](dir);
  });
}
```

Verified with that patch applied: `pnpm exec vitest run --project react` (183 files, 947 tests) and `--project solid` (6 files, 8 tests) pass, `pnpm exec tsc -b` and `pnpm lint` are clean, and the broken-example case now fails with the transform error instead of skipping.

For the record, this is pre-existing behavior moved verbatim from `vitest.setup.ts`, not something this PR introduces.

**Assessment**

- Severity: High (verified), because a broken example entry produces a green suite that silently covers nothing.
- Priority: P1 (estimate, 70% confidence), because the fix is contained and lands in a file this PR already rewrites, though the behavior predates it.
- Breakage likelihood: Possible (estimate, 65% confidence), because it needs an example entry to break in a way its own test does not otherwise catch — most likely during a refactor that touches many `index.react.tsx` files at once.
- Fix effort: Easy (verified), because the patch above is ~10 lines and passes both framework suites plus type check and lint.

diegohaz (commented):
Initial review of `500f8ac4ed629f520bc9506d20d4fe936664533d`, diffed from the merge base `520acb9092f4670a18f2715938664db4a17869ef` and then expanded to the surrounding config, scripts, CI workflows, and suites.

The direction is right. Named projects with explicit environments are a better boundary than an environment variable read at config-evaluation time, and moving the whole `packages/ariakit-scripts` suite to a `node` project generalizes the targeted fix from #6952 instead of leaving a per-file directive behind.

## Verification performed

I reproduced the parity claim rather than taking it from the description.

- `vitest list --filesOnly` on both refs. After stripping the new `[project]` prefix, the collected file lists are byte-identical to `origin/main`: 61 files for `core`+`node` vs. the old default suite, 183 for `react`, 6 for `solid`.
- `pnpm test` (61 files, 723 tests), `pnpm test-react` (183 files, 947 tests), `pnpm test-solid` (6 files, 8 tests) — all pass.
- `pnpm exec tsc -b` and `pnpm lint` — clean.
- `pnpm exec vitest run --project core --project node --shard 1/2` and `2/2` split 31+30 files and 384+339 tests, so the `--shard` path used by `Main / Test` still works with two projects selected.
- `vitest run --project core --project node <file> --coverage` works, and `vitest run packages/ariakit-scripts/src/perf.test.ts` with no `--project` runs the file exactly once, in the `node` project.
- Resolved each project's config through `createVitest` to check what `extends: true` actually merges.

Also checked and fine:

- No `ARIAKIT_TEST_LOADER` or `test-loader.ts` references survive anywhere in the tree, and nothing in `.github/workflows` referenced either.
- `ariakit react18` rsyncs the repository root with a blocklist rather than a file allowlist, so the three new root-level setup files reach the React 18 workspace.
- The environment change reaches nine `packages/ariakit-scripts` test files beyond `perf.test.ts`; none of them touch DOM globals, and the whole `node` project passes.
- No changeset is needed — this does not affect shipped code.

## Findings published inline

1. `vitest.config.ts:91-94` and `105-108` — the `react` and `solid` projects re-declare `vitest.setup.ts`, which `extends: true` already inherits, so the shared setup module is evaluated twice for every test file in those projects.
2. `vitest.config.ts:70` — `exclude: testExcludes` is repeated in all four projects; hoisting it to the root `test` block removes the repetition.
3. `vitest.config.ts:64` — the happy-dom rationale comment from `main` is dropped with no replacement, including the pointer to the `// @vitest-environment <name>` escape hatch that `packages/ariakit-test/src/*.jsdom.test.ts` still depends on.

## Relevant unresolved threads validated against this head

- `packages/ariakit-scripts/src/perf.test.ts` (`We can remove this test.`) — replied. Confirmed the guard still earns its place: with the `node` project flipped to happy-dom, the #6948 `ECONNREFUSED` signature returns, 24 tests still pass, and `runs without DOM shims` is the only failure.
- `vitest.framework.setup.ts:5-8` (`tryImport` swallowing example import errors) — replied. Reproduced with a deliberately broken example entry: the test is skipped and the run exits 0. Pre-existing and moved verbatim, but the reply carries a validated patch that propagates the error, including the reason a naive inline `import()` breaks under `dynamic-import-vars`.
- `vitest.framework.setup.ts:1` (naming the setup files `vitest.setup.framework.ts`, `vitest.setup.react.ts`, `vitest.setup.solid.ts`) — still applies to this head and is accurate as written, so no reply. Nothing to add beyond it.

One documentation follow-up that lives outside this repository was registered separately.

## Not fully validated

- `App / Test Chrome` failed on this head while `main` is green. The failure is `src/sandbox/combobox-select-nextjs-app-router/test-browser.ts:101:3 › react › opens a multi-value option in a new tab @react`, a `browserContext.waitForEvent` timeout that also failed its retry, with 740 other tests passing. Nothing in this diff participates in the Playwright suite — the changed root scripts are `test`, `test-react`, and `test-solid`, while `test-browser` and `test-chrome` are untouched — so a causal link looks implausible, but I did not prove the test is flaky. Worth a re-run before merge.
- I did not run `pnpm build`, the app or website builds, or any browser suite locally, since nothing in the diff feeds them.

packages/ariakit-scripts/src/perf.test.ts
diegohaz: We should add a comment then

vitest.config.ts
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): restored a concise Happy DOM rationale, the Ariakit test shim and React 18 context, and the per-file environment override escape hatch.

vitest.framework.setup.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): framework imports now propagate evaluation errors while retaining the opaque path helper required by Vite. A targeted failing fixture now fails with Error: framework import failed, and the React and Solid suites pass.

vitest.setup.framework.ts
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): renamed the setup entries to vitest.setup.framework.ts, vitest.setup.react.ts, and vitest.setup.solid.ts.

vitest.config.ts
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): the React and Solid projects now declare only their framework setup entry; extends: true supplies the shared vitest.setup.ts exactly once. I verified the resolved config and both suites.

packages/ariakit-scripts/src/perf.test.ts
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): kept the executable environment guard and added a comment explaining that Happy DOM can crash Playwright source-map loading in a child process while the parent suite still passes.

vitest.config.ts
diegohaz: Fixed in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da): moved the common testExcludes to the root test config. Each project inherits them, while core declares only coreTestExcludes.

diegohaz:
Addressed the current review feedback in [529f074a3](https://github.com/ariakit/ariakit/commit/529f074a3751540893ed1b0e1f70c064f65b18da).

**Fixed**

- Kept the executable Node-environment guard and documented why it matters: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672017092)
- Propagated framework import errors while preserving the opaque Vite import path: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672016961)
- Renamed the framework setup files: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672017086)
- Removed the duplicate inherited root setup entries: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672017091)
- Hoisted common excludes into the root test config: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672017113)
- Restored the Happy DOM rationale and per-file override guidance: [reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672016908)

**Verification**

- pnpm test — 61 files, 723 tests
- pnpm test-react — 183 files, 947 tests
- pnpm test-react18 — 183 files, 947 tests
- pnpm test-solid — 6 files, 8 tests
- pnpm lint
- pnpm tsc
- A temporary broken framework entry failed with Error: framework import failed instead of being skipped

packages/ariakit-scripts/src/perf.test.ts
```diff
@@ -36,6 +34,8 @@ const originalValues = new Map(
   envNames.map((name) => [name, process.env[name]]),
 );
 
+// Happy DOM makes Playwright source-map loading crash a child process while
+// this parent suite still passes, so retain an executable environment guard.
```
diegohaz: Keeping the guard is the right call, but this comment names the wrong mechanism, and the mechanism is the whole value of the comment.

The crash happens at module import, before any test body runs — so source-map loading is never reached. Running the file under happy-dom with a filter that matches nothing still crashes:

```
$ pnpm exec vitest run --project node packages/ariakit-scripts/src/perf.test.ts -t "zzzz-no-such-test"
AggregateError [ECONNREFUSED]: ...
    Error: connect ECONNREFUSED ::1:3000
    Error: connect ECONNREFUSED 127.0.0.1:3000
Node.js v24.18.0

 Test Files  1 skipped (1)
      Tests  25 skipped (25)
```

I isolated it with three probe files in the `node` project, each run with `environment` temporarily flipped to `"happy-dom"`:

- importing only `@playwright/test` — crashes
- importing only `./perf.ts`, which itself imports `errors` from `@playwright/test` — crashes
- importing only `vitest` — clean, passes

And the port matches happy-dom's own default origin, not anything in `perf.ts`: logging `globalThis.location.href` from a `core`-project test prints `http://localhost:3000/`, which is exactly the `::1:3000` / `127.0.0.1:3000` target in the trace.

`resolveScriptProfileSourceMaps` and `defaultLoadText` do call `fetch`, so the current wording is a plausible-looking wrong answer — someone debugging a recurrence would start there and find nothing. It also hides the part that generalizes: any node-environment file in this package that imports Playwright has the same failure, which is precisely why the environment is worth guarding.

```suggestion
// Importing `@playwright/test` under happy-dom fires a request at happy-dom's
// default origin (http://localhost:3000). The refused connection emits an
// unhandled 'error' event that kills the process before any test in this file
// runs, while the suite still reports success, so keep this guard executable.
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because the guard itself is correct and no test behavior depends on the comment.
- Priority: P2 (estimate, 80% confidence), because the comment was added in this PR specifically to explain a non-obvious guard, and correcting it now costs one edit.
- Breakage likelihood: Possible (estimate, 70% confidence), because the reader it misleads is exactly the person who hits this test failing and needs to know where to look.
- Fix effort: Easy (verified), because it is a comment.

diegohaz (commented):
Re-review of `529f074a3751540893ed1b0e1f70c064f65b18da`. Baseline is `500f8ac4ed629f520bc9506d20d4fe936664533d`, taken from the marker on my previous review, so the incremental scope is the single commit `Address Vitest project review feedback` plus any surrounding issue it newly exposes.

Every finding from the previous round is addressed, and I verified each one against this head rather than relying on the resolution:

- The `react` and `solid` projects now declare only their framework setup entry. Resolved `setupFiles` are `[vitest.setup.ts, vitest.setup.react.ts]` and `[vitest.setup.ts, vitest.setup.solid.ts, …jest-dom/vitest.mjs]`, with no duplicate.
- `exclude: testExcludes` is hoisted to the root `test` block. Resolved excludes are 3 entries for `node`, `react`, and `solid`, and 9 for `core`, so `coreTestExcludes` merges on top rather than replacing.
- The happy-dom rationale is back above `projects`, including the `// @vitest-environment <name>` escape hatch.
- `importDefault` propagates evaluation errors and keeps the opaque parameter. Appending a syntax error to `app/src/sandbox/group-semantics/index.react.tsx` now fails the test instead of skipping it, which is the inverse of what I recorded last round.
- The setup files are renamed to `vitest.setup.framework.ts`, `vitest.setup.react.ts`, and `vitest.setup.solid.ts`, and no reference to the old names survives anywhere in the tree.

Re-ran the full verification at this head, since the exclude hoist and the setup-file changes both touch collection:

- `vitest list --filesOnly` for `core`+`node`, `react`, and `solid` is still byte-identical to `origin/main` (61, 183, 6 files).
- `pnpm test` (61 files, 723 tests), `pnpm test-react` (183 files, 947 tests), `pnpm test-solid` (6 files, 8 tests) — all pass.
- `pnpm exec tsc -b` and `pnpm lint` — clean.

`App / Test Chrome` is green on this head, which settles the `combobox-select-nextjs-app-router` failure I flagged last round as unrelated flake.

## New finding

One, published inline: the comment added above `test("runs without DOM shims")` in `packages/ariakit-scripts/src/perf.test.ts` attributes the crash to Playwright source-map loading. It is actually an import-time failure — the file crashes under happy-dom with all 25 tests skipped, a probe importing only `@playwright/test` reproduces it, a probe importing only `vitest` does not, and the refused port matches happy-dom's default `http://localhost:3000/` origin rather than anything in `perf.ts`. The comment carries a corrected version.

Nothing else in the incremental diff raised a finding, and no unresolved threads remain to validate — all seven are resolved.

packages/ariakit-scripts/src/perf.test.ts
diegohaz: Fixed in [`0fc4e4d94`](https://github.com/ariakit/ariakit/commit/0fc4e4d948555420b030476fced1f07b94eeffb2): corrected the rationale to describe the import-time request to the Happy DOM default origin and the unhandled connection error emitted before any test runs.

diegohaz:
**Fixed**

- [`0fc4e4d94`](https://github.com/ariakit/ariakit/commit/0fc4e4d948555420b030476fced1f07b94eeffb2) ([reply](https://github.com/ariakit/ariakit/pull/6959#discussion_r3672245051)): Corrected the guard rationale to identify the import-time Playwright request to the Happy DOM default origin. Verified with isolated import and origin probes, the 25-test performance suite, and lint.

PR #6961, opened 2026-07-29
Stabilize flaky Chrome tests

diegohaz:
## Motivation

`Test Chrome` has been intermittently retrying or failing on unrelated pull requests. The failures in [job 90505813021](https://github.com/ariakit/ariakit/actions/runs/30429305765/job/90505813021?pr=6959) and its [successful rerun](https://github.com/ariakit/ariakit/actions/runs/30429305765/job/90507199203?pr=6959) include offscreen Combobox items that became DOM-visible before their effect-driven collection registration was published.

That race makes the next keyboard interaction run against a stale collection, producing intermittent `data-active-item` and `data-offscreen` assertion failures.

## Solution

- Wait for the relevant offscreen item to materialize.
- Use the shared `flushFrames` helper to cross the collection's following animation-frame publication before the next keyboard interaction:

  ```ts
  await expect(item).not.toHaveAttribute("data-offscreen");
  await flushFrames(page);
  ```

- Keep the existing keyboard and focus coverage unchanged.

## Testing

- `pnpm lint-fix`
- `pnpm tsc`
- 89/89 focused Chrome matrix cases with no retries
- 890/890 focused Chrome matrix cases across ten repetitions, four workers, and no retries

## Scope reassessment

<details>
<summary>Keep only the evidence-backed offscreen matrix checkpoint</summary>

I compared eight recent Chrome matrices from unrelated pull requests: seven were completely clean, and one had two retries in this same offscreen matrix. None had popup retries. By contrast, popup changes on this branch still failed or retried on four of seven Chrome heads; [job 90575795326](https://github.com/ariakit/ariakit/actions/runs/30451790707/job/90575795326?pr=6961) had one hard middle-click popup failure and two popup retries.

The same final workflow also repeated the Safari offscreen-value typeahead flake after the added frame checkpoint, so that change was not sufficient either.

The popup fixtures and the Safari offscreen-value fixture are therefore restored exactly to `main`. This PR keeps only the small matrix registration checkpoint that maps directly to the observed race and survived the focused stress run.

</details>


app/src/sandbox/combobox-select-offscreen-matrix/test-chrome.ts
```diff
@@ -22,6 +22,16 @@ async function renderMatrixCase(page: Page, label: string) {
   await page.getByRole("combobox", { name: "Matrix case" }).selectOption(label);
 }
 
+async function waitForItemRegistration(page: Page, item: Locator) {
+  await expect(item).not.toHaveAttribute("data-offscreen", {
+    timeout: 15_000,
```
diegohaz: The 15-second deadline is undocumented and does not fit the budget of the test it runs in.

These matrix tests never call `test.slow()`, and `app/playwright.config.ts` sets no `timeout` override, so each one runs under Playwright's default 30-second test timeout. The 8 searchable `defaultValue` cases with `autoSelect: false` (2 offscreen modes x 2 `unmountOnHide` x 2 group values) call `waitForItemRegistration` twice: once in the "move with keyboard" step and once in the "typeahead" step. Two 15-second waits alone consume the whole 30-second budget, so a genuinely stuck registration reports `Test timeout of 30000ms exceeded` rather than naming the `data-offscreen` assertion that actually failed. That is the same diagnosability loss the comment deleted from `app/src/sandbox/combobox-links/test-browser.ts` was guarding against.

It is also unclear why the value moved at all. The PR description attributes the flake to the frame-delayed collection publish, not to a slow `data-offscreen` removal, and this assertion already passed under the default 5-second expect timeout before this change. The new `flushFrames` call is what addresses the described race.

Concretely, one of:

- Drop the explicit timeout and let the default 5 seconds apply, so a stuck registration fails with the attribute assertion in the message.
- Keep the extra headroom but add `test.slow()` to the matrix tests and a comment recording why 15 seconds is needed and how it fits the budget, the way `flushFrames(page, 3)` in `app/src/sandbox/combobox-select-offscreen-value/test-browser.ts` documents its own non-default value.

**Consumer impact**: this hits maintainers triaging a red `Test Chrome` job. A stuck offscreen registration currently surfaces as a bare 30-second test timeout with no indication that `data-offscreen` was the blocker, which is the exact triage cost this PR is trying to reduce.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the test still fails correctly; only the failure message loses the specific cause.
- Priority: P2 (estimate, 80% confidence), because it degrades diagnosability rather than correctness.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because it needs a registration slow enough to exhaust the first 15-second wait under CI contention.
- Fix effort: Easy (estimate, 90% confidence), because it is a timeout value plus a comment, or `test.slow()`.

app/src/sandbox/combobox-links/test-browser.ts
```diff
@@ -71,30 +112,32 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
-        await expect(page).toHaveURL(/https:\/\/ariakit\.com/);
+        await expect(page).toHaveURL(url);
       }).toPass();
     } else {
-      // Keep the page-event deadline inside the 90-second slow-test budget so
-      // a missing tab names the event instead of timing out the whole test.
-      const newPagePromise = context.waitForEvent("page", {
+      await expectNewPage({
+        context,
         timeout: 60_000,
```
diegohaz, line 86: This 60-second deadline lost the comment that explained it. The removed lines were:

```ts
// Keep the page-event deadline inside the 90-second slow-test budget so
// a missing tab names the event instead of timing out the whole test.
```

That rationale is what makes the number non-obvious: 60 seconds is chosen to stay inside the `test.slow()` budget set on line 106, so a missing popup fails with "waiting for event page" instead of a bare test timeout. Nothing at the new call site records that constraint, so a future reader has no way to know the value is coupled to `test.slow()` and would look wrong if either changed.

The same number was added in `app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts:124` with no comment at all, where it depends on the identical `test.slow()` on line 107 of that file.

Suggested fix: restore the comment above the `expectNewPage` call (or attach it to `timeout` in `ExpectNewPageParams`, since it is the property that carries the constraint) and add the equivalent note in the Next.js test.

**Consumer impact**: this affects the next maintainer who touches either test's timeout, including anyone tuning these numbers after a future CI flake. Without the rationale, dropping `test.slow()` or lowering the timeout silently reintroduces the uninformative-timeout failure mode.

**Assessment**

- Severity: Low (estimate, 90% confidence), because it costs future maintenance context, not runtime behavior.
- Priority: P2 (estimate, 85% confidence), because it is cheap to restore now and expensive to reconstruct later.
- Breakage likelihood: Likely (estimate, 70% confidence), because any future reader of these two call sites hits the undocumented value.
- Fix effort: Easy (estimate, 95% confidence), because it is a two-line comment in each file.

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
```diff
@@ -110,22 +110,30 @@ withFramework(import.meta.dirname, async ({ id, query, test }) => {
     await status.click();
     await test.expect(status).toHaveAttribute("aria-expanded", "true");
     await page.mouse.move(0, 0);
-    await page.keyboard.press("d");
-    await page.keyboard.press("Enter");
+    await status.press("d");
```
diegohaz: Switching this to `status.press()` drops the implicit focus coverage this line used to provide.

`locator.press()` focuses the element before sending the key. With `page.keyboard.press("d")`, a regression that moved DOM focus off the trigger when the popup opened would send the typeahead key elsewhere and fail at `page.waitForURL(hasSearchParam("status", "draft"))` on line 115. Now Playwright silently restores focus first, so that regression passes here and only shows up later, if at all.

Every other press converted in this PR is preceded by a focus assertion, which matches the PR description's stated rule: "Send keyboard input through the control already asserted to be focused". This call site is the exception. Between `await status.click()` on line 110 and this line there is no `toBeFocused()` check; line 118 asserts focus only after the selection has already gone through.

```suggestion
    await test.expect(status).toBeFocused();
    await status.press("d");
```

**Consumer impact**: the scenario is a user opening the Status select in the Next.js App Router sandbox and typing to filter. If a future change makes `ComboboxSelect` hand DOM focus to the popover on open, keyboard typeahead breaks for that user, and this test is the coverage that should catch it.

**Assessment**

- Severity: Low (estimate, 80% confidence), because other tests in this file still assert focus on the combobox at later points.
- Priority: P2 (estimate, 80% confidence), because the coverage gap is narrow and the fix is one line.
- Breakage likelihood: Rare (estimate, 85% confidence), because it requires a future focus regression on open that no other assertion catches first.
- Fix effort: Easy (estimate, 95% confidence), because it is a single added assertion.

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
```diff
@@ -110,22 +110,30 @@ withFramework(import.meta.dirname, async ({ id, query, test }) => {
...
-        ? q.option("Archived").click({ modifiers: [modifier] })
-        : page.keyboard.press(`${modifier}+Enter`),
-    ]);
+    const newPagePromise = page.context().waitForEvent("page", {
+      timeout: 60_000,
+    });
+    if (browserName === "webkit") {
+      await q.option("Archived").click({ modifiers: [modifier] });
+    } else {
+      await page.keyboard.down(modifier);
```
diegohaz: This PR creates a second copy of the modifier-hold sequence. The same five statements now exist in `app/src/sandbox/combobox-links/test-browser.ts:131-136` (only the locator name and indentation differ), and both files are changed by this PR, so the duplication is introduced here rather than inherited.

`#app/test-utils/preview.ts` is the natural home: it already owns the shared browser-timing helper this PR consumes (`flushFrames`), and both call sites import from it. Something like:

```ts
/**
 * Presses `key` on `target` while `modifier` is held down, so the browser sees
 * a real modified key press instead of a synthesized chord. Releases the
 * modifier even when the press fails.
 */
export async function pressWithModifier(
  page: Page,
  target: Locator,
  modifier: string,
  key: string,
) {
  await page.keyboard.down(modifier);
  try {
    await target.press(key);
  } finally {
    await page.keyboard.up(modifier);
  }
}
```

While you are there, the modifier detection is already duplicated between `getNewTabModifier` on lines 4-7 of this file and `getClickModifier` on lines 41-46 of `combobox-links/test-browser.ts` — identical bodies under two names. That pair predates this PR, but it is the obvious companion for the same move, and consolidating both would leave one place to fix when platform detection needs to change.

**Consumer impact**: this affects maintainers of the open-in-new-tab coverage. `navigator.platform` is deprecated and the `finally`-release detail is easy to drop when copying; with three copies across two files, a future correction has to be found and applied three times or the tests diverge silently per browser.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the duplicated code is correct today.
- Priority: P2 (estimate, 80% confidence), because it is maintainability, not correctness, and is cheapest to fix while both files are already open.
- Breakage likelihood: Possible (estimate, 65% confidence), because platform detection and new-tab key handling are areas that do get revisited.
- Fix effort: Easy (estimate, 85% confidence), because it is one exported helper plus two call-site updates.

diegohaz (commented):
## Review summary

Initial review. Baseline: merge base `193f7392b003b7b0dbbde9fbe74bce92c44ce534` to head `682078362436415e930ad5945a9810e75f6439e9`. No prior review threads or reviews existed on this PR, so there were no unresolved candidate findings to validate.

### Scope

All three changed test files, plus the surrounding context they depend on: `app/playwright.config.ts` (project matching, retries, no `timeout` override), `app/src/test-utils/preview.ts` and `fixtures.ts` (`flushFrames`, `withFramework`, `q`), the three fixtures under test (`combobox-links/index.react.tsx`, `combobox-select-offscreen-matrix/{index,config}.react.tsx`, `nextjs/app/combobox-select-nextjs-app-router/*`), and the library mechanism the PR's stated root cause rests on (`collection-item-offscreen.tsx` and the `renderedItems` publish in `collection-store.ts`).

The root-cause analysis in the description checks out. `data-offscreen` clears when `useCollectionItemOffscreen` flips to active in a `requestIdleCallback`, the real item then registers in an effect, and the collection only publishes the sorted `renderedItems` inside `requestAnimationFrame` (`packages/ariakit-components/src/collection/collection-store.ts:216`). So the DOM can show a materialized item a frame before composite navigation can reach it, which is exactly the window the old `data-offscreen`-only wait left open.

### Findings published inline

Four, all maintainability or coverage, none blocking:

1. Undocumented `timeout: 15_000` in `waitForItemRegistration` that two calls cannot fit inside the default 30-second test budget.
2. Deleted rationale comment for `timeout: 60_000`, and the same bare number added in the Next.js test.
3. Missing `toBeFocused()` before `status.press("d")`, which converts an implicit focus check into a forced focus.
4. Second copy of the modifier-hold press sequence instead of a shared helper in `#app/test-utils/preview.ts`.

### Candidates evaluated and not published

**Replacing the `href` rewrite with `context.route()` interception — rejected, refuted by local evidence.** Rewriting the DOM under test is in tension with the "do not mock real behavior" guidance for these fixtures, so I prototyped network interception as the less invasive alternative: keep the real `href` values and stub the third-party responses instead. Standalone script with the repo's Playwright 1.62.0, a local page containing `<a href="https://ariakit.com">` and a `target="_blank"` link, and `context.route(/^https:\/\/ariakit\.com/, fulfill)`. Same-tab navigation and `target="_blank"` popups were intercepted in all three engines, but WebKit's modifier-click popup was not: the new page's main-frame document request never reached the route handler, and the handler then received roughly 40 `https://ariakit.com/_next/static/...` subresource requests, meaning the real page had been fetched from the network. Interception therefore does not cover the popup case in WebKit, which is precisely what this PR needs to stop contacting. The `prepareNavigation` approach is the right call; asserting the real `href` before rewriting it keeps the fixture's destinations covered.

**`flushFrames(page)` at the default 2 frames may under-wait — suspected, not validated.** The comment justifies two frames as the effect commit plus the publishing frame, which matches the straight-line path. But the publish rAF is cancelled and rescheduled whenever the collection's own `IntersectionObserver` fires (`collection-store.ts:236-237`), and neighbouring offscreen items materialize on staggered `requestIdleCallback` ticks, so the publish can slip past the second frame while the surrounding items are still registering. Sibling offscreen tests already needed non-default counts for related reasons (`flushFrames(page, 3)` in `combobox-select-offscreen-value/test-browser.ts:78`, `flushFrames(page, 5)` in `combobox-select-3837/test-browser.ts:82`). Confirming or refuting this needs many repeated CI-contended runs, which I could not do here, and the 200/200 local repetitions reported in the description are real evidence in the other direction. Flagging it only so the frame count is a conscious choice if `Test Chrome` flakes again in this file.

**`locator.press()` implies a scroll — checked, no impact found here.** Playwright's `focusNode` calls `node.focus()` with no `preventScroll`, and per spec the scroll step runs even when the element is already focused, so `page.keyboard.press()` to `locator.press()` is not behaviour-neutral in a viewport-sensitive test. I checked the affected targets: in `SearchableCase` the `ComboboxInput` is a sibling of the scrolling `ComboboxList`, not inside it, and the select-only targets are the triggers outside the popover, so none of them can scroll the list back after a `PageDown`. Combined with Chromium's center-if-needed alignment for focus scrolling, this looks harmless, but it is the mechanism to look at first if `toBeInViewport` assertions in this file start flaking.

**Remaining `page.keyboard.press` calls in the Next.js test — declined non-goal.** Lines 38 and 40 of `combobox-select-nextjs-app-router/test-browser.ts` press `PageDown` and `Space` with no preceding focus assertion, which is a stronger instance of the pattern this PR converts elsewhere. This PR neither introduces nor worsens it, and converting it would widen the change without evidence that test flakes.

**`newPage.close()` reordering — no problem found.** `expectNewPage` asserts the popup URL and closes it before the caller's `listbox`/`combobox` assertions, whereas the old code asserted those while the popup was still open. Closing a popup returns focus to the opener and can fire focus events the combobox reacts to, but nothing in the fixture hides the popover on refocus, and the reported 6/6 Chrome and 12/12 Firefox/Safari runs cover it. Recording it only as the ordering that changed.

**`prepareNavigation` durability under re-render — no problem found.** The `setAttribute` could in principle be undone by React, but React only writes attributes whose props changed between renders, `href` never changes, and the items cannot remount while the popover stays open with `unmountOnHide`. A remount would fail loudly on the URL assertion rather than silently pass.

No follow-up issues were registered: none of the above is a confirmed, separable piece of work the project needs to track.

### Commands run

- `git diff <merge-base>..<head>` and full reads of the three changed files plus their fixtures.
- GraphQL `reviewThreads` and `reviews` pagination: both empty.
- Standalone Playwright 1.62.0 interception experiment described above, across chromium, firefox, and webkit. No repository files were modified.
- Read `playwright-core`'s bundled `focusNode` implementation to confirm the `focus()` semantics behind `locator.press()`.

### Not validated

I did not run the Playwright suites locally: they need the Astro `preview-lite` build plus the OpenNext Next.js preview, and the perf and visual paths are CI-only. Behavioural verification rests on the runs reported in the description and on CI, where `App / Test Chrome` and `App / Test Firefox` were still pending at review time. No changeset is expected for a test-only change.

app/src/sandbox/combobox-select-offscreen-matrix/test-chrome.ts
diegohaz: Fixed in [`bf22a5615`](https://github.com/ariakit/ariakit/commit/bf22a56154abb4b0765555628a91010014968121). I removed the explicit 15-second timeout and kept `flushFrames(page)` for the frame-delayed collection publish. A stuck registration now fails through the matcher’s default 5-second deadline, and the focused Chrome matrix run passed all 89 cases.

app/src/sandbox/combobox-links/test-browser.ts
diegohaz: Fixed in [`5b9ac7385`](https://github.com/ariakit/ariakit/commit/5b9ac7385fe544bddf0a8394c5db2b7392dddd77). I restored the two-line rationale immediately above the 60-second page wait in this test and added the same explanation to the Next.js test, so both values now document their coupling to the 90-second `test.slow()` budget.

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
diegohaz: Fixed in [`5b9ac7385`](https://github.com/ariakit/ariakit/commit/5b9ac7385fe544bddf0a8394c5db2b7392dddd77). The test now asserts `status` is focused immediately before `status.press("d")`, preserving the focus coverage that `page.keyboard.press("d")` previously provided. The affected Next.js cases pass in Chrome, Firefox, and Safari.

app/src/sandbox/combobox-select-nextjs-app-router/test-browser.ts
diegohaz: Fixed in [`5b9ac7385`](https://github.com/ariakit/ariakit/commit/5b9ac7385fe544bddf0a8394c5db2b7392dddd77). I moved both `pressWithModifier` and `getNewTabModifier` into `#app/test-utils/preview.ts`, then updated both test files to use them. The press helper keeps the explicit `try`/`finally` release and uses an object parameter to follow the repository’s four-parameter style rule. All 39 affected desktop-browser cases pass.

diegohaz:
Addressed all current review feedback.

**Fixed**

- Restored the default matcher deadline for offscreen registration while keeping frame synchronization ([thread](https://github.com/ariakit/ariakit/pull/6961#discussion_r3672681506), [`bf22a5615`](https://github.com/ariakit/ariakit/commit/bf22a56154abb4b0765555628a91010014968121)).
- Documented both 60-second page-event deadlines against their 90-second slow-test budgets ([thread](https://github.com/ariakit/ariakit/pull/6961#discussion_r3672684627)).
- Restored explicit focus coverage before Next.js Status typeahead ([thread](https://github.com/ariakit/ariakit/pull/6961#discussion_r3672688113)).
- Shared modifier detection and modified key presses through the preview test utilities ([thread](https://github.com/ariakit/ariakit/pull/6961#discussion_r3672690603), [`5b9ac7385`](https://github.com/ariakit/ariakit/commit/5b9ac7385fe544bddf0a8394c5db2b7392dddd77)).

**Validation**

- `pnpm lint-fix`
- `pnpm tsc`
- Offscreen Chrome matrix: 89 passed
- Affected new-tab tests in Chrome, Firefox, and Safari: 39 passed

app/src/test-utils/preview.ts
```diff
@@ -50,6 +50,38 @@ export function flushFrames(page: Page, frames = 2) {
   );
 }
 
+/** Returns the platform modifier that opens links in a new tab. */
+export function getNewTabModifier(page: Page) {
```
diegohaz: Playwright resolves this modifier itself, so the helper does not need to exist. Passing the literal `"ControlOrMeta"` (Meta on macOS, Control on Windows and Linux) covers every consumer, which drops this function, its four call sites, and the dependency on `navigator.platform`.

I verified all three shapes the callers use with the repository's Playwright 1.62.0, across chromium, firefox, and webkit on macOS:

- `click({ modifiers: ["ControlOrMeta"] })` opened the popup in all three engines, identical to `["Meta"]`. Covers `combobox-links/test-browser.ts:90` and `combobox-select-nextjs-app-router/test-browser.ts:67` and `:128`.
- `keyboard.press("ControlOrMeta+Enter")` delivered `metaKey: true` in all three engines, identical to `"Meta+Enter"`. Covers `combobox-links/test-browser.ts:119`.
- `keyboard.down("ControlOrMeta")` set the sticky modifier state: the page observed `metaKey: true` on the following `target.press(...)`, and the held-modifier Enter opened the popup in all three engines. This is the case `pressWithModifier` depends on, and worth confirming rather than assuming, because `keyboard.down`'s own documentation only names `Shift`, `Meta`, `Control`, and `Alt` in its sticky-modifier note even though it lists `ControlOrMeta` among the supported shortcuts.

While the helper is being reshaped: `page` is derivable from `target.page()`, so dropping both `page` and `modifier` puts `pressWithModifier` under the four-parameter threshold and lets `PressWithModifierParams` go away too.

```ts
/**
 * Presses a key on a locator while the new-tab modifier is held, and releases
 * it even when the press fails. `ControlOrMeta` resolves to Meta on macOS and
 * Control elsewhere, so callers never detect the platform themselves.
 */
export async function pressWithNewTabModifier(target: Locator, key: string) {
  const page = target.page();
  await page.keyboard.down("ControlOrMeta");
  try {
    await target.press(key);
  } finally {
    await page.keyboard.up("ControlOrMeta");
  }
}
```

Both tests then drop their `const modifier = await getNewTabModifier(page);` lines and pass the literal directly.

**Consumer impact**: this affects maintainers of the open-in-new-tab coverage. `navigator.platform` is deprecated in the HTML spec, and it reports the emulated navigator rather than the host Playwright is driving. If a future WebKit build or user-agent reduction stops reporting a `Mac` prefix, the macOS Safari job would silently press Control and these tests would fail as an unrelated-looking new-tab timeout, with nothing in the failure pointing at platform detection. `ControlOrMeta` is resolved by Playwright from the host, so it cannot drift with the page's navigator.

**Assessment**

- Severity: Low (estimate, 85% confidence), because behaviour is equivalent today and the gain is fewer moving parts plus no deprecated API.
- Priority: P2 (estimate, 80% confidence), because this PR is the moment the helper becomes a shared exported utility, so not adding it is cheaper than removing it later.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because `navigator.platform` still reports `MacIntel` in current WebKit builds; the exposure is future drift, not a present failure.
- Fix effort: Easy (estimate, 90% confidence), because it deletes one helper, simplifies another, and updates five call sites.

diegohaz (commented):
## Re-review summary

Re-review. Baseline taken from the previous marked review's `head_sha=682078362436415e930ad5945a9810e75f6439e9`; incremental diff to the current head `5b9ac7385fe544bddf0a8394c5db2b7392dddd77`, covering `bf22a5615` and `5b9ac7385`. Findings first discovered in this run were reviewed from those two commits plus the surrounding code they newly expose, and the four unresolved candidates from the previous review were independently validated against the current head rather than taken from the replies.

### Prior findings: all four confirmed fixed at the current head

Each was re-read in the current files, not accepted from its reply.

- **15-second offscreen deadline** (`combobox-select-offscreen-matrix/test-chrome.ts:26`, thread `PRRT_kwDOBjT0Fc6UrwB6`): resolved. The explicit timeout is gone and the assertion is back on the matcher default, which is the same deadline the pre-PR code used for this assertion, so nothing new is exposed. `flushFrames(page)` and its rationale comment remain, which is the part that addresses the described race.
- **Deleted 60-second rationale** (`combobox-links/test-browser.ts:127`, thread `PRRT_kwDOBjT0Fc6UrwDT`): resolved. The two-line comment is restored verbatim above the `expectNewPage` call at lines 123-124, and the equivalent comment now sits above the Next.js `waitForEvent` at lines 122-123. Both sit above the code they describe and both files still carry the `test.slow()` the value is coupled to.
- **Missing focus assertion** (`combobox-select-nextjs-app-router/test-browser.ts:112`, thread `PRRT_kwDOBjT0Fc6UrwEx`): resolved. `await test.expect(status).toBeFocused()` now sits at line 111, between the mouse move and `status.press("d")`, so the press documents focus instead of establishing it.
- **Duplicated modifier-hold sequence** (thread `PRRT_kwDOBjT0Fc6UrwGh`): resolved. `getNewTabModifier` and `pressWithModifier` moved into `#app/test-utils/preview.ts` alongside `flushFrames`, the local copies in both tests are gone, and `pressWithModifier` uses an object parameter with a `Params` interface as the four-parameter rule requires.

One, on the newly shared helper: Playwright's own `"ControlOrMeta"` modifier makes `getNewTabModifier` unnecessary, which also removes the dependency on the deprecated `navigator.platform`. Verified against all three consumer shapes in chromium, firefox, and webkit before publishing; details and evidence are on the thread.

### Candidate evaluated and not published

**`pressWithModifier` placement in `preview.ts` — declined non-goal.** The module's name suggests preview navigation rather than general Playwright helpers, but it already exports `flushFrames` as a generic browser-timing utility, so the new helpers follow the established precedent rather than breaking it. Splitting out a separate helpers module would be a repository-wide move that this PR should not carry.

**PR description does not cover the shared helpers — worth a small edit, not a finding.** The Solution section still describes only the three test files; `app/src/test-utils/preview.ts` is now a fourth changed file exporting two new test utilities, and the Testing section's run counts predate `bf22a5615` and `5b9ac7385`. The offscreen snippet was already refreshed to drop the 15-second timeout, so this is just the remaining gap.

### Carried forward from the previous review, still open and still unvalidated

**`flushFrames(page)` at the default two frames may under-wait.** Unchanged by these commits. The publish rAF is cancelled and rescheduled whenever the collection's own `IntersectionObserver` fires (`packages/ariakit-components/src/collection/collection-store.ts:236-237`), and neighbouring offscreen items materialize on staggered `requestIdleCallback` ticks, so the publish can slip past the second frame. Still not something I can confirm or refute without many CI-contended runs, and the reported local repetitions plus the 89 passing matrix cases are evidence in the other direction. Recording it again only so the frame count stays a conscious choice if this file flakes again.

### Commands run

- `git diff 682078362..5b9ac7385` plus full reads of the changed regions in all four files.
- GraphQL `reviewThreads` pagination to exhaustion: four threads, all resolved, each read in full including replies.
- Standalone Playwright 1.62.0 experiments across chromium, firefox, and webkit for the `ControlOrMeta` finding: popup opening via `click({ modifiers })`, the `"ControlOrMeta+Enter"` chord through `keyboard.press`, and sticky modifier state through `keyboard.down` observed as `metaKey` on the following key event. No repository files were modified.
- Confirmed `Locator.page()` exists in the pinned Playwright types before proposing it.

### Not validated

The Playwright suites were not run locally; they need the Astro `preview-lite` build plus the OpenNext Next.js preview. At review time CI for this head had `Main / Type check`, `Main / Lint`, `App / Test Firefox`, and `App / Test Safari` passing, with `App / Test Chrome` still pending. The type check passing is what confirms `getNewTabModifier`'s return type still narrows to the literal union the `modifiers` array requires.

app/src/test-utils/preview.ts
diegohaz: Fixed in [`614de7100`](https://github.com/ariakit/ariakit/commit/614de710058fd942bee2fc74137021f27f8354e9): replaced page-level platform detection with Playwright’s `ControlOrMeta` across the click and keyboard flows, and simplified the press helper to derive its page from the locator.

diegohaz:
**Fixed**

- [`614de7100`](https://github.com/ariakit/ariakit/commit/614de710058fd942bee2fc74137021f27f8354e9) ([reply](https://github.com/ariakit/ariakit/pull/6961#discussion_r3673383889)): Replaced deprecated page-level platform detection with Playwright’s `ControlOrMeta`, simplified the held-modifier helper, and verified all 39 affected tests across Chrome, Firefox, and Safari.

app/src/test-utils/preview.ts
```diff
@@ -50,6 +50,44 @@ export function flushFrames(page: Page, frames = 2) {
...
+  modifier: string;
+  page: Page;
+}
+
+/**
+ * Runs an action while a modifier is held, and releases the modifier even when
+ * the action fails. Unlike per-action pointer modifiers, this can keep the key
+ * down until related browser work, such as opening a new page, has settled.
+ */
+export async function withModifier<Result>({
```
diegohaz: This helper adds a mechanism the PR has not shown it needs, and it is the third layer of machinery added to this direction.

At `5b9ac7385` both click sites used `click({ modifiers: [modifier] })`, and that form is the one the PR reports as passing: the Testing section still lists "6/6 link tests in Chrome and 12/12 across Firefox and Safari with no retries", and the reply on the shared-helpers thread reported "All 39 affected desktop-browser cases pass". `e52011b82` then replaced both with a held-modifier wrapper. Neither that commit message nor the description records a failure that motivated it; the justification exists only as the assertion in this JSDoc that a pointer modifier may be released before "related browser work, such as opening a new page, has settled".

I could not reproduce a difference at the browser level. With the repository's Playwright 1.62.0 on a plain anchor, 15 iterations per strategy per engine, in chromium, firefox and webkit:

| strategy | popups |
| --- | --- |
| `press("ControlOrMeta+Enter")` | 15/15 in all three |
| `keyboard.down` + `press` + `keyboard.up` | 15/15 in all three |
| `click({ modifiers: ["ControlOrMeta"] })` | 15/15 in all three |
| `keyboard.down` + `click()` + `keyboard.up` | 15/15 in all three |

That matches what Playwright documents for `modifiers`: "Ensures that only these modifiers are pressed during the operation, and then restores current modifiers back" — the operation-scoped guarantee this wrapper is meant to add. The disposition also travels in the click event's own modifier flags rather than live keyboard state, so a release after dispatch cannot change it. This is browser-level evidence on a plain anchor, not a run of the Ariakit fixtures, so it does not rule out a fixture-specific reason; it does mean the reason has to be named rather than assumed.

The cost is concrete:

- A new exported generic `withModifier<Result>` plus `WithModifierParams<Result>`, whose `modifier` is `"ControlOrMeta"` at all three call sites.
- `combobox-select-nextjs-app-router/test-browser.ts:64-73` turned a flat `Promise.all` into a nested closure that re-declares `newPage` inside the callback, shadowing the `const newPage` the statement is initialising, purely to hand the popup back out.
- `combobox-links/test-browser.ts:86-93` now holds the modifier across the whole of `expectNewPage`, which includes `expect(newPage).toHaveURL(url)` and `newPage.close()`. Neither has anything to do with the click, so the key stays down for the popup's entire load and teardown.

The same file also still trusts the chord form in the WebKit branch at `combobox-links/test-browser.ts:122` (`page.keyboard.press("ControlOrMeta+Enter")`), so the codebase is internally inconsistent about whether a modifier released with the key press is reliable.

Either name the failure this fixes, ideally in the description alongside the run that reproduced it, or revert the two click sites to `click({ modifiers: ["ControlOrMeta"] })` and drop `withModifier`. If the chord result above also holds for the fixtures, `pressWithNewTabModifier` collapses into `target.press("ControlOrMeta+Enter")` and `preview.ts` needs no new exports at all.

**Consumer impact**: this affects maintainers of these tests. A held-modifier wrapper is load-bearing-looking code that nothing documents a reason for, so a future reader cannot tell whether flattening it back to `click({ modifiers })` is safe, and the Next.js call site now costs a nested closure and a shadowed binding to express what one line expressed before.

**Assessment**

- Severity: Low (estimate, 80% confidence), because the tests behave correctly either way; the cost is unexplained machinery and reduced readability.
- Priority: P1 (estimate, 70% confidence), because a de-flaking PR should not land a mechanism whose need is unstated, and the reassessment cadence below makes this the right moment to settle it.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because holding the modifier longer is not itself incorrect; the risk is that the wrapper hides which property the tests actually depend on.
- Fix effort: Easy (estimate, 85% confidence), because it is a revert of two call sites plus deleting one export, or one paragraph in the description.

diegohaz (commented):
## Re-review summary

Re-review. Baseline taken from the previous marked review's `head_sha=5b9ac7385fe544bddf0a8394c5db2b7392dddd77`; incremental diff to the current head `614de710058fd942bee2fc74137021f27f8354e9`, covering `2ba1df8dd` (merge of main), `e52011b82`, and `614de7100`. The merge brought in `1c33b7d66` and touches `vitest.*`, `.github/workflows/*`, `package.json`, and `packages/ariakit-scripts/src/perf.test.ts`; against the new merge base this PR still owns only four files, so those were treated as inherited rather than reviewed as PR changes. Findings first discovered in this run come from the two PR commits plus the surrounding code they newly expose, and the previous review's finding was independently validated against the current head rather than taken from its reply.

### Prior finding: confirmed fixed at the current head

**`ControlOrMeta` replacing `getNewTabModifier`** (thread `PRRT_kwDOBjT0Fc6UsrTy`): resolved, and adopted in full. `getNewTabModifier` is gone, `navigator.platform` no longer appears anywhere in these tests, the four `page.evaluate` round-trips are removed, and `pressWithNewTabModifier(target, key)` now derives its page from `target.page()`, so `PressWithModifierParams` is gone too. All five former consumers use the literal: `click({ modifiers: ["ControlOrMeta"] })` in the Next.js WebKit branch, `page.keyboard.press("ControlOrMeta+Enter")` in the `combobox-links` WebKit branch, and the helper elsewhere.

One, on `withModifier` in `app/src/test-utils/preview.ts`: `e52011b82` replaced two working `click({ modifiers })` call sites with a held-modifier wrapper, without recording the failure that motivated it. I measured all four modifier strategies at the browser level across chromium, firefox, and webkit and found no difference; details, the measurement table, and the concrete costs are on the thread.

### Implementation review reassessment

This is the third finding-bearing review cycle on one materially unchanged direction (cycle 1: four findings; cycle 2: one finding; cycle 3: this one), which hits the cadence in `ariakit-general-review-reassessment`. Note one deviation from that skill: it requires spawning a fresh read-only subagent with parent context disabled to assess complexity in parallel, and this session is under a standing instruction not to use subagents, so the assessment below was performed only in the main task. Treat it as one assessment, not the two-source comparison the skill asks for.

**Assessment.** The original task is narrow: stop `Test Chrome` from retrying on unrelated PRs. Issue severity is low (flaky CI, no user-facing behaviour), expected frequency was intermittent, supported scope is three test files, and user impact is zero — this ships no library change and needs no changeset. Against that, the direction has accumulated `prepareNavigation`, `expectNewPage`, `waitForItemRegistration`, `withModifier`, and `pressWithNewTabModifier`, of which the last two are newly exported utilities in a shared module. The repeated cycles have clustered on exactly one thing: undocumented or unproven mechanism (a bare 15-second timeout, a deleted rationale comment, platform detection that Playwright already does, and now a held-modifier wrapper). Each individual fix has been correct and has genuinely reduced the code — cycle 2 deleted a helper outright.

**Decision: continue the direction, but stop adding mechanism.** The core of the PR is well founded: the offscreen race is real and traced to the rAF publish in `collection-store.ts:216`, and removing third-party navigation from the link tests is justified by evidence I gathered in cycle 1. Those parts should land. The one piece that should be settled before merge is `withModifier`, which is the only accumulated machinery whose need is not demonstrated, and it arrived after the form it replaced had been reported passing. Reverting it also removes the nested closure and shadowed `newPage` binding at the Next.js call site. No follow-up issues registered; nothing here is separable work the project needs to track.

Per that skill, the PR description needs an `## Implementation review reassessment` section carrying this cycle's entry as a collapsed `<details>` block before the next push completes.

### Description drift, worth fixing in the same pass

The Solution section still describes only the three test files. `app/src/test-utils/preview.ts` is now a fourth changed file exporting two new test utilities, and the Testing counts still predate `bf22a5615`, `5b9ac7385`, `e52011b82`, and `614de7100` — the "6/6 link tests in Chrome and 12/12 across Firefox and Safari" line specifically describes the `click({ modifiers })` form that `e52011b82` replaced, which is what makes the new wrapper look already-verified when it is not. The offscreen snippet was refreshed correctly in an earlier cycle.

### Carried forward, still open and still unvalidated

**`flushFrames(page)` at the default two frames may under-wait.** Unchanged by these commits, and unchanged in reasoning: the publish rAF is cancelled and rescheduled whenever the collection's own `IntersectionObserver` fires (`packages/ariakit-components/src/collection/collection-store.ts:236-237`). Still not confirmable without many CI-contended runs.

### Commands run

- `git diff 5b9ac7385..614de7100 -- app/` for the PR's own changes, plus `git diff --stat` against the new merge base `1c33b7d66` to separate inherited files, plus `git show e52011b82` and `git show 5b9ac7385:<path>` to establish what the wrapper replaced.
- GraphQL `reviewThreads` pagination to exhaustion: five threads, all resolved, read in full.
- Standalone Playwright 1.62.0 experiment across chromium, firefox, and webkit: 15 iterations each of chord press, `keyboard.down`/`press`/`keyboard.up`, `click({ modifiers })`, and held-modifier `click()`, counting popup creation. All 12 combinations were 15/15. No repository files were modified.

### Not validated

The Playwright suites were not run locally; they need the Astro `preview-lite` build plus the OpenNext Next.js preview. CI for this head had `Main / Lint` passing, with `Main / Type check`, `App / Test Chrome`, `App / Test Firefox`, and `App / Test Safari` all still pending at review time, so nothing about this head is confirmed green yet. The browser-level modifier measurement was run against a plain anchor, not the Ariakit fixtures, so it constrains the mechanism rather than the fixtures.

app/src/test-utils/preview.ts
diegohaz: Reassessed in [`cbc8b4460`](https://github.com/ariakit/ariakit/commit/cbc8b4460): broader CI evidence contradicted the popup experiment, so I removed that work entirely instead of keeping the simplification as the apparent final fix. The links and Next.js popup fixtures now match `main` exactly, no generic modifier helper remains in the cumulative diff, and the PR is scoped only to the offscreen matrix registration checkpoint.

diegohaz:
**Final scope**

- [`cbc8b4460`](https://github.com/ariakit/ariakit/commit/cbc8b4460) ([popup reassessment](https://github.com/ariakit/ariakit/pull/6961#discussion_r3673716868), [keyboard reassessment](https://github.com/ariakit/ariakit/pull/6961#discussion_r3674201168)): Removed the popup and Safari typeahead experiments after broader CI contradicted them. Those fixtures now match `main` exactly. Kept only the offscreen matrix registration checkpoint using the shared `flushFrames` helper; it passed 89/89 focused cases and 890/890 stressed cases with four workers and no retries.
- [`af052b9a3`](https://github.com/ariakit/ariakit/commit/af052b9a3): Merged the latest `main` into the branch.

**Final workflow**

- Chrome: 741 passed with no retries.
- Firefox: 512 passed with no retries.
- Safari: 443 passed with one retry in unchanged `popover-anchor-3729`; all 16 visual tests passed.
- The workflow gate and all required checks passed.

app/src/sandbox/combobox-links/test-browser.ts
```diff
@@ -67,34 +100,30 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
       }).toPass();
     } else {
       // Keep the page-event deadline inside the 90-second slow-test budget so
       // a missing tab names the event instead of timing out the whole test.
-      const newPagePromise = context.waitForEvent("page", {
+      await expectNewPage({
+        context,
         timeout: 60_000,
+        url,
+        action: () => combobox.press("ControlOrMeta+Enter"),
```
diegohaz: This one line reverts a stabilization that `main` already carries, and the evidence used to justify the revert does not cover it.

At the merge base `1c33b7d66` this branch read:

```ts
await page.keyboard.down(modifier);
try {
  await combobox.press("Enter");
} finally {
  await page.keyboard.up(modifier);
}
```

That shape came from #6913 (`c483d81de`), merged into `main` one day before this PR, whose Solution bullet is "Dispatch modifier+Enter with an explicit modifier-key lifetime and allow slow new tabs to finish hydrating" and whose Testing section records "Firefox new-tab tests: 20/20 sequential; modifier+Enter 30/30 parallel with no retries". So the explicit lifetime was a named, separately evidenced fix for this exact test, not incidental style.

My previous comment was scoped to the two click sites, where `withModifier` had been layered on top of a `click({ modifiers })` form the PR already reported passing. I conditioned the keyboard collapse on the chord result also holding for the fixtures, and that condition has not been demonstrated. The failures this PR documents in [job 90532896840](https://github.com/ariakit/ariakit/actions/runs/30438667490/job/90532896840?pr=6961) are described in the Motivation as occurring "in both modified-click cases" — the click paths. Nothing in the PR reports the keyboard path failing, so the reassessment's conclusion is being carried from click evidence onto a path the click evidence never touched.

The two forms also are not equivalent in the property #6913 named. Measuring the modifier keydown-to-keyup interval in the page, 10 iterations per form with the repository's Playwright 1.62.0:

| engine | `press("ControlOrMeta+Enter")` | `down` + `press` + `up` |
| --- | --- | --- |
| chromium | median 2.7ms (min 2.6, max 6.9) | median 8.3ms (min 5.4, max 11.1) |
| webkit | median 8.0ms (min 4.0, max 11.0) | median 13.0ms (min 9.0, max 17.0) |
| firefox | median 29.0ms (min 26.0, max 31.0) | median 31.0ms (min 30.0, max 32.0) |

The explicit form holds the modifier roughly two to three times longer in Chromium and WebKit, because each additional `await` is another protocol round-trip. That is measurement on an idle machine; the inference beyond it is that those round-trips get slower on a contended runner, so the gap widens exactly when contention is highest, which is the condition #6913 and this PR are both targeting.

Worth separating from this: the Next.js site is not affected. At the merge base it was a page-level `page.keyboard.press(\`${modifier}+Enter\`)`, so `status.press("ControlOrMeta+Enter")` there is strictly better than what `main` has. This concerns only `combobox-links`.

Either keep the explicit hold for this one keyboard site while the click sites stay flattened, or re-run #6913's modifier+Enter stress at this head and record it in Testing, so the removal rests on the same class of evidence that introduced it. The green Chrome job on this head is one sequential run, not the 30/30 parallel evidence #6913 used.

**Consumer impact**: maintainers watching `Test Chrome`. If Chrome starts retrying on unrelated PRs again after this merges, the visible cause will look like the offscreen registration work, while the actual change would be a keyboard-timing stabilization silently reverted from a different PR — precisely the attribution problem this PR exists to end.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because it risks reintroducing the flake class in the job this PR exists to stabilize.
- Priority: P1 (estimate, 75% confidence), because reverting another PR's evidenced fix should be settled before merge, not after.
- Breakage likelihood: Possible (estimate, 60% confidence), because it needs a contended runner; the millisecond gap is real but its effect on popup materialization is not established in either direction.
- Fix effort: Easy (estimate, 90% confidence), because it is either restoring five lines at one call site or running an existing stress command and recording the result.

diegohaz (commented):
## Re-review summary

Re-review. Baseline taken from the previous marked review's `head_sha=614de710058fd942bee2fc74137021f27f8354e9`; incremental diff to the current head `ee4a11b14c64035945be42aa0f131eec6e0ace89`, covering `ee4a11b14`. Findings first discovered in this run come from that commit plus the surrounding history it touches, and the previous review's finding was independently validated against the current head rather than taken from its reply.

This is the fourth finding-bearing cycle, so it is not a reassessment boundary under `ariakit-general-review-reassessment`; the cycle-3 entry in the description satisfies the cadence, and its format matches what that skill asks for.

### Prior finding: confirmed fixed, and then some

**`withModifier` machinery** (thread `PRRT_kwDOBjT0Fc6UuI_7`): resolved, and resolved more thoroughly than requested. Both exports are gone, `app/src/test-utils/preview.ts` is byte-identical to the merge base again, so this PR now owns three files instead of four. The Next.js call site is back to a flat `Promise.all` with no shadowed binding, and the `combobox-links` click site is back to `click({ modifiers: ["ControlOrMeta"] })` inside `expectNewPage` rather than wrapping it.

The description work is the stronger half of this commit. The Motivation now names [job 90532896840](https://github.com/ariakit/ariakit/actions/runs/30438667490/job/90532896840?pr=6961) as continuing popup failure after local URLs removed third-party latency, and states plainly that this does not demonstrate the held-modifier hypothesis. The stale "Register the Next.js popup event before the action" bullet is gone, `app/src/test-utils/preview.ts` correctly no longer appears since it is no longer changed, and the reassessment entry is a collapsed `<details>` with the cycle count in its summary.

One, at `combobox-links/test-browser.ts:123`: the same commit also flattened the keyboard path, and that path is not covered by the evidence used to justify the flattening. At the merge base it carried #6913's explicit modifier-key lifetime, which that PR named as a solution bullet and backed with 30/30 parallel modifier+Enter runs. Measurements of the actual modifier lifetime for both forms are on the thread.

### Correction to my cycle-3 comment

I cited the WebKit branch at `combobox-links/test-browser.ts:113` still using the chord form as evidence that the codebase was "internally inconsistent about whether a modifier released with the key press is reliable". That was weaker evidence than I presented it as: the WebKit branch does not create a popup at all, since Safari navigates in place there, so it says nothing about whether a chord is reliable for popup creation. The rest of that comment's evidence stands, and the click-site conclusion it drove was correct.

### Candidate not published inline: the residual popup flake needs a durable record

The reassessment states that "Browser popup creation under heavily contended Chrome CI remains outside this PR's supported fix until fixture-specific evidence identifies a deterministic mechanism". That is the right call for this PR's scope, and the failure behind it is confirmed rather than suspected: job 90532896840 shows one hard failure and one retry in the modified-click cases with local URLs already in place.

Under Finding Dispositions in `ariakit-general-workflow` that is a Registered follow-up, and registration is complete only when the disposition names a GitHub issue URL — a pull request note is explicitly not the durable record. I searched the open issues in this repository and found no existing match, so one would need creating. I have not created it: this review is comment-only and I do not open issues on the author's behalf without being asked. The content it should carry is already written in the Motivation and the reassessment block, plus the job link and the two affected test names.

Half of this PR's stated motivation is new-tab navigation flakiness, and that half is now explicitly out of scope. Without an issue, the next person who sees `Test Chrome` retry has nothing linking it to the analysis already done here.

### Carried forward, still open and still unvalidated

**`flushFrames(page)` at the default two frames may under-wait.** Unchanged by this commit and unchanged in reasoning: the publish rAF is cancelled and rescheduled whenever the collection's own `IntersectionObserver` fires (`packages/ariakit-components/src/collection/collection-store.ts:236-237`). Still not confirmable without many CI-contended runs.

### Commands run

- `git diff 614de7100..ee4a11b14`, plus `git diff --stat` against the merge base `1c33b7d66` to confirm `preview.ts` is no longer part of this PR.
- `git log -S 'keyboard.down(modifier)'` to find the origin of the reverted hold, then `git show c483d81de` and the #6913 description to establish what it was for and what evidence backed it.
- `git show 1c33b7d66:<path>` for both test files to establish each keyboard path's pre-PR state, which is what separates the `combobox-links` regression from the Next.js improvement.
- GraphQL `reviewThreads` pagination to exhaustion: six threads, all resolved, read in full.
- Standalone Playwright 1.62.0 measurement across chromium, firefox, and webkit: modifier keydown-to-keyup interval, 10 iterations per form. No repository files were modified.
- Searched open issues in this repository for an existing popup-flake follow-up; none found.

### Not validated

The Playwright suites were not run locally. CI for this head has `Main / Lint`, `Main / Type check`, `App / Test Chrome`, and `App / Test Firefox` passing, with `App / Test Safari` still pending. The green Chrome job is a single sequential run and does not substitute for the parallel stress evidence #6913 used for the keyboard path. The lifetime measurement was taken on an idle machine against a plain anchor, so it establishes that the two forms differ in modifier lifetime, not what that difference does to popup materialization under CI contention.

app/src/sandbox/combobox-links/test-browser.ts
diegohaz: Reassessed in [`cbc8b4460`](https://github.com/ariakit/ariakit/commit/cbc8b4460): restored `combobox-links` exactly to `main`, which preserves #6913's explicit modifier lifetime, and dropped the rest of the unproven popup experiment. The cumulative PR is now matrix-only.

PR #6965, opened 2026-07-29
Run framework-neutral tests in Node

diegohaz:
## Motivation

The framework-neutral `core` Vitest project runs every generic test in happy-dom, even though many of those tests only need Node. The source plugin is also inherited by every project although only the source-plugin integration test imports `?source` modules.

The framework ownership exclusions additionally repeat React and Solid globs, making it easier for their positive and negative project boundaries to drift apart.

## Solution

Make the project names describe their actual environments:

```ts
projects: [
  { test: { name: "node", environment: "node" } },
  { test: { name: "dom", environment: "happy-dom" } },
  // react and solid
];
```

The `node` project owns generic tests by default. The `dom` project owns explicit `*.dom.test.ts` files and tests in non-framework, non-scripts `ariakit-*` packages. React and Solid keep their framework-specific ownership, while explicit DOM markers take precedence even inside framework or scripts packages.

Project includes and exclusions now come from a shared `projectTestClaims` abstraction backed by `getFrameworkTestIncludes`, so the positive framework/example claims also define the other projects' boundaries. Empirical collection verifies that every current test file is claimed without hard-coding the inventory into the configuration.

The StackBlitz suite uses the `.dom.test.ts` convention. Imported Select legacy test support modules use the existing `.test-helper.ts` convention so they remain type-checked as test code without being collected as standalone entries. `sourcePlugin` is registered only on the Node project that contains its integration test.

The root `pnpm test` command continues to run all Vitest projects together.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test --run` (1,678 tests; 250 project assignments across 248 unique files)
- Test collection: Node 32, DOM 29, React 183, Solid 6; no missing or extra files


vitest.config.ts
```diff
@@ -7,23 +7,18 @@ import { sourcePlugin } from "./app/src/lib/source-plugin.ts";
...
-  // also match the generic *test.* include above, so keep them out of the core
-  // project explicitly.
-  "**/*react.test.{ts,tsx}",
-  "**/*solid.test.{ts,tsx}",
-  "app/src/{examples,sandbox}/**/test.{ts,tsx}",
+const domTestIncludes = [
+  "app/src/lib/stackblitz.test.{ts,tsx}",
+  "packages/ariakit-components/src/{collection,combobox,form,popover,select,tab}/**/*test.{ts,tsx}",
+  "packages/ariakit-test/src/**/*test.{ts,tsx}",
+  "packages/ariakit-utils/src/{dom,events,focus}.test.{ts,tsx}",
```
diegohaz: We could make this more generic. All `ariakit-*` packages, along with any `*.dom.test.ts` files, can run in the `dom` project (rename the StackBlitz test file). We'll naturally exclude paths covered by other projects, such as `ariakit-${framework}` packages and `ariakit-scripts`.

vitest.config.ts
```diff
@@ -44,13 +39,21 @@ function getFrameworkTestIncludes(framework: Framework) {
...
+const frameworkTestExcludes = frameworks.flatMap((framework) => [
+  `packages/ariakit-${framework}*/src/**/*test.{ts,tsx}`,
+  `**/*${framework}.test.{ts,tsx}`,
+]);
 
-const sourcePluginInstance = sourcePlugin(join(rootDir, "app/src/examples/"));
+const nodeTestExcludes = [
+  ...domTestIncludes,
+  ...frameworkTestExcludes,
+  "app/src/{examples,sandbox}/**/test.{ts,tsx}",
```
diegohaz: Are you sure this can't be inferred from `getFrameworkTestIncludes` (`...exampleTests`)? Have you tested this empirically?

I still feel like we're hard-coding too many paths here. It's becoming clear that files from one project should be excluded from the others. `react` and `solid` would be exceptions because they can run the same test files. It seems like we could use some smarter logic here instead of repeating all these paths.

Please think more deeply about this and come up with a better abstraction.

vitest.config.ts
diegohaz: Fixed in [`827a5876b`](https://github.com/ariakit/ariakit/commit/827a5876b37b7b18c09ae58fb1614c0ef69dc721): the DOM claim now covers non-framework, non-scripts `ariakit-*` packages plus explicit `*.dom.test.*` files, and the StackBlitz suite is now `stackblitz.dom.test.ts`. Explicit DOM markers also take precedence inside scripts and framework packages.

vitest.config.ts
diegohaz: Fixed in [`827a5876b`](https://github.com/ariakit/ariakit/commit/827a5876b37b7b18c09ae58fb1614c0ef69dc721): `projectTestClaims` now centralizes the DOM, React, and Solid includes, with React/Solid claims coming from `getFrameworkTestIncludes` (including `exampleTests`). Node and DOM exclusions are derived from those claims, while explicit DOM markers have defined precedence. Empirically, all 248 discovered test files are claimed (Node 32, DOM 29, React 183, Solid 6), with only the two intentional React/Solid overlaps. The imported Select modules use `.test-helper.ts`, so no broad framework exclusion is needed for them.

diegohaz:
**Fixed**

- [`827a5876b`](https://github.com/ariakit/ariakit/commit/827a5876b37b7b18c09ae58fb1614c0ef69dc721) ([reply](https://github.com/ariakit/ariakit/pull/6965#discussion_r3674360639)): Made DOM ownership convention-driven for non-framework `ariakit-*` packages and explicit `.dom.test` files, including the StackBlitz rename and explicit-marker precedence.
- [`827a5876b`](https://github.com/ariakit/ariakit/commit/827a5876b37b7b18c09ae58fb1614c0ef69dc721) ([reply](https://github.com/ariakit/ariakit/pull/6965#discussion_r3674364389)): Derived project boundaries from shared claims and empirically verified the complete 248-file corpus.
- [`b2c1e5664`](https://github.com/ariakit/ariakit/commit/b2c1e5664e4f6ebebb864a34a11d86d7fcdb7915) ([reply](https://github.com/ariakit/ariakit/pull/6965#discussion_r3674435186)): Removed the Vitest config test and its test-only export as requested.

**No longer applicable**

- [`b2c1e5664`](https://github.com/ariakit/ariakit/commit/b2c1e5664e4f6ebebb864a34a11d86d7fcdb7915) ([reply](https://github.com/ariakit/ariakit/pull/6965#discussion_r3674438707)): The Windows path-separator suggestion targeted the removed config test.

vitest.config.test.ts
```diff
@@ -0,0 +1,77 @@
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
```
diegohaz: We don't need a test file for the vitest config lol

diegohaz: Fixed in [`b2c1e5664`](https://github.com/ariakit/ariakit/commit/b2c1e5664e4f6ebebb864a34a11d86d7fcdb7915): removed `vitest.config.test.ts` and made `getProjectTestPatterns` private again. The empirical collection audit still accounts for all 248 test files.

vitest.config.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  "assigns $file to $projects",
+  ({ file, projects: expectedProjects }) => {
+    const actualProjects = projects.filter((project) => {
+      const { exclude, include } = getProjectTestPatterns(project);
+      const files = globSync(include, {
+        cwd: fixturePath,
+        ignore: exclude,
+        nodir: true,
+      });
+      return files.includes(file);
```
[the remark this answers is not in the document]
diegohaz: No longer applicable after [`b2c1e5664`](https://github.com/ariakit/ariakit/commit/b2c1e5664e4f6ebebb864a34a11d86d7fcdb7915): `vitest.config.test.ts` was removed, so the platform-specific fixture comparison no longer exists.

PR #6973, opened 2026-07-30
Run visual regression tests independently

diegohaz:
## Motivation

Visual regression tests currently run after non-visual browser tests within the same App workflow matrix shards. This serializes independently runnable suites and means a browser-test failure skips visual coverage for that shard.

## Solution

Move visual regression coverage into dedicated matrix entries:

```yaml
- name: Visual
  os: ubuntu-latest
  engines: chromium firefox
  projects: chrome firefox
- name: Visual (Safari)
  os: macos-latest
  engines: webkit
  projects: safari
```

Both visual jobs depend only on the app and Next.js builds, so they start concurrently with the browser matrix. The screenshot finalizer waits for both the browser and visual matrices before write-back, preventing its generated commit from triggering `cancel-in-progress` while browser jobs are active. Its screenshot update decision still depends on the visual matrix result.

The local Playwright installer accepts multiple space-separated engines and normalizes their cache-key segment with dashes, allowing the Ubuntu visual job to install Chromium and Firefox in one action call.

Each visual runner packs one WebP archive per browser project and uploads them together under an OS-specific artifact, such as `app-visual-ubuntu-latest`. Upload only occurs after every configured project packs successfully, and the finalizer derives the expected OS artifacts from the upload step names before replacing committed baselines. Failed Playwright output uses the separate `app-test-results-*` namespace.

## Testing

- `pnpm test packages/ariakit-scripts/src/ci-visual.test.ts`
- `pnpm lint-fix`
- `pnpm tsc`
- YAML parsing and formatting checks
- `actionlint` v1.7.12

<details>
<summary>Cycle 3: Keep concurrent visual execution and coordinate screenshot write-back</summary>

The core direction remains sound. On the base branch, visual tests run serially after non-visual tests in each browser shard, so an ordinary browser failure suppresses visual coverage. The split adds the requested two visual jobs and lets them start concurrently with the browser jobs after the shared builds.

The cycle 3 finding is valid and moderate severity. When a baseline mismatch finishes before browser tests, an early automated push can trigger the repository's `cancel-in-progress` policy and cancel the older run. This is infrequent across all runs but plausible on baseline-changing pull requests. Its impact is wasted runner time and delayed browser feedback, not baseline corruption.

The correction restores only the finalizer relationship:

```yaml
visual-finalize:
  needs: [test, visual]
  if: always() && needs.test.result != 'cancelled' && needs.visual.result != 'cancelled'
```

The visual and browser matrices remain independent because both still depend only on the shared builds. Ordinary browser failures do not suppress visual execution or screenshot write-back; cancellation prevents the finalizer from starting a replacement run while work is still active.

This is a low-complexity correction with no new helper, state, or workflow. The earlier portability and configuration simplifications remain justified, and the atomic grouped-artifact guard remains the smallest protection against partial screenshot replacement. Fork write-back and recovery from partial artifacts remain outside this pull request's scope.

Decision: keep the split and previous dispositions, restore only the finalizer dependency and cancellation guard, and narrow the claim from full pipeline independence to independent visual execution with coordinated write-back.

</details>


.github/workflows/app.yml
```diff
@@ -137,11 +132,87 @@ jobs:
...
+      - name: Set up environment
+        uses: ./.github/workflows/setup
+        with:
+          filters: app...
+
+      - name: Install Playwright
+        uses: ./.github/workflows/install-playwright
+        with:
+          engine: ${{ matrix.engine }}
+          workspace: app
```
diegohaz: We should probably allow for multiple engines in the `install-playwright` action so we don't need to call it twice and can reduce the number of matrix fields as well.

.github/workflows/app.yml
```diff
@@ -137,11 +132,87 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            name: Visual
+            engine: chromium
+            secondary_engine: firefox
+            project: chrome
+            secondary_project: firefox
```
diegohaz: We shouldn't need these secondary fields. We can use `engines` and `projects` and accept many.

.github/workflows/app.yml
```diff
@@ -137,11 +132,87 @@ jobs:
...
+            projects: chrome firefox
+            test: xvfb-run pnpm -F app run test-visual
+            results: visual
```
diegohaz: Is this `results` needed? Can't it be inferred?

packages/ariakit-scripts/src/ci-visual.test.ts
```diff
@@ -1,11 +1,11 @@
 import { expect, test } from "vitest";
 import { getVisualProject } from "./ci-visual.js";
 
-test("gets the browser project from direct and reusable workflow job names", () => {
-  expect(getVisualProject("Test Chrome")).toBe("chrome");
-  expect(getVisualProject("App / Test Firefox")).toBe("firefox");
+test("gets the browser project from screenshot upload steps", () => {
+  expect(getVisualProject("Upload screenshots (chrome)")).toBe("chrome");
+  expect(getVisualProject("Upload screenshots (Safari)")).toBe("safari");
```
diegohaz: Shouldn't it be `(ubuntu-latest)` and `(macos-latest)`?

.github/workflows/app.yml
diegohaz: Implemented in `9a68ca7`. The local action now accepts space-separated `engines`, installs them as separate Playwright arguments, and normalizes the cache-key segment to dashes. The Ubuntu visual job installs `chromium firefox` with one action call, and every repository call site has been migrated to the plural input.

.github/workflows/app.yml
diegohaz: Implemented in `9a68ca7`. The visual matrix now has only `engines`, `projects`, and `test` alongside the runner metadata; the primary/secondary engine and project fields and the second installer/upload steps are gone.

.github/workflows/app.yml
diegohaz: Removed in `9a68ca7`. Failed visual output is inferred from `matrix.os` as `app-test-results-${{ matrix.os }}`, which also keeps it outside the `app-visual-*` screenshot namespace.

packages/ariakit-scripts/src/ci-visual.test.ts
diegohaz: Yes, updated in `9a68ca7`. The upload steps and helper tests now use `Upload screenshots (ubuntu-latest)` and `Upload screenshots (macos-latest)`, and the finalizer expects the matching OS-grouped artifacts.

.github/workflows/app.yml
```diff
@@ -150,70 +206,79 @@ jobs:
...
-  test-finalize:
+  visual-finalize:
     # This job intentionally assumes PRs come from branches on this repository.
     # Fork-safe write-back logic was removed with the member-only PR model.
-    name: Visual
-    needs: test
-    if: always() && needs.test.result != 'cancelled'
+    name: Update screenshots
+    needs: visual
+    if: always() && needs.visual.result != 'cancelled'
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`d9507ccaf`](https://github.com/ariakit/ariakit/commit/d9507ccaf41ad549a48fc51bb8c460465670cc47): restored the browser-test dependency only for screenshot write-back, so visual suites still run concurrently while generated pushes cannot cancel unfinished browser jobs. Evidence gap: GitHub-hosted `cancel-in-progress` behavior cannot be red-checked locally; the dependency graph, cancellation guard, and workflow YAML were verified statically.

diegohaz:
**Fixed**

- [`d9507ccaf`](https://github.com/ariakit/ariakit/commit/d9507ccaf41ad549a48fc51bb8c460465670cc47) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680291979)): Waited for browser tests only before screenshot write-back, preserving concurrent visual execution while preventing generated pushes from canceling unfinished browser jobs.

**Evidence gaps**

- GitHub-hosted `cancel-in-progress` behavior cannot be red-checked locally; the dependency graph, cancellation guard, and workflow YAML were verified statically.

.github/workflows/app.yml
```diff
@@ -150,70 +206,79 @@ jobs:
...
-  test-finalize:
+  visual-finalize:
     # This job intentionally assumes PRs come from branches on this repository.
     # Fork-safe write-back logic was removed with the member-only PR model.
-    name: Visual
-    needs: test
-    if: always() && needs.test.result != 'cancelled'
+    name: Update screenshots
+    needs: [test, visual]
+    if: always() && needs.test.result != 'cancelled' && needs.visual.result != 'cancelled'
```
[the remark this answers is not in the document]
diegohaz: See https://github.com/ariakit/ariakit/pull/6973#discussion_r3680227673

.github/workflows/app.yml
```diff
@@ -137,11 +132,72 @@ jobs:
...
         id: test-visual
-        if: matrix.visual != ''
         # -x stops after the first failing test so we can rerun once with -u below.
-        run: ${{ matrix.visual }} --project ${{ matrix.project }} --grep @visual --pass-with-no-tests -x
+        run: |
+          projects=()
+          for project in $VISUAL_PROJECTS; do
+            projects+=(--project "$project")
+          done
+          ${{ matrix.test }} "${projects[@]}" --grep @visual --pass-with-no-tests -x
```
diegohaz: Both visual steps rebuild the same `--project` argument array in shell, but `playwright test` already accepts a list: its CLI declares `--project <project-name...>`, so `--project chrome firefox` is valid. The array is the only reason `Test visual` and `Update screenshots` (lines 209-213) need block scalars at all, and the two copies can drift independently.

Verified locally with this repository's Playwright 1.62.0 against a throwaway three-project config:

```
$ playwright test --project chrome firefox --grep @visual --list
  [chrome] > a.spec.mjs:2:1 > visual thing @visual
  [firefox] > a.spec.mjs:2:1 > visual thing @visual
  Total: 2 tests in 1 file
```

Interpolating `matrix.projects` adds no injection surface that isn't already there: `${{ matrix.test }}` is interpolated into the same command line, and both values are workflow-authored matrix literals. `VISUAL_PROJECTS` is still needed for the `Remove stale screenshots` loop and the `Pack screenshots` Python block, so it stays.

I applied both edits locally and confirmed `app.yml` still parses and both steps render as expected.

```suggestion
        run: ${{ matrix.test }} --project ${{ matrix.projects }} --grep @visual --pass-with-no-tests -x
```

The same collapse applies to `Update screenshots`:

```yaml
      - name: Update screenshots
        id: update-screenshots
        if: failure() && steps.test-visual.outcome == 'failure'
        # Record the pre-update timestamp so we can delete stale screenshots
        # that are no longer regenerated by this rerun.
        run: |
          touch -t "$(date -u -d '-1 second' '+%Y%m%d%H%M.%S' 2>/dev/null || date -u -v-1S '+%Y%m%d%H%M.%S')" "${{ runner.temp }}/pre-update-marker"
          ${{ matrix.test }} --project ${{ matrix.projects }} --grep @visual -u
```

**Assessment**

- Severity: Low (verified, 95% confidence), because this is duplicated glue with no behavior difference.
- Priority: P2 (estimate, 85% confidence), because the PR is already reshaping these two steps, so collapsing them now avoids landing the duplication.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because breakage requires a future edit that updates one copy of the loop and not the other.
- Fix effort: Easy (verified, 95% confidence), because it removes eight lines from two steps and I confirmed the replacement parses and matches Playwright's CLI contract.

packages/ariakit-scripts/src/ci-visual.js
```diff
@@ -1,8 +1,7 @@
...
- * @param {string} jobName
+ * @param {string} stepName
  */
-export function getVisualProject(jobName) {
-  const localName = jobName.split(" / ").at(-1);
-  if (!localName?.startsWith("Test ")) return;
-  return localName.slice(5).toLowerCase();
+export function getVisualPlatform(stepName) {
+  const match = stepName.match(/^Upload screenshots \(([^)]+)\)$/);
+  return match?.[1]?.toLowerCase();
```
diegohaz: Lowercasing the captured platform can now only introduce a mismatch, never fix one. Both sides of the comparison derive from the same `matrix.os` value: `.github/workflows/app.yml` line 252 names the step `Upload screenshots (${{ matrix.os }})` and line 256 names the artifact `app-visual-${{ matrix.os }}` — raw, unnormalized. The finalizer then expects `app-visual-${getVisualPlatform(stepName)}`, which is the lowercased form.

The normalization made sense for `getVisualProject`, which derived `chrome` from the job name `Test Chrome`. That derivation is gone in this PR.

Real-world impact: `runs-on` labels are case-insensitive and `macOS-latest` is a spelling GitHub uses in its own docs, so a maintainer editing the visual matrix (pinning a runner, adding a Windows entry) can reasonably write a label containing uppercase. The finalizer would then expect `app-visual-macos-latest` while the run uploaded `app-visual-macOS-latest`:

```
expected: [ 'app-visual-ubuntu-latest', 'app-visual-macos-latest' ]
found:    [ 'app-visual-ubuntu-latest', 'app-visual-macOS-latest' ]
missing:  [ 'app-visual-macos-latest' ] => ready: false
```

I reproduced that with a local simulation of the `check-artifacts` script. The failure is silent: `hasCompleteScreenshots` is `false`, the finalize job still succeeds, and the only signal is a `Missing screenshot artifacts:` log line. Automated screenshot updates just stop happening on every baseline-changing PR until someone reads the finalize log.

Dropping the normalization is the right direction rather than normalizing the artifact name, because GitHub expressions have no lowercase function, so `app-visual-${{ matrix.os }}` cannot be normalized at the upload site.

```suggestion
  return match?.[1];
```

Worth noting for `ci-visual.test.ts`: both current inputs (`ubuntu-latest`, `macos-latest`) are already lowercase, so the normalization has no coverage either way. If you keep it, it needs a case that actually exercises it; if you drop it, the existing assertions stay correct as raw-boundary assertions.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because screenshot write-back stops entirely and fails silently, but committed baselines are not corrupted.
- Priority: P2 (estimate, 75% confidence), because the current matrix values are lowercase, so nothing breaks today; this is removing a trap the PR introduces.
- Breakage likelihood: Unlikely (estimate, 80% confidence), because it requires someone to add or edit a visual matrix entry with an uppercase runner label.
- Fix effort: Easy (verified, 95% confidence), because it deletes one method call and the existing tests still pass.

.github/workflows/visual.yml
```diff
@@ -107,41 +107,44 @@ jobs:
...
+                    platform: getVisualPlatform(step.name),
+                    conclusion: step.conclusion,
+                  };
+                });
+            });
+            // Valid platform steps still count when skipped so an interrupted
+            // visual shard cannot make its screenshot artifact optional.
+            const invalidScreenshotSteps = screenshotSteps.filter(
+              (step) => !step.platform && step.conclusion !== "skipped",
+            );
```
diegohaz: The `conclusion !== "skipped"` exemption is now unsafe in exactly the way the comment above it warns about, and the comment is attached to the wrong statement.

`screenshotSteps` is already filtered to names starting with `Upload screenshots (` (line 116), and `getVisualPlatform` matches `/^Upload screenshots \(([^)]+)\)$/`. So `step.platform` can only be falsy when the parenthesized value is empty — that is, when a visual matrix entry contributes an empty `os`. In that case this filter drops the step when it is `skipped`, the step also never lands in `expectedScreenshotArtifacts` (line 132 filters on `step.platform`), and the finalizer proceeds with an incomplete expectation set. Because `Prepare screenshots` (line 194) deletes the whole `__screenshots__` tree before extracting, a write-back with a missing platform commits the deletion of that platform's baselines.

The exemption was load-bearing for the previous `matrix.secondary_project` shape, where Safari intentionally rendered `Upload screenshots ()`. This PR removes that field, so the only remaining reachable cause of an unparseable name is a mistake, and a mistake should block write-back rather than be exempted.

Separately, the comment on lines 126-127 explains why steps are counted regardless of conclusion, which is a property of `expectedScreenshotArtifacts` below, not of this filter.

```suggestion
            const invalidScreenshotSteps = screenshotSteps.filter(
              (step) => !step.platform,
            );
            // Steps count regardless of their conclusion so an interrupted
            // visual shard cannot make its screenshot artifact optional.
```

With that change, `conclusion: step.conclusion` on line 122 becomes unused and can be dropped from the mapped object.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because the exempted path skips a baseline platform during a full-tree replacement, which commits deletions rather than just skipping an update.
- Priority: P2 (estimate, 80% confidence), because the path is unreachable with the matrix as written in this PR; this removes dead-but-misleading tolerance and fixes the comment.
- Breakage likelihood: Rare (estimate, 85% confidence), because it requires a future visual matrix entry with an empty `os` value.
- Fix effort: Easy (verified, 90% confidence), because it removes one condition and one unused field; I applied it locally and `pnpm test packages/ariakit-scripts/src/ci-visual.test.ts` still passes.

diegohaz (commented):
Initial review of the full PR. Baseline: merge base `777c51061` to head `d9507ccaf`, then expanded to the surrounding CI context (`ci.yml` gate wiring, `commit-or-pr`, `app/playwright.config.ts`, the committed `__screenshots__` tree, and every `install-playwright` call site).

The direction is right. Splitting visual coverage into its own matrix removes the real coupling where a browser-test failure suppressed visual coverage for that shard, and keeping only the finalizer dependent on `test` is a smaller correction than serializing the matrices again. Grouping one artifact per OS with an all-or-nothing pack is a genuine improvement over the previous per-project artifacts, because `Prepare screenshots` replaces the whole tree and a partial upload would commit deletions.

## Checks run

- `pnpm install --frozen-lockfile`, `pnpm test packages/ariakit-scripts/src/ci-visual.test.ts` (2 passed) on the PR head and again on each local prototype below.
- Confirmed `playwright test --project` is variadic (`--project <project-name...>`, `playwright/lib/program.js:211`) and exercised it against this repo's Playwright 1.62.0 with a throwaway three-project config in a scratch directory: `--project chrome firefox --grep @visual --list` listed both projects.
- Prototyped all three inline suggestions locally, re-parsed `app.yml` and `visual.yml` with the `yaml` package, re-ran the focused test, then restored the tree (`git diff` clean).
- Simulated the `check-artifacts` expectation/artifact matching in Node with the current `getVisualPlatform`, including an uppercase runner label, to confirm the divergence described in the `ci-visual.js` comment.
- Verified every `install-playwright` caller was migrated to the plural input (`app.yml` x2, `og-images.yml`, `perf.yml`, `plus.yml`); no `engine:` action input remains, and the remaining `engine:` occurrences are the browser matrix's own field name.
- Verified `Gate` is the only required status check on `main` (ruleset `7448583`), so renaming `test-finalize` to `visual-finalize` and the job display name to `Update screenshots` cannot orphan a required check. `Gate` still gates on the aggregate `app` result, which now includes the new `visual` matrix.
- Checked the artifact plumbing: the grouped upload uses a glob, and `actions/upload-artifact` roots a single matched file at its own directory, so the macOS artifact still contains a bare `app-visual-safari.tar.gz`; `merge-multiple: true` plus the `app-visual-*.tar.gz` extract loop reassembles all three projects with no name overlap today.
- Checked cache behavior in the composite action: the new `chromium firefox` key can only partial-hit `playwright-Linux-`, and the install step is guarded on `cache-hit != 'true'`, so a partial restore still installs both engines.

## Prior review state

All seven existing review threads on this PR are resolved, so none were re-adjudicated. The two CodeRabbit items that lived in its review body rather than in threads were checked against the current head anyway: the `VISUAL_PROJECTS`-in-Python suggestion is already applied (`app.yml:239`), and the `matrix.secondary_project` documentation suggestion is obsolete because that field no longer exists.

## Findings published inline

- `.github/workflows/app.yml`: `Test visual` and `Update screenshots` each rebuild the same shell array; Playwright's `--project` already takes a list.
- `packages/ariakit-scripts/src/ci-visual.js`: `.toLowerCase()` normalizes one side of a comparison whose other side is the raw `matrix.os` value, so it can only introduce a silent mismatch.
- `.github/workflows/visual.yml`: the `conclusion !== "skipped"` exemption for unparseable step names is the hazard the adjacent comment warns about, and the comment describes the statement below it rather than the one it sits on.

## Candidates not published inline

- **`if-no-files-found: ignore` on `Upload screenshots` is now unreachable** — Declined non-goal. `Pack screenshots` raises when any configured project produces no screenshots, and the upload is gated on `steps.pack-screenshots.outcome == 'success'`, so the glob always matches. Leaving `ignore` is benign: if it somehow matched nothing, `check-artifacts` would see a missing expected artifact and block write-back, which is the safe outcome. Changing it to `error` would only move where the failure surfaces.
- **Test-results artifact names use two shapes** — Declined non-goal. The browser matrix uploads `app-${{ matrix.project }}-test-results` while the visual matrix uploads `app-test-results-${{ matrix.os }}`. Both stay clear of the `app-visual-*` namespace, so this is naming preference only.
- **Per-project tarball names are only unique because no project runs on two runners** — Declined non-goal. Inside `merge-multiple: true`, two OS artifacts containing the same `app-visual-<project>.tar.gz` would silently overwrite each other while `check-artifacts` still saw both OS artifacts present. That needs a future matrix entry running an existing project on a second OS, which contradicts how projects map to engines today. Worth remembering if a Windows visual runner is ever added.
- **Added CI cost** — not a defect, but naming the trade-off: the split adds two jobs, one of them a macOS runner, each repeating `Set up environment`, Playwright install, and both artifact downloads. That is inherent to the stated goal rather than something to fix here.

## Not fully validated

While tracing the moved-but-unchanged `Update screenshots` and `Remove stale screenshots` steps I could not reconcile their contract with Playwright 1.62's flag semantics, and I could not settle it either way, so I am recording it as a suspicion rather than a finding.

`-u` resolves to `--update-snapshots=changed` (`playwright/lib/program.js:226` declares `preset: "changed"`), and in that mode `toHaveScreenshot` returns `handleMatching()` without writing when the comparison passes. Files that still match therefore keep their checkout mtime, which is older than `pre-update-marker`, so `find ... ! -newer "$marker" -delete` would delete them — contradicting the step comment that "the update run only touches files that still exist". Under that reading, a single mismatching screenshot would leave only the regenerated files in the artifact, and the full-tree replacement in `visual.yml` would commit the rest as deletions.

Against that reading, the actual `Update screenshots` commits on `main` show no such deletions (`831ea80e0`: 144 modified, 0 deleted; `d84fd8717`: 6 added, 0 deleted), even though `commit-or-pr` stages with `git add -A` and both steps have existed since `b4ca5807c` (March 2026). So either my reading of the update mode is wrong or those runs happened to regenerate every screenshot for the affected project. Testing this properly needs a CI run, since visual screenshots require `VISUAL_TEST=true` and `CI=true` and are OS-dependent. Either way it predates this PR and does not block it; flagging it because this PR is the natural moment to confirm the assumption.

.github/workflows/app.yml
diegohaz: Fixed in [`d3ad62c12`](https://github.com/ariakit/ariakit/commit/d3ad62c1243d63d2de23fd7c21618a934a23e885): replaced the duplicated shell arrays with Playwright’s variadic `--project` argument while preserving every existing flag; the repository CLI selected all 32 Chrome and Firefox visual tests.

packages/ariakit-scripts/src/ci-visual.js
diegohaz: Fixed in [`cab044980`](https://github.com/ariakit/ariakit/commit/cab044980c7374d99f200bdc2973867626e806bd): preserved the exact platform label shared by the upload step and artifact name. Added targeted mixed-case raw-boundary coverage in [`cab044980`](https://github.com/ariakit/ariakit/commit/cab044980c7374d99f200bdc2973867626e806bd).

.github/workflows/visual.yml
diegohaz: Fixed in [`1a1b341ac`](https://github.com/ariakit/ariakit/commit/1a1b341acdc864548f6ce13bdd3aec949121aa13): removed the skipped-step exemption so every malformed screenshot-upload step blocks write-back while valid steps remain required regardless of conclusion. Evidence gap: this logic runs inline against GitHub-hosted job and artifact metadata, so no durable local Actions test was added; the unsafe branch was verified with a red/green simulation plus YAML and static graph checks.

diegohaz:
**Fixed**

- [`d3ad62c12`](https://github.com/ariakit/ariakit/commit/d3ad62c1243d63d2de23fd7c21618a934a23e885) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680475756)): Removed duplicate shell arrays and passed the visual project list directly to Playwright's variadic `--project` option.
- [`cab044980`](https://github.com/ariakit/ariakit/commit/cab044980c7374d99f200bdc2973867626e806bd) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680479171)): Preserved exact platform labels between upload step names and artifact names, with mixed-case raw-boundary regression coverage.
- [`1a1b341ac`](https://github.com/ariakit/ariakit/commit/1a1b341acdc864548f6ce13bdd3aec949121aa13) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680481675)): Made every malformed screenshot-upload step block write-back, including skipped steps.

**Evidence gaps**

- [`1a1b341ac`](https://github.com/ariakit/ariakit/commit/1a1b341acdc864548f6ce13bdd3aec949121aa13) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680481675)): The inline GitHub-hosted job/artifact path has no durable local Actions test; the unsafe branch was verified with a red/green simulation plus YAML and static graph checks.

.github/workflows/app.yml
```diff
@@ -150,70 +201,75 @@ jobs:
         # that are no longer regenerated by this rerun.
         run: |
           touch -t "$(date -u -d '-1 second' '+%Y%m%d%H%M.%S' 2>/dev/null || date -u -v-1S '+%Y%m%d%H%M.%S')" "${{ runner.temp }}/pre-update-marker"
-          ${{ matrix.visual }} --project ${{ matrix.project }} --grep @visual -u
+          ${{ matrix.test }} --project ${{ matrix.projects }} --grep @visual -u
```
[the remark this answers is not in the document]
diegohaz: The Playwright premise here is right, but the conclusion does not hold, because stale detection never depended on Playwright rewriting files. `visual()` stamps every screenshot it asserts:

```ts
await expect(page).toHaveScreenshot(fileSnapshotName, { ...screenshotOptions });
// Touch the screenshot file so the CI stale-detection step
// (which deletes files older than a pre-run marker) knows
// this screenshot is still expected by a test.
touchScreenshot(testInfo, fileSnapshotName);
```

`app/src/test-utils/visual.ts:398-404`, where `touchScreenshot` calls `utimesSync(screenshotPath, now, now)` (lines 122-141). In the `-u` rerun, `toHaveScreenshot` passes both when it matches (`handleMatching`) and when it rewrites, so the stamp always runs. Matching Firefox screenshots therefore end up newer than `pre-update-marker`, survive the `find ... ! -newer` loop, and the pack step still finds all 172.

I red/green checked this rather than reasoning from the source alone. Minimal Playwright project in a scratch directory, Playwright 1.62.0 from this repo, one `toHaveScreenshot` assertion, baseline mtime forced to 2026-01-01, marker created one second before the rerun:

```
without helper touch -> playwright exit=0, would_be_stale_deleted=1
with helper touch    -> playwright exit=0, would_be_stale_deleted=0
```

The first line is your premise reproduced: bare `-u` really does leave a matching snapshot unwritten. The second is the actual repository path.

Two follow-on notes:

The suggested `all` mode is not a reliable fix on its own. `toHaveScreenshot` under `all` sends `expected: undefined` to `_expectScreenshot` and then rewrites only when `compareBuffersOrStrings(actual, expected)` reports a byte difference; a byte-identical screenshot still falls through to `handleMatching()`. Same experiment with `--update-snapshots=all` and no helper stamp:

```
update-snapshots=all, no helper touch -> would_be_stale_deleted=1
```

So switching modes would trade a working mechanism for one that happens to work only when rendering is non-deterministic, and it would also drop the baseline from the browser-side comparison.

The combined Ubuntu shard does not change this. It does mean a Chrome-only mismatch now runs `-u` over Firefox as well, which was not true when each project had its own shard, but every Firefox screenshot is re-stamped in that rerun, so the blast radius is unchanged.

Disposition: invalid, no change required. Recording it here rather than silently resolving because this is the second time in this review cycle that the `-u` semantics have been read as the stale-detection contract; I raised the same suspicion in my previous summary and it was wrong for the same reason.

.github/workflows/app.yml
```diff
@@ -150,70 +201,75 @@ jobs:
...
 
       - name: Remove stale screenshots
-        if: matrix.visual != '' && always() && steps.update-screenshots.outcome == 'success'
+        if: always() && steps.update-screenshots.outcome == 'success'
         # The update run only touches files that still exist, so remove any
-        # older screenshots for this browser that were left behind.
-        run: find app/src -path "*/__screenshots__/*-${{ matrix.project }}.webp" ! -newer "${{ runner.temp }}/pre-update-marker" -delete
+        # older screenshots for these browsers that were left behind.
```
diegohaz: This comment is accurate but leaves out the part that makes it true, and that gap is now demonstrably expensive: within this review cycle alone, two independent reviewers read "the update run only touches files that still exist" as a claim about Playwright rewriting snapshots, concluded the loop deletes every matching baseline, and proposed changes based on that. See the thread on line 204 for the second instance.

What actually stamps the files is `visual()`, which calls `utimesSync` on each screenshot right after asserting it (`app/src/test-utils/visual.ts:398-404`). That side of the invariant is documented; this side is not, and nothing in this file points at it. The failure mode is not hypothetical: acting on the misreading means switching to `--update-snapshots=all` (which does not reliably rewrite matching files either) or deleting `touchScreenshot` as unused, and the latter turns this loop into a mass baseline deletion with no test or type error to catch it.

Since this PR is already rewording this comment for the multi-project loop, it is the cheapest moment to close the loop.

```suggestion
        # The update run re-stamps every screenshot a test still asserts
        # (`visual()` in app/src/test-utils/visual.ts), so remove any
        # older screenshots for these browsers that were left behind.
```

**Assessment**

- Severity: Medium (estimate, 70% confidence), because the invariant is invisible from both the workflow and any test, so a change made against the misreading destroys committed baselines rather than failing loudly.
- Priority: P2 (estimate, 80% confidence), because nothing is broken at this head; this prevents a plausible future regression and the PR is already editing the line.
- Breakage likelihood: Possible (verified, 90% confidence), because two reviewers reached the wrong conclusion from this comment during this PR's own review, and one of them proposed acting on it.
- Fix effort: Easy (verified, 95% confidence), because it is a comment edit with no behavior change.

packages/ariakit-scripts/src/ci-visual.test.ts
```diff
@@ -1,11 +1,18 @@
...
+  expect(getVisualPlatform("Upload screenshots (macOS-latest)")).toBe(
+    "macOS-latest",
+  );
 });
 
-test("ignores unrelated workflow jobs", () => {
-  expect(getVisualProject("App / Build app")).toBeUndefined();
+test("ignores unrelated workflow steps", () => {
+  expect(getVisualPlatform("Upload test results to GitHub")).toBeUndefined();
 });
```
diegohaz, line 19: `1a1b341ac` promoted `invalidScreenshotSteps` from a tolerated case to a hard blocker: any step whose name reaches `getVisualPlatform` and returns `undefined` now forces `hasCompleteScreenshots` to `false` and stops screenshot write-back. That guard has exactly one reachable trigger in production, and it is not covered here.

`Upload test results to GitHub` does not exercise it, because `visual.yml:115-117` filters on `step.name.startsWith("Upload screenshots (")` before calling the helper. The only input that both survives that filter and returns `undefined` is `Upload screenshots ()` — that is, a visual matrix entry whose `os` renders empty. That is the case the workflow guard exists for, and the case a future matrix edit would produce.

```suggestion
test("ignores unrelated workflow steps", () => {
  expect(getVisualPlatform("Upload test results to GitHub")).toBeUndefined();
  expect(getVisualPlatform("Upload screenshots ()")).toBeUndefined();
});
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because the missing case protects a guard rather than shipped behavior.
- Priority: P2 (estimate, 85% confidence), because the guard is correct at this head; this locks the contract `visual.yml` now depends on.
- Breakage likelihood: Rare (estimate, 85% confidence), because it requires someone to loosen the regex or the empty-parenthesis handling without noticing the workflow relies on it.
- Fix effort: Easy (verified, 95% confidence), because it is one assertion in an existing test.

diegohaz (commented):
Re-review. Incremental baseline is `d9507ccaf`, the head recorded in my previous review marker, to the current head `1a1b341ac`. New findings below come from the three commits since that SHA plus surrounding issues they newly expose, and the one unresolved thread relevant to this head was independently validated against the current code rather than taken at face value.

## Previous findings

All three are fixed as suggested, and I re-verified each against the current head rather than the commit messages.

- `d3ad62c12` collapses both shell arrays to Playwright's variadic `--project`. Both steps now render as a single command, and `VISUAL_PROJECTS` correctly survives for the two consumers that still need a list (`app.yml:211`, `app.yml:230`).
- `cab044980` drops the normalization and adds `Upload screenshots (macOS-latest)` returning `macOS-latest`. That is the raw-boundary assertion the previous shape was missing: it now fails if anyone reintroduces case folding on one side of the artifact-name comparison.
- `1a1b341ac` removes the skipped-step exemption and the unused `conclusion` field, and moves the counting comment onto the statement it describes.

`pnpm test packages/ariakit-scripts/src/ci-visual.test.ts` passes at this head.

## Unresolved thread validated

**"Rewrite matching snapshots before stale cleanup"** (`chatgpt-codex-connector`, `app.yml:204`) — **invalid, no change required**; replied on the thread with the evidence.

Its Playwright premise is correct and I reproduced it: bare `-u` resolves to the `changed` preset and does not rewrite a matching snapshot, so the file keeps its old mtime. Its conclusion does not follow, because the stale detection never relied on Playwright doing the rewriting. `visual()` calls `utimesSync` on every screenshot immediately after asserting it (`app/src/test-utils/visual.ts:398-404`), and that assertion passes in the `-u` rerun whether the snapshot matched or was rewritten.

I red/green checked the mechanism in a scratch Playwright project on 1.62.0, forcing the baseline mtime to 2026-01-01 and creating the marker one second before the rerun:

```
without helper touch                  -> would_be_stale_deleted=1
with helper touch                     -> would_be_stale_deleted=0
update-snapshots=all, no helper touch -> would_be_stale_deleted=1
```

The third line matters for the proposed fix: `all` sends `expected: undefined` to `_expectScreenshot` and then rewrites only on a byte difference, so a byte-identical screenshot still falls through to `handleMatching()`. Switching modes would replace a working mechanism with one that only works when rendering is non-deterministic.

This also retires the suspicion I recorded in my previous summary under "Not fully validated" — same claim, wrong for the same reason. The `main` history I could not reconcile then (`831ea80e0`: 144 modified, 0 deleted) is explained: those screenshots were re-stamped by the helper, not by Playwright.

## New findings published inline

Both are consequences of the fixes above rather than defects in them.

- `.github/workflows/app.yml:208-209` — the `Remove stale screenshots` comment states the invariant without naming what upholds it. Two independent reviewers misread it during this PR's own review cycle, and one proposed acting on the misreading, so the gap has a measured cost. This PR is already rewording the line.
- `packages/ariakit-scripts/src/ci-visual.test.ts:16-18` — `1a1b341ac` made `invalidScreenshotSteps` a hard write-back blocker, and its only reachable production trigger (`Upload screenshots ()`, from a matrix entry with an empty `os`) has no coverage. The existing negative case never reaches the helper in production because `visual.yml:115-117` filters on the `Upload screenshots (` prefix first.

## Unchanged from the previous review

The declined non-goals from my previous summary still stand at this head and are unaffected by these commits: the vestigial `if-no-files-found: ignore`, the two test-results artifact naming shapes, the per-project tarball collision that needs a project on two runners, and the added macOS runner cost. No new candidate in this round was withheld from publication.

.github/workflows/app.yml
diegohaz: Fixed in [`384b7eba5`](https://github.com/ariakit/ariakit/commit/384b7eba5b8ed8e79e0d1c58a0cc51271b0809d4): the stale-cleanup comment now points to `visual()` re-stamping every still-asserted screenshot before the mtime sweep.

packages/ariakit-scripts/src/ci-visual.test.ts
diegohaz: Already handled at `packages/ariakit-scripts/src/ci-visual.js:5`: the regex requires a non-empty platform, so empty parentheses return `undefined` and block write-back. Added targeted coverage in [`c8d9462fb`](https://github.com/ariakit/ariakit/commit/c8d9462fb486839fca8d7b407ab9c212c14e8b5b).

diegohaz:
**Fixed**

- [`384b7eba5`](https://github.com/ariakit/ariakit/commit/384b7eba5b8ed8e79e0d1c58a0cc51271b0809d4) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680653194)): Documented that `visual()` re-stamps every still-asserted screenshot before stale-file cleanup.

**Already resolved**

- [`c8d9462fb`](https://github.com/ariakit/ariakit/commit/c8d9462fb486839fca8d7b407ab9c212c14e8b5b) ([reply](https://github.com/ariakit/ariakit/pull/6973#discussion_r3680657012)): The helper already rejects empty platform names; added exact `Upload screenshots ()` boundary coverage and red-checked it against a loosened regex.

diegohaz (commented):
Re-review. Incremental baseline is `1a1b341ac`, the head recorded in my previous review marker, to the current head `c8d9462fb`. No new findings. Every review thread on this PR is resolved and none were reopened by these commits.

## Previous findings

Both are fixed, verified against the current code rather than the commit messages.

- `384b7eba5` names the mechanism in the `Remove stale screenshots` comment (`app.yml:208-210`). The wording is accurate against the helper: `visual()` stamps each screenshot immediately after `toHaveScreenshot` passes, and that assertion passes in the `-u` rerun whether the snapshot matched or was rewritten, so "every screenshot still asserted through `visual()`" is the right scope. It also reads correctly against the `if:` guard above it, since the loop only runs when the update step succeeded.
- `c8d9462fb` adds `expect(getVisualPlatform("Upload screenshots ()")).toBeUndefined()`. That is the input class the `invalidScreenshotSteps` blocker actually sees in production, so the guard `1a1b341ac` introduced is now pinned.

## Checks run at this head

- `pnpm test packages/ariakit-scripts/src/ci-visual.test.ts` — 2 passed.
- `pnpm lint` — clean (oxlint plus oxfmt check, 2028 files).
- Re-parsed `app.yml`, `visual.yml`, `install-playwright/action.yml`, `og-images.yml`, `perf.yml`, and `plus.yml`, and re-read the job graph from the parsed document: the visual matrix is the intended two entries, and `visual-finalize` still carries `needs: [test, visual]` with the cancellation guard on both.
- Confirmed no `engine:` action input remains anywhere; the five surviving occurrences are the browser matrix's own field name, which is passed through as `engines: ${{ matrix.engine }}`.
- Simulated the composite action's normalization step under `bash -eo pipefail` for each real call-site value. Single-engine callers produce `chromium` and `webkit`, byte-identical to the keys they used before this PR, so the migration to the plural input introduces no Playwright cache churn for `og-images`, `perf`, `plus`, or the browser matrix; only the new `chromium-firefox` key is cold on first run, and the install step is guarded on `cache-hit != 'true'` so a `playwright-Linux-` prefix restore still installs both engines.

## Standing dispositions

The declined non-goals from the first review still hold and are unaffected by these commits: the vestigial `if-no-files-found: ignore`, the two test-results artifact naming shapes, the per-project tarball collision that would need one project running on two runners, and the added macOS runner cost. The Codex thread on `-u` semantics remains invalid for the reason established in the previous round, and `384b7eba5` is the durable record of why.

Nothing else surfaced across the three rounds. From my side this is ready.

PR #6976, opened 2026-07-30
Center the selected ComboboxSelect item on open

diegohaz:
## Motivation

Opening a `ComboboxSelect` with a far selected value currently stops as soon as
the option reaches the nearest popup edge. In the long-list fixture, Watermelon
remains 43 px below the popup midpoint. The initial presentation should center
the selected option without scrolling outside the popup, while later keyboard
movement should keep nearest-edge behavior.

Fixes #6838.

## Solution

`Composite` now accepts a private scrolling callback so component-specific
presentation behavior stays outside the generic composite primitive:

```tsx
props = useComposite({
  store,
  unstable_scrollIntoView: scrollIntoView,
  ...props,
});
```

`ComboboxSelect` and its input-owned `Combobox` provide that callback. It checks
the live store to ensure the newly rendered item is still the selected value,
then centers it for the initial presentation. The normal `Composite` path still
handles every other item, so later keyboard navigation keeps nearest-edge
scrolling. Delayed item registration is covered as well: when the selected
offscreen item mounts, the pending presentation completes without a stale
second pass.

The centering implementation stays internal to Combobox. It snapshots scroll
positions above the popover boundary, delegates alignment to the browser once,
and restores those positions in a `finally` block:

```ts
element.scrollIntoView({ block: "center", inline: "nearest" });
```

The ancestor walk follows ordinary DOM parents, a ShadowRoot host, and
same-origin iframe elements. This keeps outer panels, the document, and parent
frames stationary while the browser handles transforms, nested scrollports,
scroll padding, and axis-specific overflow.

The dedicated `combobox-center-into-view` sandbox now tests this behavior only
through real Combobox components, including a shadow-slotted list and a
same-origin iframe. The `combobox-select-offscreen-value` sandbox remains
focused on `ComboboxItem offscreenMode`.

## Preview

Before: the selected Watermelon is visible at the bottom edge instead of
centered.

Selected Watermelon remains at the bottom of the popup below its annotated midpoint

After [`0b51dc09e`](https://github.com/ariakit/ariakit/commit/0b51dc09eb4a3baf66136acbde4967c67ae0f9ff):
the selected value is presented near the popup midpoint on open, while
real-focus keyboard movement continues to use nearest-edge scrolling.

ComboboxSelect opens with its far selected Watermelon presented near the popup midpoint

[Video: opening ComboboxSelect presents the selected Watermelon](https://github.com/user-attachments/assets/8d2e2da5-1963-4ec7-b77d-be639aed7c88)

## Testing

[...]

PR #6984, opened 2026-07-31
Avoid document scroll writes when opening ComboboxSelect

diegohaz:
## Motivation

Opening or reopening a searchable `ComboboxSelect` with a mouse ran document-scroll preservation while presenting the selected item. When the page coordinates had not changed, its unconditional correction still called:

```ts
documentScroller.scrollTo({ left, top, behavior: "instant" });
```

On macOS with overlay scrollbars hidden while idle, that no-op write could briefly reveal the page scrollbar.

Removing the preservation wrapper is unsafe, though. The selected option's focus handler redirects focus to the combobox input, and that redirect could move a scrolled page while the popover was still being placed. In the sandbox at 800×400, the first version of this PR changed `window.scrollY` from `272` to `0`.

## Solution

Prevent the underlying virtual-focus redirect from scrolling at its source:

```ts
baseElement.focus({ preventScroll: true });
```

Keep document-scroll preservation around selected-item presentation, but only restore the document when its coordinates actually changed. When the viewport has smooth scrolling, the helper temporarily applies `scroll-behavior: auto !important` to `document.documentElement` so root movement is synchronous and cancellable, then restores the exact prior inline value and priority. Coordinates continue to come from `document.scrollingElement`, which also handles quirks-mode documents where that element is `body`. Nested list scrolling is unchanged.

The browser regression scrolls the page, enables `html { scroll-behavior: smooth !important }`, and performs mouse open → outside click → mouse reopen. It asserts that:

- The selected option remains visible inside the popover.
- The document never leaves its starting coordinates, including intermediate frames.
- No equal-coordinate `scrollTo` write reaches the document scroller.
- The same guarantees hold on reopen.

The no-op detector always forwards real calls. This matters on WebKit, which may require a real restoration to preserve the page while still producing zero redundant writes.

Focused helper tests separately cover the no-op guard and the quirks-mode `body` scroller / root-style mapping.

No changeset is included because the affected select-mode `ComboboxPopover` behavior has not been released yet.

This intentionally does not change selected-item centering or the broader behavior under discussion in #6976.

## Preview

Mouse open → outside click → mouse reopen with root smooth scrolling enabled. The overlay labels show `window.scrollY = 272px` throughout.

[pr6984-scroll-preserved.webm](https://github.com/user-attachments/assets/e1bc8b5c-e5d3-4b01-852f-404583ea8e9c)

## Testing

- `pnpm lint`
- `pnpm tsc`
- `pnpm -F app build-lite`
- `pnpm test packages/ariakit-react-components/src/composite/utils.test.ts --project react`: 2 passed
- Chromium related scroll, offscreen-item, animation, and tabs suites: 32 passed
- Firefox related suites: 32 passed
- WebKit related suites, including smooth scrolling: 33 passed
- Final focused suites after the no-op and quirks-mode follow-up: Chromium 3 passed, Firefox 3 passed, WebKit 4 passed

### DIFF
--- a/.changeset/300-combobox-select-document-scroll.md
+++ b/.changeset/300-combobox-select-document-scroll.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed a select [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) with a [`Combobox`](https://ariakit.com/reference/combobox) inside writing to the document scroll position when opened, which could briefly reveal overlay scrollbars on macOS.
--- a/.changeset/300-combobox-select-document-scroll.md
+++ b/.changeset/300-combobox-select-document-scroll.md
@@ -1,6 +1,0 @@
----
-"@ariakit/react-components": patch
-"@ariakit/react": patch
----
-
-Fixed a select [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) with a [`Combobox`](https://ariakit.com/reference/combobox) inside writing to the document scroll position when opened, which could briefly reveal overlay scrollbars on macOS.
--- a/app/src/sandbox/combobox-select-tabs/test-browser.ts
+++ b/app/src/sandbox/combobox-select-tabs/test-browser.ts
@@ -3,46 +3,52 @@ import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
 const cases = ["Select with Tab", "Select with Combobox and Tab"];
 
 withFramework(import.meta.dirname, async ({ test }) => {
-  test("reopens the searchable select without writing the document scroll position", async ({
+  test("opens and reopens the searchable select without scrolling the page", async ({
     page,
     q,
   }) => {
-    const select = q.combobox("Select with Combobox and Tab");
+    await page.setViewportSize({ width: 800, height: 400 });
+
+    const select = q.combobox("Select with Combobox and manual Tab");
     const dialog = q.dialog();
 
-    await select.click();
-    await test.expect(dialog).toBeVisible();
-    await page.mouse.click(1, 1);
-    await test.expect(dialog).toHaveCount(0);
-
-    await page.evaluate(() => {
-      const scroller = document.scrollingElement;
-      if (!scroller) throw new Error("Document scrolling element is missing");
-      const scrollTo = scroller.scrollTo.bind(scroller);
-      document.documentElement.dataset.documentScrollCalls = "0";
-      Object.defineProperty(scroller, "scrollTo", {
-        configurable: true,
-        value: (...args: unknown[]) => {
-          const { dataset } = document.documentElement;
-          const calls = Number(dataset.documentScrollCalls);
-          dataset.documentScrollCalls = String(calls + 1);
-          Reflect.apply(scrollTo, undefined, args);
-        },
+    await select.scrollIntoViewIfNeeded();
+    const scrollY = await page.evaluate(() => window.scrollY);
+    test.expect(scrollY).toBeGreaterThan(0);
+    await page.evaluate((scrollY) => {
+      const style = document.createElement("style");
+      style.textContent = "html { scroll-behavior: smooth !important }";
+      document.head.append(style);
+      window.addEventListener("scroll", () => {
+        if (window.scrollY === scrollY) return;
+        document.documentElement.dataset.scrollMoved = "true";
       });
-    });
+    }, scrollY);
 
     await select.click();
     await test.expect(dialog).toBeVisible();
     await test.expect(q.option("main")).toBeInViewport();
-    await flushFrames(page, 3);
-
+    await flushFrames(page);
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(scrollY);
     test
       .expect(
-        await page.evaluate(
-          () => document.documentElement.dataset.documentScrollCalls,
-        ),
+        await page.evaluate(() => document.documentElement.dataset.scrollMoved),
       )
-      .toBe("0");
+      .toBeUndefined();
+
+    await page.mouse.click(1, 1);
+    await test.expect(dialog).toHaveCount(0);
+
+    await select.click();
+    await test.expect(dialog).toBeVisible();
+    await test.expect(q.option("main")).toBeInViewport();
+    await flushFrames(page);
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(scrollY);
+    test
+      .expect(
+        await page.evaluate(() => document.documentElement.dataset.scrollMoved),
+      )
+      .toBeUndefined();
   });
 
   for (const label of cases) {
--- a/app/src/sandbox/combobox-select-tabs/test-browser.ts
+++ b/app/src/sandbox/combobox-select-tabs/test-browser.ts
@@ -1,8 +1,50 @@
-import { withFramework } from "#app/test-utils/preview.ts";
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
 
 const cases = ["Select with Tab", "Select with Combobox and Tab"];
 
 withFramework(import.meta.dirname, async ({ test }) => {
+  test("reopens the searchable select without writing the document scroll position", async ({
+    page,
+    q,
+  }) => {
+    const select = q.combobox("Select with Combobox and Tab");
+    const dialog = q.dialog();
+
+    await select.click();
+    await test.expect(dialog).toBeVisible();
+    await page.mouse.click(1, 1);
+    await test.expect(dialog).toHaveCount(0);
+
+    await page.evaluate(() => {
+      const scroller = document.scrollingElement;
+      if (!scroller) throw new Error("Document scrolling element is missing");
+      const scrollTo = scroller.scrollTo.bind(scroller);
+      document.documentElement.dataset.documentScrollCalls = "0";
+      Object.defineProperty(scroller, "scrollTo", {
+        configurable: true,
+        value: (...args: unknown[]) => {
+          const { dataset } = document.documentElement;
+          const calls = Number(dataset.documentScrollCalls);
+          dataset.documentScrollCalls = String(calls + 1);
+          Reflect.apply(scrollTo, undefined, args);
+        },
+      });
+    });
+
+    await select.click();
+    await test.expect(dialog).toBeVisible();
+    await test.expect(q.option("main")).toBeInViewport();
+    await flushFrames(page, 3);
+
+    test
+      .expect(
+        await page.evaluate(
+          () => document.documentElement.dataset.documentScrollCalls,
+        ),
+      )
+      .toBe("0");
+  });
+
   for (const label of cases) {
     test(`${label} restores the selected option after switching tabs`, async ({
       page,
--- a/app/src/sandbox/combobox-select-tabs/test-browser.ts
+++ b/app/src/sandbox/combobox-select-tabs/test-browser.ts
@@ -19,6 +19,35 @@ withFramework(import.meta.dirname, async ({ test }) => {
       const style = document.createElement("style");
       style.textContent = "html { scroll-behavior: smooth !important }";
       document.head.append(style);
+      const documentScroller = document.scrollingElement;
+      if (documentScroller) {
+        document.documentElement.dataset.documentNoopScrollWrites = "0";
+        documentScroller.scrollTo = new Proxy(
+          documentScroller.scrollTo.bind(documentScroller),
+          {
+            apply(target, thisArg, args) {
+              const [first, second] = args;
+              const options =
+                typeof first === "number"
+                  ? { left: first, top: second }
+                  : (first ?? {});
+              const left = options.left ?? documentScroller.scrollLeft;
+              const top = options.top ?? documentScroller.scrollTop;
+              if (
+                documentScroller.scrollLeft === left &&
+                documentScroller.scrollTop === top
+              ) {
+                const value =
+                  document.documentElement.dataset.documentNoopScrollWrites ??
+                  "0";
+                document.documentElement.dataset.documentNoopScrollWrites =
+                  String(Number(value) + 1);
+              }
+              return Reflect.apply(target, thisArg, args);
+            },
+          },
+        );
+      }
       window.addEventListener("scroll", () => {
         if (window.scrollY === scrollY) return;
         document.documentElement.dataset.scrollMoved = "true";
@@ -35,6 +64,13 @@ withFramework(import.meta.dirname, async ({ test }) => {
         await page.evaluate(() => document.documentElement.dataset.scrollMoved),
       )
       .toBeUndefined();
+    test
+      .expect(
+        await page.evaluate(
+          () => document.documentElement.dataset.documentNoopScrollWrites,
+        ),
+      )
+      .toBe("0");
 
     await page.mouse.click(1, 1);
     await test.expect(dialog).toHaveCount(0);
@@ -49,6 +85,13 @@ withFramework(import.meta.dirname, async ({ test }) => {
         await page.evaluate(() => document.documentElement.dataset.scrollMoved),
       )
       .toBeUndefined();
+    test
+      .expect(
+        await page.evaluate(
+          () => document.documentElement.dataset.documentNoopScrollWrites,
+        ),
+      )
+      .toBe("0");
   });
 
   for (const label of cases) {
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -359,7 +359,7 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
         // this focus event to propagate so consumers can use the onFocus prop
         // on <Composite>.
         else {
-          baseElement.focus();
+          baseElement.focus({ preventScroll: true });
         }
       };
 
--- a/packages/ariakit-react-components/src/composite/composite.tsx
+++ b/packages/ariakit-react-components/src/composite/composite.tsx
@@ -192,25 +192,25 @@ function useScheduleFocus(store: CompositeStore) {
         return;
       withBaseScrollPreserved(store, () => {
         if (activeElement.hasAttribute("data-autofocus")) {
-          if (focus) {
-            const state = store.getState();
-            const selectMode =
-              "selectElement" in state && !!state.selectElement;
-            const { baseElement } = state;
-            if (selectMode && baseElement) {
-              activeElement.focus({ preventScroll: true });
-              scheduleScroll(baseElement, true);
-            } else {
-              withDocumentScrollPreserved(activeElement, () => {
+          withDocumentScrollPreserved(activeElement, () => {
+            if (focus) {
+              const state = store.getState();
+              const selectMode =
+                "selectElement" in state && !!state.selectElement;
+              const { baseElement } = state;
+              if (selectMode && baseElement) {
+                activeElement.focus({ preventScroll: true });
+                scheduleScroll(baseElement, true);
+              } else {
                 focusIntoView(activeElement);
+              }
+            } else {
+              activeElement.scrollIntoView({
+                block: "nearest",
+                inline: "nearest",
               });
             }
-          } else {
-            activeElement.scrollIntoView({
-              block: "nearest",
-              inline: "nearest",
-            });
-          }
+          });
         } else if (focus) {
           activeElement.focus({ preventScroll: true });
         }
--- a/packages/ariakit-react-components/src/composite/composite.tsx
+++ b/packages/ariakit-react-components/src/composite/composite.tsx
@@ -192,25 +192,25 @@ function useScheduleFocus(store: CompositeStore) {
         return;
       withBaseScrollPreserved(store, () => {
         if (activeElement.hasAttribute("data-autofocus")) {
-          withDocumentScrollPreserved(activeElement, () => {
-            if (focus) {
-              const state = store.getState();
-              const selectMode =
-                "selectElement" in state && !!state.selectElement;
-              const { baseElement } = state;
-              if (selectMode && baseElement) {
-                activeElement.focus({ preventScroll: true });
-                scheduleScroll(baseElement, true);
-              } else {
-                focusIntoView(activeElement);
-              }
+          if (focus) {
+            const state = store.getState();
+            const selectMode =
+              "selectElement" in state && !!state.selectElement;
+            const { baseElement } = state;
+            if (selectMode && baseElement) {
+              activeElement.focus({ preventScroll: true });
+              scheduleScroll(baseElement, true);
             } else {
-              activeElement.scrollIntoView({
-                block: "nearest",
-                inline: "nearest",
+              withDocumentScrollPreserved(activeElement, () => {
+                focusIntoView(activeElement);
               });
             }
-          });
+          } else {
+            activeElement.scrollIntoView({
+              block: "nearest",
+              inline: "nearest",
+            });
+          }
         } else if (focus) {
           activeElement.focus({ preventScroll: true });
         }
--- a/packages/ariakit-react-components/src/composite/utils.test.ts
+++ b/packages/ariakit-react-components/src/composite/utils.test.ts
@@ -1,0 +1,55 @@
+import { afterEach, expect, test, vi } from "vitest";
+import { withDocumentScrollPreserved } from "./utils.ts";
+
+afterEach(() => {
+  document.documentElement.style.removeProperty("scroll-behavior");
+  document.body.style.removeProperty("scroll-behavior");
+  document.body.scrollLeft = 0;
+  document.body.scrollTop = 0;
+  vi.restoreAllMocks();
+});
+
+test("does not write the document scroll position when it does not change", () => {
+  const scrollTo = vi.spyOn(document.documentElement, "scrollTo");
+
+  withDocumentScrollPreserved(document.body, () => {});
+
+  expect(scrollTo).not.toHaveBeenCalled();
+});
+
+test("uses the document root scroll behavior for a body scroller", () => {
+  vi.spyOn(document, "scrollingElement", "get").mockReturnValue(document.body);
+  const scrollTo = vi
+    .spyOn(document.body, "scrollTo")
+    .mockImplementation(() => {});
+
+  document.documentElement.style.setProperty(
+    "scroll-behavior",
+    "smooth",
+    "important",
+  );
+  document.body.scrollTop = 10;
+
+  withDocumentScrollPreserved(document.body, () => {
+    expect(
+      document.documentElement.style.getPropertyValue("scroll-behavior"),
+    ).toBe("auto");
+    expect(
+      document.documentElement.style.getPropertyPriority("scroll-behavior"),
+    ).toBe("important");
+    expect(document.body.style.getPropertyValue("scroll-behavior")).toBe("");
+    document.body.scrollTop = 20;
+  });
+
+  expect(scrollTo).toHaveBeenCalledWith({
+    left: 0,
+    top: 10,
+    behavior: "instant",
+  });
+  expect(
+    document.documentElement.style.getPropertyValue("scroll-behavior"),
+  ).toBe("smooth");
+  expect(
+    document.documentElement.style.getPropertyPriority("scroll-behavior"),
+  ).toBe("important");
+});
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -1,5 +1,5 @@
 import * as Core from "@ariakit/components/composite/composite-store";
-import { getDocument, isTextField } from "@ariakit/utils";
+import { getDocument, getWindow, isTextField } from "@ariakit/utils";
 import type { CompositeStore } from "./composite-store.ts";
 
 export const flipItems = Core.flipItems;
@@ -20,10 +20,41 @@ export function withDocumentScrollPreserved(
   }
   const left = documentScroller.scrollLeft;
   const top = documentScroller.scrollTop;
+  const style = (documentScroller as HTMLElement).style;
+  const scrollBehaviorProperty = "scroll-behavior";
+  const inlineScrollBehavior = style.getPropertyValue(scrollBehaviorProperty);
+  const inlineScrollBehaviorPriority = style.getPropertyPriority(
+    scrollBehaviorProperty,
+  );
+  const smooth =
+    getWindow(documentScroller).getComputedStyle(documentScroller)
+      .scrollBehavior === "smooth";
+  if (smooth) {
+    style.setProperty(scrollBehaviorProperty, "auto", "important");
+  }
   try {
     callback();
   } finally {
-    documentScroller.scrollTo({ left, top, behavior: "instant" });
+    try {
+      if (
+        documentScroller.scrollLeft !== left ||
+        documentScroller.scrollTop !== top
+      ) {
+        documentScroller.scrollTo({ left, top, behavior: "instant" });
+      }
+    } finally {
+      if (smooth) {
+        if (inlineScrollBehavior) {
+          style.setProperty(
+            scrollBehaviorProperty,
+            inlineScrollBehavior,
+            inlineScrollBehaviorPriority,
+          );
+        } else {
+          style.removeProperty(scrollBehaviorProperty);
+        }
+      }
+    }
   }
 }
 
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -20,14 +20,15 @@ export function withDocumentScrollPreserved(
   }
   const left = documentScroller.scrollLeft;
   const top = documentScroller.scrollTop;
-  const style = (documentScroller as HTMLElement).style;
+  const documentElement = element.ownerDocument.documentElement;
+  const style = documentElement.style;
   const scrollBehaviorProperty = "scroll-behavior";
   const inlineScrollBehavior = style.getPropertyValue(scrollBehaviorProperty);
   const inlineScrollBehaviorPriority = style.getPropertyPriority(
     scrollBehaviorProperty,
   );
   const smooth =
-    getWindow(documentScroller).getComputedStyle(documentScroller)
+    getWindow(documentElement).getComputedStyle(documentElement)
       .scrollBehavior === "smooth";
   if (smooth) {
     style.setProperty(scrollBehaviorProperty, "auto", "important");


diegohaz:
Preview: mouse open → outside click → mouse reopen in the `combobox-select-tabs` sandbox. The regression test separately verifies that the document scrolling element receives zero `scrollTo` calls.

[ariakit-combobox-scroll-fix-final.webm](https://github.com/user-attachments/assets/21193c05-58aa-4112-812e-571e0f73609c)

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -1,8 +1,50 @@
...
+      Object.defineProperty(scroller, "scrollTo", {
+        configurable: true,
+        value: (...args: unknown[]) => {
+          const { dataset } = document.documentElement;
+          const calls = Number(dataset.documentScrollCalls);
+          dataset.documentScrollCalls = String(calls + 1);
+          Reflect.apply(scrollTo, undefined, args);
+        },
+      });
+    });
```
diegohaz: Isn't this testing implementation details?

.changeset/300-combobox-select-document-scroll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed a select [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) with a [`Combobox`](https://ariakit.com/reference/combobox) inside writing to the document scroll position when opened, which could briefly reveal overlay scrollbars on macOS.
```
diegohaz: I don't think we need a changeset. The select `ComboboxPopover` isn't published yet.

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -1,8 +1,50 @@
...
+
+    await select.click();
+    await test.expect(dialog).toBeVisible();
+    await test.expect(q.option("main")).toBeInViewport();
+    await flushFrames(page, 3);
```
diegohaz: Read the flushFrames JSDoc

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -192,25 +192,25 @@ function useScheduleFocus(store: CompositeStore) {
...
-                focusIntoView(activeElement);
-              }
+          if (focus) {
+            const state = store.getState();
+            const selectMode =
+              "selectElement" in state && !!state.selectElement;
+            const { baseElement } = state;
+            if (selectMode && baseElement) {
+              activeElement.focus({ preventScroll: true });
+              scheduleScroll(baseElement, true);
```
diegohaz: Dropping the document scroll preservation here makes the page jump to the top when a searchable select is opened with the mouse.

`activeElement.focus({ preventScroll: true })` does not leave the document scroll position alone. Focusing the item synchronously runs `CompositeItem`'s focus handler, which reaches `redirectFocusToBaseElement` and calls `baseElement.focus()` with no `preventScroll` (`packages/ariakit-react-components/src/composite/composite-item.tsx:362`) whenever focus did not arrive from the composite. On a mouse open, `relatedTarget` is the select button, so that branch runs while the freshly mounted popover is still at its pre-placement position at the top of the document, and the browser scrolls the page there. The wrapper removed in this hunk is exactly what used to undo it.

Traced with an instrumented build, the whole jump happens inside the `focus` call on line 201:

```
[scrollprobe] focus input#…[role=combobox] []                              150 -> 0
[scrollprobe] focus div#….ak-option[role=option] [{"preventScroll":true}]  150 -> 0
```

Measured on the unmodified `combobox-select-tabs` sandbox (`Select with Combobox and manual Tab`, no injected styles, no reopen needed) in Chromium at 800x400, `window.scrollY` after a single mouse open:

| | `main` | this PR |
| --- | --- | --- |
| `window.scrollY` | 272 | 0 |

Firefox behaves the same (272 -> 272 on `main`, 272 -> 0 here). With ordinary page content around the select it is not a small-viewport edge case either: 1280x720 gives 456 -> 0, 800x600 gives 516 -> 8, 800x320 gives 656 -> 288, while `main` preserves every one of them. Re-adding the wrapper around only this `selectMode && baseElement` branch restores `main`'s behavior in all of them (5/5 runs per viewport).

Real-world impact: any scrollable page rendering a searchable `ComboboxSelect` (a `ComboboxSelect` plus a `ComboboxPopover` containing a `Combobox`, with `unmountOnHide`) throws the user back to the top of the page every time they click the select. The new regression test cannot see it because it never scrolls the page first, so a restoring `scrollTo({ left: 0, top: 0 })` and "no scroll at all" are indistinguishable to it.

**Assessment**

- Severity: High (verified, 95% confidence), because the page scroll position is lost on an ordinary open, and the popover is left visually detached from its anchor until the next placement pass.
- Priority: P1 (verified, 95% confidence), because it is a regression this PR introduces against `main` on a supported interaction.
- Breakage likelihood: Likely (verified, 90% confidence), because it needs only a scrolled page and a mouse click on a searchable `ComboboxSelect`, reproduces on the first open, and reproduces in both Chromium and Firefox.
- Fix effort: Easy (verified, 90% confidence), because the alternative below is a five-line change in one helper and it removes the need for this hunk entirely.

The stated motivation is that the two writes happen "even though the page coordinates never changed". That is fully addressed by not writing when the coordinates did not change, which keeps the restore for the cases that still need it:

```diff
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -21,7 +21,12 @@
   const left = documentScroller.scrollLeft;
   const top = documentScroller.scrollTop;
   try {
     callback();
   } finally {
-    documentScroller.scrollTo({ left, top, behavior: "instant" });
+    if (
+      documentScroller.scrollLeft !== left ||
+      documentScroller.scrollTop !== top
+    ) {
+      documentScroller.scrollTo({ left, top, behavior: "instant" });
+    }
   }
 }
```

I validated that locally on top of `main`'s `composite.tsx`: the new `reopens the searchable select without writing the document scroll position` test still passes with `0` calls, `window.scrollY` is preserved in all four viewports above, and the 23 Chromium tests across `combobox-select-scroll-into-view`, `combobox-select-offscreen-value`, `combobox-select-animation`, and `combobox-select-tabs` pass. It also lets this whole `composite.tsx` restructuring drop out of the diff, so the conditional nesting introduced here (`if (autofocus) { if (focus) … } else if (focus) …`) goes away with it.

I could not attach the screen recording of the jump: the browser upload path is unavailable in this session, so this comment relies on the measurements above rather than video. To reproduce in a few seconds, serve the app, open `/react/previews/combobox-select-tabs/` at 800x400, scroll `Select with Combobox and manual Tab` into view, click it, and read `window.scrollY`.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -192,25 +192,25 @@ function useScheduleFocus(store: CompositeStore) {
...
+              withDocumentScrollPreserved(activeElement, () => {
+                focusIntoView(activeElement);
               });
             }
-          });
+          } else {
+            activeElement.scrollIntoView({
+              block: "nearest",
+              inline: "nearest",
+            });
```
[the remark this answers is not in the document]
diegohaz: The page-jump risk is real, but I could not reproduce it from this `scrollIntoView`, and the branch above is where it actually happens.

I built the oversized-popover case described here (`combobox-select-tabs` at 800x400 with `max-height: none; overflow: visible` forced on the popover and its descendants, selection moved to the last option so presenting it needs a real scroll) and compared three builds. `main`, this PR, and this PR with `withDocumentScrollPreserved` restored around only this `scrollIntoView` all produced the same result: `window.scrollY` 436 -> 276. So in that scenario the removed wrapper here had no observable effect; something later re-scrolls regardless. I also re-ran the normal fixture with only this branch re-wrapped and the regression was unchanged.

What does reproduce is one branch up, at the `selectMode && baseElement` focus path. On the unmodified sandbox at 800x400, a single mouse open on `Select with Combobox and manual Tab` leaves `window.scrollY` at 272 on `main` and at 0 here, and re-wrapping only that branch restores `main`'s behavior. Details and the cross-browser/cross-viewport numbers are in https://github.com/ariakit/ariakit/pull/6984#discussion_r3687966441.

The remedy suggested at the end of this comment is the right one and I validated it: skipping the restoration write when the coordinates did not change keeps the new zero-write regression test green while preserving the page scroll. I have proposed exactly that patch in the linked thread.

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: It is, and the cost is concrete rather than stylistic: this assertion misses the user-visible regression the change actually introduces.

Two gaps come from measuring the call instead of the outcome. The test never scrolls the page, so `document.scrollingElement` sits at `0, 0` the whole time and a restoring `scrollTo({ left: 0, top: 0 })` is indistinguishable from no scroll at all. And only `scrollTo` is counted, so a jump produced by `focus()`, `scrollIntoView`, or a `scrollTop` write does not register. Both gaps hit at once here: with this branch, opening `Select with Combobox and manual Tab` at 800x400 on a page scrolled to 272 leaves `window.scrollY` at 0 while `main` keeps 272, and this test still passes. Details in https://github.com/ariakit/ariakit/pull/6984#discussion_r3687966441.

The user-facing property is "opening the select does not move the page", which is directly observable and covers strictly more:

```ts
await select.scrollIntoViewIfNeeded();
const scrollY = await page.evaluate(() => window.scrollY);
await select.click();
await test.expect(dialog).toBeVisible();
await test.expect(q.option("main")).toBeInViewport();
await test.expect.poll(() => page.evaluate(() => window.scrollY)).toBe(scrollY);
```

I ran that exact test in this sandbox: it fails on this branch, passes on `main`, and passes alongside the existing zero-write test in Chromium, Firefox, and WebKit once the helper is changed to skip the no-op restoration write. It needs no `Object.defineProperty` patching of `scrollTo` and no longer writes a `data-` attribute onto `<html>` in the page under test. `expect.poll` also replaces the fixed `flushFrames(page, 3)` wait with a retrying assertion, so the check does not depend on the jump landing within a specific number of frames.

If the intent is specifically to assert that no redundant write reaches the document scroller, that is worth keeping as a secondary assertion, but it should sit behind a real scroll-preservation assertion rather than standing in for one.

diegohaz (commented):
## Review summary

Initial review of all three changed files at `46c786c`, diffed from the merge base `f047a0a`, with the surrounding `useScheduleFocus` / `CompositeItem` focus path, the `combobox-select-*` sandboxes, and the existing page-scroll coverage inspected beyond the diff.

### Verification performed

- Confirmed the new regression test is genuine red/green coverage for its own assertion: it reports `2` `scrollTo` calls on `main` and `0` on this branch.
- Reproduced the branch behavior against `main` with a scripted Chromium/Firefox/WebKit harness on the unmodified `combobox-select-tabs` and `combobox-select-scroll-into-view` previews, across 800x320, 800x400, 800x600, 1280x720, and 1280x900.
- Traced the actual scroll write with an instrumented build (patched `Element.prototype.scrollIntoView`, `HTMLElement.prototype.focus`, and the document scroller's `scrollTo` / `scrollTop`).
- Ran `pnpm -F app test --project=chrome` over `combobox-select-scroll-into-view`, `combobox-select-offscreen-value`, `combobox-select-animation`, and `combobox-select-tabs`: 23 passed, matching the PR description. None of them covers the regression below.
- Prototyped and validated the proposed alternative locally, including a candidate replacement test, in Chromium, Firefox, and WebKit. All experiment changes have been reverted.

### Findings published inline

- [`composite.tsx` L200-L202](https://github.com/ariakit/ariakit/pull/6984#discussion_r3687966441): removing the document scroll preservation around the select-mode focus path makes the page jump to the top when a searchable `ComboboxSelect` is opened with the mouse. `activeElement.focus({ preventScroll: true })` synchronously reaches `CompositeItem`'s focus redirect, which calls `baseElement.focus()` without `preventScroll`, and the popover is still at its pre-placement position at that moment. On the unmodified `combobox-select-tabs` preview at 800x400, `window.scrollY` goes 272 -> 272 on `main` and 272 -> 0 here, on the first open, in both Chromium and Firefox. Severity High, Priority P1, Breakage likelihood Likely, Fix effort Easy. The suggested fix is to keep both wrappers as on `main` and instead make `withDocumentScrollPreserved` skip the restoration write when the coordinates did not change; that keeps the new zero-write test green, preserves the page scroll in every viewport measured, and removes the need for the `composite.tsx` restructuring in this diff.

### Existing unresolved threads validated against this head

- [r3687311352](https://github.com/ariakit/ariakit/pull/6984#discussion_r3687311352) (Codex, `composite.tsx` L212, oversized popovers): partly confirmed, replied. The underlying risk is real but the anchored claim is not: I built the oversized, non-internally-scrollable popover case and `main`, this branch, and this branch with the wrapper restored around only that `scrollIntoView` all produced the same `436 -> 276`, and re-wrapping only that branch left the real regression unchanged. Its closing recommendation, skipping the restoration write when the coordinates did not change, is confirmed and adopted in the inline finding above.
- [r3687897633](https://github.com/ariakit/ariakit/pull/6984#discussion_r3687897633) (implementation-detail test): confirmed, replied with evidence. Counting `scrollTo` calls on a page that is never scrolled cannot distinguish a restoring `scrollTo({ left: 0, top: 0 })` from no scroll at all, and it only observes `scrollTo`, so the jump this branch introduces passes the test. A `window.scrollY` assertion after opening on a scrolled page fails here, passes on `main`, and passes with the proposed helper fix across all three desktop engines.
- [r3687899055](https://github.com/ariakit/ariakit/pull/6984#discussion_r3687899055) (changeset not needed): confirmed, no reply needed. `.changeset/200-combobox-select.md` is still pending on `main` and the latest published `@ariakit/react` is `0.4.35`, so `ComboboxSelect` and the select-mode `ComboboxPopover` are unreleased. Every branch this PR touches is reachable only in select mode (`scheduleScroll` returns early when `selectElement` is absent), so there is no consumer-observable change in a published API and `.changeset/300-combobox-select-document-scroll.md` should be dropped.
- [r3687899695](https://github.com/ariakit/ariakit/pull/6984#discussion_r3687899695) (`flushFrames`): left to the author, no separate finding. The replacement test suggested on the test thread uses a retrying `expect.poll` assertion instead, which removes the fixed frame wait entirely.

### Not published as separate findings

- The conditional nesting introduced in `useScheduleFocus` (`if (autofocus) { if (focus) … } else if (focus) …`) reads worse than the branch it replaces, but it is not worth a standalone comment: if the inline finding is addressed as proposed, the whole `composite.tsx` hunk reverts and the nesting disappears with it. Folded into that comment rather than raised on its own.
- The new test lives in `combobox-select-tabs` rather than next to the existing "without scrolling the page" tests in `combobox-select-scroll-into-view`. I checked whether the scenario reproduces in the latter and it does not: its searchable select has no `unmountOnHide`, so the popover is not remounted pre-placement on reopen. The current placement is the correct call.
- No happy-dom duplicate was added. Browser-only coverage is justified here because the behavior depends on real layout, placement, and scrolling.

### Not fully validated

WebKit is inconclusive for the regression itself. My harness could not put the page at a comparable starting scroll position there (`96` versus `272` in the other engines), so I am reporting the regression as confirmed in Chromium and Firefox only. It reproduces on the first mouse open in both.

I was unable to attach the screen recording of the page jump: the browser upload path is unavailable in this session, so the inline finding relies on the measurements and the instrumented trace rather than video. The reproduction steps in that comment take a few seconds to run locally.

packages/ariakit-react-components/src/composite/composite.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e): restored document-scroll preservation around the complete selected-item presentation path. The helper now skips a no-op restoration, but temporarily makes only the document scroller synchronous while the callback runs, so real movement—including queued smooth scrolling—is still detected and restored without changing nested popover scrolling. Added targeted open/outside-click/reopen coverage under `scroll-behavior: smooth !important` in [the same commit](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e).

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: Fixed in [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e): replaced the `scrollTo` interception with user-visible outcome coverage. The test scrolls the page, opens the select with a mouse, verifies the selected option is in the viewport, and asserts that the document never leaves its starting coordinates—including intermediate frames—before repeating the outside-click/reopen sequence under `scroll-behavior: smooth !important`.

.changeset/300-combobox-select-document-scroll.md
diegohaz: Fixed in [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e): removed the changeset because this select-mode `ComboboxPopover` behavior is still unreleased.

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: Fixed in [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e): the regression now uses the documented default `flushFrames(page)` behavior, with no arbitrary frame count. The wait lets a pending smooth scroll manifest before the test checks the observable final coordinates and the intermediate-movement marker.

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Fixed in [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e): restored the preservation wrapper and prevented the underlying redirect at its source with `baseElement.focus({ preventScroll: true })`. The helper conditionally restores only after real movement and handles root smooth scrolling without weakening nested option presentation. The replacement regression reproduces the scrolled mouse-open flow that previously changed `272 → 0`, then covers outside-click/reopen under `smooth !important`.

diegohaz:
**Fixed**

- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) restored document-scroll preservation while making no-op corrections conditional and smooth-root restoration synchronous ([oversized popover thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688161352), [page-jump thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688162538)).
- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) replaced the implementation-detail interception with observable open/outside-click/reopen coverage, including intermediate movement and `smooth !important` ([test thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688161669)).
- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) removed the unnecessary changeset for the unreleased behavior ([changeset thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688161942)).
- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) switched to the documented default `flushFrames(page)` behavior and outcome assertions ([frame-wait thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688162207)).
- [`ff628c648`](https://github.com/ariakit/ariakit/commit/ff628c6485d6a4ac3d0360f396920885e8220222) added a secondary browser assertion for zero equal-coordinate document writes while forwarding necessary restorations ([no-op-write thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688218729)).
- [`ff628c648`](https://github.com/ariakit/ariakit/commit/ff628c6485d6a4ac3d0360f396920885e8220222) moved viewport smooth-behavior handling to `document.documentElement` and added quirks-mode `BODY` scroller coverage ([quirks-mode thread](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688218948)).

**Verified**

- [`ff628c648`](https://github.com/ariakit/ariakit/commit/ff628c6485d6a4ac3d0360f396920885e8220222) passed `pnpm lint`, `pnpm tsc`, `pnpm -F app build-lite`, helper tests 2/2, and the final focused suites: Chromium 3/3, Firefox 3/3, WebKit 4/4.
- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) passed the broader related browser suites: Chromium 32, Firefox 32, WebKit 33.
- [`342520ad7`](https://github.com/ariakit/ariakit/commit/342520ad74d637d046c9dc437d9da84af112fc1e) includes a refreshed [browser recording](https://github.com/user-attachments/assets/e1bc8b5c-e5d3-4b01-852f-404583ea8e9c) in the PR description showing `window.scrollY = 272px` before open, after open, after outside click, and after reopen.

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -1,8 +1,56 @@
...
+    const scrollY = await page.evaluate(() => window.scrollY);
+    test.expect(scrollY).toBeGreaterThan(0);
+    await page.evaluate((scrollY) => {
+      const style = document.createElement("style");
+      style.textContent = "html { scroll-behavior: smooth !important }";
+      document.head.append(style);
+      window.addEventListener("scroll", () => {
+        if (window.scrollY === scrollY) return;
+        document.documentElement.dataset.scrollMoved = "true";
+      });
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`ff628c648`](https://github.com/ariakit/ariakit/commit/ff628c6485d6a4ac3d0360f396920885e8220222): the browser test now proxies `document.scrollingElement.scrollTo`, counts equal-coordinate calls, and asserts zero redundant writes after both open and reopen. The proxy always forwards the call, so the existing user-visible page-position and option-visibility assertions still exercise real behavior. This distinction is necessary on WebKit, which may perform one real restoration (`272 → 96 → 272`) while still producing zero no-op writes.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -20,10 +20,41 @@ export function withDocumentScrollPreserved(
...
+  const scrollBehaviorProperty = "scroll-behavior";
+  const inlineScrollBehavior = style.getPropertyValue(scrollBehaviorProperty);
+  const inlineScrollBehaviorPriority = style.getPropertyPriority(
+    scrollBehaviorProperty,
+  );
+  const smooth =
+    getWindow(documentScroller).getComputedStyle(documentScroller)
+      .scrollBehavior === "smooth";
+  if (smooth) {
+    style.setProperty(scrollBehaviorProperty, "auto", "important");
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`ff628c648`](https://github.com/ariakit/ariakit/commit/ff628c6485d6a4ac3d0360f396920885e8220222): `withDocumentScrollPreserved` now reads and temporarily overrides `document.documentElement` for viewport `scroll-behavior`, while continuing to snapshot and restore coordinates on `document.scrollingElement`. Added a focused helper test that simulates a `BODY` scroller, verifies the root receives `auto !important`, and verifies exact priority restoration. In a real Chromium BackCompat reproduction, `BODY` computed `auto` while `HTML` computed `smooth`; the uncorrected path stayed at `500` immediately and later moved to `1871`, while the root override made the movement synchronous.

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -1,8 +1,99 @@
...
+    }, scrollY);
+
+    await select.click();
+    await test.expect(dialog).toBeVisible();
+    await test.expect(q.option("main")).toBeInViewport();
+    await flushFrames(page);
```
diegohaz, line 60: You didn't read the `flushFrames` JSDocs.

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -1,8 +1,99 @@
...
+  }) => {
+    await page.setViewportSize({ width: 800, height: 400 });
+
+    const select = q.combobox("Select with Combobox and manual Tab");
+    const dialog = q.dialog();
+
+    await select.scrollIntoViewIfNeeded();
+    const scrollY = await page.evaluate(() => window.scrollY);
+    test.expect(scrollY).toBeGreaterThan(0);
+    await page.evaluate((scrollY) => {
```
diegohaz, line 18: This looks like we're testing implementation details. How would a user reproduce this on the sandbox?

packages/ariakit-react-components/src/composite/utils.test.ts
```diff
@@ -0,0 +1,55 @@
+import { afterEach, expect, test, vi } from "vitest";
```
diegohaz, line 1: This test file shouldn't exist. You're over-egineering.

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -359,7 +359,7 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
         // this focus event to propagate so consumers can use the onFocus prop
         // on <Composite>.
         else {
-          baseElement.focus();
+          baseElement.focus({ preventScroll: true });
```
diegohaz, line 362: Why this?

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -20,10 +20,42 @@ export function withDocumentScrollPreserved(
   }
   const left = documentScroller.scrollLeft;
   const top = documentScroller.scrollTop;
+  const documentElement = element.ownerDocument.documentElement;
+  const style = documentElement.style;
+  const scrollBehaviorProperty = "scroll-behavior";
+  const inlineScrollBehavior = style.getPropertyValue(scrollBehaviorProperty);
+  const inlineScrollBehaviorPriority = style.getPropertyPriority(
+    scrollBehaviorProperty,
+  );
```
diegohaz, line 29: This whole helper change no longer has a problem to solve. #6976 merged into `main` at `1e1fa6f97` after this branch's last push, and it removed the select-mode path from `withDocumentScrollPreserved` entirely: `present()` now short-circuits on `canScrollIntoView` before reaching the wrapper, so opening a `ComboboxSelect` never enters this helper.

I measured every `scrollTo` call on `document.scrollingElement` during a mouse open and reopen of `Select with Combobox and manual Tab` at 800x400, on the `combobox-select-tabs` preview:

| base | calls per open | no-op calls |
| --- | --- | --- |
| `f047a0a` (this branch's base) | 2: real `0 → 272`, then no-op `272 → 272` | 1 |
| `1e1fa6f9` (current `main`) | 1: real, non-no-op | **0** |

Same result in Chromium, Firefox, and WebKit. The redundant write this PR set out to remove is already gone.

The decisive check: I applied this branch's `composite-item.tsx` and `utils.ts` on top of current `main` and compared against applying only the `composite-item.tsx` one-liner.

| rebased onto `1e1fa6f9` | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| one-liner only | 272 → 272 | 272 → 272 | 272 → 96 |
| one-liner + this `utils.ts` | 272 → 272 | 272 → 272 | 272 → 96 |

Identical. The 36 lines here and the 55-line `utils.test.ts` change nothing once rebased.

They are not free either. `getComputedStyle` runs on every invocation, and on smooth-scrolling documents the helper writes `scroll-behavior: auto !important` onto `<html>` and then restores it around every callback. That inline write is observable to the page: `MutationObserver`s on the root `style` attribute fire twice per pass, anything reading `documentElement.style.scrollBehavior` during the callback sees the override, and a callback that sets that property on `<html>` itself gets clobbered by the restore. The nested `try`/`finally` exists only to make that mutation safe.

**Assessment**

- Severity: Low (verified, 90% confidence), because the code is inert on current `main` rather than incorrect; the cost is the maintenance surface and the root-style mutation, not a user-facing defect.
- Priority: P1 (estimate, 85% confidence), because shipping a shared helper plus a mock-based unit test for a case that no longer occurs is the kind of thing that is much cheaper to drop now than to remove later.
- Breakage likelihood: Unlikely (estimate, 75% confidence) for the root-style mutation itself; it needs a page that sets `scroll-behavior: smooth` and observes or writes root inline styles.
- Fix effort: Easy (verified, 95% confidence), because the change is to delete this hunk and `utils.test.ts`.

Suggested resolution: rebase onto `1e1fa6f9`, revert `utils.ts` and delete `utils.test.ts`, and keep only the `composite-item.tsx` one-liner plus one user-visible regression test. Full reasoning and the remaining WebKit gap are in the review summary.

packages/ariakit-react-components/src/composite/composite-item.tsx
diegohaz: This line is the root cause of the page jump, and after #6976 it is the only part of this PR that still fixes a live bug.

When an item receives focus from outside the composite, `redirectFocusToBaseElement` lands in this `else` branch and focuses the base element. Without `preventScroll` the browser scrolls that element into view, and on a mouse open the popover has not been placed yet, so it is still at the top of the document and the page scrolls there. The comment directly above says this branch exists so the focus event propagates for the `onFocus` prop, not to scroll anything, and the sibling branch already uses `focusSilently`, which passes `preventScroll: true`. So this makes the two branches consistent with the stated intent.

Measured on `combobox-select-tabs` at 800x400, mouse open of `Select with Combobox and manual Tab` with the page at `window.scrollY` 272:

| base | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| `f047a0a` (pre-#6976) | 272 → 272 | 272 → 272 | 272 → 272 |
| `1e1fa6f9` (current `main`) | 272 → **0** | 272 → **0** | 272 → **96** |
| current `main` + this line | 272 → 272 | 272 → 272 | 272 → **96** |

So #6976 removed the `withDocumentScrollPreserved` wrapper that had been masking this, and the regression is live on `main` today. This line fixes Chromium and Firefox. WebKit still moves from a separate cause that this line does not reach, and I have not diagnosed it.

I ran `src/sandbox/{combobox,select,menu,composite,tab,toolbar}*` across Chromium, Firefox, and WebKit with only this line applied to current `main`: 984 passed, 1 flaky, and the only failures were the `*-nextjs-app-router` files, which need the real Next.js server my local run did not have.

Two consequences worth deciding on:

1. This changes released behavior. `redirectFocusToBaseElement` runs for every `virtualFocus` composite, so `Combobox`, `Select`, and `Menu` all stop scrolling their base element into view on this path. That is not the unreleased select-mode `ComboboxPopover` scope that justified dropping the changeset earlier, so a patch changeset for `@ariakit/react-components` and `@ariakit/react` is needed again. There is no Solid equivalent of this branch, so Solid is unaffected.
2. Existing coverage does not protect it. Current `main` passes `combobox-select-scroll-into-view`, `combobox-center-into-view`, `combobox-select-offscreen-value`, and `combobox-select-tabs` on all three engines (80 passed) while the regression is present, so this PR should keep one user-visible page-scroll regression test.

packages/ariakit-react-components/src/composite/utils.test.ts
diegohaz: Agreed, and it should go along with the `utils.ts` change it covers.

Three separate reasons, in increasing order of weight:

1. It duplicates a contract the browser test already asserts. The sandbox test forces `scroll-behavior: smooth !important` and checks the same smooth-scrolling path end to end, so the same behavior is pinned at two layers.
2. It mocks the behavior under test. `vi.spyOn(document, "scrollingElement", "get").mockReturnValue(document.body)` fabricates a quirks-mode scroller, `vi.spyOn(document.body, "scrollTo").mockImplementation(() => {})` replaces the operation being verified, and the assertions then check the inline `scroll-behavior` value and priority *inside* the callback. Nothing here is observable behavior; it is a transcript of the implementation.
3. It is testing code that no longer runs. On current `main` (`1e1fa6f9`), #6976 short-circuits the select-mode path before `withDocumentScrollPreserved`, so the flow this PR is about never enters the helper. I applied this branch's `composite-item.tsx` plus `utils.ts` on top of `1e1fa6f9` and compared with the one-liner alone: identical results on all three engines (Chromium and Firefox 272 → 272, WebKit 272 → 96). The helper change buys nothing after a rebase, so its test has nothing to protect.

Deleting this file and reverting `utils.ts` removes 91 of the 182 added lines and leaves the part that still fixes a live bug.

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: Confirmed, and the answer to "how would a user reproduce this on the sandbox" is that they cannot: no-op `scrollTo` calls have no rendered consequence a test can observe, which is why this block has to reach for a `Proxy` and `data-` attributes on `<html>`.

A user *can* reproduce the part that matters. Open the `combobox-select-tabs` preview at 800x400, scroll `Select with Combobox and manual Tab` into view, click it, and watch the page jump to the top on current `main`. That is one observable property and it needs no instrumentation:

```ts
await select.scrollIntoViewIfNeeded();
const scrollY = await page.evaluate(() => window.scrollY);
test.expect(scrollY).toBeGreaterThan(0);
await select.click();
await test.expect(dialog).toBeVisible();
await test.expect(q.option("main")).toBeInViewport();
await test.expect.poll(() => page.evaluate(() => window.scrollY)).toBe(scrollY);
```

That covers the regression, drops the `Proxy` and the two `data-` attributes written onto the page under test, and removes the duplicated assertion blocks: the open and reopen halves of this test differ only by the outside click between them, so the shared part belongs in one local helper called twice. It also drops the `scroll-behavior: smooth` scenario, and with it the `flushFrames` wait; if that scenario is kept, see the `flushFrames` thread for why the wait has to stay and be justified rather than be replaced by a retrying assertion.

What it gives up is the no-op-write assertion. On current `main` that is no longer a real loss: after #6976, opening a `ComboboxSelect` never enters `withDocumentScrollPreserved` at all, and I measured zero no-op `scrollTo` calls on open and reopen in Chromium, Firefox, and WebKit. Details are in https://github.com/ariakit/ariakit/pull/6984#discussion_r3688329012.

One caveat on the replacement: on current `main` the WebKit case settles at 96 rather than 272 for a reason neither this branch nor the `composite-item.tsx` one-liner addresses, so the assertion above passes on Chromium and Firefox but not yet on WebKit.

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: Confirmed. The JSDoc that #6976 added to `flushFrames` in `app/src/test-utils/preview.ts` on `1e1fa6f9` reads:

> Use sparingly. Prefer a retrying assertion against existing observable
> state. When no such state exists, add a comment explaining why waiting for
> frames is necessary.

Both halves are violated here. Observable state exists (`window.scrollY`), so a retrying assertion is available and should be used instead; and the call carries no comment explaining why frames are necessary.

The earlier reply on this thread claimed the wait "lets a pending smooth scroll manifest before the test checks the observable final coordinates". It does not. `flushFrames(page)` waits two animation frames; a smooth scroll animates over far more than that, so the assertion can sample a pre-drift value. I measured this directly by removing the smooth handling from the helper and opening the same select under `scroll-behavior: smooth !important`:

| engine | `window.scrollY` after `flushFrames(page)` | after settling |
| --- | --- | --- |
| Chromium | 272 | 272 |
| Firefox | 272 | 272 |
| WebKit | 272 | **96** |

The WebKit row is the failure mode: two frames after the open the page still reads its original position, and the assertion at line 61 would pass while the page goes on to move 176px. The `dataset.scrollMoved` listener has the same problem, since scroll events are delivered at the next rendering opportunity rather than synchronously.

For the regression this PR needs to cover, the jump is immediate, and `await test.expect.poll(() => page.evaluate(() => window.scrollY)).toBe(scrollY)` is the right tool: it keeps sampling until the value matches or the timeout expires, so it fails on a jump regardless of which frame the jump lands on, and it needs no explanatory comment because it is not waiting on frames at all. I verified that version fails on the broken state and passes on the fixed one.

To be precise about its limit, since it is the opposite of the case above: a retrying assertion passes as soon as the value matches, so it proves the page did not jump but it cannot prove the page will not drift later. Ruling out a *pending* smooth scroll genuinely needs a settle condition, and that is the case the JSDoc carves out. If this test keeps the `scroll-behavior: smooth` scenario, the frame wait should stay and carry a comment saying it is waiting out a queued smooth scroll, and the frame count should be justified rather than left at the default that this measurement shows is too small.

app/src/sandbox/combobox-select-tabs/test-browser.ts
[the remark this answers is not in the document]
diegohaz: The mechanical point is right: a `scroll` listener cannot observe a no-op `scrollTo`, because the coordinates never change and no event is dispatched. Instrumenting `scrollTo` is the only way to assert that property from a browser test.

It is obsolete now, though. #6976 merged into `main` at `1e1fa6f9` after this branch's last push and removed the select-mode path from `withDocumentScrollPreserved`, so opening a `ComboboxSelect` no longer enters the helper at all. Measuring every `scrollTo` call on `document.scrollingElement` during a mouse open and reopen of `Select with Combobox and manual Tab`:

| base | calls per open | no-op calls |
| --- | --- | --- |
| `f047a0a` (this branch's base) | 2: real `0 → 272`, then no-op `272 → 272` | 1 |
| `1e1fa6f9` (current `main`) | 1: real, non-no-op | 0 |

Same in Chromium, Firefox, and WebKit. So after a rebase there is no unwanted write left for this instrumentation to guard, and the recommendation here would be pinning a property that the code no longer produces. Details in https://github.com/ariakit/ariakit/pull/6984#discussion_r3688329012.

packages/ariakit-react-components/src/composite/utils.ts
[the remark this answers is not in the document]
diegohaz: The CSS point is correct. Per CSSOM View, the viewport's used `scroll-behavior` comes from the root element and the value on the HTML `body` is not propagated, so a quirks-mode document with `body` as `document.scrollingElement` would have been read wrong by the earlier version.

The direction it pushed the fix in is the problem, not the fact. Following it added a temporary `scroll-behavior: auto !important` write onto `<html>` around every callback, a nested `try`/`finally` to unwind it, and a 55-line mock-based unit test for the unwinding, in a shared helper.

That machinery has no effect after a rebase. #6976 merged into `main` at `1e1fa6f9` and short-circuits the select-mode path before `withDocumentScrollPreserved`, so this flow no longer reaches the helper. I applied this branch's `composite-item.tsx` and `utils.ts` on top of `1e1fa6f9` and compared with the `composite-item.tsx` one-liner alone: Chromium 272 → 272, Firefox 272 → 272, WebKit 272 → 96 in both cases, identical.

If the no-op-write optimization is ever revisited outside this PR, the cheap form of this correction is to read the root element's computed `scroll-behavior` and restore unconditionally when it is `smooth`, instead of overriding it. That is about five lines, mutates nothing, and matches today's behavior exactly on smooth-scrolling documents. I measured it: it holds the page at 272 in all three engines under `scroll-behavior: smooth !important`, where dropping the smooth handling entirely lets WebKit drift to 96. It trades away zero-no-op-writes on smooth-scrolling pages only. Full context in https://github.com/ariakit/ariakit/pull/6984#discussion_r3688329012.

diegohaz (commented):
## Re-review summary

Re-review of `ff628c648`. Incremental baseline is `46c786ca0`, the head recorded by the previous marked review, so the reviewed range is `342520ad7..ff628c648` plus the surrounding behavior those commits expose. This run also re-validated every relevant unresolved thread against the current head.

The headline is not in the diff: **`main` moved.** #6976 merged at `1e1fa6f9` after this branch's last push, and it changes both halves of this PR's premise.

### The original problem is gone; a bigger one took its place

#6976 rewrote `useScheduleFocus` and removed the select-mode path from `withDocumentScrollPreserved` (`present()` now short-circuits on `canScrollIntoView`). Measuring every `scrollTo` call on `document.scrollingElement` during a mouse open and reopen of `Select with Combobox and manual Tab` on the `combobox-select-tabs` preview at 800x400:

| base | `scrollTo` calls per open | no-op calls | `window.scrollY` |
| --- | --- | --- | --- |
| `f047a0a` (this branch's base) | 2: real `0 → 272`, then no-op `272 → 272` | 1 | 272 → 272 |
| `1e1fa6f9` (current `main`) | 1: real, non-no-op | **0** | 272 → **0** |

Identical in Chromium, Firefox, and WebKit, except that WebKit lands on 96 rather than 0.

So the redundant write this PR was opened for no longer happens, and the wrapper that had been masking `CompositeItem`'s unqualified `baseElement.focus()` is gone, which means **opening a searchable `ComboboxSelect` on a scrolled page moves the page on `main` today**. Existing coverage does not see it: `combobox-select-scroll-into-view`, `combobox-center-into-view`, `combobox-select-offscreen-value`, and `combobox-select-tabs` all pass on current `main` across the three desktop engines (80 passed, 1 flaky).

### The rest of the branch stops earning its keep after a rebase

Applying this branch's library changes on top of `1e1fa6f9` and comparing against the `composite-item.tsx` one-liner alone:

| rebased onto `1e1fa6f9` | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| `composite-item.tsx` one-liner only | 272 → 272 | 272 → 272 | 272 → 96 |
| one-liner + this branch's `utils.ts` | 272 → 272 | 272 → 272 | 272 → 96 |

Identical. The 36 added lines in `utils.ts` and the 55-line `utils.test.ts` change nothing once rebased, while adding a `getComputedStyle` call per invocation and a temporary `scroll-behavior: auto !important` write onto `<html>` around every callback.

<details>
<summary>2026-07-31, cycle 3: Collapse to the root-cause fix and rebase onto 1e1fa6f9</summary>

**Assessment**

- Issue severity. What the branch set out to fix was a cosmetic macOS overlay-scrollbar flash from a redundant `scrollTo`. What it now needs to fix is a page-position jump on open, which is a functional defect.
- Expected frequency. The cosmetic case no longer occurs in this flow at all after #6976 (0 no-op writes measured). The jump occurs on every mouse open of a searchable `ComboboxSelect` on a scrolled page, in all three desktop engines.
- Supported scope. `redirectFocusToBaseElement` runs for every `virtualFocus` composite, so the fix reaches released `Combobox`, `Select`, and `Menu`, not only the unreleased select-mode `ComboboxPopover`.
- Likely user impact. High for the jump, none for the redundant write on current `main`.
- Implementation and maintenance complexity. The branch is now +182/−4 across four files: a 36-line rewrite of a shared helper with a nested `try`/`finally` and a temporary root inline-style mutation, a 55-line unit test built on `vi.spyOn` mocks of `document.scrollingElement` and `Element.scrollTo` that asserts inline-style state mid-callback, and a 93-line browser test that proxies `scrollTo` and writes two `data-` attributes onto the page under test. Three of the four files exist to serve a motivation that no longer applies.

**Decision**

Rebase onto `1e1fa6f9` and collapse the branch to:

1. `composite-item.tsx`: `baseElement.focus({ preventScroll: true })`. Verified to fix Chromium and Firefox on current `main`; `src/sandbox/{combobox,select,menu,composite,tab,toolbar}*` across all three engines gives 984 passed, 1 flaky, with only `*-nextjs-app-router` failures from the missing local Next.js server.
2. One user-visible browser regression test asserting the page does not move on open and reopen, with no `scrollTo` instrumentation and no `data-` attributes on `<html>`, and the duplicated open/reopen block extracted into one local helper.
3. A patch changeset for `@ariakit/react-components` and `@ariakit/react`. The earlier decision to drop it was correct for a select-mode-only change and no longer holds now that the fix is in `CompositeItem`. No Solid equivalent exists, so Solid is unaffected.
4. Revert `utils.ts` and delete `utils.test.ts`.

Intentionally unsupported for now: eliminating the redundant `scrollTo` write on the paths that still use `withDocumentScrollPreserved` (non-select autofocus composites, and the Safari branch in `composite-item.tsx`). If that is ever worth revisiting, the cheap form is to read the root element's computed `scroll-behavior` and restore unconditionally when it is `smooth`, rather than overriding it; I measured that at about five lines with no style mutation, holding 272 in all three engines under `scroll-behavior: smooth !important`, where dropping smooth handling entirely lets WebKit drift to 96.

</details>

This entry belongs in the pull request description when the collapsed version is pushed; I have not edited the description as part of this review.

### Open question the collapse does not answer

On current `main`, WebKit settles at 96 instead of 272 with or without this branch's helper changes, and the `composite-item.tsx` one-liner does not reach it. Pre-#6976 `main` held 272 on WebKit, so this is also a #6976 side effect. I did not diagnose the cause. It needs a decision before the regression test can be written for all three engines: either fix it in the collapsed PR, or scope the test and register a follow-up issue. I have not created an issue, since that is a repository write beyond a review.

- [r3688241672](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688241672) (`composite-item.tsx` "Why this?"): answered. This is the root cause and, after #6976, the only part of the branch that still fixes a live bug. Raised the changeset consequence there.
- [r3688238165](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688238165) (`utils.test.ts` should not exist): confirmed, replied. Duplicates the browser test's contract, mocks the behavior under test, and covers code that no longer runs after a rebase.
- [r3688233724](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688233724) (implementation details): confirmed, replied. A user cannot reproduce a no-op `scrollTo`; they can reproduce the page jump. Gave the instrumentation-free replacement.
- [r3688232029](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688232029) (`flushFrames` JSDoc): confirmed, replied. #6976's JSDoc asks for a retrying assertion against observable state, or a comment justifying the wait; this call has neither. Measured that two frames do not let a smooth scroll manifest: WebKit read 272 after `flushFrames(page)` and settled at 96.
- [r3688164627](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688164627) (Codex, assert writes not events): mechanically correct, now obsolete, replied. There is no unwanted write left in this flow to guard.
- [r3688164629](https://github.com/ariakit/ariakit/pull/6984#discussion_r3688164629) (Codex, quirks-mode viewport element): the CSS fact is correct, replied. The machinery it prompted has no effect after a rebase, and the cheap form of the correction is noted for any future revisit.

Findings from the previous review are all resolved on threads by `342520ad7`; I did not re-adjudicate them.

### Verification performed

Reproductions on `f047a0a`, `1e1fa6f9`, `1e1fa6f9` + the one-liner, and `1e1fa6f9` + the full branch, each in Chromium, Firefox, and WebKit, measuring `window.scrollY` and every `document.scrollingElement.scrollTo` call with its coordinates. `pnpm -F app test` for `combobox-select-tabs` and the `combobox-select-*` scroll suites at this head and under three helper variants; the broad `src/sandbox/{combobox,select,menu,composite,tab,toolbar}*` sweep across all three engines with the one-liner. All experiment patches have been reverted.

### Not fully validated

The WebKit residual above. The `*-nextjs-app-router` failures in my sweeps are environmental (no local Next.js server) and were excluded by inspection rather than by re-running against a real one.

diegohaz:
Closing in favor of #6985.

The premise of this branch changed after #6976 merged. The redundant `scrollTo`
write it was opened to remove no longer happens: measuring every
`document.scrollingElement.scrollTo` call during an open and reopen of
`Select with Combobox and manual Tab`, this branch's base `f047a0a` issues two
(one real, one no-op) while current `main` issues one, with zero no-op writes.
So the `utils.ts` rewrite and its unit test no longer change anything, which
matches the objections already raised on those threads.

What #6976 did expose is a worse problem that this branch's
`composite-item.tsx` one-liner correctly identified: `withDocumentScrollPreserved`
had been masking two unqualified scroll requests in `redirectFocusToBaseElement`,
and opening a searchable `ComboboxSelect` on a scrolled page now moves the page
(300 → 0 in Chromium and Firefox, 300 → 96 in WebKit, at 1280x720 on the
`combobox-select-tabs` preview).

#6985 keeps the root-cause fix and drops the rest. It changes only
`composite-item.tsx`: `baseElement.focus({ preventScroll: true })`, plus always
wrapping the Safari `scrollIntoView` fallback in `withDocumentScrollPreserved`,
which is what fixes the WebKit half and which this branch did not reach. It adds
one browser regression test with no `Proxy` instrumentation and no `data-*`
attributes on the page under test, verified red on all three desktop engines
before the fix.

PR #6985, opened 2026-07-31
Fix the page scrolling when a ComboboxSelect opens

diegohaz:
## Motivation

Opening a searchable `ComboboxSelect` on a scrolled page moves the page on `main` today. Scroll the picker into view and click it: the page jumps to the top in Chromium and Firefox, and part of the way up in WebKit.

The popup is still at its pre-placement origin when focus enters it. Floating UI's `computePosition` is async, so the popup sits at document `0,0` until it resolves. Instrumenting the open sequence with the page scrolled to 628, reading the popup's `data-placing` attribute and its viewport rect at every `focus()` call:

```
focus  role=option    placing=true   popoverTop=-628   composite.tsx      (useScheduleFocus)
focus  role=combobox  placing=true   popoverTop=-628   composite-item.tsx (redirectFocusToBaseElement)
focus  role=combobox  placing=null   popoverTop=228    dialog.tsx         (autoFocusOnShow, after positioning)
```

`useScheduleFocus` gives the selected item DOM focus while `popoverTop` is negative, and `CompositeItem` synchronously redirects that focus to the composite element, which is the `Combobox` rendered inside the same unplaced popup. The browser scrolls the document to where that element currently is, which is the top of the page.

This is a regression from #6976. Before it, that focus call ran inside `withDocumentScrollPreserved`:

```ts
withDocumentScrollPreserved(activeElement, () => {
  activeElement.focus({ preventScroll: true });
  scheduleScroll(baseElement, true);
});
```

#6976 lifted the call out of the wrapper while restructuring the branch, so the redirect's document scroll became visible.

## Solution

Put the focus call back inside `withDocumentScrollPreserved`, on the `unstable_scrollIntoView` path only.

```ts
withDocumentScrollPreserved(activeElement, focusActiveItem);
```

The consumers on that path already own presentation scrolling: both route to `scrollSelectedItemIntoView`, whose `centerIntoView` snapshots and restores every scroll ancestor of the popup, including the document scroller. So the document should never move on the composite's behalf here. Nested scrollers still move, which is what centers the selected item, and the later scheduled scroll runs on its own pass and is unaffected.

Measured on the `combobox-select-open-page-scroll` preview at 1280x720, with the page scrolled to 628 before opening:

| | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| `main` | 628 &rarr; 243 | 628 &rarr; 243 | 628 &rarr; 127 |
| this branch | 628 &rarr; 628 | 628 &rarr; 628 | 628 &rarr; 628 |

The restore is synchronous, in the same task as the focus, so nothing is painted at the intermediate position.

### Why not fix it in `CompositeItem`

The first version of this branch did, by passing `preventScroll: true` to `baseElement.focus()` and wrapping the Safari `data-autofocus` fallback. Three problems, all measured:

- `preventScroll` alone, keeping `main`'s `if (fromComposite)` split, leaves WebKit at 628 &rarr; 127, because the Safari fallback runs first and moves the document by itself.
- Wrapping that fallback unconditionally fixes all three engines but stops presenting an offscreen inline composite whose item has `autoFocus` on WebKit, where the manual scroll is the only thing that presents it. `main` scrolls to 1372 there; the wrapped version left the page at 0.
- Scoping the wrapper with `fromComposite || closest("[data-placing]")` fixes both, but it changes released `CompositeItem` behavior for a bug that is not reachable in a released version, and it makes `composite/**` read a popover-private attribute.

Fixing the unreleased regression where it was introduced avoids all of that. `composite-item.tsx` is untouched on this branch.

## Preview

Both recordings open `Select with Combobox and manual Tab` on the `combobox-select-tabs` preview at 800x400 with the page scrolled down, in Chromium. They show the user-visible behavior, which is the same for every version of the fix on this branch.

Before: the page jumps to the top as the popup opens.

[before-broken.webm](https://github.com/user-attachments/assets/8da47b17-3b66-41cc-bcdc-3ea5597f9686)

After: the page stays where it was and the popup opens under the trigger.

[after-fixed.webm](https://github.com/user-attachments/assets/57e8437b-ad95-4f92-b088-1d04ca8b930f)

## Tests

`app/src/sandbox/combobox-select-open-page-scroll` is a dedicated sandbox. The first version of this branch put the test in `combobox-select-tabs`, but that fixture owns its props for an unrelated tab-restoration scenario, so an edit there could have left the test green while it stopped reproducing the bug.

The fixture reproduces the jump only when three things hold together, and each is pinned by an assertion that fails if it is removed:

- `virtualFocus`, without which the composite never redirects focus at all.
- The `Combobox` rendered inside the popup rather than next to the select. That is what leaves the popup's initial focus unresolved, so an item is focused first.
- `unmountOnHide`, so the composite element only mounts as the popup opens. Without it, the only thing that can schedule the focus is the dialog's autofocus, which `usePopover` already gates on `positioned`, so the bug does not reproduce.

Red checks, each confirmed in Chromium, Firefox and WebKit:

| Change | Result |
| --- | --- |
| `composite.tsx` reverted to `main` | test fails in all three |
| `unmountOnHide` removed from the fixture | test fails in all three |
| `Combobox` moved outside the popup | test fails in all three |

The final scroll read is deliberately non-retrying, so a correction that lands after paint would not be accepted.

## No changeset

The page jump is not reachable in a released version. The whole `unstable_scrollIntoView` branch was added by #6976, which is unreleased, and its only library consumers are gated on `ComboboxSelect`, which has not shipped either. Nothing this branch changes is observable through a released API.

That was also checked against the real published package rather than only by reading it. A standalone app on `@ariakit/react@0.4.35` running the documented `SelectPopover` plus `Combobox` pattern, where `SelectItem` marks the selected item with no consumer opt-in, was measured across Chromium, Firefox and WebKit, two viewports, with and without `unmountOnHide`, four scroll positions, and both the first open and a reopen. The page did not move in any of the 96 measurements, and the popup was already positioned when the item received focus.

## Follow-up

The ordering behind this bug is worth revisiting on its own, tracked in #6986. `Popover` already waits for placement before autofocusing, but the composite's scheduled focus does not, and neither does `CompositeFocusOnMove`. Gating `useScheduleFocus` on `data-placing` was prototyped here and does fix the jump in all three engines with the suite green, but `data-placing` is also permanently set on closed-but-mounted popovers, so the predicate cannot tell "opening" from "closed", and the gate turns a synchronous focus move into an unbounded wait with no cancellation. That belongs in a change that can take the whole composite focus presentation into account, for both real and virtual focus.

## Review gate reassessment

<details>
<summary>Round 3: Relocate the fix to the commit that regressed it</summary>

**Assessment**

Three consecutive pre-commit gate rounds produced findings on the regression test for the `CompositeItem` direction. Every substantive finding was about the test's ability to fail rather than about the library change, which is what a negative assertion costs: "the page did not move" is only meaningful while the fixture still reproduces the movement.

The reassessment surfaced a direction that had not been measured. The PR's own description already said the calls were previously masked by `withDocumentScrollPreserved` in `useScheduleFocus`; `git show 1e1fa6f97` confirms #6976 lifted the focus call out of that wrapper. Restoring it is smaller than every alternative, touches only unreleased code, cannot regress the inline WebKit case because `composite-item.tsx` is not modified, and makes the changeset decision unambiguous instead of resting on 96 negative measurements against the published package.

**Decision**

Relocate. `composite-item.tsx` reverted to `main`, the fix moved to `composite.tsx`, the sandbox narrowed to the popup case, and the changeset removed. The inline-composite coverage added for the previous direction was dropped with it, since nothing in this branch can affect that path. The wider focus-ordering redesign is registered as #6986.

</details>

### DIFF
--- a/.changeset/300-composite-item-focus-scroll.md
+++ b/.changeset/300-composite-item-focus-scroll.md
@@ -1,6 +1,0 @@
----
-"@ariakit/react-components": patch
-"@ariakit/react": patch
----
-
-Fixed [`CompositeItem`](https://ariakit.com/reference/composite-item), and components based on it such as [`ComboboxItem`](https://ariakit.com/reference/combobox-item) and [`SelectItem`](https://ariakit.com/reference/select-item), scrolling the page when it receives focus inside a [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) composite widget rendered in a popup that hasn't been positioned yet.
--- a/.changeset/300-composite-item-focus-scroll.md
+++ b/.changeset/300-composite-item-focus-scroll.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`CompositeItem`](https://ariakit.com/reference/composite-item), and components based on it such as [`ComboboxItem`](https://ariakit.com/reference/combobox-item) and [`SelectItem`](https://ariakit.com/reference/select-item), scrolling the page when it receives focus inside a [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) composite widget rendered in a popup that hasn't been positioned yet.
--- a/app/src/sandbox/combobox-select-tabs/test-browser.ts
+++ b/app/src/sandbox/combobox-select-tabs/test-browser.ts
@@ -39,4 +39,22 @@ withFramework(import.meta.dirname, async ({ test }) => {
       await test.expect(lastOption).toBeInViewport();
     });
   }
+
+  // https://github.com/ariakit/ariakit/pull/6976
+  test("opens the searchable select with an autofocused item without scrolling the page", async ({
+    page,
+    q,
+  }) => {
+    await page.evaluate(() => {
+      document.body.style.paddingTop = "400px";
+      window.scrollTo({ top: 300 });
+    });
+    await q.combobox("Select with Combobox and manual Tab").click();
+    await test.expect(q.combobox("Find a branch or tag")).toBeFocused();
+    await test.expect(q.option("main")).toBeInViewport();
+    // Read the scroll position once instead of retrying. A retrying assertion
+    // would also pass on an implementation that scrolls the page and then
+    // scrolls it back.
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(300);
+  });
 });
--- a/app/src/sandbox/combobox-select-tabs/test-browser.ts
+++ b/app/src/sandbox/combobox-select-tabs/test-browser.ts
@@ -39,22 +39,4 @@ withFramework(import.meta.dirname, async ({ test }) => {
       await test.expect(lastOption).toBeInViewport();
     });
   }
-
-  // https://github.com/ariakit/ariakit/pull/6976
-  test("opens the searchable select with an autofocused item without scrolling the page", async ({
-    page,
-    q,
-  }) => {
-    await page.evaluate(() => {
-      document.body.style.paddingTop = "400px";
-      window.scrollTo({ top: 300 });
-    });
-    await q.combobox("Select with Combobox and manual Tab").click();
-    await test.expect(q.combobox("Find a branch or tag")).toBeFocused();
-    await test.expect(q.option("main")).toBeInViewport();
-    // Read the scroll position once instead of retrying. A retrying assertion
-    // would also pass on an implementation that scrolls the page and then
-    // scrolls it back.
-    test.expect(await page.evaluate(() => window.scrollY)).toBe(300);
-  });
 });
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -324,24 +324,28 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
           }
           return;
         }
-        // The composite may live in a popup that hasn't been positioned yet, so
-        // neither this fallback nor the focus redirect below may scroll the
-        // document, otherwise the page would jump.
-        //
+        const fromComposite =
+          relatedTarget === baseElement || isItem(store, relatedTarget);
         // Safari doesn't scroll the element into view when another element is
-        // immediately focused. So we have to do it manually here. Restoring the
-        // document scroll afterwards leaves only nested scrollers moved.
+        // immediately focused. So we have to do it manually here.
         if (isSafari() && currentTarget.hasAttribute("data-autofocus")) {
-          withDocumentScrollPreserved(currentTarget, () => {
+          // Virtual focus immediately returns to an already focused composite.
+          // Keep the fallback scoped to nested scrollers in that case.
+          if (fromComposite) {
+            withDocumentScrollPreserved(currentTarget, () => {
+              currentTarget.scrollIntoView({
+                block: "nearest",
+                inline: "nearest",
+              });
+            });
+          } else {
             currentTarget.scrollIntoView({
               block: "nearest",
               inline: "nearest",
             });
-          });
+          }
         }
         hasFocusedComposite.current = true;
-        const fromComposite =
-          relatedTarget === baseElement || isItem(store, relatedTarget);
         // If the previously focused element is a composite or composite item
         // component, we'll transfer focus silently to the composite element.
         // That's because this is just a transition event, the composite
@@ -353,10 +357,9 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
 
         // Otherwise, the composite element is likely not focused, so we need
         // this focus event to propagate so consumers can use the onFocus prop
-        // on <Composite>. Propagation is controlled by the silent focus flag,
-        // not by the scroll option.
+        // on <Composite>.
         else {
-          baseElement.focus({ preventScroll: true });
+          baseElement.focus();
         }
       };
 
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -324,28 +324,24 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
           }
           return;
         }
-        const fromComposite =
-          relatedTarget === baseElement || isItem(store, relatedTarget);
+        // The composite may live in a popup that hasn't been positioned yet, so
+        // neither this fallback nor the focus redirect below may scroll the
+        // document, otherwise the page would jump.
+        //
         // Safari doesn't scroll the element into view when another element is
-        // immediately focused. So we have to do it manually here.
+        // immediately focused. So we have to do it manually here. Restoring the
+        // document scroll afterwards leaves only nested scrollers moved.
         if (isSafari() && currentTarget.hasAttribute("data-autofocus")) {
-          // Virtual focus immediately returns to an already focused composite.
-          // Keep the fallback scoped to nested scrollers in that case.
-          if (fromComposite) {
-            withDocumentScrollPreserved(currentTarget, () => {
-              currentTarget.scrollIntoView({
-                block: "nearest",
-                inline: "nearest",
-              });
-            });
-          } else {
+          withDocumentScrollPreserved(currentTarget, () => {
             currentTarget.scrollIntoView({
               block: "nearest",
               inline: "nearest",
             });
-          }
+          });
         }
         hasFocusedComposite.current = true;
+        const fromComposite =
+          relatedTarget === baseElement || isItem(store, relatedTarget);
         // If the previously focused element is a composite or composite item
         // component, we'll transfer focus silently to the composite element.
         // That's because this is just a transition event, the composite
@@ -357,9 +353,10 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
 
         // Otherwise, the composite element is likely not focused, so we need
         // this focus event to propagate so consumers can use the onFocus prop
-        // on <Composite>.
+        // on <Composite>. Propagation is controlled by the silent focus flag,
+        // not by the scroll option.
         else {
-          baseElement.focus();
+          baseElement.focus({ preventScroll: true });
         }
       };
 
--- a/app/src/sandbox/combobox-select-open-page-scroll/index.react.tsx
+++ b/app/src/sandbox/combobox-select-open-page-scroll/index.react.tsx
@@ -1,0 +1,96 @@
+import * as Ariakit from "@ariakit/react";
+import type { CSSProperties } from "react";
+import { useState } from "react";
+
+const branches = [
+  "main",
+  "0.14-stable",
+  "16.8.6",
+  "builds/facebook-www",
+  "devtools-v4-merge",
+  "fabric-cleanup",
+  "fabric-focus-blur",
+  "gh-pages",
+  "nativefb-enable-cache",
+  "rsckeys",
+];
+
+const selectedBranch = "fabric-focus-blur";
+
+const popoverStyle = {
+  background: "white",
+  border: "1px solid gray",
+  display: "flex",
+  flexDirection: "column",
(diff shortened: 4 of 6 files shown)


.changeset/300-composite-item-focus-scroll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`CompositeItem`](https://ariakit.com/reference/composite-item), and components based on it such as [`ComboboxItem`](https://ariakit.com/reference/combobox-item) and [`SelectItem`](https://ariakit.com/reference/select-item), scrolling the page when it receives focus inside a [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) composite widget rendered in a popup that hasn't been positioned yet.
```
diegohaz: Is this bug present in the published version of `@ariakit/react`? If it is, then this changeset is fine. Otherwise, we shouldn't create changesets when we fix bugs that haven't been published.

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -324,28 +324,24 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
...
-          // Keep the fallback scoped to nested scrollers in that case.
-          if (fromComposite) {
-            withDocumentScrollPreserved(currentTarget, () => {
-              currentTarget.scrollIntoView({
-                block: "nearest",
-                inline: "nearest",
-              });
-            });
-          } else {
+          withDocumentScrollPreserved(currentTarget, () => {
```
diegohaz: One thing the original PR fixed that this one doesn't is that we seem to be calling `scrollIntoView` even when it isn't needed, causing the browser to briefly show the scrollbar.

I wonder if we should use the non-standard [`scrollIntoViewIfNeeded`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded) API instead, both here and perhaps in other places where we currently call `Element.scrollIntoView()`. It's supported by Chrome and Safari, and we could have a cheap fallback for browsers that don't support it, currently Firefox.

Would that simplify things? Would `withDocumentScrollPreserved` still be needed in that case?

app/src/sandbox/combobox-select-tabs/test-browser.ts
```diff
@@ -39,4 +39,22 @@ withFramework(import.meta.dirname, async ({ test }) => {
       await test.expect(lastOption).toBeInViewport();
     });
   }
+
+  // https://github.com/ariakit/ariakit/pull/6976
+  test("opens the searchable select with an autofocused item without scrolling the page", async ({
```
diegohaz: I'm not sure we should reuse the `combobox-select-tabs` sandbox for this. If I understand correctly, this bug occurs in a specific scenario. I'm particularly concerned that a prop in this sandbox might change, such as removing `autoFocus` from the combobox items. That could cause the test to pass even if we break this feature again in the future. Is that correct?

We should probably have a dedicated sandbox for this.

I also wonder whether the real fix belongs elsewhere. If I understand correctly, the Popover component triggers autofocus before it has finished positioning. Shouldn't we fix it there instead? Shouldn't we wait for the popover to finish positioning before triggering autofocus? I'm afraid we might be fixing a specific use case, but until we fix the generic issue, we might have other problems with Popover + autoFocus.

packages/ariakit-react-components/src/composite/composite-item.tsx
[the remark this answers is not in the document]
diegohaz: Superseded by [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe): this line is back to `main`, so the unconditional wrapper is gone.

The report was correct, though, and it's what redirected the fix. Confirmed on WebKit with an offscreen inline `virtualFocus` composite whose item has `autoFocus`: `main` scrolls the page to 1372 to present it, and the unconditionally wrapped version left the page at 0 with the composite out of the viewport. Chromium and Firefox were unaffected, because they run the implicit focus scroll before dispatching the focus event.

The prescribed fix wouldn't have worked. `fromComposite` is about where focus came from, not popup versus inline, and the popup case this PR fixes also arrives from outside the composite, so restoring that split measures 628 to 127 on WebKit.

Tracing the open sequence showed the jump starts one step earlier, in the composite's scheduled focus, and that it's a regression from unreleased #6976, which lifted that focus call out of `withDocumentScrollPreserved`. Restoring the wrapper there fixes all three engines and leaves `CompositeItem` untouched, so this branch no longer changes the behavior you flagged.

.changeset/300-composite-item-focus-scroll.md
diegohaz: Fixed in [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe): removed it. The bug isn't reachable in a published version.

The whole `unstable_scrollIntoView` branch this now fixes was added by #6976, and its only library consumers are gated on `ComboboxSelect`, so nothing here is observable through a released API.

I also checked the earlier claim that the two offending calls in `CompositeItem` made this reachable in 0.4.35, and it doesn't hold. A standalone app on `@ariakit/react@0.4.35` running the documented `SelectPopover` plus `Combobox` pattern, where `SelectItem` marks the selected item with no consumer opt-in, measured across Chromium, Firefox and WebKit, two viewports, with and without `unmountOnHide`, four scroll positions, first open and reopen: the page didn't move in any of the 96 runs. Instrumentation showed the popup already positioned when the item received focus, because `usePopover` gates the dialog's autofocus on `positioned`.

packages/ariakit-react-components/src/composite/composite-item.tsx
diegohaz: `scrollIntoView` is being called when it isn't needed, but not on this line, and `scrollIntoViewIfNeeded` wouldn't have helped here.

Instrumenting every `Element.prototype.scrollIntoView` call while opening the searchable select, recording whether the target was already fully visible and whether any scroller actually moved:

| call site | target already visible | document moved |
| --- | --- | --- |
| `combobox.tsx`, `block: "nearest"` | yes | no |
| `centerIntoView`, `block: "center"` | yes | yes, 300 to 284, then restored |

The `block: "nearest"` calls already behave like `scrollIntoViewIfNeeded`: they resolve to the current position when the element is fully visible. The one that genuinely scrolls when it doesn't need to is `centerIntoView`, which asks for `block: "center"` even when the selected item is already fully visible, and then `restoreScrollPositions` puts every ancestor back. That's the likeliest source of the scrollbar flash, and it's #6976 code rather than this line.

At this line the item genuinely was offscreen, since the popup sits at its pre-placement origin, so on WebKit the fallback moved the document 300 to 96. `scrollIntoViewIfNeeded` would still have scrolled there. So no, it wouldn't have simplified this, and yes, something would still have had to preserve the document scroll.

That's moot now: this line is back to `main`. The fix moved to the composite's scheduled focus, which is where the regression came from, and `withDocumentScrollPreserved` is used there instead.

One thing worth recording about the no-op write guard #6984 had and this branch doesn't. I tried adding it, and it breaks `combobox-select-animation/test-safari.ts` ("does not scroll with smooth scrolling", from #6832) deterministically. Under `scroll-behavior: smooth`, `scrollIntoView` starts an animated scroll, so the scroller hasn't moved yet when the guard reads it back; the guard skips the restore and the smooth scroll lands. The unconditional write is what cancels it, which is why #6984 needed the `scroll-behavior: auto !important` override alongside the guard.

app/src/sandbox/combobox-select-tabs/test-browser.ts
diegohaz: Fixed in [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe): moved to a dedicated sandbox, `combobox-select-open-page-scroll`.

You were right, and the fixture dependency was worse than the `autoFocus` case you named. Measuring which props were actually load-bearing: `autoFocus` isn't one. `ComboboxItem` derives `data-autofocus` from the selected value whenever there's a `selectElement`, so removing it from `combobox-select-tabs` changes nothing and the bug still reproduces. `unmountOnHide` is the real invariant: without it the popup is already positioned when focus arrives and the bug doesn't reproduce at all, so the old test would have gone green. The new fixture pins `virtualFocus`, the `Combobox` being inside the popup, and `unmountOnHide`, each red-checked by removing it and watching the test fail in all three engines.

On whether the fix belongs in `Popover`: it already waits. `usePopover` keeps a `positioned` state that only flips after the first awaited `computePosition`, passes `autoFocusOnShow: positioned && autoFocusOnShow` to `useDialog`, and publishes `data-placing` so other components can wait; `Combobox` observes it for `autoSelect`. What doesn't wait is the composite's own scheduled focus. Reading `data-placing` and the popup rect at each `focus()` call with the page at 628:

```
focus  role=option    placing=true   popoverTop=-628   composite.tsx      (useScheduleFocus)
focus  role=combobox  placing=true   popoverTop=-628   composite-item.tsx (redirectFocusToBaseElement)
focus  role=combobox  placing=null   popoverTop=228    dialog.tsx         (autoFocusOnShow)
```

So the generic gate exists, the composite just wasn't using it. I prototyped gating `useScheduleFocus` on `data-placing`, and it does fix the jump in all three engines with the suite green. I didn't ship it: `data-placing` is also permanently set on closed-but-mounted popovers (six of them on the `combobox-select-6347` preview with everything closed), so the predicate can't tell "opening" from "closed", and the gate turns a synchronous focus move into an unbounded wait with no cancellation. `CompositeFocusOnMove` is ungated too.

What the branch does instead is narrower: this is a regression from #6976, which lifted the scheduled focus out of `withDocumentScrollPreserved`, so the fix restores that wrapper and `CompositeItem` is untouched. Your broader point stands, and the redesign is tracked in #6986.

diegohaz:
The review moved the fix. Tracing the open sequence showed the page jump is a regression from #6976, which lifted the composite's scheduled focus out of `withDocumentScrollPreserved`, so [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe) restores that wrapper and reverts `composite-item.tsx` to `main`. The PR description and title are updated.

**Fixed**

- [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe) ([reply](https://github.com/ariakit/ariakit/pull/6985#discussion_r3690471034)): moved the regression test to a dedicated `combobox-select-open-page-scroll` sandbox. Measuring the fixture showed `unmountOnHide`, not `autoFocus`, was the invariant the old test silently depended on. `Popover` does already wait for placement; the composite's scheduled focus didn't.
- [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe) ([reply](https://github.com/ariakit/ariakit/pull/6985#discussion_r3690470826)): removed the changeset. The bug isn't reachable in `@ariakit/react@0.4.35`, verified against the published package across 96 measurements in three engines.

**Obsolete**

- [`575b3bdfe`](https://github.com/ariakit/ariakit/commit/575b3bdfe) ([reply](https://github.com/ariakit/ariakit/pull/6985#discussion_r3690470721)): the unconditional wrapper is gone with the rest of the `composite-item.tsx` change. The report was accurate and is what redirected the fix, but the suggested `fromComposite` split would have left WebKit at 628 to 127.

**Declined**

- ([reply](https://github.com/ariakit/ariakit/pull/6985#discussion_r3690470942)): `scrollIntoViewIfNeeded` wouldn't have helped at that line, since the item genuinely was offscreen at the popup's pre-placement origin. The call that scrolls when it doesn't need to is `centerIntoView`. Also recorded why #6984's no-op write guard can't land on its own: it breaks `combobox-select-animation/test-safari.ts` under `scroll-behavior: smooth`.

**Tracked**

- [#6986](https://github.com/ariakit/ariakit/issues/6986) ([reply](https://github.com/ariakit/ariakit/pull/6985#discussion_r3690471034)): revisit composite focus presentation as a whole, including the `data-placing` gate prototype and why it wasn't shipped, the ungated `CompositeFocusOnMove` path, and keeping real and virtual focus both supported.

Issue #6986, opened 2026-07-31
Revisit composite focus presentation for real and virtual focus


diegohaz:
### Context

Found while addressing review feedback on https://github.com/ariakit/ariakit/pull/6985, which fixes a page jump when a `ComboboxSelect` opens on a scrolled page. That PR patches the point where the scroll happens. The review asked whether the real fix belonged in `Popover`, since it looked like autofocus was running before positioning finished. Investigating that turned up a broader problem in the composite focus logic that is too large and too delicate to resolve inside a bug fix, so it is recorded here.

The composite focus machinery now has several independent paths that move DOM focus, each with its own scroll-presentation policy, and no shared notion of when it is safe to move focus into a popup. It also has to keep serving both real focus and virtual focus. These differ in DOM focus ownership, not in whether the active item is presented to the user: real focus leaves DOM focus on the item, while virtual focus leaves primary DOM focus on the base element through `aria-activedescendant`.

### What the ordering actually is

`Popover` already waits for positioning before it autofocuses. `usePopover` keeps a `positioned` state that only flips after the first awaited `computePosition`, passes `autoFocusOnShow: positioned && autoFocusOnShow` to `useDialog`, and publishes `data-placing` on the popup so other components can wait too. `Combobox` observes `data-placing` for its `autoSelect` behavior, with the comment "prevent the focus from moving to the first item while the popover is still calculating its position, which could cause a scroll jump".

The composite's own scheduled focus does not observe it. Opening `Select with Combobox and manual Tab` on the `combobox-select-tabs` preview with the page at scrollY 300, captured by patching `HTMLElement.prototype.focus` and reading the popup's `data-placing` and viewport rect at each call:

```
focus  role=option    placing=true   popoverTop=-300   composite.tsx      (useScheduleFocus -> present)
focus  role=combobox  placing=true   popoverTop=-300   composite-item.tsx (redirectFocusToBaseElement)
focus  role=combobox  placing=null   popoverTop=228    dialog.tsx         (autoFocusOnShow, after positioning)
```

`popoverTop=-300` is the popup still sitting at its pre-placement origin, which is document `0,0`. The first two calls move DOM focus into it there. Only the third waits.

### The gate was prototyped, and it works

[...]

Issue #6838, opened 2026-07-24
Center the selected ComboboxSelect item on open


diegohaz:
## Context

Discovered while reviewing #6832. ComboboxSelect currently keeps a selected item visible when the popover opens, including asynchronous and virtualized lists, while deliberately avoiding page scroll.

## Work

When the popover opens and receives initial focus, scroll the selected item toward the middle of its nearest scrollable viewport. Preserve nearest-edge scrolling for later arrow-key movement and never scroll the document. Add a browser fixture with a long list and a far-down selected value that asserts both the initial midpoint and unchanged `window.scrollY`.

## Expected outcome

A newly opened ComboboxSelect presents the selected option near the center of the popup viewport without disruptive recentering during navigation or page-scroll jumps.

This is independent of #6832 because the current ensure-visible behavior is correct and sufficient for the unified select to ship.

## Workaround

See the [corrected temporary userland workaround](https://github.com/ariakit/ariakit/issues/6838#issuecomment-5131243236).

diegohaz:
The implementation merged in #6976 is being reverted. The centered selected-item goal remains valid, but the follow-up work in #6994 showed that extending the current Composite-based presentation path required disproportionate scheduling, cancellation, navigation-ordering, and outer-scroll work.

Until the focus-presentation logic is revisited holistically, `ComboboxSelect` will return to the previous nearest-edge behavior, which keeps the selected item visible without the centering machinery. The centering requirement, coverage matrix, and constraints learned from #6994 are now tracked in #6986, so this issue remains closed as superseded by that broader work.

PR #7003, opened 2026-08-01
Revert ComboboxSelect selected-item centering

diegohaz:
## Motivation

The selected-item centering added in #6976 expanded the shared Composite focus scheduler and introduced document scroll side effects. Follow-up #6994 showed that making it reliable required open-cycle, cancellation, navigation-ordering, popup-lifecycle, and ancestor-scroll policy that was disproportionate to the presentation benefit. The behavior has not been published.

## Solution

Restore the pre-#6976 nearest-edge presentation path, remove the private centering callback and helper, and delete center-specific fixtures while preserving later changes from main. The selected-item path is again handled by the existing nearest-edge presentation:

```ts
activeElement.scrollIntoView({
  block: "nearest",
  inline: "nearest",
});
```

Add the independent `unmountOnHide` focus-cancellation case originally proposed in the unmerged #6994. The broader presentation redesign, including centered selected items, is tracked in #6986.

## Preview

Opening the filterable ComboboxSelect keeps the selected Watermelon visible at the nearest edge without centering it.

Selected Watermelon visible at the nearest edge of the open ComboboxSelect popup

[Interaction recording: open the filterable ComboboxSelect and keep Watermelon at the nearest edge](https://github.com/user-attachments/assets/1e5371e6-6242-4d5c-8af6-56a391ea2e17)

## Tests

- `pnpm lint-fix`
- `pnpm build`
- `pnpm clean`
- `pnpm tsc`
- Focused Vitest on current React and React 18: 2 files, 3 tests in each run
- `pnpm build-app-lite`
- Desktop Playwright in Chrome, Firefox, and Safari: 54 tests
- iOS Playwright: 1 test

## Notes

No changeset is included because the centered behavior has not been published.


app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
```diff
@@ -78,81 +49,35 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+    await test.expect(q.option("Watermelon")).not.toBeInViewport();
   });
 
-  // https://github.com/ariakit/ariakit/issues/6838
-  test("keeps nearest-edge scrolling after virtual focus movement", async ({
+  // Locks in the same invariant for a popup that remounts on every open, where
+  // the select element is briefly the composite again. This is invariant
+  // coverage rather than a regression test: it also passes without the
+  // presentation-cycle handling added in
+  // https://github.com/ariakit/ariakit/pull/6994
```
diegohaz: This comment cites presentation-cycle handling "added in" #6994, but #6994 was closed without merging, so that handling was never added to this repository. `gh pr view 6994 --json state,mergedAt` returns `CLOSED` with `mergedAt: null`, and searching every reachable commit for the mechanism that PR introduced (`git log --all -S "unstable_presentActiveItem"`) returns only [`e2d8dd42d`](https://github.com/ariakit/ariakit/commit/e2d8dd42d), which is #6994's own branch head and is not reachable from `main`.

The sentence therefore presents a counterfactual as history. A reader auditing this test follows the link to a closed, unmerged pull request and cannot tell what the "also passes without" claim was measured against, because the named handling has never existed in any tree this test can run against. The pull request description carries the same framing ("Keep the independent `unmountOnHide` focus-cancellation coverage from #6994"), which reads as retained coverage rather than coverage this pull request newly adds.

Separately, this test's body is identical to `cancels presentation when real focus moves` directly above it apart from the combobox accessible name, and I could not find a mutation that separates the two. Removing the `hasFocus(scrollBaseElement)` cancellation guard in `useScheduleFocus`, and separately removing the `data-autofocus` guard in `scheduleScroll`, each left all five tests in this file passing in Chrome. Together with the comment's own statement that this is not a regression test, the new `unmountOnHide` fixture and this test are extra surface in a pull request whose stated motivation is that this machinery's cost was disproportionate to its benefit.

I did not exhaustively search for a discriminating mutation, so this is evidence that the two tests overlap rather than proof that the new one can never fail alone.

Two reasonable paths:

1. Drop the `unmountOnHide` fixture and this test, keeping the revert limited to restoring the pre-#6976 state.
2. Keep them, and rewrite the comment to state what the test guards in this tree without citing an unmerged pull request.

**Assessment**

- Severity: Low (verified), because it misdescribes repository history rather than changing runtime behavior.
- Priority: P1 (estimate, 90% confidence), because this pull request introduces the text and the fix is a comment rewrite or a small deletion before merge.
- Breakage likelihood: Likely (verified), because every future reader of this test meets the claim and the dead link.
- Fix effort: Easy (verified, 95% confidence), because it is a comment rewrite or removal of one fixture and one test.

diegohaz (commented):
Initial review. There were no existing review threads on this pull request, so the baseline is the merge base [`370291c3a`](https://github.com/ariakit/ariakit/commit/370291c3a) to head [`9d38c52f9`](https://github.com/ariakit/ariakit/commit/9d38c52f9), expanded into the surrounding code, the reverted pull requests, and the affected fixtures and suites.

## What I verified

**The library revert is exact.** Diffing head against the pre-#6976 tree ([`f4aedae2e`](https://github.com/ariakit/ariakit/commit/f4aedae2e)) for `composite.tsx` leaves exactly one difference: the `A focus handoff must scroll before ...` comment, whose head wording already matches `main` after #6995 removed the offscreen observer. Worth noting for other reviewers, since it surprised me at first: the `selectMode` branches in `useScheduleFocus` are restored pre-#6976 code, not new combobox knowledge being pushed into `Composite`. #6985's `composite.tsx` change lived entirely on the `canScrollIntoView` path that #6976 created, so it correctly disappears with it.

**No leftovers.** Repository-wide searches for `unstable_scrollIntoView`, `scrollSelectedItemIntoView`, `centerIntoView`, and `combobox-center-into-view` return nothing.

**`composite-late-item-focus` matches** what #6976 added there, 21 fixture lines and 24 test lines, removed exactly.

**Keeping #6976's `flushFrames` JSDoc in `app/src/test-utils/preview.ts` is the right call.** It is generic guidance about preferring retrying assertions, not center-specific.

**The "no changeset" claim holds.** #6976 shipped no changeset, and no file under `.changeset/` mentions the centering behavior, so no pending entry describes behavior this pull request removes.

Commands run on head, with an Astro dev server started from this checkout on a dedicated port and confirmed by the server process working directory:

- `pnpm tsc`: 0 errors.
- `pnpm lint`: clean.
- `pnpm test combobox-select composite-late-item-focus combobox`: 65 files, 413 tests passed.
- `pnpm -F app test-chrome` over `combobox-select-selected-value-presentation`, `composite-late-item-focus`, and `combobox-select-scroll-into-view`: 17 passed.
- `pnpm -F app test-ios combobox-select-selected-value-presentation`: 1 passed.

## Published inline

One finding, on the new `cancels presentation when real focus moves in an unmounted popup` test and its comment.

## Other candidates and dispositions

**Fixture insertion order. Declined non-goal.** The new `Focus moving unmounted fruit` fixture is inserted above `Touch filterable fruit` instead of appended where `Virtual focus fruit` was removed, which pushes the iOS test's target further down the document. I chased this because `test-ios.ts` asserts `window.scrollY` is still `100` after a second click, and a target below the fold would make Playwright's actionability scroll break that assertion. It does not break: the iOS test passes on head, and the target keeps comfortable slack inside the 746px iPhone 13 Pro Max viewport at `scrollY` 100. Appending rather than inserting would have kept every existing fixture at its original offset for free, but there is no defect to fix.

**Happy-dom coverage for #6858 is not restored. Declined non-goal, flagged so the decision is explicit.** #6976 deleted `app/src/sandbox/combobox-select-scroll-into-view/test.ts`, and this revert does not bring it back. I restored it temporarily and ran it twice: it passes against head, and it also passes against `main`'s library files. Its deletion was therefore not forced by centering, so this revert does not have to undo it. The authoritative browser coverage for that issue is still in place at `combobox-select-scroll-into-view/test-browser.ts`, and the deleted duplicate asserted through `vi.spyOn(HTMLElement.prototype, "scrollIntoView")` rather than through user-visible state. Happy to see it restored if you would rather keep the duplicate.

**Issue #6838 is still closed while its implementation is reverted. No change needed here.** #6986 is open and its "Selected-item presentation inherited from #6838" section already records that #6976 is being reverted and folds the centering goal into the redesign, so the follow-up is durably registered and no new issue is warranted. Someone landing on #6838 itself still sees a closed issue with no note that its implementation was reverted, so a short cross-link comment there would close the loop outside this pull request.

## Not fully validated

- Firefox, Safari, and Android Playwright projects. I ran Chrome and iOS locally; the description reports 54 desktop tests across the three desktop engines.
- The React 18 suite. The description reports it; I ran only the current React suite.
- I did not exhaustively search for a mutation that separates the two `cancels presentation` tests, as noted inline.

The revert itself is clean, complete, and well evidenced. The one newly discovered finding concerns the added test rather than the reverted code.

app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
diegohaz: Fixed in [`3b28b5e24`](https://github.com/ariakit/ariakit/commit/3b28b5e249627b6e69201f86fbb2c3be3add5caa): rewrote the comment to state that the case originated in unmerged #6994 and to describe the current-tree popup-remount/composite invariant. I kept the test because `unmountOnHide` exercises a distinct lifecycle, and updated the PR description with the same provenance.

diegohaz:
**Fixed**

- [`3b28b5e24`](https://github.com/ariakit/ariakit/commit/3b28b5e249627b6e69201f86fbb2c3be3add5caa) ([reply](https://github.com/ariakit/ariakit/pull/7003#discussion_r3696098663)): Clarified that the popup-remount lifecycle case originated in unmerged #6994, described the current-tree invariant, and updated the PR description with the same provenance.

PR #7006, opened 2026-08-02
Unify composite focus presentation

### LINKED ISSUE #6986: Revisit composite focus presentation for real and virtual focus

diegohaz:
## Motivation

Composite focus was presented through separate paths for initial selection, keyboard movement, virtual-focus redirects, `autoFocus`, and Dialog focus. When a Popover positioned asynchronously, those paths could race: focus entered popup content too early, native focus scrolling moved the page or unrelated ancestors, and real and virtual focus produced different results.

The legacy website exposed one more composition: its inner-page header opens a searchable `SelectPopover` through a non-composite `MenuList`, while `useDeferredValue` mounts the combobox input after the popup. The popup initially focused its dialog container and never retargeted the late input, both when clicking the current-page control and when pressing Command/Ctrl+K.

This replaces those paths with one presentation policy. Closes #6986.

## Solution

Popover now owns a private presentation scope that publishes its open cycle, positioning readiness, owner, and logical scrollport. Every focus source delegates to the same abortable transaction:

```ts
scheduleFocusPresentation({
  getTarget,
  getScopeTarget,
  isValid,
  focus: true,
  scroll: "nearest",
});
```

The transaction follows one sequence:

`intent → bind target and scope cycle → wait for readiness and settlement → focus with preventScroll → revalidate ownership → scroll the owned port`

This gives real focus, virtual focus, Composite movement, Focusable/Dialog autofocus, and item-to-base handoffs the same cancellation and presentation semantics. A virtual-focus redirect remains part of the same focus transaction, so its intentional blur cannot cancel the remaining scroll step.

Selected items in `ComboboxSelect` and `Select` popovers use the same coordinator with `focus: false` and `scroll: "center"` on open. Later navigation uses nearest-edge scrolling. The coordinator does not depend on `data-placing` or other DOM markers for correctness.

The owned scrolling helper handles nested scrollports, scroll padding and margins, transformed/fractional geometry, oversized targets, RTL, and vertical or sideways writing modes without scrolling unrelated ancestors.

`ComboboxPopover` and composed `SelectPopover` now share one internal focus policy for autofocus enablement, the initial focus target, and final focus restoration. `SelectPopover` keeps its existing first-tabbable behavior when the input participates in the popup's initial presentation. If the input registers only after focus has fallen back to the popup container, the policy promotes that late input once for the current open cycle. An intentional initial focus on another child remains authoritative.

The follow-up sandbox reproduces the legacy `SelectPopover → MenuList → Combobox` composition, including the shared stores, deferred input, and both opening paths. Its explicit `initialFocus` workaround is removed on the final library implementation.

### Deferred legacy select menu

When a `Combobox` mounts after an already-open `SelectPopover`, capture the input element in state and pass it to `initialFocus`:

```tsx
// TODO: Remove after https://github.com/ariakit/ariakit/issues/6986 is fixed.
const [inputElement, setInputElement] = useState<HTMLInputElement | null>(null);

<Combobox ref={setInputElement} />
<SelectPopover initialFocus={inputElement}>
  {/* ... */}
</SelectPopover>
```

The state callback ref is important because it rerenders the popover when the deferred input mounts. A stable object ref whose `current` value changes would not trigger the `initialFocus` update. This workaround only targets the deferred-input `SelectPopover` composition; the production fix coordinates it without consumer state.

### Composite page-scroll workaround

For consumers pinned to an older release, the complete temporary [`useCompositeFocusWorkaround`](https://github.com/ariakit/ariakit/blob/302064d294f1947def676708dc02a618ed4c786c/app/src/sandbox/combobox-select-open-page-scroll/composite-focus-workaround.react.ts) remains available at the workaround commit:

```tsx
// TODO: Remove after https://github.com/ariakit/ariakit/issues/6986 is fixed.
const workaround = useCompositeFocusWorkaround();

<ComboboxSelect ref={workaround.selectRef} />
<ComboboxPopover
  ref={workaround.popoverRef}
  autoFocusOnShow={workaround.autoFocusOnShow}
  onFocusCapture={workaround.onFocusCapture}
>
  <Combobox aria-label="Search" />
  <ComboboxList ref={workaround.scrollportRef}>
    {items.map((item) => (
      <ComboboxItem key={item} value={item} />
    ))}
  </ComboboxList>
</ComboboxPopover>
```

The composite workaround is intentionally limited: it observes Ariakit's private `data-placing` marker, temporarily removes `data-autofocus`, handles one supplied vertical scrollport, and cannot coordinate every Composite/Dialog focus path. The production fix replaces it entirely.

## Testing

The regression tests were first verified red against the previous implementation, including premature popup focus, page/ancestor scrolling, incorrect selected-item alignment, a late virtual-focus item canceling its own presentation, and the legacy deferred input leaving focus on the popup container.

The deferred legacy regression was committed red in [`56bf031df`](https://github.com/ariakit/ariakit/commit/56bf031df80bb417cb480143bf066f8274dc8111) and verified green with the public userland workaround in [`042e584fd`](https://github.com/ariakit/ariakit/commit/042e584fdc9047f9af5ce322305686cce92b4346). The final library implementation removes that workaround and passes:

- React 19 StrictMode: 185 files, 984 tests.
- React 18.3.1 StrictMode: 185 files, 984 tests.
- Chromium, Firefox, and Safari/WebKit: 96/96 focused browser tests across the three engines.
- Repository-wide lint and formatting checks.
- TypeScript and Astro checks with zero diagnostics.
- The production-lite app build.
- Direct verification on the React 18 legacy website: both the current-page control and Command+K focus the `Search examples` combobox and accept immediate typing.

Coordinator, ShadowRoot, cancellation, nested scrollport, scroll geometry, selected-value presentation, deferred input focus, and existing legacy Select behavior all have focused regression coverage.

## Preview

[Legacy website: click and Command+K focus the deferred combobox input](https://github.com/user-attachments/assets/a421c05c-f96c-40b9-8fb7-e1b82a2aecdf)

## Changeset

Adds patch changesets for `@ariakit/react-components` and `@ariakit/react`.


diegohaz:
Closing in favor of #7009, which takes the same core insight (focus with `preventScroll`, then present the item as a separate step that waits for placement) and closes #6986 with a much smaller surface.

Both branches address every defect #6986 enumerates. The difference is cost and blast radius.

## Why #7009 instead

#7009 is net negative on implementation code (+187 / -203 in `packages/**`, no new modules, five mechanisms deleted against two added), where this branch is +1721 / -349 across 26 package files and five new modules totalling ~1440 lines.

Readiness is the clearest divergence. #7009 publishes one `@private` boolean on the popover store from a single `sync(store, ["mounted"], ...)` writer, and StrictMode, `unmountOnHide: false`, and popups with nothing to position all fall out of that one rule. This branch coordinates a five-field snapshot written by four independent producers, two monotonic counters, and a microtask-deferred owner detach.

Layering: #7009 adds no `composite/**` to `popover/**` dependency and reads placement by duck-typing the store key, the same way `contentElement` already is. This branch relocates Popover lifecycle vocabulary (`open`, `cycle`, `positionRun`, `beginPositioning`) into `focusable/`, the lowest primitive in the library, and `select/select-popover.tsx` ends up importing its focus policy from `combobox/combobox-popover-presentation.ts`.

Performance is measured, not theoretical. The perf run on this branch flags four confirmed significant regressions: combobox close 10.4ms to 17.6ms scripting (+69%), combobox move across items +10%, combobox-select move across items +12%, composite move across items +14%. The drift also reaches widgets this change is not about, such as tooltip move without tooltips (+7%) and tab move (+8%), because `focusable.tsx` gained per-element work. #7009 does not touch `focusable.tsx` at all and reports no significant changes.

`scroll-into-view.ts` reimplements the CSSOM-View algorithm (scroll padding and margin, transform scale, writing modes). This branch's own code falls back to native `scrollIntoView` when there is no scope, so the engine is not needed for correctness. It exists to center inside the popup without cascading to the viewport, which re-lands the behavior #7003 reverted from `main` two hours before both branches were cut. #7009 registers that as #7011 instead, so the geometry can arrive together with the centering decision rather than ahead of it.

## Two regressions found on this branch

**`alwaysVisible` composites stop moving DOM focus.** Running #7009's `app/src/sandbox/menu-always-visible` fixture unchanged against all three refs: `main` passes, #7009 passes, this branch fails. `data-active-item` advances correctly but DOM focus stays on the first item, so a roving tabindex menu becomes keyboard-unnavigable and screen readers lose the cursor. The cause is the scope gate in `focusable/focus-presentation.tsx`, which runs `if (!snapshot.open) { cancel(); return; }` before the focus call, so a composite inside a rendered-but-closed popup cancels the whole presentation rather than just the scroll. This is the same class of defect as #6986 point (1).

**The deferred `SelectPopover` combobox failure was self-inflicted.** Running this branch's own `menu-combobox-deferred-mount` test against `main` passes 2/2, and against #7009 passes 2/2, but against [`b2fc8a2b`](https://github.com/ariakit/ariakit/commit/b2fc8a2b5d623a20b8fcfcf29f8057338c52994b) it fails 2/2. So four of the eight commits here, plus `combobox-popover-focus.ts`, repair a regression this branch introduced rather than one that existed on `main`. The real React 18 website may still have had a separate issue, but the sandbox as written does not reproduce anything against `main`.

## Also worth carrying over

`@ariakit/react-components/composite/utils` is a published subpath, and this branch removes `withDocumentScrollPreserved`, `focusSilently`, and `silentlyFocused` from it under a `patch` changeset. #7009 removes one of the three and is equally undisclosed, so that still needs a changeset note there. Repository precedent is `.changeset/100-remove-offscreen-items.md`, which used `minor` with an explicit **BREAKING** label.

The 1228 lines of package-source tests do not carry forward: `focus-presentation.test.ts` renders React components under `packages/ariakit-react-components/src/**`, which app-owned fixtures are supposed to cover, and both files were added in the same commit as the modules they test, so they are not regression tests.

What this branch did better and should be kept as practice is the red proof. [`85ee2bfe`](https://github.com/ariakit/ariakit/commit/85ee2bfe9330ba21de3e07c9aac7d329a8365077) recorded 5 genuine CI failures across Chrome, Firefox, Safari, and both React versions against unmodified library source, and [`042e584f`](https://github.com/ariakit/ariakit/commit/042e584fdc9047f9af5ce322305686cce92b4346) closed a clean red to green pair with a userland workaround alone. #7009's regression tests were force-pushed alongside the fix, so its red state was never publicly recorded even though the commit ordering is correct. That verification loop is worth repeating regardless of which implementation lands.

Centering is tracked in #7011 and the remaining non-presenting focus moves in #7012.

PR #7008, opened 2026-08-02
Add composite element state

diegohaz:
## Motivation

Composite stores expose their composite DOM node and arrow-key inclusion through `baseElement` and `includesBaseElement`, which do not clearly describe those contracts. Consumers need explicit names without breaking existing store configuration or connected-store behavior.

## Solution

Add `compositeElement`, `setCompositeElement`, and `compositeElementInFocusOrder`, with the old APIs retained as synchronized, soft-deprecated aliases on initialized stores. Migrate internal call sites and examples and add coverage for initialization precedence, direct state updates on initialized stores, connected stores, and public deprecation metadata.

```tsx
<CompositeProvider compositeElementInFocusOrder>
  <Composite />
</CompositeProvider>
```

<details>
<summary>Cycle 3: Simplify Composite alias synchronization</summary>

**Assessment**

The original direction added a Composite-specific helper and six order-sensitive setter spreads solely to synchronize aliases on derived stores before initialization. That edge case is expected to be rare, is outside established store-composition semantics, and introduced disproportionate implementation and maintenance complexity for otherwise low-impact behavior.

**Decision**

Use the same local normalization pattern as Combobox's `inputValue`/`value` compatibility inside `createCompositeStore`. Keep setup listeners for initialized and connected stores, remove the helper and derived-store spreads, and treat alias convergence on uninitialized derived stores as unsupported. Normal React usage and initialized core or connected stores remain fully synchronized.

</details>

<details>
<summary>Follow-up: Rely on initialized store synchronization</summary>

**Assessment**

The local alias-normalization branches in Composite and Combobox primarily provide convergence before store initialization. Initialized stores already have setup listeners for two-way synchronization. However, React attaches refs before layout-effect initialization, so a legacy `setBaseElement` ref callback must preserve the element across initialization.

**Decision**

Remove custom alias handling from both public `setState` paths and rely on the existing setup listeners after initialization. Route the deprecated `setBaseElement` setter through the canonical `compositeElement` key so pre-initialization ref writes survive. Keep Combobox's remaining `setState` wrapper for select-option and selected-value ownership. Raw writes to alias keys on an uninitialized core store may temporarily diverge, which is intentionally unsupported.

</details>

<details>
<summary>Cycle 6: Remove deprecated-only pre-init tracking</summary>

**Assessment**

A connected `@ariakit/store` containing only deprecated Composite keys can overwrite a newer update made between child-store creation and initialization. This is reachable through the public store utilities, but it requires a legacy-only state shape and a write in that narrow lifecycle window; current Ariakit stores contain both aliases. The possible user impact is stale element or focus-order state in that path. Protecting it required two tracked values, two reconciliation branches, four listener assignments, and a dedicated regression contract, partially reintroducing the temporal machinery removed in earlier simplifications. The modern reference-order assertion added alongside that machinery was also inert because the extractor sorts deprecated props last and then alphabetically.

**Decision**

Treat deprecated-only connected-store writes before initialization as outside the compatibility contract. Remove the tracking and its dedicated test, and keep a concise support-boundary comment next to the initialization listeners. Preserve initialized core, derived, and connected-store synchronization, plus the deprecated `setBaseElement` pre-init ref path. Remove the inert modern extractor assertion while retaining the declaration-order-sensitive legacy extractor assertion.

</details>


packages/ariakit-components/src/composite/__utils.ts
```diff
@@ -0,0 +1,50 @@
...
+      return;
+    }
+    store.setState(key, value);
+  };
+
+  return {
+    setState,
+    setCompositeElement: (element) => setState("compositeElement", element),
+    setBaseElement: (element) => setState("compositeElement", element),
+  };
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`7158db0c2`](https://github.com/ariakit/ariakit/commit/7158db0c2eb192e916b671382ff8ba07e72fa8b6): `setBaseElement` now delegates through the `baseElement` state key while alias synchronization preserves two-way compatibility.

diegohaz:
**Fixed**

- [`7158db0c2`](https://github.com/ariakit/ariakit/commit/7158db0c2eb192e916b671382ff8ba07e72fa8b6) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698076461)): `setBaseElement` now delegates through the `baseElement` state key while preserving two-way alias synchronization.

packages/ariakit-components/src/composite/__utils.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { SetState } from "@ariakit/utils";
+import type { CompositeStoreState } from "./composite-store.ts";
+
+interface CompositeStoreSetters<S extends CompositeStoreState> {
+  setState: Store<S>["setState"];
+  setCompositeElement: SetState<S["compositeElement"]>;
+  setBaseElement: SetState<S["baseElement"]>;
+}
+
+export function createCompositeStoreSetters<S extends CompositeStoreState>(
```
diegohaz: Is this the best way to sync the two states? What about what we're doing with `inputValue`/`value`? Shouldn't we use the same pattern?

packages/ariakit-components/src/combobox/combobox-store.ts
```diff
@@ -452,16 +478,20 @@ export function createComboboxStore({
...
     }
-    combobox.setState(key, value);
+    compositeSetters.setState(key, value);
   };
 
   return {
     ...popover,
     ...composite,
     ...combobox,
+    ...compositeSetters,
```
diegohaz: Why do we need a specific `...compositeStoreSetters` here? Why we can't just get it from `...composite`?

packages/ariakit-components/src/composite/__utils.ts
diegohaz: Fixed in [`6cb882ce2`](https://github.com/ariakit/ariakit/commit/6cb882ce25a7400ab3a2f53899ccb65a38b6486b): moved the two-way aliases into `createCompositeStore` using the same local normalization pattern as Combobox's `inputValue`/`value`, while retaining setup synchronization for initialized and connected stores.

packages/ariakit-components/src/combobox/combobox-store.ts
diegohaz: Fixed in [`6cb882ce2`](https://github.com/ariakit/ariakit/commit/6cb882ce25a7400ab3a2f53899ccb65a38b6486b): removed `compositeStoreSetters` and the derived-store spreads; initialized derived stores now use the setters from `...composite`.

diegohaz:
**Fixed**

- [`6cb882ce2`](https://github.com/ariakit/ariakit/commit/6cb882ce25a7400ab3a2f53899ccb65a38b6486b) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698328597)): Simplified Composite alias synchronization to the local Combobox pattern.
- [`6cb882ce2`](https://github.com/ariakit/ariakit/commit/6cb882ce25a7400ab3a2f53899ccb65a38b6486b) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698328711)): Removed the derived-store helper and setter spreads.

packages/ariakit-components/src/combobox/combobox-store.ts
```diff
@@ -705,7 +728,7 @@ export interface ComboboxStoreOptions<
   extends
     StoreOptions<
       ComboboxStoreState<T>,
-      | "includesBaseElement"
+      | "compositeElementInFocusOrder"
```
diegohaz, line 731: Does removing `includesBaseElement` from here break any TypeScript code that relies on `includesBaseElement` today or it's fine?

packages/ariakit-react-components/src/composite/composite-store.ts
```diff
@@ -71,17 +75,49 @@ export interface CompositeStoreItem extends Core.CompositeStoreItem {}
 export interface CompositeStoreState<
   T extends CompositeStoreItem = CompositeStoreItem,
 >
-  extends Core.CompositeStoreState<T>, CollectionStoreState<T> {}
+  extends Core.CompositeStoreState<T>, CollectionStoreState<T> {
+  /**
+   * The composite element itself.
+   *
+   * @deprecated Use `compositeElement` instead.
+   */
```
diegohaz: Shouldn't TypeScript automatically infer these JSDocs in `packages/ariakit-react-components/src/composite/composite-store.ts` since we're extending `Core.CompositeStoreState` already?

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -503,8 +554,10 @@ export function createCompositeStore<
   return {
     ...collection,
     ...composite,
-
-    setBaseElement: (element) => composite.setState("baseElement", element),
+    setCompositeElement: (element) =>
+      composite.setState("compositeElement", element),
+    setBaseElement: (element) =>
+      composite.setState("compositeElement", element),
```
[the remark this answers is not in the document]
diegohaz: Not a real problem.

packages/ariakit-components/src/combobox/combobox-store.ts
```diff
@@ -436,22 +460,15 @@ export function createComboboxStore({
 
   const setInputValue: ComboboxStore["setInputValue"] = (value) => {
     combobox.setState("inputValue", value);
-    combobox.setState("value", combobox.getState().inputValue);
   };
```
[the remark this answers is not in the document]
diegohaz: We don't support pre-init.

packages/ariakit-components/src/combobox/combobox-store.ts
diegohaz: Already handled at packages/ariakit-components/src/combobox/combobox-store.ts:743: `ComboboxStoreOptions` still extends `CompositeStoreOptions<ComboboxStoreItem>`, which explicitly declares the deprecated optional `includesBaseElement` property. A strict TypeScript compile confirmed this remains valid:

```ts
const options: ComboboxStoreOptions = { includesBaseElement: true };
```

packages/ariakit-react-components/src/composite/composite-store.ts
diegohaz: Fixed in [`7c97e215d`](https://github.com/ariakit/ariakit/commit/7c97e215d71ad65305e2b8b0715c563c0b4d23b7): removed the redundant React wrapper redeclarations and rely on the inherited Core interfaces. TypeScript and both the modern and legacy reference extractors preserve the deprecation metadata.

diegohaz:
**Fixed**

- [`7c97e215d`](https://github.com/ariakit/ariakit/commit/7c97e215d71ad65305e2b8b0715c563c0b4d23b7) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698547785)): Removed the redundant React Composite declarations; inherited Core JSDoc remains available to TypeScript and both reference extractors.

**Already resolved**

- [`ComboboxStoreOptions`](https://github.com/ariakit/ariakit/blob/7c97e215d71ad65305e2b8b0715c563c0b4d23b7/packages/ariakit-components/src/combobox/combobox-store.ts#L743) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698546787)): The deprecated `includesBaseElement` option remains inherited through `CompositeStoreOptions`; a strict TypeScript compile confirmed compatibility.

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -306,6 +323,29 @@ export function createCompositeStore<
...
+        );
+      }),
+      sync(composite, ["includesBaseElement"], (state) => {
+        composite.setState(
+          "compositeElementInFocusOrder",
+          state.includesBaseElement,
+        );
+      }),
+    ),
+  );
```
diegohaz, line 349: These mirrors overwrite a connected store's newer deprecated value at initialization, and push the stale value back into that store.

`sync` invokes its listener immediately at registration, so when this setup runs, the `compositeElement` → `baseElement` direction always wins. But `storeInit` pushes a connected store's values down *before* it runs the registered `setup` callbacks, so anything that reached `baseElement`/`includesBaseElement` from the connected store after `createCompositeStore` computed `initialState` is immediately reverted to the canonical value captured at creation time. Because these are plain `composite.setState` calls, the stale value is also fanned back out to the connected store, silently rewriting state the consumer owns.

The consumer scenario is a store carrying only the deprecated keys, which is exactly the shape the new `syncs deprecated aliases with connected stores` test builds, being written to before the composite store is initialized. In React that ordering is normal: ref callbacks run before the layout effect that calls `init`.

I ran this probe against `origin/main` (`fdcf2038a`) and this head (`7c97e215d`):

```ts
const first = document.createElement("div");
const second = document.createElement("div");
const source = createStore({ baseElement: first, includesBaseElement: true });
const store = createCompositeStore({ store: source });

source.setState("baseElement", second);
source.setState("includesBaseElement", false);
init(store);
```

|                                        | `main`   | this head |
| -------------------------------------- | -------- | --------- |
| `store.getState().baseElement`         | `second` | `first`   |
| `store.getState().includesBaseElement` | `false`  | `true`    |
| `source.getState().baseElement`        | `second` | `first`   |
| `source.getState().includesBaseElement`| `false`  | `true`    |

Swapping the registration order does not fix it, it only moves the problem: the deprecated key would then win and clobber a pre-init `setCompositeElement`, which is the case `preserves a base element set before initialization` covers. The mirrors have to know which alias actually moved. Tracking the last mirrored value, the same approach `combobox-store.ts` already uses for `syncedSelectElement` and `syncedAnchorElement`, restores `main`'s behavior:

```ts
let syncedElement = compositeElement;
let syncedInFocusOrder = compositeElementInFocusOrder;

setup(composite, () => {
  const state = composite.getState();
  if (state.baseElement !== syncedElement) {
    composite.setState("compositeElement", state.baseElement);
  }
  if (state.includesBaseElement !== syncedInFocusOrder) {
    composite.setState(
      "compositeElementInFocusOrder",
      state.includesBaseElement,
    );
  }
  return chain(
    sync(composite, ["compositeElement"], (state) => {
      syncedElement = state.compositeElement;
      composite.setState("baseElement", state.compositeElement);
    }),
    // ...the other three mirrors, each assigning its tracked value first
  );
});
```

I applied that locally and the probe matches `main` on all four rows, with the full `pnpm test` suite green (248 files, 1670 tests).

If the extra bookkeeping isn't worth it for this store shape, the alternative is to narrow the claim rather than the code: rename or drop the connected-store test so it doesn't imply that a deprecated-only store stays authoritative across `init`, and say so where the support boundary is documented.

**Assessment**

- Severity: Medium (verified), because the composite store silently reverts state owned by the consumer's own store, with no warning and no way to observe it before the next write.
- Priority: P2 (estimate, 65% confidence), because the trigger needs a connected store that carries only the deprecated keys, but this PR ships a test for exactly that shape.
- Breakage likelihood: Rare (verified, 85% confidence), because it also requires that store to change between `createCompositeStore` and `init`. Stores created by this version always carry both aliases and therefore take `storeInit`'s ordered per-key `sync` path, which stays consistent; I confirmed that variant is unaffected.
- Fix effort: Easy (verified), because the tracked-value variant above is local to this block and passes the full suite.

packages/ariakit-react-components/src/composite/composite-store.ts
```diff
@@ -24,7 +24,11 @@ export function useCompositeStoreProps<T extends Core.CompositeStore>(
 ) {
   store = useCollectionStoreProps(store, update, props);
   useStoreProps(store, props, "activeId", "setActiveId");
-  useStoreProps(store, props, "includesBaseElement");
+  const focusOrderProps = {
+    compositeElementInFocusOrder:
+      props.compositeElementInFocusOrder ?? props.includesBaseElement,
+  };
+  useStoreProps(store, focusOrderProps, "compositeElementInFocusOrder");
```
diegohaz, line 31: Nothing tests this fallback after the fixture migration, so the deprecated React prop can silently stop working.

Every composite-derived store hook routes through `useCompositeStoreProps` (`radio`, `toolbar`, `menubar`, `tab`, `combobox`, `menu`, `select`, `tag`), which makes this `??` the single line that keeps `includesBaseElement` working as a prop on `CompositeProvider`, `ComboboxProvider`, `TabProvider`, and the rest. This PR migrated every fixture that exercised it (`combobox-radix-select`, `combobox-tabs-old`, `dialog-combobox-4825`, `dialog-combobox-command-menu`, `menu-nested-combobox`, `menu-nested-search-interactions`), and `grep -rn includesBaseElement app examples` now only matches `app/src/lib/jsdoc-loader.test.ts`.

I verified the gap by collapsing this block to `useStoreProps(store, props, "compositeElementInFocusOrder")`, dropping the deprecated prop entirely: the full Vitest suite still passes (248 files, 1670 tests). The browser suites don't cover it either, since the migrated fixtures now pass the new prop.

The new core-store tests cover `createCompositeStore({ includesBaseElement })` and `store.setState("includesBaseElement", ...)`, but nothing covers the React prop plumbing, which is where consumers actually pass the deprecated option.

The behavior does work today; I confirmed it with a throwaway React test: `useCompositeStore({ includesBaseElement: true, focusLoop: true })` sets both `compositeElementInFocusOrder` and `includesBaseElement` to `true`, `ArrowRight` past the last item lands on the composite element with `activeId === null`, and toggling the prop to `false` propagates to both keys. That behavior just isn't pinned by anything in the repo.

Per `ariakit-ariakit-workflow` this belongs in an app fixture rather than a package-source React test. The cheapest option is to leave one fixture on the deprecated prop instead of migrating it, and `app/src/sandbox/combobox-tabs-old/` is a natural home given what its name already signals.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because a silent break removes the deprecated option's effect entirely, and `includesBaseElement={false}` is load-bearing for command-menu style widgets where the active item must not blur when the input is clicked.
- Priority: P2 (estimate, 80% confidence), because the code is correct today and this only guards against future regressions.
- Breakage likelihood: Possible (estimate, 60% confidence), because an alias mapping with no failing test is exactly what a later cleanup deletes once the migration looks complete.
- Fix effort: Easy (verified), because it means not migrating one existing fixture.

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -851,9 +926,17 @@ export interface CompositeStoreOptions<
...
       | "activeId"
     > {
+  /**
+   * Whether the composite element is in the arrow-key focus order.
+   *
+   * @deprecated Use
+   * [`compositeElementInFocusOrder`](https://ariakit.com/reference/composite-provider#compositeelementinfocusorder)
+   * instead.
+   */
+  includesBaseElement?: CompositeStoreState<T>["includesBaseElement"];
```
diegohaz: Redeclaring `includesBaseElement` here duplicates the state JSDoc and ranks the deprecated option above its replacement on the reference pages.

`CompositeStoreState` already declares this member with the same description and the same `@deprecated` message, and `StoreOptions` is just `Partial<Pick<S, K>>`, so adding the key to the pick produces an identical type with no second doc block to keep in sync.

It also fixes ordering in the generated reference. Both extractors emit the most-derived interface's own members first, so on this head `CompositeProvider` and `useCompositeStore` list the deprecated option second while its replacement sits eleven entries lower:

```
setActiveId, includesBaseElement, defaultActiveId, defaultItems, items, id,
virtualFocus, orientation, rtl, focusLoop, focusWrap, focusShift,
compositeElementInFocusOrder, activeId, setItems, store
```

`ComboboxProvider` already reads the way you'd want, `compositeElementInFocusOrder` immediately followed by `includesBaseElement`, precisely because there the deprecated key is inherited instead of redeclared.

```suggestion
      | "compositeElementInFocusOrder"
      | "includesBaseElement"
      | "activeId"
    > {
```

I applied this locally: `pnpm tsc` passes, and `composite-store.test.ts`, `combobox-store.test.ts`, `jsdoc-loader.test.ts`, and `reference-utils.test.ts` all pass. The legacy extractor then emits `..., focusShift, compositeElementInFocusOrder, includesBaseElement (deprecated), activeId, ...` with the description and the deprecation message unchanged, since both now resolve through `CompositeStoreState`.

**Assessment**

- Severity: Low (verified), because the effect is limited to reference-page ordering and a duplicated doc block that can drift.
- Priority: P2 (estimate, 85% confidence), because it is cheapest to fix while the declaration is being introduced.
- Breakage likelihood: Likely (verified), in the sense that the ordering is deterministic and every reader of the Composite reference pages sees it.
- Fix effort: Easy (verified), because the suggestion above is the entire change.

.changeset/200-composite-element-state.md
```diff
@@ -0,0 +1,20 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Composite element state
+
+Composite stores now expose the [`compositeElement`](https://ariakit.com/reference/use-composite-store#compositeelement) state and [`setCompositeElement`](https://ariakit.com/reference/use-composite-store#setcompositeelement) method. The new [`compositeElementInFocusOrder`](https://ariakit.com/reference/composite-provider#compositeelementinfocusorder) option controls whether arrow key navigation can move focus to the composite element:
```
diegohaz: The `#compositeelement` anchor won't exist, so this link lands at the top of the page instead of the state member.

Store state members are not rendered as anchored items on `use-composite-store`. On the deployed page the only `*element*` ids are the setters:

```
$ curl -s https://ariakit.com/reference/use-composite-store | grep -o 'id="[a-z0-9-]*"'
... id="includesbaseelement" ... id="setbaseelement" ...
```

There is no `id="baseelement"`, and none of the other element state members are anchored either (`use-combobox-store` only has `setanchorelement`, `setarrowelement`, `setbaseelement`, `setcontentelement`, `setdisclosureelement`, `setpopoverelement`). The new app extractor won't produce it either: `getReferenceItemId` in `app/src/lib/reference.ts` builds state anchors as `` `${kind}-${slugify(name)}` ``, so the id there would be `state-composite-element`.

The two other links in this paragraph are fine, and I checked both formats against the live site: `#setcompositeelement` matches the existing `setbaseelement` return-prop anchor, and `#compositeelementinfocusorder` matches the existing `includesbaseelement` prop anchor on `composite-provider`. Only the state link is affected, so dropping the anchor is enough:

```suggestion
Composite stores now expose the [`compositeElement`](https://ariakit.com/reference/use-composite-store) state and [`setCompositeElement`](https://ariakit.com/reference/use-composite-store#setcompositeelement) method. The new [`compositeElementInFocusOrder`](https://ariakit.com/reference/composite-provider#compositeelementinfocusorder) option controls whether arrow key navigation can move focus to the composite element:
```

Worth noting the same reasoning supports the bare `` `compositeElement` `` in the `@deprecated` tag on `baseElement` in `composite-store.ts`: there is no reference anchor to point at, so leaving it as inline code is the right call even though the sibling deprecations are linked.

**Assessment**

- Severity: Low (verified), because the link still reaches the correct page, just not the member.
- Priority: P2 (estimate, 90% confidence), because it is a one-token edit to a changelog entry that has not shipped yet.
- Breakage likelihood: Likely (verified), because the anchor is deterministic and absent in both generators.
- Fix effort: Easy (verified), because the suggestion above is the whole change.

diegohaz (commented):
Initial review of `7c97e215d`, diffed from the merge base `fdcf2038a` across all 42 changed files, then expanded into the surrounding store, React, reference-generator, and fixture code.

## Scope and method

- Read the whole diff, then `packages/ariakit-store/src/index.ts` (`storeInit` ordering, `sync`/`setup`/`batch` semantics, parent fan-out), `packages/ariakit-react-store/src/index.tsx` (`useStoreProps`), and every composite-derived store (`combobox`, `menu`, `menubar`, `radio`, `select`, `tab`, `tag`, `toolbar`) to check that the alias handling reaches all of them and that no `pick`/`omit` key list was missed. `select-store.ts` and `tab-store.ts` are the only ones that needed new keys, and both were updated.
- `grep` across the repo confirms the migration is complete: outside the store implementations, their tests, and the two reference-extractor tests, nothing still references `baseElement` or `includesBaseElement`.
- Checked the changeset against `@ariakit/react`'s public entrypoints. The widget list (Combobox, Menu, Menubar, Radio, Select, Tab, Toolbar) is exactly the set of composite-derived stores that package exports; `Tag` is correctly omitted since it isn't re-exported there. All linked reference pages return 200.

## Commands and experiments

- `pnpm test` twice on unmodified head-adjacent trees: 248 files, 1670 tests, green.
- `pnpm tsc`, plus focused runs of `composite-store.test.ts`, `combobox-store.test.ts`, `app/src/lib/jsdoc-loader.test.ts`, and `website/build-pages/reference-utils.test.ts`.
- Dumped both reference extractors' output for `CompositeProvider`, `useCompositeStore`, `ComboboxProvider`, and `useComboboxStore` to inspect prop ordering, descriptions, and deprecation placement rather than relying on the assertions alone.
- Store-level probes for alias precedence, connected stores, pre-init writes, function-updater setters, and `next()` option precedence, run against both `origin/main` and this head for comparison.
- A throwaway React test exercising `includesBaseElement` as a prop, both at mount and on update.
- Prototyped and validated the fixes proposed in the two suggestion-bearing comments below.

## Findings

Four inline comments, no blocking issues:

1. `composite-store.ts` alias mirrors overwrite a connected store's newer deprecated value at `init` and push the stale value back out to that store. Reproduced against `main` and this head; the in-version store shapes take a different, unaffected code path.
2. `useCompositeStoreProps`'s deprecated-prop fallback has no coverage. Deleting it entirely leaves the full Vitest suite green.
3. `CompositeStoreOptions` redeclares `includesBaseElement` instead of picking it, which duplicates the state JSDoc and lists the deprecated option second on the Composite reference pages while its replacement sits eleven entries lower.
4. The changeset links a `#compositeelement` anchor that neither reference generator produces.

## Candidates reviewed and not published

- **Pre-init alias divergence on uninitialized core stores** (threads r3698511772 and r3698511778, both resolved, both already adjudicated by the author as unsupported). Validated independently rather than taken on trust: on this head `createComboboxStore().setInputValue("abc")` leaves `value` at `""` until `init`, and `setState("includesBaseElement", true)` before `init` is reverted rather than merely left diverged. Both converge correctly once initialized. Recording the outcome only; not re-litigating a decision already made on those threads.
- **Double state writes per alias change.** Every `compositeElement` or `compositeElementInFocusOrder` write now dispatches a second `setState` for its alias. Not a finding: both keys change at most a handful of times per widget lifetime, `runListeners` short-circuits on the empty keyed-listener bucket now that no internal code subscribes to `baseElement`, and the extra `defaultValue` call in `getNextId` is one rest-array allocation per keypress. Disposition: declined non-goal.
- **A separate `290-` soft-deprecation changeset.** Considered and not raised: `ariakit-ariakit-deprecation` limits soft deprecations to the `@deprecated` tag and its migration message, and explicitly says not to add a changeset for them. Disposition: invalid.
- **Contributor credit.** No issue, reporter, or external contributor is involved, and bot reviewers are excluded, so no `Thanks to` sentence or `Co-authored-by` trailer is required.
- **Stray whitespace.** The blank line between the import block and `type Orientation` in `composite-store.ts` was dropped. Cosmetic only, lint is green. Disposition: declined non-goal.
- **OG image churn.** `app/public/og-image/react_components_disclosure.png` was rewritten twice on this branch (`a3d7ec72e` 49636 → 48144 bytes, then `2844be956` 48144 → 50381 bytes) with no Disclosure-related source change, while on `main` that directory has only ever changed in two commits. That points at nondeterminism in the OG generator rather than anything in this PR, and it leaves an unrelated binary diff here. It exists independently of this change, so it is a follow-up rather than PR work; I have not opened an issue for it yet and no matching open issue exists.

## Not fully validated

- Playwright suites were not run locally. CI's Chrome, Safari (2/2), iOS, Android, Visual, and Visual (Safari) jobs passed on this head; Firefox, Safari (1/2), and three performance jobs were still pending when this review was written.
- The rendered reference pages were inspected through both extractors' output rather than a local `build-website-lite` render. CI's Main / Build website and Docs / Generate docs both passed.
- Visual snapshots were not touched, per the repository's CI-only policy.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`9fc9a998c`](https://github.com/ariakit/ariakit/commit/9fc9a998c4099dcecf1d1cd602e8c487f611b8c5). The setup now tracks the last mirrored values and adopts a legacy-only connected store's changed `baseElement`/`includesBaseElement` before installing the two-way mirrors:

```ts
if (state.baseElement !== syncedCompositeElement) {
  composite.setState("compositeElement", state.baseElement);
}
```

Added a regression test for the exact create → parent update → init sequence; it fails against the previous mirror ordering and passes with the fix.

packages/ariakit-react-components/src/composite/composite-store.ts
diegohaz: Fixed in [`9fc9a998c`](https://github.com/ariakit/ariakit/commit/9fc9a998c4099dcecf1d1cd602e8c487f611b8c5). I kept `combobox-tabs-old` on `includesBaseElement` and added browser-primary plus happy-dom coverage that toggles the deprecated prop after mount:

```tsx
<CompositeProvider
  focusLoop
  includesBaseElement={includesCompositeElement}
>
```

The test focuses the last item, presses ArrowRight, and expects focus on the composite element. It fails if the `?? props.includesBaseElement` fallback is removed, and passes in Chrome, Firefox, Safari, React 19, and React 18.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`9fc9a998c`](https://github.com/ariakit/ariakit/commit/9fc9a998c4099dcecf1d1cd602e8c487f611b8c5). I removed the duplicate property/JSDoc and now select it through `StoreOptions`:

```ts
| "compositeElementInFocusOrder"
| "includesBaseElement"
| "activeId"
```

Both reference extractor suites now assert that the replacement is listed before the deprecated option while preserving inherited deprecation metadata.

.changeset/200-composite-element-state.md
diegohaz: Fixed in [`9fc9a998c`](https://github.com/ariakit/ariakit/commit/9fc9a998c4099dcecf1d1cd602e8c487f611b8c5). The changeset now links `compositeElement` to the `use-composite-store` page without the nonexistent `#compositeelement` anchor:

```md
[`compositeElement`](https://ariakit.com/reference/use-composite-store)
```

diegohaz:
Addressed the four review findings in [`9fc9a998c`](https://github.com/ariakit/ariakit/commit/9fc9a998c4099dcecf1d1cd602e8c487f611b8c5):

- preserve newer legacy-only connected-store values across initialization;
- add post-mount `includesBaseElement` React coverage in happy-dom and Chrome/Firefox/Safari;
- inherit `includesBaseElement` option metadata and pin replacement-before-deprecated ordering in both extractors;
- remove the invalid `#compositeelement` changeset anchor.

Validation: `pnpm lint-fix`, `pnpm build`/`pnpm clean`, `pnpm tsc`, 75 focused Vitest tests, React 18 focused tests, `pnpm build-website-lite`, `pnpm build-app-lite`, and the focused Playwright test in Chrome, Firefox, and Safari.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Reassessed against `9fc9a998c`: the tracking works, but I now agree it isn't worth keeping. Reverting it is the right call.

The claim in my original comment is still technically accurate, and `preserves legacy connected store updates before initialization` does fail without the tracking. What I under-weighted is how narrow the trigger is. The clobber only happens when the connected store carries `baseElement` **without** `compositeElement`. Any store that carries both makes `storeInit` take its per-key `sync` path (`keys.length === stateKeys.length`), which pushes in `getKeys(initialState)` order, where `compositeElement` precedes `baseElement`, so nothing diverges. Every store this version can produce carries both, and `createStore` is not re-exported by any public `@ariakit/react` or `@ariakit/solid` entrypoint, so building a deprecated-only store means reaching into `@ariakit/store` directly or sharing a store across majors. That is not a use case worth carrying bookkeeping for.

Reverting is cheap and self-contained. I restored `composite-store.ts` to `7c97e215d` and ran the file's suite: only `preserves legacy connected store updates before initialization` fails (1 failed, 22 passed). Both `syncs deprecated aliases with connected stores` and `preserves a base element set before initialization` still pass, so the post-init two-way sync and the pre-init ref case stay covered.

One thing to be deliberate about: without the tracking the composite store doesn't merely diverge from such a store, it writes its own creation-time value back into it at `init`. If that's accepted as out of contract, it's worth a sentence somewhere durable, so the next reader doesn't rediscover it as a bug and re-add the bookkeeping.

On the combobox side there's nothing equivalent to remove. `syncedSelectElement` and `syncedAnchorElement` are runtime ownership tracking, not pre-init compatibility: they record whether the current value is the fallback the listener assigned, so an explicit anchor or select keeps ownership afterwards. I confirmed that by dropping the ownership branch from the anchor listener and always assigning the fallback, which fails three runtime tests: `syncs the combobox element with the anchor element`, `preserves an anchor that later becomes the disclosure`, and `preserves an explicit anchor matching a previous fallback`. Combobox's `inputValue`/`value` mirrors never had this tracking, so they are already in the state a revert would produce.

Disposition: declined non-goal. The behavior difference is real but the scenario is unreachable for consumers, so it does not justify the machinery.

app/src/lib/jsdoc-loader.test.ts
```diff
@@ -287,6 +307,47 @@ test("loads Combobox input value metadata", async () => {
...
+  );
+
+  const store = getReference(entries, "react/composite/use-composite-store");
+  for (const reference of [provider, store]) {
+    const names = getParamProps(reference).map((prop) => prop.name);
+    const compositeElementIndex = names.indexOf("compositeElementInFocusOrder");
+    const includesBaseElementIndex = names.indexOf("includesBaseElement");
+    expect(compositeElementIndex).toBeGreaterThanOrEqual(0);
+    expect(includesBaseElementIndex).toBeGreaterThan(compositeElementIndex);
+  }
```
diegohaz: This ordering assertion can't fail, so it doesn't guard the change it was added for.

The app extractor sorts props deprecated-last, then alphabetically (`app/src/lib/jsdoc-loader.ts:1137-1141`):

```ts
return props.sort((a, b) => {
  if (a.deprecated && !b.deprecated) return 1;
  if (!a.deprecated && b.deprecated) return -1;
  return a.name.localeCompare(b.name);
});
```

So `includesBaseElement` lands after `compositeElementInFocusOrder` no matter how `CompositeStoreOptions` declares the key, and it would still land there even if the `@deprecated` tag were dropped, since `c` sorts before `i` anyway. I dumped the param names against `7c97e215d`, before the pick change, and got the identical order both times:

```
activeId, compositeElementInFocusOrder, defaultActiveId, defaultItems, focusLoop,
focusShift, focusWrap, id, items, orientation, rtl, setActiveId, setItems, store,
virtualFocus, includesBaseElement
```

The companion assertion in `website/build-pages/reference-utils.test.ts` is the real guard: reverting the same file to `7c97e215d` fails it with `expected 1 to be greater than 6`, because the legacy extractor is the one that orders by declaration site, and it is what ariakit.com currently serves. That also corrects something I got wrong in my earlier comment, where I said both extractors order that way.

Dropping this loop keeps the file honest; the deprecation metadata for this param is already asserted a few lines above via `includesBaseElement.deprecated`. If you'd rather keep an app-side check, it needs to assert something the sort doesn't already guarantee.

**Assessment**

- Severity: Low (verified), because the assertion is inert rather than wrong, and the behavior it describes is genuinely guaranteed by the extractor.
- Priority: P2 (estimate, 80% confidence), because it only misrepresents where the coverage lives.
- Breakage likelihood: Rare (verified), because an inert assertion cannot fail; the cost is a future reader trusting it.
- Fix effort: Easy (verified), because it is a deletion.

diegohaz (commented):
Re-review of `9fc9a998c`. Baseline is `7c97e215d`, the head recorded in the previous marked review, so the incremental diff is the single `Address composite alias review feedback` commit. Newly discovered findings come from those changes plus the surrounding code they expose; the one unresolved thread from the previous run was independently revalidated against this head rather than taken as fixed.

## Previous findings

- **Alias mirrors clobber a connected store at `init`** (r3698606849, unresolved). Revalidated: the tracking works and `preserves legacy connected store updates before initialization` is a genuine red/green regression test, failing on `7c97e215d` and passing here. On reassessment I no longer think the machinery is worth keeping, and replied on the thread with the reasoning and the combobox comparison the question there asked for. Disposition changed to declined non-goal.
- **Deprecated React prop had no coverage.** Fixed and verified as a real regression guard. Collapsing `useCompositeStoreProps` to `useStoreProps(store, props, "compositeElementInFocusOrder")`, dropping the deprecated prop entirely, now fails `updates focus order through the deprecated prop` and nothing else across the full suite (1 failed, 1671 passed). The new `composite-navigation` section is a good shape for it: the checkbox exercises the update path rather than just the mount path, `tabIndex={0}` on the control is right for Safari's click-focus behavior per `ariakit-ariakit-example-playwright-test`, and both the browser test and the happy-dom duplicate drive it through accessible queries. Reverting `combobox-tabs-old` additionally keeps the deprecated prop exercised through a derived store (`TabProvider`), which the new fixture doesn't cover.
- **`CompositeStoreOptions` redeclaring `includesBaseElement`.** Fixed. Verified the legacy extractor now emits `..., focusShift, compositeElementInFocusOrder, includesBaseElement, activeId, ...` with the description and deprecation message intact, and that the new website-side ordering assertion fails against `7c97e215d` with `expected 1 to be greater than 6`.
- **Changeset anchor.** Fixed; the remaining two anchors match live ids on `use-composite-store` and `composite-provider`.

## New finding

One inline comment: the ordering assertion added to `app/src/lib/jsdoc-loader.test.ts` is inert. The app extractor sorts props deprecated-last then alphabetically, so it passes identically against `7c97e215d`; the website-side assertion is the one that actually guards the change. That also corrects a claim in my previous review: only the legacy extractor orders by declaration site, and it is the one ariakit.com currently serves.

## Commands and experiments

- `pnpm test` on the full suite, and focused runs of `composite-store.test.ts`, `combobox-store.test.ts`, `jsdoc-loader.test.ts`, `reference-utils.test.ts`, and `app/src/sandbox/composite-navigation/test.ts`.
- Red checks for each new test by restoring the relevant file to `7c97e215d`, or by removing the code under test, and confirming the expected failure and only that failure.
- Dumped the app extractor's param ordering for `CompositeProvider` and `useCompositeStore` against both heads to check whether the new app-side assertion discriminates.
- Removed combobox's anchor-ownership branch to confirm `syncedAnchorElement` is runtime state rather than pre-init compatibility.
- Restored every experiment; the worktree matches `9fc9a998c` exactly.

## Not fully validated

- Playwright suites were not run locally. All of them passed on this head in CI, including the new `composite-navigation/test-browser.ts` across Chrome, Firefox, both Safari shards, iOS, and Android.
- Type checking and the website build were taken from CI's green Main / Type check, Main / Build website, and Docs / Generate docs jobs.
- The OG generator churn noted in the previous review did not recur on this commit, so there is still no follow-up issue for it.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`466c16baa`](https://github.com/ariakit/ariakit/commit/466c16baa9e207053dab4b9a74a149bd3ca85af7): removed the tracked values, pre-init reconciliation branches, and dedicated regression contract for the intentionally unsupported deprecated-only connected-store timing. The supported boundary is now recorded next to the setup listeners:

```ts
// Alias synchronization starts at initialization. Writes made before then by
// connected stores with only deprecated keys are intentionally unsupported.
```

Initialized connected and derived stores still synchronize both aliases, and the deprecated `setBaseElement` pre-init ref path remains covered.

app/src/lib/jsdoc-loader.test.ts
diegohaz: Fixed in [`466c16baa`](https://github.com/ariakit/ariakit/commit/466c16baa9e207053dab4b9a74a149bd3ca85af7): removed the inert modern ordering loop and restored the direct `getParamProp` helper. The app suite still verifies the meaningful contract:

```ts
expect(includesBaseElement.deprecated).toEqual(
  expect.stringContaining("compositeElementInFocusOrder"),
);
```

The declaration-order-sensitive assertion remains in the legacy extractor suite.

diegohaz:
**Fixed**

- [`466c16baa`](https://github.com/ariakit/ariakit/commit/466c16baa9e207053dab4b9a74a149bd3ca85af7) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698897734)): removed deprecated-only pre-init connected-store tracking and its dedicated test, while documenting the intentionally unsupported lifecycle window and preserving initialized synchronization plus the `setBaseElement` ref path.
- [`466c16baa`](https://github.com/ariakit/ariakit/commit/466c16baa9e207053dab4b9a74a149bd3ca85af7) ([reply](https://github.com/ariakit/ariakit/pull/7008#discussion_r3698898485)): removed the inert modern reference-order assertion and helper, while retaining deprecation coverage and the declaration-order-sensitive legacy extractor assertion.

Validation: `pnpm lint-fix`, 61 focused Vitest tests, `pnpm build`, generated declaration inspection, `pnpm clean`, `pnpm tsc`, and `pnpm build-website-lite`.

diegohaz (commented):
Re-review of `f70b3bd76`. No new findings, and both open items from the previous round are resolved.

## Baseline

The branch was rebased since the last marked review, so the recorded baseline `9fc9a998c` is no longer on it. I established the correspondence with `git range-diff` instead of diffing across the rebase: `9fc9a998c` maps to `852a74db1`, and the only content delta the rebase introduced is the `composite-navigation` fixture adapting to `b6368e942`, which added `DistantComposite` and the sandbox's first `test-browser.ts` on `main`. Everything else in the PR is byte-identical to what was already reviewed. The genuinely new work is `f70b3bd76`.

## Previous findings

- **Alias mirrors clobber a connected store at `init`.** Reverted, as agreed on that thread, and the comment above the `setup` block now records the boundary explicitly: alias synchronization starts at initialization, and pre-init writes from connected stores carrying only the deprecated keys are unsupported. That is the durable note I asked for, so the next reader doesn't rediscover it as a bug. The two remaining connected-store tests still pass, so post-init two-way sync and the pre-init ref case stay covered.
- **Inert app-side ordering assertion.** Removed, along with the now-unused `getParamProps` helper. The dual-generator split still holds: the website test keeps the ordering guard, and the app test keeps the metadata it is uniquely able to check, since state props only exist in that extractor.

## Verification

- `pnpm test`: 249 files, 1672 tests, green. All 48 CI checks pass on this head (Build styles skipped as usual).
- Re-ran the red checks for both remaining new guards on this head, since a rebase can quietly invalidate a regression test:
  - Collapsing `useCompositeStoreProps` to `useStoreProps(store, props, "compositeElementInFocusOrder")` fails `updates focus order through the deprecated prop` and nothing else (1 failed, 1671 passed).
  - Restoring the redeclared `includesBaseElement` option fails the website ordering assertion with `expected 1 to be greater than 6`.
- Checked the fixture merge for interference with the coverage `main` just added. `LegacyFocusOrderComposite` sits before `DistantComposite`, so the first tab stop stays in `BasicComposite` as that fixture's comment requires, and pushing `DistantComposite` further down the page only strengthens the `not.toBeInViewport` precondition in the scroll test.
- Swept for leftovers: the only remaining `includesBaseElement` usages outside the store implementations and the reference tests are the two deliberate fixture usages in `combobox-tabs-old` and `composite-navigation`, and nothing on the newer `main` reintroduced `baseElement` for this PR to migrate.
- Every experiment was restored; the worktree matches `f70b3bd76` exactly.

## Considered and not raised

- `q.button("Legacy two").focus()` in the browser test as a setup step rather than a keyboard path. `ariakit-ariakit-example-playwright-test` only warns against swapping `.click()` for `.focus()` to dodge Safari's click-focus behavior, which is not what is happening here, and the sibling scroll test on `main` opens with `near.focus()` the same way. Disposition: invalid.
- Keeping `combobox-tabs-old` on the deprecated prop now that a dedicated fixture exists. It is not redundant: it is the only place the deprecated prop reaches a derived store (`TabProvider`) rather than `CompositeProvider`. Disposition: declined non-goal.

## Not fully validated

- Playwright suites were not run locally. All six projects passed on this head in CI, including the new `composite-navigation` browser test.
- Type checking and the website build were taken from CI's green Main / Type check, Main / Build website, and Docs / Generate docs jobs.
- The OG generator churn from the first review has not recurred; there is still no follow-up issue for it.

Issue #7020, opened 2026-08-02
Composite: item-owned presentation handoff never sees a focus escape


diegohaz:
Follow-up from https://github.com/ariakit/ariakit/pull/7009, raised on `composite/composite-item.tsx:354`.

The presentation `CompositeItem` starts when it hands DOM focus back to the composite element passes neither `focus` nor `requireFocus`, so `owner` is `null` in `presentItem` and `stillOwnsFocus` returns `true` unconditionally. Of the four abandon reasons the PR establishes, this request honours three: popup closed, active item changed, item disconnected. Focus escape is the one it does not see, and it is the only presentation in the change that does not.

The consequence is only user-visible when the item's nearest scrollport is not the popup, which is the case for a popup that does not scroll internally and sits below the fold. Then a request that outlives a focus escape moves the page after the user has already left.

The existing `cancels presentation when virtual focus leaves the popup` test in `combobox-select-selected-value-presentation` does not cover this. It asserts `scrollY` and the absence of a `document` scroll event, and a surviving item-owned request scrolls the item's nearest scrollport, which for an in-popup option is the popup's own list. That test pins the composite-owned request, which does pass `requireFocus: true`.

## Why it was not simply fixed

This request exists to finish a handoff that moves focus twice on its own, so a naive focus check abandons it every time: `owner` would be the item, focus then moves to the composite element, and with `focus: false` the `stillOwnsFocus` fast paths do not apply. Passing `requireFocus: true` together with `focus: false` is therefore not the right shape.

Deciding the cancellation policy for a deliberately focus-agnostic request is a design call. The exception is documented at the call site in the meantime.

Related: the focus-ownership contract question in #7018, since a narrower notion of ownership would change what a focus check here could mean.

PR #7048, opened 2026-08-03
Prevent Dialog delayed auto-focus from reclaiming escaped focus

### LINKED ISSUE #7033: Dialog: auto focus on show pulls focus back after the user has already left
Found while fixing #7020. `Dialog`'s auto-focus-on-show waits for the popover to be positioned and then focuses its target without rechecking where focus went in the meantime, so an app that moves focus out of the widget during the placing window has that focus move silently undone.

This is the observation the [triage on #7020](https://github.com/ariakit/ariakit/issues/7020#issuecomment-5160252491) recorded but did not file, now verified against the ordinary sub-frame placing window rather than only against a slow `updatePosition`.

## Mechanism

`autoFocusOnShow` is gated on `positioned`, and the effect queues a microtask that scrolls and focuses:

```jsx
if (isFocusable(element)) {
  element.scrollIntoView({ block: "nearest", inline: "nearest" });
}
element.focus({ preventScroll: true });
```

Nothing between scheduling and running asks whether focus is still where it was when the dialog opened.

## Reproduction

Instrumented against a `ComboboxSelect` whose popup stays open (`hideOnInteractOutside={false}`) and whose focus escapes to a page-level button from a handler during the open. Timings are from one Firefox run, in milliseconds:

```
t=2346  composite presentation focuses the marked item
t=2347  the item hands focus back to the combobox input
t=2354  the app moves focus to a button outside the widget
t=2358  both pending presentations correctly abandon, focus is outside
t=2366  Dialog auto focus on show runs and focuses the input again
```

At `t=2366` focus is pulled back into a popup the user has already left. In that fixture it also restarts the presentation chain, which moves the page several hundred pixels, but the focus theft is the defect on its own: it happens whether or not anything scrolls afterwards.

## What has to be true

1. A popup whose `autoFocusOnShow` is active, which is the default for `ComboboxSelect` and `Select`.
2. The popup stays open while focus leaves, through `hideOnInteractOutside={false}` or an `open` the app keeps true.
3. Focus leaves after the open but before the auto focus runs, so the escape has to be programmatic at the default placing window. A custom async `updatePosition` widens that window until an ordinary user interaction fits inside it.

## Workaround

Set [`autoFocusOnShow`](https://ariakit.com/reference/combobox-popover#autofocusonshow) to `false` and move focus yourself, or pass a callback that returns `false` once focus has already left the widget:

```jsx
<Ariakit.ComboboxPopover autoFocusOnShow={false} hideOnInteractOutside={false}>
```

The cost is that the popup no longer takes focus on open at all, so a widget that relies on that has to do it explicitly.

## Expected outcome

Auto focus on show should not move focus when focus has already moved somewhere the dialog does not own since the open that scheduled it. Ordinary opens, where focus is still on the disclosure or wherever it was, must keep working exactly as they do now.

diegohaz:
## Motivation

`Dialog` queues its initial auto-focus until the popup has been positioned. If focus leaves an open popup while it is waiting, the delayed callback can reclaim focus and create an unexpected focus loop in components such as `ComboboxPopover`.

Fixes #7033.

## Solution

`Dialog` now records whether the current store's dialog has received focus. React focus capture covers native `autoFocus` before layout effects, while a native `focusin` listener covers later focus moves and non-React descendants. The delayed callback suppresses auto-focus only when that exact store has received focus and focus is now on another focusable destination outside the current dialog, disclosure, and owned tree.

The callback reads the current `open` and `disclosureElement` state when its microtask runs. Active-element ownership starts in the Dialog document, where focus inside a same-origin iframe remains represented by the containing `iframe`. Existing open-shadow traversal then uses the deepest focused node for focusability and checks ownership against both the document's shallow active host and that deepest node. This uses the narrow dialog, disclosure, and owned-tree predicate; unlike outside-interaction handling, it intentionally does not classify arbitrary `data-focus-trap` elements as owned. Outside-interaction behavior keeps a separate content-scoped focus flag because replacing the rendered dialog element must reset that state without erasing the store's auto-focus history.

The `combobox-select-selected-value-presentation` sandbox includes visible histories for the original focus escape, an external focus-trap escape, native auto-focus into a nested open shadow root, a dialog mounted inside an open shadow root, focus entering a contained iframe, store replacement, and disclosure replacement. For `Focus escaping fruit` and `Focus trap escaping fruit`, the broken behavior records `input → focus target → input → focus target`; the fixed behavior ends at `input → focus target` and leaves the popup visible. Their history panels now stay in normal flow at a readable width, so each result remains attributable to its fixture. The contained-iframe dialog separately shows `dialog focus → iframe focus → initial focus`. Browser tests cover placement and browsing-context behavior, while reliable store, disclosure, content-replacement, and shadow-root dialog cases are duplicated in happy-dom for React 18 coverage.

<details>
<summary>Cycle 3: Keep store focus history separate from outside-interaction state</summary>

**Assessment**

The original active-element-only guard was too broad and suppressed legitimate initial focus in Menu and Combobox compositions. The required distinction is whether this exact `DialogStore` has already received focus, not merely whether some focusable element is active outside. Outside-interaction listeners have a different content-element reset boundary, so their focused state cannot safely stand in for store history.

**Decision**

Continue with one store-valued history ref for delayed auto-focus and one local content-scoped boolean for outside-interaction behavior. Keep the open-shadow traversal local to `Dialog` because no shared utility needs the broader behavior. This adds no public API and preserves existing content-replacement semantics.

</details>

<details>
<summary>Cycle 9: Continue store-scoped history with document-local ownership</summary>

**Assessment**

The original bug has moderate overall severity and high accessibility impact when it occurs: delayed auto-focus can reclaim focus after an application or user has intentionally left an open popup. Its expected frequency remains timing-dependent, but it affects stable Dialog behavior used by Select and ComboboxSelect and has been reproduced in Chrome and Firefox.

The current implementation has meaningful but bounded private complexity: one exact-store focus-history ref, React and native focus observation paths, and a local ownership check. Removing exact-store history was tested and regressed existing Menu and Combobox behavior, so that state remains load-bearing.

The confirmed same-origin iframe regression is narrower and likely less frequent than the original issue, but it silently drops the documented `initialFocus` behavior and initial scroll relative to the merge base. `getActiveElement` follows same-origin frames, while ownership must be decided in the Dialog's own document. Reading that document's active element leaves a focused frame represented by its containing `iframe`, after which the existing open-shadow traversal and shallow/deep ownership checks remain valid. This replaces an over-broad lookup and adds no state, branch, helper, public API, or shared abstraction.

The test and sandbox footprint is now the larger maintenance concern. The specialized regressions protect distinct decisions, and the two original focus-history panels now stay in normal flow so they cannot overlap during manual verification.

**Decision**

Continue the exact-store design. Address the confirmed iframe regression by resolving the active element from the Dialog's document, retaining both shallow-host and deepest-open-shadow ownership checks, and adding one focused browser regression.

Keep the auto-focus ownership predicate narrow: an arbitrary external `data-focus-trap` remains a legitimate escape. Do not replace exact-store history with an active-element-only guard.

Keep each original focus-history panel in normal flow so a browser user can identify the bug as `input → focus target → input` instead of the expected `input → focus target` without another fixture covering the result.

This is a local completion of the current direction, not a redesign, reversion, or support narrowing, so the implementation-cycle count does not reset.

**Intentionally unsupported**

Closed-shadow internals; null, body, disconnected, or non-focusable active targets; outside focus before the exact current store's dialog has received focus; arbitrary external focus traps as owned elements; and cross-browsing-context focus transitions where the Dialog itself lives in a child frame remain unsupported without a confirmed reproduction. Frame documents are not traversed for ownership; the containing `iframe` in the Dialog's document is the supported boundary.

</details>

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the narrow store-scoped focus design</summary>

**Assessment**

The bug has high accessibility impact when it occurs because focus can be reclaimed after the user or application has intentionally moved it. Its expected frequency is lower because it requires delayed placement to overlap a focus transition. Review rounds identified pre-layout native auto-focus, store replacement, nested open shadow roots, and timing-proof requirements. These are distinct supported focus paths rather than speculative branches.

**Decision**

Continue exact-store focus history and local open-shadow focusability handling. Do not broaden shared DOM or outside-interaction machinery. Closed shadow roots, body or null active targets, disconnected or non-focusable destinations, and outside focus before the current store's dialog receives focus remain intentionally unsupported because they do not provide a reliable user-selected focus destination.

</details>

<details>
<summary>Cycle 6: Continue store-scoped focus history with local shadow handling</summary>

**Assessment**

Overall severity is moderate, with high accessibility disruption per occurrence and timing-dependent frequency. The production change remains private and modest: one store-valued ref, one existing content-scoped boolean with its original reset semantics, and one local shadow traversal helper. No public API or shared DOM abstraction is added. Later review findings focused on browser authority, React 18 duplication, internal call shape, dead guards, and provenance, which shows convergence rather than a failing implementation direction.

**Decision**

Keep the current design. Use an object parameter for the expanded internal hook, assign required refs unconditionally, retain both browser and happy-dom coverage where the behavior is reliable in each environment, and preserve the same intentionally unsupported cases listed above. Reverting to an active-element guard, using a store-independent boolean, or sharing one ref across both reset boundaries would reintroduce confirmed regressions.

</details>

## Workaround

For consumers pinned to a release before this fix, disable the delayed auto-focus for this composition:

```tsx
<ComboboxPopover
  // TODO: Remove after https://github.com/ariakit/ariakit/issues/7033 is fixed.
  autoFocusOnShow={false}
/>
```

This keeps focus on the escape target but opts out of Dialog initial focus behavior for the popup.

## Preview

The [original focus-escape evidence](https://github.com/ariakit/ariakit/pull/7048#issuecomment-5167123335) and [shadow-root follow-up](https://github.com/ariakit/ariakit/pull/7048#issuecomment-5177509096) include screenshots and short recordings of the visible focus-history panels.

The latest sandbox keeps the original focus history in a readable panel next to its own fixture. After opening `Focus escaping fruit`, the fixed history ends at `input → focus target`:

Focus escaping fruit fixture showing the fixed input to focus target history in its own non-overlapping panel

### DIFF
--- a/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
+++ b/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
@@ -152,7 +152,71 @@ withFramework(import.meta.dirname, async ({ test }) => {
     // for focus not being stolen once that microtask has run.
     await flushFrames(page);
     await test.expect(escapeTarget).toBeFocused();
+    await test.expect(q.listbox()).toBeVisible();
     await test.expect(focusHistory).toHaveText("input → focus target");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("does not refocus after native auto-focus moves focus into a shadow root", async ({
+    page,
+    q,
+  }) => {
+    const disclosure = q.button("Native auto-focus dialog");
+    const escapeTarget = q.textbox("Native auto-focus shadow target");
+    const focusHistory = q.status("Native auto-focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await disclosure.click();
+
+    // React runs native auto-focus before layout effects. The input's focus
+    // handler moves focus out synchronously, before Dialog can install a DOM
+    // listener to observe that the dialog had already received focus.
+    await test.expect(escapeTarget).toBeFocused();
+    // Dialog's queued auto-focus has no observable completion marker, so cross
+    // its frame checkpoint before asserting that focus remained outside.
+    await flushFrames(page);
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(focusHistory).toHaveText("input → shadow target");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("focuses a dialog after its store changes", async ({ q }) => {
+    const focusHistory = q.status("Store swap focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await q.button("Open store A").click();
+    await test.expect(q.textbox("Store A initial focus")).toBeFocused();
+
+    await q.button("Open store B").click();
+
+    await test.expect(q.textbox("Store B initial focus")).toBeFocused();
+    await test
+      .expect(focusHistory)
+      .toHaveText(
+        "open store A → store A input → open store B → store B input",
+      );
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("uses the current disclosure in delayed auto-focus", async ({
+    page,
+    q,
+  }) => {
+    const previousDisclosure = q.button("Open disclosure swap dialog");
+    const focusHistory = q.status("Disclosure swap focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await previousDisclosure.click();
+
+    await test.expect(previousDisclosure).toBeFocused();
+    // Dialog's queued auto-focus has no observable completion marker, so cross
+    // its frame checkpoint before asserting that focus remained outside.
+    await flushFrames(page);
+    await test.expect(previousDisclosure).toBeFocused();
+    await test.expect(focusHistory).toHaveText("input → previous disclosure");
   });
 
   // This case originated in the unmerged PR below. It covers the same
--- a/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
+++ b/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
@@ -157,6 +157,26 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   // https://github.com/ariakit/ariakit/issues/7033
+  test("does not refocus an external focus trap", async ({ page, q }) => {
+    const select = q.combobox("Focus trap escaping fruit");
+    const escapeTarget = q.button("Focus trap escaping fruit focus target");
+    const focusHistory = q.status("Focus trap escaping fruit focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await select.click();
+
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(q.listbox()).not.toHaveAttribute("data-placing");
+    // Dialog queues auto-focus after placement, and there is no positive state
+    // for focus not being stolen once that microtask has run.
+    await flushFrames(page);
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(q.listbox()).toBeVisible();
+    await test.expect(focusHistory).toHaveText("input → focus target");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
   test("does not refocus after native auto-focus moves focus into a shadow root", async ({
     page,
     q,
@@ -178,6 +198,23 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await flushFrames(page);
     await test.expect(escapeTarget).toBeFocused();
     await test.expect(focusHistory).toHaveText("input → shadow target");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("honors initial focus when the dialog is inside a shadow root", async ({
+    q,
+  }) => {
+    const focusHistory = q.status("Shadow-root dialog focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await q.button("Open shadow-root focus dialog").click();
+
+    await test
+      .expect(q.textbox("Shadow-root initial focus field"))
+      .toBeFocused();
+    await test.expect(q.dialog("Shadow-root focus dialog")).toBeVisible();
+    await test.expect(focusHistory).toHaveText("app focus → initial focus");
   });
 
   // https://github.com/ariakit/ariakit/issues/7033
--- a/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
+++ b/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
@@ -132,6 +132,29 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect(q.option("Watermelon")).not.toBeInViewport();
   });
 
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("does not refocus after focus leaves an open popup", async ({
+    page,
+    q,
+  }) => {
+    const select = q.combobox("Focus escaping fruit");
+    const escapeTarget = q.button("Focus escaping fruit focus target");
+    const focusHistory = q.status("Focus escaping fruit focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await select.click();
+
+    // The app moved focus out of the popup while it was opening.
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(q.listbox()).not.toHaveAttribute("data-placing");
+    // Dialog queues auto-focus after placement, and there is no positive state
+    // for focus not being stolen once that microtask has run.
+    await flushFrames(page);
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(focusHistory).toHaveText("input → focus target");
+  });
+
   // This case originated in the unmerged PR below. It covers the same
   // focus-move cancellation invariant when the popup remounts on open and the
   // select element briefly becomes the composite again.
--- a/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
+++ b/app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
@@ -1,7 +1,7 @@
 import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
 import { recordScrollEvents } from "#app/test-utils/scroll.ts";
 
-withFramework(import.meta.dirname, async ({ test }) => {
+withFramework(import.meta.dirname, async ({ query, test }) => {
   // https://github.com/ariakit/ariakit/pull/6832
   test("focuses a far item reached through typeahead", async ({ page, q }) => {
     const select = q.combobox("Selected fruit");
@@ -215,6 +215,28 @@ withFramework(import.meta.dirname, async ({ test }) => {
       .toBeFocused();
     await test.expect(q.dialog("Shadow-root focus dialog")).toBeVisible();
     await test.expect(focusHistory).toHaveText("app focus → initial focus");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7033
+  test("honors initial focus after focus enters a contained iframe", async ({
+    page,
+    q,
+  }) => {
+    const frame = query(
+      page.frameLocator("iframe[title='Initial focus frame']"),
+    );
+    const focusHistory = q.status("Iframe dialog focus history");
+
+    await test.expect(focusHistory).toHaveText("none");
+
+    await q.button("Open iframe focus dialog").click();
+
+    await test.expect(q.textbox("Iframe initial focus field")).toBeFocused();
+    await test.expect(q.dialog("Iframe focus dialog")).toBeVisible();
+    await test.expect(frame.textbox("Iframe app focus field")).toBeVisible();
+    await test
+      .expect(focusHistory)
+      .toHaveText("dialog focus → iframe focus → initial focus");
   });
 
   // https://github.com/ariakit/ariakit/issues/7033
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -500,7 +500,10 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
       // focusable element outside while the dialog is waiting to be
       // positioned. Preserve that user choice, but allow a dialog that hasn't
       // received focus yet to perform its initial focus move.
-      const activeElement = getActiveElement(element);
+      const documentActiveElement = getDocument(contentElement).activeElement;
+      const activeElement = isElement(documentActiveElement)
+        ? documentActiveElement
+        : null;
       const deepestActiveElement =
         activeElement && getDeepestActiveElement(activeElement);
       if (
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -508,7 +508,12 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
         activeElement &&
         deepestActiveElement &&
         isFocusable(deepestActiveElement) &&
-        !isElementInDialog(activeElement, contentElement, disclosureElement)
+        !isElementInDialog(activeElement, contentElement, disclosureElement) &&
+        !isElementInDialog(
+          deepestActiveElement,
+          contentElement,
+          disclosureElement,
+        )
       ) {
         return;
       }
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -31,6 +31,7 @@ import type {
   ComponentPropsWithRef,
   ElementType,
   FC,
+  FocusEvent as ReactFocusEvent,
   ReactElement,
   KeyboardEvent as ReactKeyboardEvent,
   RefObject,
@@ -87,6 +88,16 @@ function isAlreadyFocusingAnotherElement(dialog?: HTMLElement | null) {
   if (dialog && contains(dialog, activeElement)) return false;
   if (isFocusable(activeElement)) return true;
   return false;
+}
+
+function getDeepestActiveElement(element: Element) {
+  let activeElement = element;
+  while (activeElement.shadowRoot) {
+    const nextElement = activeElement.shadowRoot.activeElement;
+    if (!isElement(nextElement)) break;
+    activeElement = nextElement;
+  }
+  return activeElement;
 }
 
 function isElementInDialog(
@@ -222,18 +233,28 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
   // (e.g., onClose calling event.preventDefault), async closes with
   // animations, or when autoFocusOnHide is disabled.
   const interactedOutsideRef = useRef(false);
+  const focusedStoreRef = useRef<DialogStore | null>(null);
   useSafeLayoutEffect(() => {
     return sync(store, ["open"], (state) => {
-      if (!state.open) return;
+      if (!state.open) {
+        focusedStoreRef.current = null;
+        return;
+      }
       interactedOutsideRef.current = false;
+      // Native auto-focus may run before layout effects on the initial open
+      // render, so preserve focus captured during that commit. Focus history
+      // from another store doesn't belong to this open cycle.
+      if (focusedStoreRef.current === store) return;
+      focusedStoreRef.current = null;
     });
   }, [store]);
-  useHideOnInteractOutside(
+  useHideOnInteractOutside({
     store,
     hideOnInteractOutside,
     domReady,
     interactedOutsideRef,
-  );
+    focusedStoreRef,
+  });
 
   const { wrapElement, nestedDialogs } = useNestedDialogs(store);
   props = useWrapElement(props, wrapElement, [wrapElement]);
@@ -264,6 +285,10 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
   // dialog is opened.
   useSafeLayoutEffect(() => {
     if (!open) return;
+    // Native auto-focus can focus the dialog and synchronously move focus
+    // outside before layout effects run. In that case, the current element is
+    // an escape target rather than the element that disclosed the dialog.
+    if (focusedStoreRef.current === store) return;
     const dialog = ref.current;
     const activeElement = getActiveElement(dialog, true);
     if (!activeElement) return;
@@ -463,21 +488,26 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
       contentElement;
     const isElementFocusable = isFocusable(element);
     if (!autoFocusOnShowProp(isElementFocusable ? element : null)) return;
-    const { disclosureElement } = store.getState();
     setAutoFocusEnabled(true);
     queueMicrotask(() => {
+      const { open, disclosureElement } = store.getState();
       // If the dialog was closed between scheduling and executing this
       // microtask, skip the focus call. Otherwise, focusing a now-hidden
       // element could steal focus from the disclosure that focusOnHide already
       // restored.
-      if (!store.getState().open) return;
-      // Focus may have moved to another focusable element outside the dialog
-      // while it was waiting to be positioned. Preserve that user choice, but
-      // allow focus to remain on the disclosure or another owned element.
+      if (!open) return;
+      // Once the dialog has received focus, focus may move to another
+      // focusable element outside while the dialog is waiting to be
+      // positioned. Preserve that user choice, but allow a dialog that hasn't
+      // received focus yet to perform its initial focus move.
       const activeElement = getActiveElement(element);
+      const deepestActiveElement =
+        activeElement && getDeepestActiveElement(activeElement);
       if (
+        focusedStoreRef.current === store &&
         activeElement &&
-        isFocusable(activeElement) &&
+        deepestActiveElement &&
+        isFocusable(deepestActiveElement) &&
         !isElementInDialog(activeElement, contentElement, disclosureElement)
       ) {
         return;
@@ -506,6 +536,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     preserveTabOrder,
     store,
     autoFocusOnShowProp,
+    focusedStoreRef,
   ]);
 
   const mayAutoFocusOnHide = !!autoFocusOnHide;
@@ -612,6 +643,16 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
 
   const onKeyDownProp = props.onKeyDown;
   const onKeyDownCaptureProp = props.onKeyDownCapture;
+  const onFocusCaptureProp = props.onFocusCapture;
+
+  const onFocusCapture = useEvent((event: ReactFocusEvent<HTMLType>) => {
+    onFocusCaptureProp?.(event);
+    if (!store.getState().open) return;
+    const target = event.target;
+    if (!isNode(target)) return;
+    if (!contains(event.currentTarget, target)) return;
+    focusedStoreRef.current = store;
+  });
 
   const acceptEscape = useEvent((event: KeyboardEvent) => {
     if (event.key !== "Escape") return false;
@@ -795,6 +836,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     "data-dialog-portal": hasDefaultModalPortal ? portalNode?.id : undefined,
     id,
     ref: useMergeRefs(ref, props.ref),
+    onFocusCapture,
     onKeyDown,
     onKeyDownCapture,
   };
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -62,6 +62,7 @@ import {
   markAndDisableTreeOutside,
 } from "./utils/disable-tree.ts";
 import {
+  isElementInside,
   isElementMarked,
   markTreeInside,
   markTreeOutside,
@@ -86,6 +87,16 @@ function isAlreadyFocusingAnotherElement(dialog?: HTMLElement | null) {
   if (dialog && contains(dialog, activeElement)) return false;
   if (isFocusable(activeElement)) return true;
   return false;
+}
+
+function isElementInDialog(
+  element: Element,
+  dialog: Element,
+  disclosureElement: Element | null,
+) {
+  if (contains(dialog, element)) return true;
+  if (disclosureElement && contains(disclosureElement, element)) return true;
+  return isElementInside(element, dialog);
 }
 
 function getElementFromProp(
@@ -452,6 +463,7 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
       contentElement;
     const isElementFocusable = isFocusable(element);
     if (!autoFocusOnShowProp(isElementFocusable ? element : null)) return;
+    const { disclosureElement } = store.getState();
     setAutoFocusEnabled(true);
     queueMicrotask(() => {
       // If the dialog was closed between scheduling and executing this
@@ -459,6 +471,17 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
       // element could steal focus from the disclosure that focusOnHide already
       // restored.
       if (!store.getState().open) return;
+      // Focus may have moved to another focusable element outside the dialog
+      // while it was waiting to be positioned. Preserve that user choice, but
+      // allow focus to remain on the disclosure or another owned element.
+      const activeElement = getActiveElement(element);
+      if (
+        activeElement &&
+        isFocusable(activeElement) &&
+        !isElementInDialog(activeElement, contentElement, disclosureElement)
+      ) {
+        return;
+      }
       // The browser's own focus scroll is unreliable here: Safari drops it
       // when a focus handler immediately moves focus elsewhere, which is what
       // virtual focus does. Scroll explicitly instead, so every engine behaves
--- a/packages/ariakit-react-components/src/dialog/dialog.tsx
+++ b/packages/ariakit-react-components/src/dialog/dialog.tsx
@@ -500,6 +500,9 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
       // focusable element outside while the dialog is waiting to be
       // positioned. Preserve that user choice, but allow a dialog that hasn't
       // received focus yet to perform its initial focus move.
+      // `getActiveElement` follows same-origin frames into another document.
+      // Ownership must stay in the dialog's document, where focus in a frame is
+      // represented by the frame element inside the dialog.
       const documentActiveElement = getDocument(contentElement).activeElement;
       const activeElement = isElement(documentActiveElement)
         ? documentActiveElement
--- a/.changeset/300-dialog-auto-focus-escape.md
+++ b/.changeset/300-dialog-auto-focus-escape.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Dialog`](https://ariakit.com/reference/dialog) and components built on it, such as [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover), so delayed auto-focus no longer pulls focus back after focus has moved outside the dialog.
--- a/app/src/sandbox/combobox-select-selected-value-presentation/index.react.tsx
+++ b/app/src/sandbox/combobox-select-selected-value-presentation/index.react.tsx
@@ -1,5 +1,5 @@
 import * as Ariakit from "@ariakit/react";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 const fruits = [
   "Apple",
@@ -216,11 +216,9 @@ function Fixture({
           style={{
             background: "white",
             border: "1px solid gray",
+            boxSizing: "border-box",
             padding: 8,
-            position: "fixed",
-            right: 8,
-            top: 8,
-            zIndex: 1,
+            width: 320,
           }}
         >
           Open the select. Expected focus history: input → focus target. An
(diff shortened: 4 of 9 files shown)


diegohaz:
## Browser evidence

In the sandbox, open the Focus escaping fruit fixture and click the select once. The fixed panel at the top makes the bug visible: the broken behavior records input → focus target → input → focus target → input → focus target, while the expected behavior ends after input → focus target.

ariakit-7033-repro-fixed-panel-after
[Short recording of the focus being reclaimed](https://github.com/user-attachments/assets/6e8283e9-6b59-41d1-ae60-37fb980a4be5)

app/src/sandbox/combobox-select-selected-value-presentation/test-browser.ts
```diff
@@ -132,6 +132,29 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+    await select.click();
+
+    // The app moved focus out of the popup while it was opening.
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(q.listbox()).not.toHaveAttribute("data-placing");
+    // Dialog queues auto-focus after placement, and there is no positive state
+    // for focus not being stolen once that microtask has run.
+    await flushFrames(page);
+    await test.expect(escapeTarget).toBeFocused();
+    await test.expect(focusHistory).toHaveText("input → focus target");
```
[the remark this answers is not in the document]
diegohaz: Fixed in [b2e59c3d6](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37). The browser regression now asserts that the listbox remains visible after the delayed auto-focus checkpoint, so it cannot pass because the popup closed.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -452,13 +463,25 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+      // while it was waiting to be positioned. Preserve that user choice, but
+      // allow focus to remain on the disclosure or another owned element.
+      const activeElement = getActiveElement(element);
+      if (
+        activeElement &&
+        isFocusable(activeElement) &&
+        !isElementInDialog(activeElement, contentElement, disclosureElement)
+      ) {
+        return;
+      }
```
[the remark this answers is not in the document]
diegohaz: Fixed in [b2e59c3d6](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37). The delayed callback now reads the current `open` and `disclosureElement` values inside the microtask. Browser and happy-dom disclosure-swap regressions verify that the previous disclosure keeps focus.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -88,6 +89,16 @@ function isAlreadyFocusingAnotherElement(dialog?: HTMLElement | null) {
   return false;
 }
 
+function isElementInDialog(
+  element: Element,
+  dialog: Element,
+  disclosureElement: Element | null,
+) {
+  if (contains(dialog, element)) return true;
+  if (disclosureElement && contains(disclosureElement, element)) return true;
```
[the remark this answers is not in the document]
diegohaz: Fixed in [b2e59c3d6](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37). Auto-focus suppression now requires focus history for the exact current store, so a virtual-focus owner present before the dialog receives focus no longer blocks legitimate initial focus. The Menu and Combobox regression suites, including React 18, pass with this narrower gate.

diegohaz:
**Fixed**

- [Popup visibility is now asserted after the delayed auto-focus checkpoint.](https://github.com/ariakit/ariakit/pull/7048#discussion_r3707212764)
- [The delayed callback reads the current disclosure inside its microtask.](https://github.com/ariakit/ariakit/pull/7048#discussion_r3707214462)
- [Exact-store focus history preserves legitimate virtual-focus initial focus.](https://github.com/ariakit/ariakit/pull/7048#discussion_r3707215515)

Implemented in [b2e59c3d6](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37). Main, React 18, type-checking, package and app builds, and the affected Chrome browser suites pass. The existing [browser evidence](https://github.com/ariakit/ariakit/pull/7048#issuecomment-5167123335) shows the user-visible focus history.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -454,11 +490,28 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+      // received focus yet to perform its initial focus move.
+      const activeElement = getActiveElement(element);
+      const deepestActiveElement =
+        activeElement && getDeepestActiveElement(activeElement);
+      if (
+        focusedStoreRef.current === store &&
+        activeElement &&
+        deepestActiveElement &&
+        isFocusable(deepestActiveElement) &&
+        !isElementInDialog(activeElement, contentElement, disclosureElement)
```
diegohaz: `isElementInDialog` decides ownership from `getActiveElement(element)`, which returns `document.activeElement`. When focus is inside an open shadow root, that is the outermost shadow **host**, not the focused element. For a dialog that itself lives inside an open shadow root, the host is an ancestor of `contentElement`, so `contains(dialog, element)` is false, `contains(disclosureElement, element)` is false, and `isElementInside(element, dialog)` is false too. Focus that is genuinely inside the dialog is therefore classified as an escape, and the guard suppresses the placement-gated auto-focus: `initialFocus` is never applied and the `scrollIntoView` above never runs.

`getDeepestActiveElement` already resolves the real focused node, but it only feeds `isFocusable`, never the ownership check, so the shallow host decides the outcome.

**Real-world use case**

Apps that mount Ariakit inside a web component or a micro-frontend shadow root (an established pattern this repository already covers in `dialog-persistent-elements`). Open any `Dialog` there whose content takes focus before placement finishes, for example an app effect that focuses a field on mount, or a user clicking inside a popup during a slow custom `updatePosition`. On `main` the dialog then moves focus to its `initialFocus` target; on this branch it does not, and the target is never scrolled into view either.

**Evidence**

I added a temporary sandbox (`Dialog` portaled into an open shadow root, `initialFocus` on the second field, a child effect focusing the first field on mount) plus a Chrome test, ran it against both revisions, and then removed it. Green on `main` ([`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb)), red on this head.

`main`, focus history reaches the `initialFocus` target:

Dialog inside an open shadow root on main: focus history reads first field then initial focus field, and the initialFocus target has the focus ring

This branch, auto-focus is suppressed and the history stops at the app-focused field:

Same dialog on this branch: focus history stops at first field and the initialFocus target never receives focus

**Assessment**

- Severity: Medium (estimate, 85% confidence), because the dialog still holds focus somewhere inside itself, but the documented `initialFocus` contract and the initial scroll are silently dropped.
- Priority: P1 (estimate, 85% confidence), because this branch introduces the regression in a configuration the repository already supports and tests, and the fix is local.
- Breakage likelihood: Possible (estimate, 75% confidence), because it needs Ariakit mounted inside an open shadow root plus focus reaching the dialog before placement completes; it does not need any shadow root inside the dialog.
- Fix effort: Easy (estimate, 90% confidence), because the ownership check just has to consider the resolved focused node, and an existing predicate already does the rest.

**Suggested fix**

The narrow version is to run the ownership check on the resolved node as well:

```diff
-        !isElementInDialog(activeElement, contentElement, disclosureElement)
+        !isElementInDialog(activeElement, contentElement, disclosureElement) &&
+        !isElementInDialog(
+          deepestActiveElement,
+          contentElement,
+          disclosureElement,
+        )
```

Both checks are needed: the shallow one still covers a nested shadow root *inside* the dialog, where `contains` cannot cross the boundary but the host is a descendant.

The version I would prefer also removes a near-duplicate. `isElementInDialog` is `isElementWithinDialog` from `use-hide-on-interact-outside.ts` minus the `aria-activedescendant` resolution and the `data-focus-trap` case, and both of those are relevant here: a virtual-focus owner and a focus trap around a portaled dialog are just as much "not an escape" for auto-focus as they are for outside interaction. Exporting `isElementWithinDialog`, calling it for both `activeElement` and `deepestActiveElement`, and deleting `isElementInDialog` (plus the then-unused `isElementInside` import) covers this bug and drops the duplicated predicate.

**Validation**

I applied both versions locally. Each one turns the repro green and keeps everything else green: `pnpm test` (254 files, 1699 tests), `pnpm test-react18` (187 files, 965 tests), and `pnpm test-chrome` over `combobox-select-selected-value-presentation`, `dialog-persistent-elements`, `menu-placing-pass`, and `combobox-select-menu-filters`. For the shared-predicate version I also ran `pnpm tsc` and `pnpm lint` clean. Notably the new "does not refocus after native auto-focus moves focus into a shadow root" test still passes, because there the resolved node is outside the dialog and both checks agree.

diegohaz (commented):
Initial review. Baseline is the merge base [`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb) to head [`b2e59c3`](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37), reviewed in a dedicated worktree and expanded beyond the diff into `dialog.tsx`, the dialog `utils/` predicates, `focusable.tsx` prop chaining, and the fixtures the changed sandboxes already own.

## Unresolved review threads

There are none. All three existing threads (the two CodeRabbit ones and the Codex virtual-focus one) are `isResolved == true` on the current head, so this run had no candidate findings to revalidate from prior review history.

## What I ran

- `pnpm test`: 254 files, 1699 tests, all passing.
- `pnpm test-react18`: 187 files, 965 tests, all passing.
- `pnpm test-chrome` over `combobox-select-selected-value-presentation` and `dialog-persistent-elements`: 28 passing.
- Regression proof by reverting only `packages/ariakit-react-components/src/dialog/` to the baseline, in both environments.

## Regression proof

The red check is solid in the browser. With the production change reverted, four Chrome tests fail: "does not refocus after focus leaves an open popup", "does not refocus after native auto-focus moves focus into a shadow root", "uses the current disclosure in delayed auto-focus", and the pre-existing #7020 test "cancels the item handoff presentation when focus leaves an open popup", which now also depends on this fix because `autoFocusOnShow={false}` was removed from "Page escaping fruit" as its old comment anticipated. In happy-dom, "uses the current disclosure in delayed auto-focus" fails on the baseline.

Worth recording so it is not mistaken for regression proof: "focuses a dialog after its store changes" and `dialog-persistent-elements` "closes on a newly inserted outside element after replacing the dialog" both pass on the baseline. They are guard tests for regressions this change could have introduced, which is a legitimate and useful role, just not evidence for the reported bug.

## Finding published inline

One, on the ownership check in the delayed auto-focus guard: `isElementInDialog` classifies from `document.activeElement`, so a dialog that lives inside an open shadow root has its own inside-focus read as an escape, and `initialFocus` plus the initial `scrollIntoView` are silently skipped. Confirmed with a temporary sandbox and Chrome test that is green on the baseline and red on this head, with before and after screenshots in the comment. The experiment has been removed and the worktree is back at the PR head.

## Candidates reviewed and not published

**Alternative design, dropping the store focus history entirely.** The obvious simplification is to delete `focusedStoreRef`, `onFocusCapture`, the `useHideOnInteractOutside` signature change, and the new disclosure-effect guard, and instead widen the predicate to the existing `isElementWithinDialog`, which already resolves `aria-activedescendant` exactly as the Codex thread asked. I built and ran that version. It regresses three existing tests: `combobox-select-menu-filters` "checkbox menu removes an uncommitted keyboard filter" and "click menu removes an uncommitted keyboard filter", and `menu-placing-pass` "keeps focus out of a popup that mounts after its store is open". So this is not a finding; it is measured support for the design in the PR description. The exact-store gate is doing work that no active-element predicate can do on its own, and the inline comment is careful to keep it.

**Suspected, not validated: the same ownership gap across a frame boundary.** `getActiveElement` recurses into same-origin iframes, so when focus is inside an iframe rendered *inside* the dialog, `activeElement` is the inner element and `contains(contentElement, activeElement)` is false across documents. That reads as an escape by the same mechanism as the shadow case, and the iframe element itself receiving focus in the outer document should be enough to set the store focus history first. I did not build a fixture for it, so I am not claiming it as confirmed, and the fix I suggested inline does not address it either since `contains` cannot cross a frame boundary in either direction. Flagging it as the natural next question rather than as current-PR work.

**Fixture duplication, declined non-goal.** "Focus escaping fruit" is prop-identical to "Page escaping fruit" apart from `label` and `showFocusHistory`, so the two fixtures now exercise the same escape. Folding the history panel into the existing fixture would remove the clone, but that fixture carries the #7020 assertions on exact `scrollY` and on no document scroll events, and a `position: fixed` panel pinned at the top right is exactly the kind of overlay that can disturb an actionability check there. A sibling variant is what the fixture-lifecycle guidance asks for in this situation, so the duplication looks like the right trade rather than something to collapse.

## Other things checked

The changeset targets `@ariakit/react-components` and `@ariakit/react` at `patch`, which matches how consumers receive the behavior, and both reference links resolve. `useFocusable` reads `props.onFocusCapture` and chains it before its own handler runs, so the new dialog handler survives the later `useFocusable` and `usePortal` passes rather than being overwritten. The `contains(event.currentTarget, target)` guard in that handler is load-bearing: without it, focus inside a React portal rendered by a dialog descendant would mark the dialog as focused. Making `interactedOutsideRef` a required option and dropping the three `if (interactedOutsideRef)` guards is behavior-preserving, since `useDialog` is the only caller. `getDeepestActiveElement` always descends strictly, so it terminates.

## Not fully validated

Browser coverage was Chrome only; I did not run the Firefox, Safari, iOS, or Android projects, and visual tests are CI-only by policy. I did not run `pnpm build` or inspect packed output, since nothing here changes published file layout or exports. CI on this head is green (48 successful, 1 skipped).

diegohaz:
## Browser evidence for `e02491ed1`

Open **Shadow-root focus dialog** in the sandbox. The visible history shows `app focus → initial focus`: native app focus lands inside a dialog mounted in an open shadow root, then Ariakit correctly applies `initialFocus`. Before the fix, the history stopped at `app focus`.

Shadow-root Dialog moves from app autofocus to initialFocus

[Short recording of the shadow-root initial-focus transition](https://github.com/user-attachments/assets/423168e0-94c9-4902-85bd-47da57208f11)

The adjacent **Focus trap escaping fruit** fixture covers the inverse boundary: focus moved to an external `data-focus-trap` remains there and is not reclaimed.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`e02491ed1`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151): delayed auto-focus now checks the narrow dialog ownership predicate against both the shallow active host and the deepest focused node. The shallow check still covers a nested shadow root inside the dialog, while the deepest check lets a dialog mounted inside an open shadow root retain `initialFocus`. I kept the predicate local so an external `data-focus-trap` remains a legitimate escape, and added a red-checked browser regression for that boundary. The sandbox now exposes both behaviors through visible focus histories, with [a screenshot and short recording](https://github.com/ariakit/ariakit/pull/7048#issuecomment-5177509096); the shadow-root case is also covered in happy-dom for React 18.

diegohaz:
**Fixed**

- [`e02491ed1`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151) → [review reply](https://github.com/ariakit/ariakit/pull/7048#discussion_r3711339878): delayed auto-focus now recognizes Dialog focus through either side of an open shadow boundary without treating an external focus trap as owned. Added visible sandbox histories plus browser and happy-dom regressions; [browser evidence](https://github.com/ariakit/ariakit/pull/7048#issuecomment-5177509096).

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -454,11 +490,33 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
       // microtask, skip the focus call. Otherwise, focusing a now-hidden
       // element could steal focus from the disclosure that focusOnHide already
       // restored.
-      if (!store.getState().open) return;
+      if (!open) return;
+      // Once the dialog has received focus, focus may move to another
+      // focusable element outside while the dialog is waiting to be
+      // positioned. Preserve that user choice, but allow a dialog that hasn't
+      // received focus yet to perform its initial focus move.
+      const activeElement = getActiveElement(element);
```
diegohaz: `getActiveElement` follows same-origin frames, so when focus is inside an iframe it returns the element in the *frame's* document rather than the frame element in the dialog's document. `getDeepestActiveElement` then only resolves shadow roots, so both ownership checks below run against a node in another document, and `contains` cannot cross a document boundary any more than it can cross a shadow boundary. An iframe that lives inside the dialog is therefore classified as an escape, and the delayed auto-focus is suppressed: `initialFocus` is never applied and the `scrollIntoView` never runs.

This is the same class of bug as the shadow-root case fixed in [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), and that fix does not reach it, because the deepest node is in the frame's document too.

I flagged this direction as unvalidated in my previous review. It reproduces, so I am reporting it as confirmed now.

**Real-world use case**

A dialog containing a same-origin iframe, which is how embedded payment fields, editors, andmedia widgets are usually delivered. Open the dialog, let the app focus one of the dialog's own fields (this is what records the store focus history), then let the embedded frame take focus during the placing window. On `main` the dialog still moves focus to its `initialFocus` target; on this head it does not. No artificial delay or custom `updatePosition` is needed; the ordinary mount flow is enough.

**Evidence**

Temporary sandbox: a `Dialog` with `initialFocus` on the last field, a plain field the content focuses on mount, and an imperatively created same-origin iframe whose input focuses itself. Green on the merge base [`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb), red on this head. Instrumenting the guard on this head shows exactly why it fires:

```
storeMatch: true, deepFocusable: true,
activeLabel: "Iframe field", deepLabel: "Iframe field",
shallowIn: false, deepIn: false
```

This head, the history stops at the frame and the `initialFocus` target never receives focus:

Dialog containing an iframe on this head: focus history stops at dialog field then iframe field, and the initialFocus target is never focused

With the change below, the history completes:

Same dialog with the fix: focus history reads dialog field, iframe field, initial focus, and the initialFocus target has the focus ring

**Assessment**

- Severity: Medium (estimate, 85% confidence), because focus stays inside the dialog, but the `initialFocus` contract and the initial scroll are silently dropped.
- Priority: P1 (estimate, 80% confidence), because this head introduces the regression relative to the merge base, and it is the same predicate that was just corrected for shadow roots.
- Breakage likelihood: Possible (estimate, 70% confidence), because it needs an embedded same-origin frame inside the dialog that takes focus while auto-focus is still pending, which embedded payment and editor widgets do on mount.
- Fix effort: Easy (estimate, 85% confidence), because the ownership lookup just has to stay in the dialog's own document.

**Suggested fix**

Resolve the active element in the dialog's document instead of following frames. A focused frame is then the frame element sitting inside the dialog, which `contains` handles, while the existing shadow traversal keeps working:

```diff
-      const activeElement = getActiveElement(element);
+      // `getActiveElement` follows same-origin frames, so it would report an
+      // element in another document. Ownership has to be decided in the
+      // dialog's own document, where a focused frame is still the frame
+      // element sitting inside the dialog.
+      const documentActiveElement = getDocument(contentElement).activeElement;
+      const activeElement = isElement(documentActiveElement)
+        ? documentActiveElement
+        : null;
       const deepestActiveElement =
         activeElement && getDeepestActiveElement(activeElement);
```

Focus in an *unrelated* frame still suppresses correctly, because that frame's element is outside the dialog and focusable. `getDocument` and `isElement` are already imported here.

**Validation**

With that change: the repro turns green, `pnpm test` stays at 254 files and 1700 tests, `pnpm test-chrome` over `combobox-select-selected-value-presentation` and `dialog-persistent-elements` stays at 30 passing (including all four tests added on this head), and `pnpm tsc` and `pnpm lint` are clean.

diegohaz (commented):
Re-review. The incremental baseline is the head recorded in my previous review marker, [`b2e59c3`](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37), to the current head [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), which adds one commit. Newly discovered findings come from that commit plus surrounding behavior it newly exposes, and I revalidated the previous finding against the current head rather than trusting the reply on its thread.

## Previous finding

Confirmed fixed. The guard now runs the ownership predicate against both the shallow active host and the deepest focused node, which is the narrow variant I validated last time. I re-ran the red check: with only `dialog.tsx` reverted to [`b2e59c3`](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37), the new "honors initial focus when the dialog is inside a shadow root" test fails in both happy-dom and Chrome, and passes on this head. That is genuine regression proof for the shadow-root case, in both environments.

Its thread is resolved, so it was not re-adjudicated as a candidate.

## Unresolved review threads

None. All four threads on this PR are resolved, so this run had no unresolved candidates to validate.

## What I ran on this head

- `pnpm test`: 254 files, 1700 tests, all passing.
- `pnpm test-chrome` over `combobox-select-selected-value-presentation` and `dialog-persistent-elements`: 30 passing.
- Red checks against [`b2e59c3`](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37) and against the merge base [`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb).

## Finding published inline

One, on the same ownership predicate: `getActiveElement` follows same-origin frames, so an iframe *inside* the dialog is classified as an escape and the delayed auto-focus is suppressed, dropping `initialFocus` and the initial scroll. This is the frame-boundary twin of the shadow-root bug just fixed, and the deepest-node fix does not reach it. I raised this direction as unvalidated in my previous summary; it now reproduces, green on the merge base and red on this head, with instrumented guard output and before and after screenshots in the comment. A one-line change to resolve the active element in the dialog's own document fixes it and keeps every suite green. The experiment has been removed and the worktree is back at this head.

## Candidates reviewed and not published

**Suspected, not validated: the dialog's own focus trap is treated as an escape.** The reply on the previous thread explains keeping the predicate local so that "an external `data-focus-trap` remains a legitimate escape", and the new "does not refocus an external focus trap" test encodes that. The implementation cannot actually tell external from own, though: `isElementInDialog` never reads `data-focus-trap` at all, so a trap this dialog itself rendered is classified as an escape too, even though its whole job is to hand focus back into the dialog. `packages/ariakit-react-components/src/dialog/utils/is-focus-trap.ts` already expresses the ownership-aware form, `isFocusTrap(element, ...ids)`, matching the attribute value against specific ids, and `disable-tree.ts` uses it that way. Reaching it needs focus to land on this dialog's own trap while auto-focus is still pending, which I did not manage to stage, so I am flagging it rather than claiming it.

**Test observation, not a defect.** "does not refocus an external focus trap" has the same body as the "does not refocus after focus leaves an open popup" test above it, differing only in the fixture labels, and the `data-focus-trap` attribute that names it is never read by the code under test. It does guard something real, that the auto-focus predicate must not adopt the `data-focus-trap` handling from `isElementWithinDialog`, but nothing in the test or fixture says so, so the next reader is likely to file it as a copy and delete it. One sentence naming what it protects would carry that. Not published inline because it is a comment on an otherwise correct test.

**Fixture duplication, still declined.** "Focus trap escaping fruit" is now the third prop-identical instance of the same escape fixture, after "Page escaping fruit" and "Focus escaping fruit". Each clone is a dozen lines of props on a shared parameterized component, and merging them would put a fixed-position panel into the fixture that carries the #7020 scroll assertions, so the trade still looks right. Recording it because the count is growing, not as work for this PR.

## Coverage note

The new happy-dom duplicate reads `shadowRoot.activeElement` directly, which is unavoidable since `toHaveFocus` resolves against `document.activeElement` and would see only the host. It is paired with the user-visible focus-history assertion, so the raw probe is supplementary rather than the primary signal, which matches the guidance.

## Not fully validated

Chrome only, as before; I did not run Firefox, Safari, iOS, or Android, and visual tests remain CI-only. I did not run `pnpm build`, since nothing here changes published packaging.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`8cfec15f7`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4): active-element ownership now starts in the Dialog document, where focus inside a contained iframe remains represented by its host. The existing shallow/deep shadow checks continue from that local element, while an unrelated iframe host remains an outside escape. A red-checked browser regression verifies `dialog focus → iframe focus → initial focus` in Chrome, Firefox, and Safari.

diegohaz:
**Fixed**

- Kept delayed auto-focus ownership in the Dialog document, so focus inside a contained same-origin iframe is represented by its host without weakening outside-frame escape handling.
- Added a red-checked browser regression for `dialog focus → iframe focus → initial focus`; the complete sandbox passes in Chrome, Firefox, and Safari.
- Kept the original #7033 focus histories in readable normal-flow panels so fixtures no longer cover one another during manual verification.
- Updated the PR description with cycle 9 of the implementation reassessment and the latest browser evidence.

Implemented in [`8cfec15f7`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4). All review threads are resolved.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -454,11 +490,36 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
       // microtask, skip the focus call. Otherwise, focusing a now-hidden
       // element could steal focus from the disclosure that focusOnHide already
       // restored.
-      if (!store.getState().open) return;
+      if (!open) return;
+      // Once the dialog has received focus, focus may move to another
+      // focusable element outside while the dialog is waiting to be
+      // positioned. Preserve that user choice, but allow a dialog that hasn't
+      // received focus yet to perform its initial focus move.
+      const documentActiveElement = getDocument(contentElement).activeElement;
```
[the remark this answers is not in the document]
diegohaz: I reproduced this and the described behavior is real, but it is not caused by this change, so I am not treating it as a finding against this PR.

I built the exact scenario: a non-modal `Dialog` portaled into a same-origin iframe's body, focus entering a field inside the dialog first (which records the store focus history), then moving to a button in the parent document before the microtask runs. Instrumenting the guard on the current head confirms your mechanism precisely:

```
storeMatch: true, docActiveTag: "BODY", deepTag: "BODY", deepFocusable: false
```

The child document reports `BODY`, `isFocusable` is false, the guard does not fire, and `element.focus()` reclaims focus. The visible history ends `dialog field → parent escape → initial focus`:

Dialog inside an iframe: focus history reads dialog field, parent escape, initial focus, showing focus reclaimed from the parent document

Where the report is inaccurate is calling this "the regression this change aims to prevent". I ran the same fixture against three revisions with only `packages/ariakit-react-components/src/dialog/` swapped:

- [`8cfec15`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4) (current head): `dialog field → parent escape → initial focus`
- [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151) (before this commit): identical
- [`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb) (merge base, before the whole PR): identical

The screenshots captured on the current head and on the merge base are byte-identical, same md5 `0197095647...`. That follows from the code: the previous line was `getActiveElement(element)`, which also starts at `getDocument(element).activeElement`, the same child document, and only ever recurses *downwards* into frames. Reading the owner document is not what loses the parent target; neither version could ever see it.

So this configuration is unchanged by the PR rather than broken by it, and it is the case the PR description already lists as intentionally unsupported ("body or null active targets"). Handling it would mean walking *up* the readable frame chain, which is a different capability from the downward traversal that exists today.

I am leaving it as an unfixed corner of #7033 rather than filing it, since the non-goal is the author's call. Happy to open a separate issue with this reproduction if you would rather track it.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -454,11 +490,36 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+      const activeElement = isElement(documentActiveElement)
+        ? documentActiveElement
+        : null;
```
diegohaz, line 509: This deliberately avoids `getActiveElement`, but nothing here says so, and `getActiveElement` is still imported and used twice in this same file (lines 86 and 293). The difference that matters is invisible: `getActiveElement` follows same-origin frames, so it would report a node in the frame's document and make both `isElementInDialog` checks fail for an iframe that is inside the dialog. That is the bug this commit fixes.

The risk is that the next reader sees a hand-rolled `getDocument(...).activeElement` plus an `isElement` narrowing sitting two lines away from a helper that appears to do the same job, and consolidates it back. The browser test would catch it, but only after someone spends the time to work out why a frame test is failing on an unrelated cleanup.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the code is correct today and the regression it guards against is covered by a browser test.
- Priority: P2 (estimate, 85% confidence), because it protects a non-obvious decision cheaply, but nothing is broken without it.
- Breakage likelihood: Possible (estimate, 65% confidence), because the idiomatic helper is right there and this file is actively refactored; a future consolidation pass is the realistic trigger.
- Fix effort: Easy (estimate, 95% confidence), because it is one comment.

**Suggested fix**

```diff
+      // `getActiveElement` follows same-origin frames, so it would report an
+      // element in another document. Ownership has to be decided in the
+      // dialog's own document, where a focused frame is still the frame
+      // element sitting inside the dialog.
       const documentActiveElement = getDocument(contentElement).activeElement;
```

diegohaz (commented):
Re-review. The incremental baseline is the head recorded in my previous review marker, [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), to the current head [`8cfec15`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4), which adds one commit.

## Previous finding

Confirmed fixed. Ownership now resolves through the Dialog document, so a contained iframe is represented by its host element. I re-ran the red check: with only `dialog.tsx` reverted to [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), the new "honors initial focus after focus enters a contained iframe" test fails in Chrome and passes on this head. Its thread is resolved, so it was not re-adjudicated.

## Unresolved thread validated this run

One, the Codex report on line 503 about escaped focus outside an iframe. I reproduced the scenario, confirmed its mechanism exactly (`docActiveTag: "BODY"`, `deepFocusable: false`, focus reclaimed), and then established that it is not caused by this change: the same fixture produces the identical focus history on this head, on [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), and on the merge base [`0beb71c`](https://github.com/ariakit/ariakit/commit/0beb71cf9fc05af051679fe0b714e4afdf5d0beb), with byte-identical screenshots between the current head and the merge base. The previous line, `getActiveElement(element)`, read the same child document and only ever recursed downwards into frames, so neither version could see a parent-document target.

Disposition: declined non-goal for this PR. The behavior is real but pre-existing, unchanged by the PR, and in the class the PR description already lists as intentionally unsupported ("body or null active targets"). Fixing it needs upward frame-chain resolution, which is a different capability from the downward traversal that exists today. I did not open an issue, because declaring the non-goal is the author's call; the reply on the thread offers to file one with the reproduction if wanted.

- Severity: Medium (estimate, 75% confidence), because focus is reclaimed after the user left, the same user impact as #7033.
- Priority: P2 (estimate, 80% confidence), because it is pre-existing and independent of this change.
- Breakage likelihood: Rare (estimate, 75% confidence), because it needs the dialog itself hosted in a same-origin frame and the escape target in the parent document.
- Fix effort: Moderate (estimate, 70% confidence), because it needs readable frame-chain traversal plus ownership rules for hosts along the way.

Details and evidence are on that thread.

## What I ran on this head

- `pnpm test`: 254 files, 1700 tests, all passing.
- `pnpm test-react18`: 187 files, 966 tests, all passing.
- `pnpm test-chrome` over `combobox-select-selected-value-presentation` and `dialog-persistent-elements`: 31 passing.
- Red check against [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), plus three-revision comparisons for the Codex scenario.

## Finding published inline

One, low priority: the new `getDocument(contentElement).activeElement` plus `isElement` narrowing carries no comment explaining why it deliberately avoids `getActiveElement`, which is still imported and used twice in the same file. The distinction (frame recursion) is exactly what this commit fixes, so an uncommented hand-rolled lookup next to the idiomatic helper invites a future consolidation that would silently reintroduce the bug.

## Checked and not raised

**Browser-only coverage for the iframe case is correct, and I verified it rather than assuming.** The shadow-root case got a happy-dom duplicate but this one did not, so I wrote the duplicate and ran it. It passes on this head *and* on [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), which is the broken revision, so happy-dom does not model the frame focus semantics the browser test depends on and the duplicate would be vacuous. Leaving this browser-only is the right call.

**The new fixture follows the repository's iframe guidance.** `IframeDialogFixture` sets the portal target from the iframe ref callback, renders through `createPortal`, and clears the target when the ref is cleared, rather than mounting a nested root. The browser test uses the injected `query` helper with `frameLocator` instead of raw `getByRole` on a locator.

**Focus-history panel is now in normal flow.** Dropping `position: fixed` removes the overlay that made me justify the sibling fixtures last time, and every suite stays green. Worth noting that the three prop-identical escape fixtures could now plausibly be collapsed, since the reason to keep them apart was that overlay. Still not work for this PR.

**Staging through `autoFocusOnShow`.** The new fixture again drives focus from inside the `autoFocusOnShow` callback, consistent with `DisclosureSwapFixture`. My own reproduction reached the same window from an ordinary child effect, so a more app-like staging exists, but the callback is deterministic and the behavior it exercises is real. Not worth changing.

## Not fully validated

Chrome only for the browser runs, as in my earlier reviews; the author reports the new test green in Firefox and Safari as well, which I did not verify. Visual tests remain CI-only, and I did not run `pnpm build`, since nothing here affects published packaging.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`a5cd577b2`](https://github.com/ariakit/ariakit/commit/a5cd577b2397ac159595b7287e4c927c08f83804): documented that `getActiveElement` follows same-origin frames into another document, while this ownership check must retain the frame element from the Dialog's document. The existing red-checked browser regression protects this invariant in Chrome, Firefox, and Safari.

diegohaz:
**Fixed**

- [`a5cd577b2`](https://github.com/ariakit/ariakit/commit/a5cd577b2397ac159595b7287e4c927c08f83804) ([reply](https://github.com/ariakit/ariakit/pull/7048#discussion_r3712621130)): Documented why delayed focus ownership must use the Dialog document's active element instead of `getActiveElement`, preserving a contained iframe as its frame element. The existing red-checked browser regression protects the invariant across Chrome, Firefox, and Safari.

**Declined**

- `Non-goal` ([reply](https://github.com/ariakit/ariakit/pull/7048#discussion_r3712727584)): Closed-shadow internals remain intentionally unsupported because the document exposes only the host, which cannot distinguish an external escape from focus inside a Dialog mounted in that same closed root. The merge base already reclaims focus in this case.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -88,6 +90,26 @@ function isAlreadyFocusingAnotherElement(dialog?: HTMLElement | null) {
   return false;
 }
 
+function getDeepestActiveElement(element: Element) {
+  let activeElement = element;
+  while (activeElement.shadowRoot) {
+    const nextElement = activeElement.shadowRoot.activeElement;
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because closed-shadow internals are intentionally unsupported: the document exposes only the host, so treating an external-looking host as an escape would also suppress legitimate initial focus when the Dialog itself is mounted inside that same closed root. The merge base always reclaims focus in this case, so the PR does not worsen it, and the PR description already records closed-shadow internals as outside the supported boundary.

diegohaz (commented):
Re-review. The incremental baseline is the head recorded in my previous review marker, [`8cfec15`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4), to the current head [`a5cd577`](https://github.com/ariakit/ariakit/commit/a5cd577b2397ac159595b7287e4c927c08f83804), which adds one commit.

**No newly discovered findings.** The incremental change is three comment lines and nothing else: `git diff --stat` reports one file and three insertions, with no deletions and no change under `app/`. The comment is accurate about the behavior it documents, and its longest line is 80 characters, within the wrap guidance.

## Previous finding

Confirmed addressed. [`a5cd577`](https://github.com/ariakit/ariakit/commit/a5cd577b2397ac159595b7287e4c927c08f83804) documents why the ownership lookup deliberately avoids `getActiveElement`, which was the whole point of the comment I asked for. Its thread is resolved, so it was not re-adjudicated.

## Unresolved review threads

None. Every thread on this PR is now resolved, including the closed-shadow-root report opened since my last review, so this run had no unresolved candidates to validate.

I did read that resolved thread for context and its reasoning holds up. Treating any external-looking active host as an escape is not separable from the legitimate case, because a Dialog mounted inside that same closed root also presents as an outside host from the document's point of view, which is the exact mirror of the open-shadow-root case fixed earlier in this PR. And the sub-case where the closed host is itself focusable is handled better here than on the merge base rather than worse, since the guard then fires and preserves the escape. Recording this only as agreement with a resolved decision, not as a new finding.

## What I ran on this head

- `pnpm test`: 254 files, 1700 tests, all passing.
- `pnpm test-react18`: 187 files, 966 tests, all passing.
- `pnpm test-chrome` over `combobox-select-selected-value-presentation` and `dialog-persistent-elements`: 31 passing.

I did not repeat the red checks this round. They were established against [`b2e59c3`](https://github.com/ariakit/ariakit/commit/b2e59c3d6f0927f4c4eaaf5ddc1e65d0f22f7c37), [`e02491e`](https://github.com/ariakit/ariakit/commit/e02491ed123d868cfaa503de9695aabed66b3151), and [`8cfec15`](https://github.com/ariakit/ariakit/commit/8cfec15f78012d65f14497ee2f7905fb1d9330a4) in my earlier reviews, and a comment-only change cannot affect them.

## Whole-change pass

Since this looks close to final, I re-read the full production diff from the merge base rather than only the increment. It reads coherently: one store-valued focus-history ref with its reset boundaries, one React capture handler chained ahead of `useFocusable`, the object-parameter form of `useHideOnInteractOutside` with `interactedOutsideRef` now required, and the ownership predicate resolving shallow host, deep shadow node, and frame boundary. `use-hide-on-interact-outside.ts` has not changed since my first review. The changeset still describes the shipped behavior correctly; the shadow-root and iframe corrections were regressions introduced and resolved inside this PR, so they do not need separate entries.

## Evidence note

No visual evidence accompanies this review because nothing browser-visible changed. The behavior screenshots on the earlier threads still describe the current head.

## Not fully validated

Chrome only for the browser runs, as in my earlier reviews; the author reports the regressions green in Firefox and Safari, which I did not verify. Visual tests remain CI-only, and I did not run `pnpm build`, since nothing here affects published packaging.

PR #7062, opened 2026-08-03
Ignore inherited properties in TagList and store selectors

### LINKED ISSUE #7058: TagList copies inherited enumerable properties onto its listbox element
## Context

Found while reviewing [#7034](https://github.com/ariakit/ariakit/pull/7034), which guards `removeUndefinedValues` in `@ariakit/utils` so it copies own properties only. That guard stops Ariakit from materializing inherited enumerable `Object.prototype` properties as own props when it builds an element's props.

`TagList` re-introduces the same problem downstream of that guard through its own unguarded `for...in`.

## Problem

`packages/ariakit-react-components/src/tag/tag-list.tsx:121` iterates caller-controlled props with an unguarded `for...in`:

```ts
const listboxProps: typeof props = {};
for (const key in props) {
  if (key === "role" || key.startsWith("aria-")) {
    const prop = key as keyof typeof props;
    listboxProps[prop] = props[prop];
    delete props[prop];
  }
}
```

`listboxProps` is then spread onto the internal listbox element.

When `Object.prototype` carries an enumerable property, `for...in` still walks the prototype chain, so an inherited `role` or `aria-*` key is read through the prototype and assigned as an **own** property of `listboxProps`. The matching `delete props[prop]` is a no-op for an inherited key, so the value is copied rather than moved.

An inherited `aria-hidden` would therefore be rendered on the listbox element, hiding the tags from assistive technology.

This violates the Property Ownership rule the project applies to caller-controlled prop records, which calls for `Object.hasOwn` or an established helper instead of an unguarded `for...in`. `mergeProps` in `packages/ariakit-react-utils/src/misc.ts:66-67` and, after #7034, `removeUndefinedValues` both follow that rule.

## Pre-existing on `main`

Verified: this is **pre-existing on `main`** and is not introduced by #7034.

`git show origin/main:packages/ariakit-react-components/src/tag/tag-list.tsx` contains the identical unguarded loop at line 121, and `git log origin/main..HEAD -- packages/ariakit-react-components/src/tag/tag-list.tsx` on the #7034 branch is empty, so that branch never touches the file. The fix does not need to wait for #7034 to merge.

## Why #7034 can finish without it

#7034 closes the exposure it introduced, plus the pre-existing one in `removeUndefinedValues`, in the helper the whole component system funnels through. The `TagList` loop is a separate code path that predates it and is unaffected by that change either way, so it is out of scope for that pull request. Its changeset names representative components rather than claiming a blanket fix.

## Expected outcome

Guard the loop so only own properties are considered, matching `mergeProps`:

```diff
 for (const key in props) {
+  if (!hasOwnProperty(props, key)) continue;
   if (key === "role" || key.startsWith("aria-")) {
```

Add focused coverage coming from an inherited enumerable prototype key, per the Property Ownership testing guidance.

## Workaround

See [the verified workaround and reproduction](https://github.com/ariakit/ariakit/issues/7058#issuecomment-5169881347). The pull request's [Workaround section](https://github.com/ariakit/ariakit/pull/7062#workaround) documents its scope and limitations.

diegohaz:
## Motivation

`TagList` and `useStoreStateObject` iterate caller-controlled records with unguarded `for...in` loops. Inherited enumerable properties can therefore replace the internal TagList listbox semantics, become own keys in returned store snapshots, execute as selector functions, or create unrelated subscriptions.

Fixes #7058

## Solution

Guard the three caller-controlled `for...in` loops with Ariakit's `hasOwnProperty` helper. `TagList` also checks ownership before using `aria-label` to suppress `aria-labelledby`, and `useStoreStateObject` writes own snapshot keys even when an inherited value compares equal.

The `tag-list-7058` sandbox covers inherited `aria-hidden`, `aria-label`, and `role` values through the rendered listbox. The `store-keyed-subscriptions` sandbox covers exact own snapshot keys, including keys colliding with `Object.prototype`, inherited selector execution, and subscription work caused by an inherited direct store key. Every behavior has happy-dom coverage and browser-authoritative checks in Chrome, Firefox, and WebKit.

## Preview

Current-main browser reproduction showing TagList aria-hidden and inherited selector processing

Current main copies the inherited `aria-hidden` value and loses the listbox from the accessibility tree.

Fixed TagList preserving its accessible listbox semantics

With the fix, the inherited values are ignored: `aria-hidden` is absent, the role remains `listbox`, and the accessible name remains `Tags`.

## Workaround

If a library adds enumerable properties to `Object.prototype`, make each known configurable extension non-enumerable before rendering Ariakit components:

```js
// TODO: Remove after https://github.com/ariakit/ariakit/issues/7058 is fixed.
const key = "aria-hidden";
const descriptor = Object.getOwnPropertyDescriptor(Object.prototype, key);

if (descriptor?.configurable) {
  Object.defineProperty(Object.prototype, key, {
    ...descriptor,
    enumerable: false,
  });
}
```

Direct `useStoreStateObject` consumers can copy own selector entries into a null-prototype object:

```js
// TODO: Remove after https://github.com/ariakit/ariakit/issues/7058 is fixed.
const safeSelectors = Object.assign(Object.create(null), selectors);
const state = useStoreStateObject(store, safeSelectors);
```

The first workaround requires knowing the polluted keys and descriptors to be configurable. The second only protects direct `useStoreStateObject` usage; it does not change third-party code that passes inherited values elsewhere.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the narrow own-property design</summary>

**Assessment**

Issue #7058 is a stable bug. Enumerable prototype extensions are uncommon, but the impact includes lost TagList semantics or accessible names, unrelated store subscriptions and selector execution, and malformed snapshot shapes. The supported scope is own enumerable string-keyed entries in caller-controlled prop and selector records.

The runtime fix remains local: it reuses the established ownership helper, adds no abstraction or persistent state, and preserves existing behavior for valid own entries. The repeated findings were omitted applications of the same ownership invariant plus release-metadata compliance, not evidence of excessive implementation complexity.

**Decision**

Continue the narrow own-property design and describe the shared release through the linked public `CompositeItem` surface. Symbols and non-enumerable properties remain outside the existing `for...in` contract; no follow-up issue is needed.

</details>

### DIFF
--- a/.changeset/300-inherited-record-properties.md
+++ b/.changeset/300-inherited-record-properties.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/react-store": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`CompositeItem`](https://ariakit.com/reference/composite-item) and similar components treating inherited enumerable properties as caller-supplied entries.
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -303,7 +303,15 @@ function InheritedObjectSelectors({ store }: StoreProps) {
       ),
     [],
   );
-  const selectorValues = useStoreStateObject(store, ["foo"], selectorObject);
+  const safeSelectorObject = useMemo(
+    () => Object.assign(Object.create(null), selectorObject),
+    [selectorObject],
+  );
+  const selectorValues = useStoreStateObject(
+    store,
+    ["foo"],
+    safeSelectorObject,
+  );
 
   const updateInheritedKey = () => {
     const calls = selectorCalls.current;
--- a/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
+++ b/app/src/sandbox/store-keyed-subscriptions/index.react.tsx
@@ -8,9 +8,13 @@ import type { Store } from "@ariakit/store";
 import type { SetState } from "@ariakit/utils";
 import { useEffect, useMemo, useReducer, useRef, useState } from "react";
 
+const objectToString = Reflect.get(Object.prototype, "toString");
+const objectValueOf = Reflect.get(Object.prototype, "valueOf");
+
 interface TestState {
   foo: number;
   bar: number;
+  prototypeMethod: typeof Object.prototype.toString;
 }
 
 interface SelectorCalls {
@@ -253,7 +257,7 @@ function DirectValues({ store }: StoreProps) {
 }
 
 function DynamicSelectors({ store }: StoreProps) {
-  const [key, setKey] = useState<keyof TestState>("foo");
+  const [key, setKey] = useState<"foo" | "bar">("foo");
   const value = useStoreState(store, [key], (state) => state[key]);
   const object = useStoreStateObject(store, [key], {
     value: (state) => state[key],
@@ -292,6 +296,8 @@ function InheritedObjectSelectors({ store }: StoreProps) {
             selectorCalls.current += 1;
             return state.foo;
           },
+          toString: "prototypeMethod" as const,
+          valueOf: () => objectValueOf,
         },
         {
           inheritedDirect: "bar" as const,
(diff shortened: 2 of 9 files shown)


diegohaz:
This is too edge case. Not worth the tests.

PR #7034, opened 2026-08-03
Fix explicitly undefined props overriding computed defaults

diegohaz:
## Motivation

Many Ariakit component hooks compute a default for a prop they do not destructure, then spread the caller's props over it:

```ts
// If the hovercard is modal, we should always autoFocus on show.
const autoFocusOnShow = useStoreState(
  store,
  ["autoFocusOnShow"],
  (state) => modal || state.autoFocusOnShow,
);

props = usePopover({
  store,
  modal,
  portal,
  autoFocusOnShow,
  ...props,
  portalRef,
  // ...
});
```

An own key holding `undefined` still wins that spread, so the downstream hook falls back to its own destructuring default (`autoFocusOnShow = true` in `usePopover`) instead of the value the component computed for itself. `<Hovercard autoFocusOnShow={undefined} />` therefore behaves differently from `<Hovercard />`.

That matters because the most common way to wrap an Ariakit component in React forwards optional props positionally:

```tsx
function MyHovercard({ autoFocusOnShow, ...props }: MyHovercardProps) {
  return <Ariakit.Hovercard autoFocusOnShow={autoFocusOnShow} {...props} />;
}
```

A consumer who never sets `autoFocusOnShow` on `MyHovercard` gets `true` rather than the context sensitive default, so the hovercard starts pulling keyboard focus into itself as soon as the pointer touches the anchor.

The same shape appears wherever a hook computes a value for a prop it does not destructure. Three cases are reproduced here:

| Component | Prop | Computed default | With a forwarded `undefined` |
| --- | --- | --- | --- |
| `Hovercard` | `autoFocusOnShow` | `modal \|\| state.autoFocusOnShow` | `true`, so the card steals focus on hover |
| `TabPanel` | `focusable` | `!store.composite && !hasTabbableChildren` | `true`, so the panel becomes a stray tab stop |
| `Checkbox` | `clickOnEnter` | `!nativeCheckbox` | `true`, so Enter toggles a native checkbox |

The reported component, `ComboboxPopover`, does not reproduce. `useComboboxList` returns `removeUndefinedValues(props)` earlier in its chain, so the key is stripped before it can reach `usePopover`. That protection is incidental to hook ordering rather than deliberate, which is what this pull request addresses.

Fixes #7028

## Solution

`forwardRef` now drops props holding `undefined` before handing them to the component:

```ts
const Role = React.forwardRef(
  (props, ref) => render(removeUndefinedValues({ ...props, ref })),
);
```

[...]

PR #7064, opened 2026-08-04
Remove redundant undefined prop filtering

### LINKED ISSUE #7037: The removeUndefinedValues calls at hook returns are largely redundant for the component path
## Context

Found while fixing #7028. That fix strips explicitly undefined props at the component boundary, in `forwardRef`, so a prop the caller never set is gone before the first hook runs.

## Problem

Around 20 hooks end with `return removeUndefinedValues(props);`. The pattern was introduced in [`939fca4`](https://github.com/ariakit/ariakit/commit/939fca43edb3a1f759abf5dd3c60f323affed0f2) so a hook's own undefined values could not clobber a computed default further down the chain, and it has been added one hook at a time since.

For the component path, the #7028 fix now subsumes most of that: the caller's undefined props never enter the chain, so the remaining work these calls do is stripping values the hook itself computed as `undefined`. Some of those are still load-bearing, but many are not, and the calls run on every render of every affected component.

Removing the redundant ones would delete real code and is a measurable win on a hot path: each call allocates a fresh object and walks every key.

## Why #7028 can close without it

#7028 needed the smallest change that fixes the reported behavior, and its own review already narrowed it once. Auditing 20 hook returns is a separate risk profile from a one line policy change, and getting it wrong reintroduces exactly the clobbering the fix removes.

The audit is also not purely mechanical. These calls still cover the path #7028 deliberately does not: a consumer calling a component hook directly, such as `useHovercard({ autoFocusOnShow: undefined })`, since `createHook` is unchanged. Any removal has to establish per hook that no such caller depends on it.

## Expected outcome

Audit each `return removeUndefinedValues(props)` call site and classify it: still needed because the hook itself emits `undefined` values that a downstream hook computes, still needed for the direct hook call path, or redundant now. Remove the redundant ones with regression coverage, and consider whether a single documented rule about where the helper belongs would prevent the pattern from being re-added ad hoc.

Relates to #7036, which covers the other uncovered path, the `render` element merge.

diegohaz:
## Motivation

#7034 moved undefined-prop filtering to the stable component boundary, but 24 experimental hook returns still called `removeUndefinedValues`, allocating a new object and walking its keys. Most calls are redundant on the component path, while intermediate hook boundaries must still remove undefined sentinels before later hooks compute defaults.

Fixes #7037

## Solution

Removed 18 redundant return-level filters and imports. Kept six filters in `useCheckbox`, `useRadio`, `useCompositeContainer`, `useCompositeTypeahead`, `useDisclosureContent`, and `useFocusable`, where later hook composition would otherwise let an undefined property overwrite a computed default.

Consolidated the coverage in the module-first `react-components-undefined-props` sandbox and added issue-linked browser and happy-dom tests for direct combobox, dialog, and toolbar hooks plus stable `FormCheckbox` and composed `FormRadio` behavior. Removed ambiguous issue links from unrelated legacy menubar tests.

## Tests

- `pnpm build`
- `pnpm clean`
- `pnpm tsc`
- Focused current React tests (1 file, 12 tests)
- Focused React 18 tests (1 file, 12 tests)
- Desktop Playwright on Chrome, Firefox, and Safari (36 tests)
- Full Vitest (253 files and 1,699 relevant tests passed; two unrelated local child-process readiness tests could not start, while the published PR head's complete CI passed them)
- Red proof: removing the retained `useCheckbox`, `useRadio`, `useCompositeContainer`, or `useDisclosureContent` boundary filters makes the corresponding targeted regression fail; prior proof also covers `useFocusable` and `useCompositeTypeahead`

## Performance

The exact published-head performance workflow reported no significant Chrome or Node changes.

<details>
<summary>Cycle 3: Continue the narrowed filter audit and simplify fixture naming</summary>

**Assessment**

Issue #7037 is a broad hot-path allocation audit. The first reviewed state removed all return filters too broadly and regressed stable `FormCheckbox` behavior plus direct Dialog and Toolbar hook composition. The current state restores the six confirmed chain boundaries while removing 18 redundant filters. Runtime complexity remains lower than the base: no new state, branches, helpers, abstractions, or public machinery were introduced. Maintenance cost is limited to targeted scenarios duplicated across the required browser and happy-dom surfaces.

**Decision**

Continue the narrowed implementation without reverting or narrowing supported behavior. Preserve stable component behavior and later computed defaults at the six retained boundaries. Do not claim a quantified end-to-end performance improvement without benchmark evidence.

Rename the consolidated sandbox to `react-components-undefined-props`. The prior component enumeration became stale, while the behavior-only alternative `undefined-props-computed-defaults` does not satisfy the repository's component-or-module-first naming rule. Undefined-valued keys remain intentionally filtered at load-bearing intermediate boundaries; the change does not promise uniform undefined-key preservation from every experimental hook.

</details>


packages/ariakit-react-components/src/checkbox/checkbox.tsx
```diff
@@ -176,12 +176,12 @@ export const useCheckbox = createHook<TagName, CheckboxOptions>(
...
     props = useCommand<TagName>({ clickOnEnter: !nativeCheckbox, ...props });
 
-    return removeUndefinedValues({
+    return {
       name: nativeCheckbox ? name : undefined,
       value: nativeCheckbox ? valueProp : undefined,
       checked: isChecked,
       ...props,
-    });
+    };
```
diegohaz: Dropping the filter here breaks [`FormCheckbox`](https://ariakit.com/reference/form-checkbox) on the plain component path, not just in direct hook composition.

`useFormCheckbox` calls this hook without a `name` option:

```tsx
props = useCheckbox({ store: checkboxStore, value, checked, ...props });

props = useFormControl({
  store: form,
  name,
  "aria-labelledby": undefined,
  ...props,
});
```

So `name` on line 180 evaluates to `undefined` and this return now ships it as an **own key**. `useFormControl` places its real `name` *before* `...props`, so that own `undefined` key wins the spread, and `useFormControl` destructures `name: nameProp` as `undefined`. [`useFormRadio`](https://github.com/ariakit/ariakit/blob/110923faf8f69149c1efea31b70cbd1a515c013d/packages/ariakit-react-components/src/form/form-radio.tsx#L67) is unaffected only because it passes `name` into `useRadio`.

This is the general hazard behind the change: `forwardRef` filters props **entering** a component, but nothing filters props **leaving** a hook. These return-level filters were resolving the `key: cond ? x : undefined` sentinels the hooks synthesize themselves, which is a different job from the one #7034 moved to the component boundary.

Measured on this fixture, rendered with `<Form>` / `<FormLabel name={form.names.accept}>` / `<FormCheckbox name={form.names.accept} />`:

```
main    <label id="_r_0_" for="_r_2_">Accept terms</label>
        <input aria-invalid="false" aria-checked="false" id="_r_2_" type="checkbox" name="accept">

110923f <label id="_r_0_">Accept terms</label>
        <input aria-invalid="false" aria-checked="false" id="_r_2_" type="checkbox" name="undefined">
```

Real-world impact for anyone using `<Form>` with a checkbox field: the control submits under the literal field name `undefined` instead of `accept`, so native submission and `FormData` carry the wrong key; `FormLabel` loses its `for`, leaving the checkbox with no accessible name at all (`getByRole("checkbox", { name: "Accept terms" })` no longer matches it); and the field registers under the wrong name for validation and touched state. The existing `app/src/sandbox/form-checkbox-integer-like-name` fixture nests the input inside the label, so implicit labelling keeps it green and the suite never sees this.

Restoring the filter on this return alone fixes all three symptoms (verified locally). Passing `name` through, mirroring `useFormRadio`, fixes this chain too but leaves the `value` sentinel and the same hazard elsewhere. If the goal is to keep the allocation savings, the durable fix is to stop synthesizing `undefined` sentinels here and only create the keys that should exist:

```ts
const ownProps: typeof props = { checked: isChecked };
if (nativeCheckbox) {
  ownProps.name = name;
  ownProps.value = valueProp;
}
return { ...ownProps, ...props };
```

**Assessment**

- Severity: Critical (verified, 95% confidence), because a stable `@ariakit/react` component silently submits the wrong field name and loses its accessible name.
- Priority: P0 (estimate, 90% confidence), because it is a shipped-behavior regression on the default component path with no opt-out.
- Breakage likelihood: Likely (verified, 90% confidence), because it needs only `<FormCheckbox name={...} />` with a non-wrapping `FormLabel`, which is the documented usage.
- Fix effort: Easy (verified, 90% confidence), because restoring this one filter resolves it; the sentinel rework is Moderate.

packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
```diff
@@ -292,7 +292,7 @@ export const useDisclosureContent = createHook<
     style,
   };
 
-  return removeUndefinedValues(props);
+  return props;
```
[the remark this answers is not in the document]
diegohaz: Confirmed. I reproduced this against both refs with a direct hook fixture:

```tsx
function HookDialog(props: DialogProps) {
  const htmlProps = useDialog(props);
  return <ak.Role {...htmlProps} />;
}

<ak.DialogProvider open>
  <HookDialog aria-label="Hook dialog" disabled aria-disabled={undefined} modal={false} portal={false}>
    Dialog content
  </HookDialog>
</ak.DialogProvider>
```

Asserting `aria-disabled="true"` on the dialog passes on `main` (0beb71cf9) and fails on this head (110923faf), where the attribute is absent.

The chain is the one you describe. `useDialog` composes `useDisclosureContent` and then `useFocusable`, and `useFocusable` places its computed value before the spread:

```ts
props = {
  "data-focus-visible": (focusable && focusVisible) || undefined,
  "data-autofocus": autoFocus || undefined,
  "aria-disabled": disabled || undefined,
  ...props,
```

`disabled` there is `focusable && disabledFromProps(props)`, so it is `true`, but the own `aria-disabled` key holding `undefined` wins the spread and `removeUndefinedValues` at the return then drops the key entirely. Since the dialog defaults to a `div`, `disabled` is not rendered either, so the element exposes no disabled state.

One correction to the framing: the retained filter in `useFocusable` cannot cover this, because it runs *after* the spread. It protects hooks composed around `useFocusable` from the sentinels `useFocusable` itself emits; it does not protect `useFocusable`'s own computed defaults from an incoming key.

Disposition: Address now, but not by restoring only this filter. The same mechanic is confirmed on the component path in [`useCheckbox`](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711017565), so the fix should be chosen for the class rather than per chain.

**Assessment**

- Severity: Medium (verified, 85% confidence), because a disabled dialog composed through the hook exposes no disabled semantics to assistive technology.
- Priority: P1 (estimate, 75% confidence), because it is a behavior change against `main` on an experimental but documented hook surface.
- Breakage likelihood: Possible (estimate, 70% confidence), because it needs direct `useDialog` composition that forwards an explicitly `undefined` optional prop, which is the wrapper pattern the sandbox in this PR already models.
- Fix effort: Easy (verified, 85% confidence), because restoring one return-level filter resolves it.

packages/ariakit-react-components/src/composite/composite-container.tsx
```diff
@@ -250,7 +249,7 @@ export const useCompositeContainer = createHook<
     onClick,
   };
 
-  return removeUndefinedValues(props);
+  return props;
```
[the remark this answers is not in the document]
diegohaz: Confirmed. Reproduced against both refs with a direct hook fixture:

```tsx
function HookToolbarContainer(props: ToolbarContainerProps) {
  const htmlProps = useToolbarContainer(props);
  return <ak.Role {...htmlProps} />;
}

<ak.Toolbar aria-label="Hook toolbar">
  <HookToolbarContainer aria-label="Hook container" disabled aria-disabled={undefined}>
    <input type="text" aria-label="Inner input" />
  </HookToolbarContainer>
</ak.Toolbar>
```

Asserting `aria-disabled="true"` on the container passes on `main` (0beb71cf9) and fails on this head (110923faf).

The composition is `useToolbarContainer` -> `useCompositeContainer` -> `useToolbarItem` -> `useCompositeItem` -> `useCommand` -> `useFocusable`, and after this PR nothing in that chain filters before `useFocusable` computes `"aria-disabled": disabled || undefined` ahead of its `...props` spread. The own key holding `undefined` wins, then the return-level filter drops the key. Because the container defaults to a `div`, no native `disabled` is rendered either, so the element exposes no disabled state.

Disposition: Address now, together with the other occurrences rather than by restoring this single filter. The same mechanic is confirmed on the plain component path in `useCheckbox`, where `FormCheckbox` ends up rendering `name="undefined"`, so the class needs one decision.

**Assessment**

- Severity: Medium (verified, 85% confidence), because a disabled toolbar container exposes no disabled semantics to assistive technology.
- Priority: P1 (estimate, 75% confidence), because it is a behavior change against `main` on an experimental but documented hook surface.
- Breakage likelihood: Possible (estimate, 65% confidence), because it needs direct `useToolbarContainer` composition forwarding an explicitly `undefined` optional prop.
- Fix effort: Easy (verified, 85% confidence), because restoring one return-level filter resolves it.

.changeset/300-component-hook-undefined-values.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ariakit/react-components": patch
+---
+
+Updated experimental component hooks to preserve properties whose values are `undefined` in their returned props unless removing them is required to protect a later computed default.
```
diegohaz, line 5: Two problems with this entry, both downstream of the correctness finding on `useCheckbox`.

**The frontmatter is incomplete for the behavior this branch currently ships.** `FormCheckbox` is re-exported by `@ariakit/react` (`packages/ariakit-react/src/form.ts:10`), and its rendered `name` attribute and label association change on this head, so `@ariakit/react` receives the change through its public entrypoint and belongs in the frontmatter. If the `useCheckbox` regression is fixed so no stable component behavior changes, this becomes moot and the opposite question applies: with the observable effect limited to experimental hook returns, this may be an internal refactor that needs no changeset at all.

**The entry describes the mechanism instead of the outcome, and the mechanism is the regression vector.** "preserve properties whose values are `undefined` in their returned props" is what causes `name="undefined"` to reach the DOM, and "unless removing them is required to protect a later computed default" asks readers to reason about internal hook ordering they cannot see.

This is also motivated as a performance change (#7037, "allocating a new object and walking its keys"), but neither the changeset nor the PR description carries a measurement. `ariakit-general-changeset` asks performance entries to quantify the result with a percentage or multiplier and name the benchmark or workload. Worth landing a number here, since it is what justifies trading these filters away.

**Assessment**

- Severity: Low (estimate, 80% confidence), because it misroutes changelog and versioning rather than breaking runtime behavior.
- Priority: P1 (estimate, 80% confidence), because a missing dependent package means `@ariakit/react` consumers get the changed behavior without a released note.
- Breakage likelihood: Likely (verified, 90% confidence), because the frontmatter as written cannot version `@ariakit/react` at all.
- Fix effort: Easy (estimate, 90% confidence), because it is a frontmatter line plus a rewritten sentence, once the correctness decision is made.

app/src/sandbox/menubar-interactions/test.ts
```diff
@@ -1,6 +1,7 @@
 import { click, hover, press, q, sleep, type } from "@ariakit/test";
 import { expect, test } from "vitest";
 
+// https://github.com/ariakit/ariakit/issues/7037
```
diegohaz: This file has seven tests, and the issue URL sits above only the first one, so it reads as a group-level comment for the whole file rather than traceability for a specific regression. `ariakit-ariakit-example-playwright-test` asks for the URL immediately above every test it documents, repeated per test instead of one group-level comment.

It is also not clear which tests this issue actually documents. The PR describes these as the negative proof for the retained `useFocusable` filter, and that proof was reported as 36 menu failures, so the link most likely belongs above the specific tests that go red without the filter rather than above `show/hide on click` alone. Same question applies to the single link added in `test-browser.ts`.

Worth confirming which tests in this file actually regress without the retained filter and annotating exactly those.

**Assessment**

- Severity: Low (estimate, 85% confidence), because it only affects how a future reader traces the regression back to its issue.
- Priority: P2 (estimate, 85% confidence), because it is documentation traceability, not behavior.
- Breakage likelihood: Rare (estimate, 80% confidence), because it misleads only when someone later edits these tests and needs to know what they protect.
- Fix effort: Easy (estimate, 90% confidence), because it is comment placement once the affected tests are identified.

diegohaz (commented):
## Review summary

Initial review. Baseline is the merge base `0beb71cf9` to head `110923faf` (single commit), then expanded into the surrounding hook compositions, the `forwardRef` boundary from #7034, and the app fixtures that cover this behavior.

### Headline

The premise needs one correction, and it is load-bearing for the whole change. `forwardRef` filters props **entering** a component; it does not filter props **leaving** a hook. #7034 moved the first job to the component boundary, but these 22 return-level filters were doing the second one: resolving the `key: cond ? x : undefined` sentinels that the hooks synthesize themselves. Removing them is safe only where no later hook in the same composition computes a default for that key, and that condition does not hold everywhere.

The most serious consequence is on the plain component path, not in direct hook composition: **`FormCheckbox` now renders `name="undefined"` and loses its label association.**

```
main    <label id="_r_0_" for="_r_2_">Accept terms</label>
        <input aria-invalid="false" aria-checked="false" id="_r_2_" type="checkbox" name="accept">

110923f <label id="_r_0_">Accept terms</label>
        <input aria-invalid="false" aria-checked="false" id="_r_2_" type="checkbox" name="undefined">
```

`useFormCheckbox` calls `useCheckbox` without a `name`, so the new unfiltered return ships an own `name: undefined` that beats the real field name `useFormControl` places before its spread. `useFormRadio` escapes only because it passes `name` into `useRadio`. Restoring the filter on that one return fixes all three symptoms locally.

### Published inline

- `packages/ariakit-react-components/src/checkbox/checkbox.tsx:184` — `FormCheckbox` renders `name="undefined"`, loses `FormLabel`'s `for`, and registers the field under the wrong name. Critical / P0.
- `.changeset/300-component-hook-undefined-values.md:5` — frontmatter omits `@ariakit/react` for behavior that reaches it, the entry describes the mechanism rather than the outcome, and the performance motivation carries no measurement. Low / P1.
- `app/src/sandbox/menubar-interactions/test.ts:4` — one issue URL above the first of seven tests reads as a group-level comment, and it is unclear which tests the issue documents. Low / P2.

Both Codex threads are **confirmed**, reproduced against both refs, and answered on their threads.

- `disclosure-content.tsx:295` (thread `PRRT_kwDOBjT0Fc6WQl54`) — direct `useDialog` composition with `{ disabled: true, "aria-disabled": undefined }` loses the computed `aria-disabled`. Passes on `0beb71cf9`, fails on `110923faf`. Address now, as part of the class rather than by restoring this one filter. One correction to the thread's framing: the retained `useFocusable` filter cannot cover this, because it runs after the spread; it protects hooks composed around `useFocusable`, not `useFocusable`'s own computed defaults.
- `composite-container.tsx:252` (thread `PRRT_kwDOBjT0Fc6WQl57`) — same mechanic through `useToolbarContainer` -> `useCompositeContainer` -> `useToolbarItem` -> `useCompositeItem` -> `useCommand` -> `useFocusable`. Passes on `0beb71cf9`, fails on `110923faf`. Address now.

### Direction

Per-chain auditing is what let these through, and it leaves no invariant behind that a future hook reordering would trip. Two coherent options:

1. Restore the return-level filters and keep the guarantee, accepting the allocations. Safe, but gives up the stated goal.
2. Keep the removals and stop synthesizing `undefined` sentinels inside the hooks, building each object with only the keys that should exist. That reaches the perf goal without depending on a downstream filter, e.g. in `useCheckbox`:

```ts
const ownProps: typeof props = { checked: isChecked };
if (nativeCheckbox) {
  ownProps.name = name;
  ownProps.value = valueProp;
}
return { ...ownProps, ...props };
```

Whichever direction, the missing regression coverage is worth closing: the only `FormCheckbox` fixture (`app/src/sandbox/form-checkbox-integer-like-name`) nests the input inside its `FormLabel`, so implicit labelling keeps it green with `name="undefined"` and the 1697-test run never sees the break. A fixture with a non-wrapping `FormLabel` asserting the rendered `name` and the `for`/`id` association would have caught it.

### Checked and clear

Swept every hook whose filter was removed for a following hook call in the same composition, then diffed rendered markup between `0beb71cf9` and `110923faf` for `FormRadio`, `MenuItemCheckbox`, `MenuItemRadio`, `Tab`, and `TabPanel`. All identical. The `render` prop path is also unaffected: `mergeProps` skips `undefined` overrides and nested Ariakit elements re-enter through `forwardRef`.

The two retained filters are justified and correctly placed. `usePopover` computes `autoFocusOnShow` before its `...props`, so the `useCompositeTypeahead` filter is genuinely load-bearing for `useComboboxPopover`, and the new sandbox test covers exactly that. The fixture consolidation from `hovercard-tab-panel-checkbox-7028` to `combobox-hovercard-tab-panel-checkbox-undefined-props` follows the fixture lifecycle correctly, and per-test issue URLs are preserved in the renamed fixture.

### Not fully validated

I did not measure the performance benefit, so I cannot say what the 22 removed filters are worth against the regression risk. No browser or React 18 run either; findings above come from the react happy-dom project plus static analysis of the composition order.

Commands: `pnpm install`, `pnpm test` against a temporary sandbox exercising the affected chains on both refs, `pnpm test form-checkbox-integer-like-name`. The temporary fixture was removed; no experiment changes remain.

packages/ariakit-react-components/src/disclosure/disclosure-content.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`2ca2656fa`](https://github.com/ariakit/ariakit/commit/2ca2656fa19ede173b1d84f2758c04eeb6a237dc): restored the `useDisclosureContent` boundary filter before `useDialog` reaches `useFocusable`, and added targeted direct-`useDialog` coverage for the computed `aria-disabled` state.

packages/ariakit-react-components/src/composite/composite-container.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`2ca2656fa`](https://github.com/ariakit/ariakit/commit/2ca2656fa19ede173b1d84f2758c04eeb6a237dc): restored the `useCompositeContainer` boundary filter before toolbar composition reaches `useFocusable`, and added targeted direct-`useToolbarContainer` coverage for the computed `aria-disabled` state.

packages/ariakit-react-components/src/checkbox/checkbox.tsx
diegohaz: Fixed in [`2ca2656fa`](https://github.com/ariakit/ariakit/commit/2ca2656fa19ede173b1d84f2758c04eeb6a237dc): restored the `useCheckbox` boundary filter so `FormCheckbox` keeps its computed field name, with targeted stable-component coverage for `name="newsletter"` and the external accessible label.

.changeset/300-component-hook-undefined-values.md
diegohaz: Already handled in [`.changeset/300-component-hook-undefined-values.md`](https://github.com/ariakit/ariakit/blob/2ca2656fa19ede173b1d84f2758c04eeb6a237dc/.changeset/300-component-hook-undefined-values.md): after fixing the stable component regressions, the remaining change is the consumer-observable return-key shape of public experimental `@ariakit/react-components` hooks, so the original patch changeset remains correctly scoped and does not claim a measured performance result.

app/src/sandbox/menubar-interactions/test.ts
diegohaz: Fixed in [`2ca2656fa`](https://github.com/ariakit/ariakit/commit/2ca2656fa19ede173b1d84f2758c04eeb6a237dc): removed the ambiguous #7037 links from the legacy menubar tests. The confirmed chains now have issue links immediately above their targeted happy-dom and browser regressions.

diegohaz:
**Fixed**

- Restored the four required intermediate filters in [`2ca2656fa`](https://github.com/ariakit/ariakit/commit/2ca2656fa19ede173b1d84f2758c04eeb6a237dc): `useCheckbox`, `useRadio`, `useCompositeContainer`, and `useDisclosureContent`. Targeted happy-dom and browser regressions now cover each concrete composition chain.
- Replied with the direct-hook evidence for [`useDisclosureContent`](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711950379) and [`useCompositeContainer`](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711953574), plus stable [`FormCheckbox`](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711956612) coverage.
- Removed the ambiguous issue links from legacy menubar tests and documented the traceability outcome [in the thread](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711963913).

**Already resolved**

- Retained the original `@ariakit/react-components` patch changeset because the public experimental hook return-key shape remains consumer-observable after stable component regressions are fixed. The final scope and rationale are recorded [in the changeset thread](https://github.com/ariakit/ariakit/pull/7064#discussion_r3711960268).

The final inventory is 18 redundant filters removed and six composition boundaries retained. Verification passed for focused current React (12/12), isolated React 18 (12/12), desktop Chrome/Firefox/Safari (36/36), build, clean, and TypeScript. Full Vitest passed 253 files and 1,699 tests; only the same two unrelated local child-process readiness timeouts remained.

app/src/sandbox/react-components-undefined-props/index.react.tsx
```diff
@@ -63,6 +84,61 @@ export default function Example() {
...
+          role="group"
+          aria-label="Hook toolbar container"
+          aria-disabled={undefined}
+          disabled
+        >
+          <input aria-label="Inner input" />
+        </HookToolbarContainer>
+      </ak.Toolbar>
+
+      <h2>Form checkbox with an external label</h2>
```
diegohaz, line 120: This commit adds dialog, toolbar container, form checkbox, and form radio scenarios, but the directory name still enumerates only `combobox-hovercard-tab-panel-checkbox`. `ariakit-ariakit-workflow` asks the folder name to start with the components or modules the sandbox uses, ordered by importance, so the name is now stale for four of its eleven scenarios, including the two that cover the `FormCheckbox` field-name regression.

The practical cost is discoverability: the fixture lifecycle in `ariakit-ariakit-example-playwright-test` tells the next person to search existing fixtures for the component and behavior before creating a new one, and someone searching for `FormCheckbox` or `Form` coverage will not find this directory. That is the same failure mode that produced the duplicate-fixture problem the rename was meant to solve.

Since the component list is now long enough that enumerating it is not useful, the behavior suffix is doing the real identifying work. Something like `undefined-props-computed-defaults` describes what every scenario here has in common and stays stable as more chains are added.

I recognize this fixture was already renamed once in this PR and that renaming again means touching the preview ID, both test files, and any filters. Reasonable to decline on churn grounds; flagging it so the choice is deliberate rather than accidental.

**Assessment**

- Severity: Low (estimate, 85% confidence), because it affects discoverability rather than behavior or coverage.
- Priority: P2 (estimate, 85% confidence), because nothing breaks and the coverage itself is correct.
- Breakage likelihood: Possible (estimate, 70% confidence), because it misleads only when someone later searches for form or dialog coverage before adding a fixture.
- Fix effort: Easy (estimate, 80% confidence), because it is a directory rename plus path references, though it is the second rename this cycle.

diegohaz (commented):
## Re-review summary

Re-review. Incremental baseline is `110923faf` from my previous marked review, so new findings come from `2ca2656fa` ("Preserve later hook defaults") plus surrounding code that commit newly exposes.

**Every finding from the previous review is fixed and independently verified.** The four restored filters are the right call, the new coverage is proven regression coverage, and I found no new defects.

### Prior findings, revalidated against `2ca2656fa`

All five threads were resolved by the author. I re-verified each against the current head rather than relying on the replies.

- **`useCheckbox` / `FormCheckbox` (Critical, P0)** — Fixed. `<FormCheckbox name={form.names.accept} />` again renders `name="accept"` with `<label for="…">` restored, so the control submits under the right key and recovers its accessible name.
- **`useDisclosureContent` / direct `useDialog`** — Fixed. `aria-disabled="true"` is present again.
- **`useCompositeContainer` / direct `useToolbarContainer`** — Fixed. `aria-disabled="true"` is present again.
- **Changeset frontmatter** — Obsolete, and the author's reasoning checks out. My concern was explicitly conditional on stable component behavior changing; it no longer does, so `@ariakit/react` correctly stays out of the frontmatter. Confirmed by diffing rendered markup, not by inference.
- **Menubar issue links** — Fixed. Both ambiguous links removed, and the confirmed chains now carry issue links directly above their own regressions.

`useRadio` was also restored, which goes beyond what I reported. That turns out to be load-bearing: `useFormRadio` does pass `name` into `useRadio`, so the native input case I tested was clean, but a `FormRadio render={<div />}` makes `nativeRadio` false and the hook emits `name: undefined` again. The new "composed form radio" scenario covers exactly that, and it fails with `name="undefined"` without the restored filter. Good catch.

### Verification performed

- **Broad markup differential.** Built a fixture rendering 76 elements across `Form`, `Menu` (item, checkbox, radio, group), `Select`, `Combobox` (group, item, item value), `Tab`/`TabPanel`, `Radio`, `Checkbox`/`CheckboxCheck`, `Popover` (arrow, disclosure arrow, heading, description), `Composite` (row, item, input), `Group`/`GroupLabel`, plus the direct `useDialog` and `useToolbarContainer` hooks. Dumped every element's sorted attributes on `0beb71cf9` and `2ca2656fa`: **byte-identical**. The component path now matches `main` exactly.
- **Red proof of the new tests.** Checked out `packages/` at `110923faf` while keeping the new fixture and tests, then ran the suite. Exactly the four new tests fail and the other eight pass, each for the expected reason: missing `aria-disabled` on the dialog and the toolbar container, the form checkbox not resolvable by accessible name, and `name="undefined"` on the composed form radio. They pass on `2ca2656fa`. That satisfies the regression-proof bar in `ariakit-general-testing`.
- **Full Vitest.** 254 files, 1701 tests, all passing locally, including the two child-process readiness tests the PR description notes could not start locally.
- **Residual-hole sweep.** Re-derived every composition where a hook whose filter is still removed is followed by another hook call. The only remaining sibling chains are `useCompositeItem -> useCompositeHover` in `menu-item`, `select-item`, and `combobox-item`. `useCompositeHover` spreads `...props` first and computes nothing ahead of it, so an incoming undefined key cannot displace a default there. No remaining holes in the 18 removals.

One clarification worth recording, since it explains why only these chains were affected: a return-level filter never protected its own hook's internal `useCommand`/`useFocusable` calls, because it runs after them. It only protected *sibling* hooks composed after it. That is why a direct `useCompositeItem({ "aria-disabled": undefined, disabled: true })` loses the computed value on `main` too and is not a regression, while the four fixed chains were.

### New finding

- `app/src/sandbox/…-undefined-props/index.react.tsx:120` — the directory name no longer covers the dialog, toolbar, and form scenarios this commit added, which hurts fixture discoverability. Low / P2, and reasonable to decline on rename churn.

### Open trade-off, author's call

The PR now reports that the performance workflow "reported no significant Chrome or Node changes" — an honest disclosure, and worth taking seriously against the motivation. As it stands the change removes 18 filters to eliminate allocations that measurement cannot detect, and doing so already cost three confirmed regressions, one of them Critical on a stable public component.

That is not an argument that the current state is unsafe; it is well covered now. It is a question of whether the residual maintenance property is worth an unmeasurable gain. The safety of the 18 removals rests on an invariant nothing enforces: no hook may return an own key holding `undefined` when a sibling hook composed after it computes a default for that key. Today six filters and eleven fixture scenarios encode that by hand, and a future reordering, a new `key: cond ? x : undefined` sentinel, or a new hook inserted into a chain can silently reintroduce exactly this bug with no failing test.

Two ways to make that durable, if you want it:

1. Land a targeted micro-benchmark on a workload where the allocations should show up (a large `Composite` or `Combobox` list re-rendering) so the changelog and #7037 rest on a number rather than on reasoning about redundancy.
2. Remove the invariant instead of documenting it, by not synthesizing `undefined` sentinels in the six retained hooks and building those objects with only the keys that should exist. That lets all 24 filters go and makes the property structural rather than audited.

Either is fine, as is deciding the current balance is good enough. Flagging it because the measurement changes the cost/benefit that motivated the change, not because anything here is broken.

### Not validated

No browser or React 18 run this round; the PR reports 36 desktop Playwright tests and a 12-test React 18 run passing. I did not reproduce the performance workflow, so I am taking the "no significant change" result as reported.

app/src/sandbox/react-components-undefined-props/index.react.tsx
diegohaz: Fixed in [`bf842bedd`](https://github.com/ariakit/ariakit/commit/bf842bedd8a0ca992bee972272c0a5d7ea867ac9): renamed the fixture to `react-components-undefined-props`, which starts with the audited Ariakit module and keeps the stable shared behavior suffix without enumerating every scenario. The renamed preview collected all 12 happy-dom tests and all 36 Chrome, Firefox, and Safari cases.

diegohaz:
**Fixed**

- [`bf842bedd`](https://github.com/ariakit/ariakit/commit/bf842bedd8a0ca992bee972272c0a5d7ea867ac9) ([reply](https://github.com/ariakit/ariakit/pull/7064#discussion_r3712461643)): Renamed the consolidated fixture to the module-first `react-components-undefined-props` path. The new preview ID passes 12 happy-dom tests and 36 desktop browser cases.

diegohaz (commented):
## Re-review summary

Re-review. Incremental baseline is `2ca2656fa` from my previous marked review, so new findings come from `bf842bedd` ("Clarify undefined props sandbox name").

**No new findings.** The only outstanding item from the last round is resolved, and nothing else changed.

### What changed

`bf842bedd` is a pure rename, `R100` on all three files with zero content change: `combobox-hovercard-tab-panel-checkbox-undefined-props` becomes `react-components-undefined-props`. `packages/` and `.changeset/` are byte-identical to `2ca2656fa`, so the library verification from my previous review still stands unchanged.

### Prior finding, revalidated

- **Fixture naming (Low, P2)** — Resolved, and resolved better than I proposed. I suggested `undefined-props-computed-defaults`; your reassessment note correctly points out that a behavior-only name does not satisfy the repository's component-or-module-first rule in `ariakit-ariakit-workflow`. `react-components-undefined-props` leads with the module and keeps the behavior suffix doing the identifying work, which satisfies the convention where my suggestion would have broken it. Withdrawing the suggestion.

### Verification performed

- **Rename completeness.** No references to either the old consolidated name or the original `hovercard-tab-panel-checkbox-7028` name survive anywhere in the repo, across TS, TSX, JSON, MD, MDX, Astro, and YAML.
- **Preview routing.** `getPreviewId` in `app/src/test-utils/preview.ts` derives the id from the directory path relative to the preview roots, and `getPreviewFrameworksSync` infers frameworks from the `index.react.tsx` entry file, so the preview becomes `/react/previews/react-components-undefined-props/` with no manifest to update. Confirmed no generated file enumerates sandbox ids; discovery is dynamic.
- **Test collection.** The renamed file is still collected and passes, 12/12 at the new path, so the vitest entry-file glob followed the rename.
- **Full Vitest.** 254 files and 1701 tests passing, identical to the counts I measured on `2ca2656fa`, so no suite lost collection in the move.
- **Final inventory.** 18 filters removed and 6 retained, in `useCheckbox`, `useRadio`, `useCompositeContainer`, `useCompositeTypeahead`, `useDisclosureContent`, and `useFocusable`. Matches the PR description exactly.
- **CI.** 45 checks passing, 1 skipped, and `App / Test Safari (2/2)` still pending at review time. No failures.

### Open trade-off from the last round, closed

I raised whether 18 removals earn their keep when the workflow measures no significant change. That is now addressed on the record rather than left implicit: the perf bot reports no significant changes for both Chrome (27 tests) and Node, the PR carries a Performance section stating it, and the cycle 3 reassessment explicitly decides not to claim a quantified end-to-end improvement without benchmark evidence.

That is a legitimate resolution of the question I asked. The motivation is removing provably redundant work rather than a measurable win, the changelog does not overclaim, and the six retained boundaries plus eleven fixture scenarios document where the property is load-bearing. Recording it as a deliberate, documented decision, not an unresolved concern.

### Not validated

No browser or React 18 run locally this round; the PR reports 36 desktop Playwright cases and 12 React 18 tests passing at the renamed path, and CI corroborates Chrome, Firefox, Safari shard 1, iOS, and Android. I did not reproduce the performance workflow and am taking the bot's published result as reported.

PR #7065, opened 2026-08-04
Fix composite presentation when focused item is replaced

### LINKED ISSUE #7042: Composite: a presentation drops a replaced item when the item itself held DOM focus
Raised on https://github.com/ariakit/ariakit/pull/7030, which fixes the same drop for virtual-focus composites. That pull request scopes its claims to leave this case out rather than widening `stillOwnsFocus`, because the outcome here is identical before and after it.

A presentation created with `focus: true` and `requireFocus: true` moves DOM focus onto its item. When the app then replaces that item's node under the same id, the browser parks focus on `<body>`, and the ownership check reads that as focus escaping rather than as the request's own item being torn out from under it. The request cancels, so the popup stays scrolled where it was.

`Menu` is the case this reaches by default, because `menu-store.ts` never sets `virtualFocus` and inherits the `false` default from `createCompositeStore`. Items therefore take real DOM focus, and `CompositeFocusOnMove` passes `requireFocus: ownsFocus(store)`, which is `true` for any move made while focus is already inside the composite.

## Evidence

With #7030 applied and `presentItem` instrumented, a `Menu` whose `updatePosition` resolves on a timer, opened and then navigated with `ArrowUp`, `Home`, `End`, with every item node re-keyed under stable ids:

```
{"event":"pass",         "requireFocus":true,"focused":true,"activeTag":"BODY","hasElement":true,"connected":false}
{"event":"reresolve",    "requireFocus":true,"focused":true,"activeTag":"BODY","found":true}
{"event":"cancelByFocus","requireFocus":true,"focused":true,"activeTag":"BODY","target":"replacing-actions-rename"}
```

| Variant | `menu.scrollTop` after positioning finishes |
| --- | --- |
| Control, item nodes kept | `142` |
| Item nodes replaced under stable ids | `0` |

The re-resolution added by #7030 succeeds, and `stillOwnsFocus` then cancels: `activeElement` is neither the owner nor the replacement, and `ownsFocus` is false because the body is inside neither the composite element nor the popup.

## Why it is separable

This is not a regression. Before #7030 the same wake cancelled at the disconnected-element check instead, so a replaced item was never presented in this configuration either. #7030 changes the virtual-focus path only, which is `Combobox` and `Select`.

Reachability is also unestablished. The reproduction above widened the placing window deliberately; nobody has shown that a default synchronous placement leaves enough room for a replacement to land inside it.

## What to decide

Whether "the focused node was removed" should be distinguished from "the user moved focus away". The browser parks focus on the body only in the first case, so when `focused` is already true and the active element is the body at the moment the cached element is found disconnected, the escape is this request's own replacement. Handing focus to the replacement would restore the user's keyboard position as well as the scroll, which is the stronger fix.

The cost is a focus-provenance signal inside a function whose ordering is already delicate, and which just gained a focus-escape guarantee in #7031, so it needs a fixture that pins both behaviors at once. Today no fixture combines real focus inside the composite with item replacement: `composite-presentation-cancel`'s `HeldMenu` drives everything from buttons outside the composite on purpose, so `requireFocus` is false there.

Related: #7018 on how wide "the composite" should be for this same ownership check.

### Workaround

See the [current workaround and its limitations](https://github.com/ariakit/ariakit/issues/7042#issuecomment-5166186239).

diegohaz:
## Motivation

When an application replaces the focused DOM node for a menu item under the same logical id while its popup is still positioning, the browser moves focus to the document body. The pending presentation treated that as a real focus escape and stopped, so the replacement kept the active state but received neither focus nor the delayed scroll.

```tsx
<Fragment key={generation}>
  <Ariakit.MenuItem id="stable-action-id">Action</Ariakit.MenuItem>
</Fragment>
```

Fixes #7042

## Solution

Track the brief ref-cleanup window for each composite item and listen for focus changes directly on the element a presentation request focuses. If React disconnects that focused node during its ref cleanup, the request re-resolves the same logical item, restores focus to the replacement, and completes its pending scroll presentation. A blur that occurs outside that replacement window remains terminal, so an explicit focus escape is never turned into a new focus request.

This stays local to the active presentation request: a transient `WeakSet` marker and paired item `blur` and `focus` listeners replace the broader document-level focus state used in the earlier attempt. The focus listener keeps the request's focus-left state current when an item blurs and then regains focus before replacement.

The public `Menu` regression uses stable item ids and a held positioning promise. It verifies the relevant behavior:

- replacing the focused node preserves focus immediately and brings the replacement into view when positioning finishes;
- explicitly blurring before replacement leaves focus outside and does not scroll the replacement into view;
- blurring and refocusing the item before replacement restores focus to the same-id replacement.

Removing the explicit-blur guard or the refocus listener makes the corresponding happy-dom and browser cancellation tests fail, confirming that coverage protects both distinctions.

## Workaround

Prefer keeping item nodes stable while the popup is positioning. If replacement is unavoidable on a version without this fix, reissue the active move after the new nodes render:

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7042 is fixed.
useEffect(() => {
  const { activeId, open } = menu.getState();
  if (!open) return;
  if (activeId == null) return;
  menu.move(activeId);
}, [menu, itemsGeneration]);
```

Tie `itemsGeneration` specifically to actual item-node replacement. `menu.move()` schedules a complete presentation and moves DOM focus as well as scrolling, so this workaround is narrower than the final library behavior.

## Verification

- `pnpm lint-fix`
- `pnpm build`
- `pnpm clean`
- `pnpm tsc`
- React 19: 254 files, 1700 tests
- React 18: 187 files, 966 tests
- Focused browser tests: Chrome 8/8, Firefox 8/8, WebKit 8/8
- Focused happy-dom tests: React 19 5/5, React 18 5/5
- `pnpm build-app-lite`

## Preview

[Exact-head browser recording: after the item blurs and regains focus, the same-id replacement receives focus when positioning finishes.](https://github.com/user-attachments/assets/56baba84-3bdb-4259-b47c-61b6ab344bf2)

<details>
<summary>Cycle 3: Narrow cleanup-time blur provenance</summary>

**Assessment**

The original issue is a confirmed stable-API `Menu` bug: with a pending public `updatePosition`, replacing the focused same-id item drops focus to the document body and leaves the active item offscreen. This has medium impact for keyboard and assistive-technology users, although the timing is uncommon.

Cycle 1 narrowed the unmount marker to the focused item. Cycle 2 found that the one-shot blur latch stayed terminal after refocus, so the implementation now keeps persistent focus and blur listeners for the active request. A separate direct-ref change attempted to distinguish a callback-ref cleanup that deliberately calls `blur()`, but Cycle 3 showed that this guarantee would not cover the equally public render-element ref path.

The cleanup-time blur scenario is rare, and complete support would require cross-cutting final-ref ordering machinery or a risky global merge-order change.

**Decision**

Keep the focused-only marker and the refocus listener pair. Treat `blur()` invoked from callback-ref teardown as unsupported focus-escape provenance regardless of whether the callback comes from a direct ref or a render-element ref. The partial direct-ref behavior was removed, keeping both public ref surfaces coherent without a cross-cutting abstraction.

</details>

### DIFF
--- a/app/src/sandbox/menu-placing-pass/test-browser.ts
+++ b/app/src/sandbox/menu-placing-pass/test-browser.ts
@@ -23,9 +23,6 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
     await q.button("Replace focused action").dispatchEvent("click");
     const replacement = withinMenu.menuitem("Action 30");
     await test.expect(replacement).toHaveAttribute("data-active-item");
-    await test.expect
-      .poll(() => page.evaluate(() => document.activeElement?.tagName))
-      .toBe("BODY");
 
     await q
       .button("Finish replacement actions positioning")
--- a/app/src/sandbox/menu-placing-pass/test-browser.ts
+++ b/app/src/sandbox/menu-placing-pass/test-browser.ts
@@ -1,7 +1,41 @@
 import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
 import { recordScrollEvents } from "#app/test-utils/scroll.ts";
 
-withFramework(import.meta.dirname, async ({ test }) => {
+withFramework(import.meta.dirname, async ({ query, test }) => {
+  // https://github.com/ariakit/ariakit/issues/7042
+  test("presents a focused item after its node is replaced", async ({
+    page,
+    q,
+  }) => {
+    const menu = q.menu("Replacement actions");
+    const withinMenu = query(menu);
+
+    await q.button("Show replacement actions").click();
+    await test.expect(menu).toBeVisible();
+    await test.expect(menu).toHaveAttribute("data-placing");
+
+    await withinMenu.menuitem("Action 1").focus();
+    await page.keyboard.press("End");
+    await test.expect(withinMenu.menuitem("Action 30")).toBeFocused();
+
+    // Dispatching without a native click preserves the removed item as the
+    // browser's focus provenance while React commits its replacement.
+    await q.button("Replace focused action").dispatchEvent("click");
+    const replacement = withinMenu.menuitem("Action 30");
+    await test.expect(replacement).toHaveAttribute("data-active-item");
+    await test.expect
+      .poll(() => page.evaluate(() => document.activeElement?.tagName))
+      .toBe("BODY");
+
+    await q
+      .button("Finish replacement actions positioning")
+      .dispatchEvent("click");
+    await test.expect(replacement).toBeFocused();
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBeGreaterThan(0);
+  });
+
   // A custom `updatePosition` that calls the supplied default and then keeps
   // working owns the whole pass: the popup is only where it belongs once that
   // callback returns. Publishing readiness when the inner default pass resolves
--- a/app/src/sandbox/menu-placing-pass/test-browser.ts
+++ b/app/src/sandbox/menu-placing-pass/test-browser.ts
@@ -23,6 +23,7 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
     await q.button("Replace focused action").dispatchEvent("click");
     const replacement = withinMenu.menuitem("Action 30");
     await test.expect(replacement).toHaveAttribute("data-active-item");
+    await test.expect(replacement).toBeFocused();
 
     await q
       .button("Finish replacement actions positioning")
@@ -31,6 +32,34 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
     await test.expect
       .poll(() => menu.evaluate((element) => element.scrollTop))
       .toBeGreaterThan(0);
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7042
+  test("abandons a replaced item after focus leaves", async ({ page, q }) => {
+    const menu = q.menu("Replacement actions");
+    const withinMenu = query(menu);
+
+    await q.button("Show replacement actions").click();
+    await withinMenu.menuitem("Action 1").focus();
+    await page.keyboard.press("End");
+    const item = withinMenu.menuitem("Action 30");
+    await test.expect(item).toBeFocused();
+
+    await item.blur();
+    await test.expect
+      .poll(() => page.evaluate(() => document.activeElement?.tagName))
+      .toBe("BODY");
+    await q.button("Replace focused action").dispatchEvent("click");
+    const replacement = withinMenu.menuitem("Action 30");
+    await test.expect(replacement).toHaveAttribute("data-active-item");
+
+    await q
+      .button("Finish replacement actions positioning")
+      .dispatchEvent("click");
+    await test.expect(replacement).not.toBeFocused();
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBe(0);
   });
 
   // A custom `updatePosition` that calls the supplied default and then keeps
--- a/app/src/sandbox/menu-placing-pass/test-browser.ts
+++ b/app/src/sandbox/menu-placing-pass/test-browser.ts
@@ -74,8 +74,14 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
     await withinMenu.menuitem("Action 1").focus();
     await page.keyboard.press("End");
     const item = withinMenu.menuitem("Action 30");
+    await test.expect(item).toBeFocused();
+    await test.expect(menu).toHaveAttribute("data-placing");
     await item.blur();
+    await test.expect
+      .poll(() => page.evaluate(() => document.activeElement?.tagName))
+      .toBe("BODY");
     await item.focus();
+    await test.expect(item).toBeFocused();
 
     await q.button("Replace focused action").dispatchEvent("click");
     const replacement = withinMenu.menuitem("Action 30");
@@ -84,6 +90,9 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
       .button("Finish replacement actions positioning")
       .dispatchEvent("click");
     await test.expect(replacement).toBeFocused();
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBeGreaterThan(0);
   });
 
   // A custom `updatePosition` that calls the supplied default and then keeps
--- a/app/src/sandbox/menu-placing-pass/test-browser.ts
+++ b/app/src/sandbox/menu-placing-pass/test-browser.ts
@@ -62,6 +62,30 @@ withFramework(import.meta.dirname, async ({ query, test }) => {
       .toBe(0);
   });
 
+  // https://github.com/ariakit/ariakit/pull/7065#discussion_r3711847856
+  test("restores focus when a refocused item is replaced", async ({
+    page,
+    q,
+  }) => {
+    const menu = q.menu("Replacement actions");
+    const withinMenu = query(menu);
+
+    await q.button("Show replacement actions").click();
+    await withinMenu.menuitem("Action 1").focus();
+    await page.keyboard.press("End");
+    const item = withinMenu.menuitem("Action 30");
+    await item.blur();
+    await item.focus();
+
+    await q.button("Replace focused action").dispatchEvent("click");
+    const replacement = withinMenu.menuitem("Action 30");
+
+    await q
+      .button("Finish replacement actions positioning")
+      .dispatchEvent("click");
+    await test.expect(replacement).toBeFocused();
+  });
+
   // A custom `updatePosition` that calls the supplied default and then keeps
   // working owns the whole pass: the popup is only where it belongs once that
   // callback returns. Publishing readiness when the inner default pass resolves
--- a/packages/ariakit-react-components/src/composite/composite-item.tsx
+++ b/packages/ariakit-react-components/src/composite/composite-item.tsx
@@ -50,6 +50,8 @@ import {
   focusSilently,
   getEnabledItem,
   isItem,
+  markItemMounted,
+  markItemUnmounting,
   selectTextField,
   usePresentItem,
 } from "./utils.ts";
@@ -166,6 +168,13 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
 
     const id = useId(props.id);
     const ref = useRef<HTMLType>(null);
+    const mountedElementRef = useRef<HTMLType>(null);
+    const markUnmountingRef = useCallback((element: HTMLType | null) => {
+      const mountedElement = mountedElementRef.current;
+      if (!element && mountedElement) markItemUnmounting(mountedElement);
+      if (element) markItemMounted(element);
+      mountedElementRef.current = element;
+    }, []);
     const row = useContext(CompositeRowContext);
     const disabled = disabledFromProps(props);
     const trulyDisabled = disabled && !props.accessibleWhenDisabled;
@@ -527,7 +536,7 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
       "data-active-item": isActiveItem || undefined,
       ...props,
       id,
-      ref: useMergeRefs(ref, props.ref),
+      ref: useMergeRefs(ref, markUnmountingRef, props.ref),
       tabIndex: isTabbable ? props.tabIndex : -1,
       onFocus,
       onBlurCapture,
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -14,6 +14,19 @@ export const flipItems = Core.flipItems;
 export const findFirstEnabledItem = Core.findFirstEnabledItem;
 export const groupItemsByRows = Core.groupItemsByRows;
 
+const unmountingItems = new WeakSet<Element>();
+
+/** Marks the brief window between a composite item's ref cleanup and removal. */
+export function markItemUnmounting(element: Element) {
+  unmountingItems.add(element);
+  queueMicrotask(() => unmountingItems.delete(element));
+}
+
+/** Clears a transient ref cleanup when React keeps the same node mounted. */
+export function markItemMounted(element: Element) {
+  unmountingItems.delete(element);
+}
+
 /**
  * Runs a callback while preserving the composite element's own scroll position.
  *
@@ -107,6 +120,7 @@ function presentItem({
   let element: HTMLElement | null = null;
   let resolvedId: string | undefined;
   let focused = false;
+  let focusLeftElement = false;
   let done = false;
   let wasMounted = false;
   let wasOpen = false;
@@ -185,20 +199,27 @@ function presentItem({
     resolvedId = item.id;
     return item.element;
   };
+  let removeBlurListener: (() => void) | undefined;
   let unsubscribe: (() => void) | undefined;
   const cancel = () => {
     done = true;
+    removeBlurListener?.();
     unsubscribe?.();
   };
   const present = () => {
     if (done) return;
     const state = store.getState();
     if (abandonedByState(state)) return cancel();
+    let restoreFocus = false;
     if (!element) {
       // The item may not have rendered yet, so keep waiting for it.
       element = resolveElement(state);
       if (!element) return;
     } else if (!element.isConnected) {
+      restoreFocus =
+        focused &&
+        !focusLeftElement &&
+        getActiveElement(element) === getDocument(element).body;
       // The element left the DOM, but React can replace an item's node while
       // keeping its id, so the logical item can still be there and still be
       // worth presenting. Resolve it again here rather than going back to
@@ -221,27 +242,27 @@ function presentItem({
     // is nothing to compare focus against, and an option that has just focused
     // itself isn't even recognizable as an item yet.
     //
-    // A request whose item still holds DOM focus when that item is replaced
-    // ends here, because the browser parks focus on the body when it removes
-    // the focused node. It only gets that far with an owner to compare against,
-    // and only where the item keeps DOM focus: a virtual-focus item that hands
-    // focus straight back to the composite element moves no focus at all when
-    // its node is replaced. A `Menu` moved while focus is already inside it
-    // therefore keeps the behavior a replaced item had before this, while the
-    // same menu moved from outside has no owner and gets this far. Only as far
-    // as the scroll, though: the focus step below is latched to the element the
-    // request first resolved, so a replacement is brought into view without
-    // being focused in its place.
-    //
-    // Telling a removed focused node apart from focus escaping is deliberately
-    // not attempted here: the signal would have to be carried down from the
-    // re-resolution above, into an ordering that just gained a focus-escape
-    // guarantee.
-    // See https://github.com/ariakit/ariakit/issues/7042
-    if (!stillOwnsFocus(element)) return cancel();
+    // Removing the focused node parks focus on the document body. Restore
+    // focus only when the item's ref cleanup marked that blur as removal. An
+    // explicit blur is latched below, so replacing the item afterward does not
+    // turn a real focus escape into a new focus request.
+    if (restoreFocus) {
+      focused = false;
+    } else if (!stillOwnsFocus(element)) {
+      return cancel();
+    }
     if (focus && !focused) {
       focused = true;
+      focusLeftElement = false;
       const itemElement = element;
+      removeBlurListener?.();
+      const onBlur = () => {
+        focusLeftElement = !unmountingItems.has(itemElement);
+      };
+      itemElement.addEventListener("blur", onBlur, { once: true });
+      removeBlurListener = () => {
+        itemElement.removeEventListener("blur", onBlur);
+      };
       withCompositeScrollPreserved(store, () => {
         itemElement.focus({ preventScroll: true });
       });
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -200,11 +200,11 @@ function presentItem({
     resolvedId = item.id;
     return item.element;
   };
-  let removeBlurListener: (() => void) | undefined;
+  let removeFocusListeners: (() => void) | undefined;
   let unsubscribe: (() => void) | undefined;
   const cancel = () => {
     done = true;
-    removeBlurListener?.();
+    removeFocusListeners?.();
     unsubscribe?.();
   };
   const present = () => {
@@ -256,13 +256,18 @@ function presentItem({
       focused = true;
       focusLeftElement = false;
       const itemElement = element;
-      removeBlurListener?.();
+      removeFocusListeners?.();
       const onBlur = () => {
         focusLeftElement = !unmountingItems.has(itemElement);
       };
-      itemElement.addEventListener("blur", onBlur, { once: true });
-      removeBlurListener = () => {
+      const onFocus = () => {
+        focusLeftElement = false;
+      };
+      itemElement.addEventListener("blur", onBlur);
+      itemElement.addEventListener("focus", onFocus);
+      removeFocusListeners = () => {
         itemElement.removeEventListener("blur", onBlur);
+        itemElement.removeEventListener("focus", onFocus);
       };
       withCompositeScrollPreserved(store, () => {
         itemElement.focus({ preventScroll: true });
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -245,8 +245,9 @@ function presentItem({
     //
     // Removing the focused node parks focus on the document body. Restore
     // focus only when the item's ref cleanup marked that blur as removal. An
-    // explicit blur is latched below, so replacing the item afterward does not
-    // turn a real focus escape into a new focus request.
+    // explicit blur is latched below until the same item takes focus again, so
+    // replacing an item the user left does not turn a real focus escape into a
+    // new focus request, while an item they came back to is still restored.
     if (restoreFocus) {
       focused = false;
     } else if (!stillOwnsFocus(element)) {
--- a/packages/ariakit-react-components/src/composite/utils.ts
+++ b/packages/ariakit-react-components/src/composite/utils.ts
@@ -16,8 +16,9 @@ export const groupItemsByRows = Core.groupItemsByRows;
 
 const unmountingItems = new WeakSet<Element>();
 
-/** Marks the brief window between a composite item's ref cleanup and removal. */
+/** Marks the brief window between a focused item's ref cleanup and removal. */
 export function markItemUnmounting(element: Element) {
+  if (getActiveElement(element) !== element) return;
   unmountingItems.add(element);
   queueMicrotask(() => unmountingItems.delete(element));
 }
--- a/.changeset/300-menu-item-replacement-focus.md
+++ b/.changeset/300-menu-item-replacement-focus.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Menu`](https://ariakit.com/reference/menu) to preserve DOM focus and bring a logical item into view when React replaces its element while the popup is positioning.
--- a/app/src/sandbox/menu-placing-pass/index.react.tsx
+++ b/app/src/sandbox/menu-placing-pass/index.react.tsx
@@ -1,6 +1,6 @@
 import * as Ariakit from "@ariakit/react";
 import type { ComponentProps } from "react";
-import { Fragment, useRef, useState } from "react";
+import { Fragment, useEffect, useRef, useState } from "react";
 
 type MenuProps = ComponentProps<typeof Ariakit.Menu>;
 type UpdatePosition = NonNullable<MenuProps["updatePosition"]>;
@@ -27,6 +27,14 @@ function FocusedReplacementMenu() {
   const menu = Ariakit.useMenuStore();
   const releaseRef = useRef<(() => void) | null>(null);
   const [generation, setGeneration] = useState(0);
+
+  // TODO: Remove once https://github.com/ariakit/ariakit/issues/7042 is fixed.
+  useEffect(() => {
+    const { activeId, open } = menu.getState();
+    if (!open) return;
+    if (activeId == null) return;
+    menu.move(activeId);
+  }, [menu, generation]);
 
   const updatePosition: UpdatePosition = () =>
     new Promise<void>((resolve) => {
--- a/app/src/sandbox/menu-placing-pass/index.react.tsx
+++ b/app/src/sandbox/menu-placing-pass/index.react.tsx
@@ -1,6 +1,6 @@
 import * as Ariakit from "@ariakit/react";
 import type { ComponentProps } from "react";
-import { useRef } from "react";
+import { Fragment, useRef, useState } from "react";
 
 type MenuProps = ComponentProps<typeof Ariakit.Menu>;
 type UpdatePosition = NonNullable<MenuProps["updatePosition"]>;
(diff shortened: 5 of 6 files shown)


packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -14,6 +14,19 @@ export const flipItems = Core.flipItems;
 export const findFirstEnabledItem = Core.findFirstEnabledItem;
 export const groupItemsByRows = Core.groupItemsByRows;
 
+const unmountingItems = new WeakSet<Element>();
+
+/** Marks the brief window between a composite item's ref cleanup and removal. */
+export function markItemUnmounting(element: Element) {
+  unmountingItems.add(element);
+  queueMicrotask(() => unmountingItems.delete(element));
+}
```
diegohaz, line 24: Every composite item that unmounts gets a `WeakSet` insert plus its own `queueMicrotask` closure here, but the mark is only ever read from the one-shot `blur` listener in `presentItem`, and only an element that holds DOM focus can fire `blur`. For every item except the focused one, the entry is therefore written and cleaned up without ever being observable.

The invariant the mark actually serves is narrower than what it records: "the item that is going away is the one holding focus". Testing that at the call site keeps the marked set to at most one element per commit and makes the intent explicit at the point where it is established, rather than leaving it implicit in the blur handler.

```suggestion
export function markItemUnmounting(element: Element) {
  // Only an item that holds focus can blur when it goes, and the mark is only
  // ever read from that blur, so nothing else needs one.
  if (getActiveElement(element) !== element) return;
  unmountingItems.add(element);
  queueMicrotask(() => unmountingItems.delete(element));
}
```

`getActiveElement` is already imported in this file, and `markItemMounted` is still required for the ref-reattach case, where a focused item's merged ref changes identity and React detaches and reattaches it on the same node.

I validated this locally on top of the PR head. `app/src/sandbox/menu-placing-pass` stays green in happy-dom (4/4), Chrome (7/7), Firefox (7/7) and WebKit (7/7), and the full `pnpm test` run stays at 254 files / 1698 tests. I also confirmed the guard does not weaken the regression proof: replacing the `unmountingItems` check with `focusLeftElement = false` still turns `abandons a replaced item after focus leaves` red in both happy-dom and Chrome.

On magnitude, I want to be honest rather than oversell it: a microbenchmark of 1000 marks put the per-item path at roughly 0.09ms versus 0.02ms for a batched variant, so the saving on a large virtualized unmount is a fraction of a millisecond against a React commit that costs tens of milliseconds. The reason to do it is that the recorded state matches the invariant, not the timing.

**Assessment**

- Severity: Low (verified, evidence: behavior is identical across happy-dom, Chrome, Firefox, WebKit and the full suite), because nothing user-visible changes either way.
- Priority: P2 (estimate, 80% confidence), because it tightens newly introduced state rather than fixing a defect.
- Breakage likelihood: Not applicable, because this is not a latent bug; the current code is correct, just broader than it needs to be. The wasted work is Likely (verified) in the narrow sense that it happens on every composite item unmount, including bulk unmounts in `CollectionRenderer`-backed comboboxes.
- Fix effort: Easy (verified, evidence: the one-line guard above passing all four environments plus the full suite).

diegohaz (commented):
Initial review. No prior review threads existed on this PR, so the baseline is the merge base `0beb71cf9` through head `df4659bed`, plus the surrounding composite focus and presentation code, the `menu-placing-pass` fixture contract, and the changeset conventions.

## Verdict

The approach holds up. I tried to replace it with something simpler twice and both attempts failed or were not worth it, which is the strongest thing I can say for it. One inline comment, on the breadth of the new marker rather than on its correctness.

## What I ran

- `pnpm test app/src/sandbox/menu-placing-pass/test.ts`: 4/4.
- `pnpm test`: 254 files, 1698 tests, all passing.
- `pnpm test-react18 app/src/sandbox/menu-placing-pass/test.ts`: 4/4.
- `pnpm test-chrome`, `pnpm test-firefox`, `pnpm test-safari` on `menu-placing-pass`: 7/7 each, against an `APP_PORT=4323` server started with `pnpm dev-app` from this review worktree and verified to serve this branch's fixture.

## Regression proof, verified both ways

- Reverting `composite/utils.ts` and `composite/composite-item.tsx` to `0beb71cf9` turns `restores focus after a focused item is replaced` red at `expect(replacement).toHaveFocus()`, and leaves the other three green. The red is for the stated reason.
- Neutralizing the explicit-blur guard (`focusLeftElement = false` in the blur handler) turns `leaves focus outside after a focused item is replaced` red in happy-dom and `abandons a replaced item after focus leaves` red in Chrome. The claim in the PR description is accurate.

## Alternatives I tested and rejected

**Dropping the marker for `itemElement.isConnected` at blur time.** This would have removed `unmountingItems`, both exported helpers, and the whole `composite-item.tsx` change. It passes happy-dom but fails Chrome. I probed the underlying DOM behavior directly across engines to understand why:

| Engine | `blur` on removal of the focused node | `isConnected` inside that handler | Order vs. a microtask queued at removal |
| --- | --- | --- | --- |
| Chromium | fires synchronously | `true` | `blur` first |
| Firefox | never fires | n/a | n/a |
| WebKit | never fires | n/a | n/a |

So Chromium runs the focus fixup before detaching the node, and `isConnected` cannot distinguish a removal blur from an explicit one there. The ref-cleanup marker is doing real work, and it is Chromium where it is load-bearing: on Firefox and WebKit no blur fires on removal, so `focusLeftElement` stays false without any marker. The probe also confirms `queueMicrotask` is a safe cleanup window, since the removal blur lands before it.

**Batching the per-item microtask into one per commit.** Measured at roughly 0.09ms versus 0.02ms per 1000 marks, which is negligible against the React commit that produces them, and the perf CI on this PR reports no significant change. Declined non-goal: not worth the indirection.

The finding I did publish is the third variant, guarding the mark on the element actually holding focus. That one keeps every environment green and narrows the recorded state to the invariant the blur handler relies on.

## Correctness paths I walked and did not find problems in

- React detaches refs for a deleted subtree before `removeChild` of its topmost host node in both React 18 and 19, so the mark is always set before the removal that triggers the blur. `useMergeRefs` covers both React 19 ref shapes: when no child ref returns a cleanup the merged ref returns `undefined` and React calls it with `null`, and when one does, the merged cleanup still calls `setRef(ref, null)` for the others.
- Every composite item component routes through `useCompositeItem`, so no item type is left unmarked.
- Only two call sites pass `focus: true`, so the new blur listener has a narrow blast radius. The virtual-focus site hands focus straight back to the composite element, which latches `focusLeftElement`; that is harmless, because `restoreFocus` also requires the active element to be the body, and DOM focus sits on the composite element in that flow. The #7030 virtual-focus behavior is unaffected.
- Focus stealing: an ordinary escape (clicking the page background, tabbing away, moving to another control) blurs the item while it is still connected and unmarked, so `focusLeftElement` latches and the request still cancels.

One residual I want to name rather than publish as a finding: `markItemUnmounting` cannot tell a deletion detach from an identity detach, so an item whose `props.ref` changes identity on every render is marked during the window between React's mutation-phase detach and its layout-phase reattach. A focus move landing inside that window would be misread as a removal, and a later replacement could then pull focus back. This needs an unstable ref, a presentation in flight, a focus move inside a window spanning only sibling layout effects, the active element ending on the body, and a node replacement. I could not construct it, and React exposes no signal that would separate the two detach kinds, so the fix would mean abandoning the ref as the source. Declined non-goal at this rarity; the guard in the inline comment does not change it either way, since only a focused element can blur in the first place.

## Minor notes, not published inline

- `composite-item.tsx:174-175` uses single-line `if` bodies that are not early exits, which `ariakit-general-code-style` reserves braces for. There are only three other instances in `packages/ariakit-react-components`, so this is new code moving against the local grain. Low value, hence not inline.
- The changeset names `Menu` only. That is accurate for default configurations, since `Menu` inherits `virtualFocus: false` while `Select` and `Combobox` default to `true`, but a consumer setting `virtualFocus={false}` on `Select` reaches the same fix. Naming one representative derived component is within the single-paragraph guidance, so this is an observation rather than a request. Contributor credit is correctly absent: issue #7042 was opened by `ariakit-bot` and its only commenters are bots.
- `dispatchEvent("click")` in the browser test is a synthetic driver rather than a user click, but it is necessary and documented: a real click would move focus off the item and defeat both scenarios. The fixture extension also does not disturb the existing scenarios, and the five pre-existing browser tests stay green.

## Not fully validated

- Android and iOS projects, and the visual suite, which are CI-only here. All are green on CI for this head.
- The perf numbers, which I read from the CI report rather than reproducing. The report flags no significant change, and the movements go in both directions.

packages/ariakit-react-components/src/composite/utils.ts
diegohaz: Fixed in [`754479456`](https://github.com/ariakit/ariakit/commit/7544794561b1bfaa7d723fe97e724d080947eec2): unmount tracking now records only the item that owns DOM focus, while ref reattachment still clears the transient mark.

diegohaz:
**Fixed**

- [`754479456`](https://github.com/ariakit/ariakit/commit/7544794561b1bfaa7d723fe97e724d080947eec2) ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3711808378)): Limited transient unmount tracking to the item that owns DOM focus while preserving same-node ref reattachment cleanup.

packages/ariakit-react-components/src/composite/utils.ts
diegohaz: Confirmed resolved at `754479456`. I re-ran the guard against this head rather than carrying over the earlier result: `menu-placing-pass` is green in happy-dom (4/4) and in Chrome, Firefox and WebKit (7/7 each), the full `pnpm test` run is 254 files / 1698 tests, and `pnpm tsc` and `pnpm lint` are clean.

The narrowing also holds where it matters: `markItemMounted` is still doing its job for the ref-reattach case, since a focused item whose merged ref changes identity is marked on detach and cleared on reattach.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -221,27 +243,27 @@ function presentItem({
...
+    }
     if (focus && !focused) {
       focused = true;
+      focusLeftElement = false;
       const itemElement = element;
+      removeBlurListener?.();
+      const onBlur = () => {
+        focusLeftElement = !unmountingItems.has(itemElement);
+      };
+      itemElement.addEventListener("blur", onBlur, { once: true });
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and it reproduces outside happy-dom. I added a temporary case to the existing fixture that focuses `Action 30` through the move, blurs it, focuses it again, and only then replaces the nodes while positioning is still held:

```ts
await blur(item);
expect(document.activeElement).toBe(document.body);
await focus(item);
expect(item).toHaveFocus();

await dispatch.click(q.button("Replace focused action"));
// Fails on this head: the replacement is not focused.
expect(replacement).toHaveFocus();
```

On `754479456` that fails in happy-dom, Chrome and Firefox. The item held DOM focus at the moment its node was replaced, which is exactly the situation this PR sets out to fix, yet focus lands on the body and the replacement is never scrolled into view. The outcome depends on a blur that happened earlier and that the user cannot perceive, so two identical-looking interactions diverge.

What makes it terminal is that `focusLeftElement` records "a blur happened at some point" rather than "focus is not on this element now", and `{ once: true }` removes the only listener that could ever revise it. Tracking the return leg fixes it directly:

```suggestion
      const onBlur = () => {
        focusLeftElement = !unmountingItems.has(itemElement);
      };
      const onFocus = () => {
        focusLeftElement = false;
      };
      itemElement.addEventListener("blur", onBlur);
      itemElement.addEventListener("focus", onFocus);
      removeBlurListener = () => {
        itemElement.removeEventListener("blur", onBlur);
        itemElement.removeEventListener("focus", onFocus);
      };
```

With that applied, the repro passes in happy-dom and in Chrome, Firefox and WebKit (8/8 each), the four existing sandbox tests stay green, the full `pnpm test` run stays green, and `pnpm tsc` is clean. `leaves focus outside after a focused item is replaced` still passes, because focus never returns to the item there, so the latch stays set and the request still cancels.

Worth adding a regression test for the return leg alongside the two existing cases, since nothing currently pins it.

**Assessment**

- Severity: Medium (verified, evidence: the replacement loses both DOM focus and the pending scroll in three engines), because the user's keyboard position is dropped in the exact scenario the PR targets.
- Priority: P1 (estimate, 85% confidence), because it is an incomplete case of this PR's own contract rather than separable work, and the fix is contained.
- Breakage likelihood: Possible (estimate, 70% confidence). It needs a blur and a return to the same item inside the placing window, which is short in ordinary use but is exactly what a menu held open by a slow or custom `updatePosition` widens, and clicking the popup background and then back onto an item is an ordinary way to produce it.
- Fix effort: Easy (verified, evidence: the listener pair above passing every environment plus the full suite).

packages/ariakit-react-components/src/composite/composite-item.tsx
```diff
@@ -527,7 +536,7 @@ export const useCompositeItem = createHook<TagName, CompositeItemOptions>(
       "data-active-item": isActiveItem || undefined,
```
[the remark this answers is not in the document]
diegohaz: The mechanism is real, and I reproduced it. I gave the fixture's items a consumer callback ref whose cleanup blurs the node:

```tsx
ref={(element) => {
  if (!element) return;
  return () => element.blur();
}}
```

On `754479456`, `restores focus after a focused item is replaced` still passes with that ref in place, so the consumer's `blur()` is indeed absorbed: `useMergeRefs` walks its refs in order on both the React 19 cleanup path and the `null` path, so `markUnmountingRef` marks the node before `props.ref` gets to run. Reordering it last makes that test fail instead, which is the behavior you describe, and the reorder does not disturb the normal path: with the fixture back to its own shape, the four existing sandbox tests stay green.

```suggestion
      ref: useMergeRefs(ref, props.ref, markUnmountingRef),
```

That also composes with the focused-item guard now in `markItemUnmounting`: once the consumer's cleanup has blurred the node, `getActiveElement(element) !== element`, so the mark is skipped and the blur stays terminal without any extra branch.

Where I would temper the framing is the severity. In a plain removal the node is being detached either way and focus would reach the body regardless, so treating a `blur()` there as an intentional escape is a judgment call rather than a clear defect. The case that convinces me is the other one: when the detach is a ref identity change and the element stays mounted, the consumer's blur is a genuine escape on a live node, and the current order records it as a removal. No Ariakit code path calls `.blur()` from a ref cleanup, so reaching this needs consumer code shaped that way.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because focus is pulled back into the popup against an explicit consumer blur.
- Priority: P2 (estimate, 75% confidence), because no in-repo path reaches it, though the fix is a single-line reorder that costs nothing to take now.
- Breakage likelihood: Rare (verified for the mechanism, estimate for the exposure, 80% confidence), because it needs a callback ref on a composite item whose cleanup blurs, coinciding with a presentation in flight and a same-id replacement.
- Fix effort: Easy (verified, evidence: the reorder above leaving the existing sandbox tests green).

diegohaz (commented):
Re-review. The previous review by this workflow ran against `df4659bed`, so the incremental baseline is `df4659bed..754479456`, which is the single commit `Limit unmount marker to focused item`. Newly discovered findings were reviewed from that range plus surrounding behavior it exposes, and every relevant unresolved thread was independently validated against the current head rather than taken on its author's word.

## Status

The narrowing I asked for landed and holds up. Two unresolved bot threads turned out to describe real behavior, and I reproduced both rather than accepting or dismissing them on inspection. One of them is worth fixing before merge; the other is a cheap refinement whose framing I would soften.

## Threads validated this run

| Thread | Claim | Outcome | Disposition |
| --- | --- | --- | --- |
| `utils.ts:24` (mine, `df4659bed`) | Marker recorded for every unmounting item, not just the focused one | Already resolved in `754479456` | Address now, done |
| `utils.ts:263` (Codex) | One-shot blur latch is never cleared when the item regains focus | Confirmed, reproduced in happy-dom, Chrome and Firefox | Address now |
| `composite-item.tsx:536` (Codex) | Marker is set before the consumer ref cleanup runs | Confirmed as a mechanism; severity framing softened | Address now, cheap, but see the caveat on the thread |

Full detail is on each thread. Summarised:

**The blur latch is the one I would not merge without.** `focusLeftElement` records that a blur happened rather than that focus is currently elsewhere, and `{ once: true }` removes the only listener that could revise it. So an item that is blurred, focused again, and then replaced while positioning is still pending loses both DOM focus and its pending scroll, even though it held focus at the moment of replacement. That is the same user-visible situation the PR exists to fix, and the divergence hinges on an earlier blur the user cannot perceive. I reproduced it on this head in happy-dom, Chrome and Firefox, and validated a listener-pair fix that keeps every environment green.

**The ref ordering claim is real but narrower than its wording suggests.** `useMergeRefs` walks its refs in order on both React 19 paths, so `markUnmountingRef` marks the node before `props.ref` gets to run, and a consumer ref cleanup that calls `blur()` is absorbed. I verified that by giving the fixture items such a ref. In a plain removal the node is going away regardless, so calling that an intentional escape is a judgment call; the case that persuades me is a ref identity change, where the element stays mounted and the consumer's blur is a genuine escape on a live node. No Ariakit path calls `.blur()` from a ref cleanup, so exposure needs consumer code shaped that way. The one-line reorder is free and composes with the new focused-item guard.

Both fixes are compatible. With the reorder and the listener pair applied together: `menu-placing-pass` is 8/8 in Chrome, Firefox and WebKit including my temporary repro, happy-dom is 5/5, the full `pnpm test` run is 254 files / 1699 tests, and `pnpm tsc` is clean. All experiments were reverted; the worktree is back at `754479456`.

## What I ran this round

- `pnpm test app/src/sandbox/menu-placing-pass/test.ts`, `pnpm test` (254 files / 1698 tests on the unmodified head), `pnpm tsc`, `pnpm lint`.
- `pnpm test-chrome`, `pnpm test-firefox`, `pnpm test-safari` on `menu-placing-pass`, against an Astro server started with `pnpm dev-app` from this review worktree and confirmed by process cwd and served fixture content to belong to this checkout.
- Temporary repros for both bot claims, plus prototypes of both fixes, all reverted afterwards.

## Carried forward from the previous review

Still accurate against this head, and not re-litigated here: the marker is load-bearing on Chromium specifically, because Chromium fires the removal blur synchronously with `isConnected` still `true` while Firefox and WebKit fire no blur at all on removal; `queueMicrotask` remains a safe cleanup window; and every composite item component routes through `useCompositeItem`, so no item type is left unmarked.

The minor notes from last round are unchanged and still not published inline: the two single-line non-early-exit `if` bodies at `composite-item.tsx:174-175`, and the changeset naming `Menu` only, which is accurate for default configurations since `Menu` inherits `virtualFocus: false` while `Select` and `Combobox` default to `true`.

One update to the residual I named last time. I described the deletion-versus-identity detach ambiguity as a race I could not construct. The Codex ref-ordering thread supplies a deterministic way into it through a consumer ref cleanup, so it is no longer only a race, and the reorder addresses that entry point. The general ambiguity remains, since React exposes no signal separating the two detach kinds.

## Not fully validated

- Android, iOS and the visual suite, which are CI-only here.
- Perf, read from the CI report rather than reproduced. Nothing in `754479456` changes the per-item cost characteristics beyond reducing them.

packages/ariakit-react-components/src/composite/composite-item.tsx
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because callback-ref teardown-time `blur()` is not a supported focus-escape provenance signal. Fully distinguishing it would require cross-cutting final-ref ordering across direct and render-element refs, while the reported sequence is a rare teardown edge. Direct and render refs therefore retain the same behavior. Browser evidence gap: this deliberately unsupported teardown provenance is not demonstrated; the [exact-head recording](https://github.com/user-attachments/assets/56baba84-3bdb-4259-b47c-61b6ab344bf2) covers the supported blur, refocus, and replacement flow instead.

packages/ariakit-react-components/src/composite/utils.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`6472ab9`](https://github.com/ariakit/ariakit/commit/6472ab95cd1a254c75c1d74f555f669d4afae5a0): pending presentations now refresh their focus-left state when the same item regains focus. Added targeted happy-dom and cross-browser regression coverage in the same commit. [Exact-head browser recording](https://github.com/user-attachments/assets/56baba84-3bdb-4259-b47c-61b6ab344bf2).

diegohaz:
**Fixed**

- [`6472ab9`](https://github.com/ariakit/ariakit/commit/6472ab95cd1a254c75c1d74f555f669d4afae5a0) ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3712500623)): Kept the pending item focus state current across blur and refocus cycles, with happy-dom and three-browser coverage. [Exact-head recording](https://github.com/user-attachments/assets/56baba84-3bdb-4259-b47c-61b6ab344bf2).

**Declined**

- `Non-goal` ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3712499966)): Kept callback-ref teardown blur outside the supported focus-escape contract instead of adding cross-cutting final-ref ordering.

**Already resolved**

- [`754479456`](https://github.com/ariakit/ariakit/commit/7544794561b1bfaa7d723fe97e724d080947eec2) ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3711808378)): Limited unmount tracking to the item that owns DOM focus.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -221,27 +243,32 @@ function presentItem({
...
-    // Telling a removed focused node apart from focus escaping is deliberately
-    // not attempted here: the signal would have to be carried down from the
-    // re-resolution above, into an ordering that just gained a focus-escape
-    // guarantee.
-    // See https://github.com/ariakit/ariakit/issues/7042
-    if (!stillOwnsFocus(element)) return cancel();
+    // Removing the focused node parks focus on the document body. Restore
+    // focus only when the item's ref cleanup marked that blur as removal. An
+    // explicit blur is latched below, so replacing the item afterward does not
+    // turn a real focus escape into a new focus request.
```
diegohaz: This paragraph still describes the `{ once: true }` behavior that `6472ab95c` replaced. "An explicit blur is latched below" now reads as unconditional, but the latch is released as soon as the same item takes focus again, which is exactly the case that commit added. A reader auditing the focus-escape rules from the comments would conclude that the refocus case is deliberately terminal, and the test added right below proves the opposite.

The gap is worth closing in this file specifically, because every other non-obvious decision here carries its rationale, and the new `onFocus` listener is currently the only part of the mechanism without one. Nothing in the code says why the latch has to be released rather than left set.

```suggestion
    // Removing the focused node parks focus on the document body. Restore
    // focus only when the item's ref cleanup marked that blur as removal. An
    // explicit blur is latched below until the same item takes focus again, so
    // replacing an item the user left does not turn a real focus escape into a
    // new focus request, while an item they came back to is still restored.
```

**Assessment**

- Severity: Low (verified, evidence: the comment contradicts the behavior the sibling test pins), because nothing shipped behaves differently.
- Priority: P2 (estimate, 80% confidence), because the current work introduced the mismatch and this file's comments are load-bearing for anyone reasoning about the focus-escape ordering.
- Breakage likelihood: Not applicable as a runtime defect. The risk is a future change made against the stale rule, which is Possible (estimate, 60% confidence) given how much of this function's contract lives in prose.
- Fix effort: Easy (verified), because it is comment-only.

app/src/sandbox/menu-placing-pass/test-browser.ts
```diff
@@ -1,7 +1,91 @@
...
+    await item.blur();
+    await item.focus();
+
+    await q.button("Replace focused action").dispatchEvent("click");
+    const replacement = withinMenu.menuitem("Action 30");
+
+    await q
+      .button("Finish replacement actions positioning")
+      .dispatchEvent("click");
+    await test.expect(replacement).toBeFocused();
```
diegohaz, line 92: This pins focus but not the pending scroll, and the scroll was the other half of what went wrong. The sibling test at the top of this file asserts both:

```ts
await test.expect(replacement).toBeFocused();
await test.expect
  .poll(() => menu.evaluate((element) => element.scrollTop))
  .toBeGreaterThan(0);
```

I checked that the refocus path really does restore the scroll on this head by adding that assertion temporarily; it passes in Chrome. So the coverage is available for free, and without it a future change that restores focus but drops the delayed `scrollIntoView` on this path would stay green here while the non-refocus test caught the same defect.

Two smaller things in the same test. It never asserts that `End` reached `Action 30` or that the item is focused before `blur()`, which its siblings both do, so a failure earlier in the sequence surfaces as a confusing final assertion rather than at the step that broke. It also omits `await test.expect(menu).toHaveAttribute("data-placing")`, which is what makes the test's premise explicit: the whole scenario only means something while positioning is still held.

```suggestion
    const item = withinMenu.menuitem("Action 30");
    await test.expect(item).toBeFocused();
    await test.expect(menu).toHaveAttribute("data-placing");
    await item.blur();
    await item.focus();
    await test.expect(item).toBeFocused();

    await q.button("Replace focused action").dispatchEvent("click");
    const replacement = withinMenu.menuitem("Action 30");

    await q
      .button("Finish replacement actions positioning")
      .dispatchEvent("click");
    await test.expect(replacement).toBeFocused();
    await test.expect
      .poll(() => menu.evaluate((element) => element.scrollTop))
      .toBeGreaterThan(0);
```

The happy-dom duplicate cannot carry the scroll assertion, but it can carry the intermediate focus assertions.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because the untested half is a real user-facing outcome: the item ends up focused but out of view.
- Priority: P2 (estimate, 75% confidence), because the behavior is correct today and this only protects it, but the test is new in this PR so the gap is cheapest to close now.
- Breakage likelihood: Possible (estimate, 65% confidence), because the focus and scroll halves are deliberately decoupled in `presentItem`, so a future change can move one without the other.
- Fix effort: Easy (verified, evidence: the scroll assertion passing in Chrome when added to this test).

diegohaz (commented):
Re-review. The previous run by this workflow was against `754479456`, so the incremental baseline is `754479456..6472ab95c`, the single commit `Track composite item refocus`. Findings below come from that range plus the surrounding code it changes, and the resolved threads were treated as settled rather than reopened.

## Status

The P1 from last round is fixed, and correctly. Two P2s this time, both about the change describing and protecting itself rather than about behavior: a comment that still documents the superseded rule, and a new regression test that pins half of what it was written for.

## The refocus fix

`6472ab95c` replaces the one-shot blur listener with a blur and focus pair, so `focusLeftElement` tracks whether focus is currently elsewhere rather than whether a blur ever happened. That is the right shape, and I verified the regression coverage rather than trusting the commit message: reverting `utils.ts` to `754479456` while keeping the new tests turns `restores focus when a refocused item is replaced` red in happy-dom and in Chrome, and it passes with the commit applied. The tests fail for the stated reason.

Dropping `{ once: true }` does not strand listeners. Every terminal path in `present` routes through `cancel`, the re-arm branch removes the previous pair before installing a new one, and `usePresentItem` cancels on unmount, so the only requests holding listeners are the parked ones that are meant to.

Validation at this head: `menu-placing-pass` 5/5 in happy-dom and 8/8 in Chrome, Firefox and WebKit; full `pnpm test` 254 files / 1699 tests; `pnpm tsc` 0 errors; `pnpm lint` clean. Servers were started with `pnpm dev-app` from this review worktree and confirmed by process cwd to belong to this checkout.

## Findings

Both published inline.

- `utils.ts:246-249` still says "An explicit blur is latched below", which is the behavior this commit removed. The latch is now released when the same item regains focus, so the comment states the opposite of what the new test pins. The new `onFocus` listener is also the only part of this mechanism without a stated rationale, in a function where the rest of the contract lives in prose.
- The new browser test asserts focus but not the delayed scroll, while its sibling asserts both. I confirmed by temporary assertion that the scroll is restored on this path in Chrome, so the coverage is free. It also skips the intermediate `toBeFocused` and `data-placing` assertions its siblings use to make the premise explicit.

## Resolved threads

Both bot threads I validated last round are now resolved, and I am not reopening either.

The refocus thread was fixed as described. On the ref-ordering thread I want to correct my own prior recommendation rather than let it stand: the decline is well founded and my suggested reorder was a partial fix. `system.tsx:55` merges the render element's ref after `props.ref` (`useMergeRefs(props.ref, getRefProperty(render))`), so reordering inside `useCompositeItem` would move the marker after a direct `ref` prop but still leave it ahead of a `render={<div ref={...} />}` ref. That would have made the two composition paths disagree, which is exactly the cross-cutting ordering problem the reply cites. Taking the one-line reorder on my recommendation would have been worse than leaving it.

My own marker thread from the first review remains open on GitHub but was answered and confirmed resolved last round; nothing in this commit touches it, so it needs no further reply.

## Not fully validated

- Android, iOS and the visual suite, which are CI-only here.
- Perf, unchanged in character by this commit: it swaps one listener for two per focused presentation and adds no per-item work.

packages/ariakit-react-components/src/composite/utils.ts
diegohaz: Fixed in [`50dcedd03`](https://github.com/ariakit/ariakit/commit/50dcedd03252e747704e24d025d50c75d68a366b): the presentation rationale now states that explicit blur remains terminal only until the same item regains focus.

app/src/sandbox/menu-placing-pass/test-browser.ts
diegohaz: Fixed in [`50dcedd03`](https://github.com/ariakit/ariakit/commit/50dcedd03252e747704e24d025d50c75d68a366b): the refocus regression now asserts the held placing state, each focus transition, replacement focus, and delayed scrolling. The matching happy-dom test covers the non-layout focus transitions; the browser case passes in Chrome, Firefox, and WebKit.

diegohaz:
**Fixed**

- [`50dcedd03`](https://github.com/ariakit/ariakit/commit/50dcedd03252e747704e24d025d50c75d68a366b) ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3712873296)): Updated the refocus rationale to match the focus listener.
- [`50dcedd03`](https://github.com/ariakit/ariakit/commit/50dcedd03252e747704e24d025d50c75d68a366b) ([reply](https://github.com/ariakit/ariakit/pull/7065#discussion_r3712874016)): Completed the refocus browser regression with placing, focus-transition, and delayed-scroll assertions.

diegohaz (commented):
Re-review. The previous run by this workflow was against `6472ab95c`, so the incremental baseline is `6472ab95c..50dcedd03`, the single commit `Complete composite refocus coverage`. Both findings from that run are addressed, and I re-read the accumulated diff from the merge base to make sure the six commits still add up to one coherent change.

## Status

No new code findings. One factual correction to the description, below. The implementation, its coverage, and its rationale now agree with each other, which was not true two rounds ago.

## What `50dcedd03` does

The comment rewrite and both test changes match what I asked for. The browser test now pins the placing state, each focus transition, and the delayed scroll; the happy-dom duplicate carries the intermediate assertions it can express. The test also picked up a body-focus poll after `blur()` that I had not suggested, which is the right addition: it makes the blur step's effect explicit rather than assumed.

Validation at this head: `menu-placing-pass` 5/5 in happy-dom and 8/8 in Chrome, Firefox and WebKit; full `pnpm test` 254 files / 1699 tests; `pnpm test-react18` 187 files / 965 tests; `pnpm tsc` 0 errors; `pnpm lint` clean. Servers came from `pnpm dev-app` in this review worktree, confirmed by process cwd.

## One correction to the description

The Verification section overstates both suite totals by exactly one:

| Claimed | Measured |
| --- | --- |
| React 19: 254 files, 1700 tests | 254 files, 1699 tests |
| React 18: 187 files, 966 tests | 187 files, 965 tests |

The file counts are right; only the test counts are off. This is not a local-environment artifact: the `Main / Test` and `Main / Test React 18` jobs on CI for `50dcedd03` report `1699 passed (1699)` and `965 passed (965)`, matching two local runs here. Worth correcting since the section is a record of what was actually run.

Everything else in the description now tracks the implementation, including the paired listeners, the three regression cases, and the reassessment note explaining why cleanup-time `blur()` provenance stays unsupported across both ref surfaces.

## Accumulated state, re-read

The final shape holds together: a focused-only unmount marker, a presentation-scoped blur and focus listener pair, and a `restoreFocus` branch gated on the request having focused the item, no live focus escape, and the active element being the body. `markItemMounted` still covers ref reattachment. Nothing in the later commits reintroduced the per-item cost or the terminal latch that earlier rounds removed. CI is green across every check on this head.

The changeset is unchanged since the first round and still accurate: `Menu` is the component reached by default, since it inherits `virtualFocus: false` while `Select` and `Combobox` default to `true`.

## Threads

Both findings from the previous run are resolved by this commit. The only thread still open on GitHub is my original marker finding, which was answered and confirmed resolved two rounds ago; nothing since has touched it, so it needs no further reply.

The one observation I have raised before and am not escalating: `composite-item.tsx:174-175` still uses single-line `if` bodies that are not early exits, against the local convention. It stays a note rather than a request.

## Not fully validated

- Android, iOS and the visual suite, which are CI-only here and green on this head.
- Perf, unchanged in character since `754479456`; this commit touches tests and one comment.

PR #7066, opened 2026-08-04
Prevent menu item hover from scrolling

### LINKED ISSUE #7012: Hovering a menu item can scroll the page
### Context

Found during review of https://github.com/ariakit/ariakit/pull/7009, which makes every focus move that presents a composite item use `preventScroll` and hands the scrolling to one explicit step. Several focus moves in the same widgets were deliberately left out of that change because they do not present the active item, and they still let the browser scroll.

### Problem

These call sites move DOM focus without `preventScroll`, so the browser scrolls the focused element into view, centered:

- `packages/ariakit-react-components/src/menu/menu-item.tsx:162` and `:170`, which focus a hovered submenu button
- `packages/ariakit-react-components/src/combobox/combobox-item.tsx:207` and `:211`
- `packages/ariakit-react-components/src/combobox/combobox-list.tsx:83`
- `packages/ariakit-react-components/src/composite/composite-container.tsx:76`, `:202`, `:214` and `:226`

The clearest user-facing shape is the first one: hovering a submenu button in a menubar on a scrolled page can move the page under the pointer. That is the behavior https://github.com/ariakit/ariakit/issues/2590 removed for composite hover, and #7009 applies the same rule to `CompositeHover`, so `MenuItem` is now inconsistent with the component it is built on.

### Why #7009 did not cover it

#7009 replaces the presentation policy for the active item. These moves have a different purpose, and changing them means deciding, per call site, whether the target should still be brought into view and by what alignment. That is a separate behavioral decision rather than a consequence of unifying the presentation, and none of them contributes to the page-scroll defect that PR fixes.

### Expected outcome

Each of these focus moves is either given `preventScroll` with an explicit presentation where one is wanted, or left as it is with the reason recorded. Hovering should not move the page in any of them, matching #2590.

### Workaround

https://github.com/ariakit/ariakit/issues/7012#issuecomment-5160221874

https://github.com/ariakit/ariakit/issues/7012#issuecomment-5181206564

diegohaz:
## Motivation

Hovering a partially visible submenu button or an item in an active menubar moves DOM focus without suppressing the browser's implicit focus scroll. The page or scrollable menu can therefore move under a stationary pointer.

Fixes #7012.

## Solution

Pass `preventScroll: true` to the two `MenuItem` focus calls that run when hover moves DOM focus to a submenu button or active menubar item. This keeps the page and scrollable menus stationary under the pointer while preserving focus, active-item state, menu switching, and Ariakit's explicit keyboard presentation.

The seven focus moves triggered by keyboard or click interactions remain unchanged because presenting those destinations is expected behavior.

Real-browser regressions cover both confirmed paths. The menubar case runs in Chromium and WebKit, while the scrollable submenu case runs in Chromium, matching the engines where each defect was reproduced.

## Workaround

Until a release containing this fix is available, wrap the hovered item's own `focus` method so every direct focus call on that element sets `preventScroll`. Ariakit's keyboard presentation still uses its explicit `scrollIntoView` path, so keyboard navigation continues to bring off-screen items into view.

```tsx
// TODO: Remove after https://github.com/ariakit/ariakit/issues/7012 is fixed.
function preventScrollOnFocus(element: HTMLElement | null) {
  if (!element) return;
  const focus = element.focus.bind(element);
  element.focus = (options) => focus({ ...options, preventScroll: true });
}

<Ariakit.MenuItem
  ref={preventScrollOnFocus}
  render={<Ariakit.MenuButton />}
>
  Edit
</Ariakit.MenuItem>;
```

Apply the ref to the submenu and menubar items rendered as `MenuButton` elements that can receive hover focus. This is a temporary element-level override, so it also suppresses implicit scrolling for other direct `focus()` calls on those elements; Ariakit's explicit item presentation remains intact.

## Preview

A focused 800×450 recording was captured from the failing revision and reviewed locally, showing the page move after hover. The available GitHub API cannot upload user attachments and the browser-upload control is unavailable in this execution environment, so the recording could not be attached; the automated tests below provide the exact before-and-after scroll measurements.

## Testing

Without the workaround or library fix, `APP_PORT=4321 pnpm test-browser app/src/sandbox/menu-item-hover-7012` fails on current `main` as expected:

- Menubar hover changes `scrollY` from 50 to 36 in Chromium and WebKit.
- Submenu hover changes `scrollTop` from 128 to 156 in Chromium.

With the sandbox workaround applied, the same command passes all three cases. With the workaround removed and the final library fix applied, it also passes all three cases.

Additional final verification:

- `pnpm build`
- `pnpm tsc`
- `pnpm lint-fix`

### DIFF
--- a/app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
+++ b/app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
@@ -1,0 +1,26 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/7012
+  test("hovering a partially visible menubar item does not scroll", async ({
+    page,
+    q,
+  }) => {
+    await q.menuitem("File").click();
+    await test.expect(q.menu("File")).toBeVisible();
+
+    await page.mouse.wheel(0, 50);
+    await test.expect.poll(() => page.evaluate(() => scrollY)).toBe(50);
+
+    const edit = q.menuitem("Edit");
+    const box = await edit.boundingBox();
+    if (!box) {
+      throw new Error("Edit menu item is not visible");
+    }
+    await page.mouse.move(box.x + box.width / 2, 8);
+
+    await test.expect(edit).toBeFocused();
+    await test.expect(q.menu("Edit")).toBeVisible();
+    await test.expect.poll(() => page.evaluate(() => scrollY)).toBe(50);
+  });
+});
--- a/app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
+++ b/app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
@@ -13,11 +13,12 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect.poll(() => page.evaluate(() => scrollY)).toBe(50);
 
     const edit = q.menuitem("Edit");
+    await test.expect(edit).not.toBeInViewport({ ratio: 1 });
     const box = await edit.boundingBox();
     if (!box) {
       throw new Error("Edit menu item is not visible");
     }
-    await page.mouse.move(box.x + box.width / 2, 8);
+    await page.mouse.move(box.x + box.width / 2, box.y + box.height - 8);
 
     await test.expect(edit).toBeFocused();
     await test.expect(q.menu("Edit")).toBeVisible();
--- a/app/src/sandbox/menu-item-hover-7012/test-chrome.ts
+++ b/app/src/sandbox/menu-item-hover-7012/test-chrome.ts
@@ -29,10 +29,10 @@ withFramework(import.meta.dirname, async ({ test }) => {
     if (!shareBox) {
       throw new Error("Share menu item is not visible");
     }
-    await page.mouse.move(
-      shareBox.x + shareBox.width / 2,
-      menuBox.y + menuBox.height - 8,
-    );
+    const menuBottom = menuBox.y + menuBox.height;
+    test.expect(shareBox.y).toBeLessThan(menuBottom);
+    test.expect(shareBox.y + shareBox.height).toBeGreaterThan(menuBottom);
+    await page.mouse.move(shareBox.x + shareBox.width / 2, menuBottom - 8);
 
     await test.expect(share).toBeFocused();
     await test.expect
--- a/app/src/sandbox/menu-item-hover-7012/test-chrome.ts
+++ b/app/src/sandbox/menu-item-hover-7012/test-chrome.ts
@@ -1,0 +1,42 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/7012
+  test("hovering a partially visible submenu button does not scroll", async ({
+    page,
+    q,
+  }) => {
+    const actions = q.button("Actions");
+    await actions.scrollIntoViewIfNeeded();
+    await actions.click();
+
+    const menu = q.menu("Actions");
+    const menuBox = await menu.boundingBox();
+    if (!menuBox) {
+      throw new Error("Actions menu is not visible");
+    }
+    await page.mouse.move(
+      menuBox.x + menuBox.width / 2,
+      menuBox.y + menuBox.height / 2,
+    );
+    await page.mouse.wheel(0, 128);
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBe(128);
+
+    const share = q.menuitem("Share");
+    const shareBox = await share.boundingBox();
+    if (!shareBox) {
+      throw new Error("Share menu item is not visible");
+    }
+    await page.mouse.move(
+      shareBox.x + shareBox.width / 2,
+      menuBox.y + menuBox.height - 8,
+    );
+
+    await test.expect(share).toBeFocused();
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBe(128);
+  });
+});
--- a/packages/ariakit-react-components/src/menu/menu-item.tsx
+++ b/packages/ariakit-react-components/src/menu/menu-item.tsx
@@ -154,6 +154,9 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
         if (!store) return false;
         if (!getFocusOnHover()) return false;
         const { compositeElement, items } = store.getState();
+        // Hovering is not a request to move the page or menu; the pointer is
+        // already on what the user cares about. Keyboard navigation presents
+        // items through the composite's own scroll step.
         // If the menu item is also a submenu button, we should move actual DOM
         // focus to it so that the submenu will not close when the user moves
         // the cursor back to the menu button.
--- a/packages/ariakit-react-components/src/menu/menu-item.tsx
+++ b/packages/ariakit-react-components/src/menu/menu-item.tsx
@@ -159,7 +159,7 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
         // the cursor back to the menu button.
         if (isWithinMenu) {
           if (event.currentTarget.hasAttribute("aria-expanded")) {
-            event.currentTarget.focus();
+            event.currentTarget.focus({ preventScroll: true });
           }
           return true;
         }
@@ -167,7 +167,7 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
         // the menu item if focus is somewhere on the widget. Without this, the
         // open menus in the menu bar wouldn't close.
         if (menuHasFocus(compositeElement, items, event.currentTarget)) {
-          event.currentTarget.focus();
+          event.currentTarget.focus({ preventScroll: true });
           return true;
         }
         return false;
--- a/.changeset/300-menu-item-hover-scroll.md
+++ b/.changeset/300-menu-item-hover-scroll.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`MenuItem`](https://ariakit.com/reference/menu-item) hover focus so partially visible menubar and submenu items no longer scroll the page or menu under the pointer.
--- a/app/src/sandbox/menu-item-hover-7012/index.react.tsx
+++ b/app/src/sandbox/menu-item-hover-7012/index.react.tsx
@@ -6,6 +6,12 @@ const itemStyle = {
   padding: "12px 16px",
 } as const;
 
+function preventScrollOnFocus(element: HTMLElement | null) {
+  if (!element) return;
+  const focus = element.focus.bind(element);
+  element.focus = (options) => focus({ ...options, preventScroll: true });
+}
+
 function MenuContent({ label }: { label: string }) {
   return (
     <Ariakit.Menu gutter={4} portal>
@@ -24,6 +30,7 @@ function MenubarExample() {
           {["File", "Edit", "View"].map((label) => (
             <Ariakit.MenuProvider key={label}>
               <Ariakit.MenuItem
+                ref={preventScrollOnFocus}
                 render={<Ariakit.MenuButton />}
                 style={itemStyle}
               >
@@ -51,7 +58,11 @@ function ScrollableMenuExample() {
             </Ariakit.MenuItem>
           ))}
           <Ariakit.MenuProvider>
-            <Ariakit.MenuItem render={<Ariakit.MenuButton />} style={itemStyle}>
+            <Ariakit.MenuItem
+              ref={preventScrollOnFocus}
+              render={<Ariakit.MenuButton />}
+              style={itemStyle}
+            >
               Share
             </Ariakit.MenuItem>
             <MenuContent label="Share" />
--- a/app/src/sandbox/menu-item-hover-7012/index.react.tsx
+++ b/app/src/sandbox/menu-item-hover-7012/index.react.tsx
@@ -1,0 +1,72 @@
+import * as Ariakit from "@ariakit/react";
+
+const itemStyle = {
+  border: "1px solid #ddd",
+  height: 56,
+  padding: "12px 16px",
+} as const;
+
+function MenuContent({ label }: { label: string }) {
+  return (
+    <Ariakit.Menu gutter={4} portal>
+      <Ariakit.MenuItem style={itemStyle}>{label} one</Ariakit.MenuItem>
+      <Ariakit.MenuItem style={itemStyle}>{label} two</Ariakit.MenuItem>
+    </Ariakit.Menu>
+  );
+}
+
+function MenubarExample() {
+  return (
+    <section>
+      <h2>Menubar</h2>
+      <Ariakit.MenubarProvider>
+        <Ariakit.Menubar style={{ display: "flex", gap: 8 }}>
+          {["File", "Edit", "View"].map((label) => (
+            <Ariakit.MenuProvider key={label}>
+              <Ariakit.MenuItem
+                render={<Ariakit.MenuButton />}
+                style={itemStyle}
+              >
+                {label}
+              </Ariakit.MenuItem>
+              <MenuContent label={label} />
+            </Ariakit.MenuProvider>
+          ))}
+        </Ariakit.Menubar>
+      </Ariakit.MenubarProvider>
+    </section>
+  );
+}
+
+function ScrollableMenuExample() {
+  return (
+    <section style={{ marginTop: 1200 }}>
+      <h2>Scrollable menu</h2>
+      <Ariakit.MenuProvider>
+        <Ariakit.MenuButton>Actions</Ariakit.MenuButton>
+        <Ariakit.Menu gutter={4} style={{ maxHeight: 180, overflow: "auto" }}>
+          {["Cut", "Copy", "Paste", "Delete", "Duplicate"].map((label) => (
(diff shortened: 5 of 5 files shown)


app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    await page.mouse.wheel(0, 50);
+    await test.expect.poll(() => page.evaluate(() => scrollY)).toBe(50);
+
+    const edit = q.menuitem("Edit");
+    const box = await edit.boundingBox();
+    if (!box) {
+      throw new Error("Edit menu item is not visible");
+    }
+    await page.mouse.move(box.x + box.width / 2, 8);
```
diegohaz: The hover point is a fixed viewport coordinate, and nothing in the test asserts the precondition the regression actually depends on: that `Edit` is clipped by the viewport top. The browser only scrolls a focused element that isn't already fully in view, so once the item happens to be fully visible, the implicit scroll has nothing to do and this test passes whether or not `preventScroll` is there.

The margin is thin. Measured on this branch at 1280x720, `Edit` sits at document top `36`, so after `wheel(0, 50)` its box is `y: -14, height: 56` — clipped by 14px — while `y = 8` stays inside the item for any document top up to 58.

This isn't theoretical. With the library fix reverted to `event.currentTarget.focus()` on both lines, and a single `paddingTop: 20` added to the sandbox's first `<section>`, this test goes green against the unfixed library:

```
✓  1 [chrome] › src/sandbox/menu-item-hover-7012/test-chrome-safari.ts:5:3 › react › hovering a partially visible menubar item does not scroll @react (2.3s)
```

Unmodified, the same test does fail on the reverted library (`scrollY` `50` → `36`), so the coverage is real today. It just isn't pinned to the condition that makes it real, and the failure mode of losing it is silence rather than a red test.

The real-world consumer here is the next person who touches this sandbox or the preview chrome around it: a 20px change anywhere above the menubar disarms the guard for [#7012](https://github.com/ariakit/ariakit/issues/7012) without any signal, and the page-scroll-under-the-pointer bug can come back unnoticed.

Asserting the precondition fixes it. `toBeInViewport({ ratio: 1 })` is the same assertion the neighbouring scroll fixture already uses in `app/src/sandbox/composite-page-scroll/test-browser.ts`, and deriving the hover point from the box removes the second layout assumption at the same time:

```suggestion
    const edit = q.menuitem("Edit");
    // The browser only scrolls a focused element that is not already fully in
    // view, so the regression only exists while this item is clipped.
    await test.expect(edit).not.toBeInViewport({ ratio: 1 });
    const box = await edit.boundingBox();
    if (!box) {
      throw new Error("Edit menu item is not visible");
    }
    await page.mouse.move(box.x + box.width / 2, box.y + box.height - 8);
```

I ran this exact version locally: it passes with the fix, still fails without it for the original reason (`scrollY` `50` → `36`), and under the 20px shift it now fails loudly instead of passing:

```
Error: expect(locator).not.toBeInViewport() failed
Expected: not in viewport
Received: in viewport
```

**Assessment**

- Severity: Medium (verified), because the shipped behavior fix is correct and what is at risk is only the guard that keeps #7012 from silently regressing.
- Priority: P1 (estimate, 80% confidence), because the fixture is new in this PR and the assertion is one line, so it is much cheaper now than after the coverage has already lapsed.
- Breakage likelihood: Possible (verified), because a 20px shift in the sandbox or in the preview chrome is enough and I reproduced exactly that.
- Fix effort: Easy (verified), because it is one assertion plus reusing the box already read on line 16.

app/src/sandbox/menu-item-hover-7012/test-chrome.ts
```diff
@@ -0,0 +1,42 @@
...
+    await page.mouse.wheel(0, 128);
+    await test.expect
+      .poll(() => menu.evaluate((element) => element.scrollTop))
+      .toBe(128);
+
+    const share = q.menuitem("Share");
+    const shareBox = await share.boundingBox();
+    if (!shareBox) {
+      throw new Error("Share menu item is not visible");
+    }
```
diegohaz, line 31: Same gap as the menubar test, one scrollport down: nothing here asserts that `Share` is actually clipped by the menu's scrollport, which is the only reason the browser has anything to scroll. If the menu's content ever fits, or `maxHeight` grows, this test keeps passing against a broken library instead of failing.

Measured on this branch at 1280x720, the margin is 28px: the menu box is `y: 412, height: 180` (bottom `592`) and `Share` is `y: 564, height: 56` (bottom `620`). That is also exactly the scroll the bug produces, since `scrollTop` maxes out at `156` and the unfixed library moves it `128` → `156`.

The precondition is one assertion away, and `menuBox` is already in scope:

```suggestion
    const share = q.menuitem("Share");
    const shareBox = await share.boundingBox();
    if (!shareBox) {
      throw new Error("Share menu item is not visible");
    }
    // The browser only scrolls a focused element that is not already fully in
    // view, so the regression only exists while this item is clipped by the
    // menu's scrollport.
    test
      .expect(shareBox.y + shareBox.height)
      .toBeGreaterThan(menuBox.y + menuBox.height);
```

I ran this version locally: it passes with the fix and still fails without it for the original reason (`scrollTop` `128` → `156`).

The consumer impact matches the sibling test: a later change to the item heights or to the menu's `maxHeight` in this sandbox quietly retires the guard for [#7012](https://github.com/ariakit/ariakit/issues/7012) in the scrollable-submenu path, which is the path that only runs in Chromium and therefore has no second engine covering it.

**Assessment**

- Severity: Medium (estimate, 85% confidence), because the behavior fix is correct and only the regression guard is at risk.
- Priority: P1 (estimate, 75% confidence), because it is the same one-line change while the fixture is still new.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because the 28px margin is wider than the menubar test's 14px and the sizes here are set explicitly in the sandbox rather than inherited from the preview chrome.
- Fix effort: Easy (verified), because both boxes are already read in the test.

packages/ariakit-react-components/src/menu/menu-item.tsx
```diff
@@ -159,15 +159,15 @@ export const useMenuItem = createHook<TagName, MenuItemOptions>(
         // the cursor back to the menu button.
         if (isWithinMenu) {
           if (event.currentTarget.hasAttribute("aria-expanded")) {
-            event.currentTarget.focus();
+            event.currentTarget.focus({ preventScroll: true });
```
diegohaz, line 165: Both `preventScroll` flags land here without a rationale, which makes them look like incidental options rather than the deliberate policy they encode. The sibling call site this change is aligning with does carry one, in `composite-hover.tsx`:

```ts
// Hovering an item is not a request to move the page; the pointer is
// already on what the user cares about.
compositeElement.focus({ preventScroll: true });
```

It matters more here than usual for two reasons. First, [#7012](https://github.com/ariakit/ariakit/issues/7012) states the expected outcome as each of these focus moves being "either given `preventScroll` with an explicit presentation where one is wanted, or left as it is with the reason recorded" — the decision for the two sites that changed is only recorded in this PR's description, and the code is where the next reader will look. Second, the non-obvious half of the decision is the part that isn't visible here at all: hover deliberately gets no explicit presentation to replace the suppressed scroll, because keyboard navigation still presents the item through the composite's own `scrollIntoView` step in `composite/utils.ts`. Without that noted, a later reader can reasonably read the missing presentation as an oversight and "fix" it back into a scroll on hover.

```suggestion
          if (event.currentTarget.hasAttribute("aria-expanded")) {
            // Hovering is not a request to move the page or the menu; the
            // pointer is already on what the user cares about. No explicit
            // presentation replaces the suppressed scroll here, because
            // keyboard navigation still presents the item through the
            // composite's own step.
            event.currentTarget.focus({ preventScroll: true });
```

The menubar branch on line 170 deserves the same treatment, though a shorter note that points back here would do.

**Assessment**

- Severity: Low (estimate, 85% confidence), because nothing user-facing depends on it today.
- Priority: P2 (estimate, 80% confidence), because it is a maintainability note rather than a defect, but it is cheap while the change is fresh.
- Breakage likelihood: Unlikely (estimate, 65% confidence), because it needs a later contributor to revisit these lines without reading #7012 or this PR.
- Fix effort: Easy (verified), because it is a comment on two adjacent branches.

diegohaz (commented):
Initial review of the full PR, diffed from the merge base `60e87c92b6d06fd9de5e903253f8081437f01efa` to head `4bb57fe`, then expanded into the surrounding menu, composite, and app-test context. No prior review threads existed on this PR, so there were no unresolved candidate findings to validate.

## Verdict on the change itself

The two-line fix is the right shape and the right size. `focusOnHover` runs only from `useCompositeHover`'s `onMouseMove`, so both call sites are hover-only and no keyboard or click path changes. It matches the policy `composite-hover.tsx` already applies one level down, and it correctly adds no explicit presentation to replace the suppressed scroll, since `presentItem` in `composite/utils.ts` only brings in an item marked with `data-autofocus` and hover never marks one. `withCompositeScrollPreserved` isn't applicable here because it only engages for a virtual-focus text-field composite. I also checked the rest of `src/menu/**` and `src/menubar/**`: after this change there are no remaining `focus()` calls in either directory without `preventScroll`. There is no Solid menu package, so React-only scope is right.

I reproduced the regression proof locally rather than taking the description's numbers on trust. Reverting both lines to `event.currentTarget.focus()` turns both tests red for the stated reasons, and restoring them turns both green:

- menubar: `scrollY` expected `50`, received `36`
- submenu: `scrollTop` expected `128`, received `156`

Browser-only coverage is the correct call: implicit focus scrolling is a real-browser layout behavior with no happy-dom counterpart, so a duplicate could not fail against the unfixed library. I confirmed the composite filename maps as intended — `--list` collects `test-chrome-safari.ts` under the `safari` project and collects nothing under `firefox`.

## Findings published inline

Both regression tests assert their outcome but not their precondition, which is what makes the outcome meaningful. Details, measurements, and validated suggestions are on the two test files. The menubar one is confirmed rather than argued: with the library fix reverted and a single `paddingTop: 20` added to the sandbox's first `<section>`, that test passes green against the broken library. A third, lower-priority comment asks for the rationale behind the two `preventScroll` flags to live in the code.

## Other candidates, with dispositions

**Scroll events are not recorded, only final offsets.** `app/src/test-utils/scroll.ts` exists precisely because "final offsets alone don't prove that nothing moved", and the neighbouring `composite-page-scroll` fixture pairs `recordScrollEvents` with `flushFrames`. These tests assert offsets only. Declined non-goal for this PR: the defect under test moves the offsets and leaves them moved, which the current assertions catch, as the red run above shows. The recorder guards a different shape (scroll and restore inside one task, visible as a scrollbar flash) that this change cannot produce. Worth folding in if you touch these tests anyway, but not something to add on my say-so.

**Seven of the nine call sites listed in #7012 are unchanged.** I checked each one rather than relying on the description. `combobox-item.tsx:207` and `:211` are in the printable-key `onKeyDownCapture`; `composite-container.tsx:76` is reached from `open()`, which is called only from `onKeyDown` and `onClick`; `:202`, `:214`, and `:226` are the input-listener, Escape, and Enter branches. None is hover-triggered, so the issue's stated outcome — hovering must not move the page, matching #2590 — is met by this PR and closing it is defensible. One wording nit for the description rather than the code: `combobox-list.tsx:83` runs from the list's `onFocus` handler, so "triggered by keyboard or click interactions" describes it only indirectly. Declined non-goal, no code change requested.

**Changeset does not name components built on `MenuItem`.** I considered asking for `MenuItemCheckbox` and `MenuItemRadio`, since both call `useMenuItem` and both are re-exported by `@ariakit/react`. Invalid on inspection: the submenu branch needs `aria-expanded` on the item, and the menubar branch needs a store without `hideAll`, which those two cannot bind to because they resolve their store from the menu context. Naming them would be an unverified behavioral claim, so leaving the entry at `MenuItem` is correct. The rest of the changeset checks out — `300-` prefix for a patch, both affected packages in the frontmatter, single-paragraph past-tense form, and `https://ariakit.com/reference/menu-item` returns `200`. No contributor credit is due, since the issue author is the PR author.

**New issue-number fixture instead of extending a semantic one.** I checked `menubar-interactions`, `composite-hover-virtual-focus-events`, and `composite-page-scroll`. The closest is `composite-page-scroll`, but it owns the `Composite`-level presentation policy from #6986, and this scenario needs a scrolled page plus a menubar and a scrollable menu, which would change that fixture's layout invariants. The lifecycle explicitly allows a separate fixture in that case, and `menu-item-hover-7012` follows the naming rule. No change requested.

## Not fully validated

I ran the Chrome project locally against this worktree's own dev servers; Safari, Firefox, mobile, and visual runs are covered only by this PR's green CI. I did not re-run `pnpm tsc`, `pnpm build`, or the happy-dom suites locally, since nothing in the diff should reach them and CI is green on all of them. The local experiments described above were reverted; the worktree is back at `4bb57fe` with a clean tree.

app/src/sandbox/menu-item-hover-7012/test-chrome-safari.ts
diegohaz: Fixed in [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd): the test now asserts that Edit is clipped and derives the hover point from its bounding box.

app/src/sandbox/menu-item-hover-7012/test-chrome.ts
diegohaz: Fixed in [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd): the test now asserts that Share crosses the menu scrollport edge and derives the hover point from that edge.

packages/ariakit-react-components/src/menu/menu-item.tsx
diegohaz: Fixed in [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd): the shared inline rationale now explains why hover suppresses implicit scrolling and why keyboard navigation still presents items explicitly.

diegohaz:
**Fixed**

- [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd) ([reply](https://github.com/ariakit/ariakit/pull/7066#discussion_r3714338056)): pinned the menubar clipping precondition and item-relative hover point.
- [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd) ([reply](https://github.com/ariakit/ariakit/pull/7066#discussion_r3714340225)): pinned the submenu scrollport clipping precondition and hover point.
- [`9c749e382`](https://github.com/ariakit/ariakit/commit/9c749e3825f81f83db988d0128c5fce81a558cbd) ([reply](https://github.com/ariakit/ariakit/pull/7066#discussion_r3714342497)): documented the hover-focus and keyboard-presentation policy.

diegohaz (commented):
Re-review of `9c749e38`, incremental from the head my previous review recorded, `4bb57fe`. All three threads from that review were already resolved, so there were no unresolved candidate findings to validate; I verified the three fixes directly against the new commit instead of taking the resolutions on trust, and re-examined the surrounding code for anything the changes newly expose.

**No new findings.** All three points are genuinely addressed, and the strengthened tests are better than what I suggested.

## What I verified

I re-ran the regression proof end to end against this head, on this worktree's own dev servers.

Green as-is, including the engine that only this file covers:

```
✓  2 [chrome] › .../test-chrome-safari.ts:5:3 › react › hovering a partially visible menubar item does not scroll @react
✓  1 [chrome] › .../test-chrome.ts:5:3 › react › hovering a partially visible submenu button does not scroll @react
✓  1 [safari] › .../test-chrome-safari.ts:5:3 › react › hovering a partially visible menubar item does not scroll @react
```

Still red with both `preventScroll` flags reverted, and — this is the part worth calling out — red on the *original* assertions, `scrollY` `50` → `36` and `scrollTop` `128` → `156`. A precondition assertion added in the wrong place can fire first and mask the signal it was meant to protect; these don't.

Both preconditions are load-bearing rather than decorative, which I checked by breaking each one deliberately:

- Menubar, with the library fix reverted and `paddingTop: 20` on the sandbox's first `<section>` — the exact edit that used to produce a silent green — now fails loudly: `expect(locator).not.toBeInViewport() failed / Expected: not in viewport / Received: in viewport`.
- Submenu, with the menu's `maxHeight` raised to `208`, the one value where `scrollTop` `128` is still reachable *and* `Share` exactly fits, fails on the new assertion at its precise boundary: `Expected: > 592 / Received: 592`.

The two-assertion form you used on `test-chrome.ts` is stronger than what I proposed. Bracketing `Share` on both sides of `menuBottom` also proves the hover point `menuBottom - 8` lands inside the item, which my single `toBeGreaterThan` left implicit, and hoisting `menuBottom` removes the duplicated expression the original had. Worth noting for anyone reading later: because `Share` is the last item, any shrink of that menu's content drops max `scrollTop` below `128` and trips the wheel assertion first, so this test has two independent guards against quietly ceasing to reproduce.

I also re-checked that nothing in the incremental diff disturbs the fix itself: the two `focus({ preventScroll: true })` calls are unchanged, and the added comment is the only change in package source.

## One minor observation, not a request

The new rationale at `menu-item.tsx:157-159` sits directly against the pre-existing `// If the menu item is also a submenu button...` comment with no blank line, so the six lines read as one block belonging to the `isWithinMenu` branch. The rationale is shared by both focus calls, and a reader arriving at the menubar branch on line 173 sees only that branch's own comment. A blank line between the two blocks would signal the wider scope. Declined non-goal on my side rather than something to push: it is comment formatting, the two call sites are ten lines apart, and the substance I asked for is there.

## Not validated locally

Firefox, mobile, and visual runs, plus `pnpm tsc` and `pnpm build`, are covered only by this PR's CI, which is now green across all checks including both Safari shards and Gate. My local experiments were reverted; the worktree is clean at `9c749e38`.

PR #7072, opened 2026-08-06
Refine code comments across the repository

diegohaz:
## Motivation

Comments across the active workspaces vary from useful rationale to redundant narration and oversized implementation histories. The cleanup should reduce noise without making regexes, parsers, state machines, effect lifecycles, or regression fixtures harder to understand.

## Solution

Remove comments that repeat nearby code or preserve dead snippets, condense long explanations, and retain concise guideposts for cryptic syntax, ordering, browser behavior, and setup or cleanup invariants. Keep issue and pull request links where they provide useful extended context, preserve useful effect-block narration, and leave the legacy `website/` workspace unchanged.

For example, rendered task-list markup keeps the rationale for a regex that would otherwise be cryptic:

```ts
// Task-list markup may wrap the disabled checkbox in rendered HTML.
const checkboxRegex =
  /<input\s+[^>]*type=["']checkbox["'][^>]*disabled[^>]*>/i;
```

## Testing

- `pnpm lint`
- `pnpm lint-css`
- `pnpm tsc`
- `node app/src/lib/build-styles.ts --check`
- Focused `pnpm test --run` across 15 affected test files (173 tests)
- `git diff --check`

No changeset is needed because this is a comment-only cleanup with no user-facing behavior change.

## Review gate reassessment

<details>
<summary>Cycle 3 (2026-08-06): Continue the targeted comment cleanup</summary>

**Assessment**

At the cycle-3 boundary, staged snapshot `2ab9dc35…` spanned 128 active-repository files but added no behavioral contract, state, helpers, abstractions, or test machinery. Its net reduction of 1,104 lines lowered the maintenance surface.

All three finding-bearing cycles identified isolated low-severity comment-policy issues: inaccurate narration and line wrapping, comment placement, then one remaining inline restatement. This pattern reflected the breadth and subjective judgment of a repository-wide comment audit, not excessive implementation complexity. Expected frequency and user impact were low and limited to maintainer clarity; no runtime defect emerged.

**Decision**

Continue the current direction without reverting, redesigning, or narrowing scope. The cycle-3 finding was addressed by deleting only the redundant `// Skip files that were merged` narration. No replacement, test, or follow-up issue was warranted, and the resulting exact snapshot received a fresh clean review.

Intentionally excluded: the legacy `website/` workspace, blanket removal of explanatory comments for cryptic code, removal of useful effect-block narration, behavioral refactors, and new public contracts.

</details>


app/src/styles/styles.json
```diff
@@ -7458,10 +7458,6 @@
               "name": "--ak-list-border-top",
               "value": "calc(\n    var(--ak-list-leading) + var(--ak-list-border-gap) + var(--ak-frame-padding)\n  )"
             },
-            {
-              "name": "/* Border segment */",
-              "value": {}
-            },
             {
               "name": "&::after",
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`86da42a57`](https://github.com/ariakit/ariakit/commit/86da42a5722d5b3884e1951328118effb703cb39): removed the stale `/* Border segment */` line from the inline snapshot so it matches the generated style output. The focused styles suite now passes all 21 tests.

diegohaz:
**Fixed**

- [`86da42a57`](https://github.com/ariakit/ariakit/commit/86da42a5722d5b3884e1951328118effb703cb39) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3729894311)): Updated the `styleDefToCss` inline snapshot to match the generated `ak-list-item-ol-border` CSS after the redundant comment removal. The focused styles suite passes all 21 tests.

app/astro.config.ts
```diff
@@ -80,18 +78,9 @@ export default defineConfig({
...
-    // content.config.ts, astro-remote in markdown.astro). With the Cloudflare
-    // adapter they get optimized lazily during the first cold dev request, and
-    // the mid-request optimizer reload resets React's hook dispatcher,
-    // producing "Invalid hook call" errors. Pre-including them forces the
-    // optimizer to bundle them at startup so no reload happens mid-request. The
-    // Cloudflare adapter merges this list into the SSR environment's
-    // optimizeDeps.include.
+    // Pre-optimize bare SSR imports so the Cloudflare adapter cannot reload
+    // React mid-request.
+    // https://github.com/withastro/astro/issues/17166
```
diegohaz, line 85: This PR deletes 23 of the repository's 27 `TODO` comments. Counting with `git grep -c TODO` over `*.ts`, `*.tsx`, `*.astro`, `*.css`, `*.yml` excluding `website/`: 27 on the merge base `38da662a7`, 4 on this head. The description does not mention TODOs — its stated scope is comments that repeat nearby code, dead snippets, and over-long explanations, and a TODO is none of those. It is the only durable record that some work was deliberately deferred.

The two in this file are the clearest loss, because both marked a workaround that should be **deleted** when an upstream condition is met:

- `// TODO: Remove this workaround once Astro isolates optimizer cache writes across dev and check commands.`
- `// TODO: Remove this workaround once https://github.com/withastro/astro/issues/17166 is fixed.`

The upstream link survived, but the removal trigger did not. As rewritten, this pre-include list reads as permanent design: when withastro/astro#17166 ships, nothing in the repository says this block can go, and the same applies to the separate `cacheDir` workaround above.

Full inventory of the removed TODOs:

| File | TODO |
| --- | --- |
| `app/astro.config.ts` | Remove workaround once Astro isolates optimizer cache writes |
| `app/astro.config.ts` | Remove workaround once withastro/astro#17166 is fixed |
| `app/src/examples/combobox-group/test.ts` | Add composition util to `@ariakit/test` |
| `app/src/sandbox/combobox-group-basic/test.ts` | Add composition util to `@ariakit/test` |
| `packages/ariakit-components/src/select/select-store.ts` | Revisit; probably deprecate with ComboboxProvider as SelectProvider parent |
| `packages/ariakit-react-components/src/combobox/combobox.tsx` | Try derived state |
| `packages/ariakit-react-components/src/combobox/combobox.tsx` | Test this (selection-range restore) |
| `packages/ariakit-react-components/src/combobox/combobox.tsx` | Test this (async items active value reset) |
| `packages/ariakit-react-components/src/composite/composite-container.tsx` | Add data-attribute to indicate expanded |
| `packages/ariakit-react-components/src/composite/composite.tsx` | Test this (composed store blur) |
| `packages/ariakit-react-components/src/dialog/dialog.tsx` | Do this more generically in DisclosureContent |
| `packages/ariakit-react-components/src/focusable/focusable.tsx` | Test Focusable contentEditable |
| `packages/ariakit-react-components/src/hovercard/hovercard.tsx` | Maybe use `state.anchorElement` directly |
| `packages/ariakit-react-components/src/menu/menu.tsx` | Test this (aria-owns sibling tab order) |
| `packages/ariakit-react-components/src/portal/portal.tsx` | Warn about portals as the `document.body` element |
| `packages/ariakit-react-components/src/radio/radio.tsx` | Maybe do this in the radio store directly |
| `packages/ariakit-solid-components/src/as/as.tsx` (x2) | Replace with LazyDynamic |
| `packages/ariakit-solid-components/src/heading/heading.tsx` | Replace with LazyDynamic |
| `packages/ariakit-solid-utils/src/system.tsx` (x2) | Replace with LazyDynamic |
| `packages/ariakit-test/src/press.ts` (x2) | Implement repeat |

Two of these also documented an unimplemented public option: `press.ts` accepts `KeyboardEventInit`, whose `repeat` flag is forwarded onto the event but never produces repeated keydowns. `// TODO: Implement repeat` was the only place that said so.

What makes this look like collateral rather than policy is that the PR keeps four TODOs and actively improves three of them — `focus-trap-region.tsx:38`, `role.tsx:57`, and `system.tsx:32` were rewritten with a permalink, and `shims.ts:270` keeps `// TODO: Remove this shim once the upstream fix ships.` Those four show the intended treatment; the other 23 did not get it.

Suggested resolution: restore them in the same style as the four that were kept (one line plus a permalink where one exists), or register the ones worth tracking as GitHub issues and reference those issues from the code, so the deferred work survives as something other than a deletion. For this file specifically:

```suggestion
    // TODO: Remove this workaround once withastro/astro#17166 is fixed.
    // Pre-optimize bare SSR imports so the Cloudflare adapter cannot reload
    // React mid-request.
    // https://github.com/withastro/astro/issues/17166
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because nothing breaks at runtime; the cost is deferred work that no longer has an owner or a trigger.
- Priority: P1 (estimate, 80% confidence), because the deletions are introduced here, are invisible in the description, and are cheap to reverse now but expensive to reconstruct later.
- Breakage likelihood: Likely (verified, 90% confidence), for the upstream-gated workarounds specifically: `astro/issues/17166` and the happy-dom composition gap will eventually resolve, and nothing in the tree will then prompt the cleanup.
- Fix effort: Easy (estimate, 90% confidence), because it is a restore-or-register decision per TODO, with no code changes.

packages/ariakit-react-components/src/combobox/combobox-select.tsx
```diff
@@ -172,6 +172,8 @@ export const useComboboxSelect = createHook<TagName, ComboboxSelectOptions>(
       }
     });
 
+    // Mirror the custom select with a native control so browser autofill can
+    // update the combobox store.
     props = useWrapElement(
```
diegohaz: This comment now describes the wrong wrapper. On the merge base it sat directly above the `useWrapElement` call that renders the hidden native `<select>`; here it was moved up to the `useWrapElement` at line 177, which only renders `<ComboboxScopedContextProvider>` and has nothing to do with a native control or with autofill. The wrapper that actually mirrors the custom select — line 213, the one rendering `<select style={getVisuallyHiddenStyle()} … onChange={…}>` — is now the uncommented one.

This is the failure mode the PR is otherwise removing: a reader scanning for why a native control exists finds the explanation attached to a context provider, and the code that needs it has none.

Moving it back:

```suggestion
    props = useWrapElement(
```

and restoring it above the wrapper at line 213:

```ts
    // Mirror the custom select with a native control so browser autofill can
    // update the combobox store.
    props = useWrapElement(
      props,
      (element) => {
        if (!name) return element;
```

**Assessment**

- Severity: Low (verified, 95% confidence), because it misleads readers only; no runtime behavior depends on it.
- Priority: P1 (estimate, 85% confidence), because the PR introduced it, it contradicts the PR's own goal, and the fix is a two-line move.
- Breakage likelihood: Likely (verified, 95% confidence), in the sense that every reader of this file now meets the wrong pairing — the mismatch is unconditional, not situational.
- Fix effort: Easy (verified, 95% confidence), because it is a comment relocation with no code change.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -523,15 +513,9 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
-      // virtual focus does. Scroll explicitly instead, so every engine behaves
-      // the same whether or not focus stayed put. It happens before the focus
-      // so a focus handler that presents something else scrolls last and wins.
-      // Re-read focusability here rather than trusting the snapshot taken in
-      // the effect body: the scroll is conditional on it while the focus below
-      // is not, so a stale `true` would scroll to an element that can no
-      // longer be focused.
+      // Safari may drop focus scrolling when virtual focus redirects it. Scroll
+      // first, then re-check focusability so stale state cannot move the page
+      // toward an element that no longer accepts focus.
```
diegohaz: "Scroll first, then re-check focusability" states the opposite of what the code does: `isFocusable(element)` is the guard, and `scrollIntoView` runs inside it. The focusability check comes first.

The condensation dropped the referent that made the original ordering claim true. On the merge base, "It happens before the focus" meant the *scroll* happens before `element.focus({ preventScroll: true })` on line 522 — two different subjects, scroll-vs-focus, not scroll-vs-check. Merging them into one sentence produced a sequence that no longer matches either.

The merge base also carried a second reason that is now gone: the re-read exists because *the scroll is conditional on focusability while the focus below is not*, so a stale `true` from the effect body would scroll to an element that can no longer be focused. "stale state cannot move the page" gestures at that but never says which state is stale or why only the scroll is guarded.

```suggestion
      // Safari may drop the browser's own focus scroll when virtual focus
      // immediately redirects focus, so scroll explicitly before focusing.
      // Re-read focusability here rather than trusting the effect body's
      // snapshot: only the scroll is conditional on it, so a stale `true`
      // would scroll to an element that can no longer be focused.
```

**Assessment**

- Severity: Low (verified, 90% confidence), because the comment is wrong but the code is unchanged and correct.
- Priority: P1 (estimate, 80% confidence), because a comment that inverts the order of its own guard is worse than the narration this PR removes elsewhere, and it is introduced here.
- Breakage likelihood: Possible (estimate, 75% confidence), because a maintainer reordering these lines to match the comment would drop the focusability guard and scroll to a non-focusable element.
- Fix effort: Easy (verified, 95% confidence), because it is a rewrite of three comment lines.

packages/ariakit-scripts/src/perf.ts
```diff
@@ -1429,27 +1405,15 @@ async function measureIteration(
...
-/**
- * Whether `error` is a Playwright timeout thrown by a navigation. Only
- * navigations are retried: other timeouts, such as interaction or verify
- * steps, can reflect real behavior of the code under measurement. Page-load
- * measures are the one case where the measured interaction is itself a
- * navigation, so a page load that legitimately exceeds the navigation bound
- * is retried too; that stays safe because the failed attempt records nothing
- * and the retry fails the same way.
- */
+/** Returns whether a retryable navigation—not interaction—timed out. */
```
diegohaz: "navigation—not interaction" is false for page-load measurements, and that was the one case the merge-base comment existed to cover.

`measurePageLoad` (line 1585) passes `gotoAndSettle` as the measured callback with `loadPage: false`, so for those tests the measured interaction *is* a `page.goto`. A page load that legitimately exceeds `PLAYWRIGHT_TEST_TIMEOUT` therefore returns `true` here and gets retried, which the new one-liner says does not happen. The merge base named the exception and, more importantly, said why it stays sound: the failed attempt records nothing, so the retry is an independent sample rather than a second chance at a slow result.

That safety argument is the part worth keeping — it is what stops someone from "fixing" this predicate to exclude page-load tests and, in doing so, turning an infra wedge into a hard failure.

```suggestion
/**
 * Whether `error` is a Playwright timeout thrown by a navigation. Interaction
 * and verify timeouts are not retried, since they can reflect real behavior of
 * the code under measurement. Page-load measures are the exception — the
 * measured interaction is itself a navigation — and retrying them stays safe
 * because the failed attempt records nothing.
 */
```

**Assessment**

- Severity: Low (verified, 90% confidence), because the predicate itself is unchanged; only its documented contract is now wrong.
- Priority: P2 (estimate, 75% confidence), because it misleads rather than breaks, but it sits on retry semantics that perf CI correctness depends on.
- Breakage likelihood: Possible (estimate, 70% confidence), because the mismatch only bites when someone edits this predicate or debugs a retried page-load measurement.
- Fix effort: Easy (verified, 95% confidence), because restoring the exception costs a few JSDoc lines.

packages/ariakit-test/src/press.ts
```diff
@@ -379,9 +379,6 @@ function isPressable(element: Element) {
   return isFocusable(element) || element.tagName === "BODY";
 }
 
-// Fires `keydown` and applies the browser's default keydown behavior for the key
-// (moving focus with `Tab`, the caret with the arrow keys, etc.). Returns whether
-// the default action was allowed so the caller can gate the keyup behavior on it.
 async function pressKeyDown(
```
diegohaz, line 385: The comment removed from above this function carried a return contract that neither the name nor the signature expresses:

> Fires `keydown` and applies the browser's default keydown behavior for the key (moving focus with `Tab`, the caret with the arrow keys, etc.). Returns whether the default action was allowed so the caller can gate the keyup behavior on it.

`pressKeyDown` returns `Promise<boolean>`, and the boolean is not "did it run" — it is `defaultAllowed`, which `press` reads on line 481 and forwards to `pressKeyUp` on line 490 to decide whether the keyup default behavior (clicking buttons, checkboxes, radios with `Space`) may run at all. The repo's own comment guidance keeps JSDoc "when it adds contract, parameter, return, or non-obvious behavioral information that the name and types do not already express," which is exactly this case: a bare `boolean` with an inverted-sounding polarity risk.

The sibling `pressKeyUp` lost its comment too, which matters less — it returns nothing — but the keydown one is worth keeping. A condensed form is enough:

```suggestion
// Fires `keydown` and applies the key's default behavior (focus for `Tab`, the
// caret for arrows). Returns whether the default was allowed, which gates the
// matching keyup behavior.
async function pressKeyDown(
```

**Assessment**

- Severity: Low (verified, 90% confidence), because it costs comprehension time in an internal helper, not correctness.
- Priority: P2 (estimate, 75% confidence), because it is a removal of genuine contract information rather than narration, but the call sites are close enough to reconstruct it.
- Breakage likelihood: Possible (estimate, 70% confidence), because a caller that reads the boolean as "handled" instead of "default allowed" would gate keyup backwards.
- Fix effort: Easy (verified, 95% confidence), because it restores three comment lines.

packages/ariakit-test/src/shims.test.ts
```diff
@@ -545,14 +516,11 @@ test("preserves form and select proxy identity across ancestor APIs", () => {
...
-      // Elements outside the parent are still reported as outside.
       expect(parent.contains(outside)).toBe(false);
       expect(parent.contains(container)).toBe(false);
       expect(parent.contains(null)).toBe(false);
-      // The other direction never regressed, but it must keep working.
       expect(container.contains(parent)).toBe(true);
       expect(container.contains(grandchild)).toBe(true);
-      // The divergence outlives the subtree's removal from the document.
       container.remove();
       expect(parent.contains(grandchild)).toBe(true);
```
diegohaz, line 529: The `container.remove()` on line 524 is now unexplained test setup, and it reads as a stray duplicate of the cleanup in the `finally` block on line 527. The removed comment — "The divergence outlives the subtree's removal from the document" — was the only thing saying that the removal is part of the assertion: it proves the repaired parent pointer survives detachment, which is why line 525 asserts `contains` still holds afterwards.

The repo's comment guidance calls this case out directly ("Explain unusual test setup: state the condition the setup creates"), and mid-`try` mutation that mirrors the cleanup is exactly the kind of line a future reader deletes as redundant.

Two other removals in this file drop the same class of information:

- Lines 13–17: `width: 1, height: 1` and `toHaveLength(0)` are magic values produced by the shim. The removed pair ("reports a 1x1 rect for connected, visible elements…" / "…and an empty list for elements that aren't visible") named the rule; without it the expectations look arbitrary, and `disconnected` alone does not convey that visibility, not connection, is the shim's criterion.
- Line 521: `expect(container.contains(parent)).toBe(true)` lost "The other direction never regressed, but it must keep working," which was the only signal that this assertion is a guard rather than the case under test.

```suggestion
      expect(container.contains(parent)).toBe(true);
      expect(container.contains(grandchild)).toBe(true);
      // The divergence outlives the subtree's removal from the document.
      container.remove();
      expect(parent.contains(grandchild)).toBe(true);
```

The same pattern appears in `packages/ariakit-components/src/composite/composite-store.test.ts`. In `test("scans within the active item's row on grids")` (line 316), `expect(store.next({ activeId: "a1" })).toBe("a3")` and `expect(store.next({ activeId: "a3" })).toBeUndefined()` now require the reader to reconstruct both the disabled-item skip and the row boundary from the fixture above; the deleted one-liners stated which rule each assertion covers.

**Assessment**

- Severity: Low (verified, 90% confidence), because coverage is unchanged; the cost is comprehension and the risk of a future reader deleting load-bearing setup.
- Priority: P2 (estimate, 70% confidence), because these are removals of genuine intent rather than narration, though none blocks the change.
- Breakage likelihood: Possible (estimate, 70% confidence), because the unexplained `container.remove()` is a realistic candidate for a "duplicate cleanup" simplification that would silently drop the detachment assertion.
- Fix effort: Easy (verified, 95% confidence), because it restores a handful of one-line comments.

app/src/lib/source.ts
```diff
@@ -118,7 +112,8 @@ export type ImportPathType =
   | "export-type"
   | "export-all";
 
-// Central registry of patterns and their classification
+// Text and module-specifier helpers
```
diegohaz: This section header lands on the wrong declaration. The next thing below it is `PATTERNS`, a `RegExp` → `ImportPathType` registry — not a text or module-specifier helper. Those helpers (`sortStrings`, `getOrCreate`, `cleanupText`, `getPathFromGroups`) start further down, after `PATTERNS`.

Two edits collided here: the banner replacement inserted the new header at the old `// ====== Types ======` boundary, and the same hunk deleted `// Central registry of patterns and their classification`, which was accurate and non-obvious — it is what tells a reader that this array, not each individual regex constant, is the single place where a pattern gets its `ImportPathType`.

Splitting them keeps both correct:

```suggestion
// Central registry of patterns and their classification
```

and moving `// Text and module-specifier helpers` down to just above `sortStrings` (line 128), where the helpers actually begin.

**Assessment**

- Severity: Low (verified, 90% confidence), because it only misdirects navigation in a 700-line module.
- Priority: P2 (estimate, 70% confidence), because it is introduced here and cheap to fix, but harmless at runtime.
- Breakage likelihood: Likely (verified, 90% confidence), in that every reader using these headers to navigate hits the mismatch; the header is unconditionally attached to the wrong block.
- Fix effort: Easy (verified, 95% confidence), because it is a one-line restore plus moving one comment line.

diegohaz (commented):
## Review summary

Initial review. Baseline: merge base `38da662a7` → head `86da42a5`, 129 files, +532/−1637. Reviewed the whole diff file by file, plus surrounding context in the modules each comment describes.

### What I checked

- `pnpm tsc` (pass), `pnpm lint` (pass, 2047 files formatted correctly), `node app/src/lib/build-styles.ts --check` (pass).
- `pnpm test --run` on `app/src/lib/styles.test.ts`, `packages/ariakit-components/src/composite/composite-store.test.ts`, `packages/ariakit-test/src/shims.test.ts`, `app/src/lib/jsdoc-loader.test.ts` — 4 files, 68 tests, all passing.
- **Link verification.** All 40 newly added URLs were resolved against the GitHub API. Every issue/PR number exists, every title is topically consistent with the code it annotates, and all ten `#discussion_r…` permalinks resolve to real review comments whose `path` matches the file the link was added to. Two are anchored in a different file than their original thread (`…/6404#discussion_r3454880546` from `dispatch.ts` now cited in `__utils.ts`, `…/6256#discussion_r3366022385` from `__utils.ts` now cited in `shims.ts`) — both correct, since each thread is about moving that very code. No fabricated or mismatched links.
- **Comment-only verification.** Scanned every changed line that is not a comment. Four non-comment changes, all behavior-preserving: the empty-branch collapse in `app/src/components/code-block-content.astro:241` (`!(a >= x || b <= y)` → `a < x && b > y`, verified equivalent), `} else if` reflow where a comment used to sit between the brace and `else` (`composite-container.tsx`, `tab-store.ts`), a `.map()` reflow in `docs.ts`, and moving `createdAt: string; // ISO string` to a preceding line. Worth noting in the description, which currently says comment-only.
- **Generated output.** The CSS comment removals do change `app/src/styles/styles.json` and the CSS that `styleDefToCss` renders. I confirmed `styleDefsToCss` has no consumers in `app/src` outside `lib/styles.ts` and its test, so nothing user-visible changes and no changeset is needed. `website/` is untouched (0 files), as stated.

### Existing threads

One review thread exists (`PRRT_kwDOBjT0Fc6XBYPI`, Codex on `app/src/styles/styles.json`, stale `/* Border segment */` snapshot) and it is resolved, so it was not a candidate finding. I independently confirmed the fix in `86da42a57` holds: `app/src/lib/styles.test.ts` passes on this head.

### Findings published inline

1. **23 of 27 `TODO` comments deleted repo-wide** (`app/astro.config.ts`) — the largest issue here, out of the description's stated scope, and invisible in it. Includes two upstream-gated "remove this workaround once…" triggers and the only record that `press`'s `repeat` option is unimplemented. The four TODOs that survived (three of them upgraded with permalinks) show the intended treatment.
2. **`combobox-select.tsx:175`** — the native-select/autofill rationale was moved onto the `ComboboxScopedContextProvider` wrapper; the wrapper that renders `<select>` (line 213) is now the uncommented one.
3. **`dialog.tsx:516`** — "Scroll first, then re-check focusability" inverts the code's actual order; the original's subject was scroll-before-*focus*, and the reason the re-read exists was dropped.
4. **`perf.ts:1408`** — "navigation—not interaction" is false for `measurePageLoad`, where the measured interaction is itself a `page.goto`; the retry-safety argument went with it.
5. **`press.ts:382`** — `pressKeyDown`'s `defaultAllowed` return contract, which gates keyup behavior, is no longer documented anywhere.
6. **`shims.test.ts:524`** — `container.remove()` is now unexplained setup that reads as a duplicate of the `finally` cleanup; also covers the 1x1 magic rect and the `composite-store.test.ts` grid assertions.
7. **`source.ts:115`** — `// Text and module-specifier helpers` sits above the `PATTERNS` registry, and the accurate `// Central registry of patterns and their classification` was dropped in the same hunk.

### Candidates not published, with dispositions

- **`composite-store.ts:562`, removed `// move() does nothing`** — Declined non-goal. The comment marked why `move(undefined)` returns early while `move(null)` is meaningful (`combobox-cancel.tsx` relies on the `null` path). The strict `===` already signals the distinction and the `null` behavior is part of the documented public API, so the loss is preferential rather than actionable.
- **`combobox.tsx`, removed "See tag-combobox test named …" pointers (x2)** — Declined non-goal. These pointed at the regression tests covering the deselection guards, but the replacement comments do state the rule, and test-name pointers go stale on rename.
- **`table-of-contents.astro:80`, dropped directional rule** — Declined non-goal. "scrolling down → last visible link, scrolling up → first visible link" is no longer stated, only implied by "directionally relevant." Derivable from the `isScrollingDown` branch a few lines below, so I left it out of the inline set; worth folding in if you touch the file for the other fixes.
- **`app/src/styles/*.css`, removed grouping headers** — Declined non-goal. Some (`/* ak-list configuration */`, `/* ak-list variants */`, `/* ak-list-item elements and variants */`) were section markers spanning several utilities rather than restatements of the `@utility` line that follows, so scannability drops in the 300-line files. This is a navigation preference, not a correctness issue.
- **`vitest.setup.framework.ts:51`** — Declined non-goal. The rewritten block keeps `/* … */` where the repo's comment standard asks for `//` on every line of a non-JSDoc comment. Pre-existing style that this PR did not worsen, though the file was already being rewritten. I did verify the new pointer is accurate: the Playwright targets it defers to are really defined in `app/playwright.config.ts` (`chrome`, `firefox`, `safari`, `ios`, `android`, plus the `browser`/`mobile` groups).
- **`perf.ts:32`, `playwright.config.ts:13`** — Declined non-goal. The rationale for the exact magic values (30s/10s timeouts, `0` meaning "let the preview server pick") is thinner now, but each replacement still explains the purpose.

No follow-up issues were registered; every confirmed finding is either resolvable in this PR or explicitly declined above.

### Where the judgment is good

Worth saying explicitly, since the bulk of this diff is sound: the `dialog-nested-variants` fixture labels were correctly deleted — every `{/* Sibling no portal portal */}` was already restated by both the state variable and the `<DialogHeading>` immediately below. The `examples/` and `app/src/sandbox/` duplicate pairs (`menu-nested-combobox`, `combobox-radix-select`, `combobox-tabs`) were kept in sync. Dead commented-out code (`orchestrate.ts` debug block, `supports-inert.ts` toggle, the SVG `radialGradient`) is a clean removal, and the "summarize locally, link the discussion" pattern is applied consistently and with correct links.

### Not fully validated

Browser and visual Playwright suites were not run locally; Firefox, Safari, and the two perf jobs were still in progress at review time. Comment-only edits should not reach them, and `lint`/`tsc`/`build-styles --check` cover the CSS and generated-output paths, but CI remains the authority there. `pnpm build` and packed-contents verification were not run and are not required — no `files`, `exports`, or packaging changes.

app/astro.config.ts
diegohaz: Fixed in [`da93b5bd8`](https://github.com/ariakit/ariakit/commit/da93b5bd842e8b4fa2d26bdfc76ede3eeeb866c5): restored all 23 deferred-work TODOs with concise wording, including the two removal triggers in this file.

packages/ariakit-react-components/src/combobox/combobox-select.tsx
diegohaz: Fixed in [`ba50d8771`](https://github.com/ariakit/ariakit/commit/ba50d8771db921ace1b9a020fce566343c6f80eb): moved the autofill explanation to the wrapper that renders the hidden native select.

packages/ariakit-react-components/src/dialog/dialog.tsx
diegohaz: Fixed in [`da93b5bd8`](https://github.com/ariakit/ariakit/commit/da93b5bd842e8b4fa2d26bdfc76ede3eeeb866c5): corrected the Safari ordering explanation and restored why focusability is re-read before the conditional scroll.

packages/ariakit-scripts/src/perf.ts
diegohaz: Fixed in [`9bbad82d3`](https://github.com/ariakit/ariakit/commit/9bbad82d338dbaeccf4b0eeac23d6f5d6e73f1ce): documented the page-load navigation exception and why a failed retry remains an independent sample.

packages/ariakit-test/src/press.ts
diegohaz: Fixed in [`da93b5bd8`](https://github.com/ariakit/ariakit/commit/da93b5bd842e8b4fa2d26bdfc76ede3eeeb866c5): restored the `defaultAllowed` return contract that gates the matching keyup behavior.

packages/ariakit-test/src/shims.test.ts
diegohaz: Fixed in [`2afb2920f`](https://github.com/ariakit/ariakit/commit/2afb2920f9dab70af571a718c43eae89dfb60a56): restored concise setup and assertion intent for the shim and composite grid tests.

app/src/lib/source.ts
diegohaz: Fixed in [`40b34b8fa`](https://github.com/ariakit/ariakit/commit/40b34b8fa1b956b88b0306d9c4c92b380a97f19e): labeled the pattern registry accurately and moved the helper heading above the helper functions.

diegohaz:
**Fixed**

- [`da93b5bd8`](https://github.com/ariakit/ariakit/commit/da93b5bd842e8b4fa2d26bdfc76ede3eeeb866c5) ([TODO reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730259635), [dialog reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730261497), [press reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730263158)): Restored all 23 deferred-work TODOs, corrected the Safari scroll and focus rationale, and reinstated the keydown return contract.
- [`ba50d8771`](https://github.com/ariakit/ariakit/commit/ba50d8771db921ace1b9a020fce566343c6f80eb) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730260723)): Moved the native-select autofill explanation to the hidden-select wrapper.
- [`9bbad82d3`](https://github.com/ariakit/ariakit/commit/9bbad82d338dbaeccf4b0eeac23d6f5d6e73f1ce) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730262122)): Restored the page-load navigation exception and independent-sample retry rationale.
- [`2afb2920f`](https://github.com/ariakit/ariakit/commit/2afb2920f9dab70af571a718c43eae89dfb60a56) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730263987)): Restored concise intent comments for unusual shim setup and composite grid boundaries.
- [`40b34b8fa`](https://github.com/ariakit/ariakit/commit/40b34b8fa1b956b88b0306d9c4c92b380a97f19e) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730264764)): Corrected the pattern-registry label and moved the helper heading above the helpers.

packages/ariakit-components/src/composite/composite-store.test.ts
```diff
@@ -369,24 +363,19 @@ test("moves vertically in one-dimensional composites", () => {
...
   expect(store.down({ activeId: "one" })).toBe("three");
-  // Vertical loops wrap around just like horizontal ones.
   expect(store.down({ activeId: "three", focusLoop: true })).toBe("one");
   expect(store.up({ activeId: "one", focusLoop: true })).toBe("three");
 });
 
 test("handles falsy item ids and row ids", () => {
-  // An item with an empty string id must not return itself when looping.
   const single = createComposite([{ id: "" }]);
   expect(single.next({ activeId: "", focusLoop: true })).toBeUndefined();
```
diegohaz, line 375: The restoration in `2afb2920f` covered `test("scans within the active item's row on grids")` — the test I named — but the four tests below it still have bare assertions, so the file is now inconsistent with itself: one grid test explains each expectation, the next one does not.

The strongest remaining case is this one. `createComposite([{ id: "" }])` followed by `next({ activeId: "", focusLoop: true })` returning `undefined` looks like a bug on its face — a single-item loop that refuses to return its only item. The deleted comment ("An item with an empty string id must not return itself when looping") was the only thing marking it as the intended contract rather than a falsy-id defect waiting to be "fixed."

Line 379 has the same problem: `createComposite([{ id: "b", rowId: "" }, { id: "a" }])` is testing that `rowId: ""` is treated like `rowId: undefined`, which is why `down` from `b` reaches `a`. Nothing in the assertions says the empty string is the point.

```suggestion
  // An item with an empty string id must not return itself when looping.
  const single = createComposite([{ id: "" }]);
  expect(single.next({ activeId: "", focusLoop: true })).toBeUndefined();
```

and at line 379:

```ts
  // Empty string row ids behave like undefined row ids when moving vertically.
  const store = createComposite([{ id: "b", rowId: "" }, { id: "a" }]);
```

One more worth restoring is line 345, `expect(store.previous({ activeId: "a1", focusLoop: true })).toBe("a2")`: the backward loop wraps past the disabled `a3` to land on `a2`, which is not readable from the assertion without cross-referencing the fixture five lines up.

**Assessment**

- Severity: Low (verified, 90% confidence), because coverage is unchanged and every test passes; the cost is that intentional edge-case contracts read as accidents.
- Priority: P2 (estimate, 75% confidence), because it is the same class as the finding already accepted for this file and cheap to finish, but nothing depends on it.
- Breakage likelihood: Possible (estimate, 70% confidence), because the falsy-id expectations are the ones a future contributor is most likely to read as a defect and change.
- Fix effort: Easy (verified, 95% confidence), because it restores three one-line comments.

diegohaz (commented):
## Re-review summary

Re-review. Baseline: the head of my previous review, `86da42a5` → current head `40b34b8f` (5 commits, 22 files, +54/−7). Newly discovered findings come from those commits plus surrounding code they expose; I also re-verified each earlier finding against the current head rather than relying on the replies.

### Verification of the seven earlier findings

All seven are genuinely fixed. What I checked, beyond reading the replies:

1. **TODOs** — `git grep -c TODO` outside `website/` is back to **27 on this head, matching the merge base**. I diffed all 23 restorations against their originals: every one preserves the original intent, and several are sharper (`Try derived state` → `Try deriving canInline instead of updating it in an effect`; `Implement repeat` → `Implement repeated keydown simulation`; `Test this` → named coverage gaps). Both upstream-gated removal triggers are back in `app/astro.config.ts`, above the explanations rather than replacing them.
2. **`combobox-select.tsx`** — the autofill rationale now sits at line 211, directly above the `useWrapElement` that renders `<select>`; the context-provider wrapper is correctly bare.
3. **`dialog.tsx:517`** — the rewrite states the real order ("scroll explicitly before focusing") and restores the reason for the re-read ("only scrolling is conditional"). Matches the code: `isFocusable` guard, `scrollIntoView` inside it, `focus({ preventScroll: true })` after.
4. **`perf.ts:1408`** — the page-load exception and the "failed attempts record nothing" safety argument are both back.
5. **`press.ts:382`** — the `defaultAllowed` return contract is restored above `pressKeyDown`.
6. **`shims.test.ts`** — all three spots addressed. I checked the new wording against the shim itself: `isHidden` returns true for `!isConnected` and for `hidden`/`display:none`/`visibility:hidden`, so both "Connected visible elements receive a synthetic 1x1 client rect" and "Disconnected elements have no client rects" are accurate for the cases asserted.
7. **`source.ts`** — solved better than I proposed: `// Import pattern classification` labels `PATTERNS`, and `// Text and module-specifier helpers` moved down to line 128 where the helpers actually start. Both headers now describe what follows them.

Both newly added permalinks resolve correctly and anchor to the right files (`#discussion_r3730034764` → `dialog.tsx`, `#discussion_r3730036563` → `perf.ts`).

### Checks run on this head

`pnpm tsc` (pass), `pnpm lint` (pass), and `pnpm test --run` over `composite-store.test.ts`, `shims.test.ts`, `styles.test.ts` — 3 files, 62 tests, all passing. No added comment line exceeds 80 characters, and the incremental diff contains no non-comment changes.

- **`composite-store.test.ts:372`** — the test-comment restoration stopped at the one test I named. The four tests below it still have bare assertions, so the file now explains its first grid test and none of the others. The falsy-id expectations are the sharpest case: `next({ activeId: "", focusLoop: true })` returning `undefined` for a single-item composite reads as a defect without the comment that marked it as the contract.

### Observation not published inline

- **Self-referential permalinks in `dialog.tsx:521` and `perf.ts:1411`** — Declined non-goal. Both now cite review comments on this PR. That is valid and the local comments are self-sufficient, but it differs from the pattern elsewhere in the codebase, where a `#discussion_r…` link points at the review of the PR that *introduced the code* (`popover.tsx` → #7009, `composite/utils.ts` → #7030). The dialog scroll block traces to #7048, which `dialog.tsx` already links a few lines above for the adjacent ownership logic. Not worth a change on its own; worth preferring the origin PR if either comment is touched again.

Nothing else in the five new commits introduced an inaccuracy. No follow-up issues registered.

### Not fully validated

Same as the previous review: browser, visual, and perf Playwright suites were not run locally and remain CI's responsibility. `pnpm build` and packed-contents verification are still not applicable — no `files`, `exports`, or packaging changes.

packages/ariakit-components/src/composite/composite-store.test.ts
diegohaz: Fixed in [`5e90b7e04`](https://github.com/ariakit/ariakit/commit/5e90b7e04352b2ee4afa5f39008b40988a0c0b2b): restored concise intent comments for the backward disabled-item loop and the two falsy-ID edge-case contracts while keeping routine neighboring narration removed.

diegohaz:
**Fixed**

- [`5e90b7e04`](https://github.com/ariakit/ariakit/commit/5e90b7e04352b2ee4afa5f39008b40988a0c0b2b) ([reply](https://github.com/ariakit/ariakit/pull/7072#discussion_r3730521367)): Restored the three non-obvious composite edge-case comments while keeping routine neighboring narration removed.

packages/ariakit-components/src/composite/composite-store.test.ts
diegohaz: Confirmed fixed in `5e90b7e04`. I re-derived each restored comment from its fixture rather than reading them alone:

- Line 345 — row `a` is `a1, a2, a3(disabled)`, so a backward loop from `a1` wraps past the disabled `a3` and lands on `a2`. "Skip the disabled item when looping backward" is exactly what makes `a2` the expected value.
- Line 373 — `createComposite([{ id: "" }])` with `focusLoop` returning `undefined` is the empty-string-id contract, now stated.
- Line 381 — item `a` has no `rowId` and item `b` has `rowId: ""`, and vertical movement crosses between them, which is the equivalence the comment names.

`pnpm lint` passes, the file's 22 tests pass, and the diff is comment-only with no line over 80 characters.

On the assertions you deliberately left bare: the standard you applied — comment what the test name does not already say — holds up, and I agree with it. The two remaining bare assertions in `loops within the active item's row on grids` are direct demonstrations of "loops within the active item's row," so the name carries them.

One asymmetry I noticed and am **not** asking you to change: line 343's forward loop from `a2` also skips the disabled `a3` to reach `a1`, so the skip is not unique to the backward case that line 345 calls out. Disposition is Declined non-goal — the current state is readable, the expected values are derivable from a fixture five lines up, and a third round over one line of test narration would be churn against a policy you have already stated clearly.

No further action needed on this thread from my side.

diegohaz (commented):
## Re-review summary

Re-review. Baseline: the head of my previous review, `40b34b8f` → current head `5e90b7e0` (1 commit, 1 file, +3/−0). No new findings.

### Candidate validated this run

- **`composite-store.test.ts:372` (thread `PRRT_kwDOBjT0Fc6XDEKL`, unresolved at review time)** — Confirmed fixed in `5e90b7e04`; replied on the thread. All three comments I named are restored and each is accurate against its fixture: the backward loop from `a1` really does skip the disabled `a3` to land on `a2`; the empty-string-id contract and the `rowId: ""`/`undefined` equivalence are both stated where the expected values depend on them.

The eight older threads remain resolved and were re-checked against this head — the earlier fixes are all still in place, including the TODO count, which stays at 27 and matches the merge base.

### Checks run on this head

`pnpm lint` (pass) and `pnpm test --run` on `composite-store.test.ts` (22 tests, pass). The incremental diff is three comment lines with no non-comment changes and no line over 80 characters, so the broader suites are unaffected since the previous review, where `pnpm tsc` and the wider focused run both passed.

### Observation recorded, no action requested

- **Line 343 in `loops within the active item's row on grids`** — Declined non-goal. The forward loop from `a2` skips the same disabled `a3` that line 345's restored comment calls out, so the disabled-skip note is asymmetric across three consecutive assertions. The values stay derivable from the fixture directly above, the "comment what the test name does not already say" standard applied here is coherent, and another round over one line of test narration would be churn. Recorded so the asymmetry is on the record, not as a request.

### Status

From my side this PR is done: every finding I raised across three reviews is fixed and independently verified against the code, and nothing in the latest commit introduced a new issue. Remaining verification belongs to CI — the browser, visual, and perf Playwright suites were never run locally in any of these reviews.

PR #7077, opened 2026-08-07
Extract shared combobox control behavior

### LINKED ISSUE #6830: Extract shared ComboboxControl behavior for Combobox and ComboboxSelect
### Motivation

Issue #6796 suggested a shared abstraction for the behavior used by both `Combobox`/`ComboboxInput` and `ComboboxSelect`. During implementation it was agreed to build `ComboboxSelect` on its own first and factor the shared part out once both components exist (see the discussion in #6796). The shared surface is now visible: store resolution, popover scoped context, typeahead composition, `aria-activedescendant` exposure, and the open-state keyboard mapping that mirrors `CompositeItem`.

This internal abstraction represents the primary `role="combobox"` control rendered by either `Combobox`/`ComboboxInput` or `ComboboxSelect`. It is distinct from `ComboboxDisclosure`, which is an auxiliary popup toggle and does not use this behavior.

The current implementation is correct without the extraction; this is a refactoring follow-up to avoid the two components drifting apart.

### Expected outcome

Shared behavior between `Combobox`/`ComboboxInput` and `ComboboxSelect` lives in one internal `ComboboxControl` abstraction (for example, `useComboboxControl` and shared options), with no behavior changes. `ComboboxDisclosure` remains separate.

diegohaz:
## Motivation

Closes #6830.

`Combobox`/`ComboboxInput` and `ComboboxSelect` both render the primary `role="combobox"` control, and both duplicate the behavior that makes it one: the combobox popup identity in the accessibility tree, and the composite widget composition that owns the combobox items. `ComboboxSelect` was built on its own first, as agreed in #6796, so the two components now describe the same control twice and can drift apart.

`ComboboxDisclosure` is not part of this. It is an auxiliary popup toggle, so it keeps using `PopoverDisclosure` directly.

## Solution

Add an internal `ComboboxControl` module at `packages/ariakit-react-components/src/combobox/__combobox-control.tsx`, with the `__` prefix that keeps a module out of the generated package exports, the same as `__hovercard-trigger.tsx` and `combobox/__utils.ts`.

It owns two pieces:

- `getComboboxControlProps(contentElement)` returns the control's identity in the accessibility tree, which is `role="combobox"` plus `aria-haspopup` resolved with a `listbox` fallback rather than the `dialog` fallback that generic dialog disclosures apply.
- `useComboboxControl` composes `useComposite` with the combobox item scrolling from `getScrollItemIntoView` and the `focusable` default the control needs to own DOM focus and expose `aria-activedescendant`.

Both components now spread the shared props and call the shared hook:

```tsx
const htmlProps = {
  ...getComboboxControlProps(contentElement),
  "aria-autocomplete": ariaAutoComplete,
  // ...
};

props = useComboboxControl<TagName>({ store, ...props });
```

The split between a plain function and a hook is deliberate. `ComboboxSelect` is also a popover disclosure, so `usePopoverDisclosure` runs between its own props and the control. `useButton` always writes a `role` key and `useDialogDisclosure` defaults `aria-haspopup` to the `dialog` popup role, and in Ariakit's layered props each hook does `props = { ...defaults, ...props }`, so an earlier layer wins. A hook-level ARIA default applied last would therefore be silently overridden for `ComboboxSelect` while working for `Combobox`. Returning those props from a plain function lets each component apply them ahead of every intervening hook, which keeps one definition and one behavior for both.

Store resolution stays per-component on purpose: `Combobox` falls back to the scoped context so a `ComboboxInput` nested inside `ComboboxPopover` still finds its store, which `ComboboxSelect` does not need as the trigger. Typeahead, the `aria-activedescendant` gating on `mounted`, and the open-state keyboard proxying remain in `ComboboxSelect`, because moving any of them into the shared control would change `Combobox` behavior.

## Changes

No behavior changes, no public API changes, and no changeset. `ComboboxOptions` still extends `CompositeOptions`, unchanged from `main`.

## Verification

- `pnpm tsc` and `pnpm lint` pass.
- `pnpm test` passes (254 files, 1708 tests) and `pnpm test-react18` passes (187 files, 974 tests).
- Playwright Chrome passes for `src/sandbox/combobox` and `src/examples/combobox` (220 tests), and the full app Chrome suite passes (732 passed, 1 skipped; 5 unrelated tooltip and tab-panel tests were flaky and green on retry).
- Generated public reference metadata was dumped from the app `jsdoc-loader` for all 27 combobox references on `main` and on this branch, and the two are byte-identical.

## Acknowledgments

Thanks to [@georgekaran](https://github.com/georgekaran) for proposing the sequencing in #6796 that built `ComboboxSelect` first and deferred this extraction, and for implementing the `ComboboxSelect` behavior this change factors out.

### DIFF
--- a/packages/ariakit-react-components/src/combobox/__combobox-control.tsx
+++ b/packages/ariakit-react-components/src/combobox/__combobox-control.tsx
@@ -1,0 +1,55 @@
+import { createHook } from "@ariakit/react-utils";
+import { getPopupRole } from "@ariakit/utils";
+import type { ElementType } from "react";
+import type { CompositeOptions } from "../composite/composite.tsx";
+import { useComposite } from "../composite/composite.tsx";
+import { getScrollItemIntoView } from "./__utils.ts";
+import type { ComboboxStore } from "./combobox-store.ts";
+
+const TagName = "div" satisfies ElementType;
+type TagName = typeof TagName;
+
+/**
+ * Returns the props that identify the combobox control in the accessibility
+ * tree. The popup role falls back to `listbox`, unlike the `dialog` fallback
+ * that generic dialog disclosures apply.
+ */
+export function getComboboxControlProps(contentElement?: HTMLElement | null) {
+  return {
+    role: "combobox",
+    "aria-haspopup": getPopupRole(contentElement, "listbox"),
+  } as const;
+}
+
+/**
+ * Returns props that turn an element into the composite widget for the
+ * combobox items. This is the primary `role="combobox"` element rendered by
+ * either
+ * [`Combobox`](https://ariakit.com/reference/combobox)/[`ComboboxInput`](https://ariakit.com/reference/combobox-input)
+ * or [`ComboboxSelect`](https://ariakit.com/reference/combobox-select). It's
+ * distinct from
+ * [`ComboboxDisclosure`](https://ariakit.com/reference/combobox-disclosure),
+ * which is an auxiliary popup toggle.
+ * @see https://ariakit.com/components/combobox
+ */
+export const useComboboxControl = createHook<TagName, ComboboxControlOptions>(
+  function useComboboxControl({ store, focusable = true, ...props }) {
+    return useComposite<TagName>({
+      store,
+      unstable_scrollIntoView: getScrollItemIntoView(store),
+      focusable,
+      ...props,
+    });
+  },
+);
+
+export interface ComboboxControlOptions<
+  T extends ElementType = TagName,
+> extends CompositeOptions<T> {
+  /**
+   * Object returned by the
+   * [`useComboboxStore`](https://ariakit.com/reference/use-combobox-store)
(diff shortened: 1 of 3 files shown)


diegohaz:
Closing this without merging.

Auditing the two hooks to size the extraction showed that most of the surface this was meant to unify is select-specific for structural reasons, not because `ComboboxSelect` happened to be built first. The details are in https://github.com/ariakit/ariakit/issues/6830.

What remains genuinely shared and movable without changing behavior is `role="combobox"`, the `listbox` fallback for `aria-haspopup`, the `focusable` default, and passing `unstable_scrollIntoView` to `useComposite`. The only non-trivial piece of that, `getScrollItemIntoView`, is already shared through `combobox/__utils.ts`, so a dedicated module, hook, and options interface adds more indirection than it removes.

The branch is kept for reference in case the extraction is worth revisiting alongside behavior changes.

PR #7067, opened 2026-08-04
Center the selected item when a select-shaped popup opens

diegohaz:
## Motivation

When a select-shaped popup opens, Ariakit aligns the selected item to the nearest edge of a long list. Centering it on the initial open shows the surrounding values, while later navigation should retain the platform's nearest-edge behavior.

Fixes #7011.

## Solution

Center the initially selected `ComboboxSelect` item when the popup has one vertical scrollport, without moving the document or ancestors outside the popup. Opening through click or keyboard, reopening, real and virtual focus, remounted popups, and asynchronously replaced options all use the same presentation path.

Later keyboard and typeahead navigation continues using nearest-edge scrolling. Layouts with no internal scrollport, multiple internal scrollports, or a non-horizontal writing mode fall back to native nearest-edge scrolling so the item remains visible.

The implementation passes a private item-scrolling callback through the Composite components. `ComboboxSelect` records one movement baseline per open cycle, so item callbacks, including callbacks from virtualized items, can distinguish the opening presentation from later navigation without subscribing or re-rendering every item.

## Workaround

The [temporary observer-based workaround](https://github.com/ariakit/ariakit/issues/7011#issuecomment-5181943654) can center an item when the popup element itself is the vertical scrollport. It is narrower than the library fix and depends on the private, version-sensitive `data-placing` marker.

## Preview

The initially selected `16.8.6` item opens centered with surrounding options visible:

16.8.6 centered in the opened select-shaped popup

## Verification

- 99 affected browser tests passed in Chrome, Firefox, and Safari.
- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build`
- Fresh three-round local Chrome performance comparison against the exact base reported no significant changes.
- Independent pre-commit review of the exact staged snapshot.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue with one ComboboxSelect-owned opening baseline</summary>

**Assessment**

The third gate cycle raised a low-severity, possible alignment error for a selected virtualized item mounted after navigation. Corrected `ComboboxRenderer` reproduction showed that the pre-fix code only ran the root nearest-edge callback and already kept that item at the edge, so the reported runtime mechanism was rejected. The exercise still exposed avoidable maintenance and runtime complexity: every item owned an `open` listener solely to read the same movement baseline.

[...]

PR #6877, opened 2026-07-27
Migrate example tests to app sandbox

diegohaz:
## Motivation

The Main workflow still runs browser jobs for the legacy root `examples` folder. This duplicates coverage that belongs to the app sandbox and spreads example-test ownership across two systems.

## Solution

- Audit all 154 root example directories and their existing browser, mobile, and Vitest behavior.
- Keep the 56 README-backed examples and 21 additional examples referenced by root guides, while deleting the 77 unreferenced sandbox examples.
- Remove test files from every retained root example after porting or confirming equivalent app coverage.
- Retain the complete legacy `Select` compatibility matrix in `app/src/sandbox/select-legacy`.
- Add focused, top-level `combobox-select-*` sandboxes for modern root-example behavior.
- Keep behavior-required styling in inline Tailwind classes; this migration adds no stylesheet files.
- Move Next.js fixture source to the root Next app while keeping its metadata and tests app-owned:

  ```text
  nextjs/app/<sandbox>/...
  app/src/sandbox/<sandbox>/preview.json
  app/src/sandbox/<sandbox>/test-browser.ts
  ```

- Give app Playwright, test utility, and perf files a Node TypeScript project, and keep React-specific source out of the Workerd project through `.react.ts`/`.react.tsx` ownership.
- Limit framework test discovery to app examples and sandboxes.
- Remove the root-example browser jobs from Main CI and run mobile behavior coverage through the App workflow.

## Verification

- `pnpm lint`
- `pnpm lint-css`
- `CI=1 pnpm tsc-clean && CI=1 pnpm tsc`
- `pnpm test` — 61 files, 679 tests
- `CI=1 pnpm test-react --run` — 184 files, 943 tests
- `CI=1 pnpm test-react18` — 184 files, 852 tests
- `pnpm test-solid` — 6 files, 8 tests
- `pnpm --filter ./app check`
- `CI=1 pnpm build-app-lite`
- `pnpm build-website-lite`
- `packages/ariakit-scripts/src/ci.test.ts` — 16 tests
- Focused migrated-owner matrix — 159 tests across Chromium, Firefox, WebKit, iOS, and Android
- Corrected ComboboxSelect and Next.js owners — 11/11 on Chromium, Firefox, and WebKit
- Restored `select-legacy` integration suite — 29/29
- Consolidated `select-legacy` Playwright coverage — 1/1 on Chromium
- Select/Combobox tab scroll restoration — 6/6 across Chromium, Firefox, and WebKit without retries
[...]

PR #7079, opened 2026-08-07
Define focus and presentation for default-open ComboboxSelect popups

### LINKED ISSUE #7068: Define focus and presentation for default-open ComboboxSelect popups
### Problem

A default-open [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) popup can mount while focus remains on the document body. In that state, the popup does not run the interactive focus-presentation path, so a selected item below the fold can remain outside the visible part of the list and keyboard navigation is not available from the control.

Issue #7011 originally included default-open popups in its expected coverage. Pull request #7067 centers selected items for interactive click and keyboard opening, but deliberately removes a generic `Composite` effect that centered default-open items without resolving focus.

### Expected outcome

Define and implement Combobox-specific focus and presentation behavior for default-open and other focusless programmatic openings. The solution should make the selected item visible without introducing popup-only state into generic `Composite` or relying on commit-order retries.

Coverage should establish the intended focus owner, selected-item visibility, asynchronously rendered items, and abandonment when focus moves elsewhere.

### Context

On the current base, a default-open select with a far selected value leaves `document.activeElement` on `BODY` and the list at `scrollTop: 0`. An attempted presentation-only path centered the item while focus still remained on `BODY`, creating a partial state where the user could see the item but could not navigate the list.

diegohaz:
## Motivation

A select-shaped [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) popup that opens without passing through the select never reaches the focus-driven presentation path. A selected value below the fold stays outside the visible part of the list, and the control offers no keyboard navigation.

```jsx
// The list opens parked on Artichoke, and arrow keys do nothing.
<Ariakit.ComboboxProvider defaultOpen defaultSelectedValue="Onion">
  <Ariakit.ComboboxSelectLabel>Vegetable</Ariakit.ComboboxSelectLabel>
  <Ariakit.ComboboxSelect />
  <Ariakit.ComboboxPopover>{/* Onion is 26 options down */}</Ariakit.ComboboxPopover>
</Ariakit.ComboboxProvider>
```

[`useComboboxPopover`](https://github.com/ariakit/ariakit/blob/9c4143ac66b5a9fa0e619aac1d77d2ca1b6bf566/packages/ariakit-react-components/src/combobox/combobox-popover.tsx#L217-L230) derived `autoFocusOnShow` as `hasSelect && (!!inputElement || selectOwnsFocus)`, so a select-only popup took focus only when the select already held it. A `defaultOpen` mount or a `show()` call therefore left focus where it was and skipped the presentation.

The asymmetry is the bug: the `!!inputElement` term already made a filterable `ComboboxSelect` autofocus on every open, regardless of who owned focus. #7067 centered the selected item for interactive opens and deliberately left this case here.

Fixes #7068.

## Solution

Showing a popup is a request to interact with it, so a select-shaped popup now takes focus on every open, including a default or programmatic one, even when another element owns focus. That makes the select-only popup behave like the input-backed one and restores both keyboard navigation and the selected-item presentation.

```diff
-      autoFocusOnShow: hasSelect && (!!inputElement || selectOwnsFocus),
+      autoFocusOnShow: hasSelect,
```

The fix removes machinery rather than adding it. `selectOwnsFocus` was the hook's only render-time `getActiveElement` read, so the derivation is now a pure function of store state and no longer toggles on unrelated re-renders. Nothing was added to generic `Composite`, and there are no commit-order retries: the existing focus handoff already presents the selected item, including one that registers after the open.

`ComboboxPopoverOptions` now documents the resulting computed default. The reference page previously inherited Dialog's unconditional `= true`, which is wrong for a popover with no `ComboboxSelect`.

A userland workaround of the same shape existed in `app/src/sandbox/combobox-select-menu-filters`, added with that fixture in #6877. It is now redundant and removed, so its existing focus assertion covers the new default on a fixture that predates this work.

## Coverage

The reproduction landed first, before any library change, so CI recorded the red state.

`app/src/sandbox/combobox-select-default-open` grows a scrollable list with a far selected value and asserts the contract: the select owns focus, the selected item is centered in the listbox, and arrow keys navigate from the control. The tests #6832 added for the previous "does not take focus" invariant are rewritten, since this redefines exactly that invariant.

`app/src/sandbox/combobox-select-programmatic-open` is new and covers openings that are not `defaultOpen`: a global `F2` shortcut shows the popup while focus is still wherever the user left it. It exercises options that register after the open, abandonment when focus moves elsewhere while a presentation is still pending, and an open that happens while another element owns focus.

The two scenarios need separate pages, because a `defaultOpen` popup already owns focus by the time a same-page shortcut could fire. The shortcut rather than a focus-refusing button opens the second fixture because Safari falls back to the last mousedown target for the disclosure element, which would give that open an opener in one engine and not the others.

`expectVerticallyCentered` moves to `app/src/test-utils/scroll.ts` rather than being copied a third time, and the option list is shared through `combobox-select-default-open/vegetables.ts`.

## Workaround

For anyone on a released version, opt the popup into the focus it already takes when the select itself opens it:

```jsx
<Ariakit.ComboboxProvider defaultOpen defaultSelectedValue="Watermelon">
  <Ariakit.ComboboxSelect />
  {/* TODO: Remove once a programmatic open takes focus on its own.
      https://github.com/ariakit/ariakit/issues/7068 */}
  <Ariakit.ComboboxPopover autoFocusOnShow>
    {/* items */}
  </Ariakit.ComboboxPopover>
</Ariakit.ComboboxProvider>
```

Assumptions and limitations:

- Apply it only to a popup that renders a `ComboboxSelect`. On a plain `Combobox` popover with no select it produces behavior the fix never will: with virtual focus, an input to popover and back focus round trip on every open; with `virtualFocus={false}`, focus lands on an item and typing stops working.
- It is a no-op for a filterable `ComboboxSelect`, whose popup already autofocuses.
- The popup takes focus on every open, including controlled and data-driven ones. That is the intended contract, not a side effect.
- Each open pays one tabbable scan of the popup, because any truthy `autoFocusOnShow` defeats the dialog's early-out. The fix pays the same cost for the same composition.

## Verification

- Red on `9c4143ac6`: 7 browser tests and 6 happy-dom tests failing. Green with the fix.
- Full happy-dom suite: 255 files, 1713 tests.
- Playwright `combobox select`: 241 in Chrome, 228 in Firefox, 179 in Safari.
- Both reference generators inspected for the new prop documentation, plus `jsdoc-loader` and `reference-utils` tests and `pnpm build-website-lite`.
- `pnpm lint-fix`, `pnpm tsc`.

Two of the tests do not red-prove their distinctive content, and cannot before the fix exists: the abandonment pair fails at its premise, because pre-fix there is no presentation to abandon and deleting the premise would make them pass vacuously.

## Follow-ups

- A programmatically opened `ComboboxSelect` announces `aria-expanded="false"` while its listbox is open, because `Dialog` reassigns the disclosure element to whoever owned focus at open time. Pre-existing on `main` and not fixed here, since it is a disclosure-ownership contract rather than a focus one: #7080.
- Cover `defaultOpen` with a competing focus owner at mount. Two `defaultOpen` popovers on one page would force relabeling the existing `q.listbox()` queries, so it is better added on its own.
- `examples/select-combobox-tab/select.tsx` carries the same now-redundant `autoFocusOnShow`. Left alone because it belongs to the legacy root Playwright suite and is behavior-identical either way.


packages/ariakit-react-components/src/combobox/combobox-popover.tsx
```diff
@@ -342,6 +338,19 @@ export interface ComboboxPopoverOptions<T extends ElementType = TagName>
...
+   * [`initialFocus`](https://ariakit.com/reference/combobox-popover#initialfocus)
+   * prop can be used to set a different element to receive focus.
+   *
+   * Defaults to `true` when a
+   * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) is
+   * rendered, and `false` otherwise, since a standalone
+   * [`ComboboxInput`](https://ariakit.com/reference/combobox-input) keeps focus
+   * while its popup is open.
+   */
+  autoFocusOnShow?: PopoverOptions<T>["autoFocusOnShow"];
```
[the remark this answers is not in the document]
diegohaz: Confirmed against the current head.

The sentence is inaccurate for exactly the composition this JSDoc documents as the `true` default. In the canonical select-shaped popup (a virtual-focus listbox with no separately tabbable content), the popup hands DOM focus to the `ComboboxSelect`, which is outside the popover. This PR's own coverage asserts that end state: `app/src/sandbox/combobox-select-default-open/test-browser.ts` carries the comment "Virtual focus keeps DOM focus on the select, not on the active item." and then asserts `q.combobox("Vegetable")` is focused. I also measured it directly on a throwaway sandbox of the same shape: after a `defaultOpen` mount, `document.activeElement` has `role="combobox"`, not an element inside the popover.

This is worth correcting rather than tolerating, because publishing an accurate computed default on the reference page is the whole reason the prop is redeclared here. I verified that both generators ship this text: it appears in the app's `.astro/data-store.json`, and calling `getReferences` from `website/build-pages/reference-utils.js` on `combobox-popover.tsx` returns the prop with this description and `defaultValue: null`.

Note that "inside the popover" is not universally wrong. With `virtualFocus={false}`, or when the popup contains a tabbable element, focus does land inside. So the fix is to stop promising a target inside the popover and name where focus actually goes in the default case, not to swap one absolute for another.

The second paragraph is accurate and should stay as is.

```suggestion
  /**
   * Determines whether the popup takes focus when it opens. With virtual focus,
   * which is the default for a select-shaped popup, DOM focus lands on the
   * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) element
   * rather than inside the popover, while the selected item becomes the active
   * item. The
   * [`initialFocus`](https://ariakit.com/reference/combobox-popover#initialfocus)
   * prop can be used to set a specific element to receive focus.
   *
   * Defaults to `true` when a
   * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) is
   * rendered, and `false` otherwise, since a standalone
   * [`ComboboxInput`](https://ariakit.com/reference/combobox-input) keeps focus
   * while its popup is open.
   */
  autoFocusOnShow?: PopoverOptions<T>["autoFocusOnShow"];
```

**Assessment**

- Severity: Low (estimate, 90% confidence), because the copy has no runtime effect, but it teaches the wrong focus owner on a public reference page.
- Priority: P2 (estimate, 85% confidence), because this PR introduces the sentence and it ships alongside the behavior it describes.
- Breakage likelihood: Likely (verified), because I reproduced the mismatch on the documented default and confirmed both reference generators emit this exact text.
- Fix effort: Easy (verified), because the change is confined to one JSDoc block and no test depends on the wording.

.changeset/300-select-popup-programmatic-focus.md
```diff
@@ -0,0 +1,21 @@
...
+  <ComboboxSelect />
+  <ComboboxPopover>
+    <ComboboxItem value="Apple" />
+    {/* Watermelon becomes the active item, ready for the arrow keys. */}
+    <ComboboxItem value="Watermelon" />
+  </ComboboxPopover>
+</ComboboxProvider>
+```
+
+To keep a popup from taking focus, pass [`autoFocusOnShow={false}`](https://ariakit.com/reference/combobox-popover#autofocusonshow).
```
diegohaz: A popup below the fold now scrolls the page when it opens, and neither this changeset nor the PR description says so.

I measured it on a throwaway sandbox that is the `combobox-select-default-open` fixture placed after a 1500px spacer, run in Chromium against the app dev server from this branch:

- On `9c4143ac6`: `window.scrollY` is `0` after mount, and nothing is focused.
- On `31d1191e7`: `window.scrollY` is `811` after mount, and the `ComboboxSelect` is focused.

This follows from the contract rather than contradicting it. Focus moves to an off-screen widget, so it gets scrolled into view the same way `autofocus` does, and the alternative (keyboard focus on an invisible open listbox) would be worse. So this is a disclosure gap, not a defect in the fix.

The gap is worth closing because "takes focus on every open" reads as a DOM-focus change. A consumer whose page renders a `defaultOpen` picker far down, or opens one from a global shortcut while the user is reading elsewhere, gets a page that arrives already scrolled, and nothing in the entry points at `autoFocusOnShow={false}` as the lever for that. The repo already treats page movement on open as its own contract, with a dedicated fixture and browser suite in `app/src/sandbox/combobox-select-open-page-scroll` from #6985, so new movement is better stated than discovered.

One sentence on the existing opt-out line covers it:

```suggestion
To keep a popup from taking focus, pass [`autoFocusOnShow={false}`](https://ariakit.com/reference/combobox-popover#autofocusonshow). This also applies to the scroll: since the popup takes focus, one that sits below the fold is scrolled into view when it opens.
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because the only user-visible effect is the page position, and it lands on the widget that just opened rather than somewhere arbitrary.
- Priority: P2 (estimate, 80% confidence), because the behavior is intended and the missing piece is one line of changelog copy, not a code change.
- Breakage likelihood: Possible (verified), because it needs a `defaultOpen` or programmatically opened `ComboboxSelect` below the fold, which is uncommon but ordinary, and I measured the movement on exactly that shape.
- Fix effort: Easy (verified), because it is one sentence in a changeset that already documents the opt-out.

diegohaz (commented):
Initial review. Baseline is the merge base [`9c4143ac6`](https://github.com/ariakit/ariakit/commit/9c4143ac66b5a9fa0e619aac1d77d2ca1b6bf566) to head [`31d1191e7`](https://github.com/ariakit/ariakit/commit/31d1191e729c7e5a81988a989b4f0889e823e2fa), all 13 changed files, then the surrounding context: `combobox-popover.tsx`, `combobox-list.tsx`, `combobox-item.tsx`, `combobox/__utils.ts`, `dialog.tsx`, `popover.tsx`, `menu.tsx`, `select-popover.tsx`, and the ComboboxSelect sandboxes. Reviewed in a dedicated worktree checked out at the head commit.

## Verdict

The direction holds up, and the diff is smaller than the behavior change. Dropping `selectOwnsFocus` removes the hook's only render-time `getActiveElement` read, so `autoFocusOnShow` becomes a pure function of store state instead of something that can flip on an unrelated re-render. That claim in the description checks out.

One argument the description does not make, which I think is the strongest one for the approach: `useSelectPopover` never routed through `useComboboxPopover`, and it has always inherited Popover's `autoFocusOnShow = true`. So `autoFocusOnShow: hasSelect` returns `ComboboxSelect` to parity with the deprecated `Select` it replaces, rather than inventing a contract. I confirmed the legacy path is untouched by this change.

## What I ran

- `pnpm test`: 255 files, 1713 tests, all passing. Matches the PR's number.
- `pnpm test-react18 app/src/sandbox/combobox-select`: 28 files, 138 tests, passing. Not in the PR's verification list, and worth having, since focus behavior can differ under the React 18 dependency set.
- Red-state proof: reverted `combobox-popover.tsx` to the base and re-ran. All 6 new and rewritten happy-dom tests fail, plus 2 in `combobox-select-menu-filters`. That second pair is the useful part, because it shows the removed userland workaround was load-bearing and is genuinely replaced by the library default rather than being dead weight. I also confirmed the "abandons the presentation" pair fails at its premise on `app/src/sandbox/combobox-select-programmatic-open/test.ts:50` rather than at its distinctive assertion, which matches the PR's own disclosure. Restored the file and verified a clean tree afterwards.
- Playwright against an app dev server started from this same worktree, with the server process working directory verified before running: `chrome` 40 passed across `combobox-select-default-open`, `combobox-select-programmatic-open`, `combobox-select-menu-filters`, and `combobox-select-selected-value-presentation`; `safari` 7 passed; `firefox` 7 passed.
- Reference metadata, checked separately in both consumers rather than trusting one. The new `autoFocusOnShow` description appears in the app's `.astro/data-store.json`, and calling `getReferences` from `website/build-pages/reference-utils.js` on `combobox-popover.tsx` returns the prop with the new description and `defaultValue: null`, so the inherited `= true` is gone from the legacy output too.
- Changeset checks: `combobox-popover`, `combobox-provider`, `combobox-select`, and `combobox-input` all return 200, and `#autofocusonshow` and `#defaultopen` resolve on their pages. Every API named in the entry is exported from `@ariakit/react`, so the description matches the public surface its frontmatter versions. No contributor credit is due, since #7068 was authored by `diegohaz` and the only other participants are bots.

## Unresolved thread validated in this run

- [`#discussion_r3737774561`](https://github.com/ariakit/ariakit/pull/7079#discussion_r3737774561) (CodeRabbit, `combobox-popover.tsx` 341-353), on the new `autoFocusOnShow` JSDoc. Outcome: confirmed. Disposition: Address now. Replied at [`#discussion_r3738126089`](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738126089) with evidence and a suggestion, including the nuance the original comment missed: "inside the popover" is correct under `virtualFocus={false}` or when the popup holds a tabbable element, so the fix is to name where focus goes in the default case rather than swap one absolute for another.

## Published inline

One comment, on `.changeset/300-select-popup-programmatic-focus.md`: a popup below the fold now scrolls the page when it opens, which I measured as `window.scrollY` going from `0` on the base to `811` on this head, and which no user-facing text mentions.

## Discovered here, not published

1. **Tabbable content in a select popup diverting the autofocus. Disproven.** I suspected that with `initialFocus` resolving to `undefined` for a select-only popup, the dialog's `getFirstTabbableIn` fallback would focus a button rendered in the popup instead of handing off to the select, which would make "restores both keyboard navigation and the selected-item presentation" conditional on popup content. It does not. On a throwaway fixture with a `Clear` button before the list, focus still landed on the `ComboboxSelect` and `ArrowDown` still moved the active item, because the dialog's `[data-autofocus=true]` lookup runs before the tabbable fallback and `combobox-item.tsx:262` marks the selected item with that attribute. Recording it because the mechanism is not obvious from the diff and the next reader may have the same doubt. Invalid, so no assessment.
2. **Comment length at `combobox-popover.tsx:215-221`.** Five prose lines plus a link for a one-line assignment, against the "normally no more than three short lines" guidance in `ariakit-general-code-style`. Both halves carry a real constraint, the contract and why this must stay a boolean rather than a callback, and shortening would most likely drop the second. Declined non-goal.
3. **`examples/select-combobox-tab/select.tsx:76` still carries the now-redundant `autoFocusOnShow`** while the equivalent sandbox workaround was removed in this PR. The PR discloses the asymmetry and it is behavior-identical either way. Declined non-goal.
4. **Two of the three Follow-ups bullets have no durable record.** #7080 is registered. "Cover `defaultOpen` with a competing focus owner at mount" and the legacy example cleanup live only in the PR body, and `ariakit-general-workflow` treats a PR note as insufficient for a Registered follow-up. The coverage bullet is the one that reads like confirmed work the project intends to pursue. I did not open an issue, since these are your own scope calls on your own PR; happy to file the coverage one if you want it tracked.

## Not fully validated

A `disabled` `ComboboxSelect` with `defaultOpen`. `combobox-list.tsx:78` bails out of the focus redirect when the composite element is not focusable, which would leave DOM focus on the listbox itself, the state its own comment warns about, and a focusless default open could not reach that before. I judged the composition too exotic to justify a fixture and did not test it, so this is a note rather than a finding.

packages/ariakit-react-components/src/combobox/combobox-popover.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`16965d9`](https://github.com/ariakit/ariakit/commit/16965d9991f94d500ee31e18ae12218836deeb7d): the description no longer claims focus lands inside the popover.

I went further than the suggested wording, because that replacement is also inaccurate. It says DOM focus lands on the `ComboboxSelect` under virtual focus, but `initialFocus: hasSelect ? inputElement : undefined` means a popup rendering a `ComboboxInput` focuses that input, inside the popover, with virtual focus fully on. `combobox-select-open-page-scroll/test-browser.ts` asserts exactly that. The same is true when the popup holds any tabbable element and no item carries `data-autofocus`, since `getFirstTabbableIn` wins and the list's self-target redirect never runs.

So the placement has at least three outcomes depending on the popup's contents and `virtualFocus`, and naming any one of them here produces another false absolute. The prop governs whether the popup takes focus; `initialFocus` governs where. The description now says that and points at both.

The second paragraph is unchanged, as you noted. I also dropped "while the selected item becomes the active item": that resolution lives in the store, keyed on `["items", "mounted", "open", "selectedValue", "selectElement"]` with no focus input, so it happens under `autoFocusOnShow={false}` too and documenting it here suggests opting out gives it up.

.changeset/300-select-popup-programmatic-focus.md
diegohaz: Fixed in [`6501e42`](https://github.com/ariakit/ariakit/commit/6501e42dfb685c5a31c05ed0475ad58522b29c33): the entry now states the scroll before the opt-out that avoids it. Added targeted coverage in [`1b06e01`](https://github.com/ariakit/ariakit/commit/1b06e01ab6efd49716571f822ac3d19d2ad1e7bb).

I reproduced your measurement independently before changing anything: with the `combobox-select-default-open` fixture behind a 1500px block spacer, `window.scrollY` is `811` after mount on this branch with the select focused, versus `0` on `9c4143ac6`. Only the document and the popup's own list scrolled. Agreed that this is disclosure rather than a defect, and that it does not conflict with the `combobox-select-open-page-scroll` contract, whose tests all scroll the select into view before opening.

I wrote the sentence the other way around, since the escape hatch read oddly before the behavior it escapes, and used "outside the viewport" to match the existing changelog phrasing.

I also added the coverage rather than leaving this documentation-only. My initial reasoning against it was wrong: `combobox-select-programmatic-open` is not `defaultOpen`, so nothing competes for focus there, and `expectVerticallyCentered` compares two viewport-relative boxes to each other, so page scroll cannot perturb it. The new browser test starts the picker outside the viewport and asserts the page moves when `F2` opens it. Red-checked by narrowing the fix to `autoFocusOnShow: hasSelect && !!inputElement`, where it fails on `toBeInViewport()` with the popup still `aria-expanded="true"`, so it fails on the scroll rather than on a premise. This matters because the scroll rests on two independent focus call sites; a later `preventScroll` could remove it while every focus assertion stayed green and this changelog sentence became false.

diegohaz:
**Fixed**

- [`16965d9`](https://github.com/ariakit/ariakit/commit/16965d9991f94d500ee31e18ae12218836deeb7d) ([reply](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738422420)): the `autoFocusOnShow` description no longer claims focus lands inside the popover. I dropped the placement claim entirely instead of applying the suggested wording, because that replacement is false too: a popup rendering a `ComboboxInput` focuses that input inside the popover under virtual focus, as `combobox-select-open-page-scroll/test-browser.ts` asserts. The prop governs whether the popup takes focus; `initialFocus` governs where.
- [`6501e42`](https://github.com/ariakit/ariakit/commit/6501e42dfb685c5a31c05ed0475ad58522b29c33) ([reply](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738424420)): the changeset now discloses that a popup opening outside the viewport is scrolled into view, stated before the opt-out that avoids it. I reproduced the reported `scrollY` of `811` independently and agree it is disclosure rather than a defect.
- [`1b06e01`](https://github.com/ariakit/ariakit/commit/1b06e01ab6efd49716571f822ac3d19d2ad1e7bb) ([reply](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738424420)): added browser coverage for that scroll in `combobox-select-programmatic-open`, red-checked so it fails on the scroll assertion rather than on a premise. The behavior rests on two independent focus call sites, so a later `preventScroll` could have removed it while every focus assertion stayed green.

Verified on the pushed head: `pnpm lint-fix`, `pnpm tsc`, both reference-extractor suites, and Playwright `combobox select` at 242 in Chrome plus the changed sandbox at 5/5 in Firefox and Safari.

app/src/sandbox/combobox-select-programmatic-open/index.react.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const loadRemainingOptions = () => setOptions(vegetables);
+
+  return (
+    // A block wrapper, because the preview lays its children out in a row and a
+    // spacer would otherwise push the picker sideways instead of below the fold.
+    <div>
+      {/* Starts the picker outside the viewport, so the scroll that taking
+      focus performs is observable. */}
+      <div style={{ height: 1500 }} />
+      <p>Press F2 to open the vegetable picker.</p>
```
diegohaz: The spacer sits above the only line that tells a person how to drive this sandbox, so the preview now opens as a blank screen.

Everything in the fixture, including `Press F2 to open the vegetable picker.`, is 1500px down. Opening `/react/previews/combobox-select-programmatic-open/` by hand gives you an empty viewport with no affordance, and the one shortcut that does anything here is not discoverable from the page.

Moving the paragraph above the spacer fixes that without weakening the new test, since the select stays 1500px below the fold either way. I ran it: with the reorder applied locally, all 5 tests in this file pass on `chrome`, `safari`, and `firefox`, including `scrolls the picker into view when it opens outside the viewport` and its `window.scrollY` and `not.toBeInViewport()` preconditions.

```suggestion
      <p>Press F2 to open the vegetable picker.</p>
      {/* Starts the picker outside the viewport, so the scroll that taking
      focus performs is observable. */}
      <div style={{ height: 1500 }} />
```

For what it is worth, I checked the obvious second-order worry and it does not apply: `app/src/tests/previews-browser.ts` screenshots with `fullPage` defaulting to `false` in `app/src/test-utils/visual.ts`, which would make a blank baseline, but `getPreviewIndexPaths` in `app/src/lib/preview-index.ts` filters sandbox previews out of that sweep. So this is purely about using the sandbox by hand.

**Assessment**

- Severity: Low (verified), because it affects only how easy the sandbox is to drive manually, and no test or shipped behavior depends on the order.
- Priority: P2 (estimate, 85% confidence), because the fixture is new in this PR and the fix is free now, but nothing breaks if it waits.
- Breakage likelihood: Likely (verified), because every manual visit to this preview lands on the blank region.
- Fix effort: Easy (verified), because I applied the reorder and ran the full file on all three desktop engines.

diegohaz (commented):
Re-review. Incremental baseline is [`31d1191e7`](https://github.com/ariakit/ariakit/commit/31d1191e729c7e5a81988a989b4f0889e823e2fa), the head recorded in my previous review marker, to the current head [`1b06e01ab`](https://github.com/ariakit/ariakit/commit/1b06e01ab6efd49716571f822ac3d19d2ad1e7bb). Three new commits, four files. I reviewed those, then re-checked the surrounding fixture and preview machinery they touch.

## Both prior findings are addressed

- [`#discussion_r3737774561`](https://github.com/ariakit/ariakit/pull/7079#discussion_r3737774561) (JSDoc focus target), resolved by [`16965d9`](https://github.com/ariakit/ariakit/commit/16965d9991f94d500ee31e18ae12218836deeb7d). Verified fixed at the current head.
- [`#discussion_r3738128075`](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738128075) (undisclosed opening scroll), resolved by [`6501e42`](https://github.com/ariakit/ariakit/commit/6501e42dfb685c5a31c05ed0475ad58522b29c33) and covered by [`1b06e01`](https://github.com/ariakit/ariakit/commit/1b06e01ab6efd49716571f822ac3d19d2ad1e7bb). Verified fixed at the current head.

On the first one, your pushback was right and my suggested replacement was wrong. Saying DOM focus lands on the `ComboboxSelect` under virtual focus does not hold for a filterable select, because `initialFocus: hasSelect ? inputElement : undefined` sends focus to the `ComboboxInput` inside the popover, which is exactly what `combobox-select-open-page-scroll/test-browser.ts` asserts. The shipped wording, which says the destination depends on contents and `virtualFocus` and points at `initialFocus`, is accurate where both my version and the original were not. Dropping the selected-item clause was also correct, since item resolution does not depend on `autoFocusOnShow`.

## Verification at this head

- Red-state proof for the new test: reverted `combobox-popover.tsx` to `9c4143ac6` and ran `scrolls the picker into view when it opens outside the viewport` in Chromium. It fails at `toBeInViewport()` on line 49, with the locator resolving to the select carrying `aria-expanded="true"`. So the popup opens and the page simply does not move, which is the intended failure rather than a premise failure. Restored and confirmed a clean tree.
- Playwright at this head for `combobox-select-programmatic-open` and `combobox-select-default-open`: 8 passed on `chrome`, 8 on `safari`, 8 on `firefox`. The four pre-existing tests in the modified fixture still pass with the spacer in place, so adding it did not disturb the invariants they rely on. Server started from this worktree with its working directory verified.
- `pnpm test`: 255 files, 1713 tests, all passing. `pnpm tsc` and `pnpm lint`: clean.
- Reference metadata in both consumers again, since the JSDoc changed: the new text is in the app's `.astro/data-store.json`, and `getReferences` from `website/build-pages/reference-utils.js` returns the prop with the new description and `defaultValue: null`.
- The new `#virtualfocus` anchor resolves on the deployed `combobox-provider` page, so the added link is not a dead fragment.

The changeset sentence is accurate, including the implicit claim that the opt-out removes the scroll too. With no focus there is no document scroll and no list presentation, since both hang off the focus handoff.

Putting the scroll coverage in `combobox-select-programmatic-open` rather than `combobox-select-default-open` was the right call, and worth noting because it is not the obvious one. A `defaultOpen` fixture has already scrolled by the time the test starts, so there is no way to assert the `scrollY === 0` and `not.toBeInViewport()` preconditions that make this test meaningful.

## Published inline

One comment, on `app/src/sandbox/combobox-select-programmatic-open/index.react.tsx`: the spacer sits above the `Press F2` instruction, so the preview opens blank with no affordance. Reordering keeps the test's preconditions intact, which I verified on all three desktop engines.

## Investigated and dropped

I expected the spacer to also blank out this sandbox's visual baseline, since `previews-browser.ts` screenshots with `fullPage` defaulting to `false` in `app/src/test-utils/visual.ts`. It does not, because `getPreviewIndexPaths` in `app/src/lib/preview-index.ts` filters `SANDBOX_PREVIEW_KIND` out of the sweep entirely. Invalid, so it stays out of the inline comment and carries no assessment.

## Carried over unchanged

The items from my previous summary that were dispositioned rather than published still stand as they were: the comment length at `combobox-popover.tsx:215-221` and the redundant `autoFocusOnShow` at `examples/select-combobox-tab/select.tsx:76` remain declined non-goals, and the offer to file a tracking issue for the "Cover `defaultOpen` with a competing focus owner at mount" follow-up is still open. The `disabled` `ComboboxSelect` with `defaultOpen` note remains untested and is still a note rather than a finding.

app/src/sandbox/combobox-select-programmatic-open/index.react.tsx
diegohaz: Fixed in [`c21bb92`](https://github.com/ariakit/ariakit/commit/c21bb9276f19e0ea684a1f10d89fe5c911c8e17e): the paragraph now sits above the spacer.

Confirmed the complaint before changing anything, and confirmed the result after. On the fixed fixture in Chromium, `window.scrollY` is `0` on load, the hint paints at `getBoundingClientRect().top === 12`, and the `ComboboxSelect` sits at `1584`, so it stays 864px below the 720px desktop viewport that all three `test-browser.ts` projects use. The scroll test's `scrollY === 0` and `not.toBeInViewport()` preconditions keep the same margin they had.

The reorder is layout-neutral for everything below the spacer, since the boxes above the label sum to the same height either way, so the select's document offset is unchanged. All 5 tests in the file pass on Chrome, Firefox, and Safari, matching your run.

This was worth catching: the paragraph exists only because an earlier round found the `F2` shortcut undiscoverable in the preview, and the spacer I added for the scroll coverage quietly undid that. I extended the spacer comment to record the ordering constraint so a future edit does not reintroduce it.

No coverage is owed here. The affordance is a preview-authoring concern with no shipped behavior attached, sandbox previews are filtered out of the visual sweep by `getPreviewIndexPaths`, and the behavior the spacer exists to prove is already asserted by `scrolls the picker into view when it opens outside the viewport`.

diegohaz:
**Fixed**

- [`c21bb92`](https://github.com/ariakit/ariakit/commit/c21bb9276f19e0ea684a1f10d89fe5c911c8e17e) ([reply](https://github.com/ariakit/ariakit/pull/7079#discussion_r3738535221)): moved `Press F2 to open the vegetable picker.` above the spacer in `combobox-select-programmatic-open`, so the preview no longer loads on a blank screen. The spacer added for the scroll coverage had been placed above the one line that tells a person how to drive the sandbox, undoing an affordance an earlier round asked for. Verified on the fixed fixture: the hint paints at `top: 12` with `scrollY: 0`, while the select stays at `1584`, keeping the scroll test 864px of margin against the 720px viewport. All 5 tests in the file pass on Chrome, Firefox, and Safari.

PR #7082, opened 2026-08-07
Define focus for default-open menus

diegohaz:
## Motivation

A `Menu` that is already open when it mounts never receives focus. Focus stays wherever it was, while the first item is highlighted as if it were the active one.

```jsx
// Focus stays on the body. Arrow keys do nothing. "Rename" looks active anyway.
<Ariakit.MenuProvider defaultOpen>
  <Ariakit.Menu>
    <Ariakit.MenuItem>Rename</Ariakit.MenuItem>
    <Ariakit.MenuItem>Duplicate</Ariakit.MenuItem>
  </Ariakit.Menu>
</Ariakit.MenuProvider>
```

`Menu` builds on `Hovercard`, so it inherits an [`autoFocusOnShow` state that starts as `false`](https://github.com/ariakit/ariakit/blob/897182fbb250a656e8821a3c1ca7f479bd3670ff/packages/ariakit-components/src/hovercard/hovercard-store.ts#L36). That default is right for a hovercard, which opens on hover and must not steal focus. Only [`MenuButton`'s click and keydown handlers](https://github.com/ariakit/ariakit/blob/897182fbb250a656e8821a3c1ca7f479bd3670ff/packages/ariakit-react-components/src/menu/menu-button.tsx#L153) ever flip it to `true`, so a menu that mounts already open never runs them and [`useMenu` bails before resolving a focus target](https://github.com/ariakit/ariakit/blob/897182fbb250a656e8821a3c1ca7f479bd3670ff/packages/ariakit-react-components/src/menu/menu.tsx#L98).

The state is overloaded. `false` means both "a hover opened this, leave focus alone" and "nobody has said anything yet". A menu that mounts open lands in the second bucket and is treated as the first. `Select` and `ComboboxSelect` are `Popover`-based and never inherit that default, which is why #7079 had to answer the same question for a different composition.

## Solution

The two components want opposite defaults, so each one states its own and the exceptions opt out.

A hovercard opens on hover, so it keeps leaving focus alone and its disclosure opts in. A menu is the other way around: showing one is a request to interact with it, so the menu store arms itself and the hover trigger opts out.

```diff
  // in createMenuStore, after the hovercard store is created
+ hovercard.setState(
+   "autoFocusOnShow",
+   defaultValue(syncState.autoFocusOnShow, true),
+ );
```

```diff
  // in MenuButton's showOnHover, once the hover is known to open the menu
+ store?.setAutoFocusOnShow(false);
```

The value is written onto the hovercard store rather than the menu's own `initialState` because `createStore` pushes each parent store's value for shared keys into the child at init, so an `initialState` entry would be overwritten. `MenuButton` is the only menu hover trigger, so the opt-out is complete: `MenuAnchor` only positions.

Three alternatives were tried and rejected:

- **Removing the two reset-on-close effects in `Hovercard`.** Breaks `menu-placing-pass`: with the flag left `true` while a menu is open, the container stays an armed focus target, so replacing a focused item's DOM node lands focus on the container instead of restoring it to the item.
- **Disarming on close from the store, with a `sync` on `open`.** Breaks `menu-motion-transitions`. That fixture is a controlled menu, and a controlled round trip briefly writes the stale `open: false` back into the store, so the flag was disarmed before the menu actually opened.
- **Making `autoFocusOnShow` a store creation option.** Wiring it through `useStoreProps` turned it into a permanent controlled pin that reverted every internal write within a microtask, which silently disabled `HovercardDisclosure`'s opt-in. The escape hatch stays the existing [`autoFocusOnShow={false}`](https://ariakit.com/reference/menu#autofocusonshow) prop on `Menu`, matching #7079.

Focus lands on the menu container, which is where the store's own `initialFocus: "container"` default points and where a mouse click on `MenuButton` already puts it. Modal menus that mount open previously landed on the first item, through `Dialog`'s generic first-tabbable fallback rather than any menu decision, so this redefines that case to match. The issue body's claim that modal already focuses the container is inaccurate and was corrected in [this comment](https://github.com/ariakit/ariakit/issues/2946#issuecomment-4977621514).

## Coverage

Two sandboxes, because a menu that is open at page load already owns focus and cannot share a page with click-driven variants.

`app/src/sandbox/menu-default-open` is the reported scenario: a menu open from the first render, with no user gesture and no prior focus owner. It asserts that focus lands on the menu container with no item highlighted, and that ArrowDown still starts roving from there, so the menu is operable and not merely focused.

`app/src/sandbox/menu-conditional-mount` is the flow that motivates the pattern in practice: clicking **Add item** reveals the next available actions without a second click. Here the menu takes focus away from a live, focused button rather than from a blank body, which is a different branch in `Dialog`. It covers the non-modal case, the modal case, and the `autoFocusOnShow={false}` opt-out.

The reproduction landed first, before any library change, so CI recorded the red state: 4 failing and 1 passing, in happy-dom, React 18, Chrome, Firefox, and Safari.

| Scenario | Before | After |
| --- | --- | --- |
| Non-modal, open at page load | Focus on `body`, ArrowDown does nothing | Container focused, ArrowDown reaches the first item |
| Non-modal, mounted open after a click | Focus stays on the clicked button | Container focused |
| Modal, mounted open after a click | First item focused | Container focused |
| `autoFocusOnShow={false}` | Focus stays put | Unchanged |

The opt-out test deliberately does not pin whether an item stays highlighted while focus is declined. Nothing has decided that, and asserting it would freeze the exact half-state the issue calls wrong.

`menubar-navigation` is the red check for the hover opt-out: those menus render with `unmountOnHide`, so they are still armed at their first open, and the test asserts focus stays on `body` while hovering. It fails without the `MenuButton` hunk. `menu-placing-pass` and `menu-motion-transitions` guard the two rejected alternatives.

## Verification

- Full happy-dom suite: 257 files, 1718 tests. React 18: 190 files, 984 tests.
- Chrome `menu hovercard tooltip menubar`: 749 passed. Firefox: 607. Safari: 497.
- `pnpm tsc` and `pnpm lint` clean.
- Both reference generators checked. The only generated metadata change repo-wide is `use-menu-store`'s `autoFocusOnShow` default flipping from `false` to `true`, which the new `/** @default true */` override on `MenuStoreState` makes truthful.
- A store-replacement focus steal was suspected during review and disproved with an A/B probe: with and without the fix, replacing an open menu's store closes the menu and leaves focus on the clicking button, identically.

## Known gap

Menus that render while closed are disarmed by `Hovercard`'s reset-on-close effect, so a later bare `store.show()` on them does not move focus. A `<Menu unmountOnHide>` that has never opened does not render at all, so that effect never runs and its **first** non-gesture open does take focus, while later opens do not. Every `MenuButton` path is unaffected, since click and keydown arm, and focus and hover opt out.

Closing that window means moving where the reset runs relative to the `unmountOnHide` unmount gate, in shared `Hovercard` code that `menu-placing-pass` and `menu-motion-transitions` are sensitive to. It is left as is rather than fixed here.

## Workaround

Until the fix ships, a menu that intentionally mounts already open can enable the state itself after mount. `setAutoFocusOnShow` is only reachable from a store reference, and `useMenuStore` has no equivalent option, so the store has to be hoisted out of the provider.

```jsx
const menu = Ariakit.useMenuStore({ defaultOpen: true });

useEffect(() => {
  // Menu inherits Hovercard's `autoFocusOnShow: false`, and only a MenuButton
  // gesture flips it, so a menu that mounts already open never focuses its
  // container.
  // TODO: Remove this, and the store hoist it needs, once
  // https://github.com/ariakit/ariakit/issues/2946 is fixed.
  menu.setAutoFocusOnShow(true);
}, [menu]);

return (
  <Ariakit.MenuProvider store={menu}>
    <Ariakit.Menu>{/* items */}</Ariakit.Menu>
  </Ariakit.MenuProvider>
);
```

Assumptions and limitations:

- It assumes this menu should take focus. A default-open menu meant to stay passive, such as one restored from persisted state on page load, should use `<Ariakit.Menu autoFocusOnShow={false}>` instead.
- It only covers the initial mount-open. The effect runs once, and `Hovercard` resets the state to `false` on close. Ordinary open and close cycles keep working because `MenuButton` sets it again on click and keydown, but a later programmatic `store.show()` with no button gesture opens without focus again and needs its own `setAutoFocusOnShow(true)`.
- Focus lands on the menu container with no item highlighted, because the store's `initialFocus` stays at its `"container"` default. Add `store.setInitialFocus("first")` to highlight the first item instead. For a modal menu this changes today's behavior, which focuses the first item.
- An explicit `autoFocusOnShow={false}` on `Menu` still wins over the store state, so the workaround does not override a deliberate opt-out.
- Focus moves in an effect, one commit after mount, so it settles slightly later than the fix does.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the current direction</summary>

**Assessment**

The bug is a real WAI-ARIA menu-pattern violation: keyboard and assistive-technology users are stranded outside an open menu with no navigable focus. Impact is high per occurrence and low in aggregate, since `defaultOpen` menus are not the dominant pattern, and a documented workaround exists. The affected surface is public and supported.

Complexity at the boundary was 2 fixtures, 6 files, 5 scenarios, 226 lines, no library code, no helpers, no shared utilities, and no public contract. The merged precedent for the structurally identical `ComboboxSelect` problem, #7079, shipped 424 lines, 2 fixtures, 14 test bodies, a new shared helper, and 2 changesets. Measured against that precedent, this snapshot is under-built rather than over-built. Each scenario maps to a distinct branch of `canAutoFocusOnShow`: page load with no prior focus owner, mount while a live button holds focus, modal, explicit opt-out, and post-focus operability.

Finding severity across the three cycles was strictly decreasing: cycle 1 found a scope defect (the reported page-load scenario was never exercised), cycle 2 found a false-green defect (an unasserted premise), and cycle 3 found only comments and one test title. No finding in any cycle challenged the contract being encoded. #6737's closure reason does not transfer, since that concerned accumulating workarounds in library code and this snapshot contains none.

**Decision**

Continue. Two removals were applied at the boundary: a redundant `tabIndex={0}` on three `Ariakit.Button`s, which `Focusable` already supplies on Safari through `safariTabIndex`, verified by a green Safari run of the opt-out guard; and a dead `label` prop on the fixture's menu component. One proposed assertion was declined as a non-goal rather than applied, as described above.

</details>

## Acknowledgments

Thanks to [@ciampo](https://github.com/ciampo) for reporting the issue with a reproduction and for the original workaround.

### DIFF
--- a/packages/ariakit-components/src/menu/menu-store.ts
+++ b/packages/ariakit-components/src/menu/menu-store.ts
@@ -86,13 +86,13 @@ export function createMenuStore({
     hideTimeout: defaultValue(props.hideTimeout, syncState.hideTimeout, 0),
   });
 
-  // Hovercard opens on hover, so it leaves focus alone by default. A menu that
-  // starts open has no gesture to speak for it, and rendering it open is still
-  // a request to interact with it, so it takes focus instead.
+  // Hovercard leaves focus alone by default and its disclosure opts in. Menus
+  // do the opposite: they arm here and the hover trigger opts out. It has to be
+  // the hovercard store, whose value wins when the menu store initializes.
   // https://github.com/ariakit/ariakit/issues/2946
   hovercard.setState(
     "autoFocusOnShow",
-    defaultValue(syncState.autoFocusOnShow, hovercard.getState().open),
+    defaultValue(syncState.autoFocusOnShow, true),
   );
 
   const initialState: MenuStoreState = {
@@ -169,6 +169,8 @@ export interface MenuStoreState<T extends MenuStoreValues = MenuStoreValues>
   placement: HovercardStoreState["placement"];
   /** @default 0 */
   hideTimeout?: HovercardStoreState["hideTimeout"];
+  /** @default true */
+  autoFocusOnShow: HovercardStoreState["autoFocusOnShow"];
   /**
    * Determines the element that should be focused when the menu is opened.
    */
--- a/packages/ariakit-components/src/menu/menu-store.ts
+++ b/packages/ariakit-components/src/menu/menu-store.ts
@@ -86,6 +86,15 @@ export function createMenuStore({
     hideTimeout: defaultValue(props.hideTimeout, syncState.hideTimeout, 0),
   });
 
+  // Hovercard opens on hover, so it leaves focus alone by default. A menu that
+  // starts open has no gesture to speak for it, and rendering it open is still
+  // a request to interact with it, so it takes focus instead.
+  // https://github.com/ariakit/ariakit/issues/2946
+  hovercard.setState(
+    "autoFocusOnShow",
+    defaultValue(syncState.autoFocusOnShow, hovercard.getState().open),
+  );
+
   const initialState: MenuStoreState = {
     ...composite.getState(),
     ...hovercard.getState(),
--- a/packages/ariakit-react-components/src/menu/menu-button.tsx
+++ b/packages/ariakit-react-components/src/menu/menu-button.tsx
@@ -253,6 +253,9 @@ export const useMenuButton = createHook<TagName, MenuButtonOptions>(
         };
         const canShowOnHover = getShowOnHover();
         if (!canShowOnHover) return false;
+        // Menus take focus when they open, but a hover is the passive way to
+        // get there, so this one opts out.
+        store?.setAutoFocusOnShow(false);
         const parent = parentIsMenubar ? parentMenubar : parentMenu;
         if (!parent) return true;
         // When hovering over a menu button shows a menu and the menu button is
--- a/.changeset/300-menu-default-open-focus.md
+++ b/.changeset/300-menu-default-open-focus.md
@@ -1,0 +1,24 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Menus that mount already open take focus
+
+A [`Menu`](https://ariakit.com/reference/menu) that is already open when it mounts, through [`defaultOpen`](https://ariakit.com/reference/menu-provider#defaultopen) or a controlled open state, now moves focus into the menu. Previously focus stayed where it was while the first item still appeared highlighted, leaving keyboard and screen reader users outside an open menu with no way to navigate it.
+
+```tsx
+<MenuProvider defaultOpen>
+  <MenuButton>Actions</MenuButton>
+  {/* The menu now takes focus, and no item is highlighted. */}
+  <Menu>
+    <MenuItem>Rename</MenuItem>
+    <MenuItem>Duplicate</MenuItem>
+  </Menu>
+</MenuProvider>
+```
+
+Focus lands on the menu itself, which is where clicking [`MenuButton`](https://ariakit.com/reference/menu-button) with a mouse already puts it. Modal menus opened this way used to focus the first item instead, so both now behave the same. Menus opened by hovering a [`MenuButton`](https://ariakit.com/reference/menu-button), such as submenus and menubar menus, still leave focus alone. To open a menu without moving focus, for example one restored from saved state on page load, pass [`autoFocusOnShow={false}`](https://ariakit.com/reference/menu#autofocusonshow).
+
+Thanks to [@ciampo](https://github.com/ciampo) for reporting the issue and providing the workaround that informed this solution.
--- a/.changeset/300-menu-default-open-focus.md
+++ b/.changeset/300-menu-default-open-focus.md
@@ -4,21 +4,4 @@
 "@ariakit/react": patch
 ---
 
-Menus that mount already open take focus
-
-A [`Menu`](https://ariakit.com/reference/menu) that is already open when it mounts, through [`defaultOpen`](https://ariakit.com/reference/menu-provider#defaultopen) or a controlled open state, now moves focus into the menu. Previously focus stayed where it was while the first item still appeared highlighted, leaving keyboard and screen reader users outside an open menu with no way to navigate it.
-
-```tsx
-<MenuProvider defaultOpen>
-  <MenuButton>Actions</MenuButton>
-  {/* The menu now takes focus, and no item is highlighted. */}
-  <Menu>
-    <MenuItem>Rename</MenuItem>
-    <MenuItem>Duplicate</MenuItem>
-  </Menu>
-</MenuProvider>
-```
-
-Focus lands on the menu itself, which is where clicking [`MenuButton`](https://ariakit.com/reference/menu-button) with a mouse already puts it. Modal menus opened this way used to focus the first item instead, so both now behave the same. Menus opened by hovering a [`MenuButton`](https://ariakit.com/reference/menu-button), such as submenus and menubar menus, still leave focus alone. To open a menu without moving focus, for example one restored from saved state on page load, pass [`autoFocusOnShow={false}`](https://ariakit.com/reference/menu#autofocusonshow).
-
-Thanks to [@ciampo](https://github.com/ciampo) for reporting the issue and providing the workaround that informed this solution.
+Fixed the focus behavior of [`Menu`](https://ariakit.com/reference/menu) when it's rendered already open, for example with [`defaultOpen`](https://ariakit.com/reference/menu-provider#defaultopen): focus now moves to the menu container with no item highlighted, in both non-modal menus, which previously left focus alone, and modal menus, which previously focused the first item. Thanks to [@ciampo](https://github.com/ciampo).
--- a/app/src/sandbox/menu-conditional-mount/index.react.tsx
+++ b/app/src/sandbox/menu-conditional-mount/index.react.tsx
@@ -1,0 +1,73 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+
+interface NewItemActionsProps {
+  modal?: boolean;
+  autoFocusOnShow?: boolean;
+  onClose: () => void;
+}
+
+function NewItemActions({
+  modal,
+  autoFocusOnShow,
+  onClose,
+}: NewItemActionsProps) {
+  return (
+    <Ariakit.MenuProvider
+      defaultOpen
+      // Unmounting on close lets the preview be replayed by hand without
+      // reloading the page.
+      setOpen={(open) => {
+        if (open) return;
+        onClose();
+      }}
+    >
+      <Ariakit.MenuButton className="button">Actions</Ariakit.MenuButton>
+      <Ariakit.Menu
+        modal={modal}
+        autoFocusOnShow={autoFocusOnShow}
+        gutter={8}
+        className="menu"
+      >
+        <Ariakit.MenuItem className="menu-item">Rename</Ariakit.MenuItem>
+        <Ariakit.MenuItem className="menu-item">Duplicate</Ariakit.MenuItem>
+      </Ariakit.Menu>
+    </Ariakit.MenuProvider>
+  );
+}
+
+type Variant = "default" | "modal" | "quiet";
+
+// Adding an item mounts its actions already open, the way a real flow reveals
+// the next available actions without a second click. The button that was
+// clicked keeps focus, so the menu has to take focus away from a live element
+// rather than from a blank body.
+export default function Example() {
+  const [added, setAdded] = useState<Variant | null>(null);
+  const close = () => setAdded(null);
+
+  return (
+    <div>
+      <div>
+        <Ariakit.Button className="button" onClick={() => setAdded("default")}>
+          Add item
+        </Ariakit.Button>
+        {added === "default" && <NewItemActions onClose={close} />}
+      </div>
+      <div>
+        <Ariakit.Button className="button" onClick={() => setAdded("modal")}>
+          Add modal item
+        </Ariakit.Button>
+        {added === "modal" && <NewItemActions modal onClose={close} />}
+      </div>
+      <div>
+        <Ariakit.Button className="button" onClick={() => setAdded("quiet")}>
+          Add quiet item
+        </Ariakit.Button>
+        {added === "quiet" && (
+          <NewItemActions autoFocusOnShow={false} onClose={close} />
+        )}
+      </div>
+    </div>
+  );
+}
--- a/app/src/sandbox/menu-conditional-mount/index.react.tsx
+++ b/app/src/sandbox/menu-conditional-mount/index.react.tsx
@@ -1,5 +1,5 @@
 import * as Ariakit from "@ariakit/react";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 
 interface NewItemActionsProps {
   modal?: boolean;
@@ -12,27 +12,16 @@ function NewItemActions({
   autoFocusOnShow,
   onClose,
 }: NewItemActionsProps) {
-  const menu = Ariakit.useMenuStore({
-    defaultOpen: true,
-    // Unmounting on close lets the preview be replayed by hand without
-    // reloading the page.
-    setOpen: (open) => {
-      if (open) return;
-      onClose();
-    },
-  });
-
-  useEffect(() => {
-    // Menu inherits Hovercard's `autoFocusOnShow: false`, and only a MenuButton
-    // gesture flips it, so a menu that mounts already open never focuses its
-    // container.
-    // TODO: Remove this, and the store hoist it needs, once
-    // https://github.com/ariakit/ariakit/issues/2946 is fixed.
-    menu.setAutoFocusOnShow(true);
-  }, [menu]);
-
   return (
-    <Ariakit.MenuProvider store={menu}>
+    <Ariakit.MenuProvider
+      defaultOpen
+      // Unmounting on close lets the preview be replayed by hand without
+      // reloading the page.
+      setOpen={(open) => {
+        if (open) return;
+        onClose();
+      }}
+    >
       <Ariakit.MenuButton className="button">Actions</Ariakit.MenuButton>
       <Ariakit.Menu
         modal={modal}
--- a/app/src/sandbox/menu-conditional-mount/index.react.tsx
+++ b/app/src/sandbox/menu-conditional-mount/index.react.tsx
@@ -1,5 +1,5 @@
 import * as Ariakit from "@ariakit/react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 
 interface NewItemActionsProps {
   modal?: boolean;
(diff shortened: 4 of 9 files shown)


packages/ariakit-components/src/menu/menu-store.ts
```diff
@@ -86,6 +86,15 @@ export function createMenuStore({
...
   });
 
+  // Hovercard leaves focus alone by default and its disclosure opts in. Menus
+  // do the opposite: they arm here and the hover trigger opts out. It has to be
+  // the hovercard store, whose value wins when the menu store initializes.
+  // https://github.com/ariakit/ariakit/issues/2946
+  hovercard.setState(
+    "autoFocusOnShow",
+    defaultValue(syncState.autoFocusOnShow, true),
+  );
```
diegohaz, line 96: This arms `autoFocusOnShow` for every menu store, not only the ones this PR is about. For a `<Menu unmountOnHide>` the arming is never undone, so the change reaches well past "rendered already open".

Both of `Hovercard`'s disarm paths miss that case. I instrumented them on this head with a plain `<Menu unmountOnHide>` opened by `menu.show()` from an ordinary button, closed with Escape:

```
PROBE mounted flag=true
RESET-EFFECT domReady=true open=true
UNMOUNT-CLEANUP openRef=true
PROBE open1 flag=true focused=true
UNMOUNT-CLEANUP openRef=true
PROBE closed flag=true
```

The `open`-gated effect never runs with `open === false`, because the menu unmounts in the same commit that flips it, and the unmount cleanup reads `useLiveRef(open)`, whose layout effect does not run for an unmounting commit, so it still sees `true`. The store stays armed for the component's whole life.

Measured against `origin/main` with the same fixture:

| `store.show()` on | main | this head |
| --- | --- | --- |
| `<Menu unmountOnHide>` | focus stays put | focus moves into the menu container, on every open |
| `<Menu>` (stays mounted) | focus stays put | focus stays put |

Real-world consumer: an app that opens a menu from its own trigger instead of `MenuButton`, such as a context menu on `contextmenu` with `MenuAnchor`, or a menu opened from a keyboard shortcut. Adding or removing `unmountOnHide`, which consumers set for animation or DOM-weight reasons, silently flips whether that call moves focus. Neither behavior is described in the changeset and neither has a test.

The unconditional `true` is also what forces the `MenuButton` hunk: removing that hunk alone fails three `menubar-navigation` tests in Chrome, which I reproduced locally.

The previous commit on this branch, [`49621f0`](https://github.com/ariakit/ariakit/commit/49621f0281d47b4f4ef40baec179762176b88a09), already scoped this:

```suggestion
  // Hovercard opens on hover, so it leaves focus alone by default. A menu that
  // starts open has no gesture to speak for it, and rendering it open is still
  // a request to interact with it, so it takes focus instead.
  // https://github.com/ariakit/ariakit/issues/2946
  hovercard.setState(
    "autoFocusOnShow",
    defaultValue(syncState.autoFocusOnShow, hovercard.getState().open),
  );
```

For any menu closed at creation this writes exactly what `createHovercardStore` already computed, so behavior outside the reported scenario is unchanged by construction rather than by test coverage. I ran it locally: the five new tests pass, `menu menubar hovercard tooltip` is 134/134 in Chrome with the `MenuButton` hunk removed, and `menu menubar hovercard tooltip dialog popover combobox select` is 812/812 in happy-dom. [`49621f0`](https://github.com/ariakit/ariakit/commit/49621f0281d47b4f4ef40baec179762176b88a09) was itself green on all 44 CI checks.

If the broader "showing a menu is a request to interact with it" contract is the goal, it should reach mounted menus too, which the Known gap says it does not, and it needs its own changeset entry and coverage rather than arriving only for menus that happen to set `unmountOnHide`.

**Assessment**

- Severity: Medium (estimate, 75% confidence), because focus lands somewhere the consumer did not ask for rather than being lost or trapped.
- Priority: P1 (verified), because the scope decides what ships here and whether the second hunk needs to exist at all.
- Breakage likelihood: Possible (estimate, 70% confidence), because it needs a menu opened without a `MenuButton` gesture and with `unmountOnHide` set; every `MenuButton` path is unaffected.
- Fix effort: Easy (verified), because [`49621f0`](https://github.com/ariakit/ariakit/commit/49621f0281d47b4f4ef40baec179762176b88a09)'s one-liner is already written and green, and applying it deletes the `MenuButton` hunk.

packages/ariakit-react-components/src/menu/menu-button.tsx
```diff
@@ -253,6 +253,9 @@ export const useMenuButton = createHook<TagName, MenuButtonOptions>(
         };
         const canShowOnHover = getShowOnHover();
         if (!canShowOnHover) return false;
+        // Menus take focus when they open, but a hover is the passive way to
+        // get there, so this one opts out.
+        store?.setAutoFocusOnShow(false);
```
diegohaz, line 258: This writes store state from inside `showOnHover`, whose job is to answer whether this hover should show the menu. `useHovercardTrigger` calls the predicate on every qualifying `mousemove` and only then schedules the show, so on a submenu, where `timeout` defaults to `150`, the flag is already `false` for hovers that never open anything: a `mouseleave` inside that window clears the timeout and the menu stays closed.

I could not build a user-visible failure out of that, because `onFocus`, `onClick`, and `onKeyDown` all rewrite the flag on the paths that matter, so treat the placement as a robustness point rather than a bug. The `parent.setActiveId` call just below is the local precedent for a side effect here, but it is idempotent and re-derivable from the same hover, while this one is the only thing standing between the unconditional `true` in `createMenuStore` and hover-opened menus taking focus. I verified that: with this hunk removed and the store hunk kept, three `menubar-navigation` tests fail in Chrome.

With the store default scoped to menus created open, as suggested in https://github.com/ariakit/ariakit/pull/7082#discussion_r3740647065, this hunk is not needed at all. I ran `menu menubar hovercard tooltip` in Chrome that way and all 134 tests pass without it.

If you want to keep it as an independent statement that hover is the passive path, rather than as compensation for the store default, it deserves that framing in the comment and a test that fails without it for its own reason, since today it is only covered by a fixture that fails for the store default's sake. Moving the call next to `store.setDisclosureElement`, where the show is actually committed, would also keep the predicate a predicate.

**Assessment**

- Severity: Low (estimate, 80% confidence), because the known reachable paths rewrite the flag before it is read.
- Priority: P2 (estimate, 75% confidence), because it is a placement and coverage concern rather than a defect, and it disappears if the store default is scoped.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because reaching it needs an armed store, a cancelled hover, and a later non-gesture open, and I could not assemble those into a failing case.
- Fix effort: Easy (verified), because it is either deleted with the scoped store default or moved a few lines down.

packages/ariakit-components/src/menu/menu-store.ts
```diff
@@ -160,6 +169,8 @@ export interface MenuStoreState<T extends MenuStoreValues = MenuStoreValues>
   placement: HovercardStoreState["placement"];
   /** @default 0 */
   hideTimeout?: HovercardStoreState["hideTimeout"];
+  /** @default true */
+  autoFocusOnShow: HovercardStoreState["autoFocusOnShow"];
```
diegohaz, line 173: This publishes a `true` default for `autoFocusOnShow` on the `useMenuStore` reference, but for a menu rendered closed, which is the common case, `Hovercard`'s reset effect writes `false` in the first effect flush after mount. On a `<Menu>` that stays mounted I measured the state as `false` at the moment `store.show()` opened it, and focus did not move.

So the reference will tell a reader the menu store starts armed, while the only value they can observe from an effect or an event handler is `false`, and `store.show()` on that menu does not focus it. That is the Known gap, and the generated reference is where consumers are least likely to learn about it.

If the store default gets scoped to menus created open, this line needs different wording anyway, something like the default being `true` only when the menu is created open. If the unconditional default stays, it is worth saying here that the state is reset when the menu is hidden, so the reference does not promise behavior the component immediately undoes.

**Assessment**

- Severity: Low (verified), because it misleads rather than breaks anything.
- Priority: P2 (estimate, 80% confidence), because it ships with the reference metadata this PR regenerates and is cheap to correct now.
- Breakage likelihood: Likely (verified), because every reader of the `useMenuStore` reference sees the default, and it is wrong for every menu rendered closed.
- Fix effort: Easy (verified), because it is one JSDoc line.

diegohaz (commented):
Initial review of the full diff from the merge base [`897182f`](https://github.com/ariakit/ariakit/commit/897182fbb250a656e8821a3c1ca7f479bd3670ff) to head [`d062ae6`](https://github.com/ariakit/ariakit/commit/d062ae6a2f7a122081404c80bc83348724a63ef9), plus the surrounding menu, hovercard, and dialog code the change depends on. There were no existing review threads on this pull request, so there was nothing to revalidate.

The fix is real and the reproduction is honest. I reverted both library hunks and confirmed the red state locally in happy-dom: 4 failed, 1 passed, with the opt-out scenario as the passing one, which matches what the branch recorded on [`dc23df1`](https://github.com/ariakit/ariakit/commit/dc23df1f6170c8f47f262f4339064afd6ae249b1). My findings are about how far the store default reaches, not about whether the bug is worth fixing.

## Findings

1. https://github.com/ariakit/ariakit/pull/7082#discussion_r3740647065 on `menu-store.ts`: the unconditional `autoFocusOnShow: true` reaches past menus rendered open. For `<Menu unmountOnHide>` neither of `Hovercard`'s disarm paths ever runs, so every programmatic `store.show()` now moves focus, while the same call on a mounted menu still does not. Not in the changeset, not covered, and it is what forces the second hunk. P1.
2. https://github.com/ariakit/ariakit/pull/7082#discussion_r3740647680 on `menu-button.tsx`: the disarm is a state write inside a boolean predicate that also runs for hovers that never open a menu. I could not turn that into a user-visible failure, so it is a placement and coverage point, and it disappears entirely under the scoped store default. P2.
3. https://github.com/ariakit/ariakit/pull/7082#discussion_r3740648118 on `MenuStoreState`: the new `@default true` will publish a `useMenuStore` default that no menu rendered closed ever exposes to its consumer. P2.

## Not published inline

The **Known gap** section describes the `unmountOnHide` case as "its **first** non-gesture open does take focus, while later opens do not". I measured the opposite on this head: the flag is `true` before the first open and still `true` after closing, so every later open takes focus too. The instrumentation is in finding 1. This matters because the section is the record of what ships, and the behavior it understates is exactly the one with no changeset entry and no test.

- Severity: Low (verified), because the code behaves consistently and only the description is wrong.
- Priority: P1 (estimate, 85% confidence), because it is the only place the broadened scope is written down.
- Breakage likelihood: Likely (verified), because anyone reading the section to decide whether the gap affects them gets the wrong answer.
- Fix effort: Easy (verified), because it is a paragraph, and it becomes moot if the scoped default lands.

## Alternatives weighed

The three alternatives listed in the description all check out as described, and I did not retry them. The one I did try is the scoped default that [`49621f0`](https://github.com/ariakit/ariakit/commit/49621f0281d47b4f4ef40baec179762176b88a09) already had on this branch, and it is the substance of finding 1. Reading `props.autoFocusOnShow` in `createHovercardStore` next to `placement` and `timeout`, rather than writing the state after creation, would match the surrounding idiom, but it turns an internal default into a public store option on `useHovercardStore` and `useMenuStore`, so I do not think it is worth the API surface. Declined non-goal.

## Checked and holding

- Red and green proof: both hunks reverted gives 4 failed and 1 passed in happy-dom, matching the branch history.
- The `MenuButton` hunk is genuinely load-bearing for the chosen approach: removing it alone fails `show menus by hovering over items without moving focus`, `show menus by hovering over items and subitems`, and `moving between menus with arrow keys after hovering over subitems` in Chrome.
- `MenuAnchor` really is position-only, and `useHovercardTrigger` has exactly one consumer in the menu package, so the hover opt-out has no second site to cover.
- `@ariakit/solid-components` has no menu, so the changeset frontmatter is complete at `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/react`.
- Changeset links resolve: `/reference/menu` and `/reference/menu-provider` both return 200, and the deployed page carries the `defaultopen` anchor. The single-paragraph credit form matches the simple-entry rule.
- Dropping `tabIndex={0}` from the fixture buttons holds up: the opt-out test asserts focus on the clicked `Ariakit.Button`, and all five new tests pass in Safari.
- Sandbox split, naming, issue URLs per test, the `flushFrames` comment, and the happy-dom duplicates all follow the fixture and waiting rules.

## Commands run

`pnpm test` on `menu menubar hovercard tooltip dialog popover combobox select` (812 passed), `pnpm test-chrome` on the same four groups (134 passed), `pnpm test-safari` on both new sandboxes, and targeted red and green runs with each hunk reverted in turn. All experiments were reverted; the worktree is back at [`d062ae6`](https://github.com/ariakit/ariakit/commit/d062ae6a2f7a122081404c80bc83348724a63ef9) with a clean status.

## Not fully validated

I suspected the hover disarm might also fix a pre-existing case on `main`, where an `unmountOnHide` menu stays armed after a click-open closed by an outside click and a later hover reopens it focused. I could not build it: the `menubar-navigation` items do not open on click, so the probe never reached the state it was meant to measure. If that case is real, the hunk has standalone value beyond compensating for the store default, which is why finding 2 is P2 rather than a request to delete it.

I did not run Firefox, iOS, or Android locally, and I did not re-verify the reference generator diff or the store-replacement A/B probe described in the Verification section.

No follow-up issues were registered: every finding above belongs to this change rather than to separable work.

diegohaz:
Closing this without merging.

The review threads on the last head showed that this change reaches well past the reported scenario, and that the machinery it leans on is not reliable enough to lean on.

[#discussion_r3740647065](https://github.com/ariakit/ariakit/pull/7082#discussion_r3740647065) is the one that decides it. Arming `autoFocusOnShow` for every menu store depends on `Hovercard` disarming the menus that should not be armed, and both of its disarm paths miss `<Menu unmountOnHide>`: the `open`-gated effect never runs with `open === false`, because the menu unmounts in the same commit that flips it, and the unmount cleanup reads a live ref whose layout effect does not run for an unmounting commit, so it still sees `true`. The store stays armed for the component's whole life, and `store.show()` from a consumer's own trigger moves focus on those menus and not on menus that stay mounted. Whether `unmountOnHide` is set, which consumers choose for animation or DOM weight, silently decides a focus behavior.

The other two threads follow from the same root. [#discussion_r3740647680](https://github.com/ariakit/ariakit/pull/7082#discussion_r3740647680) notes that the `MenuButton` hunk exists only to compensate for the unconditional default, and writes store state from inside a predicate whose job is to answer whether a hover should show the menu. [#discussion_r3740648118](https://github.com/ariakit/ariakit/pull/7082#discussion_r3740648118) notes that the published `@default true` is wrong for every menu rendered closed, because the reset writes `false` in the first effect flush after mount.

Taken together, the problem is not the default value. It is that `autoFocusOnShow` is a single boolean whose lifecycle is spread across the store default, two `Hovercard` effects that undo it, and four `MenuButton` handlers that rewrite it, with no single place that owns the question "should this particular open move focus?". Every attempt to change the default collided with a different consumer of that machinery:

| Attempt | Outcome |
| --- | --- |
| Seed the menu's own `initialState` | Overwritten at `storeInit`, which pushes parent store values into the child |
| Remove `Hovercard`'s two reset-on-close effects | `menu-placing-pass` fails: an armed container outranks focus restoration when a focused item's node is replaced |
| Disarm from the store with a `sync` on `open` | `menu-motion-transitions` fails: a controlled round trip briefly writes the stale `open: false` and disarms before the menu opens |
| Make `autoFocusOnShow` a store option via `useStoreProps` | Becomes a permanent controlled pin that reverts internal writes in a microtask, disabling `HovercardDisclosure`'s opt-in |
| Default `true` for menus, hover opts out | Green everywhere, but arms `unmountOnHide` menus for their whole life, per the thread above |

So this needs an architecture pass over how `Hovercard` and `Menu` express auto-focus intent, rather than another adjustment to where the default is written. Something that models the intent of a given open, instead of a sticky flag that different components arm and disarm at different times, would let the menu case fall out without the compensating hunks this PR accumulated.

Worth keeping from this branch, which stays available at `define-menu-default-open-focus`:

- `dc23df1f6` adds the reproduction as two sandboxes, `menu-default-open` for the page-load case and `menu-conditional-mount` for the click-to-mount case, covering non-modal, modal, and the `autoFocusOnShow={false}` opt-out. It recorded the red state on CI: 4 failing and 1 passing across happy-dom, React 18, Chrome, Firefox, and Safari.
- `49621f028` is the narrow variant, where the seed is conditioned on the store's initial `open` state. It touches one file, needs no `MenuButton` hunk, and was green on all 44 CI checks. It fixes the reported scenario without claiming the broader contract, so it is the smallest thing that could ship if the overhaul turns out to be far off.
- The failure evidence in the table above, so the next attempt does not have to rediscover it.

A summary is going on #2946 so this is findable from the issue. The workaround there remains valid.

PR #7085, opened 2026-08-08
Announce a trigger as expanded while its popup is open

diegohaz:
## Motivation

A trigger that controls a popup keeps reporting `aria-expanded="false"` while that popup is visible, whenever the open did not pass through the trigger and another element owned focus at that moment. An app-wide shortcut, an effect, a timer, or an async result calling `store.show()` all reach this.

```jsx
// Focus is in the note field, then a shortcut opens the popup.
const popover = Ariakit.usePopoverStore();

useEffect(() => {
  const onKeyDown = (event) => {
    if (event.key !== "F2") return;
    popover.show();
  };
  document.addEventListener("keydown", onKeyDown);
  return () => document.removeEventListener("keydown", onKeyDown);
}, [popover]);
```

[`Dialog`](https://ariakit.com/reference/dialog) records the element that owns focus when the popup opens as the store's disclosure element. That is deliberate and supported: it is what lets a plain [`Button`](https://ariakit.com/reference/button) act as the disclosure without being a [`DialogDisclosure`](https://ariakit.com/reference/dialog-disclosure). [`Disclosure`](https://ariakit.com/reference/disclosure) then reports itself expanded only while its own element still holds that slot, so the announced state is tied to a slot that also serves focus restoration and outside-interaction classification. When those two roles disagree, the announced state is the one that ends up wrong.

The same gate also splits two triggers that share one store: whichever one holds the slot announces the content, and the other reports itself collapsed while the very content it controls is visible.

On `MenuButton` this is not announcement-only. Menubar triggers read each other's `aria-expanded` back off the DOM to decide whether hovering a sibling opens its menu, so a trigger that under-reports its own open menu takes that away from every sibling.

## Solution

[`Disclosure`](https://ariakit.com/reference/disclosure) now sets `aria-expanded` from the store's `open` state instead of from whether its own element still holds the disclosure slot. Every component built on it inherits the fix: [`DialogDisclosure`](https://ariakit.com/reference/dialog-disclosure), [`PopoverDisclosure`](https://ariakit.com/reference/popover-disclosure), [`MenuButton`](https://ariakit.com/reference/menu-button), and the deprecated [`Select`](https://ariakit.com/reference/select). It is the same thing [`Combobox`](https://ariakit.com/reference/combobox) and [`ComboboxDisclosure`](https://ariakit.com/reference/combobox-disclosure) already did for the popup they control, and that #7081 did for [`ComboboxSelect`](https://ariakit.com/reference/combobox-select).

This is a removal rather than an addition. The `useState` that held the announced value and the effect branch that computed it are gone, so the attribute is now a pure function of the open state. It is therefore correct on the first render too, which also removes a tick where a default-open disclosure announced itself collapsed.

The effect that assigns the disclosure element is untouched, so nothing derived from that slot changes: focus still returns to the element the user was on, Escape still accepts the same targets, and outside interactions are classified as before.

That answers both questions the issue asked to settle. Focus returning to where the user was typing is the supported behavior for an open that never passed through a trigger, so it stays, and the sandbox pins it. And several triggers over one store all announce the content, because they all do control it.

### Why every trigger announces

The change does not make each trigger an owner of the content. It takes ownership out of the announcement. The store's disclosure element stays singular and untouched, and it keeps answering the question it is good at: where focus came from, and where it goes back. Whether the thing a button controls is open is a different question, and it has a different answer per button.

That a content may have several disclosure buttons is what makes the previous behavior wrong rather than what argues for it. `aria-expanded` and `aria-controls` are set side by side in the same props object, from the same store. Before this change only the trigger holding the slot said `true`, so every other trigger pointed `aria-controls` at a visible element while announcing it collapsed. Those two attributes can no longer disagree.

This is also not a new model for the library. [`Combobox`](https://ariakit.com/reference/combobox) and [`ComboboxDisclosure`](https://ariakit.com/reference/combobox-disclosure) are two controls for one popup, and both have long announced from the popup's open state. In the shipped `combobox-disclosure` example the input and the button both report `aria-expanded="true"` at the same time. [`Disclosure`](https://ariakit.com/reference/disclosure) was the outlier.

### Reviewed alternatives

The registry sketched in #7080's triage was re-derived and rejected. It cannot repair two triggers over one store, since its fallback is the single registered disclosure and neither is the active element. It moves the disclosure element, which takes focus restoration away from the element the user was on, contradicting the behavior this pull request pins. And it adds a store-level subsystem where a deletion suffices.

Tracking every connected disclosure element only to decide the announcement was considered too. Every rendered `Disclosure` would register, so the test would reduce to the open state that this change already reads, and the registry would carry no information the component does not have.

### Coverage

`app/src/sandbox/disclosure-expanded-state` is a note editor whose popups are opened by app-wide shortcuts while focus stays in the note field. Its six tests, duplicated in happy-dom and in the browser, failed against the previous behavior and pass with the fix. They cover the four triggers the issue reports, the two triggers over one store, where focus lands after the popup closes, and the menubar hover behavior that reads the attribute back off the DOM.

Framework component regressions belong in app-owned fixtures rather than package source, so there is no package-level test for this; the sandbox is the regression surface.

## Preview

Verified by hand in Chromium on `app/src/sandbox/disclosure-expanded-state`. With the caret in the note field, pressing F8 leaves focus there and the accessibility tree reports the trigger as expanded:

```yaml
textbox "Note" [active]: Draft notes
menubar:
  menuitem "Format" [expanded]
  menu "Format":
    menuitem "Bold"
    menuitem "Italic"
  menuitem "Insert"
```

Hovering the sibling trigger then opens its menu, which the announced state was blocking, and leaves the first menu open, which is #7086:

```yaml
menubar:
  menu "Format"
  menuitem "Insert" [expanded]
  menu "Insert":
    menuitem "Image"
    menuitem "Table"
```

No screenshot or recording is attached. The change itself is an ARIA attribute with no rendered form, and its one visible consequence lives in a sandbox that carries no styling by convention, so an open menu paints over the text beneath it and a capture of that state reads as overlapping text rather than as two menus. The accessibility tree above is the same browser state, read where the change is actually observable.

## Workaround

Until the fix ships, supply the attribute directly on the trigger, reading the popup's open state from its store.

```tsx
const popover = Ariakit.usePopoverStore();
const open = Ariakit.useStoreState(popover, "open");

// TODO: Remove once the trigger derives this on its own.
// https://github.com/ariakit/ariakit/issues/7083
<Ariakit.PopoverDisclosure store={popover} aria-expanded={open} />;
```

The same shape works for [`Disclosure`](https://ariakit.com/reference/disclosure), [`DialogDisclosure`](https://ariakit.com/reference/dialog-disclosure), [`MenuButton`](https://ariakit.com/reference/menu-button), and the deprecated [`Select`](https://ariakit.com/reference/select). Every layer between those components and `Disclosure` spreads consumer props after its own defaults, so the value reaches the DOM. `open` is the same value [`Combobox`](https://ariakit.com/reference/combobox) and [`ComboboxDisclosure`](https://ariakit.com/reference/combobox-disclosure) already expose for the popup they control, and [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) derives it too since #7081, so those three need no override.

When several triggers share one store, apply it to each of them. They all control the same content, so they all announce the same state.

If your tree uses a provider without a store variable, read the store from context in a child component instead of hoisting it:

```tsx
function Trigger() {
  const popover = Ariakit.usePopoverContext();
  const open = Ariakit.useStoreState(popover, "open");
  // TODO: Remove once the trigger derives this on its own.
  // https://github.com/ariakit/ariakit/issues/7083
  return <Ariakit.PopoverDisclosure aria-expanded={open} />;
}
```

`useDisclosureContext`, `useDialogContext`, `useMenuContext`, and `useSelectContext` are the equivalents for the other components.

This corrects the announced state only. The popup still treats the element that owned focus at open time as its disclosure element, so everything else derived from that slot is unchanged rather than fixed: where focus returns on close, which targets Escape accepts, how an outside interaction is classified, and which elements a hovercard counts as its own. The library fix in this pull request keeps that boundary too.

Reclaiming the slot with `setDisclosureElement` was considered and rejected. It repairs the same attribute, but it takes focus restoration away from the element the user was on, which this pull request's own tests pin.

In a menubar there is one visible consequence, and it applies to the fix as well as to the workaround. Correcting the announced state is what lets a sibling trigger open its menu on hover again, but the menu that was already open stays open, because a menubar menu decides to hide on hover from the disclosure element rather than from the announced state. Two menus are then visible at once. That is tracked separately in #7086, which carries its own workaround.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the userland override and the derive-from-open fix</summary>

**Assessment**

Issue severity is medium to high. All five affected triggers are exported from `@ariakit/react`, `aria-expanded` is a required contract for a button controlling a popup, and the attribute is read back off the DOM by `hasActiveItem` and `menuHasFocus`, so a wrong value silently disables menubar behavior rather than only misannouncing it. Expected frequency is moderate: any `store.show()` that does not pass through the trigger reaches it, but only while another element owns focus. Supported scope is one sentence in the issue, plus the two questions it asks to settle, both of which are now pinned by committed assertions. Likely user impact is silent, which is what makes it worth fixing.

Implementation complexity is very low and the direction subtracts rather than adds. All five call sites funnel into `useDisclosure`, so the fix removes the `useState`/`setExpanded` pair and its effect branch instead of introducing a branch, and it also removes a tick where a default-open disclosure renders collapsed on first paint. The workaround is one prop per trigger and calls no store method.

Three consecutive gate rounds produced findings, which is what triggered this reassessment. Reading them together, every finding across the three rounds landed on prose, a comment, comment syntax, process ordering, or a GitHub issue body, and none touched the workaround's mechanism, which two rounds independently re-derived and confirmed. The driver was fan-out rather than intricacy: round 1 discovered a second defect while validating the workaround, and one nuanced claim about it had to be stated correctly on five surfaces. Round 3's finding had an external cause, a concurrent write from the repository's own triage automation that clobbered an issue-body edit.

**Decision**

Continue unchanged. The registry sketched in #7080's triage was re-derived and rejected on three independent grounds: it does not repair two triggers over one store, since its fallback is the single registered disclosure and neither is active; it moves the disclosure element, which takes focus restoration away from the element the user was on, contradicting an assertion this pull request already committed; and it adds a store-level subsystem to fix what a deletion fixes.

Registering the menubar residual as #7086 rather than fixing it here stands. The fix does not create that logic, it removes an unrelated gate that happened to mask it, and that mask is the reported bug. The two live in different files with different owners, and repairing the residual means redesigning when a menubar menu hides on hover-outside.

Treat the residual's prose as settled. It is now correct on every surface, and further polish would restart the churn this reassessment diagnoses.

</details>


.changeset/300-disclosure-expanded-state.md
```diff
@@ -0,0 +1,12 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+A trigger announces its popup as expanded while the popup is open
```
diegohaz: This could be a single sentence changeset.

diegohaz: Fixed in [`38d2db9`](https://github.com/ariakit/ariakit/commit/38d2db95c99b6ebc08a4953bdb9f80ec0f4d65e2): collapsed the entry to one sentence, naming `PopoverDisclosure` and `MenuButton` as the two representative derived components instead of listing all of them.

diegohaz:
**Fixed**

- [`38d2db9`](https://github.com/ariakit/ariakit/commit/38d2db95c99b6ebc08a4953bdb9f80ec0f4d65e2) ([reply](https://github.com/ariakit/ariakit/pull/7085#discussion_r3740503729)): collapsed the changeset to a single sentence, naming `PopoverDisclosure` and `MenuButton` as the two representative derived components rather than listing every one.

diegohaz:
Closing this without landing.

The direction was rejected: `aria-expanded` can be used as a CSS selector to target the disclosure button that was actually used to show the content, and the buttons for one content may live in different parts of an app. Deriving the attribute from the open state makes every one of them match, which collapses that distinction.

[`dc9117b`](https://github.com/ariakit/ariakit/commit/dc9117b73) reverts everything this pull request added, so the branch tree is byte-identical to [`8ad072d`](https://github.com/ariakit/ariakit/commit/8ad072da596677e7aee42188cf727705a730775e) and the diff against `main` is empty. The commits ahead of the revert are kept as a record of the approach and why it was not taken.

#7083 stays open. The announcement part of it is now a question about what `aria-expanded` should mean rather than a defect, since the attribute is deliberately reporting the button that was used. What survives independently of that choice is the menubar consequence: after a programmatic open, hovering a sibling trigger does not open its menu, because `hasActiveItem` reads the announced state off the DOM to decide. Fixing that without touching the attribute means giving the menubar a way to know which of its items have open menus instead of inferring it from `aria-expanded`.

PR #7090, opened 2026-08-08
Keep an explicitly assigned disclosure element when a Dialog opens

### LINKED ISSUE #7087: Dialog overwrites an explicitly assigned disclosure element when it opens
### Problem

`setDisclosureElement` is the natural way for an app to say which trigger is opening a popup it shows programmatically. Assigning it before `show()` has no effect, because `Dialog` records the focused element into the same slot in a layout effect right after the open, overwriting whatever the app set.

```jsx
// Focus is in the note field, then a shortcut opens the popup and names its
// trigger as the opener.
popover.setDisclosureElement(triggerRef.current);
popover.show();
```

https://github.com/ariakit/ariakit/blob/0e69adb5a17a6ce32802e5473a9eaa0a950f9919/packages/ariakit-react-components/src/dialog/dialog.tsx#L284-L309

### Observed

Measured with focus in a text field and a [`PopoverDisclosure`](https://ariakit.com/reference/popover-disclosure) as the intended opener, on `0e69adb5a`:

| What the app does | Trigger ends up as the disclosure element |
| --- | --- |
| `setDisclosureElement(trigger)`, then `show()` | **no** |
| `show()`, then `setDisclosureElement(trigger)` in the same tick | **no** |
| `show()`, then `setDisclosureElement(trigger)` in a microtask | yes |

The first two lose to the capture, which runs in a layout effect after the render that `show()` schedules. Only an assignment deferred past that effect survives, which is not something a reader would guess from the method name.

### Deferring is not a substitute

The microtask form assigns the slot, but the popup has already computed its outside state from the slot it held when it opened. Measured in the same fixture:

| Open route | Escape from the element that had focus at open time closes the popup |
| --- | --- |
| `show()`, then `setDisclosureElement` in a microtask | **no** |
| Click on the trigger | yes |

So the only form that assigns the slot successfully also takes Escape away from the element the user is actually on. That is the same cost recorded in https://github.com/ariakit/ariakit/issues/7080#issuecomment-5223449158 when this shape was rejected as a workaround.

### Expected outcome

An explicit `setDisclosureElement` before `show()` should survive the open, and the popup should compute its outside state from it. The capture is a fallback for when nothing said otherwise, and opens with focus on `BODY` already skip it, so it should not overwrite an assignment the app made deliberately.

### Context

This came out of #7083. Deriving `aria-expanded` from the popup's open state was rejected there, because the attribute is meant to mark the trigger that was actually used, and a content may have several triggers in different parts of an app. That makes `setDisclosureElement` the idiomatic way to name the trigger for a programmatic open, which is what makes this worth fixing.

Worth settling alongside it: with the trigger holding the slot, focus returns to the trigger on close rather than to where the user was when the popup opened.

diegohaz:
## Motivation

`setDisclosureElement` is how an application names the trigger a popup belongs to when it shows that popup itself. Assigning it before `show()` had no effect, because `Dialog` captures the currently active element into the same slot in a layout effect right after the open, overwriting whatever the application had set.

```ts
// The caret is in the note field, and a shortcut opens the popup naming the
// button it belongs to as the opener.
popover.setDisclosureElement(triggerRef.current);
popover.show();
```

The named button kept `aria-expanded="false"`, the popup positioned itself and computed its outside state from the field the caret happened to be in rather than from that button, and dismissing the popup from inside returned focus to the field.

Deferring the assignment until after the capture is not a workaround either. By the time a microtask runs, the popup has already derived which elements count as outside it from the slot it held when it opened, so Escape from the field the caret is in stops dismissing it, and so does clicking that field.

Fixes #7087

## Solution

The open-time capture now returns early when the store already holds a connected disclosure element outside the dialog that this capture did not itself write. Recording the focused element stays as the fallback for when nothing named an opener.

Telling those apart needs one bit the store cannot supply: `setDisclosureElement` is a plain `setState` that skips writes which don't change the value, and a mounted disclosure button claims the slot as soon as it renders, so an explicit assignment and a leftover look identical. `Dialog` therefore marks the elements it captured itself, in a module-level `WeakSet`. The mark travels with the element rather than with a dialog or a store, because both a dialog and the store it derives are replaced when its component remounts while the store the application owns keeps the value.

Because a mounted trigger's occupancy is indistinguishable from an explicit assignment, it counts as naming the opener too. Showing content programmatically while a `PopoverDisclosure` or similar component is rendered now keeps that component as the opener rather than the element that happened to have focus. That consequence reaches applications that never call `setDisclosureElement`, so it gets its own changeset alongside the one for the fix.

`setDisclosureElement` also gained documentation saying a programmatic show should name its trigger, which the issue asked for.

## Reproduction

`app/src/sandbox/popover-disclosure-7087/` is a note editor where "Formatting" and "Quick format" both open the same popup. Typing `*` in a field opens it too and names "Formatting" as the opener, while "Suggest formatting" and "Summarize note" open a popup naming nobody.

Three tests fail on `main` and pass with the fix. Each starts by opening the popup from "Quick format", so the store already holds the other button when the shortcut runs and the explicit assignment has to overwrite it rather than inherit it. They fail because the named button reports `aria-expanded="false"`, because dismissing the popup from inside returns focus to the field instead of the named button, and because an open that names nothing hands the popup the focused field instead of the button already in the store.

Four more cover behavior the fix has to preserve, and each was red-checked against a wrong implementation:

- Escape from the field the caret is in still dismisses the popup. This is the only consequence derived from the store when the popup opens rather than read live, so it fails against a fix that captures and then restores the value afterwards.
- An open that captured nothing, because focus was on the body, must not leave the previous fallback looking like somebody's choice. This fails against a fix that forgets its own capture on the paths that record nothing.
- With no button mounted at all, the popup still falls back to the focused field, and a later open from another field replaces that fallback. This fails against a fix that simply keeps any opener it finds.
- A popup whose own store is derived by a wrapper component still replaces the fallback after that wrapper remounts. This fails against a fix that remembers the capture per store object, because the derived store is replaced while the application's store keeps the value.

Both popups use `unmountOnHide`, so every case is exercised across a remount.

Verified in happy-dom and under Chromium, Firefox and WebKit.

## Preview

Typing `*` in the note, with the caret left where it was. On `main` the "Formatting" button is not styled as expanded and the popup opens against the note field, over the buttons below it. With the fix the button is styled as expanded and the popup opens against it. The fixture styles the buttons from `[aria-expanded]`, so the highlight is the attribute.

Before:

The note field is focused and contains a typed asterisk. The Formatting button is unstyled, and the formatting popup is anchored to the note field, overlapping the buttons underneath it.

After:

The note field is still focused and contains the typed asterisk. The Formatting button is highlighted because it now reports itself as expanded, and the formatting popup is anchored to that button.

## Workaround

Until this ships, an application can put the intended opener back after the capture and force the popup to re-derive what it computed from the overwritten value.

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7087 is fixed.
const [treeSnapshotKey, setTreeSnapshotKey] = useState(0);

const showFormatting = (trigger) => {
  const opener = trigger ?? popover.getState().disclosureElement;
  popover.show();
  queueMicrotask(() => {
    if (opener?.isConnected) {
      popover.setDisclosureElement(opener);
    }
    setTreeSnapshotKey((key) => key + 1);
  });
};

<Ariakit.Popover store={popover} unstable_treeSnapshotKey={treeSnapshotKey} />;
```

Both halves are load-bearing. Removing only the key bump turned two of the tests in this PR red: Escape from the field the caret is in stopped dismissing the popup, and clicking that field stopped dismissing it too. Both consequences are derived once from `disclosureElement` when the popup opens, so restoring the value alone does not repair them.

- `unstable_treeSnapshotKey` is a private, unstable prop. It is typed on `DialogOptions` but marked `@private`, so it does not appear in the reference documentation and can change or disappear in any release. It is the only consumer-reachable way to make the popup recompute its outside elements.
- The microtask assumes `show()` runs in a discrete React event, such as a click or keydown handler, so React's synchronous commit lands before the restore. Reached from a `setTimeout`, a promise continuation after an `await`, or `startTransition`, the restore runs first and the capture overwrites it again, leaving the original bug in place with no visible sign the workaround was applied.
- Every open costs an extra render, and rebuilds both the tree snapshot and the outside marks. Only the non-modal case was validated. On a modal dialog this tears down and re-applies `aria-hidden` and `inert` across the outside tree.
- Validated on a non-modal, portaled `Popover` with `autoFocusOnShow={false}` and no enter or exit animations. With an animated popup the open and the portal commit are separated, so the restore can land on either side of the first tree effect. Re-verify there.
- For one microtask the store holds the focused field, so anything reading `disclosureElement` synchronously in that window sees the wrong opener, and `aria-expanded` lands one microtask plus one render late.
- The popup anchors to the field for that microtask and then re-anchors to the button, because `anchorElement` follows `disclosureElement` when no `PopoverAnchor` is set. Tab order follows it too.
- It only covers `show()` calls routed through the helper. Any other `store.show()` in the application keeps the original behavior, and clicks on a `PopoverDisclosure` do not need it.
- Calling it without a trigger re-asserts whatever the store already held, without checking that the element is still a mounted disclosure.
- Once the button really is the opener, closing the popup returns focus to it rather than to where the user was.

## Review gate reassessment

<details>
<summary>Reproduction fixture, cycle 3: continue the direction</summary>

**Assessment**

The issue is labeled `bug` and `a11y`, reproduces on `main`, and covers a stable public API. The reproduction is a sandbox plus two suites, well under the median size for a Popover or Dialog behavior change in this repository, and it covers three separable contract facets. The three finding-bearing rounds were surface polish on a sound artifact rather than a flawed direction: of ten findings, five were artifacts of remedies applied in a previous round and none challenged the fixture's shape or contract. Measurement corrected one reviewer premise: the last mounted `PopoverDisclosure`, not the first, claims the store, because the self-heal effect closes over the render-time value.

**Decision**

Continue. Apply the round-3 findings in reduced form, and treat further Low-severity wording, naming, or fixture-realism findings as declined non-goals unless they make an assertion wrong or a test unreliable.

</details>

<details>
<summary>Reproduction fixture, cycle 3 of the reset count: add the Escape coverage, then commit</summary>

**Assessment**

Cycles 2 and 3 changed comments only, but cycle 3 found a genuine gap rather than noise: Escape from the field the caret is in is the one consequence derived from the store when the popup opens rather than read live at event time, and it is the axis the issue measured the deferred-assignment workaround losing. A fix shaped as "capture, then restore afterwards" would have kept every other assertion green while silently breaking it.

**Decision**

Add that one test to both suites, change nothing else, then take the clean pass. Nothing in the fixture is removable: every element is load-bearing for exactly one assertion.

</details>

<details>
<summary>Library fix, cycle 3: freeze the mechanism and commit</summary>

**Assessment**

The fix is one module-level set and about ten lines inside an effect that already existed. It buys a family-wide accessibility correctness fix across `Popover`, `Menu`, `SelectPopover`, `ComboboxPopover`, `Hovercard` and `Tooltip`, for a defect with no behavior-preserving userland workaround. The two defects found in earlier rounds were both storage-lifetime bugs of the same memo, a component ref that `unmountOnHide` destroyed and a premature delete, not errors in the discrimination rule, which has survived an exhaustive audit of all ten exits of the effect. Finding severity across the three rounds went High, then Medium, then copy only.

Five simpler alternatives were weighed. The only materially simpler one, never overwriting a connected outside element, makes the fallback permanent and breaks two tests in this PR, trading the reported bug for a worse one. Clearing the state on hide has a wider blast radius and still needs the same bit. The rest add public state or a new subsystem.

**Decision**

Freeze the mechanism. Apply copy edits only: the fixture header, the changeset title, the code fence language, and the closing sentence, which also gained the positioning consequence that earlier rounds deferred here. Accepted limitation: an element that any dialog captured as a fallback stays marked for as long as it lives, so explicitly naming that same element afterwards is still overwritten. Marks cannot be dropped, because two dialogs can share an element and one clearing it would leave the other reading its own fallback as a name. This is never worse than before the fix, which dropped the assignment on every open, and it is tracked in #7095.

</details>

<details>
<summary>Review response, cycle 2: keep the direction, move the provenance onto the element</summary>

**Assessment**

Review found that keying the capture record by store object regressed against `main`: an application wrapper that derives its own store loses the record whenever it remounts, while the application's store keeps the value, so the stale fallback reads as a deliberate name and stays that way permanently. That is worse than the bug this PR fixes, which at least re-derived the opener on every open.

Three defects have now been found in this provenance machinery, so the alternative of removing it was weighed again rather than hardened a third time. Removing it means never overwriting a connected outside element, which makes the fallback permanent for every application rather than only for the wrapper shape, and it breaks three tests in this PR. The element is the only identity available today that survives both `unmountOnHide` and a wrapper remount, because the store exposes no parent chain.

**Decision**

Track the mark on the element. This removes the derived key and its effect dependency rather than adding machinery, and it fixes a strong reference the previous `WeakMap` held on captured elements. Marks are never dropped, since two dialogs can share an element and one clearing it would leave the other reading its own fallback as a name. The cost is recorded above and tracked in #7095.

</details>


.changeset/200-dialog-assigned-disclosure-element.md
```diff
@@ -0,0 +1,17 @@
+---
+"@ariakit/components": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+`Dialog` keeps the disclosure element it was given
```
diegohaz: This should be a single sentence changeset. If this covers more than one change, just split it into separate changesets.

packages/ariakit-components/src/disclosure/disclosure-store.ts
```diff
@@ -195,8 +195,22 @@ export interface DisclosureStoreFunctions extends Pick<
   /**
    * Sets the `disclosureElement` state.
    *
+   * This is the element reported as expanded while the content is open. Content
+   * built on [`Dialog`](https://ariakit.com/reference/dialog), such as
+   * [`Popover`](https://ariakit.com/reference/popover) and
+   * [`Menu`](https://ariakit.com/reference/menu), also derives the elements that
+   * count as outside it, and where focus returns on close, from it.
```
diegohaz: This explanation is too complicated. I'm the author of the library and I can't even understand it.

.changeset/200-dialog-assigned-disclosure-element.md
diegohaz: Fixed in [`ee6f36c`](https://github.com/ariakit/ariakit/commit/ee6f36ce5): split into two single-sentence entries.

It did cover two changes. One is the fix itself. The other is that a mounted trigger now keeps the opener on a programmatic open, which reaches applications that never call `setDisclosureElement`, so it got its own entry at `200-` rather than being folded into the fix.

I also renamed the fix entry to `300-` and dropped `@ariakit/components` from its frontmatter, since the only change there is a doc comment and the behavior ships through `@ariakit/react-components`.

packages/ariakit-components/src/disclosure/disclosure-store.ts
diegohaz: Fixed in [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51b): cut to one sentence, "When showing the content programmatically, set this to the element that should act as its trigger."

Everything else was explaining the fallback machinery rather than telling the reader what to do, and the `@example` below it already shows the ordering.

diegohaz:
**Fixed**

- [`ee6f36c`](https://github.com/ariakit/ariakit/commit/ee6f36ce5) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741351978)): split the changeset into two single-sentence entries, one for the fix and one for a mounted trigger now keeping the opener on a programmatic open. Renamed the fix entry to `300-` and dropped `@ariakit/components` from its frontmatter.
- [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51b) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741352241)): cut the `setDisclosureElement` documentation to one sentence.

No code behavior changed, so the existing test coverage and CI results still stand.

app/src/sandbox/popover-disclosure-7087/test-browser.ts
```diff
@@ -0,0 +1,163 @@
+import type { query } from "@ariakit/test/playwright";
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+
+type Query = ReturnType<typeof query>;
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // The first PopoverDisclosure claims the store at mount, so opening from
+  // "Quick format" first moves the incumbent off that default and makes each
+  // test's expected opener a genuine outcome rather than the mount default.
```
diegohaz: This premise is inverted, and it changes what one of the tests proves. `Quick format` is the **last** `PopoverDisclosure` in the fixture, and it is the one that owns the store at mount, not `Formatting`.

`useDisclosure`'s self-heal effect closes over the render-time `disclosureElement`, which is still `null` for both buttons on the first commit, so each one reassigns the store and the later one wins. The PR description already records this ("the last mounted `PopoverDisclosure`, not the first, claims the store"), but the comment kept the earlier premise.

I measured it on `96cf1f5` by opening the popup from `Suggest formatting` with no prior interaction, and got `Formatting=false QuickFormat=true` in both happy-dom and Chromium.

The consequence is that `openFromQuickFormatAndClose` does not move the incumbent off the mount default, because `Quick format` already is that default. In `keeps a mounted button as the opener when the open names none`, the asserted opener is therefore exactly the mount default, which is the outcome this comment claims to have ruled out. The test is still a real regression, since on `main` the capture replaces the incumbent with `Suggest formatting`, but it does not show that an incumbent established by a real interaction survives a later open.

Swapping the two buttons in `index.react.tsx` makes the sentence true and strengthens that test, because `Formatting` becomes the mount default and `Quick format` becomes a genuinely established incumbent. I applied that swap locally: the mount default flipped to `Formatting=true QuickFormat=false`, and all six happy-dom tests and all six Chromium tests still passed.

`app/src/sandbox/popover-disclosure-7087/test.ts:4` repeats the same sentence and needs the same correction.

There is no consumer-facing impact here; the finding is about the guarantee the regression suite documents for itself.

**Assessment**

- Severity: Low (verified, 90% confidence), because the suite still fails on `main` and passes with the fix, so only one assertion's strength and the comment's accuracy are affected.
- Priority: P1 (estimate, 80% confidence), because both suites repeat the wrong premise and the correct measurement is already in the PR description, so it should not ship as the fixture's documented rationale.
- Breakage likelihood: Likely (verified, 95% confidence) to mislead a future edit of these tests, since the sentence is the stated reason the helper exists. No runtime behavior is at risk.
- Fix effort: Easy (verified, 95% confidence), because swapping the two `PopoverDisclosure` blocks and correcting one sentence per suite is enough; I ran both suites after the swap.

packages/ariakit-react-components/src/hovercard/__hovercard-trigger.tsx
```diff
@@ -77,8 +77,8 @@ export const useHovercardTrigger = createHook<TagName, HovercardTriggerOptions>(
         }
         store.show();
         queueMicrotask(() => {
-          // We need to set the trigger as the hovercard disclosure only when the
-          // hovercard is shown so it isn't replaced by the dialog component.
+          // Name the trigger again once the open has settled, in case anything
+          // else claimed the store while the hovercard was being shown.
           store.setDisclosureElement(element);
         });
```
diegohaz, line 86: The comment this replaces named the exact mechanism the `queueMicrotask` defends against: the dialog's open-time capture replacing the trigger. This PR removes that mechanism on the ordinary path, and the new text points at an unnamed third party instead of at the capture that can still overwrite the trigger, so it no longer tells a reader why the deferral has to stay.

After the fix, the capture only overwrites a named trigger when `store.getState().disclosureElement === previousCapture`, which for a hovercard means this store's dialog previously captured that same element as a fallback and the `setDisclosureElement` call on line 68 no-ops the equal write. That narrow path is the only remaining reason the microtask exists, and it is precisely what someone needs to know before deleting it.

I removed the microtask on `96cf1f5` and the full Vitest suite still passed (256 files, 1720 tests), so nothing here is protected by tests either way. That makes an accurate comment the only guard.

Either name the remaining mechanism, or drop the deferral if that case is not worth defending. If it stays:

```suggestion
        queueMicrotask(() => {
          // The dialog's open-time capture can still replace the trigger when
          // it's the element that dialog last captured as a fallback, because
          // the store no-ops the equal write above.
          // https://github.com/ariakit/ariakit/issues/7087
          store.setDisclosureElement(element);
        });
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because runtime behavior is unchanged and only the maintainability of a non-obvious deferral is affected.
- Priority: P2 (estimate, 80% confidence), because it is cheap to correct now and the PR is what invalidated the original explanation.
- Breakage likelihood: Possible (estimate, 70% confidence). A future cleanup that trusts the vague comment and deletes the microtask would reintroduce the trigger being overwritten in the equal-capture case, and no test would catch it.
- Fix effort: Easy (verified, 90% confidence), because it is a comment edit in one place.

.changeset/200-dialog-mounted-disclosure-opener.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Changed [`Dialog`](https://ariakit.com/reference/dialog) and components built on it, such as [`Popover`](https://ariakit.com/reference/popover) and [`Menu`](https://ariakit.com/reference/menu), to keep a mounted [`PopoverDisclosure`](https://ariakit.com/reference/popover-disclosure) or similar trigger as the opener when shown programmatically, instead of the element that happened to have focus.
```
diegohaz: This entry describes the trigger being kept "as the opener", but for the applications it is aimed at, the ones that never call `setDisclosureElement`, the opener is not a label. It decides where the popup is positioned and where focus goes when the popup closes, and neither consequence is visible from this sentence.

`anchorElement` follows `disclosureElement` unless an anchor was set explicitly (`packages/ariakit-components/src/popover/popover-store.ts:75-82`), and this PR's own `returns focus to the named button when dismissed from inside` test shows focus restoration following it too.

So an application that shows a popup from a keyboard shortcut while a `PopoverDisclosure` is mounted now gets the popup anchored to that button instead of the caret, and focus landing on that button instead of back where the user was typing, without changing any of its own code. That is the part a reader scanning the changelog needs in order to recognize the change in their own app, and the PR description already spells it out under limitations.

```suggestion
Changed [`Dialog`](https://ariakit.com/reference/dialog) and components built on it, such as [`Popover`](https://ariakit.com/reference/popover) and [`Menu`](https://ariakit.com/reference/menu), to keep a mounted [`PopoverDisclosure`](https://ariakit.com/reference/popover-disclosure) or similar trigger as the opener when shown programmatically, instead of the element that happened to have focus, which also means the content is positioned against that trigger and returns focus to it when it closes.
```

**Assessment**

- Severity: Medium (estimate, 75% confidence), because focus landing somewhere the user did not leave it is disruptive for keyboard and screen reader users, and the changelog is where an affected app would find the cause.
- Priority: P2 (estimate, 80% confidence), because the behavior itself is intended and only the description is incomplete.
- Breakage likelihood: Possible (estimate, 70% confidence). It needs an app that shows a `Popover`, `Menu`, or `Select` popup programmatically while its trigger is mounted, which is the same scenario the entry is written for.
- Fix effort: Easy (verified, 90% confidence), because it is one trailing clause in the existing sentence.

diegohaz (commented):
## Review summary

Initial review. Baseline is the merge base [`9aa0b05`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) to head [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66), expanded into the surrounding dialog, disclosure, popover, menu, select and hovercard code. There is no prior review from this workflow on the PR, so nothing was reviewed incrementally.

- **Codex, `dialog.tsx:293`, share fallback provenance across synchronized stores.** Confirmed, and reachable through a smaller setup than the comment describes. Replied on the thread with a reproduction and a base comparison. This is the one finding that should block merge.
- **Codex, `dialog.tsx:324-326`, clear the Safari mousedown fallback before returning.** The mechanical observation holds, the consequence does not belong to this change. Declined as a non-goal, with the reasoning on the thread. No follow-up issue, because I could not establish a case where the current behavior yields a wrong opener rather than the intended one.

The two older `diegohaz` threads are resolved and were not re-adjudicated.

### Findings

The blocking one is the provenance defect above. `captureKey` is `storeProp || context || store`, which is stable only when the application hands its own store straight to the component, as `usePopover` does. Put an ordinary `usePopoverStore({ store })` in a wrapper that can remount and the key changes on every mount, so `previousCapture` is always missing and the previous fallback passes `hasNamedDisclosure()` forever. The failure is not a one-open glitch: the popup keeps anchoring to the stale element and returning focus there regardless of where the user opens it from, which is worse than the bug being fixed, since the base at least re-derived the opener each time. Details, reproduction and assessment are on the Codex thread.

Three smaller findings are inline: an inverted premise in the test-suite comments about which `PopoverDisclosure` owns the store at mount, a comment in `__hovercard-trigger.tsx` that no longer names the mechanism it guards, and a changeset entry that omits the positioning and focus-restoration consequences of the behavior change.

Everything else about the approach held up. Recording the capture per store rather than per component is the right response to `unmountOnHide`, the discrimination rule survives all the exits of the effect, and the fix does what it claims on the shapes the sandbox covers.

### What I ran

All in a dedicated worktree at [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66), restored to a clean tree afterwards.

- `pnpm test`: 256 files, 1720 tests, all passing.
- `pnpm test-react18`: 189 files, 986 tests, all passing. Worth checking because the fix turns on layout effect ordering.
- `pnpm tsc`, `pnpm lint`: clean.
- `pnpm --filter app test-chrome popover-disclosure-7087` against this checkout's own dev servers: 6 passing.
- Red check: reverting only `dialog.tsx` to the base leaves exactly three failures, each for the reason the PR claims. The regression proof holds.
- Scratch sandbox for the provenance defect, run against both base and head. Numbers on the Codex thread.
- Measured which `PopoverDisclosure` owns the store at mount, in happy-dom and Chromium.
- Removed the hovercard `queueMicrotask` and reran the full suite.
- Verified `anchorElement` follows `disclosureElement` in `popover-store.ts:75-82`, and that `MenuButton` and `Select` both route through `usePopoverDisclosure`, so the `200-` entry's claim about `Menu` is accurate and `Select` is affected the same way.
- Checked all five changeset reference links, including that the deployed `use-disclosure-store` page really exposes a `#setdisclosureelement` anchor.
- Inspected the generated reference metadata for `setDisclosureElement`: description, one `examples` entry and the live example all extract correctly, and the unfenced `@example` matches the convention already used by `setOpen` in the same file.

### Candidates reviewed and not published

- **Equal-write blind spot.** When an application names the exact element this dialog last captured as a fallback, the store no-ops the equal write, `previousCapture` matches, and the capture replaces it anyway. Confirmed, and already recorded as an accepted limitation in the PR description. Declined non-goal: it is strictly better than the base, which dropped the assignment on every path. Severity: Low (estimate, 80% confidence). Priority: P2 (estimate, 75% confidence). Breakage likelihood: Rare (estimate, 80% confidence). Fix effort: Hard (estimate, 70% confidence), because it needs the store to distinguish an explicit write from an equal one.
- **`Suggest formatting` has no `tabIndex={0}`.** On Safari a click may not focus it, so `keeps a mounted button as the opener when the open names none` reaches the `lastMousedownRef` branch there rather than the focused-element branch. Declined non-goal: the assertion holds either way and the test is still red on `main` under both branches. `Summarize note` correctly carries `tabIndex={0}`, since its `blur()` depends on having been focused.
- **`capturedDisclosures` is typed `WeakMap<object, Element>`** while every value written is an `HTMLElement`. Declined non-goal, no behavioral or type-safety consequence.
- **Retention.** The map holds one element strongly per store until superseded. Invalid as a finding: the store's own `disclosureElement` already retains the same element for the same lifetime.
- **Whether the `200-` entry should carry `**BREAKING**`.** Reviewed and agreed with the current classification; no consumer has to change code. What it is missing is the description of the consequences, which is the inline comment.

### Not fully validated

- Only the Chromium project ran locally. Firefox, WebKit and the mobile projects are on CI and on the author's reported runs.
- The legacy `website/build-pages` generator was covered only by its focused tests inside `pnpm test`, plus the modern app's generated output; I did not run `pnpm build-website-lite`. The JSDoc change reuses the exact shape of `setOpen` in the same interface, so the risk is low.
- No follow-up issues were registered. The provenance defect belongs in this PR rather than after it, and nothing else here is separable work the project would pursue independently.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -281,15 +287,48 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
     }, [domReady]);
   }
 
+  // The store the application owns outlives this component, so it keys the
+  // captured element above. Falls back to the local store when the dialog owns
+  // it, which is also the only case where nothing can unmount it separately.
+  const captureKey = storeProp || context || store;
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and reachable through a smaller setup than the one described here. Sibling dialogs are not required: one wrapper component that derives its own store and remounts is enough, because every remount produces a new `captureKey` while the application-owned store keeps `disclosureElement`.

I reproduced it on [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66) with a scratch sandbox built around the wrapper shape an application would plausibly write:

```tsx
function MyPopover(props: { store: Ariakit.PopoverStore }) {
  const popover = Ariakit.usePopoverStore({ store: props.store });
  return <Ariakit.Popover store={popover} autoFocusOnShow={false} portal>...</Ariakit.Popover>;
}

// The app remounts its own wrapper between opens.
<MyPopover key={wrapperKey} store={popover} />
```

Opening from the note captures it as a fallback, then remounting the wrapper and reopening from the title reads `disclosureElement` through a `captureKey` that has no record, so the stale note passes `hasNamedDisclosure()` and is preserved:

- Base [`9aa0b05`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161): `first=[opener=Note] second=[opener=Title]`
- Head [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66): `first=[opener=Note] second=[opener=Note]`

I also reproduced the two-synchronized-stores shape from this comment, with the same result once the dialog holding the record unmounts. My first attempt at that one looked clean only because I was reading focus after Escape, which with `autoFocusOnShow={false}` never left the field and so never reflected `disclosureElement`; reading the state directly showed the stale opener.

This matters because the fallback does not just go stale for one open, it becomes permanent. Nothing can dislodge it: every later open finds a connected outside element with no record, so the popup keeps anchoring to the note and returning focus there no matter which field the user opens it from. That is a worse outcome than the bug this PR fixes, which at least re-derived the opener on each open.

The provenance really does need to follow the synchronized state rather than the nearest store object. `packages/ariakit-react-components/src/dialog/dialog.tsx:293` picks `storeProp || context || store`, which is stable only when the application passes its own store straight through, as `usePopover` does. Any intermediate `useDialogStore({ store })` or `usePopoverStore({ store })` in a component that can unmount reintroduces the churn the `WeakMap` was added to survive.

Worth noting for whichever direction you take: the sandbox in this PR cannot catch this, because it passes `usePopoverStore()` results directly to `Popover`, which is the one shape where the key is stable. A regression for this needs a fixture with a derived store in a remounting component.

**Assessment**

- Severity: High (verified, 85% confidence), because the opener becomes permanently wrong, which misplaces the popup, sends focus to an unrelated field on close, and reports `aria-expanded` on the wrong element.
- Priority: P1 (estimate, 85% confidence), because it is a regression this PR introduces against the base, in the same area the PR is correcting.
- Breakage likelihood: Possible (verified, 75% confidence). It needs an intermediate derived store in a component that remounts, which is not the common shape but is ordinary composition and is not documented as unsupported.
- Fix effort: Moderate (estimate, 60% confidence), because the record has to key off something shared by the whole sync chain rather than the immediate store object, and that identity is not exposed today.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -281,15 +287,48 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+      // The disclosure element can't be inside the dialog.
+      if (dialog && contains(dialog, disclosureElement)) return false;
+      return true;
+    };
+    // Only forget the previous capture once something has superseded it. The
+    // paths below can leave it in the store without capturing anything, and
+    // dropping the record there would make it look named on the next open.
+    if (hasNamedDisclosure()) {
+      capturedDisclosures.delete(captureKey);
+      return;
```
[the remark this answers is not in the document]
diegohaz: The mechanical part is accurate: the new early return does leave `lastMousedownRef` set. The consequence drawn from it is not specific to this change, so I am not treating it as something to fix here.

On the base [`9aa0b05`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161), `lastMousedownRef.current = null` runs only inside the `BODY` branch. Every other exit already leaves it set, including `if (!activeElement) return`, the `contains(dialog, activeElement)` guard, and the ordinary successful capture that ends the effect. So a pending mousedown target surviving into a later open cycle is the existing behavior of this ref on the common path, not something the named-disclosure return introduces.

It is also the intended semantics rather than a leak. The ref is maintained by a document-level `mousedown` listener that runs on Safari for every press, so it holds whatever the user last pressed. When a later open finds `activeElement` on `BODY`, using the last pressed element is the fallback that branch exists to provide, and it is still gated on `isConnected`, `isFocusable`, and not being inside the dialog.

Disposition: declined as a non-goal for this PR. Clearing the ref on one of several equivalent exits would not establish any property the others break, and tightening the ref's lifetime across all of them is a separate change to pre-existing Safari behavior with no reported symptom behind it. I did not open a follow-up issue, since I could not establish a concrete case where the current behavior produces a wrong opener rather than the intended one.

packages/ariakit-react-components/src/dialog/dialog.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`3499890`](https://github.com/ariakit/ariakit/commit/3499890f9): the mark now belongs to the element rather than to a store object.

I reproduced your wrapper shape before changing anything. The suggestions popup in the sandbox now lives behind a `SuggestionsPopover` that derives its own store with `usePopoverStore({ store })`, plus a "Reload suggestions" button that remounts it, and the new test `replaces the fallback after the popup's own store is replaced` is red on the previous implementation in both happy-dom and Chromium.

On the direction: this is the third defect in the same provenance machinery, so I weighed removing it rather than hardening it again. Removing it means never overwriting a connected outside element, which makes the fallback permanent for every application instead of only the wrapper shape, and it turns three tests in this PR red. The element is the only identity that survives both `unmountOnHide` and a wrapper remount, since the store exposes no parent chain. Keying on it also removed `captureKey` and its effect dependency, so the fix is smaller than before, and it drops a strong reference the `WeakMap` held on captured elements.

One thing worth flagging, because my first attempt at this got it wrong. I initially unmarked the element a capture replaced, so that naming it later would still count. Review caught that the delete was decided from one store's value while mutating a global set, so two dialogs holding the same captured element meant whichever moved on first erased a mark the other still needed. Marks are never dropped now.

The cost is that a mark is permanent for the element's lifetime, so explicitly naming an element that any dialog once captured is still overwritten. That is never worse than `main`, which dropped the assignment on every open, and it is tracked in #7095.

app/src/sandbox/popover-disclosure-7087/test-browser.ts
diegohaz: Fixed in [`3499890`](https://github.com/ariakit/ariakit/commit/3499890f9): swapped the two buttons so "Formatting" is last, and corrected the sentence in both suites to say the last one claims the store.

I measured the premise rather than taking the swap on trust. Clicking "Suggest formatting" with no prior interaction now reports `Formatting=true QuickFormat=false`, so "Formatting" is the mount default and `openFromQuickFormatAndClose` genuinely moves the incumbent.

The helper also asserts it now, so the premise is pinned rather than only described: it checks that "Quick format" reports `aria-expanded="true"` while the popup is open, before closing it.

packages/ariakit-react-components/src/hovercard/__hovercard-trigger.tsx
diegohaz: Fixed in [`d284f9d`](https://github.com/ariakit/ariakit/commit/d284f9d53).

I took your mechanism but not the wording, because the implementation changed underneath it in the same push. Provenance is now a mark on the element rather than a per-store record, so the trigger can be treated as a fallback because some dialog captured it, not necessarily this one, and the equal write is no longer what decides it. The comment now says the mark belongs to the element and is never dropped, so naming the trigger before the open cannot rescue it and the re-name has to land after the capture.

.changeset/200-dialog-mounted-disclosure-opener.md
diegohaz: Fixed in [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e56): applied your suggestion verbatim.

One thing to flag rather than silently change: "positioned against that trigger" holds for the popover-based components through the `anchorElement` sync you cite, but not for a plain `Dialog`, which has no positioning and is the first API named in the sentence. I left the text as you wrote it rather than rescoping a sentence you have already asked to keep short. Happy to narrow it to `Popover` and `Menu` if you would rather it be exact.

diegohaz:
**Fixed**

- [`3499890`](https://github.com/ariakit/ariakit/commit/3499890f9) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741561930)): moved the capture provenance onto the element, so it survives a wrapper component remounting the store it derives. Reproduced first, and the new test `replaces the fallback after the popup's own store is replaced` is red on the previous implementation in happy-dom and Chromium. This removed `captureKey` rather than adding machinery.
- [`3499890`](https://github.com/ariakit/ariakit/commit/3499890f9) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741562122)): swapped the two disclosure buttons and corrected both suites, which said the first claims the store rather than the last. Measured the premise, and the helper now asserts it instead of only describing it.
- [`d284f9d`](https://github.com/ariakit/ariakit/commit/d284f9d53) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741562400)): named the mechanism the hovercard trigger defers around, adjusted for the element-scoped mark.
- [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e56) ([reply](https://github.com/ariakit/ariakit/pull/7090#discussion_r3741562663)): added the positioning and focus consequences to the mounted-opener changeset.

**Tracked**

- [#7095](https://github.com/ariakit/ariakit/issues/7095): a mark is permanent for the element's lifetime, so explicitly naming an element that any dialog once captured is still overwritten. Never worse than `main`, which dropped the assignment on every open, and it needs store-level provenance that does not exist today.

**Indeterminate**

- The Safari `lastMousedownRef` thread already carries a declining reply and no code change, so it is left alone. I cannot rule out a resolved-to-unresolved transition since that reply, because I have no observation of the thread between then and now.

Verified: sandbox suite 7 passed, browser suite 21 passed across Chromium, Firefox and WebKit, full suite 256 files and 1721 tests, lint and tsc clean.

packages/ariakit-react-components/src/dialog/dialog.tsx
```diff
@@ -282,14 +292,33 @@ export const useDialog = createHook<TagName, DialogOptions>(function useDialog({
...
+      if (capturedDisclosures.has(disclosureElement)) return false;
+      if (!disclosureElement.isConnected) return false;
+      // The disclosure element can't be inside the dialog.
+      if (dialog && contains(dialog, disclosureElement)) return false;
+      return true;
+    };
+    if (hasNamedDisclosure()) return;
+    const captureDisclosure = (element: HTMLElement) => {
+      capturedDisclosures.add(element);
+      store.setDisclosureElement(element);
```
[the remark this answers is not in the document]
diegohaz: Confirmed against the current head [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e5616213c3e81fa91a928eef611088d4b3), and already registered as a follow-up rather than something to fix here: [#7095](https://github.com/ariakit/ariakit/issues/7095).

The mechanics are exactly as described. `capturedDisclosures` is a module-level `WeakSet` whose marks are never removed, so once any dialog has captured an element as a fallback, `hasNamedDisclosure()` returns `false` for that element forever and the capture below replaces it with whatever holds focus. The narrowing is that an explicit assignment is only lost when the named element is not also the focused one, since otherwise the capture re-writes the same value.

The reason it is not being fixed in this PR is that neither remedy suggested here is reachable today.

Per-store provenance is what the previous revision of this PR did, keyed by `storeProp || context || store`. It has to be abandoned because a wrapper component that derives its own store with `usePopoverStore({ store })` gets a new key on every remount, which made the fallback permanently sticky instead. That regression is what the `replaces the fallback after the popup's own store is replaced` test now guards, and I verified it is red against the previous revision and green here.

Letting an explicit `setDisclosureElement` supersede the mark needs `Dialog` to observe that a write was deliberate. It cannot: `setState` skips writes that do not change the value, so re-naming the element already in the slot leaves no trace, and there is no identity that both scopes a record to one store tree and survives the remount above. `createStore` and `mergeStore` expose no parent chain, and `getKeys` is `Object.keys`, so a symbol key would not sync and the record would have to become a public `DisclosureStoreState` member.

Dropping the mark instead is unsafe for the reason the code comment gives: two dialogs can hold the same captured element, and whichever moved on first would erase a mark the other still needs, leaving it to read its own stale fallback as a deliberate name.

The suggested regression test belongs with the fix in [#7095](https://github.com/ariakit/ariakit/issues/7095) rather than here, since it would encode the current behavior as expected.

Disposition: registered follow-up, tracked in [#7095](https://github.com/ariakit/ariakit/issues/7095). It is narrower than the bug this PR fixes and never worse than the base, which overwrote the assignment unconditionally on every open.

**Assessment**

- Severity: Medium (verified, 85% confidence), because the affected assignment is silently ignored, with no diagnostic and no way for an application to see the history it depends on.
- Priority: P2 (estimate, 80% confidence), because the correct fix changes the store rather than `Dialog`, and shipping this PR strictly improves the current behavior.
- Breakage likelihood: Unlikely (estimate, 70% confidence). It needs the same element to serve first as a fallback capture and later as a name while something else holds focus; in the common shapes the named element is also the focused one, so the capture rewrites the same value.
- Fix effort: Hard (verified, 80% confidence), because it requires store-level provenance that does not exist today, as spelled out in [#7095](https://github.com/ariakit/ariakit/issues/7095).

diegohaz (commented):
## Re-review summary

Re-review. The previous review from this workflow marked head [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66), so that is the incremental baseline; the current head is [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e5616213c3e81fa91a928eef611088d4b3). Three commits since then: [`3499890`](https://github.com/ariakit/ariakit/commit/3499890f9), [`d284f9d`](https://github.com/ariakit/ariakit/commit/d284f9d53) and [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e5616213c3e81fa91a928eef611088d4b3). I also re-checked the surrounding capture logic rather than only the changed lines, because the mechanism was replaced rather than patched.

No new findings. All four findings from the previous round are fixed, and the replacement mechanism holds up under the checks below.

### Previous findings, verified fixed

- **Provenance keyed by the store.** Resolved by replacing the `WeakMap<object, Element>` keyed by `storeProp || context || store` with a module-level `WeakSet<Element>`. The mark now travels with the element, which is the identity that survives both `unmountOnHide` and a wrapper remounting the store it derives. This is a smaller mechanism than the one it replaces: `captureKey` and its effect dependency are gone, the effect is back to `[store, open]`, and the strong reference the old `WeakMap` held on captured elements is gone with it. I re-ran my own reproduction from the last round and it now behaves like the base.
- **Regression coverage for it.** `replaces the fallback after the popup's own store is replaced` exercises exactly that shape through a `SuggestionsPopover` wrapper that derives its store, plus a `Reload suggestions` button that remounts it. Red-checked: with `dialog.tsx` reverted to [`96cf1f5`](https://github.com/ariakit/ariakit/commit/96cf1f51bcc1f78b64e7ccac41d576ed9878fb66) it is the one failure out of seven, and it passes here.
- **Inverted premise in the suite comments.** Fixed, and stronger than what I suggested. The buttons were reordered so `Formatting` is the mount default, the comment now says the last `PopoverDisclosure` to mount claims the store, and the helper additionally pins `aria-expanded="true"` on `Quick format` so the starting incumbent is asserted rather than assumed.
- **Hovercard comment.** Now names the actual mechanism, that the mark belongs to the element and is never dropped, so naming the trigger before `show()` cannot rescue it. That is accurate against the new implementation.
- **Changeset consequences.** The `200-` entry now names positioning and focus restoration, still as one sentence.

### Unresolved thread validated in this run

- **CodeRabbit, `dialog.tsx:308-320`, preserve a later explicit assignment for a previously captured element.** Confirmed as accurate, dispositioned as a registered follow-up rather than current-PR work, and replied on the thread. It is tracked in [#7095](https://github.com/ariakit/ariakit/issues/7095), whose write-up already establishes why neither suggested remedy is reachable today and that the behavior exists only on this branch. Assessment is on the thread.

The seven older threads are resolved and were not re-adjudicated.

### What I ran

At [`c91a38e`](https://github.com/ariakit/ariakit/commit/c91a38e5616213c3e81fa91a928eef611088d4b3) in a dedicated worktree, restored clean afterwards.

- `pnpm test`: 256 files, 1721 tests, all passing, one more than the previous round.
- `pnpm test-react18`: 189 files, 987 tests, all passing.
- `pnpm tsc`, `pnpm lint`: clean.
- `pnpm --filter app test-chrome popover-disclosure-7087` against this checkout's own servers: 7 passing.
- Red check on the new test against the previous revision, as described above.

### One optional suggestion

The permanent-mark limitation is currently visible only inside the collapsed review-gate blocks at the bottom of the description. Since it is a behavioral limitation an application cannot detect, and [#7095](https://github.com/ariakit/ariakit/issues/7095) exists for it, a sentence in the `## Solution` section saying the mark is never dropped and linking that issue would put it where someone deciding to merge will read it. Not a blocker, and the `## Reproduction` section is already accurate about the seven tests and what each one guards.

### Not fully validated

- Only the Chromium project ran locally. Firefox, WebKit and the mobile projects are on CI and on the author's reported runs.
- The legacy `website/build-pages` generator was covered only by its focused tests inside `pnpm test`. The reference-affecting JSDoc has not changed since the previous round, where I inspected its generated output directly.

PR #7096, opened 2026-08-08
Treat Safari Option+Tab as keyboard modality

### LINKED ISSUE #7094: Safari Option+Tab is treated as pointer modality, so keyboard focus is not focus-visible
### Problem

`Focusable` decides between keyboard and pointer modality from a global `keydown` listener that bails out on any `Alt`-modified key:

https://github.com/ariakit/ariakit/blob/9aa0b05e3103055b716396156d3ed9f772f0b161/packages/ariakit-react-components/src/focusable/focusable.tsx#L192-L197

On macOS that discards a real keyboard navigation gesture. When the system-wide keyboard navigation setting (`System Settings` > `Keyboard` > `Keyboard navigation`, exposed in Safari as `Press Tab to highlight each item on a webpage`) is off, Safari moves focus between all focusable elements with `Option+Tab` instead of plain `Tab`. Safari delivers that as a `keydown` with `altKey: true` and `key: "Tab"`, so Ariakit never sets `isKeyboardModality`, and the element that receives focus is treated as if it had been clicked.

Because `isKeyboardModality` starts as `true` and only flips to `false` on `mousedown`, this shows up as soon as the user has touched the mouse once, which is the normal case.

### Observed

Measured on [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) in headed WebKit driven by Playwright, where Ariakit's own `isSafari()` detection returns `true`. A `TooltipAnchor` rendered as a `Button` is focused by keyboard after one mouse click elsewhere on the page:

| Gesture | `data-focus-visible` on the button | Tooltip shown |
| --- | --- | --- |
| `Tab` | true | yes |
| `Option+Tab` | **false** | **no** |

Focus lands on the same button either way, so this is not a focus-movement problem. Only the modality classification differs, and everything keyed off it is lost: the focus ring that Ariakit components draw from `data-focus-visible`, and focus-driven behavior such as `Tooltip`.

Applying the candidate fix from #7091 (treat `Option+Tab` as keyboard modality when `isSafari()`) makes `Option+Tab` behave identically to `Tab` in the same run.

### Steps to reproduce

1. On macOS, make sure `System Settings` > `Keyboard` > `Keyboard navigation` is off, which is the default. In Safari this corresponds to `Press Tab to highlight each item on a webpage` being unchecked.
2. Open https://ariakit.com/examples/tooltip in Safari.
3. Click anywhere on the page background with the mouse.
4. Press `Option+Tab` until focus reaches the tooltip anchor.

Focus moves to the anchor, but no focus ring is drawn and the tooltip never appears. Doing the same with the setting on, using plain `Tab`, shows both.

### Expected outcome

`Option+Tab` is keyboard navigation in Safari, so a component focused that way should get `data-focus-visible` and trigger focus-driven behavior, exactly as with plain `Tab`. For users who have not enabled full keyboard access, `Option+Tab` is the only way to reach non-form controls, so today those users never get focus-visible styling at all.

### Workaround

Re-announce the gesture as an unmodified `keydown` so the global listener records keyboard modality. Verified on [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) in headed WebKit: with this effect mounted, `Option+Tab` produces `data-focus-visible` and shows the tooltip.

```jsx
import { useEffect } from "react";

function useOptionTabKeyboardModality() {
  useEffect(() => {
    const onKeyDown = (event) => {
      if (!event.isTrusted) return;
      if (!event.altKey) return;
      if (event.key !== "Tab") return;
      document.dispatchEvent(new KeyboardEvent("keydown", { key: "Tab" }));
    };
    document.addEventListener("keydown", onKeyDown, true);
    return () => document.removeEventListener("keydown", onKeyDown, true);
  }, []);
}
```

### Possible solutions

Keep the `Alt` bail-out, but make an exception for `Option+Tab` on Safari, which is what #7091 prototyped:

```ts
if (event.altKey) {
  if (!isSafariBrowser) return;
  if (event.key !== "Tab") return;
}
```

Note that `useFocusable` has a second `altKey` guard, in `onKeyDownCapture`, for typing on an element that already has focus. That one is not part of this problem, since the `Option+Tab` keydown is delivered to the element focus is leaving.

### Context

Found while reviewing #7091. The finding is pre-existing on `main` at [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161), independent of that branch.

Verification used Playwright's WebKit build rather than the Safari application. WebKit is the engine Safari ships and Ariakit's `isSafari()` check returns `true` there, but a confirmation in the Safari app with `Keyboard navigation` off would still be worth having.

diegohaz:
## Motivation

`Focusable` decides between keyboard and pointer modality from a global `keydown` listener that bails out on any `Alt`-modified key:

```ts
function onGlobalKeyDown(event: KeyboardEvent) {
  if (event.metaKey) return;
  if (event.ctrlKey) return;
  if (event.altKey) return;
  isKeyboardModality = true;
}
```

On macOS that discards a real keyboard navigation gesture. When the system-wide keyboard navigation setting (`System Settings` > `Keyboard` > `Keyboard navigation`, exposed in Safari as `Press Tab to highlight each item on a webpage`) is off, which is the default, Safari moves focus between all focusable elements with `Option+Tab` instead of plain `Tab`. Safari delivers that as a `keydown` with `altKey: true` and `key: "Tab"`, so Ariakit never sets `isKeyboardModality`, and the element that receives focus is treated as if it had been clicked.

Because `isKeyboardModality` starts as `true` and only flips to `false` on `mousedown`, this surfaces as soon as the user has touched the mouse once, which is the normal case. Focus lands on the same element either way, so this is not a focus movement problem. Only the modality classification differs, and everything keyed off it is lost: the focus ring components draw from `data-focus-visible`, and focus-driven behavior such as `Tooltip`.

For users who have not enabled full keyboard access, `Option+Tab` is the only way to reach non-form controls, so today those users never get focus-visible styling at all.

## Solution

Keep the `Alt` bail-out, but make an exception for `Option+Tab` on Safari, so the gesture is recorded as keyboard modality exactly like plain `Tab`:

```ts
if (event.altKey) {
  if (!isSafariBrowser) return;
  if (event.key !== "Tab") return;
}
```

The exception stays deliberately narrow in both directions. It is gated on Safari because `Alt+Tab` is the window switcher elsewhere, and restricted to `Tab` because `Option` is the special-character modifier on macOS, so `Option+e` and friends are typing rather than navigating. Every other `Alt`-modified key keeps the current modality, exactly as before.

The second `altKey` guard in `onKeyDownCapture` is intentionally left alone. It covers typing on an element that already has focus, and the `Option+Tab` keydown is delivered to the element focus is leaving, so it is not part of this problem.

## Preview

Safari reaching the tooltip anchor with `Option+Tab` after a mouse click elsewhere on the page. The anchor is focused, draws its focus ring from `data-focus-visible`, and shows its tooltip. Before this change the same gesture produced neither.

Safari preview showing the Bookmark button focused with a visible focus ring and its "Add to bookmarks" tooltip after pressing Option+Tab

## Workaround

Until this ships, consumers can re-announce the gesture as an unmodified `keydown` so the global listener records keyboard modality:

```jsx
import { useEffect } from "react";

// TODO: Remove once https://github.com/ariakit/ariakit/issues/7094 is released.
function useOptionTabKeyboardModality() {
  useEffect(() => {
    const onKeyDown = (event) => {
      if (!event.isTrusted) return;
      if (!event.altKey) return;
      if (event.key !== "Tab") return;
      // Ariakit reads only the modifier keys here, so the re-announced event
      // carries no key. Naming it Tab would make other document listeners
      // handle one gesture twice, which can move focus twice in a focus trap.
      document.dispatchEvent(new KeyboardEvent("keydown"));
    };
    document.addEventListener("keydown", onKeyDown, true);
    return () => document.removeEventListener("keydown", onKeyDown, true);
  }, []);
}
```

This workaround is deliberately narrower and, in one respect, broader than the library fix, so it is not a complete substitute for it.

- **It is not Safari-gated.** The fix applies the exception only on Safari, while the snippet reacts to `Alt+Tab` on any platform. Gating it in userland would mean a user agent sniff, or adding `@ariakit/utils` just to import `isSafari()`, which is not worth it for code you delete after the release. The practical consequence is rare: on macOS Chrome and Firefox, `Option+Tab` reaches the page but does not move focus, so the flag flips with no matching focus event, and a later window blur and refocus can paint a focus ring on an element the user had clicked. On Windows and most Linux window managers the shell consumes `Alt+Tab` before the page sees it.
- **It only covers `Option+Tab`**, including `Option+Shift+Tab`. Any other `Alt`-modified key still leaves modality untouched.
- **It restores only the global modality flag, not per-element keyboard handling.** The synthetic event targets `document`, and propagation only visits the target's ancestors, so it never reaches element handlers. The second `altKey` guard in `onKeyDownCapture` is unaffected, so `Option+Tab` still does not make an already-focused element focus-visible in place. Only the element that subsequently receives focus benefits, which is what this bug needs.
- **It only reacts to trusted events**, so it is a no-op under testing libraries that dispatch synthetic events, such as happy-dom and jsdom. It can only be covered in a real browser.
- **It only covers the top-level document.** Ariakit registers its global listeners on same-origin child frames too, so `Option+Tab` inside an iframe is not re-announced.
- **It should be installed once, near the app root.** Several mounted copies each dispatch one extra event per gesture, which is harmless because the flag is idempotent.
- **Other document-level `keydown` listeners observe one extra untrusted event.** It carries no key, so handlers that branch on `event.key` ignore it.
- **The listener is installed in an effect**, so a gesture that arrives before hydration is not re-announced.

## Regression test

`app/src/sandbox/focusable-tooltip-7094` renders a `TooltipAnchor` as a `Button`, which is the reported scenario. `test-safari.ts` clicks the page text to switch to pointer modality, presses `Option+Tab`, and then asserts that the anchor is focused, carries `data-focus-visible`, and shows its tooltip. A second test clicks the anchor directly and asserts that pointer modality still does not produce `data-focus-visible`, so the first test's failure is attributable to modality classification rather than to the fixture never applying focus visibility at all.

Both tests wait for the `tabindex` that `Focusable` adds on Safari after hydration before interacting. The listener that records pointer modality is installed from an effect, so without that gate a click landing before the island hydrates would leave modality at its `true` default and let the first test pass on broken code.

The test is red on the failing-repro commit and green with the fix. Coverage is Playwright-only. A happy-dom duplicate is not possible here: happy-dom hardcodes `navigator.vendor` to an empty string, so `isSafari()` is always false there, which means such a test would stay red both before and after a Safari-gated fix.

### Evidence gap

The two negative branches of the guard are not covered by automated tests, and both were judged impractical rather than overlooked.

`Alt+Tab` in a non-Safari browser has no observable effect to assert, because no other engine moves focus on that gesture, so there is no focus event to check.

An `Alt`-modified non-`Tab` key is technically observable, since composite widgets move focus on `Alt+ArrowRight`, but asserting today's outcome would pin behavior that is itself questionable. That inconsistency is pre-existing and independent of this change, and is filed separately as #7099.

Fixes #7094

### DIFF
--- a/.changeset/300-safari-option-tab-focus-visible.md
+++ b/.changeset/300-safari-option-tab-focus-visible.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) treating Safari's `Option+Tab` navigation as a pointer interaction, so components built on it such as [`Button`](https://ariakit.com/reference/button) and [`TooltipAnchor`](https://ariakit.com/reference/tooltip-anchor) now receive [`data-focus-visible`](https://ariakit.com/guide/styling#data-focus-visible) when focus reaches them that way. On macOS, `Option+Tab` is how Safari moves focus between all focusable elements while the system keyboard navigation setting is off.
--- a/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
+++ b/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
@@ -1,6 +1,28 @@
 import * as Ariakit from "@ariakit/react";
+import { useEffect } from "react";
+
+// TODO: Remove once https://github.com/ariakit/ariakit/issues/7094 is released.
+// Safari navigates with Option+Tab when the macOS keyboard navigation setting
+// is off, but Ariakit ignores every Alt-modified key when it records keyboard
+// modality. Re-announcing the gesture without the modifier makes it count.
+function useOptionTabKeyboardModality() {
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (!event.isTrusted) return;
+      if (!event.altKey) return;
+      if (event.key !== "Tab") return;
+      // Ariakit reads only the modifier keys here, so the re-announced event
+      // carries no key. Naming it Tab would make other document listeners
+      // handle one gesture twice, which can move focus twice in a focus trap.
+      document.dispatchEvent(new KeyboardEvent("keydown"));
+    };
+    document.addEventListener("keydown", onKeyDown, true);
+    return () => document.removeEventListener("keydown", onKeyDown, true);
+  }, []);
+}
 
 export default function Example() {
+  useOptionTabKeyboardModality();
   return (
     <div>
       <p>Click here, then press Option+Tab to focus the button.</p>
--- a/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
+++ b/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
@@ -2,7 +2,7 @@ import * as Ariakit from "@ariakit/react";
 
 export default function Example() {
   return (
-    <div>
+    <div className="flex flex-col items-center gap-12">
       <p>Click here, then press Option+Tab to focus the button.</p>
       <Ariakit.TooltipProvider>
         <Ariakit.TooltipAnchor render={<Ariakit.Button />}>
--- a/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
+++ b/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
@@ -1,0 +1,15 @@
+import * as Ariakit from "@ariakit/react";
+
+export default function Example() {
+  return (
+    <div>
+      <p>Click here, then press Option+Tab to focus the button.</p>
+      <Ariakit.TooltipProvider>
+        <Ariakit.TooltipAnchor render={<Ariakit.Button />}>
+          Bookmark
+        </Ariakit.TooltipAnchor>
+        <Ariakit.Tooltip>Add to bookmarks</Ariakit.Tooltip>
+      </Ariakit.TooltipProvider>
+    </div>
+  );
+}
--- a/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
+++ b/app/src/sandbox/focusable-tooltip-7094/index.react.tsx
@@ -1,28 +1,6 @@
 import * as Ariakit from "@ariakit/react";
-import { useEffect } from "react";
-
-// TODO: Remove once https://github.com/ariakit/ariakit/issues/7094 is released.
-// Safari navigates with Option+Tab when the macOS keyboard navigation setting
-// is off, but Ariakit ignores every Alt-modified key when it records keyboard
-// modality. Re-announcing the gesture without the modifier makes it count.
-function useOptionTabKeyboardModality() {
-  useEffect(() => {
-    const onKeyDown = (event: KeyboardEvent) => {
-      if (!event.isTrusted) return;
-      if (!event.altKey) return;
-      if (event.key !== "Tab") return;
-      // Ariakit reads only the modifier keys here, so the re-announced event
-      // carries no key. Naming it Tab would make other document listeners
-      // handle one gesture twice, which can move focus twice in a focus trap.
-      document.dispatchEvent(new KeyboardEvent("keydown"));
-    };
-    document.addEventListener("keydown", onKeyDown, true);
-    return () => document.removeEventListener("keydown", onKeyDown, true);
-  }, []);
-}
 
 export default function Example() {
-  useOptionTabKeyboardModality();
   return (
     <div>
       <p>Click here, then press Option+Tab to focus the button.</p>
--- a/app/src/sandbox/focusable-tooltip-7094/test-safari.ts
+++ b/app/src/sandbox/focusable-tooltip-7094/test-safari.ts
@@ -1,0 +1,47 @@
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/7094
+  test("Option+Tab focuses the anchor as keyboard navigation", async ({
+    page,
+    q,
+  }) => {
+    const anchor = q.button("Bookmark");
+    // The listener that records pointer modality is installed when the island
+    // hydrates, and Safari's tabindex is applied by a later effect in the same
+    // component, so waiting for it proves the click reaches that listener. It
+    // only proves that while the fixture passes no tabIndex of its own.
+    await test.expect(anchor).toHaveAttribute("tabindex", "0");
+
+    // Safari moves focus with Option+Tab instead of Tab when the macOS keyboard
+    // navigation setting is off. Clicking first is what makes the regression
+    // observable: modality starts as keyboard until a pointer gesture drops it.
+    await q
+      .text("Click here, then press Option+Tab to focus the button.")
+      .click();
+    await page.keyboard.press("Alt+Tab");
+
+    await test.expect(anchor).toBeFocused();
+    await test.expect(anchor).toHaveAttribute("data-focus-visible", "true");
+    await test.expect(q.tooltip("Add to bookmarks")).toBeVisible();
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7094
+  test("clicking the anchor is still pointer navigation", async ({
+    page,
+    q,
+  }) => {
+    const anchor = q.button("Bookmark");
+    // Same hydration gate as above. Without it a pre-hydration click would
+    // leave the attribute absent for the wrong reason.
+    await test.expect(anchor).toHaveAttribute("tabindex", "0");
+
+    await anchor.click();
+    await test.expect(anchor).toBeFocused();
+    // Focus visibility is applied from a queueBeforeEvent callback that falls
+    // back to requestAnimationFrame, so the attribute would also be absent
+    // while it is still pending. Cross those frames to prove it never arrives.
+    await flushFrames(page);
+    await test.expect(anchor).not.toHaveAttribute("data-focus-visible");
+  });
+});
--- a/packages/ariakit-react-components/src/focusable/focusable.tsx
+++ b/packages/ariakit-react-components/src/focusable/focusable.tsx
@@ -192,7 +192,15 @@ function onGlobalMouseDown(event: MouseEvent) {
 function onGlobalKeyDown(event: KeyboardEvent) {
   if (event.metaKey) return;
   if (event.ctrlKey) return;
-  if (event.altKey) return;
+  if (event.altKey) {
+    // Safari moves focus with Option+Tab when the macOS keyboard navigation
+    // setting is off, so that gesture is keyboard navigation rather than a
+    // shortcut. Every other Alt-modified key still keeps the current modality,
+    // since Option types special characters on macOS.
+    // https://github.com/ariakit/ariakit/issues/7094
+    if (!isSafariBrowser) return;
+    if (event.key !== "Tab") return;
+  }
   isKeyboardModality = true;
 }
 


diegohaz (commented):
Initial review, baselined on the merge base [`9aa0b05`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) through head [`d194773`](https://github.com/ariakit/ariakit/commit/d194773aab6f447f9ff36115c5fa6bc00e3bb46d). This PR has no review threads, so there was nothing to revalidate. I read the diff and then the surrounding modality machinery in `focusable.tsx`, `addGlobalEventListener`, the Playwright project and workflow wiring, the changeset conventions with existing entries for comparison, and the linked issues.

No findings are published inline. The library change is correct, minimal, and well scoped, and I could not construct a failing case against it. Three lesser findings are dispositioned at the end.

## What I verified

**The regression proof, from CI rather than the description.** On [`a10d182`](https://github.com/ariakit/ariakit/commit/a10d18295c35a50919c5091dd214ed6be55045a0), which adds the test with neither the workaround nor the fix, `App / Test Safari (2/2)` failed at `app/src/sandbox/focusable-tooltip-7094/test-safari.ts:25` on all four attempts. That is the `data-focus-visible` assertion, and line 24's `toBeFocused()` passed, so focus reached the anchor and only the modality classification differed, exactly as #7094 describes. The control test passed in the same run. [`4f90d54`](https://github.com/ariakit/ariakit/commit/4f90d54ea24939ebd6c2e7725803fd4ceec0eed3) (sandbox workaround) and [`41acea7`](https://github.com/ariakit/ariakit/commit/41acea7fa055c1d94a6bf0263b82fd596899335d) (library fix) are both green, so the workaround and the fix are each demonstrated independently. This is stronger evidence than a local run, which is why I did not reproduce it locally.

**The guard's placement and its narrowing in both directions.** `onGlobalKeyDown` is installed through `addGlobalEventListener`, which attaches to `scope.document` and every same-origin child frame, so the exception applies wherever the existing listener already runs. `Cmd+Option+Tab` still bails on the `metaKey` check above it, a bare `Option` press reports `key: "Alt"` and bails, and `Option+Shift+Tab` reports `key: "Tab"` and is covered.

**Test wiring and project mapping.** `test-safari.ts` matches only `testMatchersFor("safari", "browser")` in `app/playwright.config.ts`, and that project runs on `macos-latest` in `.github/workflows/app.yml`, where `isSafari()` is true. It does not match the `ios` project, which is right, since the gesture does not exist there.

**The hydration gate is not vacuous.** For a native button with no `tabIndex` prop, `getTabIndex` returns `tabIndexProp`, so no `tabindex` attribute is rendered until `safariTabIndex` flips in the effect at `focusable.tsx:439`. That effect is declared after the global listener effect at `focusable.tsx:234`, so waiting on `tabindex="0"` genuinely proves the `mousedown` listener is installed before the click lands. The comment's caveat about a fixture passing its own `tabIndex` is accurate.

**Browser-only coverage is the right call.** happy-dom's `Navigator` returns `''` from its `vendor` getter, so `isSafari()` can never be true there and a happy-dom duplicate would stay red before and after the fix.

**The changeset.** Both `@ariakit/react-components` and `@ariakit/react` are correct, every named API (`Focusable`, `Button`, `TooltipAnchor`) is exported from `@ariakit/react`, `300-` matches a patch, and naming two derived components is the limit for a single-paragraph entry. All four links return `200`, and the styling guide does carry the `data-focus-visible` anchor. The two-sentence single-paragraph shape matches existing entries such as `.changeset/300-composite-focus-presentation.md`. No contributor credit is due: #7094 was opened by `ariakit-bot` and its only comment is from a bot.

## Alternatives weighed

Dropping the Safari gate and treating every `Alt+Tab` as keyboard navigation is worse for the reason the description gives, and I agree with it. Playwright injects key events into the renderer rather than through the window manager, so macOS Chrome and Firefox do see the gesture and do not move focus, which would flip modality with no matching focus event.

Gating on `isApple()` instead of `isSafari()` has the same defect, since it would include macOS Chrome and Firefox. Matching `event.code` instead of `event.key` buys nothing, because `Tab` has no Option-composed character on macOS and `event.key` is the convention here.

Broadening the exception to every `Alt`-modified key that actually moves focus is correctly out of scope and registered as #7099, which also records that `Option+ArrowRight` moves composite focus in every browser while the same listener ignores it.

Leaving the second `altKey` guard in `onKeyDownCapture` alone is right for this bug: the `Option+Tab` keydown lands on the element focus is leaving, and the element that receives focus goes through `onFocusCapture`. I did look for a case where `Option+Tab` does not move focus at all, which would leave that element non-focus-visible. Ariakit dialogs use focus trap sentinels (`focus-trap.tsx`, `focus-trap-region.tsx`), so focus does leave and return and fires a real focus event. I did not execute that path in a browser, so I am not raising it as a finding, only recording that I looked.

## Findings

None of these is published inline. The first has no line in this diff to anchor to, and the other two are not current-PR work under the dispositions below.

### 1. Issue #7094 still publishes the superseded workaround

The workaround on #7094, which carries the `has workaround` label and is the copy users will find, dispatches `new KeyboardEvent("keydown", { key: "Tab" })`. The version in this PR's description, matching what [`4f90d54`](https://github.com/ariakit/ariakit/commit/4f90d54ea24939ebd6c2e7725803fd4ceec0eed3) actually verified in the sandbox, dispatches `new KeyboardEvent("keydown")` with no key, and its comment states why: naming it `Tab` makes other document-level listeners handle one gesture twice, which can move focus twice in a focus trap. Both variants set the modality flag, since `onGlobalKeyDown` reads only the modifiers, so dropping the key costs nothing and avoids the double handling.

Ariakit itself has no document-level `Tab` handler to double-handle, so the exposure is consumer code and third-party focus trap libraries, which do listen on `document` and branch on `event.key === "Tab"`.

Worth syncing the issue body to the refined snippet while this is fresh. The triage automation has overwritten issue bodies before, so it is worth reading the body back after the edit rather than trusting the exit status.

**Assessment**

- Severity: Low (estimate, 85% confidence), because it only affects consumers who apply the workaround in the window before this ships.
- Priority: P2 (estimate, 80% confidence), because it is an issue edit rather than a change to this PR, though it is cheap to do now.
- Breakage likelihood: Possible (estimate, 70% confidence), because it needs the consumer to also run a document-level `keydown` listener that branches on `event.key === "Tab"`, which is exactly what focus trap libraries do.
- Fix effort: Easy (estimate, 95% confidence), because it is a single snippet replacement in the issue body.

### 2. The stated rationale for the second test does not hold on its own

The description says the second test exists "so the first test's failure is attributable to modality classification rather than to the fixture never applying focus visibility at all". A test that asserts an absence cannot separate those two: if the fixture never applied focus visibility under any circumstance, the second test would still pass. What actually rules that out is the first test passing with the fix, and the CI evidence above shows exactly that pairing.

The test itself is worth keeping. Its real value is proving `data-focus-visible` is not unconditionally present on the anchor, which is what makes the first test's positive assertion meaningful. Only the sentence in the description overstates what it establishes on its own.

**Assessment**

- Severity: Low (verified), because it is a description claim with no effect on shipped behavior or on the tests.
- Priority: P2 (estimate, 90% confidence), because it is a wording fix in the PR body.
- Breakage likelihood: Not applicable (verified), because no code or test depends on the claim.
- Fix effort: Easy (verified), because it is one sentence.

### 3. The non-Safari branch is observable, contrary to the evidence gap note

The evidence gap section says `Alt+Tab` in a non-Safari browser has "no observable effect to assert, because no other engine moves focus on that gesture, so there is no focus event to check". The gesture does not move focus, but the modality flag it would wrongly set is still observable through a later programmatic focus: click to drop into pointer modality, press `Alt+Tab`, then call `.focus()` on the anchor and assert `data-focus-visible` is absent. Since Playwright injects key events into the renderer rather than through the window manager, the keydown does reach the listener in Chromium and Firefox, so such a test would not pass vacuously.

I am not asking for it in this PR. Using `.focus()` as a probe cuts against the repo's own guidance to reserve it for cases where programmatic focus is the behavior under test, and the fixture has no non-Safari hydration gate, since the `tabindex="0"` gate exists only on Safari. Adding one would mean fixture state built for the test. Disposition is declined non-goal, but the reason is the cost and shape of the test, not that the branch is unobservable, so the note in the description is worth rewording.

**Assessment**

- Severity: Low (estimate, 80% confidence), because the untested guard's failure mode is a spurious focus ring on macOS Chrome and Firefox after `Option+Tab` followed by a window refocus.
- Priority: P2 (estimate, 75% confidence), because nothing in this change makes the gap worse.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because it takes a later edit that removes the `isSafariBrowser` check to reach it.
- Fix effort: Moderate (estimate, 70% confidence), because the fixture needs a non-Safari hydration gate before the test is trustworthy.

## Not fully validated

I did not run Playwright or the Vitest suites locally; the review worktree has no install, and this PR's own macOS Safari runs, including the red one on the test-only commit, are better evidence than a local run would be. I also did not exercise real macOS Safari with the system keyboard navigation setting off, so the reported `altKey` and `key` values for the gesture rest on the WebKit CI runs and on #7094's measurements rather than on my own observation.

PR #7098, opened 2026-08-08
Keep focus on a collapsed ComboboxSelect

### LINKED ISSUE #7093: ComboboxSelect moves focus into the popup when typeahead runs while the popup is closed
### Problem

[`ComboboxSelect`](https://ariakit.com/reference/combobox-select) turns itself into the composite widget when the popup has no [`ComboboxInput`](https://ariakit.com/reference/combobox), and it never narrows `focusOnMove`:

https://github.com/ariakit/ariakit/blob/9aa0b05e3103055b716396156d3ed9f772f0b161/packages/ariakit-react-components/src/combobox/combobox-select.tsx#L318-L328

`useComposite` defaults `focusOnMove` to `composite`, so it stays `true` even while the popup is closed. Every closed move, including typeahead, therefore asks `CompositeFocusOnMove` to focus the active item:

https://github.com/ariakit/ariakit/blob/9aa0b05e3103055b716396156d3ed9f772f0b161/packages/ariakit-react-components/src/composite/composite.tsx#L153-L168

This is usually invisible, because a closed popup is `display: none` and `element.focus()` on a hidden element is a no-op. It stops being invisible as soon as the items are still rendered and visible while the combobox is collapsed, which is exactly what [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) does, and what an exit transition leaves behind for the duration of the animation. Typing a letter on the collapsed select then pulls DOM focus out of the select and into a popup option.

### Observed

Measured on [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) in headed Chromium with a `ComboboxSelect` whose `ComboboxList` is `alwaysVisible`. The select is focused, the popup is collapsed (`aria-expanded="false"`), and `g` is pressed once to jump to `Grape`:

| Store config | `focusin` events | Active element afterwards |
| --- | --- | --- |
| `virtualFocus={false}` | `div[role=option][Grape]` | **the option**, while the select is still collapsed |
| default `virtualFocus` | `div[role=option][Grape]`, then back to the select | the select, after a spurious blur and refocus round trip |

With `virtualFocus={false}` the select simply loses focus to an option of a popup that is not open. With the default virtual focus the DOM focus bounces, which still fires `blur` and `focus` on the select and on the option, so consumer handlers such as blur validation, `onFocus` analytics and `data-focus-visible` styling all run when the user only typed a letter.

Checking out the same repro with the candidate fix from #7091 applied (`focusOnMove: open` in `useComboboxSelect`, plus the replay guard in `CompositeFocusOnMove`) produces no `focusin` events at all in either configuration, and focus stays on the select.

### Steps to reproduce

```jsx

const fruits = ["Apple", "Banana", "Grape", "Orange"];

  return (
    <Ariakit.ComboboxProvider defaultSelectedValue="Apple" virtualFocus={false}>
      <Ariakit.ComboboxSelect aria-label="Fruit" />
      <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
        {fruits.map((value) => (
          <Ariakit.ComboboxItem key={value} value={value} />
        ))}
      </Ariakit.ComboboxList>
    </Ariakit.ComboboxProvider>
  );
}
```

1. Focus the select with `Tab`. Do not open it.
2. Press `g`.

The selected value becomes `Grape` as expected, but `document.activeElement` is now the `Grape` option instead of the select, and `aria-expanded` is still `false`.

### Expected outcome

While the popup is closed, typeahead should only update the active item and the selected value. Focus belongs to the select, the same way a native `<select>` keeps focus on the control when typeahead changes its value.

### Workaround

Restrict `focusOnMove` to the open state on the select itself. That snippet is incomplete: it does not type-check, and it leaves the focus-time trigger in place. Use the corrected version and its limitations instead:

https://github.com/ariakit/ariakit/issues/7093#issuecomment-5229254268

### Possible solutions

Pass `focusOnMove: open` from `useComboboxSelect` so the closed select never asks for item presentation. That alone introduces a second problem, because enabling `focusOnMove` on open makes the presentation effect replay the move that already completed while it was disabled, so the guard has to remember the last move seen while disabled. Both halves were prototyped in #7091, which is not going to be merged.

### Context

Found while reviewing #7091. The finding is pre-existing on `main` at [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161), independent of that branch.

diegohaz:
## Motivation

[`ComboboxSelect`](https://ariakit.com/reference/combobox-select) becomes the composite widget when its popup has no [`ComboboxInput`](https://ariakit.com/reference/combobox), and it never narrows `focusOnMove`. `useComposite` defaults `focusOnMove` to `composite`, so the collapsed select keeps asking for its active item to be focused.

That is invisible while the popup is `display: none`, because `element.focus()` on a hidden element is a no-op. It stops being invisible as soon as the options stay on screen while the select is collapsed, which is what [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) does.

```jsx
<Ariakit.ComboboxProvider defaultSelectedValue="Apple" virtualFocus={false}>
  <Ariakit.ComboboxSelect aria-label="Fruit" />
  <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
    {["Apple", "Banana", "Grape", "Orange"].map((value) => (
      <Ariakit.ComboboxItem key={value} value={value} />
    ))}
  </Ariakit.ComboboxList>
</Ariakit.ComboboxProvider>
```

Focusing the collapsed select and pressing `g` selects `Grape` as expected, but `document.activeElement` becomes the `Grape` option while the select still reports `aria-expanded="false"`.

Typeahead is not the only trigger. With the default `virtualFocus`, `Composite`'s own focus handler queues a presentation whenever the container is focused, so simply tabbing into the collapsed select focuses the selected option and hands focus straight back, firing `focusin` and `focusout` on an option of a listbox that never opened. Handlers on the items see both. The same happens to a plain [`Combobox`](https://ariakit.com/reference/combobox) input that has an active item and a standalone `alwaysVisible` list.

## Solution

`presentItem` now withholds DOM focus when giving it would take focus off the control the user is on and put it inside a popup that has not opened. Two conditions have to hold: focus started on the composite element, and the item is inside a popup that the composite element is not in.

The second condition is containment rather than open state. An `alwaysVisible` popup that is its own composite, such as a [`Menu`](https://ariakit.com/reference/menu), navigates with DOM focus while its store is closed, and [#6986](https://github.com/ariakit/ariakit/issues/6986) settled that its items have to keep receiving focus. A `ComboboxSelect` is the opposite shape: the control sits outside the popup, so focus moving into the popup takes it away from the widget the user is on.

The first condition is latched once, when the request is created, and it is what keeps a closed popup receiving focus in the two cases where that is still right. A move made from outside the widget presents immediately, the way `focusOnMove` documents, instead of leaving a request whose focus lands on whoever is focused when the popup next opens. And once focus is already on an option, arrow keys keep moving it between options of a collapsed list, because that is navigation inside the popup rather than focus leaving the control. So the composite being its own popup is not the only reason a closed popup keeps receiving focus; being where the user already is counts too.

Only the focus half is withheld. The list still scrolls to the active option, and the request stays pending rather than being cancelled, so opening the popup still presents the option that was made active while it was closed. That pending lifetime is load-bearing: opening with `ArrowDown` moves the active item while the popup is still closed and only calls `show` on `keyup`.

A withheld request is retired once the active item moves away from the one it resolved. Unlike a request parked on positioning, which the popup is about to finish opening for, this one can wait indefinitely, so the user has time to pick another option and the focus it still owes would otherwise land on the wrong one.

## Testing

- `pnpm test` (1721 tests), `pnpm tsc`, `pnpm lint-fix`
- `pnpm -F app test` for chrome (753), firefox (612) and safari (501)
- Every new assertion was red-checked against the behavior it protects, including the ownership latch in both directions, the withheld-request guard, and the preserved scroll
- `pnpm test-react18` could not be run locally: it fails in its temp-workspace `prepare` step with a husky `MODULE_NOT_FOUND` before any test loads, and reproduces on a clean checkout, so it is environmental. The `Main / Test React 18` CI job covers it.

## Preview

Focusing the collapsed select and pressing `g` in `app/src/sandbox/combobox-select-collapsed-focus/`, with `virtualFocus={false}` so the lost focus is visible. Before, focus lands on the `Grape` option of a listbox that is still collapsed; after, it stays on the select.

Before:

Collapsed select showing Grape, with the focus ring on the Grape option in the list below and the readout Focused options: Grape

After:

Collapsed select showing Grape with the focus ring on the select itself and the readout Focused options: none

## Workaround

Until this is released, a collapsed select can be kept out of its own list with two userland levers, applied to the sandbox in [`b4f64fc`](https://github.com/ariakit/ariakit/commit/b4f64fc53) and reverted by the fix commit.

```jsx
function FruitSelect() {
  const store = Ariakit.useComboboxStore({ defaultSelectedValue: "Apple" });
  // `ComboboxSelect` forwards composite options to `useComposite`, but it
  // doesn't list them in its own props type, so they are spread in.
  const focusOnMoveProps = {
    focusOnMove: open,
  } satisfies Ariakit.CompositeOptions;

  return (
    <Ariakit.ComboboxProvider store={store}>
      <Ariakit.ComboboxSelect
        aria-label="Fruit"
        // TODO: remove both workarounds once
        // https://github.com/ariakit/ariakit/issues/7093 is fixed.
        {...focusOnMoveProps}
        onFocus={(event) => {
          const state = store.getState();
          if (state.open) return;
          if (!state.virtualFocus) return;
          event.preventDefault();
        }}
      />
      <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
        {/* ... */}
      </Ariakit.ComboboxList>
    </Ariakit.ComboboxProvider>
  );
}
```

`focusOnMove` covers the move trigger and the `onFocus` handler covers the virtual-focus trigger. This is narrower than the fix, so it comes with limitations:

- `focusOnMove` is not declared on `ComboboxSelectOptions`, even though `useComboboxSelect` forwards it to `useComposite`. Passing it directly as a JSX attribute is a TypeScript excess-property error, hence the `satisfies`-guarded spread.
- The `onFocus` guard is deliberately limited to virtual focus, so the roving-tabindex branch keeps clearing `activeId` when the select itself is focused.
- Preventing the default focus handling also drops the legitimate presentation that would have run when the popup opens from a pointer click with no prior move. With a bare `ComboboxList`, clicking to open then leaves the list scrolled where it was instead of at the selected item. A `ComboboxPopover` is unaffected. The fix does not share this limitation, because it keeps the request pending instead of dropping it.
- `preventDefault` on a focus event propagates, so an ancestor Ariakit component that bails on `defaultPrevented` in its own `onFocus`, such as `Hovercard` or `CompositeContainer`, is skipped too. There is no userland way to scope it.

## Review gate reassessment

<details>
<summary>Cycle 3: Narrow the contract to popup containment</summary>

**Assessment**

The bug is a real but narrow a11y defect: DOM focus lands on an option of a listbox reporting `aria-expanded="false"`. It is only observable when the options are rendered and visible while the control is collapsed, so `alwaysVisible` or an exit transition. A documented userland workaround exists.

The direction under review was "DOM focus must not reach an item whose popup has not opened". Review found that `app/src/sandbox/menu-always-visible/` (added for [#6986](https://github.com/ariakit/ariakit/issues/6986)) asserts the opposite for a composite whose element is the popup: "DOM focus is the accessibility cursor for a roving tabindex composite like this one, so it has to follow the active item." That contract would have gone red.

Review also found a second trigger of the same defect that `focusOnMove: open` does not reach, because `Composite`'s own focus handler queues a presentation under virtual focus.

**Decision**

Narrow the contract to popup containment: a composite element that sits outside its own popup must not hand DOM focus to items inside that popup while the popup is closed. That covers both triggers, keeps `#6986` green, and removes the need for the `focusOnMove` replay guard that the issue anticipated. Legacy `Select` stays uncovered because its composite element is inside its popup; that is a separate defect, and it was declined as a non-goal.

</details>

<details>
<summary>Cycle 4: Remove the scroll-once latch</summary>

**Assessment**

Withholding focus while keeping the request pending meant the scroll could run again on later store notifications, and a `scrolled` latch was added for it. The next round found that latch could never reset, because its reset site is guarded by a flag the withheld path never sets. Two consecutive rounds of findings clustered on machinery that existed only to defend one edge case.

That edge case is negligible. While the popup is closed, `openingMovesByStore` holds no entry, so the scroll always takes the nearest-edge branch, and bringing an already-visible item into view does nothing. Observing a difference requires manually scrolling a closed, always-visible list between two item registrations.

**Decision**

Remove the latch and decline the edge case. Keep the gate and the pending lifetime: the pending lifetime is load-bearing, because opening with `ArrowDown` moves the active item while the popup is still closed and only calls `show` on `keyup`, and the scroll itself is existing behavior that a plain early return would have regressed.

</details>

<details>
<summary>Cycle 3 (second count): Narrow the gate to visible popups</summary>

**Assessment**

Reassessment found that a bare `ComboboxList` keeps `contentElement` assigned and only adds `hidden` and `display: none` when closed, so the gate also applied to ordinary hidden popups where the defect is provably invisible. There, a focus that used to be a harmless no-op followed by a cancel became a request that survived and delivered a real focus at the next open, in a composition no fixture covers.

The rest of the direction is converging: this round's other findings were all comments and changelog copy, and the library diff was unchanged by them.

**Decision**

Narrow the predicate with a visibility check, positioned last because it reads style. This makes the focus half share the precondition the scroll half already applies, restores the previous behavior for every hidden closed popup, and confines the change to the shape the issue describes.

Intentionally unsupported: a withheld request survives an explicit `activeId` clear and still presents its pinned item when the popup opens. Tightening the retirement guard to cover that would cancel the request when a `virtualFocus={false}` select is clicked, because focusing it clears the active item, and nothing else would present it on open.

</details>

Fixes #7093

### DIFF
--- a/.changeset/300-combobox-select-collapsed-focus.md
+++ b/.changeset/300-combobox-select-collapsed-focus.md
@@ -1,0 +1,10 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+Collapsed combobox controls keep focus
+
+[`ComboboxSelect`](https://ariakit.com/reference/combobox-select) and [`Combobox`](https://ariakit.com/reference/combobox) no longer move DOM focus into their popup while it is closed. This was visible whenever the options stayed on screen next to the collapsed control, such as with an [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) [`ComboboxList`](https://ariakit.com/reference/combobox-list), where focusing either control, or typing a letter on the select, moved DOM focus onto an option of a listbox that still reported `aria-expanded="false"`, firing focus and blur on it and, without virtual focus, leaving focus there.
+
+The list still scrolls to the active option, and opening the popup still presents the option that was made active while it was closed.
--- a/.changeset/300-combobox-select-collapsed-focus.md
+++ b/.changeset/300-combobox-select-collapsed-focus.md
@@ -3,8 +3,4 @@
 "@ariakit/react-components": patch
 ---
 
-Collapsed combobox controls keep focus
-
-[`ComboboxSelect`](https://ariakit.com/reference/combobox-select) and [`Combobox`](https://ariakit.com/reference/combobox) no longer move DOM focus into their popup while it is closed. This was visible whenever the options stayed on screen next to the collapsed control, such as with an [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) [`ComboboxList`](https://ariakit.com/reference/combobox-list), where focusing either control, or typing a letter on the select, moved DOM focus onto an option of a listbox that still reported `aria-expanded="false"`, firing focus and blur on it and, without virtual focus, leaving focus there.
-
-The list still scrolls to the active option, and opening the popup still presents the option that was made active while it was closed.
+Fixed [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) and [`Combobox`](https://ariakit.com/reference/combobox) moving DOM focus off the collapsed control and into their popup, which was visible when the options stayed on screen next to it, such as with an [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) [`ComboboxList`](https://ariakit.com/reference/combobox-list).
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -3,6 +3,22 @@ import { useState } from "react";
 
 const fruits = ["Apple", "Banana", "Grape", "Orange"];
 
+interface FocusedOptionsProps {
+  label: string;
+  items: string[];
+}
+
+function FocusedOptions({ label, items }: FocusedOptionsProps) {
+  return (
+    <p>
+      Focused options:{" "}
+      <output aria-label={`${label} focused options`}>
+        {items.join(", ") || "none"}
+      </output>
+    </p>
+  );
+}
+
 interface FruitSelectProps {
   label: string;
   virtualFocus?: boolean;
@@ -10,35 +26,14 @@ interface FruitSelectProps {
 
 function FruitSelect({ label, virtualFocus }: FruitSelectProps) {
   const [focusedItems, setFocusedItems] = useState<string[]>([]);
-  const store = Ariakit.useComboboxStore({
-    defaultSelectedValue: "Apple",
-    virtualFocus,
-  });
-  const open = Ariakit.useStoreState(store, "open");
-  // `ComboboxSelect` forwards composite options to `useComposite`, but it
-  // doesn't list them in its own props type, so they are spread in.
-  const focusOnMoveProps = {
-    focusOnMove: open,
-  } satisfies Ariakit.CompositeOptions;
 
   return (
     <div>
-      <Ariakit.ComboboxProvider store={store}>
-        <Ariakit.ComboboxSelect
-          aria-label={label}
-          // TODO: remove both workarounds once
-          // https://github.com/ariakit/ariakit/issues/7093 is fixed. The prop
-          // stops the closed moves from presenting an item, and the handler
-          // stops the presentation that virtual focus queues when the
-          // collapsed select itself is focused.
-          {...focusOnMoveProps}
-          onFocus={(event) => {
-            const state = store.getState();
-            if (state.open) return;
-            if (!state.virtualFocus) return;
-            event.preventDefault();
-          }}
-        />
+      <Ariakit.ComboboxProvider
+        defaultSelectedValue="Apple"
+        virtualFocus={virtualFocus}
+      >
+        <Ariakit.ComboboxSelect aria-label={label} />
         {/* The list stays visible while the select is collapsed, which is
             also what an exit transition leaves behind while it runs. */}
         <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
@@ -55,12 +50,57 @@ function FruitSelect({ label, virtualFocus }: FruitSelectProps) {
           ))}
         </Ariakit.ComboboxList>
       </Ariakit.ComboboxProvider>
-      <p>
-        Focused options:{" "}
-        <output aria-label={`${label} focused options`}>
-          {focusedItems.join(", ") || "none"}
-        </output>
-      </p>
+      <FocusedOptions label={label} items={focusedItems} />
+    </div>
+  );
+}
+
+// Enough options that the typeahead target starts outside the scrollport, so
+// the list has somewhere to scroll while the select is collapsed.
+const codes = [
+  ...Array.from({ length: 29 }, (_, index) => `Alpha ${index + 1}`),
+  "Zulu",
+];
+
+function CodeSelect() {
+  return (
+    <Ariakit.ComboboxProvider defaultSelectedValue="Alpha 1">
+      <Ariakit.ComboboxSelect aria-label="Code" />
+      <Ariakit.ComboboxList
+        alwaysVisible
+        aria-label="Code options"
+        style={{ maxHeight: 80, overflowY: "auto" }}
+      >
+        {codes.map((value) => (
+          <Ariakit.ComboboxItem key={value} value={value} />
+        ))}
+      </Ariakit.ComboboxList>
+    </Ariakit.ComboboxProvider>
+  );
+}
+
+// The same contract without a select: the input is the composite element and
+// the options live in a list outside it, so a collapsed input must not focus
+// them either.
+function FilterCombobox() {
+  const [focusedItems, setFocusedItems] = useState<string[]>([]);
+
+  return (
+    <div>
+      <Ariakit.ComboboxProvider defaultActiveId="filter-banana">
+        <Ariakit.Combobox aria-label="Filter" />
+        <Ariakit.ComboboxList alwaysVisible aria-label="Filter options">
+          {fruits.map((value) => (
+            <Ariakit.ComboboxItem
+              key={value}
+              id={`filter-${value.toLowerCase()}`}
+              value={value}
+              onFocus={() => setFocusedItems((items) => [...items, value])}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+      <FocusedOptions label="Filter" items={focusedItems} />
     </div>
   );
 }
@@ -70,6 +110,8 @@ export default function Example() {
     <>
       <FruitSelect label="Fruit" />
       <FruitSelect label="Fruit without virtual focus" virtualFocus={false} />
+      <CodeSelect />
+      <FilterCombobox />
     </>
   );
 }
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -79,6 +79,42 @@ function CodeSelect() {
   );
 }
 
+const stages = ["Draft", "Review", "Final"];
+
+// Driven from outside the widget, the way an "apply preset" button or a router
+// effect would, so the move happens while focus is elsewhere on the page.
+function StageSelect() {
+  const store = Ariakit.useComboboxStore({
+    defaultSelectedValue: "Draft",
+    virtualFocus: false,
+  });
+
+  return (
+    <div>
+      <Ariakit.ComboboxProvider store={store}>
+        <Ariakit.ComboboxSelect aria-label="Stage" />
+        <Ariakit.ComboboxList alwaysVisible aria-label="Stage options">
+          {stages.map((value) => (
+            <Ariakit.ComboboxItem key={value} value={value} />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+      {/* Safari only focuses a clicked button when it has an explicit
+          tabindex, and these tests assert where the click left focus. */}
+      <button
+        type="button"
+        tabIndex={0}
+        onClick={() => store.move(store.last())}
+      >
+        Apply stage preset
+      </button>
+      <button type="button" tabIndex={0} onClick={() => store.show()}>
+        Open stage list
+      </button>
+    </div>
+  );
+}
+
 // The same contract without a select: the input is the composite element and
 // the options live in a list outside it, so a collapsed input must not focus
 // them either.
@@ -111,6 +147,7 @@ export default function Example() {
       <FruitSelect label="Fruit" />
       <FruitSelect label="Fruit without virtual focus" virtualFocus={false} />
       <CodeSelect />
+      <StageSelect />
       <FilterCombobox />
     </>
   );
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -1,0 +1,54 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+
+const fruits = ["Apple", "Banana", "Grape", "Orange"];
+
+interface FruitSelectProps {
+  label: string;
+  virtualFocus?: boolean;
+}
+
+function FruitSelect({ label, virtualFocus }: FruitSelectProps) {
+  const [focusedItems, setFocusedItems] = useState<string[]>([]);
+
+  return (
+    <div>
+      <Ariakit.ComboboxProvider
+        defaultSelectedValue="Apple"
+        virtualFocus={virtualFocus}
+      >
+        <Ariakit.ComboboxSelect aria-label={label} />
+        {/* The list stays visible while the select is collapsed, which is
+            also what an exit transition leaves behind while it runs. */}
+        <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
+          {fruits.map((value) => (
+            // With virtual focus, DOM focus lands on the item and is handed
+            // straight back to the select, and the select's own blur event is
+            // stopped on the way out. Recording it here is the only way to
+            // observe that an item was focused at all.
+            <Ariakit.ComboboxItem
+              key={value}
+              value={value}
+              onFocus={() => setFocusedItems((items) => [...items, value])}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+      <p>
+        Focused options:{" "}
+        <output aria-label={`${label} focused options`}>
+          {focusedItems.join(", ") || "none"}
+        </output>
+      </p>
+    </div>
+  );
+}
+
+export default function Example() {
+  return (
+    <>
+      <FruitSelect label="Fruit" />
+      <FruitSelect label="Fruit without virtual focus" virtualFocus={false} />
+    </>
+  );
+}
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -10,14 +10,35 @@ interface FruitSelectProps {
 
 function FruitSelect({ label, virtualFocus }: FruitSelectProps) {
   const [focusedItems, setFocusedItems] = useState<string[]>([]);
+  const store = Ariakit.useComboboxStore({
+    defaultSelectedValue: "Apple",
+    virtualFocus,
+  });
+  const open = Ariakit.useStoreState(store, "open");
+  // `ComboboxSelect` forwards composite options to `useComposite`, but it
+  // doesn't list them in its own props type, so they are spread in.
+  const focusOnMoveProps = {
+    focusOnMove: open,
+  } satisfies Ariakit.CompositeOptions;
 
   return (
     <div>
-      <Ariakit.ComboboxProvider
-        defaultSelectedValue="Apple"
-        virtualFocus={virtualFocus}
-      >
-        <Ariakit.ComboboxSelect aria-label={label} />
+      <Ariakit.ComboboxProvider store={store}>
+        <Ariakit.ComboboxSelect
+          aria-label={label}
+          // TODO: remove both workarounds once
+          // https://github.com/ariakit/ariakit/issues/7093 is fixed. The prop
+          // stops the closed moves from presenting an item, and the handler
+          // stops the presentation that virtual focus queues when the
+          // collapsed select itself is focused.
+          {...focusOnMoveProps}
+          onFocus={(event) => {
+            const state = store.getState();
+            if (state.open) return;
+            if (!state.virtualFocus) return;
+            event.preventDefault();
+          }}
+        />
         {/* The list stays visible while the select is collapsed, which is
             also what an exit transition leaves behind while it runs. */}
         <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
--- a/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
+++ b/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
@@ -105,6 +105,58 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await test.expect(q.option("Alpha 25")).toHaveAttribute("data-active-item");
   });
 
+  // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859
+  // Withholding belongs to the widget that owns focus. A move made from
+  // outside keeps presenting immediately, so the focus it moves is
+  // attributable to that call instead of landing on whoever is focused when
+  // the list next opens.
+  test("a move from outside the widget still presents immediately", async ({
+    page,
+    q,
+  }) => {
+    const openList = q.button("Open stage list");
+
+    await q.button("Apply stage preset").click();
+
+    await test.expect(q.option("Final")).toBeFocused();
+    await test
+      .expect(q.combobox("Stage"))
+      .toHaveAttribute("aria-expanded", "false");
+
+    await openList.click();
+
+    await test
+      .expect(q.combobox("Stage"))
+      .toHaveAttribute("aria-expanded", "true");
+    // Opening is observable above, but a presentation left over from the move
+    // would arrive from a passive effect after paint, so cross those frames
+    // before asserting that the click kept its focus.
+    await flushFrames(page);
+    await test.expect(openList).toBeFocused();
+  });
+
+  // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859
+  // Once focus is on an option, moving between options is navigation inside
+  // the list, not focus leaving the control, so the focus ring has to keep up
+  // with the active item even though the list is collapsed.
+  test("arrow keys keep moving focus between options in a collapsed list", async ({
+    page,
+    q,
+  }) => {
+    const list = q.listbox("Stage options");
+    await query(list).option("Draft").click();
+    await test.expect(query(list).option("Draft")).toBeFocused();
+    await test
+      .expect(q.combobox("Stage"))
+      .toHaveAttribute("aria-expanded", "false");
+
+    await page.keyboard.press("ArrowDown");
+
+    const review = query(list).option("Review");
+    await test.expect(review).toHaveAttribute("data-active-item");
+    await test.expect(review).toBeFocused();
+  });
+
   // https://github.com/ariakit/ariakit/issues/7093
   test("a collapsed combobox input never focuses its list", async ({
     page,
--- a/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
+++ b/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
@@ -44,7 +44,82 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
       await test
         .expect(query(list).option("Grape"))
         .toHaveAttribute("data-active-item");
-      await test.expect(focusedOptions).toContainText("Grape");
+      await test.expect(focusedOptions).toHaveText("Grape");
     });
   }
+
+  // https://github.com/ariakit/ariakit/issues/7093
+  // Only virtual focus asks for the active item when the select itself is
+  // focused, which happens while the list is still closed. That request is the
+  // one that has to survive until the list opens, so a pointer open with no
+  // move before it still presents the item.
+  test("opening the collapsed select with a pointer presents its item", async ({
+    q,
+  }) => {
+    const select = q.combobox("Fruit");
+    const focusedOptions = q.status("Fruit focused options");
+
+    await select.click();
+
+    await test.expect(select).toHaveAttribute("aria-expanded", "true");
+    await test.expect(focusedOptions).toHaveText("Apple");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7093
+  // Withholding focus must not also withhold the rest of the presentation: a
+  // list that is already on screen still belongs at the active item.
+  test("typeahead scrolls a collapsed list to the active item", async ({
+    q,
+  }) => {
+    const select = q.combobox("Code");
+    const list = q.listbox("Code options");
+    const scrollTop = () => list.evaluate((element) => element.scrollTop);
+
+    await select.focus();
+    test.expect(await scrollTop()).toBe(0);
+
+    await select.press("z");
+
+    const zulu = q.option("Zulu");
+    await test.expect(zulu).toHaveAttribute("data-active-item");
+    await test.expect.poll(scrollTop).toBeGreaterThan(0);
+    await test.expect(zulu).toBeInViewport();
+    await test.expect(select).toBeFocused();
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7093
+  // Focusing the collapsed select leaves a presentation waiting for the list to
+  // open. Picking another option before that happens has to retire it, or
+  // opening the list would present the option that was active on focus.
+  test("picking an option before opening retires the pending presentation", async ({
+    q,
+  }) => {
+    const select = q.combobox("Code");
+    await select.focus();
+    await q.option("Alpha 25").click();
+    await test.expect(select).toHaveText("Alpha 25");
+
+    await select.click();
+
+    await test.expect(select).toHaveAttribute("aria-expanded", "true");
+    await test.expect(q.option("Alpha 25")).toHaveAttribute("data-active-item");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7093
+  test("a collapsed combobox input never focuses its list", async ({
+    page,
+    q,
+  }) => {
+    const combobox = q.combobox("Filter");
+
+    await combobox.focus();
+
+    await test.expect(combobox).toHaveAttribute("aria-expanded", "false");
+    // The composite can present its active item from a queued microtask or
+    // from a passive effect that runs after paint, so cross those frames
+    // before asserting that focus never reached the list.
+    await flushFrames(page);
+    await test.expect(combobox).toBeFocused();
+    await test.expect(q.status("Filter focused options")).toHaveText("none");
+  });
 });
--- a/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
+++ b/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
@@ -1,0 +1,50 @@
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+
+const labels = ["Fruit", "Fruit without virtual focus"];
+
+withFramework(import.meta.dirname, async ({ test, query }) => {
+  for (const label of labels) {
+    // https://github.com/ariakit/ariakit/issues/7093
+    test(`${label}: a collapsed select never focuses its list`, async ({
+      page,
+      q,
+    }) => {
+      const select = q.combobox(label);
+      const list = q.listbox(`${label} options`);
+      const focusedOptions = q.status(`${label} focused options`);
+
+      await select.focus();
+
+      await test.expect(select).toHaveAttribute("aria-expanded", "false");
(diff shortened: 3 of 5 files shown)


.changeset/300-combobox-select-collapsed-focus.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+Collapsed combobox controls keep focus
```
diegohaz: Use a single sentence changeset.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
```diff
@@ -0,0 +1,125 @@
...
+    await select.click();
+
+    await test.expect(select).toHaveAttribute("aria-expanded", "true");
+    await test.expect(focusedOptions).toHaveText("Apple");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7093
+  // Withholding focus must not also withhold the rest of the presentation: a
+  // list that is already on screen still belongs at the active item.
+  test("typeahead scrolls a collapsed list to the active item", async ({
```
diegohaz, line 71: I don't understand why it scrolls when the list is collapsed. Isn't that a waste of processing power?

Typeahead on a collapsed select should generally update the selected value, but not open/scroll the item in the list.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
```diff
@@ -0,0 +1,125 @@
...
+  // https://github.com/ariakit/ariakit/issues/7093
+  // Focusing the collapsed select leaves a presentation waiting for the list to
+  // open. Picking another option before that happens has to retire it, or
+  // opening the list would present the option that was active on focus.
+  test("picking an option before opening retires the pending presentation", async ({
+    q,
+  }) => {
+    const select = q.combobox("Code");
+    await select.focus();
+    await q.option("Alpha 25").click();
```
diegohaz, line 99: How are we clicking before showing the list? Is this actually testing real use case?

app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
```diff
@@ -0,0 +1,117 @@
...
+        aria-label="Code options"
+        style={{ maxHeight: 80, overflowY: "auto" }}
+      >
+        {codes.map((value) => (
+          <Ariakit.ComboboxItem key={value} value={value} />
+        ))}
+      </Ariakit.ComboboxList>
+    </Ariakit.ComboboxProvider>
+  );
+}
```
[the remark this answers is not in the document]
diegohaz: Validated against the current head. The tests do not pass when the stale request survives, so the extra recorder is not needed.

I disabled the retirement guard in `packages/ariakit-react-components/src/composite/utils.ts` (the `activeId !== resolvedId` cancel at lines 247-250) and ran `pnpm test combobox-select-collapsed-focus`. `picking an option before opening retires the pending presentation` went red:

```
FAIL app/src/sandbox/combobox-select-collapsed-focus/test.ts > picking an option before opening retires the pending presentation
expect(element).toHaveAttribute("data-active-item")
Expected the element to have attribute: data-active-item
Received: null
  ❯ app/src/sandbox/combobox-select-collapsed-focus/test.ts:69:32
```

The surviving request focuses `Alpha 1`, and under virtual focus that focus is exactly what moves `data-active-item` back onto `Alpha 1`, so the existing `Alpha 25` assertion already fails. Restoring the guard returns the file to 5 passed. The same test is the one that goes red in the browser project too.

`CodeSelect` deliberately has no focus recorder: the observable it would add is the one `data-active-item` already carries for that fixture, and the recorder in `FruitSelect` exists only because the virtual-focus round trip there hands focus back before any attribute settles.

Invalid: the premise that the assertions are blind to a stale pending request does not hold at this head.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -166,6 +166,31 @@ function presentItem({
...
+    state: ReturnType<typeof store.getState>,
+    target: HTMLElement,
+  ) => {
+    if (!("open" in state)) return false;
+    if (state.open) return false;
+    const popup = getPopupElement(store);
+    if (!popup?.contains(target)) return false;
+    const { compositeElement } = state;
+    if (!compositeElement) return false;
+    if (popup.contains(compositeElement)) return false;
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and it is worth fixing here, though the suggested predicate does not work as written.

I built a throwaway sandbox: an `alwaysVisible` `ComboboxList` with `virtualFocus={false}`, plus two buttons outside the widget, one calling `store.move("fruit-grape")` and one calling `store.show()`. Recording `document.activeElement` after each step:

| | after the programmatic move | after clicking "Open list" |
| --- | --- | --- |
| base `9aa0b05e3` | the `Grape` option | the "Open list" button |
| this head `5f8ffc99` | the "Move to Grape" button | **the `Grape` option** |

So the withheld request outlives the interaction that created it. The move itself no longer steals focus, which is the improvement, but the focus it still owes is delivered whenever the popup next opens, and it lands on whoever holds focus then. In the second column the user clicked a button and focus left that button for a listbox option.

Real-world shape: an app that syncs a collapsed `ComboboxSelect` from outside the widget, such as an "apply preset" or "restore filters" button, or a router effect, calls `store.move(id)` while the user is elsewhere on the page. Later anything opens the list, and focus is pulled out of the control the user just activated. `focusOnMove` semantics say a programmatic move from outside keeps moving focus, so the base behavior is at least immediate and attributable to the call; the deferred version is not.

The predicate as suggested is not a drop-in. I first re-read focus ownership inside `entersClosedPopup` on every pass, which makes the gate flip between passes. `picking an option before opening retires the pending presentation` then failed in chrome, and not on the pending-request assertion but earlier, on the click itself:

```
Error: expect(locator).toHaveText(expected) failed
Expected: "Alpha 25"
Received: "Alpha 1"
  at app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts:100:31
```

A pass where focus was momentarily off the composite element flipped `focusWithheld` to `false` and delivered the focus that the click was supposed to replace.

Latching the decision once, next to `owner`, holds up:

```ts
const owner = requireFocus
  ? getActiveElement(store.getState().compositeElement)
  : null;
// The withheld focus is only this widget's to withhold. A programmatic move
// made while focus is already outside keeps its focus-on-move behavior, so a
// pending request can't outlive that call and land on an unrelated control.
const startedInside = ownsFocus(store);
```

and in `entersClosedPopup`, before the `isVisible` read:

```ts
if (!startedInside) return false;
```

Verified locally on this head: `pnpm test` 1720 passed, and all 6 chrome tests for this sandbox pass. Both reported triggers stay fixed, because typeahead and the virtual-focus focus handler both run with focus on the select.

No fixture covers a programmatic move made while focus is outside the widget, so this needs a regression alongside the fix.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because DOM focus leaves a control the user just used, while the selected value and ARIA state stay correct.
- Priority: P1 (estimate, 80% confidence), because this PR introduces the deferral that creates the window, and the correction is two lines in the same function.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because it needs a programmatic `move` issued while focus is outside the widget, a popup that stays visible while closed, and a later open.
- Fix effort: Easy (verified), because the latch above is two lines and both suites stay green with it applied.

.changeset/300-combobox-select-collapsed-focus.md
diegohaz: Confirmed. This is a bug fix with no public API change, so the single-sentence form applies and the title plus two explanation paragraphs are more than the entry needs. The second paragraph is also implementation rationale rather than changelog copy.

Suggested replacement for the whole body below the frontmatter:

```md
Fixed [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) and [`Combobox`](https://ariakit.com/reference/combobox) moving DOM focus into their popup while it was still closed, which was visible when the options stayed on screen next to the collapsed control, such as with an [`alwaysVisible`](https://ariakit.com/reference/combobox-list#alwaysvisible) [`ComboboxList`](https://ariakit.com/reference/combobox-list).
```

Link check: `combobox-select`, `combobox`, and `combobox-list` all return 200 on the deployed reference, and `combobox-list` does carry `id="alwaysvisible"`, so the anchor resolves.

No contributor credit sentence is needed here. #7093 was filed by `ariakit-bot`, which the credit policy excludes.

The frontmatter is right as it stands: the change lives in `@ariakit/react-components` and reaches consumers through the `@ariakit/react` re-exports, and both named components are exported from `@ariakit/react`.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
diegohaz: Three things I checked, then the trade-off, because I think your instinct is right.

**It never runs for an ordinary collapsed popup.** `if (!isVisible(element)) return;` at `utils.ts:286` sits above the scroll, and a closed `ComboboxList` is `display: none`. So there is no cost on the normal select at all; the work only exists when the list is genuinely on screen while the control reports `aria-expanded="false"`, which is the `alwaysVisible` and exit-transition shape.

**It is not new here.** On `9aa0b05e3` the same collapsed typeahead already focused the option *and* scrolled the list. This PR removes the focus half and leaves the scroll half exactly where it was, so keeping it is the status-quo option, not an addition.

**It does repeat while withheld.** Because `cancel()` is skipped, the scroll re-runs on later notifications of `activeId`, `items`, `mounted`, `open`, and `unstable_placing`. I instrumented the scroll site and ran this sandbox's happy-dom file: seven withheld scrolls across the five tests, never more than two for the same item. Bounded and cheap in this fixture, but it scales with how many of those notifications arrive before the popup opens, and nothing caps it.

If you want typeahead to leave the collapsed list alone, that is a simplification rather than just a behavior tweak. All three `presentItem` hunks collapse into one early return:

```ts
const focusWithheld = focus && entersClosedPopup(state, element);
if (focusWithheld) {
  const { activeId } = state;
  if (activeId != null && activeId !== resolvedId) return cancel();
  return;
}
```

with `if (focus && !focused) {`, `if (markedOnly && !element.hasAttribute("data-autofocus")) return cancel();`, and the bare `cancel()` all restored to their base form. I ran that variant: `pnpm test` 1720 passed, and the other four tests in this file stayed green. Only `typeahead scrolls a collapsed list to the active item` would have to go, because it is the sole guard for the scroll half.

The one thing lost is that `data-active-item` still moves to the typed option, so a collapsed `alwaysVisible` list would show a highlight the user cannot see until they open it. Opening still scrolls to it, since the pending request completes then.

I would drop it. The invisible-highlight case only exists in the same narrow composition this PR is already narrowing, it removes about twenty lines plus the "keep the request alive so it can scroll again" rationale and the comment defending a repeat scroll, and it matches the native model you described: collapsed typeahead changes the value and nothing else. Your call, though, since keeping it is the strictly behavior-preserving option.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
diegohaz: It is reachable, and it is the interaction the guard actually needs. Details, since the "before showing the list" phrasing is doing the confusing part.

`CodeSelect` renders its `ComboboxList` with `alwaysVisible`, so the options are laid out, hit-testable and clickable while the select still reports `aria-expanded="false"`. Nothing is being clicked through a hidden popup. Given that fixture, clicking a visible option is the ordinary way a user picks from a list that is already on screen, and it is the same premise the rest of the sandbox and #7093 rest on.

It is also not a dead path in the test. The click selects, and line 100 asserts the select reads `Alpha 25` before anything opens. The later `select.click()` then asserts `aria-expanded="true"`, which incidentally proves the option click left the select collapsed rather than toggling it open.

The reason it has to be a click, rather than a second typeahead press, is ownership of the pending request. `Composite` and `CompositeFocusOnMove` share one `usePresentItem`, so a second keypress would create a new presentation that cancels the first through `usePresentItem`'s own `cancel()`, and the retirement guard would never be consulted. `CompositeItem` has its own `usePresentItem` instance, so focusing an item leaves the composite's pending request untouched and only moves `activeId`. That is precisely the state the `activeId !== resolvedId` guard exists for, and it is why disabling the guard turns this test red and no other.

Worth saying plainly: a permanently visible list beside a collapsed select is an unusual composition, so this reads as synthetic. But it is the composition the issue is about, and the alternative shapes I considered either cancel the request by another route or do not move `activeId` at all. If the sandbox comment said that the options stay clickable because the list never unmounts, the test would stop looking like it clicks something invisible.

diegohaz (commented):
Initial review. Baseline is the merge base `9aa0b05e3` to head `5f8ffc99`, expanded to `presentItem`'s call sites in `composite.tsx` and `composite-item.tsx`, the `usePresentItem` ownership model, and the base behavior of every path the new gate touches. Reviewed in a dedicated worktree on the PR branch; no changes were left behind.

The direction is sound and the containment predicate is the right axis: it fixes both reported triggers, keeps #6986 green, and the pending lifetime really is load-bearing for the `ArrowDown` open. One confirmed issue in the deferral it introduces, and one design question of yours that I think should be answered by removing code.

## What I ran

- `pnpm test` on the head: 1719 passed. Re-run at 1720 with a temporary probe sandbox added.
- `pnpm test combobox-select-collapsed-focus`: 5 passed.
- `APP_PORT=4321 NEXTJS_PORT=3000 pnpm -F app test --project=chrome combobox-select-collapsed-focus`: 6 passed. Servers came from `pnpm dev-app` in this worktree, confirmed by fetching the branch-only preview at `/react/previews/combobox-select-collapsed-focus/`.
- Red check of the retirement guard: disabling the `activeId !== resolvedId` cancel turns `picking an option before opening retires the pending presentation` red in happy-dom and in chrome.
- Behavior probe: a throwaway `alwaysVisible` sandbox with external buttons calling `store.move()` and `store.show()`, run against both `9aa0b05e3` and this head.
- Two prototypes, each run against `pnpm test` and the chrome project: withholding the scroll as well as the focus, and gating the predicate on focus ownership.
- Instrumented the withheld scroll site to count repeat scrolls.

## Validation outcomes

**Codex, `composite/utils.ts` L184-186 — confirmed, address now.** The predicate ignores whether the widget owns focus, so a programmatic move made from outside now defers its focus instead of dropping it, and delivers it whenever the popup next opens. Measured against base: on `9aa0b05e3` focus lands on the option at move time and the later open leaves the external button focused; on this head the move is clean but the open pulls focus off the button the user just clicked. The suggested predicate is not a drop-in, though: re-reading ownership on every pass makes the gate flip mid-flight and broke the option click in chrome. Latching `ownsFocus(store)` once beside `owner` works, and `pnpm test` (1720) plus all 6 chrome tests stay green with it. Needs a regression too, since no fixture drives a programmatic move from outside the widget. Assessment is on the thread.

**Your changeset comment — confirmed, address now.** Bug fix with no API change, so the single-sentence form applies. Proposed replacement is on the thread, with all four reference links verified (three 200s, and `combobox-list` really does carry `id="alwaysvisible"`). No credit sentence: #7093 was filed by `ariakit-bot`.

**Your collapsed-scroll question — confirmed simplification, disposition is yours.** The scroll never runs for an ordinary collapsed popup, because `!isVisible(element)` returns above it, so the cost only exists for a list that is genuinely on screen while collapsed. It is also pre-existing on `9aa0b05e3` rather than added here. It does repeat while withheld: seven withheld scrolls across this file's five tests, at most two per item, uncapped in principle. Dropping it collapses all three `presentItem` hunks into a single early return; I ran that variant and only `typeahead scrolls a collapsed list to the active item` would need to go. I lean toward dropping it, for the reasons on the thread, but keeping it is the strictly behavior-preserving choice.

**Your click-before-open question — resolved, no change needed.** The options are clickable because `alwaysVisible` keeps them laid out while `aria-expanded="false"`; the click really selects, and the following `select.click()` proves the list was still collapsed. It has to be a click rather than a second keypress: `Composite` and `CompositeFocusOnMove` share one `usePresentItem`, so another keypress would cancel the pending request through `usePresentItem` itself and never reach the guard, whereas `CompositeItem` owns a separate instance and only moves `activeId`. Worth a sandbox comment saying the list never unmounts, so the test stops reading as a click on something invisible.

**CodeRabbit, `index.react.tsx` L65-80 — invalid.** The claim that the assertions are blind to a stale pending request does not hold. With the guard disabled the test fails on the existing `data-active-item` assertion, because the surviving request focuses `Alpha 1` and virtual focus moves the attribute with it. Evidence on the thread.

## Not published inline

No new inline comments. Every substantive finding I reached was already anchored accurately by an unresolved thread, so the detail is in the replies rather than duplicated.

Two minor observations, both declined as non-goals rather than findings: `entersClosedPopup` ends on `isVisible(target)` and line 286 reads `isVisible(element)` again on the same pass, and `getPopupElement(store)` re-reads `store.getState()` although `state` is already a parameter. Both are trivial and only on the withheld-eligible path.

I also confirmed the gate cannot reach the hover or roving-tabindex presentations: neither `composite-item.tsx` call site passes `focus`, so only `CompositeFocusOnMove` and `Composite`'s virtual-focus focus handler can ever create a withheld request. `@ariakit/solid-components` has no composite yet, so there is no cross-framework counterpart to fix.

## Not validated

Firefox, safari, ios and android Playwright projects; `pnpm test-react18`; `pnpm tsc`; `pnpm lint-fix`; and the app build. The PR description reports these green at this head and I did not re-run them.

packages/ariakit-react-components/src/composite/utils.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`249efbea0`](https://github.com/ariakit/ariakit/commit/249efbea0): the withholding decision is now latched once from whether focus started on the composite element, so a move made while focus is elsewhere keeps presenting immediately instead of leaving a request that lands on whoever is focused when the popup next opens.

It landed narrower than the `ownsFocus(store)` latch proposed above. `ownsFocus` also counts focus that is already on an item, so it withheld navigation *inside* the list: with focus on an option of the collapsed list, arrow keys moved the active item while the focus ring stayed put. Measured in Chromium on the `Stage` fixture, clicking `Draft` and pressing ArrowDown left the value at `Review` with DOM focus still on `Draft`. The latch is therefore `compositeAtStart?.contains(activeAtStart)`, and it is applied at the `focusWithheld` call site rather than inside `entersClosedPopup`, which stays a pure location predicate.

Both directions have regressions in `app/src/sandbox/combobox-select-collapsed-focus/`, in both harnesses: `a move from outside the widget still presents immediately` and `arrow keys keep moving focus between options in a collapsed list`. Each was red-checked against the exact code it protects, the first by removing the latch and the second by widening it back to `ownsFocus(store)`.

.changeset/300-combobox-select-collapsed-focus.md
diegohaz: Fixed in [`c3ac3bdce`](https://github.com/ariakit/ariakit/commit/c3ac3bdce): reduced to the single suggested sentence.

One wording change from the suggestion: it now says focus moved *off the collapsed control* rather than *into the popup while it was closed*. The same PR adds a test where a move made from outside the widget deliberately does focus an option while the combobox still reports `aria-expanded="false"`, so the stricter phrasing would have overstated the fix.

diegohaz:
**Fixed**

- [`249efbea0`](https://github.com/ariakit/ariakit/commit/249efbea0) ([reply](https://github.com/ariakit/ariakit/pull/7098#discussion_r3742633775)): withholding focus is now latched once from whether focus started on the composite element, so a move made from outside the widget keeps presenting immediately. Landed narrower than the suggested `ownsFocus(store)` latch, which also withheld arrow-key navigation between options once focus was already inside the list. Regressions for both directions in `combobox-select-collapsed-focus`, in both harnesses, each red-checked.
- [`c3ac3bdce`](https://github.com/ariakit/ariakit/commit/c3ac3bdce) ([reply](https://github.com/ariakit/ariakit/pull/7098#discussion_r3742634324)): changeset reduced to a single sentence, phrased as focus moving off the collapsed control so it does not overstate the fix.

**Evidence gaps**

- `pnpm test-react18` could not be run locally: it fails in its temp-workspace `prepare` step with a husky `MODULE_NOT_FOUND` before any test loads, and reproduces on a clean `HEAD` with no working-tree changes, so it is environmental. The `Main / Test React 18` CI job covers it.

packages/ariakit-react-components/src/composite/utils.ts
```diff
@@ -125,9 +125,13 @@ function presentItem({
...
-  const owner = requireFocus
-    ? getActiveElement(store.getState().compositeElement)
-    : null;
+  const compositeAtStart = store.getState().compositeElement;
+  const activeAtStart = getActiveElement(compositeAtStart);
+  const owner = requireFocus ? activeAtStart : null;
+  // Only focus that would leave the control itself is this widget's to
+  // withhold, and that is decided once rather than per pass.
+  // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859
+  const startedOnComposite = !!compositeAtStart?.contains(activeAtStart);
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR, because fixing it at this one call site would break the invariant that makes the latch safe.

The premise is right: `getActiveElement` resolves through iframes and `aria-activedescendant` but not shadow roots (`packages/ariakit-utils/src/dom.ts:42`), so with focus inside a shadow root it returns the host and the containment test is false.

But the same blindness already governs `ownsFocus`, which uses that helper and the same containment test, and `ownsFocus` is what computes `requireFocus` at this very call site (`composite.tsx:164`). Composite focus ownership is shadow-blind repo-wide today, not as of this change.

That matters here specifically. The latch is safe because `startedOnComposite` implies `requireFocus`, so every withheld request has a non-null `owner` and `stillOwnsFocus` can abandon it. Resolving the active element through shadow roots only in `startedOnComposite` would make it true while `ownsFocus`, and therefore `requireFocus`, stayed false, leaving a withheld request in a shadow root unguarded, which is the deferred-focus-steal window this latch was added to close.

So the real fix is to make focus ownership shadow-aware in `getActiveElement` and `ownsFocus` together, which changes every composite rather than this predicate. Happy to file that separately if composites inside shadow roots are something Ariakit wants to support; there is no fixture for one today, and the shadow-root sandboxes that exist cover Dialog focus rather than composite navigation.

diegohaz:
**Declined**

- [`#discussion_r3742647603`](https://github.com/ariakit/ariakit/pull/7098#discussion_r3742647603) ([reply](https://github.com/ariakit/ariakit/pull/7098#discussion_r3742662075)): shadow-root focus detection. The premise is correct, but `ownsFocus` is shadow-blind in the same way and is what computes `requireFocus` at this call site, so resolving the active element only in `startedOnComposite` would leave a withheld request in a shadow root unguarded by `stillOwnsFocus`. That is the exact window the latch closes. Making focus ownership shadow-aware belongs in `getActiveElement` and `ownsFocus`, which affects every composite.

packages/ariakit-react-components/src/composite/utils.ts
diegohaz, line 134: This condition is not reflected in the PR description, and it changes the contract the description states.

The Solution section still says "Containment, not open state, is the predicate", and describes withholding purely in terms of a composite element that sits outside its popup. At this head that is only half the rule: withholding also requires that focus started on the composite element. That second half produces two behaviors a reader of the description would not predict, both of which this PR now has tests for:

- a move made from outside the widget presents immediately, instead of leaving a request that lands on whoever is focused when the popup next opens
- arrow keys keep moving DOM focus between options while the list is still collapsed, because focus that is already inside the popup is navigation rather than focus leaving the control

The second is the one worth writing down. At `5f8ffc99` that navigation was broken by the gate, and only this condition restores it, yet the description's framing implies that `#6986`'s "the composite is its own popup" shape is the sole reason a closed popup ever keeps receiving focus. It now also keeps receiving focus when the user is already in it.

The Testing section is stale too. On this head `pnpm test` is 1721, not 1719, and the chrome, firefox and safari totals each grew by the two new tests.

No consumer scenario applies here: this is the merge record and the reassessment log rather than shipped behavior, so I am not inventing a user-facing impact for it.

**Assessment**

- Severity: Low (estimate, 85% confidence), because it misleads readers of the PR record rather than affecting anything that ships.
- Priority: P2 (estimate, 80% confidence), because it should be corrected before merge but blocks nothing.
- Breakage likelihood: Verified present, because I read the current description at this head and the condition is absent from it; there is no runtime breakage.
- Fix effort: Easy (estimate, 90% confidence), because it is a paragraph in the Solution section plus refreshed counts.

diegohaz (commented):
Re-review. Baseline is `5f8ffc99`, the head recorded in the previous review marker, to the current head `c3ac3bdc`, plus the surrounding `presentItem` behavior the new condition reaches. Every review thread on this PR is now resolved, so there were no unresolved candidates to revalidate; this round is the two new commits and what they expose.

Both findings from the last round are fixed, and one of them turned out to be hiding a second defect.

## Verification

- `pnpm test`: 1721 passed.
- `pnpm tsc`: 0 errors.
- `pnpm -F app test --project=chrome` and `--project=safari` for this sandbox: 8 passed each. Servers from `pnpm dev-app` in this worktree on 4322/3001, confirmed by fetching the branch-only `Apply stage preset` markup from the preview.
- Red check of `startedOnComposite`: dropping it from `focusWithheld` turns exactly the two new tests red, and nothing else.
- Vacuity check on the happy-dom duplicate: with the two immediate-presentation assertions removed and the latch dropped, `a move from outside the widget still presents immediately` still fails on its final `expect(openList).toHaveFocus()`. That assertion is doing real work rather than sampling before the regression could appear, so the absence of a `sleep()` there is correct.

## Outcomes

**`249efbea`, the focus-origin latch — fixed and verified.** It landed narrower than the `ownsFocus(store)` latch I suggested, and the narrower version is the better one. `compositeAtStart?.contains(activeAtStart)` treats focus that is already inside the popup as navigation rather than as focus this widget should withhold, which `ownsFocus` would have swallowed. Extracting `compositeAtStart` and `activeAtStart` also leaves `owner` byte-for-byte equivalent to what it computed before, so nothing else in `stillOwnsFocus` shifted.

The second defect is the one worth calling out: `arrow keys keep moving focus between options in a collapsed list` goes red at `5f8ffc99`. The previous head broke arrow-key navigation inside a collapsed `alwaysVisible` list, and I did not catch that last round. It is covered now in both suites.

`StageSelect` is a good fixture for this. The buttons sit outside `ComboboxProvider`, so the move really is external rather than staged through an internal handle, and the `tabIndex={0}` on them is the documented Safari click-focus requirement, which I confirmed by running the safari project rather than taking it on faith.

**`c3ac3bdce`, the changeset — fixed and verified.** Single past-tense sentence, all four reference links present. The wording change from the suggestion is right, and is the more accurate description now: "off the collapsed control" rather than "into their popup while it was closed" is what the new arrow-key test would otherwise contradict.

**The collapsed-scroll question — resolved as-is.** Kept the current behavior, which is the strictly behavior-preserving option and was always yours to call. No new evidence this round, so I am not reopening it.

**The click-before-open question and the CodeRabbit thread — closed, no change needed.** CodeRabbit withdrew its finding after the red check.

**The shadow-root thread — resolved, not re-adjudicated.** I read the reasoning rather than re-litigating a resolved thread, and it does not change anything here: when `getActiveElement` returns a host outside the composite, `startedOnComposite` is `false`, so the request stops withholding and falls back to the pre-existing behavior instead of producing a new failure mode. Degrading toward base is the safe direction for a gate like this.

## New finding

One, published inline: the Solution section of the PR description still presents containment as the entire predicate and never mentions the focus-origin condition, even though that condition is what makes an external move present immediately and what keeps collapsed arrow-key navigation working. The Testing counts are stale too (1721, not 1719). Low severity, P2, and it is about the merge record rather than shipped behavior.

## Not validated

Firefox, ios and android Playwright projects, `pnpm test-react18`, `pnpm lint-fix`, and the app build. I ran chrome and safari for this sandbox only, on the reasoning that the new assertions are focus- and click-behavior assertions and safari is where those diverge.

packages/ariakit-react-components/src/composite/utils.ts
diegohaz: Fixed in the PR description, since the correction is to the record rather than to shipped behavior.

The Solution section now opens with both conditions instead of one, and the containment paragraph is explicitly the second of the two. A third paragraph covers the first condition and the two behaviors you named: a move made from outside presents immediately the way `focusOnMove` documents, and arrow keys keep moving DOM focus between options of a collapsed list because that is navigation inside the popup rather than focus leaving the control. It closes by saying outright that the composite being its own popup is no longer the only reason a closed popup keeps receiving focus, so the `#6986` shape is not read as the sole case.

Testing counts refreshed against this head, all re-run rather than adjusted arithmetically: `pnpm test` 1721, chrome 753, firefox 612, safari 501. Your arithmetic matched exactly in each browser.

One change beyond what you asked for: the `pnpm test-react18` count is gone, replaced by a stated evidence gap. That suite cannot run locally at the moment because its temp workspace fails `prepare` with a husky `MODULE_NOT_FOUND` before any test loads, and it reproduces on a clean checkout, so I could not honestly claim a number for it. The `Main / Test React 18` CI job covers it and passes.

diegohaz:
**Fixed**

- PR description ([reply](https://github.com/ariakit/ariakit/pull/7098#discussion_r3743774936)): the Solution section now states both withholding conditions instead of only containment, and adds a paragraph for the ownership condition covering the outside-move and collapsed-list-navigation behaviors, so the `#6986` shape is not read as the only reason a closed popup keeps receiving focus. Testing counts re-run against this head: `pnpm test` 1721, chrome 753, firefox 612, safari 501.

**Evidence gaps**

- `pnpm test-react18` is now listed as a gap in the description rather than a count. It fails in its temp-workspace `prepare` step with a husky `MODULE_NOT_FOUND` before any test loads, and reproduces on a clean checkout, so it is environmental. The `Main / Test React 18` CI job covers it.

PR #7101, opened 2026-08-08
Treat modified navigation keys as keyboard modality

### LINKED ISSUE #7099: Alt-modified keys that move composite focus are treated as pointer modality
### Problem

`Focusable` decides between keyboard and pointer modality from a global `keydown` listener that ignores any `Alt`-modified key:

https://github.com/ariakit/ariakit/blob/9aa0b05e3103055b716396156d3ed9f772f0b161/packages/ariakit-react-components/src/focusable/focusable.tsx#L192-L197

Composite navigation does not apply the same filter. `CompositeItem` dispatches its arrow key actions from a map keyed on `event.key` alone, with no modifier check:

https://github.com/ariakit/ariakit/blob/9aa0b05e3103055b716396156d3ed9f772f0b161/packages/ariakit-react-components/src/composite/composite-item.tsx#L464-L489

So `Option+ArrowRight` (or `Alt+ArrowRight`) really does move focus between composite items, in every browser, while the modality listener ignores the same keystroke. Once the user has touched the mouse, which is the normal case since modality only flips to pointer on `mousedown`, focus moves by keyboard but the element that receives it is still classified as pointer-focused.

### Observed

Measured on the `toolbar-keyboard-navigation` sandbox, in both headless WebKit and headless Chromium driven by Playwright, using its `Basic formatting` toolbar. `Undo` is clicked first so modality is pointer, which is the normal state once the mouse has been used. Focus then moves to `Bold`, skipping the disabled `Redo`, and `data-focus-visible` is read after the animation frame that applies it:

| Gesture | Focus moves to `Bold` | `data-focus-visible` on `Bold` |
| --- | --- | --- |
| `ArrowRight` | yes | true |
| `Alt+ArrowRight` | yes | **false** |

Both browsers produce identical results, so this is not engine-specific.

Both gestures perform the same focus movement through the same code path. Only the modality classification differs, so the focus ring that components draw from `data-focus-visible`, and focus-driven behavior such as `Tooltip`, are lost for the modified gesture.

This is not limited to `Toolbar`. It applies to every composite widget whose items move focus with arrow keys, and to `Home`, `End`, `PageUp`, and `PageDown` for the same reason.

### Expected outcome

A keystroke that Ariakit itself uses to move focus should count as keyboard navigation. Either the modality listener should recognize those gestures, or composite navigation should ignore `Alt`-modified keys so that the two surfaces agree.

The second option is a behavior change for consumers who rely on `Alt+Arrow` navigating a composite today, so the first is likely preferable, but the two surfaces should be reconciled deliberately rather than left inconsistent.

### Context

Found while reviewing the fix for #7094, which covers the related but narrower case of Safari's `Option+Tab`. That fix scopes its exception to `key === "Tab"` on Safari, so it neither introduces nor resolves this one.

This finding is pre-existing on `main` at [`9aa0b05e3`](https://github.com/ariakit/ariakit/commit/9aa0b05e3103055b716396156d3ed9f772f0b161) and is independent of #7096. The `Alt`-modified non-`Tab` code path is identical before and after that pull request, because those keys return at the same point in both versions.

#7096 deliberately does not add regression coverage for this behavior. Asserting the current outcome would lock in the inconsistency described above, which should be resolved here first.

diegohaz:
## Motivation

`Focusable` decides between keyboard and pointer modality from a global `keydown` listener that ignores every modified keystroke:

```ts
function onGlobalKeyDown(event: KeyboardEvent) {
  if (event.metaKey) return;
  if (event.ctrlKey) return;
  if (event.altKey) return;
  isKeyboardModality = true;
}
```

Composite navigation applies no such filter. [`CompositeItem`](https://ariakit.com/reference/composite-item) and [`Composite`](https://ariakit.com/reference/composite) dispatch their arrow, `Home`, `End`, `PageUp` and `PageDown` actions from a map keyed on `event.key` alone, and the `Home` and `End` handlers deliberately implement `Ctrl+Home` and `Ctrl+End` for grids, which is the standard grid pattern.

So `Alt+ArrowDown` really does move focus between composite items, and `Ctrl+Home` really does jump to the first cell of a grid, while the modality listener ignores both gestures. Once the user has touched the mouse, which is the normal case since modality only flips to pointer on `mousedown`, focus moves by keyboard but the element that receives it is still classified as pointer-focused. Everything keyed off that classification is lost: the focus ring components draw from [`data-focus-visible`](https://ariakit.com/guide/styling#data-focus-visible), and focus-driven behavior such as [`Tooltip`](https://ariakit.com/reference/tooltip).

The same inconsistency exists in a second place. `Focusable` has another modifier guard in `onKeyDownCapture`, which applies `data-focus-visible` to an element that already has focus. A plain `ArrowDown` pressed on the last item of a non-looping composite makes that item focus-visible today, while `Alt+ArrowDown` on the same item does not, even though neither gesture moves focus. That guard is also the only mechanism available to composites with [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus), where DOM focus stays on the composite element and no focus event ever fires to re-apply the attribute.

This applies to every composite widget whose items move focus with arrow keys, including [`Toolbar`](https://ariakit.com/reference/toolbar), [`Menu`](https://ariakit.com/reference/menu), [`Tab`](https://ariakit.com/reference/tab), [`Select`](https://ariakit.com/reference/select) and [`Combobox`](https://ariakit.com/reference/combobox).

## Solution

Treat the keys Ariakit itself navigates with as keyboard navigation regardless of modifiers, since those keys are never typing. A single predicate in `packages/ariakit-react-components/src/focusable/__utils.ts` holds the allowlist, and both guards use it:

```ts
// Keys that composite widgets move focus with. They're never typing, so they
// count as keyboard navigation whether or not a modifier is held. Keep in sync
// with the key maps in composite.tsx and composite-item.tsx.
export function isCompositeMoveKey(key: string) {
  return (
    key === "ArrowUp" ||
    key === "ArrowRight" ||
    key === "ArrowDown" ||
    key === "ArrowLeft" ||
    key === "Home" ||
    key === "End" ||
    key === "PageUp" ||
    key === "PageDown"
  );
}
```

`combobox-select.tsx` had a verbatim copy of that predicate, so it now imports the shared one and there is a single definition. The module is `__`-prefixed, which the build skips when collecting public files, so it adds no published entry point, and it imports nothing, so it adds no dependency.

In `onGlobalKeyDown` it sets keyboard modality above the existing modifier guards, which are otherwise untouched. In `onKeyDownCapture` the three modifier early returns become one condition:

```ts
const hasModifier = event.metaKey || event.altKey || event.ctrlKey;
if (hasModifier && !isCompositeMoveKey(event.key)) return;
```

The change is a no-op for every keystroke without a modifier, in both guards. Only the eight allowlisted keys behave differently, and only when a modifier is held. `Ctrl+S`, `Cmd+C` and `Option+e` still leave the modality untouched.

`Tab` is deliberately left out of the allowlist, so the Safari `Option+Tab` exception added in #7096 stays necessary and is preserved exactly, including `Ctrl+Alt+Tab` and `Cmd+Tab` still returning early.

The fix covers all three modifiers rather than only `Alt`. The issue is framed around `Alt`, but `Ctrl+Home` and `Ctrl+End` are deliberate, implemented grid gestures that `app/src/sandbox/combobox-select-grid/test.ts` already exercises, so an `Alt`-only fix would repair the rarer gesture and leave the standard one silently ring-less.

Because `onKeyDownCapture` belongs to every `Focusable`, the behavior is not limited to composite items: a mouse-focused plain [`Button`](https://ariakit.com/reference/button) now becomes focus-visible on a modified navigation key too. That is parity with the unmodified keys, which that guard has always allowed, since it has no key filter at all.

## Preview

Chromium on the docs code-block tab list, which draws its focus style from `data-focus-visible`. A tab is clicked with the mouse, so the modality is pointer, and then `Alt+ArrowRight` moves focus to the next tab. Both states select the tab; only the fixed build draws the focus style.

Code-block tab list before and after the fix. In both, the checkbox-card.tsx tab is selected after pressing Alt+ArrowRight. Before the fix it renders as an ordinary selected tab with no focus style; after the fix it renders with the blue focus style.

## Workaround

Until this ships, consumers can re-announce the gesture on the focused item as an unmodified, key-less `keydown` from the composite's capture phase. Ariakit then records keyboard modality and applies focus-visible on both surfaces, and because the replayed event carries no key, composite navigation ignores it and cannot move focus twice.

```tsx
import type { KeyboardEvent as ReactKeyboardEvent } from "react";

const navigationKeys = new Set([
  "ArrowUp",
  "ArrowRight",
  "ArrowDown",
  "ArrowLeft",
  "Home",
  "End",
  "PageUp",
  "PageDown",
]);

// TODO: Remove once https://github.com/ariakit/ariakit/issues/7099 is released.
function announceKeyboardModality(event: ReactKeyboardEvent<HTMLElement>) {
  if (!event.altKey && !event.ctrlKey && !event.metaKey) return;
  if (!navigationKeys.has(event.key)) return;
  const target = event.target as HTMLElement;
  const view = target.ownerDocument.defaultView;
  if (!view) return;
  target.dispatchEvent(new view.KeyboardEvent("keydown", { bubbles: true }));
}

<Ariakit.Composite onKeyDownCapture={announceKeyboardModality}>
  {/* composite items */}
</Ariakit.Composite>;
```

Verified in happy-dom and in Chromium, Firefox and WebKit against the reproduction in this pull request.

This workaround is deliberately narrower than the library fix in some respects and broader in others, so it is not a complete substitute for it.

- **It covers only the eight composite navigation keys.** Modified activation keys are not covered: `Command`'s Enter and Space path does not check modifiers, so `Alt+Enter` still activates a composite item while the item stays classified as pointer-focused.
- **It infers modality from the keystroke, not from whether Ariakit actually navigated.** It marks keyboard modality even when the gesture is a no-op for that widget: an `orientation: "vertical"` composite receiving a modified `ArrowLeft` or `ArrowRight`, `Home` and `End` when the composite element is a text field, and `PageUp` and `PageDown` when there is no scrolling element. In those configurations the snippet produces a focus ring where Ariakit did nothing.
- **It is unsuitable on a text-field composite or a textbox item.** [`Combobox`](https://ariakit.com/reference/combobox) renders an `input`, and Ariakit deliberately declines to navigate when the caret is not at the edge of the value. Applying the snippet there makes `Option+ArrowLeft`, the macOS word-jump, flip the modality flag and mark a mouse-focused input focus-visible while the user is only moving a caret.
- **It has no self-target check.** As a capture-phase handler on the composite, it also fires when the target is a focusable descendant that is not a composite item, such as a button inside an item. Ariakit declines those, but the snippet still replays and marks that inner element focus-visible. Gate on the target being the active item, or move the handler onto the item, if that matters.
- **It is verified for roving tabindex, not for [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus).** When the composite base element is the keydown target, the replay also passes through Ariakit's keyboard event proxy, which focuses the active item and re-fires the key-less event on it before the real key is handled. By inspection this is duplicated work rather than new behavior, but it is untested here.
- **Every covered gesture emits one extra untrusted `keydown`** that bubbles to `document`, so app hotkey managers, analytics and extensions observe it. It carries no key, so handlers that branch on `event.key` ignore it, and it is inert inside Ariakit.
- **It does not cross a shadow root.** The replay is `composed: false`, unlike a real key press, so with a composite inside a shadow root the global half silently does nothing and only the per-element half still works.
- **It does not check `event.defaultPrevented`,** so an ancestor that calls `preventDefault()` without stopping propagation still gets a replay.
- **It is per-composite, not global,** and must be attached to every composite that needs it.
- **It flips the process-wide modality flag,** so until the next `mousedown` any element focused programmatically anywhere on the page becomes focus-visible. A real key press does the same, so this is expected rather than a scoped change.
- **Do not preserve `event.key` in the replay.** The empty key is what keeps it inert; forwarding the real key matches the navigation map and moves focus twice.
- **It relies on Ariakit not filtering untrusted keydowns,** which is true today. It also fires once per auto-repeat while a key is held.

## Regression test

`app/src/sandbox/composite-navigation` gains four tests, in the authoritative Playwright file and the happy-dom duplicate. Each clicks first so the modality is pointer, which is the normal state once the mouse has been used.

1. `Alt+ArrowDown` moves focus from `Apple` to `Grape`, which must be focus-visible. This covers the global listener.
2. `Alt+ArrowDown` on `Orange`, the last item of the non-looping composite, moves nothing and fires no focus event, so only the item's own keydown handling can apply the attribute. This covers `onKeyDownCapture`, which the first test cannot reach.
3. `Ctrl+Home` on a grid moves focus to the first cell, which must be focus-visible. This covers the modifier generalization on a gesture Ariakit implements deliberately.
4. `Ctrl+c` on `Orange` must leave both the item and the next element to receive focus non-focus-visible. This is a boundary control rather than a failing repro: it passes today and pins the half of the contract the fix must not break, so deleting the modifier guards outright cannot pass.

The first three fail on the `data-focus-visible` assertion and never on focus movement, in happy-dom and in all three desktop Playwright projects, which attributes the failure to modality classification rather than to focus never moving.

Those four tests all pass with the userland workaround as well, so they do not uniquely exercise the library change. `app/src/sandbox/focusable-click-focus`, whose fixture contains no composite at all, therefore also gains a test in both files: click the plain `Button`, assert it is not focus-visible, press `Alt+ArrowDown`, and assert it is. The composite-scoped workaround could not have made that pass, so it pins the fix to `Focusable` itself. It was red-checked in both environments by temporarily reverting the library change.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the navigation-key allowlist direction</summary>

**Assessment**

The formal gate produced findings on three consecutive rounds for the reproduction commit, which triggered the mandatory reassessment. Of the nine findings, one was a design-level challenge, resolved as invalid in its central claim because it assumed the fix would drop `onKeyDownCapture`'s modifier guard rather than add an allowlist to it. The remaining eight were comment accuracy, one missing boundary control, one test-title overclaim, and file ordering. No round asked to restructure the guards or increased library complexity, so the repeated cycles reflect convergence on test quality rather than over-complexity.

The library cost is one pure predicate, one statement hoisted above the existing modifier returns in `onGlobalKeyDown`, and one condition in `onKeyDownCapture`. No new state, no new public API, no new branching in composite code. The issue is a confirmed accessibility bug on a stable surface, and its impact is wider than the title suggests because `Ctrl+Home` and `Ctrl+End` are standard grid gestures Ariakit implements deliberately.

**Decision**

Continue. Keep all four tests, keep `composite-navigation` as the fixture, and keep the all-modifier scope, since narrowing to `Alt` would leave a tested, standard grid gesture broken. Keep the near-identical `isCompositeMoveKey` predicate in `combobox-select.tsx` private rather than extracting a shared helper, because sharing would either invert layering or add a new export to the published `@ariakit/utils`. One simplification adopted: every comment is self-contained and names its own mechanism, replacing the cross-reference pattern that produced two of the nine findings. No disposition was reverted; the rationale for keeping the second test was corrected to rest on the measured parity result and on `virtualFocus` composites rather than on a misread sentence in the issue.

</details>

<details>
<summary>Cycle 8: Continue, with the library diff and test set frozen</summary>

**Assessment**

Three further finding-bearing rounds ran on the library-fix commit. Severity trended down sharply: round 1 produced one substantive coverage finding and one structural challenge, round 2 produced a rename and a serial comma, and round 3 produced a single false sentence in the changeset. The library code has not changed behaviorally since the first round of this commit; every round since moved a name or a comment. No round has challenged the design on its merits since cycle 3.

The proportionality conclusion from cycle 3 still holds. The fix is 20 lines in one file, and its blast radius is narrower than it looks: `isAlwaysFocusVisible` already forces focus-visible on text fields, `select`, `contenteditable` and Ariakit's own combobox role, so the change cannot newly affect them, and for the elements it can affect, `onKeyDownCapture` has never had a key filter, so an unmodified `ArrowDown` on a mouse-focused button already applied the attribute.

**Decision**

Superseded in part by the review response below, which adopted the extraction after review surfaced a third option neither objection covered. The rest of this decision still stands.

Continue, and freeze the library diff and the test set. The remaining changeset edit is the only landing change. The duplicated predicate stays private for the second time: `packages/ariakit-react-components/package.json` publishes `./composite/utils`, and that module imports `@ariakit/store`, so hosting the predicate there would pull the composite store graph into every `Focusable` and `Button` consumer, while `@ariakit/utils` would mean a new export on a published package. The remedy that was adopted instead is discoverability, by giving the two copies the same name. The non-composite coverage in `focusable-click-focus` is kept: the issue-fix workflow requires focused regression coverage when the sandbox tests only prove the behavior is repairable by either the workaround or the fix, and package-level component regression tests are forbidden by the repository's own rules, so a non-composite sandbox test is the only permitted remedy.

One process gap is recorded rather than hidden: the preserved artifact files captured the findings and dispositions for the reproduction commit's first three rounds and for all of the library-fix commit's rounds, but not for the reproduction commit's fourth round or the workaround commit's first round. Both were addressed in full at the time, and both are described in this pull request, but they were not written to the durable artifact before later work replaced the snapshot.

</details>

## Review response

Two review threads were addressed after the reassessments above, so the two decisions they touch are recorded here rather than by rewriting the point-in-time entries.

The changeset was three sentences of implementation rationale. A single-paragraph entry should be one short past-tense summary sentence, and the rationale already lives in this description, so it was shortened.

The duplicated predicate was extracted after all. Both earlier reassessments declined extraction because the only homes considered were `composite/utils`, which imports `@ariakit/store` and would pull the composite store graph into every `Focusable` and `Button` consumer, and `@ariakit/utils`, which would add an export to a published package. Review pointed out a third home neither objection covers: a `__`-prefixed private module. `packages/ariakit-scripts/src/build.ts` skips those while collecting public files, so `focusable/__utils.ts` adds no published entry point, and the predicate imports nothing, so it adds no dependency. `checkbox/__utils.ts`, `combobox/__utils.ts` and `popover/__utils.ts` are existing precedent, and `combobox-select.tsx` already imported `../popover/__utils.ts`. Verified after the change: the generated `exports` maps are unchanged, and `npm pack` ships the new file exactly like the three existing private modules.

That closes the drift hazard the earlier decision accepted. `combobox-select.tsx` gates `shouldProxyCompositeEvent` on the same predicate, so a ninth navigation key added to the composite key maps would previously have reached only one copy, silently leaving `ComboboxSelect` proxying the wrong set with no test failing.

Fixes #7099


.changeset/300-modified-navigation-key-focus-visible.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) treating the keys that move composite focus as a pointer interaction when a modifier was held, so components built on it such as [`CompositeItem`](https://ariakit.com/reference/composite-item) and [`ToolbarItem`](https://ariakit.com/reference/toolbar-item) now receive [`data-focus-visible`](https://ariakit.com/guide/styling#data-focus-visible) when focus reaches them that way. Arrow keys and `Home`, `End`, `PageUp`, and `PageDown` move focus through a composite whether or not a modifier is held, including `Ctrl+Home` and `Ctrl+End` on a grid, so they now count as keyboard navigation. This applies to every [`Focusable`](https://ariakit.com/reference/focusable) component, not only composite items.
```
diegohaz: This changeset is too big.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -189,14 +189,38 @@ function onGlobalMouseDown(event: MouseEvent) {
   }
 }
 
+// Keys that composite widgets move focus with. They're never typing, so they
+// count as keyboard navigation whether or not a modifier is held. Keep in sync
+// with the key maps in composite.tsx and composite-item.tsx.
+function isCompositeMoveKey(key: string) {
```
diegohaz: This is a verbatim copy of `isCompositeMoveKey` at `packages/ariakit-react-components/src/combobox/combobox-select.tsx:56`, same eight comparisons in the same order, in the same package. Neither copy points at the other: this comment names `composite.tsx` and `composite-item.tsx` as the sync targets, and the `combobox-select.tsx` copy carries no comment at all, so the only thing linking them is the shared name.

Concretely, if a later change teaches `Composite` and `CompositeItem` a ninth navigation key, whoever adds it follows this comment to the two key maps, updates this copy, and has nothing pointing them at `combobox-select.tsx`. That copy gates `shouldProxyCompositeEvent`, which returns `isCompositeMoveKey(event.key)` once a `ComboboxSelect` popup is open and an input exists, so the new key would silently stop being proxied to the composite: a user pressing it in an open combobox-select would get the input's default behavior instead of moving through the list, while `Focusable` would already be treating the same key as keyboard navigation. Nothing in either feature's tests would fail, because each copy is only exercised through its own feature.

The rationale recorded in the PR description rules out `composite/utils` (it imports `@ariakit/store`, so it would pull the composite store graph into every `Focusable` and `Button` consumer) and `@ariakit/utils` (a new export on a published package). Both objections hold, but there is a third home neither covers. `packages/ariakit-scripts/src/build.ts:135` skips any entry whose name starts with `__` while collecting public files, so private modules never reach the generated `exports` map: `checkbox/__utils.ts`, `combobox/__utils.ts`, and `popover/__utils.ts` appear nowhere in `packages/ariakit-react-components/package.json`. And `combobox-select.tsx` already reaches across directories for exactly this kind of helper, on lines 24 and 28:

```ts
import { withDefaultButtonType } from "../button/utils.ts";
import { getBasePlacement } from "../popover/__utils.ts";
```

So a new `src/focusable/__utils.ts` holding the predicate adds no published export and no dependency, since the predicate imports nothing, and it gives the "keep in sync" comment a single home that both call sites import.

```ts
// packages/ariakit-react-components/src/focusable/__utils.ts

// Keys that composite widgets move focus with. They're never typing, so they
// count as keyboard navigation whether or not a modifier is held. Keep in sync
// with the key maps in composite.tsx and composite-item.tsx.
export function isCompositeMoveKey(key: string) {
  return (
    key === "ArrowUp" ||
    key === "ArrowRight" ||
    key === "ArrowDown" ||
    key === "ArrowLeft" ||
    key === "Home" ||
    key === "End" ||
    key === "PageUp" ||
    key === "PageDown"
  );
}
```

If you would rather keep two copies, the cheap version of the same fix is to have each comment name the other file, so the hazard is discoverable from either side instead of resting on the shared name.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the two copies agree today and nothing user-visible changes until one of them drifts.
- Priority: P2 (estimate, 80% confidence), because this PR is what creates the duplication, but the extraction is behavior-preserving and safe to land separately.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because it requires a future change to the composite navigation key set, and each copy is exercised only through its own feature, so a drifting edit would not turn either suite red.
- Fix effort: Easy (estimate, 90% confidence), because it is one new private module plus two imports, with no behavior change and no packaging change.

.changeset/300-modified-navigation-key-focus-visible.md
diegohaz: Confirmed, and it is a guideline problem rather than a taste one: a single-paragraph changeset should be one short past-tense summary sentence, and this entry is three sentences and 81 words. The second sentence explains the mechanism (which keys move composite focus, and that `Ctrl+Home` and `Ctrl+End` are grid gestures) and the third states the scope. Both read as implementation rationale, which is what the PR description is for.

Everything factual in the entry does check out, so this is a length problem only. `Focusable`, `CompositeItem`, and `ToolbarItem` are all re-exported by `@ariakit/react` (`packages/ariakit-react/src/focusable.ts:5`, `composite.ts:26`, `toolbar.ts:21`), so both frontmatter packages can name them. All four links resolve: the three reference pages return `200`, and `https://ariakit.com/guide/styling` does contain `id="data-focus-visible"`. No contributor credit is missing either, since #7099 was opened by you and only bots commented on it.

Suggested replacement, one sentence, past tense, keeping the two representative derived components the guideline allows for a single-paragraph entry:

```md
Fixed [`Focusable`](https://ariakit.com/reference/focusable) components, including [`CompositeItem`](https://ariakit.com/reference/composite-item) and [`ToolbarItem`](https://ariakit.com/reference/toolbar-item), not receiving [`data-focus-visible`](https://ariakit.com/guide/styling#data-focus-visible) when focus moved with a modified navigation key, such as `Alt+ArrowDown` between composite items or `Ctrl+Home` on a grid.
```

Naming `Focusable` first already implies the third sentence's point. If you want to keep that point explicit, because a mouse-focused plain [`Button`](https://ariakit.com/reference/button) now becomes focus-visible on `Alt+ArrowDown` too, the guideline's other shape fits: a short title line, then an explanation paragraph, rather than one long paragraph.

**Assessment**

- Severity: Low (verified, because the only impact is changelog readability for the next release).
- Priority: P1 (estimate, 85% confidence), because the changeset ships with this PR and rewriting it now costs one line.
- Breakage likelihood: Likely (verified, because the entry is already in the diff and will render in the changelog exactly as written).
- Fix effort: Easy (estimate, 95% confidence), because it is a single-line rewrite of the changeset body with no code or test impact.

diegohaz (commented):
Initial review. Baseline is the merge base [`d82fbbac`](https://github.com/ariakit/ariakit/commit/d82fbbaccabbcf9de40478f33f9691bb1743a98e) to head [`3c714111`](https://github.com/ariakit/ariakit/commit/3c71411155bdeebf9c49e78aa5ac44d0359c9e71), all six changed files, plus the surrounding context: `focusable.tsx` in full, the `Composite` and `CompositeItem` key maps, `combobox-select.tsx`, both sandbox fixtures, the Vitest project and fixture-loader setup, the `@ariakit/test` `press` signatures, the `@ariakit/react` re-exports, and the packaging rules in `packages/ariakit-scripts/src/build.ts`. No prior review from this workflow exists on this PR, so there was no incremental baseline to use.

The library change holds up. The allowlist matches the key maps exactly: `composite.tsx:501-508` and `composite-item.tsx:465-486` both dispatch on `ArrowUp`, `ArrowRight`, `ArrowDown`, `ArrowLeft`, `Home`, `End`, `PageUp`, and `PageDown` with no modifier filter, and `Home` and `End` branch on `event.ctrlKey` for grids, so the description's premise is accurate rather than inferred. `Tab` staying out of the allowlist keeps the Safari `Option+Tab` path from #7096 intact, and `app/src/sandbox/focusable-tooltip-7094/test-safari.ts` still covers it.

## Validation performed

- Ran `pnpm test app/src/sandbox/composite-navigation/test.ts app/src/sandbox/focusable-click-focus/test.ts` on the PR head: 10 passed.
- Reproduced the red check independently. Restoring `focusable.tsx` to its baseline content and rerunning the same command fails 4 tests, and every failure is on the `data-focus-visible` assertion rather than on focus movement or `toHaveFocus`. The fourth test, `keeps pointer modality on a modified non-navigation key`, passes in both states, which is what makes it a real boundary control: deleting the modifier guards outright could not pass it. The worktree was restored to the PR head afterwards and `git status` is clean.
- Did not rerun the Playwright suites locally, because all of `App / Test Chrome`, `Test Firefox`, `Test Safari (1/2)`, `Test Safari (2/2)`, `Test Android`, and `Test iOS` are green on this head.
- Verified the changeset's public surface and links: `Focusable`, `CompositeItem`, and `ToolbarItem` are re-exported by `@ariakit/react` (`packages/ariakit-react/src/focusable.ts:5`, `composite.ts:26`, `toolbar.ts:21`); the three reference pages return `200`; and `https://ariakit.com/guide/styling` contains `id="data-focus-visible"`.
- Checked the `data-focus-visible` documentation in `app/src/guides/ariakit-react/2-styling.mdx:165-171`. It describes the attribute as applying on keyboard interaction without enumerating keys, so the fix makes behavior match the docs more closely and no docs change is owed.

## Unresolved thread validated in this run

`.changeset/300-modified-navigation-key-focus-visible.md` ([discussion_r3742241850](https://github.com/ariakit/ariakit/pull/7101#discussion_r3742241850)), "This changeset is too big": confirmed, disposition Address now, replied on the thread with a one-sentence replacement. The entry is three sentences and 81 words in a single paragraph, where the guideline asks for one short past-tense summary sentence. Its content is accurate and its links resolve, and no contributor credit is missing since #7099 was opened by you and only bots commented, so the finding is about length alone.

## Newly discovered finding published inline

One, on the duplicated `isCompositeMoveKey` predicate. The version added in `focusable.tsx` is a verbatim copy of the one at `combobox-select.tsx:56`, and neither copy references the other, so the sync hazard the new comment warns about now spans three files with only the shared name linking two of them. The PR description rules out `composite/utils` and `@ariakit/utils` for reasons that both hold, but `packages/ariakit-scripts/src/build.ts:135` skips any entry whose name starts with `__` when collecting public files, so a private `focusable/__utils.ts` would add no published export and no dependency. `combobox-select.tsx:24,28` already imports across directories from `../button/utils.ts` and `../popover/__utils.ts`, so the pattern is established in that exact file. Assessment is in the inline comment.

## Candidates examined and not published

- **Modality precision as an alternative design.** The obvious alternative is to have `Composite` and `CompositeItem` announce keyboard modality only when they actually navigate, instead of `Focusable` inferring it from the key. That is more precise, but it cannot produce the behavior the second regression test pins: on the last item of a non-looping composite, Ariakit deliberately does not move, and the PR wants parity with the unmodified key there. Reproducing that would require the composite to announce whenever the key merely maps to an action, which is the same allowlist evaluated one layer up, at the cost of a new cross-module export and wiring in two components. Invalid as a finding: the PR's approach is the smaller one for the same behavior.
- **Global modality flipping on browser-level chords.** `Ctrl+PageUp` and `Ctrl+PageDown` switch browser tabs, `Alt+ArrowLeft` and `Alt+ArrowRight` are back and forward, and `Option+ArrowLeft` is the macOS word jump. All of them now set keyboard modality where they previously did not. Not a finding: `onGlobalKeyDown` has never filtered by key, so the unmodified `ArrowLeft`, `PageUp`, and `PageDown` already flip modality in exactly those situations, and the change only removes an inconsistency between the modified and unmodified forms of the same gesture.
- **No coverage for `Meta`-modified navigation keys.** The tests cover `Alt` and `Ctrl` but not `Cmd`. Declined non-goal: `isCompositeMoveKey` is modifier-agnostic and both guards test the same predicate, so a third modifier case would exercise no distinct code path.
- **Wording of the new `onKeyDownCapture` comment.** "This guard also matters when the key moves nothing" reads as describing the early return, when the load-bearing thing is the allowlist that now lets those keys through it. Declined non-goal: the rest of the sentence makes the intent recoverable, and this is preferential.

## Not fully validated

- The `virtualFocus` claim in the description, that `onKeyDownCapture` is the only mechanism available there because no focus event fires. It is plausible by inspection and nothing in the diff depends on it, but there is no test for it in this PR and I did not build a fixture to confirm it.
- Real-browser behavior beyond what CI covers, since I relied on the green Playwright jobs rather than rerunning them locally.

No follow-up issues were registered, because neither finding is separable from this PR's diff.

.changeset/300-modified-navigation-key-focus-visible.md
diegohaz: Fixed in [`f3c749a`](https://github.com/ariakit/ariakit/commit/f3c749aba05ad645f5eb62359d21b770752c2533): rewrote the entry as one past-tense summary sentence and moved the mechanism and scope rationale into the pull request description.

One deliberate deviation from your suggested text: it says `on modified navigation keys` rather than `when focus moved with a modified navigation key`. The fix also covers the case where the key moves nothing, such as `Alt+ArrowDown` on the last item of a composite that doesn't loop, where no focus event fires and only `onKeyDownCapture` can apply the attribute, so the wording avoids implying that focus has to move.

I kept the one-sentence shape rather than the title-plus-paragraph form. Naming `Focusable` first does carry the scope point, as you noted.

packages/ariakit-react-components/src/focusable/focusable.tsx
diegohaz: Fixed in [`d75a318`](https://github.com/ariakit/ariakit/commit/d75a3181434ea261f24b91cdb58daf997a3b0c5d): extracted the predicate into `packages/ariakit-react-components/src/focusable/__utils.ts`, imported by both `focusable.tsx` and `combobox-select.tsx`. There is now a single definition, so the drift hazard on `shouldProxyCompositeEvent` is gone rather than merely discoverable.

The private-module argument is what changed my mind, and I verified each premise before adopting it: `isPrivateModule` at `packages/ariakit-scripts/src/build.ts:56` skips `__`-prefixed entries at line 135 while collecting public files; `__utils` appears zero times in `packages/ariakit-react-components/package.json` even though `checkbox/__utils.ts`, `combobox/__utils.ts` and `popover/__utils.ts` all exist; and `combobox-select.tsx` already imported `../button/utils.ts` and `../popover/__utils.ts`. After the change, `pnpm build` followed by `npm pack --dry-run --json` shows the new file shipping exactly like those three, and the generated `exports` maps are unchanged.

I put it in `focusable/` rather than `composite/` so both call sites import downward, since `Focusable` is the lowest layer of focusable, composite, combobox. `focusable/` has no sibling-directory imports today, so it cannot pick up a dependency that would flow back into `Button` consumers.

The pull request description has been updated: both reassessment entries are kept as point-in-time records, with a `## Review response` section recording that the extraction was adopted and why the earlier objections did not apply to this third option.

diegohaz:
**Fixed**

- [`f3c749a`](https://github.com/ariakit/ariakit/commit/f3c749aba05ad645f5eb62359d21b770752c2533) ([reply](https://github.com/ariakit/ariakit/pull/7101#discussion_r3742332513)): rewrote the changeset as one past-tense summary sentence and moved the mechanism and scope rationale into the description. Kept the one-sentence shape, and used `on modified navigation keys` so it also covers the case where the key moves nothing.
- [`d75a318`](https://github.com/ariakit/ariakit/commit/d75a3181434ea261f24b91cdb58daf997a3b0c5d) ([reply](https://github.com/ariakit/ariakit/pull/7101#discussion_r3742332736)): extracted `isCompositeMoveKey` into `focusable/__utils.ts`, imported by both `focusable.tsx` and `combobox-select.tsx`, leaving a single definition. The `__`-prefixed private module adds no published entry point and no dependency, which is what the two earlier objections had missed.

The description now records the adopted extraction in a `## Review response` section, keeping both reassessment entries as point-in-time records.

Verification on the pushed head: full Vitest 257 files and 1726 tests, `pnpm test-chrome` on `composite-navigation`, `focusable-click-focus` and `combobox-select` 169 passed, `pnpm lint-fix` and `pnpm tsc` clean, and `pnpm build` plus `npm pack --dry-run` confirming the generated `exports` maps are unchanged.

packages/ariakit-react-components/src/focusable/focusable.tsx
```diff
@@ -190,13 +191,21 @@ function onGlobalMouseDown(event: MouseEvent) {
 }
 
 function onGlobalKeyDown(event: KeyboardEvent) {
+  // Composite items navigate on these keys without checking modifiers, and
+  // grids treat Ctrl+Home and Ctrl+End as first and last item, so the modified
+  // gesture moves focus just like the unmodified one.
+  // https://github.com/ariakit/ariakit/issues/7099
+  if (isCompositeMoveKey(event.key)) {
+    isKeyboardModality = true;
+    return;
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR, because the scenario is pre-existing behavior for the unmodified gesture rather than something this change introduces, and the suggested remedy cannot be implemented where the finding places it.

`onGlobalKeyDown` has never had a key filter for unmodified keystrokes. With no modifier held, `metaKey`, `ctrlKey` and `altKey` are all false, so control falls through to `isKeyboardModality = true` for *any* key. On `main` today, a user who clicks a text field and presses plain `ArrowLeft`, or simply types a letter, already flips the process-wide flag, and a button focused programmatically afterwards already gets `data-focus-visible`. This change makes `Option+ArrowLeft` behave like `ArrowLeft` in exactly that respect; it does not create a new class of artifact. The flag records that the user is interacting with the keyboard, not that a composite consumed the key.

The text field itself is unaffected either way, since `isAlwaysFocusVisible` already returns true for `TEXTAREA`, `SELECT`, text-type `INPUT` and contenteditable, so those are focus-visible on pointer focus regardless of modality.

On the remedy: `onGlobalKeyDown` is registered on `document` in the capture phase, so it runs strictly before any composite handler could consume the key. It cannot observe whether a navigation action was performed. Deferring the classification until after a composite acts would also leave the second guard unfixed, since `onKeyDownCapture` belongs to every `Focusable` and applies on plain `Button` elements where nothing consumes the key, and it would reintroduce the disagreement between the two surfaces that #7099 asks to resolve.

The broader consequence is deliberate and disclosed in the description: a mouse-focused plain `Button` now becomes focus-visible on a modified navigation key, which is parity with the unmodified keys that guard has always allowed.

diegohaz:
**Declined**

- `chatgpt-codex-connector` on `focusable.tsx` ([reply](https://github.com/ariakit/ariakit/pull/7101#discussion_r3742361494)): scope modified-key modality to actual composite navigation. Declined. `onGlobalKeyDown` has never filtered unmodified keystrokes, so on `main` today a plain `ArrowLeft` caret move, or typing any character, already sets the process-wide flag; this change only brings the modified gesture to parity. The suggested remedy is also not implementable there, since the listener is a document capture-phase handler that runs before any composite could consume the key, and it would leave `onKeyDownCapture` unfixed on plain `Focusable` elements.

No code change. No new commits since [`d75a318`](https://github.com/ariakit/ariakit/commit/d75a3181434ea261f24b91cdb58daf997a3b0c5d), which is green on all 48 checks.

diegohaz (commented):
Re-review. Incremental baseline is [`3c714111`](https://github.com/ariakit/ariakit/commit/3c71411155bdeebf9c49e78aa5ac44d0359c9e71), the head recorded in the previous marked review, to the current head [`d75a3181`](https://github.com/ariakit/ariakit/commit/d75a3181434ea261f24b91cdb58daf997a3b0c5d). Two new commits: [`f3c749ab`](https://github.com/ariakit/ariakit/commit/f3c749aba05ad645f5eb62359d21b770752c2533) and [`d75a3181`](https://github.com/ariakit/ariakit/commit/d75a3181434ea261f24b91cdb58daf997a3b0c5d). Both findings from the previous review are fixed, and I found no new substantive findings.

## Both prior findings verified fixed

**Duplicated predicate.** `isCompositeMoveKey` now lives in `packages/ariakit-react-components/src/focusable/__utils.ts` and is imported by `focusable.tsx:33` and `combobox-select.tsx:28`. The moved body is byte-identical to the two copies it replaces, comment included, so the extraction is behavior-preserving. `rg isCompositeMoveKey` finds one definition and two call sites, and searching the wider key list turns up no other copy in `packages/`. No packaging cost, as expected: `__utils` appears in neither `packages/ariakit-react-components/src/index.ts` nor the `exports` map in its `package.json`, and the PR adds no `package.json` change. The "keep in sync" note now has a single home that both call sites resolve to.

**Changeset size.** Now one past-tense summary sentence, down from three. By the same `wc -w` count I used before, the body went from 81 to 25. Links re-checked on this head: the three reference pages return `200` and `https://ariakit.com/guide/styling` still contains `id="data-focus-visible"`. Naming `Focusable` first, with `CompositeItem` and `ToolbarItem` as the two representative derived components, keeps the "not only composite items" point implicit, which was the intent of dropping the third sentence.

## Validation performed on this head

- Ran `pnpm test combobox-select app/src/sandbox/composite-navigation/test.ts app/src/sandbox/focusable-click-focus/test.ts`: 149 passed across 30 files. `combobox-select.tsx` is the file that lost its local copy, so its whole sandbox family was included rather than just the two files this PR touches.
- Re-ran the red check at this head, since the extraction moved code inside `focusable.tsx`. Restoring only `focusable.tsx` to its merge-base content, leaving the new `__utils.ts` in place so `combobox-select.tsx` still resolves, fails the same 4 tests, each on the `data-focus-visible` assertion, with `keeps pointer modality on a modified non-navigation key` still passing. The worktree was restored afterwards and `git status` is clean.
- Confirmed the two guards in `focusable.tsx:192-200` and `focusable.tsx:354-361` are unchanged apart from the import, so no behavior moved with the predicate.
- All 48 non-skipped checks are green on this head, with only `Build styles` skipped. That includes `Main / Build packages`, which runs the export generation the private-module claim depends on, plus every Playwright project.

## Resolved threads, recorded rather than reopened

All three threads on this PR are resolved, so none were candidates. Recording the outcomes since two of them are mine:

- The changeset thread and the `isCompositeMoveKey` thread each carry a "Fixed in" reply, and both fixes are verified above.
- [discussion_r3742338825](https://github.com/ariakit/ariakit/pull/7101#discussion_r3742338825) from the Codex connector, "Scope modified-key modality to actual composite navigation", was resolved with a reply that matches the conclusion my previous review reached independently. I checked its two load-bearing claims rather than taking them on trust: `isAlwaysFocusVisible` at `focusable.tsx:64-78` does return true for non-readonly `TEXTAREA`, `SELECT`, and text-type `INPUT`, for `isContentEditable`, and for Ariakit's own combobox role, so the text field in that scenario is focus-visible on pointer focus regardless of modality; and `focusable.tsx:247` registers the listener with `capture` true, so it does run before any composite handler and genuinely cannot observe whether a navigation action was consumed. The dismissal holds.

## New findings

None. The incremental diff is a behavior-preserving extraction plus a changelog rewrite, and neither exposes a broader issue in the surrounding code. Two things I considered and did not raise: the shared comment's middle sentence states the `Focusable` modality rationale while `combobox-select.tsx` uses the predicate for event proxying, which is accurate for both consumers since the modifier-agnostic key set is what each one needs; and the plain `//` comment on an exported helper matches the mixed style already present in `checkbox/__utils.ts`, `combobox/__utils.ts`, and `popover/__utils.ts`.

## Not fully validated

Unchanged from the previous review: the `virtualFocus` claim in the description is plausible by inspection but has no test here, and real-browser behavior rests on the green Playwright jobs rather than a local run.

PR #7107, opened 2026-08-09
Wait for preview hydration in browser tests

diegohaz:
## Motivation

Firefox browser tests could sample or interact with server-rendered preview controls before Astro completed its client hydration handoff, producing intermittent render-count and cross-realm focus failures.

## Solution

Keep generic browser navigation limited to bounded load and network-idle settling. For `withFramework` preview routes, wait for Astro to hand every eager `client:load` island to its renderer when JavaScript is enabled:

```ts
await gotoAndSettle(page, previewUrl);
if (javaScriptEnabled) {
  await page.waitForFunction(
    () => !document.querySelector("astro-island[ssr]"),
  );
}
```

Gate shadow-root and iframe fixture disclosures on their actual focus targets, and leave performance page-load measurements unchanged.

<details>
<summary>Cycle 3: Scope hydration waiting to preview routes</summary>

**Assessment**

The original failure is a recurring CI race on eager preview routes. Applying the wait to generic navigation introduced exceptions for full-site pages with offscreen `client:visible` islands, including the reference and checkbox-card pages, while the homepage renders no islands at all.

**Decision**

Scope the hydration handoff wait to `withFramework`, where generated preview routes contain only eager islands, and remove the navigation option plus all three full-site opt-outs. Keep observable readiness assertions for state that React creates after Astro's handoff.

</details>


app/src/tests/preview-hydration-browser.ts
```diff
@@ -0,0 +1,33 @@
+import { test } from "#app/test-utils/fixtures.ts";
```
diegohaz: We don't need this file. We don't need to test these test utils.

app/src/test-utils/preview.ts
```diff
@@ -11,25 +11,41 @@ import {
...
   await page
     .waitForLoadState("networkidle", { timeout: 5_000 })
     .catch((error) => {
       if (!(error instanceof errors.TimeoutError)) throw error;
     });
+  if (options.waitForHydration) {
+    await page.waitForFunction(
+      () => !document.querySelector("astro-island[ssr]"),
+    );
+  }
```
diegohaz: Doesn't this new `waitForHydration` replace the `.waitForLoadState("networkidle", { timeout: 5_000 })`? The `networkidle` was there mostly for waiting for hydration.

app/src/sandbox/store-keyed-subscriptions/test-chrome.ts
```diff
@@ -11,7 +11,7 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page.on("pageerror", (error) => hydrationErrors.push(error.message));
 
     // Re-navigate with listeners attached to capture errors from hydration.
-    await gotoAndSettle(page, page.url());
+    await gotoAndSettle(page, page.url(), { waitForHydration: true });
```
diegohaz: I believe `waitForHydration` should be `true` by default in `gotoAndSettle`.

app/src/tests/preview-hydration-browser.ts
diegohaz: Confirmed against the current head: the file is still there, and I agree it should go. Two things I found while validating it that make the case stronger than "we don't test test utils".

The final assertion cannot fail. `gotoAndSettle(page, url, { waitForHydration: true })` only returns once `document.querySelector("astro-island[ssr]")` is null, so `test.expect(await page.locator("astro-island[ssr]").count()).toBe(0)` just restates the postcondition of the call above it. Everything the test proves is already encoded in "the previous line did not time out".

The part that does carry signal, the init script, is coupled to Astro's private client bootstrap. `astro-island`'s `start()` reads `Astro[this.getAttribute("client")]` and, when it is `undefined`, waits for an `astro:<directive>` event (see `astro-island.prebuilt-dev.js` in `astro@7.1.6`). The init script here rides on that exact shape by installing a setter on `self.Astro.load`. If Astro changes how the directive globals are published, the delay silently stops being installed, the page hydrates at normal speed, and the test still passes green while covering nothing. A test that cannot fail for the reason it exists is worse than no test.

There is also a small cost: the filename matches `/\/tests\/[^/]*-browser/` in `app/playwright.config.ts`, so it runs in the `chrome`, `firefox`, and `safari` projects, each paying the scripted 2s delay.

Disposition: address in this PR by deleting `app/src/tests/preview-hydration-browser.ts`. If you want to keep a red/green record of the wait, the PR description is the right place for it, not a permanent browser test.

app/src/test-utils/preview.ts
diegohaz: It replaces `networkidle` as a hydration *signal*, but I would not drop the `networkidle` wait yet. I checked both halves of this.

First, `networkidle` is still the only settle step on every `waitForHydration: false` path: `theme-browser.ts:5`, `theme-browser.ts:28`, `reference-hovercards-browser.ts:38`, and the two `withFrameworkBeforeHydration` suites. So it cannot be removed outright, only made conditional.

Second, and more important: on the `waitForHydration: true` path the island wait resolves earlier than the name suggests, so dropping `networkidle` would widen the race this PR is closing rather than narrow it. In `astro@7.1.6`, `astro-island` does `await u(this.Component, ...)` and then `this.removeAttribute("ssr")`, where `u` is the renderer's hydrate callback. `@astrojs/react@6.0.2`'s `client.js` returns immediately after `startTransition(() => { hydrateRoot(element, componentEl, renderOptions) })`, and React 19's `hydrateRoot` ends at `ensureRootIsScheduled(...)` (`react-dom-client.development.js:28111`) without flushing. Inside `startTransition` the update takes a transition lane, so the hydration render commits in a later Scheduler task, after the microtask in which Astro removed `ssr`.

So for React previews, `astro-island[ssr]` disappearing means "the island module loaded and the renderer was invoked", not "the DOM is hydrated". The bounded `networkidle` wait is currently the slack that usually covers that remaining gap, which is also why this PR still needs the `disabled` guards in the combobox fixture and the `visibilitychange` retry loop in `form-6309`.

Solid is the opposite case: `@astrojs/solid-js` calls `hydrate(fn, element, { renderId })` synchronously, so Solid previews genuinely are hydrated when the wait resolves.

My recommendation: keep both waits for now and make the JSDoc say what the island wait actually guarantees (I left a separate comment on that). If you want to drop `networkidle` for previews later, the prerequisite is a signal that means "the framework committed", for example a marker set from an effect in the generated preview wrapper, which would also let the fixture guards go away.

app/src/sandbox/store-keyed-subscriptions/test-chrome.ts
diegohaz: Confirmed against the current head (`gotoAndSettle` still defaults to `{}`, so the wait is off unless a caller opts in), and I agree.

The strongest argument is that all three opt-in call sites sit *inside* a `withFramework` block whose `beforeEach` already navigated with the wait enabled: `store-keyed-subscriptions/test-chrome.ts:14`, `menu-3214/test-chrome.ts:27`, and `form-6309/test-browser.ts:20`. Having to restate the option on a re-navigation to the same URL is exactly the kind of thing that gets dropped when a test is copied, and dropping it fails silently and intermittently, which is the failure mode this PR exists to remove.

Flipping the default costs three opt-outs, all on full site pages, and they are all real: `theme-browser.ts:5`, `theme-browser.ts:28`, and `reference-hovercards-browser.ts:38`. `/react/examples/checkbox-card/` renders code blocks through `code-block.astro:18` (`client:visible`) and `code-block-tabs.astro:114` (`client:visible={{ rootMargin: "50%" }}`), and the reference page's hovercard anchors are `client:idle`, which the test already notes at `reference-hovercards-browser.ts:43`. An island that never becomes visible keeps its `ssr` attribute forever, so those three would hang without the opt-out.

That trade is worth taking, because the two failure modes are not symmetric: forgetting an opt-out fails loudly with a `waitForFunction` timeout on the first run, while forgetting an opt-in fails intermittently and only on a contended runner.

```ts
export async function gotoAndSettle(
  page: Page,
  url: string,
  { waitForHydration = true }: PreviewNavigationOptions = {},
) {
```

While you are there, note that `withFramework`'s `options: PreviewNavigationOptions = { waitForHydration: true }` is a whole-object default, so `withFramework(dir, cb, {})` silently turns the wait off. Destructuring the property default in `gotoAndSettle` fixes that too, because `withFramework` can then forward `options` unchanged.

Preview pages are safe for the flipped default: `preview-codegen.ts:118` only ever emits `client:load` islands, and `[...preview].astro` renders the island alone with no site chrome, so there is no lazy island on a preview route.

app/src/test-utils/preview.ts
```diff
@@ -11,25 +11,41 @@ import {
...
  * Navigates to `url` and waits for it to be ready for a browser test. Waits for
  * the bounded `load` event rather than `networkidle`: under CI contention,
  * networkidle's unbounded "no requests for 500ms" wait can stall past the test
  * timeout. After `load`, still wait for network idle so late work (such as
  * `client:load` island hydration) settles, but cap it so a stalled or chatty
- * request degrades to a short wait instead of consuming the test budget. Only
+ * request degrades to a short wait instead of consuming the test budget. Then
+ * wait for Astro to hand every island to its renderer so tests do not interact
+ * with server-rendered controls when `waitForHydration` is enabled. Keep that
+ * disabled for full site pages, which can contain lazy offscreen islands. Only
```
diegohaz: This promises more than the wait delivers for React previews, and the difference is exactly why the rest of the PR still needs its guards.

In `astro@7.1.6`, `astro-island` does `await u(this.Component, m, n, { client })` and only then `this.removeAttribute("ssr")`, where `u` is the value returned by the renderer's client entry. `@astrojs/react@6.0.2`'s `client.js` returns as soon as it has called `startTransition(() => { hydrateRoot(element, componentEl, renderOptions) })`, and React 19's `hydrateRoot` finishes at `ensureRootIsScheduled(...)` (`react-dom-client.development.js:28111`) without flushing anything. Because the update is enqueued inside `startTransition`, it lands on a transition lane and the hydration render is performed by a Scheduler task, which runs after the microtask in which Astro dropped the `ssr` attribute.

So on a React preview this wait means "the island module loaded and the renderer was invoked". It does not mean the DOM stopped being server-rendered markup, and a test that interacts on the very next Playwright command is still, in principle, racing. What it does remove is the large, variable part of the gap (module fetch and evaluate), which is why it fixes the observed flakiness. Solid is genuinely stronger: `@astrojs/solid-js`'s `client.js` calls `hydrate(fn, element, { renderId })` synchronously, so a Solid island really is hydrated when `ssr` disappears.

The concrete risk is a future contributor reading this paragraph, concluding the wait covers the whole handoff, and deleting the things that actually close the remaining gap: `disabled={!portalElement}` and `disabled={!frameBody}` in the combobox fixture, or the `visibilitychange` retry loop in `form-6309/test-browser.ts`. The Firefox flake then comes back, and it comes back as an intermittent CI failure rather than as a review comment.

Suggested wording change: say the wait covers Astro handing each island to its renderer, and add one line noting that React commits the hydration render in a later task, so callers that touch state produced by effects still need their own readiness signal.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because the mistaken reading removes working race guards rather than breaking anything today.
- Priority: P1 (estimate, 80% confidence), because the comment is introduced by this PR and correcting it is a few words.
- Breakage likelihood: Possible (estimate, 75% confidence), because it needs a later contributor to act on the stronger reading; the underlying gap itself is verified from the Astro and React sources cited above.
- Fix effort: Easy (verified), because only the JSDoc changes.

app/src/test-utils/preview.ts
```diff
@@ -91,9 +108,17 @@ export function withFramework(
...
   }
 }
+
+/** Runs framework previews without waiting for client-side hydration. */
+export function withFrameworkBeforeHydration(
+  dirname: string,
+  callback: (params: WithFrameworkCallbackParams) => Promise<void>,
+) {
+  return withFramework(dirname, callback, { waitForHydration: false });
+}
```
diegohaz: Both call sites of this wrapper pair it with `test.use({ javaScriptEnabled: false })` (`composite-item-tabbable/test-chrome-js-disabled.ts:4` and `toolbar-pre-hydration/test-chrome-js-disabled.ts:4`), and that is not a coincidence: with JavaScript disabled the island can never hydrate, so the wait can never be satisfied. Playwright already exposes that as an option fixture, so the beforeEach can derive the answer instead of asking every author to remember a second, separately named entry point.

```ts
test.beforeEach(async ({ page, javaScriptEnabled }) => {
  await gotoAndSettle(page, `/${framework}/previews/${id}/`, {
    waitForHydration: javaScriptEnabled,
    ...options,
  });
});
```

That lets `withFrameworkBeforeHydration` and the two files importing it go back to plain `withFramework`, and it closes a real footgun: a future JS-disabled preview test written with `withFramework` today gets no error at authoring time and no useful error at run time. I verified both halves locally in a scratch Playwright project against `@playwright/test@1.62.1`:

- A `beforeEach` registered *before* the callback that calls `test.use({ javaScriptEnabled: false })`, which is exactly the order `withFramework` uses, still observes `javaScriptEnabled === false`. A sibling describe without the override observes `true`.
- With `javaScriptEnabled: false`, `page.waitForFunction(() => !document.querySelector("astro-island[ssr]"))` never resolves and ends in `page.waitForFunction: Timeout ... exceeded`. In `app/playwright.config.ts` the non-perf projects set no `timeout`, so that is Playwright's 30s default, which is also the whole test budget: the failure surfaces as a timeout inside `beforeEach` with no hint about JavaScript being off.

If you would rather keep an explicit escape hatch for a pre-hydration test that runs *with* JavaScript enabled, `...options` above preserves it while still defaulting correctly, and the named wrapper can go.

**Assessment**

- Severity: Medium (verified), because the failure consumes the full 30s test budget and names `waitForFunction` rather than the actual cause.
- Priority: P2 (estimate, 75% confidence), because nothing in the current tree hits it; it is a trap for the next JS-disabled preview test.
- Breakage likelihood: Possible (estimate, 70% confidence), because it requires someone to add a JS-disabled preview suite and reach for the obvious helper.
- Fix effort: Easy (verified), because it is one fixture parameter plus deleting the wrapper and two imports.

packages/ariakit-scripts/src/perf.ts
```diff
@@ -1044,6 +1045,11 @@ async function gotoAndSettle(page: Page, url: string) {
     .catch((error) => {
       if (!(error instanceof errors.TimeoutError)) throw error;
     });
+  await page.waitForFunction(
+    () => !document.querySelector("astro-island[ssr]"),
+    undefined,
+    { timeout: ITERATION_NAVIGATION_TIMEOUT },
+  );
```
diegohaz: Two things worth recording here, because this copy is no longer the same function as the one it says it is kept in sync with.

First, the divergence itself. `app/src/test-utils/preview.ts` now gates the wait behind `waitForHydration`, while this copy always waits. That is fine in practice, and I checked why: every navigation that reaches this function is a preview URL (`measureIteration` uses `params.url`, and `createPerfPageLoadMeasure` uses `page.url()` from a `withFramework` beforeEach), `preview-codegen.ts:118` only emits `client:load` islands, and `[...preview].astro` renders the island with no site chrome. So there is no lazy offscreen island on a perf route and the wait cannot stall. Worth saying that out loud in the comment, since the sync note now invites a reader to "fix" the difference by copying the option across.

Second, and the reason I am flagging it rather than just noting it: `createPerfPageLoadMeasure` passes `gotoAndSettle` itself as the measured interaction, so adding a step inside it changes what the `page load` measurement covers. `app/src/sandbox/ariakit-tailwind/perf-chrome.ts:5` is the one test using it. `.github/workflows/perf.yml` builds the baseline from a worktree at `github.event.pull_request.base.sha` (see the "Set up baseline worktree" step, `git worktree add --detach "$BASELINE_DIR" "$BASE_SHA"`), so the baseline side of this PR's pair runs the old function and the current side runs the new one. Per `ariakit-ariakit-performance`, that makes any movement in that row a benchmark-definition change, not a product signal.

This PR's perf run came back clean, so nothing was misattributed, but the `ariakit-tailwind` `page load` row in it is not an apples-to-apples comparison and the first post-merge run is the first valid baseline for it. Calling that out in the PR description (or landing the `perf.ts` hunk separately) keeps the record straight.

For what it is worth, the direction is right: `ariakit-ariakit-performance` asks the measured callback not to resolve before the UI work completes, and ending page-load measurement at "islands handed to the renderer" is closer to that than ending it at network idle.

**Assessment**

- Severity: Low (estimate, 85% confidence), because it affects how one perf row is interpreted, not correctness of the product or the suite.
- Priority: P2 (estimate, 80% confidence), because the run is already clean and the fix is a note.
- Breakage likelihood: Likely (verified), because the mismatched pair is inherent to how the baseline worktree is built and already happened in this PR's run.
- Fix effort: Easy (estimate, 90% confidence), because it is a comment plus a line in the PR description, unless you prefer to split the hunk into its own PR.

app/src/sandbox/combobox-select-selected-value-presentation/index.react.tsx
```diff
@@ -611,16 +611,16 @@ function NativeAutoFocusFixture() {
...
-      focusTargetRef.current = null;
+      setFocusTarget(null);
       input.removeEventListener("focus", onFocus);
       nestedHost.remove();
     };
   }, []);
   return (
     <>
-      <Ariakit.DialogDisclosure store={dialog}>
+      <Ariakit.DialogDisclosure store={dialog} disabled={!focusTarget}>
```
diegohaz: This fixture change and the assertion added at `test-browser.ts:517` guard the same race twice, and unlike the shadow-root and iframe fixtures this one already had an observable readiness signal.

`setFocusTarget(input)` runs in the same synchronous effect body as `root.append(nestedHost)`, which is what makes `q.textbox("Native auto-focus shadow target")` resolvable (Playwright pierces open shadow roots, and the test already relies on that for `toBeFocused` further down). So `await test.expect(escapeTarget).toBeVisible()` alone proves the effect ran, and the value is only read later, inside the dialog input's `onFocus`, a full click round trip after that assertion.

Keeping both means the fixture's rendered contract changed (the disclosure is now `disabled` in the server-rendered markup) to close a window that the test already closes, and it converts a `useRef` into state that exists purely to drive that attribute. I would keep the `toBeVisible` wait and revert the fixture here, or keep the fixture change and drop the now-redundant `toBeEnabled` assertion, but not both.

The `disabled={!portalElement}` and `disabled={!frameBody}` guards in the other two fixtures are a different case and I am not suggesting changing those: neither has an accessible signal available before the dialog opens, and exposing readiness as real disabled state is what `ariakit-ariakit-example-integration-test` asks for over an internal probe.

**Assessment**

- Severity: Low (estimate, 85% confidence), because both guards are correct; the cost is redundant code and a changed fixture contract.
- Priority: P2 (estimate, 80% confidence), because it is a cleanup rather than a defect.
- Breakage likelihood: Rare (estimate, 85% confidence), because nothing currently asserts this fixture's pre-hydration markup.
- Fix effort: Easy (verified), because it is reverting one of the two halves.

diegohaz (commented):
Initial review of all 10 changed files at [`7b2fff5`](https://github.com/ariakit/ariakit/commit/7b2fff5233ee56d5419f8e69bb18a54ed8c86e18), diffed against the merge base [`6a9e599`](https://github.com/ariakit/ariakit/commit/6a9e5995a599e52100c35a8fc40b0072c8d0672c), plus the surrounding preview, perf, and Astro island machinery. Reviewed in a dedicated worktree at the PR head. All checks on this head are green.

The direction is right and the diagnosis holds: waiting for the island handoff removes the large, variable part of the pre-hydration window, and the two pre-hydration suites are correctly moved onto an explicit opt-out path rather than silently losing the wait. My findings are about how much the wait is claimed to guarantee, and about two knobs that could be one.

## Validation performed

- Read `astro@7.1.6`'s `astro-island.prebuilt-dev.js`, `@astrojs/react@6.0.2`'s `client.js`, `@astrojs/solid-js@7.0.1`'s `client.js`, and React 19.2.8's `hydrateRoot` to establish exactly when `astro-island[ssr]` is removed relative to the framework's hydration commit.
- Traced every `gotoAndSettle` caller and every `withFramework` navigation target, and confirmed preview routes only ever carry `client:load` islands (`preview-codegen.ts:118`, `[...preview].astro`) while the three non-preview callers land on pages with `client:idle` and `client:visible` islands.
- Ran a scratch Playwright project against `@playwright/test@1.62.1` to check two claims behind the `withFrameworkBeforeHydration` finding: that a `beforeEach` registered before `test.use({ javaScriptEnabled: false })` still observes the override (it does), and what `page.waitForFunction` does with JavaScript disabled (it never resolves and times out).
- Read `.github/workflows/perf.yml` to confirm the perf baseline is built from a worktree at the PR base SHA, which is what makes the `page load` row a benchmark-definition change here.

- `app/src/tests/preview-hydration-browser.ts` ("we don't need this file"): **confirmed**, replied on the thread. Address now. Beyond the scope argument, the file's final assertion cannot fail (it restates `gotoAndSettle`'s postcondition) and its only real coverage depends on Astro's private directive-global shape, so it can silently degrade to a no-op and still pass.
- `app/src/test-utils/preview.ts:48` ("doesn't this replace `networkidle`?"): **confirmed as a real question, but the answer is no for now**, replied on the thread. It replaces `networkidle` as a hydration signal, but `networkidle` is still the only settle step on the opt-out paths, and because the island wait resolves before React commits, it is currently also the slack covering that gap. Removing it would widen the race this PR narrows.
- `app/src/sandbox/store-keyed-subscriptions/test-chrome.ts:14` ("`waitForHydration` should default to `true`"): **confirmed**, replied on the thread. Address now. All three opt-in sites re-navigate inside a `withFramework` block that already waited, and the two failure modes are asymmetric: a forgotten opt-out fails loudly, a forgotten opt-in fails intermittently. The three opt-outs it would require are all real and all on full site pages.

## Newly published findings

- `app/src/test-utils/preview.ts:24-27`: the JSDoc claims tests will not interact with server-rendered controls, but on React previews the wait resolves before the hydration render commits. P1, easy, wording only.
- `app/src/test-utils/preview.ts:118-124`: `withFrameworkBeforeHydration` can be derived from Playwright's `javaScriptEnabled` option, which also removes a 30s silent-timeout trap for the next JS-disabled preview test. P2.
- `packages/ariakit-scripts/src/perf.ts:1048-1052`: this copy now diverges from the one it is documented to be in sync with, and because `createPerfPageLoadMeasure` measures `gotoAndSettle` itself, the `ariakit-tailwind` `page load` row in this PR's perf report compares two different measured intervals. P2.
- `app/src/sandbox/combobox-select-selected-value-presentation/index.react.tsx:623`: the `useRef` to state conversion plus `disabled` guards the same race as the new `toBeVisible` assertion in the test. Keep one. The equivalent guards in the shadow-root and iframe fixtures are justified and I am not asking for those to change. P2.

## Other candidates and dispositions

- CodeRabbit's outside-diff note on `perf.ts` (gate the wait behind an option because lazy offscreen islands could exhaust `ITERATION_NAVIGATION_TIMEOUT`): **invalid as stated**. Perf navigations only ever reach preview routes, which carry no lazy islands. The sync-divergence half of the concern is real and is folded into my `perf.ts` comment.
- The changeset bot warning: **not applicable**. Both `@ariakit/scripts` and `app` are `private: true`, and this is test infrastructure with no shipped-code effect.
- A stronger readiness signal, for example a marker set from an effect in the generated preview wrapper, would let the fixture `disabled` guards, the `form-6309` retry loop, and eventually the `networkidle` wait all go away. I raised it in the `networkidle` thread as the prerequisite for that cleanup rather than as work for this PR, and I did not open a follow-up issue because it is a design option rather than confirmed, independently scoped work.
- The comment at `menu-3214/test-chrome.ts:30` ("this also confirms the island hydrated") is now partly redundant with the navigation wait, but it still describes what the assertion proves about item registration, so I did not raise it.

## Not fully validated

I did not run the app Playwright suites against these changes. The review worktree has no installed dependencies, and a faithful local run needs a full install plus `pnpm dev-app` with matching `APP_PORT` and `NEXTJS_PORT`. I relied on this PR's CI instead, which is fully green at this head, including Chrome, Firefox, Safari, iOS, Android, Visual, and the perf matrix. I also did not reproduce the original Firefox flakiness, so the claim that these specific waits are what fixed it rests on the mechanism plus the green run, not on a red-to-green demonstration.

app/src/tests/preview-hydration-browser.ts
diegohaz: Fixed in [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7): removed the brittle test-utils browser test.

app/src/test-utils/preview.ts
diegohaz: This is not an issue because the bounded network-idle step remains the only settle signal when hydration waiting is disabled, and Astro removing the `ssr` marker only records renderer handoff; React can commit in a later task. I kept both waits and clarified the distinction in [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7).

app/src/sandbox/store-keyed-subscriptions/test-chrome.ts
diegohaz: Fixed in [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7): `waitForHydration` now defaults to `true`, with explicit opt-outs for the three lazy full-site pages.

app/src/test-utils/preview.ts
diegohaz: Fixed in [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7): the JSDoc now describes Astro renderer handoff and calls out React's later commit and the need for effect-specific readiness signals.

app/src/test-utils/preview.ts
diegohaz: Fixed in [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7): `withFramework` now derives hydration waiting directly from Playwright's `javaScriptEnabled` fixture, and the special wrapper was removed.

packages/ariakit-scripts/src/perf.ts
diegohaz: Superseded by [`ae3ab57`](https://github.com/ariakit/ariakit/commit/ae3ab578ebd87745ca1509a4f1319074a6211e74): the performance hydration wait was removed, so the measured page-load interval remains unchanged; the comments now document only the shared bounded navigation steps.

app/src/sandbox/combobox-select-selected-value-presentation/index.react.tsx
diegohaz: Fixed in [`384f587`](https://github.com/ariakit/ariakit/commit/384f587a0397dcfb84b61c29285dc345994ccdfb): restored the native focus target ref and retained the existing visible target assertion as the sole readiness signal.

diegohaz:
**Fixed**

- [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744869923)): Removed the brittle test-utils browser test.
- [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744870380)): Made hydration waiting the safe default with explicit lazy-page opt-outs.
- [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744870551)): Documented Astro renderer handoff and React's later commit accurately.
- [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744870842)): Derived no-JavaScript behavior from Playwright's fixture configuration.
- [`384f587`](https://github.com/ariakit/ariakit/commit/384f587a0397dcfb84b61c29285dc345994ccdfb) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744871382)): Kept the native shadow target's existing visible readiness signal and removed redundant fixture state.

**Obsolete**

- [`ae3ab57`](https://github.com/ariakit/ariakit/commit/ae3ab578ebd87745ca1509a4f1319074a6211e74) ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744871231)): Removed the performance hydration wait, preserving the existing page-load measurement definition.

**Invalid**

- `No change` ([reply](https://github.com/ariakit/ariakit/pull/7107#discussion_r3744870129)): Kept bounded network idle because it still settles late work on opt-out paths, while Astro's marker only guarantees renderer handoff.

diegohaz (commented):
Re-review at [`384f587`](https://github.com/ariakit/ariakit/commit/384f587a0397dcfb84b61c29285dc345994ccdfb). Baseline is the head recorded in my previous review marker, [`7b2fff5`](https://github.com/ariakit/ariakit/commit/7b2fff5233ee56d5419f8e69bb18a54ed8c86e18). The branch was force-pushed and rebased onto [`aa7a556`](https://github.com/ariakit/ariakit/commit/aa7a556986dc2a75c73a061815491a2f4a7f40c9), so I diffed the two heads directly for the three new commits and separately re-checked the evidence that the rebase could have invalidated.

Every finding from the previous review is addressed, and each fix is the right one rather than a workaround. I have no new findings, so there are no new inline comments. All seven threads were already resolved with fix replies, so I verified them against the code instead of re-opening them.

## Verification of each fix

- [`020077d`](https://github.com/ariakit/ariakit/commit/020077de5e010c9ca18d39946fa21b348fbe6cf7) deletes `app/src/tests/preview-hydration-browser.ts`. Confirmed gone, and nothing else referenced it.
- The JSDoc now says the wait covers Astro handing each island to its renderer and that React can commit in a later task, so callers depending on effect-created state still need their own readiness signal. That matches the mechanism, and the sync note is correctly narrowed to the bounded load and network-idle steps, which are now the only part the two helpers share.
- `waitForHydration` defaults to `true` through a destructured property default, which also removes the partial-object footgun: `gotoAndSettle(page, url, {})` now waits instead of silently skipping. The three former opt-ins are back to plain `gotoAndSettle(page, page.url())`, and the three full-site pages opt out explicitly. I re-walked every caller; all nine navigation sites land on the intended side.
- `withFramework` derives the wait from `javaScriptEnabled` and the extra parameter and the `withFrameworkBeforeHydration` wrapper are gone, with both JS-disabled suites back on plain `withFramework`. The two `test.use({ javaScriptEnabled: false })` calls are the first statement in their callbacks, so they still register synchronously inside the describe body before the hook runs. `App / Test Chrome` is the project that owns both `test-chrome-js-disabled.ts` files, and it passed on this head, which exercises this path directly.
- [`ae3ab57`](https://github.com/ariakit/ariakit/commit/ae3ab578ebd87745ca1509a4f1319074a6211e74) removes the hydration wait from the perf helper entirely, which is a better answer than the one I suggested. `packages/ariakit-scripts/src/perf.ts` is now behaviorally identical to its state at the base for both call sites, so the measured `page load` interval in `createPerfPageLoadMeasure` is unchanged and this PR's `ariakit-tailwind` row is a valid comparison again. `ITERATION_NAVIGATION_TIMEOUT` is still used by `context.setDefaultNavigationTimeout`, so nothing was left dangling.
- [`384f587`](https://github.com/ariakit/ariakit/commit/384f587a0397dcfb84b61c29285dc345994ccdfb) restores `focusTargetRef` and drops the redundant `toBeEnabled` assertion, keeping `await test.expect(escapeTarget).toBeVisible()` as the single readiness signal. That is the half I would have kept: it is set in the same synchronous effect body, and the value is only read a full click round trip later. The `disabled={!portalElement}` and `disabled={!frameBody}` guards in the shadow-root and iframe fixtures are untouched, which is correct since neither has an accessible signal before the dialog opens.

The PR description was updated to match, including dropping the earlier claim that performance tests wait for islands. Description and implementation now agree.

## Re-verified after the rebase

The rebase moved the base onto the Astro dependency update, taking `astro` from 7.1.6 to 7.2.0 and `@astrojs/solid-js` from 7.0.1 to 7.0.2. My previous review's mechanism argument, which the new JSDoc wording now encodes, was based on 7.1.6, so I re-checked it against the published 7.2.0 build rather than assuming it carried over.

`astro@7.2.0`'s `astro-island.prebuilt.js` still does `await this.hydrator(this)(this.Component, u, n, { client: ... }), this.removeAttribute("ssr"), this.dispatchEvent(new CustomEvent("astro:hydrate"))`, so the attribute still drops one microtask after the renderer callback returns. `@astrojs/react` is unchanged at 6.0.2 and still returns right after `startTransition(() => hydrateRoot(...))`, and `@astrojs/solid-js@7.0.2` still calls `hydrate(fn, element, { renderId })` synchronously. The wording holds on the new base.

The hydration `waitForFunction` still has no explicit timeout, so it inherits Playwright's 30s default, which for the non-perf projects is the entire test budget. A preview that never hydrates therefore fails as a `page.waitForFunction` timeout inside `beforeEach`. I considered suggesting a shorter bound for a clearer failure, but the failure is already loud and immediate rather than intermittent, and a shorter bound only changes how long a doomed test takes to report. Not worth the extra knob.

Perf iterations still navigate without a hydration wait, both for the unmeasured `measureIteration` setup navigation and for the measured page-load callback. That is exactly the behavior at the base, so nothing regressed here and it is out of scope for this PR. Recording it only as a pointer: if perf ever shows load-related noise, adding the wait to the unmeasured navigation while keeping it out of `createPerfPageLoadMeasure` is the shape that buys determinism without breaking comparability. I have not measured that, so I am not claiming it is currently a problem and did not open a follow-up issue.

## Not fully validated

Same limitation as last time: I did not run the app Playwright suites locally, since the review worktree has no installed dependencies and a faithful run needs a full install plus `pnpm dev-app`. I relied on this PR's CI on this head instead. I also still have not reproduced the original Firefox flake, so the fix claim rests on the mechanism plus a green run rather than a red-to-green demonstration.

## Status

All 47 checks are green on this head, including `App / Test Chrome`, `App / Test Firefox`, both Safari shards, iOS, Android, Visual, and the full performance matrix. Firefox and the Chrome JS-disabled suites are the two that would have caught a mistake in the default flip or the `javaScriptEnabled` derivation, and both passed.

Looks good to me.

diegohaz (commented):
Re-review at [`1c60b08`](https://github.com/ariakit/ariakit/commit/1c60b08ebb41872715b19ae18d5ffe3ba871904d). Baseline is the head recorded in my previous review marker, [`384f587`](https://github.com/ariakit/ariakit/commit/384f587a0397dcfb84b61c29285dc345994ccdfb). One new commit, a fast-forward, so I reviewed it directly plus the surrounding call sites it changes the guarantees for. All seven earlier threads remain resolved and no new threads were opened, so there was nothing to re-adjudicate.

No new findings and no new inline comments. This iteration is better than the one it replaces, and I would take it as final.

## What changed and why it holds

[`1c60b08`](https://github.com/ariakit/ariakit/commit/1c60b08ebb41872715b19ae18d5ffe3ba871904d) moves the hydration wait out of `gotoAndSettle` and into `withFramework`'s hook, deleting `PreviewNavigationOptions`, the navigation option, and all three full-site opt-outs.

That is the right place for it. The wait's precondition is "this route contains only eager islands", and that is a property of generated preview routes, not of navigation in general: `preview-codegen.ts` emits every island as `client:load` and `[...preview].astro` renders the island with no site chrome. Putting the wait behind an option meant restating that precondition at every call site and inverting it at three others, and a forgotten opt-out would have hung a full-site test for the whole 30s budget. Scoping it to `withFramework` makes the invariant unstateable elsewhere, which is stronger than defaulting it correctly.

It also finishes off the divergence I raised in the first review. `gotoAndSettle` and the perf helper in `packages/ariakit-scripts/src/perf.ts` are now behaviorally identical again, so the "kept in sync" note is fully true rather than true of a subset, and the measured page-load interval is still untouched.

The comment revert in `form-6309/test-browser.ts` is correct and I checked it rather than assuming: "`gotoAndSettle` doesn't strictly guarantee" hydration is now literally accurate, since the helper no longer waits for it at all.

The PR description was rewritten to match the new shape, and the reassessment section documenting why cycle 3 changed direction is a good record to leave behind.

## One candidate I considered and am not publishing

The wait is now inline in the hook and unreachable from anywhere else, so the three tests that re-navigate to their own preview route get bounded load and network idle only: `store-keyed-subscriptions/test-chrome.ts:14`, `form-6309/test-browser.ts:20`, and `menu-3214/test-chrome.ts:27`. The guarantee `withFramework` establishes is dropped by the re-navigation with no way to re-establish it, which is the same trade-off from the other direction that opened this discussion.

I checked all three and none of them is affected. Each follows its re-navigation with a retrying assertion that cannot pass before hydration commits: `not.toHaveText("0")` on the render counters, `toHaveCount(3)` on the registered items, and the `toPass()` loop that re-fires `visibilitychange` until the status leaves "Idle". That is also the correct pattern independent of any navigation helper, and it is unchanged from `main`.

The residual risk is narrow: a future preview test that re-navigates and then interacts immediately would race, because server-rendered markup is fully actionable to Playwright while React has not attached its listeners yet, so the click lands and does nothing. If that ever comes up, extracting the five-line wait into an exported `waitForPreviewHydration(page)` is the small fix.

Disposition: not addressed in this PR. Nothing is broken today, no current test needs the helper, and this axis has now been through three design cycles. Adding a fourth turn on it would be churn rather than improvement, which is the call `ariakit-general-review-reassessment` asks for here. I am recording it rather than opening a follow-up issue, because it is a latent sharp edge with no confirmed failure behind it.

## Status

All 48 checks are green on this head, including Chrome, Firefox, both Safari shards, iOS, Android, Visual, and the full performance matrix. I did not run the suites locally, for the same reason as the previous two rounds: the review worktree has no installed dependencies and a faithful run needs a full install plus `pnpm dev-app`. I also still have not reproduced the original Firefox flake, so the fix claim continues to rest on the mechanism plus green runs rather than a red-to-green demonstration.

Looks good to me.

Issue #6837, opened 2026-07-24
Enable ComboboxItem focusOnHover by default


diegohaz:
## Context

Discovered while reviewing #6832. ComboboxItem now defaults `focusOnHover` to true in ComboboxSelect and standalone ComboboxList compositions, matching Select behavior, but ordinary input-backed Combobox items still keep the longstanding false default.

## Work

Evaluate and implement `focusOnHover` as the default for standard Combobox items. Add browser-primary and integration coverage showing that hovering an item without an explicit prop makes it active while preserving keyboard and virtual-focus behavior. Document and add a changeset for the behavior change.

## Expected outcome

Combobox items follow a consistent hover-focus default across standard, select, and standalone-list compositions.

This is independent of #6832 because changing the longstanding standard Combobox default is not required for the new ComboboxSelect API to work correctly.

diegohaz:
Although this isn't a breaking change from the user's perspective, I think it's safer to delay it until v0.5.0.

georgekaran:
hey @diegohaz can I work on this?

diegohaz:
> hey @diegohaz can I work on this?

@georgekaran of course! 👍

georgekaran:
hey @diegohaz I've implemented here: https://github.com/georgekaran/ariakit/tree/feat/combobox-focus-on-hover

let me know what you think

georgekaran:
@diegohaz @ariakit-bot 
Changes are in [040b28f](https://github.com/georgekaran/ariakit/commit/040b28fd96586471c140766266f2f05cc3036642)

PR #7119, opened 2026-08-12
Withhold ComboboxItem hover activation while the combobox is collapsed

### LINKED ISSUE #7118: ComboboxItem focusOnHover can move focus while the combobox is collapsed
## Current behavior

An explicit [`focusOnHover`](https://ariakit.com/reference/combobox-item#focusonhover) value of `true`, or a callback that returns `true`, bypasses the combobox open-state check. In an `alwaysVisible` list, pointer hover can make an item active and move DOM focus to the combobox owner while the owner is collapsed. The owner stays collapsed with `aria-expanded="false"`.

This behavior exists on current `main` at [`a0426ed`](https://github.com/ariakit/ariakit/commit/a0426ed547d95b84c9d53033053e51baeaca4aaa). It was added as an explicit override in [`6f79732`](https://github.com/ariakit/ariakit/commit/6f7973240). It exists before the work in #6837.

The recording starts with focus on another combobox. Pointer hover on `Explicit hover second` then moves focus to the collapsed `Always-visible hover select` combobox.

https://github.com/user-attachments/assets/177d1d1a-0c78-42bf-aa9e-8b0f3863230e

## Steps to reproduce the bug

The repository has a [current-main reproduction](https://github.com/ariakit/ariakit/tree/a0426ed547d95b84c9d53033053e51baeaca4aaa/app/src/sandbox/combobox-select-open-active-item). This smaller example has the same behavior:

```tsx

  return (
    <>
      <button type="button">Other control</button>
      <Ariakit.ComboboxProvider virtualFocus={false}>
        <Ariakit.ComboboxSelect aria-label="Fruit" />
        <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
          <Ariakit.ComboboxItem focusOnHover value="Apple" />
          <Ariakit.ComboboxItem focusOnHover value="Banana" />
        </Ariakit.ComboboxList>
      </Ariakit.ComboboxProvider>
    </>
  );
}
```

1. Focus `Other control`.
2. Confirm that the `Fruit` combobox is collapsed.
3. Move the pointer over `Banana` without a click.
4. Observe that focus moves to the collapsed combobox and `Banana` becomes active.

The same result occurs with `focusOnHover={() => true}`.

## Expected behavior

`focusOnHover` must not move focus or change the active item while the combobox store is closed. The closed state must apply to the default value, an explicit boolean, and a callback in the same way. Hover can activate an item after the store opens.

## Workaround

Return the live open state from the callback:

```tsx
const combobox = Ariakit.useComboboxStore();

<Ariakit.ComboboxProvider store={combobox}>
  <Ariakit.ComboboxSelect aria-label="Fruit" />
  <Ariakit.ComboboxList alwaysVisible>
    <Ariakit.ComboboxItem
      focusOnHover={() => combobox.getState().open}
      value="Apple"
    />
  </Ariakit.ComboboxList>
</Ariakit.ComboboxProvider>;
```

Assumptions, limitations, and a variant that preserves authored hover logic: https://github.com/ariakit/ariakit/issues/7118#issuecomment-5261165785

## Possible solutions

Gate the final hover predicate with the live `open` state. Do not put this check only in the prop default:

```ts
const focusOnHoverProp = useBooleanEvent(focusOnHover ?? true);

props = useCompositeHover({
  store,
  ...props,
  focusOnHover(event) {
    if (!focusOnHoverProp(event)) return false;
    return store.getState().open;
  },
});
```

This matches the current `SelectItem` rule. Tests must cover explicit `true` and a callback on a closed, always-visible list.

> AI bot report: An AI coding agent prepared this report from source inspection, automated tests, and browser verification. It is not authoritative. A maintainer should verify the evidence and decide the final behavior.

diegohaz:
## Motivation

An explicit [`focusOnHover`](https://ariakit.com/reference/combobox-item#focusonhover) value on `ComboboxItem`, either a boolean `true` or a callback that returns `true`, bypasses the combobox open-state check. In an `alwaysVisible` list, pointer hover can activate an item and move DOM focus to the combobox owner while the owner is collapsed with `aria-expanded="false"`:

```tsx
<button type="button">Other control</button>
<Ariakit.ComboboxProvider virtualFocus={false}>
  <Ariakit.ComboboxSelect aria-label="Fruit" />
  <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
    <Ariakit.ComboboxItem focusOnHover value="Apple" />
    <Ariakit.ComboboxItem focusOnHover value="Banana" />
  </Ariakit.ComboboxList>
</Ariakit.ComboboxProvider>
```

With focus on `Other control`, merely moving the pointer over `Banana` steals focus from the button and activates the item, even though the `Fruit` combobox is collapsed. The override was introduced by 6f79732407cebf3ffda58d8c74be5fb14e03969e, which honors authored item options unconditionally instead of gating the final hover predicate on the live `open` state the way `SelectItem` does.

Fixes #7118

## Solution

`useComboboxItem` now gates the whole `focusOnHover` predicate on the live combobox `open` state, applying the same closed-state rule that `SelectItem` already enforces: the closed state applies to the default value, an explicit boolean, and a callback in the same way, and hover activates an item once the store opens.

```ts
focusOnHover(event) {
  if (!store.getState().open) return false;
  if (!focusOnHoverProp(event)) return false;
  return store.getState().open;
},
```

The `open` check runs before the authored predicate so callbacks with side effects, such as the `move()` pattern from the [select-grid example](https://ariakit.com/examples/select-grid), don't run at all while the list is collapsed, and it is re-read after the predicate runs so a callback that closes the list from inside the predicate can't activate an item with a stale result. The gate covers both hover directions: hovering an option no longer activates it, and moving the pointer off an option no longer clears the active item or pulls focus back into the widget while the list is closed. The `focusOnHover` reference docs now state both directions of the closed-state rule, and a changeset covers `@ariakit/react-components` and `@ariakit/react`.

The reproduction and its guard rails live in the `combobox-select-collapsed-focus` sandbox, which owns the collapsed-focus contract: failing tests for an authored boolean and an authored callback on a collapsed always-visible list (red in CI on the first commit), a red-checked test proving a side-effectful callback isn't invoked while collapsed, a hover-end test proving leaving an option in a collapsed list keeps the active item and focus, and green guards pinning that both authored forms still take hover once the list opens plus the default predicate on a collapsed select list and an open plain combobox. The sandbox workaround from the second commit is reverted in the final commit.

This change also removes the `ClosedAlwaysVisibleSelect` fixture and its "honors focusOnHover on a closed always-visible list" test from the `combobox-select-open-active-item` sandbox. That previously green test pinned the contract added by 6f79732407cebf3ffda58d8c74be5fb14e03969e (#6832), which this change reverses; its reproduction details and the honored-override intent now live in the `HoverSelect` fixture and the open-list guard tests.

The same predicate-before-open-check exposure exists in the soft-deprecated `SelectItem` and is registered as #7120; it is pre-existing on `main` and independent of this change.

## Preview

The recording below shows the fixed behavior in the `combobox-select-collapsed-focus` sandbox with the fix applied, with only the `Hover fruit` widget visible and helper styles injected for legibility: real DOM focus is outlined in crimson and the active item is painted gold. Hovering the options of the collapsed always-visible list changes nothing while focus sits on the outside button, hovering activates items once the select opens, and after closing the list by clicking Grape, moving the pointer away keeps Grape active and focused. The issue's own [recording](https://github.com/user-attachments/assets/177d1d1a-0c78-42bf-aa9e-8b0f3863230e) shows the pre-fix behavior, where the same hover steals focus from the button.

[Recording of the fixed hover behavior on a collapsed always-visible combobox list](https://github.com/user-attachments/assets/9e8b6c10-9dcf-4975-849a-f74a141c26e8)

## Workaround

Until the fix is released, gate the authored `focusOnHover` value on the live open state from userland. The second commit applies this in the sandbox and all repro tests pass with the library unchanged:

```tsx
// TODO: Remove this store hoist and the open-state gate in focusOnHover
// below once https://github.com/ariakit/ariakit/issues/7118 is fixed.
const combobox = Ariakit.useComboboxStore({ virtualFocus: false });

<Ariakit.ComboboxProvider store={combobox}>
  <Ariakit.ComboboxSelect aria-label="Fruit" />
  <Ariakit.ComboboxList alwaysVisible aria-label="Fruit options">
    <Ariakit.ComboboxItem
      value="Apple"
      focusOnHover={(event) => {
        if (!combobox.getState().open) return false;
        // Defer to whatever value you would otherwise author.
        return true;
      }}
    />
  </Ariakit.ComboboxList>
</Ariakit.ComboboxProvider>;
```

Assumptions and limitations:

- The snippet assumes `focusOnHover` is always authored on the item. If your code sometimes leaves the prop `undefined`, returning `false` from the wrapper in that case is narrower than the component default (`selectMode && open`), which keeps hover activation on while a select-shaped list is open. Only apply the gate where you author the value.
- The gate also suppresses the hover-end blur while the list is collapsed: a pointer leaving an item after the list has closed no longer clears the active item or returns focus to the combobox. This matches the reading that nothing hover-driven should change state while the list is collapsed, and the library fix follows the same rule `SelectItem` already applies to both paths.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the stage-1 reproduction direction</summary>

**Assessment**

The formal pre-commit gate produced findings in rounds 1 (6), 2 (3), and 3 (1) for one unchanged direction: extend the semantic `combobox-select-collapsed-focus` sandbox with the reproduction and remove the contract-reversing test from `combobox-select-open-active-item`. Every finding was hygiene-grade — redundant waits, query-helper style, comment wording, commit-message provenance, and an unrun verification command — and none questioned the fixture placement, the contract shape, or the deletion decision. The underlying bug is a confirmed, user-visible focus-stealing defect (`bug`, `a11y`) in a stable API, so the reproduction work is required by the issue-fix workflow. The direction adds one small fixture component and six test behaviors duplicated across the two mandated test surfaces, with no new machinery or public contracts.

**Decision**

Continue the current direction. The finding counts are strictly decreasing and converging on hygiene, which indicates reviewer thoroughness rather than a design problem. Alternatives were weighed and rejected: an isolated issue-number sandbox would contradict the fixture lifecycle preference for coherent semantic fixtures, and dropping the default-contract guard tests would leave the fix free to silently flip the default predicate. An independent fresh-context complexity subagent performed the same reassessment in parallel and reached the same recommendation; the assessments were compared before recording.

Intentionally uncovered, recorded so later rounds do not expand scope onto them: a `virtualFocus` (aria-activedescendant) hover variant (same predicate branch; the real-focus variant is the stronger signal and the one the issue demonstrates), an authored `focusOnHover` on a plain combobox with a closed `alwaysVisible` list (identical authored branch to the covered select case), and a Solid duplicate (the sandbox has no Solid entry, so `withFramework` runs React only).

</details>

<details>
<summary>Cycle 6: Continue; close the hover-end coverage gap with one capped test pair</summary>

**Assessment**

Cycles 4 through 6 (workaround pass 1, fix pass 1, fix pass 3) each produced findings for the same unchanged direction: a userland gate in the sandbox, then a `SelectItem`-parity open gate in `useComboboxItem`. Cycle 5 was substantive and improved the fix shape: the open check was hoisted before the authored predicate so side-effectful callbacks (the select-grid `move()` pattern) don't run while collapsed, pinned by a red-checked regression pair. The remaining cycle-6 finding was the last uncovered branch of the same single predicate: the hover-end direction claimed by the changeset had no regression pin. The fix itself stayed three lines plus a JSDoc note and a changeset; the test surface growth is bounded, with every test mapping to a distinct consumer-visible claim in the changeset or JSDoc.

**Decision**

Continue; address the hover-end gap with exactly one test per duplicate against the existing `HoverSelect` fixture (no new fixture), red-checked against the pre-fix predicate (`data-active-item` cleared to `null` on hover-away while collapsed). An independent fresh-context complexity subagent reached the same recommendation and the same scope cap; the assessments were compared before recording. Narrowing the changeset claim instead was rejected because the leave path is genuinely gated, so dropping the clause would misdescribe shipped behavior. Recorded without action in this PR: the fix comment no longer claims literal statement-order parity with `SelectItem` (the reorder is deliberate), and `SelectItem`'s own predicate-before-open-check exposure is registered as #7120 (pre-existing on `main`; that component is soft-deprecated in favor of `useComboboxItem`). The default-predicate guards do not foreclose #6837 (enabling `focusOnHover` by default for plain comboboxes in v0.5); they make any future default flip deliberate rather than accidental.

</details>

<details>
<summary>Cycle 10: Continue the review-response work; delete the disputed wait instead of rewording it</summary>

**Assessment**

The gate cycles responding to this PR's two inline review threads produced one substantive outcome (the `focusOnHover` post-callback re-read with its red-checked `HideHoverSelect` coverage, requested by the review thread) and then two successive comment-accuracy findings on a single `flushFrames` justification in the new browser test. Each proposed wording named a deferral mechanism that source inspection could not locate on this fixture's path: hover activation goes through `setActiveId`, which never bumps `moves`, so no presentation passive effect can fire, and a list-only composition has no post-hide focus restoration. The finding pattern is documentation convergence on one comment, not design churn; severities are monotonically decreasing and no cycle questioned the fix, the fixture, or the response scope. A cycle-count correction is also recorded: an earlier finding-bearing pass was missed in the count, so the prior reassessment ran one cycle later than its nominal boundary; counting continues from the recorded boundaries.

**Decision**

Continue; no revert, fixture removal, or contract narrowing. The cycle disposition changed from rewording the comment to deleting the `flushFrames` call and its comment from that one test: unlike its siblings, the test awaits a positive `aria-expanded` transition as its settle point, and the buggy shape commits the activation and focus steal synchronously in the same handler that hides the list, which the re-run browser red check confirmed (still solidly red without the wait). An independent fresh-context complexity subagent performed the same reassessment in parallel and recommended exactly this disposition change; the assessments were compared and the disagreement investigated against source before recording. Declared out of scope for this PR: refactoring the pre-existing duplicated `flushFrames` comments from the #7093 tests.

</details>

### DIFF
--- a/packages/ariakit-react-components/src/combobox/combobox-item.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-item.tsx
@@ -283,16 +283,19 @@ export const useComboboxItem = createHook<TagName, ComboboxItemOptions>(
       },
     });
 
-    const focusOnHoverProp = useBooleanEvent(focusOnHover ?? false);
+    const focusOnHoverProp = useBooleanEvent(focusOnHover ?? selectMode);
 
     props = useCompositeHover({
       store,
       ...props,
+      // Disable focusOnHover when the popup is closed, even for authored
+      // values, so pointer hover can't activate an item or move focus while
+      // the combobox is collapsed. The open check comes first so authored
+      // callbacks with side effects don't run at all while closed.
+      // https://github.com/ariakit/ariakit/issues/7118
       focusOnHover(event) {
-        if (focusOnHover !== undefined) {
-          return focusOnHoverProp(event);
-        }
-        return selectMode && store.getState().open;
+        if (!store.getState().open) return false;
+        return focusOnHoverProp(event);
       },
     });
 
@@ -430,6 +433,9 @@ export interface ComboboxItemOptions<T extends ElementType = TagName>
   /**
    * Defaults to `false`, or `true` when the item is used with a
    * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select).
+   *
+   * Regardless of the value, hovering an item never activates it or moves
+   * focus while the combobox is closed.
    */
   focusOnHover?: CompositeHoverOptions["focusOnHover"];
 }
--- a/packages/ariakit-react-components/src/combobox/combobox-item.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-item.tsx
@@ -291,11 +291,13 @@ export const useComboboxItem = createHook<TagName, ComboboxItemOptions>(
       // Disable focusOnHover when the popup is closed, even for authored
       // values, so pointer hover can't activate an item or move focus while
       // the combobox is collapsed. The open check comes first so authored
-      // callbacks with side effects don't run at all while closed.
+      // callbacks with side effects don't run at all while closed, and it's
+      // re-read afterwards because the callback itself may close the list.
       // https://github.com/ariakit/ariakit/issues/7118
       focusOnHover(event) {
         if (!store.getState().open) return false;
-        return focusOnHoverProp(event);
+        if (!focusOnHoverProp(event)) return false;
+        return store.getState().open;
       },
     });
 
--- a/packages/ariakit-react-components/src/combobox/combobox-item.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-item.tsx
@@ -436,8 +436,10 @@ export interface ComboboxItemOptions<T extends ElementType = TagName>
    * Defaults to `false`, or `true` when the item is used with a
    * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select).
    *
-   * Regardless of the value, hovering an item never activates it or moves
-   * focus while the combobox is closed.
+   * Regardless of the value, hover has no effect while the combobox is
+   * closed: hovering an item never activates it or moves focus, and moving
+   * the pointer off an item never clears the active item or moves focus back
+   * to the combobox.
    */
   focusOnHover?: CompositeHoverOptions["focusOnHover"];
 }
--- a/.changeset/300-combobox-item-focus-on-hover-closed.md
+++ b/.changeset/300-combobox-item-focus-on-hover-closed.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`ComboboxItem`](https://ariakit.com/reference/combobox-item) so an explicit [`focusOnHover`](https://ariakit.com/reference/combobox-item#focusonhover) boolean or callback value neither activates an item on hover nor clears the active item on hover end while the combobox is closed.
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -151,23 +151,33 @@ interface HoverSelectProps {
 // which the default predicate alone never reaches.
 // https://github.com/ariakit/ariakit/issues/7118
 function HoverSelect({ label, focusOnHover }: HoverSelectProps) {
+  // TODO: Remove this store hoist and the open-state gate in focusOnHover
+  // below once https://github.com/ariakit/ariakit/issues/7118 is fixed.
+  const combobox = Ariakit.useComboboxStore({
+    defaultSelectedValue: "Apple",
+    virtualFocus: false,
+  });
+
   return (
     <div>
       {/* Safari needs an explicit tabindex to focus the clicked button. */}
       <button type="button" tabIndex={0}>
         {label} other control
       </button>
-      <Ariakit.ComboboxProvider
-        defaultSelectedValue="Apple"
-        virtualFocus={false}
-      >
+      <Ariakit.ComboboxProvider store={combobox}>
         <Ariakit.ComboboxSelect aria-label={label} />
         <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
           {fruits.map((value) => (
             <Ariakit.ComboboxItem
               key={value}
               value={value}
-              focusOnHover={focusOnHover}
+              focusOnHover={(event) => {
+                if (!combobox.getState().open) return false;
+                if (typeof focusOnHover === "function") {
+                  return focusOnHover(event);
+                }
+                return focusOnHover ?? false;
+              }}
             />
           ))}
         </Ariakit.ComboboxList>
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -141,6 +141,41 @@ function FilterCombobox() {
   );
 }
 
+interface HoverSelectProps {
+  label: string;
+  focusOnHover: Ariakit.ComboboxItemProps["focusOnHover"];
+}
+
+// The pointer side of the same contract. An explicit boolean and a callback
+// are authored here so the collapsed gate is exercised on the authored path,
+// which the default predicate alone never reaches.
+// https://github.com/ariakit/ariakit/issues/7118
+function HoverSelect({ label, focusOnHover }: HoverSelectProps) {
+  return (
+    <div>
+      {/* Safari needs an explicit tabindex to focus the clicked button. */}
+      <button type="button" tabIndex={0}>
+        {label} other control
+      </button>
+      <Ariakit.ComboboxProvider
+        defaultSelectedValue="Apple"
+        virtualFocus={false}
+      >
+        <Ariakit.ComboboxSelect aria-label={label} />
+        <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
+          {fruits.map((value) => (
+            <Ariakit.ComboboxItem
+              key={value}
+              value={value}
+              focusOnHover={focusOnHover}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+    </div>
+  );
+}
+
 export default function Example() {
   return (
     <>
@@ -149,6 +184,8 @@ export default function Example() {
       <CodeSelect />
       <StageSelect />
       <FilterCombobox />
+      <HoverSelect label="Hover fruit" focusOnHover />
+      <HoverSelect label="Callback hover fruit" focusOnHover={() => true} />
     </>
   );
 }
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -216,6 +216,46 @@ function MoveHoverSelect() {
   );
 }
 
+// An authored callback that closes the list from inside the predicate and
+// still returns true. The gate must re-read the live open state after the
+// callback, or the stale pre-check result would activate the item and steal
+// focus right as the list collapses.
+// https://github.com/ariakit/ariakit/issues/7118
+function HideHoverSelect() {
+  const combobox = Ariakit.useComboboxStore({
+    defaultSelectedValue: "Apple",
+    virtualFocus: false,
+  });
+
+  return (
+    <div>
+      {/* Safari needs an explicit tabindex to focus the clicked button. */}
+      <button type="button" tabIndex={0} onClick={() => combobox.show()}>
+        Open hide hover fruit list
+      </button>
+      <Ariakit.ComboboxProvider store={combobox}>
+        <Ariakit.ComboboxSelect aria-label="Hide hover fruit" />
+        <Ariakit.ComboboxList
+          alwaysVisible
+          aria-label="Hide hover fruit options"
+        >
+          {fruits.map((value) => (
+            <Ariakit.ComboboxItem
+              key={value}
+              value={value}
+              focusOnHover={(event) => {
+                if (event.type === "mouseleave") return false;
+                combobox.hide();
+                return true;
+              }}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+    </div>
+  );
+}
+
 export default function Example() {
   return (
     <>
@@ -227,6 +267,7 @@ export default function Example() {
       <HoverSelect label="Hover fruit" focusOnHover />
       <HoverSelect label="Callback hover fruit" focusOnHover={() => true} />
       <MoveHoverSelect />
+      <HideHoverSelect />
     </>
   );
 }
--- a/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
+++ b/app/src/sandbox/combobox-select-collapsed-focus/index.react.tsx
@@ -151,8 +151,37 @@ interface HoverSelectProps {
 // which the default predicate alone never reaches.
 // https://github.com/ariakit/ariakit/issues/7118
 function HoverSelect({ label, focusOnHover }: HoverSelectProps) {
-  // TODO: Remove this store hoist and the open-state gate in focusOnHover
-  // below once https://github.com/ariakit/ariakit/issues/7118 is fixed.
+  return (
+    <div>
+      {/* Safari needs an explicit tabindex to focus the clicked button. */}
+      <button type="button" tabIndex={0}>
+        {label} other control
+      </button>
+      <Ariakit.ComboboxProvider
+        defaultSelectedValue="Apple"
+        virtualFocus={false}
+      >
+        <Ariakit.ComboboxSelect aria-label={label} />
+        <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
+          {fruits.map((value) => (
+            <Ariakit.ComboboxItem
+              key={value}
+              value={value}
+              focusOnHover={focusOnHover}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxProvider>
+    </div>
+  );
+}
+
+// An authored callback with a side effect: the consumer moves the composite
+// from inside the predicate, like the select-grid example does. The closed
+// gate must keep the callback from running at all, or the move would still
+// activate the item and steal focus.
+// https://github.com/ariakit/ariakit/issues/7118
+function MoveHoverSelect() {
   const combobox = Ariakit.useComboboxStore({
     defaultSelectedValue: "Apple",
     virtualFocus: false,
@@ -162,21 +191,22 @@ function HoverSelect({ label, focusOnHover }: HoverSelectProps) {
     <div>
       {/* Safari needs an explicit tabindex to focus the clicked button. */}
       <button type="button" tabIndex={0}>
-        {label} other control
+        Move hover fruit other control
       </button>
       <Ariakit.ComboboxProvider store={combobox}>
-        <Ariakit.ComboboxSelect aria-label={label} />
-        <Ariakit.ComboboxList alwaysVisible aria-label={`${label} options`}>
+        <Ariakit.ComboboxSelect aria-label="Move hover fruit" />
+        <Ariakit.ComboboxList
+          alwaysVisible
+          aria-label="Move hover fruit options"
+        >
           {fruits.map((value) => (
             <Ariakit.ComboboxItem
               key={value}
               value={value}
               focusOnHover={(event) => {
-                if (!combobox.getState().open) return false;
-                if (typeof focusOnHover === "function") {
-                  return focusOnHover(event);
-                }
-                return focusOnHover ?? false;
+                if (event.type === "mouseleave") return false;
+                combobox.move(event.currentTarget.id);
+                return true;
               }}
             />
           ))}
@@ -196,6 +226,7 @@ export default function Example() {
       <FilterCombobox />
       <HoverSelect label="Hover fruit" focusOnHover />
       <HoverSelect label="Callback hover fruit" focusOnHover={() => true} />
+      <MoveHoverSelect />
     </>
   );
 }
--- a/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
+++ b/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
@@ -203,6 +203,92 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
   }
 
   // https://github.com/ariakit/ariakit/issues/7118
+  // The other hover direction: after the list closes by clicking an option,
+  // moving the pointer off that option must not clear the active item or move
+  // focus while the list is collapsed.
+  test("hover end on a collapsed list keeps the active item", async ({
+    page,
+    q,
+  }) => {
+    const select = q.combobox("Hover fruit");
+    const list = q.listbox("Hover fruit options");
+    const grape = query(list).option("Grape");
+    const other = q.button("Hover fruit other control");
+
+    await select.click();
+    await test.expect(select).toHaveAttribute("aria-expanded", "true");
+    await grape.hover();
+    await test.expect(grape).toHaveAttribute("data-active-item");
+
+    await grape.click();
+    await test.expect(select).toHaveAttribute("aria-expanded", "false");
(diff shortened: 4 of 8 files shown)


packages/ariakit-react-components/src/combobox/combobox-item.tsx
```diff
@@ -430,6 +433,9 @@ export interface ComboboxItemOptions<T extends ElementType = TagName>
   /**
    * Defaults to `false`, or `true` when the item is used with a
    * [`ComboboxSelect`](https://ariakit.com/reference/combobox-select).
+   *
+   * Regardless of the value, hovering an item never activates it or moves
+   * focus while the combobox is closed.
```
diegohaz: This new sentence documents only half of the closed-state rule. The same gate also covers the hover-end direction: while the combobox is closed, moving the pointer off an item no longer clears the active item or moves focus back to the combobox. The changeset states both halves ("neither activates an item on hover nor clears the active item on hover end"), and the "hover end on a collapsed list keeps the active item" tests pin the behavior, but a consumer reading only the `focusOnHover` reference entry cannot learn the hover-end half. That half is arguably the more surprising one in practice: after clicking an option in an always-visible list, the item stays active and focused even as the pointer moves away, and nothing in these docs says so.

**Assessment**

- Severity: Low (verified), because this is a reference-docs gap only; the shipped behavior is correct and covered by red-checked tests.
- Priority: P2 (estimate, 80% confidence), because it is a one-sentence docs completion that fits naturally in this PR.
- Breakage likelihood: Possible (estimate, 70% confidence), in the sense that consumers who read only the reference entry would not expect the active item to persist on hover end while collapsed; there is no runtime impact.
- Fix effort: Easy (verified), because the suggestion below is the whole change.

```suggestion
   * Regardless of the value, hovering an item never activates it or moves
   * focus while the combobox is closed, and moving the pointer off an item
   * never clears the active item or moves focus back to the combobox.
```

diegohaz (commented):
## Review summary

Initial review of the full diff from the merge base [`a0426ed`](https://github.com/ariakit/ariakit/commit/a0426ed547d95b84c9d53033053e51baeaca4aaa) to head [`77ed9c7`](https://github.com/ariakit/ariakit/commit/77ed9c74730caee2fe18ed627ca76d72a8d80dad), plus the surrounding context: `useCompositeHover` (both hover directions), `useSelectItem` (the parity claim), `useBooleanEvent` reactivity, `useMenuItem` (its own hover predicate, unaffected), the [`6f79732`](https://github.com/ariakit/ariakit/commit/6f79732407cebf3ffda58d8c74be5fb14e03969e) / #6832 history that introduced the honored override, issue #7118, and follow-up #7120. There were no existing review threads to validate.

## Validation performed

- `pnpm test combobox-select-collapsed-focus combobox-select-open-active-item` at head: 30/30 passed.
- Red-check experiment: with only `combobox-item.tsx` reverted to `origin/main`, exactly the four #7118 regression tests failed for the expected reasons (`data-active-item` set on hover while collapsed for the boolean, callback, and side-effectful callback cases; `data-active-item` cleared to `null` on hover end). The 12 guard tests stayed green, so none of the new tests pass vacuously, including the hover-end test whose Playwright variant depends on the tracker re-arming move. The file was restored afterwards; the branch is untouched.
- CI: the first commit [`a479ab5`](https://github.com/ariakit/ariakit/commit/a479ab5cb) recorded the red check across happy-dom, React 18, Chrome, Firefox, and Safari, with failure annotations matching the new assertions. Head is 44/44 green, including type check, lint, docs generation, and all browser matrices, so I did not rerun the Playwright suites locally.
- Changeset: `https://ariakit.com/reference/combobox-item` returns 200 and the deployed page has the `#focusonhover` anchor. Frontmatter is correct: `ComboboxItem` is publicly re-exported by `@ariakit/react`. No contributor credit is required because #7118 was reported by `ariakit-bot`.
- Confirmed `@ariakit/solid-components` has no combobox port, so the React-only `withFramework` coverage is right, and confirmed #7120 is open and records the `SelectItem` exposure as pre-existing on `main`.

## Alternatives weighed

- **Predicate-first order** (the `SelectItem` shape and the sketch in #7118): the PR's open-check-first order is strictly safer because authored callbacks with side effects, like the select-grid `move()` pattern, never run while the list is collapsed, and a red-checked test pins that. The divergence from `SelectItem`'s literal order is deliberate, and #7120 tracks the `SelectItem` side. Best available trade-off.
- **Gating in `useCompositeHover` instead**: not viable; the composite layer has no disclosure state, which is owned by the combobox/select layer, so the item-level gate is the right altitude.
- **Keeping an authored escape hatch while closed** (the #6832 honored-override contract): declined non-goal. Issue #7118 defines the closed-state rule for the default, an explicit boolean, and a callback alike, and a consumer who wants a permanently interactive always-visible list can keep the store open. Removing the `ClosedAlwaysVisibleSelect` fixture and its test is coherent because that test pinned the exact contract this PR reverses, and the scenario now lives in the new sandbox's `HoverSelect` fixtures with the fixed expectations.

## Findings

One inline comment: the new `focusOnHover` JSDoc documents only the hover-activation half of the closed-state rule, while the changeset and tests also establish the hover-end half. Docs-only, with a one-sentence suggestion attached.

No other findings. The library change is minimal and reduces branching relative to `main` (the old two-branch predicate collapses into one gated line with the default folded into `useBooleanEvent(focusOnHover ?? selectMode)`, which stays reactive via `useEvent`). The sandbox extension is additive and preserves the existing #7093/#7098 fixture invariants, waits follow the documented mechanisms, and every new test maps to a consumer-visible claim.

## Not fully validated

I did not run the dual reference-doc generators because the JSDoc change is wording-only and the Docs CI job is green on head. I also did not exercise the `virtualFocus` hover variant or a plain-combobox authored value on a collapsed list; the PR records both as intentionally uncovered, and they share the gated branch that the covered cases exercise.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
diegohaz: Confirmed against the current head, so this remains **Address now**. `useCompositeHover` evaluates this same predicate on `mouseleave` before it clears the active item and focuses the composite owner. The happy-dom and browser tests both pin that behavior, but the new public JSDoc still describes only pointer entry. Both reference extractors expose this JSDoc while the inherited `blurOnHoverEnd` text remains generic, so a reference-only reader cannot discover the closed-state pointer-exit rule. The existing suggestion accurately completes the contract.

**Assessment**

- Severity: Low (verified), because the runtime behavior is correct and the omission is limited to public reference documentation.
- Priority: P2 (estimate, 90% confidence), because this PR introduces and claims to document the two-direction closed-state contract.
- Breakage likelihood: Possible (estimate, 75% confidence), when a consumer uses an always-visible list, moves the pointer off an active item after the list closes, and relies on the reference entry to predict focus behavior.
- Fix effort: Easy (verified), because the existing one-sentence suggestion is sufficient and both extractors already propagate the JSDoc.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
```diff
@@ -283,16 +283,19 @@ export const useComboboxItem = createHook<TagName, ComboboxItemOptions>(
...
+      // the combobox is collapsed. The open check comes first so authored
+      // callbacks with side effects don't run at all while closed.
+      // https://github.com/ariakit/ariakit/issues/7118
       focusOnHover(event) {
-        if (focusOnHover !== undefined) {
-          return focusOnHoverProp(event);
-        }
-        return selectMode && store.getState().open;
+        if (!store.getState().open) return false;
+        return focusOnHoverProp(event);
```
diegohaz: The `open` guard can become stale while the authored callback runs. If the list starts open and `focusOnHover` synchronously calls the public `store.hide()` method before returning `true`, the first check passes, the callback closes the store, and this return still lets `useCompositeHover` focus the composite owner and set the hovered item active. I reproduced this on the current head with an always-visible list: after hover, `aria-expanded` was `false`, the hovered item had `data-active-item`, and focus had moved from an outside button to the collapsed select.

This is a supported callback shape, not a purity-only hypothetical. The public API accepts callbacks, this PR explicitly covers side-effectful callbacks, and the select-grid example calls `store.move()` inside one. A custom select grid or command picker that dismisses itself from its hover predicate can therefore move keyboard and assistive-technology context into a collapsed widget, contrary to the new JSDoc and changeset contract.

Please keep the pre-check so callbacks never run when the list starts closed, then recheck the live state after the callback:

```ts
focusOnHover(event) {
  if (!store.getState().open) return false;
  if (!focusOnHoverProp(event)) return false;
  return store.getState().open;
},
```

The same focused regression passes with this guard. It should be duplicated in the existing browser and happy-dom fixture surfaces.

**Assessment**

- Severity: Medium (estimate, 90% confidence), because the failure changes active state and steals focus after the widget has become collapsed.
- Priority: P1 (estimate, 90% confidence), because the implementation does not fully enforce this PR's stated accessibility contract.
- Breakage likelihood: Rare (estimate, 85% confidence), because it requires an open list plus a callback that synchronously closes the store and returns `true`.
- Fix effort: Easy (verified by the local experiment), because one post-callback state check and a focused regression cover it.

diegohaz (commented):
## Re-review summary

This re-review used the latest prior marked head, [`77ed9c7`](https://github.com/ariakit/ariakit/commit/77ed9c74730caee2fe18ed627ca76d72a8d80dad), as the incremental baseline. There are no new commits since that review. I independently re-inspected the full PR diff from merge base [`a0426ed`](https://github.com/ariakit/ariakit/commit/a0426ed547d95b84c9d53033053e51baeaca4aaa) through the same current head, plus `useCompositeHover`, `useBooleanEvent`, store update ordering, the select-grid callback pattern, both sandbox test surfaces, the removed reversed-contract fixture, public documentation, changeset scope, issue #7118, and follow-up #7120.

## Findings

- One new inline finding: the pre-callback `open` check can become stale if the authored predicate synchronously closes the store and returns `true`. This reproduces the same collapsed-widget active-state and focus-stealing failure that the PR intends to prevent. Disposition: Address now.
- The unresolved documentation thread at `#discussion_r3763724100` is confirmed against the current head. The public JSDoc omits the hover-end half of the closed-state rule that the changeset and tests establish. I replied on that thread with the validation result and Address now disposition.
- No other correctness, compatibility, performance, security, test-design, or substantive simplification findings.

## Validation performed

- `pnpm test combobox-select-collapsed-focus combobox-select-open-active-item`: 30 of 30 tests passed.
- `pnpm test-react18 combobox-select-collapsed-focus combobox-select-open-active-item`: 30 of 30 tests passed.
- Temporary public-API regression experiment: with an initially open always-visible list, outside focus, and `focusOnHover={() => { store.hide(); return true; }}`, the current code closed the list but still activated the item and moved focus to the collapsed select. Adding a post-callback live `open` check made the same test pass. All experiment changes were removed, and the worktree is clean.
- The original test-only commit [`a479ab5`](https://github.com/ariakit/ariakit/commit/a479ab5cbe7e2d68cf1b38fc940ab6e261a3318e) failed on the intended boolean and callback `data-active-item` assertions across happy-dom, React 18, Chrome, Firefox, and Safari. Current head CI is green across all checks, including the browser matrix, type check, lint, docs, builds, and performance.
- `git diff --check` passed. The existing browser recording resolves as a WebM attachment.

## Alternatives weighed

Checking `open` only after the callback would reintroduce the documented side-effect problem because callbacks would run when the list starts closed. Keeping only the current pre-check leaves the open-to-closed race reproduced above. Checking both before and after the callback is the smallest complete gate. Moving disclosure logic into `useCompositeHover` remains the wrong layer because the composite hook does not own the combobox open state.

## Not fully validated

I did not rerun the local Playwright suites because the current head has a fully green Chrome, Firefox, Safari, Android, and iOS CI matrix. The new event-order experiment ran in the public happy-dom surface; its active-state and DOM-focus effects are observable there without browser-only APIs.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
diegohaz: Fixed in [`6ee5e1762`](https://github.com/ariakit/ariakit/commit/6ee5e1762775b06074a91cfc700d7fbb551dd325): the JSDoc now states both halves of the closed-state rule. The suggestion's wording was adjusted so the closed condition governs both clauses; appended verbatim, the second clause would read as unconditional and be false while the combobox is open, where hover end does clear the active item.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
diegohaz: Fixed in [`4bd851cb5`](https://github.com/ariakit/ariakit/commit/4bd851cb55c1f63afa135b026e37ea7f7f445f8b): the pre-check stays so callbacks never run while the list starts closed, and the live open state is re-read after the callback runs. The same commit adds the HideHoverSelect fixture with one regression per test surface, red-checked in both the browser and happy-dom duplicates against the previous predicate shape (the hovered item kept `data-active-item` after the list collapsed).

diegohaz:
**Fixed**

- [`4bd851cb5`](https://github.com/ariakit/ariakit/commit/4bd851cb55c1f63afa135b026e37ea7f7f445f8b) ([reply](https://github.com/ariakit/ariakit/pull/7119#discussion_r3780046548)): the `focusOnHover` predicate now re-reads the live open state after the authored callback runs, so a callback that closes the list from inside the predicate can no longer activate the item or steal focus with a stale result. Includes the `HideHoverSelect` fixture and one regression per test surface, red-checked in both the browser and happy-dom duplicates.
- [`6ee5e1762`](https://github.com/ariakit/ariakit/commit/6ee5e1762775b06074a91cfc700d7fbb551dd325) ([reply](https://github.com/ariakit/ariakit/pull/7119#discussion_r3780045724)): the `focusOnHover` reference docs now state both halves of the closed-state rule, with the closed condition scoped to govern both clauses rather than the suggestion's verbatim wording, whose second clause would read as unconditional and be false while the combobox is open.

diegohaz (commented):
## Re-review summary

Re-review from the latest prior marked review baseline [`77ed9c7`](https://github.com/ariakit/ariakit/commit/77ed9c74730caee2fe18ed627ca76d72a8d80dad) (review `pr-7119-77ed9c747-20260813T010644Z`) to the current head [`6ee5e17`](https://github.com/ariakit/ariakit/commit/6ee5e1762775b06074a91cfc700d7fbb551dd325). The incremental diff is the two commits responding to the prior inline threads, and I also re-inspected the surrounding predicate, fixture, and test context for newly exposed issues.

## Validation of the fix commits

Both prior findings are resolved correctly on the current head; their threads are resolved and needed no further reply.

- [`4bd851c`](https://github.com/ariakit/ariakit/commit/4bd851cb55c1f63afa135b026e37ea7f7f445f8b) (stale open guard): the predicate keeps the pre-check, so authored callbacks still never run while the list starts closed, and re-reads the live open state after the callback, so a callback that synchronously hides the store can no longer activate the item with a stale result. The two `store.getState()` reads are semantically required, not duplication. I verified the red check locally: with only `combobox-item.tsx` reverted to the previous predicate shape, exactly the new "a callback that closes the list on hover activates nothing" test fails, with `data-active-item` present after the list collapsed, and the other 16 tests stay green. The new tests also cannot pass vacuously, because the asserted `aria-expanded="false"` flip itself requires the authored callback to have run.
- [`6ee5e17`](https://github.com/ariakit/ariakit/commit/6ee5e1762775b06074a91cfc700d7fbb551dd325) (JSDoc hover-end gap): the reference docs now state both directions with the closed condition governing both clauses. The wording is an improvement over the original suggestion, whose appended clause would have read as unconditional and been false while the combobox is open.

## Checks run

- `pnpm test combobox-select-collapsed-focus combobox-select-open-active-item` at head: 31/31 passed.
- Red-check experiment described above; the working tree was restored afterwards and the branch is untouched.
- CI on head: 44/44 checks green, including type check, lint, docs generation, and the Chrome, Firefox, Safari, and mobile browser matrices, so I did not rerun the Playwright suites locally.
- Confirmed the PR description now describes the post-callback re-read and the `HideHoverSelect` coverage, and that the changeset needs no update, since the guard enforces the already-documented closed-state contract.

## Findings

No new findings. The `HideHoverSelect` fixture models the reported hazard precisely, the browser test's settle point is the observable collapse rather than an arbitrary wait, and the predicate remains minimal for the contract it enforces.

PR #7136, opened 2026-08-14
Update Astro

diegohaz:
## Motivation

The docs app should use the current Astro dependency group. Astro 7.2.1 also validates content references after loaders finish, which exposed three example metadata entries that point to component pages that do not exist.

## Solution

Update `astro` and `@astrojs/cloudflare`, regenerate and deduplicate the pnpm lockfile, and remove the stale `dialog`, `checkbox`, and `combobox` component references. The `combobox-group` example retains its valid `popover` reference, while examples with no valid component references use the schema's empty-array default.

```yaml
# Before
components:
  - combobox
  - popover

# After
components:
  - popover
```

The same unchanged source produced no invalid-reference diagnostics with Astro 7.2.0. Astro 7.2.2 reported all three stale references, which confirms that the metadata cleanup responds to Astro's new validation.

## Dependencies

| Package | Workspace | Old | New | Sources |
| `astro` | `app` | `7.2.0` | `7.2.2` | [7.2.2 notes](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#722), [7.2.1 notes](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#721), [compare](https://github.com/withastro/astro/compare/astro%407.2.0...astro%407.2.2) |
| `@astrojs/cloudflare` | `app` | `14.2.0` | `14.2.1` | [14.2.1 notes](https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/CHANGELOG.md#1421), [compare](https://github.com/withastro/astro/compare/%40astrojs%2Fcloudflare%4014.2.0...%40astrojs%2Fcloudflare%4014.2.1) |

Astro 7.2.2 includes fixes for Cloudflare incremental image builds, content-entry styles in fallback development environments, and large image builds. The Cloudflare adapter update corrects its Astro peer range to `^7.2.0` and updates its redirects helper.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm build-app-lite`
- `pnpm --filter ./app check` on unchanged source with Astro 7.2.0 for the diagnostic comparison


diegohaz:
Astro 7.2.1 added validation for content references after loaders finish. I ran the app check against the same unchanged source with both dependency states: Astro 7.2.0 reported no invalid references, while Astro 7.2.2 reported `dialog-framer-motion -> dialog`, `checkbox-card -> checkbox`, and `combobox-group -> combobox`.

The migration removes only those unresolved IDs. It retains `combobox-group -> popover`, which resolves to an existing component entry, and relies on the schema's empty-array default where no valid component reference remains.

This source migration makes the dependency group non-mechanical, so the PR requires human review and has no automerge.

PR #7154, opened 2026-08-18
Fix middle click on disabled elements rendered as links

### LINKED ISSUE #7153: Disabled elements rendered as links open their destination on middle click
### Current behavior

A disabled element rendered as a link still opens its destination on a middle click.

[`Focusable`](https://ariakit.com/reference/focusable) blocks activation by disabling three events in the capture phase: `click`, `mousedown`, and `keypress`. A middle click doesn't fire `click`, it fires `auxclick`, and nothing in the library listens for it. Calling `preventDefault()` on `mousedown` doesn't stop it either.

When the element is truly disabled, `Focusable` also applies `pointer-events: none`, so the pointer never reaches it and the gesture is blocked by accident rather than by design. Setting [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) removes `pointer-events: none` to keep the element reachable, and the middle click goes through.

This means [`Tab`](https://ariakit.com/reference/tab) is affected at its default settings, because `Tab` defaults `accessibleWhenDisabled` to `true`:

```ts
// packages/ariakit-react-components/src/tab/tab.tsx:103-104
const accessibleWhenDisabled =
  selected || (props.accessibleWhenDisabled ?? true);
```

Measured with Playwright 1.62.1 against the current `main`, dispatching a real middle click at the element's centre:

| Element | `pointer-events` | Chromium | Firefox | WebKit |
| `<Focusable disabled render={<a href />}>` | `none` | blocked | blocked | blocked |
| `<Focusable disabled accessibleWhenDisabled render={<a href />}>` | `auto` | opens a new tab | opens a new tab | navigates in place |
| `<Tab disabled render={<a href />}>` (default settings) | `auto` | opens a new tab | opens a new tab | navigates in place |

Every other gesture is already handled correctly: left click, <kbd>Enter</kbd>, and modifier clicks are all blocked by the existing capture-phase handlers in all three engines.

2. Replace the contents of `src/app.tsx` with:

```jsx

  return (
    <Ariakit.TabProvider>
      <Ariakit.TabList>
        <Ariakit.Tab disabled render={<a href="https://ariakit.org" />}>
          Disabled tab
        </Ariakit.Tab>
      </Ariakit.TabList>
    </Ariakit.TabProvider>
  );
}
```

3. Middle click the disabled tab.
4. It opens https://ariakit.org in a new tab. Expected: nothing happens.

The same result can be reproduced with `<Ariakit.Focusable disabled accessibleWhenDisabled render={<a href="https://ariakit.org" />}>`.

**Note for whoever writes the regression test:** Playwright's `click({ button: "middle" })` will not reproduce this. Its actionability check refuses elements carrying `aria-disabled="true"` and fails with `element is not enabled` before delivering the event. The test has to bypass that, either with `force: true` or by driving `page.mouse` at the element's coordinates. Using raw mouse events is preferable, because it still performs real hit testing and therefore also verifies that `pointer-events: none` blocks the default case.

### Expected behavior

A disabled element should not navigate through any gesture, whether or not [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) is enabled. `pointer-events: none` shouldn't be the only thing standing between a disabled link and its destination, since it's absent by design whenever the element must stay reachable.

### Workaround

Pass the capture handler yourself. `Focusable` forwards it to the element, so this works today:

```jsx
<Ariakit.Tab
  disabled
  onAuxClickCapture={(event) => {
    event.stopPropagation();
    event.preventDefault();
  }}
  render={<a href="https://ariakit.org" />}
>
  Disabled tab
</Ariakit.Tab>
```

Verified to block the middle click in Chromium, Firefox, and WebKit, while an enabled link in the same page still opens normally.

Alternatively, omit `href` while the element is disabled, which removes the destination entirely and closes every gesture at once, including the context menu and dragging the link to the address bar:

```jsx
<Ariakit.Tab disabled={disabled} render={<a href={disabled ? undefined : url} />}>
  Tab
</Ariakit.Tab>
```

Additional assumptions and limitations: https://github.com/ariakit/ariakit/issues/7153#issuecomment-5329143814

### Possible solutions

Add a fourth `useDisableEvent` call next to the three that already exist in `packages/ariakit-react-components/src/focusable/focusable.tsx:299-307`:

```ts
const onAuxClickCapture = useDisableEvent(props.onAuxClickCapture, disabled);
```

Then include `onAuxClickCapture` in the props object returned at the end of `useFocusable`, alongside the existing `onClickCapture` and `onMouseDownCapture`.

This benefits every component built on `Focusable`, not just links, and it costs one more handler per focusable element. `onAuxClickCapture` is typed in React 18 and 19, and React has delegated `auxClick` since version 17, so `pnpm test-react18` should be run as well.

Two related gestures stay unguarded and are deliberately out of scope here, since removing the destination is the only real fix for them: the context menu (`Open link in new tab`, `Copy link address`) and dragging the link to the address bar. Both are blocked by `pointer-events: none` in the default case and both remain available under `accessibleWhenDisabled`. Note also that `focusable.tsx` spreads the consumer's `style` after its own, so a `style={{ pointerEvents: "auto" }}` written to get `cursor: not-allowed` silently reopens all three gestures.

This was found while designing the `Link` component in #7112. `Link` isn't affected, because it withholds `href` while disabled, but it can't withhold an `href` that a `render` element supplies, so the guard matters there too.

diegohaz:
## Motivation

A disabled element rendered as a link still opens its destination on a middle click.

[`Focusable`](https://ariakit.com/reference/focusable) blocks activation by disabling `click`, `mousedown`, and `keypress` in the capture phase. A middle click does not fire `click`, it fires `auxclick`, and nothing in the library listens for it. Calling `preventDefault()` on `mousedown` does not stop it either.

When the element is truly disabled, [`Focusable`](https://ariakit.com/reference/focusable) also applies `pointer-events: none`, so the pointer never reaches it and the gesture is blocked by accident rather than by design. [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) removes `pointer-events: none` to keep the element reachable, and the middle click goes through. That makes [`Tab`](https://ariakit.com/reference/tab) affected at its default settings, because it defaults [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) to `true`.

```tsx
<Ariakit.Tab disabled render={<a href="https://ariakit.org" />}>
  Disabled tab
</Ariakit.Tab>
```

Middle clicking that tab opens the destination in a new tab in Chromium and Firefox, and navigates in place in WebKit. Every other gesture is already blocked: left click, <kbd>Enter</kbd>, and modifier clicks.

Fixes #7153

## Solution

A fourth capture guard covers `auxclick`, beside the three that already cover `click`, `mousedown`, and `keypress`:

```ts
const onAuxClickCapture = useDisableEvent(props.onAuxClickCapture, disabled);
```

It reuses the existing `useDisableEvent` helper, so it behaves exactly like its siblings: the consumer's own handler runs first, and the event is cancelled only when the element is disabled and that handler has not already prevented it. Every component built on [`Focusable`](https://ariakit.com/reference/focusable) inherits the guard, not only links.

The guard is not restricted to the middle button, which matches the unconditional form of the three guards beside it. Back and forward mouse buttons are unaffected: driving them through CDP over a disabled element showed that Chromium dispatches no `auxclick` for them at all, and history navigation still happened.

Like its siblings, the guard stops the event as well as preventing it, so on a disabled element an `auxclick` no longer reaches the handlers that would have run after the guard. This mirrors how the existing `click` guard already suppresses `onClick` on a disabled element, and it is the only behavior change here that a consumer can notice without having hit the bug.

Two gestures stay open by design, because neither fires `auxclick`: the context menu (`Open link in new tab`, `Copy link address`) and dragging the link to the address bar. Removing the destination is the only thing that closes those, and the issue puts both out of scope.

### Tests

The failing tests landed first, in [`2296063a4`](https://github.com/ariakit/ariakit/commit/2296063a4221bddefa0cb648c205a3814b358433), so CI showed the reproduction red before any fix, then a userland workaround in [`f2389b838`](https://github.com/ariakit/ariakit/commit/f2389b838) turned them green from consumer code alone. The workaround is reverted in this commit, so they now pass because of the library change.

`app/src/sandbox/focusable-tab-7153` renders the affected shapes as real links: a [`Focusable`](https://ariakit.com/reference/focusable) that is only [`disabled`](https://ariakit.com/reference/focusable#disabled), one that is also [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), and a routed tab list with an enabled and a disabled [`Tab`](https://ariakit.com/reference/tab).

The browser test delivers the gesture with `click({ button: "middle", force: true })`. Forcing is required because Playwright's actionability check treats `aria-disabled` as not enabled, so an unforced click waits for the element to become enabled and then fails with `element is not enabled`, never dispatching the gesture. Forcing skips Playwright's own checks and not the browser's hit testing, so the truly disabled link is still blocked by `pointer-events: none`, exactly as a real user would experience it.

Two enabled controls, a link and a tab, assert that the same gesture still navigates. Without them a fix that blocked middle click on every tab, enabled or not, would leave the suite green.

## Workaround

Until this ships, [`Focusable`](https://ariakit.com/reference/focusable) forwards `onAuxClickCapture` to the element, so the guard can be passed from userland:

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7153 is fixed.
<Ariakit.Tab
  disabled={disabled}
  onAuxClickCapture={
    disabled
      ? (event) => {
          event.stopPropagation();
          event.preventDefault();
        }
      : undefined
  }
  render={<a href="/billing" />}
>
  Billing
</Ariakit.Tab>
```

Assumptions and limitations:

- Attach the handler only while the element is disabled. The library's own guards check [`disabled`](https://ariakit.com/reference/focusable#disabled) before they cancel anything, but a handler passed from userland runs whenever it is attached.
- Apply it to every disabled link, not only the ones that set [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled). The default case is blocked by `pointer-events: none` rather than by design, and that protection disappears as soon as the element has to stay reachable, or the consumer sets `pointerEvents` in its own `style`, which [`Focusable`](https://ariakit.com/reference/focusable) merges over its own.
- The handler cancels `auxclick` for every non-primary button, not only the middle one, so a right-button aux click on the element stops reaching ancestor handlers too.
- It does not close the two gestures that never fire `auxclick`: the context menu (`Open link in new tab`, `Copy link address`) and dragging the link to the address bar. Withholding `href` while the element is disabled closes those too, but on a plain [`Focusable`](https://ariakit.com/reference/focusable) it also costs the element its link role and its place in the tab order, because [`Focusable`](https://ariakit.com/reference/focusable) decides native tabbability from the tag name alone.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the current test design</summary>

**Assessment**

Issue severity is moderate: `disabled` should block activation through every gesture, and middle click is the only remaining hole. Expected frequency is low per session but on a default configuration, since [`Tab`](https://ariakit.com/reference/tab) supplies `accessibleWhenDisabled` by default and middle click is common on link-shaped navigation. Supported scope is narrow: `@ariakit/react` `Focusable` and its descendants, desktop pointer only. Likely user impact is a visible but recoverable navigation to a destination the app deliberately blocked.

All three rounds circled one root decision: the observable under test is the absence of a browser-process side effect that no DOM state records. Nothing in the page changes when the gesture is correctly blocked. That forces a timed window, a timed window makes the failure direction a false green, and a window implemented as a promise created before the gesture has a fragile lifetime. The alternative would be surfacing fixture state for the test to watch, which both test skills explicitly forbid.

Implementation complexity fell rather than rose across the three rounds. `test-browser.ts` went from 91 to 94 lines and `test.ts` from 35 to 38, while a 12-line hand-rolled `page.mouse` driver, a `browserName` branch, and `page`/`context` threaded through four call sites were all removed, and two enabled controls plus two hit-target pins were added. Maintenance complexity is low: no production code, no public API, no CI configuration, all inside one sandbox directory.

**Decision**

Continue the current design. Reverting to the round 1 approach would restore an underived window, the hand-rolled driver, and the missing enabled-tab control. No machinery can be removed without losing coverage: dropping the `sameTab` race hangs the WebKit path to a test timeout, and dropping either hit-target pin or either enabled control reopens an earlier finding.

Two proposals are recorded as declined non-goals so later reviews do not re-raise them. Replacing the `Promise.race` in `expectNavigation` with a `browserName === "webkit"` branch is preferential; the current form is verified green on all three engines. Extending the hit-target pin to the truly disabled case is impossible by construction, since `pointer-events: none` is what that case asserts.

The absence budget stays at 5s. The repository's 60s new-tab waits are presence waits, where a large budget is free because the event arrives early; an absence budget is a hard wall-clock floor paid on every run. The local measurement behind the 5s value is corroborated by CI on the reproduction commit: while the bug was live, every delivery that actually occurred was caught inside the window on chrome, firefox, and safari.

</details>

<details>
<summary>Cycle 6: Simplify the workaround documentation by deletion</summary>

**Assessment**

The workaround commit changed one file and drew no finding in three consecutive gate rounds. Every finding in cycles 4 to 6 landed on the pull request description instead, and each one was a genuine inaccuracy about `Focusable` internals: a snippet that attached the guard unconditionally, an alternative that silently removed keyboard reachability, and a style-merge claim that is key-wise rather than wholesale.

The generator was not the design but added prose specificity. Each correction introduced a new claim, and each new claim became the next round's target. Cycle 6's two findings were direct descendants of cycle 3 and cycle 5 fixes: an evidentiary sentence about retries and flake status, and an accessibility caveat that had drifted away from the example it described.

**Decision**

Continue the implementation unchanged and simplify the prose by deletion rather than correction.

The duplicate `href`-withholding snippet is removed. It restated the alternative already in the issue body, and the accessibility cost stated after it applies to a plain [`Focusable`](https://ariakit.com/reference/focusable) rather than to the [`Tab`](https://ariakit.com/reference/tab) it was illustrated with, since `CompositeItem` pins a non-active item's `tabIndex` before `Focusable` computes one. One scoped sentence carries the cost instead.

The CI corroboration above is reduced to the claim that carries the argument. Attempt counts, per-project retry counts, and flake status were not load-bearing and were expensive to keep accurate.

The remaining limitations stay. Stage 4 of the issue-fix workflow requires them whenever the workaround is narrower than normal component behavior, and they are republished on the issue, where users act on them without this pull request for context. They are now a bulleted list so each limitation stays bound to its own scope.

</details>

### DIFF
--- a/.changeset/300-focusable-disabled-middle-click.md
+++ b/.changeset/300-focusable-disabled-middle-click.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link. This affects elements that stay reachable through [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), which [`Tab`](https://ariakit.com/reference/tab) enables by default.
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link, including one kept reachable by [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), which [`Tab`](https://ariakit.com/reference/tab) enables by default.
--- a/.changeset/300-focusable-disabled-middle-click.md
+++ b/.changeset/300-focusable-disabled-middle-click.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link. This affects elements that stay reachable through [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), which [`Tab`](https://ariakit.com/reference/tab) enables by default.
--- a/app/src/sandbox/focusable-tab-7153/index.react.tsx
+++ b/app/src/sandbox/focusable-tab-7153/index.react.tsx
@@ -1,14 +1,4 @@
 import * as Ariakit from "@ariakit/react";
-import type { MouseEvent } from "react";
-
-// `Focusable` disables click, mousedown, and keypress while the element is
-// disabled, but not the auxclick a middle click fires, so a disabled link still
-// opens its destination.
-// TODO: Remove once https://github.com/ariakit/ariakit/issues/7153 is fixed.
-function blockAuxClick(event: MouseEvent) {
-  event.stopPropagation();
-  event.preventDefault();
-}
 
 // The tab list has no tab panels because routed tabs render their panel from
 // the destination page.
@@ -18,17 +8,12 @@ export default function Example() {
       <Ariakit.Focusable render={<a href="?link=enabled" />}>
         Enabled link
       </Ariakit.Focusable>
-      <Ariakit.Focusable
-        disabled
-        onAuxClickCapture={blockAuxClick}
-        render={<a href="?link=disabled" />}
-      >
+      <Ariakit.Focusable disabled render={<a href="?link=disabled" />}>
         Disabled link
       </Ariakit.Focusable>
       <Ariakit.Focusable
         disabled
         accessibleWhenDisabled
-        onAuxClickCapture={blockAuxClick}
         render={<a href="?link=accessible" />}
       >
         Accessible disabled link
@@ -38,12 +23,7 @@ export default function Example() {
           <Ariakit.Tab id="overview" render={<a href="?tab=overview" />}>
             Overview
           </Ariakit.Tab>
-          <Ariakit.Tab
-            id="billing"
-            disabled
-            onAuxClickCapture={blockAuxClick}
-            render={<a href="?tab=billing" />}
-          >
+          <Ariakit.Tab id="billing" disabled render={<a href="?tab=billing" />}>
             Billing
           </Ariakit.Tab>
         </Ariakit.TabList>
--- a/app/src/sandbox/focusable-tab-7153/index.react.tsx
+++ b/app/src/sandbox/focusable-tab-7153/index.react.tsx
@@ -1,4 +1,14 @@
 import * as Ariakit from "@ariakit/react";
+import type { MouseEvent } from "react";
+
+// `Focusable` disables click, mousedown, and keypress while the element is
+// disabled, but not the auxclick a middle click fires, so a disabled link still
+// opens its destination.
+// TODO: Remove once https://github.com/ariakit/ariakit/issues/7153 is fixed.
+function blockAuxClick(event: MouseEvent) {
+  event.stopPropagation();
+  event.preventDefault();
+}
 
 // The tab list has no tab panels because routed tabs render their panel from
 // the destination page.
@@ -8,12 +18,17 @@ export default function Example() {
       <Ariakit.Focusable render={<a href="?link=enabled" />}>
         Enabled link
       </Ariakit.Focusable>
-      <Ariakit.Focusable disabled render={<a href="?link=disabled" />}>
+      <Ariakit.Focusable
+        disabled
+        onAuxClickCapture={blockAuxClick}
+        render={<a href="?link=disabled" />}
+      >
         Disabled link
       </Ariakit.Focusable>
       <Ariakit.Focusable
         disabled
         accessibleWhenDisabled
+        onAuxClickCapture={blockAuxClick}
         render={<a href="?link=accessible" />}
       >
         Accessible disabled link
@@ -23,7 +38,12 @@ export default function Example() {
           <Ariakit.Tab id="overview" render={<a href="?tab=overview" />}>
             Overview
           </Ariakit.Tab>
-          <Ariakit.Tab id="billing" disabled render={<a href="?tab=billing" />}>
+          <Ariakit.Tab
+            id="billing"
+            disabled
+            onAuxClickCapture={blockAuxClick}
+            render={<a href="?tab=billing" />}
+          >
             Billing
           </Ariakit.Tab>
         </Ariakit.TabList>
--- a/app/src/sandbox/focusable-tab-7153/index.react.tsx
+++ b/app/src/sandbox/focusable-tab-7153/index.react.tsx
@@ -1,0 +1,33 @@
+import * as Ariakit from "@ariakit/react";
+
+// The tab list has no tab panels because routed tabs render their panel from
+// the destination page.
+export default function Example() {
+  return (
+    <div className="flex flex-col items-start gap-4">
+      <Ariakit.Focusable render={<a href="?link=enabled" />}>
+        Enabled link
+      </Ariakit.Focusable>
+      <Ariakit.Focusable disabled render={<a href="?link=disabled" />}>
+        Disabled link
+      </Ariakit.Focusable>
+      <Ariakit.Focusable
+        disabled
+        accessibleWhenDisabled
+        render={<a href="?link=accessible" />}
+      >
+        Accessible disabled link
+      </Ariakit.Focusable>
+      <Ariakit.TabProvider defaultSelectedId="overview">
+        <Ariakit.TabList aria-label="Sections">
+          <Ariakit.Tab id="overview" render={<a href="?tab=overview" />}>
+            Overview
+          </Ariakit.Tab>
+          <Ariakit.Tab id="billing" disabled render={<a href="?tab=billing" />}>
+            Billing
+          </Ariakit.Tab>
+        </Ariakit.TabList>
+      </Ariakit.TabProvider>
+    </div>
+  );
+}
--- a/app/src/sandbox/focusable-tab-7153/test-browser.ts
+++ b/app/src/sandbox/focusable-tab-7153/test-browser.ts
@@ -1,0 +1,94 @@
+import type { Locator } from "@ariakit/test/playwright";
+import { errors, expect } from "@ariakit/test/playwright";
+import { withFramework } from "#app/test-utils/preview.ts";
+
+// Playwright's actionability check treats `aria-disabled` as not enabled and
+// refuses to click, so the gesture has to be forced. Forcing skips Playwright's
+// own checks, not the browser's hit testing, which is what blocks the truly
+// disabled link through `pointer-events: none`.
+function middleClick(locator: Locator) {
+  return locator.click({ button: "middle", force: true });
+}
+
+async function expectNavigation(locator: Locator, destination: RegExp) {
+  const page = locator.page();
+  const url = page.url();
+  // Chromium and Firefox open the destination in a new tab, while WebKit
+  // follows the middle click in the current one. Race both so neither engine
+  // needs its own branch.
+  const newTab = page
+    .context()
+    .waitForEvent("page")
+    .catch(() => null);
+  const sameTab = page
+    .waitForURL(destination)
+    .then(() => null)
+    .catch(() => null);
+  await middleClick(locator);
+  const opened = await Promise.race([newTab, sameTab]);
+  await expect(opened ?? page).toHaveURL(destination);
+  if (!opened) return;
+  await expect(page).toHaveURL(url);
+}
+
+async function expectNoNavigation(locator: Locator) {
+  const page = locator.page();
+  const url = page.url();
+  // A blocked gesture leaves nothing behind to assert on, so wait out the
+  // window in which the browser would have delivered the destination. The
+  // slowest delivery measured across the three engines was 1.4s, so this
+  // budget leaves room for a loaded runner without waiting on the clock for
+  // longer than the gesture can plausibly take.
+  const newTab = page
+    .context()
+    .waitForEvent("page", { timeout: 5_000 })
+    .catch((error) => {
+      if (!(error instanceof errors.TimeoutError)) throw error;
+      return null;
+    });
+  const [opened] = await Promise.all([newTab, middleClick(locator)]);
+  expect(opened, "the link opened in a new tab").toBeNull();
+  await expect(page).toHaveURL(url);
+}
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  test("middle click opens an enabled link", async ({ q }) => {
+    await expectNavigation(q.link("Enabled link"), /\?link=enabled$/);
+  });
+
+  test("middle click opens an enabled tab rendered as a link", async ({
+    q,
+  }) => {
+    await expectNavigation(q.tab("Overview"), /\?tab=overview$/);
+  });
+
+  // Truly disabled elements get `pointer-events: none`, so the pointer lands on
+  // the wrapper and never reaches this link. That also makes the link
+  // impossible to hover, which is why this case alone has no hit-target pin.
+  // https://github.com/ariakit/ariakit/issues/7153
+  test("middle click does not open a disabled link", async ({ q }) => {
+    await expectNoNavigation(q.link("Disabled link"));
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7153
+  test("middle click does not open an accessible disabled link", async ({
+    q,
+  }) => {
+    const link = q.link("Accessible disabled link");
+    // Forcing the click skips Playwright's hit-target check, so hover first to
+    // keep it. Without that pin, a later fixture change could move the gesture
+    // off this link and leave the test passing for the wrong reason.
+    await link.hover();
+    await expectNoNavigation(link);
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7153
+  test("middle click does not open a disabled tab rendered as a link", async ({
+    q,
+  }) => {
+    const tab = q.tab("Billing");
+    // Same hit-target pin as above.
+    await tab.hover();
+    await expectNoNavigation(tab);
+  });
+});
--- a/app/src/sandbox/focusable-tab-7153/test.ts
+++ b/app/src/sandbox/focusable-tab-7153/test.ts
@@ -1,0 +1,38 @@
+import { dispatch, q } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+// A middle click activates a link through `auxclick`, and `@ariakit/test` has
+// no middle-click utility, so dispatch the single event that carries the
+// activation. `dispatch` resolves to `false` when it was prevented.
+function middleClick(element: Element | null) {
+  const event = new MouseEvent("auxclick", {
+    bubbles: true,
+    cancelable: true,
+    composed: true,
+    detail: 1,
+    button: 1,
+  });
+  return dispatch(element, event);
+}
+
+// Only the browser test covers the truly disabled link: `dispatch` routes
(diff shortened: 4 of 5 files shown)


.changeset/300-focusable-disabled-middle-click.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link. This affects elements that stay reachable through [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), which [`Tab`](https://ariakit.com/reference/tab) enables by default.
```
diegohaz: This should be a single sentence changeset.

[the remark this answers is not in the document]
diegohaz: Validated against the current head `712f10f9` and dispositioned as invalid for this repository's changeset conventions, so no change is planned here.

A changeset is the changelog entry for the release that already contains the fix. A reader who reaches this line is on a version where the middle click is blocked, so a pre-fix workaround has no audience there. `ariakit-general-changeset` defines a simple entry as a short past-tense summary sentence describing what changed, and the three sibling entries queued for the same release (`.changeset/300-combobox-item-focus-on-hover-closed.md`, `.changeset/300-select-item-focus-on-hover-closed.md`, `.changeset/300-tooltip-disabled-anchor-hover.md`) all follow that shape.

The proposed text would also make this a multi-paragraph entry, and the same convention then requires the first paragraph to become a short title with no links, which is a larger rewrite than the suggestion shows.

Two smaller accuracy points, in case the wording is reused elsewhere: the documented workaround calls `event.stopPropagation()` as well as `event.preventDefault()`, and it must be attached only while the element is disabled, because a handler passed from userland runs whenever it is attached while the library's own guards check `disabled` first.

The workaround is already published where users look for it: the body of https://github.com/ariakit/ariakit/issues/7153, which carries the `has workaround` label, and the `## Workaround` section of this pull request.

diegohaz: Confirmed against the current head `712f10f9`: the entry is still one paragraph made of two sentences, so this is unaddressed.

`ariakit-general-changeset` specifies a short past-tense summary sentence for a simple entry, and every other entry queued for this release is a single sentence: `.changeset/300-combobox-item-focus-on-hover-closed.md`, `.changeset/300-select-item-focus-on-hover-closed.md`, and `.changeset/300-tooltip-disabled-anchor-hover.md`. This is the only two-sentence one.

Dropping the second sentence keeps every convention the entry already satisfies: past tense, a period, two representative derived components rather than an exhaustive list, and a link on every API mention.

```suggestion
Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link.
```

What that loses is the pointer to `accessibleWhenDisabled` as the reason the bug is reachable on a default `Tab`. If that nuance is worth keeping in the changelog, it fits in one sentence:

> Fixed [`Focusable`](https://ariakit.com/reference/focusable) and components built on it, such as [`Tab`](https://ariakit.com/reference/tab) and [`Button`](https://ariakit.com/reference/button), so a middle click no longer opens the destination of a [`disabled`](https://ariakit.com/reference/focusable#disabled) element rendered as a link, including one kept reachable by [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled), which [`Tab`](https://ariakit.com/reference/tab) enables by default.

I verified that every reference URL used in both variants resolves: `/reference/focusable`, `/reference/tab`, and `/reference/button` all return `200`, and the deployed `focusable` page carries both `id="disabled"` and `id="accessiblewhendisabled"`.

**Assessment**

- Severity: Low (verified, 95% confidence), because it affects changelog prose only and no shipped behavior.
- Priority: P1 (estimate, 85% confidence), because the changeset ships with this pull request and the convention is uniform across the entries already queued for the same release.
- Breakage likelihood: Likely (verified, 100% confidence), in the sense that the deviation is present right now in the released text rather than conditional on any input.
- Fix effort: Easy (verified, 95% confidence), because it is a one-line edit to a single changeset file with no code or test impact.

diegohaz (commented):
Initial review. Baseline is the merge base [`b7a53f16`](https://github.com/ariakit/ariakit/commit/b7a53f16722b2397d1bd32b0677bbdfb7f8e6608), head is [`712f10f9`](https://github.com/ariakit/ariakit/commit/712f10f9780af995366419e677454135ab1778f2), covering all 5 changed files plus the surrounding `Focusable` internals, the `@ariakit/test` dispatch path, the app Playwright configuration, the sibling changesets, and issue [#7153](https://github.com/ariakit/ariakit/issues/7153).

## What I ran

Reviewed in a dedicated worktree on the PR head.

- `pnpm tsc`: clean.
- `pnpm test app/src/sandbox/focusable-tab-7153/test.ts`: 4 passed.
- `pnpm test-react18 app/src/sandbox/focusable-tab-7153/test.ts`: 4 passed, which closes the React 18 check the issue asked for.
- Red-check reproduction: restored `focusable.tsx` from the merge base, re-ran the happy-dom file, and got exactly the two regression tests failing (`expected true to be false` on both disabled cases) while the two enabled controls stayed green, then restored the fix and confirmed a clean tree. The enabled controls are doing their job.
- Cross-checked the description's CI claims against the check runs: [`2296063a4`](https://github.com/ariakit/ariakit/commit/2296063a4221bddefa0cb648c205a3814b358433) is `failure` on `App / Test Chrome`, `App / Test Firefox`, and `App / Test Safari (2/2)`, and [`f2389b838`](https://github.com/ariakit/ariakit/commit/f2389b838) is `success` on all six device jobs. The reproduction was genuinely red on three engines before any library change.
- Verified every reference URL in the changeset: `/reference/focusable`, `/reference/tab`, and `/reference/button` return `200`, and the deployed `focusable` page carries both `id="disabled"` and `id="accessiblewhendisabled"`.

I did not run the Playwright suite locally. Standing it up needs `pnpm dev-app`, and a local run can silently reuse another worktree's server, so I relied on the CI evidence above plus the green run on the current head.

- `.changeset/300-focusable-disabled-middle-click.md` line 6, thread `PRRT_kwDOBjT0Fc6aIefS` (`coderabbitai`), asking the changeset to document the pre-fix workaround. Invalid for this repository's conventions. Replied on the thread.
- `.changeset/300-focusable-disabled-middle-click.md` line 6, thread `PRRT_kwDOBjT0Fc6aTMoa` (`diegohaz`), asking for a single-sentence changeset. Confirmed and still unaddressed at the current head. Replied on the thread with a suggestion and the assessment.

## Newly discovered finding

**No browser evidence in the description, and no stated evidence gap.**

`ariakit-general-browser-evidence` asks for at least one focused visual artifact in GitHub content that communicates a browser-visible result, normally under a `## Preview` section, and requires the gap to be stated explicitly when the artifact cannot be captured. This description communicates a browser-visible result and carries neither.

The effect is an absence, so a screenshot cannot show it. A short recording is the artifact that works here: middle click the disabled tab in `app/src/sandbox/focusable-tab-7153` before and after the guard, with the browser tab strip in frame so the new tab that used to appear is visibly gone, while the enabled link in the same fixture still opens one. The same recording doubles as the evidence for the `Tab` default that makes this reachable without any opt-in.

This is the only newly discovered finding, and it has no diff line to anchor to, which is why it is here rather than inline.

**Assessment**

- Severity: Low (estimate, 90% confidence), because it affects how the change is communicated and not what ships.
- Priority: P2 (estimate, 75% confidence), because the written explanation, the measured table in the issue, and green Playwright runs on three engines already carry the verification, and the skill positions visual evidence as a supplement to those rather than a replacement.
- Breakage likelihood: Likely (verified, 100% confidence), in the sense that the omission is present now rather than conditional on any input.
- Fix effort: Easy (estimate, 80% confidence), because the fixture already renders every affected shape, so one short recording of the existing sandbox covers it.

## Considered and not raised

- **Guard breadth.** The guard cancels `auxclick` for every non-primary button, not just the middle one, so a right-button aux click on a disabled element also stops reaching ancestor handlers. This matches the unconditional form of the three guards beside it and the description names it as intended. It does not cost the context menu, since `contextmenu` is dispatched from the mouse press in all three engines and is a separate event from `auxclick`. Declined non-goal.
- **Back and forward mouse buttons.** The description's claim rests on a Chromium CDP measurement. Firefox does dispatch DOM mouse events for buttons 3 and 4 on Windows and Linux, and Playwright cannot drive those buttons on that engine, so I could not verify it there. It cannot be a regression introduced by this PR either way: any `auxclick` for those buttons is preceded by a `mousedown` that the pre-existing `onMouseDownCapture` guard already cancels on the same element. Not fully verifiable, not a finding.
- **The 5s absence budget in `expectNoNavigation`.** Three tests times three desktop projects is a fixed wall-clock cost on every run. The cycle 3 reassessment already dispositions this with a local measurement corroborated by the red CI run, and I found no new evidence against it, so I am not reopening it.
- **The hand-built `auxclick` in `test.ts`.** I checked whether `dispatch.auxClick` would replace it. `@testing-library/dom@10.4.1`'s event map has no `auxClick` entry, so that helper does not exist and the documented `dispatch(element, event)` form is the correct call. The comment above it is accurate.
- **The truly disabled browser case.** It exercises `pointer-events: none` rather than the new guard, which the test comment says plainly. It still earns its place, because that style is a key-wise merge (`{ pointerEvents: "none", ...styleProp }`) that a consumer `style` can override, so it is the case that regresses if that merge ever changes.

## What checked out

- The library change is the minimal correct one. `disabled` rather than `trulyDisabled` is the right flag, matching the three siblings, and `onAuxClickCapture` sits after the `...props` spread like them, so the consumer handler still runs first through `useDisableEvent` and can pre-empt the guard.
- No parity gap. `useDisableEvent` appears in exactly one place in the repository, and `packages/ariakit-solid-components/src` has no `Focusable`, so there is no Solid counterpart to keep in sync.
- Nothing went stale. No JSDoc, guide, or reference text enumerates the blocked event list, so adding a fourth guard leaves no documentation behind.
- Fixture lifecycle. `tab-disabled` and `focusable-click-focus` do not coherently own this scenario, and giving `tab-disabled` link-rendered tabs would change the fixture contract its keyboard regression depends on. A new issue-number sandbox is the right step, and `focusable-tab-7153` orders the components correctly with the issue number as the suffix. Omitting `preview.json` is correct for a sandbox.
- Contributor credit. Issue [#7153](https://github.com/ariakit/ariakit/issues/7153) is authored by `diegohaz` and its only comment is from `ariakit-bot`; this pull request's comments are all from bots. No qualifying human contributor exists, so the absent `Thanks to` sentence and absent `## Acknowledgments` section are both correct.
- Changeset frontmatter. `@ariakit/react-components` plus `@ariakit/react` is right for a behavior change reaching consumers through the outer package, `300-` is the right prefix for a patch, and `Focusable`, `Tab`, and `Button` are all reachable through `@ariakit/react`.
- Performance. The perf comment reports no regression from the extra `useEvent` on a hook that runs on every focusable element, including `tab-perf`.

.changeset/300-focusable-disabled-middle-click.md
diegohaz: Fixed in [`e7c6cbc`](https://github.com/ariakit/ariakit/commit/e7c6cbcd68075bbd960abc08d4dc32225826cc10): condensed the entry to a single sentence, using the variant that keeps the `accessibleWhenDisabled` pointer rather than the one that drops the second sentence outright.

That clause is the entry's only signal for why a default `Tab` was exposed at all, which is the most useful fact in it. A plain `<Button disabled render={<a href />}>` was never affected, since neither `button.tsx` nor `command.tsx` sets `accessibleWhenDisabled`, so it is `trulyDisabled` and `pointer-events: none` already blocked the gesture.

diegohaz:
**Fixed**

- [`e7c6cbc`](https://github.com/ariakit/ariakit/commit/e7c6cbcd68075bbd960abc08d4dc32225826cc10) ([reply](https://github.com/ariakit/ariakit/pull/7154#discussion_r3809577822)): condensed the changeset to a single sentence, using the variant that keeps the `accessibleWhenDisabled` pointer, since that clause is the entry's only signal for why a default `Tab` was exposed.

Issue #7156, opened 2026-08-18
Events dispatched through dispatch(element, event) skip initEvent, so they lack getModifierState under happy-dom


diegohaz:
## Context

Found while reviewing the `@ariakit/test` change that adds non-primary mouse button support to [`click`](https://github.com/ariakit/ariakit/tree/main/packages/ariakit-test), which promotes `auxclick` from a `rightClick`-only implementation detail to a documented `click` option.

## Problem

`packages/ariakit-test/src/dispatch.ts` builds its named event functions (`dispatch.click`, `dispatch.keyDown`, and so on) by creating the event and then running the module-private `initEvent`, which applies `initMouseEvent`, `initUIEventModififiers`, and the other per-interface initializers.

The direct form, `dispatch(element, event)`, goes straight to `baseDispatch` and never runs `initEvent`. That form is public and documented:

```ts
await dispatch(q.textbox(), new Event("selectstart", { bubbles: true }));
```

happy-dom's `MouseEvent` implements no `getModifierState`, and `initUIEventModififiers` is what normally supplies it. So any event dispatched through the direct form lacks `getModifierState` under happy-dom, while every event dispatched through a named function has it.

A plain-DOM handler that calls `event.getModifierState("Shift")` on such an event throws `TypeError: event.getModifierState is not a function`. React consumers are insulated, because React's modifier-state getter falls back to `nativeEvent.shiftKey` when the native method is absent.

This is pre-existing on `main`, and not introduced by any open pull request. On `main`, `packages/ariakit-test/src/right-click.ts` builds its `auxclick` with a raw `MouseEvent` constructor and dispatches it through `dispatch(element, event)`, so a `rightClick` sequence already exposes `getModifierState` on `pointerdown`, `mousedown`, `contextmenu`, `pointerup`, and `mouseup`, but not on `auxclick`. `packages/ariakit-test/src/select.ts` has the same gap for its `selectstart` event.

## Why this can wait

No code in the repository reads `getModifierState`, and React consumers never hit the missing method. The gap has shipped since `rightClick` was added in #6574.

## Expected outcome

Events dispatched through the direct `dispatch(element, event)` form should be initialized the same way named event functions are, so `getModifierState` and the other normalized properties are present regardless of which form created the event.

Two approaches are worth comparing, and the fix should be made at the dispatch layer rather than special-cased in a caller:

[...]

PR #7157, opened 2026-08-18
Support non-primary mouse buttons in the @ariakit/test click helper

diegohaz:
## Motivation

`@ariakit/test` cannot simulate a middle click. `click` always dispatches a `click` event, whatever `button` the event init carries, so a test cannot reproduce a gesture that browsers deliver as `auxclick`.

```ts
// Dispatches `click`, so nothing distinguishes this from a primary click.
await click(link, { button: 1 });
```

#7154 needs exactly this gesture. It fixes a disabled element rendered as a link that still opens its destination on a middle click, and its happy-dom coverage cannot express the gesture today.

`rightClick` already simulates the secondary button and ends with `auxclick`, but it hard-codes button 2 and duplicates the entire `click` sequence, so the two paths can drift.

## Solution

`click` now reads `button` from the event init and simulates that mouse button through the whole interaction, following the rule browsers and [`user-event`](https://github.com/testing-library/user-event/blob/main/src/system/pointer/mouse.ts#L174) both apply: the primary button dispatches `click`, every other button dispatches `auxclick`.

```ts
await click(q.link("Ariakit"), { button: 1 });
```

A non-primary button also skips activation behavior, so labels and `option` elements are not forwarded, and the secondary button dispatches `contextmenu` while it is still held down.

Per-phase button state moves into a private `__mouse.ts`: the pointer moves onto the element with no button held, the press reports the held button in `buttons`, and the release reports none. `mouseDown` and `mouseUp` own their own phase, which also corrects `select`, and fixes a real gap: `mouseDown` is documented for press-and-hold testing, yet it reported `buttons: 0`, so a drag gated on the standard `event.buttons === 1` check could never be tested.

With that in place `rightClick` is exactly `click(element, { ...options, button: 2 })`, so it drops from 74 lines to 21 and can no longer drift from `click`.

## Browser evidence

Every expectation in the new tests was recorded from a real gesture, not assumed. Playwright drove `page.click(..., { button })` on Chromium, Firefox, and WebKit, recording `type button buttons detail`:

| Case | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| Middle click, enabled button | `auxclick 1 0 1` | `auxclick 1 0 1` | `auxclick 1 0 1` |
| Middle click, `disabled` button | `auxclick` fires | no `auxclick` | `auxclick` fires |
| Middle click on a `<label for>` | no forwarding | no forwarding | forwards to the control |
| Primary press | `buttons: 1` | `buttons: 1` | `buttons: 1` |

Two of those results shaped the implementation. Browsers keep firing `auxclick` on a disabled control even though they suppress `click` there, so it is not suppressed. And label activation is tied to the `click` event by the HTML specification, so a middle click does not activate a label; WebKit is the outlier and this matches Chromium and Firefox.

The primary press reporting `buttons: 1` is the third: the helper reported `0`, which no browser does. That is corrected as part of the same per-phase derivation and is described in the changeset.

## A note for #7154

Neither happy-dom nor a synthetic event in a real browser navigates on `auxclick`. happy-dom gates anchor navigation on `event.type === "click"`, and middle-click "open in new tab" is a user-agent gesture behavior rather than a DOM activation behavior. So a happy-dom test shaped as "middle click the disabled link, assert it did not navigate" would pass whether or not the fix is present.

Integration coverage for #7154 should assert something the fix actually changes, such as an `auxclick` listener on an ancestor not firing, or `defaultPrevented` observed from the test's own listener. Note that `click` discards the `auxclick` dispatch's return value, so the test needs its own listener. Real navigation suppression belongs in a Playwright test driving a genuine middle click.

## Follow-up

Events dispatched through the direct `dispatch(element, event)` form skip the initializer that named event functions run, so under happy-dom they lack `getModifierState`. This is pre-existing on `main`, predates this change, and also affects `select`'s `selectstart` event. Registered as #7156.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the current direction</summary>

**Assessment**

Issue severity is moderate. Without this, #7154 cannot get happy-dom coverage at all, and `mouseDown` cannot express a press-and-hold drag. Supported scope is narrow: `@ariakit/test`, a test utility whose blast radius is test authors rather than end users. Likely user impact is confined to test behavior. Expected frequency is low for the non-primary path, which is opt-in, but the `buttons` correction touches every `click`.

Three review rounds produced findings sharing one root cause. The change introduces a distinction between raw caller options and per-phase normalized options, and both are `PointerEventInit`, so TypeScript cannot flag a call site that was missed. Each round caught a different subset of the same enumerable list by eye. A mechanical sweep of every option-passing call site confirmed the list is now closed, with `select.ts`'s final `click` the last member. Rounds 2 and 3 each returned zero correctness defects and zero objections to prior dispositions, and severity decayed every round.

Complexity did not grow with the cycles. Production source is +126/-75 across 6 files, and `__mouse.ts` is largely a relocation of helpers that already existed inside `right-click.ts`, which loses 55 lines and drops from 74 lines to 21. The genuinely new code is a button-to-bitmask table and a small accessor. No public API is added, no new state, no platform policy.

**Decision**

Continue the current direction. Do not simplify, remove `__mouse.ts`, revert, or narrow the contract.

The two cheaper alternatives are worse. Confining the change to `click.ts` makes `mouseDown`'s own documentation false, since it is described as one step of a full `click` yet would report different `buttons`. Skipping the primary-button correction leaves `buttons` accurate for every button except the one used almost all the time, and leaves press-and-hold drags untestable.

The `getModifierState` gap is registered as #7156 rather than fixed here: it is pre-existing on `main`, has no consumer in the repository, is invisible to React consumers, and fixing it properly means restructuring the dispatch layer.

</details>


packages/ariakit-test/src/__mouse.ts
```diff
@@ -0,0 +1,61 @@
...
+): PointerEventInit {
+  return { ...options, button: getMouseButton(options), buttons: 0 };
+}
+
+// `@testing-library/dom` has no `auxclick` in its event map, so `dispatch` can't
+// build this one by name.
+export function dispatchAuxClick(element: Element, options?: MouseEventInit) {
+  const { defaultView } = element.ownerDocument;
+  const MouseEventConstructor = defaultView?.MouseEvent ?? MouseEvent;
+  const event = new MouseEventConstructor("auxclick", {
```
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because it would make `auxclick` the only event in the sequence with a different interface. `@testing-library/dom` declares `EventType: "MouseEvent"` for both `click` and `contextMenu` in its event map, so `dispatch.click` and `dispatch.contextMenu` construct `MouseEvent`. Switching only `auxclick` to `PointerEvent` would leave a right-click firing `contextmenu` as a `MouseEvent` and `auxclick` as a `PointerEvent`.

`PointerEvent` extends `MouseEvent`, so every property these helpers and their tests read (`button`, `buttons`, `detail`, and the modifier keys) is unaffected either way, and `instanceof MouseEvent` still holds.

This construction is also not new here. It is relocated verbatim from `right-click.ts`, which has used `MouseEvent` since `rightClick` shipped in #6574.

Aligning the event interfaces is a dispatch-layer concern rather than an `auxclick` one, and it overlaps the initialization gap tracked in #7156: events built by hand and sent through `dispatch(element, event)` skip `initEvent`, so they also miss `getModifierState`. Both are better fixed together there.

diegohaz:
**Declined**

- [`__mouse.ts:53`](https://github.com/ariakit/ariakit/pull/7157#discussion_r3804884717) ([reply](https://github.com/ariakit/ariakit/pull/7157#discussion_r3804884717)): building `auxclick` with `PointerEvent` instead of `MouseEvent`. `click` and `contextmenu` are `MouseEvent` in this package, so changing only `auxclick` would make it the odd one out, and `PointerEvent` extends `MouseEvent`, so no property these helpers read changes. Aligning event interfaces belongs at the dispatch layer alongside #7156.

diegohaz (commented):
## Review summary

Initial review. Baseline is the merge base `b7a53f167` to head [`f8cda48`](https://github.com/ariakit/ariakit/commit/f8cda4878f6bf6a8712c109e1b7d36d9b124366e), covering all 10 changed files plus the surrounding `packages/ariakit-test/src` interaction layer (`dispatch.ts`, `hover.ts`, `press.ts`, `tap.ts`, `select.ts`, `index.ts`).

This is a well-evidenced change. The `rightClick` collapse into `click` is the strongest part: it removes a duplicated 74-line sequence that could drift, and the three pre-existing `right-click.test.ts` cases act as real regression proof that the collapse is behavior-preserving. One finding published inline.

### Published inline

- `packages/ariakit-test/src/__mouse.ts` L33-L45: `getPressOptions` and `getReleaseOptions` overwrite `buttons` unconditionally, so a caller-supplied `buttons` is silently discarded. Verified as a regression against `main`, with a validated two-line fix.

### Verified as correct

- **`rightClick` equivalence.** Read both versions side by side and confirmed the sequence, the `settle()` placement between `contextmenu` and `mouseUp`, the post-`mouseDown` visibility retarget loop, and the trailing `sleep()` are all preserved. `settle()` becoming conditional on `tap` is a no-op for `rightClick`, which never passes `tap`. All three `right-click.test.ts` cases pass unchanged, including the prevented-`pointerdown` and `pointer-events: none` retargeting paths.
- **No public API added.** `__mouse.ts` is not re-exported from `index.ts`, so the four helpers stay private.
- **No in-repo blast radius for the `buttons` correction.** Grepped `packages` and `app` for `.buttons` reads in library and app source and found none, so the change from `0` to `1` on every primary `pointerdown`/`mousedown` cannot alter component behavior. The green `Main / Test`, `Main / Test React 18`, and browser suites agree.
- **`dispatchAuxClick` is genuinely necessary.** Confirmed independently that `auxClick` is absent from `@testing-library/dom`'s `fireEvent` map while `contextMenu` is present, so `dispatch` cannot build this event by name. The realm-safe `element.ownerDocument.defaultView?.MouseEvent ?? MouseEvent` construction is also correct.
- **Option normalization sweep is closed.** Independently enumerated every mouse and pointer `dispatch.*` call site in the package. Inside `click`, `select`, `mouseDown`, `mouseUp`, and `rightClick`, every one now receives phase-normalized options. `hover.ts` and `press.ts` are the only remaining raw-option sites, and both are correct: public `hover` should not force `buttons: 0` (that would make a drag-over untestable), and `press`'s clicks are keyboard-synthesized.
- **Changeset metadata.** `@ariakit/test` is `0.7.4`, so the `200-` minor ordering prefix with `patch` frontmatter follows the v0 rule, and no other package needs listing since no published package re-exports these helpers. Spot-checked each behavioral claim in the description against the implementation and the new tests; all hold.
- **The readme edits are generator output.** `packages/ariakit-test/readme.md` is produced from JSDoc by `packages/ariakit-scripts/src/docs.ts`, and the `Docs / Generate docs` job is green, so the prose is in sync with the source.

### Other candidates, with dispositions

**`rightClick(null)` now reports `Unable to click on null element`.** Confirmed by probe: `main` reports `Unable to rightClick on null element`. Disposition: Declined non-goal. `tap` already delegates to `click` and already inherits the same generic message on `main`, so this matches an established precedent for thin delegating wrappers rather than introducing a new inconsistency. Restoring it would mean re-adding a `wrapAsync` wrapper to keep the rejection asynchronous, which is a poor trade for a diagnostic string.

- Severity: Low (verified), because it only affects the wording of an error a test author already has to debug from a stack trace.
- Priority: P2 (estimate, 75% confidence), because `tap`'s precedent makes the current behavior defensible as-is.
- Breakage likelihood: Possible (estimate, 70% confidence), because it needs a query helper to return `null` and be passed straight to `rightClick`.
- Fix effort: Easy (estimate, 85% confidence), because it is a small wrapper around the delegating call.

**`auxclick` has no `getModifierState` under happy-dom.** Confirmed by probe: within one `click(button, { button: 1, shiftKey: true })`, `mouseup` exposes `getModifierState` returning `true`, while `auxclick` throws `TypeError: mouse.getModifierState is not a function`, because the direct `dispatch(element, event)` form skips `initEvent`. Disposition: Registered follow-up, already recorded by the author as [#7156](https://github.com/ariakit/ariakit/issues/7156). I verified that issue is open and that its problem statement covers this exact path. It is pre-existing on `main` for `rightClick`, and this PR widens its reach rather than creating it. No Ariakit component listens to `auxclick`, so the exposure is limited to external `@ariakit/test` consumers.

- Severity: Medium (verified), because a plain-DOM `auxclick` handler calling `getModifierState` throws rather than misreporting.
- Priority: P2 (estimate, 80% confidence), because the correct fix restructures the dispatch layer and belongs with the rest of #7156.
- Breakage likelihood: Unlikely (estimate, 75% confidence), because React consumers are insulated by React's `shiftKey` fallback and no in-repo component listens to `auxclick`.
- Fix effort: Moderate (estimate, 70% confidence), because it means routing hand-built events through the named-function initializer.

**`select` does not pin `button` to `0`.** `select` normalizes its hover and final click phases but still lets a caller-supplied `button` flow into `mouseDown`, so `select(text, element, { button: 2 })` would emit `pointerdown`/`mousedown` with `button: 2` and then a `click`, a sequence no browser produces for a secondary drag. Disposition: Declined non-goal. The same leak exists on `main`, `select` models a primary-button text drag by definition, and no caller passes `button` to it. Pinning `button: 0` throughout `select` would be a reasonable future hardening, not a defect this PR introduces.

**Buttons `6` and above map to `buttons: 0`.** Disposition: Invalid or obsolete. Measured identical on `main` and on this head (`pointerdown 9 0`), and the Pointer Events bitmask table defines no bit beyond the pen eraser, so no browser emits those values.

**Constructing `auxclick` as a `PointerEvent`.** Not re-adjudicated. The only review thread on this PR ([`PRRT_kwDOBjT0Fc6aI5lW`](https://github.com/ariakit/ariakit/pull/7157#discussion_r3804869464)) raised it, the author answered it, and the reporter withdrew it and resolved the thread. I read it in full and reached the same conclusion independently, so it needs no new reply.

### Commands and experiments

- `pnpm test --run packages/ariakit-test` on the PR head: 15 files, 88 tests, all passing.
- Wrote a temporary probe suite to measure `buttons` propagation, the null-element message, `getModifierState` availability on `auxclick`, out-of-range button numbers, and `auxclick` on an anchor.
- Ran that probe against the PR head, then checked out `b7a53f167`'s `packages/ariakit-test/src` and re-ran it to get the `main` baseline, then restored the head and confirmed a clean tree.
- Prototyped the inline suggestion locally and re-ran the full package suite (93 tests including probes) to confirm it is behavior-preserving.
- Verified `auxClick` is absent from `@testing-library/dom`'s `fireEvent` map by enumerating its keys.
- Removed every experiment; the review worktree ends clean against the PR head.

### Not fully validated

- The recorded cross-browser behaviors (Chromium and WebKit firing `auxclick` on disabled controls while Firefox does not, WebKit forwarding a middle click on a `<label>`, and `focus` ordering for a non-primary press) are taken from the PR's Playwright evidence. I did not re-record them, and the happy-dom suite cannot distinguish them.
- I did not run the full repository Vitest suite or the Playwright browser suites locally. Those rest on the green CI run for this head.

packages/ariakit-test/src/__mouse.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Returns the event properties for the release, once the button is no longer
+ * held down.
+ */
+export function getReleaseOptions(
+  options?: PointerEventInit,
+): PointerEventInit {
+  return { ...options, button: getMouseButton(options), buttons: 0 };
+}
```
diegohaz, line 74: `getPressOptions` and `getReleaseOptions` overwrite `buttons` unconditionally, so a caller-supplied `buttons` is silently discarded. `PointerEventInit` declares `buttons`, so the call compiles and appears to work, but the value never reaches the event. This is a capability that exists on `main` today.

I measured both sides with the same probe. On `main` (`b7a53f167`):

```
mouseDown(button, { button: 2, buttons: 3 })
-> ["pointerdown 2 3", "mousedown 2 3"]
```

On this head (`f8cda487`):

```
mouseDown(button, { button: 2, buttons: 3 })
-> ["pointerdown 2 2", "mousedown 2 2"]
```

The real-world case is a chorded gesture, which is the one thing `buttons` exists to express and `button` cannot. A test author reproducing "the user presses the right button while the left one is already held" writes `buttons: 3`, and a test author reproducing "the user releases the middle button but keeps the left one held" writes `mouseUp(element, { buttons: 1 })`. Both are expressible on `main` and become inexpressible here, with no type error and no runtime signal.

Deriving `buttons` per phase is the right default, so this only needs the derived value to be a fallback rather than an override:

```suggestion
export function getPressOptions(options?: PointerEventInit): PointerEventInit {
  const button = getMouseButton(options);
  return {
    ...options,
    button,
    buttons: options?.buttons ?? buttonsByButton[button] ?? 0,
  };
}

/**
 * Returns the event properties for the release, once the button is no longer
 * held down.
 */
export function getReleaseOptions(
  options?: PointerEventInit,
): PointerEventInit {
  return {
    ...options,
    button: getMouseButton(options),
    buttons: options?.buttons ?? 0,
  };
}
```

I applied exactly this locally and ran `pnpm test --run packages/ariakit-test`: 88 existing tests stay green, including the six-button matrix in `mouse-down.test.ts` and the three `rightClick` sequence tests, and the chorded press goes back to reporting `3`. Nothing regresses, because every existing call site leaves `buttons` unset, so the fallback always resolves to the derived value.

If the intent is instead that these helpers own `buttons` outright, the honest form is to remove it from the accepted type (for example `Omit<PointerEventInit, "buttons">` on the public helpers) so the compiler rejects it rather than the runtime ignoring it. Either resolution is fine; silently dropping a declared option is the part worth avoiding.

**Assessment**

- Severity: Low (verified), because it only removes an escape hatch for chorded gestures and cannot affect a single-button press, which every in-repo caller uses.
- Priority: P2 (estimate, 80% confidence), because this PR introduces the silent drop and the fix is one token per function, but nothing in the repository depends on it.
- Breakage likelihood: Rare (verified), because a sweep of `packages` and `app` found no caller passing `buttons` into these helpers, so only external `@ariakit/test` consumers writing chorded-gesture tests are affected.
- Fix effort: Easy (verified), because the two-line change above is validated locally against the full package suite.

diegohaz: Fixed in [`570850d`](https://github.com/ariakit/ariakit/commit/570850d8c86a60243b6ec1e1c10b852ab1f9f9c1): the derived value is now a default rather than an override, so an explicit `buttons` always wins. I reproduced the drop first and got the same numbers you did.

I applied the same fallback to `getHoverOptions`, which dropped `buttons` the same way. A chorded gesture holds a button while the pointer moves onto the element, so honoring it there restores `click(element, { buttons })` to what it did before this PR on all three phases rather than two. `button` stays forced to `0` in that helper, since one init feeds both the pointer and the mouse events and a move event reports no button change.

Two red-checked regressions cover it: `mouseDown and mouseUp keep an explicit buttons value` fails with `pointerdown 2 2` against `2 3` without the fix, and the hover test fails with `pointerover 0 0` against `0 1` when only the `getHoverOptions` line is reverted.

The changeset now documents the escape hatch, scoped to `mouseDown` and `mouseUp`. `click` cannot express a chorded gesture, because one `options` object spans press and release, so the release would report the just-released button as still held. The `mouseDown` and `mouseUp` reference docs mention it too, since stating the guarantee as absolute there would have contradicted the changeset.

diegohaz:
**Fixed**

- [`570850d`](https://github.com/ariakit/ariakit/commit/570850d8c86a60243b6ec1e1c10b852ab1f9f9c1) ([reply](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809123335)): the per-phase helpers dropped a caller-supplied `buttons`, making a chorded gesture inexpressible. The derived value is now a default rather than an override, in all three phases, including `getHoverOptions`, which dropped it the same way. Two red-checked regressions cover the press/release and hover paths, the changeset documents the escape hatch scoped to `mouseDown` and `mouseUp`, and both reference docs were updated so they no longer state the guarantee as absolute.

packages/ariakit-test/src/__mouse.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Returns the event properties for moving the pointer onto the element, before
+ * any button is pressed. One init feeds both the pointer and the mouse events,
+ * so `button` reports the mouse-event value rather than the `-1` a real
+ * `pointermove` uses when no button changed state.
+ * https://w3c.github.io/pointerevents/#the-button-property
+ */
+export function getHoverOptions(options?: PointerEventInit): PointerEventInit {
+  return { ...options, button: 0, buttons: options?.buttons ?? 0 };
```
diegohaz: Your reasoning for extending the fallback here holds for `click`, and I verified it: on `b7a53f167`, `click(button, { button: 1, buttons: 1 })` already reported `buttons: 1` on all seven events, so honoring it across all three phases restores that exactly rather than inventing anything. That part is not a finding.

`rightClick` is the case the restoration argument does not cover, because it did not delegate before. On `b7a53f167` it hard-coded the press value, so a caller-supplied `buttons` was ignored:

```
main:  rightClick(button, { buttons: 1 })
       -> pointerdown 2 2, mousedown 2 2, contextmenu 2 2, pointerup 2 0, ...

head:  rightClick(button, { buttons: 1 })
       -> pointerdown 2 1, mousedown 2 1, contextmenu 2 1, pointerup 2 1, ...
```

`pointerdown 2 1` is not a state a pointing device can be in. `button: 2` says the secondary button is the one transitioning to pressed, so bit 1 must be set in `buttons`, and `1` says only the primary is held. So `rightClick` moves from ignoring the caller's value to emitting an impossible one. `select` behaves the same way, reporting the caller's `buttons` verbatim on the press even though it never accepted a chord either.

I am not asking you to special-case `rightClick`, which would put back a branch this PR correctly removed. The cheap resolution is to pin the behavior in the test that already exercises it. `click reports the held buttons on the hover phase` is currently the only coverage of `click` with an explicit `buttons`, and it asserts the one phase where the value is coherent, so nothing records what the press does with the same call. Extending that test to the full sequence would put the acknowledged limitation where a reader meets it:

```ts
const all = recordEvents(button);
await click(button, { button: 1, buttons: 1 });

// One init spans every phase, so an explicit `buttons` is reported verbatim
// and can't describe a chord: the press can't add the button it is pressing.
// Use `mouseDown` and `mouseUp` for that.
expect(all).toEqual([
  "pointerdown 1 1 0",
  "mousedown 1 1 1",
  "focus - - 0",
  "pointerup 1 1 0",
  "mouseup 1 1 1",
  "auxclick 1 1 1",
]);
```

I measured those six values on this head, so the assertion is accurate as written.

The alternative, if you would rather the multi-phase helpers keep deriving, is to strip `buttons` once at the `click` and `select` boundary and let `mouseDown` and `mouseUp` stay the documented chord path, which is what the changeset already says. I prototyped that locally: `click(button, { button: 1, buttons: 1 })` becomes `pointerdown 1 4`, `rightClick(button, { buttons: 1 })` becomes `pointerdown 2 2`, and the only test that fails is the hover assertion added in this commit. All 92 others stay green, including `mouseDown and mouseUp keep an explicit buttons value`. I mention it only as the other coherent end state; it does trade away the pre-PR `click` behavior you deliberately restored, so the test change above is the smaller move.

**Assessment**

- Severity: Low (verified), because it only affects a call that no in-repo test makes and that neither `rightClick` nor `click` documents as accepting `buttons`.
- Priority: P2 (estimate, 75% confidence), because the `click` half is pre-existing and only the `rightClick` half is new, so a test that records the limitation is enough.
- Breakage likelihood: Rare (verified), because a sweep of `packages` and `app` found no caller passing `buttons` into any of these helpers.
- Fix effort: Easy (verified), because the assertion above is measured and drops into an existing test.

diegohaz (commented):
## Re-review summary

Re-review. Baseline is the head recorded in my previous review marker, [`f8cda48`](https://github.com/ariakit/ariakit/commit/f8cda4878f6bf6a8712c109e1b7d36d9b124366e), so the incremental scope is [`570850d`](https://github.com/ariakit/ariakit/commit/570850d8c86a60243b6ec1e1c10b852ab1f9f9c1) alone: `__mouse.ts`, `mouse-down.ts`, `mouse-up.ts`, `click.test.ts`, `mouse-down.test.ts`, the readme, and the changeset. I also re-checked the surrounding `click`, `select`, and `rightClick` paths that the change newly exposes.

Newly discovered findings were reviewed from the commits since the last reviewed SHA plus the surrounding behavior they expose. One finding published inline.

### Prior finding: resolved

The `buttons` drop I reported on [`__mouse.ts:33-62`](https://github.com/ariakit/ariakit/pull/7157#discussion_r3808930503) is fixed in [`570850d`](https://github.com/ariakit/ariakit/commit/570850d8c86a60243b6ec1e1c10b852ab1f9f9c1) and the thread is resolved, so I did not reopen it. Re-verified against this head: `mouseDown(button, { button: 2, buttons: 3 })` now reports `pointerdown 2 3` and `mouseup(button, { button: 2, buttons: 1 })` reports `pointerup 2 1`. The new `mouseDown and mouseUp keep an explicit buttons value` test pins exactly that, and both values are what a browser reports for pressing the secondary button with the primary held and then releasing it with the primary still down. The full package suite is green at 89 tests.

### Published inline

- `packages/ariakit-test/src/__mouse.ts` L21-L34: `rightClick` and `select` now forward a caller-supplied `buttons` verbatim into the press phase, which can emit a state no pointing device can be in. Includes a measured, drop-in assertion to pin the limitation in the test that already exercises it.

### Validated and withdrawn

**Extending the fallback to `getHoverOptions` does not create the `click` incoherence.** My working hypothesis on opening this re-review was that honoring `buttons` across all three phases made `click(element, { button, buttons })` newly incoherent, since one init cannot carry the three different values a chord needs. I measured `b7a53f167` before writing it up, and the reply on the prior thread is correct: `main` already reported the caller's `buttons` verbatim on every event of that call.

```
main:  click(button, { button: 1, buttons: 1 })
       -> pointerover 1 1, pointermove 1 1, pointerdown 1 1, mousedown 1 1,
          pointerup 1 1, mouseup 1 1, click 1 1

head:  click(button, { button: 1, buttons: 1 })
       -> pointerover 0 1, pointermove 0 1, pointerdown 1 1, mousedown 1 1,
          pointerup 1 1, mouseup 1 1, auxclick 1 1
```

The `buttons` column is identical, so this restores pre-PR behavior rather than introducing anything. The two `button` column differences are both intended and documented: `getHoverOptions` pins the move phase to `0`, and the auxiliary button now ends in `auxclick`. Disposition: Invalid or obsolete for `click`. The residual delta is `rightClick`, which never delegated before, and that is the published finding.

**The new `getHoverOptions` JSDoc claim.** Checked the `button: -1` rule against the linked Pointer Events section. The comment is accurate: a real `pointermove` reports `-1` when no button changed state while `mousemove` reports `0`, and one init feeding both events forces the choice. Documenting it is an improvement over the silent behavior on `main`. Not a finding.

**`getHoverOptions` pinning `button` to `0` while honoring `buttons`.** Deliberate and now explained in the source. Disposition: Declined non-goal.

### Carried forward from the previous review

- `auxclick` still has no `getModifierState` under happy-dom. Unchanged by this commit, still tracked in [#7156](https://github.com/ariakit/ariakit/issues/7156). Disposition: Registered follow-up, unchanged.
- `rightClick(null)` still reports `Unable to click on null element`. Unchanged, and still matches the `tap` precedent. Disposition: Declined non-goal, unchanged.
- `select` not pinning `button` to `0`. Unchanged. Disposition: Declined non-goal, unchanged. Its `buttons` behavior is folded into the published finding.

### Documentation and changeset

The changeset now scopes the escape hatch correctly: `Pass buttons to mouseDown and mouseUp yourself to describe a chorded gesture`. That matches where a chord can actually be expressed and is consistent with the `mouseDown` and `mouseUp` JSDoc, which no longer state the derived value as an absolute guarantee. The readme diff is the matching generator output for those two JSDoc edits, and nothing in `click`'s section changed, which is correct because `click`'s contract for `buttons` is unchanged from `main`. The `recordPressEvents` extraction in `mouse-down.test.ts` is a clean deduplication of the listener setup now shared by two tests.

### Commands and experiments

- `pnpm test --run packages/ariakit-test` on this head: 15 files, 89 tests, all passing.
- Re-measured `click`, `rightClick`, and `select` with an explicit `buttons` on this head, then checked out `b7a53f167`'s `packages/ariakit-test/src` and re-measured to get the `main` baseline, then restored this head.
- Prototyped the alternative resolution (stripping `buttons` at the `click` and `select` boundary) and ran the full package suite: only the hover assertion added in this commit fails, with 92 others green.
- Verified the exact six-event assertion offered in the inline comment by running it against this head before publishing it.
- Removed every experiment; the review worktree ends clean at `570850d`.

### Not fully validated

- The cross-browser recordings in the PR description are still taken on trust; I did not re-record them, and happy-dom cannot distinguish them.
- I did not run the full repository Vitest suite or the Playwright browser suites locally for this commit. Those rest on CI, which is complete and green for `570850d` at 44 successful check runs with `Build styles` skipped.

packages/ariakit-test/src/__mouse.ts
diegohaz: @/Users/diegohaz/.claude/jobs/b798077b/tmp/reply.md

diegohaz:
**Fixed**

- [`5a1a6d9`](https://github.com/ariakit/ariakit/commit/5a1a6d90ae77c1018acca141474cfb66d469fcb8) ([reply](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809556292)): `click`, `tap`, `rightClick`, and `select` now strip `buttons` at the boundary and let each phase derive its own, so `rightClick(button, { buttons: 1 })` no longer emits the impossible `pointerdown 2 1`. `mouseDown` and `mouseUp` stay the chord path. Coverage lands in the same commit and both strips are red-checked.

packages/ariakit-test/src/__mouse.ts
```diff
@@ -0,0 +1,89 @@
...
+    buttons: options?.buttons ?? 0,
+  };
+}
+
+// `@testing-library/dom` has no `auxclick` in its event map, so `dispatch` can't
+// build this one by name.
+export function dispatchAuxClick(element: Element, options?: MouseEventInit) {
+  const { defaultView } = element.ownerDocument;
+  const MouseEventConstructor = defaultView?.MouseEvent ?? MouseEvent;
+  const event = new MouseEventConstructor("auxclick", {
```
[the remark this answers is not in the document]
diegohaz: Validated independently and not carrying this as a finding against this pull request. The interface observation is accurate, but it is not specific to `auxclick`, and it is not introduced here.

I measured which events in this package are `PointerEvent` instances, passing pointer-specific properties through:

```
click(button, { button: 1, pointerType: "pen", pointerId: 42 })
  pointerdown  instanceofPointerEvent=true   pointerType=pen        pointerId=42
  mousedown    instanceofPointerEvent=false  pointerType=undefined  pointerId=undefined
  pointerup    instanceofPointerEvent=true   pointerType=pen        pointerId=42
  auxclick     instanceofPointerEvent=false  pointerType=undefined  pointerId=undefined

click(button, { pointerType: "pen", pointerId: 42 })
  click        instanceofPointerEvent=false  pointerType=undefined  pointerId=undefined

rightClick(button, { pointerType: "pen", pointerId: 42 })
  contextmenu  instanceofPointerEvent=false  pointerType=undefined  pointerId=undefined
```

`auxclick` behaves exactly like `click` and `contextmenu`, which both predate this pull request. All three are built as `MouseEvent` because `@testing-library/dom` declares them that way in its event map, and `dispatch` builds named events from that map. So "the `PointerEventInit` properties are discarded on the final event" is equally true of `click` on `main`; the helper has never carried `pointerType` or `pointerId` onto the activation event, whatever the button.

Changing only `auxclick` would make it the single event in the sequence with a different interface from its siblings, which is the same conclusion reached on [this thread](https://github.com/ariakit/ariakit/pull/7157#discussion_r3804869464), where the finding was withdrawn and resolved. Aligning the activation events with the `PointerEvent` interface is a dispatch-layer change that belongs with the initialization gap tracked in [#7156](https://github.com/ariakit/ariakit/issues/7156), not with the `auxclick` call site.

Disposition: invalid or obsolete for this pull request, as a duplicate of an already-dispositioned finding whose underlying cause is registered separately.

packages/ariakit-test/src/mouse-down.ts
```diff
@@ -66,16 +70,22 @@ export function mouseDown(element: Element | null, options?: PointerEventInit) {
     if (!isVisible(element)) return;
 
     const { disabled } = element as HTMLButtonElement;
+    const pressOptions = getPressOptions(options);
 
-    const pointerDefaultAllowed = await dispatch.pointerDown(element, options);
+    const pointerDefaultAllowed = await dispatch.pointerDown(
+      element,
+      pressOptions,
+    );
```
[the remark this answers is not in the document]
diegohaz: Confirmed technically, and registered as a follow-up rather than current-PR work, because the behavior is unchanged from `main`.

The specification claim checks out. From [Chorded button interactions](https://w3c.github.io/pointerevents/#chorded-button-interactions):

> Pointer Events do not fire overlapping `pointerdown` and `pointerup` events for chorded button presses (depressing an additional button while another button on the pointer device is already depressed).

So for `mouseDown(element, { button: 2, buttons: 3 })` the primary button is already held, the press does not cross the zero-buttons boundary, and a real device reports the change through `pointermove` while still firing the compatibility `mousedown`. The mirrored point about `mouseUp` with a nonzero remaining `buttons` holds for the same reason.

What decides the disposition is that this is not new here. I ran the same probe against `b7a53f167` with that commit's `packages/ariakit-test/src` checked out, and against this head, and the two are byte-identical:

```
main b7a53f167          head 5a1a6d90a
pointerdown 2 3         pointerdown 2 3
mousedown   2 3         mousedown   2 3
pointerup   2 1         pointerup   2 1
mouseup     2 1         mouseup     2 1
```

`mouseDown` and `mouseUp` have always forwarded a caller-supplied `buttons` to `dispatch.pointerDown` and `dispatch.pointerUp`. This pull request documents that input; it does not change what it does. The one thing attributable to the pull request is the wording, since the changeset presents `buttons` as a way "to describe a chorded gesture" without noting that the pointer event types are not the ones a chord produces.

Registered as [#7161](https://github.com/ariakit/ariakit/issues/7161), which records the specification quote, the measured sequences, the pre-existing verification, the `dispatch.pointerMove` workaround that composes the accurate sequence today, and the open question of whether the right outcome is to add the branch or to narrow the helpers' documented contract instead.

I am not asking for the branch in this pull request. Adding it would put a conditional pointer-event path into two helpers whose documented contract is that they fire `pointerdown` and `mousedown`, and `pointerup` and `mouseup`, for a capability with no caller in this repository.

**Assessment**

- Severity: Low (verified), because the compatibility `mousedown` and `mouseup` still fire with correct `button` and `buttons`, so only assertions on pointer event types during a chord are affected.
- Priority: P2 (verified), because the behavior is identical on `main` and independent of this pull request's goal.
- Breakage likelihood: Rare (verified), because no caller in `packages` or `app` passes `buttons` to these helpers and no library source reads `event.buttons`.
- Fix effort: Moderate (estimate, 70% confidence), because the branch is small but it changes two documented contracts and invalidates an existing test.

diegohaz (commented):
## Re-review summary

Re-review. Baseline is the head recorded in my previous marker, [`570850d`](https://github.com/ariakit/ariakit/commit/570850d8c86a60243b6ec1e1c10b852ab1f9f9c1), so the incremental scope is [`5a1a6d9`](https://github.com/ariakit/ariakit/commit/5a1a6d90ae77c1018acca141474cfb66d469fcb8) alone. I also re-checked `tap`, `rightClick`, `select`, and the public `hover` that the change routes around.

**No new findings of my own this round.** Two unresolved threads opened by another reviewer since my last pass were validated independently; one is invalid for this pull request and one is confirmed but pre-existing and now registered. Details below.

### Prior finding: resolved

The `rightClick` and `select` issue I reported on [`__mouse.ts:21-34`](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809178540) is fixed in [`5a1a6d9`](https://github.com/ariakit/ariakit/commit/5a1a6d90ae77c1018acca141474cfb66d469fcb8) and the thread is resolved. The implementation is the alternative resolution I prototyped, and the two sequences it now pins are the exact values I measured on that prototype: `pointerdown 1 4 0` for `click(paste, { button: 1, buttons: 1 })` and `pointerdown 2 2 0` for `rightClick(menu, { buttons: 1 })`.

`omitButtons` is the right shape for it. I verified the parts a spread-and-delete helper can get wrong:

- It does not mutate the caller's object. After `click(button, shared)` with `shared = { button: 1, buttons: 1, shiftKey: true }`, `shared` is unchanged.
- Modifiers survive the copy: `pointerdown shift=true ctrl=true alt=true` and `auxclick shift=true ctrl=true alt=true`.
- `omitButtons(undefined)` returns `{}` rather than `undefined`, which changes nothing downstream because every phase helper already normalized an absent init to the same values.
- `tap` inherits the derivation through `click`: `tap(button, { button: 1, buttons: 1 })` gives `pointerdown 1 4`, and no test covers `tap` directly, which is fine since it is a one-line delegation.
- The single-step chord escape hatch still works: `pointerdown 2 3` then `pointerup 2 1`.
- Public `hover` still honors `buttons`, so `hover(button, { buttons: 1 })` gives `pointerover 0 1`. That is the correct home for a mid-drag move, and it is what the previous commit was reaching for through `click`.

`select` coverage landing in `mouse-down.test.ts` is consistent with that file, which already imported `select` and exercises it at line 57.

### Validated: threads opened by another reviewer

**Construct `auxclick` with `PointerEvent`** ([thread](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809385234)). Disposition: invalid or obsolete for this pull request. Measured that `auxclick`, `click`, and `contextmenu` are all non-`PointerEvent` here and all discard `pointerType` and `pointerId`, while `pointerdown` and `pointerup` carry them. The observation is real but not specific to `auxclick` and not introduced here, it duplicates the earlier thread that was withdrawn and resolved, and the underlying cause is registered as [#7156](https://github.com/ariakit/ariakit/issues/7156). Replied on the thread with the measurements.

**Suppress pointer events for intermediate chord changes** ([thread](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809385237)). Disposition: registered follow-up, [#7161](https://github.com/ariakit/ariakit/issues/7161). Confirmed against the specification, which says Pointer Events do not fire overlapping `pointerdown` and `pointerup` for chorded button presses. Confirmed as pre-existing by running the same probe against `b7a53f167` and against this head and getting byte-identical output, so the pull request documents this input without changing it. Replied on the thread with the specification quote, both measurements, and the assessment.

### One optional polish, not published inline

The changeset presents `buttons` as a way "to describe a chorded gesture" without noting that the pointer event types are not the ones a chord produces, which is the only part of #7161 attributable to this pull request. A short caveat in the `mouseDown` and `mouseUp` prose, or a link to #7161, would close the gap between what the sentence promises and what the helpers emit. Related and equally optional: `click` and `select` now carry the "an explicit `buttons` is ignored" note in their reference sections, while `tap` and `rightClick` do not, even though the changeset names all four. `tap`'s prose already points readers at `click`; `rightClick`'s does not. Neither is a defect, so I am not publishing either inline.

<details>
<summary>Cycle 3: Continue the current direction, register the remaining edge case</summary>

**Assessment**

Three finding-bearing review cycles have now landed on one piece of machinery, namely whether callers can express `buttons` at all: the value being dropped (`f8cda48`), the value being honored where one init spans several steps (`570850d`), and now the surviving single-step escape hatch not matching the specification's chorded event types (`5a1a6d9`). Counting them together across reviewer sources puts this at the cycle 3 boundary, so I paused before acting on the dispositions.

Issue severity for the remaining item is low. The compatibility `mousedown` and `mouseup` still fire with correct `button` and `buttons`, so only assertions on pointer event types during a chord are affected. Supported scope is narrow, since `@ariakit/test` is a test utility whose blast radius is test authors. Expected frequency is rare: no caller in `packages` or `app` passes `buttons` to these helpers, and no library source reads `event.buttons` at all. Likely user impact is confined to external consumers writing chord-specific tests, who also have a workaround that composes the accurate sequence from the public `dispatch` surface today.

Complexity did not grow across the cycles; it shrank. Cycle 2's fix removed a branch rather than adding one, reverting `getHoverOptions` to a constant and replacing per-call special cases with a single `omitButtons` boundary applied once in `click` and once in `select`. The net addition for cycle 3's direction is one nine-line helper with no state and no public export, since `__mouse.ts` is private and absent from `index.ts`. Fixing the remaining edge case now would reverse that trend: it would add a conditional pointer-event path to two helpers whose documented contract is that they fire `pointerdown` and `mousedown`, and `pointerup` and `mouseup`, and would invalidate a test added in this pull request.

**Decision**

Continue the current direction. Do not add chord-aware branching to `mouseDown` and `mouseUp` here, and do not remove the escape hatch, which restores a capability that exists on `main`.

The remaining edge case is registered as [#7161](https://github.com/ariakit/ariakit/issues/7161), verified pre-existing on `main`, so it does not need to wait for this pull request. That issue deliberately leaves open whether the outcome should be the branch or a narrowed contract in which the helpers stay single-purpose and chorded sequences are composed from `dispatch`.

Intentionally unsupported: chorded gestures through `click`, `tap`, `rightClick`, and `select`, which now derive each step and ignore an explicit `buttons`. This is documented in the changeset, the `click` and `select` reference prose, and their JSDoc.

Carried follow-ups unchanged: [#7156](https://github.com/ariakit/ariakit/issues/7156) for the dispatch-layer initialization gap.

**Process note**

`ariakit-general-review-reassessment` also calls for an advisory read-only subagent to reassess in parallel with parent context disabled. This review session is configured without subagent delegation, so I performed the reassessment only in the main task, which is where the skill places the decision in any case. Recording the omission so the gap in the audit trail is visible rather than silent.

</details>

This entry belongs in the pull request description under an `## Implementation review reassessment` heading before the next push, per `ariakit-general-review-reassessment`.

### Commands and experiments

- `pnpm test --run packages/ariakit-test` on this head: 15 files, 91 tests, all passing.
- `pnpm exec tsc -p packages/ariakit-test/tsconfig.json --noEmit`: clean. `oxfmt --check` and `oxlint` on the changed sources: clean.
- Probed `tap`, public `hover`, caller-object mutation, and modifier preservation on this head.
- Probed event interfaces and pointer-property propagation for `pointerdown`, `mousedown`, `pointerup`, `click`, `contextmenu`, and `auxclick`.
- Probed the chorded press and release sequence on this head, then checked out `b7a53f167`'s `packages/ariakit-test/src` and re-ran the identical probe to establish the pre-existing baseline, then restored this head.
- Read the two cited Pointer Events sections directly rather than quoting them from the thread.
- Removed every experiment; the review worktree ends clean at `5a1a6d9`.

### Not fully validated

- The cross-browser recordings in the pull request description remain taken on trust; happy-dom cannot distinguish them and I did not re-record them.
- I did not run the full repository Vitest suite or the Playwright browser suites locally. Those rest on CI, which is complete and green for `5a1a6d9` at 44 successful check runs with `Build styles` skipped.

packages/ariakit-test/src/mouse-down.ts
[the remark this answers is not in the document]
diegohaz: Tracked in [#7161](https://github.com/ariakit/ariakit/issues/7161): the pointer event types are identical on `b7a53f167`, so correcting them is a behavioral change to `mouseDown` and `mouseUp` in their own right rather than part of adding non-primary button support.

The wording this thread identifies as the one part attributable to this pull request is fixed in [`b3e64b1`](https://github.com/ariakit/ariakit/commit/b3e64b18e). Borrowing the specification's "chorded gesture" vocabulary invites the reader to expect the specification's event sequence, so both helpers and the changeset now say which event fires:

> When another button is already held, a browser fires `pointermove` instead of `pointerdown`. This helper fires `pointerdown` in both cases.

I also corroborated the expected sequence in real browsers, since the measurements on the issue are helper-side. Driving a left press, a right press while left is held, then both releases through Playwright, Chromium, Firefox, and WebKit all produced `pointermove button=2 buttons=3` for the chord press and `pointermove button=2 buttons=1` for the chord release, with `pointerdown` and `pointerup` bracketing the whole gesture once. That trace, plus a note that the `mousedown` suppression currently keyed on a canceled `pointerdown` has to be dropped on the `pointermove` branch, is recorded in [this comment](https://github.com/ariakit/ariakit/issues/7161#issuecomment-5336585352). I had opened #7163 for the same finding before seeing #7161 and closed it as a duplicate.

diegohaz:
**Fixed**

- [`b3e64b1`](https://github.com/ariakit/ariakit/commit/b3e64b18e) ([reply](https://github.com/ariakit/ariakit/pull/7157#discussion_r3809495101)): `mouseDown` and `mouseUp` now state that a browser fires `pointermove` instead of `pointerdown` or `pointerup` when another button is already held, and that both helpers fire theirs in either case. The changeset says the same. Documentation only; no behavior change.

**Tracked**

- [#7162](https://github.com/ariakit/ariakit/issues/7162): browsers dispatch `click`, `auxclick`, and `contextmenu` as `PointerEvent`, while the dispatch layer builds all three as `MouseEvent` and drops the `PointerEventInit` properties the press phase honors. Measured on Chromium, Firefox, and WebKit, and under happy-dom. Pre-existing and uniform across all three events, so it belongs at the dispatch layer next to #7156 rather than at the `auxclick` call site.
- [#7161](https://github.com/ariakit/ariakit/issues/7161): corroborated the chorded-pointer-event finding in real browsers and recorded the trace, plus a note that the `mousedown` suppression keyed on a canceled `pointerdown` has to be dropped on the `pointermove` branch, in [this comment](https://github.com/ariakit/ariakit/issues/7161#issuecomment-5336585352). I opened #7163 for the same finding before seeing #7161 and closed it as a duplicate.

packages/ariakit-test/src/__mouse.ts
diegohaz: Fixed in [`5a1a6d9`](https://github.com/ariakit/ariakit/commit/5a1a6d90ae77c1018acca141474cfb66d469fcb8): took the alternative and stripped `buttons` once at the `click` and `select` boundary, so every phase derives its own value again and `mouseDown` and `mouseUp` stay the documented chord path. `rightClick(button, { buttons: 1 })` is back to `pointerdown 2 2` and `click(button, { button: 1, buttons: 1 })` reports `pointerdown 1 4`, with targeted coverage in the same commit: the hover assertion now passes `{ button: 1, buttons: 1 }` and still expects `0 0` on all six move events, `click and rightClick derive buttons from the button they press` pins both full sequences, and `select derives buttons from the button it presses` pins the `select` half. Both strips are red-checked: reverting the one in `click.ts` fails with `pointerdown 1 1 0` against `pointerdown 1 4 0`, and reverting the one in `select.ts` fails with `pointerdown 0 3` against `pointerdown 0 1`.

I went with the alternative rather than the smaller test change because the smaller one records the `rightClick` half instead of fixing it. `rightClick` would keep emitting `pointerdown 2 1`, and an assertion on it turns an impossible pointer state into the contract, which is harder to walk back later than the pre-PR `click` behavior it costs. That behavior was itself an artifact of `click` having no `button` support: `buttons: 1` was coherent on all seven events only because the press was always the primary one. It stops being coherent the moment `button` can be something else, so restoring it across all three phases would only reproduce the same defect on `click` that the `rightClick` example makes visible.

The changeset already said `mouseDown` and `mouseUp` are the chord path, and it now also states that `click`, `tap`, `rightClick`, and `select` ignore an explicit `buttons`. `rightClick` keeps firing the same sequence it fired before this PR; `select` passed the init through unfiltered on `b7a53f167`, so its drop is a real user-visible change and the changeset calls that out.

PR #7158, opened 2026-08-18
Render TagList as the listbox element

### LINKED ISSUE #7155: Clicking the tag field no longer focuses the tag input
## Context

Found during the pre-commit review of the change that drops the `aria-owns` hack from the experimental Tag components. That change makes `TagList` render the `listbox` element itself, so `TagInput` has to be a sibling of `TagList`, and the recommended layout wraps both in a container element and gives `TagList` a `display: contents` style.

## Problem

`useTagList` registers an `onMouseDown` handler documented as "Focus on the input element when clicking on the tag list". It is what makes clicking the empty area of a tag field focus the input, which is the standard affordance for this kind of control.

In the new recommended layout that handler can no longer run for that case:

- An element with `display: contents` generates no box, so a `mousedown` on the container's padding targets the container. The container is `TagList`'s parent, so the event never reaches `TagList`.
- Every `mousedown` that does reach `TagList` originates inside a `Tag`, and the handler returns early for those: `if (!isSelfFocusable && currentTarget.contains(focusableTarget)) return;`

The visible effect is in `app/src/sandbox/tag/`. The container carries `ak-input`, which applies `ak-frame-field` padding and `cursor-text`, so the pointer shows a text caret over the field's padding but clicking there no longer focuses the input.

## Verification against `main`

This is **not** pre-existing on `main`. Verified in an isolated worktree at `b7a53f16722b2397d1bd32b0677bbdfb7f8e6608`: on `main`, `ak-input` is applied directly to the element `Ariakit.TagList` renders (`app/src/sandbox/tag/index.react.tsx`, and the local wrapper in `app/src/sandbox/tag-combobox/tag.react.tsx`), no sandbox or JSDoc example gives `TagList` a `display: contents` style, and a padding click therefore reaches `TagList` with `event.target === event.currentTarget`, passes the guard, and focuses the input.

So this issue only becomes real once the `aria-owns` removal merges, and work on it should wait for that.

## Why the originating change can ship without this

Dropping `aria-owns` is the point of that change, and it is the right trade because Safari does not expose `aria-owns` relationships reliably. The handler is kept because it is still correct whenever `TagList` renders as a real box, which `app/src/sandbox/tag-list-6311` covers. The narrowing is documented in the `TagList` JSDoc and in the changeset, so consumers are told to handle the click on their own container.

Restoring it properly needs a design decision that is larger than that change: either a container component that owns the field styles and the click behavior, or an exported helper or prop set that consumers apply to their own container.

## Expected outcome

Clicking the empty area of a tag field focuses the input again, in the sibling layout, without consumers having to reimplement the `queueBeforeEvent` and `getClosestFocusable` logic. Whatever the shape, the tag sandbox should demonstrate it, and a browser test should cover clicking the field's padding.

diegohaz:
## Motivation

The experimental `TagList` component could not render the `listbox` element directly. `TagInput` had to be a sibling of the tags so the whole field could be styled as a single input, but the `listbox` role accepts only options as children, so an input inside it would be invalid.

To work around this, `TagList` rendered a second, visually empty element that carried the `listbox` role and referenced each tag through `aria-owns`:

```html
<div>
  <div role="listbox" aria-owns="tag-1 tag-2" style="position: fixed"></div>
  <div role="option" id="tag-1">JavaScript</div>
  <div role="option" id="tag-2">React</div>
  <input />
</div>
```

Safari has poor support for `aria-owns`, so the tags were not reliably exposed as options of that listbox.

## Solution

`TagList` now renders the `listbox` element itself and carries the tag list ARIA attributes directly, so `aria-owns` is gone:

```html
<div role="group" class="tag-list">
  <div role="listbox" style="display: contents">
    <div role="option">JavaScript</div>
    <div role="option">React</div>
  </div>
  <input />
</div>
```

Because the `listbox` role accepts only options as children, `TagInput` moves out of `TagList` and becomes its sibling. The new `TagControl` component renders the container that wraps both and takes over the field styles, and `TagList` gets a `display: contents` style, so the tags and the input still share one layout.

Before:

```tsx
<TagList className="tag-list">
  {values.map((value) => (
      {value}
    </Tag>
  ))}
  <TagInput />
</TagList>
```

After:

```tsx
<TagControl className="tag-list">
  <TagList style={{ display: "contents" }}>
    {values.map((value) => (
        {value}
      </Tag>
    ))}
  </TagList>
  <TagInput />
</TagControl>
```

`TagControl` is what makes the sibling structure work without spreading behavior across the two children. It owns the two things that belong to the whole field: clicking it focuses the input, and the undo and redo shortcuts are handled there, where key events from both the tag list and the input bubble through. Both used to live on `TagList`, which can no longer host them now that it is the listbox and the input is outside it.

It also supplies the store to both children, so a consumer who passes a `store` prop instead of using `TagProvider` passes it to `TagControl` alone.

`TagControl` renders a `group` element as well. The tags and the input look like a single field but are two separate widgets for assistive technologies, and the group is what holds them together. It also gives the field an accessible handle, so tests and consumers can reach it by role instead of by class name. Because one label now names the group, the listbox and the input, `TagListLabel` is renamed to `TagLabel`.

Chrome exposes the tag sandbox as:

```
group "Tags"
  listbox "Tags"
    option "JavaScript"
    option "React"
  textbox "Tags"
```

The name is repeated on all three nodes. That is the cost of naming the field once and letting every part of it carry that name. Passing `aria-label` or `aria-labelledby` to `TagControl` overrides it.

Closes #7155.

## Trade-offs

One behavior is narrower after this change, and it is documented in the `TagList` JSDoc and the changeset.

`aria-owns` referenced the tags by id, so any element rendered between `TagList` and the tags was invisible to assistive technologies. Now that containment is structural, a wrapper element between the listbox and its options can stop assistive technologies from seeing the tags as options. This is the cost of dropping `aria-owns`, and it is the better trade because `aria-owns` was not reliable in Safari to begin with.

## Changes

- `TagList` renders `role="listbox"` (`role="list"` on touch devices) plus `aria-live`, `aria-relevant`, `aria-atomic`, `aria-orientation`, and `aria-labelledby` on its own element. The prop-stripping loop that moved `role` and `aria-*` props onto the extra element is gone, so those props now apply to the element the consumer renders.
- `TagControl` renders the field container and owns click-to-focus, the undo and redo shortcuts, and supplying the store to both children.
- `TagControl` renders `role="group"` with `aria-labelledby` taken from the tag label, so the field is one named unit for assistive technologies.
- `TagListLabel` is renamed to `TagLabel` and its module moves to `@ariakit/react-components/tag/tag-label`, because one label now names the group, the listbox and the input.
- `TagList` documents that only `Tag` components belong inside it, and that `TagControl` is where those two behaviors live.
- Every tag sandbox renders `TagInput` as a sibling of `TagList`.
- JSDoc examples across the tag modules show the new structure.

## Verification

- New regression coverage in `app/src/sandbox/tag/test-browser.ts` and `app/src/sandbox/tag/test.ts` asserts that the tags are inside the listbox and the input is outside it. Reverting only `tag-list.tsx` to the base version turns the happy-dom test red because the `aria-owns` listbox does not contain the option.
- The behaviors that moved into `TagControl` are red-checked: disabling its focus call fails the click-to-focus browser test, and disabling its undo and redo branch fails the existing `undo/redo removing tags` and `undo/redo typing` tests.
- The `group` role is red-checked three ways: removing `role="group"` fails the new happy-dom test, removing its `aria-labelledby` fails the element-id browser test, and dropping the `aria-label` guard fails the new `aria-labelledby-aria-label` case, because `aria-labelledby` then wins over the consumer's own name.
- No sandbox passes a tag `store` prop, so `TagControl` handing the store to its children through context has no fixture. That matches `main`, where no sandbox passed one either.
- The tag sandbox's browser tests are now one `test-browser.ts` running on Chrome, Firefox and Safari rather than a Chrome-only file plus a separate one, which took the clipboard permission grant behind a `browserName` guard.
- `pnpm test` (1776 tests), `pnpm tsc`, `pnpm lint-fix`, and `pnpm build` all pass. `pnpm build` regenerates the exports map with `./tag/tag-label` in place, `npm pack --dry-run` carries no `tag-list-label` file, and the old subpath now throws `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- `pnpm test-react18` does not run in this environment: its harness fails while installing into a temporary workspace. It fails the same way on `sandbox/checkbox-interactions`, which this change does not touch, so it is environmental rather than caused by this change. CI covers that suite.
- Playwright: `sandbox/tag*`, `sandbox/combobox-select-tag-tooltip-element-id`, and `sandbox/aria-labelledby-aria-label` pass on Chrome, Firefox, and Safari.
- The `tag` and `tag-combobox` previews measure identical bounding boxes before and after the change. The field, both tags, and the input keep the same x, y, width, and height.
- Chrome exposes the tag sandbox as the tree shown above, with the textbox as a sibling of the listbox, and the page carries no `aria-owns` attribute.

One evidence gap is worth naming precisely. The unverified claim is that WebKit and Gecko expose an element with a `display: contents` style carrying an explicit `role="listbox"` as a listbox in their accessibility trees. Playwright's role queries are computed from the DOM, so the browser test cannot answer that, and this repository has no WebKit or Gecko accessibility-tree probe. Chrome is covered by the tree above. `display: contents` has a history of accessibility-tree defects across engines, so this deserves a real check before the Tag components leave experimental status.

The gap does not block the change. It is not a regression against the base, where Safari did not honour `aria-owns` and the tags were not associated with the listbox at all. The structural win, `TagList` being the listbox with the tags as its real children, holds regardless, and if an engine drops the boxless listbox from its tree, the fix is application-side CSS rather than another library break.

## Preview

The tag sandbox renders exactly as before. This screenshot of `/react/previews/tag/` and the one taken at the base revision are byte-identical, verified by SHA256, with the base confirmed to be serving `aria-owns` at capture time.

The tag sandbox: a Tags label above a bordered field containing the JavaScript and React tags, each with a remove button, and the tag input below them

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the current design</summary>

**Assessment**

Seven findings across three gate cycles. None alleged a correctness defect in the new behavior. Five were documentation (the two narrowed behaviors and the completeness of the migration guidance), one was a dead line of code, and one was a verification gap that closed by running two Playwright suites, which passed. The repeated cycles track the documentation surface of a breaking structural change, not implementation complexity.

Implementation complexity went down. `tag-list.tsx` lost 78 lines and gained 46, and most of the additions are JSDoc. Gone: a `renderedItems` subscription, the id-join for `aria-owns`, an unguarded `for...in` loop that stripped `role` and `aria-*` props off the consumer's element with `delete`, a synthetic DOM element, and a children override. The only new machinery is `handleUndoRedoShortcut`, a 12 line function holding the code that already lived inline in `TagList`. No new component, prop, or public concept. Maintenance complexity also drops, since `TagList` no longer re-renders on every item registration and there is one fewer element and ARIA relationship to keep in sync.

Two behaviors narrow. Clicking the empty area of a field no longer focuses the input when `TagList` uses `display: contents`, which is Medium severity and likely for anyone following the new docs with a padded field, though it is a pointer affordance rather than a correctness or accessibility defect. Elements rendered between the listbox and the tags can now hide the options from assistive technologies, which is Medium severity and possible for consumers with wrapper integrations. The second is the inherent price of dropping `aria-owns`: repairing it means keeping `aria-owns`, which is what the task asked to remove.

**Decision**

Continue the current design.

Keeping `useTagList`'s click-to-focus handler is right. Removing it would break the behavior wherever `TagList` still renders a real box, which `app/src/sandbox/tag-list-6311` covers on all three engines. Restoring it for the sibling layout needs either a container component or an exported helper, which is public API this task did not ask for and a design decision in its own right, so it is registered as https://github.com/ariakit/ariakit/issues/7155 rather than grown into this change.

`handleUndoRedoShortcut` stays in `tag/utils.ts`. Handling the shortcut on both focusable roots is a direct consequence of the requested sibling structure, and one shared function is the minimal way to avoid duplicating it. That module is an already-public subpath holding the same class of internal helper, and `composite/utils.ts` sets the same precedent.

The documentation added is proportionate. Three JSDoc paragraphs and three changeset paragraphs for a breaking change to a component's required markup is close to the minimum, and cycle 2 showed that an incomplete changeset actively misleads.

A parallel advisory complexity pass reached the same conclusion independently and proposed no disposition change. It measured the same shape from the other side: filtering comments and JSDoc out of the staged library diff leaves about 91 touched lines, roughly 55 of them deletions of the `aria-owns` apparatus and 12 the undo/redo body moved verbatim, so the net new logic is about two lines. It also confirmed by reading `composite/*` that the sibling structure is safe: `virtualFocus` is off and the store seeds `activeId` from the input, so `TagList` never receives a `tabIndex` and the boxless composite root is not a phantom tab stop; arrow keys are handled on the item; and `ownsFocus` already accounts for items outside the composite element. It scored the store-inheritance break Medium rather than High, since the existing `invariant` fails fast in development, which does not change its disposition.

It raised two items not previously seen, both declined here. First, an `onKeyDown` on `TagList` no longer sees the input's key events, so only a handler on `TagInput` can intercept the shortcut. That is one instance of the general fact that the input is now a sibling, which the changeset states plainly, and enumerating each event-bubbling consequence is not proportionate. Second, the paragraph about elements between the listbox and the tags could prescribe a remedy such as `role="none"` on an unavoidable wrapper. This repository has no way to verify that remedy, and the earlier "must not generate a box" wording was already rejected for being an unverified accessibility claim, so the paragraph states the constraint without prescribing a fix.

It also noted that restoring click-to-focus inside the tag sandbox alone would take about five lines. That is deliberately not done: the fix belongs in the library rather than copied into a fixture, the sandbox tests do not cover it, and demonstrating a workaround now would prejudge the design in https://github.com/ariakit/ariakit/issues/7155. Worth recording that the affordance is not fully lost either, because `TagListLabel` renders a real `label` element pointing at the input, so clicking the label still focuses it.

Intentionally unsupported: clicking the empty area of a `display: contents` field to focus the input (https://github.com/ariakit/ariakit/issues/7155); elements rendered between `TagList` and the tags; a `TagList` `onKeyDown` intercepting the input's key events; and `TagInput` inheriting a store passed only to `TagList`, which throws a clear invariant in development. One evidence gap stays open: engine accessibility-tree exposure of a boxless element carrying `role="listbox"` is confirmed only in Chrome, since this repository has no harness for asserting WebKit or Gecko accessibility trees. That is not a regression against the base, where Safari already failed to honour `aria-owns`.

No material direction change, so the cycle count does not reset.

**Superseded**

Review feedback on the pull request asked for the `TagControl` component described above. It resolves the click-to-focus narrowing this entry had accepted, and https://github.com/ariakit/ariakit/issues/7155 is closed by this pull request rather than left as follow-up work.

Three more details in this entry no longer describe the branch. `handleUndoRedoShortcut` is gone, because the undo and redo handling moved into `TagControl` rather than being shared between `TagList` and `TagInput`, so `tag/utils.ts` is now unchanged from `main`. `TagListLabel` is renamed to `TagLabel`. And `TagControl` now carries a `group` role. The rest of this entry still stands.

</details>


app/src/sandbox/tag-combobox/test.ts
```diff
@@ -12,13 +12,12 @@ function getSelectionText(element: HTMLElement | HTMLInputElement | null) {
...
-  const combobox = q.combobox("Invitees");
-  return q.option
-    .all(undefined, {
-      name(_, element) {
-        return !!element.parentElement?.contains(combobox);
-      },
-    })
+  // The combobox suggestions are options too, so only the ones inside the tag
+  // list count as tags.
+  const tagList = q.listbox.ensure("Invitees");
```
diegohaz: Is `ensure` necessary here?

app/src/sandbox/tag/test-browser.ts
```diff
@@ -0,0 +1,15 @@
+import { withFramework } from "#app/test-utils/preview.ts";
```
diegohaz, line 3: Add this to `test-chrome.ts` and rename that file to `test-browser.ts`.

packages/ariakit-react-components/src/tag/tag-input.tsx
```diff
@@ -192,22 +197,24 @@ export const useTagInput = createHook<TagName, TagInputOptions>(
...
- *       {(values) =>
- *         values.map((value) => (
- *           <Tag key={value} value={value}>
- *             {value}
- *             <TagRemove />
- *           </Tag>
- *         ))
- *       }
- *     </TagValues>
+ *   <div className="tag-list">
```
diegohaz:
```suggestion
 *   <TagControl>
```

packages/ariakit-react-components/src/tag/tag-list.tsx
```diff
@@ -98,94 +86,74 @@ export const useTagList = createHook<TagName, TagListOptions>(
...
+ * The [`TagInput`](https://ariakit.com/reference/tag-input) component must be
+ * rendered as a sibling of this component. To style them together as a single
+ * input field, wrap both in a container element and give this component a
+ * `display: contents` style.
+ *
+ * Because this component is the listbox element, any element rendered between
+ * it and the tags can stop assistive technologies from seeing the tags as
+ * options of this listbox.
+ *
+ * Clicking this element focuses the input element. This doesn't apply when the
```
diegohaz: We should probably have a `TagControl` component that handles this.

packages/ariakit-react-components/src/tag/tag-input.tsx
```diff
@@ -153,6 +157,7 @@ export const useTagInput = createHook<TagName, TagInputOptions>(
     const onKeyDown = useEvent((event: KeyboardEvent<HTMLType>) => {
       onKeyDownProp?.(event);
       if (event.defaultPrevented) return;
+      handleUndoRedoShortcut(event);
```
diegohaz: If we have `TagControl`, this `handleUndoRedoShortcut` logic can be reverted into that parent component (which I believe was handled by `TagList` before).

app/src/sandbox/tag-combobox/test.ts
diegohaz: Fixed in [`f3ca05b`](https://github.com/ariakit/ariakit/commit/f3ca05b0711f80038f4292edf6d24fb1082d0238): it was not. `q.within()` invariants on a null element, so the plain query already fails loudly. Dropped it here and in the equivalent spot in `app/src/sandbox/tag/test.ts`.

The one remaining `ensure` in the tag tests is the argument of `not.toContainElement(q.textbox.ensure("Tags"))`, which stays because a null there would pass vacuously.

app/src/sandbox/tag/test-browser.ts
diegohaz: Fixed in [`dc19ff3`](https://github.com/ariakit/ariakit/commit/dc19ff31458073eef68ee1e83824faf40543d29c): the structural test moved into `test-chrome.ts` and that file is now `test-browser.ts`, so all 11 tests run on Chrome, Firefox and Safari. Tag browser coverage went from 32 to 53 passing.

One thing had to change to make the rename work. `context.grantPermissions(["clipboard-read", "clipboard-write"])` throws `Unknown permission: clipboard-read` on Firefox, which is why the file was Chrome only. Only the grant is Chromium specific though: the tests write with `navigator.clipboard.writeText`, which a focused document is allowed to do, and paste with a real key press rather than reading the clipboard. I verified all 9 paste tests pass on all three engines without it, then kept the grant behind a `browserName` guard rather than removing it, since CI runs Chromium under `xvfb` and I could not verify that environment locally.

packages/ariakit-react-components/src/tag/tag-input.tsx
diegohaz: Fixed in [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6): applied, and the same swap went through every tag module JSDoc example plus the `useTagStore` one, so no example renders a plain container div anymore.

packages/ariakit-react-components/src/tag/tag-list.tsx
diegohaz: Fixed in [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6): added `TagControl`. It renders the container that wraps `TagList` and `TagInput`, and it owns clicking the field to focus the input, so that behavior no longer depends on `TagList` generating a box. This closes #7155, which I had opened for exactly this.

It also supplies the store to both children, so a consumer who is not using `TagProvider` now passes `store` to `TagControl` alone instead of to each child. The `tag-combobox` sandbox was switched to that pattern so its existing tests cover it: removing the provider fails all 5.

Covered by a new browser test that clicks the field padding and asserts the input is focused. Red-checked by disabling the focus call.

packages/ariakit-react-components/src/tag/tag-input.tsx
diegohaz: Fixed in [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6): the shortcut moved into `TagControl`, where key events from both the tag list and the input bubble through, so it is handled in one place again. `useTagInput` no longer touches it and the `handleUndoRedoShortcut` helper is gone, which leaves `tag/utils.ts` byte-identical to `main`.

Covered by the existing `undo/redo removing tags` and `undo/redo typing` tests, which exercise both the tag-focused and input-focused paths. Red-checked by disabling the branch in `TagControl`.

diegohaz:
**Fixed**

- [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809245866)): added `TagControl`. It renders the field container and owns clicking to focus the input, so that no longer depends on `TagList` generating a box. Closes #7155.
- [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809246487)): the undo and redo shortcut moved into `TagControl`, where events from both children bubble through. `handleUndoRedoShortcut` is gone and `tag/utils.ts` is back to its state on `main`.
- [`c67ef10`](https://github.com/ariakit/ariakit/commit/c67ef10f3b4efad346b6232d1342fc507375e1e6) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809245154)): the suggestion is applied, and every tag module's JSDoc example now renders `TagControl` instead of a container div.
- [`dc19ff3`](https://github.com/ariakit/ariakit/commit/dc19ff31458073eef68ee1e83824faf40543d29c) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809244574)): the tag sandbox's browser tests are one `test-browser.ts` again, running on Chrome, Firefox and Safari. Coverage went from 32 to 53 passing.
- [`f3ca05b`](https://github.com/ariakit/ariakit/commit/f3ca05b0711f80038f4292edf6d24fb1082d0238) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809243927)): dropped the unnecessary `ensure` calls. `q.within()` already rejects null.

**Tracked**

- [#7160](https://github.com/ariakit/ariakit/issues/7160): an empty `TagList` with a `display: contents` style is absent from Playwright's role queries, though Chrome's accessibility tree does list it. Three sandboxes with initially empty tag lists omit the style for that reason, so the fixtures and the recommended pattern disagree until that is decided.

**Notes**

`TagControl` also supplies the store to both children, so `store` goes on `TagControl` alone rather than being passed to `TagList` and `TagInput` separately. The `tag-combobox` sandbox was switched to that pattern so its existing tests cover it.

Two things needed verifying rather than assuming. Renaming the clipboard test file was blocked by `grantPermissions(["clipboard-read", ...])`, which throws on Firefox; only the grant is Chromium specific, so guarding it was enough, and I kept it rather than deleting it because CI runs Chromium under `xvfb`. Each behavior that moved into `TagControl` was red-checked by disabling it.

.changeset/100-tag-list-listbox-element.md
```diff
@@ -0,0 +1,45 @@
...
+      </Tag>
+    ))}
+  </TagList>
+  <TagInput />
+</TagControl>
+```
+
+`TagControl` takes over the two behaviors that used to live on `TagList`: clicking the field focuses the input, and the undo and redo shortcuts are handled there. Both now cover the tags and the input from a single element, so render a `TagControl` if you rely on either.
+
+`TagInput` no longer inherits the store from `TagList`. It inherits it from `TagControl` instead, so pass the `store` prop to `TagControl` when you're not using `TagProvider`.
```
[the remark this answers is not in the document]
diegohaz: Validated against the current head: this claim does not hold, so the changeset sentence is accurate as written.

`TagControl` wraps its children in `TagContextProvider` (`packages/ariakit-react-components/src/tag/tag-control.tsx:90`), and `TagContextProvider` is `ctx.ContextProvider` (`packages/ariakit-react-components/src/tag/tag-context.tsx:38`), which is the same provider `TagProvider` renders. `useTagProviderContext` is `ctx.useProviderContext` (`packages/ariakit-react-components/src/tag/tag-context.tsx:36`), and that hook reads the unscoped context, returning `undefined` only when the nearest *scoped* provider holds the same store:

```ts
const useProviderContext = () => {
  const scoped = React.useContext(scopedContext);
  const store = useContext();
  if (scoped && scoped === store) return;
  return store;
};
```

(`packages/ariakit-react-utils/src/system.tsx:118-123`)

Nothing renders a scoped tag provider between `TagControl` and `TagList`, so `TagList` resolves the control's store and the invariant never fires. The scoped provider is the one `TagList` itself renders for its own descendants (`packages/ariakit-react-components/src/tag/tag-list.tsx:47`).

There is empirical confirmation in the branch too. `app/src/sandbox/tag-combobox/tag.react.tsx` builds the field with `useTagStore` plus `<TagControl store={tag}>` and no `TagProvider`, and its five happy-dom tests pass locally (`pnpm test app/src/sandbox/tag` on this head: 4 files, 25 tests passed). The `useTagStore` JSDoc example in `packages/ariakit-react-components/src/tag/tag-store.ts` documents the same shape.

Disposition: invalid, no change needed.

app/src/sandbox/tag-combobox/tag.react.tsx
```diff
@@ -2,37 +2,52 @@ import { clsx } from "clsx";
 import * as React from "react";
 import * as Ariakit from "./ariakit-experimental.react.ts";
 
-export interface TagListProps extends Ariakit.TagListProps {
+export interface TagFieldProps extends Ariakit.TagControlProps {
```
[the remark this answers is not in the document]
diegohaz: Validated against the current head: the mechanism is real, but it is a latent type-surface quirk in this sandbox fixture rather than a defect, and it predates this change.

Real part: `TagFieldProps extends Ariakit.TagControlProps` (`app/src/sandbox/tag-combobox/tag.react.tsx:5`) inherits the optional `store` prop, and `{...props}` is spread after `store={tag}` on line 31, so a caller-supplied `store` would reach `TagControl` while `TagListLabel` on line 24 keeps the store created on line 20.

Why it is not actionable here:

- Nothing passes it. The only caller is `app/src/sandbox/tag-combobox/index.react.tsx`, which passes `label`, `value`, `setValue`, `values`, `setValues`, and children.
- The same shape exists on the base revision. `TagListProps extends Ariakit.TagListProps` also carried `store`, spread after `ref` onto `Ariakit.TagList` inside `TagProvider`, so a caller-supplied `store` split the tag list from the label and the input in exactly the same way. This change only moves which components follow the overriding store.
- It is fixture code, not shipped component behavior. `TagControl` itself resolves `store || context` and throws a development invariant when neither is present.

Disposition: declined non-goal for this pull request.

If it is worth closing cheaply anyway, the smallest fix is to drop the prop from the fixture's surface rather than to reorder the spread:

```tsx
export interface TagFieldProps
  extends Omit<Ariakit.TagControlProps, "store"> {
```

app/src/sandbox/combobox-select-tag-tooltip-element-id/index.react.tsx
```diff
@@ -84,9 +85,10 @@ function TagExample() {
     <section aria-label="Tag">
       <TagProvider>
         <DynamicTagLabel />
-        <TagList>
+        <TagControl>
+          <TagList />
```
diegohaz, line 89: Three sandboxes deliberately omit the `display: contents` style that this change's docs, changeset, and every JSDoc example recommend, and nothing in the code records why.

The split is exactly by initial tag values:

| Sandbox | Initial values | `display: contents` |
| --- | --- | --- |
| `tag` | `["JavaScript", "React"]` | yes |
| `tag-combobox` | one invitee | yes |
| `tag-list-6311` | `["JavaScript", "React"]` | yes |
| `tag-remove-6330` | `["React", "Vue"]` | yes |
| `tag-input-6294` | `[]` | no |
| `tag-input-6333` | `[]` | no |
| `combobox-select-tag-tooltip-element-id` | no tags at all | no |

https://github.com/ariakit/ariakit/issues/7160 explains it: a boxless empty listbox drops out of role-based queries, and that issue records that adding the style to this sandbox made `section.listbox("Tag label")` fail with `element(s) not found`. That reasoning lives only in the issue. Neither the code nor the pull request description points at it, so the divergence reads as an oversight in review and in `git blame`.

The concrete failure this invites: a contributor tidying the fixtures to match the documented pattern adds the style here, and `updates tag relationships when element ids change` in `app/src/sandbox/combobox-select-tag-tooltip-element-id/test-chrome.ts:50` starts failing on the `section.listbox("Tag label")` query for a reason none of the surrounding code explains. An Ariakit consumer reading these sandboxes as reference markup hits the same surprise from the other direction: they copy a `TagList` with no style, or add the style and lose the listbox from their own role-based tests.

`ariakit-general-code-style` asks for a comment on intentionally narrow solutions, with the local constraint summarized and the stable link carrying the history. A short comment above each of the three would do it:

```suggestion
          {/* Starts empty, and a boxless empty listbox drops out of
              role-based queries, so this deliberately omits the
              display: contents style the docs recommend.
              https://github.com/ariakit/ariakit/issues/7160 */}
          <TagList />
```

Linking https://github.com/ariakit/ariakit/issues/7160 from the pull request description's Trade-offs section would also help, since that section currently names only the wrapper-element trade-off.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the fixtures behave correctly today and only future edits are at risk.
- Priority: P2 (estimate, 80% confidence), because this pull request creates the divergence but nothing breaks until someone acts on it.
- Breakage likelihood: Possible (estimate, 75% confidence), because an unexplained three-way inconsistency against a pattern repeated in six JSDoc examples and the changeset is a natural cleanup target.
- Fix effort: Easy (verified, 95% confidence), because it is a comment in three files plus one line in the pull request description.

diegohaz (commented):
Initial review. No prior review carrying this workflow's marker exists on this pull request, so the baseline is the merge base `b7a53f167` through the current head `dc19ff31`, followed by inspection of the surrounding tag, composite, and store code beyond the diff.

The change is sound. The structural claim it rests on is real, the behaviors that moved to `TagControl` are each still covered, and the approach is the right one for the problem.

## What I ran

Fresh worktree on `dc19ff31`, dependencies installed from the lockfile.

- `pnpm tsc`: passes.
- `pnpm test app/src/sandbox/tag`: 4 files, 25 tests, all passing.
- Regression proof, reproduced rather than taken on trust: reverting only `packages/ariakit-react-components/src/tag/tag-list.tsx` to `b7a53f167` turns `renders the tags inside the listbox and the input outside it` red in `app/src/sandbox/tag/test.ts`, failing on line 36 because the `aria-owns` listbox does not contain the `JavaScript` option. The revert was restored and the tree is clean.
- CI on this exact head: every check passing, including Chrome, Firefox, both Safari shards, iOS, Android, Visual, Visual (Safari), Test React 18, Lint, Build packages, and Continuous Releases. I did not run Playwright locally and relied on that run instead.
- Reference links: `https://ariakit.com/reference/tag-list`, `/tag-control`, and `/tag-provider` all return `404`. The app reference collection is generated from `packages/ariakit-react` and `packages/ariakit-solid` only (`app/src/content.config.ts:99` and `:108`), and neither exports the tag modules, so no reference page exists for them yet. The changeset correctly keeps `TagList`, `TagControl`, and `TagInput` as plain inline code, and the new `ariakit.com/reference/tag-control` link in the `tag-control.tsx` JSDoc follows the convention the other tag modules already use.
- Changeset mechanics: `@ariakit/react-components` is at `0.4.1`, so `minor` frontmatter for a breaking change matches the v0 rule, and the `100-` prefix matches the major-change ordering key. `@ariakit/react` is correctly absent, since it re-exports nothing from the tag modules. #7155 is authored by an excluded participant, so no credit sentence is required.
- I verified the changeset's claims against the code individually, including the store-inheritance sentence, which is accurate.

- https://github.com/ariakit/ariakit/pull/7158#discussion_r3809267017 (`.changeset/100-tag-list-listbox-element.md`): **invalid**. `TagControl` provides the store through `TagContextProvider`, which is the same unscoped provider `TagProvider` renders, and `useTagProviderContext` returns it because no scoped tag provider sits between the control and the list. The `tag-combobox` sandbox and the `useTagStore` JSDoc both use that exact shape, and its tests pass. Replied on the thread with the mechanism and the evidence.
- https://github.com/ariakit/ariakit/pull/7158#discussion_r3809267025 (`app/src/sandbox/tag-combobox/tag.react.tsx`): **confirmed but declined non-goal**. The type surface really does let a caller-supplied `store` reach `TagControl` while `TagListLabel` keeps the internally created one, but nothing passes it, the same latent split existed on the base revision through `TagListProps`, and it is fixture code rather than shipped behavior. Replied with the reasoning and a one-line cleanup if it is worth closing anyway.

## Published inline

One finding, on `app/src/sandbox/combobox-select-tag-tooltip-element-id/index.react.tsx`: three sandboxes deliberately omit the `display: contents` style that the changeset and six JSDoc examples recommend, because their tag lists start empty, and nothing in the code or the pull request description records that. The split is exactly by initial values, and https://github.com/ariakit/ariakit/issues/7160 holds the reasoning.

## Registered follow-up

https://github.com/ariakit/ariakit/issues/7164, for the evidence gap this pull request names in its own description. Verified as not pre-existing on `main`: there, `TagList` renders the separate `position: fixed` listbox with `aria-owns`, and no sandbox gives a `TagList` a `display: contents` style, so the boxless-listbox structure exists only on this branch and cannot be measured until it merges.

One refinement to the description's reasoning is worth recording. The argument that the gap is not a regression, because the base relied on `aria-owns` and Safari does not support it well, is sound for WebKit, where the tags were not associated with the listbox to begin with. It does not settle Gecko: neither the base behavior nor the new behavior has been measured there, so both sides of that comparison are unknown for Firefox. That is a reason to scope the follow-up check to both engines, not a reason to hold the change.

- **`display: contents` as consumer boilerplate.** Every documented example now carries `style={{ display: "contents" }}`, which is a real cost on a public API. I weighed defaulting it inside `useTagList`, and it is the wrong trade: it would impose the boxless-listbox exposure risk on every consumer, and #7160 already shows a case where the style is actively wrong. Declined non-goal.
- **Where the undo and redo shortcut belongs.** Handling it on `Tag` and `TagInput` through a shared helper would cover every focus position inside the widget, since focus is always on the input or on a tag, and it would keep undo and redo working for anyone who restructures without adopting `TagControl`. The container is still the better choice: it also covers consumer-rendered focusables inside the field, and it keeps one attachment point instead of two. The current approach is justified; no change requested.
- **`TagControl` itself.** I checked whether the component is avoidable. It is not: `TagList` cannot own click-to-focus once it generates no box, and it cannot see the input's key events once the input is a sibling. Splitting children inside `TagList` is not possible without a component boundary. The new component is the minimum that makes the sibling structure work.
- **Missing happy-dom duplicate for click-to-focus.** The behavior that moved to `TagControl` has browser coverage only. A duplicate is feasible but would need a class-based probe for an element with no role or name, and would not reproduce the padding click that is the point of the test. The three-engine browser coverage plus `app/src/sandbox/tag-list-6311` is adequate.
- **Test-helper vacuity.** I checked the rewritten `tags()` helper in `app/src/sandbox/tag-combobox/test.ts`, because `expect(tags()).toEqual([])` on line 44 would be vacuous if a missing listbox produced an empty scope. `q.within` throws on a null element (`packages/ariakit-test/src/query.ts:285`), so it cannot. `toContainElement(null)` fails rather than passing, so the plain queries on lines 36 and 37 of `app/src/sandbox/tag/test.ts` are not vacuous either, and the negative assertion on line 41 correctly uses `.ensure`.
- **Composite behavior with a boxless composite root.** I confirmed the reasoning in the description independently: `ownsFocus` recognizes an item outside the composite element, `focusable` on the composite resolves from `activeId === null` while the tag store seeds `activeId` from the input element, and `startedOnComposite` in `usePresentItem` only gates `entersClosedPopup`, which needs an `open` state the tag store does not have. Nothing in composite breaks from the input moving outside the list.

## Not fully validated

- Playwright suites were not run locally; the green CI run on this head is the evidence.
- `pnpm build` and the packed contents were not run locally. The new `./tag/tag-control` export is a plain source path shaped exactly like its siblings, the build discovers public files from the source directory, and CI's package build and Continuous Releases both pass.
- Accessibility-tree exposure of the boxless listbox on Gecko and WebKit, which is what #7164 tracks.
- Whether moving the clipboard tests from a Chrome-only file to `test-browser.ts` introduces cross-engine flake. Clipboard access is historically engine-sensitive, and this expands nine paste tests to Firefox and WebKit. One green run does not settle it, and I could not measure flake rates, so this is uncertainty rather than a finding. Worth watching on the next few runs; if it does become noisy, splitting the new structural tests into `test-browser.ts` and leaving the clipboard tests in `test-chrome.ts` recovers the cross-engine coverage that this change actually needed.

app/src/sandbox/combobox-select-tag-tooltip-element-id/index.react.tsx
diegohaz: This is just a sandbox.

app/src/sandbox/tag-combobox/tag.react.tsx
```diff
@@ -2,37 +2,52 @@ import { clsx } from "clsx";
...
 
-export const TagList = React.forwardRef<HTMLDivElement, TagListProps>(
-  function TagList(
+export const TagField = React.forwardRef<HTMLDivElement, TagFieldProps>(
+  function TagField(
     { label, value, setValue, values, setValues, ...props },
     ref,
   ) {
+    // Uses the store prop instead of TagProvider, so the tag list and the tag
+    // input inherit the store from TagControl.
```
diegohaz: This doesn't make sense. TagProvider should work as well.

app/src/sandbox/tag/test-browser.ts
```diff
@@ -18,13 +17,40 @@ async function tags(page: Page) {
...
+    await expect(q.textbox("Tags")).toBeVisible();
+  });
+
+  test("focuses the input when clicking on the control", async ({
+    page,
+    q,
+  }) => {
+    // The tag list has a display: contents style and generates no box, so the
+    // control is the element that receives clicks on the field's padding.
+    const control = page.locator(".ak-tag-list");
```
diegohaz: This is not right. We should avoid locator and we should never use class names like this. Can't we get the common ancestor of the tag list and control elements instead?

I wonder if `TagControl` should have a role by default, and if we should rename `TagListLabel` to `TagLabel` so it can be used as a label for `TagControl`, `TagList` and `TagInput`?

Can we explore this path? If this makes sense, we should do this in this PR instead of a follow-up.

app/src/sandbox/tag-combobox/tag.react.tsx
diegohaz: Fixed in [`089e985`](https://github.com/ariakit/ariakit/commit/089e9853ceccaaa0b6dbfba50c81701e557c0a02): the sandbox creates the store with `TagProvider` again, the way it does on `main`.

You're right, and the comment was wrong rather than just unclear. `TagProvider` supplies the store to `TagControl` and to both of its children just as well, so passing the store explicitly only changed which component receives it, and it also forced a `store` prop onto the label. Nothing about the sibling structure needed it.

`TagControl` still hands the store down to `TagList` and `TagInput`, so the `store` prop stays supported for consumers who don't use `TagProvider`. No sandbox exercises that path now, which is also how `main` was: no fixture there passed a tag `store` prop either.

app/src/sandbox/tag/test-browser.ts
diegohaz: Fixed in [`20236c9`](https://github.com/ariakit/ariakit/commit/20236c93cdd1770ed1ea28e78266afaef44bccb5): explored it, both parts hold up, so they're in this PR rather than a follow-up.

`TagControl` now renders `role="group"` with `aria-labelledby` from the label, using the same prop shape `useTagList` and `useGroup` already use, so a consumer `role`, `aria-labelledby` or `aria-label` still wins. That gives the test the accessible ancestor you were asking for: `q.group("Tags")` replaces `page.locator(".ak-tag-list")`, and it resolves to the same element, so the click geometry is unchanged.

`TagListLabel` is renamed to `TagLabel`, since one label now names all three. Chrome's tree for the tag sandbox, measured with CDP:

```
group "Tags"
  listbox "Tags"
    option "JavaScript"
    option "React"
  textbox "Tags"
```

One thing worth deciding explicitly: the name is now on all three nodes, so entering the field announces "Tags" up to three times. I think that's the right trade, because each name is individually correct and naming only some of them leaves an unnamed node mid-traversal, but it follows from one label naming three elements rather than from the group role itself. `aria-label` or `aria-labelledby` on `TagControl` opts out.

Two notes on the rename. It's breaking for `@ariakit/react-components/tag/tag-list-label`, so it has its own changeset; I didn't add a deprecated alias, since these components are experimental and only reachable through deep subpaths. And I added `TagControl` to the `aria-labelledby-aria-label` sandbox, which is where each component's "an explicit `aria-label` wins over the label" guard is pinned. Without that guard `aria-labelledby` wins and the consumer's name silently disappears, which is what the new case red-checks.

diegohaz:
**Fixed**

- [`089e985`](https://github.com/ariakit/ariakit/commit/089e9853ceccaaa0b6dbfba50c81701e557c0a02) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809862292)): the tag combobox sandbox creates its store with `TagProvider` again, like it does on `main`. The comment claiming the `store` prop was needed for the children to inherit it is gone, because it wasn't.
- [`20236c9`](https://github.com/ariakit/ariakit/commit/20236c93cdd1770ed1ea28e78266afaef44bccb5) ([reply](https://github.com/ariakit/ariakit/pull/7158#discussion_r3809862534)): `TagControl` renders `role="group"` labeled by the tag label, so the browser test queries `q.group("Tags")` instead of `page.locator(".ak-tag-list")`. `TagListLabel` is renamed to `TagLabel`, since one label now names the group, the listbox and the input. `TagControl` was also added to the `aria-labelledby-aria-label` sandbox, where each component's "an explicit `aria-label` wins over the label" guard is pinned.

**Notes**

Chrome exposes the tag sandbox as `group "Tags"` containing `listbox "Tags"` with the two options, and `textbox "Tags"` as a sibling of the listbox. The name lands on all three nodes, so entering the field announces it up to three times. That follows from one label naming three elements rather than from the group role itself, and passing `aria-label` or `aria-labelledby` to `TagControl` opts out.

The rename is breaking for `@ariakit/react-components/tag/tag-list-label`, so it has its own changeset. There's no deprecated alias, since these components are experimental and reachable only through deep subpaths.

Each new behavior is red-checked: removing `role="group"` fails the new happy-dom test, removing `aria-labelledby` fails the element-id browser test, and dropping the `aria-label` guard fails the new fixture case, because `aria-labelledby` then wins over the consumer's own name.

Two claims in the description were corrected rather than left standing. Removing `TagControl`'s context provider no longer fails the `tag-combobox` tests, now that the sandbox uses `TagProvider`, and no fixture passes a tag `store` prop at all, which matches `main`.

**Evidence gaps**

`pnpm test-react18` does not run in this environment. Its harness fails while installing into a temporary workspace, and it fails the same way on `sandbox/checkbox-interactions`, which this change does not touch, so it is environmental rather than caused by this change. CI covers that suite.

app/src/sandbox/tag-combobox/tag.react.tsx
```diff
@@ -23,11 +23,11 @@ export const TagList = React.forwardRef<HTMLDivElement, TagListProps>(
...
       >
         {label && (
-          <Ariakit.TagListLabel className="ak-tag-list-label">
+          <Ariakit.TagLabel className="ak-tag-list-label">
             {label}
-          </Ariakit.TagListLabel>
+          </Ariakit.TagLabel>
         )}
-        <Ariakit.TagList
+        <Ariakit.TagControl
```
diegohaz, line 30: Moving this fixture to `TagProvider` took away the only coverage of `TagControl`'s store-provision responsibility. Nothing in the repository passes a `store` prop to a tag component any more, so both halves of that responsibility can now be deleted with the suite green.

Two mutations, both run locally on this head, `pnpm test app/src/sandbox/tag app/src/sandbox/combobox-select-tag app/src/sandbox/aria-labelledby`:

- Deleting the entire `useWrapElement` block that renders `TagContextProvider` in `useTagControl`: 5 files, 28 tests, all passing.
- Making `useTagControl` ignore its `store` prop, by replacing `store = store || context` with `store = context`: same 28 tests, all passing.

The first one is silent precisely because every fixture now sits under a `TagProvider` that supplies the same context from above, so the control providing it again is redundant in tests and load-bearing only in production code that follows the documented shape.

For contrast, the same `store = context` mutation applied at `dc19ff31`, the previous head, fails all 5 `tag-combobox` tests on `TagControl must receive a store prop or be wrapped in a TagProvider component`. That coverage existed one commit ago.

This matters because the `store` prop is the only way to use the tag components without `TagProvider`, and this release documents it twice: the changeset says to pass it to `TagControl`, and the `useTagStore` JSDoc in `packages/ariakit-react-components/src/tag/tag-store.ts` shows exactly that shape. A reviewer on this pull request already doubted that path works. It does, but nothing proves it any more.

The Verification section of the pull request description is now stale for the same reason. It states that removing the context provider fails all five `tag-combobox` tests, and that no longer reproduces at this head.

One fixture on the documented shape restores both halves at once, because without a `TagProvider` the list and the input can only reach the store through the control:

```tsx
const tag = Ariakit.useTagStore({ value, setValue, values, setValues });
return (
  <>
    {label && (
      <Ariakit.TagLabel store={tag} className="ak-tag-list-label">
        {label}
      </Ariakit.TagLabel>
    )}
    <Ariakit.TagControl ref={ref} store={tag} {...props} className={/* ... */} />
  </>
);
```

If you would rather keep this sandbox on `TagProvider`, a small dedicated fixture covers it just as well. Either way, dropping `store` from `TagFieldProps` keeps the shape that was raised on this file earlier from coming back.

**Assessment**

- Severity: Medium (estimate, 80% confidence), because a documented public contract of a component this pull request introduces can break with the whole suite green.
- Priority: P1 (estimate, 70% confidence), because this pull request creates the gap and its description still cites the coverage that no longer exists.
- Breakage likelihood: Possible (estimate, 70% confidence), because it takes a later edit to `useTagControl`'s store resolution, which is likely while a new component settles.
- Fix effort: Easy (verified, 90% confidence), because it is one fixture, and the previous revision of this file already had the shape.

packages/ariakit-react-components/package.json
```diff
@@ -255,9 +255,10 @@
     "./tab/tab-store": "./src/tab/tab-store.ts",
     "./tag/tag": "./src/tag/tag.tsx",
     "./tag/tag-context": "./src/tag/tag-context.tsx",
+    "./tag/tag-control": "./src/tag/tag-control.tsx",
     "./tag/tag-input": "./src/tag/tag-input.tsx",
+    "./tag/tag-label": "./src/tag/tag-label.tsx",
```
diegohaz, line 260: The rename drops the `./tag/tag-list-label` subpath entirely, so an existing import fails at module resolution rather than with a migration message. That diverges from how this package has handled its other component renames.

Both precedents in this same exports map keep the old name working:

- `MenuBar` to `Menubar`: `"./menu/menu-bar"` is still exported at line 185 of this file, and `packages/ariakit-react-components/src/menu/menu-bar.tsx:48` carries `@deprecated` with `Use [` + "`Menubar`" + `](https://ariakit.com/reference/menubar) instead.`
- `FormField` to `FormControl`: `packages/ariakit-react-components/src/form/form-field.tsx:52` carries `@deprecated` with `This component has been renamed to [` + "`FormControl`" + `](https://ariakit.com/reference/form-control). The API remains the same.`

That is the soft-deprecation shape `ariakit-ariakit-deprecation` describes, and `TagListLabel` is a shipped API rather than a branch-new one, with a released fix recorded at `packages/ariakit-react-components/CHANGELOG.md:487`.

The consumer this actually reaches is narrow but real: a read-only tag field that renders `TagListLabel`, `TagList`, and `Tag` with no `TagInput`. Nothing else in this release breaks that markup, since there is no input to move out of the list and no reason to add a `TagControl`. The rename is the only thing that breaks it, and it breaks with `Cannot find module` instead of pointing at `TagLabel`.

The counterweight is real too, so this is a decision rather than a defect: the tag components are experimental, and every consumer that renders a `TagInput` has to restructure regardless.

Both outcomes are fine, but the choice should be deliberate rather than implicit:

- Keep the clean break, and say so in the changeset, so it reads as policy for experimental components rather than an oversight against two visible precedents.
- Or follow the precedent: re-export `TagListLabel`, `useTagListLabel`, `TagListLabelOptions`, and `TagListLabelProps` from a `tag-list-label.tsx` module with `@deprecated` tags pointing at `TagLabel`, keep the subpath, and add the `290-` deprecation changeset that `ariakit-general-changeset` asks for.

**Assessment**

- Severity: Low (estimate, 75% confidence), because it is a build-time failure with a documented migration, not a runtime defect.
- Priority: P2 (estimate, 75% confidence), because either decision is defensible and only the unstated divergence from precedent is the problem.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because it needs a consumer of an experimental component using the label without a tag input.
- Fix effort: Easy (estimate, 85% confidence), because it is either one sentence in the changeset or a small alias module plus a deprecation changeset.

diegohaz (commented):
Re-review. The baseline is the head recorded in the previous marked review, `dc19ff31`, through the current head `20236c93`, plus the surrounding code the new commits touch. Two commits since then: `089e9853` moved the tag combobox sandbox back to `TagProvider`, and `20236c93` renamed `TagListLabel` to `TagLabel` and gave `TagControl` a `group` role named by that label.

The group role is a good addition and it is well covered. `app/src/sandbox/aria-labelledby-aria-label` exercises the `aria-label` branch, `app/src/sandbox/combobox-select-tag-tooltip-element-id` exercises the label-id update path on the control, and the browser click test now targets `q.group("Tags")` instead of a class locator, which also asserts the role and the name where it matters.

Two new findings, both introduced by the commits since the last review.

## Published inline

- https://github.com/ariakit/ariakit/pull/7158#discussion_r3809937619 — `TagControl`'s store provision lost its only coverage when this sandbox moved back to `TagProvider`. Nothing in the repository passes a `store` prop to a tag component any more, so the control can stop providing the store, or ignore its `store` prop, with the whole suite green. Verified with mutation runs across both heads. The Verification section of the description is stale for the same reason.
- https://github.com/ariakit/ariakit/pull/7158#discussion_r3809938861 — the rename drops the `./tag/tag-list-label` subpath outright, against two visible precedents in the same exports map that kept a deprecated alias. Either decision is defensible for an experimental component; the ask is to make it explicit rather than implicit.

## What I ran

- `pnpm tsc` on this head: passes.
- `pnpm test app/src/sandbox/tag app/src/sandbox/combobox-select-tag app/src/sandbox/aria-labelledby`: 5 files, 28 tests, all passing.
- Three mutation experiments, all restored afterwards and the tree verified clean:
  - Swapping `TagControl` from `TagContextProvider` to `TagScopedContextProvider` fails all 28 tests, because the scoped provider also sets the unscoped context and `useProviderContext` then returns `undefined` for `TagList`. That distinction is caught, which is worth recording since it was a candidate risk.
  - Deleting `TagControl`'s context provider entirely: 28 tests still pass.
  - Making `useTagControl` ignore its `store` prop: 28 tests still pass on this head, and the same mutation fails all 5 `tag-combobox` tests at `dc19ff31`.
- CI on this head: every check passing, including both Safari shards, which were still running when I started.
- Changeset mechanics for `110-tag-label.md`: `110-` orders it after the `100-` major entry, `minor` frontmatter matches the v0 rule for a breaking change, the title is a short phrase without a period, both `Before:` and `After:` blocks are present, the renamed hook and types are named, and the API names stay plain inline code, which is right while `https://ariakit.com/reference/tag-label` does not exist.
- Checked for leftovers: no code references to `TagListLabel` or `tag-list-label` remain outside the historical `CHANGELOG.md` entry.

## Threads

Every thread on this pull request was resolved before this run, including the three I opened last time. Section 3 excludes resolved threads from candidacy, so there was nothing to re-validate and I have not re-raised any of them.

Worth recording plainly: the `display: contents` divergence I reported last time was resolved without a code change, with the reply `This is just a sandbox.` That is your call and I am not reopening it. The reasoning still lives only in https://github.com/ariakit/ariakit/issues/7160.

- **`TagLabel` as a name.** I checked it against the convention where `XLabel` labels `X`, which would read as a label for the `Tag` component. It is consistent with `ComboboxLabel`, which labels the combobox widget rather than a child, and the collision with the `Tag` component is the same one `TagInput` already carries. Not a finding.
- **Naming the group, the listbox and the input with the same text.** Assistive technologies will now encounter `Invitees` three times when entering the field. Whether that is helpful grouping or verbosity is not something this repository can measure, and the alternative of leaving the group unnamed makes the group pointless. Recorded as uncertainty, not a finding.
- **`useAttribute(labelElement, "id")` now runs in both `useTagControl` and `useTagList`.** Two observers on one label element. The cost is negligible and the alternative would couple the two components, so this is the right shape.
- **No happy-dom duplicate for the new `aria-labelledby-aria-label` case.** That sandbox is browser-only by design, with a `test-chrome.ts` and no `test.ts`, so the new case follows its siblings.
- **Whether the rename needs a deprecation.** Published inline rather than suppressed here, because the precedent is concrete and the decision is yours to state.

## Still open from the previous review

https://github.com/ariakit/ariakit/issues/7164 stands unchanged: accessibility-tree exposure of a boxless element carrying `role="listbox"` is confirmed in Chrome only. The new `group` element does not affect it, since `TagControl` always generates a box.

## Not fully validated

- Playwright suites were not run locally; the green CI run on this head is the evidence.
- `pnpm build` and the packed contents were not run locally. The exports map change is a rename of one plain source path, and CI's package build and Continuous Releases both pass.
- The accessibility-tree question tracked by #7164, and the verbosity of the three-way labeling described above.

packages/ariakit-react-components/package.json
diegohaz: This is an experimental component.

PR #7191, opened 2026-08-19
Wait for preview frameworks to commit before browser tests

### LINKED ISSUE #7186: Preview hydration wait resolves before the framework commits
## Problem

`withFramework` in `app/src/test-utils/preview.ts` waits for the preview island to lose its `ssr` attribute before every browser test:

```ts
await page.waitForFunction(() => !document.querySelector("astro-island[ssr]"));
```

That attribute marks hydration as **scheduled**, not committed. `@astrojs/react`'s client entry calls `hydrateRoot` inside `startTransition` and returns immediately:

```js
startTransition(() => {
  const root = getOrCreateRoot(element, () => hydrateRoot(element, componentEl, renderOptions));
  root.render(componentEl);
});
```

`astro-island` then awaits that hydrator, removes `ssr`, and dispatches `astro:hydrate`, all before React has rendered anything.

## Evidence

Measured on `/react/previews/combobox-select-selected-value-presentation/` with a `MutationObserver` installed through `page.addInitScript`, against the Astro dev server:

| Engine | `ssr` removed | First post-hydration render | Gap |
| --- | --- | --- | --- |
| Chrome | 393 ms | 651 ms | 258 ms (16 frames) |
| Firefox | 452 ms | 826 ms | 374 ms (7 frames) |
| Safari | 523 ms | 747 ms | 224 ms (7 frames) |

Until that render lands, the preview still shows server-rendered markup and React has committed no state. In the sandbox above the item is `<div role="option" id="..." data-render-count="1">Apple</div>` before the commit and `<div role="option" tabindex="-1" id="..." data-render-count="2" aria-selected="false">Apple</div>` after it.

The gap is only hidden locally because `gotoAndSettle`'s bounded network-idle wait usually takes longer than hydration. Under CI contention that ordering inverts.

## Impact

Two shapes of test are affected.

Tests that sample or negatively assert DOM state before their first interaction can read pre-hydration markup. #7184 is a confirmed instance: `does not re-render inactive items when the popup opens` sampled the server-rendered render count and then reported the settled count as a re-render, failing roughly 3% of `app` workflow runs on `main` and exhausting all three Firefox attempts. `app/src/sandbox/react-components-undefined-props/test-browser.ts` asserts an absence as its first action after `beforeEach` and is a candidate for the same problem, though it has not been confirmed.

Tests whose first action is a click are largely insulated, because React attaches its listeners to the container inside `hydrateRoot` and replays discrete events. They are not immune though: the two tests reported as flaky (passing on retry) in the same file, `uses the current disclosure in delayed auto-focus` in Firefox and `focuses a far item reached through typeahead` in Safari, both start with a click on a server-rendered control.

## Prior art

#7107 added the current wait for exactly this failure mode. It reduced the rate but could not eliminate it, because no framework-agnostic committed-hydration signal exists to wait on: `astro:hydrate` and the `ssr` removal both fire when the hydrator returns, and React writes its container key (`__reactContainer$...`) synchronously inside `hydrateRoot`.

#7184 fixed the one confirmed failure at the test level, by asserting the settled render count instead of sampling a value after a fixed number of frames. That does not generalize.

## Expected outcome

`withFramework`'s `beforeEach` waits on a marker that only exists after the framework commits, so no test can observe pre-hydration markup.

The natural place is `generatePreviewWrapper` in `app/src/lib/preview-codegen.ts`, which already generates the `client:load` island wrapper per framework. A tiny per-framework component that sets a marker in a mount effect would give `beforeEach` a real completion signal.

Audit `app/src/**` browser tests that sample or negatively assert DOM state before their first interaction once the marker exists, and remove any frame-count waits that were only compensating for this gap.

diegohaz:
## Motivation

Closes #7186.

Astro removes the `ssr` attribute before React or Solid has necessarily committed its scheduled hydration work. As a result, `withFramework` could return while the preview still had server-rendered state.

```ts
await page.waitForFunction(
  () => !document.querySelector("astro-island[ssr]"),
);
```

## Solution

Generate a small client wrapper for each framework preview. The wrapper marks the document from a framework mount effect after the framework commit completes.

```ts
useEffect(markPreviewHydrated, []);
```

Select wrapper generation through a framework adapter registry. Each adapter owns its generated file extension and content generator, and generated artifacts use names such as `preview.client.react.tsx`. A future renderer can add its own adapter and file syntax without changing central filename or cleanup logic.

Make `withFramework` wait for the shared marker for Astro previews. Keep JavaScript-disabled and Next.js preview behavior unchanged. Cover React and Solid wrapper generation and cleanup in the preview integration tests, exercise the hydration wait through the existing browser suites, and remove the redundant local `ssr` waits found in the browser-test audit.

## Verification

- `pnpm lint`
- `pnpm tsc`
- `pnpm build-app-lite`
- `pnpm test app/src/lib/preview-integration.test.ts`
- Existing preview suites in focused Chrome, Safari, and Firefox runs
- Ariakit Tailwind Chrome suite with retries disabled


app/src/sandbox/preview-hydration-7186/index.react.tsx
```diff
@@ -0,0 +1,11 @@
+export default function Example() {
```
diegohaz: I don't think we need this sandbox. The hydration mechanism is already tested with the other passing tests.

app/src/lib/preview-hydration.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+export const PREVIEW_HYDRATED_ATTRIBUTE = "data-preview-hydrated";
```
diegohaz: Instead of exporting a const, we could export a function `isPreviewHydrated` (or something like that)?

app/src/sandbox/preview-hydration-7186/index.react.tsx
diegohaz: Fixed in [`d04cb09d7`](https://github.com/ariakit/ariakit/commit/d04cb09d7cf5d7ab13cd1e60af1c06045a1326cc): removed the synthetic hydration sandbox and kept the existing preview browser suites as integration coverage.

app/src/lib/preview-hydration.ts
diegohaz: Fixed in [`d04cb09d7`](https://github.com/ariakit/ariakit/commit/d04cb09d7cf5d7ab13cd1e60af1c06045a1326cc): replaced the exported attribute constant with the serializable `isPreviewHydrated` predicate and passed it directly to `page.waitForFunction`.

diegohaz:
**Fixed**

- [`d04cb09d7`](https://github.com/ariakit/ariakit/commit/d04cb09d7cf5d7ab13cd1e60af1c06045a1326cc) ([sandbox reply](https://github.com/ariakit/ariakit/pull/7191#discussion_r3813426799)): Removed the synthetic hydration sandbox and retained the existing browser suites as integration coverage.
- [`d04cb09d7`](https://github.com/ariakit/ariakit/commit/d04cb09d7cf5d7ab13cd1e60af1c06045a1326cc) ([predicate reply](https://github.com/ariakit/ariakit/pull/7191#discussion_r3813430406)): Encapsulated readiness in `isPreviewHydrated` and used it directly from Playwright.

Issue #7171, opened 2026-08-19
fireClickEvent dispatches synthetic clicks as MouseEvent, but browsers dispatch them as PointerEvent


diegohaz:
## Context

Found while fixing #7162, which makes `@ariakit/test` dispatch `click`, `auxclick`, and `contextmenu` as `PointerEvent`. That fix covers the test helpers only. The same gap exists in shipped library code, where it reaches real applications rather than tests.

## Problem

`fireClickEvent` in `packages/ariakit-utils/src/events.ts` builds a `MouseEvent`:

```ts
export function fireClickEvent(element: Element, eventInit?: PointerEventInit) {
  const event = new MouseEvent("click", eventInit);
  return element.dispatchEvent(event);
}
```

The parameter is typed `PointerEventInit`, but `MouseEvent`'s constructor ignores the pointer members, so a caller passing `pointerId` or `pointerType` has them silently dropped.

`Command` in `packages/ariakit-react-components/src/command/command.tsx` calls this to synthesize the click for `Enter` and `Space` activation on elements that don't click natively, so the event reaches every component built on `Command`, such as [`Button`](https://ariakit.com/reference/button).

Measured with Playwright on Chromium 151, Firefox 153, and WebKit 26.5, a click that no pointer caused is still a `PointerEvent` in every engine:

| Case | Interface | `pointerId` | `pointerType` | `isTrusted` |
| --- | --- | --- | --- | --- |
| `Enter` on a button | `PointerEvent` | `-1` | `""` | `true` |
| `Space` on a button | `PointerEvent` | `-1` | `""` | `true` |
| `element.click()` | `PointerEvent` | `-1` | `""` | `false` |

Ariakit's synthetic click is a `MouseEvent`, so consumer code that checks `event instanceof PointerEvent` or reads `event.pointerType` in a click handler sees a different event shape when a user activates an Ariakit component with the keyboard than when they activate a native `<button>` the same way. TypeScript types `click` as `PointerEvent` in `lib.dom`, so reading `event.pointerType` in that handler type-checks and then returns `undefined`.

The constructor is also taken from the ambient global rather than the element's own window, so the event belongs to the wrong realm for an element inside a same-origin iframe.

## Why this can wait

[...]

PR #7216, opened 2026-08-20
Match the synthetic Command click view and composition

### LINKED ISSUE #7192: The synthetic click behind keyboard activation reports view: null and composed: false
## Context

Found while fixing #7171, which makes `fireClickEvent` build the click [`Command`](https://ariakit.com/reference/command) synthesizes for `Enter` and `Space` as a `PointerEvent` from the element's own window. That covers the interface, `pointerId`, and `pointerType`, which is the member set #7171's expected outcome names.

## Problem

Two more members of that click still differ from the one browsers dispatch.

Measured with Playwright on Chromium 151, Firefox 153, and WebKit 26.5, comparing a native button activated with `Enter` against a non-native [`Command`](https://ariakit.com/reference/command) activated the same way:

| Member | Browsers | Ariakit's synthetic click |
| --- | --- | --- |
| `view` | the element's own window | `null` |
| `composed` | `true` | `false` |

`Command` strips `view` from the init before calling `fireClickEvent`:

```ts
const { view, ...eventInit } = event;
```

and React's synthetic event carries no `composed`, so neither member ever reaches the constructor.

`composed: false` is the consequential one. A non-composed event does not propagate out of a shadow root, so a listener outside a shadow tree never sees a keyboard activation that happened inside it, where the click from a native button would reach it.

`@ariakit/test` already sets `composed: true` on the click it fires for keyboard activation in `packages/ariakit-test/src/press.ts`, so the shipped library is the outlier here.

## Why this can wait

#7171 enumerates the members it covers, and its regression coverage asserts exactly those. `view` and `composed` are separate members with a separate consequence, and the shadow-boundary behavior needs its own browser coverage rather than riding on the pointer-member assertions.

## Expected outcome

The synthetic click reports the element's own window as `view` and `composed: true`, matching what Chromium, Firefox, and WebKit dispatch for a click no pointer caused.

diegohaz:
## Motivation

When [`Command`](https://ariakit.com/reference/command) renders a non-native element, keyboard activation dispatches a synthetic click with `view: null` and `composed: false`. Native keyboard clicks use the element's owner window and cross open shadow boundaries.

Closes #7192.

## Solution

`fireClickEvent` now layers the target element's owner window onto `view` and sets `composed: true`. The existing Proxy still preserves caller-provided event members, supports live and frozen init objects, and resets pointer contact data to native click defaults.

The regression coverage verifies document, iframe, and shadow-root activation. It also renames the existing #7171 sandbox to `command-keyboard-click` so the related keyboard-click regressions use one fixture.

The shadow-root propagation assertion runs in the browser suite because happy-dom does not reliably reproduce this React portal path.

## Workaround

Keep `Command` on its default native button so the browser dispatches the keyboard click:

```tsx
<Ariakit.Command onClick={handleClick}>Action</Ariakit.Command>
```

This preserves the native click's owner-window `view` and composed propagation. It does not apply when the interface must render a non-native host.

## Preview

The final non-native `Command` activation reports the owner window, creates a composed click, and lets Enter and Space cross the shadow boundary twice:

Non-native Command keyboard activation reports the owner window and two clicks cross the shadow boundary

### DIFF
--- a/.changeset/300-command-click-metadata.md
+++ b/.changeset/300-command-click-metadata.md
@@ -1,0 +1,7 @@
+---
+"@ariakit/utils": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed `fireClickEvent` so clicks expose the target element's owner window through `view` and are composed. This also fixed keyboard activation on non-native [`Command`](https://ariakit.com/reference/command) elements and components built on [`Command`](https://ariakit.com/reference/command), such as [`Button`](https://ariakit.com/reference/button).
--- a/.changeset/300-command-click-metadata.md
+++ b/.changeset/300-command-click-metadata.md
@@ -1,7 +1,6 @@
 ---
-"@ariakit/utils": patch
 "@ariakit/react-components": patch
 "@ariakit/react": patch
 ---
 
-Fixed `fireClickEvent` so clicks expose the target element's owner window through `view` and are composed. This also fixed keyboard activation on non-native [`Command`](https://ariakit.com/reference/command) elements and components built on [`Command`](https://ariakit.com/reference/command), such as [`Button`](https://ariakit.com/reference/button).
+Fixed keyboard activation on non-native [`Command`](https://ariakit.com/reference/command) elements and components built on [`Command`](https://ariakit.com/reference/command), such as [`Button`](https://ariakit.com/reference/button), so their synthetic clicks use the element's owner window for `view` and are composed.
--- a/.changeset/300-fire-click-event-metadata.md
+++ b/.changeset/300-fire-click-event-metadata.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Fixed `fireClickEvent` so clicks use the target element's owner window for `view` and are composed.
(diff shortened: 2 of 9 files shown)


.changeset/300-command-click-metadata.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ariakit/utils": patch
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed `fireClickEvent` so clicks expose the target element's owner window through `view` and are composed. This also fixed keyboard activation on non-native [`Command`](https://ariakit.com/reference/command) elements and components built on [`Command`](https://ariakit.com/reference/command), such as [`Button`](https://ariakit.com/reference/button).
```
diegohaz: We should probably separate utils (fireClickEvent) from components/react into two changesets.

diegohaz: Fixed in [`edfec2a38`](https://github.com/ariakit/ariakit/commit/edfec2a389c26e4ad576a6c2023c551598213f05): split the utility and React release notes into separate changesets.

diegohaz:
**Fixed**

- [`edfec2a38`](https://github.com/ariakit/ariakit/commit/edfec2a389c26e4ad576a6c2023c551598213f05) ([review thread](https://github.com/ariakit/ariakit/pull/7216#discussion_r3818828980)): Split the utility and React release notes into separate changesets.

PR #7217, opened 2026-08-20
Resolve named-property collisions in DOM helpers

### LINKED ISSUE #7215: isElement, isNode, and the realm lookups in getDocument and getWindow still read a member a named element can answer
## Context

Found while reviewing #7213, which fixes the realm helpers in `packages/ariakit-utils/src/dom.ts`. That pull request hardened `isDocument` by reading `nodeType` through a getter taken from `Node.prototype`, because a document overrides its own built-ins with the elements it names. Two reads in the same module still go through the named-property lookup.

## Problem

`isElement` and `isNode` read the member directly:

```ts
export function isElement(target: EventTarget | null | undefined): target is Element {
  return (target as Node | null)?.nodeType === 1;
}

export function isNode(target: EventTarget | null | undefined): target is Node {
  return typeof (target as Node | null)?.nodeType === "number";
}
```

Measured on Chromium 151, Firefox 153, and WebKit 26.5, a `<form>` containing `<input name="nodeType">` answers `form.nodeType` with that control, so both guards report `false` for a form that is plainly an element and a node.

This is reachable. `packages/ariakit-react-components/src/dialog/utils/use-hide-on-interact-outside.ts` bails out on `if (!isElement(target)) return;`, so pressing on such a form outside an open [`Dialog`](https://ariakit.com/reference/dialog), [`Popover`](https://ariakit.com/reference/popover), or [`Menu`](https://ariakit.com/reference/menu) would leave it open. `isPortalEvent` in `packages/ariakit-utils/src/events.ts` treats that form's events as portal events for the same reason.

`getDocument` and `getWindow` have a milder version of the same gap. `getDocument` resolves the owning document with a direct read:

```ts
const nodeDocument = isWindow(node) ? node.document : (node as Node).ownerDocument;
```

A form holding `<input name="ownerDocument">` answers that read with the control. #7213 detects that the answer is not a document and degrades to the ambient one, which is right in the ambient realm and wrong for a form inside a frame. `getWindow` reads the view the same way:

```ts
const { defaultView } = nodeDocument;
```

Measured on the same three engines, a `<form name="defaultView">` makes a document answer that lookup with the form. #7213 detects that and degrades to the ambient window, which is correct in the ambient realm but wrong for an element inside a frame, where the frame's real view is the right answer.

## Why this can wait

All four reads are pre-existing, and none is a regression from #7213. Each needs a control or element named exactly after the member, which is markup nobody writes by accident.

The two guards also need their own consideration rather than a mechanical swap. `isElement` runs on every event a component observes, so replacing a property read with a function call and a `try`/`catch` is a hot path change that wants a measurement. Reworking `getWindow` would also invert the expectations of the two cases in `packages/ariakit-utils/src/dom.test.ts` that emulate the shadowing with an instance-level `Object.defineProperty`, since a prototype getter bypasses exactly that emulation.

## Expected outcome

`isElement`, `isNode`, and the document and view lookups in `getDocument` and `getWindow` resolve through a read that named properties cannot answer, the way `isDocument` already does, with the performance of `isElement` measured rather than assumed, and with the affected tests reworked so they still prove what they claim.

### Workaround

See the [confirmed workaround](https://github.com/ariakit/ariakit/issues/7215#issuecomment-5350361864) until #7217 is released.

diegohaz:
## Motivation

HTML forms and documents expose named elements as properties. A control named `nodeType`, `ownerDocument`, or `defaultView` can therefore replace the DOM member that Ariakit reads.

After keyboard interaction, this caused a pointer-focused `Focusable` form with an input named `nodeType` to keep its keyboard focus styling. The same collision could also make `getDocument` and `getWindow` resolve the wrong realm.

## Solution

Keep direct DOM property reads for the normal path. When a named element overrides one of those properties, repeat the read from the object's prototype:

```ts
Reflect.get(Object.getPrototypeOf(value), name, value);
```

This bypasses the named property and invokes the native getter with the original object. It also works with nodes from same-origin frames without caching realm-specific constructors. The existing `isDocument` implementation remains unchanged.

Add focused utility tests for the `nodeType`, `ownerDocument`, and `defaultView` collisions. Add a real-browser `Focusable` regression test for the user-visible focus behavior.

## Workaround

Before this fix is available, use a collision-free DOM field name and remap the submitted data to the backend-facing `nodeType` key:

```html
<input name="nodeTypeField" />
```

```ts
const data = new FormData(form);
data.set("nodeType", data.get("nodeTypeField") ?? "");
data.delete("nodeTypeField");
```

This workaround requires control over both the DOM field name and the point where `FormData` is prepared.

## Preview

The focused browser suite passes 30/30 tests across Chromium, Firefox, and WebKit. A pointer-focused form containing `name="nodeType"` no longer receives keyboard focus styling.

The exact final source also has no significant change in a two-round local comparison against `main`:

- Combobox movement scripting: -21%
- Combobox select movement scripting: -14%
- Tab movement scripting: -23%

<details>
<summary>Final: bypass only the reported named properties</summary>

Earlier versions used a generic getter walker and added handling for import order, throwing proxies, and a numeric `Window.nodeType` global. Those cases increased code size and added work to hot paths beyond the scope of #7215.

The final implementation keeps normal reads unchanged and uses one prototype read only when the direct result has the named-element collision shape. It does not add special handling for unrelated proxy or numeric-window cases.

Paired local profiles identified the generic reader on the affected combobox stacks. The exact final unprofiled comparison is clean: -21% scripting for combobox movement, -14% for combobox-select movement, and -23% for tab movement.

</details>

Fixes #7215


packages/ariakit-utils/src/dom.ts
```diff
@@ -36,42 +34,76 @@ function isWindow(value: unknown): value is Window & typeof globalThis {
...
+    ownerDocumentGetter ??
+    getDOMGetter(Object.getPrototypeOf(node), "ownerDocument");
+  if (!getter) return node.ownerDocument;
+  return callDOMGetter(getter, node);
+}
+
+function getDefaultView(ownerDocument: Document) {
+  if (typeof window !== "undefined" && ownerDocument === window.document) {
+    return window;
+  }
```
diegohaz: This early return is not only a fast path, and nothing here says so. Under the repository's Vitest happy-dom environment, `document` carries an **own** `defaultView` that is the global `window`, while the `Document` interface getter answers with happy-dom's separate internal window. The two are not the same object, so this branch is what keeps `getWindow` returning the `window` that tests and app code already hold.

I verified it by deleting only these three lines and running the suite:

```
FAIL packages/ariakit-utils/src/dom.test.ts > getWindow handles a form with 'a control named self'
FAIL packages/ariakit-utils/src/dom.test.ts > getWindow handles a form with 'controls named self and document'
Tests  2 failed | 24 passed (26)
```

Both expect `getWindow(form)` to be `window` and get the internal window instead.

Every other branch in this block carries a comment explaining the constraint it encodes, so this one reads as a removable optimization next to the getter right below it. A short note keeps a later cleanup from silently changing which window object `getWindow` hands back in happy-dom based suites, including consumers' own suites.

```suggestion
  // Not only a fast path. A DOM emulation can define an own `defaultView` on
  // the document that points at the global `window` while the `Document`
  // getter answers with a separate internal window, so this pins the ambient
  // realm to the `window` callers already hold.
  if (typeof window !== "undefined" && ownerDocument === window.document) {
    return window;
  }
```

**Assessment**

- Severity: Low (estimate, 80% confidence), because the code is correct today and only a later edit turns it into a defect.
- Priority: P2 (estimate, 85% confidence), because it protects the branch rather than fixing current behavior.
- Breakage likelihood: Possible (verified, 90% confidence), because removing the branch fails two existing tests today, and an automated reviewer has already proposed rewriting this function.
- Fix effort: Easy (verified, 95% confidence), because it is a comment.

packages/ariakit-utils/src/dom.ts
```diff
@@ -36,42 +34,76 @@ function isWindow(value: unknown): value is Window & typeof globalThis {
...
-    return view.document === ownerDocument;
+    return getter.call(value);
   } catch {
-    // A cross-origin window answers only an allowlist, which `document` is not
-    // on, so refusing to answer is itself proof that it owns another document.
-    return false;
+    // The brand check rejects anything that is not a node.
+    return undefined;
   }
 }
```
diegohaz: `getNodeType` repeats `callDOMGetter`'s `try`/`catch` verbatim, and the three readers introduced here repeat the same four steps: take the cached getter, otherwise walk the supplied object's prototype chain, fall back to a plain read when neither exists, then call the getter defensively. One helper covers all three, drops the `as Node` and `as object` casts, and lets `defaultView` reuse the module level cache the other two already have.

I applied this locally: `pnpm test packages/ariakit-utils` passes 77/77, `pnpm tsc` reports no errors, and `pnpm lint` is clean.

```ts
const nodePrototype = typeof Node === "undefined" ? undefined : Node.prototype;
const documentPrototype =
  typeof Document === "undefined" ? undefined : Document.prototype;

const nodeTypeGetter = getDOMGetter(nodePrototype, "nodeType");
const ownerDocumentGetter = getDOMGetter(nodePrototype, "ownerDocument");
const defaultViewGetter = getDOMGetter(documentPrototype, "defaultView");

// The cached getter is missing only when this module is evaluated before the
// browser globals exist, so resolve one from the supplied object instead. The
// plain read is the last resort, and it is the one a named element can answer.
function readDOMMember(
  value: object,
  cachedGetter: DOMGetter | undefined,
  name: string,
) {
  const getter =
    cachedGetter ?? getDOMGetter(Object.getPrototypeOf(value), name);
  if (!getter) return Reflect.get(value, name);
  try {
    return getter.call(value);
  } catch {
    // The brand check rejects anything that does not implement the interface.
    return undefined;
  }
}

function getNodeType(value: unknown) {
  if (!value) return undefined;
  return readDOMMember(value, nodeTypeGetter, "nodeType");
}

function getOwnerDocument(node: Node) {
  return readDOMMember(node, ownerDocumentGetter, "ownerDocument");
}

function getDefaultView(ownerDocument: Document) {
  if (typeof window !== "undefined" && ownerDocument === window.document) {
    return window;
  }
  return readDOMMember(ownerDocument, defaultViewGetter, "defaultView");
}
```

Caching `defaultView` also removes the prototype walk `getWindow` currently pays on every call for a frame document. Keep the `window.document` early return in `getDefaultView` while doing it: dropping it fails two existing tests, as explained in the other comment on this file.

**Assessment**

- Severity: Low (estimate, 85% confidence), because behavior is unchanged and only readability and a small per-call cost are affected.
- Priority: P2 (estimate, 85% confidence), because the duplication ships with the new machinery and is cheapest to remove before it has callers to keep in sync.
- Breakage likelihood: Rare (estimate, 80% confidence), because the risk is three copies drifting apart in a later fix rather than a failure today.
- Fix effort: Easy (verified, 95% confidence), because the replacement above already passes the utils suite, the type check, and lint.

diegohaz (commented):
Initial review of the full diff from the merge base [`69eac27`](https://github.com/ariakit/ariakit/commit/69eac275612ee7d00359751a8118853ac82f6e92) to head [`a49ebe3`](https://github.com/ariakit/ariakit/commit/a49ebe326dab6d546f324c38df0bd529c4dc6d61), plus the surrounding call sites of the four helpers this changes.

The approach holds up. Reading through the interface getters is the only option that also covers `isElement` and `isNode`, because the result-validation trick used for `isDocument` on the base branch cannot tell a shadowing control apart from a real element: a control's own `nodeType` is `1`, so any "check what came back" guard would accept it. Dropping `ownsDocument` is correct now for the same reason the guard existed: the named getter could answer with an `<iframe name="defaultView">` window from another realm, and the `Document` getter cannot, so the only view it can return is the one that owns the document.

## What I ran

- `pnpm test packages/ariakit-utils/src/dom.test.ts`: 26 passed on this branch.
- Same suite with only `packages/ariakit-utils/src/dom.ts` reverted to the merge base: `isElement handles a form that answers nodeType with a control`, `isNode handles a form that answers nodeType with a control`, `getDocument ignores a form's named ownerDocument control`, `getWindow ignores a document's named defaultView form`, and `resolves supplied nodes when imported without DOM globals` all fail. The five new utility tests are proven regression coverage.
- Check runs on the reproduction commit [`594a62c`](https://github.com/ariakit/ariakit/commit/594a62c0e82aa482da255b0d028b7b323732b4a4) record `App / Test Chrome`, `App / Test Firefox`, and `App / Test Safari (2/2)` failing, so the browser reproduction is proven red on all three engines before the fix.
- `pnpm tsc`, `pnpm lint`, and `pnpm test packages/ariakit-utils` (77 passed) against the local prototype of the simplification proposed inline.
- Probed the happy-dom descriptors this code depends on: `nodeType`, `ownerDocument`, and `defaultView` are all prototype accessors there, so the cached-getter path is the one the suite exercises rather than the plain-read fallback.

## Findings published inline

Both are on `packages/ariakit-utils/src/dom.ts`: the undocumented early return in `getDefaultView`, which measurably carries behavior rather than only speed, and the three near-identical member readers that one helper can replace.

## Candidates validated and not published

- **Outside-interaction handlers are also affected.** `use-hide-on-interact-outside.ts` opens with `if (!isElement(target)) return;`, so a form that answers `nodeType` with a control looked like it would keep a popup open when clicked. Invalid: `getEventTargets` builds its target list with `event.composedPath().filter(isElement)`, and the form's ancestors still pass the filter, so the projected `rootTarget` stays outside the popup. A Popover fixture with a shadowing form outside it hides on click on the base branch and on this branch alike. The changeset scoping this to `Focusable` is accurate.
- **`Reflect.deleteProperty` results are unchecked in `resolves supplied nodes when imported without DOM globals`** (raised by the automated review). Accurate that the call reports failure by return value rather than by throwing, but the deletions do succeed in this environment: the same test fails on the base branch, which it could not do if the globals were still present. Declined as a non-goal for this PR, since the assertion would guard a hypothetical future environment rather than a current gap.
- **`type("element", q.textbox("Node type"))` in the sandbox's happy-dom test omits `.ensure(...)`.** `type` falls back to the focused element instead of failing on a missing target, so a null target would type into the "Add node editor" button and produce a confusing payload mismatch rather than a missing-element error. Declined as a non-goal: every existing `type(text, target)` call under `app/src/**` omits it too, so changing this one alone would not establish the convention, and the following assertion still fails loudly.
- **`await addEditor.focus()` before `await addEditor.press("Enter")`.** `locator.press()` focuses the element itself, so the explicit focus is redundant. Declined as a non-goal: it is one line in a test and it makes the keyboard-modality setup easier to read.
- **`min-height: 180px` and `justify-content: flex-end` in `style.css` carry the browser test's `position: { x: 4, y: 4 }` click.** The layout is what keeps that point on the form's own padding rather than on a child. Declined as a non-goal, though a short comment on the click position would help a later reader.

## Registered follow-up

https://github.com/ariakit/ariakit/issues/7218 covers the call sites that still read the shadowable members directly instead of going through these helpers: `getRootElement` in `portal.tsx` reads `doc.defaultView?.HTMLElement`, and `getSingleVerticalScrollport` in `combobox/__utils.ts` reads `element.ownerDocument.defaultView`. Both are present unchanged on `main` at [`da21014`](https://github.com/ariakit/ariakit/commit/da21014aecad5ab6ce1a9ea26175a2318660180e), so they are separable from this PR, which is scoped to the helpers themselves.

## Not fully validated

- The per-call timings in the description. The performance checks on this branch report no significant regressions, which is consistent with them, but I did not re-measure.
- The user-visible symptom behind issue 7218. Emulating the document named getter in the happy-dom suite also breaks `dom-accessibility-api`, which reads `document.defaultView` to resolve accessible names, so the queries fail before the Ariakit path runs. That reproduction needs a real browser fixture.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in [`726c3273b`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166): documented why the ambient-window branch preserves the exact window object when a DOM implementation exposes a distinct interface-getter result.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in [`726c3273b`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166): consolidated the three reads in `readDOMMember`, cached the ambient `defaultView` getter, and retained dynamic prototype lookup for imports without DOM globals. The fresh-import test now also removes `Document`.

diegohaz:
**Fixed**

- [`726c3273b`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166) ([thread reply](https://github.com/ariakit/ariakit/pull/7217#discussion_r3819546639)): Documented the ambient-window compatibility invariant.
- [`726c3273b`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166) ([thread reply](https://github.com/ariakit/ariakit/pull/7217#discussion_r3819549303)): Consolidated DOM member reads, cached `defaultView`, and retained fresh-import coverage.

packages/ariakit-utils/src/dom.ts
```diff
@@ -36,40 +34,74 @@ function isWindow(value: unknown): value is Window & typeof globalThis {
...
-    // on, so refusing to answer is itself proof that it owns another document.
-    return false;
+    // Web IDL getters reject values that do not implement their interface.
+    return undefined;
+  }
+}
+
+function getNodeType(value: unknown) {
+  return readDOMMember(value, nodeTypeGetter, "nodeType");
+}
```
diegohaz: The combobox "move across items" regression in the performance report is real, not noise, and the current report's `No significant performance changes detected` headline is a threshold artifact rather than a clean result.

I pulled the raw `perf-chrome-rounds-*` artifacts from both perf runs on this branch and recomputed the paired medians (my numbers reproduce the published ones exactly, so the method matches the comparator). Scripting, per round:

| Run | Test | Per-round delta | Raw support per round |
| --- | --- | --- | --- |
| [`a49ebe3`](https://github.com/ariakit/ariakit/commit/a49ebe326dab6d546f324c38df0bd529c4dc6d61) | combobox-perf > move across items | +12.2%, +13.9%, +16.8%, +15.6% | 96%, 100%, 100%, 100% |
| [`a49ebe3`](https://github.com/ariakit/ariakit/commit/a49ebe326dab6d546f324c38df0bd529c4dc6d61) | combobox-select-perf > move across items | +8.9%, +15.6%, +8.2% | 100%, 100%, 100% |
| [`726c327`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166) | combobox-perf > move across items | +7.8%, +8.9% | 84%, 100% |
| [`726c327`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166) | combobox-select-perf > move across items | +8.0%, +11.6% | 88%, 100% |

Eleven rounds across two independent runs, every one positive, with near-disjoint sample distributions. The earlier run promoted it to the top of the report as a confirmed change rather than a candidate:

```
| combobox-perf > react > move across items | 371.2ms → 427.5ms (+15%) :warning: | 510.5ms → 582.3ms (+14%) :warning: | 32ms → 32ms (+0%) | 882.6ms → 1002.4ms (+14%) :warning: |
```

The current run reads clean only because the magnitude landed at +8.3% and +9.8%, and `computeSignificance` gates on `Math.abs(percent) > THRESHOLD_PERCENT` with `THRESHOLD_PERCENT = 10`. Below that line a row cannot even become a candidate, so the direction and the raw support stop being reported.

Every other row of comparable size is noise by the same measurement: `menu-perf > move across items` scripting goes +14.3% in the first run and -4.0% in the second, and `composite-perf > move across items` disagrees with its own profiled variant in the same run (-5.6% unprofiled, +6.3% profiled). Only the two combobox rows hold their direction across both runs.

What I could not establish is the mechanism, and the obvious explanation does not survive arithmetic. I instrumented `readDOMMember` and drove the `combobox-perf` fixture in happy-dom: an `ArrowDown` costs about 75 `readDOMMember` calls and 73 getter calls, with zero brand-check throws. Across the 199 presses the measured interaction performs, that is roughly 14,500 getter calls, which cannot add 30 to 56ms of scripting at any plausible per-call cost. The earlier run also moved Rendering by +14%, and a JS accessor cannot add 70ms of layout and paint. So neither "the getters cost ~1ns each" nor "it is run-to-run noise" explains what the artifacts show, and I would not merge on either reading.

The cheapest thing to try is restoring the plain read as the fast path for the hottest helper. A named element can only answer with an element or a `RadioNodeList`, never with a number, so a numeric plain read is always the real `nodeType` and the guarded getter is only needed for the shadowed and non-node cases. That keeps every named-property test green while taking `isElement` and `isNode` off the getter for real nodes:

```suggestion
function getNodeType(value: unknown) {
  // A named element answers with an element or a `RadioNodeList`, never with a
  // number, so a numeric plain read is always the real `nodeType`. Only the
  // shadowed and non-node cases pay for the guarded getter below.
  const nodeType = (value as Node | null | undefined)?.nodeType;
  if (typeof nodeType === "number") return nodeType;
  return readDOMMember(value, nodeTypeGetter, "nodeType");
}
```

I applied that locally: `pnpm test packages/ariakit-utils` passes 78/78, the `focusable-click-focus` sandbox passes, `pnpm tsc` reports no errors, and `pnpm lint` is clean. Two caveats. It does not take `getDocument` and `getWindow` off the getter, because `ownerDocument` and `defaultView` answer with an object either way and the same type test cannot separate them. And it restores main's exposure to a cross-origin `WindowProxy` throwing on the plain read, which is unreachable from an event target in your own document but is a real difference from the current code. Most importantly, I cannot show it moves the benchmark without a perf run, so please treat it as the first experiment rather than the fix.

A profiled variant of `move across items`, the way `open combobox (script profile)` already exists in that harness, would say where the extra scripting actually goes.

**Assessment**

- Severity: Medium (estimate, 75% confidence), because it is a measurable slowdown on combobox keyboard navigation rather than a correctness defect, and 200 items is a realistic list size.
- Priority: P1 (estimate, 80% confidence), because the earlier run reported it as a confirmed regression and the current clean headline comes from the threshold, not from the evidence.
- Breakage likelihood: Likely (verified, 95% confidence), because all eleven measured rounds across two independent CI runs move the same way with 84 to 100% raw sample support.
- Fix effort: Moderate (estimate, 60% confidence), because the cause is not yet identified; the suggestion above is cheap, but confirming any fix needs a fresh perf run.

diegohaz (commented):
Re-review of [`726c327`](https://github.com/ariakit/ariakit/commit/726c3273b9187f09a9ce36447db121f62ab8d166), focused on the performance question. The branch was rebased since my last review, so the incremental baseline is the pre-rebase head [`a49ebe3`](https://github.com/ariakit/ariakit/commit/a49ebe326dab6d546f324c38df0bd529c4dc6d61) plus the new `Simplify DOM member reads` commit.

## Previous findings

Both threads from the last review are resolved and both fixes hold up on reading the new code: the ambient-window branch in `getDefaultView` now carries the constraint and a permalink, and the three readers are consolidated into `readDOMMember` with `defaultView` cached alongside the other two. The consolidation also picked up two guards mine did not have, rejecting `null` and non-object values before the prototype walk, which is a strict improvement over the old `if (!value)`. No new issues in that commit.

## The performance report is not clean

One new finding, published inline. Short version: the combobox rows are a real regression, and the rows that look similar in size are noise.

I pulled the raw `perf-chrome-rounds-*` artifacts from both perf runs on this branch rather than reading the medians, and recomputed the paired medians. My numbers reproduce the published ones exactly, so the method matches the comparator.

Across the two runs, `combobox-perf > move across items` and `combobox-select-perf > move across items` moved scripting up in **eleven rounds out of eleven**, with 84 to 100% raw sample support per round and near-disjoint distributions. The earlier run put it at the top of the report as a confirmed change, `371.2ms → 427.5ms (+15%)` scripting with matching flags on rendering and total. The current run reads `No significant performance changes detected` only because the magnitude landed at +8.3% and +9.8%, and the comparator gates on `Math.abs(percent) > 10`, so a row below that line cannot even be reported as a candidate.

By the same measurement, everything else of comparable size is noise. `menu-perf > move across items` goes +14.3% in the first run and -4.0% in the second. `composite-perf > move across items` disagrees with its own profiled variant inside the same run. Those are the rows to dismiss; the combobox pair is not one of them.

I could not identify the mechanism, and the obvious one does not hold. Instrumenting `readDOMMember` and driving the `combobox-perf` fixture in happy-dom gives about 73 getter calls per `ArrowDown` and no brand-check throws, so roughly 14,500 getter calls across the measured interaction. That cannot produce 30 to 56ms at any plausible per-call cost, and the earlier run's +14% rendering delta cannot come from a JS accessor at all. The inline comment carries the per-round table, a fast-path suggestion for `getNodeType` that I validated locally, and its caveats.

## Correcting my previous review

My first review said the performance checks reported no significant regressions. That was wrong, and I should have read the report instead of the check statuses: the run for [`a49ebe3`](https://github.com/ariakit/ariakit/commit/a49ebe326dab6d546f324c38df0bd529c4dc6d61) flagged `combobox-perf > move across items` as a confirmed regression on scripting, rendering, and total. The perf workflow does not fail the build, so a green check says nothing about the report.

## What I ran

- Recomputed paired medians, per-round agreement, and pooled pairwise raw support for all 27 Chrome rows in both runs, from the `perf-chrome-rounds-*` artifacts of runs [32336329903](https://github.com/ariakit/ariakit/actions/runs/32336329903) and [32344728543](https://github.com/ariakit/ariakit/actions/runs/32344728543).
- Recovered the earlier report from the `Performance / Post PR comment` job log, since the PR comment is updated in place.
- Confirmed the current perf run is fresh and complete for this exact revision: every Chrome job plus Node succeeded, and job manifests show balanced interleaved round ordering.
- Counted `readDOMMember` and getter calls per `ArrowDown` on the `combobox-perf` fixture with temporary instrumentation, since removed.
- `pnpm test packages/ariakit-utils` (78 passed), the `focusable-click-focus` sandbox, `pnpm tsc`, and `pnpm lint` against the local fast-path prototype.

## Not fully validated

- Whether the suggested fast path actually moves the benchmark. It is correct and keeps the named-property coverage green, but only a fresh perf run can judge it.
- Whether the regression is confined to combobox. The two combobox fixtures are the only ones that hold direction across both runs, but the harness only profiles a few tests, and `move across items` is not one of them.

packages/ariakit-utils/src/dom.ts
```diff
@@ -36,54 +34,109 @@ function isWindow(value: unknown): value is Window & typeof globalThis {
...
 
+function getNodeType(value: unknown) {
+  return readDOMMember(value, nodeTypeGetter, "nodeType");
+}
+
+function isDocument(value: unknown): value is Document {
+  // Node.DOCUMENT_NODE === 9.
+  const nodeType = (value as Node | null | undefined)?.nodeType;
+  return (
+    nodeType === 9 || (typeof nodeType === "object" && getNodeType(value) === 9)
```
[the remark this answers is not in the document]
diegohaz: Addressed in 36567272b by removing this PR change to isDocument. isDocument is now identical to main and retains the branded Node.prototype getter, so a numeric Window.nodeType global is not accepted. No additional Window guard or edge-case test is needed.

packages/ariakit-utils/src/dom.ts
```diff
@@ -36,40 +34,82 @@ function isWindow(value: unknown): value is Window & typeof globalThis {
...
+    if (typeof nodeType === "number") {
+      return nodeType;
+    }
   } catch {
-    // A cross-origin window answers only an allowlist, which `document` is not
-    // on, so refusing to answer is itself proof that it owns another document.
-    return false;
+    // A cross-origin WindowProxy can throw before the getter rejects it below.
+  }
+  return readDOMMember(value, nodeTypeGetter, "nodeType");
```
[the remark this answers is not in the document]
diegohaz: This no longer applies after 36567272b. The generic reader and no-global import fallback were removed. isElement and isNode retain main’s direct read for ordinary inputs and use the prototype only when an object-valued named element shadows nodeType. Arbitrary throwing proxies keep main’s behavior and are outside #7215.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in [`36567272b`](https://github.com/ariakit/ariakit/commit/36567272b306407cc656686719bf3b72e517f8c8) by simplifying the hot path instead of extending the generic reader. Ordinary `nodeType`, `ownerDocument`, and `defaultView` reads now stay direct; the prototype read runs only after the direct value proves that a named-property collision occurred. The PR no longer changes `isDocument` from `main`.

I measured the exact commit locally against the PR base in two interleaved rounds. Scripting for the three focused movement scenarios changed by -21% (combobox), -14% (combobox select), and -23% (tabs).

The [exact-head CI report](https://github.com/ariakit/ariakit/pull/7217#issuecomment-5351114787) is also clean. Scripting changed by -1%, -2%, and +2% respectively; total time changed by -2%, -3%, and +2%. Chrome and Node report no significant changes, and the complete exact-head CI run passed.

diegohaz:
Addressed the current review feedback in [`36567272b`](https://github.com/ariakit/ariakit/commit/36567272b306407cc656686719bf3b72e517f8c8).

- Replaced the generic DOM reader with a single prototype-read helper.
- Kept direct reads on the ordinary hot path and used the helper only after detecting a named-property collision.
- Restored `isDocument` to the implementation from `main`.
- Removed handling and tests for speculative throwing-proxy and numeric-window cases.

Verification:

- Utility tests: 25/25 passed.
- Named-collision browser tests: Chromium, Firefox, and WebKit passed.
- Lint, formatting, type checks, and the production app build passed.
- Local exact-base comparison: combobox -21%, combobox select -14%, and tabs -23% scripting.
- [Exact-head CI](https://github.com/ariakit/ariakit/actions/runs/32368012778) passed. Its [performance report](https://github.com/ariakit/ariakit/pull/7217#issuecomment-5351114787) found no significant Chrome or Node changes.

packages/ariakit-utils/src/dom.ts
```diff
@@ -59,6 +60,10 @@ function isDocument(value: unknown): value is Document {
   }
 }
 
+function getDOMProperty(value: object, name: string) {
+  return Reflect.get(Object.getPrototypeOf(value), name, value);
+}
```
diegohaz, line 67: This helper is the whole fix, and it carries no comment. Neither do the three call sites that gate on it. Two non-obvious invariants are now load-bearing and unstated.

The first is why starting the lookup at the prototype works at all. `Object.getPrototypeOf(value)` skips the named-property getter, which lives on the object itself, while the `Reflect.get` receiver keeps the native getter running against the original object. That receiver argument is also what makes the read answer for a node from a same-origin frame, since the getter comes from that frame's own prototype chain rather than a cached ambient one. Drop the receiver and the read silently starts answering for the prototype.

The second is why `typeof nodeType === "object"` is a sound collision detector, in `isElement` and `isNode`. A named-property getter can only answer with an element, a `RadioNodeList`, or a `WindowProxy`, so it can never produce a number. That is exactly what makes a numeric direct read always genuine and lets the ordinary path stay on the plain read. Nothing in the file says so, so the gate reads like a guess, and the next reader who wants to simplify it has no way to tell whether widening it to `!== "number"` or dropping it is safe.

This is also the one branch in this block that lost its explanation. The removed comment carried a third fact that still applies:

```
// Reading `nodeType` on a non-node target yields `undefined`. The numeric
// literal (Node.ELEMENT_NODE === 1) avoids referencing the `Node` global, so
// the guard stays safe even if it's ever evaluated during SSR.
```

`// Node.ELEMENT_NODE === 1.` keeps the value and drops the reason. Every other branch in this block still states the constraint it encodes, including `isWindow`, `nodeTypeGetter`, `ownsDocument`, and the `defaultView` check in `getWindow`, which is the convention this one now breaks.

I verified both invariants on Chromium 151.0.7922.34, Firefox 153.0, and WebKit 26.5 through Playwright 1.62.1, on a page holding `<form id="f1"><input name="nodeType"></form>`, `<form name="nodeType">`, and `<form name="defaultView">`. All three engines agree: `typeof form.nodeType` is `"object"` while the prototype read is `1`, the same holds for a form inside a same-origin iframe, and `window.nodeType` and `document.nodeType` are both object-valued while `isElement` and `isNode` still answer correctly for them.

```suggestion
// Reads `name` from the prototype so the named-property getter, which lives on
// the object itself, cannot answer. The `Reflect.get` receiver keeps the native
// getter running against `value`, which is also why this answers for a node
// belonging to a same-origin frame.
function getDOMProperty(value: object, name: string) {
  return Reflect.get(Object.getPrototypeOf(value), name, value);
}
```

And above the two guards, one line is enough:

```ts
// A named element answers with an element, a `RadioNodeList`, or a
// `WindowProxy`, never with a number, so an object-valued read is the
// collision and a numeric one is always the real `nodeType`. That is what
// keeps the ordinary path on the plain read.
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because the shipped behavior is correct and only a later edit turns the missing rationale into a defect.
- Priority: P2 (estimate, 80% confidence), because it is cheapest to state while the reasoning is still fresh, and this module has already attracted automated rewrite suggestions for these exact lines.
- Breakage likelihood: Possible (estimate, 70% confidence), because the gate is the kind of narrow condition a later simplification pass widens, and widening it costs the hot path the plain read this PR was reshaped to preserve.
- Fix effort: Easy (verified, 95% confidence), because it is comments only.

diegohaz (commented):
## Review summary

Initial review with this skill. There is no prior review carrying its marker, so the baseline is the merge base [`da21014`](https://github.com/ariakit/ariakit/commit/da21014aecad5ab6ce1a9ea26175a2318660180e) through head [`3656727`](https://github.com/ariakit/ariakit/commit/36567272b306407cc656686719bf3b72e517f8c8), plus the surrounding module, the `isElement` and `isNode` call sites in `@ariakit/react-components`, the sandbox fixture, and both changesets.

All five existing review threads are resolved, so none of them was a candidate finding for this run. The generic reader, the throwing-proxy case, and the numeric `Window.nodeType` case they discuss are gone from the current head, and `isDocument` is byte identical to `main`.

### What I ran

- `pnpm test packages/ariakit-utils/src/dom.test.ts`: 25/25 pass at head.
- Red check for all four new unit tests. With `packages/ariakit-utils/src/dom.ts` restored from `main` and the tests left at head, exactly the four new tests fail and the other 21 pass. Each fails for the stated reason, not incidentally.
- Red check for the new browser test. Same substitution, `pnpm run test-chrome src/sandbox/focusable-click-focus` against a dev server from this checkout: 9 passed, 1 failed, and the failure is `data-focus-visible` being present after the pointer click, with `toBeFocused` already satisfied on the line above. That rules out the vacuous shape where the form never receives focus. Restoring the head implementation returns 10/10.
- Independent cross-engine verification of the shipped implementation. I transcribed the final `getDOMProperty`, `isWindow`, `isDocument`, `ownsDocument`, `getDocument`, `getWindow`, `isElement`, and `isNode` into a page holding `<form id="f1"><input name="nodeType"><input name="ownerDocument"></form>`, `<form name="nodeType">`, `<form name="defaultView">`, and a same-origin iframe carrying the same shapes, then ran it on Chromium 151.0.7922.34, Firefox 153.0, and WebKit 26.5 through Playwright 1.62.1. All three engines agree on all twenty cases: the form is an element and a node, the frame form resolves to the frame's document and window, and `window` and `document` still answer correctly even though `window.nodeType` and `document.nodeType` are both object-valued on that page. The frame document carrying `<form name="defaultView">` still resolves to the frame's own window through the fallback.
- Verified the changeset frontmatter against the release surfaces, and that [`Focusable`](https://ariakit.com/reference/focusable) resolves rather than 404s.
- Read the exact-head performance report and the current check run. Chrome and Node report no significant changes, and `combobox-perf > react > move across items` scripting is 324.2ms to 321.6ms.

The approach itself holds up against the alternatives. A cached prototype getter, which is what `isDocument` uses, would have to be resolved per realm to answer for frame nodes; `Reflect.get` with the original receiver gets that for free. Gating the fallback on an object-valued read keeps the ordinary path on the plain read, which is what took the earlier combobox regression back out.

### Published inline

One finding, on `getDOMProperty`: the helper that implements the fix and the two guards that gate on it ship without the rationale that makes them reviewable, and `isElement` is the one branch in this block that loses its existing comment. Comments only.

### Discovered in this review, not published inline

**Registered follow-up: three reads in the same module still take the named-property answer.** Filed as [#7225](https://github.com/ariakit/ariakit/issues/7225). Measured on the same three engines: `<form name="activeElement">` makes `document.activeElement` answer with that form while the prototype read still answers with the focused button, so `getActiveElement` reports the wrong element and its `!activeElement?.nodeName` guard does not catch it; a `<form>` with `<input name="tagName">` makes `isButton` throw `TypeError: form.tagName.toLowerCase is not a function`; and a form shadowing `checkVisibility`, `offsetWidth`, `offsetHeight`, and `getClientRects` makes `isVisible` throw on `getClientRects`. All three are present unchanged on `main` at [`35439dd`](https://github.com/ariakit/ariakit/commit/35439ddfe), none is touched by this diff, and they are reached through different helpers, so this PR's guarantees do not depend on them. [#7218](https://github.com/ariakit/ariakit/issues/7218) covers the call sites outside `@ariakit/utils` but not these.

- Severity: High (verified, 90% confidence), because two of the three throw rather than degrade, and the third corrupts focus tracking.
- Priority: P2 (estimate, 85% confidence), because it is separable work on pre-existing code.
- Breakage likelihood: Rare (verified, 90% confidence), because each needs markup naming a control exactly after the member, the same precondition as this PR's own bug.
- Fix effort: Moderate (estimate, 70% confidence), because `isVisible` runs per item in the composite filters and wants the same measurement standard `isElement` got here.

**Address now, no diff anchor: the Preview section states a conclusion its own numbers contradict.** It reads "no significant change in a two-round local comparison against `main`" and then lists -21%, -14%, and -23% scripting. The comparator flags anything past 10%, so those magnitudes are not "no change", and per the performance workflow a local pair is exploratory and cannot support a published claim either way. The authoritative evidence is the exact-head report, which is genuinely clean at -1%, -2%, and +2% scripting, and it is currently reachable only from inside the collapsed reassessment block. Promoting that report into Preview and dropping the local numbers would make the section say what the evidence actually supports.

- Severity: Low (verified, 90% confidence), because the code is fine and only the description misreports.
- Priority: P2 (estimate, 80% confidence), because the description outlives the branch as the record of why this shape was chosen.
- Breakage likelihood: Likely (verified, 95% confidence), because the contradiction is in the current text.
- Fix effort: Easy (verified, 95% confidence), because it is an edit to two paragraphs.

**Address now, no diff anchor: Preview carries no visual evidence and does not state the gap.** The result being claimed is browser visible, so the browser-evidence policy asks for a screenshot or a short recording, and when one cannot be captured it asks for the gap to be named rather than left out. Worth noting that this fixture cannot carry it: sandboxes under `app/src/sandbox/**` do not style `data-focus-visible`, so a before and after screenshot of `focusable-click-focus` is identical, and the only three app surfaces that do style it, `app/src/components/code-block.react.tsx`, `app/src/examples/_lib/ariakit/select.react.tsx`, and `app/src/examples/combobox-group/combobox.react.tsx`, are tab, select, and combobox surfaces rather than a `Focusable` rendered as a form. Naming that explicitly is probably the honest fix here.

- Severity: Low (estimate, 85% confidence), because the automated coverage is strong and the missing artifact is supplementary.
- Priority: P2 (estimate, 75% confidence), because it is a description policy gap, not a behavior gap.
- Breakage likelihood: Likely (verified, 90% confidence), because Preview has no artifact today.
- Fix effort: Easy (estimate, 85% confidence) if the gap is stated, Moderate if a surface that styles the attribute has to be built.

**Declined non-goal: `getDocument` no longer validates the window branch.** The refactor turned `isWindow(node) ? node.document : ...` followed by an `isDocument` check into an unconditional `return node.document`, so a value that satisfies `isWindow` but whose `document` is not a document would now be returned from a function typed to return `Document`. I could not construct one. `isWindow` tests `value.window === value`, and a form's named controls cannot produce that self reference, so the only values that reach the branch are real windows. Not worth a guard.

**Inspected and sound: the `getWindow` fallback skips `ownsDocument`.** The direct branch needs that ownership test because the named getter can answer with a real window from another realm, namely an `<iframe name="defaultView">`. The prototype read cannot be answered by a named element at all, so there is nothing left to disambiguate. Confirmed on all three engines for a frame document that carries `<form name="defaultView">`.

**Not fully validated: the Dialog reachability that [#7215](https://github.com/ariakit/ariakit/issues/7215) describes.** The issue says pressing such a form outside an open [`Dialog`](https://ariakit.com/reference/dialog), [`Popover`](https://ariakit.com/reference/popover), or [`Menu`](https://ariakit.com/reference/menu) would leave it open. Reading `getEventTargets` in `use-previous-mouse-down-ref.ts`, `event.composedPath().filter(isElement)` drops the form but keeps its ancestors, and `rootTarget` then resolves to `<body>`, which is outside the dialog, so the dialog appears to hide anyway. I did not build a fixture to confirm that, so treat it as unconfirmed rather than as a correction. It matters only because it supports the narrower scope both changesets already take: they claim the [`Focusable`](https://ariakit.com/reference/focusable) styling effect, which I did reproduce, and do not claim a dialog effect.

### Not covered

The browser red check ran on Chrome only; Firefox and WebKit were verified green at head through the PR's own CI rather than red locally. I did not re-derive the happy-dom `defaultView` divergence from the earlier resolved thread. The removed ambient early return is no longer load bearing, because the direct read plus `ownsDocument` already pins the global window there, and the fallback is reachable only by shadowing `document.defaultView`, which breaks `dom-accessibility-api` before any Ariakit code runs. The changesets, fixture lifecycle, filename to project mapping, and the browser-only justification for skipping a happy-dom duplicate all check out.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in [`3dc9dfa54`](https://github.com/ariakit/ariakit/commit/3dc9dfa546c37c26bdae9af64cbb476d0ba9c293): documented why the prototype lookup retains `value` as the getter receiver across realms and why only an object-valued `nodeType` triggers the named-property fallback.

diegohaz:
**Fixed**

- [`3dc9dfa54`](https://github.com/ariakit/ariakit/commit/3dc9dfa546c37c26bdae9af64cbb476d0ba9c293) ([thread reply](https://github.com/ariakit/ariakit/pull/7217#discussion_r3821822470)): Documented the prototype receiver and object-valued collision invariants without changing behavior.

PR #7226, opened 2026-08-20
Resolve remaining named DOM member collisions

### LINKED ISSUE #7225: getActiveElement, isButton, and isVisible still read members a named element can answer
## Context

Found while reviewing [#7217](https://github.com/ariakit/ariakit/pull/7217), which reads `nodeType`, `ownerDocument`, and `defaultView` through the object's prototype so a named element cannot answer those lookups. That pull request covers the four reads named in [#7215](https://github.com/ariakit/ariakit/issues/7215): `isElement`, `isNode`, `getDocument`, and `getWindow`.

Three other reads in the same module still go through the named-property lookup. [#7218](https://github.com/ariakit/ariakit/issues/7218) covers the call sites outside `@ariakit/utils` that read `defaultView` directly; this issue covers the remaining reads inside `packages/ariakit-utils/src/dom.ts` itself. Two of them throw instead of degrading to a wrong answer.

## Problem

A document exposes `embed`, `form`, `iframe`, `img`, and `object` elements that carry a `name` attribute as own properties that override built-ins, and a form does the same for its listed controls. Everything below was measured on Chromium 151.0.7922.34, Firefox 153.0, and WebKit 26.5, driven through Playwright 1.62.1. All three engines agree on every result.

### `getActiveElement` reports a named form as the focused element

[`packages/ariakit-utils/src/dom.ts#L117`](https://github.com/ariakit/ariakit/blob/35439ddfe/packages/ariakit-utils/src/dom.ts#L117):

```ts
const { activeElement } = getDocument(node);
```

With `<form name="activeElement">` in the document and a `<button>` focused, `document.activeElement` answers with the form. The prototype read answers with the button:

| Read | Result |
| --- | --- |
| `document.activeElement` | `<form name="activeElement">` |
| `Reflect.get(Object.getPrototypeOf(document), "activeElement", document)` | `<button>` |

The existing IE11 guard does not catch it, because the form has a `nodeName`:

```ts
if (!activeElement?.nodeName) return null;
```

`getActiveElement` then reports the form as focused, so every caller that compares the active element against a component element gets a false answer. This read backs focus tracking across [`Composite`](https://ariakit.com/reference/composite), [`Dialog`](https://ariakit.com/reference/dialog), [`Menu`](https://ariakit.com/reference/menu), and [`Tab`](https://ariakit.com/reference/tab). The same document also answers `document.body` with a `<form name="body">`.

### `isButton` throws a `TypeError`

[`packages/ariakit-utils/src/dom.ts#L219`](https://github.com/ariakit/ariakit/blob/35439ddfe/packages/ariakit-utils/src/dom.ts#L219):

```ts
const tagName = element.tagName.toLowerCase();
```

A `<form>` containing `<input name="tagName">` answers `form.tagName` with the control, so `.toLowerCase()` is not a function. Measured message on Chromium and Firefox: `form.tagName.toLowerCase is not a function`.

`isButton` receives the rendered element in `command.tsx`, `focusable.tsx`, `button.tsx`, `composite-item.tsx`, and `composite-container.tsx`, so a [`Command`](https://ariakit.com/reference/command) derived component rendered as a `<form>` through [composition](https://ariakit.com/guide/composition) throws rather than degrading. The sandbox that [#7217](https://github.com/ariakit/ariakit/pull/7217) adds renders exactly that shape, `<Ariakit.Focusable render={<form />}>` wrapping a named control, which is the markup this read would meet.

### `isVisible` throws a `TypeError`

[`packages/ariakit-utils/src/dom.ts#L240-L247`](https://github.com/ariakit/ariakit/blob/35439ddfe/packages/ariakit-utils/src/dom.ts#L240-L247):

```ts
if (typeof element.checkVisibility === "function") {
  return element.checkVisibility();
}
const htmlElement = element as HTMLElement;
return (
  htmlElement.offsetWidth > 0 ||
  htmlElement.offsetHeight > 0 ||
  element.getClientRects().length > 0
);
```

A `<form>` with controls named `checkVisibility`, `offsetWidth`, `offsetHeight`, and `getClientRects` fails the capability test, compares two elements against `0`, and then throws on `element.getClientRects()`. Measured message: `form.getClientRects is not a function`.

Shadowing only `checkVisibility` and `getClientRects` is not enough when the form is laid out, because `offsetWidth > 0` short circuits first. `isVisible` backs `isFocusable` in `packages/ariakit-utils/src/focus.ts` and the composite item filters in `packages/ariakit-react-components/src/composite/utils.ts`.

## What was verified

All three reads are present unchanged on `main` at [`35439dd`](https://github.com/ariakit/ariakit/commit/35439ddfe), so none of this is introduced by [#7217](https://github.com/ariakit/ariakit/pull/7217). The diff of that pull request touches `getDocument`, `getWindow`, `isElement`, and `isNode` only.

The named-property behavior above was measured directly in the three engines on a plain HTML page, without Ariakit code in the picture. The end user symptom was not reproduced through a component fixture. happy-dom implements no document named getter and defines `nodeType` on the prototype chain, so a reproduction needs a real browser fixture, the way [#7217](https://github.com/ariakit/ariakit/pull/7217) added one under `app/src/sandbox/focusable-click-focus/`.

## Why 7217 can finish without this

[#7217](https://github.com/ariakit/ariakit/pull/7217) is scoped to the four reads that [#7215](https://github.com/ariakit/ariakit/issues/7215) named, and it proves each one with a red check. These three reads exist unchanged on `main`, they are reached through different helpers, and fixing them does not change what the hardened helpers guarantee.

## Expected outcome

`getActiveElement` resolves `activeElement` through a read that a named element cannot answer, in the same way [#7217](https://github.com/ariakit/ariakit/pull/7217) resolves `ownerDocument` and `defaultView`. `isButton` and `isVisible` either resolve their reads the same way or stop throwing, so a named control degrades to a wrong-but-safe answer rather than to an exception.

Add real browser coverage for each one, placing the colliding markup in the preview document. Weigh the cost on `isVisible`, which composite item filters call for every item, against the same measurement standard [#7217](https://github.com/ariakit/ariakit/pull/7217) applied to `isElement`.

### Workaround

Do not give a form control or a named `embed`, `form`, `iframe`, `img`, or `object` element a `name` that matches a DOM member Ariakit reads, such as `activeElement`, `body`, `tagName`, `checkVisibility`, `offsetWidth`, `offsetHeight`, or `getClientRects`. Use a collision free DOM field name and remap it where the form data is prepared:

```html
<input name="tagNameField" />
```

```ts
const data = new FormData(form);
data.set("tagName", data.get("tagNameField") ?? "");
data.delete("tagNameField");
```

diegohaz:
## Motivation

HTML documents and forms expose named elements as properties. Those properties can replace the DOM members that Ariakit reads.

A form named `activeElement` makes `getActiveElement` report the form instead of the focused control. Form controls named `tagName`, `checkVisibility`, `offsetWidth`, `offsetHeight`, and `getClientRects` make `isButton` and `isVisible` throw instead of returning a result.

## Solution

Read a native DOM member from its prototype when a named property shadows the normal value. `Reflect.get` keeps the original document or element as the getter receiver, including for values from another realm.

Use the existing ES2018-safe ownership helper for `activeElement`. Centralize the other direct and prototype reads in a typed `getDOMMember` helper, and keep `.call(element)` at method call sites so the hot path does not allocate bound functions.

Add utility regression tests and a real-browser sandbox with all six colliding names. A focused unit test removes `Element.prototype.checkVisibility` so the legacy geometry fallbacks also run. The sandbox renders an Ariakit `Dialog` and verifies that focus returns to its opener without reading the colliding `document.activeElement` value.

The `tagName` fix in this PR is limited to `@ariakit/utils`. Other direct DOM member reads in the React and event packages can still shadow before `isButton` runs; [#7228](https://github.com/ariakit/ariakit/issues/7228) tracks that broader audit and fix.

## Workaround

Until a release with this fix is available, use collision-free DOM names and remap them when preparing the submitted form data:

```tsx
{/* TODO: Remove this remapping after https://github.com/ariakit/ariakit/issues/7225 is released. */}
<input name="tagNameField" />
```

```ts
const data = new FormData(form);
data.set("tagName", data.get("tagNameField") ?? "");
data.delete("tagNameField");
```

Apply the same suffix-and-remap pattern to `activeElement`, `checkVisibility`, `offsetWidth`, `offsetHeight`, and `getClientRects`. This workaround requires control over both the DOM field names and the point where `FormData` is prepared.

## Preview

The final sandbox reports `Focused button`, `Not a button`, and `Visible` with the collision-prone names present. Its Ariakit `Dialog` also restores focus from `Close dialog` to `Open dialog`. The focused browser suite passes all 12 cases across Chrome, Firefox, and Safari. The utility suite passes all 82 tests, and the production-style app build includes `/react/previews/dom-7225`.

[Dialog focus-restoration recording](https://github.com/user-attachments/assets/8e32ce1c-6b02-401e-befc-c696bbc0917f)

Fixes #7225

### DIFF
--- a/app/src/sandbox/dom-7225/index.react.tsx
+++ b/app/src/sandbox/dom-7225/index.react.tsx
@@ -28,26 +28,26 @@ export default function Example() {
 
   return (
     <main>
-      <form ref={formRef} name="profileForm" aria-label="Profile">
+      <form ref={formRef} name="activeElement" aria-label="Profile">
         <label>
           Tag name
-          <input name="tagNameField" />
+          <input name="tagName" />
         </label>
         <label>
           Visibility check
-          <input name="checkVisibilityField" />
+          <input name="checkVisibility" />
         </label>
         <label>
           Width
-          <input name="offsetWidthField" />
+          <input name="offsetWidth" />
         </label>
         <label>
           Height
-          <input name="offsetHeightField" />
+          <input name="offsetHeight" />
         </label>
         <label>
           Client rectangles
-          <input name="getClientRectsField" />
+          <input name="getClientRects" />
         </label>
       </form>
 
--- a/app/src/sandbox/dom-7225/index.react.tsx
+++ b/app/src/sandbox/dom-7225/index.react.tsx
@@ -1,0 +1,73 @@
+import { getActiveElement, isButton, isVisible } from "@ariakit/utils";
+import { useRef, useState } from "react";
+
+export default function Example() {
+  const formRef = useRef<HTMLFormElement>(null);
+  const [activeElementResult, setActiveElementResult] = useState("Not checked");
+  const [buttonResult, setButtonResult] = useState("Not checked");
+  const [visibilityResult, setVisibilityResult] = useState("Not checked");
+
+  const checkActiveElement = (button: HTMLButtonElement) => {
+    const activeElement = getActiveElement(button);
+    setActiveElementResult(
+      activeElement === button ? "Focused button" : "Wrong element",
+    );
+  };
+
+  const checkButton = () => {
+    const form = formRef.current;
+    if (!form) return;
+    setButtonResult(isButton(form) ? "Button" : "Not a button");
+  };
+
+  const checkVisibility = () => {
+    const form = formRef.current;
+    if (!form) return;
+    setVisibilityResult(isVisible(form) ? "Visible" : "Hidden");
+  };
+
+  return (
+    <main>
+      <form ref={formRef} name="activeElement" aria-label="Profile">
+        <label>
+          Tag name
+          <input name="tagName" />
+        </label>
+        <label>
+          Visibility check
+          <input name="checkVisibility" />
+        </label>
+        <label>
+          Width
+          <input name="offsetWidth" />
+        </label>
+        <label>
+          Height
+          <input name="offsetHeight" />
+        </label>
+        <label>
+          Client rectangles
+          <input name="getClientRects" />
+        </label>
+      </form>
+
+      <button
+        type="button"
+        onClick={(event) => checkActiveElement(event.currentTarget)}
+      >
+        Check active element
+      </button>
+      <output aria-label="Active element result">{activeElementResult}</output>
+
+      <button type="button" onClick={checkButton}>
+        Check button type
+      </button>
+      <output aria-label="Button result">{buttonResult}</output>
+
+      <button type="button" onClick={checkVisibility}>
+        Check visibility
+      </button>
+      <output aria-label="Visibility result">{visibilityResult}</output>
+    </main>
+  );
+}
--- a/app/src/sandbox/dom-7225/index.react.tsx
+++ b/app/src/sandbox/dom-7225/index.react.tsx
@@ -1,11 +1,14 @@
+import * as Ariakit from "@ariakit/react";
 import { getActiveElement, isButton, isVisible } from "@ariakit/utils";
 import { useRef, useState } from "react";
 
 export default function Example() {
+  const dialog = Ariakit.useDialogStore();
   const formRef = useRef<HTMLFormElement>(null);
   const [activeElementResult, setActiveElementResult] = useState("Not checked");
   const [buttonResult, setButtonResult] = useState("Not checked");
   const [visibilityResult, setVisibilityResult] = useState("Not checked");
+  const [focusedElement, setFocusedElement] = useState("None");
 
   const checkActiveElement = (button: HTMLButtonElement) => {
     const activeElement = getActiveElement(button);
@@ -50,6 +53,24 @@ export default function Example() {
           <input name="getClientRects" />
         </label>
       </form>
+
+      <button
+        type="button"
+        onClick={dialog.show}
+        onFocus={() => setFocusedElement("Open dialog")}
+      >
+        Open dialog
+      </button>
+      <Ariakit.Dialog store={dialog} modal={false} aria-label="Example dialog">
+        <button
+          type="button"
+          onClick={dialog.hide}
+          onFocus={() => setFocusedElement("Close dialog")}
+        >
+          Close dialog
+        </button>
+      </Ariakit.Dialog>
+      <output aria-label="Focused element">{focusedElement}</output>
 
       <button
         type="button"
--- a/app/src/sandbox/dom-7225/index.react.tsx
+++ b/app/src/sandbox/dom-7225/index.react.tsx
@@ -28,26 +28,26 @@ export default function Example() {
 
   return (
     <main>
-      <form ref={formRef} name="activeElement" aria-label="Profile">
+      <form ref={formRef} name="profileForm" aria-label="Profile">
         <label>
           Tag name
-          <input name="tagName" />
+          <input name="tagNameField" />
         </label>
         <label>
           Visibility check
-          <input name="checkVisibility" />
+          <input name="checkVisibilityField" />
         </label>
         <label>
           Width
-          <input name="offsetWidth" />
+          <input name="offsetWidthField" />
         </label>
         <label>
           Height
-          <input name="offsetHeight" />
+          <input name="offsetHeightField" />
         </label>
         <label>
           Client rectangles
-          <input name="getClientRects" />
+          <input name="getClientRectsField" />
         </label>
       </form>
 
--- a/packages/ariakit-utils/src/dom.ts
+++ b/packages/ariakit-utils/src/dom.ts
@@ -122,7 +122,11 @@ export function getActiveElement(
   node?: Node | null,
   activeDescendant = false,
 ): HTMLElement | null {
-  const { activeElement } = getDocument(node);
+  const ownerDocument = getDocument(node);
+  // The native accessor is inherited, so an own answer may be a named element.
+  const activeElement = Object.hasOwn(ownerDocument, "activeElement")
+    ? getDOMProperty(ownerDocument, "activeElement")
+    : ownerDocument.activeElement;
   if (!activeElement?.nodeName) {
     // In IE11, activeElement might be an empty object if we're interacting
     // with elements inside of an iframe.
@@ -229,9 +233,15 @@ export function isFrame(element: Element): element is HTMLIFrameElement {
  * isButton(document.querySelector("div[role='button']")); // false
  */
 export function isButton(element: { tagName: string; type?: string }) {
-  const tagName = element.tagName.toLowerCase();
-  if (tagName === "button") return true;
-  if (tagName === "input" && element.type) {
+  const tagNameProperty = element.tagName;
+  const tagName =
+    typeof tagNameProperty === "string"
+      ? tagNameProperty
+      : getDOMProperty(element, "tagName");
+  if (typeof tagName !== "string") return false;
+  const normalizedTagName = tagName.toLowerCase();
+  if (normalizedTagName === "button") return true;
+  if (normalizedTagName === "input" && element.type) {
     return buttonInputTypes.indexOf(element.type) !== -1;
   }
   return false;
@@ -253,12 +263,29 @@ export function isVisible(element: Element) {
   if (typeof element.checkVisibility === "function") {
     return element.checkVisibility();
   }
+  const unshadowedCheckVisibility = getDOMProperty(element, "checkVisibility");
+  if (typeof unshadowedCheckVisibility === "function") {
+    return unshadowedCheckVisibility.call(element);
+  }
   const htmlElement = element as HTMLElement;
-  return (
-    htmlElement.offsetWidth > 0 ||
-    htmlElement.offsetHeight > 0 ||
-    element.getClientRects().length > 0
-  );
+  const offsetWidthProperty = htmlElement.offsetWidth;
+  const offsetWidth =
+    typeof offsetWidthProperty === "number"
+      ? offsetWidthProperty
+      : getDOMProperty(element, "offsetWidth");
+  if (typeof offsetWidth === "number" && offsetWidth > 0) return true;
+  const offsetHeightProperty = htmlElement.offsetHeight;
+  const offsetHeight =
+    typeof offsetHeightProperty === "number"
+      ? offsetHeightProperty
+      : getDOMProperty(element, "offsetHeight");
+  if (typeof offsetHeight === "number" && offsetHeight > 0) return true;
+  if (typeof element.getClientRects === "function") {
+    return element.getClientRects().length > 0;
+  }
+  const unshadowedGetClientRects = getDOMProperty(element, "getClientRects");
+  if (typeof unshadowedGetClientRects !== "function") return false;
+  return unshadowedGetClientRects.call(element).length > 0;
 }
 
 /**
--- a/packages/ariakit-utils/src/dom.ts
+++ b/packages/ariakit-utils/src/dom.ts
@@ -259,7 +259,7 @@ const buttonInputTypes = [
 /**
  * Checks if the element is visible or not.
  */
-export function isVisible(element: Element) {
+export function isVisible(element: Element): boolean {
   if (typeof element.checkVisibility === "function") {
     return element.checkVisibility();
   }
--- a/packages/ariakit-utils/src/dom.ts
+++ b/packages/ariakit-utils/src/dom.ts
@@ -74,12 +74,12 @@ interface DOMMemberTypes {
   string: string;
 }
 
-function getDOMMember<Type extends keyof DOMMemberTypes>(
-  value: object,
-  name: string,
+function getDOMMember<Value extends object, Type extends keyof DOMMemberTypes>(
+  value: Value,
+  name: Extract<keyof Value, string>,
   type: Type,
 ): DOMMemberTypes[Type] | undefined {
-  const member: unknown = Reflect.get(value, name);
+  const member: unknown = value[name];
   if (typeof member === type) {
     return member as DOMMemberTypes[Type];
   }
@@ -282,9 +282,10 @@ const buttonInputTypes = [
 export function isVisible(element: Element): boolean {
   const checkVisibility = getDOMMember(element, "checkVisibility", "function");
   if (checkVisibility) return checkVisibility.call(element);
-  const offsetWidth = getDOMMember(element, "offsetWidth", "number");
+  const htmlElement = element as HTMLElement;
+  const offsetWidth = getDOMMember(htmlElement, "offsetWidth", "number");
   if (offsetWidth != null && offsetWidth > 0) return true;
-  const offsetHeight = getDOMMember(element, "offsetHeight", "number");
+  const offsetHeight = getDOMMember(htmlElement, "offsetHeight", "number");
   if (offsetHeight != null && offsetHeight > 0) return true;
   const getClientRects = getDOMMember(element, "getClientRects", "function");
   if (!getClientRects) return false;
--- a/packages/ariakit-utils/src/dom.ts
+++ b/packages/ariakit-utils/src/dom.ts
@@ -4,7 +4,7 @@
  */
 
 import { hasOwnProperty } from "./misc.ts";
-import type { AriaHasPopup, AriaRole } from "./types.ts";
+import type { AnyFunction, AriaHasPopup, AriaRole } from "./types.ts";
 
 /**
  * It's `true` if it is running in a browser environment or `false` if it is not
@@ -62,8 +62,32 @@ function isDocument(value: unknown): value is Document {
 
 // Start at the prototype to bypass named properties while keeping `value` as
 // the getter receiver, including for values from another realm.
-function getDOMProperty(value: object, name: string) {
+function getDOMProperty(value: Document, name: "activeElement"): Element | null;
+function getDOMProperty(value: object, name: string): unknown;
+function getDOMProperty(value: object, name: string): unknown {
   return Reflect.get(Object.getPrototypeOf(value), name, value);
+}
+
+interface DOMMemberTypes {
+  function: AnyFunction;
+  number: number;
+  string: string;
+}
+
+function getDOMMember<Type extends keyof DOMMemberTypes>(
+  value: object,
+  name: string,
+  type: Type,
+): DOMMemberTypes[Type] | undefined {
+  const member: unknown = Reflect.get(value, name);
+  if (typeof member === type) {
+    return member as DOMMemberTypes[Type];
+  }
+  const unshadowedMember = getDOMProperty(value, name);
+  if (typeof unshadowedMember === type) {
+    return unshadowedMember as DOMMemberTypes[Type];
+  }
+  return undefined;
 }
 
 function ownsDocument(
@@ -124,7 +148,7 @@ export function getActiveElement(
 ): HTMLElement | null {
   const ownerDocument = getDocument(node);
   // The native accessor is inherited, so an own answer may be a named element.
-  const activeElement = Object.hasOwn(ownerDocument, "activeElement")
+  const activeElement = hasOwnProperty(ownerDocument, "activeElement")
     ? getDOMProperty(ownerDocument, "activeElement")
     : ownerDocument.activeElement;
   if (!activeElement?.nodeName) {
@@ -233,12 +257,8 @@ export function isFrame(element: Element): element is HTMLIFrameElement {
  * isButton(document.querySelector("div[role='button']")); // false
  */
 export function isButton(element: { tagName: string; type?: string }) {
-  const tagNameProperty = element.tagName;
-  const tagName =
-    typeof tagNameProperty === "string"
-      ? tagNameProperty
-      : getDOMProperty(element, "tagName");
-  if (typeof tagName !== "string") return false;
+  const tagName = getDOMMember(element, "tagName", "string");
+  if (!tagName) return false;
   const normalizedTagName = tagName.toLowerCase();
   if (normalizedTagName === "button") return true;
   if (normalizedTagName === "input" && element.type) {
@@ -260,32 +280,15 @@ const buttonInputTypes = [
  * Checks if the element is visible or not.
  */
 export function isVisible(element: Element): boolean {
-  if (typeof element.checkVisibility === "function") {
-    return element.checkVisibility();
-  }
-  const unshadowedCheckVisibility = getDOMProperty(element, "checkVisibility");
-  if (typeof unshadowedCheckVisibility === "function") {
-    return unshadowedCheckVisibility.call(element);
-  }
-  const htmlElement = element as HTMLElement;
-  const offsetWidthProperty = htmlElement.offsetWidth;
-  const offsetWidth =
-    typeof offsetWidthProperty === "number"
-      ? offsetWidthProperty
-      : getDOMProperty(element, "offsetWidth");
-  if (typeof offsetWidth === "number" && offsetWidth > 0) return true;
-  const offsetHeightProperty = htmlElement.offsetHeight;
-  const offsetHeight =
-    typeof offsetHeightProperty === "number"
-      ? offsetHeightProperty
-      : getDOMProperty(element, "offsetHeight");
-  if (typeof offsetHeight === "number" && offsetHeight > 0) return true;
-  if (typeof element.getClientRects === "function") {
-    return element.getClientRects().length > 0;
-  }
-  const unshadowedGetClientRects = getDOMProperty(element, "getClientRects");
-  if (typeof unshadowedGetClientRects !== "function") return false;
-  return unshadowedGetClientRects.call(element).length > 0;
+  const checkVisibility = getDOMMember(element, "checkVisibility", "function");
+  if (checkVisibility) return checkVisibility.call(element);
+  const offsetWidth = getDOMMember(element, "offsetWidth", "number");
+  if (offsetWidth != null && offsetWidth > 0) return true;
+  const offsetHeight = getDOMMember(element, "offsetHeight", "number");
+  if (offsetHeight != null && offsetHeight > 0) return true;
+  const getClientRects = getDOMMember(element, "getClientRects", "function");
+  if (!getClientRects) return false;
+  return getClientRects.call(element).length > 0;
 }
 
 /**
--- a/.changeset/300-react-named-active-element.md
+++ b/.changeset/300-react-named-active-element.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Composite`](https://ariakit.com/reference/composite) and [`Dialog`](https://ariakit.com/reference/dialog) focus tracking when a document exposes a form named `activeElement`.
--- a/.changeset/300-utils-named-active-button-visibility.md
+++ b/.changeset/300-utils-named-active-button-visibility.md
@@ -1,0 +1,5 @@
+---
+"@ariakit/utils": patch
+---
+
+Fixed `getActiveElement` reading a named form instead of the focused element, and fixed `isButton` and `isVisible` throwing when form controls shadow the DOM members they read.
--- a/app/src/sandbox/dom-7225/test-browser.ts
+++ b/app/src/sandbox/dom-7225/test-browser.ts
@@ -1,0 +1,30 @@
+import { withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/7225
+  test("resolves the focused element past a named form", async ({ q }) => {
+    const button = q.button("Check active element");
+    await button.focus();
+    await button.press("Enter");
+
+    await test
+      .expect(q.status("Active element result"))
+      .toHaveText("Focused button");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7225
+  test("checks a form with a control named tagName", async ({ q }) => {
+    await q.button("Check button type").click();
+
+    await test.expect(q.status("Button result")).toHaveText("Not a button");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7225
+  test("checks visibility with controls named after DOM members", async ({
+    q,
+  }) => {
+    await q.button("Check visibility").click();
+
+    await test.expect(q.status("Visibility result")).toHaveText("Visible");
+  });
+});
--- a/app/src/sandbox/dom-7225/test-browser.ts
+++ b/app/src/sandbox/dom-7225/test-browser.ts
@@ -13,6 +13,20 @@ withFramework(import.meta.dirname, async ({ test }) => {
   });
 
   // https://github.com/ariakit/ariakit/issues/7225
+  // happy-dom has no document named getter, so this integration stays in the
+  // browser suite that reproduces the platform behavior.
+  test("restores dialog focus past a named activeElement form", async ({
+    q,
+  }) => {
+    await q.button("Open dialog").click();
+    await test.expect(q.status("Focused element")).toHaveText("Close dialog");
+
+    await q.button("Close dialog").click();
+
+    await test.expect(q.status("Focused element")).toHaveText("Open dialog");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7225
   test("checks a form with a control named tagName", async ({ q }) => {
     await q.button("Check button type").click();
 
--- a/packages/ariakit-utils/readme.md
+++ b/packages/ariakit-utils/readme.md
@@ -397,7 +397,7 @@ isButton(document.querySelector("div[role='button']")); // false
 #### `isVisible`
 
 ```ts
-function isVisible(element: Element): any;
+function isVisible(element: Element): boolean;
 ```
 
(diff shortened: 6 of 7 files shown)


packages/ariakit-utils/src/dom.ts
```diff
@@ -249,16 +259,33 @@ const buttonInputTypes = [
...
  */
-export function isVisible(element: Element) {
+export function isVisible(element: Element): boolean {
   if (typeof element.checkVisibility === "function") {
     return element.checkVisibility();
   }
+  const unshadowedCheckVisibility = getDOMProperty(element, "checkVisibility");
+  if (typeof unshadowedCheckVisibility === "function") {
+    return unshadowedCheckVisibility.call(element);
+  }
```
diegohaz: I wonder if this can all be done inside `getDOMProperty` so we can just do this?

```js
const checkVisibility = getDOMProperty(element, "checkVisibility", "function");
const offsetWidth = getDOMProperty(element, "offsetWidth", "number");
```

Where we would automatically call `.bind(element)` inside `getDOMProperty` if the last parameter is `"function"`?

I'm not sure if this would impact performance, though. There's also TypeScript, so I believe `getDOMProperty` would have to use generics.

packages/ariakit-utils/src/dom.ts
```diff
@@ -229,9 +233,15 @@ export function isFrame(element: Element): element is HTMLIFrameElement {
  * isButton(document.querySelector("div[role='button']")); // false
  */
 export function isButton(element: { tagName: string; type?: string }) {
```
diegohaz, line 259: This read is fixed, but the React crash it is meant to remove is still there, because two other reads of `tagName` run first.

I served the app preview from this branch and mounted the shape the issue names, a `Command` derived component rendered as a form:

```tsx
<Ariakit.Button render={<form aria-label="Tag editor" />}>
  <input aria-label="Tag name" name="tagName" />
</Ariakit.Button>
```

On this head it still fails to mount in Chromium 151, Firefox 153, and WebKit 26.5:

```
[pageerror] element?.tagName.toLowerCase is not a function
An error occurred in the <Button> component.
```

That message is `useTagName` at `packages/ariakit-react-utils/src/hooks.ts:226`, which runs in a layout effect for every `Focusable`, `Command`, and `Button`. `packages/ariakit-react-components/src/focusable/focusable.tsx:443` reads the same member with no browser gate, and `packages/ariakit-utils/src/events.ts:39` reads it inside this very package for `isOpeningInNewTab` and `isDownloading`.

Hardening only those first two locally, on top of this head, makes the same sandbox mount and Ariakit apply the degraded answer this PR enables:

```html
<form aria-label="Tag editor" role="button" tabindex="0">
```

Reverting `isButton` to its `main` implementation while keeping those two local patches throws again with `element.tagName.toLowerCase is not a function`, so this change is necessary as well as insufficient.

I registered the remaining reads as https://github.com/ariakit/ariakit/issues/7228 rather than asking for them here, because they live in three other packages and `@ariakit/utils` exports no hardened tag name helper today, so resolving them is a public API decision.

What I would change in this PR is the framing. The description opens with "Form controls named `tagName` [...] make `isButton` and `isVisible` throw instead of returning a result", and the PR title says "remaining", but a React consumer who hits that crash still crashes after this ships. A sentence in the description scoping the fix to `@ariakit/utils` and pointing at #7228 would keep the changelog and the PR honest.

**Assessment**

- Severity: Medium (verified, 95% confidence), because the library level fix is correct and provable, and only the PR's stated user facing outcome is overstated. The crash itself is tracked in #7228.
- Priority: P1 (estimate, 80% confidence), because the scope note belongs with the description and changesets this PR ships, not in a follow up.
- Breakage likelihood: Likely (verified, 95% confidence) for the framing to mislead, because the issue this PR closes uses the React crash as its justification and the React path measurably still throws in all three engines.
- Fix effort: Easy (estimate, 90% confidence), because it is a description edit plus the existing #7228 link.

packages/ariakit-utils/src/dom.ts
```diff
@@ -249,16 +259,33 @@ const buttonInputTypes = [
...
+    typeof offsetHeightProperty === "number"
+      ? offsetHeightProperty
+      : getDOMProperty(element, "offsetHeight");
+  if (typeof offsetHeight === "number" && offsetHeight > 0) return true;
+  if (typeof element.getClientRects === "function") {
+    return element.getClientRects().length > 0;
+  }
+  const unshadowedGetClientRects = getDOMProperty(element, "getClientRects");
+  if (typeof unshadowedGetClientRects !== "function") return false;
+  return unshadowedGetClientRects.call(element).length > 0;
```
diegohaz: These three fallbacks never run in any environment this repository tests, so nothing here is covered.

The `checkVisibility` fallback two lines above always answers first. I instrumented `isVisible` with `process.stderr.write` and ran `pnpm test packages/ariakit-utils/src/dom.test.ts`. The only branch marker that ever printed was the `checkVisibility` one, across all 28 tests including the new `isVisible ignores controls named after the members it reads` case. happy-dom defines `checkVisibility` on the prototype, so the prototype read finds it and returns before the geometry reads.

Real browsers behave the same way. Driving a page with `<form>` and controls named after all five members through Playwright 1.62.1:

| | Chromium 151 | Firefox 153 | WebKit 26.5 |
| --- | --- | --- | --- |
| `typeof form.checkVisibility` | `object` | `object` | `object` |
| `typeof proto(form, "checkVisibility")` | `function` | `function` | `function` |
| `proto(form, "checkVisibility").call(form)` | `true` | `true` | `true` |

So the new browser test `checks visibility with controls named after DOM members` also stops at the `checkVisibility` fallback and never reaches these lines either. That leaves 18 of the 20 new lines in `isVisible` executed by nothing, in a helper the composite item filters call for every item.

They are only reachable below Chrome 105, Firefox 125, and Safari 17.4, where `checkVisibility` does not exist. Two ways out, either is fine:

Cover them. The cheapest honest way is to make the prototype read come back empty for one test, so the geometry path is what the assertion exercises:

```ts
// https://github.com/ariakit/ariakit/issues/7225
// Emulate an engine without `checkVisibility`, so the geometry reads below are
// what the shadowed controls have to survive.
test("isVisible reads geometry past named controls without checkVisibility", () => {
  const descriptor = Object.getOwnPropertyDescriptor(
    Element.prototype,
    "checkVisibility",
  );
  // ...delete, assert, restore
});
```

Or drop them. If the project is willing to say that a browser without `checkVisibility` may still throw on this shape, keep the legacy branch exactly as it is on `main` and let the hardening stop at `checkVisibility`. That is a smaller diff than untested code, and it is honest about what is proven.

**Assessment**

- Severity: Low (verified, 90% confidence), because the branches read correctly to me, so the exposure is unproven code rather than a known defect.
- Priority: P2 (estimate, 75% confidence), because it does not block the fix that this PR does prove.
- Breakage likelihood: Rare (estimate, 85% confidence), because reaching the code needs a browser older than Chrome 105, Firefox 125, or Safari 17.4 together with a form whose controls are named after these members.
- Fix effort: Easy (estimate, 85% confidence), because it is one focused test with a restored descriptor, or a deletion.

app/src/sandbox/dom-7225/index.react.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const checkVisibility = () => {
+    const form = formRef.current;
+    if (!form) return;
+    setVisibilityResult(isVisible(form) ? "Visible" : "Hidden");
+  };
+
+  return (
+    <main>
+      <form ref={formRef} name="activeElement" aria-label="Profile">
```
diegohaz, line 34: This sandbox calls the three utilities directly and renders no Ariakit component, so nothing here covers the `@ariakit/react-components` changeset this PR ships. That gap is what let the crash in https://github.com/ariakit/ariakit/pull/7226#discussion_r3822900104 through.

`ariakit-ariakit-workflow` asks sandboxes to show how a user meets the bug through ordinary Ariakit components and to avoid a test-only harness. The three buttons and `<output>` elements here exist only so a test can read a result. Calling `@ariakit/utils` directly is a real consumer surface, so this is defensible as utility coverage, but it is not the coverage the react changeset claims.

The issue already points at the fixture that would carry it. `app/src/sandbox/focusable-click-focus/index.react.tsx:38` renders exactly the shape #7225 describes:

```tsx
{/* https://github.com/ariakit/ariakit/issues/7215 */}
<Ariakit.Focusable
  render={<form aria-label="Node editor" style={{ padding: 16 }} />}
  tabIndex={0}
>
  <input aria-label="Node type" name="nodeType" />
</Ariakit.Focusable>
```

Adding one sibling control named `tagName` to that form, or a `Command` derived variant beside it, gives a component level red check. `packages/ariakit-react-components/src/command/command.tsx:74` calls `isButton(element)` in an unconditional mount effect, so the throw is not Safari only the way `focusable.tsx:467` is. I confirmed the red check by reverting `isButton` to its `main` implementation: the tree fails to mount in Chromium, Firefox, and WebKit, and the preview renders nothing.

The fixture lifecycle in `ariakit-ariakit-example-playwright-test` puts extension of an existing semantic fixture first, so I would keep this sandbox for the utility level cases and put the component case in `focusable-click-focus`.

The missing happy-dom `test.ts` is fine as it stands, since happy-dom implements no document named getter and defines these members on the prototype. `focusable-click-focus/test-browser.ts:33` records that reason in a comment next to the test that needs it; a similar comment here would save the next reader the check.

**Assessment**

- Severity: Medium (verified, 90% confidence), because a react changeset ships with no component level regression, and the missing coverage already hid a real crash.
- Priority: P1 (estimate, 80% confidence), because the changeset and the fixture belong to the same change.
- Breakage likelihood: Possible (estimate, 75% confidence), because a future refactor of the `isButton` call sites in `command.tsx` or `focusable.tsx` would not be caught by anything in the suite.
- Fix effort: Easy (estimate, 85% confidence), because it is one control and one test added to a fixture that already renders the shape.

packages/ariakit-utils/src/dom.ts
```diff
@@ -122,7 +122,11 @@ export function getActiveElement(
   node?: Node | null,
   activeDescendant = false,
 ): HTMLElement | null {
-  const { activeElement } = getDocument(node);
+  const ownerDocument = getDocument(node);
+  // The native accessor is inherited, so an own answer may be a named element.
+  const activeElement = Object.hasOwn(ownerDocument, "activeElement")
```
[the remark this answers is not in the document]
diegohaz: Confirmed, and the fix is a one word swap.

`tsconfig.base.json` does set `"target": "es2018"`, and `"lib": ["DOM", "ESNext"]` means TypeScript never flags `Object.hasOwn`, so this ships as written. `Object.hasOwn` is ES2022, available from Chrome 93, Firefox 92, and Safari 15.4. Below any of those, `Object.hasOwn` is `undefined` and the call throws a `TypeError` on every `getActiveElement`, whether or not any named collision exists. That is focus tracking for `Composite`, `Dialog`, `Menu`, and `Tab` failing on load.

The claim about the existing helper checks out too. `packages/ariakit-utils/src/misc.ts:126-130` is exactly this guard:

```ts
if (typeof Object.hasOwn === "function") {
  return Object.hasOwn(object, prop);
}
return Object.prototype.hasOwnProperty.call(object, prop);
```

and `dom.ts:6` already imports `hasOwnProperty` from it for `getItemRoleByPopupRole`, so no new import is needed.

I applied the swap locally and verified it:

```diff
-  const activeElement = Object.hasOwn(ownerDocument, "activeElement")
+  const activeElement = hasOwnProperty(ownerDocument, "activeElement")
```

`pnpm tsc` reports 0 errors, and `pnpm test packages/ariakit-utils/src/dom.test.ts` passes 28 of 28, including the new named `activeElement` case. The type predicate on `hasOwnProperty` narrows `"activeElement"` against `Document` without complaint.

One note on the surrounding baseline, so this is not read as stricter than it is: `.includes` appears in `packages/ariakit-react-components/src`, so the repository is not holding a pure ES2018 API line. What makes this one worth changing is narrower than the target setting. The package ships a guarded helper for this exact API, this file already imports it, and the unguarded call sits on a hot path that runs on every focus change.

The only unguarded `Object.hasOwn` in shipped source besides this one is `packages/ariakit-test/src/shims.ts:465`, and `@ariakit/test` runs in Node and happy-dom rather than in a consumer's browser, so it is not a precedent here.

**Assessment**

- Severity: Critical (verified, 90% confidence), because focus tracking fails for every affected consumer on load, not only for the named collision this PR targets.
- Priority: P1 (estimate, 85% confidence), because it is introduced by this PR and the fix is one identifier.
- Breakage likelihood: Rare (estimate, 70% confidence), because it needs a browser older than Chrome 93, Firefox 92, or Safari 15.4, and I could not find a published support floor for the repository to check that against.
- Fix effort: Easy (verified, 95% confidence), because the swap is validated above with `tsc` and the focused suite green.

packages/ariakit-utils/src/dom.ts
diegohaz: Yes, this works, and it is cheaper than what is on the branch. I prototyped it and measured the two things you were unsure about.

Generics are enough, with a lookup interface rather than a conditional type:

```ts
interface DOMMemberTypes {
  string: string;
  number: number;
  function: (...args: any[]) => any;
}

// A named property is always an element, so a member that already has its
// native type is genuine and the prototype read below is skipped.
function getDOMMember<T extends keyof DOMMemberTypes>(
  value: object,
  name: string,
  type: T,
): DOMMemberTypes[T] | undefined {
  const member = (value as Record<string, unknown>)[name];
  if (typeof member === type) return member as DOMMemberTypes[T];
  const unshadowed = getDOMProperty(value, name);
  if (typeof unshadowed === type) return unshadowed as DOMMemberTypes[T];
  return undefined;
}
```

`isVisible` then reads as the original did:

```ts
export function isVisible(element: Element): boolean {
  const checkVisibility = getDOMMember(element, "checkVisibility", "function");
  if (checkVisibility) return checkVisibility.call(element);
  const offsetWidth = getDOMMember(element, "offsetWidth", "number");
  if (offsetWidth) return true;
  const offsetHeight = getDOMMember(element, "offsetHeight", "number");
  if (offsetHeight) return true;
  const getClientRects = getDOMMember(element, "getClientRects", "function");
  if (!getClientRects) return false;
  return getClientRects.call(element).length > 0;
}
```

and `isButton` loses its ternary entirely:

```ts
const tagName = getDOMMember(element, "tagName", "string");
if (!tagName) return false;
const normalizedTagName = tagName.toLowerCase();
```

`pnpm tsc` reports 0 errors and `pnpm test packages/ariakit-utils/src/dom.test.ts packages/ariakit-utils/src/focus.test.ts` passes 39 of 39 against it.

On performance, it costs nothing. Chromium 151, 100k `isVisible` calls per sample, 500 passes over 200 laid out elements, median of 7 samples after warmup, three runs:

| Variant | Run 1 | Run 2 | Run 3 |
| --- | --- | --- | --- |
| This branch | 6.7 ms | 7.1 ms | 7.0 ms |
| `getDOMMember` with `.call(element)` | 6.0 ms | 6.5 ms | 6.6 ms |
| `getDOMMember` with `.bind(element)` | 6.9 ms | 6.8 ms | 6.8 ms |

So the consolidation is consistently the fastest of the three, and the extra call frame does not show up. The `.bind` part is the only piece that costs anything, around 5 to 10 ns per call from the closure it allocates, and `isVisible` is the helper the composite item filters run for every item. Since `.call(element)` is one word longer and allocates nothing, I would keep `.call` and skip the auto bind.

Two things the generic version also buys, beyond the line count:

It removes an `any` from the module. `getDOMProperty` returns `Reflect.get(Object.getPrototypeOf(value), name, value)`, and `Object.getPrototypeOf` returns `any`, so the whole result is `any` today. In `getActiveElement` that `any` now flows through `isFrame`, `getAttribute`, and out through `as HTMLElement | null` unchecked, and I suspect it is why `isVisible` needed the explicit `: boolean` in fb07b8b. `DOMMemberTypes[T] | undefined` restores the checking at every call site.

It also makes the untested branches visible. Flattening `isVisible` to four `getDOMMember` calls makes it obvious at a glance that everything after the first one is unreachable wherever `checkVisibility` exists, which is every engine and environment this repository tests. I left the evidence for that on a separate comment.

diegohaz (commented):
Initial review of `fb07b8b`, diffed from the merge base with `main` (`54330d8`), then expanded to the call sites, the fixtures, and the packages that consume these helpers.

The `dom.ts` change is correct and I confirmed the platform behavior it rests on. What it does not do is deliver the user facing outcome the description leads with, and I could reproduce that in a browser.

## What I ran

- `pnpm tsc` and `pnpm lint`, both clean on this head.
- `pnpm test packages/ariakit-utils/src/dom.test.ts`, 28 passed, and with `focus.test.ts`, 39 passed.
- `pnpm dev-app` from this worktree, then `APP_PORT=4323 pnpm test src/sandbox/dom-7225` in `app`. The 9 new cases pass in Chrome, Firefox, and Safari, as the description claims. I verified the server's working directory pointed at this worktree before trusting the run.
- A Playwright probe of the raw platform behavior in Chromium 151.0.7922.34, Firefox 153.0, and WebKit 26.5.
- An instrumented `isVisible` with `process.stderr.write` per branch, to see which fallbacks the suite reaches.
- A Chromium micro benchmark of three `isVisible` shapes, 100k calls per sample, median of 7 after warmup, repeated 3 times.
- Local prototypes of the two alternatives discussed in threads, each checked with `tsc` and the focused suites, then reverted. Nothing is left in the branch.

## The premise checks out

The `Object.hasOwn` guard in `getActiveElement` is the right discriminator, and it is exact in all three engines. With `<form name="activeElement">` present and a button focused:

| Read | Chromium 151 | Firefox 153 | WebKit 26.5 |
| --- | --- | --- | --- |
| `document.activeElement === form` | `true` | `true` | `true` |
| `Object.hasOwn(document, "activeElement")` | `true` | `true` | `true` |
| prototype read `=== button` | `true` | `true` | `true` |

happy-dom does not define `activeElement` as an own property, so the guard does not misfire in the test environment either. I checked, because it does define `defaultView` that way.

The choice to discriminate `activeElement` structurally while `tagName` and friends use a runtime type check is right and worth keeping: a named `activeElement` is an element and so is the real answer, so only `Object.hasOwn` can tell them apart.

## Published inline

1. **The React crash this PR is written to remove is still there.** `packages/ariakit-utils/src/dom.ts`. `useTagName` at `packages/ariakit-react-utils/src/hooks.ts:226` throws before `isButton` is reached, so `<Ariakit.Button render={<form />}>` with `<input name="tagName">` still fails to mount on this head in all three engines. I confirmed the `isButton` fix is necessary as well as insufficient by patching the React reads and then reverting `isButton` to `main`. Registered as https://github.com/ariakit/ariakit/issues/7228. What I ask for here is a scope note in the description, not the fix.
2. **The `isVisible` geometry fallbacks are executed by nothing.** `packages/ariakit-utils/src/dom.ts:271-288`. The `checkVisibility` fallback answers first in happy-dom and in all three engines, so 18 of the 20 new lines never run in any test. Either cover them by emulating an engine without `checkVisibility`, or drop them.
3. **No component level regression backs the react changeset.** `app/src/sandbox/dom-7225/index.react.tsx`. The sandbox calls the utilities directly and renders no Ariakit component. `app/src/sandbox/focusable-click-focus/` already renders the shape #7225 names, and extending it is what the fixture lifecycle asks for first. This gap is what let finding 1 through.

## Threads validated

- [`#discussion_r3822521065`](https://github.com/ariakit/ariakit/pull/7226#discussion_r3822521065), Codex, unguarded `Object.hasOwn`: **confirmed**, Address now. The target is ES2018, `misc.ts:126-130` is a guard for this exact API, and `dom.ts:6` already imports it. I applied the swap, `tsc` clean and 28 of 28 passing. Replied with the evidence and with the one qualifier the finding lacks: `.includes` is used elsewhere in shipped source, so the general baseline argument is weaker than the specific one.
- [`#discussion_r3822682800`](https://github.com/ariakit/ariakit/pull/7226#discussion_r3822682800), consolidating the type check into `getDOMProperty`: **confirmed viable**, Address now. Prototyped with a lookup interface and generics, `tsc` clean and 39 of 39 passing. Measured: the consolidated version with `.call(element)` is the fastest of the three shapes, 6.0 to 6.6 ms against 6.7 to 7.1 ms for this branch. The auto bind you asked about is the only part that costs, so keep `.call`. It also removes an `any`: `getDOMProperty` returns `any` today because `Object.getPrototypeOf` does, and in `getActiveElement` that `any` now flows out through `as HTMLElement | null` unchecked.

## Registered follow up

https://github.com/ariakit/ariakit/issues/7228 covers the `tagName` reads outside `dom.ts`, verified pre-existing on `main` at `54330d8` and untouched by this PR. I added a comment there for two reads inside `getActiveElement` itself, `activeElement.getAttribute` at line 142 and `getElementById` at line 144, both on the `activeDescendant` path `Composite` takes, plus measurements showing `contains`, `closest`, `getAttribute`, `isContentEditable`, `scrollHeight`, `clientHeight`, `document.getElementById`, `document.querySelector`, and `document.body` all answer with elements under the same mechanism.

That comment carries the one design question I would put to you before more of these land: hardening one member per report makes every new read a new report, and the set is open ended. A single accessor that all DOM reads in `@ariakit/utils` go through would close the class rather than the instances.

## Not published, with reasons

- **React changeset scope.** I started to flag `.changeset/300-react-named-active-element.md` for describing only the `activeElement` fix while the PR also changes `isButton` and `isVisible`, which back `Command`, `Focusable`, `Button`, and the composite item filters. **Invalid**: finding 1 shows the React level `isButton` crash is not actually fixed, so the narrow claim is the accurate one. The description is what overstates the scope, and that is covered inline.
- **`@ariakit/test` and Solid changesets.** Checked and not needed. `updateInternalDependencies` is `patch` and both depend on `@ariakit/utils` through `workspace:*`, and `packages/ariakit-solid-components/src` uses none of these three helpers.
- **Missing happy-dom `test.ts` for the sandbox.** **Declined non goal.** happy-dom implements no document named getter and defines these members on the prototype, so a duplicate cannot exercise the behavior. This is the real browser only exception. I did suggest, in the sandbox comment, recording that reason next to the tests the way `focusable-click-focus/test-browser.ts:33` does.
- **Comment density in the new branches.** `getActiveElement` explains its guard; the four ternaries in `isButton` and `isVisible` do not. Folded into the thread reply, since the consolidation makes the point moot.

## Not fully validated

- `isTextbox`, `getTextboxValue`, and `getScrollingElement` read `isContentEditable`, `value`, `scrollHeight`, and `clientHeight` with no guard. I measured that a form answers those names with controls, but I did not drive any of them to a thrown error through a component, so I am recording them in #7228 as audit candidates rather than as reports.
- The `activeDescendant` reads at lines 142 and 144 are confirmed shadowable by measurement, but I did not build a `Composite` fixture that reaches them, so the end to end impact is inferred from the code path rather than observed.
- No performance CI evidence for this head. The micro benchmark above is a local Chromium measurement of `isVisible` in isolation, not the repository's perf harness.

packages/ariakit-utils/src/dom.ts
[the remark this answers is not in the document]
diegohaz: Fixed in [`bf88f492a7c0c0cb1284697d829f5908f0c394c2`](https://github.com/ariakit/ariakit/commit/bf88f492a7c0c0cb1284697d829f5908f0c394c2). `getActiveElement` now uses the existing ES2018-safe helper:

```ts
hasOwnProperty(ownerDocument, "activeElement")
```

The new regression test removes `Object.hasOwn` before it calls `getActiveElement`. The pre-fix implementation throws `Object.hasOwn is not a function`; the fixed utility suite passes all 82 tests.

packages/ariakit-utils/src/dom.ts
diegohaz: Implemented in [`bf88f492a7c0c0cb1284697d829f5908f0c394c2`](https://github.com/ariakit/ariakit/commit/bf88f492a7c0c0cb1284697d829f5908f0c394c2). The typed `getDOMMember` helper now contains the native-type fast path and the prototype fallback for `isButton` and `isVisible`:

```ts
const checkVisibility = getDOMMember(element, "checkVisibility", "function");
```

Methods still use `.call(element)`, so the hot path does not allocate bound functions. The helper also keeps the general prototype result as `unknown` instead of allowing `any` to flow through these call sites.

packages/ariakit-utils/src/dom.ts
diegohaz: Covered in [`bf88f492a7c0c0cb1284697d829f5908f0c394c2`](https://github.com/ariakit/ariakit/commit/bf88f492a7c0c0cb1284697d829f5908f0c394c2). The new test deletes `Element.prototype.checkVisibility`, shadows all three geometry members, and restores the descriptor in `finally`.

The pre-fix geometry route reaches the named control and throws:

```text
element.getClientRects is not a function
```

The fixed route passes, and the complete utility suite is green with 82 tests.

app/src/sandbox/dom-7225/index.react.tsx
diegohaz: Fixed in [`9105600af9e2e30c1f5a570932494957891b9a51`](https://github.com/ariakit/ariakit/commit/9105600af9e2e30c1f5a570932494957891b9a51). The issue sandbox now renders an Ariakit `Dialog` and verifies that focus returns from its close button to the plain button that opened it while `<form name="activeElement">` is present.

The test observes `onFocus` output instead of reading the colliding `document.activeElement`. With the pre-fix lookup restored, all three browser projects stay on `Close dialog` and fail. The fixed implementation passes all 12 sandbox cases across Chrome, Firefox, and Safari.

[Dialog focus-restoration recording](https://github.com/user-attachments/assets/8e32ce1c-6b02-401e-befc-c696bbc0917f)

The wider React `tagName` reads that prevent the suggested composed-form case remain tracked in [#7228](https://github.com/ariakit/ariakit/issues/7228).

diegohaz:
## Review feedback addressed

- [`bf88f492a7c0c0cb1284697d829f5908f0c394c2`](https://github.com/ariakit/ariakit/commit/bf88f492a7c0c0cb1284697d829f5908f0c394c2) replaces the unguarded `Object.hasOwn` call with the existing ES2018-safe helper and adds a targeted compatibility test ([thread](https://github.com/ariakit/ariakit/pull/7226#discussion_r3823525285)).
- The same commit consolidates the native-type and prototype reads in typed `getDOMMember` calls while keeping `.call(element)` for methods ([thread](https://github.com/ariakit/ariakit/pull/7226#discussion_r3823527599)).
- The same commit adds a test that removes `Element.prototype.checkVisibility` and exercises all three shadowed geometry members ([thread](https://github.com/ariakit/ariakit/pull/7226#discussion_r3823530026)).
- [`9105600af9e2e30c1f5a570932494957891b9a51`](https://github.com/ariakit/ariakit/commit/9105600af9e2e30c1f5a570932494957891b9a51) adds an Ariakit `Dialog` regression that proves focus restoration past `<form name="activeElement">` ([thread and recording](https://github.com/ariakit/ariakit/pull/7226#discussion_r3823532545)).

Verification is green: 82 utility tests, 12 issue-sandbox cases across Chrome, Firefox, and Safari, package build, app/Astro check, `tsc -b`, and the production-style app build. Red checks reproduced the missing `Object.hasOwn`, direct `getClientRects`, and failed Dialog focus-restoration paths before each fix was restored.

The PR description now has the current counts, durable browser evidence, and the [#7228](https://github.com/ariakit/ariakit/issues/7228) scope note for the wider React and event-package DOM member reads.

diegohaz (commented):
Re-review of `272792d3b`, incremental from `fb07b8bb0`, the head my previous review covered. The branch was rebased onto `f076a337e`, so #7227 is now in the baseline on both sides rather than in this diff.

Everything I raised is resolved, each with a red check rather than an assertion. The performance regression that the consolidation introduced is gone, and I confirmed that from the raw round data rather than the summary line.

## The regression is fixed, confirmed in the raw rounds

`Restore fast DOM member reads` swaps `Reflect.get(value, name)` for `value[name]` and types the key as `Extract<keyof Value, string>`. That is better than what I suggested, because the key is now checked against the value's own members instead of passing through an untyped cast.

The three CI runs tell the story cleanly:

| Head | `combobox-perf > move across items` scripting | Report |
| --- | --- | --- |
| `fb07b8bb0` | 425.3ms → 402.9ms (-5%) | No significant changes |
| `9105600af` | 401ms → 478.1ms (+19%) | Confirmed regression |
| `272792d3b` | 383.2ms → 381.3ms (-0%) | No significant changes |

I did not stop at the summary. The raw rounds for this head are flat on both sides:

| Round | Baseline scripting | Current scripting |
| --- | --- | --- |
| 1 | 386.2ms | 381.3ms |
| 2 | 380.1ms | 381.3ms |

For contrast, the regressed head's rounds were baseline 372.9 / 374.7 against current 478.0 / 458.0 in the tightly interleaved first block, so that one was a real regression rather than runner drift, and this one is a real recovery.

An isolated Chromium measurement agrees on the cause. 150k `isVisible` calls over a 500 element mixed tree, median of 9 samples, three repeats:

| Fast path | Median |
| --- | --- |
| `element.checkVisibility` named access, as at `fb07b8bb0` | 8.6 to 9.5ms |
| `value[name]` keyed access, as shipped now | 8.5 to 9.1ms |
| `Reflect.get(value, name)`, as at `9105600af` | 14.2 to 15.1ms |

So `Reflect.get` cost about 1.7x on that read, and the keyed access it replaced costs nothing against plain named access. That also answers the question you raised in the earlier thread about whether centralizing would cost anything: it does not, as long as the fast path stays an ordinary property read.

## One near-miss worth naming, and why it is not real

The report carries a diagnostic line:

> Unflagged threshold-sized changes for tab-perf > react > move across tabs: Scripting (rounds 1/2, raw 1/2, pairs 60%).

The row shows 503ms → 578.2ms (+15%) scripting. It is noise from a single contaminated round, and the raw artifact shows it plainly:

| Round | Baseline median | Current median | Current raw samples |
| --- | --- | --- | --- |
| 1 | 504.3ms | 455.2ms | 549.2, 496.1, 445.5, 455.2, 453.4 |
| 2 | 501.7ms | 701.2ms | 560.6, 519.8, 736.3, 701.2, 734.5 |

The baseline is stable at ~503ms across both rounds, and current round 1 is about 10% **faster** than it. Round 2 is bimodal between 520 and 736. The comparator demoted it correctly; no action needed, and I am recording it only so the +15% in the table is not mistaken for a finding later.

## Previous findings

All four are closed, and I verified the two that make behavioral claims rather than taking the replies at face value.

- **Codex's unguarded `Object.hasOwn`.** Now `hasOwnProperty(ownerDocument, "activeElement")`, with a regression that removes `Object.hasOwn` before calling. Resolved.
- **`isVisible` geometry fallbacks executed by nothing.** The new test deletes `Element.prototype.checkVisibility` and restores the descriptor in `finally`. I instrumented `isVisible` with `process.stderr.write` per branch and re-ran the suite: `HIT:offsetWidth` and `HIT:getClientRects` both print now, where previously only the `checkVisibility` marker ever did. The dead code is genuinely covered, not just accompanied by a test.
- **No component level regression behind the react changeset.** The sandbox now renders an Ariakit `Dialog` and asserts focus returns to its opener through `onFocus` output rather than through the colliding `document.activeElement`. I reverted `getActiveElement` to its `main` implementation in my own checkout and ran the suite against a server from that same checkout: 6 failed, 6 passed, with `restores dialog focus past a named activeElement form` failing in Chrome, Firefox, and Safari, then all 12 passing with the fix restored. Genuine red check.
- **The description overstated the fix.** The scope note landed and points at #7228.

## Local verification

`pnpm tsc` 0 errors, `pnpm lint` clean, `pnpm test packages/ariakit-utils/src/dom.test.ts packages/ariakit-utils/src/focus.test.ts` 41 passed. Full CI on this head is green with no failing checks. No unresolved review threads remain, so there were no candidate findings to revalidate from other reviewers.

## No new findings

Nothing new to raise on this head. Two small things I looked at and am deliberately not filing:

- `getDOMProperty`'s `activeElement` overload declares `Element | null`, while a prototype chain without the accessor would yield `undefined`. Unreachable in practice, since the call is guarded by an own-property check and every engine and happy-dom define the accessor, and `!activeElement?.nodeName` absorbs it either way.
- Both new tests use `try`/`finally` to restore a descriptor where `ariakit-general-code-style` prefers `using`. The rule is written for objects that already implement `Symbol.dispose`, and here one would have to be built for the purpose, so the current form is fine.

diegohaz:
Closing unmerged in favour of #7231, which reverts most of this family of defenses rather than extending it.

The reasoning, in short: where a named-property collision makes Ariakit throw, the developer gets a stack trace naming the read and renaming the control fixes it, so it is not worth defending. The members hardened here are also implausible field names. Full coverage would mean auditing roughly 250 DOM member reads, and every new read would reopen the question.

#7231 keeps only the reads where the helper cannot detect the wrong answer from the value and the failure is silent. That is `nodeType`, which was already on `main`, and `activeElement`. The work in this pull request on `isButton` and `isVisible` is intentionally not carried over: both throw, which is the supported-to-fix case.

The review of this pull request was not wasted. The `Object.hasOwn` finding, the `getDOMMember` consolidation, the `Reflect.get` performance regression and its fix, and the measurements behind all of it informed the contract #7231 now states.

PR #7227, opened 2026-08-20
Fix defaultView collisions in Portal and Combobox

### LINKED ISSUE #7218: Portal and Combobox read defaultView directly, so a form named defaultView can break them
## Context

Found while reviewing [#7217](https://github.com/ariakit/ariakit/pull/7217), which reads `nodeType`, `ownerDocument`, and `defaultView` through their interface prototype getters so a named element cannot answer those lookups. That change hardens `isElement`, `isNode`, `getDocument`, and `getWindow` in `@ariakit/utils`, but it does not change the call sites that read the same members directly instead of going through those helpers.

## Problem

A document exposes elements it names as own properties that override built-ins, so a `<form name="defaultView">` makes `document.defaultView` answer with that form. `packages/ariakit-utils/src/dom.test.ts` records this as measured on Chromium 151, Firefox 153, and WebKit 26.5.

Two stable-API paths still read that member directly, so the hardened helpers do not protect them.

[`packages/ariakit-react-components/src/portal/portal.tsx#L36`](https://github.com/ariakit/ariakit/blob/da21014aecad5ab6ce1a9ea26175a2318660180e/packages/ariakit-react-components/src/portal/portal.tsx#L36):

```ts
const HTMLElementClass = doc.defaultView?.HTMLElement;
```

When the form answers, `HTMLElementClass` is `undefined`, the `instanceof` guard fails, and `getRootElement` returns `doc.body` instead of the fullscreen element. Portals then render outside the fullscreen element, where they are not visible.

[`packages/ariakit-react-components/src/combobox/__utils.ts#L16`](https://github.com/ariakit/ariakit/blob/da21014aecad5ab6ce1a9ea26175a2318660180e/packages/ariakit-react-components/src/combobox/__utils.ts#L16):

```ts
const view = element.ownerDocument.defaultView;
if (!view) return null;
if (!view.getComputedStyle(popup).writingMode.startsWith("horizontal")) {
```

When the form answers, `view` is truthy and `view.getComputedStyle` is `undefined`, so `getSingleVerticalScrollport` throws a `TypeError` while scrolling a combobox item into view.

Two lower-risk reads exist as well. [`collection-renderer.tsx`](https://github.com/ariakit/ariakit/blob/da21014aecad5ab6ce1a9ea26175a2318660180e/packages/ariakit-react-components/src/collection/collection-renderer.tsx#L297) destructures `defaultView` and `documentElement` from `scroller.ownerDocument`, which degrades to the correct result because the shadowed values fail the `scroller === documentElement` comparison. [`tab-store.ts`](https://github.com/ariakit/ariakit/blob/da21014aecad5ab6ce1a9ea26175a2318660180e/packages/ariakit-components/src/tab/tab-store.ts#L29) reads `element.ownerDocument.activeElement` on a tab item, which would have to be rendered as a form with a control named `ownerDocument` to collide.

## What was verified

Both reads are present unchanged on `main` at [`da21014`](https://github.com/ariakit/ariakit/commit/da21014aecad5ab6ce1a9ea26175a2318660180e), so this is not introduced by [#7217](https://github.com/ariakit/ariakit/pull/7217).

The end user symptom was not reproduced. Emulating the named getter in the happy-dom suite by shadowing `document.defaultView` also breaks `dom-accessibility-api`, which reads the same member to resolve accessible names, so the queries fail before the Ariakit path runs. A reproduction needs a real browser fixture with a `<form name="defaultView">` in the preview document.

## Why 7217 can finish without this

[#7217](https://github.com/ariakit/ariakit/pull/7217) is scoped to the DOM helpers in `@ariakit/utils`. These call sites bypass those helpers entirely, they exist unchanged on `main`, and fixing them does not change what the helpers guarantee.

## Expected outcome

Route both reads through `getWindow(element)`, which resolves the view through the `Document` interface getter, then add browser coverage that places a `<form name="defaultView">` in the document for the fullscreen portal path and for combobox item scrolling.

diegohaz:
## Motivation

A form named `defaultView` can replace the document property lookup in named-access browsers. This breaks fullscreen portals and Combobox item presentation.

Fixes #7218.

## Solution

Route both affected document-window reads through the hardened `getWindow(element)` helper.

The regression sandbox now renders its form with React and exercises ordinary public `Portal`, `ComboboxSelect`, `ComboboxPopover`, and `ComboboxItem` interactions. React 19 has its own direct `ownerDocument.defaultView` read during commits, so the fixture scopes the real form name collision to the native fullscreen and selected-option focus events, then restores the normal name before React flushes an update.

A separate commit proves the userland workaround before the library fix is applied.

## Workaround

Use a different form name until the fix is available:

```html
<!-- TODO: Restore name="defaultView" after https://github.com/ariakit/ariakit/issues/7218 is fixed. -->
<form name="settingsForm">
  <!-- fields -->
</form>
```

## Testing

- `pnpm test-browser portal-combobox-7218`: six passing cases across Chrome, Firefox, and Safari.
- Temporarily restoring the old production reads makes both Chrome regressions fail. The Portal stays outside fullscreen, and Combobox emits `view.getComputedStyle is not a function`.

## Preview

[Browser evidence](https://github.com/ariakit/ariakit/pull/7227#issuecomment-5357515920) captures the fixed fullscreen Portal containment and selected Combobox item visibility. The final review-response commit changes only the regression fixture, so the production implementation shown in that evidence is unchanged.

### DIFF
--- a/app/src/sandbox/portal-combobox-7218/index.react.tsx
+++ b/app/src/sandbox/portal-combobox-7218/index.react.tsx
@@ -1,80 +1,53 @@
 import * as Ariakit from "@ariakit/react";
-import type { MouseEvent, MutableRefObject } from "react";
-import { useCallback, useLayoutEffect, useRef, useState } from "react";
-import {
-  getScrollItemIntoView,
-  useTrackComboboxSelectPresentation,
-} from "../../../../packages/ariakit-react-components/src/combobox/__utils.ts";
+import type { RefObject } from "react";
+import { useRef } from "react";
 
 const items = Array.from({ length: 30 }, (_, index) => `Item ${index + 1}`);
+const formName = "preferences";
 
-interface CollisionProps {
-  formRef: MutableRefObject<HTMLFormElement | null>;
+// React 19 reads the same named property during commits. Scope the collision
+// to native browser work so its failure does not hide Ariakit's behavior.
+function setFormName(formRef: RefObject<HTMLFormElement | null>, name: string) {
+  const form = formRef.current;
+  if (!form) return;
+  form.name = name;
 }
 
-function createDefaultViewForm() {
-  const form = document.createElement("form");
-  form.name = "defaultView";
-  form.setAttribute("aria-label", "Preferences");
-  const label = document.createElement("label");
-  label.textContent = "Theme";
-  const input = document.createElement("input");
-  input.name = "theme";
-  input.defaultValue = "System";
-  label.appendChild(input);
-  form.appendChild(label);
-  return form;
-}
-
-function AddDefaultViewCollision({ formRef }: CollisionProps) {
-  useLayoutEffect(() => {
-    // Keep the real named form across the target layout effects only. React 19
-    // has its own direct lookup that would otherwise mask the Ariakit result.
-    const form = createDefaultViewForm();
-    document.body.appendChild(form);
-    formRef.current = form;
-    return () => form.remove();
-  }, [formRef]);
-
-  return null;
-}
-
-interface RemoveCollisionProps extends CollisionProps {
-  recordResult: () => void;
-}
-
-function RemoveDefaultViewCollision({
+function PreferencesForm({
   formRef,
-  recordResult,
-}: RemoveCollisionProps) {
-  useLayoutEffect(() => {
-    recordResult();
-    formRef.current?.remove();
-    formRef.current = null;
-  }, [formRef, recordResult]);
-
-  return null;
+}: {
+  formRef: RefObject<HTMLFormElement | null>;
+}) {
+  return (
+    <form ref={formRef} name={formName}>
+      <label>
+        Theme
+        <input name="theme" defaultValue="System" />
+      </label>
+    </form>
+  );
 }
 
-function FullscreenPortal() {
+function FullscreenPortal({
+  formRef,
+}: {
+  formRef: RefObject<HTMLFormElement | null>;
+}) {
   const fullscreenHostRef = useRef<HTMLDivElement>(null);
-  const formRef = useRef<HTMLFormElement>(null);
-  const [mounted, setMounted] = useState(false);
 
   const enterFullscreen = () => {
-    void fullscreenHostRef.current?.requestFullscreen();
+    const form = formRef.current;
+    const host = fullscreenHostRef.current;
+    if (!form || !host) return;
+    setFormName(formRef, "defaultView");
+    const restoreName = () => {
+      setFormName(formRef, formName);
+    };
+    host.ownerDocument.addEventListener("fullscreenchange", restoreName, {
+      once: true,
+    });
+    void host.requestFullscreen().catch(restoreName);
   };
-
-  const recordResult = useCallback(() => {
-    const portal = document.querySelector('[id="portal/fullscreen-content"]');
-    const result = document.getElementById("portal-result");
-    if (!result) return;
-    const insideFullscreen =
-      portal?.parentElement === fullscreenHostRef.current;
-    result.textContent = insideFullscreen
-      ? "Portal parent: fullscreen"
-      : "Portal parent: body";
-  }, []);
 
   return (
     <section aria-label="Fullscreen portal">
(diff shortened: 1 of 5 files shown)


diegohaz:
Browser evidence

The fullscreen Portal remains inside the fullscreen element when a form is named `defaultView`, and the selected Combobox item scrolls into view without an exception.

ariakit-7218-portal-fixed
ariakit-7218-combobox-fixed

app/src/sandbox/portal-combobox-7218/index.react.tsx
```diff
@@ -0,0 +1,195 @@
+import * as Ariakit from "@ariakit/react";
+import type { MouseEvent, MutableRefObject } from "react";
+import { useCallback, useLayoutEffect, useRef, useState } from "react";
+import {
+  getScrollItemIntoView,
+  useTrackComboboxSelectPresentation,
+} from "../../../../packages/ariakit-react-components/src/combobox/__utils.ts";
```
diegohaz: This is not acceptable. Why are we testing implementation details?

app/src/sandbox/portal-combobox-7218/index.react.tsx
```diff
@@ -0,0 +1,195 @@
...
+  form.setAttribute("aria-label", "Preferences");
+  const label = document.createElement("label");
+  label.textContent = "Theme";
+  const input = document.createElement("input");
+  input.name = "theme";
+  input.defaultValue = "System";
+  label.appendChild(input);
+  form.appendChild(label);
+  return form;
+}
```
diegohaz: Why are we creating this like this instead of using React?

app/src/sandbox/portal-combobox-7218/index.react.tsx
```diff
@@ -0,0 +1,195 @@
+import * as Ariakit from "@ariakit/react";
```
diegohaz, line 1: This whole sandbox is terrible. It doesn't represent real world usage.

app/src/sandbox/portal-combobox-7218/index.react.tsx
diegohaz: Addressed in [36bdb412d](https://github.com/ariakit/ariakit/commit/36bdb412d2c28ec47c1286f20c12e8e56ff3bfbc). I removed the internal Combobox utility imports and direct callback invocation. The tests now drive the public `ComboboxSelect` and selected `ComboboxItem`, assert the rendered selection, and fail on the original `view.getComputedStyle is not a function` page error. All six Chrome, Firefox, and Safari cases pass.

app/src/sandbox/portal-combobox-7218/index.react.tsx
diegohaz: Addressed in [36bdb412d](https://github.com/ariakit/ariakit/commit/36bdb412d2c28ec47c1286f20c12e8e56ff3bfbc). `PreferencesForm` now renders the form and its fields as normal React JSX. React 19 has its own direct `ownerDocument.defaultView` read during commits, so the fixture scopes the real `defaultView` form name to the native fullscreen and option-focus event windows, then restores it before React flushes an update.

app/src/sandbox/portal-combobox-7218/index.react.tsx
diegohaz: Rebuilt in [36bdb412d](https://github.com/ariakit/ariakit/commit/36bdb412d2c28ec47c1286f20c12e8e56ff3bfbc) as a small preferences app with a React-rendered form, a fullscreen `Portal`, and a normal `ComboboxSelect`. The manual DOM lifecycle, sandbox result statuses, internal calls, element IDs, and test-only action buttons are gone. The regression tests now follow the same public interactions a user performs.

diegohaz:
Addressed all review feedback in [36bdb412d](https://github.com/ariakit/ariakit/commit/36bdb412d2c28ec47c1286f20c12e8e56ff3bfbc).

- Replaced the internal Combobox utility harness with public `ComboboxSelect` and `ComboboxItem` interactions.
- Rendered the preferences form and fields with React.
- Removed manual DOM insertion, layout-effect collision components, result statuses, element IDs, and test-only action buttons.
- Kept the real `defaultView` collision active only for the native event windows under test so React 19's separate direct lookup does not mask the Ariakit paths.

Verification:

- `pnpm test-browser portal-combobox-7218`: 6/6 passing across Chrome, Firefox, and Safari.
- Restoring the old production reads makes both Chrome tests fail for the expected reasons: the Portal remains outside fullscreen and Combobox emits `view.getComputedStyle is not a function`.
- Focused formatter and linter checks pass.
- [Existing browser evidence](https://github.com/ariakit/ariakit/pull/7227#issuecomment-5357515920) remains valid because the final commit changes only the regression fixture, not the product fix.

PR #7233, opened 2026-08-21
Route remaining activeElement reads through the hardened accessor

### LINKED ISSUE #7230: Dialog, ComboboxList, and the tab store read activeElement without the hardened accessor
## Context

Found while reviewing the in-flight revert of the named-property defenses, which keeps exactly one of them: `getActiveElement` reads `activeElement` through an accessor taken from the prototype, because a named form and the focused element are both elements and no runtime type check can tell them apart.

Three other reads of the same member do not go through that helper, so they still answer with the form.

## Problem

A document exposes `embed`, `form`, `iframe`, `img`, and `object` elements carrying a `name` attribute as own properties that override built-ins. With `<form name="activeElement">` on the page, `document.activeElement` answers with the form. Measured on Chromium 151.0.7922.34, Firefox 153.0, and WebKit 26.5; all three agree.

[`packages/ariakit-react-components/src/dialog/dialog.tsx#L526`](https://github.com/ariakit/ariakit/blob/4a85174708/packages/ariakit-react-components/src/dialog/dialog.tsx#L526):

```ts
const documentActiveElement = getDocument(contentElement).activeElement;
```

This backs the guard that preserves focus which escaped while the dialog's placement was still pending. The read answers with the form, `isFocusable(form)` is `false`, so the early return never fires and the dialog pulls focus back from wherever the user legitimately moved it.

[`packages/ariakit-react-components/src/combobox/combobox-list.tsx#L81`](https://github.com/ariakit/ariakit/blob/4a85174708/packages/ariakit-react-components/src/combobox/combobox-list.tsx#L81):

```ts
if (getDocument(list).activeElement !== list) return;
```

The comparison is permanently false, so the redirect that moves focus from the list container to the combobox never runs.

[`packages/ariakit-components/src/tab/tab-store.ts#L29`](https://github.com/ariakit/ariakit/blob/4a85174708/packages/ariakit-components/src/tab/tab-store.ts#L29):

```ts
const activeElement = items[0]?.element?.ownerDocument.activeElement;
```

`getFocusedTab` then always returns `undefined`, so changing `selectedId` while a tab holds DOM focus takes the `setState("activeId", ...)` branch instead of moving focus, leaving DOM focus on the previously selected tab.

## What was verified

All three reads are present unchanged on `main` at [`4a85174`](https://github.com/ariakit/ariakit/commit/4a85174708000ca5d10e16dc3b398504c71095ee), and the revert touches only `packages/ariakit-utils/src/dom.ts`, its test file, two sandboxes, and changesets. None of this is introduced by it.

`getActiveElement` is not a drop-in replacement at `dialog.tsx#L526`. That site deliberately avoids recursing into frames, because it decides ownership in the dialog's own document where iframe focus is represented by the frame element.

## Why the revert can finish without this

That pull request is a revert. The read it keeps is proven by a unit case and by a browser sandbox that fails in all three engines without it. These three sites are separate call sites in two other packages, they were equally unguarded before it, and hardening them needs a new exported helper rather than a change to what the revert touches.

## Expected outcome

Export the cached accessor from `@ariakit/utils` so a document's active element can be read without frame recursion, and route these three sites through it. Something like:

```ts
export function getDocumentActiveElement(ownerDocument: Document) {
  if (!activeElementGetter) return ownerDocument.activeElement;
  return activeElementGetter.call(ownerDocument) as Element | null;
}
```

That also removes the per-call ternary from `getActiveElement`, which sits on hot focus paths.

Add real browser coverage for each site, in the shape of `app/src/sandbox/dialog-form-named-active-element/`. happy-dom implements no document named getter, so a happy-dom duplicate cannot exercise this.

### Workaround

Do not give a `form`, `embed`, `iframe`, `img`, or `object` element a `name` of `activeElement`. Use a collision free name and remap it where the form data is prepared:

```html
<form name="activeElementForm"></form>
```

When you do not control the colliding element, each affected site can be overridden from consumer code instead: https://github.com/ariakit/ariakit/issues/7230#issuecomment-5366964983

### LINKED ISSUE #7234: getActiveElement's JSDoc summary describes a read it deliberately does not perform
## Context

Found during the pre-commit review of #7233, which routes the three remaining unguarded `activeElement` reads through `getActiveElement`. Two independent reviewers flagged the helper's JSDoc summary while reviewing that change. It is not caused by it.

## Problem

[`packages/ariakit-utils/src/dom.ts#L128`](https://github.com/ariakit/ariakit/blob/fbd1b71609/packages/ariakit-utils/src/dom.ts#L128) summarizes the helper as:

```ts
/**
 * Returns `element.ownerDocument.activeElement`.
 */
export function getActiveElement(
```

That is the one thing it deliberately does not do, and it has been inaccurate in three separate ways:

- It reads an accessor taken from the `Document` prototype chain rather than the member itself, which is the whole point of [#7228](https://github.com/ariakit/ariakit/issues/7228). A document naming an element `activeElement` is exactly the case the plain member read gets wrong.
- It resolves focus that lives inside a frame into that frame's own document, so the return value is frequently not the given document's `activeElement`.
- It resolves `aria-activedescendant` when asked, returning an element that never had DOM focus at all.

The summary is generated into the published reference at `packages/ariakit-utils/readme.md`, so consumers read it there too.

## What was verified

The wording is pre-existing on `main` at [`fbd1b71`](https://github.com/ariakit/ariakit/commit/fbd1b71609f12764204e460c920b1bdc6e33c75d), where the same line reads identically. It predates #7233 and is not introduced by it. The frame recursion has been there since the helper was written, and the accessor read landed in [`1f3f406`](https://github.com/ariakit/ariakit/commit/1f3f406ac).

## Why #7233 can finish without this

That pull request hardens three call sites. It changes the helper's second parameter into an options object and documents both options with their own JSDoc and `@default` values, so the generated reference now renders a complete `GetActiveElementOptions` block directly above the summary. Rewriting the summary sentence itself is a separate docs change that would touch no call site and no behavior, and folding it in would have forced another readme regeneration for a line no caller depends on.

One thing that pull request does change: its new `@example` for `frame: false` now sits directly beneath the summary, so the contradiction is visible within a single screenful of the generated reference. That makes the existing defect easier to notice, but does not create it.

## Expected outcome

Rewrite the summary so it describes what the helper actually returns, then regenerate `packages/ariakit-utils/readme.md` with `pnpm --filter @ariakit/utils run docs`. Something that names the accessor read and the two resolutions, rather than an expression the function does not evaluate.

No changeset is needed, since no behavior or signature changes.

diegohaz:
## Motivation

A document exposes `embed`, `form`, `iframe`, `img`, and `object` elements carrying a `name` attribute as own properties that override built-ins. With `<form name="activeElement">` on the page, `document.activeElement` answers with the form instead of the focused element, in Chromium, Firefox, and WebKit alike.

`getActiveElement` in `@ariakit/utils` already defends against this by reading an accessor taken from the `Document` prototype chain, because a named form and the focused element are both elements and no runtime type check can tell them apart. Three other reads of the same member do not go through that helper, so they still answer with the form:

```ts
// packages/ariakit-react-components/src/dialog/dialog.tsx
const documentActiveElement = getDocument(contentElement).activeElement;

// packages/ariakit-react-components/src/combobox/combobox-list.tsx
if (getDocument(list).activeElement !== list) return;

// packages/ariakit-components/src/tab/tab-store.ts
const activeElement = items[0]?.element?.ownerDocument.activeElement;
```

Each read fails silently rather than throwing. [`Dialog`](https://ariakit.com/reference/dialog) pulls focus back from wherever the user legitimately moved it, [`ComboboxList`](https://ariakit.com/reference/combobox-list) never hands focus from the list back to the combobox, and changing the selected tab while a [`Tab`](https://ariakit.com/reference/tab) holds DOM focus leaves focus on the previously selected tab.

Fixes #7230
Closes #7234

## Solution

Route all three reads through `getActiveElement`, which already reads the accessor.

Two of them can call it as it stands. `combobox-list.tsx` asks whether the list itself holds focus, and `tab-store.ts` looks the answer up among the tab items, so resolving through a focused frame lands both comparisons the same way it does today.

`dialog.tsx` cannot, because it decides ownership in the dialog's own document, where focus inside a frame is represented by the frame element. Rather than give it a module-local read, which would have to duplicate the accessor lookup outside `@ariakit/utils`, `getActiveElement` now takes an options object as its second parameter, and `frame: false` stops the lookup at that document.

```ts
export interface GetActiveElementOptions {
  frame?: boolean;
  activeDescendant?: boolean;
}
```

That replaces the previous positional `activeDescendant` boolean, so the one existing caller of it now reads `getActiveElement(dialog, { activeDescendant: true })`.

The three reads become:

```ts
// packages/ariakit-react-components/src/dialog/dialog.tsx
const activeElement = getActiveElement(contentElement, { frame: false });

// packages/ariakit-react-components/src/combobox/combobox-list.tsx
if (getActiveElement(list) !== list) return;

// packages/ariakit-components/src/tab/tab-store.ts
const activeElement = getActiveElement(element);
```

The dialog site also drops its `isElement` narrowing, because `getActiveElement` already answers `null` rather than a non-element.

### Coverage

Each site gets a browser regression in the shape the issue asks for. happy-dom implements no document named getter, so a happy-dom duplicate cannot exercise this and none was added.

- `app/src/sandbox/dialog-form-named-active-element/` gains a sibling variant, next to the existing case from #7225.
- `app/src/sandbox/combobox-list-form-named-active-element/` is new.
- `app/src/sandbox/tab-form-named-active-element/` is new.

The reproduction landed first, in [`915feb86e`](https://github.com/ariakit/ariakit/commit/915feb86e), so CI recorded all three failing in Chrome, Firefox, and Safari before the fix.

`packages/ariakit-utils/src/dom.test.ts` covers the options object itself, asserting that a focused frame resolves through by default and answers with the frame element under `frame: false`, that the caller's options survive the frame hop, and that `aria-activedescendant` is resolved only when asked.

It also picks up the fix for #7234, which this pull request's own review surfaced: `getActiveElement`'s JSDoc summary named `element.ownerDocument.activeElement`, the one expression the helper deliberately never evaluates.

## Workaround

The complete workaround is to not give an `embed`, `form`, `iframe`, `img`, or `object` element a `name` of `activeElement`. Use a collision free name and remap it where the form data is prepared. This is the only option here that preserves all normal component behavior, so prefer it whenever you control the markup.

```html
<!-- TODO: The remap can go once the fix ships. -->
<!-- https://github.com/ariakit/ariakit/issues/7230 -->
<form name="activeElementForm"></form>
```

When you do not control the colliding element, for example because it comes from a third party embed, from CMS injected markup, or from a legacy form whose name is part of a serialization contract, each affected site can be overridden from consumer code instead. Every override below is narrower than the behavior it stands in for, so read the limitations before applying one.

### Dialog

Decide in your own code whether focus already escaped, and tell the dialog to skip its auto focus when it did. The `:focus` selector answers the focused element without going through the member the named form shadows.

```tsx
const dialogRef = useRef<HTMLDivElement>(null);

// TODO: Remove once the fix ships.
// https://github.com/ariakit/ariakit/issues/7230
const hasEscapedFocus = () => {
  const dialogElement = dialogRef.current;
  if (!dialogElement) return false;
  const focused = dialogElement.ownerDocument.querySelector(":focus");
  if (!focused) return false;
  if (dialogElement.contains(focused)) return false;
  if (dialog.getState().disclosureElement?.contains(focused)) return false;
  return true;
};

<Ariakit.Dialog
  ref={dialogRef}
  store={dialog}
  autoFocusOnShow={() => !hasEscapedFocus()}
/>;
```

Limitations. The callback answers earlier than the library does. Ariakit evaluates [`autoFocusOnShow`](https://ariakit.com/reference/dialog#autofocusonshow) synchronously while the dialog's effect runs, and only places focus in a microtask after that, so this override sees only escapes that already happened by the time the effect ran. An escape that lands between the effect and that microtask, for example from a parent or sibling effect that focuses an error summary or a live region, stays invisible and the dialog still pulls focus.

Returning `false` also does more than decline one focus call. It leaves the dialog's subtree with auto focus disabled, so an Ariakit [`Focusable`](https://ariakit.com/reference/focusable) descendant with `autoFocus` is suppressed too, and it skips [`initialFocus`](https://ariakit.com/reference/dialog#initialfocus), the `[data-autofocus]` lookup, the first tabbable fallback, and the fallback that focuses the dialog element itself.

The disclosure check above is required, because the disclosure normally still holds focus when the dialog opens and the dialog counts that as its own rather than as an escape. `:focus` also reads only the light tree of one document, so unlike the library it does not walk into shadow roots, and it has no counterpart for the library's nested dialog ownership check, for its focusability check on the deepest active element, or for the elements a dialog claims through [`getPersistentElements`](https://ariakit.com/reference/dialog#getpersistentelements), which the library also counts as inside itself.

### ComboboxList

Move focus back to the combobox yourself when the list receives it. `ComboboxList` calls your `onFocus` prop before its own handler, so your redirect is queued first and wins.

```tsx
const comboboxRef = useRef<HTMLInputElement>(null);

<Ariakit.ComboboxList
  onFocus={(event) => {
    if (event.target !== event.currentTarget) return;
    // TODO: Remove once the fix ships.
    // https://github.com/ariakit/ariakit/issues/7230
    queueMicrotask(() => comboboxRef.current?.focus());
  }}
/>;
```

Limitations. This skips three guards the library applies. It ignores a consumer that calls `event.preventDefault()` to opt out of the redirect, it does not check that the combobox is focusable, and it does not re-check that the list still holds focus when the microtask runs. The focusability check is the one that matters most in practice: a modal [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) may render the combobox inert, and moving focus there would fight the dialog's focus containment. Do not apply this override inside a modal popover.

### Tab

Use [`select`](https://ariakit.com/reference/use-tab-store#select), which is [`setSelectedId`](https://ariakit.com/reference/use-tab-store#setselectedid-1) plus a move, so it always follows the selection with focus.

```tsx
// TODO: Restore `setSelectedId` once the fix ships.
// https://github.com/ariakit/ariakit/issues/7230
tab.select(id);
```

Limitations. [`setSelectedId`](https://ariakit.com/reference/use-tab-store#setselectedid-1) moves focus only when a tab already holds it, while [`select`](https://ariakit.com/reference/use-tab-store#select) always moves it. Apply this only where a tab is known to hold focus, such as a shortcut handled inside the tab list. Applying it to a button outside the tabs, a router effect, or a state sync would pull focus into the tab list on every programmatic selection change, which is worse than the bug.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the failing-reproduction direction</summary>

**Assessment**

Issue severity is medium and the failure is silent: focus lands in the wrong place across `Dialog` and the components built on it, `ComboboxList`, and `Tab`. Expected frequency is rare, because it needs an `embed`, `form`, `iframe`, `img`, or `object` element named exactly `activeElement`. Supported scope is settled by the issue itself, which states that leaving three of the four read sites unguarded is an inconsistency rather than a declined edge case, and the guarded contract these reads sit under already ships.

Implementation and maintenance complexity is low and entirely test-side: two new sandbox directories plus a sibling variant in the existing `dialog-form-named-active-element` fixture, with no library machinery, no abstraction, no public contract, and no new branches.

Across three review rounds there was no correctness finding. Round 1 raised comment placement, a missing rationale, one redundant prop, and two consistency nits. Round 2 raised one durability finding about undocumented load-bearing props, resolved by a comment plus one precondition assertion after both props were settled empirically. Round 3 returned no blocking findings and two comment-accuracy nits. The finding rate decayed monotonically, so the cycles reflect thoroughness about fixture comments rather than a contested design.

Two structural causes explain the repetition, both external to the design. Playwright's `toBeFocused` compares `getRootNode().activeElement` against the node, so it can never pass on a page carrying `<form name="activeElement">`, which forces every fixture in this family away from the repository's normal focus assertion and onto an `<output>` focus-history probe. And the fixtures need several props whose necessity is invisible locally, which correctly attracts requests for in-file rationale.

Simplification levers were considered and rejected. Merging the three fixtures would put one named form on a page with three unrelated widgets and would break the component-first naming rule. Covering the combobox and tab sites through `packages/ariakit-utils/src/dom.test.ts` instead would prove the helper rather than prove that each call site routes through it, which is the regression these tests exist to catch, and happy-dom implements no document named getter, so no cheaper automated surface exists.

One residual risk is named and intentionally not addressed: the dialog fixture's red state depends on React committing `autoFocus` before the dialog's queued auto-focus. If that ordering ever inverted, the assertions would pass without deciding anything. Detecting it would need a fixture-level ordering probe, which is more machinery than a three-engine-verified red state justifies.

**Decision**

Continue the current direction. Apply the two round 3 accuracy nits and take the commit. No machinery to remove, nothing to revert, no contract to narrow, and no disposition to reverse. This is not a material direction change, so the cycle count does not reset.

</details>


.changeset/300-react-named-active-element-reads.md
```diff
@@ -0,0 +1,14 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Focus decisions on a page with a form named `activeElement`
```
diegohaz: This is an edge case bug, so we don't need a multi-paragraph changeset. Either turn this into a single sentence changeset or split into a few single sentence changesets.

packages/ariakit-utils/src/dom.ts
```diff
@@ -124,12 +124,31 @@ export function getWindow(
...
 /**
  * Returns `element.ownerDocument.activeElement`.
+ * @example
+ * // Focus inside a frame, as the frame element rather than the inner element.
+ * getActiveElement(document.getElementById("dialog"), { frame: false });
  */
 export function getActiveElement(
   node?: Node | null,
-  activeDescendant = false,
+  options?: GetActiveElementOptions,
```
diegohaz:
```suggestion
  { frame = true, activeDescendant = false }: GetActiveElementOptions = {},
```

.changeset/300-react-named-active-element-reads.md
diegohaz: Fixed in [`d3492d6c`](https://github.com/ariakit/ariakit/commit/d3492d6cb8175dfe5c67456b57cc643238ff5634): split into three single-sentence entries, one per affected component family, and matched the wording of the sibling entries already describing this bug class.

I left [`100-utils-active-element-options.md`](https://github.com/ariakit/ariakit/blob/91154013995f6c51ab5865fb7b3b53aa0487a826/.changeset/100-utils-active-element-options.md) multi-paragraph, since the breaking-change format wants the `**BREAKING**` sentence plus `Before:`/`After:` blocks. Happy to collapse that one too if you meant both.

packages/ariakit-utils/src/dom.ts
diegohaz: Fixed in [`e82ec57e`](https://github.com/ariakit/ariakit/commit/e82ec57eb093cc395cd2fec7d33d80a017519c0d): applied as suggested, so the use sites read `frame` and `activeDescendant` instead of `options?.frame !== false`.

One consequence worth flagging: the recursion can no longer forward the caller's object, so it rebuilds `{ frame, activeDescendant }` by hand. Every member of `GetActiveElementOptions` is optional, so dropping one there would type-check silently, and no test combined a frame hop with `activeDescendant`. The same commit extends the frame case in `dom.test.ts` to cover it. Red-checked by deleting `activeDescendant` from that literal: `pnpm tsc` reported 0 errors while the test failed.

I also measured the `= {}` allocation before applying this, since it sits on the focus paths in `focus.ts` and `composite/utils.ts`: 0.026 ns per call over 50M iterations, and 0.0 ns per call under `--no-opt`. The literal never escapes, so it is scalar-replaced.

diegohaz:
**Fixed**

- [`d3492d6c`](https://github.com/ariakit/ariakit/commit/d3492d6cb8175dfe5c67456b57cc643238ff5634) ([reply](https://github.com/ariakit/ariakit/pull/7233#discussion_r3828833948)): split the React changeset into three single-sentence entries, one per affected component family, matching the wording of the sibling entries already describing this bug class. The `@ariakit/utils` entry stays multi-paragraph because it is a breaking change.
- [`e82ec57e`](https://github.com/ariakit/ariakit/commit/e82ec57eb093cc395cd2fec7d33d80a017519c0d) ([reply](https://github.com/ariakit/ariakit/pull/7233#discussion_r3828834602)): destructured the `getActiveElement` options. Since the recursion now rebuilds the object by hand and every member is optional, the same commit adds coverage that the caller's options survive a frame hop, red-checked by dropping one and confirming `tsc` stays green while the test fails.

The same push also carries [`91154013`](https://github.com/ariakit/ariakit/commit/91154013995f6c51ab5865fb7b3b53aa0487a826), which fixes #7234 at your request, so this pull request now closes it.

Issue #7240, opened 2026-08-22
Adopt Oxlint React Compiler rules


diegohaz:
## Context

Oxlint 1.79 replaces the opt-in `react/react-compiler` rule with individual rules assigned to standard lint categories. Ariakit enables the `correctness` and `suspicious` categories, so updating from Oxlint 1.78 to 1.79 reports 77 errors and 30 warnings on unchanged source.

The dependency update keeps the eight rules that currently report diagnostics disabled:

```ts
"react/exhaustive-effect-dependencies": "off",
"react/hooks": "off",
"react/immutability": "off",
"react/memo-dependencies": "off",
"react/purity": "off",
"react/refs": "off",
"react/set-state-in-effect": "off",
"react/use-memo": "off",
```

The remaining React Compiler rules stay enabled, so the update preserves new coverage where current source is already clean.

## Problem

Current package, example, sandbox, and app code does not pass the eight rules above. Updating the dependency and resolving all diagnostics in one pull request would mix a broad React code migration into a tooling update.

This is pre-existing behavior exposed by Oxlint 1.79 on `main`. Oxlint 1.78 passes the same source. The dependency update can ship safely with only the reporting rules disabled because the repository retains a clean lint baseline while enabling every other newly categorized rule.

## Expected outcome

Review the diagnostics for each disabled rule, distinguish valid React Compiler constraints from upstream false positives, update code and focused tests where appropriate, then enable the rules in `oxlint.config.ts`.

Upstream context:

- [Oxlint 1.79 release notes](https://github.com/oxc-project/oxc/blob/main/npm/oxlint/CHANGELOG.md#1790---2026-08-18)
- [React Compiler rule split](https://github.com/oxc-project/oxc/pull/25500)

PR #7242, opened 2026-08-22
Adopt React effect state lint rule

diegohaz:
## Motivation

Oxlint reports 17 `react/set-state-in-effect` diagnostics across shipped package code, the app, examples, and regression sandboxes. The rule remains globally disabled, so the effect-state portion of #7240 has no repository-wide enforcement.

## Solution

This change removes only the `react/set-state-in-effect` override from `oxlint.config.ts` and audits every baseline diagnostic.

Derived or event-driven state now updates during render or in the event that owns the transition. The copy button also associates each clipboard request with its text so an older completion cannot mark a newer value as copied:

```tsx
setCopyState((currentState) => {
  if (currentState.text !== text) return currentState;
  return { text, status: "copied" };
});
```

Effects that must synchronize with an external system or a deliberate fixture lifecycle retain their behavior behind narrow, documented lint suppressions. These include DOM measurement and attributes, browser capability and media-query state, external store cycles, hydration boundaries, portals, and regression-fixture sequencing.

No changeset is included because the shipped package edits only document existing synchronization behavior. The executable refactors are private app and fixture code and preserve their existing observable behavior, apart from ignoring stale clipboard completions.

## Testing

- `pnpm lint`
- `pnpm tsc`
- Focused React 19 Vitest: 60 tests
- Focused React 18 Vitest: 48 tests
- Affected baseline Playwright suites: 291 tests across Chrome, Firefox, Safari, and iOS
- `pnpm build`


app/src/components/copy-code.react.tsx
```diff
@@ -27,18 +27,19 @@ export function CopyCode({
   ...props
 }: CopyCodeProps) {
   const [state, setState] = useState<"idle" | "copied">("idle");
+  const [previousText, setPreviousText] = useState(text);
+
+  if (previousText !== text) {
+    setPreviousText(text);
+    setState("idle");
+  }
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`4935ab4f8`](https://github.com/ariakit/ariakit/commit/4935ab4f898b6dbdf44fdfd4326a656ff50604b3): copied status is now associated with the request text, so a completion is ignored after the text changes.

diegohaz:
**Fixed**

- [`4935ab4f8`](https://github.com/ariakit/ariakit/commit/4935ab4f898b6dbdf44fdfd4326a656ff50604b3) ([reply](https://github.com/ariakit/ariakit/pull/7242#discussion_r3835688162)): Tied clipboard completion to its request text so a stale completion cannot mark a newer value as copied.

app/src/sandbox/copy-code-stale-request/index.react.tsx
```diff
@@ -0,0 +1,14 @@
+import { useState } from "react";
```
diegohaz: We don't need this sandbox (no need to test CopyCode)

diegohaz: Removed the sandbox and its test in [`493d36649`](https://github.com/ariakit/ariakit/commit/493d366497f97ecfa4ef9971b79fb5252efad8c4). I also removed the related test and recording claims from the PR description.

PR #7248, opened 2026-08-22
Adopt React ref and immutability lint rules

### LINKED ISSUE #7240: Adopt Oxlint React Compiler rules

diegohaz:
## Motivation

Oxlint 1.79 split the React Compiler check into individual rules. The earlier work for #7240 adopted the hook-contract, effect-state, dependency, and purity rules. `react/refs` and `react/immutability` remained disabled because they reported valid render-phase hazards together with deliberate test instrumentation and React-only false positives in Solid files.

Closes #7240.

An earlier revision of this pull request scoped the rules by moving the `react` plugin out of the root `plugins` list and declaring it only inside an override. That is not equivalent. In oxlint 1.79 a plugin introduced only by an override contributes nothing beyond the rules that override names, so every React rule that was not listed stopped running everywhere. Measured with `oxlint --print-config`:

| config | enabled `react/*` rules |
| --- | --- |
| before this series | 38 |
| plugin declared only in an override | 0 |
| this pull request | 40 |

The rules that silently stopped running included `react/jsx-key`, `react/no-children-prop` and `react/exhaustive-deps`. That revision removed twelve `exhaustive-deps` directives as stale, but only three of them were genuinely dead. The other nine were suppressing live errors and are restored here.

## Solution

Keep the `react` plugin at the top level, where a declared plugin does activate its category rules, and scope only the individual rule that needs scoping.

Every React rule stays on everywhere by default. Solid gets a single exemption, for the one rule measured misfiring on it: Solid holds an element reference by reassigning a local, which `react/immutability` reads as mutating a value after render.

```ts
overrides: [
  {
    files: ["**/*.solid.*", "packages/ariakit-solid*/**"],
    rules: {
      "react/immutability": "off",
    },
  },
];
```

Against the config before this series the change is: delete the two lines that parked `react/immutability` and `react/refs`, which is the stated goal of #7240, add that one override, and add two `options` entries. The three pre-existing overrides are unchanged.

- `options.reportUnusedDisableDirectives: "error"` makes a suppression that stops matching a build failure. This is the only mechanism that would have caught the regression above at config-change time, before any source line moved. It surfaced five directives that were already dead, which this pull request removes.
- `options.denyWarnings: true` makes warnings fail CI. Roughly a quarter of the enabled rules are warnings, so they could not fail a build before.

`--disable-nested-config` stays. The upstream issue it works around, oxc-project/oxc#20354, is closed but still reproduces on 1.79.0 when a nested directory carries `options.typeAware` and is linted by explicit path from the outer root.

The `createElement` call in `vitest.setup.framework.ts` now passes its child as the third argument instead of a `children` prop, which removes a `react/no-children-prop` suppression rather than restoring it.

No changeset is included because the production changes preserve existing behavior and the remaining changes affect examples, test sandboxes, or lint configuration.

## Known exceptions

- `react/immutability` is off on `**/*.solid.*` and `packages/ariakit-solid*/**`. It was already off repo-wide before this series, so Solid loses nothing relative to `main`.
- Other React rules, including other React Compiler rules and plain rules such as `react/jsx-key`, still run on Solid source and can misfire there. That behavior predates this pull request and is tracked in #7250, which also records that `denyWarnings` now makes the warning-level ones fail CI.
- `react/exhaustive-deps` does not analyze `useSafeLayoutEffect`, so 48 of the library's 132 effect call sites are unchecked by dependency linting. That gap predates this pull request and is tracked in #7253. It is the reason three of the twelve restored directives were dead.
- `no-unnecessary-type-assertion` is disabled for `app/src/lib`. CI lints without the generated `app/.astro` types, so `astro:content` resolves `CollectionEntry` to `any` there and a narrowing assertion looks unnecessary.

<details>
<summary>Cycle 3: Keep state-based key stabilization and simplify equality</summary>

**Assessment**

The first review cycles restored the existing form cache and menu render helper, then removed one unused suppression. The third cycle found that sparse or JavaScript-supplied `undefined` subscription keys made the new render-state update retry indefinitely. These keys are outside the typed `PropertyKey[]` contract and should be rare, but the pull request would turn previously non-crashing input into a render failure.

The state-based key cache remains small and removes the production render-time ref hazard. The `undefined` equality guard predates this lint work and became unnecessary when the comparator was generalized to accept unknown values.

**Decision**

Keep the state-based design and remove the obsolete guard so equality is reflexive for sparse entries. Add focused public-hook regression coverage for both keyed hooks. Do not add key validation or filtering, and do not define support for other invalid key values.

</details>

<details>
<summary>Cycle 6: Keep the scoped plugin and close semantic path gaps</summary>

**Assessment**

The sixth finding-bearing cycle showed that a narrow extension pattern missed names such as `*.react.test.tsx`, while JSX-only example scopes missed React hooks in `.ts` files. The production refactors remain contained and well covered; this finding concerns lint path coverage, not their runtime design.

Oxlint 1.79 does not apply top-level category activation to a plugin introduced only by an override. Mirroring every categorized React rule would be version-coupled and was rejected during review.

**Decision**

Keep React out of the root plugin list. Enable the eight rules tracked by #7240 in an override that matches `**/*.react.*` and the repository's React workspaces, while excluding `**/*.solid.*`. Do not restore global suppressions or rewrite Solid semantics.

</details>

## Review gate reassessment

<details>
<summary>Pass 6: Keep the scoped eight-rule design and remove stale suppressions</summary>

**Assessment**

Scoping the React plugin makes older suppressions for React rules outside #7240 inactive. This has no runtime effect, but unused directives would misrepresent lint coverage. The affected shared framework setup also contains Solid code, so applying the React plugin to that whole file would recreate the cross-framework problem.

**Decision**

Remove the stale `react/no-children-prop` and classic `exhaustive-deps` directive entries. Keep active #7240 rule names in mixed directives. Do not add file-specific exceptions, restore the global plugin, mirror all categorized React rules, or change unrelated loader behavior.

</details>

<details>
<summary>Cycle 3 (top-level plugin direction): Narrow the Solid override to the one measured rule</summary>

**Assessment**

Three consecutive gate passes produced findings, and all three landed on the same artifact: the comment above the Solid override and the rule list it described. Each rewrite replaced one false claim with another, which is the signature of a list that has no true description.

The list was the previous revision's React-override allowlist, inverted. The two sets are identical, so it encoded no property of Solid. Measuring it directly: with the Solid override removed, the only React diagnostics on Solid source anywhere in the repository are three `react/immutability` reports in `app/src/sandbox/create-ref-6349/index.solid.tsx`. Nothing from the other seven rules.

Worse, six of the eight were already enforcing on Solid before this series (`react/purity`, `react/set-state-in-effect` and `react/use-memo` as errors; `react/hooks`, `react/memo-dependencies` and `react/exhaustive-effect-dependencies` as warnings), each adopted repo-wide with no Solid exception in #7242, #7243 and #7246. Disabling them for Solid silently reversed three merged decisions and reintroduced the same silent-coverage-loss failure mode this pull request exists to fix.

Severity of the underlying issue is low and its frequency is one file, but the maintenance cost of the wrong shape is permanent: every pre-emptively disabled rule is a silent false negative on Solid, and the list would need re-auditing whenever oxlint adds a React rule. An armed rule that misfires fails loudly on the commit that introduces it.

An independent complexity review, run with no shared context, reached the same conclusion and confirmed the identical-set finding.

**Decision**

Narrow the Solid override to `react/immutability` alone, and state the mechanism rather than a category claim. Keep the reactive policy: a rule joins the list only once it actually reports. Leave the remaining React rules armed on Solid and track the known latent misfires in #7250 rather than pre-emptively disabling them.

Separately, abandon the `react/exhaustive-deps` retirement that this revision had bundled in. Measurement showed the React Compiler rules report none of its ten diagnostics, including two genuine missing dependencies, so the supersession claim did not hold. The rule stays enabled and nine suppressions are restored. Three of the merge base's twelve proved dead under `reportUnusedDisableDirectives`, each having sat on a `useSafeLayoutEffect` call, which the rule does not analyse.

</details>

## Verification

- Full lint, formatting, and TypeScript checks
- Full test suite: 268 files, 1918 tests passed
- `oxlint --print-config` comparison against the config before this series: the only rule severity changes are `react/immutability` and `react/refs` moving from off to on
- Coverage probes confirming React rules reach `packages/ariakit-react*`, `packages/ariakit-test`, framework-agnostic packages, `examples/`, `guide/`, `app/src`, `nextjs/`, `templates/react`, `scripts/`, and the repository root
- Confirmed `react/purity` still reports on Solid source, so the narrowed override loses no enforcement that `main` had
- Confirmed both new options take effect: an unused directive and a warning-only violation each fail the run

### DIFF
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -1,9 +1,19 @@
 import { defineConfig } from "oxlint";
 
 export default defineConfig({
-  plugins: ["typescript", "import"],
+  // Declaring `react` only in an override would enable just the rules that
+  // the override names, silently disabling every other React rule, including
+  // `react/jsx-key` and `react/no-children-prop`.
+  plugins: ["typescript", "react", "import"],
   options: {
     typeAware: true,
+    // A suppression that stops matching is how lint coverage disappears
+    // unnoticed, so treat an unused directive as an error rather than as
+    // cleanup.
+    reportUnusedDisableDirectives: "error",
+    // Roughly a quarter of the enabled rules are warnings, which would
+    // otherwise never fail CI.
+    denyWarnings: true,
   },
   ignorePatterns: ["website", "**/*.astro"],
   categories: {
@@ -17,8 +27,10 @@ export default defineConfig({
     "import/namespace": "off",
     "no-unsafe-type-assertion": "off",
     "no-unassigned-import": "off",
+    "react-in-jsx-scope": "off",
     "no-shadow": "off",
     "no-underscore-dangle": "off",
+    "iframe-missing-sandbox": "off",
     "consistent-return": "off",
     "no-unnecessary-type-arguments": "off",
     "consistent-type-imports": ["error", { fixStyle: "separate-type-imports" }],
@@ -42,40 +54,32 @@ export default defineConfig({
       },
     },
     {
-      // Disable this rule for the app because some types depend on the app
-      // being built first, and linting may run before the app is built.
+      // We have our own `forwardRef` implementation that doesn't need the `ref`
+      // parameter, which leads to false positives.
+      files: ["packages/ariakit-react-components/src/**/*.{ts,tsx}"],
+      rules: {
+        "forward-ref-uses-ref": "off",
+      },
+    },
+    {
+      // These types are generated into the gitignored `app/.astro` directory,
+      // so CI always lints without them and `astro:content` resolves
+      // `CollectionEntry` to `any`. That widens unions for the first rule and
+      // makes a narrowing assertion look unnecessary for the second.
       files: ["app/src/lib/**/*.ts"],
       rules: {
         "no-redundant-type-constituents": "off",
+        "no-unnecessary-type-assertion": "off",
       },
     },
     {
-      files: [
-        "**/*.react.*",
-        "{examples,nextjs,packages/ariakit-react*,packages/ariakit-test,templates/react}/**/*.{js,jsx,ts,tsx}",
-      ],
-      excludeFiles: ["**/*.solid.*"],
-      plugins: ["typescript", "react", "import"],
+      // Solid reassigns a variable to hold an element reference, which this
+      // rule reads as mutating a value after render. This override disables
+      // nothing else, so add a rule here only once it reports on Solid.
+      // https://github.com/ariakit/ariakit/issues/7250
+      files: ["**/*.solid.*", "packages/ariakit-solid*/**"],
       rules: {
-        "react/exhaustive-effect-dependencies": "warn",
-        "react/hooks": "warn",
-        "react/iframe-missing-sandbox": "off",
-        "react/immutability": "error",
-        "react/memo-dependencies": "warn",
-        "react/purity": "error",
-        "react/react-in-jsx-scope": "off",
-        "react/refs": "error",
-        "react/set-state-in-effect": "error",
-        "react/use-memo": "error",
-      },
-    },
-    {
-      // We have our own `forwardRef` implementation that doesn't need the
-      // `ref` parameter, which leads to false positives.
-      files: ["packages/ariakit-react-components/src/**/*.{ts,tsx}"],
-      plugins: ["typescript", "react", "import"],
-      rules: {
-        "react/forward-ref-uses-ref": "off",
+        "react/immutability": "off",
       },
     },
   ],
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -12,10 +12,6 @@ export default defineConfig({
     pedantic: "off",
   },
   rules: {
-    // Adopt these React Compiler checks separately from this update.
-    // https://github.com/ariakit/ariakit/issues/7240
-    "react/immutability": "off",
-    "react/refs": "off",
     // Type-only exports are incorrectly reported as missing.
     // https://github.com/oxc-project/oxc/issues/13258
     "import/namespace": "off",
@@ -41,6 +37,20 @@ export default defineConfig({
     ],
   },
   overrides: [
+    {
+      // React Compiler rules do not model Solid component semantics.
+      files: ["**/*.solid.*"],
+      rules: {
+        "react/exhaustive-effect-dependencies": "off",
+        "react/hooks": "off",
+        "react/immutability": "off",
+        "react/memo-dependencies": "off",
+        "react/purity": "off",
+        "react/refs": "off",
+        "react/set-state-in-effect": "off",
+        "react/use-memo": "off",
+      },
+    },
     {
       files: ["*.d.ts"],
       rules: {
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -1,7 +1,7 @@
 import { defineConfig } from "oxlint";
 
 export default defineConfig({
-  plugins: ["typescript", "react", "import"],
+  plugins: ["typescript", "import"],
   options: {
     typeAware: true,
   },
@@ -17,10 +17,8 @@ export default defineConfig({
     "import/namespace": "off",
     "no-unsafe-type-assertion": "off",
     "no-unassigned-import": "off",
-    "react-in-jsx-scope": "off",
     "no-shadow": "off",
     "no-underscore-dangle": "off",
-    "iframe-missing-sandbox": "off",
     "consistent-return": "off",
     "no-unnecessary-type-arguments": "off",
     "consistent-type-imports": ["error", { fixStyle: "separate-type-imports" }],
@@ -38,31 +36,9 @@ export default defineConfig({
   },
   overrides: [
     {
-      // React Compiler rules do not model Solid component semantics.
-      files: ["**/*.solid.*"],
-      rules: {
-        "react/exhaustive-effect-dependencies": "off",
-        "react/hooks": "off",
-        "react/immutability": "off",
-        "react/memo-dependencies": "off",
-        "react/purity": "off",
-        "react/refs": "off",
-        "react/set-state-in-effect": "off",
-        "react/use-memo": "off",
-      },
-    },
-    {
       files: ["*.d.ts"],
       rules: {
         "consistent-type-imports": "off",
-      },
-    },
-    {
-      // We have our own `forwardRef` implementation that doesn't need the `ref`
-      // parameter, which leads to false positives.
-      files: ["packages/ariakit-react-components/src/**/*.{ts,tsx}"],
-      rules: {
-        "forward-ref-uses-ref": "off",
       },
     },
     {
@@ -71,6 +47,35 @@ export default defineConfig({
       files: ["app/src/lib/**/*.ts"],
       rules: {
         "no-redundant-type-constituents": "off",
+      },
+    },
+    {
+      files: [
+        "**/*.react.*",
+        "{examples,nextjs,packages/ariakit-react*,packages/ariakit-test,templates/react}/**/*.{js,jsx,ts,tsx}",
+      ],
+      excludeFiles: ["**/*.solid.*"],
+      plugins: ["typescript", "react", "import"],
+      rules: {
+        "react/exhaustive-effect-dependencies": "warn",
+        "react/hooks": "warn",
+        "react/iframe-missing-sandbox": "off",
+        "react/immutability": "error",
+        "react/memo-dependencies": "warn",
+        "react/purity": "error",
+        "react/react-in-jsx-scope": "off",
+        "react/refs": "error",
+        "react/set-state-in-effect": "error",
+        "react/use-memo": "error",
+      },
+    },
+    {
+      // We have our own `forwardRef` implementation that doesn't need the
+      // `ref` parameter, which leads to false positives.
+      files: ["packages/ariakit-react-components/src/**/*.{ts,tsx}"],
+      plugins: ["typescript", "react", "import"],
+      rules: {
+        "react/forward-ref-uses-ref": "off",
       },
     },
   ],
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -62,13 +62,20 @@ export default defineConfig({
       },
     },
     {
-      // These types are generated into the gitignored `app/.astro` directory,
-      // so CI always lints without them and `astro:content` resolves
-      // `CollectionEntry` to `any`. That widens unions for the first rule and
-      // makes a narrowing assertion look unnecessary for the second.
+      // Astro generates these types into the gitignored `app/.astro`, so CI
+      // lints without them and `astro:content` resolves `CollectionEntry` to
+      // `any`, which widens every union built from it.
       files: ["app/src/lib/**/*.ts"],
       rules: {
         "no-redundant-type-constituents": "off",
+      },
+    },
+    {
+      // Same missing types, but here they make a narrowing assertion look
+      // unnecessary. This cannot be a directive because the assertion is
+      // needed once the types exist, so the directive would then be unused.
+      files: ["app/src/lib/reference-tokenizer.test.ts"],
+      rules: {
         "no-unnecessary-type-assertion": "off",
       },
     },
--- a/packages/ariakit-react-components/src/collection/collection-renderer.tsx
+++ b/packages/ariakit-react-components/src/collection/collection-renderer.tsx
@@ -227,7 +227,6 @@ function getItemSize(
     if (itemObject.itemSize) {
       return initialSize + itemObject.itemSize * items.length;
     }
-    // oxlint-disable-next-line no-unnecessary-type-arguments
     const totalSize = items.reduce<number>(
       (sum, item) => sum + getItemSize(item, horizontal),
       initialSize,
@@ -242,7 +241,6 @@ function getItemSize(
   // The nested items run along the cross axis, so the item's extent along the
   // measured axis is the largest child extent rather than the sum.
   if (items?.length && !hasSameOrientation) {
-    // oxlint-disable-next-line no-unnecessary-type-arguments
     const maxSize = items.reduce<number>(
       (max, item) => Math.max(max, getItemSize(item, horizontal)),
       0,
@@ -401,6 +399,7 @@ function useScroller(
   });
   // Keep state synchronization and automatic ancestor detection off the
   // layout path.
+  // oxlint-disable-next-line exhaustive-deps
   useEffect(() => {
     if (scrollElement === undefined) {
       inheritedController?.revalidate();
@@ -427,8 +426,7 @@ function useScroller(
     if (nextScroller === scroller) return;
     setScroller(nextScroller);
   });
-  const activeController = scrollElement === undefined ? undefined : controller;
-  return [scroller, scrollerRef, activeController] as const;
+  return [scroller, scrollerRef, controller] as const;
 }
 
 function getRendererOffset(
@@ -489,7 +487,6 @@ function getItemsEnd<T extends Item>(props: {
   const lastItemData = props.data.get(lastItemId);
   if (lastItemData?.end) return lastItemData.end + props.paddingEnd;
   if (!Array.isArray(props.items)) return defaultEnd;
-  // oxlint-disable-next-line no-unnecessary-type-arguments
   const end = props.items.reduce<number>(
     (sum, item) => sum + getItemSize(item, props.horizontal, false),
     0,
@@ -720,7 +717,7 @@ export function useCollectionRenderer<T extends Item = any>({
       ? inheritedScrollerController
       : scrollElementProp === null
         ? undefined
-        : (ownScrollerController ?? undefined);
+        : ownScrollerController;
   const offsetsRef = useRef({ start: 0, end: 0 });
 
   const processVisibleIndices = useCallback(() => {
@@ -770,6 +767,7 @@ export function useCollectionRenderer<T extends Item = any>({
       if (shallowEqual(prevIndices, indices)) return prevIndices;
       return indices;
     });
+    // oxlint-disable-next-line exhaustive-deps
   }, [
     // oxlint-disable-next-line react/memo-dependencies -- element registration signal
     elementsUpdated,
--- a/packages/ariakit-react-components/src/collection/collection-renderer.tsx
+++ b/packages/ariakit-react-components/src/collection/collection-renderer.tsx
@@ -324,9 +324,15 @@ function resolveNullScroller() {
 }
 
 interface ScrollerController {
+  disable: () => void;
+  invalidate: () => void;
+  revalidate: () => void;
+  setResolve: (resolve: () => Element | null) => void;
+}
+
+interface ScrollerControllerState {
   revalidated: boolean;
   resolve: () => Element | null;
-  revalidate: () => void;
 }
 
 function useScroller(
@@ -337,52 +343,61 @@ function useScroller(
   const [scroller, setScroller] = useState<Element | null>(null);
   const scrollerRef = useRef<Element | null>(null);
   const publishedScrollerRef = useRef<Element | null>(null);
-  const controllerRef = useRef<ScrollerController | null>(null);
   const autoResolved = useRef(false);
   const previousRendererRef = useRef(rendererRef);
-  const resolveScroller = () => {
-    const renderer = rendererRef?.current;
-    if (!renderer) return null;
-    return getScrollElement(renderer, scrollElement);
-  };
-  let controller = controllerRef.current;
-  if (!controller && scrollElement !== undefined) {
-    const nextController: ScrollerController = {
-      revalidated: false,
-      resolve: resolveNullScroller,
+  const controllerStateRef = useRef<ScrollerControllerState>({
+    revalidated: false,
+    resolve: resolveNullScroller,
+  });
+  const controller = useMemo<ScrollerController>(() => {
+    return {
+      disable: () => {
+        controllerStateRef.current = {
+          revalidated: true,
+          resolve: resolveNullScroller,
+        };
+      },
+      invalidate: () => {
+        controllerStateRef.current = {
+          ...controllerStateRef.current,
+          revalidated: false,
+        };
+      },
       revalidate: () => {
-        if (nextController.revalidated) return;
-        nextController.revalidated = true;
-        const nextScroller = nextController.resolve();
+        const { revalidated, resolve } = controllerStateRef.current;
+        if (revalidated) return;
+        controllerStateRef.current = { revalidated: true, resolve };
+        const nextScroller = resolve();
         scrollerRef.current = nextScroller;
         if (nextScroller === publishedScrollerRef.current) return;
         publishedScrollerRef.current = nextScroller;
         setScroller(nextScroller);
       },
+      setResolve: (nextResolve) => {
+        controllerStateRef.current = {
+          revalidated: false,
+          resolve: nextResolve,
+        };
+      },
     };
-    controllerRef.current = nextController;
-    controller = nextController;
-  }
+  }, []);
+  const resolveScroller = () => {
+    const renderer = rendererRef?.current;
+    if (!renderer) return null;
+    return getScrollElement(renderer, scrollElement);
+  };
   // Explicit refs and resolvers can change without their prop identity
   // changing, so resolve them during each committed layout.
   // oxlint-disable-next-line exhaustive-deps
   useSafeLayoutEffect(() => {
-    if (inheritedController) {
-      inheritedController.revalidated = false;
-    }
+    inheritedController?.invalidate();
     if (scrollElement === undefined) {
-      if (controller) {
-        controller.revalidated = true;
-        controller.resolve = resolveNullScroller;
-      }
+      controller.disable();
       publishedScrollerRef.current = null;
-      controllerRef.current = null;
       return;
     }
-    if (!controller) return;
     publishedScrollerRef.current = scroller;
-    controller.revalidated = false;
-    controller.resolve = resolveScroller;
+    controller.setResolve(resolveScroller);
     scrollerRef.current = resolveScroller();
   });
   // Keep state synchronization and automatic ancestor detection off the
@@ -407,14 +422,15 @@ function useScroller(
       autoResolved.current = false;
       // Ancestor refs attach after descendant layout effects, so resolve the
       // explicit target again once the entire layout phase has completed.
-      controller?.revalidate();
+      controller.revalidate();
       return;
     }
     const nextScroller = scrollerRef.current;
     if (nextScroller === scroller) return;
     setScroller(nextScroller);
   });
-  return [scroller, scrollerRef, controller] as const;
+  const activeController = scrollElement === undefined ? undefined : controller;
+  return [scroller, scrollerRef, activeController] as const;
 }
 
 function getRendererOffset(
@@ -923,6 +939,8 @@ export function useCollectionRenderer<T extends Item = any>({
         elementObserver?.unobserve(prevElement);
       }
       updateElements();
+      // Item refs are an imperative DOM registry coordinated by updateElements.
+      // oxlint-disable-next-line react/immutability
       elements.set(element.id, element);
       elementObserver?.observe(element);
     },
@@ -987,6 +1005,8 @@ export function useCollectionRenderer<T extends Item = any>({
     for (const [id, element] of elements) {
       if (renderedIds.has(id)) continue;
       elementObserver?.unobserve(element);
+      // Item cleanup mutates the same imperative DOM registry after commit.
+      // oxlint-disable-next-line react/immutability
       elements.delete(id);
     }
   }, [itemsProps, itemSize, elements, elementObserver]);
--- a/packages/ariakit-react-components/src/collection/collection-renderer.tsx
+++ b/packages/ariakit-react-components/src/collection/collection-renderer.tsx
@@ -388,7 +388,6 @@ function useScroller(
   };
   // Explicit refs and resolvers can change without their prop identity
   // changing, so resolve them during each committed layout.
-  // oxlint-disable-next-line exhaustive-deps
   useSafeLayoutEffect(() => {
     inheritedController?.invalidate();
     if (scrollElement === undefined) {
@@ -402,7 +401,6 @@ function useScroller(
   });
   // Keep state synchronization and automatic ancestor detection off the
   // layout path.
-  // oxlint-disable-next-line exhaustive-deps
   useEffect(() => {
     if (scrollElement === undefined) {
       inheritedController?.revalidate();
@@ -772,7 +770,6 @@ export function useCollectionRenderer<T extends Item = any>({
       if (shallowEqual(prevIndices, indices)) return prevIndices;
       return indices;
     });
-    // oxlint-disable-next-line exhaustive-deps
   }, [
     // oxlint-disable-next-line react/memo-dependencies -- element registration signal
     elementsUpdated,
--- a/packages/ariakit-react-store/src/index.tsx
+++ b/packages/ariakit-react-store/src/index.tsx
@@ -70,7 +70,6 @@ function hasSameStoreKeys(
   if (keys?.length !== otherKeys.length) return false;
   for (let index = 0; index < keys.length; index += 1) {
     const key = keys[index];
-    if (key === undefined) return false;
     if (isSameValue(key, otherKeys[index])) continue;
     return false;
   }
--- a/packages/ariakit-react-store/src/index.tsx
+++ b/packages/ariakit-react-store/src/index.tsx
@@ -80,20 +80,23 @@ function hasSameStoreKeys(
 function useStableStoreKeys<K extends StoreKey>(
   keys: readonly K[] | null,
 ): K[] | null {
-  const keysRef = React.useRef<K[] | null>(null);
-  const currentKeys = keysRef.current;
+  const [stableKeys, setStableKeys] = React.useState<K[] | null>(() =>
+    keys === null ? null : [...keys],
+  );
 
   if (keys === null) {
-    keysRef.current = null;
+    if (stableKeys !== null) {
+      setStableKeys(null);
+    }
     return null;
   }
 
-  if (hasSameStoreKeys(currentKeys, keys)) {
-    return currentKeys;
+  if (hasSameStoreKeys(stableKeys, keys)) {
+    return stableKeys;
   }
 
   const nextKeys = [...keys];
-  keysRef.current = nextKeys;
+  setStableKeys(nextKeys);
   return nextKeys;
 }
 
--- a/app/src/lib/reference-tokenizer.test.ts
+++ b/app/src/lib/reference-tokenizer.test.ts
@@ -88,10 +88,7 @@ function createRefData(
 }
 
 function makeRef(id: string, data: RefData): CollectionEntry<"references"> {
-  // Oxlint reports this assertion inconsistently depending on the file set.
-  // https://github.com/oxc-project/oxc/issues/21752
-  // https://github.com/oxc-project/tsgolint/issues/1047
-  // oxlint-disable-next-line typescript/no-unnecessary-type-assertion -- Partial fixture.
+  // Partial fixture: only the members the tokenizer reads are populated.
   return { id, data } as unknown as CollectionEntry<"references">;
 }
 
--- a/app/src/lib/styles.ts
+++ b/app/src/lib/styles.ts
@@ -26,8 +26,10 @@ import {
   toRegexFromWildcard,
 } from "./styles-shared.ts";
 
-// The JSON import keeps its generated literal type, but the resolver needs the
-// public schema shape shared with the build script.
+// `tsconfig.node.json` reaches this file through `vitest.config.ts`, without
+// `app/env.d.ts`, so the JSON import keeps its inferred literal type there and
+// is not assignable to `StylesJson`. The app projects type it as `StylesJson`.
+// oxlint-disable-next-line no-unnecessary-type-assertion
 const styles = stylesRaw as unknown as StylesJson;
 
 export type {
--- a/app/src/lib/styles.ts
+++ b/app/src/lib/styles.ts
@@ -28,7 +28,6 @@ import {
 
 // The JSON import keeps its generated literal type, but the resolver needs the
 // public schema shape shared with the build script.
-// oxlint-disable-next-line no-unnecessary-type-assertion
 const styles = stylesRaw as unknown as StylesJson;
 
 export type {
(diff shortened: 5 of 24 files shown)


oxlint.config.ts
```diff
@@ -41,6 +37,20 @@ export default defineConfig({
...
+        "react/exhaustive-effect-dependencies": "off",
+        "react/hooks": "off",
+        "react/immutability": "off",
+        "react/memo-dependencies": "off",
+        "react/purity": "off",
+        "react/refs": "off",
+        "react/set-state-in-effect": "off",
+        "react/use-memo": "off",
+      },
+    },
```
diegohaz, line 55: Instead of this, can we enable the `react` plugin for React files specifically (`.react.tsx`, react packages, etc. without hard-coding file names)?

packages/ariakit-react-store/src/index.tsx
```diff
@@ -80,20 +80,23 @@ function hasSameStoreKeys(
...
 
-  if (hasSameStoreKeys(currentKeys, keys)) {
-    return currentKeys;
+  if (hasSameStoreKeys(stableKeys, keys)) {
+    return stableKeys;
   }
 
   const nextKeys = [...keys];
-  keysRef.current = nextKeys;
+  setStableKeys(nextKeys);
```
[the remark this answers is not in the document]
diegohaz: Fixed in [`9610d366a`](https://github.com/ariakit/ariakit/commit/9610d366a31bc1c827b9b9765b62a08e7f3f7456): made subscription-key equality reflexive for sparse or `undefined` entries. The same commit adds focused happy-dom and Chrome hydration coverage for both keyed hooks.

oxlint.config.ts
diegohaz: Fixed in [`174c6c737`](https://github.com/ariakit/ariakit/commit/174c6c7372ab8f1bf8373a8a8a97e0cea4280b32): moved the React plugin out of the root config and scoped the eight #7240 rules to `*.react.*` files and semantic React workspaces, with Solid files excluded. I also removed directives made stale by the narrower scope.

diegohaz:
**Fixed**

- [`9610d366a`](https://github.com/ariakit/ariakit/commit/9610d366a31bc1c827b9b9765b62a08e7f3f7456) ([reply](https://github.com/ariakit/ariakit/pull/7248#discussion_r3836133987)): Made sparse subscription-key equality reflexive and added public-hook regression coverage.
- [`174c6c737`](https://github.com/ariakit/ariakit/commit/174c6c7372ab8f1bf8373a8a8a97e0cea4280b32) ([reply](https://github.com/ariakit/ariakit/pull/7248#discussion_r3836134371)): Scoped the React plugin and all eight #7240 rules to React files and removed stale directives.

oxlint.config.ts
```diff
@@ -63,5 +49,34 @@ export default defineConfig({
         "no-redundant-type-constituents": "off",
       },
     },
+    {
+      files: [
+        "**/*.react.*",
+        "{examples,nextjs,packages/ariakit-react*,packages/ariakit-test,templates/react}/**/*.{js,jsx,ts,tsx}",
+      ],
+      excludeFiles: ["**/*.solid.*"],
```
diegohaz: Why can't we simply use `*.react.*` and React packages on files instead of excluding solid here? The problem is that we will have to maintain this `excludeFiles` when we add more frameworks.

oxlint.config.ts
```diff
@@ -63,5 +49,34 @@ export default defineConfig({
...
+        "react/exhaustive-effect-dependencies": "warn",
+        "react/hooks": "warn",
+        "react/iframe-missing-sandbox": "off",
+        "react/immutability": "error",
+        "react/memo-dependencies": "warn",
+        "react/purity": "error",
+        "react/react-in-jsx-scope": "off",
+        "react/refs": "error",
+        "react/set-state-in-effect": "error",
+        "react/use-memo": "error",
```
diegohaz: Aren't these `error` and `warn` done automatically by the `react` plugin?

oxlint.config.ts
```diff
@@ -1,7 +1,7 @@
 import { defineConfig } from "oxlint";
 
 export default defineConfig({
-  plugins: ["typescript", "react", "import"],
+  plugins: ["typescript", "import"],
```
diegohaz: Taking `react` out of the root `plugins` list does more than scope the eight rules tracked by #7240. It also silently disables every other categorized React rule across the whole repository, because oxlint 1.79 does not apply top-level `categories` to a plugin that only an override introduces. Only the ten rule names listed in the override below are still evaluated, and only inside that override's scope.

Three rules that were actively enforcing on this codebase are now off everywhere:

- `react/exhaustive-deps` (rendered as `react-hooks(exhaustive-deps)`), previously `error`. The ten `oxlint-disable-next-line exhaustive-deps` directives this PR removes were suppressing live errors, not stale ones.
- `react/jsx-key`, previously `error`. A missing `key` prop in a list now passes `pnpm lint`.
- `react/no-children-prop`, previously `error`, which is why the directive in `vitest.setup.framework.ts` could be dropped.

Evidence, all run on this PR head with `oxlint 1.79.0`:

1. Running the base `oxlint.config.ts` from [`65b244f3e`](https://github.com/ariakit/ariakit/commit/65b244f3e11d648c7554f436e6ef3996d9646da7) against this branch's code reports 11 errors: 10 `react-hooks(exhaustive-deps)` across `use-root-dialog.ts`, `form.tsx`, `hovercard.tsx`, `hooks.ts`, and `collection-renderer.tsx`, plus 1 `react(no-children-prop)` in `vitest.setup.framework.ts`. The current config reports none of them.
2. Adding a single line, `"react/exhaustive-deps": "error"`, to the override below reproduces exactly those 10 diagnostics, including `form.tsx:103` "React Hook useEffect has a missing dependency: 'shouldFocusOnSubmit'" and `hovercard.tsx:325` "The ref's value `.current` is accessed directly in the effect cleanup function".
3. A throwaway `examples/__probe-lint.tsx` returning `items.map((item) => <li>{item}</li>)` reports `error react(jsx-key)` and exits 1 under the base config, and is silent with exit 0 under this config.

The three `warn` rules cannot cover the gap, because oxlint exits 0 when only warnings are present, so `pnpm lint` never fails on them. I confirmed separately that the chosen severities are otherwise faithful: `react/use-memo` was already `error` and `react/memo-dependencies` and `react/exhaustive-effect-dependencies` were already `warning` under the previous root-plugin config, so the eight tracked rules are preserved correctly. The gap is strictly the categorized rules outside that list.

A targeted fix avoids the version coupling that mirroring all ~87 React rules would create. Adding just the rules that were actually enforcing keeps the list short and intentional:

```ts
rules: {
  "react/exhaustive-deps": "error",
  "react/jsx-key": "error",
  // the eight rules tracked by #7240 stay as they are
},
```

I verified both ids resolve inside this override, and that `react/exhaustive-deps` is the id the configuration schema uses even though diagnostics print as `react-hooks(exhaustive-deps)`. Re-enabling `react/exhaustive-deps` requires restoring the ten directives this PR removes, so if dropping the rule is the intended trade-off instead, that decision is worth stating in the PR description, since the description currently presents those directives as stale rather than as suppressions of a rule being retired.

Separately, `vitest.setup.framework.ts` sits at the repository root and matches neither override pattern, so no React rule reaches it any more even if rules are added back to the override.

**Assessment**

- Severity: Medium (estimate, 85% confidence), because this is enforcement loss rather than a runtime defect, but it removes the guard for a broad class of stale-closure and list-reconciliation bugs in the published React packages.
- Priority: P1 (estimate, 85% confidence), because this PR is specifically about React lint rule adoption, so retiring `react/exhaustive-deps` and `react/jsx-key` should be a stated decision here rather than a side effect.
- Breakage likelihood: Likely (verified, 95% confidence), because ten existing sites already violate `react/exhaustive-deps` and any new violation of it or of `react/jsx-key` now goes unreported.
- Fix effort: Easy (estimate, 80% confidence), because two rule entries restore coverage, though re-enabling `react/exhaustive-deps` also means restoring the ten removed directives.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
```diff
@@ -407,14 +420,15 @@ function useScroller(
...
+      controller.revalidate();
       return;
     }
     const nextScroller = scrollerRef.current;
     if (nextScroller === scroller) return;
     setScroller(nextScroller);
   });
-  return [scroller, scrollerRef, controller] as const;
+  const activeController = scrollElement === undefined ? undefined : controller;
+  return [scroller, scrollerRef, activeController] as const;
```
diegohaz: `activeController` can never be `undefined` at the only call site, so this ternary and the `?? undefined` it feeds on line 723 are both dead.

`useScroller` is called once, on line 701, with `scrollElementProp` passed as `scrollElement`. The returned controller is consumed only on line 718:

```ts
const scrollerController =
  scrollElementProp === undefined
    ? inheritedScrollerController
    : scrollElementProp === null
      ? undefined
      : (ownScrollerController ?? undefined);
```

`ownScrollerController` is read only in the branch where `scrollElementProp` is neither `undefined` nor `null`. In that branch `scrollElement === undefined` is false inside `useScroller`, so `activeController` is always `controller`, and `?? undefined` can never fire. The `?? undefined` was load-bearing before this PR, when the third tuple element was `ScrollerController | null` from `controllerRef.current`, but the `useMemo` controller is never null.

```suggestion
  return [scroller, scrollerRef, controller] as const;
```

Line 723 then simplifies to `: ownScrollerController;`.

I applied both edits locally and ran `pnpm lint`, `pnpm tsc`, and the full `pnpm test` (268 files, 1918 tests). All pass, so the removal is behavior preserving.

**Assessment**

- Severity: Low (verified, 90% confidence), because no behavior depends on either expression.
- Priority: P2 (estimate, 85% confidence), because it is dead code introduced by this PR rather than a defect.
- Breakage likelihood: Rare (verified, 90% confidence), because the removal is provably unreachable code and the full suite passes without it.
- Fix effort: Easy (verified, 95% confidence), because it is a two line change validated locally.

oxlint.config.ts
diegohaz: Validated against this head. The maintenance concern is real and reproducible, but the literal alternative would lose coverage, so I looked for a formulation that keeps both.

Why `*.react.*` plus React packages alone is not enough: `examples` is a mixed root where React is the un-infixed default. It holds 93 non-infixed `.tsx` React modules alongside 77 `.react.tsx` files, and two of the non-infixed ones get suppressions in this very PR, `examples/menu-slide/menu.tsx` and `examples/tab-panel-animated/animated-tabs.tsx`. I removed the `react/refs` suppression from `examples/tab-panel-animated/animated-tabs.tsx` and confirmed it reports `error react(refs)` with the current `files`, then dropped `examples` from the roots and confirmed the same file goes silent with exit 0. So the `examples` root is load-bearing.

Two things worth knowing about the current `excludeFiles`: it is inert today, since the only in-scope Solid file is `examples/separator/index.solid.tsx`, a six line component with no hooks or refs, and removing the exclusion leaves the full repository lint clean. Your maintenance worry is nonetheless demonstrable rather than theoretical: I added a throwaway `examples/separator/index.svelte.tsx` that reads a ref during render, and the current config reports `error react(refs)` on it, which is exactly the false positive a new framework would hit.

Here is a formulation I validated that needs no update when a framework is added. It splits the override so the mixed root excludes every framework-infixed file instead of naming frameworks, and lets `**/*.react.*` in the first override opt the React ones back in:

```ts
const reactRules = {
  "react/exhaustive-effect-dependencies": "warn",
  "react/hooks": "warn",
  "react/iframe-missing-sandbox": "off",
  "react/immutability": "error",
  "react/memo-dependencies": "warn",
  "react/purity": "error",
  "react/react-in-jsx-scope": "off",
  "react/refs": "error",
  "react/set-state-in-effect": "error",
  "react/use-memo": "error",
} as const;

// ...

{
  files: [
    "**/*.react.*",
    "{nextjs,packages/ariakit-react*,packages/ariakit-test,templates/react}/**/*.{js,jsx,ts,tsx}",
  ],
  plugins: ["typescript", "react", "import"],
  rules: reactRules,
},
{
  // React modules here have no framework infix, and every other framework
  // always has one, so excluding infixed files scopes this correctly without
  // naming each framework.
  files: ["examples/**/*.{js,jsx,ts,tsx}"],
  excludeFiles: ["**/*.*.{js,jsx,ts,tsx}"],
  plugins: ["typescript", "react", "import"],
  rules: reactRules,
},
```

With that config I confirmed on this head that the full repository `oxlint` run is clean and `pnpm tsc` passes, that `examples/tab-panel-animated/animated-tabs.tsx` still reports `react/refs` once its suppression is removed, that a copy of the probe named `examples/separator/probe.react.tsx` still reports `react/refs` through the first override, and that the throwaway `.svelte.tsx` probe is no longer reported. Files matching two overrides do get both rule sets, which I checked separately by removing the `react/immutability` suppressions in `collection-renderer.tsx` and seeing both errors return despite the `forward-ref-uses-ref` override also matching.

Keeping the current single override with `excludeFiles` is also defensible, since the entry costs one glob per framework. The split above is the version that answers your objection directly.

oxlint.config.ts
diegohaz: No, and I checked this directly rather than trusting the release notes. In oxlint 1.79 a plugin that only an override introduces does not receive top-level `categories` activation, so these explicit severities are required.

Three runs with the same fixture, a `createElement` call passing `children` in the props object, and the same `categories` block this config uses:

- `react` in the top-level `plugins`, no `rules` entry: reports `error react(no-children-prop)`.
- `react` only in an override's `plugins`, no `rules` entry: reports nothing, exit 0.
- `react` in an override's `plugins` plus `"react/no-children-prop": "error"`: reports `error react(no-children-prop)`.

So the middle case is what this config does for every React rule that is not named here, which confirms the reasoning recorded in the PR description.

I also checked that the severities you chose match what the rules already had, so nothing is silently downgraded. Under the previous root-plugin config, `react/use-memo` reported as `error` and both `react/memo-dependencies` and `react/exhaustive-effect-dependencies` reported as `warning`, which is exactly what this list encodes.

The consequence of that same mechanic is the one thing I do think needs a decision here, and I left it as a separate comment on the root `plugins` line: `react/exhaustive-deps`, `react/jsx-key`, and `react/no-children-prop` were all enforcing at `error` before this PR and are now off repository wide.

diegohaz (commented):
Initial review of all 21 changed files, diffed from the merge base [`65b244f3e`](https://github.com/ariakit/ariakit/commit/65b244f3e11d648c7554f436e6ef3996d9646da7) to head [`174c6c737`](https://github.com/ariakit/ariakit/commit/174c6c7372ab8f1bf8373a8a8a97e0cea4280b32), then expanded into the surrounding code: the full `useScroller` and `useCollectionRenderer` flow, both keyed store hooks and their consumers, and the whole `oxlint.config.ts` scoping model. No prior review carries this skill's marker, so there was no incremental baseline.

## Verification

- `pnpm tsc`: clean.
- `pnpm test`: 268 files, 1918 tests passed.
- `oxlint --disable-nested-config` over the repository at head: clean, exit 0.
- Regression proof for the sparse-key fix: restoring `if (key === undefined) return false;` in `hasSameStoreKeys` fails all 6 tests in `app/src/sandbox/store-keyed-subscriptions/test.ts` with "Too many re-renders. React limits the number of renders to prevent an infinite loop.", and they pass with the fix. This independently confirms the claim in the PR description, so the coverage added in [`9610d366a`](https://github.com/ariakit/ariakit/commit/9610d366a31bc1c827b9b9765b62a08e7f3f7456) is proven, not just present.
- Lint scoping experiments: a three-way plugin and category check, the base config run against head code, re-enabling `react/exhaustive-deps` in the override, a `jsx-key` probe, a `.svelte.tsx` probe under `examples`, and a check that a file matching two overrides receives both rule sets.
- Both proposed changes were applied locally and validated before being suggested, then reverted. The worktree is back at head with no leftovers.

## Findings

Two published inline, both first discovered in this review.

1. `oxlint.config.ts` line 4: moving `react` out of the root `plugins` list silently disables every categorized React rule that is not named in the override. `react/exhaustive-deps`, `react/jsx-key`, and `react/no-children-prop` were all enforcing at `error` before this PR and now report nothing anywhere. The ten `exhaustive-deps` directives this PR removes were suppressing live errors rather than being stale.
2. `packages/ariakit-react-components/src/collection/collection-renderer.tsx` lines 430 and 431: `activeController` can never be `undefined` at the only call site, so that ternary and the `?? undefined` on line 723 are both dead.

- Thread on `oxlint.config.ts` line 57, `excludeFiles` maintenance ([`#discussion_r3836191539`](https://github.com/ariakit/ariakit/pull/7248#discussion_r3836191539)): confirmed as a real concern, and reproduced with a throwaway `.svelte.tsx` probe that the current config flags. The literal alternative is not viable, since dropping the `examples` root silences a real `react/refs` error in `examples/tab-panel-animated/animated-tabs.tsx`, one of 93 non-infixed React modules there. Replied with a validated split-override formulation that needs no per-framework maintenance. Disposition: Address now if the maintenance property is wanted, otherwise the current line is correct and merely inert today.
- Thread on `oxlint.config.ts` line 69, automatic severities ([`#discussion_r3836196355`](https://github.com/ariakit/ariakit/pull/7248#discussion_r3836196355)): the premise does not hold in oxlint 1.79, so the explicit severities are required. Replied with the three-way experiment. I also confirmed the chosen severities match what each rule had under the previous root-plugin config, so nothing is silently downgraded. Disposition: Invalid or obsolete as a change request, and the PR description's reasoning is accurate.

The two resolved threads were not re-adjudicated.

- Missing changeset: agreed, none is needed. The only published-package behavior difference is subscription-key equality for sparse or `undefined` entries, which is outside the documented `PropertyKey[]` contract. Under the released code such input caused a new keys array every render and therefore a re-subscribe every render, and it now stays stable, which is strictly better on invalid input. Disposition: Declined non-goal, consistent with the PR's own cycle 3 reassessment.
- The state-based `useStableStoreKeys` adds one extra render pass whenever the key list actually changes, where the ref version did not. It converges, it is the documented React pattern for adjusting state during render, and the full suite is green, so this is not a finding.
- The `useCallback` wrapper in `combobox-select-open-page-scroll` is a correct and minimal workaround. Passing `combobox.setAnchorElement` straight to `ref` makes the compiler treat the whole `combobox` value as a ref, which then flags `store={combobox}` and `onClick={combobox.show}` as well. That is the only site in the repository using a store setter directly as a `ref`.
- The `useScroller` refactor is behavior preserving as far as I could determine by reading. The controller state transitions map one to one onto the previous mutations, and the one timing difference, detaching the controller during render instead of one render later, is unobservable because the caller only reads `ownScrollerController` when `scrollElementProp` is neither `undefined` nor `null`.

## Not fully validated

I did not run the Playwright or Chrome suites in this review, so the scroller coordination in `collection-renderer.tsx` is covered here only by reading plus the happy-dom suite. That path is primarily exercised by the browser projects the PR description reports as green.

oxlint.config.ts
diegohaz: Fixed in [`0ccf978cf`](https://github.com/ariakit/ariakit/commit/0ccf978cf), by moving `react` back into the top-level `plugins` list rather than by adding rules to the override.

Two corrections to this comment, both from measurement after it was written.

The scale was wrong. I wrote "every other categorized React rule" and implied dozens. `oxlint --print-config` gives the exact figures: 38 enabled `react/*` rules before this series, 0 at the previous head, 40 now. So 38 rules were lost, not the larger number the wording suggested.

The count of live suppressions was wrong. I wrote that ten directives were suppressing live errors. The previous head removed twelve `exhaustive-deps` directives, of which nine were live and three were genuinely dead. All three dead ones sat on `useSafeLayoutEffect` calls, which `exhaustive-deps` does not analyze. I found that out by restoring all twelve and letting `reportUnusedDisableDirectives` reject the three that suppressed nothing, which is a better check than my transcription.

The scoping mechanism itself is confirmed. With `react` declared only inside an override naming one rule, a probe file reported that rule and stayed silent on `react/jsx-key` and `react/no-children-prop`; with the plugin at the top level all three report.

The fix keeps every React rule on everywhere and scopes a single rule, `react/immutability`, away from Solid, that being the only one measured misfiring on Solid source anywhere in the repository. The remaining gap this comment mentioned, that plain React rules still run on Solid, is tracked in #7250, and the `useSafeLayoutEffect` blind spot is tracked in #7253.

packages/ariakit-react-components/src/collection/collection-renderer.tsx
diegohaz: Applied in [`0ccf978cf`](https://github.com/ariakit/ariakit/commit/0ccf978cf). `useScroller` now returns `controller` directly and line 720 reads `: ownScrollerController;`.

Two independent reviews re-derived the reachability argument rather than taking the tests as proof: `ownScrollerController` has exactly one consumer, and it is reached only when `scrollElementProp` is neither `undefined` nor `null`, which is precisely the branch where the removed ternary evaluated to `controller`. One of them also noted the result matches what `useScroller` returned on `main` before this pull request's refactor, so this restores that shape rather than inventing a new one.

Verified with the full suite, `pnpm tsc`, and lint.

oxlint.config.ts
diegohaz: Resolved differently than proposed. [`0ccf978cf`](https://github.com/ariakit/ariakit/commit/0ccf978cf) removes this override entirely, so there is no `excludeFiles` left to maintain.

The maintenance concern you raised was correct, and the split-override formulation I suggested here was the wrong answer to it. Keeping `react` at the top level means every React rule applies everywhere by default, so no allowlist of workspaces exists at all. That also closed two holes the allowlist had: `guide/` was never listed despite `guide/package.json` declaring React, and repository-root files matched nothing.

What remains is a single override disabling one rule, `react/immutability`, for `**/*.solid.*` and `packages/ariakit-solid*/**`. That second glob matters and the earlier shape lacked it: the four Solid packages carry no `.solid.` infix, so `**/*.solid.*` alone never reached them.

On your original question, whether a new framework forces a config edit: it does, but only if that framework trips a React rule, and the failure is a visible error on the commit that introduces it rather than silent coverage loss. The eight-rule pre-emptive list that was here before has been dropped, because measurement showed it was the previous override's allowlist inverted and six of its eight rules had been enforcing on Solid on `main` all along.

oxlint.config.ts
diegohaz: Obsolete as of [`0ccf978cf`](https://github.com/ariakit/ariakit/commit/0ccf978cf). These explicit severities are gone, along with the override that held them.

The answer to the question stands: a plugin introduced only by an override does not receive top-level `categories` activation, so listing severities was required in that shape. The shape itself was the problem. With `react` declared at the top level the rules take their category severities exactly as before, which measurement confirms: the resolved severity delta against the config before this series is now exactly two entries, `react/immutability` and `react/refs` moving from off to on, and nothing else moved.

oxlint.config.ts
```diff
@@ -56,11 +62,24 @@ export default defineConfig({
...
-      // being built first, and linting may run before the app is built.
+      // These types are generated into the gitignored `app/.astro` directory,
+      // so CI always lints without them and `astro:content` resolves
+      // `CollectionEntry` to `any`. That widens unions for the first rule and
+      // makes a narrowing assertion look unnecessary for the second.
       files: ["app/src/lib/**/*.ts"],
       rules: {
         "no-redundant-type-constituents": "off",
+        "no-unnecessary-type-assertion": "off",
+      },
```
diegohaz: This comment names one mechanism, but the disable it documents covers two sites with different causes, and only one of them is the `app/.astro` case.

Measured on this head by temporarily re-arming the rule and linting `app/src/lib` in both states:

| `app/.astro` | reports |
| --- | --- |
| absent (what CI sees) | `styles.ts:31` and `reference-tokenizer.test.ts:92` |
| present (after `astro sync`) | `styles.ts:31` only |

`reference-tokenizer.test.ts:92` behaves exactly as this comment describes. `app/src/lib/styles.ts:31` does not: it reports identically in both states, so no generated type is involved, and it is a true positive.

`app/env.d.ts` already declares that JSON module as the shared schema type:

```ts
declare module "#app/styles/styles.json" {
  const styles: import("./src/lib/styles-json-types.ts").StylesJson;
  export default styles;
}
```

So `stylesRaw` is already `StylesJson`, and `as unknown as StylesJson` converts nothing. I confirmed the type instead of inferring it: adding `const __probe: number = stylesRaw;` to that file fails with `Type 'StylesJson' is not assignable to type 'number'`, which also rules out the import resolving to `any`. That makes the comment above `styles.ts:31`, "The JSON import keeps its generated literal type, but the resolver needs the public schema shape shared with the build script", a description of a conversion that is not happening.

To be clear, the directory-wide `off` is the right shape for the `CollectionEntry` case, and I checked the narrower alternative before suggesting anything. Restoring a line-level directive on the test fails with `Unused oxlint-disable directive (no problems were reported)` as soon as `app/.astro` exists, which is precisely what the new `reportUnusedDisableDirectives: "error"` is for. That half is well chosen.

The gap is the other half. At the merge base, `styles.ts` carried its own directive on the line above the assertion, so the redundancy was visible there. It is now silenced by a directory-wide rule whose recorded reason does not apply to it, and every future unnecessary assertion anywhere under `app/src/lib` inherits that silence.

Removing the redundant assertion makes the override cover only the case this comment describes. In `app/src/lib/styles.ts`, replace lines 29 to 31 with:

```ts
const styles: StylesJson = stylesRaw;
```

The two comment lines go with the assertion, since they describe the conversion being removed. This comment then needs no change: `CollectionEntry` becomes the only case it covers.

I applied that locally and re-ran with the rule armed: `app/src/lib` is completely clean with `app/.astro` present, and the only remaining report without it is `reference-tokenizer.test.ts:92`, so the override matches its stated reason exactly. `tsc -p app/tsconfig.workerd.json` reports the same 6 pre-existing errors before and after, none of them in `styles.ts`.

**Assessment**

- Severity: Low (verified, 90% confidence), because no runtime behavior changes; the cost is a silenced true positive plus a rationale that does not match what is silenced.
- Priority: P2 (estimate, 80% confidence), because it is a coverage and accuracy gap this commit introduces rather than a defect.
- Breakage likelihood: Possible (estimate, 75% confidence), because hiding something new requires someone to add a redundant assertion under `app/src/lib`. The one already hidden is verified rather than hypothetical.
- Fix effort: Easy (verified, 95% confidence), because it is a one line change plus a stale comment, validated locally with lint and the app type check.

diegohaz (commented):
Re-review. The previous review with this marker was written against [`174c6c737`](https://github.com/ariakit/ariakit/commit/174c6c7372ab8f1bf8373a8a8a97e0cea4280b32), so the incremental baseline is that SHA and the target is [`0ccf978cf`](https://github.com/ariakit/ariakit/commit/0ccf978cfed2975f8ec6ef4c64c5f330332aca23), 10 files. I also re-derived the parts of the earlier range that the new plugin-scoping model touches, since moving `react` back to the root arms more rules on more paths than the revision I reviewed before.

Reviewed in a dedicated worktree checked out at the head SHA, with a clean tree restored after every experiment.

## Verification

- `oxlint --disable-nested-config`: clean, exit 0. Run 15 consecutive times with identical results, so nothing in the current tree reports non-deterministically.
- `oxfmt --check`: clean, 2122 files.
- `pnpm test --run`: 268 files, 1918 tests passed, matching the description.
- `tsc -p app/tsconfig.workerd.json`: 6 pre-existing errors from missing `wrangler types` output, none in the files this PR touches. Used as a stable baseline for the `styles.ts` experiments.
- Local experiments, each reverted: re-arming `no-unnecessary-type-assertion` with and without `app/.astro`; restoring a line-level directive on `reference-tokenizer.test.ts`; replacing the `styles.ts` assertion; a deliberately wrong-typed probe against `stylesRaw`; removing the Solid override; and stripping all nine `exhaustive-deps` suppressions to catalogue what each one hides.

## Confirmed claims

Each of these is a load-bearing claim in the description that I measured rather than took on trust.

- The central claim holds for the top-level rule set. Diffing `oxlint --print-config` between the config at the merge base and this head, the only changes in the top-level `rules` block are `react/immutability` and `react/refs` moving from `allow` to `deny`, which is exactly #7240's goal. 40 `react/*` rules enabled (31 deny, 9 warn), matching the table.
- The Solid override is narrowed correctly. With that override removed, the only React diagnostics on Solid source anywhere in the repository are three `react/immutability` reports in `app/src/sandbox/create-ref-6349/index.solid.tsx`, exactly as stated. Every Solid source file in the repo carries the `.solid.` infix or lives under `packages/ariakit-solid*`, so the two globs have no path gap today.
- The directive census matches. The merge base had 12 `exhaustive-deps` directives and this head has 9. All three removed ones sat on a `useSafeLayoutEffect` call, as claimed: `collection-renderer.tsx`, `use-root-dialog.ts`, and `useUpdateLayoutEffect` in `hooks.ts`. The surviving nine are at the same sites as the merge base.
- No dead suppressions remain anywhere. `reportUnusedDisableDirectives: "error"` plus a clean full run proves every one of the 72 `oxlint-disable` directives in the repo is currently load-bearing. I also confirmed the unprefixed `exhaustive-deps` name resolves, by removing one directive and watching `react-hooks(exhaustive-deps)` report at that site.
- The two remaining suppressions that hide a "missing dependency" report are intentional, not latent bugs. `useWrapElement` omits `callback` because callers supply the dependency list, which its own `-- public deps API` reason states, and the `form.tsx` effect omits `shouldFocusOnSubmit` deliberately so clearing the flag does not re-run the effect, which the comment block above the preceding effect explains. Both are byte-identical to the merge base.
- The dead code removed from `useScroller` really was unreachable. I re-derived it rather than relying on the earlier thread: `ownScrollerController` has exactly two occurrences, and its only read is reachable only when `scrollElementProp` is neither `undefined` nor `null`, which is the branch where the removed ternary always evaluated to `controller`. `controller` is a `useMemo` value and never null, so the `?? undefined` could not fire either.
- `hasSameStoreKeys` is now reflexive for every value the type admits, not just for `undefined`. `isSameValue` is `value === other || (value !== value && other !== other)`, so `NaN` keys, which are valid `PropertyKey` values and would otherwise retrigger the render-phase update forever, compare equal. That closes the loop class the earlier fix opened, rather than only the reported case.

## Findings

One, published inline.

- `oxlint.config.ts` lines 65 to 73: the new `no-unnecessary-type-assertion: "off"` silences two sites with different causes, and the comment's `app/.astro` mechanism explains only one of them. `app/src/lib/styles.ts:31` reports identically with and without the generated types and is a true positive, because `app/env.d.ts` already declares that JSON module as `StylesJson`, so the assertion converts nothing. It carried its own directive at the merge base and is now silenced with no record. I verified the narrower alternative is genuinely unavailable, so the directory-wide `off` is right for the `CollectionEntry` case; the fix is to drop the redundant assertion.

## Description mismatches

Both facts below are disclosed in the Known exceptions section, so nothing is concealed. The problem is that the two summary claims a reader would use to audit the config are contradicted by the config itself, in a PR whose whole premise is that config drift must not pass unnoticed.

`oxlint --print-config` emits an `overrides` array alongside `rules`, and comparing that array across the same two configs shows two further severity changes: `app/src/lib/**/*.ts` gains `typescript/no-unnecessary-type-assertion: "allow"`, and the new fourth override sets `react/immutability: "allow"` for Solid paths. So the verification bullet, "the only rule severity changes are `react/immutability` and `react/refs` moving from off to on", holds for the top-level `rules` block but not for the output it cites, and both omissions are coverage reductions rather than additions.

For the same reason, "The three pre-existing overrides are unchanged" is not accurate: the `app/src/lib` override gained a second rule and a rewritten comment in this commit. The sentence enumerating the change against the pre-series config also omits that addition entirely.

Both are description edits with no code anchor, which is why they are here rather than inline.

- Missing changeset: agreed, none is needed. I re-checked the only published-package behavior difference, subscription-key equality for sparse or `undefined` entries, and it is outside the documented `PropertyKey[]` contract and strictly better on such input than the released code. Disposition: declined non-goal.
- `denyWarnings: true` collapses the distinction between `correctness: "error"` and `suspicious: "warn"` for CI purposes, leaving the tri-level `categories` block meaningful only for editor display. This is the stated intent and the option carries a comment saying so. Disposition: declined non-goal.
- oxc-project/oxc#21752, linked from the comment this PR removes, is still open and reports `no-unnecessary-type-assertion` firing non-deterministically under type-aware linting, with CI failing randomly on identical commits. `reportUnusedDisableDirectives: "error"` makes that failure mode bidirectional, since a flaky rule that does not fire turns its directive into a hard error, and about ten directives here target type-aware rules. I could not reproduce any variation: 15 consecutive full runs were identical. This PR also removes both directives for the specific rule named upstream, which reduces rather than increases the exposure. Not a finding, recorded as a suspected issue only.

No follow-up issues were registered, because no candidate survived as confirmed work that this PR should hand off.

## Threads and coverage

All six existing review threads are resolved, so there were no unresolved candidate findings to validate in this run. Both findings from the previous marked review were fixed in this head and I confirmed each independently rather than relying on the replies.

Not fully validated: I did not run the Playwright or browser suites locally, so the scroller coordination in `collection-renderer.tsx` is covered here by reading plus the happy-dom suite. All 30 CI checks on this head are green, including Type check, Lint, Test, Test React 18, and the Chrome, Firefox, Safari, iOS, and Android projects. I also did not run `pnpm build` or a packed-contents check, since nothing here changes packaging.

oxlint.config.ts
diegohaz: Fixed in [`0469e7762`](https://github.com/ariakit/ariakit/commit/0469e7762c29705864db3e5cc6bfd5cb194a5ff6): scoped `no-unnecessary-type-assertion` to `app/src/lib/reference-tokenizer.test.ts` and restored the directive on `styles.ts`, so the override now covers only the case its comment describes.

The two-site observation was right and the directory-wide shape was mine to fix. At the merge base `styles.ts` carried its own directive, and replacing it with a rule that names a different cause is the kind of quiet coverage loss this pull request exists to remove.

The suggested replacement does not hold, though, and this is worth recording because it is the third review to reach it.

`const styles: StylesJson = stylesRaw;` passes every app project and fails `pnpm tsc`, which is `pnpm -F app run check && tsc -b`:

```
app/src/lib/styles.ts:29:7 - error TS2322: Type '{ version: number; modules: (...)[]; index: {...} }'
  is not assignable to type 'StylesJson'.
```

`tsconfig.node.json` excludes `app`, yet still reaches this file: `vitest.config.ts` imports `app/src/lib/source-plugin.ts`, which imports `./styles.ts`. `app/env.d.ts` is not in that program, so `#app/styles/styles.json` resolves to the real JSON and keeps its inferred literal type. The `as unknown as` form is the only one that satisfies both resolutions, which is why the assertion is load-bearing rather than redundant.

`tsc -p app/tsconfig.workerd.json` passing is consistent with that, and so is the `__probe: number` check: all four app projects that contain the file type it as `StylesJson`. The disagreement is with the root project, not with any of them.

The `styles.ts` comment now names that project and the import path, so the next reader does not have to re-derive it. The assertion has already been deleted and restored once in this pull request's history.

Two follow-ups came out of this. #7254 tracks moving the ambient `#app/styles/styles.json` declaration next to `styles.ts` so every program agrees, which would delete the cast, the directive and the comment, and remove an `as unknown as T` form that `ariakit-general-code-style` disallows. #7255 tracks running `astro sync` in the Lint job so CI and local lint stop disagreeing about the generated types; with those types present, both `app/src/lib` overrides lint clean and can be deleted.

diegohaz:
Handled the one unresolved review thread on `0ccf978cf`.

**Fixed**

- [`0469e7762`](https://github.com/ariakit/ariakit/commit/0469e7762c29705864db3e5cc6bfd5cb194a5ff6) ([reply](https://github.com/ariakit/ariakit/pull/7248#discussion_r3837042777)): the previous commit disabled `no-unnecessary-type-assertion` for all of `app/src/lib` under a reason that explained only one of the two sites it silenced. Scoped it to `app/src/lib/reference-tokenizer.test.ts`, which is the site that genuinely cannot be green in both type states, and restored the line-level directive on `app/src/lib/styles.ts` with its real cause. The rule is armed again everywhere else under that directory.

**Invalid**

- ([reply](https://github.com/ariakit/ariakit/pull/7248#discussion_r3837042777)): the suggested `const styles: StylesJson = stylesRaw;` passes every app project but fails `pnpm tsc` with `TS2322`. `tsconfig.node.json` excludes `app` yet reaches the file through `vitest.config.ts` and `source-plugin.ts`, so `app/env.d.ts` is absent from that program and the JSON import keeps its inferred literal type. The `as unknown as` form is the only one that satisfies both resolutions.

**Tracked**

- [#7254](https://github.com/ariakit/ariakit/issues/7254): move the ambient `#app/styles/styles.json` declaration next to `styles.ts` so every TypeScript program agrees, which would delete the cast, the directive and the comment.
- [#7255](https://github.com/ariakit/ariakit/issues/7255): run `astro sync` in the Lint job so CI and local lint stop disagreeing about the generated Astro types. With those types present, both `app/src/lib` overrides lint clean and can be deleted.

Verified before pushing: lint clean with and without `app/.astro`, `oxfmt --check`, `pnpm tsc`, and the full test suite at 268 files and 1918 tests.

PR #5240, opened 2025-12-30
Add `@ariakit/ui` package

diegohaz:
## Motivation

The docs app styles components with the legacy `ak-*.css` utility layer. Those utilities are global class names with no typed surface, so a component's look is assembled by hand at every call site, variations are expressed by remembering the right class combination, and nothing stops a call site from drifting away from the intended recipe.

## Solution

Add a private `@ariakit/ui` package with [Clava](https://github.com/ariakit/clava) style recipes and Ariakit React components. Variant props provide a typed surface for the shared design system. The package uses `@ariakit/scripts` for build and clean, with the shared stylesheet at `@ariakit/ui/styles/ui.css`.

Keep a small docs app for the later site rewrite: the Separator and Combobox documentation, Combobox Group and its Custom Items variant, guides, sandboxes, previews, and the UI gallery. Remove the old site UI and Clerk integration, and preserve reimplementation requirements in `app/docs/reimplementation`.

```tsx
// Before: the recipe lives in the class attribute
<button class="ak-button ak-button-primary ak-frame-field/2 text-sm">

// After: the recipe lives in the style module
<Button $layer="brand" $size="sm">
```

Declarative tables reserve `key` for row identity, so stateful cells keep their state when rows move. Optional render shorthands omit `false`, `null`, and `undefined`; numeric zero and explicit empty strings remain content. Static Combobox items can request their highlight with `$highlighted`.

```tsx
<Table rows={[{ key: "ada", name: "Ada" }]} />
<ComboboxSelectItem value="Open">{showLabel && "Custom label"}</ComboboxSelectItem>
<div {...comboboxItem.jsx({ $highlighted: true })} />
```

## Preview

<details>
<summary>Stable table rows and conditional content</summary>

The reserved row key keeps Ada’s edited note with Ada after the rows are reversed and a contributor is added:

https://github.com/user-attachments/assets/1ba77bbe-a67e-4c75-a1d1-670a986cc0b6

Unkeyed rows use their position. An explicit zero key remains distinct from those positions, so the assigned task keeps its edited note:

https://github.com/user-attachments/assets/e3051f1d-834e-4603-97a7-275f267cd919

False Select labels fall back to their values. Enabling the custom labels replaces that content:

https://github.com/user-attachments/assets/b0c39125-bba2-4802-81f8-867c011c8a95

Optional navigation headings leave their links reachable. A zero label remains a working disclosure button:

https://github.com/user-attachments/assets/9e5a4812-2dd0-421c-af1e-8c04101310a1

[...]

PR #7258, opened 2026-08-22
Use one type for the styles JSON import

### LINKED ISSUE #7254: styles.json import types differently in the root and app TypeScript projects
## Problem

`app/src/lib/styles.ts` needs a laundering cast and a lint suppression only because two TypeScript programs disagree about the type of one JSON import.

```ts
// oxlint-disable-next-line no-unnecessary-type-assertion
const styles = stylesRaw as unknown as StylesJson;
```

`app/env.d.ts` declares the module as the shared schema type:

```ts
declare module "#app/styles/styles.json" {
  const styles: import("./src/lib/styles-json-types.ts").StylesJson;
  export default styles;
}
```

Every app project sees that declaration and types `stylesRaw` as `StylesJson`, which makes the assertion look unnecessary. The root `tsconfig.node.json` does not: it excludes `app`, yet it still pulls this file in through `vitest.config.ts`, which imports `app/src/lib/source-plugin.ts`, which imports `./styles.ts`. In that program the import resolves to the real JSON and keeps its inferred literal type, which is not assignable to `StylesJson`.

So the assertion is required by `pnpm tsc` and redundant everywhere else. Removing it fails the root `tsc -b` with `TS2322`; keeping it without a suppression fails lint once `options.reportUnusedDisableDirectives` is enabled, which #7248 added.

## Expected outcome

Make every program that includes `styles.ts` see the same type, most likely by moving the `#app/styles/styles.json` declaration out of `app/env.d.ts` into a declaration file next to `styles.ts` that is pulled in with a `/// <reference path>`. The two ambient declarations would collide, so the copy in `app/env.d.ts` has to go at the same time.

That would let the cast, the suppression, and the explanatory comment all be deleted. It also removes an `as unknown as T` form, which `ariakit-general-code-style` disallows for silencing type errors. The cast predates #7248; that pull request only made it visible by enabling the unused-directive check.

Discovered during the pre-commit review of #7248.

diegohaz:
## Motivation

`app/src/lib/styles.ts` is loaded by the app TypeScript projects and by the root Node project through `vitest.config.ts`. Only the app projects loaded the ambient JSON declaration from `app/env.d.ts`, so the same import had two types and required a laundering cast:

```ts
const styles = stylesRaw as unknown as StylesJson;
```

Closes #7254.

## Solution

Move the `#app/styles/styles.json` declaration into `app/src/lib/styles-json.d.ts` and reference it directly from `styles.ts`, so every TypeScript program that includes the source loads the shared `StylesJson` schema:

```ts
/// <reference path="./styles-json.d.ts" />

import styles from "#app/styles/styles.json" with { type: "json" };
```

Remove the duplicate declaration from `app/env.d.ts`, along with the cast, suppression, and workaround comment. A file-specific lint override permits this deliberate path reference only in `styles.ts`.

No changeset is included because this changes internal app type wiring without changing runtime or published-package behavior.

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test app/src/lib/styles.test.ts` (21 tests)
- Root/app regression proof: the cast-free assignment failed the root program with `TS2322` before the declaration move, while both focused programs passed afterward.

### DIFF
--- a/app/env.d.ts
+++ b/app/env.d.ts
@@ -17,11 +17,6 @@ declare module "*?source" {
   export default source;
 }
 
-declare module "#app/styles/styles.json" {
-  const styles: import("./src/lib/styles-json-types.ts").StylesJson;
-  export default styles;
-}
-
 type PlusType = import("./src/lib/schemas.ts").PlusType;
 type Framework = import("./src/lib/schemas.ts").Framework;
 type User = import("@clerk/astro/server").User;
--- a/app/src/lib/styles-json-types.ts
+++ b/app/src/lib/styles-json-types.ts
@@ -8,7 +8,7 @@
  * SPDX-License-Identifier: UNLICENSED
  */
 
-/** JSON shape for `#app/styles/styles.json` (shared by `env.d.ts` and `styles.ts`). */
+/** JSON shape for `#app/styles/styles.json`. */
 
 export type StyleType = "utility" | "variant" | "at-property";
 
--- a/app/src/lib/styles-json.d.ts
+++ b/app/src/lib/styles-json.d.ts
@@ -1,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2025-present Ariakit FZ-LLC. All Rights Reserved.
+ *
+ * This software is proprietary. See the license.md file in the root of this
+ * package for licensing terms.
+ *
+ * SPDX-License-Identifier: UNLICENSED
+ */
+
+declare module "#app/styles/styles.json" {
+  const styles: import("./styles-json-types.ts").StylesJson;
+  export default styles;
+}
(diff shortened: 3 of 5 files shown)


diegohaz:
Closing this in favor of #5240.

That pull request removes the entire generated styles JSON pipeline: `app/src/lib/styles.ts`, `styles-json-types.ts`, `build-styles.ts`, `app/src/styles/styles.json`, the ambient declaration in `app/env.d.ts`, and the `source-plugin.ts` calls that resolve styles. Once it lands, the conflicting JSON import type no longer exists.

This pull request is therefore temporary machinery for a path we plan to delete. The triple-slash reference and file-specific lint exception are not worth carrying for an internal, non-runtime issue. We can keep the existing cast and suppression until #5240 removes them.

Closing #7254 as not planned for the same reason.

PR #7259, opened 2026-08-22
Disable incompatible React lint rules for Solid source

### LINKED ISSUE #7250: React lint rules misfire on Solid source
## Problem

Plain React lint rules run on Solid source and can report defects that do not exist in Solid. This is independent of the React Compiler rules, which [#7248](https://github.com/ariakit/ariakit/pull/7248) scopes away from Solid.

`react/jsx-key` is confirmed. Solid does not need a `key` when rendering a list, but the rule reports one:

```tsx
// packages/ariakit-solid-components/src/example.tsx
export default function Example(props: { items: string[] }) {
  return (
    <ul>
      {props.items.map((item) => (
        <li>{item}</li>
      ))}
    </ul>
  );
}
```

```
example.tsx:4:20: error react(jsx-key): Missing "key" prop for element in iterator.
```

`react/no-children-prop` is suspected for the same reason, since passing `children` as a prop is ordinary in Solid. It has not been confirmed.

## Evidence that this is pre-existing

Reproduced on `main` using the `oxlint.config.ts` from [`65b244f3e`](https://github.com/ariakit/ariakit/commit/65b244f3e11d648c7554f436e6ef3996d9646da7), which declares the `react` plugin at the top level with `categories.correctness: "error"`. `react/jsx-key` is a `correctness` rule, so it has been active on every file, including Solid source, for as long as the plugin has been enabled. This is not introduced by #7248.

## Why this can wait

No committed Solid file triggers it today, so `pnpm lint` is green. Every Solid list in the repository currently uses `<For>` or does not render a list at all. The failure only appears when someone writes `.map()` in Solid source.

## Expected outcome

Decide whether these rules should be disabled for `**/*.solid.*` and `packages/ariakit-solid*/**` alongside the React Compiler rules that #7248 already disables there, or whether Solid should eventually get its own rule set. Confirm whether `react/no-children-prop` misfires as well, and check the remaining `correctness` and `suspicious` React rules against idiomatic Solid rather than adding them one incident at a time.

Discovered during the pre-commit review of #7248.

diegohaz:
## Motivation

Oxlint applies categorized React rules to Solid source, so valid Solid component and JSX patterns can fail lint with React-only assumptions. This includes one-time component setup, unkeyed list output, explicit `children` props, namespaced JSX, string styles, and Solid refs.

Fixes #7250.

## Solution

Disable the React Compiler rules and the React JSX rules whose assumptions do not hold for Solid, while keeping framework-neutral JSX checks and rules that only recognize React APIs enabled.

```ts
{
  files: ["**/*.solid.*", "packages/ariakit-solid*/**"],
  rules: disabledReactRules,
}
```

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- `pnpm test` (268 files, 1,918 tests)
- A temporary idiomatic Solid probe that failed with seven React diagnostics before the fix and passed with the audited override


oxlint.config.test.ts
```diff
@@ -0,0 +1,85 @@
+import { execFileSync } from "node:child_process";
```
diegohaz: We don't need to test this config. Remove this test file.

diegohaz: Fixed in [`eb7565df9`](https://github.com/ariakit/ariakit/commit/eb7565df90228efc5bb33d9f9b53e056c3ac2c49): removed the config audit test and its stale reference while preserving the Solid lint override.

diegohaz:
**Fixed**

- [`eb7565df9`](https://github.com/ariakit/ariakit/commit/eb7565df90228efc5bb33d9f9b53e056c3ac2c49) ([reply](https://github.com/ariakit/ariakit/pull/7259#discussion_r3837883640)): Removed the config audit test and its stale reference while preserving the Solid lint override.

PR #7288, opened 2026-08-25
Expose keyboard shortcut names to assistive technology

### LINKED ISSUE #7279: aria-label on kbd provides no accessible name in three sandboxes
## Context

Found while auditing the Claude Code auto-memory store for this repository. One memory recorded that `aria-label` does not produce an accessible name on an element whose role is `generic`, which is the ARIA mapping for `<span>` and `<kbd>`. Checking the repository for that pattern found 15 occurrences.

## Problem

`<kbd>` maps to the `generic` role, and `aria-label` is prohibited on `generic`. A browser ignores it, so the intended announcement never reaches assistive technology, and the raw glyph is announced instead.

15 occurrences across three sandboxes:

- `app/src/sandbox/combobox-tabs/combobox.react.tsx:122,125,132,135,143`
- `app/src/sandbox/combobox-tabs-old/combobox.react.tsx:142,145,152,155,163`
- `app/src/sandbox/dialog-combobox-tab-command-menu/command-menu.react.tsx:343,346,351,354,368`

Each has this shape:

```tsx
<kbd className="kbd" aria-label="Up Arrow">
```

The intent is clear and correct: announce "Up Arrow" rather than the arrow character. The mechanism does not deliver it.

These are demo surfaces, so the pattern is also copied by readers.

## Why this is filed separately

The audit that found it is complete and concerns a different repository. This is an independent repository defect and does not block anything.

## Expected outcome

Confirm the mapping first in a real browser and a screen reader, because the fix depends on what the browser actually announces for the current markup.

Then replace the prohibited attribute with a mechanism that works on a `generic` element. Two options:

1. Put visually hidden text inside the element and hide the glyph from assistive technology.
2. Give the element a role that permits a name, when that role is accurate for the content.

Apply the same fix to all three sandboxes, and prefer one shared pattern so the demos teach the working one.

### Workaround

https://github.com/ariakit/ariakit/issues/7279#issuecomment-5411998260

diegohaz:
## Motivation

The `aria-label` attribute is prohibited on the generic role that browsers map from `<kbd>`. Chrome therefore exposes the shortcut glyphs as unnamed images in three sandboxes instead of exposing names such as “Up Arrow” and “Enter”.

## Solution

Replace the prohibited labels on all 15 icon-only keys with visually hidden text inside each `<kbd>`, and hide the decorative SVGs from assistive technology.

## Workaround

Give the icon-only `<kbd>` a role that accepts an accessible name:

```tsx
// TODO: Remove this role override after https://github.com/ariakit/ariakit/issues/7279 is fixed.
<kbd role="img" aria-label="Up Arrow">
  <ArrowIcon direction="up" />
</kbd>
```

This covers icon-only shortcut keys by exposing them as named images. It changes their accessibility role and does not preserve the native `<kbd>` content semantics that the final fix must preserve.

## Verification

- On the original markup, local Chromium accessibility snapshots exposed the affected keys only as unnamed images.
- With the temporary workaround, local Chromium accessibility snapshots exposed named images.
- With the final hidden-text solution, local Chromium accessibility snapshots expose generic nodes with text such as “Up Arrow” and no image nodes.
- `pnpm tsc` passes.
- `pnpm lint-fix` passes.

## Preview

The final change affects only accessibility semantics. A focused local screenshot confirmed that no hidden shortcut text became visible, but the screenshot could not be uploaded from this environment. Local Chromium accessibility snapshots prove the name exposure. BrowserStack screen-reader output remains unverified because BrowserStack Local and the remote navigation channel were unavailable in this environment.

Fixes #7279


app/src/sandbox/combobox-tabs-old/test-chrome.ts
```diff
@@ -0,0 +1,31 @@
+import { query } from "@ariakit/test/playwright";
```
diegohaz: No need for this test file.

app/src/sandbox/combobox-tabs/test-chrome.ts
```diff
@@ -0,0 +1,31 @@
+import { query } from "@ariakit/test/playwright";
```
diegohaz: No need for this test file.

app/src/sandbox/dialog-combobox-tab-command-menu/test-chrome.ts
```diff
@@ -0,0 +1,29 @@
+import { query } from "@ariakit/test/playwright";
```
diegohaz: No need for this test file.

app/src/sandbox/combobox-tabs-old/test-chrome.ts
diegohaz: Fixed in [`f0f4e9ba6`](https://github.com/ariakit/ariakit/commit/f0f4e9ba6a4df52b92bba1c98d20cff51682d077): removed this unnecessary sandbox Chrome test file.

app/src/sandbox/combobox-tabs/test-chrome.ts
diegohaz: Fixed in [`f0f4e9ba6`](https://github.com/ariakit/ariakit/commit/f0f4e9ba6a4df52b92bba1c98d20cff51682d077): removed this unnecessary sandbox Chrome test file.

app/src/sandbox/dialog-combobox-tab-command-menu/test-chrome.ts
diegohaz: Fixed in [`f0f4e9ba6`](https://github.com/ariakit/ariakit/commit/f0f4e9ba6a4df52b92bba1c98d20cff51682d077): removed this unnecessary sandbox Chrome test file.

diegohaz:
**Fixed**

- [`f0f4e9ba6`](https://github.com/ariakit/ariakit/commit/f0f4e9ba6a4df52b92bba1c98d20cff51682d077) ([combobox-tabs-old reply](https://github.com/ariakit/ariakit/pull/7288#discussion_r3854220971), [combobox-tabs reply](https://github.com/ariakit/ariakit/pull/7288#discussion_r3854224473), [command-menu reply](https://github.com/ariakit/ariakit/pull/7288#discussion_r3854227984)): Removed the three unnecessary standalone Chrome test files and updated the PR verification text.

PR #7289, opened 2026-08-25
Remove ineffective lite preview scripts

### LINKED ISSUE #7278: preview-lite sets DISABLE_REFERENCE_LINKS after the build, where it has no effect
## Context

Found while auditing the Claude Code auto-memory store for this repository. One memory described `preview-lite` as a way to preview the app without reference links. Verifying that claim showed the script cannot do it.

## Problem

`DISABLE_REFERENCE_LINKS` is read in three places, and all three read it through `import.meta.env`:

- `app/src/components/code-block-content.astro:94`
- `app/src/components/content-link.astro:27`
- `app/src/pages/[...partials].astro:22`, inside `getStaticPaths`

The decisive one is `getStaticPaths`. It runs when the route is built, so the set of partial pages is fixed in the build output.

`app/package.json:26` defines the script as:

```json
"preview-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run preview",
```

and `preview` is `astro preview`, which serves the existing build and never re-runs the build. So `preview-lite` sets a variable that the already-built output cannot observe. The same applies to `preview-stripe-lite` in the same file, and to `preview-app-lite` in the root `package.json:15`.

The name is the problem rather than the behavior. `preview-app-lite` reads as "preview the lite build", so a reader who runs it against a full build believes they are looking at a link-free app while they are looking at whatever was built last. The working pairing is `build-lite` followed by plain `preview`, because the build is where the variable takes effect.

`dev-lite` and `build-lite` are correct, because `astro dev` and `astro build` both evaluate the variable.

## Why this is filed separately

The audit that found it is complete and concerns a different repository. This is an independent repository defect and does not block anything.

## Expected outcome

Decide between two fixes and apply one:

1. Remove `preview-lite`, `preview-stripe-lite`, and `preview-app-lite`, and document that the lite mode is selected at build time with `build-lite` or `build-app-lite`.
2. Keep the names and make them build first, so the script matches what its name promises.

Confirm the behavior before choosing. Build the app twice, once with `build-lite` and once with `build`, then serve each with plain `preview` and with `preview-lite`, and compare the rendered reference links.

diegohaz:
## Motivation

`DISABLE_REFERENCE_LINKS` is evaluated when Astro builds the app, but the lite preview scripts set it only while serving an existing build. Their output therefore depends on the previous build and can misrepresent a full build as lite.

Fixes #7278.

## Solution

Remove `preview-lite`, `preview-stripe-lite`, and `preview-app-lite`. Playwright uses ordinary preview so it serves the existing build without rebuilding, and the contributor guide documents that `build-app-lite` selects lite mode before `preview-app` serves the output.


app/package.json
```diff
@@ -23,7 +23,7 @@
     "build-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run build",
     "build-min": "cross-env DISABLE_SYNTAX_HIGHLIGHTING=true pnpm run build-lite",
     "preview": "cross-env ASTRO_PREVIEW_BACKGROUND=0 astro preview",
-    "preview-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run preview",
+    "preview-lite": "pnpm run build-lite && pnpm run preview",
     "preview-stripe": "concurrently -k -r 'pnpm:preview' 'pnpm:listen-*'",
     "preview-stripe-lite": "concurrently -k -r 'pnpm:preview-lite' 'pnpm:listen-*'",
```
diegohaz: We probably don't need those `lite` versions. We can simply call `build-lite` before preview.

package.json
```diff
@@ -12,7 +12,7 @@
     "build-app": "pnpm -F nextjs run build-worker && pnpm -F app run build",
     "build-app-lite": "pnpm -F nextjs run build-worker && pnpm -F app run build-lite",
     "preview-app": "conc -r 'pnpm -F app run preview' 'pnpm -F nextjs run preview'",
-    "preview-app-lite": "cross-env DISABLE_REFERENCE_LINKS=true pnpm run preview-app",
+    "preview-app-lite": "pnpm run build-app-lite && pnpm run preview-app",
```
diegohaz: Same here

app/package.json
diegohaz: Fixed in [`064f8dd`](https://github.com/ariakit/ariakit/commit/064f8dd0a681fed7b7548d226dbc43d20e654158): removed `preview-lite` and `preview-stripe-lite`; lite mode is selected with `build-lite` before ordinary preview. Evidence gap: No automated regression test was added at the user's request; the build/preview behavior was verified manually.

package.json
diegohaz: Fixed in [`064f8dd`](https://github.com/ariakit/ariakit/commit/064f8dd0a681fed7b7548d226dbc43d20e654158): removed `preview-app-lite`; `build-app-lite` now selects lite mode before `preview-app` serves the output. Evidence gap: No automated regression test was added at the user's request; the build/preview behavior was verified manually.

diegohaz:
**Fixed**

- [`064f8dd`](https://github.com/ariakit/ariakit/commit/064f8dd0a681fed7b7548d226dbc43d20e654158) ([app scripts reply](https://github.com/ariakit/ariakit/pull/7289#discussion_r3854096146), [root script reply](https://github.com/ariakit/ariakit/pull/7289#discussion_r3854099380)): Removed `preview-lite`, `preview-stripe-lite`, and `preview-app-lite`; documented building lite output before ordinary preview.

**Evidence gaps**

- [`064f8dd`](https://github.com/ariakit/ariakit/commit/064f8dd0a681fed7b7548d226dbc43d20e654158) ([app scripts reply](https://github.com/ariakit/ariakit/pull/7289#discussion_r3854096146), [root script reply](https://github.com/ariakit/ariakit/pull/7289#discussion_r3854099380)): No automated regression test was added at the user's request; the manual build-and-preview comparison verified the behavior.

PR #7304, opened 2026-08-28
Fix ComboboxPopover role when a nested ComboboxList has a different popup role

### LINKED ISSUE #7302: ComboboxPopover keeps role="listbox" when a nested ComboboxList has a different popup role
> [!NOTE]
> Posted by `ariakit-bot` on behalf of @diegohaz.

### Current behavior

`ComboboxList` suppresses its own `role="listbox"` when it finds a listbox already inside it, so that `ComboboxPopover > ComboboxList` does not produce nested listboxes. The detection is hardcoded to one role (`combobox-list.tsx:122`):

```ts
setHasListboxInside(!!element.querySelector("[role='listbox']"));
```

But `ComboboxList`'s own JSDoc documents four other roles as valid (`combobox-list.tsx:196-199`, and the same text at `combobox-popover.tsx:305-308`):

> The `role` prop is set to `listbox` by default, but can be overriden by any other valid combobox popup role (`listbox`, `menu`, `tree`, `grid` or `dialog`).

So when the nested list carries any role other than `listbox`, the outer popover does not detect it, keeps `role="listbox"`, and the rendered DOM is invalid. Measured output, with `style`/`id`/`aria-labelledby`/`data-*` stripped:

```html
<!-- nested ComboboxList, default role — suppression works -->
<div role="dialog"><div role="listbox"><div role="option">Apple</div></div></div>

<!-- nested ComboboxList role="tree" -->
<div role="listbox"><div role="tree"><div role="treeitem">Apple</div></div></div>

<!-- nested ComboboxList role="menu" -->
<div role="listbox"><div role="menu"><div role="menuitem">Apple</div></div></div>

<!-- nested ComboboxList role="grid" -->
<div role="listbox"><div role="grid"><div role="gridcell">Apple</div></div></div>
```

The first and second are the same component tree one role string apart. One is auto-corrected, the other is not.

axe-core 4.13.0 reports the broken shapes as critical `aria-required-children` violations — _"Element has children which are not allowed: [role=tree]"_ — and reports no violation for the `listbox` case or for the workaround below.

There is a second symptom from the same wrong role. `combobox.tsx:733` computes `aria-haspopup` from that outer element:

```ts
"aria-haspopup": getPopupRole(contentElement, "listbox"),
```

so the input advertises `aria-haspopup="listbox"` while the popup is a tree, menu or grid. With the nested list left at its default role, the same input correctly reports `dialog`.

Keyboard navigation is unaffected — `aria-activedescendant` still lands on the `treeitem`. This is a semantics and assistive-technology bug, not a behavioral one.

Setting the role on `ComboboxPopover` itself and putting items directly inside works correctly in every respect, and is the shape `examples/select-grid` uses. The nested form is needed when the popover must also hold content the popup role may not own — a live result count, a header, a toolbar — since a `tree` may only own `treeitem` and `group` children.

2. Replace `src/app.tsx` with the code below
3. Click the input and type `app`
4. Inspect the popup element in devtools

```jsx
import { useMemo, useState } from "react";

const files = [
  { id: "src", label: "src", folder: true },
  { id: "src/app.tsx", label: "app.tsx", parent: "src" },
  { id: "src/main.tsx", label: "main.tsx", parent: "src" },
  { id: "readme.md", label: "readme.md" },
];

  const [value, setValue] = useState("");
  const matches = useMemo(
    () =>
      files.filter((file) =>
        file.label.toLowerCase().includes(value.toLowerCase()),
      ),
    [value],
  );
  return (
    <Ariakit.ComboboxProvider value={value} setValue={setValue}>
      <Ariakit.Combobox placeholder="Go to file" />
      <Ariakit.ComboboxPopover gutter={4} sameWidth>
        {/* Outside the tree on purpose: a tree may only own treeitem and
            group children, so the live count cannot live inside it. */}
        <div role="status" aria-live="polite">
          {matches.length} files
        </div>
        <Ariakit.ComboboxList role="tree" aria-label="Files">
          {matches.map((file) => (
            <Ariakit.ComboboxItem
              key={file.id}
              value={file.label}
              aria-level={file.parent ? 2 : 1}
              aria-expanded={file.folder ? true : undefined}
            >
              {file.label}
            </Ariakit.ComboboxItem>
          ))}
        </Ariakit.ComboboxList>
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

The popup element is `role="listbox"` and directly contains `role="tree"`. The input is `aria-haspopup="listbox"`. Delete `role="tree"` from `ComboboxList` and the popup element becomes `role="dialog"` and the input `aria-haspopup="dialog"`.

### Expected behavior

The popover should drop its own `role="listbox"` whenever the nested list already carries a popup role — the same auto-correction it performs today when that role happens to be `listbox` — and `aria-haspopup` should report the real popup role.

### Workaround

Set `role="dialog"` on `ComboboxPopover` and keep the real role on the nested list:

```jsx
<Ariakit.ComboboxPopover role="dialog" aria-label="Files">
  <div role="status" aria-live="polite">{matches.length} files</div>
  <Ariakit.ComboboxList role="tree" aria-label="Files">
    ...
  </Ariakit.ComboboxList>
</Ariakit.ComboboxPopover>
```

That yields `dialog > tree > treeitem`, `aria-haspopup="dialog"`, no axe violations, and unchanged keyboard navigation.

Mirroring the role onto the popover instead — `<ComboboxPopover role="tree"><ComboboxList role="tree">` — is _not_ a workaround. It produces `tree > tree > treeitem`, because the outer element still only suppresses its role when it detects a listbox.

[Confirmed workaround and triage evidence](https://github.com/ariakit/ariakit/issues/7302#issuecomment-5449478841)
[Second case and limitations](https://github.com/ariakit/ariakit/issues/7302#issuecomment-5457996095)

### Possible solutions

The detection at `combobox-list.tsx:122` needs to match every popup role a nested list may legitimately carry, not just `listbox`. The surrounding effect needs no other change: the `MutationObserver` at `:124-129` already observes `subtree: true` with `attributeFilter: ["role"]`, so a broader selector re-evaluates correctly.

Widening the selector alone may not be the right fix, though, because the current query already has a scoping problem that a wider one would inherit. A `[role='listbox']` descendant that has nothing to do with the combobox already strips the outer role today:

```jsx
<Ariakit.ComboboxPopover>
  <Ariakit.ComboboxItem value="Apple" />
  <div role="listbox"><div role="option">unrelated</div></div>
</Ariakit.ComboboxPopover>
```

renders `<div role="dialog"><div role="option">Apple</div><div role="listbox">…</div></div>`, leaving the real `ComboboxItem` outside any listbox. Scoping the query to descendants that are actually `ComboboxList` elements for the same store would fix both, rather than widening an existing hazard.

Two smaller things belong with the change: `hasListboxInside` / `setHasListboxInside` (`combobox-list.tsx:103`, `:134`) want a role-neutral name, and the comment at `:112-115` still says "we can't have nested listbox roles". If the selector should be derived rather than inlined, `allowedPopupRoles` (`packages/ariakit-utils/src/dom.ts:364`) is currently module-private and would need to be exported.

Scope: `hasListboxInside` exists only in `packages/ariakit-react-components/src/combobox/combobox-list.tsx`. There is no equivalent check in `ariakit-solid-components`, `ariakit-components`, `select/select-list.tsx` or `menu/menu-list.tsx`.

One adjacent gap noticed while testing, listed so it is not conflated with this one: `itemRoleByPopupRole` (`dom.ts:366-370`) has no `grid` entry, so a `ComboboxItem` inside `role="grid"` falls back to `role="option"` unless the consumer passes `role="gridcell"` explicitly, as `examples/select-grid` does.

**Evidence note.** This defect has no visual manifestation — the two popups render identically, and the difference is only in the roles and the accessibility tree — so there is no screenshot to attach. The evidence is the DOM output and the axe-core results above, both captured from runs under vitest with happy-dom. It has not been verified in a real browser, and no screen reader was run, so the assistive-technology impact rests on the ARIA specification and axe-core's rule rather than on observed announcements.

diegohaz:
## Motivation

`ComboboxList` documents that its `role` prop accepts any valid combobox popup role (`listbox`, `menu`, `tree`, `grid` or `dialog`). It also drops its own default `role="listbox"` when it already contains a nested list, so `ComboboxPopover > ComboboxList` does not produce nested listboxes.

That detection only looks for one role:

```ts
setHasListboxInside(!!element.querySelector("[role='listbox']"));
```

So the two behaviors disagree. A nested list with any role other than `listbox` is never detected, the popover keeps `role="listbox"`, and the rendered DOM is invalid:

```html
<!-- nested ComboboxList, default role -->
<div role="dialog"><div role="listbox"><div role="option">Apple</div></div></div>

<!-- nested ComboboxList role="tree" -->
<div role="listbox"><div role="tree"><div role="treeitem">Apple</div></div></div>
```

The same component tree, one role string apart. One is auto-corrected, the other is not. axe-core reports the second shape as a critical `aria-required-children` violation. The combobox input also derives `aria-haspopup` from that element, so it advertises `listbox` while the popup is a tree, menu or grid.

The nested form is needed whenever the popover must hold content that the popup role may not own, such as a live result count, a header or a toolbar. A `tree` may only own `treeitem` and `group` children, so that content cannot live inside the list.

The same query has a second problem. It matches any `[role='listbox']` descendant, including markup the combobox does not own:

```jsx
<Ariakit.ComboboxPopover>
  <Ariakit.ComboboxItem value="Apple" />
  <div role="listbox"><div role="option">unrelated</div></div>
</Ariakit.ComboboxPopover>
```

This renders `<div role="dialog"><div role="option">Apple</div><div role="listbox">…</div></div>`, leaving the real `ComboboxItem` outside any listbox.

Fixes #7302

## Solution

Both problems come from asking the DOM the wrong question. `[role='listbox']` is neither necessary nor sufficient: it misses the four other roles a nested list may carry, and it matches elements that are not `ComboboxList` at all.

So the query now asks for component identity instead of a role. `useComboboxList` renders a `data-combobox-list` marker on every list host, and detection looks for that:

```diff
-setHasListboxInside(!!element.querySelector("[role='listbox']"));
+setHasNestedList(!!element.querySelector("[data-combobox-list]"));
```

Everything else about the existing effect is unchanged, including its `[mounted, contentElement]` dependencies and its `contentElement !== element` guard. The net change to the library is one file.

This is role-agnostic by construction, so it covers every popup role the prop accepts rather than enumerating them, and it is scoped to real `ComboboxList` elements, so unrelated markup can no longer take the popup role away from the popover.

Two properties fall out of using a DOM query rather than a React one, and both are load-bearing:

- `Element.querySelector` never matches the element it is called on, so `<ComboboxPopover render={<ComboboxList />}>`, which runs the hook twice on a single element, needs no special case.
- Only DOM descendants match, so a list rendered elsewhere through a portal cannot claim the popup role, and a list portaled *into* the popup from another part of the React tree is still found.

The observer also watches less than before. `main` passed `attributeFilter: ["role"]`, which implies `attributes: true`, so it woke on every `role` mutation anywhere in the popup subtree. The marker is static and can only enter or leave the subtree with its element, so `childList` alone is enough:

```diff
 observer.observe(element, {
   subtree: true,
   childList: true,
-  attributeFilter: ["role"],
 });
```

Per-list role reactivity is unaffected, because `useAttribute` already installs its own `role` observer on each list host.

### Notes

- `ComboboxPopover` composes `useComboboxList`, so the popover element carries `data-combobox-list` too. This matches how `data-dialog` works on `Dialog` and its derived components.
- Legacy `SelectList` renders `role="listbox"` by default only when it has no combobox store and its child store differs, and it carries no marker either way. Such an element nested inside a combobox popup no longer strips that popup's role, which is the same correction as the unrelated-listbox case above.

### Scope

- Middle lists, meaning `ComboboxPopover > ComboboxList > ComboboxList`, are not a supported composition. The supported shape is `ComboboxPopover > ComboboxList`.
- Detection is not scoped to the combobox store. A `ComboboxList` that is a DOM descendant of the popup takes the popup role regardless of which store it belongs to. For the default role this matches `main` exactly. It differs from `main` only when a foreign inline list carries a non-`listbox` role, where `main` produced invalid `listbox > tree` anyway.
- A popover whose only listbox-role content is markup the combobox does not own now keeps `role="listbox"`, so that foreign listbox becomes an invalid child of a listbox. On `main` the popover demoted itself to `dialog` and the foreign listbox was valid. That demotion is exactly what left a popover's own `ComboboxItem` children with no owning listbox, which is the second symptom this fixes, so the two cannot both be satisfied. Consumers who hit this should nest their own items in a `ComboboxList` and pass no `role` to the popover, which then renders as a `dialog`, or move the foreign widget out of the popup.

### Not covered here

- `ComboboxRow` and `ComboboxGroup` read the popup role from the popover rather than the nearest list, so a nested `ComboboxList role="grid"` loses `row` and `rowgroup`. Confirmed pre-existing on `main` and unchanged by this pull request. Tracked in #7305 and fixed in #7314, which is now merged. This branch and that change touch the same file but not the same regions; applying both and running the suites gives 1927 vitest tests and 837 Chrome tests passing.
- Before a popup is opened for the first time, the popover still renders `role="listbox"` and the input still reports `aria-haspopup="listbox"`, because detection is gated on `mounted`. Confirmed pre-existing on `main` by reverting `combobox-list.tsx` to `a8a5181` and re-measuring. Tracked in #7317.

`aria-haspopup` settles one task after the popup role resolves. Measured in Chrome, the popup element's `role` changes at `t=10434.5ms` and the input's `aria-haspopup` at `t=10444.5ms`, in separate mutation batches. `ComboboxSelect` synchronizes this with `useAttribute(contentElement, "role")`; `Combobox` does not, and #7306 declined to add it.

## Workaround

Until the fix is released, pass the popup role explicitly to `ComboboxPopover`.

When the real popup role lives on a nested `ComboboxList`, set the popover to `dialog`:

```jsx
{/* TODO: Remove once released. https://github.com/ariakit/ariakit/issues/7302 */}
<Ariakit.ComboboxPopover role="dialog" aria-label="Go to file results">
  <div role="status">{matches.length} files</div>
  <Ariakit.ComboboxList role="tree" aria-label="Files">
    ...
  </Ariakit.ComboboxList>
</Ariakit.ComboboxPopover>
```

When the popover renders its own items and an unrelated listbox strips its role, set it back to `listbox`:

```jsx
{/* TODO: Remove once released. https://github.com/ariakit/ariakit/issues/7302 */}
<Ariakit.ComboboxPopover role="listbox" aria-label="Fruit results">
  <Ariakit.ComboboxItem value="Apple" />
  <div role="listbox" aria-label="Recently used">...</div>
</Ariakit.ComboboxPopover>
```

Setting the popover to `dialog` also changes `aria-haspopup` from `listbox` to `dialog`. That is intended, and it is what the fix produces too.

This workaround is narrower than the fix:

- The popup role has to be known when the element is written. A popover that renders a nested list in one mode and direct items in another has to compute the prop, because a fixed `role="dialog"` leaves direct items outside any listbox.
- A nested `ComboboxList role="grid"` is not covered, for the reason in #7305 above.
- In the second case the popover owns its own options again, but the unrelated listbox stays an invalid child of a `listbox`. Consumers who need a clean axe run should nest their items in a `ComboboxList` and pass no `role` to the popover, which then renders as a `dialog`, or move the unrelated widget out of the popup.
- In the second case `aria-multiselectable` is not restored, because the popover's `ComboboxList` logic skips the whole block that carries it. Multi-selectable comboboxes have to pass `aria-multiselectable` explicitly. The fix restores it automatically.

## Tests

`app/src/sandbox/combobox-list-7302` covers six scenarios through the rendered accessibility tree:

- A "Go to file" combobox whose popover holds a live result count and a nested `ComboboxList role="tree"`. The popover must be a `dialog` that contains the tree, `aria-haspopup` must be `dialog`, and the role must survive a filtered and an empty result set.
- A "Run command" combobox with a nested `ComboboxList role="menu"` and a footer hint outside it, proving the behavior is not specific to `tree`.
- A fruit search whose popover renders its own options directly and also embeds a third-party widget that ships its own listbox. The popover must keep `role="listbox"` for its own options.
- A tag search whose popover shares its element with a `ComboboxList` through the `render` prop. The shared element must keep `role="listbox"`.
- An issue search with the same shared element plus a real nested `ComboboxList`. The shared element must become a `dialog` and the nested list must own the popup role.
- A docs search whose popover renders its own options and also portals a `ComboboxList` for the same store into a side panel outside the popup. The popover must keep `role="listbox"` and the pinned list must stay outside it.

The first three failed on [`b6edb36`](https://github.com/ariakit/ariakit/commit/b6edb36f52648c1a9c4c340ce600044d980ede55), where CI recorded `Main / Test`, `Main / Test React 18` and `App / Test Chrome` red.

Both properties of the final query were red-checked against the shipped code:

- Reverting the selector to `element.querySelector("[role='listbox']")` fails exactly the three scenarios that define the bug: the tree search, the menu search and the fruit search.
- Broadening the scan to `document.querySelector("[data-combobox-list]")` fails the fruit search, the tag search and the docs search, which is what pins detection to the popup's own subtree.

The issue search passes on `main` as well; it is a behavior-preservation guard for the shared-element composition rather than a regression guard.

## Evidence

There is nothing to screenshot: the popup renders identically and keyboard navigation is unchanged, as the issue notes. The observable difference is in the accessibility tree, so the evidence is a measurement rather than an image. Both runs below instrument the `combobox-list-7302` preview in Chrome with a `MutationObserver` over `role` and `aria-haspopup`, then click the "Go to file" combobox.

On `main` the popup never demotes itself. The observer recorded **zero** mutations, leaving `listbox > tree`:

| | input `aria-haspopup` | popup `role` |
| --- | --- | --- |
| before open | `listbox` | `listbox` |
| after open | `listbox` | `listbox` |

On this branch the popup demotes as it opens:

| | input `aria-haspopup` | popup `role` |
| --- | --- | --- |
| before open | `listbox` | `listbox` |
| after open | `dialog` | `dialog` |

```
t=10434.5  popup  role           listbox -> dialog
t=10444.5  input  aria-haspopup  listbox -> dialog
```

The two land in separate mutation batches, which is the `aria-haspopup` lag described above. The identical `before open` row on both sides is the pre-existing behavior tracked in #7317.

<details>
<summary>Direction change: React-context registration replaced by a component marker</summary>

**Assessment**

The first implementation replaced the DOM query with a React context (`ComboboxNestedListContext`) that nested lists registered through. It needed five separate conditions to stay correct: upward forwarding, a same-element guard, a containment check, a same-store gate, and element state to re-run registration after refs attach. Four of the five produced their own review findings: stale forwarding when a parent's registrar changed, a portaled list wrongly suppressing the role, an accidental public export on the published `./combobox/combobox-context` subpath, and an untested store gate.

Role ownership is a DOM containment property, so a React-tree mechanism had to reconstruct DOM-ness by hand. Once the maintainer removed the same-store requirement, the only property the context provided that a DOM query could not was gone.

**Decision**

Return to `main`'s query structure and change what it looks for, from a role to a `data-combobox-list` component marker. The library change went from roughly +50/−28 across two files to one file, `combobox-context.tsx` returned to its `main` state, and four of the five conditions disappeared rather than being repaired. Two findings against the context version, the portal-in case and the accidental export, were resolved for free rather than fixed.

**Intentionally unsupported**

Middle lists, and cross-store DOM descendants as described under Scope.

**Registered follow-ups**

#7305 (fixed by #7314), #7317.

</details>

## Review gate reassessment

<details>
<summary>Round 3: continue the marker design</summary>

**Assessment**

Issue severity is moderate and accessibility-only: invalid ARIA with no visual or keyboard symptom, reported by axe as a critical `aria-required-children` violation, plus a `aria-haspopup` value that does not match the popup. Expected frequency is low, because no pre-existing example or sandbox in the repository puts a non-`listbox` role on a nested `ComboboxList`; every non-default popup role in the repo sits on `ComboboxPopover` itself. The supported scope is one composition, `ComboboxPopover > ComboboxList`, in `@ariakit/react` only; there is no equivalent detection in the Solid or vanilla packages. Likely user impact is limited to screen-reader users of the affected shapes, and both workarounds above are confirmed.

Implementation complexity is now zero added branches, zero added state, zero added helpers, zero added abstractions and zero added exports. The public surface added is one internal DOM attribute, matching `data-dialog`, `data-dialog-dismiss` and `data-composite-container`, none of which are documented. The observer watches strictly less than `main`.

The three round-3 findings were a false claim in this description, a pre-existing defect now tracked in #7317, and comment verbosity. None concerned the mechanism.

**Decision**

Continue. No further simplification is available: a React context cannot express DOM containment without reconstructing it, and a widened role selector regresses inline `Popover` and `Menu` descendants, which was measured.

</details>

### DIFF
--- a/.changeset/300-combobox-list-popup-role.md
+++ b/.changeset/300-combobox-list-popup-role.md
@@ -3,19 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Nested `ComboboxList` can own any popup role
-
-[`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) drops its own `listbox` role when it contains a nested [`ComboboxList`](https://ariakit.com/reference/combobox-list), so the two elements never produce nested popup roles. This now happens whichever popup role the nested list carries, not only `listbox`.
-
-```tsx
-<Ariakit.ComboboxPopover>
-  <div role="status">{matches.length} files</div>
-  <Ariakit.ComboboxList role="tree" aria-label="Files">
-    <Ariakit.ComboboxItem value="app.tsx" />
-  </Ariakit.ComboboxList>
-</Ariakit.ComboboxPopover>
-```
-
-The popover above used to render `role="listbox"` around the tree, which is invalid, and [`Combobox`](https://ariakit.com/reference/combobox) reported `aria-haspopup="listbox"`. It now renders as a dialog and reports `aria-haspopup="dialog"`.
-
-[`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) also keeps its own popup role when an unrelated element with `role="listbox"` renders inside it. Only a real nested [`ComboboxList`](https://ariakit.com/reference/combobox-list) takes over the popup role now.
+Fixed [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) rendering an invalid `listbox` role around a nested [`ComboboxList`](https://ariakit.com/reference/combobox-list) that carries a different popup role, and dropping its own role when an unrelated element with `role="listbox"` renders inside it.
--- a/.changeset/300-combobox-list-popup-role.md
+++ b/.changeset/300-combobox-list-popup-role.md
@@ -1,0 +1,21 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Nested `ComboboxList` can own any popup role
+
+[`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) drops its own `listbox` role when it contains a nested [`ComboboxList`](https://ariakit.com/reference/combobox-list), so the two elements never produce nested popup roles. This now happens whichever popup role the nested list carries, not only `listbox`.
+
+```tsx
+<Ariakit.ComboboxPopover>
+  <div role="status">{matches.length} files</div>
+  <Ariakit.ComboboxList role="tree" aria-label="Files">
+    <Ariakit.ComboboxItem value="app.tsx" />
+  </Ariakit.ComboboxList>
+</Ariakit.ComboboxPopover>
+```
+
+The popover above used to render `role="listbox"` around the tree, which is invalid, and [`Combobox`](https://ariakit.com/reference/combobox) reported `aria-haspopup="listbox"`. It now renders as a dialog and reports `aria-haspopup="dialog"`.
+
+[`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) also keeps its own popup role when an unrelated element with `role="listbox"` renders inside it. Only a real nested [`ComboboxList`](https://ariakit.com/reference/combobox-list) takes over the popup role now.
--- a/app/src/sandbox/combobox-list-7302/test.ts
+++ b/app/src/sandbox/combobox-list-7302/test.ts
@@ -1,0 +1,60 @@
+import { click, q, type } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+// https://github.com/ariakit/ariakit/issues/7302
+test("nested tree list owns the popup role", async () => {
+  const combobox = q.combobox("Go to file");
+  await click(combobox);
+
+  const popover = q.dialog("Go to file results");
+  // ComboboxList resolves the popup role asynchronously, so aria-haspopup can
+  // still be catching up when the popover opens.
+  await expect
+    .poll(() => combobox.getAttribute("aria-haspopup"))
+    .toBe("dialog");
+
+  const tree = q.tree("Files");
+  expect(popover).toContainElement(tree);
+  expect(q.within(tree).treeitem.all()).toHaveLength(4);
+  expect(q.within(popover).status()).toHaveTextContent("4 of 4 files");
+
+  // The nested list re-renders on every keystroke, so the popup role must
+  // survive both a filtered and an empty result set.
+  await type("app", combobox);
+  expect(q.dialog("Go to file results")).toBe(popover);
+  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
+  expect(q.within(popover).status()).toHaveTextContent("1 of 4 files");
+  expect(q.within(tree).treeitem.all()).toHaveLength(1);
+
+  await type("zzz", combobox);
+  expect(q.dialog("Go to file results")).toBe(popover);
+  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
+  expect(q.within(popover).status()).toHaveTextContent("0 of 4 files");
+  expect(q.within(tree).treeitem.all()).toHaveLength(0);
+});
+
+// https://github.com/ariakit/ariakit/issues/7302
+test("nested menu list owns the popup role", async () => {
+  const combobox = q.combobox("Run command");
+  await click(combobox);
+
+  const popover = q.dialog("Command results");
+  await expect
+    .poll(() => combobox.getAttribute("aria-haspopup"))
+    .toBe("dialog");
+
+  const menu = q.menu("Commands");
+  expect(popover).toContainElement(menu);
+  expect(q.within(menu).menuitem.all()).toHaveLength(3);
+});
+
+// https://github.com/ariakit/ariakit/issues/7302
+test("unrelated listbox does not take the popup role", async () => {
+  const combobox = q.combobox("Search fruits");
+  await click(combobox);
+
+  const popover = q.listbox("Fruit results");
+  expect(combobox).toHaveAttribute("aria-haspopup", "listbox");
+  expect(popover).toContainElement(q.option("Apple"));
+  expect(popover).toContainElement(q.listbox("Recently used"));
+});
--- a/app/src/sandbox/combobox-list-7302/test.ts
+++ b/app/src/sandbox/combobox-list-7302/test.ts
@@ -58,3 +58,28 @@ test("unrelated listbox does not take the popup role", async () => {
   expect(popover).toContainElement(q.option("Apple"));
   expect(popover).toContainElement(q.listbox("Recently used"));
 });
+
+// https://github.com/ariakit/ariakit/issues/7302
+test("popover sharing an element with a list keeps the popup role", async () => {
+  const combobox = q.combobox("Search tags");
+  await click(combobox);
+
+  const popover = q.listbox("Tag results");
+  expect(combobox).toHaveAttribute("aria-haspopup", "listbox");
+  expect(popover).toContainElement(q.option("Design"));
+});
+
+// https://github.com/ariakit/ariakit/issues/7302
+test("nested list owns the popup role on a shared element", async () => {
+  const combobox = q.combobox("Search issues");
+  await click(combobox);
+
+  const popover = q.dialog("Issue results");
+  await expect
+    .poll(() => combobox.getAttribute("aria-haspopup"))
+    .toBe("dialog");
+
+  const list = q.listbox("Issues");
+  expect(popover).toContainElement(list);
+  expect(list).toContainElement(q.option("Bug report"));
+});
--- a/app/src/sandbox/combobox-list-7302/test.ts
+++ b/app/src/sandbox/combobox-list-7302/test.ts
@@ -7,11 +7,7 @@ test("nested tree list owns the popup role", async () => {
   await click(combobox);
 
   const popover = q.dialog("Go to file results");
-  // ComboboxList resolves the popup role asynchronously, so aria-haspopup can
-  // still be catching up when the popover opens.
-  await expect
-    .poll(() => combobox.getAttribute("aria-haspopup"))
-    .toBe("dialog");
+  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
 
   const tree = q.tree("Files");
   expect(popover).toContainElement(tree);
@@ -39,9 +35,7 @@ test("nested menu list owns the popup role", async () => {
   await click(combobox);
 
   const popover = q.dialog("Command results");
-  await expect
-    .poll(() => combobox.getAttribute("aria-haspopup"))
-    .toBe("dialog");
+  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
 
   const menu = q.menu("Commands");
   expect(popover).toContainElement(menu);
@@ -75,11 +69,26 @@ test("nested list owns the popup role on a shared element", async () => {
   await click(combobox);
 
   const popover = q.dialog("Issue results");
-  await expect
-    .poll(() => combobox.getAttribute("aria-haspopup"))
-    .toBe("dialog");
+  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
 
   const list = q.listbox("Issues");
   expect(popover).toContainElement(list);
   expect(list).toContainElement(q.option("Bug report"));
+});
+
+// https://github.com/ariakit/ariakit/issues/7302
+// https://github.com/ariakit/ariakit/pull/7304#discussion_r3884213906
+test("portaled list for the same store does not take the popup role", async () => {
+  const combobox = q.combobox("Search docs");
+  await click(combobox);
+
+  const popover = q.listbox("Doc results");
+  expect(combobox).toHaveAttribute("aria-haspopup", "listbox");
+  expect(popover).toContainElement(q.option("Getting started"));
+
+  // The pinned list reaches this store through context, but it renders outside
+  // the popup, so it must not take the popup role from it.
+  const pinned = q.listbox("Pinned docs");
+  expect(popover).not.toContainElement(pinned);
+  expect(pinned).toContainElement(q.option("Changelog"));
 });
--- a/packages/ariakit-react-components/src/combobox/combobox-context.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-context.tsx
@@ -15,6 +15,17 @@ export const ComboboxListRoleContext = createContext<string | undefined>(
   undefined,
 );
 
+/**
+ * Lets a nested combobox list register itself with the nearest list above it,
+ * so that list can tell whether it contains another one that owns the popup
+ * role. The registration is forwarded up the chain, so lists that share an
+ * element learn about it too. Registering returns a cleanup function, or
+ * nothing when the element is the receiving list's own element.
+ */
+export const ComboboxNestedListContext = createContext<
+  ((element: Element) => (() => void) | undefined) | null
+>(null);
+
 const ctx = createStoreContext<ComboboxStore>(
   [PopoverContextProvider, CompositeContextProvider],
   [PopoverScopedContextProvider, CompositeScopedContextProvider],
--- a/packages/ariakit-react-components/src/combobox/combobox-context.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-context.tsx
@@ -15,17 +15,6 @@ export const ComboboxListRoleContext = createContext<string | undefined>(
   undefined,
 );
 
-/**
- * Lets a nested combobox list register itself with the nearest list above it,
- * so that list can tell whether it contains another one that owns the popup
- * role. The registration is forwarded up the chain, so lists that share an
- * element learn about it too. Registering returns a cleanup function, or
- * nothing when the element is the receiving list's own element.
- */
-export const ComboboxNestedListContext = createContext<
-  ((element: Element) => (() => void) | undefined) | null
->(null);
-
 const ctx = createStoreContext<ComboboxStore>(
   [PopoverContextProvider, CompositeContextProvider],
   [PopoverScopedContextProvider, CompositeScopedContextProvider],
--- a/packages/ariakit-react-components/src/combobox/combobox-list.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-list.tsx
@@ -25,7 +25,6 @@ import { isHidden } from "../disclosure/disclosure-content.tsx";
 import {
   ComboboxHeadingContext,
   ComboboxListRoleContext,
-  ComboboxNestedListContext,
   ComboboxScopedContextProvider,
   useComboboxContext,
   useComboboxScopedContext,
@@ -101,11 +100,8 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
       ? multiSelectable || undefined
       : undefined;
 
-    const parentList = useContext(ComboboxNestedListContext);
-    // A list for another store is a separate popup that happens to render
-    // inside this one, so it does not own the outer list's popup role.
-    const registerInParentList = scopedContextSameStore ? parentList : null;
-    const [nestedListCount, setNestedListCount] = useState(0);
+    const [hasNestedList, setHasNestedList] = useState(false);
+    const contentElement = useStoreState(store, "contentElement");
     const parentHeadingContext = useContext(ComboboxHeadingContext);
     const headingState = useState<string>();
     const [headingId, setHeadingId] = parentHeadingContext || headingState;
@@ -116,31 +112,22 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
     // We support nested <ComboboxList> elements (usually in the form of
     // ComboboxPopover>ComboboxList). The innermost list owns the popup role,
     // whichever role that is, so an outer list must not render one of its own.
-    // Nested lists register through context so that only a real ComboboxList
-    // suppresses the role, not any descendant with a popup role attribute.
-    const registerNestedList = useEvent((element: Element) => {
-      // ComboboxPopover is built on this same hook, so a ComboboxList passed to
-      // its render prop shares this element rather than nesting inside it.
-      if (element === ref.current) return;
-      setNestedListCount((count) => count + 1);
-      // Both hooks install their own context, so a nested list only reaches the
-      // inner one. Forwarding stops the outer hook from rendering a popup role
-      // on the shared element.
-      const unregisterFromParentList = registerInParentList?.(element);
-      return () => {
-        setNestedListCount((count) => count - 1);
-        unregisterFromParentList?.();
-      };
-    });
-
     useSafeLayoutEffect(() => {
-      if (!registerInParentList) return;
+      if (!mounted) return;
       const element = ref.current;
       if (!element) return;
-      return registerInParentList(element);
-    }, [registerInParentList]);
+      if (contentElement !== element) return;
+      const update = () => {
+        setHasNestedList(!!element.querySelector("[data-combobox-list]"));
+      };
+      // The marker is static, so it can only come and go with its element.
+      const observer = new MutationObserver(update);
+      observer.observe(element, { subtree: true, childList: true });
+      update();
+      return () => observer.disconnect();
+    }, [mounted, contentElement]);
 
-    if (!nestedListCount) {
+    if (!hasNestedList) {
       props = {
         role: "listbox",
         "aria-multiselectable": ariaMultiSelectable,
@@ -156,18 +143,16 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
       props,
       (element) => (
         <ComboboxScopedContextProvider value={store}>
-          <ComboboxNestedListContext.Provider value={registerNestedList}>
-            <ComboboxHeadingContext.Provider value={headingContext}>
-              <DialogHeadingContext.Provider value={setHeadingId}>
-                <ComboboxListRoleContext.Provider value={role}>
-                  {element}
-                </ComboboxListRoleContext.Provider>
-              </DialogHeadingContext.Provider>
-            </ComboboxHeadingContext.Provider>
-          </ComboboxNestedListContext.Provider>
+          <ComboboxHeadingContext.Provider value={headingContext}>
+            <DialogHeadingContext.Provider value={setHeadingId}>
+              <ComboboxListRoleContext.Provider value={role}>
+                {element}
+              </ComboboxListRoleContext.Provider>
+            </DialogHeadingContext.Provider>
+          </ComboboxHeadingContext.Provider>
         </ComboboxScopedContextProvider>
       ),
-      [store, role, headingContext, registerNestedList],
+      [store, role, headingContext],
     );
 
     // When nesting ComboboxList elements, the content element should be
@@ -188,6 +173,9 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
     const labelId = headingId || labelElement?.id;
 
     props = {
+      // Marks a real ComboboxList, so the query above matches only these and
+      // not any other descendant that happens to carry a popup role.
+      "data-combobox-list": "",
       "aria-labelledby": props["aria-label"] != null ? undefined : labelId,
       hidden,
       ...props,
--- a/packages/ariakit-react-components/src/combobox/combobox-list.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-list.tsx
@@ -25,6 +25,7 @@ import { isHidden } from "../disclosure/disclosure-content.tsx";
 import {
   ComboboxHeadingContext,
   ComboboxListRoleContext,
+  ComboboxNestedListContext,
   ComboboxScopedContextProvider,
   useComboboxContext,
   useComboboxScopedContext,
@@ -100,8 +101,11 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
       ? multiSelectable || undefined
       : undefined;
 
-    const [hasListboxInside, setHasListboxInside] = useState(false);
-    const contentElement = useStoreState(store, "contentElement");
+    const parentList = useContext(ComboboxNestedListContext);
+    // A list for another store is a separate popup that happens to render
+    // inside this one, so it does not own the outer list's popup role.
+    const registerInParentList = scopedContextSameStore ? parentList : null;
+    const [nestedListCount, setNestedListCount] = useState(0);
     const parentHeadingContext = useContext(ComboboxHeadingContext);
     const headingState = useState<string>();
     const [headingId, setHeadingId] = parentHeadingContext || headingState;
@@ -110,28 +114,33 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
       [headingId, setHeadingId],
     );
     // We support nested <ComboboxList> elements (usually in the form of
-    // ComboboxPopover>ComboboxList), but we can't have nested listbox roles, so
-    // we check here if there's already a listbox element inside the current
-    // element.
+    // ComboboxPopover>ComboboxList). The innermost list owns the popup role,
+    // whichever role that is, so an outer list must not render one of its own.
+    // Nested lists register through context so that only a real ComboboxList
+    // suppresses the role, not any descendant with a popup role attribute.
+    const registerNestedList = useEvent((element: Element) => {
+      // ComboboxPopover is built on this same hook, so a ComboboxList passed to
+      // its render prop shares this element rather than nesting inside it.
+      if (element === ref.current) return;
+      setNestedListCount((count) => count + 1);
+      // Both hooks install their own context, so a nested list only reaches the
+      // inner one. Forwarding stops the outer hook from rendering a popup role
+      // on the shared element.
+      const unregisterFromParentList = registerInParentList?.(element);
+      return () => {
+        setNestedListCount((count) => count - 1);
+        unregisterFromParentList?.();
+      };
+    });
+
     useSafeLayoutEffect(() => {
-      if (!mounted) return;
+      if (!registerInParentList) return;
       const element = ref.current;
       if (!element) return;
-      if (contentElement !== element) return;
-      const callback = () => {
-        setHasListboxInside(!!element.querySelector("[role='listbox']"));
-      };
-      const observer = new MutationObserver(callback);
-      observer.observe(element, {
-        subtree: true,
-        childList: true,
-        attributeFilter: ["role"],
-      });
-      callback();
-      return () => observer.disconnect();
-    }, [mounted, contentElement]);
+      return registerInParentList(element);
+    }, [registerInParentList]);
 
-    if (!hasListboxInside) {
+    if (!nestedListCount) {
       props = {
         role: "listbox",
         "aria-multiselectable": ariaMultiSelectable,
@@ -147,16 +156,18 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
       props,
       (element) => (
         <ComboboxScopedContextProvider value={store}>
-          <ComboboxHeadingContext.Provider value={headingContext}>
-            <DialogHeadingContext.Provider value={setHeadingId}>
-              <ComboboxListRoleContext.Provider value={role}>
-                {element}
-              </ComboboxListRoleContext.Provider>
-            </DialogHeadingContext.Provider>
-          </ComboboxHeadingContext.Provider>
+          <ComboboxNestedListContext.Provider value={registerNestedList}>
+            <ComboboxHeadingContext.Provider value={headingContext}>
+              <DialogHeadingContext.Provider value={setHeadingId}>
+                <ComboboxListRoleContext.Provider value={role}>
+                  {element}
+                </ComboboxListRoleContext.Provider>
+              </DialogHeadingContext.Provider>
+            </ComboboxHeadingContext.Provider>
+          </ComboboxNestedListContext.Provider>
         </ComboboxScopedContextProvider>
       ),
-      [store, role, headingContext],
+      [store, role, headingContext, registerNestedList],
     );
 
     // When nesting ComboboxList elements, the content element should be
--- a/app/src/sandbox/combobox-list-7302/index.react.tsx
+++ b/app/src/sandbox/combobox-list-7302/index.react.tsx
@@ -28,7 +28,14 @@ function GoToFile() {
   return (
     <Ariakit.ComboboxProvider inputValue={value} setInputValue={setValue}>
       <Ariakit.Combobox aria-label="Go to file" />
-      <Ariakit.ComboboxPopover gutter={4} aria-label="Go to file results">
+      {/* TODO: ComboboxList injects role="listbox" here, so the popup role has
+          to be set explicitly. Remove once the nested list can take over the
+          popup role. https://github.com/ariakit/ariakit/issues/7302 */}
+      <Ariakit.ComboboxPopover
+        role="dialog"
+        gutter={4}
+        aria-label="Go to file results"
+      >
         {/* A tree may only own treeitem and group children, so the live count
             has to stay outside the nested list. */}
         <div role="status">
@@ -53,7 +60,14 @@ function RunCommand() {
   return (
     <Ariakit.ComboboxProvider>
       <Ariakit.Combobox aria-label="Run command" />
-      <Ariakit.ComboboxPopover gutter={4} aria-label="Command results">
+      {/* TODO: ComboboxList injects role="listbox" here, so the popup role has
+          to be set explicitly. Remove once the nested list can take over the
+          popup role. https://github.com/ariakit/ariakit/issues/7302 */}
+      <Ariakit.ComboboxPopover
+        role="dialog"
+        gutter={4}
+        aria-label="Command results"
+      >
         <Ariakit.ComboboxList role="menu" aria-label="Commands">
           {commands.map((command) => (
             <Ariakit.ComboboxItem key={command} value={command} />
@@ -84,7 +98,15 @@ function FruitSearch() {
   return (
     <Ariakit.ComboboxProvider>
       <Ariakit.Combobox aria-label="Search fruits" />
-      <Ariakit.ComboboxPopover gutter={4} aria-label="Fruit results">
+      {/* TODO: The unrelated listbox below makes ComboboxList drop its own
+          role, so the popup role has to be set explicitly. Remove once that
+          listbox stops taking over the popup role.
+          https://github.com/ariakit/ariakit/issues/7302 */}
+      <Ariakit.ComboboxPopover
+        role="listbox"
+        gutter={4}
+        aria-label="Fruit results"
+      >
         {fruits.map((fruit) => (
           <Ariakit.ComboboxItem key={fruit} value={fruit} />
         ))}
--- a/app/src/sandbox/combobox-list-7302/index.react.tsx
+++ b/app/src/sandbox/combobox-list-7302/index.react.tsx
@@ -1,0 +1,105 @@
+import * as Ariakit from "@ariakit/react";
+import { useState } from "react";
+
+interface FileEntry {
+  id: string;
+  label: string;
+  folder?: boolean;
+  parent?: string;
+}
+
+const files: FileEntry[] = [
+  { id: "src", label: "src", folder: true },
+  { id: "src/app.tsx", label: "app.tsx", parent: "src" },
+  { id: "src/main.tsx", label: "main.tsx", parent: "src" },
+  { id: "readme.md", label: "readme.md" },
+];
+
+const commands = ["Reload window", "Toggle sidebar", "Split editor"];
+
+const fruits = ["Apple", "Banana", "Cherry"];
+
+function GoToFile() {
+  const [value, setValue] = useState("");
+  const matches = files.filter((file) =>
+    file.label.toLowerCase().includes(value.toLowerCase()),
+  );
+
+  return (
+    <Ariakit.ComboboxProvider inputValue={value} setInputValue={setValue}>
+      <Ariakit.Combobox aria-label="Go to file" />
+      <Ariakit.ComboboxPopover gutter={4} aria-label="Go to file results">
+        {/* A tree may only own treeitem and group children, so the live count
+            has to stay outside the nested list. */}
+        <div role="status">
+          {matches.length} of {files.length} files
+        </div>
+        <Ariakit.ComboboxList role="tree" aria-label="Files">
+          {matches.map((file) => (
+            <Ariakit.ComboboxItem
+              key={file.id}
+              value={file.label}
+              aria-level={file.parent ? 2 : 1}
+              aria-expanded={file.folder || undefined}
+            />
+          ))}
+        </Ariakit.ComboboxList>
+      </Ariakit.ComboboxPopover>
+    </Ariakit.ComboboxProvider>
+  );
+}
+
+function RunCommand() {
+  return (
+    <Ariakit.ComboboxProvider>
+      <Ariakit.Combobox aria-label="Run command" />
+      <Ariakit.ComboboxPopover gutter={4} aria-label="Command results">
+        <Ariakit.ComboboxList role="menu" aria-label="Commands">
+          {commands.map((command) => (
+            <Ariakit.ComboboxItem key={command} value={command} />
+          ))}
+        </Ariakit.ComboboxList>
+        {/* A menu may only own menuitem, menuitemcheckbox, menuitemradio,
+            group and separator children, so the footer hint has to stay
+            outside the nested list. */}
+        <div>Press Enter to run</div>
+      </Ariakit.ComboboxPopover>
+    </Ariakit.ComboboxProvider>
+  );
+}
+
+// Stands in for a third-party widget that ships its own listbox. The combobox
+// does not own this markup and must not hand its popup role over to it.
+function RecentlyUsed() {
+  return (
+    <div role="listbox" aria-label="Recently used">
+      <div role="option" aria-selected={false}>
+        Papaya
+      </div>
+    </div>
+  );
+}
+
+function FruitSearch() {
+  return (
+    <Ariakit.ComboboxProvider>
+      <Ariakit.Combobox aria-label="Search fruits" />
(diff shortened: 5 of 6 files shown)


.changeset/300-combobox-list-popup-role.md
```diff
@@ -0,0 +1,21 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Nested `ComboboxList` can own any popup role
```
diegohaz: This should be a single-sentence changeset.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
```diff
@@ -147,16 +156,18 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
...
       (element) => (
         <ComboboxScopedContextProvider value={store}>
-          <ComboboxHeadingContext.Provider value={headingContext}>
-            <DialogHeadingContext.Provider value={setHeadingId}>
-              <ComboboxListRoleContext.Provider value={role}>
-                {element}
-              </ComboboxListRoleContext.Provider>
-            </DialogHeadingContext.Provider>
-          </ComboboxHeadingContext.Provider>
+          <ComboboxNestedListContext.Provider value={registerNestedList}>
```
diegohaz: Why do we need a separate context? Couldn't we simply check more roles in the old check? Something like:
```ts
setHasListboxInside(!!element.querySelector("[role='listbox'],[role='tree']"))
```

app/src/sandbox/combobox-list-7302/test.ts
```diff
@@ -0,0 +1,85 @@
...
+test("nested tree list owns the popup role", async () => {
+  const combobox = q.combobox("Go to file");
+  await click(combobox);
+
+  const popover = q.dialog("Go to file results");
+  // ComboboxList resolves the popup role asynchronously, so aria-haspopup can
+  // still be catching up when the popover opens.
+  await expect
+    .poll(() => combobox.getAttribute("aria-haspopup"))
+    .toBe("dialog");
```
diegohaz: The comment records an asynchrony that does not happen, and the poll it justifies weakens the assertion it wraps.

I replaced all three `expect.poll` blocks in this file with `expect(combobox).toHaveAttribute("aria-haspopup", "dialog")` and ran the file 5 times on `pnpm test` and once on `pnpm test-react18`. All 5 tests passed every time. Asserting the attribute directly right after `await click(combobox)` reports `dialog`, with no intermediate value, on both React versions.

The PR description says the same thing for the underlying behavior: "Measured to be correct immediately after the popup opens, so it is not a live defect." So the comment here contradicts the description, and a future reader has to re-derive which one is right.

The poll also costs coverage. If `Combobox` later starts settling `aria-haspopup` through an extra tick, which is exactly what #7306 tracks, `expect.poll` absorbs it and the suite stays green. A direct assertion is the thing that would catch it, and it is the assertion the other two cases in this file already make after `type(...)` on lines 25 and 31.

This anchor covers the comment and the first poll:

```suggestion
  expect(combobox).toHaveAttribute("aria-haspopup", "dialog");
```

The same replacement applies to the other two polls at lines 42-44 and 78-80.

**Assessment**

- Severity: Low (verified, 90% confidence), because no shipped behavior is affected; the cost is a weaker regression net and a misleading comment.
- Priority: P2 (estimate, 85% confidence), because the file is new in this PR and the comment is cheap to correct now.
- Breakage likelihood: Possible (estimate, 70% confidence), because #7306 describes the exact change that would make the polled assertion pass while the synchronous behavior regresses.
- Fix effort: Easy (verified, 95% confidence), because the replacement is the three-line substitution above and I confirmed the suite stays green with it.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
```diff
@@ -110,28 +114,33 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
...
-    // element.
+    // ComboboxPopover>ComboboxList). The innermost list owns the popup role,
+    // whichever role that is, so an outer list must not render one of its own.
+    // Nested lists register through context so that only a real ComboboxList
+    // suppresses the role, not any descendant with a popup role attribute.
+    const registerNestedList = useEvent((element: Element) => {
+      // ComboboxPopover is built on this same hook, so a ComboboxList passed to
+      // its render prop shares this element rather than nesting inside it.
+      if (element === ref.current) return;
+      setNestedListCount((count) => count + 1);
```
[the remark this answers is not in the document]
diegohaz: Confirmed as a regression, but the suggested remedy cannot work at this call site. Both halves are measured below.

**The problem reproduces.** Fixture with a popover that renders its own item plus a same-store list through a React portal:

```tsx
<Ariakit.ComboboxPopover aria-label="Portal results">
  <Ariakit.ComboboxItem value="Apple" />
  {createPortal(
    <Ariakit.ComboboxList aria-label="Elsewhere">
      <Ariakit.ComboboxItem value="Portaled" />
    </Ariakit.ComboboxList>,
    document.body,
  )}
</Ariakit.ComboboxPopover>
```

| | popover `role` | `aria-haspopup` | parent role of `option` "Apple" |
| --- | --- | --- | --- |
| merge base [`e5c8fbd48`](https://github.com/ariakit/ariakit/commit/e5c8fbd483b01fe168b9c9c100244a7dabc8b005) | `listbox` | `listbox` | `listbox` |
| this branch | `dialog` | `dialog` | `dialog` |

React context crosses the portal, so the portaled list registers even though it is not a DOM descendant. The popover drops `role="listbox"` and its own `option` is left with no owning `listbox`, which is the same invalid shape the "unrelated listbox does not take the popup role" test exists to prevent. The old `querySelector` never matched across a portal, so this shape is new.

**The suggested check would disable the fix.** `ref.current` is `null` in the receiving list at the moment a genuinely nested list registers. I instrumented `registerNestedList` to record `ref.current` and `ref.current?.contains(element)` on every call:

| registering element | receiving `ref.current` | `contains(element)` |
| --- | --- | --- |
| portaled list | `null` | `null` |
| genuinely nested list | `null` | `null` |
| list sharing the element | the same element | `true` |

React attaches refs bottom up, so a descendant's layout effect runs before the ancestor host element's ref is assigned. Only the shared-element case has a ref by then, which is exactly why `element === ref.current` works and why a containment test at the same point rejects every real nesting. Adding `ref.current?.contains(element)` here reverts #7302.

A containment test has to run where the receiving list's own ref is already attached, which is that list's own layout effect. That means holding the registered elements (a ref-held `Set` plus a version counter to re-run the effect) and computing the flag there from `element.contains(nested)`, instead of counting at registration time.

**Real-world case.** A consumer renders the results list into a container outside the popup (a side panel, a preview iframe) with `createPortal` while the popover still renders its own `ComboboxItem` children. Every option in the popover then has no owning `listbox`, so screen readers lose the set size and position announcements for that list.

**Assessment**

- Severity: Medium (verified, 85% confidence), because the measured DOM puts `role="option"` directly inside `role="dialog"` with no owning listbox; the assistive-technology consequence rests on the ARIA specification rather than an observed announcement.
- Priority: P2 (estimate, 70% confidence), because this branch introduces it, but it needs a composition no repository fixture uses.
- Breakage likelihood: Rare (estimate, 80% confidence), because it requires a same-store `ComboboxList` rendered through a React portal inside a popover that also renders its own items.
- Fix effort: Moderate (estimate, 70% confidence), because the check cannot stay at registration time and has to move into the receiving list's own layout effect.

.changeset/300-combobox-list-popup-role.md
diegohaz: Confirmed, and the changeset guidance agrees: the multi-paragraph form with a code example is for entries that add or change a public API, while a bug fix takes a short past-tense summary sentence. Nothing here changes the public surface, so the entry is a simple one.

One sentence that keeps both behaviors the tests cover:

```md
Fixed [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) rendering an invalid `listbox` role around a nested [`ComboboxList`](https://ariakit.com/reference/combobox-list) that carries a different popup role, and dropping its own role when an unrelated element with `role="listbox"` renders inside it.
```

Two notes on what the compression drops:

The current entry states the resulting `aria-haspopup="dialog"`. That is a consequence of the popover's own role rather than a second behavior, so folding it out is safe, but if you want it kept the sentence can end with `, which also makes [`Combobox`](https://ariakit.com/reference/combobox) report the real popup role in `aria-haspopup`.`

No contributor credit is required. #7302 was opened by `ariakit-bot` on behalf of you, and the credit policy excludes both.

I verified all three reference slugs return HTTP 200: `combobox-popover`, `combobox-list` and `combobox`.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
diegohaz: I tested the wider selector against the current branch. It regresses two ordinary compositions, so the context is doing work the selector cannot do.

`Popover` (`modal = false`, then `portal = modal` at `popover.tsx:249-250`) and `Menu` (`modal: modalProp = false`, then `portal = modalProp` at `menu.tsx:44-45`) both render inline by default, and their content stays in the DOM while hidden. So a `role="dialog"` or `role="menu"` element sits inside the popup even when nothing is open.

I patched the merge-base file to `element.querySelector("[role='listbox'],[role='menu'],[role='tree'],[role='grid'],[role='dialog']")` and measured the popup role for two fixtures whose popover renders its own `ComboboxItem` children:

| fixture | merge base | wider selector | this branch |
| --- | --- | --- | --- |
| `ComboboxPopover` containing an inline `Ariakit.Popover` | `listbox` | `dialog` | `listbox` |
| `ComboboxPopover` containing an inline `Ariakit.Menu` | `listbox` | `dialog` | `listbox` |

In both wider-selector rows the popover's own options end up with no owning `listbox`, which is a worse and far more common version of the unrelated-listbox hazard the issue already describes.

The narrow form in your comment avoids that, but `[role='listbox'],[role='tree']` only covers `tree`. The [`ComboboxList`](https://ariakit.com/reference/combobox-list) JSDoc documents `listbox`, `menu`, `tree`, `grid` and `dialog`, so it would fix one of the four non-default roles and leave `menu`, `grid` and `dialog` broken. Adding any of the remaining three is what triggers the table above.

A DOM query cannot separate "a nested `ComboboxList` that happens to carry `role='menu'`" from "an `Ariakit.Menu` rendered inside the popup", because after render they are the same attribute on the same kind of element. Registration is what supplies that identity.

The one thing the DOM query gave for free is DOM scoping, which is why the cross-store condition is now needed explicitly and why the portal case in the Codex thread regressed. If you keep the context, that thread has the measurement showing where a containment test can and cannot run.

I also red-checked all three conditions the description claims are necessary, and they hold:

- Removing `if (element === ref.current) return;` fails `popover sharing an element with a list keeps the popup role` and 3 tests in `combobox-multiple-select`.
- Removing the forwarding fails only `nested list owns the popup role on a shared element`.
- Removing the same-store gate fails only `menu-nested-search-interactions > tab in/out search menu`.

diegohaz (commented):
Initial review of all 6 changed files at [`1237d133e`](https://github.com/ariakit/ariakit/commit/1237d133e34ccbdd2557a968115e231494e3db72), against merge base [`e5c8fbd48`](https://github.com/ariakit/ariakit/commit/e5c8fbd483b01fe168b9c9c100244a7dabc8b005), plus the surrounding combobox context: `combobox-popover.tsx`, `combobox-item.tsx`, `combobox-context.tsx`, `useEvent` in `packages/ariakit-react-utils/src/hooks.ts`, and the `portal`/`modal` defaults in `popover.tsx`, `menu.tsx` and `dialog.tsx`.

The direction holds up. I tried to break it from the other side, by testing the cheaper alternative, and the cheaper alternative is worse. Details are in the thread reply.

## What I ran

- `pnpm test --run` on the head: 269 files, 1923 tests, all passing.
- `pnpm test-react18 combobox-list-7302`: passing.
- `gh pr checks 7304`: 48 pass, 1 skipping, including all browser and visual suites, so I did not repeat Playwright locally.
- Four scratch fixtures under a temporary sandbox, plus temporary patches to `combobox-list.tsx`. All removed; the tree is back at the head commit with no diff.

## Verified claims from the description

All three "a test proved each one necessary" claims hold. Removing each condition in turn:

| removed condition | failures |
| --- | --- |
| `if (element === ref.current) return;` | `popover sharing an element with a list keeps the popup role`, plus 3 in `combobox-multiple-select` |
| forwarding to `registerInParentList` | only `nested list owns the popup role on a shared element` |
| the `scopedContextSameStore` gate | only `menu-nested-search-interactions > tab in/out search menu` |

The `aria-haspopup` claim under "Not covered here" also holds: the attribute reads `dialog` immediately after the popover opens, on React 19 and React 18.

## Findings

**Portaled same-store list takes the popup role** (reply on the Codex thread, `combobox-list.tsx:124`). Confirmed as a regression against the merge base and measured both ways. Switching from a DOM query to React context also switched the scope from DOM descendants to React descendants, so a same-store `ComboboxList` behind `createPortal` now suppresses the outer role and leaves the popover's own options with no owning `listbox`. Severity Medium, Priority P2, Breakage likelihood Rare, Fix effort Moderate. The suggested containment check cannot go where Codex proposed it: I instrumented `registerNestedList` and the receiving list's `ref.current` is `null` at registration time for every genuine nesting, because React attaches refs bottom up. The reply has the table and a direction that can work.

**Unnecessary `expect.poll` and an inaccurate comment** (inline, `test.ts:10-14`). The comment states that the popup role resolves asynchronously; it does not, on either React version, across 5 runs. The poll also absorbs exactly the regression #7306 describes. Severity Low, Priority P2, Breakage likelihood Possible, Fix effort Easy.

**Changeset form** (reply on your own thread). Your note is right and matches the changeset guidance for bug fixes. I proposed a single sentence that keeps both covered behaviors, and confirmed all three reference slugs return HTTP 200 and that no contributor credit is due.

## Considered and not raised

- Dropping the `mounted` gate that the old effect had means the outer role is now correct while the popup is closed with `unmountOnHide={false}`, rather than flipping on open. Behavior change, but in the correct direction.
- Three-level nesting makes the outer `nestedListCount` reach 2 through forwarding. Only its truthiness is read, so this is harmless.
- `registerNestedList` in the `useWrapElement` dependency array is redundant, since `useEvent` returns a `useCallback(..., [])`. Harmless, not worth a change.
- Removing the `MutationObserver` and the `contentElement` subscription is a real cost reduction for filtered lists, and `Performance / Chrome (combobox-perf)` reports no significant change either way. The changeset correctly makes no performance claim.
- #7305 and #7306 are open and correctly scoped out. I confirmed both exist and did not re-verify their contents.

## Evidence gaps

No screenshots or recordings. Every finding here is a change to `role` or `aria-haspopup` with no visual manifestation, which #7302 already records, so the evidence is DOM attribute output from local runs rather than a capture.

The portal regression was measured under happy-dom only, not in a real browser, and no screen reader was run. The assistive-technology consequence rests on the ARIA specification and on the axe rule cited in #7302.

packages/ariakit-react-components/src/combobox/combobox-context.tsx
```diff
@@ -15,6 +15,17 @@ export const ComboboxListRoleContext = createContext<string | undefined>(
...
 );
 
+/**
+ * Lets a nested combobox list register itself with the nearest list above it,
+ * so that list can tell whether it contains another one that owns the popup
+ * role. The registration is forwarded up the chain, so lists that share an
+ * element learn about it too. Registering returns a cleanup function, or
+ * nothing when the element is the receiving list's own element.
+ */
+export const ComboboxNestedListContext = createContext<
```
diegohaz: This makes an implementation-only registration callback consumer-importable. `@ariakit/react-components` publishes `./combobox/combobox-context`, while `ComboboxNestedListContext` is used only by `combobox-list.tsx`. A bug fix should not add this accidental public contract.

The registration-only precedents are file-local: `NestedHovercardContext` in `hovercard.tsx`, `SelectListContext` in `select-list.tsx`, and `NestedDialogsContext` in `use-nested-dialogs.tsx`. Please follow the same shape here:

```ts
// combobox-list.tsx
const ComboboxNestedListContext = createContext<
  ((element: Element) => (() => void) | undefined) | null
>(null);
```

Assessment: Severity Low; Priority P2; breakage likelihood Possible, because a consumer can start deep-importing the new symbol; fix effort Easy.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
```diff
@@ -110,28 +114,33 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
...
+      // Both hooks install their own context, so a nested list only reaches the
+      // inner one. Forwarding stops the outer hook from rendering a popup role
+      // on the shared element.
+      const unregisterFromParentList = registerInParentList?.(element);
```
diegohaz: `useEvent` keeps this provider value stable when `registerInParentList` changes. A descendant's effect on lines 136–141 therefore does not clean up and re-register with the new parent chain.

I reproduced it with three nested lists. The outer and middle lists initially use store A, and the innermost list inherits the middle store. Changing the middle list to store B without remounting removes only the middle list's direct registration. The innermost list's old forwarded registration remains on the outer list, so the outer host has no role instead of restoring `role="listbox"`. Changing back to A leaves the stale path in place.

A reusable result section can hit this when it changes from its parent's store to an independently controlled store without remounting its list tree. Direct outer options then lose their listbox owner.

The minimum repair for this lifecycle is a dependency-sensitive callback, following the nested-hovercard pattern:

```ts
const registerNestedList = useCallback(
  (element: Element) => {
    // Existing registration and cleanup.
  },
  [registerInParentList],
);
```

The containment-aware version I tested also depends on the receiving `element`. It passes the A → B → A regression together with all eight focused cases.

Assessment: Severity Medium; Priority P2; breakage likelihood Rare, because it needs a nested store change without a remount; fix effort Easy.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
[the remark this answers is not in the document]
diegohaz: I rechecked this thread while comparing alternatives. There is a smaller safe repair than the `Set` plus version-counter design in my previous reply.

Keep component registration, but hold the receiving list's host element in state and make the registrar dependency-sensitive:

```ts
const [element, setElement] = useState<HTMLType | null>(null);

const registerNestedList = useCallback(
  (child: Element) => {
    const unregisterFromParent = registerInParentList?.(child);
    if (!element || child === element || !element.contains(child)) {
      return unregisterFromParent;
    }
    setNestedListCount((count) => count + 1);
    return () => {
      setNestedListCount((count) => count - 1);
      unregisterFromParent?.();
    };
  },
  [element, registerInParentList],
);
```

Then register this list with its parent in a layout effect keyed by both `element` and `registerInParentList`, and merge `setElement` into the host ref. The state update makes descendants retry after the ancestor ref attaches, before paint. Each ancestor applies `contains` to its own host. Forwarding must happen even when the current receiver rejects the candidate, because a list can portal outside one ancestor and a deeper list can portal back inside a higher ancestor.

I tested this form with the five PR tests plus these cases:

- A same-store list portaled to `document.body`: the outer popover keeps `role="listbox"`.
- A three-level list whose middle list changes from store A to B and back: the outer role changes `none → listbox → none`.
- Two nested lists removed one at a time and remounted: the outer role changes `dialog → dialog → listbox → dialog`, and `aria-multiselectable="true"` is restored only while the outer list owns the role.

Results: 8/8 in the focused file on the current React version, 7/7 on React 18 before the final lifecycle case was added, and 15/15 in `menu-nested-search-interactions` plus `combobox-multiple-select`. The scratch changes are removed and the worktree is back at this exact head.

This keeps the finding at Medium / P2 / Rare. The tested repair changes my fix-effort estimate from Moderate to Easy (80% confidence).

packages/ariakit-react-components/src/combobox/combobox-list.tsx
diegohaz: I tested one more DOM-based alternative: mark real `ComboboxList` hosts instead of widening the role selector.

```tsx
props = {
  "data-combobox-list": "",
  ...props,
};

const update = () => {
  setHasNestedList(!!element.querySelector("[data-combobox-list]"));
};
```

That prototype removes the registration context and forwarding. Relative to this head, it was 26 additions and 56 deletions. It passed the five PR tests, the same-store portal case, and the `2 → 1 → 0 → remount` cleanup case.

It is not equivalent to the current intended store boundary. An inline descendant for another combobox store still has the marker, so it suppresses the outer role. In my three-level A → B store-switch fixture, the outer list stayed without a role instead of restoring `listbox`; its direct option would again have no owning listbox. Preserving the current same-store rule would require extra store bookkeeping that removes most of the simplification.

So a role selector is still invalid because it cannot identify `ComboboxList`. A component marker plus `MutationObserver` is a genuinely smaller option only if physical DOM nesting is intentionally made authoritative across stores. Otherwise, the component context with a containment-aware callback is the simpler equivalent solution.

diegohaz (commented):
Same-SHA re-review of [`1237d133e`](https://github.com/ariakit/ariakit/commit/1237d133e34ccbdd2557a968115e231494e3db72) against merge base [`e5c8fbd48`](https://github.com/ariakit/ariakit/commit/e5c8fbd483b01fe168b9c9c100244a7dabc8b005). I used the [previous marked review](https://github.com/ariakit/ariakit/pull/7304#pullrequestreview-5055747540) as the baseline, read every current review thread, and inspected all six changed files plus the registration precedents in Hovercard, Select, and nested Dialogs.

## Outcome

I would not merge the current shape yet. The context direction is justified, but this implementation has two role-ownership defects:

1. The [known same-store portal regression](https://github.com/ariakit/ariakit/pull/7304#discussion_r3884213906) changes the scope from DOM descendants to React descendants.
2. A [new stale forwarded-registration case](https://github.com/ariakit/ariakit/pull/7304#discussion_r3884767380) keeps the outer role suppressed when an intermediate list changes stores without remounting.

The new context is also [exported from a published subpath](https://github.com/ariakit/ariakit/pull/7304#discussion_r3884767015), although it is an implementation-only registration channel. It should be file-local.

## Recommended repair

Keep component identity and the same-store gate. Hold each receiving list's host in state, and use a dependency-sensitive registrar that forwards first and then checks local DOM containment:

```ts
const [element, setElement] = useState<HTMLType | null>(null);

const registerNestedList = useCallback(
  (child: Element) => {
    const unregisterFromParent = registerInParentList?.(child);
    if (!element || child === element || !element.contains(child)) {
      return unregisterFromParent;
    }
    setNestedListCount((count) => count + 1);
    return () => {
      setNestedListCount((count) => count - 1);
      unregisterFromParent?.();
    };
  },
  [element, registerInParentList],
);
```

Register the current list in a layout effect keyed by `element` and `registerInParentList`, and merge `setElement` into the host ref. This is the smallest equivalent design I found. It keeps the three mechanisms that the existing tests prove necessary: store scoping, same-element exclusion, and upward forwarding. The callback identity also makes subscriptions follow store changes.

## Alternatives compared

| Design | Result | Trade-off |
| --- | --- | --- |
| State-held host plus containment-aware context | Recommended; all tested invariants pass | One extra pre-paint render when a list host attaches |
| Ref-held registration tokens plus receiver layout reconciliation | Correct in principle | More lifecycle state, versioning, and duplicate-element bookkeeping |
| `data-combobox-list` plus `MutationObserver` | 26 additions / 56 deletions in the prototype; 7/8 focused cases pass | Changes the inline cross-store rule, restores subtree scans, and exposes a marker |
| Store-scoped `WeakMap` registry | Directly models store plus element ownership | Needs a new external subscription and registry lifetime rules |
| Wider role selector | Rejected | Cannot distinguish a real `ComboboxList role="menu"` from an unrelated `Menu` or `Popover` |

`useMetadataProps` can identify two list hooks on one shared host and could remove that local special case. It does not solve portal containment or stale parent subscriptions, so it is a helper rather than a replacement.

I added the measured DOM-marker result to the [existing design thread](https://github.com/ariakit/ariakit/pull/7304#discussion_r3884767965).

## Validation

- Exact head, original focused suite: 5/5 pass.
- Exact head plus direct portal and store-switch regressions: both new tests fail.
- Recommended repair, focused suite with portal, A → B → A store switch, and `2 → 1 → 0 → remount` plus `aria-multiselectable`: 8/8 pass.
- Recommended repair on React 18 before the last lifecycle case was added: 7/7 pass.
- Recommended repair in `menu-nested-search-interactions` and `combobox-multiple-select`: 15/15 pass.
- DOM-marker prototype: 7/8 pass; only the cross-store ownership case fails.
- Current GitHub checks: 48 successful, 1 skipped.

All prototypes and added fixtures were scratch work. The worktree is restored to the exact PR head with no diff.

## Current thread disposition

- Portal containment: confirmed. I added the smaller validated repair and updated the fix-effort estimate.
- Single-sentence changeset: confirmed in the previous marked reply; no new evidence, so I did not duplicate it.
- Wider role selector: still rejected. I added the component-marker experiment because it is materially different.
- `expect.poll`: confirmed in the previous marked review; no new evidence, so I did not duplicate it.

No screenshot or recording is attached. These failures only change DOM/ARIA roles and have no visual state to capture. The new regressions were tested in happy-dom, not with a screen reader or a real browser. The existing branch browser jobs are green, but they do not contain the new portal and store-switch cases.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0): the registration context this was reported against no longer exists.

Your diagnosis was right, and it turned out to be an argument against the whole approach rather than for a repair. Role ownership is a DOM containment property, and reconstructing that from React context needed a containment check, a same-element guard and upward forwarding, each of which produced its own finding. Once the same-store requirement was dropped, nothing was left that a DOM query could not do, so detection went back to `main`'s `querySelector` with the selector changed from a role to a component marker:

```ts
setHasNestedList(!!element.querySelector("[data-combobox-list]"));
```

A portaled list is not a descendant of the popup, so it cannot take the popup role, and a list portaled *into* the popup is found, which the context version got wrong in the other direction.

The suggested `ref.current.contains(element)` check is worth recording as unusable at that call site regardless: `ref.current` is `null` in the receiving list at the moment a genuinely nested list registers, because React attaches refs bottom-up, so the check would have rejected every real nesting.

Covered by `portaled list for the same store does not take the popup role` in [`a8ae9c8`](https://github.com/ariakit/ariakit/commit/a8ae9c8963a7908f26ed65573af623c9c6fdb41a), red-checked by broadening the scan to `document.querySelector`, which fails that scenario plus the fruit and tag searches.

.changeset/300-combobox-list-popup-role.md
diegohaz: Fixed in [`18f5dc3`](https://github.com/ariakit/ariakit/commit/18f5dc35a69d40d8d7447e0a1c7de18bb8cb0942): the entry is now a single past-tense sentence.

```md
Fixed [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) rendering an invalid `listbox` role around a nested [`ComboboxList`](https://ariakit.com/reference/combobox-list) that carries a different popup role, and dropping its own role when an unrelated element with `role="listbox"` renders inside it.
```

The multi-paragraph form with a code example is for entries that add or change a public API, and nothing here changes the public surface. The `aria-haspopup` consequence is folded out because it follows from the popover's own role rather than being a second behavior. No contributor credit is owed: #7302 was opened by `ariakit-bot` on behalf of you, and the credit policy excludes both.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
diegohaz: Superseded by [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0). You were right that a separate context was not needed, and it is gone. `combobox-context.tsx` is back to its state on `main`, and the library change is now one file.

The half that does not work is widening the role list. I measured `[role='listbox'],[role='menu'],[role='tree'],[role='grid'],[role='dialog']` against two ordinary compositions: `Popover` and `Menu` both render inline by default and keep their content in the DOM while hidden, so a `ComboboxPopover` containing an inline `Ariakit.Popover` or `Ariakit.Menu` dropped its own role and left its options with no owning listbox. The narrower `[role='listbox'],[role='tree']` avoids that but fixes only one of the four non-default roles.

The problem is that a role cannot express component identity: after render, a nested `ComboboxList role="menu"` and an `Ariakit.Menu` are the same attribute on the same kind of element. So the query asks for identity instead, keeping everything else about the effect as it is on `main`:

```diff
-setHasListboxInside(!!element.querySelector("[role='listbox']"));
+setHasNestedList(!!element.querySelector("[data-combobox-list]"));
```

`data-combobox-list` follows the existing internal-marker pattern (`data-dialog`, `data-dialog-dismiss`, `data-composite-container`), including the empty-string value and its position before `...props`.

Two things then fall out for free that the context version needed explicit conditions for. `querySelector` never matches the element it is called on, so `ComboboxPopover render={<ComboboxList />}` needs no same-element guard; and only DOM descendants match, so no forwarding or store gate is required. The observer also watches less than `main`, since `attributeFilter: ["role"]` implied `attributes: true` and the marker is static.

Red-checked both directions: reverting the selector to `[role='listbox']` fails exactly the three scenarios that define the bug, and broadening the scan to `document.querySelector` fails three including the portal case.

app/src/sandbox/combobox-list-7302/test.ts
diegohaz: Fixed in [`a8ae9c8`](https://github.com/ariakit/ariakit/commit/a8ae9c8963a7908f26ed65573af623c9c6fdb41a): all three polls are direct assertions now, and the comment is gone.

One correction to the reasoning, because it matters for anyone reading this later. The asynchrony is real in a browser, it just does not survive `click()` in happy-dom. Instrumenting the live preview with a `MutationObserver` over `role` and `aria-haspopup` and clicking the combobox records:

```
t=10434.5  popup  role           listbox -> dialog
t=10444.5  input  aria-haspopup  listbox -> dialog
```

Separate batches, so the popup is already a `dialog` while the input still advertises `listbox`. That is expected: `combobox.tsx` computes `aria-haspopup` from `getPopupRole(contentElement, "listbox")` during render with no subscription to that attribute, which is exactly the synchronization `ComboboxSelect` has via `useAttribute(contentElement, "role")` and `Combobox` does not. #7306 declined to add it.

So the conclusion holds but the premise in my earlier comment did not, and the PR description no longer claims the value is "correct immediately after the popup opens". In happy-dom the strict `q.dialog(...)` query on the preceding line already forces the role to settle, and the direct assertion was stable across repeated runs on React 19 and React 18, so it is the stronger assertion to keep.

Separately, before a popup is opened for the first time the input reports `aria-haspopup="listbox"` even when a nested list will own the role, because detection is gated on `mounted`. That is pre-existing on `main` and now tracked in #7317.

packages/ariakit-react-components/src/combobox/combobox-context.tsx
diegohaz: Superseded by [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0): `ComboboxNestedListContext` no longer exists, so nothing new is exported from the published `./combobox/combobox-context` subpath. `git diff origin/main -- packages/ariakit-react-components/src/combobox/combobox-context.tsx` is empty.

The finding was correct and the file-local precedents you cited were the right shape, but the accidental export turned out to be one of several symptoms of using a React context here at all. Detection is a DOM query again, so there is no context to place either way.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
diegohaz: Superseded by [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0): there is no registration, no forwarding and no provider value left, so there is nothing that can go stale.

The finding was correct against that design, and the fix you suggested was the right one for it. I applied it first, with a `useCallback` keyed on `[registerInParentList, listElement]` following the `registerNestedHovercard` precedent, and the reproduction you described stopped happening. It only became moot afterwards, when the same-store requirement was dropped and the context could be removed entirely in favour of a DOM query.

Worth recording that the scenario it fixed, a nested list changing store without remounting, is a middle-list shape (`ComboboxPopover > ComboboxList > ComboboxList`), which is now explicitly out of the supported set.

diegohaz:
Addressed the open review threads. The implementation changed direction while doing so: the React-context registration is gone, and detection is a DOM query again with the selector changed from a role to a `data-combobox-list` component marker. The library change is now one file, `combobox-context.tsx` is back to its state on `main`, and the observer watches less than `main` did. The description has been rewritten for it.

**Fixed**

- [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885497859)): a same-store list rendered through a portal outside the popup no longer takes the popup role, and one portaled into it is now found. Covered by a new scenario in [`a8ae9c8`](https://github.com/ariakit/ariakit/commit/a8ae9c8963a7908f26ed65573af623c9c6fdb41a).
- [`a8ae9c8`](https://github.com/ariakit/ariakit/commit/a8ae9c8963a7908f26ed65573af623c9c6fdb41a) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885500328)): the three `expect.poll` blocks are direct `aria-haspopup` assertions, and the inaccurate comment is gone.
- [`18f5dc3`](https://github.com/ariakit/ariakit/commit/18f5dc35a69d40d8d7447e0a1c7de18bb8cb0942) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885498516)): the changeset is a single past-tense sentence.

**Obsolete**

- [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885499564)): the separate context is removed. Widening the role selector is still not viable, because an inline `Popover` or `Menu` inside the popup matches it and loses the popover its own role; a role cannot express component identity, which is why the marker replaces it.
- [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885500805)): `ComboboxNestedListContext` no longer exists, so nothing new is exported from the published `./combobox/combobox-context` subpath.
- [`36a4e24`](https://github.com/ariakit/ariakit/commit/36a4e243a792c5eecea0c7d141ac454615c224a0) ([reply](https://github.com/ariakit/ariakit/pull/7304#discussion_r3885501299)): there is no registration or forwarding left, so no provider value can go stale. The scenario it affected is a middle-list shape, now explicitly unsupported.

**Tracked**

- [#7317](https://github.com/ariakit/ariakit/issues/7317): before a popup is opened for the first time, the popover still renders `role="listbox"` and the input reports `aria-haspopup="listbox"`, because detection is gated on `mounted`. Confirmed pre-existing on `main` by reverting `combobox-list.tsx` to `a8a5181` and re-measuring.

**Invalid**

- A review round reported that the `MutationObserver` `childList` path had no coverage and that `observer.observe(...)` could be deleted with every test still green. Replacing that call with a no-op fails five tests, in `combobox-tabs` and `combobox-tabs-animated`, whose nested `ComboboxList` mounts inside a `TabPanel` in a later commit. No duplicate fixture was added.

**Evidence gaps**

- The `aria-haspopup` timing and the pre-open values were measured by instrumenting the preview in Chrome, not by an automated test. Neither is assertable without depending on behavior the project has declined to guarantee in #7306.
- The axe-core characterisation of the invalid shapes is taken from the rule definition; this sandbox has no axe wiring, so it was not re-measured here.

`main` picked up #7314 while this was in progress. That change and this one touch the same file but not the same regions; applying both and running the suites gives 1927 vitest tests and 837 Chrome tests passing.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
```diff
@@ -110,28 +110,24 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
...
+    // whichever role that is, so an outer list must not render one of its own.
     useSafeLayoutEffect(() => {
       if (!mounted) return;
       const element = ref.current;
       if (!element) return;
       if (contentElement !== element) return;
-      const callback = () => {
-        setHasListboxInside(!!element.querySelector("[role='listbox']"));
+      const update = () => {
+        setHasNestedList(!!element.querySelector("[data-combobox-list]"));
```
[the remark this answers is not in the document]
diegohaz: We can't ignore them as it would create an invalid accessibility tree anyway.

packages/ariakit-react-components/src/combobox/combobox-list.tsx
```diff
@@ -110,28 +110,24 @@ export const useComboboxList = createHook<TagName, ComboboxListOptions>(
...
-      const observer = new MutationObserver(callback);
-      observer.observe(element, {
-        subtree: true,
-        childList: true,
-        attributeFilter: ["role"],
-      });
-      callback();
+      // The marker is static, so it can only come and go with its element.
+      const observer = new MutationObserver(update);
+      observer.observe(element, { subtree: true, childList: true });
```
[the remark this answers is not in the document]
diegohaz: Out-of-scope.

PR #7314, opened 2026-08-29
Fix Combobox group and row roles in nested grids

### LINKED ISSUE #7305: ComboboxRow and ComboboxGroup ignore a nested ComboboxList popup role
### Current behavior

`ComboboxItem` reads the popup role from the nearest list through `ComboboxListRoleContext` (`combobox-item.tsx:145-146`), so it picks up the role of a nested `ComboboxList`. `ComboboxRow` and `ComboboxGroup` do not. Both read it from the store's `contentElement`, which is always the outermost element (`combobox-list.tsx:164-167` sets `setContentElement` to `null` for a nested list that shares the store):

```ts
// combobox-row.tsx:41-43
const contentElement = useStoreState(store, "contentElement");
const popupRole = getPopupRole(contentElement);
const role = popupRole === "grid" ? "row" : "presentation";
```

```ts
// combobox-group.tsx:42-47
const contentElement = useStoreState(store, "contentElement");
const popupRole = getPopupRole(contentElement);

if (popupRole === "grid") {
  props = { role: "rowgroup", ...props };
}
```

So when the `grid` role lives on a nested `ComboboxList` rather than on `ComboboxPopover`, the rows and groups inside it lose their grid semantics.

Measured with happy-dom on `e5c8fbd`, for this tree:

```jsx
<Ariakit.ComboboxPopover aria-label="Direction results">
  <div role="status">2 results</div>
  <Ariakit.ComboboxList role="grid" aria-label="Directions">
    <Ariakit.ComboboxGroup>
      <Ariakit.ComboboxRow>
        <Ariakit.ComboboxItem role="gridcell" value="Top Left" />
        <Ariakit.ComboboxItem role="gridcell" value="Top Right" />
      </Ariakit.ComboboxRow>
    </Ariakit.ComboboxGroup>
  </Ariakit.ComboboxList>
</Ariakit.ComboboxPopover>
```

```text
popoverRole: "listbox"
haspopup:    "listbox"
groupRole:   "group"
rowRole:     "presentation"
```

`ComboboxRow` is `presentation` and `ComboboxGroup` is `group`, so the grid has no rows and no row groups. Assistive technology cannot announce row and column position. Putting `role="grid"` directly on `ComboboxPopover`, as `examples/select-grid` and `app/src/sandbox/combobox-grid` do, works correctly. The nested form is needed when the popover must also hold content that a `grid` may not own, such as a live result count or a header.

### Expected behavior

`ComboboxRow` and `ComboboxGroup` should resolve the popup role from the nearest `ComboboxList`, the same way `ComboboxItem` already does, so a nested `ComboboxList role="grid"` produces `grid > rowgroup > row > gridcell`.

### Context

Found while reviewing https://github.com/ariakit/ariakit/pull/7304, which fixes #7302. That pull request makes `ComboboxPopover` drop its own `role="listbox"` whenever a nested `ComboboxList` carries the popup role, so the popover in the tree above becomes `dialog` instead of `listbox`.

This gap is independent of that change and survives it. `getPopupRole(contentElement)` returns `dialog` after the fix and `listbox` before it, and neither equals `grid`, so `ComboboxRow` and `ComboboxGroup` behave identically in both cases. The measurement above was taken with `packages/` untouched, so the behavior is pre-existing on `main`.

#7302 lists a related but separate gap in the same area: `itemRoleByPopupRole` (`packages/ariakit-utils/src/dom.ts:366-370`) has no `grid` entry, so a `ComboboxItem` inside `role="grid"` falls back to `role="option"` unless the consumer passes `role="gridcell"` explicitly.

### Workaround

Put the popup role on `ComboboxPopover` itself and keep the rows and items directly inside it, as `app/src/sandbox/combobox-grid` does. Content that a `grid` may not own has to move out of the popup.

diegohaz:
## Motivation

`ComboboxGroup` and `ComboboxRow` resolved their roles from the combobox store's outer `contentElement`. When a nested `ComboboxList` owned `role="grid"`, the group and row ignored that nearest list and rendered as `group` and `presentation`, so assistive technology lost the grid's row structure.

Fixes #7305

## Solution

`ComboboxListRoleContext` now carries the nearest list's store and resolved role. `ComboboxGroup` and `ComboboxRow` use that role only when the context belongs to their store. Otherwise, they read the explicit store's `contentElement`. This preserves nested list roles and prevents an outer combobox popup role from leaking into inner explicit-store hooks. `ComboboxItem` keeps its existing nearest-list behavior; #7316 tracks its pre-existing cross-store automatic-role behavior.

```ts
listRole?.store === store
  ? listRole.role
  : getPopupRole(contentElement);
```

The `combobox-grid` sandbox now covers three public role-resolution paths in happy-dom and Chrome:

- a nested `ComboboxList role="grid"`;
- explicit-store hooks with no list-role provider;
- explicit-store hooks rendered under another combobox popup.

The sandbox no longer imports the private role context or exposes hidden role-probe attributes. A patch changeset covers `@ariakit/react-components` and `@ariakit/react`.

## Workaround

Until the fix is released, put the `grid` role on `ComboboxPopover`, render the group and rows directly inside it, and move content that a grid may not own outside the popup.

```tsx
<div role="status">{results.length} results</div>
{/* TODO: Remove after https://github.com/ariakit/ariakit/issues/7305 is fixed. */}
<ComboboxPopover role="grid" aria-label="Results">
  <ComboboxGroup>
    <ComboboxRow>
      <ComboboxItem role="gridcell" value="North West" />
      <ComboboxItem role="gridcell" value="North East" />
    </ComboboxRow>
  </ComboboxGroup>
</ComboboxPopover>
```

This workaround preserves grid, row-group, row, grid-cell, focus, and keyboard behavior. It is narrower than the final fix because headers, live result counts, toolbars, and other content that a grid may not own cannot remain inside the popup.

## Tests

- `pnpm lint-fix`
- `pnpm build`
- `pnpm clean`
- `pnpm tsc`
- `pnpm test`: 268 files and 1,921 tests passed
- `pnpm test app/src/sandbox/combobox-grid/test.ts`: 6 tests passed
- `pnpm test-react18 app/src/sandbox/combobox-grid/test.ts`: 6 tests passed
- `APP_PORT=4324 pnpm -F app run test --project=chrome src/sandbox/combobox-grid/test-chrome.ts --output test-results/7314-comment-response-no-provider`: 3 tests passed
- Red check (cross-store): the cross-store test failed with `group` and `presentation` after the store-ownership condition was temporarily removed, then all 5 focused tests passed after restoration
- Red check (no provider): the standalone test failed with `group` and `presentation` after only the no-provider fallback was temporarily removed, while the other 5 focused tests passed; all 6 tests passed after restoration

## Evidence

The defect changes DOM and accessibility roles without changing the rendered UI. Automated accessibility-tree assertions provide the relevant evidence, so no visual artifact is attached.

<details>
<summary>Cycle 4: Pair list roles with owning stores and remove private probes</summary>

**Assessment**

- **Issue severity:** Medium. An outer combobox role can replace the inner store's grid role and remove row-group and row semantics for assistive technology.
- **Expected frequency:** Low across all comboboxes, but deterministic when explicit-store hooks render under another combobox list or popup.
- **Supported scope:** Nested component composition and documented explicit-store hooks must resolve roles from their own combobox store.
- **Likely user impact:** Assistive technology can receive the wrong structure for an inner grid when an outer combobox provides a nearer role context.
- **Implementation and maintenance complexity:** Low. One memoized context object carries the store and role. Two hooks use the same direct ownership check, and the change removes the three-state scalar sentinel and private test probes.

**Decision**

Change direction from the three-state scalar value to a store-owned context object. This prevents cross-store role leakage, represents a pending same-store role as `role: undefined`, and removes tests coupled to the private context shape. The public nested-combobox fixture covers the reported regression. This direction change resets the implementation review finding count. `ComboboxItem` keeps its pre-existing behavior, which is outside this PR and tracked in #7316.

</details>

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the three-state role-context design</summary>

**Assessment**

- **Issue severity:** Medium. Affected grids lose row-group and row semantics for assistive technology.
- **Expected frequency:** Low across all comboboxes, but deterministic for a grid role on a nested `ComboboxList`.
- **Supported scope:** Nested component composition and documented explicit-store hooks must work. Generic combobox scope must not replace list-role-provider detection, and a pending nearest provider must still take precedence.
- **Likely user impact:** The original bug removes grid position information. Incorrect fallback behavior could also break public hook composition or temporarily apply outer-grid semantics to an inner list.
- **Implementation and maintenance complexity:** Low to moderate. The runtime change adds no store state, observer, helper, or public prop. It adds a three-state context value and the same small precedence branch in two hooks. Each test fixture covers a distinct supported state.

**Decision**

Continue the three-state context design. It is smaller than a second presence context or wrapper object and preserves the required role precedence. Keep direct coverage for the measured role, missing provider, generic select scope, and provider-present pending-role states.

</details>

### DIFF
--- a/app/src/sandbox/combobox-grid/index.react.tsx
+++ b/app/src/sandbox/combobox-grid/index.react.tsx
@@ -1,5 +1,4 @@
 import * as Ariakit from "@ariakit/react";
-import { ComboboxListRoleContext } from "@ariakit/react-components/combobox/combobox-context";
 import { useComboboxGroup } from "@ariakit/react-components/combobox/combobox-group";
 import { useComboboxRow } from "@ariakit/react-components/combobox/combobox-row";
 
@@ -40,61 +39,14 @@ function HookGrid() {
   );
 }
 
-function SelectHookRoles({
-  store,
-  testId = "select-hook-roles",
-}: {
-  store: Ariakit.ComboboxStore;
-  testId?: string;
-}) {
-  const groupProps = useComboboxGroup({ store });
-  const rowProps = useComboboxRow({ store });
-
+function NestedHookGrid() {
   return (
-    <span
-      data-testid={testId}
-      data-group-role={groupProps.role}
-      data-row-role={rowProps.role}
-    />
-  );
-}
-
-function SelectHookGrid() {
-  const store = Ariakit.useComboboxStore();
-
-  return (
-    <>
-      <Ariakit.ComboboxLabel store={store} className="label">
-        Select direction
-      </Ariakit.ComboboxLabel>
-      <Ariakit.ComboboxSelect store={store} className="combobox">
-        Choose direction
-        <SelectHookRoles store={store} />
-        <ComboboxListRoleContext.Provider value={undefined}>
-          <SelectHookRoles store={store} testId="pending-list-hook-roles" />
-        </ComboboxListRoleContext.Provider>
-      </Ariakit.ComboboxSelect>
-      <Ariakit.ComboboxPopover
-        store={store}
-        role="grid"
-        aria-label="Select hook directions"
-        gutter={4}
-        className="popover"
-      >
-        <Ariakit.ComboboxRow className="combobox-row">
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="West"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="East"
-          />
-        </Ariakit.ComboboxRow>
+    <Ariakit.ComboboxProvider defaultOpen>
+      <Ariakit.Combobox aria-label="Outer command" />
+      <Ariakit.ComboboxPopover role="dialog" aria-label="Outer commands">
+        <HookGrid />
       </Ariakit.ComboboxPopover>
-    </>
+    </Ariakit.ComboboxProvider>
   );
 }
 
@@ -190,8 +142,7 @@ export default function Example() {
         </Ariakit.ComboboxPopover>
       </Ariakit.ComboboxProvider>
 
-      <HookGrid />
-      <SelectHookGrid />
+      <NestedHookGrid />
     </>
   );
 }
--- a/app/src/sandbox/combobox-grid/index.react.tsx
+++ b/app/src/sandbox/combobox-grid/index.react.tsx
@@ -1,4 +1,102 @@
 import * as Ariakit from "@ariakit/react";
+import { ComboboxListRoleContext } from "@ariakit/react-components/combobox/combobox-context";
+import { useComboboxGroup } from "@ariakit/react-components/combobox/combobox-group";
+import { useComboboxRow } from "@ariakit/react-components/combobox/combobox-row";
+
+function HookGrid() {
+  const store = Ariakit.useComboboxStore();
+  const groupProps = useComboboxGroup({ store });
+  const rowProps = useComboboxRow({ store });
+
+  return (
+    <>
+      <Ariakit.ComboboxLabel store={store} className="label">
+        Hook direction
+      </Ariakit.ComboboxLabel>
+      <Ariakit.Combobox store={store} className="combobox" />
+      <Ariakit.ComboboxPopover
+        store={store}
+        role="grid"
+        aria-label="Hook directions"
+        gutter={4}
+        className="popover"
+      >
+        <Ariakit.Role {...groupProps}>
+          <Ariakit.Role {...rowProps}>
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="South West"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="South East"
+            />
+          </Ariakit.Role>
+        </Ariakit.Role>
+      </Ariakit.ComboboxPopover>
+    </>
+  );
+}
+
+function SelectHookRoles({
+  store,
+  testId = "select-hook-roles",
+}: {
+  store: Ariakit.ComboboxStore;
+  testId?: string;
+}) {
+  const groupProps = useComboboxGroup({ store });
+  const rowProps = useComboboxRow({ store });
+
+  return (
+    <span
+      data-testid={testId}
+      data-group-role={groupProps.role}
+      data-row-role={rowProps.role}
+    />
+  );
+}
+
+function SelectHookGrid() {
+  const store = Ariakit.useComboboxStore();
+
+  return (
+    <>
+      <Ariakit.ComboboxLabel store={store} className="label">
+        Select direction
+      </Ariakit.ComboboxLabel>
+      <Ariakit.ComboboxSelect store={store} className="combobox">
+        Choose direction
+        <SelectHookRoles store={store} />
+        <ComboboxListRoleContext.Provider value={undefined}>
+          <SelectHookRoles store={store} testId="pending-list-hook-roles" />
+        </ComboboxListRoleContext.Provider>
+      </Ariakit.ComboboxSelect>
+      <Ariakit.ComboboxPopover
+        store={store}
+        role="grid"
+        aria-label="Select hook directions"
+        gutter={4}
+        className="popover"
+      >
+        <Ariakit.ComboboxRow className="combobox-row">
+          <Ariakit.ComboboxItem
+            role="gridcell"
+            className="combobox-item"
+            value="West"
+          />
+          <Ariakit.ComboboxItem
+            role="gridcell"
+            className="combobox-item"
+            value="East"
+          />
+        </Ariakit.ComboboxRow>
+      </Ariakit.ComboboxPopover>
+    </>
+  );
+}
 
 export default function Example() {
   return (
@@ -71,29 +169,29 @@ export default function Example() {
           Grouped direction
         </Ariakit.ComboboxLabel>
         <Ariakit.Combobox className="combobox" />
-        <div role="status">2 results</div>
-        <Ariakit.ComboboxPopover
-          role="grid"
-          aria-label="Grouped directions"
-          gutter={4}
-          className="popover"
-        >
-          <Ariakit.ComboboxGroup>
-            <Ariakit.ComboboxRow className="combobox-row">
-              <Ariakit.ComboboxItem
-                role="gridcell"
-                className="combobox-item"
-                value="North West"
-              />
-              <Ariakit.ComboboxItem
-                role="gridcell"
-                className="combobox-item"
-                value="North East"
-              />
-            </Ariakit.ComboboxRow>
-          </Ariakit.ComboboxGroup>
+        <Ariakit.ComboboxPopover gutter={4} className="popover">
+          <div role="status">2 results</div>
+          <Ariakit.ComboboxList role="grid" aria-label="Grouped directions">
+            <Ariakit.ComboboxGroup>
+              <Ariakit.ComboboxRow className="combobox-row">
+                <Ariakit.ComboboxItem
+                  role="gridcell"
+                  className="combobox-item"
+                  value="North West"
+                />
+                <Ariakit.ComboboxItem
+                  role="gridcell"
+                  className="combobox-item"
+                  value="North East"
+                />
+              </Ariakit.ComboboxRow>
+            </Ariakit.ComboboxGroup>
+          </Ariakit.ComboboxList>
         </Ariakit.ComboboxPopover>
       </Ariakit.ComboboxProvider>
+
+      <HookGrid />
+      <SelectHookGrid />
     </>
   );
 }
--- a/app/src/sandbox/combobox-grid/index.react.tsx
+++ b/app/src/sandbox/combobox-grid/index.react.tsx
@@ -2,62 +2,95 @@ import * as Ariakit from "@ariakit/react";
 
 export default function Example() {
   return (
-    <Ariakit.ComboboxProvider focusWrap={false} focusLoop={false}>
-      <Ariakit.ComboboxLabel className="label">Direction</Ariakit.ComboboxLabel>
-      <Ariakit.Combobox className="combobox" placeholder="e.g., Top, Center" />
-      <Ariakit.ComboboxPopover role="grid" gutter={4} className="popover">
-        <Ariakit.ComboboxRow className="combobox-row">
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Top Left"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Top Center"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Top Right"
-          />
-        </Ariakit.ComboboxRow>
-        <Ariakit.ComboboxRow className="combobox-row">
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Center Left"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Center"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Center Right"
-          />
-        </Ariakit.ComboboxRow>
-        <Ariakit.ComboboxRow className="combobox-row">
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Bottom Left"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Bottom Center"
-          />
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Bottom Right"
-          />
-        </Ariakit.ComboboxRow>
-      </Ariakit.ComboboxPopover>
-    </Ariakit.ComboboxProvider>
+    <>
+      <Ariakit.ComboboxProvider focusWrap={false} focusLoop={false}>
+        <Ariakit.ComboboxLabel className="label">
+          Direction
+        </Ariakit.ComboboxLabel>
+        <Ariakit.Combobox
+          className="combobox"
+          placeholder="e.g., Top, Center"
+        />
+        <Ariakit.ComboboxPopover role="grid" gutter={4} className="popover">
+          <Ariakit.ComboboxRow className="combobox-row">
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Top Left"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Top Center"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Top Right"
+            />
+          </Ariakit.ComboboxRow>
+          <Ariakit.ComboboxRow className="combobox-row">
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Center Left"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Center"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Center Right"
+            />
+          </Ariakit.ComboboxRow>
+          <Ariakit.ComboboxRow className="combobox-row">
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Bottom Left"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Bottom Center"
+            />
+            <Ariakit.ComboboxItem
+              role="gridcell"
+              className="combobox-item"
+              value="Bottom Right"
+            />
+          </Ariakit.ComboboxRow>
+        </Ariakit.ComboboxPopover>
+      </Ariakit.ComboboxProvider>
+
+      <Ariakit.ComboboxProvider>
+        <Ariakit.ComboboxLabel className="label">
+          Grouped direction
+        </Ariakit.ComboboxLabel>
+        <Ariakit.Combobox className="combobox" />
+        <Ariakit.ComboboxPopover gutter={4} className="popover">
+          <div role="status">2 results</div>
+          <Ariakit.ComboboxList role="grid" aria-label="Grouped directions">
+            <Ariakit.ComboboxGroup>
+              <Ariakit.ComboboxRow className="combobox-row">
+                <Ariakit.ComboboxItem
+                  role="gridcell"
+                  className="combobox-item"
+                  value="North West"
+                />
+                <Ariakit.ComboboxItem
+                  role="gridcell"
+                  className="combobox-item"
+                  value="North East"
+                />
+              </Ariakit.ComboboxRow>
+            </Ariakit.ComboboxGroup>
+          </Ariakit.ComboboxList>
+        </Ariakit.ComboboxPopover>
+      </Ariakit.ComboboxProvider>
+    </>
   );
 }
--- a/app/src/sandbox/combobox-grid/index.react.tsx
+++ b/app/src/sandbox/combobox-grid/index.react.tsx
@@ -71,24 +71,27 @@ export default function Example() {
           Grouped direction
         </Ariakit.ComboboxLabel>
         <Ariakit.Combobox className="combobox" />
-        <Ariakit.ComboboxPopover gutter={4} className="popover">
-          <div role="status">2 results</div>
-          <Ariakit.ComboboxList role="grid" aria-label="Grouped directions">
-            <Ariakit.ComboboxGroup>
-              <Ariakit.ComboboxRow className="combobox-row">
-                <Ariakit.ComboboxItem
-                  role="gridcell"
-                  className="combobox-item"
-                  value="North West"
-                />
-                <Ariakit.ComboboxItem
-                  role="gridcell"
-                  className="combobox-item"
-                  value="North East"
-                />
-              </Ariakit.ComboboxRow>
-            </Ariakit.ComboboxGroup>
-          </Ariakit.ComboboxList>
+        <div role="status">2 results</div>
+        <Ariakit.ComboboxPopover
+          role="grid"
+          aria-label="Grouped directions"
+          gutter={4}
+          className="popover"
+        >
+          <Ariakit.ComboboxGroup>
+            <Ariakit.ComboboxRow className="combobox-row">
+              <Ariakit.ComboboxItem
+                role="gridcell"
+                className="combobox-item"
+                value="North West"
+              />
+              <Ariakit.ComboboxItem
+                role="gridcell"
+                className="combobox-item"
+                value="North East"
+              />
+            </Ariakit.ComboboxRow>
+          </Ariakit.ComboboxGroup>
         </Ariakit.ComboboxPopover>
       </Ariakit.ComboboxProvider>
     </>
--- a/app/src/sandbox/combobox-grid/index.react.tsx
+++ b/app/src/sandbox/combobox-grid/index.react.tsx
@@ -2,7 +2,12 @@ import * as Ariakit from "@ariakit/react";
 import { useComboboxGroup } from "@ariakit/react-components/combobox/combobox-group";
 import { useComboboxRow } from "@ariakit/react-components/combobox/combobox-row";
 
-function HookGrid() {
+interface HookGridProps {
+  label: string;
+  popupLabel: string;
+}
+
+function HookGrid({ label, popupLabel }: HookGridProps) {
   const store = Ariakit.useComboboxStore();
   const groupProps = useComboboxGroup({ store });
   const rowProps = useComboboxRow({ store });
@@ -10,13 +15,13 @@ function HookGrid() {
   return (
     <>
       <Ariakit.ComboboxLabel store={store} className="label">
-        Hook direction
+        {label}
       </Ariakit.ComboboxLabel>
       <Ariakit.Combobox store={store} className="combobox" />
       <Ariakit.ComboboxPopover
         store={store}
         role="grid"
-        aria-label="Hook directions"
+        aria-label={popupLabel}
         gutter={4}
         className="popover"
       >
@@ -44,7 +49,10 @@ function NestedHookGrid() {
     <Ariakit.ComboboxProvider defaultOpen>
       <Ariakit.Combobox aria-label="Outer command" />
       <Ariakit.ComboboxPopover role="dialog" aria-label="Outer commands">
-        <HookGrid />
+        <HookGrid
+          label="Nested hook direction"
+          popupLabel="Nested hook directions"
+        />
       </Ariakit.ComboboxPopover>
     </Ariakit.ComboboxProvider>
   );
@@ -142,6 +150,7 @@ export default function Example() {
         </Ariakit.ComboboxPopover>
       </Ariakit.ComboboxProvider>
 
+      <HookGrid label="Hook direction" popupLabel="Hook directions" />
       <NestedHookGrid />
     </>
   );
--- a/packages/ariakit-react-components/src/combobox/combobox-context.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-context.tsx
@@ -11,8 +11,8 @@ import {
 } from "../popover/popover-context.tsx";
 import type { ComboboxStore } from "./combobox-store.ts";
 
-export const ComboboxListRoleContext = createContext<string | undefined>(
-  undefined,
+export const ComboboxListRoleContext = createContext<string | undefined | null>(
+  null,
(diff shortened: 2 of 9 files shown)


packages/ariakit-react-components/src/combobox/combobox-context.tsx
```diff
@@ -11,8 +11,8 @@ import {
 } from "../popover/popover-context.tsx";
 import type { ComboboxStore } from "./combobox-store.ts";
 
-export const ComboboxListRoleContext = createContext<string | undefined>(
-  undefined,
+export const ComboboxListRoleContext = createContext<string | undefined | null>(
```
diegohaz: What does null mean?

packages/ariakit-react-components/src/combobox/combobox-group.tsx
```diff
@@ -40,7 +44,9 @@ export const useComboboxGroup = createHook<TagName, ComboboxGroupOptions>(
     );
 
     const contentElement = useStoreState(store, "contentElement");
-    const popupRole = getPopupRole(contentElement);
+    const contextRole = useContext(ComboboxListRoleContext);
+    const popupRole =
+      contextRole === null ? getPopupRole(contentElement) : contextRole;
```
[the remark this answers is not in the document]
diegohaz: Confirmed. I reproduced it on this head and on the merge base.

Fixture, as `app/src/sandbox/<name>/index.react.tsx`:

```tsx
function InnerGrid() {
  const store = Ariakit.useComboboxStore({ defaultOpen: true });
  const groupProps = useComboboxGroup({ store });
  const rowProps = useComboboxRow({ store });

  return (
    <>
      <Ariakit.Combobox store={store} aria-label="Inner" />
      <Ariakit.ComboboxPopover store={store} role="grid" aria-label="Inner grid">
        <Ariakit.Role {...groupProps} data-testid="inner-group">
          <Ariakit.Role {...rowProps} data-testid="inner-row">
            <Ariakit.ComboboxItem role="gridcell" value="A" />
          </Ariakit.Role>
        </Ariakit.Role>
      </Ariakit.ComboboxPopover>
    </>
  );
}

  return (
    <Ariakit.ComboboxProvider defaultOpen>
      <Ariakit.Combobox aria-label="Outer" />
      <Ariakit.ComboboxPopover aria-label="Outer list">
        <InnerGrid />
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

happy-dom result:

| Ref | group role | row role |
| --- | --- | --- |
| [`a8a5181`](https://github.com/ariakit/ariakit/commit/a8a51813d21b848dcc6240aefc9e5ec1aa4cece0) (merge base) | `rowgroup` | `row` |
| [`2d46991`](https://github.com/ariakit/ariakit/commit/2d469917a58b5c2d41e9d4b8053174ed84f5a89d) (this head) | `group` | `presentation` |

The `useComboboxGroup` JSDoc documents calling the hook outside the popover and spreading the props inside it, so a reusable grid combobox built that way loses `grid > rowgroup > row > gridcell` as soon as it is rendered inside another combobox's list. That is the same defect #7305 describes, moved to a different composition.

Concrete case: a command palette (outer combobox, listbox popover) whose results embed a self-contained picker built as a combobox grid, such as a date or color picker. The palette author renders that picker inside `ComboboxPopover`, and the picker's rows stop being rows for assistive technology.

**Assessment**

- Severity: Medium (verified), because the grid loses row and row-group structure, which is the exact impact this PR fixes elsewhere.
- Priority: P1 (estimate, 85% confidence), because the PR's stated scope is that the documented explicit-store hook pattern keeps working, and the fix is available now.
- Breakage likelihood: Unlikely (estimate, 80% confidence), because it needs an inner combobox with a non-default popup role rendered under another combobox's list.
- Fix effort: Easy (verified), because I implemented and validated the change locally.

Pairing the role with its store resolves this and also answers "What does null mean?" on the other thread. All the states this PR needs fall out of one shape:

```ts
// combobox-context.tsx
export interface ComboboxListRole {
  store: ComboboxStore;
  role: string | undefined;
}

export const ComboboxListRoleContext = createContext<ComboboxListRole | null>(
  null,
);
```

```tsx
// combobox-list.tsx
const listRoleContext = useMemo(() => ({ store, role }), [store, role]);
// ...
<ComboboxListRoleContext.Provider value={listRoleContext}>
```

```ts
// combobox-group.tsx and combobox-row.tsx
const listRole = useContext(ComboboxListRoleContext);
const popupRole =
  listRole?.store === store ? listRole.role : getPopupRole(contentElement);
```

`ComboboxItem` becomes `getItemRole(listRole?.role)`.

`null` now means "no list above", and a provider owned by another store falls back to the explicit store's content element instead of overriding it. The pending state is preserved: same store, `role` still `undefined`, no fallback.

Note that comparing `useComboboxScopedContext()` instead does not work here. `ComboboxSelect` provides that scoped store too (`combobox-select.tsx:167`) without providing a list role, so store equality alone cannot separate "inside my list" from "inside my select".

Validated locally on this branch:

- the fixture above returns `rowgroup` and `row`
- `pnpm test app/src/sandbox/combobox-grid/test.ts`: 7 passed, once the pending fixture is written as `value={{ store, role: undefined }}` instead of `value={undefined}`
- `pnpm test combobox select`: 75 files, 496 tests passed
- `pnpm tsc`: clean

The experiment is reverted and the branch is untouched.

packages/ariakit-react-components/src/combobox/combobox-context.tsx
diegohaz: Today it packs three states into one value: `null` means no `ComboboxList` provider above, `undefined` means a provider is above but `useAttribute` has not measured its `role` yet, and a string is the measured role. Only `null` takes the `getPopupRole(contentElement)` fallback, which is what keeps the documented explicit-store hook pattern working.

That distinction is not recoverable from the type, and no comment carries it. It is also not sufficient: it cannot separate a provider that belongs to this store from one that belongs to another combobox, which is the regression confirmed on the [`useComboboxGroup` thread](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885056841).

Carrying the store next to the role removes the question instead of documenting it:

```ts
export interface ComboboxListRole {
  store: ComboboxStore;
  role: string | undefined;
}

export const ComboboxListRoleContext = createContext<ComboboxListRole | null>(
  null,
);
```

`null` reads as "no list above", `role: undefined` reads as "not measured yet", and `listRole.store === store` decides whether the nearest list is the one this hook belongs to. `ComboboxStore` is already imported in this file. I validated the full change locally, and the commands and results are on the other thread.

If the current shape stays, the sentinel needs a comment above the declaration that states what `null` and `undefined` each mean, and why only `null` falls back.

**Assessment**

- Severity: Low (estimate, 90% confidence), because it affects readability and future maintenance rather than shipped behavior.
- Priority: P2 (estimate, 80% confidence), because the store-paired shape is already required to fix the other thread, so this resolves with it.
- Breakage likelihood: Possible (estimate, 70% confidence), because the next reader of these two hooks has to reconstruct the sentinel meaning before changing either branch.
- Fix effort: Easy (verified), because the replacement shape type-checks and passes the combobox and select suites.

app/src/sandbox/combobox-grid/index.react.tsx
```diff
@@ -1,63 +1,197 @@
...
+
+  return (
+    <>
+      <Ariakit.ComboboxLabel store={store} className="label">
+        Select direction
+      </Ariakit.ComboboxLabel>
+      <Ariakit.ComboboxSelect store={store} className="combobox">
+        Choose direction
+        <SelectHookRoles store={store} />
+        <ComboboxListRoleContext.Provider value={undefined}>
```
diegohaz: This fixture reaches into a library-internal React context to fabricate a state the library already produces on its own, which couples an app sandbox to a private value shape.

Two costs, both observed on this branch:

1. An internal refactor turns into an unrelated sandbox failure. When I changed `ComboboxListRoleContext` to carry `{ store, role }` (see the [`useComboboxGroup` thread](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885056841)), `app/src/sandbox/combobox-grid/test.ts > does not fall back while a nested list role is pending` failed while the library behavior was unchanged and strictly better. The signal was false: only the hand-written provider value was stale.
2. `ComboboxListRoleContext` is reachable only through the deep `@ariakit/react-components/combobox/combobox-context` subpath and is not re-exported from `@ariakit/react`. Consuming it from a sandbox presents a private seam as supported.

`SelectHookRoles` above has the same shape. `data-group-role` and `data-row-role` on a `<span>` inside a `<button>` are not state a user or assistive technology perceives, and `role="rowgroup"` there would not be a real accessibility tree. `ariakit-ariakit-workflow` and `ariakit-ariakit-example-integration-test` both ask for the observable-UI form first, with an internal probe only as a fallback.

The pending state is reachable without the probe: a nested `<ComboboxList>` with no explicit `role` renders one commit with the role unmeasured before `useAttribute` settles it to `listbox`. If that transient cannot be observed deterministically in happy-dom, dropping the pending case costs little. Under the store-paired shape the pending behavior is structural rather than a branch that can silently invert.

**Assessment**

- Severity: Low (estimate, 85% confidence), because no shipped behavior is affected and the cost is maintenance plus misleading test signal.
- Priority: P2 (estimate, 80% confidence), because it does not block release, but it is cheapest to fix while this fixture is being written.
- Breakage likelihood: Likely (verified), because it already produced a false failure under a locally validated refactor of the same context during this review.
- Fix effort: Easy (estimate, 75% confidence), because it is a fixture and test change with no library change.

diegohaz (commented):
## Review summary

Initial review of [`2d46991`](https://github.com/ariakit/ariakit/commit/2d469917a58b5c2d41e9d4b8053174ed84f5a89d) against merge base [`a8a5181`](https://github.com/ariakit/ariakit/commit/a8a51813d21b848dcc6240aefc9e5ec1aa4cece0). No prior review from this workflow exists on this PR, so this is a full review of the whole diff plus the surrounding combobox role-resolution code.

### What I checked

- The full diff across `packages/ariakit-react-components/src/combobox/**`, `app/src/sandbox/combobox-grid/**`, and `.changeset/`.
- Surrounding context: `combobox-list.tsx` role measurement through `useAttribute`, `setContentElement` precedence for nested lists, `ComboboxScopedContextProvider` in both `combobox-list.tsx` and `combobox-select.tsx`, `getPopupRole` and `getItemRoleByPopupRole` in `packages/ariakit-utils/src/dom.ts`, and `useCompositeGroup` and `useCompositeRow` for registration side effects (there are none, so the probe fixture does not create phantom rows).
- Issue #7305 against the implementation.
- All CI checks on this head: green, including Chrome, Firefox, Safari, iOS, Android, visual, React 18, type check, and performance.

### Commands run

- `pnpm test app/src/sandbox/combobox-grid/test.ts`: 7 passed on this head.
- The same file with only the four package files reverted to the merge base: `uses the nested grid role for groups and rows` fails for the expected reason (`group` instead of `rowgroup`), so the regression test is proven red-then-green. `does not fall back while a nested list role is pending` also fails there.
- A temporary sandbox reproducing an inner explicit-store grid combobox under an outer combobox list, run on both refs.
- `pnpm test combobox select`: 75 files, 496 tests passed, with the alternative shape applied.
- `pnpm tsc`: clean, with the alternative shape applied.

All experiments were reverted. The branch was not modified or pushed.

### Findings

Both published findings landed as replies on existing unresolved threads rather than as new inline comments, since accurate threads already anchored them.

1. Cross-store role context breaks the documented explicit-store hook pattern, on [thread r3885056841](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885056841). Validation outcome: confirmed against this head, and confirmed absent on the merge base. Severity Medium (verified), Priority P1, breakage likelihood Unlikely, fix effort Easy. A store-paired context shape fixes it while keeping all four states this PR covers, and I validated that shape locally.
2. The `null` sentinel is undocumented, on [thread r3885096577](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885096577). Validation outcome: confirmed. Severity Low, Priority P2. The same store-paired shape removes the ambiguity; otherwise the declaration needs a comment.
3. The sandbox fixture depends on a library-internal context shape, published inline on `app/src/sandbox/combobox-grid/index.react.tsx`. Severity Low, Priority P2, breakage likelihood Likely (verified during this review), fix effort Easy.

### Description mismatch

The Solution section states that the design lets nested lists take precedence "without breaking the documented explicit-store hook pattern". That holds for a missing provider and for `ComboboxSelect` scope, but not when the nearest provider belongs to a different combobox store. Finding 1 has the reproduction. The description is otherwise accurate: I verified the four covered role-resolution paths, the workaround, and the listed test commands.

### Checked and clear

- Changeset: filename prefix `300-`, `patch` type for both packages, one past-tense sentence, correct reference links. No contributor credit is required, since the issue author is excluded by policy.
- Fixture lifecycle: extending the existing semantic `combobox-grid` sandbox with sibling variants is the correct step. The original grid scenario and its navigation invariants are preserved, and the issue URL sits beside the reproducing test in both the browser test and the happy-dom duplicate.
- Solid parity: not applicable, since `packages/ariakit-solid-components` has no combobox.
- Behavior for `ComboboxItem` is unchanged by the new `null` default, because `getItemRoleByPopupRole` already treated `null` and `undefined` alike and both still resolve to `option`.
- Performance: the provider value is a string, so no new render churn, and the performance job reports no significant change.

### Not treated as findings

- CodeRabbit reported no actionable comments. Its `Docstring Coverage` pre-merge warning is a generic threshold over touched functions, not a documentation gap under this repository's conventions. Declined non-goal. The substantive part of it, the missing explanation on the context declaration, is covered by finding 2.
- Placing `role="grid"` on `ComboboxPopover` with a plain nested `ComboboxList` now yields `presentation` rows where the outer content element previously produced `row`. This is a real precedence change, but nesting a listbox inside a grid is already invalid markup, and the changeset wording describes the new nearest-list rule. Declined non-goal.

### Not fully validated

I did not run the Playwright suite or `pnpm build` locally. CI covers both on this head and both pass.

packages/ariakit-react-components/src/combobox/combobox-group.tsx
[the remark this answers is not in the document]
diegohaz: **Fixed now**

Fixed in [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641): each list role is now paired with its owning store. `ComboboxGroup` and `ComboboxRow` use the contextual role only when that store matches, and otherwise fall back to the explicit store's content element.

The same commit adds a public regression case for explicit-store hooks under another combobox popup. The test fails with `group` and `presentation` when the ownership check is removed, and passes with `grid > rowgroup > row > gridcell` after restoration.

Visual evidence gap: this changes accessibility roles without changing rendered pixels. Happy-dom and Chrome role queries verify the repaired hierarchy.

packages/ariakit-react-components/src/combobox/combobox-context.tsx
diegohaz: **Fixed now**

Fixed in [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641): the context now uses `{ store, role }` instead of an ambiguous scalar sentinel.

`null` now means only that no list provider is above the component. A same-store provider whose role is not measured yet is represented as `{ store, role: undefined }`, and a provider from another store is rejected by the ownership check.

app/src/sandbox/combobox-grid/index.react.tsx
diegohaz: **Fixed now**

Fixed in [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641): the sandbox no longer imports `ComboboxListRoleContext` or exposes hidden role-probe attributes.

The replacement is a user-visible nested combobox. Happy-dom and Chrome assert its real accessibility hierarchy through public role queries, including the explicit-store case under another popup.

Visual evidence gap: this role-only change has no visible rendering difference, so the public accessibility queries are the relevant evidence.

diegohaz:
**Fixed**

- [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641) ([reply](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885205682)): scoped list roles to their owning combobox store and added public cross-store regression coverage.
- [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641) ([reply](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885206381)): replaced the ambiguous scalar role sentinel with a store-owned context value.
- [`5e4d7b636`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641) ([reply](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885207073)): removed private role-context probes from the sandbox and replaced them with public accessibility assertions.

**Declined non-goal**

- ([reply](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885273881)): the grid-specific item-role claim does not reproduce because `grid` falls back to `option` and an authored `gridcell` role wins. Other cross-store item-role behavior is pre-existing and outside #7305.

**Evidence gaps**

- Accessibility roles have no visible rendering difference. Happy-dom and Chrome role queries verify `grid > rowgroup > row > gridcell`.

packages/ariakit-react-components/src/combobox/combobox-item.tsx
```diff
@@ -142,8 +142,8 @@ export const useComboboxItem = createHook<TagName, ComboboxItemOptions>(
     setValueOnClick = setValueOnClick ?? (!selectMode && !multiSelectable);
     hideOnClick = hideOnClick ?? (value != null && !multiSelectable);
     preventScrollOnKeyDown = preventScrollOnKeyDown ?? selectMode;
-    const popupRole = useContext(ComboboxListRoleContext);
-    const role = getItemRole(popupRole);
+    const listRole = useContext(ComboboxListRoleContext);
+    const role = getItemRole(listRole?.role);
```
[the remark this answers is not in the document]
diegohaz: **Declined non-goal**

I verified this against `5e4d7b636` and `origin/main`. The grid-specific failure described here does not occur: `getItemRoleByPopupRole("grid")` has no mapping, so the computed fallback is `option` on both refs, and an authored `role="gridcell"` wins because `...props` is spread after the computed role.

A different outer `menu` or `tree` can affect an explicit-store item's automatic role, but that behavior already exists on `main`. This PR only fixes `ComboboxGroup` and `ComboboxRow`, as stated by the changeset. Fixing the broader item behavior needs separate regression coverage and a performance review because it would add a content-element subscription to every item, so I am keeping it out of #7305.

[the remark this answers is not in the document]
diegohaz: Validated independently. The behavior is real, and it is pre-existing on `main` rather than introduced by this PR, so I registered it instead of asking for it here: #7316.

Reproduction, with an outer `menu` popup and an inner `listbox` popup so the two item roles differ:

```tsx
function InnerList() {
  const store = Ariakit.useComboboxStore({ defaultOpen: true });
  const itemProps = useComboboxItem({ store, value: "A" });

  return (
    <>
      <Ariakit.Combobox store={store} aria-label="Inner" />
      <Ariakit.ComboboxPopover store={store} role="listbox" aria-label="Inner list">
        <Ariakit.Role {...itemProps} data-testid="inner-item" />
      </Ariakit.ComboboxPopover>
    </>
  );
}

  return (
    <Ariakit.ComboboxProvider defaultOpen>
      <Ariakit.Combobox aria-label="Outer" />
      <Ariakit.ComboboxPopover role="menu" aria-label="Outer menu">
        <InnerList />
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

happy-dom result, expected `option`:

| Ref | item role |
| --- | --- |
| [`a8a5181`](https://github.com/ariakit/ariakit/commit/a8a51813d21b848dcc6240aefc9e5ec1aa4cece0) (merge base) | `menuitem` |
| [`5e4d7b6`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641) (this head) | `menuitem` |

Identical, so this PR does not introduce or worsen it. Before this branch, `ComboboxListRoleContext` carried a bare role string, so the nearest provider already won regardless of store. This PR rewrote that read as `listRole?.role` for the new `{ store, role }` shape, which resolves to the same value, and the description states that `ComboboxItem` keeps its existing nearest-list behavior.

Applying the group and row treatment to items is also not a mechanical copy, which is why it needs its own change:

1. `useComboboxItem` has no `contentElement` fallback today. Adding `getPopupRole(contentElement)` would change behavior for an item with an explicit store and no provider above it at all: `option` today, `menuitem` or `treeitem` after. That affects existing consumers and needs its own regression tests and changelog entry.
2. It would not help the grid case that motivated #7305, because `itemRoleByPopupRole` (`packages/ariakit-utils/src/dom.ts:366-370`) has no `grid` entry, so grid items still need an explicit `role="gridcell"`. #7302 tracks that gap.

Disposition: registered follow-up at https://github.com/ariakit/ariakit/issues/7316. The reproduction, the pre-existence evidence, the workaround (`role` on the item wins, since it lands after `role` in the props merge at `combobox-item.tsx:250-252`), and the open decision about the no-provider case are all recorded there.

app/src/sandbox/combobox-grid/index.react.tsx
```diff
@@ -1,63 +1,148 @@
...
-          <Ariakit.ComboboxItem
-            role="gridcell"
-            className="combobox-item"
-            value="Bottom Right"
-          />
-        </Ariakit.ComboboxRow>
+    <Ariakit.ComboboxProvider defaultOpen>
+      <Ariakit.Combobox aria-label="Outer command" />
+      <Ariakit.ComboboxPopover role="dialog" aria-label="Outer commands">
+        <HookGrid />
```
diegohaz: Moving `HookGrid` under `NestedHookGrid` gained the cross-store case but dropped the only fixture that exercised the no-provider case, so the `getPopupRole(contentElement)` fallback for a plain explicit-store hook is now unguarded.

Every current fixture reaches `useComboboxGroup` with a role provider above it: the `Direction` and `Grouped direction` comboboxes match on store, and `NestedHookGrid` mismatches on store. Nothing renders a group or row with `listRole == null` any more.

Mutation evidence, run on this branch:

| Mutation applied to `useComboboxGroup` and `useComboboxRow` | at [`2d46991`](https://github.com/ariakit/ariakit/commit/2d469917a58b5c2d41e9d4b8053174ed84f5a89d) | at [`5e4d7b6`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641) |
| --- | --- | --- |
| Drop the no-provider fallback | 2 tests fail | **5 pass** |
| `listRole?.role ?? getPopupRole(contentElement)` | n/a | 1 test fails |
| `listRole ? listRole.role : getPopupRole(contentElement)` | n/a | 1 test fails |

The middle two rows show the new cross-store test is doing real work. The first row is the regression: at the previous head, removing the fallback failed `preserves grid roles with explicit-store hooks` and `preserves grid roles for hooks under select scope`. At this head the same mutation is invisible.

That fallback is the literal pattern in `useComboboxGroup`'s own JSDoc, where the hook is called at the top of a component and the props are spread inside `ComboboxPopover`. `ariakit-ariakit-example-playwright-test` asks for a source-to-target behavior mapping before removing fixture coverage, and preserving or explicitly declining every unmapped behavior. This one looks dropped rather than declined.

Rendering `HookGrid` standalone next to `NestedHookGrid` restores it. The labels have to differ so the strict `q.combobox()` and `q.grid()` queries stay unambiguous, for example by passing the label text and popup label through props.

**Assessment**

- Severity: Low (verified), because no shipped behavior is affected; the fallback works, it is only untested.
- Priority: P2 (estimate, 80% confidence), because it does not block release, but it is cheapest to restore while this fixture is open.
- Breakage likelihood: Possible (verified by mutation), because a later simplification of the ownership ternary can remove the fallback with a green suite.
- Fix effort: Easy (estimate, 85% confidence), because it is one extra render of an existing component plus one test.

diegohaz (commented):
## Re-review summary

Re-review of [`5e4d7b6`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641). Incremental baseline is [`2d46991`](https://github.com/ariakit/ariakit/commit/2d469917a58b5c2d41e9d4b8053174ed84f5a89d), the head of my previous review with this workflow, found through its marker. Merge base is unchanged at [`a8a5181`](https://github.com/ariakit/ariakit/commit/a8a51813d21b848dcc6240aefc9e5ec1aa4cece0). One new commit, `Keep combobox roles scoped to their store`.

### Previous findings

All three are fixed, and I verified each against this head rather than accepting the replies.

1. Cross-store role leak, [thread r3885056841](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885056841). Fixed. `ComboboxListRoleContext` now carries `{ store, role }` and both hooks gate on `listRole?.store === store`. My original reproduction returns `rowgroup` and `row` on this head, matching the merge base.
2. Undocumented `null` sentinel, [thread r3885096577](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885096577). Fixed. `null` now means only "no provider above". Keeping `ComboboxListRole` unexported is better than the exported form I suggested: it stays a private seam, and I confirmed declaration emit still works, since `dist/combobox/combobox-context.d.ts` emits the local interface and references it from the exported context.
3. Fixture coupled to the private context, [thread r3885130687](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885130687). Fixed. The private import and both `data-*` role probes are gone, and the replacement asserts a real accessibility hierarchy through public role queries.

### Commands run

- `pnpm test app/src/sandbox/combobox-grid/test.ts`: 5 passed.
- `pnpm test combobox select`: 75 files, 494 tests passed.
- `pnpm tsc`: clean.
- `pnpm build`, then inspected `packages/ariakit-react-components/dist/combobox/combobox-context.d.ts`, then `pnpm clean`. Working tree restored.
- Independent red check: removing the store-ownership condition fails exactly one test, `preserves grid roles for explicit-store hooks under another popup`, and nothing else. This confirms the author's reported red check.
- Two further mutations of the ownership ternary, plus a cross-store item reproduction on both refs.
- All 48 checks pass on this head, with zero non-success check runs.

All experiments were reverted and the working tree is clean at this head. The branch was not modified or pushed.

### Findings

1. Fixture coverage regression for the no-provider fallback, published inline on `app/src/sandbox/combobox-grid/index.react.tsx`. Severity Low (verified), Priority P2, breakage likelihood Possible (verified by mutation), fix effort Easy.

### Validated candidates from other reviewers

- `useComboboxItem` does not check the list role's store, [thread r3885231946](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885231946) from CodeRabbit. Validation outcome: confirmed as real behavior, and confirmed pre-existing on `main`. A reproduction with an outer `menu` popup and an inner `listbox` popup yields `menuitem` instead of `option` identically on the merge base and on this head, so this PR neither introduces nor worsens it. Disposition: registered follow-up at https://github.com/ariakit/ariakit/issues/7316, since the item fix is a separate behavior change: `useComboboxItem` has no `contentElement` fallback, and adding one would change the no-provider case from `option` to the popup-derived role. Full evidence is on the thread and in the issue.

### Declined

- The pending-role case is no longer covered by any test, which I verified: a mutation that falls back while the same-store role is still measuring passes all 5 tests. Declined non-goal. This follows the tradeoff I endorsed last round, and the mutation it misses produces a single-commit role flash before `useAttribute` settles, with no lasting effect on the accessibility tree. Recorded here so the gap is on the record rather than implicit.

### Checked and clear

- Changeset is unchanged and still accurate. The cross-store handling restores released behavior rather than changing it: my nested reproduction returns `rowgroup` and `row` on both the merge base and this head, so the only user-visible change against the last release remains the nested same-store list case the entry describes.
- The description now matches the implementation, including the explicit statement that `ComboboxItem` keeps its existing nearest-list behavior. The mismatch I reported last round is gone.
- `useMemo` keeps the provider value stable, so no render churn is introduced, and the performance job reports no significant change.
- Test renames now carry the issue URL above both the browser test and the happy-dom duplicate.

### Not fully validated

I did not run Playwright locally. CI covers Chrome, Firefox, Safari, iOS, Android, and visual on this head, and all pass.

app/src/sandbox/combobox-grid/index.react.tsx
diegohaz: Fixed in [`36f98eac4`](https://github.com/ariakit/ariakit/commit/36f98eac4a415a6175394e8b39c6eca14de07653): restored a standalone explicit-store `HookGrid` beside the nested case, with distinct accessible input and popup labels. Added targeted coverage in the same commit in both Chrome and happy-dom.

The red check removed only the no-provider fallback while preserving the cross-store fallback. The new standalone test then failed with `group` and `presentation`, while the other five focused tests passed. After restoration, all six happy-dom tests, all six React 18 tests, and all three Chrome tests passed.

diegohaz:
**Fixed**

- [`36f98eac4`](https://github.com/ariakit/ariakit/commit/36f98eac4a415a6175394e8b39c6eca14de07653) ([reply](https://github.com/ariakit/ariakit/pull/7314#discussion_r3885403896)): restored standalone explicit-store fallback coverage in happy-dom and Chrome, with a mutation red check that fails only the restored case.

PR #7319, opened 2026-08-29
Fix ComboboxItem roles with explicit stores

### LINKED ISSUE #7316: ComboboxItem resolves its role from an outer combobox popup when given an explicit store
### Current behavior

`useComboboxItem` reads the popup role from the nearest `ComboboxListRoleContext` provider without checking which combobox store owns that provider (`packages/ariakit-react-components/src/combobox/combobox-item.tsx:145-146`):

```ts
const listRole = useContext(ComboboxListRoleContext);
const role = getItemRole(listRole?.role);
```

When the hook receives an explicit `store` and is rendered under another combobox's `ComboboxList` or `ComboboxPopover`, the outer provider wins. The item then advertises the outer popup's item role instead of the role its own popup implies.

`useComboboxItem` documents that explicit-store pattern in its own JSDoc:

```jsx
const store = useComboboxStore();
const props = useComboboxItem({ store, value: "value" });
<Role {...props} />
```

```tsx
import { useComboboxItem } from "@ariakit/react-components/combobox/combobox-item";

function InnerList() {
  const store = Ariakit.useComboboxStore({ defaultOpen: true });
  const itemProps = useComboboxItem({ store, value: "A" });

  return (
    <>
      <Ariakit.Combobox store={store} aria-label="Inner" />
      <Ariakit.ComboboxPopover store={store} role="listbox" aria-label="Inner list">
        <Ariakit.Role {...itemProps} data-testid="inner-item" />
      </Ariakit.ComboboxPopover>
    </>
  );
}

  return (
    <Ariakit.ComboboxProvider defaultOpen>
      <Ariakit.Combobox aria-label="Outer" />
      <Ariakit.ComboboxPopover role="menu" aria-label="Outer menu">
        <InnerList />
      </Ariakit.ComboboxPopover>
    </Ariakit.ComboboxProvider>
  );
}
```

The inner popup is a `listbox`, so its item should be an `option`. Measured with happy-dom:

```text
expected: option
actual:   menuitem
```

The outer `menu` role reaches the inner item through the nearest provider, so assistive technology reads a menu item inside a listbox.

### Pre-existing on `main`

Verified. The same fixture produces `menuitem` on both [`a8a5181`](https://github.com/ariakit/ariakit/commit/a8a51813d21b848dcc6240aefc9e5ec1aa4cece0), the merge base of #7314, and on [`5e4d7b6`](https://github.com/ariakit/ariakit/commit/5e4d7b636b5b4e0194ed512b0df00c7897d84641), the head of #7314.

Before #7314, `ComboboxListRoleContext` carried a bare role string, so the nearest provider already won regardless of store. #7314 rewrote that read as `listRole?.role` to match a new `{ store, role }` context shape, which resolves to the same value. `ComboboxItem` behavior is unchanged by that pull request.

### Why #7314 does not fix this

#7314 scopes `ComboboxGroup` and `ComboboxRow` to their own store and states that `ComboboxItem` keeps its existing nearest-list behavior. Applying the same ownership check to items is not a mechanical copy of that change:

1. `useComboboxItem` has no `contentElement` fallback today. Adding `getPopupRole(contentElement)` would change behavior for an item that has an explicit store and no provider above it at all. It resolves to `option` today, and would instead resolve to `menuitem` or `treeitem` from the popup role. That affects existing consumers and needs its own regression tests and changelog entry.
2. It would not help the grid case that motivated #7305, because `itemRoleByPopupRole` (`packages/ariakit-utils/src/dom.ts:366-370`) has no `grid` entry, so grid items still need an explicit `role="gridcell"`. #7302 lists that gap.

#7314 repairs a distinct regression for groups and rows and is complete without this change.

### Expected behavior

`ComboboxItem` should resolve its role from its own combobox, the same way `ComboboxGroup` and `ComboboxRow` do after #7314.

The fix must decide explicitly whether an item with an explicit store and no list-role provider above it keeps resolving to `option`, or starts reading the explicit store's `contentElement`. Both cases need coverage in `app/src/sandbox/**`.

### Workaround

Pass `role` explicitly on the item. The prop takes precedence over the computed role (`packages/ariakit-react-components/src/combobox/combobox-item.tsx:250-252`).

```tsx
<Ariakit.Role {...itemProps} role="option" />
```

[Confirmed workaround and triage evidence](https://github.com/ariakit/ariakit/issues/7316#issuecomment-5459978553)

[Workaround role mapping and limitations](https://github.com/ariakit/ariakit/issues/7316#issuecomment-5460388775)

### Context

Found by CodeRabbit during review of #7314 (https://github.com/ariakit/ariakit/pull/7314#discussion_r3885231946) and confirmed independently against both refs above.

diegohaz:
## Motivation

`useComboboxItem` can inherit an outer combobox popup role when it receives an explicit store. The rendered item then exposes a role that conflicts with its own popup.

Fixes #7316

## Solution

Resolve the role from `ComboboxListRoleContext` only when that context belongs to the item's store. Otherwise, derive the popup role from the explicit store's content element. The content element joins the existing store subscription, so this does not add a second subscription.

Add happy-dom, React 18, and Chrome regression coverage for an explicit store nested under another popup and for an explicit store without a provider. Add patch release notes for `@ariakit/react-components` and `@ariakit/react`.

## Workaround

Pass the item role explicitly after spreading the props returned by `useComboboxItem`.

```tsx
const itemProps = useComboboxItem({ store, value: "Inner item" });

{/* TODO: Remove after https://github.com/ariakit/ariakit/issues/7316 is fixed. */}
<Ariakit.Role {...itemProps} role="option" />
```

The explicit role must match the popup: use `option` for a listbox, `menuitem` for a menu, `treeitem` for a tree, or `gridcell` for a grid. This workaround preserves the reported composition and ARIA semantics, but consumers must keep the role synchronized if the popup role changes.

## Preview

No visual artifact is included because the broken and corrected states render identically. The difference exists only in the DOM and accessibility roles, which the browser assertions cover.

### DIFF
--- a/.changeset/300-combobox-item-store-role.md
+++ b/.changeset/300-combobox-item-store-role.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`ComboboxItem`](https://ariakit.com/reference/combobox-item) to resolve its automatic role from its own explicit combobox store.
--- a/app/src/sandbox/combobox-item-7316/index.react.tsx
+++ b/app/src/sandbox/combobox-item-7316/index.react.tsx
@@ -1,0 +1,52 @@
+import * as Ariakit from "@ariakit/react";
+import { useComboboxItem } from "@ariakit/react-components/combobox/combobox-item";
+
+function NestedListbox() {
+  const store = Ariakit.useComboboxStore({ defaultOpen: true });
+  const itemProps = useComboboxItem({ store, value: "Inner item" });
+
+  return (
+    <>
+      <Ariakit.Combobox store={store} aria-label="Inner combobox" />
+      <Ariakit.ComboboxPopover
+        store={store}
+        role="listbox"
+        aria-label="Inner list"
+      >
+        <Ariakit.Role {...itemProps} />
+      </Ariakit.ComboboxPopover>
+    </>
+  );
+}
+
+function StandaloneMenu() {
+  const store = Ariakit.useComboboxStore({ defaultOpen: true });
+  const itemProps = useComboboxItem({ store, value: "Standalone item" });
+
+  return (
+    <>
+      <Ariakit.Combobox store={store} aria-label="Standalone combobox" />
+      <Ariakit.ComboboxPopover
+        store={store}
+        role="menu"
+        aria-label="Standalone menu"
+      >
+        <Ariakit.Role {...itemProps} />
+      </Ariakit.ComboboxPopover>
+    </>
+  );
+}
+
+export default function Example() {
+  return (
+    <>
+      <Ariakit.ComboboxProvider defaultOpen>
+        <Ariakit.Combobox aria-label="Outer combobox" />
+        <Ariakit.ComboboxPopover role="menu" aria-label="Outer menu">
+          <NestedListbox />
+        </Ariakit.ComboboxPopover>
+      </Ariakit.ComboboxProvider>
+      <StandaloneMenu />
+    </>
+  );
+}
--- a/app/src/sandbox/combobox-item-7316/index.react.tsx
+++ b/app/src/sandbox/combobox-item-7316/index.react.tsx
@@ -13,7 +13,7 @@ function NestedListbox() {
         role="listbox"
         aria-label="Inner list"
       >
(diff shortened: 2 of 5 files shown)


diegohaz:
@codex

PR #7303, opened 2026-08-28
Fix ARIA issues in modal Menu

diegohaz:
## Motivation

When `Menu` is rendered with the `modal` prop, the popup element with `role="menu"` has two ARIA problems.

Ariakit prepends a visually hidden `<button>Dismiss popup</button>` to the popup as a fallback close affordance, so the accessibility tree exposes a plain `button` as an owned child of `role="menu"`. The ARIA menu pattern only allows `menuitem`, `menuitemcheckbox`, `menuitemradio`, `group`, and `separator` as owned elements. The label is also hardcoded in English, so it can't be localized.

The `aria-labelledby` attribute on the popup points to the `MenuButton`, which the modal dialog places inside an `inert` subtree while the menu is open. Chromium doesn't resolve `aria-labelledby` across that `inert` boundary when the referenced element is named by its own text content, so the menu ends up with an empty accessible name.

```jsx
<Ariakit.MenuProvider>
  <Ariakit.MenuButton>Preview</Ariakit.MenuButton>
  <Ariakit.Menu modal>
    <Ariakit.MenuItem>Desktop</Ariakit.MenuItem>
    <Ariakit.MenuItem>Mobile</Ariakit.MenuItem>
  </Ariakit.Menu>
</Ariakit.MenuProvider>
```

Reading Chromium's own accessibility tree for the element with `role="menu"` above shows both problems at once:

```json
{
  "name": "",
  "nameSources": [
    { "type": "relatedElement", "attribute": "aria-labelledby", "invalid": true }
  ],
  "children": [
    { "role": "button", "name": "Dismiss popup" },
    { "role": "menuitem", "name": "Desktop" },
    { "role": "menuitem", "name": "Mobile" }
  ]
}
```

Fixes #4270
Fixes #7310

## Solution

A modal `Menu` now adds its `MenuButton` to the modal context, the same way [`ComboboxPopover`](https://ariakit.com/reference/combobox-popover) already does for the combobox input and its controls. With the button outside the `inert` subtree, Chromium resolves the menu's `aria-labelledby` target again and the accessible name comes back.

Only a disclosure the application named counts. `Dialog` also captures whatever element happened to have focus as a fallback opener, and that element can't be assumed to close the menu, so it stays out of the modal context.

[...]

PR #7333, opened 2026-08-30
Fix Dialog fallback dismiss ownership and staleness

### LINKED ISSUE #7321: Dialog skips its fallback dismiss when a nested popup renders a dismiss button
## Context

Found while reviewing #7303, which moves the fallback dismiss button from inside the dialog element to next to it. The detection that decides whether to render that button was rewritten in that PR, but the predicate itself was carried over unchanged, so this is not caused by #7303.

## Problem

`useDialog` decides whether a modal dialog needs its visually hidden fallback dismiss button with a single descendant query:

```ts
// packages/ariakit-react-components/src/dialog/dialog.tsx
return !dialog.querySelector("[data-dialog-dismiss]");
```

`data-dialog-dismiss` is set by every `DialogDismiss`, including the ones that `PopoverDismiss`, `MenuDismiss`, `HovercardDismiss` and `ComboboxDismiss` render. Any of those inside the dialog's DOM subtree satisfies the check, even when the button closes a different store.

This does not need an explicit `portal={false}`. A nested `Popover` is non-modal by default, `portal` defaults to `modal`, so its content renders inline inside the parent dialog. The nested content is also rendered while closed, since `unmountOnHide` is `false` by default, so a popover the user never opened is enough.

The result is a modal dialog with no dismiss control of its own and no fallback, which is exactly the state the fallback exists to prevent. A screen reader user on a touch device, who can't press Escape and can't click outside, has no way to leave it.

## Reproduction

```tsx
<Ariakit.DialogProvider>
  <Ariakit.DialogDisclosure>Compose</Ariakit.DialogDisclosure>
  <Ariakit.Dialog>
    <Ariakit.DialogHeading>Compose</Ariakit.DialogHeading>
    <Ariakit.PopoverProvider>
      <Ariakit.PopoverDisclosure>Formatting</Ariakit.PopoverDisclosure>
      <Ariakit.Popover>
        <Ariakit.PopoverHeading>Formatting</Ariakit.PopoverHeading>
        <Ariakit.PopoverDismiss>Close</Ariakit.PopoverDismiss>
      </Ariakit.Popover>
    </Ariakit.PopoverProvider>
  </Ariakit.Dialog>
</Ariakit.DialogProvider>
```

Open the "Compose" dialog without touching the popover, then look for a button named "Dismiss popup". There is none. Removing the nested `PopoverDismiss` brings it back.

## Verified pre-existing

Confirmed on `main` rather than only on the PR branch. In a worktree at the #7303 head, restoring the package source to the merge base `e5c8fbd48` and re-running the same fixture produces the same result:

| Fixture | `main` source (`e5c8fbd48`) | #7303 head (`e0c92377`) |
| --- | --- | --- |
| Dialog with no nested dismiss | fallback present | fallback present |
| Dialog containing a closed `Popover` with `PopoverDismiss` | fallback absent | fallback absent |

So #7303 can merge without this. It neither introduces nor worsens the behavior, and the button it renders is correct wherever the predicate says one is needed.

## Related second defect in the same predicate

The check runs in an effect whose dependencies are `modal`, `unstable_hiddenDismiss`, `open` and `domReady`. None of them changes when a `DialogDismiss` mounts or unmounts while the dialog stays open. So a dialog that renders its dismiss control conditionally, after loading for example, keeps whichever answer was correct at open time: no fallback when the control arrives late, or two controls both named "Dismiss popup" when it arrives after the fallback was added.

Both defects come from the same place, so they are worth fixing together.

## Expected outcome

Count only dismiss buttons that belong to this dialog. Matching on the nearest `[data-dialog]` ancestor is the straightforward form:

```ts
const dismisses = dialog.querySelectorAll("[data-dialog-dismiss]");
const ownDismiss = [...dismisses].some(
  (dismiss) => dismiss.closest("[data-dialog]") === dialog,
);
```

Then decide how the answer should track a dismiss control that appears or disappears while the dialog is open, rather than being fixed at open time.

Cover both with a regression test. `app/src/sandbox/dialog-hidden-dismiss/` is the fixture that owns this behavior after #7303, and it already has both a browser test and a happy-dom duplicate.

diegohaz:
Fixes #7321

## Motivation

A modal [`Dialog`](https://ariakit.com/reference/dialog) with no dismiss control of its own renders a visually hidden fallback dismiss button, so that assistive technology users who cannot press Escape or click outside are not trapped in it. Whether it needs one was decided by a single descendant query:

```ts
return !dialog.querySelector("[data-dialog-dismiss]");
```

`data-dialog-dismiss` is set by every [`DialogDismiss`](https://ariakit.com/reference/dialog-dismiss), including the ones that [`PopoverDismiss`](https://ariakit.com/reference/popover-dismiss), [`MenuDismiss`](https://ariakit.com/reference/menu-dismiss), `HovercardDismiss` and `ComboboxDismiss` render. Any of them anywhere in the dialog's subtree satisfied the check, even though it closes a different store. Two defects followed.

**A nested popup's dismiss suppressed the fallback.** This needs no `portal={false}`. A nested [`Popover`](https://ariakit.com/reference/popover) is non-modal by default and `portal` defaults to `modal`, so its content renders inline inside the dialog. It also stays rendered while closed, since `unmountOnHide` defaults to `false`. So a popover the user never opened was enough to leave the dialog with no dismiss control and no fallback, which is exactly the state the fallback exists to prevent.

```tsx
<Ariakit.Dialog>
  <Ariakit.DialogHeading>Compose</Ariakit.DialogHeading>
  <Ariakit.PopoverProvider>
    <Ariakit.PopoverDisclosure>Formatting</Ariakit.PopoverDisclosure>
    <Ariakit.Popover>
      {/* This closes the popover, but the dialog counted it as its own. */}
      <Ariakit.PopoverDismiss>Close formatting</Ariakit.PopoverDismiss>
    </Ariakit.Popover>
  </Ariakit.PopoverProvider>
</Ariakit.Dialog>
```

The same composition reaches a modal [`Menu`](https://ariakit.com/reference/menu) through its submenu, which is never modal and so renders inline inside its parent.

**The answer was frozen at open time.** The effect that computed it depended on `modal`, `open` and `domReady`, none of which changes when a [`DialogDismiss`](https://ariakit.com/reference/dialog-dismiss) mounts or unmounts while the dialog stays open. A dialog that renders its dismiss control conditionally therefore kept whichever answer was correct when it opened: no fallback when the control arrived late, or two dismiss controls when it arrived after the fallback was added.

## Solution

Three parts, in one commit.

**Ownership.** A new internal `hasOwnDismiss` accepts a dismiss button only when the nearest `[data-dialog]` ancestor is this dialog, so a button belonging to a popup nested inside it no longer counts.

```ts
function hasOwnDismiss(dialog: Element) {
  const dismisses = dialog.querySelectorAll("[data-dialog-dismiss]");
  for (const dismiss of dismisses) {
    if (dismiss.closest("[data-dialog]") === dialog) return true;
  }
  return false;
}
```

**Staleness.** The same effect now attaches a `MutationObserver` to the dialog, so the answer keeps up with a dismiss control that mounts or unmounts while the dialog stays open. It reads the store's `contentElement` rather than the ref, so the observer follows the element that is in the document. The fallback renders next to the dialog rather than inside it, so adding it cannot retrigger the observer.

**A stable wrapper, which is what makes the second part safe.** The wrapper that renders the fallback used to return the bare element when none was needed and a fragment when one was. Toggling that changed the returned tree's shape, so React reconciled the dialog against a different child index and remounted it along with everything the application had rendered inside it. With the observer in place, that turned a mid-open dismiss mount into a loop: the remount reset the application state that had just mounted the dismiss, which flipped the answer straight back. The wrapper is now always rendered and only the button toggles.

```tsx
<>
  {needsHiddenDismiss && <button ... />}
  {element}
</>
```

That third part also removes a remount that happened on `main` whenever the fallback appeared. The same hazard remains in the neighbouring backdrop wrapper and in `usePortal`; it is out of scope here and registered as #7335.

## Workaround

Until the fix is released, give the affected popup a dismiss control of its own. It can be visually hidden, so it does not change the design, and it satisfies the predicate the same way a visible one would.

```tsx
<Ariakit.Dialog>
  <Ariakit.DialogHeading>Compose</Ariakit.DialogHeading>
  {/* TODO: Remove once the fix lands.
      https://github.com/ariakit/ariakit/issues/7321 */}
  <Ariakit.VisuallyHidden render={<Ariakit.DialogDismiss tabIndex={-1} />} />
  <Ariakit.PopoverProvider>
    <Ariakit.PopoverDisclosure>Formatting</Ariakit.PopoverDisclosure>
    <Ariakit.Popover>
      <Ariakit.PopoverHeading>Formatting</Ariakit.PopoverHeading>
      <Ariakit.PopoverDismiss>Close formatting</Ariakit.PopoverDismiss>
    </Ariakit.Popover>
  </Ariakit.PopoverProvider>
</Ariakit.Dialog>
```

Assumptions and limitations, none of which apply to the fix itself:

- **It has to be a child of the popup that lacks a dismiss control, not of the nested popup.** A dismiss control placed inside the nested popup closes that popup instead, which is the defect.
- **`tabIndex={-1}` is required.** Without it the stand-in becomes a tab stop and takes the dialog's initial focus, since it is the first tabbable element. The fallback it replaces is not reachable with Tab.
- **For a conditionally rendered dismiss control, the stand-in must render in the exact complement of the branch that renders the real one.** The predicate ran once per open before this fix, so a stand-in present when the popup opened suppressed the fallback for that whole open cycle. Mount it too narrowly and the popup ends up with no dismiss control at all; mount it too broadly and users get two.
- **For a modal [`Menu`](https://ariakit.com/reference/menu) the stand-in must go inside the menu**, because a modal menu makes everything outside it inert except its own menu button. That returns a `button` to the menu's owned elements, which #7303 had removed. There is no userland placement that avoids this.

## Tests

`app/src/sandbox/dialog-hidden-dismiss/` gains five scenarios, each with a browser test and a happy-dom duplicate:

| Scenario | Covers | On `main` |
| --- | --- | --- |
| `Compose` | dialog holding a closed `Popover` with a `PopoverDismiss` | fails |
| `Tools` | modal `Menu` whose submenu renders a `MenuDismiss` | fails |
| `Activity` | dismiss control mounts, below a wrapper, while the dialog stays open | fails |
| `Remove photo` | dismiss control unmounts, among the dialog's own children, while it stays open | fails |
| `Share` | dialog that *does* own a dismiss, after a nested popup's in document order and inside a wrapper | passes |

`Share` passes on `main`, so it is not reproduction coverage. It is a guard against over-correction: it fails for a fix that only inspects the first `[data-dialog-dismiss]` match, or only the dialog's direct children.

The commits are staged so that CI recorded the evidence. The reproduction commit was red in `Main / Test`, `Main / Test React 18` and `App / Test Chrome`, each on the missing `Dismiss popup` fallback. The workaround commit turned all of them green from userland alone. This commit reverts the workaround and turns them green from the library.

Known coverage gap, accepted: the happy-dom duplicate settles the observer through `click()`, which drains it via `sleep()` and `flushScheduler`, so its assertions are exact. Playwright has no equivalent settle step, so the two `test-chrome.ts` guards that assert the fallback is *still* present after a nested popup opens could pass before a hypothetical future regression that deferred the recomputation to a frame or timer. The primary red proof is unaffected, because it is the first fallback assertion in each test, taken before any nested popup opens.

## Review gate reassessment

<details>
<summary>Cycle 3, reproduction commit: Continue the current design</summary>

**Assessment**

Issue severity is high for the affected users: a modal dialog with no dismiss control reachable by assistive technology is a trap, not a degradation. The supported scope is ordinary public composition, with no experimental surface and no opt-in required, and the expected frequency is moderate and rising, since #7303 removed the `unstable_hiddenDismiss` opt-out.

Complexity added by that stage was fixture and test only: five scenarios, two local components with one `useState` each, five tests duplicated across the two suites. No branches, state, helpers, abstractions, public contract or platform policy.

The findings across three gate rounds split into two species. "A wrong fix would still pass" ran 3, then 1, then 0, and converged. Comment accuracy and burden ran 1, then 2, then 1, and did not. The red set was unchanged since the first reviewed state, so three rounds produced one passing guard test, one wrapper element and comment edits. The direction itself was never in question.

Each scenario was checked for removal and none earned it: `Compose` is the issue's own reproduction, `Tools` is the separately reported instance, `Activity` and `Remove photo` are the two directions of the staleness defect that the issue asks to fix together, and `Share` is the only scenario where a dialog owns a dismiss and also contains a nested popup's.

**Decision**

Continue, with a bounded prose simplification and no scenario, assertion or test dropped. Adopted guardrail: a fixture comment states what the composition is and what the user-visible consequence is; it does not state what other modules do, and it does not enumerate implementations that would pass.

**Intentionally unsupported**

The ownership and staleness cross-product, meaning a dismiss the dialog does *not* own mounting or unmounting while the dialog stays open. Both axes are handled by independent mechanisms in any plausible fix, the composition is uncommon, and it could not be proven red at that stage.

</details>

<details>
<summary>Cycle 3, library fix commit: Continue the current design</summary>

**Assessment**

The fix is one new 11-line internal module, a few net lines inside one existing effect, and one wrapper reshaped. No public API change, no new prop, no store field, no platform policy, no new dependency. One new runtime resource, a `MutationObserver` scoped to open modal dialogs only, following the existing precedent in `combobox-list.tsx` and stricter than it, since the constructor comes from `getWindow(contentElement)` per the DOM realms rule.

Findings against the fix logic converged immediately, 2 then 0 then 0, and the three-part shape survived two consecutive rounds untouched. Every finding in rounds 2 and 3 was text. Across both gates the prose species ran 1, 2, 1 then 1, 1, 2 without converging.

The first reassessment's guardrail did not fail; its scope did not follow the work. It governed fixture comments, while rounds 2 and 3 moved to library comments, the changeset and this description.

**Decision**

Continue, keeping all three parts, all five scenarios and every assertion. Splitting the staleness half into a follow-up was considered and rejected: the reproduction commit is already pushed and CI recorded all four tests red, so shipping only the ownership half would leave committed, CI-witnessed red coverage that would have to be deleted or skipped.

Apply a bounded prose reduction instead, and extend the guardrail to every prose artifact in the commit: state the local constraint and the user-visible consequence, link the most specific permalink for the rest, and do not narrate React or DOM mechanics that the code and the file's existing comments already carry. Comment lines in the new code went from 13 to 8, with two comment-accuracy findings resolved by deletion rather than a fourth rewording.

**Intentionally unsupported**

The ownership and staleness cross-product, carried forward: the chosen fix keeps the two mechanisms independent, which is the condition under which the first reassessment said not to revisit it. A dismiss control the application portals outside the dialog element, which is correctly not an escape route from inside an inert-protected region. A dialog whose `data-dialog` marker is stripped by user props, which already breaks other internals that depend on the same marker.

**Registered follow-up**

#7335, the same shape-toggle remount hazard in the backdrop wrapper and in `usePortal`, verified pre-existing on `main` in an isolated worktree.

</details>


.changeset/300-dialog-own-dismiss.md
```diff
@@ -0,0 +1,25 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
```
diegohaz, line 5: This should be a single-sentence changeset.

packages/ariakit-react-components/src/dialog/utils/__own-dismiss.ts
```diff
@@ -0,0 +1,11 @@
+// PopoverDismiss, MenuDismiss and friends all render DialogDismiss, and a
+// nested popup renders inline and stays rendered while closed, so its dismiss
+// sits in this dialog's subtree even though it closes a different popup.
+// https://github.com/ariakit/ariakit/issues/7321
+export function hasOwnDismiss(dialog: Element) {
```
diegohaz: Why do we need a separate file for this function? Shouldn't this be on the `dialog.tsx` file instead?

.changeset/300-dialog-own-dismiss.md
diegohaz: Fixed in [`8fa04ec`](https://github.com/ariakit/ariakit/commit/8fa04ec4f2a295d985d184ffb1e53e3e43551834): rewrote it as one `Fixed ...` sentence covering both defects, matching the other bug fix entries. The multi-paragraph form with an example is what the changeset guidance reserves for features that add or change public APIs, and this one changes no API.

packages/ariakit-react-components/src/dialog/utils/__own-dismiss.ts
diegohaz: Fixed in [`e994bea`](https://github.com/ariakit/ariakit/commit/e994bea24e698f850d5dd218a977a323d1cb5372): moved it into `dialog.tsx` beside the other module scope helpers and deleted the file. It had a single call site, in `dialog.tsx` itself.

For the record on why the neighbouring `__is-hidden-dismiss.ts` stays separate: it has four call sites, and two of them (`utils/disable-tree.ts` and `utils/use-hide-on-interact-outside.ts`) are imported by `dialog.tsx`, so inlining that one would create an import cycle. No such constraint applied here.

diegohaz:
**Fixed**

- [`8fa04ec`](https://github.com/ariakit/ariakit/commit/8fa04ec4f2a295d985d184ffb1e53e3e43551834) ([reply](https://github.com/ariakit/ariakit/pull/7333#discussion_r3888971696)): rewrote the changeset as a single `Fixed ...` sentence covering both defects, matching the other bug fix entries.
- [`e994bea`](https://github.com/ariakit/ariakit/commit/e994bea24e698f850d5dd218a977a323d1cb5372) ([reply](https://github.com/ariakit/ariakit/pull/7333#discussion_r3888971877)): moved `hasOwnDismiss` into `dialog.tsx` beside the other module scope helpers and deleted `utils/__own-dismiss.ts`. It had a single call site, in `dialog.tsx` itself.

Neither comment reported a behavior change, so no test changes were needed. Verified on the pushed head: `pnpm test` 1946 passed, the sandbox suite 10 passed on both React versions, `pnpm test-chrome dialog-hidden-dismiss` 11 passed, with `pnpm tsc` and `pnpm lint` clean.

The description's `Solution` snippet was updated to drop the now incorrect `export` keyword. The dated entries under `Review gate reassessment` are left as they were, since they record decisions made at the time.

Issue #7343, opened 2026-08-30
Toggling the portal prop remounts an open Dialog or Popover and everything inside it


diegohaz:
## Context

Found while fixing #7335, which repairs the same reconciliation hazard in the neighbouring backdrop wrapper in `packages/ariakit-react-components/src/dialog/dialog.tsx`. #7335 explicitly defers this one, because `usePortal` needs its own analysis rather than the same one-line treatment.

## Problem

The wrapper in `packages/ariakit-react-components/src/portal/portal.tsx` has three structurally different returns for the same element:

```tsx
(element) => {
  element = (
    <PortalContext.Provider value={portalNode || context}>
      {element}
    </PortalContext.Provider>
  );

  if (!portal) return element;

  if (!portalNode) {
    return <span ref={refProp} id={props.id} style={{ position: "fixed" }} hidden />;
  }

  element = (
    <>
      {preserveTabOrder && <FocusTrap ... />}
      {element}
      {preserveTabOrder && <FocusTrap ... />}
    </>
  );

  element = createPortal(element, portalNode);
  // ...
};
```

Crossing the `portal` boundary swaps a provider element for a portal, so React deletes the fiber and creates a new one. The element and its whole subtree remount, and whatever the application rendered inside it is discarded.

`portal` defaults to `modal` in both `useDialog` and `usePopover`, so a dynamic `modal` flips `portal` too and remounts for this reason as well. `examples/popover-responsive` toggles `modal` and `backdrop` together on a media query, which is exactly the shape a consumer reaches for when a popover becomes a bottom sheet on small screens.

## Why #7335 can ship without this

#7335 fixes the backdrop wrapper alone. Its sandbox holds `modal` and `portal` constant per dialog, so its regression coverage isolates the backdrop wrapper and does not depend on this. The two are independent remount sources; repairing one does not repair or worsen the other.

## Verified pre-existing

The shape toggle is present on `main` at [`c806c1fb1`](https://github.com/ariakit/ariakit/commit/c806c1fb1586effa8cdc92d1ae33fe9892b00b03), the current base at the time of writing, and the #7335 branch does not touch `portal.tsx`.

## Expected outcome

Decide whether the element can keep a stable position across all three returns, and cover it with a regression test that keeps a popup open across a `portal` toggle and asserts that state inside it survives. The `preserveTabOrder` focus traps are also conditional siblings rendered before the element inside the portal, so they need the same consideration as the backdrop did.

diegohaz:
[...]

Issue #7335, opened 2026-08-30
Toggling the backdrop prop remounts an open Dialog and everything inside it


diegohaz:
## Context

Found by the pre-commit review of #7333, which fixes the same reconciliation hazard in the neighbouring hidden dismiss wrapper. That PR changes only the hidden dismiss wrapper, so this one is left as it is.

## Problem

The backdrop wrapper in `packages/ariakit-react-components/src/dialog/dialog.tsx` returns the bare element when `backdrop` is falsy, and a fragment when it is truthy:

```tsx
(element) => {
  if (!backdrop) return element;
  return (
    <>
      <DialogBackdrop ... />
      {element}
    </>
  );
},
```

React unwraps the unkeyed fragment, so the two branches hand the parent different children. Crossing the falsy/truthy boundary on an open dialog therefore unmounts and remounts the dialog element and its whole subtree, discarding whatever the application rendered in it.

This originally described the parent's children going from `[dialog]` to `[backdrop, dialog]`, with index 0 changing type. That was accurate when this was filed. #7333 landed afterwards and made the neighbouring hidden dismiss wrapper always return a fragment, so the children are now `[hidden dismiss slot, dialog]` and `[backdrop, <>[hidden dismiss slot, dialog]</>]`. The dialog sits at the same index in both, and what changes is that slot's element type. The remount is the same; only the mechanism differs.

The deciding factor is what ends up in the element's own slot. A conditional sibling rendered *after* a concrete `{element}` leaves that slot alone, which is why the hidden inputs in `packages/ariakit-react-components/src/combobox/combobox.tsx` are safe. Placed before it, the slot stays stable only while both branches return the same shape.

## Verified pre-existing

Verified on `origin/main` at `0b991e8a08911fba70524d6dd53346106e69a08d`, in an isolated worktree, with a temporary sandbox and happy-dom test that were not committed.

The fixture keeps a `Dialog` open and holds `modal` and `portal` constant, changing only `backdrop` between an element and `false`. It contains an uncontrolled `<input>` the test types into and a child component with its own `useState` counter.

| Fixture | Same dialog node | Uncontrolled input value | Child state |
| --- | --- | --- | --- |
| `modal portal`, backdrop toggled off | no | lost | reset |
| `modal portal`, backdrop toggled on | no | lost | reset |
| `modal={false} portal={false}`, backdrop toggled off | no | lost | reset |
| `modal portal`, one truthy backdrop swapped for another | yes | preserved | preserved |

[...]

PR #7352, opened 2026-08-30
Fix Select and ComboboxSelect remounts when name changes

### LINKED ISSUE #7346: Toggling the name prop remounts Select and Combobox popovers and everything inside them
## Context

Found by the pre-commit review of #7342, which fixes the same reconciliation hazard in `Dialog`'s backdrop wrapper. #7335 enumerated the family and cleared `combobox.tsx`, but did not cover these two.

## Problem

`SelectPopover`'s hidden native select in `packages/ariakit-react-components/src/select/select.tsx` renders as a sibling **before** the element, and only while `name` is truthy:

```tsx
props = useWrapElement(
  props,
  (element) => {
    if (!name) return element;
    return (
      <>
        <select /* visually hidden autofill mirror */ />
        {element}
      </>
    );
  },
  // ...
);
```

`packages/ariakit-react-components/src/combobox/combobox-select.tsx` has the identical shape.

This is the wrapper shape #7333 and #7342 fixed in `packages/ariakit-react-components/src/dialog/dialog.tsx`. Returning the bare element in one branch and an unkeyed fragment in the other changes what occupies the element's own slot, so crossing the falsy/truthy boundary makes React delete the element's fiber and rebuild it, discarding whatever the application rendered inside.

Ordering is what decides it. The hidden inputs in `packages/ariakit-react-components/src/combobox/combobox.tsx` render **after** `{element}`, which is why they are safe. Rendered before it, they are not.

## Impact

Lower than #7335. It needs a consumer that toggles `name` between falsy and truthy on a mounted `Select` or `Combobox`, which is a less natural thing to do than toggling a backdrop. A form that only assigns `name` once a field becomes part of a submitted payload would reach it.

## Verified pre-existing

Both wrappers have this shape on `main` at [`74f979289`](https://github.com/ariakit/ariakit/commit/74f9792898dfd35e4494bb98f34e183cfa35463a), and #7342 does not touch either file.

## Expected outcome

Decide whether to apply the same treatment as #7333 and #7342: always return the fragment and toggle the hidden native select inside a fixed slot, so gaining or losing it only mounts or unmounts that control. Cover it with a regression test that toggles `name` on a mounted `Select` and asserts that state inside the popover survives.

Note that `packages/ariakit-react-components/src/combobox/combobox-cancel.tsx` looks similar but is not the same hazard: there the wrapper's own element is what mounts and unmounts, which is the intent.

diegohaz:
## Motivation

Changing `name` between empty and nonempty remounts the rendered subtree of `Select` and `ComboboxSelect`. Local state and uncontrolled input values are lost. This also affects a popover composed inside the control through a portal. A popover rendered as a separate sibling is outside this subtree.

## Solution

Keep the wrapper fragment stable and render the hidden native select in a fixed conditional slot before the control. Changing `name` now mounts or unmounts only the native select, while preserving the control and its descendants.

```tsx
<>
  {!!name && <select /* native form and autofill control */ />}
  {element}
</>
```

The fix retains native form submission and autofill behavior. It adds no DOM wrapper, public API, or extra native select when `name` is empty.

## Tests

Four regressions cover both controls, initially named and unnamed. Each toggles `name` in both directions and checks an uncontrolled draft, child component state, native-select presence, and form submission. The same cases run in Chromium, React 19, and React 18.

Two additional proof cases keep `name` nonempty (`"fruit"` to `"produce"`) and check the native field name, draft, child state, and submission. All six cases pass locally in Chromium, React 19, and React 18. The two new cases also pass with the pre-fix source and fail at the draft assertion when the wrapper is temporarily keyed by `name`. Lint and type checks pass.

Local verification passed: `pnpm test` (1,956 tests), `pnpm test-react18` (1,078 tests), `pnpm tsc`, `pnpm lint-fix`, `pnpm build`, and 34 targeted browser checks across Chromium, Firefox, and Safari.

The reproduction, userland workaround, and library fix are separate commits. The final fixture uses the library fix without the workaround. CI recorded the four expected reproduction failures in `Main / Test`, `Main / Test React 18`, and `App / Test Chrome`. The workaround then passed those complete suites (1,956, 1,078, and 854 tests respectively).

Fixes #7346

## Workaround

Until the fix is released, leave `name` off the Ariakit control and submit its selected value through a separate hidden input. The control then keeps the same wrapper shape. The sandbox verifies this workaround without a library change.

```tsx
const combobox = Ariakit.useComboboxStore({ defaultSelectedValue: "Apple" });
const value = Ariakit.useStoreState(combobox, "selectedValue");

// TODO: Remove after upgrading to the fixed version.
// https://github.com/ariakit/ariakit/issues/7346
<input type="hidden" name={included ? "fruit" : undefined} value={value} />
<Ariakit.ComboboxSelect store={combobox} />
```

For `Select`, use `useSelectStore` and its `value` state. This example covers a single selected value and an enabled control inside a form. For multiple values, render one hidden input per value. Forward `form` and `disabled` to those inputs when needed. Hidden inputs do not provide the native select's autofill or `required` validation behavior; handle validation separately if the application needs it. The library fix retains those native-select features.

## Preview

Local before-and-after recordings show the draft clearing before the fix and surviving both name transitions after it. They are not attached yet because the browser upload requires explicit permission.

### DIFF
--- a/app/src/sandbox/select-combobox-name-7346/test.ts
+++ b/app/src/sandbox/select-combobox-name-7346/test.ts
@@ -38,3 +38,31 @@ for (const label of [
     }
   });
 }
+
+for (const label of ["Select named", "Combobox named"]) {
+  // Keeps name nonempty, so this passes before the fix too. It distinguishes
+  // a rename from the empty/nonempty boundary that caused the remount.
+  // https://github.com/ariakit/ariakit/issues/7346
+  test(`${label} keeps popup state when name stays nonempty`, async () => {
+    await click(q.combobox(label));
+    const dialog = q.dialog(label);
+    expect(dialog).toBeVisible();
+    await type("Buy ripe fruit", q.textbox("Note"));
+    await click(q.button("Add attachment"));
+    expect(q.form(label).querySelector("select")).toHaveAttribute(
+      "name",
+      "fruit",
+    );
+
+    await click(q.button(`Rename ${label}`));
+    expect(q.form(label).querySelector("select")).toHaveAttribute(
+      "name",
+      "produce",
+    );
+    expect(q.textbox("Note")).toHaveValue("Buy ripe fruit");
+    expect(q.status("Attachments")).toHaveTextContent("Attachments: 1");
+    expect(q.dialog(label)).toBe(dialog);
+    await click(q.button(`Submit ${label}`));
+    expect(q.status(`${label} submission`)).toHaveTextContent("Apple");
+  });
+}
--- a/app/src/sandbox/select-combobox-name-7346/test.ts
+++ b/app/src/sandbox/select-combobox-name-7346/test.ts
@@ -28,6 +28,9 @@ for (const label of [
       expect(q.textbox("Note")).toHaveValue("Buy ripe fruit");
       expect(q.status("Attachments")).toHaveTextContent("Attachments: 1");
       expect(q.dialog(label)).toBe(dialog);
+      expect(q.form(label).querySelectorAll("select")).toHaveLength(
+        included ? 1 : 0,
+      );
       await click(q.button(`Submit ${label}`));
       expect(q.status(`${label} submission`)).toHaveTextContent(
         included ? "Apple" : "Omitted",
--- a/app/src/sandbox/select-combobox-name-7346/test.ts
+++ b/app/src/sandbox/select-combobox-name-7346/test.ts
@@ -1,0 +1,37 @@
+import { click, q, type } from "@ariakit/test";
+import { expect, test } from "vitest";
+
+for (const label of [
+  "Select unnamed",
+  "Select named",
+  "Combobox unnamed",
+  "Combobox named",
+]) {
+  // https://github.com/ariakit/ariakit/issues/7346
+  test(`${label} keeps popup state when name changes`, async () => {
+    await click(q.combobox(label));
+    const dialog = q.dialog(label);
+    expect(dialog).toBeVisible();
+    await type("Buy ripe fruit", q.textbox("Note"));
+    await click(q.button("Add attachment"));
+    expect(q.status("Attachments")).toHaveTextContent("Attachments: 1");
+
+    for (const included of [
+      label.endsWith(" unnamed"),
+      !label.endsWith(" unnamed"),
+    ]) {
+      await click(q.checkbox(`Include ${label}`));
+      expect(q.checkbox(`Include ${label}`)).toHaveProperty(
+        "checked",
+        included,
+      );
+      expect(q.textbox("Note")).toHaveValue("Buy ripe fruit");
+      expect(q.status("Attachments")).toHaveTextContent("Attachments: 1");
+      expect(q.dialog(label)).toBe(dialog);
+      await click(q.button(`Submit ${label}`));
+      expect(q.status(`${label} submission`)).toHaveTextContent(
+        included ? "Apple" : "Omitted",
+      );
+    }
+  });
+}
--- a/packages/ariakit-react-components/src/combobox/combobox-select.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-select.tsx
@@ -201,7 +201,10 @@ export const useComboboxSelect = createHook<TagName, ComboboxSelectOptions>(
     // update the combobox store.
     props = useWrapElement(
       props,
-      // Keep the control in the same slot when the native select toggles.
+      // The native select toggles inside a wrapper that's always rendered.
+      // Returning the bare element in one branch and a fragment in the other
+      // would remount the control and everything inside it.
+      // https://github.com/ariakit/ariakit/issues/7346
       (element) => (
         <>
           {!!name && (
--- a/packages/ariakit-react-components/src/combobox/combobox-select.tsx
+++ b/packages/ariakit-react-components/src/combobox/combobox-select.tsx
@@ -201,10 +201,10 @@ export const useComboboxSelect = createHook<TagName, ComboboxSelectOptions>(
     // update the combobox store.
     props = useWrapElement(
       props,
-      (element) => {
-        if (!name) return element;
-        return (
-          <>
+      // Keep the control in the same slot when the native select toggles.
+      (element) => (
+        <>
+          {!!name && (
             <select
               style={getVisuallyHiddenStyle()}
               tabIndex={-1}
@@ -246,10 +246,10 @@ export const useComboboxSelect = createHook<TagName, ComboboxSelectOptions>(
                 </option>
               ))}
             </select>
-            {element}
-          </>
-        );
-      },
+          )}
+          {element}
+        </>
+      ),
       [
         store,
         label,
--- a/packages/ariakit-react-components/src/select/select.tsx
+++ b/packages/ariakit-react-components/src/select/select.tsx
@@ -198,10 +198,10 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
   // onChange event is triggered and we set the autofill state to true.
   props = useWrapElement(
     props,
-    (element) => {
-      if (!name) return element;
-      return (
-        <>
+    // Keep the control in the same slot when the native select toggles.
+    (element) => (
+      <>
+        {!!name && (
           <select
             style={getVisuallyHiddenStyle()}
             tabIndex={-1}
@@ -245,10 +245,10 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
               </option>
             ))}
           </select>
-          {element}
-        </>
-      );
-    },
+        )}
+        {element}
+      </>
+    ),
     [
       store,
       label,
--- a/packages/ariakit-react-components/src/select/select.tsx
+++ b/packages/ariakit-react-components/src/select/select.tsx
@@ -198,7 +198,10 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
   // onChange event is triggered and we set the autofill state to true.
   props = useWrapElement(
     props,
-    // Keep the control in the same slot when the native select toggles.
+    // The native select toggles inside a wrapper that's always rendered.
+    // Returning the bare element in one branch and a fragment in the other
+    // would remount the control and everything inside it.
+    // https://github.com/ariakit/ariakit/issues/7346
     (element) => (
       <>
         {!!name && (
--- a/.changeset/300-select-name-remount.md
+++ b/.changeset/300-select-name-remount.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+Fixed [`Select`](https://ariakit.com/reference/select) and [`ComboboxSelect`](https://ariakit.com/reference/combobox-select) losing descendant state when the `name` prop changes between empty and nonempty.
--- a/app/src/sandbox/select-combobox-name-7346/index.react.tsx
+++ b/app/src/sandbox/select-combobox-name-7346/index.react.tsx
@@ -32,9 +32,6 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
   const [submitted, setSubmitted] = useState("Not submitted");
   const label = `${combobox ? "Combobox" : "Select"} ${initiallyNamed ? "named" : "unnamed"}`;
   const name = included ? "fruit" : undefined;
-  const selectValue = Ariakit.useStoreState(select, "value");
-  const comboboxValue = Ariakit.useStoreState(comboboxStore, "selectedValue");
-  const value = combobox ? comboboxValue : selectValue;
   const popupProps = {
     "aria-label": label,
     role: "dialog",
@@ -65,12 +62,9 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
         />
         Include {label}
       </label>
-      {/* TODO: Remove after upgrading to the fixed version.
-      https://github.com/ariakit/ariakit/issues/7346 */}
-      <input type="hidden" name={name} value={value} />
       {combobox ? (
         <Ariakit.ComboboxProvider store={comboboxStore}>
-          <Ariakit.ComboboxSelect aria-label={label}>
+          <Ariakit.ComboboxSelect name={name} aria-label={label}>
             <Ariakit.ComboboxSelectedValue />
             {/* The portal keeps the popup outside the button in the DOM,
             while its React state belongs to the select's rendered subtree. */}
@@ -85,7 +79,7 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
         </Ariakit.ComboboxProvider>
       ) : (
         <Ariakit.SelectProvider store={select}>
-          <Ariakit.Select aria-label={label}>
+          <Ariakit.Select name={name} aria-label={label}>
             <Ariakit.SelectValue />
             <Ariakit.SelectPopover store={select} {...popupProps}>
               <Ariakit.SelectList>
--- a/app/src/sandbox/select-combobox-name-7346/index.react.tsx
+++ b/app/src/sandbox/select-combobox-name-7346/index.react.tsx
@@ -29,9 +29,10 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
     defaultSelectedValue: "Apple",
   });
   const [included, setIncluded] = useState(initiallyNamed);
+  const [fieldName, setFieldName] = useState("fruit");
   const [submitted, setSubmitted] = useState("Not submitted");
   const label = `${combobox ? "Combobox" : "Select"} ${initiallyNamed ? "named" : "unnamed"}`;
-  const name = included ? "fruit" : undefined;
+  const name = included ? fieldName : undefined;
   const popupProps = {
     "aria-label": label,
     role: "dialog",
@@ -49,7 +50,7 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
       onSubmit={(event) => {
         event.preventDefault();
         const data = new FormData(event.currentTarget);
-        const fruit = data.get("fruit");
+        const fruit = data.get(fieldName);
         setSubmitted(typeof fruit === "string" ? fruit : "Omitted");
       }}
     >
@@ -91,6 +92,9 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
           </Ariakit.Select>
         </Ariakit.SelectProvider>
       )}
+      <Ariakit.Button onClick={() => setFieldName("produce")}>
+        Rename {label}
+      </Ariakit.Button>
       <Ariakit.Button type="submit">Submit {label}</Ariakit.Button>
       <output aria-label={`${label} submission`}>{submitted}</output>
     </form>
--- a/app/src/sandbox/select-combobox-name-7346/index.react.tsx
+++ b/app/src/sandbox/select-combobox-name-7346/index.react.tsx
@@ -32,6 +32,9 @@ function Field({ combobox = false, initiallyNamed = false }: FieldProps) {
   const [submitted, setSubmitted] = useState("Not submitted");
   const label = `${combobox ? "Combobox" : "Select"} ${initiallyNamed ? "named" : "unnamed"}`;
   const name = included ? "fruit" : undefined;
+  const selectValue = Ariakit.useStoreState(select, "value");
+  const comboboxValue = Ariakit.useStoreState(comboboxStore, "selectedValue");
+  const value = combobox ? comboboxValue : selectValue;
   const popupProps = {
     "aria-label": label,
     role: "dialog",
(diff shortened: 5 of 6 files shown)


packages/ariakit-react-components/src/select/select.tsx
```diff
@@ -198,10 +198,10 @@ export const useSelect = createHook<TagName, SelectOptions>(function useSelect({
   // onChange event is triggered and we set the autofill state to true.
   props = useWrapElement(
     props,
-    (element) => {
-      if (!name) return element;
-      return (
-        <>
+    // Keep the control in the same slot when the native select toggles.
```
diegohaz: The comment states what the wrapper does, but not the constraint that forces it. A reader who does not already know the hazard sees an always-rendered fragment whose first slot is usually `false`, and `if (!name) return element;` looks strictly simpler. That is exactly the shape this repo has now had to correct in four wrappers: Dialog's fallback dismiss (#7333), Dialog's backdrop (#7335, #7342), and the two in this PR (#7346). The comment is the only local defense against a future "simplification" that reintroduces it.

The two fixes this PR follows both record the constraint and link the issue. `dialog.tsx:876-879`:

```
    // The backdrop toggles inside a wrapper that's always rendered. Returning
    // the bare element in one branch and a fragment in the other would change
    // the element type in the dialog's slot and remount everything inside it.
    // https://github.com/ariakit/ariakit/issues/7335
```

and `dialog.tsx:850-853`:

```
    // The button toggles inside a wrapper that's always rendered, rather than
    // by swapping the returned element, so gaining or losing it doesn't remount
    // the dialog.
    // https://github.com/ariakit/ariakit/issues/7335
```

`ariakit-general-code-style` also asks for the permalink when the full rationale lives in an issue. Matching that convention here:

```suggestion
    // The native select toggles inside a wrapper that's always rendered.
    // Returning the bare element in one branch and a fragment in the other
    // would change the element type in the control's slot and remount
    // everything inside it.
    // https://github.com/ariakit/ariakit/issues/7346
```

**Assessment**

- Severity: Low (estimate, 85% confidence), because nothing consumers observe changes today. The cost lands only if the construct is later reverted, which reintroduces #7346.
- Priority: P2 (estimate, 80% confidence), because this is comment-only and does not block the fix, but it is cheap and the convention is already set by the two sibling fixes in the same release.
- Breakage likelihood: Possible (estimate, 70% confidence), because the bare-return form reads as simpler and this exact shape has already regressed repeatedly. The new sandbox tests would catch a revert in CI, so the realistic cost is wasted round trips rather than a shipped regression.
- Fix effort: Easy (verified, 95% confidence), because it is two comment blocks. Evidence: the same edit applied at `combobox-select.tsx:204`.

packages/ariakit-react-components/src/combobox/combobox-select.tsx
```diff
@@ -201,10 +201,10 @@ export const useComboboxSelect = createHook<TagName, ComboboxSelectOptions>(
     // update the combobox store.
     props = useWrapElement(
       props,
-      (element) => {
-        if (!name) return element;
-        return (
-          <>
+      // Keep the control in the same slot when the native select toggles.
```
diegohaz: Same as the note on `select.tsx:201`: the comment records what the wrapper does but not why the fragment has to be unconditional, and it does not link the issue that holds the rationale.

```suggestion
      // The native select toggles inside a wrapper that's always rendered.
      // Returning the bare element in one branch and a fragment in the other
      // would change the element type in the control's slot and remount
      // everything inside it.
      // https://github.com/ariakit/ariakit/issues/7346
```

app/src/sandbox/select-combobox-name-7346/test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { expect, test } from "vitest";
+
+for (const label of [
+  "Select unnamed",
+  "Select named",
+  "Combobox unnamed",
+  "Combobox named",
+]) {
+  // https://github.com/ariakit/ariakit/issues/7346
+  test(`${label} keeps popup state when name changes`, async () => {
```
diegohaz, line 11: All four tests cross the falsy/truthy boundary, so nothing here pins down the neighbouring guarantee: changing `name` from one nonempty value to another must not remount either. `dialog-backdrop-7335` covers that case deliberately, and says why in `test.ts:39-41`:

```
// Keeps the backdrop truthy, so this passes before the fix too: it's the
// control that separates the falsy/truthy boundary from an ordinary re-render.
// https://github.com/ariakit/ariakit/issues/7335
```

It earns its place twice. It documents that the boundary, not any `name` change, is the cause, and it is the case that fails if someone later reaches for `<Fragment key={name}>` or an equivalent keyed wrapper. That variant still passes all four tests here while breaking every consumer that renames a live field.

Consumer scenario: a repeated fieldset that renumbers its fields after a row is removed, so a mounted `ComboboxSelect` goes from `items[1].fruit` to `items[0].fruit`, or a form builder that renames a field from its settings panel. Today the popup and everything a consumer rendered inside it survive that; nothing locks it in.

I verified the gap by adding a throwaway sandbox that keeps `name` truthy and switches it from `"fruit"` to `"produce"` behind a Rename button, then asserting the native select's `name` attribute, the uncontrolled draft, the child counter, and the dialog node identity. It passes both on `ccf49cf` and on `77b8081`, which is the control behavior. The fixture is removed; the worktree is clean.

The smallest addition is a rename affordance on `Field` and one more test:

```tsx
const [name, setName] = useState("fruit");
// ...
<Ariakit.Button onClick={() => setName("produce")}>Rename {label}</Ariakit.Button>
<Ariakit.Select name={included ? name : undefined} aria-label={label}>
```

```ts
// Keeps name truthy, so this passes before the fix too: it's the control that
// separates the falsy/truthy boundary from an ordinary re-render.
// https://github.com/ariakit/ariakit/issues/7346
test(`${label} keeps popup state when name is renamed`, async () => {
  // open, type, add attachment, click Rename, assert the draft and counter
  // survive and that the native select carries the new name
});
```

Worth mirroring in `test-chrome.ts`, the way `dialog-backdrop-7335` mirrors its control.

**Assessment**

- Severity: Low (estimate, 80% confidence), because this is a regression-coverage gap rather than a defect. The behavior itself is correct on this head, which I confirmed above.
- Priority: P2 (estimate, 75% confidence), because the fix and its four regressions stand on their own. This hardens them against the next change to the same wrapper.
- Breakage likelihood: Unlikely (estimate, 70% confidence), because it takes a future edit that keys the wrapper on `name` or otherwise makes any `name` change remount. The precedent shows this family of wrappers does attract exactly that kind of edit.
- Fix effort: Easy (verified, 90% confidence), because I built and ran the fixture while reviewing. Evidence: roughly 20 lines of sandbox plus one test per suite.

diegohaz (commented):
## Review summary

**Type:** initial review. No prior review carried this workflow's marker, and the PR had no review threads, so the baseline is the merge base `ccf49cf9ed5eedfa36db99e776f131936101de1b` against head `77b80810881bcd21005f2a032038e2efab3aa6b0`. Scope was the full diff (6 files) plus the surrounding wrapper family, the sibling fixes it follows, and the fixture precedent.

### Verdict

The fix is correct, minimal, and consistent with the established treatment for this hazard. I traced React's reconciliation for both transitions and confirmed the mechanism rather than assuming it. An unkeyed top-level fragment is unwrapped by `reconcileChildFibers`, so the wrapper's children become an array in both branches. `false` occupies index 0 without producing a fiber, which keeps the control at index 1 across the toggle. Going unnamed to named, the fast loop creates the `select` fiber at index 0 and reuses the control at index 1. Going named to unnamed, the boolean slot drops React into the keyed-map path, which still matches the control at index 1 and deletes only the native select. `placeChild` sees `oldIndex >= lastPlacedIndex` in both directions, so the control is not even moved in the DOM, and the native select keeps rendering before it.

Two findings, both P2 and neither blocking.

### Findings published inline

- `select.tsx:201` and `combobox-select.tsx:204` (r3890636878, r3890637228): the new comment records what the wrapper does but not the constraint that forces it, and it omits the issue permalink. `dialog.tsx:850-853` and `dialog.tsx:876-879`, the two sibling fixes this PR follows, both state the constraint and link #7335. Without that, `if (!name) return element;` reads as strictly simpler and invites a revert. Severity Low, Priority P2, Breakage likelihood Possible, Fix effort Easy.
- `test.ts:11` (r3890637465): no control case keeps `name` truthy across a change. `dialog-backdrop-7335/test.ts:39-41` adds one deliberately, and it is the case that fails if someone later reaches for a keyed wrapper such as `<Fragment key={name}>`, which would still pass all four tests here. Severity Low, Priority P2, Breakage likelihood Unlikely, Fix effort Easy.

### Candidates evaluated and not published

- **Other wrappers with the same shape.** I scanned every `{element}` inside a fragment across `packages/ariakit-react-components/src`. Seven sites render a sibling before the element: the two in this PR, `dialog.tsx:867` and `dialog.tsx:891`, `portal.tsx:306` and `portal.tsx:391`, and `focus-trap-region.tsx:61`. After this PR every one of them uses the stable-fragment pattern. `combobox.tsx:713` renders its hidden inputs after the element, which is why it was already safe. `combobox-cancel.tsx:87` swaps the wrapper's own element on purpose. `portal.tsx:271` is #7343, closed as working as expected. No unfixed site remains. Not a finding.
- **Element-first ordering as a simpler alternative.** Moving `{element}` to index 0 and the native select after it would also fix the remount and touch fewer lines, and `combobox.tsx` proves the shape works. Rejected: it would move the native select after the trigger in the DOM for every consumer, which is a rendered-output change well beyond the bug. Keeping DOM order is the better trade-off.
- **Reconciliation cost of the boolean slot.** The `false` at index 0 makes React break out of the fast child loop and build the remaining-children map on every update of an unnamed control. Real, but it is one small map per re-render, it is already the codebase-wide pattern, and this PR's `combobox-select-perf` numbers are flat (open -4%, move -0%, close +3%, all within noise). Invalid as a finding.
- **`test-chrome.ts` versus `test-browser.ts`.** The tests assert native form submission, which the playwright skill lists as platform-sensitive. The contract under test is React reconciliation, which is engine-independent, so the default single-project choice is right. Not a finding.
- **Changeset does not link the `name` prop.** The sibling entry `300-dialog-backdrop-remount.md` links `backdrop` with an anchor, so I checked whether `name` should match. Neither reference page emits a `#name` anchor, because `name` is inherited through `Pick<SelectHTMLAttributes<...>>` with no JSDoc. The skill says to keep the name as plain inline code in that case, so the changeset is correct as written. Invalid.
- **Native-select block duplicated between the two files.** Roughly 55 near-identical lines, and this PR edits both copies identically. Declined non-goal: it is pre-existing, this PR does not worsen it, and `Select` is soft-deprecated in favour of `ComboboxSelect`, so the duplication is expected to go by attrition rather than by extracting a shared hook now. No follow-up issue registered.
- **`label.endsWith(" unnamed")` recomputed with its own negation** in both test files. Iterating structured `[label, initiallyNamed]` pairs would drop the label parsing. Declined non-goal: low-value readability, and the tests read clearly as they are.

No unresolved review threads existed at the start of this run, so there was nothing to revalidate or reply to. I read every top-level review body: CodeRabbit reported no actionable comments, Codex completed with no findings comment, and the perf comment is a status report. CodeRabbit's "Docstring Coverage" warning is generic tooling policy that does not apply to these files. No follow-up issues were registered.

### What I ran

Reviewed in a dedicated worktree at the PR head.

- `pnpm test select-combobox-name-7346`: 4 passed.
- Regression proof: reverted only `select.tsx` and `combobox-select.tsx` to `ccf49cf` and re-ran. All 4 failed at `expect(q.textbox("Note")).toHaveValue("Buy ripe fruit")`, receiving an empty value, which is the remount discarding the uncontrolled draft. Restored and re-verified clean.
- `pnpm test-react18 select-combobox-name-7346`: 4 passed.
- `pnpm test select`: 41 files, 230 passed. `pnpm test combobox`: 71 files, 454 passed.
- Control experiment: a throwaway sandbox that keeps `name` truthy and renames it from `"fruit"` to `"produce"`, asserting the native select's `name` attribute, the uncontrolled draft, a child counter, and dialog node identity. It passes on both `ccf49cf` and `77b8081`, which is what makes it a control. Removed afterwards; `git status` is clean.
- Verified `https://ariakit.com/reference/select` and `https://ariakit.com/reference/combobox-select` return 200, and that neither page emits a `#name` anchor.
- Read #7346, #7335, #7333, #7342 and the `dialog-backdrop-7335` fixture to check the approach and fixture conventions against precedent.

### Checked and clean

- Changeset: `300-` prefix for a patch, correct frontmatter for the direct package plus the public re-export, single past-tense sentence, valid reference links. Issue author is excluded from credit, and the issue has no other participants, so no `Thanks to` sentence is due.
- Fixture lifecycle: no existing semantic fixture coherently owns this scenario, so a new issue-number sandbox matches step 2, exactly as `dialog-backdrop-7335` did. The issue URL sits directly above the generated tests in both suites.
- The sandbox drives everything through accessible roles and visible output with no test-only globals, and `preview.json` is correctly omitted for a sandbox that needs no preview metadata.
- No public export, JSDoc, or declaration changes, so the dual-generator reference check does not apply. No Solid counterpart exists for these components.
- SSR is unaffected: the `false` branch emits no markup, so client and server output match.

### Not fully validated

- I did not run the Playwright suite locally and did not start the app dev servers. I relied on the green `App / Test Chrome`, Firefox, Safari, iOS, and Android jobs at this head.
- Real autofill behaviour across the toggle, such as a password manager moving focus into the native select right as `name` flips, is unverified outside the synthetic `onFocus` and `onChange` handlers.
- No screen-reader verification. Nothing in the diff changes the accessible tree, since the native select stays `aria-hidden`.

packages/ariakit-react-components/src/select/select.tsx
diegohaz: Fixed in [`837cdc451`](https://github.com/ariakit/ariakit/commit/837cdc4517c49a2ec6234157a54a71a08b309ce5): the comment now explains that switching between the bare element and a fragment remounts the control subtree, and links #7346. The same explanation appears in both wrappers.

packages/ariakit-react-components/src/combobox/combobox-select.tsx
diegohaz: Fixed in [`837cdc451`](https://github.com/ariakit/ariakit/commit/837cdc4517c49a2ec6234157a54a71a08b309ce5): this wrapper now documents the same reconciliation constraint and links #7346.

app/src/sandbox/select-combobox-name-7346/test.ts
diegohaz: Already handled at `select.tsx:206` and `combobox-select.tsx:209`: changing `name="fruit"` to `name="produce"` preserves the wrapper and its descendants. Added targeted coverage in [`8de356ab2`](https://github.com/ariakit/ariakit/commit/8de356ab213a62613cfaac769bcb3a7758358661). Both controls now have rename cases in Chromium and the integration suite. They check the native field name, draft, attachment state, and submission; the integration tests also check dialog identity.

All six cases pass in React 19, React 18, and Chromium. The two new cases also pass with the pre-fix source. Temporarily adding `<Fragment key={name}>` makes both new cases fail at the draft-value assertion in React and Chromium; the stable wrappers were restored afterward.

Visual evidence gap: a local recording verifies the rename flow, but it is not attached because browser uploads require explicit permission.

diegohaz:
**Fixed**

- [`837cdc451`](https://github.com/ariakit/ariakit/commit/837cdc4517c49a2ec6234157a54a71a08b309ce5) ([Select reply](https://github.com/ariakit/ariakit/pull/7352#discussion_r3890662586), [ComboboxSelect reply](https://github.com/ariakit/ariakit/pull/7352#discussion_r3890662844)): explain why both wrappers must stay stable and link #7346.

**Already resolved**

- [`8de356ab2`](https://github.com/ariakit/ariakit/commit/8de356ab213a62613cfaac769bcb3a7758358661) ([reply](https://github.com/ariakit/ariakit/pull/7352#discussion_r3890663217)): add proof tests for the existing `name="fruit"` → `name="produce"` behavior. All six cases pass in React 19, React 18, and Chromium. Both new cases fail with a temporary keyed-wrapper regression. Lint and type checks pass.

**Evidence gaps**

- [`Local recording`](https://github.com/ariakit/ariakit/pull/7352#discussion_r3890663217) ([reply](https://github.com/ariakit/ariakit/pull/7352#discussion_r3890663217)): the rename recording is not attached because browser uploads require explicit permission.

diegohaz (commented):
## Re-review summary

**Type:** re-review. Incremental baseline is `77b80810881bcd21005f2a032038e2efab3aa6b0`, the head SHA on my previous marked review, against the current head `8de356ab213a62613cfaac769bcb3a7758358661`. Two new commits, `837cdc451` and `8de356ab2`, touching five files for +70/-4. I also re-checked the surrounding wrapper family for anything the new commits expose.

**No new findings.** Both findings from the previous review are resolved, and I verified each against the current head rather than relying on the thread replies.

### Previous findings

- **Wrapper rationale comment** (r3890636878, r3890637228, both resolved). Confirmed fixed in `837cdc451`. `select.tsx:201-204` and `combobox-select.tsx:204-207` now state the constraint and link #7346. The landed wording drops the "change the element type in the control's slot" clause from my suggestion and lands at three lines plus the link, which is the better call: it keeps the load-bearing part, and `ariakit-general-code-style` asks for no more than three `//` lines excluding the source link. Matches `dialog.tsx:850-853` and `dialog.tsx:876-879`.
- **Missing truthy-`name` control** (r3890637465, resolved). Confirmed fixed in `8de356ab2`. `fieldName` state plus a `Rename {label}` button drive `"fruit"` to `"produce"`, and both suites gained a case per control that asserts the native field name, the draft, the child counter, and submission, with `test.ts` also asserting dialog node identity.

### Verification of the reply's claims

Each claim reproduced independently in a clean worktree at `8de356ab2`.

- Six cases pass in React 19 (`pnpm test select-combobox-name-7346`) and React 18 (`pnpm test-react18`).
- Six cases pass in Chromium, run locally against `pnpm dev-app` from this same checkout on `APP_PORT=4323`. This confirms the claim before `App / Test Chrome` finished; it has since passed on CI.
- **Control property holds.** Reverting only `select.tsx` and `combobox-select.tsx` to `ccf49cf` gives `4 failed | 2 passed`: the four boundary cases fail and both rename cases pass. That is exactly the role `dialog-backdrop-7335` gives its own control, and it is what makes the four failures attributable to the empty/nonempty boundary rather than to any `name` change.
- **Keyed-wrapper experiment holds.** Replacing `<>` with `<Fragment key={name}>` in `select.tsx` alone fails `Select named keeps popup state when name stays nonempty` at `test.ts:62`, the draft assertion, as described. One precision note, not a correction: that experiment also fails the two `Select` boundary cases, so it does not by itself isolate what the new cases add. Their distinct value is the control property above. Restored afterwards and re-confirmed green.
- `pnpm test select` (232 passed) and `pnpm test combobox` (456 passed) confirm the shared-fixture edit did not disturb neighbouring coverage. All 48 CI checks pass at this head, including Chrome, Firefox, both Safari shards, iOS, Android, and Visual.

### Candidates evaluated in the new diff and not published

- **`Ariakit.Button` default type for the new Rename button.** It sits directly inside the `form`, unlike `Add attachment`, which is portaled out. Checked `withDefaultButtonType`: it applies `type="button"` when no `type` and no `render` are given, so Rename cannot submit the form. Not a finding.
- **Setup duplicated across the two loops** in both suites: open, draft, add attachment. A shared helper would remove about five lines in four places. Declined non-goal, since `dialog-backdrop-7335/test.ts` repeats the same setup across its three tests, so this matches the fixture convention rather than departing from it.
- **No `select` count assertion after the rename.** The first loop asserts the count across the toggle, the rename case asserts only the name attribute. Declined non-goal: the wrapper renders exactly one `select`, so a stale duplicate has no mechanism, and `querySelector` returning the first match already fails the attribute assertion if a stale element preceded the new one.
- **No comment on the Rename affordance in `index.react.tsx`.** `dialog-backdrop-7335` comments its equivalent control in the fixture source. Declined non-goal: both test suites carry the three-line explanation directly above the generated tests, so a third copy adds nothing. Same disposition as the analogous point in the previous review.
- **Chrome rename case omits the dialog-identity assertion** that `test.ts` makes. Not a finding: Playwright locators re-resolve lazily, the existing Chrome cases are written the same way, and `test.ts` covers identity.
- **`data.get(fieldName)`** replacing `data.get("fruit")` in `onSubmit`. Checked against the original four cases, which never click Rename, so `fieldName` stays `"fruit"` and the `"Omitted"` and `"Apple"` expectations are unchanged. Confirmed by the passing run. Not a finding.

### Description accuracy

The Tests section was updated for the new cases and every claim in it matched what I reproduced. Two small points, neither blocking:

- The earlier paragraph still reports "34 targeted browser checks", which predates the two new cases. Harmless, since the new paragraph reports the current round separately.
- The Preview section still states that the before-and-after recordings are not attached because the upload needs explicit permission. `ariakit-general-browser-evidence` asks for that evidence on a browser-visible fix, so this remains an open, author-acknowledged gap rather than something this review discovered.

### Not fully validated

- Firefox and Safari were not run locally. I relied on the green CI jobs at this head, which now include both Safari shards.
- Real autofill behaviour across a rename, such as a password manager acting on the native select as its `name` changes, is still unverified outside the synthetic handlers.

Issue #3285, opened 2023-12-30
[Combobox] It's not possible to scroll outside of focused item


honzatmn:
### Current behavior

In [multiselect example](https://ariakit.org/examples/combobox-multiple) (with `focusOnHover` enabled), when scrolling, the focus doesn't allow to scroll the item outside of the visible area and results in glitching (see video)

https://github.com/ariakit/ariakit/assets/6851247/27cc6424-bbff-4b7a-8cc7-d525fab5c717

### Steps to reproduce the bug

1. Open [multiselect example](https://ariakit.org/examples/combobox-multiple) 
2. Hover over an item
3. Try to scroll away

### Expected behavior

It's possible to scroll the list without issues

### Workaround

Set [`resetValueOnSelect`](https://ariakit.org/reference/combobox-provider#resetvalueonselect) to false and manually reset the combobox value.

diegohaz:
Thanks for the report, @honzatmn!

Are you able to reproduce this issue locally, or does it only occur on the site? I suspect that an unreleased change might have introduced this regression.

PR #7357, opened 2026-08-31
Fix delayed active item after the combobox list changes

### LINKED ISSUE #3914: [Combobox + Select] - Active item change is delayed when controlled
### Current behavior

When `activeItem` is controlled, there is a delay bewteen the list being updated, and the new active item changing. this is causing a flicker in the UI.
when `activeItem` is uncontrolled, this does not seem to happen.
Notice here that `Andora` is already at the top, but `Afghanistan` is still active
CleanShot 2024-07-03 at 16 06 49

1. Open https://stackblitz.com/edit/hx9wql?file=select-combobox-virtualized%2Fstyle.css,select-combobox-virtualized%2Fcountries.ts,select-combobox-virtualized%2Findex.tsx&theme=light
2. type `a`
3. type `n`
4. click `backspace` to remove the `n`
5. repeat steps 3 and 4

### Expected behavior

list and active item should change in the same tick

### Workaround

Feed `autoSelect` your own up-to-date matches through `getAutoSelectId`, with the assumptions and limitations listed in https://github.com/ariakit/ariakit/issues/3914#issuecomment-5474957327.

Leaving `activeId` and `setActiveId` uncontrolled does **not** avoid this, and the earlier `flushSync` workaround is obsolete on current React 18 and 19. Both are corrected in the comment above.

### Possible solutions

Prototype a Combobox-only pre-paint registration path: let `ComboboxItem` opt into layout-effect item registration so `renderedItems` is available before the next browser frame, while keeping the existing `autoSelect` effect unchanged. The exact rAF regression was red on current main and green with this prototype in React 18 and 19 Chromium.

This removes the reported painted flicker, but intermediate React commits still contain the mismatched values. If literal single-commit atomicity is required, add a renderer/store transaction that owns the prospective filtered items and auto-select target together, and stages the list until the controlled `activeId` commit acknowledges it.

Application code can achieve strict atomicity for known items by coordinating both updates:

```tsx
startTransition(() => {
  setMatches(nextMatches);
  setActiveId(nextFirstItemId);
});
```

See the [local experiment report](https://github.com/ariakit/ariakit/issues/3914#issuecomment-5064620745) for the red/green frame matrix, prototype shape, validation, and tradeoffs.

diegohaz:
## Motivation

After a keystroke changes a combobox's filtered list, the new list and the auto-selected item reach the screen in different frames. Users see the new first option while the previously active option is still highlighted.

The report describes this with a controlled [`activeId`](https://ariakit.com/reference/combobox-provider#activeid), and that is how it was noticed, but the flicker is not exclusive to the controlled case. Removing the `activeId` prop from the reproduction leaves the same six stale frames per interaction.

The cause is a timing chain. [`CollectionItem`](https://ariakit.com/reference/collection-item) registers items in a passive effect, and the store published its public `renderedItems` only from a `requestAnimationFrame`. The `autoSelect` behavior derives its target from `renderedItems`, so after each keystroke it re-asserted the previous list's first item and corrected itself a frame or more later.

Fixes #3914

## Solution

Collection stores now publish their rendered items in the batch microtask that follows the registrations, instead of waiting for an animation frame. Everything derived from those items, including the Combobox auto-selected item, now describes the list the user is about to see.

```diff
-      let raf = requestAnimationFrame(() => {
-        const { renderedItems } = collection.getState();
-        if (state.renderedItems === renderedItems) return;
+      if (state.renderedItems !== collection.getState().renderedItems) {
         sortItems(state.renderedItems);
-      });
+      }
```

The `IntersectionObserver` re-sort keeps its animation frame, because scrolling fires it constantly.

This removes the stale frame that was previously guaranteed. The corrective render still comes from a separate scheduler task, so this is a timing improvement rather than an atomicity guarantee: the list and the active item are not merged into a single React commit.

Registering `CollectionItem` in a layout effect was measured first, since that was the suspected fix. It made no difference at all, producing the same six stale frames per test: registration and publication are decoupled by the batch microtask, so moving registration earlier within the same commit does not remove the frame hop.

### Verification

- The reproduction tests pass, and the store-level test `publishes rendered items before the next animation frame` fails against the previous behavior and passes with this change.
- `pnpm test` 1959 passed, `pnpm test-react18` 1080 passed.
- Full Chrome suite, clean sequential local runs: 871 passed with 0 flaky before, 872 passed with 1 flaky after, at 2.6m and 2.5m.
- The `perf` workflow reports no confirmed regression across its 27 tests. The only flagged row is an improvement, `tab-perf > mount tabs` at -12% scripting and -11% total. The rows most exposed to the lost intra-frame sort coalescing came out flat or faster: `collection-renderer-perf > scroll nested renderers` -2%, `toggle renderer items` -5%, `composite-perf > mount composite` -4%. `composite-perf > mount controlled composite` is +9% total but stayed below the flagging gate, so it is unconfirmed and worth a second look if it repeats.

## Workaround

Until this ships, applications can feed `autoSelect` their own up-to-date matches through [`getAutoSelectId`](https://ariakit.com/reference/combobox#getautoselectid), instead of letting it read the store's rendered items:

```tsx
const matches = useMemo(
  () =>
    matchSorter(countries, searchValue).map((country) => ({
      id: `country-${country}`,
      value: country,
    })),
  [searchValue],
);

// TODO: Remove when https://github.com/ariakit/ariakit/issues/3914 is fixed.
const firstMatchId = matches[0]?.id ?? null;

<Ariakit.Combobox autoSelect getAutoSelectId={() => firstMatchId} />;
```

This snippet is deliberately narrower than the real `autoSelect` policy. It assumes:

- **The match list is derived in the same render as the input value.** Deriving the matches from a `useDeferredValue` puts the new list in a commit that arrives after the auto-select effect has already read the callback, and the stale highlight comes back. With `startTransition` around the value setter, the highlight stayed correct in the reproduction sandbox.
- **Every item has a stable, application-controlled `id` that the store can resolve.** The callback returns an id, so items rendered with automatically generated ids cannot use this. The id must also belong to an item the store already knows, because it is currently rendered or listed in `items` or `defaultItems`. Returning an id the store cannot resolve makes `autoSelect` move to it twice, which is the re-focus that #3837 avoids.
- **One flat list, in DOM order.** It returns `matches[0]`, whereas the default reads the store's DOM-ordered rendered items. With grouped results, a pinned "Create ..." option, or more than one item source, the first array entry is not the first rendered option, and the wrong option is highlighted.
- **No disabled items and no tabs.** The default skips disabled items and `role="tab"` items and falls back to the first enabled item. Add a filter such as `matches.find((item) => !item.disabled)?.id` when the list has disabled options.
- **Nothing is rendered when there are no matches.** Returning `null` moves the active item back to the input, which matches the default for an empty list. An application that renders an empty-state `ComboboxItem` instead gets the highlight cleared while an option is still on screen.
- **`autoComplete="list"`, the default.** With `"inline"` or `"both"`, the active value becomes current before the store's rendered items do, so the inline completion is not highlighted for a frame.

When the application already passes `getAutoSelectId`, merge this into the existing callback. The `renderedItems` argument cannot do the filtering, because that argument is the stale array this works around.

The snippet does not change the paths that run before `getAutoSelectId` is consulted. IME composition, the user-scrolled guard, and `preserveSelectedValue` for a combobox inside a select all behave normally.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the synchronous publish, unchanged</summary>

**Assessment**

Severity is moderate and visual: a guaranteed painted frame in which the list is new while everything derived from it, including the highlight and `aria-activedescendant`, is still the previous list. Expected frequency is high per occurrence, since it happens on every keystroke in a combobox with `autoSelect` and on every composite mount. Supported scope is core: `renderedItems` is documented as being in DOM order, but nothing in the public contract promised when it publishes, and no repository code depended on the delay. Likely user impact is broad and positive, because every collection component converges one frame earlier to the same value.

Implementation complexity is negative relative to `main`. The change deletes a `requestAnimationFrame`, its closure, and a cleanup path, and replaces a guarded return with a plain early return. It adds no state, helper, abstraction, public contract, or platform policy. Maintenance complexity concentrates in one pre-existing area: the `IntersectionObserver` re-sort, whose rationale was undocumented before this change.

Across three cycles the findings were one logic defect, one missing test, one changeset factual error, and three comment-accuracy findings. Three of those prose findings concern the legacy `IntersectionObserver` block rather than the new code, and cycle 3 was partly caused by cycle 2 replacing an imprecise comment with a false one. The changed logic itself did not move after cycle 1.

**Decision**

Continue the current design. The alternatives are worse: registering `CollectionItem` in a layout effect was measured to give zero improvement and would move per-item `setState` calls into the commit phase; keeping the animation frame alongside a synchronous publish adds machinery that always bails out; publishing unsorted breaks the documented DOM-order contract and the `autoSelect` target; and decoupling `autoSelect` from `renderedItems` is larger and would leave the other collection components stale.

Intentionally unsupported: the loss of intra-frame sort coalescing across separate tasks, which the `perf` workflow gates automatically for this path, with an O(n) already-in-order pre-check in `sortBasedOnDOMPosition` held back unless CI flags a row; a user-supplied `getAutoSelectId` returning an id the store never registered, which is pre-existing and unchanged in kind; and any consumer relying on `renderedItems` lagging by a frame.

No follow-up issues were registered.

</details>

## Acknowledgments

Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi) for reporting the issue with a reproduction.


.changeset/300-combobox-auto-select-frame-lag.md
```diff
@@ -0,0 +1,12 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed the auto-selected item lagging behind the filtered list
```
diegohaz: This should be a single-paragraph changeset.

diegohaz: Confirmed against the current head: `.changeset/300-combobox-auto-select-frame-lag.md` is still four paragraphs. `ariakit-general-changeset` reserves the multi-paragraph form for entries that need a title plus a focused code example, normally a new or changed public API. This entry is a behavior fix with no example, so the simple form applies. The first paragraph is also a complete past-tense sentence, which is the simple-entry shape rather than a title.

Collapsing it changes two things that the multi-paragraph rules were carrying:

- The credit sentence loses its role clause. A single-paragraph entry ends with `Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi).`, not `... for reporting the issue.`
- The derived-component list drops to at most two representatives, and the entry still has to say that the change reaches every component built on the base component.

Suggested rewrite:

```md
---
"@ariakit/react": patch
---

Fixed [`Combobox`](https://ariakit.com/reference/combobox) with [`autoSelect`](https://ariakit.com/reference/combobox#autoselect) keeping the previously active option highlighted for one frame after the filtered list changes. The same timing applies to all components built on [`CollectionItem`](https://ariakit.com/reference/collection-item), such as [`SelectItem`](https://ariakit.com/reference/select-item) and [`MenuItem`](https://ariakit.com/reference/menu-item). Thanks to [@ItaiYosephi](https://github.com/ItaiYosephi).
```

Checked while validating this: all five reference pages return `200`, the combobox page does expose an `autoselect` anchor, and `@ariakit/react` re-exports `CollectionItem`, `SelectItem`, `MenuItem`, and `Tab`, so the public-package scope holds in either form. `.changeset/300-collection-rendered-items-timing.md` already follows the simple form, keeps its credit sentence free of role clauses, and needs no change.

**Assessment**

- Severity: Low (verified, 95% confidence), because it affects changelog prose only and no shipped behavior.
- Priority: P1 (estimate, 90% confidence), because the entry ships with the release and the format is a stated repository convention.
- Breakage likelihood: Rare (verified, 95% confidence), because nothing consumes the file except the generated changelog.
- Fix effort: Easy (verified, 95% confidence), because it is a single-file rewrite.

packages/ariakit-components/src/collection/collection-store.test.ts
```diff
@@ -866,3 +866,31 @@ test("updates rendered items after sorting a large collection", async () => {
...
+    store.renderItem({ id: "second", element: second });
+    store.renderItem({ id: "first", element: first });
+
+    // Both registrations share the microtask queued by the first one, and the
+    // sort publishes inside it, so one await is enough and the second is
+    // margin. Awaiting microtasks rather than polling is what makes this fail
+    // when the publish waits for an animation frame, which is the delay behind
+    // https://github.com/ariakit/ariakit/issues/3914.
+    await Promise.resolve();
+    await Promise.resolve();
```
diegohaz, line 885: The second `await Promise.resolve();` is dead weight, and the comment above it already says so. Both `renderItem` calls share the microtask that the first one queues, and the synchronous `sortItems` publishes inside that flush, so `renderedItems` is already sorted after one await.

I verified this locally: deleting line 887 keeps `pnpm test collection-store` green at 31 passed, and the test still goes red for the right reason against the base store (`AssertionError: expected [] to deeply equal [ 'first', 'second' ]`). So the margin await buys no coverage.

Removing it also lets the comment shrink to the constraint that carries the test, which is that awaiting a microtask rather than polling is what makes it fail when the publish waits for an animation frame. `ariakit-general-code-style` asks for the smallest comment that preserves the context, and the sentence that explains the padding goes away with the padding.

```suggestion
    // Both registrations share the microtask queued by the first one, and the
    // sort publishes inside it. Awaiting a microtask rather than polling is
    // what makes this fail when the publish waits for an animation frame,
    // which is the delay behind
    // https://github.com/ariakit/ariakit/issues/3914.
    await Promise.resolve();
```

**Assessment**

- Severity: Low (verified, 90% confidence), because the extra await does not change what the test proves.
- Priority: P2 (estimate, 85% confidence), because it is a readability cleanup in new test code, not a coverage gap.
- Breakage likelihood: Rare (verified, 90% confidence), because the assertion is unchanged and the red check against the base store still fails.
- Fix effort: Easy (verified, 95% confidence), because it is one deleted line and a shorter comment.

diegohaz (commented):
Initial review of all 9 changed files, baselined on the merge base `bbba558c15d8f5b18776c869164c8e41a65dd251` against head `1ec20e4e`. No prior review from this workflow exists on the PR, so this is a full review rather than an incremental one.

The change holds up. It is a net deletion in `collection-store.ts`, it adds no state or abstraction, and the two published behaviors it touches are both proven by tests that go red on the base commit.

## What I ran

- `pnpm test`: 278 files, 1959 tests passed.
- `pnpm test-react18`: 206 files, 1080 tests passed.
- `pnpm tsc`: clean.
- `pnpm test-chrome` in `app`, full suite against `pnpm dev-app`: 868 passed, 1 skipped, 0 failed, 0 flaky, 3.3m.
- Red checks, by temporarily restoring the base `collection-store.ts` and then restoring the branch file:
  - `publishes rendered items before the next animation frame` fails with `expected [] to deeply equal [ 'first', 'second' ]`, and the other 30 store tests stay green, so the new test is the only thing it proves.
  - Both new Chrome tests fail with the reported flicker in the ledger, for example `{ first: "Andorra", active: "Afghanistan" }`, which is the exact frame the issue describes.
- Comment and link checks: the five reference URLs in the changesets return `200`, the combobox page exposes an `autoselect` anchor, and `@ariakit/react` re-exports `CollectionItem`, `SelectItem`, `MenuItem`, and `Tab`.

## Verified while reviewing

- **The timing claim is real, and the mechanism is the one the PR states.** `batch` listeners run from `queueMicrotask`, so calling `sortItems` in the listener body publishes to the public store inside that flush. The public JSDoc for `renderedItems` documents DOM order only and never promised a publication time, so nothing in the documented contract moves.
- **The guard is still load-bearing.** `sortBasedOnDOMPosition` returns the same array reference when the order is already correct, so a publish converges in at most two passes even without the guard. The guard still matters for the case where DOM order is not stable between passes, where it stops an unbounded microtask loop. That is the same protection the previous animation-frame bail-out gave.
- **Observer churn does not increase.** In the already-ordered case both the old and new code build the `IntersectionObserver` once, and in the reorder case both build it twice. The new ordering is slightly better, because the observer now sees the sorted array instead of the pre-sort one.
- **The `#3285` reference is accurate.** That issue is the scroll regression fixed by #3286, which is what introduced the animation frame inside `ioCallback`, so keeping it there and citing that issue matches the history.
- **The `composite-item.tsx` window did not change.** Children commit their layout effects before the parent publishes `compositeElement`, and item registration runs in a later passive effect, so `renderedItems` still cannot become non-empty before `compositeElement` is set. The composite and tab suites agree.
- **`menu-maximum-update-depth`**, the fixture added with #3286 to guard the update-depth regression, still passes.

## Findings

Two, both non-blocking, one of them yours.

1. `.changeset/300-combobox-auto-select-frame-lag.md` should be a single-paragraph entry. Validated on the existing thread with the concrete rewrite and the two follow-on rules it triggers, the credit form and the derived-component count.
2. A redundant `await Promise.resolve();` in the new store test, published inline with the local evidence.

## Not published inline

- **`composite-perf > react > mount controlled composite` at +9% total and +10% scripting.** I could not validate this in the time available, and the perf workflow left it below its flagging gate, so it is a suspected issue rather than a finding. It is worth a second look only if it repeats on a later run. No impact metadata, because I did not confirm it.
- **The `IntersectionObserver` block is disconnected and rebuilt on every publish.** Declined non-goal for this PR. It predates the change, the change does not increase how often it happens, and reworking the observer lifecycle is a separate design question.
- **The loss of intra-frame sort coalescing across separate tasks.** Declined non-goal. The PR discloses it, the perf workflow reports the exposed rows as flat or faster, and my full Chrome run showed no regression or flake.

No follow-up issues were registered, because nothing reached the bar for confirmed independent work.

## Not fully validated

- I did not re-run the `perf` workflow locally, so the performance conclusions rest on the run reported in the description plus the flat full Chrome suite.
- Firefox, Safari, and the mobile projects were not run. The new tests are Chrome-only by design, and the contract is a DOM and state contract rather than an engine-specific one, so that filename choice looks right.

.changeset/300-combobox-auto-select-frame-lag.md
diegohaz: Fixed in [`286607bc`](https://github.com/ariakit/ariakit/commit/286607bc270e2ae9a6b60a6933607a64096d1e00): collapsed the entry to the single-paragraph form, dropped the role clause from the credit sentence, and trimmed to two derived representatives while keeping the statement that the change reaches every component built on `CollectionItem`.

Verified the form before applying rather than taking it on trust. `ariakit-general-changeset` ties the multi-paragraph shape to a title plus a focused code example, normally a new or changed public API, and this entry has neither. The old first paragraph was also a complete past-tense sentence, which is the simple-entry shape rather than a title. In-repo precedent agrees: `300-dialog-own-dismiss.md`, `300-select-name-remount.md`, and `300-combobox-item-store-role.md` are all single-paragraph bug fixes.

Two deviations from your suggested text, both easy to revert if you prefer the original:

- `The same timing applies` became `The same fix applies`. Following a sentence that describes the defect, "the same timing applies" can read as those components still having the problem, while `ariakit-ariakit-changeset` asks the entry to tell readers the change reaches them.
- I kept that as a separate sentence instead of the inline appositive used by `300-dialog-own-dismiss.md` (`Fixed modal Dialog, and every component built on it such as Menu, ...`). There the symptom generalizes to the derived components. Here `autoSelect` is combobox-only, so folding `SelectItem` and `MenuItem` into the first sentence would describe a symptom they do not have.

packages/ariakit-components/src/collection/collection-store.test.ts
diegohaz: Fixed in [`1e9056c8`](https://github.com/ariakit/ariakit/commit/1e9056c810ad5cb0d4c614aeadde488a8343dc90): removed the second await and shortened the comment.

I re-ran the red check rather than assuming it still held. With a single await the test fails against the pre-fix store for the right reason, `AssertionError: expected [] to deeply equal [ 'first', 'second' ]` at 30 passed and 1 failed, and passes with the fix. It also cannot become flaky: happy-dom implements `requestAnimationFrame` with `setImmediate`, a macrotask, so a microtask await can never outrun it.

One deviation from your suggestion: I dropped the trailing `which is the delay behind` clause so the URL sits on its own line with no trailing period. That matches the example in `ariakit-general-code-style/references/comments.md`, which pairs a finished sentence with a bare link line, and the two issue links this PR adds in `collection-store.ts`. It also brings the block back to three prose lines plus the link.

diegohaz:
**Fixed**

- [`286607bc`](https://github.com/ariakit/ariakit/commit/286607bc270e2ae9a6b60a6933607a64096d1e00) ([reply](https://github.com/ariakit/ariakit/pull/7357#discussion_r3892794060)): collapsed `300-combobox-auto-select-frame-lag.md` to the single-paragraph form, dropped the role clause from the credit sentence, and trimmed to two derived representatives while keeping the statement that the change reaches every component built on `CollectionItem`.
- [`1e9056c8`](https://github.com/ariakit/ariakit/commit/1e9056c810ad5cb0d4c614aeadde488a8343dc90) ([reply](https://github.com/ariakit/ariakit/pull/7357#discussion_r3892794680)): removed the redundant second `await Promise.resolve()` in the collection store test and shortened the comment.

Both replies note a small deviation from the suggested text, each easy to revert: `The same timing applies` became `The same fix applies` in the changeset, and the test comment drops the trailing `which is the delay behind` clause so the issue link sits on its own line without a trailing period.

The red check was re-run after the test change. With a single await it still fails against the pre-fix store with `AssertionError: expected [] to deeply equal [ 'first', 'second' ]`, and passes with the fix. Verification on the pushed state: 1959 Vitest tests, 31/31 in the focused store file, `pnpm tsc` clean, `pnpm lint-fix` clean.

Issue #3232, opened 2023-12-19
No focusable buttons when first toolbar item is disabled


jeryj:
### Current behavior

If the first button in a toolbar is disabled after rerender, it does not cause the `data-active-item` and roving tabindex to update, so the disabled button remains as the he `data-active-item` and the rest of the buttons have a `tabindex=-1`, so no button in the toolbar is able to be focused.

### Steps to reproduce the bug

1. Open sandbox: https://codesandbox.io/p/sandbox/inspiring-cherry-6xwpl2
2. Tab to the toolbar
3. Focus skips the toolbar, as the active item is still set as the now disabled button, which can't receive focus.

### Expected behavior

The toolbar will rerender and allow focus to be placed on the next available non-disabled button.

### Workaround

https://github.com/ariakit/ariakit/issues/3232#issuecomment-1863533666

diegohaz:
Thanks for reporting, @jeryj!

I believe this is because Ariakit doesn't automatically shift focus to another composite item when the current one becomes disabled or is removed from the DOM. This is because elements might be temporarily removed from the DOM due to React Suspense, virtualization, and so on. Plus, the active item id might point to an item that's not yet rendered, but will be soon. I'm planning to introduce a new feature to handle removable elements in the near future.

However, we could likely build something for disabled items directly into the composite item component.

In the meantime, here are two alternatives that can be implemented in userland:

-  Use the [`accessibleWhenDisabled`](https://ariakit.org/reference/toolbar-item#accessiblewhendisabled) prop on toolbar items.

-  Implement logic within a custom `ToolbarItem` component that automatically sets the active id when the current item is active and becomes disabled:
  
    ```jsx
    function ToolbarItem(props) {
      const toolbar = Ariakit.useToolbarContext();
      if (!toolbar) throw new Error("ToolbarItem must be used within Toolbar");
    
      const defaultId = React.useId();
      const id = props.id ?? defaultId;
      const isActive = toolbar.useState((state) => state.activeId === id);
    
      React.useEffect(() => {
        if (!props.disabled) return;
        if (!isActive) return;
        toolbar.setActiveId(toolbar.next());
      }, [props.disabled, isActive, toolbar]);
    
      return <Ariakit.ToolbarItem id={id} {...props} />;
    }
    ```

jeryj:
Thank you for the workaround ideas, @diegohaz! Those are super helpful.

Issue #1672, opened 2022-07-26
Removing a composite item makes the composite not focusable


marcospassos:
### Current behavior

Removing a composite item makes the whole composite becomes not focusable anymore

https://user-images.githubusercontent.com/943036/181102500-6eadd440-f93b-416c-be4f-531ef608ffcf.mov

### Steps to reproduce the bug

1. Open sandbox: https://codesandbox.io/s/epic-hertz-skqg7t?file=/src/style.css
2. Click to remove a composite item
3. Try to focus on it using tabs

### Expected behavior

Remains focusable.

diegohaz:
The problem is that it's hard for the library to know whether an item is being removed or just transitioning between mounted and unmounted states.

In your app, you control the business logic and know exactly when the item is being removed. In this case, you should explicitly move to another item:
```jsx
const onDelete = () => {
  composite.move(composite.next() || composite.previous());
  ...
}
```

I've tried to solve this generically several times in the past. In fact, the `Composite` component in Reakit (v1) does that automatically, but it fails in many use cases, like, for example, when using React Suspense.

marcospassos:
May I suggest adding it to the docs? That's fine handling this at the application level – I just didn't expect that.

diegohaz:
Leaving this here from my notes:

---

## Disposable

```jsx
const disposable = useDisposableStore(); // collection store
// Maybe store optional
<DisposableContainer store={disposable}>
  <Disposable>
    <button>button</button>
  </Disposable>
  <Disposable>
    <button>button</button>
  </Disposable>
  <Disposable>
    <button>button</button>
  </Disposable>
</DisposableContainer>
```

## CompositeDisposable

- Aware of `activeId` so can make a better decision when items are removed (like browser tabs)
- Or maybe `activeId` is already included in the disposable store, so it inherits from the composite store and we can use `Disposable` directly.

```jsx
const composite = useCompositeStore();
<Composite store={composite} role="tablist">
  <CompositeItem role="tab" render={<CompositeDisposable />}>
    Tab 1
  </CompositeItem>
  <CompositeItem role="tab" render={<CompositeDisposable />}>
    Tab 2
  </CompositeItem>
  <CompositeItem role="tab" render={<CompositeDisposable />}>
    Tab 3
  </CompositeItem>
</Composite>
```

PR #7362, opened 2026-08-31
Fix focus loss when the active composite item becomes disabled

### LINKED ISSUE #7359: Disabling the active composite item loses focus and breaks arrow-key navigation
### Current behavior

When a focused `CompositeItem` rendered as a native button becomes disabled, DOM focus moves to `document.body`, but the composite store's `activeId` still points to that item. Arrow keys then stop navigating the composite because keyboard events no longer reach it.

The existing fallback makes the enabled items tabbable, so Tab can recover focus. It does not preserve focus through the enabled-to-disabled transition. This is not a permanent keyboard trap.

Reproduced with published `@ariakit/react@0.4.38`, React `19.2.8`, and Chrome 151 (Playwright, macOS). Also reproduced against current `main` at [`dd1577592`](https://github.com/ariakit/ariakit/commit/dd1577592b837353720532fbe405bf67050fc219).

[Minimal reproduction on StackBlitz, including a workaround toggle](https://stackblitz.com/edit/lwmsotcm?file=App.jsx). The project uses pinned dependencies and a saved lockfile. No local setup is required.

1. Open the StackBlitz project in Chrome and wait for the preview to load.
2. Leave **Keep the disabled active item focusable** unchecked.
3. Click **Previous**.
4. Press **ArrowRight** to focus **Disable me**.
5. Press **Enter**. The button becomes disabled. Its focus indicator disappears, while the visible `activeId` remains `disable`.
6. Press **ArrowRight**. **Next** does not receive focus. DOM focus is on `body`.
7. Press **Tab**. **Next** receives focus, which confirms that the failure is focus loss during the transition, not a permanently unreachable widget.

The essential trigger is a composite item that disables itself while it has focus:

```jsx
<Ariakit.CompositeItem
  id="disable"
  disabled={disabled}
  onClick={() => setDisabled(true)}
>
  Disable me
</Ariakit.CompositeItem>
```

### Expected behavior

Disabling the active item should not drop focus outside the composite. The user should be able to press ArrowRight to move to the next enabled item without first using Tab or the pointer to recover focus.

The disabled item must remain exposed as disabled and must not activate.

### Workaround

Keep the disabled item focusable while `activeId` points to it:

```jsx
const activeId = Ariakit.useStoreState(composite, "activeId");

<Ariakit.CompositeItem
  id="disable"
  disabled={disabled}
  accessibleWhenDisabled={activeId === "disable"}
  onClick={() => setDisabled(true)}
>
  Disable me
</Ariakit.CompositeItem>
```

Select **Keep the disabled active item focusable** in the reproduction and repeat steps 3 through 6. This resets the disabled state and applies the conditional prop. Verified results:

- After Enter, **Disable me** retains DOM focus and has `aria-disabled="true"` without the native `disabled` attribute.
- ArrowRight moves focus to **Next**.
- Once `activeId` changes, **Disable me** becomes natively disabled. ArrowLeft skips it and focuses **Previous**.

Setting `accessibleWhenDisabled` unconditionally also keeps disabled items focusable, but the conditional workaround limits that behavior to the active item.

This workaround keeps the item focusable while it is the active item, which also keeps it focusable after focus leaves the composite entirely. A form that keys on DOM focus instead, and the limitations of both, are in https://github.com/ariakit/ariakit/issues/7359#issuecomment-5481066365.

### Possible solutions

Automatically apply `accessibleWhenDisabled` behavior when an item is disabled and `activeId` points to it. Conceptually:

```js
const effectiveAccessibleWhenDisabled =
  accessibleWhenDisabled || (disabled && isActiveItem);
```

Compute this during render, before calculating `trulyDisabled`, and pass the same effective value through to `useCommand`/`useFocusable`. An effect that runs after the native `disabled` attribute is applied is too late to prevent focus loss.

This approach preserves the current active item without selecting a replacement or making all disabled items permanently focusable. Keep `aria-disabled` and the existing activation guards in place. Once the item is inactive, restore normal disabled-item skipping unless the caller explicitly requested `accessibleWhenDisabled`.

Before shipping a fix, cover roving and virtual focus, controlled `activeId`, an initially disabled active item, and native form controls. Also define the behavior of an explicit `accessibleWhenDisabled={false}`, since the proposed invariant would override it while the item is active. Only the native-button, roving-focus path above has been browser-verified for this report.

Related: #1672 concerns removing the active item. This report concerns an item that remains mounted and becomes disabled.

diegohaz:
## Motivation

When a focused [`CompositeItem`](https://ariakit.com/reference/composite-item) becomes disabled, it is removed from the focus order right away: through the native `disabled` attribute when it renders a button or input, and through its `tabindex` when it renders anything else. The browser then moves focus to `document.body`, while the composite store still points [`activeId`](https://ariakit.com/reference/composite-provider#activeid) at that item. Arrow keys no longer reach the composite, so keyboard navigation stops until the user presses Tab to get back in.

This is the essential trigger, an item that disables itself while it has focus:

```jsx
<Ariakit.CompositeItem
  disabled={disabled}
  onClick={() => setDisabled(true)}
>
  Disable me
</Ariakit.CompositeItem>
```

Fixes #7359.

## Solution

This pull request lands in stages so CI records the failing reproduction first.

The first commit adds the `composite-item-disabled-active` sandbox with Playwright and happy-dom coverage. Three tests fail on `main`: roving focus, a controlled [`activeId`](https://ariakit.com/reference/composite-provider#activeid), and a native form control. Three more are guards that must stay green: virtual focus, an explicit `accessibleWhenDisabled={false}` opt out, and an active item that is disabled before it ever receives focus (#3232). The final commit adds two more: one for an item that is not natively focusable and therefore loses its `tabindex` rather than gaining the native attribute, and one happy-dom test pinning that the option stays inheritable from a composing [`Focusable`](https://ariakit.com/reference/focusable).

The final commit keeps a composite item accessible while it holds DOM focus, so the native `disabled` attribute is only applied once focus has moved elsewhere. The item stays exposed as disabled through `aria-disabled` and still refuses to activate.

`useCompositeItem` now tracks whether the item holds DOM focus and derives the option from it:

```ts
const accessibleWhenDisabled =
  props.accessibleWhenDisabled ?? (disabled && focused ? true : undefined);
```

Keying on DOM focus rather than on the active item is what preserves #3232: an item that is the active item but was never focused stays natively disabled and keeps being skipped. Falling back to `undefined` rather than `false` keeps the option inheritable from a composing [`Focusable`](https://ariakit.com/reference/focusable), and an explicit `accessibleWhenDisabled={false}` still wins.

The flag is set only under roving focus, since [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) leaves DOM focus on the composite element, and only when the item itself is the focus target. It is released on blur unless the item still holds DOM focus, because Firefox fires `focusout` when the window loses focus while the element is still focused. Without that guard the native attribute would land on the focused element and drop focus to the body once the window came back. `test-firefox.ts` covers it, and fails without the guard.

## Workaround

Until this ships, keep the item accessible while it holds DOM focus, so the native `disabled` attribute is never applied to the element the user is currently on.

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7359 is fixed.
function SelfDisablingItem(props: Ariakit.CompositeItemProps) {
  const [focused, setFocused] = useState(false);
  return (
    <Ariakit.CompositeItem
      {...props}
      accessibleWhenDisabled={focused}
      onFocus={(event) => {
        props.onFocus?.(event);
        setFocused(true);
      }}
      onBlur={(event) => {
        props.onBlur?.(event);
        const { currentTarget } = event;
        // Firefox fires focusout when the window loses focus while the element
        // still holds DOM focus. Releasing the workaround then would drop focus
        // to the body once the window comes back.
        if (currentTarget.contains(currentTarget.ownerDocument.activeElement)) {
          return;
        }
        setFocused(false);
      }}
    />
  );
}
```

The item keeps DOM focus, stays exposed as disabled through `aria-disabled`, and still refuses to activate. Once focus moves away it becomes natively disabled again, arrow keys skip it, and the composite stays reachable with Tab.

This workaround is narrower than the fix in this pull request, in three ways.

Apply it only under roving focus. Under [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus) the composite element keeps DOM focus, so nothing is lost. Applying it there would also latch the flag on, because [`CompositeItem`](https://ariakit.com/reference/composite-item) cancels the intermediate blur that follows the virtual focus handoff.

Do not apply it to components that already default `accessibleWhenDisabled` to `true`, such as [`Tab`](https://ariakit.com/reference/tab) and [`FormSubmit`](https://ariakit.com/reference/form-submit). Passing an explicit `false` while the item is unfocused would override that default and make their disabled items natively disabled.

Do not key it on [`activeId`](https://ariakit.com/reference/composite-provider#activeid) instead of DOM focus. `activeId` keeps pointing at the item after focus leaves the composite, so the item would stay natively enabled and become the composite's only tab stop, which is the #3232 behavior this pull request preserves.

The helper owns `accessibleWhenDisabled`, `onFocus` and `onBlur`. It forwards a caller's `onFocus` and `onBlur`, but a caller cannot pass its own `accessibleWhenDisabled` through it.

The Firefox behavior above is reproducible with Playwright: focus the item, disable it, then open a real popup with `window.open`, bring the popup to front, close it, and bring the page back to front. Firefox logs exactly one `focusout` on the item with `relatedTarget === null` while `document.activeElement` is still that item; Chromium logs none. Note that `document.hasFocus()` stays `true` in Playwright, so it cannot be used to detect this.

## Review gate reassessment

<details>
<summary>Library fix cycle 3: Narrow the contract to items that themselves hold focus</summary>

**Assessment**

The core of the fix is small and proportionate: one state flag, one derived option, one release handler. Layered on it was a separable capability, support for items that hold focusable descendants, consisting of an acquire-side hoist, an `isFocusEventOutside` release guard, a `contains` generalization, a fixture section and three tests.

Four of the nine findings on this track concerned only that capability. The two open ones were decisive. The first: the `contains` generalization had no regression coverage, and the evidence originally recorded for it was overstated, since a later measurement showed a self-correcting transient rather than a persistent state. The second, and more serious: Firefox fires no `focusout` when a focused element is removed from the DOM, so a container item whose focused descendant unmounts latched the flag on with no event able to clear it, leaving a disabled item registered as enabled and permanently in the arrow key rotation. That is a regression this pull request would have introduced, and repairing it needed a document level listener or an observer.

**Decision**

Narrow the supported contract to items that themselves hold DOM focus. The acquire side is back under the self target check, the release keeps only the guard that covers Firefox's window blur, and the descendant machinery, its fixture section and its three tests are removed.

This introduces no regression: a disabled container behaves exactly as it does on `main`. It removes both open findings by removing their cause rather than repairing it, and the reported bug stays fully fixed.

**Corrections to the record**

An earlier version of this entry justified the narrowing partly on the claim that components rendering a `div` cannot exhibit this issue, because they never receive the native `disabled` attribute. **That claim is false**, and a later review disproved it with browser probes. `getTabIndex` returns `undefined` for a truly disabled element that is not natively tabbable, so React removes the `tabindex` attribute, and a focused `div` then loses focus to the body in Chromium and WebKit. Components such as [`MenuItem`](https://ariakit.com/reference/menu-item) do exhibit this issue, by that second mechanism.

The narrowing still holds, because it removed support for focus held by a *descendant*, which is orthogonal to the element type of the item itself. The fix is element agnostic, so a `div` rendered item that holds focus is fixed exactly as a button is. A `NonNativeComposite` fixture section and a test in each suite now cover that path; the browser test fails in all three engines without the fix.

An earlier entry also described a container "going inert while the user is typing" without the `contains` generalization. A later measurement showed a self-correcting transient of a few milliseconds. Under this decision the point is moot.

**Intentionally unsupported**

- Composite items that hold focusable descendants, such as `CompositeContainer` and [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container). Focus landing on a descendant does not mark the item as holding focus, so such an item becomes truly disabled while its descendant holds focus, exactly as on `main`. An item that holds focus itself is supported regardless of what it renders.
- A focusable descendant inside a `button` shaped item, which is invalid HTML.

</details>

<details>
<summary>Cycle 3: Continue the current test-only reproduction direction</summary>

**Assessment**

Issue severity is medium. Disabling a focused composite item drops DOM focus to `document.body` and stops arrow key navigation, which is a real keyboard and screen reader defect in a core primitive. It is recoverable, because the existing tabbable fallback lets Tab re-enter the composite, so it is not a permanent keyboard trap.

Expected frequency is medium to high. The trigger is the ordinary pattern of an action that disables itself on activation, such as mark as read, apply, submit, or undo becoming unavailable.

Supported scope is a stable public API. The bug requires roving focus. Virtual focus is unaffected, because DOM focus never rests on the item. Corrected: this entry originally added "and an element that supports the native `disabled` attribute", which a later review disproved. See the corrections in the library fix entry above.

Likely user impact is that keyboard and screen reader users lose their place mid widget and are moved to `document.body` without warning. Pointer users are unaffected.

Implementation and maintenance complexity is low. The reviewed state is test only: one sandbox directory, three files, six fixture sections, and no library change, helper, abstraction, public contract, or platform policy. The six sections are not invented scope. They are exactly the six scenarios the issue requires before shipping: roving focus, virtual focus, controlled `activeId`, an initially disabled active item, native form controls, and an explicit `accessibleWhenDisabled={false}`.

Four of the nine findings so far share one root cause: proving a transient, engine scheduled, negative state, namely focus that must not move, across three browser engines plus a DOM shim that does not model the mechanism. That root cause is intrinsic to the bug's surface rather than to the chosen design. The remaining findings are unrelated singletons about coverage completeness, coverage organization, base branch hygiene, and a documentation deferral. No reviewer in any round challenged the fixture set, the sandbox choice, the browser project selection, or the eventual fix shape, and the trend is convergent: four findings, then four, then one, with top severity falling from High to Medium to Low.

**Decision**

Continue the current design. There is no accreted machinery to simplify away, the only simplification finding raised in three rounds removed a single line, and reverting or narrowing would mean deleting coverage the issue explicitly requires before shipping.

Constraint carried into the library fix: the sketch in the issue, `accessibleWhenDisabled || (disabled && isActiveItem)`, would regress #3232. Items register into the collection as `disabled: trulyDisabled`, and `isTabbable` makes every item tabbable only while the active item is registered as disabled. Making a merely active disabled item accessible would turn an initially disabled active item into the only tabbable item, so Tab would land on it instead of skipping it. The fix must therefore key on the item actually holding DOM focus, not on it being the active item. `InitiallyDisabledComposite` polices this.

Also carried into the library fix: an explicit `accessibleWhenDisabled={false}` opts out of the new behavior, and that must be stated in the [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) documentation and in the changeset.

**Intentionally unsupported**

- Items rendered as `a href` or `summary`, where a disabled item receives `tabindex="-1"` rather than losing the attribute, so focus is not lost. Corrected: this entry originally listed `div` here, and a `div` rendered item does lose focus. See the corrections in the library fix entry above.
- Items with `focusable={false}`, which are never treated as disabled by [`Focusable`](https://ariakit.com/reference/focusable).
- Mobile projects. This is a desktop keyboard focus contract.

No follow-up issues were registered, because every finding was dispositioned as Address now within this pull request.

</details>

<details>
<summary>Cycle 6: Continue, and decline further comment-only findings</summary>

**Assessment**

Issue attributes are unchanged from cycle 3, with one correction: [`Tab`](https://ariakit.com/reference/tab) is not affected, because it already defaults `accessibleWhenDisabled` to `true`. The live surface is roving focus composites whose items do not default that prop.

Complexity is unchanged and remains test only. The staged diff grew from 315 to 422 insertions between the first reviewed state and now, and every addition traces to an accepted finding: one fixture section and its sibling button, one test per file, four `flushFrames` gates with their comments, and the assertions that pin the observable outputs of `trulyDisabled`. No abstraction was introduced, and the only simplification finding raised in six rounds removed a single line.

Twelve findings across six rounds split evenly between substantive and editorial. The substantive ones have two root causes, both now closed. The first is assertions of state that must not change, placed without crossing a boundary, which is closed by the uniform `aria-disabled` gate plus `flushFrames` at all four sites. The second is that the reproduction was written before the fix's observable shape was decided, which is closed and provably exhausted: `trulyDisabled` drives exactly four render outputs in `Focusable`, the roving test now pins three, and the fourth is not independently observable for a button because `useCompositeItem` supplies its own tab index. The collection registration is already pinned by the arrow key navigation assertions in the same test.

The cycle 3 prediction that severity would fall monotonically to comments only was wrong: rounds 4 and 5 each produced a medium finding with real power to discriminate between candidate fixes, and those are the second and third most valuable findings in the track. Continuing past cycle 3 was correct. That justification has now expired, because the substantive class is exhausted while the editorial class is self sustaining: the round 6 finding critiques the wording of a comment that the round 1 fix introduced, and every reword creates fresh comment surface.

**Decision**

Continue the current design. There is no machinery to remove, nothing to revert to, and stage 1 adds no contract to narrow.

From round 7 onward, a finding whose entire content is the wording, placement, or completeness of an explanatory comment, where no assertion, fixture section, or coverage dimension changes, is a Declined non-goal. The one carve out is a comment that states something factually false in a way that would steer the library fix wrong. This weakens no correctness, safety, security, acceptance, or supported behavior requirement, since comment phrasing touches none of them.

**Carried into the library fix**

1. The sketch in the issue, `accessibleWhenDisabled || (disabled && isActiveItem)`, would regress #3232, so the fix must key on the item actually holding DOM focus. The virtual focus and initially disabled guards both discriminate this.
2. While an item holds DOM focus it becomes fully accessible when disabled, so the native attribute, `data-truly-disabled` and the injected `pointer-events: none` all follow together. [`Hovercard`](https://ariakit.com/reference/hovercard) reads `data-truly-disabled` to decide whether a tooltip may open.
3. An explicit `accessibleWhenDisabled={false}` opts out, and that must be stated in the prop documentation and in the changeset.
4. Pre-existing on `main`: `useCompositeItem` computes `trulyDisabled` from the raw prop while [`Focusable`](https://ariakit.com/reference/focusable) resolves the same option through its context channel first, so the two can disagree when the value arrives from a parent. Verified pre-existing and triggerable in an isolated worktree on `main`, and registered as #7364. This pull request neither introduces nor worsens it, so it ships separately.
5. During `focusout`, `document.activeElement` is `body` in Chromium, Firefox and WebKit even when focus moves to a descendant, so a `contains(activeElement)` check cannot identify a descendant move. Superseded by the cycle 3 narrowing: focus held by a descendant is no longer supported, so the fix carries only the `activeElement` check that covers the Firefox window blur `focusout`, and `isFocusEventOutside` is not used.

**Also intentionally unsupported**

- [`Tab`](https://ariakit.com/reference/tab), which is immune by default.
- Grid and [`CompositeRow`](https://ariakit.com/reference/composite-row) composites, which the issue does not name.
- Removal of the active item, which the issue scopes out in favor of #1672.

</details>

### DIFF
--- a/.changeset/300-composite-item-disabled-focus.md
+++ b/.changeset/300-composite-item-disabled-focus.md
@@ -1,0 +1,20 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Focus preserved when a composite item becomes disabled
+
+A [`CompositeItem`](https://ariakit.com/reference/composite-item) that becomes [`disabled`](https://ariakit.com/reference/focusable#disabled) while it holds focus now stays focusable until focus moves elsewhere. Previously the focused element was removed from the focus order right away, through the native `disabled` attribute on elements that support it, such as a button or input, and through the loss of `tabindex` on elements that support neither native disabling nor native tabbing, such as a `div`. The browser then moved focus to the body and arrow keys stopped navigating the composite.
+
+```tsx
+<CompositeItem disabled={read} onClick={() => setRead(true)}>
+  Mark as read
+</CompositeItem>
+```
+
+The item is still exposed as disabled through `aria-disabled` and still cannot be activated. Once focus moves away it becomes fully disabled again and arrow keys skip it.
+
+This applies to all components built on [`CompositeItem`](https://ariakit.com/reference/composite-item) that use roving focus, including [`ToolbarItem`](https://ariakit.com/reference/toolbar-item), [`ToolbarContainer`](https://ariakit.com/reference/toolbar-container), [`MenuItem`](https://ariakit.com/reference/menu-item), [`MenuItemCheckbox`](https://ariakit.com/reference/menu-item-checkbox), [`MenuItemRadio`](https://ariakit.com/reference/menu-item-radio), [`Radio`](https://ariakit.com/reference/radio), and [`FormRadio`](https://ariakit.com/reference/form-radio).
+
+[`ComboboxItem`](https://ariakit.com/reference/combobox-item) and [`SelectItem`](https://ariakit.com/reference/select-item) are affected only when the composite opts out of [`virtualFocus`](https://ariakit.com/reference/composite-provider#virtualfocus), which they enable by default, since virtual focus keeps DOM focus on the composite element rather than the item. [`Tab`](https://ariakit.com/reference/tab) is unaffected, since it already keeps disabled tabs focusable by default. To opt out and keep the previous behavior, set [`accessibleWhenDisabled`](https://ariakit.com/reference/focusable#accessiblewhendisabled) to `false`.
--- a/app/src/sandbox/composite-item-disabled-active/index.react.tsx
+++ b/app/src/sandbox/composite-item-disabled-active/index.react.tsx
@@ -1,15 +1,47 @@
 import * as Ariakit from "@ariakit/react";
 import { useId, useState } from "react";
 
+// TODO: Remove this workaround once
+// https://github.com/ariakit/ariakit/issues/7359 is fixed. Keeping the item
+// accessible while it holds DOM focus stops the native `disabled` attribute
+// from moving focus to the body. It is deliberately not applied to the virtual
+// focus, opt out, and initially disabled sections, which must keep their
+// current behavior.
+function SelfDisablingItem(props: Ariakit.CompositeItemProps) {
+  const [focused, setFocused] = useState(false);
+  return (
+    <Ariakit.CompositeItem
+      {...props}
+      accessibleWhenDisabled={focused}
+      onFocus={(event) => {
+        props.onFocus?.(event);
+        setFocused(true);
+      }}
+      onBlur={(event) => {
+        props.onBlur?.(event);
+        const { currentTarget } = event;
+        // Firefox fires focusout when the window loses focus while the element
+        // still holds DOM focus. Releasing the workaround then would apply the
+        // native attribute to the focused element and drop focus to the body
+        // once the window comes back.
+        if (currentTarget.contains(currentTarget.ownerDocument.activeElement)) {
+          return;
+        }
+        setFocused(false);
+      }}
+    />
+  );
+}
+
 function RovingComposite() {
   const [read, setRead] = useState(false);
   return (
     <Ariakit.CompositeProvider>
       <Ariakit.Composite role="toolbar" aria-label="Roving message actions">
         <Ariakit.CompositeItem>Roving reply</Ariakit.CompositeItem>
-        <Ariakit.CompositeItem disabled={read} onClick={() => setRead(true)}>
+        <SelfDisablingItem disabled={read} onClick={() => setRead(true)}>
           Roving mark as read
-        </Ariakit.CompositeItem>
+        </SelfDisablingItem>
         <Ariakit.CompositeItem>Roving archive</Ariakit.CompositeItem>
       </Ariakit.Composite>
     </Ariakit.CompositeProvider>
@@ -38,9 +70,9 @@ function ControlledComposite() {
     <Ariakit.CompositeProvider activeId={activeId} setActiveId={setActiveId}>
       <Ariakit.Composite role="toolbar" aria-label="Controlled message actions">
         <Ariakit.CompositeItem>Controlled reply</Ariakit.CompositeItem>
-        <Ariakit.CompositeItem disabled={read} onClick={() => setRead(true)}>
+        <SelfDisablingItem disabled={read} onClick={() => setRead(true)}>
           Controlled mark as read
-        </Ariakit.CompositeItem>
+        </SelfDisablingItem>
         <Ariakit.CompositeItem>Controlled archive</Ariakit.CompositeItem>
       </Ariakit.Composite>
     </Ariakit.CompositeProvider>
@@ -53,7 +85,7 @@ function NativeControlComposite() {
     <Ariakit.CompositeProvider>
       <Ariakit.Composite role="toolbar" aria-label="Native message actions">
         <Ariakit.CompositeItem>Native reply</Ariakit.CompositeItem>
-        <Ariakit.CompositeItem
+        <SelfDisablingItem
           render={<input type="checkbox" />}
           aria-label="Native mark as read"
           disabled={locked}
--- a/app/src/sandbox/composite-item-disabled-active/index.react.tsx
+++ b/app/src/sandbox/composite-item-disabled-active/index.react.tsx
@@ -1,0 +1,131 @@
+import * as Ariakit from "@ariakit/react";
+import { useId, useState } from "react";
+
+function RovingComposite() {
+  const [read, setRead] = useState(false);
+  return (
+    <Ariakit.CompositeProvider>
+      <Ariakit.Composite role="toolbar" aria-label="Roving message actions">
+        <Ariakit.CompositeItem>Roving reply</Ariakit.CompositeItem>
+        <Ariakit.CompositeItem disabled={read} onClick={() => setRead(true)}>
+          Roving mark as read
+        </Ariakit.CompositeItem>
+        <Ariakit.CompositeItem>Roving archive</Ariakit.CompositeItem>
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
+
+function VirtualFocusComposite() {
+  const [read, setRead] = useState(false);
+  return (
+    <Ariakit.CompositeProvider virtualFocus>
+      <Ariakit.Composite role="toolbar" aria-label="Virtual message actions">
+        <Ariakit.CompositeItem>Virtual reply</Ariakit.CompositeItem>
+        <Ariakit.CompositeItem disabled={read} onClick={() => setRead(true)}>
+          Virtual mark as read
+        </Ariakit.CompositeItem>
+        <Ariakit.CompositeItem>Virtual archive</Ariakit.CompositeItem>
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
+
+function ControlledComposite() {
+  const [activeId, setActiveId] = useState<string | null | undefined>();
+  const [read, setRead] = useState(false);
+  return (
+    <Ariakit.CompositeProvider activeId={activeId} setActiveId={setActiveId}>
+      <Ariakit.Composite role="toolbar" aria-label="Controlled message actions">
+        <Ariakit.CompositeItem>Controlled reply</Ariakit.CompositeItem>
+        <Ariakit.CompositeItem disabled={read} onClick={() => setRead(true)}>
+          Controlled mark as read
+        </Ariakit.CompositeItem>
+        <Ariakit.CompositeItem>Controlled archive</Ariakit.CompositeItem>
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
+
+function NativeControlComposite() {
+  const [locked, setLocked] = useState(false);
+  return (
+    <Ariakit.CompositeProvider>
+      <Ariakit.Composite role="toolbar" aria-label="Native message actions">
+        <Ariakit.CompositeItem>Native reply</Ariakit.CompositeItem>
+        <Ariakit.CompositeItem
+          render={<input type="checkbox" />}
+          aria-label="Native mark as read"
+          disabled={locked}
+          onChange={() => setLocked(true)}
+        />
+        <Ariakit.CompositeItem>Native archive</Ariakit.CompositeItem>
+      </Ariakit.Composite>
+    </Ariakit.CompositeProvider>
+  );
+}
+
+// An explicit `accessibleWhenDisabled={false}` opts out of keeping the item
+// reachable, so this item becomes natively disabled even while it holds focus.
+function OptOutComposite() {
+  const [read, setRead] = useState(false);
+  return (
+    <>
+      <button type="button" tabIndex={0}>
+        Before opt out
+      </button>
+      <Ariakit.CompositeProvider>
+        <Ariakit.Composite role="toolbar" aria-label="Opt out message actions">
+          <Ariakit.CompositeItem>Opt out reply</Ariakit.CompositeItem>
+          <Ariakit.CompositeItem
+            disabled={read}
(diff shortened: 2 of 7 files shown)


diegohaz:
Closing without merging.

The behavior this pull request changes matches the platform. A native `<button>` that disables itself also drops focus to `document.body`, in Chromium, Firefox, and WebKit alike, and every one of those engines keeps the sequential focus navigation starting point, so <kbd>Tab</kbd> resumes at the next item instead of at the top of the document. Ariakit already ships the part that makes that recovery land on the next composite item (#3232). The full evidence and reasoning are in #7359.

The implementation here worked and was covered across all three engines, but it tracked focus in React state on every composite item, and that state flips twice per arrow keypress. The performance report above measures the result on the hottest navigation path of every roving-focus composite: +17% scripting on `tooltip-perf > move across items without tooltips`, which is pure `ToolbarItem` arrow navigation with no tooltips involved, +31% on the same fixture with tooltips, +11% on `menu-perf > move across items`, and +10% on `composite-perf > move across items`. Virtual-focus composites stayed flat, because the tracking was gated on roving focus, which points at the state flip rather than at measurement noise.

Trading that for a recoverable state that already matches native behavior is not a good deal.

The branch `fix-composite-item-disabled-focus` stays available if this is ever revisited. It carries a reproduction sandbox at `app/src/sandbox/composite-item-disabled-active/` with happy-dom and Playwright coverage for roving and virtual focus, controlled `activeId`, native and non-native items, inherited and explicitly opted-out `accessibleWhenDisabled`, an initially disabled active item, and a Firefox window-blur round trip.

#7364, filed while working on this, is unrelated to the decision here and stays open.

diegohaz:
Correction to my closing comment above. I wrote that Chromium, Firefox, and WebKit all keep the sequential focus navigation starting point, so <kbd>Tab</kbd> resumes at the next item. That is wrong for Safari.

Real Safari discards the starting point when the focused element becomes disabled, so <kbd>Tab</kbd> restarts from the beginning of the document rather than continuing after the item. Playwright's WebKit does not reproduce this, which is where the error came from. Verified in Safari 26.6.2 through `safaridriver`, on plain HTML with no Ariakit involved, so it is browser behavior rather than something this pull request would have been fixing.

Full measurements and mechanism: https://github.com/ariakit/ariakit/issues/7359#issuecomment-5486819865

The decision to close is unchanged, and the performance measurements above are unaffected. The Safari cost is simply higher than I stated.

Issue #7363, opened 2026-08-31
Composite composite={false} documentation overstates what it disables


diegohaz:
### Context

Found while working on #7360. Not a blocker for that fix; filing so the observation has a durable record.

### Current behavior

The [`composite`](https://ariakit.org/reference/composite#composite) prop's JSDoc says:

> If disabled, this component will stop managing focus and keyboard navigation for its items and itself.

The "for its items" half is not accurate for keyboard navigation. With `composite={false}`, arrow keys pressed on an item still move the active item:

- [`CompositeItem`'s `onKeyDown`](https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-components/src/composite/composite-item.tsx) calls `store.move(nextId)` and is wired unconditionally. It has no reference to the parent `Composite`'s `composite` prop.
- The container's own `onKeyDown` in `composite.tsx` is also wired unconditionally, although it requires `isSelfTarget`, and `composite={false}` makes the container unfocusable, so in practice it cannot be the target.
- Only `CompositeFocusOnMove` is gated on `composite`.

Observed in Chromium on a plain toolbar composite: with `composite={false}`, `ArrowRight` on an item moves `data-active-item` and the roving `tabindex` from one item to the next, while DOM focus stays put. So the component stops moving **focus**, but item-level active-item navigation continues.

### Why it seems worth recording

While writing the regression fixture for #7360, a comment describing the prop was based on this JSDoc wording rather than on the implementation, and had to be corrected during review. The sentence reads naturally as "arrow keys do nothing here", which is not what happens.

### Expected behavior

Either is reasonable, and which one is the right call is a maintainer decision:

1. Tighten the JSDoc so it describes what the component stops doing (moving focus, and applying composite ARIA attributes), without claiming item keyboard navigation stops. The surrounding sentence, "These responsibilities should be taken over by another composite component", already reads as prescriptive rather than descriptive.
2. Or, if the documented behavior is the intended one, change the implementation so item navigation is also disabled while the prop is off.

Option 1 looks more likely to be correct, since the items belong to a store that the taking-over composite drives, and that composite needs those moves to keep working.

### Workaround

https://github.com/ariakit/ariakit/issues/7363#issuecomment-5478792701

PR #7368, opened 2026-08-31
Fix Composite remount replaying an old focus request

### LINKED ISSUE #7360: Composite remount replays an old focus request and scrolls the page
### Current behavior

Remounting a `Composite` can move focus to its container and scroll the page after a previous item move, even though no new focus move was requested. The store stays mounted while the composite is removed and added again.

The standalone reproduction uses published `@ariakit/react@0.4.38` (`@ariakit/react-components@0.5.0`) and `react@19.2.8`. It uses no Menu or Popover components. In Chromium 151.0.7922.34, clicking **Show toolbar** moves focus to the toolbar container and changes the page scroll from 1119px to 329px. If the earlier ArrowRight step is skipped, the button keeps focus and the scroll stays at 1119px. Both focus outcomes were also verified in the saved StackBlitz preview after reloading the project.

Recording of the standalone reproduction: ArrowRight, hide, scroll down, and show. The toolbar takes focus and the page scrolls up. Captured at 1280 × 800; the scroll values depend on viewport size.

https://github.com/user-attachments/assets/a7c5a895-933b-4f72-97d1-a09279812f7a

The current `menubar-navigation` example has the same failure on [`dd15775`](https://github.com/ariakit/ariakit/commit/dd1577592b837353720532fbe405bf67050fc219): focus [Services in the example](https://ariakit.org/examples/menubar-navigation), press ArrowDown to Web Development, ArrowLeft to Contact, then ArrowRight. Services briefly receives focus, then the remounted menu container takes focus. On a below-fold host page, this also scrolls to the page top before the menu is positioned.

1. Open [the StackBlitz reproduction](https://stackblitz.com/edit/wuygy9iq?file=src%2FApp.jsx) and wait for its preview to load.
2. Click **First item**, then press **ArrowRight**. Second item receives focus.
3. Click **Hide toolbar**. This calls `store.setActiveId(null)` and removes the composite while keeping its store.
4. Scroll down and click **Show toolbar**.
5. Observe that the toolbar container takes focus and the page scrolls up.

Control: reload and repeat without pressing ArrowRight. **Show toolbar** keeps focus and the scroll does not change.

### Expected behavior

**Show toolbar** should keep focus. Adding the composite back should not create a new focus or scroll request.

These operations have different meanings:

```tsx
store.move(null);        // Request focus on the composite container.
store.setActiveId(null); // Clear the active item without moving focus.
```

A previous move must not turn a later active-item reset into a new container-focus request.

### Workaround

Keep the composite mounted and hide it instead of removing it: https://github.com/ariakit/ariakit/issues/7360#issuecomment-5476506121

Or clear the recorded move count once the composite goes away, which keeps it unmounting. Assumptions and limitations are listed there: https://github.com/ariakit/ariakit/issues/7360#issuecomment-5480816118

### Possible solutions

The current `Composite` container-focus effect runs when `moves` or `compositeElement` changes. It checks for a nonzero `moves` counter and `activeId === null`, then scrolls and focuses the container. A remount can meet those conditions using an old move count and a later active-item reset. See the [container-focus effect](https://github.com/ariakit/ariakit/blob/dd1577592b837353720532fbe405bf67050fc219/packages/ariakit-react-components/src/composite/composite.tsx#L171-L204) and [`move` versus `setActiveId`](https://github.com/ariakit/ariakit/blob/dd1577592b837353720532fbe405bf67050fc219/packages/ariakit-components/src/composite/composite-store.ts#L559-L564).

Track the intent and lifetime of an actual move request so a completed request is not repeated when the container changes. Keep this repair in the existing Composite behavior; it does not depend on a new component family.

Preserve explicit `store.move(null)` calls, including a request made while `composite={false}` followed by `composite={true}`. The [existing source comment](https://github.com/ariakit/ariakit/blob/dd1577592b837353720532fbe405bf67050fc219/packages/ariakit-react-components/src/composite/composite.tsx#L142-L151) records why the effect waits for the composite element. Removing that dependency alone could break that case.

Add browser coverage for the standalone remount case and the reported menubar sequence, with focus and below-fold scroll assertions. Include repeated remounts and the explicit delayed `move(null)` case. Only Chromium was verified in this investigation; other engines still need checks.

diegohaz:
## Motivation

Removing a [`Composite`](https://ariakit.com/reference/composite) from the tree and adding it back moves focus to its own container and scrolls the page to it, although nothing asked for a new focus move.

These two calls mean different things:

```tsx
store.move(null); // Request focus on the composite container.
store.setActiveId(null); // Point the active item at the container, asking for no focus.
```

`move()` increments a `moves` counter that `setActiveId()` leaves alone. `CompositeFocusOnMove` keys two effects off that counter, so a fresh instance of it reads a count left behind by a move that was already carried out and acts on that count again. A fresh instance appears in two ways: the whole `Composite` remounting, and the `composite` prop flipping from `false` to `true`, which remounts only that child while the DOM element survives.

Both effects replay, and which one does depends on the active item:

- with `activeId === null`, the container takes focus and the page scrolls to it;
- with an item still active, that item takes focus and the page scrolls to it instead.

The sequence reported against the `menubar-navigation` example is arrow into a menu, arrow out to a sibling item, then arrow back. The shared menu unmounts on the way and comes back with the earlier move still recorded and a null active item, so its container steals focus from the menubar item.

## Solution

The commits land in the order the components issue-fix workflow requires: the failing reproduction first, then a userland workaround, then the library fix.

### Reproduction

`app/src/sandbox/composite-7360/` is a formatting toolbar at the end of a long document, with the store owned by the page so it outlives every mount of the toolbar. Collapsing the toolbar points its active item back at the container, the way an app does for a widget that should come back with no item of its own selected. The toolbar is rendered last so that removing and adding it changes layout only below the controls, which keeps the browser's own scroll anchoring out of the measurements. A highlight tool that loads on demand lets the fixture ask for a tool before it exists.

Twelve browser tests run in Chromium, Firefox and WebKit, with a happy-dom duplicate of each for the focus half:

| Test | Pins |
| --- | --- |
| Untouched toolbar comes back | Harness control: mounting alone moves nothing |
| Untouched composite behavior comes back | Harness control: the `composite` flip alone moves nothing |
| Toolbar comes back after a move | The reported container replay |
| Repeated remounts | The replay does not accumulate |
| Composite behavior comes back | The sibling item replay, on a remount the element survives |
| Composite behavior comes back after an abandoned request | A request the toolbar gave up on is spent, not left behind |
| Newly moved item comes back | A move recorded while handed off must still be carried out |
| Focus command while expanded | `move(null)` still focuses and scrolls to the container |
| Focus command while collapsed | A move with no composite element stays pending |
| Focus command while handed off | The same, across `composite={false}` to `composite={true}` |
| Tool loads after being asked for | A move waiting on its item is carried out when the item arrives |
| Tool loads after composite behavior comes back | That pending move survives the instance being replaced |
| Menubar sequence | The reported `menubar-navigation` reproduction |

Four of them fail on `main` in all three engines, along with the menubar test. The rest pass and must keep passing: together they are what stops a fix from suppressing the effects wholesale.

`app/src/sandbox/composite-superseded-move/` covers the one state the toolbar cannot express. Taking a pending request over needs a composite element that can take focus while an item is active, and a roving-tabindex toolbar's container is focusable only when no item is, exactly when that branch does not run. The palette there carries `tabIndex`, so clicking its own padding focuses it and replaces the pending request with its own.

### The fix

`CompositeFocusOnMove` now records, per store, which of its instances consumed the current move, and refuses the request to any other instance. A `sync` on `moves` clears that record, so every new request and every reset of the counter starts unconsumed.

The record has to outlive the component, because the count alone cannot tell a request that is still waiting from one an earlier instance already acted on. It is keyed on the **component instance** rather than on the composite element: switching `composite` off and on again mounts a fresh instance while the element itself survives, so an element-keyed record would be a no-op on exactly that path. A ref token survives the same instance's effect re-runs, including StrictMode's double invocation, but not a remount, which is the distinction the request needs.

What the record tracks is **consumption**, not merely reading the request. A move has three outcomes, and only the last is handed on:

- **Carried out.** Focus reached the item, or the container branch focused the container. Spent, so a later instance must not repeat it.
- **Given up on.** The popup closed, the user moved to another item, focus left the composite, or a new presentation replaced this one. Also spent: the request is dead, not owed.
- **Still waiting.** The target item has not registered yet, or a closed popup has not opened. Nothing consumed it, so whichever instance is around when the wait ends carries it out.

Recording on read collapses the last two into the first and drops pending requests. Recording only on carry-out collapses the middle one into the last and replays abandoned requests. Both were caught by review on this branch, and each of the three states now has a test that fails against a version of the fix that collapses it into another.

`presentItem` reports consumption through a new `@private` `onConsume`, fired where it delivers focus and at every point where it gives up. Its returned `cancel` carries a `settle` alongside it, so `usePresentItem` can retire a live presentation it is replacing, while an owner cancelling one on unmount still hands the request on.

`consumeMove` also takes the `moves` value its effect ran with and drops a consumption that arrives after the count has moved on, so a focus handler that starts a new request does not inherit this one's consumption. That guard ships without a red check: the window needs a store write to re-enter a live presentation between a `moves` change and React flushing the effect whose cleanup would cancel it, and no reachable consumer path was found. It compares the count by value, so it does not cover a reset that returns to the same number.

## Preview

Recorded against `app/src/sandbox/composite-7360/` at 1280 × 800: click the first toolbar item, press ArrowRight, return to the controls at the top of the document, collapse the toolbar, then expand it again.

Before, the toolbar takes focus and the page jumps from 0px to 1580px to reach it.

[before.webm](https://github.com/user-attachments/assets/ce930309-66a5-4546-ace2-f12242e8ed04)

After, focus stays on the control that was clicked and the page stays at 0px.

[after.webm](https://github.com/user-attachments/assets/7ff41ba8-30c0-496e-a16b-0a51396aa9d0)

## Workaround

Until this ships, an app can clear the recorded move count itself once the composite goes away, so there is nothing left for a fresh instance to carry out.

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7360 is fixed.
const away = !expanded || handedOff;
useEffect(() => {
  if (!away) return;
  store.setState("moves", 0);
}, [away, store]);
```

For a popup that closes itself, such as the shared menu in `menubar-navigation`, put the same effect in a component inside the provider that owns the store, keyed on `mounted`:

```tsx
// TODO: Remove once https://github.com/ariakit/ariakit/issues/7360 is fixed.
function ForgetRecordedMoves() {
  const menu = Ariakit.useMenuContext();
  const mounted = Ariakit.useStoreState(menu, "mounted");
  useEffect(() => {
    if (mounted) return;
    menu?.setState("moves", 0);
  }, [mounted, menu]);
  return null;
}
```

This is narrower than the fix, and the limitations matter:

- **It clears once, on the way out, and is not a standing invariant.** A `move()` issued later, while the composite is still away, is genuinely pending and must still be carried out when it returns. That is why `away` is a single dependency: splitting it into the two flags that produce it re-runs the effect on a transition between two away states and silently drops that pending request.
- **A `move()` issued in the same commit that takes the composite away is lost**, because the effect runs after that commit. The fix preserves such a request; this does not.
- **The `away` condition must cover every way the composite goes away** — unmounting, and the [`composite`](https://ariakit.com/reference/composite#composite-1) prop becoming `false`. A missed path leaves the replay in place.
- **Do not clear on the way back in.** The container-focus effect is a layout effect, so it runs before a passive effect in the same commit.
- **`moves` is shared state, so this is not safe on every store.** It is safe on plain composite, menu and menubar stores. On a [`Tab`](https://ariakit.com/reference/tab) store it commits the active tab as selected, because that store reacts to any change of `moves` and `selectOnMove` defaults to `true`. On a [`Combobox`](https://ariakit.com/reference/combobox) store it disarms selection-follows-open and rewrites the active value. [`Select`](https://ariakit.com/reference/select) and [`Radio`](https://ariakit.com/reference/radio) ignore a reset and are unaffected.
- **`moves` stops matching its documented meaning**, the number of times [`move`](https://ariakit.com/reference/use-composite-store#move) has been called, for any consumer code that reads it.

## Review gate reassessment

<details>
<summary>Cycle 3: Continue the reproduction-fixture direction, with a comment verify-or-delete sweep</summary>

**Assessment**

Issue severity: a confirmed bug in stable `@ariakit/react`. A remounted `Composite` steals DOM focus and scrolls the page with no move requested, which is a focus-management and orientation defect rather than a cosmetic one.

Expected frequency: needs a prior move, a remount, and `activeId === null` together. That combination occurs in the shipped `menubar-navigation` example, so it is reachable through normal use rather than only through a contrived case.

Supported scope: every widget built on `Composite` that is conditionally rendered while its store outlives it, including `Toolbar`, `Menu`, `Select`, `Combobox`, and `Tab`.

Likely user impact: focus lands on a container the user never targeted, and the page jumps. Measured at 0px to 1580px in the reproduction fixture.

Implementation and maintenance complexity: this snapshot adds no library code, no public API, no new abstraction, and no shared helper. It adds one issue-number sandbox, its Playwright test, its happy-dom duplicate, and one test in an existing fixture, reusing `withFramework`, `flushFrames`, `recordScrollEvents`, and `query`. Three small local helpers, two `useState` booleans, no new state machinery.

Cause of the repeated cycles: of eleven findings across three cycles, only one (cycle 1, scroll anchoring making the page assertions unfalsifiable) required a design change, and it has not recurred. Eight were comment or naming accuracy, six of those in one specific class: a comment asserting library behavior the library does not have. Counts and severity both fall across cycles, 6 to 4 to 1. This is an authoring-accuracy pattern in prose about library internals, not a design converging on a problem.

**Decision**

Continue the current direction. There is no machinery to simplify or remove: the snapshot is the failing reproduction and contains no implementation. Every candidate narrowing would drop coverage the issue asks for by name, including the `composite={false}` to `composite={true}` case it warns must be preserved. The six browser tests are jointly satisfiable only by a fix that distinguishes a consumed move request from a pending one, so the preservation tests are load-bearing rather than padding.

The cycle-3 finding is addressed by widening the remedy to match the defect class: every falsifiable claim about library behavior in the four changed files was verified against source in one pass, or corrected. Thirteen of fourteen held. The one that did not described `composite={false}` as producing "a plain group" with "tab navigation"; the prop actually stops the component managing focus and keyboard navigation so another composite can take over, and the items keep roving tabindex. That wording, the control's label, and the matching test titles were corrected. The menubar comment was also tightened: the active item is null when the menu returns through two separate paths, so the comment now states the observable fact instead of attributing it to one.

Intentionally unsupported in this snapshot: consolidating the issue-number fixture into a semantic one. This is a declined non-goal rather than a registered follow-up. The fixture-lifecycle rules place a new isolated reproduction at step 2, and step 3 consolidation triggers only once issue-number fixtures develop stable overlap, which has not happened. No follow-up record is needed.

No registered follow-up issues arose from this track.

</details>

<details>
<summary>Cycle 6: Continue, and complete the reproduction with the sibling effect</summary>

**Assessment**

Issue severity, expected frequency, supported scope and likely user impact are unchanged from cycle 3, and cycle 6 raised the assessed impact: the defect has a second live path that the reproduction did not cover.

Implementation and maintenance complexity: still no library code, no public API, no new abstraction, and no shared helper. Cycle 6 adds one browser test and one happy-dom duplicate that reuse the fixture's existing controls, with no fixture change.

Cause of the repeated cycles: finding counts run 6, 4, 1, 2, 1, 1. Cycles 1 to 5 were one blocking harness defect plus eleven comment-accuracy and test-strength items, and they converged. Cycle 6 is a different kind of finding, the first substantive one since cycle 1: a coverage gap, found by running the reproduction against unmodified base code. So the cycles are not one problem recurring.

One process defect is worth recording: the cycle-3 remedy introduced the cycle-4 finding, because it sourced its replacement wording from the prop's public JSDoc instead of verifying against the implementation. Every remedy for a "comment asserts behavior the library does not have" finding must itself be verified against the implementation in the same cycle.

**Decision**

Continue the current direction, and include the cycle-6 coverage rather than deferring it.

`CompositeFocusOnMove` runs two effects on the same stale `moves` counter. The reproduction pinned only the container-focus one, guarded by `activeId === null`. Its sibling, "Present the active item", guarded by `activeId != null`, replays a completed move by the same mechanism. Measured on the base commit through the fixture's own controls: after moving to the second item, handing composite behavior over, and returning the page to its origin, taking that behavior back moves focus back to the item and the page from 0px to 1580px, with no new move requested.

This is the same defect through the other branch of the same effect pair, not separable work: the issue's expected behavior says adding the composite back must not create a new focus or scroll request, with no branch qualifier, and the workflow requires reproduction coverage to land before the library fix. Splitting it would either leave the fix's second half unasserted or close the issue with one branch still broken.

The scope is fenced to one browser test and one happy-dom duplicate on the existing controls. Needing new fixture state or a new sandbox would have been the signal to split instead.

This finding also falsified the planned library fix before any of it was written. The plan was to record which composite element had carried out the store's current move request. When only the `composite` prop flips, the effect remounts while the composite element is unchanged, so an element-keyed claim still matches and the replay is allowed. The discriminator has to be the component instance, which a per-instance token records. One guard then covers both effects, so this coverage adds no library machinery.

Registered follow-up: https://github.com/ariakit/ariakit/issues/7363, for the `composite` prop JSDoc, whose "stop managing focus and keyboard navigation for its items" wording is imprecise about item-level navigation. It is pre-existing on `main` and independent of this change, and its ambiguity produced the cycle-4 finding here.

Intentionally unsupported: consolidating this issue-number fixture into a semantic one, as recorded at cycle 3.

</details>

<details>
<summary>Cycle 9: Continue, with a narrowed comment policy</summary>

**Assessment**

Issue severity, supported scope, likely user impact and expected frequency are unchanged. The menubar reproduction is the frequency argument that matters: arrow into a menu, arrow out to a sibling, arrow back is plain keyboard use of a shipped composition.

Implementation and maintenance complexity: still no library code, no public contract, no new abstraction, no platform policy, no runtime cost. The helpers are ten extractions of two to four lines, each one a prior cycle asked for.

The nine-test suite is at minimum size rather than padded. Each test maps either to an explicit acceptance criterion in the issue or to a demonstrated wrong fix, and the cycle-9 reviewer stress-tested seven candidate wrong fixes without finding one that passes all nine. Removing a test would admit a wrong fix, so "simplify or remove machinery" and "narrow the supported contract" are both unavailable on the substance.

Cause of the repeated cycles: of twenty-five findings, six were test validity or coverage, fifteen comment or naming accuracy, and four readability or reuse. The comment findings cluster immediately after the cycles that added tests, because prose written for a smaller suite went stale when new callers arrived, and two of them were introduced by a previous cycle's own remedy. Cycle 9 is the first cycle where the suite did not grow, and it found no correctness, coverage or accuracy defect. The series is converging.

For the skill's low-complexity clause: this is a test-and-fixture-only snapshot whose maintenance cost is the standard, bounded cost of any sandbox in this repository. The repeated cycles do not create a meaningful complexity concern.

**Decision**

Continue, and adopt a narrowed comment policy as the direction refinement rather than a batch deletion of existing prose.

Comments here make one of three kinds of claim: why a setup exists, what the test would otherwise fail to catch, or a derived description of library internals that no assertion depends on. Every one of the fifteen accuracy findings lived in the third kind. Every comment of the first two kinds has produced no finding in nine cycles. The rule going forward: a sentence describing library behavior must be needed to understand why an assertion in that file is what it is; otherwise delete it, and when it is needed, verify it at the call site and attach a permalink where a durable record exists.

A batch cull of existing prose was considered and rejected. It would diverge from the register of the nearest sibling fixtures and of the component source itself, it would discard notes that earned themselves through blocking findings, and rewriting fifty lines of unreviewed prose at this stage is the exact mechanism that produced two of the findings being counted.

Two cycle-9 remedies were revised on this basis. The helper whose name and JSDoc promised an ordering its body lacks is fixed by shortening the JSDoc to its load-bearing clause and renaming it, not by reordering a passing test to satisfy prose. The reuse finding is scoped to the happy-dom file. The browser file's two apparent duplications are deliberate: those tests need no scroll to the top, so routing them through the shared helper would trade duplication for a which-helper-do-I-call problem.

Registered follow-up from earlier cycles, still open: https://github.com/ariakit/ariakit/issues/7363.

Intentionally unsupported: consolidating this issue-number fixture into a semantic one, as recorded at cycle 3.

</details>

<details>
<summary>Cycle 12: Continue, with the record tracking consumption rather than reading</summary>

**Assessment**

Issue severity, supported scope, likely user impact and expected frequency are unchanged from cycle 9. What changed is that the library fix now exists and has been reviewed on its own terms, over three rounds prompted by the pull request comments.

Implementation and maintenance complexity: the fix adds a per-store `WeakMap` record, two small helpers, one `sync` subscription, one `@private` `presentItem` callback, an idempotent `consume`, a `settle` wrapper over `cancel`, and a `settle` property on the value `presentItem` returns. That is more machinery than the first draft, which was a single read-time guard.

Cause of the repeated cycles: all three rounds found the same class of defect, a move request being misclassified at the moment a presentation stops. Round 1 found that claiming on read discards a request still waiting for its item, reproduced through the fixture. Round 1 also found that claiming on carry-out alone replays a request the presentation gave up on, reproduced the same way. Round 2 found that the give-up contract missed the one stop that comes from outside the presentation, a replacement through `usePresentItem`. Each round narrowed the same question rather than opening a new one, and the answer converged on a single invariant.

**Decision**

Continue, and keep the consumption contract.

Four simpler designs were evaluated and rejected on evidence rather than taste.

Claiming at read is what the first draft did. It drops any request whose item registers later, which is supported behavior with its own coverage in `app/src/sandbox/composite-late-item-focus/`.

Claiming only when the item is already connected, the smallest prototype proposed on the pull request, fixes that and nothing else. A request whose item never arrives is never claimed, so picking another item abandons it and leaves it to be replayed. This is what the abandoned-request test pins, and that test was written from a reproduction rather than from the code.

Claiming at read and releasing on cleanup when the presentation is still waiting appears to need no `presentItem` change, but it does not save anything and does not go far enough. Knowing whether a presentation is still waiting means tracking exactly the same thing at exactly the same points inside `presentItem`, since a cancelled presentation and a pending one are otherwise indistinguishable, and it adds a release path on top. It also leaves the superseded case untouched: a presentation replaced through `usePresentItem` would still be released on cleanup and replayed.

Recording a consumed `moves` count per store, which would remove the request record, the `sync` subscription and the instance token together, is unsafe. `CompositeContainer` sets `moves` back to `0`, so a reset followed by fresh moves can reach a count that was already recorded as consumed and silently retire a live request. Resetting on every change of the counter is what avoids that, and it needs the record.

The remaining machinery is the smallest that distinguishes states the store cannot distinguish on its own, because `moves` counts requests and records nothing about what became of them. The tests that pin those states are jointly satisfiable only by a design that keeps them apart: each one fails against a version of the fix that collapses its state into another, and each of those failures was measured rather than argued.

The third state has a third way in, and it needed its own fixture rather than a follow-up. A request can also be taken over, when a new presentation replaces one that is still waiting. `composite-7360` cannot reach it: the palette branch that takes a request over runs only while an item is active, and a roving-tabindex toolbar's container is focusable only while none is, so the two never meet there. That is a property of the fixture rather than of the defect, and `app/src/sandbox/composite-superseded-move/` supplies the missing shape with a listbox-style palette that carries `tabIndex`. Its reproduction test fails in all three engines when the replacement cancels instead of settling, while both of its controls pass, so the failure isolates the takeover.

Registered follow-up: https://github.com/ariakit/ariakit/issues/7378. Review found a second route to the same symptom, and this one is not a replay of a request that was carried out. A move for an item that never registers stays pending, and a later `setActiveId(null)` leaves it indistinguishable from a request for the container, so the container takes focus and the page scrolls on the next mount. It reproduces from this pull request's own fixture, and it reproduces the same way with the merge-base library files swapped in, so it is pre-existing rather than introduced here. Resolving it means recording what a move asked for, not only what became of it, which is a different question from the one this change answers.

Intentionally unsupported, as recorded at cycle 3: consolidating this issue-number fixture into a semantic one. The new fixture is semantic rather than issue-numbered, because it covers a state of the contract rather than a distinct report.

The follow-up registered at cycle 6, https://github.com/ariakit/ariakit/issues/7363, was closed on `main` by #7366.

</details>

Fixes #7360

### DIFF
--- a/.changeset/300-composite-remount-focus.md
+++ b/.changeset/300-composite-remount-focus.md
@@ -3,4 +3,4 @@
 "@ariakit/react": patch
 ---
 
-Fixed [`Composite`](https://ariakit.com/reference/composite) taking focus itself or moving focus to the item that was active before, and scrolling the page to it, when it becomes a composite widget anew after being rendered again or after the [`composite`](https://ariakit.com/reference/composite#composite-1) prop switches back on, with no new [`move`](https://ariakit.com/reference/use-composite-store#move) call behind it. This also applies to all components built on it, such as [`Menu`](https://ariakit.com/reference/menu) and [`Toolbar`](https://ariakit.com/reference/toolbar).
+Fixed [`Composite`](https://ariakit.com/reference/composite) replaying an earlier [`move`](https://ariakit.com/reference/use-composite-store#move) call, taking focus and scrolling the page, when it is mounted again or when the [`composite`](https://ariakit.com/reference/composite#composite-1) prop switches back on. This also applies to all components built on it, such as [`Menu`](https://ariakit.com/reference/menu) and [`Toolbar`](https://ariakit.com/reference/toolbar).
--- a/.changeset/300-composite-remount-focus.md
+++ b/.changeset/300-composite-remount-focus.md
@@ -1,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Composite`](https://ariakit.com/reference/composite) taking focus itself or moving focus to the item that was active before, and scrolling the page to it, when it becomes a composite widget anew after being rendered again or after the [`composite`](https://ariakit.com/reference/composite#composite-1) prop switches back on, with no new [`move`](https://ariakit.com/reference/use-composite-store#move) call behind it. This also applies to all components built on it, such as [`Menu`](https://ariakit.com/reference/menu) and [`Toolbar`](https://ariakit.com/reference/toolbar).
--- a/app/src/sandbox/composite-7360/test-browser.ts
+++ b/app/src/sandbox/composite-7360/test-browser.ts
@@ -1,0 +1,229 @@
+import type { query } from "@ariakit/test/playwright";
+import type { Page } from "@playwright/test";
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+import { recordScrollEvents } from "#app/test-utils/scroll.ts";
+
+type Query = ReturnType<typeof query>;
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  /**
+   * Moves the active item with the keyboard. The move is what arms the replay:
+   * it leaves a move count behind that a fresh effect instance reads as a focus
+   * request to carry out when the toolbar comes back, whether it comes back by
+   * mounting again or by becoming a composite again.
+   */
+  const moveToSecondItem = async (q: Query, page: Page) => {
+    await q.button("Bold").click();
+    await page.keyboard.press("ArrowRight");
+    await test.expect(q.button("Italic")).toBeFocused();
+  };
+
+  /** Rests the page at its origin, away from the toolbar at the end. */
+  const scrollToTop = async (page: Page) => {
+    await page.evaluate(() => window.scrollTo({ top: 0 }));
+    await test.expect.poll(() => page.evaluate(() => window.scrollY)).toBe(0);
+  };
+
+  const handOffControl = (q: Query) =>
+    q.checkbox("Hand off composite behavior");
+
+  /** Hands composite behavior over and rests the page at its origin. */
+  const handOff = async (q: Query, page: Page) => {
+    await handOffControl(q).click();
+    await test.expect(handOffControl(q)).toBeChecked();
+    await scrollToTop(page);
+  };
+
+  /**
+   * Takes composite behavior back and settles the point where a focus or scroll
+   * request would land.
+   */
+  const takeBackAndSettle = async (q: Query, page: Page) => {
+    await handOffControl(q).click();
+    await test.expect(handOffControl(q)).not.toBeChecked();
+    // Same checkpoint as `expandAndSettle`, reached through the `composite`
+    // prop instead of a mount.
+    await flushFrames(page);
+  };
+
+  /**
+   * Returns to the controls at the top of the document and collapses the
+   * toolbar from there, so the page rests at its origin with the toggle holding
+   * focus and the toolbar unmounted.
+   */
+  const collapseFromTop = async (q: Query, page: Page) => {
+    await scrollToTop(page);
+    await q.button("Collapse toolbar").click();
+    await test.expect(q.toolbar("Formatting")).toBeHidden();
+  };
+
+  /**
+   * Brings the toolbar back and settles the point where a focus or scroll
+   * request would land.
+   */
+  const expandAndSettle = async (q: Query, page: Page) => {
+    await q.button("Expand toolbar").click();
+    await test.expect(q.toolbar("Formatting")).toBeVisible();
+    // Focus staying put and the page not moving have no positive state to wait
+    // on, and the effect that would disturb either runs a commit after the
+    // toolbar appears, once its element reaches the store.
+    await flushFrames(page);
+  };
+
+  // Pins the harness: with no move behind it, nothing in the library asks for
+  // focus or a scroll here, so this fails only if mounting the toolbar moves
+  // focus or the page on its own.
+  test("keeps focus and page position when an untouched toolbar comes back", async ({
+    page,
+    q,
+  }) => {
+    await collapseFromTop(q, page);
+    const scroll = await recordScrollEvents(page);
+
+    await expandAndSettle(q, page);
+
+    await test.expect(q.button("Collapse toolbar")).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7360
+  test("keeps focus and page position when the toolbar comes back after a move", async ({
+    page,
+    q,
+  }) => {
+    await moveToSecondItem(q, page);
+    await collapseFromTop(q, page);
+    const scroll = await recordScrollEvents(page);
+
+    await expandAndSettle(q, page);
+
+    await test.expect(q.button("Collapse toolbar")).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // https://github.com/ariakit/ariakit/issues/7360
+  test("keeps focus and page position through repeated remounts", async ({
+    page,
+    q,
+  }) => {
+    await moveToSecondItem(q, page);
+    await collapseFromTop(q, page);
+    const scroll = await recordScrollEvents(page);
+
+    for (let cycle = 0; cycle < 3; cycle += 1) {
+      await expandAndSettle(q, page);
+      await test.expect(q.button("Collapse toolbar")).toBeFocused();
+      await q.button("Collapse toolbar").click();
+      await test.expect(q.toolbar("Formatting")).toBeHidden();
+    }
+
+    await test.expect(q.button("Expand toolbar")).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // The same pin as the control above, for the other way the toolbar comes
+  // back: whatever the `composite` prop switching off and on does on its own,
+  // it isn't what the tests below observe.
+  test("keeps focus and page position when untouched composite behavior comes back", async ({
+    page,
+    q,
+  }) => {
+    await handOff(q, page);
+    const scroll = await recordScrollEvents(page);
+
+    await takeBackAndSettle(q, page);
+
+    await test.expect(handOffControl(q)).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // The sibling request: with an item still active, what gets replayed presents
+  // that item instead of the toolbar container. Handing composite behavior over
+  // and taking it back remounts the effect without replacing the toolbar
+  // element, so this also covers a remount the element itself survives.
+  // https://github.com/ariakit/ariakit/issues/7360
+  test("keeps focus and page position when composite behavior comes back", async ({
+    page,
+    q,
+  }) => {
+    await moveToSecondItem(q, page);
+    await handOff(q, page);
+    const scroll = await recordScrollEvents(page);
+
+    await takeBackAndSettle(q, page);
+
+    await test.expect(handOffControl(q)).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // The other half of that branch: a move recorded while composite behavior is
+  // handed off has nothing to carry it out, so it stays pending and its item
+  // takes focus once the toolbar is a composite again. The move comes from the
+  // arrow key, which still moves the active item meanwhile; only focus stops
+  // following it.
+  // https://github.com/ariakit/ariakit/issues/7363
+  test("focuses a newly moved item when composite behavior comes back", async ({
+    page,
+    q,
+  }) => {
+    await moveToSecondItem(q, page);
+    await handOffControl(q).click();
+    await test.expect(handOffControl(q)).toBeChecked();
+
+    await q.button("Italic").click();
+    await page.keyboard.press("ArrowRight");
+    await takeBackAndSettle(q, page);
+
+    await test.expect(q.button("Underline")).toBeFocused();
+  });
+
+  test("focuses the toolbar when a focus command runs while it is expanded", async ({
+    q,
+  }) => {
+    // Pins the fixture geometry the two `toBeInViewport` assertions rely on: a
+    // shorter document would leave the toolbar on screen from the start and
+    // make them pass without the library scrolling anything.
+    await test.expect(q.toolbar("Formatting")).not.toBeInViewport();
+
+    await q.button("Focus toolbar").click();
+
+    await test.expect(q.toolbar("Formatting")).toBeFocused();
+    await test.expect(q.toolbar("Formatting")).toBeInViewport();
+  });
+
+  // A move that arrives with no composite element to carry it out stays
+  // pending, so the toolbar takes focus once it is back.
+  test("focuses the toolbar when a focus command runs while it is collapsed", async ({
+    page,
+    q,
+  }) => {
+    await moveToSecondItem(q, page);
+    await collapseFromTop(q, page);
+
+    await q.button("Focus toolbar").click();
+    await q.button("Expand toolbar").click();
+
+    await test.expect(q.toolbar("Formatting")).toBeFocused();
+    await test.expect(q.toolbar("Formatting")).toBeInViewport();
+  });
+
+  // The same pending move, but with the toolbar mounted the whole time and only
+  // its `composite` prop switching off and back on.
+  test("focuses the toolbar when a focus command runs while it is handed off", async ({
+    q,
+  }) => {
+    await handOffControl(q).click();
+    await test.expect(handOffControl(q)).toBeChecked();
+
+    await q.button("Focus toolbar").click();
+    await handOffControl(q).click();
+
+    await test.expect(handOffControl(q)).not.toBeChecked();
+    await test.expect(q.toolbar("Formatting")).toBeFocused();
+  });
+});
--- a/app/src/sandbox/composite-7360/test-browser.ts
+++ b/app/src/sandbox/composite-7360/test-browser.ts
@@ -1,18 +1,16 @@
-import type { query } from "@ariakit/test/playwright";
 import type { Page } from "@playwright/test";
 import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
 import { recordScrollEvents } from "#app/test-utils/scroll.ts";
 
-type Query = ReturnType<typeof query>;
-
-withFramework(import.meta.dirname, async ({ test }) => {
+withFramework(import.meta.dirname, async ({ test, query }) => {
   /**
    * Moves the active item with the keyboard. The move is what arms the replay:
    * it leaves a move count behind that a fresh effect instance reads as a focus
    * request to carry out when the toolbar comes back, whether it comes back by
    * mounting again or by becoming a composite again.
    */
-  const moveToSecondItem = async (q: Query, page: Page) => {
+  const moveToSecondItem = async (page: Page) => {
+    const q = query(page);
     await q.button("Bold").click();
     await page.keyboard.press("ArrowRight");
     await test.expect(q.button("Italic")).toBeFocused();
@@ -24,13 +22,13 @@ withFramework(import.meta.dirname, async ({ test }) => {
     await test.expect.poll(() => page.evaluate(() => window.scrollY)).toBe(0);
   };
 
-  const handOffControl = (q: Query) =>
-    q.checkbox("Hand off composite behavior");
+  const handOffControl = (page: Page) =>
+    query(page).checkbox("Hand off composite behavior");
 
   /** Hands composite behavior over and rests the page at its origin. */
-  const handOff = async (q: Query, page: Page) => {
-    await handOffControl(q).click();
-    await test.expect(handOffControl(q)).toBeChecked();
+  const handOff = async (page: Page) => {
+    await handOffControl(page).click();
+    await test.expect(handOffControl(page)).toBeChecked();
     await scrollToTop(page);
   };
 
@@ -38,9 +36,9 @@ withFramework(import.meta.dirname, async ({ test }) => {
    * Takes composite behavior back and settles the point where a focus or scroll
    * request would land.
    */
-  const takeBackAndSettle = async (q: Query, page: Page) => {
-    await handOffControl(q).click();
-    await test.expect(handOffControl(q)).not.toBeChecked();
+  const takeBackAndSettle = async (page: Page) => {
+    await handOffControl(page).click();
+    await test.expect(handOffControl(page)).not.toBeChecked();
     // Same checkpoint as `expandAndSettle`, reached through the `composite`
     // prop instead of a mount.
     await flushFrames(page);
@@ -51,7 +49,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
    * toolbar from there, so the page rests at its origin with the toggle holding
    * focus and the toolbar unmounted.
    */
-  const collapseFromTop = async (q: Query, page: Page) => {
+  const collapseFromTop = async (page: Page) => {
+    const q = query(page);
     await scrollToTop(page);
     await q.button("Collapse toolbar").click();
     await test.expect(q.toolbar("Formatting")).toBeHidden();
@@ -61,13 +60,31 @@ withFramework(import.meta.dirname, async ({ test }) => {
    * Brings the toolbar back and settles the point where a focus or scroll
    * request would land.
    */
-  const expandAndSettle = async (q: Query, page: Page) => {
+  const expandAndSettle = async (page: Page) => {
+    const q = query(page);
     await q.button("Expand toolbar").click();
     await test.expect(q.toolbar("Formatting")).toBeVisible();
     // Focus staying put and the page not moving have no positive state to wait
     // on, and the effect that would disturb either runs a commit after the
     // toolbar appears, once its element reaches the store.
     await flushFrames(page);
+  };
+
+  /**
+   * Asks for a tool that has not loaded yet. The move counts as a request like
+   * any other, but nothing can carry it out until the item registers.
+   */
+  const askForHighlight = async (page: Page) => {
+    const q = query(page);
+    await scrollToTop(page);
+    await q.button("Focus highlight tool").click();
+    await test.expect(q.button("Highlight")).toBeHidden();
+  };
+
+  const loadHighlight = async (page: Page) => {
+    const q = query(page);
+    await q.button("Load highlight tool").click();
+    await test.expect(q.button("Highlight")).toBeVisible();
   };
 
   // Pins the harness: with no move behind it, nothing in the library asks for
@@ -77,10 +94,10 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    await collapseFromTop(q, page);
+    await collapseFromTop(page);
     const scroll = await recordScrollEvents(page);
 
-    await expandAndSettle(q, page);
+    await expandAndSettle(page);
 
     await test.expect(q.button("Collapse toolbar")).toBeFocused();
     test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
@@ -92,11 +109,11 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    await moveToSecondItem(q, page);
-    await collapseFromTop(q, page);
+    await moveToSecondItem(page);
+    await collapseFromTop(page);
     const scroll = await recordScrollEvents(page);
 
-    await expandAndSettle(q, page);
+    await expandAndSettle(page);
 
     await test.expect(q.button("Collapse toolbar")).toBeFocused();
     test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
@@ -108,12 +125,12 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    await moveToSecondItem(q, page);
-    await collapseFromTop(q, page);
+    await moveToSecondItem(page);
+    await collapseFromTop(page);
     const scroll = await recordScrollEvents(page);
 
     for (let cycle = 0; cycle < 3; cycle += 1) {
-      await expandAndSettle(q, page);
+      await expandAndSettle(page);
       await test.expect(q.button("Collapse toolbar")).toBeFocused();
       await q.button("Collapse toolbar").click();
       await test.expect(q.toolbar("Formatting")).toBeHidden();
@@ -129,14 +146,13 @@ withFramework(import.meta.dirname, async ({ test }) => {
   // it isn't what the tests below observe.
   test("keeps focus and page position when untouched composite behavior comes back", async ({
     page,
-    q,
   }) => {
-    await handOff(q, page);
+    await handOff(page);
     const scroll = await recordScrollEvents(page);
 
-    await takeBackAndSettle(q, page);
+    await takeBackAndSettle(page);
 
-    await test.expect(handOffControl(q)).toBeFocused();
+    await test.expect(handOffControl(page)).toBeFocused();
     test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
     test.expect(await scroll.events()).not.toContain("document");
   });
@@ -148,15 +164,14 @@ withFramework(import.meta.dirname, async ({ test }) => {
   // https://github.com/ariakit/ariakit/issues/7360
   test("keeps focus and page position when composite behavior comes back", async ({
     page,
-    q,
   }) => {
-    await moveToSecondItem(q, page);
-    await handOff(q, page);
+    await moveToSecondItem(page);
+    await handOff(page);
     const scroll = await recordScrollEvents(page);
 
-    await takeBackAndSettle(q, page);
+    await takeBackAndSettle(page);
 
-    await test.expect(handOffControl(q)).toBeFocused();
+    await test.expect(handOffControl(page)).toBeFocused();
     test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
     test.expect(await scroll.events()).not.toContain("document");
   });
@@ -166,18 +181,18 @@ withFramework(import.meta.dirname, async ({ test }) => {
   // takes focus once the toolbar is a composite again. The move comes from the
   // arrow key, which still moves the active item meanwhile; only focus stops
   // following it.
-  // https://github.com/ariakit/ariakit/issues/7363
+  // Related: https://github.com/ariakit/ariakit/issues/7363
   test("focuses a newly moved item when composite behavior comes back", async ({
     page,
     q,
   }) => {
-    await moveToSecondItem(q, page);
-    await handOffControl(q).click();
-    await test.expect(handOffControl(q)).toBeChecked();
+    await moveToSecondItem(page);
+    await handOffControl(page).click();
+    await test.expect(handOffControl(page)).toBeChecked();
 
     await q.button("Italic").click();
     await page.keyboard.press("ArrowRight");
-    await takeBackAndSettle(q, page);
+    await takeBackAndSettle(page);
 
     await test.expect(q.button("Underline")).toBeFocused();
   });
@@ -202,8 +217,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    await moveToSecondItem(q, page);
-    await collapseFromTop(q, page);
+    await moveToSecondItem(page);
+    await collapseFromTop(page);
 
     await q.button("Focus toolbar").click();
     await q.button("Expand toolbar").click();
@@ -215,15 +230,71 @@ withFramework(import.meta.dirname, async ({ test }) => {
   // The same pending move, but with the toolbar mounted the whole time and only
   // its `composite` prop switching off and back on.
   test("focuses the toolbar when a focus command runs while it is handed off", async ({
+    page,
     q,
   }) => {
-    await handOffControl(q).click();
-    await test.expect(handOffControl(q)).toBeChecked();
+    await handOffControl(page).click();
+    await test.expect(handOffControl(page)).toBeChecked();
 
     await q.button("Focus toolbar").click();
-    await handOffControl(q).click();
+    await handOffControl(page).click();
 
-    await test.expect(handOffControl(q)).not.toBeChecked();
+    await test.expect(handOffControl(page)).not.toBeChecked();
     await test.expect(q.toolbar("Formatting")).toBeFocused();
+  });
+
+  // Pins the third kind of pending move, the one waiting on its item rather
+  // than on the toolbar: with nothing replacing the toolbar meanwhile, the
+  // request is carried out when the tool loads.
+  test("focuses a tool that loads after being asked for", async ({
+    page,
+    q,
+  }) => {
+    await askForHighlight(page);
+
+    await loadHighlight(page);
+
+    await test.expect(q.button("Highlight")).toBeFocused();
+    await test.expect(q.button("Highlight")).toBeInViewport();
+  });
+
+  // The third thing a pending move can do: be given up on. Picking another tool
+  // makes the toolbar abandon the request for the one that never arrived, and a
+  // request it gave up on is spent rather than still waiting, so it is not left
+  // behind for a later instance to carry out.
+  // https://github.com/ariakit/ariakit/issues/7360
+  test("keeps focus and page position when composite behavior comes back after an abandoned request", async ({
+    page,
+    q,
+  }) => {
+    await askForHighlight(page);
+    await q.button("Bold").click();
+    await test.expect(q.button("Bold")).toBeFocused();
+    await handOff(page);
+    const scroll = await recordScrollEvents(page);
+
+    await takeBackAndSettle(page);
+
+    await test.expect(handOffControl(page)).toBeFocused();
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // The same pending move, across the handoff that replaces what would carry it
+  // out. Only the handoff is exercised here: collapsing also points the store
+  // at the toolbar container, which retires the request instead of leaving it
+  // pending.
+  test("focuses a tool that loads after composite behavior comes back", async ({
+    page,
+    q,
+  }) => {
+    await askForHighlight(page);
+    await handOff(page);
+    await takeBackAndSettle(page);
+
+    await loadHighlight(page);
+
+    await test.expect(q.button("Highlight")).toBeFocused();
+    await test.expect(q.button("Highlight")).toBeInViewport();
   });
 });
--- a/packages/ariakit-react-components/src/composite/composite.tsx
+++ b/packages/ariakit-react-components/src/composite/composite.tsx
@@ -119,46 +119,54 @@ type PresentItem = ReturnType<typeof usePresentItem>;
 
 interface MoveRequest {
   /**
-   * The `CompositeFocusOnMove` instance that has claimed the current move, or
-   * `null` while no instance has.
+   * The `CompositeFocusOnMove` instance that consumed the current move, or
+   * `null` while none has.
    */
-  claimedBy: object | null;
+  consumedBy: object | null;
 }
 
 /**
- * Tracks, per store, which instance has claimed the current move. This has to
+ * Tracks, per store, which instance has consumed the current move. This has to
  * outlive the component: `moves` only counts requests, so a fresh instance
- * reading a nonzero count can't tell a request still waiting for a composite
- * element from one an earlier instance already acted on.
+ * reading a nonzero count can't tell a request still waiting, for its item or
+ * for a composite element, from one an earlier instance already acted on.
  */
 const moveRequests = new WeakMap<CompositeStore, MoveRequest>();
 
 function getMoveRequest(store: CompositeStore) {
   const cached = moveRequests.get(store);
   if (cached) return cached;
-  const request: MoveRequest = { claimedBy: null };
+  const request: MoveRequest = { consumedBy: null };
   moveRequests.set(store, request);
   // Every change to the count starts a new request, including the resets that
-  // cancel a pending one, so it goes back to being unclaimed.
+  // cancel a pending one, so it goes back to being unconsumed.
   sync(store, ["moves"], () => {
-    request.claimedBy = null;
+    request.consumedBy = null;
   });
   return request;
 }
 
 /**
- * Whether `instance` may act on the store's current move, claiming it when it
- * may. The claim is per component instance rather than per composite element,
- * because switching the `composite` prop off and on again mounts a fresh
- * instance while the element itself survives. Both effects claim before the
- * guard that decides whether they act, so a request one of them read is never
- * left unclaimed for a later instance to act on.
+ * Whether `instance` may act on the store's current move. An unconsumed request
+ * is available to whichever instance is around; a consumed one stays with the
+ * instance that consumed it, so its own effects can still run again.
  */
-function claimMove(store: CompositeStore, instance: object) {
-  const request = getMoveRequest(store);
-  if (request.claimedBy && request.claimedBy !== instance) return false;
-  request.claimedBy = instance;
-  return true;
+function mayActOnMove(store: CompositeStore, instance: object) {
+  const { consumedBy } = getMoveRequest(store);
+  return !consumedBy || consumedBy === instance;
+}
+
+/**
+ * Records that `instance` consumed the store's move `moves`, so no later
+ * instance replays it. It is per component instance rather than per composite
+ * element, because switching the `composite` prop off and on again mounts a
+ * fresh instance while the element itself survives. A consumption that arrives
+ * after the count has moved on is dropped, resets included, because it settles
+ * the request the count held when it started.
+ */
+function consumeMove(store: CompositeStore, instance: object, moves: number) {
+  if (store.getState().moves !== moves) return;
+  getMoveRequest(store).consumedBy = instance;
 }
 
 interface CompositeFocusOnMoveProps {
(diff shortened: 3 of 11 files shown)


.changeset/300-composite-remount-focus.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Composite`](https://ariakit.com/reference/composite) taking focus itself or moving focus to the item that was active before, and scrolling the page to it, when it becomes a composite widget anew after being rendered again or after the [`composite`](https://ariakit.com/reference/composite#composite-1) prop switches back on, with no new [`move`](https://ariakit.com/reference/use-composite-store#move) call behind it. This also applies to all components built on it, such as [`Menu`](https://ariakit.com/reference/menu) and [`Toolbar`](https://ariakit.com/reference/toolbar).
```
diegohaz: This changeset has too much details. Let's simplify it.

app/src/sandbox/composite-7360/test-browser.ts
```diff
@@ -0,0 +1,229 @@
...
+type Query = ReturnType<typeof query>;
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  /**
+   * Moves the active item with the keyboard. The move is what arms the replay:
+   * it leaves a move count behind that a fresh effect instance reads as a focus
+   * request to carry out when the toolbar comes back, whether it comes back by
+   * mounting again or by becoming a composite again.
+   */
+  const moveToSecondItem = async (q: Query, page: Page) => {
```
diegohaz: There's no need to accept `q` here and in other abstracted functions. Since we have page, we can just use the `query` function provided by withFramework callback parameter to build the `q` object.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -149,11 +194,17 @@ const CompositeFocusOnMove = memo(function CompositeFocusOnMove({
...
+  // effect re-runs of the same instance, including StrictMode's double
+  // invocation, but not a remount, which is exactly the distinction the request
+  // needs.
+  const instanceRef = useRef({});
 
   // Present the active item.
   useEffect(() => {
     if (!moves) return;
     if (!focusOnMove) return;
+    if (!claimMove(store, instanceRef.current)) return;
```
diegohaz: Claiming before the guard that decides whether to act marks a request consumed even when this instance cannot carry it out. `present()` stays pending while its target item has not registered yet, so the claim outlives a request that never happened. When the instance is replaced, its cleanup cancels that pending presentation and the replacement is refused the claim, so the move is dropped and focus never reaches the item.

Moving to an item that registers in a later commit is supported behavior with its own coverage in `app/src/sandbox/composite-late-item-focus/`. Both ways of replacing the instance that this PR already exercises reach the problem: the `composite` prop flipping off and back on, and the composite unmounting and mounting again.

**Real-world impact.** An app loads composite items asynchronously (a lazy list, or a `CollectionRenderer` fed by a request) and calls `store.move(id)` for an item that has not rendered yet, while the list remounts or hands composite behavior over as the data arrives. Before this change the item took focus once it registered. After it, focus silently stays where it was, the keyboard navigation the user asked for appears to do nothing, and nothing recovers until a new `move()` happens.

**Reproduction**, run on this branch. Sandbox `index.react.tsx`:

```tsx
  const store = Ariakit.useCompositeStore();
  const [handedOff, setHandedOff] = useState(false);
  const [expanded, setExpanded] = useState(true);
  const [loaded, setLoaded] = useState(false);
  return (
    <main>
      <button type="button" tabIndex={0} onClick={() => store.move("late")}>Move to late item</button>
      <button type="button" tabIndex={0} onClick={() => setLoaded(true)}>Load late item</button>
      <button type="button" tabIndex={0} onClick={() => setExpanded(!expanded)}>
        {expanded ? "Collapse toolbar" : "Expand toolbar"}
      </button>
      <label>
        <input type="checkbox" tabIndex={0} checked={handedOff} onChange={(event) => setHandedOff(event.currentTarget.checked)} />
        Hand off composite behavior
      </label>
      {expanded && (
        <Ariakit.Composite store={store} composite={!handedOff} role="toolbar" aria-label="Formatting">
          <Ariakit.CompositeItem id="bold">Bold</Ariakit.CompositeItem>
          {loaded && <Ariakit.CompositeItem id="late">Late</Ariakit.CompositeItem>}
        </Ariakit.Composite>
      )}
    </main>
  );
}
```

`test.ts`:

```ts
test("control: pending move is carried out when the item registers", async () => {
  await click(q.button("Move to late item"));
  await click(q.button("Load late item"));
  expect(q.button("Late")).toHaveFocus();
});

test("pending move survives a composite handoff", async () => {
  await click(q.button("Move to late item"));
  await click(q.checkbox("Hand off composite behavior"));
  await click(q.checkbox("Hand off composite behavior"));
  await click(q.button("Load late item"));
  expect(q.button("Late")).toHaveFocus();
});

test("pending move survives a remount", async () => {
  await click(q.button("Move to late item"));
  await click(q.button("Collapse toolbar"));
  await click(q.button("Expand toolbar"));
  await click(q.button("Load late item"));
  expect(q.button("Late")).toHaveFocus();
});

test("a new move after the handoff is carried out", async () => {
  await click(q.button("Move to late item"));
  await click(q.checkbox("Hand off composite behavior"));
  await click(q.checkbox("Hand off composite behavior"));
  await click(q.button("Move to late item"));
  await click(q.button("Load late item"));
  expect(q.button("Late")).toHaveFocus();
});
```

All four pass on the merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10). At [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05), `pending move survives a composite handoff` and `pending move survives a remount` fail. The control and `a new move after the handoff is carried out` pass in both, which isolates the stale claim rather than the presentation cancel on unmount.

**Fix direction**, prototyped locally: claim when the request is carried out, not when it is read. Splitting the helper into a read check and a claim, claiming in the container branch only after `isSelfActive`, and in this branch only when the active item's element is already connected, kept the nine tests in this PR green along with the whole Vitest suite (280 files, 1972 tests) and turned the two failures above green.

```ts
if (!mayActOnMove(store, instanceRef.current)) return;
const { activeId } = store.getState();
if (activeId == null) return;
// A request still waiting for its item has not been carried out, so it stays
// available to whichever instance is around when the item arrives.
if (getEnabledItem(store, activeId)?.element?.isConnected) {
  claimMove(store, instanceRef.current);
}
```

That element check is a local proxy for "the presentation will act now". A presentation withheld for a closed popup is the same class of pending request, so a completion signal from `presentItem` would cover both. Declaring the pending case unsupported and pinning the new behavior with a test is also a coherent resolution, but it contradicts what `focuses a newly moved item when composite behavior comes back` asserts for a move recorded while handed off.

No recording is attached: the evidence here is a deterministic test differential between two commits rather than a visual state, and the fixture above reproduces it in a few seconds.

**Assessment**

- Severity: Medium (estimate, 85% confidence), because a focus request is silently dropped with no error and no fallback, and keyboard users lose the navigation they asked for.
- Priority: P1 (estimate, 80% confidence), because this change introduces the behavior and the PR's own contract says a move that was not carried out must still be carried out.
- Breakage likelihood: Possible (estimate, 75% confidence), because it needs a move targeting an item that registers in a later commit plus an instance replacement before it registers. Both are supported and separately covered here, but the combination is not everyday usage.
- Fix effort: Moderate (estimate, 75% confidence), because the local proxy above is a small change that is already validated, while covering the withheld-popup case as well needs a completion signal from `presentItem`.

app/src/sandbox/composite-7360/test-browser.ts
```diff
@@ -0,0 +1,229 @@
...
+    test.expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    test.expect(await scroll.events()).not.toContain("document");
+  });
+
+  // The other half of that branch: a move recorded while composite behavior is
+  // handed off has nothing to carry it out, so it stays pending and its item
+  // takes focus once the toolbar is a composite again. The move comes from the
+  // arrow key, which still moves the active item meanwhile; only focus stops
+  // following it.
+  // https://github.com/ariakit/ariakit/issues/7363
```
diegohaz: This URL sits where every other test in the file puts the issue it reproduces, so it reads as "this test reproduces #7363". It does not. #7363 is a closed documentation issue about the `composite` prop JSDoc, and this test reproduces the #7360 pending-move behavior. The same line is in the happy-dom duplicate at `app/src/sandbox/composite-7360/test.ts:106`.

The link is doing a different job here: it backs the sentence above it about arrow keys still moving the active item while composite behavior is handed off. Keeping the two roles apart matters because the issue-number position is what a reader greps for when tracing coverage back to a report, and `ariakit-ariakit-example-playwright-test` reserves it for the issue a test reproduces.

Either attach the link to the claim it supports, for example:

```suggestion
  // arrow key, which still moves the active item meanwhile (see
  // https://github.com/ariakit/ariakit/issues/7363); only focus stops
  // following it.
```

or drop it and label the test with `// https://github.com/ariakit/ariakit/issues/7360`, which is the issue it actually covers.

**Assessment**

- Severity: Low (estimate, 90% confidence), because nothing in the behavior changes; the cost is a reader tracing coverage to the wrong report.
- Priority: P2 (estimate, 85% confidence), because it is an accuracy fix inside code this PR adds and it is cheap to do now.
- Breakage likelihood: Rare (estimate, 90% confidence), because it misleads only someone auditing which tests cover which issue.
- Fix effort: Easy (verified, 95% confidence), because it is a two-line comment edit in two files.

.changeset/300-composite-remount-focus.md
diegohaz: Confirmed, still unaddressed at [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05). The entry is one 70-word sentence carrying four separate qualifiers: what the component does wrong, that it also scrolls, the two ways it comes back, and that no new `move` is behind it. `ariakit-general-changeset` asks a single-paragraph entry to be a short past-tense summary sentence, and the mechanism detail belongs in the issue rather than the changelog.

Both reference links check out and are worth keeping: `https://ariakit.com/reference/composite` returns 200 and its page does define `id="composite-1"`, and `https://ariakit.com/reference/use-composite-store` returns 200. Naming `Menu` and `Toolbar` as representative derived components also matches the two-component limit for a single-paragraph entry.

A shorter shape that keeps the reader-facing facts:

```md
Fixed [`Composite`](https://ariakit.com/reference/composite) taking focus and scrolling the page when it is rendered again, or when the [`composite`](https://ariakit.com/reference/composite#composite-1) prop switches back on, after an earlier [`move`](https://ariakit.com/reference/use-composite-store#move) call. This also applies to all components built on it, such as [`Menu`](https://ariakit.com/reference/menu) and [`Toolbar`](https://ariakit.com/reference/toolbar).
```

No contributor credit sentence is needed here, since #7360 was filed by you and `ariakit-general-contributor-credit` excludes `diegohaz`.

app/src/sandbox/composite-7360/test-browser.ts
diegohaz: Confirmed, still unaddressed at [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05). `withFramework` does pass `query` to its callback (`app/src/test-utils/preview.ts:104` calls `callback({ id, framework, query, test })`), so the five helpers here can build their own `q` from `page` and drop both the `q` parameter and the `type Query = ReturnType<typeof query>` alias with its type-only import. `handOffControl` is the one that then needs `page` where it takes none today.

One thing worth deciding along with it: this file did not invent the pattern. `app/src/sandbox/composite-presentation-cancel-on-unmount/test-browser.ts` and `app/src/sandbox/popover-disclosure-7087/test-browser.ts` use the same `type Query = ReturnType<typeof query>` plus `(q: Query, page: Page)` shape, and the first of those is the nearest sibling fixture to this one. Changing only this file leaves three sandboxes with two conventions for the same thing. Either is fine as a decision, but it is worth making it explicitly rather than by file.

The neighbouring `app/src/sandbox/menubar-navigation/test-browser.ts` already builds `q` per test with `const q = query(page)`, which is the shape this comment is asking for.

diegohaz (commented):
## Review summary

Initial review of all six changed files, diffed from the merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10) to [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05), in a dedicated worktree on the PR head. Context beyond the diff: `CompositeFocusOnMove` and its render site, `presentItem`/`usePresentItem`, `sync` in `@ariakit/store`, every `useComposite` caller, the neighbouring composite sandboxes, and the shipped `examples/menubar-navigation` against its app sandbox copy.

### What I ran

- `pnpm test --run app/src/sandbox/composite-7360` at head: 9 passed.
- The same with only `composite.tsx` reverted to the merge base: 3 failed (`keeps focus when the toolbar comes back after a move`, `keeps focus through repeated remounts`, `keeps focus when composite behavior comes back`), 6 passed. The regression coverage is proven red-then-green, and the six preservation controls hold on base as claimed.
- `pnpm test --run` at head: 279 files, 1968 tests, all passing.
- A scratch sandbox for the pending-move case, run at head and on the merge base, plus a prototype of the fix direction with the full suite again (280 files, 1972 tests, all passing).
- Reference link checks for the changeset: `https://ariakit.com/reference/composite` (200, and its page defines `id="composite-1"`) and `https://ariakit.com/reference/use-composite-store` (200).
- CI on this head is green across Chrome, Firefox, Safari, iOS and Android, so I did not re-run the Playwright suites locally.

### Findings

One correctness finding and one comment-accuracy finding, both inline.

The correctness one is the important one: claiming a move before the guard that decides whether to act marks a request consumed even when the instance cannot carry it out. A `move(id)` whose item registers in a later commit leaves `present()` pending; replacing the instance cancels that presentation and refuses the claim to its replacement, so the move is dropped. Verified by differential: the reproduction passes 4 of 4 on the merge base and fails 2 of 4 at this head, while a fresh `move()` after the same handoff still works, which rules out the presentation cancel on unmount as the cause. A prototype that claims at carry-out time rather than at read time keeps all nine tests here green along with the whole suite, and turns the two failures green.

### Unresolved threads validated against this head

- [Thread on `.changeset/300-composite-remount-focus.md`](https://github.com/ariakit/ariakit/pull/7368#discussion_r3898810072), "This changeset has too much details": **confirmed**, still unaddressed. Disposition: Address now. Replied with the specific structural problem and a shorter shape. Severity Low (estimate, 90%), Priority P2 (estimate, 85%), Breakage likelihood Rare (estimate, 95%, it is changelog prose), Fix effort Easy (verified, 95%).
- [Thread on `app/src/sandbox/composite-7360/test-browser.ts`](https://github.com/ariakit/ariakit/pull/7368#discussion_r3898835118), on helpers taking `q`: **confirmed**, still unaddressed. `withFramework` does pass `query` to its callback, so the helpers can build `q` from `page`. Disposition: Address now. Replied, and flagged that two sibling sandboxes already use the current shape, so the decision is worth making across them rather than in this file alone. Severity Low (estimate, 90%), Priority P2 (estimate, 85%), Breakage likelihood Rare (estimate, 95%), Fix effort Easy (verified, 90%).

### Review-body pass

CodeRabbit reported no actionable comments but flagged a merge risk: "an interrupted or non-consuming instance may retain ownership and suppress a legitimate focus request." It has no thread anchor and no reproduction, so I validated it independently against this head. It is accurate, and the inline finding on `composite.tsx` is the concrete form of it with a reproduction and a measured base-versus-head differential. Codex completed with no findings; nothing else in the bot comments is review-tool feedback.

### Candidates I discovered and did not publish inline

- Replacing the `sync` subscription with a recorded `moves` value on the request, which would remove the permanent per-store listener and the lazy `getMoveRequest` setup. **Invalid**: `moves` is public state a consumer can reset, and the PR's own documented workaround does exactly that, so a recorded count is open to an ABA collision where a reset and a new `move()` land back on the same number. `sync` is the right mechanism here.
- The `sync` listener is never unsubscribed. **Invalid or obsolete**: it closes over the request object rather than the store, and the `WeakMap` is keyed by the store, so nothing outlives the store itself.
- `useRef({})` allocates a discarded object on every render. **Declined non-goal**: a preference with no measurable cost in a component that renders only on `moves` and `compositeElement` changes.
- Consolidating `composite-7360` into a semantic fixture. **Declined non-goal**, agreeing with the reassessment recorded in the description. The nearest sibling, `composite-presentation-cancel-on-unmount`, covers presentations stranded by an unmount under virtual focus, not a completed move replayed by a fresh instance, so it does not coherently own this scenario.
- The app sandbox copy of `menubar-navigation` differs from the shipped `examples/menubar-navigation` only in Tailwind class names, so the new menubar test does pin the reported composition. Not a finding.

No follow-up issues were registered from this review.

### Not fully validated

The pending-move regression is proven in happy-dom and by the base-versus-head differential; I did not reproduce it in a real browser or attach a recording, since the evidence is a deterministic test differential rather than a visual state. I also did not exercise the withheld-focus variant of the same class, where a presentation stays pending because its popup is closed rather than because its item has not registered. It looks reachable by the same mechanism, but I did not build a fixture for it.

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Confirmed and fixed in [`259030c`](https://github.com/ariakit/ariakit/commit/259030cda). Reproduced before changing anything: the pending-move scenario passes on the merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10) and fails at [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05), which isolates the stale claim exactly as described.

Recording the claim on carry-out alone turned out to be wrong in the other direction, though, so the fix went further than the prototype. A move has three outcomes, not two:

- **carried out**, when focus reaches the item or the container branch focuses the container;
- **given up on**, when the popup closes, the user moves to another item, focus leaves the composite, or a new presentation replaces this one;
- **still waiting**, when the item has not registered or a closed popup has not opened.

Only the last is handed on. Recording on read collapses the first two states into one and drops pending requests, which is this thread. Recording only on carry-out collapses the middle state into the last and replays requests the composite gave up on: asking for a tool that has not loaded and then clicking another tool leaves the abandoned request unclaimed, so taking composite behavior back moves focus to the clicked tool. I hit that while validating the prototype and it reproduces in the fixture.

So `presentItem` reports consumption rather than focus. `onConsume` fires where it delivers focus and at every point where it gives up, and its returned `cancel` carries a `settle` alongside it so `usePresentItem` can retire a live presentation it is replacing, while an owner cancelling one on unmount still hands the request on. That last part matters: `useComposite` shares one `usePresentItem` between its own `onFocus` and `CompositeFocusOnMove`, so the composite element taking focus can supersede a request that is still waiting.

I used that signal rather than the `getEnabledItem(store, activeId)?.element?.isConnected` proxy because the proxy still claims for a presentation whose focus is withheld for a closed popup: the item is connected, so the proxy records a claim, but nothing has been focused yet.

Keeping the record keyed on the component instance still matters. With a plain flag a same-instance effect re-run would be refused after consumption, and because a re-run first runs the previous cleanup and cancels the pending presentation, a request still owing a scroll would be cancelled and never restarted.

Coverage, all red-checked against the version of the fix that collapses each state:

- `app/src/sandbox/composite-7360/` gains a highlight tool that loads on demand, with a control test, this thread's regression test, and the abandoned-request test. The regression test fails at `8ecfb7b`; the abandoned-request test fails when consumption is limited to carry-out.
- `app/src/sandbox/composite-superseded-move/` covers the superseded request in [`259030c`](https://github.com/ariakit/ariakit/commit/259030cda). The toolbar cannot express it, because taking a request over needs a composite element that can take focus while an item is active and a roving-tabindex container is focusable only when none is. Its reproduction test fails in Chromium, Firefox and WebKit when the superseding call is reduced to a plain cancel, with both controls passing.

One evidence gap, stated rather than implied: `consumeMove` also drops a consumption that arrives after the move count has moved on, and that guard has no red check. It needs a store write to re-enter a live presentation between a `moves` change and React flushing the effect whose cleanup would cancel it, and I found no reachable consumer path. It compares the count by value, so it does not cover a reset that returns to the same number.

Review of the fix also turned up a second route to the #7360 symptom that this change does not close, and that reproduces the same way with the merge-base library files swapped in: a pending move whose item never registers, followed by `setActiveId(null)` and a remount, makes the container take focus and scroll. It is pre-existing rather than introduced here, and closing it means recording what a move asked for rather than only what became of it, so it is tracked in https://github.com/ariakit/ariakit/issues/7378 and noted in the description.

app/src/sandbox/composite-7360/test-browser.ts
diegohaz: Fixed in [`c9e7193`](https://github.com/ariakit/ariakit/commit/c9e71938a): the helpers now build their own `q` from `page` with the `query` that `withFramework` passes to its callback, so the `q` parameter, the `type Query = ReturnType<typeof query>` alias and its type-only import are all gone. `handOffControl` takes `page` for the same reason.

On the wider question, the explicit decision: this file moves to `query(page)` and the two pre-existing siblings keep the alias. `const q = query(page)` is the dominant convention in the app browser suite, so `composite-presentation-cancel-on-unmount` and `popover-disclosure-7087` are the two outliers rather than a second standard this file was joining. Converting them is a preference-level cleanup in files this pull request does not otherwise touch, so it stays out of scope here instead of adding an unrelated diff to a bug fix. Happy to do it as a standalone change if you would rather not leave the two behind.

app/src/sandbox/composite-7360/test-browser.ts
diegohaz: Fixed in [`c9e7193`](https://github.com/ariakit/ariakit/commit/c9e71938a): the link now sits in the sentence it supports, on its own `// Related:` line so the permalink keeps the separate-line form the code style asks for. Applied in both this file and the happy-dom duplicate at `app/src/sandbox/composite-7360/test.ts`.

One deviation from the second option you offered, which was to label this test with #7360 instead. I left the issue-number position empty here. Running the suite against the merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10) shows this test passing there, so it does not reproduce #7360; it is a preservation test, like the two `focus command while collapsed` and `while handed off` tests next to it, which carry no URL either. The four tests that do fail at the merge base keep the label. Happy to add it back if you would rather the position mark the issue a test belongs to rather than the one it reproduces.

.changeset/300-composite-remount-focus.md
diegohaz: Fixed in [`0367097`](https://github.com/ariakit/ariakit/commit/0367097ac): cut to one summary sentence, keeping the two facts a reader acts on, what the component does wrong and what brings it back, and dropping the mechanism detail, which belongs in the issue.

It reads "replaying an earlier `move` call" rather than "after an earlier `move` call", because a move that is still waiting is deliberately carried out across a remount, and only one already carried out or given up on is withheld. Reference links and the `#composite-1` and `#move` anchors were re-verified after the edit.

diegohaz:
**Fixed**

- [`259030c`](https://github.com/ariakit/ariakit/commit/259030cda) ([reply](https://github.com/ariakit/ariakit/pull/7368#discussion_r3900615099)): the claim recorded which instance had read the move, not which had done something about it, so a request still waiting for its item was dropped when that instance was replaced. It now records consumption, and a move has three outcomes rather than two: carried out, given up on, or still waiting, with only the last handed on. Validating the smaller fix showed that recording on carry-out alone replays a request the composite gave up on, so `presentItem` reports consumption through a private `onConsume` and its `cancel` carries a `settle` for a presentation being replaced rather than unmounted. Coverage in [`259030c`](https://github.com/ariakit/ariakit/commit/259030cda) and [`c9e7193`](https://github.com/ariakit/ariakit/commit/c9e71938a), red-checked against a version of the fix that collapses each state into another.
- [`c9e7193`](https://github.com/ariakit/ariakit/commit/c9e71938a) ([reply](https://github.com/ariakit/ariakit/pull/7368#discussion_r3900615666)): the browser test helpers build their own `q` from `page`, dropping the `q` parameter, the `Query` alias and its type-only import.
- [`c9e7193`](https://github.com/ariakit/ariakit/commit/c9e71938a) ([reply](https://github.com/ariakit/ariakit/pull/7368#discussion_r3900616023)): the `#7363` link moved into the sentence it supports, on its own `Related:` line. The issue-number position is now empty on the two tests that pass at the merge base, since those are preservation tests rather than reproductions.
- [`0367097`](https://github.com/ariakit/ariakit/commit/0367097ac) ([reply](https://github.com/ariakit/ariakit/pull/7368#discussion_r3900616128)): the changeset is one summary sentence, and says "replaying an earlier `move` call" because a pending move is still deliberately carried out across a remount.

**Tracked**

- [#7378](https://github.com/ariakit/ariakit/issues/7378): a second route to the same symptom, found while reviewing this change. A pending move whose item never registers, followed by `setActiveId(null)` and a remount, makes the container take focus and scroll. It reproduces the same way with the merge-base library files swapped in, so it is pre-existing rather than introduced here, and closing it means recording what a move asked for rather than only what became of it.

**Evidence gaps**

- `consumeMove` drops a consumption that arrives after the move count has moved on, and that guard has no red check: it needs a store write to re-enter a live presentation between a `moves` change and React flushing the effect whose cleanup would cancel it, and no reachable consumer path was found. It compares the count by value, so it does not cover a reset that returns to the same number.

**Verification**

1975 Vitest tests, 1096 under React 18, `tsc` clean, and 72 browser tests across Chromium, Firefox and WebKit for `composite-7360`, `composite-superseded-move` and `menubar-navigation`. A full browser run earlier in the branch was 2213 passing with no failures.

app/src/sandbox/composite-7360/test-browser.ts
```diff
@@ -0,0 +1,300 @@
+import type { Page } from "@playwright/test";
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+import { recordScrollEvents } from "#app/test-utils/scroll.ts";
+
+withFramework(import.meta.dirname, async ({ test, query }) => {
+  /**
+   * Moves the active item with the keyboard. The move is what arms the replay:
+   * it leaves a move count behind that a fresh effect instance reads as a focus
+   * request to carry out when the toolbar comes back, whether it comes back by
+   * mounting again or by becoming a composite again.
```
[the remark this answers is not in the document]
diegohaz: Validated against the current head and not reproduced. The comment does not present the replay as expected behavior: it opens with "The move is what arms the replay", which scopes everything after it to the failure mode this helper sets up. That is the "what the test would otherwise fail to catch" note, and it is why `moveToSecondItem` exists at all rather than the test clicking Bold directly.

The proposed replacement also loses accuracy. "The move is what arms the replay: the current instance carries it out" reads against itself, because a request the current instance carried out is exactly the one that must never be replayed, so it cannot also be what arms a replay. It drops the leftover move count as well, and that count is the whole reason this helper is the arming step rather than an ordinary click.

Verified by running the suite rather than by reading alone. `keeps focus when the toolbar comes back after a move`, `keeps focus through repeated remounts` and `keeps focus when composite behavior comes back` all fail on the merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10) and pass at [`0367097`](https://github.com/ariakit/ariakit/commit/0367097ac413d99ad56c447a3fc0f3648c939416), which is the arming the comment describes.

One nearby observation that is not this finding and needs no action. The file now also holds requests a fresh instance *must* carry out, through `askForHighlight`, so "a fresh effect instance reads as a focus request to carry out" could be read as a general rule rather than as the pre-fix behavior. If that is ever worth closing, the smaller edit is to scope the existing sentence in place, for example "which, before this fix, a fresh effect instance carried out when the toolbar came back", rather than to replace the paragraph.

diegohaz (commented):
## Re-review summary

Re-review against the baseline from my previous marked review, [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05), through the current head [`0367097`](https://github.com/ariakit/ariakit/commit/0367097ac413d99ad56c447a3fc0f3648c939416). Three new commits, nine files. I also re-read the whole of `presentItem` and `usePresentItem` rather than the diff alone, because the consumption seam now reaches `CompositeItem` through the same hook.

**No new findings.** Every finding from the previous review is fixed, and the fix went to the deeper of the two options I offered.

### The correctness finding is properly closed

`259030c` replaced claim-on-read with consume-on-carry-out, and routed it through a real completion signal (`onConsume`, with `settle` for a presentation that is finished with its request and `cancel` for one that hands it on). That covers a case my suggested local proxy did not: a request withheld because its popup is closed now stays unconsumed as well, since the withheld branch returns before both `consume()` and `settle()`. The gap I recorded as unvalidated last time is closed by construction, though I verified that by reading the branch rather than with a Combobox-shaped fixture.

Verified by differential rather than by inspection:

| Test | merge base [`4f4f0cd`](https://github.com/ariakit/ariakit/commit/4f4f0cd187ecac15ecd388efb737e78ed5562a10) | previous fix [`8ecfb7b`](https://github.com/ariakit/ariakit/commit/8ecfb7b7d35f626f5e1c2f7002bb9f1ecb522a05) | head [`0367097`](https://github.com/ariakit/ariakit/commit/0367097ac413d99ad56c447a3fc0f3648c939416) |
| --- | --- | --- | --- |
| toolbar comes back after a move | fail | pass | pass |
| repeated remounts | fail | pass | pass |
| composite behavior comes back | fail | pass | pass |
| after an abandoned request | fail | pass | pass |
| after the palette took the request over | fail | pass | pass |
| tool loads after composite behavior comes back (both sandboxes) | pass | **fail** | pass |

The middle column is what makes this coverage worth having: the two pending-move tests fail against the previous fix and pass on the base, so they pin the over-correction rather than the original defect, and the five in the top group pin the defect. Neither group alone is satisfiable by a wrong fix in the other direction.

### What I ran

- `pnpm test --run` at head: 280 files, 1975 tests, all passing.
- The same sandboxes with both `composite.tsx` and `utils.ts` reverted to the merge base: 5 failed, 11 passed. Reverted to `8ecfb7b` instead: 2 failed, 14 passed. Those are the columns above.
- `pnpm lint` and `pnpm tsc`: both clean.
- CI at this head is green across Chrome, Firefox, Safari, iOS, Android, React 18 and Visual.

### Previous findings, all resolved

- [`composite.tsx` claim-on-read](https://github.com/ariakit/ariakit/pull/7368#discussion_r3899583031): fixed in `259030c`, verified above. Thread resolved.
- [#7363 link placement](https://github.com/ariakit/ariakit/pull/7368#discussion_r3899583629): fixed in `c9e7193`, now on its own `// Related:` line in both the browser test and the happy-dom duplicate. Thread resolved.
- [changeset detail](https://github.com/ariakit/ariakit/pull/7368#discussion_r3898810072): fixed in `0367097`. The entry is one sentence, and "replaying an earlier `move` call" is a better choice than the "after an earlier `move` call" I suggested, because a still-pending move is deliberately not covered by the fix. Thread resolved.
- [helpers taking `q`](https://github.com/ariakit/ariakit/pull/7368#discussion_r3898835118): fixed in `c9e7193`, and the new `composite-superseded-move` tests use the same shape, so the two sandboxes added here are consistent with each other. Thread resolved.

### Unresolved thread validated this run

- [CodeRabbit on `test-browser.ts:10`](https://github.com/ariakit/ariakit/pull/7368#discussion_r3900638128), asking to rewrite the `moveToSecondItem` comment: **invalid**. The comment opens with "The move is what arms the replay", which scopes it to the failure mode the helper sets up rather than to expected behavior, and the proposed replacement contradicts itself and drops the leftover move count that makes this helper the arming step. Replied on the thread with the reasoning and with the smaller in-place edit worth considering if the general reading ever needs closing.

### Candidates I considered and did not publish

- `composite-superseded-move/test-browser.ts` asserts only focus, no layout or scroll, which `ariakit-ariakit-example-playwright-test` normally routes to `test-chrome.ts`. **Declined non-goal**: its fourth test clicks a non-native focusable container at a fixed position and expects it to take focus, which is exactly the macOS Safari click-focus contract the skill sends to `test-browser.ts`, and splitting a four-test sandbox across two project files to save two engines on the other three trades more than it buys.
- `onConsume` is a new member on the exported `PresentItemParams`, reachable through the public `./composite/utils` subpath, without an `unstable_` prefix. **Invalid**: it matches the existing `@private` convention of `scrollIntoView` in the same interface, and `usePresentItem` generates no reference page (`https://ariakit.com/reference/use-present-item` returns 404), so no public reference metadata changes.
- `Object.assign(cancel, { settle })` attaches a method to the returned canceller. **Declined non-goal**: the return value is used directly as a React effect cleanup, so it has to stay callable, and a `{ cancel, settle }` object would change every call site to buy nothing.
- Keeping both the `sync` reset and the count check inside `consumeMove` looks redundant at a glance. **Invalid**: they do different jobs. The `sync` reset makes each new request start unconsumed, and the count check drops a consumption that arrives after the count has moved on, which is what makes an asynchronous `onConsume` safe.
- The new `composite-superseded-move` sandbox rather than another case in `composite-7360`. **Invalid**: it needs a focusable composite container, and adding `tabIndex` to the existing toolbar would change that fixture's contract for its other nine tests.

### Not fully validated

The withheld-focus path, where a presentation stays pending because its popup is closed rather than because its item has not registered, is covered by reading the branch order in `presentItem` rather than by a fixture. Building one needs a Combobox-shaped composite whose element sits outside its popup, which I did not do.

PR #7380, opened 2026-09-01
Preserve Composite move targets across active item changes

### LINKED ISSUE #7378: Composite carries a pending move out for the wrong target after setActiveId
### Summary

A [`move`](https://ariakit.com/reference/use-composite-store#move) that targets an item which never registers stays pending. If [`setActiveId`](https://ariakit.com/reference/use-composite-store#setactiveid) then points the store at the container, that pending request is carried out as a request for the container: the next time the widget is a composite again, it takes focus itself and scrolls the page to itself, although nothing asked for it.

### Reproduction

Found while reviewing #7368, and reachable from the sandbox that pull request adds, `app/src/sandbox/composite-7360/`, with no new controls:

```ts
test("pending move redirected by setActiveId(null)", async () => {
  // Asks for a tool that has not loaded, so the request stays pending.
  await click(q.button("Focus highlight tool"));
  // Collapsing calls store.setActiveId(null) and unmounts the toolbar.
  await click(q.button("Collapse toolbar"));

  await click(q.button("Expand toolbar"));

  expect(q.button("Collapse toolbar")).toHaveFocus();
});
```

The toolbar takes focus instead of the button the user clicked.

### Cause

`move(id)` writes `activeId` and then increments `moves`. Nothing carries the request out while the item is missing, and `setActiveId(null)` does not retire it either: `presentItem`'s `abandonedByState` deliberately exempts a null active id, because a composite clears its own active item when its element takes focus while the item it was showing is still worth presenting.

The request therefore stays alive with `activeId === null`, which is indistinguishable from the state a real `move(null)` leaves behind. `CompositeFocusOnMove`'s container branch reads that state and focuses and scrolls to the container.

The mirror case is the same shape: a pending `move(null)`, then `setActiveId(someItemId)`, then a remount, presents that item instead.

### Expected

A request should be carried out for what it asked for. A pending `move(id)` whose target never arrives should not turn into a request for the container, and a pending `move(null)` should not turn into a request for an item.

### Scope

**Pre-existing, and not a regression from #7368.** Verified by swapping the merge-base `composite.tsx` and `utils.ts` into the branch: the reproduction above fails there too, so it behaves the same on `main`. #7368 narrows the replay that #7360 reported, where a move that was carried out was replayed after a remount, and it does not reach this case.

Popup-backed stores are unaffected, because `abandonedByState`'s `mounted` and `open` latches retire the request when the popup closes. It applies to plain composite stores, so [`Toolbar`](https://ariakit.com/reference/toolbar), [`Menubar`](https://ariakit.com/reference/menubar), [`TabList`](https://ariakit.com/reference/tab-list) and [`RadioGroup`](https://ariakit.com/reference/radio-group).

It needs a `move` for an item that never registers, a later `setActiveId` that changes the shape of the target, and a remount or a [`composite`](https://ariakit.com/reference/composite#composite-1) prop toggle, so it is not everyday usage.

### Direction

Recording what the move asked for, alongside the record #7368 already keeps per store, would let each branch check that the request is one it can carry out. The `sync` on `moves` that #7368 adds already fires at the instant `activeId` still holds the move's target, because `move` writes `activeId` first.

### Workaround

See the [verified keep-mounted workaround](https://github.com/ariakit/ariakit/issues/7378#issuecomment-5488387284).

diegohaz:
## Motivation

A pending [`move`](https://ariakit.com/reference/use-composite-store#move) can be carried out for a different target after [`setActiveId`](https://ariakit.com/reference/use-composite-store#setactiveid) changes the active item and the [`Composite`](https://ariakit.com/reference/composite) remounts. An item move can become a container move, and a container move can become an item move. In either case, focus leaves the control the user clicked even though the later state update did not request focus.

```ts
store.move("unavailable");
store.setActiveId(null);

store.move(null);
store.setActiveId("bold");
```

## Solution

Start recording move targets when the React Composite store hook creates the store, before any Composite component mounts. The request listener watches `activeId` and `moves`: a standalone active item change cancels the pending target, while a real move's following count update records its requested target, including propagated moves. The React focus handler checks that the current active item still matches that target before it focuses an item or the composite element.

The request record follows the shared core store across wrapper objects returned by derived hooks such as `useTabStore`. The sandbox covers both mirror sequences after a remount and before the first Composite mount, active item changes that cycle away and back while unmounted, and a derived Tab store before its first TabList mount.

## Workaround

Until the fix ships, keep the Composite instance mounted and hide it instead of conditionally rendering it:

```tsx
// TODO: Restore conditional rendering after https://github.com/ariakit/ariakit/issues/7378 is fixed.
<Ariakit.Composite
  hidden={!showPendingComposite}
  store={pendingStore}
  role="toolbar"
  aria-label="Pending actions"
>
  {/* items */}
</Ariakit.Composite>
```

This preserves the effect instance, so a later `setActiveId` update cannot be reinterpreted by a replacement instance. The limitation is that the component, its effects, its state, and its DOM remain mounted while hidden. Use it only when retaining those resources is acceptable.

## Preview

Before the fix, the pending item request is redirected to the remounted toolbar. The toolbar takes focus and the page scrolls after the final Show click:

[issue-7378-before.webm](https://github.com/user-attachments/assets/12f3ebaf-acc0-451a-b07c-a6172c62db9b)

After the fix, the same sequence leaves focus on the Show/Hide button and keeps the page at its original scroll position:

[issue-7378-after-complete.webm](https://github.com/user-attachments/assets/718e9659-a9cd-4d3b-888a-dbde251415dd)

<details>
<summary>Cycle 3: Continue the narrow core move-target record</summary>

**Assessment**

The issue has medium severity and low expected frequency. It requires an unavailable target, a later `setActiveId` that changes the target shape, and a remount or `composite` toggle. When it occurs, it can steal focus and scroll the page, with a high impact on keyboard and assistive-technology users. The supported scope includes the stable Composite store contract, first mounts, remounts, and composed-store propagation for Composite and derived components.

The implementation has medium-high complexity. It coordinates a private core request record with the existing React consumption record, distinguishes direct moves from propagated counters, and requires synchronized package releases. The repeated review findings came from the same constraint: `activeId` and the numeric `moves` counter do not retain durable request identity.

**Decision**

Continue the current design. Capturing the target in the core `move` path is the narrowest approach that covers requests made before the first React mount. A React-only record restores that bug, a public move-target state adds a larger contract, and cancelling every request in `setActiveId` breaks supported pending presentation behavior. No edge cases are intentionally excluded. Publish `@ariakit/components`, `@ariakit/react-components`, and `@ariakit/react` together because the React handler depends on the new private core record.

</details>

<details>
<summary>Cycle 4: Replace the cross-package record with React-local tracking</summary>

**Assessment**

The issue severity, expected frequency, supported scope, and likely user impact are unchanged. The new review finding showed that the React listener can record the target before the first Composite mount when it starts in the store hook rather than in the component.

The React-local direction has lower implementation and maintenance complexity. It removes the private core record, request flags, structural cross-package access, core-only tests, and the coordinated `@ariakit/components` release. A stable function from the shared core store identifies wrapper copies returned by derived hooks.

**Decision**

Simplify the implementation. Register the existing React request record in `useCompositeStoreProps`, record `activeId` whenever `moves` changes, and keep the exact-target check in the focus effects. A public derived-store regression verifies that wrapper copies share the pending request. This supersedes the cycle 3 core-record direction without narrowing supported behavior.

</details>

<details>
<summary>Cycle 3 after simplification: Keep React-local tracking</summary>

**Assessment**

The issue remains medium severity with low expected frequency and high impact when triggered because redirected focus or scrolling can disrupt keyboard and assistive-technology users. The supported scope includes React Composite-family stores, derived stores such as `useTabStore`, direct and propagated moves, remounts, and moves before the first Composite mount.

The React-local runtime has medium, bounded implementation and maintenance complexity. It relies on the established `activeId`-before-`moves` write order, a private WeakMap keyed by the shared `store.item` function, and exact-target checks in two focus branches. The three findings in this direction corrected passive-effect timing coverage, prevented an accidental public export, aligned desktop-browser coverage, and updated documentation for the two distinct `moves` subscriptions. They reduced test and contract risk without adding runtime state or branches.

**Decision**

Continue the React-local design. Removing the request record would restore the bug, while returning to the core-record approach would restore cross-package machinery and a coordinated release. No edge case is intentionally unsupported within the React scope. Standalone core stores do not perform the affected DOM-focus presentation, and popup-backed stores retain their existing cancellation behavior.

</details>

Fixes #7378


packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -321,8 +328,25 @@ export function createCompositeStore<
     focusShift: defaultValue(props.focusShift, syncState?.focusShift, false),
   };
 
+  const moveRequest: CompositeMoveRequest = {
+    moves: initialState.moves,
+    requested: false,
+    requesting: false,
+    targetId: initialState.activeId,
+  };
```
diegohaz: The cross-package move record can be removed. The whole mechanism — the `CompositeMoveRequest` shape, the `requested`/`requesting`/`moves` bookkeeping, the `try`/`finally` in `move`, the `__unstableCompositeMoveRequest` seam, the `CoreMoveRequest` interface that `composite.tsx` re-declares structurally, and the synchronized three-package release — exists to answer one question: *what item did the pending move ask for?*

The React package can already answer it on its own. `move` writes `activeId` before `moves`, so inside `sync(store, ["moves"], ...)` the answer is simply `store.getState().activeId`, for a direct move and for a count propagated from a connected store alike. The reason the React record cannot answer it today is not the store layer — it is that `getMoveRequest` is first called by `CompositeFocusOnMove`, so no subscription exists while no composite is mounted. Registering the same record from `useCompositeStoreProps` closes that gap: `sync` attaches a listener directly, so it works before `init`, and every composite-family store hook (`useTabStore`, `useSelectStore`, `useComboboxStore`, `useMenuStore`, `useMenubarStore`, `useToolbarStore`, `useRadioStore`, `useTagStore`) funnels through it.

`packages/ariakit-react-components/src/composite/composite-store.ts`:

```ts
  store = useCollectionStoreProps(store, update, props);
  // Start recording move targets as soon as the store exists, so a move made
  // before any composite mounts is not carried out for a later active item.
  getMoveRequest(store);
```

`packages/ariakit-react-components/src/composite/utils.ts` (the record and its helpers move here so both files can use them):

```ts
  sync(store, ["moves"], () => {
    request.consumedBy = null;
    request.targetId = store.getState().activeId;
  });
```

`packages/ariakit-react-components/src/composite/composite.tsx`, one line in each effect:

```ts
    if (activeId !== getMoveRequest(store).targetId) return;
```

`packages/ariakit-components` then needs no change at all, and the changeset drops to `@ariakit/react-components` and `@ariakit/react`.

Validated locally on `ddef361c6` with this alternative applied:

- All four new #7378 regression tests pass, in `test.ts` and in `test-browser.ts` (Chrome), including both pre-first-mount cases.
- `pnpm test` 1979 passed, `pnpm test-react18` on the sandbox 7 passed, `pnpm tsc` and `pnpm lint` clean.
- `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` 549 passed.

Two alternatives I also built and measured, for the record. Collapsing `requested`/`requesting` into a single `{moves, targetId}` pair recorded inside the `setState("moves", …)` updater keeps the core record but removes the flags, and passes the full suite — a smaller step if the record is worth keeping. Promoting the target to a `@private unstable_moveTargetId` state key (the `unstable_placing` precedent) also passes everything, but it grows `CompositeStoreState`, has to be paired with `moves` in every key list such as `tab-store.ts` `independentKeys`, and only records once the store is initialized, so it needs the `preserves the target of each move` test rewritten.

**Assessment**

- Severity: Low (verified, 95% confidence), because the current code is correct; the cost is maintenance surface, not user-visible behavior.
- Priority: P1 (estimate, 75% confidence), because the private record fixes a cross-package contract and a coordinated release cadence that the simpler form does not need, and that is cheapest to avoid before merge.
- Breakage likelihood: Possible (estimate, 70% confidence), for the untyped seam specifically: `getCoreMoveRequest` uses optional chaining, so a consumer on a newer `@ariakit/react-components` with an older `@ariakit/components` inside the caret range silently loses the fix instead of failing.
- Fix effort: Easy (verified, 90% confidence), because the working alternative is 3 files inside one package, +80/−154 lines against this branch.

packages/ariakit-components/src/composite/composite-store.ts
```diff
@@ -321,8 +328,25 @@ export function createCompositeStore<
...
+
   const composite = createStore(initialState, collection, props.store);
 
+  setup(composite, () =>
+    sync(composite, ["moves"], (state, prevState) => {
+      if (state.moves === prevState.moves) return;
```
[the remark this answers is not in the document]
diegohaz: Confirmed on `ddef361c6`. Adding this to `composite-store.test.ts` fails with `expected 'old' to be 'new'`:

```ts
test("pre-init move record vs composed state", () => {
  const source = createStore({ activeId: "old", moves: 0 });
  const store = createCompositeStore({ store: source });

  store.move("old");
  source.setState("moves", 0);
  source.setState("activeId", "new");
  source.setState("moves", 1);

  const destroy = init(store);
  expect(store.getState()).toMatchObject({ activeId: "new", moves: 1 });
  expect(getMoveTarget(store)).toBe("new"); // receives "old"
  destroy();
});
```

`storeInit` pushes the connected store's keys before it runs the setups, and the push of `moves` is a no-op here because the count already matches, so the sync never observes it and the record keeps `{moves: 1, requested: true, targetId: "old"}`.

The cause is not this early return, though, and no reconciliation at `setup` time fixes it. At that point the composed state is `{activeId: "new", moves: 1}` and the record is `{moves: 1, targetId: "old"}` — the same pair a genuine pending move produces. Reconciling would restore the #7378 bug for any move made before `init`, which the `does not redirect … before the first composite mount` cases cover. A counter carries no request identity, so these two histories are indistinguishable once they converge on the same count.

Reachability is what keeps this out of the current change. It needs `move` to run on a store that is connected to another store and not yet initialized, and the connected store to reset its count and return to the same value in that same window. Supported React usage does not reach it: `useStore` initializes in a layout effect, and user code only receives the store after the hook returns, so a move from an event handler or an effect is always post-init.

I am recording this as a declined non-goal for this PR: it is an inherent limit of counter-based identity rather than an oversight in this diff, and every alternative I built and tested behaves the same way at that point. It also shrinks under the simplification in https://github.com/ariakit/ariakit/pull/7380#discussion_r3901267323, which removes the pre-init core record entirely.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -123,25 +127,48 @@ interface MoveRequest {
...
  */
 const moveRequests = new WeakMap<CompositeStore, MoveRequest>();
 
 function getMoveRequest(store: CompositeStore) {
   const cached = moveRequests.get(store);
   if (cached) return cached;
-  const request: MoveRequest = { consumedBy: null };
+  const request: MoveRequest = {
+    consumedBy: null,
+    targetId: store.getState().activeId,
```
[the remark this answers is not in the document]
diegohaz: Not valid on `ddef361c6`. `sync` is "called immediately and synchronously whenever the store state changes" (`packages/ariakit-store/src/index.ts`), so registering the listener runs the callback once right away, with the current `moves`. That first call already applies the `coreRequest?.requested && coreRequest.moves === state.moves` condition, and it overwrites the initializer before any effect reads `targetId`.

The initializer value is therefore never observed. Replacing it with a deliberately impossible value on this branch keeps every sandbox test green:

```ts
  const request: MoveRequest = {
    consumedBy: null,
    targetId: "__SCRATCH_BOGUS_INITIALIZER__",
  };
```

`pnpm test app/src/sandbox/composite-late-item-focus/test.ts` — 7 passed, including `does not redirect an item move before the first composite mount` and `does not redirect a composite move before the first composite mount`. Those two are exactly the sequences quoted in this comment, and they pass unchanged on the current head, so the prediction that they would fail does not hold.

diegohaz (commented):
## Review

Initial review of `ddef361c6`, diffed against the merge base with `main`. The fix is correct and the coverage is right: four regression cases in both `test-browser.ts` and the happy-dom duplicate, covering the item-to-container and container-to-item redirects after a remount and before the first mount. The description matches the implementation. The one substantive finding is that the same behavior is reachable without the new cross-package contract.

### Simpler alternative, built and measured

Posted inline: https://github.com/ariakit/ariakit/pull/7380#discussion_r3901267323

The target of a pending move is already derivable in the React package. `move` writes `activeId` before `moves`, so `store.getState().activeId` inside `sync(store, ["moves"], …)` is the requested target, for a direct move and for a propagated count alike. The only reason the existing React record cannot supply it is that `getMoveRequest` is first called by `CompositeFocusOnMove`, so nothing is subscribed while no composite is mounted. Registering that record from `useCompositeStoreProps` instead closes the gap — `sync` attaches a listener directly, so it is live before `init`, and every composite-family store hook funnels through that function.

That removes from this change: `CompositeMoveRequest`, the `requested`/`requesting`/`moves` bookkeeping, the `try`/`finally` in `move`, the `__unstableCompositeMoveRequest` seam and its `@ts-expect-error`, the `CoreMoveRequest` interface `composite.tsx` re-declares structurally, the two core tests that reach through the seam, and the coordinated release of three packages. `@ariakit/components` needs no change; the changeset becomes `@ariakit/react-components` and `@ariakit/react`.

Validated with the alternative applied on top of `ddef361c6`:

- `pnpm test app/src/sandbox/composite-late-item-focus/test.ts` — 7 passed; `pnpm test-react18` on the same file — 7 passed.
- `pnpm test` — 1979 passed.
- `pnpm tsc` — 0 errors. `pnpm lint` — clean.
- `pnpm test-chrome src/sandbox/composite-late-item-focus` — 9 passed, including all four new #7378 cases.
- `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` — 549 passed.

Two further alternatives I implemented and ran, neither of which I recommend over the above:

- Keeping the core record but collapsing it to `{moves, targetId}`, recorded inside the `setState("moves", …)` updater so it lands before listeners run. Drops `requested`, `requesting` and the `try`/`finally`; full suite green. A smaller step if the record is worth keeping.
- Promoting the target to a `@private unstable_moveTargetId` state key, following the `unstable_placing` precedent. Full suite green and `pnpm tsc` clean, but it grows `CompositeStoreState`, must be paired with `moves` in every key list — `tab-store.ts` `independentKeys` needed it — and records only once the store is initialized, so `preserves the target of each move` has to be rewritten. Writing the key inside `move` also perturbed the store's notification order enough to break `combobox-tabs-old > clear input with keyboard`.

### Unresolved threads validated against this head

| Thread | Claim | Outcome |
| --- | --- | --- |
| `PRRT_kwDOBjT0Fc6d-yt7` (`composite-store.ts:342`) | A pre-init `move` record survives a connected store's count reset and reuse | **Confirmed**, reproduced with a failing store test. Declined non-goal for this PR — see the [reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3901270609) |
| `PRRT_kwDOBjT0Fc6d-1pr` (`composite.tsx:159`) | The initial `targetId` snapshot makes the pre-mount tests fail | **Invalid** — see the [reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3901271427) |

On the confirmed one: `storeInit` pushes a connected store's keys before running the setups, and the `moves` push is a no-op when the count already matches, so the record keeps the earlier target while the composed state moves on. It is an inherent limit of counter-based identity rather than a defect in this diff — reconciling at `setup` time would restore the #7378 bug for any move made before `init` — and it needs `move` to run on a connected, uninitialized store, which supported React usage does not reach, since `useStore` initializes in a layout effect and user code only receives the store after the hook returns. Assessment: Severity Low (estimate, 80% confidence), Priority P2 (estimate, 80% confidence), Breakage likelihood Rare (verified by the reproduction requiring a hand-driven counter cycle on an uninitialized store, 90% confidence), Fix effort Hard (estimate, 70% confidence, because the counter cannot separate the two histories).

I also read both top-level bot review bodies. Neither adds an actionable claim beyond the two threads above.

### Minor notes, not published inline

- The two new core tests assert through `store.__unstableCompositeMoveRequest` with `@ts-expect-error`. The behavior they protect is already covered by the four sandbox cases through the public surface. They disappear under the alternative above; if the record stays, they are acceptable as store-level contract coverage.
- `getCoreMoveRequest` uses optional chaining, so a consumer on a newer `@ariakit/react-components` with an older `@ariakit/components` inside the caret range silently loses the fix rather than failing. The changeset releases the three packages together, so this only bites on a mixed install.
- The sandbox additions append to the existing `composite-late-item-focus` fixture and sit after the existing controls, so the earlier `press.Tab()` sequences are unaffected. Confirmed by the full browser run above.

### Not fully validated

Firefox, Safari and mobile Playwright projects were not run for the alternative; only the `chrome` project. Visual tests were not run, per the local policy.

packages/ariakit-components/src/composite/composite-store.ts
diegohaz: Fixed in [`10111f7a7`](https://github.com/ariakit/ariakit/commit/10111f7a7e11143ea7c3f2efb5bf0ac5ae0bf6ce): moved move-target tracking into a private React module initialized by the store hook, removed the core seam, and added derived `useTabStore` coverage that fails with wrapper-object keying.

packages/ariakit-components/src/composite/composite-store.ts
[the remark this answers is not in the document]
diegohaz: Superseded by [`10111f7a7`](https://github.com/ariakit/ariakit/commit/10111f7a7e11143ea7c3f2efb5bf0ac5ae0bf6ce): the core move-request record no longer exists, so the pre-initialization reconciliation concern no longer applies.

diegohaz:
**Fixed**

- [`10111f7a7`](https://github.com/ariakit/ariakit/commit/10111f7a7e11143ea7c3f2efb5bf0ac5ae0bf6ce) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3901652967)): Replaced the cross-package request record with private React-local tracking and added derived-store coverage.

**Obsolete**

- [`10111f7a7`](https://github.com/ariakit/ariakit/commit/10111f7a7e11143ea7c3f2efb5bf0ac5ae0bf6ce) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3901656255)): Removing the core request record eliminated the pre-initialization reconciliation case.

app/src/sandbox/composite-late-item-focus/test-chrome.ts
```diff
@@ -0,0 +1,18 @@
+import { flushFrames, withFramework } from "#app/test-utils/preview.ts";
+
+withFramework(import.meta.dirname, async ({ test }) => {
+  // https://github.com/ariakit/ariakit/issues/7378
+  test("does not redirect a derived composite move before the first mount", async ({
```
diegohaz: This case runs only in `chrome`, while its three siblings for the same issue run in `chrome`, `firefox` and `safari` from `test-browser.ts`. All four assert the same contract — a click leaves focus on the button and a pending move is not redirected to a later active item — so the split leaves the family covered by two different project sets with nothing stating why.

I ran this test in the other two desktop projects on this head by copying it to a `test-browser`-matched filename, so either direction is safe:

```
pnpm test-firefox src/sandbox/composite-late-item-focus/test-browser-scratch.ts  # 1 passed
pnpm test-safari  src/sandbox/composite-late-item-focus/test-browser-scratch.ts  # 1 passed
```

Both directions are defensible, so it is worth choosing one deliberately:

- Move this test into `test-browser.ts` with its siblings. The assertion is that focus *stays* on a clicked `<button>`, which is the platform-sensitive path that `tabIndex={0}` on these sandbox buttons exists to serve, and `ariakit-general-pr-review`'s companion Playwright guidance lists platform-sensitive focus as the `test-browser.ts` case.
- Or move all four into `test-chrome.ts`. The asserted contract is a store-and-effect contract rather than an engine difference, and that same guidance says not to use `test-browser.ts` for focus alone when the contract is not platform-sensitive. This also drops six redundant desktop runs from CI.

What I would not keep is the current state, where the derived case silently gets a third of the coverage of the three cases it mirrors.

**Assessment**

- Severity: Low (verified, 90% confidence), because the behavior is covered in `chrome` and in the happy-dom duplicate; only cross-engine coverage differs.
- Priority: P2 (estimate, 85% confidence), because it is a coverage-scope choice rather than a defect.
- Breakage likelihood: Unlikely (verified, 85% confidence), since I ran the case in `firefox` and `safari` on this head and both passed, so no engine-specific failure is currently hidden.
- Fix effort: Easy (verified, 95% confidence), because it is moving one test between two files in the same directory.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -210,12 +181,14 @@ const CompositeFocusOnMove = memo(function CompositeFocusOnMove({
 
   // Present the active item.
   useEffect(() => {
+    const moveRequest = getMoveRequest(store);
```
diegohaz, line 184: Anchored here because the lines it concerns are outside this diff: the `CompositeFocusOnMove` docblock just above, at lines 154-156, still says

> It's only rendered when the `composite` prop is enabled, so the `moves` subscription doesn't run for non-composite widgets.

That stopped being true when the record moved to `__move-request.ts`. There are two `moves` subscriptions now, and only one is still gated by the `composite` prop:

- `useStoreState(store, "moves")` in this component, which is what keeps non-composite widgets from re-rendering. Still gated, still accurate.
- `sync(store, ["moves"], …)` inside `getMoveRequest`, which `useCompositeStoreProps` now registers for every store the composite-family hooks create — `useCompositeStore`, `useTabStore`, `useSelectStore`, `useComboboxStore`, `useMenuStore`, `useMenubarStore`, `useToolbarStore`, `useRadioStore`, `useTagStore` — whether or not any `Composite` renders with `composite` enabled.

Registering it there is what makes the fix work before the first mount, so the subscription belongs where it is. The comment is the part that drifted: it now claims something the code stopped doing, and this file is dense enough that a later reader will take it at face value.

Naming the surviving subscription is enough:

```ts
 * don't re-render this component. It's only rendered when the `composite` prop
 * is enabled, so the `moves` state subscription that drives these renders
 * doesn't run for non-composite widgets. The move request record keeps its own
 * `moves` listener, started by the store hook, so a move made before any
 * composite mounts is still recorded.
```

**Assessment**

- Severity: Low (verified, 95% confidence), because no behavior depends on the comment.
- Priority: P2 (estimate, 80% confidence), because the surrounding comments are load-bearing for anyone changing this file and an inaccurate one invites a wrong assumption about where subscriptions live.
- Breakage likelihood: Unlikely (estimate, 75% confidence), and only indirectly, through a future change made on the stale premise that nothing subscribes to `moves` for non-composite widgets.
- Fix effort: Easy (verified, 95% confidence), a comment edit.

diegohaz (commented):
## Re-review

Re-review of `10111f7a7`, incremental from my previous marked review at `ddef361c6`. The branch was also rebased onto `3d0e69ab2`, so the `popover-store.ts`, `dialog.tsx`, `disclosure-content.tsx`, `__hovercard-trigger.tsx`, `menu-button.tsx` and `composite-item.tsx` entries in a naive `ddef361c6..10111f7a7` range come from #7383 and are not part of this PR. Against the current merge base the PR is 8 files and touches one package.

The core seam is gone and the fix is now React-only. My previous main finding is resolved, and the new commit went further than what I proposed: it found and closed a real gap in my recommendation.

### The derived-store gap was mine, and the fix is right

My proposal keyed the record by the store object. `useTabStoreProps` returns `Object.assign(useMemo(() => ({ ...compositeStore, panels }), …), …)`, so the object `useCompositeStoreProps` records against is not the object `TabList` later hands to `Composite`. Keying by the store loses the record for every derived hook. Keying by `store.item` restores it, because `createCompositeStore` builds its own collection store (`composite-store.ts:275`) and every wrapper copies that function by reference, which makes it a stable identity for the underlying store and unique to it.

Verified that the key is load-bearing rather than defensive — replacing `const key = store.item` with `const key = store` on this head:

```
pnpm test app/src/sandbox/composite-late-item-focus/test.ts
× does not redirect a derived composite move before the first mount
1 failed | 7 passed (8)
```

The new `useTabStore` sandbox case is the right coverage for it, and it is a public-surface reproduction rather than a probe of the private module.

### Verification on this head

- `pnpm test` — 1980 passed. `pnpm test-react18 app/src/sandbox/composite-late-item-focus/test.ts` — 8 passed.
- `pnpm tsc` — 0 errors. `pnpm lint` — clean.
- `pnpm test-chrome src/sandbox/composite-late-item-focus` — 10 passed, all five #7378 cases included.
- `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` — 549 passed, 1 flaky.
- The flake was `menu-modal-composition/test-chrome.ts:121 › interact with modal by clicking on the Slot menu item with the keyboard`, which passed on retry. Re-ran it isolated with `--repeat-each=5 --retries=0`: 175 passed, 0 failed. Not caused by this change.
- `__move-request.ts` does not become a public subpath export: `getPublicFiles` skips `__`-prefixed entries (`packages/ariakit-scripts/src/build.ts:135`), and `@ariakit/react-components` currently generates 236 exports with no `__` entry, matching the existing `__hovercard-trigger.tsx` precedent.
- The changeset is correctly narrowed to `@ariakit/react-components` and `@ariakit/react`. The description's cycle 4 block supersedes cycle 3 explicitly, so the retained cycle 3 text is history rather than a mismatch.

### New findings

Both are P2 and neither blocks merge.

- Test project scope: the derived case runs only in `chrome` while the three cases it mirrors run in three desktop projects — https://github.com/ariakit/ariakit/pull/7380#discussion_r3901833340. I ran it in `firefox` and `safari` on this head and both passed, so nothing engine-specific is hidden; the point is to pick one scope for the family.
- Stale comment: the `CompositeFocusOnMove` docblock still says the `moves` subscription does not run for non-composite widgets, which the move to `__move-request.ts` changed — https://github.com/ariakit/ariakit/pull/7380#discussion_r3901840259.

### Previously validated candidates

All three threads from the last run are resolved and outdated, and I re-checked each against this head rather than relying on that state.

| Thread | Prior outcome | Status on `10111f7a7` |
| --- | --- | --- |
| `PRRT_kwDOBjT0Fc6d_R6z` | Simplification, Address now | Resolved — the core record, its flags, the `@ts-expect-error` seam and the third package are gone |
| `PRRT_kwDOBjT0Fc6d-yt7` | Confirmed, declined non-goal | Obsolete — the pre-initialization core record it described no longer exists |
| `PRRT_kwDOBjT0Fc6d-1pr` | Invalid | Obsolete — the code it anchored to was removed, and the reviewer confirmed the correction |

I also read the newer empty-bodied bot review wrappers; they carry no additional claim.

### Notes, not published inline

- One deliberate semantic change from the previous design is worth recording. The record now stores `store.getState().activeId` at the moment `moves` changes, where the core record stored the `id` argument passed to `move`. These differ only if a listener changes `activeId` between the two `setState` calls inside `move`. In that corner the new behavior presents the item that actually ended up active instead of suppressing the request, which fails open rather than closed and cannot reintroduce #7378, since that issue is about a `setActiveId` after the count settles.
- `const key = store.item` is guarded by the derived test, so it will not rot silently. Its comment explains why the store object is not the key, but not why `item` identifies the store. One clause about `item` being created once per `createCompositeStore` and copied by every wrapper would save the next reader the trip through `collection-store.ts`.

### Not fully validated

The broad cross-component sweep was run in `chrome` only; `firefox` and `safari` were exercised for the new derived case alone. Mobile projects and visual tests were not run, per the local policy.

app/src/sandbox/composite-late-item-focus/test-chrome.ts
diegohaz: Fixed in [`8505d7201`](https://github.com/ariakit/ariakit/commit/8505d7201e98ccbf2ba06ea8b1b80f92c4eee704): moved the derived Tab regression into `test-browser.ts`, so Chrome, Firefox, and Safari now run the same focus contract.

Current-head recording:

[issue7378-derived-current-head.webm](https://github.com/user-attachments/assets/3cc06709-cffb-42fc-a2e9-c02eab2b3782)

packages/ariakit-react-components/src/composite/composite.tsx
diegohaz: Fixed in [`8505d7201`](https://github.com/ariakit/ariakit/commit/8505d7201e98ccbf2ba06ea8b1b80f92c4eee704): the docblock now identifies the `CompositeFocusOnMove` subscription as render-driving and distinguishes it from the store hook’s separate move-request listener.

diegohaz:
**Fixed**

- [`8505d7201`](https://github.com/ariakit/ariakit/commit/8505d7201e98ccbf2ba06ea8b1b80f92c4eee704) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902243066)): Aligned the derived Tab regression with the Chrome, Firefox, and Safari focus suite.
- [`8505d7201`](https://github.com/ariakit/ariakit/commit/8505d7201e98ccbf2ba06ea8b1b80f92c4eee704) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902264958)): Corrected the subscription lifecycle documentation.

packages/ariakit-react-components/src/composite/composite.tsx
```diff
@@ -210,12 +181,14 @@ const CompositeFocusOnMove = memo(function CompositeFocusOnMove({
...
     if (!moves) return;
     if (!focusOnMove) return;
     const instance = instanceRef.current;
     if (!mayActOnMove(store, instance)) return;
     const { activeId } = store.getState();
     if (activeId == null) return;
+    if (activeId !== moveRequest.targetId) return;
```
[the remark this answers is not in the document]
diegohaz: Confirmed on `8505d7201`, and reproducible through the sandbox controls this PR already ships. The container branch needs no new fixture:

```ts
test("cycle container unmounted", async () => {
  await click(q.button("Hide pending toolbar"));
  await click(q.button("Focus pending toolbar")); // move(null)
  await click(q.button("Target bold action")); // setActiveId("pending-bold")
  await click(q.button("Target pending toolbar")); // setActiveId(null)
  await click(q.button("Show pending toolbar"));

  expect(q.button("Hide pending toolbar")).toHaveFocus(); // fails
  expect(q.toolbar("Pending actions")).not.toHaveFocus();
});
```

The same sequence without the hide/show passes, so the asymmetry is real: the toolbar steals focus from the button the user clicked only because the composite was unmounted while the active id cycled. The item branch behaves the same way — with a temporary `pendingStore.move("pending-bold")` control, `move("pending-bold")`, `setActiveId(null)`, `setActiveId("pending-bold")`, then remount, focus lands on `Bold action` instead of staying on the toggle.

This matters for the PR because it is the same user-visible harm as the issue being fixed — a conditionally rendered composite taking focus from the control that rendered it, with no update having asked for focus — and because the reassessment block currently states "No edge case is intentionally unsupported within the React scope."

An exact-target check cannot see this on its own: the id it compares against is correct by the time the composite remounts. The record has to remember that a non-move update happened. I prototyped that on this head — a `superseded` flag set when `activeId` changes without a `moves` change, cleared on every count change, and ORed into the two focus branches:

```ts
  sync(store, ["moves", "activeId"], (state, prevState) => {
    if (state.moves !== prevState.moves) {
      request.consumedBy = null;
      request.targetId = store.getState().activeId;
      request.superseded = false;
      return;
    }
    if (state.activeId !== prevState.activeId) {
      request.superseded = true;
    }
  });
```

Results with that applied: both cycle cases above pass, `pnpm test` 1984 passed, and `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` 550 passed. So the strict rule does not disturb the propagated-move, late-item, or popup-cancellation paths that the narrower target check was protecting.

One design note if you take this. Mirroring `presentItem`'s `abandonedByState` rule instead — supersede only on a change to a different non-null id — is the more conservative option, but it does not close the item-branch case, because there the intervening update is `setActiveId(null)`, which that rule deliberately ignores. The strict rule closes both.

Deferring is also defensible, since the behavior predates this PR and the current guard neither introduces nor worsens it. In that case it needs a follow-up issue and the reassessment sentence above needs qualifying, rather than being left unstated.

**Assessment**

- Severity: Medium (verified, 90% confidence), because focus and scroll move away from the control the user activated, which is disruptive for keyboard and assistive-technology users — the same impact this PR cites for #7378.
- Priority: P2 (estimate, 70% confidence), because it predates this change, though the guard added here is the natural place to close it and the fix is validated.
- Breakage likelihood: Possible (verified by the reproduction, 85% confidence), needing a conditionally rendered composite plus an active id that returns to a pending target while unmounted. Toolbars and menubars toggled by a control are the realistic shape.
- Fix effort: Easy (verified, 85% confidence), one extra field and one listener key, measured against the full unit suite and a 550-test Chrome sweep.

packages/ariakit-react-components/src/composite/__move-request.ts
```diff
@@ -0,0 +1,37 @@
...
+    consumedBy: null,
+    targetId: store.getState().activeId,
+  };
+  moveRequests.set(key, request);
+  // `move` writes `activeId` before `moves`, so the active id at this point is
+  // the target of this request, including moves propagated from another store.
+  sync(store, ["moves"], () => {
+    request.consumedBy = null;
+    request.targetId = store.getState().activeId;
+  });
```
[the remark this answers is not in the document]
diegohaz: The mechanism is confirmed, with one correction: the record no longer carries a `moves` field, so it is not `{moves: 1, targetId: "old"}`. What survives is `targetId: "old"`. The rest holds — `storeInit` pushes a connected store's keys before running the setups, `activeId` before `moves`, and a pushed count that already equals the current one notifies nothing, so this listener never runs and the target is not reconciled.

It is also the same finding as the earlier core-record thread, which I validated and declined on `ddef361c6`. Moving the record into React changed where it lives, not whether a counter can distinguish the two histories, and the new evidence offered here — that the tracker is installed before initialization — is true but does not change the disposition.

What does change the disposition is reachability, and I could not construct a reachable case. It needs all of the following inside the window between the store hook's render and its `init` layout effect: `move` runs on the store, and the connected source resets its count and returns to the same value. A descendant layout effect can occupy that window, so it is not strictly render-phase-only, but the source would have to cycle `moves` down and back in that same commit. The only non-`move` writes to `moves` in the codebase are resets to `0` (`combobox-store.ts` on close, `composite-container.tsx`), and a reset followed by a return to the identical count within one commit, on a store connected to a child that moves before its own initialization, is not a shape supported usage produces.

Contrast with the sibling finding on `composite.tsx`, which I did reproduce with the sandbox controls this PR already ships and which I am recommending be addressed. I am keeping this one as a declined non-goal on the same reasoning as before: an inherent limit of counter-based request identity rather than a defect in this diff, with no reachable consumer scenario.

**Assessment**

- Severity: Medium (estimate, 70% confidence), because if reached it suppresses a legitimate focus request rather than misdirecting one, so the failure is a missing focus move.
- Priority: P2 (estimate, 85% confidence), because no supported usage reaches it and closing it would require request identity the `moves` counter does not carry.
- Breakage likelihood: Rare (verified by failing to construct a reachable case, 85% confidence), requiring a connected source to cycle its count within the single commit that precedes the child store's initialization.
- Fix effort: Hard (estimate, 75% confidence), because reconciling at initialization cannot separate this from a genuine pending pre-mount move, which is the case the sandbox's before-first-mount tests cover.

diegohaz (commented):
## Re-review

Re-review of `8505d7201`, incremental from my previous marked review at `10111f7a7`. One commit since then, `Align Composite move target review coverage`, which addresses both P2 notes from that round. No new source behavior; the only runtime file touched is a comment.

Both of my previous findings are resolved, and I discovered nothing new of my own in the incremental diff. The substance of this round came from validating the two Codex threads opened against this head: one is confirmed and reproducible, the other is confirmed in mechanism but unreachable.

### Previous findings, verified as fixed

- Test project scope. `test-chrome.ts` is deleted and the derived Tab case now sits in `test-browser.ts` with the three cases it mirrors, so all five #7378 browser cases run in the same project set. Verified by running the sandbox in each desktop project on this head: `chrome` 10 passed, `firefox` 10 passed, `safari` 10 passed.
- Stale docblock. The `CompositeFocusOnMove` comment now says "this render-driving subscription doesn't run for non-composite widgets. The store hook tracks move requests separately", which distinguishes the two `moves` subscriptions correctly.

### Verification on this head

- `pnpm test` — 1980 passed. `pnpm tsc` — 0 errors. `pnpm lint` — clean.
- `pnpm test-chrome src/sandbox/composite-late-item-focus` — 10 passed, likewise `test-firefox` and `test-safari`.

### Open threads validated against this head

| Thread | Claim | Outcome |
| --- | --- | --- |
| `PRRT_kwDOBjT0Fc6eB661` (`composite.tsx:191`) | A move superseded by `setActiveId` revives when the active id cycles back | **Confirmed and reproduced.** Address now — [reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902398610) |
| `PRRT_kwDOBjT0Fc6eB66u` (`__move-request.ts:35`) | A pre-initialization target is not reconciled when a connected source returns to an equal count | **Confirmed in mechanism, no reachable case.** Declined non-goal — [reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902402224) |

The confirmed one is worth summarizing here because it is the one open question left on this PR. With the composite unmounted, `move(null)`, `setActiveId("pending-bold")`, `setActiveId(null)`, then remount, the toolbar takes focus from the toggle the user clicked. The identical sequence without the hide and show leaves focus alone, so it is an asymmetry rather than a consistent policy, and it is the same user-visible harm the PR sets out to fix. It reproduces through sandbox controls this PR already ships, and the item branch behaves the same way with a temporary `move("pending-bold")` control.

An exact-target check cannot detect it, because the active id is back at the requested target by the time the composite remounts; the record has to remember that a non-move update intervened. I prototyped a `superseded` flag on this head — set when `activeId` changes without a `moves` change, cleared on every count change, ORed into both focus branches — and measured it: both cycle cases fixed, `pnpm test` 1984 passed, and the composite-family Chrome sweep 550 passed. So the stricter rule does not disturb the propagated-move, late-item, or popup-cancellation paths. The reply carries the patch and the trade-off against the narrower `abandonedByState`-mirroring rule, which does not close the item branch.

The behavior predates this PR, so deferring it is defensible. If you defer, it needs a follow-up issue, and the reassessment block's "No edge case is intentionally unsupported within the React scope" needs qualifying, since this is one.

### Notes, not published inline

- The description's cycle blocks track the implementation accurately: the newest block supersedes cycle 3's core-record decision, and its account of the React-local runtime — the `activeId`-before-`moves` write order, the WeakMap keyed by `store.item`, and the two focus branches — matches the code.
- `pnpm test` reports 1980 here against 1979 at `10111f7a7`, which is the derived Tab happy-dom case that already existed plus the browser move; no unit coverage was lost in the file deletion.

### Not fully validated

The composite-family sweep was run in `chrome` only, both for this head and for the prototype. `firefox` and `safari` were exercised for the `composite-late-item-focus` sandbox alone. Mobile projects and visual tests were not run, per the local policy.

packages/ariakit-react-components/src/composite/composite.tsx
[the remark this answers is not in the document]
diegohaz: Fixed in [`3f8739886`](https://github.com/ariakit/ariakit/commit/3f8739886e9b924ba3eead82c67c2977c9e85273): standalone active ID changes now cancel the pending target, while a real move's following count update records its target. Added targeted container and item cycle coverage in the same commit.

Browser verification on the current head:

[issue7378-active-id-cycle-current-head-short.webm](https://github.com/user-attachments/assets/23dedeca-9f9b-4d80-9475-9bfe9e77383a)

diegohaz:
**Fixed**

- [`3f8739886`](https://github.com/ariakit/ariakit/commit/3f8739886e9b924ba3eead82c67c2977c9e85273) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902723906)): Cancelled pending item and container targets when non-move active ID changes cycle away and back.

packages/ariakit-react-components/src/composite/__move-request.ts
```diff
@@ -0,0 +1,43 @@
...
+  sync(store, ["moves", "activeId"], (state, prevState) => {
+    if (state.moves !== prevState.moves) {
+      request.consumedBy = null;
+      request.targetId = state.activeId;
+      return;
+    }
+    if (state.activeId !== prevState.activeId) {
+      // `move` writes `activeId` before `moves`, so its count update replaces
+      // this cancellation with the requested target.
+      request.targetId = undefined;
```
diegohaz: The cancellation marker is a value from `activeId`'s own domain, so the record no longer distinguishes "this request was superseded" from "this request recorded an unset target". `targetId` is typed `Core.CompositeStoreState["activeId"]`, and `undefined` is a legal active id, not an out-of-band value.

Both states are reachable and produce an identical record. Verified on this head:

```ts
test("a live propagated request records an undefined target", () => {
  const source = createStore({ activeId: undefined, moves: 0 });
  const store = createCompositeStore({ store: source });
  const destroy = init(store);
  const request = getMoveRequest(store);

  source.setState("moves", 1);

  expect(store.getState().activeId).toBe(undefined);
  expect(request.targetId).toBe(undefined); // live, never cancelled
  destroy();
});

test("a cancelled request is recorded the same way", () => {
  const store = createCompositeStore({ defaultActiveId: "a" });
  const destroy = init(store);
  const request = getMoveRequest(store);

  store.move("a");
  store.setActiveId("b");

  expect(request.targetId).toBe(undefined); // cancelled
  destroy();
});
```

Both pass. This is not a live defect: the two consumers in `composite.tsx` each exclude a nullish active id before they compare, `if (activeId == null) return` in the item branch and `isSelfActive` in the container branch, so a cancelled request can never match. But that safety is an invariant of the call sites, not of the record, and nothing states it. A third consumer, or a helper that folds the comparison into `store.getState().activeId === getMoveRequest(store).targetId`, reads a cancelled request as valid for any composite whose active id is unset — a composite with no rendered items resolves `activeId` to `undefined` through the store's own first-enabled-item sync.

A marker outside the value domain removes the coupling and needs no invariant. Either a dedicated flag:

```ts
interface MoveRequest {
  consumedBy: object | null;
  targetId: Core.CompositeStoreState["activeId"];
  /** Whether a non-move update superseded the request. */
  superseded: boolean;
}
```

or a module-private sentinel, so the type itself rules out collision:

```ts
const CANCELLED = Symbol("cancelled");
// targetId: Core.CompositeStoreState["activeId"] | typeof CANCELLED
```

The behavior you have shipped is right and the tests cover it; this is about the record carrying the distinction rather than borrowing a legal value for it.

**Assessment**

- Severity: Low (verified, 90% confidence), because no current consumer can observe the collision.
- Priority: P2 (estimate, 80% confidence), because it is newly introduced code and cheapest to make unambiguous now.
- Breakage likelihood: Unlikely (estimate, 75% confidence), and only through a future consumer that compares `targetId` without first excluding a nullish active id.
- Fix effort: Easy (verified, 90% confidence), one field or one sentinel; I measured the flag variant at `pnpm test` 1984 passed and a 550-test Chrome sweep before you landed this commit.

.changeset/300-composite-move-target.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react-components": patch
+"@ariakit/react": patch
+---
+
+Fixed [`Composite`](https://ariakit.com/reference/composite) carrying out a pending [`move`](https://ariakit.com/reference/use-composite-store#move) request for a different target after [`setActiveId`](https://ariakit.com/reference/use-composite-store#setactiveid) changes the active item. This affects all components built on [`Composite`](https://ariakit.com/reference/composite), including [`Toolbar`](https://ariakit.com/reference/toolbar) and [`Menubar`](https://ariakit.com/reference/menubar).
```
diegohaz: This entry now understates what the PR fixes. "for a different target" was accurate when the guard only compared targets, but `3f8739886` also cancels a request whose target is unchanged, which is the case the two new `after the active id cycles` tests cover: `move(null)`, `setActiveId("pending-bold")`, `setActiveId(null)`, then remount. The final target equals the requested one, so a reader matching their bug against this sentence would conclude it is not covered.

The distinguishing condition is no longer the target but whether any non-move update intervened:

```md
Fixed [`Composite`](https://ariakit.com/reference/composite) carrying out a pending [`move`](https://ariakit.com/reference/use-composite-store#move) request after [`setActiveId`](https://ariakit.com/reference/use-composite-store#setactiveid) changes the active item, including when a later update sets the active item back to the one the request asked for. This affects all components built on [`Composite`](https://ariakit.com/reference/composite), including [`Toolbar`](https://ariakit.com/reference/toolbar) and [`Menubar`](https://ariakit.com/reference/menubar).
```

I checked the credit requirement separately: #7378 was opened by the repository owner and the only other participants are review bots, so `ariakit-general-contributor-credit` adds no `Thanks to` sentence here.

**Assessment**

- Severity: Low (verified, 90% confidence), because it affects release notes rather than behavior.
- Priority: P2 (estimate, 85% confidence), though it is worth correcting before release since the entry is the only user-facing description of the fix.
- Breakage likelihood: Likely (verified, 90% confidence) as a documentation mismatch: any reader hitting the cycle variant is told the fix covers a different target than theirs.
- Fix effort: Easy (verified, 95% confidence), one sentence.

diegohaz (commented):
## Re-review

Re-review of `3f8739886`, incremental from my previous marked review at `8505d7201`. One commit since then, `Cancel superseded Composite move targets`, which takes the finding I recommended addressing. Merge base is unchanged at `3d0e69ab2`.

The cycle case is fixed and covered in both suites. Every thread on this PR is resolved and none of the previously validated candidates recurred, so both findings below are new and came from reading this commit.

### Previous finding, verified as fixed

The superseded-request replay is gone. Both branches are covered by mirrored tests, `does not replay a pending composite move after the active id cycles` and `does not replay a pending item move after the active id cycles`, in `test.ts` and `test-browser.ts`, and the new `Focus pending bold action` control carries `tabIndex={0}` as the Safari click-focus rule requires. The implementation differs from the flag I prototyped, cancelling by writing `undefined` to `targetId` instead, which is the subject of the first finding below.

The listener now also fires on `activeId`, so it runs on every keyboard navigation and every hover for every composite-family store rather than only on `moves`. That is the intended cost of detecting a non-move update, and the sweeps below show no behavioral fallout.

### Verification on this head

- `pnpm test` — 1982 passed. `pnpm tsc` — 0 errors. `pnpm lint` — clean.
- `pnpm test-chrome src/sandbox/composite-late-item-focus` — 12 passed; `test-firefox` — 12 passed; `test-safari` — 12 passed. All five original #7378 cases plus the two new cycle cases run in every desktop project.
- `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` — 552 passed, no flakes.

### New findings

Both are P2 and neither blocks merge.

- The cancellation marker is in-band — https://github.com/ariakit/ariakit/pull/7380#discussion_r3902831496. `undefined` is a legal `activeId`, so a cancelled request and a request that legitimately recorded an unset target are now the same record. I verified both states are reachable and produce identical values. It is not a live defect, because each of the two consumers in `composite.tsx` excludes a nullish active id before comparing, but that invariant lives at the call sites and is unstated. A flag or a module-private symbol removes the coupling.
- The changeset understates the fix — https://github.com/ariakit/ariakit/pull/7380#discussion_r3902833381. "for a different target" no longer describes the cycle case, where the final target equals the requested one. Suggested rewording in the comment. I also checked contributor credit: #7378 was opened by the repository owner and the only other participants are review bots, so `ariakit-general-contributor-credit` requires no `Thanks to` sentence.

### Notes, not published inline

- The `state.activeId` read in the count branch replaces the previous `store.getState().activeId`. These are equivalent: the store passes a freshly evaluated `getState()` to each listener, so the snapshot cannot lag a reentrant write. No action needed.
- Ordering inside `move` is what makes the cancellation safe, since the `activeId` write cancels and the `moves` write immediately re-records. The comment on those lines states exactly that, which is the right place for it.
- `consumedBy` is deliberately not reset in the cancellation branch, so a consumed request stays consumed. That reads correctly.

### Not fully validated

The cross-component sweep was run in `chrome` only; `firefox` and `safari` covered the `composite-late-item-focus` sandbox alone. Mobile projects and visual tests were not run, per the local policy. I did not measure the added `activeId` listener against the performance workflow; the reasoning that one keyed listener per store is negligible is an argument, not a measurement.

packages/ariakit-react-components/src/composite/__move-request.ts
diegohaz: Fixed in [`f4ca0acc2`](https://github.com/ariakit/ariakit/commit/f4ca0acc226aae7a1bfb5d0f707dbc7eee598c7c): replaced the legal `undefined` value with a module-private Symbol sentinel, so cancelled requests cannot collide with any `activeId` value.

.changeset/300-composite-move-target.md
diegohaz: Fixed in [`f4ca0acc2`](https://github.com/ariakit/ariakit/commit/f4ca0acc226aae7a1bfb5d0f707dbc7eee598c7c): updated the release note to cover any intervening `setActiveId` change, including when the active item later returns to the request's target.

diegohaz:
**Fixed**

- [`f4ca0acc2`](https://github.com/ariakit/ariakit/commit/f4ca0acc226aae7a1bfb5d0f707dbc7eee598c7c) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902958567)): Used a private Symbol sentinel so cancelled requests cannot collide with legal active item values.
- [`f4ca0acc2`](https://github.com/ariakit/ariakit/commit/f4ca0acc226aae7a1bfb5d0f707dbc7eee598c7c) ([reply](https://github.com/ariakit/ariakit/pull/7380#discussion_r3902962605)): Updated the changeset to cover active item changes that cycle back to the requested target.

diegohaz (commented):
## Re-review

Re-review of `f4ca0acc2`, incremental from my previous marked review at `3f8739886`. One commit since then, `Distinguish cancelled Composite move requests`, which takes both P2 findings from that round. Merge base unchanged at `3d0e69ab2`.

**No new findings.** Every thread on this PR is resolved, no previously validated candidate recurred, and I found nothing substantive in the incremental diff or in the surrounding code it touches. From my side this is ready.

### Previous findings, verified as fixed

- The cancellation marker is now `Symbol("cancelled")`, module-private, with `targetId` widened to `Core.CompositeStoreState["activeId"] | typeof cancelled`. The marker can no longer coincide with a legal active id, so the guarantee moved from an unstated call-site invariant into the type. The two consumers in `composite.tsx` still compare with `!==` and type-check unchanged. Naming matches the dominant camelCase precedent for module-private symbols in this package, alongside `symbol` in `command.tsx`, `composite-hover.tsx` and `disclosure.tsx`, and `accessibleWhenDisabledSymbol` in `focusable.tsx`.
- The changeset now describes the cycle case: "including when a later update sets the active item back to the one the request asked for". It stays a single-paragraph past-tense entry with the affected-components sentence after it, which is the shape `ariakit-general-changeset` asks for, and no credit sentence is required for the reasons recorded last round.

### Verification on this head

- `pnpm test` — 1982 passed. `pnpm tsc` — 0 errors. `pnpm lint` — clean.
- `pnpm test-chrome src/sandbox/composite-late-item-focus` — 12 passed; `test-firefox` — 12 passed; `test-safari` — 12 passed.
- `pnpm test-chrome "src/(sandbox|examples)/(composite|combobox|tab|select|menu|toolbar|menubar|radio|tag)"` — 552 passed, no flakes.
- CI run `33494940362` is on `f4ca0acc226aae7a1bfb5d0f707dbc7eee598c7c` and green across every job, including the desktop and mobile Playwright projects and both visual jobs.

### The performance question I left open is now answered

In the previous two rounds I flagged that broadening the request listener from `["moves"]` to `["moves", "activeId"]` puts it on every keyboard navigation and every hover for every composite-family store, and that my "one keyed listener is negligible" claim was reasoning rather than measurement. The performance workflow on this head reports no significant changes, and the benchmarks that exercise the new listener hardest are flat:

| Benchmark | Scripting | Total |
| --- | --- | --- |
| composite-perf > move across items | 341ms → 340.1ms (-0%) | 1280.7ms → 1277.7ms (-0%) |
| combobox-perf > move across items | 281.4ms → 277.8ms (-1%) | 467.9ms → 452.4ms (-3%) |
| menu-perf > move across items | 240.7ms → 239ms (-1%) | 440.1ms → 438.7ms (-0%) |
| tab-perf > move across tabs | 569ms → 573.1ms (+1%) | 3000.4ms → 3021.8ms (+1%) |

One thing to glance at rather than act on: the two mount rows moved the other way, `mount composite` +3% and `mount controlled composite` +4% scripting. That is directionally consistent with `getMoveRequest` now running during every `useCompositeStoreProps` render and registering a listener at store creation, so it is not implausible. The workflow classified the run as no significant change and I cannot separate a few percent on a mount benchmark from run-to-run variance with one sample, so I am recording it rather than raising it. If it repeats on the next run against a fresh baseline it would be worth a second look.

### Not fully validated

The cross-component sweep I ran locally was `chrome` only; `firefox` and `safari` covered the `composite-late-item-focus` sandbox alone. CI covers the rest, including mobile and visual. I did not re-verify packed package contents, since this commit adds no files and the `__` private-module exclusion was verified two rounds ago.

PR #7412, opened 2026-09-04
Preserve arrow keys in menubar menu controls

### LINKED ISSUE #7409: Arrow keys on a non-MenuItem element inside a Menu move focus to the next Menubar item
### Current behavior

A `Menu` inside a `Menubar` often contains focusable elements that are not `MenuItem`, such as a plain link, a plain button, or a text field.

When one of those elements has DOM focus and the user presses an arrow key, the key never reaches the focused element. Instead the open menu closes, focus jumps to the adjacent `Menubar` item, and that item's menu opens.

Which arrow keys leak depends on the menubar orientation. In the default horizontal `Menubar`, <kbd>ArrowRight</kbd> and <kbd>ArrowLeft</kbd> leak. With `orientation="vertical"`, <kbd>ArrowDown</kbd> and <kbd>ArrowUp</kbd> leak.

The horizontal case is the more damaging one, because those keys have a native meaning inside a text field. With a plain `<input>` inside a horizontal menubar menu, placing the caret at position 0 and pressing <kbd>ArrowRight</kbd> does not move the caret to position 1. It closes the menu and opens the next menubar menu instead.

This also breaks a composition Ariakit itself documents. The [Menu with Combobox](https://ariakit.com/examples/menu-combobox) example places an `Ariakit.Combobox` text input directly inside `Ariakit.Menu`. Composing that same supported pattern inside a `Menubar` breaks it: type into the search field, press <kbd>ArrowLeft</kbd>, and the menu closes and the sibling menubar menu opens instead of the caret moving.

`MenuItem` elements that are not text fields are not affected. Arrow keys pressed on a real `MenuItem` behave correctly, including the intended menubar traversal, so <kbd>ArrowRight</kbd> on a menu item in a horizontal menubar still correctly moves to the next menubar menu.

A `MenuItem` that renders a text field *is* affected, though. With `<Ariakit.MenuItem render={<input defaultValue="abc" />} />` and the caret not at the edge of the text, <kbd>ArrowRight</kbd> still closes the menu and moves to the next menubar item.

A `Menu` outside a `Menubar` is not affected. The same plain link inside a standalone `MenuProvider` keeps focus when <kbd>ArrowDown</kbd> or <kbd>ArrowUp</kbd> is pressed. The leak is specific to the combination of a `Menubar` and a focusable element inside the menu that is not a registered menu item.

2. Replace the contents of `src/app.tsx` with the code below.
3. Click the `File` item to open its menu.
4. Click the `Search` field inside the open menu, then press <kbd>Home</kbd> to put the caret at position 0.
5. Press <kbd>ArrowRight</kbd>.

The caret stays at position 0, the `File` menu closes, focus moves to the `Edit` menubar item, and the `Edit` menu opens.

```tsx

const menu = {
  border: "1px solid",
  background: "white",
  color: "black",
  padding: 4,
};

  return (
    <Ariakit.Menubar style={{ display: "flex", gap: 8 }}>
      <Ariakit.MenuProvider>
        <Ariakit.MenuItem render={<Ariakit.MenuButton />}>File</Ariakit.MenuItem>
        <Ariakit.Menu gutter={4} style={menu}>
          <Ariakit.MenuItem>New Tab</Ariakit.MenuItem>
          <input aria-label="Search" defaultValue="abc" />
        </Ariakit.Menu>
      </Ariakit.MenuProvider>
      <Ariakit.MenuProvider>
        <Ariakit.MenuItem render={<Ariakit.MenuButton />}>Edit</Ariakit.MenuItem>
        <Ariakit.Menu gutter={4} style={menu}>
          <Ariakit.MenuItem>Undo</Ariakit.MenuItem>
        </Ariakit.Menu>
      </Ariakit.MenuProvider>
    </Ariakit.Menubar>
  );
}
```

The same defect appears on the vertical axis. Add `orientation="vertical"` to the `Menubar`, replace the `input` with `<a href="#a">Plain link</a>`, open the `File` menu, press <kbd>Tab</kbd> to reach the link, then press <kbd>ArrowDown</kbd>. The menu closes, focus moves to `Edit`, and the `Edit` menu opens. Clicking the link directly instead of tabbing to it produces the same result, so the leak does not depend on how focus got there.

Verified in Chromium against `@ariakit/react@0.4.39`, the current `latest`, and separately against `main` at [`425f7014a`](https://github.com/ariakit/ariakit/commit/425f7014a).

### Expected behavior

An arrow key pressed on a focusable element inside the menu that is not a registered menu item should act on that element, exactly as it does when the same element sits inside a `Menu` with no `Menubar` ancestor.

In the reproduction, <kbd>ArrowRight</kbd> in the search field should move the caret. In the vertical variant, <kbd>ArrowDown</kbd> and <kbd>ArrowUp</kbd> on the plain link should leave focus on the link with the `File` menu still open.

Menubar traversal from a real `MenuItem` should keep working unchanged, so <kbd>ArrowRight</kbd> on a menu item in a horizontal menubar should still move to the next menubar menu.

### Workaround

Wrap the non-`MenuItem` content in an element that stops arrow keys before they reach the `Menu`.

```tsx
import type * as React from "react";

function stopArrowKeys(event: React.KeyboardEvent) {
  if (event.key.startsWith("Arrow")) {
    event.stopPropagation();
  }
}

<Ariakit.Menu gutter={4}>
  <Ariakit.MenuItem>New Tab</Ariakit.MenuItem>
  <div onKeyDown={stopArrowKeys}>
    <input aria-label="Search" defaultValue="abc" />
  </div>
</Ariakit.Menu>
```

Verified in a browser: the caret moves normally in the horizontal case, the plain link keeps focus in the vertical case, and <kbd>Escape</kbd> still closes the menu and restores focus to the menubar item.

Stop only arrow keys, so that typeahead, <kbd>Escape</kbd> and <kbd>Tab</kbd> keep working, and do not call `preventDefault`, because that would suppress the native caret movement you are trying to restore.

Putting `onKeyDown` on the `Menu` itself does not work. `useMenuList` calls the supplied `onKeyDown` first and then checks only `event.defaultPrevented`, never `event.isPropagationStopped()`, so `stopPropagation` there is ignored.

### Possible solutions

The menubar traversal block in `packages/ariakit-react-components/src/menu/menu-list.tsx` (the `if (parentMenubar)` key map) runs for any keydown that bubbles up to the menu element, with no check on where the event came from.

Every other composite keydown handler starts with a target check: `composite-item.tsx` and `composite.tsx` both open with `if (!isSelfTarget(event)) return;`. `useMenuList` has none.

Registered menu items are protected only by accident. `menu-item.tsx` sets `preventScrollOnKeyDown = true`, so `composite-item.tsx` calls `event.preventDefault()` on every arrow key it handles, and `menu-list.tsx` then bails on `event.defaultPrevented`. Anything focusable inside the menu that is not a registered composite item gets no such protection.

A bare `isSelfTarget` guard would be wrong, and this was checked in a browser. In a horizontal menubar, <kbd>ArrowRight</kbd> pressed on a focused `MenuItem` is supposed to move to the next menubar entry, and it reaches this handler with `isSelfTarget(event) === false` because the target is the item element rather than the menu. Adding that guard alone would break the intended traversal, and would break the existing assertions in `app/src/sandbox/menubar-interactions/test.ts`.

A guard that accepts the menu itself plus registered items of this menu store, using the existing `isItem` helper in `packages/ariakit-react-components/src/composite/utils.ts`, is closer. It is still incomplete on its own: a text field rendered as a `MenuItem` *is* a registered item, so that guard would still let the caret case through. The handler additionally needs to respect the caret-edge condition that `composite-item.tsx` already implements for text fields, or the traversal should move onto the items themselves so it inherits that logic naturally.

The same unguarded handler is present in the published builds of `@ariakit/react-core@0.2.17` and `0.3.6`, so this predates the `Menubar` rename and is reachable through the older public `MenuBar` component as well.

_Found while reviewing the NavigationMenu design in #7365, and filed separately because it reproduces with `Menu` and `Menubar` alone._

diegohaz:
## Motivation

Arrow keys in controls inside a `Menu` can close it and move focus to another `Menubar` item. A plain input, a Combobox input, or a text field rendered as `MenuItem` loses focus instead of moving its caret. Vertical menubars also capture Up and Down from plain links and buttons.

Fixes #7409.

## Solution

`MenuList` now accepts its own keyboard events and events from its registered items. Events from other controls retain their native behavior. For a registered text field, menu traversal checks the selection boundary, as `CompositeItem` already does. Ordinary menu items and the menu container still traverse the menubar. Registered inputs retain traversal at the text edges.

```tsx
<Ariakit.MenuItem
  hideOnClick={false}
  render={<input defaultValue="abc" />}
/>
```

With the caret inside the text, Left and Right move it. At the relevant text edge, the same key can open the adjacent menubar menu. The guard also protects controls inside menu items and submenus. The existing `menubar-interactions/index.react.tsx` now renders a separate `MenuControls` component alongside the original scenario. Its tests stay in the existing sandbox test files. The final sandbox contains no workaround.

## Validation

The original 10 cases failed in React 19, React 18, Chrome, Firefox, and Safari before the workaround. The same tests passed with the wrapper workaround and with the library fix. The added nested-input and submenu cases also fail without the fix. Navigation tests reject a menu-only event guard. Registered-input tests in a vertical menubar fail when the Up and Down selection checks are removed.

The final fix passes 160 integration tests across the menu fixtures, 46 focused checks under React 18, and 75 browser checks across Chrome, Firefox, and Safari. The package build, TypeScript, lint, and CSS checks pass.

## Preview

Before the fix, ArrowRight moves focus from Search to Edit and opens Edit:

https://github.com/user-attachments/assets/5c46a8c8-5c26-4cfc-82d3-5d79025f4699

With the library fix in the combined sandbox, Search keeps focus and Tools stays open. At the text edge in Rename, ArrowRight opens Format:

https://github.com/user-attachments/assets/a80a6a50-0b1a-4762-9115-a3ffa66ca87e

The registered input keeps focus after Up and Down in a vertical menubar:

https://github.com/user-attachments/assets/d4c82ea3-3e82-4363-bf51-8cd9057e9b2d

A temporary standalone-menu fixture verifies the release-note scope: ArrowLeft in a plain input moves the caret and keeps its submenu open. The fixture is not part of the committed sandbox:

https://github.com/user-attachments/assets/8d643a61-7f21-4071-8516-c5961df95d76

## Workaround

Wrap the affected controls in an element inside `Menu` that stops arrow-key propagation. Do not call `preventDefault`, so text editing remains available. Leave ordinary menu items outside the wrapper.

```tsx
function stopArrowKeys(event: React.KeyboardEvent) {
  // TODO: Remove after the fix for https://github.com/ariakit/ariakit/issues/7409 is released.
  if (event.key.startsWith("Arrow")) {
    event.stopPropagation();
  }
}

<Ariakit.Menu>
  <Ariakit.MenuItem>New Tab</Ariakit.MenuItem>
  <div onKeyDown={stopArrowKeys}>
    <input aria-label="Search" defaultValue="abc" />
    <Ariakit.MenuItem
      hideOnClick={false}
      render={<input defaultValue="abc" />}
    />
  </div>
</Ariakit.Menu>
```

The same wrapper works around the Combobox input, or around plain links and buttons in a vertical menubar. Put the handler on the wrapper, not on `Menu` itself. Escape still closes the popup and restores focus. For a registered input item, this temporary wrapper also stops menubar traversal at the text edges; the library fix retains that traversal. Navigation among registered menu items still runs before the wrapper handler.


app/src/sandbox/menubar-interactions/test-browser.ts
```diff
@@ -24,4 +24,106 @@ withFramework(import.meta.dirname, async ({ test }) => {
...
+        await expect(q.menu("Tools")).not.toBeVisible();
+        await test
+          .expect(next === "Format" ? q.menu(next) : q.dialog(next))
+          .toBeVisible();
+      });
+    }
+  }
+
+  for (const key of ["ArrowDown", "ArrowUp"]) {
+    for (const name of ["Help", "Action"]) {
```
diegohaz, line 130: The `ArrowUp` and `ArrowDown` half of the new textbox guard in `menu-list.tsx` (lines 113 to 116) has no regression coverage. This vertical-menubar group presses Up and Down only on the plain link and button, which already exit at the `isItem` check. No test presses Up or Down on the registered text field item (`Rename`) with the caret inside the text while the menubar is vertical.

Evidence: I removed `|| event.key === "ArrowUp"` and `|| event.key === "ArrowDown"` from the guard and ran `pnpm test app/src/sandbox/menubar-interactions/test.ts`. All 26 tests in this sandbox still passed. A temporary test that checks `Vertical menubar`, opens `Tools`, clicks `Rename`, sets the caret to position 1, and presses `ArrowDown` failed under that mutation (focus moved to `Format`), failed against the merge-base library file, and passed at the PR head. `ArrowUp` behaved the same way (focus moved to `Insert`).

Real-world impact: a consumer with a vertical `Menubar` whose menu renders a `MenuItem` as a text field is protected by the current code, but a later edit to this guard could reintroduce the vertical focus leak with a green suite.

Suggested addition next to this group, with the happy-dom duplicate beside the matching loop in `test.ts` (line 212):

```ts
for (const key of ["ArrowDown", "ArrowUp"]) {
  // https://github.com/ariakit/ariakit/issues/7409
  test(`${key} keeps the caret in Rename in a vertical menubar`, async ({
    page,
    q,
  }) => {
    await q.checkbox("Vertical menubar").check();
    await q.menuitem("Tools").click();
    const input = q.menuitem("Rename");
    await input.click();
    await input.evaluate((element: HTMLInputElement) => {
      element.setSelectionRange(1, 1);
    });
    await expect(input).toBeFocused();
    await page.keyboard.press(key);
    await expect(input).toBeFocused();
    await expect(q.menu("Tools")).toBeVisible();
  });
}
```

**Assessment**

- Severity: Medium (estimate, 80% confidence), because a silent regression would reintroduce the vertical-menubar focus leak for text field items.
- Priority: P1 (estimate, 75% confidence), because this PR introduces the branch and the existing vertical-menubar group already has the fixture it needs.
- Breakage likelihood: Rare (estimate, 80% confidence), because the current code is correct and the exposure is limited to future edits of this guard.
- Fix effort: Easy (verified), because the temporary happy-dom version of this test passed at the PR head and failed under the mutation.

.changeset/300-menubar-menu-control-arrows.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ariakit/react": patch
+"@ariakit/react-components": patch
+---
+
+Fixed arrow keys in controls inside a [`Menu`](https://ariakit.com/reference/menu) moving focus to another [`Menubar`](https://ariakit.com/reference/menubar) item instead of preserving focus and text editing.
```
diegohaz: The entry describes only the `Menubar` case, but the same guard also changes nested menus without a `Menubar`. Before this PR, the `hasParentMenu` branch of `useMenuList` ran its `hideMap` for any bubbled keydown, so `ArrowLeft` inside a text field in a submenu placed to the right closed that submenu instead of moving the caret. After this PR the submenu stays open and the caret moves.

Evidence: I temporarily added a standalone `MenuProvider` (no menubar) with a nested `Menu` that contains a plain input to this sandbox, clicked into the input at position 1, and pressed `ArrowLeft`. The check failed against the merge-base library file (the submenu closed) and passed at the PR head.

Real-world impact: the `menu-nested-combobox` example (https://ariakit.com/examples/menu-nested-combobox) composes a `Combobox` inside submenus placed to the right, so consumers who follow that pattern receive this fix and should find it in the release notes.

Suggested wording:

```md
Fixed arrow keys in controls inside a [`Menu`](https://ariakit.com/reference/menu) moving focus to another [`Menubar`](https://ariakit.com/reference/menubar) item, or closing the submenu that contains them, instead of preserving focus and text editing.
```

**Assessment**

- Severity: Low (verified), because only the release note is affected and the runtime behavior is correct.
- Priority: P2 (estimate, 80% confidence), because the note ships with this PR but the omission does not block the fix.
- Breakage likelihood: Likely (verified), because this entry is the only release note for the change.
- Fix effort: Easy (verified), because one sentence changes.

packages/ariakit-react-components/src/menu/menu-list.tsx
```diff
@@ -97,6 +104,17 @@ export const useMenuList = createHook<TagName, MenuListOptions>(
...
+      const target = event.target as HTMLElement;
+      if (!isSelfTarget(event) && !isItem(store, target)) return;
+      // Menu traversal must respect text editing just like CompositeItem,
+      // including arrows on the axis that the menu itself does not handle.
+      if (isTextbox(target)) {
+        const selection = getTextboxSelection(target);
+        const isPrevious = event.key === "ArrowLeft" || event.key === "ArrowUp";
+        const isNext = event.key === "ArrowRight" || event.key === "ArrowDown";
+        if (isPrevious && selection.start !== 0) return;
+        if (isNext && selection.end !== getTextboxValue(target).length) return;
```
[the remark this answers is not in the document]
diegohaz: Confirmed at the PR head [`5e2fac0`](https://github.com/ariakit/ariakit/commit/5e2fac01d9c3df4aa14ed764d5a535601466ceb6): with the whole value of `Rename` selected (`setSelectionRange(0, 3)`), `ArrowLeft` moves focus to `Insert` and opens its menu instead of collapsing the selection. I verified this with a temporary happy-dom test in the sandbox.

Disposition: declined for this PR. The rule is the one `CompositeItem` already applies on the menu's own axis (`composite-item.tsx`, lines 508 to 517). It moves when `selection.start` is 0 for Left and Up, or when `selection.end` equals the value length for Right and Down, without checking whether the selection is collapsed. The same input item therefore already leaves the menu on `ArrowUp` with the whole value selected. The PR does not introduce or worsen the behavior: at the merge base, `ArrowLeft` traversed the menubar from this field regardless of the caret or selection, which the same temporary test confirmed.

Returning early for every non-collapsed selection only here would make the two axes of one text field behave differently. A selection-aware boundary is a separate design decision that should change `CompositeItem` and `MenuList` together.

diegohaz (commented):
## Review summary

Initial review of head [`5e2fac0`](https://github.com/ariakit/ariakit/commit/5e2fac01d9c3df4aa14ed764d5a535601466ceb6) against merge base [`4075ee6`](https://github.com/ariakit/ariakit/commit/4075ee6dc0b52df69586b6dd0c088ef0ed34fc72). Scope: all six changed files, plus the surrounding `useMenuList`, `useCompositeItem`, `useComposite`, `useMenuStore`, menu context, and `@ariakit/utils` textbox helpers.

### What I checked

- The guard accepts the menu element and registered items of the same store, so unregistered controls keep native behavior. Submenu stores inherit the menubar store through `useMenubarContext` in `useMenuStore`, and `MenuScopedContextProvider` does not reset that context, so each submenu handles its own menubar traversal. The existing `show/hide on key down` test already proves `ArrowRight` from a submenu leaf (`Search the Web`) reaches `View`, so the new `isItem` check does not regress that path.
- Alternatives: a bare `isSelfTarget` guard breaks traversal from items (the `from item` tests), and moving traversal into `MenuItem` loses the case where the menu element itself has focus after a click (the `from menu` tests). The chosen guard is the smallest change that keeps both.
- The sandbox keeps the original scenario unchanged and adds a separately rendered `MenuControls` component, which matches the sandbox layout rule. The link and button carry `tabIndex={0}` for Safari click focus.
- Changeset frontmatter, filename prefix, tense, and reference links (`/reference/menu` and `/reference/menubar` return 200) are correct. No contributor credit is due: the issue author is excluded by policy and no other person contributed.
- CI on the head commit is green, including `App / Test Chrome`, `App / Test Firefox`, both `App / Test Safari` shards, and the performance matrix.

### Commands and experiments

- `pnpm install --frozen-lockfile`, then `pnpm test app/src/sandbox/menubar-interactions/test.ts`: 26 passed at the head.
- Regression proof: with `menu-list.tsx` swapped for the merge-base version, all 13 new tests failed, and the 6 `traverses the menubar` tests and 7 original tests still passed.
- Mutation: removing `ArrowUp` and `ArrowDown` from the textbox guard left all 26 tests green. Temporary vertical-menubar tests on `Rename` failed under the mutation and at the merge base, and passed at the head. See the inline comment on `test-browser.ts`.
- Temporary standalone nested menu without a menubar: `ArrowLeft` inside its input closed the submenu at the merge base and kept it open at the head. See the inline comment on the changeset.
- Temporary full-selection check on `Rename`: `ArrowLeft` traversed to `Insert` at both the merge base and the head.
- All temporary edits were reverted and `git status` is clean.

### Existing feedback

- Codex thread [discussion_r3938100596](https://github.com/ariakit/ariakit/pull/7412#discussion_r3938100596) (`menu-list.tsx`, lines 115 to 116): confirmed at the head. Disposition: declined for this PR. The rule matches `CompositeItem` (`composite-item.tsx`, lines 508 to 517), the same field already leaves the menu on `ArrowUp` with a full selection, and the merge base traversed regardless of selection, so the PR neither introduces nor worsens it. A selection-aware boundary would need to change `CompositeItem` and `MenuList` together. Replied on the thread.
- The CodeRabbit merge-risk note in its top-level comment repeats the same claim and has no thread anchor. Same validation and disposition as above.

### Findings published inline

- Test coverage: the Up and Down half of the textbox guard has no regression test (`test-browser.ts`, line 111).
- Release notes: the changeset omits the nested-submenu behavior change that also applies without a `Menubar` (`.changeset/300-menubar-menu-control-arrows.md`, line 6).

No correctness defect was found in the library change.

### Not published

- Nit: `test-browser.ts` mixes the imported `expect` with `test.expect` (line 103). Behavior-neutral.
- `page.getByLabel` in the browser tests is acceptable: the Playwright `q` helper has no label query, and the four targets span the `textbox`, `menuitem`, and `combobox` roles.

### Not fully validated

- I did not run Playwright locally. I relied on the green CI browser jobs for the head commit.
- I did not exercise `contenteditable` targets or RTL text fields. Both follow the pre-existing `CompositeItem` rules.

app/src/sandbox/menubar-interactions/test-browser.ts
diegohaz: Resolved in [`5e2fac01d`](https://github.com/ariakit/ariakit/commit/5e2fac01d9c3df4aa14ed764d5a535601466ceb6): the registered-textbox guard already preserves focus for Up and Down in a vertical menubar. Added targeted coverage in [`008172b82`](https://github.com/ariakit/ariakit/commit/008172b82a21b751db7bcf07eecc41902f6ea606).

The two cases pass in React 19, React 18, Chrome, Firefox, and Safari. Removing only the Up/Down alternatives from the guard makes both integration cases and all six browser cases fail at the post-key focus assertion.

The recording shows both keys retaining focus in Rename:

https://github.com/user-attachments/assets/d4c82ea3-3e82-4363-bf51-8cd9057e9b2d

.changeset/300-menubar-menu-control-arrows.md
diegohaz: Fixed in [`c600be373`](https://github.com/ariakit/ariakit/commit/c600be373a761b88b0b92a4d3cc20790b4e43b49): the release note now includes arrow keys closing a submenu.

I verified the standalone plain-input case: without the guard, ArrowLeft closes the submenu; with it, focus stays in the field and the caret moves left. The autoSelect Combobox example named above follows a separate active-item event-proxy path, so it does not establish this release-note claim.

The recording uses a temporary standalone fixture with the current library code:

https://github.com/user-attachments/assets/8d643a61-7f21-4071-8516-c5961df95d76

diegohaz:
**Already resolved**

- [`008172b82`](https://github.com/ariakit/ariakit/commit/008172b82a21b751db7bcf07eecc41902f6ea606) ([reply](https://github.com/ariakit/ariakit/pull/7412#discussion_r3938238796)): Added Up/Down coverage for registered inputs in vertical menubars. Both integration cases and all six browser cases fail when the vertical guard branches are removed.

**Fixed**

- [`c600be373`](https://github.com/ariakit/ariakit/commit/c600be373a761b88b0b92a4d3cc20790b4e43b49) ([reply](https://github.com/ariakit/ariakit/pull/7412#discussion_r3938238942)): Expanded the release note to include arrow keys closing a standalone submenu, verified with a plain input.

PR #7448, opened 2026-09-07
Add comment reflow linting

diegohaz:
## Motivation

Keep comment wrapping and placement consistent with Ariakit's code style through the existing lint workflow.

## Solution

Add `oxlint-plugin-comment-reflow@0.1.2` and enable its rule with `{ printWidth: 80, trailingComments: "always" }`. Disable Oxfmt's JSDoc formatting so the plugin owns comment reflow. Apply the initial cleanup across the files covered by Oxlint.

Use v0.1.2 to preserve standalone permalink lines and reflow prose that contains semicolons or HTML element names. Keep formulas in inline code so wrapping does not split their expressions. The release also includes the JSDoc tag separator fix; regression coverage checks the generated deprecation guidance of legacy APIs.

The cleanup preserves source syntax and comment wording, with inline-code delimiters added around two formulas. Eligible trailing comments move above their targets. Directives, code examples, and protected comment layouts remain under the plugin's preservation rules.

## Verification

- `pnpm lint-fix` followed by `pnpm lint`.
- `pnpm tsc`.
- `pnpm test run app/src/lib/jsdoc-loader.test.ts website/build-pages/reference-utils.test.ts`.
- `pnpm build-website-lite`.
- Compared source syntax and parsed JSDoc tag names against the original source, and verified standalone permalink preservation, with the autofocus reference deliberately kept inline after `See`.
- Reproduced merged permalinks and orphan lines with v0.1.1, then verified the same cases and repeated reflow with v0.1.2.
- Ran the upstream core regression suite against the published v0.1.2 build.
- Rebuilt Tailwind CSS and confirmed no generated CSS diff.
- Confirmed the deprecation tests fail with malformed tags and pass with valid metadata.
- Ran v0.1.2 on the original multiline deprecation comments and verified that it produces the current source without manual repairs.
- Inspected generated modern and legacy `ComboboxItemCheck` references, including prop descriptions, links, and legacy page anchors.

### DIFF
--- a/app/src/sandbox/collection-renderer-6337/test-browser.ts
+++ b/app/src/sandbox/collection-renderer-6337/test-browser.ts
@@ -6,8 +6,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    // Expected values follow the fixed-size formula: paddingStart + itemSize *
-    // index + gap * index.
+    // Expected values follow the fixed-size formula:
+    // `paddingStart + itemSize * index + gap * index`.
     await test.expect(q.button("Measured 1")).toHaveCSS("top", "20px");
     await test.expect(q.button("Measured 2")).toHaveCSS("top", "62px");
     await test.expect(q.button("Fixed 1")).toHaveCSS("top", "20px");
--- a/app/src/sandbox/collection-renderer-6337/test-browser.ts
+++ b/app/src/sandbox/collection-renderer-6337/test-browser.ts
@@ -6,8 +6,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    // Expected values follow the fixed-size formula:
-    // paddingStart + itemSize * index + gap * index.
+    // Expected values follow the fixed-size formula: paddingStart + itemSize *
+    // index + gap * index.
     await test.expect(q.button("Measured 1")).toHaveCSS("top", "20px");
     await test.expect(q.button("Measured 2")).toHaveCSS("top", "62px");
     await test.expect(q.button("Fixed 1")).toHaveCSS("top", "20px");
--- a/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
+++ b/app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
@@ -49,11 +49,11 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     });
   }
 
-  // https://github.com/ariakit/ariakit/issues/7093
-  // Only virtual focus asks for the active item when the select itself is
-  // focused, which happens while the list is still closed. That request is the
-  // one that has to survive until the list opens, so a pointer open with no
-  // move before it still presents the item.
+  // https://github.com/ariakit/ariakit/issues/7093 Only virtual focus asks for
+  // the active item when the select itself is focused, which happens while the
+  // list is still closed. That request is the one that has to survive until the
+  // list opens, so a pointer open with no move before it still presents the
+  // item.
   test("opening the collapsed select with a pointer presents its item", async ({
     q,
   }) => {
@@ -66,9 +66,9 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await test.expect(focusedOptions).toHaveText("Apple");
   });
 
-  // https://github.com/ariakit/ariakit/issues/7093
-  // Withholding focus must not also withhold the rest of the presentation: a
-  // list that is already on screen still belongs at the active item.
+  // https://github.com/ariakit/ariakit/issues/7093 Withholding focus must not
+  // also withhold the rest of the presentation: a list that is already on
+  // screen still belongs at the active item.
   test("typeahead scrolls a collapsed list to the active item", async ({
     q,
   }) => {
@@ -88,10 +88,10 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await test.expect(select).toBeFocused();
   });
 
-  // https://github.com/ariakit/ariakit/issues/7093
-  // Focusing the collapsed select leaves a presentation waiting for the list to
-  // open. Picking another option before that happens has to retire it, or
-  // opening the list would present the option that was active on focus.
+  // https://github.com/ariakit/ariakit/issues/7093 Focusing the collapsed
+  // select leaves a presentation waiting for the list to open. Picking another
+  // option before that happens has to retire it, or opening the list would
+  // present the option that was active on focus.
   test("picking an option before opening retires the pending presentation", async ({
     q,
   }) => {
@@ -107,10 +107,9 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
   });
 
   // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859
-  // Withholding belongs to the widget that owns focus. A move made from
-  // outside keeps presenting immediately, so the focus it moves is
-  // attributable to that call instead of landing on whoever is focused when
-  // the list next opens.
+  // Withholding belongs to the widget that owns focus. A move made from outside
+  // keeps presenting immediately, so the focus it moves is attributable to that
+  // call instead of landing on whoever is focused when the list next opens.
   test("a move from outside the widget still presents immediately", async ({
     page,
     q,
@@ -136,10 +135,10 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await test.expect(openList).toBeFocused();
   });
 
-  // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859
-  // Once focus is on an option, moving between options is navigation inside
-  // the list, not focus leaving the control, so the focus ring has to keep up
-  // with the active item even though the list is collapsed.
+  // https://github.com/ariakit/ariakit/pull/7098#discussion_r3742291859 Once
+  // focus is on an option, moving between options is navigation inside the
+  // list, not focus leaving the control, so the focus ring has to keep up with
+  // the active item even though the list is collapsed.
   test("arrow keys keep moving focus between options in a collapsed list", async ({
     page,
     q,
@@ -175,17 +174,17 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
       await grape.hover();
 
       await test.expect(select).toHaveAttribute("aria-expanded", "false");
-      // Hover activation is committed inside the mousemove handler, so there
-      // is no positive state to wait for. Cross the frames a presentation
-      // would use to reach the item before asserting that none did.
+      // Hover activation is committed inside the mousemove handler, so there is
+      // no positive state to wait for. Cross the frames a presentation would
+      // use to reach the item before asserting that none did.
       await flushFrames(page);
       await test.expect(grape).not.toHaveAttribute("data-active-item");
       await test.expect(other).toBeFocused();
     });
 
-    // https://github.com/ariakit/ariakit/issues/7118
-    // The gate is about the closed list, not about the authored value, so the
-    // same item still takes hover once the select opens.
+    // https://github.com/ariakit/ariakit/issues/7118 The gate is about the
+    // closed list, not about the authored value, so the same item still takes
+    // hover once the select opens.
     test(`${label}: hovering an option activates it once the list opens`, async ({
       q,
     }) => {
@@ -202,10 +201,10 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     });
   }
 
-  // https://github.com/ariakit/ariakit/issues/7118
-  // The other hover direction: after the list closes by clicking an option,
-  // moving the pointer off that option must not clear the active item or move
-  // focus while the list is collapsed.
+  // https://github.com/ariakit/ariakit/issues/7118 The other hover direction:
+  // after the list closes by clicking an option, moving the pointer off that
+  // option must not clear the active item or move focus while the list is
+  // collapsed.
   test("hover end on a collapsed list keeps the active item", async ({
     page,
     q,
@@ -223,30 +222,29 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await grape.click();
     await test.expect(select).toHaveAttribute("aria-expanded", "false");
     await test.expect(grape).toHaveAttribute("data-active-item");
-    // Clicking an option in an always-visible list leaves DOM focus on it
-    // even as the list collapses, so the hover end below starts from the
-    // option.
+    // Clicking an option in an always-visible list leaves DOM focus on it even
+    // as the list collapses, so the hover end below starts from the option.
     await test.expect(grape).toBeFocused();
 
     // The click's mouseup resets the mouse-movement tracker, and a hover
-    // teleport fires the boundary events before any mousemove, which would
-    // skip the hover-end handler entirely. Move within the option first the
-    // way a real pointer would, so the tracker is armed when it leaves.
+    // teleport fires the boundary events before any mousemove, which would skip
+    // the hover-end handler entirely. Move within the option first the way a
+    // real pointer would, so the tracker is armed when it leaves.
     await grape.hover({ position: { x: 4, y: 4 } });
     await other.hover();
 
-    // Retaining the attribute and focus is an absence-of-change assertion,
-    // and the mouseleave clearing would land within a frame, so cross the
-    // frames before asserting that nothing changed.
+    // Retaining the attribute and focus is an absence-of-change assertion, and
+    // the mouseleave clearing would land within a frame, so cross the frames
+    // before asserting that nothing changed.
     await flushFrames(page);
     await test.expect(grape).toHaveAttribute("data-active-item");
     await test.expect(grape).toBeFocused();
   });
 
-  // https://github.com/ariakit/ariakit/issues/7118
-  // The closed gate must keep the authored callback from running at all: this
-  // one moves the composite from inside the predicate, so merely invoking it
-  // while collapsed would activate the item and steal focus.
+  // https://github.com/ariakit/ariakit/issues/7118 The closed gate must keep
+  // the authored callback from running at all: this one moves the composite
+  // from inside the predicate, so merely invoking it while collapsed would
+  // activate the item and steal focus.
   test("a side-effectful callback does nothing on a collapsed list", async ({
     page,
     q,
@@ -262,17 +260,16 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     await grape.hover();
 
     await test.expect(select).toHaveAttribute("aria-expanded", "false");
-    // Hover activation is committed inside the mousemove handler, so there
-    // is no positive state to wait for. Cross the frames a presentation
-    // would use to reach the item before asserting that none did.
+    // Hover activation is committed inside the mousemove handler, so there is
+    // no positive state to wait for. Cross the frames a presentation would use
+    // to reach the item before asserting that none did.
     await flushFrames(page);
     await test.expect(grape).not.toHaveAttribute("data-active-item");
     await test.expect(other).toBeFocused();
   });
 
-  // https://github.com/ariakit/ariakit/issues/7118
-  // The gate is about the closed list: the same side-effectful callback keeps
-  // working once the select opens.
+  // https://github.com/ariakit/ariakit/issues/7118 The gate is about the closed
+  // list: the same side-effectful callback keeps working once the select opens.
   test("a side-effectful callback activates the option once the list opens", async ({
     q,
   }) => {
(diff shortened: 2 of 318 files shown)


app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
```diff
@@ -49,11 +49,11 @@ withFramework(import.meta.dirname, async ({ test, query }) => {
     });
   }
 
-  // https://github.com/ariakit/ariakit/issues/7093
-  // Only virtual focus asks for the active item when the select itself is
-  // focused, which happens while the list is still closed. That request is the
-  // one that has to survive until the list opens, so a pointer open with no
-  // move before it still presents the item.
+  // https://github.com/ariakit/ariakit/issues/7093 Only virtual focus asks for
```
diegohaz: The reflow treats a comment line that holds only a permalink as an ordinary word in the surrounding paragraph, so it absorbs the link into the prose whenever it fits. Here the issue link that opened the block now sits mid-sentence:

```ts
// https://github.com/ariakit/ariakit/issues/7093 Only virtual focus asks for
// the active item when the select itself is focused, which happens while the
```

`ariakit-general-code-style` says the opposite: "Put the most specific stable permalink on a separate comment line." So the new rule enforces a layout that the written code style forbids, and it will keep doing so for every comment written from now on.

This is not isolated. Counting comment lines that contain only a URL, base against head across the changed files:

- 72 permalink lines lost their dedicated line, in 40 files.
- 22 permalinks in 13 files gained one, because the arithmetic happened to fall the other way.

This file alone drops from 15 such lines to 4, and `test.ts` next to it from 14 to 4. The placement is now decided purely by whether the URL fits after the last prose word, so the same link style renders differently from one block to the next.

I confirmed the behavior on a scratch file with the PR's exact configuration. A leading link is absorbed, a trailing link is absorbed once it fits, and only a blank `//` separator keeps it apart:

```ts
// Short prose that ends early.
// https://github.com/ariakit/ariakit/issues/7093

// becomes
// Short prose that ends early. https://github.com/ariakit/ariakit/issues/7093
```

Since you own `oxlint-plugin-comment-reflow`, the durable fix is upstream: treat a comment line whose entire content is a link as its own paragraph, the way an empty `//` line already separates one. That keeps the rest of the reflow intact and leaves the 50 net permalink lines where the style guide puts them. If instead the merged form is the layout you now want, the code-style rule should change with it, so the two stop disagreeing.

**Assessment**

- Severity: Medium (estimate, 85% confidence), because it degrades a convention the repository documents and applies it across the whole linted tree rather than one call site.
- Priority: P1 (estimate, 80% confidence), because the cleanup commit is what bakes it in. Reverting after the fact costs another 40-file churn.
- Breakage likelihood: Likely (verified, evidence: 72 absorbed lines across 40 files in this diff, reproduced on a scratch file with this configuration).
- Fix effort: Moderate (estimate, 75% confidence), because it needs a plugin change plus a re-run of the cleanup, or a code-style amendment instead.

packages/ariakit-test/src/__utils.test.ts
```diff
@@ -7,7 +7,8 @@ afterEach(() => {
 
 // Resolving the realm by reading one member is only safe while nothing shadows
 // it. A form exposes its controls as named properties, and browsers let those
-// override built-ins, so a control named `ownerDocument` answers the lookup this
+// override built-ins, so a control named `ownerDocument` answers the lookup
+// this
```
diegohaz: The fixer left a one-word line here. The base line was 81 columns, one over the target, and the fix pushed only the overflowing word down instead of rewrapping the rest of the paragraph:

```ts
// override built-ins, so a control named `ownerDocument` answers the lookup
// this
// helper starts from. Measured on Chromium 151, Firefox 153, and WebKit 26.5;
```

`packages/ariakit-test/src/shims.test.ts:608` has the same shape, with `// and` alone on its line between `<button>` and `<input>`.

I reproduced the cause: joining the paragraph into one source line and rerunning `oxlint --fix` produces a clean wrap with no orphan, so the fixer is doing a per-line overflow repair rather than reflowing the paragraph.

```ts
// Resolving the realm by reading one member is only safe while nothing shadows
// it. A form exposes its controls as named properties, and browsers let those
// override built-ins, so a control named `ownerDocument` answers the lookup
// this helper starts from. Measured on Chromium 151, Firefox 153, and WebKit
// 26.5; happy-dom defines `ownerDocument` on the prototype chain, so it is
// emulated.
```

Two ways forward: rejoin these two paragraphs and let the fixer rewrap them, or make the plugin cascade the rewrap through the remaining lines of the paragraph so the cleanup cannot produce an orphan in the first place. The second also protects future edits, since any comment that lands one column over the target hits this.

**Assessment**

- Severity: Low (verified, evidence: comment layout only, no runtime or generated-documentation effect).
- Priority: P2 (estimate, 85% confidence), because it is two localized comments, but it points at fixer behavior that recurs.
- Breakage likelihood: Possible (verified, evidence: 2 occurrences in this cleanup, both from a paragraph whose last line ran one column long).
- Fix effort: Easy (verified, evidence: rejoining the paragraph and rerunning the fixer produced the clean wrap shown above).

app/src/sandbox/collection-renderer-6337/test-browser.ts
```diff
@@ -6,8 +6,8 @@ withFramework(import.meta.dirname, async ({ test }) => {
     page,
     q,
   }) => {
-    // Expected values follow the fixed-size formula:
-    // paddingStart + itemSize * index + gap * index.
+    // Expected values follow the fixed-size formula: paddingStart + itemSize *
+    // index + gap * index.
```
diegohaz: The formula was on its own line before the cleanup, and the reflow merged it into the prose and then broke it across the wrap, leaving a line that ends in a bare `*`:

```ts
// Expected values follow the fixed-size formula: paddingStart + itemSize *
// index + gap * index.
```

`packages/ariakit-tailwind/src/input.ts:393` has the same problem, ending a line with a bare `+`:

```ts
// When forbidden: blend between boundary and flipped deltas. Using a +
// x*(b-a) instead of a*(1-x) + b*x so shouldFlip appears once.
```

Both comments exist to state math that the code does not show, so splitting the expression at an operator is the one place the wrapping should not fall.

The plugin keeps inline code together, so backticks fix it. I verified both by editing the comment and rerunning `oxlint --fix`:

```suggestion
    // Expected values follow the fixed-size formula:
    // `paddingStart + itemSize * index + gap * index`.
```

For `input.ts`, `` Using `a + x*(b-a)` instead of `a*(1-x) + b*x` `` reflows to a clean two-line wrap with both expressions intact. Backticks also match how the rest of these comments already mark code, such as `` `ownerDocument` `` and `` `layer-contrast` ``.

**Assessment**

- Severity: Low (verified, evidence: comment layout only, no runtime effect).
- Priority: P2 (estimate, 80% confidence), because it is two comments, though both are the kind that only exist to carry the formula.
- Breakage likelihood: Possible (estimate, 75% confidence), because any unbackticked expression whose operator lands near the wrap point splits the same way.
- Fix effort: Easy (verified, evidence: adding backticks and rerunning `oxlint --fix` produced the layouts above).

diegohaz (commented):
## Review scope

Initial review. Baseline is the merge base `3abf1a6` against head `dc0d4a4`, 317 files.

The mechanical part of this cleanup holds up. I verified it independently rather than reading the diff, since 311 of the changed files are source files whose diff is almost entirely whitespace.

## What I verified

**Source and comment text are unchanged.** For every changed `.ts`, `.tsx`, `.js`, `.jsx`, `.mjs`, and `.cjs` file, I extracted all comment ranges with the TypeScript scanner and compared the two versions separately: code with comments removed, and comment text with whitespace collapsed. All 311 files match. The only code changes are the three intentional ones: `oxfmt.config.ts`, `oxlint.config.ts`, and the new test in `website/build-pages/reference-utils.test.ts`.

**Fenced blocks and comment structure survive.** Fenced code inside comments is byte-identical in every file. No markdown block construct, such as a list marker, heading, table row, blockquote, or four-space indent, was created or destroyed by the rewrapping.

**No directive was touched.** Nothing matching `@ts-expect-error`, `@ts-ignore`, `oxlint-disable`, `eslint-disable`, `biome-ignore`, `istanbul`, `v8 ignore`, `prettier-ignore`, `#region`, `<reference`, `@vitest-`, `@license`, or `@ts-nocheck` appears on either side of the diff.

**Generated reference metadata is safe.** I dumped every reference from the legacy extractor for all 28 modules of `packages/ariakit-react/src`, on both the base and head trees, and diffed field by field. 380 entries on both sides, 634 strings changed, all wrapping only. The two entries where markdown line starts differ are the same six list items in `Menu.props[0].description` with one item rewrapped, so the markers and their count are unchanged. Nothing gained a four-space indent, so no prose became a code block.

The `@deprecated` reindentation is real but harmless. Continuation lines in the extracted text now carry two leading spaces, so the `useState` deprecation reads `Use`, then a two-space-indented line holding the `useStoreState` link and `instead.`. Two spaces is still a paragraph continuation in markdown, and both renderers inline the text after a label, in `renderDeprecated` and in `<Markdown content={deprecated} />`. The added `reference-utils.test.ts` case fills the legacy-side gap. The modern side already had the equivalent coverage in `jsdoc-loader.test.ts`, so the two generators are now symmetric.

**The reflow is reproducible.** I checked out the base tree, applied only the config changes, and ran `oxlint --fix` plus `oxfmt`. The result reproduces head exactly, apart from the docs regeneration and the new test. That independently confirms the description's claim that v0.1.1 produces the current source with no manual repairs.

**No measurable lint cost.** `oxlint --disable-nested-config` over the whole repo: 4.97s on base, 5.00s and 5.02s on head. The JS plugin adds under 1%.

## Findings

Three, all about the output the rule produces rather than the mechanism.

1. [Permalink lines absorbed into prose](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951601333), in `combobox-select-collapsed-focus/test-browser.ts`. 72 comment lines that held only a URL lost their dedicated line, across 40 files, while 22 in 13 files gained one. `ariakit-general-code-style` requires the permalink on a separate comment line, so the rule as configured enforces the opposite of the written style. Severity Medium, priority P1.
2. [One-word orphan lines](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951602060), in `__utils.test.ts` and `shims.test.ts`. The fixer pushes an overflowing tail word onto its own line without rewrapping the rest of the paragraph. Severity Low, priority P2.
3. [Formulas split at an operator](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951603374), in `collection-renderer-6337/test-browser.ts` and `ariakit-tailwind/src/input.ts`. Backticks fix both, verified. Severity Low, priority P2.

## Candidates I checked and am not reporting

**Trailing comment moves.** `trailingComments: "always"` moves eight trailing comments above their target, and every one of the source lines already fit in 80 columns, so the width argument does not apply. I ran the whole cleanup again with the plugin's default `"overflow"` to compare: the entire behavioral difference across the repository is those eight placements. I was going to report the setting as unnecessary until I checked `ariakit-general-code-style`, which says "Do not put comments at the end of a code line." `"always"` is the setting that matches the written style, and `"overflow"` would leave eight violations in place. Correct as configured.

**Missing changeset.** The bot flagged it. The change is whitespace inside doc comments, with no behavior or API effect, and the repository has precedent: the biome to oxlint migration in #5576 reformatted the packages with no changeset. No changeset needed.

**Generated files.** No file in the diff carries a generated-file marker, so nothing should revert on the next regeneration. The docs job ran on `dc0d4a4` and the three regenerated readmes are the complete set for the three packages whose sources changed.

**`jsdoc: false` in Oxfmt.** Required, since the two formatters would otherwise fight over the same comments. It does mean Oxfmt no longer normalizes malformed JSDoc gutters and the plugin preserves nonstandard block layouts by design, so that repair is gone. Accepted tradeoff, not a finding.

## Not fully validated

**Editor feedback.** `.vscode/settings.json` already sets `source.fixAll.oxc` with format on save, which matches the plugin's recommended setup, but I did not confirm that the Oxc extension's language server loads JS plugins. Oxlint's JS plugin API is alpha. If it does not, the rule only surfaces at commit time through the hook or in CI, which is workable but worth knowing.

**Lint scope.** `ignorePatterns` still excludes `website` and `**/*.astro`, so comment style is now enforced on part of the tree only. That matches the description, which scopes the cleanup to files Oxlint covers, and it is pre-existing scope rather than something this PR changes.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
diegohaz: Filed upstream: https://github.com/diegohaz/oxlint-plugin-comment-reflow/issues/6

I reproduced it against `main` at [`d3db528`](https://github.com/diegohaz/oxlint-plugin-comment-reflow/commit/d3db5283e36cf61b8324af39fd10100ca820d662), which is the same tree as the published 0.1.1, so the behavior is not already fixed there. The issue records the minimal `reflowText` case, the counts from this diff, and the two layouts that already survive, a blank `//` separator and an angle bracket autolink, since either could be the model for keeping a link-only line in place.

packages/ariakit-test/src/__utils.test.ts
diegohaz: Filed upstream: https://github.com/diegohaz/oxlint-plugin-comment-reflow/issues/7

Reproducing it against `main` at [`d3db528`](https://github.com/diegohaz/oxlint-plugin-comment-reflow/commit/d3db5283e36cf61b8324af39fd10100ca820d662) narrowed the cause, which corrects what I wrote above. The wrapper is not doing a per-line repair. `isStructure` cuts the paragraph early, and the shortened paragraph then wraps correctly, leaving the tail word alone. Two of its alternatives fire on ordinary prose: `;$` on the line ending in `26.5;` here, and `\s*<` on the `<input>` line in `shims.test.ts`.

That also changes the local fix. Rejoining only the paragraph is not enough, since the misclassified line still cuts it. Feeding the whole comment in as one line does produce the clean wrap.

app/src/sandbox/combobox-select-collapsed-focus/test-browser.ts
diegohaz: Fixed in [`2674e8020`](https://github.com/ariakit/ariakit/commit/2674e8020cd905e9f283d7a040ea1ddcc554c059): upgraded to v0.1.2 and regenerated the cleanup from the original comments, which restores the permalink lines that v0.1.1 had already merged. Every originally standalone permalink remains on its own line, and repeated reflow is stable.

Verified the original failing case against both versions and ran the upstream core regression suite against the published v0.1.2 build.

packages/ariakit-test/src/__utils.test.ts
diegohaz: Fixed in [`2674e8020`](https://github.com/ariakit/ariakit/commit/2674e8020cd905e9f283d7a040ea1ddcc554c059): v0.1.2 treats the semicolon and HTML element-name lines as prose, so the complete paragraphs reflow together. The isolated `this` and `and` lines are gone.

Both cases reproduce with v0.1.1 and pass with v0.1.2, including a second reflow pass. The published build also passes the upstream core regression suite.

app/src/sandbox/collection-renderer-6337/test-browser.ts
diegohaz: Fixed in [`6bccdb4bc`](https://github.com/ariakit/ariakit/commit/6bccdb4bc0a0222ebc63ab87b84fa3df6916dea7): added inline-code delimiters to the renderer formula and both blend expressions.

`paddingStart + itemSize * index + gap * index`, `a + x*(b-a)`, and `a*(1-x) + b*x` stay intact after lint and formatting. Source syntax is unchanged, and the Tailwind rebuild produces no CSS diff.

diegohaz:
**Fixed**

- [`2674e8020`](https://github.com/ariakit/ariakit/commit/2674e8020cd905e9f283d7a040ea1ddcc554c059) ([reply](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951903354)): Updated to v0.1.2 and restored standalone permalink lines.
- [`2674e8020`](https://github.com/ariakit/ariakit/commit/2674e8020cd905e9f283d7a040ea1ddcc554c059) ([reply](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951903498)): Fixed paragraph wrapping around semicolons and HTML element names.
- [`6bccdb4bc`](https://github.com/ariakit/ariakit/commit/6bccdb4bc0a0222ebc63ab87b84fa3df6916dea7) ([reply](https://github.com/ariakit/ariakit/pull/7448#discussion_r3951907994)): Kept renderer and blend formulas intact with inline-code markup.

Lint, formatting, TypeScript, documentation tests, and the website build pass. The published plugin passes its upstream core regression suite. Source syntax and JSDoc tags are preserved; the Tailwind rebuild produces no CSS diff.

diegohaz (commented):
## Review scope

Re-review. Incremental baseline is the previous marked review at `dc0d4a4` against head `81c671e`, three commits and 48 files. I also re-ran every invariant from the merge base `3abf1a6`, since the cleanup was regenerated and the whole output is new.

No new findings.

## The three findings are fixed, verified independently

**Permalink lines.** Counting comment lines that hold only a URL, base against head: 1 absorbed, in `focusable.tsx`, which is the deliberate `See https://x.com/...` join in [`81c671e`](https://github.com/ariakit/ariakit/commit/81c671e189fbc530b08a6044601bc8930c4dfde6). The other 71 are back on their own line. `combobox-select-collapsed-focus/test-browser.ts` returns from 4 such lines to 15, and its sibling from 4 to 14. `menu-store.ts` and the `issues/7068` block in `combobox-popover.tsx` now match the merge base exactly.

**Orphan lines.** Both are gone, and both landed on the wrap I predicted, including the permalink staying separate:

```ts
// override built-ins, so a control named `ownerDocument` answers the lookup
// this helper starts from. Measured on Chromium 151, Firefox 153, and WebKit
// 26.5; happy-dom defines `ownerDocument` on the prototype chain, so it is
// emulated.
// https://github.com/ariakit/ariakit/issues/7209
```

Scanning the whole changed tree for a short prose line stranded mid paragraph now finds none. Every hit is a paragraph-final line followed by a preserved permalink, which is the intended shape.

**Formulas.** Both lines that ended in a bare operator are gone, and the backticked expressions survive the wrap.

The fix also reached a file the old version could not touch. `app/src/lib/collection-cache.ts` was left alone under 0.1.1 because its first line ends in a semicolon, and it now wraps correctly.

## What I re-verified at this head

**Source and comment text.** Across all 296 changed source files, code with comments removed is unchanged, and comment text with whitespace collapsed is unchanged in every file except `collection-renderer-6337/test-browser.ts` and `ariakit-tailwind/src/input.ts`, which is the intended backtick edit in [`6bccdb4`](https://github.com/ariakit/ariakit/commit/6bccdb4bc0a0222ebc63ab87b84fa3df6916dea7). Code changes remain the three intentional files.

**Comment structure.** Fenced blocks are byte-identical. No markdown block construct was created or destroyed. No directive line appears on either side of the diff.

**Generated reference metadata.** I dumped all 380 references from the legacy extractor at this head and compared with the base: 634 strings changed, wrapping only, and the same two benign list rewraps in `Menu.props[0].description`. The dump is also byte-identical to the one I took at `dc0d4a4`, so the 0.1.2 regeneration did not move any generated documentation.

**Generated readmes.** `pnpm docs` across every package produces no diff, so the three regenerated readmes are still the complete and current set.

**Reproducible and stable.** Running `oxlint --fix` at this head changes nothing. Rerunning the whole cleanup from the merge base with only the config files from this head reproduces the head exactly, apart from the two formula edits, the `focusable.tsx` join, the readmes, and the new test. No unexplained manual edit is hiding in the diff.

**Local checks.** `pnpm lint`, which includes `oxfmt --check`, passes. CI at `81c671e` has 44 checks passing and 2 Safari jobs still running, with Lint, Type check, Test, Test React 18, Docs, Build packages, Build website, and Test Plus all green.

## Notes, not findings

**Short label lines before a long link.** 9 comments now read as a label alone followed by the permalink, such as `// Regression for` in the two `dialog-persistent-elements` files and `// Based on` in `use-prevent-body-scroll.ts`. The link tokens are 67 and 144 columns, so nothing can share their line at a target of 80, and the base only avoided it by running over. The two-line form matches the code style, so this is the right tradeoff rather than a regression.

**PR description.** Updated for 0.1.2 and accurate against what I checked, including the note that the autofocus reference is inline on purpose.

**Bot coverage.** CodeRabbit still skips at 301 files, one over its limit. Codex completed on `81c671e` with no findings.

**Changeset.** Still not needed, for the reasons in the previous review.

## Existing feedback

All three threads from the previous review are resolved with replies naming the fix commits. I validated each against this head rather than relying on the replies, and all three hold. No unresolved thread remains.

PR #7464, opened 2026-09-11
Make control joining optional and preserve tab corners

diegohaz:
## Motivation

Tabs without strip padding inherit the control group's border-joining and corner rules. This squares the selected tab's inner top corner, even though each tab should keep its own shape. This is a follow-up to #5240.

## Solution

Add a boolean `$joined` variant that controls both the border overlap and the inner-corner rules. It defaults to `true` for control groups and `false` for tab lists. Joining still depends on the existing layout, gap, and padding conditions. The tab list keeps its frame cover behavior.

```tsx
<ButtonGroup $p="none" />
<ButtonGroup $p="none" $joined={false} />
<TabList /> // Independent by default; pass $joined to opt in.
```

Add an independent button-group example to the gallery and browser coverage for the defaults, explicit overrides, zero padding, spacing, and layout conditions. The tab checks cover folder, flat, and bevel tabs, plus a folder glider.

## Preview

The same unpadded folder tabs before and after the change:

| Before | After |
| --- | --- |
| The selected Preview tab has a square top-right corner | The selected Preview tab has both top corners rounded |

## Verification

- `pnpm lint-fix`
- `pnpm tsc`
- Browser regression tests pass in Chrome, Firefox, and Safari.
- Both new regression checks fail when the fix is removed and pass after restoration.
- Recipe output comparisons preserve existing button-group defaults and explicit joined tab lists.

### DIFF
--- a/app/src/pages/ariakit-ui/_ariakit-ui/button.react.tsx
+++ b/app/src/pages/ariakit-ui/_ariakit-ui/button.react.tsx
@@ -384,8 +384,8 @@ export function ButtonSection() {
 
       <Sample
         title="Groups"
-        code='ButtonGroup $layout="horizontal" | "stretch" | "vertical" | "wrap" · $gap · $p="none"'
-        description="A group sizes its buttons together. Without padding the inner corners square off so the buttons join."
+        code='ButtonGroup $layout="horizontal" | "stretch" | "vertical" | "wrap" · $gap · $p="none" · $joined={false}'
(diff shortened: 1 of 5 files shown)


diegohaz:
Closing as superseded by #7471. That PR includes the same `$joined` option and the `$joined: false` default for tab lists, and implements joining through the shared `ak-frame-join` utilities.

The additional coverage here is still useful and is not all present in #7471 yet: unpadded folder, flat, and bevel tabs; the folder glider; explicit `<TabList $joined>`; and padded independent button groups. Those regression cases and the independent-group gallery example should be carried over before #7471 merges.

PR #7490, opened 2026-09-12
Make disabled Ariakit UI controls dim consistently

### LINKED ISSUE #7474: Make disabled Ariakit UI controls dim consistently
## Context

Found while moving the Ariakit UI gallery into the `app/src/sandbox/ariakit-ui/` sandbox in #7465. The PR fixes the disabled bevel button, the disabled select field and choice cards disabled through their group. Two gaps in the shared disabled look remain: a badge slot in a disabled control keeps its full fill, and the `$disabled` prop dims less than a disabled attribute.

Both items were checked on `main` at [`a23ef72`](https://github.com/ariakit/ariakit/commit/a23ef727caa80db0769f0329d8d8ccf7c26568e2) by reading the code, and on the PR branch at [`bdfbb8b`](https://github.com/ariakit/ariakit/commit/bdfbb8b47ac8a243259994dd0c8a06280daca765) with Chromium probes on the sandbox Button page.

## Status on `main`

Both items are **pre-existing on `main`**. The PR does not change the rules involved. The boxes exist only in the PR branch, so work on this issue is easiest after #7465 merges.

### 1. A badge slot keeps its full fill in a disabled control

For badge and avatar slots, [`controlSlot`](https://github.com/ariakit/ariakit/blob/a23ef727caa80db0769f0329d8d8ccf7c26568e2/packages/ariakit-ui/src/styles/control.ts#L255) adds `group-[.disabled]/control:ak-layer-darken-5` and `group-[.disabled]/control:ak-ink-0`. In the "Disabled" box (`<Button disabled>` with `<ButtonSlot $kind="badge">3</ButtonSlot>`), the darken rule matches and sets `--_ak-layer-idle-relative-lightness: calc((5 / 100) * -1)`, but the slot background stays `oklch(0.515341 0.1546 248.516)`. That is the same fill as the enabled "Count badge" box. Only the text fades, to `oklch(1 0 0 / 0.733)`. An earlier verification found the same result for badge slots in disabled combobox rows.

### 2. `$disabled` alone dims less than the disabled attribute

The control's `$disabled` classes set `ak-ink-0`, but `@ariakit/tailwind` lowers the contrast scale only through [`:where(:disabled, [disabled], [aria-disabled="true"])`](https://github.com/ariakit/ariakit/blob/a23ef727caa80db0769f0329d8d8ccf7c26568e2/packages/ariakit-tailwind/src/input.ts#L864). With the "Disabled" button's classes on an element that has no disabled attribute, the label computes `oklch(0 0 0 / 0.542)` with `--_ak-contrast-scale: 1`. The native disabled button computes `oklch(0 0 0 / 0.433)` with `--_ak-contrast-scale: 0.25`. This affects every label-hosted control that takes the disabled look from the prop, such as a disabled choice card, whose `label` is never `:disabled` itself.

## Why the PR can finish without this

The PR does not change the disabled rules of `controlSlot` or the plugin's disabled contrast. Item 2 needs a new public utility in `@ariakit/tailwind`, a published package, which needs its own design and changeset. The sandbox shows today's look, and every disabled button box uses the native attribute.

## Expected outcome

- A badge or avatar slot visibly changes in a disabled control, through a darken that takes effect on a colored slot or through another disabled treatment.
- `@ariakit/tailwind` exposes a public utility, for example `ak-disabled`, that applies the disabled contrast scale, and `control.$disabled` uses it, so the prop path and the attribute path dim the same.

diegohaz:
## Motivation

Fixes #7474.

A badge in a disabled control keeps its full fill, and an avatar image does not fade. Also, a label styled with `$disabled` has stronger text contrast than a native disabled control because the label does not carry a disabled attribute.

## Solution

Add the public `ak-disabled` utility to `@ariakit/tailwind`. The existing disabled-attribute rule and `control.$disabled` use the same contrast settings. Descendant layers and text inherit those settings.

```tsx
<label className="ak-layer ak-disabled ak-ink-0">
  <input type="file" disabled className="sr-only" />
  Upload attachment
</label>
```

Dim badge and avatar surfaces with `ak-layer-mix-20`, matching the existing disabled choice-card treatment. The fill mixes into its parent before `ak-ink-0` computes the text color, so the text keeps its adaptive contrast. Fade image pixels separately into the slot surface.

Slots follow native and ARIA disabled states changed by scripts, disabled fieldsets, and choice cards disabled through their groups. The first legend of a disabled fieldset stays operable. It still inherits reduced contrast from the fieldset, as it does on `main`; the Tailwind README now documents that limitation.

Nested slots receive the same ink as direct children when a script changes `disabled` or `aria-disabled`. `RadioCardSlot` and `CheckboxCardSlot` now wrap badge children like `ButtonSlot`, so `<RadioCardSlot $kind="badge">3</RadioCardSlot>` gets the intended compact text size.

Choice cards and field rows establish disabled contrast on their wrapping label when the input is disabled. This makes `<CheckboxCard aria-disabled>` and `<RadioField aria-disabled>` match the `disabled` prop, and follows script changes to the input. Input wrappers apply the same settings when all their entry controls are disabled.

The button, radio, and checkbox sandboxes cover these states, the upload label, and both avatar types. Browser regressions compare native, ARIA, utility, and prop-driven disabled styles in light, dark, and high-contrast modes. The Tailwind utility has documentation and a changeset; `@ariakit/ui` is private.

The radio and checkbox card and field tests share one disabled-contrast assertion. Each sandbox supplies its own input and comparison pairs, while the helper checks native and ARIA states and exact enabled-color restoration for every pair. The card tests also retain their check that the enabled color differs from the disabled color.

## Preview

| Surface | Before | After |
| --- | --- | --- |
| Badge | Disabled button before the fix: the blue badge keeps its full fill | Disabled button after the fix: the badge has a pale blue fill and adaptive dark text |
| Avatars | Disabled avatars before the fix: the initial and image retain their fills | Disabled avatars after the fix: the initial keeps adaptive contrast and only the image pixels fade |
| Upload label | Upload label before the fix: its text is stronger than a disabled button | Upload label after the fix: its text uses native disabled contrast |
| Fieldset | Disabled fieldset before the fix: the lower badge and avatar remain opaque | Disabled fieldset after the fix: the lower slots use muted fills while the first legend stays enabled |
| Card group | Disabled card group before the fix in dark theme: the badge and avatar retain full fills | Disabled card group after the fix in dark theme: the badge and avatar use muted fills and adaptive text |

The recording shows a script setting `button.disabled` to `true`, then back to `false`. The nested badge and avatar change their fill and ink without rerendering the recipe. The blue outline marks the script target:

https://github.com/user-attachments/assets/6537a648-1327-4262-ba4e-cdf0d8579305

<details>
<summary>Review fixes: nested ink and card badge text</summary>

| Surface | Before the review fix | After |
| --- | --- | --- |
| Nested slots | Nested disabled slots before the fix have muted fills but solid ink | Nested disabled slots after the fix have muted fills and matching disabled ink |
| Radio card badge | Radio card badge before the wrapper fix has an oversized count | Radio card badge after the wrapper fix has compact count text |
| Checkbox card badge | Checkbox card badge before the wrapper fix has an oversized count | Checkbox card badge after the wrapper fix has compact count text |

</details>

<details>
<summary>Input-disabled card contrast</summary>

The Email card uses `disabled`; the Messages card uses `aria-disabled`. Before this fix, the Messages label and badge keep stronger ink. Afterward, both cards use the same disabled contrast.

| Before | After |
| --- | --- |
| ARIA-disabled card before the fix has stronger label and badge ink than the native-disabled card | Both cards after the fix have matching disabled label and badge ink |

The recording shows `input.disabled` changing from `false` to `true` and back. The blue outline marks the card whose input changes:

https://github.com/user-attachments/assets/74512e32-d2e5-43f0-ab3c-4679a0ed0896

</details>

<details>
<summary>Field rows, input wrappers, and descendant layer proof</summary>

The Email field uses `disabled`; the Messages field uses `aria-disabled`. Both fields now use the same label and description contrast. The input-wrapper comparison disables its inner input; the before image omits the new wrapper rule.

| Surface | Before | After |
| --- | --- | --- |
| Field row | ARIA-disabled field before the fix has stronger label and description text than the native-disabled field | Both disabled fields after the fix have matching label and description contrast |
| Input wrapper | Disabled input wrapper without the new contrast rule has a stronger filter icon | Disabled input wrapper after the fix dims its filter icon to the native-disabled contrast |

The recording shows `input.disabled` changing from `false` to `true` and back. The blue outline marks the field whose input changes:

https://github.com/user-attachments/assets/f0529692-80ac-4e3a-b06c-9235e0de9990

The utility badge and native-disabled badge have matching backgrounds under increased contrast. The focused test now checks this descendant surface as well as its ink:

Native-disabled and utility-disabled descendants with matching blue badge backgrounds under increased contrast

</details>

## Verification

The original disabled regressions fail against the original styles and pass with the fix. The nested ink and card badge text tests fail before their fixes and pass afterward. The button and radio checks pass in Chromium, Firefox, and WebKit, and the full Tailwind browser suite passes, including WCAG and computed-color snapshots. Visual screenshot baselines are generated by CI.

The current main merge also includes the table edge, caption, and control-size fixes and dependency updates. Their relevant browser and integration checks pass. The new radio and checkbox card contrast tests fail before the recipe fix and pass afterward in Chromium, Firefox, and WebKit, with both themes and contrast preferences.

The latest main merge preserves the disabled fixes alongside the disclosure composition, badge sizing, and navigation alignment updates. It keeps the button group examples in separate sections and uses main’s captures to resolve the screenshot conflicts. CI regenerates affected row-group baselines for the combined gallery.

The field-row and input-wrapper regressions fail before the recipe fixes and pass afterward. The descendant-background proof fails when only the utility badge background paint is removed; its text-color checks still pass. The compact badge-text checks run only in Chromium.

The shared choice assertion passes in Chromium, Firefox, and WebKit. Removing the choice contrast rule still fails all four card and field tests. Both card tests also fail when an incorrect enabled color is injected after removing the disabled state, or when the disabled color persists after enabling the input.

`pnpm lint-fix`, `pnpm tsc`, and `pnpm build` pass. `pnpm clean` restores source exports after the package build.

<details>
<summary>Cycle 3: Keep the existing disabled selectors</summary>

The slot findings affect the disabled appearance of supported badge and avatar compositions. Their severity is low. Nested slots disabled by attributes are less common than direct children, but they must receive the same fill and ink treatment. Card badge text has the wrong size whenever the composition is used.

Keep the existing layer and ink utilities and complete their disabled selectors. Match the existing badge wrapper in the radio and checkbox components. This adds no runtime state or new color calculation. A shared container-query state channel would add inheritance and reset rules across several recipes without a demonstrated benefit for this fix.

Disabled contrast inheritance, including the enabled first legend of a disabled fieldset, already exists on `main`. It can reduce the requested high-contrast boost, but requires this uncommon composition. Document that scope as the review permits. A public reset or automatic restoration is outside the current contract. No follow-up implementation is accepted in this change.

The independent reassessment and local assessment agree on the selector and component fixes. The local assessment initially proposed a reset utility; checking the original utility contract and unchanged rules on `main` supports the narrower documentation response. Continue the existing direction without resetting the cycle count.

</details>

<details>
<summary>Cycle 6: Share the choice contrast assertions</summary>

The current finding has low severity and affects test maintenance. Radio and checkbox cards and fields share the same disabled-state checks, but their duplicated loops have drifted: the field tests restore every exact enabled color, while the card tests only require a color change. The maintenance cost recurs whenever these assertions change. The extraction has no production effect.

Keep the current disabled selectors, adaptive color utilities, and supported native, ARIA, fieldset, group, nested-slot, field, and input-wrapper behavior. Share one input-state and color-comparison assertion in the existing Ariakit UI test helpers. Keep fixture locators and media loops at each call site, use exact enabled-color restoration for all pairs, and leave the differently shaped input-wrapper test local.

The independent and local assessments agree. This removes duplicated test logic without adding runtime state, a public API, or a new platform requirement. The earlier reset-utility and custom-image exclusions remain. Continue the same implementation direction at cycle 6 without resetting the count.

</details>


app/src/sandbox/ariakit-tailwind-7474/test-chrome.ts
```diff
@@ -0,0 +1,55 @@
...
+              await test.expect(element).toHaveCSS(property, expected);
+            }
+          }
+          for (const name of ["badge", "colored text"]) {
+            const color = await q
+              .text(`Native ${name}`)
+              .evaluate((element) => getComputedStyle(element).color);
+            await test
+              .expect(q.text(`Upload ${name}`))
+              .toHaveCSS("color", color);
```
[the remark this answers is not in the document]
diegohaz: Don't we already have `@visual` regression tests for that?

packages/ariakit-ui/src/styles/control.ts
```diff
@@ -294,7 +294,8 @@ export const controlSlot = cv({
       variants.$kind === "badge" || variants.$kind === "avatar";
     if (!paintsSurface) return;
     addClass([
-      "group-[.disabled]/control:ak-layer-darken-5",
+      // Fade the whole slot so colored fills and avatar images dim together.
+      "group-[.disabled]/control:opacity-50",
```
diegohaz: I'm not sure `opacity-50` is correct. We should try to dim the elements using `@ariakit/tailwind` utilities so it's context-aware and contrast-safe.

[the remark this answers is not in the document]
diegohaz: Fixed in [`dce45d315666f46e5ec2aacfdf7f815a8ba4265b`](https://github.com/ariakit/ariakit/commit/dce45d315666f46e5ec2aacfdf7f815a8ba4265b): badge and avatar slots now follow native and ARIA disabled states, disabled fieldsets, and choice cards disabled through their groups. For example, `button.disabled = true` updates the surface without a recipe render. The first legend stays enabled.

The same commit adds targeted browser coverage. All button and radio checks pass in Chromium, Firefox, and WebKit. The [updated screenshots and recording](https://github.com/ariakit/ariakit/pull/7490) show the fieldset, card-group, and script-driven states.

diegohaz: Fixed in [`dce45d315666f46e5ec2aacfdf7f815a8ba4265b`](https://github.com/ariakit/ariakit/commit/dce45d315666f46e5ec2aacfdf7f815a8ba4265b): the slot surface now uses `ak-layer-mix-20`, matching the existing disabled choice-card treatment. `ak-ink-0` computes text color from the mixed surface, with no whole-slot opacity to reduce that contrast again. Only image pixels fade into their slot's layer.

The tests cover both themes and more-contrast settings, and reject the previous whole-slot opacity treatment. The [updated screenshots and recording](https://github.com/ariakit/ariakit/pull/7490) show the resulting fills and text.

diegohaz:
**Fixed**

- [`dce45d315`](https://github.com/ariakit/ariakit/commit/dce45d315666f46e5ec2aacfdf7f815a8ba4265b) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3995293574)): Native/ARIA, fieldset, and card-group disabled states now dim slots.
- [`dce45d315`](https://github.com/ariakit/ariakit/commit/dce45d315666f46e5ec2aacfdf7f815a8ba4265b) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3995294876)): Layer mixing preserves adaptive text contrast; image fading stays scoped to image pixels.

packages/ariakit-ui/src/styles/control.ts
```diff
@@ -294,8 +294,19 @@ export const controlSlot = cv({
...
     if (!paintsSurface) return;
     addClass([
-      "group-[.disabled]/control:ak-layer-darken-5",
+      // Mix the fill into its parent before ink is calculated, so disabled
+      // text keeps the contrast chosen for the resulting surface.
+      "group-[.disabled]/control:ak-layer-mix-20",
+      // Native and ARIA states can change without rerendering the recipe.
+      "group-ui-disabled/control:ak-layer-mix-20",
+      // A choice card reads disabled state from the input inside its label.
+      "group-ui-disabled-within/choice:ak-layer-mix-20",
```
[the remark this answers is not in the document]
diegohaz: **Confirmed on `fd0ff44`, with a different trigger than the one described.**

An `aria-disabled` badge is not the case that breaks, because `badge` sets no `group/choice` and is never `:disabled` itself. The case that breaks is a badge or avatar slot that is **not a direct child** of a button-based control which becomes disabled through the attribute rather than the `$disabled` prop.

`controlSlot` now mixes the fill for three signals (lines 299, 301, 303) but adds `ak-ink-0` for only one (line 304). The other two rely on the consumer recipes, and those do not cover the same elements:

- `button.ts:61` has `ui-disabled:*:ak-ink-0!`, which reaches **direct children only**.
- `choice.ts:200` has `ui-disabled-within:**:ak-ink-0`, which reaches **every descendant**.

A slot inside `ButtonContent` is a supported composition (the `controlSlot` `$mx` JSDoc and the `choiceCard` `$orientation` JSDoc both describe it), so a button-based control leaves it with a dimmed fill and full-strength ink.

Measured in Chromium on `/react/previews/ariakit-ui-button/` by cloning the "Count badge" button, nesting its badge slot one level deeper and setting `disabled`:

| Slot | `background-color` | `color` |
| --- | --- | --- |
| enabled | `oklch(0.515341 0.1546 248.516)` | `oklch(1 0 0)` |
| `[disabled]`, slot as direct child | `oklch(0.90804 0.0314768 247.264)` | `oklch(0 0 0 / 0.494732)` |
| `[disabled]`, slot nested one level | `oklch(0.90804 0.0314768 247.264)` | `oklch(0 0 0)` |

Three Inbox buttons with a count badge: enabled with a solid blue badge and white count; disabled with the slot nested one level, showing a pale badge and solid black count; disabled with the slot as a direct child, showing a pale badge and a dimmed count

Before this PR the nested dynamic path was consistently undimmed. The new mix rules make it half-dimmed, which is the inconsistency this PR sets out to remove, so the ink rules belong with them.

**Assessment**

- Severity: Low (estimate, 85% confidence), because the badge text stays readable in both color schemes; only the disabled appearance is inconsistent.
- Priority: P1 (estimate, 70% confidence), because the PR's stated outcome is that the attribute path and the prop path dim the same, and this leaves one path where they do not.
- Breakage likelihood: Possible (estimate, 75% confidence), because it needs a badge or avatar slot below the control's direct children plus an attribute-driven disabled state, such as a script, `fieldset[disabled]`, or `aria-disabled`. No current fixture nests a painted slot, so no test covers it.
- Fix effort: Easy (verified, 95% confidence): I applied the two class strings locally and the six new disabled tests still pass in Chromium.

**Disposition: Address now.** The current work introduces the half-dimmed state, and the changed contract depends on the resolution.

```diff
       "group-[.disabled]/control:ak-ink-0",
+      "group-ui-disabled/control:ak-ink-0",
+      "group-ui-disabled-within/choice:ak-ink-0",
```

With that applied, the nested case measures `oklch(0 0 0 / 0.494732)`, identical to the direct-child case. `group-ui-disabled/control:ak-ink-0` alone closes the observed gap; the `/choice` line is redundant with `choice.ts:200` today but keeps the three ink rules in step with the three mix rules above them.

Worth considering separately: the effects and the signals are now a 3 x 3 matrix that must be kept in sync by hand, and this gap is exactly what a missing cell looks like. A single state channel, for example `--control-disabled` set by `control`, `button` and `choiceCard` and read by one `@container style(...)` variant in `ui.css` next to `ui-list-last-row`, would reduce `controlSlot` to three classes and make a missing cell impossible. I did not prototype that, so treat it as a direction rather than a validated alternative.

app/src/sandbox/ariakit-ui-radio/index.react.tsx
```diff
@@ -588,6 +591,41 @@ export default function RadioExamples() {
...
+          {disabled ? "Enable card grid" : "Disable card grid"}
+        </Button>
+        <RadioProvider defaultValue="jane">
+          <RadioCardGrid aria-label="Assignee" disabled={disabled}>
+            <RadioCard value="jane">
+              <RadioCardSlot $kind="avatar" $layer="brand">
+                J
+              </RadioCardSlot>
+              <RadioCardLabel>Jane Doe</RadioCardLabel>
+              <RadioCardSlot $kind="badge">3</RadioCardSlot>
```
diegohaz, line 624: The badge count is a bare text node, so it never gets the badge text size. `controlSlot` sizes a badge through `*:text-[0.8125em]`, which needs an element child, and the `$kind` JSDoc says so: "When you use the `badge` kind, wrap the text in a `<span>` element so it’s styled correctly." `ButtonSlot` and `BadgeSlot` insert that wrapper themselves, but `RadioCardSlot` renders `<ak.Role.span {...radioCardSlot.jsx(variantProps)} {...rest} />` with no wrapper, so the fixture has to supply it.

Measured in Chromium on `/react/previews/ariakit-ui-radio/`: this slot computes `font-size: 16px` and has zero element children, while the equivalent `<ButtonSlot $kind="badge">3</ButtonSlot>` in the button sandbox computes `13px` on its inner span. The count reads visibly larger in its pill:

Disabled card slots box: the card shows a J avatar, the label Jane Doe and a pale badge whose 3 fills most of the pill

Disabled box in the button sandbox: the Delete button shows a pale badge whose 3 is noticeably smaller inside the pill

This is the only `RadioCardSlot $kind="badge"` in the repository, so it becomes the reference for that composition, and CI is generating the `ariakit-ui-radio` visual baselines from it.

```diff
-              <RadioCardSlot $kind="badge">3</RadioCardSlot>
+              <RadioCardSlot $kind="badge">
+                <span>3</span>
+              </RadioCardSlot>
```

The `code` string above the fixture needs the same wrapper, since it is what a reader copies.

The root cause is that `choiceCardSlot` is `buttonSlot`, yet `RadioCardSlot` and `CheckboxCardSlot` do not wrap the badge child the way `ButtonSlot` and `BadgeSlot` do. That divergence is pre-existing and independent of this change.

**Assessment**

- Severity: Low (verified, 90% confidence), because the defect is a wrong font size in one sandbox box, with no functional effect.
- Priority: P2 (estimate, 85% confidence), because nothing is blocked, but the box is new in this PR and its baseline is being recorded now.
- Breakage likelihood: Likely (verified, 95% confidence), because it is the rendered result today in every browser and color scheme: `16px` instead of `13px`.
- Fix effort: Easy (verified, 95% confidence), because the fixture fix is one wrapper element in the rendered example and its `code` string.

packages/ariakit-tailwind/src/input.ts
```diff
@@ -865,12 +865,17 @@ const root = rule(
   at.variant("contrast-more", set(vars.contrast, CONTRAST_HIGH)),
 );
 
-const disabled = rule(
-  ':where(:disabled, [disabled], [aria-disabled="true"])',
+utility(
+  "disabled",
   set(vars.contrastScale, vars.disabledContrastScale),
   set(vars.textContrastScale, vars.disabledTextContrastScale),
 );
```
diegohaz, line 872: `ak-disabled` ships without a counterpart that restores the normal contrast. Both properties it writes are registered with `inherits: true`, so a disabled ancestor lowers the contrast of every descendant, and an enabled descendant has no public way to opt back out. The `@apply ak-disabled` rule below gives the same reach to `:disabled`, `[disabled]` and `[aria-disabled="true"]`.

The case is now in this PR's own fixture. `<fieldset disabled>` matches that selector, but its first legend stays enabled, so the "Disabled fieldset" box renders an enabled button that inherits the disabled contrast settings. Measured in Chromium on `/react/previews/ariakit-ui-button/`, the legend button computes `--_ak-contrast-scale: 0.25` and `--_ak-text-contrast-scale: 0.8`, against `1` and `1` for another enabled button on the same page.

Under `prefers-contrast: more` that difference paints. The enabled legend badge computes `oklch(0.427841 0.1546 248.516)` where the same badge outside the fieldset computes `oklch(0.2 0.1546 248.516)`, so a high-contrast user gets a quarter of the requested boost on a control they can still operate:

Two identical enabled Actions buttons with a count badge under prefers-contrast more: the one outside a disabled fieldset has a deep navy badge, the one in the first legend of a disabled fieldset has a much lighter blue badge

To be clear about attribution: the inheritance and the `:where(:disabled, [disabled], [aria-disabled="true"])` rule are both pre-existing on `main`; the diff here only moves the two declarations into the utility. What is new is that the utility makes the scope a documented public API, and the fixture makes the gap concrete. A `readme.md` note, or a reset such as `ak-disabled-none` restoring the registered initial `1`, would close it.

**Assessment**

- Severity: Medium (verified, 75% confidence), because an operable control loses most of the contrast boost that `prefers-contrast: more` asked for, which is an accessibility regression for the user who set that preference.
- Priority: P2 (estimate, 70% confidence), because the behavior is pre-existing and narrow, and this PR neither introduces nor worsens it.
- Breakage likelihood: Rare (estimate, 80% confidence), because it needs an enabled control inside a disabled ancestor, which in practice means the first legend of `fieldset[disabled]`, together with `prefers-contrast: more` for the visible case.
- Fix effort: Moderate (estimate, 70% confidence), because a public reset utility needs a name, `readme.md` coverage, a changeset and a regression in the `ariakit-tailwind-7474` sandbox; a documentation-only note is Easy.

diegohaz (commented):
## Review summary

Initial review of the full change, baseline `082a3ea` (merge base) to `fd0ff44`, all 33 files, plus the surrounding recipes in `packages/ariakit-ui/src/styles` and the plugin pipeline in `packages/ariakit-tailwind/src/input.ts`.

The approach is right. Reusing the existing contrast channel instead of a new opacity treatment keeps the disabled look inside the adaptive pipeline, and mixing the slot fill before the ink is computed is the correct order. `@apply ak-disabled` is a behavior-preserving refactor of the plugin rule, not a rewrite, which is the cheapest way to make the scope public.

### Findings

1. [`packages/ariakit-ui/src/styles/control.ts:303`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3995425692) (reply on the open CodeRabbit thread): the three new mix signals have only one matching ink rule. A painted slot below the control's direct children dims its fill but keeps full-strength ink when the control is disabled through the attribute. **Address now**; a two-line fix is validated in the reply.
2. [`app/src/sandbox/ariakit-ui-radio/index.react.tsx:624`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997181250): the new `RadioCardSlot $kind="badge"` passes a bare text node, so the count renders at `16px` instead of `13px`. `RadioCardSlot` does not insert the wrapper that `ButtonSlot` and `BadgeSlot` insert.
3. [`packages/ariakit-tailwind/src/input.ts:868`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997181509): `ak-disabled` ships without a way to opt back out of the inherited contrast settings. Pre-existing behavior, but the new public utility and the new fieldset fixture make it a documented scope with no escape hatch.

### Existing feedback validated against this head

| Thread | Outcome | Disposition |
| --- | --- | --- |
| `PRRT_kwDOBjT0Fc6ht_9z` (CodeRabbit, `control.ts:303`, open) | Confirmed, with a corrected trigger: the break is a non-direct-child slot, not an `aria-disabled` badge | Address now, replied on the thread |
| `PRRT_kwDOBjT0Fc6htL5U` (Codex, resolved) | The surface half is fixed and verified across the script, fieldset and card-group paths | Invalid or obsolete for the surface; the ink residue is folded into the thread above, so the thread was not reopened |
| `PRRT_kwDOBjT0Fc6htMSq` (CodeRabbit, resolved, withdrawn by its author) | Agreed: the `ariakit-tailwind-7474` test already compares `background-color` for the badge | Invalid or obsolete |
| `PRRT_kwDOBjT0Fc6htYDW` (`opacity-50` on slots, resolved) | Verified: slot surfaces now use `ak-layer-mix-20`, and `opacity-50` survives only on avatar `` pixels, which no adaptive utility can express | Invalid or obsolete |

No follow-up issues were registered from this run. The Codex connector review body is boilerplate, and both CodeRabbit review bodies only restate their inline findings, so the top-level bodies added no separate claims.

### Verified

- **Regression proof.** Reverting `packages/ariakit-ui/src/styles/control.ts` to `082a3ea` makes all six new disabled tests fail in Chromium; restoring it makes them pass. The worktree was restored and is clean.
- **Assertions are not vacuous.** The fieldset test compares the disabled avatar against the enabled legend badge, which is only meaningful because both compute `oklch(0.515341 0.1546 248.516)` when enabled. Confirmed by reading computed styles.
- **`@apply` is behavior preserving.** Compiling `src/output.css` emits exactly the two former declarations inside `:where(:disabled, [disabled], [aria-disabled="true"])`, with no wrapper and no reordering, on both `tailwindcss@4.3.3` (the devDependency) and `tailwindcss@4.0.3` (the floor of the `^4.0.3` peer range). Nothing else in the plugin writes `--_ak-contrast-scale` or `--_ak-text-contrast-scale`, so moving the source into `@layer utilities` creates no cascade conflict.
- **Generated output is in sync.** `pnpm -F @ariakit/tailwind build` reproduces the committed `output.css` with an empty diff.
- **Sandbox wiring.** `app/src/sandbox/ariakit-tailwind-7474/style.css` is identical to the other focused sandboxes, and `app/src/styles/global.css:13` already excludes the new directory through `@source not "../sandbox/ariakit-tailwind*"`.
- **Changeset.** `200-` prefix for a minor change, `patch` frontmatter for a `v0` package, a title that is not a sentence, a `tsx` example, and plain inline code for the utility name, which matches the other `@ariakit/tailwind` entries. `@ariakit/ui` is private and listed in the `ignore` array, so it correctly gets no entry. Issue #7474 was opened by `diegohaz`, whom the credit policy excludes, so no `Thanks to` sentence is required.
- **Commands run.** `pnpm -F @ariakit/tailwind build`; `APP_PORT=… pnpm -F app test-chrome ariakit-tailwind-7474`; `APP_PORT=… pnpm -F app test-chrome ariakit-ui-button ariakit-ui-radio -g "dims|gives"`; the same suite against the reverted recipe and against a local prototype of the finding 1 fix; Chromium probes of computed styles on both sandbox previews at `no-preference` and `more` contrast.

### Checked and not reported

- **`opacity-50` on avatar images.** It is the one value in the change that does not follow the contrast preference, but raster pixels cannot go through `color-mix`, and the code comment states the constraint. Not a finding.
- **`ariakit-tailwind-7474` as a directory name.** The fixture lifecycle permits an issue-number suffix while a reproduction is isolated, so the semantic rename is optional, not required.
- **`test-browser.ts` for the new button and radio tests.** They assert computed `oklch` and `color-mix` output, which differs between engines, so the three-project cost is justified rather than a fallback.
- **`image` shadowed inside the `for (const [button, slot, image] of controls)` loop** in `app/src/sandbox/ariakit-ui-button/test-browser.ts`, where the outer `image` is an article scope and the inner one is an `` locator. Readability only, below the reporting bar.

### Not fully validated

- Firefox and WebKit were not exercised locally; the CI `App / Test Firefox` and `App / Test Safari` jobs pass on `fd0ff44`.
- Visual screenshots were not captured locally, per the repository policy. The baselines in this PR come from CI.
- `pnpm tsc`, `pnpm lint-fix`, `pnpm build` and `pnpm test` were not run locally; the corresponding CI jobs pass on this head.
- Performance was not measured locally. `Performance / Chrome (ariakit-tailwind)` passes, but that sandbox does not compile the `@ariakit/ui` recipes, so the six class strings added to every painted slot are not covered by any current measurement. I have no evidence of a regression, only no evidence either way.
- The unification direction sketched at the end of the finding 1 reply, a single `--control-disabled` channel read by one `@container style(...)` variant, was not prototyped.

One prose note, tied to finding 1: "Slots follow native and ARIA disabled states changed by scripts, disabled fieldsets, and choice cards disabled through their groups" in the description holds for the fill but not yet for the ink of a nested slot. Resolving finding 1 makes the sentence accurate as written.

app/src/sandbox/ariakit-ui-radio/index.react.tsx
diegohaz: We should do this in the radio/checkbox framework components like other components do (when the slot is badge).

diegohaz: Fixed in [`a9662c2fd9214dc4d53978f0a39fc39de53f333e`](https://github.com/ariakit/ariakit/commit/a9662c2fd9214dc4d53978f0a39fc39de53f333e): `RadioCardSlot` and `CheckboxCardSlot` now wrap badge children like `ButtonSlot`, so `<RadioCardSlot $kind="badge">3</RadioCardSlot>` receives the compact text size.

The same commit adds targeted browser coverage for both components. Both tests fail at `16px` before the fix and pass at `13px` afterward. The [updated preview](https://github.com/ariakit/ariakit/pull/7490) includes before-and-after images for both card badges.

packages/ariakit-tailwind/src/input.ts
diegohaz: Fixed in [`cc6d5f0ad3583f838c952e64562e161bd628fe7e`](https://github.com/ariakit/ariakit/commit/cc6d5f0ad3583f838c952e64562e161bd628fe7e): documented that enabled descendants also inherit the disabled contrast settings, including controls in the first legend of a disabled fieldset and `prefers-contrast: more`. The README also states that there is no public reset utility.

This takes the documentation option in the review. The inheritance and automatic disabled selector already exist on `main`; their behavior remains unchanged. The browser probe confirms the contrast difference, while the fieldset test confirms that the first-legend control stays enabled.

diegohaz:
**Fixed**

- [`a9662c2fd`](https://github.com/ariakit/ariakit/commit/a9662c2fd9214dc4d53978f0a39fc39de53f333e) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997294674)): Wrap badge children in the radio and checkbox card components so counts use compact text. Both browser regressions fail before the fix and pass afterward.
- [`cc6d5f0ad`](https://github.com/ariakit/ariakit/commit/cc6d5f0ad3583f838c952e64562e161bd628fe7e) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997295109)): Document that enabled descendants, including the first legend of a disabled fieldset, inherit reduced contrast. This follows the documentation option in the review; the existing behavior remains unchanged.

The [PR preview](https://github.com/ariakit/ariakit/pull/7490) includes the card badge comparisons.

packages/ariakit-ui/src/styles/control.ts
```diff
@@ -294,8 +294,23 @@ export const controlSlot = cv({
...
+      // Slots can sit below a content wrapper, beyond the control's direct
+      // child ink rules.
       "group-[.disabled]/control:ak-ink-0",
+      "group-ui-disabled/control:ak-ink-0",
+      "group-ui-disabled-within/choice:ak-ink-0",
+      // Images cover the layer. Fade only their pixels into it; fading the
+      // whole slot would also reduce its text's adaptive contrast.
+      "group-[.disabled]/control:[&_img]:opacity-50",
+      "group-ui-disabled/control:[&_img]:opacity-50",
+      "group-ui-disabled-within/choice:[&_img]:opacity-50",
```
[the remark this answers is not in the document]
diegohaz: We don't need to support this unless it's a very simple fix (like one line).

packages/ariakit-ui/src/styles/control.ts
```diff
@@ -294,8 +294,23 @@ export const controlSlot = cv({
...
+      "group-[.disabled]/control:ak-layer-mix-20",
+      // Native and ARIA states can change without rerendering the recipe.
+      "group-ui-disabled/control:ak-layer-mix-20",
+      // A choice card reads disabled state from the input inside its label.
+      "group-ui-disabled-within/choice:ak-layer-mix-20",
+      // Slots can sit below a content wrapper, beyond the control's direct
+      // child ink rules.
       "group-[.disabled]/control:ak-ink-0",
+      "group-ui-disabled/control:ak-ink-0",
+      "group-ui-disabled-within/choice:ak-ink-0",
```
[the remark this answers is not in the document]
diegohaz: Is this really an issue? Do we support this kind of composition?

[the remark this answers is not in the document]
diegohaz: Fixed in [`7f5ecd2efe1238fbb9c4011f4f98f023b0a972a3`](https://github.com/ariakit/ariakit/commit/7f5ecd2efe1238fbb9c4011f4f98f023b0a972a3): choice cards now apply `ak-disabled` on the label when their input is disabled.

Yes, this composition is supported. `RadioCard` and `CheckboxCard` pass `aria-disabled` to the Ariakit input, and `disabledFromProps` uses it to disable the control. The label only mirrored the `disabled` prop, so `<CheckboxCard aria-disabled>` missed the inherited contrast settings.

The new radio and checkbox regressions fail before the fix and pass afterward in Chromium, Firefox, and WebKit, with both themes and contrast preferences. They cover the card, label, and badge ink, plus script-driven input changes. The [updated PR preview](https://github.com/ariakit/ariakit/pull/7490) includes the comparison and a short recording.

packages/ariakit-ui/src/styles/control.ts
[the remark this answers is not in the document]
diegohaz: Not changing this in the current PR because preserving arbitrary custom image opacity is outside the requested scope. There is no simple one-line change here that composes with both existing opacity and existing filters; replacing opacity with a filter would move the conflict to custom filters. This follows the scope in your reply and adds no new composition contract.

diegohaz:
**Fixed**

- [`7f5ecd2ef`](https://github.com/ariakit/ariakit/commit/7f5ecd2efe1238fbb9c4011f4f98f023b0a972a3) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997557766)): Give input-disabled radio and checkbox cards the same contrast as prop-disabled cards. `<CheckboxCard aria-disabled>` is supported; both new regressions fail before the fix and pass in all three browser engines afterward. The [PR preview](https://github.com/ariakit/ariakit/pull/7490) includes the comparison and recording.

**Declined**

- [`Custom image opacity`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997497991) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997557917)): Leave arbitrary opacity composition outside this fix, as requested. A filter replacement would require rules for existing filters too.

packages/ariakit-ui/src/styles/choice.ts
```diff
@@ -196,6 +196,7 @@ export const choiceCard = cv({
     // it reads the state from its input and draws what $disabled below draws.
     // The label itself is never disabled, so its hover and press, which only
     // skip a disabled element, are turned off here too.
+    "ui-disabled-within:ak-disabled",
```
diegohaz, line 201: `choiceField` needs the same line. It carries the identical treatment and the identical reasoning ("The input is the only part that is ever `:disabled`, so the row reads its state from it as well as from the `$disabled` prop"), but only `ak-ink-0`:

```ts
// choice.ts, choiceField
"ui-disabled-within:cursor-not-allowed",
"ui-disabled-within:ak-ink-0 ui-disabled-within:**:ak-ink-0",
```

`RadioField` mirrors the input's disabled state exactly the way `RadioCard` does, with `$disabled: variantProps.$disabled ?? rest.disabled`, so `<RadioField aria-disabled>`, `<CheckboxField aria-disabled>`, and a script disabling the input all reach the row through CSS only. Before this PR every path was equally undimmed, so there was nothing to diverge; now the card is consistent with native and the row is not.

Measured in Chromium on `/react/previews/ariakit-ui-radio/`, on the row's label text:

| Row | `--_ak-text-contrast-scale` | `color` |
| --- | --- | --- |
| `RadioField` with `disabled` (the `$disabled` prop path) | `0.8` | `oklch(0 0 0 / 0.433475)` |
| `RadioField` with `aria-disabled` on its input | `1` | `oklch(0 0 0 / 0.541843)` |

Those are the same two numbers that issue #7474 names in its second problem ("the label computes `oklch(0 0 0 / 0.542)` with `--_ak-contrast-scale: 1`. The native disabled button computes `oklch(0 0 0 / 0.433)`"), so this is the reported defect still present on the field row.

Two identical disabled RadioField rows side by side: the one disabled through the prop has lighter label and description text than the one disabled through aria-disabled on its input

```diff
     // The input is the only part that is ever :disabled, so the row reads
     // its state from it as well as from the $disabled prop, which a radio
     // disabled through its group never receives. The description paints its
     // own ink, so the dimming has to reach every descendant.
+    "ui-disabled-within:ak-disabled",
     "ui-disabled-within:cursor-not-allowed",
     "ui-disabled-within:ak-ink-0 ui-disabled-within:**:ak-ink-0",
```

With that line the `aria-disabled` row measures `oklch(0 0 0 / 0.433475)`, identical to the prop path, and the 35 tests in the button, radio and checkbox sandboxes still pass in Chromium.

The `input` recipe has the third instance of the same shape: `ui-field-disabled:ak-ink-0 ui-field-disabled:**:ak-ink-0` with no `ak-disabled`. There the gap is limited to the `:has()` branch, a wrapper whose entry controls are disabled, because a wrapper that is itself `:disabled` or `[aria-disabled]` already gets the settings from the plugin rule. No fixture exercises that today, so I did not measure it.

**Assessment**

- Severity: Medium (verified, 85% confidence), because a disabled field row reached through ARIA or a script keeps stronger text than every other disabled control, which is the defect this PR exists to remove.
- Priority: P1 (estimate, 75% confidence), because the same change makes the card path correct and leaves the sibling row path wrong in the same file, which is harder to notice later than now.
- Breakage likelihood: Possible (verified, 90% confidence for the mechanism; estimate for frequency), because it needs `aria-disabled`, a group-driven state, or a script rather than the `disabled` prop. `RadioFieldProps extends ak.RadioProps`, so `aria-disabled` is a supported prop on the row.
- Fix effort: Easy (verified, 95% confidence), because it is one class string; I applied it locally and re-ran the three sandboxes.

app/src/sandbox/ariakit-tailwind-7474/test-chrome.ts
```diff
@@ -0,0 +1,55 @@
...
+          }
```
diegohaz, line 54: **Validating the CodeRabbit claim that could not be posted inline** ([review 5187736642](https://github.com/ariakit/ariakit/pull/7490#pullrequestreview-5187736642), "Assert the disabled badge surface against an enabled badge"). It has no thread, so there is nowhere to reply; the outcome is **confirmed**, with a smaller fix than the one proposed.

This sandbox exists to prove that `ak-disabled` reproduces the attribute's effect, including on descendants. The loop above proves that for the elements that carry the class, across `color`, `background-color` and `border-top-color`. This loop proves it for descendants only through `color`, so a regression that stopped the utility from reaching a descendant **layer** would pass.

The descendant background does depend on the inherited setting. On `/react/previews/ariakit-tailwind-7474/` under `prefers-contrast: more`, overriding `--_ak-contrast-scale: 1` on `Upload badge` alone moves its background from `oklch(0.407575 0.245 262.881)` to `oklch(0.2 0.245 262.881)` while `Native badge` stays put. Today the two badges match in all four colour-scheme and contrast combinations, so comparing the property is stable:

| Contrast | `Native badge` background | `Upload badge` background |
| --- | --- | --- |
| `no-preference` | `oklch(0.495075 0.245 262.881)` | `oklch(0.495075 0.245 262.881)` |
| `more` | `oklch(0.407575 0.245 262.881)` | `oklch(0.407575 0.245 262.881)` |

Adding an enabled badge to the fixture, as the bot proposed, also works, but the comparison that matters here is utility against attribute rather than disabled against enabled, and the loop already has both operands:

```diff
-          for (const name of ["badge", "colored text"]) {
-            const color = await q
-              .text(`Native ${name}`)
-              .evaluate((element) => getComputedStyle(element).color);
-            await test
-              .expect(q.text(`Upload ${name}`))
-              .toHaveCSS("color", color);
+          for (const name of ["badge", "colored text"]) {
+            for (const property of ["color", "background-color"]) {
+              const expected = await q
+                .text(`Native ${name}`)
+                .evaluate(
+                  (element, property) =>
+                    getComputedStyle(element).getPropertyValue(property),
+                  property,
+                );
+              await test
+                .expect(q.text(`Upload ${name}`))
+                .toHaveCSS(property, expected);
+            }
           }
```

`colored text` is `ak-text`, so its background is transparent in both, which makes that half a cheap no-op rather than a wrong assertion.

**Assessment**

- Severity: Low (verified, 90% confidence), because this is missing coverage rather than a defect; the descendant background is correct today.
- Priority: P2 (estimate, 75% confidence), because the sandbox is the designated proof for a newly public utility and the gap is in the half of its contract that the PR description states.
- Breakage likelihood: Rare (estimate, 70% confidence), because it takes a future change to the inheritance of `--_ak-contrast-scale` or `--_ak-text-contrast-scale` to expose it, and only under `prefers-contrast: more`.
- Fix effort: Easy (verified, 90% confidence), because the loop already resolves both elements.

**Disposition: Address now.** The changed contract is that descendant layers inherit the settings, and this sandbox is where that claim is proven.

app/src/sandbox/ariakit-ui-checkbox/test-browser.ts
```diff
@@ -8,6 +8,65 @@ import {
 } from "#app/test-utils/ariakit-ui.ts";
 
 withCaptures(import.meta.dirname, async ({ query, test }) => {
+  // https://github.com/ariakit/ariakit/issues/7474
+  // https://github.com/ariakit/ariakit/pull/7490#discussion_r3997200806
+  test("uses compact text for a card badge", async ({ q }) => {
+    const badge = query(q.article("Card badge")).text("3");
+    await test.expect(badge).toHaveCSS("font-size", "13px");
+  });
```
diegohaz: This assertion runs in Chrome, Firefox and Safari for a contract that no engine can disagree about: `*:text-[0.8125em]` against a `16px` parent is `13px` everywhere. `ariakit-ui-checkbox/test-chrome.ts` already exists beside this file, so moving the test costs nothing.

The same applies to the copy in `app/src/sandbox/ariakit-ui-radio/test-browser.ts:12`, where the sandbox has no `test-chrome.ts` yet.

The neighbouring tests in this file earn their three projects: `matches native card contrast for ARIA and script disabled states` compares computed `oklch` and `color-mix` output, which is exactly the kind of result that differs between engines.

**Assessment**

- Severity: Low (verified, 90% confidence), because the only cost is CI time; the assertion itself is correct.
- Priority: P2 (estimate, 85% confidence), because nothing is blocked and the waste is two fast assertions per extra engine.
- Breakage likelihood: Rare (estimate, 80% confidence), because the risk is a flaky Safari or Firefox run failing a test that never needed those projects.
- Fix effort: Easy (verified, 90% confidence), because it is a move into the existing `test-chrome.ts` for checkbox and a new one-test file for radio.

diegohaz (commented):
## Re-review summary

Re-review. Incremental baseline `fd0ff44`, the head of my previous marked review, to `dd94968`. The PR also moved its merge base from `082a3ea` to `20053a2`, so I re-read the whole change against the new base and confirmed the merges carried nothing unintended: the PR diff against `20053a2` is exactly the 17 disabled-related files plus baselines, with no combobox, tabs or dependency edits left in it.

All three findings from the previous review were addressed, and `d593fc3` and `7f5ecd2` both come with regressions that fail without their fix. The remaining work is one instance of the same pattern that `7f5ecd2` fixed, one place where the missing cell now sits.

### Findings

1. [`packages/ariakit-ui/src/styles/choice.ts:199`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997782168): `choiceField` did not get the `ui-disabled-within:ak-disabled` line that `choiceCard` just got, so `<RadioField aria-disabled>`, `<CheckboxField aria-disabled>` and a script-disabled field input still measure `oklch(0 0 0 / 0.541843)` against `oklch(0 0 0 / 0.433475)` for the `disabled` prop. Those are the two numbers issue #7474 names. One-line fix, validated.
2. [`app/src/sandbox/ariakit-tailwind-7474/test-chrome.ts:39`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997782368): confirming CodeRabbit's claim that could not be posted inline. Descendants are compared only for `color`, so a regression that stopped `ak-disabled` reaching a descendant layer would pass. **Address now**, with a smaller fix than the bot proposed.
3. [`app/src/sandbox/ariakit-ui-checkbox/test-browser.ts:13`](https://github.com/ariakit/ariakit/pull/7490#discussion_r3997782559): the two `uses compact text for a card badge` tests assert an engine-independent `font-size` from `test-browser.ts`, and `test-chrome.ts` already exists beside one of them.

### Previous findings, verified on this head

| Finding | Commit | Outcome |
| --- | --- | --- |
| Nested slot ink (`PRRT_kwDOBjT0Fc6ht_9z`) | `d593fc3` | Fixed as suggested, plus a `Nested badge and avatar` fixture. Removing the two ink lines fails `dims nested slot ink when a script changes disabled` and its `aria-disabled` twin in Chromium; restoring them passes. |
| Radio card badge wrapper (`PRRT_kwDOBjT0Fc6hyjY8`) | `a9662c2` | Fixed one level deeper than I proposed: `RadioCardSlot` and `CheckboxCardSlot` now wrap the badge child like `ButtonSlot` and `BadgeSlot`, which removes the divergence rather than patching the fixture. `13px` confirmed for both card badges, and the radio surface assertion correctly moved to `.locator("..")` now that the slot has an element child. |
| Inherited disabled contrast (`PRRT_kwDOBjT0Fc6hyjbr`) | `cc6d5f0` | Documented in `readme.md`, which was the option I offered. The paragraph states the first-legend case and that no public reset exists, so the contract is now honest about its scope. |

### Other feedback validated during this run

| Source | Outcome | Disposition |
| --- | --- | --- |
| Codex `PRRT_kwDOBjT0Fc6hy3RW`, card contrast from within | Confirmed and fixed in `7f5ecd2`. Removing `ui-disabled-within:ak-disabled` fails both new `matches native card contrast for ARIA and script disabled states` tests in Chromium. | Already resolved by the author |
| Codex `PRRT_kwDOBjT0Fc6hy3RY`, avatar image opacity below 50% | Technically accurate: the group-variant rule wins on specificity and would raise a lower opacity to `0.5`. | Declined non-goal, as the author recorded; no adaptive utility fades raster pixels, and a filter swap moves the same conflict onto filters |
| CodeRabbit review body, "a disabled badge can return to its full-strength fill without the browser regression detecting it" | Invalid against this head. `dims badge and avatar surfaces in disabled controls`, both script tests, and the radio card-grid test all assert `background-color` against an enabled reference. | Invalid or obsolete |
| CodeRabbit review body, docstring coverage 33.33% | The functions it counts are `RadioCardSlot` and `CheckboxCardSlot`, which already carry JSDoc, and the rest are sandbox fixtures. | Declined non-goal |
| CodeRabbit outside-diff comment on `ariakit-tailwind-7474/test-chrome.ts` | Confirmed; published as finding 2 because the claim has no thread to reply on. | Address now |

No follow-up issues were registered from this run.

### Verified

- **Regression proof for both new fixes.** Removing `"group-ui-disabled/control:ak-ink-0"` and `"group-ui-disabled-within/choice:ak-ink-0"` fails the two nested-ink tests; removing `"ui-disabled-within:ak-disabled"` fails the two card-contrast tests. Both were restored and the worktree is clean.
- **No regression from the finding 1 fix.** With `ui-disabled-within:ak-disabled` added to `choiceField`, all 35 tests across the button, radio and checkbox sandboxes pass in Chromium.
- **Merge integrity.** `control.ts` carries both this PR's disabled classes and the avatar branch from #7489, and the PR diff against the new merge base contains no file from the merged PRs.
- **CI on this head.** 30 check runs and 4 statuses on `dd94968`, all successful. `Performance / Chrome (ariakit-tailwind)` reports no significant change, which narrows but does not close the gap I noted last time, since that sandbox still does not compile the `@ariakit/ui` recipes.
- **Commands run.** `APP_PORT=… pnpm -F app test-chrome ariakit-ui-radio ariakit-ui-checkbox ariakit-ui-button`; the same suite filtered to `matches native card contrast` and `dims nested slot ink` against each reverted recipe; Chromium probes of computed styles and of a mutated `--_ak-contrast-scale` on the radio and `ariakit-tailwind-7474` previews at `no-preference` and `more` contrast.

### One note on the description

The `Card group` row of the Preview table pairs a dark-theme Before with a light-theme After, while every other row keeps one theme across the pair, so the difference the caption claims is not readable from the two images. The After cell also points at `4a800d39-114b-42e5-b6ad-43c2e9036f71`, the same asset the `Radio card badge` row uses for its own After, which looks like the card-group capture was overwritten when the review-fix table was added.

### Not fully validated

- Firefox and WebKit were not exercised locally; the CI `App / Test Firefox` and `App / Test Safari` jobs pass on `dd94968`.
- Visual screenshots were not captured locally, per repository policy.
- `pnpm tsc`, `pnpm lint-fix`, `pnpm build` and `pnpm test` were not run locally; the corresponding CI jobs pass on this head.
- The third instance of the missing `ak-disabled`, the `:has()` branch of `ui-field-disabled` in `input.ts`, was reasoned about but not measured, because no fixture puts the class on a wrapper around a disabled entry control.

packages/ariakit-ui/src/styles/choice.ts
diegohaz: Fixed in [`fe09c8443`](https://github.com/ariakit/ariakit/commit/fe09c844346cf7ed065207194d769a453946a682): choice fields and input wrappers now apply the existing `ak-disabled` settings when their inputs supply the disabled state. `<RadioField aria-disabled>` and `<CheckboxField aria-disabled>` match the native-disabled label and description; script changes also update wrapper and icon contrast.

The new regressions fail before these two recipe classes and pass afterward in Chromium, Firefox, and WebKit, across both themes and contrast preferences. They also check that removing each attribute restores the enabled colors. The [field and wrapper before/after images and toggle recording](https://github.com/ariakit/ariakit/pull/7490#issue-5431157018) show the result.

app/src/sandbox/ariakit-tailwind-7474/test-chrome.ts
diegohaz: Resolved in [`9fb60990b`](https://github.com/ariakit/ariakit/commit/9fb60990b6f4e6b9005c4c2cf53130fdc17a228b): `ak-disabled` already gives descendant badges the same background as the native-disabled reference. Added targeted coverage in [`1f48a45ef`](https://github.com/ariakit/ariakit/commit/1f48a45ef9e42edd5539718cd2c12bbf68dfce17).

The assertion compares badge `background-color` in all four theme/contrast combinations. Colored text keeps its ink assertion. A temporary background-only mutation fails all four new checks while the existing text-color checks pass. The [descendant background preview](https://github.com/ariakit/ariakit/pull/7490#issue-5431157018) shows the current matching surfaces.

app/src/sandbox/ariakit-ui-checkbox/test-browser.ts
diegohaz: Fixed in [`327c2068f`](https://github.com/ariakit/ariakit/commit/327c2068f75f624dc5f151aee752e0e554337d32): moved both compact badge-text checks to `test-chrome.ts`, with the same fixtures and `13px` assertions. The full browser run collects each check only in Chromium; computed-color checks still run in all three engines.

diegohaz:
**Fixed**

- [`fe09c8443`](https://github.com/ariakit/ariakit/commit/fe09c844346cf7ed065207194d769a453946a682) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3998711054)): field rows and input wrappers now match native-disabled contrast, including ARIA and script changes. The new regressions pass in all three engines. [Before/after previews and recording](https://github.com/ariakit/ariakit/pull/7490#issue-5431157018).
- [`327c2068f`](https://github.com/ariakit/ariakit/commit/327c2068f75f624dc5f151aee752e0e554337d32) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3998711616)): both compact badge-text checks now run only in Chromium, with their original assertions.

**Already resolved**

- [`1f48a45ef`](https://github.com/ariakit/ariakit/commit/1f48a45ef9e42edd5539718cd2c12bbf68dfce17) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r3998711470)): added direct badge-background coverage for the existing native/utility match. A background-only mutation fails the new assertion in all four theme/contrast combinations. [Matching surfaces](https://github.com/ariakit/ariakit/pull/7490#issue-5431157018).

app/src/sandbox/ariakit-ui-checkbox/test-browser.ts
```diff
@@ -8,6 +8,120 @@ import {
...
+        );
+      });
+    }
+  });
+
+  // https://github.com/ariakit/ariakit/pull/7490#discussion_r3997782168
+  test("matches native field contrast for ARIA and script disabled states", async ({
+    page,
+    q,
+  }) => {
```
diegohaz, line 51: This block and the card block above it are now duplicated verbatim in the radio sandbox. Diffing the two sandboxes:

- `matches native field contrast…`: 60 lines each, differing on 2 lines (`box.checkbox(...)` against `box.radio(...)`).
- `matches native card contrast…`: 50 lines each, differing on the same 2 lines.

That is 110 lines of identical state-toggling logic kept in step by hand across two files, and the two copies have already drifted: the card version ends with a negative `not.toHaveCSS("color", disabledColor)`, while the field version captures the enabled colors up front and asserts the exact values come back. The field version is the stronger check, so the card path currently has the weaker one for no reason.

`app/src/test-utils/ariakit-ui.ts` is the established home for this ("Helpers for the tests of the ariakit-ui-\* sandboxes") and already exports small shared assertions such as `expectFocusVisible`, `hoverOver` and `expectMedia`. Each call site keeps what is actually specific to it, its `pairs`, and the toggling loop moves once:

```ts
// app/src/test-utils/ariakit-ui.ts
export async function expectDisabledContrastFollowsInput(
  input: Locator,
  pairs: readonly (readonly [Locator, Locator])[],
) { /* capture enabled colors, toggle aria-disabled then disabled, restore */ }
```

```ts
// each sandbox, after building pairs
await test.expect(input).toBeDisabled();
await expectDisabledContrastFollowsInput(input, pairs);
```

I applied this locally across all four tests: 130 lines removed from the two sandboxes, 46 added to the helper, net 72 fewer lines. All four tests pass in Chromium, and all four still fail when `ui-disabled-within:ak-disabled` is removed from `choice.ts`, so the regression proof survives the move. Adopting the field version's round trip for the card tests is what closes the drift.

The `ariakit-ui-input` wrapper test stays where it is: it compares one wrapper and its icon against a different reference and already uses that file's local `setDisabled` helper, so it does not fit the same shape.

**Assessment**

- Severity: Low (verified, 90% confidence), because the tests are correct today; the cost is maintenance and the drift that has already appeared between the two copies.
- Priority: P2 (estimate, 80% confidence), because nothing is blocked, but a third choice component or a fourth state would copy the block again.
- Breakage likelihood: Possible (verified, 85% confidence), because the divergence between the card and field round trips is present in this PR and is exactly the failure mode duplication produces.
- Fix effort: Easy (verified, 95% confidence), because I ran the extraction end to end, including the reverted-recipe check.

diegohaz (commented):
## Re-review summary

Re-review. Incremental baseline `dd94968`, the head of my previous marked review, to `f4f5b6f0`. The merge base moved again, from `20053a2` to `77afdf1`, taking in the screenshot row-group split, so I re-read the whole change against the new base: 21 source files plus baselines, with nothing from the merged PRs left in the diff.

All three findings from the previous round are fixed, each with a regression I confirmed fails without it, and the `input.ts` instance I could only reason about last time was measured and fixed too. One finding remains, and it is about the tests rather than the behaviour.

### Findings

1. [`app/src/sandbox/ariakit-ui-checkbox/test-browser.ts:64`](https://github.com/ariakit/ariakit/pull/7490#discussion_r4000754343): the card and field contrast tests are now 110 lines duplicated verbatim between the radio and checkbox sandboxes, differing on two lines each, and the two copies have already drifted in how they assert the return to the enabled state. Extraction into `app/src/test-utils/ariakit-ui.ts` is applied and verified locally, including the reverted-recipe check.

### Previous findings, verified on this head

| Finding | Commit | Outcome |
| --- | --- | --- |
| `choiceField` missing `ak-disabled` | `fe09c844` | Fixed, and extended to the `input` recipe, which I had flagged as the third instance but could not measure. Removing `ui-disabled-within:ak-disabled` from `choiceField` and `ui-field-disabled:ak-disabled` from `input` fails all three new tests in Chromium: both `matches native field contrast for ARIA and script disabled states` and `matches native disabled contrast on an input wrapper`. The wrapper test also covers the leading icon, which is the part the `**:ak-ink-0` rule reaches. |
| Descendant badge background unverified | `1f48a45` | Fixed. `background-color` is now compared for the badge and skipped for colored text, which is right: `ak-text` paints no background, so asserting it there would be a no-op rather than coverage. |
| Compact badge text in `test-browser.ts` | `327c206` | Fixed. Both assertions moved to `test-chrome.ts`, with a new one-test file for the radio sandbox, and the computed-colour tests correctly stayed in all three projects. |

No new feedback arrived from either bot on these commits, and every review thread on the PR is resolved.

### Verified

- **Regression proof.** Removing the two new recipe lines fails the three new field and wrapper tests; restoring them passes. All 33 tests across the radio, checkbox, input and `ariakit-tailwind-7474` sandboxes pass in Chromium on this head.
- **The extraction keeps the proof.** With the shared helper in place and the card tests switched to the field version's round trip, all four tests still fail when `ui-disabled-within:ak-disabled` is removed from `choice.ts`.
- **Baselines.** The screenshot changes are coherent with the row-group split from `main`: six new button `rows-25-25` images, six new checkbox `rows-10-10`, and radio `rows-7-8` replaced by `rows-7-9` as the new examples extend the grid. No `ariakit-ui-input` baseline changed, which is correct: `ui-field-disabled:ak-disabled` is a no-op where the class sits on a `:disabled` input, and no existing fixture puts it on a wrapper around one.
- **CI on this head.** 30 check runs on `f4f5b6f0`, all successful, and no non-passing row in `gh pr checks`.
- **Commands run.** `APP_PORT=… pnpm -F app test-chrome ariakit-ui-radio ariakit-ui-checkbox ariakit-ui-input ariakit-tailwind-7474`; the same suite filtered to the new tests against the two reverted recipe lines; the same filtered suite against a local prototype of the extraction, both with and without the recipe lines; a mutated `--_ak-contrast-scale` on the `ariakit-tailwind-7474` badge.

### One note on the new descendant-background check

I could not reproduce a mutation that fails it without the existing ink check failing first. Setting `--_ak-contrast-scale: 1` on `Upload badge` fails at `color` (`oklch(1 0 0)` against the expected `oklch(1 0 0 / 0.776987)`) before the run reaches `background-color`, because the layer and the ink both read `--_ak-ct`. The two are coupled from the fixture side, so the added assertion earns its place only against a plugin-internal regression, where the layer stops reading `--_ak-ct` and the ink keeps doing so. That is a real class and worth covering; I am noting it so the check is not mistaken for independent coverage of the fixture-level case.

### Carried over from the previous summary

The `Card group` row of the Preview table still pairs a dark-theme Before with a light-theme After, and its After cell still points at `4a800d39-114b-42e5-b6ad-43c2e9036f71`, the same asset the `Radio card badge` row uses for its own After. The other rows keep one theme across the pair.

### Not fully validated

- Firefox and WebKit were not exercised locally; the CI `App / Test Firefox` and `App / Test Safari` jobs pass on `f4f5b6f0`.
- Visual screenshots were not captured locally, per repository policy.
- `pnpm tsc`, `pnpm lint-fix`, `pnpm build` and `pnpm test` were not run locally; the corresponding CI jobs pass on this head.
- Performance was not measured locally. `Performance / Chrome (ariakit-tailwind)` still does not compile the `@ariakit/ui` recipes, so the class strings added to every painted slot remain outside any current measurement, in either direction.

app/src/sandbox/ariakit-ui-checkbox/test-browser.ts
diegohaz: Fixed in [`1a87fc1cc`](https://github.com/ariakit/ariakit/commit/1a87fc1cc6d7c7554f722b72cd6f4e0c0d861d85): the four radio/checkbox card and field tests now share `expectDisabledContrastFollowsInput`. Each sandbox keeps its input and comparison pairs. The card tests also require the enabled card color to differ from its disabled color. Every pair must return to its exact enabled color after either disabled attribute is removed; the input-wrapper test stays local.

All four tests still fail when the choice contrast rule is removed. Both card tests also fail when the disabled color persists or an incorrect enabled color is injected after re-enabling. All 69 selected checks pass in Chromium, Firefox, and WebKit. The [card and field previews](https://github.com/ariakit/ariakit/pull/7490#issue-5431157018) show the tested behavior; the Card group comparison now uses matching dark themes.

diegohaz:
**Fixed**

- [`1a87fc1cc`](https://github.com/ariakit/ariakit/commit/1a87fc1cc6d7c7554f722b72cd6f4e0c0d861d85) ([reply](https://github.com/ariakit/ariakit/pull/7490#discussion_r4000886985)): Shared the four choice contrast checks and retained the card color-change check plus exact enabled-color restoration.

